PR #1162, opened 2014-08-18
Multi Source / Multi Target routing

felixguendling:
This branch contains code to enable OSRM to do a "one to many" and "many to one" shortest path search.

For the one to many case:
The forward heap keeps the distances from the source and gets reused for each target.
The target heap keeps the distances from the target and gets cleared for each target.

The many to one case ist just the other way around (use all edges in the opposite direction).

Difference to many-to-many:
- The many-to-many search seems to label the "whole graph" (not really because of Contraction Hierarchies) with backward routing steps. This search is more like the regular shortest path search (alternating: backward, forward).
- The many-to-many search does not calculate distances (just times).

I'm new to OSRM and to the GitHub flow. I'm sorry if I got something wrong. I'm open for feedback.

Is the way, the distances are computed in RoutingAlgorithms/MultiTargetRouting.h (line 167 ff.) correct?
Is it correct to use IncrementalFindPhantomNodeForCoordinate here (instead of the non-incremental version)?


ZupoLlask:
Is this PR ready to be merged?

felixguendling:
No, it's not. I need to evaluate (*correctness*, performance) a version with the latest changes. If I understand emiltin and DennisOSRM correctly, they are more interested in a N:M routing, not a 1:N/N:1 routing (like it's implemented here).

DennisOSRM:
> they are more interested in a (more general) N:M routing, not a 1:N/N:1 routing (like it's implemented here).

Yes, and we need a solid implementation of the (generalized) Knopp et al. algorithm already used in the distance table computation.

felixguendling:
Okay. Then there should be a new PR for a N:M search. I don't think that it makes sense to include this feature here.

I evaluated the current 1:N/N:1 code. It's seems to be correct now (1000 results match the distance table results) and still outperforms the N:N (distance table) approach by a factor of about two (see graphics above from Sep 1, 2014). So for my own purposes, it's not reasonable to switch to the distance table approach.

[I accidentally merged the wrong commit in this PR. But since I'm closing it now, that doesn't matter anymore. I can provide a clean  implementation of the 1:N/N:1 approach somewhere else if someone is interested.]

DennisOSRM:
> I assume, you refer to this concept: http://algo2.iti.kit.edu/schultes/hwy/distTable.pdf

yes, that's the method.

> I evaluated the current 1:N/N:1 code. It seems to be correct now (1k results match the distance table results) and still outperforms the N:N (distance table) approach by a factor of about two

The interested reader should note that the 1:N method is only twice as fast while the full distance table computes on the order of N more entries. It's a bit comparing apples and oranges.

Thanks again for your work @felixguendling

emiltin:
but in cases where you only need to compute 1:n, then it's twice as fast as doing a n:m with n=1?

felixguendling:
> The interested reader should note that the 1:N method is only twice as fast while the full distance table computes on the order of N more entries. It's a bit comparing apples and oranges.

As I stated in the main evaluation post from Sep 1, 2014, I modified the distance table algorithm to stop after one source (respectively one target for N:1) by putting conditional 'break' statements in the calculation loops:

> I did a comparison of a version of the distance-table implementation which calculates only one row (respectively one column for comparison with many-to-one) of the distance table.

emiltin:
right. i think it could make sense to have this special case as a service, since it's faster than using the general request with n or m set to 1.

DennisOSRM:
> right. i think it could make sense to have this special case as a service, since it's faster than using the general request with n or m set to 1.

A proper implementation of the general algorithm will be faster.

felixguendling:
> A proper implementation of the general algorithm will be faster.

What's wrong with the distance table implementation? The concept seems to match the approach described in the paper. Adding a distance calculation and changing the loops so that they won't iterate over the same N targets (first loop over one 'targets' vector, the second over the 'sources' vector) won't change the performance characteristics.

I think one difference is that the N:M approach requires an additional data store ("buckets"). For the N:1/1:N case I can use (and reuse) the query heap of the "1" in "1:N"/"N:1".

DennisOSRM:
It's not wrong. It's a quite good implementation. Please don't get me wrong here.

Doing one-to-many (or many-to-one) queries with data from the heaps is alright. But we are looking for a general implementation and there is some tuning opportunity with a general implementation over the distance tables as well, e.g. always doing a (less entries)-to-(more entries) search with reversed direction if necessary and similar things. You already did some of these things. The point was that it makes sense to _properly_ build out a complete n:m search that combines all these features.

norcis:
Hi. I'm new at Project-OSRM and looking to try it with many-to-one calculations.
I can't do it as localhost library straightforward (as server code is in .NET) and REST calls (like /table?loc=54.676377,25.290076&loc=54.67720,25.28781&loc=54.676377,25.290076&z=14) is a preferred way.

Could you tell me the steps how to try it?

1. Download and compile from this pull (reference in wiki Building OSRM, Running OSRM)
2. ?

felixguendling:
I can provide a working version of my implementation. Currently, the source code is not publicly accessible because I deleted the forked repository. I am maintaining my modifications in a private Git. I can provide my implementation if you are interested.

norcis:
Thanks felixguendling, it would be great!

mantesat:
Felixguendling, I'd be interested in your implementation too. Can you also send it to me? Thanks in advance!

felixguendling:
@norcis @mantesat 

I'm sorry for answering late. I'm a bit busy these days.

The code is now online:
https://github.com/felixguendling/osrm-backend/tree/n2o-o2n
*("n2o-o2n" = N to one and one to N = multi source / multi target)*

Clone:

    git clone git@github.com:felixguendling/osrm-backend.git -b n2o-o2n

It's not based on the latest OSRM development branch but It should work (at least for me it does). It should be easy to extract the changes and throw them into a current release of OSRM or merge the Git branches.

Build process, dependencies, etc. are the same as for OSRM. You can stick to the official guide.

The new Multi Source and Multi Target plugins come with a REST API similar to that known from the Table Routing or Via Routing plugin. It takes a list of coordinates and uses the first coordinate as source (for multi-target routing) or target (for multi-source routing).

    http://localhost:5000/multitarget?loc=src_lat,src_lng&loc=target1_lat,target1_lng&loc=...
    http://localhost:5000/multitarget?loc=target_lat,target_lng&loc=src_lat,src_lng&loc=...

If you find any issues (related to the N:1/1:N implementation) feel free to open a ticket at the felixguendling/osrm-backend repo.

mantesat:
@felixguendling

Thanks for the code. However, I tried to build it as is, and it fails because it looks for the multi_target.cpp file in the plugins dir (I searched the file and found it in the Plugins (capital P) directory). Moving the file to the plugins dir helped complete the build, but when running the server all requests failed with an internal error.

I'm not familiar with the building process beyond following the basic instructions, so I'm kind of stuck on this. Do you have any suggestions? (I can open an issue to your gitg repo if that's preferable).

felixguendling:
I fixed the bug you mentioned. The Mac OS (as well as Windows) filesystem does not make a difference between upper case and lower case letters. So for me this source compiled.

I think it makes sense to move to another repository (open a new issue there). I don't want to disturb anybody here through spam mails concerning a dead pull request where off-topic issues are discussed (since the original functionality was rejected by project officials).

jornedeblaere:
@felixguendling
i'm intrested as well in your implementation. Is it possible to have a look as well?

felixguendling:
Clone https://github.com/felixguendling/osrm-backend/
Build & run instructions can be found in the wiki here.

jornedeblaere:
This is what i'm looking for, thanks for sharing! on the other hand, querying is limited to 250locations more or less, is there a work around to increase the limmit?

felixguendling:
Basically, the algorithms (N:1/1:N as well as via-route and distance table) are not limited to a certain number of locations (i.e. 250). I've tested the algorithm with >1000 locations using OSRM as a library in C++ code. I assume you mean the limit imposed by using HTTP GET requests for querying OSRM?

jornedeblaere:
yes indeed changing the limit for http requests, sinds now the buffer gets full around 250 locations in a single http get request

felixguendling:
I can't see where the server limits the number of locations. The HTTP server code does not limit the URI length. Can you please point to a specific code segment that does not allow for >250 locations?

jornedeblaere:
I tried it with wget. when i use 250 location it works, I tried using 300 locations and got a bad connection error

felixguendling:
Can you please check whether it is a client side or a server side problem and what exactly causes the restriction? Otherwise, I can recommend using OSRM as a library.

jornedeblaere:
Although my C++ skills are restricted to a bare minimum. I have a good feeling that it is a server side problem. Using the development branch of osrm you can now specify the max length of the matrix. I can do calls over 35k locations. But I really can’t isolate the problem because unfortunately I haven’t got the skills. On the other hand is it hard to compile the project as a library?

felixguendling:
Seems like nobody has a solution here (despite the problem seems to be fixed . Please open an issue (providing an example query) here:
https://github.com/felixguendling/osrm-backend/


> On the other hand is it hard to compile the project as a library?

It is not that hard. An example is provided in https://github.com/Project-OSRM/osrm-backend/blob/master/tools/simpleclient.cpp

felixguendling:
@jornedeblaere

Have a look at the `develop` branch of https://github.com/felixguendling/osrm-backend/.
I merged my multi-target patch into the current develop branch.

PR #1178, opened 2014-09-09
[not ready] Compute route sketches

TheMarex:
This PR implements an experimental idea of displaying a "routing sketch" instead of a textual representation of turn instructions.

This is just an initial tracking PR to get travis/appveyor builds.

This will not work without a modified front-end. The current (hacky) implementation lives in features/route-sketch.


Ylannl:
Interesting. Makes me think of [LineDrive](http://graphics.stanford.edu/papers/routemaps/).

Issue #1077, opened 2014-06-11
Distance matrix limited to 100 stops?


mck-:
Is there any way to increase this limit? All other distances are returned as INT_MAX

drewlesueur:
For people coming to this now, it looks like this is now an argument you can pass in to osrm-routed:
--max-table-size

https://github.com/Project-OSRM/osrm-backend/wiki/osrm-routed.1

PR #1236, opened 2014-10-22
Distance table plugin, set maximum size of input before truncation from cmake #1077

frodrigo:
Distance table plugin, set maximum size of input before truncation from cmake #1077


DennisOSRM:
This gets implemented in #1298, right? Can we close here?

PR #1273, opened 2014-11-17
Feature/protobuf output

DennisOSRM:
Adds a protobuf descriptor and supersedes #1161.

TODO: 
- [x] rebase on develop
- [x] fix protoc call on Windows
- [x] add protobuffer output for 
 - [x] viaroute
 - [x] distance table
 - [x] locate
 - [x] nearest
- [x] remove camel case from response.proto
- [ ] add test to check for correct output
 - [x] viaroute
 - [ ] distance table
 - [ ] locate
 - [ ] nearest
- [ ] make sure JSON and pbf field names are similar/same.
- [ ] micro benchmarks for 
  - [ ] speed 
  - [x] file size


TheMarex:
Rebased on develop and push to my fork:
https://github.com/TheMarex/Project-OSRM/tree/feature/protobuf-output
I expect this will need some rework after #1292 is merged?

DennisOSRM:
@TheMarex yep, One #1292 is merged, this is following suit. 

Want to convert this to a feature branch in the main repo, and replace this PR?

DennisOSRM:
rebased onto develop. Now that the common interface between libOSRM and the rest of the stack is a `json_container`object, the question is whether it makes sense to refactor the ProtoBuffer generation of the output into a renderer that traverses the container.

DennisOSRM:
Windows compile works fine now. Thanks @alex85k for providing protoc binaries.

DennisOSRM:
@emiltin how hard is it to parse protobuffer encoded responses in cucumber tests?

emiltin:
Haven't tried, it but I gues it shouldn't be too hard. There are of course Ruby bindings for protobu https://code.google.com/p/ruby-protobuf. Someone is also maintaining RSpec matchers that can be used with cucumber: https://github.com/connamara/protobuf_spec.

DennisOSRM:
@emiltin thanks sounds easy enough. What do you think about checking the JSON as well the protobuffer encoded response with each query run during tests.

emiltin:
I think it would be better to have a set of tests that specifically exercise the protobuf response format, and have the rest of the tests use only JOSN (or protobuffer if that's faster, which I doubt).

DennisOSRM:
yeah, that probably suffices.

DennisOSRM:
basic implementation is more or less done. Now it's time to get this through the last lap, i.e. add tests and verify the implementation is complete.

emiltin:
what's the benefit of protobuf output - smaller size?

TheMarex:
@emiltin Deserialization should also be faster. Should be interesting for people who use libOSRM and compute a lot of queries/minute.

emiltin:
i'm looking into adding tests for pbf output, but i'm getting empty outputs:

```
~/code/osrm-backend$ (feature/protobuf-output *) curl "http://127.0.0.1:5000/viaroute?loc=1.0,1.0&loc=1.0,1.0008990679362704&alt=false&instructions=true&output=json" -I
HTTP/1.0 200 OK
Content-Length: 558
Content-Type: application/json; charset=UTF-8
Content-Disposition: inline; filename="response.json"

~/code/osrm-backend$ (feature/protobuf-output *) curl "http://127.0.0.1:5000/viaroute?loc=1.0,1.0&loc=1.0,1.0008990679362704&alt=false&instructions=true&output=pbf" -I
HTTP/1.0 200 OK
Content-Length: 0
Content-Type: application/x-protobuf; charset=UTF-8
Content-Disposition: inline; filename="response.json"
```

it also seems odd that the filename includes ".json"?
am i doing something wrong?

DennisOSRM:
@emiltin the code path is probably not correctly wired. I'll get to that Monday.

emiltin:
if you're using osrm as a lib, can you use the basic c++ structures as response format, ie bypass pbf/json?

DennisOSRM:
The lib interface is called _JSON container_. It's an abstract data structure that behaves very similar to what JSON feels like.

emiltin:
ok. just wondering why you would want to use pbf if you use osrm as a lib. wouldn't the encoding/decoding be unnecessary?

DennisOSRM:
yes, and no. You are right that it doesn't make too much sense when you use the lib. The use case is when you don't use the lib, but run requests over http.

TheMarex:
Ah we now expose the json container in the library interface as well? For some reason I thought it was still http::Reply there. Cool.

DennisOSRM:
First benchmarks have come in. The numbers should be taken with a bit of caution but they indicate good file sizes: 
 
     | uncompr. | compr. | compressibility | compr. file size over JSON
----|--------------|-----------|---------------------| ---------------------
json | 1.00     |  0.46     | 0.46                  | 0.46 
pbf| 0.72        | 0.66     | 0.43                  | 0.31

The numbers are relative to the respective JSON size. What we see is that the uncompressed protocol buffer encoded outputs are smaller and compress a little worse. This is expected and the overall compressed file size is still smaller by a good margin.

@emiltin 0384c21 should have proper output. Please note that it is binary encoded data and that it may appear as empty when printing to the console.

emiltin:
do you mean the compressed json is smaller (0.46) than both raw pbf (0.72) and compresed pbf (0.66)?

DennisOSRM:
it's a bit confusing. File sizes are JSON > PBF > compr. JSON > compr. PBF.

emiltin:
is there a reason the field naming is different in pbf responses than in json? it's a bit annoying when testing

DennisOSRM:
Hmm. Thats a but annoying. Should be (more or less) the same as for the other format. Will fix that

emiltin:
added plumming for testing protobuffer output. run 'bundle' to install the required gems.

however, i'm getting an error about an invalid character. not sure if this is real, of a quirk of some ruby lib?

```
~/code/osrm-backend$ (feature/protobuf-output) cuke -t @format -p all
Ruby version 2.1.4
Using default port 5000
Using default timeout 2
Using the all profile...
@routing @pbf @testbot @format
Feature: Outputting protobuffer format

  Background:                   # features/testbot/protobuffer_output.feature:4
    Given the profile "testbot" # features/step_definitions/data.rb:1
    And the query options       # features/step_definitions/options.rb:55
      | output | pbf |

  @todo
  Scenario: Testbot - Protobuffer import, nodes and ways # features/testbot/protobuffer_output.feature:10
    Given the node map                                   # features/step_definitions/data.rb:27
      | a | b |
    And the ways                                         # features/step_definitions/data.rb:66
      | nodes | name |
      | ab    | xyz  |
    When I route I should get                            # features/step_definitions/routing.rb:1
      | from | to | route |
      | a    | b  | ab    |
      invalid character at "\b\x00\x12\x1AFound " (ActiveSupport::OkJson::Error)
      ./features/step_definitions/routing.rb:49:in `block (3 levels) in <top (required)>'
      ./features/step_definitions/routing.rb:5:in `each'
      ./features/step_definitions/routing.rb:5:in `each_with_index'
      ./features/step_definitions/routing.rb:5:in `block (2 levels) in <top (required)>'
      ./features/support/launch.rb:111:in `block in load'
      ./features/support/launch.rb:108:in `chdir'
      ./features/support/launch.rb:108:in `load'
      ./features/support/launch.rb:129:in `/^I route I should get$/'
      features/testbot/protobuffer_output.feature:18:in `When I route I should get'

Failing Scenarios:
cucumber -p all features/testbot/protobuffer_output.feature:10 # Scenario: Testbot - Protobuffer import, nodes and ways

1 scenario (1 failed)
5 steps (1 failed, 4 passed)
0m0.163s
```

the failing scenarios is marked with @todo, so use "-p all" if you want to run it, as above.

at the moment, the protobuffer message formats haven been statically compiled to the ruby file features/support/response.pb.rb. preferably, it would be automatically recompiled whenever the source is modified. i also have a problem with the names of classes with underscores requiring my to manually change the ruby file a bit, so that would have to be solved if we want to do automatic recompiles.

emiltin:
is the "position" in the pbf route_instruction the index, eg 0,1,2,3,4... etc? sometimes i see some indexes missing, ie instructions have positions 0,2,4.

DennisOSRM:
The index references the location in the coordinate array.

DennisOSRM:
Plenty of tests are currently failing with more or less the same error message:

```log
undefined local variable or method `json' for #<Object:0x00000003367a60> (NameError)
./features/step_definitions/routing.rb:91:in `block (3 levels) in <top (required)>'
./features/step_definitions/routing.rb:5:in `each'
./features/step_definitions/routing.rb:5:in `each_with_index'
./features/step_definitions/routing.rb:5:in `block (2 levels) in <top (required)>'
./features/support/launch.rb:111:in `block in load'
./features/support/launch.rb:108:in `chdir'
./features/support/launch.rb:108:in `load'
./features/support/launch.rb:129:in `load'
./features/step_definitions/routing.rb:4:in `/^I route I should get$/'
features/testbot/time.feature:230:in `When I route I should get'
```

emiltin:
yeah. i'm changing the protobuffer gem and will push an update in a second.

DennisOSRM:
cool!

emiltin:
you can now test protobuf output. try "cucumber -t @pbf_out".

as mentioned, the pbf data has to be squeezed into shape to match the json output. nothing wrong with that as such, but would cleaner if the fields of the two outputs matched.

emiltin:
there is still something odd with the index that needs fixing i think. but it's coming along.

DennisOSRM:
What's odd with the index?

emiltin:
fix the index issues, and added a small suite of tests for various pbf outputs.

emiltin:
tests should be added for the other request types - tables, locate, etc.

emiltin:
one of the travis builds stalled due to a network error during gcc install

DennisOSRM:
restarted the failed build

emiltin:
appveyor failed with:

curl: (7) Failed connect to build.project-osrm.org:80; No error
Command exited with code 7

a random network error i suppose?

DennisOSRM:
yeah, nothing serious but it's happening more often recently. Already restarted the build.

DennisOSRM:
@emiltin did you add a test for the other responses than the table call?

emiltin:
cuke tests for protobuffer output currently only cover viaroute calls: https://github.com/Project-OSRM/osrm-backend/blob/feature/protobuf-output/features/testbot/protobuffer_output.feature

do you plan to converge to field names a bit more or is the difference due to pbf requirements?

DennisOSRM:
Thanks for the reminder to make the names consistent. Will do that today.

emiltin:
if you do, you might need to update this cuke helper: https://github.com/Project-OSRM/osrm-backend/blob/feature/protobuf-output/features/support/parse_output.rb

DennisOSRM:
cool, thanks. first a rebase and then renaming the other things

DennisOSRM:
After a rebase I am seeing the following error:

```
      Group is deprecated. (NotImplementedError)
      ./features/support/parse_output.rb:4:in `parse_output'
      ./features/step_definitions/routing.rb:46:in `block (3 levels) in <top (required)>'
      ./features/step_definitions/routing.rb:5:in `each'
      ./features/step_definitions/routing.rb:5:in `each_with_index'
      ./features/step_definitions/routing.rb:5:in `block (2 levels) in <top (required)>'
      ./features/support/launch.rb:111:in `block in load'
      ./features/support/launch.rb:108:in `chdir'
      ./features/support/launch.rb:108:in `load'
      ./features/support/launch.rb:129:in `/^I route I should get$/'
      features/testbot/protobuffer_output.feature:171:in `When I route I should get'
```

@emiltin do you have an idea what's causing this?

emiltin:
a guess: if you changed the protobuffer structures, you need to update/recompile the ruby protobuffer definition features/support/response.pp.rb

DennisOSRM:
ah! i'll give that a try. thanks

emiltin:
note that i had some kind of issue with the automatic naming of the ruby classes due to camel case etc. when compiling. i had to rename them a bit manually. that was with another gem but now you know.

DennisOSRM:
Which tool did you use to generate the ruby code?

emiltin:
if i recall correctly, i used the tool described at https://github.com/localshred/protobuf/wiki/Compiling-Definitions. not working?

DennisOSRM:
not sure why it is not working. keep getting the same error message that `Group is deprecated. (NotImplementedError)`.

emiltin:
did you regenerate the .pp.rb file?

DennisOSRM:
Right, regenerated but there were no significant differences to the existing one.

emiltin:
the protobuffer documenation (google, not the gem) states:

Groups
Note that this feature is deprecated and should not be used when creating new message types – use nested message types instead.

so either the protobuffer definition uses these deprecated, or is somehow intepreted wrong?

emiltin:
response.pb.cc includes references to WIRETYPE_END_GROUP, which is causing exceptions at https://github.com/localshred/protobuf/blob/a5e6efcca4a8ac79bc21566411b3faeed8a07804/lib/protobuf/decoder.rb#L28.

could it be osmium using pbf groups?

TheMarex:
Dropping this PR because it only increases the surface of code to maintain. We should support JSON output only for simplicity and maintainability. If you are worried about data size of the response consider using `node-osrm` and compress the response. The compression of `0.31` of protobuf over compressed JSON `0.46` is not enough to warrant this feature.

PR #1297, opened 2014-11-26
[not ready] Integrate RangeTable into StaticGraph

TheMarex:
StaticGraph uses an adjacency array data structure that can easily be mapped to RangeTable. The idea is so save some memory here.

Preliminary tests have shown no big improvement: For Brandenburg we only saved 2MB (about 5%).

TODO:
- [ ] Fix SharedDatafacade and datastore
- [ ] Fix StaticGraph test
- [ ] Write a benchmark for StaticGraph to test performance
- [x] Actually calculate the amount of memory that could be theoretically saved

Just did a small estimate the potential memory that could be saved is about 400MB for the whole planet, so sadly not that much.


TheMarex:
Obsolete.

PR #1298, opened 2014-11-27
Add a commande line option to osrm-routed for max locations supported in distance table query

frodrigo:
Due to issue (with gcc or luabind depends on computer) to compile develop branch this patch is not tested on this branch. Nevertheless this patch work on last release. #1236

PR #1335, opened 2015-01-08
Feature/frodrigo max locations distance table

DennisOSRM:
supersedes #1298 that got into a quirky state somehow.

### DIFF
--- a/features/options/routed/help.feature
+++ b/features/options/routed/help.feature
@@ -25,7 +25,8 @@ Feature: osrm-routed command line options: help
         And stdout should contain "--port"
         And stdout should contain "--threads"
         And stdout should contain "--sharedmemory"
-        And stdout should contain 22 lines
+        And stdout should contain "--max_locations_distance_table"
+        And stdout should contain 25 lines
         And it should exit with code 0
 
     Scenario: osrm-routed - Help, short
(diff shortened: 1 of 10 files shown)


DennisOSRM:
Thanks again @frodrigo

features/options/routed/help.feature
```diff
@@ -49,7 +50,8 @@ Feature: osrm-routed command line options: help
         And stdout should contain "--port"
         And stdout should contain "--threads"
         And stdout should contain "--sharedmemory"
-        And stdout should contain 22 lines
+        And stdout should contain "--max_locations_distance_table"
```
emiltin: a minor comment: "sharedmemory" doesn't use underscore, while "max_locations_distance_table". it would be nice to be consistent. i also feel max_locations_distance_table is quite long for an option name?

TheMarex: What about "--distance-table-location-limit"?

emiltin: or --max-table-size?
but it still wouldn't fix the inconsistency with "sharedmemory" which doesn't use dash or underscore.

DennisOSRM: thanks, oversight on my behalf. good catch @emiltin. will go with `--max-table-size`.

emiltin: alright. going to change "--sharedmemory" to "--shared-memory" as well or leave for now?

DennisOSRM: yeah, going to add it as well. about to push in 10 minutes.

emiltin: just remember to update wiki and let people know in due time. could break peoples toolchains otherwise :-)

DennisOSRM:
thanks for the reminder. will send a note to the mailing list in a few moments.

DennisOSRM:
Fixed with b1a1276 and email sent.

PR #1338, opened 2015-01-09
Black list access=psv for car profile

frodrigo:
access=psv is like access=forestry but for psv


DennisOSRM:
What does `psv` stand for?

frodrigo:
psv for public service vehicles, less frequent than forestry or agricultural but in use
http://taginfo.openstreetmap.org/keys/access?filter=ways#values

DennisOSRM:
Cool, good catch! Will merge once the CIs come back positive.

DennisOSRM:
Thanks

PR #1339, opened 2015-01-09
remove explicit appveyor os spec to get scheduled on hyper-v

DennisOSRM:


DennisOSRM:
Cc @alex85k seems we can safely remove this flag

PR #1343, opened 2015-01-12
use static graph for component exploration, closes #1288

### LINKED ISSUE #1288: StaticGraph constructor not used
https://github.com/Project-OSRM/osrm-backend/blob/fc1db35f27d03df8dfcf9ef5fa000435acd5e065/DataStructures/StaticGraph.h#L89

This constructor is not used (apart from the ```StaticGraphTest```). Instead the necessary values are build
externally and directly serialized:

https://github.com/Project-OSRM/osrm-backend/blob/75f9c0844550261d5e9c842d0a10b8f82eac2965/Contractor/Prepare.cpp#L241

Should we drop the constructor in favour of a factory function that can create the node array and edge array (so we can use it in ```Prepare.cpp``` and ```StaticGraphTest.cpp```)?

Similar ```InputEdge``` is also not used anymore (but various typdefs reference).

BTW. Do you know any utility to detect similar dead code?

/cc @DennisOSRM

DennisOSRM:
should be faster and use less RAM than the dynamic graph.

### DIFF
--- a/tools/components.cpp
+++ b/tools/components.cpp
@@ -27,7 +27,7 @@ SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
 #include "../typedefs.h"
 #include "../algorithms/tiny_components.hpp"
-#include "../data_structures/dynamic_graph.hpp"
+#include "../data_structures/static_graph.hpp"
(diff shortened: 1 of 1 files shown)


DennisOSRM:
ready to be merged once CIs come back green.

PR #1347, opened 2015-01-15
Fix ComputePerpendicularDistance convinience function

TheMarex:
- [x] Wait for travis
- [x] Fix strange formating errors
- [x] Add unit test for FixedPointCoordinate


emiltin:
what errors did this cause?

TheMarex:
@emiltin This function is used in StaticRTree to compute the key for the priority queue in the ```IncrementalFindPhantomNode``` function. In some cases it gave values that are much higher than the real distance.
I don't think it did cause noticeable bugs in the current develop branch, but I use a variation of this function in another branch that needs precise distance values.

DennisOSRM:
:+1:

PR #1350, opened 2015-01-20
speed up nearest neighbor query by pruning

DennisOSRM:
- an upper bound of the k largest element is derived when going thru the _stream_ of edges in nearest neighbor lookup. The lookup of nearest neighbors is sped up considerably:

scenario |  before | after | ratio
---------|----------|---------------------|-------
incremental 1 | 1.98 ms | 1.22 ms | 0.61
incremental 5 | 2.11 ms | 1.26 ms | 0.64

- Rule of thumb: every query is now a millisecond faster. Mileage may vary depending on CPU and SSD/HDD.

- moves a lot of logic when handling coordinates away from the library interface


emiltin:
:+1:

TheMarex:
Cool. :+1:

PR #1352, opened 2015-01-21
test VS2015 build

DennisOSRM:
background: http://www.appveyor.com/blog/2015/01/20/visual-studio-2015-ctp-image

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -37,7 +37,7 @@ build_script:
   - SET P=c:/projects/osrm
   - set TBB_INSTALL_DIR=%P%/tbb
   - set TBB_ARCH_PLATFORM=intel64/vc12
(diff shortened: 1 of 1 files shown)


emiltin:
Travis failed due to a ruby timeout when launching osrm-routed. The timeout is currently set yo 60 seconds. We can increase it if we want.

DennisOSRM:
thanks @emiltin. I'll rerun for the sake of having an all green. The Travis environment is rather fishy.

DennisOSRM:
@alex85k would love to have your opinion on this

alex85k:
Swithcing to VS2015 is may be harder. cmake redefinesthe compiler, ``-G "Visual Studio 14 2015 Win64" ``` is needed. Currently it still uses MSVC 18.0 from VS2013...

Most likely, the dependencies pack will be incompatible (however, the libs from https://github.com/BergWerkGIS/mapnik-dependencies are tested on VS2015).

Switching to VS2015 preview will certainly be needed after VS2015 Express release but I am not sure we should drop VS2013+CTP (most popular) version now... Better have both but that means two dependencies packs.

DennisOSRM:
Alright, then let's revisit once the software is actually released.

Issue #1369, opened 2015-01-31
accessing way's id in lua


bjtaylor1:
I used to (probably at least several months ago) be able to access a way's id in lua by just  "way.id" but this doesn't seem to work now, it is always "nil".

Any way I can do it?

I am using a fairly recent update of master branch.

emiltin:
it should be possible. if not, sounds like a bug

bjtaylor1:
I think it is a bug.
I just tried it with the standard car profile and put 

```lua
function way_function (way, result)
  if way.id ~= nil then
    print("got way id!"..way.id)
  else
    print ("no way id")
  end
...
end
```
in the top of the way function, and it always prints "no way id".

emiltin:
i would start by lookin in https://github.com/Project-OSRM/osrm-backend/blob/725b93a961625a7b04d54806d7e0f80252a6bcd0/extractor/scripting_environment.cpp where the luabind link to c++ is defined

bjtaylor1:
ok cheers, that works.
I can access it as a function and it returns the id.
Should I submit a merge request?

emiltin:
maybe the access by function is intended. @DennisOSRM? if so, then the example should be updated.

TheMarex:
This seems to be a regression from the switch to libosmium. @bjtaylor1 you added a binding for osmium::Way::id() ? Pull requests are always welcome! :+1:

bjtaylor1:
Yes. I just created a pull request for it.

PR #1375, opened 2015-02-04
added lua binding for osmium::Way::id()

bjtaylor1:
as discussed here: https://github.com/Project-OSRM/osrm-backend/issues/1369

### DIFF
--- a/extractor/scripting_environment.cpp
+++ b/extractor/scripting_environment.cpp
@@ -118,6 +118,7 @@ void ScriptingEnvironment::init_lua_state(lua_State* lua_state)
         luabind::class_<osmium::Way>("Way")
         .def("get_value_by_key", &osmium::Way::get_value_by_key)
         .def("get_value_by_key", &get_value_by_key<osmium::Way>)
+				.def("id", &osmium::Way::id)
     ];
 
     if (0 != luaL_dofile(lua_state, file_name.c_str()))


DennisOSRM:
Thanks for the PR. I'm going to merge this manually into develop branch.

emiltin:
is this ready to merge?

DennisOSRM:
Already cherry-picked into develop branch.

DennisOSRM:
It's here: https://github.com/Project-OSRM/osrm-backend/commit/d77a88c21e6b847c02b2d52af0ee7519c634a851

emiltin:
did the syntax change? when i try:

```lua
id_plus_one = 1 + way.id
```

i get:

```
attempt to concatenate field 'id' (a function value)
```

emiltin:
way.id() doesn't work either i get:

terminate called after throwing an instance of 'osrm::exception'
  what():  ERROR occured in profile script:
No matching overload found, candidates:
long id(Way const&)
Aborted (core dumped)
*** Exited with code 134.

DennisOSRM:
reopening to track whats missing in this issue

DennisOSRM:
shouldn't it be `way:id()` rather than `way.id()`

TheMarex:
So, this does actually work with ```way:id()``` right? Can we close?

DennisOSRM:
There was an issue binding members from a superclass into Lua. It does not yet work.

bjtaylor1:
The binding I added has to be used like `way:id()` in lua. It worked ok when I did it....

TheMarex:
Uhm, adding this to profile.lua
```lua
  local id_test = way:id() + 1
  io.write("id: ", id_test, "\n")
```
Has exactly the indented effect. So it does work, doesn't it? Can we close @DennisOSRM?

DennisOSRM:
Hmm.. perhaps my notebook is not up to date. I'll revisit tomorrow morning and update/close the issue as needed.

kemets:
Is it possible that this is no longer working? Using `way:id()` I get a `std::runtime_error: 'Trying to use unregistered class'` . The following minimal `profile.lua` demonstrates the problem:

    function node_function(node)
      return 1
    end

    function way_function (way)
      local id_test = way:id() + 1
      io.write("id: ", id_test, "\n")
      return 1
    end

I am using OSRM on Mac OS X with the dependencies installed using homebrew.

emiltin:
not sure if it's the reason for your error, but the node and way functions take two arguments:
https://github.com/Project-OSRM/osrm-backend/blob/master/profiles/car.lua#L213

kemets:
You're correct, but it doesn't seem to matter. I started from https://github.com/Project-OSRM/osrm-backend/blob/master/profiles/car.lua , but the example I gave in this thread was based on https://github.com/Project-OSRM/osrm-backend/blob/master/profiles/examples/postgis.lua#L48 which only takes one argument for the node and way functions. I adapted my example to take two arguments, but the error remains.

systemed:
Same `'Trying to use unregistered class'` error for me (also OS X, Homebrew).

kemets:
Sorry for the late follow up.  The exception is caused by the return type of the `id()` function defined for `osmium::OSMObject` (i.e., the parent class for all derived classes such as `osmium::Way` and `osmium::Node`). 

Looking at the definition of `osmium::OSMObject::id()`:

    /// Get ID of this object.
    object_id_type id() const noexcept {
        return m_id;
    }

The `object_id_type` is a typedef for `int64_t`:

    typedef int64_t  object_id_type;

The definition of `int64_t` depends on the platform. On Ubuntu Trusty (where we the exception is not thrown) it is defined in `/usr/include/stdint.h`:

    # if __WORDSIZE == 64
    typedef long int                int64_t;
    # else
    __extension__
    typedef long long int           int64_t;

However, the definition on OS X is in `/usr/include/sys/_types/_int64_t.h`:

    typedef long long               int64_t;

which seems to be the cause of the exception. 

We're deploying on Linux so switched our development environment to a virtual machine to circumvent this problem. Not sure where this should be solved, I expect at the luabind side?

TheMarex:
@kemets so you are running on a 32 bit system? Your systems standard library headers should define `int64_t` correctly as 64bit integer. Is it a bug in luabind on OSX?

kemets:
Both my OS X 10.9.5 and Ubuntu Trusty are 64 bit. Did a quick test also:

    sizeof(long int): 8
    sizeof(long long): 8
    sizeof(long long int): 8
    sizeof(int64_t): 8

Might it have something to do with the `LUABIND_NUMBER_CONVERTER` definitions in https://github.com/luabind/luabind/blob/master/luabind/detail/policy.hpp#L690 ?

TheMarex:
@kemets oh yeah type deduction will most likely not figure out that `long long int` == `long int`. Did you try it with adding `int64_t` to the list?

daniel-j-h:
As reported today on IRC, `way::id()` still crashes OSRM. User reported this for Ubuntu.

TheMarex:
@daniel-j-h you mean `way:id()` ? Is this a luabind version thing?

daniel-j-h:
Yes, sorry for the typo.

On OSX @danpat reported the following exception being thrown:

    /usr/local/include/luabind/detail/make_instance.hpp
    80:        throw std::runtime_error("Trying to use unregistered class");

On IRC (Ubuntu, Release build) a user reported:

    [info]   vehicle
    terminate called after throwing an instance of 'osrm::util::exception'
    terminate called recursively
    terminate called recursively
    Aborted (core dumped)

which looks like a throwing destructor to me (bad). Could actually be the same issue.

TheMarex:
@daniel-j-h Can't reproduce this on Arch Linux with the following profile modification:

```
diff --git a/profiles/car.lua b/profiles/car.lua
index f6bfd23..2adcabd 100644
--- a/profiles/car.lua
+++ b/profiles/car.lua
@@ -212,6 +212,7 @@ function node_function (node, result)
 end
 
 function way_function (way, result)
+  io.write("id:" .. way:id())
   local highway = way:get_value_by_key("highway")
   local route = way:get_value_by_key("route")
   local bridge = way:get_value_by_key("bridge")
```

Can the user try this and report the error?

Issue #1379, opened 2015-02-11
Long requests fail to parse


neilbu:
If a viaroute/table call is made with a large number of query string parameters (i.e. 1500+ characters), parsing will periodically fail because the request parser is constructed on demand.  Any state from parsing the previous part of the request is lost...apologies if this is a deliberate thing, but thought I'd mention it since it (occasionally) mars what is otherwise a work of pure genius!

DennisOSRM:
Got an example call?

neilbu:
Sorry - yes:

[...]

PR #1380, opened 2015-02-12
Made request_parser a member of Connection

neilbu:
As discussed in issue #1379 - a simple (and quite probably naive) change to allow long strings to be parsed.

Previously, the request parser was created on demand...but this could
cause requests to be flagged as bad requests if the string length was
large enough to cause the string to be split up into chunks by boost's
read_some method.


neilbu:
In case it's not obvious, commit e48b927 shouldn't be part of this pull request...I don't know why it's showing up!  Sorry...

TheMarex:
@neilbu you can create a new branch and then do:

> git reset HEAD^
> git checkout -f

To remove the latest commit.

neilbu:
Thank you!

PR #1381, opened 2015-02-12
Made the request parser a member of the Connection

neilbu:
As discussed in issue #1379 - a simple (and quite probably naive) change to allow long strings to be parsed.

Previously, the request parser was created on demand...but this could
cause requests to be flagged as bad requests if the string length was
large enough to cause the string to be split up into chunks by boost's
read_some method.

(Apologies for creating a second pull request!)


neilbu:
Debug code removed :)

DennisOSRM:
Cool, I'll get this reviewed today.

DennisOSRM:
thanks, much appreciated.

PR #1396, opened 2015-02-25
Small change in order to be able to use OSRM as submodule in other projects

RockLobster:
Replaced CMAKE_SOURCE_DIR in osrm root dir's cmake file with CMAKE_CURRENT_SOURCE_DIR

=> Allows OSRM to be used as a cmake subproject as well


DennisOSRM:
Looking good. Will merge within the hour.

DennisOSRM:
thanks

PR #1397, opened 2015-02-25
JSON string escape fix

TheMarex:
Moves the escaping to json_renderer


DennisOSRM:
There's a function called `EscapeJSON' in util/string_util.hpp that has much better complexity. Calling insert() on a string has bad complexity as it reallocates and copies all suffixes after the insert.

I'll run point on this and also make sure the escaped street names don't get escaped twice. Some tests are failing.

DennisOSRM:
Thanks, @TheMarex. Great detective work.

PR #1401, opened 2015-02-26
lint corrections

DennisOSRM:
lint corrections
    
    - fix license header
    - let shared_ptr autodestruct as it's shared
    - rename fences to resemble file names
    - reorder includes into lexicographic order

@TheMarex


TheMarex:
:+1:

PR #1407, opened 2015-03-03
Map Matching

TheMarex:
Implements map matching based on ```"Hidden Markov Map Matching Through Noise and Sparseness"; P. Newson and J. Krumm; 2009; ACM```.
This can be used to match GPS traces to a road network. Note: Currently this works best with traces that have a sample frequency of about 10 samples/minute. You should make sure to remove blobs (accumulated GPS points, when standing still) and reduce the sample to 10 samples/minute.
The ```gps_precision``` and ```matching_beta``` parameters can be tuned to adapt the matching to your dataset.

API additions:

```
t = UNIX_TIMESTAMP
gps_precision = GPS_STANDART_DEVIATION (in meters, default: -1 which means auto)
matching_beta = DOUBLE  (see paper for details, default: -1 which means auto)
classify = BOOL (default: false enables computing confidence for the trace)
```

Usage:

```
/match?loc=lat1,lon1&t=0&loc=lat2,lon2&t=7&loc=lat3,lon3&t=14&geometry=false&classify=true
```

routed parameter additions:

```
--max-locations-map-matching (default: -1 which means disabled)
```


emiltin:
great!

algorithms/bayes_classifier.hpp
```diff
@@ -0,0 +1,114 @@
+/*
+
+Copyright (c) 2015, Project OSRM, Dennis Luxen, others
```
DennisOSRM: Copyright (c) 2015, Project OSRM contributors

algorithms/bayes_classifier.hpp
```diff
@@ -0,0 +1,114 @@
...
+
+#include <vector>
+#include <cmath>
+
+struct NormalDistribution
+{
+    NormalDistribution(const double mean, const double standard_deviation)
+    : mean(mean)
+    , standard_deviation(standard_deviation)
+    {
```
DennisOSRM: please reformat with clang-format 3.5+ and the .clang-format file from the repository root.

algorithms/bayes_classifier.hpp
```diff
@@ -0,0 +1,114 @@
...
+    }
+
+    // FIXME implement log-probability version since its faster
+    double probabilityDensityFunction(const double val)
```
DennisOSRM: could this function be const?

algorithms/bayes_classifier.hpp
```diff
@@ -0,0 +1,114 @@
...
+    : location(location)
+    , scale(scale)
+    {
+    }
+
+    // FIXME implement log-probability version since its faster
+    double probabilityDensityFunction(const double val) const
+    {
+        const double x = std::abs(val - location);
+        return 1.0 / (2*scale) * std::exp(-x / scale);
```
DennisOSRM: use `2.` to hint that this is a floating point multiplication

algorithms/bayes_classifier.hpp
```diff
@@ -0,0 +1,114 @@
...
+ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+(INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+*/
+
+#ifndef BAYES_CLASSIFIER_HPP
+#define BAYES_CLASSIFIER_HPP
+
+#include <vector>
```
DennisOSRM: reorder include lexicographically in the order of local includes -> c headers -> STL headers

data_structures/static_rtree.hpp
```diff
@@ -796,20 +793,32 @@ class StaticRTree
         return !result_phantom_node_vector.empty();
     }
 
-    // implementation of the Hjaltason/Samet query [3], a BFS traversal of the tree
```
DennisOSRM: Refrain from C-style comments (`/* .. */`), especially over multiple lines, use C++-style comments instead (`//`)

library/osrm_impl.cpp
```diff
@@ -42,6 +42,7 @@ class named_mutex;
 #include "../plugins/nearest.hpp"
 #include "../plugins/timestamp.hpp"
 #include "../plugins/viaroute.hpp"
+#include "../plugins/map_matching.hpp"
```
DennisOSRM: reorder includes to be alphabetically ordered

plugins/map_matching.hpp
```diff
@@ -0,0 +1,314 @@
...
+    {
+        double distance_feature = -std::log(trace_length) + std::log(matched_length);
+
+        // matched to the same point
+        if (!std::isfinite(distance_feature))
+        {
+            return std::make_pair(ClassifierT::ClassLabel::NEGATIVE, 1.0);
+        }
+
+        auto label_with_confidence =  classifier.classify(distance_feature);
```
DennisOSRM: can this line be moved to the return statement, where `label_with_confidence` is used?

plugins/map_matching.hpp
```diff
@@ -0,0 +1,314 @@
...
+                unsigned compact_size = candidates.size();
+                for (const auto i : osrm::irange(0u, compact_size))
+                {
+                    // Split edge if it is bidirectional and append reverse direction to end of list
+                    if (candidates[i].first.forward_node_id != SPECIAL_NODEID
+                     && candidates[i].first.reverse_node_id != SPECIAL_NODEID)
+                    {
+                        PhantomNode reverse_node(candidates[i].first);
+                        reverse_node.forward_node_id = SPECIAL_NODEID;
+                        candidates.push_back(std::make_pair(reverse_node, candidates[i].second));
```
DennisOSRM: using `emplace_back()` here should allow to get rid of the surrounding/inner `make_pair()` call.

plugins/map_matching.hpp
```diff
@@ -0,0 +1,314 @@
...
+    }
+
+    bool getCandiates(const std::vector<FixedPointCoordinate>& input_coords, std::vector<double>& sub_trace_lengths, Matching::CandidateLists& candidates_lists)
+    {
+        double last_distance = coordinate_calculation::great_circle_distance(
+            input_coords[0],
+            input_coords[1]);
+        sub_trace_lengths.resize(input_coords.size());
+        sub_trace_lengths[0] = 0;
+        for (const auto current_coordinate : osrm::irange<std::size_t>(0, input_coords.size()))
```
DennisOSRM: `osrm::irange(0u, input_coords.size())` should work to make the code more readable.

util/json_util.hpp
```diff
@@ -0,0 +1,101 @@
...
+    }
+    if (-std::numeric_limits<T>::infinity() == d)
+    {
+        return -std::numeric_limits<T>::max();
+    }
+
+    return d;
+}
+
+void append_to_array(osrm::json::Array &a) {}
```
DennisOSRM: can we reuse the functionality from `util/container.hpp`'s `append_to_container()`?

plugins/map_matching.hpp
```diff
@@ -0,0 +1,314 @@
...
+            subtrace.values["confidence"] = sub.confidence;
+
+        if (route_parameters.geometry)
+        {
+            DescriptionFactory factory;
+            FixedPointCoordinate current_coordinate;
+            factory.SetStartSegment(
+                raw_route.segment_end_coordinates.front().source_phantom,
+                raw_route.source_traversed_in_reverse.front());
+            for (const auto i : osrm::irange<std::size_t>(0, raw_route.unpacked_path_segments.size()))
```
DennisOSRM: `osrm::irange(0u, raw_route.unpacked_path_segments.size())` should work

plugins/map_matching.hpp
```diff
@@ -0,0 +1,314 @@
...
+            return 400;
+        }
+
+        // enforce maximum number of locations for performance reasons
+        if (max_locations_map_matching > 0 && static_cast<int>(input_coords.size()) > max_locations_map_matching)
+        {
+            return 400;
+        }
+
+        bool found_candidates = getCandiates(input_coords, sub_trace_lengths, candidates_lists);
```
DennisOSRM: can we make the bool const?

plugins/map_matching.hpp
```diff
@@ -0,0 +1,314 @@
...
+                                        route_parameters.gps_precision,
+                                        sub_matchings);
+
+        if (1 > sub_matchings.size())
+        {
+            return 400;
+        }
+
+        osrm::json::Array matchings;
+        for (auto& sub : sub_matchings)
```
DennisOSRM: would it make sense to use ``std::transform()`` here to make the intent more clear that the sub_matching container is to be changed instead of iterated?

plugins/map_matching.hpp
```diff
@@ -0,0 +1,314 @@
...
+        {
+            // classify result
+            if (route_parameters.classify)
+            {
+                double trace_length = sub_trace_lengths[sub.indices.back()] - sub_trace_lengths[sub.indices.front()];
+                TraceClassification classification = classify(trace_length,
+                                                              sub.length,
+                                                              (sub.indices.back() - sub.indices.front() + 1) - sub.nodes.size());
+                if (classification.first == ClassifierT::ClassLabel::POSITIVE)
```
DennisOSRM: Would it make sense to switch the two compared items? That way there wouldn't be an lvalue on the left-hand side and accidental assignments would (or could) not happen.

plugins/map_matching.hpp
```diff
@@ -0,0 +1,314 @@
...
+                }
+            }
+
+            BOOST_ASSERT(sub.nodes.size() > 1);
+
+            // FIXME we only run this to obtain the geometry
+            // The clean way would be to get this directly from the map matching plugin
+            InternalRouteResult raw_route;
+            PhantomNodes current_phantom_node_pair;
+            for (unsigned i = 0; i < sub.nodes.size() - 1; ++i)
```
DennisOSRM: Does it make sense to use an range-based for with an irange object? This way, variable `i` could be const.

plugins/map_matching.hpp
```diff
@@ -0,0 +1,314 @@
...
+            }
+            search_engine_ptr->shortest_path(
+                raw_route.segment_end_coordinates,
+                std::vector<bool>(raw_route.segment_end_coordinates.size(), true),
+                raw_route);
+
+            matchings.values.emplace_back(submatchingToJSON(sub, route_parameters, raw_route));
+        }
+
+        if (osrm::json::Logger::get())
```
DennisOSRM: add curly braces after the if statement header

plugins/map_matching.hpp
```diff
@@ -0,0 +1,314 @@
...
+        }
+
+        bool found_candidates = getCandiates(input_coords, sub_trace_lengths, candidates_lists);
+        if (!found_candidates)
+        {
+            return 400;
+        }
+
+        // setup logging if enabled
+        if (osrm::json::Logger::get())
```
DennisOSRM: add curly braces after the if statement header

routing_algorithms/map_matching.hpp
```diff
@@ -0,0 +1,627 @@
...
+
+*/
+
+#ifndef MAP_MATCHING_HPP
+#define MAP_MATCHING_HPP
+
+#include "routing_base.hpp"
+
+#include "../data_structures/coordinate_calculation.hpp"
+#include "../util/simple_logger.hpp"
```
DennisOSRM: reorder includes

routing_algorithms/map_matching.hpp
```diff
@@ -0,0 +1,627 @@
...
+#include <osrm/json_container.hpp>
+#include <variant/variant.hpp>
+
+#include <fstream>
+
+#include <algorithm>
+#include <iomanip>
+#include <numeric>
+
+namespace Matching
```
DennisOSRM: namespaces are lowercase and should be nested inside an outer `osrm` namespace.

routing_algorithms/map_matching.hpp
```diff
@@ -0,0 +1,627 @@
...
+    std::vector<PhantomNode> nodes;
+    std::vector<unsigned> indices;
+    double length;
+    double confidence;
+};
+
+using CandidateList = std::vector<std::pair<PhantomNode, double>>;
+using CandidateLists = std::vector<CandidateList>;
+using SubMatchingList = std::vector<SubMatching>;
+constexpr static const unsigned max_number_of_candidates = 10;
```
DennisOSRM: can we move these const's to the top of the namespace?

routing_algorithms/map_matching.hpp
```diff
@@ -0,0 +1,627 @@
...
+// implements a hidden markov model map matching algorithm
+template <class DataFacadeT> class MapMatching final
+    : public BasicRoutingInterface<DataFacadeT, MapMatching<DataFacadeT>>
+{
+    using super = BasicRoutingInterface<DataFacadeT, MapMatching<DataFacadeT>>;
+    using QueryHeap = SearchEngineData::QueryHeap;
+    SearchEngineData &engine_working_data;
+
+    // FIXME this value should be a table based on samples/meter (or samples/min)
+    constexpr static const double default_beta = 10.0;
```
DennisOSRM: can we move these const's to the top of the class?

routing_algorithms/map_matching.hpp
```diff
@@ -0,0 +1,627 @@
...
+    constexpr static const double default_sigma_z = 4.07;
+    constexpr static const double log_2_pi = std::log(2 * M_PI);
```
DennisOSRM: Does the standard define `std::log` as constexpr?

routing_algorithms/map_matching.hpp
```diff
@@ -0,0 +1,627 @@
...
+    struct HiddenMarkovModel
+    {
+        std::vector<std::vector<double>> viterbi;
+        std::vector<std::vector<std::pair<unsigned, unsigned>>> parents;
+        std::vector<std::vector<float>> path_lengths;
+        std::vector<std::vector<bool>> pruned;
+        std::vector<bool> breakage;
+
+        const Matching::CandidateLists &candidates_list;
+        const EmissionLogProbability& emission_log_probability;
```
DennisOSRM: The position of the ampersand is inconsistent

routing_algorithms/map_matching.hpp
```diff
@@ -0,0 +1,627 @@
...
+            clear(0);
+        }
+
+        void clear(unsigned initial_timestamp)
+        {
+            BOOST_ASSERT(viterbi.size() == parents.size() &&
+                         parents.size() == path_lengths.size() &&
+                         path_lengths.size() == pruned.size() && pruned.size() == breakage.size());
+
+            for (unsigned t = initial_timestamp; t < viterbi.size(); t++)
```
DennisOSRM: a range-based for loop with an `irange` would allow to make t const, also `t++` should be `++t`.

routing_algorithms/map_matching.hpp
```diff
@@ -0,0 +1,627 @@
...
+                viterbi.emplace_back(l.size());
+                parents.emplace_back(l.size());
+                path_lengths.emplace_back(l.size());
+                pruned.emplace_back(l.size());
+            }
+
+            clear(0);
+        }
+
+        void clear(unsigned initial_timestamp)
```
DennisOSRM: - making the parameter a `std::size_t` removes a number of implicit up-casts below.
- would it make sense to use a default parameter of zero?

routing_algorithms/map_matching.hpp
```diff
@@ -0,0 +1,627 @@
...
+            std::fill(breakage.begin() + initial_timestamp, breakage.end(), true);
+        }
+
+        unsigned initialize(unsigned initial_timestamp)
+        {
+            BOOST_ASSERT(initial_timestamp < candidates_list.size());
+
+            do
+            {
+                for (auto s = 0u; s < viterbi[initial_timestamp].size(); ++s)
```
DennisOSRM: a range-based for loop with an irange would allow to make `s` const

routing_algorithms/map_matching.hpp
```diff
@@ -0,0 +1,627 @@
...
+            if (logger)
+            {
+                object = &logger->map->at("matching");
+            }
+        }
+
+        void initialize(const Matching::CandidateLists& candidates_list)
+        {
+            // json logger not enabled
+            if (!logger)
```
DennisOSRM: use curly braces.

routing_algorithms/map_matching.hpp
```diff
@@ -0,0 +1,627 @@
...
+                return;
+
+            osrm::json::Array states;
+            for (unsigned t = 0; t < candidates_list.size(); t++)
```
DennisOSRM: a range-based for loop with an irange would allow to make t const

routing_algorithms/map_matching.hpp
```diff
@@ -0,0 +1,627 @@
...
+            {
+                osrm::json::Array timestamps;
+                for (unsigned s = 0; s < candidates_list[t].size(); s++)
```
DennisOSRM: - a range-based for loop with an irange would allow to make `s` const
- `s++` should be `++s`

routing_algorithms/map_matching.hpp
```diff
@@ -0,0 +1,627 @@
...
+                                 const unsigned prev_state,
+                                 const unsigned current_state,
+                                 const double prev_viterbi,
+                                 const double emission_pr,
+                                 const double transition_pr,
+                                 const double network_distance,
+                                 const double great_circle_distance)
+        {
+            // json logger not enabled
+            if (!logger)
```
DennisOSRM: use curly braces.

routing_algorithms/map_matching.hpp
```diff
@@ -0,0 +1,627 @@
...
+                .get<mapbox::util::recursive_wrapper<osrm::json::Array>>()
+                .get().values.push_back(transistion);
+
+        }
+
+        void set_viterbi(const std::vector<std::vector<double>>& viterbi,
+                         const std::vector<std::vector<bool>>& pruned)
+        {
+            // json logger not enabled
+            if (!logger)
```
DennisOSRM: use curly braces

routing_algorithms/map_matching.hpp
```diff
@@ -0,0 +1,627 @@
...
+                return;
+
+            for (auto t = 0u; t < viterbi.size(); t++)
```
DennisOSRM: - a range-based for loop with an irange would allow to make `t` const
- `t++` should be `++t`

routing_algorithms/map_matching.hpp
```diff
@@ -0,0 +1,627 @@
...
+            {
+                for (auto s_prime = 0u; s_prime < viterbi[t].size(); ++s_prime)
```
DennisOSRM: - a range-based for loop with an irange would allow to make `s_prime` const
- `s_prime++` should be `++s_prime`

routing_algorithms/map_matching.hpp
```diff
@@ -0,0 +1,627 @@
...
+                    osrm::json::get(*object, "states", t, s_prime, "viterbi") = osrm::json::clamp_float(viterbi[t][s_prime]);
+                    osrm::json::get(*object, "states", t, s_prime, "pruned") = static_cast<unsigned>(pruned[t][s_prime]);
+                }
+            }
+        }
+
+        void add_chosen(const unsigned t, const unsigned s)
+        {
+            // json logger not enabled
+            if (!logger)
```
DennisOSRM: use curly braces

routing_algorithms/map_matching.hpp
```diff
@@ -0,0 +1,627 @@
...
+                return;
+
+            osrm::json::get(*object, "states", t, s, "chosen") = true;
+        }
+
+        void add_breakage(const std::vector<bool>& breakage)
+        {
+            // json logger not enabled
+            if (!logger)
```
DennisOSRM: use curly braces

routing_algorithms/map_matching.hpp
```diff
@@ -0,0 +1,627 @@
...
+
+        DebugInfo debug(osrm::json::Logger::get());
+        debug.initialize(candidates_list);
+
+        unsigned breakage_begin = std::numeric_limits<unsigned>::max();
+        std::vector<unsigned> split_points;
+        std::vector<unsigned> prev_unbroken_timestamps;
+        prev_unbroken_timestamps.reserve(candidates_list.size());
+        prev_unbroken_timestamps.push_back(initial_timestamp);
+        for (auto t = initial_timestamp + 1; t < candidates_list.size(); ++t)
```
DennisOSRM: - a range-based for loop with an irange would allow to make `t` const

routing_algorithms/map_matching.hpp
```diff
@@ -0,0 +1,627 @@
...
+
+            auto &current_viterbi = model.viterbi[t];
+            auto &current_pruned = model.pruned[t];
+            auto &current_parents = model.parents[t];
+            auto &current_lengths = model.path_lengths[t];
+            const auto &current_timestamps_list = candidates_list[t];
+            const auto &current_coordinate = trace_coordinates[t];
+
+            // compute d_t for this timestamp and the next one
+            for (auto s = 0u; s < prev_viterbi.size(); ++s)
```
DennisOSRM: - a range-based for loop with an irange would allow to make `s` const

routing_algorithms/map_matching.hpp
```diff
@@ -0,0 +1,627 @@
...
+                prev_unbroken_timestamps.push_back(new_start);
+                // Important: We potentially go back here!
+                // However since t > new_start >= breakge_begin
+                // we can only reset trace_coordindates.size() times.
+                t = new_start+1;
+            }
+
+            unsigned prev_unbroken_timestamp = prev_unbroken_timestamps.back();
+
+            const auto &prev_viterbi = model.viterbi[prev_unbroken_timestamp];
```
DennisOSRM: does it make sense to capture this state information in a state object?

routing_algorithms/map_matching.hpp
```diff
@@ -0,0 +1,627 @@
...
+            auto &current_pruned = model.pruned[t];
+            auto &current_parents = model.parents[t];
+            auto &current_lengths = model.path_lengths[t];
+            const auto &current_timestamps_list = candidates_list[t];
+            const auto &current_coordinate = trace_coordinates[t];
+
+            // compute d_t for this timestamp and the next one
+            for (auto s = 0u; s < prev_viterbi.size(); ++s)
+            {
+                if (prev_pruned[s])
```
DennisOSRM: use curly braces

routing_algorithms/map_matching.hpp
```diff
@@ -0,0 +1,627 @@
...
+                    continue;
+
+                for (auto s_prime = 0u; s_prime < current_viterbi.size(); ++s_prime)
```
DennisOSRM: - a range-based for loop with an irange would allow to make `s_prime` const

routing_algorithms/map_matching.hpp
```diff
@@ -0,0 +1,627 @@
...
+                {
+                    // how likely is candidate s_prime at time t to be emitted?
+                    const double emission_pr =
+                        emission_log_probability(candidates_list[t][s_prime].second);
+                    double new_value = prev_viterbi[s] + emission_pr;
+                    if (current_viterbi[s_prime] > new_value)
```
DennisOSRM: use curly braces

routing_algorithms/map_matching.hpp
```diff
@@ -0,0 +1,627 @@
...
+                        get_network_distance(prev_unbroken_timestamps_list[s].first,
+                                             current_timestamps_list[s_prime].first);
+                    const auto great_circle_distance =
+                        coordinate_calculation::great_circle_distance(prev_coordinate,
+                                                                      current_coordinate);
+
+                    const auto d_t = std::abs(network_distance - great_circle_distance);
+
+                    // very low probability transition -> prune
+                    if (d_t > 500)
```
DennisOSRM: use curly braces

routing_algorithms/map_matching.hpp
```diff
@@ -0,0 +1,627 @@
...
+        {
+            Matching::SubMatching matching;
+
+            // find real end of trace
+            // not sure if this is really needed
+            unsigned parent_timestamp_index = sub_matching_end - 1;
+            while (parent_timestamp_index >= sub_matching_begin &&
+                   model.breakage[parent_timestamp_index])
+            {
+                parent_timestamp_index--;
```
DennisOSRM: use prefix decrement

routing_algorithms/map_matching.hpp
```diff
@@ -0,0 +1,627 @@
...
+            if (reconstructed_indices.size() < 2)
+            {
+                sub_matching_begin = sub_matching_end;
+                continue;
+            }
+
+            matching.length = 0.0f;
+            matching.nodes.resize(reconstructed_indices.size());
+            matching.indices.resize(reconstructed_indices.size());
+            for (auto i = 0u; i < reconstructed_indices.size(); ++i)
```
DennisOSRM: a range-based for loop with an irange would allow to make `i` const

routing_algorithms/map_matching.hpp
```diff
@@ -0,0 +1,627 @@
...
+
+                matching.indices[i] = timestamp_index;
+                matching.nodes[i] = candidates_list[timestamp_index][location_index].first;
+                matching.length += model.path_lengths[timestamp_index][location_index];
+
+                debug.add_chosen(timestamp_index, location_index);
+            }
+
+            sub_matchings.push_back(matching);
+
```
DennisOSRM: empty line necessary for formatting/readability?

server/api_grammar.hpp
```diff
@@ -62,6 +63,14 @@ template <typename Iterator, class HandlerT> struct APIGrammar : qi::grammar<Ite
                     qi::double_)[boost::bind(&HandlerT::addCoordinate, handler, ::_1)];
         hint = (-qi::lit('&')) >> qi::lit("hint") >> '=' >>
                stringwithDot[boost::bind(&HandlerT::addHint, handler, ::_1)];
+        timestamp = (-qi::lit('&')) >> qi::lit("t") >> '=' >>
```
DennisOSRM: which of these fields are optional or necessary in combination with others. Right now, the grammar accepts any (wild) combination.

server/api_grammar.hpp
```diff
@@ -83,7 +92,7 @@ template <typename Iterator, class HandlerT> struct APIGrammar : qi::grammar<Ite
 
     qi::rule<Iterator> api_call, query;
     qi::rule<Iterator, std::string()> service, zoom, output, string, jsonp, checksum, location,
-        hint, stringwithDot, stringwithPercent, language, instruction, geometry, cmp, alt_route, u,
+        hint, timestamp, matching_beta, gps_precision, classify, stringwithDot, stringwithPercent, language, instruction, geometry, cmp, alt_route, u,
```
DennisOSRM: The order here should reflect the order in the grammar.

server/data_structures/internal_datafacade.hpp
```diff
@@ -415,6 +416,22 @@ template <class EdgeDataT> class InternalDataFacade final : public BaseDataFacad
             input_coordinate, resulting_phantom_node_vector, number_of_results);
     }
 
+    bool
+    IncrementalFindPhantomNodeForCoordinateWithMaxDistance(const FixedPointCoordinate &input_coordinate,
```
DennisOSRM: mark as `const override final`

server/data_structures/shared_datafacade.hpp
```diff
@@ -403,6 +404,22 @@ template <class EdgeDataT> class SharedDataFacade final : public BaseDataFacade<
             input_coordinate, resulting_phantom_node_vector, number_of_results);
     }
 
+    bool
+    IncrementalFindPhantomNodeForCoordinateWithMaxDistance(const FixedPointCoordinate &input_coordinate,
```
DennisOSRM: mark as `const override final`

util/json_logger.hpp
```diff
@@ -0,0 +1,77 @@
+/*
+
+Copyright (c) 2015, Project OSRM, Dennis Luxen, others
```
DennisOSRM: Copyright (c) 2015, Project OSRM contributors

util/json_logger.hpp
```diff
@@ -0,0 +1,77 @@
...
+#include <osrm/json_container.hpp>
+
+#include <boost/thread.hpp>
+
+namespace osrm
+{
+namespace json
+{
+
+class Logger
```
DennisOSRM: This class needs a comment or two to make the intent / usage more clear.

util/json_logger.hpp
```diff
@@ -0,0 +1,77 @@
...
+{
+    using MapT = std::unordered_map<std::string, osrm::json::Value>;
+
+  public:
+    static Logger* get()
+    {
+        static Logger logger;
+
+#ifdef NDEBUG
+        return nullptr;
```
DennisOSRM: This is potentially evil. Calling function from a `nullptr` is undefined behavior and thus needs checking code. Ideally. this class becomes empty when NDEBUG is defined.

util/json_logger.hpp
```diff
@@ -0,0 +1,77 @@
...
+    }
+
+    boost::thread_specific_ptr<MapT> map;
+};
+
+
+}
+}
+
+#endif /* SIMPLE_LOGGER_HPP */
```
DennisOSRM: JSON not SIMPLE ;-)

util/json_util.hpp
```diff
@@ -0,0 +1,101 @@
...
+(INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
+LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON
+ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+(INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+*/
+
+// based on
+// https://svn.apache.org/repos/asf/mesos/tags/release-0.9.0-incubating-RC0/src/common/json.hpp
```
DennisOSRM: is it?

util/json_util.hpp
```diff
@@ -0,0 +1,101 @@
...
+
+#ifndef JSON_UTIL_HPP
+#define JSON_UTIL_HPP
+
+#include <osrm/json_container.hpp>
+
+#include <limits>
+#include <cmath>
```
DennisOSRM: reorder includes.

util/json_util.hpp
```diff
@@ -0,0 +1,101 @@
...
+// Make sure we don't have inf and NaN values
+template <typename T> T clamp_float(T d)
+{
+    if (std::isnan(d) || std::numeric_limits<T>::infinity() == d)
+    {
+        return std::numeric_limits<T>::max();
+    }
+    if (-std::numeric_limits<T>::infinity() == d)
+    {
+        return -std::numeric_limits<T>::max();
```
DennisOSRM: use `std::numeric_limits<T>::lowest()`

CMakeLists.txt
```diff
@@ -69,14 +69,14 @@ file(GLOB PrepareGlob contractor/*.cpp data_structures/hilbert_value.cpp util/co
 set(PrepareSources prepare.cpp ${PrepareGlob})
 add_executable(osrm-prepare ${PrepareSources} $<TARGET_OBJECTS:FINGERPRINT> $<TARGET_OBJECTS:GITDESCRIPTION> $<TARGET_OBJECTS:COORDINATE> $<TARGET_OBJECTS:IMPORT> $<TARGET_OBJECTS:LOGGER> $<TARGET_OBJECTS:RESTRICTION> $<TARGET_OBJECTS:EXCEPTION> $<TARGET_OBJECTS:MERCATOR>)
 
-file(GLOB ServerGlob server/*.cpp)
+file(GLOB ServerGlob server/*.cpp util/compute_angle.cpp)
```
DennisOSRM: use an object library to avoid double-compile.

TheMarex:
:+1: for the refactor changes. Will have a go at the remaining points later.

DennisOSRM:
i am going to do an additional change of copy-edits today. We should we in a good place soon

routing_algorithms/map_matching.hpp
TheMarex: t can't be const, because it needs to be changed by the breakage-recovery-subroutine.

util/json_logger.hpp
TheMarex: Just making the methods empty does not work. The actually juicy part is the wrapped unordered_map. So we would need to have to wrap any access to that as well. 
What we want in Release mode, is a class that is basically a mock of this class and can completely be optimized out.

server/api_grammar.hpp
TheMarex: timestamp (```t```) behaves exactly like ```hint``` or ```u```. To fix the fact that you can specify e.g. ```classify``` more than once, we would need to change the query grammar in line 43.
Though currently all parameters behave that way.

DennisOSRM: ok, fair enough. IIRC, the reasoning back then was that garbage in -> garbage out. I think we can leave that as is.

DennisOSRM:
Happy to see how quickly this is moving along. Two sticking points remain though:

- The json logger component should work in a way that doesn't require checking for nullptrs. It should just do nothing if not initialized/c'ted.
- Do we have any tests that cover any of the new functionality?

TheMarex:
>- The json logger component should work in a way that doesn't require checking for nullptrs. It should just do nothing if not initialized/c'ted.

Let me think about this a little bit. Probably we can wrap the acces to the stored JSON::Objects and replace them by mock-functions in release mode.
Though I wouldn't say this is a blocker, as the code is not exactly widely used.

> Do we have any tests that cover any of the new functionality?

No, afaik we would need to extend the cucumber framework by the following functionality:
- Match query
- A way to specify an input trace with(out) timestamps

emiltin:
is matching documented in the wiki yet?

DennisOSRM:
> is matching documented in the wiki yet?

not yet. good point

emiltin:
what amount of gps points is this intended for? i guess using POST will quickly be required.

DennisOSRM:
yeah, the best way of accessing is thru nodejs bindings

TheMarex:
I was able to run queries with >100 points + timestamps without any problems so far. But yes, using the node bindings is highly advised in this case.

TheMarex:
@DennisOSRM added the tests. If travis comes back positive, its probbaly good to merge?

DennisOSRM:
:+1: restarted Travis, will merge once its fine.

DennisOSRM:
Great work!

PR #1409, opened 2015-03-05
FIX:

RockLobster:
Moved the GeometryCompressor's free_list_maximum into a member variable of the class.
(+UniqueNumber() commented out since it is not being used anywhere in all of OSRM)

There was absolutely no need for free_list_maximum to be a static variable and it caused major issues in my code.
Short description:
- I ran osrm-prepare's Prepare().Process(..) function in my own code multiple times for multiple files
	=> After the first file is done, free_list_maximum is still on the same size, but a new GeometryCompressor will start at a size of 100 again.
	=> Indices broken.

### DIFF
--- a/contractor/geometry_compressor.cpp
+++ b/contractor/geometry_compressor.cpp
@@ -35,8 +35,7 @@ SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 #include <limits>
 #include <string>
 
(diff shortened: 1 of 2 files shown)


DennisOSRM:
Thanks, once the CIs come back green, we can move ahead with this PR.

DennisOSRM:
Thanks again. Removed the lint manually.

Issue #1432, opened 2015-04-10
How to get node location in profile script


vesse:
I need to check external resource in `node_function` and would need the coordinates and ID of a node to perform the check. Is the location of a node somehow accessible in the Lua script? Are the members and methods of the `way` and `node` objects documented somewhere?

Thanks!

TheMarex:
Thanks for reporting this.Just added a PR to fix this. In general we don't have good documentation on this. You can find most information here https://github.com/Project-OSRM/osrm-backend/blob/develop/extractor/scripting_environment.cpp#L79

Maybe you want to help improving the documentation by adding that to the wiki?

vesse:
Thanks! Maybe add also `.def("id", &osmium::Node::id)` while the PR is still open?

I'll see if I could help with the documentation once I have more experience with OSRM.

DennisOSRM:
Closing here, tracking progress in #1433

PR #1433, opened 2015-04-10
Make node location accessible

TheMarex:
See issue #1432.


DennisOSRM:
From #1432: 

> Maybe add also `.def("id", &osmium::Node::id)` while the PR is still open?

DennisOSRM:
@TheMarex Made the node available, too. This should be good to merge.

ujwalaz:
Can i get node id / coordinates of nodes in profile scripts? This is required for matching purpose to update speed for particular way. I tried to update from process_segment but there also i can update weight and duration but not exact speed. Please let me know how can i get location of way.

PR #1437, opened 2015-04-13
Feature/matching classification

TheMarex:
Update the parameters used in the classification code. Should now work better after the trace splitting addition.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -27,6 +27,7 @@ if(WIN32 AND MSVC_VERSION LESS 1800)
   message(FATAL_ERROR "Building with Microsoft compiler needs Visual Studio 2013 or later (Express version works too)")
 endif()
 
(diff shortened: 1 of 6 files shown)


emiltin:
how does classification work?

TheMarex:
@emiltin The classification is important when trying to figure out if the map matching failed. This will happen if the trace for example uses a road that is not in the OSRM data. Sometimes the trace will also be too noisy and the resulting matching will have a lot of strange loops. The classification basically compares the difference between the total length of the trace and the total length of the matching. If both are about the same, the matching has a high probability of being correct.

The actual parameters for the classifier where derived from manually labeled traces using ```bin/calibrate_classification.py``` in https://github.com/mapbox/osrm-matching-inspection.
I can't disclose the actual traces I used to derive this values because of the usual copyright and privacy concerns.

emiltin:
thanks

PR #1440, opened 2015-04-16
fix incorrect SCC generation

### LINKED ISSUE #1427: Issue with nearest-neighbor search, i.e. SCC traversal of the road network
Currently blocking the release of 4.6.0:

During preprocessing, the road network is explored to identify _all_ strongly connected components. There appears to be an issue that does not properly identify all segments/SCCs correctly. Here's an example. The entire one-way should be discovered as a string of SCCs:

screenshot 2015-04-01 17 18 26

The code didn't change drastically over time. So, the issue is most probably something minor such as an off-by-one or similar error.

DennisOSRM:
fixes #1427 and also introduces a tool to check graph implementation behavior on real data files.

CC:  @TheMarex


TheMarex:
Will review this later today and merge once travis comes back positive. :+1:

TheMarex:
While this fixed the problem, returning ```EndEdges``` is a bad idea, it breaks ```FindEdgeInEigtherDirection```. Some parts of the code where even assuming that ```DynamicGraph``` would return ```SPECIAL_EDGEID``` if no edge was found. Somehow it looked like a merge gone bad in ```EdgeBasedGraphFactory```.

Anyway usage should now be consistent cross the whole code base. Hopefully nothing broke in the process.

DennisOSRM:
The semantics used to be that it returned an ID greater or equal to `EndEdges()`, but returning `SPECIAL_EDGEID` may make the intent more clear.

PR #1456, opened 2015-05-01
Fix travis links

willwhite:
Fix travis links in README after project rename.


TheMarex:
Merged this by hand e2954211ecacb0b265820daae654b114f58a3298 as this PR was against master and not develop :-) Otherwise :+1:

Issue #1454, opened 2015-04-29
Testing setup


lbud:
@emiltin — 

I'm eager to get tests up and running/100% passing in order to work on contributing and have been hitting some walls. Yesterday @springmeyer and I worked on bootstrapping the initial setup on the [`masonize`](https://github.com/Project-OSRM/osrm-backend/compare/masonize) branch, which led us down quite a few Lua/Luabind/LuaJIT wormholes that seemed *maybe* OS-specific (we are both running OSX). Now I'm running Lua 5.1.0 with Luabind and without LuaJIT; Lua 5.3.0 led to Lua compile errors failing all cucumber tests. After figuring out which versions of Lua/Luabind to use my `cucumber` results (sans nice colors) are [here](https://gist.github.com/lbud/814f98e35c144dd9a2ab) — you can see at the bottom I've got
```
369 scenarios (33 failed, 336 passed)
1539 steps (33 failed, 1506 passed)
```
I'm trying to figure out the following:
* Why would my results be differing/not 100% passing? (for example: )
![image](https://cloud.githubusercontent.com/assets/3170312/7403191/4688527c-eea1-11e4-9650-c26567ec3887.png)
![image](https://cloud.githubusercontent.com/assets/3170312/7403198/51ea8662-eea1-11e4-93a3-9b9179136f9e.png)
* Could this be related to Lua or cucumber versions? (I also had a hard time getting cucumber to run on Ruby 2.+ and had to downgrade to 1.9.3 to run it)
* Could this be OS-specific?

Curious to see if you have any hunches here. Thanks!!
cc @TheMarex

emiltin:
it looks to me like wrong routes are in fact returned, rather than a problem with ruby or the test framework.

could there be different version of osrm running?
could there be different/wrong datasets loaded through osrm-datastore?

latest develop branch works for me on both ubuntu 14 and os x. 10.9, with all tests passing. i'm using lua 5.2 on both.

emiltin:
one major difference seems to be that i'm running lua 5.2 while you're using 5.1.

springmeyer:
@emiltin thanks for the feedback and ideas! We'll try using Lua 5.2 as a next step. The reason we chose 5.1 was that I we realized that luajit would be a good thing to try eventually and luajit says it is compatible with lua 5.1.

@emiltin - as far as your general OS X setup: what version of ruby do you use, and how do you install it and other osrm-backend deps? Do you mainly use homebrew or source installs?

[...]

Issue #1442, opened 2015-04-17
cmake/FindLuaJIT.cmake does not find library on Ubuntu 14.04


woodbri:
The lib files get installed at:

```
/usr/lib/x86_64-linux-gnu/libluajit-5.1.a
/usr/lib/x86_64-linux-gnu/libluajit-5.1.so
/usr/lib/x86_64-linux-gnu/libluajit-5.1.so.2
/usr/lib/x86_64-linux-gnu/libluajit-5.1.so.2.0.2
/usr/lib/x86_64-linux-gnu/pkgconfig/luajit.pc
```

I think cmake/FindLuaJIT.cmake line 58+ needs to change some how but not sure what is best to do here so this path gets checked also.

This package is installed via the apt-get package manager.

```
$ apt-show-versions | grep luajit
libluajit-5.1-2:amd64/trusty 2.0.2+dfsg-1 uptodate
libluajit-5.1-common:all/trusty 2.0.2+dfsg-1 uptodate
libluajit-5.1-dev:amd64/trusty 2.0.2+dfsg-1 uptodate
```

DennisOSRM:
Works fine on our development and CI machines on Ubuntu 14.04. Are you using develop or master branch?

woodbri:
Using develop branch

woodbri:
```
[develop] ~/work/osrm-backend/build$ cmake ..
-- The C compiler identification is GNU 4.8.2
-- The CXX compiler identification is GNU 4.8.2
-- Check for working C compiler: /usr/bin/cc
-- Check for working C compiler: /usr/bin/cc -- works
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Check for working CXX compiler: /usr/bin/c++
-- Check for working CXX compiler: /usr/bin/c++ -- works
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Building on a 64 bit system
-- Configuring OSRM in release mode
-- Performing Test LTO_AVAILABLE
-- Performing Test LTO_AVAILABLE - Success
-- Performing Test LTO_WORKS
-- Performing Test LTO_WORKS - Success
-- LTO working
-- Performing Test HAS_COLOR_FLAG
-- Performing Test HAS_COLOR_FLAG - Failed
-- Boost version: 1.54.0
-- Found the following Boost libraries:
--   date_time
--   filesystem
--   iostreams
--   program_options
--   regex
--   system
--   thread
[...]

PR #1457, opened 2015-05-01
Modify profiles to use Lua 5.2+ without needing compatibility flags

lbud:
Since the deprecation of `module` in Lua 5.2+, the current approach only works in 5.2 with compatibility flags, and doesn't work in Lua 5.3 at all. This alternate approach allows us to use Lua 5.2 with or without compatibility flags and fixes issues raised in https://github.com/Project-OSRM/osrm-backend/issues/1454.

Note: the [change](https://github.com/Project-OSRM/osrm-backend/blob/f6cc4a652060be504e4fbce79936aee3a7989d44/profiles/lib/access.lua#L12) I introduce here to `lib/access.lua`'s nil return reflects the way it [had been inlined in car.lua](https://github.com/Project-OSRM/osrm-backend/blob/11bdfba37e3e1e2b4e3490ccec492d59b4b42397/profiles/car.lua#L157) — before changing the `lib/access.lua` value, two car barrier tests (features/car/barrier.feature) were failing; after the change all cucumber tests pass. I'm not entirely confident as to why this is as I am by no means a lua expert… @emiltin would you mind taking a look?

I'm curious to see how upgrading affects travis (especially wrt questions about using LuaJIT, https://github.com/Project-OSRM/osrm-backend/issues/1442) (I'm developing on OSX locally).
cc @TheMarex @springmeyer @willwhite

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,7 +7,7 @@ install:
  - sudo apt-add-repository -y ppa:ubuntu-toolchain-r/test
  - sudo add-apt-repository -y ppa:boost-latest/ppa
  - sudo apt-get update >/dev/null
- - sudo apt-get -q install protobuf-compiler libprotoc-dev libprotobuf7 libprotobuf-dev libbz2-dev libstxxl-dev libstxxl1 libxml2-dev libzip-dev lua5.1 liblua5.1-0-dev rubygems libtbb-dev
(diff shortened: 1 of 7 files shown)


TheMarex:
Todo:
- [ ] Figure out why travis does not find lua
- [ ] Change luaAddScriptFolderToLoadPath to check if it is a symbolic link and also append the expanded path

lbud:
@TheMarex ah, the cmake files are still looking for 5.1 — I'll fix.

TheMarex:
Sorry dropped the ball here. Will rebase and add following symbolic links.

TheMarex:
Looks good. Merging.

PR #1455, opened 2015-04-30
upgrade to cucumber 2.0

emiltin:
Upgrades to cucumber 2.0.
Remember to run 'bundle' upgrade cucumber and related gems.
Tested with ruby 2.1.5p273 and 2.2.0p0, all tests passing.

- [ ] test with ruby 1.9
- [ ] test with ruby 2.2
- [ ] check if travis and appveyor can use ruby 2.2, if yes then we could require ruby 2.2 in the Gemfile to avoid incompatibilities with older ruby versions
- [ ] make sure CIs pass all tests

PR #1465, opened 2015-05-09
fix cucumber tests running on Windows

alex85k:
https://github.com/Project-OSRM/osrm-backend/pull/1455

Running tests on Windows (Ruby 1.9.3 32 bit) caused the errors in file checking (.exe was missing).
Here is the possible patch.


alex85k:
Forgot to mention: all the tests passed successfully with the patch applied (Release build, Visual Studio 2015 RC, Boost 1.58).

TheMarex:
Looks good. :+1:

PR #1478, opened 2015-05-15
Cleanup the profiles

TheMarex:
@emiltin could you have a look at the changes I made to the bike profile? I added safety penalties and disables the usage of public transportation by default.

### DIFF
--- a/profiles/bicycle.lua
+++ b/profiles/bicycle.lua
@@ -12,10 +12,10 @@ access_tags_hierachy = { "bicycle", "vehicle", "access" }
 cycleway_tags = {["track"]=true,["lane"]=true,["opposite"]=true,["opposite_lane"]=true,["opposite_track"]=true,["share_busway"]=true,["sharrow"]=true,["shared"]=true }
 service_tag_restricted = { ["parking_aisle"] = true }
 restriction_exception_tags = { "bicycle", "vehicle", "access" }
+unsafe_highway_list = { ["primary"] = true, ["secondary"] = true, ["tertiary"] = true, ["primary_link"] = true, ["secondary_link"] = true, ["tertiary_link"] = true}
 
-default_speed = 15
-
-walking_speed = 6
+local default_speed = 15
+local walking_speed = 6
 
 bicycle_speeds = {
   ["cycleway"] = default_speed,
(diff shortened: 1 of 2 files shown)


emiltin:
Adding a penally for all primary, secondary and tertiary because they're unsafe seems a bit drastic to me. For example, in Copenhagen, most of such road have bike lanes (separated from car traffic with a raised curbs), and would not be considered unsafe. Most adults would choose them for going to school and work, for example.

Nørrebrogade is one of the busiest bike streets in Copenhagen, and is tagged as a tertiary: http://www.openstreetmap.org/way/27432854. Parts of it don't even allow cars.

One of the busiest car roads is also tagged as tertiary: http://www.openstreetmap.org/way/8122273. Even this road is not considered unsafe for bikes, since there are bike tracks, and in many places the bike tracks are even separated from car lanes with trees.

For this reason I would recommend considering at least the 'cycleway' tag. Note that this tag is particular tricky since you can have 'opposite_lane/track' and this can be combimed with oneway tags.

You migh also consider access tags for cars and other motor vehicles; if they're not allowed, it should be safer. Maxspeed for cars would be another thing to look at.

I would also treat primary/secondary/tertiary different.

Yes, ike routing is generally more complex that cars :-)

Another issues with all this is the fact that speed and impedance is not separated, so the only way you can prefer certain streets is to manipulate the speed, resulting in abnormal travel times. Just because a street is unsafe doesn't mean to go slow.

emiltin:
Bringing bikes on trains and metros is debatable, but I think it's fine to disable that be default. In any case it's not really working as it should, I think because some train lines are tagged using relations rather than using the route tag.

emiltin:
Btw, I think it would be great to refactor the profiles and create some common structure for the usual processing flow of access, oneway, surface, speeds, etc.

TheMarex:
Thanks for the feedback @emiltin! I don't really want to change the default behavior, just make it easier in general to configure stuff. I'm not too happy about the penalties as well, as it skews the ETAs a lot and I hope we can separate the travel time and edge weight in the future. But it is seems like good heuristic, especially in the US.

@systemed has probably a lot of experience tuning penalties like that for GB.

> For this reason I would recommend considering at least the 'cycleway' tag. Note that this tag is particular tricky since you can have 'opposite_lane/track' and this can be combimed with oneway tags.

The penalty only gets applied when the road is not tagged with `cycleway`, because the speed gets overwritten [here](https://github.com/Project-OSRM/osrm-backend/blob/a309fbc7658a080f64731715e645cd24dfbe0994/profiles/bicycle.lua#L366).

But yeah the control flow is totally confusing. I wonder how we could make this hierarchy more explicit.

TheMarex:
Tests pass, I'm going to merge this first version.

PR #1485, opened 2015-05-21
Return error message when lua error occurs.

### LINKED ISSUE #1467: Show Lua runtime errors during extraction
At present Lua runtime errors encountered during extraction just throw a generic "Lua runtime error".

During contraction, however, the runtime error is reported fully:

    if (0 != luaL_dofile(lua_state, profile_path.string().c_str()))
    {
        std::cerr << lua_tostring(lua_state, -1) << " occured in scripting block" << std::endl;
        return false;
    }

I think OSRM used to show Lua errors during extraction too - I found this code in an old version:

	try {
		...
	} catch (const luabind::error &er) {
		std::cerr << er.what() << std::endl;
		lua_State* Ler = er.state();
		report_errors(Ler, -1);
	}

	...
	void BaseParser::report_errors(lua_State *L, const int status) const {
	    if( 0!=status ) {
	        std::cerr << "-- " << lua_tostring(L, -1) << std::endl;
	        lua_pop(L, 1); // remove error message
	    }
	}

danpat:
This change gets Lua error messages visible on the console again.  When inside the `lua_error_callback()`, the error message may not be the first item on the stack.

This fixes #1467.

### DIFF
--- a/extractor/restriction_parser.cpp
+++ b/extractor/restriction_parser.cpp
@@ -44,7 +44,7 @@ namespace
 {
 int lua_error_callback(lua_State *lua_state)
 {
-    luabind::object error_msg(luabind::from_stack(lua_state, -1));
+    std::string error_msg = lua_tostring(lua_state, -1);
     std::ostringstream error_stream;
     error_stream << error_msg;
     throw osrm::exception("ERROR occured in profile script:\n" + error_stream.str());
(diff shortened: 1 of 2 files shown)


TheMarex:
Perfect @danpat :+1:

systemed:
\o/

PR #1493, opened 2015-05-24
Check for barriers and turn restrictions in TarjanSCC

TheMarex:
This fixes missing small components.

### DIFF
--- a/algorithms/tiny_components.hpp
+++ b/algorithms/tiny_components.hpp
@@ -97,20 +97,20 @@ template <typename GraphT> class TarjanSCC
         TIMER_START(SCC_RUN);
         const NodeID max_node_id = m_node_based_graph->GetNumberOfNodes();
 
-        // The following is a hack to distinguish between stuff that happens
-        // before the recursive call and stuff that happens after
         std::stack<TarjanStackFrame> recursion_stack;
-        // true = stuff before, false = stuff after call
         std::stack<NodeID> tarjan_stack;
         std::vector<TarjanNode> tarjan_node_list(max_node_id);
         unsigned component_index = 0, size_of_current_component = 0;
         unsigned index = 0;
+        // The following is a hack to distinguish between stuff that happens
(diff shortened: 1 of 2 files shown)


TheMarex:
The [Tarjan algorithm](https://en.wikipedia.org/wiki/Tarjan's_strongly_connected_components_algorithm) for finding strongly connected components is not compatible with turn restrictions. Counter example below:

Graph:
```
a----b----c
     |
     d
```
With an `only_straight_on` restriction on (a,b).

If we start the algorithm at `a` it will find `b`, add it to the stack. We continue with `b` which finds `c` but will skip `d` since a the turn `(a, b, d)` is not allowed. We continue with `c` which will find `b` and update its component index. Back to `b` which finds `a` and updates its component index. At `a` we have completed the component search and `d` is never visited, even though there is path `(a, b, c, b, d)` and `(d, b, a)` in the component.

The main problem is that we check if there is a restriction regarding `(a, b, d)` (th at is a restriction starting at the parent of `b` and using `b` as a via). If we would know that there is a neighbor (e.g. `c`) of `b` that belongs to the same SCC and for which the turn `(c, b, d)` is allowed, we could safely explore `d`.

There are two points in the code where we know we found an neighbor in the same SCC:
- While we scan the edges of `b`, we find a neighbor that is already on the stack
- When we get back from the recursion and update the component id of the parent (which always is a neighbor)

In each case it would cost us an additional scan of all neighbor and turn restriction lookups. The main problem is that I'm not sure if this is _sufficient_ to find all ways.

TheMarex:
I added a breaking test case.

TheMarex:
Took another look at this, looks like we have to run TarjanSCC on the edge-expanded graph. This requires some bigger restructuring (for one we have to construct the actual DynamicGraph outside of the contractor). This might also have some ramification on the memory usage.

TheMarex:
Closing superseded by https://github.com/Project-OSRM/osrm-backend/pull/1565

Issue #1429, opened 2015-04-03
Route calculation is dependant on route changes in unrelated area


MaartenDeen:
I apologise for the maybe cryptic subject. I am noticing a change in the route calculation in unrelated areas when you change the parameters of the route in other areas.
Consider these two routes (screenshots are also attached): 
1) http://www.osrm.at/bDs
2) http://www.osrm.at/bDt

1) is the route that osrm calculates without intermediate points and takes a shortest path in the village Lierop. For 2) I added an intermediate point in an area that is not related to the route in the village (it is not on a track that should be affected by the new point) and the new route calculates a different route in the village Lierop which is about 500 meter longer.
Looking at the speed information from the point of change, 1) is about 340 meters @ 30 km/h, 2) is about 145 m @ 50 km/h and 680 m @ 30 km/h. This can never be a faster route.

Why does the route calculation change when I put an intermediate point in a non-related area?

![route1](https://cloud.githubusercontent.com/assets/1825970/6980781/a6eafec6-d9f7-11e4-99e9-9ebcfb5a8947.png)
![route2](https://cloud.githubusercontent.com/assets/1825970/6980783/a8bf92de-d9f7-11e4-8b21-5b24d4362636.png)

danpat:
Possibly related test case.  Here, the intermediate marker is moved around a corner that is on the path in both cases.  In the second case, the routing seems reasonable, in the first case, the routing does a complete lap of the block to arrive at the point.

![screen shot 2015-04-13 at 5 20 57 pm](https://cloud.githubusercontent.com/assets/1892250/7128039/920217b2-e201-11e4-8a42-6b57cf556a84.png)

![screen shot 2015-04-13 at 5 22 03 pm](https://cloud.githubusercontent.com/assets/1892250/7128043/ab23317c-e201-11e4-9194-453b61a704da.png)

woodbri:
Please supply links from the [Generate Link] link so we can have a link example of these last two. Thanks.

danpat:
Here you go:

http://osrm.at/bOG
http://osrm.at/bOF

danpat:
Here's a minimal case from the original report.  By moving the intermediate marker slightly, the route changes in unexpected ways.

I loaded a 10x10km OSM extract for this region of the Netherlands into a local OSRM install, and I can't reproduce the problem based on HEAD from master.  If I place markers in the same spots, the routing does *not* jump around like it does on http://map.project-osrm.org/.

[...]

Issue #1424, opened 2015-03-26
Strange error in route with intermediate point and one-way street


fpacheco:
One-way street with an intermediate point located in the same block as the destination. 

http://osrm.at/bys

How can it be avoided?

TheMarex:
Sorry, I don't understand what is wrong with that route. Can you elaborate a bit more? Thanks!

fpacheco:
Sorry, I sent the wrong link!.

This is one that correctly describes the problem: http://osrm.at/byx
Or this other: http://osrm.at/byy

Thank you. Fernando.

fpacheco:
For the first link I expect something like this ...  
![osrm_test_3](https://cloud.githubusercontent.com/assets/89121/6856915/dc36f8f8-d3e1-11e4-9ede-412349b08d47.png)

DennisOSRM:
Interesting bug. We'll look into that.

DennisOSRM:
I have found a (minor) issue that caused this odd behavior. I am working on a fix for this now. With that fix, routes will look like this

![screenshot from 2015-03-27 10 18 13](https://cloud.githubusercontent.com/assets/1067895/6865102/c5e1eff2-d46a-11e4-967c-0c305414cef9.png)

instead of this

![screenshot from 2015-03-27 10 18 30](https://cloud.githubusercontent.com/assets/1067895/6865101/c14b0a78-d46a-11e4-93e3-1c5b6439f12e.png)

I'll push the fix later today to develop branch but will also take some extra care to make sure that the fix doesnt have any unexpected side effects.

DennisOSRM:
Fix has been published. Will close when the next full update cycle completes and the fix is live on the demo site. 

Thanks again for reporting @fpacheco

DennisOSRM:
All good now:

![screenshot 2015-03-30 09 20 10](https://cloud.githubusercontent.com/assets/1067895/6891752/12446410-d6be-11e4-9f48-b688932e6d57.png)
![screenshot 2015-03-30 09 20 20](https://cloud.githubusercontent.com/assets/1067895/6891753/1269c71e-d6be-11e4-9fcf-59d27e99e639.png)

danpat:
Looks like this change created #1429.

DennisOSRM:
@danpat thanks for the pointer. Will investigate.

TheMarex:
Has to revert the fix because it caused regression.

woodbri:
[...]

PR #1497, opened 2015-05-26
Partially fix via-routing when vias are on the same way

danpat:
This small change partly addresses #1424, and does not seem to re-introduce #1429.

When two vias are placed on the same way (NodeID), it causes the reverse heap test (`WasInserted`) to return true in the first round call to `RoutingStep`.  When a previous leg has a distance offset in place, the loop thinks the two ends of the bi-directional Dijkstra search have met and no further edges are visited, the two vias are routed directly, even if the way is marked as one-way, and the vias are not visitable in their placed order.

This change adds an explicit test for vias on the same way.  When detected, the distance offset is not applied to the nodes inserted into the search heaps.  This leads to proper traversal of the edge graph.

One point of note:  the `raw_route_data.shortest_path_length` value is not correct, further changes are needed to make that bit work too.

Additionally, while the tests in `features/testbot/via.feature` address this bug specifically, I could not get them all to work.   Only two of the tests in the `Via points on ring on the same oneway` scenario work, the other three are returning no results to the test suite, although manual testing of these scenarios seems to work OK.  I'm not familiar enough with cucumber to dig all the way into it at this stage.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -74,6 +74,7 @@ file(GLOB DatastructureGlob data_structures/search_engine_data.cpp data_structur
 list(REMOVE_ITEM DatastructureGlob data_structures/Coordinate.cpp)
 file(GLOB CoordinateGlob data_structures/coordinate*.cpp)
 file(GLOB AlgorithmGlob algorithms/*.cpp)
+file(GLOB RoutingAlgorithmGlob routing_algorithms/*.hpp)
 file(GLOB HttpGlob server/http/*.cpp)
 file(GLOB LibOSRMGlob library/*.cpp)
 file(GLOB DataStructureTestsGlob unit_tests/data_structures/*.cpp data_structures/hilbert_value.cpp)
@@ -85,6 +86,7 @@ set(
   ${DescriptorGlob}
   ${DatastructureGlob}
   ${AlgorithmGlob}
+  ${RoutingAlgorithmGlob}
   ${HttpGlob}
 )
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -74,7 +74,6 @@ file(GLOB DatastructureGlob data_structures/search_engine_data.cpp data_structur
 list(REMOVE_ITEM DatastructureGlob data_structures/Coordinate.cpp)
 file(GLOB CoordinateGlob data_structures/coordinate*.cpp)
 file(GLOB AlgorithmGlob algorithms/*.cpp)
-file(GLOB RoutingAlgorithmGlob routing_algorithms/*.hpp)
 file(GLOB HttpGlob server/http/*.cpp)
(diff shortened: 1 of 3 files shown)


TheMarex:
@danpat awesome work figuring this one out. Will start to review this now.

> One point of note: the raw_route_data.shortest_path_length value is not correct, further changes are needed to make that bit work too.

Hm, just subtracting the edge offsets from the overall distance should work. But I need to check that.

danpat:
My gut tells me that this solution is a hack, and it should be possible to make it work more in-line with Dennis' original change, using the phantom node weights to get proper ordering.  However, my brain isn't quite grokking all the interactions and I couldn't come up with anything else that would work.

The source of the "weird" paths seen in #1429 is because when cumulative `distance1` and `distance2` are used, the post-loop swap here:

https://github.com/Project-OSRM/osrm-backend/blob/master/routing_algorithms/shortest_path.hpp#L314

is sometimes called incorrectly.

TheMarex:
Hm, so the reason your fix works is, that it triggers [this condition](https://github.com/danpat/osrm-backend/blob/17dd39609f7325defbeecc2217804e0773cc0e76/routing_algorithms/routing_base.hpp#L81), so the `upper_bound` gets not set to 0 and we don't terminate the search.

If the street segment is actually bidirectional this should lead to incorrect results. Because we need to actually terminate at that position, right?

Regarding the cumulative `distance1` and `distance2` values, I think I get what he was trying to do:
- Dijkstra does not care which initial distance value the start node has: it is just an offset
- You want to get the overall shortest route, not just the shortest routes between two vias. But that depends on the current best distance to the forward and reverse node of the source
- But if you want to accumulate the `local_upper_bound`s (the distance between the vias) you can't offset the initial distance value, thus the removal

Ideas why Dennis fix is broken:
There is something that actually relies on `distance1` and `distance2` being the distance to the previous via's forward/reverse node. Maybe the wrong via leg is chosen?

TheMarex:
In conclusion I think we should maybe start working from the original patch and exactly figure out why it is broken: That is try to construct a minimal lua test case. I still have an extract from Lierop which is mentioned here https://github.com/Project-OSRM/osrm-backend/issues/1429, maybe I can reduce the data in JOSM to a test case.

danpat:
Agreed.  Now that I'm a bit more familiar with the cucumber test framework, I'll see if I can build a failing test.

TheMarex:
@danpat This is a smaller extract from Lierop where you could try to find a test case for #1429 https://drive.google.com/file/d/0B0Q8fvc1VRlDNW9hVE5pd09xVzQ/view?usp=sharing

emiltin:
note that you can load osm or pbf files from cucumber. that way you can start with an actual .osm extract, and reduce it in steps.

TheMarex:
Closing in favor of new PR.

Issue #1293, opened 2014-11-25
accept HTTP POST requests


Carsten64:
Hello together,

I'm facing a problem with the length of URL by requesting the OSRM services , I get a HTTP Error 414, that means that my URL is too long. I was reviewing the code and found in the file [Connection.h] (https://github.com/Project-OSRM/osrm-backend/blob/master/Server/Connection.h) in line 98 the following code snippet:
<b>boost::array<char, 8192> incoming_data_buffer; </b>.

Is this the place where I can change the value to more KB per request, for example 32 kb?

kind regards

Carsten

emiltin:
duplicate of #205? 
maybe rename this issue to soemthing like "accept HTTP POSTrequests"?

DennisOSRM:
renamed and closed the other issue.

solidsnake2013:
oh! we every time use point of 300-400.
and we We also need to use [table] and [viaroute].

There is no other way?

TheMarex:
@solidsnake2013 at this point I would suggest building every serious application on top of `node-osrm` and not using the HTTP interface directly. `osrm-routed` is super convenient for debugging and small deployments, but does not scale to large requests.

emiltin:
is ndoe-osrm described in the wiki?

TheMarex:
@emiltin [yep](https://github.com/Project-OSRM/osrm-backend/wiki/Library-api#nodejs-language-bindings) but I should move that on the main page.

TheMarex:
This was implemented.

PR #1502, opened 2015-05-28
Accepting HTTP Post Request

agruss:
This is a solution for accepting HTTP post requests. You can send them described as here: http://en.wikipedia.org/wiki/POST_%28HTTP%29#Use_for_submitting_web_forms

Still missing are tests for the test suite (there seems to be some problems with the methods, the only way to send a post request is via post_form() which malformed the requests).


TheMarex:
@agruss awesome, thanks! Will review that later today. :+1:

@emiltin any idea why `post_form` is broken in this case?

emiltin:
How is the request sent by post_form() malformed?

agruss:
@emiltin When I try to send a post_form() the values from the data parameter are not the same, e.g. I want to send "loc=1.0,1.9876543", but Wireshark shows me some other value was sent like "loc=1.0,19800032" or it cuts them halfway off. It's kinda strange, this never happend when I tested it with the javascript implementation I wrote.

TheMarex:
@agruss can you post an example function call? That would be extremely helpful for debugging. Otherwise I'm :+1: on this. This will fix https://github.com/Project-OSRM/osrm-backend/issues/1293

emiltin:
sounds weird with post_form(), not sure why that would happen.

agruss:
@TheMarex I can give you an example function for javascript.
The one I used is:
```
function sendRequest(service, parameters) {	
	var xmlhttp = new XMLHttpRequest();
	xmlhttp.open("POST","http://127.0.0.1:5000/" + service + "?",true);
	var serverResponse;
	xmlhttp.onreadystatechange = function() {
		if (xmlhttp.readyState === 4)  {
			serverResponse = xmlhttp.responseText;
			console.log(serverResponse);
		}
	};
	xmlhttp.send(parameters);
}
		
var service = "locate";
var parameters = "loc=51.984080000,7.637920000";
sendRequest(service, parameters);
```
You can change easily the service and parameters and try a little bit.
About the empty request: I think you mean when you try to send a http://127.0.0.1:5000/match or another service and it gives you just an empty json object back? It's the same behavior as the get request. You need a ? at the end. To be honest I dont't really know why you need it, the parser never checks for a ? at this position, only for a space.

agruss:
Okay I found something out about the behavior from post_form():
When I try to send a POST request Ruby seems to "separate" the message body and the header fields. To be clear, it sends two packages. The first one has no message body but the header fields, the second one on the other hand has the message body, but no header fields. This is a problem, because the parser goes first through the whole header fields and then through the message body for the POST request.
I'm not really into Ruby so I have no idea how to send only one package with all the informations in it.

danpat:
@agruss: The following patch solves the spanning-multiple packet problem:

https://github.com/danpat/osrm-backend/commit/672dbddb0528d501ade931c92f51b13f142cfb45

Spanning multiple packets is perfectly valid behaviour, the packet-parsing termination condition wasn't quite right, we could receive any number of bytes for each call to `RequestParser::parse`.

Two changes:

  - Need to inject a `?` character into the `current_request.uri` when staring to read the POST body.  The API grammer is setup to parse `GET` request URIs, which have a `?` separator.  We're piggybacking on that behaviour, so we need to insert the missing piece.
  - Honour the `Content-Length` header for reading the `POST` body, rather than assuming that we're done when we've read the current TCP packet (or whatever `TCP_socket.async_read_some` gives us, which could be any number of bytes).

Couple of things need to be done to complete this:

  1.  It does not handle `POST` requests where `Content-Type` is `multipart/form-data`.  That's probably OK.
  2.  Need to validate that `Content-Length` is a positive number, or 0.

emiltin:
we should also have cucumber tests

danpat:
Re-reading my patch, the end-of-HTTP-request termination logic still isn't quite right.  Under most sane circumstances, we'll probably get the request headers in a single packet.  However, if that doesn't happen (i.e. very small network MTU) for some reason, my patch might still report the packet completed when all the headers haven't been read yet.  I think that condition in `RequestParser::parse` should read:

    if (state == internal_state::post_request && content_length <= 0)
    {
        result = osrm::tribool::yes;
    }

This scenario is a bit more difficult to create though, it requires sending the HTTP request in partial chunks so that `TCP_socket.async_read_some` triggers even before the headers are complete.

@agruss if you integrate my patch into your pull request, I'd update `RequestParser::parse` with the code snippet above.

agruss:
@danpat Oh sorry I didn't read your last comment. I updated your code with the last commit.
I also tried a Content-Type check. Is a ```return osrm::tribool::no;``` enough? The server should respond  with a bad request, shouldn't it?

danpat:
Yes.   We should only accept `application/x-www-form-urlencoded` encoded `POST` requests, because those are the only ones we can parse.  I think `return osrm::tribool::no;` should be right.

TheMarex:
@agruss Do you now want to try again to add a cucumber test for this? This should work now with multi-package support, right? Apart from that this looks good to merge. :rocket:

agruss:
Okay I added some tests... I'm not really satisfied with the code I wrote maybe someone can tell me what I can improve (or is it okay for this purpose)?

emiltin:
i think there's a bit too much code duplication. it migh be better to add a step definition that simply let's you choose between get/post, like you can already choose input format, etc.

agruss:
@emiltin  Thanks for the help. I changed my tests. 
Now you can do 
```
And the query options
| post | true |
```
and the test will be a POST request.
I think this is what you suggested?

emiltin:
better!
but i think it can be improved a bit more by defining something like:

```ruby
Given /^the HTTP method "([^"]*)"$/ do |method|
  @http_method = method
end
```

that way you can use it like below. note that you can place it in the background section if you want to use it for all scenarios in the feature file:

```gherkin
    Background:
        Given the profile "testbot"
        And the HTTP method "POST"
```

also the code that handles post/get is still duplicated, and should preferable be factored out to a common class or method.

TheMarex:
@agruss tests look much better. :+1: ready to merge @emiltin?

RhinoDevel:
It is really great that POST requests are supported (although [#1293](https://github.com/Project-OSRM/osrm-backend/issues/1293) is still open).

@agruss's [JavaScript sample code above](https://github.com/Project-OSRM/osrm-backend/pull/1502#issuecomment-106726784) needs to be modified just a little bit to work with (e.g.) OSRM v4.9.0 - here is an updated version:

    var req = new XMLHttpRequest();

    req.open('POST', 'http://127.0.0.1:5000/nearest', true);
    req.setRequestHeader('Content-Type', 'application/x-www-form-urlencoded');
    req.onreadystatechange = function()
    {
        if(req.readyState===4)
        {
            console.log(req.responseText);
        }
    }
    req.send('loc=53.552858,10.006695');

Required changes:
- replaced `locate` with supported service `nearest`
- removed `?` at end of URL
- explicitly set `Content-Type` to `application/x-www-form-urlencoded`

TheMarex:
@RhinoDevel thanks for clearing this up. :+1: Good find with the issue.

PR #1509, opened 2015-05-31
Matching between separate splits

agruss:
Okay, I don't know if this is really a fix, because I'm not quite sure why it is implemented in this way (so this pull request may be unnecessary).
The map matching algorithm combines the geometry from each split by just linking them together. If there is an obstacle, e.g. a building between these two splits, the geometry would go through it.
This fix (if you can really call it like that) matches all the geometries by collecting the whole data first and then finding the shortest path.


TheMarex:
Hm, so your PR does a shortest path calculation across all submatchings? This might be a result of some missing documentation. Let me clear that up why there are "missing" geometry segments.

If the matching algorithm decides that a segment of the input trace can not be matched, it is removed from the input trace and the matching is split at that point. That means we have no idea what happens between two end-points of sub-matchings.

Just joining sub-matchings with a shortest-path calculation has the assumption that the vehicle took that route. But really, the whole point why we had to split it, is because we have no data to prove that.

You should threat each sub-matching as completely independent and don't join their geometries.

agruss:
@TheMarex Ah okay! The results of my solution would be something like "assume that the vehicle took that route, but in the real route it went a completely other way"?

TheMarex:
Yep. Could be that the car took a small detour because the road was blocked, but you won't be able to know this. Using that approach for stuff like computing average speed values of roads will just mess up your data.

agruss:
@TheMarex Alright, thanks for the explanation!

PR #1512, opened 2015-06-01
Routing Instructions for match-service

agruss:
This allows the match-service to generate routing instructions.


TheMarex:
@agruss Okay I guess the idea is you want to get the street names of the ways the coordinates where matched to?

In general you have the right idea: its exactly the same as including the turn instructions. But it might be a little confusing from a user perspective to get turn instructions for a trace matching. :wink:

A few idea:
- don't try to integrate it into `json_descriptor.hpp`, instead what about having a look at the code that generates the turn instructions and only use the part that generates the street names? You can put that into`plugins/match.hpp`
- the names should be stored in for each submatching as an array. what about the name `ways`?
- you can use the default `match` step_definition and extent it to check if `row['ways']` is set

Otherwise I'm :+1: for this feature. (actually I wanted to add that myself for some time)

BTW. If you like hit me up at patrick@mapbox.com, I'm curious what you are building.

BharathMG:
Even I was looking to get turn instructions in /match endpoint. :+1:  !

agruss:
@TheMarex I've sent you an email why I implemented the turn instructions for the matching. ;-)
But I think that this may be too specific for my case, so I updated the version (BTW: great hint to get the names in the match.hpp instead from the json_description, thanks again!)

silviokuehn:
Looks good. That's what I'm looking for years. Looking forwards! ;)

PR #1515, opened 2015-06-02
Abstracted OSRM intermediate file writer spec for edges and nodes.

daniel-j-h:
The OSRM intermediate file spec is currently implicitly written down in form of source code, copy and pasted throughout the source base. This pull request starts to abstract the writing process, in order to e.g. later easily be able to change to format to using something like protobuf instead of the hand-crafted byte-wise writing.

The main idea is this: writing can be split into three concepts:

* The file header
* a number of item and
* a finalization step which prefix-length encodes the file by reserving a location for the size and in the end jumping back writing it

### DIFF
--- a/data_structures/hilbert_value.cpp
+++ b/data_structures/hilbert_value.cpp
@@ -57,7 +57,7 @@ uint64_t HilbertCode::BitInterleaving(const uint32_t latitude, const uint32_t lo
 
 void HilbertCode::TransposeCoordinate(uint32_t *X) const
 {
-    uint32_t M = 1 << (32 - 1), P, Q, t;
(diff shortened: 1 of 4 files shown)


daniel-j-h:
Undefined Behavior commit was cherry-picked; I/O abstraction never made it (although I find the policy-based design beautiful). For the record: in case we want real I/O abstractions based on a schema, this is how it can be done with protobuf:

- length-prefix write messages via https://github.com/google/protobuf/pull/710
- zero copy stream backed by a file (via file descriptor)
- chunk up to 64MB of messages into a repeated field and flush after each chunk

Issue #1308, opened 2014-12-09
When the same coordinate is supplied twice via_indices is shorter than via_points


paulbalomiri:
Hi,
When supplying a viaroute request like this (two locations are equal)
```
viaroute?loc=39.8138634,-86.0116111loc=39.8051002,-86.09197929999999&loc=39.8051002,-86.09197929999999
```
the result contains:
```
...
"via_indices":[0,158],
"via_points":[[39.813873,-86.009827],[39.806225,-86.092972],[39.806225,-86.092972]]
...
```
Thus the result is inconsistent within (`via_indices.length!=via_points.length`) and with the input (3 locations).

TheMarex:
This is indeed a bug:

http://router.project-osrm.org/viaroute?loc=39.8138634,-86.0116111loc=39.8051002,-86.09197929999999&loc=39.8051002,-86.09197929999999

Response
```
[...]

PR #1518, opened 2015-06-02
Fix multiple vias with same coordinate

TheMarex:
This fixes adding multiple vias with the same coordinate. It is an edge case, but should not produce invalid data.


TheMarex:
This fixes https://github.com/Project-OSRM/osrm-backend/issues/1308

Issue #1451, opened 2015-04-24
odd route in Monte Video


emiltin:
This is a reduction of an issue reported by on the mailinglist by @woodbri.

This route is normal:
http://osrm.at/bYn
![screen shot 2015-04-24 at 08 04 19](https://cloud.githubusercontent.com/assets/66034/7313353/97eded68-ea58-11e4-87fe-0cbec4276e08.png)

Adding a via point makes osrm take an odd detour:
http://osrm.at/bYm
![screen shot 2015-04-24 at 08 04 38](https://cloud.githubusercontent.com/assets/66034/7313354/9bab8938-ea58-11e4-8a8d-9bddd11832c0.png)

dkastl:
Similar issue (but here the position of the via marker even makes a difference).

* Direct connection
![1](https://cloud.githubusercontent.com/assets/227762/7654944/7922cf02-fb5b-11e4-8f7b-c2d4af5e32ed.png)

* Odd detour as in the previous example
![2](https://cloud.githubusercontent.com/assets/227762/7654959/916c7388-fb5b-11e4-8b7f-47f332c47e30.png)
http://osrm.at/cmT

* Correct result again, when the via point is moved slightly to the right
![3](https://cloud.githubusercontent.com/assets/227762/7654973/a3fcef0a-fb5b-11e4-9b42-e3f368ab61b3.png)

This is very simple residential area with no restrictions, one-ways or complicated road types.
It wasn't difficult to find such an issue. Just moved to an area I know and tried a few short routes with via points.

woodbri:
This looks like it is working correctly on map.project-osrm.org now on develop after revering 11c6713

PR #1522, opened 2015-06-08
illustrating #1451 and #1429

TheMarex:
Can we rebase and merge this as regression test?


emiltin:
it might be better to use a test with table data, rather than a pbf file?

TheMarex:
Should now be covered by ebff45f8037cefb76859e664b43579af92c63e9d.

PR #1529, opened 2015-06-16
Generating matching instructions

agruss:
This is based on the pull request #1512.
This allows to generate via ```match?{locs}&instructions=true``` routing instructions for the matched route.

The instructions are now generated in the ```plugin/match.hpp```. Instead of getting the instructions for every submatching, the instructions are in an array ```route_instructions``` for the whole route, just like the viaroute service provides it. So it is possible to match and then getting routing instructions for this route.

This service is kinda wanted for companies who wants routing instructions for recorded (or generated) routes (this is also a reason why I reopened this PR, because there seems to be an interest in this feature).

### DIFF
--- a/descriptors/description_factory.cpp
+++ b/descriptors/description_factory.cpp
@@ -39,14 +39,16 @@ DescriptionFactory::DescriptionFactory() : entire_length(0) { via_indices.push_b
 
 std::vector<unsigned> const &DescriptionFactory::GetViaIndices() const { return via_indices; }
 
-void DescriptionFactory::SetStartSegment(const PhantomNode &source, const bool traversed_in_reverse)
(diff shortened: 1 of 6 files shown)


arnekaiser:
I think it would be a great feature to get routing instructions for GPX tracks. Many of our users (www.naviki.org) are interested in this feature.

TheMarex:
We can get this into `develop` quickly. About a day's worth of work. Response is going to look like multiple `via_route` responses.

hmaeda:
Getting in the `viaroute` type responses would be amazing! Love the power of open source and especially the community behind osrm! Looking forward to seeing this built and merged into `develop`!

TheMarex:
Tracked in new PR.

PR #1530, opened 2015-06-16
Generate a more direct geometry for match service

agruss:
This is an editing for the geometry of the matching service. You can turn it on via ```match?{locs}&cut=true```.

The actual problem is that some recorded routes may be very inaccurate:
matching_bad_ex_1
matching_bad_ex_2

The green line is the recorded route, the blue line is the actual result of the matching service. The bicycle profile was used for this.
The gps_precision and matching_beta parameter can get rid of some of these segments, but not all of them.

So the actual geometry is not good enough for e.g. navigation purpose, because it leads to dead ends where the actual recorded route didn't go (see the pictures above; the first picture you start at the bottom and should just turn left; the second picture you start also from the bottom and should just go straight and not turn left).
This solution generates a new geometry for the matching service, a geometry which is more direct:
matching_cut_1

It cuts the geometry when the result does these dead ends. Furthermore, dead ends which are quite big (>= 200m) are excluded, since it is very unlikely that these paths are the results of inaccurate coordiantes.

A real recorded route is linked for an example (the track is from Germany, North Rhine-Westphalia):
[Click here](http://www.file-upload.net/download-10694597/weg3.gpx.html)

### DIFF
--- a/data_structures/route_parameters.cpp
+++ b/data_structures/route_parameters.cpp
@@ -131,3 +131,5 @@ void RouteParameters::addCoordinate(
         static_cast<int>(COORDINATE_PRECISION * boost::fusion::at_c<0>(received_coordinates)),
         static_cast<int>(COORDINATE_PRECISION * boost::fusion::at_c<1>(received_coordinates)));
 }
+
+void RouteParameters::setCutFlag(const bool flag) { cut = flag; }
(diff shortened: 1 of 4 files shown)


TheMarex:
Looks very interesting, We decided to solve this client side using https://github.com/mapbox/geojson-tidy

I'm reluctant to add this server side, as trace filtering (especially removing blobs like above) and tuning the `gps_precision` parameter should have the same effect. Also it adds a lot of complexity to the matching. I see why it is a practical solution in your case, but I don't think we want this included.

Thanks for making the effort though. :+1:

Issue #1102, opened 2014-06-25
Importing and interpolating 2D data


emiltin:
It would be great if OSRM would have a generic way to import and interpolate 2D data. 

Data could include elevation, polution, amount of forest/parks, or anything else that's relevant when routing.

OSRM would store the data in an efficient internal format, for example as a number of 3D points (2D location + value). Data could be trimmed to only cover areas where there are road network. Raster data could be sampled, while vector or point data could perhaps be used more directly, or maybe resampled/reduced.

The 2D points could then be interpolated efficiently (using some wave function) at any desired point (in which case you get back a single value) or along any line or polyline (in which case you get a list of distance/value pairs which you can further interpolate if you want). 

Interpolation would be easily accessible from LUA, both when preprocesssing data, and during (the proposed) postprocessing of computed routes.

As an example, the interpolation could be used to compute elevation penalties when preprocessing data; for each way you process in LUA, you pass the way to the  interpolator and get a list of distance/value pairs and use it to compute a "hill slowdown/speedup". You have the flexibility to compute the penalty in LUA how you see fit based on the profle for the way segment.

Later during route queries, once a route has been computed, you can pass the polyline to the interpolator and get back a list of distance/value pairs representing the elevation profile, and pass that along with the route, so the front-end can display it.

You could have several of these data layers, representing different data types. Since the interpolator would be generic and not only for elevation, you could choose to also pass back several profiles to the front-end, for example evalation and polution.

The road network can be viewed as 2.5D. It's mostly flat, but at certain points and areas you have several layers, like bridges or parking houses. A way to handle this would be to have several data layers each datatype, and choose from LUA which one you query for each part of the route.

Since data is represented as 2D points, not much extra space would be needed. For example, consider elevation. Most of the data is at ground level, but bridges are cross on top of that, and at these points you have two elevations for the same point.

You would have several layers: "elevation_layer0" would cover all of the normal road network. "elevation_layer1" would cover only ways tagged with layer=1, and would therefore be tiny (since it only has 2D points where there are bridges). "elevation_layer2" would be even smaller, etc. When you preprocess in LUA you look at the layer tags of the way and interpolate from the relevant data layer. The result would be that when you go under the bridge you get one elevation, and when you go on the bridge, you get another.

TheMarex:
If we add elevation data support to OSRM, it should be in form of something similar. Goals for this are:
- Make it easy to include raster-data as information source for edge weight calculation
[...]

PR #1532, opened 2015-06-17
Support raster data in profiles

lbud:
Ref https://github.com/Project-OSRM/osrm-backend/issues/1102
@TheMarex — here's a prototype that works for e.g. a SF Bay Area OSM extract and a SRTM extract of a similar (but smaller) extent. I've been tweaking a copy of the bicycle profile to run an elevation-based bicycle routing instance and while I have yet to get the heuristics as good as I'd like them, it does work. In a lua profile I added
``` lua
function source_function ()
    load_raster_data("sfbay-srtm.asc", "srtm_id", -122.604861111, -122.323472222, 37.6881944444, 37.8673611113)
end
```
(that's `<source_path>, <user-defined ID string>, extent`)
and something like (this heuristic needs work)
``` lua
function segment_function (edge, distance)
  local elev_delta =
    get_raster_interpolate("srtm_id", edge.source_coordinate.lon, edge.source_coordinate.lat) -
    get_raster_interpolate("srtm_id", edge.target_coordinate.lon, edge.target_coordinate.lat)

  local penalize = elev_delta / distance

  if penalize > 0 then
     edge.speed = edge.speed * math.max((1 - penalize * 2), 0.1)
  end
end
```

I've got several questions/issues about how this should be designed:

_Default profiles_
* Should we not integrate this into any default profiles, since they're dependent on local files?

_File types_ — probably my biggest question
* The file type I've been using is basically an Arc ASCII file with the first…6?…lines of metadata cut off, so it looks like [this](https://raw.githubusercontent.com/Project-OSRM/osrm-backend/a3961af4afd13a8574d51fcedfc8811a969cc3da/unit_tests/fixtures/raster_data.asc) — just a plan ASCII grid. What should we standardize on?
* Do you have any ideas as to how this might work for…bigger datasets? :earth_africa: 

_Performance_
* Right now loaded sources are stored [in main memory](https://github.com/Project-OSRM/osrm-backend/blob/a3961af4afd13a8574d51fcedfc8811a969cc3da/data_structures/raster_source.hpp#L127-L128). I'm not sure this is an idea that scales…
* I haven't benchmarked this processing vs the whole-world-processing but should

_Handling coordinates outside of a raster's bounds_
* I quickly realized that [throwing exceptions](https://github.com/Project-OSRM/osrm-backend/blob/a3961af4afd13a8574d51fcedfc8811a969cc3da/data_structures/raster_source.hpp#L73) was a bad idea here as you may frequently have a [semi-]localized datasource. But returning -1, or any value, is also not a good idea since this is so source-agnostic. Do you have ideas on how to handle this more gracefully?

_Code review_
* I'm sure I've got all my number types wrong…these could definitely use a good scrutiny
* Lua: part of me says "there's no way [overloading no-ops with lua functions](https://github.com/Project-OSRM/osrm-backend/blob/a73b3a6585d09e5ad70a3f3506d0c2e4f831b199/extractor/scripting_environment.cpp#L72-L73) and [redefining number types](https://github.com/Project-OSRM/osrm-backend/blob/a73b3a6585d09e5ad70a3f3506d0c2e4f831b199/extractor/scripting_environment.cpp#L145-L167) is the right way to do this," but that same part of me tried a million other ways. Does this look rightish/wrongish?
* [This unit test error](https://github.com/Project-OSRM/osrm-backend/blob/a3961af4afd13a8574d51fcedfc8811a969cc3da/unit_tests/data_structures/raster_source.cpp#L55-L56) seems to be a Boost issue, but makes me nervous for zero values…need to do further digging on this

I've got all of the raster_source datastructure code in a header file right now; I assume I probably should move most of it out of the header but it seemed faster to develop this way — once this is near ready I'll do that and then squash all these commits


TheMarex:
Awesome. :+1: Reviewing this in detail will take some time. There are a number of smaller issues I spotted while skimming the diff, but we can tackle that later. :sparkles:  Some things right off the bat:

What are the implications of doing something like:

> local raster_id = load_raster_data("sfbay-srtm.asc", -122.604861111, -122.323472222, 37.6881944444, 37.8673611113)

in the global of a lua profile? I guess it gets executed for each separate lua context. But we check if the source is already loaded. As long as the same `raster_id` is returned I think that should work, right? This would have the benefit of avoiding calling another function directly from outside of lua.

I don't see a need for a user defined ID. The filename already uniquely identifies the resource from a user perspective. For speaking to the API an internal (integer) handle suffices. Like this:

> get_raster_interpolate(raster_id, edge.source_coordinate.lon, edge.source_coordinate.lat)

Passing strings around is really slow (relative to other operations in C++), so people try to avoid it at all cost. :wink:

> Should we not integrate this into any default profiles, since they're dependent on local files?

I think we should include a demo profile (or extend the bicycle profile `local use_elevation=false`) but make it explicit that it is "bring your own data".

> Right now loaded sources are stored in main memory. I'm not sure this is an idea that scales…

My current idea would be to simply mmap the input file if we can get random access. But this is stuff for later.

> The file type I've been using is basically an Arc ASCII file with the first…6?…lines of metadata cut off, so it looks like this — just a plan ASCII grid. What should we standardize on?

For now don't worry about scaling this data-wise. In the long run we want to use something binary and not ASCII based. Most convenient would be have GeoTIFF support, but I guess the format is too complex and not suitable for random access (compression). There are lots of clever and complicated ways we can make this scale if we have the basic infrastructure set (most of these involve a separate pre-processing step for raster data into our own internal format).

After re-reading this, it sounds a lot like "don't worry", but we need to spec out each of these issues in separate steps and iterate on a basic implementation. :rocket:

Also could you rebase on the recent `develop` branch?

lbud:
> What are the implications of doing something like:
> 
> > local raster_id = load_raster_data("sfbay-srtm.asc", -122.604861111, -122.323472222, 37.6881944444, 37.8673611113)
> 
> in the global of a lua profile? I guess it gets executed for each separate lua context. But we check if the source is already loaded. As long as the same raster_id is returned I think that should work, right? This would have the benefit of avoiding calling another function directly from outside of lua.
> 
> I don't see a need for a user defined ID. The filename already uniquely identifies the resource from a user perspective. For speaking to the API an internal (integer) handle suffices. Like this:
> 
> > get_raster_interpolate(raster_id, edge.source_coordinate.lon, edge.source_coordinate.lat)
> 
> Passing strings around is really slow (relative to other operations in C++), so people try to avoid it at all cost. :wink:

Right, I started with 
``` lua
local raster_id = load_raster_data("sfbay-srtm.asc", -122.604861111, -122.323472222, 37.6881944444, 37.8673611113)
```
but `raster_id` isn't present in the other contexts by the time you need it. But you're right that the user-defined IDs are weird — what we could do is
``` lua
-- global pathname:
srtm_source = "sfbay-srtm.asc"

function source_function ()
  load_raster_data(srtm_source, -122.604861111, -122.323472222, 37.6881944444, 37.8673611113)
end
```
``` lua
function segment_function (edge, distance)
  get_raster_interpolate(srtm_source, edge.source_coordinate.lon, edge.source_coordinate.lat)
  ...
```

Will make that change and rebase on `develop` now :+1:

TheMarex:
> but raster_id isn't present in the other contexts by the time you need it.

Can you elaborate on that? You mean it is causing a race-condition, because multiple threads are trying to register a raster source simultaneously? It is absolutely okay to add locking here. I would suggest something like http://en.cppreference.com/w/cpp/thread/lock_guard

We want to really avoid passing a string to `get_raster_interpolate` because we will do that operation million of times. Looking a string up in a hash-table is not for free (string needs to be copied from lua to C++, hashed (e.g. find the bucket index) and the a linear scan of the bucket), looking something up with an integer index to a vector is for free (simple addition of memory addresses).

lbud:
Based on changes introduced in the rebased commits that refactored extraction into more modular functions — in particular, changes to the structure of InternalExtractorEdge — my example `segment_function` is a little funny now:

``` lua
function segment_function (source, target, distance)
  -- `source` has properties .source_coordinate.{lat, lon} and 
  --      .speed (corresponds to InternalExtractorEdge)
  -- `target` has properties .lat, .lon (corresponds to ExtractionNode)
  local elev_delta =
    (get_raster_interpolate(raster_source, source.source_coordinate.lon, source.source_coordinate.lat) -
    get_raster_interpolate(raster_source, target.lon, target.lat))

  local penalize
  if distance * 1 ~= 0 then
      penalize = elev_delta / distance
  else
      penalize = 0
  end

  if penalize ~= 0 then
     source.weight_data.speed = source.weight_data.speed * math.max((1 - penalize * 10), 0.1)
  end
end
```

TheMarex:
@lbud the interface looks much better. :+1: We should swap the order `weight` and `distance` though, since the "return value" is always the last parameter by C++ convention.

Not sure if we should pass `WeightData` directly to lua. Having a way of eigther setting speed or duration is great, but I fear people will
forget to set the `.type` property accordingly. Also, how do enums map to lua?

Otherwise I think you nailed it on functionality. Some smaller issues like the `double` problem remain. That _should_ work out of the box. We need to figure out why not.

After that we can do a final code review to polish this up and merge this version. :rocket:

data_structures/raster_source.cpp
```diff
@@ -0,0 +1,176 @@
...
+
+#include <osrm/coordinate.hpp>
+
+#include <unordered_map>
+#include <sstream>
+#include <cmath>
+#include <iostream>
+
+// LoadedSources stores sources in memory; LoadedSourcePaths maps them to indices
+std::vector<RasterSource> LoadedSources;
```
TheMarex: This state should be wrapped in a class e.g. `SourceContainer`, which has the free functions below as members.

extractor/scripting_environment.cpp
```diff
@@ -73,13 +76,37 @@ void ScriptingEnvironment::init_lua_state(lua_State *lua_state)
     // open utility libraries string library;
     luaL_openlibs(lua_state);
 
+    const auto sourceNoOp = []
+    {
+    };
+    const auto segmentNoOp = [](const FixedPointCoordinate *, const ExternalMemoryNode *,
```
TheMarex: I don't think this is needed. You can use `lua_function_exists(lua_state, "segment_function");`

extractor/scripting_environment.cpp
```diff
@@ -122,9 +149,39 @@ void ScriptingEnvironment::init_lua_state(lua_State *lua_state)
...
+        luabind::class_<ExternalMemoryNode>("EdgeTarget")
+            .property("lat", &ExternalMemoryNode::lat)
+            .property("lon", &ExternalMemoryNode::lon),
+        luabind::class_<FixedPointCoordinate>("Coordinate")
+            .property("lat", &FixedPointCoordinate::lat)
+            .property("lon", &FixedPointCoordinate::lon),
+        luabind::class_<RasterDatum>("RasterDatum")
+            .property("datum", &RasterDatum::datum)
+            .def("invalid_data", &RasterDatum::get_invalid),
+        luabind::class_<double>("double")
```
TheMarex: What is the error message you get then you remove that? We should aim to get double to work out-of-the-box with luabind.

lbud:
@TheMarex re your comments before:
* I was able to remove the `double` bindings when by changing it from a `boost::cref()` wrapper to passing the raw double to Lua
* Didn't see `lua_function_exists` before, thanks!
* There is now a single instance of `SourceContainer` bound to a lua state if `source_function` exists (the same lua state is used to call `segment_function`. The sources API from lua scripts has thus changed slightly: 

_loading raster sources:_ 
```lua
raster_source = sources:load(RASTER_SOURCE_PATH, LON_MIN, LON_MAX, LAT_MIN, LAT_MAX, NROWS, NCOLS)
```
(`sources` is the instance bound to this lua state's globals table — this is bound only _if_ `source_function` exists, so it but does not have to be used in a `segment_function` (i.e. if for some reason you wanted to do something with segments without loading a raster source, you could still include a `segment_function`, but could not use the following two APIs as there would be no `sources` `SourceContainer` instance))
_nearest neighbor query:_
``` lua
local sourceData = sources:query(raster_source, source.lon, source.lat)
```
_bilinear interoplation query:_
``` lua
local targetData = sources:interpolate(raster_source, target.lon, target.lat)
```

lbud:
I rebased on `develop` before I pushed this most recent round of commits so I'm thinking this is…readyish, once I squash and also figure out the following:
* Not sure how to fix appveyor (is this something that is only fixed on `develop`?)
* Should I look into supporting more raster formats first (dig into libtiff)? Or should that be a post-merge as-requested thing?
* Do you think I should look into stubbing out some scenarios for a few cucumber tests?

TheMarex:
> Not sure how to fix appveyor (is this something that is only fixed on develop?)

This should be fixed in the PR. Looking at the error message it seems like you hard-coded a path in UNIX fashion somewhere? You can use `boost::filesystem::path` to translate between UNIX and Windows paths.

> Should I look into supporting more raster formats first (dig into libtiff)? Or should that be a post-merge as-requested thing?

Let's focus on this post-merge. Goal is to get the minimum viable feature merge end of this month.

> Do you think I should look into stubbing out some scenarios for a few cucumber tests?

Yes, absolutely. Cucumber tests are pretty vital at making sure we don't break any work-flows and profile features.

lbud:
Cucumber tests have been added. They're a bit weird as this feature is both an extract step and affects routing, so there's a good extract test, a bad extract test, and a routing test. Right now cucumber tests are failing as I used speeds that depend on https://github.com/Project-OSRM/osrm-backend/pull/1625 merging, and appveyor tests are failing because it looks like some appveyor things moved on S3 in https://github.com/Project-OSRM/osrm-backend/commit/b025179a8c31a365b803c6a800ec307c2d80105b on PR https://github.com/Project-OSRM/osrm-backend/pull/1612. Once those merge I'll rebase, squash and then this is good to go :rocket:

TheMarex:
@lbud should be good for a rebase now.

data_structures/raster_source.hpp
```diff
@@ -0,0 +1,185 @@
...
+    RasterGrid(const RasterGrid &) = default;
+    RasterGrid &operator=(const RasterGrid &) = default;
+
+    RasterGrid(RasterGrid &&) = default;
+    RasterGrid &operator=(RasterGrid &&) = default;
+
+    std::int32_t operator()(std::size_t x, std::size_t y) { return _data[y * xdim + x]; }
+    const std::int32_t operator()(std::size_t x, std::size_t y) const
+    {
+        return _data[(y-1) * xdim + (x-1)];
```
daniel-j-h: `x` and `y` are of type `std::size_t` here, that is a unsigned-ish typedef. If you do `x-1` and `x` is `0`, it wraps around to `std::numeric_limits<std::size_t>::max()`. The same goes for `y`. Accessing `_data` is then out of bounds, thus undefined behavior.

Also, the `const` `operator()` should not be different than the non-`const` `operator()` one line above.

@lbud: is it because lua does 1-based indexing and your segment function somehow passes the values through to this function?

data_structures/raster_source.cpp
```diff
@@ -0,0 +1,183 @@
...
+{
+    BOOST_ASSERT(count > 0);
+    return (max - min) / count;
+}
+
+// Query raster source for nearest data point
+RasterDatum RasterSource::getRasterData(const float lon, const float lat) const
+{
+    int _lon = static_cast<int>(lon * COORDINATE_PRECISION);
+    int _lat = static_cast<int>(lat * COORDINATE_PRECISION);
```
daniel-j-h: These two can be `const`. And `auto`. Just make them `const auto` :)

data_structures/raster_source.cpp
```diff
@@ -0,0 +1,183 @@
...
+
+RasterSource::RasterSource(RasterGrid _raster_data,
+                           std::size_t _width,
+                           std::size_t _height,
+                           int _xmin,
+                           int _xmax,
+                           int _ymin,
+                           int _ymax)
+    : xstep(calcSize(_xmin, _xmax, _width)), ystep(calcSize(_ymin, _ymax, _height)),
+      raster_data(_raster_data), width(_width), height(_height), xmin(_xmin), xmax(_xmax),
```
daniel-j-h: As `RasterGrid` is moveable, you could do `raster_data(std::move(_raster_data))` to save a expensive copy operation here.

data_structures/raster_source.cpp
```diff
@@ -0,0 +1,183 @@
...
+    int yth = ((ythP - floor(ythP)) > (ystep / 2.) ? floor(ythP) : ceil(ythP));
+
+    return {raster_data(xth, yth)};
+}
+
+// Query raster source using bilinear interpolation
+RasterDatum RasterSource::getRasterInterpolate(const float lon, const float lat) const
+{
+    int _lon = static_cast<int>(lon * COORDINATE_PRECISION);
+    int _lat = static_cast<int>(lat * COORDINATE_PRECISION);
```
daniel-j-h: `const auto`

data_structures/raster_source.cpp
```diff
@@ -0,0 +1,183 @@
...
+    int right = ceil(xthP);
+
+    float x = (_lon - left * xstep + xmin) / xstep;
+    float y = (ymax - top * ystep - _lat) / ystep;
+    float x1 = 1.0 - x;
+    float y1 = 1.0 - y;
+
+    return {static_cast<std::int32_t>(
+        raster_data(left, top) * (x1 * y1) + raster_data(right, top) * (x * y1) +
+        raster_data(left, bottom) * (x1 * y) + raster_data(right, bottom) * (x * y))};
```
daniel-j-h: This looks like a bilinear interpolation and probably is. Would be great to derive it from the code at first sight.
Maybe introduce some intermediary variables and give them appropriate names.

data_structures/raster_source.cpp
```diff
@@ -0,0 +1,183 @@
...
+                                      double xmax,
+                                      double ymin,
+                                      double ymax,
+                                      unsigned nrows,
+                                      unsigned ncols)
+{
+    int _xmin = static_cast<int>(xmin * COORDINATE_PRECISION);
+    int _xmax = static_cast<int>(xmax * COORDINATE_PRECISION);
+    int _ymin = static_cast<int>(ymin * COORDINATE_PRECISION);
+    int _ymax = static_cast<int>(ymax * COORDINATE_PRECISION);
```
daniel-j-h: `const auto`

data_structures/raster_source.cpp
```diff
@@ -0,0 +1,183 @@
...
+
+    auto itr = LoadedSourcePaths.find(source_path);
```
daniel-j-h: `const` as you do not modify the iterator. Maybe even a `const_iterator` as you also do not modify the element it is pointing to.

data_structures/raster_source.cpp
```diff
@@ -0,0 +1,183 @@
...
+    if (itr != LoadedSourcePaths.end())
+    {
+        std::cout << "[source loader] Already loaded source '" << source_path << "' at source_id "
```
daniel-j-h: Do we have logging infrastructure in place? Or does everything go to `std::cout`? (Or better `std::cerr`?)

data_structures/raster_source.cpp
```diff
@@ -0,0 +1,183 @@
...
+    TIMER_START(loading_source);
+
+    if (!boost::filesystem::exists(source_path.c_str()))
+    {
+        throw osrm::exception("error reading: no such path");
+    }
+
+    RasterGrid rasterData{source_path, ncols, nrows};
+
+    RasterSource source{rasterData,
```
daniel-j-h: You do not need the `rasterData` variable from here on, as you pass it to the `source` variable's constructor (sink). Therefore, instead of 1/ a copy operation into the `source` constructor and then 2/ the `rasterData` goes out of scope and gets destructed anyway just do a single move:

    RasterSource source{std::move(rasterData), ..

data_structures/raster_source.cpp
```diff
@@ -0,0 +1,183 @@
...
+                        static_cast<std::size_t>(ncols),
+                        static_cast<std::size_t>(nrows),
```
daniel-j-h: If you have to cast them to `std::size_t` here anyway, would it make sense to change the arguments from `unsigned` to `std::size_t`?

data_structures/raster_source.cpp
```diff
@@ -0,0 +1,183 @@
...
+RasterDatum SourceContainer::getRasterDataFromSource(unsigned int source_id, int lon, int lat)
+{
+    if (LoadedSources.size() < source_id + 1)
+    {
+        throw osrm::exception("error reading: no such loaded source");
+    }
+
+    const auto &found = LoadedSources[source_id];
+    return found.getRasterData(float(lon) / COORDINATE_PRECISION,
+                               float(lat) / COORDINATE_PRECISION);
```
daniel-j-h: `static_cast<T>(x)` instead of old-style `T(x)` C-casts. The reason for this is that C-style casts are implicitely trying "safe casts" (what `static_cast` is doing) and if there is no way to do this, they silently fall back to casts such as `reinterpret_cast` (interpreting bit patterns) and `dynamic_cast` (up/down casting in a type hierarchy), that you really do not want normally.

In addition, C-casts are hard to grep for.


Note: I know OSRM uses C-style casts in a lot of places --- does not make it better though :)

data_structures/raster_source.cpp
```diff
@@ -0,0 +1,183 @@
...
+SourceContainer::getRasterInterpolateFromSource(unsigned int source_id, int lon, int lat)
+{
+    if (LoadedSources.size() < source_id + 1)
+    {
+        throw osrm::exception("error reading: no such loaded source");
+    }
+
+    const auto &found = LoadedSources[source_id];
+    return found.getRasterInterpolate(float(lon) / COORDINATE_PRECISION,
+                                      float(lat) / COORDINATE_PRECISION);
```
daniel-j-h: `static_cast<T>(x)`. Also shares code with `SourceContainer::getRasterDataFromSource`.

data_structures/raster_source.hpp
```diff
@@ -0,0 +1,185 @@
...
+class RasterGrid
+{
+  public:
+    RasterGrid(const std::string &filepath, std::size_t _xdim, std::size_t _ydim)
+    {
+        xdim = _xdim;
+        ydim = _ydim;
+        _data.reserve(ydim * xdim);
+
+        boost::filesystem::ifstream stream(boost::filesystem::path{filepath});
```
daniel-j-h: Hmm here we're checking for `exists` before reading the file:

https://github.com/Project-OSRM/osrm-backend/pull/1532/files#diff-ec33cc38b2649821f0afd9f8c800b198R135

I know the stdlib's `std::ifstream` crashes when initialized with a path to a directory instead of a file, like `"."` --- not sure if boost's ifstream behaves differently. Could you check this and more importantly use a single way of doing this throughout the code?

data_structures/raster_source.hpp
```diff
@@ -0,0 +1,185 @@
...
+                         double xmin,
+                         double xmax,
+                         double ymin,
+                         double ymax,
+                         unsigned nrows,
+                         unsigned ncols);
+
+    RasterDatum getRasterDataFromSource(unsigned int source_id, int lon, int lat);
+
+    RasterDatum getRasterInterpolateFromSource(unsigned int source_id, int lon, int lat);
```
daniel-j-h: `lon` and `lat` can be easily mixed here --- could you at least add some asserts to the definitions, validating for valid input?

extractor/extraction_containers.cpp
```diff
@@ -264,6 +269,16 @@ void ExtractionContainers::PrepareEdges()
...
             node_iterator->lat, node_iterator->lon);
 
+        if (lua_function_exists(segment_state, "segment_function"))
+        {
+            luabind::call_function<void>(
+                segment_state, "segment_function",
+                boost::cref((*edge_iterator).source_coordinate),
+                boost::cref(*node_iterator),
+                distance,
+                boost::ref((*edge_iterator).weight_data));
```
daniel-j-h: `(*it).x` is the same as `it->x` :)

unit_tests/data_structures/raster_source.cpp
```diff
@@ -0,0 +1,87 @@
...
+{
+    SourceContainer sources;
+    int source_id =
+        sources.loadRasterSource("../unit_tests/fixtures/raster_data.asc", 0, 0.09, 0, 0.09, 10, 10);
+    BOOST_CHECK_EQUAL(source_id, 0);
+
+    BOOST_CHECK_EQUAL(sources.getRasterDataFromSource(0, normalize(0.02), normalize(0.02)).datum, 0);
+    // BOOST_CHECK_EQUAL(sources.getRasterDataFromSource(0, 0.0, 0.00), 0);
+    // TODO: 'unknown location:0: fatal error in "raster_test": signal: SIGSEGV, si_code: 0 (memory
+    // access violation at address: 0x00000000)' - seems to be boost problem w 0
```
daniel-j-h: Maybe http://stackoverflow.com/a/16131381 --- but the `0x00000000` segfault looks more like a null pointer dereferencing issue.

unit_tests/data_structures/raster_source.cpp
```diff
@@ -0,0 +1,87 @@
...
+#include "../../typedefs.h"
+#include "../../util/osrm_exception.hpp"
+
+#include <osrm/coordinate.hpp>
+
+#include <boost/filesystem.hpp>
+
+#include <boost/test/unit_test.hpp>
+#include <boost/test/test_case_template.hpp>
+#include <boost/mpl/list.hpp>
```
daniel-j-h: Do you really need those includes? I can't find anything `mpl` related for example.

unit_tests/fixtures/raster_data.asc
```diff
@@ -0,0 +1,10 @@
+0  0  0  0  0  2  4  6  8  10
+0  0  0  0  2  4  6  8  10 12
+0  0  0  2  4  6  8  10 12 14
+0  0  2  4  6  8  10 12 14 16
+0  2  4  6  8  10 12 14 16 18
+0  0  2  4  6  8  10 12 14 16
+0  0  0  2  4  6  8  10 12 14
+0  0  0  0  2  4  6  8  10 8
+0  0  0  0  0  2  4  6  8  6
+0  0  0  0  0  0  2  4  6  4
```
daniel-j-h: I would change the edge values to something different than `0`. This way when your grid subscript `operator()()` does an out of bounds access and does not get killed but accesses memory to the left/right instead which may be zeroed out, you can detect this more easily.

lbud:
@emiltin thanks so much for all your notes! Really helpful; I had not used cucumber before. I was dynamically writing a rasterbot profile so as not to clutter the `profiles/` directory with more profiles to maintain, but actually I think you're right that it's much less confusing to just use a static profile in that directory — fixed in https://github.com/Project-OSRM/osrm-backend/commit/0ae1e3d9670b740c4d8ac633f98d0c633aeff8bd.

emiltin:
you're welcome, glad you could use the comments.

emiltin:
maybe we can move test profiles to a folder like profiles/test/

data_structures/raster_source.hpp
lbud: @daniel-j-h yeah, definitely. I switched this in haste last week (https://github.com/Project-OSRM/osrm-backend/commit/dd796a3d6b970d1ebcca3626920ea897f68bfe7a) trying to get AppVeyor to pass before I realized that wasn't necessarily the cause. I'll try to iron this out throughout these files.

data_structures/raster_source.hpp
lbud: The best we could do to assert this, I think, is just checking that -90 < lat < 90…right? :\

daniel-j-h: Yeah, that's probably the best we can do for now.

In an optimal world we would have a wrapper type around lon, lat that provides only sensible member functions. After all, with the current design what does it even mean to do `lon / lat` or `lon ^ lat`, that's just nonsense but the compiler allows it since it's fine for their int types.

By the way that's the reason there is a path type in the boost filesystem library, instead of just passing file paths wrapped in strings around, since there are operations on strings that just do not make sense on paths (and the other way round).

data_structures/raster_source.hpp
lbud: @daniel-j-h thanks to this comment a found some more math problems :) fixed in latest commit

profiles/rasterbot.lua
```diff
@@ -0,0 +1,42 @@
...
+    result.name = name
+  end
+
+  result.forward_speed = 15
+  result.backward_speed = 15
+end
+
+function source_function ()
+  raster_source = sources:load(
+    "../test/fixtures/rastersource.asc",
```
emiltin: wouldn't it be better if this path was somehow passed in as a program option when running osrm-prepare? eg --raster elevation.asc

data_structures/raster_source.hpp
lbud: @daniel-j-h — just spent a little more time digging around. Since Lua's types are fairly primitive, the API that in the bindings (`SourceContainer::loadRasterSource`) needs to keep a `std::string` as its function signature. Moreover, the `std::unordered_map LoadedSourcePaths` template can't take a `boost::filesystem::path`, so paths need to be a regular `std::string` for that as well.

So: no, it's impossible to use _only_ `boost::filesystem::path`s throughout this data structure — at some point it needs to construct a `boost::filesystem::path` from an original `std::string`. I changed this slightly in https://github.com/Project-OSRM/osrm-backend/commit/96ff3f17dabaed1ad71ca4fa7dd696fae9b498af so that it constructs a `boost` path only once and a `RasterGrid` constructor takes a boost path, but that's kind of the best we can do here, I think :\

(Also, I'd like to keep the `exists` check and the `stream` check both — an `exists` error implies a user-created problem where the path was specified incorrectly in a profile and therefore belongs in the loading API, whereas a `stream` error speaks maybe to a more system-level error if a path does indeed exist, I think. Just my intuition.)

profiles/rasterbot.lua
lbud: @emiltin possibly, but to do that this feature would become limited to one data source. When everything is done in the profiles like this you can include multiple sources, each of which are assigned their own int ID (in this case `raster_source` is `0`) and can be queried separately (so e.g. you could layer elevation data _and_ pollution data, or you could geographically patchwork multiple sources together, etc).

emiltin: but still, it means hardcoding paths in code. each time you move a file you then need to change your profile code. i think moving paths to something like a settings file would be nice. eg:

# raster_source.ini
elevation /data/raster_4w3452.asc
pollution /data/raster_84hhw.asc
layer2 /data/raster_cnn322.asc

TheMarex: I don't see a problem with paths in the profile. They are ment to be modified by the user. Configuration files have a terrible usability - let's not start adding them again.

data_structures/raster_source.hpp
daniel-j-h: > Also, I'd like to keep the exists check and the stream check both — an exists error implies a user-created problem where the path was specified incorrectly in a profile and therefore belongs in the loading API, whereas a stream error speaks maybe to a more system-level error if a path does indeed exist, I think. Just my intuition.

Yes, sounds reasonable. Then grep your code and make sure to check for `exists()` everywhere you create a stream from a filepath. :+1:

TheMarex:
@lbud seems like all tests are passing. :rocket: Good to merge after a round of commit squashing?

lbud:
@TheMarex yes! I wanted to have a solid draft of the documentation for this ready to add to the wiki: https://gist.github.com/lbud/05dcaf02685db0db80e8 . Just squashed into https://github.com/Project-OSRM/osrm-backend/commit/bac6703f8e0d089aa58b7fefee8bba7204e4e492 … :rocket:

TheMarex:
@lbud great documentation. Yes duration values are typically only used for ferries or moving bridges and stuff like that.

Merging into develop. :rocket:

PR #1565, opened 2015-07-09
Compute components on edge expanded graph

TheMarex:


TheMarex:
The main problem at the moment with this approach is that sometimes request A->B will not work but B->A because of oneway streets. If you run an SCC algorithm on that it will identify A and B to be in different strictly connected components (by definition SCC means that there is always a path from A->B and B->A in each component). My idea is to first run a connected component analysis (two points are in the same connected component if there exists either a path A->B or B->A) and within each connected component a SCC search. If two edges are in the same CC and the corresponding SCC id of A is smaller than the corresponding SCC id of B there must be a path A->B but not B->A.

TheMarex:
Current problem with this PR is that the search would never snap to oneways leading into a parking lot - even if there is a path between both endpoints.

I think what I want to build here might actually be impossible (in constant space). Reading a little bit in the literature, what I basically want is a Reachability Oracle: `(s, t) \in O <=> There is a directed path from s to t`.

This can be solved trivially if you have a lookup table that tells you the distance between every pair of nodes, which would need `O(n^2)` space. Hub Labels promise near-linear lookup time, but also have high space requirements. (could be easily constructed from the CH though)

One idea would be to try to figure out the reachability only between SCC's of the same CC - which would involve running a search between each pair - probably extremely slow.

TheMarex:
Gave this a little more though:

- The only case where this change will cause problems is routing on oneway strees that lead into a CC
- In most cases there will be a way to and from an island, so snapping to brides will still work
- I would like to give this a spin as it should only improve the current situation

Issue #1575, opened 2015-07-22
Total route duration !== sum of all maneuver durations


peterqliu:
http://osrm.at/dLQ

This trip has a total duration of 95s, but its first maneuver takes 169s.

cc/ @lbud

TheMarex:
Best guess would a bug in the edge compression. Note that there should be a turn announced.

TheMarex:
Okay found the root cause of this:

Path unpacking is broken for the first and last segment if these segments are compressed edges. In that case `PathData` will contain the full length of the compressed edge - not only the length of the part that is covered by the `PhantomNode`.

This has some bigger implications on the query code and description generation code. Currently working on fixing this and simplifying the code along the way.

danpat:
This has been fixed.

Current API request:

https://router.project-osrm.org/route/v1/driving/-122.98851,37.995975;-122.998009,37.99736?overview=false&alternatives=true&steps=true

sum of steps durations == summary duration.

PR #1586, opened 2015-08-05
[not ready] Fix Point Reyer

TheMarex:
This PR aims to address #1575.

- [x] Change `InternalDataStructre.unpacked_route_segments` to be a continuous array
- [ ] Change `UnpackPath()` to only generate a list of edge-based-nodes (currently it also does decompression + handles phantom nodes)
- [ ] Create  `UncompressPath()` that takes a list of edge-based nodes and de-compresses the corresponding edges into `PathData` (needs access to phantom nodes)


TheMarex:
Outdated.

PR #1594, opened 2015-08-11
gps_precision & matching_beta as float

agruss:
In the server api the value for both parameters is described as a float.
In the current version both parameters can only get short integer values which seems to be a mistake (I guess?).
This PR fixes the values for both parameters.


TheMarex:
Good catch! :+1:

PR #1603, opened 2015-08-18
Modernize the code base to C++11 standards and beyond.

daniel-j-h:
Apply `clang-modernize` (based on Clang 3.6) transformations to the
codebase while making sure to support Clang>=3.4 and GCC>=4.8.

We apply the transformations in parallel to speed up the quite
time consuming process, and use our `clang-format` style file
to automatically format the code respecting our coding conventions.

We use the following self-explanatory transformations:

* AddOverride
* LoopConvert
* PassByValue
* ReplaceAutoPtr
* UseAuto
* UseNullptr

This required a `compile_commands.json` compilation database, e.g.

    ccmake .. -DCMAKE_BUILD_TYPE=Release -DCMAKE_EXPORT_COMPILE_COMMANDS=1

for CMake or check Bear for a Makefile based solution (or even Ninja).

    git ls-files -x '*.cpp|*.h' | \
      xargs -I{} -P $(nproc) clang-modernize -p build -final-syntax-check -format -style=file -summary -for-compilers=clang-3.4,gcc-4.8 -include . -exclude third_party {}

Boom!

References:

* http://clang.llvm.org/extra/clang-modernize.html
* http://clang.llvm.org/extra/ModernizerUsage.html

### DIFF
--- a/data_structures/segment_information.hpp
+++ b/data_structures/segment_information.hpp
@@ -34,6 +34,7 @@ SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 #include "../typedefs.h"
 
 #include <osrm/coordinate.hpp>
+#include <utility>
 
 // Struct fits everything in one cache line
 struct SegmentInformation
@@ -48,7 +49,7 @@ struct SegmentInformation
     bool necessary;
     bool is_via_location;
 
-    explicit SegmentInformation(const FixedPointCoordinate &location,
+    explicit SegmentInformation(FixedPointCoordinate location,
                                 const NodeID name_id,
                                 const EdgeWeight duration,
                                 const float length,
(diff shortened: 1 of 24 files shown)


data_structures/segment_information.hpp
```diff
@@ -56,20 +57,20 @@ struct SegmentInformation
...
-    explicit SegmentInformation(const FixedPointCoordinate &location,
+    explicit SegmentInformation(FixedPointCoordinate location,
                                 const NodeID name_id,
                                 const EdgeWeight duration,
                                 const float length,
                                 const TurnInstruction turn_instruction,
                                 const TravelMode travel_mode)
-        : location(location), name_id(name_id), duration(duration), length(length), bearing(0),
-          turn_instruction(turn_instruction), travel_mode(travel_mode),
+        : location(std::move(location)), name_id(name_id), duration(duration), length(length),
```
TheMarex: Will this cause copy (for the parameter `location`) and one move parameter -> attribute (well since this are two `int` values it will be like a copy anyway)?

TheMarex:
Do you have more information on the `const&` -> pass-by-value + `std::move` pattern? Can't this incur an additional copy?

extractor/internal_extractor_edge.hpp
```diff
@@ -68,18 +69,26 @@ struct InternalExtractorEdge
...
+                 target,
+                 name_id,
+                 0,
+                 forward,
+                 backward,
+                 roundabout,
+                 access_restricted,
+                 travel_mode,
+                 is_split),
+          weight_data(std::move(weight_data))
```
TheMarex: This might also be a problem.

data_structures/segment_information.hpp
daniel-j-h: `location` is constructed in the `SegmentInformation` constructor from either an lvalue or an rvalue. If it's a rvalue it gets moved in (which probably is implemented as a copy for primitive types, no guarantees). The member attribute is then initialized by invoking its move constructor (which probably is implemented as a copy for primitives types, again).

Detailed explanation will follow in the PR.

daniel-j-h:
The arguments for passing sink values by value is explained in detail in "Modern Effective C++", Item 26 and Item 27. Here's a article that basically has more or less the same explanations:

> https://dzone.com/articles/c-sink-parameter-passing

The reasoning goes like this:

You want to provide the option to "steal" from an rvalue if the user passes one in. In the lvalue case you need a copy anyway. That is, the optimal solution would be to write both `T(const U&)` and `T(U&&)`.

This is fine for a single parameter, for more there is a space explosion, as you would have to write all combinations, such as `T(const U&, const V&)`,`T(U&&, const V&)` and so on.

You could use what Scott Meyer's calls "universal references", that is `T&&` in a type-deducing context, but this has downsides of 1/ requiring a type-deducing context, 2/ being favoured by the compiler over copy constructors for non-const lvalues, 3/ ugly error messages.

In contrast passing by value solves the problem in the following way:

If it's an rvalue we move from it, if it's an lvalue we do a copy (that we need anyway). This requires the move to be rather cheap. The standard provides a few features that make it cheap even in cases where the type is not moveable:

* [Copy Elision](
http://en.cppreference.com/w/cpp/language/copy_elision): the copy / move from a temporary into the member attribute is **ommitted** --- if you don't want this you have to disable this with a compiler switch (smth. smth. fno-elision)
* (Named)RVO: even without move semantics NRVO kicks in, constructing objects in place

In our particular situation (i.e. primitive types) there should be no difference.

TheMarex:
Alright, performance impact would probably be minimal anyway.

PR #1610, opened 2015-08-19
Add roundtrip/TSP plugin

chaupow:
This PR is not ready yet. But soon. Stay tuned.


zavan:
Nice! :+1:

library/osrm_impl.hpp
```diff
@@ -28,6 +28,10 @@ SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 #ifndef OSRM_IMPL_HPP
 #define OSRM_IMPL_HPP
 
+// #if __cplusplus > 199711L
```
TheMarex: This needs to be removed.

unit_tests/routing_algorithms/tsp.cpp
```diff
@@ -0,0 +1,45 @@
+/*
```
TheMarex: I would not include empty unit tests - gives you a false sense of security.

TheMarex:
AppVayor seems to be failing because of this line:

`using NodeIDIter = typename std::vector<NodeID>::iterator;`

Maybe try without the `typename`

TODO:

- [ ] Fix the small remarks that I added above
- [ ] Run `clang-format -i path/to/file` on your code
- [ ] Move the `trip_` files from `routing_algorithms` to `algorithms` since they don't implement the `BasicRoutingInterface`

Otherwise this looks great. :+1:

data_structures/matrix_graph_wrapper.hpp
```diff
@@ -0,0 +1,65 @@
...
+
+#include <vector>
+
+//This Wrapper provides all methods that are needed for TarjanSCC, when the graph is given in a
+//matrix representation (e.g. as output from a distance table call)
+
+template <typename T> class MatrixGraphWrapper {
+public:
+
+    MatrixGraphWrapper(std::vector<T> table, const std::size_t number_of_nodes) : table_(table), number_of_nodes_(number_of_nodes) {};
```
daniel-j-h: `#include <cstddef>` for `std::size_t` used here

data_structures/matrix_graph_wrapper.hpp
```diff
@@ -0,0 +1,65 @@
...
+
+    std::size_t GetNumberOfNodes() const {
+        return number_of_nodes_;
+    }
+
+    std::vector<T> GetAdjacentEdgeRange(const NodeID node) const {
+        std::vector<T> edges;
+        for (auto i = 0; i < number_of_nodes_; ++i) {
+            if (*(std::begin(table_) + node * number_of_nodes_ + i) != INVALID_EDGE_WEIGHT) {
```
daniel-j-h: `#include <iterator>` for `std::begin` / `std::end`

data_structures/matrix_graph_wrapper.hpp
```diff
@@ -0,0 +1,65 @@
...
+                edges.push_back(i);
+            }
+        }
+        return edges;
+    }
+
+    EdgeWeight GetTarget(const EdgeWeight edge) const {
```
daniel-j-h: `#include "typedefs.h"` for `EdgeWeight`

data_structures/matrix_graph_wrapper.hpp
```diff
@@ -0,0 +1,65 @@
...
+        return edge;
+    }
```
daniel-j-h: I don't understand `GetTarget`, it is the identity function. It does nothing.

data_structures/matrix_graph_wrapper.hpp
daniel-j-h: If you take the `std::vector<T>` table by-value, you have to `std::move` it into the sink `table_`, otherwise you always have an extra copy.

That is, do

    table_(std::move(table)), ..

And then `#include <utility>` for `std::move`.

data_structures/matrix_graph_wrapper.hpp
```diff
@@ -0,0 +1,65 @@
...
+
+private:
+    std::vector<T> table_;
```
daniel-j-h: Can this be const, too?

data_structures/matrix_graph_wrapper.hpp
```diff
@@ -0,0 +1,65 @@
...
+
+    MatrixGraphWrapper(std::vector<T> table, const std::size_t number_of_nodes) : table_(table), number_of_nodes_(number_of_nodes) {};
+
+    std::size_t GetNumberOfNodes() const {
+        return number_of_nodes_;
+    }
+
+    std::vector<T> GetAdjacentEdgeRange(const NodeID node) const {
+        std::vector<T> edges;
+        for (auto i = 0; i < number_of_nodes_; ++i) {
```
daniel-j-h: `auto i = 0;` deduces `i` to type `int`, you want `std::size_t` though.

data_structures/matrix_graph_wrapper.hpp
daniel-j-h: If I get this line correctly, you want to filter a specific range for valid edges. The (including, excluding) range is defined by:

    table_[node * number_of_nodes + 0] - table_[node * number_of_nodes + number_of_nodes]

It would be 1/ more clear what you are doing here and 2/ more idiomatic to write something along the lines of:

    auto it = // iterator to start of range;
    auto last = // iterator to end of range;

    std::copy_if(it, last, std::back_inserter(edges), validEdge);

with `validEdge` being a lambda checking this:

    auto validEdge = [](const auto& edge){ return edge != INVALID_EDGE_WEIGHT; };

And rip out your hand-written loop logic.

library/osrm_impl.hpp
daniel-j-h: I guess it's because @chaupow has a old boost version installed globally and recent compiler versions (rightfully) warn for the deprecation of the register keyword in C++11.

@chaupow can you upgrade your boost installation? Or install a recent version locally, e.g. in your home directory and then tell CMake the path to this directory.

plugins/trip.hpp
```diff
@@ -0,0 +1,332 @@
...
+
+#include <cstdlib>
+#include <algorithm>
+#include <memory>
+#include <unordered_map>
+#include <string>
+#include <utility>
+#include <vector>
+#include <limits>
+#include <map>
```
daniel-j-h: You do not use anything from the `map` header. If so remove it. What about the `unordered_map` header? Could you please check all of those and only include what you use? (Or use the tool `include-what-you-use`)

plugins/trip.hpp
```diff
@@ -0,0 +1,332 @@
...
+        // in_component: all NodeIDs sorted by component ID
+        // in_range: index where a new component starts
+        //
+        // example: NodeID 0, 1, 2, 4, 5 are in component 0
+        //          NodeID 3, 6, 7, 8    are in component 1
+        //          => in_component = [0, 1, 2, 4, 5, 3, 6, 7, 8]
+        //          => in_range = [0, 5]
+        SCC_Component(std::vector<NodeID> in_component,
+                      std::vector<size_t> in_range)
+                      : component(in_component),
```
daniel-j-h: You take the two `std::vector<T>`s by-value, therefore you have to `std::move` them into their sinks, otherwise there exist (potentially) expensive copies.

plugins/trip.hpp
```diff
@@ -0,0 +1,332 @@
...
+                             "invalid scc component range");
+            BOOST_ASSERT_MSG([&in_range](){
+                                 for (std::size_t r = 0; r < in_range.size() - 1; ++r) {
+                                     if (in_range[r] > in_range[r+1]) {
+                                         return false;
+                                     }
+                                 }
+                                 return true;
+                             }(),
+                             "invalid component ranges");
```
daniel-j-h: Remove all of this; replace with `std::is_sorted(begin(in_range), end(in_range));`.

plugins/trip.hpp
```diff
@@ -0,0 +1,332 @@
...
+        //          => in_component = [0, 1, 2, 4, 5, 3, 6, 7, 8]
+        //          => in_range = [0, 5]
+        SCC_Component(std::vector<NodeID> in_component,
+                      std::vector<size_t> in_range)
+                      : component(in_component),
+                        range(in_range) {
+            range.push_back(in_component.size());
+            BOOST_ASSERT_MSG(in_component.size() >= in_range.size(),
+                             "scc component and its ranges do not match");
+            BOOST_ASSERT_MSG(*std::max_element(in_range.begin(), in_range.end()) < in_component.size(),
```
daniel-j-h: You never check for `.size() > 0`. For example, if `in_range` is empty, dereferencing the iterator returned by `std::max_element` here is undefined behavior. Same with `std::min_element` below.

plugins/trip.hpp
```diff
@@ -0,0 +1,332 @@
...
+                                     }
+                                 }
+                                 return true;
+                             }(),
+                             "invalid component ranges");
+          };
+
+        // constructor to use when whole graph is one single scc
+        SCC_Component(std::vector<NodeID> in_component)
+                      : component(in_component), range({0, in_component.size()}) {
```
daniel-j-h: By-value, thus move it!

    : component(std::move(in_component)), ..

plugins/trip.hpp
```diff
@@ -0,0 +1,332 @@
...
+                phantom_node_vector[i].erase(std::begin(phantom_node_vector[i]));
+            }
+            BOOST_ASSERT(phantom_node_vector[i].front().is_valid(facade->GetNumberOfNodes()));
+        }
+    }
+
+    // Object to hold all strongly connected components (scc) of a graph
+    // to access all graphs with component ID i, get the iterators by:
+    // auto start = std::begin(scc_component.component) + scc_component.range[i];
+    // auto end = std::begin(scc_component.component) + scc_component.range[i+1];
```
daniel-j-h: I think the interface is somewhat ugly.

1/ can this `SCC_Component` be reused outside of your TSP code? If so, there is no need for inlining it here.

2/ can you come up with a more elegant way to iterate, especially one that does not need to expose internal implementation details? Like a function `begin_for(ComponentId id);` and `end_for(ComponentId id);`, or something similar?

plugins/trip.hpp
```diff
@@ -0,0 +1,332 @@
...
+                             "invalid component ranges");
+          };
+
+        // constructor to use when whole graph is one single scc
+        SCC_Component(std::vector<NodeID> in_component)
+                      : component(in_component), range({0, in_component.size()}) {
+                      };
+
+        std::size_t GetNumberOfComponents() const{
+            return range.size() - 1;
```
daniel-j-h: `range.size()` can be `0`, in this case this wraps around to `std::numeric_limits<std::size_t>::max()`.

plugins/trip.hpp
daniel-j-h: What happens in case `in_component` is empty?

plugins/trip.hpp
```diff
@@ -0,0 +1,332 @@
...
+                                         return false;
+                                     }
+                                 }
+                                 return true;
+                             }(),
+                             "invalid component ranges");
+          };
+
+        // constructor to use when whole graph is one single scc
+        SCC_Component(std::vector<NodeID> in_component)
```
daniel-j-h: There is a technique, called delegating constructors, in which you call a different constructor from your own one.

Like this:

    struct S {
        S(int x, int y) { }
        S() : S(1, 2) { } // delegates to ^
    };

If you use this technique here, you still have all the assertions and validation done by the constructor above.

plugins/trip.hpp
```diff
@@ -0,0 +1,332 @@
...
+        scc.run();
+
+        std::vector<size_t> range_insertion;
+        std::vector<size_t> range;
+        range_insertion.reserve(scc.get_number_of_components());
+        range.reserve(scc.get_number_of_components());
+
+        std::vector<NodeID> components(number_of_locations, 0);
+
+        auto prefix = 0;
```
daniel-j-h: This is of type `int`, but you probably want `std::size_t`?

plugins/trip.hpp
```diff
@@ -0,0 +1,332 @@
...
+    // and returns an SCC_Component
+    SCC_Component SplitUnaccessibleLocations(const std::size_t number_of_locations,
+                                             const DistTableWrapper<EdgeWeight> & result_table) {
+
+        // Run TarjanSCC
+        auto wrapper = std::make_shared<MatrixGraphWrapper<EdgeWeight>>(result_table.GetTable(), number_of_locations);
+        auto scc = TarjanSCC<MatrixGraphWrapper<EdgeWeight>>(wrapper);
+        scc.run();
+
+        std::vector<size_t> range_insertion;
```
daniel-j-h: `std::size_t` instead of `size_t` everywhere if you want to be pedantic.

plugins/trip.hpp
```diff
@@ -0,0 +1,332 @@
...
+        std::vector<size_t> range;
+        range_insertion.reserve(scc.get_number_of_components());
+        range.reserve(scc.get_number_of_components());
+
+        std::vector<NodeID> components(number_of_locations, 0);
+
+        auto prefix = 0;
+        for (size_t j = 0; j < scc.get_number_of_components(); ++j){
```
daniel-j-h: You have multiple `scc.get_number_of_components()` calls here. I would help the compiler and do common subexpression elimination here, that is store the result as `const` somewhere, instead of repeated function calls.

plugins/trip.hpp
```diff
@@ -0,0 +1,332 @@
...
+    // takes the number of locations and its distance matrix,
+    // identifies and splits the graph in its strongly connected components (scc)
+    // and returns an SCC_Component
+    SCC_Component SplitUnaccessibleLocations(const std::size_t number_of_locations,
+                                             const DistTableWrapper<EdgeWeight> & result_table) {
+
+        // Run TarjanSCC
+        auto wrapper = std::make_shared<MatrixGraphWrapper<EdgeWeight>>(result_table.GetTable(), number_of_locations);
+        auto scc = TarjanSCC<MatrixGraphWrapper<EdgeWeight>>(wrapper);
+        scc.run();
```
daniel-j-h: Traversing components from TarjanSCC seem like a reasonable feature best implemented directly in [tiny_components](https://github.com/Project-OSRM/osrm-backend/blob/300d901618dd259cfba408d0e3d96aeec4f40369/algorithms/tiny_components.hpp).

plugins/trip.hpp
```diff
@@ -0,0 +1,332 @@
...
+            range.push_back(prefix);
+            prefix += scc.get_component_size(j);
+        }
+
+        for (size_t i = 0; i < number_of_locations; ++i) {
+            components[range_insertion[scc.get_component_id(i)]] = i;
+            ++range_insertion[scc.get_component_id(i)];
+        }
+
+        return SCC_Component(components, range);
```
daniel-j-h: You are passing a copy of both `components` and `range` to `SCC_Component`'s constuctor, as it takes its arguments by-value. But then both `components` and `range` go out of scope here, therefore are destroyed.

Safe those two copies and extra work by `std::move`ing both arguments in!

plugins/trip.hpp
```diff
@@ -0,0 +1,332 @@
...
+        json_result.values["permutation"] = json_permutation;
+    }
+
+    void ComputeRoute(const PhantomNodeArray & phantom_node_vector,
+                      const RouteParameters & route_parameters,
+                      const std::vector<NodeID> & trip,
+                      InternalRouteResult & min_route) {
+        // given he final trip, compute total distance and return the route and location permutation
+        PhantomNodes viapoint;
+        for (auto it = std::begin(trip); it != std::end(trip); ++it) {
```
daniel-j-h: This calls `std::end(trip)` on every iteration. Improve this by storing the end iterator (because it does not change):

    for (auto it = std::begin(trip), end = std::end(trip); it != end; ++it)

Or even better, more idiomatic and describes intend (and it really cool):

Let's say you have a trip `1 -> 2 -> 3 -> 4 -> 1`. What you really want is to connect all elements with their following one and the last one to the first one. This can be done by taking the `trip` vector, rotating it to the left by one, then zip-ing both ranges with a function that generates the `viapoint` and outputs it into the `segment_end_coordinates`.

    trip:        1  2  3  4
                  \  \  \  \
    out:           +  +  +  +
                  /  /  /  /
    trip_rot:    2  3  4  1

Here are the two key lines you need for this:

    std::rotate(begin(trip_rot), begin(trip_rot) + 1, end(trip_rot));
    std::transform(begin(trip), end(trip), begin(trip_rot), out, toViapoint);

where `out` is e.g. a `std::back_inserter` applied to the `segment_end_coordinates` like:

    auto out = std::back_inserter(min_route.segment_end_coordinates)

and `toViapoint` is a lambda that takes to `NodeId`s and returns a `PhantomNodes` containing the vias, like:

    auto toViapoint = [&] (const NodeId& lhs, const NodeId& rhs) {
        return PhantomNodes{phantom_node_vector[lhs][0], phantom_node_vector[rhs][0]};
    };

Hope that helps.

Refs.
 - http://en.cppreference.com/w/cpp/algorithm/rotate
 - http://en.cppreference.com/w/cpp/algorithm/transform

algorithms/tarjan_scc.hpp
```diff
@@ -52,6 +52,7 @@ SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 #include <unordered_map>
 #include <unordered_set>
 #include <vector>
+#include <map>
```
daniel-j-h: From what I can tell TarjanSCC needs the following includes:

- `<cstddef>` for `std::size_t`
- `<vector>` for `std::vector`
- `<memory>` for `std::shared_ptr`
- `<stack>` for `std::stack`
- `<algorithm>` for `std::min`, `std::count_if`

but non of those:

- `<unordered_map>`
- `<unordered_set>`
- `<map>`

plugins/trip.hpp
```diff
@@ -0,0 +1,332 @@
...
+    {
+        // check if all inputs are coordinates
+        if (!check_all_coordinates(route_parameters.coordinates)) {
+            return 400;
+        }
+
+        // get phantom nodes
+        PhantomNodeArray phantom_node_vector(route_parameters.coordinates.size());
+        GetPhantomNodes(route_parameters, phantom_node_vector);
+        auto number_of_locations = phantom_node_vector.size();
```
daniel-j-h: Can this be `const`?

plugins/trip.hpp
```diff
@@ -0,0 +1,332 @@
...
+
+        // compute the distance table of all phantom nodes
+        const auto result_table = DistTableWrapper<EdgeWeight>(*search_engine_ptr->distance_table(phantom_node_vector),
+                                                               number_of_locations);
+        if (result_table.size() == 0){
+            return 400;
+        }
```
daniel-j-h: You can already check this before wrapping the distance table into your `DistTableWrapper`.

plugins/trip.hpp
```diff
@@ -0,0 +1,332 @@
...
+
+        const constexpr std::size_t BF_MAX_FEASABLE = 10;
+        BOOST_ASSERT_MSG(result_table.size() > 0, "Distance Table is empty.");
```
daniel-j-h: `check_all_coordinates` from above already checks for at least two points. Can the distance table then ever be empty? In cases where there is no path, it should still return the distance table but with `std::numeric_limit<EdgeWeight>::max()` indicating invalid entries.

I think a better (read: more strict) check here would be to check for a distance table that has a size of `N x N` for `N` points.

plugins/trip.hpp
```diff
@@ -0,0 +1,332 @@
...
+        // get scc components
+        SCC_Component scc = [&](){
+            if (*std::max_element(result_table.begin(), result_table.end()) == INVALID_EDGE_WEIGHT) {
+                // compute all scc with tarjan
+                return SplitUnaccessibleLocations(number_of_locations, result_table);
+            } else {
+                // whole graph is one scc
+                std::vector<NodeID> location_ids(number_of_locations);
+                std::iota(std::begin(location_ids), std::end(location_ids), 0);
+                return SCC_Component(location_ids);
```
daniel-j-h: `location_ids` is copied into `SCC_Components`, but then goes out of scope ayway. Therefore, `std::move` it in!

plugins/trip.hpp
```diff
@@ -0,0 +1,332 @@
...
+        }
+
+        const constexpr std::size_t BF_MAX_FEASABLE = 10;
+        BOOST_ASSERT_MSG(result_table.size() > 0, "Distance Table is empty.");
+
+        // get scc components
+        SCC_Component scc = [&](){
+            if (*std::max_element(result_table.begin(), result_table.end()) == INVALID_EDGE_WEIGHT) {
+                // compute all scc with tarjan
+                return SplitUnaccessibleLocations(number_of_locations, result_table);
```
daniel-j-h: As I understad it, you check for a single `INVALID_EDGE_WEIGHT`, indicating disconnected components.

Instead of `std::max_element`, what you are really doing is to search for (at least one) invalid weights:

    if (std::find(begin(result_table), end(result_table), INVALID_EDGE_WEIGHT) != std::end(result_table))

This breaks early without the need to go through the whole `result_table` if it finds an invalid entry before the end.

plugins/trip.hpp
```diff
@@ -0,0 +1,332 @@
...
+            } else {
+                // whole graph is one scc
+                std::vector<NodeID> location_ids(number_of_locations);
+                std::iota(std::begin(location_ids), std::end(location_ids), 0);
+                return SCC_Component(location_ids);
+            }
+        }();
```
daniel-j-h: Using the lambda-trick here is fine, but I can also think of this being a `DistTableWrapper` member function, returning its scc. This would eliminate the inline code here and encapsulate it where it logically should belong.

plugins/trip.hpp
```diff
@@ -0,0 +1,332 @@
...
+
+
+        using NodeIDIterator = typename std::vector<NodeID>::const_iterator;
```
daniel-j-h: Strictly speaking you don't need the `typename` here, as `NodeID` is a concrete type.

Where you do have to tell the compiler "hey this really is a type and not a value" with `typename` is in the context of dependent types, such as:

    template <typename T>
    using Iter = std::vector<T>::const_iterator; // fails, needs typename in front of it

Take a look here: https://github.com/cplusplus/draft/commit/253e03fe84cf0c55c95c9f779fadfa0d026b701b (C++14 standard pdf), p360, 14.6/2.

Ref. http://en.cppreference.com/w/cpp/language/dependent_name

plugins/trip.hpp
```diff
@@ -0,0 +1,332 @@
...
+            BOOST_ASSERT_MSG(component_size >= 0,"invalid component size");
+
+            if (component_size > 1) {
+                std::vector<NodeID> scc_route;
+                NodeIDIterator start = std::begin(scc.component) + scc.range[k];
+                NodeIDIterator end = std::begin(scc.component) + scc.range[k+1];
+
+                if (component_size < BF_MAX_FEASABLE) {
+                    scc_route = osrm::trip::BruteForceTrip(start, end, number_of_locations, result_table);
+                    route_result.push_back(scc_route);
```
daniel-j-h: Unfortunately you need `scc_route` for logging below, otherwise you could `std::move` it here, avoiding a copy. Maybe restructure the logging? Or create a temp. string for the logging and keep it around for a bit.

plugins/trip.hpp
```diff
@@ -0,0 +1,332 @@
...
+                } else {
+                    scc_route = osrm::trip::FarthestInsertionTrip(start, end, number_of_locations, result_table);
+                    route_result.push_back(scc_route);
```
daniel-j-h: Same for `std::move` here, see ^: https://github.com/Project-OSRM/osrm-backend/pull/1610/files#r38188689

plugins/trip.hpp
```diff
@@ -0,0 +1,332 @@
...
+                // use this if output if debugging of route is needed:
+                SimpleLogger().Write() << "Route #"
+                                       << k << ": "
+                                       << [&scc_route](){
+                                               std::string s = "";
+                                               for (auto x : scc_route) {
+                                                   s += std::to_string(x) + " ";
+                                               }
+                                               return s;
+                                           }();
```
daniel-j-h: You could avoid the move-assignment operator above, by doing it like this:

    route_result.push_back(osrm::trip::FarthestInsertionTrip(start, end, number_of_locations, result_table));

And then do the logging by accessing `route_result.back()`.

Also, a nice way of streaming out iterables with a separator is the following:

    std::copy(begin(vec), end(vec), std::ostream_iterator<std::string>(std::cout, " "));

plugins/trip.hpp
```diff
@@ -0,0 +1,332 @@
...
+            } else {
+                // if component only consists of one node, add it to the result routes
+                route_result.push_back({scc.component[scc.range[k]]});
```
daniel-j-h: The `{ .. }` are for creating a temporary --- what you could instead do is `.emplace_back( .. )` to construct an item inplace, without: temporary -> move from temporary -> destruct temporary.

plugins/trip.hpp
```diff
@@ -0,0 +1,332 @@
...
+            }
+        }
+
+        // compute all round trip routes
+        std::vector<InternalRouteResult> comp_route (route_result.size());
+        for (std::size_t r = 0; r < route_result.size(); ++r) {
+            ComputeRoute(phantom_node_vector, route_parameters, route_result[r], comp_route[r]);
+        }
+
+        TIMER_STOP(trip);
```
daniel-j-h: You never actually use the timers, e.g. by means of calculations the difference --- can it be removed?

plugins/trip.hpp
```diff
@@ -0,0 +1,332 @@
...
+            } else {
+                // if component only consists of one node, add it to the result routes
+                route_result.push_back({scc.component[scc.range[k]]});
+            }
+        }
+
+        // compute all round trip routes
+        std::vector<InternalRouteResult> comp_route (route_result.size());
+        for (std::size_t r = 0; r < route_result.size(); ++r) {
+            ComputeRoute(phantom_node_vector, route_parameters, route_result[r], comp_route[r]);
```
daniel-j-h: Instead of out-references better return the min. route by value (that is, move it out) in `ComputeRoute`.

This way, you can use something like this here:

    std::transform(begin(route_result), end(route_result), begin(comp_route), getRoute);

with `getRoute` being a lambda, like:

    auto getRoute = [&] (const auto& route_result) {
      return ComputeRoute(phantom_node_vector, route_parameters, route_result);
    };

plugins/trip.hpp
```diff
@@ -0,0 +1,332 @@
...
+        }
+
+        TIMER_STOP(trip);
+
+        // prepare JSON output
+
+
+        // create a json object for every trip
+        osrm::json::Array trip;
```
daniel-j-h: I would be cautious with naming this `trip` when your timer above is also "named" `trip`. The timer macro may internally rename / prefix the local variable it creates, but this is an implementation detail.

plugins/trip.hpp
```diff
@@ -0,0 +1,332 @@
...
+        for (std::size_t i = 0; i < route_result.size(); ++i) {
+            std::unique_ptr<BaseDescriptor<DataFacadeT>> descriptor;
+            descriptor = osrm::make_unique<JSONDescriptor<DataFacadeT>>(facade);
```
daniel-j-h: Instead of creating a "empty" base descriptor `unique_ptr` and then initializing it later, you can merge those two lines.

Let's take a step back here. What's the reason for using a `unique_ptr` here? From what I can see it goes out of scope a few lines later anyway, without you transferring ownership to somewhere else.

Is it just for the polymorphic runtime function dispatch? You can do this with references, too, like:

    struct base { };
    struct derived : base { };

    derived d;
    base& b = d;

Now calling functions on `b` (the reference to a `base` type) are under runtime dispatch and will call `derived` functions when they override `base` ones. This does not require `unique_ptr`s at all.

plugins/trip.hpp
```diff
@@ -0,0 +1,332 @@
...
+            descriptor->SetConfig(route_parameters);
+
+            osrm::json::Object scc_trip;
+
+            // set permutation output
+            SetLocPermutationOutput(route_result[i], scc_trip);
+            // set viaroute output
+            descriptor->Run(comp_route[i], scc_trip);
+
+            trip.values.push_back(scc_trip);
```
daniel-j-h: `scc_trip` is not needed anymore, `std::move` it in!

plugins/trip.hpp
```diff
@@ -0,0 +1,332 @@
...
+        }
+
+        json_result.values["trips"] = trip;
```
daniel-j-h: `trip` is not needed anymore --- if this is a map, you can `.insert` and then `std::move()` it in, as a map's `operator[]` does not provide an rvalue overload.

routing_algorithms/trip_brute_force.hpp
```diff
@@ -0,0 +1,104 @@
...
+(INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
+LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON
+ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+(INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+*/
+
+#ifndef Trip_BRUTE_FORCE_HPP
+#define Trip_BRUTE_FORCE_HPP
```
daniel-j-h: Everything upper case for include guards (?)

routing_algorithms/trip_brute_force.hpp
```diff
@@ -0,0 +1,104 @@
...
+namespace osrm
+{
+namespace trip
+{
+
+// computes the distance of a given permutation
+EdgeWeight ReturnDistance(const DistTableWrapper<EdgeWeight> & dist_table,
+                          const std::vector<NodeID> & location_order,
+                          const EdgeWeight & min_route_dist,
+                          const std::size_t & component_size) {
```
daniel-j-h: Passing `std::size_t` by `const&` --- if anything at all --- makes the compiler's job in doing optmizations harder. Just pass fundamental types such as bools, ints, doubles, etc. by value.

routing_algorithms/trip_brute_force.hpp
```diff
@@ -0,0 +1,104 @@
...
+    EdgeWeight route_dist = 0;
+    std::size_t i = 0;
+    while (i < location_order.size() && (route_dist < min_route_dist)) {
+        route_dist += dist_table(location_order[i], location_order[(i+1) % component_size]);
+        ++i;
+    }
+
+    BOOST_ASSERT_MSG(route_dist != INVALID_EDGE_WEIGHT, "invalid route found");
```
daniel-j-h: I think what you want to check is this condition, but in every iteration. That is, in case a single distance table entry is invalid, you want the assert to trigger. If you don't check this inside the loop, then adding a valid edge weight to an invalid one does not result in an invalid edge weight, but in a valid one that is way off (because of wrap around).

routing_algorithms/trip_brute_force.hpp
```diff
@@ -0,0 +1,104 @@
...
+    std::vector<NodeID> perm(start, end);
+    std::vector<NodeID> route;
+    route.reserve(component_size);
+
+    EdgeWeight min_route_dist = INVALID_EDGE_WEIGHT;
+
+    // check length of all possible permutation of the component ids
+
+    BOOST_ASSERT_MSG(*(std::max_element(std::begin(perm), std::end(perm))) < number_of_locations, "invalid node id");
+    BOOST_ASSERT_MSG(*(std::min_element(std::begin(perm), std::end(perm))) >= 0, "invalid node id");
```
daniel-j-h: 1/ you need to `#include <iterator>` for the freestanding functions `std::begin` and `std::end`.

2/ dereferencing `std::max_element` requires that `perm`'s size is `> 0`, otherwise this is undefined behavior. Maybe assert this (on the `component_size` above).

data_structures/matrix_graph_wrapper.hpp
chaupow: Yea, you're right but the wrapper needs to provide the function in order to use the Tarjan Algorithm.

plugins/trip.hpp
chaupow: i don't know why, but when i do that, it breaks the plugin. you got a hunch, why @daniel-j-h ?

data_structures/matrix_graph_wrapper.hpp
chaupow: @daniel-j-h: I have tried and thought about this the last time you told me this. but i don't want to copy the value that is in the table (which is the edge weight) but i want to add the node id of the other node to which i check whether the distance is valid.

i cant figure out how to do that using `copy_if`. do you have a better idea?

plugins/trip.hpp
chaupow: @daniel-j-h: do you mean the struct `SCC_Component` or the function `SplitUnaccessibleLocations `? i'm not sure about this. the struct might come handy in other cases too, that's true. i know little about the actual work that tarjan_scc is doing. 

@TheMarex what do you think? would it fit to include the struct to tarjan_scc?

plugins/trip.hpp
chaupow: @daniel-j-h ?

plugins/trip.hpp
chaupow: @daniel-j-h : no idea why but i cant get this to work.

shouldnt it be just 

```
SCC_Component(std::vector<NodeID> in_component)
: SCC_Component(in_component, {0, in_component.size()}) {};
```

?

algorithms/trip_farthest_insertion.hpp
```diff
@@ -0,0 +1,211 @@
...
+        auto to_node = std::next(from_node);
+        if (to_node == end)
+        {
+            to_node = start;
+        }
+
+        const auto dist_from = dist_table(*from_node, new_loc);
+        const auto dist_to = dist_table(new_loc, *to_node);
+        const auto trip_dist = dist_from + dist_to - dist_table(*from_node, *to_node);
+        ;
```
daniel-j-h: Suspicious additional `;` here.

algorithms/trip_farthest_insertion.hpp
```diff
@@ -0,0 +1,211 @@
...
+        auto next_node = -1;
+        NodeIDIter next_insert_point;
+
+        // find unvisited loc i that is the farthest away from all other visited locs
+        for (auto i = start; i != end; ++i)
+        {
+            // find the shortest distance from i to all visited nodes
+            if (!visited[*i])
+            {
+                auto insert_candidate =
```
daniel-j-h: You don't modify `insert_candidate` from here on, this can be `const`.

algorithms/trip_nearest_neighbour.hpp
```diff
@@ -0,0 +1,122 @@
...
+        EdgeWeight trip_dist = 0;
+        for (std::size_t via_point = 1; via_point < component_size; ++via_point)
+        {
+            EdgeWeight min_dist = INVALID_EDGE_WEIGHT;
+            NodeID min_id = SPECIAL_NODEID;
+
+            // 2. FIND NEAREST NEIGHBOUR
+            for (auto next = start; next != end; ++next)
+            {
+                auto curr_dist = dist_table(curr_node, *next);
```
daniel-j-h: You don't modify `curr_dist` from here on, this can be `const`.

algorithms/trip_nearest_neighbour.hpp
```diff
@@ -0,0 +1,122 @@
...
+            trip_dist += min_dist;
+            curr_node = min_id;
+        }
+
+        // check round trip with this starting point is shorter than the shortest round trip found
+        // till now
+        if (trip_dist < shortest_trip_distance)
+        {
+            shortest_trip_distance = trip_dist;
+            route = curr_route;
```
daniel-j-h: This calls the copy assignment operator, but it should be possible to call the move assignment operator by `std::move`ing `curr_route`, as you dont't need it anymore from here on.

data_structures/matrix_graph_wrapper.hpp
daniel-j-h: Ah, I see. Hmm yeah that's a different problem then, indeed. You could make use of some iterator utilities from boost, like `irange`, or `indexed` --- but I guess this would be overkill here.


References;

- http://www.boost.org/doc/libs/1_46_1/libs/range/doc/html/range/reference/ranges/irange.html
- http://www.boost.org/doc/libs/1_46_1/libs/range/doc/html/range/reference/adaptors/reference/indexed.html

data_structures/matrix_graph_wrapper.hpp
daniel-j-h: Ah! Is there an interface for that? I can see `GetTarget` member function being used in the `TarjanSCC` implementation, but where do you get the information from which member functions are required? That is, what are the requirements for the `GraphT` type?

plugins/trip.hpp
daniel-j-h: What `breaks`? And how exactly? It is perfectly safe to do here:

    : component(std::move(in_component)), ..

But then also make sure to use `component` instead of `in_component` in the constructor, as `in_component` is an an unspecified invalid state after the move, and may e.g. return `true` for `.empty()` (or `0` for `.size()` and so on). That is, just don't access an object that you moved from.

plugins/trip.hpp
daniel-j-h: Yes, except that the extra `;` at the end is not needed and you make an extra copy when taking `in_component` by-value (but a move would be problematic here since you need the vector's size). Otherwise this *looks* fine.

plugins/trip.hpp
daniel-j-h: Currently iterating over components can be done manually using [these member functions](https://github.com/Project-OSRM/osrm-backend/blob/300d901618dd259cfba408d0e3d96aeec4f40369/algorithms/tiny_components.hpp#L227-L236) from `TarjanSCC`.

If you check your code that follows, it does exactly that. I think you could move the manual iteration out to `TarjanSCC`, wrap it once with iterators, and then use it here, Because I guess iterating over components is a reasonable feature for `TarjanSCC` that is not only needed here.

plugins/trip.hpp
daniel-j-h: Storing the raw distance table from:

    ptr->distance_table(phantom_node_vector)

Then doing the size check, and only then moving the distance table into your wrapper.

zavan:
Hello, I am very interested on this feature. If I compile the develop branch, is this going to be available on the json api already, or is it not ready yet?

danpat:
@zavan Yep, it's in `develop` now.  I don't think the API wiki has been updated yet, but the request format should be something like:

    /trip?loc=lat1,lon1&loc=lat2,lon2&loc=lat3,lon3

similar to the other API calls.

zavan:
Thank you very much!
Gonna try it

zavan:
Works perfectly, great work @chaupow, this will help me a lot on a near future.

But unfortunately, right now, that won't do for me, I need a optmized route passing through ~300 waypoints, without the return trip, and the starting point needs to be fixed. Does OSRM happen to have something like this?

danpat:
Hi @zavan The plugin is new, so it doesn't have that feature yet.  Feel free to submit a patch!

zavan:
Unfortunately I don't yet have the knowledge (C++, C++ testing and OSRM internals) to create this, and I'm pretty busy, but I'll talk to some people on the company and see if I can get them to work on this.

Thanks for the help @danpat, and again, great work @chaupow, this is very promising!

chaupow:
Hey @zavan,

glad you like it. I've already been thinking about the problem you described and I have some ideas how that could be implemented but I have not found the time to try them out (yet? hopefully).

I'm just curious: How is the runtime for your application? 300 are quite a few waypoints. Is it working for you? :)

zavan:
Right now we're using Google Maps API Enterprise for this, so it's veeery slow, we have to queue the requests, which for now is not a problem, but it will become soon, that's why we're trying an in-house OSRM solution.

Our use case, which I described on my last comment, is, I believe, a specialization of this one, I just forgot to mention that only our starting point is fixed, our end point is not fixed, it can be any of the ~300 points, provided that the overall route is one of the shortest we can get with OSM data.

We need someone with the knowledge to help us with this. If you're available for this job, @chaupow, we could negotiate something, email me if you're interested: felipe@apeiara.com.

Issue #1611, opened 2015-08-19
Use new libosmium version 2.3.0


joto:
https://github.com/osmcode/libosmium/tree/v2.3.0

Has faster PBF reading!

daniel-j-h:
We are working on merging this / this is basically done and works now with Windows issues fixed. Closing here.

PR #1612, opened 2015-08-19
 Pull in latest libosmium release.

daniel-j-h:
This updates the current `libosmium` version 2.2.0 dependency to the
recent 2.3.0 release.

Interesting for us:

- mapbox/protozero based protobuf handling
- empty version in pbf files

References:

- #1611
- https://github.com/osmcode/libosmium/blob/c43f8db50d93912a8bec5cd9fea733f7fec05549/CHANGELOG.md#230---2015-08-18
- osmcode/libosmium#102


TheMarex:
Seems like `libosmium` broke on windows.

https://ci.appveyor.com/project/DennisOSRM/osrm-backend/build/1.0.1082#L1573

TheMarex:
@BergWerkGIS it seems like some of the dependencies do something like [this](http://www.enchantedage.com/node/231) and thus `libosmium` produces this error:

```
 C:\projects\osrm\third_party\libosmium\include\osmium/osm/types_from_string.hpp(53): error C2039: '_strtoi64': is not a member of 'std' [C:\projects\osrm\build\osrm-extract.vcxproj]
```

@danpat suggest `cl.exe /E` might help here to print the expanded file with all macro definitions?

joto:
Okay, this proves that we have this problem with the define. I don't think this is something to fix in libosmium, because libosmium builds fine alone. So newer versions of Visual Studio obviously support `strtoll()`. I suggest you find the culprit in the the other dependencies and `#undef stroll` after including it and report it as a bug in their code.

springmeyer:
one note: the osrm appveyor scripts are building with vs 2015 but libosmium appveyor is still only testing with vs 2014. So there may be a chance that this is replicable with libosmium itself if we were to text with 2015 (which we should) /cc @BergWerkGIS

BergWerkGIS:
Y, osrm and libosmium are bit out of sync and behind on the move to VS2015.
I will look into those after the current mapnik/Mapbox Studio push.

joto:
@springmeyer that's why i tried libosmium with VS2015 yesterday in https://github.com/osmcode/libosmium/pull/106 and it compiled fine. So this is not an VS2014 vs VS2015 issue.

springmeyer:
@joto - ah, okay.

BergWerkGIS:
@TheMarex 
AppVeyor green: https://ci.appveyor.com/project/DennisOSRM/osrm-backend/build/1.0.1117

TheMarex:
Rebasing this on `develop` currently creates a huge merge conflict. Looking [at the diff](https://github.com/Project-OSRM/osrm-backend/commit/5714233989fec46c06c49b9c22c8d9818158e08b) more carefully you can see that some files in the repository root are overwritten with the ones from`libosmium`. e.g. [this file](https://github.com/Project-OSRM/osrm-backend/blob/5714233989fec46c06c49b9c22c8d9818158e08b/CHANGELOG.md).

@daniel-j-h can you try that again maybe?

daniel-j-h:
Hmm let me do this again and verify the update script works as expected.

daniel-j-h:
@TheMarex this should be fine: https://github.com/Project-OSRM/osrm-backend/pull/1626. Please check it, should only touch the directory `third_party`. Not sure how it ended up overwriting files outside of it.

joto:
Could you also test with current libosmium master? I will release a new version of libosmium soon, so would be good to see if there are any last-minute problems.

daniel-j-h:
@joto with [latest master](https://github.com/osmcode/libosmium/tree/e081a516c256a49145fd15699715ca862bff4213):

- [x] osrm compiles
- [x] unit tests ok
- [x] cucumber tests ok

daniel-j-h:
Windows issues resolved here, doing a clean pull in https://github.com/Project-OSRM/osrm-backend/pull/1626. Closing this.

PR #1613, opened 2015-08-19
Add script to update subtree-ed third party dependencies more easily.

daniel-j-h:
Note: this updates the subtrees immediately.

Discussion: would it make sense to do something along the lines of:

```bash
$ http --body https://api.github.com/repos/mapbox/variant/releases/latest | jq ".tag_name"
"v1.0"
```

And warn the user if the latest release tag is not the tag the update
script was called with. Or at least ask for confirmation?

### DIFF
--- a/update_depdendencies.sh
+++ b/update_depdendencies.sh
@@ -16,6 +16,6 @@ read -p "Looks good? (Y/n) " ok
 
 if [[ $ok =~ [yY] ]]
 then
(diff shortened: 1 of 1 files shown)


TheMarex:
Hm, yeah showing the availability of new version would be nice - but only if this is like a one line thing. Otherwise this will degenerate to a package manager. :wink:

Issue #77, opened 2012-01-14
Make travel time and edge weights independent


emiltin:
i think it will be imporant to be able to set 'attractiveness' on ways depending on certain tags.

this should be separate from the speed, since it's still important to have the correct speeds for each sections, so that a estimated travel time can be computed.

for example, we want to prioritize streets with bike tracks over streets without, even though the speed might be the same, or slightly lower. the same goes for green routes, which people often prefer over a big road, even though the speed might be similar or lower. 

one way it could be done simply is by hardcoding the tag combinations in osrm, but allowing the weight to be set in the speedprofile. something like:

   if(accesTag=='bicycle' &&  cycleway=='track')
        attractiveness = w.keyVals.Find("bike_track_attractiveness");    //value set in speedprofile, say 1.4
   else if(accesTag=='bicycle' &&  cycleway=='lane')
        attractiveness = w.keyVals.Find("bike_lane_attractiveness");    //value set in speedprofile, say 1.2
   else if(accesTag=='bicycle' &&  route != nil)
        attractiveness = w.keyVals.Find("bike_route_attractiveness");    //value set in speedprofile, say 1.5
  //other cases goes here....
  else
        attractiveness = 1.0;

eleu:
I'd ask the same question as @emiltin about any updates on this one, agreeing that this would be a very useful feature.

emiltin:
+1

emiltin:
hi @TheMarex, what are the plans for separating travel time and impedance? it's quite akward having to always modify travel time to prioritized certain ways over other.

for example, we're implemeting green routes for bikes, which follows parks, lakes etc. currently the only way to do that is to modify the travel time, resulting in weird travel times being returned by OSRM. this means we  we need to fall back on hacks like ignoring what osrm provides and instead computing a generic travel time in the front-end as distance/average_bike_speed.

TheMarex:
@emiltin This is certainly desirable, _but_ would mean a lot higher memory usage. An idea would be to hold auxiliary data for each edge out of main memory (for example street name, OSM id, height difference) that would not get used for routing but could be returned with the response to augment the route.

emiltin:
yes it would require more memory, but only during preparing, right? also adding something like 8 bit for each edge might not be a huge increase compared to what's curently used?

not sure isee how storing auxilary data would solve the problem of  speed and impedance not being separated, but maybe i misunderstand you?

emiltin:
[...]

PR #1614, opened 2015-08-19
[Not ready] Weight edges

Tristramg:
We want to be able to compute the optimal route according to an arbitrary objective and not only the duration (e.g. avoid a road with heavy traffic on a bike).

The usual trick is to modify the speed to penalize unwanted ways. The problem is that is produces bad duration estimations.

This PR fixes that by keeping the duration to output it correctly.

In the lua profile, it is possible to define a `forward_weight` (or `backward_weight`) on the way. That value will be used for the route computation, but the duration will be returned. If that weight is not given, it fallbacks on the speed. Existing profiles continue to work.

It is not ready:

- [ ] phantom nodes are not evaluated
- [ ] memory consumption should be measured and minimized
- [x] tests are broken :sweat_smile: 
- [x] it is not tested
- [ ] maybe it could be more generic


emiltin:
nice, this is a long-standing issue :-) #77

contractor/edge_based_graph_factory.cpp
```diff
@@ -120,27 +120,27 @@ void EdgeBasedGraphFactory::InsertEdgeBasedNode(const NodeID node_u,
...
 
-        std::vector<int> forward_dist_prefix_sum(forward_geometry.size(), 0);
-        std::vector<int> reverse_dist_prefix_sum(reverse_geometry.size(), 0);
+        std::vector<int> forward_weight_prefix_sum(forward_geometry.size(), 0);
+        std::vector<int> reverse_weight_prefix_sum(reverse_geometry.size(), 0);
 
         // quick'n'dirty prefix sum as std::partial_sum needs addtional casts
         // TODO: move to lambda function with C++11
-        int temp_sum = 0;
+        int temp_weight_sum = 0;
```
daniel-j-h: Is there a possibility of an overflow (i.e. UB) here?

data_structures/import_edge.hpp
```diff
@@ -51,6 +52,7 @@ struct NodeBasedEdge
     NodeID target;
     NodeID name_id;
     EdgeWeight weight;
+    EdgeWeight duration;
```
daniel-j-h: Why the manual padding below, but not doing proper padding here?

data_structures/import_edge.hpp
```diff
@@ -73,12 +75,14 @@ struct EdgeBasedEdge
...
                            const EdgeWeight weight,
+                           const EdgeWeight duration,
                            const bool forward,
                            const bool backward);
     NodeID source;
     NodeID target;
     NodeID edge_id;
-    EdgeWeight weight : 30;
+    EdgeWeight weight;
+    EdgeWeight duration : 30;
```
daniel-j-h: This is further constraining the duration type --- possibility of overflows?

data_structures/node_based_graph.hpp
```diff
@@ -80,8 +81,9 @@ NodeBasedDynamicGraphFromEdges(int number_of_nodes, const std::vector<NodeBasedE
     auto edges_list = directedEdgesFromCompressed<NodeBasedDynamicGraph::InputEdge>(input_edge_list,
         [](NodeBasedDynamicGraph::InputEdge& output_edge, const NodeBasedEdge& input_edge)
         {
-            output_edge.data.distance = static_cast<int>(input_edge.weight);
-            BOOST_ASSERT(output_edge.data.distance > 0);
+            output_edge.data.weight = static_cast<int>(input_edge.weight);
+            output_edge.data.duration = static_cast<int>(input_edge.duration);
+            BOOST_ASSERT(output_edge.data.weight > 0);
```
daniel-j-h: Are those casts safe? Those are narrowing conversions.

Discussion: maybe we should think about either providing a `safe_cast` or using `boost::numeric_cast` --- not just for these two casts but globally.

contractor/edge_based_graph_factory.cpp
Tristramg: The idea is that the weight is of the order of magnitude of the duration (so with an UB of ~100 days). However, the weight has now its own 32bits, so the actual UB is even higher.

If someone decides that the weight are 1% of the duration across the world, it might be a problem. I’m not sure where that should be checked.

data_structures/import_edge.hpp
Tristramg: The padding below allows to have a smaller struct. We wouldn’t gain anything from a memory standpoint. However from a coherence point of view do we want the same padding everywhere.

data_structures/import_edge.hpp
Tristramg: Until now, the weight was the duration. The upper bounds are known and this won’t break things. However if someone has strange ideas concerning the weight, the are more chances that bigger values are obtained. That is why I kept the duration padded and not the weight.

Tristramg:
Small update concerning the memory consumption. I could not measure a significant difference. This is a screenshot of htop with the old and new instance with BadenWurttemberg data. Only 2Mb difference, ~1%.

screenshot from 2015-08-21 14 11 55

Using valgrind massif (that has significantly less overhead than other valgrind tools), I could not see a difference either.

It says that 93% of the memory usage is in:

`->93.23% (99,238,752B) 0x47377C: InternalDataFacade<QueryEdge::EdgeData>::InternalDataFacade(std::unordered_map<std::string, boost::filesystem::path, std::hash<std::string>, std::equal_to<std::string>, std::allocator<std::pair<std::string const, boost::filesystem::path> > > const&) [clone .local.97] (in /home/tristram/mapbox/osrm-backend/build-old/osrm-routed)`

The QueryEdge::EdgeData was not modified and has not an extra field. That is why adding the duration doesn’t change much the general memory usage.

I think I missed some very important thing. Are compressed edges written to disk and not loaded in `osrm-routed` and read directly from disk when build the path that will be presented?

On disk, those are the only differences between the old and new:

```
< 116M	data/baden-wuerttemberg-latest.osrm
---
> 130M	data/baden-wuerttemberg-latest.osrm

< 158M	data/baden-wuerttemberg-latest.osrm.fileIndex
---
> 185M	data/baden-wuerttemberg-latest.osrm.fileIndex
```

data_structures/query_edge.hpp
```diff
@@ -50,7 +50,7 @@ struct QueryEdge
         }
         NodeID id : 31;
         bool shortcut : 1;
-        int distance : 30;
```
TheMarex: You forgot to extend this: You need two fields here, one for `distance` and one for `duration`.

routing_algorithms/routing_base.hpp
```diff
@@ -261,7 +261,7 @@ template <class DataFacadeT, class Derived> class BasicRoutingInterface
                                                    TurnInstruction::NoTurn, 0, travel_mode);
                     }
                     unpacked_path.back().turn_instruction = turn_instruction;
-                    unpacked_path.back().segment_duration = ed.distance;
```
TheMarex: This needs to use `duration` instead of `distance`

routing_algorithms/routing_base.hpp
```diff
@@ -236,7 +236,7 @@ template <class DataFacadeT, class Derived> class BasicRoutingInterface
                 {
                     BOOST_ASSERT(!facade->EdgeIsCompressed(ed.id));
                     unpacked_path.emplace_back(facade->GetGeometryIndexForEdgeID(ed.id), name_index,
-                                               turn_instruction, ed.distance, travel_mode);
```
TheMarex: This needs to use `duration`.

TheMarex:
The low memory usage is due to the fact that the `duration` value never reaches the server component. You forgot to extend `QueryEdge`.

emiltin:
maybe 30 bits is not needed for duration/speed?

routing_algorithms/routing_base.hpp
Tristramg: Aha! That’s was what I overlooked. Thank you. I will fix that ICE back to Paris.

Tristramg:
@emiltin for a single edge indeed 30 bits are plenty. However the contraction hierarchies generates edges that are the result of many contracted edges, and according to @TheMarex there are some situations where 30 bits are required (a walking edge from south Chile to Canada). The shortcuts generated are weird :)

emiltin:
i see

Tristramg:
I changed it. The memory consumption evolves as predicted: on BW, it increases from 188Mb to 220 measured by valgrind massif.

I added a test that catches my mistake but it still fails. I am not sure when I’ll have some time to find out what I forgot.

PR #1625, opened 2015-08-22
[cucumber] Use rounding for speed

lbud:
In writing [cucumber tests for raster sources](https://github.com/Project-OSRM/osrm-backend/blob/07a28c18a95e57cb4f31ecad4908e7cf80da6353/features/raster/weights.feature#L40-L50) I noticed something weird was happening with speeds; namely, edges marked `highway=primary` and evaluated with `testbot` heuristics (primary = 36mph) were sometimes resulting in `speed=36 km/h` and sometimes `speed=35 km/h`. In investigating it seems this is the result of `35.999.to_i` — ruby truncates decimals to the nearest int rather than rounding. This PR changes from `to_i` to `round` and also modifies any tests that subsequently failed. (This change to routing.rb fixes the failing cucumber tests in the raster source branch as well.)

@emiltin — at quick glance I imagine these test fixtures were written based on the results of the speed conversion in routing.rb, so I assume it's okay to change them based on switching from `to_i` to `round`. Is this okay?

cc @TheMarex

### DIFF
--- a/features/bicycle/bridge.feature
+++ b/features/bicycle/bridge.feature
@@ -42,6 +42,6 @@ Feature: Bicycle - Handle movable bridge
         When I route I should get
             | from | to | route       | modes | speed   |
             | a    | g  | abc,cde,efg | 1,5,1 | 5 km/h |
-            | b    | f  | abc,cde,efg | 1,5,1 | 3 km/h |
+            | b    | f  | abc,cde,efg | 1,5,1 | 4 km/h |
             | c    | e  | cde         | 5     | 2 km/h |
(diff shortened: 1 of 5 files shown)


emiltin:
yes makes sense.

TheMarex:
AppVeyor failture might be related to https://github.com/Project-OSRM/osrm-backend/commit/40443d1e25a5053d543621d03e106cffbfa37ff4 Going to pull that into `develop` and re-base this branch.

TheMarex:
There we go. Cucumber is green. -> merging by hand.

PR #1626, opened 2015-08-25
Use libosmium 2.3.0 (noconflict)

daniel-j-h:
Used the `update_dependencies.sh` script to update the `third_party` libraries.

Making sure it touches only the `third_party` directory:

    git show --pretty="format:" --name-only |grep -v third_party

Ref. https://github.com/Project-OSRM/osrm-backend/pull/1612#issuecomment-134537811


TheMarex:
@daniel-j-h can you push an empty commit to re-trigger AppVeyor? Seems to be brolen.

TheMarex:
You are missing https://github.com/Project-OSRM/osrm-backend/commit/b025179a8c31a365b803c6a800ec307c2d80105b to from the other PR to make AppVeyor work.

daniel-j-h:
Cherry-picked the commit you mentioned.

I'm not used to AppVeyor, anything I have to look out for? What's the general starting point / way for debugging those issues / taking a look at results?

springmeyer:
@daniel-j-h - looks like you correctly cherry-picked the commit which seemed to fix this error (`error C2039: '_strtoi64': is not a member of 'std'`) previously. I wonder if somehow appveyor pulled the wrong code for that commit. Want to try pushing one more empty commit just to re-trigger things? /cc @BergWerkGIS for any further ideas.

BergWerkGIS:
Solution might be same as @joto suggested last time:
https://github.com/Project-OSRM/osrm-backend/pull/1612#issuecomment-132926847

springmeyer:
@BergWerkGIS - nice hint, thank you. I now understand what @joto was talking about and it makes sense. So the fix should be easy. What I gather is happening is that:

  - `libosmium` code is fine
  - what changed however is that new `libosmium` started using `std::strtoll` only recently
  - the breakage is coming from `stxxl`: https://github.com/stxxl/stxxl/blob/master/include/stxxl/bits/msvc_compatibility.h#L34
  - a quick solution should be possible by avoiding that `stxxl` header from being used or at least included before the `libosmium` header

TheMarex:
@springmeyer Wow awesome for figuring this out. Lets go with switching headers and hoping for the best.

@BergWerkGIS  What is `_MSC_VER` for VS2015 so that we can get a patch upstream to prevent this from breaking stuff? Does MS publish a list of newly supported functions somewhere? Would the complete header be obsolete for VS2015? You can just PR that and it should get merged - Timo is usually pretty responsive.

joto:
The patch upstream should be easy: Just #undef those #defines again at the end. That should work in all versions. And it would be proper etiquette anway not to have stray macros defined outside your own header file.

springmeyer:
@TheMarex `_MSC_VER` is `1900` I think for VS 2015.

BergWerkGIS:
I've updated the binary deps package containing the `stxxl` fix and the build worked for me locally.

AppVeyor still running ATM: https://ci.appveyor.com/project/DennisOSRM/osrm-backend/build/1.0.1155

I don't have time for a proper fix (PR for stxxl) now as I'm about to jump on the plane for my vacation, but added my TODOs to https://github.com/Project-OSRM/osrm-backend/issues/1628

BergWerkGIS:
```diff
@@ -25,15 +25,17 @@ inline double log2(double x)
 }
 
 // http://msdn.microsoft.com/en-us/library/2ts7cx93.aspx
 #define snprintf _snprintf
 
+#if defined(_MSC_VER) && (_MSC_VER < 1900)
 // http://msdn.microsoft.com/en-us/library/h80404d3.aspx
 #define strtoll _strtoi64
 
 // http://msdn.microsoft.com/en-us/library/85zk715d.aspx
 #define strtoull _strtoui64
+#endif
 
 #endif // STXXL_MSVC
 
 #endif // !STXXL_MSVC_COMPATIBILITY_HEADER
 // vim: et:ts=4:sw=4

```

TheMarex:
Had do re-do this on `develop` because rebasing failed. Closing here.

PR #1630, opened 2015-08-27
Make map matching more robust

TheMarex:
Main problem with map matching is that the algorithm currently assumes your sample rate is around 5-10s. This is because a few thresholds where hard-coded and optimized for that.

Still TODO is the re-calibration of the `confidence` classifier - most traces will have a very low confidence if the same rate is very low.

- [x] Make trace splitting threshold sample rate dependent
- [x] Make transition pruning sample rate dependent
- [x] Re-measure `matching_beta` for low sample rate (100sec -> ~200)
- [x] Calculate candidate query radius based on `gps_accuracy`
- [x] Filter candidates by compressed edge id
- [x] Fix broken splitting test case


arnekaiser:
There is already a pull request (#1529).

PR #1634, opened 2015-08-30
Enable turn penalties on car profile

danpat:
These values were derived by comparing several thousand real-world driving routes, and comparing ETA estimates from OSRM with the real-world travel times.  Enabling turn penalties narrows the variance between OSRM estimates and real-world values, which implies better edge weighting, and thus, more realistic selection of routes.

For now, this has a right-side driving bias.  Further work is needed to bias properly in left-side-driving locations.

### DIFF
--- a/features/car/ferry.feature
+++ b/features/car/ferry.feature
@@ -41,7 +41,7 @@ Feature: Car - Handle ferry routes
 
         When I route I should get
             | from | to | route       | modes | speed   |
-            | a    | g  | abc,cde,efg | 1,2,1 | 26 km/h |
+            | a    | g  | abc,cde,efg | 1,2,1 | 25 km/h |
             | b    | f  | abc,cde,efg | 1,2,1 | 20 km/h |
             | c    | e  | cde         | 2     | 12 km/h |
             | e    | c  | cde         | 2     | 12 km/h |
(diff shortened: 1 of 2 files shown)


emiltin:
I originally picked that formula (angle*angle*k/turn_bias) for the bike profile, because it seemed a natural way to penalize sharp turns.

However it wasn't based on empirical data. Were you able to use the real-world data to assess the formula as well, or just the parameters?

It's interesting that you've chosen a penalty of 10, while the bike profile uses 60.

danpat:
Yes, these numbers are based on empirical data.  I modified the map-matching plugin to return the travel time for a matched route, which lets us compare real routes to what OSRM would've returned for the same route.

The premise of the function turn function seems reasonably sound, so all I did was tweak values and compare against a cohort of about 6000 trips that I had available.  Without the function in place, ETAs were about 85% of reality (i.e. OSRM goes too fast) with a σ=26%.  Introducing the turn penalty and finding the "best" values reduced σ to about 22% and brought the mean up to about 92%.  An incremental improvement, to be sure, but a step in the right direction.  The data itself shows σ≈10% on similar routes during similar time periods.

We're running some analysis on the data we have right now to see if we can extract turning movements.  If so, we might be able to come up with an improved `turn_function`, but for now, it works OK and does improve results.

TheMarex:
Two tests are failing because the turn penalties increase the overall route duration. Just update the threshold and this should be good to merge. :+1:

Issue #1617, opened 2015-08-20
Pedestrian roads with motorcar=destination are unroutable


planemad:
Roads tagged `highway=pedestrian`+`motorcar=destination` are currently not routable using the car profile. Ideally this should be routable.

http://osrm.at/f5s

The wiki definition of `access=destination` is:
> Only when travelling to this element; e.g., local traffic only.

If any of the routing points are at an element with access=destination, it should definitely route as the user intends to visit that point and not drive by. 

cc @TheMarex

daniel-j-h:
For this we would have to know if it's a route's source or destination. Otherwise we still don't want to route through it.

I had a talk with @danpat about it: it seems to be not possible right now: the profile's callbacks are only triggered on a node, way (and segment for the raster branch) basis.

@TheMarex feel free to churn in here, as you tagged it for the `4.8.0` release. Maybe we're missing something.

danpat:
We discussed something similar here for private areas (i.e. gated communities).  We should allow routing when the destination/source is on one of these restricted areas, but not route through them for others.

This could be achieved with https://github.com/Project-OSRM/osrm-backend/pull/1614  by adding a bit invisible penalty to private->non-private or access/limited-access transitions.

TheMarex:
This ticket is just about making it routable in general - even in the middle of the route. This seldom causes problems as `destination` roads are dead-ends in most cases anyway. It will also get a very low  speed, so it is unlikely to be used as shortcut.

It is possible to do this correctly with CH (for specific travel mode changes), I can dig up the paper for that. The idea is basically to not contract nodes that allow to change the travel mode.

daniel-j-h:
Taking a look at this, I found that in the `car.lua`profile both [`restriction_exception_tags`](https://github.com/Project-OSRM/osrm-backend/blob/f6a90e9b429b4e2fdfa231580ae404d7c0696b14/profiles/car.lua#L13) and [`get_exceptions`](https://github.com/Project-OSRM/osrm-backend/blob/f6a90e9b429b4e2fdfa231580ae404d7c0696b14/profiles/car.lua#L149-L153) are unused (i.e. dead code). @TheMarex any idea to what this was meant for? I will add a destination exception by hand now.

daniel-j-h:
Solved, closing here.

kdano:
"as destination roads are dead-ends in most cases anyway"
That's not true at all. Not where I live.
[...]

PR #1639, opened 2015-09-02
Make pedestrian roads marked as destination routable with car profile.

daniel-j-h:
Discussion: it seems like the `@destination` tests are not executed by default, seeing as they are also marked `@todo`? --- I would love to get some eyes on this before merging.

From what I can tell, [access_tag_whitelist](https://github.com/Project-OSRM/osrm-backend/blob/f6a90e9b429b4e2fdfa231580ae404d7c0696b14/profiles/car.lua#L301) is used only once to determine speed values for the way. I started out adding `destination` to the white list (which works), with the idea that I could add an exception [here](https://github.com/Project-OSRM/osrm-backend/blob/f6a90e9b429b4e2fdfa231580ae404d7c0696b14/profiles/car.lua#L362). But I'm not totally sure yet how the "restriction flag" influences routing?

The solution right now adds speed values for pedestrian destination roads, which seems to make them routable, at least that's what the tests are telling. But I have a strange feeling that this may not be a correct solution..

/cc @TheMarex 

---------------------------------------------------

Check provided tests with:

    cucumber --tags @access -p verify

References:

- https://github.com/Project-OSRM/osrm-backend/issues/1617
- http://wiki.openstreetmap.org/wiki/Tag:highway%3Dpedestrian
- http://wiki.openstreetmap.org/wiki/Key:motorcar
- http://wiki.openstreetmap.org/wiki/Key:access

### DIFF
--- a/profiles/car.lua
+++ b/profiles/car.lua
@@ -4,7 +4,7 @@ local find_access_tag = require("lib/access").find_access_tag
 
 -- Begin of globals
 barrier_whitelist = { ["cattle_grid"] = true, ["border_control"] = true, ["checkpoint"] = true, ["toll_booth"] = true, ["sally_port"] = true, ["gate"] = true, ["lift_gate"] = true, ["no"] = true, ["entrance"] = true }
-access_tag_whitelist = { ["yes"] = true, ["motorcar"] = true, ["motor_vehicle"] = true, ["vehicle"] = true, ["permissive"] = true, ["designated"] = true }
+access_tag_whitelist = { ["yes"] = true, ["motorcar"] = true, ["motor_vehicle"] = true, ["vehicle"] = true, ["permissive"] = true, ["designated"] = true, ["destination"] = true }
 access_tag_blacklist = { ["no"] = true, ["private"] = true, ["agricultural"] = true, ["forestry"] = true, ["emergency"] = true, ["psv"] = true }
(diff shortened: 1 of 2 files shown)


profiles/car.lua
```diff
@@ -297,8 +297,12 @@ function way_function (way, result)
         result.backward_speed = highway_speed
       end
     else
+      -- make an exception for restricted highway=pedestrian with motorcar=destination
+      local motorcar = way:get_value_by_key("motorcar")
+      local pedestrian_road = motorcar and "destination" == motorcar and "pedestrian" == highway
+
       -- Set the avg speed on ways that are marked accessible
-      if access_tag_whitelist[access] then
+      if access_tag_whitelist[access] or pedestrian_road then
```
TheMarex: It is fine to add this to `access_tag_whitelist`.

TheMarex:
The `restricted` flag is not yet used for routing. I think the original idea was to add an indication if the returned routed uses restriced streets.

TheMarex:
As per discussion, we should just add `destination` to the access restriction whitelist. It is the behavior described in the wiki and when it breaks, we should fix it.

TheMarex:
Good to merge. :+1:

daniel-j-h:
Merged, closing here.

PR #1640, opened 2015-09-02
Remove unneeded semicola from profiles.

daniel-j-h:
Nothing fancy, does what it says.


TheMarex:
:+1: looks good to me.

PR #1641, opened 2015-09-02
Don't mix signed and unsigned in comparisons as signed is converted first to unsigned.

daniel-j-h:
Check out this: http://coliru.stacked-crooked.com/a/0df657f3205f295b

This is true:

    -1 > 1u

because the integer literal `-1` is first converted to a large unsigned
value and then compared to the unsigned `1`.

This patch fixes several of those isses in the farthest insertion
algorithm.

`-Wsign-compare` catches those issues.

References:

- http://stackoverflow.com/a/5416498
- C++14 standard

/cc @chaupow please make sure to understand this. Sidenote: there was also an include for `BOOST_ASSERT` missing.


TheMarex:
:+1: good to merge.

Issue #1601, opened 2015-08-18
Sanitize all the things: test builds with dynamic runtime sanitizers


daniel-j-h:
This is a proposal to add instrumented binaries to the CI infrastructure, in order to improve code quality and detect hard to find bugs. The idea is as follows: build instrumented binaries, run tests, check reports.
#### Sanitizers

Here is a quick overview of what we should use:
- [AddressSanitizer](http://clang.llvm.org/docs/AddressSanitizer.html): detects e.g. double frees, buffer overflows, ...
- [ThreadSanitizer](http://clang.llvm.org/docs/ThreadSanitizer.html): detects races, posible deadlocks, ...
- [MemorySanitizer](http://clang.llvm.org/docs/MemorySanitizer.html): detects access to uninitialized memory
- [UndefinedBehaviorSanitizer](http://clang.llvm.org/docs/UsersManual.html#controlling-code-generation): detects undefined shifts, integer overflows, ...

For using all the sanitizer's features we need a recent Clang or GCC compiler (GCC lacks a few features compared to Clang). Also note that not all sanitizer combinations are allowed or possible, that is we should build separate instrumented binaries for each sanitizer.
#### Why not Valgrind

Valgrind only provides a subset of what the sanitizers listed above can detect. Running binaries under valgrind is also orders of magnitude slower than using the sanitizers and may not be feasible at all under the CI environemnt constraints.

---

Please at least skim the slides to the sanitizers presentation linked below.

References:
- Overview talk: [video](https://channel9.msdn.com/Events/CPP/C-PP-Con-2014/Sanitize-your-CPP-code), [slides](https://github.com/CppCon/CppCon2014/raw/7b15ec44ac01de0ff3e65a7194b84aca2d4e2366/Presentations/Sanitize%20your%20C%2B%2B%20code/Sanitize%20your%20C%2B%2B%20code%20-%20Kostya%20Serebryany%20-%20CppCon%202014.pdf)
- [-fsanitize in latest Clang](http://clang.llvm.org/docs/UsersManual.html#controlling-code-generation)
- [-fsanitize in latest GCC](https://gcc.gnu.org/onlinedocs/gcc/Debugging-Options.html)

daniel-j-h:
Playground for Travis setup is here: https://github.com/daniel-j-h/cpp-travis-setup (warning: quite noisy).

This is work in progress, but the following are major accomplishments for enabling this ticket's goals:

- container based architecture, faster startup, install, builds
[...]

PR #1642, opened 2015-09-02
[not ready] Use Boost where it provides low overhead features

daniel-j-h:
This Pull Request tracks the progress. Do not merge yet.

Detailed descriptions are in the commit messages, please make sure to read them for discussion.

The basic idea is to rip out hand-written features that have a more or less 1:1 feature parity with plain Boost. From a quick glance sometimes our hand-written code provides less (which can make sense to speed up compilation). The main problem is they are not as thoroughly tested and do not have the level of assertions that Boost provides.

We should make sure to include only what we need, e.g. `<boost/range/irange.hpp>` instead of `<boost/range.hpp>` and keep an eye on compilation times.

This means also removing features we use from Boost which we can use from C++11 / C++14, while testing for portability issues.


daniel-j-h:
After half an hour travis complained about the `irange` switch at https://github.com/Project-OSRM/osrm-backend/commit/f3d6b8ae98f844cc31648c4f2e1a0b5fe0c0083d. Seems like the cucumber tests are failing in large quantity. From a quick glance over the commit, I'm unsure why. Have to investigate this.

daniel-j-h:
Snippet comparing both versions: http://coliru.stacked-crooked.com/a/4a9317c65dcf00b8

daniel-j-h:
The only difference I can come up with is the following edge case:

    irange(x, y);

where `x > y`: http://coliru.stacked-crooked.com/a/4eba09b985445e5e

In this case, Boost's `irange` (rightfully) triggers an assert while our implementation directly jumps into its `operator!=` where it returns `false` instead, signalling the end iterator.

This should be checked at call site and not in `irange`. There might be issues in the code using it, since it silently returns an empty range in those edge cases.

/cc @TheMarex if you have the time to take a look (not important)

daniel-j-h:
Update to the irange issue:

```c++
auto r = boost::irange<unsigned int>(0, 15);
std::cout << r.size() << std::endl;
```

results in `-4294967281` on my Boost 1.55, on recent Boost versions this seems to be fixed:

http://coliru.stacked-crooked.com/a/8d0093d1191d022f

The problem seems to be the `integer_range`s `difference_type`, which is `long int`.

TheMarex:
Ah yeah I remember that problem now. Ran into that a few times, I guess that is why we switched.

daniel-j-h:
@TheMarex Gnaa could have saved me a couple of hours of pain and suffering there :P reverting, but then also adding some asserts on negative sized ranges and so on.

danpat:
How often do we run the test suite through a debug build that will actually have asserts enabled?  If we wanted to be thorough, we'd have Travis do both a Release and Debug build, and test both and publish the Release build if both worked.

daniel-j-h:
@danpat this is something I'd like to do in combination with https://github.com/Project-OSRM/osrm-backend/issues/1601.
That's additional high load for the CI though.

daniel-j-h:
@TheMarex okay if I merge this? Want to get the queue of pull requests down.

One issue that is scoped to this branch but tracked in https://github.com/Project-OSRM/osrm-backend/issues/1692#issuecomment-142971706: if I get the okay to switch in `Boost.Variant`, I could rip out `third_party/variant` completely. If that's too much for now, I'm totally fine with merging this branch as-is.

TheMarex:
@daniel-j-h Just leave `mapbox/variant`. We need to investigate more if removing it makes sense. The variant type is part of the hot code paths and as such requires an in-depth look before we can do significant changes.

daniel-j-h:
> Otherwise this is good to merge after a rebase.

Meh, this wasn't good after a merge. At least some build configurations seem to fail, indicating undefined behavior or worse. Investigating right now in hotfix https://github.com/Project-OSRM/osrm-backend/pull/1712. I should have cought it earlier :/

Issue #1643, opened 2015-09-03
Add block barriers to foot whitelist


planemad:
`barrier=block` are meant to stop larger vehicles according to the [wiki](http://wiki.openstreetmap.org/wiki/Tag:barrier%3Dblock).

This is a possible candidate for the barrier whitelist in the foot and bicycle profiles. We're seeing Strava cycling/walking activity on roads with blocks. 

Example location:https://www.openstreetmap.org/node/959123752

TheMarex:
Something for the `foot` whitelist at least. Not sure about bikes.

emiltin:
i would also whitelist for bicycles. my guess is that barrier=block is used mainly when it's blocks with space between, rather that a low wall. a bicycle doesn't need much space to pass.

daniel-j-h:
I solved this in https://github.com/Project-OSRM/osrm-backend/pull/1651 --- merged, closing here.

PR #1651, opened 2015-09-07
Make an exception for block barriers in bicycle and foot profile.

daniel-j-h:
This adds `barrier=block` exceptions to the respective white lists.

In addition this adds tests to check for the exception in bicycle and
foot profiles and makes sure cars are still not able to cross them.

Checked with:

    cucumber --tags @barrier -p verify

References:

- https://github.com/Project-OSRM/osrm-backend/issues/1643
- http://wiki.openstreetmap.org/wiki/Tag:barrier%3Dblock


TheMarex:
:+1: Good to merge.

PR #1659, opened 2015-09-08
Don't pass by const-value for a read-only view.

daniel-j-h:
I can't see a reason why we pass by const-value here.

Note: strictly speaking this changes the API because of the `route_parameters` publicly facing header.


TheMarex:
Good catch :+1:

daniel-j-h:
> Note: strictly speaking this changes the API because of the route_parameters publicly facing header.

@TheMarex  anything on that, otherwise I'll merge it.

TheMarex:
It's not a breaking api change so this is fine.

PR #1661, opened 2015-09-09
[not ready] Refactor utilities and respective call sites

daniel-j-h:
This Pull Request tracks the progress. Do not merge yet.

The high level goal is to improve the utilities in regard to warnings, build times, compiler and linker specific settings and features.

Please read every commit's description, as they have detailed explanations attached. This pull request just tracks the overall progress.


daniel-j-h:
Please read the commit messages, I explain the rationale in detail there.

Had a good laugh here: [before](https://github.com/Project-OSRM/osrm-backend/blob/develop/server/data_structures/internal_datafacade.hpp#L261-L344) and [after](https://github.com/Project-OSRM/osrm-backend/blob/87153ff1b49a74f1e08c2f2a2643aa1bdefa7d48/server/data_structures/internal_datafacade.hpp#L260-L293).

server/data_structures/internal_datafacade.hpp
```diff
@@ -258,90 +257,40 @@ template <class EdgeDataT> class InternalDataFacade final : public BaseDataFacad
...
+
         SimpleLogger().Write() << "loading geometries";
-        AssertPathExists(geometries_path);
-        LoadGeometries(geometries_path);
+        LoadGeometries(file_for("geometries"));
+
         SimpleLogger().Write() << "loading r-tree";
-        AssertPathExists(ram_index_path);
-        AssertPathExists(file_index_path);
+        // XXX(daniel-j-h): it says this ^ but doesn't load the rtree here
```
daniel-j-h: /cc @TheMarex ^

daniel-j-h:
The [last commit](https://github.com/Project-OSRM/osrm-backend/commit/b086323e0eb09454ce904f4b967ea8acc15ad8e1) rips out all hand-written conversion code. Please read the commit message for details.

Right now there is at least one cucumber test that is failing, because of the `to_string_with_precision` function template. To me it seems like [in the old code here](https://github.com/Project-OSRM/osrm-backend/blob/bed0598530540c1b2459bb271727511ebedaa80e/util/cast.hpp#L159-L163) we're doing a `trim_right(output, ".0")`, which I do not yet fully understand. It would make sense to me if wo do not want trailing zeros, but this is removing only a single trailing zero. What's the reason for this?

Here is a small self contained test case:
http://coliru.stacked-crooked.com/a/6f3ee9d79b1f81d0

And I/O manipulation flag reference:
http://en.cppreference.com/w/cpp/io/ios_base/fmtflags


Except for this small issue it Just Works (tm) --- after some hours :P.

server/data_structures/internal_datafacade.hpp
TheMarex: The `StaticRTree` is a thread specific pointer that is initialized on demand the first time it is used. It needs to be thread specific because it contains a stream object which is not thread safe. Which make me think we should maybe explore `mmap`, since virtual memory accesses is thread safe.

TheMarex:
@daniel-j-h I think the problem is that JSON (and javascript) in general does not have integers. So our internal JSON representation will cast everything to double values for the number type. To emphasis in a response that this value is _not_ meant to be a floating point value we remove the `.0` (e.g. for array indices).

server/data_structures/internal_datafacade.hpp
daniel-j-h: > It needs to be thread specific because it contains a stream object which is not thread safe.

Oh, if it's just for the stream's thread safety, there's Herb Sutter's [`monitor<T>`](https://gist.github.com/daniel-j-h/a3c3cf172b00f186b85b#file-monitor-hpp) and [`concurrent<T>`](https://gist.github.com/daniel-j-h/a3c3cf172b00f186b85b#file-concurrent-hpp) (blocking and non-blocking, respectively --- this is exactly what you get with Clojure's [Refs](http://clojure.org/Refs) and [Agents](http://clojure.org/agents)), that make it very easy:

    monitor<std::ostream&> out{std::cout};

Boom, global thread safe everything-synchronization that blocks (stream in this case). Usage:

    out([](auto& stream) { stream << "Hello"; });

Change `monitor` with `concurrent`, boom global thread safe everything-synchronization that does not block. Call site stays the same.

Check out his talk [here](https://channel9.msdn.com/Shows/Going+Deep/C-and-Beyond-2012-Herb-Sutter-Concurrency-and-Parallelism) if you have some time. High quality content.

TheMarex:
@daniel-j-h the only blocker here is the added `.0`? If so, just try to emulate that behavior: it is important for indices.

PR #1663, opened 2015-09-09
[not ready] Parallelize more: nested parallelism, rely on task scheduler

daniel-j-h:
This Pull Request tracks the progress. Do not merge yet.

The high level goal is to parallelize more using Intel TBB's scheduler for:

- nested parallelism
- not oversubscribing

Please read every commit's description, as they have detailed explanations attached. This pull request just tracks the overall progress.

[Here's](https://github.com/Project-OSRM/osrm-backend/commit/6b2bf4954aa87402e3d64c0278b26c4a61e8e667) additional explanation on this, as in the following commit messages.

PR #1664, opened 2015-09-09
Ownership: vector already owns, no need for wrapping in unique_ptr.

daniel-j-h:
Removes the pointless `std::unique_ptr<std::vector<T>>` usage,
as a `std::vector` already owns its resources and manages them.

Results in one indirection less (hint: good).

Sidenote: I also don't get things like [this here](https://github.com/Project-OSRM/osrm-backend/blob/db092c828e09e55570e5f1ac15dd63a8735f2012/contractor/processing_chain.cpp#L207-L209). This is introducing an additional move, even though taking by rvalue reference `&&` means we have it exclusively for us. If at all the function should accept the argument by-value and then the call site can decide ito transfer ownership with a move or to create a copy.

(And sorry for the clang-format noise)

### DIFF
--- a/contractor/processing_chain.cpp
+++ b/contractor/processing_chain.cpp
@@ -230,7 +230,7 @@ std::size_t Prepare::WriteContractedGraph(unsigned max_node_id,
     const unsigned crc32_value = CalculateEdgeChecksum(node_based_edge_list);
 
     // Sorting contracted edges in a way that the static query graph can read some in in-place.
-    tbb::parallel_sort(contracted_edge_list);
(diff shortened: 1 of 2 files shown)


daniel-j-h:
Seems like Travis CI is failing since it's using a old `libtbb-dev` that does not have the [Container overload](https://www.threadingbuildingblocks.org/docs/help/index.htm#reference/algorithms/parallel_sort_func.htm) for `parallel_sort`, but only the iterator taking one.

This should also affect [this pull request](https://github.com/Project-OSRM/osrm-backend/pull/1663).

PR #1666, opened 2015-09-10
Adds basic Doxygen support

danpat:
Run:

    mkdir build
    cd build
    cmake ..
    make doc

and output will be in

    build/html


daniel-j-h:
Now with callgraphs (required `dot`) and more verbose, intended to be used by us developers (e.g. private member functions).

danpat:
Should add `find_package(dot)` to CMakeLists.txt if we're going require DOT for doxygen.

daniel-j-h:
Thanks @danpat, there's indeed an easy way to [check for Dot](http://www.cmake.org/cmake/help/v3.0/module/FindDoxygen.html).

Preview:

Inheritance diagrams:

inheritance

Callgraphs:

callgraph

TheMarex:
@daniel-j-h the docs seem to include our unit tests. See the entries `RandomGraphFixture` etc.

daniel-j-h:
Yeah we definitely have add some more exclude patterns, it's also trying to parse the `README.md`.

daniel-j-h:
All done here!

Latest changes exclude unwanted files / directories, and we no longer have a hard dependency on `dot` for generating docs (availability of `dot` enables graphs, though).

TheMarex:
@daniel-j-h So the option of just generating `.md` files did not work out? Otherwise this looks good to merge.

daniel-j-h:
@TheMarex what do you mean with generating `.md` files? I did the following:

- use doxygen to extract comments and associated symbols as xml
- use sphinx with breathe plugin to generate beautiful looking docs

the idea was to then let readthedocs build the documentation automatically on every commit (similarly to how Travis does it), resulting in good looking Python-ish docs.

Unfortunately there are two main issues blocking this:

- readthedocs does not support doxygen in the pipeline, there is a wontfix ticket on this
- no autodoc support, we would have to explicitly activate docs generation for every symbol

what I then did was to improve the doxygen setup as good as possible, while keeping the option open to change the layout: we are able to support custom html, css and js files after all.

I'm currently talking to @amyleew about this. Maybe we can create a developer-friendly and beautiful layout once and then re-use it also in projects like libosmium.

amyleew:
Per chat with @daniel-j-h, mocked up simple layout with Project OSRM customization @TheMarex. I should be able to touch on this in the next few days. Should take too long, should I cut a new ticket in this /repo for this?

image

daniel-j-h:
@amyleew it would be interesting for us is if we are able to not only style the doxygen output, but to make it more easy to navigate and work with it --- feel free to create a ticket for this, in which we can discuss specifics.

Issue #1606, opened 2015-08-18
Strange routing : any ideas ?


the-knife:
Hello,

Any ideas why this way seems to be avoided in [this itinerary](http://osrm.at/f34) ?
Coudn't find anything strange in the openstreetmap data.

Thanks in advance!
the_knife

danpat:
There is a turn restriction in OSM:

<img width="725" alt="screen shot 2015-08-18 at 12 14 26 pm" src="https://cloud.githubusercontent.com/assets/1892250/9340099/bfb5c80c-45a2-11e5-87b7-d557af7b1200.png">

Looking on Google Streetview, it appears that there is a turn lane and I can't see a sign restricting turns here.  It looks like an error in OSM data, you should correct that if it's wrong.

the-knife:
Thanks for searching.

The relation https://www.openstreetmap.org/relation/3724506
indicates that bicycles are able to turn to right even if the traffic signal is red.

As the "no_right_turn" tag is not present, osrm should not prevent to turn, should it?

danpat:
Ah, thanks for double-checking.

Yes, looks like a bug in the restrictions-parsing code.  If we find any relation with a key that begins with "restriction", will will add a restriction for it, and we only look at the "from" "via" and "to" attributes, the actual value of the `<tag k='restriction' v='...'/>` tag is only looked at to see if it begins with `only`.

In short, restriction parsing is broken here, this is a bug.

To fix this however, we would need to move restriction processing and parsing into Lua so that it could be fully customized for bike/car/walk profiles.

Updating title and marking this as a bug.

@the-knife The fix needs to go into `extractor/restrictions_parser.cpp` if you want to have a go at it :-)

MoKob:
fixed

PR #1668, opened 2015-09-10
Treat `restriction:*` tags specifically based on profiles

lbud:
Fixes https://github.com/Project-OSRM/osrm-backend/issues/1606.

This treats any keys in [`restriction_exception_tags`](https://github.com/Project-OSRM/osrm-backend/blob/master/profiles/car.lua#L14) as candidates for specific restriction keys (so the only `restriction:*` tags applied in the car profile would be `restriction:motorcar`, `restriction:motor_vehicle`, `restriction:vehicle`).

@daniel-j-h or @TheMarex would love a second set of eyes on the way I approached this. Also, I think I changed the line endings on lines I added in restriction_parser.cpp — what should I do about that?

### DIFF
--- a/extractor/restriction_parser.cpp
+++ b/extractor/restriction_parser.cpp
@@ -35,6 +35,7 @@ SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
 #include <boost/algorithm/string.hpp>
 #include <boost/algorithm/string/regex.hpp>
+#include <boost/algorithm/string/predicate.hpp>
 #include <boost/ref.hpp>
 #include <boost/regex.hpp>
 
@@ -51,8 +52,7 @@ int lua_error_callback(lua_State *lua_state)
 }
 }
 
-RestrictionParser::RestrictionParser(lua_State *lua_state)
-    : use_turn_restrictions(true)
+RestrictionParser::RestrictionParser(lua_State *lua_state) : use_turn_restrictions(true)
 {
     ReadUseRestrictionsSetting(lua_state);
 
@@ -141,16 +141,30 @@ RestrictionParser::TryParse(const osmium::Relation &relation) const
 
     bool is_only_restriction = false;
 
-    for (auto iter = fi_begin; iter != fi_end; ++iter)
+    for (; fi_begin != fi_end; ++fi_begin)
     {
-        if (std::string("restriction") == iter->key() ||
-            std::string("restriction::hgv") == iter->key())
-        {
-            const std::string restriction_value(iter->value());
+        const std::string key(fi_begin->key());
+        const std::string value(fi_begin->value());
 
-            if (restriction_value.find("only_") == 0)
+        if (value.find("only_") == 0)
+        {
+            is_only_restriction = true;
+        }
+
+        // if the "restriction*" key is longer than 11 chars, it is a conditional exception (i.e.
+        // "restriction:<transportation_type>")
+        if (key.size() > 11)
+        {
+            const auto ex_suffix = [&](const std::string &exception)
             {
-                is_only_restriction = true;
+                return boost::algorithm::ends_with(key, exception);
+            };
+            bool is_actually_restricted =
+                std::any_of(begin(restriction_exceptions), end(restriction_exceptions), ex_suffix);
+
+            if (!is_actually_restricted)
+            {
+                return mapbox::util::optional<InputRestrictionContainer>();
             }
         }
     }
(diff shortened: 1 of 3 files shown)


extractor/restriction_parser.cpp
```diff
@@ -140,11 +141,11 @@ RestrictionParser::TryParse(const osmium::Relation &relation) const
     }
 
     bool is_only_restriction = false;
+    const char RESTRICTION_PREFIX[] = {"restriction"};
 
     for (auto iter = fi_begin; iter != fi_end; ++iter)
```
daniel-j-h: You don't have to create a copy here, just use `for (; fi_begin != fi_end; ++fi_begin)`.

extractor/restriction_parser.cpp
```diff
@@ -140,11 +141,11 @@ RestrictionParser::TryParse(const osmium::Relation &relation) const
     }
 
     bool is_only_restriction = false;
+    const char RESTRICTION_PREFIX[] = {"restriction"};
 
     for (auto iter = fi_begin; iter != fi_end; ++iter)
     {
-        if (std::string("restriction") == iter->key() ||
-            std::string("restriction::hgv") == iter->key())
+        if (boost::algorithm::starts_with(iter->key(), RESTRICTION_PREFIX))
```
daniel-j-h: I _think_ this check amd the `RESTRICTION_PREFIX` is redundant as the `KeyPrefixFilter` above already filters out everything that starts with `restriction`.

extractor/restriction_parser.cpp
```diff
@@ -140,11 +141,11 @@ RestrictionParser::TryParse(const osmium::Relation &relation) const
...
     bool is_only_restriction = false;
+    const char RESTRICTION_PREFIX[] = {"restriction"};
 
     for (auto iter = fi_begin; iter != fi_end; ++iter)
     {
-        if (std::string("restriction") == iter->key() ||
-            std::string("restriction::hgv") == iter->key())
+        if (boost::algorithm::starts_with(iter->key(), RESTRICTION_PREFIX))
         {
             const std::string restriction_value(iter->value());
```
daniel-j-h: No need for a copy here, `const auto&` should work fine I guess.

extractor/restriction_parser.cpp
```diff
@@ -152,6 +153,21 @@ RestrictionParser::TryParse(const osmium::Relation &relation) const
             {
                 is_only_restriction = true;
             }
+            if (strlen(iter->key()) > 11)
```
daniel-j-h: Why `11` here? Because of the prefix length above? I also would just use string's `.size()` member function.

Hmm are both `key()` and `value()` return values C-strings and not `std::string`? Then I would create a `std::string` for both of them at the beginning an just use these.

extractor/restriction_parser.cpp
```diff
@@ -152,6 +153,21 @@ RestrictionParser::TryParse(const osmium::Relation &relation) const
...
                 is_only_restriction = true;
             }
+            if (strlen(iter->key()) > 11)
+            {
+                bool is_actually_restricted = false;
+                for (const auto &exception : restriction_exceptions)
+                {
+                    if (boost::algorithm::ends_with(iter->key(), exception))
+                    {
+                        is_actually_restricted = true;
```
daniel-j-h: Can you return early here? This looks like

```
const auto ex_suffix = [&](const auto& key){ return boost::algorithm::ends_with(key, exception)) };
is actually_restricted = std::any_of(begin(restriction_exceptions), end(restrictions_exceptions), ex_suffix);

```

extractor/restriction_parser.cpp
lbud: Yep, `11` because `restriction` is length 11, so if it's longer than 11 it's a [conditional restriction](http://wiki.openstreetmap.org/wiki/Conditional_restrictions#Transportation_mode) (`restriction:<transportation_mode>`)

danpat: For readability sake, maybe replace `11` with `const std::string("restriction").size()`.  Throwing the `const` in there should avoid any runtime overhead, and it shows the intent a bit more clearly.

Alternatively, throw a comment in there that describes why 11, like `// If key is > 11, then it will be a conditional restriction`

lbud:
Thanks for the comments @daniel-j-h! Fixed everything in https://github.com/Project-OSRM/osrm-backend/commit/b562ad1c7cc8da7bac84ea594a434fce0beacb46.

daniel-j-h:
Perfect! :+1:

TheMarex:
Awesome @lbud :+1: needs a rebase on `develop` and then it is good to merge.

lbud:
@TheMarex :+1: done

TheMarex:
Merged.

PR #1677, opened 2015-09-11
Integration scripts for Clang's Modernize and Tidy tool.

daniel-j-h:
New directory: `scripts/`, in which small scripts for developers reside.

- `modernize`: runs all cpp files through `clang-modernize`, respecting
  out targeted compiler versions, applying C++11 transformations, doing
  syntax checks and formatting --- in parallel.

- `tidy`: runs all cpp files through `clang-tidy`, with selected
  warnings only, since we do not want to warn on every small detail.

Please check the talk slides for `clang-tidy` linked in the references!

Note: there might be a static analyzer script coming! :)

References:

- http://clang.llvm.org/extra/clang-tidy/
- http://llvm.org/devmtg/2014-04/PDFs/Talks/clang-tidy%20LLVM%20Euro%202014.pdf
- http://clang.llvm.org/extra/clang-tidy/checks/list.html
- https://github.com/Project-OSRM/osrm-backend/pull/1603

### DIFF
--- a/scripts/modernize.sh
+++ b/scripts/modernize.sh
@@ -1,0 +1,6 @@
+#!/usr/bin/env bash
+
+# Runs the Clang Modernizer in parallel on the code base.
+# Requires a compilation database in the build directory.
+
+git ls-files '*.cpp' | xargs -I{} -P $(nproc) clang-modernize -p build -final-syntax-check -format -style=file -summary -for-compilers=clang-3.4,gcc-4.8 -include . -exclude third_party {}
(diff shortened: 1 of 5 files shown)


daniel-j-h:
Static Analyzer Report:

report

scripts/modernize.sh
```diff
@@ -0,0 +1,6 @@
+#!/usr/bin/env bash
+
+# Runs the Clang Modernizer in parallel on the code base.
+# Requires a compilation database in the build directory.
+
+git ls-files '*.cpp' | xargs -I{} -P $(nproc) clang-modernize -p build -final-syntax-check -format -style=file -summary -for-compilers=clang-3.4,gcc-4.8 -include . -exclude third_party {}
```
TheMarex: Does this also work for headers?

daniel-j-h: The reason we have to pipe the `.cpp` file into it is that the compilation database generated by CMake only contains the compiler commands for a single TU. That is, you have commands in there like `g++ main.cc -o main`, which only match the `.cpp` files, since the headers are included by the preprocessor.

That said, the tool *does* work on headers we just don't have to specify them, check the initial run here:
https://github.com/Project-OSRM/osrm-backend/pull/1603/files

TheMarex:
This is good to merge. :+1: Can you ticket the non-false positives?

daniel-j-h:
Yeah, not sure how to do a proper integration, though. Would love to have a Travis build running for the modernizer at least, exiting with an error if it finds something. The tidy and analyzer tools are too noisy for continuous integration, probably good to do this once every other month.

PR #1684, opened 2015-09-14
Script for fully automated test bisecting.

daniel-j-h:
Automate cucumber tests bisecting by providing a `git bisect` script.

Because it is stored in source control, but bisecting changes the HEAD,
it is advised to first copy over the script to a place outside source
control, e.g. `/tmp`.

Usage:

    git bisect start HEAD HEAD~10
    git bisect run /tmp/bisect_cucumber.sh

    # wait till it gives you the fault-introducing commit

    git bisect reset

This automatically configures and builds OSRM, spawns the cucumber tests
and communicates with `git bisect` based on its return code.

Note: uses the same `scripts` directory as proposed in https://github.com/Project-OSRM/osrm-backend/pull/1677, but does not depend on pull request.

Reference:

- man git-bisect

### DIFF
--- a/scripts/bisect_cucumber.sh
+++ b/scripts/bisect_cucumber.sh
@@ -1,0 +1,27 @@
+#!/usr/bin/env bash
+
+# Automates bisecting cucumber tests in a portable way; usage:
+#
+#     git bisect start GOODGITSHA BADGITSHA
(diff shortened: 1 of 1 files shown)


TheMarex:
Good to merge. :+1:

PR #1687, opened 2015-09-14
Caches iterators instead of invoking function calls on every iteration.

daniel-j-h:
This caches iterators, i.e. especially the end iterator when possible.

The problem:

    for (auto it = begin(seq); it != end(seq); ++it)

this has to call `end(seq)` on every iteration, since the compiler is
not able to reason about the call's site effects (too bad, huh).

Instead do it like this:

    for (auto it = begin(seq), end = end(seq); it != end; ++it)

caching the end iterator.

Of course, still better would be:

    for (auto&& each : seq)

if all you want is to capture the value (not the iterator).

And still better: use algorithms that express intend, like `std::count_if`, `std::accumulate`, and so on. Hand-written loops are the root of all evil!

Why `auto&&` you may ask? Because it binds to everything and never copies!

Skim the referenced proposal (that was rejected, but nevertheless) for a
detailed explanation on range-based for loops and why `auto&&` is great.

Reference:

- http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2014/n3853.htm

### DIFF
--- a/contractor/contractor.hpp
+++ b/contractor/contractor.hpp
@@ -312,7 +312,7 @@ class Contractor
         tbb::parallel_for(tbb::blocked_range<int>(0, number_of_nodes, InitGrainSize),
                           [&remaining_nodes](const tbb::blocked_range<int> &range)
                           {
-                              for (int x = range.begin(); x != range.end(); ++x)
+                              for (int x = range.begin(), end = range.end(); x != end; ++x)
                               {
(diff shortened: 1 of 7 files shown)


TheMarex:
:+1:

daniel-j-h:
Did a rebase followed by a fast-forward merge. Done!

PR #1691, opened 2015-09-16
Include (road) name of matched nodes in addition to coordinate.

danpat:
This came up as part of customer project.  They wanted to know the names of the roads the matched points were on.  We discussed using the matched-points to call `/nearest` which returns the name, but even though we know the matched coordinate is on a real geometry, we can't guarantee that `/nearest` will select the correct edge direction.

This PR is to discuss the approach, and whether we want to include this feature upstream.


TheMarex:
We can merge for now. Will likely be replaced by the `5.0.0` API rework. :+1: if rebased on `develop` and merged by hand.

PR #1695, opened 2015-09-18
Test that alternative route exists

freenerd:
Complement to a6b44a14709bf760425cfef288d1f14a6914f569

Also I really got to like the cucumber tests @emiltin :smile:

### DIFF
--- a/features/step_definitions/routing.rb
+++ b/features/step_definitions/routing.rb
@@ -82,6 +82,10 @@ When /^I route I should get$/ do |table|
         end
         if table.headers.include? 'route'
           got['route'] = (instructions || '').strip
+          if table.headers.include?('alternative')
+            raise "*** No alternative found ***" unless json['found_alternative']
+            got['alternative'] = way_list json['alternative_instructions'].first
+          end
           if table.headers.include?('distance')
             if row['distance']!=''
               raise "*** Distance must be specied in meters. (ex: 250m)" unless row['distance'] =~ /\d+m/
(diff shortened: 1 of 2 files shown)


features/step_definitions/routing.rb
```diff
@@ -82,6 +82,10 @@
         end
         if table.headers.include? 'route'
           got['route'] = (instructions || '').strip
+          if table.headers.include?('alternative')
+            raise "*** No alternative found ***" unless json['found_alternative']
```
emiltin: setting got['alternative'] to nil is probably better than raising?

emiltin:
:-)

emiltin:
maybe move the test for alt routes to a new scenario file like testbot/alternative.scenario? i also recommend tagging the scenario with something like ```@alternative```, like other scenario files.

features/testbot/via.feature
```diff
@@ -52,6 +52,31 @@ Feature: Via points
...
+            | hi    |
+            | ij    |
+            | jz    |
+
+        And the query options
+            | alt | true |
+
+        When I route I should get
+            | waypoints | route       | alternative    |
+            | a,z       | ab,bc,cd,dz | ag,gh,hi,ij,jz |
```
emiltin: since you have no via points, i would use 'from' and 'to' columns instead of 'waypoints'. it's easier to read and is used everywhere else when there are no via points. note that this scenario file is meant for testing via points specifically.

PR #1701, opened 2015-09-22
Refactor alternative route test

freenerd:
Picking up comments by @emiltin from https://github.com/Project-OSRM/osrm-backend/pull/1695

I've added another test for the alt flag set to false. @emiltin other tests you can think of? @TheMarex are there more cases where we want to test for that an alternative exists?

### DIFF
--- a/features/step_definitions/routing.rb
+++ b/features/step_definitions/routing.rb
@@ -83,8 +83,12 @@ When /^I route I should get$/ do |table|
         if table.headers.include? 'route'
           got['route'] = (instructions || '').strip
           if table.headers.include?('alternative')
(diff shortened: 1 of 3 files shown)


TheMarex:
@freenerd you made this PR against `master` not `develop`.

For more advanced tests we would need to have a more deterministic alternative path algorithm. This is maybe something we want to investigate, but not really high priority.

freenerd:
@TheMarex ah sorry, re-opening against develop. can you set the "base branch" to `develop` at https://github.com/Project-OSRM/osrm-backend/settings/branches ?

emiltin:
maybe a test that shows when osrm delibaretly does not provide an alternative?

PR #1702, opened 2015-09-22
Refactor alternative route test

freenerd:
Re-rolling https://github.com/Project-OSRM/osrm-backend/pull/1701 because was PRed against wrong branch

Picking up comments by @emiltin from https://github.com/Project-OSRM/osrm-backend/pull/1695

I've added another test for the alt flag set to false. @emiltin other tests you can think of? @TheMarex are there more cases where we want to test for that an alternative exists?


TheMarex:
Good to merge. :+1: Please merge by hand (ff-merge).

PR #1707, opened 2015-09-24
Bugfix: allow POST request without POST data

arnekaiser:
Since commit a87d89302fb412f4b64a1e77a91fa8cecc70fbfd OSRM returns errors like

*{"status_message":"Query string malformed close to position 6","status":400}*

if you send a POST request without POST data. For example: 

*curl -X POST http://OSRM_SERVER/hello*

With this change the question mark will only be added if there is any POST data.


TheMarex:
Thanks! :+1:

PR #1708, opened 2015-09-24
[not ready] Enable the usage of traffic data for edge weights

TheMarex:
@danpat is currently heading the development here. This is currently very much in the prototype stage. The idea is as follows:

- Get traffic identifiers for each edge from lua and save an internal mapping to the graph
- All operations that don't need to be done after the contraction are moved to `osrm-extract`
- You can use the mapping to directly overwrite edge weights for edges
- Hence by running direct edge weight updates and `osrm-prepare` in a loop you can get down to real-time-ish processing times
- The newly added partial contraction parameter can be used to tweak this

Problems:
- Geometry compression is a lot less effective, since one OSM way might be split into several traffic identifiers (which cannot be merged)
- Traffic ids are pulled into the server and increase the memory usage
- The added overhead of using a lookup table might make pre-processing slower

TODO:
- [ ] squash a lot of the partial commits
- [ ] rethink our strategy to pulling traffic identifiers in the server: might be too much memory overhead
- [ ] investigate the memory overhead by using common traffic identifiers (TMC)


TheMarex:
Capturing from voice call:

- `traffic_id` is getting pulled from the server-side code. encoding it as name is fine
- traffic data ingestion will happen as follows:
  - `osrm-extract` will write out a mapping: `node_osm_id -> internal_id` 
  - `osrm-extract` will create the (compressed) edge expanded graph
  - `osrm-extract` will write out a mapping: `edge_based_edge_id -> (segment_start_node_id, segment_end_node_id, length)`. Note that an edge based edge will cover multiple of such segments.
  - `osrm-extract` will write out the penalties added for each edge based edge
  - `osrm-update` will be able to modify the weights of the edge-based graph [optional step]
  - `osrm-prepare` will read edge based graph from file and do the contraction.

PR #1710, opened 2015-09-24
Feature/bearing selection

danpat:
This adds a new `b=` parameter to `match` and `viaroute` requests which restricts Nearest Neighbor selection when finding phantom nodes.

If bearing hints are available to the user (i.e. from vehicle GPS traces), this can help get better snapping by not selecting road lanes going in the opposite direction to what was actually being travelled.

The parameter is optional, and if omitted, the current behaviour remains unchanged.

Like the `timestamp` parameter, if present, one `b` parameter is required for each coordinate pair.

The filter limits are currently hard-coded, but could probably be made configurable.  The intent at this stage is mostly to reduce matching edges going opposite to that which is desired, so a fairly broad matching angle is permissible.

The filter is implemented in the StaticRTree, so this could be trivially exposed for `/nearest`, and all other APIs that provide coordinates.

### DIFF
--- a/data_structures/route_parameters.cpp
+++ b/data_structures/route_parameters.cpp
@@ -28,6 +28,7 @@ SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 #include <boost/fusion/container/vector.hpp>
 #include <boost/fusion/sequence/intrinsic.hpp>
 #include <boost/fusion/include/at_c.hpp>
+#include <boost/spirit/include/qi.hpp>
 
 #include <osrm/route_parameters.hpp>
 
@@ -114,6 +115,18 @@ void RouteParameters::addTimestamp(const unsigned timestamp)
     if (!timestamps.empty())
     {
         timestamps.back() = timestamp;
+    }
+}
+
+void RouteParameters::addBearing(const int bearing, boost::spirit::qi::unused_type unused, bool& pass)
+{
+    bearings.resize(coordinates.size());
+    pass = false;
+    if (bearing < 0 || bearing > 359) return;
+    if (!bearings.empty())
+    {
+        bearings.back() = bearing;
+        pass = true;
     }
 }
 
--- a/data_structures/static_rtree.hpp
+++ b/data_structures/static_rtree.hpp
@@ -642,10 +642,55 @@ class StaticRTree
         return result_coordinate.is_valid();
     }
 
+    /**
+     * Checks whether A is between B-range and B+range, all modulo 360
+     * e.g. A = 5, B = 5, range = 10 == true
+     *      A = -6, B = 5, range = 10 == false
+     *      A = -2, B = 355, range = 10 == true
+     *      A = 6, B = 355, range = 10 == false
+     *      A = 355, B = -2, range = 10 == true
+     *
+     * @param A the bearing to check, in degrees, 0-359, 0=north
+     * @param B the bearing to check against, in degrees, 0-359, 0=north
+     * @param range the number of degrees either side of B that A will still match
+     * @return true if B-range <= A <= B+range, modulo 360
+     * */
+    static bool IsBearingWithinBounds(const int A,
+                                      const int B,
+                                      const int range)
+    {
+
+        if (range >= 180) return true;
+        if (range <= 0) return false;
+
+        // Map both bearings into positive modulo 360 space
+        const int normalized_B = (B < 0)?(B % 360)+360:(B % 360);
+        const int normalized_A = (A < 0)?(A % 360)+360:(A % 360);
+
+        if (normalized_B - range < 0)
+        {
+            return (normalized_B - range + 360 <= normalized_A && normalized_A < 360) ||
+                   (0 <= normalized_A && normalized_A <= normalized_B + range);
+        }
+        else if (normalized_B + range > 360)
+        {
+            return (normalized_B - range <= normalized_A && normalized_A < 360) ||
+                   (0 <= normalized_A && normalized_A <= normalized_B + range - 360);
+        }
+        else
+        {
+            return normalized_B - range <= normalized_A && normalized_A <= normalized_B + range;
+        }
+    }
+
+
+
     bool IncrementalFindPhantomNodeForCoordinate(
         const FixedPointCoordinate &input_coordinate,
         std::vector<PhantomNode> &result_phantom_node_vector,
         const unsigned max_number_of_phantom_nodes,
+        const int filter_bearing = 0,
+        const int filter_bearing_range = 180,
         const float max_distance = 1100,
         const unsigned max_checked_elements = 4 * LEAF_NODE_SIZE)
     {
@@ -737,9 +782,34 @@ class StaticRTree
                     m_coordinate_list->at(current_segment.v), input_coordinate,
                     projected_coordinate, foot_point_coordinate_on_segment, current_ratio);
 
+                const float forward_edge_bearing = coordinate_calculation::bearing(
+                                m_coordinate_list->at(current_segment.u),
+                                m_coordinate_list->at(current_segment.v));
+
+                const float backward_edge_bearing = (forward_edge_bearing + 180) > 360 
+                                                      ? (forward_edge_bearing - 180) 
+                                                      : (forward_edge_bearing + 180);
+
+                const bool forward_bearing_valid = IsBearingWithinBounds(forward_edge_bearing, filter_bearing, filter_bearing_range);
+                const bool backward_bearing_valid = IsBearingWithinBounds(backward_edge_bearing, filter_bearing, filter_bearing_range);
+
+                if (!forward_bearing_valid && !backward_bearing_valid)
+                {
+                    continue;
+                }
+
                 // store phantom node in result vector
                 result_phantom_node_vector.emplace_back(current_segment,
                                                         foot_point_coordinate_on_segment);
+
+                if (!forward_bearing_valid)
+                {
+                    result_phantom_node_vector.back().forward_node_id = SPECIAL_NODEID;
+                } 
+                else if (!backward_bearing_valid)
+                {
+                    result_phantom_node_vector.back().reverse_node_id = SPECIAL_NODEID;
+                }
 
                 // Hack to fix rounding errors and wandering via nodes.
                 FixUpRoundingIssue(input_coordinate, result_phantom_node_vector.back());
@@ -788,6 +858,8 @@ class StaticRTree
         const FixedPointCoordinate &input_coordinate,
         std::vector<std::pair<PhantomNode, double>> &result_phantom_node_vector,
         const double max_distance,
+        const int filter_bearing = 0,
+        const int filter_bearing_range = 180,
         const unsigned max_checked_elements = 4 * LEAF_NODE_SIZE)
     {
         unsigned inspected_elements = 0;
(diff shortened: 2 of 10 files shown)


data_structures/route_parameters.cpp
```diff
@@ -117,6 +117,15 @@ void RouteParameters::addTimestamp(const unsigned timestamp)
...
 }
 
+void RouteParameters::addBearing(const int bearing)
+{
+    bearings.resize(coordinates.size());
+    if (!bearings.empty())
+    {
+        bearings.back() = bearing;
+    }
+}
```
daniel-j-h: Let me see if I understand this correctly: this is the handler for the `b=` spirit grammar, called with ints. It implicitly relies on interleaving coordinates and bearings. In the case of multiple bearings such as `b=1&b=2&b=3` only the last `b=3` is stored.

data_structures/static_rtree.hpp
```diff
@@ -642,11 +642,56 @@ class StaticRTree
...
+    /**
+     * Checks whether A is between B-range and B+range, all modulo 360
+     * e.g. A = 5, B = 5, range = 10 == true
+     *      A = -6, B = 5, range = 10 == false
+     *      A = -2, B = 355, range = 10 == true
+     *      A = 6, B = 355, range = 10 == false
+     *      A = 355, B = -2, range = 10 == true
+     *
+     * @param A the bearing to check, in degrees, 0-359, 0=north
+     * @param B the bearing to check against, in degrees, 0-359, 0=north
```
daniel-j-h: The documentation states as precondition for `A` and `B` to be in the range 0-359 (both including I suppose?).
But then you do modulo calculation to bring `A` and `B` in this range in case it is outside of it.

Following that:

- either the precondition is wrong and I am allowed to pass in values outside of the stated range, or
- the precondition is right, then assert this instead of doing modulo calculation

data_structures/static_rtree.hpp
```diff
@@ -642,11 +642,56 @@ class StaticRTree
...
+     * @param range the number of degrees either side of B that A will still match
+     * @return true if B-range <= A <= B+range, modulo 360
+     * */
+    static bool IsBearingWithinBounds(const int A,
+                                      const int B,
+                                      const int range)
```
daniel-j-h: Right now `range` is known at compile-time. Will this ever change?

If not tell the compiler this useful information, such as:

    template <int Range>
    fn() {
        static_assert(range >= 0,"");
        //
    }

Also: if the range is never allowed to be negative, then either assert this everywhere or use an integral unsigned type.

data_structures/static_rtree.hpp
```diff
@@ -642,11 +642,56 @@ class StaticRTree
...
+
+
+
     bool IncrementalFindPhantomNodeForCoordinate(
         const FixedPointCoordinate &input_coordinate,
         std::vector<PhantomNode> &result_phantom_node_vector,
         const unsigned max_number_of_phantom_nodes,
         const float max_distance = 1100,
+        const int filter_bearing = 0,
+        const int filter_bearing_range = 180,
```
daniel-j-h: In case you change the range argument to be a compile-time constant, you have to say `const constexpr int` here.

plugins/match.hpp
```diff
@@ -126,8 +127,12 @@ template <class DataFacadeT> class MapMatchingPlugin : public BasePlugin
             }
 
             std::vector<std::pair<PhantomNode, double>> candidates;
+            // Use bearing values if supplied, otherwise fallback to 0,180 defaults
+            int bearing = input_bearings.size() > 0 ? input_bearings[current_coordinate] : 0;
+            int range = input_bearings.size() > 0 ? 10 : 180;
```
daniel-j-h: Those can be `const`. Also most variables _could_ be `auto` ..

plugins/match.hpp
```diff
@@ -261,12 +266,19 @@ template <class DataFacadeT> class MapMatchingPlugin : public BasePlugin
...
             json_result.values["status"] = "Number of timestamps does not match number of coordinates .";
             return 400;
         }
 
+        if (input_bearings.size() > 0 && input_coords.size() != input_bearings.size())
+        {
+            json_result.values["status"] = "Number of bearings does not match number of coordinates .";
+            return 400;
+        }
+
```
daniel-j-h: I see how this checks at least that the sizes are the same. But what happens if I do not interleave coordinated with bearings, but instead do a request such as:

> coordinate, bearing, coordinate, coordinate, bearing, bearing

then this check won't catch the mistake and from the last two bearings, only the last one will be used and there exists a bearing that is default initialized but never assigned to?

data_structures/route_parameters.cpp
danpat: Correct, I've just mimicked the behaviour of `timestamp`.  In theory we could supply bearings only on some coordinates, but I didn't dig that far into the logic for now.

data_structures/static_rtree.hpp
danpat: Yep.  I'm being restrictive in the documentation, but forgiving in the implementation.  Because of edge cases around 0 degrees, it turned out to be easier to implement negative values anyway.  Docs are just to encourage users to keep it to 0-359.

server/data_structures/internal_datafacade.hpp
```diff
@@ -447,13 +448,15 @@ template <class EdgeDataT> class InternalDataFacade final : public BaseDataFacad
         }
 
         return m_static_rtree->IncrementalFindPhantomNodeForCoordinate(
-            input_coordinate, resulting_phantom_node_vector, number_of_results);
+            input_coordinate, resulting_phantom_node_vector, number_of_results, 1100, bearing, range);
```
daniel-j-h: Meh, it's bad enough that we hard-code the max. distance, but having to hard-code it multiple times is just ugly.

server/data_structures/shared_datafacade.hpp
```diff
@@ -418,13 +419,15 @@ template <class EdgeDataT> class SharedDataFacade final : public BaseDataFacade<
         }
 
         return m_static_rtree->second->IncrementalFindPhantomNodeForCoordinate(
-            input_coordinate, resulting_phantom_node_vector, number_of_results);
+            input_coordinate, resulting_phantom_node_vector, number_of_results, 1100, bearing, range);
```
daniel-j-h: Same here with hard-coding the max. distance. I understand how this is required here for being able to pass bearing and range with values other than the default ones. But maybe reorder the arguments then?

server/data_structures/internal_datafacade.hpp
danpat: Arg, I cringed as I typed that, I meant to re-order those parameters so it didn't need to be specified.  I'll fix this.

unit_tests/data_structures/static_rtree.cpp
```diff
@@ -487,4 +493,100 @@ BOOST_AUTO_TEST_CASE(rectangle_test)
...
+    BOOST_CHECK_EQUAL(false, TestStaticRTree::IsBearingWithinBounds(344, -5, 10));
+    BOOST_CHECK_EQUAL(false, TestStaticRTree::IsBearingWithinBounds(6, -5, 10));
+
+    BOOST_CHECK_EQUAL(true, TestStaticRTree::IsBearingWithinBounds(-1, 5, 10));
+    BOOST_CHECK_EQUAL(false, TestStaticRTree::IsBearingWithinBounds(-6, 5, 10));
+
+    BOOST_CHECK_EQUAL(true, TestStaticRTree::IsBearingWithinBounds(-721, 5, 10));
+    BOOST_CHECK_EQUAL(true, TestStaticRTree::IsBearingWithinBounds(719, 5, 10));
+
+}
```
daniel-j-h: Hmm reading those tests I have to think about edge-cases like `std::numeric_limits<int>::max()`. It should be safe right now, but would make for great tests when the implementation changes, as we handle values from the user (outside) here.

emiltin:
very nice.  for viaroute, can you pass a bearing just for A, not B?

data_structures/static_rtree.hpp
```diff
@@ -737,10 +782,33 @@ class StaticRTree
...
+                    !IsBearingWithinBounds(backward_edge_bearing, filter_bearing, filter_bearing_range))
+                {
+                    continue;
+                }
+
                 // store phantom node in result vector
                 result_phantom_node_vector.emplace_back(current_segment,
                                                         foot_point_coordinate_on_segment);
 
+                if (!IsBearingWithinBounds(forward_edge_bearing, filter_bearing, filter_bearing_range))
```
TheMarex: Can we cache the result from the first if above for this?

data_structures/static_rtree.hpp
```diff
@@ -893,6 +978,15 @@ class StaticRTree
                         current_segment.forward_travel_mode, current_segment.backward_travel_mode),
                     current_perpendicular_distance);
 
+                if (!IsBearingWithinBounds(forward_edge_bearing, filter_bearing, filter_bearing_range))
```
TheMarex: Same as in the other call, I think we should cache the result here.

plugins/match.hpp
TheMarex: Actually this will work. What probably won't work would be `bearing, bearing, bearing, coordinate, coordinate, coordinate`. We should actually fix this on a grammar level.

server/data_structures/shared_datafacade.hpp
TheMarex: :+1: for re-ordering.

unit_tests/data_structures/static_rtree.cpp
TheMarex: We should handle validation of route parameters on the server level (e.g. `route_parameters.cpp`).

This is definitely already garbage-in garbage-out territory.

daniel-j-h: Right, but not garbage-in, undefined-behavior-out.

data_structures/static_rtree.hpp
danpat: I foresee changing this at runtime at some future time, but not in this PR. I'll `constexpr` these.

unit_tests/data_structures/static_rtree.cpp
danpat: I added a range check to `route_parameters.cpp`.

data_structures/static_rtree.hpp
```diff
@@ -642,10 +642,55 @@ class StaticRTree
...
+                                      const Angle B,
+                                      const int range)
+    {
+
+        if (range >= 180) return true;
+        if (range <= 0) return false;
+
+        // Map both bearings into positive modulo 360 space
+        const int normalized_B = (int(B) < 0)?(int(B) % 360)+360:(int(B) % 360);
+        const int normalized_A = (int(A) < 0)?(int(A) % 360)+360:(int(A) % 360);
```
daniel-j-h: Please no C-style casts, a la `(int)`. Those will try all casts, with even falling back to `reinterpret_cast` (which is bad). Do a `static_cast` here, maybe cache its value.

Or better: strongly typedef-ing does not prevent you from extending the type.
You could still add functions, such as:

    inline int normalize(const Angle& a) { return /* my normalization code */; }

and then reuse that throughout the codebase. You can even put your typedef and those functions into their own namespace. With ADL you only have to say `normalize(myns::Angle(90))`, as normalize is being searched in the `msns` namespace automatically (this is ADL).

strong_typedef.hpp
```diff
@@ -0,0 +1,92 @@
...
+    struct D                                                        \
+        : boost::totally_ordered1< D                                \
+        , boost::totally_ordered2< D, T                             \
+        > >                                                         \
+    {                                                               \
+        T val;                                                      \
+        explicit D(const T t_) : val(t_) {};                        \
+        D(): val() {};                                              \
+        D(const D & t_) : val(t_.val){}                             \
+        D & operator=(const D & rhs) { val = rhs.val; return *this;}  \
```
daniel-j-h: This prevents the compiler from generating the special member move functions. Argh, I'm not sure this Boost typedef thing is up to date at all (re. C++11), after looking at the details.

Reference:

- http://accu.org/content/conf2014/Howard_Hinnant_Accu_2014.pdf p28

PR #1712, opened 2015-09-28
Do not use an incomplete type with value semantics

daniel-j-h:
> Any of the following contexts requires class T to be complete:
>   - definition or function call to a function with return type T or argument type T; 

Let's wait for Travis on this one!

Reference:

- http://en.cppreference.com/w/cpp/language/type#Incomplete_type


daniel-j-h:
Boom! Hitting the error in the first try! Travis gives okay! Merging this hotfix now!

PR #1713, opened 2015-09-29
Don't count ferries as public transportation

TheMarex:
This allows ferries even if you disable public transportation.


emiltin:
tests? :-)

TheMarex:
@emiltin problem is that `use_public_transportation` is set to true by default in upstream OSRM. No sure how to properly test this without running `sed` on the profile or something.

emiltin:
ah ok. i guess we don't aim at testing the behaviour of profiles when you modify flags.

PR #1715, opened 2015-09-30
Run Modernizer

daniel-j-h:
Does what it says.

We probably should add a Travis build for this with the latest Clang, but oh well..

### DIFF
--- a/datastore.cpp
+++ b/datastore.cpp
@@ -219,8 +219,8 @@ int main(const int argc, const char *argv[]) try
     }();
 
     // Allocate a memory layout in shared memory, deallocate previous
-    SharedMemory *layout_memory = SharedMemoryFactory::Get(layout_region, sizeof(SharedDataLayout));
-    SharedDataLayout *shared_layout_ptr = new (layout_memory->Ptr()) SharedDataLayout();
+    auto *layout_memory = SharedMemoryFactory::Get(layout_region, sizeof(SharedDataLayout));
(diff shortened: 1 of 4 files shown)


daniel-j-h:
Only the Travis GCC 4.8 Debug build fails on this. How is it even possible..? The changes look good.
Notably, some cucumber tests fail, with the note that osrm isn't running.

datastore.cpp
```diff
@@ -229,7 +229,7 @@ int main(const int argc, const char *argv[])
         // Allocate a memory layout in shared memory, deallocate previous
         SharedMemory *layout_memory =
             SharedMemoryFactory::Get(layout_region, sizeof(SharedDataLayout));
-        SharedDataLayout *shared_layout_ptr = new (layout_memory->Ptr()) SharedDataLayout();
```
TheMarex: Uhm shouldn't this be `auto*`?

daniel-j-h: Nope, `auto` type deduction works like template type deduction. And `T` vs `T*` matters only for human readability. Here's an example with a nice trick to force the compiler to show me the infered types. In both cases they are `char*`: http://coliru.stacked-crooked.com/a/302af5710763fb14

TheMarex:
Other than the failing travis build, this should be good. Restarted, looks like a blip. :+1:

c0nk:
What about trying to revert the changes one-by-one until it passes?

PR #1716, opened 2015-09-30
Use ccache by default if available and a suitable compiler is used.

daniel-j-h:
This checks if `ccache` is available, and if so uses it.
The user can stil disable it via the ccache env variable, quoting:

    disable (CCACHE_DISABLE) [boolean]
        When true, ccache will just call the real compiler, bypassing the cache completely. The default is false.

At least Clang required `CCACHE_CPP2`.

The user does not have to set up anything, just to install ccache.
Of course, things like the cache's max size, its location and so on can
be configured.

References:

- https://ccache.samba.org/manual.html


daniel-j-h:
Apparently works on osx, too. Wohoo!

PR #1717, opened 2015-10-01
osrm-components: remove unused param

freenerd:
We don't use restrictions anymore

/cc @TheMarex


TheMarex:
@freenerd `develop` not `master`. :-)

freenerd:
@TheMarex can you change the default branch from master to develop?

Issue #1709, opened 2015-09-24
Fast method for updating edge weights from alternate speed/travel time data sources


danpat:
We're looking at ways to ingest data from real-time traffic sources and update edge weights to affect routing.  @TheMarex and I discussed some of the technicalities by voice this morning, this ticket is to capture what we talked about and create a forum for discussion before I go ahead and implement it.

# What we want to do
We want to update edge weights across the road graph when fed data from a variety of sources

# Why it's hard
 - We need to re-run `osrm-prepare` (specifically, the contraction phase) if the edge weights are changed
 - We want to avoid re-running the fairly expensive early parts of `osrm-extract`
 - To minimize turnaround times, we only want to re-run the parts we absolutely have after to we import the latest traffic data
 - Modifying edge compression could lead to an increase in memory usage, which we really want to avoid
 - Identifying the correct edges to update after data has been converted between structures a few times is currently impossible

# What we came up with
We will modify the workflow and insert two new steps.  It will now look like this:

`osrm-extract`
  - Parse OSM
  - Construct node-based graph
  - Compress edges
  - Perform edge expansion **NEW** (moved here from `osrm-prepare`)
  - Save segment-to-expanded-edge lookup table  **NEW**.  "segment" means a coordinate pair that are joined by a navigable edge in the original OSM data source, i.e. `nodeA -> nodeB`

`osrm-update-edges`  **NEW** optional phase to quickly update edge weights
  - Modify edge-expanded graph weights using segment-to-expanded-edge mapping file
  - This may or may not be a separate tool to `osrm-prepare`, TBD.

`osrm-prepare`
  - ~~Perform edge expansion~~ Moved to `osrm-extract`
  - Load edge-expanded graph
  - Contract graph

`osrm-routed`
  - Unchanged

With this in place, it will be possible to quickly update edge weights on a partially processed datafile.  I expect `osrm-update-edges` should take < 60s to update a full-sized planet, it will be a linear-ish operation.  Combined with new `-k` parameter to `osrm-prepare`, should enable processing turnaround times fast enough to ingest traffic data and get usable data back to users of the service in enough time to be useful.  We're aiming for total turnaround speeds of <15 minutes for USA sized road networks (tunable).

The segment lookup table will look something like this:

Expanded EdgeID | OSM From NodeID | OSM To NodeID | Length (metres)
---|---|---|---
12376 | 1234 |  3423 | 342
12376 | 3423 |  8237 | 983
[...]

PR #1719, opened 2015-10-01
Refactor edge expansion into extract phase

danpat:
This PR is the first step towards of #1709

It moves all the code for RTree generation, edge expansion and graph compression out of `osrm-prepare` and into `osrm-extract`.  A new intermediate file is written out (with the suffix `.ebg`, short for 'edge-based-graph`) and that new file is used by `osrm-prepare` to load the edge-based-graph for compression.

This refactor does it in the simplest-possible way.  No attempt was made to keep the node-based-graph in memory for conversion.  The STXXL structures are destroyed via RAII, and the node-based-graph data is re-read.  The code is as deliberately unchanged as possible.  The additional I/O introduced by re-reading the node-based-graph in between parsing the OSM and constructing the edge-based-graph should be relatively trivial in the grand scheme of things, but could probably be optimized at a later date.


contractor/processing_chain.cpp
```diff
@@ -130,81 +100,40 @@ int Prepare::Run()
...
+    edge_based_edge_list.resize(number_of_edges);
+    SimpleLogger().Write() << "Reading " << number_of_edges << " edges from the edge based graph";
+
+    // TODO: can we read this in bulk?  DeallocatingVector isn't necessarily
+    // all stored contiguously
+    for (;number_of_edges > 0; --number_of_edges) {
+        EdgeBasedEdge inbuffer;
+        input_stream.read((char *) &inbuffer, sizeof(EdgeBasedEdge));
+        edge_based_edge_list.emplace_back(std::move(inbuffer));
     }
```
daniel-j-h: I don't think you need to rely on `DeallocatingVector` here. Simply chunk into buffer and emplace that one by one then. Rough idea:

- local `std::array<EdgeBasedEdge, ChunkSize> buffer;`
- read `n = min(ChunkSize, number_of_edges)` into buffer
- `std::move(begin(buffer), begin(buffer)+n, back_inserter(edge_based_edge_list))`

note: the last move is not the usual single-argument move but the [range-move](http://en.cppreference.com/w/cpp/algorithm/move), (also possible with insert and move-iterators).

extractor/extractor.cpp
```diff
@@ -253,6 +262,57 @@ int extractor::run()
 
         TIMER_STOP(extracting);
         SimpleLogger().Write() << "extraction finished after " << TIMER_SEC(extracting) << "s";
+    }
+    catch (std::exception &e)
```
daniel-j-h: Always catch by const-ref. All the member functions you want to invoke (you can invoke) are marked const.

extractor/extractor.cpp
```diff
@@ -253,6 +262,57 @@ int extractor::run()
...
+        // movement (e.g. turn from A->B, and B->A) becomes an edge
+        //
+        //
+        //    // Create a new lua state
+
+        SimpleLogger().Write() << "Generating edge-expanded graph representation";
+
+        TIMER_START(expansion);
+
+        auto node_based_edge_list = osrm::make_unique<std::vector<EdgeBasedNode>>();;
```
daniel-j-h: I fixed this and similar issues in https://github.com/Project-OSRM/osrm-backend/pull/1664. Please don't introduce them back into the code base.

`std::vector` already owns and thus manages its resources. There is no need for `std::unique_ptr<std::vector<>>`, as this brings no benefits, but adds one level of indirection. If you want to have a read-only view of a vector, take it by const-ref. If you take ownership of a vector, take it by-value leaving the call site with the choice to either pass in a copy or use `std::move` to transfer ownership.

extractor/extractor.cpp
```diff
@@ -262,5 +322,309 @@ int extractor::run()
         SimpleLogger().Write(logWARNING) << e.what();
         return 1;
     }
+
     return 0;
 }
+
+unsigned extractor::CalculateEdgeChecksum(std::unique_ptr<std::vector<EdgeBasedNode>> node_based_edge_list)
```
daniel-j-h: ag -iQ "unique_ptr<std::vec"

I want this to return not a single match :P

extractor/extractor.cpp
```diff
@@ -262,5 +322,309 @@ int extractor::run()
...
+    SimpleLogger().Write() << "CRC32: " << crc32_value;
+
+    return crc32_value;
+}
+
+/**
+    \brief Setups scripting environment (lua-scripting)
+    Also initializes speed profile.
+*/
+void extractor::SetupScriptingEnvironment(lua_State *lua_state, SpeedProfileProperties &speed_profile)
```
daniel-j-h: Hmm I know this code only got moved around, but I wonder why we take `lua_state` by pointer, when a reference would be good enough. Either that, or at least `assert(lua_state);`.

extractor/extractor.cpp
```diff
@@ -262,5 +322,309 @@ int extractor::run()
...
+
+    // adjust lua load path
+    luaAddScriptFolderToLoadPath(lua_state, config.profile_path.string().c_str());
+
+    // Now call our function in a lua script
+    if (0 != luaL_dofile(lua_state, config.profile_path.string().c_str()))
+    {
+        std::stringstream msg;
+        msg << lua_tostring(lua_state, -1) << " occured in scripting block";
+        throw osrm::exception(msg.str());
```
daniel-j-h: All the `stringstreams` here are unnecessary, you can just concat a C-string and a stdlib string:

    { const char* cstring = "abc";  string s;  s += cstring;  cout << s; }

Note: `std::string` has a C-String taking constructor, that invokes undefined behavior if the C-String is null. So if in doubt just default construct it (empty string) and then concat. Otherwise everything works as expected.

contractor/processing_chain.cpp
```diff
@@ -130,81 +100,40 @@ int Prepare::Run()
...
+
+    edge_based_edge_list.resize(number_of_edges);
+    SimpleLogger().Write() << "Reading " << number_of_edges << " edges from the edge based graph";
+
+    // TODO: can we read this in bulk?  DeallocatingVector isn't necessarily
+    // all stored contiguously
+    for (;number_of_edges > 0; --number_of_edges) {
+        EdgeBasedEdge inbuffer;
+        input_stream.read((char *) &inbuffer, sizeof(EdgeBasedEdge));
+        edge_based_edge_list.emplace_back(std::move(inbuffer));
```
TheMarex: Easiest way to speed this up is to add a method to `DeallocationVector` that return a reference to a bucket and directly read into this bucket.

extractor/extractor.cpp
danpat: I'd like to fix this stuff separately and keep this PR (and thus, commit) focused just on moving the functionality around.

TheMarex:
@danpat Rebase on `develop` and then it is good to merge. :+1:

PR #1720, opened 2015-10-02
Fixed a bug when extracting data from a file with cut streets (with m…

jordanmarkov:
Fixed a bug when extracting data from a file with cut streets (with missing nodes).
This could happen when trying to use part of highly connected road network, like part
of a city, and the osm file is missing data outsite the bounding box, as is the case when
using `osmosis` with `--bounding-polygon` option.

### DIFF
--- a/extractor/extraction_containers.cpp
+++ b/extractor/extraction_containers.cpp
@@ -250,6 +250,16 @@ void ExtractionContainers::PrepareEdges(lua_State *segment_state)
         edge_iterator->source_coordinate.lon = node_iterator->lon;
         ++edge_iterator;
     }
+
+    // Remove all remaining edges. They are invalid because there are no corresponding nodes for
+    // them. This happens when using osmosis with bbox or polygon to extract smaller areas.
+    auto markSourcesInvalid = [](InternalExtractorEdge &edge)
+    {
+        SimpleLogger().Write(LogLevel::logWARNING) << "Found invalid node reference "
+                                                   << edge.result.source;
+        edge.result.source = SPECIAL_NODEID;
+    };
+    std::for_each(edge_iterator, all_edges_list_end, markSourcesInvalid);
     TIMER_STOP(set_start_coords);
     std::cout << "ok, after " << TIMER_SEC(set_start_coords) << "s" << std::endl;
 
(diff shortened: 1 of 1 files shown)


daniel-j-h:
Oh great! We also ran into this problem once, I'm not sure how we fixed it back then (/cc @danpat @TheMarex), but at least OSRM wasn't segfaulting any longer than.

Small nitpicking here: I would prefer showing intend with something like:

```c++
auto markSourceInvalid = [](const auto& edge){ edge.result.source = SPECIAL_NODEID; }
for_each(it, end, markSourceInvalid);
```

(and if you need the number of elements in that range `distance(it, end)` is in `O(1)`)

jordanmarkov:
I did do this, but it's not so clean as I'd like it to be. Apparentrly I can't use auto lambda parameters in C++11.

daniel-j-h:
Rright, you have to specify the type with C+11. Also, algorithms also work fine with empty ranges :P.

extractor/extraction_containers.cpp
```diff
@@ -250,6 +250,20 @@ void ExtractionContainers::PrepareEdges(lua_State *segment_state)
         edge_iterator->source_coordinate.lon = node_iterator->lon;
         ++edge_iterator;
     }
+
+    // Remove all remaining edges. They are invalid because there are no corresponding nodes for
+    // them. This happens when using osmosis with bbox or polygon to extract smaller areas.
+    if (0 < std::distance(edge_iterator, all_edges_list_end))
```
TheMarex: Check can be removed since `std::for_each` can handle empty ranges.

TheMarex:
If you remove the check and rebase on `develop` this is good to merge!

Thanks for catching this. Didn't think about this special case. :+1:

jordanmarkov:
I could remove the check, but then the log message would be lost.

TheMarex:
It okay to pull the message in the lambda and warn for every node like here
https://github.com/Project-OSRM/osrm-backend/pull/1720/files#diff-8c1bec9519b9760fa906373c44d539f2R222

jordanmarkov:
Is it possible to restart the AppVeyor build? It failed with a timeout, but the code should be fine.

TheMarex:
All builds pass. Perfect!. :+1:

PR #1730, opened 2015-10-09
We are on v4.8.1.

daniel-j-h:
Missing that is not cool :/ How do we handle this in the future? Context: the reason we ripped out the automatic Gitsha retrieval was that we tagged on master but built from develop, and that other issues.


TheMarex:
:+1:

Issue #1698, opened 2015-09-19
How to run Europe on 8GB? - part 2


maximd1:
Please, don't close (like https://github.com/Project-OSRM/osrm-backend/issues/1697). I really need to have Europe running on machine with 8GB of RAM. I have a lot of swap space, but osrm-routed kills every time I try to start it.

Here is a log:

[info] Reading options from: server.ini
[info] starting up engines, HEAD-HASH-NOTFOUND, compiled at Aug 11 2015, 22:58:44
[info] HSGR file:       "/mnt/pbf/europe-latest.osrm.hsgr"
[info] loading graph data
[info] loading graph from /mnt/pbf/europe-latest.osrm.hsgr
[info] loaded 86031783 nodes and 417438212 edges
[info] Data checksum is 3545495023
[info] loading egde information
Killed

/var/log/messages:

Sep 19 22:17:04 OSRM kernel: Purging GPU memory, 0 bytes freed, 24576 bytes still pinned.
Sep 19 22:17:05 OSRM kernel: osrm-routed invoked oom-killer: gfp_mask=0x280da, order=0, oom_adj=0, oom_score_adj=0
Sep 19 22:17:05 OSRM kernel: osrm-routed cpuset=/ mems_allowed=0
Sep 19 22:17:05 OSRM kernel: Pid: 1557, comm: osrm-routed Not tainted 2.6.32-573.3.1.el6.x86_64 #1
Sep 19 22:17:05 OSRM kernel: Call Trace:
Sep 19 22:17:05 OSRM kernel: [<ffffffff810d6dd1>] ? cpuset_print_task_mems_allowed+0x91/0xb0
Sep 19 22:17:05 OSRM kernel: [<ffffffff8112a5d0>] ? dump_header+0x90/0x1b0
Sep 19 22:17:05 OSRM kernel: [<ffffffff81232cbc>] ? security_real_capable_noaudit+0x3c/0x70
Sep 19 22:17:05 OSRM kernel: [<ffffffff8112aa52>] ? oom_kill_process+0x82/0x2a0
Sep 19 22:17:05 OSRM kernel: [<ffffffff8112a991>] ? select_bad_process+0xe1/0x120
Sep 19 22:17:05 OSRM kernel: [<ffffffff8112ae90>] ? out_of_memory+0x220/0x3c0
Sep 19 22:17:05 OSRM kernel: [<ffffffff8113786c>] ? __alloc_pages_nodemask+0x93c/0x950
[...]

PR #1731, opened 2015-10-09
Conditional memlock

daniel-j-h:
Note: the noise comes from the 2nd commit, flattening plus clang-format.

-------------------------------------------------------------------------------

Only lock the virtual address space when shared memory was requested.

In addition, some improvements:

- unlock only when locking succeeded
- scoped exception safe RAII locker

Reference:

- #1698 (comment)

### DIFF
--- a/routed.cpp
+++ b/routed.cpp
@@ -91,11 +91,23 @@ int main(int argc, const char *argv[])
         }
 
 #ifdef __linux__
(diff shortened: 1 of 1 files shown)


hoerup:
It looks to me that you don't test for the value of shouldLock

daniel-j-h:
You are right, that went out with introducing a raii type --- will push a fix.

PR #1740, opened 2015-10-18
Fix various bugs

c0nk:


daniel-j-h:
Great, this looks good to me! Could you elaborate in which cases `std::size_t` is smaller than `std::int64_t`? I know that technically speaking we don't have any guarantees there, but on the other hand we only support 64 bit systems because of the `unsigned` mess that is used throughout the codebase instead of fixed width integral types. On 64 bit systems `std::size_t` should be 64 bit, though.

c0nk:
The issue is specifically on 32-bit systems where sizeof(size_t) == 4. (size_t)(0) - 1 evaluates to to (size_t)-1. If sizeof(size_t) == 4 then (int64_t)(size_t)-1 == (int64_t)UINT32_MAX == UINT32_MAX which causes buffer overflows. For 64-bit systems where sizeof(size_t) == sizeof(int64_t) then (int64_t)(size_t)-1 == (int64_t)UIN64_MAX == (int64_t)-1 (which i think itself relies on implementation-defined behavior). The way its written now works for both 32 and 64 bit systems and avoids implementation-defined behavior except when .size() returns > INT64_MAX which practically can't happen.

daniel-j-h:
Please note that we do not really support 32 bit systems. This is most prominently due to the unfortunate decision to use `unsigned` throughout the code base.

You are right about the `min(size_t{x}, size_t{y}) - 1` issue, good catch! Just thinking about if it's logically possible at all for both to be empty.

c0nk:
I need this to work on 32-bit systems. I am happy to submit patches for other problems related to uses of 'unsigned' that I run into.

It is possible that they are both empty. I am running OSRM on real 32-bit devices and this bug really happens.

TheMarex:
@c0nk using OSRM with 32bit system does not make sense. For non-trivial datasets you will already reach the limits of 32bit addressable memory (< 4GB).
What are you trying to do?

util/json_renderer.hpp
```diff
@@ -76,7 +76,7 @@ struct Renderer : mapbox::util::static_visitor<>
     void operator()(const Array &array) const
     {
         out << "[";
-        for (auto it = array.values.cend(), end = array.values.cend(); it != end;)
```
TheMarex: Uhm, wow this is a bad error. That would mean simpleclient did never work?

c0nk:
Even for 64-bit systems, that change avoids implementation-defined behavior for the size() == 0 case.

I am using OSRM to generate routes offline on embedded devices like phones/tablets which may be 32-bit. The largest dataset I need this to work for currently is Germany.

TheMarex:
> I am using OSRM to generate routes offline on embedded devices like phones/tablets which may be 32-bit. The largest dataset I need this to work for currently is Germany.

Interesting. Though on mobile devices I would suggest using a routing engine that could handle external memory (e.g. storing the data only on SD card and not in RAM). If you only want a datasets the size of Germany this will however work fine I guess.

If you rebase on `develop` this is good to merge. :+1: Thanks again!

c0nk:
This is based off of Develop. What else should I do?

Hey do you know of any other free routing engines that use external memory?

TheMarex:
@c0nk just run `git rebase develop` with an updated develop and a `git push -f origin`. It's mainly to get a cleaner git history. :-)

There is http://wiki.navit-project.org/index.php/Main_Page but I'm not sure how active they are anymore.

c0nk:
Rebased. Enjoy.

TheMarex:
:tada:

util/json_renderer.hpp
daniel-j-h: @TheMarex I did a git blame and it looks like I accidentally introduced that when doing the global [cache end iterator](https://github.com/Project-OSRM/osrm-backend/pull/1687) thing. Ouch! At least now we know how many people are using the json renderer with the current release. :P

Issue #1729, opened 2015-10-09
Provide better out-of-memory messages


emiltin:
When osrm bins run out of memory, they will often exit with messages that users intepret as a crash.

- Perform simple upfront checks to avoid starting task that can obvious not be completed due to lack of memory.
- Provide clearer messages in case memory runs out.

daniel-j-h:
There are a few sides to this problem, e.g. we may get killed by the Kernel OOM killer, e.g. due to fragmentation as seen in https://github.com/Project-OSRM/osrm-backend/issues/1698. In case we throw a `std::bad_alloc` there should be no reason why not to catch it at top level and translate it to a better human readable message.

danpat:
Ref #1736

daniel-j-h:
`std::bad_alloc` handled in https://github.com/Project-OSRM/osrm-backend/issues/1741 --- feel free to change the messages or add more information.

daniel-j-h:
Nothing actionable here, closing. We did not implement heuristics for file sizes, memory usage or related. If we detect that we are out of memory, we simply print human readable messages. Tracking I/O system failures such as out-of-disk-capacity in https://github.com/Project-OSRM/osrm-backend/issues/1682.

PR #1741, opened 2015-10-18
Catch `std::bad_alloc` at the top and translate to human readable messages

daniel-j-h:
This catches `std::bad_alloc` in the executables' main entry points and provides basic human readable error messages. Feel free to improve those messages.

Also, for the datastore I would like to get in https://github.com/Project-OSRM/osrm-backend/pull/1731 first, otherwise the locking / unlocking is ugly.

Note: noise from flattening plus clang-format, see specific commit.

References:

- https://github.com/Project-OSRM/osrm-backend/issues/1729

### DIFF
--- a/datastore.cpp
+++ b/datastore.cpp
@@ -93,360 +93,437 @@ void delete_region(const SharedDataType region)
     }
 }
 
-int main(const int argc, const char *argv[])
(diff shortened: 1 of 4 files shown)


TheMarex:
Rebased and waiting for travis before merging. :+1:

PR #1742, opened 2015-10-18
Provide a Makefile wrapping the beast that CMake is

daniel-j-h:
This provides an easy to use Makefile, that wraps the build process
behind a simple `make` command. In particular, it builds in release mode
and in parallel by default. It also separates release builds from debug
builds, in the build directory.

No more

    mkdir build
    cd build
    cmake .. -DCMAKE_BUILD_TYPE=Release
    cmake --build -- .

from now on just do

    make

Discussion: should be update all our documentation, readmes and wikis,
to simply invoke `make`?

To be done:

- the raster unit tests hardcode "../unit_tests/" as path prefix; this
  breaks with a build directory such as "build/Release/"

- the cucumber setup assumes the osrm binaries are located in "build/"

/cc @lbud @emiltin

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,26 @@
+default: all
+
+BUILD_TYPE ?= Release
+BUILD_DIR := build/${BUILD_TYPE}
+ROOT_DIR := $(shell pwd)
+
+NPROCS:=1
+OS:=$(shell uname -s)
+
+ifeq ($(OS),Linux)
+	NPROCS := $(shell nproc)
+else ifeq ($(OS),Darwin)
+	NPROCS := $(shell sysctl hw.ncpu | awk '{print $$2}')
+endif
+
+${BUILD_DIR}/Makefile: CMakeLists.txt
+	@mkdir -p ${BUILD_DIR}
+	@cd ${BUILD_DIR} && cmake ${ROOT_DIR} -DCMAKE_BUILD_TYPE=${BUILD_TYPE}
+
+all: ${BUILD_DIR}/Makefile
+	@${MAKE} -C ${BUILD_DIR} -j ${NPROCS}
+
+clean:
+	@${RM} -rf ${BUILD_DIR}
+
+.PHONE: all clean


alex85k:
Interesing idea. But why not to use simpler shell scripts?
THere may be ineresting problems on BSD-etc like gmake calls cmake calls make. Or maybe it calls gmake again...

daniel-j-h:
With a makefile we can include dependencies, such as the tests depends on the osrm executables and so on. The main idea is to make the common task of building osrm with some sane default described above easy and the not so common task such as customizing the cmake buildsystem still available for "advanced users". I would say, for the most use cases a simple `make` as described above does The Right Thing (tm).

danpat:
:+1: Works great on OSX (with BSD-like `make` and no `gmake` installed)

alex85k:
By the way, you can add "make prepare" with OS detection (Ubuntu or Redhat or MacOS) and things like
```
apt-get install .....
gem install ....
```
(better not ot call it automatically, however).

woodbri:
A nice feature for the makefile would be to make target "all" default to "help" and to have "help" target just echo out usage info, like a list of targets and what they do. This helps the occasional user remember how to use it.

```
all: help
help:
    @echo "Usage: make help|install-deps|build|install|..."
    @echo "    help - this message"
    @echo "    install-deps - install dependent packages"
    @echo "    build - compile OSRM"
    ...
```

daniel-j-h:
Wait wait wait. This is neither a dependency manager, nor a replacement for the Docs or Wiki. It's just to make the oh-so-common task of compiling OSRM easier. In fact, I did this in the first place because I'm lazy and like to call just `make` or `make CXX=clang++` instead of every time checking for a build directory, invoking CMake, building it and so on.

emiltin:
you can already build just by running 'rake'. see the Rakefile, or run 'rake -T':

```
rake build     # Build using CMake
rake clean     # Delete preprocessing files
rake crop      # Crop OSM data
rake cucumber  # Run Cucumber features
rake default   # Rebuild and run tests
rake down      # Stop the routing server
rake download  # Download OSM data
rake extract   # Extract OSM data
rake kill      # Kill all osrm-extract, osrm-prepare and osrm-routed processes
rake pid       # Get PIDs of all osrm-extract, osrm-prepare and osrm-routed processes
rake prepare   # Prepare OSM data
rake process   # Reprocess OSM data
rake run       # Run the routing server in the terminal
rake setup     # Setup config files
rake sweep     # Remove test cache files
rake test      # Run all cucumber test
rake up        # Launch the routing server in the background
rake update    # Stop, reprocess and restart
```

the Rakefile probably needs an update though...

TheMarex:
I'm on the fence here. Personally I hate it when people add another layer above cmake I first need to figure out if I really need to use the Makefile or can just go with vanilla cmake. Just having a `CMakeList.txt` makes it explicit that this is just a cmake project and you can build it like any other cmake project.

TheMarex:
Going to shelf this.

PR #1743, opened 2015-10-19
Don't generate files in source tree

c0nk:


CMakeLists.txt
```diff
@@ -32,13 +32,16 @@ endif()
 option(ENABLE_JSON_LOGGING "Adds additional JSON debug logging to the response" OFF)
 option(BUILD_TOOLS "Build OSRM tools" OFF)
 
+include_directories(${CMAKE_CURRENT_SOURCE_DIR})
```
TheMarex: Why do we need `${CMAKE_CURRENT_SOURCE_DIR}` here?

TheMarex:
Good idea. :+1:

Could we do `${CMAKE_CURRENT_BIN_DIR}/generated` instead to make it explicit that this was generated by cmake? @daniel-j-h any reservations here?

c0nk:
The `include_directories(${CMAKE_CURRENT_SOURCE_DIR})` is needed because `fingerprint_impl.hpp` includes files in the source tree.

I would recommend to include `${CMAKE_CURRENT_SOURCE_DIR}` anyway and write all includes relative to the root source directory. So instead of writing things like `#include "../util/ini_file.hpp"` you always simply write `#include "util/ini_file.hpp"`. One advantage of this is that generated files and non-generated files may always be included in the same way.

Of course this rule wouldn't apply to #includes from public headers that get installed.

c0nk:
Putting the generated files in a `gen` or a `generated` dir is very reasonable. Let me know if I should change that.

daniel-j-h:
Good catch with `util/version.hpp`: this was accidentally committed and should never live in the repository as it is generated by CMake based on the version information in the `CMakeLists.txt`.

Looks good to me, and setting the include directory to `${CMAKE_CURRENT_SOURCE_DIR}` probably makes sense for how this repository is structured right now.

c0nk:
Merge? :-)

TheMarex:
Thanks again @c0nk! :+1:

PR #1452, opened 2015-04-26
[NOT READY] General Many-To-Many plugin

TheMarex:
This plugin is in a very unfinished state. This pull request is just for tracking the progress.

TODO:

- [ ] Rebase on develop
- [ ] Evaluate what is actually implemented
- [ ] Add serialization to distance/time table
- [ ] Clean up code

Issue #1447, opened 2015-04-22
Make distance_table consistent regarding different phantom nodes


chaupow:
Currently the distance table is getting a list of locations that might have more than one phantom node per input coordinate.
Distance table should be changes such that

1) when computing the distance table, for every location the same phantom node is used 
2) when returning output, there should not only be the information about the distances but also which phantom node was used in the process

TheMarex:
More background here:

The main problem is [this part of the code](https://github.com/Project-OSRM/osrm-backend/blob/develop/routing_algorithms/many_to_many.hpp#L90-L104). The for _every location_ the `phantom_node_vector` that is returned by `StaticRTree` is inserted.  This vector usually contains up to 2 phantom nodes, one from a big and one from a small component.

Example: We have two locations `loc1`: `{a, b}` and `loc2`: `{c, d}`. The result of the query should be a table of size `2x2`. First step is to initialize the backward heap and run the backward search for each location. For `loc1` that would mean we insert `a` _and_ `b` in the heap and run the backward search. All nodes that are found by the backward search are marked with the distance to `a` _or_ `b`, depending on which one is shorter. But we don't save from which node the distance actually is.

- Retain the information for each node. When we do the backward search we retain information about the "original parent" in the heap data. This would mean modifying [this line](https://github.com/Project-OSRM/osrm-backend/blob/develop/routing_algorithms/many_to_many.hpp#L224) and [this line](https://github.com/Project-OSRM/osrm-backend/blob/develop/routing_algorithms/many_to_many.hpp#L230) to set the `parent` not to `node` but to parent of `node`. For all nodes found in the backward search in the example above that would yield nodes either marked `a` or `b`.

- Decide on exactly one phantom node. This would need similar logic as in [viaroute](https://github.com/Project-OSRM/osrm-backend/blob/develop/plugins/viaroute.hpp#L152-L160). Running this on the list of coordinates would discard all the small components if they are not from the same small component.

danpat:
So, to clarify the behaviour we're going for here:

  - If all nodes snap to a big component, we simply route those
  - If some components snap to small components *and* have nearby big components, and everything snaps to a big component, snap to to the big components (i.e. city parks < 1100m from a big components).
  - If all components snap to the same small component, use that.
  - If some components snap *only* to a small component, use that (i.e. islands more than 1100m from a big component).

TheMarex:
This was implemented to be consistent with `viaroute`. It uses the same snapping behavior.

Issue #1075, opened 2014-06-11
Travel time matrix with origins and destinations?


emiltin:
Would it be possible to provide origins and destinations separately to the travel time matrix calculation?

For example, you could provide a list of homes and a list of stations, and get all travel times between home-station, but not between home-home or station-station.

yusufumac:
Any update on this feature? I would be great to query with NXM matrices.

TheMarex:
@yusufumac https://github.com/Project-OSRM/osrm-backend/pull/1452
If you want to lend a helping hand, let me know.

rCarto:
+1

frodrigo:
Released in 4.9.0 can be closed.

PR #1760, opened 2015-11-05
Support rectangular matrix with less sources than targets

fab-girard:
Hello,
I've modified table plugin to support rectangular matrix and return mapped coordinates.
One simple test has been added with a 1x4 matrix.
If you are ok with this pull request, here the help to replace in osrm wiki:

## Service `table`

This computes distance tables for the given via points. Please note that the distance in this case is the _travel time_ which is the default metric used by OSRM.
If only `loc` parameter is used, all pair-wise distances are computed.
If `src` parameter is combined with `loc` parameter, only pairs between scr/loc are computed.

### Query

```
http://{server}/table?loc={lat,lon}&loc={lat,lon}<&loc={lat,lon} ...>
```

|Parameter    |Values                    |Description                                                   |
|-------------|--------------------------|--------------------------------------------------------------|
|loc          |lat,lon                   |Location of the via point                                     |
|src          |lat,lon                   |Location of the source point                                  |
|mapped_points|`true`, `false` (default) |Return the mapped coordinates                                 |

### Response

* `distance_table` array of arrays that stores the matrix in row-major order. `distance_table[i][j]` gives the travel time from
  the i-th via to the j-th via point. Values are given in 10th of a second.
* `target_mapped_coordinates` array of arrays that contains the `[lat, lon]` pair of the snapped coordinate
* `source_mapped_coordinates` array of arrays that contains the `[lat, lon]` pair of the snapped coordinate

### Example

```
http://router.project-osrm.org/table?loc=52.554070,13.160621&loc=52.431272,13.720654&loc=52.554070,13.720654&loc=52.554070,13.160621
```

```JSON
{
    "distance_table": [
        [
            0,
            31089,
            31224,
            0
        ],
        [
            31248,
            0,
            13138,
            31248
        ],
        [
            31167,
            13188,
            0,
            31167
        ],
        [
            0,
            31089,
            31224,
            0
        ]
    ]
}
```

```
http://router.project-osrm.org/table?loc=52.554070,13.160621&loc=52.431272,13.720654&loc=52.554070,13.720654&src=52.554070,13.160621
```

```JSON
{
    "distance_table": [
        [
            0,
            31089,
            31224
        ]
    ]
}
```

[Review on Reviewable](https://reviewable.io/reviews/project-osrm/osrm-backend/1760)

### DIFF
--- a/data_structures/route_parameters.cpp
+++ b/data_structures/route_parameters.cpp
@@ -36,7 +36,7 @@ SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
 RouteParameters::RouteParameters()
     : zoom_level(18), print_instructions(false), alternate_route(true), geometry(true),
-      compression(true), deprecatedAPI(false), uturn_default(false), classify(false),
+      compression(true), deprecatedAPI(false), uturn_default(false), classify(false), mapped_points(true),
       matching_beta(5), gps_precision(5), check_sum(-1), num_results(1)
 {
 }
@@ -144,6 +144,14 @@ void RouteParameters::addCoordinate(
     const boost::fusion::vector<double, double> &received_coordinates)
 {
     coordinates.emplace_back(
+        static_cast<int>(COORDINATE_PRECISION * boost::fusion::at_c<0>(received_coordinates)),
+        static_cast<int>(COORDINATE_PRECISION * boost::fusion::at_c<1>(received_coordinates)));
+}
+
+void RouteParameters::addDestination(
+    const boost::fusion::vector<double, double> &received_coordinates)
+{
+    destinations.emplace_back(
         static_cast<int>(COORDINATE_PRECISION * boost::fusion::at_c<0>(received_coordinates)),
         static_cast<int>(COORDINATE_PRECISION * boost::fusion::at_c<1>(received_coordinates)));
 }
--- a/data_structures/route_parameters.cpp
+++ b/data_structures/route_parameters.cpp
@@ -148,8 +148,18 @@ void RouteParameters::addCoordinate(
         static_cast<int>(COORDINATE_PRECISION * boost::fusion::at_c<1>(received_coordinates)));
 }
 
+void RouteParameters::addSource(
+    const boost::fusion::vector<double, double> &received_coordinates)
+{
+    sources.emplace_back(
+        static_cast<int>(COORDINATE_PRECISION * boost::fusion::at_c<0>(received_coordinates)),
+        static_cast<int>(COORDINATE_PRECISION * boost::fusion::at_c<1>(received_coordinates)));
+}
+
 void RouteParameters::getCoordinatesFromGeometry(const std::string &geometry_string)
 {
     PolylineCompressor pc;
     coordinates = pc.decode_string(geometry_string);
 }
+
+void RouteParameters::setMappedPointsFlag(const bool flag) { mapped_points = flag; }
--- a/plugins/distance_table.hpp
+++ b/plugins/distance_table.hpp
@@ -70,23 +70,32 @@ template <class DataFacadeT> class DistanceTablePlugin final : public BasePlugin
     int HandleRequest(const RouteParameters &route_parameters,
                       osrm::json::Object &json_result) override final
     {
-        if (!check_all_coordinates(route_parameters.coordinates) ||
-            (route_parameters.sources.size() && !check_all_coordinates(route_parameters.sources, 1)))
+        const bool useSameTgtSrc = route_parameters.coordinates.size() ? true : false;
+        if ((useSameTgtSrc && route_parameters.destinations.size()) || (useSameTgtSrc && route_parameters.sources.size()))
+        {
+            return 400;
+        }
+
+        if ((useSameTgtSrc && !check_all_coordinates(route_parameters.coordinates)) ||
+            (!useSameTgtSrc && !check_all_coordinates(route_parameters.destinations, 2) && !check_all_coordinates(route_parameters.sources, 1)))
         {
             return 400;
         }
 
         const auto &input_bearings = route_parameters.bearings;
-        if (input_bearings.size() > 0 && route_parameters.coordinates.size() + route_parameters.sources.size() != input_bearings.size())
+        unsigned nb_coordinates = useSameTgtSrc ? route_parameters.coordinates.size() : (route_parameters.destinations.size() + route_parameters.sources.size());
+        if (input_bearings.size() > 0 && nb_coordinates != input_bearings.size())
         {
             json_result.values["status"] = "Number of bearings does not match number of coordinates .";
             return 400;
         }
 
         const bool checksum_OK = (route_parameters.check_sum == facade->GetCheckSum());
-        unsigned max_locations =
+        unsigned max_locations = useSameTgtSrc ?
             std::min(static_cast<unsigned>(max_locations_distance_table),
-                     static_cast<unsigned>(route_parameters.coordinates.size()));
+                     static_cast<unsigned>(route_parameters.coordinates.size())) :
+            std::min(static_cast<unsigned>(max_locations_distance_table),
+                     static_cast<unsigned>(route_parameters.destinations.size()));
 
         PhantomNodeArray phantom_node_target_vector(max_locations);
         for (const auto i : osrm::irange(0u, max_locations))
@@ -104,14 +113,14 @@ template <class DataFacadeT> class DistanceTablePlugin final : public BasePlugin
             }
             const int bearing = input_bearings.size() > 0 ? input_bearings[i].first : 0;
             const int range = input_bearings.size() > 0 ? (input_bearings[i].second?*input_bearings[i].second:10) : 180;
-            facade->IncrementalFindPhantomNodeForCoordinate(route_parameters.coordinates[i],
+            facade->IncrementalFindPhantomNodeForCoordinate(useSameTgtSrc ? route_parameters.coordinates[i] : route_parameters.destinations[i],
                                                             phantom_node_target_vector[i], 1, bearing, range);
 
             BOOST_ASSERT(phantom_node_target_vector[i].front().is_valid(facade->GetNumberOfNodes()));
         }
-        unsigned nb_coordinates = route_parameters.coordinates.size();
+        unsigned shift_coordinates = (useSameTgtSrc) ? 0 : route_parameters.destinations.size();
         max_locations = 0;
-        if (route_parameters.sources.size())
+        if (!useSameTgtSrc)
         {
             max_locations = std::min(static_cast<unsigned>(max_locations_distance_table),
                                      static_cast<unsigned>(route_parameters.sources.size()));
@@ -119,19 +128,19 @@ template <class DataFacadeT> class DistanceTablePlugin final : public BasePlugin
         PhantomNodeArray phantom_node_source_vector(max_locations);
         for (const auto i : osrm::irange(0u, max_locations))
         {
-            if (checksum_OK && i < route_parameters.hints.size() - route_parameters.coordinates.size() &&
-                !route_parameters.hints[i+route_parameters.coordinates.size()].empty())
+            if (checksum_OK && i < route_parameters.hints.size() - shift_coordinates &&
+                !route_parameters.hints[i + shift_coordinates].empty())
             {
                 PhantomNode current_phantom_node;
-                ObjectEncoder::DecodeFromBase64(route_parameters.hints[i+route_parameters.coordinates.size()], current_phantom_node);
+                ObjectEncoder::DecodeFromBase64(route_parameters.hints[i + shift_coordinates], current_phantom_node);
                 if (current_phantom_node.is_valid(facade->GetNumberOfNodes()))
                 {
                     phantom_node_source_vector[i].emplace_back(std::move(current_phantom_node));
                     continue;
                 }
             }
-            const int bearing = input_bearings.size() > 0 ? input_bearings[nb_coordinates + i].first : 0;
-            const int range = input_bearings.size() > 0 ? (input_bearings[nb_coordinates + i].second?*input_bearings[nb_coordinates + i].second:10) : 180;
+            const int bearing = input_bearings.size() > 0 ? input_bearings[i + shift_coordinates].first : 0;
+            const int range = input_bearings.size() > 0 ? (input_bearings[i + shift_coordinates].second?*input_bearings[i + shift_coordinates].second:10) : 180;
             facade->IncrementalFindPhantomNodeForCoordinate(route_parameters.sources[i],
                                                             phantom_node_source_vector[i], 1, bearing, range);
 
--- a/plugins/distance_table.hpp
+++ b/plugins/distance_table.hpp
@@ -70,13 +70,14 @@ template <class DataFacadeT> class DistanceTablePlugin final : public BasePlugin
     int HandleRequest(const RouteParameters &route_parameters,
                       osrm::json::Object &json_result) override final
     {
-        if (!check_all_coordinates(route_parameters.coordinates))
+        if (!check_all_coordinates(route_parameters.coordinates) ||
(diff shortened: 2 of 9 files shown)


fab-girard:
This pull request is linked to #1075 #1447 and #1452

TheMarex:
Great work @fab-girard! What about adding both `src` and `dst` as parameters and `loc` meaning both `src` and `dst`? I like the addition of mapped coordinates: I don't think we need an API parameter for this, this can be on by default.

fab-girard:
It's not simple to change loc parameter... every osrm plugins seem to use this `loc` parameter.
1/ I don't want to break all plugin infrastructure
2/ The test infrastructure is strongly linked to this `loc` paramater

The parameter for mapped coordinated is added with the same idea : not to break existing behavior...

Fab

TheMarex:
> 1/ I don't want to break all plugin infrastructure

You misunderstood me. What would be great is:
- `loc` add coordinate to both `sources` and `targets`
- `src` add coordinate to `sources`
- `dst` add coordinate to `targets`

This way you are 100% backwards compatible since we only supported symmetric matrices that use `loc` for both source and targets.

emiltin:
great! a few more tests would be good, covering different number of source/destinations. the existing test has a as both src and dest, but i guess it's also possible to do ie. [a] -> [b,c,d] ?

fab-girard:
I agree with you: my first idea was to deliver more tests with for instance a 2x3 matrix. But tests' infrastructure is not enough free... I cannot use more than one `src` url param for query because it is stored in a hash where key is `src`. `loc` url param has a special treatment and, as I explained to @TheMarex, it's not simple with existing tests to avoid mandatory `loc` parameter.
I think it's possible to change tests' infra but I've not enough time for it now.

Note that code for rectangular and square matrix are the same. So existing tests are yet testing rectangular matrix ;) 

Fab

hoerup:
@fab-girard This is great i have a project where I only need one-to-many, so I did a test run with your patches last night. It reduced the total calc time from 4 hours to 2 hours :)

fab-girard:
In my case too, this enhancement has reduced by half the time for one job. This was my intended goal.
Hope this pull request will be merged soon ;) 

Fab

frodrigo:
@TheMarex do you want more tests by rework the tests infrastructure or do you think is enough like this? We can work on this point, as we like this pull request will be mainstreamed.

TheMarex:
@frodrigo yes I would really prefer the more general approach here as it is an API extension. Would love to have this in 4.9.0!

fab-girard:
@TheMarex what do you mean by general appoach ?
Is if ok if:
- I keep `loc` param mandatory and `src` optionnal: it's the most general approach according to me (and if it's not the case I'm afraid by changing all tests infra... :( )
- I try to add few more tests (for this point I will need to improve test infra in order to support more than one url multi param)

TheMarex:
> I keep loc param mandatory and src optionnal: it's the most general approach according to me (and if it's not the case I'm afraid by changing all tests infra... :( )

It is not the most general as it doesn't cover cases like one target and multiple sources or a disjunct set of multiple sources and multiple targets. If you need help updating the test infrastructure we are happy to help in any way.
I think you already found the relevant code locations that need to be modified:
- `api_grammar.cpp`:
    - Drop `mapped_points` and make it the default behavior
    - Make `src` work like `loc` (can be used multiple times)
    - Add `dst` and make it work like `src`
- Modify the plugin to deal with the general case of arbitrary list of sources and targets.
- Add tests for `src` and `dst`. You can have a look at the tests for the `match` plugin that has similar logic for the `t` parameter. If anything is unclear don't hesitate to ask.

fab-girard:
Sorry, I'm not agree with you (maybe we don't understand what other is saying) : what I've already pushed authorizes multiple `loc` AND multiple `src` parameter... Replace `loc`/`src` by `dst`/`src` will enhance nothing... it will only make code more complex :( 

To be clear:
>Drop mapped_points and make it the default behavior
ok (for info, this paramater was added to keep original behavior and improve output file's size for network if necessary)

>Make src work like loc (can be used multiple times)
my pull request authorizes already today the general case for any kind of rectangular matrix (#N>1 for `loc` and #M>1 for `src`)

>Add dst and make it work like src
unnecessary and make code uselessly more complex (see the begining of my answer)

>Modify the plugin to deal with the general case of arbitrary list of sources and targets.
It's already the case... (see my previous point)

>Add tests for src and dst. You can have a look at the tests for the match plugin that has similar logic for the t parameter.
ok

hoerup:
@fab-girard as I read your last comment - in your opinion, the only missing item is tests ?

fab-girard:
Yes it's my opinion, I'm waiting the GO from @TheMarex. If he's ok, I will just provide more tests in this PR ;) and maybe change default value of `mapped_points` param...
Fab

hoerup:
@fab-girard  While we wait for @TheMarex, could you do a rebase to latest so it at least can build with travis ?

fab-girard:
@hoerup : done ;) Sorry for the delay.

hoerup:
@emiltin @TheMarex Could you please re-visit this PR so @fab-girard can get an overview of what needs to be done ?

TheMarex:
>  Replace loc/src by dst/src will enhance nothing... it will only make code more complex :( 

- `loc`: Coordinate is both source and destination
- `src`: Coordinate is only source
- `dst`: Coordinate is only destination

Please move on the points above before I revisit this.

fab-girard:
New commit:
- use now `loc`/`dst`/`src`
- mapped points are returned by default
- one test added (2x3 matrix with one of sources which is not in destinations)

Linked to other PR / commit: https://github.com/Project-OSRM/node-osrm/pull/119

Fab

hoerup:
Nice work, @fab-girard

fab-girard:
Thx for your code review  @hoerup, I've added error status as requested.
Fab

hoerup:
@TheMarex What do you think of this PR now ?

TheMarex:
@fab-girard great work. I'm currently caught up in some refactoring, but tagged this for `4.9.0`.

data_structures/route_parameters.cpp
```diff
@@ -36,7 +36,7 @@ SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
 RouteParameters::RouteParameters()
     : zoom_level(18), print_instructions(false), alternate_route(true), geometry(true),
-      compression(true), deprecatedAPI(false), uturn_default(false), classify(false),
+      compression(true), deprecatedAPI(false), uturn_default(false), classify(false), mapped_points(true),
```
TheMarex: Please remove the `mapped_points` parameter entirely. It has no significant implications on size or performance. I'm only talking about the parameter for the API not the feature as such - it should just be the normal behavior and not user configurable.

plugins/distance_table.hpp
```diff
@@ -70,24 +70,34 @@ template <class DataFacadeT> class DistanceTablePlugin final : public BasePlugin
...
     {
-        if (!check_all_coordinates(route_parameters.coordinates))
+        const bool useSameTgtSrc = route_parameters.coordinates.size() ? true : false;
+        if ((useSameTgtSrc && route_parameters.destinations.size()) || (useSameTgtSrc && route_parameters.sources.size()))
+        {
+            return 400;
+        }
+
+        if ((useSameTgtSrc && !check_all_coordinates(route_parameters.coordinates)) ||
+            (!useSameTgtSrc && !check_all_coordinates(route_parameters.destinations, 2) && !check_all_coordinates(route_parameters.sources, 1)))
```
TheMarex: `useSameTgtSrc` is unneeded. This can be replaced with:

`if (!check_all_coordinates(route_parameter.coordinates) || !check_all_coordinates(route_parameter.sources) || !check_all_coordinates(route_parameters.destinations)) { return 400; }`

plugins/distance_table.hpp
```diff
@@ -70,24 +70,34 @@ template <class DataFacadeT> class DistanceTablePlugin final : public BasePlugin
...
         {
             return 400;
         }
 
         const auto &input_bearings = route_parameters.bearings;
-        if (input_bearings.size() > 0 && route_parameters.coordinates.size() != input_bearings.size())
+        unsigned nb_coordinates = useSameTgtSrc ? route_parameters.coordinates.size() : (route_parameters.destinations.size() + route_parameters.sources.size());
```
TheMarex: 1. Don't use abbreviations in variable names.
2. Expected logic would be: `auto number_of_coordinates = route_parameters.coordinates.size() + route_parameters.sources.size() + route_parameters.destinations.size();`

data_structures/route_parameters.cpp
frodrigo: This as option was made to avoid network transfer of data, if not needed by client. With huge matrix in mind.

plugins/distance_table.hpp
```diff
@@ -70,24 +70,34 @@ template <class DataFacadeT> class DistanceTablePlugin final : public BasePlugin
...
+        if (input_bearings.size() > 0 && nb_coordinates != input_bearings.size())
         {
             json_result.values["status"] = "Number of bearings does not match number of coordinates .";
             return 400;
         }
 
         const bool checksum_OK = (route_parameters.check_sum == facade->GetCheckSum());
-        unsigned max_locations =
+        unsigned max_locations = useSameTgtSrc ?
```
TheMarex: I guess it would be cleaner to just return `400` for tables that exceed a size of `max_locations_distance_table*max_locations_distance_table` so this should be more like:

```c++
if (number_of_sources*number_of_destinations > max_locations_distance_table*max_locations_distance_table)
{
    json_result.values["status_message"] = std::string("Table exceeds maximum number of entries of ") + std::to_string(max_locations_distance_table*max_locations_distance_table);
    return 400;
}
```

Where `auto number_of_sources = route_parameters.coordinates.size() + route_parameters.sources.size()` and `auto number_of_destinations = route_parameters.coordinates.size() + route_parameters.destinations.size();`

plugins/distance_table.hpp
```diff
@@ -70,24 +70,34 @@ template <class DataFacadeT> class DistanceTablePlugin final : public BasePlugin
...
+            std::min(static_cast<unsigned>(max_locations_distance_table),
+                     static_cast<unsigned>(route_parameters.coordinates.size())) :
             std::min(static_cast<unsigned>(max_locations_distance_table),
-                     static_cast<unsigned>(route_parameters.coordinates.size()));
+                     static_cast<unsigned>(route_parameters.destinations.size()));
 
-        PhantomNodeArray phantom_node_vector(max_locations);
+        PhantomNodeArray phantom_node_target_vector(max_locations);
```
TheMarex: `max_locations` is wrong here as it would be `number_of_destinations`.

plugins/distance_table.hpp
```diff
@@ -70,24 +70,34 @@ template <class DataFacadeT> class DistanceTablePlugin final : public BasePlugin
...
         for (const auto i : osrm::irange(0u, max_locations))
```
TheMarex: You can move this loop to a utility function and like this:
```c++
GetCoordinates(const std::vector<FixedPointCoordinate>& inputs, PhantomNodeArray& output)
{
....
}
```

The code should then look somethings like:
```c++
GetCoordinates(route_parameters.coordinates, phantom_node_vector);
std::insert(phantom_node_source_vector.end(), phantom_node_vector.begin(), phantom_node_vector.end());
std::insert(phantom_node_destination_vector.end(), phantom_node_vector.begin(), phantom_node_vector.end());
GetCoordinates(route_parameters.sources, phantom_node_source_vector);
GetCoordinates(route_parameters.destinations, phantom_node_destination_vector);

File: plugins/distance_table.hpp [hunk 7]
```diff
@@ -97,21 +107,49 @@ template <class DataFacadeT> class DistanceTablePlugin final : public BasePlugin
...
+                                     static_cast<unsigned>(route_parameters.sources.size()));
+        }
+        PhantomNodeArray phantom_node_source_vector(max_locations);
+        for (const auto i : osrm::irange(0u, max_locations))
+        {
+            if (checksum_OK && i < route_parameters.hints.size() - shift_coordinates &&
+                !route_parameters.hints[i + shift_coordinates].empty())
+            {
+                PhantomNode current_phantom_node;
+                ObjectEncoder::DecodeFromBase64(route_parameters.hints[i + shift_coordinates], current_phantom_node);
```
TheMarex: We sadly need to drop `hint` support here. `5.0.0` will restructure the API a little, I think we can live with this for one release cycle.

plugins/distance_table.hpp
```diff
@@ -119,17 +157,41 @@ template <class DataFacadeT> class DistanceTablePlugin final : public BasePlugin
...
-            auto row_begin_iterator = result_table->begin() + (row * number_of_locations);
-            auto row_end_iterator = result_table->begin() + ((row + 1) * number_of_locations);
+            auto row_begin_iterator = result_table->begin() + (row * number_of_targets);
+            auto row_end_iterator = result_table->begin() + ((row + 1) * number_of_targets);
             json_row.values.insert(json_row.values.end(), row_begin_iterator, row_end_iterator);
-            json_array.values.push_back(json_row);
+            matrix_json_array.values.push_back(json_row);
+        }
+        json_result.values["distance_table"] = matrix_json_array;
+        if (route_parameters.mapped_points) {
```
TheMarex: Remove if, since the parameters needs to be dropped. (Code in general is fine.) I would go with `source_coordinates` and `target_coordinates` here.

plugins/plugin_base.hpp
```diff
@@ -45,9 +45,9 @@ class BasePlugin
     virtual const std::string GetDescriptor() const = 0;
     virtual int HandleRequest(const RouteParameters &, osrm::json::Object &) = 0;
     virtual bool
-    check_all_coordinates(const std::vector<FixedPointCoordinate> &coordinates) const final
+    check_all_coordinates(const std::vector<FixedPointCoordinate> &coordinates, const unsigned min = 2) const final
```
TheMarex: Remove the min check here. Should be handled on caller side.

routing_algorithms/many_to_many.hpp
```diff
@@ -67,11 +67,12 @@ class ManyToManyRouting final
     ~ManyToManyRouting() {}
 
     std::shared_ptr<std::vector<EdgeWeight>>
-    operator()(const PhantomNodeArray &phantom_nodes_array) const
+    operator()(const PhantomNodeArray &phantom_targets_array, const PhantomNodeArray &phantom_sources_array = PhantomNodeArray(0)) const
```
TheMarex: Don't provide default values here. Parameter order should be `const PhantomNodeArray& sources, const PhantomNodeArray& targets`.

routing_algorithms/many_to_many.hpp
```diff
@@ -113,7 +114,7 @@ class ManyToManyRouting final
 
         // for each source do forward search
         unsigned source_id = 0;
-        for (const std::vector<PhantomNode> &phantom_node_vector : phantom_nodes_array)
+        for (const std::vector<PhantomNode> &phantom_node_vector : (phantom_sources_array.size() ? phantom_sources_array : phantom_targets_array))
```
TheMarex: This should not have a switch. Always use `phantom_sources_array`. Caller needs to take care to populate both arrays accordingly.

data_structures/route_parameters.cpp
TheMarex: For really huge matrices you should avoid going over the network (uncompressed) anyway. You can take a look at our `node-osrm` bindings. Since this is all 100% numerical values you can get a way with about 4 bytes per entry (just save the raw matrix and the size).

TheMarex:
@fab-girard Added a few notes. Let me know if you have time to address this, otherwise I'll pull your changes in a new branch and spend some time tomorrow to address them.

TheMarex:
I'm going to grab this. Opened a new PR on a branch in this repository.

fab-girard:
Thank you @TheMarex to have included the PR in 4.9.0 !
If I understood, you have updated my commit to take into account your review.

I've updated the wiki with your last changes in 4.9.0: https://github.com/Project-OSRM/osrm-backend/wiki/Server-api

Fab

PR #1774, opened 2015-11-13
Add a summary to each match result

danpat:
Useful for comparing real-world data against the OSRM internal edge weighting.


TheMarex:
Looks good to me. Merge away. :+1:

PR #1723, opened 2015-10-06
New Travis CI integration

daniel-j-h:
Dummy pull request for trying out the new Travis CI integration. Details follow.

PR #1775, opened 2015-11-16
Use new install script for cmake. Upgrades to 3.4.0

TheMarex:


daniel-j-h:
Instead of an additional level of indirection to some dubious gists, what about just writing the dependency resolution into the `.travis.config`, like I started it over at this pull request: https://github.com/Project-OSRM/osrm-backend/pull/1723/files#diff-354f30a63fb0907d4ad57269548329e3R139

TheMarex:
Obsolete.

PR #1782, opened 2015-11-18
Fix pkgconfig cmake template

fab-girard:
Hello,
I've fixed pkgconfig.in template file in order to be able to use `node-pre-gyp --build-from-source` in node-osrm project.
'osrm'  is still referenced in includes from node-osrm poject:
- json_v8_renderer.hpp
- node_osrm.cpp

Fabien


TheMarex:
@fab-girard if you rebase on develop this should pass the builds and can be merged. :+1:

danpat:
There's a test in `develop` that's currently not passing, but if it's the only thing that fails, we should be good.

fab-girard:
@TheMarex: done ;)

PR #1783, opened 2015-11-18
Preserve information about big components and exit early

TheMarex:
This uses a bit flag to differenciate between small and big components
and keeps the ids for both. This makes it easy to decide if two
coordinates are on different continets.

### DIFF
--- a/data_structures/phantom_node.hpp
+++ b/data_structures/phantom_node.hpp
@@ -47,6 +47,7 @@ struct PhantomNode
                 int forward_offset,
                 int reverse_offset,
                 unsigned packed_geometry_id,
+                bool is_tiny_component,
                 unsigned component_id,
                 FixedPointCoordinate &location,
                 unsigned short fwd_segment_position,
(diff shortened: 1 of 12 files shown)


data_structures/phantom_node.hpp
```diff
@@ -85,7 +88,11 @@ struct PhantomNode
     int forward_offset;
     int reverse_offset;
     unsigned packed_geometry_id;
-    unsigned component_id;
+    struct ComponentType {
+        uint32_t id : 31;
+        bool is_tiny : 1;
+    } component;
+    static_assert(sizeof(ComponentType) == 4, "ComponentType needs to 4 bytes big");
```
TheMarex: @daniel-j-h any idea how this can possibly fail on VS2015?

daniel-j-h: Implementation defined size, since this is a struct with bit fields §9.6 Bit-fields N3936

PR #1785, opened 2015-11-18
[not ready] Adds ability to dump GeoJSON with detailed edge weight data

danpat:
Useful for debugging.  

Adds a new compile-time flag to enable:  `cmake -DDEBUG_GEOMETRY=ON`

This unlocks two new runtime options.
For `osrm-extract`, it's `--debug-turns <filename>`
For `osrm-prepare`, it's `--debug-geometry <filename>`

This will give you point data for turns (with heading and value meta-data in the `properties` of each GeoJSON), and lines for each edge, with weight data attached.  There is one two-point line segment for every node-based edge in the graph, and they are not bidirectional (i.e. you get two for each one-way street, oriented in the direction of travel).

Still todo:
 - [x] Move code into utility functions to minimize the `#ifdef` impact

### DIFF
--- a/build-local.bat
+++ b/build-local.bat
@@ -11,7 +11,8 @@ SET CONFIGURATION=Release
 FOR /F "tokens=*" %%i in ('git rev-parse --abbrev-ref HEAD') do SET APPVEYOR_REPO_BRANCH=%%i
 ECHO APPVEYOR_REPO_BRANCH^: %APPVEYOR_REPO_BRANCH%
 
-SET PATH=C:\mb\windows-builds-64\tmp-bin\cmake-3.1.0-win32-x86\bin;%PATH%
+::SET PATH=C:\mb\windows-builds-64\tmp-bin\cmake-3.1.0-win32-x86\bin;%PATH%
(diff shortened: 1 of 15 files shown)


build-local.bat
```diff
@@ -11,7 +11,8 @@ SET CONFIGURATION=Release
 FOR /F "tokens=*" %%i in ('git rev-parse --abbrev-ref HEAD') do SET APPVEYOR_REPO_BRANCH=%%i
 ECHO APPVEYOR_REPO_BRANCH^: %APPVEYOR_REPO_BRANCH%
 
-SET PATH=C:\mb\windows-builds-64\tmp-bin\cmake-3.1.0-win32-x86\bin;%PATH%
+::SET PATH=C:\mb\windows-builds-64\tmp-bin\cmake-3.1.0-win32-x86\bin;%PATH%
```
TheMarex: remove commented out clines

contractor/contractor.hpp
```diff
@@ -55,6 +55,14 @@ SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
 class Contractor
 {
+  public:
```
TheMarex: How did this happen? Should not be part of the PR.

PR #1787, opened 2015-11-19
Add pre-turn bearing as last field to instruction

TheMarex:


data_structures/segment_information.hpp
```diff
@@ -43,7 +43,8 @@ struct SegmentInformation
     NodeID name_id;
     EdgeWeight duration;
     float length;
-    short bearing; // more than enough [0..3600] fits into 12 bits
+    short pre_turn_bearing; // more than enough [0..3600] fits into 12 bits
+    short post_turn_bearing;
     TurnInstruction turn_instruction;
```
danpat: Should we start being more explicit about data sizes in cases where we know the range is fixed?  `int16_t` here?

Issue #322, opened 2012-07-04
64bit node / edge ids


karme:
have seen some int overflow trying a world import using nearly all tracks

is the code ready for 64Bit node / edge ids? (is it sufficient to adjust typedefs.h)?

PS: i first thought there was already an issue for that one, but I didn't find it - sorry if i missed it.

4x4falcon:
As I'm having problems with id's > 32bit i gather that this is still open.

I have data with node id's commencing at 1000000000000 (decimal) and osrm-extract crashes, I'm assuming that this is the problem as I have not run with gdb.

I've looked at the code and changed typedefs.h to unsigned long however this creates lots of other errors when compiling, which I have not sorted through yet.

Are there any plans to change osrm to 64 bit ids. I know there becomes memory issues but osm data is about 12months away from going over 4294967295 based on current id of 3727297211 and that it's been 18 months since passing node id 2147483647.

If you can point me to all places that need changing would be great as it's hard to work out code that has virtually no comments, by just sorting through make errors.

Otherwise great work.

TheMarex:
Yes this is still an issues. Relevant parts of the code are:

- The node id list https://github.com/Project-OSRM/osrm-backend/blob/bb1428eeb14889890977d514dee84b78d400acb0/extractor/extraction_containers.hpp#L70 
- The remapping from OSM ids to internal ids https://github.com/Project-OSRM/osrm-backend/blob/bb1428eeb14889890977d514dee84b78d400acb0/extractor/extraction_containers.hpp#L70
- The temporary edges that contain references to OSM ids https://github.com/Project-OSRM/osrm-backend/blob/bb1428eeb14889890977d514dee84b78d400acb0/extractor/internal_extractor_edge.hpp#L71-L72

The important thing to note is that we only need 64bit for _external_ OSM ids, we renumber them anyway in `osrm-extract` since we only use a small number of them. So we need to take care to not confuse them in `osrm-extract` (and only use 64bit where necessary, otherwise memory usage might explode)

@daniel-j-h want to take a look at this?

4x4falcon:
Thanks.  I'll see what I can find to get it to work.

Really like the routing and interface work just would like to get it to work with our data.

danpat:
Done in 9005fe2f61aa5478c5bf0bef5ec76843af04f93e.

Note that this is just support for 64bit *node* IDs, not way or relation IDs.  It's also *not* support for >2^32 actual nodes, just node ID values higher than 2^32.

PR #1793, opened 2015-11-26
Add support for 64bit OSM node identifiers

danpat:
This PR is to address https://github.com/Project-OSRM/osrm-backend/issues/322

It adds support for 64bit OSM node id values.  Without this change, OSM id values greater than 2^32 will overflow, and potentially conflict with other values.  I.e. currently, OSRM sees 4294967296 as the same as 1.  We expect to hit the unsigned 2^32 ceiling in the next few months.

This adds some overhead during processing: we need extra space to store the 64bit identifiers.  The main increases are:

  - during `osrm-extract`, the `all_nodes_list` and `used_nodes_list` will double in size.  These are both STXXL vectors, so the main impact will be on disk space and I/O during sorting.

  - the same applies for turn restrictions - we will now hold 64bit IDs in memory instead of 32bit IDs until we can re-number the identifiers.

Run-time memory performance for `osrm-routed` and `osrm-prepare` should be relatively unchanged.  By this stage of the pipeline, we are using mostly internal 32bit identifiers.  There will be some additional I/O when loading the `.nodes` file, because it contains the original OSM 64bit IDs (this file will double in size), but this cost should be relatively small.

Still TODO:

 - [ ] 64bit ID test - I have manually constructed a test file that demonstrates the conflicting IDs, but I still need to integrate it into the test suite.

Because of the ubiquitous use of the `NodeID` type, I implemented a strictly typed (based on `BOOST_STRICT_TYPEDEF`) `NodeID64` which does not allow automatic type conversion.  This should help avoid any type confusion in the future, as we'll get compiler errors if someone tries to assign a NodeID to a NodeID64, or vice-versa.  At least with GCC4.9 and `clang` from XCode 7 on OSX, the strict typedef object is fully optimized away under `-O3`, so it incurs no runtime overhead for production systems.

### DIFF
--- a/data_structures/external_memory_node.cpp
+++ b/data_structures/external_memory_node.cpp
@@ -31,7 +31,7 @@ SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 #include <limits>
 
 ExternalMemoryNode::ExternalMemoryNode(
-    int lat, int lon, unsigned int node_id, bool barrier, bool traffic_lights)
+    int lat, int lon, OSMNodeID node_id, bool barrier, bool traffic_lights)
     : QueryNode(lat, lon, node_id), barrier(barrier), traffic_lights(traffic_lights)
 {
 }
@@ -40,13 +40,13 @@ ExternalMemoryNode::ExternalMemoryNode() : barrier(false), traffic_lights(false)
 
 ExternalMemoryNode ExternalMemoryNode::min_value()
 {
-    return ExternalMemoryNode(0, 0, 0, false, false);
+    return ExternalMemoryNode(0, 0, MIN_OSM_NODEID, false, false);
 }
 
 ExternalMemoryNode ExternalMemoryNode::max_value()
 {
     return ExternalMemoryNode(std::numeric_limits<int>::max(), std::numeric_limits<int>::max(),
-                              std::numeric_limits<unsigned>::max(), false, false);
+                              MAX_OSM_NODEID, false, false);
 }
 
 bool ExternalMemoryNodeSTXXLCompare::operator()(const ExternalMemoryNode &left,
--- a/data_structures/node_based_graph.hpp
+++ b/data_structures/node_based_graph.hpp
@@ -75,7 +75,7 @@ using NodeBasedDynamicGraph = DynamicGraph<NodeBasedEdgeData>;
 /// The since DynamicGraph expects directed edges, we need to insert
 /// two edges for undirected edges.
 inline std::shared_ptr<NodeBasedDynamicGraph>
-NodeBasedDynamicGraphFromEdges(int number_of_nodes, const std::vector<NodeBasedEdge> &input_edge_list)
+NodeBasedDynamicGraphFromEdges(std::size_t number_of_nodes, const std::vector<NodeBasedEdge> &input_edge_list)
 {
     auto edges_list = directedEdgesFromCompressed<NodeBasedDynamicGraph::InputEdge>(input_edge_list,
         [](NodeBasedDynamicGraph::InputEdge& output_edge, const NodeBasedEdge& input_edge)
--- a/data_structures/node_id.hpp
+++ b/data_structures/node_id.hpp
@@ -32,10 +32,10 @@ SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
 struct Cmp
 {
-    using value_type = NodeID;
-    bool operator()(const NodeID left, const NodeID right) const { return left < right; }
-    value_type max_value() { return 0xffffffff; }
-    value_type min_value() { return 0x0; }
+    using value_type = OSMNodeID;
+    bool operator()(const value_type left, const value_type right) const { return left < right; }
+    value_type max_value() { return MAX_OSM_NODEID; }
+    value_type min_value() { return MIN_OSM_NODEID; }
 };
 
 #endif // NODE_ID_HPP
--- a/data_structures/query_node.hpp
+++ b/data_structures/query_node.hpp
@@ -38,32 +38,32 @@ SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
 struct QueryNode
 {
-    using key_type = NodeID; // type of NodeID
+    using key_type = OSMNodeID; // type of NodeID
     using value_type = int;  // type of lat,lons
 
-    explicit QueryNode(int lat, int lon, NodeID node_id) : lat(lat), lon(lon), node_id(node_id) {}
+    explicit QueryNode(int lat, int lon, OSMNodeID node_id) : lat(lat), lon(lon), node_id(node_id) {}
     QueryNode()
         : lat(std::numeric_limits<int>::max()), lon(std::numeric_limits<int>::max()),
-          node_id(std::numeric_limits<unsigned>::max())
+          node_id(SPECIAL_OSM_NODEID)
     {
     }
 
     int lat;
     int lon;
-    NodeID node_id;
+    OSMNodeID node_id;
 
     static QueryNode min_value()
     {
         return QueryNode(static_cast<int>(-90 * COORDINATE_PRECISION),
                          static_cast<int>(-180 * COORDINATE_PRECISION),
-                         std::numeric_limits<NodeID>::min());
+                         MIN_OSM_NODEID);
     }
 
     static QueryNode max_value()
     {
         return QueryNode(static_cast<int>(90 * COORDINATE_PRECISION),
                          static_cast<int>(180 * COORDINATE_PRECISION),
-                         std::numeric_limits<NodeID>::max());
+                         MAX_OSM_NODEID);
     }
 
     value_type operator[](const std::size_t n) const
--- a/data_structures/restriction_map.cpp
+++ b/data_structures/restriction_map.cpp
@@ -33,10 +33,16 @@ RestrictionMap::RestrictionMap(const std::vector<TurnRestriction> &restriction_l
     // a pair of starting edge and a list of all end nodes
     for (auto &restriction : restriction_list)
     {
+        // This downcasting is OK because when this is called, the node IDs have been
+        // renumbered into internal values, which should be well under 2^32
+        // This will be a problem if we have more than 2^32 actual restrictions
+        BOOST_ASSERT(restriction.from.node < std::numeric_limits<NodeID>::max());
+        BOOST_ASSERT(restriction.via.node < std::numeric_limits<NodeID>::max());
         m_restriction_start_nodes.insert(restriction.from.node);
         m_no_turn_via_node_set.insert(restriction.via.node);
 
-        RestrictionSource restriction_source = {restriction.from.node, restriction.via.node};
+        // This explicit downcasting is also OK for the same reason.
+        RestrictionSource restriction_source = {static_cast<NodeID>(restriction.from.node), static_cast<NodeID>(restriction.via.node)};
 
         std::size_t index;
         auto restriction_iter = m_restriction_map.find(restriction_source);
@@ -62,6 +68,7 @@ RestrictionMap::RestrictionMap(const std::vector<TurnRestriction> &restriction_l
             }
         }
         ++m_count;
+        BOOST_ASSERT(restriction.to.node < std::numeric_limits<NodeID>::max());
         m_restriction_bucket_list.at(index)
             .emplace_back(restriction.to.node, restriction.flags.is_only);
     }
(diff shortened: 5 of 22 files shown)


data_structures/node_based_graph.hpp
```diff
@@ -75,7 +75,7 @@ using NodeBasedDynamicGraph = DynamicGraph<NodeBasedEdgeData>;
 /// The since DynamicGraph expects directed edges, we need to insert
 /// two edges for undirected edges.
 inline std::shared_ptr<NodeBasedDynamicGraph>
-NodeBasedDynamicGraphFromEdges(int number_of_nodes, const std::vector<NodeBasedEdge> &input_edge_list)
+NodeBasedDynamicGraphFromEdges(std::size_t number_of_nodes, const std::vector<NodeBasedEdge> &input_edge_list)
```
MoKob: missing cstddef for std::int (@daniel-j-h are you proud of me yet?)

data_structures/node_id.hpp
```diff
@@ -32,10 +32,10 @@ SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
 struct Cmp
 {
-    using value_type = NodeID;
-    bool operator()(const NodeID left, const NodeID right) const { return left < right; }
-    value_type max_value() { return 0xffffffff; }
-    value_type min_value() { return 0x0; }
+    using value_type = NodeID64;
+    bool operator()(const value_type left, const value_type right) const { return left < right; }
+    value_type max_value() { return INVALID_NODEID64; }
```
MoKob: Could lead to confusion, errors down the line. Is INVALID_NODEID64 always going to be the maximum?

data_structures/query_node.hpp
```diff
@@ -38,32 +38,32 @@ SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
...
-                         std::numeric_limits<NodeID>::min());
+                         NodeID64(0));
     }
 
     static QueryNode max_value()
     {
         return QueryNode(static_cast<int>(90 * COORDINATE_PRECISION),
                          static_cast<int>(180 * COORDINATE_PRECISION),
-                         std::numeric_limits<NodeID>::max());
+                         INVALID_NODEID64);
```
MoKob: see above

data_structures/restriction_map.cpp
```diff
@@ -33,10 +33,14 @@ RestrictionMap::RestrictionMap(const std::vector<TurnRestriction> &restriction_l
     // a pair of starting edge and a list of all end nodes
     for (auto &restriction : restriction_list)
     {
+        // This downcasting is OK because when this is called, the node IDs have been
+        // renumbered into internal values, which should be well under 2^32
+        // This will be a problem if we have more than 2^32 actual restrictions
         m_restriction_start_nodes.insert(restriction.from.node);
         m_no_turn_via_node_set.insert(restriction.via.node);
```
MoKob: Down cast should be asserted to be correct

extractor/edge_based_graph_factory.cpp
```diff
@@ -518,6 +518,7 @@ void EdgeBasedGraphFactory::GenerateEdgeExpandedEdges(
                 BOOST_ASSERT(SPECIAL_NODEID != edge_data1.edge_id);
                 BOOST_ASSERT(SPECIAL_NODEID != edge_data2.edge_id);
 
+                // NOTE: potential overflow here if we hit 2^32 routable edges
                 m_edge_based_edge_list.emplace_back(edge_data1.edge_id, edge_data2.edge_id,
```
MoKob: assert that the space is enough

extractor/extraction_containers.cpp
```diff
@@ -217,21 +221,21 @@ void ExtractionContainers::PrepareEdges(lua_State *segment_state)
 
     while (edge_iterator != all_edges_list_end && node_iterator != all_nodes_list_end)
     {
-        if (edge_iterator->result.source < node_iterator->node_id)
+        if (edge_iterator->result.osm_source_id < node_iterator->node_id)
         {
             SimpleLogger().Write(LogLevel::logWARNING) << "Found invalid node reference " << edge_iterator->result.source;
             edge_iterator->result.source = SPECIAL_NODEID;
```
MoKob: SPECIAL_NODEID is an invalid node, is it? Would love it, if we could use consistent names here (just like INVALID_NODEID64)

extractor/extraction_containers.cpp
```diff
@@ -464,7 +469,7 @@ void ExtractionContainers::PrepareEdges(lua_State *segment_state)
 
 void ExtractionContainers::WriteEdges(std::ofstream& file_out_stream) const
 {
-    std::cout << "[extractor] Writing used egdes       ... " << std::flush;
+    std::cout << "[extractor] Writing used edges       ... " << std::flush;
     TIMER_START(write_edges);
     // Traverse list of edges and nodes in parallel and set target coord
     unsigned number_of_used_edges = 0;
```
MoKob: would prefer a 64 bit type here for the moment when we actually have more than 2^32 edges, or at least an assert to make sure we do not miss that type.

extractor/first_and_last_segment_of_way.hpp
```diff
@@ -58,19 +59,15 @@ struct FirstAndLastSegmentOfWay
...
-                std::numeric_limits<NodeID>::max(),
-                std::numeric_limits<NodeID>::max(),
-                std::numeric_limits<NodeID>::max(),
-                std::numeric_limits<NodeID>::max()};
+        return {INVALID_OSM_WAYID,
+                INVALID_NODEID64,
+                INVALID_NODEID64,
+                INVALID_NODEID64,
+                INVALID_NODEID64};
     }
```
MoKob: as above, would prefer an additional definition for MAX_ID or keep usage of numeric_limits. INVALID just looks wrong to me in a max statement.

extractor/internal_extractor_edge.hpp
```diff
@@ -101,19 +101,36 @@ struct InternalExtractorEdge
...
-        return InternalExtractorEdge(0, 0, 0, WeightData(), false, false, false,
+        return InternalExtractorEdge(NodeID64(0), NodeID64(0), 0, WeightData(), false, false, false,
                                      false, TRAVEL_MODE_INACCESSIBLE, false);
     }
-    static InternalExtractorEdge max_value()
+    static InternalExtractorEdge max_osm_value()
     {
-        return InternalExtractorEdge(SPECIAL_NODEID, SPECIAL_NODEID, 0, WeightData(), false,
+        return InternalExtractorEdge(INVALID_NODEID64, INVALID_NODEID64, 0, WeightData(), false,
                                      false, false, false, TRAVEL_MODE_INACCESSIBLE, false);
```
MoKob: s.a. max vs invalid

typedefs.h
```diff
@@ -38,6 +39,16 @@ SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 #endif
 #endif
 
+// OpenStreetMap node ids are higher than 2^32
+OSRM_STRONG_TYPEDEF(uint64_t, NodeID64)
+OSRM_STRONG_TYPEDEF(uint32_t, OSMWayID)
```
MoKob: how high are the osm way ids? shouldn't we experience similar problems there as well?

typedefs.h
```diff
@@ -38,6 +39,16 @@ SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
...
+
+static const NodeID64 INVALID_NODEID64 = NodeID64(std::numeric_limits<uint64_t>::max());
+static const OSMWayID INVALID_OSM_WAYID = OSMWayID(std::numeric_limits<uint32_t>::max());
+
+using NodeID64_weak = uint64_t;
+using EdgeID64_weak = uint64_t;
+
```
MoKob: EdgeID64_weak is 64 bit and OSMWayID is 32?

typedefs.h
```diff
@@ -38,6 +39,16 @@ SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 #endif
 #endif
 
+// OpenStreetMap node ids are higher than 2^32
+OSRM_STRONG_TYPEDEF(uint64_t, NodeID64)
```
TheMarex: Instead of `NodeID64` we should use `OSMNodeID`.

typedefs.h
```diff
@@ -38,6 +39,16 @@ SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
...
+OSRM_STRONG_TYPEDEF(uint32_t, OSMWayID)
+
+static const NodeID64 INVALID_NODEID64 = NodeID64(std::numeric_limits<uint64_t>::max());
+static const OSMWayID INVALID_OSM_WAYID = OSMWayID(std::numeric_limits<uint32_t>::max());
+
+using NodeID64_weak = uint64_t;
```
TheMarex: should be `WeakOSMNodeId`

TheMarex:
In general this looks good. Added a few comments on the diff. Otherwise this should be good to merge. :+1:

While the extension of `QueryNode` is problematic, it has no impact on the server memory usage (as it is read object-wise and only the `lat`/`lon` is used). We should think about re-working our data format (there is a weired disconnect between what data is written to disk and what is actually used later by the server component).

typedefs.h
daniel-j-h: should be fine

http://wiki.openstreetmap.org/wiki/Stats#Nodes.2C_ways_and_relations

typedefs.h
daniel-j-h: If possible, set the fixed integer types once and then use type inference, so that in case you change the types, the code will adapt itself, like:

    std::numeric_limits<NodeID64>::max()

    std::numeric_limits<decltype(node)>::max()

data_structures/external_memory_node.cpp
```diff
@@ -40,13 +40,13 @@ ExternalMemoryNode::ExternalMemoryNode() : barrier(false), traffic_lights(false)
...
 {
-    return ExternalMemoryNode(0, 0, 0, false, false);
+    return ExternalMemoryNode(0, 0, NodeID64(0), false, false);
 }
 
 ExternalMemoryNode ExternalMemoryNode::max_value()
 {
     return ExternalMemoryNode(std::numeric_limits<int>::max(), std::numeric_limits<int>::max(),
-                              std::numeric_limits<unsigned>::max(), false, false);
+                              NodeID64(std::numeric_limits<unsigned>::max()), false, false);
```
joto: This looks strange. Why is max of `unsigned` used here? Shouldn't this be max of whatever is inside the `NodeID64`? Maybe have a `static const NodeID64::max_value;` in the `NodeID64` or something like that?

typedefs.h
joto: There are only about 320 million ways in OSM currently, so no problem for a while.

extractor/extraction_containers.cpp
```diff
@@ -288,20 +293,20 @@ void ExtractionContainers::PrepareEdges(lua_State *segment_state)
             continue;
         }
 
-        if (edge_iterator->result.target < node_iterator->node_id)
+        if (edge_iterator->result.osm_target_id < node_iterator->node_id)
         {
-            SimpleLogger().Write(LogLevel::logWARNING) << "Found invalid node reference " << edge_iterator->result.target;
+            SimpleLogger().Write(LogLevel::logWARNING) << "Found invalid node reference " << NodeID64_to_uint64_t(edge_iterator->result.osm_target_id);
```
joto: Where is `NodeID64_to_uint64_t()` defined? And wouldn't it be better to implement a conversion operator so we can write `uint64_t(...)`?

extractor/extraction_containers.cpp
```diff
@@ -479,6 +484,8 @@ void ExtractionContainers::WriteEdges(std::ofstream& file_out_stream) const
             continue;
         }
 
+        // IMPORTANT!!!!
+        // Here, we depend on slicing to write out the base class (not NodeBasedEdgeWithOSM)
         file_out_stream.write((char*) &edge.result, sizeof(NodeBasedEdge));
         number_of_used_edges++;
```
daniel-j-h: Hmm slicing is something different. Slicing is when you have `base` and `child`, and instead of binding a child to a base with references or pointers for polymorphism like `base& myBase = myChild;` you accidentally say `base b = myChild;`, then `b` really is a `base` slicing off its `child` part.

What you are doing here is just writing out the first n bytes of the base, hoping the memory layout is base-first-then-child. I haven't looked deeper into it yet, but I can show you at least a counterexample, of when this breaks:

https://gist.github.com/daniel-j-h/71e33b698234e3deada2#file-virtual-txt-L33

I also think you can guard against this via static asserting on [std::is_polymorphic](http://en.cppreference.com/w/cpp/types/is_polymorphic).

It's probably a good idea to encapsulate and hide all of this.

extractor/extraction_containers.cpp
danpat: Yeah, agreed SPECIAL is a strange choice.  To keep things consistent for this PR, I've renamed INVALID_OSM_NODEID to SPECIAL_OSM_NODEID.  This at least gets things consistent, and doesn't pollute this PR with a bunch of unrelated changes.  We can fix the SPECIAL/INVALID naming separately.

extractor/extraction_containers.cpp
danpat: I changed this to create a copy of the base class, and *really* depend on slicing to get just the data we want.

Issue #1546, opened 2015-06-24
Loop between intermediate points


etorrano:
I wrote the same issue in bug #1424  that seems to be fixed, but in the following case, that appeared as an example of an U-turn problem, now appears a loop instead of the U-turn :

![30b85f12-0096-11e5-912e-80be204d6e40](https://cloud.githubusercontent.com/assets/9082428/8060539/586a4ea8-0e9c-11e5-9458-ef58cdbbd4b8.PNG)

http://osrm.at/cQt

![ruteos0](https://cloud.githubusercontent.com/assets/9082428/8062206/e1661342-0ea4-11e5-8cfb-c6472ebec3ce.png)

When I put an intermediate point in order to get a better route I get this:

http://osrm.at/cQu

![ruteos1](https://cloud.githubusercontent.com/assets/9082428/8062207/e188470a-0ea4-11e5-9572-17ed778e59f9.png)

I think this is the right route.

Andygol:
@TheMarex 
May there is the same issue like reported above. I found it in ukrainian branch of the [forum.osm.org](http://forum.openstreetmap.org/viewtopic.php?pid=562312#p562312)

All depends on the angle of the line, if the angle is obtuse, the route goes around the circle, and if the angle is acute - all right

**obtuse angle**
![open source routing machine demo 2015-11-20 16-50-07](https://cloud.githubusercontent.com/assets/369696/11302678/748daa44-8fa7-11e5-8ca6-b70ced302d93.png)

**sharp angle**
![open source routing machine demo 2015-11-20 16-57-28](https://cloud.githubusercontent.com/assets/369696/11302779/fcebe126-8fa7-11e5-9284-54f9d10f98d9.png)



![osrm-weird-behavior](https://cloud.githubusercontent.com/assets/369696/11302144/49f21da4-8fa4-11e5-8cd4-f2a1c123c29f.gif)
http://map.project-osrm.org/?z=18&center=50.402454%2C30.485365&loc=50.403225%2C30.485097&loc=50.401926%2C30.486749&loc=50.401744%2C30.483579&hl=en&ly=&alt=&df=&srv=

TheMarex:
Fixed on the demo server.

etorrano:
I tested the bug through the links used above, in order to detect it, and it is all ok, thanks for the fix!

PR #1795, opened 2015-12-01
Refactoring shortest path/direct shortest path queries

TheMarex:
This pull request addresses https://github.com/Project-OSRM/osrm-backend/issues/1546. In the process I had to structurally change `shortest_path` and `direct_shortest_path` and used it as an opportunity for a refactor.

/cc @danpat for code review

[Review on Reviewable](https://reviewable.io/reviews/project-osrm/osrm-backend/1795)

### DIFF
--- a/routing_algorithms/direct_shortest_path.hpp
+++ b/routing_algorithms/direct_shortest_path.hpp
@@ -64,143 +64,39 @@ class DirectShortestPathRouting final
     {
         (void)uturn_indicators; // unused
 
-        engine_working_data.InitializeOrClearFirstThreadLocalStorage(
-            super::facade->GetNumberOfNodes());
-        engine_working_data.InitializeOrClearSecondThreadLocalStorage(
-            super::facade->GetNumberOfNodes());
-
-        QueryHeap &forward_heap = *(engine_working_data.forward_heap_1);
-        QueryHeap &reverse_heap = *(engine_working_data.reverse_heap_1);
-
-        QueryHeap &forward_core_heap = *(engine_working_data.forward_heap_2);
-        QueryHeap &reverse_core_heap = *(engine_working_data.reverse_heap_2);
-
         // Get distance to next pair of target nodes.
         BOOST_ASSERT_MSG(1 == phantom_nodes_vector.size(),
                                          "Direct Shortest Path Query only accepts a single source and target pair. Multiple ones have been specified.");
-
         const auto& phantom_node_pair = phantom_nodes_vector.front();
 
+        engine_working_data.InitializeOrClearFirstThreadLocalStorage(
+            super::facade->GetNumberOfNodes());
+        QueryHeap &forward_heap = *(engine_working_data.forward_heap_1);
+        QueryHeap &reverse_heap = *(engine_working_data.reverse_heap_1);
         forward_heap.Clear();
         reverse_heap.Clear();
-        int distance = INVALID_EDGE_WEIGHT;
-        NodeID middle = SPECIAL_NODEID;
 
-        const EdgeWeight min_edge_offset =
-            std::min(-phantom_node_pair.source_phantom.GetForwardWeightPlusOffset(),
-                     -phantom_node_pair.source_phantom.GetReverseWeightPlusOffset());
+        int distance;
+        std::vector<NodeID> packed_leg;
 
-        // insert new starting nodes into forward heap, adjusted by previous distances.
-        if (phantom_node_pair.source_phantom.forward_node_id != SPECIAL_NODEID)
+        if (super::facade->GetCoreSize() > 0)
         {
-            forward_heap.Insert(
-                phantom_node_pair.source_phantom.forward_node_id,
-                -phantom_node_pair.source_phantom.GetForwardWeightPlusOffset(),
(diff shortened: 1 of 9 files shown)


daniel-j-h:
Reviewed 9 of 9 files at r1.
Review status: all files reviewed at latest revision, 4 unresolved discussions, some commit checks failed.

---

*[plugins/viaroute.hpp, line 123 \[r1\]](https://reviewable.io:443/reviews/project-osrm/osrm-backend/1795#-K4YBE-K73_a9ERUJrLK:-K4YBE-K73_a9ERUJrLL:-804150890) ([raw file](https://github.com/project-osrm/osrm-backend/blob/e4d60d70d20f46d207d82499d6653e353e892482/plugins/viaroute.hpp#L123)):*
No need for constructing an expensive stream, use

    const auto thisIsAStdString = "hello world" + std::to_string(integral);

I think the whole file has this problem.

---

*[routing_algorithms/routing_base.hpp, line 194 \[r1\]](https://reviewable.io:443/reviews/project-osrm/osrm-backend/1795#-K4YCCEaceB29BNnixg-:-K4YCCEaceB29BNnixg0:1250862421) ([raw file](https://github.com/project-osrm/osrm-backend/blob/e4d60d70d20f46d207d82499d6653e353e892482/routing_algorithms/routing_base.hpp#L194)):*
If this really is a model of the `RandomIterator` concept, there is no need for `std::prev` or similar, you can just use `--` and `++`.

Maybe `static_assert` on the iterator tag via iterator traits.

---

*[routing_algorithms/shortest_path.hpp, line 129 \[r1\]](https://reviewable.io:443/reviews/project-osrm/osrm-backend/1795#-K4YDwxX4qUBPbeUeZL8:-K4YDwxX4qUBPbeUeZL9:1660723191) ([raw file](https://github.com/project-osrm/osrm-backend/blob/e4d60d70d20f46d207d82499d6653e353e892482/routing_algorithms/shortest_path.hpp#L129)):*
what
why
what is this monster

can you handle 14 parameters, let alone the`2^4=16` possible states from the 4 bools alone --- because I can't

---

*[routing_algorithms/shortest_path.hpp, line 231 \[r1\]](https://reviewable.io:443/reviews/project-osrm/osrm-backend/1795#-K4YEWLYKZTFOflAk05E:-K4YEWLYKZTFOflAk05F:2101406251) ([raw file](https://github.com/project-osrm/osrm-backend/blob/e4d60d70d20f46d207d82499d6653e353e892482/routing_algorithms/shortest_path.hpp#L231)):*
see above

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/project-osrm/osrm-backend/1795)*

TheMarex:
*[routing_algorithms/routing_base.hpp, line 194 \[r1\]](https://reviewable.io:443/reviews/project-osrm/osrm-backend/1795#-K4YCCEaceB29BNnixg-:-K4YGObJ2Vr8c45SHsQh:-850313398) ([raw file](https://github.com/project-osrm/osrm-backend/blob/e4d60d70d20f46d207d82499d6653e353e892482/routing_algorithms/routing_base.hpp#L194)):*
Doesn't that map to the same thing anyway?

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/project-osrm/osrm-backend/1795)*

routing_algorithms/direct_shortest_path.hpp
```diff
@@ -64,143 +65,71 @@ class DirectShortestPathRouting final
...
-                super::RoutingStep(forward_core_heap, reverse_core_heap, &middle, &distance,
-                                   min_edge_offset, true, false);
-            }
-            if (!reverse_core_heap.Empty())
-            {
-                super::RoutingStep(reverse_core_heap, forward_core_heap, &middle, &distance,
-                                   min_edge_offset, false, false);
-            }
+            super::Search(forward_heap, reverse_heap, distance, packed_leg);
+            std::copy(packed_leg.begin(), packed_leg.end(), std::ostream_iterator<NodeID>(std::cout, ", "));
```
TheMarex: Foo forgot this debug output.

routing_algorithms/shortest_path.hpp
```diff
@@ -51,294 +53,482 @@ class ShortestPathRouting final
...
+                const int total_distance_to_forward,
+                const int total_distance_to_reverse,
+                int &new_total_distance_to_forward,
+                int &new_total_distance_to_reverse,
+                std::vector<NodeID> &leg_packed_path_forward,
+                std::vector<NodeID> &leg_packed_path_reverse) const
+    {
+        // silence unused warning
+        (void) search_from_forward_node;
+        (void) search_from_reverse_node;
```
danpat: Might be neater to do this in the signature instead:

    const bool /* search_from_forward_node */,
    const bool /* search_from_reverse_node */,

Why are these unused?

TheMarex: They are used, see the assertions below.

danpat: They're only used during debug builds, but they're polluting the function signature under normal use.  I think the assertion might belong at the call site for `SearchLoop`, there only seems to be one.

TheMarex:
Merged after fixing the issues.

Issue #1796, opened 2015-12-01
Don't snap to ferries


TheMarex:
As @danpat pointed out we currently snap against all modes of transportation, even ferries.

This creates non-nonsensical results like below:

http://map.project-osrm.org/?z=6&center=57.308724,20.632324&loc=55.078367,16.391602&loc=55.961501,16.029053&hl=en&ly=&alt=&df=&srv=

We should remove all non-default mode segments from the RTree.

TheMarex:
So we have two ways to fix this:

1. On the client side: Add a new parameter to the query that filters the found edges by mode of transportation
2. During pre-processing: For each mode of transportation add an additional flag that specifies if its `"snapable"`.

My original approach just disallowed snapping to non-default mode of transportations and uses the second approach. This is problematic in two ways:

- (a) The semantic value of the mode of transportation value is user defined. `forward_mode=2` could mean anything. It could make sense to snap to that.
- (b) This breaks a lot of tests since we test routability on streets with non-default travel mode for the profile tests. (which relies on snapping to that road in the first place)

The second approach would add a new API parameter and yet another query mode. This is less problematic after the `StaticRTree` refactor which moves the application complexity outside of the query code.However from a user standpoint this does not really help. They would need to know that the streets they want to snap to have `mode=0` or `mode=1`. Instead it should just Do The Right Thing (tm).

We can try to fix (b) by changing the way we do routability tests.

emiltin:
the mode is user defined, but is usually known by both back and front end. that's how the front displays eg 'ferry' when applicable. so the front end (and user) would know what mode to use for eg bikeways.

MoKob:
Related: https://github.com/Project-OSRM/osrm-backend/issues/1726

TheMarex:
Resolved.

PR #1797, opened 2015-12-01
Don't include non-default modes of transportation in the StaticRTree

TheMarex:
Addresses https://github.com/Project-OSRM/osrm-backend/issues/1796

[Review on Reviewable](https://reviewable.io/reviews/project-osrm/osrm-backend/1797)

### DIFF
--- a/extractor/extractor.cpp
+++ b/extractor/extractor.cpp
@@ -579,9 +579,14 @@ void extractor::BuildRTree(std::vector<EdgeBasedNode> &&node_based_edge_list_,
 
     // Filter out all segments that are not traversible by the default mode of transportation
     // For example this filters out ferry routes and such.
-    std::remove_if(node_based_edge_list.begin(), node_based_edge_list.end(), [](const EdgeBasedNode& segment) {
-                return segment.forward_travel_mode != TRAVEL_MODE_DEFAULT && segment.backward_travel_mode != TRAVEL_MODE_DEFAULT;
-            });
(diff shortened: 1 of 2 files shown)


extractor/extractor.cpp
```diff
@@ -572,9 +572,17 @@ void extractor::WriteNodeMapping(const std::vector<QueryNode> & internal_to_exte
...
                          const std::vector<QueryNode> &internal_to_external_node_map)
 {
+    std::vector<EdgeBasedNode> node_based_edge_list(std::move(node_based_edge_list_));
+
+    // Filter out all segments that are not traversible by the default mode of transportation
+    // For example this filters out ferry routes and such.
+    std::remove_if(node_based_edge_list.begin(), node_based_edge_list.end(), [](const EdgeBasedNode& segment) {
+                return segment.forward_travel_mode != TRAVEL_MODE_DEFAULT && segment.backward_travel_mode != TRAVEL_MODE_DEFAULT;
+            });
+
```
daniel-j-h: Hmm looks like you forgot to `erase`? `remove_if` will only shuffle items satisfying the predicate to the end.

TheMarex: Good catch. It actually does not shuffle, it overwrites. But this means we have duplicate entries.

TheMarex:
Superseded by https://github.com/Project-OSRM/osrm-backend/pull/1802

PR #1798, opened 2015-12-02
Distance table consistent snapping

danpat:
This commit is to address https://github.com/Project-OSRM/osrm-backend/issues/1447

It makes a few changes to the way the `distance_table` plugin selects phantom nodes.  The new rules are:

  - Only one is selected for each candidate location
  - If all candidates are big components, things proceed as before
  - If some candidates are on a small component, but are near a big component, they are snapped to the big component
  - If some candidates are on a small component and are *not* near a big component, they remain on their small components
  - If all candidates are on small components, then these are used, we do not search for the nearest big component.

A couple of extra changes had to happen to make this work:

  - Needed to expose the `max_checked_elements` in `IncrementalFindPhantomNodeForCoordinate` so that we would *not* return big components that are outside `max_distance`
  - Made the `small_component_size` configurable on the command line, to support testability.  It still defaults to the previously hard-coded value of 1000

I'm not 100% confident that this is the behaviour we want, but the test case included should demonstrate the new behaviour (this test fails with the old behaviour).  Please review and discuss the approach.

[Review on Reviewable](https://reviewable.io/reviews/project-osrm/osrm-backend/1798)

### DIFF
--- a/extractor/extractor.cpp
+++ b/extractor/extractor.cpp
@@ -436,7 +436,7 @@ void extractor::FindComponents(unsigned max_edge_id,
                          component_search.get_component_id(node.reverse_edge_based_node_id));
 
         const unsigned component_size = component_search.get_component_size(forward_component);
-        node.component.is_tiny = component_size < 1000;
+        node.component.is_tiny = component_size < config.small_component_size;
         node.component.id = 1 + forward_component;
     }
 }
--- a/extractor/extractor_options.cpp
+++ b/extractor/extractor_options.cpp
@@ -64,7 +64,11 @@ ExtractorOptions::ParseArguments(int argc, char *argv[], ExtractorConfig &extrac
         "Number of threads to use")(
             "generate-edge-lookup",boost::program_options::value<bool>(
                                                 &extractor_config.generate_edge_lookup)->implicit_value(true)->default_value(false),
-                                 "Generate a lookup table for internal edge-expanded-edge IDs to OSM node pairs");
+                                 "Generate a lookup table for internal edge-expanded-edge IDs to OSM node pairs")(
+        "small-component-size",
+        boost::program_options::value<unsigned int>(&extractor_config.small_component_size)
+            ->default_value(1000),
+        "Number of nodes required before a strongly-connected-componennt is considered big (affects nearest neighbor snapping)");
 
 #ifdef DEBUG_GEOMETRY
         config_options.add_options()("debug-turns",
--- a/features/testbot/distance_matrix.feature
+++ b/features/testbot/distance_matrix.feature
@@ -135,3 +135,30 @@ Feature: Basic Distance Matrix
             |   | b   | e   | f   |
             | a | 100 | 200 | 300 |
             | b | 0   | 100 | 200 |
+
+    # There is a 1100m limit when searching for nearest neighbours
+    # so we set the grid size to something that makes it easy to
+    # put nodes inside/outside that limit
+    Scenario: Testbog - Check snapping on on small components
+        Given a grid size of 300 meters
+        Given the extract extra arguments "--small-component-size 4"
+        Given the node map
+            | a | b |  |   | m |  | x |
+            | d | e |  |   | n |  | y |
+
+        And the ways
+            | nodes |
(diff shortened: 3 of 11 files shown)


extractor/extractor_options.cpp
```diff
@@ -64,7 +64,11 @@ ExtractorOptions::ParseArguments(int argc, char *argv[], ExtractorConfig &extrac
         "Number of threads to use")(
             "generate-edge-lookup",boost::program_options::value<bool>(
                                                 &extractor_config.generate_edge_lookup)->implicit_value(true)->default_value(false),
-                                 "Generate a lookup table for internal edge-expanded-edge IDs to OSM node pairs");
+                                 "Generate a lookup table for internal edge-expanded-edge IDs to OSM node pairs")(
+        "small-component-size",
+        boost::program_options::value<unsigned int>(&extractor_config.small_component_size)
+            ->default_value(1000),
+        "Number of nodes required before a strongly-connected-componennt is considered big (affects nearest neighbor snapping)");
```
daniel-j-h: Cool.. but how does it affect the snapping? From the description I can not see how it would affect snapping if I increase the size, and how it affects the snapping if I decrease the size. I think we can be more user-friendly here --- in the end it's not just us who wants to use this.

TheMarex:
> If all candidates are on small components, then these are used, we do not search for the nearest big component

You mean on the same small component?

> Needed to expose the max_checked_elements in IncrementalFindPhantomNodeForCoordinate so that we would not return big components that are outside max_distance

I would roll back all changes to `StaticRTree` in this PR since its going to be overwritten by my refactor anyway.

danpat:
The current behaviour was only to fetch 1 small component and 1 big component.  The alternative would be to implement something like "most common component" and snap all coordinates to that, but to do that right would require identifying component ids for big components, which we don't do right now.

extractor/extractor_options.cpp
danpat: Ideally, this is something that's better documented in a manpage/user manual, and the description here would be shorter.

extractor/extractor.cpp
```diff
@@ -436,7 +436,7 @@ void extractor::FindComponents(unsigned max_edge_id,
                          component_search.get_component_id(node.reverse_edge_based_node_id));
 
         const unsigned component_size = component_search.get_component_size(forward_component);
-        node.component.is_tiny = component_size < 1000;
+        node.component.is_tiny = component_size < config.small_component_size;
```
TheMarex: Yes good idea here.

features/testbot/distance_matrix.feature
```diff
@@ -100,3 +100,30 @@ Feature: Basic Distance Matrix
             | y | 500 | 0   | 300 | 200 |
             | d | 200 | 300 | 0   | 300 |
             | e | 300 | 400 | 100 | 0   |
+
+    # There is a 1100m limit when searching for nearest neighbours
```
TheMarex: I plan to remove this limit completely. It seldom hits anyway and it has been a constant source of grievance while debugging.

plugins/distance_table.hpp
```diff
@@ -104,11 +106,44 @@ template <class DataFacadeT> class DistanceTablePlugin final : public BasePlugin
...
+
+            if (phantom_node_vector[i].size() == 1)
+            {
+                at_least_one_only_big_component |= !phantom_node_vector[i].back().component.is_tiny;
+            }
 
             BOOST_ASSERT(phantom_node_vector[i].front().is_valid(facade->GetNumberOfNodes()));
         }
 
+        // If we have one phantom node that only exists in a big component, then we will force all others
```
TheMarex: Why not use the logic from `viaroute`?

1. Check if phantom nodes are all in the same component (doesn't matter if small or big)
2. If they are not try to use the big component for each node

I think it should be consistent.

TheMarex:
We should merge this after https://github.com/Project-OSRM/osrm-backend/pull/1802 is merged.

TheMarex:
Rebased and made the behavior consistent with `viaroute` (uses the same function to snap the target and source array)

PR #1802, opened 2015-12-04
Refactor StaticRTree

TheMarex:
This PR does the following thing:
- Remove all application dependent query code outside of StaticRTree
- Add lambda functions to interactively filter while searching
- Add convenience layer `GeospatialQuery` that implements application specific code
- Reduce the number of supported queries in `StaticRTree` to exactly one.
- In the process the `locate` plugin (snap to edge endpoints was removed) as this functionality is not needed in any other plugin and the API does not have any significant usage

Future work:
- Add mode filter API parameter
- Implement mode filter in `GeospatialQuery`

/cc @danpat @daniel-j-h for code review.

[Review on Reviewable](https://reviewable.io/reviews/project-osrm/osrm-backend/1802)


algorithms/geospatial_query.hpp
```diff
@@ -0,0 +1,180 @@
...
+#include <osrm/coordinate.hpp>
+
+#include <vector>
+#include <memory>
+#include <algorithm>
+
+// Implements complex queries on top of an RTree and builds PhantomNodes from it.
+//
+// Only holds a weak reference on the RTree!
+template <typename RTreeT> class GeospatialQuery
```
daniel-j-h: Does the rtree type really has to be templated?

algorithms/geospatial_query.hpp
```diff
@@ -0,0 +1,180 @@
...
+                return num_results > 0 && has_big_component;
+            });
+
+
+        if (results.size() == 0)
+        {
+            return std::make_pair(PhantomNode {}, PhantomNode {});
+        }
+
+        BOOST_ASSERT(results.size() > 0);
```
daniel-j-h: This can not trigger since `.size()` returns an unsigned `size_t` and we check for the zero case.

algorithms/geospatial_query.hpp
```diff
@@ -0,0 +1,180 @@
...
+        return std::make_pair(MakePhantomNode(input_coordinate, results.front()).second, MakePhantomNode(input_coordinate, results.back()).second);
```
daniel-j-h: Oh I see, maybe leave the assert in, since we use `.front` here.

algorithms/geospatial_query.hpp
```diff
@@ -0,0 +1,180 @@
...
+    }
+
+  private:
+    inline std::vector<std::pair<double, PhantomNode>>
```
daniel-j-h: No need to mark member functions `inline` that are defined in the class definition. They are already inline!

algorithms/geospatial_query.hpp
```diff
@@ -0,0 +1,180 @@
...
+        }
+        return transformed;
+    }
+
+    inline std::pair<bool, bool> checkSegmentBearing(const EdgeData &segment,
+                                                     const float filter_bearing,
+                                                     const float filter_bearing_range)
+    {
+        const float forward_edge_bearing =
+            coordinate_calculation::bearing(coordinates->at(segment.u), coordinates->at(segment.v));
```
daniel-j-h: Why the `.at` member function here instead of the subscript operator `[]`? We are sure that it's there, no need to use `.at` for bounds checking?

benchmarks/static_rtree.cpp
```diff
@@ -65,99 +67,65 @@ FixedPointCoordinateListPtr LoadCoordinates(const boost::filesystem::path &nodes
     return coords;
 }
 
-void Benchmark(BenchStaticRTree &rtree, unsigned num_queries)
+template <typename QueryT>
+void BenchmarkQuery(const std::vector<FixedPointCoordinate> &queries,
+                    std::string name,
```
daniel-j-h: const-ref

benchmarks/static_rtree.cpp
```diff
@@ -65,99 +67,65 @@ FixedPointCoordinateListPtr LoadCoordinates(const boost::filesystem::path &nodes
...
+{
+    std::mt19937 mt_rand(RANDOM_SEED);
+    std::uniform_int_distribution<> lat_udist(WORLD_MIN_LAT, WORLD_MAX_LAT);
+    std::uniform_int_distribution<> lon_udist(WORLD_MIN_LON, WORLD_MAX_LON);
+    std::vector<FixedPointCoordinate> queries;
+    for (unsigned i = 0; i < num_queries; i++)
     {
-        PhantomNode phantom;
-        rtree.FindPhantomNodeForCoordinate(q, phantom, 3);
+        queries.emplace_back(FixedPointCoordinate(lat_udist(mt_rand), lon_udist(mt_rand)));
```
daniel-j-h: No need to specify `FixedPointCoordinate`, that's the point of `emplace_back`, to only pass arguments and construct the container element inplace! The current code does exactly the opposite, having no benefits at all!

data_structures/hidden_markov_model.hpp
```diff
@@ -140,7 +140,7 @@ template <class CandidateLists> struct HiddenMarkovModel
             for (const auto s : osrm::irange<std::size_t>(0u, viterbi[initial_timestamp].size()))
             {
                 viterbi[initial_timestamp][s] =
-                    emission_log_probability(candidates_list[initial_timestamp][s].second);
+                    emission_log_probability(candidates_list[initial_timestamp][s].first);
```
daniel-j-h: Out of scope. This PR is supposed to refactor the static rtree. If that's a HMM bug, handle separately.

data_structures/static_rtree.hpp
```diff
@@ -567,589 +342,130 @@ class StaticRTree
...
-    {
-        const bool ignore_tiny_components = (zoom_level <= 14);
-        EdgeDataT nearest_edge;
-
-        float min_dist = std::numeric_limits<float>::max();
-        float min_max_dist = std::numeric_limits<float>::max();
-
-        std::priority_queue<QueryCandidate> traversal_queue;
-        traversal_queue.emplace(0.f, 0);
+                results.emplace_back(std::move(current_segment));
```
daniel-j-h: When the container stores Ts and you want to move a T into it use the `push_back(&&)` overload and move it in.
When the container stores Ts that can be constructed from U, V (has a constructor T(U, V)) and you want to move a U or V into it, use `emplace_back(Args&&...)` and move the U and V arguments in.

data_structures/static_rtree.hpp
```diff
@@ -567,589 +342,130 @@ class StaticRTree
...
-        }
-        if (SPECIAL_NODEID != result_phantom_node.reverse_node_id)
-        {
-            const auto new_weight =
-                static_cast<TreeWeightType>(result_phantom_node.reverse_weight * (1.f - ratio));
-            result_phantom_node.reverse_weight = new_weight;
-        }
-    }
+        LeafNode current_leaf_node;
+        LoadLeafFromDisk(leaf_id, current_leaf_node);
```
daniel-j-h: Shouldn't we check for errors..

data_structures/static_rtree.hpp
```diff
@@ -567,589 +342,130 @@ class StaticRTree
...
         {
-            result_phantom_node.location.lon = input_coordinate.lon;
-        }
-        if (1 == std::abs(input_coordinate.lat - result_phantom_node.location.lat))
-        {
-            result_phantom_node.location.lat = input_coordinate.lat;
+            auto &current_edge = current_leaf_node.objects[i];
+            const float current_perpendicular_distance =
+                coordinate_calculation::perpendicular_distance_from_projected_coordinate(
+                    m_coordinate_list->at(current_edge.u), m_coordinate_list->at(current_edge.v),
```
daniel-j-h: `.at` vs `[]`

data_structures/static_rtree.hpp
```diff
@@ -1170,18 +485,11 @@ class StaticRTree
         BOOST_ASSERT_MSG(leaves_stream.good(), "Reading from leaf file failed.");
```
daniel-j-h: Clearing the goodbit is a horrible idea --- there is a reason a stream is in a bad state, it does not make sense to simply ignore it! Also, we don't throw or otherwise handle errors here, and the callsite ^ does neither.

extractor/extractor.cpp
```diff
@@ -573,11 +573,20 @@ void extractor::WriteNodeMapping(const std::vector<QueryNode> & internal_to_exte
...
+                           const std::vector<QueryNode> &internal_to_external_node_map)
 {
+    SimpleLogger().Write() << "constructing r-tree of " << node_based_edge_list.size()
+                           << " edge elements build on-top of " << internal_to_external_node_map.size()
+                           << " coordinates";
+
+    TIMER_START(construction);
     StaticRTree<EdgeBasedNode>(node_based_edge_list, config.rtree_nodes_output_path.c_str(),
                                config.rtree_leafs_output_path.c_str(),
                                internal_to_external_node_map);
```
daniel-j-h: Hmm I thought the static rtee's constructor takes `std::string`, so no need to go from string -> c-string -> string here

daniel-j-h:
In general, the code is really hard to follow, not only because all the templated parameters, but also because a lot of types are simply pairs. And kv.first - kv.second * -kv.first says nothing on first sight. Would love some small helper structs, that 1/ bring in more rigorous typing and 2/ show intend.

algorithms/geospatial_query.hpp
TheMarex: Yes because the RTree type depends on the value that is saved inside it. For the tests we want to use a different value type. We could however just template the value type?

algorithms/geospatial_query.hpp
TheMarex: `(*coordinates)[...]` is kind of awkward to write?

data_structures/hidden_markov_model.hpp
TheMarex: No the order of the parameter changed. Used to be `(coord, distance)` now its `(distance, coord)`.

Issue #1808, opened 2015-12-09
STXXL corrupts name lists when they are large


danpat:
TL;DR: `STXXLVector<std::string>` is producing undefined behaviour when things get paged out.  The STXXL container types only support POD objects.  `extraction_containers.hpp` has an `STXXLVector<string>` that breaks when large numbers of strings are inserted that trigger paging behaviour.  It works if not enough data is inserted to trigger paging, which seems to be what we've been doing up to now.

If you use the `name` attribute on ways as a storage mechanism for other metadata, this generates a much larger `.names` file than if just regular street names are used.

Under circumstances where there are very large numbers of unique names (in our test data set, around 67M unique names), STXXL seems to corrupt the name strings upon retrieval, which leads to a segfault at the end of the first stage of `osrm-extract`.

The stack trace looks like this:

```
[extractor] Sorting used ways         ... ok, after 35.3939s
[extractor] Sorting 0 restriction. by from... ok, after 0.000661s
[extractor] Fixing restriction starts ... ok, after 0s
[extractor] Sorting restrictions. by to  ... ok, after 0.000308s
[extractor] Fixing restriction ends   ... ok, after 0s
[info] usable restrictions: 0
[extractor] writing street name index ...
Program received signal SIGSEGV, Segmentation fault.
(gdb) bt
   #0  __memcpy_sse2_unaligned () at ../sysdeps/x86_64/multiarch/memcpy-sse2-unaligned.S:36
   #1  0x00007ffff607d430 in std::basic_streambuf<char, std::char_traits<char> >::xsputn(char const*, long) () from /usr/lib/x86_64-linux-gnu/libstdc++.so.6
   #2  0x00007ffff6074772 in std::ostream::write(char const*, long) () from /usr/lib/x86_64-linux-gnu/libstdc++.so.6
   #3  0x00000000004df7a3 in ExtractionContainers::WriteNames(std::string const&) const ()
   #4  0x000000000054a4a9 in ExtractionContainers::PrepareData(std::string const&, std::string const&, std::string const&, lua_State*) ()
   #5  0x000000000049dcc7 in extractor::run() ()
   #6  0x000000000046f965 in main ()
   (gdb)
```

After lots of debugging, it turns out that when we retrieve the `name` values from the `external_name_list` here:

https://github.com/Project-OSRM/osrm-backend/blob/c38183423f82bd1e0568af04201f12599966d2fe/extractor/extraction_containers.cpp#L136-L138

[...]

Issue #1732, opened 2015-10-11
OSRM creates instructions with invalid road names


mk112014:
On Windows, OSRM has no problem with small data, but with large data it creates instructions with invalid road names.
Data: North America (http://download.geofabrik.de/north-america-latest.osm.pbf)
Query: viaroute?loc=37.3575027358159,-121.882495880127&loc=37.3570315637053,-121.880419850349&instructions=true
![query](https://cloud.githubusercontent.com/assets/15073906/10416522/f6b28148-7044-11e5-8865-61738ea45bb0.jpg)
Instruction:
![instruction](https://cloud.githubusercontent.com/assets/15073906/10416523/fe208fd8-7044-11e5-9b97-468f46c4b08f.jpg)

BergWerkGIS:
Looks good on [map.project-osrm.org](http://map.project-osrm.org/?z=18&center=37.357824%2C-121.881594&loc=37.357554%2C-121.882399&loc=37.357554%2C-121.880795&hl=en&ly=&alt=&df=&srv=):

![image](https://cloud.githubusercontent.com/assets/4549888/10416567/2dbea4a6-701c-11e5-8512-1476c82586cf.png)

BergWerkGIS:
I am able to reproduce the problem.

* osrm binaries downloaded yesterday
* north-america-latest.osm.pbf downloaded today

Output of `osrm-extract.exe` and `osrm-prepare.exe`:
https://gist.github.com/BergWerkGIS/18f4433d849b444ea29c

Wrong route (see below):
http://localhost:5000/viaroute?loc=37.3575027358159,-121.882495880127&loc=37.3570315637053,-121.880419850349&instructions=true

Correct route (see below):
http://router.project-osrm.org/viaroute?loc=37.3575027358159,-121.882495880127&loc=37.3570315637053,-121.880419850349&instructions=true


#### localhost
```json
{
  "status_message": "Found route between points",
  "status": 0,
  "found_alternative": false,
  "via_points": [
    [
      37.357498,
      -121.8825
    ],
    [
      37.357037,
      -121.880409
    ]
  ],
  "route_geometry": "uzbgfA|vbngFmIaUiZqx@tCoC~|@im@",
  "route_instructions": [
    [
      "10",
      "South Shiloh Court",
      131,
      0,
      17,
      "131m",
      "NE",
      60,
      1
    ],
    [
      "3",
      "Alkali Springs Ro",
      139,
      2,
      15,
[...]

PR #1809, opened 2015-12-09
Use a std::vector in place of stxxl:vector for the names list

rparanjpe-tesla:
-For large datasets with very many unique names, stxxl::vector can corrupt
 data. Technically, we should only be using stxxl:vectors with POD. Other
 types might lead to strange/unpredictable behavior as we noticed here.
-See http://algo2.iti.kit.edu/dementiev/stxxl/trunk/FAQ.html

@danpat @TheMarex

### DIFF
--- a/extractor/extraction_containers.cpp
+++ b/extractor/extraction_containers.cpp
@@ -55,7 +55,8 @@ ExtractionContainers::ExtractionContainers()
 {
     // Check if stxxl can be instantiated
     stxxl::vector<unsigned> dummy_vector;
-    name_list.push_back("");
+    name_char_data.push_back('\0');
+    name_offsets.push_back(1);
 }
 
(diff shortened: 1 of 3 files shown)


danpat:
This is related to https://github.com/Project-OSRM/osrm-backend/issues/1808

TheMarex:
We should use a `stxxl::vector<char> names_data;` + `stxxl::vector<unsigned> names_offset` instead. This would only use POD and memory usage would not really be increased much.

We would then only need to write the values in sequence to disk [here](https://github.com/Project-OSRM/osrm-backend/blob/develop/extractor/extraction_containers.cpp#L112). That would actually be much cleaner. I'm unsure why its a vector of strings at all.

danpat:
Closes #1808, #1732

PR #1813, opened 2015-12-12
Configurable snapping

TheMarex:
This makes it configurable from profiles whether to not snap to road or not. This is useful for ferries and the like where snapping makes no sense.

- [x] Rebase on develop
- [x] Write test to verify the behaviour
- [x] Fix routability cucumber tests that rely on snapping to ferry roads _doesn't seem to be an issue for the car profile_

### DIFF
--- a/data_structures/import_edge.hpp
+++ b/data_structures/import_edge.hpp
@@ -44,6 +44,7 @@ struct NodeBasedEdge
                            bool backward,
                            bool roundabout,
                            bool access_restricted,
+                           bool startpoint,
                            TravelMode travel_mode,
                            bool is_split);
 
@@ -55,6 +56,7 @@ struct NodeBasedEdge
     bool backward : 1;
     bool roundabout : 1;
     bool access_restricted : 1;
+    bool startpoint : 1;
     bool is_split : 1;
     TravelMode travel_mode : 4;
 };
@@ -69,9 +71,10 @@ struct NodeBasedEdgeWithOSM : NodeBasedEdge
                            bool backward,
                            bool roundabout,
                            bool access_restricted,
+                           bool startpoint,
                            TravelMode travel_mode,
                            bool is_split)
-        : NodeBasedEdge(SPECIAL_NODEID, SPECIAL_NODEID, name_id, weight, forward, backward, roundabout, access_restricted, travel_mode, is_split),
+        : NodeBasedEdge(SPECIAL_NODEID, SPECIAL_NODEID, name_id, weight, forward, backward, roundabout, access_restricted, startpoint, travel_mode, is_split),
         osm_source_id(source), osm_target_id(target) {}
 
     OSMNodeID osm_source_id;
--- a/data_structures/node_based_graph.hpp
+++ b/data_structures/node_based_graph.hpp
@@ -47,10 +47,10 @@ struct NodeBasedEdgeData
 
     NodeBasedEdgeData(int distance, unsigned edge_id, unsigned name_id,
             bool access_restricted, bool reversed,
-            bool roundabout, TravelMode travel_mode)
+            bool roundabout, bool startpoint, TravelMode travel_mode)
         : distance(distance), edge_id(edge_id), name_id(name_id),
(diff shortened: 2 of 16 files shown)


data_structures/node_based_graph.hpp
```diff
@@ -60,6 +60,7 @@ struct NodeBasedEdgeData
     bool access_restricted : 1;
     bool reversed : 1;
     bool roundabout : 1;
+    bool startpoint : 1;
     TravelMode travel_mode : 4;
```
MoKob: Is this struct defined at multiple locations? Would prefer to move it to an include and only define it once.

data_structures/import_edge.hpp
```diff
@@ -55,6 +56,7 @@ struct NodeBasedEdge
     bool backward : 1;
     bool roundabout : 1;
     bool access_restricted : 1;
+    bool startpoint : 1;
     bool is_split : 1;
     TravelMode travel_mode : 4;
 };
```
MoKob: multiply defined? Just missing is_split here?

extractor/edge_based_graph_factory.cpp
```diff
@@ -78,6 +78,11 @@ void EdgeBasedGraphFactory::GetEdgeBasedNodes(std::vector<EdgeBasedNode> &nodes)
     nodes.swap(m_edge_based_node_list);
 }
 
+void EdgeBasedGraphFactory::GetStartPointMarkers(std::vector<bool> &node_is_startpoint)
+{
+    m_edge_based_node_is_startpoint.swap(node_is_startpoint);
+}
```
MoKob: Effective C++, Item 25 (https://books.google.de/books?id=U7lTySXdFk0C&pg=PT139&lpg=PT139&dq=prefer+std::swap+over+member&source=bl&ots=nO6Jrdv9kb&sig=nX3EPlZUTHahOvHTCCSoSba8RmM&hl=en&sa=X&ved=0ahUKEwjCzfvp_trJAhVE_Q4KHRZpA4YQ6AEIOjAF#v=onepage&q=prefer%20std%3A%3Aswap%20over%20member&f=false).

You should consider adding `using std::swap` and call `swap(node_is_starpoint, m_edge_base_node_is_startpoint)` as a general way to invoke swap.

extractor/extractor.cpp
```diff
@@ -529,32 +531,31 @@ extractor::BuildEdgeExpandedGraph(std::vector<QueryNode> &internal_to_external_n
...
-            config.edge_segment_lookup_path,
-            config.edge_penalty_path,
-            config.generate_edge_lookup
+                                 config.edge_segment_lookup_path, config.edge_penalty_path,
+                                 config.generate_edge_lookup
 #ifdef DEBUG_GEOMETRY
-            , config.debug_turns_path
+                                 ,
+                                 config.debug_turns_path
 #endif
```
MoKob: why in two lines? Would prefer consistent style with ',' at begin of the line

extractor/extractor.cpp
```diff
@@ -572,26 +573,44 @@ void extractor::WriteNodeMapping(const std::vector<QueryNode> & internal_to_exte
 
     Saves tree into '.ramIndex' and leaves into '.fileIndex'.
  */
-void extractor::BuildRTree(const std::vector<EdgeBasedNode> &node_based_edge_list,
+void extractor::BuildRTree(std::vector<EdgeBasedNode> node_based_edge_list,
+                           std::vector<bool> node_is_startpoint,
                            const std::vector<QueryNode> &internal_to_external_node_map)
```
MoKob: We are not taking ownership of these vectors. So sink parameter passing does not apply. Shouldn't these be passed by const reference?

data_structures/node_based_graph.hpp
TheMarex: It's not the struct, just shares the same member names for consistency.

extractor/extractor.cpp
TheMarex: Needs to be sink because we filter it, look in the diff a few lines down.

PR #1817, opened 2015-12-14
Cleanup library setup

TheMarex:
Removes unneeded code and uses C++11 idioms.


library/osrm.hpp
```diff
@@ -49,9 +49,9 @@ class OSRM
     std::unique_ptr<OSRM_impl> OSRM_pimpl_;
 
   public:
-    explicit OSRM(libosrm_config &lib_config);
+    explicit OSRM(LibOSRMConfig lib_config);
```
daniel-j-h: If you take `LibOSRMConfig` by value here it can not be an [incomplete type](http://en.cppreference.com/w/cpp/language/type#Incomplete_type). If you instead take it by const-ref, you can simply forward-declare `LibOSRMConfig` above, and then remove the appropriate include, keeping this header clean by breaking dependencies.

library/osrm_impl.cpp
```diff
@@ -171,11 +159,11 @@ void OSRM_impl::increase_concurrent_query_count()
 }
 
 // proxy code for compilation firewall
-OSRM::OSRM(libosrm_config &lib_config) : OSRM_pimpl_(osrm::make_unique<OSRM_impl>(lib_config)) {}
+OSRM::OSRM(LibOSRMConfig lib_config) : OSRM_pimpl_(osrm::make_unique<OSRM_impl>(std::move(lib_config))) {}
```
daniel-j-h: You need to define the special member functions in the implementation file.

See discussion over at https://github.com/mapbox/mapbox-gl-native/issues/3254#issuecomment-164055377

PR #1818, opened 2015-12-15
A side_road tag support for the OSRM car profile.

gardster:
This PR adds support for the side_road tag. http://wiki.openstreetmap.org/wiki/Key:side_road
Without this processing the router tends to use these roads instead of main roads. 
http://map.project-osrm.org/?z=15&center=59.845456%2C30.276743&loc=59.851963%2C30.282558&loc=59.843198%2C30.276518&hl=en&ly=&alt=&df=&srv=
The actual speed on side roads is usually lower than the speed limit.

### DIFF
--- a/profiles/car.lua
+++ b/profiles/car.lua
@@ -131,6 +131,8 @@ maxspeed_table = {
 traffic_signal_penalty          = 2
 use_turn_restrictions           = true
 
+side_road_speed_multiplier      = 0.8
+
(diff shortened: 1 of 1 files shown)


daniel-j-h:
Thanks, this looks good. From what I can see on taginfo, there are only about 600 `side_road` keys:
http://taginfo.openstreetmap.org/keys/side_road#values

PR #1822, opened 2015-12-16
name_lengths std::vector --> stxxl::vector

rparanjpe-tesla:
@TheMarex 

This is following through with a suggestion you had made earlier on making this `name_lengths` vector stxxl as well. This PR does that and links the STXXL libs to the `datastructure-tests` so that the tests will succeed.


TheMarex:
Thanks :+1:

PR #1826, opened 2015-12-17
Add max values for viaroute and trip and reorganize return code handling

TheMarex:
"status" is now always:

* 200 if the request was successful
* 207 if the result is empty (no path found)
* 208 if no edge for snapping was found
* 400 if the request is invalid

 viaroute and trip now have a maximum of 500 and 100 locations
 respectively. Override with the `--max-viaroute-size` and `--max-trip-size`
 parameters.

### DIFF
--- a/features/options/routed/help.feature
+++ b/features/options/routed/help.feature
@@ -25,9 +25,11 @@ Feature: osrm-routed command line options: help
         And stdout should contain "--port"
         And stdout should contain "--threads"
         And stdout should contain "--shared-memory"
+        And stdout should contain "--max-viaroute-size"
+        And stdout should contain "--max-trip-size"
         And stdout should contain "--max-table-size"
         And stdout should contain "--max-matching-size"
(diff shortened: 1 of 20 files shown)


perliedman:
If I understand correctly, this PR has broken all sites using Leaflet Routing Machine, including OSRM's own demo site: http://map.project-osrm.org/?z=13&center=38.899500%2C-77.026900&loc=38.913075%2C-77.056732&loc=38.908667%2C-77.017250&hl=en&ly=&alt=&df=&srv=

I guess it also breaks any other code actually checking the status code.

While I agree the new codes make a lot of sense, and understand that the demo server is just that: a demo you can't rely on, it would be nice with a heads up for API breaking changes like this.

Have this been discussed somewhere that I have missed?

TheMarex:
Sorry for the missing heads up. Merged this a little bit more quickly than I planed to. I'll post something on the mailing list.

PR #1828, opened 2015-12-17
Via way restrictions [WIP]

rparanjpe-tesla:
This is a branch under development to incorporate via way restrictions, currently with the 'no_' case with a 3-member restriction (from, via, to). 

This will work as follows:
- [x] Read in restrictions if they are a "no_*" restriction on three ways: `a->b->c`
- [x] Write out the turn restrictions to the `.restrictions` file along with the regular restrictions in a special format (documented in the code)
- [x] Parse in restrictions and during the edge based graph building phase
- [ ] Create extra edge based nodes for the middle way (the via), which essentially retain the state of coming from the "from" way.

Concretely, let's say we have three ways in the restriction: `a->b->c` (`a` is the from, `b` is the via, `c` is the to). We create a new edge based node (`b'`) and take all of the *outgoing* edge_based_edges of `b` and duplicate them so that `b'` also has those outgoing edges. Next we delete the edge_based_edge connecting `a` to `b` and create an identically weighted one from `a` to `b'`. Finally we delete the edge from `b'` to `c`.

I'll be adding commits to this branch, but would love feedback as I continue development.

@TheMarex @danpat

### DIFF
--- a/data_structures/restriction.hpp
+++ b/data_structures/restriction.hpp
@@ -79,7 +79,7 @@ struct TurnRestriction
 
 /**
  * This is just a wrapper around TurnRestriction used in the extractor.
- * 
+ *
  * Could be merged with TurnRestriction. For now the type-destiction makes sense
  * as the format in which the restriction is presented in the extractor and in the
  * preprocessing is different. (see restriction_parser.cpp)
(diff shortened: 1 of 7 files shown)


TheMarex:
@rparanjpe-tesla The approach seems sound. Only concern seems the nearest neighbor search. Currently we assume that the nearest neighbor is unique - in the case of overlapping segments it obviously wouldn't be the case. As a result we might snap to (b') instead of (b) if we start a search on that segment and won't find a path to (c).

TheMarex:

PR #1832, opened 2015-12-18
modified local state of u_turn_penalty_variable

MoKob:
The current profiles do not result in an applied path penalty.
The local state of the variable results in a penalty of 0, no matter what value is specified.
For correct u-turn penalisation, the state needs to be switched to a global variable.


TheMarex:
@MoKob good catch!

Issue #1636, opened 2015-09-01
Feature Request - Add bearing to Response Message


Selikoff:
The "/match" and "/locate" services are excellent for map-matching a vehicle to a road.  Would there be much effort in updating these services to return a bearing value (0 to 360) in the list of response fields?  For bi-directional roads or intersections, I can understand this being ambiguous for the "/locate" service, but for one-way streets or places where a list of points was passed to "/match", the direction should be able to be determined using the road and history data.

TheMarex:
@Selikoff you get the full geometry of the road in the response of the `match` plugin (and an index into the coordinate array to the point that corresponds with the input trace point). You can extract the bearing from the adjacent coordinates of the road.

Selikoff:
Thanks @TheMarex, I'll give that a shot. Either way, it would be useful to have it as an output parameter in the response message.

Selikoff:
@TheMarex Can you provide more details about how this would work?  Would I just take the last two geometry points and compute the bearing between them?  Also, how do the geometry points differ from the map matched points?

TheMarex:
@Selikoff I'm sorry I actually confused the `match` response with `viaroute`. It is a little it more complicated:
- pass `instructions=true` and `geometry=true` to your `match` query
- use the instructions array to derive the location of your matched point in the geometry: look for the `ReachViaLocation` (numeric values is `9`) instructions as outlined [here](https://github.com/Project-OSRM/osrm-backend/wiki/Server-api#response-2)
- Each of those via instructions corresponds to a matched point (they are `vias` of the route)
- The 4th array entry `idx` of corresponds to the index in the coordinate array represented by `geometry`
- Depending of what bearing value you want to use compute `bearing(geometry[idx-1], geometry[idx])`

Selikoff:
Thanks @TheMarex, this is part of why it would be great if this was part of the /match service.  This is a lot of steps to perform by the client.

daniel-j-h:
I'm not sure if I fully understand your use-case but by now we have
- bearing_before: The clockwise angle from true north to the direction of travel immediately before the maneuver.
- bearing_after: The clockwise angle from true north to the direction of travel immediately after the maneuver.

https://github.com/Project-OSRM/osrm-backend/blob/master/docs/http.md#stepmaneuver-object

Is that good enough for your usage?

retorquere:
[...]

PR #1841, opened 2015-12-26
use double precision calculations instead of mixing double and float

mortada:
I noticed that the `match` API can have the `matched points` visibly off the way geometry. After digging into the code this appears to be a numerical precision issue. Some of the internal calculations are `float` based while others are `double` based. Here's an example showcasing the current behavior:

osrm-precision-current

Notice that the matched points (plotted in red) are not on the way polyline. After changing all calculations to be `double` based, here's the new behavior with this PR

osrm-precision-fixed

### DIFF
--- a/include/engine/geospatial_query.hpp
+++ b/include/engine/geospatial_query.hpp
@@ -30,7 +30,7 @@ template <typename RTreeT> class GeospatialQuery
     // Does not filter by small/big component!
     std::vector<PhantomNodeWithDistance>
     NearestPhantomNodesInRange(const FixedPointCoordinate &input_coordinate,
-                               const float max_distance,
+                               const double max_distance,
                                const int bearing = 0,
                                const int bearing_range = 180)
     {
@@ -40,7 +40,7 @@ template <typename RTreeT> class GeospatialQuery
                           {
                               return checkSegmentBearing(data, bearing, bearing_range);
                           },
-                          [max_distance](const std::size_t, const float min_dist)
+                          [max_distance](const std::size_t, const double min_dist)
                           {
                               return min_dist > max_distance;
                           });
@@ -61,7 +61,7 @@ template <typename RTreeT> class GeospatialQuery
                                      {
                                          return checkSegmentBearing(data, bearing, bearing_range);
                                      },
-                                     [max_results](const std::size_t num_results, const float)
+                                     [max_results](const std::size_t num_results, const double)
                                      {
                                          return num_results >= max_results;
                                      });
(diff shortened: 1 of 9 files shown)


mortada:
Note that the loss of precision happens immediately with `float` even though internally there's the fixed precision conversion to `int` (i.e. `FixedPointCoordinate`). A simple example to illustrate this:

```
#include <boost/lexical_cast.hpp>
#include <string>
#include <iomanip>
#include <iostream>

int main()
{
  std::string lat = "37.387442";

  float COORDINATE_PRECISION = 1000000.f;
  float lat_float = boost::lexical_cast<float>(lat);
  int lat_int = lat_float * COORDINATE_PRECISION;
  float lat_float_after = lat_int / COORDINATE_PRECISION;
  std::cout << std::setprecision(8) << "lat as string " << lat << " \n";
  std::cout << std::setprecision(8) << "lat as float " << lat_float << " \n";
  std::cout << std::setprecision(8) << "lat as int " << lat_int << " \n";
  std::cout << std::setprecision(8) << "lat as float (after fixed precision conversion) " << lat_float_after << " \n";

  double COORDINATE_PRECISION2 = 1000000.0;
  double lat_double = boost::lexical_cast<double>(lat);
  int lat_int2 = lat_double * COORDINATE_PRECISION2;
  double lat_double_after = lat_int2 / COORDINATE_PRECISION2;
  std::cout << std::setprecision(8) << "lat as double " <<  lat_double << " \n";
  std::cout << std::setprecision(8) << "lat as double (after fixed precision conversion) " << lat_double_after << " \n";
}
```

this outputs
```
lat as string 37.387442
lat as float 37.387444
lat as int 37387444
lat as float (after fixed precision conversion) 37.387444
lat as double 37.387442
lat as double (after fixed precision conversion) 37.387442
```

as can be seen above there's a loss of precision when doing the conversion in `float`, while using `double` is ok

TheMarex:
@mmehyar Sorry missed this PR. Can you try to rebase on`develop`?

In general :+1: as this will not increase memory usage and runtime overhead is neglectable.

mortada:
@TheMarex thanks a lot, I've rebased this PR

mortada:
@TheMarex thanks a lot! So I actually have a related question - what do you think would be the easiest way to also return the index of the way segment for each matched point? For instance, in this case the indices may be [2, 9, 10]. I'd like to be able to compute how far each matched point is into the matched way polyline.  

This is similar to the feature request here: https://github.com/Project-OSRM/osrm-backend/issues/1636

Issue #1789, opened 2015-11-20
Add option to disable hint_data output.


ardanika:
When using the Match service, the response contains hint_data. This additional data doubles the size of the response from the match query and is not needed unless you're also using the viaroute service. It would be nice to have a command line option or service request parameter that disables the output of hint_data at least for the Match service.

daniel-j-h:
The v5 API still sends the hint every time:
https://github.com/Project-OSRM/osrm-backend/blob/master/docs/http.md

Did we forget to add a switch for that during the v4 -> v5 transition or what is the idea behind it? @TheMarex

PR #1842, opened 2015-12-27
compute hint_data only if geometry or print_instructions is true

mortada:
related to this: https://github.com/Project-OSRM/osrm-backend/issues/1789

seems like it'd be nice to have a way to not compute `hist_data` (and therefore also `raw_route`)


TheMarex:
The API will be rewritten for `5.0` so I'm going to table this change for now. In general I will be aiming for a appraoch that is consistent across all services.

PR #1848, opened 2016-01-03
The big format

TheMarex:
Applies automated tools such as:

- [x] clang-tidy
- [x] clang-format
- [x] cncc

### DIFF
--- a/.cncc.style
+++ b/.cncc.style
@@ -1,0 +1,14 @@
+# Kind-specific patterns to check AST nodes against. Both python-clang and
+# libclang docs explain CursorKind, with differences in detail. See also:
+#  - https://github.com/llvm-mirror/clang/blob/aca4fe314a55cacae29e1548cb7bfd2119c6df4c/bindings/python/clang/cindex.py#L599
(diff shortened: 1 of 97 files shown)


.cncc.style
```diff
@@ -0,0 +1,14 @@
...
+#  - https://docs.python.org/2/library/re.html#regular-expression-syntax
+
+class_decl: '^([A-Z][a-z]+)+$'
+struct_decl: '^([A-Z][a-z]+)+$'
+field_decl: '^[a-z_]+$'
+var_decl: '^[a-z]+[a-z0-9_]*$'
+parm_decl: '^[a-z]*[a-z0-9_]*$'
+namespace: '^[a-z_]*$'
+cxx_method: '^([A-Z][a-z]+)+$'
+function_decl: '^[a-z]+([A-Z][a-z]+)*$'
```
daniel-j-h: Can and should namespaces and field decls start with `_` and contain `_`? If field decls can start with `_` I think it should be stricter like `_?[a-z]`.

PR #1854, opened 2016-01-05
Update in-tree libosmium dependency to 2.5.4

daniel-j-h:
We're at libosmium 2.3 at the moment, time to upgrade to 2.5!

The latest releases have some critical fixes, see the changelog:
https://github.com/osmcode/libosmium/blob/v2.5.4/CHANGELOG.md


daniel-j-h:
@joto: I think there is an issue in libosmium 2.5.4 with Windows:

```
  extractor.obj : error LNK2019: unresolved external symbol __imp_ntohl referenced in function "private: unsigned int __cdecl osmium::io::detail::PBFParser::read_blob_header_size_from_file(void)" (?read_blob_header_size_from_file@PBFParser@detail@io@osmium@@AEAAIXZ) [C:\projects\osrm\build\extractor-tests.vcxproj]
  C:\projects\osrm\build\Release\extractor-tests.exe : fatal error LNK1120: 1 unresolved externals [C:\projects\osrm\build\extractor-tests.vcxproj]
 ```

Function: [read_blob_header_size_from_file](https://github.com/osmcode/libosmium/blob/1d12e0902a1c620ef57d7f9e4892f92f9a083a55/include/osmium/io/detail/pbf_input_format.hpp#L113).

I think you need the following headers:

Windows: [Winsock2.h](https://msdn.microsoft.com/en-us/library/windows/desktop/ms740069%28v=vs.85%29.aspx)
Linux: arpa/inet.h per posix

And on Windows link against Ws2.

**Comment by BergWerkGIS** [comment]
@daniel-j-h @joto We had a similar error before. I fixed that within stxxl:
https://github.com/mapbox/windows-builds/blob/master/patches/stxxl-msc_ver-1900.diff

Will have a look during the next days.

**Comment by joto** [comment]
Includes for `winsock2.h` and `netinet/in.h`, respectively, are done in in https://github.com/osmcode/libosmium/blob/master/include/osmium/io/detail/pbf.hpp#L39-L44 and the tests of osmium-tool (which test pbf reading and writing) all run through on Linux and Windows. https://github.com/osmcode/libosmium/blob/master/cmake/FindOsmium.cmake#L105 (which you should use when you include libosmium) adds the `ws2_32` library on Windows.

So I don't think libosmium is to blame here.

**Comment by daniel-j-h** [comment]
Interesting, so we _have_ to use the libosmium cmake module? From what I can see we only add the inlcude path to libosmium and call it a day: https://github.com/Project-OSRM/osrm-backend/blob/73d40ce02abe57a40f2db43e7b6bc9ca5bbbcfc2/CMakeLists.txt#L39

**Comment by joto** [comment]
@daniel-j-h  Well, you don't have to, but it is the recommended way, because it does all the work for you.

**Comment by daniel-j-h** [comment]
@joto my [latest commit](https://github.com/Project-OSRM/osrm-backend/commit/cb37c0d135d724d58fe1e295352b4b46d6d0d2d3) now uses `FindOsmium.cmake`. Unfortunately I have to [set `OSMIUM_INCLUDE_DIR`](https://github.com/Project-OSRM/osrm-backend/commit/cb37c0d135d724d58fe1e295352b4b46d6d0d2d3#diff-af3b638bc2a3e6c650974192a53c7291R229), otherwise the module can not find the include dir:

```
CMake Error at /usr/local/share/cmake-3.4/Modules/FindPackageHandleStandardArgs.cmake:148 (message):
  Could NOT find Osmium (missing: OSMIUM_INCLUDE_DIR)
  /usr/local/share/cmake-3.4/Modules/FindPackageHandleStandardArgs.cmake:388 (_FPHSA_FAILURE_MESSAGE)
  third_party/libosmium/cmake/FindOsmium.cmake:262 (find_package_handle_standard_args)
  CMakeLists.txt:229 (find_package)
```

BergWerkGIS:
> So I don't think libosmium is to blame here.

I second that.
I also get the `__imp_ntohl reference error` with branch `develop` when I build locally.

Interesting that it seems to work on AppVeyor:
https://ci.appveyor.com/project/DennisOSRM/osrm-backend/build/1.0.2083

TheMarex:
Looks good. :+1: Merge away!

PR #1860, opened 2016-01-06
Remove taginfo project file

daniel-j-h:
The taginfo.json file was supposed to capture all tag usage in order to
make it visible for mappers:

> http://wiki.openstreetmap.org/wiki/Taginfo/Projects

Unfortunately this file is horribly outdated and we can not reasonably
provide such a file with car, foot, bike profiles and the user's ability
to bring-their-own profiles.

For example, compare the taginfo.json with car.lua:

> https://github.com/Project-OSRM/osrm-backend/blob/v4.9.0/profiles/car.lua


joto:
@daniel-j-h As taginfo maintainer I'd be sorry if this would be gone. I think it provides a valuable service to OSMers and OSRM users alike. Would it really be so difficult to keep this file up-to-date? I see that you can't provide this for user profiles, but it should be possible for the most often used profiles.

daniel-j-h:
By now the profiles have a massive amount of tags they use in different situations. The `taginfo.json` would have to be updated accordingly to include the union of all of them. This is doable, of course.

TheMarex:
Hm we could automatically scrape the profiles to fill the key fields in `taginfo.json`. However looking at the required information keeping the `value` field up to date would be a pain.

TheMarex:
Closing in favor of #1980

Issue #1863, opened 2016-01-06
Broken Routing for Loops on One-Way Streets


MoKob:
Following up the research on https://github.com/Project-OSRM/osrm-backend/issues/1861, it seems like the routing on one-way streets is broken.

Given the simple set-up:
 * a --> b
 * b --> c
 * c --> a

This would translate into an edge based graph **(ab) -> (bc), (bc) -> (ca), (ca) -> (ab)**.

Starting at the end of the one-way street (ab) and going to the beginning, the query has to find a self-loop within the graph **(ab) -> (bc) -> (ca) -> (ab)**, as both nodes map to the same segment (ab).

In the current system, this is partly handled by a special case [here](https://github.com/Project-OSRM/osrm-backend/blob/develop/include/engine/routing_algorithms/shortest_path.hpp#L88).
This case works only partly though, as we cannot deduct all neighbours to a vertex in a CH.

If **(ab)** is contracted last, we cannot find any arcs for **(ab)** in the CH anymore. This results in the foremost mentioned query to fail.
Due to the random nature of the contraction, this can result in random queries to fail, as long as a loop may be required (e.g. due to via points).

A fix for this problem could include self-loops within the graph for one-way streets. Instead of using the dedicated [special case handling](https://github.com/Project-OSRM/osrm-backend/blob/develop/include/engine/routing_algorithms/shortest_path.hpp#L88), we would have to check for self-loop arcs in the search.

emiltin:
is all cucumber testing disable for os x? why not just disable the offending test, if you can't simply fix the problem yet?

MoKob:
At the moment it looks like even all builds are disabled for OSX (https://github.com/Project-OSRM/osrm-backend/blob/develop/.travis.yml#L77).

Its not that I've disabled them, its that I found out about that today.

danpat:
Here's the original ticket for this:

https://github.com/Project-OSRM/osrm-backend/issues/1778

MoKob:
To document my findings so far.
The problem goes even deeper than originally though.

The problem is that in OSRM a path from a node to itself can have non-zero length.
These paths originate in the idea that every node represents a path in the original graph, combined with a turn at the end (edge-based graph).
When searching, OSRM uses phantom nodes with offsets. Lets name the node we are searching from `a`.
A phantom node contains a reference to `a` and an offset along the path, `L`.
[...]

PR #1869, opened 2016-01-08
Shortest path loops

MoKob:
Addressing https://github.com/Project-OSRM/osrm-backend/issues/1863, this pull request fixes a problem were routing actually requires a loop to work correctly.

Opposed to node based routing, starting and ending on a road can require more than just driving along the road itself. In case of one-way streets, we can end up requiring a loop that starts late on the segment and arrives early on the segment. (E.g. how do I get to the beginning of that oneway street).

### DIFF
--- a/features/testbot/via.feature
+++ b/features/testbot/via.feature
@@ -93,8 +93,6 @@ Feature: Via points
             | 1,3,2     | ab,bc,cd,cd,de,ef,fa,ab,bc | 1600m +-1 | head,straight,straight,via,right,right,right,right,straight,destination |
             | 3,2,1     | cd,de,ef,fa,ab,bc,bc,cd,de,ef,fa,ab | 2400m +-1 | head,right,right,right,right,straight,via,straight,right,right,right,right,destination |
 
-    # TODO: Remove this ignore when https://github.com/Project-OSRM/osrm-backend/issues/1863 gets fixed
-    @ignore-platform-mac
     Scenario: Via points on ring on the same oneway
     # xa it to avoid only having a single ring, which cna trigger edge cases
         Given the node map
@@ -117,7 +115,6 @@ Feature: Via points
             | 1,2,3     | ab,ab                      | 200m +-1  | head,via,destination                                             |
             | 1,3,2     | ab,ab,bc,cd,da,ab          | 1100m +-1 | head,via,right,right,right,right,destination                     |
             | 3,2,1     | ab,bc,cd,da,ab,ab,bc,cd,da,ab | 1800m     | head,right,right,right,right,via,right,right,right,right,destination |
-
 
     # See issue #1896
     Scenario: Via point at a dead end with oneway
@@ -162,3 +159,45 @@ Feature: Via points
             | waypoints | route            |
             | a,1,c     | abc,bd,bd,bd,abc |
             | c,1,a     | abc,bd,bd,bd,abc |
+
+    Scenario: Via points on ring on the same oneway, forces one of the vertices to be top node
+        Given the node map
+            | a | 1 | 2 | b |
+            | 8 |   |   | 3 |
+            | 7 |   |   | 4 |
+            | d | 6 | 5 | c |
+
+        And the ways
+            | nodes | oneway |
+            | ab    | yes    |
+            | bc    | yes    |
+            | cd    | yes    |
+            | da    | yes    |
+
+        When I route I should get
+            | waypoints | route                      | distance  | turns                                                            |
+            | 2,1       | ab,bc,cd,da,ab             | 1100m +-1  | head,right,right,right,right,destination                         |
+            | 4,3       | bc,cd,da,ab,bc             | 1100m +-1  | head,right,right,right,right,destination                         |
+            | 6,5       | cd,da,ab,bc,cd             | 1100m +-1  | head,right,right,right,right,destination                         |
+            | 8,7       | da,ab,bc,cd,da             | 1100m +-1  | head,right,right,right,right,destination                         |
+
+    Scenario: Multiple Via points on ring on the same oneway, forces one of the vertices to be top node
+        Given the node map
+            | a | 1 | 2 | 3 | b |
+            |   |   |   |   | 4 |
+            |   |   |   |   | 5 |
+            |   |   |   |   | 6 |
+            | d | 9 | 8 | 7 | c |
+
+        And the ways
+            | nodes | oneway |
+            | ab    | yes    |
+            | bc    | yes    |
+            | cd    | yes    |
+            | da    | yes    |
+
+        When I route I should get
+            | waypoints | route                      | distance  | turns                                                            |
+            | 3,2,1     | ab,bc,cd,da,ab,ab,bc,cd,da,ab | 3000m +-1    | head,right,right,right,right,via,right,right,right,right,destination |
+            | 6,5,4     | bc,cd,da,ab,bc,bc,cd,da,ab,bc | 3000m +-1    | head,right,right,right,right,via,right,right,right,right,destination |
+            | 9,8,7     | cd,da,ab,bc,cd,cd,da,ab,bc,cd | 3000m +-1    | head,right,right,right,right,via,right,right,right,right,destination |
--- a/include/contractor/contractor.hpp
+++ b/include/contractor/contractor.hpp
@@ -145,8 +145,11 @@ class Contractor
     }
 
     template <class ContainerT>
-    Contractor(int nodes, ContainerT &input_edge_list, std::vector<float> &&node_levels_)
-        : node_levels(std::move(node_levels_))
+    Contractor(int nodes,
+               ContainerT &input_edge_list,
+               std::vector<float> &&node_levels_,
+               std::vector<EdgeWeight> &&node_weights_)
+        : node_levels(std::move(node_levels_)), node_weights(std::move(node_weights_))
     {
         std::vector<ContractorEdge> edges;
         edges.reserve(input_edge_list.size() * 2);
@@ -203,8 +206,7 @@ class Contractor
             forward_edge.data.shortcut = reverse_edge.data.shortcut = false;
             forward_edge.data.id = reverse_edge.data.id = id;
             forward_edge.data.originalEdges = reverse_edge.data.originalEdges = 1;
-            forward_edge.data.distance = reverse_edge.data.distance =
-                std::numeric_limits<int>::max();
+            forward_edge.data.distance = reverse_edge.data.distance = INVALID_EDGE_WEIGHT;
             // remove parallel edges
             while (i < edges.size() && edges[i].source == source && edges[i].target == target)
             {
@@ -223,7 +225,7 @@ class Contractor
             // merge edges (s,t) and (t,s) into bidirectional edge
             if (forward_edge.data.distance == reverse_edge.data.distance)
             {
-                if ((int)forward_edge.data.distance != std::numeric_limits<int>::max())
+                if ((int)forward_edge.data.distance != INVALID_EDGE_WEIGHT)
                 {
                     forward_edge.data.backward = true;
                     edges[edge++] = forward_edge;
@@ -231,11 +233,11 @@ class Contractor
             }
             else
             { // insert seperate edges
-                if (((int)forward_edge.data.distance) != std::numeric_limits<int>::max())
+                if (((int)forward_edge.data.distance) != INVALID_EDGE_WEIGHT)
                 {
                     edges[edge++] = forward_edge;
                 }
-                if ((int)reverse_edge.data.distance != std::numeric_limits<int>::max())
+                if ((int)reverse_edge.data.distance != INVALID_EDGE_WEIGHT)
                 {
                     edges[edge++] = reverse_edge;
                 }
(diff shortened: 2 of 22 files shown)


MoKob:
General Problem is fixed.
The different search types have to be adjusted to work correctly.

 * Routing Base
- [ ] Search
- [ ] SearchWithCore
- [ ] get_network_distance

 * Alternative Path
- [ ] operator()

 * Many To Many
- [ ] operator()

 * DirectShortestPath
- [x] operator()

 * ShortestPath
- [x] operator()

MoKob:
Turns out that this problem is even more severe then I though. Self loops are a even more general thing. Will have to do a lot more work than I originally thought.

MoKob:
@daniel-j-h could you give this a look. It contains some code to serialize/deserialize flags that I did not need anymore after I had to adjust for all vertices. Maybe that part can still be helpful?

features/testbot/via.feature
```diff
@@ -117,3 +115,67 @@ Feature: Via points
             | 1,2,3     | ab,ab                      | 200m +-1  | head,via,destination                                             |
             | 1,3,2     | ab,ab,bc,cd,da,ab          | 1100m +-1 | head,via,right,right,right,right,destination                     |
             | 3,2,1     | ab,bc,cd,da,ab,ab,bc,cd,da,ab | 1800m     | head,right,right,right,right,via,right,right,right,right,destination |
+
+    Scenario: Via points on ring on the same oneway, forces one of the vertices to be top node
+    # xa it to avoid only having a single ring, which cna trigger edge cases
```
daniel-j-h: Hmm.. I do not understand that sentence :)

include/contractor/contractor.hpp
```diff
@@ -718,6 +718,50 @@ class Contractor
...
+            // Found a shorter Path -> Update distance
+            else if (to_distance < heap.GetKey(to))
+            {
+                heap.DecreaseKey(to, to_distance);
+                heap.GetData(to).hop = current_hop;
+            }
+        }
+    }
+
+    inline int FindSelfLoop(const NodeID node, int max_distance, ContractorThreadData *const data)
```
daniel-j-h: Why by pointer and not ref?

include/contractor/contractor.hpp
```diff
@@ -718,6 +718,50 @@ class Contractor
...
+    {
+        ContractorHeap &heap = data->heap;
```
daniel-j-h: Can be auto

include/contractor/contractor.hpp
```diff
@@ -718,6 +718,50 @@ class Contractor
...
+        heap.Clear();
+        heap.Insert(node, INT_MAX, ContractorHeapData(0, true)); // mark the node itself as a target
```
daniel-j-h: Hmm int max here depends on the value type used for representation.. this is kind of ugly :/

include/contractor/contractor.hpp
```diff
@@ -718,6 +718,50 @@ class Contractor
...
+        RelaxNode(node, SPECIAL_NODEID, 0, heap);                // add its neighbors to the heap
+        Dijkstra(max_distance, 1, 1000, data, SPECIAL_NODEID);
```
daniel-j-h: Where do the magic constants for number of targets and especially max nodes come from?
Would be great to at least symbolize them.

include/contractor/contractor.hpp
```diff
@@ -833,6 +852,50 @@ class Contractor
                 continue;
             }
 
+            // Check for Potential Self-Loops, only add if self-loop is the shortest loop
+            for (auto out_edge : contractor_graph->GetAdjacentEdgeRange(node))
+            {
+                const ContractorEdgeData &out_data = contractor_graph->GetEdgeData(out_edge);
```
daniel-j-h: Could be `const auto&`

include/contractor/contractor.hpp
```diff
@@ -833,6 +852,50 @@ class Contractor
...
+                    if (source == target)
+                    {
+                        // TODO we could prune some of these self-loops if we new about the weight
+                        // along the
+                        // edge represented by `node` itself. If the self-loop is longer than the
+                        // edge,
+                        // the self loop can be pruned as well
+                        // Check for Self Loops
+                        const int this_loop_distance = in_data.distance + out_data.distance;
+                        int self_loop_distance = FindSelfLoop(source, this_loop_distance, data);
```
daniel-j-h: As far as I can see could also be const

include/contractor/contractor.hpp
```diff
@@ -833,6 +852,50 @@ class Contractor
...
+                            {
+                                inserted_edges.emplace_back(source, target, this_loop_distance,
+                                                            out_data.originalEdges +
+                                                                in_data.originalEdges,
+                                                            node, true, true, false);
+
+                                inserted_edges.emplace_back(target, source, this_loop_distance,
+                                                            out_data.originalEdges +
+                                                                in_data.originalEdges,
+                                                            node, true, false, true);
```
daniel-j-h: This and above is hard to follow as I have to 1/ look for inserted_edges' value type, 2/ then check which position in the ContractorEdgeData constructor these values correspond to. Hmm probably better to symbolize them, that is create some named booleans and then pass those (especially because only source/target and backward changes).

Would be great to have named parameters á la: fn(forward: true, backward: false) --- oh well :/

include/engine/routing_algorithms/alternative_path.hpp
```diff
@@ -382,10 +382,12 @@ class AlternativeRouting final
         int upper_bound_s_v_path_length = INVALID_EDGE_WEIGHT;
         new_reverse_heap.Insert(via_node, 0, via_node);
         // compute path <s,..,v> by reusing forward search from s
+        const bool constexpr STALLING_ENABLED = true;
+        const bool constexpr DO_NOT_FORCE_LOOPS = false;
```
daniel-j-h: Nice!

include/engine/routing_algorithms/routing_base.hpp
```diff
@@ -573,29 +634,46 @@ template <class DataFacadeT, class Derived> class BasicRoutingInterface
...
-            RetrievePackedPathFromHeap(forward_heap, reverse_heap, middle, packed_leg);
+            // we need to unpack sub path from core heaps
+            if (facade->IsCoreNode(middle))
+            {
+                std::vector<NodeID> packed_core_leg;
+                RetrievePackedPathFromHeap(forward_core_heap, reverse_core_heap, middle,
+                                           packed_core_leg);
+                BOOST_ASSERT(packed_core_leg.size() > 0);
+                RetrievePackedPathFromSingleHeap(forward_heap, packed_core_leg.front(), packed_leg);
+                std::reverse(packed_leg.begin(), packed_leg.end());
```
daniel-j-h: needs algorithm header; looking at the file, several headers are missing, e.g. vector, iterator, utility, limits, cstddef

include/engine/routing_algorithms/routing_base.hpp
```diff
@@ -573,29 +634,46 @@ template <class DataFacadeT, class Derived> class BasicRoutingInterface
...
     }
 
+    //Requires the heaps for be empty
+    //If heaps should be adjusted to be initialized outside of this function,
+    //the addition of force_loop parameters might be required
     double get_network_distance(SearchEngineData::QueryHeap &forward_heap,
                                 SearchEngineData::QueryHeap &reverse_heap,
                                 const PhantomNode &source_phantom,
-                                const PhantomNode &target_phantom) const
+                                const PhantomNode &target_phantom ) const
```
daniel-j-h: clang-format on all the files

include/engine/routing_algorithms/routing_base.hpp
```diff
@@ -573,29 +634,46 @@ template <class DataFacadeT, class Derived> class BasicRoutingInterface
...
     {
+        BOOST_ASSERT( forward_heap.Size() == 0 );
+        BOOST_ASSERT( reverse_heap.Size() == 0 );
```
daniel-j-h: .Empty

include/engine/routing_algorithms/shortest_path.hpp
```diff
@@ -255,7 +169,10 @@ class ShortestPathRouting final
             BOOST_ASSERT(forward_heap.Size() > 0);
             BOOST_ASSERT(reverse_heap.Size() > 0);
             super::Search(forward_heap, reverse_heap, new_total_distance_to_reverse,
-                          leg_packed_path_reverse);
+                          leg_packed_path_reverse, false,
+                          source_phantom.reverse_node_id == target_phantom.reverse_node_id &&
+                              source_phantom.GetReverseWeightPlusOffset() >
+                                  target_phantom.GetReverseWeightPlusOffset());
```
daniel-j-h: Wait.. this and above is your forceLoop lambda body, isn't it?

include/extractor/extractor.hpp
```diff
@@ -11,14 +11,15 @@ namespace osrm
...
-class extractor
+class Extractor
 {
   public:
-    extractor(ExtractorConfig extractor_config) : config(std::move(extractor_config)) {}
+    Extractor(ExtractorConfig extractor_config) : config(std::move(extractor_config)) {}
     int run();
 
-  private:
+  protected:
```
daniel-j-h: Why protected?

include/util/io.hpp
```diff
@@ -0,0 +1,70 @@
...
+             ++chunk_bit, ++bit_nr)
+            chunk_bitset[chunk_bit] = flags[bit_nr];
+
+        chunk = chunk_bitset.to_ulong();
+        ++chunk_count;
+        flag_stream.write(reinterpret_cast<const char *>(&chunk), sizeof(chunk));
+    }
+    util::SimpleLogger().Write() << "Wrote " << number_of_bits << " bits in " << chunk_count
+                                 << " chunks (Flags).";
+    return !flag_stream.fail();
```
daniel-j-h: Hmm I would be careful with explicitely checking states here; if you just want to verify the I/O operation went fine and the stream is okay, then force the stream to a boolean context, such as `if(stream) ok();`

include/util/io.hpp
```diff
@@ -0,0 +1,70 @@
...
+    for (std::size_t chunk_id = 0; chunk_id < chunks; ++chunk_id)
+    {
+        flag_stream.read(reinterpret_cast<char *>(&chunk), sizeof(chunk));
+        std::bitset<32> chunk_bits(chunk);
+        for (std::size_t bit = 0; bit < 32 && bit_position < number_of_bits; ++bit, ++bit_position)
+            flags[bit_position] = chunk_bits[bit];
+    }
+    util::SimpleLogger().Write() << "Read " << number_of_bits << " bits in " << chunks
+                                 << " Chunks from disk.";
+    return !flag_stream.fail();
```
daniel-j-h: Same here

src/contractor/processing_chain.cpp
```diff
@@ -143,10 +147,12 @@ std::size_t Prepare::LoadEdgeExpandedGraph(
...
-    size_t number_of_edges = 0;
-    size_t max_edge_id = SPECIAL_EDGEID;
-    input_stream.read((char *)&number_of_edges, sizeof(size_t));
-    input_stream.read((char *)&max_edge_id, sizeof(size_t));
+    // TODO std::size_t can vary on systems. Our files are not transferable, but we might want to
+    // consider using a fixed size type for I/O
+    std::size_t number_of_edges = 0;
+    std::size_t max_edge_id = SPECIAL_EDGEID;
+    input_stream.read((char *)&number_of_edges, sizeof(std::size_t));
+    input_stream.read((char *)&max_edge_id, sizeof(std::size_t));
```
daniel-j-h: Looking at `ag -Q 'sizeof(unsigned'` we're writing size prefixes as unsigneds most if not all of the time

include/contractor/contractor.hpp
MoKob: mostly due to consistency. It was used like this everywhere else in the contractor. That thing really requires a refactor.

MoKob:
While this works, it significantly increases preprocessing time (Germany 662 -> 781 seconds).

A possible way to improve this would be to remember self-loop prune distances based on the length of the ways represented in the edge based nodes, and INFTY for one-ways. This requires updating these values in case of traffic updates as well. @TheMarex @danpat @daniel-j-h opinions on this?

MoKob:
Got the increase down to 697.7 from 662 (previously 781) on the German road network.
After review, this should solve the loop problems.

MoKob:
With the new values, sadly the performance is back down to 746.107 seconds.

MoKob:
Last commit got it down to 700s again. Yay me.

include/contractor/contractor.hpp
```diff
@@ -846,22 +869,65 @@ class Contractor
...
+                            stats->original_edges_added_count +=
+                                2 * (out_data.originalEdges + in_data.originalEdges);
+                        }
+                        else
+                        {
+                            // CAREFUL: This only works due to the independent node-setting. This
+                            // guarantees that source is not connected to another node that is
+                            // contracted
+                            node_weights[source] = loop_weight + 1;
+                            node_weights[source] = loop_weight; // make sure to prune better
```
TheMarex: This looks like a typo, ` node_weights[source] ` gets assigned twice.

include/contractor/contractor.hpp
```diff
@@ -846,22 +869,65 @@ class Contractor
...
+
+                            inserted_edges.emplace_back(
+                                target, source, loop_weight,
+                                out_data.originalEdges + in_data.originalEdges, node, SHORTCUT_ARC,
+                                FORWARD_DIRECTION_DISABLED, REVERSE_DIRECTION_ENABLED);
+                        }
+                    }
+                    continue;
+                }
                 const int path_distance = in_data.distance + out_data.distance;
```
TheMarex: `loop_weight` and `path_distance` seem to always have the same value, but are declared in different scopes. Move `path_distance` up.

include/engine/routing_algorithms/many_to_many.hpp
```diff
@@ -143,10 +143,30 @@ class ManyToManyRouting final
                     (*result_table)[source_id * number_of_targets + target_id];
                 // check if new distance is better
                 const EdgeWeight new_distance = source_distance + target_distance;
-                if (new_distance >= 0 && new_distance < current_distance)
+                if (new_distance < 0)
                 {
-                    (*result_table)[source_id * number_of_targets + target_id] =
-                        (source_distance + target_distance);
+                    for (auto edge : super::facade->GetAdjacentEdgeRange(node))
```
TheMarex: I feel this should be a utility function in `routing_base.hpp`. Something like:

`UpdateLoopWeight(NodeId node, EdgeWeight& loop_weight)`

Also can't we guarantee that each node has at most one relevant loop? In that case we can optimize this.

MoKob: We could increase the memory overhead to have a lookup. Since this should only happen at the start of a segment, I guess the overhead to compute it, is worth it to save the additional entry for every node.

Moved the lookup to Routing base.

PR #1870, opened 2016-01-08
Make classes with only a single funtion free standing functions

daniel-j-h:
- [X] DouglasPeuker
- [x] PolylineCompressor
- [x] PolylineFormatter
- [x] RouteNameExtraction
- [x] TurnInstructionsClass
- [x] ComputeAngle


src/engine/douglas_peucker.cpp
```diff
@@ -59,13 +59,15 @@ struct CoordinatePairCalculator
 };
 }
 
-void DouglasPeucker::Run(std::vector<SegmentInformation> &input_geometry, const unsigned zoom_level)
+void DouglasPeucker(std::vector<SegmentInformation>::iterator begin,
```
TheMarex: Leading char should be small case like: `douglasPeucker` for free functions.

daniel-j-h: Urgh yeah have to get used to that again. Also having different conventions between member functions and free standing function is somewhat ugly for refactoring :/

daniel-j-h:
Done. Converted to free standing functions. At the same time:

- fixing includes in the files I touched
- fixing two occurrences of Undefined Behavior when instantiating templates with incomplete type
- made sure compile-time constants are guaranteed to be ROM-able

/cc @TheMarex

TheMarex:
@daniel-j-h seems like it broke builds on windows. Investigating.

daniel-j-h:
> 'not': undeclared identifier

https://github.com/Project-OSRM/osrm-backend/pull/1871

TheMarex:
Rebased on #1871, merge after that PR.

PR #1871, opened 2016-01-08
replaces and/not with &&/! for windows

MoKob:
just as the title says...


daniel-j-h:
Interesting. Does Windows not support these?
http://en.cppreference.com/w/cpp/language/operator_alternative

MoKob:
Well. According to all the failing AppVeyor builds, it strongly looks like it. At least visual studio does not support them, as far as I remember from my personal experience.

danpat:
According to this:

http://stackoverflow.com/a/6006540

Visual C++ will support them when it's in "standards compliance" mode, enabled by adding `/Za` to the command line.

Looks like we're building under `/Ze` by default, we might want to switch to `/Za`:

https://msdn.microsoft.com/en-us/library/34h23df8.aspx#Anchor_9

PR #1874, opened 2016-01-11
Fix comment format warning

giraldeau:
Multi-line comments with slash prefixed should not contain slashes. (-Wcomment)

Signed-off-by: Francis Giraldeau <francis.giraldeau@gmail.com>


TheMarex:
Thanks for your pull request @giraldeau!. We only accept pull requests to `develop` and it should be already fixed there. Closing here. Sorry for not catching this before `4.9.0`.

giraldeau:
Thanks for the follow-up!

PR #1878, opened 2016-01-11
Inline mercator transformations; no need for separate translation unit

daniel-j-h:
Does what it says: no need for mercator to be a struct with two static functions, a namespace suffices (looks like we missed that in the other PR). No need for a header / impl. split for these one liners; let the optimizer see the body. Oh, and lat2y -> latToY, y2lat -> yToLat.

### DIFF
--- a/include/util/mercator.hpp
+++ b/include/util/mercator.hpp
@@ -1,17 +1,25 @@
 #ifndef MERCATOR_HPP
 #define MERCATOR_HPP
 
+#include <cmath>
+
 namespace osrm
 {
 namespace util
 {
-
-struct mercator
+namespace mercator
 {
-    static double y2lat(const double value) noexcept;
 
-    static double lat2y(const double latitude) noexcept;
-};
+inline double yToLat(const double value) noexcept
+{
+    return 180. * M_1_PI * (2. * std::atan(std::exp(value * M_PI / 180.)) - M_PI_2);
+}
+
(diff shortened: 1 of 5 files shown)


include/util/mercator.hpp
```diff
@@ -1,17 +1,25 @@
...
 {
-    static double y2lat(const double value) noexcept;
+    return 180. * M_1_PI * (2. * std::atan(std::exp(value * M_PI / 180.)) - M_PI_2);
+}
 
-    static double lat2y(const double latitude) noexcept;
-};
+inline double latToY(const double latitude) noexcept
+{
+    return 180. * M_1_PI * std::log(std::tan(M_PI_4 + latitude * (M_PI / 180.) / 2.));
```
MoKob: `180. * M_1_PI * std::log(std::tan(M_PI_4 + latitude * (M_PI / 90.)))` ?

daniel-j-h: Yes, but does not make a difference since both sub expressions are calculated at compile time and
```python
>>> (math.pi / 180) / 2.
0.008726646259971648
>>> math.pi / 360.
0.008726646259971648
```
so we do not even get better precision out of it.

MoKob: Not really. Just a readability thing. But I am fine with a merge either way.

TheMarex:
:+1: Looks good to me.

PR #1879, opened 2016-01-11
Remove headers with no reverse dependencies

daniel-j-h:
I initially wanted to fix an issue in `FixedPointCoordinate`, where `const T&&` is never a universal (forwarding) reference, but always a rvalue reference, and therefore `std::forward` does not make sense. See [here](https://isocpp.org/blog/2012/11/universal-references-in-c11-scott-meyers) why). Turns out, it has no reverse dependencies.

Found another one with

```
daniel@x1c /t/o/i/util> for header in *.hpp
                            echo -n $header:
                            ag -iQ $header ../../ --noheading | wc -l
                        end
...
fixed_point_number.hpp:0
range_algorithms.hpp:0
...
```

Only two. Surprises me to be honest.
But we're cleaning up for quite a bit now.


TheMarex:
@daniel-j-h who came up with `FixedPointNumber`? I think the idea was to use it in `FixedPointCoordinate` to make the encoding as `int` transparent.

I would like to use a similar approach like in `rtd`.

TheMarex:
Is there a reliable way to find utility functions that are not used? Some headers contain quite a few of them.

daniel-j-h:
Not that I know of; there is iwyu for headers that works most of the time. Hmm time to build something libclang-based?

TheMarex:
@daniel-j-h hm. this should be possible with the AST right? Track all symbols references by a file, collect them over the code base and check against all declared symbols.

daniel-j-h:
I would have to look into this again and in detail, but it sounds doable.
The only problem that I encountered in using libclang is that the pre-processor already ran.
That is, you operate on the granularity of translation units and for example can no longer differentiate between header and implementation file. Hooking the pre-processor is not doable with libclang, we would have to use LLVM directly for that. And that's ugly since they have no API stability guarantees, also lots and lots of C++ boilerplate needed, and so on. I will think about a basic approach using libclang tomorrow.

daniel-j-h:
@TheMarex this is what I came up with: https://gist.github.com/daniel-j-h/6f7b58236bf0517167f1

- get all function symbols S
- get all function call expressions C
- pipe through `sort | uniq`, then `comm` to get S \ C the set of unused symbols

this is quite hacky but works on my local small example.
Running it on osrm I can see function calls missing --- investigating.

daniel-j-h:
Merge incoming! In case we really switch to a fixed point number type, we can either pull the wrapper out of Git or evaluate an implementation for it. AST matching postponed to another PR.

Issue #1604, opened 2015-08-18
Rename import_edge.hpp


TheMarex:
It contains two structs and neither is named `ImportEdge`.

PR #1880, opened 2016-01-11
Refactor import edge

daniel-j-h:
Splits import_edge.hpp into node_based_edge.hpp, edge_based_edge.hpp. No separate cpp files, since the definitions are small anyway. Fixes a few includes while we're at it. Closes https://github.com/Project-OSRM/osrm-backend/issues/1604.


TheMarex:
This is great.:+1: Ready to merge.

Issue #1660, opened 2015-09-08
`noexcept` specifier placed on functions that may indeed throw


daniel-j-h:
First example I could find: [bearing.hpp](https://github.com/Project-OSRM/osrm-backend/blob/fd6daa580a7d3c7e2074a9c6ca1da41268274d26/util/bearing.hpp#L35), marks the `get` function with `noexcept`, guaranteeing to never throw. If the function does throw, the program gets terminated.

The problem here is that `std::string` can indeed throw in its constructor (e.g. in the out of memory case). This is just an example based on the first occurrence I could fine --- have to check all of them.

References:

- C++14 standard: 15.4 "Exception specifications" [except.spec]
- Effective Modern C++: "Item 14: Declare functions noexcept if they won’t emit exceptions."
- http://en.cppreference.com/w/cpp/language/noexcept_spec

TheMarex:
Can we check this in an automated case?

daniel-j-h:
From the top of my head I don't think there is an automated way to do it.

There's also the [`noexcept` operator](http://en.cppreference.com/w/cpp/language/noexcept) that allows for conditionally `noexcept`, such as `noexcept(fn())`.

daniel-j-h:
I fixed this in #1882.

PR #1882, opened 2016-01-11
Rips out dead code from the container.hpp and creates for_each_pair.hpp

daniel-j-h:
Note: after merging https://github.com/Project-OSRM/osrm-backend/issues/1660 is resolved; close the issue.

I manually checked all the occurrences in the code base (phew..); where it is placed but an exception might get thrown, we might as well take the `std::terminate` that comes with it (e.g. in not being able to log). I also added two TODOs that I would like to address in the future.

### DIFF
--- a/include/util/for_each_pair.hpp
+++ b/include/util/for_each_pair.hpp
@@ -1,0 +1,45 @@
+#ifndef FOR_EACH_PAIR_HPP
+#define FOR_EACH_PAIR_HPP
+
+#include <numeric>
+#include <iterator>
+
+namespace osrm
+{
+namespace util
+{
+
+// TODO: check why this is not an option here:
+// std::adjacent_find(begin, end, [=](const auto& l, const auto& r){ return function(), false; });
+template <typename ForwardIterator, typename Function>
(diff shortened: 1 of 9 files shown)


include/util/for_each_pair.hpp
```diff
@@ -0,0 +1,45 @@
...
+    }
+
+    auto next = begin;
+    next = std::next(next);
+
+    while (next != end)
+    {
+        function(*begin, *next);
+        begin = std::next(begin);
+        next = std::next(next);
```
MoKob: just nit-pitting, but I've read somewhere that it would be best to bring these kind of functions into the namespace (`using std::next`) so that the compiler can decide which next to use, if there should be some better fitting one.
Just as you've done below for begin and end.

daniel-j-h: Hmm for begin / end I can understand the need for Koenig lookup (so that it e.g. works on raw arrays); next should eventually dispatch to operator++ (for forward iterators), and that the iterator concept guarantees for and lets the user implement it in any way she wishes.

PR #1887, opened 2016-01-12
Define compile-time constants for coordinate calculation only once

### LINKED ISSUE #1327: douglas peucker: move rad and earth_radius into header file
constants are repeated: move into header file that encapsulates constants in an (anonymous) namespace.

daniel-j-h:
In addition, make them more precise (probably does not matter). Closes #1327.


TheMarex:
Good find. :+1: Looks good to me. Please merge.

PR #1889, opened 2016-01-13
Fix race condition in osrm-routed HTTP server

daniel-j-h:
Stumbled upon this one by accident.

It goes like this:

- `output_buffer` is function-local
- we pass it to async_write and leave the scope
- `output_buffers` goes out of scope
- bad things happen, sometimes

The fix is to, again, put it inside the connection that handles itself
via the enable_shared_from_this idiom.

We had a similar issue a few month ago --- argh routed!

/cc @danpat @TheMarex unrelated to the current datastore problem

References:

- http://www.boost.org/doc/libs/1_59_0/doc/html/boost_asio/reference/async_write/overload1.html
- https://github.com/Project-OSRM/osrm-backend/pull/1690

### DIFF
--- a/include/server/connection.hpp
+++ b/include/server/connection.hpp
@@ -63,6 +63,8 @@ class Connection : public std::enable_shared_from_this<Connection>
     http::request current_request;
     http::reply current_reply;
     std::vector<char> compressed_output;
+    // Header compression_header;
+    std::vector<boost::asio::const_buffer> output_buffer;
 };
 }
 }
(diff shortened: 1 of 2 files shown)


TheMarex:
Uhm didn't we fix _exactly_ this bug already? Did it get reverted by accident.

daniel-j-h:
Yeah there was the same issue in more or less the same location; really liked the high level Python server approach that we went for in rtd. Maybe not perfect, but at least not another source of critical issues.

PR #1890, opened 2016-01-13
Fix debian package upgrades.

zimathias:
The postinst script fails on package upgrade given that the symlink
already exists.

Note that symlinks should *not* be handled in a postinst scripts.
Instead dh_links should be used so that symlinks are compliant with the
Debian policy.


TheMarex:
Thanks! :+1:

PR #1891, opened 2016-01-13
Fix wrong assertion and emplace_back bug in trip plugin

TheMarex:


include/engine/plugins/trip.hpp
```diff
@@ -291,13 +287,14 @@ template <class DataFacadeT> class RoundTripPlugin final : public BasePlugin
         {
             const auto component_size = scc.range[k + 1] - scc.range[k];
 
-            BOOST_ASSERT_MSG(component_size >= 0, "invalid component size");
+            BOOST_ASSERT_MSG(component_size > 0, "invalid component size");
+
+            std::vector<NodeID> scc_route;
```
MoKob: as this is only used to transfer it into route_result, wouldn't it be more efficient to directly emplace_back the different choices (e.g. `route_results.emplace_back(trip::BruteFroceTrip())`) instead of first constructing them into scc_route and then moving them out?

TheMarex:
Arg `osrm-routed` is crashing when build on clang. WTF. @daniel-j-h any idea what could be going on here? Works for me locally.

include/engine/plugins/trip.hpp
```diff
@@ -145,7 +137,7 @@ template <class DataFacadeT> class RoundTripPlugin final : public BasePlugin
             // whole graph is one scc
             std::vector<NodeID> location_ids(number_of_locations);
             std::iota(std::begin(location_ids), std::end(location_ids), 0);
-            return SCC_Component(std::move(location_ids));
+            return SCC_Component(std::move(location_ids), {0, location_ids.size()});
```
MoKob: This is the same problem we had in the prototype. Moving in the second part breaks. The SCC_Component will see {0,0} instead of {0. location_ids.size()}.

MoKob:
An additional problem I've seen during my research of what is going wrong here:

https://github.com/Project-OSRM/osrm-backend/blob/fix/trip_two_coords/include/engine/plugins/trip.hpp#L213

If the route has length 0, this part does an access violation.

include/engine/plugins/trip.hpp
MoKob: To solve this, you have to explicitly construct the range parameter (second parameter) and move it in.

include/engine/plugins/trip.hpp
daniel-j-h: `route_results` is a `vector<vector<>>`, so `.emplace_back` would only construct a `vector<>` inplace, and as far as I can see, there is no way a `vector<>` could be meaningfully initialized via `BruteForceTrip()`.

`vector<vector<>>` are not perfect anyway, since they potentially over allocate in both dimensions (fortunately we can `.reserve` here so that's not too bad)

TheMarex:
@MoKob great catch! Should be fixed now hopefully.

include/engine/plugins/trip.hpp
```diff
@@ -145,7 +137,7 @@ template <class DataFacadeT> class RoundTripPlugin final : public BasePlugin
             // whole graph is one scc
             std::vector<NodeID> location_ids(number_of_locations);
             std::iota(std::begin(location_ids), std::end(location_ids), 0);
-            return SCC_Component(std::move(location_ids));
+            return SCC_Component(std::move(location_ids), std::vector<size_t> {{0, location_ids.size()}});
```
MoKob: Given that the cases still fail, I remember that even constructing like this did not work in the prototype.
I had to explicitly construct the vector prior to the return statement and move it in.
`std::vector<size_t> range = {0.location_ids.size()};`
`return SCC_Component(std::move(location_ids), std::move(range));`

daniel-j-h: Wait, now the problem is a bit different, isn't it? What we are now running into is the difference between

    vector<int> x(1, 42); // (count, value) ctor - vector of one initalized int with value 42

and

    vector<int> x{1, 42}; // vector of two ints with value 1 and 42

FYI, small trick to make things like this easier to test: on freenode there are bots - `geordi` and `clang` - that let you quickly test those small snippets, e.g.

    { vector<size_t> x = {0, 42}; cout << x; }
    <geordi> {0, 42}

Just in case you're on freenode anyway.

Reference:
- http://en.cppreference.com/w/cpp/container/vector/vector

MoKob: Yeah. Typing in code and typing in git... In the code I tested I had `vector<size_t> range{0,size}`. But you are right.

Nevertheless, create vector prior to moving it in.

Issue #1888, opened 2016-01-13
Deadlock using `osrm-datastore` while requests are being served


danpat:
Under some circumstances, it seems that `osrm-datastore` can deadlock if routing requests are being served at the same time.

To reproduce the problem:

  1. Run `osrm-datastore whatever.osrm` to load the initial dataset
  2. Run `osrm-routed -s` to load data from shared memory.
  3. Set up a request running in a rapid loop: `while true ; do curl 'http://localhost:5000/viaroute?loc=39.22693426244916,-75.59280395507812&loc=39.06451486901886,-75.465087890625' ; done`
  4. Now while that loop is running, do the same with `osrm-datastore`, like so:  `while true ; ./osrm-datastore whatever.osrm ; done`

In a perfect world, the locking would work properly and routing wouldn't be interrupted.  However, very shortly after starting the `osrm-datastore` loop, I start seeing errors like this:

![screen shot 2016-01-12 at 3 58 55 pm](https://cloud.githubusercontent.com/assets/1892250/12280774/69d8ad0c-b945-11e5-8851-3042cccd271b.png)

![screen shot 2016-01-12 at 4 03 12 pm](https://cloud.githubusercontent.com/assets/1892250/12280855/0c02063c-b946-11e5-95c5-dccc067364b1.png)

Then, within a few seconds, everything stops completely, and `osrm-routed` appears to be stuck in a busy loop.

Killing the `curl` loop does not unlock things.  `osrm-routed` responds to requests with `{"status": 500,"status_message":"Internal Server Error"}` and does not close the socket (leading to `curl` hanging).

Killing the `osrm-datastore` loop does not resurrect `osrm-routed`.

Killing the `osrm-datastore` loop and restarting `osrm-routed` does seem to bring things back to life, so it doesn't look like any routing data is getting corrupted, just locking state.

danpat:
Simpler reproduction recipe:

   1.  Run `osrm-datastore whatever.osrm`
   2.  Run `osrm-routed -s`
   3.  Run `osrm-datastore whatever.osrm`
   4.  Run `osrm-datastore whatever.osrm`
   5.  Run `curl http://localhost:5000/viaroute?loc=1,1&loc=2,2`

You can now kill `osrm-routed`, and restarting it (without running further `osrm-datastore` commands) will lead to errors like this:

```
danpat@rundle:~/mapbox/osrm-backend/build$ ./osrm-routed -s
[info] starting up engines, v4.9.0
[...]

PR #1892, opened 2016-01-14
Don't delete shared memory segments if we're about to use them

danpat:
If `osrm-datastore` is run an even number of times, it will set the layout back to the currently used one.  This code changes the `SharedDataFacade` to check if the new layout is still referring to the same one.  The re-initialization is only done if the timestamp has also changed.

This is a partial fix to https://github.com/Project-OSRM/osrm-backend/issues/1888

It fixes the "simple reproduction recipe", but I can still cause errors if I run `osrm-datastore` and `curl` in tight loops.  However, running `osrm-datastore` in a tight loop seems like a really oddball case.

### DIFF
--- a/include/datastore/shared_memory_factory.hpp
+++ b/include/datastore/shared_memory_factory.hpp
@@ -77,6 +77,7 @@ class SharedMemory
 
     SharedMemory() = delete;
     SharedMemory(const SharedMemory &) = delete;
+    SharedMemory &operator=(const SharedMemory &) = delete;
 
     template <typename IdentifierT>
     SharedMemory(const boost::filesystem::path &lock_file,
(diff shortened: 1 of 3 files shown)


danpat:
Hmm, this doesn't seem to fix all cases.  If I put `osrm-routed` under heavy load, then running `osrm-datastore` can still cause `osrm-routed` to hang (with high CPU usage).

danpat:
Additional changes clear up a lot of the problems.  However, if I run `osrm-datastore` very rapidly, I still seem to be able to trigger exceptions in `osrm-routed` if it's also under heavy load.  However, these instances are massively decreased compared to before, so this fix is useful.

This adds slight locking overhead to queries when using the SharedDataFacade.  Unfortunately, it's unavoidable overhead if we want correct behaviour, unless we want to investigate writing some kind of lock-free implementation, which is major work.

PR #1897, opened 2016-01-15
U turns at via

MoKob:
The current route description does not emit a u-turn instruction in case of a u-turn at a via node.

This pull request fixes this issue and adjusts the test-cases accordingly.


include/engine/guidance/segment_list.hpp
```diff
@@ -102,6 +102,23 @@ SegmentList<DataFacadeT>::SegmentList(const InternalRouteResult &raw_route,
                    raw_route.segment_end_coordinates[raw_index].target_phantom,
                    raw_route.target_traversed_in_reverse[raw_index],
                    raw_route.is_via_leg(raw_index), facade);
+            if (raw_route.is_via_leg(raw_index))
```
TheMarex: Indentation seems off?

TheMarex:
Good catch @MoKob! Looks good to me. :+1:

include/engine/guidance/segment_list.hpp
TheMarex: Ah was actually fine. :+1:

Issue #1896, opened 2016-01-14
Via-Routing error in Disney Land


TheMarex:
http://map.project-osrm.org/?z=15&center=28.376732,-81.501732&loc=28.367453,-81.519472&loc=28.367986,-81.519370&loc=28.367627,-81.518576&hl=en&ly=&alt=&df=&srv=

![screenshot from 2016-01-14 19-19-19](https://cloud.githubusercontent.com/assets/606968/12333320/d124192e-baf3-11e5-80b8-54041980fae5.png)

Seemy like it never does a U-Turn on this way:

https://www.openstreetmap.org/way/340766783

Even tough its bi-directional. Entering leaving the way works:

http://map.project-osrm.org/?z=18&center=28.367932,-81.517793&loc=28.367453,-81.519472&loc=28.367986,-81.519370&hl=en&ly=&alt=&df=&srv=

http://map.project-osrm.org/?z=18&center=28.367719,-81.519445&loc=28.367986,-81.519370&loc=28.367453,-81.519472&hl=en&ly=&alt=&df=&srv=

Maybe the `EdgeBasedGraphFactory` doesn't add a U-Turn edge? Or the via-route code is broken.

TheMarex:
Further investigation shows that there is now U-Turn edge that would allow to turn at the via, thus completing this route. Probably a bug in `osrm-extract`

MoKob:
The u-turn is part of the routing algorithm, isn't it? The phantom node should capture both forward and backward segment. In the via-part of the [query](https://github.com/Project-OSRM/osrm-backend/blob/develop/include/engine/routing_algorithms/shortest_path.hpp#L356) , u-turns at `vias` have to be explicitly allowed. Could this be related to u-turns not being allowed?

MoKob:
Was able to confirm it: its not a routing error but the avoidance of u-turns at via-nodes. Using
http://router.project-osrm.org/viaroute?z=15&loc=28.367453,-81.519472&loc=28.367986,-81.519370&u=true&loc=28.367627,-81.518576
will find a route.

TheMarex:
Should be fixed now.

PR #1899, opened 2016-01-15
Fix/uturn entrance oneways

TheMarex:
Fixes issue #1896.


danpat:
Indenting on this change seems off.  Is that intended?

TheMarex:
@danpat good catch.

PR #1904, opened 2016-01-18
Explicitely run modernizer on repo. files

daniel-j-h:
Fixes running the modernizer only on cpp files; also full run on codebase with my local Clang 3.6 installation --- @TheMarex or @MoKob do you have a Clang 3.8 / trunk installation around by chance and want to give the scripts a go, just to see if anything would change?


TheMarex:
:+1: merging

Issue #1908, opened 2016-01-19
Random size of elemens in route_instructions, no keys too


maximd1:
Hi Team,

I use osrm-backend 4.8.0 for my application. The issue I am facing with is that route_instructions of server response contains an array of arrays with random number of sub-elements - I catch IndexOutOfBoundsException when I try to access them.

I work with json.simple to parse JSON response and iterate through the each sub-element of route_instructions. According to https://github.com/Project-OSRM/osrm-backend/wiki/Server-api every entry can contain up to 11 elements. Example shows that there are only 9 (and even 8!) elements in the response. My live server also doesn't return full response - only 9 fields. Why (and which?) rest of fields are ignored?

The problem compounded by the fact that ordering of fields in the response not guaranteed too - I am not sure that, for example, driving directions code is a strictly first parameter. So, keys are very, very desirable. Maybe, there is a special undocumented parameter for the /viaroute service to switch them on in a response?

Thank you.

danpat:
Hi @maximd1 

Parameters 10 and 11 are new in 4.9.0, I'll add that to the wiki page, sorry that it's not clear.

The `instruction` entry with only 8 values (10 in 4.9.0) is the last instruction - this looks like a bug here:   https://github.com/Project-OSRM/osrm-backend/blob/develop/include/engine/guidance/textual_route_annotation.hpp#L115-L124, it looks like we've forgotten an entry for the `travel_mode`.

/cc @TheMarex

maximd1:
@danpat 
Thank you for your help, the issue is resolved now.

PR #1909, opened 2016-01-19
Include travel mode in last instruction

danpat:
The last instruction in the `instructions` block is missing the `travel_mode` parameter.  This change copies that value from the previous instruction.

This is to address https://github.com/Project-OSRM/osrm-backend/issues/1908

### DIFF
--- a/include/engine/guidance/textual_route_annotation.hpp
+++ b/include/engine/guidance/textual_route_annotation.hpp
@@ -41,6 +41,7 @@ inline util::json::Array AnnotateRoute(const std::vector<SegmentInformation> &ro
 
     round_about = {std::numeric_limits<std::int32_t>::max(), 0, 0};
     std::string temp_dist, temp_length, temp_duration, temp_bearing, temp_instruction;
+    extractor::TravelMode last_travel_mode = TRAVEL_MODE_DEFAULT;
 
     // Generate annotations for every segment
     for (const SegmentInformation &segment : route_segments)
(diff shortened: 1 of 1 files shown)


maximd1:
@danpat
Any workaround for 4.8.0, please?

Thank you.

TheMarex:
@maximd1 upgrade to `4.9.0` please.

danpat:
@maximd1 the order of the variables is consistent.  As long as you're not using the 8th or 9th parameters, you should be fine.  If you need them for some reason, and you can't upgrade to the latest codebase, you could backport the change I made here and re-compile - it's a fairly trivial fix.

maximd1:
@danpat 
Thank you for your update. Unfortunately I didn't find a textual_route_annotation.hpp in the 4.8.0 codebase - could you advise where shall I add your fix, please?

P.S. right now we can't upgrade to 4.9.0 yet, because we use data for 4.8.0 we've paid for. Next AWS processing will be accomplished within the 4.9.0 - but for the next several months we will continue using a previous build.

danpat:
@maximd1 The code got moved for 4.9.0.

Here's the equivalent code location in 4.8.1:

https://github.com/Project-OSRM/osrm-backend/blob/8f6fc0146ba76d34d20c5b7a87b75249bbb12b82/descriptors/json_descriptor.hpp#L384

This is where the last instruction is generated, there same `emplace_back` is missing there.

maximd1:
yes, this did the trick. Thank you very much!

danpat:
> P.S. right now we can't upgrade to 4.9.0 yet, because we use data for 4.8.0 we've paid for.

@maximd1 Consider using AWS spot pricing, it can drastically reduce the cost of running some of the large instances.  I wouldn't expect to spend more than $2 or $3 in EC2 costs to re-process the planet.

Issue #1504, opened 2015-05-30
Replace BOOST_ASSERT with own assertion


TheMarex:
Goal: We want to run assertions on large datasets to be sure we didn't miss any edge cases.
Problem:
 - To get `BOOST_ASSERT` in Release builds you need to undefine `NDEBUG`
 - We want syntax checks for assertion expressions in every build modus

How to:
- Enable by default in Debug build
- Disable by default in Release build
- Enable when `USE_ASSERTIONS=1` is passed as option
- `util/assertion.hpp` implemented as inline functions `osrm::assert(EXP)` and `osrm::assert_msg(EXP, MSG)`

daniel-j-h:
What about

    #define BOOST_ENABLE_ASSERT_HANDLER

because we do not want to change all the files in which we're using Boost.Assert, we could define it via CMake, that is passing `-DBOOST_ENABLE_ASSERT_HANDLER` to the compiler wrapped behind an `option(USE_ASSERTIONS OFF)`.

Then, in an implementation file, we define the handlers (they are already declared in Boost):

```c++
namespace detail {
  void assertion_failed_msg(char const* expr, char const* msg, char const* function, char const* file, long line) {
    std::ostringstream fmt;
    fmt << file << ":" << line << "\nin: " << function << ": " << expr << "\n" << msg;
    throw AssertionException{fmt.str().c_str()};
  }
}

// Boost.Assert only declares the following two functions and let's us define them.
// We throw to guarantee for stack-unwinding and therefore our destructors being called.
namespace boost {
void assertion_failed(char const* expr, char const* function, char const* file, long line) {
  ::detail::assertion_failed_msg(expr, "", function, file, line);
}

void assertion_failed_msg(char const* expr, char const* msg, char const* function, char const* file, long line) {
  ::detail::assertion_failed_msg(expr, msg, function, file, line);
}
}
```

This may seem a bit hacky (and probably also is), but allows us to 1/ change the behavior globally with almost no modifications 2/ properly do stack unwinding by throwing an exception.

References:

- http://www.boost.org/doc/libs/1_59_0/libs/assert/assert.html

TheMarex:
@daniel-j-h yes that is excatly what we need. Would like to ship this to the demo server, so we can get assertions in `osrm-extract` and `osrm-prepare` as well.

PR #1913, opened 2016-01-20
Provide a way to selectively enable assertions in release mode

daniel-j-h:
- Throwing an assertion exception for proper stack unwinding, making
  sure destructors are called

- On by default in Debug mode, in Release mode enable via:

        cmake -DCMAKE_BUILD_TYPE=Release -DENABLE_ASSERTIONS=ON

Current problem that I'm seeing is that some code is not catching
exceptions or worse silently swallowing them. Would like to check the
whole pipeline before merging this in.

Or `std::terminate` as plain old `assert` would do, without deterministic
stack unwinding and without our destructors to be called.

Resolves: https://github.com/Project-OSRM/osrm-backend/issues/1504

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -32,6 +32,7 @@ endif()
 option(ENABLE_JSON_LOGGING "Adds additional JSON debug logging to the response" OFF)
 option(DEBUG_GEOMETRY "Enables an option to dump GeoJSON of the final routing graph" OFF)
 option(BUILD_TOOLS "Build OSRM tools" OFF)
+option(ENABLE_ASSERTIONS OFF)
 
 include_directories(BEFORE ${CMAKE_CURRENT_BINARY_DIR}/include/)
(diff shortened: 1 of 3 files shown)


TheMarex:
Looks good merged!

daniel-j-h:
Noo, it's still using the `assertionError` exception derived from `std::logic_error`, so gets swallowed by both `catch(...)` and `catch(const std::exception&)`.

TheMarex:
@daniel-j-h yes but we only catch these at the top-level, no?

daniel-j-h:
Yes, we should. Check with

    ag 'catch(.*exception.*)' --ignore third_party/

Also we have make sure the logger is unmuted as soon as possible - I just tried adding a assertion at a tool's top for the fun of it. You won't get messages at all then.

TheMarex:
@daniel-j-h hm okay lets just fail hard in debug and only use exceptions in release mode

PR #1914, opened 2016-01-20
[experimental - do not merge] reverse name ids

MoKob:
Adds reverse name IDs to phantom nodes.

Allows for different names in forward/reverse direction.
This could be used to implement features like [name:left, name:right](http://wiki.openstreetmap.org/wiki/Key:name).


TheMarex:

PR #1917, opened 2016-01-21
Make XORFastHash great again

daniel-j-h:
- Initialize tables in clean way
- Use array with known bounds
- Assert in-bounds subscript operator
- Explicitely type integrals
- Remove unused XORMiniHash
- Cut down memory usage in half for two tables and storage positions

Warning: do not merge yet, as alternative.feature is failing! Running with -fsanitize=undefined revealed
> polyline_compressor.cpp:33 "left shift of negative value -106"

but I couldn't figure out yet why this happens.

When this is done, we should see a noticeable performance increase as it's currently used in the heap implementation. Benchmarking needed.

### DIFF
--- a/src/engine/polyline_compressor.cpp
+++ b/src/engine/polyline_compressor.cpp
@@ -1,6 +1,7 @@
 #include "engine/polyline_compressor.hpp"
 
 #include <cstddef>
+#include <cstdlib>
 #include <cmath>
 
 namespace osrm


include/util/xor_fast_hash.hpp
```diff
@@ -29,64 +37,30 @@ namespace util
 */
 class XORFastHash
 { // 65k entries
-    std::vector<unsigned short> table1;
-    std::vector<unsigned short> table2;
+    std::array<std::uint16_t, (1u << 16u)> table1;
```
TheMarex: The the size value should be template parameter.

daniel-j-h: Hmm the max size also depends on the `uint32_t` type in the `operator()` and how we split it into two `uint16_t` there. That is, the highest index a `uint16_t` can hold is `65535`, therefore we are only able to index `65536` items in each table.

 Therefore if we template the max size but keep the `uint32_t` for hashable types, we should at least

    static_assert(MaxSize <= (1u << 16u), "");

daniel-j-h:
@MoKob I adapted the alternative test to your reduced alternative test, but

    cucumber -p verify features/testbot/alternative.feature

is still failing.

In debug mode the undefined behavior described above still triggers in the polyline compressor. Debug mode on develop does not trigger the undefined behavior, so it seems like these changes potentially introduced it.

Any ideas @MoKob @TheMarex?

TheMarex:
@daniel-j-h this needs a deep dive into `contractor_graph.hpp`:

- get `.osm` data generated by test from `test/cache`
- run `osrm-extract`
- run `osrm-prepare` with version before/after your changes
- print out the order in which nodes are contracted

My gut tells me there is something going wrong with the heap in the contraction step, that leads to a different contraction order, that causes OSRM to not find an alternative path.

include/util/xor_fast_hash_storage.hpp
```diff
@@ -34,29 +34,35 @@ template <typename NodeID, typename Key> class XORFastHashStorage
...
+    explicit XORFastHashStorage(size_t) : positions(1u << 16u), current_timestamp(0) {}
 
     HashCell &operator[](const NodeID node)
     {
-        unsigned short position = fast_hasher(node);
+        std::uint16_t position = fast_hasher(node);
         while ((positions[position].time == current_timestamp) && (positions[position].id != node))
         {
-            ++position %= (2 << 16);
+            ++position %= (1u << 16u);
```
TheMarex: hard coded size again

include/util/xor_fast_hash_storage.hpp
```diff
@@ -34,29 +34,35 @@ template <typename NodeID, typename Key> class XORFastHashStorage
...
 
     // peek into table, get key for node, think of it as a read-only operator[]
     Key peek_index(const NodeID node) const
     {
-        unsigned short position = fast_hasher(node);
+        std::uint16_t position = fast_hasher(node);
         while ((positions[position].time == current_timestamp) && (positions[position].id != node))
         {
-            ++position %= (2 << 16);
+            ++position %= (1u << 16u);
```
TheMarex: This loop should at most wrap around once, if the array is full this could lead to endless loops.

include/util/xor_fast_hash_storage.hpp
```diff
@@ -66,13 +72,13 @@ template <typename NodeID, typename Key> class XORFastHashStorage
         if (std::numeric_limits<unsigned>::max() == current_timestamp)
         {
             positions.clear();
-            positions.resize(2 << 16);
+            positions.resize(1u << 16u);
         }
```
MoKob: Guess this is the final position missing the size?

daniel-j-h:
Can't get the alternative cucumber test to work; same with @MoKob's [reduced alternative test](https://gist.github.com/MoKob/b80acfc794169c1671f6).

Next steps would be to take a look at the order that's being generated I guess.

I _think_ it's because of the `random_shuffle` vs. the explicit `shuffle` with a mersenne twister.

MoKob:
The problem with this test is its strong dependence of the contraction order. The vertices in between cannot be contracted first. If source and target end up at the top of the hierarchy, an alternative can easily be missed. Unless you specify a contraction order or tailor the graph to perfectly reflect the priority calculation, there is no way of guaranteeing that case to pass.

TheMarex:
@daniel-j-h the alternative fix is fine. What we want to test in that case is the code path of alternative routes. We can't really check the route itself as @MoKob pointed out its more or less a heuristic.

@daniel-j-h you were working on a clean re-implementation of the poly line encoder/decoder?

daniel-j-h:
Re. polyline algorithm: yes, but it isn't done yet and I wanted to get the XORFastHash fix in with this pull request, so I quickly hotfixed the undefined behavior in the existing polyline compressor to not being blocked by it here.

TheMarex:
@daniel-j-h ok good to merge :+1:

PR #1928, opened 2016-01-25
make it possible to disable max_locations in map_matching

mortada:
Currently there's no way to disable the `max_locations_map_matching` check in the `match` plugin. Setting the `--max-matching-size=-1` in `osrm-routed`  will result in the following exception

```
[warn] [exception] Max location for map matching must be at least two`
```

this PR makes it possible to disable the max locations check by setting `--max-matching-size=-1`


TheMarex:
Good catch! :+1:

PR #1929, opened 2016-01-25
Send the Connection: close response header

zimathias:
so that proxy servers don't try to reuse the connection. 

Given that osrm-routed does *not* support keepalive the Connection: close response header should be set.


TheMarex:
@zimathias :+1: looks good.

daniel-j-h:
@zimathias just a word of caution: osrm-routed is more or less a development server, to make debugging osrm easy. It certainly has its problems, for example no keep alive support.

For deploying osrm we therefore recommend using libosrm with this lightweight Node wrapper: https://github.com/Project-OSRM/node-osrm

PR #1930, opened 2016-01-25
improve comments for gps_precision

mortada:
fixed some typos and changed >95% to >99% (3 sigma of normal distribution)

One thing to note is that the assumption about the relationship between GPS precision and the correct search radius may be a bit misleading, since the search radius should really also depend on the width of the road. Let's say a road has a large width (e.g. large number of lanes) and a trace is on a lane that's the farthest away from the road line. Then one would still want the search radius to reach the road even if the GPS precision is really high.

### DIFF
--- a/include/engine/plugins/match.hpp
+++ b/include/engine/plugins/match.hpp
@@ -85,8 +85,9 @@ template <class DataFacadeT> class MapMatchingPlugin : public BasePlugin
     {
         CandidateLists candidates_lists;
 
-        // assuming the gps_precision is the standart-diviation of normal distribution that models
-        // GPS noise (in this model) this should give us the correct candidate with >0.95
+        // assuming gps_precision is the standard deviation of a normal distribution that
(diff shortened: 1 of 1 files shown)


TheMarex:
> Then one would still want the search radius to reach the road even if the GPS precision is really high.

I guess what we really want is to incorporate road width in the nearest neighbor search in high precision cases like this. That would mean line segments need to be replaced by boxes. In general this would increase the index size quite a bit.

Do we have the data for this to actually test and see if this would be a worthwhile investment?

mortada:
@TheMarex I don't actually have any particular data in mind, and I'm not necessarily proposing a box search instead of a line search. But it just seems like the assumption that the GPS precision of a trace can determine the appropriate search radius for map matching is a bit misleading. 

Perhaps one possibility is to let users specify a desired search radius instead of a gps_precision

emiltin:
can't you still use a line search, but move the line according to the road width (for each direction)?

PR #1932, opened 2016-01-27
Takes care of proper special member generation globally, fixes #1689

### LINKED ISSUE #1689: Check special member generation for mistakes in implicitely generated ones
See for eample [server/http/header.hpp](https://github.com/Project-OSRM/osrm-backend/blob/5094bad838f11d5dbd08b443d552bf471f82a75e/server/http/header.hpp#L38-L41): writing special members is horrible here.

There are rules on how the compiler generates special member functions.
You have to understand them. You *really* have to understand them.

Check Howard Hinnant's talk slides in the references (he's the libc++ dev).
Read it, understand it, and then print out page 28. Seriously, just do it.

For the example above, the best solution is this:

```c++
struct header { std::string name, value; };
```

Tada, the compiler does the rest!
Isn't this awesome and so much more clear then trying to write them manually?

Note: a `pair<string, string>` or then a `map<string, string>` alias would work out just fine, too.

Reference:

- http://accu.org/content/conf2014/Howard_Hinnant_Accu_2014.pdf (p 28)

daniel-j-h:
Phew, a lot of classes were affected by this. The rationale for the
changes are as follows:

- When a type X declares any constructor, the default constructor is
  not declared, so there is no need for X() = delete there. In fact,
  there is brutal difference between those two: deleted members
  participate in overload resolution, but not-declared members do not!

- When a type X wants to be non-copyable (e.g. to be only movable, like
  threads, unique_ptrs, and so on), you can either do it by inheriting
  from boost::noncopyable (the old way), or better declare both (!) the
  copy constructor _and_ the copy assignment operator as deleted:

      X(X const&) = delete;
      X& operator=(X const&) = delete;

  We had tons of types with deleted copy constructors that were lacking
  a corresponding deleted copy assignment operator, making them still
  copyable and you wouldn't even notice (read: scary)!

References:

- http://accu.org/content/conf2014/Howard_Hinnant_Accu_2014.pdf
- http://www.boost.org/doc/libs/master/libs/core/doc/html/core/noncopyable.html

Note: I know, I'm quoting Hinnant's extraordinary slides a lot, but
getting the sematic right here is so incredibly important.

### DIFF
--- a/include/engine/engine.hpp
+++ b/include/engine/engine.hpp
@@ -46,7 +46,10 @@ class Engine final
 
   public:
     Engine(EngineConfig &config_);
+
     Engine(const Engine &) = delete;
+    Engine &operator=(const Engine &) = delete;
+
     int RunQuery(const RouteParameters &route_parameters, util::json::Object &json_result);
 
   private:
--- a/include/engine/routing_algorithms/routing_base.hpp
+++ b/include/engine/routing_algorithms/routing_base.hpp
@@ -42,10 +42,11 @@ template <class DataFacadeT, class Derived> class BasicRoutingInterface
     DataFacadeT *facade;
 
   public:
-    BasicRoutingInterface() = delete;
-    BasicRoutingInterface(const BasicRoutingInterface &) = delete;
     explicit BasicRoutingInterface(DataFacadeT *facade) : facade(facade) {}
     ~BasicRoutingInterface() {}
+
+    BasicRoutingInterface(const BasicRoutingInterface &) = delete;
+    BasicRoutingInterface &operator=(const BasicRoutingInterface &) = delete;
 
     /*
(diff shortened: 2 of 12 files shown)


daniel-j-h:
@TheMarex @MoKob @danpat could you make sure you're running clang-format on changes before pushing them? A few weeks ago we had a perfectly formatted code base and I'm already seeing unrelated changed due to clang-format here. Would be great to keep the signal to noise ratio high.

daniel-j-h:
Note: this PR does not remove explicitly specified special members where implicit ones would suffice. No need to change that right now, it's more important that we're fixing what is broken.

So hopefully now we're no longer accidentally copying objects around :)

TheMarex:
:+1: good to merge.

PR #1933, opened 2016-01-27
[WIP] introduce barrier penalties to handle gated areas

MoKob:
Many gated areas (like gated communities, factories ... ) exist in the graph. Usually they are separated from the rest of the graph by a barrier of some form.
In the current way of handling, we have no way of avoiding them.

This pull requests uses the turn_instruction `access_restriction_penalty`  to introduce a behaviour in which we can route **out/into** such areas but **not pass through** them.

This pull request has two problems:
 * It only works for barriers at the moment, but at least in Germany there exists a traffic sign requesting no through traffic (Anlieger frei -- essentially saying only enter if you have a good reason to go here)
 * It cannot handle the many-to-many plugin, as we do not extract paths in that scenario. To handle the many-to-many plugin, we would need https://github.com/Project-OSRM/osrm-backend/issues/77

### DIFF
--- a/profiles/car.lua
+++ b/profiles/car.lua
@@ -3,10 +3,10 @@
 local find_access_tag = require("lib/access").find_access_tag
 
 -- Begin of globals
-barrier_whitelist = { ["cattle_grid"] = true, ["border_control"] = true, ["checkpoint"] = true, ["toll_booth"] = true, ["sally_port"] = true, ["gate"] = true, ["lift_gate"] = true, ["no"] = true, ["entrance"] = true }
-access_tag_whitelist = { ["yes"] = true, ["motorcar"] = true, ["motor_vehicle"] = true, ["vehicle"] = true, ["permissive"] = true, ["designated"] = true, ["destination"] = true }
-access_tag_blacklist = { ["no"] = true, ["private"] = true, ["agricultural"] = true, ["forestry"] = true, ["emergency"] = true, ["psv"] = true }
-access_tag_restricted = { ["destination"] = true, ["delivery"] = true }
+barrier_whitelist = { ["border_control"] = true, ["checkpoint"] = true, ["toll_booth"] = true, ["sally_port"] = true, ["no"] = true, ["entrance"] = true }
+access_tag_whitelist = { ["yes"] = true, ["motorcar"] = true, ["motor_vehicle"] = true, ["vehicle"] = true, ["permissive"] = true }
+access_tag_blacklist = { ["no"] = true, ["agricultural"] = true, ["forestry"] = true, ["emergency"] = true, ["psv"] = true }
+access_tag_restricted = { ["destination"] = true, ["delivery"] = true, ["private"] = true }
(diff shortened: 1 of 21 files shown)


profiles/testbot.lua
```diff
@@ -52,6 +52,17 @@ function node_function (node, result)
...
+  if barrier and "" ~= barrier then
+  --  make an exception for rising bollard barriers
+     local bollard = node:get_value_by_key("bollard")
+     local rising_bollard = bollard and "rising" == bollard
+
+     if not rising_bollard then
+         result.barrier = true
+         result.access_restricted = false --allow access, but only at a penalty
+     end
+  end
```
MoKob: This is a temporary insertion. It allows the barrier.feature to pass but may be responsible for some cases breaking (routability is changed by this request)

profiles/car.lua
```diff
@@ -3,10 +3,10 @@
...
-barrier_whitelist = { ["cattle_grid"] = true, ["border_control"] = true, ["checkpoint"] = true, ["toll_booth"] = true, ["sally_port"] = true, ["gate"] = true, ["lift_gate"] = true, ["no"] = true, ["entrance"] = true }
-access_tag_whitelist = { ["yes"] = true, ["motorcar"] = true, ["motor_vehicle"] = true, ["vehicle"] = true, ["permissive"] = true, ["designated"] = true, ["destination"] = true }
-access_tag_blacklist = { ["no"] = true, ["private"] = true, ["agricultural"] = true, ["forestry"] = true, ["emergency"] = true, ["psv"] = true }
-access_tag_restricted = { ["destination"] = true, ["delivery"] = true }
+barrier_whitelist = { ["border_control"] = true, ["checkpoint"] = true, ["toll_booth"] = true, ["sally_port"] = true, ["no"] = true, ["entrance"] = true }
+access_tag_whitelist = { ["yes"] = true, ["motorcar"] = true, ["motor_vehicle"] = true, ["vehicle"] = true, ["permissive"] = true }
+access_tag_blacklist = { ["no"] = true, ["agricultural"] = true, ["forestry"] = true, ["emergency"] = true, ["psv"] = true }
+access_tag_restricted = { ["destination"] = true, ["delivery"] = true, ["private"] = true }
 access_tags = { "motorcar", "motor_vehicle", "vehicle" }
```
MoKob: manipulating these checks is partly responsible for breaking some of the routability checks in cucumber

src/extractor/edge_based_graph_factory.cpp
```diff
@@ -426,38 +428,27 @@ void EdgeBasedGraphFactory::GenerateEdgeExpandedEdges(
...
 
-                if (is_barrier_node)
-                {
-                    if (node_u != node_w)
-                    {
-                        ++skipped_barrier_turns_counter;
-                        continue;
-                    }
-                }
-                else
```
MoKob: Not omitting turns through barriers breaks parts of the routability checks in cucumber.

MoKob:
while somewhat successful, these penalties should be handled via the separation of eta and routing-weight. closing here.

PR #1934, opened 2016-01-28
Fixes Alternative Routing when including loops

MoKob:
The alternative route plugin fix for shortest path loops still had a bug.

This pull request fixes this bug and avoids additional errors that originate in the overlap computation when using the original middle node as via candidate.

### DIFF
--- a/include/engine/routing_algorithms/alternative_path.hpp
+++ b/include/engine/routing_algorithms/alternative_path.hpp
@@ -156,8 +156,10 @@ class AlternativeRouting final
         std::vector<NodeID> packed_forward_path;
         std::vector<NodeID> packed_reverse_path;
 
-        if (upper_bound_to_shortest_path_distance !=
-            forward_heap1.GetKey(middle_node) + reverse_heap1.GetKey(middle_node))
+        const bool path_is_a_loop =
+            upper_bound_to_shortest_path_distance !=
+            forward_heap1.GetKey(middle_node) + reverse_heap1.GetKey(middle_node);
+        if (path_is_a_loop)
         {
             // Self Loop
             BOOST_ASSERT(forward_heap1.GetData(middle_node).parent == middle_node &&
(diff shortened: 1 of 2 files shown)


MoKob:
An example of the broken routing prior to the patch: http://map.project-osrm.org/?z=17&center=37.352130%2C-122.055402&loc=37.351908%2C-122.058331&loc=37.352070%2C-122.058551&hl=en&ly=&alt=&df=&srv=

include/engine/routing_algorithms/alternative_path.hpp
```diff
@@ -156,8 +156,9 @@ class AlternativeRouting final
         std::vector<NodeID> packed_forward_path;
         std::vector<NodeID> packed_reverse_path;
 
-        if (upper_bound_to_shortest_path_distance !=
-            forward_heap1.GetKey(middle_node) + reverse_heap1.GetKey(middle_node))
+        bool path_is_a_loop = upper_bound_to_shortest_path_distance !=
```
TheMarex: Make const

include/engine/routing_algorithms/alternative_path.hpp
```diff
@@ -239,6 +240,8 @@ class AlternativeRouting final
         std::vector<NodeID> preselected_node_list;
         for (const NodeID node : via_node_candidate_list)
         {
+            if( node == middle_node )
```
TheMarex: Run clang-format with correct style file, this uses two spaces and no enclosing `{}`.

TheMarex:
Other than the added remarks this looks good to merge! Good catch @MoKob! Do you think we can add a regression test for this?

MoKob:
The problem with a regression test is that it near impossible to check for alternative routes in a guaranteed way. We can use a normal loop test with alternatives true to check for correct use of the loops. Whether an invalid alternative is found is not possible to check in my view.

PR #1935, opened 2016-01-28
[NOT READY] OSRM 5.0 API

TheMarex:
This PR is meant to track the implementation of the new OSRM 5.0 API.

TODO
- [x] Fix api_parameter parser for new query URL
- [x] Port plugins to new response format:
   - [x] viaroute
   - [x] trip
   - [x] table
   - [x] match
   - [x] nearest
- [ ] Unit Tests https://github.com/Project-OSRM/osrm-backend/pull/2088 @daniel-j-h
- [x] Adapt cucumber tests to new API:
  - [x] Fix the support code
  - [x] Remove code that tests for old API behavior

bug fixes
- [x] Return simplified overview geometry display (too granular at the moment, only returns origin/destination)
- [ ] Investigate dead-end snapping bearings bug https://github.com/Project-OSRM/osrm-backend/pull/1935#issuecomment-191311809
- [x] Refactor `uturns` parameter functionality https://github.com/mapbox/api-directions/issues/453
- [x] Fix `geometries` parameter handling - step geometries are only returned as polyline even with `geojson` geometry requests
- [x] Return true `NoMatch` response in match service https://github.com/Project-OSRM/osrm-backend/pull/1935#issuecomment-191449259
- [x] Rename no match response code in match service to `NoMatch` from `NoMatchings` (parity with public api)
- [x] Fix max integer values in distance table to return `null` rather than `2147483647`
- [x] Fix the polyline numeric problems
- [x] investigate mis-numbered roundabout exits (https://github.com/Project-OSRM/osrm-backend/pull/2007#issuecomment-192458899)
- [x] Rename `alternative` parameter to `alternatives`
 
For reference the specification that we want to implement: https://github.com/Project-OSRM/osrm-backend/wiki/New-Server-api

/cc @daniel-j-h @MoKob would be interested about your opinion of the code structure.


include/engine/guidance/step_maneuver.hpp
```diff
@@ -0,0 +1,26 @@
...
+namespace guidance
+{
+
+struct StepManeuver
+{
+    util::FixedPointCoordinate location;
+    double heading_before;
+    double heading_after;
+    extractor::TurnInstruction instruction;
+};
```
MoKob: Would love to have the opportunity to add something like a hint to every maneuver. Essentially something like **Turn right at name**, or **turn right onto ramp**

TheMarex: Oh this is captured by the `RouteStep` by the `way_name`. Sorry should document the relation better here. `StepManeuver` is the maneuver used to navigate on the given `RouteStep`

emiltin:
what about simplification of instructions? for example, crossing a big intersection on bike often results in multiple 'miniature' instructions because of the way the bike paths is modelled. but all you want to know as a user is a simple 'continue on ...'

include/engine/guidance/step_maneuver.hpp
MoKob: This only allows a single name, does it? I was thinking about something like `take exit **number** towards **direction**`. Is there a way do give these additional hints or am I supposed to encode these into some of the fields? I guess way_name would be the exit number (as a name for the ramp?), but where would I add the additional hints for the direction name?

Using the name could be a possibility, but it would somewhat break the expected result if such hints are encoded into the name field.

include/engine/douglas_peucker.hpp
```diff
@@ -44,14 +44,16 @@ const constexpr auto DOUGLAS_PEUCKER_THRESHOLDS_SIZE =
 // Input is vector of pairs. Each pair consists of the point information and a
 // bit indicating if the points is present in the generalization.
 // Note: points may also be pre-selected*/
-void douglasPeucker(std::vector<SegmentInformation>::iterator begin,
-                    std::vector<SegmentInformation>::iterator end,
-                    const unsigned zoom_level);
+std::vector<util::FixedPointCoordinate>
+douglasPeucker(std::vector<util::FixedPointCoordinate>::const_iterator begin,
+               std::vector<util::FixedPointCoordinate>::const_iterator end,
+               const unsigned zoom_level);
```
daniel-j-h: Why the change in the Douglas Peucker interface here? (Not saying it's bad, it looks better, just curious)

include/engine/engine.hpp
```diff
@@ -47,7 +48,7 @@ class Engine final
   public:
     Engine(EngineConfig &config_);
     Engine(const Engine &) = delete;
-    int RunQuery(const RouteParameters &route_parameters, util::json::Object &json_result);
+    Status RunQuery(const RouteParameters &route_parameters, util::json::Object &json_result);
```
daniel-j-h: tl;dr: returning an incomplete type is UB

3.9/5 http://eel.is/c++draft/basic.types#5

> A class that has been declared but not defined, an enumeration type in certain contexts ([dcl.enum]), or an array of unknown size or of incomplete element type, is an incompletely-defined object type.45 Incompletely-defined object types and cv void are incomplete types ([basic.fundamental]).

7.2/6 http://eel.is/c++draft/dcl.enum#6

> An enumeration whose underlying type is fixed is an incomplete type from its point of declaration ([basic.scope.pdecl]) to immediately after its enum-base (if any), at which point it becomes a complete type. An enumeration whose underlying type is not fixed is an incomplete type from its point of declaration to immediately after the closing } of its enum-specifier, at which point it becomes a complete type.

3.10/4 http://eel.is/c++draft/basic.lval#4

> Unless otherwise indicated ([expr.call]), prvalues shall always have complete types or the void type; in addition to these types, glvalues can also have incomplete types. [ Note: class and array prvalues can have cv-qualified types; other prvalues always have cv-unqualified types. See Clause [expr].  — end note ]

include/engine/guidance/geometry_assembler.hpp
```diff
@@ -0,0 +1,85 @@
...
+#include "engine/internal_route_result.hpp"
+#include "engine/phantom_node.hpp"
+#include "engine/guidance/route_step.hpp"
+#include "engine/guidance/leg_geometry.hpp"
+#include "util/coordinate_calculation.hpp"
+#include "util/coordinate.hpp"
+#include "extractor/turn_instructions.hpp"
+#include "extractor/travel_mode.hpp"
+
+#include <vector>
```
daniel-j-h: include utility for move

include/engine/douglas_peucker.hpp
TheMarex: To circumvent creating a vector for each route step, the whole leg geometry is saved in one large vector and each step only saves indices into that. Technically decided now that per-step simplification is probably not a good idea, but this is a left-over from the previous design.

include/engine/engine.hpp
TheMarex: Oh interesting. Thanks for the heads up. :+1:

include/engine/guidance/leg_assembler.hpp
```diff
@@ -0,0 +1,160 @@
...
+#define ENGINE_GUIDANCE_LEG_ASSEMBLER_HPP_
+
+#include "engine/guidance/route_leg.hpp"
+#include "engine/guidance/route_step.hpp"
+#include "engine/guidance/leg_geometry.hpp"
+#include "engine/internal_route_result.hpp"
+
+#include <boost/range/irange.hpp>
+
+#include <vector>
```
daniel-j-h: cstdint for size_t, uint32_t
numeric for accumulate
algorithm for transform, sort, min
string for, well, string :)
iterator for next

include/engine/guidance/leg_assembler.hpp
```diff
@@ -0,0 +1,160 @@
+#ifndef ENGINE_GUIDANCE_LEG_ASSEMBLER_HPP_
+#define ENGINE_GUIDANCE_LEG_ASSEMBLER_HPP_
+
+#include "engine/guidance/route_leg.hpp"
+#include "engine/guidance/route_step.hpp"
+#include "engine/guidance/leg_geometry.hpp"
+#include "engine/internal_route_result.hpp"
+
+#include <boost/range/irange.hpp>
```
daniel-j-h: unused, also older Boost's irange (e.g. 1.55) show disturbing behavior https://github.com/Project-OSRM/osrm-backend/pull/1642#issuecomment-137205439

include/engine/guidance/leg_assembler.hpp
```diff
@@ -0,0 +1,160 @@
...
+        // make sure the segments are sorted by position
+        const constexpr std::size_t MAX_USED_SEGMENTS = 2;
+        segments.resize(std::min(segments.size(), MAX_USED_SEGMENTS));
+        std::sort(segments.begin(), segments.end(),
+                  [](const NamedSegment &lhs, const NamedSegment &rhs)
+                  {
+                      return lhs.position < rhs.position;
+                  });
+
+        std::string summary;
```
daniel-j-h: `boost::algorithm::join` does the next 10 lines; take a look at rtd/include/util/geojson

    join(segments | transformed(toName), ", ");

include/engine/guidance/leg_geometry.hpp
```diff
@@ -0,0 +1,54 @@
+#ifndef ENGINE_GUIDANCE_LEG_GEOMETRY_HPP
+#define ENGINE_GUIDANCE_LEG_GEOMETRY_HPP
+
+#include "util/coordinate.hpp"
+#include "util/integer_range.hpp"
+
+#include <boost/assert.hpp>
+#include <boost/range/iterator_range.hpp>
```
daniel-j-h: unused, see comment above

include/engine/guidance/overview_assembler.hpp
```diff
@@ -0,0 +1,45 @@
+#ifndef ENGINE_GUIDANCE_OVERVIEW_ASSEMBLER_HPP
+#define ENGINE_GUIDANCE_OVERVIEW_ASSEMBLER_HPP
+
+#include "engine/guidance/leg_geometry.hpp"
+#include "engine/douglas_peucker.hpp"
+
+#include <boost/range/irange.hpp>
```
daniel-j-h: unused

include/engine/guidance/overview_assembler.hpp
```diff
@@ -0,0 +1,45 @@
...
+
+#include <vector>
+
+namespace osrm
+{
+namespace engine
+{
+namespace guidance
+{
+class OverviewAssembler
```
daniel-j-h: Why is this a class when the small helper types above are all structs?

include/engine/guidance/overview_assembler.hpp
```diff
@@ -0,0 +1,45 @@
...
+{
+  public:
+    std::vector<util::FixedPointCoordinate>
+    operator()(const std::vector<LegGeometry> &leg_geometries, const unsigned zoom_level) const
```
daniel-j-h: Also with a single operator() this could be a free standing function?

include/engine/guidance/route.hpp
```diff
@@ -0,0 +1,18 @@
...
+#define ROUTE_HPP
+
+namespace osrm {
+namespace engine {
+namespace guidance {
+
+struct Route
+{
+    double duration;
+    double distance;
```
daniel-j-h: Units?

include/engine/guidance/route_assembler.hpp
```diff
@@ -0,0 +1,42 @@
...
+namespace osrm
+{
+namespace engine
+{
+namespace guidance
+{
+class RouteAssembler
+{
+  public:
+    Route operator()(const std::vector<RouteLeg> &route_legs) const
```
daniel-j-h: Same as above, struct, operator()

include/engine/guidance/route_assembler.hpp
```diff
@@ -0,0 +1,42 @@
+#ifndef ENGINE_GUIDANCE_ROUTE_ASSEMBLER_HPP
+#define ENGINE_GUIDANCE_ROUTE_ASSEMBLER_HPP
+
+#include "engine/guidance/route_leg.hpp"
+#include "engine/guidance/route.hpp"
+
+#include <boost/range/irange.hpp>
+
+#include <vector>
```
daniel-j-h: numeric for accumulate

include/engine/guidance/route_leg.hpp
```diff
@@ -0,0 +1,28 @@
...
+{
+namespace guidance
+{
+
+struct RouteLeg
+{
+    double duration;
+    double distance;
+    std::string summary;
+    boost::optional<std::vector<RouteStep>> steps;
```
daniel-j-h: Does a optional<vector<T>> make sense here? A empty vector<T> already signals an abundance of values. Making use of the Free Monad and all that.

include/engine/guidance/route_step.hpp
```diff
@@ -0,0 +1,39 @@
+#ifndef ROUTE_STEP_HPP
+#define ROUTE_STEP_HPP
+
+#include "extractor/travel_mode.hpp"
+#include "engine/guidance/step_maneuver.hpp"
+
+#include <string>
+#include <vector>
```
daniel-j-h: cstdint for size_t

include/engine/guidance/route_step.hpp
```diff
@@ -0,0 +1,39 @@
...
+// this struct saves the information of the segment b,c.
+// Notable exceptions are Departure and Arrival steps.
+// Departue: s --> a --> b. Represents the segment s,a with location being s.
+// Arrive: a --> b --> t. The segment (b,t) is already covered by the previous segment.
+struct RouteStep
+{
+    unsigned name_id;
+    std::string way_name;
+    double duration;
+    double distance;
```
daniel-j-h: instead of duration and distance, why not use the Route type defined above that itself holds a duration and distance?

include/engine/guidance/step_assembler.hpp
```diff
@@ -0,0 +1,140 @@
...
+#include "engine/guidance/step_maneuver.hpp"
+#include "engine/guidance/leg_geometry.hpp"
+#include "engine/internal_route_result.hpp"
+#include "engine/phantom_node.hpp"
+#include "util/coordinate_calculation.hpp"
+#include "util/coordinate.hpp"
+#include "extractor/turn_instructions.hpp"
+#include "extractor/travel_mode.hpp"
+
+#include <boost/range/irange.hpp>
```
daniel-j-h: unused

include/engine/guidance/step_assembler.hpp
```diff
@@ -0,0 +1,140 @@
...
+
+#include <vector>
```
daniel-j-h: utility for move
cstddef for size_t

include/engine/guidance/step_assembler.hpp
```diff
@@ -0,0 +1,140 @@
...
+            // PathData saves the information we need of the segment _before_ the turn,
+            // but a RouteStep is with regard to the segment after the turn.
+            // We need to skip the first segment because it is already covered by the
+            for (const auto &path_point : leg_data)
+            {
+                if (path_point.turn_instruction != extractor::TurnInstruction::None)
+                {
+                    auto name = facade->get_name_for_id(path_point.name_id);
+                    const auto distance = leg_geometry.segment_distances[segment_index];
+                    steps.push_back(RouteStep{path_point.name_id, std::move(name),
```
daniel-j-h: I think you can emplace_back here

include/engine/guidance/step_assembler.hpp
```diff
@@ -0,0 +1,140 @@
...
+                                              path_point.travel_mode, maneuver,
+                                              leg_geometry.FrontIndex(segment_index),
+                                              leg_geometry.BackIndex(segment_index) + 1});
+                    maneuver = ManeuverFromGeometry(path_point.turn_instruction, leg_geometry,
+                                                    segment_index);
+                    segment_index++;
+                }
+            }
+            const auto distance = leg_geometry.segment_distances[segment_index];
+            steps.push_back(RouteStep{
```
daniel-j-h: same here

include/engine/guidance/step_assembler.hpp
```diff
@@ -0,0 +1,140 @@
...
+                leg_geometry.BackIndex(segment_index) + 1});
+        }
+        else
+        {
+            //
+            // |-----s source_duration
+            // |-------------t target_duration
+            // x---*---*---*---z compressed edge
+            //       |-------| duration
+            steps.push_back(RouteStep{
```
daniel-j-h: same here

include/engine/guidance/step_assembler.hpp
```diff
@@ -0,0 +1,140 @@
...
+                source_node.name_id, facade->get_name_for_id(source_node.name_id),
+                target_duration - source_duration, leg_geometry.segment_distances[segment_index],
+                source_mode,
+                StepManeuver{source_node.location, 0., 0., extractor::TurnInstruction::Depart},
+                leg_geometry.FrontIndex(segment_index), leg_geometry.BackIndex(segment_index) + 1});
+        }
+
+        BOOST_ASSERT(segment_index == number_of_segments - 1);
+        // This step has length zero, the only reason we need it is the target location
+        steps.push_back(RouteStep{
```
daniel-j-h: same here

include/engine/hint.hpp
```diff
@@ -0,0 +1,45 @@
...
+    }
+
+    std::string ToBase64() const
+    {
+        std::string encoded;
+        ObjectEncoder::EncodeToBase64(*this, encoded);
+        return encoded;
+    }
+
+    static Hint FromBase64(std::string base64Hint)
```
daniel-j-h: Take by const ref?

include/engine/internal_route_result.hpp
```diff
@@ -17,26 +17,15 @@ namespace engine
```
daniel-j-h: cstddef for size_t

include/engine/plugins/plugin_base.hpp
```diff
@@ -2,6 +2,7 @@
 #define BASE_PLUGIN_HPP
 
 #include "engine/phantom_node.hpp"
+#include "engine/status.hpp"
```
daniel-j-h: utility for pair
iterator for begin, end, back_inserter

include/engine/plugins/plugin_base.hpp
```diff
@@ -48,6 +41,13 @@ class BasePlugin
         return true;
     }
 
+    Status Error(const std::string& code, const std::string& message, util::json::Object &json_result) const
+    {
+        json_result.values["code"] = code;
+        json_result.values["message"] = message;
+        return Status::Error;
+    }
```
daniel-j-h: Take the strings by value and move them into the values

include/engine/plugins/viaroute.hpp
```diff
@@ -35,10 +39,133 @@ template <class DataFacadeT> class ViaRoutePlugin final : public BasePlugin
...
+        util::json::Array json_routes;
+        // currently the raw_route has a weird format that boundles routes
+        auto number_of_routes = raw_route.has_alternative() ? 2 : 1;
+        json_routes.values.reserve(number_of_routes);
+
+        std::vector<guidance::RouteLeg> legs;
+        std::vector<guidance::LegGeometry> leg_geometries;
+        legs.reserve(raw_route.segment_end_coordinates.size());
+        leg_geometries.reserve(raw_route.segment_end_coordinates.size());
+        for (auto idx : boost::irange(0UL, raw_route.segment_end_coordinates.size()))
```
daniel-j-h: See comment and ref. above why we do not use irange. Dangerous.

include/engine/polyline_compressor.hpp
```diff
@@ -11,13 +10,14 @@ namespace osrm
...
+using CoordVectorForwardIter = std::vector<FixedPointCoordinate>::const_iterator;
 // Encodes geometry into polyline format.
 // See: https://developers.google.com/maps/documentation/utilities/polylinealgorithm
-std::string polylineEncode(const std::vector<SegmentInformation> &geometry);
+std::string encodePolyline(CoordVectorForwardIter begin, CoordVectorForwardIter end);
 
 // Decodes geometry from polyline format
 // See: https://developers.google.com/maps/documentation/utilities/polylinealgorithm
-std::vector<util::FixedPointCoordinate> polylineDecode(const std::string &polyline);
+std::vector<util::FixedPointCoordinate> decodePolyline(const std::string &polyline);
```
daniel-j-h: Hm so global convention is `verbNoun`? I think that's my fault, I changed it when refactoring the code last week.

include/engine/response_objects.hpp
```diff
@@ -0,0 +1,257 @@
...
+        route_leg.values["steps"] = std::move(*steps);
+    }
+    return route_leg;
+}
+
+util::json::Array makeRouteLegs(std::vector<guidance::RouteLeg> &&legs,
+                                const std::vector<guidance::LegGeometry> &leg_geometries)
+{
+    util::json::Array json_legs;
+    for (const auto idx : boost::irange(0UL, legs.size()))
```
daniel-j-h: Careful here! Same as above. Check the whole code base for those please.

include/engine/guidance/step_assembler.hpp
TheMarex: But that would mean adding a custom constructor to `RouteStep` no?

include/engine/response_objects.hpp
```diff
@@ -0,0 +1,257 @@
...
+        token = "depart";
+        break;
+    case extractor::TurnInstruction::Arrive:
+        token = "arrive";
+        break;
+    default:
+        BOOST_ASSERT("Unsupported instruction reached");
+        break;
+    }
+    return token;
```
daniel-j-h: Hmm it's ugly to do the translation here. What about this (concept, untested):

    #define TURNS \
      X(Arrive, "arrive") \
      X(Depart, "depart")

and then

    #define X(sym, repr) sym

    enum class TurnInst {
        TURNS
    };

    #undef X

and for the representation

    #define X(sym, repr) case sym: return repr;
    switch (instruction) {
        TURNS
    }
    #undef X

This trick is called X-macro. It's a hack, you probably shouldn't use it too often. And I love it.

include/engine/response_objects.hpp
```diff
@@ -0,0 +1,257 @@
...
+#include "engine/polyline_compressor.hpp"
+#include "extractor/turn_instructions.hpp"
+#include "extractor/travel_mode.hpp"
+#include "util/coordinate.hpp"
+#include "util/json_container.hpp"
+#include "util/typedefs.hpp"
+
+#include <boost/assert.hpp>
+#include <boost/optional.hpp>
+#include <boost/range/irange.hpp>
```
daniel-j-h: Same as above

include/engine/response_objects.hpp
```diff
@@ -0,0 +1,257 @@
...
+    util::json::Object step_maneuver;
+    step_maneuver.values["type"] = detail::instructionToString(maneuver.instruction);
+    step_maneuver.values["location"] = detail::coordinateToLatLon(maneuver.location);
+    step_maneuver.values["heading_before"] = maneuver.heading_before;
+    step_maneuver.values["heading_after"] = maneuver.heading_after;
+    return step_maneuver;
+}
+
+util::json::Object makeRouteStep(guidance::RouteStep &&step,
+                                 boost::optional<util::json::Value> geometry)
```
daniel-j-h: Why rvalue-ref for step? Do we really need to take ownership?

include/engine/routing_algorithms/routing_base.hpp
```diff
@@ -293,8 +293,8 @@ template <class DataFacadeT, class Derived> class BasicRoutingInterface
                 if (!facade->EdgeIsCompressed(ed.id))
                 {
                     BOOST_ASSERT(!facade->EdgeIsCompressed(ed.id));
-                    unpacked_path.emplace_back(facade->GetGeometryIndexForEdgeID(ed.id), name_index,
-                                               turn_instruction, ed.distance, travel_mode);
+                    unpacked_path.push_back(PathData {facade->GetGeometryIndexForEdgeID(ed.id), name_index,
```
daniel-j-h: Just curious here, but why wasn't emplace_back an option here?

include/extractor/edge_based_node.hpp
```diff
@@ -75,10 +75,16 @@ struct EdgeBasedNode
...
+    // given the following geometry: x->y->u->v->w->z
+    // that is compressed to an edge x->z we need the distance
+    // x->u (forward_offset) and v<-z (reverse_offset) to initialite
+    // the heap with the correct offset to the end points.
+    // We keep the weight of the segment in forward and reverse direction around to
+    // support start points in the middle of a segment (PhantomNode).
+    int forward_weight; // u->v
+    int reverse_weight; // u<-v
+    int forward_offset; // distance x->y->u in the above example
+    int reverse_offset; // distance v<-w<-z in the above example
```
daniel-j-h: Yes, theses are the comments I've been longing for. I want more of this!

include/extractor/turn_instructions.hpp
```diff
@@ -8,35 +8,40 @@ namespace extractor
...
     StayOnRoundAbout,
-    StartAtEndOfStreet,
-    ReachedYourDestination,
-    EnterAgainstAllowedDirection,
-    LeaveAgainstAllowedDirection,
-    InverseAccessRestrictionFlag = 127,
-    AccessRestrictionFlag = 128,
-    AccessRestrictionPenalty = 129
+    Depart,
+    Arrive
```
daniel-j-h: X-macro

include/osrm/osrm.hpp
```diff
@@ -62,7 +60,7 @@ class OSRM
   public:
     OSRM(EngineConfig &lib_config);
     ~OSRM(); // needed because we need to define it with the implementation of OSRM_impl
-    int RunQuery(const RouteParameters &route_parameters, json::Object &json_result);
+    engine::Status RunQuery(const RouteParameters &route_parameters, json::Object &json_result);
```
daniel-j-h: Same here, incomplete type as return type is bad.

src/engine/douglas_peucker.cpp
```diff
@@ -1,8 +1,9 @@
 #include "engine/douglas_peucker.hpp"
 #include "util/coordinate_calculation.hpp"
+#include "util/coordinate.hpp"
 
 #include <boost/assert.hpp>
-#include "osrm/coordinate.hpp"
+#include <boost/range/irange.hpp>
```
daniel-j-h: Danger

src/engine/engine.cpp
```diff
@@ -1,14 +1,15 @@
...
-#include "engine/plugins/timestamp.hpp"
-#include "engine/plugins/trip.hpp"
+//#include "engine/plugins/distance_table.hpp"
+//#include "engine/plugins/hello_world.hpp"
+//#include "engine/plugins/nearest.hpp"
+//#include "engine/plugins/timestamp.hpp"
+//#include "engine/plugins/trip.hpp"
 #include "engine/plugins/viaroute.hpp"
-#include "engine/plugins/match.hpp"
+//#include "engine/plugins/match.hpp"
```
daniel-j-h: wip? same below

include/engine/guidance/step_assembler.hpp
daniel-j-h: Oh, yes you are correct. Hmm `push_back({args})` should work though.

include/engine/response_objects.hpp
TheMarex: I would rather just have a global static array that translates from consecutive Instructions to strings. I don't really like macro tricks like this.

include/engine/polyline_compressor.hpp
TheMarex: Yeah probably happened when you switched from the noun `PolylineDecoder` to `polylineDecode`. `decodePolyline` sounds better to my ear?

include/engine/guidance/route_step.hpp
TheMarex: Because it would be semantically incorrect. `Route` contains `RouteLegs` which contain `RouteSteps`

include/engine/guidance/route_leg.hpp
TheMarex: You are right, don't know why I wrapped this.

include/engine/guidance/overview_assembler.hpp
TheMarex: Yep, transforming these. Doesn't make much sense when you don't capture the Datafacade.

daniel-j-h:
- [ ] adapt `include/engine/plugins/{trip, match}.hpp`
- [ ] implement `src/engine/plugins/{nearest, trip, match}.cpp`
- [ ] implement `src/server/service/{nearest, trip, match}_service.cpp`
- [ ] implement `includeserver/api/{nearest, trip, match}_grammar.hpp`
- [ ] implement `include/engine/api `

daniel-j-h:
- [ ] Update Wiki (https://github.com/Project-OSRM/osrm-backend/wiki/Library-api)

TheMarex:
TODO:

- [ ] Switch internally from lat/lng to lng/lat
- [ ] Fix handling of `destinations` and `sources` in `many_to_many`:
   - empty arrays means `all`

TheMarex:
I just merged the switch from `lat, lng` to `lng, lat` in this.

:warning: **THIS BREAKS THE DATA FORMAT** :warning:

freenerd:
:tada:

karenzshea:
Found a bug with the `geometries` option - when passed `geojson` the overview geometry is returned as geojson but the step geometries are still polyline

cc @TheMarex

freenerd:
Looks like `overview: simplified` is a bit too simplistic:

screen shot 2016-02-25 at 16 27 05

TheMarex:
:warning: **PROFILES NEED TO BE UPDATED ** :warning: 

Fixing the travel mode required exiting all the profiles.

lbud:
@TheMarex: I'm hitting [this boost assertion error](https://travis-ci.org/Project-OSRM/node-osrm/jobs/111890322#L1560) in node-osrm's debug build. It looks like when a radius limits a search enough to not find a node, it trips this `target_phantom.IsValid()` assertion, when I'm assuming it should return something about being unable to find a route before it tries to assert this?

Reproducible with a `route` request, OSRM built in debug mode, 
coordinates: [13.43864,52.51993], [13.415852,52.513191],
radiuses: [10, 10],
alternative: false

```
# route: routes Berlin with valid radius values
Error: /home/travis/build/Project-OSRM/node-osrm/deps/osrm-backend-Debug/include/engine/routing_algorithms/direct_shortest_path.hpp:61 in: void osrm::engine::routing_algorithms::DirectShortestPathRouting<DataFacadeT>::operator()(const std::vector<osrm::engine::PhantomNodes>&, osrm::engine::InternalRouteResult&) const [with DataFacadeT = osrm::engine::datafacade::BaseDataFacade]: target_phantom.IsValid() 
  ---
    operator: error
    expected: |-
      undefined
    actual: |-
      [Error: /home/travis/build/Project-OSRM/node-osrm/deps/osrm-backend-Debug/include/engine/routing_algorithms/direct_shortest_path.hpp:61
in: void osrm::engine::routing_algorithms::DirectShortestPathRouting<DataFacadeT>::operator()(const std::vector<osrm::engine::PhantomNodes>&, osrm::engine::InternalRouteResult&) const [with DataFacadeT = osrm::engine::datafacade::BaseDataFacade]: target_phantom.IsValid()
]
    stack: |-
      Error: /home/travis/build/Project-OSRM/node-osrm/deps/osrm-backend-Debug/include/engine/routing_algorithms/direct_shortest_path.hpp:61
      in: void osrm::engine::routing_algorithms::DirectShortestPathRouting<DataFacadeT>::operator()(const std::vector<osrm::engine::PhantomNodes>&, osrm::engine::InternalRouteResult&) const [with DataFacadeT = osrm::engine::datafacade::BaseDataFacade]: target_phantom.IsValid()

```

TheMarex:
- [ ] This request has broken snapping `http://localhost:8080/directions/v5/mapbox/driving/-6.778477191925049,62.012351989746094;-6.77706241607666,62.01187777993707.json?geometries=geojson&bearings=350,10;`

karenzshea:
API has a test that makes a request with a body like
```
    {
      "type": "Feature",
      "properties": {},
      "geometry": {
        "type": "LineString",
        "coordinates": [ 
          [ -6.755433082580566, 62.0140175455414 ],
          [ -6.756570339202881, 62.0155882895177 ]
        ]
      }
    }
```
(supposed to be a messy trace) and a `gps_precision=1` parameter expects a `NoMatchings` code response but instead gets back
```
{ code: 'ok', matchings: [], tracepoints: [ null, null ] }
```

@TheMarex

src/extractor/guidance/turn_analysis.cpp
```diff
@@ -1659,87 +1663,84 @@ std::vector<TurnCandidate> TurnAnalysis::getTurnCandidates(const NodeID from_nod
...
 
     for (const EdgeID onto_edge : node_based_graph.GetAdjacentEdgeRange(turn_node))
     {
-        bool turn_is_valid = true;
+        // reverse edges are never valid turns because the resulting turn would look like this:
+        // from_node --via_edge--> turn_node <--onto_edge-- to_node
         if (node_based_graph.GetEdgeData(onto_edge).reversed)
         {
-            turn_is_valid = false;
+            continue;
```
MoKob: This breaks turn handling, since invalid turns are part of the turn situation but are not reported anymore

TheMarex:
:warning: Rebased on `develop` to pull in changes that change the indices.

** THIS BREAKS THE DATA FORMAT **

TheMarex:
:warning: Rebased on `develop`

@karenzshea @willwhite the recent merge of the uturns branch introduces a new file `.properties` that needs to be passed through to the server. Also the profiles need to be updated.

daniel-j-h:
Breaking out the not yet finished base64 changeset and merging it in here was a Bad Idea (tm).

I assume the Hint mechanism being broken after the merge. See https://github.com/Project-OSRM/osrm-backend/pull/2139#issuecomment-200752364

daniel-j-h:
Should be fixed in https://github.com/Project-OSRM/osrm-backend/pull/2088/commits/e7cee33101951a4b5e2e2106af0b505c2ef2dca5. Please cherry-pick if you want to keep the base64 changeset in this branch.

Note: the base64 tests still only compile on my `unit-tests-for-v5` branch as you only pulled the basic base64 changes over here, but not adaptions e.g. to the mock datafacade.

TheMarex:
:warning: rebased on `develop` to pull in the cucumber-js changes.

unit_tests/library/route.cpp
```diff
@@ -15,6 +16,52 @@
...
+                                                      {"name", ""},
+                                                      {"mode", "driving"},
+                                                      {"maneuver", json::Object{{
+                                                                       {"type", "depart"},
+                                                                       {"location", json::Array{}},
+                                                                       {"modifier", ""},
+                                                                       {"bearing_before", 0.},
+                                                                       {"bearing_after", 0.},
+                                                                       {"exit", 0},
+                                                                   }}}}}}}}}}}}}}}}}}}};
```
daniel-j-h: :100: Lisp :P

unit_tests/library/equal_json.hpp
```diff
@@ -0,0 +1,27 @@
+#ifndef UNIT_TESTS_JSON_EQUAL
+#define UNIT_TESTS_JSON_EQUAL
+
+#include <boost/test/included/unit_test.hpp>
+
+#include "util/json_deep_compare.hpp"
+
+boost::test_tools::predicate_result compareJSON(const osrm::util::json::Value &reference,
```
daniel-j-h: Needs inline otherwise Undefined Behavior by violating ODR

include/util/json_deep_compare.hpp
```diff
@@ -0,0 +1,158 @@
...
+#include <set>
+
+namespace osrm
+{
+namespace util
+{
+namespace json
+{
+
+struct Comparator : mapbox::util::static_visitor<bool>
```
daniel-j-h: Can't we just provide `operator==` for out types in json_container and then use mapbox/variant's `operator==`? https://github.com/mapbox/variant/blob/3f025adbf599d8dd9bfca02d45b37e49a2cae841/variant.hpp#L813-L829

include/util/json_deep_compare.hpp
```diff
@@ -0,0 +1,158 @@
...
+{
+    Comparator(std::string &reason_, const std::string &lhs_path_, const std::string &rhs_path_)
```
daniel-j-h: Ah I see we want the reason / message why it failed.

include/util/json_deep_compare.hpp
```diff
@@ -0,0 +1,158 @@
...
+      bool is_same = lhs.value == rhs.value;
+      if (!is_same)
+      {
+        reason = lhs_path + " (= \"" + lhs.value + "\") != " + rhs_path + " (= \"" + rhs.value + "\")";
+      }
+      return is_same;
+    }
+
+    bool operator()(const Number &lhs, const Number &rhs) const {
+      bool is_same = lhs.value == rhs.value;
```
daniel-j-h: This compares all numbers for `==`, what if we want to test routes that e.g. do not return `0` as `duration` / `distance`? That is, we need a way of saying: compare the JSON but don't complain on e.g. on the `distance` field.

Or what is your take on this?

include/engine/routing_algorithms/shortest_path.hpp
```diff
@@ -84,6 +84,12 @@ class ShortestPathRouting final
         BOOST_ASSERT(forward_heap.Size() > 0);
         BOOST_ASSERT(reverse_heap.Size() > 0);
 
+        // this is only relevent if source and target are on the same compressed edge
+        auto is_oneway_source = !(search_from_forward_node && search_from_reverse_node);
+        auto is_oneway_target = !(search_to_forward_node && search_to_reverse_node);
+        // we only enable loops here if we can't search from forward to backward node
+        auto needs_loop_forwad = is_oneway_source && super::NeedsLoopForward(source_phantom, target_phantom);
+        auto needs_loop_backwards = is_oneway_target && super::NeedsLoopBackwards(source_phantom, target_phantom);
```
daniel-j-h: const for all

lbud:
Tracking tests that are currently failing (noting fixes; will check off as those fixes merge into this branch):

- [ ] features/bicycle/cycleway.feature:43 # Scenario: Bike - Left/right side cycleways on implied oneways
- [ ] _features/bicycle/roundabout.feature:7 # Scenario: Bicycle - Roundabout instructions_ **[** fix in #2166 **]**
- [ ] _features/car/roundabout.feature:7 # Scenario: Car - Roundabout instructions_ **[** fix in #2166 **]**
- [ ] features/guidance/motorway.feature:8 # Scenario: Ramp Exit Right
- [ ] features/guidance/motorway.feature:23 # Scenario: Ramp Exit Right Curved Right
- [ ] features/guidance/motorway.feature:39 # Scenario: Ramp Exit Right Curved Left
- [ ] features/guidance/motorway.feature:57 # Scenario: Ramp Exit Left
- [ ] features/guidance/motorway.feature:72 # Scenario: Ramp Exit Left Curved Left
- [ ] features/guidance/motorway.feature:88 # Scenario: Ramp Exit Left Curved Right
- [ ] features/guidance/motorway.feature:104 # Scenario: On Ramp Right
- [ ] features/guidance/motorway.feature:119 # Scenario: On Ramp Left
- [ ] features/guidance/motorway.feature:134 # Scenario: Highway Fork
- [ ] features/guidance/motorway.feature:150 # Scenario: Fork After Ramp
- [ ] features/guidance/motorway.feature:167 # Scenario: On And Off Ramp Right
- [ ] features/guidance/motorway.feature:185 # Scenario: On And Off Ramp Left
- [ ] _features/guidance/roundabout.feature:8 # Scenario: Enter and Exit_ **[** fix in #2166 **]**
- [ ] _features/guidance/roundabout.feature:39 # Scenario: Only Enter_ **[** fix in #2166 **]**
- [ ] _features/guidance/roundabout.feature:74 # Scenario: Only Exit_ **[** fix in #2166 **]**
- [ ] _features/guidance/roundabout.feature:109 # Scenario: Drive Around_ **[** fix in #2166 **]**
- [ ] _features/guidance/roundabout.feature:140 # Scenario: Mixed Entry and Exit_ **[** fix in #2166 **]**
- [ ] _features/raster/extract.feature:7 # Scenario: osrm-extract on a valid profile_ **[** fix in #2177 **]**
- [ ] _features/raster/weights.feature:33 # Scenario: Weighting not based on raster sources_ **[** fix in #2177 **]**
- [ ] _features/raster/weights.feature:45 # Scenario: Weighting based on raster sources_ **[** fix in #2177 **]**
- [ ] _features/raster/weights.feature:63 # Scenario: Weighting based on raster sources_ **[** fix in #2177 **]**
- [ ] features/testbot/64bit.feature:5 # Scenario: 64bit overflow conflicts
- [ ] features/testbot/alternative_loop.feature:7 # Scenario: Alternative Loop Paths
- [ ] features/testbot/bad.feature:47 # Scenario: Routing close to the north/south pole
- [ ] features/testbot/bearing_param.feature:8 # Scenario: Testbot - Intial bearing in simple case
- [ ] features/testbot/bearing_param.feature:24 # Scenario: Testbot - Intial bearing in simple case
- [ ] features/testbot/bearing_param.feature:44 # Scenario: Testbot - Initial bearing on split way
- [ ] features/testbot/bearing_param.feature:73 # Scenario: Testbot - Initial bearing in all direction
- [ ] features/testbot/bearing.feature:7 # Scenario: Bearing when going northwest
- [ ] features/testbot/bearing.feature:20 # Scenario: Bearing when going west
- [ ] features/testbot/bearing.feature:102 # Scenario: Bearings on an east-west way.
- [ ] features/testbot/bearing.feature:143 # Scenario: Bearings at high latitudes
- [ ] features/testbot/bearing.feature:178 # Scenario: Bearings at high negative latitudes
- [ ] features/testbot/compression.feature:7 # Scenario: Compressed segments have correct order
- [ ] features/testbot/distance_matrix.feature:40 # Scenario: Testbot - Travel time matrix with fuzzy match
- [ ] features/testbot/distance.feature:48 # Scenario: 10m distances
- [ ] features/testbot/distance.feature:67 # Scenario: 100m distances
- [ ] features/testbot/duration.feature:7 # Scenario: Duration of ways
- [ ] features/testbot/geometry.feature:9 # Scenario: Route retrieving geometry
- [ ] features/testbot/graph.feature:23 # Scenario: Turn instructions on compressed road network geometry
- [ ] features/testbot/loop.feature:7 # Scenario: Weired sidestreet loops
- [ ] features/testbot/loop.feature:74 # Scenario: Avoid weird loops 3
- [ ] features/testbot/matching_turns.feature:7 # Scenario: Turn directions
- [ ] features/testbot/matching_turns.feature:78 # Scenario: Turn directions
- [ ] features/testbot/mode.feature:68 # Scenario: Testbot - Modes in each direction, same forward/backward speeds
- [ ] features/testbot/mode.feature:97 # Scenario: Testbot - Modes for oneway, same forward/backward speeds
- [ ] features/testbot/mode.feature:125 # Scenario: Testbot - Modes for reverse oneway, same forward/backward speeds
- [ ] features/testbot/opposite.feature:7 # Scenario: Testbot - Going against the flow
- [ ] features/testbot/origin.feature:24 # Scenario: North-south oneways close to the origin
- [ ] features/testbot/origin.feature:59 # Scenario: North-south oneways crossing the origin
- [ ] features/testbot/penalty.feature:133 # Scenario: Routing between signals on the same way should not incur a delay
- [ ] features/testbot/planetary.feature:4 # Scenario: Approximated Longitudinal distances at equator
- [ ] features/testbot/planetary.feature:18 # Scenario: Approximated Longitudinal distances at latitude 45
- [ ] features/testbot/planetary.feature:32 # Scenario: Approximated Longitudinal distances at latitude 80
- [ ] features/testbot/planetary.feature:46 # Scenario: Approximated Latitudinal distances at longitude 0
- [ ] features/testbot/planetary.feature:60 # Scenario: Approximated Latitudinal distances at longitude 45
- [ ] features/testbot/planetary.feature:74 # Scenario: Approximated Latitudinal distances at longitude 80
- [ ] features/testbot/projection.feature:24 # Scenario: Projection onto way at high latitudes, 1km distance
- [ ] features/testbot/projection.feature:34 # Scenario: Projection onto way at high latitudes, no distance
- [ ] _features/testbot/roundabout.feature:7 # Scenario: Testbot - Roundabout_ *fix in #2166*
- [ ] features/testbot/speed.feature:20 # Scenario: Testbot - Speed on rivers, map
- [ ] features/testbot/status.feature:38 # Scenario: Malformed requests
- [ ] features/testbot/time.feature:11 # Scenario: Basic travel time, 10m scale
- [ ] features/testbot/time.feature:40 # Scenario: Basic travel time, 100m scale
- [ ] features/testbot/time.feature:69 # Scenario: Basic travel time, 1km scale
- [ ] features/testbot/time.feature:197 # Scenario: Time of travel on part of a way
- [ ] features/testbot/turn_angles.feature:26 # Scenario: Slight Turn
- [ ] features/testbot/turns.feature:92 # Scenario: Turn instructions at high latitude
- [ ] features/testbot/uturn.feature:7 # Scenario: U-turns at via points disabled by default
- [ ] features/testbot/uturn.feature:26 # Scenario: Query param to allow U-turns at all via points
- [ ] features/testbot/uturn.feature:71 # Scenario: u-turn mixed with non-uturn vias
- [ ] features/testbot/via.feature:7 # Scenario: Simple via point
- [ ] features/testbot/via.feature:19 # Scenario: Simple via point with core factor
- [ ] features/testbot/via.feature:34 # Scenario: Via point at a dead end
- [ ] features/testbot/via.feature:70 # Scenario: Duplicate via point
- [ ] features/testbot/via.feature:85 # Scenario: Via points on ring of oneways
- [ ] features/testbot/via.feature:110 # Scenario: Via points on ring on the same oneway
- [ ] features/testbot/via.feature:134 # Scenario: Via point at a dead end with oneway
- [ ] features/testbot/via.feature:177 # Scenario: Via points on ring on the same oneway, forces one of the vertices to be top node
- [ ] features/testbot/via.feature:198 # Scenario: Multiple Via points on ring on the same oneway, forces one of the vertices to be top node
- [ ] features/timestamp/timestamp.feature:4 # Scenario: Request timestamp

MoKob:
@lbud can you exclude `features/guidance` from these tests? They are new tests that are previously non-working and are part of my efforts in https://github.com/Project-OSRM/osrm-backend/pull/2133

TheMarex:
All tests pass. Merged into deveop. :tada: :tada:

Issue #1567, opened 2015-07-10
Foot profile using subway and railway


kaligrafy:
The foot profile does route on subway lines and railways...

Somebody has any idea why?

TheMarex:
Ah I forgot to port the `use_public_transport` option over from the bike profile. https://github.com/Project-OSRM/osrm-backend/blob/master/profiles/bicycle.lua#L105

PR #1936, opened 2016-01-28
The Big Cleanup II

daniel-j-h:
wip


include/util/exception.hpp
```diff
@@ -1,5 +1,5 @@
-#ifndef OSRM_EXCEPTION_HPP
-#define OSRM_EXCEPTION_HPP
+#ifndef EXCEPTION_HPP
+#define EXCEPTION_HPP
```
MoKob: I deem these common header guard names dangerous if they do not at least include some directory/other identifier. Good chance of header guard conflicts.

daniel-j-h: Yes you are right. Will revert to the osrm prefix again. :+1:

TheMarex:
@daniel-j-h can we maybe keep the TODO list in this PR? I'm currently trying to split some smaller items out for @karenzshea to learn about C++ compiler errors :smile_cat:

daniel-j-h:
@TheMarex sure! @karenzshea hit me up if you need help in fighting compilers :hocho:

daniel-j-h:
- [x] RouteParameter should not depend on boost spirit
- [x] ~~RouteParameter should either be a struct or have getter-setter, not both~~ <- is done this way to act as a Spirit handler, setters are Spirit callbacks
- [ ] Drop the weak coupling of plugins over service name + registration
  -> libosrm interface should be _typed_ like node-osrm
- [ ] Rename external_memory_node.hpp
- [x] FixedPointNumber is not used anywhere
- [x] Remove `node_id.hpp`
- [x] Remove check-hsgr
- [ ] FIXME `xor_fast_has_shorage.hpp` only has space for 2^16 nodes
- [x] coordinate_calculation should be a namespace not struct
- [x] Remove reference to `crc32_processor.hpp` from `extractor.hpp`
- [x] Make douglas_peuker a function
- [x] Make PolylineCompressor a function
- [x] Make PolylineFormatter a function
- [x] Make RouteNameExtraction a function
- [ ] Remove SearchEngineData, fold into SearchEngine
- [ ] Drop gpx support
- [x] ~~Make TravelMode an enum class~~ not possible, would increase node sizes b/c enum classes are not bit-packable
- [x] Rename TurnInstrauctionsClass -> should be a function
- [ ] rename OSRM class -> Engine
- [x] ComputeAngle should be a free function and in coordinate_calculation
- [ ] unlint util/:
 - [ ] floating_point.hpp is it used?
 - [ ] interger_range.hpp is it used?
 - [ ] create json/ subfolder
- [x] mercator should be a namespace in coordinate_calculation
- [x] rename osrm_exception.hpp to exception.hpp
- [x] check if range_algorithms is used somewhere
- [x] remove tribool and fold into server code
- [ ] remove xml renderer
- [ ] remove typedefs.hpp -> No global NodeIDs and EdgeIDs !!!
- [ ] replace SimpleLogger with RTD_LOG
- [ ] Move debug_geometry to extractor
- [ ] routing_algorithms -> routing
- [ ] use sort_unique_resite or remove
- [ ] move matching_debug_info to engine/map_matching
- [ ] hash implementation in processing_chain should use util/std_hash
- [ ] replace DistTableWrapper with more generic implementation (2d vector wrapper or something)
- [ ] move TarjanSCC from extractor to util
- [ ] Remove DescriptorTable
- [ ] Rename FixedPointCoordinate -> Coordinate
- [ ] extraction_helper_functions does not follow style guide, should be cpp file
- [ ] lua_function_exists needs to be renamed
- [ ] NodeBasedDynamicGraphFromEdges needs to be renamed
- [ ] read_file_lower_content needs to be renamed
- [ ] abstract routed_options and fold into tools/routed.cpp
- [ ] rename geometry_string test to polyline
- [ ] move duration_parsing to extractor
- [ ] remove the config file option for osrm-extract and osrm-prepare
- [ ] change extractor class name to Extractor
- [x] HilbertCode should be function
- [x] polyline_compressor.cpp:33 "left shift of negative value -106", UB: found with -fsanitize=undefined

emiltin:
refactor lua profiles, with base class/functions containing code that's reused between profiles?

daniel-j-h:
@emiltin I would love if you could tackle that. The profiles desperately need some refactoring and small fixes like for this ticket: https://github.com/Project-OSRM/osrm-backend/issues/1567.

TheMarex:
@daniel-j-h can you please rebase + merge this as we shifted to API work?

daniel-j-h:
Okay. Ticketing remaining issues up in https://github.com/Project-OSRM/osrm-backend/issues/1971.

Waiting for Travis, then will merge.

daniel-j-h:
Hopefully this should fix Travis. Waiting for ok, then will merge.

PR #1937, opened 2016-01-29
Merge tribool into request_parser

karenzshea:
`tribool` is only being used by `request_parser.cpp`, and down the include chain in `connection.cpp`. This moves the definition into `request_parser.hpp` and renames it to `RequestStatus`.

Also made the `RequestStatus` values more descriptive.

cc @TheMarex


src/server/connection.cpp
```diff
@@ -41,13 +41,13 @@ void Connection::handle_read(const boost::system::error_code &error, std::size_t
 
     // no error detected, let's parse the request
     http::compression_type compression_type(http::no_compression);
-    util::tribool result;
+    RequestStatus result;
```
TheMarex: This needs to be `RequestParser::RequestStatus`. Remember this is a nested type and we are outside of `RequestParser`.

include/server/request_parser.hpp
```diff
@@ -22,11 +21,18 @@ class RequestParser
   public:
     RequestParser();
 
-    std::tuple<util::tribool, http::compression_type>
+    std::tuple<RequestStatus, http::compression_type>
     parse(http::request &current_request, char *begin, char *end);
 
+    enum class RequestStatus : char
```
TheMarex: Move this type declaration before the function declaration. The C++ compiler reads from top to bottom and doesn't know about stuff that is yet to come.

PR #1938, opened 2016-01-29
finding better turn angles and instructions

MoKob:
Improves the turn angle discovery for osrm-extract.

Considers not simply the first vertex of a way but looks a bit further into the way. Not necessarily perfect, but may improve turn emission.

Further improvements:

 - [x] consider all turns at the intersection
 - [x] make slight turn angle narrower

 - [x] need to fix: head on at name changes may break ([*head on* to *slight left*](http://map.project-osrm.org/?z=18&center=34.070494%2C-118.234885&loc=34.070498%2C-118.234676&loc=34.070342%2C-118.235110&hl=en&ly=&alt=&df=&srv=))

 - [x] need to fix: identical instructions:  [*go straight* has to become *slight right*](http://map.project-osrm.org/?z=18&center=34.034724%2C-118.336776&loc=34.034630%2C-118.339738&loc=34.034653%2C-118.337109&hl=en&ly=&alt=&df=&srv=)

- [x] need to fix: continue breaks: [*go straight* cannot become *turn right*](http://map.project-osrm.org/?z=18&center=32.788154%2C-116.832626&loc=32.788046%2C-116.832830&loc=32.788204%2C-116.832454&hl=en&ly=&alt=&df=&srv=)

- [x] continue on street only if its [most obvious](http://map.project-osrm.org/?z=18&center=34.076443%2C-118.215659&loc=34.077581%2C-118.215294&loc=34.076323%2C-118.216141&hl=en&ly=&alt=&df=&srv=)

- [x] continue needs to become [slight right](http://map.project-osrm.org/?z=18&center=34.062584%2C-117.811744&loc=34.061992%2C-117.812002&loc=34.063161%2C-117.811364&hl=en&ly=&alt=&df=&srv=)

- [x] Improve conflict handling to better consider invalid turns

- [x] consider road type in conflict resolution.
- [x] consider conflict resolution when deciding on turn go straight

- [x] [review 47](https://github.com/MoKob/turn-report/blob/master/case_47.png)
- [x] [review 58](https://github.com/MoKob/turn-report/blob/master/case_58.png)
- [x] [review 66](https://github.com/MoKob/turn-report/blob/master/case_66.png)
- [x] [review 85](https://github.com/MoKob/turn-report/blob/master/case_85.png)
- [x] [review 109](https://github.com/MoKob/turn-report/blob/master/case_109.png) - don-t emit no-turn for equal category on very similar turns
- [x] [review 118](https://github.com/MoKob/turn-report/blob/master/case_118.png) - sharp region might be to large
- [x] [review 143](https://github.com/MoKob/turn-report/blob/master/case_143.png) - find separated roads by identifying  same name streets
- [x] [review 200](https://github.com/MoKob/turn-report/blob/master/case_200.png) - tie breaking seems off here
- [x] [review 202](https://github.com/MoKob/turn-report/blob/master/case_202.png) [and 204](https://github.com/MoKob/turn-report/blob/master/case_204.png) - go straight should be slight-right
- [x] [review 237](https://github.com/MoKob/turn-report/blob/master/case_237.png) what to do about multiple valid u-turns?
- [x] [review 255](https://github.com/MoKob/turn-report/blob/master/case_255.png) go straight should be slight left
- [x] [review 294](https://github.com/MoKob/turn-report/blob/master/case_294.png) and [299](https://github.com/MoKob/turn-report/blob/master/case_299.png) should not emit turn
- [x] [review 319](https://github.com/MoKob/turn-report/blob/master/case_319.png) should not emit turn, review obviousness
- [x] [review 352](https://github.com/MoKob/turn-report/blob/master/case_352.png) make sure that no-turn is correct
- [x] [review 359](https://github.com/MoKob/turn-report/blob/master/case_359.png) should not be no-turn
- [x] [review 378](https://github.com/MoKob/turn-report/blob/master/case_378.png) should not force go straight
- [x] [review 391](https://github.com/MoKob/turn-report/blob/master/case_391.png) should be slight left
- [x] [review 392](https://github.com/MoKob/turn-report/blob/master/case_392.png) exclude name_id==0
- [x] [review 455](https://github.com/MoKob/turn-report/blob/master/case_455.png) should not be go straight
- [x] [review 493](https://github.com/MoKob/turn-report/blob/master/case_493.png)
- [x] [review 596](https://github.com/MoKob/turn-report/blob/master/case_594.png) 
- [x] [review 595](https://github.com/MoKob/turn-report/blob/master/case_595.png) should not be no-turn
- [x] [review 620](https://github.com/MoKob/turn-report/blob/master/case_595.png) should be slight-right
- [x] [review 707](https://github.com/MoKob/turn-report/blob/master/case_707.png) may want to keep the turn-right?

### DIFF
--- a/features/testbot/loop.feature
+++ b/features/testbot/loop.feature
@@ -92,6 +92,6 @@ Feature: Avoid weird loops caused by rounding errors
             | cf    | primary     |
 
         When I route I should get
-            | waypoints | route             | turns                                      |
-            | a,2,d     | ab,be,ef,ef,cf,cd | head,left,right,via,right,left,destination |
-            | a,1,d     | ab,be,ef,ef,cf,cd | head,left,right,via,right,left,destination |
+            | waypoints | route             | turns                                            |
+            | a,2,d     | ab,be,ef,ef,cf,cd | head,left,straight,via,straight,left,destination |
+            | a,1,d     | ab,be,ef,ef,cf,cd | head,left,straight,via,straight,left,destination |
--- a/features/testbot/matching_turns.feature
+++ b/features/testbot/matching_turns.feature
@@ -17,21 +17,13 @@ Feature: Turn directions/codes
         And the ways
             | nodes |
             | xa    |
-            | xb    |
             | xc    |
-            | xd    |
             | xe    |
-            | xf    |
             | xg    |
-            | xh    |
             | xi    |
-            | xj    |
             | xk    |
-            | xl    |
             | xm    |
-            | xn    |
             | xo    |
-            | xp    |
 
         When I match I should get
             | trace | route | turns                         | matchings |
@@ -40,7 +32,7 @@ Feature: Turn directions/codes
             | ia    | xi,xa | head,straight,destination     | ia        |
             | ic    | xi,xc | head,slight_right,destination | ic        |
             | ie    | xi,xe | head,right,destination        | ie        |
-            
+
             | ko    | xk,xo | head,left,destination         | ko        |
             | ka    | xk,xa | head,slight_left,destination  | ka        |
             | kc    | xk,xc | head,straight,destination     | kc        |
@@ -96,21 +88,13 @@ Feature: Turn directions/codes
         And the ways
             | nodes |
             | xa    |
-            | xb    |
             | xc    |
-            | xd    |
             | xe    |
-            | xf    |
             | xg    |
-            | xh    |
             | xi    |
-            | xj    |
             | xk    |
-            | xl    |
             | xm    |
-            | xn    |
             | xo    |
-            | xp    |
 
         When I match I should get
             | trace | route | turns                         | matchings | duration |
--- a/features/testbot/turn_angles.feature
+++ b/features/testbot/turn_angles.feature
@@ -1,0 +1,74 @@
+@routing @testbot @via
+Feature: Via points
+
+    Background:
+        Given the profile "testbot"
+
+        And a grid size of 4 meters
+
+    Scenario: Basic Right Turn
+        Given the node map
+            | a | b | c | d | e | f | g |
+            |   |   |   |   |   | h |   |
+            |   |   |   |   |   | i |   |
+            |   |   |   |   |   | j |   |
+            |   |   |   |   |   | k |   |
+
+        And the ways
+            | nodes   | oneway |
+            | abcdefg | yes    |
+            | ehijk   | yes    |
+
+        When I route I should get
+            | from | to | route                      | distance  | turns                  |
+            | a    | k  | abcdefg,ehijk              |  34m +-1  | head,right,destination |
+
+    Scenario: Slight Turn
+        Given the node map
+            | a | b | c | d | e | f | g |   |
+            |   |   |   |   |   | h | i |   |
+            |   |   |   |   |   |   |   | j |
+            |   |   |   |   |   |   |   | k |
+
+        And the ways
+            | nodes   | oneway |
+            | abcdefg | yes    |
+            | ehijk   | yes    |
+
+        When I route I should get
+            | from | to | route                      | distance  | turns                         |
+            | a    | k  | abcdefg,ehijk              |  34m +-1  | head,slight_right,destination |
+
+    Scenario: Nearly Slight Turn
+        Given the node map
+            | a | b | c | d | e | f | g |   |
+            |   |   |   |   |   | h |   |   |
+            |   |   |   |   |   |   | i |   |
+            |   |   |   |   |   |   |   | j |
+            |   |   |   |   |   |   |   | k |
+
+        And the ways
+            | nodes   | oneway |
+            | abcdefg | yes    |
+            | ehijk   | yes    |
+
+        When I route I should get
+            | from | to | route                      | distance  | turns                         |
+            | a    | k  | abcdefg,ehijk              |  37m +-1  | head,right,destination        |
+
+    Scenario: Nearly Slight Turn (Variation)
+        Given the node map
+            | a | b | c | d | e | f | g |   |
+            |   |   |   |   |   | h |   |   |
+            |   |   |   |   |   |   | i |   |
+            |   |   |   |   |   |   | j |   |
+            |   |   |   |   |   |   |   | k |
+
+        And the ways
+            | nodes   | oneway |
+            | abcdefg | yes    |
+            | ehijk   | yes    |
+
+        When I route I should get
+            | from | to | route                      | distance  | turns                         |
+            | a    | k  | abcdefg,ehijk              |  37m +-1  | head,right,destination        |
--- a/include/extractor/edge_based_graph_factory.hpp
+++ b/include/extractor/edge_based_graph_factory.hpp
@@ -16,6 +16,8 @@
 #include "extractor/restriction_map.hpp"
 
 #include <algorithm>
+#include <cstdint>
+#include <cstddef>
 #include <iosfwd>
 #include <memory>
 #include <queue>
@@ -23,6 +25,7 @@
 #include <unordered_map>
 #include <unordered_set>
 #include <vector>
+#include <string>
 
 #include <boost/filesystem/fstream.hpp>
 
@@ -70,10 +73,18 @@ class EdgeBasedGraphFactory
 
     unsigned GetHighestEdgeID();
 
-    TurnInstruction
-    AnalyzeTurn(const NodeID u, const NodeID v, const NodeID w, const double angle) const;
+    // Basic analysis of a turn (u --(e1)-- v --(e2)-- w)
+    // with known angle.
+    // Handles special cases like u-turns and roundabouts
+    // For basic turns, the turn based on the angle-classification is returned
+    TurnInstruction AnalyzeTurn(const NodeID u,
+                                const EdgeID e1,
+                                const NodeID v,
+                                const EdgeID e2,
+                                const NodeID w,
+                                const double angle) const;
 
-    int GetTurnPenalty(double angle, lua_State *lua_state) const;
+    std::int32_t GetTurnPenalty(double angle, lua_State *lua_state) const;
 
   private:
     using EdgeData = util::NodeBasedDynamicGraph::EdgeData;
@@ -123,8 +134,54 @@ class EdgeBasedGraphFactory
 
     void FlushVectorToStream(std::ofstream &edge_data_file,
                              std::vector<OriginalEdgeData> &original_edge_data_vector) const;
+
+    struct TurnCandidate
+    {
+        EdgeID eid; // the id of the arc
+        bool valid; // a turn may be relevant to good instructions, even if we cannot take the road
+        double angle;                // the approximated angle of the turn
+        TurnInstruction instruction; // a proposed instruction
+        double confidence;           // how close to the border is the turn?
+
+        std::string toString() const
+        {
+            std::string result = "[turn] ";
+            result += std::to_string(eid);
+            result += " valid: ";
+            result += std::to_string(valid);
+            result += " angle: ";
+            result += std::to_string(angle);
+            result += " instruction: ";
+            result += std::to_string(static_cast<std::int32_t>(instruction));
+            result += " confidence: ";
+            result += std::to_string(confidence);
+            return result;
+        }
+    };
+
+    // Use In Order to generate base turns
+
+    // cannot be const due to the counters...
+    std::vector<TurnCandidate> getTurnCandidates(NodeID from, EdgeID via_edge);
+    std::vector<TurnCandidate> optimizeCandidates(NodeID via_edge,
+                                                  std::vector<TurnCandidate> turn_candidates) const;
+    std::vector<TurnCandidate> suppressTurns(EdgeID via_edge,
+                                             std::vector<TurnCandidate> turn_candidates) const;
+
+    QueryNode getRepresentativeCoordinate(const NodeID src,
+                                          const NodeID tgt,
+                                          const EdgeID via_eid,
+                                          bool INVERTED) const;
+
+    bool isObviousChoice(EdgeID coming_from_eid,
+                         std::size_t turn_index,
+                         const std::vector<TurnCandidate> &turn_candidates) const;
+
+    std::size_t restricted_turns_counter;
+    std::size_t skipped_uturns_counter;
+    std::size_t skipped_barrier_turns_counter;
 };
-}
-}
+} // namespace extractor
+} // namespace osrm
 
 #endif /* EDGE_BASED_GRAPH_FACTORY_HPP_ */
--- a/include/extractor/turn_instructions.hpp
+++ b/include/extractor/turn_instructions.hpp
@@ -1,6 +1,11 @@
 #ifndef TURN_INSTRUCTIONS_HPP
 #define TURN_INSTRUCTIONS_HPP
 
+#include <algorithm>
+#include <cmath>
+
+#include <boost/assert.hpp>
+
 namespace osrm
 {
 namespace extractor
@@ -24,6 +29,7 @@ enum class TurnInstruction : unsigned char
     StayOnRoundAbout,
     StartAtEndOfStreet,
     ReachedYourDestination,
+    NameChanges,
     EnterAgainstAllowedDirection,
     LeaveAgainstAllowedDirection,
     InverseAccessRestrictionFlag = 127,
@@ -31,37 +37,87 @@ enum class TurnInstruction : unsigned char
     AccessRestrictionPenalty = 129
 };
 
+// shiftable turns to left and right
+const constexpr bool shiftable_left[] = {false, false, true, true, true, false, false, true, true};
+const constexpr bool shiftable_right[] = {false, false, true, true, false, false, true, true, true};
+
+inline TurnInstruction shiftTurnToLeft(TurnInstruction turn)
+{
+    BOOST_ASSERT_MSG(static_cast<int>(turn) < 9,
+                     "Shift turn only supports basic turn instructions");
+    if (turn > TurnInstruction::TurnSlightLeft)
+        return turn;
+    else
+        return shiftable_left[static_cast<int>(turn)]
+                   ? (static_cast<TurnInstruction>(static_cast<int>(turn) - 1))
+                   : turn;
+}
+
+inline TurnInstruction shiftTurnToRight(TurnInstruction turn)
+{
+    BOOST_ASSERT_MSG(static_cast<int>(turn) < 9,
+                     "Shift turn only supports basic turn instructions");
+    if (turn > TurnInstruction::TurnSlightLeft)
+        return turn;
+    else
+        return shiftable_right[static_cast<int>(turn)]
+                   ? (static_cast<TurnInstruction>(static_cast<int>(turn) + 1))
+                   : turn;
+}
+
+inline double angularDeviation(const double angle, const double from)
+{
+    const double deviation = std::abs(angle - from);
+    return std::min(360 - deviation, deviation);
+}
+
+inline double getAngularPenalty(const double angle, TurnInstruction instruction)
+{
+    BOOST_ASSERT_MSG(static_cast<int>(instruction) < 9,
+                     "Angular penalty only supports basic turn instructions");
+    const double center[] = {180, 180, 135, 90, 45,
+                             0,   315, 270, 225}; // centers of turns from getTurnDirection
+    return angularDeviation(center[static_cast<int>(instruction)], angle);
+}
+
+inline double getTurnConfidence(const double angle, TurnInstruction instruction)
+{
+
+    // special handling of U-Turns and Roundabout
+    if (instruction >= TurnInstruction::HeadOn || instruction == TurnInstruction::UTurn ||
+        instruction == TurnInstruction::NoTurn || instruction == TurnInstruction::EnterRoundAbout ||
+        instruction == TurnInstruction::StayOnRoundAbout || instruction == TurnInstruction::LeaveRoundAbout )
+        return 1.0;
+
+    BOOST_ASSERT_MSG(static_cast<int>(instruction) < 9,
+                     "Turn confidence only supports basic turn instructions");
+    const double deviations[] = {10, 10, 35, 50, 45, 0, 45, 50, 35};
+    const double difference = getAngularPenalty(angle, instruction);
+    const double max_deviation = deviations[static_cast<int>(instruction)];
+    return 1.0 - (difference / max_deviation) * (difference / max_deviation);
+}
+
 // Translates between angles and their human-friendly directional representation
 inline TurnInstruction getTurnDirection(const double angle)
 {
-    if (angle >= 23 && angle < 67)
-    {
+    // An angle of zero is a u-turn
+    // 180 goes perfectly straight
+    // 0-180 are right turns
+    // 180-360 are left turns
+    if (angle > 0 && angle < 60)
         return TurnInstruction::TurnSharpRight;
-    }
-    if (angle >= 67 && angle < 113)
-    {
+    if (angle >= 60 && angle < 140)
         return TurnInstruction::TurnRight;
-    }
-    if (angle >= 113 && angle < 158)
-    {
+    if (angle >= 140 && angle < 170)
         return TurnInstruction::TurnSlightRight;
-    }
-    if (angle >= 158 && angle < 202)
-    {
+    if (angle >= 170 && angle <= 190)
         return TurnInstruction::GoStraight;
-    }
-    if (angle >= 202 && angle < 248)
-    {
+    if (angle > 190 && angle <= 220)
         return TurnInstruction::TurnSlightLeft;
-    }
-    if (angle >= 248 && angle < 292)
-    {
+    if (angle > 220 && angle <= 300)
         return TurnInstruction::TurnLeft;
-    }
-    if (angle >= 292 && angle < 336)
-    {
+    if (angle > 300 && angle < 360)
         return TurnInstruction::TurnSharpLeft;
-    }
     return TurnInstruction::UTurn;
 }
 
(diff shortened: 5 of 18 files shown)


MoKob:
A few examples of changed behavior:

 * [*slight right* becomes *right*](http://map.project-osrm.org/?z=18&center=32.978784%2C-116.779652&loc=32.978653%2C-116.779448&loc=32.978937%2C-116.779824&hl=en&ly=&alt=&df=&srv=)
 * [*right* becomes *sharp right*](http://map.project-osrm.org/?z=18&center=32.978797%2C-116.779926&loc=32.978928%2C-116.779851&loc=32.978923%2C-116.780001&hl=en&ly=&alt=&df=&srv=)
* [*slight left* becomes *left*](http://map.project-osrm.org/?z=18&center=35.130035%2C-120.600647&loc=35.129864%2C-120.600690&loc=35.130171%2C-120.600834&hl=en&ly=&alt=&df=&srv=)
* [*sharp right* becomes *right*](http://map.project-osrm.org/?z=18&center=35.130088%2C-120.600534&loc=35.130092%2C-120.600330&loc=35.130171%2C-120.600834&hl=en&ly=&alt=&df=&srv=)

features/testbot/turn_angles.feature
```diff
@@ -0,0 +1,78 @@
+@routing @testbot @via
```
TheMarex: Yes, very good idea to add regression tests for this. :+1:

src/extractor/edge_based_graph_factory.cpp
```diff
@@ -394,6 +395,52 @@ void EdgeBasedGraphFactory::GenerateEdgeExpandedEdges(
     util::DEBUG_TURNS_START(debug_turns_path);
 #endif
 
+    const auto getRepresentativeCoordinate = [this](const NodeID src, const NodeID tgt,
```
TheMarex: What do `src` and `tgt` represent here?

```
u------v
       |
       w
```

Is `src = u` and `tgt = w` here?

TheMarex:
Great to see you tackling this! :+1:

src/extractor/edge_based_graph_factory.cpp
MoKob: Will clear this up in the process. Currently I am more in a Beta kind of stage. Have to take another approach to reasonably get correct results. But I will make sure to get reliable names in there.

include/extractor/turn_instructions.hpp
```diff
@@ -38,6 +38,40 @@ inline TurnInstruction getTurnDirection(const double angle)
...
+    }
+    if (angle >= 292 && angle < 336)
+    {
+        return TurnInstruction::TurnSharpLeft;
+    }
+    return TurnInstruction::UTurn;
+}
+
+// Translates between angles and their human-friendly directional representation
+inline TurnInstruction getOldTurnDirection(const double angle)
```
daniel-j-h: Is this for keeping backwards-compatibility while working on this PR? Can be removed later I assume?

include/extractor/turn_instructions.hpp
```diff
@@ -38,6 +38,40 @@ inline TurnInstruction getTurnDirection(const double angle)
...
+    if (angle >= 230 && angle < 292)
+    {
+        return TurnInstruction::TurnLeft;
+    }
+    if (angle >= 292 && angle < 336)
+    {
+        return TurnInstruction::TurnSharpLeft;
+    }
+    return TurnInstruction::UTurn;
+}
```
daniel-j-h: By the way, can't we just divide 360° by number of instructions and then divide the angle by this and use this as instruction id / index into the instructions?

include/extractor/turn_instructions.hpp
MoKob: Exactly. Currently I have this in there to output changes I do. I can then compare locations with unexpected turn changes (e.g. UTurn -> GoStraight) and check for reasons...

include/extractor/turn_instructions.hpp
MoKob: Not any more. I've modified the ranges to be not of the same size any more (e.g. go straight is narrower than left/right)

features/testbot/matching_turns.feature
```diff
@@ -17,21 +17,13 @@ Feature: Turn directions/codes
...
             | xg    |
-            | xh    |
             | xi    |
-            | xj    |
             | xk    |
-            | xl    |
             | xm    |
-            | xn    |
             | xo    |
-            | xp    |
```
MoKob: Technically, this makes the test a lot simpler, but its only designated to check for turns, not the matching.
We have to deal with a single node of degree 16, though. This makes it impossible to handle the instructions reasonably with only 8 instructions. Maintaining this, to handle more angles than these 8 seems unreasonable.

features/testbot/matching_turns.feature
```diff
@@ -96,21 +88,13 @@ Feature: Turn directions/codes
...
```
MoKob: see above.

include/extractor/turn_instructions.hpp
```diff
@@ -9,21 +13,22 @@ namespace extractor
...
+    TurnLeft,         // 7
+    TurnSlightLeft,   // 8
+    ReachViaLocation, // 9
+    HeadOn,           // 10
     EnterRoundAbout,
     LeaveRoundAbout,
     StayOnRoundAbout,
     StartAtEndOfStreet,
     ReachedYourDestination,
+    NameChanges,
```
MoKob: This NameChanges instruction has to be handled in front-ends, if we decide to merge this, or we would have to fake it/deactivate it.

MoKob:
@daniel-j-h @TheMarex I am just doing some checks to make sure cleaning up did not break everything. Could you have a look at the code?

features/testbot/loop.feature
```diff
@@ -93,5 +93,5 @@ Feature: Avoid weird loops caused by rounding errors
 
         When I route I should get
             | waypoints | route             | turns                                      |
-            | a,2,d     | ab,be,ef,ef,cf,cd | head,left,right,via,right,left,destination |
-            | a,1,d     | ab,be,ef,ef,cf,cd | head,left,right,via,right,left,destination |
+            | a,2,d     | ab,be,ef,ef,cf,cd | head,left,name_changes,via,right,left,destination |
+            | a,1,d     | ab,be,ef,ef,cf,cd | head,left,name_changes,via,right,left,destination |
```
daniel-j-h: @emiltin probably wants all those aligned :P check all feature changes

features/testbot/turn_angles.feature
```diff
@@ -0,0 +1,78 @@
...
+Feature: Via points
+
+    Background:
+        Given the profile "testbot"
+
+        And a grid size of 4 meters
+
+    Scenario: Via points on ring of oneways
+    # xa it to avoid only having a single ring, which cna trigger edge cases
```
daniel-j-h: What about those comments? (Also below) 

> xa it to avoid only having a single ring, which cna trigger edge cases

I don't understand.

include/extractor/edge_based_graph_factory.hpp
```diff
@@ -70,8 +72,12 @@ class EdgeBasedGraphFactory
...
     unsigned GetHighestEdgeID();
 
-    TurnInstruction
-    AnalyzeTurn(const NodeID u, const NodeID v, const NodeID w, const double angle) const;
+    TurnInstruction AnalyzeTurn(const NodeID u,
+                                const EdgeID e1,
+                                const NodeID v,
+                                const EdgeID e2,
+                                const NodeID w,
+                                const double angle) const;
```
daniel-j-h: This is

       e1     e2
    u ---- v ---- w

I assume? So this only takes the immediate neighbours into account?

include/extractor/edge_based_graph_factory.hpp
```diff
@@ -123,8 +129,41 @@ class EdgeBasedGraphFactory
...
+        double confidence;           // how close to the border is the turn?
+
+        std::string toString() const
+        {
+            std::ostringstream oss;
+            oss << "[turn] " << eid << " valid: " << valid << " angle: " << angle
+                << " Instruction: " << static_cast<int>(instruction)
+                << " Confidence: " << confidence;
+            return oss.str();
+        }
```
daniel-j-h: I wouldn't go the extra roundabout through streams, what about

    std::string fmt:
    fmt += "test"
    fmt += var:
    fmt += std::to_string(integral):
    return fmt:

include/extractor/edge_based_graph_factory.hpp
```diff
@@ -123,8 +129,41 @@ class EdgeBasedGraphFactory
...
+    };
+
+    // Use In Order to generate base turns
+    std::vector<TurnCandidate>
+    getTurnCandidates(NodeID from, EdgeID via_edge); // cannot be const due to the counters...
```
daniel-j-h: Is this clang-formatted? Maybe run it on all of the files you touched (`git diff --name-only`).

include/extractor/edge_based_graph_factory.hpp
```diff
@@ -123,8 +129,41 @@ class EdgeBasedGraphFactory
...
+    void optimizeCandidates(NodeID via_edge, std::vector<TurnCandidate> &turn_candidates) const;
+    void suppressTurns(EdgeID via_edge, std::vector<TurnCandidate> &turn_candidates) const;
```
daniel-j-h: Why out-refs? Why not take by-value, let the user move in when possible, return by-value, moving the result out.

include/extractor/turn_instructions.hpp
```diff
@@ -9,21 +13,22 @@ namespace extractor
...
+    GoStraight,       // 1
+    TurnSlightRight,  // 2
+    TurnRight,        // 3
+    TurnSharpRight,   // 4
+    UTurn,            // 5
+    TurnSharpLeft,    // 6
+    TurnLeft,         // 7
+    TurnSlightLeft,   // 8
+    ReachViaLocation, // 9
+    HeadOn,           // 10
```
daniel-j-h: What are the comments for here? If you want to be explicit, why not saying `GoStrainght = 1,` and so on?

Also what about the `127`, `129` below? Do their numeric values have specific meaning?

include/extractor/turn_instructions.hpp
```diff
@@ -31,37 +36,98 @@ enum class TurnInstruction : unsigned char
     AccessRestrictionPenalty = 129
 };
 
-// Translates between angles and their human-friendly directional representation
-inline TurnInstruction getTurnDirection(const double angle)
+inline TurnInstruction shiftTurnToLeft(TurnInstruction turn)
```
daniel-j-h: The next 100 lines or so look like a small state machine to me. Are there better ways of expressing those?

include/extractor/turn_instructions.hpp
```diff
@@ -31,37 +36,98 @@ enum class TurnInstruction : unsigned char
...
         return TurnInstruction::GoStraight;
+    default:
+        return turn;
     }
-    if (angle >= 202 && angle < 248)
-    {
+}
+
+inline double angularDeviation( const double angle, const double from ){
+  const double deviation = fabs( angle - from );
```
daniel-j-h: C has no function overloading, C++ has. There should be no need for `fabs` here, `std::abs` should suffice.

include/extractor/turn_instructions.hpp
```diff
@@ -31,37 +36,98 @@ enum class TurnInstruction : unsigned char
...
+                             0,   315, 270, 225}; // centers of turns from getTurnDirection
+    return angularDeviation( center[static_cast<int>(instruction)], angle);
+}
+
+inline double getTurnConfidence(const double angle, TurnInstruction instruction)
+{
+    // special handling of U-Turns and Roundabout
+    if (instruction >= TurnInstruction::HeadOn || instruction == TurnInstruction::UTurn ||
+        instruction == TurnInstruction::NoTurn)
+        return 1.0;
```
daniel-j-h: Oh? It surprises me that the `>=` on instances of enum classes works?! Never knew that. Do you really want to make this dependant on the instruction order in the enum? I can see things breaking when we add or remove instructions.

src/extractor/edge_based_graph_factory.cpp
```diff
@@ -12,15 +13,42 @@
...
+const double constexpr NARROW_TURN_ANGLE = 25.;
+// angle between roads that classifies them as fully distinct
+const double constexpr MINIMAL_DISTINCTION_ANGLE = 85.;
+// angle difference that can be classified as straight, if its the only narrow turn
+const double constexpr FUZZY_STRAIGHT_ANGLE = 15.;
+
+// Configuration to find representative candidate for turn angle calculations
+const double constexpr MINIMAL_SEGMENT_LENGTH = 1.;
+const double constexpr DESIRED_SEGMENT_LENGTH = 10.;
+} // namespace
```
daniel-j-h: You don't need the anonymous namespace for `const constexpr`, they are only visible in that TU anyway.

src/extractor/edge_based_graph_factory.cpp
```diff
@@ -363,6 +391,10 @@ void EdgeBasedGraphFactory::GenerateEdgeExpandedEdges(
 
     unsigned node_based_edge_counter = 0;
     unsigned original_edges_counter = 0;
+    restricted_turns_counter = 0;
+    skipped_uturns_counter = 0;
+    skipped_barrier_turns_counter = 0;
+    unsigned compressed = 0;
```
daniel-j-h: Is there the possibility of an overflow when using `unsigned` here?

src/extractor/edge_based_graph_factory.cpp
```diff
@@ -638,6 +599,398 @@ void EdgeBasedGraphFactory::GenerateEdgeExpandedEdges(
                                  << " turns over barriers";
 }
 
+// requires sorted candidates
```
daniel-j-h: Can you assert `std::is_sorted` for debug builds here?

src/extractor/edge_based_graph_factory.cpp
```diff
@@ -638,6 +599,398 @@ void EdgeBasedGraphFactory::GenerateEdgeExpandedEdges(
...
+    {
+        if (turn_candidates[getLeft(0)].instruction == TurnInstruction::UTurn)
+            turn_candidates[getLeft(0)].instruction = TurnInstruction::TurnSharpLeft;
+        if (turn_candidates[getRight(0)].instruction == TurnInstruction::UTurn)
+            turn_candidates[getRight(0)].instruction = TurnInstruction::TurnSharpRight;
+    }
+
+    const auto keepStraight = [](double angle)
+    {
+        return fabs(angle - 180) < 5;
```
daniel-j-h: `std::abs`

src/extractor/edge_based_graph_factory.cpp
```diff
@@ -638,6 +599,398 @@ void EdgeBasedGraphFactory::GenerateEdgeExpandedEdges(
...
+                            resolveTransitive(
+                                candidate_at_begin.instruction, instruction_right_of_begin,
+                                turn_candidates[getRight(getRight(conflict_begin))].instruction,
+                                RESOLVE_TO_RIGHT);
+                    }
+                }
+            }
+        }
+    }
+}
```
daniel-j-h: Wow that was some complex ugly function :/ maybe split it into smaller logical parts?

src/extractor/edge_based_graph_factory.cpp
```diff
@@ -638,6 +599,398 @@ void EdgeBasedGraphFactory::GenerateEdgeExpandedEdges(
...
+
+void EdgeBasedGraphFactory::suppressTurns(EdgeID via_edge,
+                                          std::vector<TurnCandidate> &turn_candidates) const
```
daniel-j-h: by-value / move-out

src/extractor/edge_based_graph_factory.cpp
```diff
@@ -638,6 +599,398 @@ void EdgeBasedGraphFactory::GenerateEdgeExpandedEdges(
...
+
+        auto confidence = getTurnConfidence(angle, turn);
+        if (!turn_is_valid)
+            confidence *= 0.8; // makes invalid turns more likely to be resolved in conflicts
+
+        turn_candidates.push_back({onto_edge, turn_is_valid, angle, turn, confidence});
+    }
+    using std::sort;
+    using std::begin;
+    using std::end;
```
daniel-j-h: I don't think you need any of those, the type you're sorting is a `std::vector`, so `begin/end(vec)` finds `std::begin/end` due to argument dependent lookup (ADL), then you can call `std::sort` on those iterators.

src/extractor/edge_based_graph_factory.cpp
```diff
@@ -638,6 +595,474 @@ void EdgeBasedGraphFactory::GenerateEdgeExpandedEdges(
...
+                                RESOLVE_TO_RIGHT);
+                    }
+                }
+            }
+        }
+    }
+    return turn_candidates;
+}
+
+bool EdgeBasedGraphFactory::isObviousChoice(EdgeID via_eid,
```
MoKob: @daniel-j-h did some updates. Could you review this function and suppressTurns again. After that we should be mostly done for now.

src/extractor/edge_based_graph_factory.cpp
```diff
@@ -700,19 +1125,89 @@ TurnInstruction EdgeBasedGraphFactory::AnalyzeTurn(const NodeID node_u,
...
         }
     }
-
-    return getTurnDirection(angle);
-}
-}
-}
+    // default: If the edge is very short, or we do not have a compressed geometry
+    return m_node_info_list[INVERTED ? src : tgt];
+};
```
daniel-j-h: Extra semicolon?

include/extractor/edge_based_graph_factory.hpp
```diff
@@ -123,8 +134,52 @@ class EdgeBasedGraphFactory
...
+                                                  std::vector<TurnCandidate> turn_candidates) const;
+    std::vector<TurnCandidate> suppressTurns(EdgeID via_edge,
+                                             std::vector<TurnCandidate> turn_candidates) const;
+
+    QueryNode getRepresentativeCoordinate(const NodeID src,
+                                          const NodeID tgt,
+                                          const EdgeID via_eid,
+                                          bool INVERTED) const;
+
+    bool isObviousChoice( EdgeID coming_from_eid, std::size_t turn_index, const std::vector<TurnCandidate> &turn_candidates ) const;
```
daniel-j-h: clang-format all the files you touched

include/extractor/turn_instructions.hpp
```diff
@@ -1,6 +1,10 @@
 #ifndef TURN_INSTRUCTIONS_HPP
 #define TURN_INSTRUCTIONS_HPP
 
+#include <algorithm>
+#include <cmath>
+#include <iostream>
```
daniel-j-h: remove

include/extractor/turn_instructions.hpp
```diff
@@ -31,37 +36,80 @@ enum class TurnInstruction : unsigned char
     AccessRestrictionPenalty = 129
 };
 
+namespace
+{
+ //shiftable turns to left and right
+const bool shiftable_left[] = {false, false, true, true, true, false, false, true, true};
+const bool shiftable_right[] = {false, false, true, true, false, false, true, true, true};
+} // namespace
```
daniel-j-h: anon. ns in header is not a good idea since every TU gets a copy of the vars: just make them const constexpr

include/extractor/turn_instructions.hpp
```diff
@@ -31,37 +36,80 @@ enum class TurnInstruction : unsigned char
...
+
+inline TurnInstruction shiftTurnToLeft(TurnInstruction turn)
+{
+    if (turn > TurnInstruction::TurnSlightLeft)
+        return turn;
+    else
+        return shiftable_left[static_cast<int>(turn)]
+                   ? (static_cast<TurnInstruction>(static_cast<int>(turn) - 1))
+                   : turn;
```
daniel-j-h: Hmm maybe assert in-bounds here

include/extractor/turn_instructions.hpp
```diff
@@ -31,37 +36,80 @@ enum class TurnInstruction : unsigned char
...
+}
+
+inline TurnInstruction shiftTurnToRight(TurnInstruction turn)
+{
+    if (turn > TurnInstruction::TurnSlightLeft)
+        return turn;
+    else
+        return shiftable_right[static_cast<int>(turn)]
+                   ? (static_cast<TurnInstruction>(static_cast<int>(turn) + 1))
+                   : turn;
```
daniel-j-h: Same

include/extractor/turn_instructions.hpp
```diff
@@ -31,37 +36,80 @@ enum class TurnInstruction : unsigned char
...
+{
+    const double deviation = std::abs(angle - from);
+    return std::min(360 - deviation, deviation);
+}
+
+inline double getAngularPenalty(const double angle, TurnInstruction instruction)
+{
+    const double center[] = {180, 180, 135, 90, 45,
+                             0,   315, 270, 225}; // centers of turns from getTurnDirection
+    return angularDeviation(center[static_cast<int>(instruction)], angle);
```
daniel-j-h: Same

include/extractor/turn_instructions.hpp
```diff
@@ -31,37 +36,80 @@ enum class TurnInstruction : unsigned char
...
+inline double getTurnConfidence(const double angle, TurnInstruction instruction)
+{
+    // special handling of U-Turns and Roundabout
+    if (instruction >= TurnInstruction::HeadOn || instruction == TurnInstruction::UTurn ||
+        instruction == TurnInstruction::NoTurn)
+        return 1.0;
+
+    const double deviations[] = {10, 10, 35, 50, 45, 0, 45, 50, 35};
+    const double difference = getAngularPenalty(angle, instruction);
+    const double max_deviation = deviations[static_cast<int>(instruction)];
```
daniel-j-h: Same

src/extractor/edge_based_graph_factory.cpp
```diff
@@ -638,6 +595,475 @@ void EdgeBasedGraphFactory::GenerateEdgeExpandedEdges(
...
+                                    {
+                                        return left.angle < right.angle;
+                                    }),
+                     "Turn Candidates not sorted by angle.");
+    const auto end = std::stable_partition(turn_candidates.begin(), turn_candidates.end(),
+                                           [](const TurnCandidate &candidate)
+                                           {
+                                               return candidate.valid;
+                                           });
+    turn_candidates.resize(std::distance(turn_candidates.begin(), end));
```
daniel-j-h: I don't understand the need for stable_partition+resize here. What about erase+remove_if?

include/extractor/turn_instructions.hpp
```diff
@@ -31,37 +37,87 @@ enum class TurnInstruction : unsigned char
     AccessRestrictionPenalty = 129
 };
 
+// shiftable turns to left and right
+const constexpr bool shiftable_left[] = {false, false, true, true, true, false, false, true, true};
+const constexpr bool shiftable_right[] = {false, false, true, true, false, false, true, true, true};
```
TheMarex: Add static assertion for the size of the array `w.r.t` to the number of turn instructions. You can extend the enum to include `NumberOfTurnInstructions` to get the size automatically.

MoKob: Its asserted for a different size than the number of actual instructions, since we do not shift all instructions

src/extractor/edge_based_graph_factory.cpp
```diff
@@ -405,71 +428,22 @@ void EdgeBasedGraphFactory::GenerateEdgeExpandedEdges(
             }
 
             ++node_based_edge_counter;
+            auto turn_candidates = getTurnCandidates(node_u, e1);
```
TheMarex: Rename `e1` to `incoming_edge` or something.

src/extractor/edge_based_graph_factory.cpp
```diff
@@ -638,6 +595,475 @@ void EdgeBasedGraphFactory::GenerateEdgeExpandedEdges(
...
+        if (turn_candidates[getRight(0)].instruction == TurnInstruction::UTurn)
+            turn_candidates[getRight(0)].instruction = TurnInstruction::TurnSharpRight;
+    }
+
+    const auto keepStraight = [](double angle)
+    {
+        return std::abs(angle - 180) < 5;
+    };
+
+    if (turn_candidates.size() > 1)
```
TheMarex: Change to early-exit style `if (turn_candidate <= 1) { return turn_candidates; } BOOST_ASSERT(turn_candidates.size() > 1);` to circumvent the indentation depth.

src/extractor/edge_based_graph_factory.cpp
```diff
@@ -638,6 +595,475 @@ void EdgeBasedGraphFactory::GenerateEdgeExpandedEdges(
...
+                auto &candidate_at_begin = turn_candidates[conflict_begin];
+                if (conflict_size == 2)
+                {
+                    if (turn.instruction == TurnInstruction::GoStraight)
+                    {
+                        if (instruction_left_of_end != TurnInstruction::TurnSlightLeft &&
+                            instruction_right_of_begin != TurnInstruction::TurnSlightRight)
+                        {
+                            std::int32_t resolved_count = 0;
+                            if (!keepStraight(candidate_at_end.angle) &&
```
TheMarex: Why the inverted logic here? `if (keepStraight() || resolve())` and switch branches.

src/extractor/edge_based_graph_factory.cpp
```diff
@@ -638,6 +595,475 @@ void EdgeBasedGraphFactory::GenerateEdgeExpandedEdges(
...
+                                !resolve(candidate_at_end.instruction, instruction_left_of_end,
+                                         RESOLVE_TO_LEFT))
+                                util::SimpleLogger().Write()
```
TheMarex: This should either be silent, an assertion or debug output.

src/extractor/edge_based_graph_factory.cpp
MoKob: Not the same due to the side effects. Resolve updates its instructions, but we call if only when we do not want to keep the instruction at GoStraight.

TheMarex:
:+1: good to merge if you rebase on `develop`.

PR #1945, opened 2016-02-01
Pedantic clang-format checking

daniel-j-h:
A few weeks after we touched literally every file in the code base and neatly formatted them adhering to our .clang-format style we again already have 36 files that violate the style.

This small pull request makes Travis angry when your files violate the style. :angry:

Discussion: I put in the check at the end, so that even when the style check fails at least we can see that it's only coming from those checks and the build still would succeed otherwise.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -166,5 +166,5 @@ script:
   - popd
   - |
     if [ -n "$RUN_CLANG_FORMAT" ]; then
-      ./scripts/format.sh
+      ./scripts/format.sh || true # we don't want to fail just yet
     fi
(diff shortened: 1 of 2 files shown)


daniel-j-h:
Oops, I added the script unconditionally at the bottom to every build. Let's wait for Travis (the GCC builds will fail since they do not have clang-format), then I will adapt it. I will also run a clang-format to get the builds to pass.

TheMarex:
Yup good idea. :+1: However there are a few checks now that actually are not dependent on the build matrix. I'm thinking about maybe moving these scirpt-only checks to a different CI to parallelization?

daniel-j-h:
With different CI you mean something like Circle CI? We could in addition at least also run the modernizer. Tidy and analyze is probably too much noise.

daniel-j-h:
Only enabled on latest Clang build (3.8 at the moment); last commit runs it once to get the build to pass.

TheMarex:
@daniel-j-h can we wait to merge here until the API refactor is merged? We should:
- [ ] Add the tool
- [ ] Run the tool, but output a warning

PR #1952, opened 2016-02-05
Shm vector wrapper cleanup and undefined behavior

daniel-j-h:
I started looking at the facades and especially how the shm datafacade wraps the memory region it owns. Turns out the shared memory wrapper we're using is more or less

```cpp
boost::const_multi_array_ref<T, 1> array_view{first, boost::extents[last - first]};
```
That is, a non-owning [array_view](http://www.boost.org/doc/libs/1_60_0/libs/multi_array/doc/reference.html#const_multi_array_ref
) wrapper around a memory range; iteratable, size, data, and so on.

This pull request contains the following fixes:
- refactors the shm vector wrapper's swap semantics
- makes swap noexcept (as it always should be!)
- fixes misuse of member function swaps
- fixes undefined behavior in the shm vector wrapper by shifting into the sign bit (reminder: integer literals are signed by default!)

And sorry for the noise, this comes from clang-format. We have to get the pedantic clang-format pull request merged already!

### DIFF
--- a/include/engine/datafacade/datafacade_base.hpp
+++ b/include/engine/datafacade/datafacade_base.hpp
@@ -14,8 +14,11 @@
 
 #include "osrm/coordinate.hpp"
 
+#include <cstddef>
+
+#include <vector>
+#include <utility>
 #include <string>
-#include <boost/optional.hpp>
 
 namespace osrm
 {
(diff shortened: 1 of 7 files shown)


TheMarex:
:+1: looks good and can be merged.

PR #1954, opened 2016-02-07
Fix 'egdes' typo, replace with 'edges'.

sebastic:
The lintian QA tool reported this spelling error for the recent rebuild of the OSRM 4.9.1 Debian package with GDAL 1.11.4 and 2.0.2.


TheMarex:
:+1:

PR #1964, opened 2016-02-10
Update libosmium to v2.6.0

springmeyer:
https://github.com/osmcode/libosmium/blob/master/CHANGELOG.md#260---2016-02-04


springmeyer:
This is passing on Travis, so this should be good to merge. But I'll let @TheMarex decide when to merge since I figure it would make sense to test this on larger data to ensure no speed regressions.

TheMarex:
:+1: Looks good to me.

PR #1965, opened 2016-02-11
Support passing OSRM_SHARED_LIBRARY_PATH env variable for cucumber tests

springmeyer:
The need for this is that:

  - On OS X if libraries are stored in custom locations (not /usr/lib or /usr/local)
    and they do not embed an rpath (which is common to not have) then the developer
    needs to set DYLD_LIBRARY_PATH to ensure that osrm tools can find dependent libraries
    at runtime (this is normal and common).
  - But as of OS X > 10.11 DYLD_LIBRARY_PATH no longer is inherited. While it works in
    the main shell when running a command like `osrm-extract` you will find that when
    `osrm-extract` is run by cucumber (ruby child process) then DYLD_LIBRARY_PATH
    is blocked and the command cannot start.
  - So, this introduces the ability to pass in a variable that the cucumber tests will
    understand and can manually forward along to ensure that DYLD_LIBRARY_PATH is respected
    where it counts.

The intended usage of this is therefore:

    # set the environment variable
    export OSRM_SHARED_LIBRARY_PATH=${DYLD_LIBRARY_PATH}
    # then run cucumber tests
    cucumber -p verify

### DIFF
--- a/features/support/data.rb
+++ b/features/support/data.rb
@@ -258,7 +258,7 @@ def extract_data
   Dir.chdir TEST_FOLDER do
     log_preprocess_info
     log "== Extracting #{osm_file}.osm...", :preprocess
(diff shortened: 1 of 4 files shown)


TheMarex:
:+1: looks good to me.

Issue #1956, opened 2016-02-09
OSRM debugging GeoJSON should be fully formed


danpat:
Currently, the debugging output from `osrm-prepare` requires a second pass in order to replace node id values with real coordinates.

The debugging code should be modified to load the `.nodes` file so that nodes can be looked up as the add the output is written.

/cc @karenzshea

PR #1966, opened 2016-02-11
Include real coordinates in debug geojson mode

karenzshea:
https://github.com/Project-OSRM/osrm-backend/issues/1956

For building with the `DEBUG_GEOMETRY=ON`, previously the output geojson only has OSM node ids in the coordinates field, this uses the osrm.nodes file to lookup real lon/lat coordinates and write them.

@danpat @TheMarex

### DIFF
--- a/include/util/debug_geometry.hpp
+++ b/include/util/debug_geometry.hpp
@@ -4,6 +4,7 @@
 #include "contractor/contractor_config.hpp"
 #include "extractor/query_node.hpp"
 #include "osrm/coordinate.hpp"
+#include <boost/filesystem/fstream.hpp>
 
 #ifndef DEBUG_GEOMETRY
 
@@ -69,6 +70,8 @@ boost::filesystem::ofstream dg_debug_turns_file;
 bool dg_output_turn_debug = false;
 bool dg_first_turn_debug = true;
 
+std::unordered_map<OSMNodeID, util::FixedPointCoordinate> node_lookup_map;
+
 inline void DEBUG_GEOMETRY_START(const contractor::ContractorConfig &config)
 {
     time_t raw_time;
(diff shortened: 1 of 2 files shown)


danpat:
@karenzshea Can you squash these commits and update the commit message to be the same as the message for this PR?

include/util/debug_geometry.hpp
```diff
@@ -77,12 +81,25 @@ inline void DEBUG_GEOMETRY_START(const contractor::ContractorConfig &config)
...
+
+    extractor::QueryNode current_node;
+    unsigned number_of_coordinates = 0;
+    nodes_input_stream.read((char *)&number_of_coordinates, sizeof(unsigned));
+
+    for (unsigned i = 0; i < number_of_coordinates; ++i)
+    {
+        nodes_input_stream.read((char *)&current_node, sizeof(extractor::QueryNode));
+        node_lookup_map[current_node.node_id] = util::FixedPointCoordinate(current_node.lat, current_node.lon);
+    } nodes_input_stream.close();
```
daniel-j-h: Run `clang-format` (similat to gofmt for example) that re-formats against our `.clang-format` specs.

include/util/debug_geometry.hpp
```diff
@@ -4,6 +4,8 @@
 #include "contractor/contractor_config.hpp"
 #include "extractor/query_node.hpp"
 #include "osrm/coordinate.hpp"
+#include <ios>
```
TheMarex: What is this library?

include/util/debug_geometry.hpp
```diff
@@ -69,6 +71,8 @@ boost::filesystem::ofstream dg_debug_turns_file;
 bool dg_output_turn_debug = false;
 bool dg_first_turn_debug = true;
 
+std::unordered_map<OSMNodeID, util::FixedPointCoordinate> node_lookup_map;
```
TheMarex: needs to be static? /cc @daniel-j-h

danpat: Shouldn't need `static` in this case.  AFAIK, `static` on global variables doesn't turn them into singletons, that only applies for function-scoped and class-member static stuff.

MoKob: that screams Multiple Defined Symbol all over, doesn't it?

daniel-j-h: @MoKob you're still right in that is undefined behavior, violating the ODR (one definition rule). If multiple TUs include that header and then get linked together :boom:

Needs an `extern` declaration in the header and a single definition in exactly one implementation file.

daniel-j-h: Hm actually the same applies to the two bools above. For those we could simply make them `const constexpr auto` in the header and call it a day.

danpat: Can't make them `const`, they need to be modified.  These are just their initial values.

MoKob:
why did we just merge more undefined behaviour into the codebase?

TheMarex:
This is code that is only used for debug purposes. It's not even compiled in production.

MikeLeonard:
Hi

I've just posted an issue related to this (issue #2024) and am hoping that flagging it here is ok. This thread seems to contain the people who have touched the debug geoJSON mode code so I thought flagging it here would make sense (but I've written a separate comment so as not to clog up this pull request).

@TheMarex , @danpat, @daniel-j-h  in particular... I'm getting a blank geoJSON file out when running osrm-prepare with the option --debug-geometry. The option is there and --debug-turns is working fine for me with osrm-extract. I'd would really appreciate any help if you have time to check out issue #2024.  Thanks!

Mike

PR #1975, opened 2016-02-12
C++14. Let's do this.

daniel-j-h:
Reference: https://github.com/mapbox/windows-builds/issues/78

@BergWerkGIS there's no flag for Visual Studio, is there? The default will be good enough?

Let's see how both Appveyor and Travis like this.


BergWerkGIS:
`VS 2015 Update 2` is not yet available as RTM and will still take some time (cannot say how long) for all C++14 features to arrive.

Also, I haven't yet looked into this error:

```
C:\projects\osrm\src\engine\engine.cpp(90): error C2039: 'shared_lock': is not a member of 'boost'
```

https://ci.appveyor.com/project/DennisOSRM/osrm-backend/build/1.0.2564#L1705

Has anything related to `boost` changed in `osrm-backend` lately?

daniel-j-h:
Yes https://github.com/Project-OSRM/osrm-backend/commit/3075f6cea71c18de93d34f8c4d8f943abdf2b7ce

daniel-j-h:
Disabled the OSX shared lib build. In addition to not finding libosrm, it had a single cucumber test failing: https://travis-ci.org/Project-OSRM/osrm-backend/jobs/108934582#L481

I could not reproduce it locally via

    cucumber -p verify features/options/routed/invalid.feature

and _all_ the other builds are fine, too.

daniel-j-h:
Travis gives okay!

Disclaimer: requires C++14 compiler and stdlib. That is, gcc<=4.8 is out. Also clang<=3.4 is out since it would require `-std=c++1y` and is probably not mature enough.

Let's require gcc>=4.9, clang>=3.5 and re-write the docs to simply install the gcc 5.3 ppa for older Ubuntus. gcc 4.9 is in Debian Jessie, so we're fine there.

BergWerkGIS:
I got `develop` and this branch to compile locally with this change:


```diff
diff --git a/src/engine/engine.cpp b/src/engine/engine.cpp
index c9fe44e..2949de1 100644
--- a/src/engine/engine.cpp
+++ b/src/engine/engine.cpp
@@ -22,6 +22,7 @@
 #include <boost/assert.hpp>
 #include <boost/interprocess/sync/named_condition.hpp>
 #include <boost/interprocess/sync/scoped_lock.hpp>
+#include <boost/thread/lock_types.hpp>

 #include <algorithm>
 #include <fstream>

```

Also had to patch boost as it disables some c++11 and c++14 features for Visual Studio by default:

e.g.:
```
BOOST_NO_CXX11_CONSTEXPR
```

Uploaded an updated binary deps package.

----

Tests are failing as it seems that `algorithm-test.exe` and `datastructure-tests.exe` don't get generated anymore.

springmeyer:
Lock types fix landed in https://github.com/Project-OSRM/osrm-backend/pull/1994

TheMarex:
Urg looks like the auto merging f-ed up in this case after rebasing. We run against gcc 5.0 and gcc 4.9 now, right?

daniel-j-h:
Yes, packages are gcc-5 (which really is gcc 5.3) and gcc-4.9.

TheMarex:
Blocked by https://github.com/Project-OSRM/node-osrm/pull/153/files

daniel-j-h:
Rebased onto develop.

TheMarex:

PR #1976, opened 2016-02-13
correct handling of travel modes

MoKob:
At the moment, travel modes are handled in a very unintuitive way.
The travel mode of a segment cannot be deduced from its own ID but only from the turn-id it has been reached from (and the phantom node itself).

This comes with additional implied behaviour for the turn-by-turn generation.

To allow fine grained weights (like in https://github.com/Project-OSRM/osrm-backend/pull/1970) along edges, we need the travel mode to not run into problems with this implied behaviour.

In addition, it allows us to do better instructions based on travel modes (e.g. in the tunnel, take the fork on the right) or similar stuff.

### DIFF
--- a/include/engine/guidance/segment_list.hpp
+++ b/include/engine/guidance/segment_list.hpp
@@ -154,7 +154,7 @@ void SegmentList<DataFacadeT>::AddLeg(const std::vector<PathData> &leg_data,
                                       const bool is_via_leg,
                                       const DataFacade *facade)
 {
-    for (const PathData &path_data : leg_data)
+    for (const auto &path_data : leg_data)
     {
         AppendSegment(facade->GetCoordinateOfNode(path_data.node), path_data);
     }
@@ -196,32 +196,7 @@ template <typename DataFacadeT>
 void SegmentList<DataFacadeT>::AppendSegment(const FixedPointCoordinate coordinate,
                                              const PathData &path_point)
 {
-    // if the start location is on top of a node, the first movement might be zero-length,
-    // in which case we dont' add a new description, but instead update the existing one
-    if ((1 == segments.size()) && (segments.front().location == coordinate))
-    {
-        if (path_point.segment_duration > 0)
-        {
-            segments.front().name_id = path_point.name_id;
-            segments.front().travel_mode = path_point.travel_mode;
-        }
-        return;
-    }
-
-    // make sure mode changes are announced, even when there otherwise is no turn
-    const auto getTurn = [](const PathData &path_point, const extractor::TravelMode previous_mode)
-    {
-        if (extractor::TurnInstruction::NoTurn == path_point.turn_instruction &&
-            previous_mode != path_point.travel_mode && path_point.segment_duration > 0)
-        {
-            return extractor::TurnInstruction::GoStraight;
-        }
-        return path_point.turn_instruction;
-    };
-
-    // TODO check why we require .front() here
-    const auto turn = segments.size() ? getTurn(path_point, segments.front().travel_mode)
-                                      : path_point.turn_instruction;
+    const auto turn = path_point.turn_instruction;
 
     segments.emplace_back(coordinate, path_point.name_id, path_point.segment_duration, 0.f, turn,
                           path_point.travel_mode);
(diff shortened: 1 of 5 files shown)


include/engine/routing_algorithms/routing_base.hpp
```diff
@@ -675,11 +679,11 @@ template <class DataFacadeT, class Derived> class BasicRoutingInterface
         BOOST_ASSERT_MSG((SPECIAL_NODEID != middle && INVALID_EDGE_WEIGHT != distance),
                          "no path found");
 
-        if (distance != forward_heap.GetKey(middle) + reverse_heap.GetKey(middle))
```
TheMarex: This looks like an unrelated bug fix, can we break this out?

include/extractor/edge_based_node.hpp
```diff
@@ -9,6 +9,7 @@
 #include "osrm/coordinate.hpp"
 
 #include <limits>
+#include <string>
```
TheMarex: We don't need string here.

include/engine/guidance/segment_list.hpp
```diff
@@ -235,6 +211,30 @@ void SegmentList<DataFacadeT>::Finalize(const bool extract_alternative,
...
+    }
+
+    // announce mode changes
+    for (std::size_t i = 0; i + 1 < segments.size(); ++i)
+    {
+        auto &segment = segments[i];
+        const auto next_mode = segments[i + 1].travel_mode;
+        if (segment.travel_mode != next_mode &&
+            segment.turn_instruction == extractor::TurnInstruction::NoTurn)
+        {
```
TheMarex: Couldn't we handle this in pre-processing in the turn generation?

MoKob: Not without adding additional information to the edges.

In turning from (a)->(b), we want to know both about the modes of (a) for all information stored along (a), as well as (b). The turn announces the mode change to the mode of (b) at (a). So you either have to store both the current and the next travel mode, or you have to traverse the path to find the travel mode.

include/engine/guidance/segment_list.hpp
```diff
@@ -153,7 +154,7 @@ void SegmentList<DataFacadeT>::AddLeg(const std::vector<PathData> &leg_data,
                                       const bool is_via_leg,
                                       const DataFacade *facade)
 {
-    for (const PathData &path_data : leg_data)
+    for (const auto path_data : leg_data)
```
TheMarex: Why the copy here?

MoKob: Because typo performance bug. Thats why. Good catch!

include/engine/segment_information.hpp
```diff
@@ -8,6 +8,8 @@
 #include "osrm/coordinate.hpp"
 #include <utility>
 
+#include <string>
```
TheMarex: Not needed by `SegmentInformation`.

include/extractor/edge_based_node.hpp
```diff
@@ -80,6 +80,7 @@ struct EdgeBasedNode
     int forward_offset;          // prefix sum of the weight up the edge TODO: short must suffice
     int reverse_offset;          // prefix sum of the weight from the edge TODO: short must suffice
     unsigned packed_geometry_id; // if set, then the edge represents a packed geometry
+
```
TheMarex: Remove whitespace change.

include/engine/routing_algorithms/routing_base.hpp
```diff
@@ -614,6 +616,8 @@ template <class DataFacadeT, class Derived> class BasicRoutingInterface
         std::sort(reverse_entry_points.begin(), reverse_entry_points.end(), entry_point_comparator);
 
         NodeID last_id = SPECIAL_NODEID;
+        forward_core_heap.Clear();
+        reverse_core_heap.Clear();
```
daniel-j-h: Wait how does this come in here, or better why was this missing before? Bug?

daniel-j-h:
Looks good code-wise. All the manual looping, erasing, shifting is a bit ugly (and the erasing somewhat dangerous), but that's because of the existing code / approach.

PR #1981, opened 2016-02-15
fix routing when requiring loop at a core node

MoKob:
just as title says, taken out of https://github.com/Project-OSRM/osrm-backend/pull/1976


TheMarex:
:+1: looks good to merge.

PR #1986, opened 2016-02-17
[WIP] Advanced guidance vocabulary

MoKob:
Updates our (just improved) guidance to a broader vocabulary using `turn types` and `modifiers`.

 - [x] add information on destination (left/right/ahead)
 - [ ] announce ramps
 - [ ] handle forks
 - [ ] handle end of road
 - [ ] come up with other instructions to handle


MoKob:
A preview of whats to come:

[this](http://map.project-osrm.org/?z=16&center=52.473331%2C13.345342&loc=52.476011%2C13.347573&loc=52.470626%2C13.342016&hl=en&ly=&alt=&df=&srv=) will become 
screen shot 2016-02-17 at 23 00 34

MoKob:
First example of a complex instruction [this](http://map.project-osrm.org/?z=17&center=52.470054%2C13.312731&loc=52.471410%2C13.313928&loc=52.468528%2C13.310580&hl=en&ly=&alt=&df=&srv=) will emit an 'end of road', because at least one intersection has been crossed, [this](http://map.project-osrm.org/?z=18&center=52.468745%2C13.311189&loc=52.468894%2C13.311481&loc=52.468528%2C13.310580&hl=en&ly=&alt=&df=&srv=) will not. 

screen shot 2016-02-17 at 23 54 45

Also notice the new information: your destination is on the left.

src/guidance/translator.cpp
```diff
@@ -0,0 +1,150 @@
+#include <cstddef>
```
TheMarex: OSRM output should be symbolic and language agnostic-only. It is okay to expose the symbolic modifiers, but we should not include textual information here.

Problems I'm seeing with that would be:
- Translations
- HTML wrapped text
- Typo fix would mean we need to roll a new OSRM release

This code should be easy to translate for `api-directions`/`osrm-frontend` though.

src/extractor/edge_based_graph_factory.cpp
```diff
@@ -418,8 +420,49 @@ void EdgeBasedGraphFactory::GenerateEdgeExpandedEdges(
     util::DEBUG_TURNS_START(debug_turns_path);
 #endif
 
+    struct CompareTurnPossibilities
```
TheMarex: Should be a lambda.

src/guidance/translator.cpp
MoKob: I still have to see what I am exposing. I am not saying that this should ever make it outside. It is part of my hack to actually test my guidance and see how it would look like in an application...

src/extractor/edge_based_graph_factory.cpp
MoKob: Again, temporary feature about to be deleted in the end...

emiltin:
for bike guidance, complicated intersections often result in many small turns, even though all you really need to do is soemthing like turn left. this is caused by the way intersections are often modellen for bikes and pedestrians. it would be really useful is osrm was smarter about  merging these close-by instructions.

MoKob:
@emiltin its on the list for down the line. Though I am not yet sure on how good bike navigation would have to sound like. I guess it would benefit much more from points of interests along the road.
Whenever you have some good proposals, I'd gladly see what I can do.

emiltin:
Here are som ideas for improving bike guidance:

- avoiding "miniature routing" in intersections by combining small steps
- provide alternative cues when ways/paths have no name, which is often the case for bike networks:
 - use way type (turn right into tunnel)
 - use intersection type (turn right at sideroad / intersection / fork)
 - use surface types (continue on gravel path / cobblestones)
 - use oneway tag to distinguish way (turn left into oneway street)
 - use elevation (turn left on path going up hill / down hill)
 - use tags on nodes (turn left after/before gate / signalised intersection)
- implement routing across squares/area (cross square and continue on ... )
- improved guidance when changing levels (take elevator up/down)

looking at polygons would up the possibility for much more context-sensitive instructions, by looking at nearby parks, houses, etc. but it's not possible with the current code unless you you an external spatial tool like PostGIS.

TheMarex:
@MoKob :warning: rebasing on `develop`

emiltin:
Maybe left/right side could also be announced beforehand? Eg. "You will reach your destination in 500m. It will be on your right side."

MoKob:
@emiltin when to announce is a different problem, isn't it? we give the length of the next segment and the instruction. So it's up to the application when to announce it.

emiltin:
I think you're right.

PR #1994, opened 2016-02-17
include boost/thread/lock_types.hpp

BergWerkGIS:


springmeyer:
This passed travis and appveyor (It does not look that way because I cancelled builds to get others going). Going to merge - thanks for the quick windows fix @BergWerkGIS!

Issue #1953, opened 2016-02-05
OSRM debugging tileserver


danpat:
When trying to debug routing problems with OSRM, it's helpful to have some insight into the underlying routing graph.

There's a compile time feature `-DDEBUG_GEOMETRY` that enables dumping of a big GeoJSON file with lots of short linestrings with properties showing edge weights and turn penalties.  It can be used to make maps that look like this:

![screen shot 2016-02-05 at 12 34 38 pm](https://cloud.githubusercontent.com/assets/1892250/12858434/0e3456f4-cc05-11e5-979d-144c2b9815eb.png)

The problem is that for large areas, the GeoJSON output is huge: >140GB for North America.

This ticket is to roughly describe building a custom tileserver that can generate debugging tiles for slippy maps on-the-fly.

Output could be raster files for use with Leaflet, or vector tiles for use with something like Mapbox GL JS.

Because of the discarding of data during `osrm-prepare`, I expect this would need to be a standalone server that knows how to load the node-based graph, and potentially any traffic data updates that have been applied by `osrm-prepare`.

danpat:
Done!

Issue #1857, opened 2016-01-05
Offsets from StaticRTree are incorrect after traffic data updates


danpat:
# Background

The `StaticRTree` stores an `EdgeBasedNode` struct, which contains `forward_offset`, `reverse_offset`, `forward_weight` and `reverse_weight`.  These values are calculated during `osrm-extract`.

If the traffic update features of `osrm-prepare` are used to update weights, the values stored in the `StaticRTree` are not currently updated.  This means that the `PhantomNode` offset values will be incorrect, leading to incorrect route lengths.

Practically, for a 2 coordinate route, the effect is probably quite small.  Traffic speeds that are close to the original esimates will mean that the offsets will only be wrong by a small amount, and in most cases, won't lead to different routing.

However, routes with lots of vias will be more significantly affected.  If traffic data is used, the `PhantomNode`s returned by the `StaticRTree` will not be in quite the right place, and the offset errors will compound along the route.

# The naive fix
The naive fix is to update the `StaticRTree` when new weights are loaded during `osrm-prepare`.  However, this will be a terribly expensive operation, as we would need to re-scan each compressed edge for each update to an offset made on a segment on that edge.  This will massively blow out the runtime for `osrm-prepare`, which is very undesirable for traffic updates.

# Proposed solution
The alternative is to calculate the offsets on the fly.  This will require retrieving segment durations (which will be up-to-date after traffic data imports), and calculating the offset values after the `PhantomNodes` are found.

https://github.com/Project-OSRM/osrm-backend/blob/1da1fa5a3174c8cc95b1b25e71323c9fc41fcad4/include/engine/datafacade/internal_datafacade.hpp#L427-L436

Similar to ^^, we could also fetch segment weights, then calculate the offsets.  This call would happen for each `via` in the request.

MoKob:
would also love to stay in the loop here.

TheMarex:
### Serialize out the information we actually need.

Luckily `CompressedEdgeContainer` already has the information we need: the duration for each segment. However we discard it on serialization [here](https://github.com/Project-OSRM/osrm-backend/blob/1da1fa5a3174c8cc95b1b25e71323c9fc41fcad4/src/extractor/compressed_edge_container.cpp#L78) (note that we only use `current_node.first`).

So we need to serialize out `second` as well.

### Remove `{forward,reverse}_{offset,weight}`

[...]

PR #2000, opened 2016-02-20
Vector tile plugin

danpat:
This PR is to address #1953, which started as a crazy idea, and then turned out to be not all that difficult to get working.

It adds:

   - a bounding-box search in the StaticRTree
   - a new `tile` plugin that will serve vector tiles at the `/tile?z={z}&x={x}&y={y}` URI

It's mostly clean, except for how the PBF is actually rendered by `osrm-routed`.  I had to add a special-case kludge in `request_handler.hpp` to render the `pbf` JSON part directly to the response stream, rather than going through one of the existing renderers.  I'm open to ideas on how to make this cleaner, but there's a built-in assumption that plugins will return JSON, and this plugin breaks that assumption.

The included `viewer.html` file is added to the repository root, but needs a better home.

Other notes:

   - `protozero` in `third_party/libosmium` was updated to include proper PBF write support
   - a new Buffer type was added to util::json so that downstream `node-osrm` can recognize string vs byte objects and render them directl

Known problems:

  - should probably implement limits on what zoom levels are honored, or cap the number if items returned in an RTree search - a tile that covers the whole USA would be....large
  - line-clipping hasn't been implemented, so long lines that make it a long way past the edges of the tiles sometimes experience a coordinate overflow and go off in strange directions.  This is most noticable at maximum zooms.
   - this plugin currently displays the `forward/reverse_weight` from the `EdgeBasedNodes` in the `StaticRTree`.  As per https://github.com/Project-OSRM/osrm-backend/issues/1857, these values are stale after traffic updates, so this plugin will need to be updated in line with the work done in https://github.com/Project-OSRM/osrm-backend/tree/dynamic_offset_calc
   - there are currently no tests
   - it would be nice to expose more data points, like turn costs (idea: once `dynamic_offset_calc` lands, turn costs are edge-based-node weight minus sum of segment weights (maybe, although last segment may have turn cost included, should it?)...

Please review.  I'm offline from Feb 20 to Feb 27, but I'd like to polish this and merge it when I get back, as i think we'll find it super useful.


willwhite:
What's next for getting this merged? I think there are two options:

- Merge into develop and then port the changes to https://github.com/Project-OSRM/osrm-backend/pull/1935
- Just rebase against 5.0 now and PR against https://github.com/Project-OSRM/osrm-backend/pull/1935

daniel-j-h:
There's a token in it, remove it and take care of it.

TheMarex:
> Merge into develop and then port the changes to #1935

I would like to do this as it means we have a stepping stone on the old API. This branch needs a proper cleanup first:

- [ ] Remove all the HTML debugging stuff
- [ ] Quick code review by @daniel-j-h
- [ ] Quick code review by @TheMarex

Some thoughts:
- Extention through `json::Buffer` is ugly but in the old architecture not avoidable

@daniel-j-h can you look into getting this merged? It seems like it also breaks the tests.

After this in `develop` I will rebase the 5.0 API to include this plugin.

alexshires:
Hi all,

I'm working on a osrm based project and have just discovered the recently added DEBUG_GEOMETRY option, which is very useful for our use case. However, from following your recent commits, you seem to plan to remove it in order to go straight to the final viewing option for debugging. Would it be possible to keep the debugging geojson output available please, and if so, can I do anything to help the development of this?

Regards,

Alex

daniel-j-h:
@TheMarex this needs some more work before merging.

In particular patching third_party/libosmium for protozero changes is ugly, and I would rather pull in the latest libosmium release. Doing this next. Will then refactor the code a bit.

Should we put the debug-viewer into a separate repository in the Project-OSRM organization? Or even incorporate this in the frontend in some optional way?

willwhite:
@daniel-j-h I rebased this on the `develop` branch to make sure 69bc8a8b2214da798ffd39f6ebd98d82a1bf5103 was included here. Just a heads up in case you start cleaning this up in the morning.

TheMarex:
@alexshires Yes this will replace the geojson output as it just does not scale over trivial data set sizes. Decoding vector tiles should not be too hard - also its easier to visualize as it requires no intermediate step.

alexshires:
@TheMarex thanks, no problem, I'd have liked to keep it for small enough datasets but we'll try and convert OSM to our own formatting.

daniel-j-h:
- [ ] decide where to put `debug-viewer` and pull it out of the Git history
- [ ] there's still duplicate code in `tile.hpp`, refactor it if possible
- [ ] use portable math constants as we do in the whole code base
- [ ] where is the c&p code from? at least mapnik and somewhere else? are there license issues?
- [ ] `Buffer` is a string containing bytes (what about `\0`?) that we escape and put out as JSON? Wat?
- [ ] there are lines simply commented out in the changes which I do not understand; clarify
- [ ] there are todos in the code, e.g. templating `RectangleInt2D`; want to do this?

daniel-j-h:
Seeing a few issues on the Travis OSX instances wrt. `bind<void>`. Let's see if switching to the latest XCode helps as @MoKob reported. If this works out, I'm handing this back to @danpat.

TheMarex:
Test passed, and rebased on develop, merging. :+1:

willwhite:
:clap:

PR #2010, opened 2016-02-24
Rip out mapbox/variant

daniel-j-h:
As discussed last week @springmeyer I can see static assertions with this trivial change though, which do not make sense to me. Care to look into it?


daniel-j-h:
Note: do not merge yet, need to adapt user code (i.e. example, node-osrm, ...).

springmeyer:
@daniel-j-h I'm not sure why you would see different behavior, but here are a few questions to consider:

  - Why did you also upgrade `OSMIUM_TAG`: https://github.com/Project-OSRM/osrm-backend/pull/2010/files#diff-4174b0d99cc9c12d5bc7503ff6663c6aR4
  - What is the purpose of the code changes around `QueryCandidate` and the addition of the assert at https://github.com/Project-OSRM/osrm-backend/pull/2010/files#diff-82d60e4743c741c744969b6ac2f682b9R402?

daniel-j-h:
- when you updated `libosmium` last week you did not update / use the the `update_dependencies.sh` script to pull the subtree, so `OSMIUM_TAG` was still referring to the version before your update. I removed variant from the script and saw this mismatch. The script is for manual usage only, never called by Travis or someone else.
- there is no longer a `is<T>` function, the way to do it is to check if `get<T>(&t)` is a `nullptr`, but since you can not say `if (const T* ptr = get<T>(&t))` the two cases have to be hoisted outside the `if/else`. And because we want to fail hard in case someone extends the variant but does not adapts the code in cases the type is neither of the two we crash in debug.

daniel-j-h:
The issue still remains, now the question is should we try to resolve it in time for v5 or do a mapbox/variant update for v5 as ticketed in https://github.com/Project-OSRM/osrm-backend/issues/2140.

TheMarex:

PR #2011, opened 2016-02-24
Don't drive or ride on whatever with highway and access=yes tags

frodrigo:
access=yes implied we can go on this highway but don't suppose anything about vehicles. So avoid hiking path with car like this http://map.project-osrm.org/?z=13&center=42.824302%2C-0.273714&loc=42.798739%2C-0.333281&loc=42.830739%2C-0.261676


emiltin:
Wrong tagging? Isn't "access" highest in the hierachy of transport modes, and as such implies access for all modes? http://wiki.openstreetmap.org/wiki/Key:access

frodrigo:
I don't think so, according to the wiki:
> yes 	The public has an official, legally-enshrined right of access; i.e., it's a right of way

TheMarex:
This is wrong tagging. Explicitly adding `access=yes` implies its accessible for all modes of transportation - as documented in the wiki if you read a little bit more down.

Yes this is unfortunate - but sadly its an unintuitive tagging scheme. Matching it against the default mode of transportation breaks down in cases like `highway=pedestrian` + `access=yes` which is quite common in city centres.

PR #2014, opened 2016-02-24
Run tests for all branches.

willwhite:
It's useful to have travis build all branches for feedback during
development.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,11 +5,6 @@ dist: trusty
 notifications:
   email: false
 
-branches:
-  only:
-    - master
-    - develop
-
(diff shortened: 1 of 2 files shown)


willwhite:
This also avoids the need to open a PR in order for travis to check your work. It helps people experiment without clogging up the PR queue.

TheMarex:
We have PRs for each branch on purpose, it makes it easier to track who is working on what and people can comment on the diff, without fearing that a force push will delete comments by other people.

The problem here is that enabling this means that each commit in a PR is build twice by travis. This will quickly clog up the build queue.

PR #2020, opened 2016-02-26
Vector tile plugin refactoring

daniel-j-h:
Branching off of https://github.com/Project-OSRM/osrm-backend/pull/2000 and working on it here, as the libosmium subtree was broken and I don't want to force-push to @danpat's branch. Once this is done we can still rewrite history.


daniel-j-h:
Merged back into `debug_tileserver` --- Github not picking it up. Closing.

PR #2041, opened 2016-03-02
Fixes critical error in table plugin accessing uninitialized memory

daniel-j-h:
Although we check for valid coordinates in the table plugin via
`check_all_coordinates`, we do not check for `#srcs > 0` and `#dsts > 0`.

This would be fine as the grammar parser combines adding coordinates and
setting their `is_source` and `is_destination` property, which makes
adding coordinates without specifying source or destination impossible.

See: [route_parameters.cpp, AddSource, AddDestination](https://github.com/Project-OSRM/osrm-backend/blob/becae462968a9f6e6b72b409d2d073f13138b55a/src/engine/route_parameters.cpp#L115-L140), and [api_grammar.hpp](https://github.com/Project-OSRM/osrm-backend/blob/becae462968a9f6e6b72b409d2d073f13138b55a/include/server/api_grammar.hpp#L110-L111)

In contrast, the Polyline codepath does not do this! In fact, it only lets
you set coordinates, but not their `is_source` or `is_destination` property.

See: [route_parameters.cpp, SetCoordinatesFromGeometry](https://github.com/Project-OSRM/osrm-backend/blob/becae462968a9f6e6b72b409d2d073f13138b55a/src/engine/route_parameters.cpp#L142-L145)

Therefore, the following queries only set coordinates:

    http 'http://localhost:5000/table?locs=s_hhFg{arEgEfEgEfEgEfEgEfEgEfEgEfEgEfEgEfEgEfE'
    http 'http://localhost:5000/table?locs=_p~iF~ps|U_ulLnnqC_mqNvxq`@'

but fail to specify sources and targets!

The distance table plugin now assumes `is_course` and `is_destination`
is the same size as `coordinates`.

And happily accesses [uninitialized memory](https://github.com/Project-OSRM/osrm-backend/blob/becae462968a9f6e6b72b409d2d073f13138b55a/include/engine/plugins/distance_table.hpp#L126).

References:
- https://gist.github.com/daniel-j-h/d770c4e3f0a74a8bd588

### DIFF
--- a/include/engine/plugins/distance_table.hpp
+++ b/include/engine/plugins/distance_table.hpp
@@ -59,18 +59,31 @@ template <class DataFacadeT> class DistanceTablePlugin final : public BasePlugin
             return Status::Error;
         }
 
-        auto number_of_sources =
-            std::count_if(route_parameters.is_source.begin(), route_parameters.is_source.end(),
-                          [](const bool is_source)
-                          {
-                              return is_source;
-                          });
-        auto number_of_destination =
-            std::count_if(route_parameters.is_destination.begin(),
-                          route_parameters.is_destination.end(), [](const bool is_destination)
-                          {
-                              return is_destination;
-                          });
+        // The check_all_coordinates guard above makes sure we have at least 2 coordinates.
+        // This guard makes sure is_source, is_destination, coordinates are parallel arrays.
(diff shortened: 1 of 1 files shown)


TheMarex:
:+1: on merging when travis passes.

PR #2043, opened 2016-03-02
[merge into rewrite/new-api] Fixes coordinate, source and destination validation by means of backporting #2041

daniel-j-h:
Read https://github.com/Project-OSRM/osrm-backend/pull/2041 first please.

Not sure if this is the right place to check for it.. should node-osrm handle it? /cc @lbud


TheMarex:
Closing as obsolete.

PR #2070, opened 2016-03-08
adds a relative position to the phantom node while removing it from the hint

MoKob:
To compute the relative position of the queried location to the end of the path, the phantom node needs to be augmented with the queried coordinate.

Re-enables the relative position that got broken in https://github.com/Project-OSRM/osrm-backend/commit/8a459c3ddbea26b3947a2406a772afb3747affcf

TODO:

 - [x] adjust hint to not contain a second copy
 - [x] @daniel-j-h takes a look

### DIFF
--- a/include/engine/api/base_api.hpp
+++ b/include/engine/api/base_api.hpp
@@ -8,6 +8,7 @@
 #include "engine/hint.hpp"
 
 #include <boost/assert.hpp>
+#include <boost/range/algorithm/transform.hpp>
 
 #include <vector>
 
@@ -33,24 +34,22 @@ class BaseAPI
 
         util::json::Array waypoints;
         waypoints.values.resize(parameters.coordinates.size());
-        waypoints.values[0] = MakeWaypoint(parameters.coordinates.front(),
-                                           segment_end_coordinates.front().source_phantom);
+        waypoints.values[0] = MakeWaypoint(segment_end_coordinates.front().source_phantom);
 
-        auto coord_iter = std::next(parameters.coordinates.begin());
         auto out_iter = std::next(waypoints.values.begin());
-        for (const auto &phantoms : segment_end_coordinates)
-        {
-            *out_iter++ = MakeWaypoint(*coord_iter++, phantoms.target_phantom);
-        }
+        boost::range::transform(segment_end_coordinates, out_iter,
+                                [this](const PhantomNodes &phantom_pair)
+                                {
+                                    return MakeWaypoint(phantom_pair.target_phantom);
+                                });
         return waypoints;
     }
 
   protected:
-    util::json::Object MakeWaypoint(const util::Coordinate input_coordinate,
-                                    const PhantomNode &phantom) const
+    util::json::Object MakeWaypoint(const PhantomNode &phantom) const
     {
         return json::makeWaypoint(phantom.location, facade.get_name_for_id(phantom.name_id),
-                                  Hint{input_coordinate, phantom, facade.GetCheckSum()});
+                                  Hint{phantom, facade.GetCheckSum()});
     }
 
     const datafacade::BaseDataFacade &facade;
--- a/include/engine/api/table_api.hpp
+++ b/include/engine/api/table_api.hpp
@@ -17,6 +17,8 @@
 
 #include "util/integer_range.hpp"
 
+#include <boost/range/algorithm/transform.hpp>
+
 namespace osrm
 {
 namespace engine
@@ -72,13 +74,12 @@ class TableAPI final : public BaseAPI
         util::json::Array json_waypoints;
         json_waypoints.values.reserve(phantoms.size());
         BOOST_ASSERT(phantoms.size() == parameters.coordinates.size());
-        auto phantom_iter = phantoms.begin();
-        auto coordinate_iter = parameters.coordinates.begin();
-        for (; phantom_iter != phantoms.end() && coordinate_iter != parameters.coordinates.end();
-             ++phantom_iter, ++coordinate_iter)
-        {
-            json_waypoints.values.push_back(BaseAPI::MakeWaypoint(*coordinate_iter, *phantom_iter));
-        }
+
+        boost::range::transform(phantoms, std::back_inserter(json_waypoints.values),
+                                [this](const PhantomNode &phantom)
+                                {
+                                    return BaseAPI::MakeWaypoint(phantom);
+                                });
         return json_waypoints;
     }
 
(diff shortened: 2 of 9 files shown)


TheMarex:
What is the actual feature that you want to get from this?If its on the left or right side of the street, right? So in that case you should add exactly that - `POSITION_LEFT` `POSITION_RIGHT`.

MoKob:
the position depends on the found path in the end, so it can only be computed after the path is found.

TheMarex:
@MoKob no you can encode the position of the input coordinate in relation the forward direction of the segment. All you need to know downstream then is if the segment was traversed in forward direction or reverse direction. This is information that is already contained in the `InternalRouteResult`.

That way you only need to extend `PhantomNode` by at most one byte (alignment will probably not even make it matter). `PhantomNode` size is important for the `Hint` that we return to the client, we want minimize the size there.

MoKob:
I switched it to a much more involved pre-calculated relative position. This will make it impossible, however, to accurately provide the location when we eliminate very short segments at the end of the route. Hopefully, these can be guessed in a reasonable way, but it might at least introduce some inconsistencies.

screen shot 2016-03-09 at 10 41 43

include/engine/geospatial_query.hpp
```diff
@@ -348,6 +348,56 @@ template <typename RTreeT, typename DataFacadeT> class GeospatialQuery
                 coordinates->at(data.u), coordinates->at(data.v), input_coordinate,
                 point_on_segment, ratio);
 
+        const auto translateAngleToLocation = [](const double angle)
```
TheMarex: No need to go over all this angle calculations. We can use what is already there. This can be simplified to something like:

```cpp
if (ratio < std::numeric_limits<double>::epsilon())
   return RelativeLocation::BEHIND;

if (ratio-1. < std::numeric_limits<double>::epsilon())
   return RelativeLocation::AHEAD;

auto distance = coordinate_caluclation::havensineDistance(point_on_segment, input_coordinate);
if (distance > 100. || distance < 3.)
    return RelativeLocation::INVALID;

bool is_left_of = coordinate_calculatetion::is_left_of(coordinates[data.u], coordinates[data.v], coordinate);

if (is_left_of)
  return RelativeLocation::Left;

return RelativeLocation::Right;
```
You can port `is_left_of` over from [here](https://github.com/TheMarex/deberg/blob/master/geometry.hpp#L135-L148). Coordinates need to be transformed to mercator first in the helper function.

MoKob: That proposal does not result in the same information that I want to have in the search.
                      A
              
***************        would result in `Ahead`, but should be `Left`, same with `Behind`.

If the ratio thing is ignored,
                                                  A
***************
would result in `Left` instead of `Ahead`.

My proposal how to handle this and other conflicts down the line remains to push the coordinate through the phantom node and do the computations in assemble_steps.

This would require a fix in the hint, apparently, since it is computed 1:1 from the phantom node. In my view, it should not include this information, if we have to pass it via the phantom node.

If we want to fix unnecessary segments during post-processing (snapped to the wrong phantom node at an intersection), we require the coordinate for correct calculations.

MoKob:
I reverted back to my original proposal. To avoid any effect on the hint, I could move the location out of the hint and only use the one in the phantom node.

This also reduces some complexity along the basic API waypoint construction.

MoKob:
One additional thing to discuss: I added a default parameter to the hint calculation. Might be that we want to reduce potential errors by explicitly specifying it whenever the coordinate does not matter?

include/engine/api/base_api.hpp
```diff
@@ -33,24 +33,21 @@ class BaseAPI
...
-        waypoints.values[0] = MakeWaypoint(parameters.coordinates.front(),
-                                           segment_end_coordinates.front().source_phantom);
+        waypoints.values[0] = MakeWaypoint(segment_end_coordinates.front().source_phantom);
 
-        auto coord_iter = std::next(parameters.coordinates.begin());
         auto out_iter = std::next(waypoints.values.begin());
         for (const auto &phantoms : segment_end_coordinates)
         {
-            *out_iter++ = MakeWaypoint(*coord_iter++, phantoms.target_phantom);
+            *out_iter++ = MakeWaypoint(phantoms.target_phantom);
```
daniel-j-h: these constructs are the perfect candidates for algorithms instead of raw loops

include/engine/api/table_api.hpp
```diff
@@ -72,12 +72,9 @@ class TableAPI final : public BaseAPI
...
         json_waypoints.values.reserve(phantoms.size());
         BOOST_ASSERT(phantoms.size() == parameters.coordinates.size());
-        auto phantom_iter = phantoms.begin();
-        auto coordinate_iter = parameters.coordinates.begin();
-        for (; phantom_iter != phantoms.end() && coordinate_iter != parameters.coordinates.end();
-             ++phantom_iter, ++coordinate_iter)
+        for ( const auto & phantom : phantoms )
         {
-            json_waypoints.values.push_back(BaseAPI::MakeWaypoint(*coordinate_iter, *phantom_iter));
+            json_waypoints.values.push_back(BaseAPI::MakeWaypoint(phantom));
```
daniel-j-h: same here ^ also if you emplace_back you don't have to repeat the verbose type

include/engine/phantom_node.hpp
```diff
@@ -177,7 +186,7 @@ struct PhantomNode
 };
 
 #ifndef _MSC_VER
-static_assert(sizeof(PhantomNode) == 52, "PhantomNode has more padding then expected");
+static_assert(sizeof(PhantomNode) == 60, "PhantomNode has more padding then expected");
```
MoKob: @BergWerkGIS I cannot verify on my system whether this size captures the size on a Windows system correctly and the AppVeiyor builds are currently not working. Could you maybe verify whether this `static assert` works?

BergWerkGIS: Will look into this, but it might take a while, as I have to get a build going first.
Serach for `: error` in https://ci.appveyor.com/project/DennisOSRM/osrm-backend/build/1.0.2784
e.g. have to adjust [`windows-builds`](https://github.com/mapbox/windows-builds) to build  `boost_unit_test_framework` and create a new binary deps package.

include/engine/phantom_node.hpp
```diff
@@ -169,6 +177,7 @@ struct PhantomNode
     static_assert(sizeof(ComponentType) == 4, "ComponentType needs to 4 bytes big");
 #endif
     util::Coordinate location;
+    util::Coordinate input_coordinate;
```
TheMarex: should be `input_location` to be consistent.

include/engine/phantom_node.hpp
```diff
@@ -144,6 +151,7 @@ struct PhantomNode
         component.is_tiny = other.component.is_tiny;
 
         location = foot_point;
+        this->input_coordinate = input_coordinate;
```
TheMarex: `foot_point` -> `location_`. `input_location` -> `input_location_`. This eliminates the `this->`.

TheMarex:
Create! Rebasing and merging.

PR #2073, opened 2016-03-09
[not ready] Modular Buildsystem

daniel-j-h:
Root `CMakeLists.txt` calls `add_subdirectory`. Sub directories provide their own `CMakeLists.txt` setting up dependencies, include paths and libraries to link in.

This is a first try to modularize the 500 lines of `CMakeLists.txt` mess we're in.

### DIFF
--- a/unit_tests/CMakeLists.txt
+++ b/unit_tests/CMakeLists.txt
@@ -33,7 +33,11 @@ add_executable(util-tests
 set(AllBoostLibrariesExceptUnitTest ${Boost_LIBRARIES})
 
 find_package(Boost 1.49.0 REQUIRED COMPONENTS unit_test_framework)
-add_definitions(-DBOOST_TEST_DYN_LINK)
+
+if(NOT WIN32)
+  add_definitions(-DBOOST_TEST_DYN_LINK)
+endif()
+
 include_directories(SYSTEM ${Boost_INCLUDE_DIRS})
 
 
--- a/unit_tests/CMakeLists.txt
+++ b/unit_tests/CMakeLists.txt
@@ -1,0 +1,69 @@
+set(ExtractorTest_SOURCES
+	extractor_tests.cpp
+	extractor/compressed_edge_container.cpp
+	extractor/graph_compressor.cpp
+	extractor/raster_source.cpp)
+
+set(EngineTest_SOURCES
+	engine_tests.cpp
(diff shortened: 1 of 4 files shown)


unit_tests/CMakeLists.txt
```diff
@@ -0,0 +1,69 @@
...
+	util/bearing.cpp
+	util/binary_heap.cpp
+	util/coordinate.cpp
+	util/duration_parsing.cpp
+	util/dynamic_graph.cpp
+	util/io.cpp
+	util/range_table.cpp
+	util/static_graph.cpp
+	util/static_rtree.cpp
+	util/string_util.cpp)
```
daniel-j-h: We had some `IOTest` globs in here before, but we never built an executable from it.
What was the plan for that, @TheMarex?

Also the globs just silently fail even in the case when the directory is not even there. So being explicit here is a good thing.

TheMarex: I want to avoid exactly this, hard coding file names in builds. Please revert to globs again. This just means every time you add move a file you need to remember to rename it here as well. Double book-keeping like this is just error prone and globs just solve this for us.

daniel-j-h: In my opinion you should know _exactly_ what files your building an object file with. That's also how we did it in rtd. In the worst case you get a linker error. Anyway, it's not worth it bike-shedding on smth. like this. Will re-work with globs.

TheMarex:
:warning: rebased on `develop` to get travis builds running again

TheMarex:
Strangely travis fails with https://travis-ci.org/Project-OSRM/osrm-backend/jobs/115616684#L1712 seems like it can't find lua5?

daniel-j-h:
:sigh: this is unrelated to the changes _I think_. CMake should pick up a specific Lua version and then link against e.g. `-llua5.2`. I can not reproduce it locally.

daniel-j-h:
It seems like we need this per https://github.com/Project-OSRM/osrm-backend/issues/2197#issuecomment-204864938.

daniel-j-h:
Rebases and adapted to changes on develop, let's see how Travis likes this now.

daniel-j-h:
Closing in favor of https://github.com/Project-OSRM/osrm-backend/pull/2227 which works on top of v5 since we diverged too much in the last weeks.

PR #2075, opened 2016-03-10
report depart/arrive instead of waypoints

MoKob:
Adds Depart/Arrive to the API response.

I kept this not as additional instruction, as this would remove space for turn-instructions stored at arcs.

### DIFF
--- a/include/engine/guidance/step_maneuver.hpp
+++ b/include/engine/guidance/step_maneuver.hpp
@@ -4,6 +4,8 @@
 #include "util/coordinate.hpp"
 #include "extractor/guidance/turn_instruction.hpp"
 
+#include <cstdint>
+
 namespace osrm
 {
 namespace engine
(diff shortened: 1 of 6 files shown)


include/engine/guidance/step_maneuver.hpp
```diff
@@ -11,12 +13,21 @@ namespace engine
 namespace guidance
 {
 
+enum class WaypointType : std::uint8_t
+{
+    None,
+    Arrive,
+    Depart,
+    Waypoint
```
TheMarex: This does not exist anymore because of the leg split. I would expect the instructions for a leg to be:
`Depart, [...], Arrive`. I don't think we need a `Waypoint` type - for instructions our biggest unit is a leg, hence there are no via.

TheMarex:
Okay this is fine. I see your point about `Location` modifier not being used in the Extractor, hence it should not be part of `TurnInstruction`. :+1:

Merging.

PR #2078, opened 2016-03-10
Fix OSX Travis builds for recent images

daniel-j-h:
Our Travis OSX builds are broken since a few days (?). It seems like the OXS image now comes with libxml2, GDAl and boost installed by default. Installing libxml2 and GDAL only shows a warning, trying to install Boost throws an error. Here's an example log: https://travis-ci.org/Project-OSRM/osrm-backend/jobs/115030045#L60

The trivial fix here is to assume libxml2, GDAL and boost as implicit dependencies (ugly).

Is there a way to pin versions or at least specify all dependencies explicitly? From searching around this is a wontfix in brew. /cc OSX users for ideas @MoKob @danpat @springmeyer

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -127,7 +127,8 @@ install:
       export PATH=${DEPS_DIR}/cmake/bin:${PATH}
 
     elif [[ "${TRAVIS_OS_NAME}" == "osx" ]]; then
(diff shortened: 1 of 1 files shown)


TheMarex:
PR against wrong branch. Otherwise :+1: Reopening.

TheMarex:
Cherry-picked as c2e49bd15270ae3e9decc00101b969712016e110

PR #2080, opened 2016-03-10
static turn instructions

MoKob:
For this branch to be merged, the following needs to happen:

 - [x] do not expose the internal conflict resolution
 - [x] do not expose `new name` on near identical names
 - [x] add road classes (e.g. [primary vs residential](http://www.openstreetmap.org/edit#map=18/41.69060/-120.37531)) to fork detection
 - [x] clean-up response to only return valid turn candidates
 
~~- [ ] add cucumber tests for static turn instructions~~
~~- [x] motorways~~
~~- [ ] fork~~
~~- [ ] end of road~~
~~- [ ] new name~~
~~- [ ] suppressed~~
~~- [ ] continue~~
~~- [ ] merge~~
~~- [ ] ramp~~
 - [x] test roundabouts
      - [x] enter
      - [x] immediately exit
      - [x] enter and exit
      - [x] only exit
 - [x] review, I am looking at you @daniel-j-h

### DIFF
--- a/include/extractor/guidance/toolkit.hpp
+++ b/include/extractor/guidance/toolkit.hpp
@@ -339,8 +339,12 @@ inline bool requiresNameAnnounced(const std::string &from, const std::string &to
     // this uses the encoding of references in the profile, which is very BAD
     // Input for this function should be a struct separating streetname, suffix (e.g. road,
     // boulevard, North, West ...), and a list of references
-    std::string from_name = "", from_ref = "", to_name = "", to_ref = "";
+    std::string from_name;
+    std::string from_ref;
+    std::string to_name;
+    std::string to_ref;
 
+    // Split from the format "{name} ({ref})" -> name, ref
     auto split = [](const std::string &name, std::string &out_name, std::string &out_ref)
     {
         const auto ref_begin = name.find_first_of('(');
@@ -352,7 +356,6 @@ inline bool requiresNameAnnounced(const std::string &from, const std::string &to
         else
         {
             out_name = name;
-            out_ref = "";
         }
     };
 
@@ -360,32 +363,20 @@ inline bool requiresNameAnnounced(const std::string &from, const std::string &to
     split(to, to_name, to_ref);
 
     // check similarity of names
-    if (from_name != "" && to_name != "")
-    {
-        if ((from_name.back() >= '0' && from_name.back() <= '9') ||
-            (to_name.back() >= '0' && to_name.back() <= '9'))
-            return from_name != to_name;
-        if (from.find("Weg ") == 0 && to_name.find("Weg ") == 0)
-            return from_name != to_name;
-        auto from_itr = from_name.begin();
-        auto to_itr = to_name.begin();
-        for (; from_itr != from_name.end() && to_itr != to_name.end() && *from_itr == *to_itr;
-             ++to_itr, ++from_itr)
-        {
-            /* do nothing */
-        }
+    auto names_are_empty = from_name.empty() && to_name.empty();
+    auto names_are_equal = from_name == to_name;
+    auto name_is_removed = !from_name.empty() && to_name.empty();
+    // references are contained in one another
+    auto refs_are_empty = from_ref.empty() && to_ref.empty();
+    auto ref_is_contained =
+        !from_ref.empty() && !to_ref.empty() &&
+        (from_ref.find(to_ref) != std::string::npos || to_ref.find(from_ref) != std::string::npos);
+    auto ref_is_removed = !from_ref.empty() && to_ref.empty();
 
-        const auto common_length = std::distance(from_name.begin(), from_itr);
-        return (100 * common_length / std::min(to_name.length(), from_name.length())) < 80;
-    }
-    else if (from_ref != "" && to_ref != "")
-    {
-        // references are contained in one another
-        if (from_ref.find(to_ref) != std::string::npos ||
-            to_ref.find(from_ref) != std::string::npos)
-            return false;
-    }
-    return true;
+    auto obvious_change = ref_is_contained || names_are_equal ||
+                          (names_are_empty && refs_are_empty) || name_is_removed || ref_is_removed;
+
+    return !obvious_change;
 }
 
 inline int getPriority( const FunctionalRoadClass road_class )
--- a/include/extractor/guidance/toolkit.hpp
+++ b/include/extractor/guidance/toolkit.hpp
@@ -12,9 +12,12 @@
 #include "extractor/guidance/classification_data.hpp"
 #include "extractor/guidance/turn_instruction.hpp"
 
+#include <algorithm>
 #include <map>
 #include <cmath>
 #include <cstdint>
+#include <string>
+#include <iostream>
 
 namespace osrm
 {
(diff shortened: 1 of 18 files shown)


include/extractor/guidance/turn_instruction.hpp
```diff
@@ -43,8 +43,14 @@ enum TurnType // at the moment we can support 32 turn types, without increasing
     NewName,                // no turn, but name changes
```
TheMarex: Can we merge `NewName` and `Notification`? `Notification` just tells the user that a street attribute (like name or travel mode) changed. This would be forward compatible if we add more attributes for which we might want to emit a `Notification.

include/extractor/guidance/turn_instruction.hpp
```diff
@@ -43,8 +43,14 @@ enum TurnType // at the moment we can support 32 turn types, without increasing
...
     Continue,               // remain on a street
     Turn,                   // basic turn
+    FirstTurn,              // First of x turns
+    SecondTurn,             // Second of x turns
+    ThirdTurn,              // Third of x turns
     Merge,                  // merge onto a street
     Ramp,                   // special turn (highway ramp exits)
+    FirstRamp,              // numbered ramp
```
TheMarex: Do you see any way of moving this additional information outside the turn type? I mean "only" handling four ramps at the same node is fine with me, but how will this evolve once we also emit intersection shapes?

TheMarex:
:warning: rebased on the new `rewrite/new-api`.

TheMarex:

include/extractor/guidance/turn_instruction.hpp
MoKob: Right Now Notification is Unused due to data not being exposed. I wouldn't worry about that one right now. The moment we get more data into the system, we have to open a discussion of what to expose.

include/extractor/guidance/toolkit.hpp
```diff
@@ -12,9 +12,12 @@
 #include "extractor/guidance/classification_data.hpp"
 #include "extractor/guidance/turn_instruction.hpp"
 
+#include <algorithm>
 #include <map>
 #include <cmath>
 #include <cstdint>
+#include <string>
+#include <iostream>
```
daniel-j-h: can be removed?

include/extractor/guidance/toolkit.hpp
```diff
@@ -315,6 +323,84 @@ inline bool isLowPriorityRoadClass(const FunctionalRoadClass road_class)
...
+
+    if ((second + 1) % detail::num_direction_modifiers == first)
+        return false;
+
+    return true;
+}
+
+inline bool requiresNameAnnounced(const std::string &from, const std::string &to)
+{
+    // FIXME, handle in profile to begin with?
```
daniel-j-h: yes we should ticket that; also the "Weg" handling below is ugly here. Maybe for post-v5?

include/extractor/guidance/toolkit.hpp
```diff
@@ -315,6 +323,84 @@ inline bool isLowPriorityRoadClass(const FunctionalRoadClass road_class)
...
+            to_ref.find(from_ref) != std::string::npos)
+            return false;
+    }
+    return true;
+}
+
+inline int getPriority( const FunctionalRoadClass road_class )
+{
+    const constexpr int road_priority[] = {10, 0, 10, 2, 10, 4, 10, 6, 10, 8, 10, 11, 10, 12, 10, 14};
+    return road_priority[static_cast<int>(road_class)];
```
daniel-j-h: Where did you get those priorities from? I can see them coming from the functional classes, but it's hard to follow here.

include/extractor/guidance/toolkit.hpp
```diff
@@ -315,6 +323,84 @@ inline bool isLowPriorityRoadClass(const FunctionalRoadClass road_class)
...
+}
+
+inline bool canBeSeenAsFork(const FunctionalRoadClass first, const FunctionalRoadClass second)
+{
+    // forks require similar road categories
+    return std::abs(getPriority(first) - getPriority(second)) <= 1;
```
daniel-j-h: Why? Care to elaborate how this influences decisions? I also can't find users in this pull request. Is this for future use?

include/extractor/guidance/turn_analysis.hpp
```diff
@@ -20,42 +23,61 @@ namespace extractor
...
+// For the turn analysis, we require a full list of all connected roads to determine the outcome.
+// Invalid turns can influence the perceived angles
+//
+// aaa(2)aa
+//          a - bbbbb
+// aaa(1)aa
+//
+// will not be perceived as a turn from (1) -> b, and as a U-turn from (1) -> (2).
+// In addition, they can influence whether a turn is obvious or not.
+struct ConnectedRoad : public TurnOperation
```
daniel-j-h: Is this semantically correct? A connected road is a turn operation?

include/extractor/guidance/turn_analysis.hpp
```diff
@@ -82,10 +106,9 @@ class TurnAnalysis
     //
     // The treatment results in a straight turn angle of 180º rather than a turn angle of approx
     // 160
-    std::vector<TurnCandidate>
-    mergeSegregatedRoads(const NodeID from_node,
-                         const EdgeID via_eid,
-                         std::vector<TurnCandidate> turn_candidates) const;
+    std::vector<ConnectedRoad> mergeSegregatedRoads(const NodeID from_node,
+                                                    const EdgeID via_eid,
+                                                    std::vector<ConnectedRoad> intersection) const;
```
daniel-j-h: fyi, intersection is passed by value, is this what you want to do here?

include/extractor/guidance/turn_analysis.hpp
```diff
@@ -93,91 +116,69 @@ class TurnAnalysis
...
     // Produces instructions to enter/exit a roundabout or to stay on it.
     // Performs the distinction between roundabout and rotaries.
-    std::vector<TurnCandidate> handleRoundabouts(const NodeID from,
+    std::vector<ConnectedRoad> handleRoundabouts(const NodeID from,
                                                  const EdgeID via_edge,
                                                  const bool on_roundabout,
                                                  const bool can_enter_roundabout,
                                                  const bool can_exit_roundabout,
-                                                 std::vector<TurnCandidate> turn_candidates) const;
+                                                 std::vector<ConnectedRoad> intersection) const;
```
daniel-j-h: same here ^

include/extractor/guidance/turn_instruction.hpp
```diff
@@ -42,8 +42,16 @@ enum TurnType // at the moment we can support 32 turn types, without increasing
     NewName,                // no turn, but name changes
     Continue,               // remain on a street
     Turn,                   // basic turn
+    FirstTurn,              // First of x turns
+    SecondTurn,             // Second of x turns
+    ThirdTurn,              // Third of x turns
+    FourthTurn,             // Fourth of x turns
```
daniel-j-h: per the zero-one-infinity rule: why do we handle 1-4 explicitly?

include/util/name_table.hpp
```diff
@@ -0,0 +1,26 @@
...
+namespace util
+{
+class NameTable
+{
+  private:
+    //FIXME should this use shared memory
+    RangeTable<16, false> m_name_table;
+    ShM<char, false>::vector m_names_char_list;
+  public:
+    NameTable( const std::string &filename );
```
daniel-j-h: format

include/util/name_table.hpp
```diff
@@ -0,0 +1,26 @@
...
+#include "util/shared_memory_vector_wrapper.hpp"
+#include "util/range_table.hpp"
+
+#include <string>
+
+namespace osrm
+{
+namespace util
+{
+class NameTable
```
daniel-j-h: what is a name table good for, what's its purpose? skimming the class says nothing about it. I assume for street names?

src/extractor/edge_based_graph_factory.cpp
```diff
@@ -302,8 +302,8 @@ void EdgeBasedGraphFactory::GenerateEdgeExpandedEdges(
     // Three nested loop look super-linear, but we are dealing with a (kind of)
     // linear number of turns only.
     util::Percent progress(m_node_based_graph->GetNumberOfNodes());
-    guidance::TurnAnalysis turn_analysis( *m_node_based_graph, m_node_info_list,
-                                       *m_restriction_map, m_barrier_nodes, m_compressed_edge_container );
+    guidance::TurnAnalysis turn_analysis(*m_node_based_graph, m_node_info_list, *m_restriction_map,
```
daniel-j-h: Hm why is restriction map wrapped in a shared_ptr?

src/util/name_table.cpp
```diff
@@ -0,0 +1,52 @@
...
+#include <boost/filesystem/fstream.hpp>
+
+namespace osrm
+{
+namespace util
+{
+
+NameTable::NameTable(const std::string &filename)
+{
+    boost::filesystem::ifstream name_stream(filename, std::ios::binary);
```
daniel-j-h: if (!name_stream) throw

src/util/name_table.cpp
```diff
@@ -0,0 +1,52 @@
...
+
+    name_stream >> m_name_table;
+
+    unsigned number_of_chars = 0;
+    name_stream.read(reinterpret_cast<char *>(&number_of_chars), sizeof(number_of_chars));
+    BOOST_ASSERT_MSG(0 != number_of_chars, "name file broken");
+    m_names_char_list.resize(number_of_chars + 1); //+1 gives sentinel element
```
daniel-j-h: What is the sentinel needed for? the number_of_chars + 1 elements are default inserted, which 1/ allocated memory and 2/ initializes them to zero. Do you only care for the last zero?

MoKob: I honestly didn't think about it. I just copied it from the datafacade. The problem here, I require the names which have been serialized to disk and cannot be accessed later without reeding them back in. I cannot think of a reasonable way to have this in here, but @TheMarex reasonably said that some of these tests should be performed in pre-processing, if possible

include/extractor/guidance/toolkit.hpp
MoKob: They are used in the not shown `TurnAnalsyis.cpp`. It essentially decides that we do not discover a fork for a service road leaving a motorway.

include/extractor/guidance/turn_analysis.hpp
MoKob: Technically it contains a possible turn operation with a flag. We could set the Turn Operation as variable for that struct. Should not change much, but would elongate all accesses down the line.

include/extractor/guidance/turn_analysis.hpp
MoKob: I tried to follow your guidance for inplace modification + move in this pull request. So yes, it should be this way.

include/extractor/guidance/turn_instruction.hpp
MoKob: because this is the only thing that happens in not fully messed up data-modelling. 
Haven;t seen anything that has more than four turns on a single side and wasn't a data modelling problem. They will probably make room for something better down the line.

src/extractor/edge_based_graph_factory.cpp
MoKob: Good question. It has been there and I didn't wand to mix clean-up with new features too much.

MoKob:
Moved the remaining tests here: https://github.com/Project-OSRM/osrm-backend/issues/2114

TheMarex:
:warning: rebasing on `rewrite/new-api`.

include/util/name_table.hpp
```diff
@@ -0,0 +1,31 @@
...
+{
+  private:
+    // FIXME should this use shared memory
+    RangeTable<16, false> m_name_table;
+    ShM<char, false>::vector m_names_char_list;
+
+  public:
+    NameTable(const std::string &filename);
+    std::string get_name_for_id(const unsigned name_id) const;
```
TheMarex: Needs to be CamelCase.

include/extractor/guidance/toolkit.hpp
TheMarex: Agreed. We can move the more complex merging out of this PR to more sophisticated methods. I'm going to simplify the code below and PR the diff to the current code.

TheMarex:
Okay this is in good shape now. Good work @MoKob! Merging back to the rewrite branch.

PR #2085, opened 2016-03-10
[DON'T MERGE] Test the minimal linux image

TheMarex:


daniel-j-h:
For the record, documentation is here: https://docs.travis-ci.com/user/trusty-ci-environment/#Image-differences-from-Precise and it seems to be a new feature :tada: nice!

PR #2086, opened 2016-03-11
Switch example to monaco

TheMarex:
This should save a lot of time building.


daniel-j-h:
Yes, thanks for catching this. Fixed in https://github.com/Project-OSRM/osrm-backend/commit/5b30f926776598ab7d7e8c1ce55258fa54ca02f8

@TheMarex would love if you could leave pull requests open for a bit before merging :+1:

PR #2088, opened 2016-03-12
[Not Ready] Unit Tests for v5

daniel-j-h:
Tracks progress on unit tests for v5 https://github.com/Project-OSRM/osrm-backend/pull/1935

- [X] fix all tests and benchmarks to compile again
- [x] make all tests pass again for v5 (except rtree test)
- [ ] guidance assembly should be tested
  - [ ] assemble_overview
  - [ ] assemble_geometry
  - [ ] assemble_steps
  - [ ] assemble_route
  - [ ] assemble_legs
- [ ] typed services need to be tested
  - [ ] `osrm.Route()`
  - [ ] `osrm.Match()`
  - [ ] `osrm.Table()`
  - [x] `osrm.Nearest()`
  - [ ] `osrm.Trip()`
  - [x] `osrm.Tile()`
- [ ] `parameter_parser.cpp` The current unit tests don't test all parameter types
- [x] `Hint::fromBase64` and `Hint::toBase64` need to be tested


TheMarex:
> make all tests pass again for v5

There is a StaticRTree test that is not supposed to pass - it documents the math problems that we currently have in the StaticRTree.

daniel-j-h:
Thanks for flagging this. It's only a single server unit test not passing. For the record, I could not base64-decode the hints in https://github.com/Project-OSRM/osrm-backend/blob/unit-tests-for-v5/unit_tests/server/parameters_parser.cpp#L172-L202, more information in https://github.com/Project-OSRM/osrm-backend/commit/bcf7c3130aa3be9a16d1e79bb9981592b7e7d973.

daniel-j-h:
RFC 4648 Test Vectors failing: https://github.com/Project-OSRM/osrm-backend/commit/154c6e3fd0eb0bdb0d1be5a1cd1f8c3944cbe867#diff-9646fe27bfa4361c1e0889b6bbbaf10cR6.
Seems like our Base64 encoding is broken.

daniel-j-h:
Base64 encoding/decoding was broken in so many ways. For example we forgot to take care of C-string null-terminated literals and byte-wise dumped everything even into `std::string` (which is bad).

I re-worked it completely in https://github.com/Project-OSRM/osrm-backend/commit/33af45a83cba1239578351ba3e6da32c9b385fb0 now the API does not shoot you in the foot by default any longer. Which is progress at least. The API is now as follows:

```c++
// Encoding Implementation

// Encodes a chunk of memory to Base64.
inline std::string encodeBase64(const unsigned char *first, std::size_t size);

// Overload for signed char catches (not only but also) C-string literals.
inline std::string encodeBase64(const char *first, std::size_t size);

// Convenience specialization, encoding from string instead of byte-dumping it.
inline std::string encodeBase64(const std::string &x);

// Encode any sufficiently trivial object to Base64.
template <typename T> std::string encodeBase64Bytewise(const T &x);

// Decoding Implementation

// Decodes into a chunk of memory that is at least as large as the input.
template <typename OutputIter> void decodeBase64(const std::string &encoded, OutputIter out);

// Convenience specialization, filling string instead of byte-dumping into it.
inline std::string decodeBase64(const std::string &encoded);

// Decodes from Base 64 to any sufficiently trivial object.
template <typename T> T decodeBase64Bytewise(const std::string &encoded);
```

With a `Bytewise` suffix indicating byte-wise dumping the user-provided `T`. We could probably SFINAE out the suffix and let the compiler detect strings and literals, but writing `Bytewise` instead of implementing horrible SFINAE magic is good enough here.

daniel-j-h:
The [last two tests I added](https://github.com/Project-OSRM/osrm-backend/blob/d30f7741630af5de9681d063d46e365393f66f6a/unit_tests/engine/base64.cpp#L41-L77) are failing at the moment.

Looks like the last couple of bytes are wrong:

```
AAAAgAAAAID___________________9_____fwAAAAAAAAAA__________8AAAAAAAAAgAAAAIAAAAAAAAAAAA==
AAAAgAAAAID___________________9_____fwAAAAAAAAAA__________8AAAAAAAAAgAAAAIAAAAAA3W-8Mg==
```

TheMarex:

daniel-j-h:
[Table config limit unit test I added](https://github.com/Project-OSRM/osrm-backend/blob/26b27bffaf201a3cf80a460ee41584740184a079/unit_tests/library/limits.cpp#L77-L104) is currently failing to respect the v5 API spec, ticket is over at https://github.com/Project-OSRM/osrm-backend/issues/2100.

daniel-j-h:
The [last commit](https://github.com/Project-OSRM/osrm-backend/pull/2088/commits/898049a6d5ffe3c8882b2a7a80f7ac33f11bb8fd) changes the Table plugin to return the correct `TooBig` error code on engine limit violations as per [the v5 server API](https://github.com/Project-OSRM/osrm-backend/wiki/New-Server-api#response). @lbud @karenzshea would you be so kind and make sure this does not break your code in `libosrm` consumer projects.

daniel-j-h:
[Nearest no-coordinates-test I added](https://github.com/Project-OSRM/osrm-backend/blob/5e7a0eb7ddcc9f6278d1dd4cce4e53209a087d17/unit_tests/library/nearest.cpp#L45-L60) is currently failing to respect the v5 API spec, ticket is over at https://github.com/Project-OSRM/osrm-backend/issues/2108.

- [ ] resolve in nearest plugin

daniel-j-h:
Tile plugin and parameters need some improvements, ticket is over at https://github.com/Project-OSRM/osrm-backend/issues/2109.

- [ ] resolve in tile plugin and tile parameters

daniel-j-h:
[Route plugin initial test I added](https://github.com/Project-OSRM/osrm-backend/pull/2088/commits/f47a6b85c22b3f0d4b8d60cd0bfeccd3faa1c1ae) has some tests failing where the v5 spec is unclear or ambiguous.

- [ ] @TheMarex clarifies v5 spec wrt. failing tests
- [ ] adapt tests to make them pass or adapt spec and impl.

daniel-j-h:
[The Tile plugin fix I added](https://github.com/Project-OSRM/osrm-backend/pull/2088/commits/0baaa6304bbc5c3becb3c244616ff6a1a998c44c) only validates in debug mode / for debug builds as we do for all parameter validation. We expect the caller to adhere to the parameter's preconditions. This also means:

- [ ] routed needs to validate /tile(x,y,z) these before passing them through
- [ ] node-osrm needs to validate /tile(x,y,z) before passing them through /cc @lbud @karenzshea

Precondition and references are in the [commit](https://github.com/Project-OSRM/osrm-backend/pull/2088/commits/0baaa6304bbc5c3becb3c244616ff6a1a998c44c).

unit_tests/library/route.cpp
```diff
@@ -22,15 +22,124 @@ BOOST_AUTO_TEST_CASE(test_route)
...
     RouteParameters params;
-
+    params.coordinates.emplace_back(util::FloatLongitude{}, util::FloatLatitude{});
     params.coordinates.emplace_back(util::FloatLongitude{}, util::FloatLatitude{});
     params.coordinates.emplace_back(util::FloatLongitude{}, util::FloatLatitude{});
 
     json::Object result;
-
     const auto rc = osrm.Route(params, result);
+    BOOST_CHECK(rc == Status::Ok);
```
TheMarex: Try to do these checks fixture based. That means you need:

1. Code that just creates the full expected response as util::json::Object (hard coded)
2. Code that validates if the response and the given object are the same

Hard coding in tests is exactly what you want. There should be as little actual code as possible to minimize surprises. Tests need to be as stupid as possible.
I feel a lot of the tests that I wrote in the past that included extensive clever setup code are pretty bad and break often.

daniel-j-h:
Should now compile again after adapting the code to the changes pulled in by the rebase.

[Hint's operator==](https://github.com/Project-OSRM/osrm-backend/pull/2088/commits/62060a2c95d70f54d3e14995fa98b8753106c619#diff-0b15ceebf2a721738feb576a870a35cbL53) should be fixed on the original branch, after looking over the changes again.

include/util/viewport.hpp
```diff
@@ -38,7 +38,10 @@ unsigned getFittedZoom(util::Coordinate south_west, util::Coordinate north_east)
     const double height_ratio = (max_y - min_y) / detail::VIEWPORT_HEIGHT;
     const auto zoom = detail::MAX_ZOOM - std::max(std::log(width_ratio), std::log(height_ratio)) * detail::INV_LOG_2;
 
-    return std::max<unsigned>(detail::MIN_ZOOM, zoom);
+    if (std::isfinite(zoom))
```
TheMarex: I feel we should catch this earlier in the code above. There needs to be a check if `height_ratio` and `width_ratio` are bigger than 0. In that case `MAX_ZOOM` (18) needs to be returned.

include/util/coordinate.hpp
```diff
@@ -50,22 +52,30 @@ OSRM_STRONG_TYPEDEF(double, FloatLongitude)
...
+    const auto longitude = static_cast<double>(floating);
+    const auto fixed = boost::numeric_cast<std::int32_t>(longitude * COORDINATE_PRECISION);
+    return FixedLongitude(fixed);
 }
 
 inline FloatLatitude toFloating(const FixedLatitude fixed)
 {
-    return FloatLatitude(static_cast<int32_t>(fixed) / COORDINATE_PRECISION);
+    const auto latitude = static_cast<std::int32_t>(fixed);
+    const auto floating = boost::numeric_cast<double>(latitude * COORDINATE_PRECISION);
```
daniel-j-h: Should be `/`

include/util/coordinate.hpp
```diff
@@ -50,22 +52,30 @@ OSRM_STRONG_TYPEDEF(double, FloatLongitude)
...
+    return FloatLatitude(floating);
 }
 
 inline FloatLongitude toFloating(const FixedLongitude fixed)
 {
-    return FloatLongitude(static_cast<int32_t>(fixed) / COORDINATE_PRECISION);
+    const auto longitude = static_cast<std::int32_t>(fixed);
+    const auto floating = boost::numeric_cast<double>(longitude * COORDINATE_PRECISION);
```
daniel-j-h: Should be `/` - fsck c&p!

include/util/viewport.hpp
daniel-j-h: @TheMarex we're using `std::log` on both `width_ratio` and `height_ratio`. For `log(x)`:

- x < 0 => NaN
- x = 0 => -inf

http://en.cppreference.com/w/cpp/numeric/math/log

Care to elaborate what you want to check instead?

TheMarex:
:warning: rebased on `rewrite/new-api` again and broke out the base 64 fixes

include/engine/base64.hpp
```diff
@@ -16,12 +17,26 @@
 
 // RFC 4648 "The Base16, Base32, and Base64 Data Encodings"
 // See: https://tools.ietf.org/html/rfc4648
-// Implementation adapted from: http://stackoverflow.com/a/28471421
```
daniel-j-h: The tips here on how to use the Boost serialization iterators are wrong and dangerous! Padding needs to happen _before_ iterating over stream!

include/engine/base64.hpp
```diff
@@ -16,12 +17,26 @@
...
+                                               6,            // get view of 6 bit
+                                               8             // from sequence of 8 bit
+                                               >>;
+
+using BinaryFromBase64 = boost::archive::iterators::transform_width<
+    boost::archive::iterators::binary_from_base64<std::string::const_iterator>,
+    8, // get a view of 8 bit
+    6  // from a sequence of 6 bit
+    >;
+} // ns detail
```
daniel-j-h: these two transform_widths require the stream length to be a multiple of 24bits / 3 bytes, that's why we pad it _before_ using them for iteration!

include/engine/base64.hpp
```diff
@@ -32,18 +47,24 @@ namespace engine
...
-    using namespace boost::archive::iterators;
+    std::vector<unsigned char> bytes{first, first + size};
+    BOOST_ASSERT(!bytes.empty());
 
-    const std::string bytes{first, first + size};
+    std::size_t bytes_to_pad{0};
 
-    using Iter = base64_from_binary<transform_width<std::string::const_iterator, 6, 8>>;
+    while (bytes.size() % 3 != 0)
+    {
```
MoKob: I would prefer some note on why we use 3 here, since the connection to the 6/8 part is not obvious.

TheMarex:
Merging this back into `rewrite/new-api` to circumvent branch sprawl.

PR #1921, opened 2016-01-22
[suggestion] performance fix for map-matching on core chs

MoKob:
Introduces a pruning weight for the `get_network_distance` function.
This addresses time-outs for map-matching on a core-ch with a large core.

PR #2090, opened 2016-03-14
Fixes the core search stopping criterion

MoKob:
Adds the fix that got left out on the rewrite of https://github.com/Project-OSRM/osrm-backend/pull/1921 in https://github.com/Project-OSRM/osrm-backend/commit/adc3bacea960b7bdd8d01f24c21454a9c0f1fff5.

When the search reaches the core, there is no notion of upward/downward anymore.
If one of the heaps runs empty, it is equivalent to no distance existing.

`HUGE-CORE-GRAPH - s  ||  t` (s connected to the huge graph, t in a small component)

The current search continues, if no pruning value is specified, until the huge core graph has been explored. This can result in slow queries for non-map-matching queries.

Since the core search could be performed unidirectional, we can modify the stopping criterion as in this pull request.

### DIFF
--- a/include/engine/routing_algorithms/routing_base.hpp
+++ b/include/engine/routing_algorithms/routing_base.hpp
@@ -501,10 +501,10 @@ template <class DataFacadeT, class Derived> class BasicRoutingInterface
                 std::vector<NodeID> &packed_leg,
                 const bool force_loop_forward,
                 const bool force_loop_reverse,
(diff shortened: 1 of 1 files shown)


TheMarex:
Good catch! :+1: Did you do any measurements of how much this increases the speed?

Ready to merge from my side.

MoKob:
It should, on average, not affect the speed at all. It just catches bad cases were no route exists quicker. So our worst case is not going to suffer as much

PR #2093, opened 2016-03-14
Bugfix: set initial uturn indicators

arnekaiser:
Viaroute requests in the following style will crash the server, because the uturn indicators are not set.

Example: /viaroute?locs=mqb|H}}rm@ofd@lz{@?? (Note: The geometry has a coordinate precision of five, but this will also crash a server with a coordinate precision of six)

This pull request will fix this issue. 

Question: Should "is_source" and "is_destination" also be set in "SetCoordinatesFromGeometry"?


TheMarex:
:+1: LGTM. Note this will be obsolete once #1935 is merged.

daniel-j-h:
Related: https://github.com/Project-OSRM/osrm-backend/pull/2041

PR #2098, opened 2016-03-15
Support multiple traffic update files, and include filename in debug tile metadata

danpat:
This change adds support for `--segment-speed-file file1 --segment-speed-file file2` to `osrm-contract`.

It also saves the filenames, and remembers which file was the source of the data used to update each edge weight.

If the `--segment-speed-file` option is not used, there is no overhead to this feature.  When it is used, there is an extra 0.25 * (size of compressed geometries).

Two new files are written out when this feature is used:

  `.osrm.datasource_indexes`  <- these link the compressed geometry segments to datasource name that sourced their data
  `.osrm.datasource_names`  <- this is the list of datasource names used, in order of priority (least to most)

### DIFF
--- a/include/engine/datafacade/shared_datafacade.hpp
+++ b/include/engine/datafacade/shared_datafacade.hpp
@@ -80,6 +80,11 @@ class SharedDataFacade final : public BaseDataFacade
     util::ShM<unsigned, true>::vector m_geometry_indices;
     util::ShM<extractor::CompressedEdgeContainer::CompressedEdge, true>::vector m_geometry_list;
     util::ShM<bool, true>::vector m_is_core_node;
+    util::ShM<uint8_t, true>::vector m_datasource_list;
+
+    util::ShM<char, true>::vector m_datasource_name_data;
+    util::ShM<std::size_t, true>::vector m_datasource_name_offsets;
+    util::ShM<std::size_t, true>::vector m_datasource_name_lengths;
 
     boost::thread_specific_ptr<std::pair<unsigned, std::shared_ptr<SharedRTree>>> m_static_rtree;
     boost::thread_specific_ptr<SharedGeospatialQuery> m_geospatial_query;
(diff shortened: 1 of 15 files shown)


TheMarex:

include/engine/datafacade/shared_datafacade.hpp
```diff
@@ -632,6 +669,52 @@ class SharedDataFacade final : public BaseDataFacade
...
+    {
+        std::string result;
+
+        if (m_datasource_name_offsets.empty() || datasource_name_id > m_datasource_name_offsets.size())
+        {
+            if (datasource_name_id == 0) return "lua profile";
+            return "UNKNOWN";
+        }
+
+        std::copy(m_datasource_name_data.begin() + m_datasource_name_offsets[datasource_name_id],
```
TheMarex: Apply `clang-format`.

src/storage/storage.cpp
```diff
@@ -325,6 +341,38 @@ int Storage::Run()
...
+    {
+        std::string name;
+        while (std::getline(datasource_names_input_stream, name))
+        {
+            m_datasource_name_offsets.push_back(m_datasource_name_data.size());
+            std::copy(name.c_str(), name.c_str() + name.size(), std::back_inserter(m_datasource_name_data));
+            m_datasource_name_lengths.push_back(name.size());
+        }
+
+        shared_layout_ptr->SetBlockSize<char>(SharedDataLayout::DATASOURCE_NAME_DATA, m_datasource_name_data.size());
```
TheMarex: Needs formating

src/storage/storage.cpp
```diff
@@ -325,6 +341,38 @@ int Storage::Run()
...
+        size_t number_of_compressed_datasources = 0;
+        geometry_datasource_input_stream.read(reinterpret_cast<char *>(&number_of_compressed_datasources), sizeof(size_t));
+        shared_layout_ptr->SetBlockSize<uint8_t>(SharedDataLayout::DATASOURCES_LIST, number_of_compressed_datasources);
+    }
+
+    // Load datasource name sizes.  This file is optional, and it's non-fatal if it doesn't
+    // exist
+    std::ifstream datasource_names_input_stream(datasource_names_path.c_str(), std::ios::binary);
+    std::vector<char> m_datasource_name_data;
+    std::vector<size_t> m_datasource_name_offsets;
```
TheMarex: `size_t` -> `std::size_t`

PR #2099, opened 2016-03-16
Adapts routed to always print port and ip it listens on, fixed #2097

### LINKED ISSUE #2097: Print host and port on which osrm-routed listen
After osrm-routed boostrap add host and port:

Replace:
```
[info] running and waiting for requests
```

By:
```
[info] running and waiting for requests on http://0.0.0.0:5000
```

Explanation: system can choose an available port by using the value '0' like this
```
./build/osrm-routed --port 0 andorra-latest.osrm
```

daniel-j-h:


TheMarex:
Looks good to merge. :+1:

Issue #2101, opened 2016-03-16
U-Turn assertion triggered in osrm-extract


TheMarex:
When running #1935 on an data extract from the bay area the following assertion is triggered:

```
#6  0x00000000007a6a7e in osrm::extractor::EdgeBasedGraphFactory::Run (this=this@entry=0x7fffffffad50, 
    original_edge_data_filename="/home/patrick/Code/osrm-data/bay_area_2015-08-14-2.osrm.edges", lua_state=lua_state@entry=0xabbb88, 
    edge_segment_lookup_filename="/home/patrick/Code/osrm-data/bay_area_2015-08-14-2.osrm.edge_segment_lookup", 
    edge_penalty_filename="/home/patrick/Code/osrm-data/bay_area_2015-08-14-2.osrm.edge_penalties", generate_edge_lookup=<optimized out>)
    at /home/patrick/Code/osrm-temp/src/extractor/edge_based_graph_factory.cpp:178
#7  0x000000000072c965 in osrm::extractor::Extractor::BuildEdgeExpandedGraph (this=this@entry=0x7fffffffe1d0, 
    internal_to_external_node_map=std::vector of length 892232, capacity 1048576 = {...}, node_based_edge_list=std::vector of length 0, capacity 0, 
    node_is_startpoint=std::vector<bool> of length 0, capacity 0, edge_based_node_weights=std::vector of length 0, capacity 0, edge_based_edge_list=...)
    at /home/patrick/Code/osrm-temp/src/extractor/extractor.cpp:536
#8  0x000000000072df9e in osrm::extractor::Extractor::run (this=this@entry=0x7fffffffe1d0) at /home/patrick/Code/osrm-temp/src/extractor/extractor.cpp:272
#9  0x00000000006e6d2b in main (argc=4, argv=0x7fffffffe4e8) at /home/patrick/Code/osrm-temp/src/tools/extract.cpp:147
(gdb) f 3
#3  0x00000000007cdd4d in osrm::extractor::guidance::TurnAnalysis::handleTwoWayTurn (this=this@entry=0x7fffffffa320, from=from@entry=686395, via_edge=via_edge@entry=1461569, 
    turn_candidates=std::vector of length 2, capacity 4 = {...}) at /home/patrick/Code/osrm-temp/src/extractor/guidance/turn_analysis.cpp:843
843     BOOST_ASSERT(turn_candidates[0].angle < 0.001);
(gdb) p turn_candidates 
[...]

PR #2102, opened 2016-03-16
fix merging of turn instructions

MoKob:
Fix the currently broken merge of segregated roads

adresses https://github.com/Project-OSRM/osrm-backend/issues/2101

### DIFF
--- a/src/extractor/guidance/turn_analysis.cpp
+++ b/src/extractor/guidance/turn_analysis.cpp
@@ -1679,7 +1679,7 @@ std::vector<TurnCandidate> TurnAnalysis::getTurnCandidates(const NodeID from_nod
 
     for (const EdgeID onto_edge : node_based_graph.GetAdjacentEdgeRange(turn_node))
     {
-
+        BOOST_ASSERT( onto_edge != SPECIAL_EDGEID );
         const NodeID to_node = node_based_graph.GetTarget(onto_edge);
 
         bool turn_is_valid =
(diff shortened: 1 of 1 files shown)


src/extractor/guidance/turn_analysis.cpp
```diff
@@ -1846,10 +1848,12 @@ std::vector<TurnCandidate> TurnAnalysis::mergeSegregatedRoads(
...
         for (std::size_t i = 2; i < turn_candidates.size(); ++i)
             turn_candidates[i].angle += correction_factor;
-        turn_candidates[1].angle = 0;
+        turn_candidates[0] = merge(turn_candidates[0],turn_candidates[1]);
+        turn_candidates[0].angle = 0;
+        turn_candidates.erase(turn_candidates.begin() + 1);
     }
 
-    for (std::size_t index = 0; index < turn_candidates.size(); ++index)
+    for (std::size_t index = 2; index < turn_candidates.size(); ++index)
```
danpat: Can this be a descriptive constant instead of `2` ?

MoKob: what would you have in mind? the first two indices are handled prior. just don't merge anything including 0

danpat: Maybe just a comment on the loop as to why it starts at "2".  I can see it now reading the code, but it'd be nice if it was a bit more obvious.

PR #2106, opened 2016-03-17
Fixes the edge-based-graph factory's edge counter serialization.

daniel-j-h:
The counter for original edges is of type `std::size_t`, but we
serialized `sizeof(unsigned)` number of bytes out to the `.osrm.edges`
file.

We should probably check all writes (analogous for reads) and make the
count parameter dependent on `sizeof(variable)`.

    ag '\.write\((.*), sizeof\((.*)\)\);'

References:
- Moritz's original report at https://gist.github.com/MoKob/76287bafd3afcf32a947

### DIFF
--- a/src/extractor/edge_based_graph_factory.cpp
+++ b/src/extractor/edge_based_graph_factory.cpp
@@ -10,6 +10,7 @@
 #include "util/exception.hpp"
 
 #include <boost/assert.hpp>
+#include <boost/numeric/conversion/cast.hpp>
 
 #include <algorithm>
 #include <cmath>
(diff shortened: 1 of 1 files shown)


danpat:
If we're going to write `size_t` instead of `unsigned`, then we also need to update the locations where this data is read, otherwise we're going to end up with off-by-N errors reading data.

There are at least two:

https://github.com/Project-OSRM/osrm-backend/blob/rewrite/new-api/src/storage/storage.cpp#L218-L219

https://github.com/Project-OSRM/osrm-backend/blob/rewrite/new-api/include/engine/datafacade/internal_datafacade.hpp#L147-L148

and there might be more.....

daniel-j-h:
You're correct! Do we need `std::size` for the edge counter anyway?

danpat:
Probably not, as we don't really support edge or node IDs > `std::numeric_limits<unsigned>::max()`

Let's go back to writing an `unsigned` to the file, cast the value properly, and put in an assert to make sure that the `size_t` doesn't overflow `unsigned`.

daniel-j-h:
Last commit does that.

TheMarex:
@daniel-j-h :+1: can be merged into `develop`.

PR #2107, opened 2016-03-17
fix negative weight

MoKob:
Fixes the negative weight that can occur in current routing request.

I've found the reason for the problem. Its a combination of an error made by me based on a comment of @TheMarex in `assemble_leg.hpp` and a variation in the paradigm assumed by @TheMarex due to the fixes for the offsets in the R-Tree done by @danpat .

To try and explain the situation:

Assemble Leg assumes positive weights to be located at only certain elements (https://github.com/Project-OSRM/osrm-backend/blob/fix/negative-cost/include/engine/guidance/assemble_leg.hpp#L141-L142) . Due to the offsets by @danpat , this assumption is broken. I tried to keep it true and deleted the travel-time when no turn instruction was issued. This is fixed [here](https://github.com/Project-OSRM/osrm-backend/pull/2107/files#diff-99adcb8bccf8e94a176bdf92509ff070).

The other problem is that the new offsets are corrected for the source-phantom already. So if the duration of the path crosses an intersection, subtracting the source_duration ([see here](https://github.com/Project-OSRM/osrm-backend/blob/fix/negative-cost/include/engine/guidance/assemble_leg.hpp#L150)) is no longer valid.

This pull request offers a way of dealing with it, I am unsure whether we should rather correctly handle the offset weights, or if other parts of the code are affected of this as well.

### DIFF
--- a/src/engine/guidance/post_processing.cpp
+++ b/src/engine/guidance/post_processing.cpp
@@ -198,6 +198,7 @@ std::vector<std::vector<PathData>> postProcess(std::vector<std::vector<PathData>
     // silence silent turns for good
     for (auto &path_data : leg_data)
     {
+        unsigned last = path_data.back().duration_until_turn;
         for (auto &data : path_data)
         {
             if (isSilent(data.turn_instruction) || (leavesRoundabout(data.turn_instruction) &&
@@ -208,6 +209,7 @@ std::vector<std::vector<PathData>> postProcess(std::vector<std::vector<PathData>
                 data.exit = 0;
             }
         }
+        path_data.back().duration_until_turn = last;
     }
 
     return leg_data;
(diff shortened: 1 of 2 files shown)


src/engine/guidance/post_processing.cpp
```diff
@@ -198,6 +198,7 @@ std::vector<std::vector<PathData>> postProcess(std::vector<std::vector<PathData>
     // silence silent turns for good
     for (auto &path_data : leg_data)
```
TheMarex: Can you explain to me what you do here?

The fix should be not set `data.duration_until_turn = 0;` Or merge the duration with the previous segment.

MoKob: https://github.com/Project-OSRM/osrm-backend/blob/rewrite/new-api/include/engine/guidance/assemble_leg.hpp#L141-L142

It removes the time value from all non-turn items

TheMarex:
Addressed by https://github.com/Project-OSRM/osrm-backend/pull/1935/commits/9b75ff98c9f7372d1900de197fa5616d30ba8324 and following commits.

PR #2116, opened 2016-03-18
Move PostProcessing onto Route Steps

MoKob:
This pull request tracks the progress of moving the Guidance Post-Processing options onto the Route Step.

It consists of the following task:
 ~~-[]Adjust turn collapsing in [assemble_leg.hpp](https://github.com/Project-OSRM/osrm-backend/blob/rewrite/new-api/include/engine/guidance/assemble_leg.hpp#L36-L54): This process assumes name-ids to be the only decision that decides whether a turn is kept or not. This needs to be adjusted to compress only no-turns. Some instructions are actually required to stay on a road.~~
 Since this is only a heuristic for the route name and would require a lot of changes, keeping the name as only source of naming the route seems like the better option.

- [x] Move Post Processing away from `PathData` and past `assemble_steps.cpp` to work on route-steps. This would require the insertion of temporary route steps for silent instructions that can then be collapsed, or a full integration of the post-processing with assemble_steps. I'd prefer the former (@themarex, whats your take on this?)

- [x] Post-Process Temporary Route Steps: Due to the adjustment above, some silent instructions might remain in the route-steps. Post-processing has to combine these accordingly.
     - [x] we have to count exits for roundabouts
     - [x] In addition, we can add information on intersections in between the instructions. Every silent turn can be added to a list of intersections with respective distances along the current step geometry (thank you for the proposal @TheMarex)

- [x] clean up silent instructions -- remove silent instructions from output to prevent them from being shown in the json result.
    - [x] remove silent instructions from output
    - [x] potentially sync leg-geometry with new instructions?

- [x] https://github.com/Project-OSRM/osrm-backend/pull/2080 needs to be merged prior to this

- [x] find out why path length and segments do not match 

- [x] get @daniel-j-h 's blessing

### DIFF
--- a/include/engine/guidance/assemble_steps.hpp
+++ b/include/engine/guidance/assemble_steps.hpp
@@ -27,8 +27,7 @@ namespace detail
 StepManeuver stepManeuverFromGeometry(extractor::guidance::TurnInstruction instruction,
                                       const WaypointType waypoint_type,
                                       const LegGeometry &leg_geometry,
-                                      const std::size_t segment_index,
-                                      const unsigned exit);
+                                      const std::size_t segment_index);
 } // ns detail
 
 template <typename DataFacadeT>
@@ -78,7 +77,7 @@ std::vector<RouteStep> assembleSteps(const DataFacadeT &facade,
         StepManeuver maneuver = detail::stepManeuverFromGeometry(
             extractor::guidance::TurnInstruction{extractor::guidance::TurnType::NoTurn,
                                                  initial_modifier},
-            WaypointType::Depart, leg_geometry, segment_index, INVALID_EXIT_NR);
+            WaypointType::Depart, leg_geometry, segment_index);
 
         // PathData saves the information we need of the segment _before_ the turn,
         // but a RouteStep is with regard to the segment after the turn.
@@ -89,7 +88,7 @@ std::vector<RouteStep> assembleSteps(const DataFacadeT &facade,
         {
             segment_duration += path_point.duration_until_turn;
 
-            if (path_point.turn_instruction != extractor::guidance::TurnInstruction::NO_TURN())
+            if (path_point.turn_instruction.type != extractor::guidance::TurnType::NoTurn)
             {
                 BOOST_ASSERT(segment_duration >= 0);
                 const auto name = facade.GetNameForID(path_point.name_id);
@@ -104,7 +103,7 @@ std::vector<RouteStep> assembleSteps(const DataFacadeT &facade,
                                           leg_geometry.BackIndex(segment_index) + 1});
                 maneuver = detail::stepManeuverFromGeometry(path_point.turn_instruction,
                                                             WaypointType::None, leg_geometry,
-                                                            segment_index, path_point.exit);
+                                                            segment_index);
                 segment_index++;
                 segment_duration = 0;
             }
@@ -136,6 +135,7 @@ std::vector<RouteStep> assembleSteps(const DataFacadeT &facade,
                                  extractor::guidance::TurnInstruction{
                                      extractor::guidance::TurnType::NoTurn, initial_modifier},
                                  WaypointType::Depart,
+                                 INVALID_EXIT_NR,
                                  INVALID_EXIT_NR};
         int duration = target_duration - source_duration;
         BOOST_ASSERT(duration >= 0);
@@ -174,6 +174,7 @@ std::vector<RouteStep> assembleSteps(const DataFacadeT &facade,
                                extractor::guidance::TurnInstruction{
                                    extractor::guidance::TurnType::NoTurn, final_modifier},
                                WaypointType::Arrive,
+                               INVALID_EXIT_NR,
                                INVALID_EXIT_NR},
                   leg_geometry.locations.size(),
                   leg_geometry.locations.size()});
--- a/include/engine/guidance/assemble_steps.hpp
+++ b/include/engine/guidance/assemble_steps.hpp
@@ -25,9 +25,13 @@ namespace guidance
 namespace detail
 {
 StepManeuver stepManeuverFromGeometry(extractor::guidance::TurnInstruction instruction,
-                                      const WaypointType waypoint_type,
                                       const LegGeometry &leg_geometry,
                                       const std::size_t segment_index);
+
+StepManeuver stepManeuverFromGeometry(extractor::guidance::TurnInstruction instruction,
+                                      const WaypointType waypoint_type,
+                                      const LegGeometry &leg_geometry);
+
 } // ns detail
 
 template <typename DataFacadeT>
@@ -39,7 +43,7 @@ std::vector<RouteStep> assembleSteps(const DataFacadeT &facade,
                                      const bool source_traversed_in_reverse,
                                      const bool target_traversed_in_reverse)
 {
-    const double constexpr ZERO_DURACTION = 0., ZERO_DISTANCE = 0., NO_BEARING = 0.;
+    const double constexpr ZERO_DURACTION = 0., ZERO_DISTANCE = 0.;
     const EdgeWeight source_duration =
         source_traversed_in_reverse ? source_node.reverse_weight : source_node.forward_weight;
     const auto source_mode = source_traversed_in_reverse ? source_node.backward_travel_mode
@@ -67,8 +71,7 @@ std::vector<RouteStep> assembleSteps(const DataFacadeT &facade,
         distance_to_start >= MINIMAL_RELATIVE_DISTANCE &&
                 distance_to_start <= MAXIMAL_RELATIVE_DISTANCE
             ? angleToDirectionModifier(util::coordinate_calculation::computeAngle(
-                  source_node.input_location, leg_geometry.locations[0],
-                  leg_geometry.locations[1]))
+                  source_node.input_location, leg_geometry.locations[0], leg_geometry.locations[1]))
             : extractor::guidance::DirectionModifier::UTurn;
 
     if (leg_data.size() > 0)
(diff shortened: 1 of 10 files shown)


MoKob:
@TheMarex I have a question regarding the leg_geometry.

After https://github.com/Project-OSRM/osrm-backend/blob/fix/guidance/postprocessing/src/engine/guidance/post_processing.cpp#L232-L241, which removes temporary instructions in post-processing to count roundabouts/intersections, the leg-geometry segments are out of sync with the actual turn-instructions.

Should we undertake the effort to set the correct indices again? Essentially translating the turn-instruction geometry indices back into the segments of the geometry? The coordinates itself should be unaffected.

MoKob:
An additional item to discuss: to output the intersections in between, should we output the duration/distance along the currently looked at segment, or the distances/durations between the different intersections themselves?

So for (S) --a -- (I) -- b -- (I2) -- c -- (T), should we rather output:

Option 1: `Depart S, (a+b+c), Intersections: [(I,a), (I2,a+b)]` 

or

Option 2: `Depart S, (a+b+c), Intersections: [(I,a), (I2,b)]`

In both cases, we would end up adding some additional information down the line on the intersection shape.

include/engine/guidance/assemble_steps.hpp
```diff
@@ -25,10 +25,13 @@ namespace guidance
 namespace detail
 {
 StepManeuver stepManeuverFromGeometry(extractor::guidance::TurnInstruction instruction,
-                                      const WaypointType waypoint_type,
                                       const LegGeometry &leg_geometry,
-                                      const std::size_t segment_index,
-                                      const unsigned exit);
+                                      const std::size_t segment_index);
```
daniel-j-h: From what I can see I _think_ you only need the instruction to pass it on (and you do not move it), so a const-ref (or const value since it should be cheap to copy it) should be fine, too.

include/engine/guidance/assemble_steps.hpp
```diff
@@ -25,10 +25,13 @@ namespace guidance
...
+
+StepManeuver stepManeuverFromGeometry(extractor::guidance::TurnInstruction instruction,
```
daniel-j-h: Same ^

include/engine/guidance/assemble_steps.hpp
```diff
@@ -40,7 +43,7 @@ std::vector<RouteStep> assembleSteps(const DataFacadeT &facade,
                                      const bool source_traversed_in_reverse,
                                      const bool target_traversed_in_reverse)
 {
-    const double constexpr ZERO_DURACTION = 0., ZERO_DISTANCE = 0., NO_BEARING = 0.;
+    const double constexpr ZERO_DURACTION = 0., ZERO_DISTANCE = 0.;
```
daniel-j-h: -C

include/engine/guidance/assemble_steps.hpp
```diff
@@ -162,21 +162,17 @@ std::vector<RouteStep> assembleSteps(const DataFacadeT &facade,
...
-                               extractor::guidance::TurnInstruction{
-                                   extractor::guidance::TurnType::NoTurn, final_modifier},
-                               WaypointType::Arrive,
-                               INVALID_EXIT_NR},
-                  leg_geometry.locations.size(),
-                  leg_geometry.locations.size()});
+    auto final_maneuver = detail::stepManeuverFromGeometry(
+        extractor::guidance::TurnInstruction{extractor::guidance::TurnType::NoTurn, final_modifier},
+        WaypointType::Arrive, leg_geometry);
+    steps.push_back(RouteStep{target_node.name_id,
```
daniel-j-h: If you .emplace_back here you don't have to specify the RouteStep type again

src/engine/api/json_factory.cpp
```diff
@@ -149,6 +149,12 @@ util::json::Object makeStepManeuver(const guidance::StepManeuver &maneuver)
     step_maneuver.values["bearing_after"] = maneuver.bearing_after;
     if (maneuver.exit != 0)
         step_maneuver.values["exit"] = maneuver.exit;
+
+    //TODO currently we need this to comply with the api.
+    //We should move this to an additional entry, the moment we
+    //actually compute the correct locations of the intersections
+    if (!maneuver.intersections.empty() && maneuver.exit == 0 )
```
daniel-j-h: format `== 0 )`

src/engine/guidance/assemble_steps.cpp
```diff
@@ -12,15 +12,49 @@ namespace guidance
...
+            util::coordinate_calculation::bearing(pre_turn_coordinate, turn_coordinate);
+    }
+    return StepManeuver{
+        turn_coordinate,
+        pre_turn_bearing,
+        post_turn_bearing,
+        instruction,
+        waypoint_type,
+        INVALID_EXIT_NR,
+        {} // no intermediate intersections
```
daniel-j-h: We leave the function here, maybe move in the turn coordinate?

src/engine/guidance/assemble_steps.cpp
```diff
@@ -32,7 +66,15 @@ StepManeuver stepManeuverFromGeometry(extractor::guidance::TurnInstruction instr
     const double post_turn_bearing =
         util::coordinate_calculation::bearing(turn_coordinate, post_turn_coordinate);
 
-    return StepManeuver{turn_coordinate, pre_turn_bearing, post_turn_bearing, instruction, waypoint_type, exit};
+    return StepManeuver{
+        turn_coordinate,
```
daniel-j-h: same ^

src/engine/guidance/post_processing.cpp
```diff
@@ -20,194 +22,248 @@ namespace guidance
...
+        [](RouteStep into, const RouteStep &last_step, const RouteStep &intersection)
     {
-        if (not path_data.empty())
-            path_data[0].exit = carry_exit;
+        into.maneuver.intersections.push_back(
+            {last_step.duration, last_step.distance, intersection.maneuver.location});
 
-        for (std::size_t data_index = 0; data_index + 1 < path_data.size(); ++data_index)
+        into = detail::forwardInto(std::move(into), intersection);
+        return into;
```
daniel-j-h: merge: return detail::fwdInto()

include/engine/guidance/post_processing.hpp
```diff
@@ -1,7 +1,7 @@
 #ifndef ENGINE_GUIDANCE_POST_PROCESSING_HPP
 #define ENGINE_GUIDANCE_POST_PROCESSING_HPP
 
-#include "engine/internal_route_result.hpp"
+#include "engine/guidance/route_step.hpp"
```
daniel-j-h: missing: utility, cstddef

src/engine/guidance/post_processing.cpp
```diff
@@ -4,6 +4,8 @@
 #include "engine/guidance/toolkit.hpp"
 
 #include <boost/assert.hpp>
+#include <boost/range/algorithm_ext/erase.hpp>
+
```
daniel-j-h: missing: utility, cstddef
also: vector is already included in the header (interface requires vectors as in/out types)

src/engine/guidance/post_processing.cpp
```diff
@@ -20,194 +22,248 @@ namespace guidance
...
 
-std::vector<std::vector<PathData>> postProcess(std::vector<std::vector<PathData>> leg_data)
+// Every Step Maneuver consists of the information until the turn.
+// This list contains a set of instructions, called silent, which should
+// not be part of the final output.
+// They are required for maintenance purposes. We can calculate the number
+// of exits to pass in a roundabout and the number of intersections
+// that we come across.
+
+std::vector<RouteStep> postProcess(std::vector<RouteStep> steps)
```
daniel-j-h: This is a 200 loc function and is somewhat hard to follow --- I would love if we could refactor it a bit

src/engine/guidance/post_processing.cpp
```diff
@@ -20,194 +22,248 @@ namespace guidance
...
+    // Two valid NO_TURNs exist in each leg in the form of Depart/Arrive
+
+    // keep valid instructions
+    const auto not_is_valid = [](const RouteStep &step)
+    {
+        return step.maneuver.instruction == TurnInstruction::NO_TURN() &&
+               step.maneuver.waypoint_type == WaypointType::None;
+    };
+
+    boost::remove_erase_if(steps, not_is_valid);
```
daniel-j-h: shiny! the stdlib should have things like this by default! :D

src/engine/guidance/post_processing.cpp
```diff
@@ -20,194 +22,248 @@ namespace guidance
...
+
+#if PRINT_DEBUG
```
daniel-j-h: Maybe prefix with `OSRM_`, to avoid potential clashes

include/engine/guidance/assemble_steps.hpp
MoKob: ??

daniel-j-h: should be duration instead of dura**c**tion ? :P

MoKob: sure. looked like you added a random comment from a console command, though. like `make -C` ...

MoKob:
Might be related: https://github.com/Project-OSRM/osrm-backend/issues/1575

MoKob:
Path length in front-end are rounded to large differences (426->450). This results in the differences in the path data length I've seen. The values seem correct, though.

include/engine/guidance/assemble_steps.hpp
daniel-j-h: For the record, this does not work. We don't know why ¯\\\_(ツ)\_/¯

src/engine/api/json_factory.cpp
daniel-j-h: seems to be still in there

MoKob:
@TheMarex this one is all yours

TheMarex:
@MoKob looks great! Going to rebase and merge.

PR #2118, opened 2016-03-18
Ready to merge: make Windows build again

BergWerkGIS:


daniel-j-h:
Looking at Appveyor it seems like we have at least two additional problems:

```
C:\projects\osrm\src\extractor\edge_based_graph_factory.cpp(140): error C2672: 'osrm::util::irange': no matching overloaded function found [C:\projects\osrm\build\EXTRACTOR.vcxproj]
C:\projects\osrm\src\extractor\edge_based_graph_factory.cpp(140): error C2782: 'osrm::util::range<Integer> osrm::util::irange(const Integer,const Integer,std::enable_if<std::is_integral<_Ty>::value,void>::type *) noexcept': template parameter 'Integer' is ambiguous [C:\projects\osrm\build\EXTRACTOR.vcxproj]
  C:\projects\osrm\include\util/integer_range.hpp(49): note: see declaration of 'osrm::util::irange'
  C:\projects\osrm\src\extractor\edge_based_graph_factory.cpp(140): note: could be 'unsigned __int64'
  C:\projects\osrm\src\extractor\edge_based_graph_factory.cpp(140): note: or       'unsigned long'
C:\projects\osrm\src\extractor\edge_based_graph_factory.cpp(140): error C2784: 'osrm::util::range<Integer> osrm::util::irange(const Integer,const Integer,std::enable_if<std::is_integral<_Ty>::value,void>::type *) noexcept': could not deduce template argument for 'const Integer' from 'const unsigned __int64' [C:\projects\osrm\build\EXTRACTOR.vcxproj]
  C:\projects\osrm\include\util/integer_range.hpp(49): note: see declaration of 'osrm::util::irange'
C:\projects\osrm\src\extractor\edge_based_graph_factory.cpp(146): error C2065: 'i': undeclared identifier [C:\projects\osrm\build\EXTRACTOR.vcxproj]
C:\projects\osrm\src\extractor\edge_based_graph_factory.cpp(146): error C2228: left of '.node_id' must have class/struct/union [C:\projects\osrm\build\EXTRACTOR.vcxproj]
C:\projects\osrm\src\extractor\edge_based_graph_factory.cpp(146): error C2789: 'current_edge_target_coordinate_id': an object of const-qualified type must be initialized [C:\projects\osrm\build\EXTRACTOR.vcxproj]
  C:\projects\osrm\src\extractor\edge_based_graph_factory.cpp(146): note: see declaration of 'current_edge_target_coordinate_id'
C:\projects\osrm\src\extractor\edge_based_graph_factory.cpp(154): error C2065: 'i': undeclared identifier 
```

and the linker not being able to find the Boost unit test framework library:

```
LINK : fatal error LNK1104: cannot open file 'boost_unit_test_framework-vc140-mt-1_59.lib' [C:\projects\osrm\build\util-tests.vcxproj]
```

daniel-j-h:
For the first issue: looks like the zero [here](https://github.com/Project-OSRM/osrm-backend/blob/windows-build/src/extractor/edge_based_graph_factory.cpp#L140) has to be of the same type as `geometry_size`. I _think_ that's a `std::size_t`, so we can make it work via `irange(std::size_t{0}, geometry_size)` or via `irange(decltype(geometry_size){0}, geometry_size)`.

We should provide an `irange` overload with a single argument (implying end) as e.g. in Python, though.

BergWerkGIS:
yep, that's why it's `[WIP]!!!` :smirk: 
opened the PR just to see if AppVeyor get's the same errors as I do locally.
Problem with the boost version we are using is that it thinks that VS2015 is not yet capable of c++11/14 features and disables those. I tried to override, but to now avail till now.

BergWerkGIS:
Builds are working now for me locally, with one strange twist:

* When within Visual Studio all projects in the solution build successfully.
* When I build the very same solution on the console I get the [same error as on AppVeyor](https://ci.appveyor.com/project/DennisOSRM/osrm-backend/build/1.0.2817#L2933):
```
LINK : fatal error LNK1104: cannot open file 'boost_unit_test_framework-vc140-mt-1_59.lib'
```

BTW the name of the lib is _"**lib**boost_unit_test_framework-vc140-mt-1_59.lib"_ and not just _"boost_unit_test_framework-vc140-mt-1_59.lib"_.

I suppose Visual Studio is maybe setting some env var or #define that we don't in the build scripts/CMakeLists and that's why the library name is not resolved correctly.

BergWerkGIS:
It fails for these 3 projects:
* `engine-tests`
* `extractor-tests`
* `util-tests`

BergWerkGIS:
Still trying to find out where the reference to the wrong `boost_unit_test_framework-vc140-mt-1_59.lib` without the `lib` at the beginning is coming from.

The correct name and full path gets passed to the linker:

image

but linking still fails with the same error:

image

daniel-j-h:
I don't know about the Windows situation in specific but I read about `BOOST_TEST_DYN_LINK` and `BOOST_TEST_NO_LIB`. Setting the latter prevents auto-linking.

Could it be that it automatically links against that library? The CMakeLists.txt already contains a Windows-specific definition --- maybe that's something to look into.

daniel-j-h:
Digging a bit deeper this looks more and more suspicious for what you are seeing:

https://github.com/boostorg/test/blob/boost-1.60.0/include/boost/test/unit_test.hpp#L23-L37

(especially take note of the define BOOST_LIB_NAME boost_unit_test_framework)

and then the auto-link header

https://github.com/boostorg/config/blob/boost-1.60.0/include/boost/config/auto_link.hpp#L20-L21

(and the prefix define below)

BergWerkGIS:
Thanks, will look into the links tomorrow.
What bugs me is that it builds fine within Visual Studio, but not on the console.
In the end Visual Studio also shells out to MSBuild, like we do.

BergWerkGIS:
Tried passing `BOOST_ALL_NO_LIB` and `BOOST_TEST_NO_LIB` to circumvent `Auto Linking` via
* `cmake` commandline and
* `CMakeList.txt`

but the build is still looking for `boost_unit_test_framework-vc140-mt-1_59.lib` instead of **`lib`**`boost_unit_test_framework-vc140-mt-1_59.lib`.

----

In the meantime I've found out why the build worked within Visual Studio and not from the commandline:

The commandline builds the whole solution, whereas the default in VS builds the startup project of the solution.
Project `ALL_BUILD` is defined as startup project but it doesn't have the failing projects (`engine-tests`, `extractor-tests` & `util-tests`) as dependencies.
So it worked because it doesn't reference the failing projects.

daniel-j-h:
For a quick and ugly workaround could you try setting the prefix as described here: https://github.com/boostorg/config/blob/boost-1.60.0/include/boost/config/auto_link.hpp#L53. I'm also not sure about this:

> "lib" for static libraries otherwise "".

This implies we're not getting the `lib` prefix _by design_.

BergWerkGIS:
Waiting on AppVeyor: https://ci.appveyor.com/project/DennisOSRM/osrm-backend/build/1.0.2838

Latest commit https://github.com/Project-OSRM/osrm-backend/pull/2118/commits/526f4c265fef4e7763ccfb3b4daeeb19997f86db worked for me locally.

BergWerkGIS:
hm, AppVeyor now fails because of missing `Doxygen`:
https://ci.appveyor.com/project/DennisOSRM/osrm-backend/build/1.0.2838#L99
Didn't do so before:
https://ci.appveyor.com/project/DennisOSRM/osrm-backend/build/1.0.2791#L93

BergWerkGIS:
@daniel-j-h 

[On AppVeyor](https://ci.appveyor.com/project/DennisOSRM/osrm-backend/build/1.0.2843#L66) Cmake complains that it cannot find the static library `boost_atomic`:

<pre>
CMake Error at C:/Program Files (x86)/CMake/share/cmake-3.5/Modules/FindBoost.cmake:1647 (message):
  Unable to find the requested Boost libraries.
 
  Boost version: 1.59.0
 
  Boost include path: C:/projects/osrm/osrm-deps/boost/include/boost-1_59
 
<b>  Could not find the following static Boost libraries:
 
          boost_atomic</b>
 
  Some (but not all) of the required Boost libraries were found.  You may
  need to install these additional Boost libraries.  Alternatively, set
  BOOST_LIBRARYDIR to the directory containing Boost libraries or BOOST_ROOT
  to the location of Boost.
  CMakeLists.txt:238 (find_package)
</pre>

But the same build scripts and same binary dependency package works for me locally.
And I don't have a `boost_atomic` lib, only the headers (same as on AppVeyor):

image

Any ideas?

daniel-j-h:
We [try to find Doxygen](https://github.com/Project-OSRM/osrm-backend/blob/ca0d21eca9562b3d54568052be6be350c707d03e/CMakeLists.txt#L424-L433) but it's not required; you just don't get the `doc` target.

BergWerkGIS:
@daniel-j-h doxygen is not the problem (sorry for the confusion): `boost` is not found.

daniel-j-h:
Oh I just saw your updated comment, sorry for that. Boost.Atomic has a library part to it that we link against. It's called `libboost_atomic` (`.a` for it's static lib and `.so` for its shared lib on Unix systems).

From the screenshots I can see the headers, and this is one part of the library. The second part is the compiled library. Where do you get the Boost distribution from?

BergWerkGIS:
> Where do you get the Boost distribution from?

Building myself. And I don't have the atomic lib locally either - same scripts and same binary deps package as on AppVeyor, but locally build succeeds.

BergWerkGIS:
this is how it looks locally:

```
-- Found Osmium: C:/mb/windows-builds-64/packages/osrm-backend/third_party/libosmium/include  
-- Boost version: 1.59.0
-- Found the following Boost libraries:
--   date_time
--   filesystem
--   iostreams
--   program_options
--   regex
--   system
--   thread
--   unit_test_framework
--   date_time
--   chrono
--   zlib
-- Found Intel TBB

```

BergWerkGIS:
hm, `CMakeCache.txt` looks like this.

```
//Boost atomic library (debug)
Boost_ATOMIC_LIBRARY_DEBUG:FILEPATH=Boost_ATOMIC_LIBRARY_DEBUG-NOTFOUND

//Boost atomic library (release)
Boost_ATOMIC_LIBRARY_RELEASE:FILEPATH=Boost_ATOMIC_LIBRARY_RELEASE-NOTFOUND
```

Maybe a `CMake` version problem?
I tried with 3.1.0 and 3.5.1, AppVeyor is on 3.5.0.
Maybe I try to match that.

daniel-j-h:
We don't use Boost.Atomic ourselves, that's why we do not specify it in the CMakeLists.txt.

I can imagine the following:

- Boost.Thread requiring Boost.Atomic in case there is no C++11 `<atomic>` support. Your local Boost build probably does not build Boost.Atomic by default since you _do have_ `<atomic>` support.

- CMake thinking it knows best and links in Boost.Atomic just to be sure.

Looking at the error above, I got myself the CMake 3.5.0 sources. This is where CMake fscks up:

    cmake-3.5.0/Modules/FindBoost.cmake

Line 696, setting deps for Boost 1.59 (which you seem to use from the screenshot above):

    elseif(NOT Boost_VERSION VERSION_LESS 105900 AND Boost_VERSION VERSION_LESS 106000)
      ...
      set(_Boost_THREAD_DEPENDENCIES chrono system date_time atomic)

So for Boost 1.59, CMake 3.5.0 indeed tries to link in Boost.Atomic.

daniel-j-h:
Going down the rabbit hole, I can't find Boost.Atomic on the list of libraries you have to build on Windows:

http://www.boost.org/doc/libs/1_59_0/more/getting_started/windows.html

Can you try setting this definition for Boost.Thread and see if it helps:

http://www.boost.org/doc/libs/1_59_0/doc/html/thread/build.html#thread.build.configuration.move0

If that does not help, get yourself the Boost sources and (after bootstrapping) do:

    b2 --show-libraries

and then append the following to your build command:

    --with-atomic

(I'm just assuming here the library is called "atomic" in b2)

BergWerkGIS:
>  Your local Boost build probably does not build Boost.Atomic by default since you do have <atomic> support.

Good catch: Only difference I can think of: I'm on `Visual Studio 2015 Update 1` and AppVeyor is still on stock `Visual Studio 2015`.

Will check.

BergWerkGIS:
Not building atomic yet:
https://github.com/mapbox/windows-builds/blob/e7681c95436769384f79ed608e50b5ebb6c5772d/scripts/build_boost.bat#L100-L108

----

AppVeyor and my local machine are on the same VS 2015 Update 1 Version:

<pre>
Microsoft (R) C/C++ Optimizing Compiler Version <b>19.00.23506</b> for x64
Microsoft (R) Build Engine version <b>14.0.24720.0</b>
</pre>

BergWerkGIS:
Yes, it's a problem with CMake.
Can reproduce problem locally with `CMake@3.5.0`.

BergWerkGIS:
@daniel-j-h AppVeyor is :white_check_mark: now (pulling in `CMake@3.5.1` manually):
https://ci.appveyor.com/project/DennisOSRM/osrm-backend/build/1.0.2850#L3049

One curiosity:
`engine-tests`, `extractor-tests` and `util-tests` succeed on AppVeyor, but I get an error locally:

```
running engine-tests.exe ...
Running 3 test cases...

*** No errors detected
running extractor-tests.exe ...
Running 8 test cases...
. 10% . 20% . 30% . 40% . 50% . 60% . 70% . 80%  100%
. 10% . 20% . 30% . 40% . 50% . 60% . 70% . 80%  100%
. 10% . 20% . 30% . 40% . 50% . 60% . 70% . 100%
. 10% . 20% . 30% . 40% . 50% . 60% . 70% . 80%  100%
. 10% . 20% . 30% . 40% . 50% . 60% . 70% . 80%  100%
unknown location(0): fatal error: in "raster_source/raster_test": class osrm::util::exception: Failed to parse raster source correctly.
C:\mb\windows-builds-64\packages\osrm-backend\unit_tests\extractor\raster_source.cpp(27): last checkpoint: "raster_test" entry.

*** 1 failure is detected in the test module "extractor tests"
```

daniel-j-h:
Wohoo it works! :)

Regarding the issue: I quickly skimmed the files in question and I _think_ it's `raster_source.hpp:45` not specifying the `binary` mode. Can you verify if that's the case? I think we already have an issue open with the stream modes on Windows, don't we?

BergWerkGIS:
@daniel-j-h Looking good on AppVeyor :white_check_mark: 
https://ci.appveyor.com/project/DennisOSRM/osrm-backend/build/1.0.2860

[`std::ios::binary`](https://github.com/Project-OSRM/osrm-backend/pull/2118/commits/fb3e0bb9eeeb4836b8ced5c771d45866a2155b36?diff=split#diff-b45d00b0678c271de432a95e8663feafR45) did the trick for me locally - thanks, I had totally forgotten about that.

BTW, I've added the build output of MSBuild as text artifacts for easy offline reading:
https://ci.appveyor.com/project/DennisOSRM/osrm-backend/build/1.0.2860/artifacts

daniel-j-h:
Phew, thanks a lot for your efforts! :tada:

Handing this over to @TheMarex for for a quick look then we're good to go I think!

BergWerkGIS:
No worries. Was quite a deep dive into boost #defines and Cmake though.

src/extractor/edge_based_graph_factory.cpp
```diff
@@ -138,7 +138,7 @@ void EdgeBasedGraphFactory::InsertEdgeBasedNode(const NodeID node_u, const NodeI
     NodeID current_edge_source_coordinate_id = node_u;
 
     // traverse arrays from start and end respectively
-    for (const auto i : util::irange(0UL, geometry_size))
+    for (const auto i : util::irange(std::size_t{ 0 }, geometry_size))
```
TheMarex: What is the problem here?

daniel-j-h: I think this is due to the long literal as long is 4 byte on 64bit Windows, 8 byte on 64bit Unix, but I'm unsure why the compiler complains only here.

BergWerkGIS: These are the error messages: https://github.com/Project-OSRM/osrm-backend/pull/2118#issuecomment-199802077
I think VS is not able to deduct which type `0UL` is.

daniel-j-h: Exactly, both types to `irange` have to be the same. On 64 bit Unix `size_t` is 64 bit / 8 bytes therefore the types match, but on 64 bit Windows the types are different.

TheMarex:
Rebased and merging into `develop`. :+1:

Issue #2012, opened 2016-02-24
Only have global uturns parameter


freenerd:
Currently the `viapoint` API supports enabling or disabling turns on coordinates. In https://github.com/Project-OSRM/osrm-backend/issues/519#issuecomment-188355482 the question of how these concretely work came up.

Previous uturn work:

- https://github.com/Project-OSRM/osrm-backend/issues/100
- https://github.com/Project-OSRM/osrm-backend/issues/729
- https://github.com/Project-OSRM/osrm-backend/issues/1922

#### My current understanding:

- U turns are 180 degree turns of the vehicle on the same way
- U turns over separated ways are always allowed
- U turns in dead-end ways are always allowed
- U turns on ways that are not dead-ends are never allowed
- U turns for the start coordinate are always allowed (since you may
- U turns for the start coordinate with bearings
- U turns do not matter for the last coordinate, since a vehicle only has to reach it regardless from direction
- U turns can be enabled/disabled for intermediate coordinates for cases where the arrival bearing is 180 degrees different from the departure bearing

#### Example with U Turn for the intermediate coordinate enabled

![c261c2be-0c3b-11e4-99d5-1a63e854d66d](https://cloud.githubusercontent.com/assets/25713/13297157/30b83ffc-db31-11e5-925a-51be717cc2a9.png)

#### Use cases

1. Vehicle Type
  - Bikes/Pedestrians can uturn easily everywhere, cars/trucks can't
  - Should be solved via Lua profile, which are vehicle dependent
2. Local Law
  - U-Turn laws seem to be very different per locality (and even street) per 
  - Should be solvable in Lua profile based on tags in OSM? Potentially even a mapping between locality and laws?
3. User Preference (from org running OSRM)
  - Should be a global Lua parameter to set default to false/true
4. User Preference (from user requesting OSRM)
  - Needs to be solved via request parameter as is right now

Looking at the use cases, 4. is currently solved, which transitively also solved the other cases but delegates this to the user. 1. and 3. could be solved by being able to set the default in the Lua profile. 2 is a tough nut to crack.

#### Next actions

- [ ] Expose this API/Library documentation better. Should this be a whole new section?
- [ ] Allow a Lua parameter to globally set uturn default

#### Way out there

- Think allowing uturns on segments https://github.com/Project-OSRM/osrm-backend/issues/729 (or rather not think about it just now)

vvk95:
> Should be solvable in Lua profile based on tags in OSM?
 
[...]

PR #2121, opened 2016-03-19
Global uturns parameter

TheMarex:
Addresses https://github.com/Project-OSRM/osrm-backend/issues/2012 and https://github.com/Project-OSRM/osrm-backend/issues/2127


include/engine/api/route_parameters.hpp
```diff
@@ -60,10 +60,10 @@ struct RouteParameters : public BaseParameters
                     const bool alternatives_,
                     const GeometriesType geometries_,
                     const OverviewType overview_,
-                    std::vector<boost::optional<bool>> uturns_,
+                    boost::optional<bool> uturns_,
                     Args... args_)
         : BaseParameters{std::forward<Args>(args_)...}, steps{steps_}, alternatives{alternatives_},
-          geometries{geometries_}, overview{overview_}, uturns{std::move(uturns_)}
+          geometries{geometries_}, overview{overview_}, uturns{uturns_}
```
daniel-j-h: When not moving from it, uturns can be a const-ref or const val

TheMarex:
@daniel-j-h this should be done from my side. Can you review?

include/engine/api/route_parameters.hpp
```diff
@@ -60,10 +60,10 @@ struct RouteParameters : public BaseParameters
                     const bool alternatives_,
                     const GeometriesType geometries_,
                     const OverviewType overview_,
-                    std::vector<boost::optional<bool>> uturns_,
+                    boost::optional<bool> uturns_,
```
daniel-j-h: Comment below still holds for this line ^

include/engine/datafacade/datafacade_base.hpp
```diff
@@ -143,6 +143,8 @@ class BaseDataFacade
     virtual std::size_t GetCoreSize() const = 0;
 
     virtual std::string GetTimestamp() const = 0;
+
+    virtual bool GetUTurnsDefault() const = 0;
```
daniel-j-h: Implement dummy for mock facade! On my unit-tests-for-v5 branch I had to hunt down a ton of extra member functions as no one cared to implement them for too long!

include/engine/datafacade/internal_datafacade.hpp
```diff
@@ -662,6 +678,8 @@ class InternalDataFacade final : public BaseDataFacade
     }
 
     std::string GetTimestamp() const override final { return m_timestamp; }
+
+    bool GetUTurnsDefault() const override final { return m_profile_properties.allow_u_turn_at_via; }
```
daniel-j-h: I don't think there are benefits for putting `final` on all those member functions, as the facade class itself is already marked `final`.

include/engine/datafacade/shared_datafacade.hpp
```diff
@@ -68,6 +69,7 @@ class SharedDataFacade final : public BaseDataFacade
     std::unique_ptr<storage::SharedMemory> m_layout_memory;
     std::unique_ptr<storage::SharedMemory> m_large_memory;
     std::string m_timestamp;
+    extractor::ProfileProperties* m_profile_properties;
```
daniel-j-h: Why is this a raw pointer? Who owns it?

/edit oh, I see below this is coming from shared memory.

include/extractor/extractor.hpp
```diff
@@ -40,6 +40,8 @@ namespace osrm
 namespace extractor
 {
 
+class ProfileProperties;
```
daniel-j-h: `struct` otherwise clang complains about mismatched tag

include/extractor/profile_properties.hpp
```diff
@@ -0,0 +1,46 @@
...
+
+    void SetTrafficSignalPenalty(const double traffic_signal_penalty_)
+    {
+        traffic_signal_penalty = static_cast<int>(traffic_signal_penalty_ * 10.);
+    }
+
+    //! penalty to cross a traffic light in deci-seconds
+    int traffic_signal_penalty;
+    //! penalty to do a uturn in deci-seconds
+    int u_turn_penalty;
```
daniel-j-h: Can we use `boost::numeric_cast<To>(from)` for casting and storing into `int`s in this class? Makes it possible to detect overflows and other issues (throwing an exception by default).

include/extractor/restriction_parser.hpp
```diff
@@ -19,6 +19,8 @@ namespace osrm
 namespace extractor
 {
 
+class ProfileProperties;
```
daniel-j-h: `struct`

include/extractor/scripting_environment.hpp
```diff
@@ -23,18 +26,28 @@ namespace extractor
...
+        lua_State *state;
+    };
+
     explicit ScriptingEnvironment(const std::string &file_name);
 
     ScriptingEnvironment(const ScriptingEnvironment &) = delete;
     ScriptingEnvironment &operator=(const ScriptingEnvironment &) = delete;
 
-    lua_State *GetLuaState();
+    Context &GetContex();
```
daniel-j-h: Can this be `const`?

src/extractor/edge_based_graph_factory.cpp
```diff
@@ -438,20 +441,16 @@ void EdgeBasedGraphFactory::GenerateEdgeExpandedEdges(
...
-                luabind::call_function<double>(lua_state, "turn_function", 180. - angle);
-            return static_cast<int>(penalty);
-        }
-        catch (const luabind::error &er)
-        {
-            util::SimpleLogger().Write(logWARNING) << er.what();
-        }
+        // call lua profile to compute turn penalty
+        double penalty = luabind::call_function<double>(lua_state, "turn_function", 180. - angle);
+        return static_cast<int>(penalty);
```
daniel-j-h: `boost::numeric_cast<To>(from);`

src/extractor/extraction_containers.cpp
```diff
@@ -288,6 +288,8 @@ void ExtractionContainers::PrepareEdges(lua_State *segment_state)
     const auto all_edges_list_end_ = all_edges_list.end();
     const auto all_nodes_list_end_ = all_nodes_list.end();
 
+    auto has_segment_function = util::lua_function_exists(segment_state, "segment_function");
```
daniel-j-h: `const`

src/extractor/graph_compressor.cpp
```diff
@@ -13,8 +13,7 @@ namespace osrm
 namespace extractor
 {
 
-GraphCompressor::GraphCompressor(SpeedProfileProperties speed_profile)
-    : speed_profile(std::move(speed_profile))
+GraphCompressor::GraphCompressor()
```
daniel-j-h: This can be `GraphCompressor() = default;` in the header.
I can see this or the same with dtors all over the code base.

src/extractor/scripting_environment.cpp
```diff
@@ -51,122 +52,137 @@ int luaErrorCallback(lua_State *state)
 }
 }
 
+ScriptingEnvironment::Context::Context() : state(luaL_newstate()) {}
+
+ScriptingEnvironment::Context::~Context() { lua_close(state); }
```
daniel-j-h: I feel like we should abstract away the Lua state resource ownership and then hold onto an object by-value via the Rule Of Zero (which means we don't need ctor/dtor/{move,copy}{assign,construct} any longer).

What I mean is e.g. doing https://gist.github.com/daniel-j-h/7ca2bf73047c788f3d9b#file-luaraii-cc-L14-L24
and then holding on to a state by value and composing it with other member attributes.

src/extractor/scripting_environment.cpp
```diff
@@ -51,122 +52,137 @@ int luaErrorCallback(lua_State *state)
...
+             .def_readwrite("is_startpoint", &ExtractionWay::is_startpoint)
+             .def_readwrite("duration", &ExtractionWay::duration)
+             .property("forward_mode", &ExtractionWay::get_forward_mode,
+                       &ExtractionWay::set_forward_mode)
+             .property("backward_mode", &ExtractionWay::get_backward_mode,
+                       &ExtractionWay::set_backward_mode),
+         luabind::class_<osmium::Way>("Way")
+             .def("get_value_by_key", &osmium::Way::get_value_by_key)
+             .def("get_value_by_key", &get_value_by_key<osmium::Way>)
+             .def("id", &osmium::Way::id),
```
daniel-j-h: Every now and then users ask for means to iterate over nodes in ways. Maybe we should register a [Way's .nodes()](http://docs.osmcode.org/libosmium/latest/classosmium_1_1Way.html) member function (and then probably also the `WayNodeList` for iteration) exposing this to the profiles.

src/extractor/scripting_environment.cpp
```diff
@@ -51,122 +52,137 @@ int luaErrorCallback(lua_State *state)
...
     std::lock_guard<std::mutex> lock(init_mutex);
     bool initialized = false;
     auto &ref = script_contexts.local(initialized);
     if (!initialized)
     {
-        std::shared_ptr<lua_State> state(luaL_newstate(), lua_close);
+        std::shared_ptr<Context> state = std::make_shared<Context>();
         ref = state;
-        InitLuaState(ref.get());
+        InitContext(*ref);
```
daniel-j-h: Why not `ref = make_shared(); InitContext(*ref)`? I see no need for incrementing and then immediately afterwards decrementing the ref count.

daniel-j-h:
New file added:

- [ ] add documentation to the Wiki
- [ ] make sure the example still builds (I _think_ you need to adapt the server paths?)

include/engine/datafacade/shared_datafacade.hpp
TheMarex: Ownership is clearly managed actually. None of the types here have ownership of their containing data. Its all managed by the shared memory segment. We could copy, but that would be inconsistent with everything else.

include/extractor/scripting_environment.hpp
TheMarex: Yes should be a unique_ptr actually. Function can't be const because this modifies the locks.

PR #2131, opened 2016-03-22
improves consistency of fork handling

MoKob:
Small fix that improves the consistency of fork handling.

Previously, one side of a potential fork could emit a fork, while the other side would emit no instruction.
This is not the case anymore.


TheMarex:
:+1: Merged.

PR #2132, opened 2016-03-22
Only allow to specify the common base path

TheMarex:
This PR removes the ability to specify separate paths for each input file an thus greatly simplifies path handling.


include/engine/datafacade/internal_datafacade.hpp
```diff
@@ -215,31 +215,36 @@ class InternalDataFacade final : public BaseDataFacade
         }
     }
 
-    void LoadDatasourceInfo(const boost::filesystem::path &datasource_names_file,
-                            const boost::filesystem::path &datasource_indexes_file)
+    void LoadDatasourceInfo(const std::string &datasource_names_file,
+                            const std::string &datasource_indexes_file)
```
daniel-j-h: Why suddenly strings? boost::fs::path should be fine as we're using boost::fs::ifstream anyway which is the same as std::ifstream but in addition has a ctor from boost::fs::path.

include/engine/datafacade/internal_datafacade.hpp
```diff
@@ -215,31 +215,36 @@ class InternalDataFacade final : public BaseDataFacade
...
-                                    sizeof(std::size_t));
-            if (number_of_datasources > 0)
-            {
-                m_datasource_list.resize(number_of_datasources);
-                datasources_stream.read(reinterpret_cast<char *>(&(m_datasource_list[0])),
-                                        number_of_datasources * sizeof(uint8_t));
-            }
+            throw util::exception("Could not open " + datasource_indexes_file + " for reading!");
         }
+        BOOST_ASSERT(datasources_stream);
```
daniel-j-h: redundant with the exception branch above

include/engine/datafacade/internal_datafacade.hpp
```diff
@@ -215,31 +215,36 @@ class InternalDataFacade final : public BaseDataFacade
...
+            datasources_stream.read(reinterpret_cast<char *>(&(m_datasource_list[0])),
+                                    number_of_datasources * sizeof(uint8_t));
+        }
+
+        boost::filesystem::ifstream datasourcenames_stream(datasource_names_file, std::ios::binary);
+        if (!datasourcenames_stream)
+        {
+            throw util::exception("Could not open " + datasource_names_file + " for reading!");
+        }
+        BOOST_ASSERT(datasourcenames_stream);
```
daniel-j-h: redundant

include/engine/datafacade/internal_datafacade.hpp
```diff
@@ -215,31 +215,36 @@ class InternalDataFacade final : public BaseDataFacade
...
+        std::string name;
+        while (std::getline(datasourcenames_stream, name))
+        {
+            m_datasource_names.push_back(name);
```
daniel-j-h: move(name)

include/engine/datafacade/shared_datafacade.hpp
```diff
@@ -684,16 +684,10 @@ class SharedDataFacade final : public BaseDataFacade
...
-            datasource_name_id > m_datasource_name_offsets.size())
-        {
-            if (datasource_name_id == 0)
-                return "lua profile";
-            return "UNKNOWN";
-        }
+        BOOST_ASSERT(m_datasource_name_offsets.size() >= 1);
+        BOOST_ASSERT(m_datasource_name_offsets.size() > datasource_name_id);
 
+        std::string result;
```
daniel-j-h: reserve as size is known; back_inserter does not reserve in advance as it only get one element at a time

daniel-j-h: looking at the code, the same applies to the for_each a few lines above --- maybe check the code you touched for this

include/engine/engine_config.hpp
```diff
@@ -41,24 +42,9 @@ namespace engine
...
-                       {"edgesdata", base.string() + ".edges"},
-                       {"coredata", base.string() + ".core"},
-                       {"geometries", base.string() + ".geometry"},
-                       {"timestamp", base.string() + ".timestamp"},
-                       {"datasource_names", base.string() + ".datasource_names"},
-                       {"datasource_indexes", base.string() + ".datasource_indexes"},
-                       {"namesdata", base.string() + ".names"}}
-    {
-    }
+    bool IsValid() const;
```
daniel-j-h: What we do here is implementing constructors and their invariant establishment ourselves; what about having a proper constructor that throws in case its invariant could not be established (not isValid)? At the moment responsibility for checking validity is pushed to the user.

include/storage/storage_config.hpp
```diff
@@ -0,0 +1,33 @@
...
+
+namespace osrm
+{
+namespace storage
+{
+struct StorageConfig
+{
+    StorageConfig() = default;
+    StorageConfig(const boost::filesystem::path &base);
+    bool IsValid() const;
```
daniel-j-h: Same here with isValid

src/engine/engine.cpp
```diff
@@ -136,8 +135,11 @@ Engine::Engine(EngineConfig &config)
     }
     else
     {
-        util::populate_base_path(config.server_paths);
-        query_data_facade = util::make_unique<datafacade::InternalDataFacade>(config.server_paths);
+        if (!config.storage_config.IsValid())
+        {
+            throw util::exception("Invalid file paths given!");
+        }
```
daniel-j-h: This is the ctors job

src/engine/engine_config.cpp
```diff
@@ -0,0 +1,27 @@
...
+        storage_config.datasource_indexes_path.empty() && storage_config.names_data_path.empty();
+
+    const bool limits_valid =
+        (max_locations_distance_table == -1 || max_locations_distance_table > 2) &&
+        (max_locations_map_matching == -1 || max_locations_map_matching > 2) &&
+        (max_locations_trip == -1 || max_locations_trip > 2) &&
+        (max_locations_viaroute == -1 || max_locations_viaroute > 2);
+
+    return ((use_shared_memory && all_path_are_empty) || storage_config.IsValid()) && limits_valid;
+}
```
daniel-j-h: ctor's job

src/storage/storage.cpp
```diff
@@ -72,10 +72,12 @@ void deleteRegion(const SharedDataType region)
     }
 }
 
-Storage::Storage(const DataPaths &paths_) : paths(paths_) {}
+Storage::Storage(StorageConfig config_) : config(std::move(config_)) {}
 
 int Storage::Run()
 {
+    BOOST_ASSERT_MSG(config.IsValid(), "Invalid storage config");
```
daniel-j-h: Invalid config should not be allowed

src/storage/storage.cpp
```diff
@@ -211,12 +125,12 @@ int Storage::Run()
...
     shared_layout_ptr->SetBlockSize<char>(SharedDataLayout::FILE_INDEX_PATH,
-                                          file_index_path.length() + 1);
+                                          config.file_index_path.length() + 1);
 
     // collect number of elements to store in shared memory object
-    util::SimpleLogger().Write() << "load names from: " << names_data_path;
+    util::SimpleLogger().Write() << "load names from: " << config.names_data_path;
     // number of entries in name index
-    boost::filesystem::ifstream name_stream(names_data_path, std::ios::binary);
+    boost::filesystem::ifstream name_stream(config.names_data_path, std::ios::binary);
```
daniel-j-h: if(!name_stream) { throw oops(); }

src/storage/storage.cpp
```diff
@@ -230,7 +144,7 @@ int Storage::Run()
     shared_layout_ptr->SetBlockSize<char>(SharedDataLayout::NAME_CHAR_LIST, number_of_chars);
 
     // Loading information for original edges
-    boost::filesystem::ifstream edges_input_stream(edges_data_path, std::ios::binary);
+    boost::filesystem::ifstream edges_input_stream(config.edges_data_path, std::ios::binary);
```
daniel-j-h: same

src/storage/storage.cpp
```diff
@@ -287,12 +201,12 @@ int Storage::Run()
 
     // load timestamp size
     std::string m_timestamp;
-    if (boost::filesystem::exists(timestamp_path))
+    if (boost::filesystem::exists(config.timestamp_path))
```
daniel-j-h: Should deriving suffixes from base path _and_ file existence be checked in the config's ctor?

src/storage/storage_config.cpp
```diff
@@ -0,0 +1,36 @@
...
+           boost::filesystem::is_regular_file(file_index_path) &&
+           boost::filesystem::is_regular_file(hsgr_data_path) &&
+           boost::filesystem::is_regular_file(nodes_data_path) &&
+           boost::filesystem::is_regular_file(edges_data_path) &&
+           boost::filesystem::is_regular_file(core_data_path) &&
+           boost::filesystem::is_regular_file(geometries_path) &&
+           boost::filesystem::is_regular_file(timestamp_path) &&
+           boost::filesystem::is_regular_file(datasource_names_path) &&
+           boost::filesystem::is_regular_file(datasource_indexes_path) &&
+           boost::filesystem::is_regular_file(names_data_path);
```
daniel-j-h: is_regular_file is checked here, so why the file_exists check above?

daniel-j-h: From the docs:

> [Note: regular_file implies appropriate <fstream> operations would succeed, assuming no hardware, permission, access, or file system race errors.

That's interesting as it means we're fscked on all those listed cases, as we're not checking .read() and .write().

src/tools/routed.cpp
```diff
@@ -41,6 +46,111 @@ BOOL WINAPI console_ctrl_handler(DWORD ctrl_type)
...
 
+const static unsigned INIT_OK_START_ENGINE = 0;
+const static unsigned INIT_OK_DO_NOT_START_ENGINE = 1;
+const static unsigned INIT_FAILED = -1;
+
+// generate boost::program_options object for the routing part
+inline unsigned
+generateServerProgramOptions(const int argc,
+                             const char *argv[],
+                             boost::filesystem::path &base_path,
```
daniel-j-h: const?

include/engine/engine_config.hpp
TheMarex: We need this for the same reason, we need the `IsValid()` for `*Parameters`. The object is filled element-by element (by boost program options). Also you can modify the member variables since they are public - giving no guarantees at all.

include/engine/datafacade/shared_datafacade.hpp
TheMarex: Actually there is already a reserve there.

src/storage/storage.cpp
TheMarex: Done.

CMakeLists.txt
```diff
@@ -80,7 +80,7 @@ add_executable(osrm-extract src/tools/extract.cpp)
 add_executable(osrm-contract src/tools/contract.cpp)
 add_executable(osrm-routed src/tools/routed.cpp $<TARGET_OBJECTS:SERVER> $<TARGET_OBJECTS:UTIL>)
 add_executable(osrm-datastore src/tools/store.cpp $<TARGET_OBJECTS:UTIL>)
-add_library(osrm src/osrm/osrm.cpp $<TARGET_OBJECTS:ENGINE> $<TARGET_OBJECTS:UTIL>)
+add_library(osrm src/osrm/osrm.cpp $<TARGET_OBJECTS:ENGINE> $<TARGET_OBJECTS:UTIL> $<TARGET_OBJECTS:STORAGE>)
```
daniel-j-h: trailing space

daniel-j-h: ENGINE now depends on STORAGE, but we do not specify that dependency.

For example: https://gist.github.com/daniel-j-h/c0310687bf9664a7bbf1

Required me to add STORAGE to the engine tests, too, because ENGINE now depends on STORATGE in https://github.com/Project-OSRM/osrm-backend/pull/2088/commits/9f9ec2a45c830f01a9899b7d7d0d126b8cdf00c1.

PR #2135, opened 2016-03-23
[ready] adds a distinction between roundabouts and rotaries

MoKob:
Some larger rotaries have distinct names and should be handled differently from the pure roundabouts.
A roundabout is usually a small circle at an intersection.

This will unlock instructions of the form `at Logan Circle take the 4th exit` and the roundabout-turns (`at the roundabout turn left`).

Tasks:
 - [x] compute radius of roundabout
       - [x] implement radius of a circle
       - [x] find consistent roundabout size

 - [x] determine threshold / rules for distinction between rotary and roundabout (see below)
 - [x] implement distinction in `handleRoundabouts` moved to a dedicated function

 - [x] reviewed @daniel-j-h
 - [ ] on merge: update API for new instruction/new data and handle in API (@karenzshea)
       Proposal: `rotary | a special (often named) roundabout -- can offer additional_name as attribute`
       or: `additional_name` a roundabout may specify an `additional_name` to indicate the name of the roundabout itself.

### DIFF
--- a/include/extractor/guidance/toolkit.hpp
+++ b/include/extractor/guidance/toolkit.hpp
@@ -350,7 +350,8 @@ inline bool requiresNameAnnounced(const std::string &from, const std::string &to
         const auto ref_begin = name.find_first_of('(');
         if (ref_begin != std::string::npos)
         {
-            out_name = name.substr(0, ref_begin);
+            //we might need to trim a space, if there is a reference
+            out_name = name.substr(0, ref_begin - (ref_begin > 0 && name[ref_begin-1] == ' '));
             out_ref = name.substr(ref_begin + 1, name.find_first_of(')') - 1);
         }
         else
--- a/include/extractor/guidance/turn_analysis.hpp
+++ b/include/extractor/guidance/turn_analysis.hpp
@@ -112,7 +112,8 @@ class TurnAnalysis
     // Processing of roundabouts
     // Produces instructions to enter/exit a roundabout or to stay on it.
     // Performs the distinction between roundabout and rotaries.
-    std::vector<ConnectedRoad> handleRoundabouts(const EdgeID via_edge,
+    std::vector<ConnectedRoad> handleRoundabouts(const bool is_rotary,
+                                                 const EdgeID via_edge,
                                                  const bool on_roundabout,
                                                  const bool can_exit_roundabout,
                                                  std::vector<ConnectedRoad> intersection) const;
@@ -152,8 +153,8 @@ class TurnAnalysis
                                                  std::vector<ConnectedRoad> intersection) const;
 
     // Any Junction containing motorways
-    std::vector<ConnectedRoad> handleMotorwayJunction(
-        const EdgeID via_edge, std::vector<ConnectedRoad> intersection) const;
+    std::vector<ConnectedRoad>
+    handleMotorwayJunction(const EdgeID via_edge, std::vector<ConnectedRoad> intersection) const;
 
     std::vector<ConnectedRoad> handleFromMotorway(const EdgeID via_edge,
                                                   std::vector<ConnectedRoad> intersection) const;
@@ -193,6 +194,7 @@ class TurnAnalysis
                                                 std::vector<ConnectedRoad> intersection,
                                                 const std::size_t up_to) const;
 
+    bool isRotary(const NodeID nid) const;
 }; // class TurnAnalysis
 
 } // namespace guidance
--- a/include/extractor/guidance/turn_instruction.hpp
+++ b/include/extractor/guidance/turn_instruction.hpp
@@ -89,24 +89,32 @@ struct TurnInstruction
         return TurnInstruction(TurnType::NoTurn, DirectionModifier::UTurn);
     }
 
-    static TurnInstruction REMAIN_ROUNDABOUT(const DirectionModifier modifier)
+    static TurnInstruction REMAIN_ROUNDABOUT(bool is_rotary, const DirectionModifier modifier)
     {
+        (void)is_rotary; // staying does not require a different instruction at the moment
         return TurnInstruction(TurnType::StayOnRoundabout, modifier);
     }
 
-    static TurnInstruction ENTER_ROUNDABOUT(const DirectionModifier modifier)
+    static TurnInstruction ENTER_ROUNDABOUT(bool is_rotary, const DirectionModifier modifier)
     {
-        return TurnInstruction(TurnType::EnterRoundabout, modifier);
+        return {is_rotary ? TurnType::EnterRotary : TurnType::EnterRoundabout, modifier};
     }
 
-    static TurnInstruction EXIT_ROUNDABOUT(const DirectionModifier modifier)
+    static TurnInstruction EXIT_ROUNDABOUT(bool is_rotary, const DirectionModifier modifier)
     {
-        return TurnInstruction(TurnType::ExitRoundabout, modifier);
+        return {is_rotary ? TurnType::ExitRotary : TurnType::ExitRoundabout, modifier};
+    }
+
+    static TurnInstruction ENTER_AND_EXIT_ROUNDABOUT(bool is_rotary,
+                                                     const DirectionModifier modifier)
+    {
+        return {is_rotary ? TurnType::EnterAndExitRotary : TurnType::EnterAndExitRoundabout,
+                modifier};
     }
 
     static TurnInstruction SUPPRESSED(const DirectionModifier modifier)
     {
-        return TurnInstruction{TurnType::Suppressed, modifier};
+        return {TurnType::Suppressed, modifier};
     }
 };
 
--- a/include/util/coordinate_calculation.hpp
+++ b/include/util/coordinate_calculation.hpp
@@ -4,6 +4,7 @@
 #include "util/coordinate.hpp"
 
 #include <boost/math/constants/constants.hpp>
+#include <boost/optional.hpp>
 
 #include <utility>
 
@@ -60,6 +61,16 @@ double bearing(const Coordinate first_coordinate, const Coordinate second_coordi
 
 // Get angle of line segment (A,C)->(C,B)
 double computeAngle(const Coordinate first, const Coordinate second, const Coordinate third);
+
+// find the center of a circle through three coordinates
+boost::optional<Coordinate> circleCenter(const Coordinate first_coordinate,
+                                         const Coordinate second_coordinate,
+                                         const Coordinate third_coordinate);
+
+// find the radius of a circle through three coordinates
+double circleRadius(const Coordinate first_coordinate,
+                    const Coordinate second_coordinate,
+                    const Coordinate third_coordinate);
 
 // factor in [0,1]. Returns point along the straight line between from and to. 0 returns from, 1
 // returns to
--- a/src/engine/api/json_factory.cpp
+++ b/src/engine/api/json_factory.cpp
@@ -1,16 +1,16 @@
 #include "engine/api/json_factory.hpp"
 
-#include "engine/polyline_compressor.hpp"
 #include "engine/hint.hpp"
+#include "engine/polyline_compressor.hpp"
 
 #include <boost/assert.hpp>
-#include <boost/range/irange.hpp>
 #include <boost/optional.hpp>
+#include <boost/range/irange.hpp>
 
-#include <string>
-#include <utility>
 #include <algorithm>
 #include <iterator>
+#include <string>
+#include <utility>
 #include <vector>
 
 using TurnType = osrm::extractor::guidance::TurnType;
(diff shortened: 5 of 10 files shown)


MoKob:
The consistent roundabout size requires a, otherwise unseen, traversal of the full roundabout.
To make sure we get the same radius for all settings, we have to fully uncover all roundabout vertices and select three of them in a consistent way.

src/util/coordinate_calculation.cpp
```diff
@@ -322,7 +396,8 @@ double degreeToPixel(FloatLatitude lat, unsigned zoom)
 }
 
 // Converts a WMS tile coordinate (z,x,y) into a wsg84 bounding box
-void xyzToWSG84(const int x, const int y, const int z, double &minx, double &miny, double &maxx, double &maxy)
+void xyzToWSG84(
+    const int x, const int y, const int z, double &minx, double &miny, double &maxx, double &maxy)
 {
```
MoKob: I get the strong feeling that my editor/yours simply does not recognise the same line-length...

daniel-j-h: Is this again a clang-format issue?

daniel-j-h: I built a local LLVM 3.8, can use that for clang-format'ing.

In case you also want one, here's how to build it: https://gist.github.com/daniel-j-h/8ee8c511f89ebe80821b (needs some adaption on OSX, I think there's no sha256sum and no realpath, otherwise should be fine)

MoKob: Don't know whether it was formatted correctly prior. But I force the use of clang-format-3-6 now and I get this changes... no idea why.

MoKob:
I've now implemented the following discovery mechanism:

A rotary has a dedicated name (no street connects to it that has the same name).
It is at least 30 meters in diameter.

This heuristic might end up leaving some smaller rotaries to stay roundabouts. We might have to reconsider the 30 meters. But based on my tests on Berlin, the setting worked quite well, only reporting reasonable results as rotary.

MoKob:
This request also fixes a, previously non-existent, problem in post-processing that only occurs if we actually have rotaries.

Furthermore, I found an error in the -- over my initial version -- simplified method to determine whether a name change is trivial (required to find name changes in roundabouts based on references).

include/extractor/guidance/toolkit.hpp
```diff
@@ -350,7 +350,8 @@ inline bool requiresNameAnnounced(const std::string &from, const std::string &to
         const auto ref_begin = name.find_first_of('(');
         if (ref_begin != std::string::npos)
         {
-            out_name = name.substr(0, ref_begin);
+            //we might need to trim a space, if there is a reference
+            out_name = name.substr(0, ref_begin - (ref_begin > 0 && name[ref_begin-1] == ' '));
```
daniel-j-h: substr copies the string and returns a new one; maybe use Boost's in-place `trim_right` which also shows intent.

daniel-j-h: By the way, the stdlib's string type is horribly broken for such things (it pre-dates the stdlib included in the standard), that's why there is string_ref / string_view or hand-made equivalents e.g. in the Chromium code base.

- https://groups.google.com/a/chromium.org/forum/#!topic/chromium-dev/EUqoIz2iFU4
- https://news.ycombinator.com/item?id=8704318

MoKob: this would require a change in the signature to copy the string there. At the moment, trim-right wouldn't help, would it?

include/extractor/guidance/turn_analysis.hpp
```diff
@@ -112,7 +112,8 @@ class TurnAnalysis
     // Processing of roundabouts
     // Produces instructions to enter/exit a roundabout or to stay on it.
     // Performs the distinction between roundabout and rotaries.
-    std::vector<ConnectedRoad> handleRoundabouts(const EdgeID via_edge,
+    std::vector<ConnectedRoad> handleRoundabouts(const bool is_rotary,
```
daniel-j-h: This function takes three bools; it handles all those 2^3 possible states?

include/extractor/guidance/turn_instruction.hpp
```diff
@@ -89,19 +89,27 @@ struct TurnInstruction
...
+    static TurnInstruction EXIT_ROUNDABOUT(bool is_rotary, const DirectionModifier modifier)
     {
-        return TurnInstruction(TurnType::ExitRoundabout, modifier);
+        return {is_rotary ? TurnType::ExitRotary : TurnType::ExitRoundabout, modifier};
+    }
+
+    static TurnInstruction ENTER_AND_EXIT_ROUNDABOUT(bool is_rotary,
+                                                     const DirectionModifier modifier)
+    {
+        return TurnInstruction(
```
daniel-j-h: Is there a difference between the `return { }` above and suddenly the `return TurnInst( )` here?

MoKob: no, Just cleaned that part up for the above functions and missed one.

include/util/coordinate_calculation.hpp
```diff
@@ -24,14 +25,14 @@ const constexpr double EARTH_RADIUS_WGS84 = 6378137.0;
...
-    const constexpr double MAXEXTENT = EARTH_RADIUS_WGS84 * boost::math::constants::pi<double>();
-    // ^ math functions are not constexpr since they have side-effects (setting errno) :(
-    const double MAX_LATITUDE = RAD_TO_DEGREE * (2.0 * std::atan(std::exp(180.0 * DEGREE_TO_RAD)) - boost::math::constants::half_pi<double>());
-    const constexpr double MAX_LONGITUDE = 180.0;
+// earth circumference devided by 2
+const constexpr double MAXEXTENT = EARTH_RADIUS_WGS84 * boost::math::constants::pi<double>();
+// ^ math functions are not constexpr since they have side-effects (setting errno) :(
+const double MAX_LATITUDE = RAD_TO_DEGREE * (2.0 * std::atan(std::exp(180.0 * DEGREE_TO_RAD)) -
+                                             boost::math::constants::half_pi<double>());
+const constexpr double MAX_LONGITUDE = 180.0;
```
daniel-j-h: What changed here? This is noise from formatting as far as I can see?

include/extractor/guidance/turn_analysis.hpp
MoKob: All relevant states. the first is simply switch between two sets of instructions.

include/util/coordinate_calculation.hpp
```diff
@@ -69,6 +70,16 @@ double bearing(const Coordinate first_coordinate, const Coordinate second_coordi
 // Get angle of line segment (A,C)->(C,B)
 double computeAngle(const Coordinate first, const Coordinate second, const Coordinate third);
 
+// find the cnter of a circle through three coordinates
```
daniel-j-h: +e

src/engine/api/json_factory.cpp
```diff
@@ -164,6 +164,8 @@ util::json::Object makeRouteStep(guidance::RouteStep step, util::json::Value geo
     route_step.values["distance"] = std::move(step.distance);
     route_step.values["duration"] = std::move(step.duration);
     route_step.values["name"] = std::move(step.name);
+    if( step.additional_name != "" )
```
daniel-j-h: spaces

src/engine/guidance/post_processing.cpp
```diff
@@ -308,7 +312,7 @@ LegGeometry resyncGeometry(LegGeometry leg_geometry, const std::vector<RouteStep
         leg_geometry.segment_offsets.push_back(step.geometry_end - 1);
     }
 
-    //remove the data fromt the reached-target step again
+    // remove the data fromt the reached-target step again
```
daniel-j-h: -t

include/util/coordinate_calculation.hpp
MoKob: Didn't touch it. So formatting. The question remains how we can get a consistent formatting in...

src/extractor/guidance/turn_analysis.cpp
```diff
@@ -95,6 +97,109 @@ inline bool isRampClass(EdgeID eid, const util::NodeBasedDynamicGraph &node_base
...
 
+bool TurnAnalysis::isRotary(const NodeID nid) const
+{
+    // translate a node ID into its respective coordinate stored in the node_info_list
+    const auto getCoordinate = [this](const NodeID node)
+    {
+        return util::Coordinate(node_info_list[node].lon, node_info_list[node].lat);
+    };
+
+    unsigned roundabout_name = 0;
```
daniel-j-h: maybe add suffix _id since this is not the name but the name id

src/extractor/guidance/turn_analysis.cpp
```diff
@@ -95,6 +97,109 @@ inline bool isRampClass(EdgeID eid, const util::NodeBasedDynamicGraph &node_base
...
+    std::set<unsigned> connected_names;
```
daniel-j-h: It does not have to be ordered, does it? If not, use `std::unordered_set`.

src/extractor/guidance/turn_analysis.cpp
```diff
@@ -95,6 +97,109 @@ inline bool isRampClass(EdgeID eid, const util::NodeBasedDynamicGraph &node_base
...
+        return continue_edge;
+    };
+
+    // the roundabout radius has to be the same for all locations we look at it from
+    // to guarantee this, we search the full roundabout for its vertices
+    // and select the three smalles ids
+    std::vector<NodeID> roundabout_nodes;
+    roundabout_nodes.reserve(40);
+    roundabout_nodes.push_back(nid);
+    while (roundabout_nodes.size() < 40) // hard abort if something is wrong
```
daniel-j-h: 40? Why 40?

src/extractor/guidance/turn_analysis.cpp
```diff
@@ -95,6 +97,109 @@ inline bool isRampClass(EdgeID eid, const util::NodeBasedDynamicGraph &node_base
...
+    {
+        auto eid = getNextOnRoundabout(roundabout_nodes.back());
```
daniel-j-h: const

src/extractor/guidance/turn_analysis.cpp
```diff
@@ -95,6 +97,109 @@ inline bool isRampClass(EdgeID eid, const util::NodeBasedDynamicGraph &node_base
...
+        roundabout_nodes.size() >= 3
+            ? util::coordinate_calculation::circleRadius(getCoordinate(roundabout_nodes[0]),
+                                                         getCoordinate(roundabout_nodes[1]),
+                                                         getCoordinate(roundabout_nodes[2]))
+            : 0.5 * util::coordinate_calculation::haversineDistance(
+                        getCoordinate(roundabout_nodes[0]), getCoordinate(roundabout_nodes[1]));
+
+    // check whether the circle computation has gone wrong
+    if (std::isinf(radius))
+        return false;
```
daniel-j-h: When can this happen? Shouldn't this be an assertion?

src/extractor/guidance/turn_analysis.cpp
MoKob: this value is chosen at random to avoid getting stuck in invalid constructs. I can think of no situation in which we would get 40 intersections in a roundabout, but if there is a strange loop within one, this hard-abort stops it from happening.

src/extractor/guidance/turn_analysis.cpp
MoKob: It can happen if the roundabout-nodes have identical coordinates. We could, instead, do a filtering of the coordinate values based on the coordinates.
Whats your take on this?

daniel-j-h: Hm at least add a comment for that particular situation :+1:

include/extractor/guidance/toolkit.hpp
daniel-j-h: No you are right.

src/extractor/guidance/turn_analysis.cpp
```diff
@@ -1051,7 +1165,7 @@ void TurnAnalysis::handleDistinctConflict(const EdgeID via_edge,
         }
         else
         {
-            // FIXME this should possibly know about the actual roads?
+            // FIXME this should possibly know aboat the actual roads?
```
daniel-j-h: I don't understand that change

src/util/coordinate_calculation.cpp
```diff
@@ -257,6 +257,81 @@ double computeAngle(const Coordinate first, const Coordinate second, const Coord
...
+
+double circleRadius(const Coordinate C1, const Coordinate C2, const Coordinate C3)
+{
+    // a circle by three points requires thee distinct points
+    auto center = circleCenter(C1, C2, C3);
+    if (center)
+        return haversineDistance(C1, *center);
+    else
+        return std::numeric_limits<double>::infinity();
+}
```
daniel-j-h: If you use boost's `numeric_cast<To>(from)` you can catch subtle errors (=> exception)

src/extractor/guidance/turn_analysis.cpp
danpat: I think it's Canadian.  Although in that case, I really would've expected `aboot`.

karenzshea:
While poking around on cleaning up some text instruction code in api-directions, I was wondering, is it possible to guarantee that a roundabout step will never be returned without an `exit` modifier?

MoKob:
@karenzshea depends on how you want to handle simply entering the roundabout. Technically not a reasonable route, but we shouldn't fail on them.

src/engine/api/json_factory.cpp
```diff
@@ -164,6 +165,9 @@ util::json::Object makeRouteStep(guidance::RouteStep step, util::json::Value geo
     route_step.values["distance"] = std::round(step.distance * 10) / 10.;
     route_step.values["duration"] = std::round(step.duration * 10) / 10.;
     route_step.values["name"] = std::move(step.name);
+    if (step.additional_name != "")
```
TheMarex: Use `.empty()`

src/extractor/guidance/turn_analysis.cpp
```diff
@@ -95,6 +98,111 @@ inline bool isRampClass(EdgeID eid, const util::NodeBasedDynamicGraph &node_base
...
+    {
+        return false;
+    }
+
+    // calculate the radius of the roundabout/rotary. For two coordinates, we assume a minimal
+    // circle
+    // with both vertices right at the other side (so half their distance in meters).
+    // Otherwise, we construct a circle through the first tree vertices.
+    auto node_itr = roundabout_nodes.begin();
+    const double radius = roundabout_nodes.size() >= 3
```
TheMarex: An easy approximation for the circle diameter would be to calculate the distance of all coordinates to the first coordinates and take the maximum. That would remove the need for having a special function which needs tests and needs to be maintained.

MoKob: I would agree without the compressed geometry. If the geometry is is removed, the diameter can be approximated very badly if the entry/exit vertices are not evenly distributed.

TheMarex:
This looks good. Rebasing and merging.

Issue #2151, opened 2016-03-26
Douglas Peucker algorithm uses bad approximation


TheMarex:
Currently the Douglas Peucker algorithm we use for geometry simplification uses an approximate distance metric that leads to results like this:

http://bl.ocks.org/d/867d579ec59bfc965bf1

The turn coordinate should not be removed at this zoom level, however the geometry contains almost co-linear points that are preserved.

danpat:
You think it would make sense to only simplify the geometry between turn coordinates?

TheMarex:
No we need to replace the approximation by the correct method. We are not dependent on the precise turn coordinates being present in the geometry anymore since we moved that to the `Maneuver` object.

TheMarex:
The offending code is [here](https://github.com/Project-OSRM/osrm-backend/blob/rewrite/new-api/src/engine/douglas_peucker.cpp#L21-L57) this needs to be replaced by the point-segment distance from `coordinate_calculation`.

TheMarex:
This is fixed.

PR #2153, opened 2016-03-28
Use correct perpendicular distance measure when simplifying

danpat:
This changes the Douglas-Peucker line simplification algorithm to use the same measurements as other point-near-line calculations in the codebase (`util::coordinate_calculation::perpendicularDistance`).

This corrects line-simplification problems like the one pointed out in https://github.com/Project-OSRM/osrm-backend/issues/2151

Example:
screen shot 2016-03-28 at 2 55 49 pm

Blue = old problematic simplification, pink = new correct simplification.

### DIFF
--- a/src/engine/douglas_peucker.cpp
+++ b/src/engine/douglas_peucker.cpp
@@ -16,53 +16,6 @@ namespace osrm
 namespace engine
 {
 
-namespace
-{
-// FIXME This algorithm is a very naive approximation that leads to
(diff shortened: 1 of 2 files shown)


TheMarex:
@danpat great work. :+1: we might want to optimize this a little bit later and move out the coordinate projection (only needs to be done for the points in between the segment endpoints)

danpat:
@TheMarex agreed, I'll ticket this and stick it in a future milestone.

Issue #2096, opened 2016-03-15
Length Zero Segments


MoKob:
https://github.com/Project-OSRM/osrm-backend/blob/rewrite/new-api/include/engine/guidance/assemble_steps.hpp#L62-L143 does a step-back from OSRM-4.9.

Given a coordinate that snaps to an intersection of two roads, we have to select a phantom node pre-route. This can result in an additional turn-instruction both at the front and at the end of the route.

E.g
(1)
a ----- b
 |
 | 
 c
(2)

This could result in `Head West on (ab) for 0 m`, `Turn Left onto (ac)`, `Arrive at (2)`, when routing from `(1)` to `(2)`.

In my view, the removal of this additional segment in the front/back is something that we should keep.
The Route (1)-(2) should result in `Head South on (ac)`, `Arrive at (2)`.

/cc @TheMarex

TheMarex:
This should be solved in post-processing not during initial assembly to reduce code complexity. This is just a special case of segment merging - we should detect similar cases of absurdly short segments (say below 2 meters) and merge them to the previous/next segment.

How do you see this integrated into `include/guidance/post_processing.cpp`? I see that you operate on `PathData` instead of `RouteStep` objects. It would be beneficial if you could port this to use `RouteStep`, which would enable easier layering (also the `PathData` is already filtered and assembled into segments, which is what you are actually interested at that point, no?).

danpat:
@TheMarex we should consider http://wiki.openstreetmap.org/wiki/Tag:highway%3Dstop#Tagging_minor_road_stops
before doing any merging.

TheMarex:
@danpat this would happen post-routing when the after instructions are generated. We don't even have knowledge about stop signs at this level.

MoKob:
Wherever you solve it, it requires to adjust stuff thats stored in the phantom node in ways that are hard to do, when you use preprocessed data in that phantom-node.

TheMarex:
@MoKob this was addressed by moving `post-processing` to `RouteStep`, right?

MoKob:
No. It just no becomes possible to solve it. So far the segments are processed as they come from `assemble_steps` and `assemble_geometry`. The post-processing still needs to handle unnecessary segments in the front-back.

MoKob:
fixed with https://github.com/Project-OSRM/osrm-backend/issues/2159

PR #2159, opened 2016-03-29
Fix/zero length segments

MoKob:
Addresses https://github.com/Project-OSRM/osrm-backend/issues/2096

A minor problem with this request, for very short routes the summary can end up not being contained in the route instructions at all.

 - [X] @daniel-j-h says :+1:

### DIFF
--- a/include/engine/api/route_api.hpp
+++ b/include/engine/api/route_api.hpp
@@ -106,7 +106,8 @@ class RouteAPI : public BaseAPI
                  * Using post-processing on basis of route-steps for a single leg at a time
                  * comes at the cost that we cannot count the correct exit for roundabouts.
                  * We can only emit the exit nr/intersections up to/starting at a part of the leg.
-                 * If a roundabout is not terminated in a leg, we will end up with a enter-roundabout
+                 * If a roundabout is not terminated in a leg, we will end up with a
+                 *enter-roundabout
                  * and exit-roundabout-nr where the exit nr is out of sync with the previous enter.
                  *
                  *         | S |
@@ -118,16 +119,22 @@ class RouteAPI : public BaseAPI
                  *         |   |
                  *         |   |
                  *
-                 * Coming from S via V to T, we end up with the legs S->V and V->T. V-T will say to take
+                 * Coming from S via V to T, we end up with the legs S->V and V->T. V-T will say to
+                 *take
                  * the second exit, even though counting from S it would be the third.
-                 * For S, we only emit `roundabout` without an exit number, showing that we enter a roundabout
+                 * For S, we only emit `roundabout` without an exit number, showing that we enter a
+                 *roundabout
                  * to find a via point.
                  * The same exit will be emitted, though, if we should start routing at S, making
                  * the overall response consistent.
                  */
 
-                 leg.steps = guidance::postProcess(std::move(steps));
-                 leg_geometry = guidance::resyncGeometry(std::move(leg_geometry),leg.steps);
+                leg.steps = guidance::postProcess(std::move(steps));
+                guidance::trimShortSegments(leg.steps, leg_geometry);
+                leg.steps = guidance::assignRelativeLocations(std::move(leg.steps), leg_geometry,
+                                                              phantoms.source_phantom,
+                                                              phantoms.target_phantom);
+                leg_geometry = guidance::resyncGeometry(std::move(leg_geometry), leg.steps);
             }
 
             leg_geometries.push_back(std::move(leg_geometry));
--- a/include/engine/guidance/post_processing.hpp
+++ b/include/engine/guidance/post_processing.hpp
@@ -1,6 +1,7 @@
 #ifndef ENGINE_GUIDANCE_POST_PROCESSING_HPP
 #define ENGINE_GUIDANCE_POST_PROCESSING_HPP
 
+#include "engine/phantom_node.hpp"
 #include "engine/guidance/route_step.hpp"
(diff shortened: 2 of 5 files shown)


include/engine/guidance/post_processing.hpp
```diff
@@ -16,6 +17,15 @@ namespace guidance
...
 
+// trim initial/final segment of very short length. Function set up for in-place modification
+void trimShortSegments(std::vector<RouteStep> &steps, LegGeometry &geometry);
+
+//assign relative locations to depart/arrive instructions
+void assignRelativeLocations(std::vector<RouteStep> &steps,
+                             const LegGeometry &geometry,
+                             const PhantomNode &source_node,
+                             const PhantomNode &target_node);
+
```
daniel-j-h: Hm the other functions here take a vector by value, modify it and move it out again. Should we be consistent here? 

(Not that big of a deal for sure)

src/engine/guidance/post_processing.cpp
```diff
@@ -292,6 +296,124 @@ std::vector<RouteStep> postProcess(std::vector<RouteStep> steps)
...
+    //
+    // a --- b
+    //       |
+    //       c
+    //
+    // If a route from b to c is requested, both a--b and b--c could be selected as start segment.
+    // In case of a--b, we end up with an unwanted turn saying turn-right onto b-c.
+    // These cases start off with an initial segment which is of zero length.
+    // We have to be careful though, since routing that starts in a roundabout has a valid
+    // initial segment of length zero and we cannot delete the upcoming segment.
```
daniel-j-h: :+1: for explanations!

src/engine/guidance/post_processing.cpp
```diff
@@ -292,6 +296,124 @@ std::vector<RouteStep> postProcess(std::vector<RouteStep> steps)
...
+
+        // finally remove the initial (now duplicated move)
+        steps.erase(steps.begin());
+
+        // and update the leg geometry indices for the removed entry
+        std::for_each(steps.begin(), steps.end(), [](RouteStep &step)
+                      {
+                          --step.geometry_begin;
+                          --step.geometry_end;
+                      });
```
daniel-j-h: format

src/engine/guidance/post_processing.cpp
```diff
@@ -292,6 +296,124 @@ std::vector<RouteStep> postProcess(std::vector<RouteStep> steps)
...
+    }
+
+    auto &next_to_last_step = *(steps.end() - 2);
```
daniel-j-h: With the erase above only the guard at the top of the function guarantees for this not being an out of bounds access; maybe assert on size here, too just to be sure (e.g. for when we incorporate future changes).

src/engine/guidance/post_processing.cpp
MoKob: thats what I get from clang-format...

src/engine/guidance/post_processing.cpp
```diff
@@ -292,6 +296,130 @@ std::vector<RouteStep> postProcess(std::vector<RouteStep> steps)
...
+    const auto distance_to_start = util::coordinate_calculation::haversineDistance(
+        source_node.input_location, leg_geometry.locations[0]);
+    const auto initial_modifier =
+        distance_to_start >= MINIMAL_RELATIVE_DISTANCE &&
+                distance_to_start <= MAXIMAL_RELATIVE_DISTANCE
+            ? angleToDirectionModifier(util::coordinate_calculation::computeAngle(
+                  source_node.input_location, leg_geometry.locations[0], leg_geometry.locations[1]))
+            : extractor::guidance::DirectionModifier::UTurn;
+
+    steps.front().maneuver.instruction.direction_modifier = initial_modifier;
```
daniel-j-h: assert on size at the top

src/engine/guidance/post_processing.cpp
```diff
@@ -292,6 +296,130 @@ std::vector<RouteStep> postProcess(std::vector<RouteStep> steps)
...
+
+    const auto distance_from_end = util::coordinate_calculation::haversineDistance(
+        target_node.input_location, leg_geometry.locations.back());
+    const auto final_modifier =
+        distance_from_end >= MINIMAL_RELATIVE_DISTANCE &&
+                distance_from_end <= MAXIMAL_RELATIVE_DISTANCE
+            ? angleToDirectionModifier(util::coordinate_calculation::computeAngle(
+                  leg_geometry.locations[leg_geometry.locations.size() - 2],
+                  leg_geometry.locations[leg_geometry.locations.size() - 1],
```
daniel-j-h: also on location size

daniel-j-h: Note: If you're fine with getting an exception (instead of UB) you can use `.at()` which always bounds-checks instead of the subscript operator `[]`.

include/engine/api/route_api.hpp
```diff
@@ -127,6 +127,8 @@ class RouteAPI : public BaseAPI
                  */
 
                  leg.steps = guidance::postProcess(std::move(steps));
+                 guidance::trimShortSegments(leg.steps,leg_geometry);
+                 leg.steps = guidance::assignRelativeLocations(std::move(leg.steps),leg_geometry,phantoms.source_phantom,phantoms.target_phantom);
                  leg_geometry = guidance::resyncGeometry(std::move(leg_geometry),leg.steps);
```
TheMarex: Needs format for sure.

TheMarex:
Good work @MoKob! :+1: Merging.

Issue #2054, opened 2016-03-04
Expose node ids


TheMarex:
We want to expose the OSM ids in the response. The only problem is added memory consumption in the server. We need to minimize this in order to not break the bill.
A global dataset currently has 800 million nodes for the bike profile. That means about 6GB additional memory if we would 64bit OSM IDs.

However we already safe the node ids sorted by OSM ID. If we would use a block based prefix encoding we can probably safe a lot of memory. `RangeTable` already does something similar for the name index.

We can run some numbers on this but I would assume that the jump between used OSM ids is probably less than `2^16-1` as such we could use `unsigned short`.

Using a block lengths of 16 ids per block we get to about 1.85GB.

zscore:
This would be very useful! What kind of work needs to be done to get this going?

danpat:
@zscore It should be reasonably straightforward, if you can figure out how to navigate the codebase :-)

The `InternalDataFacade` needs to be modified here:

https://github.com/Project-OSRM/osrm-backend/blob/develop/include/engine/datafacade/internal_datafacade.hpp#L138-L145

This loop reads the `.nodes` file.  The `QueryNode.node_id` member has the OSM Node ID we're looking for, currently we only use the lon/lat values that get read, the `.node_id` is being discarded.

We would need to add a new vector and accessor methods to the `*DataFacade` classes to store the node IDs.

The above needs to be repeated for the `SharedDataFacade`.

Once that is done, when the API response is constructed, we would need to fetch the OSM ids IDs (probably at the same place we get the coordinate array) and place it in the response object.

This would give us a list of OSM node IDs that correspond to the coordinates we return in the response.  The OSM way IDs are never stored in OSRM, there is no way to expose them.

@TheMarex 's description above regarding prefix encoding would be a memory optimization - not necessary for an initial draft of this feature, but would certainly be desirable.

zscore:
Cool! If no one more qualified steps up to do this, I may give this a try.

TheMarex:
@freenerd @danpat w.r.t to exposing this in the API my feeling here is per-step:

- makes it possible to have a per-instruction view of the route
- consistent with the uncompressed geometry
- exposing node ids for the overview geometry cold cause confusion in case the overview geometry is simplified

danpat:
[...]

PR #2165, opened 2016-03-30
[WIP] Feature/expose node ids

danpat:
This is to address #2054

This change loads the original OSM node ID values into memory along with coordinates, instead of just discarding them.

They are then appended to the `route` object in `trip` and `match` responses, so that the route through the original OSM data can be followed.  These values could then be used with an offline service to get detailed metadata about the route that OSRM does not have the capacity to store/expose.

Current status: 

  - works only with the `InternalDataFacade`
  - will always be 2 nodes short - the nodes just before and just after the PhantomNodes - not sure whether to include these or not, or include a null value so that the node array aligns with the coordinate array.


TheMarex:

PR #2166, opened 2016-03-30
[ready] Roundabout fix / handle separate exits only and special cases correctly

MoKob:
This pull request addresses perceived counting in roundabouts which have a separate exit connected to the entry point.

The current model will break for left-sided driving and needs to be addressed if we add the possibility to find out about left-sided driving.

 - [x] only count separate exits
 - [x] fix only exits
 - [x] fix only enter

### DIFF
--- a/src/engine/guidance/post_processing.cpp
+++ b/src/engine/guidance/post_processing.cpp
@@ -50,7 +50,7 @@ void fixFinalRoundabout(std::vector<RouteStep> &steps)
          --propagation_index)
     {
         auto &propagation_step = steps[propagation_index];
-        if (entersRoundabout(propagation_step.maneuver.instruction))
+        if (propagation_index == 0 || entersRoundabout(propagation_step.maneuver.instruction))
         {
             propagation_step.maneuver.exit = 0;
             propagation_step.geometry_end = steps.back().geometry_begin;
@@ -141,6 +141,7 @@ void closeOffRoundabout(const bool on_roundabout,
             {
                 // TODO at this point, we can remember the additional name for a rotary
                 // This requires some initial thought on the data format, though
+
                 propagation_step.maneuver.exit = step.maneuver.exit;
                 propagation_step.geometry_end = step.geometry_end;
                 propagation_step.name = step.name;
@@ -205,6 +206,7 @@ std::vector<RouteStep> postProcess(std::vector<RouteStep> steps)
 #endif
     // Count Street Exits forward
     bool on_roundabout = false;
+    bool has_entered_roundabout = false;
 
     // adds an intersection to the initial route step
     // It includes the length of the last step, until the intersection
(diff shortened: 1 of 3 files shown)


src/extractor/guidance/turn_analysis.cpp
```diff
@@ -117,14 +117,22 @@ std::vector<TurnOperation> TurnAnalysis::getTurns(const NodeID from, const EdgeI
...
         }
-        else
+        // Exiting roundabouts at an entry point is technically a data-modelling issue.
+        // This workaround handles cases in which an exit follows the entry.
+        // To correctly represent perceived exits, we only count exits leading to a
+        // separate vertex than the one we are coming from that are in the direction of
+        // the roundabout.
+        // The sorting of the angles represents a problem for left-sided driving, though.
+        // FIXME in case of left-sided driving, we have to check whether we can enter the
+        // roundabout later in the cycle, rather than prior.
```
daniel-j-h: The profiles already expose left / right side driving, do you want to pass through the scripting environment to the guidance framework at some point in time?

MoKob: I have no background on how this is handled currently. I just know that the decision whether an exit exists past the point of entry depends on the direction in which the roundabout can be entered. If an exit is between entry and origin, we need to count it. This `between` depends on whether we drive clockwise or counterclockwise in the roundabout

src/engine/guidance/post_processing.cpp
```diff
@@ -319,11 +329,10 @@ void trimShortSegments(std::vector<RouteStep> &steps, LegGeometry &geometry)
...
     // In case of a--b, we end up with an unwanted turn saying turn-right onto b-c.
     // These cases start off with an initial segment which is of zero length.
-    // We have to be careful though, since routing that starts in a roundabout has a valid
-    // initial segment of length zero and we cannot delete the upcoming segment.
+    // We have to be careful though, since routing that starts in a roundabout has a valid.
+    // To catch these cases correctly, we have to perform trimming prior to the post-processing
 
-    if (steps.front().distance <= std::numeric_limits<double>::epsilon() &&
-        !entersRoundabout((steps.begin() + 1)->maneuver.instruction))
+    if (steps.front().distance <= std::numeric_limits<double>::epsilon())
```
daniel-j-h: assert on steps size, for .front

MoKob: This is checked a few lines prior

src/engine/guidance/post_processing.cpp
```diff
@@ -298,6 +303,10 @@ std::vector<RouteStep> postProcess(std::vector<RouteStep> steps)
 
 void trimShortSegments(std::vector<RouteStep> &steps, LegGeometry &geometry)
 {
+#if OSRM_POST_PROCESSING_PRINT_DEBUG
```
TheMarex: Can we please not have commented out debug logging? I know this might be convenient for debugging but littering the code base with it makes it harder to read and maintain.

TheMarex:
Good to go once the leftover debug code is gone.

src/engine/guidance/post_processing.cpp
MoKob: The problem here is, that we need to merge some stuff while it is augmented further.
When I remove this, I have to add it for all further improvements and remove it again.

MoKob:
Removed the debug output from post-processing.

PR #2169, opened 2016-03-30
[wip] fixes local path unpacking when start/end are on same segment

MoKob:
This pull request fixes an issue if both source/target phantom snap to the same segment. In these cases, the unpacking would include the weight of the target node twice.

It uncovers a problem with the weight vectors, though. @danpat The weight we see [here](https://github.com/Project-OSRM/osrm-backend/blob/fix/advisory/include/engine/routing_algorithms/routing_base.hpp#L412-L430) should theoretically be equal to zero in all cases. For some reason it isn't, though, showing that https://github.com/Project-OSRM/osrm-backend/blob/fix/advisory/include/engine/routing_algorithms/routing_base.hpp#L403 is off. It looks like the weight-vectors might include the turn-penalty, which leads to inconsistencies with the phantom-nodes.


MoKob:
messed up base branch, closing

Issue #2162, opened 2016-03-29
Speed bug related to…bearing? End segment?


lbud:
@MoKob @TheMarex  — I would have suspected this related to #2155, but that is fixed and this is not:

features/car/advisory.feature has a test as follows:
```
    Scenario: Car - Advisory speed overwrites backwards maxspeed
        Given the node map
            | a | b | c |

        And the ways
            | nodes | highway       | maxspeed:backward | maxspeed:advisory:backward |
            | ab    | residential   | 90                | 45                         |
            | bc    | residential   |                   | 45                         |

        When I route I should get
            | from | to | route | speed        |
            | b    | a  | ab    | 47 km/h +- 1 |
            | c    | b  | bc    | 47 km/h +- 1 |
```

b-a fails with `19 km/h`; c-b succeeds.
Now, if I add an additional segment to the left side of the requested segment, i.e.
```
    Scenario: Car - Advisory speed overwrites backwards maxspeed
        Given the node map
            | a | b | c | d |

        And the ways
            | nodes | highway       | maxspeed:backward | maxspeed:advisory:backward |
            | ab    | residential   |                   | 45                         |
            | bc    | residential   | 90                | 45                         |
            | cd    | residential   |                   | 45                         |

        When I route I should get
            | from | to | route | speed        |
            | c    | b  | bc    | 47 km/h +- 1 |
            | d    | c  | cd    | 47 km/h +- 1 |
```

this succeeds.

The difference between these routes is that on the failing test, where `ab` is the leftmost segment, the departure and arrival are both on `ab`, and the departure has no bearing before or after. On the passing test, where we instead test the middle segment with the same attributes, the departure and arrival are on different segments, and the departure has a bearing_before of 0 and a bearing_after of 270:

*FAILS:* 
```
[ { distance: 999.8,
    duration: 192.7,
    name: 'ab',
    maneuver: 
     { type: 'depart',
       location: [Object],
       bearing_before: 0,
       bearing_after: 0 },
    geometry: '_ibEeadE???dw@',
    mode: 'driving' },
  { distance: 0,
    duration: 0,
    name: 'ab',
    maneuver: 
     { type: 'arrive',
       location: [Object],
       bearing_before: 270,
       bearing_after: 0 },
    geometry: '',
    mode: 'driving' } ]
[ { distance: 999.9,
    duration: 76.6,
    name: 'bc',
    maneuver: 
     { type: 'depart',
       location: [Object],
       bearing_before: 0,
[...]

Issue #2161, opened 2016-03-29
Which bearing should cucumber test?


miccolis:
Looks like we've got more detailed bearing info in the new API which confuses the tests when:

```
  Scenario: Bearing when going northwest   # features/testbot/bearing.feature:7
    Given the profile "testbot"            # features/step_definitions/data.js:8
    Given the node map                     # features/step_definitions/data.js:47
      | b |   |
      |   | a |
    And the ways                           # features/step_definitions/data.js:121
      | nodes |
      | ab    |
    When I route I should get              # features/step_definitions/routing.js:4
      | from | to | route | bearing |
      | a    | b  | ab    | 315     |
      Tables were not identical:
          from |     to |     route |     bearing
      | (-) a | (-) b | (-) ab | (-) 315 |
      | (+) a | (+) b | (+) ab | (+) 0 |
```

And instructions like

```
[ { distance: 141.5,
    duration: 28.2,
    name: 'ab',
    maneuver: 
     { type: 'depart',
       location: [Object],
       bearing_before: 0,
       bearing_after: 0 },
    geometry: 'kcbEsnbE??sDrD',
    mode: 'driving' },
  { distance: 0,
    duration: 0,
    name: 'ab',
    maneuver: 
     { type: 'arrive',
       location: [Object],
       bearing_before: 315,
       bearing_after: 0 },
    geometry: '',
    mode: 'driving' } ]
```

I suspect we want to be using the `bearing_after` for the depart step here. Does the test fixture need to be updated or? /cc @MoKob

MoKob:
@TheMarex you are correct. Sorry, I missed updating this here as well. #2170 fixes this issue (its a problem introduced mainly in tests due to phantom-node snapping). After the merge, both bearings should be consistent.

TheMarex:
Passes now. :tada:

PR #2170, opened 2016-03-30
[ready] fixes local unpacking/identical coordinate issues

MoKob:
This pull request fixes an issue if both source/target phantom snap to the same segment. In these cases, the unpacking would include the weight of the target node twice.

It uncovers a problem with the weight vectors, though. @danpat The weight we see [here](https://github.com/Project-OSRM/osrm-backend/blob/fix/advisory/include/engine/routing_algorithms/routing_base.hpp#L412-L430) should theoretically be equal to zero in all cases. For some reason it isn't, though, showing that https://github.com/Project-OSRM/osrm-backend/blob/fix/advisory/include/engine/routing_algorithms/routing_base.hpp#L403 is off. It looks like the weight-vectors might include the turn-penalty, which leads to inconsistencies with the phantom-nodes.

This pull request solves: https://github.com/Project-OSRM/osrm-backend/issues/2162
and https://github.com/Project-OSRM/osrm-backend/issues/2161 (at least offers more consistency)

 - [X] @daniel-j-h :eyes: 
 - [x] requires https://github.com/Project-OSRM/osrm-backend/pull/2166 merged

### DIFF
--- a/include/engine/routing_algorithms/routing_base.hpp
+++ b/include/engine/routing_algorithms/routing_base.hpp
@@ -89,11 +89,10 @@ template <class DataFacadeT, class Derived> class BasicRoutingInterface
             const std::int32_t new_distance = reverse_heap.GetKey(node) + distance;
             if (new_distance < upper_bound)
             {
+                // if loops are forced, they are so at the source
                 if (new_distance >= 0 &&
-                    (!force_loop_forward ||
-                     forward_heap.GetData(node).parent !=
-                         node) // if loops are forced, they are so at the source
-                    && (!force_loop_reverse || reverse_heap.GetData(node).parent != node))
+                    (!force_loop_forward || forward_heap.GetData(node).parent != node) &&
+                    (!force_loop_reverse || reverse_heap.GetData(node).parent != node))
                 {
                     middle_node_id = node;
                     upper_bound = new_distance;
@@ -320,8 +319,11 @@ template <class DataFacadeT, class Derived> class BasicRoutingInterface
                 for (std::size_t i = start_index; i < end_index; ++i)
                 {
                     unpacked_path.push_back(
-                        PathData{id_vector[i], name_index, weight_vector[i],
-                                 extractor::guidance::TurnInstruction::NO_TURN(), travel_mode});
+                        PathData{id_vector[i],
+                                 name_index,
+                                 weight_vector[i],
+                                 extractor::guidance::TurnInstruction::NO_TURN(),
+                                 travel_mode});
                 }
                 BOOST_ASSERT(unpacked_path.size() > 0);
                 unpacked_path.back().turn_instruction = turn_instruction;
@@ -339,6 +341,7 @@ template <class DataFacadeT, class Derived> class BasicRoutingInterface
                     // However the first segment duration needs to be adjusted to the fact that the
                     // source phantom is in the middle of the segment.
                     // We do this by subtracting v--s from the duration.
+
                     BOOST_ASSERT(unpacked_path.front().duration_until_turn >= source_weight);
                     unpacked_path.front().duration_until_turn -= source_weight;
                 }
@@ -369,8 +372,10 @@ template <class DataFacadeT, class Derived> class BasicRoutingInterface
         }
 
         std::size_t end_index = phantom_node_pair.target_phantom.fwd_segment_position;
+        const std::size_t delta = target_traversed_in_reverse ? 1 : 0;
         if (target_traversed_in_reverse)
         {
+            start_index += 1;
             std::reverse(id_vector.begin(), id_vector.end());
             std::reverse(weight_vector.begin(), weight_vector.end());
             end_index = id_vector.size() - phantom_node_pair.target_phantom.fwd_segment_position;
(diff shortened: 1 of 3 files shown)


include/engine/routing_algorithms/routing_base.hpp
```diff
@@ -360,7 +363,7 @@ template <class DataFacadeT, class Derived> class BasicRoutingInterface
             if (target_traversed_in_reverse)
             {
                 start_index =
-                    id_vector.size() - phantom_node_pair.source_phantom.fwd_segment_position - 1;
+                    id_vector.size() - phantom_node_pair.source_phantom.fwd_segment_position - 1 ;
```
daniel-j-h: space ` `

include/engine/routing_algorithms/routing_base.hpp
```diff
@@ -393,7 +398,9 @@ template <class DataFacadeT, class Derived> class BasicRoutingInterface
             BOOST_ASSERT(i < id_vector.size());
             BOOST_ASSERT(phantom_node_pair.target_phantom.forward_travel_mode > 0);
             unpacked_path.emplace_back(
-                PathData{id_vector[i], phantom_node_pair.target_phantom.name_id, weight_vector[i],
+                PathData{id_vector[i],
+                         phantom_node_pair.target_phantom.name_id,
+                         weight_vector[i-delta],
```
daniel-j-h: If we emplace back we don't need the `PathData` constructor. In case this was the situation where emplace_back was failing without the constructor, use push_back, as there is an rvalue overload.

src/engine/guidance/post_processing.cpp
```diff
@@ -319,41 +328,63 @@ void trimShortSegments(std::vector<RouteStep> &steps, LegGeometry &geometry)
...
-    // We have to be careful though, since routing that starts in a roundabout has a valid
-    // initial segment of length zero and we cannot delete the upcoming segment.
+    // We have to be careful though, since routing that starts in a roundabout has a valid.
+    // To catch these cases correctly, we have to perform trimming prior to the post-processing
 
-    if (steps.front().distance <= std::numeric_limits<double>::epsilon() &&
-        !entersRoundabout((steps.begin() + 1)->maneuver.instruction))
+    BOOST_ASSERT(geometry.locations.size() >= steps.size());
+    const bool zero_length_step = steps.front().distance <= std::numeric_limits<double>::epsilon();
+    const bool duplicated_coordinate = geometry.locations.at(0) == geometry.locations.at(1);
```
daniel-j-h: We assert on `steps.size() >= 2` and `geometry.locations.size() >= steps.size()`, thus `geometry.locations.size() >= 2`, therefore you can change `.at(x)` to `[x]` for x in 0, 1.

src/engine/guidance/post_processing.cpp
```diff
@@ -319,41 +328,63 @@ void trimShortSegments(std::vector<RouteStep> &steps, LegGeometry &geometry)
...
+        {
+            // move offsets to front
+            BOOST_ASSERT(geometry.segment_offsets[1] == 1);
+            // geometry offsets have to be adjusted. Move all offsets to the front and reduce by
+            // one.
+            std::transform(geometry.segment_offsets.begin() + 1, geometry.segment_offsets.end(),
+                           geometry.segment_offsets.begin(), [](const std::size_t val)
+                           {
+                               return val - 1;
+                           });
```
daniel-j-h: * see below

src/engine/guidance/post_processing.cpp
```diff
@@ -319,41 +328,63 @@ void trimShortSegments(std::vector<RouteStep> &steps, LegGeometry &geometry)
...
+        }
+        else
+        {
+            steps.front().geometry_begin = 1;
+            // reduce all offsets by 1
+            std::transform(geometry.segment_offsets.begin(), geometry.segment_offsets.end(),
+                           geometry.segment_offsets.begin(), [](const std::size_t val)
+                           {
+                               return val - 1;
+                           });
```
daniel-j-h: \* duplicate see above, maybe put a lambda at the top and name it appropriately

src/engine/guidance/post_processing.cpp
```diff
@@ -381,7 +413,25 @@ void trimShortSegments(std::vector<RouteStep> &steps, LegGeometry &geometry)
             TurnInstruction::NO_TURN(), WaypointType::Arrive, geometry);
         steps.pop_back();
         // the geometry indices of the last step are already correct;
+    } else if( geometry.locations.at( geometry.locations.size() - 2) == geometry.locations.at( geometry.locations.size() - 1 ) )
```
daniel-j-h: format

daniel-j-h: we assert on size, [] should be fine

src/engine/guidance/post_processing.cpp
```diff
@@ -401,7 +451,8 @@ std::vector<RouteStep> assignRelativeLocations(std::vector<RouteStep> steps,
         distance_to_start >= MINIMAL_RELATIVE_DISTANCE &&
                 distance_to_start <= MAXIMAL_RELATIVE_DISTANCE
             ? angleToDirectionModifier(util::coordinate_calculation::computeAngle(
-                  source_node.input_location, leg_geometry.locations.at(0), leg_geometry.locations.at(1)))
+                  source_node.input_location, leg_geometry.locations.at(0),
+                  leg_geometry.locations.at(1)))
```
daniel-j-h: same

src/engine/guidance/post_processing.cpp
MoKob: these two are not exactly identical. One transforms in place, one moves to the front in addition. Maybe clarify with an additional comment?

TheMarex:
:+1: great work.

PR #2171, opened 2016-03-30
Fix-up cucumber timestamp handling

miccolis:
Gets us to `425 scenarios (113 failed, 312 passed)`. /cc @TheMarex @lbud


TheMarex:
:+1:

Issue #2147, opened 2016-03-25
(Trip Plugin) Segementation Fault / Issue in FarthestInsertionTrip 


K-Leon:
I just encountered an unroutable Trip in latest stable.

Using [this Route](https://router.project-osrm.org/trip?loc=46.9622733,7.4801433&loc=46.9621183,7.480145&loc=46.9621383,7.4801533&loc=46.9621633,7.4801717&loc=46.9621617,7.480165&loc=46.9621883,7.48015&loc=46.96225,7.4802083&loc=46.9622783,7.4801783&loc=46.9622517,7.4802083&loc=46.9616883,7.4803667&loc=46.96228,7.4801117) results in a crash.

These trips resolve more or less to these two segments. So we have several Waypoints without a real distance between each other.
46.962396,7.480171
46.961509,7.479702

**Results in:**
**osrm-routed: /src/library/../plugins/../algorithms/trip_farthest_insertion.hpp:212:** std::vector<unsigned int> osrm::trip::FarthestInsertionTrip(const NodeIDIterator&, const NodeIDIterator&, std::size_t, const DistTableWrapper<int>&) [with NodeIDIterator = __gnu_cxx::__normal_iterator<const unsigned int*, std::vector<unsigned int> >; std::size_t = long unsigned int]: **Assertion `max_from >= 0' failed.**

Anyone knows what could have caused this?

daniel-j-h:
I can reproduce the error on latest develop and an Switzerland extract from Geofabrik.

--------------------------------------------------------------------------------------------------------------------------------

For the record, this is the v5 query:

    curl 'http://127.0.0.1:5000/trip/v1/driving/46.9622733,7.4801433;46.9621183,7.480145;46.9621383,7.4801533;46.9621633,7.4801717;46.9621617,7.480165;46.9621883,7.48015;46.96225,7.4802083;46.9622783,7.4801783;46.9622517,7.4802083;46.9616883,7.4803667;46.96228,7.4801117'

and I can not reproduce the issue with v5.

--------------------------------------------------------------------------------------------------------------------------------

What seems odd, though is the following:

Here's the `Trip` query for [46.962396,7.480171](http://www.openstreetmap.org/#map=18/46.962396,7.480171) and [46.96245/7.47965](http://www.openstreetmap.org/#map=18/46.96245/7.47965):

    http --body 'http://127.0.0.1:5000/trip/v1/driving/46.962396,7.480171;46.961509,7.479702'

[...]

PR #2172, opened 2016-03-30
Handle all distances between coordinates being zero in Farthest Insertion algorithm.

daniel-j-h:
Assertions for additional safety and sanity.

References:
- resolves https://github.com/Project-OSRM/osrm-backend/issues/2147 for v4

### DIFF
--- a/include/engine/trip/trip_farthest_insertion.hpp
+++ b/include/engine/trip/trip_farthest_insertion.hpp
@@ -3,6 +3,7 @@
 
 #include "engine/search_engine.hpp"
 #include "util/dist_table_wrapper.hpp"
+#include "util/typedefs.hpp"
 
 #include "osrm/json_container.hpp"
 #include <boost/assert.hpp>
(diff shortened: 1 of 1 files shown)


daniel-j-h:
This no longer crashes routed. There is the possibility that we return "empty" trips as in
https://gist.github.com/daniel-j-h/a7de7e143b2a063805cbe95fe30aa1e3

Is this the expected behavior for Trip with e.g. two times the same coordinates (read: two locations that snap to the same coordinates)?

Note also I only worked on the Farthest Insertion algorithm, I have not yet looked at others. Should we ticket that up as potential problems? I don't feel like I want to spent a lot of time with this at the moment.

daniel-j-h:
Also needs trivial v5 backport / cherry-pick when this is ready to be merged.

daniel-j-h:
I'm also curious why I can not reproduce this on v5 --- did we change something in the snapping code? That is, the fix should definitely be backported to v5, the coordinates in the original issue just do not trigger the issue.

TheMarex:
> did we change something in the snapping code?

In general yes, but I didn't remember that I had to pull this in the trip plugin as well.

Going to hold off merging here because I need to coordinate the merge with a rebase of `v5`. Most likely EOD.

PR #2177, opened 2016-03-30
Fix raster_source + tests

lbud:
* scripting_environment now casts coordinates' `lat` and `lon` properties to doubles, so this modifies the function signature for query functions called from lua profiles
* adds forward_ and backward_mode to raster test profiles (required to return a valid edge)
* URLs were being called "malformed" with hyphens in them, so I had to rename `rasterbot-interp.lua` to `rasterbotinterp.lua`

@TheMarex — :eyes: & :+1: ?


TheMarex:
:+1: going to merge

Issue #2163, opened 2016-03-29
Dropping POST support in 5.0


TheMarex:
URL based POST encoding doesn't work with the new API anymore, since the coordinates are part of the path now, e.g. `http://127.0.0.1/route/v1/driving/lon,lat;lon,lat;lon,lat?option=value`. As such there is little sense to remove the (short) option part to POST.

Since we can't support two APIs that means POST is out for `osrm-routed`. For big requests we have been advocating using the C++ library directly or using `node-osrm` since some time now. As such this is conscious removal of a feature.

- [x] Remove POST tests
- [x] Remove code that handles POST

daniel-j-h:
I can take a look at that.

Also for the record: here is an example project integrating `libosrm` with a battle-tested HTTP server (e.g. providing KeepAlive amongst other features). That should get you started:

https://github.com/daniel-j-h/libosrm-http-casablanca

I wanted to do the same with Facebook's Proxygen but haven't had the time so far.

kostadin24:
What about distance matrix? Without POST it is not possible to solve matrix with 5000 elements.
URL become too long.

daniel-j-h:
You can either split your request and re-assemble the responses (I _think_ I saw a project that does that transparently), or use the [node-osrm](https://github.com/Project-OSRM/node-osrm) bindings which we recommend for production use anyway, or use `libosrm` for table queries without the HTTP overhead.

----------------------------------------------------------------------------------------------------------------

Here's a quick brain dump of how to use `libosrm` in the upcoming v5 release (currently in branch `rewrite/new-api`). I will add this and more details to the Wiki at some point in time.

Take a look at the [example code (that lives in example/example.cc)](https://github.com/Project-OSRM/osrm-backend/blob/ca0d21eca9562b3d54568052be6be350c707d03e/example/example.cpp#L30). Here is all you ever wanted to know about `libosrm`, that is a short description of what the types do and where to find documentation on it:

- [`EngineConfig`](https://github.com/Project-OSRM/osrm-backend/blob/ca0d21eca9562b3d54568052be6be350c707d03e/include/engine/engine_config.hpp#L43-L53) - for initializing an OSRM instance we can configure certain properties and constraints. E.g. the storage config is the base path such as `france.osm.osrm` from which we derive and load `france.osm.osrm.*` auxiliary files. This also lets you set constraints such as the maximum number of locations allowed for specific services.

[...]

PR #2182, opened 2016-03-31
Removes server-side support for POST requests on v5

daniel-j-h:
Resolves https://github.com/Project-OSRM/osrm-backend/issues/2163.

@TheMarex we seriously need to talk about routed. I just took a deeper look at the request parser and we have a freaking character-level state machine in there. :wat:

I can totally see this coming from the [Boost.ASIO examples](http://www.boost.org/doc/libs/1_55_0/doc/html/boost_asio/examples/cpp11_examples.html).

Just compare the following two files with ours:

- http://www.boost.org/doc/libs/1_55_0/doc/html/boost_asio/example/cpp11/http/server/request_parser.hpp
- http://www.boost.org/doc/libs/1_55_0/doc/html/boost_asio/example/cpp11/http/server/request_parser.cpp

This is not a priority with v5 or any release in particular but we need a clear upgrade path. And I'm fine with removing our hand-written C++ code e.g. with a node-using routed (or anything more serious, heck look at [how beautiful it could be](https://nghttp2.org/documentation/libnghttp2_asio.html#server-api)).

### DIFF
--- a/include/server/request_parser.hpp
+++ b/include/server/request_parser.hpp
@@ -64,17 +64,11 @@ class RequestParser
         space_before_header_value,
         header_value,
         expecting_newline_2,
-        expecting_newline_3,
-        post_O,
-        post_S,
-        post_T,
-        post_request
+        expecting_newline_3
     } state;
 
     http::header current_header;
     http::compression_type selected_compression;
-    bool is_post_header;
-    int content_length;
 };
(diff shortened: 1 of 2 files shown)


TheMarex:
:+1: looks good to me. Rebasing and merging.

PR #2185, opened 2016-04-01
fixes via-feature issues, including an error in post-processing

MoKob:
Minor error in post-processing that can end up removing a coordinate that should not be removed.


MoKob:
wrong base branch

PR #2186, opened 2016-04-01
Fixes Via Feature / Post Processing Bug triggered in it

MoKob:
Minor error in post-processing that can end up removing a coordinate that should not be removed.


MoKob:
superseeded by https://github.com/Project-OSRM/osrm-backend/pull/2187

Issue #2173, opened 2016-03-30
Coordinate parsing should allow integers


miccolis:
Currently a path like `/route/v1/bicycle/80,80;80,0.json?steps=true&alternatives=false` will get `{ code: 'InvalidQuery', message: 'Query string malformed close to position 11’ }`

I noticed this when running `./node_modules/.bin/cucumber.js features/testbot/planetary.feature `

daniel-j-h:
Here's a small self-contained reproducible test-case:

```c++
#include <vector>
#include <string>
#include <iterator>
#include <iostream>

#include <boost/spirit/include/qi.hpp>

using namespace boost::spirit;
using namespace boost::spirit::qi;

int main() {
  const std::string input{"4.0.json"};

  auto first = begin(input);
  const auto last = end(input);

  std::vector<double> into;

  const auto ok = parse(first, last,                  //
                        (int_ | double_) >> ".json",  //
                        into);

  if (ok && first == last) {
    std::cout << "Ok, got: ";
    for (auto&& each : into) std::cout << each << " ";
    std::cout << "\n";
  } else {
    const auto pos = std::distance(first, last);
    std::cerr << "Error at position: " << pos << "\n\n";
    std::cerr << input << "\n" << std::string(pos, ' ') << "^\n";
  }
}
```

Output:

```
Error at position: 8

4.0.json
        ^
```

daniel-j-h:
For the record, this is what I got so far going down the Spirit :rabbit: hole:

For parsing `1,2;3,4.json` ideally we could use `((double_ % ",") % ";") >> ".json"`.

Spirit's `parser % seperator` DSL expands to `parser >> *(seperator >> parser)`. We can simplify the test case to a trivial `double_ >> ".json"`; once this works the other pieces just fall into place.

`double_ >> ".json"` parses `4.1.json` fine but fails on `4.json` since the primitive `double_` parser is too greedy.

Spirit's DSL maps to PEG (Parsing Expression Grammar); it may look like EBNF but the choice operator `|` is evaluated from left to right in PEG. So the next logical step is to make use of this:

Unfortunately `(int_ | double_) >> ".json"` also fails to parse `4.json`.

danpat:
@daniel-j-h Tried `(double_ | int_) >> ".json"` ?

Assuming those are tried in-order, maybe that will work....

daniel-j-h:
Nope does not help either. Currently looking into Spirit v2 details and reaching out to some guys.

[...]

Issue #2168, opened 2016-03-30
Request handler's position hint in error reporting is broken


daniel-j-h:
Breaking out of https://github.com/Project-OSRM/osrm-backend/issues/2164: the [iterator pointing to the position where parsing failed](https://github.com/Project-OSRM/osrm-backend/blob/ca0d21eca9562b3d54568052be6be350c707d03e/src/server/request_handler.cpp#L78-L79) is never being incremented when parsing actually fails:

```json
daniel@x1c /t/o/build> curl 'http://127.0.0.1:5000/Xroute/v1/driving/'
{"message":"URL string malformed close to position 0: \"v\"\"\/Xr\"","code":"InvalidUrl"}

daniel@x1c /t/o/build> curl 'http://127.0.0.1:5000/route/Xv1/driving/'
{"message":"URL string malformed close to position 0: \"\/ro\"","code":"InvalidUrl"}

daniel@x1c /t/o/build> curl 'http://127.0.0.1:5000/route/v1/driving/'
{"message":"URL string malformed close to position 0: \"vin\/ro\"","code":"InvalidUrl"}
```

The [code above (`parseURL`) dispatches to Spirit's `parse`](https://github.com/Project-OSRM/osrm-backend/blob/ca0d21eca9562b3d54568052be6be350c707d03e/src/server/api/url_parser.cpp#L77) with its documentation here:

http://www.boost.org/doc/libs/1_55_0/libs/spirit/doc/html/spirit/qi/reference/parse_api/iterator_api.html#spirit.qi.reference.parse_api.iterator_api.synopsis

We pass the first iterator by reference, so it seems like we're using the interface how it was intended.

Expected behavior: for

```json
curl 'http://127.0.0.1:5000/route/Xv1/driving/13.388860,52.517037;13.397634,52.529407'
```

the iterator points to the "X" (or surrounding) from which we can construct the `context` string object.

daniel-j-h:
Small self-contained reproducible snippet:

```c++
#include <vector>
#include <string>
#include <iterator>
#include <iostream>

#include <boost/spirit/include/qi.hpp>

using namespace boost::spirit;
using namespace boost::spirit::qi;

int main() {
  const std::string input{"/route/Xv1/driving/"};

  auto first = begin(input);
  const auto last = end(input);

  std::string service;
  unsigned version;
  std::string profile;

  const auto ok = parse(first, last,                //
                        "/" >> +alpha               //
[...]

PR #2188, opened 2016-04-01
[Not Ready] Fix Parsers for v5

daniel-j-h:
Working on 
- [x] https://github.com/Project-OSRM/osrm-backend/issues/2168
- [X] https://github.com/Project-OSRM/osrm-backend/issues/2173

WIP.

### DIFF
--- a/src/server/api/url_parser.cpp
+++ b/src/server/api/url_parser.cpp
@@ -62,7 +62,7 @@ boost::optional<ParsedURL> parseURL(std::string::iterator &iter, const std::stri
 {
     using It = std::decay<decltype(iter)>::type;
 
-    static URLParser<It, ParsedURL> const parser;
+    static URLParser<It, ParsedURL()> const parser;
     ParsedURL out;
 
     const auto ok = boost::spirit::qi::parse(iter, end, parser, out);
--- a/src/server/api/url_parser.cpp
+++ b/src/server/api/url_parser.cpp
@@ -55,9 +55,10 @@ struct URLGrammar : boost::spirit::qi::grammar<Iterator>
         profile_rule = +alpha_numeral;
         query_rule = +all_chars;
 
-        url_rule = qi::lit('/') >> service_rule[set_service] >> qi::lit('/') >> qi::lit('v') >>
-                   version_rule[set_version] >> qi::lit('/') >> profile_rule[set_profile] >>
-                   qi::lit('/') >> query_rule[set_query];
+        url_rule = qi::lit('/') >> service_rule[set_service]                    //
+                   >> qi::lit('/') >> qi::lit('v') >> version_rule[set_version] //
+                   >> qi::lit('/') >> profile_rule[set_profile]                 //
+                   >> qi::lit('/') >> query_rule[set_query];
     }
 
     ParsedURL parsed_url;
--- a/src/server/api/url_parser.cpp
+++ b/src/server/api/url_parser.cpp
@@ -62,7 +62,7 @@ boost::optional<ParsedURL> parseURL(std::string::iterator &iter, const std::stri
 {
     using It = std::decay<decltype(iter)>::type;
 
-    URLParser<It, ParsedURL> parser;
+    static URLParser<It, ParsedURL> const parser;
     ParsedURL out;
 
     const auto ok = boost::spirit::qi::parse(iter, end, parser, out);
--- a/src/server/api/url_parser.cpp
+++ b/src/server/api/url_parser.cpp
@@ -1,17 +1,55 @@
 #include "server/api/url_parser.hpp"
-
 #include "engine/polyline_compressor.hpp"
 
-#include <boost/bind.hpp>
-#include <boost/spirit/include/qi_char_.hpp>
-#include <boost/spirit/include/qi_grammar.hpp>
-#include <boost/spirit/include/qi_uint.hpp>
-#include <boost/spirit/include/qi_real.hpp>
-#include <boost/spirit/include/qi_lit.hpp>
-#include <boost/spirit/include/qi_action.hpp>
-#include <boost/spirit/include/qi_as_string.hpp>
-#include <boost/spirit/include/qi_operator.hpp>
-#include <boost/spirit/include/qi_plus.hpp>
+//#define BOOST_SPIRIT_DEBUG
+#include <boost/spirit/include/qi.hpp>
+
+#include <string>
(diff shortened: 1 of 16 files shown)


daniel-j-h:
@TheMarex today @artemp told me this about grammars:

> grammar should be/can be shared between threads and should not hold any state

Storing results (e.g. parameters in the parameters parser) in the grammar is a really bad idea.

daniel-j-h:
I started resolving the grammar issues, here are a few tricks I like to share which makes working with Spirit so much easier. Maybe that's of interest to you @TheMarex:

1/

    #define BOOST_SPIRIT_DEBUG

provides output such as

```
<start>
  <try>/route/Xv1/driving/7</try>
  <service>
    <try>route/Xv1/driving/7.</try>
    <success>/Xv1/driving/7.41635</success>
    <attributes>[[r, o, u, t, e]]</attributes>
  </service>
  <fail/>
</start>
```

for rules registered with

    BOOST_SPIRIT_DEBUG_NODES

2/ 

Spirit is able to co-operate with Fusion, i.e. transparently parsing into a Sequence (Fusion concept) is as easy as: 1/ [adapt type to Sequence concept](https://github.com/Project-OSRM/osrm-backend/blob/ab28360fab497a7433216c86176d8a2962d1b239/include/server/api/parsed_url.hpp#L30) (enables for compile time introspection, with no runtime costs; useful for for things like [this](https://github.com/daniel-j-h/argparse-generic)) 2/ [provide type representation parser](https://github.com/Project-OSRM/osrm-backend/blob/ab28360fab497a7433216c86176d8a2962d1b239/src/server/api/url_parser.cpp) 3/ profit.

daniel-j-h:
Is not ready but in a mergeable state in case you want the improvements in the rc @TheMarex.

TheMarex:
@daniel-j-h can you scope what is missing here? The expectation point fix?

src/server/api/url_parser.cpp
```diff
@@ -1,88 +1,78 @@
...
-#include <boost/bind.hpp>
-#include <boost/spirit/include/qi_char_.hpp>
-#include <boost/spirit/include/qi_grammar.hpp>
-#include <boost/spirit/include/qi_uint.hpp>
-#include <boost/spirit/include/qi_real.hpp>
-#include <boost/spirit/include/qi_lit.hpp>
-#include <boost/spirit/include/qi_action.hpp>
-#include <boost/spirit/include/qi_as_string.hpp>
-#include <boost/spirit/include/qi_operator.hpp>
-#include <boost/spirit/include/qi_plus.hpp>
```
TheMarex: In my experience specifying features explicitly cuts down on memory usage and compile time, especially since we have a lot of grammars now.

daniel-j-h:
- [ ] expectation points for all sequence `>>`-using parsers
- [ ] add the `.json` fix to all parameter grammars


Strictly speaking we have to make all grammars stateless; I started this with the URL parser grammar, as I was implementing fixes for it anyway. But doing this properly for all additional gramars is probably too much for this pull request.

src/server/api/url_parser.cpp
daniel-j-h: In theory yes. In practice, the grammar is only used in this single cpp file and therefore is TU-local.

For grammars in headers what helps is explicitely instantiating the grammar as in http://stackoverflow.com/a/5507332.

daniel-j-h:
https://github.com/Project-OSRM/osrm-backend/issues/2168 is solved with this. Strictly speaking what's missing is
- expectation points for other grammars (`*_parameters_grammar.hpp`)
- `radiuses=` problem, broken out in https://github.com/Project-OSRM/osrm-backend/pull/2231 since it currently segfaults

@TheMarex this ready for merging in case you want to speed up pushing a RC 2 out.

daniel-j-h:
https://github.com/Project-OSRM/osrm-backend/pull/2188/commits/e48fd7c08adcdee3cab49a7c3a4d93f95ab1e2d3 adapts grammars to expectation parsers. The only problem we're still having is the parameter parsers not knowing about the url part that's getting parsed in the url parser. That is, in a query:

    http 'http://localhost:5000/table/v1/driving/7.416351,43.731205;7.420363,43.736189?radiuses=unlimiteX;5'

the parameters parser only gets to see

    7.416351,43.731205;7.420363,43.736189?radiuses=unlimiteX;5

We now report position 47, correctly pointing out `=unlimiteX` _if you start counting after the base url path_, since this gets already split off in the url parser.

daniel-j-h:
Rebased onto v5. There are some unit tests failing, but after a quick look it seems like they always test the failure to be at position 0. That is, the unit tests are simply wrong. Will adapt them.

- [ ] fix unit tests

src/server/api/url_parser.cpp
daniel-j-h: For the record, changing from explicitly specifying Qi headers to the single Qi header in the `*_grammars.hpp` does not result in measurable increased compile times.

daniel-j-h:
Done. /cc @TheMarex

src/server/api/url_parser.cpp
springmeyer: I've found the same thing in Mapnik. So for simplicity in mapnik we only include `qi.hpp` since individual includes are no faster and are a pain to maintain. Plan to revisit this in spirit x3.

daniel-j-h: Great to hear you're getting the same results. Spirit v2 is really hard for getting above a certain level of understanding, it's almost as if I have to learn by failing. Had some hard times the last two days.

I took a look at x3 for a few side-projects and their approach to always generate an AST and move away from semantic actions and grammar is refreshing.

daniel-j-h:
Hrm a single cucumber test is failing (thanks Travis):
https://travis-ci.org/Project-OSRM/osrm-backend/jobs/121704945#L1725

daniel-j-h:
Especially this is concerning:

    http 'http://localhost:5000/nonsense' | jq '.message'
    "URL string malformed close to position 9: \"nse\""

https://github.com/Project-OSRM/osrm-backend/blob/url-parser-for-v5/src/server/request_handler.cpp#L114

TheMarex:
Looking into the failing tests only the one you mentioned seems actually strange. However I think this stems from the fact that it would be expecting an `/` at that position to make sense. I feel thse error messages are already a lot better and fine with me.

I'm fixing and then merging.

Issue #2150, opened 2016-03-25
Wrong car routing through motor_vehicle=delivery


marcows:
See http://map.project-osrm.org/?z=17&center=49.318213%2C8.572083&loc=49.316863%2C8.575462&loc=49.316517%2C8.574700&hl=en&alt=0
The two streets connecting each service area with the street over the motorway are tagged with motor_vehicle=delivery, so they shouldn't be used.

Further info:
https://lists.openstreetmap.org/pipermail/talk/2016-February/075526.html

See also @Fujnky's [comment](https://github.com/Project-OSRM/osrm-backend/issues/256#issuecomment-145653584) on issue #256 illustrating a similar problem.

daniel-j-h:
Thank you for reporting this. I looked into it and this is what I found:

If you take a look at the [car profile](https://github.com/Project-OSRM/osrm-backend/blob/0d291cb68ec5cb800b71856382963e55c8e7a98a/profiles/car.lua) you can see `access=motor_vehicle` in the whitelist, so we're currently routing over it unconditionally. `motor_vehicle=delivery` is something different. Here's its [taginfo](http://taginfo.openstreetmap.org/keys/motor_vehicle).

If you take a look at the [`access_tag_restricted`](https://github.com/Project-OSRM/osrm-backend/blob/0d291cb68ec5cb800b71856382963e55c8e7a98a/profiles/car.lua#L9) list you can find `access=delivery` in there. But `access_tag_restricted` is not used at the moment and will only be used in the future to augment the route response e.g. for guidance.

marcows:
`access=motor_vehicle` is something different, we only talk about `motor_vehicle` as a key, not as a value (which catches wrong mapping like `access=motor_vehicle` I guess - or are these preprocessed tags, not directly related to OSM tags?).

I just tried and learned OSRM also routes through roads with `motor_vehicle=destination`. Is this on purpose because detection of whether the routing target is inside a destination area is not implemented yet?

I've read the car profile lua code (which I only skimmed through before writing this issue) and conclude that `access_tag_whitelist` only has any influence for highway values not included in `speed_profile` (highway=track or unknown highway values) - and then only on the speed, not on access. For prohibiting access, `access_tag_blacklist` has to contain the concerned access value.

> But access_tag_restricted is not used at the moment and will only be used in the future to augment the route response e.g. for guidance.

I don't understand what "to augment the route response e.g. for guidance" means.

[...]

PR #2202, opened 2016-04-04
Removes unused access_tag list from car profile

daniel-j-h:
Looking into the profiles for https://github.com/Project-OSRM/osrm-backend/issues/2150#issuecomment-203595229 I found
`access_tags` being neither used in the profile, nor
through luabind. It is dead code.

Using my time machine (which sadly can only go back
in history), I found the introducing commit here:

https://github.com/Project-OSRM/osrm-backend/commit/74cc50f52b79cdb06d5c61fabc1fdc9ca5979fcc#diff-ded22cbfae0ae88f6359ccff1c3355b0R106

Seems like we no longer use it, and it was still left in the car profile.


TheMarex:
Good catch, ready to merge.

PR #2203, opened 2016-04-04
Spelling: Hierachy vs Hierarchy

daniel-j-h:
Seen while looking at the profiles for https://github.com/Project-OSRM/osrm-backend/issues/2150#issuecomment-203595229.


TheMarex:
It is a hard word to spell. :+1: Should be trivially mergable on `develop` as well.

PR #2204, opened 2016-04-04
Do not route over delivery ways and nodes --- v5

daniel-j-h:
See https://github.com/Project-OSRM/osrm-backend/issues/2150.

PR #2205, opened 2016-04-04
Do not route over delivery ways and nodes --- v4 backport

daniel-j-h:
Cherry-picked! :cherries:

See https://github.com/Project-OSRM/osrm-backend/issues/2150, https://github.com/Project-OSRM/osrm-backend/pull/2204.


TheMarex:
I think we are good to merge this in `develop` and then rebase.

daniel-j-h:
Yep, I forgot we already have the cucumber tests in develop and thought there might be something different compared to v5. But it's trivially mergeable.

PR #2210, opened 2016-04-04
Suppress pedantic ISO C++ warning about argv[0] size check

daniel-j-h:
Boost.Filesystem immediately does a size check in the path headers, triggering this.


TheMarex:
wow that was a weird one. :+1: for merge

Issue #2215, opened 2016-04-04
Disable steps and alternatives by default


TheMarex:
We inherited the default setting of `steps=true` and `alternatives=true` from upstream. In practice we see that most people don't actually use the steps or alternative, but won't bother to change the default settings. That means the server does a lot of computational work for nothing.

Proposal:
- `steps=false` by default
- `alternatives=false` by default

/cc @freenerd

daniel-j-h:
Done in https://github.com/Project-OSRM/osrm-backend/issues/2217.

PR #2217, opened 2016-04-05
Disable steps and alternatives by default

daniel-j-h:
For https://github.com/Project-OSRM/osrm-backend/issues/2215.

I can see the following issues still:
- [x] before we asserted on an [alternative being found](https://github.com/Project-OSRM/osrm-backend/blob/rewrite/new-api/unit_tests/library/route.cpp#L105-L108), now although I set [`alteratives=true`](https://github.com/Project-OSRM/osrm-backend/blob/disable-steps-alternatives-by-default/unit_tests/library/route.cpp#L74-L76) the [assertion for an alternative triggers](https://github.com/Project-OSRM/osrm-backend/blob/disable-steps-alternatives-by-default/unit_tests/library/route.cpp#L110-L111)

(Note: `server-tests` are failing, but I checked `rewrite/new-api` and they are already failing there)

### DIFF
--- a/features/step_definitions/trip.js
+++ b/features/step_definitions/trip.js
@@ -110,6 +110,7 @@ module.exports = function () {
                 } else {
                     var params = this.queryParams,
                         waypoints = [];
+                    params['steps'] = 'true';
                     if (row.from && row.to) {
                         var fromNode = this.findNodeByName(row.from);
                         if (!fromNode) throw new Error(util.format('*** unknown from-node "%s"', row.from));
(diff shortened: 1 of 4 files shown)


daniel-j-h:
Github has some hiccups at the moment (https://status.github.com/messages); I pushed a commit fixing the remaining issue above, but it's not visible via the Github user interface (yet?).

Alternatives are completely random, we can not assert on one being found, not even for our fixed Monaco dataset.

TheMarex:
Looks good to me. :+1:

Issue #2009, opened 2016-02-24
Boost Spirit's floating point parsers happily accept `inf`, `nan` and related by default


daniel-j-h:
Quick and dirty proof of concept (with X3, same applies to Spirit 2/Qi):

```c++
#include <string>
#include <vector>
#include <iostream>
#include <iterator>

#include <boost/spirit/home/x3.hpp>
#include <boost/spirit/home/x3/version.hpp>

static_assert(SPIRIT_X3_VERSION >> 12u == 3u, "Boost.Spirit 3.x.x required");

int main(int argc, char** argv) {
  namespace x3 = boost::spirit::x3;

  const std::vector<std::string> args{argv, argv + argc};

  if (args.size() != 2)
    return EXIT_FAILURE;

  auto first = begin(args[1]);
  auto last = end(args[1]);

  const auto parser = x3::double_;
  double into;
  const auto ok = x3::parse(first, last, parser, into);

  if (first == last and ok)
    std::cout << into << std::endl;
  else
    return EXIT_FAILURE;
}
```

Triggers with

    ./a.out nan
    ./a.out inf
    ./a.out -infinity
    ./a.out -nan

Here's the [`RealPolicies`](http://www.boost.org/doc/libs/1_60_0/libs/spirit/doc/html/spirit/qi/reference/numeric/real.html#spirit.qi.reference.numeric.real._code__phrase_role__identifier__realpolicies__phrase___code_): we should probably provide our own and _at least_ disallow `parse_exp(f, l)`, `parse_exp_n(f, l, n)`, `parse_nan(f, l, n)`, `parse_inf(f, l, n)`.

I quickly tested a few of those edge cases against routed and _it seems like_ we're triggering a invalid coordinate path by means of `PluginBase::check_all_coordinates`, which dispatches to `FixedPointCoordinate.IsValid()`. I would still prefer disabling parsing those as valid coordinates already in the api grammars.

References:
- https://github.com/boostorg/spirit/blob/develop/test/x3/real1.cpp
- https://github.com/boostorg/spirit/blob/develop/test/x3/real2.cpp

lbud:
> it seems like we're triggering a invalid coordinate path by means of `PluginBase::check_all_coordinates`, which dispatches to `FixedPointCoordinate.IsValid()`. I would still prefer disabling parsing those as valid coordinates already in the api grammars.

[...]

PR #2222, opened 2016-04-05
Fix/2173/no trailing dotjson policy

oxidase:
Hello, 
it is another type fix  for #2173 with a customized real policy: it drops real number parsing if finds ".json" and returns an integer part if got_a_number is true otherwise rule fails.  Output here:
```
http 'http://localhost:5000/route/v1/driving/7.416351,43.731205;7.420363,43.736189;7.420363,43.736189;80,0...json'   | jq '.code'
"InvalidQuery"
http 'http://localhost:5000/route/v1/driving/7.416351,43.731205;7.420363,43.736189;7.420363,43.736189;80,0..json'   | jq '.code'
"Ok"
http 'http://localhost:5000/route/v1/driving/7.416351,43.731205;7.420363,43.736189;7.420363,43.736189;80,0.json'   | jq '.code'
"Ok"
http 'http://localhost:5000/route/v1/driving/7.416351,43.731205;7.420363,43.736189;7.420363,43.736189;80,0json'   | jq '.code'
"InvalidQuery"
 http 'http://localhost:5000/route/v1/driving/7.416351,43.731205;nan,-inf;7,.736189;1e2,2.json'   | jq '.code'
"InvalidValue"
http 'http://localhost:5000/route/v1/driving/7.416351,43.731205;7,.736189;1e2,2.json'   | jq '.code'
"Ok"
```

But it will fail if the last entry has an exponent part
```
http 'http://localhost:5000/route/v1/driving/7.416351,43.731205;7,.736189;1e2,2e4.json'   | jq '.code'
"InvalidValue"
```

### DIFF
--- a/include/server/api/parsed_url.hpp
+++ b/include/server/api/parsed_url.hpp
@@ -27,6 +27,11 @@ struct ParsedURL final
 } // server
 } // osrm
 
-BOOST_FUSION_ADAPT_STRUCT(osrm::server::api::ParsedURL, service, version, profile, query)
+BOOST_FUSION_ADAPT_STRUCT(osrm::server::api::ParsedURL,
+    (std::string, service),
+    (unsigned, version),
+    (std::string, profile),
+    (std::string, query)
+)
 
 #endif
--- a/include/server/api/base_parameters_grammar.hpp
+++ b/include/server/api/base_parameters_grammar.hpp
@@ -29,10 +29,51 @@ namespace api
 {
 
 namespace qi = boost::spirit::qi;
+
+template <typename T, char... Fmt>
+struct no_trailing_dot_policy : qi::real_policies<T> {
+
+    template <typename Iterator>
+    static bool parse_dot(Iterator& first, Iterator const& last) {
+        if (first == last || *first != '.')
+            return false;
+
+        static const constexpr char fmt[sizeof...(Fmt)] = {Fmt...};
+
+        if (first + sizeof(fmt) < last
+            && std::equal(fmt, fmt + sizeof(fmt), first + 1u))
+            return false;
+
+        ++first;
+        return true;
+    }
+
+    template <typename Iterator>
+    static bool parse_exp(Iterator&, const Iterator&) {
+        return false;
+    }
+
+    template <typename Iterator, typename Attribute>
+    static bool parse_exp_n(Iterator&, const Iterator&, Attribute&) {
+        return false;
+    }
+
+    template <typename Iterator, typename Attribute>
+    static bool parse_nan(Iterator&, const Iterator&, Attribute&) {
+        return false;
+    }
+
+    template <typename Iterator, typename Attribute>
+    static bool parse_inf(Iterator&, const Iterator&, Attribute&) {
+        return false;
+    }
+};
+
 struct BaseParametersGrammar : boost::spirit::qi::grammar<std::string::iterator>
 {
     using Iterator = std::string::iterator;
     using RadiusesT = std::vector<boost::optional<double>>;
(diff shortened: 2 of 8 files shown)


daniel-j-h:
> But it will fail if the last entry has an exponent part

Strictly speaking we have to provide a custom real policy for inf, nan and related anyway: https://github.com/Project-OSRM/osrm-backend/issues/2009

include/server/api/parsed_url.hpp
```diff
@@ -27,6 +27,11 @@ struct ParsedURL final
 } // server
 } // osrm
 
-BOOST_FUSION_ADAPT_STRUCT(osrm::server::api::ParsedURL, service, version, profile, query)
+BOOST_FUSION_ADAPT_STRUCT(osrm::server::api::ParsedURL,
+    (std::string, service),
+    (unsigned, version),
+    (std::string, profile),
+    (std::string, query)
+)
```
daniel-j-h: I see, this is needed for older Boost versions. Good catch!

daniel-j-h:
Looks good, the only problem I'm seeing is coupling the format with a policy. When we want to support multiple formats we now have to provide multiple formats / a generic policy.

daniel-j-h:
@oxidase what do you think of

```c++
template <typename T, char... Fmt>
struct no_trailing_dot_policy : qi::real_policies<T> {

  template <typename Iterator>
  static bool parse_dot(Iterator& first, Iterator const& last) {
    if (first == last || *first != '.') return false;

    static const constexpr char fmt[sizeof...(Fmt)] = {Fmt...};

    if (first + sizeof(fmt) < last &&
        std::equal(fmt, fmt + sizeof(fmt), first + 1u))
      return false;

    ++first;
    return true;
  }

  template <typename Iterator>
  static bool parse_exp(Iterator&, const Iterator&) {
    return false;
  }

  template <typename Iterator, typename Attribute>
  static bool parse_exp_n(Iterator&, const Iterator&, Attribute&) {
    return false;
  }

  template <typename Iterator, typename Attribute>
  static bool parse_nan(Iterator&, const Iterator&, Attribute&) {
    return false;
  }

  template <typename Iterator, typename Attribute>
  static bool parse_inf(Iterator&, const Iterator&, Attribute&) {
    return false;
  }
};
```

and then providing convenience aliases such as

```c++
using json_policy = no_trailing_dot_policy<double, 'j', 's', 'o', 'n'>;
```

which makes it easy to extend in the future.

Also when we provide a custom policy we can solve https://github.com/Project-OSRM/osrm-backend/issues/2009 quite nicely.

TheMarex:
Sorry for the rebase @oxidase, didn't see your PR open.

Can you do:
```
git remote add upstream https://github.com/Project-OSRM/osrm-backend.git
git checkout rewrite/new-api
git fetch upstream
git reset --hard upstream/rewrite/new-api
git checkout fix/2173/no_trailing_dotjson_policy
git rebase -i rewrite/new-api
```

Sorry for the trouble.

oxidase:
@daniel-j-h  it looks very good for me.
I was not aware of #2009, but imho stricter parsing policy would be better than parsing nans and  throwing an exception instead of rejecting during the parsing stage.

@TheMarex ok, no problem.  I have updated my pull request into url-parser-for-v5 branch.

include/server/api/parsed_url.hpp
oxidase: Yes, for libboost1.58-dev:amd64 parser returns empty strings and server responses "InvalidService"

daniel-j-h:
Merged! Perfect @oxidase I appreciate your effort! :+1:

PR #2225, opened 2016-04-06
The C compiler is used for several CMake tests, match with C++ compiler

daniel-j-h:
Otherwise we have CMake detecting gcc 4.8 as C compiler and does some compilation checks with it, but then uses e.g. clang 3.8 for actual compilation :boom:.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -28,7 +28,7 @@ matrix:
         apt:
           sources: ['ubuntu-toolchain-r-test']
           packages: ['g++-5', 'libbz2-dev', 'libstxxl-dev', 'libstxxl1', 'libxml2-dev', 'libzip-dev', 'lua5.1', 'liblua5.1-0-dev', 'libtbb-dev', 'libgdal-dev', 'libluabind-dev', 'libboost-all-dev']
-      env: COMPILER='g++-5' BUILD_TYPE='Debug'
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -104,22 +104,22 @@ matrix:
...
     #  osx_image: xcode6.4
     #  compiler: clang
-    #  env: COMPILER='clang++' BUILD_TYPE='Release' BUILD_SHARED_LIBS=ON
+    #  env: CCOMPILER='clang' CXXCOMPILER='clang++' BUILD_TYPE='Release' BUILD_SHARED_LIBS=ON
 
     #- os: osx
     #  osx_image: xcode7
     #  compiler: clang
-    #  env: COMPILER='clang++' BUILD_TYPE='Release' BUILD_SHARED_LIBS=ON
+    #  env: CCOMPILER='clang' CXXCOMPILER='clang++' BUILD_TYPE='Release' BUILD_SHARED_LIBS=ON
```
daniel-j-h: What's the plan with those OSX tests?

TheMarex: We disabled shared library builds on OSX because 1/ nobody uses that and it increases build time and 2/ it had some weird linking issues that nobody wanted to debug.

TheMarex: Ah I see there a few `xcode 6.3` based builds. that is super outdated and can go.

Issue #2197, opened 2016-04-02
Release 5.0.0 RC1


TheMarex:
```
E-Mail

We pushed out the first RC for 5.0.0. As the version indicates this release breaks the backwards compatibility of the public API. The HTTP API was completely rewritten and is now a little bit less verbose. We refactored the C++ library interface as well, making it a lot easier to consume. Explore `example/example.cpp` on how to use it. Please let us know if you find any bugs that go beyond the known issues below.

- Renamed osrm-prepare into osrm-contract
- osrm-contract does not need a profile parameter anymore
- New public HTTP API, find documentation [here](https://github.com/Project-OSRM/osrm-backend/wiki/New-Server-api)
- POST support is discontinued, please use library bindings for more complex requests
- Removed timestamp plugin
- Coordinate order is now Longitude,Latitude
- Cucumber tests now based on Javascript (run with `npm test`)
- Profile API changed:
   - `forward_mode` and `backward_mode` now need to be selected from a pre-defined list
   - Global profile properties are now stored in a global `properties` element. This includes:
     - `properties.traffic_signal_penalty`
     - `properties.use_turn_restrictions`
     - `properties.u_turn_penalty`
     - `properties.allow_u_turn_at_via`

Known issues:
- Units uncovered that zero-length trips might have a duration bigger than zero #2134
- pkg-config does not work as expected, include paths might be insufficient #2195
- Position hint on requests with syntax error might be off #2193
- Queries with integer coordinates might fail to parse if `.json` is added #2173

As always this release is in sync with a `node-osrm` release. `npm install https://github.com/Project-OSRM/node-osrm/archive/v5.0.0-rc.1.tar.gz` (we don't publish pre-releases because of he semantic versioning bug in npm were 4.9.1 < 5.0.0-rc.1, that means new users would suddenly download the RC)

This release was made possible by: @daniel-j-h, @danpat, @karenzshea, @lbud, @miccolis, @mokob, @springmeyer, @themarex, @willwhite

We expect to push out at least another RC before we tag `5.0.0`.

Cheers,
Patrick
```

- [x] Tag `osrm-backend` `v5.0.0-rc.1`
- [x] Tag `node-osrm` `v5.0.0-rc.1`

TheMarex:
Getting a link error while building node-osrm, delaying until monday, if someone doesn't figure it out.

springmeyer:
I see two new link errors.

1) boost_test_main

[...]

Issue #2065, opened 2016-03-08
Avoid unnecessary linking to libboost_unit_test_framework.so 


springmeyer:
The command line tools all link libboost_unit_test_framework even though it is not used by osrm-extract, osrm-contract, etc. This is harmless for local development. But when distributing binaries of the command line tools it becomes problematic bloat: unless this unnecessary shared library is also distributed the tools will fail to run.

So, overall it would be ideal to ensure each tool only links the libraries it actually needs. This might also have the added benefit that linking will be faster.

daniel-j-h:
> But when distributing binaries of the command line tools it becomes problematic bloat: unless this unnecessary shared library is also distributed the tools will fail to run.

Really? I can not see `libboost_unit_test_framework.so` in the dynamic sections for `osrm-extract` or even `libosrm.so`:

```
daniel@x1c /t/o/build> readelf -d osrm-extract|ag needed
 0x0000000000000001 (NEEDED)             Shared library: [libc.so.6]
 0x0000000000000001 (NEEDED)             Shared library: [libosrm_extract.so]
 0x0000000000000001 (NEEDED)             Shared library: [libboost_filesystem.so.1.59.0]
 0x0000000000000001 (NEEDED)             Shared library: [libboost_program_options.so.1.59.0]
 0x0000000000000001 (NEEDED)             Shared library: [libboost_system.so.1.59.0]
 0x0000000000000001 (NEEDED)             Shared library: [libtbb.so.2]
 0x0000000000000001 (NEEDED)             Shared library: [libstdc++.so.6]
 0x0000000000000001 (NEEDED)             Shared library: [libgcc_s.so.1]
```

```
daniel@x1c /t/o/build> readelf -d libosrm.so|ag needed
 0x0000000000000001 (NEEDED)             Shared library: [libboost_filesystem.so.1.59.0]
 0x0000000000000001 (NEEDED)             Shared library: [libboost_system.so.1.59.0]
 0x0000000000000001 (NEEDED)             Shared library: [libboost_thread.so.1.59.0]
 0x0000000000000001 (NEEDED)             Shared library: [libpthread.so.0]
 0x0000000000000001 (NEEDED)             Shared library: [librt.so.1]
 0x0000000000000001 (NEEDED)             Shared library: [libstdc++.so.6]
 0x0000000000000001 (NEEDED)             Shared library: [libm.so.6]
 0x0000000000000001 (NEEDED)             Shared library: [libgcc_s.so.1]
[...]

PR #2227, opened 2016-04-06
Start modularizing the CMake buildsystem

daniel-j-h:
The main reason for modularizing the unit tests was to split off
libboost_unit_test_framework from the osrm toolchain binaries.

Now only the unit test binaries link against it, unblocking
binary distribution without unit test library dependencies.

I started this on v4 a couple of weeks ago and hit a -llua5 issue on
Travis. By now v5 diverged quite a bit (especially in the code that I
have to stare at for debugging the issues).

https://github.com/Project-OSRM/osrm-backend/pull/2073

Let's bring this to v5+ only.

This is blocking:

- https://github.com/Project-OSRM/osrm-backend/issues/2065
- https://github.com/Project-OSRM/osrm-backend/issues/2197#issuecomment-204864938

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -161,10 +161,10 @@ script:
       sudo ldconfig
     fi
   - echo "travis_fold:start:UNIT_TESTS"
-  - ./extractor-tests
-  - ./engine-tests
-  - ./util-tests
-  - ./server-tests
+  - ./unit_tests/extractor-tests
+  - ./unit_tests/engine-tests
+  - ./unit_tests/util-tests
+  - ./unit_tests/server-tests
   - echo "travis_fold:end:UNIT_TESTS"
   - popd
   - echo "travis_fold:start:CUCUMBER"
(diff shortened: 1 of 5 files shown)


daniel-j-h:
Lua linking issues are gone :tada:

I think we should ticket modularizing other subprojects like the libraries and tools (similar to how we did it in rtd) for when v5 is releases. Right now this already unblocks the tickets mentioned above.

daniel-j-h:
@springmeyer does this unblock your issues I linked above? That is can you ship v5 with this patch?

TheMarex:
Looks good to me. Going to rebase and then merge.

springmeyer:
> @springmeyer does this unblock your issues I linked above? That is can you ship v5 with this patch?

@daniel-j-h - thanks, looks like it will. I tried to quickly confirm but hit https://github.com/Project-OSRM/osrm-backend/pull/2234 and https://github.com/Project-OSRM/osrm-backend/pull/2233 and not had a change to full test since those changes.

PR #2230, opened 2016-04-07
Add license to storage and storage config exposed in public API, closes #2036

### LINKED ISSUE #2036: License preamble missing in some headers we install
Currently we have a license preamble in _some_ libosrm header. Due to the problems outlined in https://github.com/Project-OSRM/osrm-backend/issues/2031 we install internal headers that do not have a license preamble in it.

daniel-j-h:


TheMarex:
Great. :+1: Merging.

PR #2231, opened 2016-04-07
Add unlimited keyword to radiuses grammar

daniel-j-h:
Splitting this off of my branch, since I can't get it to work properly.

### DIFF
--- a/include/server/api/base_parameters_grammar.hpp
+++ b/include/server/api/base_parameters_grammar.hpp
@@ -8,6 +8,7 @@
 #include "engine/polyline_compressor.hpp"
 
 //#define BOOST_SPIRIT_DEBUG
+#include <boost/optional.hpp>
 #include <boost/spirit/include/qi.hpp>
 
 #include <string>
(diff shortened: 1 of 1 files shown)


daniel-j-h:
Segfaults with

```
boost::spirit::qi::detail::tst_node<char, boost::optional<double> >::find<__gnu_cxx::__normal_iterator<char*, std::string>, boost::spirit::qi::tst_pass_through> (
    start=<optimized out>, first=59 ';', last=0 '\000', filter=...)
    at /usr/local/include/boost/spirit/home/qi/string/detail/tst.hpp:82
82                      if (c == p->id)
```

oxidase:
Hi, the segfault is due to local scope of `unlimited` in constructor, and the rule stores just a reference to `unlimited`, so out scope of the constructor it will be a stale reference.

Here is patch [sf2231.txt](https://github.com/Project-OSRM/osrm-backend/files/209322/sf2231.txt)
but I would not commit as is because server response is
```
[info] 07-04-2016 23:57:05 127.0.0.1 - HTTPie/0.9.2 /nearest/v1/driving/13.388860,52.517037?number=1&radiuses=unlimited
[warn] [server error] code: /usr/include/boost/optional/optional.hpp:1004
in: boost::optional<T>::reference_const_type boost::optional<T>::get() const [with T = double; boost::optional<T>::reference_const_type = const double&]: this->is_initialized()
, uri: /nearest/v1/driving/13.388860,52.517037?number=1&radiuses=unlimited
[info] 08-04-2016 00:16:34 127.0.0.1 - HTTPie/0.9.2 /nearest/v1/driving/13.388860,52.517037?number=1&radiuses=42
 42
```
somewhere server should also check for optional is_initialized.

Other possibility would be to drop optional and make
```
using RadiusesT = std::vector<double>;
unlimited.add("unlimited", std::numeric_limits<double>::infinity());
qi::symbols<char, double> unlimited;
```

daniel-j-h:
Thank you, we want to remove the `optional<double>` wrapping anyway, so this is good for now.

Closing, as this landed in: https://github.com/Project-OSRM/osrm-backend/commit/f2ac4ff1c3edd159cbe0e4ca010124f265a2f010

PR #2233, opened 2016-04-07
Add cmake option to disable -DBOOST_TEST_DYN_LINK

springmeyer:
This is needed when building the unit tests against a static `libboost_unit_test_framework` otherwise you will hit a linking error like:

```
undef: __ZN5boost9unit_test14unit_test_mainEPFbvEiPPc
Undefined symbols for architecture x86_64:
  "boost::unit_test::unit_test_main(bool (*)(), int, char**)", referenced from:
      _main in engine_tests.cpp.o
ld: symbol(s) not found for architecture x86_64
```

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -225,7 +225,7 @@ include_directories(SYSTEM ${OSMIUM_INCLUDE_DIRS})
 
 
 find_package(Boost 1.49.0 REQUIRED COMPONENTS ${BOOST_COMPONENTS})
-if(NOT WIN32)
+if(NOT WIN32 AND NOT Boost_USE_STATIC_LIBS)
   add_definitions(-DBOOST_TEST_DYN_LINK)
 endif()
 add_definitions(-DBOOST_SPIRIT_USE_PHOENIX_V3)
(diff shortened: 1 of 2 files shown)


TheMarex:
Instead of adding a custom option coud you simply check for this option for [FindBoost.cmake](https://github.com/Kitware/CMake/blob/master/Modules/FindBoost.cmake#L109)

That way if you pass `cmake .. -DBoost_USE_STATIC_LIBS=ON` you will always observe the correct behaviour.

springmeyer:
@TheMarex - confirming what I'm hearing: you are recommending passing `-DBoost_USE_STATIC_LIBS=ON` and then checking for that in the main CMakeLists.txt to disable `-DBOOST_TEST_DYN_LINK`? (I tried setting `-DBoost_USE_STATIC_LIBS=ON` and that alone is not enough to disable it).

TheMarex:
@springmeyer yes I just pushed that change. Can you give it a spin?

springmeyer:
@TheMarex - Thanks! Tested and that works great.

springmeyer:
Rebased and ready to merge /cc @TheMarex

TheMarex:

Issue #1673, opened 2015-09-10
Integrate coverage report with coveralls.io


daniel-j-h:
Try out integration with https://coveralls.io.

Here's the user-contributed C++ support repository:
https://github.com/eddyxu/cpp-coveralls

Steps to take:

- [x] investigate coveralls.io setup
- [x] build system option to pass down cxxflags `-fprofile-arcs -ftest-coverage`
- [x] travis integration

lbud:
@springmeyer was working on getting this working with the `masonize` branch a few months ago. We left off here: https://coveralls.io/github/Project-OSRM/osrm-backend and can probably convince him to come back and help out :)

daniel-j-h:
This landed in https://github.com/Project-OSRM/osrm-backend/pull/2235.

PR #2235, opened 2016-04-07
[NOT READY] Integrate converage reporting

TheMarex:


CMakeLists.txt
```diff
@@ -133,6 +134,10 @@ if(CMAKE_BUILD_TYPE MATCHES Release)
   endif()
 endif()
 
+if (COVERAGE)
+  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -ftest-coverage")
```
daniel-j-h: Do we need `-g -O0` or at least `-fno-inline` in order for this to work properly?

daniel-j-h: Ah, and `-fprofile-arcs` for storing arcs between basic blocks in the data flow graph.

daniel-j-h:
See https://github.com/Project-OSRM/osrm-backend/issues/1673, and close ticket if you merge this in.

CMakeLists.txt
TheMarex: I'm going to add a check for `CMAKE_BUILD_TYPE == Debug`. Because that is the only sensible combination.

TheMarex:
:tada: :

Issue #2145, opened 2016-03-25
Low driving speed guessed for maxspeed=none


utack:
Your debug map confirmed the suspicion I had about [this route](https://www.openstreetmap.org/directions?engine=osrm_car&route=49.50363%2C11.29023%3B49.47557%2C11.16616#map=13/49.4897/11.2282)
The highway is avoided because the speeds on it are estimated far too low
[Here](http://map.project-osrm.org/debug/#13.6/49.4841/11.2519) you can see that [maxspeed=none on this part](https://www.openstreetmap.org/way/248868540#map=14/49.4816/11.2636) results in a far underestimed real speed of 83km/h and [maxspeed=120 on this part](https://www.openstreetmap.org/way/308241043) results in a bit more realistic speed of 107km/h (still a bit low, but the difference between the two is what really stands out)
Seems like [this rule](https://github.com/Project-OSRM/osrm-backend/blob/develop/profiles/car.lua#L121) is not really working

TheMarex:
Yes you are right, this should not fall back to motorway defaults as it apparently does. Currently we don't seem to support `maxspeed=none` at all. The `source:maxspeed=signals` seems curious? Shouldn't that either be `de:motorway` or `sign`?

utack:
Could you consider this as 5.0 milestone?
You probably have all sorts of internal restructuring going on, but this is a user facing bug that can reduce the quality of the calculated routes significantly.

PR #2239, opened 2016-04-08
Add maxspeed=none tag to car profile.

daniel-j-h:
Pull request to keep discussion from https://github.com/Project-OSRM/osrm-backend/issues/2145 moving. /cc @utack @TheMarex @danpat 

maxspeed=none means there is no explicit maxspeed limit. Set to
guestimate for driving on the Autobahn.

References:

 - https://github.com/Project-OSRM/osrm-backend/issues/2145
 - http://wiki.openstreetmap.org/wiki/Key:maxspeed
 - http://taginfo.openstreetmap.org/tags/maxspeed=none
 - http://wiki.openstreetmap.org/wiki/OSM_tags_for_routing/Maxspeed

### DIFF
--- a/profiles/car.lua
+++ b/profiles/car.lua
@@ -124,7 +124,8 @@ maxspeed_table = {
   ["gb:motorway"] = (70*1609)/1000,
   ["uk:nsl_single"] = (60*1609)/1000,
   ["uk:nsl_dual"] = (70*1609)/1000,
-  ["uk:motorway"] = (70*1609)/1000
+  ["uk:motorway"] = (70*1609)/1000,
+  ["none"] = 140
 }
 
 -- set profile properties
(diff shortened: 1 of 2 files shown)


TheMarex:
@daniel-j-h there is a typo in your profile. Missing a `,` in the previous line. If this passes and works, we can include it in the next RC.

daniel-j-h:
> @daniel-j-h there is a typo in your profile. Missing a , in the previous line. If this passes and works, we can include it in the next RC.

I can't see what you mean here, it looks fine to me and runs both locally as well as on Travis. Maybe you already ammended my commit?

Travis is green now. We had some cucumber timeouts on OSX Clang builds --- Travis' fault.

TheMarex:
I fixed your commit. :-) You can merge this. :+1:

daniel-j-h:
This landed. @utack we pinned the demo server to a osrm-backend snapshot pre-v5, so you will see this change on the demo server as soon as we spin up a dedicated v5 demo server. This will probably take a couple of days.

utack:
Sounds good, thanks for fixing that so quickly. Looking forward to testing some routes then.

PR #2240, opened 2016-04-08
On successful travis build, trigger node-osrm build

springmeyer:
This triggers a travis build of `node-osrm` when `osrm-backend` builds and runs successfully.

 - I've chosen one `osrm-backend` job to be the triggering job (the g++-5 release build seemed like a good one but it could be anything)
 - The triggered builds overrides the `.travis.yml` matrix for `node-osrm` to only run one job: a `clang release` job. The idea here is to be as fast as possible and not to overload travis.

TODO before merging:

  - [x] Need to add a secure variable to encode `$TRAVIS_TOKEN`

Note: anyone with `$TRAVIS_TOKEN` defined can also run this locally to trigger `node-osrm` builds easily.:

```bash
./scripts/trigger-downstream.sh
```

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -24,6 +24,17 @@ matrix:
 
   include:
 
+    # Release / Triggered build
+    # This is listed first so it runs first to enable
+    # quickly knowing if downstream apps are broken
+    - os: linux
+      compiler: gcc
+      addons: &gcc5
+        apt:
+          sources: ['ubuntu-toolchain-r-test']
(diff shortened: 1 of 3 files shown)


springmeyer:
This is green on travis now. Ready to merge after last :eyes: from @TheMarex

TheMarex:

springmeyer:
Note: Now that we have cron job builds for node-osrm I'm not planning on restoring this branch (https://github.com/Project-OSRM/node-osrm/issues/203).

PR #2245, opened 2016-04-11
One-Definition-Rule violation in viewport

daniel-j-h:


TheMarex:
:+1: good catch, merging.

PR #2254, opened 2016-04-12
Unit tests -- table service

karenzshea:
Tests for symmetric and asymmetric matrices and source/destination options

cc @TheMarex @daniel-j-h

### DIFF
--- a/unit_tests/library/match.cpp
+++ b/unit_tests/library/match.cpp
@@ -4,6 +4,7 @@
 #include "args.hpp"
 #include "fixture.hpp"
 #include "coordinates.hpp"
+#include "waypoint_check.hpp"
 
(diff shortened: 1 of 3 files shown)


karenzshea:
@TheMarex @daniel-j-h I've added all the unit tests I see fit for table but I'm getting this failure on a `tile` test that I'm not sure I understand

> unknown location(0): fatal error in "test_tile": memory access violation at address: 0x7f130c1cd810: no mapping at fault address

ideas?

daniel-j-h:
> I'm getting this failure on a tile test that I'm not sure I understand

Tile test? I can't see you touching tile tests here, is this unrelated?

daniel-j-h:
I checked out, compiled and ran your branch. Looks fine to me. Please ping me with more information.

PR #2261, opened 2016-04-13
fix r-tree abort location

MoKob:
Fixes the R-Tree abort location to correctly handle cases without return.

This made it into the code with my unit-test fixes :(

### DIFF
--- a/include/util/static_rtree.hpp
+++ b/include/util/static_rtree.hpp
@@ -441,6 +441,14 @@ class StaticRTree
                 // inspecting an actual road segment
                 auto &current_candidate = current_query_node.node.template get<CandidateSegment>();
 
+                // to allow returns of no-results if too restrictive filtering, this needs to be done here
+                // even though performance would indicate that we want to stop after adding the first candidate
+                if (terminate(results.size(), current_candidate))
+                {
+                    traversal_queue = std::priority_queue<QueryCandidate>{};
+                    break;
+                }
+
                 auto use_segment = filter(current_candidate);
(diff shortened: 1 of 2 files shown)


unit_tests/util/static_rtree.cpp
```diff
@@ -335,6 +335,34 @@ BOOST_AUTO_TEST_CASE(regression_test)
     BOOST_CHECK_EQUAL(result_ls.front().v, result_rtree.front().v);
 }
 
+// Bug: If you querry a point with a narrow radius, no result should be returned
```
freenerd: Why does this say `bug`? It looks like the code does quite that, not returning results?

MoKob: It checks this explicit bug that vertices are returned even though none exist

include/util/static_rtree.hpp
```diff
@@ -441,6 +441,14 @@ class StaticRTree
                 // inspecting an actual road segment
                 auto &current_candidate = current_query_node.node.template get<CandidateSegment>();
 
+                // to allow returns of no-results if too restrictive filtering, this needs to be done here
+                // even though performance would indicate that we want to stop after adding the first candidate
+                if (terminate(results.size(), current_candidate))
```
TheMarex: Technically we should probably use two separate termination criterion instead of cramming it in one function call.

MoKob: We could technically filter the candidate based on its distance, not only abort. We already have the two parts. The abort would have to be a bit more sophisticated and we would duplicate the distance computation, however.

PR #2248, opened 2016-04-11
[merge-ready] Refactor/guidance

MoKob:
Further Refactoring of Guidance Work

This PR further refactors/simplifies the `turn_handler`.

 - [x] refactor three way turns
 - [x] refactor larger turns
       - [x] refactor assign left turns -> now works as a hand-off to right turns
       - [x] refactor assign right turns

PR #2268, opened 2016-04-15
[ready] Fix/roundabout segregated

MoKob:
Technically more an data-issue than anything else, but currently we could issue instructions against the turning direction of a roundabout for segregated roads merging into a roundabout at a single point.

This PR addresses the issue.

 - [ ] requires https://github.com/Project-OSRM/osrm-backend/pull/2248
 - [x] review done


src/engine/guidance/post_processing.cpp
```diff
@@ -25,6 +25,28 @@ namespace engine
...
+        const auto type = static_cast<int>(step.maneuver.instruction.type);
+        const auto modifier = static_cast<int>(step.maneuver.instruction.direction_modifier);
+
+        std::cout << "\t[" << ++segment << "]: " << type << " " << modifier
+                  << " Duration: " << step.duration << " Distance: " << step.distance
+                  << " Geometry: " << step.geometry_begin << " " << step.geometry_end
+                  << " exit: " << step.maneuver.exit
+                  << " Intersections: " << step.maneuver.intersections.size() << " [";
+
+        for (auto intersection : step.maneuver.intersections)
```
daniel-j-h: `const auto intersection&`

src/extractor/guidance/intersection_generator.cpp
```diff
@@ -207,6 +208,16 @@ Intersection IntersectionGenerator::mergeSegregatedRoads(Intersection intersecti
     if (intersection.size() == 1)
         return intersection;
 
+    bool is_connected_to_roundabout = false;
```
daniel-j-h: can be const and initialized via a lambda?

src/extractor/guidance/roundabout_handler.cpp
```diff
@@ -79,6 +80,43 @@ detail::RoundaboutFlags RoundaboutHandler::getRoundaboutFlags(
...
+        // only check actual outgoing edges
+        if (edge_data.reversed)
+        {
+
+            // remember whether we have seen the roundabout in-part
+            if (edge_data.roundabout)
+                past_roundabout_angle = true;
+
+            continue;
+        }
```
daniel-j-h: Maybe flatten to

    if reversed && roundabout

MoKob: Wouldn't be the same, would  it? I continue on all reverse edges, but only set the flag if its also a roundabout.

TheMarex:

Issue #2278, opened 2016-04-18
Suppress `mode: movable bridge`


bsudekum:
> "Continue straight on I-95 over the bridge"

Telling the user they are going to go over a bridge seems redundant. Ideally, any event `mode: 'movable bridge'` would be removed from the response. 

/cc @MoKob @karenzshea @TheMarex

MoKob:
These mode changes are a profile decision. Its less a decision to suppress it but to configure the profile to supply reasonable modes for the driving part. I guess bridges are among the first to consider here as a mode not to emit.

TheMarex:
I think we should just remove it as a mode of transportation. IMHO bridges don't really qualify as such.

PR #2281, opened 2016-04-19
Remove movable bridge travel mode

karenzshea:
https://github.com/Project-OSRM/osrm-backend/issues/2278

Movable bridge is no longer returned as a unique mode type. The mode change is still factored into duration calculations.

cc @bsudekum @TheMarex


TheMarex:
Looks good to me. Merging.

PR #2291, opened 2016-04-21
Add functionality to write turn lookups and read turn penalty files

### LINKED ISSUE #1830: Load arbitrary turn costs from data files
The edge-based graph edge weights are made up of two parts: cost values from traversing node-based-edges (segments), and costs from turning movements.

OSRM already has the ability to load arbitrary segment speed values from files prior to graph contraction [see here](https://github.com/Project-OSRM/osrm-backend/blob/da11113bf1becaf0de2c59ad9d157ec5edb19aa3/contractor/processing_chain.cpp#L177-L190).  However, the turning movement costs are currently only set by the Lua profile, and are not adjustable on a per-intersection basis.

If real measurements are available, we should be able to load them in at the same time.

OSRM identifies a *turning movement* using 3 node-based-nodes: u, v, w

```
    u
    |
    v---w
    |
   ...
```

The angle between uv and uw is what is currently passed to the `turn_function` in the Lua profile.

I propose we take a similar approach here to what was done for the segment speed lookups:

  1.   During generation of the edge-based-graph, write out a lookup file between the three nodes involved in the uvw turn, and the new edge-based-edge ID.  Node order is important, as it specifies which turn through an intersection has been taken.
  2.  An offline process can use the original OSM geometry and output a file that contains:
    `u,v,w,duration` (perhaps CSV to keep things simple?)
  3.  During `osrm-prepare`, as the edge-based graph is loaded, we can use the lookup file from (1) in combination with data generated offline in (2) to set a "real" value for the turn penalty, rather than simply basing it on the results of the `turn_function`.  This would slot right in [here](https://github.com/Project-OSRM/osrm-backend/blob/da11113bf1becaf0de2c59ad9d157ec5edb19aa3/contractor/processing_chain.cpp#L262).

Out of scope for this feature is discussing exactly what the weight value represents or how these turn costs are actually measured.  It would be great to see some exploration in this space

lbud:
Fixes #1830.
cc @danpat @TheMarex — this feels ready, though a lot of the contractor code I added was copied from the speed segment functionality and I would love if someone could take a second look 👀

### DIFF
--- a/include/engine/routing_algorithms/routing_base.hpp
+++ b/include/engine/routing_algorithms/routing_base.hpp
@@ -302,7 +302,6 @@ template <class DataFacadeT, class Derived> class BasicRoutingInterface
 
                 BOOST_ASSERT(weight_vector.size() == id_vector.size());
                 // ed.distance should be total_weight + penalties (turn, stop, etc)
-                BOOST_ASSERT(ed.distance >= total_weight);
                 const bool is_first_segment = unpacked_path.empty();
 
                 const std::size_t start_index =
--- a/src/contractor/contractor.cpp
+++ b/src/contractor/contractor.cpp
@@ -554,11 +554,13 @@ std::size_t Contractor::LoadEdgeExpandedGraph(
             if (turn_iter != turn_penalty_lookup.end())
             {
                 int new_turn_weight = static_cast<int>(turn_iter->second.first * 10);
-                new_weight += new_turn_weight;
+                BOOST_ASSERT(new_turn_weight + new_weight > 0);
+                inbuffer.weight = new_turn_weight + new_weight;
             }
-
-            BOOST_ASSERT(static_cast<int>(fixed_penalty) + new_weight > 0);
-            inbuffer.weight = fixed_penalty + new_weight;
+            else
+            {
+                inbuffer.weight = fixed_penalty + new_weight;
+            }
         }
 
         edge_based_edge_list.emplace_back(std::move(inbuffer));
--- a/src/contractor/contractor.cpp
+++ b/src/contractor/contractor.cpp
@@ -29,6 +29,7 @@
 #include <memory>
 #include <thread>
 #include <iterator>
+#include <tuple>
 
 namespace std
 {
@@ -38,6 +39,16 @@ template <> struct hash<std::pair<OSMNodeID, OSMNodeID>>
     std::size_t operator()(const std::pair<OSMNodeID, OSMNodeID> &k) const
     {
         return static_cast<uint64_t>(k.first) ^ (static_cast<uint64_t>(k.second) << 12);
+    }
+};
+
+template <> struct hash<std::tuple<OSMNodeID, OSMNodeID, OSMNodeID>>
+{
+    std::size_t operator()(const std::tuple<OSMNodeID, OSMNodeID, OSMNodeID> &k) const
+    {
+        return (static_cast<uint64_t>(std::get<0>(k)) << 24) ^
+               (static_cast<uint64_t>(std::get<1>(k)) << 12) ^
+               (static_cast<uint64_t>(std::get<2>(k)));
     }
 };
 }
@@ -71,9 +82,9 @@ int Contractor::Run()
 
     std::size_t max_edge_id = LoadEdgeExpandedGraph(
         config.edge_based_graph_path, edge_based_edge_list, config.edge_segment_lookup_path,
-        config.edge_penalty_path, config.segment_speed_lookup_paths, config.node_based_graph_path,
-        config.geometry_path, config.datasource_names_path, config.datasource_indexes_path,
-        config.rtree_leaf_path);
+        config.edge_penalty_path, config.segment_speed_lookup_paths,
+        config.turn_penalty_lookup_paths, config.node_based_graph_path, config.geometry_path,
+        config.datasource_names_path, config.datasource_indexes_path, config.rtree_leaf_path);
 
     // Contracting the edge-expanded graph
 
@@ -129,6 +140,7 @@ std::size_t Contractor::LoadEdgeExpandedGraph(
     const std::string &edge_segment_lookup_filename,
     const std::string &edge_penalty_filename,
     const std::vector<std::string> &segment_speed_filenames,
+    const std::vector<std::string> &turn_penalty_filenames,
     const std::string &nodes_filename,
     const std::string &geometry_filename,
     const std::string &datasource_names_filename,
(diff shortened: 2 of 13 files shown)


src/contractor/contractor.cpp
```diff
@@ -494,6 +541,23 @@ std::size_t Contractor::LoadEdgeExpandedGraph(
...
+
+            auto turn_iter = turn_penalty_lookup.find(std::make_tuple(from_id, via_id, to_id));
+            if (turn_iter != turn_penalty_lookup.end())
+            {
+                int new_turn_weight = static_cast<int>(turn_iter->second.first * 10);
+                new_weight += new_turn_weight;
+            }
+
+            BOOST_ASSERT(static_cast<int>(fixed_penalty) + new_weight > 0);
             inbuffer.weight = fixed_penalty + new_weight;
```
danpat: We should use *either* `new_turn_weight`, or `fixed_penalty` here.

The `fixed_penalty` value is the original turn_penalty + stop_sign penalty + u_turn_penalty.  To be absolutely correct, we should split `fixed_penalty` into those three parts, save them, then add the proper ones back here.

However, I suspect very few intersections have all three values applied at the same time, so we're probably save to just use `new_turn_weight` instead of `fixed_penalty` when we find an intersection match.

src/extractor/edge_based_graph_factory.cpp
```diff
@@ -431,6 +431,23 @@ void EdgeBasedGraphFactory::GenerateEdgeExpandedEdges(
...
+
+                    // We also now write out the mapping between the edge-expanded edges and the
+                    // original nodes. Since each edge represents a possible maneuver, external
+                    // programs can use this to quickly perform updates to edge weights in order
+                    // to penalize certain turns.
+
+                    const auto &from_node = m_node_info_list[node_u];
+                    const auto &via_node =
+                        m_node_info_list[m_node_based_graph->GetTarget(edge_from_u)];
+                    const auto &to_node = m_node_info_list[m_node_based_graph->GetTarget(turn.eid)];
```
danpat: I don't think `to_node` is correct in all cases here.

The `node_based_graph->GetTarget` gets the *last* node in a node-based-edge.  In the case where this is a simple edge, with just one segment, we'll get the correct value here.

However, in the case where the node-based-edge is compressed, what we want is the *second* node.

To get `to_node`, we'll need to do a lookup in the compressed geometry container, get the compressed geometry bucket, then get the first `node_id` in that list.  This will work for all types of edges.

You can write a test case for this by creating ways that have more than two segments, like:

```
| a | b | c | d |
| e |   |   |   |
| f |   |   |   |

Ways:
abcd
aef
```

include/engine/routing_algorithms/routing_base.hpp
```diff
@@ -302,7 +302,6 @@ template <class DataFacadeT, class Derived> class BasicRoutingInterface
 
                 BOOST_ASSERT(weight_vector.size() == id_vector.size());
                 // ed.distance should be total_weight + penalties (turn, stop, etc)
-                BOOST_ASSERT(ed.distance >= total_weight);
```
MoKob: If this is not true anymore, either the comment has to go as well or we need an adjustment of both the comment and the assertion

src/contractor/contractor.cpp
```diff
@@ -172,37 +185,74 @@ std::size_t Contractor::LoadEdgeExpandedGraph(
...
-                segment_speed_lookup[std::make_pair(OSMNodeID(from_node_id),
-                                                    OSMNodeID(to_node_id))] =
-                    std::make_pair(speed, file_id);
+                util::SimpleLogger().Write()
+                    << "Segment speed data supplied, will update edge weights from "
+                    << segment_speed_filename;
+                io::CSVReader<3> csv_in(segment_speed_filename);
+                csv_in.set_header("from_node", "to_node", "speed");
+                uint64_t from_node_id{};
+                uint64_t to_node_id{};
```
MoKob: `std::uint64_t`

src/contractor/contractor.cpp
```diff
@@ -172,37 +185,74 @@ std::size_t Contractor::LoadEdgeExpandedGraph(
...
     // for each segment
     std::vector<uint8_t> m_geometry_datasource;
 
-    if (update_edge_weights)
+    if (update_edge_weights || update_turn_penalties)
     {
-        uint8_t file_id = 1;
-        for (auto segment_speed_filename : segment_speed_filenames)
+        uint8_t segment_file_id = 1;
+        uint8_t turn_file_id = 1;
```
MoKob: std::uint8_t

src/contractor/contractor.cpp
```diff
@@ -172,37 +185,74 @@ std::size_t Contractor::LoadEdgeExpandedGraph(
...
+                                                        OSMNodeID(to_node_id))] =
+                        std::make_pair(speed, segment_file_id);
+                }
+                ++segment_file_id;
+
+                // Check for overflow
+                if (segment_file_id == 0)
+                {
+                    throw util::exception(
+                        "Sorry, there's a limit of 254 segment speed files; you supplied too many");
```
MoKob: any reason to start at `1`, count until `0` and allow only `254`? Currently this should at least say 255?  (1...255 are valid values)

The same could be done with a proper `256` count, starting at 0.

src/contractor/contractor.cpp
```diff
@@ -172,37 +185,74 @@ std::size_t Contractor::LoadEdgeExpandedGraph(
...
+                ++turn_file_id;
+
+                // Check for overflow
+                if (turn_file_id == 0)
+                {
+                    throw util::exception(
+                        "Sorry, there's a limit of 254 turn penalty files; you supplied too many");
+                }
             }
         }
```
MoKob: looks like the `if( update...)` loop should end here? it includes the full a lot of stuff that shouldn't be in here, as well?

src/contractor/contractor.cpp
```diff
@@ -412,8 +461,7 @@ std::size_t Contractor::LoadEdgeExpandedGraph(
         std::ofstream datasource_stream(datasource_indexes_filename, std::ios::binary);
         if (!datasource_stream)
         {
-            throw util::exception("Failed to open " + datasource_indexes_filename +
-                                  " for writing");
+            throw util::exception("Failed to open " + datasource_indexes_filename + " for writing");
         }
```
MoKob: just out of interest. Why are we writing in a function called `loadEdgeExpandedGraph`?

src/contractor/contractor.cpp
MoKob: After further looking at the context, it seems we are fine for the update. I am still a bit confused on how its working, though. Especially the part of writing out data in a read raises alarm bells in my head.

TheMarex:

src/contractor/contractor.cpp
danpat: @MoKob 0 is reserved for the "lua profile" datasource name.

src/contractor/contractor.cpp
danpat: @MoKob performance - it's done here because (a) all the we need is available, and (b) to avoid looping over the same data again.  I did it this way to minimize the additional time added to `osrm-contract`

src/contractor/contractor.cpp
MoKob: @danpat in that case its still 255 files, isn't it?

danpat: Yes, you're right, the error should be updated.

PR #2293, opened 2016-04-22
[review] suppress suffix changes

MoKob:
Adds the possibility to define a list of suffixes in the profile.

On obvious maneuvers (e.g. going straight onto the next road), the names are not only compared for equality but also checked for their possible suffixes. If such a change is detected, which a human would probably not hold important to know about, the suffix change will not be announced.


TheMarex:

PR #2298, opened 2016-04-22
enable reverse name IDs

MoKob:
It can occur that a single road can be assigned different names, based on the direction of travel.

This experimental PR is used to test the implications of enabling these reverse name ids.


MoKob:
Not actionable right now. We can keep the branch alive, but the engine has changed quite a bit. The basic principles might still work, though.

PR #2299, opened 2016-04-22
[ready] Improve Tests for Guidance

MoKob:
This PR adds a few tests to the guidance framework.

Furthermore, it fixes a few issues in guidance and re-enables the respective tests:

It enables the tests currently disabled in `features/guidance/turn.feature` as well as `features/guidance/fork.feature`.

- [x] merge https://github.com/Project-OSRM/osrm-backend/pull/2297 first

PR #2300, opened 2016-04-22
[review] fix discovery of obvious turns, triple forks

MoKob:
- [ ] merge https://github.com/Project-OSRM/osrm-backend/pull/2299 first

Fixes a few small bugs in turn discovery, handling of tripple forks


MoKob:
seems to be in with another PR already

PR #2304, opened 2016-04-22
[review] suppress trivial name changes

MoKob:
Continuing the work on https://github.com/Project-OSRM/osrm-backend/issues/2263, this PR tackles bridges and tunnels.
Both often are listed as `...bridge` or `...tunnel` within OSM data. Suppressing `new name` instructions for these types of name changes, we can reduce the set of unnecessary instructions.

 - [ ] requires merge of https://github.com/Project-OSRM/osrm-backend/pull/2302
 - [x] review

PR #2305, opened 2016-04-22
[review] suppress suffix changes

MoKob:
Adds the possibility to define a list of suffixes in the profile.

On obvious maneuvers (e.g. going straight onto the next road), the names are not only compared for equality but also checked for their possible suffixes. If such a change is detected, which a human would probably not hold important to know about, the suffix change will not be announced.

- [X] @daniel-j-h reviewed
- [ ] requires merge of https://github.com/Project-OSRM/osrm-backend/pull/2304

### DIFF
--- a/include/engine/guidance/assemble_steps.hpp
+++ b/include/engine/guidance/assemble_steps.hpp
@@ -79,7 +79,7 @@ std::vector<RouteStep> assembleSteps(const DataFacadeT &facade,
 
         // some name changes are not announced in our processing. For these, we have to keep the
         // first name on the segment
-        unsigned step_name_id = source_node.name_id;
+        auto step_name_id = source_node.name_id;
         for (std::size_t leg_data_index = 0; leg_data_index < leg_data.size(); ++leg_data_index)
         {
             const auto &path_point = leg_data[leg_data_index];
--- a/include/engine/guidance/assemble_steps.hpp
+++ b/include/engine/guidance/assemble_steps.hpp
@@ -14,6 +14,7 @@
 #include "util/coordinate_calculation.hpp"
 
 #include <boost/optional.hpp>
+#include <cstddef>
 #include <vector>
 
 namespace osrm
@@ -75,20 +76,30 @@ std::vector<RouteStep> assembleSteps(const DataFacadeT &facade,
         // We need to skip the first segment because it is already covered by the
         // initial start of a route
         int segment_duration = 0;
-        for (const auto &path_point : leg_data)
+
+        // some name changes are not announced in our processing. For these, we have to keep the
+        // first name on the segment
+        unsigned step_name_id = source_node.name_id;
+        for (std::size_t leg_data_index = 0; leg_data_index < leg_data.size(); ++leg_data_index)
         {
+            const auto &path_point = leg_data[leg_data_index];
             segment_duration += path_point.duration_until_turn;
 
             // all changes to this check have to be matched with assemble_geometry
             if (path_point.turn_instruction.type != extractor::guidance::TurnType::NoTurn)
             {
                 BOOST_ASSERT(segment_duration >= 0);
-                const auto name = facade.GetNameForID(path_point.name_id);
+                const auto name = facade.GetNameForID(step_name_id);
                 const auto distance = leg_geometry.segment_distances[segment_index];
-                steps.push_back(RouteStep{path_point.name_id, name, NO_ROTARY_NAME,
+                steps.push_back(RouteStep{step_name_id, name, NO_ROTARY_NAME,
                                           segment_duration / 10.0, distance, path_point.travel_mode,
                                           maneuver, leg_geometry.FrontIndex(segment_index),
                                           leg_geometry.BackIndex(segment_index) + 1});
+                if (leg_data_index + 1 < leg_data.size()){
+                    step_name_id = leg_data[leg_data_index + 1].name_id;
+                } else {
+                    step_name_id = target_node.name_id;
+                }
                 maneuver = detail::stepManeuverFromGeometry(path_point.turn_instruction,
                                                             leg_geometry, segment_index);
                 segment_index++;
@@ -98,7 +109,7 @@ std::vector<RouteStep> assembleSteps(const DataFacadeT &facade,
         const auto distance = leg_geometry.segment_distances[segment_index];
         const int duration = segment_duration + target_duration;
         BOOST_ASSERT(duration >= 0);
-        steps.push_back(RouteStep{target_node.name_id, facade.GetNameForID(target_node.name_id),
+        steps.push_back(RouteStep{step_name_id, facade.GetNameForID(step_name_id),
                                   NO_ROTARY_NAME, duration / 10., distance, target_mode, maneuver,
                                   leg_geometry.FrontIndex(segment_index),
                                   leg_geometry.BackIndex(segment_index) + 1});
@@ -132,7 +143,7 @@ std::vector<RouteStep> assembleSteps(const DataFacadeT &facade,
     BOOST_ASSERT(!leg_geometry.locations.empty());
     steps.push_back(RouteStep{target_node.name_id, facade.GetNameForID(target_node.name_id),
                               NO_ROTARY_NAME, ZERO_DURATION, ZERO_DISTANCE, target_mode,
-                              final_maneuver, leg_geometry.locations.size()-1,
+                              final_maneuver, leg_geometry.locations.size() - 1,
                               leg_geometry.locations.size()});
 
     return steps;
--- a/include/engine/guidance/toolkit.hpp
+++ b/include/engine/guidance/toolkit.hpp
@@ -1,5 +1,5 @@
-#ifndef OSRM_UTIL_GUIDANCE_TOOLKIT_HPP_
-#define OSRM_UTIL_GUIDANCE_TOOLKIT_HPP_
+#ifndef OSRM_ENGINE_GUIDANCE_TOOLKIT_HPP_
+#define OSRM_ENGINE_GUIDANCE_TOOLKIT_HPP_
 
 #include "extractor/guidance/turn_instruction.hpp"
 #include "util/bearing.hpp"
@@ -25,9 +25,12 @@ inline bool entersRoundabout(const extractor::guidance::TurnInstruction instruct
 {
     return (instruction.type == extractor::guidance::TurnType::EnterRoundabout ||
             instruction.type == extractor::guidance::TurnType::EnterRotary ||
+            instruction.type == extractor::guidance::TurnType::EnterRoundaboutIntersection ||
             instruction.type == extractor::guidance::TurnType::EnterRoundaboutAtExit ||
             instruction.type == extractor::guidance::TurnType::EnterRotaryAtExit ||
+            instruction.type == extractor::guidance::TurnType::EnterRoundaboutIntersectionAtExit ||
             instruction.type == extractor::guidance::TurnType::EnterAndExitRoundabout ||
+            instruction.type == extractor::guidance::TurnType::EnterAndExitRotary ||
             instruction.type == extractor::guidance::TurnType::EnterAndExitRotary);
 }
 
@@ -35,8 +38,10 @@ inline bool leavesRoundabout(const extractor::guidance::TurnInstruction instruct
 {
     return (instruction.type == extractor::guidance::TurnType::ExitRoundabout ||
             instruction.type == extractor::guidance::TurnType::ExitRotary ||
+            instruction.type == extractor::guidance::TurnType::ExitRoundaboutIntersection ||
             instruction.type == extractor::guidance::TurnType::EnterAndExitRoundabout ||
-            instruction.type == extractor::guidance::TurnType::EnterAndExitRotary);
+            instruction.type == extractor::guidance::TurnType::EnterAndExitRotary ||
+            instruction.type == extractor::guidance::TurnType::EnterAndExitRoundaboutIntersection);
 }
 
 inline bool staysOnRoundabout(const extractor::guidance::TurnInstruction instruction)
@@ -68,4 +73,4 @@ inline double angularDeviation(const double angle, const double from)
 } // namespace engine
 } // namespace osrm
 
-#endif /* OSRM_UTIL_GUIDANCE_TOOLKIT_HPP_ */
+#endif /* OSRM_ENGINE_GUIDANCE_TOOLKIT_HPP_ */
--- a/include/engine/guidance/toolkit.hpp
+++ b/include/engine/guidance/toolkit.hpp
@@ -65,8 +65,8 @@ inline extractor::guidance::DirectionModifier angleToDirectionModifier(const dou
 
 inline double angularDeviation(const double angle, const double from)
 {
-        const double deviation = std::abs(angle - from);
-            return std::min(360 - deviation, deviation);
+    const double deviation = std::abs(angle - from);
+    return std::min(360 - deviation, deviation);
 }
 
 } // namespace guidance
--- a/include/extractor/guidance/roundabout_handler.hpp
+++ b/include/extractor/guidance/roundabout_handler.hpp
@@ -1,14 +1,17 @@
 #ifndef OSRM_EXTRACTOR_GUIDANCE_ROUNDABOUT_HANDLER_HPP_
 #define OSRM_EXTRACTOR_GUIDANCE_ROUNDABOUT_HANDLER_HPP_
 
+#include "extractor/compressed_edge_container.hpp"
 #include "extractor/guidance/intersection.hpp"
 #include "extractor/guidance/intersection_handler.hpp"
+#include "extractor/guidance/roundabout_type.hpp"
 #include "extractor/query_node.hpp"
 
 #include "util/name_table.hpp"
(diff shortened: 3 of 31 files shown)


include/engine/guidance/assemble_steps.hpp
```diff
@@ -75,20 +76,30 @@ std::vector<RouteStep> assembleSteps(const DataFacadeT &facade,
         // We need to skip the first segment because it is already covered by the
         // initial start of a route
         int segment_duration = 0;
-        for (const auto &path_point : leg_data)
+
+        // some name changes are not announced in our processing. For these, we have to keep the
+        // first name on the segment
+        unsigned step_name_id = source_node.name_id;
```
daniel-j-h: auto

include/engine/guidance/post_processing.hpp
```diff
@@ -17,6 +17,13 @@ namespace guidance
 // passed as none-reference to modify in-place and move out again
 std::vector<RouteStep> postProcess(std::vector<RouteStep> steps);
 
+// Multiple possible reasons can result in unnecessary/confusing instructions
+// A prime example would be a segregated intersection. Turning around at this
+// intersection would result in two instructions to turn left.
+// Collapsing such turns into a single turn instruction, we give a clearer
+// set of instructionst that is not cluttered by unnecessary turns/name changes.
+std::vector<RouteStep> collapseTurns(std::vector<RouteStep> steps);
```
daniel-j-h: :+1: on comments!

include/engine/guidance/toolkit.hpp
```diff
@@ -56,6 +56,12 @@ inline extractor::guidance::DirectionModifier angleToDirectionModifier(const dou
     return extractor::guidance::DirectionModifier::Left;
 }
 
+inline double angularDeviation(const double angle, const double from)
+{
+        const double deviation = std::abs(angle - from);
+            return std::min(360 - deviation, deviation);
```
daniel-j-h: format

include/extractor/guidance/roundabout_handler.hpp
```diff
@@ -37,18 +37,25 @@ class RoundaboutHandler : public IntersectionHandler
...
-    Intersection operator()(const NodeID from_nid, const EdgeID via_eid, Intersection intersection) const override final;
+    Intersection operator()(const NodeID from_nid,
+                            const EdgeID via_eid,
+                            Intersection intersection) const override final;
 
   private:
-    detail::RoundaboutFlags getRoundaboutFlags(const NodeID from_nid, const EdgeID via_eid, const Intersection &intersection) const;
+    detail::RoundaboutFlags getRoundaboutFlags(const NodeID from_nid,
+                                               const EdgeID via_eid,
+                                               const Intersection &intersection) const;
```
daniel-j-h: These are all formatting changes from what I can tell. Was the format off before? Just asking not that we're accidentally introducing formatting with a different clang-format version again.

include/extractor/guidance/toolkit.hpp
```diff
@@ -356,12 +363,24 @@ inline bool requiresNameAnnounced(const std::string &from, const std::string &to
...
 
+    const auto getCommonLength = [](const std::string &first, const std::string &second) {
+        for (std::size_t common_end = 0; common_end < std::min(first.size(), second.size());
+             ++common_end)
+        {
+            if (first[common_end] != second[common_end])
+                return common_end;
+        }
+        return std::min(first.size(), second.size());
+    };
```
daniel-j-h: std::mismatch (?)

daniel-j-h: (or boost::mismatch(fst, snd) of course)

include/extractor/guidance/toolkit.hpp
```diff
@@ -356,12 +363,24 @@ inline bool requiresNameAnnounced(const std::string &from, const std::string &to
...
+
     split(from, from_name, from_ref);
     split(to, to_name, to_ref);
 
     // check similarity of names
     const auto names_are_empty = from_name.empty() && to_name.empty();
-    const auto names_are_equal = from_name == to_name;
+    const auto name_is_contained =
+        boost::starts_with(from_name, to_name) || boost::starts_with(to_name, from_name);
```
daniel-j-h: :+1: :)

include/extractor/guidance/toolkit.hpp
```diff
@@ -370,9 +389,39 @@ inline bool requiresNameAnnounced(const std::string &from, const std::string &to
...
+                : "";
+        boost::algorithm::to_lower(second_suffix);
+
+        const auto first_delta_is_suffix =
+            first_suffix.empty() || suffix_table.isSuffix(first_suffix);
+        const auto second_delta_is_suffix =
+            second_suffix.empty() || suffix_table.isSuffix(second_suffix);
+
+        return first_delta_is_suffix && second_delta_is_suffix;
+    };
```
daniel-j-h: This lambda is a bit ugly, and has dup. code for fist / second suffix. Maybe refactor a bit.

include/extractor/guidance/toolkit.hpp
```diff
@@ -397,6 +446,25 @@ inline bool canBeSeenAsFork(const FunctionalRoadClass first, const FunctionalRoa
     return std::abs(getPriority(first) - getPriority(second)) <= 1;
 }
 
+// To simplify handling of Left/Right hand turns, we can mirror turns and write an intersection
+// handler only for one side. The mirror function turns a left-hand turn in a equivalent right-hand
+// turn and vice versa.
```
daniel-j-h: Great idea!

include/extractor/suffix_table.hpp
```diff
@@ -0,0 +1,31 @@
...
+namespace extractor
+{
+// A table containing suffixes.
+// At the moment, it is only a front for an unordered set. At some point we might want to make it
+// country dependent and have it behave accordingly
+class SuffixTable final
+{
+  public:
+    // has to be called prior to suffix detection
+    void fill(lua_State *lua_state);
```
daniel-j-h: Then why not passing the state to the ctor? => ctor establishes invariant, user is able to call member functions

profiles/car.lua
```diff
@@ -11,6 +11,9 @@ access_tags_hierarchy = { "motorcar", "motor_vehicle", "vehicle", "access" }
 service_tag_restricted = { ["parking_aisle"] = true }
 restriction_exception_tags = { "motorcar", "motor_vehicle", "vehicle" }
 
+-- A list of suffixes to suppress in name change instructions
+suffix_list = { "N", "NE", "E", "SE", "S", "SW", "W", "NW", "road", "street" }
```
daniel-j-h: What's your take on name canonicalization / normalization? Can this be more generic?

src/engine/guidance/post_processing.cpp
```diff
@@ -27,6 +27,29 @@ namespace guidance
...
 namespace detail
 {
+
+void print(const std::vector<RouteStep> &steps)
+{
+    std::cout << "Path\n";
+    int segment = 0;
+    for (const auto &step : steps)
+    {
+        const auto type = static_cast<int>(step.maneuver.instruction.type);
```
daniel-j-h: You can use std::underlying_type to make the code work independently of your enum's representation

src/engine/guidance/post_processing.cpp
```diff
@@ -44,6 +67,13 @@ RouteStep forwardInto(RouteStep destination, const RouteStep &source)
     return destination;
 }
 
+// invalidate a step and set its content to nothing
+inline void invalidateStep(RouteStep &step)
```
daniel-j-h: This is a translation unit local helper function, right? Put it either in an anonymous namespace or make it static (instead of inline).

src/engine/guidance/post_processing.cpp
```diff
@@ -174,28 +204,155 @@ void closeOffRoundabout(const bool on_roundabout,
...
+    }
+    return step;
+}
 
-        for (auto intersection : step.maneuver.intersections)
-            std::cout << "(" << intersection.duration << " " << intersection.distance << ")";
+// A check whether two instructions can be treated as one. This is only the case for very short
+// maneuvers that can, in some form, be seen as one. The additional in_step is to find out about
+// a possible u-turn.
+inline bool collapsable(const RouteStep &step)
```
daniel-j-h: anon ns / or static

src/engine/guidance/post_processing.cpp
```diff
@@ -174,28 +204,155 @@ void closeOffRoundabout(const bool on_roundabout,
...
+    const constexpr double MAX_COLLAPSE_DISTANCE = 25;
+    return step.distance < MAX_COLLAPSE_DISTANCE;
+};
+
+void collapseTurnAt(std::vector<RouteStep> &steps,
+                    const std::size_t two_back_index,
+                    const std::size_t one_back_index,
+                    const std::size_t step_index)
+{
+    const auto &current_step = steps[step_index];
```
daniel-j-h: Maybe assert on steps.size() and indices

src/engine/guidance/post_processing.cpp
```diff
@@ -174,28 +204,155 @@ void closeOffRoundabout(const bool on_roundabout,
...
+    // An instruction is invalid, if its NO_TURN and has WaypointType::None.
+    // Two valid NO_TURNs exist in each leg in the form of Depart/Arrive
+
+    // keep valid instructions
+    const auto not_is_valid = [](const RouteStep &step) {
+        return step.maneuver.instruction == TurnInstruction::NO_TURN() &&
+               step.maneuver.waypoint_type == WaypointType::None;
+    };
+
+    boost::remove_erase_if(steps, not_is_valid);
```
daniel-j-h: :+1: :P

src/engine/guidance/post_processing.cpp
```diff
@@ -287,20 +444,65 @@ std::vector<RouteStep> postProcess(std::vector<RouteStep> steps)
...
+        --index;
+        while (index > 0 && steps[index].maneuver.instruction == TurnInstruction::NO_TURN())
+            --index;
+
+        return index;
     };
 
-    boost::remove_erase_if(steps, not_is_valid);
+    // first and last instructions are waypoints that cannot be collapsed
+    for (std::size_t step_index = 2; step_index < steps.size(); ++step_index)
```
daniel-j-h: Maybe assert in steps.size() and index

src/extractor/edge_based_graph_factory.cpp
```diff
@@ -320,8 +321,11 @@ void EdgeBasedGraphFactory::GenerateEdgeExpandedEdges(
     // Three nested loop look super-linear, but we are dealing with a (kind of)
     // linear number of turns only.
     util::Percent progress(m_node_based_graph->GetNumberOfNodes());
+    SuffixTable street_name_suffix_table;
+    street_name_suffix_table.fill(lua_state);
```
daniel-j-h: pass state to ctor to not allow partially-formed objects

src/extractor/guidance/intersection_handler.cpp
```diff
@@ -283,6 +288,18 @@ void IntersectionHandler::assignFork(const EdgeID via_edge,
...
+void IntersectionHandler::assignTrivialTurns(const EdgeID via_eid,
+                                             Intersection &intersection,
+                                             const std::size_t begin,
+                                             const std::size_t end) const
+{
+    for (std::size_t index = begin; index != end; ++index)
+        if (intersection[index].entry_allowed)
+            intersection[index].turn.instruction = {
+                findBasicTurnType(via_eid, intersection[index]),
+                getTurnDirection(intersection[index].turn.angle)};
```
daniel-j-h: We don't need the index, do we? What about

    for (auto& each : intersection)
      if (pred(each))
        each.attribute = {};

daniel-j-h: Ha, screw that, we loop over begin end --- sorry for that

src/extractor/guidance/turn_handler.cpp
```diff
@@ -503,253 +415,31 @@ std::size_t TurnHandler::findObviousTurn(const EdgeID via_edge,
...
-        return intersection;
-    // handle single turn
-    if (intersection.size() - starting_at == 1)
-    {
-        if (!intersection[starting_at].entry_allowed)
-            return intersection;
+    BOOST_ASSERT(!intersection.empty());
+    BOOST_ASSERT(starting_at <= intersection.size());
+    for (auto &road : intersection)
+        road = mirror(road);
```
daniel-j-h: missing move?

src/extractor/guidance/turn_handler.cpp
```diff
@@ -503,253 +415,31 @@ std::size_t TurnHandler::findObviousTurn(const EdgeID via_edge,
...
-                    findBasicTurnType(via_edge, intersection[starting_at + 1]), second_direction};
-            if (intersection[starting_at + 2].entry_allowed)
-                intersection[starting_at + 2].turn.instruction = {
-                    findBasicTurnType(via_edge, intersection[starting_at + 2]), third_direction};
-        }
-        else if (2 >= (intersection[starting_at].entry_allowed +
-                       intersection[starting_at + 1].entry_allowed +
-                       intersection[starting_at + 2].entry_allowed))
-        {
+    std::reverse(intersection.begin() + 1, intersection.end());
```
daniel-j-h: assert on !empty()

src/extractor/guidance/turn_handler.cpp
```diff
@@ -503,253 +415,31 @@ std::size_t TurnHandler::findObviousTurn(const EdgeID via_edge,
...
-                util::SimpleLogger().Write(logWARNING)
-                    << "\troad: " << toString(road) << " Name: " << out_data.name_id
-                    << " Road Class: " << (int)out_data.road_classification.road_class;
-            }
+    // account for the u-turn in the beginning
+    const auto count = intersection.size() - starting_at + 1;
+
+    intersection = assignRightTurns(via_edge, std::move(intersection), count);
+
+    std::reverse(intersection.begin() + 1, intersection.end());
```
daniel-j-h: assert !empty()

src/extractor/guidance/turn_handler.cpp
```diff
@@ -503,253 +415,31 @@ std::size_t TurnHandler::findObviousTurn(const EdgeID via_edge,
...
+
+    for (auto &road : intersection)
+        road = mirror(road);
```
daniel-j-h: missing move here, too? also dup code

src/extractor/suffix_table.cpp
```diff
@@ -0,0 +1,48 @@
...
+    }
+    catch (const luabind::error &er)
+    {
+        util::SimpleLogger().Write(logWARNING) << er.what();
+    }
+
+    for (auto suffix : suffixes_vector)
+    {
+        boost::algorithm::to_lower(suffix);
+        suffix_set.insert(suffix);
```
daniel-j-h: I would do a bulk to_lower + insert instead if working on one item after the other.

    auto suffix_view = suffixes_vector | transformed(to_lower);
    set.insert(begin(suffixes_view), end(suffixes_view);

or similar --- this still only does a single pass over the vector and the set knows the size in advance

src/extractor/suffix_table.cpp
```diff
@@ -0,0 +1,48 @@
...
+    std::vector<std::string> suffixes_vector;
+    try
+    {
+        // call lua profile to compute turn penalty
+        luabind::call_function<void>(lua_state, "get_name_suffix_list",
+                                     boost::ref(suffixes_vector));
+    }
+    catch (const luabind::error &er)
+    {
+        util::SimpleLogger().Write(logWARNING) << er.what();
```
daniel-j-h: Do we really only want to print a warning here? Assuming our code handles empty suffix tables, not providing the corresponding Lua profile features is a user-error. Maybe check how we handle other missing Lua properties? We should at least be consistent.

include/extractor/guidance/roundabout_handler.hpp
MoKob: well, formatting changed with at least you switching to 3.8. I am on the version that you and I compared for equality, so ... your guess is as good as mine

profiles/car.lua
MoKob: I don't get what you are looking for. Could you elaborate?

src/extractor/suffix_table.cpp
MoKob: well. this is why I check for the function before. If you don't provide it, it does not fail, does it?

MoKob:
superseeded by https://github.com/Project-OSRM/osrm-backend/pull/2327

profiles/car.lua
1ec5: A change from Road to Street (Avenue, Drive) is usually significant. If the problem is that a road alternates between the two every block, then that's likely a data issue.

What about prefixes? Most cities in the U.S. use prefixes such as North or South instead of suffixes to indicate the current city quadrant. There's more ambiguity with prefixes – North Bend Road could be named after North Bend – but a change from North to South in such cases would be prettg unambiguous.

MoKob: If you feel these are significant, I can remove road/street. I could easily do a suffix check as well and do them exclusively, so not prefix + suffix. That would enable north/south changes but not 
North Bend Road to Bend.

PR #2306, opened 2016-04-23
Add functionality to write turn lookups and read turn penalty files

lbud:
Reopened from https://github.com/Project-OSRM/osrm-backend/pull/2291.

### DIFF
--- a/include/engine/routing_algorithms/routing_base.hpp
+++ b/include/engine/routing_algorithms/routing_base.hpp
@@ -301,8 +301,6 @@ template <class DataFacadeT, class Derived> class BasicRoutingInterface
                 auto total_weight = std::accumulate(weight_vector.begin(), weight_vector.end(), 0);
 
                 BOOST_ASSERT(weight_vector.size() == id_vector.size());
-                // ed.distance should be total_weight + penalties (turn, stop, etc)
-                BOOST_ASSERT(ed.distance >= total_weight);
                 const bool is_first_segment = unpacked_path.empty();
 
                 const std::size_t start_index =
@@ -350,7 +348,8 @@ template <class DataFacadeT, class Derived> class BasicRoutingInterface
                 start_index =
                     id_vector.size() - phantom_node_pair.source_phantom.fwd_segment_position - 1;
             }
-            end_index = id_vector.size() - phantom_node_pair.target_phantom.fwd_segment_position - 1;
+            end_index =
+                id_vector.size() - phantom_node_pair.target_phantom.fwd_segment_position - 1;
         }
         else
         {
@@ -396,8 +395,17 @@ template <class DataFacadeT, class Derived> class BasicRoutingInterface
             // However the first segment duration needs to be adjusted to the fact that the source
             // phantom is in the middle of the segment. We do this by subtracting v--s from the
             // duration.
-            BOOST_ASSERT(unpacked_path.front().duration_until_turn >= source_weight);
-            unpacked_path.front().duration_until_turn -= source_weight;
+
+            // Since it's possible duration_until_turn can be less than source_weight here if
+            // a negative enough turn penalty is used to modify this edge weight during
+            // osrm-contract, we clamp to 1 here so as not to return a negative duration
+            // for this segment.
+
+            // TODO this creates a scenario where it's possible the duration from a phantom
+            // node to the first turn would be the same as from end to end of a segment,
+            // which is obviously incorrect and not ideal...
+            unpacked_path.front().duration_until_turn =
+                std::max(unpacked_path.front().duration_until_turn - source_weight, 0);
         }
 
         // there is no equivalent to a node-based node in an edge-expanded graph.
--- a/src/contractor/contractor.cpp
+++ b/src/contractor/contractor.cpp
@@ -20,6 +20,7 @@
 
 #include <boost/assert.hpp>
 #include <boost/filesystem/fstream.hpp>
+#include <boost/functional/hash.hpp>
 
 #include <tbb/parallel_sort.h>
 
@@ -29,6 +30,7 @@
 #include <memory>
 #include <thread>
 #include <iterator>
+#include <tuple>
 
 namespace std
 {
@@ -38,6 +40,18 @@ template <> struct hash<std::pair<OSMNodeID, OSMNodeID>>
     std::size_t operator()(const std::pair<OSMNodeID, OSMNodeID> &k) const
     {
         return static_cast<uint64_t>(k.first) ^ (static_cast<uint64_t>(k.second) << 12);
+    }
+};
+
+template <> struct hash<std::tuple<OSMNodeID, OSMNodeID, OSMNodeID>>
+{
+    std::size_t operator()(const std::tuple<OSMNodeID, OSMNodeID, OSMNodeID> &k) const
+    {
+        std::size_t seed = 0;
+        boost::hash_combine(seed, static_cast<uint64_t>(std::get<0>(k)));
+        boost::hash_combine(seed, static_cast<uint64_t>(std::get<1>(k)));
+        boost::hash_combine(seed, static_cast<uint64_t>(std::get<2>(k)));
+        return seed;
     }
 };
 }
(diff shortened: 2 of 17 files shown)


lbud:
k @danpat @TheMarex @MoKob and/or @daniel-j-h 🌴 have at it

src/extractor/edge_based_graph_factory.cpp
```diff
@@ -431,6 +431,33 @@ void EdgeBasedGraphFactory::GenerateEdgeExpandedEdges(
...
+                    // to penalize certain turns.
+
+                    const bool isCompressed =
+                        m_compressed_edge_container.ContainsCompressedGeometries(edge_from_u);
+
+                    const auto &from_node =
+                        isCompressed
+                            ? m_node_info_list[m_compressed_edge_container.GetLastEdgeSourceID(
+                                  edge_from_u)]
+                            : m_node_info_list[node_u];
```
danpat: The behaviour here looks right, but I think the wording could be changed.  I think if I were to come back to this in a few months, `isCompressed` and `ContainsCompressedGeometries` might confuse me, as technically, all edges are in the `CompressedGeometryContainer`.

How about `isTrivial` and `isTrivial()`, and the some comments in the `CompressdGeometryContainer` to explain what `Trivial` means in this context.

include/engine/routing_algorithms/routing_base.hpp
```diff
@@ -301,8 +301,6 @@ template <class DataFacadeT, class Derived> class BasicRoutingInterface
                 auto total_weight = std::accumulate(weight_vector.begin(), weight_vector.end(), 0);
 
                 BOOST_ASSERT(weight_vector.size() == id_vector.size());
-                // ed.distance should be total_weight + penalties (turn, stop, etc)
-                BOOST_ASSERT(ed.distance >= total_weight);
```
TheMarex: Why was this assertion removed? I don't see why this should be wrong unless there are negative penalties?

src/contractor/contractor.cpp
```diff
@@ -40,6 +41,16 @@ template <> struct hash<std::pair<OSMNodeID, OSMNodeID>>
...
     }
 };
+
+template <> struct hash<std::tuple<OSMNodeID, OSMNodeID, OSMNodeID>>
+{
+    std::size_t operator()(const std::tuple<OSMNodeID, OSMNodeID, OSMNodeID> &k) const
+    {
+        return (static_cast<uint64_t>(std::get<0>(k)) << 24) ^
+               (static_cast<uint64_t>(std::get<1>(k)) << 12) ^
+               (static_cast<uint64_t>(std::get<2>(k)));
```
TheMarex: Instead of an own hash functio you should consider using `boost:hash_combine`:

http://www.boost.org/doc/libs/1_55_0/doc/html/hash/combine.html

src/extractor/compressed_edge_container.cpp
```diff
@@ -250,16 +250,25 @@ CompressedEdgeContainer::GetBucketReference(const EdgeID edge_id) const
...
 NodeID CompressedEdgeContainer::GetFirstEdgeTargetID(const EdgeID edge_id) const
 {
     const auto &bucket = GetBucketReference(edge_id);
-    BOOST_ASSERT(bucket.size() >= 2);
     return bucket.front().node_id;
 }
+NodeID CompressedEdgeContainer::GetLastEdgeTargetID(const EdgeID edge_id) const
+{
+    const auto &bucket = GetBucketReference(edge_id);
+    return bucket[bucket.size() - 1].node_id;
```
TheMarex: Use `bucket.back().node_id` instread.

src/extractor/compressed_edge_container.cpp
```diff
@@ -250,16 +250,25 @@ CompressedEdgeContainer::GetBucketReference(const EdgeID edge_id) const
...
+}
 NodeID CompressedEdgeContainer::GetLastEdgeSourceID(const EdgeID edge_id) const
 {
     const auto &bucket = GetBucketReference(edge_id);
-    BOOST_ASSERT(bucket.size() >= 2);
```
TheMarex: Don't remove but correct assertion for uncompressed case: `BOOST_ASSERT(bucket.size() >= 1);`

include/engine/routing_algorithms/routing_base.hpp
danpat: @TheMarex yes, that's exactly the case this was removed to support.

The situation where this might arise would be where segment speeds are being calculated using GPS probe data.  At an intersection where the majority of the traffic makes a turn, the average speed will be primarily that of turning traffic.  However, if the intersection also allows unimpeded straight-through movements, a negative "turn" value can be used to correct traversal time for traffic that proceeds straight and unimpeded.  This type of situation is reasonably common on the outskirts of urban areas across the US, example: http://www.openstreetmap.org/#map=18/38.79196/-121.35847

Really though, it would apply in any situation where an average measurement does not represent the fastest traffic through an intersection, and a "speedup" needs to be applied for some movements.

lbud:
Added a few comments for clarity, replaced the removed assertions, renamed the new compressed geometry fns/vars to `isTrivial`/`IsTrivial()`, and rebased on master. As long as https://github.com/Project-OSRM/osrm-backend/pull/2306#discussion_r60853416 is okay I think this is ready — 🚦✅ and I'll squash and merge.

TheMarex:
@lbud tests on travis seem to be failing. Otherwise :+1: for merging if that is fixed.

include/engine/routing_algorithms/routing_base.hpp
```diff
@@ -397,8 +396,17 @@ template <class DataFacadeT, class Derived> class BasicRoutingInterface
...
+            // Since it's possible duration_until_turn can be less than source_weight here if
+            // a negative enough turn penalty is used to modify this edge weight during
+            // osrm-contract, we clamp to 1 here so as not to return a negative duration
+            // for this segment.
+
+            // TODO this creates a scenario where it's possible the duration from a phantom
+            // node to the first turn would be the same as from end to end of a segment,
+            // which is obviously incorrect and not ideal...
+            unpacked_path.front().duration_until_turn =
+                std::max(unpacked_path.front().duration_until_turn - source_weight, 1);
```
lbud: @TheMarex would love another set of 👀  on this — it feels like maybe sort of a hacky solution to potentially negative edge durations, but… ??
cc @danpat

src/contractor/contractor.cpp
```diff
@@ -494,7 +545,34 @@ std::size_t Contractor::LoadEdgeExpandedGraph(
...
+
+                if (new_turn_weight + new_weight < original_node_count)
+                {
+                    util::SimpleLogger().Write(logWARNING)
+                        << "turn penalty " << turn_iter->second.first << " for turn " << from_id
+                        << ", " << via_id << ", " << to_id
+                        << " is too negative: clamping turn weight to " << original_node_count;
+                }
+
+                inbuffer.weight = std::max(new_turn_weight + new_weight, original_node_count);
```
lbud: @TheMarex @daniel-j-h @MoKob @danpat 

I wasn't sure whether to clamp this to `original_node_count` (same as `num_osm_nodes` before it gets modified), `original_node_count * 10`, or `1` (or something else)… 🍝 👾  help?

lbud:
Travis tests were failing because I was testing a debug assertion in cucumber, so all release builds weren't erroring where I expected they should — which exposed the issue that we really needed to catch any negative edge weight regardless of build type, hence https://github.com/Project-OSRM/osrm-backend/pull/2306/commits/663c28ee6b92b90f23eec5b6f0312aedd2c513ff. This raised some additional questions which I've added in comments here.

TheMarex:
@lbud can you rebase again on `master`? Seems like the travis breakages are mostly due to me accidentally breaking master.

lbud:
👌 @TheMarex just rebased after renaming `original_node_count` to `compressed_edge_nodes`

include/engine/routing_algorithms/routing_base.hpp
TheMarex: Actually, this explains why the test fails above. If the phantom node snapps to a segment end point the result of this used to be `0`. using `0` instead of `1` here should yield the expected behavior.

lbud:
Just forced pushed after squash/rebase and then I'll merge.

src/extractor/compressed_edge_container.cpp
```diff
@@ -250,16 +250,32 @@ CompressedEdgeContainer::GetBucketReference(const EdgeID edge_id) const
...
+    BOOST_ASSERT(bucket.size() >= 1);
+    return bucket.back().node_id;
+}
 NodeID CompressedEdgeContainer::GetLastEdgeSourceID(const EdgeID edge_id) const
 {
     const auto &bucket = GetBucketReference(edge_id);
-    BOOST_ASSERT(bucket.size() >= 2);
+    BOOST_ASSERT(bucket.size() >= 1);
     return bucket[bucket.size() - 2].node_id;
 }
```
MoKob: `bucket.size() >=1` and access to `bucket.size()-2` seems like a bad idea?

TheMarex: Derp. That might have been my ill-advised comment to consider that non-compressed geometries. Yes this should be `2`.

PR #2312, opened 2016-04-24
Stateless parameters grammars and fix for PhantomNode bitfields packing in MSVC

oxidase:
Hello,
i have made a small refactoring of parameters grammars by switching signatures from "void" to "void(Parameters&)". Parameter structures now can be passed by references into grammars and grammars can be defined as "static const". It brings performance improvement in release mode from 
2933ns to 372ns for small queries, like
```
BOOST_AUTO_TEST_CASE(time)
{
    std::string str{"1,2;3,4;5,6.json"};

    int n = 100000;
    std::chrono::steady_clock::time_point start = std::chrono::steady_clock::now();
    for (int i=0;i<n;++i)
        api::parseParameters<engine::api::TripParameters>(str);
    std::chrono::steady_clock::time_point end = std::chrono::steady_clock::now();
    std::cout << "time is "<< std::chrono::duration_cast<std::chrono::nanoseconds>((end - start)/n).count() << "ns.\n";
}
```
because most of the time (80%-90%) was spent in construction and destruction of grammars.

The behavior is not changed but for two cases: "&geometries=foo" and "&overview=foo"  fail now at "foo"  instead of "&geometries=foo" and "&overview=foo". Also there some fixes listed in the commit message that do not change behavior but make grammars simpler.

The second commit fixes packing in MSVS to make `valid_route_urls` test pass because grammar expects ENCODED_HINT_SIZE characters that was different in MSVS. The fix just changes bool to uint32_t for 1 bit flags to make tight packing, because MSVS for any new type (here bool) will start at byte alignment and can give 1,2 or 4 additional bytes in the structure memory footprint.

### DIFF
--- a/include/engine/hint.hpp
+++ b/include/engine/hint.hpp
@@ -63,18 +63,9 @@ struct Hint
     friend std::ostream &operator<<(std::ostream &, const Hint &);
 };
 
-#ifndef _MSC_VER
 static_assert(sizeof(Hint) == 60 + 4, "Hint is bigger than expected");
 constexpr std::size_t ENCODED_HINT_SIZE = 88;
-static_assert(ENCODED_HINT_SIZE / 4 * 3 >= sizeof(Hint),
-              "ENCODED_HINT_SIZE does not match size of Hint");
-#else
-// PhantomNode is bigger under windows because MSVC does not support bit packing
-static_assert(sizeof(Hint) == 72 + 4, "Hint is bigger than expected");
-constexpr std::size_t ENCODED_HINT_SIZE = 104;
-static_assert(ENCODED_HINT_SIZE / 4 * 3 >= sizeof(Hint),
-              "ENCODED_HINT_SIZE does not match size of Hint");
-#endif
+static_assert(ENCODED_HINT_SIZE / 4 * 3 >= sizeof(Hint), "ENCODED_HINT_SIZE does not match size of Hint");
 }
 }
 
--- a/include/engine/phantom_node.hpp
+++ b/include/engine/phantom_node.hpp
@@ -149,13 +149,11 @@ struct PhantomNode
     unsigned reverse_packed_geometry_id;
     struct ComponentType
     {
-        uint32_t id : 31;
-        bool is_tiny : 1;
+        std::uint32_t id         : 31;
+        std::uint32_t is_tiny    : 1;
     } component;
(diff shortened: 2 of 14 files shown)


daniel-j-h:
Awesome work :+1:

Stateless grammars is exactly what we need. The fix for `uint32_t; bool` vs `uint32_t; uint32_t` is interesting --- from what I can find in the standard this is implementation defined (cf. [9.2.4/1](http://eel.is/c++draft/class.bit#1)).

oxidase:
Hi, 4363fd6 is  a proposal rather than a solution, so i made it separate. It is very MSVC-specific, anyway i just removed  MSC_VER conditions and static asserts must guard from different field packings.  Other possibility would be to make the failing test platform-dependent and keep different sizes of hints, but it brings even more troubles, because size of Hint will depend on the servers platform and a gcc-compiled server will reject cl-compiled hints and vice versa.  

Imho, the correct solution would be to avoid using bitfields, use fixed size types (here uint32_t), and make  explicit getter/setters for bits. I can provide a quick fix for this.

TheMarex:
Great work. :+1: ~10x speedup is something we don't get every day.

Having a bit-field method that works across gcc, clang and MSVC is good enough I think, even if its implementation defined. Thanks for making this work, this saves tons of nasty logic! 

There might be a few other locations were we can utilize this to make MSVC a first class compile target.

oxidase:
Thanks! For quick fix it does not look for me like "quick" anymore, because I have underestimated efforts to change to smth like
```
struct SegmentID
{
    SegmentID(const NodeID id, const bool enabled)
    {
        BOOST_ASSERT(!enabled || id != SPECIAL_SEGMENTID);
        osrm::util::setBits<0, 31, uint32_t>(data, id);
        osrm::util::setBits<31, 1, uint32_t>(data, enabled);
    }

    NodeID id() const
    {
        return static_cast<NodeID>(osrm::util::getBits<0, 31>(data));
    }

    bool enabled() const
    {
        return static_cast<bool>(osrm::util::getBits<31, 1>(data));
    }

    void setEnabled(const bool value)
    {
        osrm::util::setBits<31, 1, uint32_t>(data, value);
    }
private:
    uint32_t data;
};
```
It would require too many substitutions, like s/id/id()/, but brings almost no benefits over bitfields, just fixed type size and explicit bits positions.

PR #2313, opened 2016-04-24
Do not copy geometries in loop

daniel-j-h:
Again, copying geometries all over the place. This time in 1/ simplification and 2/ overview.


TheMarex:
Good catch. :+1: Good to merge after rebase.

daniel-j-h:
Almost two hours later and Travis is still not done building this after the rebase + forced push.

I'm not sure if requiring those checks is a good idea. Just think about what happens if during those two hours someone made a commit on master. Then I have to rebase and force push again, waiting another two hours hoping that _this time_ no one makes a commit on master?

PR #2314, opened 2016-04-24
[DONT MERGE] Debug coveralls

TheMarex:


TheMarex:
We switched to codecov

Issue #2309, opened 2016-04-23
V5 - Route Algorithm Crash => Assert start_index <= end_index


K-Leon:
Hello,

I just gave the new V5 Version a spin ( great work btw! ) and stumbled over an issue in Route Plugin.

> [warn] [server error] code: /var/osrm/osrm-backend/include/engine/routing_algorithms/routing_base.hpp:376
> in: void osrm::engine::routing_algorithms::BasicRoutingInterface<DataFacadeT, Derived>::UnpackPath(RandomIter, RandomIter, const osrm::engine::PhantomNodes&, std::vector<osrm::engine::PathData>&) const [with RandomIter = __gnu_cxx::__normal_iterator<unsigned int*, std::vector<unsigned int> >; DataFacadeT = osrm::engine::datafacade::BaseDataFacade; Derived = osrm::engine::routing_algorithms::DirectShortestPathRouting<osrm::engine::datafacade::BaseDataFacade>]: start_index <= end_index
> , uri: /route/v1/driving/8.7475533,47.4976083;8.7475417,47.4976133?geometries=polyline&steps=false

URL: [router.project-osrm.org/route/v1/driving/8.7475533,47.4976083;8.7475417,47.4976133?geometries=polyline&steps=false](http://router.project-osrm.org/route/v1/driving/8.7475533,47.4976083;8.7475417,47.4976133?geometries=polyline&steps=false)

@daniel-j-h I think it could be very similar to issue #2147 you fixed for v4 trip Plugin (thanks again). These coords also resolve more or less to the same Location as far as i can tell. 

One of these 'situations' in a long Routing Request is enough.

Best Regards,
Leon

daniel-j-h:
@MoKob it seems like you introduced the assertion in https://github.com/Project-OSRM/osrm-backend/commit/ef1fb08723565bc968a64d0cba4b947d1a6746ef#diff-ffa2e7ac24f3bebb547c6fe24552431bR376 --- care to take a look?

MoKob:
Should work with: https://github.com/Project-OSRM/osrm-backend/pull/2316

TheMarex:
This is a critical bug and will be included in `5.0.1`.

PR #2316, opened 2016-04-25
fix local path loos

MoKob:
Addressing https://github.com/Project-OSRM/osrm-backend/issues/2309.

I reverted the change to the loop. It seems to be a border case that I sadly cannot reproduce in cucumber. Nevertheless, the function should work fine this way. Even if its only necessary for looking up a single coordinate.

### DIFF
--- a/include/engine/routing_algorithms/routing_base.hpp
+++ b/include/engine/routing_algorithms/routing_base.hpp
@@ -373,8 +373,7 @@ template <class DataFacadeT, class Derived> class BasicRoutingInterface
         // t: fwd_segment 3
         // -> (U, v), (v, w), (w, x)
         // note that (x, t) is _not_ included but needs to be added later.
-        BOOST_ASSERT(start_index <= end_index);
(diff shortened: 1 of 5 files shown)


include/engine/routing_algorithms/routing_base.hpp
```diff
@@ -374,7 +374,7 @@ template <class DataFacadeT, class Derived> class BasicRoutingInterface
         // -> (U, v), (v, w), (w, x)
         // note that (x, t) is _not_ included but needs to be added later.
         BOOST_ASSERT(start_index <= end_index);
-        for (std::size_t i = start_index; i != end_index; ++i)
+        for (std::size_t i = start_index; i != end_index; (start_index < end_index ? ++i : --i))
```
daniel-j-h: How does this fix the `start_index <= end_index` assertion on line 376 being triggered?

MoKob: this fixes the case, I had the assertion in to try and generate a cucumber test and forgot to take it out again :(

PR #2323, opened 2016-04-26
[Experimental] Option to use TCMalloc

daniel-j-h:
Only for osrm-routed at the moment. Experimental. Needs benchmarking. Feel free to customize.

(check the refs. at the bottom, they also provide a heap profiler, heap checker and cpu profiles)

References:
- https://github.com/gperftools/gperftools/wiki
- https://github.com/gperftools/gperftools
- http://htmlpreview.github.io/?https://github.com/gperftools/gperftools/blob/master/doc/tcmalloc.html

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -29,6 +29,7 @@ if(WIN32 AND MSVC_VERSION LESS 1800)
   message(FATAL_ERROR "Building with Microsoft compiler needs Visual Studio 2013 or later (Express version works too)")
 endif()
 
+option(ENABLE_TCMALLOC "Use the TCMalloc memory allocator" OFF)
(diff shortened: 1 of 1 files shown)


TheMarex:
I benchmarked this locally. Surprisingly TCMalloc is consistently slower than the buildin malloc. Maybe their allocator strategy is not really what we need?

daniel-j-h:
Interesting to know. Initially I wanted to use jemalloc to combat fragmentation but it's a pain to integrate. Feel free to close here and let's not move on with this experimental idea.

TheMarex:
Yeah its good to have this in mind for the future. I guess custom allocators won't win us too much. :-/

PR #2324, opened 2016-04-26
Fix syntax error position indicators in parameters queries

### LINKED ISSUE #2193: Query syntax error position indicators
When there's a syntax error in the query part of a URL, but the service/version/profile part parses OK, the error position in the message refers to the error offset within the query part only, not the whole URL.

This is a change from previous behaviour, the syntax error position was previously relative to the first char in the PATH_INFO part of the URL.

Example:

`route/v1/driving`
gives
`URL string malformed close to position 0: "/ro"`

but

`route/v1/driving/1,1,1`
gives
`Query string malformed close to position 3`

It's not technically wrong, but it's not obvious what `position 3` is relative to.

oxidase:
To fix #2193 prefix_length member variable has been added to ParsedURL
that is set to the length of "/service/version/profile/" prefix
when the prefix is accepted by the parser.
Also BOOST_FUSION_ADAPT_STRUCT for osrm::server::api::ParsedURL
has been moved from header to url_parser.cpp to speed up compilation
of CUs that do not use the fusion adaption.

### DIFF
--- a/features/testbot/status.feature
+++ b/features/testbot/status.feature
@@ -55,13 +55,13 @@ Feature: Status messages
             | ?                           | 400    | URL string malformed close to position 1: "/?"    |
             | route/v1/driving            | 400    | URL string malformed close to position 17: "ing"  |
             | route/v1/driving/           | 400    | URL string malformed close to position 18: "ng/"  |
-            | route/v1/driving/1          | 400    | Query string malformed close to position 1        |
+            | route/v1/driving/1          | 400    | Query string malformed close to position 19       |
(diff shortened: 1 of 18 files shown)


TheMarex:
@oxidase Not pretty but I also don't really see a more elegant way. Thanks for the great work. :+1:

TheMarex:
Can you rebase? I would like to merge tomorrow for the `5.1.0` release that is shaping up quickly.

oxidase:
ok, rebased. I also could not find any nicer way to adjust the error position without big changes in the URL dispatcher.

Issue #2267, opened 2016-04-15
Visual Indicators / Signs


MoKob:
To further improve upon our guidance, we want to handle sign information and visual indicators. These indicators can be a list of symbols (e.g. airport), or destination signs / highway names. Wherever these `signs` are located, they can help to disambiguate guidance. Think `follow the signs towards San Francisco` over `turn right and take the ramp`.

### Modelled Data

 - [destination=*](http://wiki.openstreetmap.org/wiki/Key:destination) Seems to be the method of choice. It can be found at `nodes`, `ways`, and `relations`.
 

 - [Relation:destination_sign](http://wiki.openstreetmap.org/wiki/Relation:destination_sign), rarely used right now. 99 occurrences according to taginfo. Can probably be ignored.
 
 
## Tasks

 - [x] inspect data / find out how to parse data correctly
 	- [ ] nodes
 	- [x] ways
 	- [ ] relations
 	
Based on the results of the inspection, we have to find the best way to store the information. Right now all information for turns is stored with the edge-based edges. Most of them however, will not have destination signs assigned to them.

A possible way could be to store a lookup-table based on the edge-based edge-ids to additional turn-information. Only if such a sign is present, the data can be added to the turn in the postprocessing.

Storing the data directly with the edges is not reasonable, unless we manage to do some variable length encoding. The reasoning behind my assumption here is that the vast majority of turns (e.g. residential areas) does not contain destination signs.

 - [x] store turn data, either as look-up table or
 - [x] implement variable length encoding of edge data
 
Finally, the data has to be exposed to the API. This is a simple look-up in the final stages of the API when we generate the response. Both for having a dedicated look-up table or for variable length-encoding.
 
 - [x] expose sign data

emiltin:
can this be combined with #1342?

willwhite:
> Think `follow the signs towards San Francisco` over `turn right and take the ramp`.

Would this include incorporating exit numbers into instructions as well? E.g. `take exit 18 and follow the signs towards San Francisco`

MoKob:
@willwhite essentially any part of data that we can get should be included in the response.

so `exit nr`, `visual indicators`, and `destination names` would be on my list right now. How the API expresses the available data into strings is something to discuss on the frontend-side of things.

pratikyadav:
> destination=* Seems to be the method of choice. It can be found at nodes, ways, and relations.

[...]

PR #2328, opened 2016-04-27
Basic destination sign handling.

daniel-j-h:
This first part in our Exit Sign / Destination Sign quest re-wires
the `destination=` and `destination:ref=` tag to the `ref=` tag (in case
it's the highway does not already have a `ref=` tag).

Doing some analysis on both Berlin and San Francisco ([here](https://github.com/Project-OSRM/osrm-backend/issues/2267#issuecomment-215001531)) it looks like this
will add a couple of thousand `ref=` tags that we will announce in guidance.

References:
- https://github.com/Project-OSRM/osrm-backend/issues/2267#issuecomment-214771682
- http://wiki.openstreetmap.org/wiki/Key:destination
- http://wiki.openstreetmap.org/wiki/Proposed_features/Destination_details
- http://taginfo.openstreetmap.org/keys/destination
- http://taginfo.openstreetmap.org/keys/destination%3Aref
- http://taginfo.openstreetmap.org/search?q=destination

### DIFF
--- a/features/guidance/destination-signs.feature
+++ b/features/guidance/destination-signs.feature
@@ -15,19 +15,21 @@ Feature: Destination Signs
           | m | n |
 
         And the ways
-          | nodes | name | ref | destination | destination:ref |
-          | ab    | AB   | E1  |             |                 |
-          | cd    | CD   |     | Berlin      |                 |
-          | ef    | EF   |     | Berlin      | A1              |
-          | gh    |      |     | Berlin      | A1              |
-          | ij    |      |     | Berlin      |                 |
-          | kl    | KL   | E1  | Berlin      | A1              |
+          | nodes | name | ref | destination    | destination:ref |
+          | ab    | AB   | E1  |                |                 |
+          | cd    | CD   |     | Berlin         |                 |
+          | ef    | EF   |     | Berlin         | A1              |
+          | gh    |      |     | Berlin         | A1              |
+          | ij    |      |     | Berlin         |                 |
+          | kl    | KL   | E1  | Berlin         | A1              |
+          | mn    | MN   |     | Berlin;Hamburg | A1;A2           |
 
         When I route I should get
-          | from | to | route                           |
-          | a    | b  | AB (E1),AB (E1)                 |
-          | c    | d  | CD (Berlin),CD (Berlin)         |
-          | e    | f  | EF (A1: Berlin),EF (A1: Berlin) |
-          | g    | h  | ,                               |
-          | i    | j  | ,                               |
-          | k    | l  | KL (E1),KL (E1)                 |
+          | from | to | route                                                     |
+          | a    | b  | AB (E1),AB (E1)                                           |
+          | c    | d  | CD (Berlin),CD (Berlin)                                   |
+          | e    | f  | EF (A1: Berlin),EF (A1: Berlin)                           |
+          | g    | h  | ,                                                         |
+          | i    | j  | ,                                                         |
+          | k    | l  | KL (E1),KL (E1)                                           |
+          | m    | n  | MN (A1, A2: Berlin, Hamburg),MN (A1, A2: Berlin, Hamburg) |
--- a/features/guidance/destination-signs.feature
+++ b/features/guidance/destination-signs.feature
@@ -14,6 +14,7 @@ Feature: Destination Signs
           | k | l |
           | m | n |
           | o | p |
+          | q | r |
 
         And the ways
           | nodes | name | ref | destination    | destination:ref | oneway | #                                    |
@@ -25,6 +26,7 @@ Feature: Destination Signs
           | kl    | KL   | E1  | Berlin         | A1              | yes    |                                      |
           | mn    | MN   |     | Berlin;Hamburg | A1;A2           | yes    |                                      |
           | op    | OP   |     | Berlin;Hamburg | A1;A2           | no     | mis-tagged destination: not a oneway |
+          | qr    | QR   |     |                | A1;A2           | yes    |                                      |
 
         When I route I should get
           | from | to | route                                                     | #                         |
@@ -36,3 +38,4 @@ Feature: Destination Signs
           | k    | l  | KL (E1),KL (E1)                                           |                           |
           | m    | n  | MN (A1, A2: Berlin, Hamburg),MN (A1, A2: Berlin, Hamburg) |                           |
           | o    | p  | OP,OP                                                     | guard against mis-tagging |
+          | q    | r  | QR (A1, A2),QR (A1, A2)                                   |                           |
--- a/features/guidance/destination-signs.feature
+++ b/features/guidance/destination-signs.feature
@@ -1,0 +1,33 @@
+@routing @guidance
+Feature: Destination Signs
+
+    Background:
+        Given the profile "car"
+
+    Scenario: Car - route name assembly with destination signs
+        Given the node map
+          | a | b |
+          | c | d |
+          | e | f |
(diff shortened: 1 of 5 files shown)


features/guidance/destination-signs.feature
```diff
@@ -0,0 +1,33 @@
...
+          | ij    |      |     | Berlin      |                 |
+          | kl    | KL   | E1  | Berlin      | A1              |
+
+        When I route I should get
+          | from | to | route                           |
+          | a    | b  | AB (E1),AB (E1)                 |
+          | c    | d  | CD (Berlin),CD (Berlin)         |
+          | e    | f  | EF (A1: Berlin),EF (A1: Berlin) |
+          | g    | h  | ,                               |
+          | i    | j  | ,                               |
```
daniel-j-h: Still not sure if we can and should use `destination=` and `destination:ref=` when there's neither a `name=` nor a `ref=` on the highway. It's bad tagging then anyway, and I don't want to announce a street as "Berlin" just because it is its destination.

profiles/car.lua
```diff
@@ -343,15 +344,19 @@ function way_function (way, result)
...
   local has_ref = ref and "" ~= ref
   local has_name = name and "" ~= name
+  local has_destination = destination and "" ~= destination
 
   if has_name and has_ref then
     result.name = name .. " (" .. ref .. ")"
   elseif has_ref then
     result.name = ref
+  elseif has_name and has_destination then
+    result.name = name .. " (" .. destination .. ")"
```
daniel-j-h: - [ ] Do the same for foot / walk profile

I only now realized I hooked the lib function in only for car.

profiles/lib/destination.lua
```diff
@@ -0,0 +1,27 @@
+local Destination = {}
+
+function Destination.get_destination(way)
+  local destination = way:get_value_by_key("destination")
+  local destination_ref = way:get_value_by_key("destination:ref")
```
daniel-j-h: In theory we could make use of some less-popular tags here, too:
http://taginfo.openstreetmap.org/search?q=destination

And some of the proposed features:
http://wiki.openstreetmap.org/wiki/Proposed_features/Destination_details

But for a start this should give of the most popular ones.

profiles/lib/destination.lua
```diff
@@ -0,0 +1,27 @@
...
+
+  -- Assemble destination as: "A59: Düsseldorf, Köln"
+  --          destination:ref  ^    ^  destination
```
daniel-j-h: This is starting to get ugly: we try to encode destinations into the street name as

    $name ($destination:ref: $destination)

in case we can not use the default of

    $name ($ref)

profiles/lib/destination.lua
```diff
@@ -0,0 +1,27 @@
...
+  if destination_ref and destination_ref ~= "" then
+    rv = rv .. destination_ref
+  end
+
+  if destination and destination ~= "" then
+      if rv ~= "" then
+          rv = rv .. ": "
+      end
+
+      rv = rv .. string.gsub(destination, ";", ", ")
```
daniel-j-h: `destination=*` can have `;` separated values, but it's ugly for the user.
Not sure if encoding it this way makes that much more sense.

features/guidance/destination-signs.feature
```diff
@@ -0,0 +1,35 @@
...
+          | kl    | KL   | E1  | Berlin         | A1              |
+          | mn    | MN   |     | Berlin;Hamburg | A1;A2           |
+
+        When I route I should get
+          | from | to | route                                                     |
+          | a    | b  | AB (E1),AB (E1)                                           |
+          | c    | d  | CD (Berlin),CD (Berlin)                                   |
+          | e    | f  | EF (A1: Berlin),EF (A1: Berlin)                           |
+          | g    | h  | ,                                                         |
+          | i    | j  | ,                                                         |
```
MoKob: do we really want to expose a `,` to the outside here?

daniel-j-h: I thought this was cucumber assembling the responses? I got this from https://github.com/Project-OSRM/osrm-backend/blob/e88106e990336e6b38dfadbdd85f147c6f20a0f8/features/guidance/ramp.feature#L228-L229 as I could not find documentation on it.

daniel-j-h:
- [ ] `destination=*` only in combination with `oneway=yes`, everything else is a data issue

daniel-j-h:
Issue I was talking about: https://github.com/Project-OSRM/osrm-backend/pull/2328/commits/52613727615271cbdee1284668595d3e04affecc => no `exit_to` node tags.

profiles/lib/destination.lua
MoKob: does this work for items with `destination:ref` and no `destination` tag?

features/guidance/destination-signs.feature
```diff
@@ -0,0 +1,38 @@
...
+          | nodes | name | ref | destination    | destination:ref | oneway | #                                    |
+          | ab    | AB   | E1  |                |                 | yes    |                                      |
+          | cd    | CD   |     | Berlin         |                 | yes    |                                      |
+          | ef    | EF   |     | Berlin         | A1              | yes    |                                      |
+          | gh    |      |     | Berlin         | A1              | yes    |                                      |
+          | ij    |      |     | Berlin         |                 | yes    |                                      |
+          | kl    | KL   | E1  | Berlin         | A1              | yes    |                                      |
+          | mn    | MN   |     | Berlin;Hamburg | A1;A2           | yes    |                                      |
+          | op    | OP   |     | Berlin;Hamburg | A1;A2           | no     | mis-tagged destination: not a oneway |
+
```
MoKob: what about `destination=""`, `destination:ref=A1`

profiles/lib/destination.lua
daniel-j-h: Yes, in that case you only get the `destination:ref` as in "Berlin (A5)". I'm missing a test for this combination, thanks for pointing it out! Will add one.

features/guidance/destination-signs.feature
daniel-j-h: Yep that combination is covered by the lib function but not in the tests --- thanks, will add a cucumber test for this.

daniel-j-h:
This is done, Part II will happen in its own PR as it's conceptually different.
@TheMarex feel free to take a look and then merge.

profiles/lib/destination.lua
TheMarex: You are adding a new dependency to `car.lua`. This means all scripts that are fetching `car.lua` from OSRM (`node-osrm`, `api-directions`, `directions-processing`) will need to get adjusted. This warrants a changelog entry.

/cc @karenzshea @willwhite @miccolis

src/extractor/scripting_environment.cpp
```diff
@@ -43,6 +44,11 @@ template <class T> double lonToDouble(T const &object)
     return static_cast<double>(util::toFloating(object.lon));
 }
 
+// Luabind does not like memr funs: instead of casting to the function's signature (mem fun ptr) we simply wrap it
```
TheMarex: This function is not used? Either split out or remove.

daniel-j-h: You are right, this was my try at making `osmium::Node`s available in the `way_function`.
Which is not possible, all we get are `osmium::NodeRef`s and they only have an id but no tags and accessing their location crashes since we're not using `libosmium`'s `NodeLocationsForWays` handler :sigh:. Splitting out the commit.

PR #2333, opened 2016-04-27
Revert "Revert "switched to space separated suffix/prefix only""

TheMarex:
Reopening of #2332 because it broke master. Sorry about that, I though travis had already green lighted this branch.


MoKob:
Yeah. Should have said that we would need for travis. Just proves to show, don't code when you really need to sleep. Issue is fixed now

PR #2337, opened 2016-04-28
[NOT READY] Feature/expose node ids

TheMarex:
Reopening #2165 

# This is to address #2054

This change loads the original OSM node ID values into memory along with coordinates, instead of just discarding them.

They are then appended to the `route` object in `trip` and `match` responses, so that the route through the original OSM data can be followed.  These values could then be used with an offline service to get detailed metadata about the route that OSRM does not have the capacity to store/expose.

Current status: 

  - works only with the `InternalDataFacade`
  - will always be 2 nodes short - the nodes just before and just after the PhantomNodes - not sure whether to include these or not, or include a null value so that the node array aligns with the coordinate array.

Edit by @TheMarex:

Todo:
- [ ] Implement shared memory loading
- [ ] Investigate memory overhead on big datasets
- [ ] Write tests
- [ ] Write addition to API spec and discuss (API parameter? default?)
- [ ] Write changelog entry

### DIFF
--- a/include/engine/api/json_factory.hpp
+++ b/include/engine/api/json_factory.hpp
@@ -77,7 +77,8 @@ util::json::Object makeRouteStep(guidance::RouteStep step,
 
 util::json::Object makeRoute(const guidance::Route &route,
                              util::json::Array legs,
-                             boost::optional<util::json::Value> geometry);
+                             boost::optional<util::json::Value> geometry,
+                             util::json::Array nodes);
 
 util::json::Object
 makeWaypoint(const util::Coordinate location, std::string name, const Hint &hint);
--- a/include/engine/api/json_factory.hpp
+++ b/include/engine/api/json_factory.hpp
@@ -73,12 +73,12 @@ util::json::Object makeGeoJSONGeometry(ForwardIter begin, ForwardIter end)
 util::json::Object makeStepManeuver(const guidance::StepManeuver &maneuver);
 
 util::json::Object makeRouteStep(guidance::RouteStep step,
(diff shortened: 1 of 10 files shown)


TheMarex:
My biggest questions circle around the API changes:

1. Should there be an option to enabled/disable node ids?

2. Should node ids be per-step or per-route?
    - For `Route`: Otherwise you have to always pass `steps=true` to get node ids. This would mean paying the full guidance overhead, even if you don't really need it.
    - Against `Route`: The geometry in the `Route` object can be simplified, creating a mismatch between node ids and geometry.
   - For `RouteLeg`: This is the common use-case for map matching: Getting the ids between two trace points.
   - For `RouteStep`: Geometry is always unsimplfied, always matches node ids. 
   - Against `RouteStep`: I can't imagine an application where you would need step level granularity

3. Should we only allow array encoding? Or more space efficient base64 based encodings?

I'm leaning to:

1. Yes because it makes the transport size explode.
2. I'm torn. I would lean for `RouteLeg` motivated by the `match` use-case.
3. Transport size would be an argument for more space efficient encoding, but seeing the problems we are having with pushing different geometry formats to all layers, I would lean to `no`.

danpat:
For PR https://github.com/Project-OSRM/osrm-backend/pull/2390, we added an `annotation` object to the RouteLeg.  This would be a fine place to include this data.

lbud:
Closing here as I've recreated this against master at https://github.com/Project-OSRM/osrm-backend/pull/2418.

Issue #1403, opened 2015-02-28
Code style and naming conventions


TheMarex:
Where are we going with the class/function naming conventions? Currently the code is in a flux between camel-case and underscore.
Lets capture this in an issue to avoid any unnecessary style fixes.

Current proposal:
- Capitalized camel-case class names
- Lower-case underscore variable names
- Lower-case underscore function names
- Lower-case underscore namespace names

TheMarex:
Using something similar to the ROS C++ coding guidelines would be best. http://wiki.ros.org/CppStyleGuide

Seeing that almost all class names are CamelCase we should enforce this and make this consistent until the next release over the whole codebase.

daniel-j-h:
Tasks:

- [ ] decide on style
- [ ] implement in a cncc style file
- [ ] update code base globally using cncc

References:

- [cncc](https://github.com/mapbox/cncc)
- [default cncc style file](https://gist.github.com/daniel-j-h/da8722321366764acef1) with all available kinds
- [dump_ast.sh](https://github.com/mapbox/cncc/blob/b6f5c61c314a2a2bd6ccf1e5a19b3c1b3e835d0e/util/dump_ast.sh) for generating the Clang AST

TheMarex:
> decide on style

Given the current distribution of styles, I think we should favor CamelCase LLVM-like styling before std-like styling.

TheMarex:
@daniel-j-h Let's do this in one go after @danpat merged his traffic data branch, to not make merging more painful than necessary.

TheMarex:
Running this out post-release.

TheMarex:
Was part of this https://github.com/Project-OSRM/osrm-backend/pull/1848 but I think `cncc` missed a fair amount of misnamed symbols. Any idea @daniel-j-h?

daniel-j-h:
Hmm how did you invoke it? I can still find violations simply by doing

    find src -type f | xargs cncc --style .cncc.style --dbdir build

daniel-j-h:
We might also have dead code that never gets compiled and thus never ends up in the compile commands db. Never thought about this actually.

daniel-j-h:
Something went wrong here, look what I got from running it:

    $ wc -l cncc.txt
    461 cncc.txt

Here's the output / todo list:

https://gist.github.com/daniel-j-h/85acaf3f5d233c86dce7

TheMarex:
[...]

PR #2338, opened 2016-04-28
Fix naming conventions across the codebase

TheMarex:
Addresses #1403


daniel-j-h:
Haven't had a look at cncc for a while --- you had some problems the last time (reading the linked ticket) with it not looking at headers, but it seems to do so now?

PR #2340, opened 2016-04-28
Fix summary generation when empty strings are present.

### LINKED ISSUE #2339: Route summary can include empty strings
If ways with no names end up being the largest part of the route, then the summary may end as an odd-looking string like:

`, M-30`

In this case, the first summary item are a collection of ways with no names.

Example route against http://router.project-osrm.org/route/v1/car/-3.663499999999999,40.4679;-3.7053999999999974,40.4067?steps=true

We should exclude all empty road names from summary calculation.

danpat:
Also corrects step ordering when steps have equal durations.

Fixes #2339

### DIFF
--- a/features/testbot/summary.feature
+++ b/features/testbot/summary.feature
@@ -39,7 +39,7 @@ Feature: Basic Routing
 
         When I route I should get
             | from | to | route          | summary  |
-            | e    | a  | de,,bc,ab   | de, bc   |
+            | e    | a  | de,,bc,ab,ab   | de, bc   |
 
     @smallest @todo
     Scenario: Summaries when routing on a simple network
--- a/features/testbot/summary.feature
+++ b/features/testbot/summary.feature
@@ -58,19 +58,20 @@ Feature: Basic Routing
     @repeated
     Scenario: Check handling empty values
         Given the node map
-            | f |   |   |   |
-            | b | c | d | e |
-            | a |   |   |   |
+            | f |   |   | x |   |   |
+            | b | c | d | e | 1 | g |
+            | a |   |   | y |   |   |
 
         And the ways
-            | nodes | name   |
-            | ab    | first  |
-            | bc    | first  |
-            | cd    | first  |
-            | de    | second |
-            | bf    | third  |
+            | nodes | name   | # |
+            | ab    | first  |   |
+            | bc    | first  |   |
+            | cd    | first  |   |
+            | deg   | second |   |
+            | bf    | third  |   |
+            | xey   | cross  |we need this because phantom node segments are not considered for the summary |
 
         When I route I should get
             | from | to | route                     | summary       |
-            | a    | e  | first,first,second,second | first, second |
+            | a    | 1  | first,first,second,second | first, second |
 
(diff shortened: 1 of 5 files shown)


include/engine/guidance/assemble_leg.hpp
```diff
@@ -34,14 +34,22 @@ struct NamedSegment
...
 
 template <std::size_t SegmentNumber>
+
 std::array<std::uint32_t, SegmentNumber> summarizeRoute(const std::vector<PathData> &route_data)
 {
     // merges segments with same name id
     const auto collapse_segments = [](std::vector<NamedSegment> &segments)
     {
         auto out = segments.begin();
         auto end = segments.end();
```
MoKob: `const`

features/testbot/summary.feature
```diff
@@ -0,0 +1,54 @@
...
+        And the ways
+            | nodes | name |
+            | ab    | ab   |
+            | bc    | bc   |
+            | cd    | cd   |
+            | de    | de   |
+            | ef    |      |
+
+        When I route I should get
+            | f    | a  | ,de,cd,bc,ab,ab   | de, bc   |
```
MoKob: missing header?

MoKob:
I added another testcase and fixed one of yours. After the fixing (added header), cucumber fails. Please also review my testcase.

include/engine/guidance/assemble_leg.hpp
```diff
@@ -54,7 +62,8 @@ std::array<std::uint32_t, SegmentNumber> summarizeRoute(const std::vector<PathDa
                 *out = *in;
             }
         }
-        return out;
+        BOOST_ASSERT(out != end);
+        return ++out;
```
daniel-j-h: Is this `std::unique(first, last, sameNameid)` ?

MoKob: close. It accumulates duration and does something else. the something else seems like it should be unique, but it does not work. I don't know exactly what it is supposed to be.

features/testbot/summary.feature
```diff
@@ -0,0 +1,77 @@
...
+            | nodes | name |
+            | ab    | ab   |
+            | bc    | bc   |
+            | cd    |      |
+            | de    | de   |
+            | df    | df   |
+
+        When I route I should get
+            | from | to | route          | summary  |
+            | e    | a  | de,,bc,ab   | de, bc   |
```
TheMarex: This returns a route that is named:

> de,,ab,ab 

Note no `bc`. But the s ummary is named `de,bc`. It seems the post-processing changes the name of the `bc` segment. @MoKob can you take a look at this tomorrow?

MoKob: It's not a postprocessing issue but a generation issue. The part of the name announcement has an error for unnamed-roads. Fixing it and adding test-cases.

MoKob: Well, this is not really a bug, at least not in post processing. It is how to announce name-changes with no-name.

I added https://github.com/Project-OSRM/osrm-backend/pull/2367 to handle this situation a bit better.

In addition, the string needs to be `de,,bc,ab,ab` (currently ,,->bc is not announced, will change https://github.com/Project-OSRM/osrm-backend/pull/2367)

PR #2342, opened 2016-04-29
make AppVeyor publish binaries again

BergWerkGIS:


TheMarex:
Great! Seems to have published again.

PR #2344, opened 2016-04-29
Makes bicycle and foot profile's name and ref tag handling consistent with car profile

daniel-j-h:
In light of https://github.com/Project-OSRM/osrm-backend/pull/2328 I found the bicycle and foot profile to be inconsistent with how name and ref tags are assembled for a way's name.

Changing to

    Name (Ref)

for consistency.

/cc @systemed for visible changes in bicycle profile responses

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 
    - Profile Changes:
     - introduced a suffix_list / get_name_suffix_list to specify name suffices to be suppressed in name change announcements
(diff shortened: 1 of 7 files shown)


TheMarex:
This need an entry in `CHANGELOG.md`

daniel-j-h:
Oh yes, done in https://github.com/Project-OSRM/osrm-backend/pull/2344/commits/760e11a49c903f439075eb45c5c3b73cfeaea198

Issue #2295, opened 2016-04-22
Intersection Classifications


MoKob:
Currently, some turn instructions might be unclear on complex intersections.

If multiple possible turns exist that similar to each other. A possible example would be taking a right turn at this intersection.

<img width="149" alt="screen shot 2016-04-22 at 12 42 29" src="https://cloud.githubusercontent.com/assets/12932279/14739436/c2505d40-0887-11e6-971e-0ae5b85cf6de.png">

Even if we could distinguish them, saying `right` and `sharp right`, the same could be done with `right` and `slight right` as well as `sharp right` and `slight right`.

This ambiguity indicates a reason to augment turns with intersection classifications.

In a first step, this classification should contain a list of available turns at the location, as well as an indication of they are possible turns or not. In combination with the in/out bearings, we could enable a front-end to supply an improved visual representation of the turn itself and what roads are the ones that participate in it.

The resulting representation of the intersection could be used to generate Turn Symbols:
![representation](https://cloud.githubusercontent.com/assets/12932279/15011877/9485ad2c-11f4-11e6-9372-ab68d2128dcd.png)

The output for this turn could look like this:

```
"maneuver":
 {
    type: turn,
    modifier: right,
    bearing_before: 90,
    bearing_after:  130,
    location":[xx.xxxx,yy.yyyy]
    intersection:
   {
      out_bearing: 135,
      connections: [ 
       {
         entry_allowed: true,
         bearing: 45
       },
      {
         entry_allowed: true,
         bearing: 90
       },
       {
         entry_allowed: true,
         bearing: 135
       },
       {
          entry_allowed: true,
          bearing: 225
       },
       {
         entry_allowed: false,
         bearing: 270
       }
    ]
}
```
The discretisation results in differences between `intersection:out_bearing` and `bearing_after` would be a result of discretization.

We could also imagine giving out angles and normalising the output to the direction of travel (always have a 0-angle for the u-turn and all discretised turn_angles instead of the bearing values).
This would represent the view of the intersection a driver would actually have.

The Implementation requires the following tasks:

 - [x] classify all turns during preprocessing. This has to be done prior to the merging of segregated roads. Initial classification should consist of
   - [x] available roads and their bearings
   - [x] allowed entry flag

 - [x] add the classification ID to all intersections (if it turns out that keeping contracted nodes results in more nodes than ways, we should consider storing it with the edges).

 - [x] report intersection data in the API

[...]

PR #2345, opened 2016-04-29
[review] intersection classification

MoKob:
Work towards https://github.com/Project-OSRM/osrm-backend/issues/2295

Main goal is to create a method to indicate the form of an intersection at the moment of a turn instruction.


TheMarex:
Because of timing constraints with `5.2.0` I think we should make `bearing_before` and `bearing_after` just deprecated and remove with the next major bump. Removing these is not really urgent.

PR #2347, opened 2016-04-30
Optimize

TheMarex:
This needs a cleanup since there is also benchmarking code in there.

 85a51b7 brings a real performance boost and  006ce65 cuts longer routes down by ~15ms.

### DIFF
--- a/include/util/timing_util.hpp
+++ b/include/util/timing_util.hpp
@@ -1,54 +1,13 @@
 #ifndef TIMING_UTIL_HPP
 #define TIMING_UTIL_HPP
 
-#include <atomic>
 #include <chrono>
 #include <cstdint>
-#include <mutex>
-#include <unordered_map>
 
 namespace osrm
 {
 namespace util
 {
-
-struct GlobalTimer
-{
-    GlobalTimer() : time(0) {}
-    std::atomic<uint64_t> time;
-};
-
-class GlobalTimerFactory
-{
-  public:
-    static GlobalTimerFactory &get()
-    {
-        static GlobalTimerFactory instance;
-        return instance;
-    }
-
-    GlobalTimer &getGlobalTimer(const std::string &name)
-    {
-        std::lock_guard<std::mutex> lock(map_mutex);
-        return timer_map[name];
-    }
-
-  private:
-    std::mutex map_mutex;
-    std::unordered_map<std::string, GlobalTimer> timer_map;
-};
-
-#define GLOBAL_TIMER_AQUIRE(_X)                                                                    \
-    auto &_X##_global_timer = GlobalTimerFactory::get().getGlobalTimer(#_X)
-#define GLOBAL_TIMER_RESET(_X) _X##_global_timer.time = 0
-#define GLOBAL_TIMER_START(_X) TIMER_START(_X)
-#define GLOBAL_TIMER_STOP(_X)                                                                      \
-    TIMER_STOP(_X);                                                                                \
-    _X##_global_timer.time += TIMER_NSEC(_X)
-#define GLOBAL_TIMER_NSEC(_X) static_cast<double>(_X##_global_timer.time)
-#define GLOBAL_TIMER_USEC(_X) (_X##_global_timer.time / 1000.0)
-#define GLOBAL_TIMER_MSEC(_X) (_X##_global_timer.time / 1000.0 / 1000.0)
-#define GLOBAL_TIMER_SEC(_X) (_X##_global_timer.time / 1000.0 / 1000.0 / 1000.0)
 
 #define TIMER_START(_X) auto _X##_start = std::chrono::steady_clock::now(), _X##_stop = _X##_start
 #define TIMER_STOP(_X) _X##_stop = std::chrono::steady_clock::now()
--- a/src/benchmarks/CMakeLists.txt
+++ b/src/benchmarks/CMakeLists.txt
@@ -1,5 +1,5 @@
-file(GLOB RTreeBenchmarkSources
-    *.cpp)
+file(GLOB RTreeBenchmarkSources static_rtree.cpp)
+file(GLOB MatchBenchmarkSources match.cpp)
 
 add_executable(rtree-bench
 	EXCLUDE_FROM_ALL
@@ -15,6 +15,18 @@ target_link_libraries(rtree-bench
 	${CMAKE_THREAD_LIBS_INIT}
 	${TBB_LIBRARIES})
 
+add_executable(match-bench
+	EXCLUDE_FROM_ALL
+	${MatchBenchmarkSources}
+	$<TARGET_OBJECTS:UTIL>)
+
+target_link_libraries(match-bench
+	osrm
+	${Boost_LIBRARIES}
+	${CMAKE_THREAD_LIBS_INIT}
+	${TBB_LIBRARIES})
+
 add_custom_target(benchmarks
 	DEPENDS
-	rtree-bench)
+	rtree-bench
+	match-bench)
--- a/src/benchmarks/match.cpp
+++ b/src/benchmarks/match.cpp
@@ -1,0 +1,234 @@
+#include "util/timing_util.hpp"
+
+#include "osrm/match_parameters.hpp"
+
+#include "osrm/coordinate.hpp"
+#include "osrm/engine_config.hpp"
+#include "osrm/json_container.hpp"
+
+#include "osrm/status.hpp"
+#include "osrm/osrm.hpp"
+
+#include <boost/assert.hpp>
+
+#include <string>
+#include <utility>
+#include <iostream>
+#include <exception>
+
+#include <cstdlib>
+
+int main(int argc, const char *argv[]) try
+{
+    if (argc < 2)
+    {
+        std::cerr << "Usage: " << argv[0] << " data.osrm\n";
+        return EXIT_FAILURE;
+    }
+
+    using namespace osrm;
+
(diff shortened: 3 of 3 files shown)


TheMarex:
During this optimization spree I mainly focused on optimizing map matching performance. Key to fast map matching is the coordinate lookup, which used to occupy 85% of the run-time when doing map matching (at least for high-resolution traces).

I used a simple benchmark query on a small dataset, so threat this numbers as relative not real-world values. My test trace went from:

```
49.3284ms/req at 82 coordinate
0.601566ms/coordinate
```

To:
```
26.841ms/req at 82 coordinate
0.327329ms/coordinate
```

The StaticRTree is still the bottle neck here. ~20% are spend in the actual map matching. ~70% in the StaticRTree, the rest is 5% shortest path for the final route and lots of small change. (note this benchmark excluded step generation and geometry generation)

oxidase:
Hi, i have created PR #2348 to make leaf node memory mapped file more system-friendly. util-tests are failing but i can fix it soon.

CMakeLists.txt
```diff
@@ -99,6 +99,7 @@ if(CMAKE_BUILD_TYPE MATCHES Debug)
 endif()
 if(CMAKE_BUILD_TYPE MATCHES Release)
   message(STATUS "Configuring OSRM in release mode")
+  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -fno-omit-frame-pointer")
```
daniel-j-h: Please make sure to remove this before merging.

(I usually set  `CXXFLAGS` in the env for flags like needed by perf, then I don't have to modify the build system)

include/engine/routing_algorithms/routing_base.hpp
```diff
@@ -779,18 +779,50 @@ template <class DataFacadeT, class Derived> class BasicRoutingInterface
...
+            static_cast<double>(toFloating(target_phantom.location.lon)) * DEGREE_TO_RAD;
+        const double current_cos = std::cos(current_lat);
+
+        const double sin_dlon = std::sin((prev_lon - current_lon) / 2.0);
+        const double sin_dlat = std::sin((prev_lat - current_lat) / 2.0);
+
+        const double aharv = sin_dlat * sin_dlat + prev_cos * current_cos * sin_dlon * sin_dlon;
+        const double charv = 2. * std::atan2(std::sqrt(aharv), std::sqrt(1.0 - aharv));
+        distance += EARTH_RADIUS * charv;
+
```
daniel-j-h: You know what's sad? All of this cmath-function heavy code can't be properly optimized as most if not all of those functions potentially set a global `errno` variable. If it turns out that code like this is the bottleneck we might want to move away from legacy cmath functions (in those rare situations where it makes sense!).

(It's also the reason why the cmath functions can't be `constexpr` and thus are not usable at compile time)

include/util/coordinate.hpp
```diff
@@ -181,6 +199,9 @@ bool operator==(const Coordinate lhs, const Coordinate rhs);
 bool operator==(const FloatCoordinate lhs, const FloatCoordinate rhs);
 std::ostream &operator<<(std::ostream &out, const Coordinate coordinate);
 std::ostream &operator<<(std::ostream &out, const FloatCoordinate coordinate);
+
+inline Coordinate::Coordinate(const FloatCoordinate &other) : Coordinate(toFixed(other.lon), toFixed(other.lat)) {}
```
daniel-j-h: If you fwd declare `FloatCoordinate` you can put the function definition into the class where it belongs to and get the inlining for free.

include/util/coordinate_calculation.hpp
```diff
@@ -25,16 +25,49 @@ const constexpr long double EARTH_RADIUS = 6372797.560856;
...
 
 double greatCircleDistance(const Coordinate first_coordinate, const Coordinate second_coordinate);
 
-std::pair<double, FloatCoordinate>
-projectPointOnSegment(const FloatCoordinate &projected_xy_source,
-                      const FloatCoordinate &projected_xy_target,
-                      const FloatCoordinate &projected_xy_coordinate);
+inline std::pair<double, FloatCoordinate> projectPointOnSegment(const FloatCoordinate &source,
+                                                         const FloatCoordinate &target,
+                                                         const FloatCoordinate &coordinate)
```
daniel-j-h: Should we pass all coordinate types by value? They're wrappers around 2*64bit (`FloatCoordinate`) and 2*32bit (`Coordinate`). Looking at the coordinates why is `FloatLongitude` and `FloatLatitude` `double` anyway, do we really need the precision?

include/util/coordinate_calculation.hpp
```diff
@@ -25,16 +25,49 @@ const constexpr long double EARTH_RADIUS = 6372797.560856;
...
+    const double normed_ratio = unnormed_ratio / squared_length;
+    double clamped_ratio = normed_ratio;
+    if (clamped_ratio > 1.)
+    {
+        clamped_ratio = 1.;
+    }
+    else if (clamped_ratio < 0.)
+    {
+        clamped_ratio = 0.;
+    }
```
daniel-j-h: I searched the codebase for a `clamp` function and found one in `json_util.hpp` (which is never used by the way :P) --- worth it to provide a utility function for it?

include/util/static_rtree.hpp
```diff
@@ -297,15 +309,17 @@ class StaticRTree
             throw exception("mem index file is empty");
         }
 
-        leaves_stream.open(leaf_file, std::ios::binary);
-        leaves_stream.read((char *)&m_element_count, sizeof(uint64_t));
+        m_leaves_region.open(leaf_file);
```
daniel-j-h: Handle failure?

include/util/static_rtree.hpp
```diff
@@ -258,6 +265,11 @@ class StaticRTree
         BOOST_ASSERT_MSG(0 < size_of_tree, "tree empty");
         tree_node_file.write((char *)&size_of_tree, sizeof(std::uint32_t));
         tree_node_file.write((char *)&m_search_tree[0], sizeof(TreeNode) * size_of_tree);
+
+        m_leaves_region.open(leaf_node_filename);
```
daniel-j-h: Handle failure.

include/util/static_rtree.hpp
```diff
@@ -318,8 +332,10 @@ class StaticRTree
             throw exception("mem index file is empty");
         }
 
-        leaves_stream.open(leaf_file, std::ios::binary);
-        leaves_stream.read((char *)&m_element_count, sizeof(uint64_t));
+        m_leaves_region.open(leaf_file);
```
daniel-j-h: Handle failure? This is three times the same code ^

include/util/static_rtree.hpp
```diff
@@ -491,9 +518,8 @@ class StaticRTree
                 projected_input_coordinate_fixed, projected_nearest);
             // distance must be non-negative
             BOOST_ASSERT(0. <= squared_distance);
-            traversal_queue.push(
-                QueryCandidate{squared_distance, CandidateSegment{Coordinate{projected_nearest},
-                                                                  std::move(current_edge)}});
+            traversal_queue.push(QueryCandidate{squared_distance, SegmentIndex {static_cast<std::uint32_t>(candidate_cache.size())}});
```
daniel-j-h: Format All The Things

src/engine/polyline_compressor.cpp
```diff
@@ -13,9 +13,8 @@ namespace engine
 namespace /*detail*/ // anonymous to keep TU local
 {
 
-std::string encode(int number_to_encode)
+void encode(int number_to_encode, std::string& output)
```
daniel-j-h: Why the change here?

TheMarex: `perf` showed a lot of append operations here. The char-wise append can't be helped but appending the sub-strings just created a lot of unnecessary copies.

include/util/coordinate_calculation.hpp
TheMarex: Using `floats` yields numerical instabilities when doing conversion from/to fixed point representation. There was a bug report on this a while back, especially yielding to problems with high resolution map matched traces (all the coordinates are projected using this function).

We should benchmark if there is actually a difference between passing by value or const ref.

include/util/coordinate.hpp
TheMarex: It is forward declared. It just needs to be a complete type when using it here because we access the members. There is a cyclic dependency.

include/engine/routing_algorithms/routing_base.hpp
TheMarex: And whta math functions could we use instead? We do spend like 10% of our runtime in `sin` / `cos` / `atan`.

daniel-j-h: There are a couple of constexpr math projects out there, that's why I know about this problem. Never used one of those, probably needs some investigation. I checked Boost.Math and they seem to provide policies for setting errno. But from skimming the source it seems like they're also using cmath in the implementation.

include/util/coordinate_calculation.hpp
TheMarex: They do different things. The json clamp removes values that can't be handled by JSONs number encoding like NaN.

TheMarex:
The functional part of this was broken out in #2385. I would like to move my benchmark to `src/benchmarks` for future reference and then close this.

daniel-j-h:
=> src/CMakeLists.txt :) modularization comes step by step

PR #2348, opened 2016-05-01
Make LeafNode aligned to memory pages.

oxidase:
Hi, i would like to propose the following PR
* LeafNode is aligned to LEAF_PAGE_SIZE.
  Alignment brings 24 bytes memory overhead for 4096, but reduces
  cache misses rate. It will not be seen on small benchmarks, but for large memory consumptions it will be require less systems works to drop least recently used pages.
* Removed unused m_element_count from leaf nodes file.
  The size is computed as m_leaves_region.size() / LEAF_PAGE_SIZE.
* Added try/catch for mmap exceptions messages.
Exceptions will cover non-existing and empty files, but also permissions, memory errors etc

### DIFF
--- a/include/util/static_rtree.hpp
+++ b/include/util/static_rtree.hpp
@@ -201,6 +201,8 @@ class StaticRTree
             leaf_node_file.write((char *)&current_leaf, sizeof(current_leaf));
             processed_objects_count += current_leaf.object_count;
         }
+        leaf_node_file.flush();
+        leaf_node_file.close();
 
         std::uint32_t processing_level = 0;
         while (1 < tree_nodes_in_level.size())
--- a/include/util/static_rtree.hpp
+++ b/include/util/static_rtree.hpp
@@ -44,13 +44,17 @@ template <class EdgeDataT,
           class CoordinateListT = std::vector<Coordinate>,
           bool UseSharedMemory = false,
           std::uint32_t BRANCHING_FACTOR = 128,
-          std::uint32_t LEAF_NODE_SIZE = 128>
+          std::uint32_t LEAF_PAGE_SIZE = 4096>
 class StaticRTree
 {
   public:
     using Rectangle = RectangleInt2D;
     using EdgeData = EdgeDataT;
     using CoordinateList = CoordinateListT;
+
+    static_assert(LEAF_PAGE_SIZE >= sizeof(uint32_t) + sizeof(EdgeDataT), "LEAF_PAGE_SIZE is too small");
+    static_assert(((LEAF_PAGE_SIZE - 1) & LEAF_PAGE_SIZE) == 0, "LEAF_PAGE_SIZE is not a power of 2");
+    static constexpr std::uint32_t LEAF_NODE_SIZE = (LEAF_PAGE_SIZE - sizeof(uint32_t)) / sizeof(EdgeDataT);
 
     struct CandidateSegment
     {
@@ -72,7 +76,9 @@ class StaticRTree
         LeafNode() : object_count(0), objects() {}
         uint32_t object_count;
         std::array<EdgeDataT, LEAF_NODE_SIZE> objects;
+        unsigned char leaf_page_padding[LEAF_PAGE_SIZE - sizeof(object_count) - sizeof(objects)];
     };
+    static_assert(sizeof(LeafNode) == LEAF_PAGE_SIZE, "LeafNode size does not fit the page size");
 
   private:
     struct WrappedInputElement
@@ -110,8 +116,6 @@ class StaticRTree
     };
 
     typename ShM<TreeNode, UseSharedMemory>::vector m_search_tree;
-    uint64_t m_element_count;
-    const std::string m_leaf_node_filename;
     std::shared_ptr<CoordinateListT> m_coordinate_list;
 
     boost::iostreams::mapped_file_source m_leaves_region;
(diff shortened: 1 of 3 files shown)


include/util/static_rtree.hpp
```diff
@@ -73,7 +76,9 @@ class StaticRTree
         LeafNode() : object_count(0), objects() {}
         uint32_t object_count;
         std::array<EdgeDataT, LEAF_NODE_SIZE> objects;
+        unsigned char __alignment[LEAF_PAGE_SIZE - sizeof(uint32_t) - sizeof(EdgeDataT) * LEAF_NODE_SIZE];
```
daniel-j-h: Could you change the naming for `__alignment`. Symbols starting with `_ _` are reserved for the stdlib implementation and I feel like "alignment" is generic enough that there is the probability for a clash somewhere somehow.

include/util/static_rtree.hpp
```diff
@@ -299,43 +299,29 @@ class StaticRTree
...
+    void MapLeafNodesFile(const boost::filesystem::path &leaf_file)
+    {
+        // open leaf node file and return a pointer to the mapped leaves data
+        try {
+            m_leaves_region.open(leaf_file);
+            std::size_t num_leaves = m_leaves_region.size() / sizeof(LeafNode);
+            m_leaves.reset(reinterpret_cast<const LeafNode*>(m_leaves_region.data()), num_leaves);
+        } catch (std::exception& exc) {
+            throw exception(boost::str(boost::format("Leaf file %1% mapping failed: %2%") % leaf_file % exc.what()));
+        }
```
daniel-j-h: :+1: for error handling.

src/contractor/contractor.cpp
```diff
@@ -330,16 +330,16 @@ std::size_t Contractor::LoadEdgeExpandedGraph(
...
-            std::ifstream leaf_node_file(rtree_leaf_filename, std::ios::binary | std::ios::in);
+            std::ifstream leaf_node_file(rtree_leaf_filename, std::ios::binary | std::ios::in | std::ios::ate);
             if (!leaf_node_file)
             {
                 throw util::exception("Failed to open " + rtree_leaf_filename);
             }
-            uint64_t m_element_count;
-            leaf_node_file.read((char *)&m_element_count, sizeof(uint64_t));
+            std::size_t leaf_nodes_count = leaf_node_file.tellg() / sizeof(LeafNode);
+            leaf_node_file.seekg(0, std::ios::beg);
```
daniel-j-h: I see, instead of reading a size prefix we seek to the end (ate) and calculate the size depending on the node size.

I _think_ this pattern is used in a lot of places, but don't know why in case we have a single `|prefix|items|` block. /cc @TheMarex Note: this breaks the file format. That is, we need to document it in the changelog when doing a minor release.

include/util/static_rtree.hpp
```diff
@@ -43,16 +43,19 @@ namespace util
 template <class EdgeDataT,
           class CoordinateListT = std::vector<Coordinate>,
           bool UseSharedMemory = false,
-          std::uint32_t BRANCHING_FACTOR = 128,
-          std::uint32_t LEAF_NODE_SIZE = 128,
-          std::uint32_t MAX_NUM_CANDIDATES = LEAF_NODE_SIZE * 10>
+          std::uint32_t BRANCHING_FACTOR = 64,
```
TheMarex: Can you share why you changed the `BRANCHING_FACTOR` back to `64`? If we lower the node size it will double the  number of leaves which will create another level in the RTree doubling memory consumption again. Increase the `BRANCHING_FACTOR` keeps the hierarchy small and overhead smaller.

include/util/static_rtree.hpp
```diff
@@ -432,11 +418,6 @@ class StaticRTree
         std::priority_queue<QueryCandidate> traversal_queue;
         traversal_queue.push(QueryCandidate{0, TreeIndex{0}});
 
-        // we need to limit the size of this vector to MAX_NUM_CANDIDATES
```
TheMarex: Why did you remove this?

TheMarex:
:+1: Good idea, it makes a lot more sense to give the actual page size.

Would be great if you could clarify some of the changes I commented on.

include/util/static_rtree.hpp
oxidase: Hi, here it is a pure rebase issue, i will keep 128. For better performance, it can be run some benchmarks for the full planet R-tree in 2D-space  [BRANCHING_FACTOR, LEAF_NODE_SIZE] and pick the "optimal" value.

TheMarex: That would be awesome. :+1: Last benchmarks I did were only spot checks and that was before some bigger structural changes in the RTree so we might see different results.

include/util/static_rtree.hpp
oxidase: I have removed caching, it was the reason for failing tests with "tiny" r-trees. With mapped files only a pair of indices and fixed_projected_coordinate is needed in QueryCandidate without full copy of EdgeData.

include/util/static_rtree.hpp
oxidase: Ok, I will change to leaf_page_padding

include/util/static_rtree.hpp
TheMarex: That is a good point. Actually if we make this all const, we could get around the copy-logic to create `PhantomNode` from this. Also we could slim down the `Hint` by just storing these indices. I'll ticket it.

src/contractor/contractor.cpp
oxidase: Yes, it will break the file format,

TheMarex:
Great work. Merged it in my branch so I can rebase on `master`.

daniel-j-h:
I just stumbled upon this again and was wondering if we simply can use

    struct alignas(pageSize) LeafNode { .. }

instead of manually computing the padding. Probably less error prone and shows intent.

oxidase:
Great! alignas can be used http://coliru.stacked-crooked.com/a/cec372d2baf66735 to avoid explicit padding, but with 64 or 128 alignment.

Issue #2349, opened 2016-05-01
SegFault for some Requests


adrienjuguet:
Hello,
I used your work since a week.
I created a server but for some combinations of coordinates the server stops with a segmentation fault.
I try with 2 different profiles (car and foot) and this problem appears just with car profile.

Example of a bad query :
http://127.0.0.1:5000/trip/v1/driving/4.7886357,45.756145;4.8008547,45.74967 

Example of a good query :
http://127.0.0.1:5000/trip/v1/driving/4.7878857,45.756145;4.8007174,45.749783

adrienjuguet:
I noticed that this problem is with trip service and not with route.

MoKob:
Turns out this is due to https://github.com/Project-OSRM/osrm-backend/blob/master/src/engine/plugins/trip.cpp#L140. The route fails if no-turn is allowed at one of the vias in between.
The connected roads to one of the two locations seem to be tagged wrong. Still the implementation should be able to find the u-turn at this point: http://www.openstreetmap.org/#map=18/45.75589/4.78896 .
Investigating further

MoKob:
@adrienjuguet so technically this is a data issue. OSRM in general is designed to not take u-turns, unless we are at a cul de sac. The one way-point in question is located at a drive through (not considered routable) which is tagged incorrectly and at a not routable oneway street (not connected to the road network). Nevertheless, the debug assertion shouldn't trigger.

MoKob:
The problem here is how we discover location in which we allow u-turns.
Just as https://github.com/Project-OSRM/osrm-backend/blob/master/features/testbot/via.feature#L134-L149 shows, this is currently only the case if we have a oneway segment that is a dead end. 

Our current fix is more than hacked and only works for a single segment.

```
    Scenario: Via point at a dead end with oneway
        Given the node map
            | a | b | c |
            |   | d |   |
            |   | e |   |
            |   | f |   |

        And the ways
            | nodes | oneway |
            | abc   |  no    |
            | bd    |  no    |
            | de    |  yes   |
            | ef    |  yes   |

        When I route I should get
            | waypoints | route                |
            | a,d,c     | abc,bd,bd,bd,abc,abc |
            | c,d,a     | abc,bd,bd,bd,abc,abc |
```
Results in 
```
    When I route I should get                     # features/step_definitions/routing.js:4
      | waypoints | route                |
      | a,d,c     | abc,bd,bd,bd,abc,abc |
[...]

PR #2350, opened 2016-05-02
Fix/2349

MoKob:
Symptomatic fix for https://github.com/Project-OSRM/osrm-backend/issues/2349.

For a full and correct handling, we would require to know about small components at the creation of the edge-based-graph.

### DIFF
--- a/features/testbot/via.feature
+++ b/features/testbot/via.feature
@@ -148,6 +148,47 @@ Feature: Via points
             | a,d,c     | abc,bd,bd,bd,abc,abc |
             | c,d,a     | abc,bd,bd,bd,abc,abc |
 
+    # See issue #2349
+    Scenario: Via point at a dead end with oneway
+        Given the node map
+            | a | b | c |
+            |   | d |   |
+            |   | e |   |
+
+        And the ways
+            | nodes | oneway |
+            | abc   |  no    |
+            | bd    |  no    |
+            | ed    |  yes   |
+
+        When I route I should get
+            | waypoints | route                |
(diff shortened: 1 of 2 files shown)


features/testbot/via.feature
```diff
@@ -148,6 +148,24 @@ Feature: Via points
             | a,d,c     | abc,bd,bd,bd,abc,abc |
             | c,d,a     | abc,bd,bd,bd,abc,abc |
 
+    # See issue #2349
```
TheMarex: I feel I added something like this in https://github.com/Project-OSRM/osrm-backend/pull/1899/files did this get removed by accident again?

TheMarex:
In general I'm okay with this as a hot fix.

We should figure out how big the problem here actually is. My hunch tells me that roads that lead into nothingness will only consist of one compressed edge anyway.

features/testbot/via.feature
MoKob: No, its an addition. Your test is still in there, but is only for exiting oneways, not entering ones.

MoKob:
In general, if your assumption is true, the addition of this fix should handle the problem in general. It is only good as long as the assumption is true. Other cases will still fail.

TheMarex:
@MoKob Yes. Approach here should be:

1. We add a test case that still make this crash (force 2 compressed edges?) and tag with `@bug`
2. We open an issue about this specific case and point to the failing test.
3. We merge this.

Does this sound good to you @MoKob?

MoKob:
Added the testcase

Issue #2351, opened 2016-05-02
Trip Plugin does not handle continue_straight


MoKob:
Related to: https://github.com/Project-OSRM/osrm-backend/issues/2349

The assertion https://github.com/Project-OSRM/osrm-backend/blob/master/src/engine/plugins/trip.cpp#L140 fails on trips that require a `u-turn` at a via-node.

The reason for this is that https://github.com/Project-OSRM/osrm-backend/blob/master/src/engine/plugins/trip.cpp#L176-L177 does not consider the `continue_straight` parameter.

If a route is found without `continue_straight`, roads only connected to small components can end up resulting in no possible trip to exist.

To handle this, we have to either consider small components in the generation of the edge-based-graph or adjust the distance table calculation for continue_straight.

MoKob:
Handled by #2355

PR #2355, opened 2016-05-02
Trip plugin does not have a continue_straight parameter, use false

TheMarex:


TheMarex:
Hrm. tests are failing because of:

> arn] [server error] code: /home/patrick/Code/osrm-temp/include/engine/guidance/assemble_steps.hpp:130
in: std::vector<osrm::engine::guidance::RouteStep> osrm::engine::guidance::assembleSteps(const DataFacadeT&, const std::vector<osrm::engine::PathData>&, const osrm::engine::guidance::LegGeometry&, const osrm::engine::PhantomNode&, const osrm::engine::PhantomNode&, bool, bool) [with DataFacadeT = osrm::engine::datafacade::BaseDataFacade]: duration >= 0

Needs more digging tomorrow.

MoKob:
Just linking to ensure issue gets closed: https://github.com/Project-OSRM/osrm-backend/issues/2351

Issue #2359, opened 2016-05-03
Windows compilation Error C2899: typename cannot be used outside a template declaration


jperaltamalvar:
I'm compiling on Windows 10, and I'm getting this error:
C:\Dev\OSRM\osrm-backend\src\storage\storage.cpp(42): error C2899: typename cannot be used outside a template declaration [C:\Dev\OSRM\osrm-backend\build\STORAGE.vcxproj]

[build_errors.txt](https://github.com/Project-OSRM/osrm-backend/files/246608/build_errors.txt)

daniel-j-h:
Hotfix: remove `typename` from [storage.cpp:42](https://github.com/Project-OSRM/osrm-backend/blob/30af544919cb86f0a75bae7edbd9aefeb3fd7520/src/storage/storage.cpp#L42):

```diff
diff --git a/src/storage/storage.cpp b/src/storage/storage.cpp
index 67fd122..43967dc 100644
--- a/src/storage/storage.cpp
+++ b/src/storage/storage.cpp
@@ -39,7 +39,7 @@ namespace osrm
 namespace storage
 {
 
-using RTreeLeaf = typename engine::datafacade::BaseDataFacade::RTreeLeaf;
+using RTreeLeaf = engine::datafacade::BaseDataFacade::RTreeLeaf;
 using RTreeNode =
     util::StaticRTree<RTreeLeaf, util::ShM<util::Coordinate, true>::vector, true>::TreeNode;
 using QueryGraph = util::StaticGraph<contractor::QueryEdge::EdgeData>;
```

daniel-j-h:
Does this solve your compilation issues? For the record, which Visual Studio version are you using? Our Appveyor continuous integration did not complain on that one from what I can see.

daniel-j-h:
The changes in https://github.com/Project-OSRM/osrm-backend/issues/2360 should do the trick.

Your linking issue is coming from a missing xml dependency. Did you follow the instructions at https://github.com/Project-OSRM/osrm-backend/wiki/Windows-Compilation?

Looping in @BergWerkGIS here.

jperaltamalvar:
Thanks @daniel-j-h

I am using Visual Studio Community 2015 (14.0.23107.0 D14REL). Now compiling with changes in  https://github.com/Project-OSRM/osrm-backend/pull/2360 and https://github.com/Project-OSRM/osrm-backend/pull/2363. 

[...]

PR #2360, opened 2016-05-03
Remove typename prefix from non-dependent type alias

daniel-j-h:
I think VS is pedantic but actually right here.

Fixes https://github.com/Project-OSRM/osrm-backend/issues/2359.

References:
- http://eel.is/c++draft/namespace.udecl#19


TheMarex:
Go ahead an merge this if this compiles. :+1:

daniel-j-h:
Compiles locally, will wait on Travis and then merge.

There are some changes in there that I thought were dependent types, turns out they aren't
http://eel.is/c++draft/temp.dep

PR #2361, opened 2016-05-03
introduce exit_ramps into the set of instructions

MoKob:
Currently we emit only a single ramp type. Off-ramps are conceived differently than on-ramps.
While an on-ramp is essentially a normal turn with the additional information of a ramp,

e.g. 
screen shot 2016-05-03 at 11 12 30

For which we would expect something like `Go straight onto the ramp towards *destination*`,

or 
screen shot 2016-05-03 at 13 41 59

For which I would expect an instruction like `Turn left/right onto the ramp towards *destination*`

Off ramps are essentially only required for their side of the highway.

screen shot 2016-05-03 at 13 44 18

Here I would expect `Take exit NR towards *destination*` if the exit is on the normal side (probably hard to do right now) or `Take exit NR on the left towards *destination*`, in case of an unusual exit that is located on the left.
screen shot 2016-05-03 at 13 47 10

- [x] update readme

- [ ] integrate downstream ( @karenzshea , new instruction coming down )

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,8 @@
 # 5.1.0
    - API:
-     - added roundabout-turn instruction. The instruction indicates a small roundabout that is treated as an intersection
+     - added StepManeuver type `roundabout turn`. The type indicates a small roundabout that is treated as an intersection
         (turn right at the roundabout for first exit, go straight at the roundabout...)
+     - added StepManeuver type `on ramp` and `off ramp` to distinguish between ramps that enter and exit a highway.
      - reduced new name instructions for trivial changes
      - combined multiple turns into a single instruction at segregated roads`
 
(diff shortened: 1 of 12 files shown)


TheMarex:
@MoKob the documentation now lives in `docs/http.md`, please add there.

We agreed on the convention that new turn types should be treated as `turn` types by the application. Will instructions generated for the new turn type `exit_ramp` work with that scheme?

Amending the specification for an important type like `ramp` so late in the process is a real problem. It is not only that we add a new turn type, we change the semantics of an existing on e as well. Can't we just add a new conditional field like `"ramp_type": ["entry"|"exit"]`?

/cc @freenerd @karenzshea

MoKob:
@TheMarex we can adjust the json factory to treat the exit/ramp part and add an additional field, will update the PR

TheMarex:
Capturing from chat, changing to `on_ramp` and `off_ramp` and deprecating the old identifier.

This removed the changed semantics without a custom field.

TheMarex:
Actually had to do some homogenization changes that slipped through:

- We never use `-`or `_` in turn types but spaces
- Changed `roundabout-turn` to `roundabout turn`
- Changed `on_ramp` to `on ramp` same with `off_ramp` and `off ramp`

PR #2367, opened 2016-05-04
fix new-name on empty strings

MoKob:
Handling new-name instructions for empty segments offers undesired behaviour right now.

Coming from an empty segment, the new name is not announced.

I propose the following announcements:

From `empty name` to `name`: announce new-name = name. (Even though technically most likely a data error, the announcement makes sure we can get the most out of it in post-processing).

From `name` to `empty name`: don't announce anything (instruction would be: Road looses road-name which would be a data-error anyhow).


TheMarex:
Looks good. :+1: Merging.

PR #2368, opened 2016-05-04
Register a way's .nodes() function for use in the profile's way_function

daniel-j-h:
Ripped out from: https://github.com/Project-OSRM/osrm-backend/pull/2328

Can be used as in:

    function way_function(way, result)
      for node in way:get_nodes() do
        local id = node:id()
        io.write(">>> " .. id .. "\n")
      end

      -- ...
    end

Warning: we do not use libosmium's location cache, therefore .location()
on the nodes will crash at runtime. Once we switch to using libosmium's
cache this should work. Until then, you can use the node id now.

### DIFF
--- a/src/extractor/scripting_environment.cpp
+++ b/src/extractor/scripting_environment.cpp
@@ -14,6 +14,7 @@
 #include "util/typedefs.hpp"
 
 #include <luabind/tag_function.hpp>
+#include <luabind/iterator_policy.hpp>
 #include <luabind/operator.hpp>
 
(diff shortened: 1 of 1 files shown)


TheMarex:
This already slipped into master by accident. @_@

https://github.com/Project-OSRM/osrm-backend/commit/363d2145bdd28f389284855ca0b9618f62271669

daniel-j-h:
Wait what? I remember rebasing -i my branch, removing the commit and splitting it off to this pr. Hm.

PR #2376, opened 2016-05-05
Addition to fix #2356

### LINKED ISSUE #2356: Hint option with v5 api
I tried to get the hint option included in the v5 api request for trip and route. The request fails if the last location in the string is an empty element. In the way we use it, this is a common use case. I'm quite new to this, so maybe there is an easy solution (which I'm obviously missing) on how to pass an empty hint element for the last location. Is there a way to pass an empty hint element for the last location?

oxidase:
Hi, #2356 fixes only for the last empty hint, but it ignores empty hints in the middle
so for result_10->hints will contain only 2 hints, but expected number is 4
by https://github.com/Project-OSRM/osrm-backend/blob/master/include/engine/api/base_parameters.hpp#L72

I have added an explicit emplacement of boost::none if the optional string is none.

### DIFF
--- a/include/server/api/base_parameters_grammar.hpp
+++ b/include/server/api/base_parameters_grammar.hpp
@@ -75,11 +75,15 @@ struct BaseParametersGrammar : boost::spirit::qi::grammar<Iterator, Signature>
     BaseParametersGrammar(qi::rule<Iterator, Signature> &root_rule)
         : BaseParametersGrammar::base_type(root_rule)
     {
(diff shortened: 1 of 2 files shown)


TheMarex:
Good catch! Thanks for this. :+1: Should have tested more than the breaking case...

TheMarex:
Darn I forgot to include this in RC1. Can you rebase this on master so I can merge asap?

oxidase:
Ok, rebased

PR #2380, opened 2016-05-05
Approximation of latToY function

oxidase:
Hello, 
i have made an approximation of the latToY function by the Pade approximation. 
The range is [-70,70] degrees, the error is bounded by 4e-12, so it almost twice coordinate precision.
The approximation order [11/11] is the most possible for double precision.
The approximation now requires 22 multiplications and additions, and one division instead of log and sin.
I also changed in  yToLat to rounding to the coordinate precision instead of truncation, this will improve average value of quadratic quantization error by factor 2.

Here is the approximation error
pade11_deg
and maxima code
```
P:[0.00000000000000000000000000e+00, 1.00000000000089108431373566e+00, 2.34439410386997223035693483e-06, -3.21291701673364717170998957e-04, -6.62778508496089940141103135e-10, 3.68188055470304769936079078e-08, 6.31192702320492485752941578e-14, -1.77274453235716299127325443e-12, -2.24563810831776747318521450e-18, 3.13524754818073129982475171e-17, 2.09014225025314211415458228e-23, -9.82938075991732185095509716e-23];
Q:[1.00000000000000000000000000e+00, 2.34439410398970701719081061e-06, -3.72061271627251952928813333e-04, -7.81802389685429267252612620e-10, 5.18418724186576447072888605e-08, 9.37468561198098681003717477e-14, -3.30833288607921773936702558e-12, -4.78446279888774903983338274e-18, 9.32999229169156878168234191e-17, 9.17695141954265959600965170e-23, -8.72130728982012387640166055e-22, -3.23083224835967391884404730e-28];

horner3(p,x):=rreduce(lambda([a,y],x*y+a),p);
r(x):=horner3(P,x)/horner3(Q,x);
f(x):= 180./%pi*log(tan(%pi / 4. + (%pi*x/180.) / 2.));
plot2d([f(x)-r(x)], [x, -70, 70])$
```

The approximation lowers  in test_benchmark part of ExploreLeafNode from 48% to 30% and in time from 490us to 311us. In pure rtree benchmark with 10000 requests time per coordinate is lowered from 320us to 55us.

### DIFF
--- a/include/util/static_rtree.hpp
+++ b/include/util/static_rtree.hpp
@@ -74,9 +74,9 @@ class StaticRTree
     struct LeafNode
     {
         LeafNode() : object_count(0), objects() {}
-        uint32_t object_count;
+        std::uint32_t object_count;
         std::array<EdgeDataT, LEAF_NODE_SIZE> objects;
-        unsigned char leaf_page_padding[LEAF_PAGE_SIZE - sizeof(object_count) - sizeof(objects)];
+        unsigned char leaf_page_padding[LEAF_PAGE_SIZE - sizeof(std::uint32_t) - sizeof(std::array<EdgeDataT, LEAF_NODE_SIZE>)];
     };
     static_assert(sizeof(LeafNode) == LEAF_PAGE_SIZE, "LeafNode size does not fit the page size");
 
--- a/include/util/web_mercator.hpp
+++ b/include/util/web_mercator.hpp
@@ -35,7 +35,7 @@ inline FloatLatitude yToLat(const double y)
     const double normalized_lat =
         detail::RAD_TO_DEGREE * 2. * std::atan(std::exp(clamped_y * detail::DEGREE_TO_RAD));
 
-    return FloatLatitude(normalized_lat - 90.);
+    return FloatLatitude(std::round(normalized_lat * COORDINATE_PRECISION) / COORDINATE_PRECISION - 90.);
 }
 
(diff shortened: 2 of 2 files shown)


TheMarex:
Incredible work. This will also speed up other algorithms like Douglas-Peucker. Can't wait to do a full-stack benchmark with this.

emiltin:
excellent. but what happens outside of the -70 to 70 range?

danpat:
@emiltin well, things start to go a little haywire after about +/- 70.5:

screenshot from 2016-05-05 13-16-18

@oxidase @TheMarex while we don't really have much routing +/-70, maybe we should wrap a conditional around this conversion for values >70 && <-70.  The poor folks at McMurdo might get totally lost.

TheMarex:
It already falls back to the old code for everything outside the range.

oxidase:
@danpat yes, outside the range [-70,70] the minmax error bound is no more valid and extrapolation error will be very bad. 

The full range [-90, 90] is not possible to approximate by rational polynoms because of infinities at the endpoints, so i decided that [-70,70] is a good range for approximation and for [-90,-70] and [70,90] it falls back to the old code. Another design parameter is the order of nominator/denominator, i took 11/11, because for order 12/12 double precision is not enough and the error looks like
pade12_deg
that looks weird for |x|>57 degrees but it is just a truncation error.

danpat:
My bad for not reading properly :-P  @emiltin there's your answer :-)

PR #2381, opened 2016-05-06
Use memory disk 200Mi for testing

oxidase:
Hi, i just want to check if the testing server starves in disk IO, and using a pure memory disk could help to speed it up. 512Mi is from stxxl 64Mi per vector * 8 vectors in osrm-extract.
If it will not work i will remove PR


oxidase:
After some try-and-error attempts, looks like 200Mi for the stxxl memory disk is enough to reduce disk IO, pass default timeouts for all jobs but Xcode and not be killed by travis OOM.

danpat:
@oxidase I often run with `disk=empty,20M,memory` locally when executing the test suite.  Most of the test cases are miniscule and don't need much data.

danpat:
@oxidase also, note that these builds are disabled because they're low-priority, not because they're individually slow.  Travis lets us run 4 (I think) builds in parallel, so we reduce the build set to the minimum useful that would keep the total build duration low.

oxidase:
@danpat ok, sure PR can be removed. I had just an idea to check by PR that timeouts on the travis server are related to the memory consumption by stxxl and travis VM limits. For my local build setting size of a memory disk to 512M apparently solved timeouts for debug builds.

Issue #2383, opened 2016-05-06
Wrong include path for storage_config.hpp


giraldeau:
I tried to #include <osrm/engine_config.hpp> in external C++ prog, but the include path to storage_config.hpp was not found.

The problem can be fixed by using the relative path for storage_config.hpp:

```
diff --git a/include/engine/engine_config.hpp b/include/engine/engine_config.hpp
index eb62719..1696876 100644
--- a/include/engine/engine_config.hpp
+++ b/include/engine/engine_config.hpp
@@ -28,7 +28,7 @@ SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 #ifndef ENGINE_CONFIG_HPP
 #define ENGINE_CONFIG_HPP
 
-#include "storage/storage_config.hpp"
+#include "../storage/storage_config.hpp"
 
 #include <boost/filesystem/path.hpp>
```

giraldeau:
Converted to pull request.

Issue #2031, opened 2016-02-29
Proper libosrm API without aliasing header and external -> internal dependencies


daniel-j-h:
Currently we provide aliasing header for our public API. This is ugly as it is confusing to the user, requires us to implement [include path hack](https://github.com/Project-OSRM/osrm-backend/commit/97e47d64dfe81ee2e27bbe9c7dc05fcee761dcd6a) for the public API's headers to work at all (which by the way pollutes the users includes: when we're using osrm/X and there's a library called X we cause problems).

We do this since not only our internal headers depend on the external ones (which is fine), but in addition, our external headers also depend on our internal ones (which is ugly and nothing more than a hack). This requires us to e.g. chase down the base_parameters.hpp include headers and also install the hint.hpp header and its includes onto the user's system.

For a clean solution, the public libosrm headers must not know about the internal headers at all.
This requires us re-thinking and properly designing the API (e.g. how to break dependencies for the hints: context object, void*?) but would result in a clean and useful libosrm shared library.

TheMarex:
I think we should delay here to `5.1.0`. This is a pain to do right and doesn't win us anything immediately.

TheMarex:
@daniel-j-h if we would move `engine::` into `osrm::` and make `coordinate.hpp` header only, this should yield something that does not break the `include/<Library name>` scheme, which would allow use to remove the extra include path, right?

daniel-j-h:
Do you mean only renaming namespaces? We still have to ship e.g. `typedefs.hpp` because some API header requires it. Same for `phantom_node.hpp` and so on. Just check the output of

    find /usr/local/include/osrm/ -type f -name '*.hpp'

(or where ever you installed libosrm to)

TheMarex:
@daniel-j-h yes namespaces and like the directory structure.

- We can make `Hint` not depend on `PhantomNode` by saving individual fields so we can construct a phantom node from it. (e.g. we can compute the weight/offset at runtime) What we basically need to mirror is a `EdgeBasedNode`.
- `typedefs.hpp` needs to die. Each interface should specify its own ID type. (`static_graph.hpp`, `dynamic_graph.hpp`, ...)

PR #2384, opened 2016-05-06
Fix include for storage_config.hpp

giraldeau:
Fixes: https://github.com/Project-OSRM/osrm-backend/issues/2383

Signed-off-by: Francis Giraldeau <francis.giraldeau@gmail.com>


TheMarex:
Include paths are never relative. We give all paths relative to `include/`. You need to make sure that `/usr/local/include/osrm` is in your include path when you install it.

giraldeau:
OK, I see: 

```
$ pkg-config --cflags libosrm 
-I/usr/local/include -I/usr/local/include/osrm
```

I was thinking that /usr/local/include was sufficient. Thanks!

daniel-j-h:
For the record: this is due to https://github.com/Project-OSRM/osrm-backend/issues/2031.

PR #2385, opened 2016-05-07
Faster rtree

TheMarex:
Breaks out the functional changes from #2347 to upstream them. This PR makes the StaticRTree factor 3 faster and in the process fixes some long standing issues:

- StaticRTree now uses mmap, that mean no explicit file reading
- StaticRTree now is const
- StaticRTree does not need to be thread-local. Hence its updated with every other dataset on data update
- StaticRTree is x3 faster thanks to less unnecessary copies and approximate projection by @oxidase 

This PR breaks the data format for `osrm-extract` because the leaf size changes and is now page aligned.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@
 
    - Infrastructure
     - BREAKING: reordered internal instruction types. This breaks the **data format**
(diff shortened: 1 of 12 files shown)


emiltin:
great speed up!

PR #2386, opened 2016-05-08
Missing two commits in StaticRTree changes 

oxidase:
Hi, looks like two commits are missing in PR #2385 from #2380.
The PR is just a rebase onto master.


TheMarex:
Good catch. Weird. Maybe I forgot to pull before rebasing? Urg.

PR #2390, opened 2016-05-10
Optional route annotation data

danpat:
This PR adds a new element to the route part of `route` and `match` responses, in the form of:

```
  "annotation" : { "duration" : [ 12,13,7,2,5 ],
                    "distance" : [ 80,90,30,2,4 ] }
```

where each item in these arrays is the duration/distance of each segment in the route geometry.

This optional extra data adds a very small per-query cost (accumulating a vector in `assemble_geometry.hpp`).  It is not included in the response by default, the user must supply the `annotation=true` parameter to enable it.

There is no additional data required for storage, and the on-disk data format does not change.  This feature is purely additive.

Exposing this data makes it easy to plot how time accumulates on a route.  When compared against the same data for a real-world GPS trace, this helps visually identify areas where the routing engine disagrees with travel-times in the real world.

screenshot from 2016-05-10 00-16-09

In the future, we could extend this for any other meta-data we have available on a per-coordinate basis (traffic data source data comes to mind).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,10 @@
+# 5.2
+   Changes from 5.1
+
+   - API:
+     - new parameter `annotate` for `route` and `match` requests.  Returns additional data about each
+       coordinate along the selected/matched route line.
+
 # 5.1.0
    Changes with regard to 5.0.0
 
--- a/docs/http.md
+++ b/docs/http.md
@@ -145,6 +145,7 @@ In addition to the [general options](#general-options) the following options are
 |------------|------------------------------------------|-------------------------------------------------------------------------------|
 |alternatives|`true`, `false` (default)                 |Search for alternative routes and return as well.\*                            |
 |steps       |`true`, `false` (default)                 |Return route steps for each route leg                                          |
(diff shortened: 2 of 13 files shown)


MoKob:
@danpat I feel like this is essentially a duplicate to what is happening by introducing intersection classes for both the turns and intermediate intersections.
Is there any benefit in having this as a separate array rather having it as part of the steps output?

It would be easy (I believe it is already maintained within the steps) to output a time and duration until an intersection is reached, next to the output highlighted in: https://github.com/Project-OSRM/osrm-backend/issues/2295#issuecomment-218444416

I feel that this addition is separated from what it is representing, not being tied to the route-steps. After all, the different segments should correspond to the different steps?

danpat:
@MoKob and I discussed this in another forum - this feature is per-geometry point, but the data he's going to be exposing is per intersection.  These offer differing granularity.

TheMarex:
Capturing from chat:

- [ ] Rename `progression={true|false}` -> `annotate={true|false}`
- [ ] Rename `RouteLeg.progression` to `RouteLeg.annotation`
- [ ] Add note to `OSM ID` PR use this scheme as well
- [ ] Add documentation
- [ ] Add changelog entry

TheMarex:
Great waiting for travis to give :+1:

Issue #2393, opened 2016-05-11
osrm build fails on ubuntu16.04


stvno:
I'm trying to build osrm 5.1.0 on a 64bit ubuntu 16.04 machine. I have installed all the dependencies for 15.10 as per https://github.com/Project-OSRM/osrm-backend/wiki/Building-on-Ubuntu (since the 16.04 list is not yet available).
However it fails while building osrm-components
```
[ 95%] Linking CXX executable osrm-components
/tmp/ccSgjAcD.ltrans0.ltrans.o: In function `main':
<artificial>:(.text.startup.main+0x1aef): undefined reference to `OGRLineString::addPoint(double, double)'
<artificial>:(.text.startup.main+0x1b27): undefined reference to `OGRLineString::addPoint(double, double)'
collect2: error: ld returned 1 exit status
CMakeFiles/osrm-components.dir/build.make:123: recipe for target 'osrm-components' failed
make[2]: *** [osrm-components] Error 1
CMakeFiles/Makefile2:555: recipe for target 'CMakeFiles/osrm-components.dir/all' failed
make[1]: *** [CMakeFiles/osrm-components.dir/all] Error 2
Makefile:127: recipe for target 'all' failed
make: *** [all] Error 2
```
I have tried both from the latest git and the official release. I'm running GDAL 2.1 which might be an issue?
``` 
ogrinfo --version
GDAL 2.1.0, released 2016/04/25
```

danpat:
@stvno Yeah, it won't compile against GDAL 2, see https://github.com/Project-OSRM/osrm-backend/issues/1738

The only thing that links against GDAL is the `osrm-components` tool, and I think it's unlikely that you actually need that.  Try running `cmake` with:

`cmake -DBUILD_COMPONENTS=OFF ..`

this will disable the `osrm-components` tool build and skip any GDAL-related problems.

FWIW, I'm using GDAL 1.11.3 on Ubuntu 16.04 and everything builds cleanly using GCC 5.3.

stvno:
It indeed builds without osrm-components so it is a duplicate of #1738

Issue #1738, opened 2015-10-17
Make osrm-backend work with GDAL 2


daniel-j-h:
@joto just mentioned to me that Debian wants to update their GDAL package to version 2. It seems like this breaks osrm-backend.

From a quick glance at the code, we only need GDAL for the tools (only for the components actually?).

Investigate.

TheMarex:
@daniel-j-h Yes only used for the `osrm-components` tool. Actually I would like to remove this dependency. Can we maybe switch to GeoJSON or something trivial like that? We don't really need shapefiles.

BergWerkGIS:
> Can we maybe switch to GeoJSON or something trivial like that?

My 0.02$ to just let you know what's generally moving on the shapefile vs. GeoJSON front.
Still investigating if GeoJSON is a feasible format, as it can grow really big really fast and is a nightmare to parse.
To be processed it has to be read into memory as a whole and sequentially parsed, although @artemp is working on making this problem go away with custom spatial indices.

@freenerd provided a **1GB shapefile**, that I think came out of the `osrm-components` tool.
Converting this results in a **1.7GB** to **2.3GB GeoJSON** (depending on formatting: line breaks, indentation, decimal places).
A 2GB GeoJSON is almost always too big to be opened by conventional means (GIS) except there is a lot of memory to burn and even if it can be opened, display might be exceptionally slow.

daniel-j-h:
@BergWerkGIS what about osrm-backend outputting geojson, which we then convert to shapefile before giving it out? I understand the problems in large geojson files, but as an intermediary format to convert from, it should suffice I guess. Would this be feasible?

daniel-j-h:
More updates in this regard:

The old demo server uses the components tool to produce a shapefile and pushes that to a Geofabrik server. We then query this Geofabrik server e.g. here: http://project-osrm.org/osrm-frontend-v2/ (components layer, bottom left).

That is, the components layer is still the outdated one from months ago, because the demo server does not do any updates any more!

/cc @freenerd @danpat re. the new demo server: is it possible that we generate the components (shapefile, geojson, ...) ourselves? I talked to Fred from Geofabrik, and he is open to integrate this then e.g. at their inspector that that a lot of osm people are using at http://tools.geofabrik.de/osmi/.

daniel-j-h:
Even more updates:

[...]

PR #2394, opened 2016-05-11
Disable building components tool by default until GDAL 2 compatibility lands

daniel-j-h:
Until GDAL 2 compat. #1738 lands this is only causing compile errors for users, e.g. see #2393.

If you need the components tool (like the demo server) you can pass the appropriate CMake flag.


TheMarex:
Looks good to me. Merge at will. :+1:

TheMarex:
Rebased on master. Waiting on travis :+1: to merge

PR #2395, opened 2016-05-11
emit turn-straight for obvious turns where the main road continues

MoKob:
In deciding which turn is obvious, the current turn classification does not consider names of other roads.
In situations like this

where the main road takes a turn, this can lead to going straight being reported as a new-name. Since the main road is taking a turn, we should not announce it as a new name but rather a turn.

In the concept of OSRM turn instructions, `turn` would be the instruction to emit.

We should consider handling the case `turn straight` specifically in a frontend/client to avoid the wording `turn straight onto name`. /cc @karenzshea @bsudekum

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,5 @@
-# 5.2
-   Changes from 5.1
+# 5.2.0
+   Changes from 5.1.0
 
    - API:
      - new parameter `annotate` for `route` and `match` requests.  Returns additional data about each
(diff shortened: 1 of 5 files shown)


CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+# 5.1.1
```
TheMarex: This should be `5.2.0`.

src/extractor/guidance/turn_handler.cpp
```diff
@@ -3,8 +3,8 @@
 #include "extractor/guidance/toolkit.hpp"
 #include "extractor/guidance/turn_handler.hpp"
 
-#include "util/simple_logger.hpp"
 #include "util/guidance/toolkit.hpp"
+#include "util/simple_logger.hpp"
```
TheMarex: Unused include.

PR #2396, opened 2016-05-11
Switch to codecov

TheMarex:


daniel-j-h:
Reason for this? What was wrong with coveralls? Just trying to understand this changeset.

TheMarex:
Well coveralls doesn't work. They don't seem to support recent gcc versions.

PR #2397, opened 2016-05-11
Support SO_REUSEPORT option

zimathias:
so that multiple osrm-routed processes can bind to the same port.


include/server/server.hpp
```diff
@@ -49,7 +49,12 @@ class Server
         boost::asio::ip::tcp::endpoint endpoint = *resolver.resolve(query);
 
         acceptor.open(endpoint.protocol());
+#ifdef SO_REUSEPORT
+        int one = 1;
```
daniel-j-h: Can be const?

include/server/server.hpp
```diff
@@ -49,7 +49,12 @@ class Server
...
+        setsockopt(acceptor.native_handle(), SOL_SOCKET, SO_REUSEADDR | SO_REUSEPORT, &one, sizeof(one));
```
daniel-j-h: This relies on the particular Boost.ASIO header we're including to transitively include sys/types.h, sys/socket.h.

zimathias:
@daniel-j-h let me know if this PR addresses your comments now.

daniel-j-h:
The unconditional includes currently break Appveyor, I would just ifdef them.

include/server/server.hpp
```diff
@@ -13,6 +13,9 @@
 
 #include <zlib.h>
 
+#include <sys/types.h>
```
TheMarex: needs `#ifndef _WIN32` around this since `sys/` is platform specific.

TheMarex:
I'm curious why do you use multiple processes over multiple threads?

Can you rebase this on `master`? There seem to be a few stray merge conflicts in your branch.

zimathias:
@TheMarex this allows to hot reload the osrm-routed binary without dropping requests. Given SO_REUSEPORT allows multiple processes to bind to the same port the binary hot reload process becomes:
1. start new process with new binary
2. wait for the new process to startup and serve requests
3. stop the old process with the old binary

TheMarex:
Closing in favor of https://github.com/Project-OSRM/osrm-backend/pull/2417

Issue #1353, opened 2015-01-21
Return distance for table queries


mantesat:
Would it be possible to add distances along with travel times to the matrix requests?

TheMarex:
Adding the distance should be doable in doable in the near future. The current M2M implementation needs some work however for this to be supported.
Note: This is not routing bases on shortest distance! Until we can support that it is still some time off.

jornedeblaere:
+1 for this feature request! This would be a huge breakthrough for me

mantesat:
Any news/timeframe/estimation for this?

mattkrick:
+1, I think there's a lot of value here for routing based on shortest time and displaying the distance.

DennisOSRM:
While we are asking for your patience at this point, please be assured that this is very much on our radar.

MannusEtten:
cool! i am looking also for the distance in this matrix so I can complete a OD-matrix :+1:

rCarto:
+1

jellevictoor:
What is the current status of this issue?

zavan:
That would be very welcome :smile:

zavan:
Guys, I wrote a NodeJS script based on OSRM NodeJS bindings that can serve both time and distance tables, if anywone is interested: https://gist.github.com/zavan/2aea4881922f9de2f371

zavan:
#1033 #1248 #544

faheemmughal:
+1

RhinoDevel:
You can find a [v4.9.0](https://github.com/Project-OSRM/osrm-backend/releases/tag/v4.9.0)-based branch including additional **distances in meters** [here](https://github.com/RhinoDevel/osrm-backend/tree/v4.9.0-meters).

Please take a look at the [changes made](https://github.com/Project-OSRM/osrm-backend/compare/v4.9.0...RhinoDevel:v4.9.0-meters).

The [table service](https://github.com/Project-OSRM/osrm-backend/wiki/Server-api) response's `.distance_table` is changed this way (sample):

From [v4.9.0](https://github.com/Project-OSRM/osrm-backend/releases/tag/v4.9.0) (only 1/10 seconds distance values): `[ [ 0, 24302 ], [ 25352, 0 ] ]`
[...]

PR #2399, opened 2016-05-12
Allow abitrary edge weights

TheMarex:
This PR aims to finally implement #77.

Where we are now:

- `ExtractionWay` has a new property `weight_per_meter` that we pull through the whole pipeline to the server. It is exposed over the fields named `weight` (analogous to `duration`).
- `ProfileProperties` has a new property `weight_name`

## TODO

- [x] This needs a rebase, probably conflicts with the Datafacade changes.
- [x] `weight_per_meter` is a **wrong** name. This was an oversight from my part: It actually encodes `weight_per_second` (analogous to `speed` being `meters_per_second`) We need to rename this. /cc @karenzshea
- [x] We need to integrate #3487 to check the profile version:
   - The interface between `ScriptingEnvironment` and `EdgeBasedGraphFactory` [here](https://github.com/Project-OSRM/osrm-backend/pull/2399/files#diff-341ad719a96c9c990d1255635d7f3dbdR541) needs to work with both API version
   - The interface between `ScriptingEnvironment` and `ExtractionContainers` [here](https://github.com/Project-OSRM/osrm-backend/pull/2399/files#diff-c1989cffadbf020c61f56e4026547b3eR434) also needs to work with both API version.
   - Profile API version `0`: We expect the current implementation of `turn_function` and `segment_function` and silently fallback to `weight = duration` (this needs wrapper code in `ScriptingEnvironment::ProcessTurn` and `ScriptingEnvironment::ProcessSegment`)
  - Profile API version `1`: We expect `turn_function` and `segment_function` to follow the scheme already implemented in this PR


danpat:
Capturing from chat:

>  what we really want from this is to not need to worry any more about what metric we put in and still get correct durations.

  - Let's worry about memory consumption later - let's see how useful this feature turns out first before we optimize.

  - The duration data needs to support turn penalties.  If we do time-as-weight routing, we need to support them there too.  We can store them at the end of the `GetUncompressedWeights` array, it'll add 4 bytes for each compressed geometry.

  - Let's only support a `weight_per_distance` value from the user, rather than arbitrary total values.  Everything gets scaled by the length of the road.

  - The `way_function` can return two values per way - `speed` and `weight_per_distance` (note that both these fields are `per-distance` measurements).

  - The `turn_function` can return two values - `time_penalty` and `weight_penalty`

For traffic data imports, let's look at some use cases:

  - if we're doing fastest-time routing, we will need to update both the routing weight *and* the stored durations
  - if we're doing shortest-distance routing, but adjusting durations for traffic, we need to update the stored durations, but not the weight
  - if we're trying to support "minimize toll costs" routing, but without traffic data, we would need to update the weights, but not the durations

The last case is a bit of a stretch, I couldn't think of a better example.  But the point is that we probably need to be able to update both with the CSV import.  I think we could do:

  - update the CSV file to have 4 columns, `nodeA,nodeB,speed,weight_per_distance`
  - supply a second CSV file with `nodeA,nodeB,weight_per_distance`

I think if we simply don't supply data, we can efficiently skip either the duration and/or weight updates.

emiltin:
wow, this has been a long time coming! it will be very useful for bike routing

TheMarex:
> nodeA,nodeB,speed,weight_per_distance

We can update the data with regard to the names given in the csv header column.

include/contractor/graph_contractor.hpp
```diff
@@ -154,7 +154,7 @@ class GraphContractor
         for (auto diter = input_edge_list.dbegin(); diter != dend; ++diter)
         {
             BOOST_ASSERT_MSG(static_cast<unsigned int>(std::max(diter->weight, 1)) > 0,
```
MoKob: Isn't `max(weight,1)` always `>0`, especially since we cast to an unsigned?

include/engine/api/json_factory.hpp
```diff
@@ -72,23 +72,17 @@ util::json::Object makeGeoJSONGeometry(ForwardIter begin, ForwardIter end)
     return geojson;
 }
 
-util::json::Object makeStepManeuver(const guidance::StepManeuver &maneuver);
-
-util::json::Object makeRouteStep(guidance::RouteStep step,
-                                 boost::optional<util::json::Value> geometry);
-
 util::json::Object makeRoute(const guidance::Route &route,
```
MoKob: this looks like an unrelated change?

include/engine/api/json_factory.hpp
```diff
@@ -72,23 +72,17 @@ util::json::Object makeGeoJSONGeometry(ForwardIter begin, ForwardIter end)
...
                              util::json::Array legs,
                              boost::optional<util::json::Value> geometry);
 
 util::json::Object
 makeWaypoint(const util::Coordinate location, std::string name, const Hint &hint);
 
-util::json::Object makeRouteLeg(guidance::RouteLeg leg, util::json::Array steps);
-
```
MoKob: also: unrelated?

include/engine/guidance/assemble_steps.hpp
```diff
@@ -43,15 +43,19 @@ inline std::vector<RouteStep> assembleSteps(const datafacade::BaseDataFacade &fa
                                             const bool source_traversed_in_reverse,
                                             const bool target_traversed_in_reverse)
 {
-    const double constexpr ZERO_DURATION = 0., ZERO_DISTANCE = 0.;
+    const double constexpr ZERO_DURATION = 0., ZERO_DISTANCE = 0., ZERO_WEIGHT = 0.0;
```
MoKob: nitpick: keep consistent `0.` with `duration` and `distance`

include/engine/guidance/assemble_steps.hpp
```diff
@@ -109,6 +116,7 @@ inline std::vector<RouteStep> assembleSteps(const datafacade::BaseDataFacade &fa
                                           NO_ROTARY_NAME,
                                           segment_duration / 10.0,
                                           distance,
```
MoKob: is there a reason why we store distance in meters and duration/weight in 10ths of (seconds/number)? Shouldn't we keep this consistent?

include/extractor/extraction_way.hpp
```diff
@@ -44,9 +47,16 @@ struct ExtractionWay
     void set_backward_mode(const TravelMode m) { backward_travel_mode = m; }
     TravelMode get_backward_mode() const { return backward_travel_mode; }
 
+    // speed in km/h
     double forward_speed;
     double backward_speed;
+    // weight per meter
+    double forward_weight_per_meter;
+    double backward_weight_per_meter;
```
MoKob: are these per meter, or per decimeter as indicated by the division by ten in assembly?

include/engine/guidance/assemble_steps.hpp
danpat: @MoKob the reasons are lost in the archives of history, but I believe the thinking was "what's the smallest unit we would want to keep" and "let's store it as an integer".  Combine these ideas, and you get "meters" and "deciseconds" 😛

include/extractor/profile_properties.hpp
```diff
@@ -10,10 +11,17 @@ namespace extractor
 
 struct ProfileProperties
 {
+    static const constexpr std::size_t MAX_WEIGHT_NAME_SIZE = 256;
```
MoKob: This is a missdirection as the maximum size is 255, actually. Should be 255 and array should have size `MAX + 1`

profiles/bicycle.lua
```diff
@@ -274,10 +275,6 @@ function way_function (way, result)
     -- regular ways
     result.forward_speed = bicycle_speeds[highway]
     result.backward_speed = bicycle_speeds[highway]
-    if safety_penalty < 1 and unsafe_highway_list[highway] then
```
MoKob: unrelated?

MoKob: never mind.

include/extractor/extractor_config.hpp
```diff
@@ -71,7 +71,9 @@ struct ExtractorConfig
         rtree_nodes_output_path = basepath + ".osrm.ramIndex";
         rtree_leafs_output_path = basepath + ".osrm.fileIndex";
         edge_segment_lookup_path = basepath + ".osrm.edge_segment_lookup";
-        edge_penalty_path = basepath + ".osrm.edge_penalties";
+        turn_duration_penalties_path = basepath + ".osrm.turn_weight_penalties";
+        turn_weight_penalties_path = basepath + ".osrm.turn_duration_penalties";
```
stwa: Looks like the values are swapped:
turn_weight_penalties → turn_duration_penalties and vice versa

TheMarex: good catch :+1:

TheMarex:
Things to do here:
- [ ] Split interleaved `CompressedGeometryContainer` serialization into separate vectors
- [ ] Only load the duration vector if weight doesn't contain the same values (safes memory consumption)
- [ ] Make duration values in `osrm-contract` updatable

src/extractor/edge_based_graph_factory.cpp
```diff
@@ -416,56 +425,79 @@ void EdgeBasedGraphFactory::GenerateEdgeExpandedEdges(
...
+                std::uint16_t weight_penalty = 0;
+                std::uint16_t duration_penalty = 0;
+                if (m_traffic_lights.find(node_v) != m_traffic_lights.end())
+                {
+                    // FIXME we can't assume this here!
+                    weight_penalty += profile_properties.traffic_signal_penalty;
+                    duration_penalty += profile_properties.traffic_signal_penalty;
+                }
+
+                if (use_turn_function)
```
oxidase: use_turn_function is not defined, may the changeset it is incomplete?

src/extractor/edge_based_graph_factory.cpp
```diff
@@ -416,56 +425,79 @@ void EdgeBasedGraphFactory::GenerateEdgeExpandedEdges(
...
+                {
+                    const auto extracted_turn = ProcessTurn(turn, lua_state);
```
oxidase: `ProcessTurn(ExtractionTurn &turn) ` must be adjusted here, also `lua_state` is not defined

oxidase:
Hi, i have almost finished rebase of the PR to master, but some parts are missing. If it is interesting, i can create a duplicate PR with rebased patches.

TheMarex:
@oxidase Let me just give your contributor access so you can push directly to this branch.

oxidase:
Please review a forced rebase of the branch to master.  I have modified some commits,  changed some properties in profiles and changed TurnPenalty  to int16_t to make all tests green.

src/extractor/edge_based_graph_factory.cpp
```diff
@@ -417,56 +426,72 @@ void EdgeBasedGraphFactory::GenerateEdgeExpandedEdges(
...
+                TurnPenalty weight_penalty = 0;
+                TurnPenalty duration_penalty = 0;
+                if (m_traffic_lights.find(node_v) != m_traffic_lights.end())
+                {
+                    // FIXME we can't assume this here!
+                    weight_penalty += profile_properties.traffic_signal_penalty;
+                    duration_penalty += profile_properties.traffic_signal_penalty;
+                }
+
+                // MKR if (use_turn_function)
```
TheMarex: Probably a left-over. ProcessTurn checks internally if this function exists. I think the if-encapsulation can be removed.

TheMarex:
@oxidase rebase looks good to me. You are right turn penalties need to be signed.

Biggest remaining issue is that we still use quite a bit more memory even if weight == duration. For this we need to modify the way we safe `CompressedEdgeContainer::CompressedEdge` on disk.
To enable conditional loading of the duration data for each segment, we need to split the interleaved storage of `(node, duration, weight)`  to three independent vectors.
Since this touches a lot of code around the `CompressedEdgeContainer` I think we should focus first on getting https://github.com/Project-OSRM/osrm-backend/pull/2717 in `master`, rebasing again and then attempting the split.

rkcpi:
Commit 4f59078ae0bbe2ffb58c424616c18550c0760fda breaks duration calculation with this line:

```
using TurnPenalty = std::int16_t; // turn penalty in 100ms units
```

With this in action, durations are calculated as negative numbers. Changing it to

```
using TurnPenalty = std::uint16_t;
```
fixes the calculation.

TheMarex:
@rkcpi can you elaborate on what specifically is broken?

stwa:
@TheMarex A "manual" test fails for us, when using osrm-frontend, all durations are negative numbers.
We had no time to debug this any further yet, but we suspect that there is still a type mismatch somewhere in the code. On a quick look I spotted differences for turn penalties between internal and shared datafacade (signed/unsigned), just for example.

include/engine/datafacade/datafacade_base.hpp
```diff
@@ -72,6 +72,10 @@ class BaseDataFacade
 
     virtual unsigned GetGeometryIndexForEdgeID(const unsigned id) const = 0;
 
+    virtual TurnPenalty GetWeightPenaltyForEdgeID(const unsigned id) const = 0;
```
MoKob: Is the type off here? Edge-Weights/Edge-Durations should be getting much larger than TurnPenalties. Limiting them here could be a possible overflow issue.

include/engine/datafacade/internal_datafacade.hpp
```diff
@@ -137,9 +139,24 @@ class InternalDataFacade final : public BaseDataFacade
...
+        if (!turn_penalties_stream)
+        {
+            throw util::exception("Could not open " + turn_penalties_path.string() +
+                                  " for reading.");
+        }
+        extractor::io::TurnPenaltiesHeader header;
+        turn_penalties_stream.read(reinterpret_cast<char *>(&header), sizeof(header));
+        turn_penalties.resize(header.number_of_penalties);
+        turn_penalties_stream.read(reinterpret_cast<char *>(turn_penalties.data()),
+                                   sizeof(std::uint16_t) * header.number_of_penalties);
```
MoKob: should be TurnPenalty (?), to keep type consistent

pongo710:
Can someone with push priviliges rebase to current master so we have the fix of #2852 included?

stwa:
@oxidase Are you still working on this? Can you rebase on current master?

oxidase:
@stwa I have rebased the branch to master and  squashed some commits to have "green" tests during rebase. I also unified TurnPenalty types to be int16_t that is required by some tests with negative turn duration updates. For @MoKob comment, i did not extend for now int16 type but placed assertions for penalties to be in range of int16. 
At the moment i am not working on the issue, it requires #2717
Please could you check if there is still a problem with your tests.

stwa:
@oxidase Thanks for your help! We will repeat our tests with the new fixes applied.

pongo710:
Thanks for rebasing.

It seems there is a bug in the current status of this pull request.

I set a weight of 10000 to a way. When routing into the way i get a duration time of 1h 7 min. When i route the same distance but other way round i get a duration of only 30 sec.

The images below are showing this behaviour. The weight is set to the way where the green marker is set in the first image.

bildschirmfoto 2016-09-22 um 17 40 19
bildschirmfoto 2016-09-22 um 17 40 38


So if routing from a way with weight it works, but routing to a way with weight, it does not work.
I think some of the previous changes have introduced that behaviour.

@oxidase Could you have a look at this? I will also look into but my experience in this codebase is very small.

oxidase:
@pongo710 thanks for reporting! I can reproduce it but with slightly different durations: 42min vs 30s by 
```
  if way:id() == 284515678 then
     result.weight = 10000
  end
```

oxidase:
@pongo710 please check 1e21e0d with fix of json output for duration

pongo710:
@oxidase Thanks for the quick fix. It resolved our issue.

TheMarex:
Heads up, I'm currently rebasing this on `master`.

TheMarex:
@freenerd I have an API question:

This PR enables the user to set the name of the weight that is used for routing. That enables us to return the weight-value correctly named, for example inside the `RouteStep` object.
Say you would set in the profile `properties.weight_name = "safety"` this object would look like this:

```js
{
  "distance": 23.3,
  "duration": 55.2,
  "safety": 450
  /* rest of RouteStep object */
}
```

Making API property names depending on profile values is unusual, but might be more user-friendly then using an abstract term like `weight`. Any concerns from an api-directions standpoint here?

freenerd:
- Assumptions:
  - This is per `RouteStep` and not per `edge`
  - There is only one global weight name defined in the profile, and not different weight names

Dynamic key names feel like bad design, since it's not stable where a user might programmatically find the weight edge value. Instead I would split weight name and value:

```js
{
  "distance": 23.3,
  "duration": 55.2,
  "weight": {
    "name": "safety",
    "value": 450
  }
  /* rest of RouteStep object */
}
```

TheMarex:
Thanks for the input @freenerd.

> This is per RouteStep and not per edge

The name of the weight is global to the profile. My main concern with the explicit key-value encoding is repetition. Since the name of the weight is global, we should aim to not encode it that often.

We could include a property `weight_name`, `weight_type` or similar at the `Route` level. Using the `Route` level and not above (say global to the JSON object) would enable us to return routes that used different weights in the same API response. Inside all objects below that level we would use the key `weight` to reference it.

Example for what that would like with our current car profile (note: weight is the same as duration here):

```js
{
"waypoints": [Waypoint, ..],
"routes": [
  {
    "duration": 155,
    "distance": 55,
    "weight": 155.0,
    "weight_name": "duration",
    "legs": [
      {
         "annotation": {
           "nodes": [...],
           "datasources": [...],
           "duration": [...],
           "distance": [...],
           "weight": [...]
         },
         "steps":  [
         {
            "duration": 10,
            "distance": 5,
            "weight": 5
         }
         ... ]
      }
    ]
  }
]
}
```
Would that work for you as well?

freenerd:
> enable us to return routes that used different weights in the same API response

So even if we'd allow multiple weights in the graph, we'd only ever make them switchable per route request, but not within one route request? Then having `weight_name` per route makes sense to me 👍 

>  (note: weight is the same as duration here):

Would we even want to include `weight` at all then? I guess for most use cases, `weight === duration` will be true, thus we would not need to represent `weight` in the response.

daniel-j-h:
Having to do this two-step lookup through `weight_name` is horrible for all statically-typed APIs. I'm not saying this is a blocker here just something to keep in mind.

Example:

- https://github.com/daniel-j-h/hosrm-qa/blob/master/src/Response.hs
- https://github.com/daniel-j-h/hosrm-qa/blob/master/src/Api.hs

freenerd:
@daniel-j-h can you post an example of what you mean by `two-step lookup through weight_name`?

daniel-j-h:
Making field names (not values) in the response JSON configurable where clients have to do e.g.

    weight_name = getWeightFieldFromProfile(); // returns "safety"
    print(result[weight_name]);

or even

    weight_name = getWeightFieldFromResult(result); // returns "safety"
    print(result[weight_name]);

freenerd:
@daniel-j-h so I guess you are advocating against https://github.com/Project-OSRM/osrm-backend/pull/2399#issuecomment-251640010 and I agree that we should not do that.

TheMarex:
> Would we even want to include weight at all then? I guess for most use cases, weight === duration will be true, thus we would not need to represent weight in the response.

My concern is that it would make it less consistent, since it would then again depend on what kind of weight was used in the profile if the property exists or not.

Allowing to specify a property name had the nice property that if the weight was "duration" or "distance" it would not be encoded twice (because there are already fields with that name anyway). However I see how this would make it difficult for statically typed languages to have a native interface.

> So even if we'd allow multiple weights in the graph ...

The idea is actually exactly the opposite. We could allow to return multiple `Route` objects with different weights. Thinking about something like `/v1/route/distance+duration/{lon},{lat};{lon},{lat}` could return routes for distance and duration profiles, etc. But sure this is very much future talk.

So to conclude I would propose the static identifier `weight`, which is present unconditionally and a new property `weight_name` for the `Route` objects that makes it possible for the client to know what the weight value represents. Is that okay with everyone?

danpat:
My two cents a would be to use the term `cost` instead of `weight`. If the concern is to use a term that'll be somewhat meaningful to people that don't work on graphs all day, then I think `cost` conveys a clearer message (fuel cost, time cost, energy cost, distance cost, etc).

include/engine/routing_algorithms/routing_base.hpp
```diff
@@ -562,14 +561,14 @@ template <class DataFacadeT, class Derived> class BasicRoutingInterface
...
-                        int &distance,
+                        int &weight,
                         std::vector<NodeID> &packed_leg,
                         const bool force_loop_forward,
                         const bool force_loop_reverse,
                         int duration_upper_bound = INVALID_EDGE_WEIGHT) const
     {
         NodeID middle = SPECIAL_NODEID;
-        distance = duration_upper_bound;
+        weight = duration_upper_bound;
```
danpat: `weight_upper_bound` perhaps?

src/extractor/edge_based_graph_factory.cpp
```diff
@@ -241,8 +247,7 @@ unsigned EdgeBasedGraphFactory::RenumberEdges()
             // oneway streets always require this self-loop. Other streets only if a u-turn plus
             // traversal
             // of the street takes longer than the loop
-            m_edge_based_node_weights.push_back(edge_data.distance +
-                                                profile_properties.u_turn_penalty);
+            m_edge_based_node_weights.push_back(edge_data.weight);
```
jakepruitt: @TheMarex question about this line here - why was the `u_turn_penalty` removed from the `edge_based_node_weights`? On the branch, this test [in guidance](https://github.com/Project-OSRM/osrm-backend/blob/master/features/guidance/turn-angles.feature#L1091-L1120) fails to create a route unless the extra 200 is added to the edge based node weights. I can't tell why this one line makes a difference, but I do know if I add an arbitrary `200` to each node weight again, that test passes again.

MoKob: @jakepruitt this might actually be a bug due to self-loops in our code. The test forces our path to the right.
When checking for possible shorter paths during contraction, we probably don't consider these `forced` directions. As long as the u-turn is reasonably fast, we could end up omitting a required path.

But this is just a guess here. You could check by (temporarily) adjusting https://github.com/Project-OSRM/osrm-backend/blob/master/include/contractor/graph_contractor.hpp#L824 to always insert edges. If that fixes the problem, we sadly have a bug in relation to bearings and self-loops.

jakepruitt: 🙌 ❤️ thanks @MoKob I'll look into that.

src/extractor/extraction_containers.cpp
```diff
@@ -390,6 +390,8 @@ void ExtractionContainers::PrepareEdges(ScriptingEnvironment &scripting_environm
 
         weight = scripting_environment.ProcessSegment(
             edge_iterator->source_coordinate, target_coord, distance, weight);
+        duration = scripting_environment.ProcessSegment(
```
TheMarex: This probably needs an interface change for `ProcessSegment` to return both a weight and duration value. Maybe we should adapt the same interface other `Process*` functions have: It takes a `ExtractionSegment` object and modifies that.

TheMarex:
I don't think we should wait for `6.0` anymore to merge this. In my original implementation I broke the interface of the `turn_function` which is the reason we had to bump this to a major release. As always I underestimated the inertia to get breaking changes merged, hence this was a bad idea.

However we could make this work by still supporting the old `turn_function` interface but throw an error if it is used in conjunction with specifying the `weight` manually.

frodrigo:
If I not wrong, with this feature only weight will be available on table plugin and no more time, doesn't it ?

oxidase:
Finished branch rebase and added tests to cover functionality.

Still some things to do, but out scope of the PR:
* Cleanup of LoadEdgeExpandedGraph in contractor.cpp.
* FIXME https://github.com/Project-OSRM/osrm-backend/blob/feature/edge_weights/include/engine/guidance/assemble_geometry.hpp#L95
* Adjust tile plugin, e.g. https://github.com/Project-OSRM/osrm-backend/blob/feature/edge_weights/src/engine/plugins/tile.cpp#L796 weight is used as duration that is correct only if weight_name="duration".

Points for discussion:
* API naming: `weight` vs `score`, ...
* `(forward|backward)_weight_per_meter`: here name can be changed to `(forward|backward)_rate` or similar, because "weight_per_meter" is a synonym of speed and used as
`segment_weight = (weight[=ValueByEdge] / #segments) if weight>0  else (distance / forward_weight_per_meter[=ValueByMeter])`. This makes weight values similar to durations. Also there is an implicit restriction `0 < weight_per_meter <= 1` to avoid an underflow in division. 
* I have impression that it is better to decouple meaning of weights and durations and use integer values for weights in API and avoid division by 10, so if weight_name is "duration" and duration is 5.6 than weight will be 56, but it depends on use cases for weights if weight_name is not "duration".

TheMarex:
> API naming: weight vs score, ...

Yes this needs to be decided and the naming is hard. Maybe we can have a look at what other people call this?

> FIXME https://github.com/Project-OSRM/osrm-backend/blob/feature/edge_weights/include/engine/guidance/assemble_geometry.hpp#L95

This seems to be broken in `master` as well. This definitely needs a switch on `target_reversed` from what I can see. Not sure about the local path, best would be to re-read the logic in `UnpackPath` that generates that array. I think we might already account for that.

https://github.com/Project-OSRM/osrm-backend/blob/master/include/engine/guidance/assemble_geometry.hpp#L92-L97

> I have impression that it is better to decouple meaning of weights and durations and use integer values for weights in API and avoid division by 10, so if weight_name is "duration" and duration is 5.6 than weight will be 56, but it depends on use cases for weights if weight_name is not "duration".

For a long time we actually exposed the raw "weight" (deci-seconds) to the users and this has been a major source of confusion. To allow typical units for the weight (seconds, meters) I think it makes sense to at least allow for fixed-point values. Maybe we should surface the precision explicitly to the user in the profile? Like `weight_precision = 1`?

> Adjust tile plugin, e.g. https://github.com/Project-OSRM/osrm-backend/blob/feature/edge_weights/src/engine/plugins/tile.cpp#L796 weight is used as duration that is correct only if weight_name="duration".

Not sure what to do at that location. /cc @danpat

danpat:
Internally, I think `weight` and `duration` make sense.

In the Lua profiles, I propose the following change to the interface/naming:

  - `way.forward_routing_speed`  - used to set the `weight`
  - `way.forward_eta_speed`  - used to set the `duration`
  - `properties.routing_name` or `properties.routing_type`

I think this simplifies how we reason about adjusting these values - it's fairly intuitive that if you slow down the routing speed, affected ways will be avoided.  Keeping the scale of the two properties the same makes it easy to understand.

For turns, I propose a similar approach:

  - `turn.eta_duration`
  - `turn.routing_duration`

I don't see a need to make the properties in the profiles completely unit-less - calling both values a `speed` is OK, and much easier to comprehend.  Keeping to `weight` and `duration` inside the code is fine, but we could consider using `routing_duration` and `eta_duration` there as well.

danpat:
For the debug tiles, I'd like to expose both values on the tiles: `weight` and `duration` - we will want to be able to visualize both the routing data and the ETA data for debugging.

danpat:
I missed @oxidase 's mention of `forward_rate` as an approach - I retract my proposal and put my vote behind that 👍 

So, within the Lua profile, we should see:
```
Values used for routing:
turn.weight
way.forward_rate

Time-based values that contribute to ETA:
turn.duration
way.forward_speed
```

TheMarex:
Capturing from slack conversation with @danpat we are going with @oxidase proposal:

- `{forward,backward}_rate` and `{forward,backward}_speed`
- `weight` and `duration`

These have the advantage:

1. `rate` is not an overloaded term: It literally only refers to a ratio of values which is as much as we know about it anyway. Hopefully this will lead people to actually read the docs and don't assume stuff.

2. `duration` and `speed` have already very specific meaning that we don't want users to confuse here.

Going to `sed` this.

TheMarex:
I explored the `ProfileProperties::weight_precision` option and major blocker is that it is hard to circumvent https://github.com/Project-OSRM/osrm-backend/blob/feature/edge_weights/src/engine/api/json_factory.cpp#L245. There was a reason we needed this rounding here and I suspect it had something to do with response size. Maybe we can just remove it?

oxidase:
@TheMarex no, the reason only to have similar to the `duration` value. Originally i decided to return integer values as `route_step.values["weight"] = step.weight` but it is even more confusing

oxidase:
for api_version=1 tests pass, but for api_version=0 34 tests fail, can be related to https://github.com/Project-OSRM/osrm-backend/pull/2399/commits/1198689fd5913ac2ea80b4ecd2d67723538f6487 as to new weight calculation that can be a regression for api_version=0

oxidase:
Finished back-compatibility testing. There are 4 tests that fail for profiles with api_version=0:
* tests [Bike - Cycle tracks/lanes should enable biking](https://github.com/Project-OSRM/osrm-backend/blob/feature/edge_weights/features/bicycle/cycleway.feature#L8) and [Bike - Left/right side cycleways on implied oneways](https://github.com/Project-OSRM/osrm-backend/blob/feature/edge_weights/features/bicycle/cycleway.feature#L43) hit newly introduced assertion at https://github.com/Project-OSRM/osrm-backend/blob/feature/edge_weights/src/extractor/extractor_callbacks.cpp#L149
and require `result.backward_speed = bicycle_speeds["cycleway"]` for api_version=0 at https://github.com/Project-OSRM/osrm-backend/blob/feature/edge_weights/profiles/bicycle.lua#L369-L376

* tests [Weighting based on raster sources](https://github.com/Project-OSRM/osrm-backend/blob/feature/edge_weights/features/raster/weights.feature#L63-L79) and [Weighting based on raster sources](https://github.com/Project-OSRM/osrm-backend/blob/feature/edge_weights/features/raster/weights.feature##L45-L6) fail due to incompatible `ProcessSegment` function at https://github.com/Project-OSRM/osrm-backend/blob/feature/edge_weights/src/extractor/scripting_environment_lua.cpp#L630 because `InternalExtractorEdge::WeightData` was changed to `osrm::extractor::detail::ByEdgeOrByMeterValue`, so keeping back-compatibility will require struct redefinition old type as `api_version0::InternalExtractorEdge::WeightData`.

If it will not cause problems i would change the API version 0 to: the 4th parameter of segment_function is a reference to double value `duration`.

Besides of these two points i think PR is ready to be squashed and merged.

TheMarex:
> If it will not cause problems i would change the API version 0 to: the 4th parameter of segment_function is a reference to double value duration.

Yes that sounds like the best approach to me, good catch!

oxidase:
Rebased and squashed into one commit. After tests update, just found another interface inconsistency:
weights in CSV speed files must be in `weight*10^weight_precision` otherwise for https://github.com/Project-OSRM/osrm-backend/blob/feature/edge_weights/src/contractor/contractor.cpp#L270 and https://github.com/Project-OSRM/osrm-backend/blob/feature/edge_weights/src/contractor/contractor.cpp#L818 `weight_precision` from `ProfileProperties` must be loaded. The same is for turn weights at https://github.com/Project-OSRM/osrm-backend/blob/feature/edge_weights/src/contractor/contractor.cpp#L847 that are multiplied by 10.

TheMarex:
@oxidase yes good catch. We need to adjust these from the values that are contained in `ProfileProperties` as well.

I went through this thread to see if we didn't forget something and this comment stuck out:

> If I not wrong, with this feature only weight will be available on table plugin and no more time, doesn't it ?

Which is absolutely correct. https://github.com/Project-OSRM/osrm-backend/blob/feature/edge_weights/include/engine/routing_algorithms/many_to_many.hpp#L196 However we label this as `durations` value which is not true with this PR anymore: https://github.com/Project-OSRM/osrm-backend/blob/feature/edge_weights/include/engine/api/table_api.hpp#L68-L69

I think the original disconnect stems from the fact that this was always only designed to be used with the `route` plugin in combination with car navigation, since then the scope has widened. Our options here are:

1. Ignore this and emit wrong distance tables for profiles that use this new feature. (`ship with known regression`)
2. Extend the data we store in the graph https://github.com/Project-OSRM/osrm-backend/blob/feature/edge_weights/include/contractor/query_edge.hpp#L34 to include a `durations` property, so we don't have to unpack each path to get the duration. This will mean about 4-8GB more memory consumption for the server and we need to adapt the contractor to compute this value on each node contraction (probably slight slow-down). Adapt [this](https://github.com/Project-OSRM/osrm-backend/blob/feature/edge_weights/include/contractor/graph_contractor.hpp#L813) for duration summation and extend [ContractorEdge](https://github.com/Project-OSRM/osrm-backend/blob/feature/edge_weights/include/contractor/graph_contractor.hpp#L55).
3. Adapt the `table` plugin to unpack each of the `N x M` paths in the matrix and use the same approach as for the route plugin. (slow query)

/cc @danpat for input on our options.

danpat:
@TheMarex I'm inclined to go for #2 - and while we're at it, we should probably just bite the bullet and support https://github.com/Project-OSRM/osrm-backend/issues/1353, because it's basically the same thing.

My opinion is that we should probably admit defeat and spend the memory here.  If it can be made zero cost if we're not using it, then most folks won't be worse off.  I think it's OK to have significantly larger RAM requirements for the split weight/duration feature - it's a significant feature and comes at a cost.

oxidase:
@TheMarex `EdgeBasedEdge` also must have the `duration` value because durations are propagated from https://github.com/Project-OSRM/osrm-backend/pull/2399/commits/510d9a688fb1d095ef0122484b80cc8baa18966b#diff-341ad719a96c9c990d1255635d7f3dbdR549
to `ContractorEdge` and than to `QueryEdge`

It would require even more memory, but on the other side will fix duration updates at https://github.com/Project-OSRM/osrm-backend/pull/2399/commits/510d9a688fb1d095ef0122484b80cc8baa18966b#diff-5d25f965a92c5af3a4dfd03cf4f7616aR852

At the moment the regression test `features/testbot/distance_matrix.feature:223` in the branch does not work because at https://github.com/Project-OSRM/osrm-backend/pull/2399/commits/510d9a688fb1d095ef0122484b80cc8baa18966b#diff-b6276cd8f352289cbb3a81503ddc2fe9R226 for edges 0->0 and 1->1 from snapped points https://github.com/Project-OSRM/osrm-backend/pull/2399/commits/510d9a688fb1d095ef0122484b80cc8baa18966b#diff-b6276cd8f352289cbb3a81503ddc2fe9L69-L111 have duration is 0.  The only way that i see is to add `Get(Forward|Reversed)DurationPlusOffset` and extend `HeapData` with  duration. 
Please could you check if it is ok? It would be nice to have easier solution.

weight suffixes `_E`, `_C`, `_Q`, `_N` are just for review to see data propagation and will be removed.

oxidase:
@TheMarex  added `duration` field to edges `EdgeBasedEdge`, `UncontractedEdgeData`, `ContractorEdgeData`, `QueryEdge`, `NodeBucket` and added a new heap `ManyToManyHeapData` for the table plugin.

Still has to be decided if `SegmentBlock` has to be updated https://github.com/Project-OSRM/osrm-backend/blob/024f1e10e50231acd693a599a635488a380a17ff/include/extractor/edge_based_graph_factory.hpp#L64 to include `duration` data. This extension will  increase memory consumption but also allows to have correct durations at https://github.com/Project-OSRM/osrm-backend/blob/024f1e10e50231acd693a599a635488a380a17ff/src/contractor/contractor.cpp#L852

TheMarex:
> Still has to be decided if SegmentBlock has to be updated https://github.com/Project-OSRM/osrm-backend/blob/024f1e10e50231acd693a599a635488a380a17ff/include/extractor/edge_based_graph_factory.hpp#L64 to include duration data.

Yes it needs to be extended too and we need to make duration update-able via CSV. 👍

oxidase:
updated the branch with durations in SegmentBlock and unrelated fixes for windows tests that could hit the MAX_PATH limit as `c:\work\osrm-backend\test\cache\guidance\directional-ref-destination-signs.feature\2e90234f3bb5f7de24d2c7feef552243\a26e9bc926552aea0548abf0693c5a42\7_car_route_name_assembly_with_destination_signs_accounting_for_directionalref_tags.osrm.turn_duration_penalties` is 261 characters.

@danpat here a list of changes in memory consumption

struct | size before | size after | file | planet size before | planet size after | change
 --- | --- | --- | --- | --- | --- | --- 
 SegmentBlock        | 20 |  24 | ".edge_segment_lookup" | 55362898684 | 66359954476 | ~20% 
 QueryEdge::EdgeData |  8 |  12 | ".hsgr"                | 19260214496 | 25412531156 | ~32% 
 EdgeBasedEdge       | 16 |  20 | ".ebg"                 |  8840019396 | 11050439664 | ~25% 
 PhantomNode         | 80 | 104 | hint                   |             |             | ~30%

src/extractor/guidance/turn_discovery.cpp
```diff
@@ -41,7 +41,7 @@ bool findPreviousIntersection(const NodeID node_v,
      * PREVIOUS_ID. To verify that find, we check the intersection using our PREVIOUS_ID candidate
      * to check the intersection at NODE for via_edge
      */
-    const constexpr double COMBINE_DISTANCE_CUTOFF = 30;
+    const constexpr double COMBINE_WEIGHT_CUTOFF = 30;
```
TheMarex: This should actually be `COMBINE_DISTANCE_CUTOFF`, looking at the coordinate code below.

src/extractor/scripting_environment_lua.cpp
```diff
@@ -253,24 +336,54 @@ void Sol2ScriptingEnvironment::InitContext(LuaScriptingContext &context)
...
-                                                      "weight_data",
-                                                      &InternalExtractorEdge::weight_data);
+                                                      "weight",
+                                                      &InternalExtractorEdge::weight_data,
+                                                      "duration",
+                                                      &InternalExtractorEdge::duration_data);
 
-    context.state.new_usertype<InternalExtractorEdge::WeightData>(
-        "WeightData", "speed", &InternalExtractorEdge::WeightData::speed);
+    // context.state.new_usertype<InternalExtractorEdge::WeightData>(
```
TheMarex: Can be removed?

unit_tests/library/route.cpp
```diff
@@ -50,6 +50,7 @@ BOOST_AUTO_TEST_CASE(test_route_same_coordinates_fixture)
           json::Array{{json::Object{
               {{"distance", 0.},
                {"duration", 0.},
+               {"weight_name", "duration"},
```
TheMarex: Looks like we are not exposing the `weight` property for the `Route` and `RouteLeg` objects. This needs an adjustment in `json_factory.cpp` and `src/engine/guidance/assemble_route.cpp`, `include/engine/guidance/assemble_leg.hpp`

src/extractor/scripting_environment_lua.cpp
oxidase: sure

src/extractor/guidance/turn_discovery.cpp
oxidase: ok

include/engine/search_engine_data.hpp
```diff
@@ -17,24 +17,41 @@ struct HeapData
...
+
+    using ManyToManyHeapPtr = boost::thread_specific_ptr<ManyToManyQueryHeap>;
+
     static SearchEngineHeapPtr forward_heap_1;
     static SearchEngineHeapPtr reverse_heap_1;
     static SearchEngineHeapPtr forward_heap_2;
     static SearchEngineHeapPtr reverse_heap_2;
     static SearchEngineHeapPtr forward_heap_3;
     static SearchEngineHeapPtr reverse_heap_3;
+    static ManyToManyHeapPtr many_to_many_heap;
```
oxidase: ok,will check and remove

include/extractor/profile_properties.hpp
```diff
@@ -12,11 +14,18 @@ const constexpr auto DEFAULT_MAX_SPEED = 180 / 3.6; // 180kmph -> m/s
 
 struct ProfileProperties
 {
+    enum
```
oxidase: i don't remember, will check `static const constexpr std::sizt_t`

features/testbot/side_bias.feature
```diff
@@ -9,13 +9,10 @@ Feature: Testbot - side bias
         """
 
     Scenario: Left hand bias
-        Given the profile file "testbot" extended with
+        Given the profile file "car" extended with
```
oxidase: no, i think it is `testbot` related, i check it

features/lib/osrm_loader.js
```diff
@@ -78,9 +78,10 @@ class OSRMDirectLoader extends OSRMBaseLoader {
         if (this.osrmIsRunning()) return callback(new Error("osrm-routed already running!"));
 
         this.child = this.scope.runBin('osrm-routed', util.format("%s -p %d", this.inputFile, this.scope.OSRM_PORT), this.scope.environment, (err) => {
-          if (err && err.signal !== 'SIGINT') {
-              throw new Error(util.format('osrm-routed %s: %s', errorReason(err), err.cmd));
-          }
+            if (err && err.signal !== 'SIGINT') {
+                this.child = null;
```
oxidase: Yes, it should be in the master. I changed SIGTERM to SIGINT because for windows npm tests it is possible to check that the OSRM process was killed by `osrmDown` and do not report an error. If test run was successful than `this.child` will be updated in `osrmUp` with the next process. 

This line fixes the problem that if the test failed and not killed than the new process will not be started because `osrmIsRunning()` will return true in `osrmUp`. 

So if one test failed than all next tests will fail also due to 5sec timeout and this is killer to wait appveyor jod for ~ 5000 seconds if almost the first test fails.

If `this.child` set to null here that is ok because the process is already exited then the new process will be started in `osrmUp`.

features/lib/osrm_loader.js
```diff
@@ -116,6 +117,7 @@ class OSRMDatastoreLoader extends OSRMBaseLoader {
 
         this.child = this.scope.runBin('osrm-routed', util.format('--shared-memory=1 -p %d', this.scope.OSRM_PORT), this.scope.environment, (err) => {
             if (err && err.signal !== 'SIGINT') {
+                this.child = null;
```
oxidase: ^

features/options/profiles/version0.feature
```diff
@@ -71,7 +72,7 @@ end
 
 function turn_function (angle)
   print('turn_function ' .. angle)
-  return angle == 0 and 0 or 42
+  return angle == 0 and 0 or 17
```
oxidase: just a dummy number to have expectations 15+1.7 not so close to 20 as 15+4.2 for 42.
Basically here i did not get why a-e path is 20s, but a-b and a-d is 15+turn_penalty, 
i would expect 20+turn_penalty

CHANGELOG.md
```diff
@@ -1,3 +1,10 @@
+# 6.0.0
+  - Profiles:
```
oxidase: ok, will update changelog, should be tagged as 5.6.0 change?

TheMarex: Yep also should go under 5.6

include/extractor/profile_properties.hpp
oxidase: `static const` requires a definition some where in cpp file, but i would keep a constant as a value in an unnamed enum in scope of the struct

oxidase:
@TheMarex updated and rebased the brach, but in some places i did not make changes:

* [traffic_light_penalty](https://github.com/Project-OSRM/osrm-backend/blob/8d1f156f67c5e90b47760ac6cd5ae9d3e8d4c33a/profiles/testbot.lua#L125-L127) is still used in `turn_function` to have old behavior, because [TODO](https://github.com/Project-OSRM/osrm-backend/blob/8d1f156f67c5e90b47760ac6cd5ae9d3e8d4c33a/profiles/testbot.lua#L46) fix is required
* `MAX_WEIGHT_NAME_LENGTH` is still a value in enum
* `forward_heap_3` and `reverse_heap_3` are still in use
* [FIXME](https://github.com/Project-OSRM/osrm-backend/blob/8d1f156f67c5e90b47760ac6cd5ae9d3e8d4c33a/include/engine/guidance/assemble_geometry.hpp#L95-L97) is still there but now related only the situation when source and target phantoms on one edge

include/extractor/profile_properties.hpp
TheMarex: Hrm just looked this up because I was sure I had seen stuff like `static const constexpr std::size_t value = 123;` in class definitions before.

I'm trying to interpret [this](http://stackoverflow.com/a/34053854/1944243) comment. From my reading this would allow for doing this as long as the `static const constexpr` member is of integral type.

Is my reading of the standard correct here?

TheMarex:
> traffic_light_penalty is still used in turn_function to have old behavior, because TODO fix is required

You are right, I confused myself with my own changelog entry. Currently it is:

> - `traffic_light_penalty` is deprecated. `traffic light penalties` now need to be set over in the node function in `result.weight_penalty` and `result.duration_penalty` fields

Correct however would be:

> - `properties.traffic_light_penalty` is deprecated. Traffic light penalties now need to be set over in the turn function. Each turn with a traffic light is marked with `ExtractionTurn::has_traffic_light = true`.

This looks good to me. I think we can merge this (wow so weird).

oxidase:
looks like std::size_t is not integral type? will change to `static constexpr int MAX_WEIGHT_NAME_LENGTH = 255;`

djschilling:
Thanks for all the effort you are putting into this.
This pull request getting merged is really great for us.

danpat:
@djschilling @emiltin and anyone else who has been watching this ticket - we've merge into master but feedback on this feature would be really helpful.  Anyone who has the time to test it out, please do and open tickets with any problems/confusion you encounter.

djschilling:
We are using an earlier state of the pull request since the start of the year and it does exactly what it should.
We use it to correctly handle `access:destination` tags, by putting a high weight to `access:destination` ways.

One thing that we encountered is a performance loose from about 40%. Its not critical for us, but i think its worth mentioning.
I have no time now to test the currently merged version, so feedback from someone else would probably be good.

Issue #2066, opened 2016-03-08
Debug tiles need turn penalty data


danpat:
Currently, the debugging tiles only expose the node-based-edge weights.

However, when we route, we use the edge-based graph, and the weights on the edge-based graph are the sum of the node-based-edge weights, *plus* some additional turn/stop/u-turn costs.

In the previous iteration of graph debugging, when we were printing out GeoJSON, we dumped turn data here:

https://github.com/Project-OSRM/osrm-backend/blob/d0644bace9e121963dce87d935e1e2555842ef7a/src/extractor/edge_based_graph_factory.cpp#L470

https://github.com/Project-OSRM/osrm-backend/blob/d0644bace9e121963dce87d935e1e2555842ef7a/src/extractor/edge_based_graph_factory.cpp#L459

(and it looks like we were missing output for this spot:  https://github.com/Project-OSRM/osrm-backend/blob/d0644bace9e121963dce87d935e1e2555842ef7a/src/extractor/edge_based_graph_factory.cpp#L473, although the code for this *did* exist at one point:  https://github.com/Project-OSRM/osrm-backend/blob/d0644bace9e121963dce87d935e1e2555842ef7a/include/util/debug_geometry.hpp#L199).

But now, we can't see these numbers when debugging.

@daniel-j-h Can you take a look and see if you can expose vector tile `Point` elements with some of these properties?

Brain dump of thoughts:

   1. Getting the turn angles is going to be difficult - it's going to require walking the graph, getting packed geometries and finding the first/last segments for each node-based-edge, then calculating the angle.
   2. Turn weights are also going to need to be calculated - add up the weights of the node-based-edge, then subtract from the edge-edge.

I suspect u-turn and stop-light penalties are going to be impossible to expose individually - best we can do is identify the "turn" and the "penalty" that goes along with it.

Ideally, the output we're going for could be rendered something like this:

https://api.mapbox.com/v4/danpat.f040b156/page.html?access_token=pk.eyJ1IjoiZGFucGF0IiwiYSI6ImI0WThCVWMifQ.BP_j7Kiv5SNL-p-kTfXNxg#19/37.77515/-122.41548

(This map is missing traffic-light penalties, which apply to straight-through intersections).

TheMarex:
How urgent is this? Can it wait until `5.1.0`?

PR #2400, opened 2016-05-12
Debug tiles: turn penalties

lbud:
Closes https://github.com/Project-OSRM/osrm-backend/issues/2066.

image
😁 

@danpat want to have a look?

TODO:

- [x] Mapbox GL can't do data driven symbol rotations yet
- [x] We might be missing some turns in the data as per @MoKob


lbud:
@danpat ok now 👌 fixed the tests

include/engine/routing_algorithms/routing_base.hpp
```diff
@@ -484,6 +484,62 @@ template <class DataFacadeT, class Derived> class BasicRoutingInterface
         unpacked_path.emplace_back(t);
     }
 
+    void UnpackEdgeToEdges(const NodeID s, const NodeID t, std::vector<EdgeData> &unpacked_path) const
```
danpat: I'd love to see us start adding docs to functions.  Good examples happening in `node-osrm` here:

https://github.com/Project-OSRM/node-osrm/blob/api-documentation/src/node_osrm.cpp#L73-L90

OSRM is woefully commented, let's fix it :-)

src/engine/plugins/tile.cpp
```diff
@@ -162,6 +197,25 @@ FixedLine coordinatesToTileLine(const util::Coordinate start,
 
     return tile_line;
 }
+
+FixedPoint coordinatesToTilePoint(const util::Coordinate point, const detail::BBox &tile_bbox)
+{
+    FloatPoint geo_point{static_cast<double>(util::toFloating(point.lon)),
+                         static_cast<double>(util::toFloating(point.lat))};
+
+    double px_merc = geo_point.x * util::web_mercator::DEGREE_TO_PX;
```
danpat: `const` on all these variables

src/engine/plugins/tile.cpp
```diff
@@ -205,10 +298,79 @@ Status TilePlugin::HandleRequest(const api::TileParameters &parameters, std::str
...
+                        const auto coord_c = facade.GetCoordinateOfNode(possible_next_node.first);
+                        const auto c_bearing = static_cast<uint64_t>(
+                            util::coordinate_calculation::bearing(coord_b, coord_c));
+
+                        const auto angle_out_offset = use_point_value(c_bearing);
+                        const auto angle_weight_offset = use_point_value(possible_next_node.second);
+
+                        // TODO this is not as efficient as it could be because of repeated
+                        // angles_in
+                        edge_turn_data.emplace_back(detail::TurnData{
```
danpat: Good point.  Ideally, each value would only show up once in the layer attribute table.  But let's just get this shipped and make a ticket to remove duplicated values from the attribute table.

src/engine/plugins/tile.cpp
```diff
@@ -246,207 +407,326 @@ Status TilePlugin::HandleRequest(const api::TileParameters &parameters, std::str
...
+                            field.add_element(point_turn_data.in_angle_offset);
+                            field.add_element(1); // "bearing_out" tag key offset
+                            field.add_element(point_turn_data.out_angle_offset);
+                            field.add_element(2); // "weight" tag key offset
+                            field.add_element(point_turn_data.weight_offset);
+                        }
+                        {
+                            protozero::packed_field_uint32 geometry(
+                                feature_writer, util::vector_tile::FEATURE_GEOMETRIES_TAG);
+                            encodePoint(tile_point, geometry, start_x, start_y);
```
danpat: The `start_x` and `start_y` values probably aren't needed here.  Encoding points doesn't need to know the location of the last cursor command, each new feature starts at (0,0) already.  We can remove these parameters and make `encodePoint` just zigzag encode the `tile_point.x` and `tile_point.y` values directly.

Tracking the `start_x` and `start_y` values is only needed for multi-coordinate shapes, like lines and polygons.

The vector tile spec describes how the cursor move commands work here:
https://github.com/mapbox/vector-tile-spec/tree/master/2.1#43-geometry-encoding

lbud: @danpat cool — I sort of suspected as much but figured I could just start with the copypasta and then left it in there, will remove 👍

include/engine/routing_algorithms/routing_base.hpp
```diff
@@ -484,6 +484,69 @@ template <class DataFacadeT, class Derived> class BasicRoutingInterface
         unpacked_path.emplace_back(t);
     }
 
+    /**
+     * A duplicate of the above `UnpackEdge` function, but returning full EdgeData
+     * objects for the unpacked path. Used in the tile plugin to find outgoing
+     * edges from a given turn.
+     */
+
```
MoKob: As already said in your comment, this feel like unnecessary duplication of the unpacking routine.
We can find the `EdgeID` for two vertices s/t.
With the `EdgeID`, you can get access to the path data.
I would prefer if we not add an additional `unpack` routine but use the existing one and transform the result into EdgeData at the location it is needed.

src/engine/plugins/tile.cpp
```diff
@@ -113,6 +130,19 @@ inline bool encodeLinestring(const FixedLine &line,
     return true;
 }
 
+// from mapnik-vctor-tile
+// Encodes a point
+inline bool encodePoint(const FixedPoint &pt, protozero::packed_field_uint32 &geometry)
+{
+    geometry.add_element(9);
```
MoKob: some explanation of why this is `9` would be great. Either named (`const constexpr precision = 9`, `gemometry.add_element(precision)`) or some comment.

src/engine/plugins/tile.cpp
```diff
@@ -181,10 +230,46 @@ Status TilePlugin::HandleRequest(const api::TileParameters &parameters, std::str
     // This hits the OSRM StaticRTree
     const auto edges = facade.GetEdgesInBox(southwest, northeast);
 
-    std::vector<int> used_weights;
-    std::unordered_map<int, std::size_t> weight_offsets;
+    std::vector<int> used_line_ints;
```
MoKob: what is a `line int`? explanation of what this is supposed to be would be great/or a clearer name

include/engine/plugins/tile.hpp
```diff
@@ -23,8 +25,13 @@ namespace plugins
...
 class TilePlugin final : public BasePlugin
 {
+  private:
+    routing_algorithms::BasicRoutingInterface<
+        datafacade::BaseDataFacade,
+        routing_algorithms::ShortestPathRouting<datafacade::BaseDataFacade>> routing_base;
+
   public:
-    TilePlugin(datafacade::BaseDataFacade &facade) : BasePlugin(facade) {}
+    TilePlugin(datafacade::BaseDataFacade &facade) : BasePlugin(facade), routing_base(&facade) {}
```
MoKob: access to `routing_base` seems like the wrong approach to me. The plugin should not handle routing-related questions.
As discussed later, I feel that using the normal unpack routine and getting access to edge-data via the normal graph and locating the respective edge should be the way to go.

include/engine/routing_algorithms/routing_base.hpp
MoKob: To further clarify. NodeBasedDynamic graph offers findSmallestEdge (https://github.com/Project-OSRM/osrm-backend/blob/master/include/util/dynamic_graph.hpp#L254) which adds exactly the functionality used here in the modified `unpack`.

So a normal `unpack` and following use of `findSmallestEdge` should result in just the same output, which means the functionality here is technically already implemented.

danpat: `FindSmallestEdge` isn't quite the same, but I'm not clear on whether the difference is important:

UnpackPath uses:

```
        if ((facade->GetTarget(edge_id) == edge.second) && (weight < edge_weight) &&
            facade->GetEdgeData(edge_id).forward)
        {
            smaller_edge_id = edge_id;
            edge_weight = weight;
        }
```

whereas `FindSmallestEdge` uses:

```
    if (target == to && weight < smallest_weight)
    {
        smallest_edge = edge;
        smallest_weight = weight;
    }
```

MoKob: This looks like a shortcoming in FindSmallestEdge in that case. I would say that we have to add that behaviour (template parameter?) to FindSmallestEdge. In that progress, we could also use FindSmallestEdge in UnpackPath over the duplicated code that is used there at the moment.

Technically it should not make a difference, though, as forward/backward only are relevant if two equally valid paths exist.

Consider:

```
a -------------- b -------------- c
|                                 |
---------------- d ----------------
```
It could be that for a route `a` to `c`, both the path using `b` and the path using `d` have the same length. One might only be part of the backward search space, the other one of the forward search space.
In that case the path that is found depends on the order the queues are traversed, but the result is equally valid.
Since the search-graphs only contain traversable edges, both paths are valid.
If you are looking to find both possibilities in a specific way, we would need the distinction using `.forward`

TheMarex:
We punted on this for the moment:

- [ ] Mapbox GL can't do data driven symbol rotations yet
- [ ] We might be missing some turns in the data as per @MoKob

danpat:
Some progress:

  - I fixed some encoding problems for negative angles
  - To help make rendering easier, the tiles now encodes a `turn_angle`, rather than `bearing_out`, which saves some math on the rendering side.
  - Support in Mapbox GL is pending here: https://github.com/mapbox/mapbox-gl-js/tree/text_rotate_dds

With a local hacked-up-copy of mbgl, I get maps like this:

screen shot 2016-08-16 at 10 31 22 pm

Due to the CH structure we get *many* turn values, but not all.  That can't be rectified without loading the un-contracted CH into memory, or coming up with a separate data structure just to display this map.  This is as good as we'll get with the current data in RAM.

Next up:

  - [ ] update test cases
  - [ ] merge!

danpat:
@TheMarex @MoKob I've done some refactoring and tidy-up based on previous comments.

While the turn labelling isn't complete, I'd like to get this merge so that we can use it to try to debug *some* routing, even if it's not perfect.

@MoKob any chance we could use the ICD data to find more turn data?

MoKob:
@danpat can you elaborate what you imagine here?

Without knowing more about your intentions, my intuition would say the following: You can probably find out if you are missing some roads. So you can even guarantee in some cases that you are missing one. But you wouldn't have any additional data to test which ones are missing but to query the r-tree.

In addition, some intersections are represented a bit different in the ICD data than in OSM. If you have a road splitting into two separate ways, it can end up as a single item in the ICD.

include/util/static_graph.hpp
```diff
@@ -148,19 +148,32 @@ template <typename EdgeDataT, bool UseSharedMemory = false> class StaticGraph
...
+                                  const FilterFunction filter) const
     {
         EdgeIterator smallest_edge = SPECIAL_EDGEID;
         EdgeWeight smallest_weight = INVALID_EDGE_WEIGHT;
         for (auto edge : GetAdjacentEdgeRange(from))
         {
             const NodeID target = GetTarget(edge);
-            const EdgeWeight weight = GetEdgeData(edge).distance;
-            if (target == to && weight < smallest_weight)
+            const auto data = GetEdgeData(edge);
```
TheMarex: Using a const-ref might be better here. This will copy the edge data.

include/engine/edge_unpacker.hpp
```diff
@@ -0,0 +1,104 @@
...
+ * from beginning to end.
+ *
+ * @param packed_path_begin iterator pointing to the start of the NodeID list
+ * @param packed_path_end iterator pointing to the end of the NodeID list
+ * @param callback void(const std::pair<NodeID, NodeID>, const EdgeData &) called for each
+ * original edge found.
+ */
+
+template <typename DataFacadeT, typename RandomIter, typename Callback>
+inline void UnpackCHEdge(DataFacadeT *facade,
```
MoKob: Name and parameters don't match up for me. Packed path indicates more than a single CH edge?

include/engine/edge_unpacker.hpp
```diff
@@ -0,0 +1,104 @@
...
+                  RandomIter packed_path_begin,
+                  RandomIter packed_path_end,
+                  const Callback &callback)
```
MoKob: Is there a reason to use a two-parameter callback. In basics, this gets edges, right? So anything that needs more than the edges should query this from the edges, or not?
Otherwise we pass around additional data all the time.

include/engine/routing_algorithms/routing_base.hpp
```diff
@@ -217,84 +218,33 @@ template <class DataFacadeT, class Derived> class BasicRoutingInterface
...
+                                           const EdgeData &data) {
+
+                BOOST_ASSERT_MSG(!data.shortcut, "original edge flagged as shortcut");
+                unsigned name_index = facade->GetNameIndexFromEdgeID(data.id);
+                const auto turn_instruction = facade->GetTurnInstructionForEdgeID(data.id);
                 const extractor::TravelMode travel_mode =
                     (unpacked_path.empty() && start_traversed_in_reverse)
                         ? phantom_node_pair.source_phantom.backward_travel_mode
-                        : facade->GetTravelModeForEdgeID(ed.id);
+                        : facade->GetTravelModeForEdgeID(data.id);
```
MoKob: Nitpicking here, but data is a very general name. Would prefer something more verbose.

danpat:
I had a small brainwave regarding finding all turns.

We have the nodeID for everything visible on a tile from the RTree query.  The problem is that if we explore the CH, we don't necessarily find all maneuvers.

While it would be more expensive than what we're doing, we could do a nearest-neighbour search near the end of each piece of geometry, and attempt to find a path between the two.

screen shot 2016-08-24 at 6 53 00 pm

For the edge we're examining, if we calculate a coordinate 1m from the end, then do a nearest-neighbour search with a 2m radius, we should be able to find (with a reasonable confidence) all the connected segments.  These segments come along with the corresponding IDs, and we can attempt to find a route between our candidate edge and the seemingly connected edges.

In some cases there will be 1-ways, turn restrictions, true disconnections and we will either not find a path, or will find a very long path (greater than 2m).  We can discard these as non-possible turns.

@MoKob Opinions on the feasibility of this ?  It'd be fairly slow, but we could disable it for low zoom levels, only including this data in tiles that are of a limited geographic area to cap the load each tile creates on the server.

MoKob:
@danpat can you elaborate what exactly you are trying to solve as a problem? For me it is not entirely clear why you need these searches. What I pointed out is that we don't get all turns from edges within the bounding box. I can't find the concerns here in the PR anymore, but the situation is the following:

Given any `edge_based_node` `v` that we can find in the search graph, we can find turns to all other `edge_based_nodes` higher up in CH hierarchy. From the viewpoint of `v`, we don't see turns to any lower nodes.

Given a bounding box search that includes all `edge_based_nodes` around the intersection, the view from these lower nodes includes the missing turns.
Where does all of this become a problem? If we look at a node that is right on top of a bounding-box-border, we will miss out on these turns, since the lower level `edge_based_nodes` will not be included in the tile.
This is mainly an overlap issue, though.
So if we have hard tile borders, we should be fine searching in a slightly larger bounding box and restricting the data again post extraction.

include/engine/routing_algorithms/routing_base.hpp
```diff
@@ -217,84 +218,33 @@ template <class DataFacadeT, class Derived> class BasicRoutingInterface
...
+            packed_path_end,
+            [this,
+             &unpacked_path,
+             &phantom_node_pair,
+             &start_traversed_in_reverse,
+             &target_traversed_in_reverse](std::pair<NodeID, NodeID> & /* edge */,
+                                           const EdgeData &data) {
+
+                BOOST_ASSERT_MSG(!data.shortcut, "original edge flagged as shortcut");
+                unsigned name_index = facade->GetNameIndexFromEdgeID(data.id);
```
daniel-j-h: const auto

include/util/static_graph.hpp
```diff
@@ -148,19 +148,32 @@ template <typename EdgeDataT, bool UseSharedMemory = false> class StaticGraph
...
+     * tested or not.
+     *   Takes `EdgeData` as a parameter.
+     * @return the ID of the smallest edge if any were found that satisfied *filter*, or
+     * `SPECIAL_EDGEID` if no
+     *   matching edge is found.
+     */
+    template <typename FilterFunction>
+    EdgeIterator FindSmallestEdge(const NodeIterator from,
+                                  const NodeIterator to,
+                                  const FilterFunction filter) const
```
daniel-j-h: Taking filter by const-value does not make sense; see comment above: either by && and forwarding or by value (since the functor could be stateful).

include/util/typedefs.hpp
```diff
@@ -85,6 +85,8 @@ static const EdgeWeight INVALID_EDGE_WEIGHT = std::numeric_limits<EdgeWeight>::m
 
 using DatasourceID = std::uint8_t;
 
+using DatasourceID = std::uint8_t;
```
daniel-j-h: Needs to be removed; is same as two lines above, or am I missing something? Why does re-aliasing the same name even compile?

include/util/vector_tile.hpp
```diff
@@ -13,19 +13,22 @@ namespace vector_tile
...
 const constexpr std::uint32_t GEOMETRY_TYPE_LINE = 2;
 const constexpr std::uint32_t FEATURE_ATTRIBUTES_TAG = 2;
 const constexpr std::uint32_t FEATURE_GEOMETRIES_TAG = 4;
-const constexpr std::uint32_t VARIANT_TYPE_UINT32 = 5;
+const constexpr std::uint32_t VARIANT_TYPE_UINT64 = 5;
+const constexpr std::uint32_t VARIANT_TYPE_SINT64 = 6;
 const constexpr std::uint32_t VARIANT_TYPE_BOOL = 7;
 const constexpr std::uint32_t VARIANT_TYPE_STRING = 1;
 const constexpr std::uint32_t VARIANT_TYPE_DOUBLE = 3;
+const constexpr std::uint32_t VARIANT_TYPE_FLOAT = 2;
```
daniel-j-h: re-order?

src/engine/plugins/tile.cpp
```diff
@@ -68,9 +71,36 @@ struct point_type_i final
...
-using FloatLine = std::vector<detail::Point<double>>;
+// Used to accumulate all the information we want in the tile about
+// a turn.
+struct TurnData final
+{
+    TurnData(const util::Coordinate coordinate_,
+             const std::size_t _in,
+             const std::size_t _out,
+             const std::size_t _weight)
+        : coordinate(coordinate_), in_angle_offset(_in), turn_angle_offset(_out),
```
daniel-j-h: take coordinate by value, move in here

src/engine/plugins/tile.cpp
```diff
@@ -92,15 +122,20 @@ inline bool encodeLinestring(const FixedLine &line,
         return false;
     }
 
-    const unsigned line_to_length = static_cast<const unsigned>(line_size) - 1;
+    const unsigned LINETO_count = static_cast<const unsigned>(line_size) - 1;
```
daniel-j-h: all upper / all lower?

src/engine/plugins/tile.cpp
```diff
@@ -113,6 +148,29 @@ inline bool encodeLinestring(const FixedLine &line,
...
+inline bool encodePoint(const FixedPoint &pt, protozero::packed_field_uint32 &geometry)
+{
+    const constexpr int MOVETO_COMMAND = 9;
+    geometry.add_element(MOVETO_COMMAND);
+    const std::int32_t dx = pt.x;
+    const std::int32_t dy = pt.y;
+    // Manual zigzag encoding.
+    geometry.add_element(protozero::encode_zigzag32(dx));
+    geometry.add_element(protozero::encode_zigzag32(dy));
+    return true;
```
daniel-j-h: Why does this always return true? What's the point?

src/engine/plugins/tile.cpp
```diff
@@ -162,8 +220,56 @@ FixedLine coordinatesToTileLine(const util::Coordinate start,
...
+    std::array<NodeID, 2> path{{from, to}};
+    UnpackCHEdge(&facade,
+                 path.begin(),
+                 path.end(),
+                 [&unpacked_path](const std::pair<NodeID, NodeID> & /* edge */,
+                                  const datafacade::BaseDataFacade::EdgeData &data) {
+                     unpacked_path.emplace_back(data);
+                 });
+}
+} // ::detail namespace
```
daniel-j-h: This can be an anonymous namespace to keep the symbols hidden in this TU, right?

src/engine/plugins/tile.cpp
```diff
@@ -181,20 +287,318 @@ Status TilePlugin::HandleRequest(const api::TileParameters &parameters, std::str
...
-    // when we encode each feature in the tile.
+    // And again for integer values used by points.
+    std::vector<int> used_point_ints;
+    std::unordered_map<int, std::size_t> point_int_offsets;
+
+    // And again for float values used by points
+    std::vector<float> used_point_floats;
+    std::unordered_map<float, std::size_t> point_float_offsets;
+
+    uint8_t max_datasource_id = 0;
```
daniel-j-h: std::

src/engine/plugins/tile.cpp
```diff
@@ -181,20 +287,318 @@ Status TilePlugin::HandleRequest(const api::TileParameters &parameters, std::str
...
+
+    // This is where we accumulate information on turns
+    std::vector<detail::TurnData> all_turn_data;
+
+    // Helper function for adding a new value to the line_ints lookup table.  Returns
+    // the index of the value in the table, adding the value if it doesn't already
+    // exist
+    const auto use_line_value = [&used_line_ints, &line_int_offsets](const int &value) {
```
daniel-j-h: by value

src/engine/plugins/tile.cpp
```diff
@@ -181,20 +287,318 @@ Status TilePlugin::HandleRequest(const api::TileParameters &parameters, std::str
...
+                {
+                    f->second.push_back(edge.forward_segment_id.id);
+                }
+                else
+                {
+                    outgoing_edges[edge.u] = {edge.forward_segment_id.id};
+                }
+
+                // Add this edge-based-nodeid as an incoming to the target intersection
+                f = incoming_edges.find(edge.v);
```
daniel-j-h: This is dangerous, since it's easy to compare the first map's iterator to the second map's iterator (which is UB) here; make separate var.

src/engine/plugins/tile.cpp
```diff
@@ -181,20 +287,318 @@ Status TilePlugin::HandleRequest(const api::TileParameters &parameters, std::str
...
+                    edge.v, edge.u, edge.reverse_packed_geometry_id};
+            }
+        }
+
+        // Now, for every edge-based-node that we discovered (edge-based-nodes are sources
+        // and targets of turns).  EBN is short for edge-based-node
+        for (const auto &source_ebn : edge_based_node_info)
+        {
+            // Grab a copy of the geometry leading up to the intersection.
+            std::vector<NodeID> first_geometry;
```
daniel-j-h: This allocates and then deallocates a vector for every eb-node, what about hoisting it outside the loop?

src/engine/plugins/tile.cpp
```diff
@@ -181,20 +287,318 @@ Status TilePlugin::HandleRequest(const api::TileParameters &parameters, std::str
...
+                // due to oneways or turn restrictions. Given the edge-based-nodes that
+                // we're examining here, we *should* only find directly-connected edges, not
+                // shortcuts
+                if (smaller_edge_id != SPECIAL_EDGEID)
+                {
+                    // Check to see if it was a shortcut edge we found.  This can happen
+                    // when exactly?  Anyway, unpack it and get the first "real" edgedata
+                    // out of it, which should represent the first hop, which is the one
+                    // we want to find the turn.
+                    auto data = facade.GetEdgeData(smaller_edge_id);
```
daniel-j-h: const? or even const&?

src/engine/plugins/tile.cpp
```diff
@@ -181,20 +287,318 @@ Status TilePlugin::HandleRequest(const api::TileParameters &parameters, std::str
...
+                    if (data.shortcut)
+                    {
+                        std::vector<contractor::QueryEdge::EdgeData> unpacked_shortcut;
```
daniel-j-h: Same vector alloc/dealloc here

daniel-j-h: Hmm thinking of this, there is a trick we can do without re-writing the UnpackEdges function: use a boost function output iterator with a lambda instead of writing to a vector and then using only the first one (`boost::make_function_output_iterator`).

src/engine/plugins/tile.cpp
```diff
@@ -181,20 +287,318 @@ Status TilePlugin::HandleRequest(const api::TileParameters &parameters, std::str
...
+                        detail::UnpackEdgeToEdges(facade, source_ebn.first, target_ebn, unpacked_shortcut);
+                        data = unpacked_shortcut.front();
+                    }
+                    BOOST_ASSERT_MSG(!data.shortcut, "Connecting edge must not be a shortcut");
+
+                    // This is the geometry leading away from the intersection
+                    // (i.e. the geometry of the target edge-based-node)
+                    std::vector<NodeID> second_geometry;
```
daniel-j-h: Same here and below

src/engine/plugins/tile.cpp
```diff
@@ -181,20 +287,318 @@ Status TilePlugin::HandleRequest(const api::TileParameters &parameters, std::str
...
+                    facade.GetUncompressedGeometry(
+                        edge_based_node_info.at(target_ebn).packed_geometry_id, second_geometry);
+
+                    // Now, calculate the sum of the weight of all the segments.
+                    std::vector<EdgeWeight> forward_weight_vector;
+                    facade.GetUncompressedWeights(source_ebn.second.packed_geometry_id,
+                                                  forward_weight_vector);
+                    const auto sum_node_weight = std::accumulate(
+                        forward_weight_vector.begin(), forward_weight_vector.end(), 0);
```
daniel-j-h: `EdgeWeight{0}`?

src/engine/plugins/tile.cpp
```diff
@@ -224,272 +624,361 @@ Status TilePlugin::HandleRequest(const api::TileParameters &parameters, std::str
...
 
-                int forward_weight = 0;
-                int reverse_weight = 0;
+                    int forward_weight = 0;
+                    int reverse_weight = 0;
 
-                uint8_t forward_datasource = 0;
-                uint8_t reverse_datasource = 0;
+                    uint8_t forward_datasource = 0;
+                    uint8_t reverse_datasource = 0;
```
daniel-j-h: std::

src/engine/plugins/tile.cpp
```diff
@@ -224,272 +624,361 @@ Status TilePlugin::HandleRequest(const api::TileParameters &parameters, std::str
...
 
-                if (edge.forward_packed_geometry_id != SPECIAL_EDGEID)
-                {
-                    std::vector<EdgeWeight> forward_weight_vector;
-                    facade.GetUncompressedWeights(edge.forward_packed_geometry_id,
-                                                  forward_weight_vector);
-                    forward_weight = forward_weight_vector[edge.fwd_segment_position];
+                    if (edge.forward_packed_geometry_id != SPECIAL_EDGEID)
+                    {
+                        std::vector<EdgeWeight> forward_weight_vector;
```
daniel-j-h: vector alloc/dealloc here and below

src/engine/plugins/tile.cpp
```diff
@@ -224,272 +624,361 @@ Status TilePlugin::HandleRequest(const api::TileParameters &parameters, std::str
...
+                            const std::size_t duration,
+                            const DatasourceID datasource,
+                            const std::size_t name,
+                            std::int32_t &start_x,
+                            std::int32_t &start_y) {
+                            // Here, we save the two attributes for our feature: the speed and the
+                            // is_small
+                            // boolean.  We only serve up speeds from 0-139, so all we do is save
+                            // the
+                            // first
```
daniel-j-h: we're quite a few scopes deep here..

src/engine/plugins/tile.cpp
```diff
@@ -224,272 +624,361 @@ Status TilePlugin::HandleRequest(const api::TileParameters &parameters, std::str
...
+                                // index into the "values" array (also written later).  We're
+                                // not writing the actual speed or bool value here, we're saving
+                                // an index into the "values" array.  This means many features
+                                // can share the same value data, leading to smaller tiles.
+                                protozero::packed_field_uint32 field(
+                                    feature_writer, util::vector_tile::FEATURE_ATTRIBUTES_TAG);
+
+                                field.add_element(0); // "speed" tag key offset
+                                field.add_element(std::min(
+                                    speed_kmh, 127u)); // save the speed value, capped at 127
```
daniel-j-h: Why 127?

src/engine/plugins/tile.cpp
```diff
@@ -224,272 +624,361 @@ Status TilePlugin::HandleRequest(const api::TileParameters &parameters, std::str
...
+                                field.add_element(1);  // "is_small" tag key offset
+                                field.add_element(
+                                    128 + (edge.component.is_tiny ? 0 : 1)); // is_small feature
```
daniel-j-h: 128? Also below..

src/engine/plugins/tile.cpp
```diff
@@ -224,272 +624,361 @@ Status TilePlugin::HandleRequest(const api::TileParameters &parameters, std::str
...
+                        if (!tile_line.empty())
+                        {
+                            encode_tile_line(tile_line,
+                                             speed_kmh,
+                                             line_int_offsets[reverse_weight],
+                                             reverse_datasource,
+                                             name_offsets[name],
+                                             start_x,
+                                             start_y);
+                        }
```
daniel-j-h: Duplicate code with above

src/engine/plugins/tile.cpp
```diff
@@ -224,272 +624,361 @@ Status TilePlugin::HandleRequest(const api::TileParameters &parameters, std::str
...
+            protozero::pbf_writer point_layer_writer(tile_writer, util::vector_tile::LAYER_TAG);
+            point_layer_writer.add_uint32(util::vector_tile::VERSION_TAG, 2); // version
+            point_layer_writer.add_string(util::vector_tile::NAME_TAG, "turns"); // name
+            point_layer_writer.add_uint32(util::vector_tile::EXTENT_TAG,
+                                          util::vector_tile::EXTENT); // extent
+
+            // Begin writing the set of point features
+            {
+                // Start each features with an ID starting at 1
+                int id = 1;
```
daniel-j-h: But why?

include/engine/edge_unpacker.hpp
```diff
@@ -0,0 +1,104 @@
...
+ * the original route
+ * from beginning to end.
+ *
+ * @param packed_path_begin iterator pointing to the start of the NodeID list
+ * @param packed_path_end iterator pointing to the end of the NodeID list
+ * @param callback void(const std::pair<NodeID, NodeID>, const EdgeData &) called for each
+ * original edge found.
+ */
+
+template <typename DataFacadeT, typename RandomIter, typename Callback>
```
MoKob: Is this a naming question? `RandomIter` is after all only a template parameter.

TheMarex:
I'm missing some context here: Why was this merged despite outstanding comments from @daniel-j-h review? Would be great if we could capture that here.

daniel-j-h:
@TheMarex we're using the new Github review workflow https://github.com/blog/2256-a-whole-new-github-universe-announcing-new-tools-forums-and-features but instead of me formally "Accepting" @MoKob's changes via the Github UI we discussed the last finishing touches on chat. So the comments are still visible although @MoKob made changes to the code.

We should have kept the discussion in the tickets sorry about that.

PR #2403, opened 2016-05-14
Refactoring of iso_8601_grammar

oxidase:
* formats 'hh:mm:ss' and ISO 8601 are merged into one grammar.
* iso_8601_grammar is changed to static const.
* iso_8601_grammar supports formats P[n]DT[n]H[n]M[n]S, P[n]W,
  and part of alternative PTHHMMSS and extended PTHH:MM:SS formats.

The PR makes possible to parse duration tags like in http://www.openstreetmap.org/way/141294301


TheMarex:
Great. Two things because this adds a feature:

- [ ] Rebase on master
- [ ] Add a changelog entry that outlines the extended support for ISO8601

Did you do any ad-hoc benchmarks on this? I would guess the custom grammar is faster then reg-exp. This is not really a critical code path, so no worries here - I'm just curious.

oxidase:
Ok, rebased and added a line into ChangeLog. For the pure synthetic parsing test common_durations_get_translated the speed-up is 100 fold: from 512620ns to 4345ns. But the osrm-extract shows no statistically significant speed-up with p-value 0.81 for parsing of 1870996 ways.

PR #2405, opened 2016-05-14
Create pidfile once osrm-routed is ready to serve request.

zimathias:
Given that osrm-routed can take some time to load its data and be ready
to serve request create a pidfile only once requests will be served.
This is useful when hot reloading osrm-routed: a supervising process (eg
rainbow-saddle) can start a new osrm-routed process and watch for the
creation of the new pid file. Then the old osrm-routed can be shutdown.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -36,7 +36,7 @@ matrix:
       addons: &gcc5
         apt:
           sources: ['ubuntu-toolchain-r-test']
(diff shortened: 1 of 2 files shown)


danpat:
@zimathias For reloading data, you can use `osrm-datastore` and `osrm-routed -s`.

See https://github.com/Project-OSRM/osrm-backend/wiki/Configuring-and-using-Shared-Memory

You can load a new dataset with `osrm-datastore`, and `osrm-routed` will hot-swap to the new dataset, even under load (there is a small pause in serving requests, but it's just a few seconds, even for whole-planet sized datasets).

Issue #2296, opened 2016-04-22
Lane Guidance in OSRM


MoKob:
For some turns, the turn direction itself is not enough but we require additional information on the required lane.

This is the case if we have an ambiguous set of lanes at a turn, only one of which is actually valid for a complex maneuver.

Consider the following set of instructions: `turn left onto road A`, `after 20 meters turn right onto road B`.
If The left turn allows for a choice between two lane, only the right of the two will allow for a straightforward turn onto `road B`.

To allow lane guidance in OSRM, we will first implement a straightforward approach.

A turn maneuver and the required lines can be described as a `tuple` of `(first_lane_in,first_lane_out,#lanes)`. This creates a mapping, counting from right to left, from lanes on `road A` to lanes on `road B`.

The work on lane guidance is separated into subtasks:

### Preprocessing
 - [x] parse lane information in (http://wiki.openstreetmap.org/wiki/Key:turn:lanes)
    - [x] ways
    - [x]  ~~relations~~
    - [x] handle bus lanes
 - [x] create a list of possible tuples
      - [x] Implement a General Look-up Table to work with ShM and Internal Memory Facade
      - [x] add turn ID to turn-instructions
 - [x] associate every turn with its appropriate tuple

For the moment, we will not perform further memory optimisations for this task.

### Initial Exposure

For fast integration, in a first step we will expose ambiguous lane guidance, without any narrowing.

 - [x] define a good format for exposure @karenzshea any input?
 - [x] expose lane guidance
       - [x] expose non-narrowed guidance

### Postprocessing

In a second step, we can disambiguate lane information, wherever necessary.

 - [x] disambiguate turns in maneuvers (see starting example)
       - [x] separate route into logical segments
           - [x] segments with single lane
           - [x] longer segments
      - [x] implement algorithm for connecting lanes optimally within a segment
 - [x] expose narrowed lane information to the api

MoKob:
A problematic case:

Detection of Bus-Lanes in turn-lane data.
http://www.openstreetmap.org/#map=19/52.50632/13.28037, coming from the west on Masurenallee. The turn lanes are given as `none|through|through|left|left` where `none` refers to an unusable Bus-lane. These cases have to be addressed using schemes provided in OSM (http://wiki.openstreetmap.org/wiki/Bus_lanes)

MoKob:
What we need to discuss: How to expose turn-lanes for staying on a road.

[...]

PR #2410, opened 2016-05-16
[ready] Turn Lane Guidance

MoKob:
Implementing https://github.com/Project-OSRM/osrm-backend/issues/2296

- [x] extraction
    - [x] parse turn lane data during extraction
          - [x] turn:lane*
          - [x] destination:lanes*
    - [x] match turn lane string to possible turns
    - [x] create turn lane tupels (lanes at turn, lanes possible for instruction, first possible lane from the right)
    - [x] get turn lanes through data pipeline

- [x] routing
    - [x] expose available lanes to the API
    - [x] implement concept of independent turns
    - [x] implement post-processing to restrict multiple possible lanes to fewer lanes, when needed

- [x] cucumber
    - [x] expose lane guidance
    - [x] write lane guidance tests

- [x] docs/changelog
    - [x] document changes in doc/http.md
    - [x] update changelog


include/extractor/guidance/toolkit.hpp
```diff
@@ -470,6 +470,52 @@ inline ConnectedRoad mirror(ConnectedRoad road)
...
+        for (std::size_t i = 0; i < count_left; ++i)
+            // this is adjusted for our fake pipe. The moment cucumber can handle multiple escaped
+            // pipes, the '&' part can be removed
+            if (lane_string[i] != '|' && lane_string[i] != '&')
+            {
+                sane = false;
+                break;
+            }
+
+        if( sane )
```
daniel-j-h: clang-format38

include/util/guidance/turn_lanes.hpp
```diff
@@ -0,0 +1,95 @@
...
+} // namespace std
+
+namespace osrm
+{
+namespace util
+{
+namespace guidance
+{
+
+// The mapping of turn lanes can be done using a tupel as described here.
```
daniel-j-h: Keeping track of this so we don't forget about it before merging:

- [ ] remove outdated example or adapt to new representation

src/contractor/contractor.cpp
```diff
@@ -76,7 +76,7 @@ int Contractor::Run()
 #ifdef WIN32
 #pragma message("Memory consumption on Windows can be higher due to different bit packing")
 #else
-    static_assert(sizeof(extractor::NodeBasedEdge) == 20,
+    static_assert(sizeof(extractor::NodeBasedEdge) == 24,
                   "changing extractor::NodeBasedEdge type has influence on memory consumption!");
```
MoKob: windows size might be off

include/extractor/guidance/turn_instruction.hpp
```diff
@@ -146,16 +147,18 @@ struct TurnInstruction
     }
 };
 
-static_assert(sizeof(TurnInstruction) == 1, "TurnInstruction does not fit one byte");
+static_assert(sizeof(TurnInstruction) == 3, "TurnInstruction does not fit three byte");
```
MoKob: its currently a typo, but @daniel-j-h might have to modify this anyways. @daniel-j-h can you check this size when you have the switch and make sure to fix the typo when you are on it?

.clang-format
```diff
@@ -2,7 +2,7 @@
 Language:        Cpp
 # BasedOnStyle:  LLVM
 AccessModifierOffset: -2
-AlignAfterOpenBracket: Align
+AlignAfterOpenBracket: true
```
TheMarex: Where are these changes coming from? This should be at least an own commit.

docs/http.md
```diff
@@ -437,30 +437,42 @@ step.
...
+- `lanes`: the available turn lanes at the turn
+  - `marked`: markings on the road, following the OSM scheme (e.g. left, slight_right, or through)
+  - `take`: a boolean flag indicating whether the lane is a possible choice in the maneuver
 - `pronunciation`: The pronunciation hint of the way name. Will be `undefined` if there is no pronunciation hit.
 - `destinations`: The destinations of the way. Will be `undefined` if there are no destinations.
 - `mode`: A string signifying the mode of transportation.
 - `maneuver`: A `StepManeuver` object representing the maneuver.
 - `intersections`: A list of `Intersections` that are passed along the segment, the very first belonging to the StepManeuver
 
+Currently, the supported lane tags are: `sharp_left, left, slight_left, sharp_right, right, slight_right, through, reverse, none`.
```
TheMarex: This needs to be a table like the turn type with descriptions for each.

docs/http.md
```diff
@@ -437,30 +437,42 @@ step.
   | geojson    | [GeoJSON `LineString`](http://geojson.org/geojson-spec.html#linestring) or [GeoJSON `Point`](http://geojson.org/geojson-spec.html#point) if it is only one coordinate (not wrapped by a GeoJSON feature)|
   
 - `name`: The name of the way along which travel proceeds.
+- `lanes`: the available turn lanes at the turn
+  - `marked`: markings on the road, following the OSM scheme (e.g. left, slight_right, or through)
```
TheMarex: `marked` is a strange property name for this. If its marking on the road or a sign does not really matter, the interesting think is what type of lane this is. As such I would propose `type`.

docs/http.md
```diff
@@ -437,30 +437,42 @@ step.
   | geojson    | [GeoJSON `LineString`](http://geojson.org/geojson-spec.html#linestring) or [GeoJSON `Point`](http://geojson.org/geojson-spec.html#point) if it is only one coordinate (not wrapped by a GeoJSON feature)|
   
 - `name`: The name of the way along which travel proceeds.
+- `lanes`: the available turn lanes at the turn
```
TheMarex: Are these only valid at the turn point or for the whole segment? If it is only valid at the turn this need to be part of `StepManeuver`.

docs/http.md
```diff
@@ -437,30 +437,42 @@ step.
...
+  - `marked`: markings on the road, following the OSM scheme (e.g. left, slight_right, or through)
+  - `take`: a boolean flag indicating whether the lane is a possible choice in the maneuver
```
TheMarex: Should be named `valid`.

docs/http.md
```diff
@@ -437,30 +437,42 @@ step.
...
+Currently, the supported lane tags are: `sharp_left, left, slight_left, sharp_right, right, slight_right, through, reverse, none`.
+
 #### Example
 
 ```
 {
  "distance":152.3,
  "duration":15.6,
  "name":"Lortzingstraße",
+ "lanes":{
```
TheMarex: `lanes` is and Array and this needs to use `[]` braces.

include/extractor/guidance/debug.hpp
```diff
@@ -0,0 +1,46 @@
+#ifndef OSRM_EXTRACTOR_GUIDANCE_DEBUG_HPP_
```
TheMarex: Can we put all this debug logic in a central debug header (`util/debug.hpp`)? We don't need to be worried about a star topology with dependencies since this header should never be included in production code.

TheMarex:
I started a review on this PR. I already added some smaller suggestions. Will post back tomorrow with some more involved questions/concerns.

docs/http.md
MoKob: What kind of description would you expect? This would be: left - an arrow pointing to the left. Would a link to the OSM definition be acceptable? These are directly passed through from OSM.

.clang-format
MoKob: This fixes an error about an unsupported value for this choice. Without it clang-format reformats close to the entire code-base for me. Without, I only get a format of my own files.

Will open a dedicated PR for it.

MoKob: Hm. I was getting all kinds with YAML errors with this. No idea why I can't reproduce it right now. Removing

docs/http.md
MoKob: Well. Its really just the marking on the road. So `lane` does not really fit in my view. Would `marking` be more to your liking?

docs/http.md
MoKob: This is in fact a documentation mix-up. They are part of `StepManeuver` in the code already.

TheMarex:
Capturing from chat:

- [ ] Loading the original lane tag information as strings into the server needs to go
- [ ] The lane tags will be replaced by normalized enums and not just pass through OSM values
- [ ] Instead of using a `std::vector<char>` for storing lane tags this means it will be `std::vector<LaneType>`

docs/http.md
```diff
@@ -481,6 +487,13 @@ step.
   direction of travel immediately after the maneuver.
 - `type` A string indicating the type of maneuver. **new identifiers might be introduced without API change**
    Types  unknown to the client should be handled like the `turn` type, the existance of correct `modifier` values is guranteed.
+- `lanes`: the available turn lanes at the turn
+  - `indication`: markings on the road, following the OSM scheme (e.g. left, slight_right, or through). Markings mey be combined into a list separated by ';'.
+  - `valid`: a boolean flag indicating whether the lane is a valid choice in the maneuver
+
+Currently, the supported lane tags are: `sharp_left, left, slight_left, sharp_right, right, slight_right, through, reverse, none`.
```
TheMarex: We should make this consistent with our modifiers. We don't use underscore there. Suggestion:

| `direction`      | Description                               |
| ------------------- | ---------------------------------------|
| uturn             | lane used for uturn maneuver          |
| sharp right       | lane for a sharp right turn                        |
| right             | lane for  a normal turn to the right                |
| slight right      | lane for a slight turn to the right                |
| straight          | lane to go straight           |
| slight left       | lane for a slight turn to the left                 |
| left              | lane for a normal turn to the left                 |
| sharp left        | lane for a sharp turn to the left                  |
| all | lane allows turns in all direction |

We should also change `indication` -> `direction` or `type`.

TheMarex:
API response for a `RouteStep` will look like this if my current suggestions are implemented:

```
{
 "distance":152.3,
 "duration":15.6,
 "name":"Lortzingstraße",
 "maneuver":{
     "type":"turn",
     "modifier":"right",
     "lanes":[
         {"direction":"sharp left", "valid":"false"},
         {"direction":"straight", "valid":"true"},
         {"direction":"right", "valid":"true"}
     ]},
 "geometry":"{lu_IypwpAVrAvAdI",
 "mode":"driving",
 "intersections":[
    {"location":[13.39677,52.54366],
    "in":2,
    "out":1,
    "bearings":[10,184,270],
    "entry":[false,"true","true"]},
    {"location":[13.394718,52.543096],
    "in":0,
    "out":2,
    "bearings":[60,150,240,330],
    "entry":["false","true","true","true"]
    }
]}
```

docs/http.md
MoKob: In my current rewrite I already have the variation without underscore.
I feel that `all` is not a valid representation, because even if there is no marking (which would be all), it may only be allowed to go straight. The markings here represent the actual markings on the street.

The same way I would feel badly about direction. What we report here is nothing more than an indication, which is the current name. Direction would mean that it can be used only for a specific direction. A slight-right arrow can be used for exiting a round-about in two ways, for example.
We could go for type, but it would be confusing, since it is not a turn-road. So type would feel to me that it is a dedicated (separated) road. But as I said, indication is all we get.

docs/http.md
```diff
@@ -481,6 +485,10 @@ step.
   direction of travel immediately after the maneuver.
 - `type` A string indicating the type of maneuver. **new identifiers might be introduced without API change**
    Types  unknown to the client should be handled like the `turn` type, the existance of correct `modifier` values is guranteed.
+- `lanes`: the available turn lanes at the turn
+  - `indication`: markings on the road, following the OSM scheme (e.g. left, slight_right, or through). Markings mey be combined into a list separated by ';'.
```
TheMarex: > Markings mey be combined into a list separated by ';'.

I don't think this documentation is valid anymore. The Table needs to be moved up here.

docs/http.md
```diff
@@ -535,6 +543,20 @@ step.
   | else                   | Indicates the number of intersections passed until the turn. Example instruction: `at the fourth intersection, turn left` |
   
 
+- `indication`
+  
+  | `type`                 | Description                                                                                                               |
```
TheMarex: I think this table is broken. Member should be named `indication` to be consistent with your other namings.

include/engine/datafacade/shared_datafacade.hpp
```diff
@@ -79,18 +81,25 @@ class SharedDataFacade final : public BaseDataFacade
...
     util::ShM<NodeID, true>::vector m_via_node_list;
     util::ShM<unsigned, true>::vector m_name_ID_list;
+    util::ShM<LaneDataID, true>::vector m_lane_data_id;
     util::ShM<extractor::guidance::TurnInstruction, true>::vector m_turn_instruction_list;
     util::ShM<extractor::TravelMode, true>::vector m_travel_mode_list;
     util::ShM<char, true>::vector m_names_char_list;
+    util::ShM<char, true>::vector m_turn_string_char_list;
     util::ShM<unsigned, true>::vector m_name_begin_indices;
+    util::ShM<char, true>::vector m_lane_string_char_list;
+    util::ShM<unsigned, true>::vector m_lane_string_begin_indices;
```
TheMarex: These are left-over from the old code.

include/engine/datafacade/shared_datafacade.hpp
```diff
@@ -79,18 +81,25 @@ class SharedDataFacade final : public BaseDataFacade
     util::PackedVector<OSMNodeID, true> m_osmnodeid_list;
     util::ShM<NodeID, true>::vector m_via_node_list;
     util::ShM<unsigned, true>::vector m_name_ID_list;
+    util::ShM<LaneDataID, true>::vector m_lane_data_id;
     util::ShM<extractor::guidance::TurnInstruction, true>::vector m_turn_instruction_list;
     util::ShM<extractor::TravelMode, true>::vector m_travel_mode_list;
     util::ShM<char, true>::vector m_names_char_list;
+    util::ShM<char, true>::vector m_turn_string_char_list;
```
TheMarex: Also not used anymore.

include/engine/datafacade/internal_datafacade.hpp
```diff
@@ -78,14 +79,19 @@ class InternalDataFacade final : public BaseDataFacade
     util::ShM<NodeID, false>::vector m_via_node_list;
     util::ShM<unsigned, false>::vector m_name_ID_list;
     util::ShM<extractor::guidance::TurnInstruction, false>::vector m_turn_instruction_list;
+    util::ShM<LaneDataID, false>::vector m_lane_data_id;
+    util::ShM<util::guidance::LaneTupelIdPair, false>::vector m_lane_tupel_id_pairs;
     util::ShM<extractor::TravelMode, false>::vector m_travel_mode_list;
     util::ShM<char, false>::vector m_names_char_list;
+    util::ShM<char, false>::vector m_lanes_char_list;
```
TheMarex: Not used anymore.

include/extractor/guidance/turn_lane_types.hpp
```diff
@@ -0,0 +1,90 @@
...
+namespace TurnLaneType
+{
+namespace detail
+{
+const constexpr std::size_t num_supported_lane_types = 11;
+} // namespace detail
+
+typedef std::uint16_t Mask;
+const constexpr Mask empty = 0;
+const constexpr Mask none = 1 << 0;
```
TheMarex: When using bitshift operators it is good to use the unsigned post-fixes for numbers to make sure we don't accidentally introduce undefined behavior by shifting into the sign bit. Like: `1u << 0u`

include/extractor/guidance/toolkit.hpp
```diff
@@ -476,6 +483,77 @@ inline bool hasRoundaboutType(const TurnInstruction instruction)
...
+// are not specifically marked for left/right turns. This function can be used from the profile to
+// trim the lane string appropriately
+//
+// left|throught|
+// in combination with lanes:psv:forward=1
+// will be corrected to left|throught, since the final lane is not drivable.
+// This is in contrast to a situation with lanes:psv:forward=0 (or not set) where left|through|
+// represents left|through|through
+inline std::string
+trimLaneString(std::string lane_string, std::int32_t count_left, std::int32_t count_right)
```
TheMarex: Is this function unit tested? This seems like a good candidate for some hardening because it handles user input.

include/extractor/guidance/turn_lane_matcher.hpp
```diff
@@ -0,0 +1,59 @@
...
+// the best possible match for every tag in a given intersection, considering a few corner cases
+// introduced to OSRM handling u-turns
+typename Intersection::const_iterator findBestMatch(const TurnLaneType::Mask &tag,
+                                                    const Intersection &intersection);
+// Reverse is a special case, because it requires access to the leftmost tag. It has its own
+// matching function as a result of that. The leftmost tag is required, since u-turns are disabled
+// by default in OSRM. Therefor we cannot check whether a turn is allowed, since it could be
+// possible that it is forbidden. In addition, the best u-turn angle does not necessarily represent
+// the u-turn, since it could be a sharp-left turn instead on a road with a middle island.
+typename Intersection::const_iterator findBestMatchForReverse(const TurnLaneType::Mask &leftmost_tag,
```
TheMarex: Minor nit-pick I would put these (good!) documentation where the code lives not in the definition header.

include/storage/shared_datatype.hpp
```diff
@@ -113,11 +151,13 @@ struct SharedDataLayout
             bool end_canary_alive = std::equal(CANARY, CANARY + sizeof(CANARY), end_canary_ptr);
             if (!start_canary_alive)
             {
-                throw util::exception("Start canary of block corrupted.");
+                throw util::exception(std::string("Start canary of block corrupted. (") +
```
TheMarex: Hah good change. :+1:

include/storage/shared_datatype.hpp
```diff
@@ -49,6 +83,10 @@ struct SharedDataLayout
         BEARING_BLOCKS,
         BEARING_VALUES,
         ENTRY_CLASS,
+        LANE_DATA_ID,
+        TURN_LANE_DATA,
+        LANE_DESCRIPTION_OFFSETS,
+        LANE_DESCRIPTION_MASKS,
```
TheMarex: Add an static_assert here that check that `NUM_BLOCKS` and the size of `block_id_to_name` are the same. otherwise we might have errors in the error reporting.

include/util/debug.hpp
```diff
@@ -0,0 +1,88 @@
+#ifndef OSRM_ENGINE_GUIDANCE_DEBUG_HPP_
+#define OSRM_ENGINE_GUIDANCE_DEBUG_HPP_
+
```
TheMarex: Jup this is a good change. We have lots of other debug code (all the operator<< implementations for various types) that we could move in here.

include/util/group_by.hpp
```diff
@@ -0,0 +1,38 @@
...
+{
+
+// Runs fn on consecutive items in sub-ranges determined by pred.
+//
+// Example:
+//   vector<int> v{1,2,2,2,3,4,4};
+//   group_by(first, last, even, print);
+//   >>> 2,2,2
+//   >>> 4,4
+template <typename Iter, typename Pred, typename Fn>
```
TheMarex: Is there a unit-test for this?

src/engine/api/json_factory.cpp
```diff
@@ -56,17 +57,43 @@ const constexpr char *waypoint_type_names[] = {"invalid", "arrive", "depart"};
...
     return turn_type_names[static_cast<std::size_t>(type)];
 }
 
+util::json::Array lanesFromManeuver(const guidance::StepManeuver &maneuver)
+{
+    BOOST_ASSERT(maneuver.lanes.lanes_in_turn >= 1);
+    util::json::Array result;
+    LaneID lane_id = maneuver.lane_description.size();
+
+    for (auto iter = maneuver.lane_description.begin(); iter != maneuver.lane_description.end(); ++iter)
```
TheMarex: This can be a for-each loop now: `for(auto desc : maneuver.lane_description)`

src/engine/guidance/post_processing.cpp
```diff
@@ -31,16 +32,18 @@ namespace guidance
 
 namespace
 {
-const constexpr double MAX_COLLAPSE_DISTANCE = 25;
+const constexpr double MAX_COLLAPSE_DISTANCE = 30;
```
TheMarex: Why the constant changes here?

src/engine/guidance/post_processing.cpp
```diff
@@ -406,6 +408,7 @@ void collapseTurnAt(std::vector<RouteStep> &steps,
     else if (one_back_step.distance <= MAX_COLLAPSE_DISTANCE &&
              isCollapsableInstruction(current_step.maneuver.instruction))
     {
+        // TODO check for lanes
```
TheMarex: What is up with this TODO?

include/util/group_by.hpp
daniel-j-h: For the record, this is basically
https://docs.python.org/3.5/library/itertools.html#itertools.groupby

src/engine/api/json_factory.cpp
daniel-j-h: => `scripts/modernize.sh`

TheMarex:
@MoKob looks much better. :+1: Only remaining issues are outdated docs and some minor stuff.

include/util/group_by.hpp
TheMarex: Yeah we can add the tests in another PR, this is not blocking. But it would make sense to have tests for utility functions.

src/engine/guidance/post_processing.cpp
MoKob: This is related to https://github.com/Project-OSRM/osrm-backend/issues/2553

src/engine/guidance/post_processing.cpp
MoKob: Well. Thats the thing with arbitrary constants. Some time you find out that your first guess wasn't large enough.

include/extractor/guidance/toolkit.hpp
MoKob: since recently (@danpat and my late night debugging session) it has a dedicated cucumber test for its edge-cases

include/storage/shared_datatype.hpp
MoKob: done.

docs/http.md
```diff
@@ -481,6 +485,10 @@ step.
   direction of travel immediately after the maneuver.
 - `type` A string indicating the type of maneuver. **new identifiers might be introduced without API change**
    Types  unknown to the client should be handled like the `turn` type, the existance of correct `modifier` values is guranteed.
+- `lanes`: the available turn lanes at the turn
+  - `indication`: a indication (e.g. marking on the road) specifying the turn lane. Multiple indications may be combined with `;` (e.g. straight;left).
```
TheMarex: @MoKob is it still the case that ```multiple indications may be combined with `;````?

TheMarex:
There is just the one clarification I would like to have with regard to multiple indications. Otherwise we can merge this as soon as possible.

If there can be multiple indications per lane (why?), then we would need to change the API response format because of the same anti-pattern as before: Arrays encoded as strings in JSON.

So it would be:
```json
{
   "indications": ["left", "straight"],
   "valid": true,
}
```

MoKob:
The why is combination markers: arrow for going straight and going left.

TheMarex:
@mokob looks good now. We can merge once travis is green.

PR #2412, opened 2016-05-16
Make Traffic Updates Faster

daniel-j-h:
Parallelizes loading, parsing, saving. Switches speed segments to a flat map + binary search for initial creation and then lookups only use-case.

/cc @miccolis


daniel-j-h:
@danpat just :hand: to write a Boost.Spirit based CSV parser :P

daniel-j-h:
Rebase fsck up somehow https://codecov.io/gh/Project-OSRM/osrm-backend/commit/cf9ad9a7dede5b212bfe4949902e4fb033cc0154 is missing.

PR #2416, opened 2016-05-17
Signal the parent process when we're ready to serve requests.

zimathias:
This can be used for a hot reload of the process binary. The supervising
process can be told when the new process is ready to serve requests.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,7 @@
 
    - Infrastructure:
      - Better support for osrm-routed binary upgrade on the fly [UNIX specific]:
-       - Open sockets with SO_REUSEPORT to allow multiple servers connecting to the same port.
+       - Open sockets with SO_REUSEPORT to allow multiple osrm-routed processes serving requests from the same port.
        - Add SIGNAL_PARENT_WHEN_READY environment variable to enable osrm-routed signal its parent with USR1 when it's running and waiting for requests.
 
 # 5.1.0
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,9 @@
      - duration parser now accepts P[n]DT[n]H[n]M[n]S, P[n]W, PTHHMMSS and PTHH:MM:SS ISO8601 formats.
 
    - Infrastructure:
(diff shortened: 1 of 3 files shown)


zimathias:
#2397 is useful in this context.

danpat:
@zimathias Can you enumerate the advantage of this over the existing `osrm-datastore` and `osrm-routed -s` functionality?

Ref: https://github.com/Project-OSRM/osrm-backend/wiki/Configuring-and-using-Shared-Memory

zimathias:
The share memory option allows to hot swap the osrm *data*. However it doesn't help with hot swapping the osrm-routed *binary*. When a new osrm-routed binary is rolled for now we have to 1. stop the old binary and 2. start the new binary. #2397 allows for both the new and old binaries to listen and server requests on the same port. This PR allows a supervising process to be notified when the new process has started and is accepting requests.

The shared memory does speed up the startup process a lot - so that a new binary would listen in a few seconds.

danpat:
@zimathias Any reason not to use something like HAProxy for this?  It's almost certainly more flexible and covers more edge cases than `osrm-routed`.

zimathias:
@danpat a frontend proxy (such as haproxy or nginx) can definitely be used as well. There a multiple ways to implement the binary hot reload feature. It is very dependent on the infrastructure osrm-routed is running in. This approach works well for ours.

I'm fine if this PR is not accepted - we'll carry the patch in our code base. On the other hand it's a 3 line patch which is disabled by default. I'll leave it up to you.

danpat:
@zimathias I'm mostly just playing the devil's advocate.

If you can add docs for this to https://github.com/Project-OSRM/osrm-backend/blob/master/docs/http.md (please add a "Environent Variables" section at the top), and a test case, we can go ahead and merge.

TheMarex:
Merged this by hand into `master`: https://github.com/Project-OSRM/osrm-backend/commit/b5ffb51e1843a1665f3a75fef4efbf35de2dddea

@zimathias for future reference you should try to rebase your branches on `master`;

```
git checkout master
git pull
git checkout YOUR_BRANCH
git rebase -i master
git push -f
```

PR #2418, opened 2016-05-18
Expose node IDs

lbud:
This recreates https://github.com/Project-OSRM/osrm-backend/pull/2337 against current master to address #2054. 

Cribbing the TODO list from that PR:

- [x] Implement shared memory loading

  Based on preliminary checks, it does now work with shared memory. I was a bit flying blind on shared memory knowledge, though, so although it looked straightforward to me, I definitely could have done something wonky…

- [x] Investigate memory overhead on big datasets
- [x] Write tests
- [x] Write addition to API spec and discuss (API parameter? default?)

  Rather than adding node IDs to step objects, this now adds them to the optional `annotation` object introduced as an optional (off by default) API parameter recently.

- [ ] Write changelog entry
- [x] Integrate packed vector
- [x] Tests for packed vector
- [x] Re-investigate mem usage after packed vector to confirm

@TheMarex @danpat

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
      - paramater `annotate` was renamed to `annotations`.
      - `annotation` as accidentally placed in `Route` instead of `RouteLeg`
      - Support for destination signs. New member `destinations` in `RouteStep`, based on `destination` and `destination:ref`.
+     - Add `nodes` property to `annotation` in `RouteLeg` containing the ids of nodes covered by the route
 
    - Profile changes:
      - `result.destinations` allows you to set a way's destinations
--- a/docs/http.md
+++ b/docs/http.md
@@ -413,7 +413,8 @@ With `steps=false` and `annotations=true`:
   "steps": []
   "annotation": {
     "distance": [5,5,10,5,5],
-    "duration": [15,15,40,15,15]
+    "duration": [15,15,40,15,15],
+    "nodes": [49772551,49772552,49786799,49786800,49786801,49786802]
   }
 }
 ```
--- a/features/step_definitions/matching.js
+++ b/features/step_definitions/matching.js
@@ -34,7 +34,8 @@ module.exports = function () {
                         turns = '',
                         route = '',
                         duration = '',
-                        annotation = '';
+                        annotation = '',
+                        OSMIDs = '';
 
 
                     if (res.statusCode === 200) {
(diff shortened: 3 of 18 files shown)


danpat:
@TheMarex do we bump the semver if the shared memory layout changes?  Bumping processing from 5.2.0 to 5.2.1, but running against `osrm-routed -s` from 5.2.0 would break.

TheMarex:
@danpat I would count this similar to file format changes, as it requires a minor release (5.2 -> 5.3)

lbud:
> Investigate memory overhead on big datasets

@TheMarex @danpat …

lbud:
After measuring a 12% memory increase I've added a sketch of a packed vector in https://github.com/Project-OSRM/osrm-backend/pull/2418/commits/90de9fea72633ff63739cf07296bac82a1891d8e. TODO: refine, test, integrate — updated OP list to reflect as much.

lbud:
Re-investigated memory usage after adding the packed vector and it is now an increase (as expected) of only about 6% 🎉 

@daniel-j-h would you mind taking a look at this? 👀 🙇

daniel-j-h:
Sure, looking at [packed_vector.hpp](https://github.com/Project-OSRM/osrm-backend/blob/dfe2df1f08960d6defc12729256d8ab6e123fa34/include/util/packed_vector.hpp):

Do we want to hardcode the storage or can the packed vector work with any vector-like storage? For example, a [std::stack](http://en.cppreference.com/w/cpp/container/stack) is templated to store `T`s and to use a generic `Container` as underlying storage. I think we can do the same here, completely decoupling the packed vector from any OSRM types.

The constants at the top should be inside the packed vector (same with the helper functions), and if only the implementation and not the users need those they should live in a `private:` section.

`PackedVector() = default;` is not needed here, check the first row on page 28 here:
http://accu.org/content/conf2014/Howard_Hinnant_Accu_2014.pdf

You should probably provide a `T at(Index) const;` otherwise users are not able to call `.at` on a `const packed_vector v`.
 
You should overload `[]` so users can do `packed_vector v; use(v[index]);`:
http://en.cppreference.com/w/cpp/language/operators#Array_subscript_operator

lbud:
> You should overload `[]` so users can do `packed_vector v; use(v[index]);`:
http://en.cppreference.com/w/cpp/language/operators#Array_subscript_operator

@TheMarex and I talked about this briefly but because `[]` returns a reference to an element (right?) it seemed we couldn't really do that here, as storage elements aren't equivalent to their stored numbers…

danpat:
@lbud Only if you declare it as `value_t& operator[](std::size_t idx)`.  If you get rid of the `&`, then it can return a copy.

lbud:
@daniel-j-h @TheMarex @danpat this is rebased on master and I think ready to go — though, @danpat, did you want to do more perf testing before merging?

(To address @daniel-j-h's above comments, we discussed in person and some points were moot (helper functions are used outside of the class as well), and further templating can be done if needed down the road, I think…)

🍏 ?

TheMarex:
@lbud as soon a the tests are passing this should be good to merge.

daniel-j-h:
Currently failing long before the tests in the APT phase:

> The command "sudo -E apt-add-repository -y "ppa:ubuntu-toolchain-r/test"" failed and exited with 1 during .

Looks like a glitch on Travis/ppas, I rebased onto master and force pushed.

PR #2419, opened 2016-05-18
Fix #2413 by making enum-base explicitly unsigned char

### LINKED ISSUE #2413: Routing with Steps Crashes OSRM Server
Entering URL: "http://127.0.0.1:5000/route/v1/driving/103.8960784,1.3676869;103.880996,1.375114?overview=full&steps=true" will cause the server to crash.
Using URL:"http://127.0.0.1:5000/route/v1/driving/103.8960784,1.3676869;103.880996,1.375114?overview=full&steps=false" is fine though

A "osrm-routed.exe has stopped working" warning box will appear.

What is going on?
What solutions is there?

oxidase:
call stack at the crash point
```
osrm-routed.exe!osrm::engine::api::json::detail::instructionModifierToString(osrm::extractor::guidance::DirectionModifier modifier) Line 80    C++
     osrm-routed.exe!osrm::engine::api::json::makeStepManeuver(const osrm::engine::guidance::StepManeuver & maneuver) Line 158    C++
     osrm-routed.exe!osrm::engine::api::json::makeRouteStep(osrm::engine::guidance::RouteStep step, mapbox::util::variant<osrm::util::json::String,osrm::util::json::Number,mapbox::util::recursive_wrapper<osrm::util::json::Object>,mapbox::util::recursive_wrapper<osrm::util::json::Array>,osrm::util::json::True,osrm::util::json::False,osrm::util::json::Null> geometry) Line 185    C++
     osrm-routed.exe!osrm::engine::api::json::makeRouteLegs::__l6::<lambda>(osrm::engine::guidance::RouteStep step) Line 237    C++
...
```

where maneuver.instruction.direction_modifier = -2, but must be maneuver.instruction.direction_modifier = Left (6 '\x6')

It happens because of  enum-base type (that is 7.2/5 "if not explicitly specified, the underlying type of a scoped enumeration type is int.") sign extension. MSVC implicitly extends 1xx to 11..11xx and than at the crash point the sign-extended value is casted to size_t and used as an index.

### DIFF
--- a/include/engine/api/json_factory.hpp
+++ b/include/engine/api/json_factory.hpp
@@ -33,8 +33,8 @@ namespace json
 namespace detail
 {
 
-std::string instructionTypeToString(extractor::guidance::TurnType type);
-std::string instructionModifierToString(extractor::guidance::DirectionModifier modifier);
+std::string instructionTypeToString(extractor::guidance::TurnType::Enum type);
+std::string instructionModifierToString(extractor::guidance::DirectionModifier::Enum modifier);
 
 util::json::Array coordinateToLonLat(const util::Coordinate coordinate);
 
--- a/include/engine/guidance/toolkit.hpp
+++ b/include/engine/guidance/toolkit.hpp
@@ -49,7 +49,7 @@ inline bool staysOnRoundabout(const extractor::guidance::TurnInstruction instruc
     return instruction.type == extractor::guidance::TurnType::StayOnRoundabout;
 }
 
-inline extractor::guidance::DirectionModifier angleToDirectionModifier(const double bearing)
+inline extractor::guidance::DirectionModifier::Enum angleToDirectionModifier(const double bearing)
 {
     if (bearing < 135)
     {
--- a/include/extractor/guidance/intersection_handler.hpp
+++ b/include/extractor/guidance/intersection_handler.hpp
@@ -49,7 +49,7 @@ class IntersectionHandler
     std::size_t countValid(const Intersection &intersection) const;
 
     // Decide on a basic turn types
-    TurnType findBasicTurnType(const EdgeID via_edge, const ConnectedRoad &candidate) const;
+    TurnType::Enum findBasicTurnType(const EdgeID via_edge, const ConnectedRoad &candidate) const;
 
     // Get the Instruction for an obvious turn
     TurnInstruction getInstructionForObvious(const std::size_t number_of_candidates,
--- a/include/extractor/guidance/toolkit.hpp
+++ b/include/extractor/guidance/toolkit.hpp
@@ -145,13 +145,13 @@ getRepresentativeCoordinate(const NodeID from_node,
 }
 
 // shift an instruction around the degree circle in CCW order
-inline DirectionModifier forcedShiftCCW(const DirectionModifier modifier)
+inline DirectionModifier::Enum forcedShiftCCW(const DirectionModifier::Enum modifier)
 {
-    return static_cast<DirectionModifier>((static_cast<std::uint32_t>(modifier) + 1) %
-                                          detail::num_direction_modifiers);
+    return static_cast<DirectionModifier::Enum>((static_cast<std::uint32_t>(modifier) + 1) %
+                                                detail::num_direction_modifiers);
 }
 
-inline DirectionModifier shiftCCW(const DirectionModifier modifier)
+inline DirectionModifier::Enum shiftCCW(const DirectionModifier::Enum modifier)
 {
     if (detail::shiftable_ccw[static_cast<int>(modifier)])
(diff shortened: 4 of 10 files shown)


TheMarex:
Great work - looks good to me to merge. Could you rebase on `master`?

oxidase:
Hi, please don't merge for a moment. I have just found gcc warnings due to 
http://stackoverflow.com/questions/22606392/is-this-warning-related-to-enum-class-size-wrong
here is the example http://coliru.stacked-crooked.com/a/5e14cd3ebd852809
Also without enum-base type the size of TurnInstruction is 4.
I will try to find where exactly MSVC extends sign.

oxidase:
Hi, i would like to propose drop enum bit packing. In the following table
values of sizeof(TurnInstruction) and outcomes for MSVC and gcc are presented:

|                     | no enum-base     | enum-base uint8_t |
|---------------------|------------------|------------------|
|   with bit-packing  MSVC|   4, crash due to 7.2/7  |  1, ok       |
| with bit-packing gcc | 4, ok      |  1, warning due to 7.2/8 |
|                     |                  |                  |
|   w/o bit-packing  MSVC |  8, ok      | 2, ok       |
| w/o bit-packing gcc | 8, ok      | 2, ok       |

So without an enum-base type but with bit packing gcc requires not 1 but 4 bytes.
For enum-base uni8_t it will be just 2 bytes, bit-size covered by standard
and compiler-independent.

TheMarex:
@oxidase is there a way to tell gcc "we know what are doing, this fits into 3 bits"? This is definitely a regression and could explain a bump in memory usage that we have observed.

While we are on it, I think we should probably add a static assert there that checks for the size of `TurnInstruction`.

We need to have this as small as possible since we store a value for each edge in the graph - that means in the order of ~2^29 for the whole planet. There is a big difference between 1 byte and 4 bytes here - the main reason why we even bother with all this bit packing magic.

oxidase:
@TheMarex both gcc and msvc pack enums with 5 and 3 bits into 'unsigned char', but it violates c++ standard in "For an enumeration whose underlying type is fixed, the values of the enumeration are the values of the underlying type."  It means that an enum with uint8_t base type must be able to hold values 0-255 even if enum values are not defined for the whole range.

So "we know what are doing" -> disable or ignore the warning and use bit fields. But i fill uncomfortable by ignoring warnings because it could be platform- and compiler-dependent. Also the enums  are almost full and by adding new values it will easily overflow 1 byte.

oxidase:
@TheMarex possible solution to have 1 packed byte could be to avoid using enums with the range checks, define DirectionModifier and TurnType as uint8_t types and make enum values as constants with defined numerical values. 

If it is ok as a solution, i can update the commit.

TheMarex:
> If it is ok as a solution, i can update the commit.

Yes, sadly seems like we can't rely on enums here. We are doing similar things for `TravelMode`. That is unfortunate but fitting this into 1 byte is rather important. Not sure if @MoKob lane guidance work will push us over the boundary here beyond 5 bits for the turn type, but from what I see in his PR we should be good here still.

Thanks for digging here. This is a obnoxious bug.

oxidase:
@TheMarex ok, i have turned to explicit uint8_t type because it is already adopted solution for exactly the same issue. I also placed constants into  namespaces and named the underlying type `Enum`. The only disadvantage now is in typing TurnType::Enum instead of TurnType. 

Alone the change to 1 byte brings nothing, because of another issue with packing, because
```
Class osrm::extractor::OriginalEdgeData
   size=16 align=4
   base size=13 base align=4
type = struct osrm::extractor::OriginalEdgeData {
    NodeID via_node;
    unsigned int name_id;
    osrm::extractor::guidance::TurnInstruction turn_instruction;
    EntryClassID entry_classid;
    osrm::extractor::TravelMode travel_mode;
  public:
    OriginalEdgeData(NodeID, unsigned int, osrm::extractor::guidance::TurnInstruction, EntryClassID, osrm::extractor::TravelMode);
    OriginalEdgeData(void);
}
```

The problem is in EntryClassID alignment: the footprint is 
```
| NodeID 4 bytes | unsigned int 4 bytes | TurnInstruction 1 byte | 1 padding byte | EntryClassID 2 bytes | TravelMode 1 byte |
```
I have flipped TurnInstruction and EntryClassID, so now it is 12 bytes and memory savings are at least 3*#edge ids bytes in .osrm.edges file.

I also suggest to move TravelMode constants into a separate namespace in another PR to avoid namespace pollution.

TheMarex:
@oxidase this looks great. If you could rebase that would be great. Looks good to merge to me.

:+1: for finding the alignment problem.

oxidase:
ok, rebased

TheMarex:
Had to merge by hand because GH is acting up again.

PR #2420, opened 2016-05-19
Bring node-osrm bindings into main project

danpat:
At Mapbox, we primarily use OSRM via the `node-osrm` bindings.  These currently live in a separate project, but are tagged and released in lockstep with `osrm-backend`.

This PR brings the `node-osrm` codebase into `osrm-backend`, and builds it at the same time as the rest of the code.  This should reduce the friction we experience internally when needing to make `osrm-backend` features quickly available for NodeJS-based projects for folks not familiar with building C++ projects.

The only extra tools required are CMake 3.1, and `npm install node-cmake`.

TODO:
  - [x] make it compile and link against the current build (DONE: `cmake -DNODEJS=ON ..`, off by default)
  - [x] import test suite and make it work (DONE: `npm run node-test`)
  - [ ] enable static linking of dependencies for hassle-free `npm install` support on various platforms
  - [ ] support easy publishing of binaries/build-from-source support (via https://github.com/mapbox/node-pre-gyp/issues/205)
  - [ ] CMake 3.1 will be required - update Travis build and dependency documentation.

@TheMarex please don't hate me.


daniel-j-h:
Even Ubuntu 14.04 (LTS) does not ship with CMake 3.1+, so it's pretty bad wrt. building only `osrm-backend` or `libosrm` on different distributions. Could we at least decouple the root `CMakeLists.txt` from the Node.js binding's one only having the latter require CMake 3.1?

danpat:
@daniel-j-h yeah, that's not a bad idea.  Our `node-osrm` Travis build already pulls down `cmake 3.2.x`, so I'm OK requiring folks to get a newer cmake to get Node stuff.

TheMarex:
Closing in favor of tracking in an issue since the code here is stale anyway. https://github.com/Project-OSRM/osrm-backend/issues/3183

PR #2436, opened 2016-05-20
Support loading dynamic libraries from custom locations on OS X

springmeyer:
This builds on #1965 which did not cover all cases where that fix is needed. This fixes the command `make -C test/data benchmark` to work on OS X when:

  - libtbb.dylib is on a custom path (not /usr/lib or /usr/local/lib) and
  - the rpath is not embedded inside the tools (with `install_name_tool`)

Because the above workarounds are not always possible or easy.

### DIFF
--- a/features/support/launch_classes.js
+++ b/features/support/launch_classes.js
@@ -86,7 +86,7 @@ var OSRMDirectLoader = class extends OSRMBaseLoader {
             fs.appendFile(this.scope.OSRM_ROUTED_LOG_FILE, data, (err) => { if (err) throw err; });
         };
 
-        var child = spawn(util.format('%s%s/osrm-routed', this.scope.LOAD_LIBRARIES, this.scope.BIN_PATH), [this.inputFile, util.format('-p%d', this.scope.OSRM_PORT)]);
+        var child = spawn(util.format('%s/osrm-routed', this.scope.BIN_PATH), [this.inputFile, util.format('-p%d', this.scope.OSRM_PORT)]);
         this.scope.pid = child.pid;
         child.stdout.on('data', writeToLog);
         child.stderr.on('data', writeToLog);
@@ -122,7 +122,7 @@ var OSRMDatastoreLoader = class extends OSRMBaseLoader {
             fs.appendFile(this.scope.OSRM_ROUTED_LOG_FILE, data, (err) => { if (err) throw err; });
         };
 
-        var child = spawn(util.format('%s%s/osrm-routed', this.scope.LOAD_LIBRARIES, this.scope.BIN_PATH), ['--shared-memory=1', util.format('-p%d', this.scope.OSRM_PORT)]);
+        var child = spawn(util.format('%s/osrm-routed', this.scope.BIN_PATH), ['--shared-memory=1', util.format('-p%d', this.scope.OSRM_PORT)]);
         this.child = child;
         this.scope.pid = child.pid;
         child.stdout.on('data', writeToLog);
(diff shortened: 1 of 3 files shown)


springmeyer:
Note: as a followup I will be working on a PR that starts testing `libtbb.dylib` in a custom location so we'll see if this regresses again.

TheMarex:
@springmeyer this seems to have broken the cucumber tests.

springmeyer:
@TheMarex yes, forwarding the `osrm-routed <args>` along right here: https://github.com/Project-OSRM/osrm-backend/pull/2436/commits/fd4880528c41d22fa9eeeb876aa19e0dab2857be#diff-f5e5cb2942b044b80ea4df90379e273dR16 should not have changed anything but I think the args must end up being interpreted differently. Not sure why yet.

springmeyer:
Turns out the problem (now fixed) was broken `pid` tracking. I tried calling a bash script instead to be able to set `DYLD_LIBRARY_PATH` on the fly, which worked except it broke the pid tracking.

springmeyer:
@TheMarex this is rebased and ready to merge. To recap:

 - We use `spawn` for osrm-routed to track the `pid`. It is not possible therefore to pass `DYLD_LIBRARY_PATH=<val>` as the first argument to spawn, since it expects the actual binary. The solution then is to not pass it and just ensure `osrm-routed` can be run without needed `DYLD_LIBRARY_PATH` set. In an upcoming PR to support mason-packaged libraries I will work around this by using `install_name_tool` to set the right path
 - The remaining fixes in the PR address the ability to pass `DYLD_LIBRARY_PATH` to all the other command line tools. This greatly eases the ability to use shared libraries at custom paths without having to use `install_name_tool` to set the right path in for all the command line tools.

springmeyer:
Per chat with @TheMarex - this is on hold. Going to try an alternative strategy of using `rpath` to avoid needing to depend on `DYLD_LIBRARY_PATH` on OS X.

springmeyer:
> Going to try an alternative strategy of using rpath to avoid needing to depend on DYLD_LIBRARY_PATH on OS X.

Got this working. And I'll plan to go with this approach. So, instead of needing test changes like above, we'll just eventually need some scripts to run on travis to set rpath correctly when building against dynamic libraries in custom places on OS X (incoming PR).

It is going to look something like this:

```sh
cp ${MASON_HOME}/lib/libtbb* ./build/
for tool in $(ls build/osrm-*); do
  install_name_tool -change libtbb.dylib @loader_path/libtbb.dylib ${tool}
  install_name_tool -change libtbbmalloc.dylib @loader_path/libtbbmalloc.dylib ${tool}
done

for tool in $(ls build/libosrm*.dylib); do
  install_name_tool -change libtbb.dylib @loader_path/libtbb.dylib ${tool}
  install_name_tool -change libtbbmalloc.dylib @loader_path/libtbbmalloc.dylib ${tool}
done

for tool in $(ls build/unit_tests/*-tests); do
  install_name_tool -change libtbb.dylib @loader_path/../libtbb.dylib ${tool}
  install_name_tool -change libtbbmalloc.dylib @loader_path/../libtbbmalloc.dylib ${tool}
done

if [[ -f build/libosrm.dylib ]]; then
  install_name_tool -change libtbb.dylib @loader_path/libtbb.dylib build/libosrm.dylib
  install_name_tool -change libtbbmalloc.dylib @loader_path/libtbbmalloc.dylib build/libosrm.dylib
fi

install_name_tool -change libosrm.dylib @loader_path/../../build/libosrm.dylib example/build/osrm-example
install_name_tool -change libtbb.dylib @loader_path/../../build/libtbb.dylib example/build/osrm-example
install_name_tool -change libtbbmalloc.dylib @loader_path/../../build/libtbbmalloc.dylib example/build/osrm-example

```

springmeyer:
-> #2478

PR #2437, opened 2016-05-20
Resurrect Flat Map + Binary Search commit from PR 2412

daniel-j-h:
For https://github.com/Project-OSRM/osrm-backend/pull/2412#issuecomment-220657034. /cc @TheMarex


TheMarex:
Okay I promise to not make this go :boom: again.

Issue #2434, opened 2016-05-20
Log number of loaded speed segments


freenerd:
We are currently showing that a segment speed file has been loaded like below:

```
[info] Segment speed data supplied, will update edge weights from /tmp/segments.csv
```

It would be great to show the number of lines that matched, as a double check that it actually matched segments correctly, like this:

```
[info] Segment speed data updated 102948 edge weights from /tmp/segments.csv
```

/cc @willwhite @danpat

PR #2440, opened 2016-05-20
Contractor: log stats for loaded speed files

freenerd:
Fixes https://github.com/Project-OSRM/osrm-backend/issues/2434

Output looks like this:

```
[info] Opening /tmp/extract.osrm.ebg
[info] Reading 702 edges from the edge based graph
[info] Loaded speed file /tmp/speeds.csv with 2 speeds
[info] In total loaded 1 speed file(s) with a total of 2 unique values
[info] Used 3382 speeds from LUA profile or input map
[info] Used 2 speeds from /tmp/speeds.csv
```

Todo:

- [x] Figure out what the high number in speeds from LUA profile means (one way streets are a thing)
- [ ] Code review

### DIFF
--- a/src/contractor/contractor.cpp
+++ b/src/contractor/contractor.cpp
@@ -32,6 +32,7 @@
 #include <tbb/spin_mutex.h>
 
 #include <algorithm>
+#include <atomic>
 #include <bitset>
 #include <cstdint>
 #include <fstream>
@@ -244,6 +245,9 @@ parse_segment_lookup_from_csv_files(const std::vector<std::string> &segment_spee
             local.push_back(std::move(val));
         }
 
+        util::SimpleLogger().Write() << "Loaded speed file " << filename << " with " << local.size()
+                                     << " speeds";
+
         {
             Mutex::scoped_lock _{flatten_mutex};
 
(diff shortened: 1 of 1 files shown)


TheMarex:
@freenerd build is failing on travis. Seems like you forgot to include `atomic` somewhere.

daniel-j-h:
For the record: https://gcc.gnu.org/ml/libstdc++/2015-01/msg00121.html
We need to workaround GCC 4.8's `libstdc++` not defining `atomic_init`.

daniel-j-h:
@freenerd if you want to use relaxed memory orders:
https://github.com/Project-OSRM/osrm-backend/pull/2440/commits/08f3a0b70df666a1bd10e499c4738a21760a0139#diff-5d25f965a92c5af3a4dfd03cf4f7616aR501
then it would make sense to `fetch_add` for incrementing the counters with relaxed instead of `+= 1`:
http://en.cppreference.com/w/cpp/atomic/atomic/fetch_add

freenerd:
@daniel-j-h using `std::memory_order_seq_cst` which should work okay?

daniel-j-h:
`store(0);` for initialization is good enough, as `store` has sequential consistency as default second argument already: http://en.cppreference.com/w/cpp/atomic/atomic/store

Issue #2426, opened 2016-05-19
More continue instructions on master


TheMarex:
Using this route `http://127.0.0.1:5000/route/v1/driving/13.411474227905273,52.52786686274604;13.424413204193115,52.52342816902858?overview=false&alternatives=true&steps=true` on `master` yields two `continue` instructions that I currently don't see on the demo server yet.

Old:

```
{
    "code": "Ok",
    "routes": [
        {
            "legs": [
                {
                    "steps": [
                        {
                            "geometry": "cjr_IylzpAcAgA",
                            "maneuver": {
                                "bearing_after": 33,
                                "location": [
                                    13.411491,
                                    52.527859
                                ],
                                "bearing_before": 0,
                                "type": "depart"
                            },
                            "duration": 9.2,
                            "distance": 45.1,
                            "name": "Zolastra\u00dfe",
                            "mode": "driving"
                        },
                        {
                            "geometry": "glr_IaozpARgAfBwJF]DOh@yAHSf@aADGR_@",
                            "maneuver": {
                                "bearing_after": 114,
                                "type": "end of road",
                                "modifier": "right",
                                "bearing_before": 33,
                                "location": [
                                    13.411851,
                                    52.5282
                                ]
                            },
                            "duration": 22,
                            "distance": 282.4,
                            "name": "Torstra\u00dfe",
                            "mode": "driving"
                        },
                        {
                            "geometry": "gdr_Ige{pA\\s@N[BENa@Zw@`@oA^gABIFUH[^sAJ_@Nq@Ps@DQRq@Ng@^sADKHYb@kALa@@CT{@Ri@Z_AJ[Lk@\\}Al@sCp@uCH_@He@H]BKJg@TkA",
                            "maneuver": {
                                "exit": 7,
                                "bearing_after": 135,
                                "type": "new name",
                                "modifier": "straight",
                                "bearing_before": 136,
                                "location": [
                                    13.415404,
                                    52.526921
                                ]
                            },
                            "duration": 57,
                            "distance": 725,
                            "name": "Mollstra\u00dfe",
                            "mode": "driving"
                        },
                        {
                            "geometry": "snq_Iw}|pA",
                            "maneuver": {
                                "bearing_after": 0,
                                "location": [
                                    13.424439,
                                    52.523461
                                ],
                                "bearing_before": 116,
                                "type": "arrive"
                            },
                            "duration": 0,
                            "distance": 0,
                            "name": "Mollstra\u00dfe",
[...]

PR #2444, opened 2016-05-23
fix continue on obvious

MoKob:
Addressing https://github.com/Project-OSRM/osrm-backend/issues/2426.

Handles continue instructions on roads with a lot of new-names better.

The problem here was, that we cannot see an straight turn as obvious, if the road itself takes a turn.
The discovery for this was off, for cases where the name is repeated a lot.


TheMarex:
Rebased and merging. :+1:

PR #2451, opened 2016-05-23
[Travis] Disable spotlight indexing on osx to speed up 'npm test'

springmeyer:
I noticed that OS X spotlight indexing is on by default on travis. And due to the large number of files that `npm test` creates it keeps spotlight indexing quite busy. This leads to more CPU being used by spotlight that by osrm-backend during the tests.

This PR disables spotlight indexing. This speeds up the `npm test` on OSX Debug builds from 13 minutes to 9 minutes.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -112,6 +112,7 @@ matrix:
       #- env: CCOMPILER='clang-3.8' CXXCOMPILER='clang++-3.8' BUILD_TYPE='Release' BUILD_SHARED_LIBS=ON
 
 before_install:
+  - if [[ $(uname -s) == 'Darwin' ]]; then sudo mdutil -i off /; fi;
   - source ./scripts/install_node.sh 4
   - npm install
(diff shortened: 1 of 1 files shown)


TheMarex:
Great find!

springmeyer:
More notes on this: I've tried to figure out a way to programmatically disable for just the specific ./test/cache directory such that we could recommend all users do this on OS X. This can be done like:

```
VOLUMNE=<set to the name of your harddrive>
sudo defaults write  /Volumes/${VOLUME}/.Spotlight-V100/VolumeConfiguration.plist Exclusions -array-add $(pwd)/test/cache
```

However I still see mds_store showing up with very high CPU in Activity Monitor when running `npm test`, so I'm not sure if this really works as I would expect. Same result (not seeming to reduce mds cpu) with the `.noindex` and `.metadata_never_index` methods mentioned at http://apple.stackexchange.com/questions/87090/how-to-programatically-add-folder-to-spotlight-do-not-index

PR #2453, opened 2016-05-24
[Travis] Install cmake from more reliable s3 endpoint

springmeyer:
We upgrade cmake on travis to allow us to depend on newer cmake features in the `CMakeLists.txt`.

The existing method of upgrading `cmake` has been failing randomly and fairly frequently due to network outage. This moves to using a cmake binary from s3 published by [mason](https://github.com/mapbox/mason/tree/master/scripts/cmake/3.5.2) which should be less likely to fail.


TheMarex:
Looks good. :+1:

PR #2464, opened 2016-05-25
Distinguish between offramps and sliproads.

danpat:
This change converts some `take the ramp` instructions into simpler `turn` instructions.

In situations where a major road has a `_link` to another road, *but* those two major roads are also directly connected (i.e. there's no overpass), then we will annouce a simple turn, rather than a 'take the ramp'.  An example of this situation is here:

http://www.openstreetmap.org/#map=20/51.08164/-113.98831&layers=H

### DIFF
--- a/src/engine/guidance/post_processing.cpp
+++ b/src/engine/guidance/post_processing.cpp
@@ -41,8 +41,8 @@ void print(const std::vector<RouteStep> &steps)
     int segment = 0;
     for (const auto &step : steps)
     {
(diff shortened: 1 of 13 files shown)


src/engine/guidance/post_processing.cpp
```diff
@@ -587,10 +606,41 @@ std::vector<RouteStep> collapseTurns(std::vector<RouteStep> steps)
         const auto two_back_index = getPreviousIndex(one_back_index);
         BOOST_ASSERT(two_back_index < steps.size());
 
+        print(steps);
```
MoKob: remove

Issue #1342, opened 2015-01-12
include way tags in instructions


emiltin:
in our "I Bike CPH" bike router app for ios and android, we provide GPS turn-by-turn navigation. we would like to use things like way type, surface, etc. to provide additional cues. this is relevant for bikes, since many bike paths don't have names.

a workaround is to encode the tags as part of the way name in the lua profile, for example "Vesterbrogade {'highway':'primary'}" or "{'highway':'cycleway','surface':'gravel'}. but this takes up extra space. 

a more efficient way of storing tags and passing them to the front-end would be great. one idea would be to store the encoded tags as a separate way 'name', ie store "Vesterbrogade" and "{'highway':'primary'}" separately. since many ways would have the same tags, it would take up less space.

emiltin:
related to #282.

TheMarex:
What you basically want is a a mini key-value storage engine inside OSRM that maps edge/node ids to properties. Could be worthwhile to invest some time in this, as this often comes up. However implementing this is not trivial.
Should be doable without any runtime/memory penalty *if you don't use it*.

emiltin:
right. preferably, the lua script will be able set the keys/values for each way.

systemed:
I use the way name workaround at cycle.travel. The results are heavily tokenised (one character for surface type, etc.) which saves some space and download time. But yes, a PBF-like key-val store wouldn't be a bad thing.

TheMarex:
This would also unblock https://github.com/Project-OSRM/osrm-backend/issues/77 since we could move the speed inside that field.

marcioaguiar:
I'm also interested in this feature, specifically on the surface property. Maybe it could be customized on the profile?

danpat:
If you can handle a second network round-trip, you can use this to get tag data:

https://github.com/mapbox/route-annotator

PR #2465, opened 2016-05-26
Add pronuncation to RouteStep

TheMarex:


emiltin:
this is a breaking change, since it requires front-ends to parse the name string instead of simply displaying it?

emiltin:
if possible i think it would be better to implement this via #1342.

we're already have to pack things like the highway tag into the name (using "My Way {'highway':'primary'}" for cases where the name has no name. now we're also packing name:pronounciation and would end up with "MyWay#meyeway{'highway':'primary'}".

taginfo.json
```diff
@@ -57,6 +57,7 @@
         {"key": "maxspeed:advisory"},
         {"key": "maxspeed:advisory:forward"},
         {"key": "maxspeed:advisory:backward"},
+        {"key": "name:pronounciation", "description": "Pronounciation hit for names", "object_types": ["way"]},
```
daniel-j-h: _hint_

systemed:
Should presumably be `name:pronunciation` (not `name:pronounciation`).

TheMarex:
@systemed damn my bad spelling. Should be fixed now.

@emiltin I found a way to do this without the hack and make it compatible with all clients. There is now a new property `pronunciation`.

include/extractor/extraction_way.hpp
```diff
@@ -46,6 +46,7 @@ struct ExtractionWay
     double backward_speed;
     double duration;
     std::string name;
+    std::string pronunciation;
```
daniel-j-h: You also have to adapt the `.clear()` function for this new member attribute.

include/engine/datafacade/internal_datafacade.hpp
```diff
@@ -587,6 +587,11 @@ class InternalDataFacade final : public BaseDataFacade
         return result;
     }
 
+    std::string GetPronunciationForID(const unsigned name_id) const override final
+    {
+        return GetNameForID(name_id + 1);
```
daniel-j-h: Doing the subsequent copies into the char data and using the name_id as base here is _really_ clever --- we should document that somewhere!

include/extractor/extraction_way.hpp
TheMarex: Good catch!

Issue #2466, opened 2016-05-26
Confusing INVALID_NAME_ID and INVALID_NAMEID constants


danpat:
We have two confusingly similar constants:

`INVALID_NAMEID == std::numeric_limits<unsigned>::max()`
https://github.com/Project-OSRM/osrm-backend/blob/master/include/util/typedefs.hpp#L72

`INVALID_NAME_ID == 0`
https://github.com/Project-OSRM/osrm-backend/blob/master/include/extractor/guidance/constants.hpp#L23

We should reconcile the logic for these, and only have one.  Possible simple option: rename `INVALID_NAME_ID` to `EMPTY_NAME_ID`.

/cc @MoKob

daniel-j-h:
This landed in https://github.com/Project-OSRM/osrm-backend/issues/2469! :tada:

PR #2469, opened 2016-05-26
fix INVALID_NAME_ID vs INVALID_NAMEID

MoKob:
Fixing https://github.com/Project-OSRM/osrm-backend/issues/2466

### DIFF
--- a/src/extractor/guidance/intersection_generator.cpp
+++ b/src/extractor/guidance/intersection_generator.cpp
@@ -176,7 +176,7 @@ Intersection IntersectionGenerator::mergeSegregatedRoads(Intersection intersecti
         const auto &first_data = node_based_graph.GetEdgeData(intersection[first].turn.eid);
         const auto &second_data = node_based_graph.GetEdgeData(intersection[second].turn.eid);
 
-        return first_data.name_id != INVALID_NAME_ID && first_data.name_id == second_data.name_id &&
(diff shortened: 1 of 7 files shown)


src/extractor/guidance/intersection_generator.cpp
```diff
@@ -176,7 +176,7 @@ Intersection IntersectionGenerator::mergeSegregatedRoads(Intersection intersecti
         const auto &first_data = node_based_graph.GetEdgeData(intersection[first].turn.eid);
         const auto &second_data = node_based_graph.GetEdgeData(intersection[second].turn.eid);
 
-        return first_data.name_id != INVALID_NAME_ID && first_data.name_id == second_data.name_id &&
+        return first_data.name_id != INVALID_NAMEID && first_data.name_id == second_data.name_id &&
```
danpat: @MoKob Are you sure this is the correct change?  Should we be testing for an INVALID_NAMEID (UINT_MAX), or the empty name (0)?

MoKob: The new INVALID_NAMEID which you changed (your PR has to be merged first) is `0`. So INVALID_NAME_ID was zero and will, after your PR also be 0 again. Right now it is broken.

MoKob:
Requires https://github.com/Project-OSRM/osrm-backend/pull/2464 for INVALID_NAMEID == 0 instead of UINT_MAX

Issue #2472, opened 2016-05-26
Reduce ramIndex size


oxidase:
Hello, in https://github.com/Project-OSRM/osrm-backend/blob/master/include/util/static_rtree.hpp#L431 the bottom-most level is used only to keep only one LeafNode index value  and the rectangle. 
I would suggest to remove the last level by moving the bounding rectangle into LeafNode because there are still 32 unused padding bytes in 4096 bytes pages. 

The negative side is only in moving child_is_on_disk into children id so it will reduce number of possible tree and leaf nodes to 2^31. But this should not be a problem, because the current number of leaf nodes is  22G/4K < 5400000.

The change must reduce the ramIndex size by factor 2.

TheMarex:
This is a good suggestion. Breaking the data format for this seems fine to me. :+1: Also have no reservations with 2^31 limit on leaf nodes.

oxidase:
@TheMarex i have created PR. i have underestimated the reduction factor, so it  must be about 128.

BTW, is there any way to check online planet statistics like in #2242 ?

danpat:
@oxidase Here are the current filesizes from `router.project-osrm.org`:

```
    4 May 27 02:31 latest.osrm.core
    8 May 26 22:58 latest.osrm.datasource_indexes
   12 May 26 21:23 latest.osrm.properties
   12 May 26 22:58 latest.osrm.datasource_names
   20 May 26 21:23 latest.osrm.timestamp
  15M May 26 21:23 latest.osrm.restrictions
 121M May 26 21:23 latest.osrm.names
 998M May 27 02:31 latest.osrm.level
 998M May 26 22:52 latest.osrm.enw
 2.3G May 26 22:50 latest.osrm.icd
 2.8G May 26 22:56 latest.osrm.ramIndex
 5.9G May 26 22:49 latest.osrm.edges
 7.9G May 26 22:58 latest.osrm.ebg
 8.9G May 26 22:57 latest.osrm.nodes
  11G May 26 21:51 latest.osrm.geometry
  18G May 27 02:31 latest.osrm.hsgr
  22G May 26 22:56 latest.osrm.fileIndex
  25G May 26 21:23 latest.osrm
  33G May 26 17:16 latest.osm.pbf
```

TheMarex:
This landed in `master` :tada:

PR #2476, opened 2016-05-27
Refactor ramIndex #2472

oxidase:
I have removed the last level of almost empty leaves where out of 532 bztes only 4 are used.
The size of ramIndex file must reduce by a factor ~ 128. For me
-rw-rw-r-- 1 miha miha 6859612 May 27 11:47 oberbayern-latest.osrm.ramIndex
is reduced to 
-rw-rw-r-- 1 miha miha 53736 May 27 15:13 oberbayern-latest.osrm.ramIndex
The size of fileIndex will be the same because 16 bytes out of unused 32 padding bytes now used for the leaf node minimum bounding rectangle.

Negative side of the improvement is the number of tree and leaf nodes is reduced by factor 2 to 2^31-1.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,9 @@
     - paramater `annotate` was renamed to `annotations`.
     - `annotation` as accidentally placed in `Route` instead of `RouteLeg`
 
+   - Infrastructure
+    - BREAKING: Changed the on-disk encoding of the StaticRTree to reduce ramIndex file size. This breaks the **data format**
+
 # 5.2.0 RC1
    Changes from 5.1.0
 
--- a/features/support/config.js
+++ b/features/support/config.js
@@ -50,21 +50,27 @@ module.exports = function () {
         };
 
         var hashExtract = (cb) => {
-            this.hashOfFiles(util.format('%s/osrm-extract%s', this.BIN_PATH, this.EXE), (hash) => {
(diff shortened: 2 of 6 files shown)


TheMarex:
@oxidase great work as always. Three things:

- [ ] Changelog entry: the data format broke again with this change.
- [ ] Rebase on master: seems like the shared library tests are failing for whatever reason, maybe a rebase will fix that.
- [ ] Can you run `rtree-bench` on this? I'm curious if this increased RAM locality and hence request times.

oxidase:
@TheMarex updated ChangeLog and rebased, but now it requires  #2476, so another  rebase probably will be needed. I will check shared library test if it fail again, locally it works for me. 

I checked performance changes with rtree-bench, but i can not tell you whether changed RAM locality changes timing, because my osrm file set completely fits my RAM, so pages can be easily cached by OS. I will try to check with larger OSM regions. Also commit 57bc4b4 sightly improves performance about 3-5%, because QueryCandidates are 25%  smaller and vector emplacements in PQ take less time.

oxidase:
@TheMarex just updated the branch to pass the shared library tests. The fail was in stale hashing of binaries that do not include changes for the travis build. So i added also hashing of .so libraries if any.

TheMarex:
@oxidase good catch with the hashing failing for shared library changes. :+1:

After a rebase on `master` I feel good about merging this and including it in 5.2.0 RC2.

oxidase:
@TheMarex ok, rebased

Issue #2189, opened 2016-04-01
Range analysis plugin


roroettg:
Hello,

I am currently working on a plugin, which should be giving me all ways from a starting point within a given distance. I have worked my way through the code and read about Contraction Hierarchies, but I have still no clue if or how i can implement it.

As far as I know every algorithm already implemented to solve the shortest path problem uses a bidirectional Dijkstra. In my case I think the best way to solve the problem is a normal Dijkstra, but i could not find any information if Contration Hierarchies are able to be used that way. 

Any hints or ideas where to start are greatly appreciated.

daniel-j-h:
Hey hey, dropping a few related resources:

@danpat's `osrm-isochrone` tool is available in this branch:
https://github.com/Project-OSRM/osrm-backend/tree/isochrone_tool

@TheMarex has an experimental osrm isochrone project here:
https://github.com/mapbox/osrm-isochrone

@emilymdubois can probably tell you more about isochrones:
https://www.mapbox.com/blog/isodistances-part-1/
https://www.mapbox.com/blog/isodistances-travel-times/

danpat:
@roroettg Take a look at the `src/tools/isochrone.cpp` for an example of how to do plain Dijkstra with OSRM data.

Unfortunately, you can't perform a simple Dijkstra search on a Contraction Hierarchy - the graph structure is optimized for 1:1 routing, not 1:N.

The `isochrone.cpp` loads the pre-contracted node-based graph into RAM and uses that for Dijkstra - but I haven't yet worked out if we can include this in the main OSRM routing daemon - it doubles the memory requirements :-(  I suspect I'll make it a completely separate tool/daemon.

roroettg:
Thank you very much. 

I will keep this updated, once i have some results to show. :-)

roroettg:
Hello,

I was able to create a first implementation of an Isochrone-Plugin based on danpats's code for the current version of osrm. As @danpat  mentioned the memory increase is significant. Is there a way to reduce the memory or is there any possibility to perform these searches on the contracted graph?

Another issue for me is the testing infrastructure. As far as I understand the tests use the shared memory function of osrm, but right now I am only able the to load the uncontracted graph from a file. Is the uncontracted graph also stored in the shared memory? If so, how can i access it?

As I am new to opensource projects is this a thing I should make a pull request for once i got some tests running?

danpat:
[...]

PR #2477, opened 2016-05-27
Isochrone Plugin #2189

roroettg:
I hope I am doing this right. 

As requested in #2189 this pullrequest contains the isochrone plugin I am working on. 
- The plugin is optional and can be set with parameter -I or --isochrone
-  No additional data gets loaded then the plugin is not used


include/engine/routing_algorithms/one_to_many.hpp
```diff
@@ -0,0 +1,128 @@
...
+        util::SimpleLogger().Write() << phantomSource;
+        // explore search space
+        while (!query_heap.Empty())
+        {
+//            ForwardRoutingStep(query_heap, distance, predecessorMap, distanceMap);
+            ForwardRoutingStep(query_heap, backward_heap, )
+        }
+    }
+
+    void ForwardRoutingStep(QueryHeap &query_heap,
```
MoKob: If I understand your code correctly, it will not in-fact compute an actual isochrone.
Since you are operating on a graph hierarchy, the OSRM graph will only contain upwards facing edges.
If you start your search on the topmost node, your search will return an empty set of edges.
The rest will depend severely on the hierarchy itself.

I recommend reading up on Contraction Hierarchies (CH) in general. In short: to find actual distances in a CH, you require a bidirectional search. Variants of One-To-Many exist, but they require the addition of downwards-facing graphs (which are not available in OSRM).
The output of this search is more or less random.

roroettg:
@MoKob 
You have actually been looking at the wrong code. I thought i had deleted the one_to_many.hpp. This was my first try on implementing an algorithm. The code I am using now for the isochrones is in **/src/engine/plugins/isochrone.cpp**. There you can see that I load my own graph, which isn't contracted.

Here are some screenshots from my current status, where I use Leaflet.js and/or Mapbox GL to visualise my data. I switched to Mapbox GL to be able to visualise bigge data-sets.

Legend:
- blue are the routes that can be driven
- red is a convex hull calculated by monotone chain algorithm [**include/util/monotone_chain.hpp**]
- green is a concave hull calculated by an algorithm based on the paper "**A New Concave Hull Algorithm and Concaveness Measure for n-dimensional Datasets***" [http://www.iis.sinica.edu.tw/page/jise/2012/201205_10.pdf]

screenshot from 2016-07-20 10 52 23
screenshot from 2016-07-20 10 52 45
screenshot from 2016-07-20 10 53 50
screenshot from 2016-07-20 10 53 31

MoKob:
@roroettg well. Given that OSRM uses a CH, you have to introduce a parallel graph for your plugin to work. I assume that you are using Dijkstras algorithm on the plane Edge-Based-Graph?

I have to say that I am probably not the best person to decide how this is going to work. But I feel that a plugin that requires a full parallel graph might be hard to integrate fully into OSRM. /cc @TheMarex

TheMarex:
@roroettg that looks pretty cool. Thanks for this contribution. I would like to get this properly reviewed and then think about if (and how) we can move this forward to be included in upstream. Biggest blocker here is currently time constraints on our side.

The biggest problem that we face is how we could handle optional plugins. Currently our library interface exposes a function for each service:

```cpp
class OSRM final
{
  public:
    explicit OSRM(EngineConfig &config);

    ~OSRM();

    // Moveable but not copyable
    OSRM(OSRM &&) noexcept;
    OSRM &operator=(OSRM &&) noexcept;

    Status Route(const RouteParameters &parameters, json::Object &result);
    Status Table(const TableParameters &parameters, json::Object &result);
    Status Nearest(const NearestParameters &parameters, json::Object &result);
    Status Trip(const TripParameters &parameters, json::Object &result);
    Status Match(const MatchParameters &parameters, json::Object &result);
    Status Tile(const TileParameters &parameters, std::string &result);

  private:
    std::unique_ptr<engine::Engine> engine_;
};
```

If we now disable a service that would mean we have a function that would do nothing? @daniel-j-h do you have an idea how we could solve this in a C++-y way?

Conditionally loading additional data is something that we should figure out down the road anyway (why would you load lane guidance if you only want to run distance table requests etc..), so I'm not too concerned here.

The command line flag is a first good step, but we might want a more general interface here.

Just skimmed the diff, did you run `script/format.sh`? It only works correctly with clang-format 3.8 I think so you would probably need to return this.

danpat:
I have a possible strategy that does not require loading the edge-based-graph into memory at all - while we can't explore the CH directly, we can re-create the parts of the edge-based-graph we need on-the-fly.

The recent work I did on the debug tile plugin to add turn penalty data involved the re-derivation of the edge-based graph within the tile being queried.  This is possible by fetching all data in a geographic area, then re-connecting the dots.  See [this code](https://github.com/Project-OSRM/osrm-backend/blob/3f07a830e67c07d5ad63b8a22c17a0ab687a4eaa/src/engine/plugins/tile.cpp#L384-L610).

The rough idea would be to come up with a reasonable "search tile" size (my gut says 5km x 5km would be good starting point).  The rough algorithm would be:

  1.  Fetch a "tile" of data around our isochrone start point.
  2.  Construct the local edge-based-graph from that data
  3.  Begin our isochrone search
  4.  If we hit the edge of our tile without terminating, fetch the neighbor tile and extend our edge-based-mini-graph
  5.  Continue the isochrone search, repeating (4) as necessary until the search ends.

Performance would be dictated by (a) the cost of the bbox RTree search to fetch tile data, and (b) the cost of re-constructing the local edge-based-graph from that tile.  For the example ranges above by @roroettg , I suspect we could fetch all data in a single "tile".

danpat:
Ok, I just had a go at this to see what kind of performance we'd get.  It was relatively easy to copy the logic from the tile plugin to create the edge-based-graph we need.

100 second radius
screen shot 2016-12-04 at 1 32 55 am
200 second radius
screen shot 2016-12-04 at 1 32 35 am
300 second radius
screen shot 2016-12-04 at 1 32 42 am

A 400-second  isochrone on the car profile is rendered in about 200ms, including the generation of 1.2MB of GeoJSON.  Its not doing the tile loading, but it's loading a 0.2 degree x 0.2 degree box (about 10km x 10km in Münster).

@roroettg For your use-case, how large were you planning on allowing the isochrones?  If I push my code up, do you think you could adapt your concave hull to it?

Komzpa:
@danpat just in case, we need 30 minute car isochrones with RPS of 50k. It looks like to achieve that some caching is needed.

danpat:
@Komzpa RPS == Requests Per Second, or https://xkcd.com/645/ ?

danpat:
screen shot 2016-12-04 at 11 47 30 am

@Komzpa ^^ this is a 30 minute driving isochrone centered on the same point as the above examples.

On my laptop (with an i7 CPU), it takes about 6.3 seconds, broken down into:

  - Fetch a 2x2 degree bounding box from the RTree - 102.185ms
  - Create an initial lookup table of edges - 283ms
  - Constrct the edge-based-graph - 383ms
  - Perform the dijkstra search on the edge-based-graph - 4869ms
  - Geojson serialization - ~900ms

This approach is basically the worst case for isochrone generation - it employs no speedups at all.

I'm of the opinion that even a limited isochrone feature would be good to have - the demo server can easily restrict the allowed isochrone size to something that won't eat up all our CPU, and folks that want to do bigger ones can set up their own servers.

Additional memory use is 🙅  for the demo server, but I think isochrone sizes would be just fine.

roroettg:
@danpat Nice Work. For our use-case isochrones with the size of up to two hours for a bicycle profile would be perfect. With my current approach the memory is most of the time the limit when the isochrones get too big. On a different branch I separated the isochrone-service from the current routed-service to avoid having two graphs at runtime. 

@danpat Yes, probably i can adapt my code.

danpat:
Yeah, the memory problem....I think we just need to make the limit configurable at runtime - if the average travel speed is slow, then 2 hrs may not actually be all that big of a maximum radius.

I'll push my changes up to a separate branch here and you can take a look.

danpat:
Pushed to here: https://github.com/Project-OSRM/osrm-backend/tree/ch_isochrone_poc

This doesn't implement the tiled approach, it simply fetches a bounding box of data using a maximum-possible-radius calculation (200km/h for cars).  The nice thing about this is that it scales down quite well - small isochrones are quite fast.

My code is simply returning the geometry of all edges inside the isochrone boundary - there's no data on points for age, and the data is not tree-like in anyway.

That said, these geometry adaptations should be relatively straight forward.  It could also probably get quite a bit faster with some work on cache locality.

I spent a bit of time over the weekend reading up on one-to-many speedup approaches - I think there are some options for future work.  For now, I'd just like to get a workable plugin in, and get folks using it, then we can get feedback on what needs to work better.

One thing I'd be very interested in knowing:  do folks care for exact isochrone trees, or are approximate bounding polygons more useful?

Komzpa:
@danpat 
Thanks for the benchmark! :)
We do care for exact isochrone trees. At very least, there are different travel times to left and right sides of the two-way streets, up to serveral minutes - we need to distinguish those.

DrewRay:
Happy New year, is this still blocked by merge issues?  is it even a priority to keep working on?

danpat:
@DrewRay as far as I know, this isn't a priority for anyone, thus the lack of progress.

If you've got time and are willing to work on it, please feel free :-)  For my branch a couple of comments back, the thing I got stalled on was making a reasonably compact JSON-like serialization of the isochrone tree data.  GeoJSON doesn't really work for this.

frodrigo:
> One thing I'd be very interested in knowing: do folks care for exact isochrone trees, or are approximate bounding polygons more useful?

On my side, each time I need isochrones it is for the hull, never for the tree.

frodrigo:
It would be nice to have the maxspeed as parameter.

DrewRay:
Agreed, I can use a hull to overlay and get the network again.

frodrigo:
I test this branch and got two issues:
- the selected graph component can be a (very) small one
- the output tree can be disjoint. I generate many isochrone with this branche and I got this one: http://bl.ocks.org/anonymous/raw/94631c4ea04286efc1fb6f03c69d3744/ using the custom profile

fijemax:
Hi,
I create this PR : #3613 to correct @frodrigo problem.

include/engine/plugins/isochrone.hpp
```diff
@@ -0,0 +1,99 @@
...
+                          std::vector<SimpleEdge> &graph_edge_list);
+
+    void dijkstraByDuration(IsochroneVector &set, NodeID &source, int duration);
+    void dijkstraByDistance(IsochroneVector &isochroneSet, NodeID &source, double distance);
+    void update(IsochroneVector &s, IsochroneNode node);
+
+  public:
+    explicit IsochronePlugin(const std::string base);
+
+    Status HandleRequest(const std::shared_ptr<datafacade::BaseDataFacade> facade,
```
fijemax: This can't compile `datafacade::BaseDataFacade` need to be `const`.

fijemax:
The `ch_isochrone_poc` branch doesn't work since last rebased on master branch.

danpat:
@fijemax I've rebased and removed some WIP code from that branch, and opened a new PR at https://github.com/Project-OSRM/osrm-backend/pull/3652

mloskot:
Is this proposal deprecated (to-be-closed) as superseded by #3652 or both are being developed as alternative approaches?

TheMarex:
Super-seeded by https://github.com/Project-OSRM/osrm-backend/pull/3652

mloskot:
@TheMarex Does keeping the #3652 open mean the isochrone project is being reactivated?

PR #1865, opened 2016-01-07
[not ready] We do not require sudo on Travis

daniel-j-h:
Waiting for Travis to give thumbs up.

PR #2479, opened 2016-05-27
[WIP] Sudoless install + add job testing sudo:false

springmeyer:


daniel-j-h:
How is this sudoless install, when https://github.com/Project-OSRM/osrm-backend/pull/2479/files#diff-354f30a63fb0907d4ad57269548329e3R42 adds sudo: required too all our envs anyway? I don't understand the difference here.

springmeyer:
@daniel-j-h sorry for the confusion. I need to get in the habit of explaining pull requests when I create them. I've been tending to create pull requests before explaining them so I can trigger travis to build (travis only builds for PR's not for branch pushes, right?) to first confirm things work. Anyway, I'll try to avoid that habit for now on.

As far as your question, the motivation behind this PR is to start testing on travis `sudo:false` machines. Because the  `sudo:false` machines ban sudo I think it would be ideal, for consistency, to move all jobs to avoid using sudo even though some can.

The `sudo:false` machines boot significantly faster that the `dist:trusty` machines and I'm planning to add the sanitizers and static library deps to the new `sudo:false` job eventually (as I'll use mason for the static libs there is no need for the trusty machine). I've marked this pull as WIP to more clearly indicate that this is just a first step so far.

daniel-j-h:
The problem with this is that your `sudo: false` build does not run on Trusty.

https://github.com/Project-OSRM/osrm-backend/pull/2479/files#diff-354f30a63fb0907d4ad57269548329e3R33

Which from reading the docs is not possible at the moment:

https://docs.travis-ci.com/user/ci-environment/#Virtualization-environments
https://docs.travis-ci.com/user/trusty-ci-environment/

We had to switch to Trusty because of some packaging / build issues from what I can recall.

https://github.com/Project-OSRM/osrm-backend/pull/1865 was the second try at this, can't find the first pull request though.

springmeyer:
@TheMarex @daniel-j-h My overall goal is to get an osrm-backend job that builds/tests against static libs and this PR was a precursor for testing. But I'll close this now as it has gone stale and when I get time again - anticipating that will be mid-week this week - I'll refresh the necessary bits of this into another PR.

PR #2371, opened 2016-05-04
Destination Signs Quest II

daniel-j-h:
WIP. For https://github.com/Project-OSRM/osrm-backend/issues/2267.

PR #2482, opened 2016-05-27
Destinations / Exit Signs Feature

daniel-j-h:
For https://github.com/Project-OSRM/osrm-backend/issues/2267.

This makes https://github.com/Project-OSRM/osrm-backend/pull/2371 obsolete. Adds an API `destinations` field passing through destination / exit sign metadata using @MoKob and @TheMarex clever trick of using the name table.


emiltin:
is there any reason do duplicate the destination names in the name and the destinations fields? e.g. one of the cucumber tests show that this will be returned:

```
name = EF (A1: Berlin)
destinations = A1: Berlin
```

wouldn't it be simpler to return:
```
name = EF
destinations = A1: Berlin
```

or even just keep the osm tags:
```
name = EF
destination = Berlin
destination:ref = A1
```

and let the front-end construct the name string as preferred?

daniel-j-h:
This is due to us simply re-writing destination tags to refs in case the street has no name at all.

This allows us to be backwards compatible with the frontends out there at the moment.

PR #2483, opened 2016-05-28
Fix gcc compile

oxidase:
Addition to #82bfd4e:
- vector::emplace_back requires MoveInsertable and EmplaceConstructible types,
- deque::emplace_back requires EmplaceConstructible types.
- std::atomic is neither copyable nor movable, but
  may be instantiated with any TriviallyCopyable type.

### DIFF
--- a/src/contractor/contractor.cpp
+++ b/src/contractor/contractor.cpp
@@ -35,6 +35,7 @@
 #include <atomic>
 #include <bitset>
 #include <cstdint>
+#include <deque>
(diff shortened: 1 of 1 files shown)


TheMarex:
Thanks for your fix :+1: However this was made obsolete with https://github.com/Project-OSRM/osrm-backend/pull/2484

Both approaches work, but using atomic instructions was a little overkill for a simple debug message.

PR #2484, opened 2016-05-28
Remove the use of atomics and use thread-local accumulation instead

TheMarex:


oxidase:
@TheMarex  #2483 solves pretty the same issue. There is nothing bad in using atomics, but vector is a bad container because it requires copyable or movable types and copying is by itself a non-atomic operation, so it is not possible to make vector of atomics. Using deque with emplace_back
would solve the issue because atomics need not be moved or copied.

TheMarex:
IMHO trying to solve this with atomics was a little overkill. This can be perfectly well done in a simple two-step pattern, that will not break due to compiler strangeness. Thanks for you fix though @oxidase :+1:

PR #2488, opened 2016-05-29
Add feature name to vector tiles

vkurchatkin:


TheMarex:
@danpat do you think this makes sense w.r.t. to vector tile size?

vkurchatkin:
@TheMarex it probably makes sense to include names only on larger zoom levels

danpat:
I think this is probably better achieved by using a separate basemap layer from a regular tileserver.  The debug tiles are really just to expose OSRM-internal data that's not available elsewhere (speeds, durations, turn costs, etc).  This is what the debug map at http://map.project-osrm.org/debug/ does, it allows plenty of context.

The tiles are already pretty heavy - adding name data to every single segment will only make that worse.  Additionally, the way the geometries are segmented means that label placement will be really terrible - labels won't follow curves, and there will be lots of collisions leading to pretty random label appearance.

@vkurchatkin Thanks for the contribution, but I think we'll hold off on this change for the moment.

vkurchatkin:
@danpat thanks for the feedback!

>  The debug tiles are really just to expose OSRM-internal data that's not available elsewhere (speeds, durations, turn costs, etc).

Name is kind of internal, because it can be changed in profile script. I was trying to use it to attach  some data (way id, for example) and needed a way to visualize.

> the way the geometries are segmented means that label placement will be really terrible

Probably true, I've tried only to access it programmatically on mouse events.

danpat:
@vkurchatkin ah, I see, hmm, interesting use-case.

With that in mind, I'm 👍  on this change.  Merging now :-)

PR #2490, opened 2016-05-30
Let osrm-routed log status code, fixes #512.

### LINKED ISSUE #512: log result codes
osmr-routed logs the incoming request. it would be useful to add the result code too.

daniel-j-h:
Disclaimer: osrm-routed is not recommended for production usage.
Use node-osrm instead: https://github.com/Project-OSRM/node-osrm


TheMarex:
:+1: Looks good to merge.

PR #2491, opened 2016-05-31
highway=motorway_link does not imply oneway, fixes #1662

### LINKED ISSUE #1662: motorway_link is not implied oneway
According to https://github.com/gravitystorm/openstreetmap-carto/issues/1820 and the wiki OSRM routing rules are incorrect.

daniel-j-h:
For #1662.

References:
- http://wiki.openstreetmap.org/wiki/Tag:highway%3Dmotorway_link#Tagging_oneway
- http://wiki.openstreetmap.org/wiki/Tag:highway%3Dmotorway_link
- http://wiki.openstreetmap.org/wiki/Tag:highway%3Dmotorway

### DIFF
--- a/features/guidance/motorway.feature
+++ b/features/guidance/motorway.feature
@@ -11,9 +11,9 @@ Feature: Motorway Guidance
             |   |   |   | f | g |
 
         And the ways
-            | nodes  | highway       |
-            | abcde  | motorway      |
-            | bfg    | motorway_link |
+            | nodes | highway       | oneway |
+            | abcde | motorway      |        |
+            | bfg   | motorway_link | yes    |
 
        When I route I should get
             | waypoints | route         | turns                                |
@@ -27,9 +27,9 @@ Feature: Motorway Guidance
             |   |   |   | g | e |
 
         And the ways
-            | nodes  | highway       |
-            | abcde  | motorway      |
-            | bfg    | motorway_link |
+            | nodes | highway       | oneway |
+            | abcde | motorway      |        |
+            | bfg   | motorway_link | yes    |
 
        When I route I should get
             | waypoints | route         | turns                         |
@@ -44,9 +44,9 @@ Feature: Motorway Guidance
 
 
         And the ways
-            | nodes  | highway       |
-            | abcde  | motorway      |
-            | cfg    | motorway_link |
+            | nodes | highway       | oneway |
+            | abcde | motorway      |        |
+            | cfg   | motorway_link | yes    |
 
        When I route I should get
             | waypoints | route         | turns                                |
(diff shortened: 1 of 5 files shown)


features/guidance/motorway.feature
```diff
@@ -176,11 +176,11 @@ Feature: Motorway Guidance
...
        When I route I should get
-            | waypoints | route           | turns                                |
-            | a,e       | abcde,abcde     | depart,arrive                        |
-            | f,e       | fgc,abcde,abcde | depart,merge slight left,arrive      |
-            | a,i       | abcde,chi,chi   | depart,off ramp slight right,arrive |
-            | f,i       | fgc,chi,chi     | depart,off ramp right,arrive        |
+            | waypoints | route           | turns                              |
+            | a,e       | abcde,abcde     | depart,arrive                      |
+            | f,e       | fgc,abcde,abcde | depart,merge slight left,arrive    |
+            | a,i       | abcde,chi,chi   | depart,on ramp slight right,arrive |
```
daniel-j-h: a -> i should always have been on ramp and not off ramp, or am I missing something here?

features/guidance/motorway.feature
```diff
@@ -194,11 +194,11 @@ Feature: Motorway Guidance
...
        When I route I should get
-            | waypoints | route           | turns                               |
-            | a,e       | abcde,abcde     | depart,arrive                       |
-            | f,e       | fgc,abcde,abcde | depart,merge slight right,arrive    |
-            | a,i       | abcde,chi,chi   | depart,off ramp slight left,arrive |
-            | f,i       | fgc,chi,chi     | depart,off ramp left,arrive        |
+            | waypoints | route           | turns                             |
+            | a,e       | abcde,abcde     | depart,arrive                     |
+            | f,e       | fgc,abcde,abcde | depart,merge slight right,arrive  |
+            | a,i       | abcde,chi,chi   | depart,on ramp slight left,arrive |
```
daniel-j-h: same here /cc @MoKob can you clarify?

features/guidance/motorway.feature
MoKob: You are exiting the motorway abcde at this point. So chi is an off ramp (its not about getting on / off the ramp but whether you are leaving / entering a motorway)

features/guidance/motorway.feature
MoKob: its just a mirror of the scenario above, so same comment applies.

daniel-j-h:
Fixed the on/off ramp test cases relying on implied oneway.

TheMarex:
This will probably degrade the routing performance initially.

@planemad keep a look out to routing bugs that would occur here after the next OSRM release. Do you think we could add the non-tagged `motorway_link` roads to to-fix?

planemad:
@TheMarex unless this causes routing failures, it might be hard to keep a track of routing bugs unless explicitly reported. In the meanwhile, we can detect such ways and load into to-fix for a community review.

PR #2495, opened 2016-06-01
handle short summaries correctly

MoKob:
During my test-updates for collapsed instructions I came around this problem:

If a route consists simply of the phantom node and/or the phantom node segment at the end is significant, the summary can be off, depending on phantom node snapping.

This PR fixes the problem.

Grid:
| a | b |

with waypoints:
|a,b|

Could result in an empty summary. This is no longer the case with this PR

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,9 @@
    - Infrastructure
     - BREAKING: Changed the on-disk encoding of the StaticRTree to reduce ramIndex file size. This breaks the **data format**
 
+   - Bugfixes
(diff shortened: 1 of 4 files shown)


daniel-j-h:
This breaks the route library test on all Travis builds.
Seems like the test checks for the wrong behavior, though.

daniel-j-h:
Fixed in https://github.com/Project-OSRM/osrm-backend/pull/2495/commits/d8533209bf4e9292bbd1a5aad9f8b9a3dde2789d. This landed!

PR #2496, opened 2016-06-01
Install clang-3.8 via mason binary

springmeyer:
We needed to disable clang-3.8 builds in https://github.com/Project-OSRM/osrm-backend/commit/e7159adf59dde64a8e10a330595c7ef76224122e due to LLVM apt servers going offline (http://lists.llvm.org/pipermail/llvm-foundation/2016-June/000025.html).

This brings back testing of clang-3.8 via a mason package downloaded from s3, which should be fast and stable.

Bonus: this package supports `-fsanitize=<things>` like `address` and `integer`.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -49,16 +49,13 @@ matrix:
           packages: ['g++-4.8', 'libbz2-dev', 'libstxxl-dev', 'libstxxl1', 'libxml2-dev', 'libzip-dev', 'lua5.1', 'liblua5.1-0-dev', 'libtbb-dev', 'libgdal-dev', 'libluabind-dev', 'libboost-all-dev', 'ccache']
       env: CCOMPILER='gcc-4.8' CXXCOMPILER='g++-4.8' BUILD_TYPE='Debug'
 
-  # LLVM APT servers switched off, commenting Clang builds out as they will always fail now
-  # http://lists.llvm.org/pipermail/llvm-foundation/2016-June/000025.html
-  #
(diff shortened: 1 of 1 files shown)


springmeyer:
This is working now after fixing the clang-3.8 builds to be [proper `Release` builds](https://github.com/mapbox/mason/issues/183) and disabling openmp, which is not working yet in the clang-3.8 binaries.

As far as openmp, we have a few routes:

  - 1) keep the above patch that disables openmp
  - 2) rebuild clang-3.8 binaries again to fully remove partial/broken openmp support

I am out of the office till next Monday. When I'm back I can take a look at 2).

springmeyer:
> 2) rebuild clang-3.8 binaries again to fully remove partial/broken openmp support

Was able to do this today, so I reverted the disabling of openmp, and the builds are working now. This is ready to merge once travis is all green.

daniel-j-h:
Interested in how OpenMP support is broken in LLVM 3.8. Care to elaborate?

springmeyer:
> Interested in how OpenMP support is broken in LLVM 3.8. Care to elaborate?

@daniel-j-h it was compiled into the binary but then programs that linked it needed to load the `libomp.so` library at runtime. That would not work with a mason provided clang++ since the `libomp.so` is not in the default shared library paths. We could get it working by setting `LD_LIBRARY_PATH` to the custom `libomp.so` location. But since clang++ builds have not been using omp previously I did not want to address that in this pull.

PR #2318, opened 2016-04-25
Coverage-based Fuzz Testing

daniel-j-h:
Follow up from https://github.com/Project-OSRM/osrm-backend/pull/2251, PR against master.

PR #2497, opened 2016-06-02
guides for good tests

MoKob:
Started of with a guide on stuff to consider when writing cucumber tests.

- [x] cucumber best practice
- [x] boost unit tests guide @daniel-j-h can you take this one?


daniel-j-h:
https://github.com/Project-OSRM/osrm-backend/pull/2318 is still not merged, otherwise we could add a section for it, too.

MoKob:
You could add a todo on that in #2318 . Would prefer to have this merged and then linked in the wiki

PR #2503, opened 2016-06-04
fixes for building on armv7

noblige:
Compiles and runs on target.
https://github.com/Project-OSRM/osrm-backend/wiki/Building-for-ARMv7

@danpat

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -118,7 +118,7 @@ if(CMAKE_BUILD_TYPE MATCHES Release)
   message(STATUS "Configuring OSRM in release mode")
   # Check if LTO is available
   check_cxx_compiler_flag("-flto" LTO_AVAILABLE)
-  if(LTO_AVAILABLE)
+  if(LTO_AVAILABLE AND NOT NOLTO)
     set(OLD_CXX_FLAGS ${CMAKE_CXX_FLAGS})
     # GCC in addition allows parallelizing LTO
     if("${CMAKE_CXX_COMPILER_ID}" STREQUAL "GNU")
@@ -247,7 +247,6 @@ list(APPEND CMAKE_MODULE_PATH "${CMAKE_CURRENT_SOURCE_DIR}/third_party/libosmium
 set(OSMIUM_INCLUDE_DIR "${CMAKE_CURRENT_SOURCE_DIR}/third_party/libosmium/include")
 find_package(Osmium REQUIRED COMPONENTS io)
 add_dependency_includes(${OSMIUM_INCLUDE_DIR})
-
 
 find_package(Boost 1.49.0 REQUIRED COMPONENTS ${BOOST_COMPONENTS})
 
--- a/include/util/compat.hpp
+++ b/include/util/compat.hpp
@@ -1,0 +1,39 @@
+#ifndef OSRM_COMPAT_HPP
+#define OSRM_COMPAT_HPP
+
+#if defined(__ARM_ARCH_7A__)
+
+#include <string>
+#include <sstream>
+
+namespace std
+{
+    // See discussion https://answers.launchpad.net/gcc-arm-embedded/+question/238327
+    template <typename T> std::string to_string(const T& n)
+    {
+        std::ostringstream stm;
+        stm << n;
+        return stm.str();
+    }
+
+    // These might be fixed with a more recent gcc. See discussion on w.r.t. to ChromeOS gcc:
+    // https://code.google.com/p/android/issues/detail?id=54418
+    inline double round(double d)
+    {
+        return ::round(d);
+    }
+
+    inline double hypot(double x, double y)
+    {
+        return ::hypot(x, y);
+    }
+
+    inline long long llabs(long long n)
+    {
+        return ::llabs(n);
+    }
+}
+
+#endif
+
+#endif // OSRM_COMPAT_HPP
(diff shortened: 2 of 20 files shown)


include/util/compat.hpp
```diff
@@ -0,0 +1,36 @@
...
+#define OSRM_COMPAT_HPP
+
+#if defined(__ARM_ARCH_7A__)
+
+#include <string>
+#include <sstream>
+
+namespace std
+{
+    template <typename T> std::string to_string(const T& n)
```
TheMarex: Add comment:
```
See discussion https://answers.launchpad.net/gcc-arm-embedded/+question/238327
```

include/util/compat.hpp
```diff
@@ -0,0 +1,36 @@
...
+    {
+        std::ostringstream stm;
+        stm << n;
+        return stm.str();
+    }
+
+    inline double round(double d)
```
TheMarex: Add comment.
```
These might be fixed with a more recent gcc. See discussion on w.r.t. to ChromeOS gcc:
https://code.google.com/p/android/issues/detail?id=54418
```

CMakeLists.txt
```diff
@@ -322,27 +335,29 @@ target_link_libraries(osrm-extract osrm_extract ${Boost_LIBRARIES})
...
-set(CONTRACTOR_LIBRARIES
-    ${Boost_LIBRARIES}
-    ${CMAKE_THREAD_LIBS_INIT}
-    ${LUABIND_LIBRARY}
-    ${USED_LUA_LIBRARIES}
-    ${STXXL_LIBRARY}
-    ${TBB_LIBRARIES}
-    ${MAYBE_RT_LIBRARY}
-    ${MAYBE_COVERAGE_LIBRARIES})
+if(NOT ROUTED_STANDALONE)
```
TheMarex: If we really want to support this, we need to split the CMakeList not add a bunch of switches all over the place.

@daniel-j-h you did this for the `unit_tests` what are you experiences w.r.t to complexity here?

CMakeLists.txt
```diff
@@ -245,9 +248,13 @@ endif()
...
 set(OSMIUM_INCLUDE_DIR "${CMAKE_CURRENT_SOURCE_DIR}/third_party/libosmium/include")
-find_package(Osmium REQUIRED COMPONENTS io)
-add_dependency_includes(${OSMIUM_INCLUDE_DIR})
-
+if(NOT ROUTED_STANDALONE)
+  find_package(Osmium REQUIRED COMPONENTS io)
+  add_dependency_includes(${OSMIUM_INCLUDE_DIR})
+else()
+  # transitive dependency
+  include_directories(SYSTEM ${OSMIUM_INCLUDE_DIR})
```
TheMarex: why do you have a switch here? Adding OSMIUM include path to the search path while not calling find_package will be a no-op.

CMakeLists.txt
```diff
@@ -215,6 +215,9 @@ if("${LINKER_VERSION}" MATCHES "GNU gold" OR "${LINKER_VERSION}" MATCHES "GNU ld
 else()
   message(STATUS "Using unknown linker, not setting linker optimizations")
 endif ()
+if (STATIC_LIBSTDCXX)
+  set(LINKER_FLAGS "${LINKER_FLAGS} -static-libstdc++")
+endif()
```
daniel-j-h: CMake picks up your environment's `CXXFLAGS` and `LDFLAGS` --- there should be no need for this switch here, just export the appropriate flags.

CMakeLists.txt
daniel-j-h: The unit tests are currently doing this: provide a recursive CMakeLists.txt and register it at the end of the top level CMakeLists.txt. This way, we only require the Boost Test library when unit tests are requested.

noblige:
Addressed comments except for `CMakeList` refactoring, which feels like a bigger task than I'm comfortable with :) I wonder if we can do this as the separate next step?

TheMarex:
@noblige in that case please remove the flag to build `osrm-routed` standalone. We can fix this in another iteration. :+1:

noblige:
@TheMarex reverted `OSRM_STANDALONE` change - good to go now

include/util/compat.hpp
```diff
@@ -0,0 +1,39 @@
+#ifndef OSRM_COMPAT_HPP
+#define OSRM_COMPAT_HPP
+
+#if defined(__ARM_ARCH_7A__)
```
oxidase: Fixes are not really related to ARM architecture but to a specific libstdc++ version of the used toolchain. I would suggest to make it as a patch and apply it before building with the specific toolchain. Otherwise the patch brings redefinition errors for newer libstc++
```
[  1%] Building CXX object CMakeFiles/ENGINE.dir/src/engine/polyline_compressor.cpp.o
In file included from /home/miha/foss/osrm-backend/src/engine/polyline_compressor.cpp:3:0:
/home/miha/foss/osrm-backend/include/util/compat.hpp: In function ‘double std::round(double)’:
/home/miha/foss/osrm-backend/include/util/compat.hpp:21:33: error: ‘double std::round(double)’ conflicts with a previous declaration
     inline double round(double d)
                                 ^
In file included from /usr/arm-linux-gnueabihf/include/math.h:83:0,
                 from /usr/arm-linux-gnueabihf/include/c++/5/cmath:44,
```

noblige: @oxidase thanks! I will close this for now

PR #2504, opened 2016-06-05
Fixes for i686 and armhf platforms

oxidase:
Hello, i have fixed some issues to make 32 bit Debug/Release tests running and produce the same output as 64 bit. I will try to make Travis build for 32bit release.


.travis.yml
```diff
@@ -114,6 +118,9 @@ matrix:
       #- env: CCOMPILER='clang-3.8' CXXCOMPILER='clang++-3.8' BUILD_TYPE='Release' BUILD_SHARED_LIBS=ON
 
 before_install:
+  - if [[ ! -z $APT_ARCH ]] ; then sudo dpkg --add-architecture $APT_ARCH ; fi
```
TheMarex: This is interesting. We should talk to travis about adding this flag to their builds.

oxidase: If travis will have ARCH parameter it will be nice, because almost all my PRs to include i386 packages into the whitelist were rejected due to duplications. So it is mainly a workaround.
It does not change the main architecture. My local xenial:
```
dpkg --print-architecture
amd64
dpkg --print-foreign-architectures
i386
armhf
```

oxidase:
Just added armhf build, can be related to #2513
I also can add tests running for arm, but don't know how slow it will be.

TheMarex:
Looks very good. Great work to get ARM builds working on travis! We should try to coordinate this with the merge of https://github.com/Project-OSRM/osrm-backend/pull/2503

I'm aiming at the 6.x release series for inclusion here. Awesome work. :+1:

TheMarex:
@oxidase can you rebase? Since 5.2 is out of the door I would like to merge this now.

oxidase:
@TheMarex ok, rebased.

PR #2514, opened 2016-06-08
improve slipway handling to allow multiple styles of turn lanes

MoKob:
Sliproads can occur in multiple styles.

This PR adjusts their treatment to be able to handle them in more cases and makes the check for their existence more reliable.

See the cucumber tests for updated behaviour.
All cases listed previously returned two instructions instead of one.


MoKob:
Superseeded by: https://github.com/Project-OSRM/osrm-backend/pull/2521

PR #2517, opened 2016-06-08
fix interpolation sign error in RasterSource::GetRasterInterpolate me…

kaligrafy:
Line 16 of src/extractor/raster_source.cpp should be:
const float fromLeft = (lon - left * xstep - xmin) / xstep;
instead of
const float fromLeft = (lon - left * xstep + xmin) / xstep;

Can somebody confirm it should be a minus sign before xmin?

### DIFF
--- a/src/extractor/raster_source.cpp
+++ b/src/extractor/raster_source.cpp
@@ -62,7 +62,7 @@ RasterDatum RasterSource::GetRasterInterpolate(const int lon, const int lat) con
     const std::size_t right = static_cast<std::size_t>(fmin(ceil(xthP), width - 1));
 
     // Calculate distances from corners for bilinear interpolation
-    const float fromLeft = (lon - left * xstep + xmin) / xstep;
+    const float fromLeft = (lon - left * xstep - xmin) / xstep;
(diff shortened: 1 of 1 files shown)


src/extractor/raster_source.cpp
```diff
@@ -62,7 +62,7 @@ RasterDatum RasterSource::GetRasterInterpolate(const int lon, const int lat) con
     const std::size_t right = static_cast<std::size_t>(fmin(ceil(xthP), width - 1));
 
     // Calculate distances from corners for bilinear interpolation
-    const float fromLeft = (lon - left * xstep + xmin) / xstep;
+    const float fromLeft = (lon - left * xstep - xmin) / xstep;
```
TheMarex: Looking at the formular for bilinear interpolation this should be:

```
fromLeft = xthP - left; // will be < 1 and denotes the fraction of the x coordinate of the pixel
fromTop = ythP - top; // will be < 1 and denotes the fraction of the y coordinate of the pixel
```

@lbud can you confirm this?

TheMarex:
Addressed in a simplified version in the other PR.

PR #2520, opened 2016-06-09
[Not Ready] Guidance: Lane Anticipation

daniel-j-h:
Post-processing for turn lane anticipation during a range of subsequent quick turns.
This is part of Lane Guidance (https://github.com/Project-OSRM/osrm-backend/pull/2410).

Lane anticipation means: turn lanes later on in the route depend on turn lanes earlier.
Quick turns means: a series of turns where there is not much time to change lanes between turns.

The goal of this changeset is to guide the user to already take the correct lanes in anticipation of the following turns, reducing lane changes to a minimum. See the following scenario for an example.

There are two lanes going off the highway:

m1

http://mapillary.com/map/im/4IgMQQY0Q04iMLefQh4htw/photo

Shortly after the user can turn left or right at the following fork. Depending on this turn (left/right), we want the user to already take the correct lane (left/right) in the photo above going off the highway.

m2

http://mapillary.com/map/im/D7Bd8Sem5nZYXpIsE50MXA/photo

Here are the two possible routes for which we want the user to take the appropriate exit lane:

a2

http://map.project-osrm.org/?z=17&center=52.317494%2C13.552870&loc=52.318475%2C13.541604&loc=52.317228%2C13.553470&hl=en&alt=0

a1

http://map.project-osrm.org/?z=17&center=52.317494%2C13.552870&loc=52.318475%2C13.541604&loc=52.319701%2C13.553009&hl=en&alt=0

Here is the OSM Data Layer for this scenario:
https://www.openstreetmap.org/#map=17/52.31831/13.54993&layers=D

Overpass Turbo, Lanes in Berlin:
http://overpass-turbo.eu/map.html?Q=[out%3Ajson][timeout%3A25]%3B%0A%28%0A%20%20way[%22turn%3Alanes%22]%2852.23915380439046%2C12.938461303710936%2C52.79404304457785%2C13.838653564453125%29%3B%0A%29%3B%0Aout%20body%3B%0A%3E%3B%0Aout%20skel%20qt%3B

This changeset aims to implement lane anticipation for an arbitrary number of subsequent quick turns.

--------------------------------------------------------------------------------------------------------------------------

For the implementation the basic idea is this: we post-process all route steps, looking for series of quick turns. We then work on those series of quick turns independently as follows: we walk over all adjacent turns backwards _(turn_n, turn_n-1), (turn_n-1, turn_n-2), ..._ possibly constraining turn lanes for the previous turn dependent on the current turn. Constraining lanes works heuristically and assumes for example no crossing lanes during a turn, that is with two lanes before the turn the right lane will end up right of the left lane after the turn.

### DIFF
--- a/features/guidance/turn-lanes.feature
+++ b/features/guidance/turn-lanes.feature
@@ -290,26 +290,6 @@ Feature: Turn Lane Guidance
             | a,e       | road,through,through | depart,new name straight,arrive | ,1,   |
             | a,f       | road,right,right     | depart,turn right,arrive        | ,0,   |
 
-    Scenario: Anticipate Lane Change
-        Given the node map
-            | a |   | b |   | x |
-            |   |   |   |   |   |
-            |   |   | c |   | d |
-            |   |   |   |   |   |
-            |   |   | y |   |   |
-
-        And the ways
-            | nodes | turn:lanes:forward   | turn:lanes:backward |
-            | ab    | through\|right&right |                     |
-            | bx    |                      | left\|left&through  |
-            | bc    | left\|through        | left\|right         |
-            | cd    |                      | left\|right         |
-            | cy    |                      |                     |
-
-       When I route I should get
-            | waypoints | route       | turns                                            | lanes |
-            | d,a       | cd,bc,ab,ab | depart,end of road right,end of road left,arrive | ,0,1, |
-
      Scenario: Turn at a traffic light
         Given the node map
             | a | b | c | d |
--- a/include/engine/api/route_api.hpp
+++ b/include/engine/api/route_api.hpp
@@ -149,6 +149,7 @@ class RouteAPI : public BaseAPI
                                                               leg_geometry,
                                                               phantoms.source_phantom,
                                                               phantoms.target_phantom);
+                leg.steps = guidance::anticipateLaneChange(std::move(leg.steps));
                 leg_geometry = guidance::resyncGeometry(std::move(leg_geometry), leg.steps);
             }
 
--- a/src/engine/guidance/post_processing.cpp
+++ b/src/engine/guidance/post_processing.cpp
@@ -1,10 +1,11 @@
-#include "extractor/guidance/turn_instruction.hpp"
-#include "engine/guidance/debug.hpp"
 #include "engine/guidance/post_processing.hpp"
+#include "extractor/guidance/turn_instruction.hpp"
 
 #include "engine/guidance/assemble_steps.hpp"
 #include "engine/guidance/toolkit.hpp"
 
+#include "util/for_each_pair.hpp"
+#include "util/group_by.hpp"
 #include "util/guidance/toolkit.hpp"
 
 #include <boost/assert.hpp>
@@ -14,6 +15,7 @@
 #include <cmath>
 #include <cstddef>
 #include <iostream>
+#include <iterator>
 #include <limits>
 #include <utility>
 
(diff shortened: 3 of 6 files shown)


features/guidance/turn-lanes.feature
```diff
@@ -379,3 +379,25 @@ Feature: Turn Lane Guidance
...
+            | ab    | none\|none&none&slight_right&slight_right          |   5   | motorway      |        |
+            | bc    |                                                    |   2   | motorway_link | yes    |
+            | cd    | slight_left\|slight_left&slight_right&slight_right |   3   | motorway_link | yes    |
+            | di    |                                                    |   2   | motorway_link | yes    |
+            | dj    |                                                    |   2   | motorway_link | yes    |
+
+
+       When I route I should get
+            | waypoints | route          | turns                                         | lanes    |
+            | a,i       | ab,bc,cd,di,di | depart,off ramp right,turn slight left,arrive | ,1,,2 3, |
```
daniel-j-h: Note for adding this to our todo list for tomorrow.

daniel-j-h:
- [ ] refactor (`TurnInstruction` related functionality => into class, ...)
- [ ] more cucumber scenarios (fan-in / fan-out combinations, ending in multiple-lanes, ...)
- [ ] check out behavior on real world use-cases
- [ ] review @TheMarex @MoKob

features/guidance/turn-lanes.feature
```diff
@@ -290,25 +290,264 @@ Feature: Turn Lane Guidance
             | a,e       | road,through,through | depart,new name straight,arrive | ,1,   |
             | a,f       | road,right,right     | depart,turn right,arrive        | ,0,   |
 
-    Scenario: Anticipate Lane Change
+    @anticipate
+    Scenario: Anticipate Lane Change for subsequent multi-lane intersections
```
MoKob: I'd prefer to have a separate file for the base-case and the anticipation feature. In addition, you should try and handle special cases.

Collapsed instructions and roundabouts spring to mind.

include/engine/api/route_api.hpp
```diff
@@ -149,6 +149,7 @@ class RouteAPI : public BaseAPI
                                                               leg_geometry,
                                                               phantoms.source_phantom,
                                                               phantoms.target_phantom);
+                leg.steps = guidance::anticipateLaneChange(std::move(leg.steps));
                 leg_geometry = guidance::resyncGeometry(std::move(leg_geometry), leg.steps);
```
MoKob: For roundabout handling, we might either have to handle the lanes within the roundabout post-processing already or prior to it, since otherwise intermediate instructions are lost.
Is your logic in a state that would allow handling of arbitrary instruction streams having no-lane information in between?

src/engine/guidance/post_processing.cpp
```diff
@@ -726,6 +729,13 @@ std::vector<RouteStep> collapseTurns(std::vector<RouteStep> steps)
         return index;
     };
 
+    // TODO needs to check whether lane data is required!
```
MoKob: this looks like a messed up rebase? The function was moved to the top of the file to be accessible from multiple locations

src/engine/guidance/post_processing.cpp
```diff
@@ -1131,6 +1141,94 @@ std::vector<RouteStep> assignRelativeLocations(std::vector<RouteStep> steps,
     return steps;
 }
 
+std::vector<RouteStep> anticipateLaneChange(std::vector<RouteStep> steps)
+{
+    const constexpr auto MIN_DURATION_NEEDED_FOR_LANE_CHANGE = 15.;
+
+    // Postprocessing does not strictly guarantee for only turns
```
MoKob: Both New Name and Notification can have valid lane items that need to be handled. Why do restrict yourself here?

daniel-j-h:
This was merged into the `features/guidance/lane` branch! :tada:

PR #2521, opened 2016-06-09
turn link handling

MoKob:
Some turn links are longer than a normal intersection. This PR augments what is called a obvious turn based on road classes for same-name turns and introduces handling of link-roads that offer no choice at the next turn.

See modified turn.feature for an example.

Previous outputs (in order):

`depart,continue slight right,arrive`
`depart,continue straight,turn left,arrive`
`depart,arrive`

Also Includes:
Sliproads can occur in multiple styles.

This PR adjusts their treatment to be able to handle them in more cases and makes the check for their existence more reliable.

See the cucumber tests for updated behaviour.
All cases listed previously returned two instructions instead of one.


MoKob:
Situation in osm: https://www.openstreetmap.org/search?query=52.458091%2013.288796#map=19/52.45809/13.28880

PR #2526, opened 2016-06-10
Fix scenario for steep "off ramp" which generated "on ramp" 

daniel-j-h:
Test scenario (currently failing) for steep `off ramp` generating `on ramp`. /cc @MoKob


MoKob:
Fixed it.

TheMarex:
Looks good to me. Coverage degrees means only 71% of the lines in the diff are hit, but that is mostly due to currently uncovered lines hitting formatting/namespace changes.

PR #2542, opened 2016-06-13
Copy data to beginning of buffer, not end.

danpat:
033dc0a72d9930d51fe97ec4010c7b176974631a introduced some faulty logic when copying the PBF result from the tile plugin to the HTTP response buffer.

This was causing EACCESS segfaults because we were copying data memory past the end of our buffer.


daniel-j-h:
Thank you for fixing this! :+1:

PR #2544, opened 2016-06-14
Use BaseDataFacades in guidance and prevent ODR violations with inline

daniel-j-h:


TheMarex:
:+1: looks good to me merging.

Issue #2546, opened 2016-06-14
Service match omits the last point's coordinates in the geometry field.


rayras:
Service `match` omits the last point's coordinates in the geometry field.

curl "https://router.project-osrm.org/match/v1/driving/-3.870895,40.486134;-3.868323,40.484933;-3.868434,40.484077?geometries=geojson&overview=full"
```javascript
{
    "code": "Ok",
    "matchings": [
        {
            "confidence": 0.977545,
            "distance": 355.8,
            "duration": 25.5,
            "geometry": {
                "coordinates": [
                	[-3.870895, 40.486133 ], [-3.870728, 40.486055 ], [-3.870542, 40.485969 ], [-3.870008, 40.485724 ], [-3.869471, 40.485478 ], [-3.869186, 40.485348 ], [-3.86901, 40.485267 ], [-3.868374, 40.484975 ], [-3.868323, 40.484933 ],
                	[-3.868323, 40.484932 ], [-3.868366, 40.484814 ], [-3.868336, 40.484715 ], [-3.868264, 40.484645 ], [-3.868259, 40.48451 ]
                ],
                "type": "LineString"
            },
            "legs": [
                {"distance": 255.6, "duration": 18.3, "steps": [], "summary": ""},
                {"distance": 100.2, "duration": 7.2, "steps": [], "summary": ""}
            ]
        }
    ],
    "tracepoints": [
        {
            "hint": "xhhRg____38giwYATgAAAFoAAAALAAAAAAAAACF6KQUgeikFk7QAAFHvxP_1xGkCUe_E__bEaQIBAAEBRix7VA==",
            "location": [-3.870895, 40.486133 ],
            "matchings_index": 0,
            "name": "Avenida de España",
            "waypoint_index": 0
        },
        {
            "hint": "ojdPg____3-xqAYAAAAAAAoAAAAAAAAADwAAADr3AgNfTwgHk7QAAF35xP9EwGkCXfnE_0XAaQIAAAEBRix7VA==",
            "location": [-3.868323, 40.484932 ],
            "matchings_index": 0,
            "name": "Avenida de los Reyes Católicos",
            "waypoint_index": 1
        },
        {
            "hint": "ozdPg____3-xqAYAIwAAAEkAAAALAAAAkQAAACd6KQUcApUCk7QAAO74xP_tvGkC7vjE_-28aQIBAAEBRix7VA==",
            "location": [-3.868434, 40.484077 ],
            "matchings_index": 0,
[...]

PR #2549, opened 2016-06-14
Fix omitting the last point's coordinates in the geometry field

oxidase:
#2546 leg_index is not incremented, so the last point of the last leg was always missing if number of legs > 1

### DIFF
--- a/features/step_definitions/matching.js
+++ b/features/step_definitions/matching.js
@@ -1,5 +1,6 @@
 var util = require('util');
 var d3 = require('d3-queue');
+var polyline = require('polyline');
 
 module.exports = function () {
     this.When(/^I match I should get$/, (table, callback) => {
@@ -35,6 +36,7 @@ module.exports = function () {
                         route = '',
                         duration = '',
                         annotation = '',
+                        geometry = '',
                         OSMIDs = '';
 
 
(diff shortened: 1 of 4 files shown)


danpat:
@oxidase nice catch.  Do you have the bandwidth to make a test case for this?

oxidase:
@danpat added a test for geometry overviews. One thing that i don't get in the expectation result is the duplicated point b. Two geometries from assembleGeometry are [1,1;1.000089,1] and  [1.000089,1;1.000089,1;1.000089,0.99991], so the second leg has b as a start node and a middle node.

features/support/route.js
```diff
@@ -163,6 +163,10 @@ module.exports = function () {
         return instructions.legs.map(l => {return zip(l.annotation.duration, l.annotation.distance).map(p => { return p.join(':'); }).join(','); }).join(',');
     };
 
+    this.geometryList = (instructions) => {
+        return instructions.geometry.coordinates.map(p =>{ return p[0] + ',' + p[1]; }).join(';');
```
TheMarex: Relying on setting `geometries=geojson` here is bad. Easy to be messed up by the calling code. Instead I would suggest decoding the default (polyline) using https://github.com/mapbox/polyline

Also the parameter name `instructions` is misleading: it is a `Route` object.

oxidase: Ok, changed to decode geometry polyline response and compare decoded strings

TheMarex:
>  One thing that i don't get in the expectation result is the duplicated point b

Hm did you fix that? I observe this in the test case. Going to merge this to make it into `5.2.4`.

Issue #2550, opened 2016-06-15
No turn announced, but there probably should be


danpat:
On IRC in #osm, this route was pointed out:

http://map.project-osrm.org/?z=18&center=51.406740%2C-3.280218&loc=51.406314%2C-3.279966&loc=51.407274%2C-3.281205&hl=en&alt=0

No instruction is announced, but one was expected.

/cc @muzer

daniel-j-h:
For the record, testing this locally against master emits a "new name, slight right".

Looking a bit more into it, it looks like this area was changed a couple of hours ago in
http://www.openstreetmap.org/changeset/40029188

Here are the changes visualized
http://osmhv.openstreetmap.de/changeset.jsp?id=40029188

![before-after](https://cloud.githubusercontent.com/assets/527241/16074912/f5642786-32ec-11e6-8083-ccea8c6b2f7c.png)

(red before, green after)

The demo server still routes on old geometry as of this writing since we "only" update once a day. @MoKob is looking into the specific instruction generated.

MoKob:
While this is obviously an issue of the out-of-date data (give it a day until you can start to expect new data to be reflected in OSRM, the difference in the geometry gives the idea), it surfaced an error that I fixed in #2551. The turn onto Jenner Road itself, as modelled in the old data, is a nearly straight move onto a road, which is announced as a new-name instruction. 

With the new modelling, it should emit the correct turn left/turn right instructions. Closing here, as nothing actionable. Feel free to re-open if the problem should persist.

PR #2551, opened 2016-06-15
improve handling of obvious for end-of-road situations

MoKob:
Even though it turns out that https://github.com/Project-OSRM/osrm-backend/issues/2550 is a data issue that is using non-up-to-date mapping, it surfaced a problem in handling end-of-road situations.

This PR considers a situation in which a main road is turning and a small road is merging onto it.
The major road now does not print a turn-instruction to follow it.

See the testcase for the exact behaviour.

### DIFF
--- a/src/engine/guidance/post_processing.cpp
+++ b/src/engine/guidance/post_processing.cpp
@@ -419,6 +419,20 @@ void collapseTurnAt(std::vector<RouteStep> &steps,
                      current_step.name_id == steps[two_back_index].name_id)
             {
                 steps[one_back_index].maneuver.instruction.type = TurnType::Continue;
+
+                const auto getBearing = [](bool in, const RouteStep &step)
+                {
+                    const auto index =
+                        in ? step.intersections.front().in : step.intersections.front().out;
+                    return step.intersections.front().bearings[index];
+                };
+
+                // If we Merge onto the same street, we end up with a u-turn in some cases
+                if (bearingsAreReversed(
+                        util::bearing::reverseBearing(getBearing(true, one_back_step)),
+                        getBearing(false, current_step)))
+                    steps[one_back_index].maneuver.instruction.direction_modifier =
+                        DirectionModifier::UTurn;
             }
             else if (TurnType::Merge == one_back_step.maneuver.instruction.type &&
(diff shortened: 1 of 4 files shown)


src/engine/guidance/post_processing.cpp
```diff
@@ -414,6 +414,14 @@ void collapseTurnAt(std::vector<RouteStep> &steps,
                      current_step.name_id == steps[two_back_index].name_id)
             {
                 steps[one_back_index].maneuver.instruction.type = TurnType::Continue;
+                // If we Merge onto the same street, we end up with a u-turn in some cases
+                if (bearingsAreReversed(util::bearing::reverseBearing(
+                                            one_back_step.intersections.front()
```
TheMarex: - maybe add an alias for `one_back_step.intersections.front()` here. Could make this if more readable.
- move the `TurnType::Continue` assignment in the else block to make it explicit that these are exclusive.

src/engine/guidance/post_processing.cpp
```diff
@@ -627,7 +635,7 @@ std::vector<RouteStep> collapseTurns(std::vector<RouteStep> steps)
     for (std::size_t step_index = 1; step_index + 1 < steps.size(); ++step_index)
     {
         const auto &current_step = steps[step_index];
-        if( current_step.maneuver.instruction.type == TurnType::NoTurn )
+        if (current_step.maneuver.instruction.type == TurnType::NoTurn)
```
TheMarex: hrm it is weird that our `scripts/format.sh` did not complain about this.

TheMarex:
Looks good to me. Added some minor comments. After that this is good to merge. :+1:

src/engine/guidance/post_processing.cpp
TheMarex: @MoKob  to clarify:

the pattern currently is `var = default; if (condition) var = othervalue;` it is better to do `if (condition) var = othervalue else var = default;`

MoKob: The assignments target two different parts. The `type` is always changed, the `direction_modifier` only in case of the `if`

PR #2563, opened 2016-06-21
Fix/roundabout

MoKob:
Currently, routes like [this](http://map.project-osrm.org/?z=18&center=38.909811%2C-77.043715&loc=38.912003%2C-77.050831&loc=38.909277%2C-77.042516&hl=en&alt=0) can result in crashes an the first roundabout [see here](http://map.project-osrm.org/?z=18&center=38.912257%2C-77.049353&loc=38.913175%2C-77.051754&loc=38.911381%2C-77.048407&hl=en&alt=0) can be treated wrongly, while breaking the geometry.

screen shot 2016-06-21 at 14 40 38

Sadly, I've not been able so far to reproduce the failure in a cucumber issue.
Nevertheless, this PR enables a correct handling, by making sure roundabouts are treated correctly both in relation to intersection classifications and sliproads.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+# 5.2.4
+  - Bugfixes:
+    - Fixed in issue that arised on roundabouts in combination with intermediate intersections and sliproads
+
 # 5.2.3
   - Bugfixes:
     - Fixed an issue with name changes in roundabouts that could result in crashes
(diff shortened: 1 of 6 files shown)


MoKob:
An issue requesting a regression test can be found here: https://github.com/Project-OSRM/osrm-backend/issues/2564

PR #2565, opened 2016-06-21
Show user hint on wrong Node.js version, fixes #2435.

### LINKED ISSUE #2435: Better error output if wrong node version is used
If node v0.10.x is used, running `make -C test/data benchmark` will give:

```
/Users/dane/projects/osrm-backend/scripts/poly2req.js:9
let fs = require('fs');
^^^
SyntaxError: Unexpected strict mode reserved word
```

Ideally we should detect non-supported node versions and provide a friendly error.

daniel-j-h:
For https://github.com/Project-OSRM/osrm-backend/issues/2435.

This is more of a hack but looks like the best we can do.

The reason for this is the following:

1/ we require strict mode for our v4 script

2/ v0.10 in strict mode marks 'let' as a reserved keyword,
   aborting before execution **any** code in the script. Thus,
   we can't assert e.g. on process.versions['node'] !== '4'.

### DIFF
--- a/scripts/poly2req.js
+++ b/scripts/poly2req.js
@@ -2,7 +2,7 @@
 
 'use strict';
 
-let fs = require('fs');
+let fs = require('fs');  // Node 4.x required!
 
 let VERSION = "route_5.0";
 let SAMPLE_SIZE = 20;


TheMarex:
@daniel-j-h we can actually specify this in the package.json:

https://docs.npmjs.com/files/package.json#engines

daniel-j-h:
We already do this, see: https://github.com/Project-OSRM/osrm-backend/blob/b98431e8e63846672dd0bf1ea6bde6f1d63bdaf7/package.json#L34

I think @springmeyer's issue is more about catching the shell picking up the wrong node version.

PR #2566, opened 2016-06-21
Let Travis fail fast if CMake/Clang download fails, fixes #2433

### LINKED ISSUE #2433: [travis] if cmake download fails the build jobs should immediately error
Travis will error with `CMake 2.8.8 or higher is required.  You are running version 2.8.7` if cmake is not upgraded. I noticed that the upgrade may fail due to network issues, but still continues on without stopping:

```
$ if [[ "${TRAVIS_OS_NAME}" == "linux" ]]; then
    CMAKE_URL="http://www.cmake.org/files/v3.5/cmake-3.5.1-Linux-x86_64.tar.gz"
    mkdir cmake && travis_retry wget --quiet -O - ${CMAKE_URL} | tar --strip-components=1 -xz -C cmake
    export PATH=${DEPS_DIR}/cmake/bin:${PATH}
  elif [[ "${TRAVIS_OS_NAME}" == "osx" ]]; then
    # implicit deps, but seem to be installed by default with recent images: libxml2 GDAL boost
    brew install cmake libzip libstxxl lua51 luabind tbb md5sha1sum ccache
  fi
  
The command "wget --quiet -O - http://www.cmake.org/files/v3.5/cmake-3.5.1-Linux-x86_64.tar.gz" failed. Retrying, 2 of 3.
The command "wget --quiet -O - http://www.cmake.org/files/v3.5/cmake-3.5.1-Linux-x86_64.tar.gz" failed. Retrying, 3 of 3.
The command "wget --quiet -O - http://www.cmake.org/files/v3.5/cmake-3.5.1-Linux-x86_64.tar.gz" failed 3 times.
gzip: stdin: unexpected end of file
tar: Child returned status 1
tar: Error is not recoverable: exiting now
$ cd ${TRAVIS_BUILD_DIR}
... continues ...
```

daniel-j-h:
For https://github.com/Project-OSRM/osrm-backend/issues/2433.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -129,13 +129,13 @@ before_install:
   - DEPS_DIR="${TRAVIS_BUILD_DIR}/deps"
   - export PATH=${DEPS_DIR}/bin:${PATH} && mkdir -p ${DEPS_DIR}
   - CMAKE_URL="https://mason-binaries.s3.amazonaws.com/${TRAVIS_OS_NAME}-x86_64/cmake/3.5.2.tar.gz"
-  - travis_retry wget --quiet -O - ${CMAKE_URL} | tar --strip-components=1 -xz -C ${DEPS_DIR}
+  - travis_retry wget --quiet -O - ${CMAKE_URL} | tar --strip-components=1 -xz -C ${DEPS_DIR} || exit 1
   - |
     if [[ ${CLANG_VERSION:-false} != false ]]; then
       export CCOMPILER='clang'
       export CXXCOMPILER='clang++'
       CLANG_URL="https://mason-binaries.s3.amazonaws.com/${TRAVIS_OS_NAME}-x86_64/clang/${CLANG_VERSION}.tar.gz"
-      travis_retry wget --quiet -O - ${CLANG_URL} | tar --strip-components=1 -xz -C ${DEPS_DIR}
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -128,14 +128,14 @@ before_install:
   - npm install
   - DEPS_DIR="${TRAVIS_BUILD_DIR}/deps"
   - export PATH=${DEPS_DIR}/bin:${PATH} && mkdir -p ${DEPS_DIR}
-  - CMAKE_URL="https://mason-binaries.s3.amazonaws.com/${TRAVIS_OS_NAME}-x86_64/cmake/3.5.2.tar.gz"
-  - travis_retry wget --quiet -O - ${CMAKE_URL} | tar --strip-components=1 -xz -C ${DEPS_DIR}
+  - CMAKE_URL="https://mason-binaries.s3.amazonaws.com/${TRAVIS_OS_NAME}-x86_64/cmake/3.5.3.tar.gz"
```
daniel-j-h: Incrementing CMake version to check if it fails on the Travis build. If it does, I will decrement again and force push.

daniel-j-h:
Just tested this with an invalid URL on Travis (and then ammended this changeset): works as expected.

PR #2569, opened 2016-06-21
Fix raster bilinear interpolation

TheMarex:


TheMarex:
Supersedes https://github.com/Project-OSRM/osrm-backend/pull/2517

PR #2577, opened 2016-06-22
option to disable LTO

noblige:
Recreating https://github.com/Project-OSRM/osrm-backend/pull/2503 with minimal scope - with latest toolchain, this is the only change required to compile


daniel-j-h:
This is also great to have for incremental builds / development, since LTO is so darn slow. :+1: 

Can you add an `ENABLE_LTO` [at the top](
https://github.com/noblige/osrm-backend/blob/7e6d16e0b627d01a379352b2c5ea5d7b574b3b79/CMakeLists.txt#L48) (e.g. similar to `ENABLE_CCACHE`) and default it to `ON`.

noblige:
@daniel-j-h made it so

daniel-j-h:
:+1: thanks!

Issue #2579, opened 2016-06-23
Hard-abort on critical exceptions


daniel-j-h:
At the moment we catch-all exceptions at the top e.g. coming from `boost::numerical_cast`.

This gives us no information as to where the issue occured, all we get is a
> [warn] overflow happened

(or similar)

Here's what we should do:

- `-DENABLE_ASSERTIONS` already enables custom Boost assertion handler
- we already overwrite provide a custom assertion handler [here](https://github.com/Project-OSRM/osrm-backend/blob/6dedd9cb724633756602c23523ecc54d3e46c892/src/util/assert.cpp#L13)
- with catch-alls littered throughout the code we should either throw a different exception type and then not catch it (which then calls terminate, which calls abort) or hard abort ourselves directly

-----------------------------------------------------------------------------------------

![termination](https://raw.githubusercontent.com/adishavit/Terminators/master/termination_graph.png) (from https://github.com/adishavit/Terminators)

TheMarex:
This landed with #2579.

daniel-j-h:
For the record, it's only enabled with `-DENABLE_ASSERTIONS`. Without this (e.g. in release builds, node-osrm binaries, ..) we still simply swallow such errors..

PR #2580, opened 2016-06-23
Hard-abort on assertion and show detailed information, solves #2579

daniel-j-h:
For #2579.

From

> [warn] *swallows any useful information*

to

> [assert] /tmp/osrm-backend/src/extractor/extractor.cpp:79
> in: int osrm::extractor::Extractor::run(): false
> terminate called without an active exception

(I added an `BOOST_ASSERT(false)` to show-case this to the extractor)


TheMarex:
Looks good to me. :+1:

Issue #2548, opened 2016-06-14
Provide OSM node ids in `nearest` response


vkurchatkin:
It would be useful to have node ids of matched segment in `nearest` response. To be specific, that could be used to simplify building traffic data CSV from raw GPS data

TheMarex:
For an existing solution try using the `match` service. Since `5.2` we expose the OSM ids for each `RouteLeg` if you set `annotations=true`.

vkurchatkin:
@TheMarex already doing that, but I thought it would be nice to have an easier way to do this. If you are interested in such a feature, I would like to take a stab

danpat:
@vkurchatkin Go for it :-)

SiarheiFedartsou:
Done in https://github.com/Project-OSRM/osrm-backend/pull/4436/files

PR #2581, opened 2016-06-23
add osm nodes to `nearest` response

vkurchatkin:
To-do list:
- [ ] add tests
- [ ] update docs

Fixes: https://github.com/Project-OSRM/osrm-backend/issues/2548

@danpat please, take a look if you have time. Am I going in the right direction?

### DIFF
--- a/include/engine/api/nearest_api.hpp
+++ b/include/engine/api/nearest_api.hpp
@@ -38,8 +38,33 @@ class NearestAPI final : public BaseAPI
                        phantom_nodes.front().end(),
                        waypoints.values.begin(),
                        [this](const PhantomNodeWithDistance &phantom_with_distance) {
-                           auto waypoint = MakeWaypoint(phantom_with_distance.phantom_node);
+                           auto& phantom_node = phantom_with_distance.phantom_node;
+                           auto waypoint = MakeWaypoint(phantom_node);
                            waypoint.values["distance"] = phantom_with_distance.distance;
+
+                           util::json::Array nodes;
+
+                           std::vector<NodeID> reverse_geometry;
+                           facade.GetUncompressedGeometry(phantom_node.reverse_packed_geometry_id, reverse_geometry);
+                           auto from_node = facade.GetOSMNodeIDOfNode(reverse_geometry[reverse_geometry.size() - phantom_node.fwd_segment_position - 1]);
+
+                           std::vector<NodeID> forward_geometry;
+                           facade.GetUncompressedGeometry(phantom_with_distance.phantom_node.forward_packed_geometry_id, forward_geometry);
+                           auto to_node = facade.GetOSMNodeIDOfNode(forward_geometry[phantom_with_distance.phantom_node.fwd_segment_position]);
+
+                           if (!phantom_node.forward_segment_id.enabled)
+                           {
+                               nodes.values.push_back(static_cast<std::uint64_t>(to_node));
+                               nodes.values.push_back(static_cast<std::uint64_t>(from_node));
+                           }
+                           else
+                           {
+                               nodes.values.push_back(static_cast<std::uint64_t>(from_node));
+                               nodes.values.push_back(static_cast<std::uint64_t>(to_node));
+                           }
+
+                           waypoint.values["nodes"] = std::move(nodes);
+
                            return waypoint;
                        });
 


danpat:
@vkurchatkin You're on the right track.  Need to make a couple of changes though:

The `forward_packed_geometry` and `reverse_packed_geometry` are basically duplicates, but reverse of each other, you only need to fetch one.

That said, `forward_packed_geometry_id` and `reverse_packed_geometry_id` could be set to `SPECIAL_EDGEID`, which means they're invalid.  You'll need to check for that.  This could happen for oneways - only one direction is valid.

There is a question here as to what order you return the node IDs in.  I think it probably doesn't matter.  I would check first the `forward_packed_geometry_id != SPECIAL_EDGE_ID`, and if that fails, try the `reverse_packed_geometry_id`.  Take a look at the tile plugin here:

https://github.com/Project-OSRM/osrm-backend/blob/master/src/engine/plugins/tile.cpp#L294-L324

This shows checking the forward and reverse directions, and the different logic needed to find the array position depending on the direction.  The `PhantomNode` only has the `fwd_segment_position`, so the calculation is different when you're finding it in the `reverse_packed_geometry_id`.  I see you've already basically got that logic in there.

danpat:
One further note:  the one difference between the forward and reverse packed geometries is that they're missing the *first* node in the sequence.  The packed geometries contain nodes 1 to N.  If the `fwd_segment_position` is 0, then you'll to do some spelunking to find the first (or last) node ID.

vkurchatkin:
> The forward_packed_geometry and reverse_packed_geometry are basically duplicates, but reverse of each other, you'll only need to fetch one to get the two nodes you need.

Got it. To be honest

> If the fwd_segment_position is 0, then you'll to do some spelunking

Hmm, I guess if this is the case and there is no reverse geometry, there is not much I can do?

> That said, forward_packed_geometry_id and reverse_packed_geometry_id could be set to SPECIAL_EDGEID, which means they're invalid. You'll need to check for that. 

I assumed that I could check `segment_id.enabled` instead, is that correct?

> There is a question here as to what order you return the node IDs in. I think it probably doesn't matter.

It matters if `bearings` parameter is provided or segment is one way. I specifically implemented it by checking `phantom_node.forward_segment_id.enabled`.

TheMarex:
@vkurchatkin This looks good to me.

> - add parameter to calculate nodes;

I think it is fine if nodes are returned by default.

Do you want to push this over the finish line so we could include it in `5.4.0`.

vkurchatkin:
> I think it is fine if nodes are returned by default.

ok!

> Do you want to push this over the finish line so we could include it in 5.4.0.

I guess only testing is left then. When is the deadline for 5.4.0?

TheMarex:
> I guess only testing is left then. When is the deadline for 5.4.0?

There is no deadline, but I would expect a release in the coming weeks. This is of course not critical, but would love to see this included, I think it makes a good addition.

MoKob:
Moving this to 5.5 since its missing tests and documentation, yet. We are pushing out 5.4 to include the many changes we did the past weeks. 5.5 will not be far off, though.

daniel-j-h:
Ping. What's the state of this?

It's scheduled for 5.5 but there's some work left to do.

I'd love to get it into the next release.

vkurchatkin:
@daniel-j-h I'll try to finish over the weekend. Sorry for the delay!

daniel-j-h:
Just letting you know we're planing to do the next release (5.5) at the end of the month. We can move it to 5.6 if you need more time for it but keep in mind 5.6 will then probably take until next year.

daniel-j-h:
Ping - what's the state of this? @vkurchatkin still interested in working on this?

vkurchatkin:
@daniel-j-h working on it right now. A lot has changed so it doesn't compile yet. Also have to wrap my head around @danpat's comments

vkurchatkin:
As far as I can see there is just `packed_geometry_id` now so @danpat's feedback is mostly outdated

quinnchr:
@vkurchatkin What is the purpose of returning two node ids here? For most use cases wouldn't the closest node be the only one of interest?

danpat:
@quinnchr The two node IDs represent the start/end of the line segment that the node snapped to.  You can be quite close to a line, but quite far from the endpoints.

daniel-j-h:
Ping - what's the state of this? It's scheduled for the 5.7 release. Still actionable?

- [Here](https://github.com/Project-OSRM/osrm-backend/blob/273fd689ce531ff7eea0b671dc64435ba3b71c6f/include/engine/phantom_node.hpp#L177) is the `packed_geometry_id`
- [Here](https://github.com/Project-OSRM/osrm-backend/blob/a572719484cd86dc066500af92ac3270460b0ae1/include/engine/guidance/assemble_geometry.hpp#L41-L57) and below is an example how to go from `PhantomNode` to OSM Id

vkurchatkin:
Closing in favour of https://github.com/Project-OSRM/osrm-backend/pull/4436

PR #2582, opened 2016-06-24
Use mmap instead of read during edge weight updates.

danpat:
This changes the edge-weight-update loop to use `mmap`-ed data instead of using `read` on `::ifstream` objects.

When processing California-sized datasets, this change reduces the edge-weight-update loop from 7 seconds to 1.4 seconds.   On larger datasets where we see 150+ second durations for this code block, this should have a decent impact on the time it takes to turn around a dataset.

The file formats *should* remain compatible between releases here, hence only the minor version bump.

### DIFF
--- a/include/extractor/edge_based_graph_factory.hpp
+++ b/include/extractor/edge_based_graph_factory.hpp
@@ -43,6 +43,34 @@ namespace osrm
 namespace extractor
 {
 
+namespace lookup
+{
+// Set to 1 byte alignment
+struct SegmentHeaderBlock
+{
+    std::uint32_t num_osm_nodes;
+    OSMNodeID previous_osm_node_id;
+} __attribute ((packed));
+static_assert(sizeof(SegmentHeaderBlock) == 12, "SegmentHeaderBlock is not packed correctly");
+
+struct SegmentBlock
+{
+    OSMNodeID this_osm_node_id;
+    double segment_length;
+    std::int32_t segment_weight;
+} __attribute ((packed));
+static_assert(sizeof(SegmentBlock) == 20, "SegmentBlock is not packed correctly");
+
+struct PenaltyBlock
+{
+    std::uint32_t fixed_penalty;
+    OSMNodeID from_id;
+    OSMNodeID via_id;
+    OSMNodeID to_id;
+} __attribute ((packed));
+static_assert(sizeof(PenaltyBlock) == 28, "PenaltyBlock is not packed correctly");
+}
+
 class EdgeBasedGraphFactory
 {
   public:
--- a/src/contractor/contractor.cpp
+++ b/src/contractor/contractor.cpp
@@ -3,6 +3,7 @@
 #include "contractor/graph_contractor.hpp"
 
 #include "extractor/compressed_edge_container.hpp"
+#include "extractor/edge_based_graph_factory.hpp"
 #include "extractor/node_based_edge.hpp"
 
 #include "util/exception.hpp"
@@ -252,7 +253,7 @@ parse_segment_lookup_from_csv_files(const std::vector<std::string> &segment_spee
                 throw util::exception{"Segment speed file " + filename + " malformed"};
 
             SegmentSpeedSource val{
-                {static_cast<OSMNodeID>(from_node_id), static_cast<OSMNodeID>(to_node_id)},
+                {OSMNodeID{from_node_id}, OSMNodeID{to_node_id}},
                 {speed, static_cast<std::uint8_t>(file_id)}};
 
             local.push_back(std::move(val));
@@ -338,7 +339,7 @@ parse_turn_penalty_lookup_from_csv_files(const std::vector<std::string> &turn_pe
                 throw util::exception{"Turn penalty file " + filename + " malformed"};
 
             map[std::make_tuple(
-                OSMNodeID(from_node_id), OSMNodeID(via_node_id), OSMNodeID(to_node_id))] =
+                OSMNodeID{from_node_id}, OSMNodeID{via_node_id}, OSMNodeID{to_node_id})] =
                 std::make_pair(penalty, file_id);
         }
     };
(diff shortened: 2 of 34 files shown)


include/extractor/edge_based_graph_factory.hpp
```diff
@@ -42,6 +42,33 @@ namespace osrm
 namespace extractor
 {
 
+namespace lookup
+{
+// Set to 1 byte alignment
+#pragma pack(push, r1, 1)
```
daniel-j-h: Care to explain?

src/contractor/contractor.cpp
```diff
@@ -366,36 +367,56 @@ EdgeID Contractor::LoadEdgeExpandedGraph(
         throw util::exception("Limit of 255 segment speed and turn penalty files each reached");
 
     util::SimpleLogger().Write() << "Opening " << edge_based_graph_filename;
-    boost::filesystem::ifstream input_stream(edge_based_graph_filename, std::ios::binary);
-    if (!input_stream)
-        throw util::exception("Could not load edge based graph file");
+
+    const boost::interprocess::file_mapping edge_based_graph_mapping{
```
daniel-j-h: Maybe just do

    using boost::interprocess::file_mapping;
    ...

otherwise this is super verbose and makes reading hard with our small column length.

src/contractor/contractor.cpp
```diff
@@ -366,36 +367,56 @@ EdgeID Contractor::LoadEdgeExpandedGraph(
...
+        edge_based_graph_filename.c_str(), boost::interprocess::read_only};
```
daniel-j-h: The following code is duplicated at least three times. Can we abstract over it providing a lambda / function that takes a file name and a madvise hint, and returns a region?

src/contractor/contractor.cpp
```diff
@@ -366,36 +367,56 @@ EdgeID Contractor::LoadEdgeExpandedGraph(
...
+            return region;
         }
-    }
+        return boost::interprocess::mapped_region();
+    }();
 
     const util::FingerPrint fingerprint_valid = util::FingerPrint::GetValid();
-    util::FingerPrint fingerprint_loaded;
-    input_stream.read((char *)&fingerprint_loaded, sizeof(util::FingerPrint));
+    const util::FingerPrint fingerprint_loaded = *(reinterpret_cast<const util::FingerPrint *>(
```
daniel-j-h: In the following you're using `get_address` a lot for first / last pointers.
Maybe store const first / last pointers once and then use those.

src/contractor/contractor.cpp
```diff
@@ -366,36 +367,56 @@ EdgeID Contractor::LoadEdgeExpandedGraph(
...
+        reinterpret_cast<const char *>(edge_based_graph_region.get_address()) +
+        edge_based_graph_pos));
     fingerprint_loaded.TestContractor(fingerprint_valid);
+    edge_based_graph_pos += sizeof(util::FingerPrint);
```
daniel-j-h: Instead of pos, you can just increment `first` can't you?

Store

    auto first = ..
    const auto last = ..

once and advance first.

src/contractor/contractor.cpp
```diff
@@ -713,92 +734,84 @@ EdgeID Contractor::LoadEdgeExpandedGraph(
...
+        extractor::EdgeBasedEdge inbuffer = *edge_based_edge_ptr;
+        edge_based_edge_ptr++;
+
         if (update_edge_weights || update_turn_penalties)
         {
-            // Processing-time edge updates
-            unsigned fixed_penalty;
-            edge_fixed_penalties_input_stream.read(reinterpret_cast<char *>(&fixed_penalty),
-                                                   sizeof(fixed_penalty));
+            auto header = reinterpret_cast<const extractor::lookup::SegmentHeaderBlock *>(
```
daniel-j-h: This should be `const auto*`

src/contractor/contractor.cpp
```diff
@@ -713,92 +734,84 @@ EdgeID Contractor::LoadEdgeExpandedGraph(
...
+                edge_segment_byte_ptr);
+            edge_segment_byte_ptr += sizeof(extractor::lookup::SegmentHeaderBlock);
 
+            auto previous_osm_node_id = header->previous_osm_node_id;
             int new_weight = 0;
+            int compressed_edge_nodes = static_cast<int>(header->num_osm_nodes);
+
+            auto segmentblocks =
```
daniel-j-h: Same here

src/contractor/contractor.cpp
```diff
@@ -713,92 +734,84 @@ EdgeID Contractor::LoadEdgeExpandedGraph(
...
-            OSMNodeID previous_osm_node_id;
-            edge_segment_input_stream.read(reinterpret_cast<char *>(&previous_osm_node_id),
-                                           sizeof(previous_osm_node_id));
-            OSMNodeID this_osm_node_id;
-            double segment_length;
-            int segment_weight;
-            int compressed_edge_nodes = static_cast<int>(num_osm_nodes);
-            --num_osm_nodes;
-            for (; num_osm_nodes != 0; --num_osm_nodes)
+            for (unsigned i = 0; i < header->num_osm_nodes - 1; i++)
```
daniel-j-h: Why `-1`, can you elaborate.

daniel-j-h:
:+1: nice work Danpat!

include/extractor/edge_based_graph_factory.hpp
danpat: This is a trick to avoid wasted bytes in the files on disk.

https://en.wikipedia.org/wiki/Data_structure_alignment#Default_packing_and_.23pragma_pack

`#pragma pack` allow you to set the structure word alignment, and here I'm setting it to 1 byte, so there will be no padding.

The downside to doing this is that operations on structs that are not word aligned in memory could mean extra memory access and some calculation to join un-aligned data.

In this particular case, that overhead is dwarfed by the I/O we're doing, and for the data structures that aren't properly aligned, saves us a bunch of bytes on disk.  `SegmentBlock` is the only one that gets some padding in the default word alignment on x86-64 (4 extra bytes).

I'll add a clearer explanation as to why this is done.

src/contractor/contractor.cpp
danpat: We can only advance by a fixed amount if we do that - we're reading several different variable types from the file here, of different sizes.

I *could* bundle these "header" values into a temporary struct and grab them with one read.

src/contractor/contractor.cpp
danpat: I've seen your screen, you have a tiny font :-)

include/extractor/edge_based_graph_factory.hpp
daniel-j-h: Yeah sure, but I can't seem to find the `pack(push, r1, 1)` syntax:
https://gcc.gnu.org/onlinedocs/gcc-6.1.0/gcc/Structure-Layout-Pragmas.html

I guess it's just a named stack to push / pop values on / off (?). Maybe just do

    struct __attribute__ ((packed))

on the one struct where it is needed and `static_assert` on all of the struct's sizes.

(Unrelated: it seems like GCC 6+ supports a pragma for endianess. Interesting.)

TheMarex: If there is a non-pragma approach I would prefer that. Also static_assert on size is important to document platform specific breakages.

Issue #2533, opened 2016-06-11
roundabout exist counts are confusing for cyclists


emiltin:
Because you can usually push bikes against oneways on bike, dual carriageways are counted as 2 exists in roundabouts, leading to confusing exit counts.

For example you're told to take the 3rd exit in the following situation:

![screen shot 2016-06-11 at 23 11 58](https://cloud.githubusercontent.com/assets/66034/15987483/1a98d0e4-302a-11e6-800f-abb3e26c6032.png)

I'm not sure what can be done about this. OSRM could count only exist that you can cycle on. This would make the common case of dual carriageways more correct, but would require special handling of edge cases like exiting on a footway.

emiltin:
related to #2529. ie you could maybe tell the user "At roundabout, head straight on Birch Street" or "At roundabout, head left onto Summer Street",  instead of counting exits

MoKob:
The problem I see here is how would you tell someone to exit into one of these streets. Consider a foot-path that is actually your destination. If you don't count it, you have either an undefined exit or, if you then simply add 1, `exit two` can both be the path where you need to push and the actual street.

What I could imagine is that we can treat roundabouts as we do with cars, not allowing exits against the direction, and have to announce something without an exit like `exit the roundabout and push your bike` for all exits that don't follow the convention.

This will require both special treatment in pre and post processing, since we need to handle cases that exit the roundabout at a exit that is not part of the roundabout and we need to consider travel modes of bikes in roundabouts.
This needs https://github.com/Project-OSRM/osrm-backend/issues/2113 to be implemented reasonably without introducing yet more profile dependent settings.

emiltin:
the demo server seems to have a problem t the moment (no route is computed), but i was looking at these three roundabouts. there's clear main way that continues straight ahead in all of them.

http://map.project-osrm.org/?z=15&center=55.655110%2C12.587092&loc=55.659650%2C12.583508&loc=55.653016%2C12.588658&hl=en&alt=0

emiltin:
Right, it's not obvious how to best guide cyclist in these situations. The question is how meaningful it is to count exits if you can exit pushing bikes against oneways or at footways.

For small roundabouts, simply announcing a normal turn is probably better, and is already done with the roundabout turn instruction.

For normal/large roundabouts, maybe it would be better to announce the exit as a normal turn? Ie. in 20m exit the roundabout onto ...?

I guess pedestrian routing have the same issue with roundabouts.

[...]

Issue #2527, opened 2016-06-10
should not fork slight left when road goes straight


emiltin:
osrm emits a fork left instruction when the way continues straight ahead (same bearing before/after).

it should probably suppress the instruction (if the forking road is minor) or tell you to go straight at the fork.

situation:
![screen shot 2016-06-10 at 13 51 06](https://cloud.githubusercontent.com/assets/66034/15963455/78b71b04-2f12-11e6-9e40-fcf96f214234.png)

response:
```json
{
  "waypoints": [{
    "location": [12.620201, 55.661265],
    "hint": "HgsAgP___38zBAAAhgAAAPIAAAAAAAAAAAAAAMs9AAAQaQAAVwcAAKmRwADRUlEDqJHAANZSUQMAAAYGwiCf8g==",
    "name": "Øresundsvej"
  }, {
    "location": [12.6194, 55.661196],
    "hint": "fzMAgP___38zBAAAxAAAAAYBAAAAAAAAAAAAAA9pAAAKZQAAVwcAAIiOwACMUlEDiI7AAJBSUQMAAAYGwiCf8g==",
    "name": "Øresundsvej"
  }],
  "routes": [{
    "distance": 50.8,
    "duration": 30.4,
    "legs": [{
      "distance": 50.8,
      "steps": [{
        "distance": 18,
        "duration": 10.8,
        "name": "Øresundsvej",
        "maneuver": {
          "type": "depart",
          "location": [12.620201, 55.661265],
          "bearing_before": 0,
          "bearing_after": 261
        },
        "geometry": "}ivrIg{_lADv@",
        "mode": "pushing bike",
        "intersections": [{
          "location": [12.620201, 55.661265],
          "bearings": [261],
          "entry": [true],
          "out": 0
        }]
      }, {
        "distance": 32.9,
        "duration": 19.6,
        "name": "Øresundsvej",
        "maneuver": {
          "type": "fork",
          "modifier": "slight left",
          "bearing_before": 261,
          "bearing_after": 261,
          "location": [12.619918, 55.661241]
        },
        "geometry": "wivrIoy_lAFfB",
        "mode": "pushing bike",
        "intersections": [{
          "location": [12.619918, 55.661241],
          "bearings": [75, 255, 300],
          "entry": [false, true, true],
          "in": 0,
          "out": 1
        }]
      }, {
        "distance": 0,
        "duration": 0,
        "name": "Øresundsvej",
        "maneuver": {
[...]

Issue #2113, opened 2016-03-18
Move Guidance Decisions into Profile


MoKob:
The new static turn instructions include a few hard-coded decisions that need to be moved out into the profile.
- [x] [Fork discovery](https://github.com/Project-OSRM/osrm-backend/blob/features/guidance-features-5-0/include/extractor/guidance/toolkit.hpp#L391-L407): Forks can only happen if one road isn't clearly superior to the other. The current priority solution is very tentative. We should move the discovery function to the profiles in the profile overhaul to make sure these can be set for different preferences
- [x] [Road Classification and Ramp Discovery](https://github.com/Project-OSRM/osrm-backend/blob/features/guidance-features-5-0/include/extractor/guidance/classification_data.hpp#L21-L48) Most road classifications can have impact on how to treat a road. We treat motorways and ramps differently from other types, although only trunk and motorway ramps are treated as actual ramps. Both the road classification and what to treat as a ramp should be moved to the profile as well

We could also discuss in how far we want to allow configuration of [internal constants](https://github.com/Project-OSRM/osrm-backend/blob/features/guidance-features-5-0/include/extractor/guidance/classification_data.hpp#L21-L48).
- [ ] internal constants
- [ ] further hard-coded decisions?

For Bikes/and or Walking we have to consider further improvements:
- [ ] allow different classifications per direction (e.g. classification for pushing over cycling)

emiltin:
Good idea. Fore example bike infrastructure can also include forks and merging.

MoKob:
A first step is implemented in https://github.com/Project-OSRM/osrm-backend/pull/2586. It allows to set a fixed number of road classes to experiment with desired behaviour.
We can define link-classes, what is seen as motorway-like, roads that offer only connectivity.

daniel-j-h:
- [ ] Threshold for [Staggered Intersections](https://github.com/Project-OSRM/osrm-backend/pull/2738)
- [ ] Threshold for [Lane Anticipation](https://github.com/Project-OSRM/osrm-backend/blob/7f71f0ed120e11ada57496da16e3f3c17f80ee13/include/engine/guidance/lane_processing.hpp#L23)

oxidase:
Copied from #4075 
> the current behaviour does not work for foot and bike routing. even for cars i'm not sure i think the current behaviour makes sense, e.g. in the example no matter what exit you take, you get 'take 1st exit':
[...]

PR #2586, opened 2016-06-24
[ready] move guidance decisions into profiles

MoKob:
With the list of advanced guidance features getting longer, the amount of decisions that has been hard-coded in advance grows.
While it is not possible right now to put all decisions into configuration files, at least some information can be listed.

This PR aims at moving as much as possible into the profiles to make guidance more configurable and adjustable for walking/biking/driving.

It is related to problems seen in: 
 - https://github.com/Project-OSRM/osrm-backend/issues/2527
 - https://github.com/Project-OSRM/osrm-backend/issues/2533
 - implements: https://github.com/Project-OSRM/osrm-backend/issues/2113


include/extractor/guidance/classification_data.hpp
```diff
@@ -18,63 +15,129 @@ namespace extractor
...
-    TERTIARY,
-    TERTIARY_LINK,
-    UNCLASSIFIED,
-    RESIDENTIAL,
-    SERVICE,
-    LIVING_STREET,
-    LOW_PRIORITY_ROAD // a road simply included for connectivity. Should be avoided at all cost
-};
+    // a class that behaves like a motorway (separated directions)
+    bool motorway_class : 1;
```
TheMarex: Bitfields and bool's are apparently not cross-platform compatible (read: non-gcc this will break). The easiest solution that I have seen is to use `std::uint32_t` for all members that get packed into a 32bit block.

include/extractor/guidance/classification_data.hpp
```diff
@@ -18,63 +15,129 @@ namespace extractor
...
+        return std::string() + (motorway_class ? "motorway" : "normal") +
+               (link_class ? "_link" : "") + (may_be_ignored ? " ignorable" : " important") +
+               std::to_string(priority);
+    }
+
+    inline void set(const osmium::Way &way)
+    {
+        const static auto class_hash = [] {
+            std::unordered_map<std::string, unsigned> hash;
+            hash["motorway"] = 0;
```
TheMarex: What is your thinking behind dropping the enums?

TheMarex:
This is a very good idea. :+1:

include/extractor/guidance/classification_data.hpp
MoKob: The problem with the enums is that this is a intermediate step (to test if the main thing still works).
In the end, it should move to the profile where we need to handle the osm-strings and these respective flags should be set by the tags.

MoKob:
One problem I see right now is the road priorities. I've provided enums to give an indication of how they are used right now and what kind of distinctions are made.
The addition of roads for bicycles already improves some behaviour which is currently far from optimal (e.g. #2527), while maintaining same behaviour on cars.

In upcoming PRs (#2113), we will have to address further issues and potentially revisit the exposed values.

docs/profiles.md
```diff
@@ -32,3 +32,16 @@ Given an OpenStreetMap way, the way_function will either return nothing (meaning
 All other calculations stem from that, including the returned timings in driving directions, but also, less directly, it feeds into the actual routing decisions the engine will take (a way with a slow traversal speed, may be less favoured than a way with fast traversal speed, but it depends how long it is, and... what it connects to in the rest of the network graph)
 
 Using the power of the scripting language you wouldn't typically see something as simple as a `result.forward_speed = 20` line within the way_function. Instead a way_function will examine the tagging (e.g. `way:get_value_by_key("highway")` and many others), process this information in various ways, calling other local functions, referencing the global variables and look-up hashes, before arriving at the result.
+
+## Guidance
+
+The guidance parameters in profiles are currently a work in progress. They can and will change without any major version change.
```
TheMarex: This is not how semantic versioning works. If we break the API we need to do a version bump. Please remove this paragraph.

profiles/lib/guidance.lua
```diff
@@ -0,0 +1,88 @@
+local Guidance = {}
+
+-- Guidance: Default Mapping from roads to types/priorities
+highway_classes = { ["motorway"] = road_priority_class.motorway, ["motorway_link"] = road_priority_class.link_road, ["trunk"] = road_priority_class.trunk, ["trunk_link"] = road_priority_class.link_road,
```
TheMarex: I think this would be a little bit more readable as a table, maybe reformat it?

include/extractor/guidance/road_classification.hpp
```diff
@@ -0,0 +1,107 @@
...
+    // default construction
+    RoadClassification() : motorway_class(0), link_class(0), may_be_ignored(1), road_priority_class(RoadPriorityClass::CONNECTIVITY) {}
+
+    RoadClassification(bool motorway_class, bool link_class, bool may_be_ignored, RoadPriorityClass::Enum road_priority_class)
+        : motorway_class(motorway_class), link_class(link_class), may_be_ignored(may_be_ignored),
+          road_priority_class(road_priority_class)
+    {
+    }
+
+    inline bool isMotorwayClass() const { return (0 != motorway_class) && (0 == link_class); }
```
TheMarex: Member functions need to start with a capital letter to be consistent with our other classes.

src/extractor/extractor_callbacks.cpp
```diff
@@ -140,12 +141,8 @@ void ExtractorCallbacks::ProcessWay(const osmium::Way &input_way, const Extracti
...
 
     // FIXME this need to be moved into the profiles
-    const char *data = input_way.get_value_by_key("highway");
-    guidance::RoadClassificationData road_classification;
-    if (data)
-    {
-        road_classification.road_class = guidance::functionalRoadClassFromTag(data);
-    }
+    const guidance::RoadClassification road_classification = parsed_way.road_classification;
+    std::cout << "Classification: " << road_classification.toString() << std::endl;
```
TheMarex: Left-over debugging string?

TheMarex:
Looks good to me. But needs a changelog entry because the profiles now depend on a new file `guidance.lua` which needs to be added to our profile update scripts (like in node-osrm).

include/extractor/guidance/road_classification.hpp
```diff
@@ -0,0 +1,121 @@
...
+    std::uint32_t motorway_class : 1;
+    // all types of link classes
+    std::uint32_t link_class : 1;
+    // a low priority class is a pure connectivity way. It can be ignored in multiple decisions
+    // (e.g. fork on a primary vs service will not happen)
+    std::uint32_t may_be_ignored : 1;
+    // the road priority is used as an indicator for forks. If the roads are of similar priority
+    // (difference <=1), we can see the road as a fork. Else one of the road classes is seen as
+    // obvious choice
+    RoadPriorityClass::Enum road_priority_class : 5;
```
daniel-j-h: If we bitpack this, it's a good idea to static assert on the size

include/extractor/guidance/road_classification.hpp
```diff
@@ -0,0 +1,121 @@
...
+               may_be_ignored == other.may_be_ignored &&
+               road_priority_class == other.road_priority_class;
+    }
+
+    inline std::string ToString() const
+    {
+        return std::string() + (motorway_class ? "motorway" : "normal") +
+               (link_class ? "_link" : "") + (may_be_ignored ? " ignorable " : " important ") +
+               std::to_string(road_priority_class);
+    }
```
daniel-j-h: as discussed in chat a couple of days ago: member functions defined inside a class are inline by default

include/extractor/guidance/road_classification.hpp
MoKob: done

include/extractor/guidance/road_classification.hpp
MoKob: adressed.

MoKob:
With the current rebase, some of the anticipate lane-tests are failing (even though they should be mostly unaffected).

Holding off on fixing this until https://github.com/Project-OSRM/osrm-backend/pull/2642 and similar by @daniel-j-h have landed.

TheMarex:
Rebased on master, which results in broken cucumber tests.

TheMarex:
Never mind. This seems to have been some broken local caching.

PR #2587, opened 2016-06-24
reduce number of end-of-road instructions in unnecessary cases

MoKob:
screen shot 2016-05-24 at 15 40 50

Sometimes, end-of-road can feel way to verbose. If there has not been at least a single intersection in between, the benefit gained from `end of road` is slim.

This PR reduces the verbosity of our current guidance output. The T-intersection itself could, if necessary, still be deduced from the by now available intersection classifications.


TheMarex:
Looks good! merging.

PR #2588, opened 2016-06-25
bike: surface should never increase speed

emiltin:
The bike profile sets the speed according to the surface tag, even if the speed is faster than what's already determined by the way type.

This cause incorrect behaviour where you push the bike (like footways and oneways and surface=asphalt, since the speed will end up being full bike speed. This again to incorrect routing in a lot cases.

This PR fixes the problem and adds a cucumber test (which fails without the fix).

The updated behaviour mirrors the car profile.


TheMarex:
Looks good! Thanks @emiltin!

Issue #2590, opened 2016-06-26
way:id() from lua causes Abort trap: 6


emiltin:
Getting the OSM way id from lua used to work, but now I can't get it to work anymore with v5.2.6. Using this way function:

```lua
function way_function (way, result)
  local id = way:id()
end
```

I get this error when extracting:

```bash
~/code/osrm-backend/sandbox$ (green_v5 *) ../build/osrm-extract mid.osm.pbf --profile ../profiles/testbot.lua
[info] Using script ../profiles/testbot.lua
[info] Input file: mid.osm.pbf
[info] Profile: testbot.lua
[info] Threads: 4
[STXXL-MSG] STXXL v1.4.1 (prerelease/Release)
[STXXL-ERRMSG] Warning: no config file found.
[STXXL-ERRMSG] Using default disk configuration.
[STXXL-MSG] Disk '/var/tmp/stxxl' is allocated, space: 1000 MiB, I/O implementation: syscall autogrow delete_on_exit queue=0 devid=0
[info] Parsing in progress..
[info] input file generated by 0.44.1
[info] timestamp: n/a
[info] Found no exceptions to turn restrictions
libc++abi.dylib: terminating with uncaught exception of type osrm::util::exception: ERROR occurred in profile script:
std::runtime_error: 'Trying to use unregistered class'
Abort trap: 6
```

scripting_environment.cpp does seem to link id() to an osmium method:
```c++
         luabind::class_<osmium::Way>("Way")
             .def("get_value_by_key", &osmium::Way::get_value_by_key)
             .def("get_value_by_key", &get_value_by_key<osmium::Way>)
             .def("id", &osmium::Way::id)
             .def("get_nodes", get_nodes_for_way, luabind::return_stl_iterator),
```

Am I missing something?

TheMarex:
@emiltin I think we had an issue about this a while ago and it basically boiled down to a wrong lua/luabind version. I can't seem to find the ticket :-/

@daniel-j-h do you remember the specifics?

emiltin:
lua versions, sigh... what version is supposed to work, 5.1? i did:

```bash
~$ brew install luabind    # installs lua5.1
~$ luarocks-5.1 install luasql-postgres
~$(green_v5 *) lua
-bash: /usr/local/bin/lua: No such file or directory
~$ (green_v5 *) lua5.1
Lua 5.1.5  Copyright (C) 1994-2012 Lua.org, PUC-Rio
>
```

[...]

PR #2593, opened 2016-06-27
[Don't Merge] Let Travis test way:id() inside the way_function

daniel-j-h:
For https://github.com/Project-OSRM/osrm-backend/issues/2590.

Delete once we have Travis results.

### DIFF
--- a/profiles/car.lua
+++ b/profiles/car.lua
@@ -218,7 +218,8 @@ function node_function (node, result)
 end
 
 function way_function (way, result)
-  io.write(way:id())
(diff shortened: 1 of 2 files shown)


MoKob:
@daniel-j-h with the issue resolved, I guess we can remove this PR / branch?

daniel-j-h:
"resolved" as in Luabind is to blame. See comments in https://github.com/Project-OSRM/osrm-backend/issues/2590 for why this is happening.

tl;dr: brew ships a luabind version from 2010.

PR #2595, opened 2016-06-27
[review] handle sliproads at traffic lights

MoKob:
To handle sliproads, we are currently comparing the upcoming intersection for the straightmost turn and any other turn.
If we encounter a traffic light, the straightmost intersection does not represent the actual intersection we want.
In cases like [here](http://map.project-osrm.org/?z=18&center=48.991729%2C8.403283&loc=48.991883%2C8.402739&loc=48.991412%2C8.403854&hl=en&alt=0), this can result in two instructions instead of one.

screen shot 2016-06-27 at 17 12 43

This PR aims at reducing the verbosity of OSRM during these kinds of situations.

Desired output: `Turn right onto Ettlinger Allee`

### DIFF
--- a/features/guidance/dedicated-turn-roads.feature
+++ b/features/guidance/dedicated-turn-roads.feature
@@ -57,20 +57,21 @@ Feature: Slipways and Dedicated Turn Lanes
 
     Scenario: Inner city expressway with on road
         Given the node map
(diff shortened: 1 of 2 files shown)


TheMarex:
Looks good to me. Merging for rc.1

Issue #2594, opened 2016-06-27
Luabind can't find trimLaneString, throws exception in osrm-extract


nigun:
ubuntu 14.04 LTS
map->http://data.gis-lab.info/osm_dump/dump/latest/RU-MOS.osm.pbf
running: ln -s ../profiles/car.lua profile.lua
running: ./osrm-extract -p profile.lua ./RU-MOS.osm.pbf 
i have error:
[info] Using script profile.lua
[info] Input file: RU-MOS.osm.pbf
[info] Profile: profile.lua
[info] Threads: 4
[STXXL-MSG] STXXL v1.3.1 (release)
[STXXL-ERRMSG] Warning: no config file found.
[STXXL-ERRMSG] Using default disk configuration.
[STXXL-MSG] 1 disks are allocated, total space: 1000 MiB
[info] Parsing in progress..
[info] input file generated by 0.44.1
[info] timestamp: n/a
[info] Found 3 exceptions to turn restrictions:
[info]   motorcar
[info]   motor_vehicle
[info]   vehicle
terminate called after throwing an instance of 'osrm::util::exception'
  what():  ERROR occurred in profile script:
No matching overload found, candidates:
std::string trimLaneString(std::string,int,int)
Aborted (core dumped)

daniel-j-h:
Can you be more specific about
- osrm-backend version
- compiler and stdlib version
- operating system
- lua and luabind version
- ldd libluabind.so to see which lua version luabind was linked against
- the cmake output

MoKob:
This looks

nigun:
osrm-backend  - curent version (clone yesterday)
stdlib version -how to look?
lua and luabind version 5.1.5 (run aptget install lua5.1 liblua5.1-0-dev libluabind-dev )
cmake output:
-- Building on a 64 bit system
-- Configuring OSRM in debug mode
-- Setting linker optimizations
-- Boost version: 1.54.0
-- Found the following Boost libraries:
--   date_time
--   filesystem
--   iostreams
--   program_options
--   regex
--   system
--   thread
-- Found Intel TBB
-- TBB interface version: 7000
-- Looking for Luabind...
-- Found Luabind: /usr/lib/libluabind.so
-- Looking for LuaJIT 5.2
-- Could NOT find LUAJIT (missing:  LUAJIT_LIBRARIES LUAJIT_INCLUDE_DIR)
-- Performing Test LUABIND_WORKS
[...]

PR #2598, opened 2016-06-28
fixing luabind / CP issue in profile

MoKob:
It seems that some luabind versions might return "nil" for empty strings.
These are probably the result of: https://github.com/Project-OSRM/osrm-backend/issues/2594.

By checking for the availability of the data, we make sure not to trigger the error.

### DIFF
--- a/features/guidance/turn-lanes.feature
+++ b/features/guidance/turn-lanes.feature
@@ -107,6 +107,24 @@ Feature: Turn Lane Guidance
             | a,d       | road,turn,turn | depart,turn right,arrive        | ,straight:false right:true, |
             | a,c       | road,road,road | depart,use lane straight,arrive | ,straight:true right:false, |
 
+    @PROFILE @LANES
+    Scenario: Turn with Bus-Lane but without lanes
+        Given the node map
+            | a |   | b |   | c |
+            |   |   |   |   |   |
+            |   |   | d |   |   |
+
+        And the ways
+            | nodes | name | lanes:psv |
+            | ab    | road | 1         |
+            | bc    | road | yes       |
+            | bd    | turn |           |
+
+        When I route I should get
+            | waypoints | route          | turns                    |
+            | a,d       | road,turn,turn | depart,turn right,arrive |
+            | a,c       | road,road      | depart,arrive            |
+
     #turn lanes are often drawn at the incoming road, even though the actual turn requires crossing the intersection first
     @todo @WORKAROUND-FIXME @bug
     Scenario: Turn Lanes at Segregated Road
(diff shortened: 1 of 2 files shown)


features/guidance/turn-lanes.feature
```diff
@@ -107,6 +107,24 @@ Feature: Turn Lane Guidance
             | a,d       | road,turn,turn | depart,turn right,arrive        | ,straight:false right:true, |
             | a,c       | road,road,road | depart,use lane straight,arrive | ,straight:true right:false, |
 
+    @PROFILE @LANES
```
MoKob: On my system, this fails to trigger any issues. I've added the test nonetheless.

profiles/car.lua
```diff
@@ -206,10 +206,10 @@ local function getTurnLanes(way)
     local turn_lanes_bw = way:get_value_by_key("turn:lanes:backward")
 
     if( fw_psv ~= 0 or bw_psv ~= 0 ) then
-        turn_lanes = trimLaneString(turn_lanes, bw_psv, fw_psv )
-        turn_lanes_fw = trimLaneString(turn_lanes_fw, bw_psv, fw_psv )
+        if( turn_lanes ) then turn_lanes = trimLaneString(turn_lanes, bw_psv, fw_psv ) end
+        if( turn_lanes_fw ) then turn_lanes_fw = trimLaneString(turn_lanes_fw, bw_psv, fw_psv ) end
         --backwards turn lanes need to treat bw_psv as fw_psv and vice versa
-        turn_lanes_bw = trimLaneString(turn_lanes_bw, fw_psv, bw_psv )
+        if( turn_lanes_bw) then turn_lanes_bw = trimLaneString(turn_lanes_bw, fw_psv, bw_psv ) end
```
daniel-j-h: Shouldn't this be

    if X and X ~= ""

on all of them?

TheMarex: Yeah empty string will technically work for trimLaneString but we might as well save ourselves the overhead of calling it.

On a side note, the canonical way of doing `if` in lua is:
```lua
if condition then
   instruction
end
```

Note the lack of parenthesis and the new line.

MoKob: well, as long as X is a `X` is a string, the strings are handled correctly in trimLaneString (https://github.com/Project-OSRM/osrm-backend/blob/master/include/extractor/guidance/toolkit.hpp#L500). This only works if the string has actual size.

nigun: +        if( turn_lanes_bw) then turn_lanes_bw = trimLaneString(turn_lanes_bw, fw_psv, bw_psv ) end.... not work.

i change 208 string: if( fw_psv and fw_psv ~= 0 or bw_psw and bw_psv ~= 0 ) then

nigun:
nil value in bw_psv variable.

MoKob:
@nigun there was indeed a copy-paste error in extracting the `psv_lanes`. I fixed the line here that is supposed to prevent the return of a `nil` value.

nigun:
@MoKob Ok. thanks :)) the problem is solved.

PR #2573, opened 2016-06-22
Prevent linker from discarding TBB symbols we need, fixes #2557

daniel-j-h:
We build `osrm_contract` (the library) linking in libtbb. We then
build `osrm-contract` (the binary) linking in `osrm_contract` (the
library).

Because we're only using TBB's `parallel_invoke` in the code for the
`osrm-contract` binary, it seems like the linker discards some symbols
in the `osrm_contract` library (since the library does not need them).

Therefore we have to link libtbb for the binary again. Even worse, the
order now matters: if we first link in `osrm_contract` and then libtbb,
we're still hitting the issue re. discarded symbols.

Therefore we first link in all dependencies (libtbb, libboost*), and
only then `osrm_contract`.

Strictly speaking, we probably have to do this for all of our binary
targets, or we will hit similar issues in the future.

-----------------------------------------------------------------------------------

I tested this in a Debian Jessie container, making sure it resolves #2557.

PR #1647, opened 2015-09-04
[not ready] Buildsystem Improvements

daniel-j-h:
This Pull Request tracks the progress. Do not merge yet.

The high level goal is to improve the build system in regard to warnings, build times, compiler and linker specific settings and features.

Please read every commit's description, as they have detailed explanations attached. This pull request just tracks the overall progress.

PR #2601, opened 2016-06-28
Use gold linker by default if available

TheMarex:


TheMarex:
Hm interesting. Building shared libraries with gold seems to cause errors when running the tests. I wonder why.

Also OSX seems to have ld.gold installed but the lua test fails with it?

/cc @daniel-j-h any ideas since you played around with this before?

daniel-j-h:
Yeah I wanted to do this back [half a year ago](https://github.com/Project-OSRM/osrm-backend/pull/1647) as it allows for [debug fission](https://gcc.gnu.org/wiki/DebugFission) but could not get it to work back then (can't remember idea why). The [Webkit common CMakeLists](https://github.com/WebKit/webkit/blob/master/Source/cmake/OptionsCommon.cmake) has some great logic for it.

You also need to set the linker flags by the way. All of them (CMake has multiple vars for it).

daniel-j-h:
`gcc5-release-shared` fails on Travis:

```
[info] Parsing in progress..
../../scripts/timer.sh: line 7: 11481 Segmentation fault      (core dumped) /bin/bash -c "$CMD"
Running osrm-contract...
../../scripts/timer.sh "osrm-contract" ../../build/osrm-contract monaco.osrm
```

Locally I'm hitting https://github.com/Project-OSRM/osrm-backend/pull/2573 as it's not in this branch.

```
/tmp/ccqg9EhU.ltrans11.ltrans.o:<artificial>:vtable for tbb::empty_task: error: undefined reference to 'tbb::empty_task::~empty_task()'
/tmp/ccqg9EhU.ltrans11.ltrans.o:<artificial>:vtable for tbb::empty_task: error: undefined reference to 'tbb::empty_task::~empty_task()'
```

Can you rebase on master and see how Travis likes it then?

daniel-j-h:
Rebased against master. Let's see how Travis likes it now.

For the record, with gold I'm still hitting the TBB issue from above even with the https://github.com/Project-OSRM/osrm-backend/commit/61ba985bc95290f6f29f70233ff345c8e15dbc09 applied.

daniel-j-h:
For the record, still fails with the error above with gcc 4.9.2. Works with gcc 6.1.0.

TheMarex:
Yeah was trying to debug this locally with docker. Not sure who is at fault here.

TheMarex:
For whatever reason gold linker and shared libraries don't work together. Tracking in https://github.com/Project-OSRM/osrm-backend/issues/2785 for now. This PR is not actionable right now.

PR #2609, opened 2016-06-30
Show realtime road segments duration in match.

gardster:
The match plugin has a "duration" field in output, but takes information for it from the road graph. It is a strange behavior when we have timestamps in the query. 
This PR rewrites segments weights at the response rendering phase with new values computed from matched timestamps.

### DIFF
--- a/include/engine/api/match_api.hpp
+++ b/include/engine/api/match_api.hpp
@@ -27,15 +27,50 @@ class MatchAPI final : public RouteAPI
     }
 
     void MakeResponse(const std::vector<map_matching::SubMatching> &sub_matchings,
-                      const std::vector<InternalRouteResult> &sub_routes,
+                      std::vector<InternalRouteResult> &sub_routes,
+                      const std::vector<unsigned int> & timestamps,
                       util::json::Object &response) const
     {
         auto number_of_routes = sub_matchings.size();
         util::json::Array routes;
         routes.values.reserve(number_of_routes);
         BOOST_ASSERT(sub_matchings.size() == sub_routes.size());
-        for (auto index : util::irange<std::size_t>(0UL, sub_matchings.size()))
+        for (auto index : util::irange<std::size_t>(0UL, number_of_routes))
         {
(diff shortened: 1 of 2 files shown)


danpat:
Ah, nice idea.  Yes, this makes sense if real timestamp data is available.

Can you add a test case for this?

daniel-j-h:
We often said to users "timestamps don't _really_ have to be timestamps, increasing integers work fine, too" --- this will change it.

TheMarex:
Thanks for looking into this! :+1: Unfortunately I think there are some misconceptions here.

> The match plugin has a "duration" field in output, but takes information for it from the road graph. It is a strange behavior when we have timestamps in the query. 

This is not the function of the `match` plugin. The result should always reflect the state of the internal data. It does not interpolate trace data, it uses it as input to compute the most probable route given its current knowledge of the road network. It is basically like asking OSRM "Hey what do _you_ think this route would look like".

If you need this behavior it is quite simple to implement this client side using some weighted scaling:
- For each pair of input coordinates compute the time between them `t_trace_segment`.
- For each step along the route compute the accumulated `t_step` duration (time since start of the `RouteLeg`)
- Scale each of those durations by: `t_step / t_leg * t_trace_segment`

Thanks again for the effort. Hope this explanation helps.

Komzpa:
@TheMarex 

"Hey what do you think this route would look like". - this is the function of "route" call.

There are both temporal and spatial components in the trace. What you offer is to ignore the temporal component of the path at all and just treat is as /route call - then:
 - why at all separate these two concepts?
 - why have timestamps at the input, if "any sequence will be fine"?

The result should always reflect the state of the internal data - that is unlikely. I'd say that result should always be dependant on inputs, and inputs are OSM graph (going to internal data, but we don't actually care - treat it as cache), speed profile, and trace with timestamps.

It is not expected in match call to get back timestamps that are not related at all to the timestamps fed into it.

Having weighted scaling doesn't resolve the issues of being stuck in traffic and moving disproportionally.

PR #2612, opened 2016-06-30
[Ready] Lane Anticipation for Multi-Hop Roundabout Instructions

daniel-j-h:
At the moment multi-hop roundabout instructions replace multiple silent instructions and therefore no lane anticipation can be done on a multi-hop roundabout instruction. The multi-hop roundabout instruction simply contains the lane information from the enter instruction.

This changeset implements Lane Anticipation on roundabouts, delimited
by enter / leave step pairs. It does not handle lane anticipation
within a roundabout.

Lane anticipation happens on the granularity of a valid roundbaout:

We discard partial roundabout (enter without exit or exit without
enter) or data issues (no roundabout, exit before enter).

Related:

- #2626 for lanes within a roundabout

- #2625 for handling going straight in lane anticipation

### DIFF
--- a/features/guidance/anticipate-lanes.feature
+++ b/features/guidance/anticipate-lanes.feature
@@ -259,6 +259,140 @@ Feature: Turn Lane Guidance
             | waypoints | route                 | turns                                                   | lanes                                                                                                                                                  |
             | a,f       | abx,bcy,cdz,dew,ef,ef | depart,turn right,turn left,turn right,turn left,arrive | ,straight:false right:false right:true right:false,left:false left:true straight:false,straight:false right:true right:false,left:true straight:false, |
 
+    @anticipate
+        Scenario: Anticipate with lanes in roundabout: roundabouts as the unit of anticipation
+        Given the node map
+            |   |   | e |   |   |
+            | a | b |   | d | f |
+            |   |   | c |   |   |
+            |   |   |   |   |   |
+            |   |   | g |   |   |
+            | k | h |   | j | l |
+            |   |   | i |   |   |
+
+        And the ways
+            | nodes | turn:lanes:forward                      | highway | junction   | #   |
+            | ab    | slight_right\|slight_right&slight_right | primary |            |     |
+            | bc    | slight_left\|slight_right&slight_right  | primary | roundabout | top |
+            | cd    |                                         | primary | roundabout | top |
+            | de    |                                         | primary | roundabout | top |
+            | eb    |                                         | primary | roundabout | top |
+            | df    |                                         | primary |            |     |
+            | cg    | slight_right\|slight_right              | primary |            |     |
+            | gh    | slight_left\|slight_right               | primary | roundabout | bot |
+            | hi    |                                         | primary | roundabout | bot |
+            | ij    | slight_left\|slight_right               | primary | roundabout | bot |
+            | jg    |                                         | primary | roundabout | bot |
+            | hk    |                                         | primary |            |     |
+            | jl    |                                         | primary |            |     |
+
+        When I route I should get
+            | #           | waypoints | route       | turns                                             | lanes                                                                                          |
+            | right-right | a,k       | ab,cg,hk,hk | depart,roundabout-exit-1,roundabout-exit-1,arrive | ,slight right:false slight right:false slight right:true,slight right:false slight right:true, |
+            | right-left  | a,l       | ab,cg,jl,jl | depart,roundabout-exit-1,roundabout-exit-2,arrive | ,slight right:false slight right:false slight right:true,slight right:false slight right:true, |
+            | todo exits  | a,f       | ab,df,df    | depart,roundabout-exit-2,arrive                   | ,slight right:false slight right:false slight right:true,                                      |
+            | todo exits  | a,e       | ab,bc,eb    | depart,roundabout-exit-undefined,arrive           | ,slight right:true slight right:true slight right:true,                                        |
+
+    @anticipate @todo
+    Scenario: Roundabout with lanes only tagged on exit
+        Given the node map
+            |   |   | e |   |   |
+            | a | b |   | d | f |
+            |   |   | c |   |   |
+
+        And the ways
+            | nodes | turn:lanes:forward                    | highway | junction   |
+            | ab    |                                       | primary |            |
+            | bc    |                                       | primary | roundabout |
+            | cd    | slight_left\|slight_left&slight_right | primary | roundabout |
+            | de    |                                       | primary | roundabout |
+            | eb    |                                       | primary | roundabout |
+            | df    |                                       | primary |            |
+
+        When I route I should get
+            | waypoints | route    | turns                                                 | lanes                                                    |
+            | a,f       | ab,df,df | depart,roundabout-exit-1,use lane slight right,arrive | ,,slight left:false slight left:false slight right:true, |
+
+    @anticipate
+    Scenario: Anticipate with lanes in roundabout where we stay on the roundabout for multiple exits
+        Given the node map
+            |   |   | a |   |   |
+            |   |   | b |   |   |
+            |   | c |   | g | h |
+            |   |   |   |   |   |
+            |   | d |   | f |   |
+            |   |   | e |   |   |
+            | x |   |   |   | y |
+
+        And the ways
+            | nodes | turn:lanes:forward         | highway | junction   |
+            | ab    | slight_right\|slight_right | primary |            |
+            | bc    |                            | primary | roundabout |
+            | cd    |                            | primary | roundabout |
+            | de    |                            | primary | roundabout |
+            | ef    |                            | primary | roundabout |
+            | fg    | slight_right               | primary | roundabout |
+            | gb    |                            | primary | roundabout |
+            | gh    |                            | primary |            |
+            | cx    |                            | primary |            |
+            | dx    |                            | primary |            |
+            | ey    |                            | primary |            |
+            | fy    |                            | primary |            |
+
+        When I route I should get
+            | waypoints | route    | turns                           | lanes                                  |
+            | a,h       | ab,gh,gh | depart,roundabout-exit-5,arrive | ,slight right:false slight right:true, |
+
+    @anticipate
+    Scenario: Departing or arriving inside a roundabout does not yet anticipate lanes
+        Given the node map
+            |   |   | a |   |   |
+            | x | b |   | d | y |
+            |   |   | c |   |   |
+
+        And the ways
+            | nodes | turn:lanes:forward         | highway | junction   | name       |
+            | xb    | slight_right\|slight_right | primary |            | xb         |
+            | dy    |                            | primary |            | dy         |
+            | ab    |                            | primary | roundabout | roundabout |
+            | bc    |                            | primary | roundabout | roundabout |
+            | cd    | left\|slight_right         | primary | roundabout | roundabout |
+            | da    |                            | primary | roundabout | roundabout |
+
+        When I route I should get
+            | waypoints | route                    | turns                                   | lanes                                  |
+            | x,y       | xb,dy,dy                 | depart,roundabout-exit-1,arrive         | ,slight right:false slight right:true, |
+            | x,c       | xb,roundabout,roundabout | depart,roundabout-exit-undefined,arrive | ,slight right:true slight right:true,  |
+            | x,a       | xb,roundabout,roundabout | depart,roundabout-exit-undefined,arrive | ,slight right:true slight right:true,  |
+
+    @anticipate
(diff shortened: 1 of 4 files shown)


daniel-j-h:
@MoKob so after trying to special-case `postProcessing` I just added an additional `anticipateLaneChange` call already in `postProcessing` when we haven't already removed silent instructions.

This means we're doing it twice, once on the silent instructions and then once again in the route api at the end of post-processing. I'm not familiar with your post-processing in between: is this a valid approach?

For the record, this breaks no tests so far. For my added scenario still gives the wrong result at the moment, though (not sure why, have to look into it) .

src/engine/guidance/post_processing.cpp
```diff
@@ -538,19 +541,86 @@ std::vector<RouteStep> removeNoTurnInstructions(std::vector<RouteStep> steps)
 // that we come across.
 std::vector<RouteStep> postProcess(std::vector<RouteStep> steps)
 {
+#pragma message remove
     // the steps should always include the first/last step in form of a location
     BOOST_ASSERT(steps.size() >= 2);
     if (steps.size() == 2)
         return steps;
 
+    // Before we invalidate and remove silent instructions, we handle roundabouts (before they're
```
MoKob: this should be in a dedicated function, if its not actually interleaved with the other processing

src/engine/guidance/post_processing.cpp
```diff
@@ -538,19 +541,86 @@ std::vector<RouteStep> removeNoTurnInstructions(std::vector<RouteStep> steps)
...
+    // constraints already to a roundabout's enter instruction.
+
+    const constexpr auto MIN_DURATION_NEEDED_FOR_LANE_CHANGE_INSIDE_ROUNDABOUT = 15.;
+
+    util::guidance::print(steps);
+
+    // Walk over all steps, find enter / leave roundabout pairs
+    for (auto it = begin(steps), last = end(steps); it != last;)
+    {
+        auto enter = std::find_if(it, last, [](const RouteStep &step) {
```
MoKob: this is unnecessarily large (O(n^2) instead of O(n)) complexity.
The enter should be the initial check. Only if *it is on enter, you need to process.

MoKob: by chat: function is correct, due to assignment and breaks below

daniel-j-h:
The basic case of roundabout lanes now seem to work. There are some open issues left. Todo:
- [ ] more tests (will probably reveal issues)
- [ ] cross-roundabout lanes, as in multiple subsequent roundabouts
- [ ] lanes within a roundabout (?)
- [ ] adding lanes to the enter instruction based on a roundabout's leave instructione
- [ ] multiple stay on roundabout instructions

features/guidance/anticipate-lanes.feature
```diff
@@ -259,6 +259,128 @@ Feature: Turn Lane Guidance
...
+            | dx    |                            | primary |            |
+            | ey    |                            | primary |            |
+            | fy    |                            | primary |            |
+
+        When I route I should get
+            | waypoints | route    | turns                           | lanes                                  |
+            | a,h       | ab,gh,gh | depart,roundabout-exit-5,arrive | ,slight right:false slight right:true, |
+
+    @anticipate-roundabout-onion
+    Scenario: Anticipation in Onion Roundabout spiraling inwards / outwards
```
MoKob: as discussed in https://github.com/Project-OSRM/osrm-backend/issues/2621, this is currently not a valid roundabout model for OSRM and cannot be handled correctly.

include/engine/guidance/toolkit.hpp
```diff
@@ -63,6 +66,39 @@ inline extractor::guidance::DirectionModifier::Enum angleToDirectionModifier(con
...
+//  - data issues: no roundabout, exit before enter
+template <typename Iter, typename Fn> inline Fn forEachRoundabout(Iter first, Iter last, Fn fn)
+{
+    while (first != last)
+    {
+        const auto enter = std::find_if(first, last, [](const RouteStep &step) {
+            return entersRoundabout(step.maneuver.instruction);
+        });
+
+        const auto leave = std::find_if(first, last, [](const RouteStep &step) {
```
MoKob: this should start at enter, not at first? Possible problem: via point in roundabout.

src/engine/guidance/post_processing.cpp
```diff
@@ -1,9 +1,11 @@
-#include "extractor/guidance/turn_instruction.hpp"
 #include "engine/guidance/post_processing.hpp"
+#include "extractor/guidance/turn_instruction.hpp"
 
 #include "engine/guidance/assemble_steps.hpp"
+#include "engine/guidance/lane_processing.hpp"
 #include "engine/guidance/toolkit.hpp"
 
+#include "util/debug.hpp"
```
MoKob: remove when done

src/engine/guidance/post_processing.cpp
```diff
@@ -497,11 +499,43 @@ void collapseTurnAt(std::vector<RouteStep> &steps,
     }
 }
 
+// Works on steps including silent and invalid instructions in order to do lane anticipation for
+// roundabouts which later on get collapsed into a single multi-hop instruction.
+std::vector<RouteStep> anticipateLaneChangeForRoundabouts(std::vector<RouteStep> steps)
+{
+    util::guidance::print(steps);
```
MoKob: also remove when done.

src/engine/guidance/post_processing.cpp
```diff
@@ -497,11 +499,43 @@ void collapseTurnAt(std::vector<RouteStep> &steps,
...
+
+    const constexpr auto MIN_DURATION_NEEDED_FOR_LANE_CHANGE_INSIDE_ROUNDABOUT = 15.;
```
MoKob: within a roundabout, I would account deem all lane changes relevant, since you are in a continuous turn.

src/engine/guidance/post_processing.cpp
```diff
@@ -543,14 +580,20 @@ std::vector<RouteStep> postProcess(std::vector<RouteStep> steps)
     if (steps.size() == 2)
         return steps;
 
+    // Before we invalidate and remove silent instructions, we handle roundabouts (before they're
+    // getting collapsed into a single multi-hop instruction) by back-propagating exit lane
+    // constraints already to a roundabout's enter instruction.
+    util::guidance::print(steps);
```
MoKob: remove when done.

src/engine/guidance/post_processing.cpp
```diff
@@ -497,11 +499,43 @@ void collapseTurnAt(std::vector<RouteStep> &steps,
...
+    const auto anticipate_lanes_in_quick_roundabouts = [&](StepIterRange roundabout) {
+        // We only care about roundabouts where there is no time to change lanes inside
+        const auto sum = [](const double v, const RouteStep &step) { return v + step.duration; };
+        const auto seconds_inside = std::accumulate(roundabout.first, roundabout.second, 0., sum);
+
+        if (seconds_inside >= MIN_DURATION_NEEDED_FOR_LANE_CHANGE_INSIDE_ROUNDABOUT)
+            return;
+
+        // We do lane anticipation on the roundabout's enter and leave step only.
+        // TODO: This means, lanes _inside_ the roundabout are ignored at the moment.
```
MoKob: only lanes on the enter instruction are reported to the outside. This paradigm has to be reworked, if it should work in the end.

include/engine/guidance/toolkit.hpp
daniel-j-h: `enter < leave` is handled a bit below separately, but you're right I can change it here :+1:

MoKob: the way it is currently handled, wouldn't it result in no roundabout being handled if it is exit, enter, exit, enter ....?

daniel-j-h: No the code jumped over the `exit`, starting at `enter` the next time --- changed it now to call `find` starting from the `enter` iterator.

MoKob: Ah right. Still I'd prefer the method proposed above, since it is easier to follow as it processes in order rather than jumping back to the front.

daniel-j-h:
Squased, rebased against master. Waiting for Travis.

features/guidance/anticipate-lanes.feature
```diff
@@ -259,6 +259,138 @@ Feature: Turn Lane Guidance
...
+            | hi    |                                         | primary | roundabout | bot |
+            | ij    | slight_left\|slight_right               | primary | roundabout | bot |
+            | jg    |                                         | primary | roundabout | bot |
+            | hk    |                                         | primary |            |     |
+            | jl    |                                         | primary |            |     |
+
+        When I route I should get
+            | #           | waypoints | route       | turns                                             | lanes                                                                                          |
+            | right-right | a,k       | ab,cg,hk,hk | depart,roundabout-exit-1,roundabout-exit-1,arrive | ,slight right:false slight right:true slight right:false,slight right:false slight right:true, |
+            | right-left  | a,l       | ab,cg,jl,jl | depart,roundabout-exit-1,roundabout-exit-2,arrive | ,slight right:false slight right:true slight right:false,slight right:false slight right:true, |
```
MoKob: add checks for all other exits, to make sure they don't show unreasonable stuff

features/guidance/anticipate-lanes.feature
```diff
@@ -259,6 +259,138 @@ Feature: Turn Lane Guidance
...
+            | ab    |                                       | primary |            |
+            | bc    |                                       | primary | roundabout |
+            | cd    | slight_left\|slight_left&slight_right | primary | roundabout |
+            | de    |                                       | primary | roundabout |
+            | eb    |                                       | primary | roundabout |
+            | df    |                                       | primary |            |
+
+        When I route I should get
+            | waypoints | route    | turns                                                 | lanes                                                    |
+            | a,f       | ab,df,df | depart,roundabout-exit-1,use lane slight right,arrive | ,,slight left:false slight left:false slight right:true, |
```
MoKob: same here

src/engine/guidance/post_processing.cpp
```diff
@@ -498,11 +499,46 @@ void collapseTurnAt(std::vector<RouteStep> &steps,
...
+    const auto anticipate_lanes_in_roundabout = [&](StepIterRange roundabout) {
+        // We do lane anticipation on the roundabout's enter and leave step only.
+        // TODO: This means, lanes _inside_ the roundabout are ignored at the moment.
+
+        auto enter = *roundabout.first;
+
+        // Although the enter instruction may be a left/right turn, for right-sided driving the
+        // roundabout is counter-clockwise and therefore we need to always set it to a left turn.
+        // FIXME: assumes right-side driving (counter-clockwise roundabout flow)
+        if (util::guidance::isRightTurn(enter.maneuver.instruction))
```
MoKob: wouldn't it be more reasonable to handle this case than to modify the turn-direction and output an invalid modifier? The modifier output cannot stay this way, at least.

daniel-j-h: Yes, instead of doing it here inside roundabouts and then later on again for all steps, I should do it in a single place instead: https://github.com/Project-OSRM/osrm-backend/issues/2628 thank you for catching this!

features/guidance/anticipate-lanes.feature
```diff
@@ -259,6 +259,140 @@ Feature: Turn Lane Guidance
...
+            | a | b |   | d | f |
+            |   |   | c |   |   |
+            |   |   |   |   |   |
+            |   |   | g |   |   |
+            | k | h |   | j | l |
+            |   |   | i |   |   |
+
+        And the ways
+            | nodes | turn:lanes:forward                      | highway | junction   | #   |
+            | ab    | slight_right\|slight_right&slight_right | primary |            |     |
```
TheMarex: Remember to fix this once https://github.com/Project-OSRM/osrm-backend/pull/2635 is merged.

TheMarex:
Does this require an update of the dataformat? Thinking about including this in 5.3.0

MoKob:
Should be data-update free, as it is post-processing only.

daniel-j-h:
Rebased against master. 

In this changeset I'm using the generic lane anticipation feature already present in master. @TheMarex you want to scheduled this for 5.3: both https://github.com/Project-OSRM/osrm-backend/pull/2642 and https://github.com/Project-OSRM/osrm-backend/pull/2636 - improving the generic lane anticipation - had to branch off of @MoKob's advanced lanes branch.

Which means you can merge this already into 5.3. Once the advanced lane guidance WIP PR (https://github.com/Project-OSRM/osrm-backend/pull/2614) lands, we can merge both https://github.com/Project-OSRM/osrm-backend/pull/2642 and https://github.com/Project-OSRM/osrm-backend/pull/2636 incrementally.

Issue #2553, opened 2016-06-15
Turn Lane Handling - Part Two


MoKob:
With https://github.com/Project-OSRM/osrm-backend/pull/2410 coming to a first close, I've marked a few items in both `anticipate-lanes.feature` and `turn-lanes.feature` as `todo`.

We want to get a first version of turn lane handling integrated into OSRM and work on further improving lane guidance in a second step.

The issues with the current implementation are:

- [ ] merge lanes: lanes merging to another lane at an intersection cannot be handled so far
- [ ] we need to decide whether to announce merge instructions when travelling along a road. This would essentially require knowledge of the current lane, which would probably only be available directly after a previous turn
- [ ] handle none simple intersections: intersections can currently only be handled if there exists a trivial mapping between available turn-lanes and turns. 
- [ ] roundabouts: turn lanes to enter a round-about and exit it are currently not handled at all
- [ ] handle turn lanes during collapse operations (segregated intersections with middle island) emit useless use-lane instructions and need to be adjusted / suppressed
- [ ] the access to a previous intersection is a hack in most cases. We need to implement complex maneuvers from multiple turns to correctly indicate lanes. This requires multi-hop turn restrictions.
- [ ] lane anticipation does not work when we have to stay on the left side during multiple right turns

MoKob:
Continues the work started for https://github.com/Project-OSRM/osrm-backend/issues/2296

MoKob:
- [ ] `destination:lanes=*` could possibly be used to find out about correct ways, if they are also found on the target way

TheMarex:
@MoKob could you please take a look if this is still uptodate and if so move it to the next milestone?

BjornRasmussen:
I've started a proposal on the OSM wiki for a new type of `from-via-to` relation that would state which lanes in `from` connect to which lanes in `to`.  How useful will these relations be, and are there any ways that they can be improved to make them more useful to navigation and routing?

https://wiki.openstreetmap.org/wiki/Proposed_features/Connectivity

PR #2614, opened 2016-07-01
advanced lanes

MoKob:
# TODO
- [x] fix: collapse-detail cases are omitting the wrong turn type (turn over continue) for some reason
- [x] fix: anticipate lanes (https://travis-ci.org/Project-OSRM/osrm-backend/jobs/154098554#L3978) #2626 regression

- [x] review
  subrequests by daniel-j-h have been merged into this PR have have been reviewed by me. An additional review could focus on
  - [x] https://github.com/Project-OSRM/osrm-backend/pull/2614/commits/d6aa56b9146f3d7f33e38ad2d5cfc25f8dc12c0d
  - [x] https://github.com/Project-OSRM/osrm-backend/pull/2614/commits/17e7642ad9efcb1f0d0cc735b4531e483c1c3de8
  - [x] https://github.com/Project-OSRM/osrm-backend/pull/2614/commits/268201a38b2ba30be0844bb623ca10ed3bf16a36

 - [x] final adjustments

Working on https://github.com/Project-OSRM/osrm-backend/issues/2553.

Initial goal: Slipway Turns.

A slipway is a small dedicated turn lane.

screen shot 2016-06-29 at 15 06 18

Even though the modelling here does not follow the slipway-paradigm, such roads can already be modelled using a separate turn way.

Usually it is larger roads, though:
screen shot 2016-07-01 at 11 52 23

In such cases, the experience of the driver deviates from the model of the road. The first step here aims at creating a lane-description that includes this slipway in the reported available lanes.

In the first example, we would currently see
`left` and `straight` on the one turn, `right` on the second turn.
The aim of this PR is to expose all three lanes in the turn-lanes.

### DIFF
--- a/features/guidance/advanced-lanes.feature
+++ b/features/guidance/advanced-lanes.feature
@@ -1,0 +1,186 @@
+@routing @guidance @turn-lanes
+Feature: Turn Lane Guidance
+
+    Background:
+        Given the profile "car"
+        Given a grid size of 3 meters
+
+    #requires https://github.com/cucumber/cucumber-js/issues/417
+    #Due to this, we use & as a pipe character. Switch them out for \| when 417 is fixed
+    @bug @WORKAROUND-FIXME
+    Scenario: Separate Turn Lanes
+        Given the node map
+            |   |   |   |   |   |   |   | e |   |
+            | a |   |   | b |   |   |   | c | g |
+            |   |   |   |   |   |   |   | d |   |
+            |   |   |   |   |   |   |   | f |   |
+
+        And the ways
+            | nodes | turn:lanes:forward | name     | oneway |
+            | ab    |                    | in       | yes    |
+            | bc    | left\|through      | in       | yes    |
+            | bd    | right              | in       | yes    |
+            | ec    |                    | cross    | no     |
+            | cd    |                    | cross    | no     |
+            | df    |                    | cross    | no     |
+            | cg    |                    | straight | no     |
+
+        And the relations
+            | type        | way:from | way:to | node:via | restriction   |
+            | restriction | bd       | cd     | d        | no_left_turn  |
+            | restriction | bc       | cd     | c        | no_right_turn |
+
+       When I route I should get
+            | waypoints | route                | turns                           | lanes                                  |
+            | a,e       | in,cross,cross       | depart,turn left,arrive         | ,left:true straight:false right:false, |
+            | a,g       | in,straight,straight | depart,new name straight,arrive | ,left:false straight:true right:false, |
+            | a,f       | in,cross,cross       | depart,turn right,arrive        | ,left:false straight:false right:true, |
+
+    @TODO @2650 @bug
+    Scenario: Sliproad with through lane
+        Given the node map
+            |   |   |   |   |   |   |   |   |   | f |   |   |   |
+            |   |   |   |   |   |   |   |   |   |   |   |   |   |
+            |   |   |   |   |   |   |   | g |   |   |   |   |   |
+            |   |   |   |   |   |   |   |   |   |   |   |   |   |
+            |   |   |   |   |   |   |   |   |   |   |   |   | e |
+            |   |   |   |   |   |   |   |   |   |   |   |   |   |
+            |   |   |   |   |   |   |   |   |   | d |   |   |   |
+            | a |   |   |   |   |   |   |   |   |   |   |   |   |
+            |   |   |   |   |   | b |   |   |   |   |   |   |   |
+            |   |   |   |   |   |   |   |   |   | c |   |   |   |
+            |   |   |   |   |   |   |   |   |   |   |   |   |   |
+            |   |   |   |   |   |   |   |   |   |   |   |   |   |
+            |   |   |   |   | h |   |   |   |   | i |   |   |   |
+
+        And the ways
+            | nodes | name   | oneway | turn:lanes:forward |
+            | ab    | ghough | yes    |                    |
+            | bc    | ghough | yes    | through\|none      |
+            | bd    | ghough | yes    | none\|through      |
+            | de    | ghough | yes    |                    |
+            | fgb   | haight | yes    |                    |
+            | bh    | haight | yes    | left\|none         |
+            | fd    | market | yes    |                    |
+            | dc    | market | yes    |                    |
+            | ci    | market | yes    |                    |
+
+        And the relations
+            | type        | way:from | way:to | node:via | restriction   |
+            | restriction | fgb      | bd     | b        | no_left_turn  |
+            | restriction | fgb      | bc     | b        | no_left_turn  |
+
+        When I route I should get
+            | waypoints | route                | turns                              | lanes |
+            | a,h       | ghough,haight,haight | depart,turn right,arrive           |       |
+            | a,i       | ghough,market,market | depart,turn right,arrive           |       |
+            | a,e       | ghough,ghough,ghough | depart,continue slight left,arrive |       |
+
+    @TODO @2650 @bug
+    Scenario: Sliproad with through lane
+        Given the node map
+            |   |   |   |   |   |   |   | f |   |   |
+            |   |   |   |   |   |   |   |   |   |   |
+            |   |   |   |   |   |   | g |   |   |   |
+            |   |   |   |   |   |   |   |   |   |   |
+            |   |   |   |   |   |   |   |   |   | e |
+            |   |   |   |   |   |   |   |   |   |   |
+            |   |   |   |   |   |   |   | d |   |   |
+            | a |   |   |   |   |   |   |   |   |   |
+            |   |   |   |   | b |   |   |   |   |   |
+            |   |   |   |   |   |   |   | c |   |   |
+            |   |   |   |   |   |   |   |   |   |   |
+            |   |   |   |   |   |   |   |   |   |   |
+            |   |   |   | h |   |   |   | i |   |   |
+
+        And the ways
+            | nodes | name   | oneway | turn:lanes:forward |
+            | ab    | ghough | yes    |                    |
+            | bc    | ghough | yes    | through\|none      |
+            | bd    | ghough | yes    | none\|through      |
+            | fgb   | haight | yes    |                    |
+            | bh    | haight | yes    | left\|none         |
+            | fd    | market | yes    |                    |
+            | dc    | market | yes    |                    |
+            | ci    | market | yes    |                    |
+
+        And the relations
+            | type        | way:from | way:to | node:via | restriction   |
+            | restriction | bd       | dc     | d        | no_right_turn |
+            | restriction | fgb      | bd     | b        | no_left_turn  |
+            | restriction | fgb      | bc     | b        | no_left_turn  |
+
+        When I route I should get
+            | waypoints | route                | turns                              | lanes |
+            | a,h       | ghough,haight,haight | depart,turn right,arrive           |       |
+            | a,i       | ghough,market,market | depart,turn right,arrive           |       |
+            | a,e       | ghough,ghough,ghough | depart,continue slight left,arrive |       |
+
+    Scenario: Separate Turn Lanes
+        Given the node map
+            |   |   |   |   |   |   |   | e |   |
+            | a |   |   | b |   |   |   | c | g |
+            |   |   |   |   |   |   |   | d |   |
+            |   |   |   |   |   |   |   | f |   |
+
+        And the ways
+            | nodes | turn:lanes:forward | name     | oneway |
+            | ab    |                    | in       | yes    |
+            | bc    | left\|through      | in       | yes    |
+            | bd    | right              | in       | yes    |
+            | ec    |                    | cross    | no     |
+            | cd    |                    | cross    | no     |
+            | df    |                    | cross    | no     |
+            | cg    |                    | straight | no     |
+
+        And the relations
+            | type        | way:from | way:to | node:via | restriction   |
+            | restriction | bd       | cd     | d        | no_left_turn  |
+            | restriction | bc       | cd     | c        | no_right_turn |
+
+       When I route I should get
+            | waypoints | route                | turns                           | lanes                                  |
+            | a,e       | in,cross,cross       | depart,turn left,arrive         | ,left:true straight:false right:false, |
+            | a,g       | in,straight,straight | depart,new name straight,arrive | ,left:false straight:true right:false, |
+            | a,f       | in,cross,cross       | depart,turn right,arrive        | ,left:false straight:false right:true, |
+
+    @guidance @lanes @sliproads
+    Scenario: Separate Turn Lanes Next to other turns
+        Given the node map
+            |   |   |   |   |   |   |   | e |   |
+            | a |   |   | b |   |   |   | c | g |
+            |   |   |   |   |   |   |   | d |   |
+            |   |   |   |   |   |   |   | f |   |
+            |   |   |   |   |   |   |   |   |   |
+            |   |   |   |   |   |   |   |   |   |
+            |   |   |   |   |   |   |   |   |   |
+            |   |   |   |   |   |   |   |   |   |
+            |   |   |   |   |   |   |   |   |   |
+            |   |   |   |   |   |   |   |   |   |
+            | i |   |   | h |   |   |   | j |   |
+
+        And the ways
+            | nodes | turn:lanes:forward | name     | oneway |
+            | ab    |                    | in       | yes    |
+            | bc    | left\|through      | in       | yes    |
+            | bd    | right              | in       | yes    |
+            | ec    |                    | cross    | no     |
+            | cd    |                    | cross    | no     |
+            | df    |                    | cross    | no     |
+            | cg    |                    | straight | no     |
+            | bh    | left\|right        | turn     | yes    |
+            | ihj   |                    | other    | no     |
+
+        And the relations
+            | type        | way:from | way:to | node:via | restriction   |
+            | restriction | bd       | cd     | d        | no_left_turn  |
+            | restriction | bc       | cd     | c        | no_right_turn |
+
+       When I route I should get
+            | waypoints | route                | turns                               | lanes                                  |
+            | a,e       | in,cross,cross       | depart,turn left,arrive             | ,left:true straight:false right:false, |
+            | a,g       | in,straight,straight | depart,new name straight,arrive     | ,left:false straight:true right:false, |
+            | a,f       | in,cross,cross       | depart,turn right,arrive            | ,left:false straight:false right:true, |
+            | a,j       | in,turn,other,other  | depart,turn right,turn left,arrive  | ,,left:true right:false,               |
+            | a,i       | in,turn,other,other  | depart,turn right,turn right,arrive | ,,left:false right:true,               |
+
--- a/features/guidance/advanced-lanes.feature
+++ b/features/guidance/advanced-lanes.feature
@@ -132,55 +132,7 @@ Feature: Turn Lane Guidance
             | a,i       | ,ksd,ksd | depart,turn right,arrive | ,left:false none:true right:true, |
 
     @todo @bug @2650 @sliproads
-    #market and haight in SF
-    #http://www.openstreetmap.org/#map=19/37.77308/-122.42238
-    Scenario: Through Street Crossing Mid-Turn
-        Given the node map
-            |   |   |   |   |   |   |   | g | j |   |
-            |   |   |   |   |   |   |   |   |   |   |
-            |   |   |   |   |   |   |   |   |   |   |
-            |   |   |   |   |   |   |   |   |   |   |
-            |   |   |   |   |   | k |   |   |   |   |
-            |   |   |   |   |   |   |   |   |   |   |
-            |   |   |   |   |   |   |   |   |   | f |
-            |   |   |   |   |   |   |   |   | e |   |
-            |   |   |   |   |   |   |   | d |   |   |
-            |   |   |   |   |   |   |   |   |   |   |
-            | a |   |   | b |   |   |   |   |   |   |
-            |   |   |   |   |   |   |   | c |   |   |
-            |   |   |   |   |   |   |   |   |   |   |
-            |   |   |   |   |   |   |   |   |   |   |
-            |   |   |   |   |   |   |   |   |   |   |
-            |   |   |   |   |   |   |   |   |   |   |
-            |   |   |   |   |   |   |   |   |   |   |
-            |   |   | l |   |   |   |   | h | i |   |
-
-        And the ways
-            | nodes | name   | highway     | oneway | turn:lanes:forward |
-            | ab    | ghough | secondary   | yes    |                    |
-            | bc    | ghough | secondary   | yes    | through\|through   |
-            | bd    | ghough | secondary   | yes    | none\|through      |
-            | def   | ghough | secondary   | yes    |                    |
-            | gd    | market | primary     | yes    |                    |
-            |  dc   | market | primary     | yes    |                    |
-            |   ch  | market | primary     | yes    |                    |
-            | iej   | market | primary     | yes    |                    |
-            | gkb   | haight | residential | yes    |                    |
-            | bl    | haight | residential | yes    | left\|none         |
-
-        And the relations
-            | type        | way:from | way:to | node:via | restriction   |
-            | relation    | bd       | dc     | d        | no_right_turn |
-
-        When I route I should get
-            | waypoints | route                | turns                              | lanes                                                    |
-            | a,l       | ghough,haight,haight | depart,turn right,arrive           | ,none:false straight:false straight:false straight:true, |
-            | a,h       | ghough,market,market | depart,turn slight right,arrive    | ,none:false straight:false straight:true straight:true,  |
-            | a,j       | ghough,market,market | depart,turn left,arrive            | ,none:true straight:false straight:false straight:false, |
-            | a,f       | ghough,ghough,ghough | depart,continue slight left,arrive | ,none:true straight:true straight:false straight:false,  |
-
-    @todo @bug @2650 @sliproads
-    #market and haight in SF
+    #market and haight in SF, restricted turn
     #http://www.openstreetmap.org/#map=19/37.77308/-122.42238
     Scenario: Market/Haight without Through Street
         Given the node map
@@ -209,8 +161,8 @@ Feature: Turn Lane Guidance
             | bd    | ghough | secondary   | yes    | none\|through      |
             | def   | ghough | secondary   | yes    |                    |
             | gd    | market | primary     | yes    |                    |
-            |  dc   | market | primary     | yes    |                    |
-            |   ch  | market | primary     | yes    |                    |
+            | dc    | market | primary     | yes    |                    |
+            | ch    | market | primary     | yes    |                    |
             | iej   | market | primary     | yes    |                    |
             | bl    | haight | residential | yes    | left\|none         |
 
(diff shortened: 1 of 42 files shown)


MoKob:
I am marking this as a first step in the process. A few additional cases are solved. One problem remains: the extract can trigger an assertion right now. Investigating the issue

TheMarex:
Does this need more work? I'm currently trying to remember why we punted on this. @MoKob would be great if you could summarize the TODO in the PR description.

MoKob:
I've had a look at the regression I see for #2626. With #2693, roundabouts should not be considered anymore.

The test that was failing was looking at roundabout lanes in between two other lanes. For me, I can get it working by switching [these two lines](https://github.com/Project-OSRM/osrm-backend/blob/3ccbe9842cd3b1a894e1b9047e1c1c7121899a69/include/engine/api/route_api.hpp#L153-L154).

In my view, doing so would ignore #2693. Since the current decision is to ignore lanes in roundabout, having this case as a todo is the correct approach. If travis succeeds now, we should be able to merge, pending review.

features/guidance/advanced-lanes.feature
```diff
@@ -0,0 +1,186 @@
...
+            |   |   |   |   |   |   |   | e |   |
+            | a |   |   | b |   |   |   | c | g |
+            |   |   |   |   |   |   |   | d |   |
+            |   |   |   |   |   |   |   | f |   |
+
+        And the ways
+            | nodes | turn:lanes:forward | name     | oneway |
+            | ab    |                    | in       | yes    |
+            | bc    | left\|through      | in       | yes    |
+            | bd    | right              | in       | yes    |
```
daniel-j-h: Wow nice, I had no idea you handle turn lanes that are modelled out like that

MoKob: Well, until I did this PR I had no idea I had to :)

features/guidance/advanced-lanes.feature
```diff
@@ -0,0 +1,186 @@
...
+            | restriction | bd       | cd     | d        | no_left_turn  |
+            | restriction | bc       | cd     | c        | no_right_turn |
+
+       When I route I should get
+            | waypoints | route                | turns                           | lanes                                  |
+            | a,e       | in,cross,cross       | depart,turn left,arrive         | ,left:true straight:false right:false, |
+            | a,g       | in,straight,straight | depart,new name straight,arrive | ,left:false straight:true right:false, |
+            | a,f       | in,cross,cross       | depart,turn right,arrive        | ,left:false straight:false right:true, |
+
+    @guidance @lanes @sliproads
```
daniel-j-h: Why tagged guidance here?

features/guidance/turn-lanes.feature
```diff
@@ -424,17 +426,19 @@ Feature: Turn Lane Guidance
...
-            | de    | road  |                              | yes    | primary   |
-            | fdcg  | cross |                              |        | secondary |
+            | nodes | name  | turn:lanes:forward                  | oneway | highway   |
+            | ab    | road  | left\|left\|through\|through\|right | yes    | primary   |
+            | bd    | road  | through\|through                    | yes    | primary   |
+            | bc    | road  | left\|left                          | yes    | primary   |
+            | de    | road  |                                     | yes    | primary   |
+            | fd    | cross |                                  |        | secondary |
+            |  dc   | cross |                                  |        | secondary |
+            |   cg  | cross |                                  |        | secondary |
```
daniel-j-h: Alignment!! :P

include/extractor/extractor.hpp
```diff
@@ -87,6 +88,16 @@ class Extractor
...
         const std::vector<util::guidance::EntryClass> &entry_classes) const;
+
+    void WriteTurnLaneData(const std::string &turn_lane_file) const;
+
+    // globals persisting during the extraction process and the graph generation process
+
+    // during turn lane analysis, we might have to combine lanes for roads that are modelled as two
+    // but are more or less experienced as one. This can be due to solid lines in between lanes, for
+    // example, that genereate a small separation between them. As a result, we might have to
+    // augment the turn lane map during processing, further adding more types.
```
daniel-j-h: Thanks for the explanation in the comment

include/extractor/guidance/turn_lane_handler.hpp
```diff
@@ -33,27 +33,75 @@ namespace lanes
...
+        SLIPROAD, // Sliproads are simple assignments that, for better visual representation should
+                  // include turns from other roads in their listings
+        MERGE,    // Merging Lanes
+        NONE,     // not a turn lane scenario at all
+        INVALID,  // some error might have occurred
+        UNKNOWN,  // UNKNOWN describes all cases that we are currently not able to handle
+        NUM_SCENARIOS
+    } TurnLaneScenario;
+
+    const constexpr static char *scenario_names[TurnLaneScenario::NUM_SCENARIOS] = {
```
daniel-j-h: I think if you leave the `static` out here the compiler is free to not include the array at all. As in: `constexpr` already makes sure the literals are ROMable, `static` requires the array itself to have an address.

include/extractor/guidance/turn_lane_handler.hpp
```diff
@@ -33,27 +33,75 @@ namespace lanes
...
+        "Simple",
+        "Partition Local",
+        "Simple Previous",
+        "Partition Previous",
+        "Sliproad",
+        "Merge",
+        "None",
+        "Invalid",
+        "Unknown"};
+
```
daniel-j-h: Also we probably should `static_assert` the array's size at compile time just to be on the safe side.

include/extractor/guidance/turn_lane_handler.hpp
```diff
@@ -33,27 +33,75 @@ namespace lanes
...
     OSRM_ATTR_WARN_UNUSED
-    Intersection assignTurnLanes(const NodeID at,
-                                 const EdgeID via_edge,
-                                 Intersection intersection,
-                                 LaneDataIdMap &id_map) const;
+    Intersection assignTurnLanes(const NodeID at, const EdgeID via_edge, Intersection intersection);
 
   private:
+    unsigned *count_handled;
+    unsigned *count_called;
```
daniel-j-h: Why pointers?

The raw pointers here and the `new` / `delete` below are a huge red flag to me.

include/extractor/guidance/turn_lane_types.hpp
```diff
@@ -95,6 +96,11 @@ struct TurnLaneDescription_hash
     }
 };
 
+typedef std::unordered_map<guidance::TurnLaneDescription,
+                           LaneDescriptionID,
+                           guidance::TurnLaneDescription_hash>
```
daniel-j-h: I just had a look at the hash struct since I'm seeing this for the first time.
There is `hash_range` providing the effect you need there:
http://www.boost.org/doc/libs/1_55_0/doc/html/hash/reference.html#idp53740664-bb

src/extractor/edge_based_graph_factory.cpp
```diff
@@ -344,13 +346,19 @@ void EdgeBasedGraphFactory::GenerateEdgeExpandedEdges(
                                          name_table,
                                          street_name_suffix_table,
                                          profile_properties);
-    guidance::lanes::TurnLaneHandler turn_lane_handler(
-        *m_node_based_graph, turn_lane_offsets, turn_lane_masks, turn_analysis);
+
+    guidance::LaneDataIdMap lane_data_map;
+    guidance::lanes::TurnLaneHandler turn_lane_handler(*m_node_based_graph,
```
daniel-j-h: The turn lane handler here isn't const anyway and I think this is the only place we're using it. Therefore the non-const function issue above should be fine.

src/extractor/extractor.cpp
```diff
@@ -93,7 +94,8 @@ int Extractor::run(ScriptingEnvironment &scripting_environment)
         util::SimpleLogger().Write() << "Threads: " << number_of_threads;
 
         ExtractionContainers extraction_containers;
-        auto extractor_callbacks = util::make_unique<ExtractorCallbacks>(extraction_containers);
+        auto extractor_callbacks =
+            util::make_unique<ExtractorCallbacks>(extraction_containers, turn_lane_map);
```
daniel-j-h: I know this is not your change, but why is this a pointer?

daniel-j-h: Hmm looks like we're calling .reset below to delete the callbacks.. it's keeping the deduplication map around, only refs otherwise. Kind of ugly..

MoKob: Ticket it. We need to seriously clean the constructions anyways.

src/extractor/extractor.cpp
```diff
@@ -444,13 +445,21 @@ Extractor::BuildEdgeExpandedGraph(ScriptingEnvironment &scripting_environment,
...
+
+    // inplace prefix sum
+    std::partial_sum(turn_lane_offsets.begin(), turn_lane_offsets.end(), turn_lane_offsets.begin());
+
+    // allocate the current masks
+    std::vector<guidance::TurnLaneType::Mask> turn_lane_masks(turn_lane_offsets.back());
+    for (auto entry = turn_lane_map.begin(); entry != turn_lane_map.end(); ++entry)
+        std::copy(entry->first.begin(),
+                  entry->first.end(),
+                  turn_lane_masks.begin() + turn_lane_offsets[entry->second]);
```
daniel-j-h: Duplicate code below.

Also could you write a bit what's the idea behind the serialization / deserialization, as in: offsets, prefix sum, ..

src/extractor/extractor.cpp
```diff
@@ -633,5 +645,44 @@ void Extractor::WriteIntersectionClassificationData(
...
```
daniel-j-h: ^ dup. here

src/extractor/extractor.cpp
```diff
@@ -633,5 +645,44 @@ void Extractor::WriteIntersectionClassificationData(
...
+
+    util::SimpleLogger().Write() << "Writing turn lane masks...";
+    TIMER_START(turn_lane_timer);
+
+    std::ofstream ofs(turn_lane_file, std::ios::binary);
```
daniel-j-h:
```cpp
if (!ofs) throw osrm::exception();
```

src/extractor/extractor_callbacks.cpp
```diff
@@ -31,11 +31,15 @@ namespace extractor
...
-    : external_memory(extraction_containers)
+ExtractorCallbacks::ExtractorCallbacks(ExtractionContainers &extraction_containers,
+                                       guidance::LaneDescriptionMap &lane_description_map)
+    : lane_description_map(lane_description_map), external_memory(extraction_containers)
 {
     // we reserved 0, 1, 2 for the empty case
     string_map[MapKey("", "")] = 0;
+
+    // The map should be empty before we start initializing it
+    BOOST_ASSERT(lane_description_map.empty());
```
daniel-j-h: Hm doesn't it make more sense then to keep the map a member attribute and add a function so that the call site eventually can get access to it? (We destroy the callbacks, see comment above, but you can always move out the map)

src/extractor/guidance/turn_discovery.cpp
```diff
@@ -50,8 +50,10 @@ bool findPreviousIntersection(const NodeID node_v,
     // previous intersection.
     const auto straightmost_at_v_in_reverse =
         findClosestTurn(node_v_reverse_intersection, STRAIGHT_ANGLE);
-    if (angularDeviation(straightmost_at_v_in_reverse->turn.angle, STRAIGHT_ANGLE) >
-        FUZZY_ANGLE_DIFFERENCE)
+
+    // TODO evaluate if narrow turn is the right criterion here... Might be that other angles are
+    // valid
```
daniel-j-h: Is there a ticket for this / for all todos? Just to make sure we're not forgetting about those :)

src/extractor/guidance/turn_discovery.cpp
```diff
@@ -66,12 +68,25 @@ bool findPreviousIntersection(const NodeID node_v,
     // if the edge is not traversable, we obviously don't have a previous intersection or couldn't
     // find it.
     if (node_based_graph.GetEdgeData(result_via_edge).reversed)
+    {
+        result_via_edge = SPECIAL_EDGEID;
+        result_node = SPECIAL_NODEID;
```
daniel-j-h: Oops, shouldn't this already be backported to master asap as a bugfix?

src/extractor/guidance/turn_discovery.cpp
```diff
@@ -66,12 +68,25 @@ bool findPreviousIntersection(const NodeID node_v,
...
+    const auto check_via_edge =
+        result_intersection.end() !=
+        std::find_if(result_intersection.begin(),
+                     result_intersection.end(),
+                     [via_edge](const ConnectedRoad &road) { return road.turn.eid == via_edge; });
+
+    if (!check_via_edge)
+    {
+        result_via_edge = SPECIAL_EDGEID;
+        result_node = SPECIAL_NODEID;
```
daniel-j-h: Same here.

src/extractor/guidance/turn_lane_augmentation.cpp
```diff
@@ -280,6 +282,13 @@ LaneDataVector handleNoneValueAtSimpleTurn(LaneDataVector lane_data,
     // we have to reduce it, assigning it to neighboring turns
     else if (connection_count < lane_data.size())
     {
+        if( connection_count+1 < lane_data.size() ){
```
daniel-j-h: clang-format --- maybe do it once on all files that you changed before merging.

src/extractor/guidance/turn_lane_data.cpp
```diff
@@ -64,10 +64,12 @@ bool TurnLaneData::operator<(const TurnLaneData &other) const
            std::find(tag_by_modifier, tag_by_modifier + 8, other.tag);
 }
 
-LaneDataVector laneDataFromDescription(const TurnLaneDescription &turn_lane_description)
+LaneDataVector laneDataFromDescription(TurnLaneDescription turn_lane_description)
 {
     typedef std::unordered_map<TurnLaneType::Mask, std::pair<LaneID, LaneID>> LaneMap;
+    //TODO need to handle cases that have none-in between two identical values
```
daniel-j-h: Ticket?

src/extractor/extractor.cpp
MoKob: done. Github doesn't detect the move into a dedicated function

src/extractor/extractor.cpp
MoKob: also done.

src/extractor/guidance/turn_discovery.cpp
MoKob: since it only makes some turn lane cases (very rare) break, it is not a urgent fix. Still one of the reasons I want to have it in 5.4

src/extractor/guidance/turn_discovery.cpp
MoKob: Same here.

src/extractor/guidance/turn_lane_data.cpp
```diff
@@ -151,6 +153,27 @@ bool hasTag(const TurnLaneType::Mask tag, const LaneDataVector &data)
...
+        if (location == superset_candidate.end())
+            return false;
+
+        // compare the number of lanes TODO this might have be to be revisited for situations where
+        // a sliproad widens into multiple lanes
+        if ((location->to - location->from) != (entry.to - entry.from))
+            return false;
+    }
+    return true;
+}
```
daniel-j-h: I just checked out the branch, grepped for `isSubsetOf` but I'm unable to find a single callsite. Is this dead code? What's the idea behind this?

src/extractor/guidance/turn_lane_handler.cpp
```diff
@@ -57,54 +77,199 @@ TurnLaneHandler::TurnLaneHandler(const util::NodeBasedDynamicGraph &node_based_g
...
+                          << (int)previous_description_id << "\n"
+                          << std::endl;
+                util::guidance::printTurnAssignmentData(
+                    at, lane_data, intersection, node_info_list);
+
+                if (previous_node != SPECIAL_NODEID)
+                    util::guidance::printTurnAssignmentData(
+                        previous_node, previous_lane_data, previous_intersection, node_info_list);
+            }
+            */
```
daniel-j-h: Debugging, do you need this any longer?

src/extractor/guidance/turn_lane_handler.cpp
```diff
@@ -132,18 +298,31 @@ Intersection TurnLaneHandler::assignTurnLanes(const NodeID at,
...
+    // The turn a-b-c gets assigned the lanes of both *left* and *through*.
+    // At b-c, we get access to either a new set of lanes, or -- via the previous intersection
+    // -- to
+    // the second part of | left | through | right |. Lane anticipation can then deduce which
+    // lanes
+    // correspond to what and suppress unnecessary instructions.
+    //
+    // For our initial case, we consider only the turns that are available at the current
+    // location,
+    // which are given by partitioning the lane data and selecting the first part.
```
daniel-j-h: What happens when this is a mis-tagged by accident?

As in `ab` is tagged `left|through|right` and `bc` is tagged `left|left|through|right`.
Or even worse only `bc` has a psv lane tag going left?

src/extractor/guidance/turn_lane_handler.cpp
```diff
@@ -155,115 +334,73 @@ Intersection TurnLaneHandler::assignTurnLanes(const NodeID at,
...
+
+    return TurnLaneScenario::UNKNOWN;
+}
+
+void TurnLaneHandler::extractLaneData(const EdgeID via_edge,
+                                      LaneDescriptionID &lane_description_id,
+                                      LaneDataVector &lane_data) const
+{
+    const auto &edge_data = node_based_graph.GetEdgeData(via_edge);
+    // TODO access correct data
```
daniel-j-h: What do you mean by this?

src/extractor/guidance/turn_lane_handler.cpp
```diff
@@ -155,115 +334,73 @@ Intersection TurnLaneHandler::assignTurnLanes(const NodeID at,
...
+                                  turn_lane_masks.begin() +
+                                      turn_lane_offsets[lane_description_id + 1])
+            : TurnLaneDescription();
+
+    if (!lane_description.empty())
+        lane_data = laneDataFromDescription(lane_description);
+
+    BOOST_ASSERT(lane_description.empty() ||
+                 lane_description.size() == (turn_lane_offsets[lane_description_id + 1] -
+                                             turn_lane_offsets[lane_description_id]));
```
daniel-j-h: What about:

    if invalid:
      id = TurnLaneDescription()
    else
      id = TurnLaneDescription(...)
      assert description
      data = fromDescription(description)

src/extractor/guidance/turn_lane_handler.cpp
```diff
@@ -414,8 +551,10 @@ std::pair<LaneDataVector, LaneDataVector> TurnLaneHandler::partitionLaneData(
      * case left) turn lane as if it were to continue straight onto the intersection and
      * look back between (1) and (2) to make sure we find the correct lane for the left-turn.
      *
-     * Intersections like these have two parts. Turns that can be made at the first intersection and
-     * turns that have to be made at the second. The partitioning returns the lane data split into
+     * Intersections like these have two parts. Turns that can be made at the first intersection
+     * and
+     * turns that have to be made at the second. The partitioning returns the lane data split
+     * into
```
daniel-j-h: Hm clang-format can't do smart breaks, right?

src/extractor/guidance/turn_lane_handler.cpp
```diff
@@ -447,7 +587,8 @@ std::pair<LaneDataVector, LaneDataVector> TurnLaneHandler::partitionLaneData(
             continue;
 
         const auto best_match = findBestMatch(turn_lane_data[lane].tag, intersection);
-        if (isValidMatch(turn_lane_data[lane].tag, best_match->turn.instruction))
+        if (best_match->entry_allowed &&
```
daniel-j-h: Can `best_match` be the end iterator?

src/extractor/guidance/turn_lane_matcher.cpp
```diff
@@ -102,29 +102,34 @@ bool isValidMatch(const TurnLaneType::Mask &tag, const TurnInstruction instructi
     return false;
 }
 
+double getMatchingQuality( const TurnLaneType::Mask tag, const ConnectedRoad &road)
+{
+    const constexpr double idealized_turn_angles[] = {0, 35, 90, 135, 180, 225, 270, 315};
+    const auto idealized_angle = idealized_turn_angles[getMatchingModifier(tag)];
```
daniel-j-h: Can we assert on the size here?

Looking into this, the array here has 8 items, but `getMatchingModifier` returns 11 different items?

MoKob: It returns 11 values, out of 8 possible modifiers. Will try and add some assertion, though.

features/guidance/advanced-lanes.feature
```diff
@@ -184,3 +103,166 @@ Feature: Turn Lane Guidance
...
+
+        And the ways
+            | nodes | name   | highway     | oneway | turn:lanes:forward |
+            | ab    | ghough | secondary   | yes    |                    |
+            | bc    | ghough | secondary   | yes    | through\|through   |
+            | bd    | ghough | secondary   | yes    | none\|through      |
+            | def   | ghough | secondary   | yes    |                    |
+            | gd    | market | primary     | yes    |                    |
+            |  dc   | market | primary     | yes    |                    |
+            |   ch  | market | primary     | yes    |                    |
```
daniel-j-h: Spaces

features/guidance/advanced-lanes.feature
```diff
@@ -184,3 +103,166 @@ Feature: Turn Lane Guidance
...
```
daniel-j-h: Same here

features/guidance/advanced-lanes.feature
```diff
@@ -184,3 +103,166 @@ Feature: Turn Lane Guidance
...
```
daniel-j-h: And here. Is this the same test as above?

features/guidance/turn-lanes.feature
```diff
@@ -5,7 +5,7 @@ Feature: Turn Lane Guidance
         Given the profile "car"
         Given a grid size of 20 meters
 
-    @bug
+    @simple
```
daniel-j-h: Love it how it changes from bug to simple :)

features/guidance/turn-lanes.feature
```diff
@@ -350,10 +358,12 @@ Feature: Turn Lane Guidance
             | bg    | right |                     | no     |
 
         When I route I should get
-            | waypoints | route            | turns                    | lanes                            |
-            | a,f       | road,turn,turn   | depart,turn left,arrive  | ,left:true straight;right:false, |
-            | a,g       | road,right,right | depart,turn right,arrive | ,left:false straight;right:true, |
+            | waypoints | route            | turns                           | lanes                            |
+            | a,f       | road,turn,turn   | depart,turn left,arrive         | ,left:true straight;right:false, |
+            | a,d       | road,road,road   | depart,use lane straight,arrive | ,left:false straight;right:true, |
+            | a,g       | road,right,right | depart,turn right,arrive        | ,left:false straight;right:true, |
```
daniel-j-h: I don't understand the `straight` here. I expected `straight: true/false` to be honest. Did something change in the Cucumber parsing logic?

daniel-j-h: Nevermind, it's `straight;right : true`, I misread

features/guidance/turn-lanes.feature
```diff
@@ -466,12 +478,12 @@ Feature: Turn Lane Guidance
...
-            | bd    | road  | through\|through                    | yes    | primary   |
+            | bd    | road  | through\|through\|right             | yes    | primary   |
             | bc    | road  | left\|left                          | yes    | primary   |
             | de    | road  |                                     | yes    | primary   |
-            | fd    | cross |                                  |        | secondary |
-            |  dc   | cross |                                  |        | secondary |
-            |   cg  | cross |                                  |        | secondary |
+            | fd    | cross |                                     |        | secondary |
+            |  dc   | cross |                                     |        | secondary |
+            |   cg  | cross |                                     |        | secondary |
```
daniel-j-h: Spaces

src/extractor/guidance/turn_lane_data.cpp
```diff
@@ -67,7 +68,7 @@ bool TurnLaneData::operator<(const TurnLaneData &other) const
 LaneDataVector laneDataFromDescription(TurnLaneDescription turn_lane_description)
 {
     typedef std::unordered_map<TurnLaneType::Mask, std::pair<LaneID, LaneID>> LaneMap;
-    //TODO need to handle cases that have none-in between two identical values
+    // TODO need to handle cases that have none-in between two identical values
```
daniel-j-h: Is this ticketed? What happens on these, we ignore them?

src/extractor/guidance/turn_lane_handler.cpp
```diff
@@ -701,15 +700,12 @@ Intersection TurnLaneHandler::simpleMatchTuplesToTurns(Intersection intersection
...
-                                    const Intersection &previous_intersection,
-                                    const LaneDescriptionID &previous_lane_description_id,
-                                    const LaneDataVector &previous_lane_data)
+Intersection TurnLaneHandler::handleSliproadTurn(Intersection intersection,
+                                                 const LaneDescriptionID lane_description_id,
+                                                 LaneDataVector lane_data,
+                                                 const Intersection &previous_intersection)
 {
-    const auto sliproad_index =
+    const std::size_t sliproad_index =
```
daniel-j-h: Was this change needed or is it only for readability? (I'm fine with both, just wondering)

src/extractor/guidance/turn_lane_augmentation.cpp
```diff
@@ -283,27 +283,18 @@ LaneDataVector handleNoneValueAtSimpleTurn(LaneDataVector lane_data,
...
         // service vehicles. Otherwise, we should not have it
-        //
-        // TODO(mokob): #2730 have a look please
-        // BOOST_ASSERT(connection_count + 1 == lane_data.size());
-        //
-        if (connection_count + 1 != lane_data.size())
+        if (connection_count + 1 == lane_data.size())
         {
-            // skip broken intersections
+            lane_data = mergeNoneTag(none_index, std::move(lane_data));
```
daniel-j-h: Hm why does this bugfix include the pull request? Messed up rebase?

MoKob: drive by reformulation.

src/extractor/guidance/turn_lane_handler.cpp
MoKob: If ab is tagged `left|through|right`, only `right` will be used for the turn right. If `left|through` are used to reach `bc`. If `bc` is tagged including the `right`, it will not be possible to assign it, since it is missing the turn.
Given `left|left|through`, the turns can be matched correctly and `bc` will be used to announce the lanes.
It will be hard for your post-processing on lanes to find these cases, it doesn't matter for local processing though.

src/extractor/guidance/turn_lane_handler.cpp
MoKob: no, added an assert on that.

features/guidance/advanced-lanes.feature
MoKob: its a copy without a turn restriction/with turn restriction. adding clarification in title

features/guidance/turn-lanes.feature
MoKob: simple is only the classification, 1:1 matching of turn lanes :), not simple to handle

src/extractor/guidance/turn_lane_handler.cpp
MoKob: sign/vs unsigned comparison warnings. distance functions return signed values, even though we know it to be unsigned here (order of iterators)

PR #2619, opened 2016-07-04
[ready] Improve/sliproads

MoKob:
At the current state, some combinations of secondary roads / links and traffic lights can cause sliproad handling to fail and introduce unnecessary instructions.

An example is given here:
screen shot 2016-07-04 at 10 31 06
screen shot 2016-07-04 at 10 31 22
screen shot 2016-07-04 at 14 40 54

With the adjustments in this PR, we handle the sliproads at the given intersection correctly:

screen shot 2016-07-04 at 14 35 36
screen shot 2016-07-04 at 14 35 41

### DIFF
--- a/src/extractor/guidance/sliproad_handler.cpp
+++ b/src/extractor/guidance/sliproad_handler.cpp
@@ -1,0 +1,192 @@
+#include "extractor/guidance/constants.hpp"
+#include "extractor/guidance/intersection_scenario_three_way.hpp"
+#include "extractor/guidance/sliproad_handler.hpp"
+#include "extractor/guidance/toolkit.hpp"
+
+#include "util/guidance/toolkit.hpp"
+
+#include <limits>
+#include <utility>
+
+#include <boost/assert.hpp>
+
+using EdgeData = osrm::util::NodeBasedDynamicGraph::EdgeData;
+using osrm::util::guidance::getTurnDirection;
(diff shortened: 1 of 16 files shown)


src/extractor/guidance/sliproad_handler.cpp
```diff
@@ -0,0 +1,194 @@
...
+                                 const util::NodeBasedDynamicGraph &node_based_graph,
+                                 const std::vector<QueryNode> &node_info_list,
+                                 const util::NameTable &name_table,
+                                 const SuffixTable &street_name_suffix_table)
+    : IntersectionHandler(node_based_graph, node_info_list, name_table, street_name_suffix_table),
+      intersection_generator(intersection_generator)
+{
+}
+
+SliproadHandler::~SliproadHandler() {}
```
daniel-j-h: You can say

    ~T() = default;

in the header instead.

MoKob:
The code here overlaps with changes in the guidance-decision into profiles PR. It will need to be adjusted after a merge of https://github.com/Project-OSRM/osrm-backend/pull/2586

TheMarex:
Is this independent of #2586 and could go into 5.3.0 or should this wait for 6.0 as well?

MoKob:
Well. It requires data-reprocessing. So I would hold it for 5.4

PR #2620, opened 2016-07-04
fix obvious turn collapsing for straight turns

MoKob:
Obvious turn handling did not consider the origin of the road.
In combination with turn collapsing, we could end up seeing `turn straight` where `new name` should have been emitted.

This PR fixes this behaviour (see new cucumber case).

Behaviour changes from `depart,turn straight,arrive` into `depart,arrive`.

### DIFF
--- a/src/extractor/guidance/turn_handler.cpp
+++ b/src/extractor/guidance/turn_handler.cpp
@@ -1,7 +1,7 @@
-#include "extractor/guidance/turn_handler.hpp"
 #include "extractor/guidance/constants.hpp"
 #include "extractor/guidance/intersection_scenario_three_way.hpp"
 #include "extractor/guidance/toolkit.hpp"
+#include "extractor/guidance/turn_handler.hpp"
 
 #include "util/guidance/toolkit.hpp"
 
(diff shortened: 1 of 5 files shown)


src/extractor/guidance/turn_handler.cpp
```diff
@@ -88,15 +88,17 @@ Intersection TurnHandler::handleThreeWayTurn(const EdgeID via_edge, Intersection
...
         const bool is_obvious_by_road_class =
-            (!is_ramp && (2 * getPriority(first_class) < getPriority(second_class))) ||
+            (!is_ramp && (2 * getPriority(first_class) < getPriority(second_class)) &&
+             in_data.road_classification.road_class == first_class) ||
             (!isLowPriorityRoadClass(first_class) && isLowPriorityRoadClass(second_class));
 
         if (is_obvious_by_road_class)
             return true;
 
         const bool other_is_obvious_by_road_flass =
```
TheMarex: `flass`?

TheMarex:
Looks good to me. I think we can include this in `5.3.0`.

src/extractor/guidance/turn_handler.cpp
danpat: Given the continued use of this below, I think this might be an example of where YCM does more harm than good :-)

Issue #2617, opened 2016-07-02
Two osrm instances generated with same profile produce different /route API response when given same coordinates input


fubupc:
As the title said: two osrm instances generated with same profile and same osm.pbf file will produce different `/route` API response (`duration`, `distance`) when given same coordinates input.

I did some experiments as following:

1. Make two osrm instances with same osm.pbf and speed profile. Run them in different port.
2. Test `/route` API of both instances with exactly the same coordinates set.
3. Compare `duration`, `distance`  of both test set, you will find that most test produce same result, but a few of them have a random difference.

Some other observations:
1. An OSRM instance always returns same result for same input. (only `/route` API tested.)
2. Most of input will give same result among different OSRM instance. But different pairwise instances disagree on different subset of input.

PS: And I wrote some scripts (with sample data) to help reproducing the problem. please check it at: https://github.com/fubupc/osrm-test

danpat:
@fubupc This is possible where there are possible routes of the same **duration**.

During pre-processing (`osrm-extract`/`osrm-contract`) there is multi-threaded work occurring where it becomes possible that data does not get processed exactly the same each time you do it.  While it should be correct (we hope!) for actually finding the fastest path, if there are two possible paths of equal duration, there is no guarantee as to which one will be selected.

Note that this happens during **pre-processing**, not during an actual routing request.  Two `osrm-routed` instances using the same `.osrm` files should produce the same results.

Can you please check that your differing results have equal `duration` values?  If not, we might have a bug.  If `durations` are equal, then the behaviour is expected.

fubupc:
@danpat some tests did show different `duration` as well :(

Below are several testings I made:

```
    id     |           start            |            end             |    time compare     |   distance compapre   
----------------------------------------------------------------------------------------------------------
1655       | 106.99289800,-6.24160800   | 106.99268600,-6.24614600   |  +0.0   77.6 77.6   |     -0.3    815.9 816.2   
2125       | 106.87364700,-6.19923800   | 106.86550400,-6.20224800   |  +1.4    379 377.6  |     +0.0   4734.4 4734.4  
2413       | 106.87440500,-6.21346800   | 106.87220000,-6.21454800   |  +1.4   84.8 83.4   |     +0.0   1120.1 1120.1  
2516       | 106.72717900,-6.19996700   | 106.71997400,-6.19612800   |  +0.0  181.2 181.2  |     +2.0   1781.1 1779.1  
[...]

Issue #610, opened 2013-04-08
wrong name when way & area share nodes


emiltin:
in this example, a street "Hørsholmsgade" share some nodes with a pedestrian area:

http://www.ibikecph.dk/#!/x5nq4.7gumq/x5oc1.7gua7

Routing along the way you get:

start on Hørsholmsgade
Continue on pedestrian
Continue on Hørsholmsgade

So somehow it seems osrm chooses the pedestrian area over the street, and thus doesn't have any street name

TheMarex:
@emiltin This should be fixed as OSRM ignores areas now, right?

emiltin:
i hope osrm doen't ignore areas with highway=*

TheMarex:
You  are right, we only do this in the car profile.

systemed:
I wonder if #1463 might be the same issue as this.

TheMarex:
@systemed that seems more like a connectivity error caused by filtering out the wrong streets.

This bug here is basically caused by the fact that OSRM only allows for one edge between nodes. So if two ways overlap, the one with the smaller travel time gets chosen (including the name etc). However that part of the code was buggy and is currently addressed in #1450.

@emiltin The problem with allowing areas is that it will route on the area boundary. So in a lot of pedestrian areas that will lead to very strange results. There is no easy way to fix this. Routing across an areas is just too different. That is why you should always include a way inside the area that can be used for routing. Where do you see a problem with disabling areas for the bike and foot profile as well?

TheMarex:
Never mind, constructing a visibility graph is a lot easier than I thought. Routing over areas seems doable. Tracking here https://github.com/Project-OSRM/osrm-backend/issues/64

emiltin:
sounds good with routing on area. some profiles like bike and foot requires areas, and currently this means following the edge of the square, which is inelegant, but better than nothing.

the problem here is that sometimes osm contains ways along the edge of areas, and then osrm routes on the edge of the area, rather than the way, resulting in confusing instructions.

PR #2624, opened 2016-07-05
Fix ambiguous data issues

oxidase:
Added disambiguations for edge weights and names by taking minimal edge weight and minimal name id for edges with equal source and target nodes.

Solves  #2617

### DIFF
--- a/features/testbot/basic.feature
+++ b/features/testbot/basic.feature
@@ -266,3 +266,24 @@ Feature: Basic Routing
             | from | to | route | time |
             | a    | b  | ,     | 10s  |
             | b    | a  | ,     | 10s  |
+
+    Scenario: Ambiguous edge names - Use lexicographically smallest name
+        Given the node map
+            | a | b | c |
+
+        And the ways
+            | nodes | highway | name |
+            | ab    | primary |      |
+            | ab    | primary | Αβγ  |
+            | ab    | primary |      |
+            | ab    | primary | Abc  |
+            | ab    | primary |      |
+            | ab    | primary | Абв  |
+            | bc    | primary | Ηθι  |
+            | bc    | primary | Δεζ  |
+            | bc    | primary | Где  |
+
+        When I route I should get
+            | from | to | route       |
+            | a    | c  | Abc,Δεζ,Δεζ |
+            | c    | a  | Δεζ,Abc,Abc |
--- a/features/testbot/basic.feature
+++ b/features/testbot/basic.feature
@@ -233,7 +233,7 @@ Feature: Basic Routing
             | d    | a  | abcd,abcd  |
             | a    | m  | aeim,aeim  |
             | m    | a  | aeim,aeim  |
-    
+
     Scenario: Testbot - Triangle challenge
         Given the node map
             |   |   |   | d |
@@ -251,3 +251,18 @@ Feature: Basic Routing
             | from | to | route |
             | d    | c  | de,ce,ce |
             | e    | d  | de,de    |
+
+    Scenario: Ambiguous edge weights - Use minimal edge weight
+        Given the node map
+            | a | b |
+
+        And the ways
+            | nodes | highway   | name |
+            | ab    | tertiary  |      |
+            | ab    | primary   |      |
+            | ab    | secondary |      |
+
+        When I route I should get
+            | from | to | route | time |
+            | a    | b  | ,     | 10s  |
+            | b    | a  | ,     | 10s  |
--- a/include/extractor/internal_extractor_edge.hpp
+++ b/include/extractor/internal_extractor_edge.hpp
@@ -142,43 +142,6 @@ struct InternalExtractorEdge
     }
 };
 
-struct CmpEdgeByInternalStartThenInternalTargetID
-{
-    using value_type = InternalExtractorEdge;
-    bool operator()(const InternalExtractorEdge &lhs, const InternalExtractorEdge &rhs) const
-    {
-        return (lhs.result.source < rhs.result.source) ||
-               ((lhs.result.source == rhs.result.source) &&
-                (lhs.result.target < rhs.result.target));
-    }
-
-    value_type max_value() { return InternalExtractorEdge::max_internal_value(); }
-    value_type min_value() { return InternalExtractorEdge::min_internal_value(); }
-};
-
-struct CmpEdgeByOSMStartID
-{
-    using value_type = InternalExtractorEdge;
-    bool operator()(const InternalExtractorEdge &lhs, const InternalExtractorEdge &rhs) const
-    {
-        return lhs.result.osm_source_id < rhs.result.osm_source_id;
-    }
-
-    value_type max_value() { return InternalExtractorEdge::max_osm_value(); }
-    value_type min_value() { return InternalExtractorEdge::min_osm_value(); }
-};
-
-struct CmpEdgeByOSMTargetID
-{
-    using value_type = InternalExtractorEdge;
-    bool operator()(const InternalExtractorEdge &lhs, const InternalExtractorEdge &rhs) const
-    {
-        return lhs.result.osm_target_id < rhs.result.osm_target_id;
-    }
-
-    value_type max_value() { return InternalExtractorEdge::max_osm_value(); }
-    value_type min_value() { return InternalExtractorEdge::min_osm_value(); }
-};
 }
 }
 
--- a/src/extractor/extraction_containers.cpp
+++ b/src/extractor/extraction_containers.cpp
@@ -453,11 +453,13 @@ void ExtractionContainers::PrepareEdges(lua_State *segment_state)
                 all_edges_list[i].result.weight < min_forward_weight)
             {
                 min_forward_idx = i;
+                min_forward_weight = all_edges_list[i].result.weight;
             }
             if (all_edges_list[i].result.backward &&
                 all_edges_list[i].result.weight < min_backward_weight)
             {
                 min_backward_idx = i;
+                min_backward_weight = all_edges_list[i].result.weight;
             }
 
             // this also increments the outer loop counter!
--- a/src/extractor/extraction_containers.cpp
+++ b/src/extractor/extraction_containers.cpp
@@ -26,6 +26,8 @@
 
 namespace
 {
+namespace oe = osrm::extractor;
+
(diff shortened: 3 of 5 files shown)


src/extractor/extraction_containers.cpp
```diff
@@ -453,11 +453,13 @@ void ExtractionContainers::PrepareEdges(lua_State *segment_state)
                 all_edges_list[i].result.weight < min_forward_weight)
             {
                 min_forward_idx = i;
+                min_forward_weight = all_edges_list[i].result.weight;
```
TheMarex: This would mean the de-duplication code has been broken for a while?

include/extractor/internal_extractor_edge.hpp
```diff
@@ -142,14 +142,16 @@ struct InternalExtractorEdge
...
 {
     using value_type = InternalExtractorEdge;
     bool operator()(const InternalExtractorEdge &lhs, const InternalExtractorEdge &rhs) const
     {
         return (lhs.result.source < rhs.result.source) ||
                ((lhs.result.source == rhs.result.source) &&
-                (lhs.result.target < rhs.result.target));
+                (lhs.result.target < rhs.result.target ||
+                 (lhs.result.target == rhs.result.target &&
+                  lhs.result.name_id < rhs.result.name_id)));
```
TheMarex: Usually having a name is better than not having a name so inverting this is probably more what we want.

TheMarex:
Good catch. This must have been a very old bug, at least going back a year.

src/extractor/extraction_containers.cpp
oxidase: Yes, always the last edge was used. Because the order of edges was (source, target),  edge weights were "randomized" by scheduler.

include/extractor/internal_extractor_edge.hpp
oxidase: Ok, i will revert the order.

systemed:
Very interesting. I've occasionally found issues similar to #610 where OSRM appears to pick the less favourable way between two nodes - for example, at http://cycle.travel/map?from=51.4691,-0.2196&to=51.4707,-0.2216, where the middle section is seemingly following a highway=pedestrian area even though the road has a better weight. Hopefully this fixes it.

oxidase:
@systemed looks like it can be solved by PR.

emiltin:
Would be really nice if this could solve #610 - which is over 3 years old :-)

include/extractor/internal_extractor_edge.hpp
```diff
@@ -142,14 +142,16 @@ struct InternalExtractorEdge
...
 {
     using value_type = InternalExtractorEdge;
     bool operator()(const InternalExtractorEdge &lhs, const InternalExtractorEdge &rhs) const
     {
         return (lhs.result.source < rhs.result.source) ||
                ((lhs.result.source == rhs.result.source) &&
-                (lhs.result.target < rhs.result.target));
+                (lhs.result.target < rhs.result.target ||
+                 (lhs.result.target == rhs.result.target &&
+                  lhs.result.name_id > rhs.result.name_id)));
```
MoKob: the parsing happens asynchronously, doesn't it? So the edges arrive in a different order which should also result in name_ids being assigned randomly? Or am I missing something? NameID should not be possible to use here?

features/testbot/basic.feature
```diff
@@ -251,3 +251,34 @@ Feature: Basic Routing
...
+            | nodes | highway | name |
+            | ab    | primary | 1    |
+            | ab    | primary |      |
+            | ab    | primary | 2    |
+            | bc    | primary | 3    |
+
+        When I route I should get
+            | from | to | route |
+            | a    | c  | 2,3,3 |
+            | c    | a  | 3,2,2 |
```
MoKob: Could fail randomly, due to parallel parsing. `ab` are all identical, so they are differentiated on by the name_id which is random due to https://github.com/Project-OSRM/osrm-backend/blob/0e4061f85816b5499896022eb151a1402ae3aad0/src/extractor/extractor.cpp#L212 and its interaction with https://github.com/Project-OSRM/osrm-backend/blob/0e4061f85816b5499896022eb151a1402ae3aad0/src/extractor/extractor.cpp#L160

src/extractor/extraction_containers.cpp
MoKob: this value is never read? And the IDX is used to choose only the minimal edge. All others are invalidated. Does this even affect anything?

include/extractor/internal_extractor_edge.hpp
oxidase: Yes, it fixes only the original issue with an empty name and non-empty one.

MoKob: Well... fixes. It randomly chooses any of the non-empty names over an empty name. In that case, your test-case is still wrong, since it offers multiple possible names (50:50 chance)

features/testbot/basic.feature
oxidase: Frankly, i have no idea what to use for stable ordering of names. Would it make sense to take a lexicographically  smallest non-empty string?

MoKob: I think that might be an option. Otherwise we end up with just another undefined behaviour.

src/extractor/extraction_containers.cpp
oxidase: Yes, without the change min_forward_weight is always std::numeric_limits<int>::max()
and "all_edges_list[i].result.weight < min_forward_weight" is always true

MoKob: Ah yes. You are right. Loops and no coffee don't seem to mix.

MoKob: Can we add a test-case with multiple parallel speed-edges and empty name and check for speed?

features/testbot/basic.feature
oxidase: Ok, i will update PR with as suggested

oxidase:
Added a lexicographical compare for names to disambiguate edges. Empty strings are sorted as highest values. The price is additional memory used for "STXXLNameOffsets name_offsets".
I also changed data type to "unsigned char" because  the utf8 comparison with  char code units will sort strings incorrectly   in order Greek, Cyrillic, Latin.

src/extractor/extraction_containers.cpp
```diff
@@ -53,6 +112,12 @@ ExtractionContainers::ExtractionContainers()
     name_lengths.push_back(0);
     name_lengths.push_back(0);
 
+    // Insert three zero offsets for empty strings and the total length
+    name_offsets.push_back(0);
+    name_offsets.push_back(0);
+    name_offsets.push_back(0);
+    name_offsets.push_back(0);
+
```
danpat: Looks like there's one extra here?

oxidase: No, the last offset is always the total length of the character data array and at the moment it is zero. Offsets are added after pushing data into the data  array.

src/extractor/extraction_containers.cpp
oxidase: Ok, I have added empty names to the test

src/extractor/extraction_containers.cpp
```diff
@@ -53,6 +112,12 @@ ExtractionContainers::ExtractionContainers()
     name_lengths.push_back(0);
     name_lengths.push_back(0);
 
+    // Insert three zero offsets for empty strings and the total length
```
MoKob: this comment needs to be adjusted, if we push four zeros. It also requires an explanation of what the different indices are supposed to be.

src/extractor/extraction_containers.cpp
```diff
@@ -35,6 +37,63 @@ struct OSMNodeIDSTXXLLess
...
+            return lhs.result.target < rhs.result.target;
+
+        if (lhs.result.name_id == 0)
+            return false;
+
+        if (rhs.result.name_id == 0)
+            return true;
+
+        BOOST_ASSERT(!name_offsets.empty() && name_offsets.back() == name_data.size());
+        const oe::ExtractionContainers::STXXLNameCharData::const_iterator data = name_data.begin();
```
oxidase: @MoKob I made the original comparison based on name ids due to performance and efforts, but it is not correct anyway, because if there more than one non-empty names for one road the scheduler will randomize names and it will be a problem for the graph compressor and guidance to decide that it is the same road.

The statistic must be checked for such cases. For Jakarta case it is below 0.1% when it gets to names comparison, so performance impact is not so dramatic.

TheMarex: We can run some statistics on this. My feeling is that branch is almost never hit. This will only be used if there are multi-edges in OSM which is quite rare. Having consistent tie breaking is definitely desirable.

MoKob: Well. We can always risk it... Its just a chance of a potential performance bottle-neck. Maybe add some commentary around this to make clear that this could impact extract performance.

The problem why I insisted on having clear distinction is the parallel-edges tests. Because they could fail randomly without this destinction. So if we have tests that have this randomness-possibility, we need to get rid of randomness.

oxidase:
I have updated PR to avoid additional memory usage. Bad place now is a conversion from offsets to lengths for RangeTable in WriteCharData. Also lexicographical order is not  good for resolving street names ambiguity, and can result in some extremely rare strange street name changes.
For example, if edges have ambiguous names (A,C), (B,C), (D,C), (E,C) than names will be A,B,C,C.

I have also checked frequency of string comparisons: for oberbayern only 4 for 1466927 edges
```
1: Fischer-von-Erlach-Straße Lohensteinstraße
2: Lohensteinstraße Fischer-von-Erlach-Straße
3: Lohensteinstraße Fischer-von-Erlach-Straße
4: Maximilianstraße Römerkesselstraße (St 2055)
```

For Germany 600 comparisons for 24882388 edges
[strings.txt](https://github.com/Project-OSRM/osrm-backend/files/355503/strings.txt)

I don't understand at the moment why so many comparisons  for some street names like "St. Roman", but for me it looks like data problems.

TheMarex:
> upd: Added avoiding special node ids, so only 35 string comparisons left for 24882388 edges in Germany

That looks good to me. I don't think this will be noticable.

W.r.t the quality of the tie breaking: While a consitent naming for a street is desirable, I don't think these edge case warrants a more advanced technique.

PR #2629, opened 2016-07-06
Fix windows compilation with regard to struct/class packing

ssuluh:


include/util/struct_packing.hpp
```diff
@@ -0,0 +1,19 @@
...
+    Macro to pack struct/class in both environment
+    Usage:  PACK( 
+        struct MyStruct 
+        { 
+            int Property1;
+        });
+*/
+
+#ifdef _MSC_VER
+#define PACK( input ) __pragma( pack(push, 1) ) input __pragma( pack(pop) )
```
oxidase: gcc supports pragma(pack), so it is safe to use with gcc  4.4.7, icc 13.0, clang 3.0 and all  above version.

ssuluh: Thanks, just pushed an update and tested it locally.  I don't have unix dev setup so have to rely on your project travis ci to check them.

include/extractor/edge_based_graph_factory.hpp
```diff
@@ -46,28 +47,34 @@ namespace extractor
 namespace lookup
 {
 // Set to 1 byte alignment
+PACK(
```
oxidase: You can place here #pragma pack(push, 1)

include/extractor/edge_based_graph_factory.hpp
```diff
@@ -46,28 +47,34 @@ namespace extractor
...
 
+PACK(
 struct PenaltyBlock
 {
     std::uint32_t fixed_penalty;
     OSMNodeID from_id;
     OSMNodeID via_id;
     OSMNodeID to_id;
-} __attribute ((packed));
+}
```
oxidase: and here #pragma pack(pop)

It is special case for compatibility with MSVC
https://gcc.gnu.org/onlinedocs/gcc-5.4.0/gcc/Structure-Packing-Pragmas.html#Structure-Packing-Pragmas

ssuluh: Thanks @oxidase , I'll update it soon.  That was the effect I want to achieve, but it bites back on my face :(

include/util/struct_packing.hpp
TheMarex: I found that wen using the pragma version, gcc would not throw warnings if the struct was not packable (at least on 6.1) - I'm not sure if the pragma just does the packing switch anyway or if this would result in undefined behavior. Hence the switch to __attribute which complains about fishy stuff correctly in gcc. Maybe we should have a compiler specific switch in the define?

ssuluh: @TheMarex , when you said struct was not packable, do you mean when struct alignment in such manner there's no different in packed or default behavior? (aka sizeof is the same packed or not).  If that's the case please check this out under gcc 6.1, it doesn't look like it complaining either way:  http://melpon.org/wandbox/permlink/S8O81vF5WLqGHhIM

oxidase: The problem is not a packing by itself and also warnings are not very helpful. The real problem is in non-aligned access like in http://coliru.stacked-crooked.com/a/9fa8abe86e51c5ca
that will be a SIGBUS on some non-Intel platforms.

Such alignment problems can be detected  in run-time with ubsan, and there is no big difference between  silent packing with pragmas and ignored warnings.

ssuluh: @oxidase Thanks man, I think I understand now after catching up on some reading about SIGBUS and ubsan.  Tell you the truth I'm a .net guy and haven't touch c++ long time and no gcc either, so this is a good thing to know.  So I use your coliru linked and compare between #pragma pack vs. attribute, turn on verbosity, I don't see anything different in the compiler message between them and it's silent (http://coliru.stacked-crooked.com/a/9682ca9a0e1bb197).  And both trigger the ubsan with the -fsanitize=undefined just like your original link.

Now back to the pull request, my intention is to fix windows build so I can use it in my project, it seems to me now the question is beyond my PR, so should we do pack to begin with ?

TheMarex:
Going to merge this, but I'm sceptical if we should use packed structs as on-disk representations at all.

ssuluh:
@TheMarex, +1.  I think it's worth the core team decision between safety and performance vs. memory and disk space savings.

danpat:
I only placed it there for the disk space savings - no other reason.

If we ever want to handle cross-platform (i.e. different byte ordering or struct packing) file formats, we're going to have to be explicit about this anyway.  There's going to be some trade-off between file saving/reading performance, complexity and cross-platform support.

This method works reasonably well for semi-portable `mmap` support, although I agree it adds a bunch of ugliness :-(

Issue #2631, opened 2016-07-07
Update our Clang scripts to LLVM 4 and run them on Travis


daniel-j-h:
@kkaefer noted running the clang-modernize / tidy scripts in parallel garbles their output.

I had a look at them and this is only one of the problems: they are outdated for LLVM 3.8:
- https://github.com/Project-OSRM/osrm-backend/blob/master/scripts/analyze.sh
- https://github.com/Project-OSRM/osrm-backend/blob/master/scripts/modernize.sh
- https://github.com/Project-OSRM/osrm-backend/blob/master/scripts/tidy.sh

Problems:
- the modernize tool is gone now, it's included as separate checks in tidy
- there are more / different checks for tidy, and our `.clang-tidy` is outdated (similar to `.clang-format` that we updated to 3.8 a couple of days ago)
- we're not automatically running them on Travis

I think at this point we can steal^W get inspired by https://github.com/mapbox/mapbox-gl-native.

daniel-j-h:
These are huge improvements to clang-tidy:

> New misc-move-forwarding-reference check
> Warns when std::move is applied to a forwarding reference instead of std::forward.

> New misc-use-after-move check
> Warns if an object is used after it has been moved, without an intervening reinitialization.

> New cppcoreguidelines-slicing check
> Flags slicing of member variables or vtable.

http://releases.llvm.org/4.0.0/tools/clang/tools/extra/docs/ReleaseNotes.html

daniel-j-h:
Blocked by having LLVM 4 in Mason https://github.com/mapbox/mason/issues/385

springmeyer:
> Blocked by having LLVM 4 in Mason mapbox/mason#385

Done as per comment at https://github.com/Project-OSRM/osrm-backend/pull/3770#issuecomment-286615122

@daniel-j-h anything else needed here after #3770 and #3820 or can this be closed?

daniel-j-h:
Yeah we should update and integrate the LLVM 4 tools into the Travis builds.

PR #2633, opened 2016-07-07
Manual modernize run with LLVM 3.8

daniel-j-h:
More is blocked by:
- https://github.com/Project-OSRM/osrm-backend/issues/2631


daniel-j-h:
Failing on Travis because of [this issue with the Google apt repos](http://askubuntu.com/questions/766883/there-is-no-public-key-available-for-the-following-key-ids-1397bc53640db551).
(Why does Travs enable them by default?!) ..

TheMarex:
Looks good to me. :+1:

PR #2639, opened 2016-07-08
Add search_radius_multiplier param to /match endpoint

kerrick-lyft:
By default, OSRM /match considers a radius of 3 * gps_precision
around the input point when looking for candidate segments. Oftentimes
the 3x multiplier is too small and results in an incorrect match or an
error. Allow the multiplier to be passed in the URL.

N.B. if you double the value of this multiplier, it will make /match
 _16x_ slower. This is because doubling the query radius means there are
4x as many candidate segments (on average), and the Viterbi algorithm
is N^2 in the average number of hidden states per observation. Use with
care!

### DIFF
--- a/docs/http.md
+++ b/docs/http.md
@@ -259,14 +259,15 @@ http://{server}/match/v1/{profile}/{coordinates}?steps={true|false}&geometries={
 In addition to the [general options](#general-options) the following options are supported for this service:
 
 
-|Option      |Values                                          |Description                                                                               |
-|------------|------------------------------------------------|------------------------------------------------------------------------------------------|
-|steps       |`true`, `false` (default)                       |Return route steps for each route                                                         |
-|geometries  |`polyline` (default), `geojson`                 |Returned route geometry format (influences overview and per step)                        |
-|annotations |`true`, `false` (default)                       |Returns additional metadata for each coordinate along the route geometry.                |
-|overview    |`simplified` (default), `full`, `false`         |Add overview geometry either full, simplified according to highest zoom level it could be display on, or not at all.|
-|timestamps  |`{timestamp};{timestamp}[;{timestamp} ...]`     |Timestamp of the input location.                                                          |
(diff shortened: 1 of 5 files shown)


daniel-j-h:
Saying "use with case!" but then exposing this O(n^4) feature without limits to the user looks like a good way to allow for (unintentional) DOS scenarios. I think at least the engine config should allow for limiting this parameter on the server side (and set a sensible default limit, too).

TheMarex:
I don't see the point here? Why can't you just increase the radius value? The value 3 comes from the normal distribution that uses the radius to define its mean. Hence 3*mean gives you all the true coordinate with a probability of 0.998

kerrick-lyft:
@daniel-j-h I'll add a limit.

@TheMarex changing the radiuses changes the emission probabilities of the candidate segments in the Viterbi algorithm. This can result in a different match result.

Using 3x isn't enough for some of our testcases; we see bad matches or failed matches (no candidate segment found) even with 5x. A 0.998 percent chance that one point mismatches translates into a huge probability of a mismatch for a long route with 100s of points.

TheMarex:
> A 0.998 percent chance that one point mismatches translates into a huge probability of a mismatch for a long route with 100s of points.

It is a 1 - 0.998 = 0.002 -> 0.2% chance. If your trace data contains noise it is absolutely necessary to adjust the `radiuses` parameter to fit your data. The probability distribution _needs_ to be adjusted otherwise the model is just wrong.

The probability for one point to not get matched is `1 - 0.998^n`. That yields a chance of less than `50%` for one point of a 500 point trace to be matched to a wrong segment. You can run some tests on your data to see how often your points don't lie in the 3-sigma radius - if the rate is higher that suggests your mean (sigma) is different - hence you need to adjust the radius parameter.

If the correct segment is not in the search radius this will usually lead to this outlier being completely filtered out, since a mismatch will also make the transition probability very small. That is the whole point of using a Hidden Markov Model based approach - we can deal with this mismatches at the price of removing trace points.

Issue #2638, opened 2016-07-07
Handle lanes with different access designation


danpat:
http://wiki.openstreetmap.org/wiki/Lanes#Crossing_with_a_designated_lane_for_bicycles

Not sure how well this is tagged OSM-wide, but we should make sure to not accidentally put drivers in bike lanes if we're using the `car` profile.

```
lanes=3
turn:lanes=left|through|through|right
vehicle:lanes=yes|yes|no|yes
bicycle:lanes=yes|no|designated|yes
```

Even though `lanes=3`, there are 4 mentioned in `turn:lanes`.  For the `car` profile, we might need to compact `turn:lanes` based on `vehicle:lanes`.

/cc @1ec5 @MoKob @daniel-j-h

MoKob:
Not a big problem (500ish tags right now), but a rather simple fix. Will implement it just to be on the safe side.

MoKob:
This is only waiting for review/merge over in #2640. The problem itself is solved.

PR #2635, opened 2016-07-07
[ready] update cucumber to 1.2.1 and remove hack

MoKob:
Adresses https://github.com/Project-OSRM/osrm-backend/issues/2632

PR #2640, opened 2016-07-08
[approved-by-daniel-j-h] handle access flags for lanes

MoKob:
Addresses https://github.com/Project-OSRM/osrm-backend/issues/2638.

When merged, https://github.com/Project-OSRM/osrm-backend/pull/2586 needs to be adjusted to move the handling into the guidance lib.

Needs to be adjusted when adding: https://github.com/Project-OSRM/osrm-backend/pull/2635 to remove '&' hack from `extractor/guidance/toolkit.hpp` and `features/guidance/turn-lanes.feature:29`

### DIFF
--- a/include/extractor/guidance/toolkit.hpp
+++ b/include/extractor/guidance/toolkit.hpp
@@ -27,6 +27,7 @@
 #include <boost/algorithm/string.hpp>
 #include <boost/algorithm/string/predicate.hpp>
 #include <boost/functional/hash.hpp>
+#include <boost/tokenizer.hpp>
 
 namespace osrm
 {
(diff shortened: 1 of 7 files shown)


TheMarex:
#2635 landed this probably needs adjusting. In general is this safe for 5.3.0?

MoKob:
You are right on the adjustment (will do that on the one location). I would keep it for 5.4, since the change would require a new dataset (or would be without effect).

//cc @karenzshea needs profile update on merge

MoKob:
The hack is gone. In addition, this will need adjustment when https://github.com/Project-OSRM/osrm-backend/pull/2586 lands, since the adjustment is not considered in there yet.

include/extractor/guidance/toolkit.hpp
```diff
@@ -533,6 +534,46 @@ trimLaneString(std::string lane_string, std::int32_t count_left, std::int32_t co
...
+    tokenizer tokens(lane_string, sep);
+    tokenizer access(access_tokens, sep);
+
+    // strings don't match, don't do anything
+    if (std::distance(std::begin(tokens), std::end(tokens)) !=
+        std::distance(std::begin(access), std::end(access)))
+        return lane_string;
+
+    std::string result_string = "";
+    const std::string yes = "yes";
```
daniel-j-h: can be static

profiles/car.lua
```diff
@@ -205,15 +205,29 @@ local function getTurnLanes(way)
...
             turn_lanes_fw = trimLaneString(turn_lanes_fw, bw_psv, fw_psv )
         end
-        --backwards turn lanes need to treat bw_psv as fw_psv and vice versa
-        if  turn_lanes_bw and turn_lanes_bw ~= ""  then
+    end
+    --backwards turn lanes need to treat bw_psv as fw_psv and vice versa
+    if turn_lanes_bw and turn_lanes_bw ~= ""  then
+        if vehicle_lanes_bw and vehicle_lanes_bw ~= "" then
+            turn_lanes_bw = applyAccessTokens(turn_lanes_bw,vehicle_lanes_bw)
+        elseif  fw_psv ~= 0 or bw_psv ~= 0  then
```
daniel-j-h: this is three times more or less the same code, can we abstract this into a function?

daniel-j-h:
Looks good!

PR #2647, opened 2016-07-11
Switch profiles from Lua to library interface

### LINKED ISSUE #1974: Switch profiles from Lua to library interface

kkaefer:
There's now an abstracted interface and no direct calls to Lua anymore.

fixes #1974 

/cc @TheMarex @daniel-j-h

### DIFF
--- a/include/extractor/scripting_environment.hpp
+++ b/include/extractor/scripting_environment.hpp
@@ -1,52 +1,70 @@
 #ifndef SCRIPTING_ENVIRONMENT_HPP
 #define SCRIPTING_ENVIRONMENT_HPP
 
+#include "extractor/guidance/turn_lane_types.hpp"
+#include "extractor/internal_extractor_edge.hpp"
 #include "extractor/profile_properties.hpp"
-#include "extractor/raster_source.hpp"
+#include "extractor/restriction.hpp"
 
-#include "util/lua_util.hpp"
+#include <osmium/memory/buffer.hpp>
 
-#include <memory>
-#include <mutex>
+#include <boost/optional/optional.hpp>
+
+#include <tbb/concurrent_vector.h>
+
 #include <string>
-#include <tbb/enumerable_thread_specific.h>
+#include <vector>
 
-struct lua_State;
+namespace osmium
+{
+class Node;
+class Way;
+}
 
 namespace osrm
 {
+
+namespace util
+{
+struct Coordinate;
+}
+
 namespace extractor
 {
 
+class RestrictionParser;
+struct ExtractionNode;
+struct ExtractionWay;
+
 /**
- * Creates a lua context and binds osmium way, node and relation objects and
- * ExtractionWay and ExtractionNode to lua objects.
- *
- * Each thread has its own lua state which is implemented with thread specific
- * storage from TBB.
+ * Abstract class that handles processing osmium ways, nodes and relation objects by applying
+ * user supplied profiles.
  */
 class ScriptingEnvironment
 {
   public:
-    struct Context
-    {
-        ProfileProperties properties;
-        SourceContainer sources;
-        util::LuaState state;
-    };
-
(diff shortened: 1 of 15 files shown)


include/extractor/scripting_environment.hpp
```diff
@@ -1,23 +1,50 @@
...
+    virtual void processSegment(const osrm::util::Coordinate &source,
+                                const osrm::util::Coordinate &target,
+                                double distance,
+                                InternalExtractorEdge::WeightData &weight) = 0;
+
+    ProfileProperties properties;
+};
+
 /**
  * Creates a lua context and binds osmium way, node and relation objects and
```
TheMarex: Doc is a little bit outdated now.

include/extractor/scripting_environment_lua.hpp
```diff
@@ -0,0 +1,69 @@
...
+{
+
+struct LuaScriptingContext final : public ScriptingContext
+{
+    LuaScriptingContext() = default;
+    ~LuaScriptingContext() override = default;
+
+    std::unordered_set<std::string> getNameSuffixList() override;
+    std::vector<std::string> getExceptions() override;
+    void setupSources() override;
```
TheMarex: OSRM ties to use the clang naming convention with capitalized member functions.

`void SetupSources()`

TheMarex:
Looks good. This shouldn't change any functionality and only add a slight overhead for virtual function calls but is much easier to reason about

Feeling good about including this after 5.3.0 since API doesn't change.

include/extractor/scripting_environment_lua.hpp
kkaefer: Changed.

include/extractor/scripting_environment.hpp
kkaefer: Changed

TheMarex:
This is included in master now. :tada:

springmeyer:
I re-read this PR in the contex of #2926. I cannot see any obvious problems that would cause an increase in memory usage. But I notice that locks are used when the `GetLuaContext` is called. Therefore I wonder if `GetLuaContext` were called a more often then more locking could cause less efficient allocation and therefore memory growth. Just a hunch, but going on this hunch I looked in the PR to see if more locks could be happening...

It seems like the answer is yes, but not many more. Would appreciate others eyes to rule this out. For example [here the context was previously accessed once and then used in `util::luaFunctionExists` and `RestrictionParser` directly](https://github.com/Project-OSRM/osrm-backend/pull/2647/files#diff-edf14d92e413123a2b940ab0b8491ffbL111). After this PR the code is much cleaner but the context is accessed once when [`SetupSources()` is called](https://github.com/Project-OSRM/osrm-backend/pull/2647/files#diff-edf14d92e413123a2b940ab0b8491ffbR107) and a second time when [`RestrictionParser` is created (due to the usage of scripting_environment.GetProfileProperties() inside its constructor)](https://github.com/Project-OSRM/osrm-backend/pull/2647/files#diff-edf14d92e413123a2b940ab0b8491ffbR133). Now, creating the context twice rather than once may have no realworld performance penalty at all. But maybe it could? Or maybe there are other spots in the code that more locks are happening that I don't notice?

Issue #2570, opened 2016-06-21
Move shared memory block swapping control to osrm-datastore


danpat:
OSRM supports a hot-swap feature - you can use `osrm-datastore` to replace an old dataset with a new one while `osrm-routed` (or `node-osrm`) remains running, with only a very brief hitch in request handling performance.

However, the design is a bit ugly - `osrm-datastore` simply loads the new data, but it's actually up to the reader process (`osrm-routed`) to perform the data exchange, mark the new block in use and release the old memory.

This has a few problems:
  - the locking logic is hard to grasp
  - you can't have multiple readers of the shared memory block, *and* support hot-swapping
  - readers need write-access to the block

We should change this.  I picture it working as follows:

  - shared memory segments must be named via the command-line, no more "magic numbers"
  - users of a shared memory block register their interest via a semaphore
  - `osrm-datastore` loads new data, then signals all registered listeners that new data is available, then waits
  - listeners swap to the new data address, and release interest in the old one
  - `osrm-datastore` removes the old data once all listeners have moved, then exits

There would be some additional locking required to avoid problems with running additional `osrm-datastore` during an exchange, and to handle the startup/shutdown of listeners during an exchange.

emiltin:
related #1221

TheMarex:
As a first step I would like to implement this without having named memory segments, since this would definitely require an API change. I believe to get this right we cannot operate on process granularity but need to consider every thread.

Currently we have a piece of code that waits until all requests are finished before it swaps the data. However that is unnecessary, what we really want to do is to wait until all requests that referenced the old dataset are finished, before we deallocate that shared memory region.

To allow this we would need to have multiple versions of the shared-memory data facade in memory. Every request holds shared ownership of that data facade. If all ownerships are given up the facade is deallocated, and we can signal for the release of the shared memory region.

Things that need to change to enable this:

1. All stored references to data facades (e.g. for routing algorithms or inside the plugins) need to be changed to non-owning references when the request is executed
2. The `CheckAndReleadFacade()` logic can't live in the shared dataface anymore, because every datafacade is immutable
3. Engine will store a `shared_ptr` to the current datafacade, which will be protected by a (process local) read-writer lock. Every time a new facade is create a write-lock will need to be aquired.
4. Every request will first copy (acquire&release read lock) that `shared_ptr` and hold it for the duration of the whole request.
[...]

PR #2653, opened 2016-07-13
Add -fsanitizer=address for debug builds in travis

TheMarex:
- [x] Fix https://github.com/Project-OSRM/osrm-backend/issues/2788
- [x] Figure out why https://travis-ci.org/Project-OSRM/osrm-backend/jobs/153525916 is using the sanitizer
- [x] Review
- [x] Address comments.


CMakeLists.txt
```diff
@@ -54,6 +54,7 @@ option(ENABLE_ASSERTIONS OFF)
 option(COVERAGE OFF)
 option(SANITIZER OFF)
 option(ENABLE_LTO "Use LTO if available" ON)
+option(ENABLE_SANITIZER "Use memory sanitizer for Debug build" OFF)
```
daniel-j-h: Instead of hardcoding options for asan / msan / tsan / ubsan / ... it's probably cleaner to just set the env var

    CXXFLAGS=-fsanitize=X

and let CMake pick it up automatically.

TheMarex:
This fails because `osrm-extract` is triggers the sanitizer https://travis-ci.org/Project-OSRM/osrm-backend/jobs/153525915#L1824-L1882 which is tracked https://github.com/Project-OSRM/osrm-backend/issues/2788

TheMarex:
We are currently hitting http://stackoverflow.com/questions/37603238/fsanitize-not-using-gold-linker-in-gcc-6-1 with GCC. Switching to clang now and see if that works.

oxidase:
@TheMarex the problem was in linking example without gold against OSRM libraries built with gold. It is only ubuntu gcc 6.1 issue. gcc 6.2 works fine. I have committed to my repository https://github.com/oxidase/osrm-backend/tree/build/add-sanitize to avoid forced push to upstream.

There were 2 problems:
- lazy TBB init without cleanup -> added explicit init
- shared memory leaks to prevent -> it is just bad implementation and must be  fixed in #2570

If it looks ok, i can squash commits and push into upstream.

TheMarex:
@oxidase thanks for looking into this! :+1: I added some comments on your branch. You can pull in the changes here if you think my comments are resolved.

oxidase:
The issue /usr/bin/ld: unrecognized option '--push-state' is due to missing '-fuse-ld=gold' in the example  that is built with gcc-6.2 from the toolchain PPA and ld 2.24 from trusty that are not fully compatible. There is a problem with binutils 2.24 that do not support the option introduced by asan. 

I have added a binutils PPA that bumps version  to 2.27 for the debug asan build. It can be removed when binutils in the toolchain PPA will be updated to at least 2.26 or travis will switch to xenial.

TheMarex:
@oxidase thanks for looking into this again. Great to see this working finally. :100: 

I'm going to do the following things here:
- Split the commit up into different commit for build system/travis changes
- Split the code changes from the commit into an own commit

If we wait for #3015 the shared memory fixes are obsolete anyway since the whole ownership mode got reworked.

Issue #2645, opened 2016-07-11
Allow Turning from None-Lanes / Consider Target Road Lanes


MoKob:
The `none` tag for lanes, while technically correct, is hard to handle in osrm backend.

At some situations, the none simply represents straight -- e.g. [here](http://www.openstreetmap.org/#map=19/52.31853/13.54370)--,

at other locations it is possible to take a left/right turn and go straight -- e.g. [here](http://www.openstreetmap.org/#map=19/52.56054/13.32152).

To be able to handle the tag correctly, we need to incorporate the number of available lanes at the destination road to compare it to the number of lanes in the turn.
If the number of lanes in the turn is smaller than the available lanes at the destination and the turn is directly next to a `none` tag, we should be able to augment the number of possible lanes to turn from.

This requires further investigation, though, since I am not sure if we can accurately capture all allowed cases and distinguish them from cases where it might be forbidden.

The drawback if we don't handle these cases: single lanes (like in the second example) would be completely excluded from our lane guidance. The example at the Kurt-Schuhmacher-Damm would end up never recommending the middle lane at all.

MoKob:
Part of the work (having lanes available at all) is already done in https://github.com/Project-OSRM/osrm-backend/pull/2779.
It would be easier to have #2779 merged before starting on this task.

PR #2656, opened 2016-07-14
adding a todo testscase / regression test-case

MoKob:
From two test-drives yesterday, @freenerd and me discovered two scenarios to consider.

This PR adds testcases for both, one preventing possible regression, one considering a scenario that is tracked in https://github.com/Project-OSRM/osrm-backend/issues/2645 (and tagged in the case accordingly).


TheMarex:
Looks good to me. :+1: Merge away.

PR #2657, opened 2016-07-14
[Ready] Emit Merge for Motorways only

MoKob:
The merge paradigm in its current implementation seems confusing. 

In cases where we take a very slight turn onto a through street, the instruction is technically a merge onto a street.
However, the merge paradigm of merging to the left on a slight right turn seems to confuse more than it helps.

This PR changes the paradigm of merge to only consider motorway-like ramps for merging and emitting an instruction to `go straight` onto these merging situations.

It is based on https://github.com/Project-OSRM/osrm-backend/pull/2586

### DIFF
--- a/include/util/debug.hpp
+++ b/include/util/debug.hpp
@@ -5,6 +5,7 @@
 #include "extractor/guidance/turn_lane_data.hpp"
 #include "extractor/query_node.hpp"
 #include "engine/guidance/route_step.hpp"
+#include "util/node_based_graph.hpp"
 #include "util/typedefs.hpp"
 
 #include <iomanip>
@@ -59,6 +60,20 @@ inline void print(const extractor::guidance::Intersection &intersection)
     std::cout << "  Intersection:\n";
     for (const auto &road : intersection)
         std::cout << "\t" << toString(road) << "\n";
+    std::cout << std::flush;
+}
+
+inline void print(const NodeBasedDynamicGraph &node_based_graph,
+                  const extractor::guidance::Intersection &intersection)
+{
+    std::cout << "  Intersection:\n";
+    for (const auto &road : intersection)
+    {
+        std::cout << "\t" << toString(road) << "\n";
+        std::cout << "\t\t"
+                  << node_based_graph.GetEdgeData(road.turn.eid).road_classification.ToString()
+                  << "\n";
+    }
     std::cout << std::flush;
 }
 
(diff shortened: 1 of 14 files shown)


src/extractor/guidance/intersection_handler.cpp
```diff
@@ -95,9 +94,43 @@ TurnInstruction IntersectionHandler::getInstructionForObvious(const std::size_t
...
+                         out_data.road_classification.IsRampClass())
+                {
+                    if (in_mode == out_mode)
+                        return {TurnType::Suppressed, getTurnDirection(road.turn.angle)};
+                    else
+                        return {TurnType::Notification, getTurnDirection(road.turn.angle)};
+                }
+                else
+                {
+                    const double constexpr MAX_COLLAPSE_DSTANCE = 30;
```
daniel-j-h: `I` missing in `DSTANCE`

include/util/debug.hpp
```diff
@@ -62,6 +63,16 @@ inline void print( const extractor::guidance::Intersection & intersection )
     std::cout << std::flush;
 }
 
+inline void print( const NodeBasedDynamicGraph &node_based_graph, const extractor::guidance::Intersection & intersection )
```
daniel-j-h: clang-format

daniel-j-h:
Resolved the two nitpicks I had with this.
@TheMarex can you have a last look, I think this is good to go in now.

src/extractor/guidance/intersection_handler.cpp
```diff
@@ -93,9 +94,43 @@ TurnInstruction IntersectionHandler::getInstructionForObvious(const std::size_t
...
+                    const double constexpr MAX_COLLAPSE_DISTANCE = 30;
+                    // in normal road condidtions, we check if the turn is nearly straight.
+                    // Doing so, we widen the angle that a turn is considered straight, but since it
+                    // is obvious, the choice is arguably better.
+
+                    // FIXME this requires https://github.com/Project-OSRM/osrm-backend/pull/2399,
+                    // since `distance` does not refer to an actual distance but rather to the
+                    // duration/weight of the traversal. We can only approximate the distance here
+                    // or actually follow the full road. When 2399 lands, we can exchange here for a
+                    // precalculated distance value.
```
TheMarex: I don't see how 2399 helps here. On the contrary the distance here seems more correct?

PR #2658, opened 2016-07-15
Correctly cache ccache directory

springmeyer:
Looking at the logs for the latest builds indicates that only the `./test/cache` is being added to the cache and not the ccache compiler data or apt data. The logs (https://travis-ci.org/Project-OSRM/osrm-backend/jobs/144440422#L1067) show only:

```
adding /home/travis/build/Project-OSRM/osrm-backend/test/cache to cache
```

The lack of apt caching may be because travis does not support this in open source repos. The docs at https://docs.travis-ci.com/user/caching/ have no mention of the `apt` keyword.

The lack of ccache caching is because of at least two reasons:

- Combining `ccache: true` with `directories` appears not to work and `ccache: true` is ignored.
- ~~Even if `ccache: true` were not ignored then the path is incorrect because it is customized by the `CCACHE_TEMPDIR` environment setting~~ False alarm: this temp directory does not change the default data path: that is still correctly `~/.ccache`.

This PR therefore fixes this situation by:

 - Removing the unsupported `apt` keyword
 - Adding the correct, default ccache data directory to the directions being cached.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,8 +15,8 @@ branches:
     - master
 
 cache:
+  ccache: true
   directories:
-    - $HOME/.ccache
     - test/cache
 
 env:
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,16 +15,16 @@ branches:
     - master
 
 cache:
-  ccache: true
-  apt: true
   directories:
+    - $HOME/.ccache
     - test/cache
 
 env:
   global:
-   - CCACHE_TEMPDIR=/tmp/.ccache-temp
    - CCACHE_COMPRESS=1
    - JOBS=4
+   - CASHER_TIME_OUT=1000
+   - CCACHE_TEMPDIR=/tmp/.ccache-temp
 
 matrix:
   fast_finish: true
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,8 +15,8 @@ branches:
     - master
 
 cache:
-  ccache: true
   directories:
+    - $HOME/.ccache
     - test/cache
 
 env:


daniel-j-h:
Hmm does it make sense to cache apt packages too, then? That is `/var/cache/apt/`.
As long as we're on the same Travis build env. packages shouldn't change for us after all.

springmeyer:
> Hmm does it make sense to cache apt packages too, then? That is /var/cache/apt/.

@daniel-j-h - this is what happens if you try to cache something that requires root access:

```
adding /var/cache/apt to cache
apt is not yet cached
/var/cache/apt/archives/lock: Permission denied
running `casher add/home/travis/.ccache test/cache /var/cache/apt` took longer than 180 seconds and has been aborted
```

https://travis-ci.org/Project-OSRM/osrm-backend/jobs/144988818#L1077

springmeyer:
Noticed another problem that was lurking: After enabling `~/.ccache` for the travis cacher it fails quietly. If you open up the `store build cache` section in the logs you will see:

```
running `casher push` took longer than 180 seconds and has been aborted
```

Hoping that ac87194 will fix this.

springmeyer:
OSX is breaking slightly differently, not yet sure if this is the same timeout problem or not:

```
store build cache
0.01s
0.51sWarning! PATH is not properly set up, '/Users/travis/.rvm/gems/ruby-2.0.0-p648/bin' is not at first place,
         usually this is caused by shell initialization files - check them for 'PATH=...' entries,
         it might also help to re-add RVM to your dotfiles: 'rvm get stable --auto-dotfiles',
         to fix temporarily in this shell session run: 'rvm use ruby-2.0.0-p648'.
/Users/travis/.casher/bin/casher:141:in `read': No such file or directory - /Users/travis/.casher/paths (Errno::ENOENT)
	from /Users/travis/.casher/bin/casher:141:in `changed?'
	from /Users/travis/.casher/bin/casher:114:in `push'
	from /Users/travis/.casher/bin/casher:52:in `block in run'
	from /Users/travis/.rvm/rubies/ruby-1.9.3-p551/lib/ruby/1.9.1/timeout.rb:69:in `timeout'
	from /Users/travis/.casher/bin/casher:52:in `run'
	from /Users/travis/.casher/bin/casher:234:in `<main>'
```

springmeyer:
> OSX is breaking slightly differently, not yet sure if this is the same timeout problem or not:

It was the same problem, so also fixed by ac87194

springmeyer:
Noting problems still:

- Fixing this means that more time is spent uploading the cache. This has not noticeably impacted the build times for linux jobs, however it caused the OSX job to go from 25m -> 40m.
- The OS X jobs is still timing out during cache unpacking: `running casher add/Users/travis/.ccache test/cache took longer than 500 seconds and has been aborted`

springmeyer:
Rebased and should be ready to merge. Will help speed up travis builds significantly. Ready for 👀  /cc @TheMarex @daniel-j-h

TheMarex:
Are you sure `ccache: true` does not work with `directories`? Looking at the output from `ccache -s` the cache seems to be used in both cases:

Before: https://travis-ci.org/Project-OSRM/osrm-backend/jobs/146043833#L1546
This PR: https://travis-ci.org/Project-OSRM/osrm-backend/jobs/146089143#L1551

springmeyer:
@TheMarex - What I see is that for that older/before job (https://travis-ci.org/Project-OSRM/osrm-backend/jobs/146043833#L1069):

  - `~/.ccache` it is not listed here during cache unpacking: https://travis-ci.org/Project-OSRM/osrm-backend/jobs/146043833#L1069 but the `test/data` directory is.
  - There is an error of `running `casher push` took longer than 180 seconds and has been aborted` at https://travis-ci.org/Project-OSRM/osrm-backend/jobs/146043833#L3988

So it seems like there are two possibilities:

  - a) `cache: ccache` is broken as I first guessed
  - b) `cache:ccache` is working (the syntax is valid). But there is no ccache data in the cache because the only existing cache that succeeded in being upload only included `./test/data` (maybe before ccache cashing was added). Then, no cache has successfully been uploaded because of the cacher timeout issue.

springmeyer:
@TheMarex I tested further, pushing two jobs to travis. They both have the Cashing timeout fix and are pulling valid caches. They differ as one used `ccache: true` with `directories` and the other used only `directories` while manually supplying the `~/.ccache` directory.

It looks like:

  - Both builds take about the same amount of time, so it is likely that the `ccache: true` is working
  - Only the build that supplies `~/.ccache` as a directory reports that `adding /home/travis/build/Project-OSRM/osrm-backend/test/cache to cache`.

So, it seems like when we provide directories we get confirmation in the logs that the cacher has found them and is unpacking them. When using `ccache: true` there is no log output confirming or denying that things are worked.

springmeyer:
Concluding that the key change in this PR is the increase in cacher timeout. I'm going to close this one - let's merge #2684 instead.

PR #2659, opened 2016-07-15
[ready] make roundabout maneuvers continuous with respect to lane changes

MoKob:
In anticipating lanes over multiple turns, we consider different segments of the route.
If two turns are far enough apart, we assume a lane change to be possible. This is, of course, only an initial heuristic and should be improved upon.

For roundabouts we are in one of these cases that are suffering from the basic structure of a fixed constant that we currently apply. Roundabouts are essentially one continuous turn. As a result, we need to -- at least -- consider longer times.

This PR adjusts the fixed constant to be a parameter as a first improvement. We probably want to be able to supply heuristic functions later on, even further moving away from the current situation.

### DIFF
--- a/src/engine/guidance/lane_processing.cpp
+++ b/src/engine/guidance/lane_processing.cpp
@@ -1,3 +1,4 @@
+#include "util/debug.hpp"
 #include "util/for_each_pair.hpp"
 #include "util/group_by.hpp"
 #include "util/guidance/toolkit.hpp"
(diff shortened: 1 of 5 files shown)


src/engine/guidance/lane_processing.cpp
```diff
@@ -1,3 +1,4 @@
+#include "util/debug.hpp"
```
TheMarex: This is left-over.

TheMarex:
Looks good to me. Can be included in 5.3.0

PR #2662, opened 2016-07-15
Use JavaScript for profiles

kkaefer:
This pull requests adds support for using Node.js/JavaScript code for profiles, eliminating the need for Lua.

It works by creating a node.js C++ module that runs the extractor, and by passing in a scripting environment. The extractor is run in a thread pool thread, and whenever any profile functions are invoked, we are dispatching to the main thread to run the assigned JS function. The C++ structures `ExtractionNode/Way` and the Osmium structures are bound to JS objects and passed as parameters.

There are a few caveats:
- Callbacks must run synchronously; there's currently no way to asynchronously process a way, node or turn penalty. This means you can't use any sort of async I/O functions, such as file operations and in particular network connections.
- JavaScript uses `camelCase` names for result properties instead of the `under_score` naming style; make sure you change the names when porting Lua profiles
- The `car.js` profile I ported takes around 28 seconds on my example OSM extract, whereas the Lua version takes only 9 seconds.
- JavaScript callbacks are only run on the main thread since Node.js is single-threaded. Lua profiles are run in their respective thread, so they run in parallel.
- Changes CMake requirement to 3.1 because it is required for `node-cmake`.
- The Node module code is a monolithic file and needs cleanup and splitting up into multiple files.

To try out the JavaScript version, run `make osrm-process`, and run `bin/osrm-extract -p profiles/car.js inputfile`

cc @daniel-j-h @TheMarex

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 2.8.11)
+cmake_minimum_required(VERSION 3.1)
 # we depend on 2.8.11 introducing target_include_directories
 
 if(CMAKE_CURRENT_SOURCE_DIR STREQUAL CMAKE_CURRENT_BINARY_DIR AND NOT MSVC_IDE)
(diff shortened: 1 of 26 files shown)


emiltin:
extraction going from 9 secs to 28 secs seems worrying? do you expect this can be improved?

vkurchatkin:
It makes sense as long as single node process is used.

TheMarex:
This was a great proof-of-concept and is superseded by https://github.com/Project-OSRM/osrm-backend/pull/2687

PR #2664, opened 2016-07-15
Features/guidance/advanced lanes part 2

MoKob:
Continued work on https://github.com/Project-OSRM/osrm-backend/issues/2553, case descriptions follow on monday


MoKob:
Outdated. In general I'd still think we could improve the turn function by considering many of the tested parameters.
Especially if you are crossing through oncoming traffic would be nice to have (but also depends on traffic light settings).

PR #2667, opened 2016-07-18
Fixes Undefined Behavior in LaneTupel (Strict Aliasing)

daniel-j-h:
Here's the best explanation for the strict aliasing rules I could find. Please read it.
http://dbp-consulting.com/tutorials/StrictAliasing.html

tl;dr: has to go through a memcpy (in C++) as in:
https://github.com/WebAssembly/binaryen/blob/184cc11cee2a65d30c7696eb3284e132099e4acb/src/support/utilities.h#L29-L40

This changeset just compares attributes memberwise.


TheMarex:
Interesting never heard of this before. Looks good to merge to me.

Issue #2269, opened 2016-04-15
Support Left-Sided Driving


MoKob:
This requires OSRM to be country aware. 

At the moment, we only support right-handed driving per default and assume it to be the general mode of driving.
This can result in broken instruction handling in left-handed driving countries.

When we get some sort of country-awareness into OSRM, we need to have a look at guidance and make sure to address left-handed driving.

danpat:
@MoKob Can you list some specific examples?

MoKob:
https://github.com/Project-OSRM/osrm-backend/blob/develop/src/extractor/guidance/roundabout_handler.cpp#L74 would be such a case.

MoKob:
In general, when I noticed it during coding, I labeled the locations with `//FIXME` items

willwhite:
In the case of the roundabout handler above, what would be broken? For example, would the guidance just say "turn right out of the traffic circle" instead of "turn left out of the traffic circle"?

MoKob:
@willwhite the count for intersections is wrong in some cases. it may say take the first instead of the second. The angles themselves are independent of the side of driving.

emiltin:
related: #333 #883

miccolis:
@MoKob any sense of what follow up work will remain after https://github.com/Project-OSRM/osrm-backend/pull/4415 merges?

MoKob:
I am not aware of many situations that actually change in left-sided conditions. We already have some roundabout-handling based on profile flags. This would need to be implemented on a `byCountryFlag` basis, rather than on global level. The main handling works, we just need to ensure the behaviour is adjusted based on countries, so we can have a single graph with both left/right handed driving.

Turn lane handling for roundabouts assumes right-sided driving. We currently have them disabled anyhow, though. When we redo them to work at all, we should take left-sided driving into account.
Other cases might be optimised for right-hand detection. The code requires a general overhaul though, which could be best done, after we have maneuvers in extraction, since lane handling on OSRMs data model is a game of chance right now.

With this flag, we would also be able to switch to a `TurnType::UTurn` and `DirectionModifier::Left\Right`. The main issue here is, that this would constitute a breaking API change, since it cannot be interpreted as a normal turn. A former uturn might suddenly be announced as a `Turn Left`.

daniel-j-h:
Lane Anticipation:

[...]

PR #2670, opened 2016-07-18
[not ready] Left hand driving 

lbud:
Starting toward https://github.com/Project-OSRM/osrm-backend/issues/2269. Needs more tests 🚑 
cc @oxidase

### DIFF
--- a/features/guidance/anticipate-lanes.feature
+++ b/features/guidance/anticipate-lanes.feature
@@ -343,6 +343,38 @@ Feature: Turn Lane Guidance
             | waypoints | route    | turns                           | lanes                                  |
             | a,h       | ab,gh,gh | depart,roundabout-exit-5,arrive | ,slight right:false slight right:true, |
 
+    @anticipate @todo
+    Scenario: Anticipate with lanes in roundabout where we stay on the roundabout for multiple exits
+        # Scenario requires correct lane handling for clockwise roundabouts
+        Given the profile "lhs"
+        And the node map
+            |   |   | a |   |   |
+            |   |   | b |   |   |
+            | h | c |   | g |   |
+            |   |   |   |   |   |
+            |   | d |   | f |   |
+            |   |   | e |   |   |
+            | x |   |   |   | y |
+
+        And the ways
+            | nodes | turn:lanes:forward         | highway | junction   |
+            | ab    | slight_left\|slight_left   | primary |            |
+            | bg    |                            | primary | roundabout |
+            | gf    |                            | primary | roundabout |
+            | fe    |                            | primary | roundabout |
+            | ed    |                            | primary | roundabout |
+            | dc    | slight_left                | primary | roundabout |
+            | cb    |                            | primary | roundabout |
+            | ch    |                            | primary |            |
+            | ex    |                            | primary |            |
+            | dx    |                            | primary |            |
+            | gy    |                            | primary |            |
+            | fy    |                            | primary |            |
+
+        When I route I should get
+            | waypoints | route      | turns                           | lanes                                  |
+            | a,h       | ab,ch,ch   | depart,roundabout-exit-5,arrive | ,slight right:false slight right:true, |
+
     @anticipate
     Scenario: Departing or arriving inside a roundabout does not yet anticipate lanes
         Given the node map
--- a/include/engine/datafacade/datafacade_base.hpp
+++ b/include/engine/datafacade/datafacade_base.hpp
@@ -162,6 +162,8 @@ class BaseDataFacade
 
     virtual bool GetContinueStraightDefault() const = 0;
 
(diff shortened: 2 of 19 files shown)


include/engine/datafacade/datafacade_base.hpp
```diff
@@ -162,6 +162,8 @@ class BaseDataFacade
 
     virtual bool GetContinueStraightDefault() const = 0;
 
+    virtual bool UseLeftSideDriving() const = 0;
```
daniel-j-h: `MockDataFacade` needs a dummy impl. for this, too. Otherwise unit tests won't compile.

daniel-j-h: (try with `make tests`)

src/engine/guidance/post_processing.cpp
```diff
@@ -520,9 +520,18 @@ std::vector<RouteStep> anticipateLaneChangeForRoundabouts(std::vector<RouteStep>
         // FIXME: assumes right-side driving (counter-clockwise roundabout flow)
         const auto enter_direction = enter.maneuver.instruction.direction_modifier;
 
-        if (util::guidance::isRightTurn(enter.maneuver.instruction))
-            enter.maneuver.instruction.direction_modifier =
-                mirrorDirectionModifier(enter_direction);
+        if (facade.UseLeftSideDriving())
```
daniel-j-h: remove FIXME comment above about right-sided assumptions

daniel-j-h: (also below in this file)

src/engine/guidance/post_processing.cpp
```diff
@@ -520,9 +520,18 @@ std::vector<RouteStep> anticipateLaneChangeForRoundabouts(std::vector<RouteStep>
...
+            if (util::guidance::isLeftTurn(enter.maneuver.instruction))
+                enter.maneuver.instruction.direction_modifier =
+                    mirrorDirectionModifier(enter_direction);
+        }
+        else
+        {
+            if (util::guidance::isRightTurn(enter.maneuver.instruction))
+                enter.maneuver.instruction.direction_modifier =
+                    mirrorDirectionModifier(enter_direction);
+        }
```
daniel-j-h: In order to test this it would be great to have clockwise-flowing roundabouts modeled as Cucumber scenarios, you can take this one as example which assumes right-sided driving (and therefore counter-clockwise roundabout flow): https://github.com/Project-OSRM/osrm-backend/blob/master/features/guidance/anticipate-lanes.feature#L317

src/extractor/guidance/roundabout_handler.cpp
```diff
@@ -66,8 +68,15 @@ detail::RoundaboutFlags RoundaboutHandler::getRoundaboutFlags(
     bool on_roundabout = in_edge_data.roundabout;
     bool can_enter_roundabout = false;
     bool can_exit_roundabout_separately = false;
-    for (const auto &road : intersection)
+
+    const bool lhs = profile_properties.left_hand_driving;
+    const int step = lhs ? -1 : 1;
+    const signed int_size = static_cast<signed>(intersection.size());
```
daniel-j-h: Why all the `signed` here and below? Just use `int` for the `-1/1` and `std::size_t` to count things.

features/guidance/anticipate-lanes.feature
```diff
@@ -343,6 +343,38 @@ Feature: Turn Lane Guidance
...
+            | cb    |                            | primary | roundabout |
+            | ch    |                            | primary |            |
+            | ex    |                            | primary |            |
+            | dx    |                            | primary |            |
+            | gy    |                            | primary |            |
+            | fy    |                            | primary |            |
+
+        When I route I should get
+            | waypoints | route      | turns                           | lanes                                  |
+            | a,h       | ab,ch,ch   | depart,roundabout-exit-5,arrive | ,slight right:false slight right:true, |
```
daniel-j-h: I think you forgot to change the expected `lanes` output here from `slight right` to slight left.
If you take a look at the way `ab` and `dc` above, they only have `slight left` lanes.

Counterclockwise vs clockwise does not change the turn lane's direction.
It only changes lane anticipation's behavior in keeping the user to the left/right.

oxidase: yes, here must be `slight left`, but the test fails because query returns an empty result, so i set `todo` state.

I will try to find out the reason during the next week. @MoKob showed me the starting place https://github.com/Project-OSRM/osrm-backend/blob/master/src/extractor/guidance/turn_lane_matcher.cpp#L54-L103

MoKob:
Given #2707 is merged and marked as a duplicate of this, any reason to keep this one around?

TheMarex:
@MoKob nope forgot to close this one. :+1:

PR #2671, opened 2016-07-18
Improves lane handling for subsequent going straight, resolves #2625

### LINKED ISSUE #2625: Handle subsequent lane changes going straight
At the moment we handle subsequent lane changes only if they happen in a [left / right turn](https://github.com/Project-OSRM/osrm-backend/blob/0e4061f85816b5499896022eb151a1402ae3aad0/src/engine/guidance/lane_processing.cpp#L83).
https://github.com/Project-OSRM/osrm-backend/pull/2612 adds roundabouts to the mix.

We should handle going straight as well.

daniel-j-h:
(Re-opening against advanced-lanes branch, as the collapse branch I based it off before got already merged into advanced-lanes; @MoKob go for the merge!)

Before we only worked on subsequent quick turns, as in:
`right, right` keeps the user on the rightmost lanes.

This changeset modifies the logic to work on any subsequent steps
that are "quick" and have lane information we can constrain later.

Because we do not have a from-lane => to-lanes mapping we take the
lanes left and right of the turn lanes into account when heuristically
assigning the leftmost / rightmost lanes.

There are some edge cases where this still does not give us the optimal
solution but it gets close to what is actually possible at the moment
without having a lane mapping in post-processing.

References:
- https://github.com/Project-OSRM/osrm-backend/issues/2625

### DIFF
--- a/features/guidance/anticipate-lanes.feature
+++ b/features/guidance/anticipate-lanes.feature
@@ -257,7 +257,170 @@ Feature: Turn Lane Guidance
 
        When I route I should get
             | waypoints | route                 | turns                                                   | lanes                                                                                                                                                  |
-            | a,f       | abx,bcy,cdz,dew,ef,ef | depart,turn right,turn left,turn right,turn left,arrive | ,straight:false right:false right:true right:false,left:false left:true straight:false,straight:false right:true right:false,left:true straight:false, |
+            | a,f       | abx,bcy,cdz,dew,ef,ef | depart,turn right,turn left,turn right,turn left,arrive | ,straight:false right:true right:false right:false,left:true left:false straight:false,straight:false right:true right:false,left:true straight:false, |
+
+       @anticipate
+       Scenario: Anticipate Lanes for through, through with lanes
+           Given the node map
+               |   |   |   | f | g |   |
+               |   |   |   |   |   |   |
+               | a | b | c | d |   | e |
+               |   |   |   |   |   |   |
+               |   |   |   | h | i |   |
+
+           And the ways
+               | nodes | turn:lanes:forward                     | name |
+               | ab    |                                        | main |
+               | bc    | left\|through\|through\|through\|right | main |
+               | cd    | left\|through\|right                   | main |
+               | de    |                                        | main |
+               | cf    |                                        | off  |
+               | ch    |                                        | off  |
+               | dg    |                                        | off  |
+               | di    |                                        | off  |
+
+          When I route I should get
+               | waypoints | route               | turns                                             | lanes                                                                                                     |
+               | a,e       | main,main,main,main | depart,use lane straight,use lane straight,arrive | ,left:false straight:false straight:true straight:false right:false,left:false straight:true right:false, |
+
+       @anticipate
+       Scenario: Anticipate Lanes for through and collapse multiple use lanes
+           Given the node map
+               |   |   | e | f | g |
+               |   |   |   |   |   |
+               | a | b | c | d |   |
+               |   |   |   |   |   |
+               |   |   | h | i | j |
+
+           And the ways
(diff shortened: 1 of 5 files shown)


features/guidance/anticipate-lanes.feature
```diff
@@ -257,7 +257,116 @@ Feature: Turn Lane Guidance
...
+               | bh    |                                        | off  |
+               | cf    |                                        | off  |
+               | ci    |                                        | off  |
+               | dg    |                                        | off  |
+               | dj    |                                        | off  |
+
+          When I route I should get
+               | waypoints | route          | turns                           | lanes                                                                |
+               | a,c       | main,main,main | depart,use lane straight,arrive | ,left:false straight:true straight:true right:false,                 |
+               | a,d       | main,main,main | depart,use lane straight,arrive | ,left:false straight:true straight:true right:false,                 |
```
MoKob: This one is already collapsed by you, is it?

MoKob:
@daniel-j-h I am seeing no tests for the combination of going straight and turns here. These should work, or not? If so, could you add some test-cases to show what behaviour is working?

In addition, since the builds are working currently. The PR seems to break Windows builds. Potential easy fix?

src/engine/guidance/lane_processing.cpp
```diff
@@ -74,34 +75,110 @@ std::vector<RouteStep> anticipateLaneChange(std::vector<RouteStep> steps,
             if (!lanes_to_constrain || !lanes_fan_in)
                 return;
 
-            // In case there is no lane information we work with one artificial lane
-            const auto current_adjusted_lanes = std::max(current_lanes.lanes_in_turn, LaneID{1});
+            // We do not have a mapping from lanes to lanes. All we have is the lanes in the turn
+            // and all the lanes at that situation. Make the best out of it.
```
MoKob: could you add some explanation what would be required to improve this and how an ideal solution would look like?

MoKob:
Apart from the few items mentioned it looks good to me.

features/guidance/anticipate-lanes.feature
daniel-j-h: Yes, please have a look [here](https://github.com/Project-OSRM/osrm-backend/blob/56dad511b4148a53ff7d0be077ef715c8145105a/src/engine/guidance/lane_processing.cpp#L166-L173), the only place where anticipation now potentially collapses steps. For this I had to adapt the [collapse function](https://github.com/Project-OSRM/osrm-backend/blob/56dad511b4148a53ff7d0be077ef715c8145105a/src/engine/guidance/post_processing.cpp#L507-L525) taking lanes into account, so we only collapse if the lane data is the same.

daniel-j-h:
> @daniel-j-h I am seeing no tests for the combination of going straight and turns here. These should work, or not? If so, could you add some test-cases to show what behaviour is working?

https://github.com/Project-OSRM/osrm-backend/blob/56dad511b4148a53ff7d0be077ef715c8145105a/features/guidance/anticipate-lanes.feature#L312-L369

MoKob:
To clarify: I am thinking about left-straight-right combinations, not going straight with just one left/right at the end.

In addition, most test-cases seem to be symmetric? This could be dangerous with respect to using the lane data and mixing up left/right order. Can we add asymmetric cases?

daniel-j-h:
> To clarify: I am thinking about left-straight-right combinations, not going straight with just one left/right at the end.

Take a look at the scenario I linked: https://github.com/Project-OSRM/osrm-backend/blob/56dad511b4148a53ff7d0be077ef715c8145105a/features/guidance/anticipate-lanes.feature#L338-L369 This has all turn/through/turn combinations covered: `left,through,left`, `left,through,right`, `right,through,right`, `right,through,left`. Am I missing something?

Will add some asymmetric lane tests :+1:

Issue #2672, opened 2016-07-19
get exception when start osrm-routed map.osrm


wxlee:
Hi,
I received a exception when I use OSRM v5.2.6 with Taiwan map.
The map file download from
http://download.geofabrik.de/asia/taiwan-latest.osm.pbf
and I follow the Quick start on https://github.com/Project-OSRM/osrm-backend.
When I use the command 'osrm-route map.osrm', it will run and stop with error
'[warn] [exception] Reading from /data/Taiwan.osrm.icd failed.'
But when I use other area map, it is ok.
If I use the OSRM v4.9.1 with this map, it is also work.
I do google lots posts on Internet, but I can not find any idea about it.
Could you help me?
or any idea about this exception?

Thanks

daniel-j-h:
Can confirm, but on the `v5.2.6` tag as well as on current master. Seems like the `.icd` file is getting generated and is 7.4 MB.

@MoKob can you take a look. This is a blocker for 5.3.

danpat:
Confirmed on OSX `master`.

danpat:
Backtrace for the exception is:

```
(lldb) bt
* thread #1: tid = 0x13253b2, 0x00007fff92620f06 libsystem_kernel.dylib`__pthread_kill + 10, queue = 'com.apple.main-thread', stop reason = signal SIGABRT
  * frame #0: 0x00007fff92620f06 libsystem_kernel.dylib`__pthread_kill + 10
    frame #1: 0x00007fff96ab44ec libsystem_pthread.dylib`pthread_kill + 90
    frame #2: 0x00007fff949f26e7 libsystem_c.dylib`abort + 129
    frame #3: 0x00007fff856daf81 libc++abi.dylib`abort_message + 257
    frame #4: 0x00007fff85700a2f libc++abi.dylib`default_terminate_handler() + 243
    frame #5: 0x00007fff859466c3 libobjc.A.dylib`_objc_terminate() + 124
    frame #6: 0x00007fff856fe19e libc++abi.dylib`std::__terminate(void (*)()) + 8
    frame #7: 0x00007fff856fdc12 libc++abi.dylib`__cxa_throw + 121
    frame #8: 0x00000001002824f9 osrm-routed`osrm::engine::datafacade::InternalDataFacade::LoadIntersectionClasses(this=0x0000000101d00a70, intersection_class_file=0x00007fff5fbfead0) + 4729 at internal_datafacade.hpp:368
[...]

Issue #1682, opened 2015-09-13
`std::basic_istream::read` may fail, we do not check for this


daniel-j-h:
All our reads via `stream.read(...)` are unchecked. That is, in case of extraction failures or premature eof, we silently go on with uninitialized data structures (hint: undefined behavior anyone?).

Two solutions are possible:

- check the read function's return value (stream&) via its `operator bool` conversion
- let streams throw on error (not on per default, to mock with developers)

```c++
int main() try {
  std::ifstream in{"bin"};
  assert(in);
  // in.exceptions(std::ifstream::failbit | std::ifstream::badbit | std::ifstream::eofbit);
  // --^

  std::vector<char> v(10);
  auto& rc = in.read(v.data(), 10);

  if(!rc) {
  // --^
    if(rc.eof()) 
      return 10;
    if(rc.bad()) 
      return 11;
    if(rc.fail()) 
      return 12;
  }

} catch(const std::exception& e) {
  std::cerr << e.what() << std::endl;
}
```

Discussion: I think we can automate this to a certain degree with a clever substitution to turn

    x.read(y,z);

into

    ensure(x.read(y,z));

relying on the returned stream's conversion to bool in a boolean context, with `ensure` being a assert-like function that also triggers in release builds and potentially throws.

References:

- http://en.cppreference.com/w/cpp/io/basic_istream/read
- http://en.cppreference.com/w/cpp/io/ios_base/iostate

daniel-j-h:
Updated ticket, same applies for writing. As of now, this is the state we're in:

```
$ ag '\.read\((.*),(.*)\);' |wc -l
62
$ ag '\.write\((.*),(.*)\);' |wc -l
50
```

danpat:
We're partway there with https://github.com/Project-OSRM/osrm-backend/pull/3383 and https://github.com/Project-OSRM/osrm-backend/pull/3321

```
$ ag '\.read\((.*),(.*)\);' |wc -l
       5
$ ag '\.write\((.*),(.*)\);'  | wc -l
      51
```

PR #2673, opened 2016-07-19
[ready] fix 2672

MoKob:
Even though the usage is the same for in/output, mixing streams and read/write seems to have bad interactions. Works for the overloaded table operator but apparently it messes up parsing the number of bearings for Taiwan.

This PR fixes the issue https://github.com/Project-OSRM/osrm-backend/issues/2672

### DIFF
--- a/src/extractor/extractor.cpp
+++ b/src/extractor/extractor.cpp
@@ -507,8 +507,8 @@ Extractor::BuildEdgeExpandedGraph(lua_State *lua_state,
 
     std::vector<std::uint32_t> turn_lane_offsets;
     std::vector<guidance::TurnLaneType::Mask> turn_lane_masks;
-    if( !util::deserializeAdjacencyArray(
-        config.turn_lane_descriptions_file_name, turn_lane_offsets, turn_lane_masks) )
+    if (!util::deserializeAdjacencyArray(
+            config.turn_lane_descriptions_file_name, turn_lane_offsets, turn_lane_masks))
     {
         util::SimpleLogger().Write(logWARNING) << "Reading Turn Lane Masks failed.";
     }
@@ -673,7 +673,7 @@ void Extractor::WriteIntersectionClassificationData(
     util::RangeTable<> bearing_class_range_table(bearing_counts);
     file_out_stream << bearing_class_range_table;
 
-    file_out_stream << total_bearings;
+    file_out_stream.write(reinterpret_cast<const char *>(&total_bearings), sizeof(total_bearings));
     for (const auto &bearing_class : bearing_classes)
     {
         const auto &bearings = bearing_class.getAvailableBearings();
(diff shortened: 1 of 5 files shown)


src/extractor/extractor.cpp
```diff
@@ -673,25 +673,24 @@ void Extractor::WriteIntersectionClassificationData(
...
-    file_out_stream << total_bearings;
+    file_out_stream.write( reinterpret_cast<const char*>(&total_bearings), sizeof(total_bearings));
     for (const auto &bearing_class : bearing_classes)
     {
         const auto &bearings = bearing_class.getAvailableBearings();
         file_out_stream.write(reinterpret_cast<const char *>(&bearings[0]),
                               sizeof(bearings[0]) * bearings.size());
     }
 
-    // FIXME
```
MoKob: @daniel-j-h You said something about removing this here. Can you elaborate what is required to fix this, is actin required here?

TheMarex:
Could this be the source of data corruption in some cases? Was this problem present in `5.2`?

MoKob:
Since it breaks the reading (exception), it should not start up in that cases. See the exception reported. Was present in 5.2, though.

src/extractor/extractor.cpp
daniel-j-h:
```
      // not implemented in stdlibc++ <= 6.0.20 (<= gcc 4.9)
#if not defined __GNUC__ or __GNUC__ > 4
      static_assert(std::is_trivially_copyable<T>::value, "bytewise writing requires trivially copyable type");
#endif
```

MoKob:
@TheMarex @daniel-j-h ok to merge this? Requires new data processing, though (cc @bsudekum)

src/extractor/extractor.cpp
```diff
@@ -673,25 +673,24 @@ void Extractor::WriteIntersectionClassificationData(
...
-    // This should be here, but g++4.8 does not have it...
-    // static_assert(std::is_trivially_copyable<util::guidance::EntryClass>::value,
-    //              "EntryClass Serialization requires trivial copyable entry classes");
+    if( !static_cast<bool>(file_out_stream) )
+        util::SimpleLogger().Write() << "Failed to write bearing class data.";
```
danpat: Is this a fatal error?  If it is, we should throw an exception so everything comes crashing down and we return a bad exit code.

src/extractor/extractor.cpp
```diff
@@ -673,25 +673,26 @@ void Extractor::WriteIntersectionClassificationData(
...
         const auto &bearings = bearing_class.getAvailableBearings();
         file_out_stream.write(reinterpret_cast<const char *>(&bearings[0]),
                               sizeof(bearings[0]) * bearings.size());
     }
 
-    // FIXME
-    // This should be here, but g++4.8 does not have it...
-    // static_assert(std::is_trivially_copyable<util::guidance::EntryClass>::value,
-    //              "EntryClass Serialization requires trivial copyable entry classes");
+    if( !static_cast<bool>(file_out_stream) )
```
daniel-j-h: clang-format

daniel-j-h:
:+1: this needs to go in as fast as possible!

src/extractor/extractor.cpp
```diff
@@ -673,25 +673,26 @@ void Extractor::WriteIntersectionClassificationData(
...
     }
 
-    // FIXME
-    // This should be here, but g++4.8 does not have it...
-    // static_assert(std::is_trivially_copyable<util::guidance::EntryClass>::value,
-    //              "EntryClass Serialization requires trivial copyable entry classes");
+    if (!static_cast<bool>(file_out_stream))
+    {
+        throw util::exception("Failed to write to " + output_file_name + ".");
+    }
```
danpat: Additionally, should this check happen *before* we try to write to `file_out_stream` ?  I assume `.write`won't crash if `file_out_stream == false`, so it's not terrible, but we could probably fail earlier.

daniel-j-h: Yeah all I/O should be checked. We're doing nothing against errors at the moment => https://github.com/Project-OSRM/osrm-backend/issues/1682

PR #2674, opened 2016-07-19
[ready] Remove generic std::exception handlers

danpat:
When we hit an exceptional circumstance in OSRM, we currently handle everything, print a simple warning message and `exit 1`.

This can make debugging a bit annoying.  Core dumps (if enabled) aren't triggered for many situations, and it takes extra steps to catch exceptions during debugging.

This change removes the generic `std::exception` handlers and just lets tools die a natural death.

### DIFF
--- a/example/example.cpp
+++ b/example/example.cpp
@@ -18,7 +18,7 @@
 
 #include <cstdlib>
 
-int main(int argc, const char *argv[]) try
+int main(int argc, const char *argv[])
 {
     if (argc < 2)
     {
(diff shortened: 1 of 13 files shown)


TheMarex:
@danpat This doesn't compile. Seems like a missing `}` somewhere.

daniel-j-h:
Some Cucumber tests are failing now since on a non-existing command line option an exception is thrown. But never caught, crashing the program hard.

TheMarex:
Merge once the tests pass.

daniel-j-h:
datastore still hard crashes on an unrecognised option

PR #2677, opened 2016-07-20
[ready] Include datasources in route annotation.

danpat:
This PR adds datasource information to the route annotation.
When using multiple traffic sources, this allows you to see the makeup of discovered route.

### DIFF
--- a/include/engine/guidance/assemble_geometry.hpp
+++ b/include/engine/guidance/assemble_geometry.hpp
@@ -50,6 +50,10 @@ inline LegGeometry assembleGeometry(const datafacade::BaseDataFacade &facade,
     geometry.osm_node_ids.push_back(facade.GetOSMNodeIDOfNode(
         reverse_geometry[reverse_geometry.size() - source_node.fwd_segment_position - 1]));
 
+    std::vector<uint8_t> forward_datasource_vector;
+    facade.GetUncompressedDatasources(source_node.forward_packed_geometry_id, forward_datasource_vector);
+
+
     auto cumulative_distance = 0.;
     auto current_distance = 0.;
     auto prev_coordinate = geometry.locations.front();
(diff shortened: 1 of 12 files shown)


include/engine/guidance/assemble_geometry.hpp
```diff
@@ -79,8 +83,14 @@ inline LegGeometry assembleGeometry(const datafacade::BaseDataFacade &facade,
...
     geometry.segment_distances.push_back(cumulative_distance);
+
+    std::vector<DatasourceID> forward_datasources;
+    facade.GetUncompressedDatasources(target_node.forward_packed_geometry_id, forward_datasources);
+
     geometry.annotations.emplace_back(
-        LegGeometry::Annotation{current_distance, target_node.forward_weight / 10.});
+        LegGeometry::Annotation{current_distance,
+                                target_node.forward_weight / 10.,
+                                forward_datasources[target_node.fwd_segment_position]});
```
daniel-j-h: Hm maybe this is unrelated, but here and above the `.emplace_back(T{arg0, arg1, ..})` calls should probably be `.push_back(T{arg0, arg1, ...})` calls --- `push_back` has an rvalue overload.

scripts/format.sh
```diff
@@ -7,8 +7,33 @@
...
+    V=$(clang-format --version)
+    if [[ $V != *3.8* ]] ; then
+        echo "clang-format is not 3.8 (returned ${V})"
+        exit 1
+    fi
+else
+    echo "No appropriate clang-format found (expected clang-format-3.8, or clang-format)"
+    exit 1
+fi
+
```
daniel-j-h: :+1:

TheMarex:
Missing here:
- [ ] Changelog entry for extended API
- [ ] Docs about annotation should be extended a little bit. I don't see node-ids mentioned there

daniel-j-h:
i686 build still breaks https://github.com/Project-OSRM/osrm-backend/blob/fd6bb5ec1f48e341218e64d015927b5b9288f94c/features/testbot/matching.feature#L124

PR #2678, opened 2016-07-21
[ready] additional turn lane scenarios

MoKob:
While scanning the current map on the demo-server, I came across a few situations that should be checked in test-cases to prevent possible regressions and other cases that need to be added at first.

This PR is growing to be a selection of possible test cases to add to OSRM. See cucumber features for details.

### DIFF
--- a/features/guidance/collapse-detail.feature
+++ b/features/guidance/collapse-detail.feature
@@ -1,0 +1,53 @@
+@routing  @guidance @collapsing
+Feature: Collapse
+
+    Background:
+        Given the profile "car"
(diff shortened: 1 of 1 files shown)


TheMarex:
What is the current status here?

MoKob:
The status is WIP

MoKob:
If you simply want to reduce the number of additional PRs, you can merge this. It was simply a branch to let me add specific cases that I want to have in the testsuite. The current ones are working, even though I suspected that they might end up not-working.

PR #2684, opened 2016-07-21
Increase travis cacher timeout

springmeyer:
Looking through logs of recent travis builds I've seen that:

  - Often caches fail to upload or download before timing out
  - This leads to very slow builds that do not benefit from the caching

This fixes the problem by increasing the timeout of the cacher to be able to handle the large objects produced by `ccache`.


TheMarex:
Looks good to me. Merging.

PR #2685, opened 2016-07-21
fix use-lane handling

MoKob:
Collapsing of use-lane instructions is broken right now. This PR corrects the segfaulting behaviour.

### DIFF
--- a/src/engine/guidance/post_processing.cpp
+++ b/src/engine/guidance/post_processing.cpp
@@ -1156,6 +1156,16 @@ std::vector<RouteStep> collapseUseLane(std::vector<RouteStep> steps)
         return (mask & tag) != extractor::guidance::TurnLaneType::empty;
     };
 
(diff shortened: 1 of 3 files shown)


src/engine/guidance/post_processing.cpp
```diff
@@ -1183,7 +1193,9 @@ std::vector<RouteStep> collapseUseLane(std::vector<RouteStep> steps)
         if (step.maneuver.instruction.type == TurnType::UseLane &&
             canCollapeUseLane(step.maneuver.lanes, step.maneuver.lane_description))
         {
-            elongate(steps[step_index - 1], steps[step_index]);
+            const auto previous = getPreviousIndex(step_index);
+            steps[previous] = elongate(steps[previous], steps[step_index]);
```
daniel-j-h: Argh why is there no warning for us ignoring the return value before? There is

    __attribute__((warn_unused_result))

but man this should be on by default..

This needs to go in as soon as possible.

PR #2687, opened 2016-07-21
V8 based profiles

kkaefer:
Adds the ability to use JavaScript with the V8 engine for profiles. Unlike https://github.com/Project-OSRM/osrm-backend/pull/2662, using pure V8 means that it can be run in parallel.

There are a few caveats:
- Callbacks must run synchronously; there's currently no way to asynchronously process a way, node or turn penalty. This means you can't use any sort of async I/O functions, such as file operations and in particular network connections. There is also no way to use `require` currently. I imlemented a bare bones `console.log/warn`, but it doesn't do the inspection-like things node.js is doing.
- JavaScript uses `camelCase` names for result properties instead of the `under_score` naming style; make sure you change the names when porting Lua profiles
- We need to add V8 as a build dependency. Currently, it uses the system-installed version of V8, but I will masonize this.

There are also a few upsides:
- The `car.js` profile I ported takes only around 5 seconds on my example OSM extract, whereas the Lua version takes 9 seconds. These are the overall `time` stats:
  
  | version | real | user | sys |
  | --- | --- | --- | --- |
  | LUA | 20.078s | 43.950s | 7.832s |
  | JS 8 threads | 15.710s | 33.244s | 3.236s |
  | JS 1 thread **(!)** | 22.149s | 18.035s | 1.376s |
  
  So JS is executing faster both in wallclock time and CPU time.
- Not supporting `require` may not be an issue; we can use preprocessing scripts that merge multi-file profiles into one (e.g. something along the lines of browserify)

There are also a few things missing:
- [x] Masonize v8
- [x] Support for segments
- [ ] Clean up implementation, maybe use https://github.com/pmed/v8pp
- [ ] Implement access to osmium Nodes when processing a way
- [ ] Port test lua files to JS and add mode for running tests with JS
- [ ] Harden use of v8, e.g. check that we're not assuming certain return types or parameters

cc @daniel-j-h @TheMarex


emiltin:
You mentioned that it's not possible to use asynchronous network connections.
We're currently using PostGIS queries from the profiles to support "green" bike profiles that prioritise parks, forests, etc. This would still be possible, righht?

kkaefer:
@emiltin Lua support is not going away at least until there's a away to do I/O (like PostGIS queries) from JS.

TheMarex:
@kkaefer You said you started with the masonization of osrm-backend?

Can you ticket the things that are blocking there? I would like to handle that in an own PR.

kkaefer:
@TheMarex no, I didn't, but there's https://github.com/Project-OSRM/osrm-backend/tree/masonize by @springmeyer

springmeyer:
Next actions on this:

  - We are going to try building this PR against v8 3.x (since 3.14 is the version available on latest ubuntu LTS: http://packages.ubuntu.com/xenial/libv8-dev)
  - We are planning on landing #3119 first, to enable v8 coming from mason.

springmeyer:
> We are planning on landing #3119 first, to enable v8 coming from mason.

Landed.

> We are going to try building this PR against v8 3.x (since 3.14 is the version available on latest ubuntu LTS: http://packages.ubuntu.com/xenial/libv8-dev)

Does not build. Looks like our v8 code is using a number of things only available in v8 5.x:

```
/Users/dane/projects/osrm-backend/include/util/v8_util.hpp:4:10: fatal error: 'libplatform/libplatform.h' file not found
#include <libplatform/libplatform.h>

/Users/dane/projects/osrm-backend/include/util/v8_util.hpp:15:32: error: no member named 'ArrayBuffer' in namespace 'v8'
class V8Allocator : public v8::ArrayBuffer::Allocator

Users/dane/projects/osrm-backend/include/util/v8_util.hpp:59:31: error: no type named 'FunctionCallback' in namespace 'v8'; did you mean 'InvocationCallback'?
    new_function_template(v8::FunctionCallback callback = 0,

In file included from /Users/dane/projects/osrm-backend/src/extractor/scripting_environment_v8.cpp:14:
/Users/dane/projects/osrm-backend/include/util/v8_util.hpp:31:47: error: no member named 'GetCurrentContext' in 'v8::Isolate'
    v8::Local<v8::Context> context = isolate->GetCurrentContext();

```

springmeyer:
v8 3.14.5.10 was fairly easy to build. But v8 5.1.281.47 (the version this branch targets) is proving very complicated to build. I'm going to punt on this for now and pick back up next week.

springmeyer:
This branch is now building on travis. So I've checked the `masonize v8` box above in the description. Next actions remain as outlined by @kkaefer: https://github.com/Project-OSRM/osrm-backend/pull/2687#issue-166859543

TheMarex:
I will try to clean this up again a little bit next week and rebase it. Things that are a little bit fuzzy on my mind are from a organizational perspective:

1. How do we migrate the profiles? We need to support both lua and JS for a while because of limitations of the v8 context. Do we also keep two versions of the profiles around?

2. This requires a more recent version of v8 then is shipped with Ubuntu 14.04 (even 16.04) can we find a PPA to at least over a migration path for Ubuntu users?

TheMarex:
Actually this PR does not need any API breaking changes. I think we should push to include this as optional in the `5.x` series.

karenzshea:
Just rebased onto current master.

springmeyer:
Per discussion with @TheMarex - we need to stay on released versions of `mason` based on how our git subtree works. This branch predates that discussion so it includes mason v8 scripts checked directly into `osrm-backend. Instead we need to:

  - Upgrade the mason version after a new version is tagged with https://github.com/mapbox/mason/pull/301
  - Remove the v8 script commits from this branch
  - Upgrade mason
  - Then the v8 script version will be available.

springmeyer:
Ugh, this is messy. I noticed that this branch is actually still missing commits from master even though it was supposed to be rebased. The git subtree merge is needed to update the mason version but that also prevents being able to rebase again. So this is problematic. I think this signals that we should move away from using git subtrees for mason. A git submodule would be much less of a hassle.

TheMarex:
@springmeyer we usually do updates in a separate PR to not run into this problem. My main motivation with not using submodules is that they require a two-step process after checking out the repository.

Grabbing this task.

springmeyer:
> @springmeyer we usually do updates in a separate PR to not run into this problem.

@TheMarex 👍 is this documented somewhere?

> Grabbing this task.

Thanks for fixing!

springmeyer:
Update on this branch status. I reviewed today and fixed a few problems. Now the `mason` builds are passing while the non-mason builds are correctly failing on not finding the v8 dependency. @karenzshea @TheMarex - do we know of a PPA for v8 5.x yet?

TheMarex:
@springmeyer there are a few inofficial PPAs. pinepain has a list of v8 versions, for example https://launchpad.net/~pinepain/+archive/ubuntu/libv8-5.1

I would be fine with using these for travis. Main problem I see is that there are no 32bit builds.

TheMarex:
Rebased and added support for `segmentFunction`. V8 is also now optional and support will be automatically disabled when it is not found.

danpat:
@TheMarex This PR is getting pretty stale.  Unless we have a concrete plan to actually roll this out, I think we should close/archive this PR, instead of just bumping it from milestone to milestone.

We've made some big improvements in the Lua-sensitive parts of the codebase (edge-based-graph generation parallelization, avoiding unnecessary node_function calls), so there is less urgency for the performance benefits of v8.

There's a lot of invested knowledge in the current Lua scripts - Lua as a language is a bit odd, but perfectly capable here.

kkaefer:
stale

PR #2691, opened 2016-07-22
Target C++14 standard

springmeyer:
Testing branch to enable c++14

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -43,14 +43,6 @@ matrix:
       env: CCOMPILER='gcc-5' CXXCOMPILER='g++-5' BUILD_TYPE='Debug' COVERAGE=ON
 
     - os: linux
-      compiler: "gcc-4.8-debug"
-      addons: &gcc48
-        apt:
-          sources: ['ubuntu-toolchain-r-test']
-          packages: ['g++-4.8', 'libbz2-dev', 'libstxxl-dev', 'libstxxl1', 'libxml2-dev', 'libzip-dev', 'lua5.1', 'liblua5.1-0-dev', 'libtbb-dev', 'libgdal-dev', 'libluabind-dev', 'libboost-all-dev', 'ccache']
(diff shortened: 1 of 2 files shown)


springmeyer:
This passed on travis except for the g++-4.8 jobs which fail, curiously, in the luabind check: https://travis-ci.org/Project-OSRM/osrm-backend/jobs/146640025#L1271

Is this worth trying to fix(the luabind check)? In other words, when we move to require or default to c++14 is there any reason to keep g++-4.8 support? /cc @TheMarex @daniel-j-h

daniel-j-h:
gcc 4.8 is probably way too old for C++14 anyway

TheMarex:
Yeah let's drop 4.8. Only reason we had it was for node-osrm.

springmeyer:
Dropped g++-4.8 from testing matrix. This is rebased and ready to merge after 👀 from @TheMarex.

daniel-j-h:
Nice! We have some gcc4.8 workarounds in the code base, would be great to rip them out.

Also what are now our requirements? gcc>=4.9, clang>=3.6?

TheMarex:
> Also what are now our requirements? gcc>=4.9, clang>=3.6?

Yes I think gcc 4.9 will work. Not sure about clang 3.6 - was that ever a stable release?

daniel-j-h:
Sure, comes with Ubuntu 15.04 for example.

.travis.yml
```diff
@@ -76,19 +68,6 @@ matrix:
       compiler: "gcc-5-release-i686"
       env: TARGET_ARCH='i686' CCOMPILER='gcc-5' CXXCOMPILER='g++-5' BUILD_TYPE='Release'
 
-    - os: linux
```
TheMarex: Ticket the ARM upgrade here https://github.com/Project-OSRM/osrm-backend/issues/2699

PR #2693, opened 2016-07-22
[Ready] Remove lanes from roundabouts, closes #2626

### LINKED ISSUE #2626: Handle lanes within a roundabout
https://github.com/Project-OSRM/osrm-backend/pull/2612 implements Lane Anticipation in roundabouts based on a roundabout's `enter` and `leave` instruction. For all other (silent) instructions inside a roundabout we currently do not get any lane information for post processing to work on.

We still want to
- issue `UseLane` instruction inside roundabouts
- do lane anticipation for lane changes within a roundabout

This depends on @MoKob's turn handler work https://github.com/Project-OSRM/osrm-backend/pull/2614.

daniel-j-h:
After half a day of looking at the tagging and the data came to the
following conclusion:

We can't keep the user to the innermost / outermost lanes depending on
the exit the route takes: we found situations where both heuristics were
wrong.

Even on popular roundabouts the tagging is often wrong or in the best
case not present at all.

There are at least two different ways to interpret roundabout
indications: 1/ where e.g. a right arrow on the lane indicates turn
restrictions for the roundabout and the need to take this lane to exit
the roundabout to the right (possibly skipping multiple exits) and 2/
where a right arrow just means this is a lane in a immediate right turn.

Example: Australia marks lanes with arrows that seem to indicate
"angles you can exit the roundabout from", for example, these two ways:
- http://www.openstreetmap.org/way/320941710
- http://www.openstreetmap.org/way/42918021

Whereas Germany marks lanes with "directions you can travel in these
lanes immediately after entering the roundabout":
- http://www.openstreetmap.org/way/52578338

These two different interpretations of how to draw the arrows on the
roads mean we have conflicting solutions to "which lanes can you use to
take exit B from entry A" based on locality.

Continuing to tag ways based on lane markings is no problem, but
unfortunately, we can't reliably resolve good advice for navigation
system users (like "use the inside lane to take the second exit at the
roundabout"), there are too many situations that would generate bad
instructions (instructions that tell users to go into a lane they
shouldn't use).

### DIFF
--- a/src/engine/guidance/lane_processing.cpp
+++ b/src/engine/guidance/lane_processing.cpp
@@ -3,6 +3,7 @@
 #include "util/guidance/toolkit.hpp"
 
 #include "extractor/guidance/turn_instruction.hpp"
+#include "engine/guidance/toolkit.hpp"
 
 #include <iterator>
 
@@ -101,6 +102,29 @@ std::vector<RouteStep> anticipateLaneChange(std::vector<RouteStep> steps,
     };
 
     std::for_each(begin(subsequent_quick_turns), end(subsequent_quick_turns), constrain_lanes);
+    return steps;
+}
+
+std::vector<RouteStep> removeLanesFromRoundabouts(std::vector<RouteStep> steps)
+{
+    using namespace util::guidance;
+
+    const auto removeLanes = [](RouteStep &step) {
+        for (auto &intersection : step.intersections)
+        {
+            intersection.lane_description = {};
+            intersection.lanes = {};
+        }
+    };
+
+    for (auto &step : steps)
+    {
+        const auto inst = step.maneuver.instruction;
+
+        if (entersRoundabout(inst) || staysOnRoundabout(inst) || leavesRoundabout(inst))
+            removeLanes(step);
+    }
+
     return steps;
 }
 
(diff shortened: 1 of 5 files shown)


src/engine/guidance/lane_processing.cpp
```diff
@@ -104,6 +105,29 @@ std::vector<RouteStep> anticipateLaneChange(std::vector<RouteStep> steps,
...
+std::vector<RouteStep> removeLanesFromRoundabouts(std::vector<RouteStep> steps)
+{
+    using namespace util::guidance;
+
+    const auto removeLanes = [](RouteStep &step) {
+        for (auto &intersection : step.intersections)
+        {
+            intersection.lane_description = {};
+            intersection.lanes = {};
+        }
```
daniel-j-h: This removes both 1/ lanes in turn as well as 2/ the lane overview from all intersections affected.

src/engine/guidance/lane_processing.cpp
```diff
@@ -104,6 +105,29 @@ std::vector<RouteStep> anticipateLaneChange(std::vector<RouteStep> steps,
...
+    };
+
+    for (auto &step : steps)
+    {
+        const auto inst = step.maneuver.instruction;
+
+        if (entersRoundabout(inst) || staysOnRoundabout(inst) || leavesRoundabout(inst))
+            removeLanes(step);
```
daniel-j-h: Strictly speaking roundabout instructions are already collapsed to a single multi-hop Enter instruction when we call this function in post-processing. But better to err on the safe side in case we re-structure post-processing later on or re-use this function.

TheMarex:
Looks good to me. :+1:

PR #2696, opened 2016-07-25
Moves invalid id warnings to debug, closes #2630

### LINKED ISSUE #2630: lots of "[warn] Found invalid node reference" in release build
When extracting, i get a lot of ""[warn] Found invalid node reference" filling up the log file. Could it maybe be moved to debug level, so it's not shown in release builds?

daniel-j-h:
For https://github.com/Project-OSRM/osrm-backend/issues/2630#issuecomment-234792301.

No longer spam stdout with invalid id logs when running on a smaller extract cut from the planet.

### DIFF
--- a/src/extractor/extraction_containers.cpp
+++ b/src/extractor/extraction_containers.cpp
@@ -96,7 +96,6 @@ struct CmpEdgeByInternalSourceTargetAndName
     const oe::ExtractionContainers::STXXLNameCharData &name_data;
     const oe::ExtractionContainers::STXXLNameOffsets &name_offsets;
 };
-
 }
 
 namespace osrm
@@ -192,7 +191,8 @@ void ExtractionContainers::WriteCharData(const std::string &file_name)
     // transforms in-place name offsets to name lengths
     BOOST_ASSERT(!name_offsets.empty());
     for (auto curr = name_offsets.begin(), next = name_offsets.begin() + 1;
-         next != name_offsets.end(); ++curr, ++next)
+         next != name_offsets.end();
+         ++curr, ++next)
     {
         *curr = *next - *curr;
     }
@@ -316,8 +316,8 @@ void ExtractionContainers::PrepareEdges(ScriptingEnvironment &scripting_environm
     {
         if (edge_iterator->result.osm_source_id < node_iterator->node_id)
         {
-            util::SimpleLogger().Write(LogLevel::logWARNING) << "Found invalid node reference "
-                                                             << edge_iterator->result.source;
+            util::SimpleLogger().Write(LogLevel::logDEBUG) << "Found invalid node reference "
+                                                           << edge_iterator->result.source;
             edge_iterator->result.source = SPECIAL_NODEID;
             ++edge_iterator;
             continue;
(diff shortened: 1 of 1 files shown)


src/extractor/extraction_containers.cpp
```diff
@@ -323,8 +323,8 @@ void ExtractionContainers::PrepareEdges(ScriptingEnvironment &scripting_environm
     {
         if (edge_iterator->result.osm_source_id < node_iterator->node_id)
         {
-            util::SimpleLogger().Write(LogLevel::logWARNING) << "Found invalid node reference "
-                                                             << edge_iterator->result.source;
+            util::SimpleLogger().Write(LogLevel::logDEBUG) << "Found invalid node reference "
```
TheMarex: Hrm this one is kind of important - if a node id is not found the graph will potentially be corrupted. We had this case once for Highway 1 in CA breaking our routing in California.

So I would not switch it to debug here.

src/extractor/extraction_containers.cpp
```diff
@@ -359,8 +359,8 @@ void ExtractionContainers::PrepareEdges(ScriptingEnvironment &scripting_environm
     // Remove all remaining edges. They are invalid because there are no corresponding nodes for
     // them. This happens when using osmosis with bbox or polygon to extract smaller areas.
     auto markSourcesInvalid = [](InternalExtractorEdge &edge) {
-        util::SimpleLogger().Write(LogLevel::logWARNING) << "Found invalid node reference "
-                                                         << edge.result.source;
+        util::SimpleLogger().Write(LogLevel::logDEBUG) << "Found invalid node reference "
```
TheMarex: Same here.

src/extractor/extraction_containers.cpp
```diff
@@ -394,7 +394,7 @@ void ExtractionContainers::PrepareEdges(ScriptingEnvironment &scripting_environm
 
         if (edge_iterator->result.osm_target_id < node_iterator->node_id)
         {
-            util::SimpleLogger().Write(LogLevel::logWARNING)
+            util::SimpleLogger().Write(LogLevel::logDEBUG)
```
TheMarex: Same here.

src/extractor/extraction_containers.cpp
```diff
@@ -462,8 +462,8 @@ void ExtractionContainers::PrepareEdges(ScriptingEnvironment &scripting_environm
     // Remove all remaining edges. They are invalid because there are no corresponding nodes for
     // them. This happens when using osmosis with bbox or polygon to extract smaller areas.
     auto markTargetsInvalid = [](InternalExtractorEdge &edge) {
-        util::SimpleLogger().Write(LogLevel::logWARNING) << "Found invalid node reference "
-                                                         << edge.result.target;
+        util::SimpleLogger().Write(LogLevel::logDEBUG) << "Found invalid node reference "
```
TheMarex: Same here.

src/extractor/extraction_containers.cpp
```diff
@@ -722,24 +722,25 @@ void ExtractionContainers::PrepareRestrictions()
         if (way_start_and_end_iterator->way_id >
             OSMWayID{static_cast<std::uint32_t>(restrictions_iterator->restriction.from.way)})
         {
-            util::SimpleLogger().Write(LogLevel::logWARNING)
+            util::SimpleLogger().Write(LogLevel::logDEBUG)
```
TheMarex: This is fine: Restrictions usually are not that important (other than turn restrictions.)

daniel-j-h:
Updated to only put invalid ids for restrictions to debug.

PR #2697, opened 2016-07-25
Lane Handling for multiple indications per lane as in left;left|, fixes #2694

### LINKED ISSUE #2694: v5 demo server hits lane assertion during extraction
The v5 demo server is currently hitting this assertion during extraction:

```
in: osrm::extractor::ExtractorCallbacks::ProcessWay(const osmium::Way&, const osrm::extractor::ExtractionWay&)::<lambda(std::string)>: (lane_mask & translated_mask) == 0
[assert] /mnt/ebs/osrm/2016-07-22-173001/src/extractor/extractor_callbacks.cpp:206
```

https://github.com/Project-OSRM/osrm-backend/blob/9e323d2d426dc6ae8ce61a1c487d9a4c25c7f41d/src/extractor/extractor_callbacks.cpp#L206

daniel-j-h:
For https://github.com/Project-OSRM/osrm-backend/issues/2694. cc @TheMarex 

Before we asserted on unique lane indications per lane. Turns out the
OSM data contains lane strings such as:

    left;left|right

Which represents two lanes as in:

    <<     >
     ||    |

The two left indications _on a single lane_ look like data issue.
And we can't represent this with our enum-approach at the moment.

We don't want to crash there, so silently swallow this and
generate a single left|right for it.

### DIFF
--- a/src/extractor/extractor_callbacks.cpp
+++ b/src/extractor/extractor_callbacks.cpp
@@ -193,7 +193,8 @@ void ExtractorCallbacks::ProcessWay(const osmium::Way &input_way, const Extracti
             for (auto token_itr = inner_tokens.begin(); token_itr != inner_tokens.end();
                  ++token_itr)
             {
-                auto position = std::find(osm_lane_strings, osm_lane_strings + num_osm_tags, *token_itr);
+                auto position =
+                    std::find(osm_lane_strings, osm_lane_strings + num_osm_tags, *token_itr);
                 const auto translated_mask =
                     masks_by_osm_string[std::distance(osm_lane_strings, position)];
                 if (translated_mask == TurnLaneType::empty)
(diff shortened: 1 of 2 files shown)


src/extractor/extractor_callbacks.cpp
```diff
@@ -203,7 +204,10 @@ void ExtractorCallbacks::ProcessWay(const osmium::Way &input_way, const Extracti
                                                          << *token_itr << "\"";
                     return {};
                 }
-                BOOST_ASSERT((lane_mask & translated_mask) == 0); // make sure the mask is valid
+
+                // In case of multiple times the same lane indicators withn a lane, as in
+                // "left;left|.."  or-ing the masks generates a single "left" enum.
+                // Which is fine since this is data issue and we can't represent it anyway.
```
daniel-j-h: Sorry for the noise, this is the line changing.

TheMarex:
Looks good to me.

PR #2708, opened 2016-07-27
No longer hard crash datastore on unrecognised options

daniel-j-h:
For https://github.com/Project-OSRM/osrm-backend/pull/2674#issuecomment-235240123. Datastore was missing.


TheMarex:
:+1: looks good to merge.

PR #2710, opened 2016-07-27
Added 'return EXIT_SUCCESS'

kshalnev:
Hello.
Please consider the fix:
added missing return EXIT_SUCCESS.

Thanks.


danpat:
This isn't strictly necessary.

See note (4) at http://en.cppreference.com/w/cpp/language/main_function

> The body of the main function does not need to contain the return statement: if control reaches the end of main without encountering a return statement, the effect is that of executing return 0;.

That said, this change makes the intent clearer.  LGTM.

PR #2715, opened 2016-07-28
Filter out edges that have any speed=0 segments.

### LINKED ISSUE #2713: Overflow in speed updates
@danpat thinks there's an overflow still in the speed updates.  If a speed of 0 is found, we set the whole edge weight to INVALID_EDGE_WEIGHT.  But then we still add the turn penalty.

Overflowing here when `new_weight == INVALID_EDGE_WEIGHT`:

https://github.com/Project-OSRM/osrm-backend/blob/master/src/contractor/contractor.cpp#L823

- write a test case for this and https://github.com/Project-OSRM/osrm-backend/pull/2688
- fix it

danpat:
Fixes #2713

:flameretardantpants: are on, I'm ready for the use-of-`goto` abuse.  I think this is the perfect situation for it.

TODO:
  - [x] don't snap to edges that are zero'd
  - [x] verify contraction on nodes that become isolated because of edge filtering

Notes: 
  - debug tiles will show edges with zero speeds, even though we won't route on them.
  - in a perfect world, setting a single segment to 0 would bisect a road, leading to two cul-de-sac like streets.  Without re-writing/recompressing geometries, we can't do that, so instead we're setting the whole edge to unroutable.  For a long road with a single 0-speed segment in the middle, this makes the whole road effectively unreachable, which is the best we can do for now I think.

### DIFF
--- a/features/car/traffic_speeds.feature
+++ b/features/car/traffic_speeds.feature
@@ -23,8 +23,8 @@ Feature: Traffic - speeds
             | fb    | primary |
         And the speed file
         """
-        1,2,27
-        2,1,27
+        1,2,0
+        2,1,0
         2,3,27
         3,2,27
         1,4,27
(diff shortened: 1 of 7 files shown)


src/contractor/contractor.cpp
```diff
@@ -828,6 +833,10 @@ EdgeID Contractor::LoadEdgeExpandedGraph(
         }
 
         edge_based_edge_list.emplace_back(std::move(inbuffer));
+
+        // This is the jump target of the `goto` a few lines back.  It's used when we find a
+        // speed=0 edge and we want to skip adding it to the edge_based_edge_list.
+        skip_this_edge: ;
```
daniel-j-h: Achievement Unlocked :older_man: :sandal:

emiltin:
goto :grinning:

features/car/traffic_speeds.feature
```diff
@@ -23,8 +23,8 @@ Feature: Traffic - speeds
             | fb    | primary |
         And the speed file
         """
-        1,2,27
-        2,1,27
+        1,2,0
+        2,1,0
```
daniel-j-h: While we're at it, could you add negative number test cases, as well.

include/engine/geospatial_query.hpp
```diff
@@ -20,6 +20,11 @@ namespace osrm
 namespace engine
 {
 
+std::pair<bool, bool> boolPairAnd(const std::pair<bool, bool> &A, const std::pair<bool, bool> &B)
```
daniel-j-h: needs to be `inline` otherwise UB due to ODR-violation

features/car/traffic_speeds.feature
danpat: Done!

danpat:
I modified the `benchmark/static_rtree` to use the GeospatialQuery rather than the StaticRTree directly, to see what impact the additional lookups have.  Seems negligible:

3 runs against master:
```
Running raw RTree queries (1 result) with 100000 coordinates: Took 2.4254 seconds (2425.4ms)  ->  0.024254 ms/query (2425.4ms)
Running raw RTree queries (1 result) with 100000 coordinates: Took 2.30604 seconds (2306.04ms)  ->  0.0230604 ms/query (2306.04ms)
Running raw RTree queries (1 result) with 100000 coordinates: Took 2.29847 seconds (2298.47ms)  ->  0.0229847 ms/query (2298.47ms)

Running raw RTree queries (10 results) with 100000 coordinates: Took 5.64783 seconds (5647.83ms)  ->  0.0564783 ms/query (5647.83ms)
Running raw RTree queries (10 results) with 100000 coordinates: Took 5.63754 seconds (5637.54ms)  ->  0.0563754 ms/query (5637.54ms)
Running raw RTree queries (10 results) with 100000 coordinates: Took 5.65011 seconds (5650.11ms)  ->  0.0565011 ms/query (5650.11ms)
```

3 runs against this branch:

```
Running raw RTree queries (1 result) with 100000 coordinates: Took 2.43107 seconds (2431.07ms)  ->  0.0243107 ms/query (2431.07ms)
Running raw RTree queries (1 result) with 100000 coordinates: Took 2.32476 seconds (2324.76ms)  ->  0.0232476 ms/query (2324.76ms)
Running raw RTree queries (1 result) with 100000 coordinates: Took 2.32447 seconds (2324.47ms)  ->  0.0232447 ms/query (2324.47ms)

Running raw RTree queries (10 results) with 100000 coordinates: Took 5.66733 seconds (5667.33ms)  ->  0.0566733 ms/query (5667.33ms)
Running raw RTree queries (10 results) with 100000 coordinates: Took 5.72267 seconds (5722.67ms)  ->  0.0572267 ms/query (5722.67ms)
Running raw RTree queries (10 results) with 100000 coordinates: Took 5.76123 seconds (5761.23ms)  ->  0.0576123 ms/query (5761.23ms)
```

The impact is very small, so I think this should be OK to merge.

src/contractor/contractor.cpp
```diff
@@ -787,9 +787,17 @@ EdgeID Contractor::LoadEdgeExpandedGraph(
...
+                        // We don't want to include it in the edge_based_edge_list, so
+                        // we jump to the end of the parent loop and skip the remaineder
+                        // of the logic.  This is effectively a `continue` for the outer loop.
+                        // This is one of the few good use-cases for `goto` (a break/continue
+                        // on a parent loop), consider carefully the extra complexity involved
+                        // if you decide to remove this.
+                        // Note, we need to increment the penaltyblock here because the goto
+                        // skips the regular incrementing lower down.
+                        penaltyblock++;
+                        goto skip_this_edge;
```
TheMarex: No `goto` calls even if they look elegant. This creates precedence I don't want to see in the code base. should be easy to replace by a boolean and a break.

Issue #2592, opened 2016-06-27
De-duplicate forward/reverse geometries


danpat:
Currently, the `.geometry` file contains lists of node IDs for compressed edges in the node-based graph.  These map directly to coordinates, and are used to return the final coordinates in a route.

However, we're storing the same geometry twice - once for the forward direction, once for the reverse.  This leads to quite a bit of waste.

Currently, each compressed geometry contains the 2nd to N nodes for the edge.  The first node is never needed, because it can be retrieved from the last node of the previous edge if a route traverses multiple edges, and is not required for the first edge, as the first PhantomNode will represent the first coordinate on a section of geometry, never the first node.

However, we're storing the almost the same geometry for the reverse version of the edge - nodes 1 to N-1, in reverse order.  This essentially doubles the size of the `.geometry` table.

We should consolidate these to include all nodes along an edge, and update logic to traverse the list either forwards or reverse.  There will be some duplication of nodes at intersection points, but this should be far less than the doubled geometries we have right now.

TheMarex:
Problem is that we need a way to figure out who to derive both forward and reverse weights from each ID. I would imagine we need to split the EdgeWeight vector into forward and reverse weights, which in turn might again need a little bit extra space.

lbud:
Braindumping my progress on understanding this @TheMarex:

Currently graph compression works such that in this basic case

```
   f1     f2
        ͢
  ____   ____  
u      v      w
  ----   ---- 
       ⃪
   r2     r1 
```
we'd basically add `uv` and then attach `vw` onto it, attach maybe `wx` onto that, etc…
and then in reverse we'd add `wv`, and then attach `vu` onto it, and then maybe `ut`, etc.

BUT in a world where we store only one `EdgeBucket` per edge with both forward and reverse weights, we'll want this scenario to translate to EdgeBuckets as
```
{
    node = v
    fwd_weight = f1
    rev_weight = r1
}
{
    node = w
    fwd_weight = f2
    rev_weight = INVALID_WEIGHT
}
{
    node = u
    rev_weight = r2
    fwd_weight = INVALID_WEIGHT
}
```

That’s all well and good, but then for the next edge in the same compressed segment, say to the right of `vw`:
```
   f1     f2     f3
        ͢
  ____   ____   ____
u      v      w      x
  ----   ----   ----
       ⃪
   r2     r1     r3
```
[...]

Issue #2575, opened 2016-06-22
Refactor compressed geometry


TheMarex:
Currently we save `node id`, `weight` and soon `duration` as interleaved structs in the `.geometry` file. We need to decompose this into three separate vectors, one for each member.

This will enable:
- [ ] DataFaces can just return ranges to the vector: zero-copy for `GetUncompressedGeometry` + `GetUncompressedWeights` + `GetUncompressedDurations`
- [ ] Updating the weight values in osrm-contract will be a lot faster: We only need to write one (or two) vector containing the weights (or durations) the structure is never changed.
- [ ] No dependency on CompressedContainer::CompressedEdge in `engine/`

danpat:
This should get included in https://github.com/Project-OSRM/osrm-backend/issues/2592

TheMarex:
Actually this is a requirement for https://github.com/Project-OSRM/osrm-backend/pull/2399 since it would enable us to conditionally only load the duration to memory if applicable.

jakepruitt:
This was closed with #3093

PR #2717, opened 2016-07-29
Merge compressed geometries

lbud:
This is working towards https://github.com/Project-OSRM/osrm-backend/issues/2592.

At a glance/in many tests it seems to work; however, there are 79 cucumber scenarios failing ([list](https://gist.github.com/lbud/6ee79543e986f02ffd935248a48a94d1)), and a lot of them are just kind of like…dropping steps.

The approach this branch takes is
* change the current `CompressedEdge` to a `OnewayCompressedEdge`; these are still created as an intermediary step but are not serialized to disk
* in `EdgeBasedGraphFactory::InsertEdgeBasedNode`, call `CompressedEdgeContainer::ZipEdges` on each forward/reverse pair, creating a `CompressedEdge` that has both forward and reverse weights

As edges formerly looked like (n=node ID, w=weight, s=start node (not IDed))
```
       w1        w2        w3
     -->       -->       -->
  s    n1        n2        n3
``` 
they now need to look like (n=node ID, f=forward weight, r=reverse weight, I=invalid weight)
```
I        f2        f3        f4
-->      -->       -->       -->
  n1       n2        n3        n4
  <--      <--       <--       <--
    r1       r2        r3        I
``` 

* serialize the newly zipped edges to disk rather than the oneway edges
* modify all `GetUncompressedWeights` `GetUncompressedGeometries` etc to have both forward and reverse functions.

I can't figure out if maybe there's something I'm missing in routing_base.hpp that broke, or what. @TheMarex @danpat @daniel-j-h do any of my changes look suspicious or conspicuously missing anything?

\* I still haven't checked to see if anything about the tile plugin is broken — will do so after everything else works.

### DIFF
--- a/include/engine/datafacade/datafacade_base.hpp
+++ b/include/engine/datafacade/datafacade_base.hpp
@@ -8,6 +8,7 @@
 #include "extractor/external_memory_node.hpp"
 #include "extractor/guidance/turn_instruction.hpp"
 #include "extractor/guidance/turn_lane_types.hpp"
+#include "extractor/original_edge_data.hpp"
 #include "engine/phantom_node.hpp"
 #include "util/exception.hpp"
 #include "util/guidance/bearing_class.hpp"
@@ -74,20 +75,22 @@ class BaseDataFacade
     virtual util::Coordinate GetCoordinateOfNode(const unsigned id) const = 0;
     virtual OSMNodeID GetOSMNodeIDOfNode(const unsigned id) const = 0;
 
-    virtual unsigned GetGeometryIndexForEdgeID(const unsigned id) const = 0;
+    virtual GeometryID GetGeometryIndexForEdgeID(const unsigned id) const = 0;
 
-    virtual void GetUncompressedGeometry(const EdgeID id,
-                                         std::vector<NodeID> &result_nodes) const = 0;
+    virtual std::vector<NodeID> GetUncompressedForwardGeometry(const EdgeID id) const = 0;
+
+    virtual std::vector<NodeID> GetUncompressedReverseGeometry(const EdgeID id) const = 0;
 
     // Gets the weight values for each segment in an uncompressed geometry.
     // Should always be 1 shorter than GetUncompressedGeometry
-    virtual void GetUncompressedWeights(const EdgeID id,
-                                        std::vector<EdgeWeight> &result_weights) const = 0;
+    virtual std::vector<EdgeWeight> GetUncompressedForwardWeights(const EdgeID id) const = 0;
+
+    virtual std::vector<EdgeWeight> GetUncompressedReverseWeights(const EdgeID id) const = 0;
 
     // Returns the data source ids that were used to supply the edge
     // weights.  Will return an empty array when only the base profile is used.
-    virtual void GetUncompressedDatasources(const EdgeID id,
-                                            std::vector<uint8_t> &data_sources) const = 0;
+    virtual std::vector<uint8_t> GetUncompressedForwardDatasources(const EdgeID id) const = 0;
+    virtual std::vector<uint8_t> GetUncompressedReverseDatasources(const EdgeID id) const = 0;
 
     // Gets the name of a datasource
     virtual std::string GetDatasourceName(const uint8_t datasource_name_id) const = 0;
--- a/include/engine/datafacade/internal_datafacade.hpp
+++ b/include/engine/datafacade/internal_datafacade.hpp
@@ -76,7 +76,7 @@ class InternalDataFacade final : public BaseDataFacade
 
     util::ShM<util::Coordinate, false>::vector m_coordinate_list;
     util::PackedVector<OSMNodeID, false> m_osmnodeid_list;
-    util::ShM<NodeID, false>::vector m_via_node_list;
+    util::ShM<GeometryID, false>::vector m_via_geometry_list;
     util::ShM<unsigned, false>::vector m_name_ID_list;
     util::ShM<extractor::guidance::TurnInstruction, false>::vector m_turn_instruction_list;
     util::ShM<LaneDataID, false>::vector m_lane_data_id;
@@ -189,7 +189,7 @@ class InternalDataFacade final : public BaseDataFacade
         boost::filesystem::ifstream edges_input_stream(edges_file, std::ios::binary);
         unsigned number_of_edges = 0;
         edges_input_stream.read((char *)&number_of_edges, sizeof(unsigned));
-        m_via_node_list.resize(number_of_edges);
+        m_via_geometry_list.resize(number_of_edges);
         m_name_ID_list.resize(number_of_edges);
         m_turn_instruction_list.resize(number_of_edges);
         m_lane_data_id.resize(number_of_edges);
@@ -201,7 +201,7 @@ class InternalDataFacade final : public BaseDataFacade
         {
             edges_input_stream.read((char *)&(current_edge_data),
                                     sizeof(extractor::OriginalEdgeData));
-            m_via_node_list[i] = current_edge_data.via_node;
+            m_via_geometry_list[i] = current_edge_data.via_geometry;
             m_name_ID_list[i] = current_edge_data.name_id;
             m_turn_instruction_list[i] = current_edge_data.turn_instruction;
             m_lane_data_id[i] = current_edge_data.lane_data_id;
@@ -670,9 +670,9 @@ class InternalDataFacade final : public BaseDataFacade
         return GetNameForID(name_id + 1);
     }
 
-    virtual unsigned GetGeometryIndexForEdgeID(const unsigned id) const override final
+    virtual GeometryID GetGeometryIndexForEdgeID(const unsigned id) const override final
     {
-        return m_via_node_list.at(id);
+        return m_via_geometry_list.at(id);
     }
 
     virtual std::size_t GetCoreSize() const override final { return m_is_core_node.size(); }
@@ -689,47 +689,128 @@ class InternalDataFacade final : public BaseDataFacade
         }
     }
 
-    virtual void GetUncompressedGeometry(const EdgeID id,
-                                         std::vector<NodeID> &result_nodes) const override final
+    virtual std::vector<NodeID> GetUncompressedForwardGeometry(const EdgeID id) const override final
     {
+        /*
+         * NodeID's for geometries are stored in one place for
+         * both forward and reverse segments along the same bi-
+         * directional edge. The m_geometry_indices stores
+         * refences to where to find the beginning of the bi-
+         * directional edge in the m_geometry_list vector.
+         * */
         const unsigned begin = m_geometry_indices.at(id);
         const unsigned end = m_geometry_indices.at(id + 1);
 
-        result_nodes.clear();
+        std::vector<NodeID> result_nodes;
+
         result_nodes.reserve(end - begin);
+
         std::for_each(m_geometry_list.begin() + begin,
                       m_geometry_list.begin() + end,
                       [&](const osrm::extractor::CompressedEdgeContainer::CompressedEdge &edge) {
                           result_nodes.emplace_back(edge.node_id);
                       });
+
+        return result_nodes;
     }
 
-    virtual void
-    GetUncompressedWeights(const EdgeID id,
-                           std::vector<EdgeWeight> &result_weights) const override final
+    virtual std::vector<NodeID> GetUncompressedReverseGeometry(const EdgeID id) const override final
     {
+        /*
+         * NodeID's for geometries are stored in one place for
+         * both forward and reverse segments along the same bi-
+         * directional edge. The m_geometry_indices stores
+         * refences to where to find the beginning of the bi-
+         * directional edge in the m_geometry_list vector.
+         * */
         const unsigned begin = m_geometry_indices.at(id);
         const unsigned end = m_geometry_indices.at(id + 1);
 
-        result_weights.clear();
+        std::vector<NodeID> result_nodes;
+
+        result_nodes.reserve(end - begin);
+
+        std::for_each(m_geometry_list.rbegin() + (m_geometry_list.size() - end),
+                      m_geometry_list.rbegin() + (m_geometry_list.size() - begin),
+                      [&](const osrm::extractor::CompressedEdgeContainer::CompressedEdge &edge) {
+                          result_nodes.emplace_back(edge.node_id);
+                      });
+
+        return result_nodes;
+    }
+
+    virtual std::vector<EdgeWeight>
+    GetUncompressedForwardWeights(const EdgeID id) const override final
+    {
+        /*
+         * EdgeWeights's for geometries are stored in one place for
+         * both forward and reverse segments along the same bi-
+         * directional edge. The m_geometry_indices stores
+         * refences to where to find the beginning of the bi-
+         * directional edge in the m_geometry_list vector. For
+         * forward weights of bi-directional edges, edges 2 to
+         * n of that edge need to be read.
+         */
+        const unsigned begin = m_geometry_indices.at(id) + 1;
+        const unsigned end = m_geometry_indices.at(id + 1);
+
+        std::vector<EdgeWeight> result_weights;
         result_weights.reserve(end - begin);
+
         std::for_each(m_geometry_list.begin() + begin,
                       m_geometry_list.begin() + end,
                       [&](const osrm::extractor::CompressedEdgeContainer::CompressedEdge &edge) {
-                          result_weights.emplace_back(edge.weight);
+                          result_weights.emplace_back(edge.forward_weight);
                       });
+
+        return result_weights;
+    }
+
+    virtual std::vector<EdgeWeight>
+    GetUncompressedReverseWeights(const EdgeID id) const override final
+    {
+        /*
+         * EdgeWeights for geometries are stored in one place for
+         * both forward and reverse segments along the same bi-
+         * directional edge. The m_geometry_indices stores
+         * refences to where to find the beginning of the bi-
+         * directional edge in the m_geometry_list vector. For
+         * reverse weights of bi-directional edges, edges 1 to
+         * n-1 of that edge need to be read in reverse.
+         */
+        const unsigned begin = m_geometry_indices.at(id);
+        const unsigned end = m_geometry_indices.at(id + 1) - 1;
+
+        std::vector<EdgeWeight> result_weights;
+        result_weights.reserve(end - begin);
+
+        std::for_each(m_geometry_list.rbegin() + (m_geometry_list.size() - end),
+                      m_geometry_list.rbegin() + (m_geometry_list.size() - begin),
+                      [&](const osrm::extractor::CompressedEdgeContainer::CompressedEdge &edge) {
+                          result_weights.emplace_back(edge.reverse_weight);
+                      });
+
+        return result_weights;
     }
 
     // Returns the data source ids that were used to supply the edge
     // weights.
-    virtual void
-    GetUncompressedDatasources(const EdgeID id,
-                               std::vector<uint8_t> &result_datasources) const override final
+    virtual std::vector<uint8_t>
+    GetUncompressedForwardDatasources(const EdgeID id) const override final
     {
-        const unsigned begin = m_geometry_indices.at(id);
+        /*
+         * Data sources for geometries are stored in one place for
+         * both forward and reverse segments along the same bi-
+         * directional edge. The m_geometry_indices stores
+         * refences to where to find the beginning of the bi-
+         * directional edge in the m_geometry_list vector. For
+         * forward datasources of bi-directional edges, edges 2 to
+         * n of that edge need to be read.
+         */
+        const unsigned begin = m_geometry_indices.at(id) + 1;
         const unsigned end = m_geometry_indices.at(id + 1);
 
-        result_datasources.clear();
+        std::vector<uint8_t> result_datasources;
         result_datasources.reserve(end - begin);
 
         // If there was no datasource info, return an array of 0's.
@@ -747,6 +828,47 @@ class InternalDataFacade final : public BaseDataFacade
                 m_datasource_list.begin() + end,
                 [&](const uint8_t &datasource_id) { result_datasources.push_back(datasource_id); });
         }
+
+        return result_datasources;
+    }
+
+    // Returns the data source ids that were used to supply the edge
+    // weights.
+    virtual std::vector<uint8_t>
+    GetUncompressedReverseDatasources(const EdgeID id) const override final
+    {
+        /*
+         * Datasources for geometries are stored in one place for
+         * both forward and reverse segments along the same bi-
+         * directional edge. The m_geometry_indices stores
+         * refences to where to find the beginning of the bi-
+         * directional edge in the m_geometry_list vector. For
+         * reverse datasources of bi-directional edges, edges 1 to
+         * n-1 of that edge need to be read in reverse.
+         */
+        const unsigned begin = m_geometry_indices.at(id);
+        const unsigned end = m_geometry_indices.at(id + 1) - 1;
+
+        std::vector<uint8_t> result_datasources;
+        result_datasources.reserve(end - begin);
+
+        // If there was no datasource info, return an array of 0's.
+        if (m_datasource_list.empty())
+        {
+            for (unsigned i = 0; i < end - begin; ++i)
+            {
+                result_datasources.push_back(0);
+            }
+        }
+        else
+        {
+            std::for_each(
+                m_datasource_list.rbegin() + (m_datasource_list.size() - end),
+                m_datasource_list.rbegin() + (m_datasource_list.size() - begin),
+                [&](const uint8_t &datasource_id) { result_datasources.push_back(datasource_id); });
+        }
+
+        return result_datasources;
     }
 
     virtual std::string GetDatasourceName(const uint8_t datasource_name_id) const override final
--- a/include/engine/datafacade/shared_datafacade.hpp
+++ b/include/engine/datafacade/shared_datafacade.hpp
@@ -79,7 +79,7 @@ class SharedDataFacade final : public BaseDataFacade
 
     util::ShM<util::Coordinate, true>::vector m_coordinate_list;
     util::PackedVector<OSMNodeID, true> m_osmnodeid_list;
-    util::ShM<NodeID, true>::vector m_via_node_list;
+    util::ShM<GeometryID, true>::vector m_via_geometry_list;
     util::ShM<unsigned, true>::vector m_name_ID_list;
     util::ShM<LaneDataID, true>::vector m_lane_data_id;
     util::ShM<extractor::guidance::TurnInstruction, true>::vector m_turn_instruction_list;
@@ -229,11 +229,12 @@ class SharedDataFacade final : public BaseDataFacade
 
     void LoadViaNodeList()
     {
-        auto via_node_list_ptr = data_layout->GetBlockPtr<NodeID>(
+        auto via_geometry_list_ptr = data_layout->GetBlockPtr<GeometryID>(
             shared_memory, storage::SharedDataLayout::VIA_NODE_LIST);
-        util::ShM<NodeID, true>::vector via_node_list(
-            via_node_list_ptr, data_layout->num_entries[storage::SharedDataLayout::VIA_NODE_LIST]);
-        m_via_node_list = std::move(via_node_list);
+        util::ShM<GeometryID, true>::vector via_geometry_list(
+            via_geometry_list_ptr,
+            data_layout->num_entries[storage::SharedDataLayout::VIA_NODE_LIST]);
+        m_via_geometry_list = std::move(via_geometry_list);
     }
 
     void LoadNames()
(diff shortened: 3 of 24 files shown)


TheMarex:
This is currently conflicting with some changes in the shared/internal datafacade. Will rebase once I finished the cucumber stuff.

include/extractor/original_edge_data.hpp
```diff
@@ -24,7 +24,7 @@ struct GeometryID
     {
     }
 
-    GeometryID() : id(std::numeric_limits<unsigned>::max()), forward(false)
+    GeometryID() : id(std::numeric_limits<unsigned>::max() >> 1), forward(false)
```
MoKob: Can we get a constant for this defined somewhere? If its an arbitrary value apart from 0, max, it is easy to mix it up in checks.

include/extractor/original_edge_data.hpp
```diff
@@ -46,7 +46,7 @@ struct OriginalEdgeData
     }
 
     OriginalEdgeData()
-        : via_geometry{std::numeric_limits<unsigned>::max(),false},
+        : via_geometry{std::numeric_limits<unsigned>::max() >> 1,false},
```
MoKob: also these additional locations are easy to mess up if the ID values should change

include/engine/datafacade/datafacade_base.hpp
```diff
@@ -70,20 +71,28 @@ class BaseDataFacade
     virtual util::Coordinate GetCoordinateOfNode(const unsigned id) const = 0;
     virtual OSMNodeID GetOSMNodeIDOfNode(const unsigned id) const = 0;
 
-    virtual unsigned GetGeometryIndexForEdgeID(const unsigned id) const = 0;
+    virtual extractor::GeometryID GetGeometryIndexForEdgeID(const unsigned id) const = 0;
 
-    virtual void GetUncompressedGeometry(const EdgeID id,
-                                         std::vector<NodeID> &result_nodes) const = 0;
+    virtual void GetUncompressedForwardGeometry(const EdgeID id,
+                                                std::vector<NodeID> &result_nodes) const = 0;
```
MoKob: given that we change this anyhow. Shouldn't we move on to `move` semantics? After all we rarely append here, right? Or do we?

include/engine/datafacade/internal_datafacade.hpp
```diff
@@ -76,7 +76,7 @@ class InternalDataFacade final : public BaseDataFacade
 
     util::ShM<util::Coordinate, false>::vector m_coordinate_list;
     util::PackedVector<OSMNodeID, false> m_osmnodeid_list;
-    util::ShM<NodeID, false>::vector m_via_node_list;
+    util::ShM<extractor::GeometryID, false>::vector m_via_node_list;
```
MoKob: why is a a node now referred to by a geometry id? Isn't the list still technically holding nodes?

include/engine/datafacade/internal_datafacade.hpp
```diff
@@ -201,7 +201,7 @@ class InternalDataFacade final : public BaseDataFacade
         {
             edges_input_stream.read((char *)&(current_edge_data),
                                     sizeof(extractor::OriginalEdgeData));
-            m_via_node_list[i] = current_edge_data.via_node;
+            m_via_node_list[i] = current_edge_data.via_geometry;
```
MoKob: see above

include/engine/datafacade/internal_datafacade.hpp
```diff
@@ -682,10 +682,11 @@ class InternalDataFacade final : public BaseDataFacade
...
     }
 
-    virtual void GetUncompressedGeometry(const EdgeID id,
-                                         std::vector<NodeID> &result_nodes) const override final
+    virtual void
+    GetUncompressedForwardGeometry(const EdgeID id,
+                                   std::vector<NodeID> &result_nodes) const override final
     {
-        const unsigned begin = m_geometry_indices.at(id);
+        const unsigned begin = m_geometry_indices.at(id) + 1;
```
MoKob: please add a comment explaining this `+1` here.

include/engine/datafacade/internal_datafacade.hpp
```diff
@@ -698,26 +699,58 @@ class InternalDataFacade final : public BaseDataFacade
     }
 
     virtual void
-    GetUncompressedWeights(const EdgeID id,
-                           std::vector<EdgeWeight> &result_weights) const override final
+    GetUncompressedReverseGeometry(const EdgeID id,
+                                   std::vector<NodeID> &result_nodes) const override final
     {
         const unsigned begin = m_geometry_indices.at(id);
+        const unsigned end = m_geometry_indices.at(id + 1) - 1;
```
MoKob: same as above. please comment on why we need the `-1`

include/engine/datafacade/internal_datafacade.hpp
```diff
@@ -698,26 +699,58 @@ class InternalDataFacade final : public BaseDataFacade
...
+
+        result_nodes.clear();
+        result_nodes.reserve(end - begin);
+        std::for_each(m_geometry_list.rbegin() + (m_geometry_list.size() - end),
```
MoKob: this looks just like a `boost::transform` to me

include/engine/datafacade/internal_datafacade.hpp
```diff
@@ -698,26 +699,58 @@ class InternalDataFacade final : public BaseDataFacade
...
+    virtual void
+    GetUncompressedReverseWeights(const EdgeID id,
+                                  std::vector<EdgeWeight> &result_weights) const override final
+    {
+        const unsigned begin = m_geometry_indices.at(id);
+        const unsigned end = m_geometry_indices.at(id + 1) - 1;
+
+        result_weights.clear();
+        result_weights.reserve(end - begin);
+        std::for_each(m_geometry_list.rbegin() + (m_geometry_list.size() - end),
```
MoKob: again `boost::transform`

include/engine/datafacade/internal_datafacade.hpp
```diff
@@ -698,26 +699,58 @@ class InternalDataFacade final : public BaseDataFacade
...
+                          result_weights.emplace_back(edge.forward_weight);
+                      });
+    }
+
+    virtual void
+    GetUncompressedReverseWeights(const EdgeID id,
+                                  std::vector<EdgeWeight> &result_weights) const override final
+    {
+        const unsigned begin = m_geometry_indices.at(id);
+        const unsigned end = m_geometry_indices.at(id + 1) - 1;
```
MoKob: documentation for the `-1`

include/engine/datafacade/internal_datafacade.hpp
```diff
@@ -698,26 +699,58 @@ class InternalDataFacade final : public BaseDataFacade
...
+                      [&](const osrm::extractor::CompressedEdgeContainer::CompressedEdge &edge) {
+                          result_nodes.emplace_back(edge.node_id);
+                      });
+    }
+
+    virtual void
+    GetUncompressedForwardWeights(const EdgeID id,
+                                  std::vector<EdgeWeight> &result_weights) const override final
+    {
+        const unsigned begin = m_geometry_indices.at(id) + 1;
```
MoKob: documentation for the `+1`

include/engine/datafacade/internal_datafacade.hpp
MoKob: given that this access happens at so many locations, possibly we should encapsulate it into a dedicated function?

include/engine/datafacade/internal_datafacade.hpp
```diff
@@ -742,6 +775,35 @@ class InternalDataFacade final : public BaseDataFacade
...
     }
 
+    // Returns the data source ids that were used to supply the edge
+    // weights.
+    virtual void
+    GetUncompressedReverseDatasources(const EdgeID id,
+                                      std::vector<uint8_t> &result_datasources) const override final
+    {
+        const unsigned begin = m_geometry_indices.at(id);
+        const unsigned end = m_geometry_indices.at(id + 1) - 1;
```
MoKob: replace by functions (see above)

include/engine/datafacade/internal_datafacade.hpp
```diff
@@ -742,6 +775,35 @@ class InternalDataFacade final : public BaseDataFacade
...
+
+        result_datasources.clear();
+        result_datasources.reserve(end - begin);
```
MoKob: since we clear and allocate anyhow, we should really move on to returning a moved out vector.
In theory I would prefer so simply have an iterator view of this, so we don't copy back and forward.

include/engine/datafacade/internal_datafacade.hpp
```diff
@@ -742,6 +775,35 @@ class InternalDataFacade final : public BaseDataFacade
...
+        if (m_datasource_list.empty())
+        {
+            for (unsigned i = 0; i < end - begin; ++i)
+            {
+                result_datasources.push_back(0);
+            }
+        }
+        else
+        {
+            std::for_each(
```
MoKob: boost::transform

include/engine/datafacade/shared_datafacade.hpp
```diff
@@ -79,7 +79,7 @@ class SharedDataFacade final : public BaseDataFacade
 
     util::ShM<util::Coordinate, true>::vector m_coordinate_list;
     util::PackedVector<OSMNodeID, true> m_osmnodeid_list;
-    util::ShM<NodeID, true>::vector m_via_node_list;
+    util::ShM<extractor::GeometryID, true>::vector m_via_node_list;
```
MoKob: again, referring to nodes by a geometry ID. They are still node ids, right? Each node also has a coordinate, though.

include/engine/datafacade/shared_datafacade.hpp
```diff
@@ -534,22 +535,56 @@ class SharedDataFacade final : public BaseDataFacade
...
 
     virtual void
-    GetUncompressedWeights(const EdgeID id,
-                           std::vector<EdgeWeight> &result_weights) const override final
+    GetUncompressedReverseGeometry(const EdgeID id,
+                                   std::vector<NodeID> &result_nodes) const override final
     {
-        const unsigned begin = m_geometry_indices.at(id);
+        const signed begin = m_geometry_indices.at(id) - 1;
+        const signed end = m_geometry_indices.at(id + 1) - 2;
```
MoKob: why `-2` here now? These accesses really need explaining

include/engine/datafacade/shared_datafacade.hpp
```diff
@@ -534,22 +535,56 @@ class SharedDataFacade final : public BaseDataFacade
...
+
+        result_nodes.clear();
```
MoKob: also, all of this looks like it is duplicated code, simply based on which datafacade we use? Doesn't this scream for a template function somewhere else that wraps the access for both datafacades, so we don't mess up two locations of code?

include/engine/datafacade/shared_datafacade.hpp
```diff
@@ -534,22 +535,56 @@ class SharedDataFacade final : public BaseDataFacade
...
+                          result_weights.emplace_back(edge.forward_weight);
                       });
     }
 
-    virtual unsigned GetGeometryIndexForEdgeID(const unsigned id) const override final
+    virtual void
+    GetUncompressedReverseWeights(const EdgeID id,
+                                  std::vector<EdgeWeight> &result_weights) const override final
+    {
+        const signed begin = m_geometry_indices.at(id) - 1;
```
MoKob: -1 here now? I am confused even more

include/engine/datafacade/shared_datafacade.hpp
```diff
@@ -534,22 +535,56 @@ class SharedDataFacade final : public BaseDataFacade
...
+        const signed end = m_geometry_indices.at(id + 1) - 2;
+
+        result_weights.clear();
+        result_weights.reserve(end - begin);
+        for (signed i = end; i > begin; --i)
```
MoKob: boost::transform

include/engine/datafacade/shared_datafacade.hpp
```diff
@@ -775,6 +810,35 @@ class SharedDataFacade final : public BaseDataFacade
...
+        if (m_datasource_list.empty())
+        {
+            for (unsigned i = 0; i < end - begin; ++i)
+            {
+                result_datasources.push_back(0);
+            }
+        }
+        else
+        {
+            std::for_each(
```
MoKob: @daniel-j-h what is your feeling here, could we make all of these accesses lazy copy views? Most times we will need all data, but for all the cases that only look at the first few items along a road, all this copying could be saved.

include/engine/phantom_node.hpp
```diff
@@ -78,7 +76,7 @@ struct PhantomNode
           reverse_segment_id{SPECIAL_SEGMENTID, false},
           name_id(std::numeric_limits<unsigned>::max()), forward_weight(INVALID_EDGE_WEIGHT),
           reverse_weight(INVALID_EDGE_WEIGHT), forward_offset(0), reverse_offset(0),
-          forward_packed_geometry_id(SPECIAL_EDGEID), reverse_packed_geometry_id(SPECIAL_EDGEID),
+          packed_geometry_id(SPECIAL_EDGEID),
```
MoKob: why is this an EDGE_ID now? We have GeometryID, which is referring to Nodes and assigned an invalid EdgeID?

include/extractor/compressed_edge_container.hpp
```diff
@@ -35,11 +45,16 @@ class CompressedEdgeContainer
     void
     AddUncompressedEdge(const EdgeID edgei_id, const NodeID target_node, const EdgeWeight weight);
 
+    void InitializeBothwayVector();
```
MoKob: Documentation. What is a `bothway` vector?

include/extractor/compressed_edge_container.hpp
```diff
@@ -35,11 +45,16 @@ class CompressedEdgeContainer
...
+    unsigned ZipEdges(const unsigned f_edge_pos, const unsigned r_edge_pos);
```
MoKob: documentation, what are we zipping, how are we doing it and what kind of access impacts does that have?
Also, is this related to the main goal of the compressed geometry deduplication? Seems like an unrelated addition to me at first glance.

include/extractor/original_edge_data.hpp
```diff
@@ -13,28 +13,47 @@ namespace osrm
...
                               TravelMode travel_mode)
-        : via_node(via_node), name_id(name_id), entry_classid(entry_classid),
+        : via_geometry(via_geometry), name_id(name_id), entry_classid(entry_classid),
           lane_data_id(lane_data_id), turn_instruction(turn_instruction), travel_mode(travel_mode)
     {
     }
 
     OriginalEdgeData()
-        : via_node(std::numeric_limits<unsigned>::max()),
+        : via_geometry{std::numeric_limits<unsigned>::max() >> 1,false},
```
MoKob: dedicated constant

src/contractor/contractor.cpp
```diff
@@ -556,102 +556,64 @@ EdgeID Contractor::LoadEdgeExpandedGraph(
                 extractor::QueryNode *u;
                 extractor::QueryNode *v;
 
-                if (leaf_object.forward_packed_geometry_id != SPECIAL_EDGEID)
+                const unsigned forward_begin =
+                    m_geometry_indices.at(leaf_object.packed_geometry_id);
+
+                u = &(internal_to_external_node_map
```
MoKob: this would be so much nicer not using pointers but iterators

src/contractor/contractor.cpp
```diff
@@ -556,102 +556,64 @@ EdgeID Contractor::LoadEdgeExpandedGraph(
...
-
-                        // count statistics for logging
-                        counters[forward_speed_iter->speed_source.source] += 1;
-                    }
-                    else
-                    {
-                        // count statistics for logging
-                        counters[LUA_SOURCE] += 1;
-                    }
+                    auto new_segment_weight =
```
MoKob: const

src/contractor/contractor.cpp
```diff
@@ -556,102 +556,64 @@ EdgeID Contractor::LoadEdgeExpandedGraph(
...
-                    else
-                    {
-                        // count statistics for logging
-                        counters[LUA_SOURCE] += 1;
-                    }
+                auto reverse_speed_iter =
+                    find(segment_speed_lookup, Segment{v->node_id, u->node_id});
+                if (reverse_speed_iter != segment_speed_lookup.end())
+                {
+                    auto new_segment_weight =
```
MoKob: const

include/engine/datafacade/internal_datafacade.hpp
jakepruitt: It used to hold nodes a few months ago, then it held nodes _or_ references to the `m_compressed_edge_container` array. For the last few months (as of https://github.com/Project-OSRM/osrm-backend/commit/21245273b414f6034fa9e686fe46ac387fa97f8f) this list has only been references to the `m_compressed_edge_container`. I think the appropriate thing to do here would be to change the name of this list to `m_via_geometry_list`.

include/engine/datafacade/datafacade_base.hpp
jakepruitt: I don't think we ever use the existing data in `result_nodes`. I'll change this to return the vector.

lbud:
git-blame-it-all-on-me, @jakepruitt if you want we can walk through all my aforementioned C++ blunders together for commenting purposes etc

include/engine/phantom_node.hpp
jakepruitt: I'm not sure whether we should use GeometryID in this situation or a raw NodeID. For the `m_via_node_geometry_ids` array in the datasources, we needed to pass through the direction from which to read the via-geometry for an edge-based-edge. I'm still trying to determine whether the same thing needs to happen for phantom nodes.

include/extractor/edge_based_graph_factory.hpp
```diff
@@ -84,7 +84,7 @@ class EdgeBasedGraphFactory
     EdgeBasedGraphFactory &operator=(const EdgeBasedGraphFactory &) = delete;
 
     explicit EdgeBasedGraphFactory(std::shared_ptr<util::NodeBasedDynamicGraph> node_based_graph,
-                                   const CompressedEdgeContainer &compressed_edge_container,
+                                   CompressedEdgeContainer &compressed_edge_container,
```
jakepruitt: @MoKob I think the EdgeBasedGraphFactory now calls methods that refactor the internal storage of the compressed geometries from the single-direction geometries to the bidirectional geometries. I think this happens in the `InsertEdgeBasedNode` method because at that point we have access to node-based edges that point to the same geometry but in differetn directions.

Is there a better time to create the internal bidirectional geometry tables?

jakepruitt:
All cucumber tests are now passing - but I'm seeing a failure in the server unit tests related to base64 encoding:

```
0.01s$ ./unit_tests/server-tests

Running 10 test cases...

[assert] /home/travis/build/Project-OSRM/osrm-backend/src/engine/hint.cpp:39

in: static osrm::engine::Hint osrm::engine::Hint::FromBase64(const std::string &): base64Hint.size() == ENCODED_HINT_SIZE

Hint has invalid sizeterminate called without an active exception

unknown location(0): fatal error in "valid_route_urls": signal: SIGABRT (application abort requested)

/home/travis/build/Project-OSRM/osrm-backend/unit_tests/server/parameters_parser.cpp(133): last checkpoint

*** 1 failure detected in test suite "server tests"
```

I think next steps here are:

- [ ] Fix that server unit test
- [ ] Look into changing the `GetUncompressed*` methods into returning iterators rather than copying and returning arrays
- [ ] Do whatever is needed to change the reverse iterator in the shared memory datafacade to work with `std::for_each` (rather than our current for loop).
- [ ] Comment all of the `+1` and `-1`
- [ ] Use `boost::transform`
- [ ] Update documentation where possible

oxidase:
@jakepruitt i have updated hints in tests. 

There is a possibility to improve doubled memory consumption by removing `m_compressed_oneway_geometries` and fill `m_compressed_geometries` directly in `CompressEdge`

I have committed into https://github.com/oxidase/osrm-backend/commit/f7dde4d4981dec7e8445a4dd9c8f5b2349db352a
removal of `GetBucketReference` from the public interface.

include/engine/phantom_node.hpp
MoKob: Whatever we use, we should be consistent with SPECIAL_EDGEID to be assigned to edge_ids, geometry_ids only using a SPECIAL_GEOMETRYID.

include/extractor/edge_based_graph_factory.hpp
MoKob: I am not entirely sure where the best location would be. Having this done here adds an additional potential for difficult debugging. If possible, I'd love to have this separated from the EdgeBasedGraphFactory.

include/engine/datafacade/shared_datafacade.hpp
jakepruitt: @MoKob I think the one thing preventing us from switching to a shared templated function is that reverse iterators don't seem to work on the shared memory `m_geometry_list` or `m_datasource_list`. I'm not sure if this is an easy fix or not.

include/engine/hint.hpp
```diff
@@ -63,8 +63,8 @@ struct Hint
     friend std::ostream &operator<<(std::ostream &, const Hint &);
 };
 
-static_assert(sizeof(Hint) == 60 + 4, "Hint is bigger than expected");
-constexpr std::size_t ENCODED_HINT_SIZE = 88;
+static_assert(sizeof(Hint) == 56 + 4, "Hint is bigger than expected");
+constexpr std::size_t ENCODED_HINT_SIZE = 80;
```
jakepruitt: @lbud do you know why the hint sizes changed here? Was this part of a bad rebase?

include/util/shared_memory_vector_wrapper.hpp
```diff
@@ -46,6 +46,34 @@ template <typename DataT> class ShMemIterator : public std::iterator<std::input_
...
+    ShMemReverseIterator operator+(std::ptrdiff_t diff)
+    {
+        ShMemReverseIterator tmp(p - diff);
+        return tmp;
+    }
+    bool operator==(const ShMemReverseIterator &rhs) { return p == rhs.p; }
+    bool operator!=(const ShMemReverseIterator &rhs) { return p != rhs.p; }
+    DataT &operator*() { return *p; }
+};
+
```
jakepruitt: @lbud do you know if this shared memory reverse iterator implementation worked out? Looks like the shared datafacade is still using a `for` loop instead of iterators. Would getting this reverse iterator working allow us to switch that over to iterators?

src/extractor/edge_based_graph_factory.cpp
```diff
@@ -442,9 +445,14 @@ void EdgeBasedGraphFactory::GenerateEdgeExpandedEdges(
 
                 distance += turn_penalty;
 
-                BOOST_ASSERT(m_compressed_edge_container.HasEntryForID(edge_from_u));
+                const NodeID smaller = node_u < node_v ? node_u : node_v;
+                const NodeID larger = node_u > node_v ? node_u : node_v;
+                const EdgeID smaller_to_larger_edge = m_node_based_graph->FindEdge(smaller, larger);
+                const EdgeData &smaller_to_larger_edge_data = m_node_based_graph->GetEdgeData(smaller_to_larger_edge);
+                const bool is_encoded_forwards = smaller_to_larger_edge_data.edge_id == SPECIAL_NODEID ? node_u > node_v : node_u < node_v;
```
jakepruitt: This logic is my least favorite part of the PR, but was the trick that ultimately fixed the errors on the branch. Edges are "zipped" together in the `InsertEdgeBasedNode` method, determining which direction is "forwards" and "backwards" by the order of parameters passed to `CompressedEdge::ZipEdges`. The [calling logic in `InsertEdgeBasedNode`](https://github.com/Project-OSRM/osrm-backend/blob/master/src/extractor/edge_based_graph_factory.cpp#L280-L283) is set up so that the edge will always be oriented from the "smaller" node id to the "larger" node id unless the edge from the smaller to larger node is non-existent, i.e. the edge_id is `SPECIAL_NODEID`. I've seen the non-existent edge happen when the edge from u to v is a reversed oneway.

@MoKob @daniel-j-h do you have any good ideas for going about this in a less-brittle way? As mentioned [here](https://github.com/Project-OSRM/osrm-backend/pull/2717/files#r79394824), we probably want to move the zipping to another part of the extraction process than `EdgeBasedGraphFactory::InsertEdgeBasedNode`. But I'm not sure there's another place in the extraction process where we have the forward and reverse edge id's for a pair of node-based-nodes.

src/extractor/edge_based_graph_factory.cpp
```diff
@@ -442,9 +445,14 @@ void EdgeBasedGraphFactory::GenerateEdgeExpandedEdges(
 
                 distance += turn_penalty;
 
-                BOOST_ASSERT(m_compressed_edge_container.HasEntryForID(edge_from_u));
+                const NodeID smaller = node_u < node_v ? node_u : node_v;
```
daniel-j-h: This and the following is min/max(nodeu,nodev) (?)

jakepruitt: @daniel-j-h yep! which all falls under what I described here: https://github.com/Project-OSRM/osrm-backend/pull/2717#discussion_r79461532

jakepruitt:
Spent a few hours trying to figure out how to return a transformed range, but ended up not succeeding. @MoKob @oxidase @daniel-j-h my attempt was here: https://github.com/Project-OSRM/osrm-backend/commit/77f844b2a0c0bc6c8f8f66904a2d12b5fe40dc56 in case you see any obvious red flags. I didn't have the benefit of setting the return value to `auto` since both of these functions needed to be defined virtually, so I was mostly poking through different permutations of boost types without success.

jakepruitt:
@oxidase sorry for not getting back to you soon enough - your commit https://github.com/oxidase/osrm-backend/commit/f7dde4d4981dec7e8445a4dd9c8f5b2349db352a looks great to pull into this PR. My only question is - does it delete the `m_compressed_oneway_geometries` anywhere? Do the `m_compressed_oneway_geometries` automatically get deleted? I know that we don't serialize the `m_compressed_oneway_geometries` which is the main goal of the PR, but it'd definitely be great to be more careful about our memory usage 👍

oxidase:
@jakepruitt the commit oxidase@f7dde4d  is not finished and  just a hint that `m_compressed_oneway_geometries` is not needed externally. Missing part is compressing of zipped edges. It will allow to remove `m_compressed_oneway_geometries` and restore constness of 
`CompressedEdgeContainer &compressed_edge_container` at https://github.com/Project-OSRM/osrm-backend/pull/2717#r78716508

include/engine/hint.hpp
oxidase: @jakepruitt this is correct and expected, because `PhantomNode`  has only one  4-bytes packed geometry id instead of two

jakepruitt:
@daniel-j-h still can't figure out what templating ✨  is needed to return a transform iterator pair from `GetUncompressedForwardGeometry`. My underlying problem is that the function is defined in `includes/engine/datafacade/datafacade_base.hpp` as:

```c++
virtual std::vector<NodeID> GetUncompressedForwardGeometry(const EdgeID id) const = 0;
```

and overridden in `includes/engine/datafacade/internal_datafacade.hpp` and `includes/engine/datafacade/shared_datafacade.hpp`. Since the `m_geom_indices` member attribute has a different type depending on internal or shared datafacade, the type of the iterator used to define the type of the boost `transform_iterator` returned by the `GetUncompressedForwardGeometry` will be different for each datafacade. When I try to leave the return type as `auto` in 

```c++
virtual auto GetUncompressedForwardGeometry(const EdgeID id) const = 0;
```

I get `error: function with deduced return type cannot be virtual`. @daniel-j-h (dumb c++ question warning) is there a magical way to let the return type of a virtual function vary depending on the implementation? Is there a way to abstract the type of the return iterator away using templating? Or are we stuck returning a copy of the internal data so that we can hide implementation differences?

daniel-j-h:
Capturing from chat: you can't make a virtual function a template (type deduction with auto), and you can't have different types as iterators depending on internal or shared facade.

What you can do is noticing the vectors are contiguous blocks of memory. So what you can do is returning a `const CompressedEdges *` for both of them since the underlying item types are the same.

```cpp
const auto *first = &*begin(geom);
const auto *last = &*end(geom);
```

(`begin(geom)` gives you an iterator, `*it` dereferences the iterator to get the item, `&item` gives you the address of the item - resulting in a pointer to the first / last item in the containers)

and then wrap those in transform iterators. Conceptually this should work, give it a try!

jakepruitt:
@daniel-j-h So your recommendation *did* work for iterators that needed to go forward through the `m_geometry` list, but iterators that needed to go *in reverse* got very borked. At the moment, the downstream client of the `GetUncompressedForwardGeometry` and `GetUncompressedReverseGeometry` expect the same return type for each of these methods, so I don't think we can return a different `reversed` iterator or transform iterator or iterator range for those methods without more serious changes. I committed what I have to the https://github.com/Project-OSRM/osrm-backend/tree/refactor/merge_geometries_3 branch, so if you get a chance to 👀 at the implementation there and help me figure out a nice way to have reversed ranges or forward ranges, I think we'll be almost there.

I have a hunch the secret is templating ✨

include/engine/datafacade/shared_datafacade.hpp
TheMarex: This is the goal of https://github.com/Project-OSRM/osrm-backend/issues/2575

It might be better to have this as a separate PR.

daniel-j-h:
@jakepruitt a couple of things here: do not use (Boost) `detail` namespaces, those are "internal" and can change anytime. They should be handled as impl. detail.

Re. reverse iterators: returning pointers allows you to return a "chunk" of memory. But raw pointers are not smart enough to handle actual iteration as they are two memory addresses only: the decision to iterate front-to-back or back-to-front over the memory needs to be done on top of those pointers / on the call sites.

src/engine/plugins/tile.cpp
```diff
@@ -527,13 +527,12 @@ Status TilePlugin::HandleRequest(const api::TileParameters &parameters, std::str
                     // This is the geometry leading away from the intersection
                     // (i.e. the geometry of the target edge-based-node)
                     second_geometry.clear();
-                    facade.GetUncompressedGeometry(
-                        edge_based_node_info.at(target_ebn).packed_geometry_id, second_geometry);
+                    second_geometry = facade.GetUncompressedReverseGeometry(
```
TheMarex: Again we only use the reverse geometry here to extract the missing node that is sliced off the forward geometry. We could remove this if we change the interface.

src/engine/plugins/tile.cpp
```diff
@@ -610,38 +609,35 @@ Status TilePlugin::HandleRequest(const api::TileParameters &parameters, std::str
         // TODO this approach of writing at least an empty vector for any segment is probably stupid
         // (inefficient)
 
-        if (edge.forward_packed_geometry_id != SPECIAL_EDGEID)
+        if (edge.packed_geometry_id != SPECIAL_GEOMETRYID)
```
TheMarex: As above this will always evaluate to true now.

src/engine/plugins/tile.cpp
```diff
@@ -610,38 +609,35 @@ Status TilePlugin::HandleRequest(const api::TileParameters &parameters, std::str
...
                 reverse_weight_vector[reverse_weight_vector.size() - edge.fwd_segment_position - 1];
 
+            forward_datasource_vector.clear();
+            forward_datasource_vector = facade.GetUncompressedForwardDatasources(edge.packed_geometry_id);
+            forward_datasource = forward_datasource_vector[edge.fwd_segment_position];
+
             reverse_datasource_vector.clear();
-            facade.GetUncompressedDatasources(edge.reverse_packed_geometry_id,
-                                              reverse_datasource_vector);
+            // TODO have not tested geom zipping with tiles yet
```
TheMarex: Do you know the context of this TODO? Did you check the tiles? :-)

src/engine/plugins/tile.cpp
```diff
@@ -698,36 +694,40 @@ Status TilePlugin::HandleRequest(const api::TileParameters &parameters, std::str
...
+                    forward_datasource = forward_datasource_vector[edge.fwd_segment_position];
+
+                    reverse_weight_vector.clear();
+                    reverse_weight_vector =
+                        facade.GetUncompressedReverseWeights(edge.packed_geometry_id);
 
                     std::string name = facade.GetNameForID(edge.name_id);
 
-                    if (edge.forward_packed_geometry_id != SPECIAL_EDGEID)
+                    if (edge.packed_geometry_id != SPECIAL_EDGEID)
```
TheMarex: Again trivial check that should be an assertin.

src/extractor/edge_based_graph_factory.cpp
TheMarex: I think what also should work is:

- Change `CompressedEdgeContainer` to have `m_forward_edge_id_to_zipped_index_map` and `m_reverse_edge_id_to_zipped_index_map`.
- Instead of `HasZippedEntryForID` implement `IsEncodedForward()` and `IsEncodedBackward()` that return true if there is an entry in the corresponding map
- Change `GetZippedPositionForID()` to be directional: `GetZippedPositionForForwardID()` .. `GetZippedPositionForReverseID()`
- You can set `is_encoded_forward` depending on the `CompressedGeometryContainer` functions

That way the only thing we rely on is that for every edge id there is an entry, either in forward or reverse encoding, no explicit numbering assumptions.

jakepruitt:
@TheMarex changed the direction-determining code and it works great! I am running into problems with the change for `GetUncompressedForwardGeometry` to return all of the node ids along an edge - I implemented the change here if you want to take a look at it: https://github.com/Project-OSRM/osrm-backend/commit/3062304f1185cd1a72328a56e70d8ca7f345cab8

In my opinion, I'm fine with the way we're returning 1 less than the path's geometry nodes. The hack in `assemble_geometries.hpp` doesn't seem as nasty as the hacks necessary in `routing_algorithms.hpp` to line up `weight_vector` with `id_vector`.

TheMarex:
@jakepruitt with hacks your mean the index offset here:

https://github.com/Project-OSRM/osrm-backend/commit/3062304f1185cd1a72328a56e70d8ca7f345cab8#diff-ffa2e7ac24f3bebb547c6fe24552431bL292

Hm. I can see how that could be error prone. However at the moment we implicitly encode the geometry by only saving segment endpoints, that is also very confusing. I would prefer to make the indexing for the geometry more explicit (e.g. rename idx in the above example to segment_index) rather then keep the geometry encoding.

jakepruitt:
Implemented changes and ready for review again, @TheMarex.

TheMarex:
@jakepruitt looked at the memory savings and I realized we will see a lot less improvement since we now save an extra entry for both weight values. This only scales if we have very long geometries in the dataset and save on that. For example Berlin size didn't reduce because there are just not that many long streets, but the local State Baden-Wüttemberg is 10% smaller since there are a lot long streets.

This might need a rebase again. Did one this afternoon but seems there are conflicts again. Other then that, I think this is good to merge. :+1:

Issue #2716, opened 2016-07-29
More Robust Roundabout Radius Calculation


daniel-j-h:
For three-or-more-node roundabouts we're currently only using the first three nodes for calculating the roundabout's radius. This results in problems when the roundabout is not perfectly round but instead nodes later on are places far outside the perfect circle.

Consider this partial example:
```
    *
 *
    *
       *
           *                                *
                *                         *
                     *                 *
                            *     *

```

or when a rotary way is modelled in detail:
```
     *
   *
      *
     *
       *
         *
            *
                *
```

The first three _roundabout nodes_ make up for a small radius, therefore we never consider this a rotary.

Location: https://github.com/Project-OSRM/osrm-backend/blob/b0873e2aa42f05825183faa2919b2f8d5502c455/src/extractor/guidance/roundabout_handler.cpp#L292-L309

oxidase:
One possibility could be an MSE fitting of a circle into coordinate points, so it will give the center and the radius of the roundabout. There are many approximations to avoid costly QP, like Kåsa fit http://people.cas.uab.edu/~mosya/cl/MATLABcircle.html

MoKob:
Just a note: For all calculations we have to keep in mind that the radius computations need to be identical coming from all entries/exits. I have to have a look at the different options, but right now my feeling would be that getting the extremes (min/max lat/lon) and selecting three fitting vertices could be a simple way.

MoKob:
An alternative approach that came to me would be to use the circumference of the roundabout rather than the radius (or guess a radius from the circumference). If the circles don't follow an actual circle, this could be more robust and less error-prone.

@daniel-j-h @oxidase Whats your take on this?

oxidase:
This sounds good, but it requires two assumptions:
* all points are in the convex hull
* points are ordered

Otherwise it would require construction of the convex hull.

MoKob:
Well, we don't need to have a correct radius. We can make the full assumption on when to announce as roundabout/rotary based on `duration` that we are in the roundabout. So if a full traversal of a roundabout takes `x` seconds, we assume it to be small. Or if it is at least `m` meters long, we assume it to be a rotary.

Using this we would not need a complex calculation of the radius but rather only traverse all edges for their length.

This might be just as robust, if not even more. The distance calculation is even already available. Currently its just not stored with the edges of the graph.

PR #2718, opened 2016-07-29
[Ready] Be more aggressive classifying Roundabout Intersections

daniel-j-h:
Roundabout Intersections are roundabouts with up to four ways and turn
angles which makes the turns obvious, e.g. as in:

```
    *
    *
* *   * *
    *
    *
```

but not e.g.

```
     *
    *
* *   *
    * *
     * *
```

For Roundabout Intersections we issue instructions such as
"turn \<direction\>" instead of "take the \<nth\> exit".

At the moment we have a limit on the radius for these Roundabout
Intersections of 5 meters. Which fails to classify a wide range of
Roundabout Intersections in the US (with the US-wide streets).

Example: http://www.openstreetmap.org/way/348812150

This changeset removes the Roundabout Intersection radius limit:

- if the roundabout is larger than a threshold and is named we classify
  it as a rotary

- if the roundabout matches our criteria for Roundabout Intersections
  we classify it as a Roundabout Intersection (note: we no longer care
  about its radius here)

- else fall back to plain old Roundabout

Up for discussion. cc @1ec5 @MoKob @TheMarex

There is a second issue with determining a roundabout's radius.
But that's for another pull request (tracking in #2716).

### DIFF
--- a/features/guidance/roundabout-bike.feature
+++ b/features/guidance/roundabout-bike.feature
@@ -5,37 +5,6 @@ Feature: Basic Roundabout
         Given the profile "bicycle"
         Given a grid size of 10 meters
 
-    Scenario: Enter and Exit
-        Given the node map
-            |   |   | a |   |   |
-            |   |   | b |   |   |
-            | h | g |   | c | d |
-            |   |   | e |   |   |
-            |   |   | f |   |   |
-
-       And the ways
-            | nodes  | junction   |
-            | ab     |            |
-            | cd     |            |
-            | ef     |            |
-            | gh     |            |
-            | bgecb  | roundabout |
-
-       When I route I should get
-           | waypoints | route    | turns                           |
-           | a,d       | ab,cd,cd | depart,roundabout-exit-3,arrive |
-           | a,f       | ab,ef,ef | depart,roundabout-exit-2,arrive |
-           | a,h       | ab,gh,gh | depart,roundabout-exit-1,arrive |
-           | d,f       | cd,ef,ef | depart,roundabout-exit-3,arrive |
-           | d,h       | cd,gh,gh | depart,roundabout-exit-2,arrive |
-           | d,a       | cd,ab,ab | depart,roundabout-exit-1,arrive |
-           | f,h       | ef,gh,gh | depart,roundabout-exit-3,arrive |
-           | f,a       | ef,ab,ab | depart,roundabout-exit-2,arrive |
-           | f,d       | ef,cd,cd | depart,roundabout-exit-1,arrive |
-           | h,a       | gh,ab,ab | depart,roundabout-exit-3,arrive |
-           | h,d       | gh,cd,cd | depart,roundabout-exit-2,arrive |
-           | h,f       | gh,ef,ef | depart,roundabout-exit-1,arrive |
-
     Scenario: Only Enter
         Given the node map
             |   |   | a |   |   |
--- a/features/guidance/roundabout-turn.feature
+++ b/features/guidance/roundabout-turn.feature
@@ -389,3 +389,105 @@ Feature: Basic Roundabout
            | a,f       | ab,ef,ef | depart,roundabout turn straight exit-2,arrive | 0->180,180->224,180->0 |
            | a,h       | ab,gh,gh | depart,roundabout turn right exit-1,arrive    | 0->180,180->224,270->0 |
 
+    Scenario: Enter and Exit - Bearings
+        Given the node map
+            |   |   |   | a |   |   |   |
+            |   |   |   |   |   |   |   |
+            |   |   | i | b | l |   |   |
+            | h |   | g |   | c |   | d |
+            |   |   | j | e | k |   |   |
+            |   |   |   |   |   |   |   |
+            |   |   |   | f |   |   |   |
+
+       And the ways
+            | nodes      | junction   |
+            | ab         |            |
+            | cd         |            |
+            | ef         |            |
+            | gh         |            |
+            | bigjekclb  | roundabout |
+
+       When I route I should get
+           | waypoints | route    | turns                                         | bearing                |
+           | a,d       | ab,cd,cd | depart,roundabout turn left exit-3,arrive     | 0->180,180->270,90->0  |
+           | a,f       | ab,ef,ef | depart,roundabout turn straight exit-2,arrive | 0->180,180->270,180->0 |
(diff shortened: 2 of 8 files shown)


features/guidance/roundabout-turn.feature
```diff
@@ -389,3 +389,44 @@ Feature: Basic Roundabout
            | a,f       | ab,ef,ef | depart,roundabout turn straight exit-2,arrive | 0->180,180->224,180->0 |
            | a,h       | ab,gh,gh | depart,roundabout turn right exit-1,arrive    | 0->180,180->224,270->0 |
 
+
+    Scenario: Large radius Roundabout Intersection and ways modelled out: East Mission St, North 7th St
```
daniel-j-h: This scenario is new, the other ones are adapted to Roundabout Intersections.

1ec5:
That sounds reasonable for the case given above, which is essentially just a traffic calming device. Do you recall how the five-meter threshold was determined? Are there counterexamples in Europe where a five-meter-wide roundabout would be a conventional junction rather than a traffic-calming device?

daniel-j-h:
I don't recall the 5 meters heuristic --- could be totally guestimated.
We have to wait for @MoKob for insight here.

My main point is: once we have determined it's not a rotary (large named roundabout), does the size actually matter for classifying these as roundabout intersections?

As long as the roundabout intersection criteria are met, I see no reason (other than maybe mis-tagged rotaries?) to not announce them as turns.

emiltin:
i agree that it makes sense to announce turns if it's not a rotary and the turns are almost 90 degree. for example this series of roundabouts:
http://www.openstreetmap.org/#map=16/55.6564/12.5864

in additional to turns being almost 90 degree, there clearly a main road continuing straight ahead, which makes "straight ahead" even more natural.

features/guidance/roundabout-bike.feature
```diff
@@ -22,19 +22,19 @@ Feature: Basic Roundabout
...
+           | a,h       | ab,gh,gh | depart,roundabout turn right exit-1,arrive    |
+           | d,f       | cd,ef,ef | depart,roundabout turn left exit-3,arrive     |
+           | d,h       | cd,gh,gh | depart,roundabout turn straight exit-2,arrive |
+           | d,a       | cd,ab,ab | depart,roundabout turn right exit-1,arrive    |
+           | f,h       | ef,gh,gh | depart,roundabout turn left exit-3,arrive     |
+           | f,a       | ef,ab,ab | depart,roundabout turn straight exit-2,arrive |
+           | f,d       | ef,cd,cd | depart,roundabout turn right exit-1,arrive    |
+           | h,a       | gh,ab,ab | depart,roundabout turn left exit-3,arrive     |
+           | h,d       | gh,cd,cd | depart,roundabout turn straight exit-2,arrive |
+           | h,f       | gh,ef,ef | depart,roundabout turn right exit-1,arrive    |
```
MoKob: if we want to have it classify as a turn, this should not only be adjusted but moved to the intersection features.

features/guidance/roundabout.feature
```diff
@@ -22,19 +22,19 @@ Feature: Basic Roundabout
...
-           | d,f       | cd,ef,ef | depart,roundabout-exit-3,arrive |
-           | d,h       | cd,gh,gh | depart,roundabout-exit-2,arrive |
-           | d,a       | cd,ab,ab | depart,roundabout-exit-1,arrive |
-           | f,h       | ef,gh,gh | depart,roundabout-exit-3,arrive |
-           | f,a       | ef,ab,ab | depart,roundabout-exit-2,arrive |
-           | f,d       | ef,cd,cd | depart,roundabout-exit-1,arrive |
-           | h,a       | gh,ab,ab | depart,roundabout-exit-3,arrive |
-           | h,d       | gh,cd,cd | depart,roundabout-exit-2,arrive |
-           | h,f       | gh,ef,ef | depart,roundabout-exit-1,arrive |
+           | waypoints | route    | turns                                         |
```
MoKob: if kept, this also needs to move location.

features/guidance/roundabout.feature
```diff
@@ -295,9 +295,9 @@ Feature: Basic Roundabout
             | df    |            |
 
         When I route I should get
-            | waypoints | route    | turns                           |
-            | a,e       | ab,ce,ce | depart,roundabout-exit-1,arrive |
-            | a,f       | ab,df,df | depart,roundabout-exit-2,arrive |
+            | waypoints | route    | turns                                         |
```
MoKob: same as above

src/extractor/guidance/roundabout_handler.cpp
```diff
@@ -329,23 +326,13 @@ RoundaboutType RoundaboutHandler::getRoundaboutType(const NodeID nid) const
         if (1 == roundabout_name_ids.size() &&
             0 == connected_names.count(*roundabout_name_ids.begin()))
             return RoundaboutType::Rotary;
-        else
-            return RoundaboutType::Roundabout;
     }
 
-    if (radius <= MAX_ROUNDABOUT_INTERSECTION_RADIUS)
```
MoKob: I find it highly dangerous not to have a radius check in here. There are roundabouts of more than 2km in range (special case, though real). Announcing all of them as turns is very dangerous.

Also, if we have a long distance traveled in the roundabout, it is likely that `straight` is not obvious anymore.

daniel-j-h: Thanks for flagging this, I simply had no context why you added the limit there. Is this a data issue or a real intersection following the roundabout rules?

MoKob: In this case it is actually a incredibly large loop that really follows roundabout rules. Not exactly sure anymore were it was, but a friend of mine working at a different navigation company told me about this case.

TheMarex: In cases like this it would be helpful IMHO to add some context to the code. Link to OSM and a few words should suffices. Best would of course be a cucumber scenario testing for this behavior.

MoKob: There were a few cucumber cases testing roundabout sizes. They have been adjusted for this change (see above).

daniel-j-h:
Adopted code to comments: added upper limit for roundabout intersections, moved the cucumber tests into the roundabout turns file, created roundabout turn bike test. Rebased against master. Waiting for Travis, then ready to go.

PR #2720, opened 2016-07-30
Fix #2706 by using correct fallback u-turn

### LINKED ISSUE #2706: v5 demo server hits another lane assertion during extraction
Bad week for the demo server, it's currently hitting another one:

```
[assert] /mnt/ebs/osrm/2016-07-26-110507/src/extractor/guidance/turn_lane_augmentation.cpp:285
in: osrm::extractor::guidance::lanes::LaneDataVector osrm::extractor::guidance::lanes::handleNoneValueAtSimpleTurn(osrm::extractor::guidance::lanes::LaneDataVector, const Intersection&): connection_count + 1 == lane_data.size()
```

https://github.com/Project-OSRM/osrm-backend/blob/c7a1576100cb947af0e970644fb3f70dde63a9eb/src/extractor/guidance/turn_lane_augmentation.cpp#L283-L286

this is on the `5.3` branch.

oxidase:
Regression #2706  is due to a combination of 08248e38534eec1c4df629d5f12505d168297fde
and http://www.openstreetmap.org/changeset/40938983
where in ways http://www.openstreetmap.org/way/27292481
and http://www.openstreetmap.org/way/432488408
nodes
4315134884 (part of way 432488408)
4315134891 (part of way 432488408)
4315134886 (part of way 432488408)
form a u-turn that has index 0 after sorting and used as an allowed one
with a reversed edge.
A u-turn that corresponds to the condition uturn_could_be_valid == true has index 1
and ignored.

If the fix is ok, i will add a features test.
At the moment i tested only with `osmosis --read-pbf egypt-latest.osm.pbf --bounding-box left=29.775009 bottom=31.038227 right=29.921951 top=31.17991 --write-pbf egypt0.osm.pbf`

### DIFF
--- a/src/extractor/guidance/intersection_generator.cpp
+++ b/src/extractor/guidance/intersection_generator.cpp
@@ -143,7 +143,20 @@ Intersection IntersectionGenerator::getConnectedRoads(const NodeID from_node,
     const auto valid_count =
         boost::count_if(intersection, [](const ConnectedRoad &road) { return road.entry_allowed; });
     if (0 == valid_count && uturn_could_be_valid)
-        intersection[0].entry_allowed = true;
+    {
+        // after intersections sorting by angles, find the u-turn with (from_node == to_node)
+        // that was inserted together with setting uturn_could_be_valid flag
+        std::size_t self_u_turn = 0;
+        while (self_u_turn < intersection.size()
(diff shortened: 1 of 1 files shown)


TheMarex:
Great work. Was looking at the same code yesterday when debugging this. Seems like the possible fix. Going to merge and deploy this to the demo server to see if this work. :+1:

PR #2721, opened 2016-07-31
[Ready] Added test for issue #2706

oxidase:
Added a test that represents parts of  http://www.openstreetmap.org/way/27292481
and http://www.openstreetmap.org/way/432488408
 to cover https://github.com/Project-OSRM/osrm-backend/blob/master/src/extractor/guidance/intersection_generator.cpp#L154


daniel-j-h:
Thanks, looks good!

The Travis issue was accidentally introduced in https://github.com/Project-OSRM/osrm-backend/commit/eda6d9dc7c99fdaae69c7f68d294f84bc146f0d6, I'm already cherry-picking this fix into master.

TheMarex:
Looks good to merge, but needs a rebase. :+1:

oxidase:
Ok, rebased

Issue #2711, opened 2016-07-28
Handle HOV lane information


daniel-j-h:
http://wiki.openstreetmap.org/wiki/Key:hov

Especially HOV-only routed are a problem at the moment as they're faster and therefore always preferred in finding the shortest path.

cc @danpat

danpat:
Just to clarify the issue here:  road sections that are HOV-only, like this bypass:

http://www.openstreetmap.org/way/11198595

should probably be excluded from the default routing profile.  The profile should make enabling/disabling use of HOV-exclusive ways configurable, and default to "off".

daniel-j-h:
This already landed in https://github.com/Project-OSRM/osrm-backend/pull/2724 disabling hov-only ways by default. Closing here.

PR #2724, opened 2016-07-31
[Ready] Makes designated HOV way's routability configurable in profiles.

daniel-j-h:
Why only `hov=designated` and not all access tags, such as `hov:yes`,
`hov=no` and so on? From the Wiki:

- designated: The way is designated to high occupancy vehicles.
- yes: High occupancy vehicles are allowed. **This by itself does not imply that other vehicles are restricted from using the way.**
- no: High occupancy vehicles are not allowed on the way. **This by itself does not imply that other vehicle types are allowed to use it**.

The primary use-case is conditionally filtering ways such as:

http://www.openstreetmap.org/way/11198593#map=19/37.82571/-122.30521&layers=D

In addition there is a notion of HOV lanes for lane handling:

http://wiki.openstreetmap.org/wiki/Key:hov#hov:lanes.3D.2A

This changeset does not handle lanes at all, only designated HOV ways.
For HOV lane support, a logic similar to the lane access handling needs
to be implemented. This needs to go hand in hand with the existing lane
handling introduced in:

https://github.com/Project-OSRM/osrm-backend/commit/7d076e93445a4d016422a575f1132b267cd13c29

References:
- #2711
- http://wiki.openstreetmap.org/wiki/Key:access
- http://wiki.openstreetmap.org/wiki/Key:hov#Values

cc @willwhite @danpat

### DIFF
--- a/features/car/access.feature
+++ b/features/car/access.feature
@@ -5,7 +5,7 @@ Feature: Car - Restricted access
     Background:
         Given the profile "car"
 
-    Scenario: Car - Access tag hierarchy    on ways
+    Scenario: Car - Access tag hierarchy on ways
         Then routability should be
             | access | vehicle | motor_vehicle | motorcar | bothw |
             |        |         |               |          | x     |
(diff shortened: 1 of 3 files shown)


TheMarex:
Looks good to me.

willwhite:
🎉

Issue #2725, opened 2016-08-01
Unncessary turn instructions at Hermannstraße/Brücke


freenerd:
@MoKob This should be silent.

<img width="836" alt="screen shot 2016-08-01 at 15 27 33" src="https://cloud.githubusercontent.com/assets/25713/17295486/a6612dfc-57fc-11e6-9f11-6492f8762b48.png">

https://map.project-osrm.org/?z=18&center=52.466712%2C13.432256&loc=52.466161%2C13.432084&loc=52.468099%2C13.430915&hl=en&alt=0

Interestingly when coordinates are reversed, no instructions are emitted.

<img width="769" alt="screen shot 2016-08-01 at 15 29 05" src="https://cloud.githubusercontent.com/assets/25713/17295504/b705309a-57fc-11e6-93a0-7b3505b832a6.png">

https://map.project-osrm.org/?z=18&center=52.466712%2C13.432256&loc=52.466161%2C13.432084&loc=52.468099%2C13.430915&hl=en&alt=0

The fork leading up to the bridge seems about symmetrical, so not sure why different directions lead to different instructions.

daniel-j-h:
Thanks for flagging, thi is interesting not only because of the issues you mentioned but because of the following (notice the instructions):

![right](https://cloud.githubusercontent.com/assets/527241/17302508/d56902c6-581c-11e6-9001-949c7936505f.png)

![left](https://cloud.githubusercontent.com/assets/527241/17302509/d56eed62-581c-11e6-815b-b9ed93004aa3.png)

flipping from right to left --- looking into this now.

daniel-j-h:
- [Mapillary Trace](https://www.mapillary.com/app/?lat=52.46648333333334&lng=13.431908333333325&z=17&focus=photo&pKey=LWXnKqoGqUNLnG0lofiO0Q)
- [OSM data layer](http://www.openstreetmap.org/#map=19/52.46747/13.43158&layers=D)

daniel-j-h:
Response `type: depart, modifier: left`:

    http --body 'https://router.project-osrm.org/route/v1/driving/13.431537,52.466593;13.430915,52.468099?steps=true&alternatives=false

Response `type: depart, modifier: right`:

    http --body 'https://router.project-osrm.org/route/v1/driving/13.432143,52.466675;13.430915,52.468099?steps=true&alternatives=false

[...]

PR #2727, opened 2016-08-02
[ready - pending 2750] Properly Collapse Turns at Oneway/Bridge Scenario

daniel-j-h:
For https://github.com/Project-OSRM/osrm-backend/issues/2725.

- [Mapillary Trace](https://www.mapillary.com/app/?lat=52.46648333333334&lng=13.431908333333325&z=17&focus=photo&pKey=LWXnKqoGqUNLnG0lofiO0Q)
- [OSM data layer](http://www.openstreetmap.org/#map=19/52.46747/13.43158&layers=D)

```cucumber
 waypoints |     turns |     route
 |     a,f |     depart,arrive |     Hermannstr,Hermannstr | 

 | (-) f,a | (-) depart,arrive | (-) Hermannstr,Hermannstr | 
 | (+) f,a | (+) depart,turn right,new name slight left,arrive | (+) Hermannstr,Hermannbruecke,Hermannstr,Hermannstr | 

 | (-) y,f | (-) depart,arrive | (-) Hermannstr,Hermannstr | 
 | (+) y,f | (+) depart,turn slight left,turn right,arrive | (+) Hermannstr,Hermannbruecke,Hermannstr,Hermannstr | 

 | (-) f,y | (-) depart,arrive | (-) Hermannstr,Hermannstr | 
 | (+) f,y | (+) depart,turn right,new name slight left,arrive | (+) Hermannstr,Hermannbruecke,Hermannstr,Hermannstr | 
```

## Requirements
 - [x] create working rebase on #2750 
 - [x] merge of https://github.com/Project-OSRM/osrm-backend/pull/2750
 - [x] adjusted for comments

### DIFF
--- a/include/extractor/guidance/intersection_generator.hpp
+++ b/include/extractor/guidance/intersection_generator.hpp
@@ -48,11 +48,12 @@ class IntersectionGenerator
     // from `from_node` via `via_eid`
     // The resulting candidates have to be analysed for their actual instructions later on.
     OSRM_ATTR_WARN_UNUSED
-    Intersection getConnectedRoads(const NodeID from_node, const EdgeID via_eid) const;
+    Intersection GetConnectedRoads(const NodeID from_node, const EdgeID via_eid) const;
 
     // check if two indices in an intersection can be seen as a single road in the perceived
     // intersection representation
-    bool canMerge(const Intersection &intersection,
+    bool CanMerge(const NodeID intersection_node,
+                  const Intersection &intersection,
                   std::size_t first_index,
                   std::size_t second_index) const;
 
@@ -68,7 +69,8 @@ class IntersectionGenerator
     // The treatment results in a straight turn angle of 180º rather than a turn angle of approx
     // 160
     OSRM_ATTR_WARN_UNUSED
-    Intersection mergeSegregatedRoads(Intersection intersection) const;
+    Intersection MergeSegregatedRoads(const NodeID intersection_node,
+                                      Intersection intersection) const;
 
     // The counterpiece to mergeSegregatedRoads. While we can adjust roads that split up at the
     // intersection itself, it can also happen that intersections are connected to joining roads.
@@ -81,8 +83,20 @@ class IntersectionGenerator
     //       b
     //
     // for the local view of b at a.
-    Intersection adjustForJoiningRoads(const NodeID node_at_intersection,
+    OSRM_ATTR_WARN_UNUSED
+    Intersection AdjustForJoiningRoads(const NodeID node_at_intersection,
                                        Intersection intersection) const;
+
+    // Graph Compression cannot compress every setting. For example any barrier/traffic light cannot
+    // be compressed. As a result, a simple road of the form `a ----- b` might end up as having an
+    // intermediate intersection, if there is a traffic light in between. If we want to look farther
+    // down a road, finding the next actual decision requires the look at multiple intersections.
+    // Here we follow the road until we either reach a dead end or find the next intersection with
+    // more than a single next road.
+    inline Intersection GetActualNextIntersection(const NodeID starting_node,
+                                                  const EdgeID via_edge,
+                                                  NodeID *resulting_from_node,
+                                                  EdgeID *resulting_via_edge) const;
 };
 
 } // namespace guidance
--- a/include/extractor/guidance/intersection_generator.hpp
+++ b/include/extractor/guidance/intersection_generator.hpp
@@ -62,7 +62,8 @@ class IntersectionGenerator
     Intersection GetConnectedRoads(const NodeID from_node, const EdgeID via_eid) const;
 
     // check if two indices in an intersection can be seen as a single road in the perceived
-    // intersection representation
+    // intersection representation See below for an example. Utility function for
+    // MergeSegregatedRoads
     bool CanMerge(const NodeID intersection_node,
                   const Intersection &intersection,
                   std::size_t first_index,
--- a/include/extractor/guidance/intersection_generator.hpp
+++ b/include/extractor/guidance/intersection_generator.hpp
@@ -35,6 +35,17 @@ class IntersectionGenerator
 
     Intersection operator()(const NodeID nid, const EdgeID via_eid) const;
 
+    // Graph Compression cannot compress every setting. For example any barrier/traffic light cannot
+    // be compressed. As a result, a simple road of the form `a ----- b` might end up as having an
+    // intermediate intersection, if there is a traffic light in between. If we want to look farther
+    // down a road, finding the next actual decision requires the look at multiple intersections.
+    // Here we follow the road until we either reach a dead end or find the next intersection with
+    // more than a single next road.
+    Intersection GetActualNextIntersection(const NodeID starting_node,
+                                           const EdgeID via_edge,
+                                           NodeID *resulting_from_node,
+                                           EdgeID *resulting_via_edge) const;
+
   private:
     const util::NodeBasedDynamicGraph &node_based_graph;
     const RestrictionMap &restriction_map;
@@ -86,17 +97,6 @@ class IntersectionGenerator
     OSRM_ATTR_WARN_UNUSED
     Intersection AdjustForJoiningRoads(const NodeID node_at_intersection,
                                        Intersection intersection) const;
-
-    // Graph Compression cannot compress every setting. For example any barrier/traffic light cannot
-    // be compressed. As a result, a simple road of the form `a ----- b` might end up as having an
-    // intermediate intersection, if there is a traffic light in between. If we want to look farther
-    // down a road, finding the next actual decision requires the look at multiple intersections.
-    // Here we follow the road until we either reach a dead end or find the next intersection with
-    // more than a single next road.
-    inline Intersection GetActualNextIntersection(const NodeID starting_node,
-                                                  const EdgeID via_edge,
-                                                  NodeID *resulting_from_node,
-                                                  EdgeID *resulting_via_edge) const;
 };
 
 } // namespace guidance
--- a/include/extractor/guidance/intersection_generator.hpp
+++ b/include/extractor/guidance/intersection_generator.hpp
@@ -50,6 +50,12 @@ class IntersectionGenerator
     OSRM_ATTR_WARN_UNUSED
     Intersection getConnectedRoads(const NodeID from_node, const EdgeID via_eid) const;
 
+    // check if two indices in an intersection can be seen as a single road in the perceived
+    // intersection representation
+    bool canMerge(const Intersection &intersection,
+                  std::size_t first_index,
+                  std::size_t second_index) const;
+
     // Merge segregated roads to omit invalid turns in favor of treating segregated roads as
     // one.
     // This function combines roads the following way:
@@ -63,6 +69,20 @@ class IntersectionGenerator
     // 160
     OSRM_ATTR_WARN_UNUSED
     Intersection mergeSegregatedRoads(Intersection intersection) const;
+
+    // The counterpiece to mergeSegregatedRoads. While we can adjust roads that split up at the
+    // intersection itself, it can also happen that intersections are connected to joining roads.
+    //
+    //     *                           *
+    //     *        is converted to    *
+    //   v   a ---                     a ---
+    //   v   ^                         +
+    //   v   ^                         +
+    //       b
+    //
+    // for the local view of b at a.
+    Intersection adjustForJoiningRoads(const NodeID node_at_intersection,
+                                       Intersection intersection) const;
 };
 
 } // namespace guidance
--- a/include/util/node_based_graph.hpp
+++ b/include/util/node_based_graph.hpp
@@ -51,7 +51,7 @@ struct NodeBasedEdgeData
     LaneDescriptionID lane_description_id;
     extractor::guidance::RoadClassification road_classification;
 
-    bool IsCompatibleToExceptForName(const NodeBasedEdgeData &other) const
+    bool IsCompatibleTo(const NodeBasedEdgeData &other) const
     {
         return (reversed == other.reversed) &&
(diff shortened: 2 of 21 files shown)


MoKob:
I've added some logic to detect these name-changes (luckily that part was simple). Whats more problematic are the intersections as we can see at the top of one testcase [here](https://github.com/Project-OSRM/osrm-backend/pull/2727/files#diff-cada8e09cfad24a11d05ae0ea67479f5R985).

The angles at that turn are very large (90 degree between the two joining roads).
Merging these kind of turns is possible. I have to try and see if there are reasonable cases that will merge even though they shouldn't.

Pending verification:
 - [ ] verify that this change does not break other intersections

include/extractor/guidance/intersection_generator.hpp
```diff
@@ -48,7 +59,14 @@ class IntersectionGenerator
     // from `from_node` via `via_eid`
     // The resulting candidates have to be analysed for their actual instructions later on.
     OSRM_ATTR_WARN_UNUSED
-    Intersection getConnectedRoads(const NodeID from_node, const EdgeID via_eid) const;
+    Intersection GetConnectedRoads(const NodeID from_node, const EdgeID via_eid) const;
+
+    // check if two indices in an intersection can be seen as a single road in the perceived
+    // intersection representation
```
TheMarex: Can you add an example here? Small ASCII pictogram should suffice.

include/util/node_based_graph.hpp
```diff
@@ -51,14 +51,19 @@ struct NodeBasedEdgeData
...
     {
-        return (name_id == other.name_id) && (reversed == other.reversed) &&
+        return (reversed == other.reversed) &&
                (roundabout == other.roundabout) && (startpoint == other.startpoint) &&
                (access_restricted == other.access_restricted) &&
                (travel_mode == other.travel_mode) &&
                (road_classification == other.road_classification);
     }
+
+    bool IsCompatibleTo(const NodeBasedEdgeData &other) const
```
TheMarex: Maybe we can find better names here reflecting where each is used. I would imagine something like:

- `IsCompressableWith` (currently `IsCompatibleTo`) since it is used by the graph compressor
- `IsCompatibleTo` (currently `IsCompatibleToExceptForName`) since it is used in the guidance code

src/engine/guidance/post_processing.cpp
```diff
@@ -815,6 +815,20 @@ std::vector<RouteStep> collapseTurns(std::vector<RouteStep> steps)
                     invalidateStep(steps[step_index]);
                 }
             }
+            else if (step_index + 2 < steps.size() &&
+                     current_step.maneuver.instruction.type == TurnType::NewName &&
+                     steps[step_index + 1].maneuver.instruction.type == TurnType::NewName &&
```
TheMarex: can you name `steps[step_index + 1]`? I think it is heard to see the realtion between `current_step` and `one_step_back` with this one.

src/extractor/guidance/intersection_generator.cpp
```diff
@@ -158,7 +162,134 @@ Intersection IntersectionGenerator::getConnectedRoads(const NodeID from_node,
...
+bool IntersectionGenerator::CanMerge(const NodeID node_at_intersection,
+                                     const Intersection &intersection,
+                                     std::size_t first_index,
+                                     std::size_t second_index) const
+{
+    const auto &first_data = node_based_graph.GetEdgeData(intersection[first_index].turn.eid);
+    const auto &second_data = node_based_graph.GetEdgeData(intersection[second_index].turn.eid);
+
+    // only merge named ids
+    if (first_data.name_id == EMPTY_NAMEID)
```
TheMarex: Hrm was your intention here to use the IsCompatibleTo function?

MoKob: Not entirely. This has been reworked from what it has been before. We could introduce a second function that checks for the same name and excludes roundabout tags.

This check is neither the compressable/compatible part.

src/extractor/guidance/intersection_generator.cpp
```diff
@@ -158,7 +162,134 @@ Intersection IntersectionGenerator::getConnectedRoads(const NodeID from_node,
...
+    };
+
+    // Only merge valid y-arms
+    if (!isValidYArm(first_index, second_index) || !isValidYArm(second_index, first_index))
+        return false;
+
+    if (angle_between < 60)
+        return true;
+
+    // return false;
```
TheMarex: Why is this commented out?

src/extractor/guidance/intersection_generator.cpp
```diff
@@ -158,7 +162,134 @@ Intersection IntersectionGenerator::getConnectedRoads(const NodeID from_node,
         intersection[self_u_turn].entry_allowed = true;
     }
 
-    return mergeSegregatedRoads(std::move(intersection));
+    return intersection;
+}
+
+bool IntersectionGenerator::CanMerge(const NodeID node_at_intersection,
```
TheMarex: This function is missing a comment that outlines which cases it targets.

src/extractor/guidance/intersection_generator.cpp
MoKob: left-over code from debugging. will remove.

src/extractor/guidance/intersection_generator.cpp
```diff
@@ -158,7 +162,134 @@ Intersection IntersectionGenerator::getConnectedRoads(const NodeID from_node,
...
+    // Only merge valid y-arms
+    if (!isValidYArm(first_index, second_index) || !isValidYArm(second_index, first_index))
+        return false;
+
+    if (angle_between < 60)
+        return true;
+
+    // return false;
+    // Finally, we also allow merging if all streets offer the same name, it is only three roads and
+    // the angle is not fully extreme:
```
TheMarex: What does "not fully extreme" mean?

src/extractor/guidance/intersection_generator.cpp
```diff
@@ -158,7 +162,134 @@ Intersection IntersectionGenerator::getConnectedRoads(const NodeID from_node,
...
+    if (intersection.size() != 3)
+        return false;
+
+    const std::size_t missing_index = [first_index, second_index]() {
```
TheMarex: What is the `missing_index`?

src/extractor/guidance/intersection_generator.cpp
MoKob: See https://github.com/Project-OSRM/osrm-backend/blob/90cc782b4581f25f3350f5c0df91c3de45e59625/include/extractor/guidance/intersection_generator.hpp#L64-L81. Can move the explanation here, if you feel this is the better location.

src/extractor/guidance/intersection_generator.cpp
```diff
@@ -158,7 +162,134 @@ Intersection IntersectionGenerator::getConnectedRoads(const NodeID from_node,
...
+    const auto y_angle_difference =
+        angularDeviation(angularDeviation(intersection[missing_index].turn.angle,
+                                          intersection[first_index].turn.angle),
+                         angularDeviation(intersection[missing_index].turn.angle,
+                                          intersection[second_index].turn.angle));
+
+    // Allow larger angles if its three roads only of the same name
+    const bool could_be_valid_y_intersection =
+        angle_between < 100 && y_angle_difference < FUZZY_ANGLE_DIFFERENCE;
+    return could_be_valid_y_intersection;
```
TheMarex: Why `could_`? This implies that we need some further checks to make this assumptions. Where are they done?

src/extractor/guidance/intersection_generator.cpp
MoKob: https://github.com/Project-OSRM/osrm-backend/pull/2727/files/90cc782b4581f25f3350f5c0df91c3de45e59625#diff-5ba817e16c4faf2ff01b312c1802f695R291

src/extractor/guidance/intersection_generator.cpp
```diff
@@ -303,6 +421,110 @@ Intersection IntersectionGenerator::mergeSegregatedRoads(Intersection intersecti
...
+        auto &road = intersection[index];
+        // to find out about the above situation, we need to look at the next intersection (at d in
+        // the example). If the initial road can be merged to the left/right, we are about to adjust
+        // the angle.
+        const auto next_intersection_along_road =
+            GetConnectedRoads(node_at_intersection, road.turn.eid);
+        if (next_intersection_along_road.size() <= 1)
+            continue;
+
+        const auto adjustAngle = [](double angle, double offset) {
```
TheMarex: This could be a general helper function for angles. `normalizeAngle(double angle)` that normalizes to `[0, 359]`. `normalizeAngle(angle + offset)` would do the same as the function here.

src/extractor/guidance/intersection_generator.cpp
```diff
@@ -303,6 +421,110 @@ Intersection IntersectionGenerator::mergeSegregatedRoads(Intersection intersecti
...
+            angle += offset;
+            if (angle > 360)
+                return angle - 360.;
+            else if (angle < 0)
+                return angle + 360.;
+            return angle;
+        };
+
+        if (CanMerge(node_based_graph.GetTarget(road.turn.eid), next_intersection_along_road, 0, 1))
```
TheMarex: Hrm I think it would help to connect this to the example above as well. What corresponds to `node_based_graph.GetTarget(road.turn.eid)` ?

src/extractor/guidance/intersection_generator.cpp
```diff
@@ -303,6 +421,110 @@ Intersection IntersectionGenerator::mergeSegregatedRoads(Intersection intersecti
...
+
+Intersection
+IntersectionGenerator::GetActualNextIntersection(const NodeID starting_node,
+                                                 const EdgeID via_edge,
+                                                 NodeID *resulting_from_node = nullptr,
+                                                 EdgeID *resulting_via_edge = nullptr) const
+{
+    // This function skips over traffic lights/graph compression issues and similar to find the next
+    // actual intersection
+    Intersection potential_result = GetConnectedRoads(starting_node, via_edge);
```
TheMarex: Same language thing here `potential_` implies we are not yet done with checking if it really is the next intersection. `next_intersection` would make more sense in the context of the function name.

src/extractor/guidance/intersection_handler.cpp
```diff
@@ -375,19 +391,29 @@ std::size_t IntersectionHandler::findObviousTurn(const EdgeID via_edge,
 
     const EdgeData &in_data = node_based_graph.GetEdgeData(via_edge);
     const auto in_classification = in_data.road_classification;
+
+    const auto obvious_by_road_class = [](const RoadClassification in_classification,
+                                          const RoadClassification obvious_candidate,
+                                          const RoadClassification compare_candidate) {
+        const bool has_high_priority =
+            2 * obvious_candidate.GetPriority() < compare_candidate.GetPriority();
```
TheMarex: This assumes a certain structure in road priorities. Can we make `2` a constant and document this relationship in the road priority related code?

src/extractor/guidance/intersection_handler.cpp
```diff
@@ -375,19 +391,29 @@ std::size_t IntersectionHandler::findObviousTurn(const EdgeID via_edge,
 
     const EdgeData &in_data = node_based_graph.GetEdgeData(via_edge);
     const auto in_classification = in_data.road_classification;
+
+    const auto obvious_by_road_class = [](const RoadClassification in_classification,
```
TheMarex: This is a pretty long lambda function that doesn't actually capture any values. Could this be a member function?

TheMarex: Ah got the scope wrong. Not that long actually.

src/extractor/guidance/intersection_handler.cpp
```diff
@@ -341,15 +344,28 @@ bool IntersectionHandler::isThroughStreet(const std::size_t index,
 {
```
TheMarex: This function could also use a small explanation. E.G. compares multiple roads at the intersection and determines if the `index` would be considered the best candidate for a `through` street.

src/extractor/guidance/intersection_handler.cpp
```diff
@@ -399,18 +425,52 @@ std::size_t IntersectionHandler::findObviousTurn(const EdgeID via_edge,
...
+    }();
+
+    const auto &best_data = node_based_graph.GetEdgeData(intersection[best].turn.eid);
+    // has no obvious continued road, or the best continuation is really bad compared to the
+    // best-one
+    if (best_continue == 0 ||
+        (num_continue_names > 2 && best_continue_deviation >= 2 * NARROW_TURN_ANGLE) ||
+        (best_deviation < FUZZY_ANGLE_DIFFERENCE &&
+         !best_data.road_classification.IsRampClass())) // ||
+    //        (node_based_graph.GetEdgeData(intersection[best_continue].turn.eid).road_classification
```
TheMarex: Why is this commented out?

src/extractor/guidance/intersection_handler.cpp
```diff
@@ -375,19 +391,29 @@ std::size_t IntersectionHandler::findObviousTurn(const EdgeID via_edge,
```
TheMarex: This function has grown in size dramatically, do you think there is a way to split the different cases out?

TheMarex:
Added some comment mostly regarding structure, naming and documentation. This needs a rebase against master and is good to go once the comments are addressed.

src/extractor/guidance/intersection_generator.cpp
TheMarex: Great. Yeah I think the docs make the most sense where they are implemented, since this is all internal API (and you end up reading the implementation more often then not).

MoKob:
From discussion in chat / voice (if travis approves) this would be ready for merge after #2750 is merged.

Issue #2675, opened 2016-07-19
Move Turn Lanes into Intersections


MoKob:
After some internal discussion, we decided to move the turn lanes to the intersection object associated with every route segment.

This allows both the announcement of turn lanes for `maneuvers` as well as a visual indication for integrated guidance systems.

The API format would look like:

```json
{
 "distance":152.3,
 "duration":15.6,
 "name":"Lortzingstraße",
 "maneuver":{
     "type":"turn",
     "modifier":"right",
     },
 "geometry":"{lu_IypwpAVrAvAdI",
 "mode":"driving",
 "intersections":[
    {"location":[13.39677,52.54366],
    "in":3,
    "out":1,
    "bearings":[10,92,184,270],
    "entry":[false,"true","true"],
    "lanes":[
         {"indications":["left","straight"], "valid":"false"},
         {"indications":["right"], "valid":"true"}
    ]},
    {"location":[13.394718,52.543096],
    "in":0,
    "out":2,
    "bearings":[60,150,240,330],
    "entry":["false","true","true","true"],
    "lanes":[
         {"indications":["straight"], "valid":"true"},
         {"indications":["right"], "valid":"false"}
     ]}
]}
```

In this format, a turn has turn-lanes if the very first intersection in this list offers turn lanes.

The type `use lane` will persist for maneuvers that require an actual action of the driver. An example would be the restriction to one lane out of a larger subset of lanes.

Did I capture this correctly?

/cc @TheMarex @freenerd @daniel-j-h

MoKob:
Required to decide when to emit use-lane: https://github.com/Project-OSRM/osrm-backend/pull/2666

freenerd:
What is the design if an intersection does not have lane information? Will there be an empty object or a `null` in the response?

Is there a guarantee that the number of intersections and the number of lanes is always the same? What happens if there is no lane information at all or if only the first intersection has lane information, but the following don't?

TheMarex:
> Is there a guarantee that the number of intersections and the number of lanes is always the same? What happens if there is no lane information at all or if only the first intersection has lane information, but the following don't?

- If there are no lanes, the `lanes` property will be `undefined`
- Each `Intersection` object either has a `lanes` property or not. they only describe the lanes _at the intersection_ hence there is no semantic link of lane information between intersections.

freenerd:
[...]

PR #2728, opened 2016-08-02
[Ready] Adapt lane dump to lanes at intersection (#2675), resolves #2709

### LINKED ISSUE #2709: Adapt debug helpers to new lane representation
`include/util/debug.hpp` does not compile at the moment since it's expecting the old lanes-on-maneuver representation.

The only reason no one noticed this is it's not included in any file for releases. Adapt this and maybe include the header somewhere to catch this not compiling at all...

daniel-j-h:
Makes the lane dumping code compile again for debugging purposes.

Let's keep the include at least in one cpp file to get warnings when things break again.

### DIFF
--- a/include/util/debug.hpp
+++ b/include/util/debug.hpp
@@ -20,11 +20,13 @@ namespace guidance
 {
 inline void print(const engine::guidance::RouteStep &step)
 {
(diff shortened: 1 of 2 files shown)


TheMarex:
Looks good. :+1:

PR #2729, opened 2016-08-02
[Ready] Warn on unused return values Guidance, resolves #2686.

### LINKED ISSUE #2686: Warn on unused return value in Guidance code prone to misuse
https://github.com/Project-OSRM/osrm-backend/pull/2685/files fixes an issue where we did

    elongate(fstStep, sndStep);

instead of

    newStep = elongate(fstStep, sndStep);

we didn't get any warnings, and from what I can see the only way to trigger a warning would be to use

```cpp
__attribute__((warn_unused_result))
```

All our guidance code us using this paradigm and is prone to this issue.

- [x] add a macro for this attribute on compiler supporting this
- [x] tag important guidance function decls with it

daniel-j-h:
https://github.com/Project-OSRM/osrm-backend/pull/2685/files

fixes an issue where we did

    elongate(fstStep, sndStep);

instead of

    newStep = elongate(fstStep, sndStep);

we didn't get any warnings.

The only way to trigger a warning here is to use

```cpp
__attribute__((warn_unused_result))
```

This changeset does exactly that: for the new guidance code prone to
these kind of issue we add such an attribute to the declaration.

### DIFF
--- a/include/engine/guidance/lane_processing.hpp
+++ b/include/engine/guidance/lane_processing.hpp
@@ -4,6 +4,7 @@
 #include <vector>
 
 #include "engine/guidance/route_step.hpp"
+#include "util/attributes.hpp"
 
 namespace osrm
 {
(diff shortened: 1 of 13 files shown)


TheMarex:
Seems fine to me. MacOS timed out because of cache stuff probably.

PR #2731, opened 2016-08-02
[Ready] Fixes components tool try-catch asymmetry

daniel-j-h:
no try-without a catch


daniel-j-h:
We fixed it already on master, my branch was behind. Sorry.

PR #2733, opened 2016-08-02
Remove unused variable node_info_list

gavinsherry:
Small tidy up of a nearby comment with extraneous white space included.

### DIFF
--- a/include/extractor/guidance/turn_lane_handler.hpp
+++ b/include/extractor/guidance/turn_lane_handler.hpp
@@ -26,7 +26,7 @@ namespace extractor
 namespace guidance
 {
 
-// Given an Intersection, the graph to access the data and  the turn lanes, the turn lane matcher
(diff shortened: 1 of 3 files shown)


daniel-j-h:
Thank you. I'm not sure what @MoKob's intentions or future plans are here, but usually all our handler sooner or later need some spacial clues. For now you're right this is unused.

TheMarex:
This commit is fairly atomic and can be reverted if needed so I'm :+1: with the cleanup here. Thanks!

MoKob:
The intention here is that I need this variable in the process whenever I want to inspect lane cases. Whenever I put in a new case, I want to have a look at the location. Having it in is essential for a simple process of writing new lane cases.

It is a major pain to get in in for every situation independently. I know it's not very clean, but kind of required while I work on this.

PR #2737, opened 2016-08-03
Fixing up the release flow docs

freenerd:
We've been bitten by changes introduced between a release candidate and the final release. I wanted to add a constraint that all minor/major releases need to have had a code-equal release candidate before. When reading the document, I saw more room for clarification so did that as well.

Some open questions:

- I've done away with the `vx.y` branches that can have patch releases. Instead it should be possible to use the lastest `vx.y.z` branch, is that correct?
- Shall/Can we pin the demo server to a specific release candidate before a release? I'm assuming the case where `master` is already continuing to move while we are still in the process of running QA for a release candidate.

@TheMarex I might be stepping on your toes here. Let's discuss and or change this PR

/cc @daniel-j-h @danpat @MoKob

### DIFF
--- a/docs/releasing.md
+++ b/docs/releasing.md
@@ -39,7 +39,7 @@ We may introduce forward-compatible changes: query parameters and response prope
 
 ## Releasing a version
 
-1. Check out the appropriate release branch `x.y.z`
+1. Check out the appropriate release branch `x.y`
 2. Make sure all tests are passing (e.g. Travis CI gives you a :thumbs_up:)
 3. Make sure `CHANGELOG.md` is up to date.
 4. Make sure the OSRM version in `CMakeLists.txt` is up to date
--- a/docs/releasing.md
+++ b/docs/releasing.md
@@ -31,7 +31,7 @@ We may introduce forward-compatible changes: query parameters and response prope
 ## Release and branch management
 
 - The `master` branch is for the bleeding edge development
-- We create and maintain release branches to control the release flow
+- We create and maintain release branches `x.y` to control the release flow
 - No minor or major version will be released without a code-equal release candidates
 - For quality assurance, release candidates will be run on the demo server for 24 hours before releaseing the version proper
 - Patch versions may be released without a release candidate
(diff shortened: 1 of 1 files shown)


docs/releasing.md
```diff
@@ -1,50 +1,49 @@
...
+1. Check out the appropriate release branch `x.y.z`
+2. Make sure all tests are passing (e.g. Travis CI gives you a :thumbs_up:)
+3. Make sure `CHANGELOG.md` is up to date.
 4. Make sure the OSRM version in `CMakeLists.txt` is up to date
 5. Use an annotated tag to mark the release: `git tag vx.y.z -a` Body of the tag description should be the changelog entries.
 6. Push tags and commits: `git push; git push --tags`
-7. Proceede with the `node-osrm` release as outlined in the repository.
-8. Write a mailing-list post to osrm-talk@openstreetmap.org to announce the release
-
+8. Proceede with the `node-osrm` release as [outlined in the repository](https://github.com/Project-OSRM/node-osrm/blob/master/docs/releasing.md).
```
TheMarex: Github supports relative paths `releasing.md` should work.

docs/releasing.md
```diff
@@ -1,50 +1,49 @@
...
+- For quality assurance, release candidates will be run on the demo server for 24 hours before releaseing the version proper
+- Patch versions may be released without a release candidate
+- We may backport fixes to older versions and release them as patch versions
 
-1. Check out the appropriate release branch x.y
-2. Make sure all fixes are listed in the changelog and included in the branch
-3. Make sure all tests are passing (e.g. Travis CI gives you a :thumbs_up:)
+## Releasing a version
+
+1. Check out the appropriate release branch `x.y.z`
```
TheMarex: We always have a release branch for each series:
`5.1`, `5.2`, `5.3` etc. having an own branch for each patch doesn't really make sense because we only do linear changes to each. For example:

```
                                                     v5.2.1                                      v5.2.2
                                                        v                                               v
5.2 :- ----- commit --- commit --- commit --- commit -- commit -- commit
```

We always guarantee this linearity and it is easy to control because we typically only cherry-pick changes from master and edit the `CHANGLOG.md`.

PR #2738, opened 2016-08-03
[Ready] Collapse Staggered Intersections.

daniel-j-h:
Staggered intersection are very short zig-zags of only a few meters,
where the street name changes only temporarily.

They are common in rural and exurban areas, especially in the US.
(In addition, these cases could as well be tagging issues)

We do not want to announce these short left-rights or right-lefts:

          * -> b      a -> *
          |       or       |       becomes  a   ->   b
     a -> *                * -> b

Here is one example:

- https://www.openstreetmap.org/edit#map=20/39.26017/-84.25182

And here are two edge-cases that we don't handle at the moment:

- http://www.openstreetmap.org/edit#map=20/38.87900/-76.98519
- http://www.openstreetmap.org/edit#map=19/45.51056/-122.63462

and probably should not handle since the distance in between is
quite long (roughly 7-15 meters). For these we want to announce
two turns to not confuse the user.

Thanks to @1ec5 for raising this issue and @karenzshea for
providing additional US examples and cultural insights.

### DIFF
--- a/features/guidance/staggered-intersections.feature
+++ b/features/guidance/staggered-intersections.feature
@@ -1,0 +1,92 @@
+@routing  @guidance @staggered-intersections
+Feature: Staggered Intersections
+
+    Background:
+        Given the profile "car"
+        Given a grid size of 1 meters
+        # Note the one meter grid size: staggered intersections make zig-zags of a couple of meters only
+
+    # https://www.openstreetmap.org/#map=19/39.26022/-84.25144
+    Scenario: Staggered Intersection: Oak St, Cedar Dr
+        Given the node map
+            |   |   | j |   |   |
+            | a | b | c |   |   |
+            |   |   | d |   |   |
+            |   |   | e | f | g |
+            |   |   | h |   |   |
+            |   |   | i |   |   |
+
+        And the ways
+            | nodes  | highway     | name     |
+            | abc    | residential | Oak St   |
+            | efg    | residential | Oak St   |
+            | jcdehi | residential | Cedar Dr |
+
+        When I route I should get
+            | waypoints | route         | turns |
+            | a,g       | Oak St,Oak St | depart,arrive |
+            | g,a       | Oak St,Oak St | depart,arrive |
+
+    Scenario: Staggered Intersection: do not collapse if long segment in between
+        Given the node map
+            |   |   | j |   |   |
+            | a | b | c |   |   |
+            |   |   |   |   |   |
+            |   |   |   |   |   |
+            |   |   | d |   |   |
+            |   |   |   |   |   |
+            |   |   |   |   |   |
+            |   |   | e | f | g |
+            |   |   | h |   |   |
+            |   |   | i |   |   |
+
+        And the ways
+            | nodes  | highway     | name     |
+            | abc    | residential | Oak St   |
+            | efg    | residential | Oak St   |
+            | jcdehi | residential | Cedar Dr |
+
+        When I route I should get
+            | waypoints | route                         | turns                              |
+            | a,g       | Oak St,Cedar Dr,Oak St,Oak St | depart,turn right,turn left,arrive |
+            | g,a       | Oak St,Cedar Dr,Oak St,Oak St | depart,turn right,turn left,arrive |
+
+    Scenario: Staggered Intersection: do not collapse if not left-right or right-left
+        Given the node map
+            |   |   | j |   |   |
+            | a | b | c |   |   |
+            |   |   | d |   |   |
+            | g | f | e |   |   |
+            |   |   | h |   |   |
+            |   |   | i |   |   |
+
+        And the ways
+            | nodes  | highway     | name     |
+            | abc    | residential | Oak St   |
+            | efg    | residential | Oak St   |
+            | jcdehi | residential | Cedar Dr |
+
+        When I route I should get
+            | waypoints | route                | turns                        |
+            | a,g       | Oak St,Oak St,Oak St | depart,continue uturn,arrive |
+            | g,a       | Oak St,Oak St,Oak St | depart,continue uturn,arrive |
+
+    Scenario: Staggered Intersection: do not collapse if the names are not the same
+        Given the node map
+            |   |   | j |   |   |
+            | a | b | c |   |   |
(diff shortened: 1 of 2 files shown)


daniel-j-h:
In theory the threshold should probably be based on the street's width, classification, .. practically speaking we don't have that information and a simple approach has to suffice.

We _could_ maybe make it configurable in the profiles on a global basis if needs be.

src/engine/guidance/post_processing.cpp
```diff
@@ -513,6 +514,33 @@ void collapseTurnAt(std::vector<RouteStep> &steps,
...
+// We do not want to announce these short left-rights or right-lefts:
+//
+//      * -> b      a -> *
+//      |       or       |       becomes  a   ->   b
+// a -> *                * -> b
+//
+bool isStaggeredIntersection(const RouteStep &previous, const RouteStep &current)
+{
+    // Base decision on distance since the zig-zag is a visual clue.
+    const constexpr auto MAX_STAGGERED_DISTANCE = 3; // debatable, but keep short to be on safe side
```
freenerd: I assume this distance is in meters, should say so

emiltin:
This will be very welcome for bike routing, where intersections are often modelled in a way that mean you do several minor turns.

But he situation is often more complicated that two turns close to 90 degree. E.g. consider this situation, if you have to go from Fredens Bro southwest-wards to Sølvgade - you follow the blue bike paths and have multiple small turns:

http://www.openstreetmap.org/#map=19/55.69090/12.57094

In any case I think distances etc. be configurable in the profiles. Cars, bikes or walking will need different distances/settings.

What about instructions like 'Turn right, then left'? Some nav systems use this to handle staggered intersections with intermediate distance betweeen.

daniel-j-h:
> What about instructions like 'Turn right, then left'? Some nav systems use this to handle staggered intersections with intermediate distance betweeen.

We can't combine instructions in the backend in the sense that a quick `Left` followed by a quick `Right` turns becomes `LeftThanRight`. This is something a client can and has to do anyway for announcing maneuvers at the right time based on the primitives we're offering.

That's one of the reasons the threshold should be kept small, to be on the safe side and to collapse only if we're sure the primitive we're getting out of it still makes sense.

features/guidance/staggered-intersections.feature
```diff
@@ -0,0 +1,72 @@
...
+            | a | b | c |   |   |
+            |   |   | d |   |   |
+            |   |   | e | f | g |
+            |   |   | h |   |   |
+            |   |   | i |   |   |
+
+        And the ways
+            | nodes  | highway     | name     |
+            | abc    | residential | Oak St   |
+            | efg    | residential | Oak St   |
```
TheMarex: Can you add a negative test with `efg` not having the `Oak St` name? Like:

```
|abc|Oak Str|
|efg|Elm Str|
|jcdehi|Cedar Dr|
```

In that case the expected behavior would still be right-left/left-right turns, right?

daniel-j-h: Yes, this hooks into the pre-existing name-oscillation logic.

Added a test for this: https://github.com/Project-OSRM/osrm-backend/pull/2738/commits/b83151a1939fc5374fc18501864dde344f6ea698#diff-3e526a8f2e6b88dc336b8b12f7ca5f52R74

TheMarex:
I used a small script to calculate all the occurrences of these staggered intersections that wold be hit by this change:

https://gist.github.com/TheMarex/a62d09dc5b8c094baa6b203d805278a9

Producing the following geojson for California:

http://bl.ocks.org/d/5b44f131d86f3710d3b0e20b210c1798

There are not a lot of cases here for a 3 meters threshold, but all I examined seemed like they would profit from this change.

:+1: from me, looks good to merge.

zgz757183190:
hey ,man . can you help me ? 

Linking CXX static library libinnobase_embedded.a
[ 40%] Built target innobase_embedded
Scanning dependencies of target blackhole
[ 40%] Building CXX object storage/blackhole/CMakeFiles/blackhole.dir/ha_blackhole.cc.o
In file included from /usr/local/mysql-5.6.12/sql/sql_plugin.h:34:0,
                 from /usr/local/mysql-5.6.12/sql/structs.h:23,
                 from /usr/local/mysql-5.6.12/sql/unireg.h:170,
                 from /usr/local/mysql-5.6.12/storage/blackhole/ha_blackhole.cc:19:
/usr/local/mysql-5.6.12/sql/field.h: 在成员函数‘void Field_blob::store_ptr_and_length(const char*, uint32)’中:
/usr/local/mysql-5.6.12/sql/field.h:3268:50: 警告：argument to ‘sizeof’ in ‘void* memmove(void*, const void*, size_t)’ call is the same pointer type ‘char*’ as the source; expected ‘char’ or an explicit length [-Wsizeof-pointer-memaccess]
     bmove(ptr + packlength, (char*) &from, sizeof(char *));
                                                  ^
/usr/local/mysql-5.6.12/include/m_string.h:52:45: 附注：in definition of macro ‘bmove’
 # define bmove(d, s, n)  memmove((d), (s), (n))
                                             ^
Linking CXX static library libblackhole.a
[ 40%] Built target blackhole
Scanning dependencies of target blackhole_embedded
[ 40%] Building CXX object storage/blackhole/CMakeFiles/blackhole_embedded.dir/ha_blackhole.cc.o
In file included from /usr/local/mysql-5.6.12/sql/sql_plugin.h:34:0,
                 from /usr/local/mysql-5.6.12/sql/structs.h:23,
                 from /usr/local/mysql-5.6.12/sql/unireg.h:170,
                 from /usr/local/mysql-5.6.12/storage/blackhole/ha_blackhole.cc:19:
/usr/local/mysql-5.6.12/sql/field.h: 在成员函数‘void Field_blob::store_ptr_and_length(const char*, uint32)’中:
/usr/local/mysql-5.6.12/sql/field.h:3268:50: 警告：argument to ‘sizeof’ in ‘void* memmove(void*, const void*, size_t)’ call is the same pointer type ‘char*’ as the source; expected ‘char’ or an explicit length [-Wsizeof-pointer-memaccess]
     bmove(ptr + packlength, (char*) &from, sizeof(char *));

Issue #2730, opened 2016-08-02
Assertions related to lanes on 5.3 and master


daniel-j-h:
Splitting from https://github.com/Project-OSRM/osrm-backend/issues/2706: we're hitting the following lane assertion during extraction on the planet on master but also already on the 5.3 branch:

```
[assert] /mnt/ebs/osrm/2016-07-26-110507/src/extractor/guidance/turn_lane_augmentation.cpp:285
in: osrm::extractor::guidance::lanes::LaneDataVector osrm::extractor::guidance::lanes::handleNoneValueAtSimpleTurn(osrm::extractor::guidance::lanes::LaneDataVector, const Intersection&): connection_count + 1 == lane_data.size()
terminate called without an active exception
```

https://github.com/Project-OSRM/osrm-backend/blob/c7a1576100cb947af0e970644fb3f70dde63a9eb/src/extractor/guidance/turn_lane_augmentation.cpp#L283-L286

I wrote up some code for printing the locations for the breaking intersections in the `fix/2706-debug` branch and deployed it. For the record, here are the first five:

```
>>> 2, 4
 Tags:
	1(none) from: 0 to: 0
	2(straight) from: 1 to: 1
	8(left) from: 2 to: 2
	256(uturn) from: 2 to: 2
>>> Broken Intersection: 
(lon:7.828823, lat:47.976944)
(lon:7.828583, lat:47.975312)
(lon:7.828631, lat:47.97538)
>>> 2, 4
 Tags:
	1(none) from: 0 to: 0
	2(straight) from: 1 to: 2
	8(left) from: 3 to: 3
	256(uturn) from: 3 to: 3
>>> Broken Intersection: 
(lon:7.832026, lat:47.992681)
(lon:7.832197, lat:47.992714)
(lon:7.832428, lat:47.993222)
(lon:7.83195, lat:47.992815)
>>> 2, 4
 Tags:
	1(none) from: 0 to: 0
	2(straight) from: 1 to: 1
	8(left) from: 2 to: 2
	256(uturn) from: 2 to: 2
>>> Broken Intersection: 
(lon:7.809372, lat:48.011799)
(lon:7.809382, lat:48.011674)
(lon:7.809655, lat:48.011597)
(lon:7.809555, lat:48.011808)
>>> 2, 4
 Tags:
	1(none) from: 0 to: 0
	2(straight) from: 1 to: 1
	8(left) from: 2 to: 3
	256(uturn) from: 3 to: 3
[...]

PR #2739, opened 2016-08-04
Workaround for simple intersections that aren't: uturn lanes with oneway combination

daniel-j-h:
For https://github.com/Project-OSRM/osrm-backend/issues/2730.

This is a workaround at best. See the issue above for more context.

We traced this down by dumping the breaking locations, trying to come up with cucumber tests reproducing the issue, simplifying the cucumber test down to a small self contained scenario and trying various things before falling back to this workaround.

Affected are 190 intersections on the planet, from what I can see in the demo server logs.

Note: see the `fix/2706-debug` branch for what we deployed on the demo server.
This changeset is the gist of it.

@MoKob @danpat @TheMarex 

Not closing the original issue since this definitely needs a follow up.

### DIFF
--- a/src/extractor/guidance/turn_lane_augmentation.cpp
+++ b/src/extractor/guidance/turn_lane_augmentation.cpp
@@ -289,10 +289,12 @@ LaneDataVector handleNoneValueAtSimpleTurn(LaneDataVector lane_data,
         //
         if (connection_count + 1 != lane_data.size())
         {
-            goto these_intersections_are_clearly_broken_at_the_moment;
+            // skip broken intersections
         }
-
-        lane_data = mergeNoneTag(none_index, std::move(lane_data));
+        else
+        {
+            lane_data = mergeNoneTag(none_index, std::move(lane_data));
+        }
     }
     // we have to rename and possibly augment existing ones. The pure count remains the
     // same.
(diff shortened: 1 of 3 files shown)


src/extractor/guidance/turn_lane_handler.cpp
```diff
@@ -352,8 +352,14 @@ bool TurnLaneHandler::isSimpleIntersection(const LaneDataVector &lane_data,
             if (lane_data.back().tag == TurnLaneType::uturn)
                 return findBestMatchForReverse(lane_data[lane_data.size() - 2].tag, intersection);
 
-            BOOST_ASSERT(lane_data.front().tag == TurnLaneType::uturn);
-            return findBestMatchForReverse(lane_data[1].tag, intersection);
+            // TODO(mokob): #2730 have a look please
+            // BOOST_ASSERT(lane_data.front().tag == TurnLaneType::uturn);
+            // return findBestMatchForReverse(lane_data[1].tag, intersection);
+            //
+            if (lane_data.front().tag == TurnLaneType::uturn)
```
MoKob: Hm. This could be a case of left-sided driving, having u-turn lanes on the right side. Good catch, but looks good to me.

src/extractor/guidance/turn_lane_augmentation.cpp
```diff
@@ -283,16 +283,26 @@ LaneDataVector handleNoneValueAtSimpleTurn(LaneDataVector lane_data,
         // a pgerequisite is simple turns. Larger differences should not end up here
         // an additional line at the side is only reasonable if it is targeting public
         // service vehicles. Otherwise, we should not have it
-        BOOST_ASSERT(connection_count + 1 == lane_data.size());
-
-        lane_data = mergeNoneTag(none_index, std::move(lane_data));
+        //
```
MoKob: This is more like a workaround. We should never arrive here, when this is not true. Not handling the none-tag here simply fails out later instead of earlier. Will have to look into it.

PR #2740, opened 2016-08-04
Makes toll=yes way's routability configurable in profiles.

daniel-j-h:
The primary use-case is conditionally filtering ways such as:

https://www.openstreetmap.org/edit#map=18/38.94198/-77.28127

which we are guiding the user onto in the route from DC to IAD:

http://map.project-osrm.org/?z=12&center=38.934443%2C-77.167969&loc=38.902656%2C-77.029095&loc=38.952210%2C-77.453424&hl=en&alt=0

There are additional toll-free ways available for this route which are
a bit slower but do not require a fee.

This changeset makes `toll=yes` configurable in the profiles, disabling
them by default. Neither do we support time-based fees nor vehicle
category fees at the moment.

References:
- http://wiki.openstreetmap.org/wiki/Key:toll

cc @willwhite

### DIFF
--- a/features/car/access.feature
+++ b/features/car/access.feature
@@ -155,3 +155,10 @@ Feature: Car - Restricted access
             | primary | designated |       |
             | primary | yes        | x     |
             | primary | no         | x     |
+
+     Scenario: Car - only toll=yes ways are ignored by default
+        Then routability should be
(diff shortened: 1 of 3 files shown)


TheMarex:
:+1: looks good to me. Good to merge.

PR #2741, opened 2016-08-04
Calculate circle radius via Taubin fit

oxidase:
Added an algebraic fitting of a circle given a set of points. I have added some artificial tests, but probably some tests with  real data with expected radii must be added, but i have no a good example as in #2716

### DIFF
--- a/src/extractor/guidance/roundabout_handler.cpp
+++ b/src/extractor/guidance/roundabout_handler.cpp
@@ -290,22 +290,16 @@ RoundaboutType RoundaboutHandler::getRoundaboutType(const NodeID nid) const
     }
 
     // calculate the radius of the roundabout/rotary. For two coordinates, we assume a minimal
-    // circle
-    // with both vertices right at the other side (so half their distance in meters).
+    // circle with both vertices right at the other side (so half their distance in meters).
     // Otherwise, we construct a circle through the first tree vertices.
     const auto getRadius = [&roundabout_nodes, &getCoordinate]() {
-        auto node_itr = roundabout_nodes.begin();
-        if (roundabout_nodes.size() == 2)
-        {
-            const auto first = getCoordinate(*node_itr++), second = getCoordinate(*node_itr++);
-            return 0.5 * util::coordinate_calculation::haversineDistance(first, second);
-        }
-        else
-        {
-            const auto first = getCoordinate(*node_itr++), second = getCoordinate(*node_itr++),
-                       third = getCoordinate(*node_itr++);
-            return util::coordinate_calculation::circleRadius(first, second, third);
-        }
+        std::vector<util::Coordinate> coordinates;
+        coordinates.reserve(roundabout_nodes.size());
+        std::transform(roundabout_nodes.begin(),
+                       roundabout_nodes.end(),
+                       std::back_inserter(coordinates),
+                       getCoordinate);
+        return util::coordinate_calculation::circleRadius(std::move(coordinates));
     };
     const double radius = getRadius();
 
--- a/src/util/coordinate_calculation.cpp
+++ b/src/util/coordinate_calculation.cpp
@@ -275,6 +275,142 @@ double circleRadius(const Coordinate C1, const Coordinate C2, const Coordinate C
         return std::numeric_limits<double>::infinity();
 }
 
+/// Newton-based Taubin circle algebraic fit in a plain.
+/// Chernov, N. Circular and Linear Regression: Fitting Circles and Lines by Least Squares, 2010
+/// Chapter 5.10 Implementation of the Taubin fit, p. 126
+/// Original code http://people.cas.uab.edu/~mosya/cl/CircleFitByTaubin.cpp
+bool circleCenterTaubin(const std::vector<Coordinate> &coords,
+                        std::pair<double, double> &center,
+                        double &estimated_radius)
+{
+    // Compute mean and covariances with the two-pass algorithm
+    double meanX = 0., meanY = 0.;
+    for (const auto &c : coords)
+    {
+        // Use the equirectangular projection
+        const auto x = static_cast<double>(toFloating(c.lon));
+        const auto y = static_cast<double>(toFloating(c.lat));
+        meanX += x;
+        meanY += y;
+    }
+    meanX /= coords.size();
+    meanY /= coords.size();
+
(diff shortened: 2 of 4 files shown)


daniel-j-h:
Great work! Can you give insights into why you decided to go with Taubin's fit here as an algebraic circle fit algorithm?

Reading a bit on circle fits after you left the comment in the ticket last week, it seems to me we want a fast algebraic circle fit algorithm such as Kåsa's, Taubin's or Pratt's circle fit.

Based on [1] Kåsa seems to be fast but has problems with incomplete circles and tends to underestimates the radius, Pratt and Taubin seem to be more stable in that regard.

Also I think the only reason we can use an equirectangular projection is this, from [1]:

> All the above circle fits have an important property – they are indepen-
> dent of the choice of the coordinate system, i.e. their results are invariant
> under translations and rotations; see a proof in [14].


1  https://arxiv.org/abs/0907.0421

src/extractor/guidance/roundabout_handler.cpp
```diff
@@ -290,22 +290,16 @@ RoundaboutType RoundaboutHandler::getRoundaboutType(const NodeID nid) const
...
-                       third = getCoordinate(*node_itr++);
-            return util::coordinate_calculation::circleRadius(first, second, third);
-        }
+        std::vector<util::Coordinate> coordinates;
+        coordinates.reserve(roundabout_nodes.size());
+        std::transform(roundabout_nodes.begin(),
+                       roundabout_nodes.end(),
+                       std::back_inserter(coordinates),
+                       getCoordinate);
+        return util::coordinate_calculation::circleRadius(std::move(coordinates));
```
daniel-j-h: `circleRadius` takes its argument by `const &`, the `std::move` here is unnecessary.

src/util/coordinate_calculation.cpp
```diff
@@ -275,6 +275,142 @@ double circleRadius(const Coordinate C1, const Coordinate C2, const Coordinate C
...
+    for (const auto &c : coords)
+    {
+        // Use the equirectangular projection
+        const auto x = static_cast<double>(toFloating(c.lon));
+        const auto y = static_cast<double>(toFloating(c.lat));
+        meanX += x;
+        meanY += y;
+    }
+    meanX /= coords.size();
+    meanY /= coords.size();
```
daniel-j-h: This is UB when `coords` is empty, we should assert that precondition or add a guard at the top.

oxidase:
Pratt/Taubin fits are  better  for  partial circles. Here is the book preprint http://citeseerx.ist.psu.edu/viewdoc/download?doi=10.1.1.117.3465&rep=rep1&type=pdf
in tables 5.4 and 5.5 for Kasa fit the arc of 180 degrees is required. Taubin fit is good also for 45 degrees. Also Taubin fit is more stable than Pratt one.

I will update and rebase PR later, but do you have a real example of  misclassified roundabout?  I would like to  add as a test case.

daniel-j-h:
Thanks for the book preprint and the explanation! I don't have a misclassified example at the moment, I just stumbled over the logic in the code and opened a ticket :) I'll be out on the weekend but can probably have a look into Cucumber tests on Monday.

oxidase:
I have found some roundabouts that can be used as examples, but in general there is not so many places in Germany:
* circle roundabout with the diameter about 38-40 meters, but calculated radius is 8.8 meters http://www.openstreetmap.org/way/27189640#map=20/53.09893/9.97955&layers=HD
   points order 012 center  (lon:9.979355, lat:53.098919) radius 8.80153129481
   points order 102 center  (lon:9.979355, lat:53.098919) radius 13.8348914186
   fitted center  (lon:9.97951, lat:53.098904) avg radius 19.85258918 estimated radius  27.2258981869 
(lon:9.979225, lat:53.098906)
(lon:9.979307, lat:53.098798)
(lon:9.979288, lat:53.099032)
(lon:9.979457, lat:53.099094)
(lon:9.979778, lat:53.099003)
(lon:9.979643, lat:53.099081)
(lon:9.979773, lat:53.098835)
(lon:9.979638, lat:53.098761)

* elliptic roundabout with the axis 40 and 24 meters but calculated radius is 9.2 meters, but can be 5.6, 8.1
http://www.openstreetmap.org/#map=20/53.60554/9.46346
   points order 012 center  (lon:9.463431, lat:53.605458) radius 9.23869790604
   points order 102 center  (lon:9.463431, lat:53.605458) radius 5.67700961195
   points order 201 center  (lon:9.463431, lat:53.605458) radius 8.17698392569
   fitted center  (lon:9.463661, lat:53.605427) avg radius 19.3453867468 estimated radius  31.2062995529
(lon:9.463462, lat:53.605539)
(lon:9.463347, lat:53.605469)
(lon:9.46338, lat:53.605391)
(lon:9.463958, lat:53.605398)
(lon:9.463959, lat:53.605494)
(lon:9.463891, lat:53.605361)

Here some problems that must be fixed by PR:
- radius of 3 points calculation depends on the points ordering, it should not be the case and probably influence of  roundoff errors.
- 3 points calculation for ellipses could both underestimate and overestimate radius

Also  estimated fitted radius is always overestimated. This is expected, but overestimation 40-70% looks bad. Here average radius looks better.

MoKob:
Is this consistent for all possible roundabout entries/exits? The radius needs to be the same for all locations to not mess up internal instruction consistency.

MoKob:
Just had a closer look. Since the roundabout-nodes are still sorted, we should be fine for consistency issues.

MoKob:
This has been superseeded by a simple circumference heuristic for roundabouts. It's less error prone and good enough for our purpose.

However, we could potentially still use this to compute better turn penalties (https://github.com/Project-OSRM/osrm-backend/pull/2912).

A circle through a series of points close to a turn could give us a good approximation of a turn-radius. Here a circumference heuristic wouldn't work. However, an approximated fit like implemented here could give us a good idea.

The benefit would be that a turn with a large turn radius would usually require less breaking for the manoeuvre than a small turn-radius, thus requiring a smaller penalty.

MoKob:
Pulled this into https://github.com/Project-OSRM/osrm-backend/pull/2912/files to replace old code for circle radius.

Even though it is not used in the computation for roundabout radii, it will be used to compute turn radii. Thank you @oxidase for this. @daniel-j-h I've adjusted the `coords.size()` error you described.

Closing here.

PR #2748, opened 2016-08-05
Catch and report errors during osrm-contract tests.

danpat:
Some tests were feeding `osrm-contract` data that caused it to exit with a non-zero return code.
We weren't catching this error, and the test was logging the problem, but continuing anyway.

Because data from the previous test was often still in shared memory, and the tests shared the same data, the test that should have failed was passing.

This change properly feeds errors to the execution callback, and modifies the problematic tests so that they use different data so this problem can't repeat here.

This test now also checks for "not 0" return codes, as the way we're throwing errors now gives platform-specific error codes (uncaught exceptions leading to `abort()` on OSX causes return code 134, but code 1 on Linux).

### DIFF
--- a/features/car/traffic_speeds.feature
+++ b/features/car/traffic_speeds.feature
@@ -2,6 +2,8 @@
 Feature: Traffic - speeds
 
     Background: Use specific speeds
+
+    Scenario: Weighting based on speed file
         Given the node locations
             | node | lat        | lon      |
             | a    | 0.1        | 0.1      |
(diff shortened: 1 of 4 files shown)


springmeyer:
There are 5 other places where the error is not handled correctly for `reprocessAndLoadData`:

```
$ grep reprocessAndLoadData features/ -r
features//step_definitions/distance_matrix.js:        this.reprocessAndLoadData(() => {
features//step_definitions/matching.js:        this.reprocessAndLoadData(() => {
features//step_definitions/nearest.js:        this.reprocessAndLoadData(() => {
features//step_definitions/requests.js:        this.reprocessAndLoadData(() => {
features//step_definitions/routability.js:            this.reprocessAndLoadData((e) => {
features//step_definitions/trip.js:        this.reprocessAndLoadData(() => {
features//support/data.js:    this.reprocessAndLoadData = (callback) => {
features//support/shared_steps.js:        this.reprocessAndLoadData((e) => {
```

PR #2750, opened 2016-08-08
[ready] Scenario for unnecessary slight left onto Stralauer Str

daniel-j-h:
Failing Cucumber test for the following scenario:

http://map.project-osrm.org/?z=18&center=52.516206%2C13.410396&loc=52.517213%2C13.408626&loc=52.516018%2C13.411050&hl=en&alt=0

v

https://www.openstreetmap.org/edit#map=20/52.51609/13.41080

v2

## Requires:
 - [x] adjusted for comments.
 - [x] Fix features/testbot/bearing_param.feature:77

### DIFF
--- a/features/guidance/collapse.feature
+++ b/features/guidance/collapse.feature
@@ -302,12 +302,12 @@ Feature: Collapse
 
     Scenario: Entering a segregated road
         Given the node map
-            |   | a | f |   |   |
-            |   |   |   |   | g |
-            |   | b | e |   |   |
-            |   |   |   |   |   |
-            |   |   |   |   |   |
-            | c | d |   |   |   |
+            |   | a | f |   |   |   | g |
+            |   |   |   |   |   |   |   |
+            |   | b | e |   |   |   |   |
+            |   |   |   |   |   |   |   |
+            |   |   |   |   |   |   |   |
+            | c | d |   |   |   |   |   |
 
         And the ways
             | nodes | highway | name   | oneway |
@@ -615,12 +615,12 @@ Feature: Collapse
 
     Scenario: Forking before a turn (forky)
         Given the node map
-            |   |   |   | g |   |   |
-            |   |   |   |   |   |   |
-            |   |   |   | c |   |   |
-            | a | b |   |   |   |   |
-            |   |   |   |   | d |   |
-            |   |   |   |   | f | e |
+            |   |   |   |   |   | g |   |   |
+            |   |   |   |   |   |   |   |   |
+            |   |   |   |   |   | c |   |   |
+            | a |   |   | b |   |   |   |   |
+            |   |   |   |   |   |   | d |   |
+            |   |   |   |   |   |   | f | e |
 
         And the ways
             | nodes | name  | oneway | highway   |
--- a/features/guidance/collapse.feature
+++ b/features/guidance/collapse.feature
@@ -601,17 +601,19 @@ Feature: Collapse
             | bd    | road  | yes    | primary   |
             | bc    | road  | yes    | primary   |
             | de    | road  | yes    | primary   |
-            | fdcg  | cross | no     | secondary |
+            | fd    | cross | no     | secondary |
+            |  dc   | cross | no     | secondary |
+            |   cg  | cross | no     | secondary |
 
         And the relations
             | type        | way:from | way:to | node:via | restriction   |
-            | restriction | bd       | fdcg   | d        | no_left_turn  |
-            | restriction | bc       | fdcg   | c        | no_right_turn |
+            | restriction | bd       | dc     | d        | no_left_turn  |
+            | restriction | bc       | dc     | c        | no_right_turn |
 
         When I route I should get
-          | waypoints | route            | turns                   |
-          | a,g       | road,cross,cross | depart,turn left,arrive |
-          | a,e       | road,road        | depart,arrive           |
+            | waypoints | route            | turns                   |
+            | a,g       | road,cross,cross | depart,turn left,arrive |
+            | a,e       | road,road        | depart,arrive           |
 
     Scenario: Forking before a turn (forky)
         Given the node map
@@ -628,12 +630,14 @@ Feature: Collapse
             | bd    | road  | yes    | primary   |
             | bc    | road  | yes    | primary   |
             | de    | road  | yes    | primary   |
-            | fdcg  | cross | no     | secondary |
+            | fd    | cross | no     | secondary |
+            |  dc   | cross | no     | secondary |
+            |   cg  | cross | no     | secondary |
 
         And the relations
             | type        | way:from | way:to | node:via | restriction   |
-            | restriction | bd       | fdcg   | d        | no_left_turn  |
-            | restriction | bc       | fdcg   | c        | no_right_turn |
+            | restriction | bd       | dc     | d        | no_left_turn  |
+            | restriction | bc       | dc     | c        | no_right_turn |
 
         When I route I should get
             | waypoints | route                 | turns                                          |
--- a/features/guidance/collapse.feature
+++ b/features/guidance/collapse.feature
@@ -325,7 +325,6 @@ Feature: Collapse
             | g,f       | second,first,first  | depart,turn right,arrive       |
             | g,c       | second,first,first  | depart,end of road left,arrive |
 
-
     Scenario: Do not collapse turning roads
         Given the node map
             |   |   | e |   |   |
@@ -333,12 +332,12 @@ Feature: Collapse
             | a |   | b | f |   |
 
         And the ways
-            | nodes | highway | name   |
-            | ab    | primary | first  |
-            | bc    | primary | first  |
-            | cd    | primary | first  |
-            | ce    | primary | second |
-            | bf    | primary | third  |
+            | nodes | highway | name   | oneway |
+            | ab    | primary | first  | yes    |
+            | bc    | primary | first  | yes    |
+            | cd    | primary | first  | yes    |
+            | ce    | primary | second | yes    |
+            | bf    | primary | third  | yes    |
 
         When I route I should get
             | waypoints | route                   | turns                                      |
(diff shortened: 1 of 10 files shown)


features/guidance/turn.feature
```diff
@@ -940,12 +940,12 @@ Feature: Simple Turns
 
         And the ways
             | nodes | name          | highway   | oneway |
-            | ab    | Molkenmarkt   | secondary | yes    |
+            | ab    | Stralauer Str | secondary | yes    |
```
daniel-j-h: `ab` models this way https://www.openstreetmap.org/way/23812656 and is not Stralauer Str but Molkenmarkt instead.

MoKob: at the intersection itself it is modelled as Stralauer Str in the data... One could argue that the modeling of this intersection which has Molkenmarkt on multiple of the roads is more than strange, though. I'd question whether the name Mokenmarkt in between is acutally referring to the roads, but rather to the space in between.

daniel-j-h: Yep but `ab` is not at the intersection, it's the way right before `bc` which is Stralauer Str.

ab - Molkenmarkt - https://www.openstreetmap.org/way/23812656
bc - Stralauer Str - https://www.openstreetmap.org/way/286653195
cd - Stralauer Str - https://www.openstreetmap.org/way/60333309
ec - Molkenmarkt - https://www.openstreetmap.org/way/156364566

MoKob:
As long as the tag is as it is right now, we could only try and handle these cases with a serious perfomance impact.
If the Stralauer Straße starts prior to the intersection and has just the same road class as Spandauer Straße, the case is not distinguishable from a merge onto another street.
The only way we could find out about this situation is if we would check the road coming from the side, check for sliproads and ignore the merge if we find the incoming road to be a sliproad.

I've marked the respective case as todo, since I cannot see a reasonable way of handling it right now. It would be easier if the road in between was a link road. In that case we might consider incoming links as non-through street and start the actual street at the intersection.

features/guidance/collapse.feature
```diff
@@ -344,7 +344,7 @@ Feature: Collapse
             | waypoints | route                   | turns                                      |
             | a,d       | first,first,first,first | depart,continue left,continue right,arrive |
             | a,e       | first,second,second     | depart,turn left,arrive                    |
-            | a,f       | first,third,third       | depart,turn straight,arrive                |
+            | a,f       | first,third,third       | depart,new name straight,arrive            |
```
TheMarex: Why did this change? Seems like the original behavior of a "turn straight" made more sense in this context.

features/guidance/collapse.feature
```diff
@@ -601,39 +601,43 @@ Feature: Collapse
             | bd    | road  | yes    | primary   |
             | bc    | road  | yes    | primary   |
             | de    | road  | yes    | primary   |
-            | fdcg  | cross | no     | secondary |
+            | fd    | cross | no     | secondary |
+            |  dc   | cross | no     | secondary |
+            |   cg  | cross | no     | secondary |
```
TheMarex: Formatting is messed up

features/guidance/continue.feature
```diff
@@ -16,9 +16,9 @@ Feature: Continue Instructions
             | bd     | primary |
 
        When I route I should get
-            | waypoints | route       | turns                       |
-            | a,c       | abc,abc,abc | depart,continue left,arrive |
-            | a,d       | abc,bd,bd   | depart,turn straight,arrive |
+            | waypoints | route       | turns                           |
+            | a,c       | abc,abc,abc | depart,continue left,arrive     |
+            | a,d       | abc,bd,bd   | depart,new name straight,arrive |
```
TheMarex: Same weird change as above.

features/guidance/continue.feature
```diff
@@ -64,9 +64,9 @@ Feature: Continue Instructions
             | bd     | primary |
 
        When I route I should get
-            | waypoints | route       | turns                        |
-            | a,c       | abc,abc,abc | depart,continue right,arrive |
-            | a,d       | abc,bd,bd   | depart,turn straight,arrive  |
+            | waypoints | route       | turns                            |
+            | a,c       | abc,abc,abc | depart,continue right,arrive     |
+            | a,d       | abc,bd,bd   | depart,new name straight,arrive  |
```
TheMarex: And again.

features/guidance/new-name.feature
```diff
@@ -125,10 +125,10 @@ Feature: New-Name Instructions
             |   |   |   |   | c |
 
         And the ways
-            | nodes  | highway     |
-            | ab     | residential |
-            | bc     | residential |
-            | bd     | service     |
+            | nodes  | highway     | oneway |
```
TheMarex: Why was oneway added here?

features/guidance/turn.feature
```diff
@@ -930,3 +930,41 @@ Feature: Simple Turns
         When I route I should get
             | waypoints | turns                                        | route                                                         |
             | a,h       | depart,off ramp right,turn sharp left,arrive | Blue Star Memorial Hwy,bcde,Centreville Road,Centreville Road |
+
+    @todo
```
TheMarex: This is still todo?

src/extractor/guidance/intersection_handler.cpp
```diff
@@ -422,26 +477,56 @@ std::size_t IntersectionHandler::findObviousTurn(const EdgeID via_edge,
...
+            std::cout << "Index to the right prevents it." << std::endl;
             return 0;
-        if (angularDeviation(intersection[(best + 1) % intersection.size()].turn.angle,
-                             STRAIGHT_ANGLE) <= FUZZY_ANGLE_DIFFERENCE)
+        }
+        if (angularDeviation(intersection[left_index].turn.angle, STRAIGHT_ANGLE) <=
+                FUZZY_ANGLE_DIFFERENCE &&
+            !obvious_to_left)
+        {
+            std::cout << "Index to the left prevents it." << std::endl;
```
TheMarex: Leftover debugging code

src/extractor/guidance/intersection_handler.cpp
```diff
@@ -422,26 +477,56 @@ std::size_t IntersectionHandler::findObviousTurn(const EdgeID via_edge,
...
-             (left_deviation > best_deviation &&
-              !intersection[(best + 1) % intersection.size()].entry_allowed)) &&
-            (right_deviation / best_deviation >= DISTINCTION_RATIO ||
-             (right_deviation > best_deviation && !intersection[best - 1].entry_allowed)))
-        {
+        if ((distinct_to_left || obvious_to_left) && (distinct_to_right || obvious_to_right))
             return best;
-        }
+
+        std::cout << "Failed: " << distinct_to_left << " " << distinct_to_right << " "
```
TheMarex: Again left-over debugging code.

src/extractor/guidance/turn_handler.cpp
```diff
@@ -165,9 +168,11 @@ Intersection TurnHandler::handleThreeWayTurn(const EdgeID via_edge, Intersection
     }
     else
     {
-        if (isObviousOfTwo(via_edge, intersection[1], intersection[2]) &&
-            (in_data.name_id != second_data.name_id || first_data.name_id == second_data.name_id))
+        std::cout << "Name IDs: " << (int)in_data.name_id << " " << (int)first_data.name_id << " "
```
TheMarex: Debugging code..

src/extractor/guidance/turn_handler.cpp
```diff
@@ -165,9 +168,11 @@ Intersection TurnHandler::handleThreeWayTurn(const EdgeID via_edge, Intersection
...
+                  << (int)second_data.name_id << std::endl;
+        if (obvious_index == 1)
         {
+            std::cout << "Assigning Obvious first" << std::endl;
```
TheMarex: Debugging code..

src/extractor/guidance/turn_handler.cpp
```diff
@@ -177,9 +182,9 @@ Intersection TurnHandler::handleThreeWayTurn(const EdgeID via_edge, Intersection
                                                 getTurnDirection(intersection[1].turn.angle)};
         }
 
-        if (isObviousOfTwo(via_edge, intersection[2], intersection[1]) &&
-            (in_data.name_id != first_data.name_id || first_data.name_id == second_data.name_id))
+        if (obvious_index == 2)
         {
+            std::cout << "Assigning Obvious Second" << std::endl;
```
TheMarex: Debugging code..

src/extractor/guidance/turn_handler.cpp
```diff
@@ -195,6 +200,10 @@ Intersection TurnHandler::handleThreeWayTurn(const EdgeID via_edge, Intersection
 Intersection TurnHandler::handleComplexTurn(const EdgeID via_edge, Intersection intersection) const
 {
     const std::size_t obvious_index = findObviousTurn(via_edge, intersection);
+    std::cout << "[intersection]\n";
```
TheMarex: Debugging code..

features/guidance/new-name.feature
MoKob: simply makes it easier to debug and does not change the tested instructions.
If we have oneways, only one intersection is considered in the output. Otherwise we get three.

TheMarex:
This was pulled in #2727 and improved. There is one scenario that is not yet fixed and marked with `@todo`. Further discussions can happen in a ticket.

MoKob:
I've pulled the adjustments from #2780 into this PR and adjusted for the comments over there. From discussion in chat/voice this would be ready to get merged.

TheMarex:
@MoKob seems like there is [one failing scenario](https://travis-ci.org/Project-OSRM/osrm-backend/jobs/153544494#L3976-L3997) in travis. Is there a fix for this in #2780 or is that scenario just broken?

MoKob:
This actually seems to be a legit error. Will investigate what is going on here.

TheMarex:
Looks good to me.

PR #2752, opened 2016-08-08
Scenario for unnecessary slight right onto Holzmarktstr due to node offset

daniel-j-h:
Failing test for sensitive node placement. cc @MoKob 

# TODO
 - [x] discussion on how to handle node offsets
 - [x] check impact of https://github.com/Project-OSRM/osrm-backend/pull/2779 on this

## Requirements
 - [x] merge of https://github.com/Project-OSRM/osrm-backend/pull/2750 (base branch)
 - [x] merge of https://github.com/Project-OSRM/osrm-backend/pull/2727 (base branch)
 - [x] review
 - [x] adjust for comments

### DIFF
--- a/src/engine/guidance/post_processing.cpp
+++ b/src/engine/guidance/post_processing.cpp
@@ -1,4 +1,5 @@
 #include "engine/guidance/post_processing.hpp"
+#include "extractor/guidance/constants.hpp"
 #include "extractor/guidance/toolkit.hpp"
 #include "extractor/guidance/turn_instruction.hpp"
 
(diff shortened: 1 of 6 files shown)


MoKob:
I've managed to improve the perception here by considering the input/exit bearings in the collapse process. Due to the strong angle (for 10 meters) in the current scenario, the angle remains as a slight right, it emits a new-name now, though.

The new-name is technically correct for this setting. The test added here expects no instruction though. @daniel-j-h I feel that this might be an oversight in generating the test?

If we want to detect intense offsets like this one as straight, we need to look at ways to detect bad node-placement in the noisy data. I feel that this could be a bad idea though, as we would be fixing broken data at that point.

daniel-j-h:
> The new-name is technically correct for this setting. The test added here expects no instruction though. @daniel-j-h I feel that this might be an oversight in generating the test?

Yes good catch.

src/engine/guidance/post_processing.cpp
```diff
@@ -567,26 +568,150 @@ void collapseTurnAt(std::vector<RouteStep> &steps,
...
+                                       .bearings[current_step.intersections.front().in]),
+                               current_step.intersections.front()
+                                   .bearings[current_step.intersections.front().out]);
+
+                const auto bearing_turn_angle =
+                    turn_angle(util::bearing::reverseBearing(
+                                   one_back_step.intersections.front()
+                                       .bearings[one_back_step.intersections.front().in]),
+                               current_step.intersections.front()
+                                   .bearings[current_step.intersections.front().out]);
```
daniel-j-h: Want to write a small function abstracting over the common pattern here?

src/extractor/guidance/turn_analysis.cpp
```diff
@@ -28,7 +28,7 @@ using EdgeData = util::NodeBasedDynamicGraph::EdgeData;
 
 bool requiresAnnouncement(const EdgeData &from, const EdgeData &to)
 {
-    return !from.IsCompatibleTo(to);
+    return !from.CanCombineWith(to);
```
daniel-j-h: Can you explain this change to me?

MoKob: It simply strengthens the check a bit.

PR #2753, opened 2016-08-08
[ready] adjusting u-turn handling in lane matching to fully fix 2706

MoKob:
https://github.com/Project-OSRM/osrm-backend/pull/2739/files#diff-e4ce6eeb194b7b13640b6b48d96a052dR355 changes behaviour of an assertion instead of fixing the root of the problem.

The problem here was that sorting u-turns in the array of lane entries. The PR open here adds a regression test-case triggering the behaviour, fixes the sorting and in addition handles right-sided u-turn lanes. 

- [x] reviewed
- [x] adjusted for comments
- [x] make sure it runs on the planet

### DIFF
--- a/features/guidance/turn-lanes.feature
+++ b/features/guidance/turn-lanes.feature
@@ -863,7 +863,7 @@ Feature: Turn Lane Guidance
             | d,a ||||
             # Note: at the moment we don't care about routes, we care about the extract process triggering assertions
 
-    @reverse @2730
+    @reverse @2730 @todo
     Scenario: Reverse on the right
         Given the node map
             | a |   |   | c |   |
--- a/features/guidance/turn-lanes.feature
+++ b/features/guidance/turn-lanes.feature
@@ -740,8 +740,8 @@ Feature: Turn Lane Guidance
 
         And the ways
             | nodes | name | highway | turn:lanes:forward |
-            | ab    | road | primary | through,right      |
-            | bc    | road | primary | through,right      |
+            | ab    | road | primary | through;right      |
+            | bc    | road | primary | through;right      |
             | cd    | road | primary |                    |
             | xa    | road | primary |                    |
             | be    | turn | primary |                    |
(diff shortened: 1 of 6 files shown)


src/extractor/guidance/intersection_generator.cpp
```diff
@@ -277,7 +277,7 @@ Intersection IntersectionGenerator::mergeSegregatedRoads(Intersection intersecti
     {
         const double correction_factor = (intersection[1].turn.angle) / 2;
         for (std::size_t i = 2; i < intersection.size(); ++i)
-            intersection[i].turn.angle += correction_factor;
+            intersection[i].turn.angle -= correction_factor;
```
TheMarex: Can you add a comment outlining why the angle needs to be adjusted? It is not obvious to me from the surrounding code. Similarly why do you subtract here but add above?

src/extractor/guidance/turn_lane_handler.cpp
```diff
@@ -119,9 +119,8 @@ Intersection TurnLaneHandler::assignTurnLanes(const NodeID at,
 
     if (!lane_data.empty() && canMatchTrivially(intersection, lane_data) &&
         lane_data.size() !=
-            static_cast<std::size_t>(
-                lane_data.back().tag != TurnLaneType::uturn && intersection[0].entry_allowed ? 1
-                                                                                             : 0) +
+            static_cast<std::size_t>((
```
TheMarex: This calculation would be easier to read outside of the if header.

src/extractor/guidance/turn_lane_handler.cpp
```diff
@@ -352,14 +351,8 @@ bool TurnLaneHandler::isSimpleIntersection(const LaneDataVector &lane_data,
             if (lane_data.back().tag == TurnLaneType::uturn)
                 return findBestMatchForReverse(lane_data[lane_data.size() - 2].tag, intersection);
```
TheMarex: Can you add more context here why you need to find a match for `lane_data[-2]` if `lane_data[-1]` is a uturn?
I think it would help to use named variables here for documentation.

src/extractor/guidance/turn_lane_handler.cpp
```diff
@@ -352,14 +351,8 @@ bool TurnLaneHandler::isSimpleIntersection(const LaneDataVector &lane_data,
...
-            // TODO(mokob): #2730 have a look please
-            // BOOST_ASSERT(lane_data.front().tag == TurnLaneType::uturn);
-            // return findBestMatchForReverse(lane_data[1].tag, intersection);
-            //
-            if (lane_data.front().tag == TurnLaneType::uturn)
-                return findBestMatchForReverse(lane_data[1].tag, intersection);
-
-            return findBestMatch(data.tag, intersection);
+            BOOST_ASSERT(lane_data.front().tag == TurnLaneType::uturn);
+            return findBestMatchForReverse(lane_data[1].tag, intersection);
```
TheMarex: Similarly here. What does `lane_data[1]` and `lane_data[0]` represent respectively?

MoKob:
@TheMarex I adjusted the PR for your comments / made some stuff easier to read. Can you have a final look to see if your concerns have been properly addressed?

features/guidance/turn-lanes.feature
```diff
@@ -740,8 +740,8 @@ Feature: Turn Lane Guidance
 
         And the ways
             | nodes | name | highway | turn:lanes:forward |
-            | ab    | road | primary | through,right      |
-            | bc    | road | primary | through,right      |
+            | ab    | road | primary | through;right      |
```
TheMarex: Is this an unrelated bug in the scenario?

include/extractor/guidance/toolkit.hpp
```diff
@@ -323,7 +323,7 @@ trimLaneString(std::string lane_string, std::int32_t count_left, std::int32_t co
         for (std::int32_t i = 0; i < count_left; ++i)
             // this is adjusted for our fake pipe. The moment cucumber can handle multiple escaped
             // pipes, the '&' part can be removed
-            if (lane_string[i] != '|' && lane_string[i] != '&')
+            if (lane_string[i] != '|')
```
TheMarex: I thought this custom separator logic got removed after we upgraded cucumber?

MoKob: apparently not completely

Issue #2754, opened 2016-08-08
Crashing requests in Portland, OR


TheMarex:
These requests cause segmentation faults on `master` and `5.3.1`:

```
http://127.0.0.1:5000/route/v1/driving/-122.790939337,45.3117507691;-122.77279825,45.317365299;-122.756437278,45.3174842382;-122.743776748,45.3262135294;-122.737294855,45.3451939719;-122.716595074,45.3584805038;-122.70681792,45.3690072941;-122.698613545,45.3791216948;-122.690187134,45.3928220412;-122.666455954,45.3832097166;-122.650200594,45.3720805608;-122.650358121,45.3735136853.json?alternatives=false&overview=false&steps=true

http://127.0.0.1:5000/route/v1/driving/-122.584629301,45.4799720832;-122.589213729,45.4848174409;-122.589268433,45.4864360812;-122.594626229,45.4887861992;-122.604808901,45.4937090585;-122.611385007,45.4986346839;-122.611028608,45.5031273002;-122.616101336,45.5033468642;-122.622637795,45.5041525327;-122.630226519,45.5048628571.json?alternatives=false&overview=false&steps=true

http://127.0.0.1:5000/route/v1/driving/-122.584629301,45.4799720832;-122.589213729,45.4848174409;-122.589268433,45.4864360812;-122.594626229,45.4887861992;-122.604808901,45.4937090585;-122.611385007,45.4986346839;-122.611028608,45.5031273002;-122.616101336,45.5033468642;-122.622637795,45.5041525327;-122.630226519,45.5048628571.json?alternatives=false&overview=false&steps=true

http://127.0.0.1:5000/route/v1/driving/-122.790939337,45.3117507691;-122.77279825,45.317365299;-122.756437278,45.3174842382;-122.743776748,45.3262135294;-122.737294855,45.3451939719;-122.716595074,45.3584805038;-122.70681792,45.3690072941;-122.698613545,45.3791216948;-122.690187134,45.3928220412;-122.666455954,45.3832097166;-122.650200594,45.3720805608;-122.650358121,45.3735136853.json?alternatives=false&overview=false&steps=true
```

Crashes at this location:

```
#6  0x0000000000dc9d3f in boost::assertion_failed (expr=expr@entry=0xe7a580 "geometry.segment_offsets[1] == 1", 
[...]

PR #2755, opened 2016-08-08
[review] fix removal of multiple geometries for first vertex

MoKob:
Working on https://github.com/Project-OSRM/osrm-backend/issues/2754.

If steps of length <= 1 get removed, we might end up removing more than a single coordinate. This is broken currently, resulting in segfaults.

The PR already fixes this for the initial step, needs to work on the last step as well.

 - [x] fix first step
 - [x] fix last step
 - [x] review
 - [x] adjust for comments

### DIFF
--- a/src/engine/guidance/post_processing.cpp
+++ b/src/engine/guidance/post_processing.cpp
@@ -1,5 +1,5 @@
-#include "engine/guidance/post_processing.hpp"
 #include "extractor/guidance/turn_instruction.hpp"
+#include "engine/guidance/post_processing.hpp"
 
 #include "engine/guidance/assemble_steps.hpp"
 #include "engine/guidance/lane_processing.hpp"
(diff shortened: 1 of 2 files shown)


MoKob:
@TheMarex given that we wan't to add this to 5.3, I guess the Changelog entry has to go at some special place? Do you wan't to add this to the correct location on merge/Cherry Pick?

- fixed a bug that occurred when trimming very short segments at the begin/end of a route (less than 1 meter)

TheMarex:
@MoKob for `master` we don't need a changelog entry. To run out `5.3.2`:

- [ ] Cherry-pick to the `5.3`
- [ ] Update `CHANGELOG.md` for `5.3.2` with the entry you proposed above
- [ ] Follow https://github.com/Project-OSRM/osrm-backend/blob/master/docs/releasing.md#releasing-a-version

src/engine/guidance/post_processing.cpp
```diff
@@ -897,17 +897,27 @@ void trimShortSegments(std::vector<RouteStep> &steps, LegGeometry &geometry)
...
+        const auto offset = zero_length_step ? geometry.segment_offsets[1] : 1;
+
         // We have to adjust the first step both for its name and the bearings
         if (zero_length_step)
         {
+            // since we are not only checking for epsilon but for a full meter, we can have multiple
+            // coordinates here
+            for (int count = 1; count < offset; ++count)
+            {
+                geometry.locations.erase(geometry.locations.begin());
```
TheMarex: Instead of the loop you can use the range version of `erase`:

`geometry.location.erase(geometry.locations.begin(), geometry.locations.begin() + offset);`

src/engine/guidance/post_processing.cpp
```diff
@@ -982,10 +992,14 @@ void trimShortSegments(std::vector<RouteStep> &steps, LegGeometry &geometry)
...
     if (next_to_last_step.distance <= 1 && steps.size() > 2)
     {
-        geometry.locations.pop_back();
-        geometry.annotations.pop_back();
-        geometry.osm_node_ids.pop_back();
         geometry.segment_offsets.pop_back();
+        while (geometry.locations.size() > (geometry.segment_offsets.back() + 1))
+        {
+            geometry.locations.pop_back();
+            geometry.annotations.pop_back();
```
TheMarex: Here you can use `geometry.locations.resize(geometry.segment_offsets.back() + 1)` which won't discard the first n elements.

PR #2756, opened 2016-08-08
JSON instruction serialization fix.

gardster:
In some cases the OSRM server returns ```{"code": "InternalError","message":"Internal Server Error"}```. 
We reproduce it on request:
```
route/v1/car/-73.92415,40.86168;-73.93144,40.86696;-73.80145,40.67074;-74.30066,40.51935;-74.38946,40.5267?steps=true&geometries=geojson&bearings=0,180;282,45;180,45;272,45;336,45&overview=false
```
This happens because TurnType::Sliproad has not translation. I fix it.

### DIFF
--- a/include/engine/guidance/step_maneuver.hpp
+++ b/include/engine/guidance/step_maneuver.hpp
@@ -17,9 +17,10 @@ namespace guidance
 
 enum class WaypointType : std::uint8_t
 {
-    None,
-    Arrive,
-    Depart,
+    None = 0,
+    Arrive = 1,
+    Depart = 2,
+    MaxWaypointType = 3
 };
 
 struct StepManeuver
--- a/include/extractor/guidance/turn_instruction.hpp
+++ b/include/extractor/guidance/turn_instruction.hpp
@@ -16,12 +16,6 @@ namespace extractor
 namespace guidance
 {
 
-namespace detail
-{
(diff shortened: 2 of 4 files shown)


daniel-j-h:
cc @MoKob 

It's so easy to forget about this, since it's in totally different code locations and the array is indexed by the enum. Can we do something about this?

Maybe adding a final `ENUM_MAX_SIZE` sentinel to the enums and then `static_assert` already at compile time. What makes this even harder is the fact that these types are not really enums but `uint8_t` instead..

MoKob:
@daniel-j-h there seems to be a deeper problem here. Sliproad should never even get to this place. In fact, none of the invalid instructions should ever get there.

gardster:
```ENUM_MAX_SIZE``` seems good. I wanted to implement this approach but was not sure that you accept it, because there is no max_size fields in other enums.

TheMarex:
> ENUM_MAX_SIZE seems good. I wanted to implement this approach but was not sure that you accept it, because there is no max_size fields in other enums.

Yes actually we should have this for every enum that is used as an index to make sure stuff like this does not happen.

@MoKob could you replicate the issue with `sliproad` being exposed?

gardster:
Added max size fields. PTAL

TheMarex:
Looks great. Thanks @gardster!

PR #2759, opened 2016-08-09
[ready] Fix map matching crash

TheMarex:
The current implementation is kind of a hack. To do this properly we would need to refactor where the loop special case handling lives (needs to know which nodes are source/destination nodes).

- [x] Figure out why this breaks tests
- [x] ~~Add test case~~
- [x] Review
- [x] Incorporate review comments

### DIFF
--- a/include/engine/datafacade/shared_datafacade.hpp
+++ b/include/engine/datafacade/shared_datafacade.hpp
@@ -276,11 +276,6 @@ class SharedDataFacade final : public BaseDataFacade
 
     void LoadCoreInformation()
     {
-        if (data_layout->num_entries[storage::SharedDataLayout::CORE_MARKER] <= 0)
-        {
-            return;
-        }
-
         auto core_marker_ptr = data_layout->GetBlockPtr<unsigned>(
             shared_memory, storage::SharedDataLayout::CORE_MARKER);
         util::ShM<bool, true>::vector is_core_node(
--- a/include/engine/routing_algorithms/routing_base.hpp
+++ b/include/engine/routing_algorithms/routing_base.hpp
@@ -90,14 +90,11 @@ template <class DataFacadeT, class Derived> class BasicRoutingInterface
             if (new_distance < upper_bound)
             {
                 // if loops are forced, they are so at the source
-                if (new_distance >= 0 &&
-                    (!force_loop_forward || forward_heap.GetData(node).parent != node) &&
-                    (!force_loop_reverse || reverse_heap.GetData(node).parent != node))
-                {
-                    middle_node_id = node;
-                    upper_bound = new_distance;
-                }
-                else
+                if ((force_loop_forward && forward_heap.GetData(node).parent == node) ||
+                    (force_loop_reverse && reverse_heap.GetData(node).parent == node) ||
+                    // in this case we are looking at a bi-directional way where the source
+                    // and target phantom are on the same edge based node
+                    new_distance < 0)
                 {
                     // check whether there is a loop present at the node
                     for (const auto edge : facade->GetAdjacentEdgeRange(node))
@@ -120,6 +117,13 @@ template <class DataFacadeT, class Derived> class BasicRoutingInterface
                             }
                         }
                     }
+                }
+                else
+                {
+                    BOOST_ASSERT(new_distance >= 0);
+
+                    middle_node_id = node;
+                    upper_bound = new_distance;
                 }
             }
         }
@@ -524,7 +528,12 @@ template <class DataFacadeT, class Derived> class BasicRoutingInterface
                                           std::vector<NodeID> &packed_path) const
     {
         NodeID current_node_id = middle_node_id;
-        while (current_node_id != search_heap.GetData(current_node_id).parent)
+        // all initial nodes will have itself as parent, or a node not in the heap
+        // in case of a core search heap. We need a distinction between core entry nodes
+        // and start nodes since otherwise start node specific code that assumes
+        // node == node.parent (e.g. the loop code) might get actived.
+        while (current_node_id != search_heap.GetData(current_node_id).parent &&
+               search_heap.WasInserted(search_heap.GetData(current_node_id).parent))
         {
             current_node_id = search_heap.GetData(current_node_id).parent;
             packed_path.emplace_back(current_node_id);
@@ -636,8 +645,9 @@ template <class DataFacadeT, class Derived> class BasicRoutingInterface
         NodeID middle = SPECIAL_NODEID;
         distance = duration_upper_bound;
 
-        std::vector<std::pair<NodeID, EdgeWeight>> forward_entry_points;
-        std::vector<std::pair<NodeID, EdgeWeight>> reverse_entry_points;
+        using CoreEntryPoint = std::tuple<NodeID, EdgeWeight, NodeID>;
+        std::vector<CoreEntryPoint> forward_entry_points;
+        std::vector<CoreEntryPoint> reverse_entry_points;
 
         // get offset to account for offsets on phantom nodes on compressed edges
         const auto min_edge_offset = std::min(0, forward_heap.MinKey());
@@ -654,7 +664,7 @@ template <class DataFacadeT, class Derived> class BasicRoutingInterface
                 {
                     const NodeID node = forward_heap.DeleteMin();
                     const int key = forward_heap.GetKey(node);
-                    forward_entry_points.emplace_back(node, key);
+                    forward_entry_points.emplace_back(node, key, forward_heap.GetData(node).parent);
                 }
                 else
                 {
@@ -675,7 +685,7 @@ template <class DataFacadeT, class Derived> class BasicRoutingInterface
                 {
                     const NodeID node = reverse_heap.DeleteMin();
                     const int key = reverse_heap.GetKey(node);
-                    reverse_entry_points.emplace_back(node, key);
+                    reverse_entry_points.emplace_back(node, key, reverse_heap.GetData(node).parent);
                 }
                 else
                 {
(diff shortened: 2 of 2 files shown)


include/engine/routing_algorithms/routing_base.hpp
```diff
@@ -90,15 +90,8 @@ template <class DataFacadeT, class Derived> class BasicRoutingInterface
...
-                if (new_distance >= 0 &&
-                    (!force_loop_forward || forward_heap.GetData(node).parent != node) &&
-                    (!force_loop_reverse || reverse_heap.GetData(node).parent != node))
-                {
-                    middle_node_id = node;
-                    upper_bound = new_distance;
-                }
-                else
-                {
+                if ((force_loop_forward && forward_heap.GetData(node).parent == node) ||
```
MoKob: The switch of the instructions here is not entirely the same.

Shouldn't the first check include a `new_distance < 0 ||`? From what I've seen you inverted the check and switched the if/else branches? But the new conditions are not entirely the same. On `< 0` distances, the now `if` branch should be executed, not the `else` part. Or not?

include/engine/routing_algorithms/routing_base.hpp
```diff
@@ -636,8 +640,9 @@ template <class DataFacadeT, class Derived> class BasicRoutingInterface
         NodeID middle = SPECIAL_NODEID;
         distance = duration_upper_bound;
 
-        std::vector<std::pair<NodeID, EdgeWeight>> forward_entry_points;
-        std::vector<std::pair<NodeID, EdgeWeight>> reverse_entry_points;
+        using CoreEntryPoint = std::tuple<NodeID, EdgeWeight>;
```
MoKob: any reason to switch to tuple here over pair? Comparison should have been here without the entry_point_comparator already. I feel the access logic for tuple is not that beautiful to read, so I'd prefer to stay at pair.

include/engine/routing_algorithms/routing_base.hpp
TheMarex: Yes the branching logic is not equivalent. Good catch. However is the new_distance < 0 case ever triggered without forcing at least one loop?

I'm currently trying to think of a scenario to make sure this is covered.

include/engine/routing_algorithms/routing_base.hpp
```diff
@@ -524,7 +524,11 @@ template <class DataFacadeT, class Derived> class BasicRoutingInterface
                                           std::vector<NodeID> &packed_path) const
     {
         NodeID current_node_id = middle_node_id;
-        while (current_node_id != search_heap.GetData(current_node_id).parent)
+        // all initial nodes will have itself as parent
+        // all incomplete paths (e.g. paths ending at the core boundary) will have SPECIAL_NODEID
+        // as parent
```
MoKob: how does a path look like that starts directly at a core node?

TheMarex: That is a good point. In the current setup we always insert the start/target nodes in the non-core search heap, with themselves as parent. When we would relax that node, we check if its a core node and add it to the list of entry points, which are then added to the core search heap with `SPECIAL_NODEID` as parent. As such these nodes would never trigger a this code path.

We need to preserve these, thanks for finding this edge case. :+1:

TheMarex:
Constructing a regression test proofs quite hard. The conditions that need to be right for this bug to hit:

- Source and target phantom nodes are not allowed to be in the core
- There need to be a node that:
   - Is adjacent to both the source and target phantom
   - has a loop edge
   - the distance to that node + loop weight needs to small enough so that the core search terminates

TheMarex:
Okay I tried for several hours now to reconstruct the exact scenario for the crash with a cucumber scenario to no avail. I don't think there is any point in spending more time on this - the circumstances are so specific it is hard to image that something else can trigger this again.

include/engine/datafacade/shared_datafacade.hpp
```diff
@@ -276,11 +276,6 @@ class SharedDataFacade final : public BaseDataFacade
 
     void LoadCoreInformation()
     {
-        if (data_layout->num_entries[storage::SharedDataLayout::CORE_MARKER] <= 0)
-        {
-            return;
-        }
-
```
MoKob: Is this general clean-up? Was that related to the bug? I am missing the context here on why this was here in the first place.

include/engine/routing_algorithms/routing_base.hpp
MoKob: looks like this comment is obsolete, since you are using the `was inserted` to check the data.

include/engine/routing_algorithms/routing_base.hpp
```diff
@@ -693,34 +702,37 @@ template <class DataFacadeT, class Derived> class BasicRoutingInterface
...
         last_id = SPECIAL_NODEID;
         for (const auto &p : reverse_entry_points)
         {
-            if (p.first == last_id)
-            {
-                continue;
-            }
-            reverse_core_heap.Insert(p.first, p.second, p.first);
-            last_id = p.first;
+            insertInCoreHeap(p, reverse_core_heap, last_id);
```
MoKob: It looks to me the last_id usage here is essentially a no-op. Have you actually encountered any duplicates here? If that were the case, it would point to a serious bug in our routing code. Given [these lines](https://github.com/Project-OSRM/osrm-backend/blob/1ba8b5d8b7443589e9b12802a2ef06640ca4783c/include/engine/routing_algorithms/routing_base.hpp#L662-L666), we should never see any node more than once here.
The correctness of Dijkstras algorithm, which is also used in a CH graph is based on looking at every minimal node exactly once.

include/engine/datafacade/shared_datafacade.hpp
TheMarex: This was the bug that was causing the seg fault in the tests. Swapping non-core enabled datasets after core-enabled datasets caused the m_is_core_node vector not to be updated, causing invalid memory access. We could break that out as own PR but it is quite atomic and a requirement anyway.

Since this is quite an edge case that we basically only will run in when we do testing, I don't think this is a critical bug on its own.

include/engine/routing_algorithms/routing_base.hpp
TheMarex: Good point. I can check the commit history on this, but I'm sure we added this later on because we were running into a bug regarding this. That might just have been a side-effect of some other problem long resolved.

If we don't even need to sort here that would be awesome - saves quite a bit overhead. I can remove it and we can see if the tests still pass.

include/engine/routing_algorithms/routing_base.hpp
```diff
@@ -524,7 +528,11 @@ template <class DataFacadeT, class Derived> class BasicRoutingInterface
                                           std::vector<NodeID> &packed_path) const
     {
         NodeID current_node_id = middle_node_id;
-        while (current_node_id != search_heap.GetData(current_node_id).parent)
+        // all initial nodes will have itself as parent
+        // all incomplete paths (e.g. paths ending at the core boundary) will have SPECIAL_NODEID
+        // as parent
+        while (current_node_id != search_heap.GetData(current_node_id).parent &&
+               search_heap.WasInserted(search_heap.GetData(current_node_id).parent))
```
MoKob: I don't get why we need this two cases. Was there anything breaking when adding the core nodes with themselves as a parent? 

From looking at [this](https://github.com/Project-OSRM/osrm-backend/blob/cb57ccacefa59321b9a921a760fa70b08d49d30d/include/engine/routing_algorithms/routing_base.hpp#L712) we where using the node without a dedicated parent. So while unpacking the path, we should stop at each core node [here](https://github.com/Project-OSRM/osrm-backend/blob/cb57ccacefa59321b9a921a760fa70b08d49d30d/include/engine/routing_algorithms/routing_base.hpp#L527)

I fail to see how this modification actually changes something, other than adding to the `if` condition here. The results should just be the same?

MoKob:
https://github.com/Project-OSRM/osrm-backend/pull/2759#discussion_r76384150 got hidden by the last commit, still valid.

include/engine/routing_algorithms/routing_base.hpp
TheMarex: The code location that actually breaks is the loop code: https://github.com/Project-OSRM/osrm-backend/blob/e91bfebf237c3adfa26b8077377c54a5e3dd9d54/include/engine/routing_algorithms/routing_base.hpp#L93-L94

If you mark the core nodes like we do with real start nodes (node.parent == node), we trigger the loop logic for them as well. If for a core node this finds a shorter path that we never wanted to search for,  this will lead to a crash in the path unpacking as it will return a path that is (middle, middle)  but middle is not the forward for backward node of the source phantom.

Hence we need to have a way to differentiate between real start nodes (for which we want to allow loops) and for core start nodes (for which we don't want loops). 

I can dropbox you the data file and breaking query as it is proprietary data.

MoKob: Than we still have a problem. When we have two core nodes for the start/end, we still need to find loops that have been contracted. Or can't this ever happen? I guess it is a rare case, but the fix right now introduces a potential problem here.

Loops should only be searched for on `distance < 0` or if forced. So it looks like we might force loops at the wrong location?
Or could this `distance < 0` result from via-routes?

I guess since the map-matchings are the bigger problem over the loops on core nodes, we can merge this. You should probably add some documentation in the code to why this is important to do as it is and open an issue for core-node loops referencing these changes.

MoKob:
Capturing from the hidden conversation: requires a comment on the core search that explains the interaction between parent-nodes and loops for the core search. Next to that we need to investigate loops on core-nodes (additional issue?).

Apart from this: rebase,squash,merge.

TheMarex:
>  Next to that we need to investigate loops on core-nodes (additional issue?).

This should still work, since for start nodes the parent will still be the node itself and the logic for enabling loops applies just like before.

Logic flow is as follows:

1. all start nodes get inserted in CH heap
2. if a start node is part of the core it will get removed before it gets relaxed by the dijkstra search
3. the core start nodes get inserted in the core heap with parent == node
4. the loop logic applies to all nodes with parent == node, even in the core heap

Let me know where you see the problem in this logic flow.

MoKob:
You are right. Since the parent is set correctly in the first heap, it should be fine. This logic flow should be documented in the code, though, since it is really hard to follow these implications.

PR #2765, opened 2016-08-12
Allow Services to support several API versions simultaneously

niemeier-PSI:
I changed the way API-Version compatiblity is checked when parsing requests.

**Old:**
The service reports the version it supports, the service_handler checks if this number matches the version number from the query.

**New:**
The service handler passes the required version number to service. The service then decides if this version is supported or not.

This has the advantage of being able to support multiple versions. 

E.g.  for the API-Changes made in https://github.com/Project-OSRM/osrm-backend/pull/2764, one could increase the API Version to 2, but have all services support both API-Versions 1 and 2, this way being fully backward compatible.

### DIFF
--- a/include/server/service/base_service.hpp
+++ b/include/server/service/base_service.hpp
@@ -28,7 +28,7 @@ class BaseService
     virtual engine::Status
     RunQuery(std::size_t prefix_length, std::string &query, ResultT &result) = 0;
 
-    virtual unsigned GetVersion() = 0;
+    virtual bool SupportsVersion(unsigned version) = 0;
 
   protected:
     OSRM &routing_machine;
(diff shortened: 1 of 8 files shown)


TheMarex:
Thanks for breaking out this change! :+1: However we never intended to support multiple API versions in the same release. The version is used for two things:

1. Enable easy HTTP based forwarding: `route/v1/..` can be easily forward to an OSRM 5.x instance, `route/v2/...` hits the 6.x endpoint
2. Explicitly notify the developer of breaking changes. Since breaking changes require intervention on the client side, we prefer an explicit failure (e.g. API version not found) over a silent failure in the client application.

PR #2769, opened 2016-08-15
Restore json logging

habibutsu:
In previous version 4.x was ability to compile ORSM with debug logging to JSON responses. In 5.x this feature was not moved although the flag **-DENABLE_JSON_LOGGING** is still exist. This pull-request restore json-logging for map-matching

### DIFF
--- a/include/engine/routing_algorithms/map_matching.hpp
+++ b/include/engine/routing_algorithms/map_matching.hpp
@@ -10,6 +10,7 @@
 #include "util/coordinate_calculation.hpp"
 #include "util/for_each_pair.hpp"
 #include "util/json_logger.hpp"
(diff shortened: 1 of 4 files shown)


TheMarex:
Thanks for catching the unused cmake option. However I dropped the debug response since I felt it was not needed anymore. Any particular use-case that you need to support?

habibutsu:
I am started working with this tools is not so long ago and it was just interesting to visualise how map-matching algorithm works. Yes, I not see any practical use-case for that.

daniel-j-h:
Thanks, but this is not actionable for us. Feel free to use it for experimentation.
In fact, I might rip out the dead code for the json logger in addition from master.

Issue #1318, opened 2014-12-20
turn penalty vs traffic signal penalty


kaligrafy:
traffic signal penalty is calculated with a factor 10 (since distances are calculated in 1/10 seconds in the code)
However, turn penalties in turn function are not multiplied by a factor 10. Maybe they should be to avoid ambiguity?

MoKob:
I've prepared a small fix here: https://github.com/Project-OSRM/osrm-backend/pull/2773

TheMarex:
This will be addressed by #1614 since the `turn_function` API changed.

TheMarex:
This was fixed.

PR #2773, opened 2016-08-16
Fix/1318

MoKob:
Penalties in the profile are currently specified in seconds, except for the turn penalty.

This PR unifies this to specify seconds in general for all time-based penalty values.

/cc @karenzshea if this gets merged, we need to adjust our internal profiles

### DIFF
--- a/src/extractor/scripting_environment_lua.cpp
+++ b/src/extractor/scripting_environment_lua.cpp
@@ -363,7 +363,9 @@ int32_t LuaScriptingEnvironment::GetTurnPenalty(const double angle)
                 luabind::call_function<double>(context.state, "turn_function", angle);
             BOOST_ASSERT(penalty < std::numeric_limits<int32_t>::max());
             BOOST_ASSERT(penalty > std::numeric_limits<int32_t>::min());
-            return boost::numeric_cast<int32_t>(penalty);
+            // OSRM operates in 10ths of a second. Penalties are specified in seconds in the
+            // profile. Here we adjust this accordingly.
+            return boost::numeric_cast<int32_t>(10 * penalty);
         }
         catch (const luabind::error &er)
         {
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
     - Profiles
       - includes library guidance.lua that offers preliminary configuration on guidance.
       - added left_hand_driving flag in global profile properties
+      - refactor of turn penalties: turn penalties are now specified in seconds, just as all other penalties. Previously tenths of a second.
     - Guidance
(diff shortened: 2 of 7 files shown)


TheMarex:
Problem is that his is a breaking API change since the `turn_function` now returns seconds. You could make it API compatible by multiplying by 10 in the `turn_function`.

MoKob:
That is the whole point of this PR not to do. We have all penalties/times in seconds, only turn-penalty in deci-seconds. If you feel that we should keep this inconsistency, you can re-open 1318 and ignore this PR. Having inconsistent units in the profile feels wrong, though.

src/extractor/scripting_environment_lua.cpp
```diff
@@ -363,7 +363,9 @@ int32_t LuaScriptingEnvironment::GetTurnPenalty(const double angle)
                 luabind::call_function<double>(context.state, "turn_function", angle);
             BOOST_ASSERT(penalty < std::numeric_limits<int32_t>::max());
             BOOST_ASSERT(penalty > std::numeric_limits<int32_t>::min());
-            return boost::numeric_cast<int32_t>(penalty);
+            // OSRM operates in 10ths of a second. Penalties are specified in seconds in the
+            // profile. Here we adjust this accordingly.
+            return boost::numeric_cast<int32_t>(10 * penalty);
```
TheMarex: Move this multiplication to `turn_function`.

TheMarex:
I think the problem was more that the local variable was specified in deci-seconds which is a usability bug. Changing the unit of this variable is fine: It is not part of the external API.

MoKob:
either I don't understand where you want to move the variable, or you are missing my point. If I should move the `10` back to the profile, I would revert the entire point of this PR.

The profile specifies: traffic signal penalty in seconds, u turn penalty in seconds, turn penalty in deciseconds.

All other penalties are adjusted just the same way in the [profile properties](https://github.com/Project-OSRM/osrm-backend/blob/master/include/extractor/profile_properties.hpp#L19-L31). 

So to me having it in the `turn function` (which is in the profile) is a usability bug to me, since it does not follow the same conventions as all other timings specified in the profile.

TheMarex:
In https://github.com/Project-OSRM/osrm-backend/pull/2399 we already address the problem of the turn function being inconsistent. We will include this with the next major version bump.

We can't merge your current implementation since it would require a major API bump, which I don't think is warranted for a small change like this. However you are right, we should state the local variable `turn_penalty` in seconds. Hence my suggestion was to move the multiplication by 10 from the EdgeBasedGraphFactory to the `turn_function` in the lua profile.

I'm going to prepare a PR that implements this, so I think we can close this, as the other PR will not directly address issue #1318 but make the profiles easier to read.

Issue #2786, opened 2016-08-19
Profiles state turn penalties in deci-seconds


TheMarex:
Currently all profiles state `turn_penalty` in deci-seconds. That is confusing and should be fixed.

Related to #1318. But this is about a profile-level change only, that does not impact the API.

daniel-j-h:
This landed with https://github.com/Project-OSRM/osrm-backend/pull/2787.

PR #2787, opened 2016-08-19
[ready] State all turn penalties in seconds

TheMarex:
# Issue

This addresses https://github.com/Project-OSRM/osrm-backend/issues/2786

## Tasklist
 - [x] review
 - [x] adjust for for comments

## Requirements / Relations

None.

### DIFF
--- a/profiles/bicycle.lua
+++ b/profiles/bicycle.lua
@@ -99,7 +99,7 @@ properties.continue_straight_at_waypoint = false
 
 local obey_oneway               = true
 local ignore_areas              = true
-local turn_penalty              = 60
+local turn_penalty              = 6
 local turn_bias                 = 1.4
 -- reduce the driving speed by 30% for unsafe roads
 -- local safety_penalty            = 0.7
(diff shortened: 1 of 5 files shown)


profiles/bicycle.lua
```diff
@@ -402,7 +402,7 @@ end
 
 function turn_function (angle)
   -- compute turn penalty as angle^2, with a left/right bias
-  k = turn_penalty/(90.0*90.0)
```
MoKob: given that this is technically a bug in the API, we should add a comment here stating where the factor 10 comes from.

profiles/car.lua
```diff
@@ -524,7 +524,7 @@ end
 
 function turn_function (angle)
   ---- compute turn penalty as angle^2, with a left/right bias
-  k = turn_penalty/(90.0*90.0)
+  k = 10*turn_penalty/(90.0*90.0)
```
MoKob: see above.

MoKob:
Apart from the comments (possibly including a link to the issue), this looks good to me.

PR #2796, opened 2016-08-22
Add local maxspeed for Netherland

frodrigo:
Report change from MAPS.ME mapsme/omim@d6cd9b77a8a99afae2ca65e40c11d3e4e51e1691

@Zverik on MAPS.ME side you don't need the line ["nl:motorway"] = 130, as 130 is the default for motorway.

### DIFF
--- a/profiles/car.lua
+++ b/profiles/car.lua
@@ -130,6 +130,8 @@ maxspeed_table = {
   ["uk:nsl_single"] = (60*1609)/1000,
   ["uk:nsl_dual"] = (70*1609)/1000,
   ["uk:motorway"] = (70*1609)/1000,
+  ["nl:rural"] = 80,
+  ["nl:trunk"] = 100,
   ["none"] = 140
 }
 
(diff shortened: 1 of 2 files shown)


Zverik:
Thank you! Could you also add these lines to [the upstream](https://github.com/Project-OSRM/osrm-backend/blob/master/profiles/car.lua#L110)?

Zverik:
Good to see the change is being pushed to the upstream. Sorry we forgot to do this. Thanks, Frédéric.

TheMarex:
@frodrigo thanks for upstreaming this. :+1: Could you add the tags to `taginfo.json`? Otherwise this will be good to merge.

frodrigo:
It is the `maxspeed` tag and it's obviously already in the `taginfo.json`

TheMarex:
@frodrigo https://github.com/Project-OSRM/osrm-backend/blob/master/taginfo.json#L112 we also add the (non-number) values we support explicitly. See for example https://github.com/Project-OSRM/osrm-backend/blob/master/taginfo.json#L112

frodrigo:
done

Issue #2798, opened 2016-08-22
Overflow in via-routes with u-turns enabled


MoKob:
In the via-route plugin, we check at https://github.com/Project-OSRM/osrm-backend/blob/master/include/engine/routing_algorithms/shortest_path.hpp#L380-L381 if an invalid edge-weight is found.
This is the case when there is no route between points.
While the search without u-turns correctly returns `invalid_edge_weight`, the search with u-turn enabled does not.
At  https://github.com/Project-OSRM/osrm-backend/blob/master/include/engine/routing_algorithms/shortest_path.hpp#L117 will not be invalidated.

PR #2799, opened 2016-08-22
[ready] fix segmentation fault for via-routes with invalid intermediate segments

MoKob:
# Issue

Fixes https://github.com/Project-OSRM/osrm-backend/issues/2798

## Tasklist
 - [x] review
 - [x] adjust for for comments

## Requirements / Relations

none

### DIFF
--- a/include/engine/routing_algorithms/shortest_path.hpp
+++ b/include/engine/routing_algorithms/shortest_path.hpp
@@ -114,7 +114,11 @@ class ShortestPathRouting final
                           needs_loop_forwad,
                           needs_loop_backwards);
         }
-        new_total_distance += std::min(total_distance_to_forward, total_distance_to_reverse);
+        // if no route is found between two parts of the via-route, the entire route becomes
+        // invalid. Adding to invalid edge weight sadly doesn't return an invalid edge weight. Here
+        // we prevent the possible overflow, faking the addition of infinity + x == infinity
+        if (new_total_distance != INVALID_EDGE_WEIGHT)
+            new_total_distance += std::min(total_distance_to_forward, total_distance_to_reverse);
     }
 
     // searches shortest path between:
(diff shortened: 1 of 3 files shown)


include/engine/routing_algorithms/shortest_path.hpp
```diff
@@ -114,7 +114,11 @@ class ShortestPathRouting final
                           needs_loop_forwad,
                           needs_loop_backwards);
         }
-        new_total_distance += std::min(total_distance_to_forward, total_distance_to_reverse);
+        //if no route is found between two parts of the via-route, the entire route becomes invalid
+        if( new_total_distance == INVALID_EDGE_WEIGHT )
```
TheMarex: Is this a typo? Shouldn't it by `total_distance_to_forward == INVALID_EDGE_WEIGHT ||  total_distance_to_reverse == INVALID_EDGE_WEIGHT`?

MoKob: That was my first intuition as well. It turns out, the way it is used here is different. If you have a look at the function header, the two weights are constants passed down here. New total distance on the other hand is used to pass down to the search functions as distance value.

This could probably use a serious refactor for better names, but I didn't want to mix this up with the actual fix here.

TheMarex: Ah okay, but the line below then `new_total_distance = INVALID_EDGE_WEIGHT;` is basically just a noop?

It might be easier to read if we just check `if( new_total_distance != INVALID_EDGE_WEIGHT )` and then execute the current else branch?

MoKob: you are right, this is a remainder from my initial approach using `total_distance_to_forward == INVALID_EDGE_WEIGHT || total_distance_to_reverse == INVALID_EDGE_WEIGHT`

TheMarex:
Looks good to me. :+1:

TheMarex:
Ah a side note, what do you think about pushing out a `5.3.3` release with this?

MoKob:
I don't think that a `5.3.3` release with this is urgent, since it is more like a border case. Via route, u-turns on and invalid segment in-between. If you feel that a release could still help out, feel free to roll one.

PR #2800, opened 2016-08-22
[ready] Don't use --gc-sections gold option for binutils < 2.25

### LINKED ISSUE #2785: Gold linker and shared libraries
When using gold and shared libraries in https://github.com/Project-OSRM/osrm-backend/pull/2601 together this yields runtime crashes.

I'm not sure which optimization is the cause of that. Maybe we have some UB somewhere.

oxidase:
# Issue

Pull request fixes #2785. The problem was in gold DCE option --gc-sections that removes .gcc_except_table. Osmium fails because _Unwind_Find_FDE fails to find frame description entries that were eliminated.  [Workaround](https://sourceware.org/ml/binutils/2013-01/msg00346.html) could be to mark .gcc_except_table as KEEP, but it is too specific for travis gold version.

## Tasklist
 - [x] Find reason for crash at third_party/libosmium/include/osmium/io/detail/pbf_input_format.hpp:85 for ENABLE_GOLD_LINKER=ON AND BUILD_SHARED_LIBS=ON
 - [x] review
 - [x] adjust for for comments

## Requirements / Relations
 Related PR #2601


CMakeLists.txt
```diff
@@ -220,7 +220,7 @@ execute_process(COMMAND ${CMAKE_CXX_COMPILER} "-Wl,--version" ERROR_QUIET OUTPUT
 # For ld.gold and ld.bfs (the GNU linkers) we optimize hard
 if("${LINKER_VERSION}" MATCHES "GNU gold" OR "${LINKER_VERSION}" MATCHES "GNU ld")
   message(STATUS "Setting linker optimizations")
-  if(NOT ${CMAKE_CXX_COMPILER_ID} STREQUAL "MSVC")
+  if(NOT (${CMAKE_CXX_COMPILER_ID} STREQUAL "MSVC" OR ${ENABLE_GOLD_LINKER} AND ${BUILD_SHARED_LIBS}))
```
danpat: @oxidase Mind throwing some `()` around the `AND` statement there?  I always get nervous about order of operations when it's not explicit.

oxidase: ok, added

TheMarex:
Looks good to me. I think adding a link to the issue in the CMakeList.txt would help to understand why we need this switch. Thanks for digging here!

daniel-j-h:
Awesome find! @oxidase could you describe a bit more detailed how you debugged this problem? As in, how did you figure out the section the gc flag was removing but libosmium needed?

oxidase:
@daniel-j-h to find the reason, i did in my branch on travis
* added in .travis.yml  `gdb -batch -ex "run" -ex "info thread" -ex "thread apply all bt full" -ex "fr 9"  -ex "info local" -ex "info arg" --args ./build/osrm-extract test/data/monaco.osm.pbf -p ./profiles/car.lua` and get that _Unwind_Find_FDE failed, but was not sure why
* bisected `--gc-sections` flag on travis between Debug and Release flags in CMakeLists.txt
```
set(CMAKE_EXE_LINKER_FLAGS    "-fuse-ld=gold -Wl,--disable-new-dtags -Wl,-O1 -Wl,--hash-style=gnu -Wl,--sort-common -Wl,--gc-sections -Wl,--print-gc-sections")
set(CMAKE_SHARED_LINKER_FLAGS "-fuse-ld=gold -Wl,--disable-new-dtags -Wl,-O1 -Wl,--hash-style=gnu -Wl,--sort-common -Wl,--gc-sections -Wl,--print-gc-sections")
```
* searched for bugs with "gold gc-sections eh_frame", found [bug](https://sourceware.org/bugzilla/show_bug.cgi?id=17639) and checked  `readelf --hex-dump=.eh_frame --debug-dump=frames build/libosrm_extract.so`  from the bug description

daniel-j-h:
Thanks for the explanation and the detailed answer Michael. Highly appreciated.

Issue #2804, opened 2016-08-23
Build breaks on current CentOS


rkcpi:
Currently the build fails on CentOS (and seemingly Ubuntu, though I haven't confirmed that myself), because CentOS comes with gcc 4.8.5 which doesn't support C++14.

The error encountered is the following:
```
CMake Error at cmake/check_luabind.cmake:38 (message):
Luabind does not work with Lua 5.1 at
/usr/lib/x86_64-linux-gnu/liblua5.1.so, no working Luabind found
```

Also see the discussion regarding this in 71d64e8cd9892beb1220511323b5459aa9809645.

james-gould:
Have you tried any fixes in #1196? This may be of help: https://github.com/Project-OSRM/osrm-backend/issues/1196#issuecomment-97706794

rkcpi:
@james-gould the problem is that there's no c++14 support in CentOS's gcc. luabind worked fine all along, the error message does not tell us the real problem

i already opened a pull request #2805 that fixes this issue

pongo710:
I think this issue can be closed. The referenced pull request was merged.

PR #2805, opened 2016-08-23
Fix build on CentOS

rkcpi:
# Issue

Fixes build on CentOS systems which doesn't have gcc ready for C++14 yet.

Fixes issue #2804


TheMarex:
Doesn't `c++1y` mean `c++11` on older compilers? We switched to C++14 a while ago and just didn't get around to clean up the legacy code that backported some 14 functionality (e.g. http://en.cppreference.com/w/cpp/memory/unique_ptr/make_unique)

MoKob:
From what I've read `c++1y` is the respective draft of `c++14` just like `c++11` was referred to by `c++0x`. Using a draft only does not seem a valid step in our progression to me.

rkcpi:
This is a change proposed by @daniel-j-h yesterday in irc:
```
11:56:40 <   danieljh> rkcpi: if it works fine with "c++1y" instead can you open a pull request with that change and see how travis likes it?
```
Is there another valid approach to still be able to build with gcc on CentOS?

TheMarex:
Merging here. Note that on newer compiler `c++1y` is just an alias for `c++14` so it doesn't change anything for our current setup.

PR #2821, opened 2016-08-30
[Ready] Adds a failing test case for 255 unicode code points

daniel-j-h:
For @karenzshea:

the OSM spec says "up to 255 full unicode chars", so we should be able
to handle 255 x 🐼 --- but we don't.

The internal limit is on 255 _bytes_ in the `std::string`, but there can
be up to four bytes in a utf8 encoded code point.

References:
- http://wiki.openstreetmap.org/wiki/API_v0.6#Tags
- http://www.fileformat.info/info/unicode/char/1f43c/browsertest.htm
- https://en.wikipedia.org/wiki/UTF-8#Description
- https://github.com/Project-OSRM/osrm-backend/blob/807aa71a7c80b714b7a89db9c99c15720572ba25/include/util/range_table.hpp#L28


daniel-j-h:
We're currently returning
```
ab,🐼🐼🐼🐼🐼🐼🐼🐼🐼🐼🐼🐼🐼🐼🐼🐼🐼🐼🐼🐼🐼🐼🐼🐼🐼🐼🐼🐼🐼🐼🐼🐼🐼🐼🐼🐼🐼🐼🐼🐼🐼🐼🐼🐼🐼🐼🐼🐼🐼🐼🐼🐼🐼🐼🐼🐼🐼🐼🐼🐼🐼🐼🐼���,🐼🐼🐼🐼🐼🐼🐼🐼🐼🐼🐼🐼🐼🐼🐼🐼🐼🐼🐼🐼🐼🐼🐼🐼🐼🐼🐼🐼🐼🐼🐼🐼🐼🐼🐼🐼🐼🐼🐼🐼🐼🐼🐼🐼🐼🐼🐼🐼🐼🐼🐼🐼🐼🐼🐼🐼🐼🐼🐼🐼🐼🐼🐼��� |
```

MoKob:
Well, currently this actually fails. So I guess not-ready?

daniel-j-h:
Ha yeah, added a todo tag, disabling the check. Will ticket this.

Issue #1735, opened 2015-10-12
U-turn where dual carriageway merges to single carriageway?


PinkDuck:
e.g. from 52.673645,0.957841 to 52.672091,0.945735 (https://goo.gl/maps/7ufJJ2FC7tC2)

While legal to make a U-turn in the above instance, it would involve slowing from 70 mph to <5 mph to cross wide hatchings then accelerate back up to 70 mph. A rather dangerous thing to do, yet the chosen route above from the public web demo.

Does the route timing currently factor in the need to slow down to make such a sharp turn?

Should such high back angle turns on high speed road simply be avoided through a separate penalty or implicit restriction?

Yes, I could add a no U-turn restriction to OSM, but it is legally possible to do it as exemplified by the emergency services.

danpat:
We do have a turn penalty included in the `car.lua` profile now:

https://github.com/Project-OSRM/osrm-backend/blob/develop/profiles/car.lua#L439-L447

But it does not take into account the current speed.  You're right, a U-turn at the location above looks like a bad idea.

We should probably think about factoring the speed values into the turn penalty for cases like this.

TheMarex:
> but it is legally possible to do it as exemplified by the emergency services.

You can also add exceptions for this case http://wiki.openstreetmap.org/wiki/Relation:restriction I'm not sure about traffic laws in England but this definitely looks like it should not be allowed for general traffic.

PinkDuck:
The England traffic laws permit u-turns anywhere except where explicitly restricted by signage. Just that most drivers would realise that it is too risky to consider under normal driving. I could add restrictions but then this is quite a universal thing for any near 180 degree turn from 70 mph to 70 mph speed limit, so worth considering as a system factor.

Issue #592, opened 2013-02-15
penalty for mode change & crossing roads?


emiltin:
it could be nice to add a penalty when you change mode, for example when change between pushing and riding a bike.

it would also be very nice to add a penalty for crossing big road without signals.

both could perhaps be handled by passing info about the from/to way, and the via node, to a lua turn penalty function. the penalty (if any) could then be based not only the angle between the two ways, but also whether the mode changes, size of ways, etc, signal or not on the via node, etc.

daniel-j-h:
@MoKob started to experiment with the approach described here in https://github.com/Project-OSRM/osrm-backend/pull/2912 - bringing all these features together will allow for a better turn penalty.

PR #2822, opened 2016-08-30
Pass additional variables to turn_function

danpat:
This PR is to address https://github.com/Project-OSRM/osrm-backend/issues/1735, and hopefully make turn penalties better by considering the approach/exit speeds for each turn.  As pointed out on #1735, doing a u-turn on a 70mph (112km/h) road should be super expensive.  On the car profile, currently we only penalize this turn by 3.3 seconds, and most 60km/h 90 degree turns are penalized by ~0.8 to ~1.2 (left vs right, depending on the bias).

This turn penalty calculation only occurs during `osrm-extract`, this PR will not include a mechanism to adjust the turn penalty if traffic data updates are used in `osrm-contract`.

My plan is to update the `turn_function` for the `car.lua` profile only.  Other profiles are probably a bit less affected by this, although @emiltin, any thoughts you have on how this might apply to the bike turn penalty would be appreciated.

## Tasklist
 - [ ] Test coverage
 - [ ] Do some test driving to get some better values for the car profile
 - [ ] Do some test riding?
 - [ ] review
 - [ ] adjust for for comments

## References
http://nacto.org/docs/usdg/effect_of_radius_of_curvature_for_right_turning_vehicles_wolfe.pdf
http://onlinepubs.trb.org/onlinepubs/conferences/2011/RSS/3/Wolfermann,A.pdf
http://www.istiee.org/te/papers/N55/ET_2013_55_1_Mehar.pdf

# Other useful factors
https://github.com/Project-OSRM/osrm-backend/issues/592

### DIFF
--- a/src/extractor/edge_based_graph_factory.cpp
+++ b/src/extractor/edge_based_graph_factory.cpp
@@ -423,8 +423,38 @@ void EdgeBasedGraphFactory::GenerateEdgeExpandedEdges(
                     distance += profile_properties.traffic_signal_penalty;
                 }
 
+                const bool isTrivial = m_compressed_edge_container.IsTrivial(edge_from_u);
+
+                const auto &approach_node =
+                        isTrivial
+                            ? m_node_info_list[node_u]
+                            : m_node_info_list[m_compressed_edge_container.GetLastEdgeSourceID(
+                                  edge_from_u)];
+                const auto &intersection_node =
+                        m_node_info_list[m_compressed_edge_container.GetLastEdgeTargetID(
+                            edge_from_u)];
+
+                const auto &exit_node =
+                        m_node_info_list[m_compressed_edge_container.GetFirstEdgeTargetID(
+                            turn.eid)];
+
+                const double approach_segment_length =
+                            util::coordinate_calculation::greatCircleDistance(approach_node, intersection_node);
+                const double exit_segment_length =
+                            util::coordinate_calculation::greatCircleDistance(intersection_node, exit_node);
+
+                const auto approach_road_segments =
+                    m_compressed_edge_container.GetBucketReference(edge_from_u);
+                const auto exit_road_segments =
+                    m_compressed_edge_container.GetBucketReference(turn.eid);
+
+                // segment_length is in metres, weight is in deciseconds, this converts those
+                // to km/h
+                const double approach_speed = approach_segment_length / (approach_road_segments.back().weight/10) * 3.6;  // km/h
+                const double exit_speed = exit_segment_length / (exit_road_segments.front().weight/10) * 3.6;  // km/h
+
                 const int32_t turn_penalty =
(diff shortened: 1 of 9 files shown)


TheMarex:
Few notes here:

- the current implementation breaks the API and would require a major version bump. (not sure how lua feels about optional arguments but that could be an answer)
- keep in mind #1614 modfies the signature of turn_function drastically

danpat:
@TheMarex AFAIK, Lua has no concept of function overloading, so the following works:

```
$ lua5.1
Lua 5.1.5  Copyright (C) 1994-2012 Lua.org, PUC-Rio
> function test(a)
>>   print "hello"
>> end
> test(1,2)
hello
```

The additional arguments passed by `luabind` to the function should just be ignored by older profiles that don't implement them.  This should only require be a backward-compatible minor version bump.

src/extractor/edge_based_graph_factory.cpp
```diff
@@ -423,8 +423,38 @@ void EdgeBasedGraphFactory::GenerateEdgeExpandedEdges(
...
+
+                const auto approach_road_segments =
+                    m_compressed_edge_container.GetBucketReference(edge_from_u);
+                const auto exit_road_segments =
+                    m_compressed_edge_container.GetBucketReference(turn.eid);
+
+                // segment_length is in metres, weight is in deciseconds, this converts those
+                // to km/h
+                const double approach_speed = approach_segment_length / (approach_road_segments.back().weight/10) * 3.6;  // km/h
+                const double exit_speed = exit_segment_length / (exit_road_segments.front().weight/10) * 3.6;  // km/h
```
daniel-j-h: clang-format-3.8 all the things

emiltin:
I think turn penalties for bikes could be improved quite a bit. however i think the type of road your turning from/to might be more important that the approach/exit speeds. The typical example is that turning left form a big road into a small road at a intersection without traffic lights should incur a much more significant penalty that turning left from a small road into a small road, since you often will have to wait for a gap in car traffic before you can turn.

Similarly, the penalty for doing a u-turn depends more on the type of road that the speed of the bike traffic, since doing a u-turn typically implies crossing the road.

This is why is suggested in #592 that the way type are passed to the penalty function. An alternative is to only depend a real-world measurements, but of course that requires sufficient data.

Turn penalties for bikes could perhaps be improved a bit by using approach/exit speeds, but I don't think it's worth it.

danpat:
Some further discussion on this one has happened in other forums. Some points that were raised:

  - it might be useful to apply no penalty when the turn instruction is "continue" - implying that the travellers is staying on the same road and likely does not need to slow down, even if the road curves at the intersection.
  - does the turn happen in across non-stopping traffic, and if so, how many lanes - this would be useful for both cars and bikes

MoKob:
I've pulled this into https://github.com/Project-OSRM/osrm-backend/pull/2912. Closing here in favour of a structure based approach (instead of passing single parameters to the function).

Issue #2824, opened 2016-08-31
Broken Testcase / Staggered Intersections might need revisiting


MoKob:
During my ongoing release of 5.3.3 I came across the following testcase:

https://github.com/Project-OSRM/osrm-backend/blob/da9d63a3845eced8c648a43ec6841611cb377274/features/testbot/via.feature#L266-L279

The cucumber scenario currently succeeds on master, but fails on 5.3.3. It should fail on both versions though, or is not helpful. What the test suggests is visible in the following picture:

![vis-invalid](https://cloud.githubusercontent.com/assets/12932279/18122330/19db21ec-6f69-11e6-90b2-3990764897e3.png)

We query for the red path, which is obviously inferior to the green path. The green path is found (which would result in `ebbdcf, abcd, ebbdcf, ebbdcf`, but is collapsed again to `ebbdcf,ebbdcf` due to `https://github.com/Project-OSRM/osrm-backend/blob/da9d63a3845eced8c648a43ec6841611cb377274/src/engine/guidance/post_processing.cpp#L789` triggering here.

For the release of 5.3.3 I am marking this case as a todo. We should revisit how we want to handle this scenario and if the staggered intersection heuristic is valid. /cc @daniel-j-h

MoKob:
@daniel-j-h since you did the implementation of staggered-intersections, could you have a look at them and see that they don't trigger in invalid situations?

daniel-j-h:
From chat: a more robust left-right / right-left heuristic for staggered intersections has to directly take a look at the turn angles. This is due to data issues (like the one described above), also turn modifiers are not a reliable way here, we want to be more stable and precisely classify staggered intersections as such.

MoKob:
Even though #2855 will increase the quality of our staggered intersection detection, we still need to investigate how to handle the test-case referenced here as it is still broken.
The problem here is that the result is confusing and only comes from the suppression of name changes. The actual path taken differs from the one shown in the test.

MoKob:
Update: Staggered Intersections have been improved. We still need to revisit the mentioned test for its goals,

PR #2825, opened 2016-08-31
mark broken case todo

MoKob:
# Issue

https://github.com/Project-OSRM/osrm-backend/issues/2824

## Tasklist
 - [x] review - the review should mainly target the issue. The PR itself is trivial
 - [x] adjust for for comments

## Requirements / Relations
none


daniel-j-h:
Pull request for disabling this is fine for now, but we need to fix the issue before the next release for sure. Commenting on the issue itself for more details.

Issue #2827, opened 2016-08-31
Ignore invalid `only_` restrictions


MoKob:
If we encounter a turn restriction that flags an `only_` turn, we forbid all turns that are not directly targeting the referred node.
In case of invalid restrictions, we disallow all turns at an intersection.
To be on the safe side, we should consider allowing all turns as long as the restriction is not valid.

PR #2828, opened 2016-08-31
[ready] ignore invalid only_* restrictions

MoKob:
# Issue

https://github.com/Project-OSRM/osrm-backend/issues/2827

## Tasklist
 - [x] review - done @daniel-j-h
 - [x] adjust for for comments

## Requirements / Relations

### DIFF
--- a/src/extractor/guidance/intersection_generator.cpp
+++ b/src/extractor/guidance/intersection_generator.cpp
@@ -55,8 +55,21 @@ Intersection IntersectionGenerator::GetConnectedRoads(const NodeID from_node,
 {
     Intersection intersection;
     const NodeID turn_node = node_based_graph.GetTarget(via_eid);
-    const NodeID only_restriction_to_node =
-        restriction_map.CheckForEmanatingIsOnlyTurn(from_node, turn_node);
+    const NodeID only_restriction_to_node = [&]() {
+        // If only restrictions refer to invalid ways somewhere far away, we rather ignore the
+        // restriction than to not route over the intersection at all.
+        const auto only_restriction_to_node =
+            restriction_map.CheckForEmanatingIsOnlyTurn(from_node, turn_node);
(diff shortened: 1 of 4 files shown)


src/extractor/restriction_map.cpp
```diff
@@ -143,11 +143,13 @@ bool RestrictionMap::CheckIfTurnIsRestricted(const NodeID node_u,
...
-        {
-            return true;
-        }
+        // The following check is actually perfomed in intersection generation where we can also
+        // verify if the only restriction is valid at all.
+        // if (node_w != restriction_target.target_node && // target not found
+        //    restriction_target.is_only)                 // and is an only restriction
+        // {
+        //    return true;
+        // }
```
daniel-j-h: Can this code be removed then?

src/extractor/guidance/intersection_generator.cpp
```diff
@@ -55,8 +55,26 @@ Intersection IntersectionGenerator::GetConnectedRoads(const NodeID from_node,
...
+            for (const EdgeID onto_edge : node_based_graph.GetAdjacentEdgeRange(turn_node))
+            {
+                BOOST_ASSERT(onto_edge != SPECIAL_EDGEID);
+                const NodeID to_node = node_based_graph.GetTarget(onto_edge);
+                if (to_node == only_restriction_to_node)
+                {
+                    return only_restriction_to_node;
+                }
+            }
+            // Ignore broken only restrictions.
```
daniel-j-h: Good catch :+1: looks good.

There are range algorithms for this pattern, too, as in

    boost::find_if(EdgeRange, [] (EdgeId e) { .. });

src/extractor/restriction_map.cpp
MoKob: Well I thought about this. I feel this comment is a good documentation to nod add this check again in the future. But we can also remove it.

Issue #2802, opened 2016-08-23
Don't add restrictions when we don't understand the tag


danpat:
In https://github.com/Project-OSRM/osrm-backend/issues/2801, @codain found a relation with `type=restriction` and `restriction=yield`, which is not a documented type.

However, our restriction parser treated this like a `restriction=no_turn`, and prevented turns from a slip-road onto a major highway.

The tagging is bad, but we should also not fail so badly when we hit a tag like this.

Currently, the restriction parser only looks for `type=restriction`, and then tests if it's `restriction=only*`.  If it's any other `restriction=*` value, we treat it like a `no_turn` and remove edges from the graph.

We should, at a minimum, check that the restriction is one of the values on the OSM wiki http://wiki.openstreetmap.org/wiki/Relation:restriction:

> If the first word is "no_", then no routing is possible from the "from" to the "to" member. If it is "only_", then you know that the only routing originating from the "from" member leads to the "to" member. 

We aren't checking for `no_*` at the moment, we should probably change that.

Codain:
Proposal:

It looks like the code is in the following function of the file `src/extractor/restriction_parser.cpp`:

```
boost::optional<InputRestrictionContainer>
RestrictionParser::TryParse(const osmium::Relation &relation) const
```

I think that changing from:

```
if (value.find("only_") == 0)
{
            is_only_restriction = true;
}
```

To:

```
if (value.find("only_") == 0)
{
            is_only_restriction = true;
}
else if (value.find("no_") == 0) // See the nota reguarding this check and whether to add it or not
{
            is_only_restriction = false;
}
else
{
            // Restriction not supported, ignore it
            return {};
}
```

Should make the trick (but I don't have a development environment to check).

Nota: The second check may increase an already existing side effect. Today in some circumstances (`restriction=no_*` lead to `false` and then `restriction:<*>=only_*` leads to `true`) the last restriction value of type `only_` will override the previous ones regarding the variable `is_only_restriction` (which will be set as `true`). With this patch it will be even more visible and the last restriction will always override `is_only_restriction`. To solve this already existing case will require some additional steps/a rewriting of the whole algorithm. But at a first step and to close this issue **ignoring the second if and adding only the else statement** will solve the issue without changing the existing behavior.

Hope this helps...

PR #2829, opened 2016-08-31
Turn restriction extraction, explicitly return false for no_* restrictions

karenzshea:
# Issue

Closes https://github.com/Project-OSRM/osrm-backend/issues/2802. This doesn't address the caveat that @codain brought up here, https://github.com/Project-OSRM/osrm-backend/issues/2802#issuecomment-241796627, but I think this would require rewriting the parser to be able to know what had been previously returned to the `InputRestrictionContainer` on the same relation. 

Also adds a test for a restriction `yield`, which is the undocumented restriction this was originally found on.

## Tasklist
 - [x] review
 - [ ] adjust for for comments

## Requirements / Relations
n/a

cc @danpat @MoKob

### DIFF
--- a/features/car/restrictions.feature
+++ b/features/car/restrictions.feature
@@ -430,3 +430,27 @@ Feature: Car - Turn restrictions
             | a    | b  | ax,xy,yb,yb |
             | b    | a  | yb,xy,ax,ax |
 
+    @specific
+    Scenario: Car - Ignore unrecognized restriction
+        Given the node map
+            |   | n |   |
+            | w | j | e |
+            |   | s |   |
+
+        And the ways
+            | nodes | oneway |
+            | sj    | yes    |
+            | nj    | -1     |
+            | wj    | -1     |
+            | ej    | -1     |
+
+        And the relations
+            | type        | way:from | way:to | node:via | restriction  |
+            | restriction | sj       | wj     | j        | yield        |
+
+        When I route I should get
+            | from | to | route    |
+            | s    | w  | sj,wj,wj |
+            | s    | n  | sj,nj,nj |
+            | s    | e  | sj,ej,ej |
+
--- a/src/extractor/restriction_parser.cpp
+++ b/src/extractor/restriction_parser.cpp
@@ -96,10 +96,13 @@ RestrictionParser::TryParse(const osmium::Relation &relation) const
         if (value.find("only_") == 0)
         {
             is_only_restriction = true;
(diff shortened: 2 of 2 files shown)


features/car/restrictions.feature
```diff
@@ -430,3 +430,27 @@ Feature: Car - Turn restrictions
...
+        And the ways
+            | nodes | oneway |
+            | sj    | yes    |
+            | nj    | -1     |
+            | wj    | -1     |
+            | ej    | -1     |
+
+        And the relations
+            | type        | way:from | way:to | node:via | restriction  |
+            | restriction | sj       | wj     | j        | yield        |
```
MoKob: can we also add tests that start with only_ stuff like `only_if_you_are_lucky` or `no_not_right_now` to make sure we only use recognised tags?

src/extractor/restriction_parser.cpp
```diff
@@ -91,10 +91,20 @@ RestrictionParser::TryParse(const osmium::Relation &relation) const
         const std::string key(fi_begin->key());
         const std::string value(fi_begin->value());
 
+        // documented OSM restriction tags start either with only_* or no_*;
+        // check and return on these values, and ignore unrecognized values
         if (value.find("only_") == 0)
```
MoKob: I feel we should go even further to check for known "only_" restrictions and known "no_" restrictions. Whats your feeling here?

features/car/restrictions.feature
daniel-j-h: But we don't right? All we care about is the `no_` /  `only_` prefix.

The actual constraint is based on the `fromWayId,viaNodeId,toWayId` ids.

src/extractor/restriction_parser.cpp
daniel-j-h: Ah yes, see above. Yes we could check for all listed restrictions and check taginfo to make sure we're not missing some.

features/car/restrictions.feature
MoKob: Well, `only_if_you_are_lucky` would pose a valid restriction. From chat, this is how OSM defines their turn restrictions, though. So we are fine here.

src/extractor/restriction_parser.cpp
MoKob: From chat: concern is invalid. This check follows OSM paradigms. Fine to merge.

Issue #2781, opened 2016-08-18
Pass options to profiles


TheMarex:
Currently all profiles are self-contained: All options are specified in the file itself. The only mechanism to pass options (say set toll-roads to true/false) is by modifying that file.

This becomes problematic if we want to test the behavior of a profile with changed options. For example it is not possible to test the behavior of `car.lua` if toll roads are disabled, if the default is to enable them.

For this to work we would need a mechanism to pass external variables to the lua profile. There are two implementation options I see:
- Use environment variables and directly query them from the lua profile (`os.getenv("OSRM_IGNORE_TOLLS")`)
- Implement a custom options parser à la cmake: `osrm-extract --variable=IGNORE_TOLLS:true`

danpat:
> Use environment variables and directly query them from the lua profile 

👍  on this approach.  It's simple to implement.

TheMarex:
The immediate need for this in the cucumber tests was removed via https://github.com/Project-OSRM/osrm-backend/pull/2954

daniel-j-h:
Can we close this, then?

We probably should open another ticket re. Cucumber tests default profile options toggled.

PR #2834, opened 2016-09-01
Allow passing options to lua profiles

karenzshea:
# Issue

Allows dynamic testing of profile changes by adding a lua function that will first look for a given variable in the run environment, and fallback to a given default.

## Tasklist
 - [x] Make regression / cucumber case pass
-  [ ] Add test for lua environment reading function
 - [ ] review
 - [ ] adjust for for comments

## Requirements / Relations
https://github.com/Project-OSRM/osrm-backend/issues/2781
https://github.com/Project-OSRM/osrm-backend/pull/2795

### DIFF
--- a/profiles/car.lua
+++ b/profiles/car.lua
@@ -3,6 +3,7 @@ local find_access_tag = require("lib/access").find_access_tag
 local get_destination = require("lib/destination").get_destination
 local set_classification = require("lib/guidance").set_classification
 local get_turn_lanes = require("lib/guidance").get_turn_lanes
+local set_option = require("lib/set_option")
 
 -- Begin of globals
 barrier_whitelist = { ["cattle_grid"] = true, ["border_control"] = true, ["checkpoint"] = true, ["toll_booth"] = true, ["sally_port"] = true, ["gate"] = true, ["lift_gate"] = true, ["no"] = true, ["entrance"] = true }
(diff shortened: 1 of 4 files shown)


karenzshea:
Going to wait on https://github.com/Project-OSRM/osrm-backend/pull/2795 before adding a new environment variable reading feature to our cucumber suite.

MoKob:
I added the dependency of the cucumber refactor to the PR.

daniel-j-h:
I guess we want to do this for all our flags and then add tests for both their true / false values.

profiles/car.lua
```diff
@@ -158,7 +159,7 @@ local turn_bias                  = properties.left_hand_driving and 1/1.2 or 1.2
 local obey_oneway                = true
 local ignore_areas               = true
 local ignore_hov_ways            = true
-local ignore_toll_ways           = false
+local ignore_toll_ways = set_option("ignore_toll_ways", false)
```
daniel-j-h: Looks like the `scripts/check-taginfo.py` does not like this ^ --- Travis failing.

daniel-j-h:
Related: @karenzshea https://github.com/Project-OSRM/osrm-backend/pull/2954#issuecomment-250004546

karenzshea:
Superseeded by https://github.com/Project-OSRM/osrm-backend/pull/2954!

Issue #2839, opened 2016-09-02
Turning Loops can emit invalid instructions


MoKob:
When following a turning circle along a road, the circle currently gets detected as a sliproad and can result in an invalid turn being emitted. We should try and detect turning circles not to be sliproads.

A possible example of such a circle can be found [here](http://www.openstreetmap.org/#map=19/52.50740/13.44824)

PR #2842, opened 2016-09-02
don't emit invalid/detect loops as sliproads

MoKob:
# Issue

resolves https://github.com/Project-OSRM/osrm-backend/issues/2839

## Tasklist
 - [x] review
 - [ ] adjust for for comments

## Requirements / Relations

### DIFF
--- a/features/guidance/dedicated-turn-roads.feature
+++ b/features/guidance/dedicated-turn-roads.feature
@@ -176,7 +176,7 @@ Feature: Slipways and Dedicated Turn Lanes
             |   |   |   |   |   |   |   |   |   |
             |   |   |   |   |   |   | e | g |   |
             |   |   |   |   |   |   |   |   |   |
-            |   |   |   |   |   |   |   |   |   |
+            |   |   |   |   |   |   | 1 |   |   |
             |   |   |   |   |   |   |   | h |   |
             |   |   |   |   |   |   |   |   |   |
             |   |   |   |   |   |   |   | f |   |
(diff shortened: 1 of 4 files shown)


features/guidance/dedicated-turn-roads.feature
```diff
@@ -194,7 +194,73 @@ Feature: Slipways and Dedicated Turn Lanes
...
+            | cdefghijklmnopc | circled | yes    | residential | 1     |
+
+     When I route I should get
+            | waypoints | bearings     | route           | turns         |
+            | b,a       | 90,10 270,10 | circled,circled | depart,arrive |
+
+    @todo
+    #due to the current turn function, the left turn bcp is exactly the same cost as pcb, a right turn. The right turn should be way faster,though
+    #for that reason we cannot distinguish between driving clockwise through the circle or counter-clockwise. Until this is improved, this case here
+    #has to remain as todo (see #
```
daniel-j-h: I think the issue ref is missing here.

src/extractor/guidance/sliproad_handler.cpp
```diff
@@ -78,10 +78,14 @@ operator()(const NodeID, const EdgeID source_edge_id, Intersection intersection)
             // a one-sided sliproad, however, the non-sliproad can be considered `obvious`. Here we
             // assume that this could be the case and check for a potential sliproad/non-sliproad
             // situation.
-            const auto intersection_following_index_one =
-                findNextIntersectionForRoad(intersection_node_id, intersection[1], NULL);
-            const auto intersection_following_index_two =
-                findNextIntersectionForRoad(intersection_node_id, intersection[2], NULL);
+            NodeID intersection_node_one, intersection_node_two;
```
daniel-j-h: It seems like `findNextIntersectionForRoad` takes a `NodeID *` writing into it. I think with this change a `NodeID & out` would be fine, too. And then you can change the logic in the lambda.

src/extractor/guidance/sliproad_handler.cpp
```diff
@@ -151,6 +157,10 @@ operator()(const NodeID, const EdgeID source_edge_id, Intersection intersection)
     const auto next_road_next_intersection =
         findNextIntersectionForRoad(intersection_node_id, next_road, &next_intersection_node);
 
+    // If we are at a traffic loop at the end of a road, don't consider it a sliproad
+    if( intersection_node_id == next_intersection_node )
```
daniel-j-h: clang-format

Issue #2846, opened 2016-09-02
Cannot detect loss of prefix


MoKob:
When detecting prefix/suffix changes, we fail to detect changes of the form `North Street` to `Street`. The issue here is that ` ` is not removed to compare the remaining strings.

PR #2847, opened 2016-09-02
Fix/prefix discovery

MoKob:
# Issue

https://github.com/Project-OSRM/osrm-backend/issues/2846

## Tasklist
 - [x] review
 - [ ] adjust for for comments

## Requirements / Relations
None

### DIFF
--- a/features/guidance/suffix-changes.feature
+++ b/features/guidance/suffix-changes.feature
@@ -3,7 +3,7 @@ Feature: Suppress New Names on dedicated Suffices
 
     Background:
         Given the profile "car"
-        Given a grid size of 10 meters
+        Given a grid size of 2000 meters
 
     Scenario: Suffix To Suffix
         Given the node map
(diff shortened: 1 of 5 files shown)


features/guidance/suffix-changes.feature
```diff
@@ -3,7 +3,7 @@ Feature: Suppress New Names on dedicated Suffices
 
     Background:
         Given the profile "car"
-        Given a grid size of 10 meters
+        Given a grid size of 2000 meters
```
daniel-j-h: Why do we need this change here? Because the suffix change only triggers at a specific distance?

include/extractor/guidance/toolkit.hpp
```diff
@@ -205,43 +205,50 @@ inline bool requiresNameAnnounced(const std::string &from,
...
-            }();
-
-            return is_prefix_change || is_suffix_change;
+    const auto checkForPrefixOrSuffixChange = [](
+        const std::string &first, const std::string &second, const SuffixTable &suffix_table) {
+
+        const auto first_prefix_and_suffixes = getPrefixAndSuffix(first);
+        const auto second_prefix_and_suffixes = getPrefixAndSuffix(second);
+        // reverse strings, get suffices and reverse them to get prefixes
+        const auto checkTable = [&](const std::string str) {
```
daniel-j-h: Should this be a `const &`?

features/guidance/suffix-changes.feature
MoKob: exactly. short segments might simply suppress name changes

PR #2849, opened 2016-09-03
Update turn penalty function to better fit some measured data.

danpat:
This PR modifies the existing `car.lua` turn function to attempt to better model car turning penalties.

I went out driving and measured a bunch of turning movements:

screen shot 2016-09-02 at 5 21 03 pm

The key observations from the sample data were:

  1. We weren't penalizing 90 degree turns enough - typical values were from the 2-4 second range.
  2. There is not as much bias from left-to-right as we applied.
  3. Turns <45 degrees incurred relatively little penalty.
  4. Traffic-control mechisms (lights, stop signs) play a big part in the cost of turns.

Our existing turn function only accepts an angle value.  This PR updates the turn function (and default parameters) for the `car.lua` profile to attempt to get a slighly better fit.  In general, turns >90 degrees should have significantly higher penalties (2x or more), and shallower angle turns should have smaller penalties.  There is a left/right bias still, shallow-angle left turns (in right-driving countries) do incur an extra cost.

As can be seen on the plot above, turns that traverse intersections with traffic control (lights and stop signs) incur a higher penalty than those that do not.  At this stage we don't have that data available to the turn function, so this change is incomplete.

This PR is intended as a stepping stone towards https://github.com/Project-OSRM/osrm-backend/pull/2822 that *will* pass additional data to the turn function.

## Tasklist
 - [ ] review
 - [ ] merge

### DIFF
--- a/profiles/car.lua
+++ b/profiles/car.lua
@@ -150,10 +150,10 @@ properties.left_hand_driving               = false
 
 local side_road_speed_multiplier = 0.8
 
-local turn_penalty               = 1
+local turn_penalty               = 7.5
 -- Note: this biases right-side driving.  Should be
 -- inverted for left-driving countries.
-local turn_bias                  = properties.left_hand_driving and 1/1.2 or 1.2
+local turn_bias                  = properties.left_hand_driving and 1/1.075 or 1.075
 
 local obey_oneway                = true
 local ignore_areas               = true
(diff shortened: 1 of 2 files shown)


daniel-j-h:
Do we want to use the turn angles not only for penalties but also e.g. to prevent uturns at a road with high speed?

Can and should we use additional heuristics to influence the turn penalty, such as the road's width, lanes and other factors that may make a difference especially when the road is not modelled out in detail in the base map?

Can we add some Cucumber tests for this making sure the values make sense (+-) for certain turn angles? Mostly to catch regressions and to make sure this important part of the pipeline stays functional.

danpat:
@daniel-j-h This PR is just a quick-fix - https://github.com/Project-OSRM/osrm-backend/pull/2822 is the place to include a lot of the extra parameters you mentioned, and it requires more than just a profile change.

Doing a CucumberJS test for just this function would add the requirement of having a Lua interpreter installed - something we don't currently require (just the runtime libs).  Given the limited scope of this change, I'd like to hold off investing in testing infrastructure until #2822 lands.

profiles/car.lua
```diff
@@ -544,12 +544,13 @@ function way_function (way, result)
 end
 
 function turn_function (angle)
-  ---- compute turn penalty as angle^2, with a left/right bias
+  -- Use a sigmoid function to return a penalty that maxes out at turn_penalty
+  -- over the space of 0-180 degrees.  Values here were chosen by fitting
```
willwhite: Does this mean this function expecting input from `0` - `180`? How does that reconcile with [this comment](https://github.com/Project-OSRM/osrm-backend/blob/b19a5cc0201f0c57532756d6b0852cf6f20e05dc/include/util/guidance/toolkit.hpp#L32-L35) about how turn angles range from `0` - `360`?

Also, [this comment](https://github.com/Project-OSRM/osrm-backend/issues/2918#issuecomment-247380039) implies that the turn function is passed negative values (e.g. `turn_function(-135)`).

In short, what is the range of possible values that can be passed to `turn_function()`?

MoKob: We have two different notations here. OSRM internally operates on turn angles from `0` to `360` with `180` going straight.

Now in the profile, we use `-180` as `u-turn` to the left, `0` as going straight, `180` as u-turn to the right.

Issue #2850, opened 2016-09-04
profiles/lib/guidance.lua throwing error trying to call `trimLaneString`


danpat:
While trying to replicate https://github.com/Project-OSRM/osrm-backend/issues/2832, I'm hitting this error using the current `master` (bb244e28bbec1513339962b4ac07d155e6eed122):

```
$ ./osrm-extract -p ../profiles/car.lua ../../unprocessed-osm.pbf
[info] input file generated by Osmium (http://wiki.openstreetmap.org/wiki/Osmium)
[info] timestamp: 2015-11-11T22:22:02Z
[info] Found 3 exceptions to turn restrictions:
[info]   motorcar
[info]   motor_vehicle
[info]   vehicle
[New Thread 0x7fffbf7fe700 (LWP 73528)]
[New Thread 0x7fffbeffd700 (LWP 73529)]
[New Thread 0x7fffbe7fc700 (LWP 73530)]
[New Thread 0x7fffc47f8700 (LWP 73531)]
[New Thread 0x7fffbdffb700 (LWP 73532)]
[New Thread 0x7fffbdbfa700 (LWP 73533)]
[New Thread 0x7fffbd3f8700 (LWP 73534)]
[New Thread 0x7fffbd7f9700 (LWP 73535)]
[New Thread 0x7fffbcbf5700 (LWP 73536)]
[New Thread 0x7fffbc7f4700 (LWP 73537)]
[New Thread 0x7fff97bfe700 (LWP 73539)]
[New Thread 0x7fff973fc700 (LWP 73541)]
[New Thread 0x7fff977fd700 (LWP 73540)]
[New Thread 0x7fff97fff700 (LWP 73538)]
[New Thread 0x7fff96bfa700 (LWP 73543)]
[New Thread 0x7fff967f9700 (LWP 73544)]
[New Thread 0x7fff96ffb700 (LWP 73542)]
[New Thread 0x7fff963f8700 (LWP 73545)]
[New Thread 0x7fff95ff7700 (LWP 73546)]
terminate called after throwing an instance of 'osrm::util::exception'
  what():  ERROR occurred in profile script:
No matching overload found, candidates:
std::string trimLaneString(std::string,int,int)
```

It looks like the problem might lie here:

https://github.com/Project-OSRM/osrm-backend/blob/master/profiles/lib/guidance.lua#L99

[...]

Issue #2702, opened 2016-07-26
Fix lanes cucumber tests


TheMarex:
Currently all lanes tests only cover the lanes for the first intersection in a step.

We need to adapt this to all intersections covered by a step.

/cc @daniel-j-h @MoKob

MoKob:
I've recently added lanes to all intersections. This is resolved.

PR #2853, opened 2016-09-05
Fix/another typo

MoKob:
# Issue

Followup on https://github.com/Project-OSRM/osrm-backend/issues/2850

## Tasklist
 - [x] review
 - [ ] adjust for for comments

### DIFF
--- a/features/guidance/turn-lanes.feature
+++ b/features/guidance/turn-lanes.feature
@@ -120,6 +120,26 @@ Feature: Turn Lane Guidance
             | a,d       | road,turn,turn | depart,turn right,arrive | ,straight:false right:true, |
             | a,c       | road,road      | depart,arrive            | ,                           |
 
+    Scenario: Turn with Bus-Lane Left
+        Given the node map
+            |   |   | d |   |   |   |   |
+            |   |   |   |   |   |   |   |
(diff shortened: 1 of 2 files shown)


features/guidance/turn-lanes.feature
```diff
@@ -120,6 +120,26 @@ Feature: Turn Lane Guidance
...
+            | ab    | road | left\|through\|     | 1                 | yes    |
+            | bc    | road |                     |                   | yes    |
+            | bd    | turn |                     |                   | yes    |
+            | cf    | turn |                     |                   | yes    |
+            | ce    | side |                     |                   | yes    |
+
+        When I route I should get
+            | waypoints | route          | turns                   | lanes                       |
+            | a,d       | road,turn,turn | depart,turn left,arrive | ,left:true straight:false,  |
+            | a,c       | road,road      | depart,arrive           | ,                           |
```
daniel-j-h: There are no lanes here since going straight is obvious and we silence lanes for this? It should be still on the intersection array and at some point we have to test for this, correct?

MoKob: yes, but unrelated change, see https://github.com/Project-OSRM/osrm-backend/issues/2702

Issue #2852, opened 2016-09-05
osrm-extract stuck in a loop


danpat:
Using `master` at bb244e28bbec1513339962b4ac07d155e6eed122, and the `unprocessed-osm.pbf` linked from here, https://github.com/Project-OSRM/osrm-backend/issues/2832, `osrm-extract` seems to take an extrodinarily long amount of time (I haven't yet seen it finish).

After a period (a couple of hours) of normal busy-ness on all CPUs, activity seems to reduce to just one CPU.  I attached a debugger and took a look at the only busy thread, here's the backtrace:

```
#0  0x00007ffff5aca7bd in __kernel_rem_pio2 (x=x@entry=0x7fffffff8df0, y=y@entry=0x7fffffff8e10, e0=e0@entry=-24, nx=nx@entry=3, prec=prec@entry=2,
    ipio2=ipio2@entry=0x7ffff5b5c620 <two_over_pi>) at ../sysdeps/ieee754/dbl-64/k_rem_pio2.c:327
#1  0x00007ffff5aeedc9 in __ieee754_rem_pio2l (x=0.91289970124339009762207131082334399, x@entry=0.91289970124339009762207131082334399, y=y@entry=0x7fffffff8e70)
    at ../sysdeps/ieee754/ldbl-96/e_rem_pio2l.c:220
#2  0x00007ffff5af0ead in __sinl (x=x@entry=0.91289970124339009762207131082334399) at ../sysdeps/ieee754/ldbl-96/s_sinl.c:78
#3  0x000000000052febb in sin (__x=<optimized out>) at /usr/include/c++/4.8/cmath:449
#4  latToY (latitude=...) at /mnt/data/x/osrm-backend/include/util/web_mercator.hpp:44
#5  osrm::util::coordinate_calculation::computeAngle (first=..., first@entry=..., second=..., third=...)
    at /mnt/data/x/osrm-backend/src/util/coordinate_calculation.cpp:172
#6  0x00000000005a76d3 in osrm::extractor::guidance::IntersectionGenerator::GetConnectedRoads (this=this@entry=0x7fffffff9688, from_node=from_node@entry=4953888,
    via_eid=via_eid@entry=11009625) at /mnt/data/x/osrm-backend/src/extractor/guidance/intersection_generator.cpp:129
#7  0x00000000005a7f45 in osrm::extractor::guidance::IntersectionGenerator::GetActualNextIntersection (this=this@entry=0x7fffffff9688,
[...]

PR #2854, opened 2016-09-05
Fix/2852

MoKob:
# Issue

https://github.com/Project-OSRM/osrm-backend/issues/2852

We should probably verify on the huge extract as well, I was able to reproduce a loop in a cucumber case though and added some code to catch it. The loops should be caught now, but it would be great to verify this on the larger extract. @danpat do you have a setup running here?

## Tasklist
 - [x] review
 - [x] adjust for for comments

### DIFF
--- a/features/guidance/bugs.feature
+++ b/features/guidance/bugs.feature
@@ -1,0 +1,35 @@
+@routing @guidance
+Feature: Features related to bugs
+
+    Background:
+        Given the profile "car"
+        Given a grid size of 5 meters
+
+    @2852
+    Scenario: Loop
+        Given the node map
+            | a | 1 |   | g |   |   | b |
+            |   |   |   |   |   |   |   |
+            |   |   |   |   |   |   |   |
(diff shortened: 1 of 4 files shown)


features/guidance/bugs.feature
```diff
@@ -0,0 +1,35 @@
...
+            | bfc   | right  | yes    |
+            | chd   | bottom | yes    |
+            | dea   | left   | yes    |
+
+        And the nodes
+            | node | highway         |
+            | g    | traffic_signals |
+            | f    | traffic_signals |
+            | h    | traffic_signals |
+            | e    | traffic_signals |
```
daniel-j-h: Is this only happening with traffic signals? It's not clear to me from the two-line fix below.

MoKob: Well. It should work just the same without. The test here covers more, though, since we check both identical names not collapsed and simple name changes in between. The general issue here is that we need a disconnected loop of some form which will never reach anything else but a two way intersection.

PR #2855, opened 2016-09-05
Be more strict classifying staggered intersections

### LINKED ISSUE #2824: Broken Testcase / Staggered Intersections might need revisiting

daniel-j-h:
# Issue

Staggered Intersections were triggering on data issues / uturn'ish and straight'ish "turns". No longer rely on direction modifier but instead use turn angles for being more strict in classification.

## Tasklist
 - [ ] review
 - [ ] adjust for for comments

## Requirements / Relations
- https://github.com/Project-OSRM/osrm-backend/issues/2824
- https://github.com/Project-OSRM/osrm-backend/pull/2825

### DIFF
--- a/src/engine/guidance/post_processing.cpp
+++ b/src/engine/guidance/post_processing.cpp
@@ -1,5 +1,5 @@
-#include "extractor/guidance/turn_instruction.hpp"
 #include "engine/guidance/post_processing.hpp"
+#include "extractor/guidance/turn_instruction.hpp"
 
 #include "engine/guidance/assemble_steps.hpp"
 #include "engine/guidance/lane_processing.hpp"
@@ -503,8 +503,8 @@ void collapseTurnAt(std::vector<RouteStep> &steps,
                 steps[one_back_index] =
                     elongate(std::move(steps[one_back_index]), steps[next_step_index]);
                 invalidateStep(steps[next_step_index]); // will be skipped due to the
-                                                       // continue statement at the
-                                                       // beginning of this function
+                                                        // continue statement at the
+                                                        // beginning of this function
             }
 
             steps[one_back_index].name = steps[two_back_index].name;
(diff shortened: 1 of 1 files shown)


features/testbot/via.feature
```diff
@@ -264,9 +264,6 @@ Feature: Via points
             | 9,8,7     | cd,da,ab,bc,cd,cd,cd,da,ab,bc,cd,cd | 3000m +-1    |
 
     # See issue #2706
-    # this case is currently broken. It simply works as put here due to staggered intersections triggering a name collapse.
-    # See 2824 for further information
-    @todo
     Scenario: Incorrect ordering of nodes can produce multiple U-turns
```
MoKob: right now this case is somewhat misleading. It only prints the way it does due to a hidden name change. I would say that the case needs to be adjusted to, in the end, print the correct path. This would require the instructions to have slight angles in them, though.
As the test is right now, it provides more confusion than anything else.

daniel-j-h: Hm as this is unrelated to staggered intersections should I just leave the todo in then?

MoKob: I guess keeping it as todo is the right thing to do for now

src/engine/guidance/post_processing.cpp
```diff
@@ -528,12 +528,22 @@ bool isStaggeredIntersection(const RouteStep &previous, const RouteStep &current
...
+    // This more closely resembles what we understand as Staggered Intersection.
+    const auto is_right = [](const double angle) { return angle > 45 && angle < 135; };
+    const auto is_left = [](const double angle) { return angle > 225 && angle < 315; };
 
-    const auto left_right = isLeftTurn(previous.maneuver.instruction) && //
-                            isRightTurn(current.maneuver.instruction);
-    const auto right_left = isRightTurn(previous.maneuver.instruction) && //
-                            isLeftTurn(current.maneuver.instruction);
+    const auto left_right = is_left(angle(previous)) && is_right(angle(current));
+    const auto right_left = is_right(angle(previous)) && is_left(angle(current));
```
MoKob: should we also check for similar deviations or are combinations of sharp/slight still valid?
Have you looked at the combinations of extremes to see if they follow your idea of a staggered intersection?

daniel-j-h: The main use-case for this was the Staggered Intersection phenomena in the US where grid-like cities are the norm.

> Staggered intersections are very common in rural and exurban areas there. Often, as an area becomes built up, traffic engineers bend the roads to link them up directly, leaving a rump street.

daniel-j-h: See the following for an analysis in the US:
http://bl.ocks.org/anonymous/raw/5b44f131d86f3710d3b0e20b210c1798/

MoKob: We still should do an analysis of the non-obvious cases. If you, as was the case prior to this PR, allow 180 and 0 degree within the turns, strange situations might get detected.
What I am concerned about are cases like these:
sharp-staggered

If at some point we should increase the threshold (3 meters might be too small), we could end up combining very non-obvious turns.

I can't see from the above analysis if such cases would be a valid choice to classify them as staggered. We should at least see if these cases can exist, where they are and if they don't even exist make the angle more restricted. I know I said in chat `like 45 to 135 degreeish`, but I feel that we should test these bounds.

daniel-j-h: Yes the turn angles could also be steeper if we only target Staggered Intersections in the US.

PR #2845, opened 2016-09-02
[up for the taking] cases for new names on motorways, requires reference changes

MoKob:
# Issue

https://github.com/Project-OSRM/osrm-backend/issues/2744
## Tasklist
 - [ ] Adjust [here](https://github.com/Project-OSRM/osrm-backend/blob/3b75308b81e8d69aa70114837d98d02040bb95b2/include/extractor/guidance/toolkit.hpp#L179-L198) when https://github.com/Project-OSRM/osrm-backend/issues/2704 has landed.
 - [x] add regression / cucumber cases (see docs/testing.md)
 - [ ] review
 - [ ] adjust for for comments

## Requirements / Relations
https://github.com/Project-OSRM/osrm-backend/issues/2704

Issue #2744, opened 2016-08-05
Suppress temporary name changes along a designated route


1ec5:
Going from Dallas to Missouri along U.S. 71/69, I encountered many name changes as I passed through small towns. The name would change to things like "Mays St." or "Main St." within the town limits before reverting to "State Hwy. 69" (which would be tagged in OSM as simply `ref=US 69`). Because the ref remains constant throughout the stretch of highway, more sophisticated routers avoid any mention of the name changes; instead, the UI and voice treat the entire road as one road named "US 69". This gives the user of a turn-by-turn navigation application a more accurate idea of the total length of highway they need to traverse, rather than the length of each individual town and the distances between them.

OSRM should similarly suppress any name change that reverts after (say) a mile, where the ref stays constant, and the road is a straightaway.

/cc @danpat @MoKob

daniel-j-h:
Route with instructions on the demo server:

http://map.project-osrm.org/?z=7&center=36.337253%2C-95.581055&loc=32.776272%2C-96.796856&loc=38.760481%2C-92.561787&hl=en&alt=0

```bash
http --body 'https://router.project-osrm.org/route/v1/driving/-96.796856,32.776272;-92.561787,38.760481?steps=true'
```
39 `NewName` instructions, 25 of those have "US 69" in their name (ref).

I think we need https://github.com/Project-OSRM/osrm-backend/issues/2704 first to have a separate name and ref in the `RouteStep`.
Then we could do this simply in post-processing.

1ec5:
This would also mitigate name changes due to bridge names in OSM.

MoKob:
The problem here is that the situation is not uniformly valid. In Germany, at least, there are a lot of roads that keep the same reference but have different names. The name might be the only displayed information, though, and the reference is only used on major directions signs. Suppressing the name change completely is difficult.

To me, this is just another case of not announcing name changes in voice but simply displaying them.

1ec5:
In OSM, at least in the U.S., it's customary to tag non-signposted references with a tag like `unsigned_ref` or `ref:unsigned`. Unfortunately, there isn't an established way to indicate that a reference is only sporadically or inconspicuously signposted, for example only at major junctions or only when you squint at the sign. That occurs often in the U.S. as well, for example on county roads.

[...]

Issue #2704, opened 2016-07-26
Split a way's "Name (Ref)" into name, ref fields


daniel-j-h:
Currently we re-write a way's name to

    "Name (ref)"

where ref could be the destination if there is no ref. This string-representation is hachky at best and makes it harder for clients to build great guidance.

Task: have independent `name`, `ref`, `destination` fields and expose them to clients.

cc @freenerd @1ec5 @bsudekum

Impl. note: will break hard coded hacks like:
- https://github.com/Project-OSRM/osrm-backend/blob/c7a1576100cb947af0e970644fb3f70dde63a9eb/include/extractor/guidance/toolkit.hpp#L166-L168

1ec5:
Should ref and destination be simple strings, or should they be something more complex like arrays of strings? It's already possible for a way to have multiple refs and multiple destinations delimited by semicolons in OSM data. It wouldn't be the end of the world for a client to split strings by this delimiter, but an array would be more semantically correct.

Looking ahead to #482, if OSRM ever gains built-in support for route relations, each ref would need to be a dictionary with separate network and (numeric) ref fields. But that's clearly a bit further off.

daniel-j-h:
We're passing name, ref, destination through the pipeline and deduplicating them based on their string representation anyway.

What we could do in post-processing is the split+trim dance on the strings already for you, returning arrays if that's more convenient.

TheMarex:
I would be more comfortable with passing through `name`, `ref` and `destination` directly from the base map. Running string processing that makes assumptions about structure in OSRM is something we want to get away from. Changing this to arrays would be a breaking API change that I'm not willing to do right now.

1ec5:
How would the use of arrays be a breaking change? Separate ref and destination fields haven't been introduced yet.

In any case, there's still a possibility of the frontend to split the strings into an array so the client doesn't have to. (That would be especially important if route relation support were on the roadmap, in which case a single string wouldn't be flexible enough.)

daniel-j-h:
We do have a `destinations` field since 5.2 (=> [Changelog](https://github.com/Project-OSRM/osrm-backend/blob/master/CHANGELOG.md#520)).

TheMarex:
We already expose `destinations` as string:

https://github.com/Project-OSRM/osrm-backend/blob/fb4834458eb0157c06075c85dca3e791a05d1d6e/docs/http.md#routestep

[...]

PR #2857, opened 2016-09-05
Return name and ref response fields separately

karenzshea:
# Issue

Ref https://github.com/Project-OSRM/osrm-backend/issues/2704
Splits name and ref into separate fields in the response object.

## Tasklist
 - [x] review
 - [x] adjust for for comments

## Requirements / Relations
Enables https://github.com/Project-OSRM/osrm-backend/issues/2744 and https://github.com/Project-OSRM/osrm-backend/pull/2845

cc @daniel-j-h @MoKob @emiltin


daniel-j-h:
@emiltin this will finally provide separate `name`, `ref`, `destination` fields in the response when it lands, no longer having to parse the weird `name (ref)` construct we're currently having.

Which also means your parsing code will probably have to change. Just keeping you in the loop here.

daniel-j-h:
Tasks
- [x] add a ref property to the `ExtractionWay` type
- [x] register this new property for Lua/Luabind in the `LuaScriptingEnvironment`
- [ ] rewrite profiles to separately write `result.name` `result.ref` `result.destinations`; no more hacks
- [ ] in extractor_callbacks.hpp adapt the string deduplication map
- [ ] in extractor_callbacks..cpp deduplicate and add ref to external_memory datastructure
- [ ] add a ref attribute to the `RouteStep` type representing a step in the response
- [ ] add function for returning the ref based on name id to datafacades (base, internal, shared, mock)
- [ ] in assemble_steps.hpp call datafacade function and add the result to the `RouteStep` object
- [ ] in json_factory.cpp serialize the `RouteStep`'s new attribute to the json::object
- [ ] adapt cucumber support code: route.js, shared_steps.js, 
- [ ] adapt cucumber tests (names.feature, refs.feature, ..) testing the new behavior
- [ ] add api docs for the new ref field in `RouteStep` to `docs/http.md`
- [ ] add a changelog entry for both profiles and api change

features/bicycle/names.feature
```diff
@@ -16,7 +16,7 @@ Feature: Bike - Street names in instructions
 
         When I route I should get
             | from | to | route                                   |
-            | a    | c  | My Way (A6),Your Way (A7),Your Way (A7) |
+            | a    | c  | My Way,Your Way,Your Way                |
```
MoKob: these have to be added as an additional check instead of being removed `| route | ref |` ... or something like this. References are an essential part of guidance checks.

emiltin:
thanks @daniel-j-h for the update

CHANGELOG.md
```diff
@@ -3,11 +3,13 @@
     - Profiles
       - includes library guidance.lua that offers preliminary configuration on guidance.
       - added left_hand_driving flag in global profile properties
+      - return `ref` and `name` as separate fields. Do no use ref or destination as fallback for name value
     - Guidance
       - Handle Access tags for lanes, only considering valid lanes in lane-guidance (think car | car | bike | car)
     - API:
       - `annotations=true` now returns the data source id for each segment as `datasources`
       - Reduced semantic of merge to refer only to merges from a lane onto a motorway-like road
+      - new `ref` field in the `RouteStep` object
```
daniel-j-h: What is is good for? Also maybe a note about "name (ref)" -> `name`, `ref`.

features/bicycle/names.feature
```diff
@@ -15,8 +15,8 @@ Feature: Bike - Street names in instructions
             | bc    | Your Way | A7  |
 
         When I route I should get
-            | from | to | route                                   |
-            | a    | c  | My Way (A6),Your Way (A7),Your Way (A7) |
+            | from | to | route                                   | ref |
+            | a    | c  | My Way,Your Way,Your Way                | A6,A7,A7 |
```
daniel-j-h: Align tables.

features/bicycle/ref.feature
```diff
@@ -25,8 +25,8 @@ Feature: Bike - Way ref
             | ab    |      | E7  |
 
         When I route I should get
-            | from | to | route |
-            | a    | b  | E7,E7 |
+            | from | to | route                               | ref   |
+            | a    | b  | {highway:primary},{highway:primary} | E7,E7 |
```
daniel-j-h: I'm not sure if we really want this as a fallback. But then again it's the bike profile..

karenzshea: @emiltin do you have thoughts on removing this fallback?

features/guidance/dedicated-turn-roads.feature
```diff
@@ -165,8 +165,8 @@ Feature: Slipways and Dedicated Turn Lanes
             | qe    | secondary_link | Ettlinger Allee    |      | yes    |
 
         When I route I should get
-            | waypoints | route                                                     | turns                    |
-            | a,o       | Schwarzwaldstrasse (L561),Ettlinger Allee,Ettlinger Allee | depart,turn right,arrive |
+            | waypoints | route                                                    | turns                    | ref        |
+            | a,o       | Schwarzwaldstrasse,Ettlinger Allee,Ettlinger Allee       | depart,turn right,arrive | L561,L561, |
```
daniel-j-h: Before it was a single ref, now it's two refs?

features/bicycle/ref.feature
daniel-j-h: Looking at the OSM Wiki for ref:
http://wiki.openstreetmap.org/wiki/Key:ref

it might make sense now to also try more namespaced refs, as in: try the following tags in order:
-  `ref=`
- `ncn_ref=`
- `rcn_ref=`
- `lcn_ref=`

for the bike profile.

A similar approach can be taken for the car profile I guess.

daniel-j-h: @emiltin what's your take on http://wiki.openstreetmap.org/wiki/Cycle_routes Shouldn't the bike profile support this?

src/engine/api/json_factory.cpp
```diff
@@ -233,6 +233,7 @@ util::json::Object makeRouteStep(guidance::RouteStep step, util::json::Value geo
     route_step.values["distance"] = std::round(step.distance * 10) / 10.;
     route_step.values["duration"] = std::round(step.duration * 10) / 10.;
     route_step.values["name"] = std::move(step.name);
+    route_step.values["ref"] = std::move(step.ref);
```
daniel-j-h: Should the `ref` field always be there and empty if there is no ref for the way? Or --- similar to pronunciation and destinations --- should the ref field be undefined and only be available if there is a ref on the way?

(The same applies to the name field: shouldn't it really be optional? Or why do we separate between how we handle non-existent names and non-existent destinations)

src/extractor/extractor_callbacks.cpp
```diff
@@ -242,16 +242,15 @@ void ExtractorCallbacks::ProcessWay(const osmium::Way &input_way, const Extracti
...
     // Otherwise fetches the id based on the name and returns it without insertion.
     const auto turn_lane_id_forward = requestId(parsed_way.turn_lanes_forward);
     const auto turn_lane_id_backward = requestId(parsed_way.turn_lanes_backward);
 
     const constexpr auto MAX_STRING_LENGTH = 255u;
-    // Get the unique identifier for the street name
-    // Get the unique identifier for the street name and destination
-    const auto name_iterator = string_map.find(MapKey(parsed_way.name, parsed_way.destinations));
+    // Get the unique identifier for the street name, destination, and ref
+    const auto name_iterator = string_map.find(MapKey(parsed_way.name, parsed_way.destinations, parsed_way.ref));
```
daniel-j-h: By now all the `GetNameForID`, `name_iterator`, `name_id`s are semantically wrong and probably should be called something like data id or metadata id.

docs/http.md
```diff
@@ -469,6 +469,7 @@ step.
   | geojson    | [GeoJSON `LineString`](http://geojson.org/geojson-spec.html#linestring) or [GeoJSON `Point`](http://geojson.org/geojson-spec.html#point) if it is only one coordinate (not wrapped by a GeoJSON feature)|
   
 - `name`: The name of the way along which travel proceeds.
+- `ref`: A reference number or code for the way.
```
MoKob: note optional/undefined status when not present

profiles/bicycle.lua
```diff
@@ -211,19 +211,20 @@ function way_function (way, result)
...
 
   -- name
-  if ref and "" ~= ref and name and "" ~= name then
-    result.name = name .. " (" .. ref .. ")"
-  elseif ref and "" ~= ref then
-    result.name = ref
-  elseif name and "" ~= name then
+  if name and "" ~= name then
     result.name = name
   -- TODO find a better solution for encoding way type
```
MoKob: what about this todo? Way types are encoded separately now. Anything else missing here?

features/bicycle/ref.feature
systemed: NCN and RCN routes are mostly tagged with route relations now rather than the older `ncn_ref` and `rcn_ref` tags on ways, though `lcn_ref` is still frequently used for city cycle routes IME.

src/engine/api/json_factory.cpp
karenzshea: We should handle `ref` like `destination`; it's optional in real life.

MoKob:
From discussion in chat. We are trying to include this in 5.4.0. I updated the milestone accordingly

src/engine/api/json_factory.cpp
karenzshea: @daniel-j-h The handling of `name` is probably fine though. In real life I don't really think it's optional for streets to not have names anymore, so an empty name field implies that we don't have the data, whereas a missing field implies it doesn't exist.

src/extractor/extractor_callbacks.cpp
karenzshea: Punting on renaming this right now.

features/guidance/dedicated-turn-roads.feature
MoKob: Looks like in post-processing, whenever name gets updated, refs are not getting updated accordingly

karenzshea: This is actually an existing bug where in post-processing we don't properly update collapsed steps with the correct destination and pronunciation data along with name. PR coming for this fix, which can then include moving `ref` in preprocessing as well.

include/extractor/extractor_callbacks.hpp
```diff
@@ -39,7 +54,7 @@ class ExtractorCallbacks
     // used to deduplicate street names and street destinations: actually maps to name ids
     using MapKey = std::tuple<std::string, std::string, std::string>;
     using MapVal = unsigned;
-    std::unordered_map<MapKey, MapVal, boost::hash<MapKey>> string_map;
+    std::unordered_map<MapKey, MapVal, std::hash<MapKey>> string_map;
```
daniel-j-h: `std::hash` is the default, you can remove this. See http://en.cppreference.com/w/cpp/container/unordered_map

features/foot/ref.feature
```diff
@@ -26,7 +26,7 @@ Feature: Foot - Way ref
 
         When I route I should get
             | from | to | route |
-            | a    | b  | E7,E7 |
+            | a    | b  | {highway:primary},{highway:primary} |
```
MoKob: align table

features/guidance/destination-signs.feature
```diff
@@ -29,13 +29,13 @@ Feature: Destination Signs
...
-          | i    | j  | ,                                                         | Berlin,Berlin                                   |                           |
-          | k    | l  | KL (E1),KL (E1)                                           | A1: Berlin,A1: Berlin                           |                           |
-          | m    | n  | MN (A1, A2: Berlin, Hamburg),MN (A1, A2: Berlin, Hamburg) | A1, A2: Berlin, Hamburg,A1, A2: Berlin, Hamburg |                           |
-          | o    | p  | OP,OP                                                     | ,                                               | guard against mis-tagging |
-          | q    | r  | QR (A1, A2),QR (A1, A2)                                   | A1, A2,A1, A2                                   |                           |
+          | from | to | route                                                     | destinations                                    | ref | #                         |
+          | a    | b  | AB,AB                                                     | ,                                               | E1,E1 |                           |
```
MoKob: align

features/guidance/roundabout.feature
```diff
@@ -283,9 +283,9 @@ Feature: Basic Roundabout
             | dmg   | roundabout |          | trunk_link | yes    |        |
 
         When I route I should get
-            | waypoints | route                                                 | turns                           |
-            | a,e       | crescent (US 130),crescent (US 130),crescent (US 130) | depart,roundabout-exit-3,arrive |
-            | j,l       | NJ 38,NJ 38,NJ 38                                     | depart,roundabout-exit-2,arrive |
+            | waypoints | route                                                 | turns                           | ref                     |
+            | a,e       | crescent,crescent,crescent                            | depart,roundabout-exit-3,arrive | US 130,US 130,US 130    |
+            | j,l       | ,,                                                    | depart,roundabout-exit-2,arrive | NJ 38,NJ 38,NJ 38      |
```
MoKob: align

Issue #2856, opened 2016-09-05
Strange Roundabout Contraption


MoKob:
http://www.openstreetmap.org/#map=19/47.23318/-1.56563 shows a roundabout that we currently can't handle correctly in osrm-backend.

This setup can break the creation of instructions.

daniel-j-h:
Throughabout http://wiki.openstreetmap.org/wiki/Throughabout
https://en.wikipedia.org/wiki/Roundabout#Hamburger_roundabout

MoKob:
The issue here was an invalid roundabout-tag on the bottom side of the roundabout.
As a result, our roundabout handling broke.

MoKob:
Defective roundabout tagging is now detected in the roundabout handler.

PR #2858, opened 2016-09-05
detect broken roundabout-taggings

MoKob:
# Issue

https://github.com/Project-OSRM/osrm-backend/issues/2856
The issue here was an invalid roundabout-tag on the bottom side of the roundabout.
As a result, our roundabout handling broke.

## Tasklist
 - [x] review
 - [x] adjust for for comments

### DIFF
--- a/src/extractor/guidance/roundabout_handler.cpp
+++ b/src/extractor/guidance/roundabout_handler.cpp
@@ -270,12 +270,29 @@ RoundaboutType RoundaboutHandler::getRoundaboutType(const NodeID nid) const
     std::set<NodeID> roundabout_nodes; // needs to be sorted
 
     // this value is a hard abort to deal with potential self-loops
+    const auto countRoundaboutFlags = [&](const NodeID at_node) {
+        // FIXME: this would be nicer as boost::count_if, but our integer range does not support
+        // these range based handlers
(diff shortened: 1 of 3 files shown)


src/extractor/guidance/roundabout_handler.cpp
```diff
@@ -270,13 +270,30 @@ RoundaboutType RoundaboutHandler::getRoundaboutType(const NodeID nid) const
     std::set<NodeID> roundabout_nodes; // needs to be sorted
 
     // this value is a hard abort to deal with potential self-loops
+    const auto countRoundaboutFlags = [&](const NodeID at_node) {
+        // this would be nicer as boost::count_if, but our integer range does not support these
+        // range based handlers
```
daniel-j-h: Let's ticket the irange deficiency and where the hacks for this are in the code.
Or at least let's tag them FIXME or something.

MoKob: Could you ticket the range problem? You have a better understanding of what we need to get our ranges working correctly.

daniel-j-h: https://github.com/Project-OSRM/osrm-backend/issues/2859

Issue #2861, opened 2016-09-06
Mark libosrm public interface threadsafe / const


daniel-j-h:
@jcoupey just noted the overly restrictive `libosrm` interface: our publicly facing functions are not marked `const` although internally we're logically const and threadsafe.

I _think_ it should be good enough sprinkling `const`-ness throughout the `OSRM` and `Engine` type.

We need to check this with our example code and more importantly with `node-osrm`.

jcoupey:
@daniel-j-h You beat me at reporting speed. ;-)

Just to elaborate on how this is restricting in my use case: I wish to have an `OSRM` object as a class member (so I can re-use it at several occasions), but I can't use it in a `const` member function. More details if required at https://github.com/VROOM-Project/vroom/issues/34#issuecomment-244816177.

daniel-j-h:
This was resolved in https://github.com/Project-OSRM/osrm-backend/pull/2862 and already merged into the v5 release, since it's only lifting restrictions from the interface. Closing as done.

PR #2862, opened 2016-09-06
Makes the OSRM interface threadsafe

daniel-j-h:
# Issue

For https://github.com/Project-OSRM/osrm-backend/issues/2861. I still need to check the `mutable unique_ptr<T>` semantics. To be honest I though it would be `unique_ptr<mutable T>` since not the `unique_ptr` should be `mutable` but the type it's storing.

Technically speaking we're changing the `libosrm` API. But since we're only lifting restrictions by marking the API threadsafe, we should be fine here. I'd like to be careful here, anyway.

## Tasklist
 - [x] check `mutable unique_ptr<T>` vs `unique_ptr<mutable T>`
 - [x] review
 - [x] adjust for for comments
 - [x] adapt node-osrm - does node-osrm needs to be adapted anyway? should be fine

@jcoupey

### DIFF
--- a/example/example.cpp
+++ b/example/example.cpp
@@ -34,7 +34,7 @@ int main(int argc, const char *argv[])
     config.use_shared_memory = false;
 
     // Routing machine with several services (such as Route, Table, Nearest, Trip, Match)
-    OSRM osrm{config};
+    const OSRM osrm{config};
 
     // The following shows how to use the Route service; configure this service
     RouteParameters params;
--- a/include/engine/engine.hpp
+++ b/include/engine/engine.hpp
@@ -55,7 +55,7 @@ class Engine final
     // Needs to be public
     struct EngineLock;
 
-    explicit Engine(EngineConfig &config);
+    explicit Engine(const EngineConfig &config);
 
     Engine(Engine &&) noexcept;
     Engine &operator=(Engine &&) noexcept;
@@ -63,12 +63,12 @@ class Engine final
     // Impl. in cpp since for unique_ptr of incomplete types
     ~Engine();
 
-    Status Route(const api::RouteParameters &parameters, util::json::Object &result);
-    Status Table(const api::TableParameters &parameters, util::json::Object &result);
-    Status Nearest(const api::NearestParameters &parameters, util::json::Object &result);
-    Status Trip(const api::TripParameters &parameters, util::json::Object &result);
-    Status Match(const api::MatchParameters &parameters, util::json::Object &result);
-    Status Tile(const api::TileParameters &parameters, std::string &result);
+    Status Route(const api::RouteParameters &parameters, util::json::Object &result) const;
+    Status Table(const api::TableParameters &parameters, util::json::Object &result) const;
+    Status Nearest(const api::NearestParameters &parameters, util::json::Object &result) const;
+    Status Trip(const api::TripParameters &parameters, util::json::Object &result) const;
+    Status Match(const api::MatchParameters &parameters, util::json::Object &result) const;
+    Status Tile(const api::TileParameters &parameters, std::string &result) const;
 
   private:
     std::unique_ptr<EngineLock> lock;
--- a/include/osrm/osrm.hpp
+++ b/include/osrm/osrm.hpp
@@ -83,7 +83,7 @@ class OSRM final
      * \return Status indicating success for the query or failure
      * \see Status, RouteParameters and json::Object
      */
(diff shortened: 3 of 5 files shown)


daniel-j-h:
I was able to remove the `mutable` hacks. Should be cleaner now.

@MoKob want to target this for the 5.4 rc already or for 5.5? @jcoupey if it does not land in 5.4 you can cherry-pick this already, since it's not that big of a change.

jcoupey:
Thanks for fixing this, will make things easier in the future.

In the meantime, I'll still have to go with the work-around you proposed in order to support the whole 5.* range.

MoKob:
@daniel-j-h so I am still unclear on how this does count as an API change. Since it should still be compatible to all use-cases (after all, our internals do not change) I feel we might be fine. Can you make sure it doesn't conflict with our guarantees in any way, though?

daniel-j-h:
Our [releasing docs](https://github.com/Project-OSRM/osrm-backend/blob/master/docs/releasing.md) only state:

> Forward-compatible C++ library API

not sure what exactly that means wrt. `const`ness but since we only add `const` to the functions we lift constrains from `libosrm`: before clients weren't able to call the functions on a `const OSRM` object, now they are.

There are probably still subtle ways this could break clients, e.g. if you inherit from the `OSRM` object and provide your own `const` overloads or crazy things like that..

MoKob:
Thats unfortunate. As long as it can break clients, we are not `forward-compatible`. So based on this, I feel we might need to hold off on this for 6.0, right?

daniel-j-h:
Yes, I'm fine with that.

Already triggering a `node-osrm` build to see if we're good there, too: https://github.com/Project-OSRM/node-osrm/pull/244

TheMarex:
Quite interesting that we can mark the library level functions `const` even though the actual implementation in the plugins like [here](https://github.com/Project-OSRM/osrm-backend/blob/master/src/engine/plugins/viaroute.cpp#L30) can't be `const` because we modify the heap. Also we potentially modify the `DataFacade` on every request [here](https://github.com/Project-OSRM/osrm-backend/blob/master/src/engine/engine.cpp#L103). (though of course everything is thread-safe)

Is that because the non-`const`-ness is in a free-function `RunQuery`? I don't understand how the const-checker lets this slip through. Can you explain to me how this can happen @daniel-j-h?

So given that we definitely modify state (e.g. swap out the whole dataset for example) do you still think `const` makes semantical sense? On the other hand I don't think we leak anything of that internal state (e.g. references to anything stored in the DataFacade or heap) so from an outside perspective it might be considered `const`.

From an API version perspective I personally would not consider this a breaking change because for one the OSRM object is marked `final` (which mean nobody can sub-classes) and @daniel-j-h already laid out that the external interface is forward compatible. I don't see how this could lead to breakage without someone abusing the API.

daniel-j-h:
> Is that because the non-const-ness is in a free-function RunQuery? I don't understand how the const-checker lets this slip through. Can you explain to me how this can happen @daniel-j-h?


In my first iteration I sprinkled `mutable` keywords around but then I noticed the following:

```cpp
struct B { void fn() {} };
struct A { B b; void fn() const { b.fn(); }; };

B b;
const A a{b};
a.fn(); // error
```

does not compile, since `A::fn` sees its members as `const` and tries to invoke `fn` on a `const B` which there is none. Now in our setup:

```cpp
struct B;
struct A { B* b; void fn() const; };

struct B { void fn() {} };
void A::fn() const { b->fn(); }

B b;
const A a{&b};
a.fn(); // works
```

now `A::fn` sees a `const B* b` (!) which is a const pointer to a `B`, not a pointer to a const B. And invoking `fn` through a const pointer to (non-const) `B` is fine. Subtle difference.



> So given that we definitely modify state (e.g. swap out the whole dataset for example) do you still think const makes semantical sense? On the other hand I don't think we leak anything of that internal state (e.g. references to anything stored in the DataFacade or heap) so from an outside perspective it might be considered const.

`const` before C++11 meant "logically const" and not "bitwise const", with C++11 `const` now also implies thread-safety. That's the reason you should mark member attribute `mutex`es and related `mutable` since they're already thread-safe by design.

Here's a great video about `const` / `mutable`:
https://channel9.msdn.com/posts/C-and-Beyond-2012-Herb-Sutter-You-dont-know-blank-and-blank

TheMarex:
> "logically const" and not "bitwise const"

Ah I see that makes sense. Thanks for the explanation.

Up to you if you want to merge this in the `5.x` series or wait for `6.x`. :+1:

PR #2864, opened 2016-09-06
Add example request to the route API esection.

danpat:
Small change, adds a route example query to the route section, as some people were missing it in the General section.


daniel-j-h:
:+1: 

We should have examples for all of them, even if we only e.g. `match` two or three locations.

PR #2866, opened 2016-09-06
calculate roundabout radius from circumference, fix #2716

### LINKED ISSUE #2716: More Robust Roundabout Radius Calculation

ansis:
# Issue

fix #2716

- radius of roundabout is calculated from the total length of the edges in the roundabout
- ~~circleRadius and circleCenter are removed~~
- tests adjusted because collinear roundabouts now have the potential to become `RoundaboutIntersection`s because this [early return](https://github.com/Project-OSRM/osrm-backend/blob/7899444135858ef3e78e2309dacc5859eee49e16/src/extractor/guidance/roundabout_handler.cpp#L344) is removed

## Tasklist
 - [x] review @MoKob 
 - [x] adjust for for comments

### DIFF
--- a/features/guidance/roundabout-turn.feature
+++ b/features/guidance/roundabout-turn.feature
@@ -243,9 +243,29 @@ Feature: Basic Roundabout
             | df    |            |
 
         When I route I should get
-            | waypoints | route    | turns                           |
-            | a,e       | ab,ce,ce | depart,roundabout-exit-1,arrive |
-            | a,f       | ab,df,df | depart,roundabout-exit-2,arrive |
+            | waypoints | route    | turns                                         |
+            | a,e       | ab,ce,ce | depart,roundabout turn right exit-1,arrive    |
+            | a,f       | ab,df,df | depart,roundabout turn straight exit-2,arrive |
+
+       Scenario: Collinear in Y
+        Given the node map
+            |   | a |
+            |   | b |
+            | e | c |
+            |   | d |
+            |   | f |
+
+        And the ways
+            | nodes | junction   |
+            | ab    |            |
+            | bcdb  | roundabout |
+            | ce    |            |
+            | df    |            |
+
+        When I route I should get
+            | waypoints | route    | turns                                         |
+            | a,e       | ab,ce,ce | depart,roundabout turn right exit-1,arrive    |
+            | a,f       | ab,df,df | depart,roundabout turn straight exit-2,arrive |
 
        Scenario: Collinear in X,Y
         Given the node map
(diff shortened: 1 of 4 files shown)


include/util/coordinate_calculation.hpp
```diff
@@ -85,16 +85,6 @@ double bearing(const Coordinate first_coordinate, const Coordinate second_coordi
 // Get angle of line segment (A,C)->(C,B)
 double computeAngle(const Coordinate first, const Coordinate second, const Coordinate third);
 
-// find the center of a circle through three coordinates
```
MoKob: I am not sure whether we should throw away that code. Doesn't really hurt to still have it around (e.g. for turn angles/radius computations).

I would strongly prefer to keep it for cases like these.

ansis: Added it back.

What's the advantage of having unused code over just getting it from git later if it's needed?

MoKob: Well. It is something I currently work on and I feel code gets lost in the depths of git if once removed. After all you have no further documentation on it, unless you keep an issue or something around that documents that this code was there and might be needed again.

Since it is most likely needed in my other PR or at least could be used to compute the turn radius for turn penalty functions, I'd rather not forget that we have it.

ansis: Makes sense. Thanks!

daniel-j-h:
Rebased against master after a couple of weeks of us having other fires to fight. Still argeted for 5.5.

@MoKob what do you think needs to be done here, adding some more tests? How do we handle https://github.com/Project-OSRM/osrm-backend/pull/2741?

MoKob:
@daniel-j-h I assigned this to me to have a closer look. I am not clear on yet on what changes, what the border cases would be to consider here. Will take a look at it tomorrow. 

#2741 could still be useful. We need to figure out curvature of turns for better turn-angles anyhow. We could possibly use it there, if it can deal with non-full circles correctly? Otherwise I would prefer having the simpler solution, as we are guessing either way.

PR #2873, opened 2016-09-07
Adds limits for Nearest service, resolves #2872.

### LINKED ISSUE #2872: Provide default limits for Nearest plugin
There should be limits for the Nearest plugin. In particular limiting
- the amount of locations you can hand it and more importantly
- the amount of results you want on a per location basis and
- the distance at which we still return a "nearest" result

@MoKob @freenerd

daniel-j-h:
# Issue

#2872 

System-wide limiting 1/ search radius and 2/ number of results for Nearest plugin.
Both can trigger whole-graph traversals.

## Tasklist
 - [x] bearing 0,0
 - [x] unit tests (limits.cpp)
 - [x] check all our services for similar issues
 - [ ] discuss default limits
 - [ ] review
 - [ ] adjust for for comments

### DIFF
--- a/docs/http.md
+++ b/docs/http.md
@@ -69,7 +69,7 @@ Where the elements follow the following format:
 
 | Element    | Values                                                 |
 |------------|--------------------------------------------------------|
-|bearing     |`{value},{range}` `integer 0 .. 360,integer 0 .. 180`  |
+|bearing     |`{value},{range}` `integer 0 .. 360,integer 1 .. 180`  |
 |radius      |`double >= 0` or `unlimited` (default)                  |
 |hint        |Base64 `string`                                         |
 
--- a/docs/http.md
+++ b/docs/http.md
@@ -69,7 +69,7 @@ Where the elements follow the following format:
 
 | Element    | Values                                                 |
 |------------|--------------------------------------------------------|
-|bearing     |`{value},{range}` `integer 0 .. 360,integer 1 .. 180`  |
+|bearing     |`{value},{range}` `integer 0 .. 360,integer 0 .. 180`  |
 |radius      |`double >= 0` or `unlimited` (default)                  |
 |hint        |Base64 `string`                                         |
 
--- a/include/engine/engine_config.hpp
+++ b/include/engine/engine_config.hpp
@@ -53,6 +53,9 @@ namespace engine
  *  - Match
  *  - Nearest
  *
+ * You can restrict the coordinate search radius when a bearing is supplied.  This helps
+ * prevent the rtree search expanding if a very narrow bearing filter is supplied.
+ *
  * In addition, shared memory can be used for datasets loaded with osrm-datastore.
  *
  * \see OSRM, StorageConfig
@@ -61,13 +64,16 @@ struct EngineConfig final
 {
     bool IsValid() const;
 
+    constexpr static const int UNLIMITED = -1;
+
     storage::StorageConfig storage_config;
-    int max_locations_trip = -1;
-    int max_locations_viaroute = -1;
-    int max_locations_distance_table = -1;
-    int max_locations_map_matching = -1;
-    int max_results_nearest = -1;
-    int max_radius_nearest = -1; // in meters
+    int max_locations_trip = UNLIMITED;
+    int max_locations_viaroute = UNLIMITED;
+    int max_locations_distance_table = UNLIMITED;
+    int max_locations_map_matching = UNLIMITED;
+    int max_results_nearest = UNLIMITED;
+
+    double max_radius_when_bearings = UNLIMITED; // in meters
 
     bool use_shared_memory = true;
 };
--- a/include/engine/engine_config.hpp
+++ b/include/engine/engine_config.hpp
@@ -51,6 +51,7 @@ namespace engine
  *  - Route
  *  - Table
  *  - Match
+ *  - Nearest
  *
  * In addition, shared memory can be used for datasets loaded with osrm-datastore.
  *
@@ -65,6 +66,9 @@ struct EngineConfig final
     int max_locations_viaroute = -1;
     int max_locations_distance_table = -1;
     int max_locations_map_matching = -1;
+    int max_results_nearest = -1;
+    int max_radius_nearest = -1; // in meters
+
     bool use_shared_memory = true;
 };
 }
--- a/include/engine/plugins/plugin_base.hpp
+++ b/include/engine/plugins/plugin_base.hpp
@@ -5,6 +5,7 @@
 #include "engine/datafacade/datafacade_base.hpp"
 #include "engine/phantom_node.hpp"
 #include "engine/status.hpp"
+#include "engine/engine_config.hpp"
 
 #include "util/coordinate.hpp"
 #include "util/coordinate_calculation.hpp"
@@ -27,7 +28,8 @@ class BasePlugin
 {
   protected:
     datafacade::BaseDataFacade &facade;
-    BasePlugin(datafacade::BaseDataFacade &facade_) : facade(facade_) {}
+    const double max_radius_when_bearings;
+    BasePlugin(datafacade::BaseDataFacade &facade_, const double max_radius_when_bearings_) : facade(facade_), max_radius_when_bearings(max_radius_when_bearings_) {}
 
     bool CheckAllCoordinates(const std::vector<util::Coordinate> &coordinates)
     {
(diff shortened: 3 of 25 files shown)


danpat:
@daniel-j-h So, I approached this from a couple of different angles, and settled on adding:

  - a `max_results` limit for the Nearest plugin, similar to all other plugins
  - a general `max_radius_when_bearings` value - this restricts the rtree search when a bearing filter is used - which was the problematic case for the nearest plugin, and has the same potential for the other plugins as well.

If no bearing filter is supplied, then I don't think we need to enforce a radius limit - we'll just select the nearest edges.  The only situation where it might search a long way is the `NearestPhantomNodeWithAlternativeFromBigComponent`, but that seems pretty unlikely as well, given that *most* edges are in the "big component".

As far as I can see, an RTree search with a bearing is the only really problematic case, and this change should put a limit on how bad that can be.  I've set the default value to 1000m, and that *only* applies when a bearing filter is supplied.

Still todo:

 - [ ] clang-format - I specifically didn't do this yet because every time I do, it seems to introduce a bunch of noise in the changeset.  Please feel free to go ahead and run it.

src/engine/plugins/nearest.cpp
```diff
@@ -44,14 +44,22 @@ Status NearestPlugin::HandleRequest(const api::NearestParameters &params,
...
+            boost::algorithm::clamp(*constraint_params.radiuses[0], 0., max_radius_m);
+    }
 
-    if (constraint_params.radiuses.empty())
-        constraint_params.radiuses.push_back(boost::make_optional(max_radius_m));
-    else if (constraint_params.radiuses[0])
-        boost::algorithm::clamp(*constraint_params.radiuses[0], 0., max_radius_m);
+    if (max_results != -1)
+    {
+        constraint_params.number_of_results = boost::algorithm::clamp(params.number_of_results, 0, max_results);
```
daniel-j-h: All other plugins return a `TooBig` status code in case the user-provided params are over the global system's limit (e.g. [viaroute](https://github.com/Project-OSRM/osrm-backend/blob/938dff011fb52fedb8b787d7403b0c8fb1d67f5a/src/engine/plugins/viaroute.cpp#L35-L43)). I think in the case of `max_results` we should do the same, since this specifically is `TooBig`'s raison d’être.

unit_tests/library/nearest.cpp
```diff
@@ -43,6 +43,67 @@ BOOST_AUTO_TEST_CASE(test_nearest_response)
...
+
+    json::Object result;
+    const auto rc = osrm.Nearest(params, result);
+    BOOST_REQUIRE(rc == Status::Ok);
+
+    const auto code = result.values.at("code").get<json::String>().value;
+    BOOST_CHECK_EQUAL(code, "Ok");
+
+    const auto &waypoints = result.values.at("waypoints").get<json::Array>().values;
+    BOOST_CHECK(waypoints.size() == 2); // make sure our results were capped at max_results_nearest
```
daniel-j-h: As outlined above I think this should return the `TooBig` status code instead of silently capping results.

unit_tests/library/limits.cpp
```diff
@@ -136,4 +138,45 @@ BOOST_AUTO_TEST_CASE(test_match_limits)
...
+    json::Object result2;
+    const auto rc2 = osrm.Nearest(params, result2);
+    BOOST_REQUIRE(rc2 == Status::Error);
+    const auto code2 = result2.values.at("code").get<json::String>().value;
+    BOOST_CHECK_EQUAL(code2, "TooBig");
+
+}
+
+
+
```
daniel-j-h: Don't we want to add a bearing 0,0 test here?

docs/http.md
```diff
@@ -69,7 +69,7 @@ Where the elements follow the following format:
 
 | Element    | Values                                                 |
 |------------|--------------------------------------------------------|
-|bearing     |`{value},{range}` `integer 0 .. 360,integer 0 .. 180`  |
+|bearing     |`{value},{range}` `integer 0 .. 360,integer 1 .. 180`  |
```
MoKob: This is an API change. Are bearing range-0 values always impossible queries? If so, why don't we simply return directly and keep the API valid for now?

include/engine/plugins/plugin_base.hpp
```diff
@@ -179,18 +200,34 @@ class BasePlugin
...
                 {
-                    phantom_nodes[i] = facade.NearestPhantomNodes(parameters.coordinates[i],
-                                                                  number_of_results,
-                                                                  *parameters.radiuses[i],
-                                                                  parameters.bearings[i]->bearing,
-                                                                  parameters.bearings[i]->range);
+                    phantom_nodes[i] = facade.NearestPhantomNodes(
+                        parameters.coordinates[i],
+                        number_of_results,
+                        max_radius_when_bearings != EngineConfig::UNLIMITED
```
MoKob: duplicated code

include/engine/plugins/plugin_base.hpp
```diff
@@ -179,18 +200,34 @@ class BasePlugin
...
                 }
                 else
                 {
-                    phantom_nodes[i] = facade.NearestPhantomNodes(parameters.coordinates[i],
-                                                                  number_of_results,
-                                                                  parameters.bearings[i]->bearing,
-                                                                  parameters.bearings[i]->range);
+                    if (max_radius_when_bearings != EngineConfig::UNLIMITED)
+                    {
+                        phantom_nodes[i] =
```
MoKob: are these two functions actually different or use each other? I don't see why we split here in two functions.
In addition, why is `!= UNLIMITED` called without radius and `== UNLIMITED` called with radius?

include/engine/plugins/plugin_base.hpp
```diff
@@ -179,18 +200,34 @@ class BasePlugin
...
+                                                       number_of_results,
+                                                       parameters.bearings[i]->bearing,
+                                                       parameters.bearings[i]->range);
+                    }
+                    else
+                    {
+                        phantom_nodes[i] =
+                            facade.NearestPhantomNodes(parameters.coordinates[i],
+                                                       number_of_results,
+                                                       max_radius_when_bearings,
```
MoKob: This would always be EngineConfig::UNLIMITED, according to the above check

include/engine/plugins/plugin_base.hpp
```diff
@@ -242,17 +279,31 @@ class BasePlugin
                     phantom_node_pairs[i] =
                         facade.NearestPhantomNodeWithAlternativeFromBigComponent(
                             parameters.coordinates[i],
-                            *parameters.radiuses[i],
+                            max_radius_when_bearings != EngineConfig::UNLIMITED
```
MoKob: again, same code

include/engine/plugins/plugin_base.hpp
```diff
@@ -242,17 +279,31 @@ class BasePlugin
...
                 }
                 else
                 {
-                    phantom_node_pairs[i] =
-                        facade.NearestPhantomNodeWithAlternativeFromBigComponent(
-                            parameters.coordinates[i],
-                            parameters.bearings[i]->bearing,
-                            parameters.bearings[i]->range);
+                    if (max_radius_when_bearings == EngineConfig::UNLIMITED)
+                    {
```
MoKob: same question as above, only `==` is correct here. Still, why the split into two functions? Why not call it with UNLIMITED?

src/engine/engine_config.cpp
```diff
@@ -15,11 +15,17 @@ bool EngineConfig::IsValid() const
...
-        (max_locations_distance_table == -1 || max_locations_distance_table > 2) &&
-        (max_locations_map_matching == -1 || max_locations_map_matching > 2) &&
-        (max_locations_trip == -1 || max_locations_trip > 2) &&
-        (max_locations_viaroute == -1 || max_locations_viaroute > 2);
+    const auto unlimited_or_more_than = [](const int v, const int limit) {
+        return v == -1 || v > limit;
+    };
+
+    const bool limits_valid = unlimited_or_more_than(max_locations_distance_table, 2) &&
+                              unlimited_or_more_than(max_locations_distance_table, 2) &&
```
MoKob: copy paste duplicated check

include/engine/engine_config.hpp
```diff
@@ -60,11 +64,17 @@ struct EngineConfig final
...
-    int max_locations_viaroute = -1;
-    int max_locations_distance_table = -1;
-    int max_locations_map_matching = -1;
+    int max_locations_trip = UNLIMITED;
+    int max_locations_viaroute = UNLIMITED;
+    int max_locations_distance_table = UNLIMITED;
+    int max_locations_map_matching = UNLIMITED;
+    int max_results_nearest = UNLIMITED;
+
+    double max_radius_when_bearings = UNLIMITED; // in meters
```
daniel-j-h: Why is this a double? Shouldn't a 32 bit int be enough meters for anybody?

MoKob: This is due to our `static_r_tree` functions being overloaded for `double` (radius) and `int` max_entries... Error-prone yeah...

daniel-j-h: Ouch yeah I can totally see us running into this..

daniel-j-h:
Superseded by
- https://github.com/Project-OSRM/osrm-backend/pull/2881
- https://github.com/Project-OSRM/osrm-backend/pull/2899

Issue #2357, opened 2016-05-02
Avoid copying EdgeBasedNode into PhantomNode


TheMarex:
Once https://github.com/Project-OSRM/osrm-backend/issues/1987 has landed we can actually reference the `EdgeBasedNode`s directly.

This would mean any access to data stored there could cause a page fault. However we access this data rarely and the chance of this getting paged out again after being accessed in the StaticRTree is low.

This would enable:

- Small `PhantomNode` (basically only offsets + snapped coordinate + pointer to `EdgeBasedNode`)
- Decoupling `Hint` from `PhantomNode`. If we expose the leaf index + edge index from the `StaticRTree` that means we only encode these two values and can reconstruct the `PhantomNode` from that

We need to benchmark this to make sure my hunch about page faults is not wrong (and we don't actually kill performance with this).

danpat:
#1987 has landed, someone can pick this up if they want.

karenzshea:
Implicit to this task is also refactoring the static rtree to only contain references to `EdgeBasedNode`s. Phantom Nodes already only use references to edge results from rtree's `Nearest()` method.

PR #2880, opened 2016-09-08
Refactor PhantomNode to only contain EdgeBaseNode references

karenzshea:
# Issue

Testing changes for https://github.com/Project-OSRM/osrm-backend/issues/2357. Because we're accessing edge data via an mmap in the static rtree, we can avoid copying edge data into phantom nodes and only store references instead. It also enables us to refactor how Hints to reconstruct phantom nodes.

This may or may not ever be mergeable though, as the phantom node reads from disk need to be benchmarked for issues with page faults.
## Tasklist
- [ ] Figure out why the size of `PhantomNode` increased.
- [ ] Refactor Hints to reconstruct rather than copy PhantomNodes
  -  [ ] Benchmark map matching
- [ ] review
- [ ] adjust for for comments
## Requirements / Relations

none

### DIFF
--- a/include/engine/api/base_api.hpp
+++ b/include/engine/api/base_api.hpp
@@ -49,7 +49,7 @@ class BaseAPI
     util::json::Object MakeWaypoint(const PhantomNode &phantom) const
     {
         return json::makeWaypoint(phantom.location,
-                                  facade.GetNameForID(phantom.name_id),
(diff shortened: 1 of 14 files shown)


daniel-j-h:
Came across this while doing a sweep over all pull requests and issues. What's the state here? It needs a rebase for sure. @danpat you worked on something similar for the rtree?

karenzshea:
Stale.

PR #2888, opened 2016-09-09
Reworks Restriction Whitelist / Blacklist, resolves #2833

### LINKED ISSUE #2833: Refactor restriction parsing to consider all tags on a relation
From https://github.com/Project-OSRM/osrm-backend/issues/2802#issuecomment-241796627

> Today in some circumstances (restriction=no_* lead to false and then restriction:<*>=only_* leads to true) the last restriction value of type only_ will override the previous ones regarding the variable is_only_restriction (which will be set as true). With this patch it will be even more visible and the last restriction will always override is_only_restriction. To solve this already existing case will require some additional steps/a rewriting of the whole algorithm. But at a first step and to close this issue ignoring the second if and adding only the else statement will solve the current issue (i.e. ignore not-supported restrictions) without changing the existing behavior.

Turn restriction extraction currently has some "last processed restriction wins" logic. My understanding is that in a scenario where a relation is tagged with both a `restriction=no_turn` and a `restriction:bicycle=only_right_turn`, if the `restriction=no_turn` is looped over last, the relation will be understood as a restricted turn, even if it's actually a valid turn for bicycles.

Solution would be to holistically consider tags on a relation to determine a restriction for the given profile.

daniel-j-h:
# Issue

For #2833.

Takes a stricter aproach for whitelisting / blacklisting restrictions:

- uses `restriction=`
- uses more specific `restriction:<type>=`
- uses `except=<type>` to invert

Where `type` is the type of transportation to restrict, e.g. `motorcar`.

## Tasklist
 - [x] review
 - [x] adjust for for comments

### DIFF
--- a/profiles/bicycle.lua
+++ b/profiles/bicycle.lua
@@ -12,7 +12,7 @@ access_tag_restricted = { ["destination"] = true, ["delivery"] = true }
 access_tags_hierarchy = { "bicycle", "vehicle", "access" }
 cycleway_tags = {["track"]=true,["lane"]=true,["opposite"]=true,["opposite_lane"]=true,["opposite_track"]=true,["share_busway"]=true,["sharrow"]=true,["shared"]=true }
 service_tag_restricted = { ["parking_aisle"] = true }
-restriction_exception_tags = { "bicycle", "vehicle", "access" }
+restrictions = { "bicycle" }
 unsafe_highway_list = { ["primary"] = true, ["secondary"] = true, ["tertiary"] = true, ["primary_link"] = true, ["secondary_link"] = true, ["tertiary_link"] = true}
 
 local default_speed = 15
@@ -121,8 +121,8 @@ local function parse_maxspeed(source)
     return n
 end
 
-function get_exceptions(vector)
-  for i,v in ipairs(restriction_exception_tags) do
+function get_restrictions(vector)
+  for i,v in ipairs(restrictions) do
     vector:Add(v)
   end
 end
--- a/src/extractor/restriction_parser.cpp
+++ b/src/extractor/restriction_parser.cpp
@@ -29,20 +29,19 @@ RestrictionParser::RestrictionParser(ScriptingEnvironment &scripting_environment
 {
     if (use_turn_restrictions)
     {
-        restriction_exceptions = scripting_environment.GetExceptions();
-        const unsigned exception_count = restriction_exceptions.size();
-        if (exception_count)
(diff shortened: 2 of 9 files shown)


profiles/bicycle.lua
```diff
@@ -12,7 +12,7 @@ access_tag_restricted = { ["destination"] = true, ["delivery"] = true }
 access_tags_hierarchy = { "bicycle", "vehicle", "access" }
 cycleway_tags = {["track"]=true,["lane"]=true,["opposite"]=true,["opposite_lane"]=true,["opposite_track"]=true,["share_busway"]=true,["sharrow"]=true,["shared"]=true }
 service_tag_restricted = { ["parking_aisle"] = true }
-restriction_exception_tags = { "bicycle", "vehicle", "access" }
+restrictions = { "bicycle" }
```
MoKob: here `restrictions` differs from `restriction_exception_tags`, below these two are identical. Why is that?

src/extractor/restriction_parser.cpp
```diff
@@ -51,8 +50,9 @@ RestrictionParser::RestrictionParser(ScriptingEnvironment &scripting_environment
  * Tries to parse a relation as a turn restriction. This can fail for a number of
  * reasons. The return type is a boost::optional<T>.
  *
- * Some restrictions can also be ignored: See the ```get_exceptions``` function
- * in the corresponding profile.
+ * Some restrictions can also be ignored: See the ```get_restrictions``` function
```
MoKob: why is `get_restrictions` describing `exceptions` to said restrictions better than `get_exceptions`?

profiles/bicycle.lua
daniel-j-h: The idea is to only use restriction exceptions in the same way we are using namespaced restrictions. If you allow `restriction:motorcar` you also want to check for `except=motorcar` after all.

Before we had the situation that we were using `restriction:*=` and then were using the exception tag to filter again.

(Also the exceptions list before was wrong e.g. there is no vehicle or access exception http://taginfo.openstreetmap.org/keys/except#values, also no such restrictions http://taginfo.openstreetmap.org/keys/restriction#values).

src/extractor/restriction_parser.cpp
daniel-j-h: The type is the same, yes. As in `restriction:<type>=` or `except=type`.

Well, `except=` also allows for a list of types, as in `except=motorcar,bus` but we handle this below, too.

http://wiki.openstreetmap.org/wiki/Relation:restriction

Issue #2891, opened 2016-09-09
External Memory does not consider references


MoKob:
https://github.com/Project-OSRM/osrm-backend/blob/a289e1cd9607e0cb4bbc6be674effe5bd543571d/src/extractor/extractor_callbacks.cpp#L258-L259 is missing the reference_size for the reservation of memory.

PR #2892, opened 2016-09-09
make sure to reserve enough external memory

MoKob:
# Issue

resolves https://github.com/Project-OSRM/osrm-backend/issues/2891

## Tasklist
 ~~- [ ] add regression / cucumber cases (see docs/testing.md)~~
>  memory overflow is not testable without parsing a huge graph

 - [x] review
 - [x] adjust for for comments


daniel-j-h:
Looks good.

Issue #2896, opened 2016-09-11
Demo server hanging during osrm-extract


danpat:
The demo server has updated to 05c1fe7f46f1903bfac8feb37f0d434efe5c2b57 and is running `osrm-extract`.

Similar to other recent cases, there's only one active thread, and it seems to be stuck in a loop.  Here are two stack traces of the working thread taken a few seconds apart.  Unfortunately it looks like we don't have debug symbols enabled on the demo server (we should change that), so no line numbers.

```
(gdb) bt
#0  0x000000000047442b in osrm::extractor::CompressedEdgeContainer::HasEntryForID(unsigned int) const ()
#1  0x00000000004a88aa in osrm::extractor::guidance::IntersectionGenerator::GetConnectedRoads(unsigned int, unsigned int) const ()
#2  0x00000000004aa1f2 in osrm::extractor::guidance::IntersectionGenerator::AdjustForJoiningRoads(unsigned int, std::vector<osrm::extractor::guidance::ConnectedRoad, std::allocator<osrm::extractor::guidance::ConnectedRoad> >) const ()
#3  0x00000000004aa54a in osrm::extractor::guidance::IntersectionGenerator::operator()(unsigned int, unsigned int) const ()
#4  0x0000000000505a82 in osrm::extractor::guidance::SliproadHandler::operator()(unsigned int, unsigned int, std::vector<osrm::extractor::guidance::ConnectedRoad, std::allocator<osrm::extractor::guidance::ConnectedRoad> >) const::{lambda(unsigned int, osrm::extractor::guidance::ConnectedRoad const&, unsigned int&)#1}::operator()(unsigned int, osrm::extractor::guidance::ConnectedRoad const&, unsigned int&) const [clone .isra.36] ()
#5  0x00000000005063e5 in osrm::extractor::guidance::SliproadHandler::operator()(unsigned int, unsigned int, std::vector<osrm::extractor::guidance::ConnectedRoad, std::allocator<osrm::extractor::guidance::ConnectedRoad> >) const ()
#6  0x00000000004a61d6 in osrm::extractor::guidance::TurnAnalysis::assignTurnTypes(unsigned int, unsigned int, std::vector<osrm::extractor::guidance::ConnectedRoad, std::allocator<osrm::extractor::guidance::ConnectedRoad> >) const ()
[...]

PR #2898, opened 2016-09-12
fix looping in sliproad handler for lanes

MoKob:
# Issue

Finding a next intersection within a dedicated loop can generate endless looping: https://github.com/Project-OSRM/osrm-backend/issues/2896

## Tasklist
~~add regression / cucumber cases (see docs/testing.md)~~
 - [x] review
 - [ ] adjust for for comments

## Requirements / Relations

### DIFF
--- a/src/extractor/guidance/sliproad_handler.cpp
+++ b/src/extractor/guidance/sliproad_handler.cpp
@@ -37,9 +37,9 @@ SliproadHandler::SliproadHandler(const IntersectionGenerator &intersection_gener
 // included for interface reasons only
 bool SliproadHandler::canProcess(const NodeID /*nid*/,
                                  const EdgeID /*via_eid*/,
-                                 const Intersection & /*intersection*/) const
+                                 const Intersection &intersection) const
 {
-    return true;
+    return intersection.size() > 2;
 }
 
 Intersection SliproadHandler::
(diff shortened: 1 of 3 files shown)


src/extractor/guidance/turn_analysis.cpp
```diff
@@ -94,7 +94,8 @@ Intersection TurnAnalysis::assignTurnTypes(const NodeID from_nid,
         }
     }
     // Handle sliproads
-    intersection = sliproad_handler(from_nid, via_eid, std::move(intersection));
+    if( sliproad_handler.canProcess(from_nid,via_eid,intersection) )
```
daniel-j-h: clang-format

src/extractor/guidance/sliproad_handler.cpp
```diff
@@ -82,6 +89,10 @@ operator()(const NodeID, const EdgeID source_edge_id, Intersection intersection)
                 intersection_node_id, intersection[1], intersection_node_one);
             const auto intersection_following_index_two = findNextIntersectionForRoad(
                 intersection_node_id, intersection[2], intersection_node_two);
+            // in case of broken roads, we return
+            if (intersection_following_index_one.empty() ||
+                intersection_following_index_two.empty())
```
daniel-j-h: When can this happen?

MoKob: Thats the result of the above addition. Theoretically, we should never be able to reach this if not for other problems. But right now our demo-server is seeing endless loops and the early return from the in my view only possible location results in empty state here.

PR #2899, opened 2016-09-12
Adds a limit for number of results returned in Nearest service, resves #2872

daniel-j-h:
# Issue

For https://github.com/Project-OSRM/osrm-backend/issues/2872.

## Tasklist
 - [x] review
 - [x] adjust for for comments

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,8 @@
       - Fixed an issue that would detect turning circles as sliproads
       - Fixed a bug where post-processing instructions (e.g. left + left -> uturn) could result in false pronunciations
       - Fixes a bug where a bearing range of zero would cause exhaustive graph traversals
+    - Infrastructure:
+      - Adds a feature to limit results in nearest service with a default of 100 in `osrm-routed`
 
 # 5.3.0
   Changes from 5.3.0-rc.3
--- a/include/engine/engine_config.hpp
+++ b/include/engine/engine_config.hpp
@@ -51,6 +51,7 @@ namespace engine
  *  - Route
  *  - Table
  *  - Match
+ *  - Nearest
  *
  * In addition, shared memory can be used for datasets loaded with osrm-datastore.
  *
@@ -65,6 +66,7 @@ struct EngineConfig final
     int max_locations_viaroute = -1;
     int max_locations_distance_table = -1;
     int max_locations_map_matching = -1;
+    int max_results_nearest = -1;
     bool use_shared_memory = true;
 };
 }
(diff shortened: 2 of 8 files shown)


CHANGELOG.md
```diff
@@ -21,6 +21,8 @@
       - Fixed an issue that would detect turning circles as sliproads
       - Fixed a bug where post-processing instructions (e.g. left + left -> uturn) could result in false pronunciations
       - Fixes a bug where a bearing range of zero would cause exhaustive graph traversals
+    - Infrastructure:
+      - Adds a feature to limit results in nearest service with a default of 100 in `osrm-routed`
```
MoKob: a default of 100 would be an `abi` change, right?

daniel-j-h: What do you mean with abi change?

`libosrm`'s `EngineConfig` now provides a new limit ([unlimited by default](https://github.com/Project-OSRM/osrm-backend/blob/a618f028fb802ceea67c1a4df189b92b4758e91a/include/engine/engine_config.hpp#L69)). It's only the `osrm-routed` binary setting the limit of `100` (and the user can still change this with [`--max-nearest-size`](https://github.com/Project-OSRM/osrm-backend/blob/a618f028fb802ceea67c1a4df189b92b4758e91a/src/tools/routed.cpp#L105-L107).

We don't guarantee ABI stability at all:
https://github.com/Project-OSRM/osrm-backend/blob/master/docs/releasing.md

Only API stability. So adding a new member in the `libosrm` header should be fine.

MoKob: Seems like I have been a bit too careful here. You are right. That should be ok.

daniel-j-h: It's actually a good point: we don't guarantee ABI stability.

@jcoupey I know you started integrating `libosrm` for VROOM. Is this something that could bite you or are you compiling `libosrm` from source anyway? (Or do you expect your users to do so?)

jcoupey: @daniel-j-h thanks for asking. `libosrm` should be compiled and installed by the user to be used within VROOM, so I guess this won't make a difference on my side?

daniel-j-h: Yes, this is fine :+1: 

What does not work is e.g. linking VROOM against `libosrm` v5.3. And then upgrading `libosrm` to v5.4. If you want to upgrade `libosrm` you have to compile/link it in again.

For the record, I started a C wrapper for FFI bindings and ABI stability some time ago, just in case someone later reads this and needs those guarantees: https://github.com/daniel-j-h/libosrmc

PR #2902, opened 2016-09-12
Test for PR #2898

oxidase:
# Issue

Test checks that osrm-extract terminates for
http://www.openstreetmap.org/way/198481519 #2896 

## Tasklist
 - [x] review
 - [ ] adjust for for comments

## Requirements / Relations
The issue is fixed by PR #2898

/cc @danpat @MoKob


danpat:
👍 This checks out locally and fails without the fix in #2898.  Merging!

Issue #2903, opened 2016-09-13
False Positives in Slip Roads


MoKob:
<img width="849" alt="screen shot 2016-09-13 at 13 02 08" src="https://cloud.githubusercontent.com/assets/12932279/18471950/80456a0a-79b4-11e6-8a06-89a3bd96f577.png">

Right now we simply check connectivity for sliproads. It turns out that this can lead to false positives (see example above).

MoKob:
Landed in Master/rc.4

PR #2904, opened 2016-09-13
fix false positives in slip roads

MoKob:
# Issue

https://github.com/Project-OSRM/osrm-backend/issues/2903

## Tasklist
 - [x] review
 - [x] adjust for for comments

### DIFF
--- a/src/extractor/guidance/sliproad_handler.cpp
+++ b/src/extractor/guidance/sliproad_handler.cpp
@@ -139,7 +139,6 @@ operator()(const NodeID, const EdgeID source_edge_id, Intersection intersection)
         return intersection;
 
     const auto source_edge_data = node_based_graph.GetEdgeData(source_edge_id);
-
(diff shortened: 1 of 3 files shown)


src/extractor/guidance/sliproad_handler.cpp
```diff
@@ -192,6 +191,21 @@ operator()(const NodeID, const EdgeID source_edge_id, Intersection intersection)
                 return intersection;
             }(intersection_node_id);
 
+            const auto link_data = node_based_graph.GetEdgeData(road.turn.eid);
+            // Check if the road continues here
+            const bool is_through_street =
+                target_intersection.end() !=
+                std::find_if(target_intersection.begin()+1,
```
daniel-j-h: Target intersection can never be empty, correct?

Also clang-format (or is the ()+1 already formatted?).

PR #2900, opened 2016-09-12
Reduce NewName Instructructions / Name Changes

daniel-j-h:
With @karenzshea's [name / ref split](https://github.com/Project-OSRM/osrm-backend/issues/2857) in master we want to make use of it and reduce `NewName` instructions when ever possible. This is a first step towards https://github.com/Project-OSRM/osrm-backend/issues/2744 by using the already existing name change heuristic from the extractor now in post-processing as well.

## Tasklist
 - [x] adapt tests currently failing
 - [x] incorporate and adapt scenarios from https://github.com/Project-OSRM/osrm-backend/pull/2845
 - [x] make `requiresNameAnnounced` usable for pre- as well as post-processing (no `SuffixTable`)
 - [x] adapt raw name id checks in guidance to `requiresNameAnnounced` https://github.com/Project-OSRM/osrm-backend/issues/2909
 - [x] think about additional post-processing step over all `RouteStep`s further reducing `NewName`s - not worth it
 - [x] review
 - [x] adjust for for comments

PR #2905, opened 2016-09-13
Fix/refs in name table

MoKob:
# Issue

Follow-up on https://github.com/Project-OSRM/osrm-backend/issues/2704

## Tasklist
 ~~add regression / cucumber cases (see docs/testing.md)~~
we already have failing items over in https://github.com/Project-OSRM/osrm-backend/pull/2900
 - [x] review
 - [x] adjust for for comments

### DIFF
--- a/src/extractor/extraction_containers.cpp
+++ b/src/extractor/extraction_containers.cpp
@@ -110,7 +110,8 @@ ExtractionContainers::ExtractionContainers()
     // Check if stxxl can be instantiated
     stxxl::vector<unsigned> dummy_vector;
 
-    // Insert three empty strings offsets for name, destination and pronunciation
+    // Insert three empty strings offsets for name, ref, destination and pronunciation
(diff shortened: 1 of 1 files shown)


daniel-j-h:
Capturing from chat: this was the reason tests were failing in https://github.com/Project-OSRM/osrm-backend/pull/2900.
Huge thanks @MoKob for spotting this :bow:.

src/extractor/extraction_containers.cpp
```diff
@@ -110,7 +110,8 @@ ExtractionContainers::ExtractionContainers()
     // Check if stxxl can be instantiated
     stxxl::vector<unsigned> dummy_vector;
 
-    // Insert three empty strings offsets for name, destination and pronunciation
+    // Insert three empty strings offsets for name, ref, destination and pronunciation
```
danpat: s/three/four/

Issue #2882, opened 2016-09-08
Dynamic Edge Update QA: Dump / Warn on High Delta


daniel-j-h:
We allow users to dynamically update edge weights via:
https://github.com/Project-OSRM/osrm-backend/wiki/Traffic.

The first time an update happens we update the internal speeds (based on profiles) with the user-provided speed files. This happens in [the contractor](https://github.com/Project-OSRM/osrm-backend/blob/eebda5c0a5c1260453e408f4ad02100066a69569/src/contractor/contractor.cpp).

It would be great to have an (optional) check for when the update changes speeds by either
- a high relative delta (as in the new speed is twice as high) and / or
- a high absolute delta (as in the new speed is 100km/h higher)

(for both too high / low deltas --- maybe configurable)

This would allow us to catch either mis-taggings or errors on the speed profile side.

MoKob:
@daniel-j-h twice as high I wouldn't find that confusing. Consider a traffic jam on the motorway. When you can normally drive 200km/h and suddenly go with less than 20km/h. This would even give you a factor of 10.
So you can check for improvements (e.g. limit the maximum you would want to see in speeds), but limiting the minimum in this way without a lot of false positives is hard, I guess.

daniel-j-h:
Yes you are right. The minimum is more important for verifying accumulated speeds, as in historic speed profiles. @karenzshea is looking into this and for a first iteration checking for speed improvements that are "too high" is a good start.

karenzshea:
Initial pass at this has been implemented.

PR #2906, opened 2016-09-13
[feature] Enable a check on edge weight updates

karenzshea:
# Issue

Ref https://github.com/Project-OSRM/osrm-backend/issues/2882

This is a QA feature that runs a check on all edge weight updates that happen during contraction that the update is a reasonable update based on the previous value. ~~It's still a todo to determine a reliable heuristic for what a reasonable update is.~~ The flag now takes a value, a `double` value that's used as a factor of change from old to new weights. It's still on the table as to how to implement an absolute value in seconds as input for the flag. Should this be done with a completely different flag?

## Tasklist
 - [x] Put feature behind a flag `--verify-weights` or so
 - [x] Print the corresponding filename from datasource_names in log output
 - [x] review
 - [x] adjust for for comments

### DIFF
--- a/features/options/contract/version.feature
+++ b/features/options/contract/version.feature
@@ -6,7 +6,7 @@ Feature: osrm-contract command line options: version
 
     Background:
         Given the profile "testbot"
-    
+
     Scenario: osrm-contract - Version, short
         When I run "osrm-contract --v"
         Then stderr should be empty
--- a/include/contractor/contractor.hpp
+++ b/include/contractor/contractor.hpp
@@ -89,7 +89,8 @@ class Contractor
                           const std::string &geometry_filename,
                           const std::string &datasource_names_filename,
                           const std::string &datasource_indexes_filename,
-                          const std::string &rtree_leaf_filename);
+                          const std::string &rtree_leaf_filename,
+                          const double log_edge_updates_factor);
 };
 }
 }
--- a/include/contractor/contractor_config.hpp
+++ b/include/contractor/contractor_config.hpp
@@ -73,6 +73,7 @@ struct ContractorConfig
     bool use_cached_priority;
 
     unsigned requested_num_threads;
+    double log_edge_updates_factor;
 
     // A percentage of vertices that will be contracted for the hierarchy.
     // Offers a trade-off between preprocessing and query time.
--- a/src/contractor/contractor.cpp
+++ b/src/contractor/contractor.cpp
@@ -81,6 +81,41 @@ inline EdgeWeight distanceAndSpeedToWeight(double distance_in_meters, double spe
     return std::max<EdgeWeight>(1, static_cast<EdgeWeight>(std::round(duration * 10)));
 }
 
+// Returns updated edge weight
+template <class IterType>
+EdgeWeight getNewWeight(IterType speed_iter,
+                        const double &segment_length,
+                        const std::vector<std::string> &segment_speed_filenames,
+                        const EdgeWeight old_weight,
+                        const double log_edge_updates_factor)
+{
+    const auto new_segment_weight =
+        (speed_iter->speed_source.speed > 0)
+            ? distanceAndSpeedToWeight(segment_length, speed_iter->speed_source.speed)
+            : INVALID_EDGE_WEIGHT;
+    // the check here is enabled by the `--edge-weight-updates-over-factor` flag
+    // it logs a warning if the new weight exceeds a heuristic of what a reasonable weight update is
+    if (log_edge_updates_factor > 0 && old_weight != 0)
+    {
+        auto new_secs = new_segment_weight / 10.0;
+        auto old_secs = old_weight / 10.0;
+        auto approx_original_speed = (segment_length / old_secs) * 3.6;
+        if (old_weight >= (new_segment_weight * log_edge_updates_factor))
+        {
+            auto speed_file = segment_speed_filenames.at(speed_iter->speed_source.source - 1);
+            util::SimpleLogger().Write(logWARNING)
+                << "[weight updates] Edge weight update from " << old_secs << "s to " << new_secs << "s"
+                << ". New speed: " << speed_iter->speed_source.speed << " kph"
+                << ". Old speed: " << approx_original_speed << " kph"
+                << ". Segment length: " << segment_length << " m"
+                << ". Segment: " << speed_iter->segment.from << ","
+                << speed_iter->segment.to << " based on " << speed_file;
+        }
+    }
+
+    return new_segment_weight;
+}
+
 int Contractor::Run()
 {
 #ifdef WIN32
@@ -113,7 +148,8 @@ int Contractor::Run()
                                                config.geometry_path,
                                                config.datasource_names_path,
                                                config.datasource_indexes_path,
-                                               config.rtree_leaf_path);
+                                               config.rtree_leaf_path,
+                                               config.log_edge_updates_factor);
 
     // Contracting the edge-expanded graph
 
@@ -370,7 +406,8 @@ EdgeID Contractor::LoadEdgeExpandedGraph(
     const std::string &geometry_filename,
     const std::string &datasource_names_filename,
     const std::string &datasource_indexes_filename,
-    const std::string &rtree_leaf_filename)
+    const std::string &rtree_leaf_filename,
+    const double log_edge_updates_factor)
 {
     if (segment_speed_filenames.size() > 255 || turn_penalty_filenames.size() > 255)
         throw util::exception("Limit of 255 segment speed and turn penalty files each reached");
@@ -556,10 +593,12 @@ EdgeID Contractor::LoadEdgeExpandedGraph(
                 extractor::QueryNode *u;
                 extractor::QueryNode *v;
 
+                // forward_packed_geometry_id is the index id for geometry, weight and duration of the segments
                 if (leaf_object.forward_packed_geometry_id != SPECIAL_EDGEID)
                 {
                     const unsigned forward_begin =
                         m_geometry_indices.at(leaf_object.forward_packed_geometry_id);
+                    const auto current_fwd_segment = &(m_geometry_list[forward_begin + leaf_object.fwd_segment_position]);
 
                     if (leaf_object.fwd_segment_position == 0)
                     {
@@ -574,23 +613,23 @@ EdgeID Contractor::LoadEdgeExpandedGraph(
                                                    leaf_object.fwd_segment_position - 1]
                                        .node_id]);
                         v = &(internal_to_external_node_map
-                                  [m_geometry_list[forward_begin + leaf_object.fwd_segment_position]
-                                       .node_id]);
+                                  [current_fwd_segment->node_id]);
                     }
                     const double segment_length = util::coordinate_calculation::greatCircleDistance(
                         util::Coordinate{u->lon, u->lat}, util::Coordinate{v->lon, v->lat});
 
                     auto forward_speed_iter =
                         find(segment_speed_lookup, Segment{u->node_id, v->node_id});
+
                     if (forward_speed_iter != segment_speed_lookup.end())
                     {
-                        auto new_segment_weight =
-                            (forward_speed_iter->speed_source.speed > 0)
-                                ? distanceAndSpeedToWeight(segment_length,
-                                                           forward_speed_iter->speed_source.speed)
-                                : INVALID_EDGE_WEIGHT;
-                        m_geometry_list[forward_begin + leaf_object.fwd_segment_position].weight =
-                            new_segment_weight;
+                        const auto new_segment_weight = getNewWeight(
+                            forward_speed_iter,
+                            segment_length,
+                            segment_speed_filenames,
+                            current_fwd_segment->weight,
+                            log_edge_updates_factor);
+                        current_fwd_segment->weight = new_segment_weight;
                         m_geometry_datasource[forward_begin + leaf_object.fwd_segment_position] =
                             forward_speed_iter->speed_source.source;
 
@@ -603,6 +642,7 @@ EdgeID Contractor::LoadEdgeExpandedGraph(
                         counters[LUA_SOURCE] += 1;
                     }
                 }
+                // reverse_packed_geometry_id is the index id for geometry, weight and duration of the segment
                 if (leaf_object.reverse_packed_geometry_id != SPECIAL_EDGEID)
                 {
                     const unsigned reverse_begin =
@@ -612,6 +652,7 @@ EdgeID Contractor::LoadEdgeExpandedGraph(
 
                     int rev_segment_position =
                         (reverse_end - reverse_begin) - leaf_object.fwd_segment_position - 1;
+                    const auto current_rev_segment = &(m_geometry_list[reverse_begin + rev_segment_position]);
                     if (rev_segment_position == 0)
                     {
                         u = &(internal_to_external_node_map[leaf_object.v]);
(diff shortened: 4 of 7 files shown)


src/contractor/contractor.cpp
```diff
@@ -95,10 +98,11 @@ EdgeWeight getNewWeight(IterType new_speed_iter, const double segment_length, Ed
     {
         auto newSecs = new_segment_weight / 10;
         auto oldSecs = oldWeight / 10;
```
MoKob: the types here are ints, right. so /10 might result in a loss of precision. I'd prefer /10.0 and a formatted output to not have hundreds of decimals in there.

src/contractor/contractor.cpp
```diff
@@ -95,10 +98,11 @@ EdgeWeight getNewWeight(IterType new_speed_iter, const double segment_length, Ed
...
+        auto speed_file = segment_speed_filenames.at(new_speed_iter->speed_source.source - 1);
```
MoKob: can speed_source.source ever be zero?

MoKob: never mind, missed the check above

karenzshea: I think not. `0`th source is reserved for lua profile set defaults, and where we assign this `source` value [only happens during edge weight updates](https://github.com/Project-OSRM/osrm-backend/blob/861314256ebaa0ce4ec5c1e5259da401d7edf584/src/contractor/contractor.cpp#L233) where `source` is advanced 1 to account for lua.

MoKob: So is it guaranteed that we can never loose coverage for an edge? So if a speed ever gets accessed here, we will never put in an update that would fall back to `lua` for this edge?

src/contractor/contractor.cpp
```diff
@@ -95,10 +98,11 @@ EdgeWeight getNewWeight(IterType new_speed_iter, const double segment_length, Ed
     {
```
MoKob: regarding this change. I am a bit sceptical for the verbosity we might see. If a traffic jam happens, we can easily see a factor of 2.5. But if that jam dissolves, we would get a strong increase again. Can we maybe add some factor here that considers our fallback on lua to check if we increase strongly above a reasonable threshold?

karenzshea: The intended purpose of this flag is more for visibility into speed updates over lua provided defaults, not necessarily changes of live traffic.

features/options/contract/version.feature
```diff
@@ -6,13 +6,6 @@ Feature: osrm-contract command line options: version
 
     Background:
         Given the profile "testbot"
-    
-    Scenario: osrm-contract - Version, short
```
MoKob: Why are we removing this scenario?

include/contractor/contractor.hpp
```diff
@@ -89,7 +89,8 @@ class Contractor
                           const std::string &geometry_filename,
                           const std::string &datasource_names_filename,
                           const std::string &datasource_indexes_filename,
-                          const std::string &rtree_leaf_filename);
+                          const std::string &rtree_leaf_filename,
+                          const double &verify_weights);
```
MoKob: `verify_weights` does not describe anything resembling a `double` to me. Is this a factor? Do I have to do a verification in every contractor? Can you choose a name that describes the purpose of this? Also, can we add a default parameter value to disable checks if not wanted?

In Addition: passing a double as const reference doesn't improve stack-usage. Passing a const value should be fine.

include/contractor/contractor_config.hpp
```diff
@@ -73,6 +73,7 @@ struct ContractorConfig
     bool use_cached_priority;
 
     unsigned requested_num_threads;
+    double verify_weights;
```
MoKob: again, this name does not fit the use-case for me. `verify_weights` sounds like a `bool` to me. My suggestion would be to improve the name and add some commentary on how it is used. Also we need to add a reasonable default.

src/contractor/contractor.cpp
```diff
@@ -81,6 +81,38 @@ inline EdgeWeight distanceAndSpeedToWeight(double distance_in_meters, double spe
     return std::max<EdgeWeight>(1, static_cast<EdgeWeight>(std::round(duration * 10)));
 }
 
+// Returns updated edge weight
+template <class IterType>
+EdgeWeight getNewWeight(IterType new_speed_iter,
+                        const double &segment_length,
+                        const std::vector<std::string> &segment_speed_filenames,
+                        EdgeWeight oldWeight,
+                        const double &verify_weights)
```
MoKob: pass by value, change name (e.g. `acceptable_relative_weight_change_limit`?)

src/contractor/contractor.cpp
```diff
@@ -81,6 +81,38 @@ inline EdgeWeight distanceAndSpeedToWeight(double distance_in_meters, double spe
...
+{
+    auto new_segment_weight =
+        (new_speed_iter->speed_source.speed > 0)
+            ? distanceAndSpeedToWeight(segment_length, new_speed_iter->speed_source.speed)
+            : INVALID_EDGE_WEIGHT;
+    // the check here is enabled by the `--verify-weights` flag
+    // it logs a warning if the new weight exceeds a heuristic of what a reasonable weight update is
+    if (verify_weights > 0)
+    {
+        if ((new_segment_weight < oldWeight) && ((oldWeight / new_segment_weight) > verify_weights))
```
MoKob: `new_segment_weight == 0` -> division by zero

MoKob: Didn't @daniel-j-h recommend on checking absolute values here as well? This check seems very sensitive to rounding issues on very small segments to me.

src/contractor/contractor.cpp
```diff
@@ -370,7 +403,8 @@ EdgeID Contractor::LoadEdgeExpandedGraph(
     const std::string &geometry_filename,
     const std::string &datasource_names_filename,
     const std::string &datasource_indexes_filename,
-    const std::string &rtree_leaf_filename)
+    const std::string &rtree_leaf_filename,
+    const double &verify_weights)
```
MoKob: name, pass by value, default value = deactivated

src/contractor/contractor.cpp
```diff
@@ -582,13 +616,16 @@ EdgeID Contractor::LoadEdgeExpandedGraph(
...
-                        auto new_segment_weight =
-                            (forward_speed_iter->speed_source.speed > 0)
-                                ? distanceAndSpeedToWeight(segment_length,
-                                                           forward_speed_iter->speed_source.speed)
-                                : INVALID_EDGE_WEIGHT;
+                        auto new_segment_weight = getNewWeight(
+                            forward_speed_iter,
+                            segment_length,
+                            segment_speed_filenames,
+                            m_geometry_list[forward_begin + leaf_object.fwd_segment_position]
```
MoKob: since this is used multiple times now, can we name it to improve readability?

features/options/contract/version.feature
karenzshea: Because of the ambiguity between a `-v` flag to mean either `--version` or `--verify-weights` we're removing the short option entirely, and therefore don't need to test for it.

include/contractor/contractor.hpp
karenzshea: :+1: to renaming

src/contractor/contractor.cpp
karenzshea: Where `new_segment_weight` is set, there is already a check that it doesn't equal 0, and if it does it is instead set to `INVALID_EDGE_WEIGHT`. I'll added a check in this if block that it doesn't equal `INVALID_EDGE_WEIGHT`

karenzshea: I've decided to punt for now on an absolute value as another filtering factor.

features/options/contract/version.feature
MoKob: Ok. Follow-Up. Why aren't we choosing a different character than `v` instead? I don't feel removing the option is the way to go. I am thinking of heaps of scripts that contain short parameters in their calls. We would break all of them for this change.

src/contractor/contractor.cpp
MoKob: hm. I don't think that we should rely on some outside check at another location. It might, for some reason, might be considered reasonable to set to zero some point down the line. If it isn't guaranteed here, we should check for it.

karenzshea: If you read like 7 lines up, it's the check that happens while `new_segment_weight` is defined, https://github.com/Project-OSRM/osrm-backend/pull/2906/files/566c96b258efc369132bc10e09df4c2d2fd136e2#diff-5d25f965a92c5af3a4dfd03cf4f7616aR93

MoKob: The check above does not check for `new_segment_weight > 0`. If you have a positive speed and a segment length of zero (which can happen in OSM). The only thing guaranteeing a size unequal to zero is https://github.com/Project-OSRM/osrm-backend/pull/2906/files#diff-5d25f965a92c5af3a4dfd03cf4f7616aL76. It's fine if you insist on not checking for this, but in that case we should at least assert it.

features/options/contract/version.feature
daniel-j-h: Good catch, strictly speaking this is a breaking change and results in us having to wait till 6.0 until we can ship this. Probably makes more sense to search for a different mnemonic.

karenzshea: :+1: I'll restore this when I rename the flag. I was going off a suggestion from someone else that we just remove the short version.

src/contractor/contractor.cpp
karenzshea: :+1:

src/contractor/contractor.cpp
karenzshea: aha, I finally see what you mean. Changed.

include/contractor/contractor.hpp
karenzshea: > Also, can we add a default parameter value to disable checks if not wanted?

As for this, is it enough that the default value is `0` and we only apply this check if the value is greater than `0`?

MoKob: Sure.

src/contractor/contractor.cpp
```diff
@@ -81,6 +81,41 @@ inline EdgeWeight distanceAndSpeedToWeight(double distance_in_meters, double spe
...
+            ? distanceAndSpeedToWeight(segment_length, speed_iter->speed_source.speed)
+            : INVALID_EDGE_WEIGHT;
+    // the check here is enabled by the `--edge-weight-updates-over-factor` flag
+    // it logs a warning if the new weight exceeds a heuristic of what a reasonable weight update is
+    if (log_edge_updates_factor > 0 && new_segment_weight != 0)
+    {
+        auto newSecs = new_segment_weight / 10.0;
+        auto oldSecs = oldWeight / 10.0;
+        auto approx_original_speed = (segment_length / oldSecs) * 3.6;
+        if ((new_segment_weight < oldWeight) && ((oldSecs / newSecs) >= log_edge_updates_factor))
```
daniel-j-h: Why

> (new_segment_weight < oldWeight) && ((oldSecs / newSecs) >= log_edge_updates_factor)

- Why checking for both weight and speed? Can't we simply log cases where `new weight > old weight * f`?
- Can `newSecs` ever be zero resulting in a div-by-zero crash here?

src/contractor/contractor.cpp
```diff
@@ -560,6 +597,7 @@ EdgeID Contractor::LoadEdgeExpandedGraph(
                 {
                     const unsigned forward_begin =
                         m_geometry_indices.at(leaf_object.forward_packed_geometry_id);
+                    auto current_fwd_segment = &(m_geometry_list[forward_begin + leaf_object.fwd_segment_position]);
```
daniel-j-h: Read-only? => `const`

src/contractor/contractor.cpp
```diff
@@ -612,6 +650,7 @@ EdgeID Contractor::LoadEdgeExpandedGraph(
 
                     int rev_segment_position =
                         (reverse_end - reverse_begin) - leaf_object.fwd_segment_position - 1;
+                    auto current_rev_segment = &(m_geometry_list[reverse_begin + rev_segment_position]);
```
daniel-j-h: Same

src/contractor/contractor.cpp
```diff
@@ -81,6 +81,41 @@ inline EdgeWeight distanceAndSpeedToWeight(double distance_in_meters, double spe
...
+    auto new_segment_weight =
+        (speed_iter->speed_source.speed > 0)
+            ? distanceAndSpeedToWeight(segment_length, speed_iter->speed_source.speed)
+            : INVALID_EDGE_WEIGHT;
+    // the check here is enabled by the `--edge-weight-updates-over-factor` flag
+    // it logs a warning if the new weight exceeds a heuristic of what a reasonable weight update is
+    if (log_edge_updates_factor > 0 && oldWeight != 0)
+    {
+        auto newSecs = new_segment_weight / 10.0;
+        auto oldSecs = oldWeight / 10.0;
```
MoKob: `old_weight`, `old_secs` and `new_secs`

src/contractor/contractor.cpp
```diff
@@ -81,6 +81,41 @@ inline EdgeWeight distanceAndSpeedToWeight(double distance_in_meters, double spe
...
 
+// Returns updated edge weight
+template <class IterType>
+EdgeWeight getNewWeight(IterType speed_iter,
+                        const double &segment_length,
+                        const std::vector<std::string> &segment_speed_filenames,
+                        EdgeWeight oldWeight,
+                        const double log_edge_updates_factor)
+{
+    auto new_segment_weight =
```
MoKob: `const`

src/contractor/contractor.cpp
```diff
@@ -574,23 +613,23 @@ EdgeID Contractor::LoadEdgeExpandedGraph(
...
                     if (forward_speed_iter != segment_speed_lookup.end())
                     {
-                        auto new_segment_weight =
-                            (forward_speed_iter->speed_source.speed > 0)
-                                ? distanceAndSpeedToWeight(segment_length,
-                                                           forward_speed_iter->speed_source.speed)
-                                : INVALID_EDGE_WEIGHT;
-                        m_geometry_list[forward_begin + leaf_object.fwd_segment_position].weight =
-                            new_segment_weight;
+                        auto new_segment_weight = getNewWeight(
```
MoKob: `const`

src/contractor/contractor.cpp
```diff
@@ -634,13 +675,13 @@ EdgeID Contractor::LoadEdgeExpandedGraph(
...
                     if (reverse_speed_iter != segment_speed_lookup.end())
                     {
-                        auto new_segment_weight =
-                            (reverse_speed_iter->speed_source.speed > 0)
-                                ? distanceAndSpeedToWeight(segment_length,
-                                                           reverse_speed_iter->speed_source.speed)
-                                : INVALID_EDGE_WEIGHT;
-                        m_geometry_list[reverse_begin + rev_segment_position].weight =
-                            new_segment_weight;
+                        auto new_segment_weight = getNewWeight(
```
MoKob: `const`

MoKob:
Looks good. Just needs a squash and can be merged.

src/contractor/contractor.cpp
```diff
@@ -81,6 +81,41 @@ inline EdgeWeight distanceAndSpeedToWeight(double distance_in_meters, double spe
     return std::max<EdgeWeight>(1, static_cast<EdgeWeight>(std::round(duration * 10)));
 }
 
+// Returns updated edge weight
+template <class IterType>
+EdgeWeight getNewWeight(IterType speed_iter,
+                        const double &segment_length,
+                        const std::vector<std::string> &segment_speed_filenames,
+                        EdgeWeight old_weight,
```
MoKob: oh. missed this. this should also be const :(, sorry

MoKob:
@karenzshea seems one case is failing on travis. Seems like a legit failure as well, related to shared libs. Can you have a look at whats going on?

Issue #2907, opened 2016-09-13
Copy Paste Error can break Roundabout Postprocessing


MoKob:
https://github.com/Project-OSRM/osrm-backend/blob/38f5adbe5c70ed0f6ccb447990ac51abef04fecf/include/util/guidance/toolkit.hpp#L115-L116

The copy paste is error is introducing dead code while the roundabout-intersection is missing. This fails to identify entering and exiting a roundabout intersection immediately as `enter` and thus breaks roundabout post-processing.

MoKob:
Landed in master/rc.4

PR #2908, opened 2016-09-13
fix roundabout-intersection on immediate exit

MoKob:
# Issue

https://github.com/Project-OSRM/osrm-backend/issues/2907

## Tasklist
 - [x] review
 - [ ] adjust for for comments

### DIFF
--- a/src/engine/guidance/post_processing.cpp
+++ b/src/engine/guidance/post_processing.cpp
@@ -1,5 +1,5 @@
-#include "engine/guidance/post_processing.hpp"
 #include "extractor/guidance/turn_instruction.hpp"
+#include "engine/guidance/post_processing.hpp"
 
 #include "engine/guidance/assemble_steps.hpp"
 #include "engine/guidance/lane_processing.hpp"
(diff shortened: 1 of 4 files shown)


src/engine/guidance/post_processing.cpp
```diff
@@ -199,20 +199,6 @@ bool setUpRoundabout(RouteStep &step)
...
-        step.maneuver.exit = 1;
-        // prevent futher special case handling of these two.
-        if (instruction.type == TurnType::EnterRotaryAtExit)
-            step.maneuver.instruction.type = TurnType::EnterRotary;
-        else if (instruction.type == TurnType::EnterRoundaboutAtExit)
-            step.maneuver.instruction.type = TurnType::EnterRoundabout;
-        else
-            step.maneuver.instruction.type = TurnType::EnterRoundaboutIntersection;
-    }
-
```
MoKob: this was duplicated code that seems to have gotten back in here in a broken rebase. Removed (see below for correct version)

danpat:
@MoKob Looks good to me.  "At the roundabout go straight" seems like it would be a weird instruction, but if these things really exist, this seems like the right thing to say.

👍

PR #2912, opened 2016-09-14
Improve Turn Function

MoKob:
# Issue

Testing PR, looking at penalising turns through traffic. This essentially adds an additional penalty to left turns if they are going through opposing traffic.

```
       d
       |
a ---- b ---- c
```

Would add an additional penalty for turning left `a-b-d` but not for `d-b-c`.
## Tasklist
- [ ] review
- [ ] adjust for for comments

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,8 @@
       - `restrictions` is now used for namespaced restrictions and restriction exceptions (e.g. `restriction:motorcar=` as well as `except=motorcar`)
       - replaced lhs/rhs profiles by using test defined profiles
       - Handle `oneway=alternating` (routed over with penalty) separately from `oneway=reversible` (not routed over due to time dependence)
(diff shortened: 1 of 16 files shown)


MoKob:
Related: https://github.com/Project-OSRM/osrm-backend/pull/2822

TheMarex:
Just to put this on your radar: I had to switch to a more structured approach in #2399 that passes an object to the turn function similar to the `way_function` and `node_function` https://github.com/Project-OSRM/osrm-backend/pull/2399/files#diff-e262c8e753c419376683e5d8b10ba793R13 Depending on when you plan to implement this you could just piggy back on this and add the additional parameters to that struct.

MoKob:
Outdated. In general I'd still think we could improve the turn function by considering many of the tested parameters.
Especially if you are crossing through oncoming traffic would be nice to have (but also depends on traffic light settings).

PR #2915, opened 2016-09-15
Fixes sign mismatch in Nearest limit comparison

daniel-j-h:
Compiling on the demo server just gave me an unsigned <-> signed comparison warning.

Turns out we're missing a cast here:

`number_of_results` is unsigned, `max_results` is signed.

Because unsigned wins by the standard `max_results` is converted to unsigned first and only then compared. That's why e.g. `2u > -1` is false.

### DIFF
--- a/src/engine/plugins/nearest.cpp
+++ b/src/engine/plugins/nearest.cpp
@@ -8,6 +8,7 @@
 #include <string>
 
 #include <boost/assert.hpp>
+#include <boost/numeric/conversion/cast.hpp>
 
 namespace osrm
 {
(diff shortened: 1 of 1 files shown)


src/engine/plugins/nearest.cpp
```diff
@@ -26,7 +26,7 @@ Status NearestPlugin::HandleRequest(const api::NearestParameters &params,
 {
     BOOST_ASSERT(params.IsValid());
 
-    if (params.number_of_results > max_results)
+    if (static_cast<std::int64_t>(params.number_of_results) > max_results)
```
MoKob: Just for the fun of it, shouldn't we use a `boost::numeric_cast` for these kinds of checks (even though safe right now) to prevent errors when the casted type changes?

daniel-j-h: In general: yes.

daniel-j-h: Fixed.

We also should check other casts, ticketing over at https://github.com/Project-OSRM/osrm-backend/issues/2916

Issue #2918, opened 2016-09-15
Investigate Turn Function


MoKob:
Our new turn function is providing very low cost for turns. Measured in seconds, we currently see the following turn cost per angle:

```
[0,11] 7.4
[12,22] 7.3
[23,28] 7.2
[29,33] 7.1
[34,36] 7.0
[37,39] 6.9
[40,42] 6.8
[43,44] 6.7
[45,46] 6.6
[47,48] 6.5
[49,49] 6.4
[50,51] 6.3
[52,52] 6.2
[53,54] 6.1
[55,55] 6.0
[56,56] 5.9
[57,57] 5.8
[58,58] 5.7
[59,59] 5.6
[60,60] 5.5
[61,61] 5.4
[62,62] 5.3
[63,63] 5.2
[64,64] 5.1
[65,65] 5.0
[66,66] 4.9
[67,67] 4.8
[68,68] 4.7
[69,69] 4.6
[70,70] 4.4
[71,71] 4.3
[72,72] 4.2
[73,73] 4.1
[74,74] 4.0
[75,75] 3.8
[76,76] 3.7
[77,77] 3.6
[78,78] 3.4
[79,79] 3.3
[80,80] 3.2
[81,81] 3.1
[82,82] 3.0
[83,83] 2.8
[84,84] 2.7
[85,85] 2.6
[86,86] 2.5
[87,87] 2.4
[88,88] 2.3
[89,89] 2.2
[90,90] 2.1
[91,91] 2.0
[92,92] 1.9
[93,93] 1.8
[94,94] 1.7
[95,95] 1.6
[96,96] 1.5
[97,97] 1.4
[98,99] 1.3
[100,100] 1.2
[101,102] 1.1
[103,103] 1.0
[104,105] .9
[106,107] .8
[108,109] .7
[110,112] .6
[113,115] .5
[116,118] .4
[119,123] .3
[124,129] .2
[130,140] .1
[141,202] .0
[203,211] .1
[212,216] .2
[217,220] .3
[221,223] .4
[224,226] .5
[227,228] .6
[229,230] .7
[231,232] .8
[233,233] .9
[234,235] 1.0
[236,236] 1.1
[237,237] 1.2
[238,238] 1.3
[239,240] 1.4
[241,241] 1.5
[242,242] 1.6
[243,243] 1.7
[244,244] 1.9
[245,245] 2.0
[246,246] 2.1
[247,247] 2.2
[248,248] 2.3
[...]

PR #2922, opened 2016-09-15
Offset turn angle into +/-180 range correctly

karenzshea:
# Issue

Fixes https://github.com/Project-OSRM/osrm-backend/issues/2918. Turn angles were being subtracted from 180 before being passed into the lua `turn_penalty` function, which is possibly how the the old `turn_penalty` function expected them (essentially flipped, as an angle of 0 became 180 and an angle of 180 became 0). The new function expects values between -180 and 180. 

h/t to @jakepruitt for help with debugging this

## Tasklist
 - [ ] add regression test
 - [ ] review
 - [ ] adjust for for comments

## Requirements / Relations
https://github.com/Project-OSRM/osrm-backend/milestone/16

### DIFF
--- a/src/extractor/edge_based_graph_factory.cpp
+++ b/src/extractor/edge_based_graph_factory.cpp
@@ -431,8 +431,9 @@ void EdgeBasedGraphFactory::GenerateEdgeExpandedEdges(
                     distance += profile_properties.traffic_signal_penalty;
                 }
 
+                const auto turn_angle = turn.angle > 180 ? 180 - turn.angle : turn.angle;
                 const int32_t turn_penalty =
-                    scripting_environment.GetTurnPenalty(180. - turn.angle);
+                    scripting_environment.GetTurnPenalty(turn_angle);
                 const auto turn_instruction = turn.instruction;
 
                 if (turn_instruction.direction_modifier == guidance::DirectionModifier::UTurn)


karenzshea:
For the record, the graphed output of the new turn function looks more or less as expected; mirrored increase of penalties for sharper angles, slightly more expensive for left turns.

screenshot from 2016-09-15 17-05-22

src/extractor/edge_based_graph_factory.cpp
```diff
@@ -431,8 +431,9 @@ void EdgeBasedGraphFactory::GenerateEdgeExpandedEdges(
                     distance += profile_properties.traffic_signal_penalty;
                 }
 
+                const auto turn_angle = turn.angle > 180 ? 180 - turn.angle : turn.angle;
                 const int32_t turn_penalty =
-                    scripting_environment.GetTurnPenalty(180. - turn.angle);
+                    scripting_environment.GetTurnPenalty(turn_angle);
```
MoKob: might I ask where this is coming from? As you can see in https://github.com/Project-OSRM/osrm-backend/issues/2918, I have made tests using exactly `180-turn.angle` and you get just the results we want. 

From my tests an angle of 180 (resulting in 0 as call argument) is getting a turn penalty of zero.
From your new call value, we would get 7.4 seconds for going straight.
Osrm assumes 180 as angle for going straight. The turn function in the profile expects 0 for going straight.
The angle change is broken here.

MoKob:
The obvious check here: https://travis-ci.org/Project-OSRM/osrm-backend/jobs/160304407#L3993 where you can see that https://github.com/Project-OSRM/osrm-backend/blob/master/features/bicycle/maxspeed.feature#L41-L48 is getting penalised now. Going straight didn't experience a penalty before, now it does.

MoKob:
As described in https://github.com/Project-OSRM/osrm-backend/issues/2918#issue-177214673, we probably need a higher based penalty and probably additional penalties.

The turn function itself is correct though. OSRM simply has a different internal model than the external model to represent turn angles. A turn-angle in OSRM is modelled from `0` (u-turn) to `180` (going straight). Having the minimal penalty at `180`, not at `0` due to `180-angle` is wanted behaviour.

jakepruitt:
> Having the minimal penalty at 180, not at 0 due to 180-angle is wanted behaviour.

Just to clarify, is this 180 the number passed into the lua function? or the angle in the edge-based-graph-factory?

Issue #2927, opened 2016-09-16
osrm-extract hangs on generating edge-expanded edges


rkcpi:
We're currently trying to use osrm-extract on the following map: http://download.geofabrik.net/europe-latest.osm.pbf.

However, osrm-extract seems to stop on the `generating edge-expanded edges` step:
```
...
[extractor] writing street name index ... ok, after 0.537011s
[info] Writing turn lane masks...
[info] done (0.000127)
[info] extraction finished after 4976.04s
[info] Generating edge-expanded graph representation
[info]  - 341223 restrictions.
[info] Importing n = 382120223 nodes 
[info]  - 1094928 bollard nodes, 294824 traffic lights
[info]  and 402322465 edges 
[info] Graph loaded ok and has 402322465 edges
. 10% . 20% . 30% . 40% . 50% . 60% . 70% . 80% . 90% . 100%
[info] Node compression ratio: 0.17176
[info] Edge compression ratio: 0.213348
. 10% . 20% . 30% . 40% . 50% . 60% . 70% . 80% . 90% . 100%
[info] Generated 402321631 nodes in edge-expanded graph
[info] generating edge-expanded edges
```
It's been like this for several hours now, usually it completes within one to two hours. It also doesn't seem to do much: load avergae is 1.00, only one CPU seems to be used.

This happens with a current master build (commit aa5f8dce23fc178a3dd8e45a9d79393d29d93986), stxxl v1.3.1 on a CentOS 7.2.1511 box.

This was fine with the build we used before this one (master d3a6b5a77ed4ec9663f68aea24f960ecfb70619a)

Might be related to #2852?

daniel-j-h:
Thanks for reporting.

Looks related to turn analysis being stuck in a loop. @MoKob can you have a look on Monday?

@rkcpi what you already can do is attach gdb to the running process and manually do some sampling: break and note the function you're in. I guess this works best with debug symbols:

    cmake .. -DCMAKE_BUILD_TYPE=Debug

This way you should be able to localize where the loop is happening exactly.

MoKob:
I will look into it. Just on a side-note: the "[info] generating edge-expanded edges" is sequential, so having only one CPU at work is expected.

MoKob:
We just merged a fix that should, hopefully solve the hang you are expecting. Can you report back and close here, if the newest master solves your problem?

pongo710:
Thanks for the fast fix. We will test and report if this fixes the issue for us.

[...]

PR #2931, opened 2016-09-19
Prevent loops in extraction based on merge

MoKob:
# Issue

https://github.com/Project-OSRM/osrm-backend/issues/2927. Turns out that we can end up in loops of degree two due to merging of segregated roads. This requires a strange set of modelled roads, but we can't  assure it will not be in the data.

This PR implements a fallback that assures we cannot loop over traffic-lights endlessly.

## Tasklist
 - [ ] review
 - [ ] adjust for for comments

### DIFF
--- a/features/guidance/dedicated-turn-roads.feature
+++ b/features/guidance/dedicated-turn-roads.feature
@@ -313,3 +313,39 @@ Feature: Slipways and Dedicated Turn Lanes
             | a,d       | new york,new york                       | depart,arrive                      | this is the sinatra route            |
             | a,j       | new york,1st street,1st street          | depart,turn left,arrive            |                                      |
             | a,1       | new york,m street,1st street,1st street | depart,turn right,turn left,arrive | this can false be seen as a sliproad |
+
+    # Merging into degree two loop on dedicated turn detection / 2927
+    Scenario: Turn Instead of Ramp
+        Given the node map
+            |   |   |   |   |   |   |   |   |   |   |   |   |   |   |   | f |
+            |   |   |   |   | g |   |   |   |   |   | h |   |   |   |   |   |
+            |   |   |   |   |   |   |   |   |   |   |   |   | d |   |   | e |
+            | i |   |   |   | c |   |   |   |   |   | j |   |   |   |   |   |
+            |   |   |   |   |   |   |   |   |   |   |   |   |   |   |   |   |
+            |   |   |   |   |   |   |   |   |   |   |   |   |   |   |   |   |
+            |   |   |   |   |   |   |   |   |   |   |   |   |   |   |   |   |
+            |   |   |   |   |   |   |   |   |   |   |   |   |   |   |   |   |
+            |   |   |   |   |   |   |   |   |   |   |   |   |   |   |   |   |
+            |   |   |   |   |   |   |   |   |   |   |   |   |   |   |   |   |
+            |   |   |   |   |   |   |   |   |   |   |   |   |   |   |   |   |
+            |   |   |   |   |   |   |   |   |   |   |   |   |   |   |   |   |
+            |   |   | b |   |   |   |   |   |   |   |   |   |   |   |   |   |
+            |   |   |   |   |   |   |   |   |   |   |   |   |   |   |   |   |
+            |   |   | a |   |   |   |   |   |   |   |   |   |   |   |   |   |
+
+        And the ways
+            | nodes | highway | name | oneway |
+            | abi   | primary | road | yes    |
+            | bcjd  | primary | loop | yes    |
+            | dhgf  | primary | loop | yes    |
+            | fed   | primary | loop | yes    |
+
+        And the nodes
+            | node | highway         |
+            | g    | traffic_signals |
+            | c    | traffic_signals |
+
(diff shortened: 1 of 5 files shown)


features/guidance/dedicated-turn-roads.feature
```diff
@@ -313,3 +313,39 @@ Feature: Slipways and Dedicated Turn Lanes
...
+            | nodes | highway | name | oneway |
+            | abi   | primary | road | yes    |
+            | bcjd  | primary | loop | yes    |
+            | dhgf  | primary | loop | yes    |
+            | fed   | primary | loop | yes    |
+
+        And the nodes
+            | node | highway         |
+            | g    | traffic_signals |
+            | c    | traffic_signals |
```
daniel-j-h: Are these required to trigger the endless looping?

features/guidance/roundabout.feature
```diff
@@ -36,6 +36,25 @@ Feature: Basic Roundabout
...
+    #2927
+    Scenario: Only Roundabout
+        Given the node map
+            |   |   | a |   |   |
+            |   |   |   |   |   |
+            |   |   |   |   |   |
+            | b |   |   |   | d |
+            |   |   |   |   |   |
+            |   |   |   |   |   |
+            |   |   | c |   |   |
```
daniel-j-h: Is this related to the issue? I thought the loop was based on merges?

src/extractor/guidance/intersection_generator.cpp
```diff
@@ -577,9 +580,13 @@ IntersectionGenerator::GetActualNextIntersection(const NodeID starting_node,
...
-    while (result.size() == 2 &&
-           node_based_graph.GetEdgeData(via_edge).IsCompatibleTo(
-               node_based_graph.GetEdgeData(result[1].turn.eid)))
+    // using a maximum lookahead, we make sure not to end up in some form of loop
+    const constexpr std::size_t max_lookahead = 10;
+    for (std::size_t lookahead = 0;
+         lookahead < max_lookahead && (result.size() == 2 &&
+                                       node_based_graph.GetEdgeData(via_edge).IsCompatibleTo(
+                                           node_based_graph.GetEdgeData(result[1].turn.eid)));
+         ++lookahead)
```
daniel-j-h: For my own understanding: you try to find the first compatible edge but now with a limit so that in edge cases (as the one triggering this) we only end up looking at a limited amount (10 here) to prevent loops. Why the 10 here instead of e.g. storing and checking for a actual loop?

src/extractor/guidance/sliproad_handler.cpp
```diff
@@ -51,27 +51,36 @@ operator()(const NodeID, const EdgeID source_edge_id, Intersection intersection)
...
-        [&](const NodeID at_node, const ConnectedRoad &road, NodeID &output_node) {
-            auto intersection = intersection_generator(at_node, road.turn.eid);
-            auto in_edge = road.turn.eid;
-            // skip over traffic lights
-            while (intersection.size() == 2)
+    const auto findNextIntersectionForRoad = [&](
+        const NodeID at_node, const ConnectedRoad &road, NodeID &output_node) {
+        auto intersection = intersection_generator(at_node, road.turn.eid);
+        auto in_edge = road.turn.eid;
+        // skip over traffic lights
```
daniel-j-h: Ah this is the reason you had traffic lights in the test case at the top?

src/extractor/guidance/sliproad_handler.cpp
```diff
@@ -51,27 +51,36 @@ operator()(const NodeID, const EdgeID source_edge_id, Intersection intersection)
...
+        // if there are more than 10 traffic lights/other items on the road, we don't consider it
+        // for sliproads anymore. This also prevents us from ending up in an endless loop.
```
daniel-j-h: This is primarily for endless loops, right? We don't really care for number of traffic lights on sliproads?

features/guidance/dedicated-turn-roads.feature
MoKob: Yes, as you found out already.

features/guidance/roundabout.feature
MoKob: This is simply an additional case where loops could theoretically happen (roundabout handler has loop checking as well). This is just additional drive-by security

src/extractor/guidance/intersection_generator.cpp
MoKob: Well, we could remember all nodes we see adding an additional data structure (e.g. hashmap) and check whether any node is repeated. Since we usually have a single traffic light to look at, this check is a simple way around broken modelings. At the same time, we don't have to allocate a dedicated `unordered_set` for every iteration.

If you feel that we should be on the save side for even situations with more than 10 traffic-lights/reasons not to compress on a single edge, we can go for an unordered_map, though.

src/extractor/guidance/sliproad_handler.cpp
MoKob: No. We simply want to prevent endless loops. The same argument as above holds here. If we want to detect endless loops here, we would need an `unordered_set` to detect these. There is simply no situation that justifies a detection as sliproad if there is like 10 or more traffic lights in between. Anything but `1` would be unexpected.

PR #2934, opened 2016-09-20
Revert #2795: Rewrite cucumber test caching, Tests Failing At Random

daniel-j-h:
The Cucumber caching PR @springmeyer merged into master yesterday (ref. https://github.com/Project-OSRM/osrm-backend/pull/2795) does not work. Invoking `npm test` multiple times (so we hit the cache) lets tests fail at random and only sometimes.

Here is an example (specific tests are not important, write after end error is):

https://gist.github.com/daniel-j-h/d73b4fcea574e61f1815505f1aae44be

Reverting this commit to get a working master again.


features/guidance/turn-lanes.feature
```diff
@@ -620,7 +620,7 @@ Feature: Turn Lane Guidance
             | a,d       | hwy,hwy       | depart,arrive                       | ,                                                                            |
             | a,e       | hwy,ramp,ramp | depart,off ramp slight right,arrive | ,straight:false straight:false straight;slight right:true slight right:true, |
 
-    @todo
+    @bug @todo
```
MoKob: seems like a few additional `bug` tags got back in here?

features/guidance/anticipate-lanes.feature
```diff
@@ -422,7 +422,7 @@ Feature: Turn Lane Guidance
                | waypoints | route               | turns                                          | lanes                                                                                                                                                             |
                | a,e       | main,main,main,main | depart,use lane straight,continue right,arrive | ,left:false straight:false straight:false straight:false straight:true straight:true right:false,straight:false straight:false right:false right:true right:true, |
 
-    @anticipate @todo @2661
+    @anticipate @todo @bug @2661
```
MoKob: Also additional bug tag

features/testbot/bugs.feature
```diff
@@ -0,0 +1,5 @@
+@routing @testbot @bug
+Feature: Known bugs
+
+    Background:
+        Given the profile "testbot"
```
MoKob: is that feature designated to do something? Right now it seems completely empty. If we want to keep it, we should move known bugs here

features/testbot/via.feature
```diff
@@ -173,7 +173,7 @@ Feature: Via points
             | c,d,a     | abc,bd,bd,bd,abc,abc |
 
     # See issue #2349
-    @todo
+    @bug @todo
```
MoKob: additional label

springmeyer:
> https://gist.github.com/daniel-j-h/d73b4fcea574e61f1815505f1aae44be

First step I tried replicating this on OS X, but can't. Next I will try linux...

springmeyer:
I'm still unable to replicate on my linux machine. Sounds like @oxidase has been able to replicate on linux only with a release build and 500ms timeout (refs https://github.com/Project-OSRM/osrm-backend/pull/2936). So, I think this revert should be paused and the focus instead put on testing the fix at https://github.com/Project-OSRM/osrm-backend/pull/2936. (/cc @daniel-j-h can you test?)

MoKob:
We are going with https://github.com/Project-OSRM/osrm-backend/pull/2936

PR #2936, opened 2016-09-20
Fix "Error: write after end" and removed @bug tags

oxidase:
# Issue

Fix the problem that is caused by too early close of a log stream at process exit, but https://nodejs.org/api/child_process.html#child_process_event_exit
```
Note that when the 'exit' event is triggered, child process stdio streams might still be open.
```

@daniel-j-h  please could you check the change of `log.end`  to `log.write`  locally. I could reproduce it only with a release build and timeout 500ms.
/cc @springmeyer 

## Tasklist
 - [x] review
 - [x] adjust for for comments

## Requirements / Relations
Supersedes #2934, if retest will be  positive. Related to #2795


springmeyer:
I've been unable to reproduce with either Debug or Release build and with either OS X or Linux. Also tried `export CUCUMBER_TIMEOUT=500 && npm test`. So, I think the idea next step here is to have @daniel-j-h - who appears to be able to reproduce easily - test this branch and merge if it solves his issue.

PR #2941, opened 2016-09-22
Prevents typedefs.hpp from polluting the user's namespace, resolves #2940

### LINKED ISSUE #2940: libosrm: global namespace pollution
In https://github.com/Project-OSRM/osrm-backend/issues/2938#issuecomment-248829107 a user reported name clashes when trying to use `libosrm`.

Turns out our [typedefs.hpp](https://github.com/Project-OSRM/osrm-backend/blob/8aa93f32ccd02ed3cdde2429715cdc9366be87f3/include/util/typedefs.hpp) header is not namespaced, but we [install it](https://github.com/Project-OSRM/osrm-backend/blob/8aa93f32ccd02ed3cdde2429715cdc9366be87f3/CMakeLists.txt#L442) to the user's include dir.

This not only pollutes the user's global namespace but also is prone to name clashes as reported in the ticket linked above when trying to use / embed `libosrm`.

cc @MoKob I'm tagging this 6.0 already since it's a breaking `libosrm` API change.

daniel-j-h:
For #2940: `typefefs.hpp` defined some types and type-aliases (such as node / way
id for example) but did so without a namespace, polluting the global
one. This can cause name clashes when using / embedding libosrm.

This changeset namespaces types and type aliases in `typedefs.hpp` into
the `osrm` namespace and adapts some callsites.

Because the `libosrm` headers we ship to clients depend on
`typedefs.hpp` and clients could already use it this change is breaking
the `libosrm` v5 API.
## Tasklist
- [x] review
- [x] adjust for for comments

### DIFF
--- a/include/util/typedefs.hpp
+++ b/include/util/typedefs.hpp
@@ -36,12 +36,13 @@ SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 #include <cstdint>
 #include <limits>
 
+namespace osrm
+{
+
 // OpenStreetMap node ids are higher than 2^32
 OSRM_STRONG_TYPEDEF(std::uint64_t, OSMNodeID)
-OSRM_STRONG_TYPEDEF_HASHABLE(std::uint64_t, OSMNodeID)
 
 OSRM_STRONG_TYPEDEF(std::uint32_t, OSMWayID)
-OSRM_STRONG_TYPEDEF_HASHABLE(std::uint32_t, OSMWayID)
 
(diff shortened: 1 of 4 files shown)


TheMarex:
Arg this is yet again a problem cause by exposing the raw `Hint` object (containing a `PhantomNode` and hence pulls in a lot of internal dependencies).

While `typedefs.hpp` definetly should be namespaced (and should die) we should fix the remaining issues https://github.com/Project-OSRM/osrm-backend/issues/2031#issuecomment-214775013 at the same time.

TheMarex:
Outdated.

PR #2942, opened 2016-09-22
prefer first result

MoKob:
# Issue

Not really an issue. This PR simply prefers the first lexicographical result instead of switching the results in our brute-force method. In addition, we can now query the durations of different trips in the cucumber tests.

## Tasklist
 - [x] review
 - [x] adjust for for comments

### DIFF
--- a/features/step_definitions/trip.js
+++ b/features/step_definitions/trip.js
@@ -1,6 +1,10 @@
 var util = require('util');
 
 module.exports = function () {
+    function add(a, b) {
+        return a + b;
+    }
+
     this.When(/^I plan a trip I should get$/, (table, callback) => {
         var got;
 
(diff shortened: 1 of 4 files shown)


features/step_definitions/trip.js
```diff
@@ -52,6 +53,16 @@ module.exports = function () {
                                 return toAdd;
                             })));
                         }
+                        if(headers.has('durations')) {
+                            var all_durations = json.trips.filter(t => !!t).map(t => t.legs).map(tl => Array.prototype.concat.apply([], tl.map((sl, i) => {
```
daniel-j-h: `!!t` is some Js double negation to filter for existing trips?

daniel-j-h:
:+1:

daniel-j-h:
Just noticed that Travis is failing because of some minor eslint issues. Otherwise good to go.

Issue #2909, opened 2016-09-13
Check Semantics of EMPTY_NAMEID Callsites


daniel-j-h:
@MoKob capturing from today's chat so we don't forget about this (feel free to add more or clarify):

We're using `EMPTY_NAMEID` in guidance; now with the ref / name split we have to make sure the semantic behind checking the id is still correct or if we rather have to check e.g. for empty name or empty ref.

TheMarex:
Maybe the name `EMPTY_STRING` would be more fitting, as the dictionary contains more then just names at this point.

MoKob:
@daniel-j-h with your name-change heuristic merged, is this resolved?

daniel-j-h:
Yes - this landed. Closing here.

PR #2945, opened 2016-09-22
Uses Name Change Heuristic Instead of Comparing Name Ids

daniel-j-h:
Uses Name Change Heuristic Instead of Comparing Name Ids

@MoKob what's your take on replacing all of these, mostly in the `IntersectionHandler` and its derived handlers to make use of the name change heuristic? Work in progress, just started out looking through the code and seeing where and how we're using raw name ids.
## Tasklist
- [x] adapt all handlers derived from `IntersectionHandler`
- [x] adapt cucumber tests
- [x] review
- [x] adjust for for comments

References:
- https://github.com/Project-OSRM/osrm-backend/issues/2857
- https://github.com/Project-OSRM/osrm-backend/pull/2900
- https://github.com/Project-OSRM/osrm-backend/issues/2909


MoKob:
Sounds like a good idea. The name-id changes were all prior to the `noticeable name change` heuristic. All name_id changes should be just fine to replace.

daniel-j-h:
Note: the single test breaking is unrelated and due to https://github.com/Project-OSRM/osrm-backend/issues/2965.

features/guidance/bridges_and_tunnels.feature
```diff
@@ -31,11 +31,11 @@ Feature: Car - Guidance - Bridges and Tunnels
...
-            | from | to | route                                      | turns                           |
-            | a    | d  | Hauptstraße,Nebenstraße,Nebenstraße        | depart,end of road left,arrive  |
-            | a    | e  | Hauptstraße,Nebenstraße,Nebenstraße        | depart,end of road right,arrive |
-            | e    | a  | Nebenstraße,Hauptstraßenbrücke,Hauptstraße | depart,turn left,arrive         |
-            | d    | a  | Nebenstraße,Hauptstraßenbrücke,Hauptstraße | depart,turn right,arrive        |
+            | from | to | route                                      | turns                    |
+            | a    | d  | Hauptstraße,Nebenstraße,Nebenstraße        | depart,turn left,arrive  |
+            | a    | e  | Hauptstraße,Nebenstraße,Nebenstraße        | depart,turn right,arrive |
+            | e    | a  | Nebenstraße,Hauptstraßenbrücke,Hauptstraße | depart,turn left,arrive  |
+            | d    | a  | Nebenstraße,Hauptstraßenbrücke,Hauptstraße | depart,turn right,arrive |
```
daniel-j-h: @MoKob what do you think about this change in behavior here? Strictly speaking it's still end of the road but with a differently named bridge at the end (?)

daniel-j-h: Btw similar changes below.

features/guidance/collapse.feature
```diff
@@ -376,8 +376,8 @@ Feature: Collapse
             | cd    | primary |        |
 
         When I route I should get
-            | waypoints | route | turns         |
-            | a,d       | ,     | depart,arrive |
+            | waypoints | route    | turns                           |
+            | a,d       | ,Bridge, | depart,new name straight,arrive |
```
daniel-j-h: Looks suspicious, too.

features/car/names.feature
```diff
@@ -91,5 +91,5 @@ Feature: Car - Street names in instructions
             | restriction | abc      | cde    | c        | no_right_turn |
 
        When I route I should get
-            | waypoints | route                | turns                    | pronunciations        |
-            | a,e       | road,trunk,trunk     | depart,turn right,arrive | roooaad,truank,truank |
+            | waypoints | route             | turns                                                | pronunciations         |
+            | a,e       | road,,trunk,trunk | depart,on ramp slight right,merge slight left,arrive | roooaad,,truank,truank |
```
MoKob: sliproad detection seems not to work here anymore. This adjustment does not seem wanted

features/guidance/bridges_and_tunnels.feature
```diff
@@ -31,11 +31,11 @@ Feature: Car - Guidance - Bridges and Tunnels
...
 
         When I route I should get
-            | from | to | route                                      | turns                           |
-            | a    | d  | Hauptstraße,Nebenstraße,Nebenstraße        | depart,end of road left,arrive  |
-            | a    | e  | Hauptstraße,Nebenstraße,Nebenstraße        | depart,end of road right,arrive |
-            | e    | a  | Nebenstraße,Hauptstraßenbrücke,Hauptstraße | depart,turn left,arrive         |
-            | d    | a  | Nebenstraße,Hauptstraßenbrücke,Hauptstraße | depart,turn right,arrive        |
+            | from | to | route                                      | turns                    |
+            | a    | d  | Hauptstraße,Nebenstraße,Nebenstraße        | depart,turn left,arrive  |
+            | a    | e  | Hauptstraße,Nebenstraße,Nebenstraße        | depart,turn right,arrive |
```
MoKob: Why is this not detected as end-of-road anymore. Fixture of testcase is not correct here

features/guidance/bridges_and_tunnels.feature
```diff
@@ -51,15 +51,15 @@ Feature: Car - Guidance - Bridges and Tunnels
...
-            | f    | e  | Anderestraße,Hauptstraße,Nebenstraße,Nebenstraße         | depart,turn left,end of road right,arrive  |
-            | g    | d  | Anderestraße,Hauptstraße,Nebenstraße,Nebenstraße         | depart,turn right,end of road left,arrive  |
-            | g    | e  | Anderestraße,Hauptstraße,Nebenstraße,Nebenstraße         | depart,turn right,end of road right,arrive |
-            | e    | f  | Nebenstraße,Hauptstraßenbrücke,Anderestraße,Anderestraße | depart,turn left,end of road right,arrive  |
-            | e    | g  | Nebenstraße,Hauptstraßenbrücke,Anderestraße,Anderestraße | depart,turn left,end of road left,arrive   |
-            | d    | f  | Nebenstraße,Hauptstraßenbrücke,Anderestraße,Anderestraße | depart,turn right,end of road right,arrive |
-            | d    | g  | Nebenstraße,Hauptstraßenbrücke,Anderestraße,Anderestraße | depart,turn right,end of road left,arrive  |
+            | from | to | route                                                    | turns                               |
+            | f    | d  | Anderestraße,Hauptstraße,Nebenstraße,Nebenstraße         | depart,turn left,turn left,arrive   |
```
MoKob: see above. should still be end-of-road

daniel-j-h:
@MoKob it looks like the motorway and sliproad handler introduce the issues we're seeing with the tests.

The changes preserve the semantics, though. Have a look at the diff there:

    git show src/extractor/guidance/{motorway,sliproad}_handler.cpp

Could you lend me some eyes here?

src/extractor/guidance/intersection_handler.cpp
```diff
@@ -60,7 +59,14 @@ TurnType::Enum IntersectionHandler::findBasicTurnType(const EdgeID via_edge,
...
 
-    if (in_data.name_id == out_data.name_id && in_data.name_id != EMPTY_NAMEID)
+    const auto same_name =
+        !util::guidance::requiresNameAnnounced(name_table.GetNameForID(in_data.name_id),
+                                               name_table.GetRefForID(in_data.name_id),
+                                               name_table.GetNameForID(out_data.name_id),
+                                               name_table.GetRefForID(out_data.name_id),
+                                               street_name_suffix_table);
+
+    if (in_data.name_id != EMPTY_NAMEID && same_name)
```
MoKob: the above check (by transitivity) checked for `in_data.name_id == out_data.name_id && in_data.name_id != EMPTY_NAME_ID && out_data.name_id != EMPTY_NAME_ID`. The final check of `out_data.name_id != EMPTY_NAME_ID` is missing now.

src/extractor/guidance/motorway_handler.cpp
```diff
@@ -120,8 +120,16 @@ Intersection MotorwayHandler::fromMotorway(const EdgeID via_eid, Intersection in
         for (const auto &road : intersection)
         {
             const auto &out_data = node_based_graph.GetEdgeData(road.turn.eid);
-            if (road.turn.angle != 0 && in_data.name_id == out_data.name_id &&
-                in_data.name_id != EMPTY_NAMEID && isMotorwayClass(road.turn.eid, node_based_graph))
+
+            const auto same_name =
```
MoKob: see my comment above. The `out_data.name_id` cannot be empty here to preserve the handling based on transitivity.

src/extractor/guidance/motorway_handler.cpp
```diff
@@ -395,10 +412,7 @@ Intersection MotorwayHandler::fromRamp(const EdgeID via_eid, Intersection inters
             {
                 BOOST_ASSERT(intersection[2].entry_allowed);
                 if (isMotorwayClass(intersection[2].turn.eid, node_based_graph) &&
-                    node_based_graph.GetEdgeData(intersection[1].turn.eid).name_id !=
-                        EMPTY_NAMEID &&
-                    node_based_graph.GetEdgeData(intersection[2].turn.eid).name_id ==
-                        node_based_graph.GetEdgeData(intersection[1].turn.eid).name_id)
+                    first_intersection_data.name_id != EMPTY_NAMEID && first_second_same_name)
```
MoKob: this check also needs to check if any of the name_ids is empty.

src/extractor/guidance/motorway_handler.cpp
```diff
@@ -367,11 +386,9 @@ Intersection MotorwayHandler::fromRamp(const EdgeID via_eid, Intersection inters
             //          0
             if (intersection[1].entry_allowed)
             {
+
                 if (isMotorwayClass(intersection[1].turn.eid, node_based_graph) &&
-                    node_based_graph.GetEdgeData(intersection[2].turn.eid).name_id !=
-                        EMPTY_NAMEID &&
-                    node_based_graph.GetEdgeData(intersection[2].turn.eid).name_id ==
-                        node_based_graph.GetEdgeData(intersection[1].turn.eid).name_id)
+                    second_intersection_data.name_id != EMPTY_NAMEID && first_second_same_name)
```
MoKob: here as well.

daniel-j-h:
Rebased onto master; worked on the empty name id issues and failing tests are down to five now.

src/extractor/guidance/intersection_generator.cpp
```diff
@@ -194,7 +200,12 @@ bool IntersectionGenerator::CanMerge(const NodeID node_at_intersection,
         return false;
 
     // need to be same name
-    if (first_data.name_id != second_data.name_id)
+    if (second_data.name_id != EMPTY_NAMEID &&
```
MoKob: never mind. above check implies it.

src/extractor/guidance/sliproad_handler.cpp
```diff
@@ -212,8 +220,18 @@ operator()(const NodeID, const EdgeID source_edge_id, Intersection intersection)
                     std::find_if(target_intersection.begin() + 1,
                                  target_intersection.end(),
                                  [this, &link_data](const ConnectedRoad &road) {
-                                     return node_based_graph.GetEdgeData(road.turn.eid).name_id ==
-                                            link_data.name_id;
+
+                                     const auto &road_edge_data =
+                                         node_based_graph.GetEdgeData(road.turn.eid);
+
+                                     const auto same_name = !util::guidance::requiresNameAnnounced(
```
MoKob: from the former checks, this check should only succeed on non-empty nameid values as well

src/extractor/guidance/sliproad_handler.cpp
```diff
@@ -253,7 +271,15 @@ operator()(const NodeID, const EdgeID source_edge_id, Intersection intersection)
     if (next_road.turn.instruction.type == TurnType::Fork)
     {
         const auto &next_data = node_based_graph.GetEdgeData(next_road.turn.eid);
-        if (next_data.name_id == source_edge_data.name_id)
+
+        const auto same_name = !util::guidance::requiresNameAnnounced(
```
MoKob: not in tests, but these cannot be empty as well

features/guidance/bridges_and_tunnels.feature
MoKob: after investigation: the Bridge in-between is not considered a intersection anymore (essentially same name, no other road). Therefore the end-of-road does not come out of the json-factory anymore. We only announce end of road if there is an actual intersection in between.

features/guidance/bridges_and_tunnels.feature
MoKob: see my comment above

daniel-j-h:
- [ ] needs to take pronunciation into account

daniel-j-h:
(painfully) rebased and got failing tests down to a single scenario - would love to get this finally in as soon as I figured out the last one.

daniel-j-h:
@MoKob this is where I left off:
- the IsUTurn check now works with the name change heuristic
- the test is still failing, making a slight left instead of a uturn
- collapseUTurn is getting called with step indices 0, 1, 2 but does not reach the forwardStepSignage branch where we expected it to remove the ref

Here's the diff

```
| (-) a,h | (-) mich,mich,mich | (-) b9,b9,b9 | (-) depart,continue uturn,arrive |
| (+) a,h | (+) mich,mich,mich | (+) b9,,b9 | (+) depart,continue slight left,arrive |
```

PR #2946, opened 2016-09-24
Unset FORTIFY_SOURCE before setting to avoid warnings for compilers w…

danpat:
Some distributions (e.g. Alpine Linux) have GCC configured to pre-define `_FORTIFY_SOURCE`.  In these cases, compiling with our current configuration leads to lots of annoying messages like:

```
<command-line>:0:0: error: "_FORTIFY_SOURCE" redefined [-Werror]
```

Under some flag configurations (i.e. `-Werror`) this can even cause compilation to fail.

This change un-defines `_FORTIFY_SOURCE` before we re-define it, which avoids the printing of these warnings.  Because we explicitly want `_FORTIFY_SOURCE=2`, there is no harm in un-defining whatever was preset.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -193,7 +193,7 @@ endif()
 
 # Configuring compilers
 if(${CMAKE_CXX_COMPILER_ID} STREQUAL "Clang")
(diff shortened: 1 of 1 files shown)


MoKob:
@daniel-j-h you are our compiler master, any objections here?

daniel-j-h:
Looks hacky but reasonable from the explanation. Ready to merge.

PR #2953, opened 2016-09-26
V5.4.0 rc.5 lua revert

springmeyer:
Testing revert of 1309dd2a0f5b for v5.4.0-rc branch


springmeyer:
Testing this branch (and the revert of #2647) did not reduce memory usage as the problem was elsewhere. Closing this now without merging.

PR #2954, opened 2016-09-26
Added profile files per test generation

oxidase:
# Issue

Testing #2399 requires profiles that slightly different than car.lua or testbot.lua. Such dummy test profiles, like lhs.lua and rhs.lua, are useful only in tests and should not be committed in profiles directory. 

## Tasklist
 - [x] review
 - [x] adjust for for comments

## Requirements / Relations
Required for testing #2399 with different functions for turn/edge durations and  weights.


daniel-j-h:
Neat!

I think this is a great way of (accidentally? :P) solving https://github.com/Project-OSRM/osrm-backend/pull/2834 on the side and therefore superseding @karenzshea's work to get get options in via environment variable in https://github.com/Project-OSRM/osrm-backend/pull/2834.

Am I right in the assumption that this would allow us to say

```lua
require 'car'
ignore_hov_ways = false
```

writing Cucumber scenarios for which we can locally overwrite such flags?

See https://github.com/Project-OSRM/osrm-backend/blob/8aa93f32ccd02ed3cdde2429715cdc9366be87f3/profiles/car.lua#L160-L161

oxidase:
@daniel-j-h unfortunately for #2834 it will not work without dropping `local` for `ignore_toll_ways` in car.lua as it works for `properties` that is in the global scope. If `local` is removed 
```
     Scenario: Car - ignore_toll_ways=true enables toll way ignoring
        Given the profile file
        """
        require 'car'
        ignore_toll_ways = true
        """
        Then routability should be
            | primary | no          | x     |
````
will work.

**Comment by oxidase** [comment]
@daniel-j-h just updated PR for #2834:
* tests must be non-intrusive
* dropping `local` is a bad idea wrt to performance
Now a profile is extended with the given footer and loaded as a string.
For #2834 test will be:
```
     Scenario: Car - ignore_toll_ways=true enables toll way ignoring
        Given the profile file "car" extended with
        """
        ignore_toll_ways = true
        """
        Then routability should be
            | primary | no          | x     |
```

**Comment by emiltin** [comment]
a detail, but what about indenting the doc string to improve readability?


```gherkin
     Scenario: Car - ignore_toll_ways=true enables toll way ignoring
        Given the profile file "car" extended with
            """
            ignore_toll_ways = true
            """
        Then routability should be
            | primary | no          | x     |
```

daniel-j-h:
Thanks Michael, this is a great addition to the Cucumber code!

Issue #2878, opened 2016-09-08
Prefer Notification over NewName


MoKob:
When entering a ferry at a turn, we currently emit a `NewName`. This is confusing since the travel-mode change should be the obvious choice.

This change has to happen [here](https://github.com/Project-OSRM/osrm-backend/blob/938dff011fb52fedb8b787d7403b0c8fb1d67f5a/src/extractor/guidance/intersection_handler.cpp#L70-L164). We need to check the travel-mode changes to prefer them, when possible.

PR #2958, opened 2016-09-27
Return `notification` turn type when turn includes mode change

karenzshea:
# Issue

https://github.com/Project-OSRM/osrm-backend/issues/2878
We want to prioritize returning the `notification` turn type over `newName` when a maneuver includes a mode change.

## Tasklist
 - [x] add more cucumber cases (see docs/testing.md)
 - [ ] review
 - [ ] adjust for for comments

## Requirements / Relations
None that I'm aware of. @MoKob does this impact anything else?

### DIFF
--- a/features/guidance/notification.feature
+++ b/features/guidance/notification.feature
@@ -1,0 +1,57 @@
+@routing @guidance @mode-change
+Feature: Notification on turn onto mode change
+
+    Background:
+        Given the profile "car"
+        Given a grid size of 400 meters
+
+    Scenario: Turn onto a Ferry
+        Given the node map
+            | f |   |   |   |   |
+            | b |   |   | d |   |
+            | a |   |   |   | e |
+
+        And the ways
+            | nodes | highway | route | name  |
+            | abf   | primary |       |       |
+            | bd    |         | ferry | ferry |
+            | de    | primary |       |       |
+
+        When I route I should get
+            | waypoints | route     | turns                                        | modes                          |
+            | a,e       | ,ferry,,  | depart,turn right,notification right,arrive  | driving,ferry,driving,driving  |
+
+    Scenario: Turn onto a Ferry
+        Given the node map
+            | h |   |   | g |
+            | a | c |   | e |
+            | b |   |   | f |
+
+        And the ways
+            | nodes | highway | route | name  |
+            | ac    | primary |       |       |
+            | bah   | primary |       |       |
+            | ec    |         | ferry | ferry |
+            | gef   | primary |       |       |
+
+        When I route I should get
+            | waypoints | route     | turns                                                      | modes                                   |
+            | g,h       | ,ferry,,, | depart,turn right,notification straight,turn right,arrive  | driving,ferry,driving,driving,driving   |
+            | b,g       | ,,ferry,, | depart,turn right,notification straight,turn left,arrive   | driving,driving,ferry,driving,driving   |
+
+    Scenario: Straight onto a Ferry
+        Given the node map
+            |   |   |   |   |   |
+            |   | c | d |   | i |
+            | a |   |   |   |   |
+            |   |   |   | f |   |
+
+        And the ways
+            | nodes | highway | route | name    |
+            | ac    | primary |       |         |
+            | dc    |         | ferry | ferry   |
+            | df    | primary |       |         |
+
+        When I route I should get
+            | waypoints | route     | turns                                                  | modes                           |
+            | a,f       | ,ferry,,  | depart,notification right,notification right,arrive    | driving,ferry,driving,driving   |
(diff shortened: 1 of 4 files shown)


features/guidance/notification.feature
```diff
@@ -0,0 +1,65 @@
...
+
+    Scenario: Turn onto a Ferry
+        Given the node map
+            |   |   |   |   |   | f |
+            | b | c |   |   | d | e |
+            |   |   |   |   |   |   |
+            | a |   |   |   |   | g |
+
+        And the ways
+            | nodes | highway | route | name  |
```
MoKob: apart from the ferry (where d is just a superfluous node), all roads are simply two-hop roads. Very short ones at that. If you don't want to have collapse interact here, we should at least use a grid distance 200 meters.

In addition, having names here for two coordinate roads only complicates reading the map, especially since the names don't have anything to do with the location. I find it very difficult to read these maps due to this, since it gives an additional information that is not actually useful.

features/guidance/notification.feature
```diff
@@ -0,0 +1,65 @@
+@routing  @guidance @mode-change
+Feature: Notification on turn onto mode change
+
+    Background:
+        Given the profile "car"
+        Given a grid size of 20 meters
+
+    Scenario: Turn onto a Ferry
+        Given the node map
+            |   |   |   |   |   | f |
```
MoKob: f is not used here, right? I guess since you didn't see any notification on the exit? But in that case it would be alright, since the turn is not obvious (left/right are more or less equal)

features/guidance/notification.feature
```diff
@@ -0,0 +1,65 @@
...
+
+        When I route I should get
+            | waypoints | route                   | turns                                                   | modes                               |
+            | a,g       | bingo,ferry,wingo,wingo | depart,notification straight,notification right,arrive  | driving,ferry,driving,driving       |
+
+    Scenario: Turn onto a Ferry
+        Given the node map
+            |   |   |   |   |   |   | g |
+            |   | a |   |   |   |   | e |
+            |   |   | c |   |   | d |   |
```
MoKob: ferry lines sharing coordinates are more or less evil to me. If this should actually be a reasonable modelling in OSM, we need to think about to handle them specially. But staying on a ferry should not emit an instruction.

features/guidance/notification.feature
```diff
@@ -0,0 +1,65 @@
...
+            |   |   |   |   |   |   |   |   |
+            | a | b | c |   |   | d |   | i |
+            |   |   |   |   |   |   |   |   |
+            |   |   |   |   |   |   | f |   |
+
+        And the ways
+            | nodes | highway | route | name    | oneway |
+            | ab    | primary |       | yonkers | yes    |
+            | bc    | primary |       | dingo   | yes    |
+            | ca    | primary |       | bingo   | yes    |
```
MoKob: again, having three roads of different names here is going to pose problems at some point. Would you assume this to be a realistic setting?

features/guidance/notification.feature
```diff
@@ -0,0 +1,65 @@
...
+            | dc    |         | ferry | ferry   | no     |
+            | df    | primary |       | maple   | yes    |
+            | fi    | primary |       | dolly   | yes    |
+
+        When I route I should get
+            | waypoints | route                            | turns                                                              | modes                                 |
+            | a,i       | yonkers,ferry,maple,dolly,dolly  | depart,turn straight,notification right,new name sharp left,arrive | driving,ferry,driving,driving,driving |
```
MoKob: so you are getting a turn-straight here. I don't know if this is being solved already by my changes in the turn-angles branch, but going straight here might be considered obvious, which should be a notification in this case.

src/extractor/guidance/intersection_handler.cpp
```diff
@@ -68,6 +68,17 @@ TurnType::Enum IntersectionHandler::findBasicTurnType(const EdgeID via_edge,
     return TurnType::Turn;
 }
 
+TurnInstruction IntersectionHandler::notificationOnModeChange(const TravelMode in_mode,
+                                                              const TravelMode out_mode,
+                                                              const double turnAngle,
```
MoKob: passing the turn angle in here does not really provide any benefit. The entire function is only a mimic of the basic `?` operator. I would prefer using `{in_mode == out_mode ? TurnType::Suppressed : TurnType::Notification, getTurnDirection(turnAngle)}` instead of having a dedicated function for it.

It's a single line, which can be grasped instantly, and on the other side a function that I have to search for first.

If you prefer to use the function, the function should return a TurnType and not care about the turn-angle.

So we would get `return {notificationOnModeChange(in_mode,out_mode,TurnType::Suppressed),getTurnDirection(turnAngle)}`.

Passing the turn-angle in adds complexity to the function using something that is not even considered except for the return.

PR #2963, opened 2016-09-28
Check explicitly for empty names when creating summary

karenzshea:
# Issue

This fixes the specific route in the faroe islands, `-6.788005,62.007014;-7.303738,62.076889` that is returning a dangling comma at the beginning ond end of summaries, but we haven't been able to find other cases of this.

As far as I can tell from reading how the street names are loaded into `m_name_table`, we aren't making any kind of checks for empty strings during summary concatenation, maybe because we used to assume that names ranges in `m_name_table` would never be empty.

I'm still wondering if we assume `name_id` of `0` is used to represent empty strings, as the original check is only checking if `name_id` is `0` or not. The `name_id` responsible for the empty string in our test case is not `0`. 

## Tasklist
 - [x] Investigate if street name data is being badly loaded
 - [x] add regression test
 - [ ] review
 - [ ] adjust for for comments

## Requirements / Relations

cc @daniel-j-h @MoKob


include/engine/guidance/assemble_leg.hpp
```diff
@@ -181,9 +181,10 @@ inline RouteLeg assembleLeg(const datafacade::BaseDataFacade &facade,
                                   summary_array.end(),
                                   facade.GetNameForID(summary_array.front()),
```
MoKob: what if this very first part is empty? We would end-up with `,name`

include/engine/guidance/assemble_leg.hpp
```diff
@@ -181,9 +181,10 @@ inline RouteLeg assembleLeg(const datafacade::BaseDataFacade &facade,
...
                                   [&facade](std::string previous, const std::uint32_t name_id) {
-                                      if (name_id != 0)
+                                      auto next_name = facade.GetNameForID(name_id);
+                                      if (name_id != 0 && !next_name.empty())
                                       {
-                                          previous += ", " + facade.GetNameForID(name_id);
+                                          previous += ", " + next_name;
```
MoKob: wasn't the plan here to include the ref, if the name is empty?

include/engine/guidance/assemble_leg.hpp
```diff
@@ -181,9 +181,10 @@ inline RouteLeg assembleLeg(const datafacade::BaseDataFacade &facade,
                                   summary_array.end(),
                                   facade.GetNameForID(summary_array.front()),
                                   [&facade](std::string previous, const std::uint32_t name_id) {
-                                      if (name_id != 0)
+                                      auto next_name = facade.GetNameForID(name_id);
+                                      if (name_id != 0 && !next_name.empty())
```
MoKob: `name_id != 0` is a superfluous check now. After all you requested the name already. In addition, we should check if the `name_id` is valid (https://github.com/Project-OSRM/osrm-backend/blob/master/include/util/typedefs.hpp#L81).

karenzshea:
@MoKob I opted to use boost's join function here. It's 
- cleaner than tooling around with std::accumulate and strings
- handles potentially empty name and ref fields better
- scales with changes to `MAX_USED_SEGMENTS`

include/engine/guidance/assemble_leg.hpp
```diff
@@ -172,21 +177,28 @@ inline RouteLeg assembleLeg(const datafacade::BaseDataFacade &facade,
...
+                const auto ref = facade.GetRefForID(name_id);
+                return ref;
+            }
+        };
+
+        const auto not_empty = [&](const std::string &name) { return !name.empty(); };
+
+        const auto summary_names = summary_array | boost::adaptors::transformed(name_id_to_string) |
+                                   boost::adaptors::filtered(not_empty);
+        summary = boost::algorithm::join(summary_names, ",");
```
daniel-j-h: Nice :+1:

Issue #2968, opened 2016-09-29
Name behavior is changing between 5.3 and 5.4


freenerd:
With the introduction of the name/ref split https://github.com/Project-OSRM/osrm-backend/issues/2704 we introduced a backwards incompatibility in the step objects:

```js
// 5.3
{
  "name": "Street Name (Street Reference)"
}

// 5.4
{
  "name": "Street Name",
  "ref": "Street Reference"
}
```

This change was motivated by a better new name detection https://github.com/Project-OSRM/osrm-backend/pull/2900 and to give clients better control over how text instructions are compiled and reference shields are displayed. In some cases, clients did regex parsing of the `name` to extract the `ref`. This was undocumented behaviour and we are okay with breaking that. Unfortunately, there are also cases where the ref is actually crucial information for guidance:

```js
// 5.3
{
  "name": "I-80"
}

// 5.4
{
  "name": "",
  "ref": "I-80"
}
```

This case we shall not break, since clients relying only on `name` will now miss information.

Let's find a backwards compatible implementation that retains the old design, but allows clients to also switch to the `name`/`ref` split now already. By Version v6 we should make the `name`/`ref` split mandatory.

freenerd:
In discussion, @MoKob @willwhite and me came to the conclusion that this is not actually a breaking change, since in the profiles developers can still decide to retain the old behaviour by using the [old name/ref code](https://github.com/Project-OSRM/osrm-backend/blob/5.3/profiles/car.lua#L414-L425).  This will not enable the new funky new name detection, but retains API compatibility for everyone who is looking for it.

PR #2969, opened 2016-09-29
Backport 5.3 behavior for name and add way_name

freenerd:
# Issue

Fixes https://github.com/Project-OSRM/osrm-backend/issues/2968

This PR aims at retaining the old 5.3 design of how `name` works, but also allows clients to benefit from the `name`/`ref` split now already. By the release of version 6, the `way_name` would be removed again and `step.name` would be use without modification in the json_factory.

For being compatible with the 5 and 6 version, we'd advise clients to use this logic:

```
name = step.way_name || step.name
```

## Tasklist
 - [ ] adjust changelog?
 - [ ] review
 - [ ] adjust for review comments
 - [ ] Get this into [5.4](https://github.com/Project-OSRM/osrm-backend/tree/5.4)

## Requirements / Relations

None

### DIFF
--- a/src/engine/api/json_factory.cpp
+++ b/src/engine/api/json_factory.cpp
@@ -234,9 +234,14 @@ util::json::Object makeRouteStep(guidance::RouteStep step, util::json::Value geo
     util::json::Object route_step;
     route_step.values["distance"] = std::round(step.distance * 10) / 10.;
     route_step.values["duration"] = std::round(step.duration * 10) / 10.;
-    route_step.values["name"] = std::move(step.name);
-    if (!step.ref.empty())
+    route_step.values["way_name"] = std::move(step.name);
+    if (step.ref.empty())
+        route.step.values["name"] = std::move(step.name);
+    else
+    {
+        route_step.values["name"] = std::move(step.name) << " (" << std::move(step.ref) << ")";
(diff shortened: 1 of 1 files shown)


src/engine/api/json_factory.cpp
```diff
@@ -234,9 +234,14 @@ util::json::Object makeRouteStep(guidance::RouteStep step, util::json::Value geo
...
-    route_step.values["name"] = std::move(step.name);
-    if (!step.ref.empty())
+    route_step.values["way_name"] = std::move(step.name);
+    if (step.ref.empty())
+        route.step.values["name"] = std::move(step.name);
+    else
+    {
+        route_step.values["name"] = std::move(step.name) << " (" << std::move(step.ref) << ")";
         route_step.values["ref"] = std::move(step.ref);
+    }
```
daniel-j-h: You're potentially moving multiple times from `step.name`; you're not allowed to do that.

Instead you have to copy the `step.name` string and (if you want to) move once at the very last.

freenerd:
This code won't work as expected, since it breaks the assumption `name_from_profile === name_in_output`.

Also https://github.com/Project-OSRM/osrm-backend/issues/2968#issuecomment-250473298

Issue #2965, opened 2016-09-29
Edge Update Tests Break on Cache


MoKob:
When running cucumber tests for edge-weight-updates for the first time, the test succeeds fine.

```
cucumber features/options/contract/edge-weight-updates-over-factor.feature:6
Node Version v4.0.0
@contract @options @edge-weight-updates-over-factor
Feature: osrm-contract command line option: edge-weight-updates-over-factor

  @contract @options @edge-weight-updates-over-factor
  Scenario: Logging weight with updates over factor of 2, long segment
    Given the node locations
      | node | lat | lon |
      | a    | 0.1 | 0.1 |
      | b    | .05 | 0.1 |
      | c    | 0.3 | 0.1 |
    And the ways
      | nodes | highway     |
      | ab    | residential |
      | ac    | primary     |
    Given the profile "testbot"
    Given the extract extra arguments "--generate-edge-lookup"
    Given the speed file
      """
      1,2,100
      1,3,100
      """
    And the data has been extracted
    When I run "osrm-contract --edge-weight-updates-over-factor 2 --segment-speed-file {speeds_file} {processed_file}"
    And stderr should contain "weight updates"
    And stderr should contain "New speed: 100 kph"
    And I route I should get
      | from | to | route | speed    |
      | a    | b  | ab,ab | 100 km/h |
      | a    | c  | ac,ac | 100 km/h |

1 scenario (1 passed)
10 steps (10 passed)
0m00.926s
```

On all succeeding runs, I consistently see the following error.

```
Node Version v4.0.0
@contract @options @edge-weight-updates-over-factor
Feature: osrm-contract command line option: edge-weight-updates-over-factor

  @contract @options @edge-weight-updates-over-factor
  Scenario: Logging weight with updates over factor of 2, long segment
    Given the node locations
      | node | lat | lon |
      | a    | 0.1 | 0.1 |
      | b    | .05 | 0.1 |
      | c    | 0.3 | 0.1 |
    And the ways
      | nodes | highway     |
      | ab    | residential |
      | ac    | primary     |
    Given the profile "testbot"
    Given the extract extra arguments "--generate-edge-lookup"
    Given the speed file
      """
      1,2,100
      1,3,100
      """
    And the data has been extracted
    When I run "osrm-contract --edge-weight-updates-over-factor 2 --segment-speed-file {speeds_file} {processed_file}"
    And stderr should contain "weight updates"
    And stderr should contain "New speed: 100 kph"
[...]

PR #2973, opened 2016-09-29
Avoid doubled reprocessing for scenarios without caching

oxidase:
# Issue

A small independent addition to  #2965: the step "the data has been saved to disk" should not trigger `reprocess` to only create stamp files for next steps. Now non-caching steps will  create stamps to prevent reprocessing at routing steps. This reduces for my local setup time from 3m16.455s to 2m56.530s.

## Tasklist
 - [x] review
 - [x] adjust for for comments

## Requirements / Relations
#2971

### DIFF
--- a/features/step_definitions/options.js
+++ b/features/step_definitions/options.js
@@ -1,5 +1,7 @@
-var assert = require('assert');
-var fs = require('fs');
+'use strict';
+
+const assert = require('assert');
+const fs = require('fs');
 
 module.exports = function () {
     this.resetOptionsOutput = () => {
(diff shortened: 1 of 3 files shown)


features/step_definitions/options.js
```diff
@@ -24,11 +24,19 @@ module.exports = function () {
     });
 
     this.When(/^I run "osrm\-extract\s?(.*?)"$/, (options, callback) => {
-        this.runAndSafeOutput('osrm-extract', options, callback);
+        let stamp = this.processedCacheFile + '.extract';
```
daniel-j-h: Using `let` here produces a syntax error - Travis failing.

oxidase: Ok, updated

Issue #2925, opened 2016-09-16
Make cucumber node maps easier to read


emiltin:
I suggest changing the format of node maps used in cucumber tests to make them easier to read/view.

Current syntax:

```gherkin
Given the node map
    |   |   | i | l |   |   |
    |   |   |   |   |   |   |
    | d |   | c | b |   | a |
    | e |   | f | g |   | h |
    |   |   |   |   |   |   |
    |   |   | j | k |   |   |
```

Proposed syntax:
```gherkin
Given the node map
    |     i l     |
    |             |
    | d   c b   a |
    | e   f g   h |
    |             |
    |     j k     |
```

The benefit is that the geometry is visually clearer and the tables are simpler. Current tables have the oddity that they are stretched to double width horizontally due to all the pipe "|" symbols.

The new syntax would also allow subtle visual queues when appropriate, ie. to indicate where ways are:

```gherkin
Given the node map
    |     i l     |
    |     . .     |
    | d . c b . a |
    | e . f g . h |
    |     . .     |
    |     j k     |
```

Such extra characters would simply be ignored, they are only to make the test easier to read.

Existing tables could be converted to the new syntax using a small script, since existing node names are fortunately limited to single characters. Both syntaxes could co-exists for a while if needed.

A possible drawback: Cucumber tables can be aligned automatically in most editors which is very practical. The new format can still be aligned, however some (all?) editors will left-align the cell content, which can mess up the map.

A way to avoid this drawback would be to enforce dots at empty node locations:

```gherking
Given the node map
    | . . i l . . |
    | . . . . . . |
    | d . c b . a |
    | e . f g . h |
    | . . . . . . |
    | . . j k . . | 
```

As long as you have nodes or dots at the left-most position, left-alignments will not change the content. It does gives you a kind of grid to work on which can be beneficial, but I feel it makes it harder to read the table.

MoKob:
I feel this is a very good idea. I have encountered this problem very often myself.

emiltin:
Cucumber trims leading spaces from the column string, ie. "    j k    "  is returned as "j k". Seems we would need to use dots for empty nodes or add quotes, which is a bit annoying.

emiltin:
There is no way to preserve leading spaces in cucumber tables. The best might to to switch to doc strings, either using a minimalist layout:

```gherkin
        Given the node map
            """
                a    
                b    
            d c   g h
                e    
                f    
            """
```

[...]

PR #2979, opened 2016-09-30
Make cucumber node maps easier to read by updating the syntax

emiltin:
# Issue

Make Cucumber node maps easier to read by changing the syntax. #2925.

## Tasklist
 - [x] update support code
 - [x] update all cucumber feature files
 - [x] review
 - [x] adjust for for comments

## Requirements / Relations
 Don't think there are any.

### DIFF
--- a/features/guidance/anticipate-lanes.feature
+++ b/features/guidance/anticipate-lanes.feature
@@ -8,11 +8,13 @@ Feature: Turn Lane Guidance
     @anticipate
     Scenario: Anticipate Lane Change for subsequent multi-lane intersections
         Given the node map
-            | a |   | b |   | x |   |   |
-            |   |   |   |   |   |   |   |
-            |   |   | c |   | d |   | z |
-            |   |   |   |   |   |   |   |
-            |   |   | y |   | e |   |   |
+            """
+            a – b – x
+                |
+                c – d – z
+                |   |
+                y   e
+            """
 
         And the ways
             | nodes | turn:lanes:forward           |
@@ -32,11 +34,13 @@ Feature: Turn Lane Guidance
     @anticipate
     Scenario: Anticipate Lane Change for quick same direction turns, staying on the same street
         Given the node map
-            | a |   | b | x |
-            |   |   |   |   |
-            |   |   | c |   |
-            |   |   |   |   |
-            | e |   | d | y |
+            """
+            a – b – x
+                |
+                c
+                |
+            e – d – y
+            """
 
         And the ways
             | nodes | turn:lanes:forward    | turn:lanes:backward | name |
@@ -55,11 +59,13 @@ Feature: Turn Lane Guidance
     @anticipate
     Scenario: Anticipate Lane Change for quick same direction turns, changing between streets
         Given the node map
-            | a |   | b | x |
-            |   |   |   |   |
-            |   |   | c |   |
-            |   |   |   |   |
-            | e |   | d | y |
+            """
+            a – b – x
+                |
+                c
+                |
+            e – d – y
+            """
 
         And the ways
             | nodes | turn:lanes:forward    | turn:lanes:backward | name |
@@ -79,9 +85,13 @@ Feature: Turn Lane Guidance
     @anticipate
     Scenario: Anticipate Lane Change for quick turns during a merge
         Given the node map
-            | a |   |   |   |   |
-            | x | b |   | c | y |
-            |   |   |   |   | d |
+            """
+            a
+              \
+            x – b – c – y
+                    |
+                    d
+            """
 
         And the ways
             | nodes | turn:lanes:forward       | name | highway       | oneway |
@@ -100,9 +110,12 @@ Feature: Turn Lane Guidance
     Scenario: Schoenefelder Kreuz
     # https://www.openstreetmap.org/way/264306388#map=16/52.3202/13.5568
         Given the node map
-            | a | b | x |   |   | i |
-            |   |   | c | d |   |   |
-            |   |   |   |   |   | j |
+            """
+            a b – x
+                \       / i
+                  c – d
+                        \ j
+            """
 
         And the ways
             | nodes | turn:lanes:forward                                  | lanes | highway       | oneway | name |
@@ -123,8 +136,11 @@ Feature: Turn Lane Guidance
     Scenario: Kreuz Oranienburg
     # https://www.openstreetmap.org/way/4484007#map=18/52.70439/13.20269
         Given the node map
-            | i |   |   |   |   | a |
-            | j |   | c | b |   | x |
+            """
+            i               a
+              ' .       . '
+            j – – c – b – – x
+            """
 
         And the ways
             | nodes | turn:lanes:forward | lanes | highway       | oneway | name |
@@ -143,11 +159,13 @@ Feature: Turn Lane Guidance
     @anticipate
     Scenario: Lane anticipation for fan-in
         Given the node map
-            | a |   | b |   | x |   |   |
-            |   |   |   |   |   |   |   |
-            |   |   | c |   | d |   | z |
-            |   |   |   |   |   |   |   |
-            |   |   | y |   | e |   |   |
+            """
+            a – b – x
+                |
+                c – d – z
+                |   |
+                y   e
+            """
 
         And the ways
             | nodes | turn:lanes:forward           | name |
@@ -166,11 +184,13 @@ Feature: Turn Lane Guidance
     @anticipate
     Scenario: Lane anticipation for fan-out
         Given the node map
-            | a |   | b |   | x |   |   |
-            |   |   |   |   |   |   |   |
-            |   |   | c |   | d |   | z |
-            |   |   |   |   |   |   |   |
-            |   |   | y |   | e |   |   |
+            """
+            a – b – x
+                |
+                c – d – z
+                |   |
+                y   e
+            """
 
         And the ways
             | nodes | turn:lanes:forward           | name |
@@ -189,11 +209,13 @@ Feature: Turn Lane Guidance
     @anticipate
     Scenario: Lane anticipation for fan-in followed by fan-out
         Given the node map
-            | a |   | b |   | x |   |   |
-            |   |   |   |   |   |   |   |
-            |   |   | c |   | d |   | z |
-            |   |   |   |   |   |   |   |
-            |   |   | y |   | e |   |   |
+            """
+            a – b – x
+                |
+                c – d – z
+                |   |
+                y   e
+            """
 
         And the ways
             | nodes | turn:lanes:forward           | name |
@@ -212,11 +234,13 @@ Feature: Turn Lane Guidance
     @anticipate
     Scenario: Lane anticipation for fan-out followed by fan-in
         Given the node map
-            | a |   | b |   | x |   |   |
-            |   |   |   |   |   |   |   |
-            |   |   | c |   | d |   | z |
-            |   |   |   |   |   |   |   |
-            |   |   | y |   | e |   |   |
+            """
+            a – b – x
+                |
+                c – d – z
+                |   |
+                y   e
+            """
 
         And the ways
             | nodes | turn:lanes:forward  | name |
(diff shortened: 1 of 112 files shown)


emiltin:
The new syntax allows the addition of eg. dots to indicate the visual layout of ways. These have not been added yet.

For reference, node maps where changed using the following Ruby script:

```ruby
def replace_file file
  puts file
  lines = File.readlines(file)
  
  table = false
  first = false
  indent = nil
  
  lines.each_with_index do |line, i|
    if table
      m = line.match /^(\s+)(\|.+\|)/     # still in table content?
      if m
        indent = m[1]
        content = m[2].scan( /[^\|]+/ ).map do |t|    # find content between pipes
          t = t.strip[0] || ' '                       # strip whitespace and use " " for empty cells
        end.join ' '                                  # join with a pace between neighbouring cells
        
        content.rstrip!
        
        changed_line = "#{indent}#{content}".rstrip
        
        if first
          # if start of table content, add doc string initiator
          str = %!#{indent}"""\n#{changed_line}\n!
        else
          str = %!#{changed_line}\n!
        end
        lines[i] = str
        first = false
      else
        # not in table content anymore, add doc string terminator
        lines[i] = %!#{indent}"""\n#{lines[i]}! unless first
        table = false
      end
    else
      if line =~ /the node map/     # start of a node map table?
        table = true
        first = true
      end
    end
  end
  
  File.open(file, 'w') { |f| f.write(lines.join) }
end


files = Dir.glob("./features/**/*.feature")
files.each do |file|
  puts file
  replace_file file
end
```
 
A few tests has comment lines using '#', these were removed manually first.

A few tests used a peculiar syntax where the node ids was set in the node map using syntax like "a:1", "b:2", etc. This has been change so node ids is set in a separate table.

emiltin:
I get a single failing test locally, but I don't understand what the test is doing.

```
1) Scenario: osrm-contract - Passing base file - features/options/contract/datasources.feature:27
   Step: Then datasource names should contain "lua profile,25_osrmcontract_passing_base_file_speeds" - features/options/contract/datasources.feature:29
   Step Definition: features/step_definitions/options.js:93
   Message:
     AssertionError: 'lua profile,27_osrmcontract_passing_base_file_speeds' == 'lua profile,25_osrmcontract_passing_base_file_speeds'
         at World.<anonymous> (/Users/emil/code/osrm-backend/features/step_definitions/options.js:95:16)
         at _combinedTickCallback (node.js:370:9)
         at process._tickCallback (node.js:401:11)
```

emiltin:
@MoKob anything additional you want me to do here?

MoKob:
At least we need to have the test-setup working again. Seems to be a minor linting error (unused variable).
If that is taken care of, we can do the full review process.

emiltin:
Travis chews on commit b8226d1, even though I committed dde3fc3 above - and it seem failing tests refer to versions that are not in my local repo anymore. Could it be because I did a forced push previously?

TheMarex:
@emiltin I would recommend rebasing and master and then force pushing again. It seems travis has had a hiccup.

emiltin:
Test setup is now working, all tests passing except the following one related to speed files:

```gherkin
    Background:
        Given the profile "testbot"
        Given the extract extra arguments "--generate-edge-lookup"
        And the node map
            """
            a b
            """
        And the ways
            | nodes |
            | ab    |
        And the speed file
        """
        1,2,27
        2,1,27
        2,3,27
        3,2,27
        1,4,27
        4,1,27
        """
        And the data has been extracted

    Scenario: osrm-contract - Passing base file
        When I run "osrm-contract --segment-speed-file {speeds_file} {processed_file}"
        Then datasource names should contain "lua profile,25_osrmcontract_passing_base_file_speeds"
        And it should exit successfully
```

I don't understand what is the following line is suppose to verify?

```gherkin
        Then datasource names should contain "lua profile,25_osrmcontract_passing_base_file_speeds"
```

The cucumber error is:
```
1) Scenario: osrm-contract - Passing base file - features/options/contract/datasources.feature:27
   Step: Then datasource names should contain "lua profile,25_osrmcontract_passing_base_file_speeds" - features/options/contract/datasources.feature:29
   Step Definition: features/step_definitions/options.js:93
   Message:
     AssertionError: 'lua profile,27_osrmcontract_passing_base_file_speeds' == 'lua profile,25_osrmcontract_passing_base_file_speeds'
         at World.<anonymous> (/Users/emil/code/osrm-backend/features/step_definitions/options.js:95:16)
         at _combinedTickCallback (node.js:370:9)
         at process._tickCallback (node.js:401:11)
```

The speed data contains the number 27 and the speed file on disk contains '27', but the test seem to look for the string '25'.

emiltin:
Do you want commits squashed?

MoKob:
I am always in favour of squashing commits.

``` Then datasource names should contain "lua profile,25_osrmcontract_passing_base_file_speeds"```
is testing for edge-lookup files. When we extract, `--generete-edge-lookup` provides a file that can be used to apply penalties/updates to the graph from other sources.

I can't see why the tests breaks, though.
Did something change the ID's of nodes generated? That would be my first guess.

File: features/guidance/anticipate-lanes.feature [hunk 1]
```diff
@@ -565,43 +609,45 @@ Feature: Turn Lane Guidance
...
-            |   |   |   |   |   |
-            |   |   |   |   |   |
-            |   |   |   |   |   |
-            |   |   |   |   |   |
-            |   |   |   |   |   |
-            |   |   | c |   |   |
+            """
+                a
+            x b   d y
+
```
MoKob: this would be a good candidate to show the usage of `.` to indicate roads. Here it is really hard to follow without the visual aid, in addition it would be good to have it in to see it working.

features/step_definitions/data.js
```diff
@@ -48,41 +48,26 @@ module.exports = function () {
...
+            cb();
         };
 
-        table.raw().forEach((row, ri) => {
-            row.forEach((name, ci) => {
-                q.defer(addNode, name, ri, ci);
+        docstring.split(/\n/).forEach( (row,ri) => {
+            row.split('').forEach( (cell,ci) => {
+                if( cell.match(/[a-z0-9]/) ) {
+                    q.defer(addNode, cell, ri, ci*0.5);
```
MoKob: This 0.5 indicates that we need a space between two nodes?

emiltin:
Why is test is looking for the string "25_osrmcontract_passing_base_file_speeds" in the speed file? Where does the 25 come from? It doesn't seem to relate to IDs, nor to the speeds specified in the file.

MoKob:
The name is for the caching infrastructure. It describes the test you are looking at.

In your case it starts on line 25 of `features/options/contract/datasources.feature` and is named `osrm-contract - Passing base file`. This is distilled into the `25_osrmcontract_passing_base_file_speeds` you see.

emiltin:
I see. It seems a a bit brittle to me to have the test include what line number it's starting on. But for now I'm just going to update the line number.

features/step_definitions/data.js
```diff
@@ -48,41 +48,26 @@ module.exports = function () {
...
-
-                cb();
+    this.Given(/^the node map$/, (docstring, callback) => {
+        var q = d3.queue();
+
+        var addNode = (name, ri, ci, cb) => {
+            var lonLat = this.tableCoordToLonLat(ci, ri);
+            if (name.match(/[a-z]/) ) {
+                if (this.nameNodeHash[name]) throw new Error(util.format('*** duplicate node %s', name));
+                this.addOSMNode(name, lonLat[0], lonLat[1], null);
```
MoKob: Is this removing the `node:id` feature here, or is this the combination with https://github.com/Project-OSRM/osrm-backend/pull/2979/files#diff-c41ecde0b11b4c1dc39c3b4a6f814478R109? Having IDs is kind of important, potentially. But I don't see that we need to do it in the docstring. 
If this is what happens in added table, I would be totally fine with it. No need to have it in the docstring.

emiltin: Previously, id could be set in the node map using the special syntax 'a:14'. This doesn't work with doc strings, since you cannot insert extra characters.
I changed it so you can specify IDs using the existing 'And the nodes' tables, using an 'id' column. There are a few tests that uses this.

features/step_definitions/data.js
emiltin: Yes. This is to ensure that the map doesn't look distorted (this was part of the original intention behind the PR). eg. that a square will look like a square and not a rectangle. Compare:

```gherkin
And the node map
    """
    a . b
    .   .
    c . d
    """

And the node map
    """
    a.b
    . .
    c.d
    """
```

With the current implementation, you're allowed to place nodes anywhere, so horizontally you can place nodes with 0.5 times the grid distance, vertically only on the grid. The following is valid and will result in distance of ac=1, cd=1, but ab=0.5:

```gherkin
And the node map
    """
    ab
    c d
    """
```

An alternative would be to enforce placement only on the grid, ie. always have spacing between nodes horizontally.

features/guidance/anticipate-lanes.feature
emiltin: Right. My approach was to not add these visual clues in the first commit, since adding them too all tests is a really big job. But I'm fine with adding them to a select set of scenarios.

MoKob: I also don't think that they are necessary everywhere. But they should be in at least a few to make sure all works as expected.

MoKob:
I guess ```1) Scenario: Obvious Index wigh very narrow turn to the right - features/guidance/turn.feature:1180
   Step: Given the node map - features/guidance/turn.feature:1181
   Step Definition: features/step_definitions/data.js:51
   Message:
     TypeError: docstring.split is not a function
         at World.<anonymous> (/home/travis/build/Project-OSRM/osrm-backend/features/step_definitions/data.js:66:19)
         at nextTickCallbackWith0Args (node.js:420:9)
         at process._tickCallback (node.js:349:13)
2) Scenario: Obvious Index wigh very narrow turn to the right - features/guidance/turn.feature:1197
   Step: Given the node map - features/guidance/turn.feature:1198
   Step Definition: features/step_definitions/data.js:51
   Message:
     TypeError: docstring.split is not a function
         at World.<anonymous> (/home/travis/build/Project-OSRM/osrm-backend/features/step_definitions/data.js:66:19)
         at nextTickCallbackWith0Args (node.js:420:9)
         at process._tickCallback (node.js:349:13)
3) Scenario: Obvious Index wigh very narrow turn to the left - features/guidance/turn.feature:1215
   Step: Given the node map - features/guidance/turn.feature:1216
   Step Definition: features/step_definitions/data.js:51
   Message:
     TypeError: docstring.split is not a function
         at World.<anonymous> (/home/travis/build/Project-OSRM/osrm-backend/features/step_definitions/data.js:66:19)
         at nextTickCallbackWith0Args (node.js:420:9)
         at process._tickCallback (node.js:349:13)
4) Scenario: Obvious Index wigh very narrow turn to the left - features/guidance/turn.feature:1232
   Step: Given the node map - features/guidance/turn.feature:1233
   Step Definition: features/step_definitions/data.js:51
   Message:
     TypeError: docstring.split is not a function
         at World.<anonymous> (/home/travis/build/Project-OSRM/osrm-backend/features/step_definitions/data.js:66:19)
         at nextTickCallbackWith0Args (node.js:420:9)
         at process._tickCallback (node.js:349:13)```

is due to a few testcases that have not beed adapted yet?

emiltin:
Yes, new scenarios that we're pulled in during latest rebase.

emiltin:
I updated docs/testing.md to reflect the new syntax. I guess I should update the wiki page  https://github.com/Project-OSRM/osrm-backend/wiki/Cucumber-Test-Suite as well? Or do you want to move that to the repo?

MoKob:
No, https://github.com/Project-OSRM/osrm-backend/wiki/Cucumber-Test-Suite will remain. The testing part is mostly a howto for writing better cucumber tests (since I've seen quite a collection of brittle tests, failing due to unrelated changes - just like your line number experience). 

When we merge this, we need to update the guide on the website as well.

For the rest, with the changes it looks fine to me. /cc @TheMarex @daniel-j-h any further input?

daniel-j-h:
Looks good to me! Thanks a lot @emiltin, great work! :+1:

TheMarex:
Yup looks good to me as well. :+1: Much more readable.

emiltin:
I'm adding visual hints to the scenarios in anticipate-lanes.feature. For some scenarios it makes sense to to use dashes and pipes:

```gherkin
        Given the node map
            """
            a
              \
            x – b – c y
                    |
                    d
            """
```

But in scenarios with ways with more unusual angles, it's better to use other symbols, eg. various dots:

```gherkin
        Given the node map
            """
            i               a
              ' .       . ' 
            j . . c . b . . x
            """
```
Each style has it's drawbacks and benefits. In light of this, do we want to stick to a single style or use what's best in each scenario?

emiltin:
I've added visual hints to anticipate-lanes.feature: https://github.com/Project-OSRM/osrm-backend/blob/dd512812583e2105c80721d06c1a05528284a206/features/guidance/anticipate-lanes.feature.  Think it's ready to merge; once that happens we can update the wiki.

PR #2991, opened 2016-10-03
Fixes forwarding refs for types derived from BaseParameters, resolves #2990

### LINKED ISSUE #2990: libosrm: *Parameters do not perfectly forward to BaseParameters
Our `BaseParameters` hold some potentially expensive vectors for coordinates, hints etc.

https://github.com/Project-OSRM/osrm-backend/blob/b821a1dd9803cb058db8861da93a84108e2db22b/include/engine/api/base_parameters.hpp#L64-L67

`RouteParameters` and others then forward their arguments

https://github.com/Project-OSRM/osrm-backend/blob/b821a1dd9803cb058db8861da93a84108e2db22b/include/engine/api/route_parameters.hpp#L79

but note this line: it's `Args... args_` and not `Args... &&args` making the `forward` below useless. We *always copy* those arguments. Fixing this by adding `&&` to all `*Parameters` breaks the API.

daniel-j-h:
For #2990: we forgot to pass parameters by forwarding references in types derived from `BaseParameters` always making (potentially expensive) copies.

Note: this breaks the `libosrm` API.

### DIFF
--- a/include/engine/api/match_parameters.hpp
+++ b/include/engine/api/match_parameters.hpp
@@ -61,7 +61,7 @@ struct MatchParameters : public RouteParameters
     }
 
     template <typename... Args>
-    MatchParameters(std::vector<unsigned> timestamps_, Args... args_)
+    MatchParameters(std::vector<unsigned> timestamps_, Args &&... args_)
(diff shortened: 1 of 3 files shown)


TheMarex:
Super-seeded by https://github.com/Project-OSRM/osrm-backend/pull/5926

Issue #2978, opened 2016-09-30
libosrm API change in v5.2


jcoupey:
I've been using `libosrm` with v5.1 and noticed my code didn't compile anymore with more recent versions.

This is caused by https://github.com/Project-OSRM/osrm-backend/commit/fa525ad610af14c03edcd366571c51c100f61d08 introducing a new `annotation_` parameter for the `osrm::RouteParameters`constructor.

I adjusted my code for this new parameter so it is now all right with `libosrm` versions starting at v5.2. Other users might need to be aware of this.

daniel-j-h:
Thanks for bringing this up: a breaking `libosrm` change should have never happened in a minor release. Seems like this one slipped through. We can't do anything about it at this point in time. Sorry.

Here's the breaking change:
https://github.com/Project-OSRM/osrm-backend/commit/fa525ad610af14c03edcd366571c51c100f61d08#diff-20ff672c2bee77817916303c66b1b98eR73

Clients have to adapt calls to the `RouteParameter`'s constructor: an additional `annotations` parameter was added as third argument for requesting metadata (OSM ids) for a route. Set `false` if you only want to make your code compile again.

TheMarex:
Urg sorry, this one slipped through. How are we going to handle additive changes to the parameters in the future? The only backward compatible change I see would either be default values for the new parameters or we disallow adding new parameters.
The latter option has the the downside that it would also translate into us not being able to add flags that would actually be backward compatible on node/HTTP level.

daniel-j-h:
In theory overloading the constructor should work, as in

v1:

```cpp
RouteParameters(bool opt0, bool opt1);
```

v2:

```cpp
RouteParameters(bool opt0, bool opt1, int breakingOpt);
RouteParameters(bool opt0, bool opt1) : RouteParameters(opt0, opt1, false) {}
```
in practice I'm not sure if the variadic template `args...` could be an issue.

daniel-j-h:
A more general idea would be to bring in something like my `libosrm` C-wrapper for which you can easily guarantee both API as well as ABI stability.

https://github.com/daniel-j-h/libosrmc/blob/master/libosrmc/osrmc.h 

Of course this comes with the need to keep the API up to date, but it's only a couple of lines per `libosrm` function, translating between typed objects and opaque pointers for the C API. This would also allow for quite easy FFI bindings.

TheMarex:
[...]

PR #2993, opened 2016-10-03
Re-introduces the old `RouteParameters` ctor for API compatibility, see #2978

daniel-j-h:
For #2978: this re-introduces the original `RouteParameters` constructor we had until 5.2 for compatibility.

Sorry we broke this between 5.2 and 5.4 :(


daniel-j-h:
Should we backport this to the 5.4 branch? I _think_ the overloaded ctor should be fine for clients, as I can see no way this would break existing client code targeting versions between 5.2 <-> 5.4.

TheMarex:
Looks good to me. :+1:

Issue #2926, opened 2016-09-16
Increased Memory Consumption: 5.4 RCs vs 5.3


daniel-j-h:
We're seeing a roughly 20% increase in memory utilization with the 5.4 RCs compared to 5.3 which we can not explain. Here's what we found so far.

---------------------------------------------------------------------------------------------------------------------------

From logs we see this happening during [extraction, parsing](https://github.com/Project-OSRM/osrm-backend/blob/70ee624e5c269052f04f3ffbdf61b6672888c245/src/extractor/extractor.cpp#L109).

It happens not only on bike but also on car (bike profile does not include lanes or turn restrictions but (!) the restriction parser is still getting called always returning [an empty optional](https://github.com/Project-OSRM/osrm-backend/blob/70ee624e5c269052f04f3ffbdf61b6672888c245/src/extractor/restriction_parser.cpp#L61-L64)).

It seems to happen on larger datasets only, e.g. on the planet. But not on smaller ones.

https://github.com/Project-OSRM/osrm-backend/issues/2704 => [String deduplication map](https://github.com/Project-OSRM/osrm-backend/blob/70ee624e5c269052f04f3ffbdf61b6672888c245/include/extractor/extractor_callbacks.hpp#L57-L61) size looks reasonable; dumping some stats about the map did not reveal anything (e.g. a lot of hash collisions).

Extractor allbacks are explicitly getting [allocated](https://github.com/Project-OSRM/osrm-backend/blob/70ee624e5c269052f04f3ffbdf61b6672888c245/src/extractor/extractor.cpp#L128) and [destroyed](https://github.com/Project-OSRM/osrm-backend/blob/70ee624e5c269052f04f3ffbdf61b6672888c245/src/extractor/extractor.cpp#L221) again.

Extraction containers are [scoped](https://github.com/Project-OSRM/osrm-backend/blob/70ee624e5c269052f04f3ffbdf61b6672888c245/src/extractor/extractor.cpp#L127) to the extractor and are cleaned up once we serialized the data out.

-----------------------------------------------------------------------------------------------------------------------------

Other possibilities we need to investigate:

Touching the restriction parser code: https://github.com/Project-OSRM/osrm-backend/pull/2829

[...]

PR #2995, opened 2016-10-03
Reworks how we search for Lua in CMake and constrain to Lua 5.1 5.2

daniel-j-h:
At the moment we have a FindLua52.cmake module that is happy with
Lua52 but also Lua51 and even Lua53. Let's instead pull in the official
one from CMake (since it's not yet in CMake 2.8 which we support).

https://github.com/Kitware/CMake/blob/master/Modules/FindLua.cmake

BSD licenced.

Requires Lua 5.1 or Lua 5.2, since seeing memory issues on 5.3: https://github.com/Project-OSRM/osrm-backend/issues/2926


daniel-j-h:
Before merging: this is a larger build system change I'd like to test with various setups to make sure we're not breaking things. Because of this, I don't want to include it into 5.4.

MoKob:
Tested it on my OSX setup. Works fine (removed build folder beforehand and started clean).

TheMarex:
Looks good to me. Removing the brittle compile check is a good step.

PR #2996, opened 2016-10-03
Only look for lua 5.1 or 5.2

karenzshea:
# Issue

This could be the uglier precursor to https://github.com/Project-OSRM/osrm-backend/pull/2995. No hard feelings if this isn't what we're looking for.

## Tasklist
 - [ ] add test?
 - [ ] review
 - [ ] adjust for for comments

## Requirements / Relations
cc @daniel-j-h

### DIFF
--- a/cmake/FindLua52.cmake
+++ b/cmake/FindLua52.cmake
@@ -29,7 +29,7 @@
 find_path(LUA_INCLUDE_DIR lua.h
   HINTS
     ENV LUA_DIR
-  PATH_SUFFIXES include/lua52 include/lua5.2 include/lua-5.2 include/lua include
+  PATH_SUFFIXES include/lua52 include/lua5.2 include/lua-5.2 include include/lua51 include/lua5.1 include/lua-5.1
   PATHS
   ~/Library/Frameworks
   /Library/Frameworks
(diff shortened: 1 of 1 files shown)


cmake/FindLua52.cmake
```diff
@@ -29,7 +29,7 @@
 find_path(LUA_INCLUDE_DIR lua.h
   HINTS
     ENV LUA_DIR
-  PATH_SUFFIXES include/lua52 include/lua5.2 include/lua-5.2 include/lua include
+  PATH_SUFFIXES include/lua52 include/lua5.2 include/lua-5.2 include include/lua51 include/lua5.1 include/lua-5.1
```
daniel-j-h: Are there situations where Lua51 or Lua52 are installed in `/usr/include/lua` which we would break now? I think the safer bet would be to fix the version parsing and bail out on version major 5, minor != 1 or 2.

daniel-j-h: Also @MoKob just reported on his OSX / brew setup:

> ls /usr/local/include/lua
> lua-5.1/    lua5.1/     luabind/    luajit-2.0/

will removing the `/include/lua` search path break this setup?

PR #2998, opened 2016-10-04
Squashes non-internal 5.4 RC changelogs into 5.4 release

daniel-j-h:
Instead of having the changelog look like

> 5.4-rc.x
>   ..
> 5.4-rc.y
>   ..
> 5.4
>   ..

this squashes RC changelog items into the single 5.4-release item

> 5.4
>   ..

Note: this also removes changelog items for bugs / issues we introduced while doing RCs and then fixed in another RC (such as the summary comma bug we had between rc6 and rc7).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,38 +1,3 @@
-# 5.4.0-rc.7
-  - Chages from 5.4.0-rc.6
-    - Bugfixes re-introduce space between two entries in summaries
-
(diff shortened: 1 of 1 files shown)


daniel-j-h:
After talk with @MoKob we want the 5.4 changelog to reflect changes between rcs; master has the cleaned up changelog for the 5.4 release.Closing here.

PR #3000, opened 2016-10-04
Implements Cucumber shared vs. static library detection, fixes #2999

### LINKED ISSUE #2999: Cucumber Rewrite Breaks with Shared Libraries / Assumes Static Libosrm Libraries
The Cucumber rewrite assumes `libosrm_*.a` libraries. Shared `libosrm_*.so` let the Tests fail with

> Error: features/support/hooks.js:11 ENOENT: no such file or directory, open '/tmp/osrm-backend/build/libosrm_extract.a'

The reason for this is the following `TODO`:

https://github.com/Project-OSRM/osrm-backend/blob/94ca7828163bfd2334fdad903b465b99703752c4/features/support/env.js#L54-L55

checking the env var only works on Travis because we're setting the var [here](https://github.com/Project-OSRM/osrm-backend/blob/32c5f14ed30cfb0f7182e2e50ce6d893337b5b9b/.travis.yml#L94) to pass it to CMake [below](https://github.com/Project-OSRM/osrm-backend/blob/32c5f14ed30cfb0f7182e2e50ce6d893337b5b9b/.travis.yml#L134).

This needs to be fixed.

daniel-j-h:
For #2999: Cucumber Rewrite Breaks with Shared Libraries / Assumes Static Libosrm Libraries.

## Tasklist
 - [ ] review
 - [ ] adjust for for comments

This needs to be cherry-picked into the `5.4` branch.

### DIFF
--- a/features/support/env.js
+++ b/features/support/env.js
@@ -51,8 +51,26 @@ module.exports = function () {
         } else {
             this.TERMSIGNAL = 'SIGTERM';
             this.EXE = '';
-            // TODO autodetect if this was build with shared or static libraries
-            this.LIB = process.env.BUILD_SHARED_LIBS && '.so' || '.a';
+
+            // heuristically detect .so/.a suffix
+            this.LIB = null;
+
+            try {
+                const dot_a = util.format('%s/libosrm%s', this.BIN_PATH, '.a');
+                fs.accessSync(dot_a, fs.F_OK);
+                this.LIB = '.a';
+            } catch(e) { /*nop*/ }
+
+            try {
+                const dot_so = util.format('%s/libosrm%s', this.BIN_PATH, '.so');
+                fs.accessSync(dot_so, fs.F_OK);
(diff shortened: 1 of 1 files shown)


features/support/env.js
```diff
@@ -51,8 +51,24 @@ module.exports = function () {
...
+            try {
+                fs.accessSync(dot_a, fs.F_OK);
+                this.LIB = '.a';
+            } catch(e) { /*nop*/ }
+
+            try {
+                fs.accessSync(dot_so, fs.F_OK);
+                this.LIB = '.so';
+            } catch(e) { /*nop*/ }
+
```
freenerd: wouldn't it be better if this would fail if neither `.a` nor `.so` are detected?

`if (!this.LIB) { throw 'could not detect neither .so nor .a'; }`

daniel-j-h: :+1:

Issue #3002, opened 2016-10-04
Missing Turn Instruction


MoKob:
I am seeing a missing turn instruction for http://map.project-osrm.org/?z=18&center=39.281743%2C-76.623046&loc=39.280778%2C-76.623325&loc=39.282198%2C-76.622933&hl=en&alt=0

<img width="735" alt="screen shot 2016-10-04 at 17 08 40" src="https://cloud.githubusercontent.com/assets/12932279/19079825/41512c88-8a55-11e6-99de-afb199d734ba.png">

PR #3006, opened 2016-10-05
don't provide turn-straight next to suppress-straight

MoKob:
# Issue

resolves, https://github.com/Project-OSRM/osrm-backend/issues/3002

## Tasklist
 - [x] review
 - [x] adjust for for comments

## Requirements / Relations
none

### DIFF
--- a/src/extractor/guidance/turn_handler.cpp
+++ b/src/extractor/guidance/turn_handler.cpp
@@ -1,7 +1,7 @@
+#include "extractor/guidance/turn_handler.hpp"
 #include "extractor/guidance/constants.hpp"
 #include "extractor/guidance/intersection_scenario_three_way.hpp"
 #include "extractor/guidance/toolkit.hpp"
-#include "extractor/guidance/turn_handler.hpp"
 
 #include "util/guidance/toolkit.hpp"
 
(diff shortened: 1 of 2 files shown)


src/extractor/guidance/turn_handler.cpp
```diff
@@ -166,15 +166,22 @@ Intersection TurnHandler::handleThreeWayTurn(const EdgeID via_edge, Intersection
...
         if (obvious_index == 1)
         {
             intersection[1].turn.instruction = getInstructionForObvious(
                 3, via_edge, isThroughStreet(1, intersection), intersection[1]);
         }
         else
         {
-            intersection[1].turn.instruction = {findBasicTurnType(via_edge, intersection[1]),
-                                                getTurnDirection(intersection[1].turn.angle)};
+            if (obvious_index != 0 && direction_at_one == direction_at_two &&
```
TheMarex: I had a little trouble following the cases here. It is not obvious what value obvious_index is (pun not intended). The readability might be improved by first dividing control flow on that value, instead of doing it in a nested manner.

```
if (obvious index == 0)
else if (obvious_index == 1)
else {
BOOST_ASSERT (obvious_index == 2);
}
```

TheMarex:
Control flow got a little harder to read, other then that this looks fine. :+1:

Issue #3003, opened 2016-10-04
Priority Roads for Profiles


daniel-j-h:
Related to [stop signs](https://github.com/Project-OSRM/osrm-backend/issues/2652) we're currently unaware of priority roads.

http://wiki.openstreetmap.org/wiki/Key:priority_road
http://taginfo.openstreetmap.org/keys/priority_road#values

Tagging:
- `priority_road:forward`
- `priority_road:backward`
- `priority_road`

on ways.

daniel-j-h:
Priority tags in Berlin:

![priority roads in Berlin](https://cloud.githubusercontent.com/assets/527241/19080582/ccbd0ccc-8a57-11e6-9ccc-28401d6d5dba.png)

daniel-j-h:
Since the concept of priority roads is non-existent in the US it would be great to also consider the complementary [`highway=give_way`](http://wiki.openstreetmap.org/wiki/Tag:highway%3Dgive_way) which is tagged similarly to stop-tag-on nodes for [Stop Signs](https://github.com/Project-OSRM/osrm-backend/issues/2652).

Issue #2652, opened 2016-07-13
Stop Signs for Profiles


daniel-j-h:
Profiles are currently unaware of stop signs.
We want them for better ETAs.

Tagging: `highway=stop` on nodes.

http://wiki.openstreetmap.org/wiki/Tag:highway%3Dstop
http://taginfo.openstreetmap.org/tags/highway=stop

1ec5:
The wiki page makes it's tagging recommendation sound authoritative, but how prevalent are tags such as [`priority_road`](http://wiki.openstreetmap.org/wiki/Key:priority_road) in places like North America, where the concept of a priority road has no formal/legal status and would not be signposted as such? In my armchair-mapping travels in the U.S., I've seen three tagging styles that closely parallel tagging for traffic signals, and usage for each is much less common than for traffic signals:

1. `highway=stop` on the intersection node, even for non-all-way stops
2. Less commonly, `highway=stop` on each approach
3. Just as commonly as 2, a [stop relation](http://wiki.openstreetmap.org/wiki/Proposed_features/Relation:type%3Dstop)

Hopefully my observations are an outlier and approach 2 is the most common generally.

MoKob:
From discussion on traffic signals (https://github.com/Project-OSRM/osrm-backend/issues/2758#issue-170197459), stop-signs cannot be modelled just like traffic signals since stop-signs give a precedence to one way while penalising another. The same can be said to some extend for traffic signals -- the phases are not necessarily of the same duration -- but nothing we have currently in our scope.

The issue here is: we need to assign the stop-sign only to a specific way direction.

danpat:
In the last couple of days, I've been doing a bunch of test driving with a high-resolution GPS and looking to improved the Lua `turn_function`.

One thing is fairly clear from the data - whether you need to stop or not makes a big difference in the turn cost.

Right-hand turn that required a stop (this intersection has a stop sign for traffic heading west, north-south is a 'priority road' with no stopping required):
![screen shot 2016-09-02 at 12 34 35 am](https://cloud.githubusercontent.com/assets/1892250/18213441/210a277a-70fd-11e6-8a25-4a18d24f7abd.png)

Right-hand turn that required no stop:
![screen shot 2016-09-02 at 12 31 32 am](https://cloud.githubusercontent.com/assets/1892250/18213457/367619f2-70fd-11e6-9aab-4555b9defe5c.png)

[...]

PR #3009, opened 2016-10-05
Road Priorities: Stop Signs, Give Ways

daniel-j-h:
For https://github.com/Project-OSRM/osrm-backend/issues/2652 and https://github.com/Project-OSRM/osrm-backend/issues/3003.

The goal is to pick up stop signs and give ways from the data in the profiles, determine the way direction the node tags apply to, pass it through the pipeline and eventually influence turn penalties.

Already opening here so we can discuss. WIP.

After double checking the tagging schemes, relation-based tagging is not really an option:
- http://taginfo.openstreetmap.org/tags/type=stop
- http://taginfo.openstreetmap.org/tags/type=give_way
## Tasklist
- [ ] detect a way's direction by checking distance to next intersection
- [ ] stop signs: tags on intersection node vs tags on node approaching intersection
- [x] `priority_road` tags skip these first two steps; attach to edges directly
- [ ] influence turn penalty
- [ ] review
- [ ] adjust for for comments


daniel-j-h:
Plan for handling arbitrary node penalties is in https://github.com/Project-OSRM/osrm-backend/issues/3021#issuecomment-254903620 - I rebased this onto master and ripped out the WIP stop sign / give way sign work. @MoKob want to have a look and incorporate this into your turn function work already?

daniel-j-h:
This was de-prioritized. It's is blocked / goes hand in hand with https://github.com/Project-OSRM/osrm-backend/pull/2912.
Not considering it for the 5.7 release.

Issue #2679, opened 2016-07-21
Lanes Left / Right of Turn Lanes


daniel-j-h:
https://github.com/Project-OSRM/osrm-backend/commit/0e017a6ce53bf471b7e5671cddd745fa3a158914 as well as https://github.com/Project-OSRM/osrm-backend/pull/2671 need to know the lanes left of / right of a turn's lanes.

Such a feature should live in a toolkit header rather than duplicating the logic.

We should do this once advanced-lanes https://github.com/Project-OSRM/osrm-backend/pull/2614 is merged into master.
Otherwise it's too much of a headache to rebase all depending branches.

daniel-j-h:
To be clear here: we should have two functions in `toolkit.hpp` for calculating the lanes to the left and to the right as in comparing the lane description and the lane tuple. See https://github.com/Project-OSRM/osrm-backend/commit/0e017a6ce53bf471b7e5671cddd745fa3a158914#diff-99adcb8bccf8e94a176bdf92509ff070R1170 for an example and the links above.

TheMarex:
This was implemented in #3010.

PR #3010, opened 2016-10-05
Add toolkit function that returns right and left turn lane info

karenzshea:
# Issue

https://github.com/Project-OSRM/osrm-backend/issues/2679

This should add a layer of abstraction atop accessing other turn lanes from a given turn lane. Eventually there will be
- function `lanesToTheRight` which returns a range of turn lane description of lanes
- function `lanesToTheLeft` which returns a range of turn lane description of lanes
- function `numLanesToTheRight`, returns number of lanes
- function `numLanesToLeft`, returns number of lanes

## Tasklist
 - [x] Write `numLanesToTheRight/Left` functions
-  [x] Implement in LaneAnticipation
-  [x] Implement in `collapseUseLane`
-  [ ] Add unit test
 - [x] review
 - [ ] adjust for for comments

### DIFF
--- a/include/extractor/guidance/toolkit.hpp
+++ b/include/extractor/guidance/toolkit.hpp
@@ -16,6 +16,8 @@
 #include "extractor/guidance/intersection.hpp"
 #include "extractor/guidance/turn_instruction.hpp"
 
+#include "engine/guidance/route_step.hpp"
+
 #include <algorithm>
 #include <cmath>
 #include <cstdint>
@@ -26,6 +28,7 @@
 
 #include <boost/algorithm/string.hpp>
 #include <boost/functional/hash.hpp>
+#include <boost/range/iterator_range.hpp>
 #include <boost/tokenizer.hpp>
 
 namespace osrm
@@ -35,8 +38,8 @@ namespace extractor
 namespace guidance
 {
 
-using util::guidance::LaneTupelIdPair;
-using LaneDataIdMap = std::unordered_map<LaneTupelIdPair, LaneDataID, boost::hash<LaneTupelIdPair>>;
+using util::guidance::LaneTupleIdPair;
+using LaneDataIdMap = std::unordered_map<LaneTupleIdPair, LaneDataID, boost::hash<LaneTupleIdPair>>;
 
 using util::guidance::angularDeviation;
 using util::guidance::entersRoundabout;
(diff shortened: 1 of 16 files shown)


include/extractor/guidance/toolkit.hpp
```diff
@@ -285,6 +288,60 @@ inline std::string applyAccessTokens(std::string lane_string, const std::string
...
 }
 
+LaneID inline numLanesToTheRight(const engine::guidance::RouteStep &step)
+{
+    return step.intersections.front().lanes.first_lane_from_the_right;
+}
+
+LaneID inline numLanesToTheLeft(const engine::guidance::RouteStep &step)
+{
+    LaneID total = step.intersections.front().lane_description.size();
```
daniel-j-h: Can be `const`

include/extractor/guidance/toolkit.hpp
```diff
@@ -285,6 +288,60 @@ inline std::string applyAccessTokens(std::string lane_string, const std::string
...
+    const auto farthest_left_turn_lane =
+        lane_description.size() - (lanes.first_lane_from_the_right + lanes.lanes_in_turn);
+    // check that lanes even exist left of the turn lanes
+    if (farthest_left_turn_lane != 0)
+    {
+        // return a range from the farthest left lane, to the first lane left of turn lanes
+        const auto first_left_of_turn_lanes = lane_description.begin() + (farthest_left_turn_lane);
+        auto out_range =
+            boost::make_iterator_range(lane_description.begin(), first_left_of_turn_lanes);
+        return out_range;
```
daniel-j-h: You can simply `return make_it_range(..)`

include/extractor/guidance/toolkit.hpp
```diff
@@ -285,6 +288,60 @@ inline std::string applyAccessTokens(std::string lane_string, const std::string
...
+{
+    LaneID total = step.intersections.front().lane_description.size();
+    return total - (step.intersections.front().lanes.lanes_in_turn +
+                    step.intersections.front().lanes.first_lane_from_the_right);
+}
+
+template <class Iterator>
+inline boost::iterator_range<Iterator>
+lanesToTheLeft(const util::guidance::LaneTuple &lanes,
+               const extractor::guidance::TurnLaneDescription &lane_description)
```
daniel-j-h: What about changing the function to
- operate on a `const RouteStep &` similarly to the functions above
- not having to say `lanesToTheLeft<Iterator>(lanes, desc)`

the return type depens on `lane_description`, so you should be able to say

    auto lanes(.., desc) -> iterator_range<decltype(desc.begin())> { make_it_range(..); }

or even

    inline auto lanes(..) { return make_it_range(..); }

then users don't have to care about the iterator type and can simply call the function with the compiler deriving the correct type.

include/extractor/guidance/toolkit.hpp
```diff
@@ -285,6 +288,60 @@ inline std::string applyAccessTokens(std::string lane_string, const std::string
...
+        const auto first_left_of_turn_lanes = lane_description.begin() + (farthest_left_turn_lane);
+        auto out_range =
+            boost::make_iterator_range(lane_description.begin(), first_left_of_turn_lanes);
+        return out_range;
+    }
+    else
+    {
+        return boost::make_iterator_range(lane_description.end(), lane_description.end());
+    }
+}
```
daniel-j-h: Conceptually, can't you use your functions above to get the number of lanes left / right of the turn and then do

    numLeft = numLanesToLeft(step);
    return make_it_range(begin(lanes), begin(lanes) + numLeft);

if numLeft is zero you get two iterators back where first == last, so an empty range.
If not you slice the lanes into a sub-range.

src/engine/guidance/post_processing.cpp
```diff
@@ -1294,33 +1299,36 @@ std::vector<RouteStep> collapseUseLane(std::vector<RouteStep> steps)
...
-                return false;
+    const auto canCollapseUseLane = [containsTag](
+        const util::guidance::LaneTuple lanes,
+        const extractor::guidance::TurnLaneDescription lane_description) {
+        // the lane description is given left to right, lanes are counted from the right.
+        // Therefore we access the lane description using the reverse iterator
+        using iter_type = extractor::guidance::TurnLaneDescription::const_iterator;
+
+        boost::iterator_range<iter_type> right_most_lanes =
+            extractor::guidance::lanesToTheRight<iter_type>(lanes, lane_description);
```
daniel-j-h: This is the ugly callsite we want to spare the user from having to write. The compiler should be able to deduce the type if you work out the notes above.

daniel-j-h:
:+1:

PR #3014, opened 2016-10-06
Silence static visitor upstream deprecation warnings

daniel-j-h:
https://github.com/mapbox/variant/blob/master/README.md#deprecations
https://github.com/mapbox/variant/blob/master/include/mapbox/variant.hpp#L75-L83


TheMarex:
LGTM!

PR #3012, opened 2016-10-06
Make DataFacade local to every request

TheMarex:
# Issue

This is the first step to having fine grained locking on data updates,
see issue #2570. I'm breaking this out of the PR since this is a huge changeset and should go into master quickly or merge chaos will ensure. There is no real functionality change here, however this is a fundamental structural change.

## Tasklist
 - [x] review
 - [x] adjust for for comments

## Requirements / Relations

Requirement of #2570.

/cc @daniel-j-h

PR #3015, opened 2016-10-07
Make SharedDataFacade immutable

### LINKED ISSUE #2570: Move shared memory block swapping control to osrm-datastore

TheMarex:
# Issue

This is the second step of #2570. ~~Again this does not really improve the locking. There might even be a little bit more locking with this, as some mutexes are redundant still. It mainly cleans up the code by separating responsibilities better.~~

Switched this to fully implement #2570 because after #3017 that was an easy step. This changes the behavior of `osrm-datastore` in case there is already an update going on: It will return with an error. We don't rely anymore on having no having no requests running, thanks to the reader-writer lock:

1. Once `osrm-datastore` is ready to do the final switch, it acquires an exclusive lock that blocks all new queries.
2. Since this update is very quick, the delay between switching and new requests picking up the dataset is very short.
3. Requests already running are not affected by this, since they retain the old memory until they are finished over are reader-lock.

Testing concurrency is quite hard in an automated fashion. The cucumber tests work, but I'll only take this a general indicator for working and will do manual tests.

## Tasklist
 - [x] run `osrm-datastore` multiple times while query is running
 - [x] Try concurrent datastore runs
     - [x] with requests
     - [x] without requests
 - [x] Try hitting the server hard with a mix of slow and long queries and switch datasets
 - [x] same with fsanitize=thread
 - [x] review
 - [x] adjust for for comments

## Requirements / Relations

This is the successor of #3012 on the quest to #2570. This closes #2570.
Merge #3017  before this.

### DIFF
--- a/include/engine/data_watchdog.hpp
+++ b/include/engine/data_watchdog.hpp
@@ -3,9 +3,9 @@
 
 #include "engine/datafacade/shared_datafacade.hpp"
 
+#include "storage/shared_barriers.hpp"
 #include "storage/shared_datatype.hpp"
 #include "storage/shared_memory.hpp"
-#include "storage/shared_barriers.hpp"
 
 #include <boost/interprocess/sync/named_upgradable_mutex.hpp>
 #include <boost/thread/lock_types.hpp>
@@ -102,8 +102,10 @@ class DataWatchdog
         boost::upgrade_to_unique_lock<boost::upgrade_mutex> unique_facade_lock(facade_lock);
 
         current_timestamp = *shared_timestamp;
-        facade = std::make_shared<datafacade::SharedDataFacade>(
-            current_timestamp.layout, current_timestamp.data, current_timestamp.timestamp);
+        facade = std::make_shared<datafacade::SharedDataFacade>(shared_barriers,
+                                                                current_timestamp.layout,
+                                                                current_timestamp.data,
+                                                                current_timestamp.timestamp);
 
         return get_locked_facade();
     }
--- a/include/engine/data_watchdog.hpp
+++ b/include/engine/data_watchdog.hpp
@@ -1,3 +1,6 @@
+#ifndef OSRM_ENGINE_DATA_WATCHDOG_HPP
+#define OSRM_ENGINE_DATA_WATCHDOG_HPP
+
 #include "engine/datafacade/shared_datafacade.hpp"
 
 #include "storage/shared_datatype.hpp"
@@ -82,3 +85,5 @@ class DataWatchdog
 };
 }
 }
+
+#endif
--- a/include/engine/data_watchdog.hpp
+++ b/include/engine/data_watchdog.hpp
@@ -6,8 +6,8 @@
 #include "storage/shared_datatype.hpp"
 #include "storage/shared_memory.hpp"
 
-#include <boost/thread/shared_mutex.hpp>
 #include <boost/thread/lock_types.hpp>
+#include <boost/thread/shared_mutex.hpp>
 
 #include <memory>
 
@@ -29,7 +29,7 @@ class DataWatchdog
   public:
     DataWatchdog()
         : shared_regions(storage::makeSharedMemoryView(storage::CURRENT_REGIONS)),
-        current_timestamp {storage::LAYOUT_NONE, storage::DATA_NONE, 0}
+          current_timestamp{storage::LAYOUT_NONE, storage::DATA_NONE, 0}
     {
     }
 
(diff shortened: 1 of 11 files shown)


include/engine/data_watchdog.hpp
```diff
@@ -0,0 +1,84 @@
+#include "engine/datafacade/shared_datafacade.hpp"
```
daniel-j-h: Missing header guards

include/engine/data_watchdog.hpp
```diff
@@ -0,0 +1,84 @@
...
+        return storage::SharedMemory::RegionExists(storage::CURRENT_REGIONS);
+    }
+
+    // Check if it might be worth to try to aquire a exclusive lock
+    bool HasNewRegion() const
+    {
+        const boost::shared_lock<boost::shared_mutex> lock(current_timestamp_mutex);
+
+        const auto shared_timestamp =
+            static_cast<const storage::SharedDataTimestamp *>(shared_regions->Ptr());
```
daniel-j-h: Hm, can 1/ the shared_regions unique_ptr ever be empty, 2/ the ptr returned here ever be a nullptr?

(same below)

include/engine/datafacade/shared_datafacade.hpp
```diff
@@ -177,6 +188,11 @@ class SharedDataFacade final : public BaseDataFacade
             coordinate_list_ptr,
             data_layout->num_entries[storage::SharedDataLayout::COORDINATE_LIST]);
 
+        for (unsigned i = 0; i < m_coordinate_list.size(); ++i)
```
daniel-j-h: std::size_t

include/engine/datafacade/shared_datafacade.hpp
```diff
@@ -367,102 +383,32 @@ class SharedDataFacade final : public BaseDataFacade
...
-            util::SimpleLogger().Write(logDEBUG) << "Updates available, getting exclusive lock";
-            const boost::lock_guard<boost::shared_mutex> lock(data_mutex);
-
-            if (CURRENT_LAYOUT != data_timestamp_ptr->layout ||
-                CURRENT_DATA != data_timestamp_ptr->data)
-            {
-                // release the previous shared memory segments
-                storage::SharedMemory::Remove(CURRENT_LAYOUT);
-                storage::SharedMemory::Remove(CURRENT_DATA);
+        data_layout = static_cast<storage::SharedDataLayout *>(m_layout_memory->Ptr());
```
daniel-j-h: can both ptrs ever be nullptr?

src/engine/engine.cpp
```diff
@@ -35,6 +36,8 @@ struct Engine::EngineLock
 {
     // will only be initialized if shared memory is used
     storage::SharedBarriers barrier;
+    // ensures that when we set facade we can do it without race conditions
+    std::mutex facade_update_mutex;
```
daniel-j-h: mutable

src/engine/engine.cpp
```diff
@@ -99,11 +103,32 @@ RunQuery(const std::unique_ptr<osrm::engine::Engine::EngineLock> &lock,
...
-    // things while the query is running
-    boost::shared_lock<boost::shared_mutex> data_lock{shared_facade.data_mutex};
+    {
+        // this lock ensures that we are never overtaken while creating a new
+        // facade and setting it.
+        // This is important since we need to ensure there is always exactly
+        // one facade per shared memory region.
+        // TODO: Remove this once the SharedDataFacade doesn't own the shared memory
+        // segment anymore.
+        std::lock_guard<std::mutex> update_lock(lock->facade_update_mutex);
```
daniel-j-h: Ah! This is important, I already was wondering about the `shared_ptr &` above! Thanks for the comment!

include/engine/engine.hpp
```diff
@@ -72,6 +74,7 @@ class Engine final
 
   private:
     std::unique_ptr<EngineLock> lock;
+    std::unique_ptr<DataWatchdog> watchdog;
```
daniel-j-h: unique_ptr to hide the impl.?

include/engine/data_watchdog.hpp
TheMarex: No `makeSharedMemory` throws an exception if it can access it and create SharedMemory object.

include/engine/engine.hpp
TheMarex: yep.

TheMarex:
Updated the description since this now fully implements #2570.

TheMarex:
Finally found a solution for the stale mutexes problems: `osrm-datastore` has a new parameter `--max-wait` that allows it to only wait for a specified amount of time before it forces a release of the locks that could be blocked by crashed queries. The default is still to wait indefinitely, but in high-availability scenarios we could specify 60 seconds, which seems sensible.

include/engine/datafacade/shared_datafacade.hpp
```diff
@@ -143,6 +145,16 @@ class SharedDataFacade final : public BaseDataFacade
...
         BOOST_ASSERT_MSG(!m_coordinate_list.empty(), "coordinates must be loaded before r-tree");
 
+        const auto file_index_ptr = data_layout->GetBlockPtr<char>(
+            shared_memory, storage::SharedDataLayout::FILE_INDEX_PATH);
+        file_index_path = boost::filesystem::path(file_index_ptr);
+        if (!boost::filesystem::exists(file_index_path))
+        {
+            util::SimpleLogger().Write(logDEBUG) << "Leaf file name " << file_index_path.string();
+            throw util::exception("Could not load leaf index file. "
+                                  "Is any data loaded into shared memory?");
```
danpat: Can we re-word this error message?  Maybe "OSRM shared memory refers to a required file that doesn't exist: filename".

src/storage/storage.cpp
```diff
@@ -94,32 +126,63 @@ int Storage::Run()
...
+        regions_layout.old_regions_mutex, boost::interprocess::defer_lock);
+
+    if (max_wait > 0)
     {
-        boost::interprocess::scoped_lock<boost::interprocess::named_mutex> pending_lock(
-            barrier.pending_update_mutex);
+        if (!regions_lock.timed_lock(boost::posix_time::microsec_clock::universal_time() +
+                                     boost::posix_time::seconds(max_wait)))
+        {
+            util::SimpleLogger().Write(logWARNING) << "Queries did not finish in " << max_wait
```
danpat: Add a comment here about the consequences to the downstream reader processes if this happens.

TheMarex:
Ticketed the broken windows builds https://github.com/Project-OSRM/osrm-backend/issues/3088 The error is not obvious and I can't investigate.

springmeyer:
For others easy reference the windows error is https://gist.github.com/springmeyer/33be8237a6c530ef2b45faf5d67491d8 (https://ci.appveyor.com/project/TheMarex/osrm-backend/build/1.0.3553#L696)

Issue #3016, opened 2016-10-07
osrm-datastore updates shared memory while queries are running


TheMarex:
[This code](https://github.com/Project-OSRM/osrm-backend/blob/master/src/storage/storage.cpp#L741-L748) is supposed to ensure that we never swap the dataset while queries are running, because that would cause bad race conditions.

However since the member `barrier.number_of_queries` is _process local_ and this always 0 this code will never execute. This means we happily update the datasets even if queries are executed on it.

This will cause race conditions if you run osrm-datastore twice while a queries is executed. (because OSRM datastore will assume that the shared memory region that is currently in use, is actually free and start to write to it)

MoKob:
This issue was solved, right?

TheMarex:
Yes fixed by e82f678.

PR #3017, opened 2016-10-07
Use a shared (!) reader writer lock to protect CURRENT_REGIONS

TheMarex:
# Issue

This fixes issue #3016 and avoids race conditions/crashes. You can verify this by trying to run osrm-datastore during a long-running query. I did some manual concurrency testing, it seems to work now, even with multiple runs of osrm-datastore at the same time.

## Tasklist
 - [x] review
 - [x] adjust for for comments

## Requirements / Relations

This fixes issue #3016.

### DIFF
--- a/include/engine/engine.hpp
+++ b/include/engine/engine.hpp
@@ -1,7 +1,6 @@
 #ifndef ENGINE_HPP
 #define ENGINE_HPP
 
-#include "storage/shared_barriers.hpp"
 #include "engine/status.hpp"
 #include "util/json_container.hpp"
 
@@ -18,6 +17,11 @@ namespace json
 {
 struct Object;
 }
+}
+
+namespace storage
+{
+struct SharedBarriers;
 }
 
 // Fwd decls
@@ -52,9 +56,6 @@ class BaseDataFacade;
 class Engine final
 {
   public:
-    // Needs to be public
-    struct EngineLock;
-
     explicit Engine(const EngineConfig &config);
 
     Engine(Engine &&) noexcept;
@@ -71,7 +72,7 @@ class Engine final
     Status Tile(const api::TileParameters &parameters, std::string &result) const;
 
   private:
-    std::unique_ptr<EngineLock> lock;
+    std::unique_ptr<storage::SharedBarriers> lock;
 
     std::unique_ptr<plugins::ViaRoutePlugin> route_plugin;
     std::unique_ptr<plugins::TablePlugin> table_plugin;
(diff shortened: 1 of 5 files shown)


include/engine/engine.hpp
```diff
@@ -71,7 +72,7 @@ class Engine final
     Status Tile(const api::TileParameters &parameters, std::string &result) const;
 
   private:
-    std::unique_ptr<EngineLock> lock;
+    mutable std::unique_ptr<storage::SharedBarriers> lock;
```
daniel-j-h: Semantically the `mutable` has to be on the barrier and not the pointer, but I don't think we need the mutable here at all. The reason for this is a `const Engine` is not able to modify the pointer but _is_ able to modify the barrier through the `const` pointer.

Also the ctor doing a make_unique can already be done in the ctor initializer list as in `Engine() : lock{ shared_mem ? make_unique() : null } { }`. Then even the unique_ptr could be `const`.

src/engine/engine.cpp
```diff
@@ -97,7 +46,7 @@ RunQuery(const std::unique_ptr<osrm::engine::Engine::EngineLock> &lock,
     }
 
     BOOST_ASSERT(lock);
-    lock->IncreaseQueryCount();
+    boost::interprocess::sharable_lock<boost::interprocess::named_sharable_mutex> query_lock(lock->query_mutex);
```
daniel-j-h: I'd love to have some comments here to why this is needed and what it does.

src/storage/storage.cpp
```diff
@@ -738,20 +738,15 @@ int Storage::Run()
...
-    data_timestamp_ptr->layout = layout_region;
-    data_timestamp_ptr->data = data_region;
-    data_timestamp_ptr->timestamp += 1;
-    deleteRegion(previous_data_region);
-    deleteRegion(previous_layout_region);
+        data_timestamp_ptr->layout = layout_region;
+        data_timestamp_ptr->data = data_region;
+        data_timestamp_ptr->timestamp += 1;
+        deleteRegion(previous_data_region);
+        deleteRegion(previous_layout_region);
```
daniel-j-h: Hm just to make sure: can `deleteRegion` ever throw an exception? Is it possible we're deleting one but not the other? What happens if deleting a region fails?

TheMarex: If there is no region it is just a NOP. There is an explicit check in the function if the region exists.

TheMarex:
@daniel-j-h please confirm if this works for you now. Waiting on travis and then merging.

Issue #3018, opened 2016-10-07
Build Error in ubuntu 14.04


ankitsoni9989:
I am building OSRM on my local machine, by the steps:

> mkdir -p build
> cd build
> cmake .. -DCMAKE_BUILD_TYPE=Release
> cmake --build .
> sudo cmake --build . --target install

but in the 3rd one 

> cmake .. -DCMAKE_BUILD_TYPE=Release

 I am getting an error:

> Could NOT find Lua: Found unsuitable version "5.2.3", but required is exact version "5.2" (found /usr/lib/x86_64-linux-gnu/liblua5.2.so;/usr/lib/x86_64-linux-gnu/libm.so)
> -- Could NOT find Lua: Found unsuitable version "5.2.3", but required is exact version "5.1" (found /usr/lib/x86_64-linux-gnu/liblua5.2.so;/usr/lib/x86_64-linux-gnu/libm.so)
> CMake Error at cmake/FindLuabind.cmake:15 (MESSAGE):
>   Lua 5.1 or 5.2 was not found.
> Call Stack (most recent call first):
>   CMakeLists.txt:308 (find_package)

I tried Installing libluabind-dev but that is not working

Please suggest me where I am going wrong.

Already have this:
```
 dpkg --get-selections | grep lua
liblua5.1-0:amd64				install
liblua5.1-0-dev:amd64				install
liblua5.2-0:amd64				install
liblua5.2-dev:amd64				install
libluabind-dev					install
libluabind0.9.1					install
libluajit-5.1-common				install
lua5.1						install
luajit						install
```

Thank you

daniel-j-h:
Thanks for reporting, looks like this commit from two days ago is making the Lua check too strict:

https://github.com/Project-OSRM/osrm-backend/pull/3011

I just created a pull request over at:

https://github.com/Project-OSRM/osrm-backend/pull/3020

can you try if this works for you by checking out the `fix-strict-lua-check` branch?

ankitsoni9989:
thanks @daniel-j-h that worked for 
`cmake .. -DCMAKE_BUILD_TYPE=Release`

but again in 
`cmake --build .`
I am getting error as:
```
make[2]: *** [CMakeFiles/ENGINE.dir/src/engine/engine.cpp.o] Error 1
make[1]: *** [CMakeFiles/ENGINE.dir/all] Error 2
make: *** [all] Error 2
```
Please let me know if I am missing something.
Thanks

[...]

PR #3020, opened 2016-10-08
Changes Lua detection to be less strict

daniel-j-h:
In https://github.com/Project-OSRM/osrm-backend/issues/3018 a user reported us being too strict in looking for Lua. Seems like like CMake's `EXACT` on _some_ (?) CMake versions means: be really strict about patch versions, too. I can't reproduce it locally with CMake 3.6, though: it's happily picking up my Lua 5.2.3.

@TheMarex is there a reason you used the `EXACT` keyword here?

Note: if leaving this keyword out will pick up Lua 5.3 we should manually check `LUA_VERSION_MAJOR` and `LUA_VERSION_MINOR`.

### DIFF
--- a/cmake/FindLuabind.cmake
+++ b/cmake/FindLuabind.cmake
@@ -4,6 +4,8 @@
 #  LUABIND_LIBRARIES
 #  LUABIND_INCLUDE_DIR, where to find luabind.hpp
 
+# First we try using EXACT but in some verison of
+# cmake this would also match patch versions
 FIND_PACKAGE(Lua 5.2 EXACT)
 IF (LUA_FOUND)
     MESSAGE(STATUS "Using Lua ${LUA_VERSION_STRING}")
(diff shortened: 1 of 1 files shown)


TheMarex:
@daniel-j-h if you omit `EXACT` it will pick anything `>=MAJOR.MINOR`, that is both checks would result in 5.3 being picked up. The major problem here is that you can't force it to ignore 5.3 even though 5.2 and 5.1 are available on the system.

[The docs](https://cmake.org/cmake/help/v3.0/command/find_package.html) say that the semantics of EXACT are depending on the specific package.

What about first trying to find a version with `EXACT` and if that fails downgrade to `find_package` without `EXACT`

TheMarex:
Since this is blocking the demo server I quickly did a fix https://github.com/mapbox/api-osrm-routed/issues/56

daniel-j-h:
Thanks for jumping in here and quickly fixing this!

PR #3022, opened 2016-10-08
Route over alternating oneways but not reversible ones, closes #2837.

### LINKED ISSUE #2837: oneway=reversible
Hej,

OSRM seems not to route through a oneway=reversible path, which was used (in my opinion) correctly at a construction site where traffic from either direction is passing through the construction area as oneway, while the other direction is stopped by traffic light. The example i found is here: https://www.openstreetmap.org/directions?engine=osrm_car&route=59.78790%2C17.66058%3B59.78684%2C17.66430#map=13/59.8161/17.6707 (the bridge in the middle can be used but one direction at a time, changing every ~5 minutes). Mapzen is routing the direct way...

Im not sure if this is the right issue tracker for a (probably) routing profile related bug. If not please point me to the right one...

Cheers,
Pascal

daniel-j-h:
For #2837.

- alternating: high frequency, route over them with penalty
- reversible: low frequency, do not route over them - time dependence

- http://wiki.openstreetmap.org/wiki/Tag:oneway%3Dreversible
- http://wiki.openstreetmap.org/wiki/Tag:oneway%3Dalternating

This distinction was made at the Elbe-Labe Meetup in Dresden, with
accompanying Wiki pages and tagging scheme. Thanks all involed!

- https://github.com/Project-OSRM/osrm-backend/issues/2837
- http://wiki.openstreetmap.org/wiki/Key:oneway

### DIFF
--- a/features/car/oneway.feature
+++ b/features/car/oneway.feature
@@ -85,3 +85,12 @@ Feature: Car - Oneway streets
         When I route I should get
             | from | to | route    |
             | a    | c  | ab,bc,bc |
+
+
+    # Reversible oneways (low frequency) vs alternating oneways (high frequency).
+    # See: https://github.com/Project-OSRM/osrm-backend/issues/2837
+    Scenario: Car - Route over alternating but not reversible oneways
+        Then routability should be
+            | highway | oneway      | forw | backw |
+            | primary | reversible  |      |       |
+            | primary | alternating | x    | x     |
--- a/features/car/speed.feature
+++ b/features/car/speed.feature
@@ -22,3 +22,16 @@ Feature: Car - speeds
             | residential    | no     | 31 km/h      |
             | living_street  | no     | 18 km/h      |
             | service        | no     | 23 km/h      |
+
+    # Alternating oneways have to take average waiting time into account.
+    Scenario: Car - scaled speeds for oneway=alternating
+        Then routability should be
+            | highway        | oneway      | junction   | bothw        | #              |
+            | tertiary       |             |            | 43 km/h      |                |
+            | tertiary       | alternating |            | 20 km/h +- 5 |                |
+            | motorway       |             |            | 82 km/h      | implied oneway |
+            | motorway       | alternating |            | 30 km/h +- 5 | implied oneway |
+            | motorway       | reversible  |            |              | unroutable     |
+            | primary        |             | roundabout | 63 km/h      | implied oneway |
+            | primary        | alternating | roundabout | 25 km/h +- 5 | implied oneway |
+            | primary        | reversible  | roundabout |              | unroutable     |
--- a/profiles/car.lua
+++ b/profiles/car.lua
@@ -308,7 +308,9 @@ function way_function (way, result)
     return
   end
 
-  -- check if oneway tag is unsupported
+  -- Reversible oneways change direction with low frequency (think twice a day):
+  -- do not route over these at all at the moment because of time dependence.
+  -- Note: alternating (high frequency) oneways are handled below with penalty.
   local oneway = way:get_value_by_key("oneway")
   if oneway and "reversible" == oneway then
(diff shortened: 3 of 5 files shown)


features/car/speed.feature
```diff
@@ -22,3 +22,10 @@ Feature: Car - speeds
             | residential    | no     | 31 km/h      |
             | living_street  | no     | 18 km/h      |
             | service        | no     | 23 km/h      |
+
+    # Alternating oneways have to take average waiting time into account.
+    Scenario: Car - scaled speeds for oneway=alternating
+        Then routability should be
+            | highway        | oneway      | bothw         |
+            | tertiary       | no          | 43 km/h       |
+            | tertiary       | alternating | 20 km/h +- 10 |
```
MoKob: why this high difference? If we assume some specific speed, I would say 10 is a bit large of a variance (50% here).

profiles/car.lua
```diff
@@ -584,6 +586,14 @@ function way_function (way, result)
     result.backward_speed = math.min(penalized_speed, scaled_speed)
   end
 
+  -- Handle high frequency reversible oneways (think traffic signal controlled, changing direction every 15 minutes).
+  -- Scaling speed to take average waiting time into account plus some more for start / stop.
+  if oneway and "alternating" == oneway then
+    local scaling_factor = 0.4
```
MoKob: should this check for valid speeds? If the speed is explicitly set to `math.huge`, scaling here could have unwanted side-effects. Right?

features/car/oneway.feature
```diff
@@ -81,3 +81,12 @@ Feature: Car - Oneway streets
...
             | a    | c  | ab,bc,bc |
+
+
+    # Reversible oneways (low frequency) vs alternating oneways (high frequency).
+    # See: https://github.com/Project-OSRM/osrm-backend/issues/2837
+    Scenario: Car - Route over alternating but not reversible oneways
+        Then routability should be
+            | highway | oneway      | forw | backw |
+            | primary | reversible  |      |       |
+            | primary | alternating | x    | x     |
```
MoKob: can this tag occur with some other items that `imply` oneway? E.g. what about `motorway` or `roundabouts`?

Not that I would think them reasonable, but I would hate to fail horribly if they should occur in the *wild*

daniel-j-h:
Blocked by working speed tests: https://github.com/Project-OSRM/osrm-backend/issues/3036#issuecomment-252872862

features/car/speed.feature
daniel-j-h: :+1:

profiles/car.lua
daniel-j-h: all arithmetic with `math.huge` will return `math.huge` - but I will add a check to make it explicit

features/car/oneway.feature
daniel-j-h: What should the behavior be for implied oneways with `oneway=alternating`? I would say alternating overwrites the implied oneway then, since it's an explicit tag you have to specify.

daniel-j-h:
I manually logged the maxspeeds for the Cucumber scenarios (until https://github.com/Project-OSRM/osrm-backend/pull/3039 lands): looks good.

MoKob:
In general, this PR looks fine with me. As stated before, I'd love if we could add some checks with `oneway = reversible/alternating` for roundabouts. But thats kind of a bonus.

Feel free to squash/merge when you feel like it.

daniel-j-h:
@MoKob here are some implied oneways scenarios (including roundabouts) https://github.com/Project-OSRM/osrm-backend/pull/3022/files#diff-3757e1244d7c5310cc328ebe44f94e83R32

MoKob:
I should probably specify. I was concerned about the instructions we will see :). If its still works for counting/not segfaulting.

PR #3029, opened 2016-10-10
Catch boost bad_numeric_cast exception and let parser return an error

### LINKED ISSUE #3007: osrm-routed should validate input coordinates better
http://router.project-osrm.org/route/v1/driving/-74.735667,5.207041;-74.721057,5.202046;-74697224,5.191564;-74.59997,5.069609?overview=false&alternatives=true&steps=true&hints=;;;

> [server error] code: bad numeric conversion: negative overflow

TheMarex:
# Issue

Fixes #3007 

## Tasklist
 - [x] review
 - [x] adjust for for comments

## Requirements / Relations

### DIFF
--- a/src/server/api/parameters_parser.cpp
+++ b/src/server/api/parameters_parser.cpp
@@ -55,6 +55,11 @@ boost::optional<ParameterT> parseParameters(std::string::iterator &iter,
         // iterator to the failing position. Extract the position from the exception ourselves.
         iter = failure.first;
     }
+    catch (const boost::numeric::bad_numeric_cast &e)
+    {
+        // this can happen if we get bad numeric values in the request, just handle
+        // as normal parser error
+    }
 
     return boost::none;
 }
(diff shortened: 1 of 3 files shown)


src/server/api/parameters_parser.cpp
```diff
@@ -55,6 +55,11 @@ boost::optional<ParameterT> parseParameters(std::string::iterator &iter,
         // iterator to the failing position. Extract the position from the exception ourselves.
         iter = failure.first;
     }
+    catch (const boost::numeric::bad_numeric_cast &e)
+    {
+        // this can happen if we get bad numeric values in the request, just handle
+        // as normal parser error
+    }
```
daniel-j-h: How does this make sure invalid coordinates making the parser fail? Relying on the Coordinate type's boost numeric_cast calls in the ctor? Shouldn't we rely on a coordinate's `.IsValid()` member function instead?

TheMarex: Invalid coordinates are caught at a higher level. This only fixes the case where the actual numeric value does not fit in the data type. I added another test there to demonstrate this.

daniel-j-h: :+1:

PR #3031, opened 2016-10-10
Enable use of other build types than Debug and Release

danpat:
# Issue

The `CMakeLists.txt` file was not allowing build types other than `Debug` and `Release`.  Other built-in CMake types like `RelWithDebInfo` and `MinRelSize` were effectively disabled.

In addition, passing `-DCMAKE_BUILD_TYPE=RelWithDebInfo` was silently being overwritten by `Release`.

This change maintains the existing `cmake ..` messages, handles the additional build types specified in the CMake docs, and prints a small warning if an unrecognized build type is supplied.

Existing uses: `cmake ..` and `cmake -DCMAKE_BUILD_TYPE=Debug ..` are unchanged.  New behaviour can be seen by supplying `cmake -DCMAKE_BUILD_TYPE=RelWithDebInfo ..` or `cmake -DCMAKE_BUILD_TYPE=FOOBAR ..`

## Tasklist
 - [x] review
 - [x] adjust for for comments

## Requirements / Relations
Reference: https://github.com/Homebrew/homebrew-core/pull/5660#pullrequestreview-3410503

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -99,24 +99,6 @@ add_library(osrm_extract $<TARGET_OBJECTS:EXTRACTOR> $<TARGET_OBJECTS:UTIL>)
 add_library(osrm_contract $<TARGET_OBJECTS:CONTRACTOR> $<TARGET_OBJECTS:UTIL>)
 add_library(osrm_store $<TARGET_OBJECTS:STORAGE> $<TARGET_OBJECTS:UTIL>)
 
-# Check the release mode
-if(NOT CMAKE_BUILD_TYPE MATCHES Debug)
-  set(CMAKE_BUILD_TYPE Release)
-endif()
-if(CMAKE_BUILD_TYPE MATCHES Debug)
-  message(STATUS "Configuring OSRM in debug mode")
-  set(ENABLE_ASSERTIONS ON)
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -136,8 +118,45 @@ if(ENABLE_GOLD_LINKER)
...
+  message(STATUS "Configuring OSRM in debug mode")
+elseif(CMAKE_BUILD_TYPE MATCHES Release)
   message(STATUS "Configuring OSRM in release mode")
+elseif(CMAKE_BUILD_TYPE MATCHES RelWithDebInfo)
+  message(STATUS "Configuring OSRM in release mode with debug flags")
+elseif(CMAKE_BUILD_TYPE MATCHES MinRelSize)
+  message(STATUS "Configuring OSRM in release mode with minimized size")
+else()
+  message(STATUS "Unrecognized build type - will use cmake defaults")
+endif()
```
daniel-j-h:
```
if not BUILD_TYPE
  message(unrecognized)
else
  message(using BUILD_TYPE)
```
?

danpat: Users can still do something like `-DCMAKE_BUILD_TYPE=FOO`, which would define CMAKE_BUILD_TYPE, but we don't have any special handling for.  The IF statement is structured to warn when that happens.

Issue #3032, opened 2016-10-10
Emitting`invalid` using bicycle profile


karenzshea:
I'm seeing some cases of the `invalid` turn type being returned, but I'm only seeing this while using the bicycle.lua profile.

DC example
- request `http://localhost:5000/route/v1/cycling/-77.05127120018005,38.94556697963155;-77.04606771469116,38.942512906501065?steps=true`
- response https://gist.github.com/karenzshea/39d93267ac641e216f7a67fe43194ba9#file-dc-invalid-json-L210
- segment in question http://www.openstreetmap.org/way/6053420

München example
- request: `http://localhost:5000/route/v1/cycling/11.549114,48.143612;11.552596092224121,48.13910076713174?steps=true`
- response https://gist.github.com/karenzshea/5c79347872c173fd67f6f5b6e1dcdaee#file-mu-invalid-json-L82
- segment in question http://www.openstreetmap.org/way/71822746#map=18/48.14428/11.54930&layers=C

It seems like in both of these cases and a few others I haven't documented as thoroughly, that we're routing through invalid ways and just using a mode change to `push bike`.

@MoKob have time to take a look?

PR #3034, opened 2016-10-11
don't emit invalid on sliproads with incompatible modes

MoKob:
# Issue

resolves https://github.com/Project-OSRM/osrm-backend/issues/3032

## Tasklist
 - [x] review
 - [ ] adjust for for comments

## Requirements / Relations
none

### DIFF
--- a/features/guidance/bicycle-sliproads.feature
+++ b/features/guidance/bicycle-sliproads.feature
@@ -1,0 +1,33 @@
+@routing @bicycle @mode
+Feature: Bike - Mode flag
+
+	Background:
+		Given the profile "bicycle"
+        Given a grid size of 5 meters
+
+    Scenario: Bike Sliproad
+        Given the node map
+            """
+                  i
+            a b - c-d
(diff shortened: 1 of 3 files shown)


features/guidance/bicycle-sliproads.feature
```diff
@@ -0,0 +1,33 @@
...
+        And the ways
+            | nodes | highway   | name   | oneway:bicycle |
+            | abcd  | cycleway  | street |                |
+            | eb    | path      |        | yes            |
+            | icef  | tertiary  | road   |                |
+            | geh   | secondary | street |                |
+
+        When I route I should get
+            | waypoints | route                           | turns                               |
+            | a,f       | street,{highway:path},road,road | depart,turn right,turn right,arrive |
```
daniel-j-h: Unrelated but I'm just seeing this here: is there a reason for us to still encode the name with the highway tag in _the default profile_? If people want to do this, fine, feel free to change the profile.

Issue #3038, opened 2016-10-11
Different durations returned for reverse and forward routes


karenzshea:
I noticed this while logging cucumber tests for https://github.com/Project-OSRM/osrm-backend/issues/3036

Routing on a test case where a way is set to
- one road class type
- not a one way

Expected outcome is that a route between two ponits on the way in either direction should return the same duration, but this is not the case. All the cases I've seen so far only differ by `0.1s`.

Specifically this test, [speed.feature](https://github.com/Project-OSRM/osrm-backend/blob/6cf99c886fcb42c4e417f5477ed2c9d2eebb5554/features/car/speed.feature)

karenzshea:
Example logged from cucumber (I haven't been able to produce on an real world route yet). Specifically note that time is pulled directly from the OSRM routed response in the internal cucumber call.

```
{ forw: 
   { query: 'http://127.0.0.1:5000/route/v1/car/10.898737978337776,1.0;10.916719337063185,1.0?steps=true&alternatives=false',
     json: { code: 'Ok', routes: [Object], waypoints: [Object] },
     status: 'x',
     route: 'w11,w11',
     summary: 'w11',
     time: 232.3,
     distance: 1999.8,
     speed: 30 },
  backw: 
   { query: 'http://127.0.0.1:5000/route/v1/car/10.916719337063185,1.0;10.898737978337776,1.0?steps=true&alternatives=false',
     json: { code: 'Ok', routes: [Object], waypoints: [Object] },
     status: 'x',
     route: 'w11,w11',
     summary: 'w11',
     time: 232.2,
     distance: 1999.8,
     speed: 31 },
  bothw: { status: 'x', time: 'diff', distance: 1999.8 } }
```

daniel-j-h:
Just to be sure here: please copy the .osm.xml cucumber generates for these tests (check the tests/cache directory), run osrm-extract, contract, routed on it and manually verify those values are already wrong in osrm.

While the distance is the same, it looks like there is a tiny difference in time - could be rounding or floating point issue.

karenzshea:
Confirmed that the above also reproduces
```
{
	code: "Ok",
	routes: [{
		legs: [{
			steps: [],
			summary: "",
			duration: 232.3,
			distance: 1999.8
		}],
		geometry: "_ibEctoaA?koB",
		duration: 232.3,
		distance: 1999.8
	}],
	waypoints: [{
[...]

Issue #3036, opened 2016-10-11
Cucumber Routability scenarios testing for speeds are broken


daniel-j-h:
In https://github.com/Project-OSRM/osrm-backend/pull/3022 I implemented alternating oneways influencing speeds. For testing this I added some routability scenarios to [features/car/speed.feature](https://github.com/Project-OSRM/osrm-backend/blob/6cf99c886fcb42c4e417f5477ed2c9d2eebb5554/features/car/speed.feature#L10-L24) with lower speeds.

Playing with these routability checks, it looks like they're broken. The tests pass even if you change speeds to e.g. 0 or 9000.

For the speed comparison [here](https://github.com/Project-OSRM/osrm-backend/blob/master/features/step_definitions/routability.js#L39-L41) printing `outputRow[direction]` and `want` reveals they're the same: we're comparing the user's input with itself and not with what we get back from osrm.

daniel-j-h:
Note: this is not specific to `bothw` speeds, all our speed tests are broken.

MoKob:
The speed does not define whether we can or cannot route on an edge. Even on speed zero or 9000, you are simply very quick or stuck on that edge. We set edge cost to a minimum of 1, so you will never be reaching infinite speed.

Zero itself is assumed to be an invalid speed, though (https://github.com/Project-OSRM/osrm-backend/blob/master/src/extractor/extraction_containers.cpp#L371), since we need to be able to do https://github.com/Project-OSRM/osrm-backend/blob/master/src/extractor/extraction_containers.cpp#L392.

I don't think that setting a speed to zero should be invalid, but right now it is. We shouldn't disable the possibility to route over these, though. We should just take really long.

daniel-j-h:
All our tests are using the `forw`, `backw`, `bothw` columns with speed values, see e.g. the [maxspeed.feature](https://github.com/Project-OSRM/osrm-backend/blob/f456d7ff86f60dc44a1742d74b57db86f50374ef/features/car/maxspeed.feature#L54-L62) which leaves those tests effectively as broken.

daniel-j-h:
@karenzshea can you take a look - this is fixed now, isn't it? (if so feel free to close it)

karenzshea:
May this stay fixed, https://github.com/Project-OSRM/osrm-backend/pull/3039

PR #3039, opened 2016-10-11
Fix cucumber routability tests comparing speeds

karenzshea:
# Issue

https://github.com/Project-OSRM/osrm-backend/issues/3036

Any routability tests relying on `forw/backw/bothw` fields to be able to diff values given as speeds in `km/h` or times in `s` weren't working. We weren't updating the outputRow (the field that is later FuzzyMatched) to the value in the result. 

Test failures ostensibly look like they are because the speed values differ even with `+-` qualifications but it's actually because of hidden differences in response diffs. The issue I've uncovered here is with inconsistent times being returned sometimes on routes that should have the same duration forwards and backwards, thus the table gets marked as erroring because time and speed fields are different. https://github.com/Project-OSRM/osrm-backend/issues/3038.

## Tasklist
 - [x] review
 - [ ] adjust for for comments

### DIFF
--- a/features/car/service.feature
+++ b/features/car/service.feature
@@ -4,11 +4,10 @@ Feature: Car - Surfaces
     Background:
         Given the profile "car"
 
-    @todo
     Scenario: Car - Surface should reduce speed
         Then routability should be
             | highway  | service           | forw       | backw       |
             | service  | alley             | 5 km/h +-1 | 5 km/h +-1  |
             | service  | emergency_access  |            |             |
-            | service  | driveway          | 15 km/h +-1| 15 km/h +-1 |
+            | service  | driveway          | 5 km/h +-1 | 5 km/h +-1  |
 
--- a/profiles/car.lua
+++ b/profiles/car.lua
@@ -42,7 +42,8 @@ speed_profile = {
 -- service speeds
 service_speeds = {
   ["alley"] = 5,
-  ["parking_aisle"] = 5
+  ["parking_aisle"] = 5,
+  ["driveway"] = 5
 }
 
 -- surface/trackype/smoothness
--- a/taginfo.json
+++ b/taginfo.json
@@ -21,6 +21,11 @@
         },
         {
             "key": "service",
+            "value": "driveway",
+            "object_types": [ "way" ]
+        },
+        {
+            "key": "service",
             "value": "alley",
             "object_types": [ "way" ]
         },


features/step_definitions/routability.js
```diff
@@ -29,8 +29,13 @@ module.exports = function () {
                                     result[direction].status.toString() : '';
                                 break;
                             case /^\d+s/.test(want):
+                                outputRow[direction] = result[direction].time ?
+                                    result[direction].time.toString()+'s' : '';
                                 break;
                             case /^\d+ km\/h/.test(want):
+                                outputRow[direction] = !isNaN(result[direction].speed) ?
+                                    result[direction].speed.toString()+' km/h' :
+                                    result[direction].speed.toString() || '';
```
daniel-j-h: When can can the speed be NaN here? Why are we appending km/h, I thought the fuzzy matcher was able to handle all of this?

karenzshea: speed can come back as the value `diff` in the `bothw` field, when `forw`/`backw` differ.

I was seeing that without appending `km/h` just the number is returned in the output table. I can double check that this isn't needed though.

daniel-j-h: From skimming the failing tests, most are +- 1 km/h; but I also spotted more severe issues:

```
| (-) track | (-) asphalt | (-) 48s | 
| (+) track | (+) asphalt | (+) 60s | 
```
```
| (-) service | (-) driveway | (-) 15 km/h +-1 | (-) 15 km/h +-1 | 
| (+) service | (+) driveway | (+) 23 km/h | (+) 23 km/h | 
```

once you have a small epsilon of 1 or 2 km/h in the comparison you'll probably only see those handful of tests failing. And then we have to check what's going on there.

karenzshea:
A few issues fixed here:
- uncovered the duration bug in https://github.com/Project-OSRM/osrm-backend/issues/3038
- shortcuts were not being compared properly, as the shortcut's represented value when passing fuzzyMatch was not being properly passed to row diffing
- `actual` speed and time values were never being compared to `expected` because they weren't passed to the fuzzyMatch part of row processing
- values with `+-` qualifications weren't passing correctly even when a value was precisely within range, (e.g. `5 km/h +- 1 vs 5 km/h`) because the table diffing would compare the rows literally, ignoring input from fuzzyMatch (fuzzyMatch was properly handling ranges)

There are still some problems with my fixes that I'm trying to clear up.

MoKob:
https://travis-ci.org/Project-OSRM/osrm-backend/jobs/167366082#L5803 seems a legit problem. If the surface should influence the speed, it does not seem to do so anymore.
(possibly related: https://travis-ci.org/Project-OSRM/osrm-backend/jobs/167366082#L6062)

https://travis-ci.org/Project-OSRM/osrm-backend/jobs/167366082#L5962 as well. From @TheMarex changes, this driveway should be slow but it isn't.

On https://travis-ci.org/Project-OSRM/osrm-backend/jobs/167366082#L6020 I don't know how to interpret `diff`. But it looks like it could be a serious problem.

In general it looks like the surface modificators might be broken right now.

karenzshea:
@danpat and I ticketed out a couple issues here where I had to `@todo` test failures for more investigation

https://github.com/Project-OSRM/osrm-backend/issues/3053
https://github.com/Project-OSRM/osrm-backend/issues/3052

Hopefully this starts passing for everything else; another review and this should be good to go.

daniel-j-h:
Please squash before merging.

Also https://travis-ci.org/Project-OSRM/osrm-backend/builds/167491021 there are three tests failing, from what I can see it's a mix of the `diff` issue and `15s` vs `15 s`.

daniel-j-h:
I just fixed the `15s` vs `15 s` issue, always enforcing a space before the unit.
We're down to one failing `diff` test.

daniel-j-h:
This landed in master. i accidentally first merged and then force-pushed this branch, that's why Github shows we closed the PR. Sorry for that.

PR #3042, opened 2016-10-12
Fixes missing memory include for make_unique; reported in #3018.

daniel-j-h:
In https://github.com/Project-OSRM/osrm-backend/issues/3018#issuecomment-253118518 a user reported compilation errors because we're using make_unique without the corresponding memory header.


MoKob:
Looks fine. Go ahead and merge

PR #3045, opened 2016-10-12
Add max zoom limit of z12 to tile plugin parameters

karenzshea:
# Issue

to improve debug endpoint performance on the demo server.

## Tasklist
 - [x] add unit test for failing tile request
 - [ ] review
 - [ ] adjust for for comments

## Requirements / Relations
n/a

### DIFF
--- a/unit_tests/server/parameters_parser.cpp
+++ b/unit_tests/server/parameters_parser.cpp
@@ -383,10 +383,22 @@ BOOST_AUTO_TEST_CASE(valid_nearest_urls)
     CHECK_EQUAL_RANGE(reference_2.coordinates, result_2->coordinates);
 }
 
+BOOST_AUTO_TEST_CASE(invalid_tile_urls)
+{
+    TileParameters reference_1{1, 2, 3};
+    auto result_1 = parseParameters<TileParameters>("tile(1,2,3).mvt");
+    BOOST_CHECK(result_1);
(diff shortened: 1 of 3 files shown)


unit_tests/server/parameters_parser.cpp
```diff
@@ -385,8 +385,8 @@ BOOST_AUTO_TEST_CASE(valid_nearest_urls)
 
 BOOST_AUTO_TEST_CASE(valid_tile_urls)
 {
-    TileParameters reference_1{1, 2, 3};
-    auto result_1 = parseParameters<TileParameters>("tile(1,2,3).mvt");
+    TileParameters reference_1{1, 2, 12};
+    auto result_1 = parseParameters<TileParameters>("tile(1,2,12).mvt");
```
MoKob: can we also check for smaller zoom levels returning valid error codes?

danpat:
Hmm.  This is nitpicky, but given that this PR was triggered by noticing a Denial-Of-Service on the demo server, it might be useful to have a test to ensure that z12 is enforced via `osrm-routed`.

If, for some reason, the call to `IsValid` gets removed from `tile_service.cpp` (e.g. during a future refactoring), then bad requests would be able to get through and no tests would catch it.

karenzshea:
Thanks @danpat opened a ticket here https://github.com/mapbox/api-osrm-routed/issues/54

Issue #3048, opened 2016-10-13
Service Roads can disturb obvious turn discovery


MoKob:
<img width="968" alt="screen shot 2016-10-13 at 14 44 04" src="https://cloud.githubusercontent.com/assets/12932279/19349505/a0e1d86e-9153-11e6-9926-5e9c3da8737e.png">

When comparing the turns next to a possible obvious turn, service roads can influence the result in a negative way, resulting in the above setting.

PR #3049, opened 2016-10-13
don't get disturbed by service roads in obvious discovery

MoKob:
# Issue

resolves https://github.com/Project-OSRM/osrm-backend/issues/3048

## Tasklist
 - [x] review
 - [ ] adjust for for comments

## Requirements / Relations
none

### DIFF
--- a/features/guidance/turn.feature
+++ b/features/guidance/turn.feature
@@ -1252,3 +1252,40 @@ Feature: Simple Turns
             | waypoints | turns                          | route     |
             | a,f       | depart,turn slight left,arrive | road,,    |
             | a,c       | depart,arrive                  | road,road |
+
+    Scenario: Non-Obvious Turn Next to service road
+        Given the node map
+            """
+                                c
+                               .
+                               .
+                               .
+                               .
+                               .
+                             .
+                             .
+                             .
+                             .
+                             .
+            a - - - - - - - b - - - d
+                            |
+                            |
+                            |
+                            |
+                            |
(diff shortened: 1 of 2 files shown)


src/extractor/guidance/intersection_handler.cpp
```diff
@@ -510,17 +510,44 @@ std::size_t IntersectionHandler::findObviousTurn(const EdgeID via_edge,
...
+            if (obvious_by_road_class(in_data.road_classification,
+                                      best_data.road_classification,
+                                      candidate_data.road_classification))
+                return (index_candidate + 1) % intersection.size();
+            else
+                return index_candidate;
+
+        }();
+        const auto right_index = [&]() {
+            const auto index_candidate = best - 1;
```
TheMarex: For documentation purposes an assertion that `best > 0` could help.

features/guidance/turn.feature
```diff
@@ -1252,3 +1252,40 @@ Feature: Simple Turns
...
+        And the ways
+            | nodes  | highway | name    |
+            | ab     | primary | in      |
+            | bc     | primary | through |
+            | be     | primary | through |
+            | bd     | service |         |
+
+       When I route I should get
+            | waypoints | route              | turns                   |
+            | a,c       | in,through,through | depart,turn left,arrive |
```
daniel-j-h: Ah nice, I think there are two similar scenarios in your turn angle branch.

PR #3056, opened 2016-10-14
Asserts on valid v5 conforming maneuver types, resolves #3035

### LINKED ISSUE #3035: Log Invalid Response Types
Breaking out from

- https://github.com/Project-OSRM/osrm-backend/pull/3034
- https://github.com/Project-OSRM/osrm-backend/issues/3032

invalid instructions do not conform to the v5 API and should in theory never occur. We should log those with a fixed log message we can filter on for automatic detection.

daniel-j-h:
For https://github.com/Project-OSRM/osrm-backend/issues/3035. At the moment we have no idea when we're accidentally returning an "invalid" type. This type is not in the v5 - let's get a notification for when this happens (=> bug).

## Tasklist
 - [x] review
 - [ ] adjust for for comments

### DIFF
--- a/src/engine/api/json_factory.cpp
+++ b/src/engine/api/json_factory.cpp
@@ -175,10 +175,18 @@ std::string modeToString(const extractor::TravelMode mode)
 util::json::Object makeStepManeuver(const guidance::StepManeuver &maneuver)
 {
     util::json::Object step_maneuver;
+
+    std::string maneuver_type;
+
     if (maneuver.waypoint_type == guidance::WaypointType::None)
-        step_maneuver.values["type"] = detail::instructionTypeToString(maneuver.instruction.type);
+        maneuver_type = detail::instructionTypeToString(maneuver.instruction.type);
     else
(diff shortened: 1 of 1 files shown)


src/engine/api/json_factory.cpp
```diff
@@ -175,10 +175,20 @@ std::string modeToString(const extractor::TravelMode mode)
...
     if (maneuver.waypoint_type == guidance::WaypointType::None)
-        step_maneuver.values["type"] = detail::instructionTypeToString(maneuver.instruction.type);
+        maneuver_type = detail::instructionTypeToString(maneuver.instruction.type);
     else
-        step_maneuver.values["type"] = detail::waypointTypeToString(maneuver.waypoint_type);
+        maneuver_type = detail::waypointTypeToString(maneuver.waypoint_type);
+
+    // These invalid responses should never happen: log if they do happen
+    const auto invalid_type = maneuver_type == "invalid";
+    BOOST_ASSERT_MSG(!invalid_type, "unexpected invalid maneuver type");
```
TheMarex: Why not just `maneuver_type == "invalid"` in the assertion?

PR #3058, opened 2016-10-14
Collapses Double OSRM <-> Engine <-> .. PImpl Indirection, Resolves #3019

### LINKED ISSUE #3019: Needless implementation hiding in osrm::Engine object
Currently we use a lot of patterns (like storing everything in unique_ptrs) in the `osrm::Engine` object to hide the implementation of members. This would only be necessary if we actually care to expose that object externally. However the `osrm::OSRM` object already hides away the `Engine` internals to every user so we don't need them.

daniel-j-h:
For https://github.com/Project-OSRM/osrm-backend/issues/3019. We already have a PImpl abstraction with `OSRM`, no need for another PImpl with `Engine`. One level of indirection collapsed; `mutable` pushed down to search heaps and algorithms (since they're threadsafe).

## Tasklist
 - [x] review
 - [x] adjust for for comments

### DIFF
--- a/src/engine/engine.cpp
+++ b/src/engine/engine.cpp
@@ -1,17 +1,8 @@
-#include "engine/api/route_parameters.hpp"
-#include "engine/data_watchdog.hpp"
 #include "engine/engine.hpp"
+#include "engine/api/route_parameters.hpp"
 #include "engine/engine_config.hpp"
 #include "engine/status.hpp"
 
-#include "engine/plugins/match.hpp"
-#include "engine/plugins/nearest.hpp"
-#include "engine/plugins/table.hpp"
-#include "engine/plugins/tile.hpp"
-#include "engine/plugins/trip.hpp"
-#include "engine/plugins/viaroute.hpp"
-
-#include "engine/datafacade/datafacade_base.hpp"
 #include "engine/datafacade/internal_datafacade.hpp"
 #include "engine/datafacade/shared_datafacade.hpp"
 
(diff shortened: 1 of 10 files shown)


TheMarex:
Looks good to me. Needs a rebase.

daniel-j-h:
Done.

src/engine/engine.cpp
```diff
@@ -60,9 +52,26 @@ namespace osrm
...
 
+// Pointer-to-Impl. hiding target, disallow all other mis-uses
+
+// Not Moveable
+Engine::Engine(Engine &&) noexcept = delete;
+Engine &Engine::operator=(Engine &&) noexcept = delete;
+
+// Not Copyable
+Engine::Engine(const Engine &) = delete;
+Engine &Engine::operator=(const Engine &) = delete;
```
daniel-j-h: These needs to be expanded in the impl. file so the compiler knows about the type's sizes.

See Item 22 in Effective Modern C++ for detailed explanation.

TheMarex:
@daniel-j-h the build is failing with some compile errors:

https://travis-ci.org/Project-OSRM/osrm-backend/jobs/168780367#L1384-L1411

daniel-j-h:
Seems to work on my gcc 4.9 but not on recent compilers (see Travis). Investigating.

daniel-j-h:
Found the issue

http://eel.is/c++draft/dcl.fct.def.delete#4

reworking to make it standard conforming ...

Issue #3043, opened 2016-10-12
Rip out custom make_unique; replace with stdlib equivalent


daniel-j-h:
In https://github.com/Project-OSRM/osrm-backend/issues/3018#issuecomment-253118518 a user reported issues with `make_unique` (see also https://github.com/Project-OSRM/osrm-backend/pull/3042).

I had a look and it seems like we're even providing our own (and not fully conforming) `make_unique` implementation here:

https://github.com/Project-OSRM/osrm-backend/blob/master/include/util/make_unique.hpp

and sometimes are already using make_unique from the stdlib.

Now that we switched to C++14 we should unconditionally use the one from the stdlib instead:

http://en.cppreference.com/w/cpp/memory/unique_ptr/make_unique

PR #3059, opened 2016-10-14
remove self-written util::make_unique and use C++14 std::make_uniqe

chaupow:
# Issue
This solves issue #3043 

## Tasklist
 - [x] review
 - [x] adjust for comments


include/engine/datafacade/shared_datafacade.hpp
```diff
@@ -252,7 +251,7 @@ class SharedDataFacade final : public BaseDataFacade
             shared_memory, storage::SharedDataLayout::NAME_CHAR_LIST);
         util::ShM<char, true>::vector names_char_list(
             names_list_ptr, data_layout->num_entries[storage::SharedDataLayout::NAME_CHAR_LIST]);
-        m_name_table = util::make_unique<util::RangeTable<16, true>>(
+        m_name_table = std::make_unique<util::RangeTable<16, true>>(
```
daniel-j-h: include memory

http://en.cppreference.com/w/cpp/memory/unique_ptr/make_unique

Issue #3027, opened 2016-10-10
Directional Destination Tags


daniel-j-h:
There are roughly 10k `destination:forward` and 10k `destination:backward` tags in OSM:
- http://taginfo.openstreetmap.org/keys/destination%3Aforward
- http://taginfo.openstreetmap.org/keys/destination%3Abackward

at the moment we
- only parse `destination` and `destination:ref` tags and
- only [set a way's destination when it's a oneway](https://github.com/Project-OSRM/osrm-backend/blob/339ebe8f819309c5c7084c448b3b7d73213725bd/profiles/car.lua#L493-L510)

this limitation is fine for most of the cases since [motorways are implied oneway](http://wiki.openstreetmap.org/wiki/Tag:highway%3Dmotorway) and [motorway links](https://wiki.openstreetmap.org/wiki/Tag:highway%3Dmotorway_link) are often if not always oneways, too.

This ticket is _not_ about storing a forward and a backward destination value per way but instead taking `destination:forward` or `destination:backward` (depending on `oneway=-1`) into account.

Tasks:
- [ ] write Cucumber scenarios for combinations of `destination`, `destination:ref`, `destination:forward`, `destination:backward`, and `oneway=yes`, `oneway=no`, `oneway=-1` (see [existing scenarios](https://github.com/Project-OSRM/osrm-backend/blob/339ebe8f819309c5c7084c448b3b7d73213725bd/features/guidance/destination-signs.feature))
- [ ] in [the profile](https://github.com/Project-OSRM/osrm-backend/blob/339ebe8f819309c5c7084c448b3b7d73213725bd/profiles/car.lua#L493-L510) (or maybe in profiles/lib/destinations.lua) pick up directional destination tag and set according to oneway direction (see e.g. [directional hov handling](https://github.com/Project-OSRM/osrm-backend/blob/339ebe8f819309c5c7084c448b3b7d73213725bd/profiles/car.lua#L283-L301))

ghoshkaj:
@daniel-j-h I've been talking to @karenzshea (thanks Karen!) to try and figure out what is supposed to happen here. My understanding so far is this:

The point of this ticket is to take into account destination tags for ways that are labelled oneway=-1. So the first task is to write (a) new cucumber test, similar to destination-signs.feature, actually gives the values -1 a destination instead of ignoring it. Then the next part is to update the code in the profile to stop ignoring the destination of any oneway=-1 and instead give it the correct forward or backward destination tag. 

[...]

PR #3061, opened 2016-10-14
Directional Destination Tags

ghoshkaj:
# Issue

 #3027

## Tasklist
- [x] write Cucumber scenarios for combinations of destination, destination:ref, destination:forward, destination:backward, and oneway=yes, oneway=no, oneway=-1 (see existing scenarios)
 - [x] review
 - [x] adjust for for comments
- [x] in the profile (or maybe in profiles/lib/destinations.lua) pick up directional destination tag and set according to oneway direction (see e.g. directional hov handling)
 - [x] review
 - [x] adjust for for comments
- [x] write Cucumber scenarios for destination:ref:forward, destination:ref:backward as well
- [x] write code to address destination:ref:forward and destination:ref:backward
- [x] review
- [x] adjust for for comments

### DIFF
--- a/features/guidance/destination-signs-forward-backward.feature
+++ b/features/guidance/destination-signs-forward-backward.feature
@@ -1,0 +1,43 @@
+@routing @guidance
+Feature: Destination Signs
+
+    Background:
+        Given the profile "car"
+
+    Scenario: Car - route name assembly with destination signs with backwards and forwards
+        Given the node map
+          """
+          a b
+          c d
+          e f
+          g h
+          i j
+          k l
+          m n
+          o p
+          q r
+          """
+
+        And the ways
+          | nodes | name | ref | destination    | destination:ref | destination:forward | destination:backward | oneway | #                                    |
+          | ab    | AB   |     | Berlin         |                 |                     |                      | yes    |                                      |
+          | cd    | CD   |     |                |                 | Berlin              |                      | yes    |                                      |
+          | ef    | EF   |     |                |                 | Berlin              |  Hamburg             | -1     |                                      |
+          | gh    | GH   |     |                | A1              |                     |                      | yes    |                                      |
+          | ij    | IJ   |     | Berlin         | A1              |                     |                      | no     | mis-tagged destination: not a oneway |
+          | kl    | KL   |     |                | A1              | Berlin              |  Hamburg             | yes    |                                      |
+          | mn    | MN   |     | Berlin         | A1              | Berlin              |  Hamburg             | yes    |                                      |
+          | op    | OP   |     | Berlin         |                 |                     |  Hamburg             | -1     |                                      |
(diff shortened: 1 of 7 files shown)


features/guidance/destination-signs-forward-backward.feature
```diff
@@ -0,0 +1,43 @@
...
+          | op    | OP   |     | Berlin         |                 |                     |  Hamburg             | -1     |                                      |
+          | qr    | QR   |     | Berlin;Hamburg | A1;A2           |                     |                      | no     | mis-tagged destination: not a oneway |
+
+        When I route I should get
+          | from | to | route                                                     | destinations                                    | ref   | #                         |
+          | a    | b  | AB,AB                                                     | Berlin,Berlin                                   | ,     |                           |
+          | c    | d  | CD,CD                                                     | Berlin,Berlin                                   | ,     |                           |
+          | e    | f  | EF,EF                                                     | Hamburg,Hamburg                                 | ,     |                           |
```
danpat: This one might fail - oneway=-1 means you can only go f->e

ghoshkaj:
Can I put the logic of forward tags and backwards tags all into the one get_destination function (https://github.com/Project-OSRM/osrm-backend/pull/3061/files#diff-b3ef18c4e3ad22970d3b6a926d7fc8bdL3)?

Right now, the logic is broken between profiles/lua.car (for backward tags) and profiles/lib/destination.lua (for forward tags) and it’s kind of ugly and unintuitive.

I’m just not sure at what level to componentize at: 

1. One function called `get_destination `that takes care of all the logic, and we just call that one function in *profiles/lua.car*
2. Four functions,`get_destination`, `get_destination_forward`,` get_destination_backward` and  `get_destination_ref`. But then we don’t really need the **profiles/lib/destination.lua** and just do all the text building within **profiles/lua.car**

It looks like nobody apart from the car profile is using this bit of text so I think it's okay if I just do Option 1.

danpat:
@ghoshkaj Let's go with option 1 for now.  No need to overcomplicate the factoring for this, it's not like we're going to be calling `get_destination_forward` from a whole lot of places.

profiles/lib/destination.lua
```diff
@@ -24,4 +28,48 @@ function Destination.get_destination(way)
   return rv
 end
 
+
+function Destination.get_destination_forward(way)
+  local destination_ref = way:get_value_by_key("destination:ref")
+  local destination_forward = way:get_value_by_key("destination:forward")
```
danpat: Just to throw a wrench in the works here:

http://taginfo.openstreetmap.org/keys/destination%3Aref%3Aforward

`destination:ref` can *also* have forward/backward versions :-P  They're not super common, but there are several thousand of them out there in the wild, so we might as well add support for them while we're at it.

Issue #2939, opened 2016-09-22
Trip Plugin Hitting Assertion


MoKob:
We just got an alarm that the trip-plugin fails to complete correctly, due to an assertion we hit.

The assertion triggered is:

```/include/engine/api/trip_api.hpp:92
in: osrm::util::json::Array osrm::engine::api::TripAPI::MakeWaypoints(const std::vector<std::vector<unsigned int> >&, const std::vector<osrm::engine::PhantomNode>&) const: !trip_index.NotUsed()
```

Our logs don't show any trip requests during the time the assertion triggers, so we are without knowledge what kind of request triggered this assertion.

MoKob:
Turns out this happens in farthest insertion when we only see same coordinates. It is an unreasonable request, but we shouldn't fail on it.

daniel-j-h:
@chaupow would you be so kind and have a look at this.

PR #3065, opened 2016-10-17
handle trips with identical coordinates, #2939

MoKob:
# Issue

resolves https://github.com/Project-OSRM/osrm-backend/issues/2939

## Tasklist
 - [ ] review
 - [ ] adjust for for comments

### DIFF
--- a/features/testbot/trip.feature
+++ b/features/testbot/trip.feature
@@ -88,7 +88,7 @@ Feature: Basic trip planning
 
         When I plan a trip I should get
             | waypoints                       | trips              |
-            | a,b,c,d,e,f,g,h,i,j,k,l,m,n,o,p | cbalkjihgfedc,mnop |
+            | a,b,c,d,e,f,g,h,i,j,k,l,m,n,o,p | defghijklabcd,mnop |
 
     # Test single node in each component #1850
     Scenario: Testbot - Trip planning with less than 10 nodes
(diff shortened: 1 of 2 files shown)


features/testbot/trip.feature
```diff
@@ -88,7 +88,7 @@ Feature: Basic trip planning
 
         When I plan a trip I should get
             | waypoints                       | trips              |
-            | a,b,c,d,e,f,g,h,i,j,k,l,m,n,o,p | cbalkjihgfedc,mnop |
+            | a,b,c,d,e,f,g,h,i,j,k,l,m,n,o,p | defghijklabcd,mnop |
```
MoKob: Due to https://github.com/Project-OSRM/osrm-backend/issues/3067

include/engine/trip/trip_farthest_insertion.hpp
```diff
@@ -182,11 +184,15 @@ std::vector<NodeID> FarthestInsertionTrip(const NodeIDIterator &start,
...
+                // don't repeat coordinates
+                if (*x == *y)
+                    continue;
+
                 const auto xy_dist = dist_table(*x, *y);
                 // SCC decomposition done correctly?
                 BOOST_ASSERT(xy_dist != INVALID_EDGE_WEIGHT);
 
-                if (xy_dist > max_dist)
+                if (xy_dist >= max_dist)
```
chaupow: @MoKob can you quickly explain why this is needed? I see that a cucumber test is failing if this is not fixed but my head cannot get around the reason what's wrong here.

PR #3066, opened 2016-10-17
fix u-turn collapsing onto empty name / in presence of turning use lane

MoKob:
# Issue

The name change heuristic introduced in post-processing makes u-turn detection less strict, resulting in the following scenario:

screen shot 2016-10-17 at 16 10 15

This PR fixes this turn scenario (partly introduced due to market-street not being named)

## Tasklist
 - [x] review
 - [x] adjust for for comments

### DIFF
--- a/src/engine/guidance/post_processing.cpp
+++ b/src/engine/guidance/post_processing.cpp
@@ -483,10 +483,12 @@ void collapseTurnAt(std::vector<RouteStep> &steps,
         const auto next_step_index = step_index + 1;
         const bool continues_with_name_change =
             (next_step_index < steps.size()) &&
-            (steps[next_step_index].maneuver.instruction.type == TurnType::UseLane ||
+            ((steps[next_step_index].maneuver.instruction.type == TurnType::UseLane &&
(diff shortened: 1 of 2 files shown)


src/engine/guidance/post_processing.cpp
```diff
@@ -1,3 +1,5 @@
+#include "util/debug.hpp"
```
daniel-j-h: Leftover from debugging

src/engine/guidance/post_processing.cpp
```diff
@@ -655,6 +659,7 @@ std::vector<RouteStep> removeNoTurnInstructions(std::vector<RouteStep> steps)
 // that we come across.
 std::vector<RouteStep> postProcess(std::vector<RouteStep> steps)
 {
+    util::guidance::print(steps);
```
daniel-j-h: ^

MoKob:
@daniel-j-h any other comments? Or do you approve?

daniel-j-h:
Looks good @MoKob !

PR #3069, opened 2016-10-17
Fix duplicated code in tile plugin

TheMarex:
# Issue

The tile plugin had some duplicated code in it.

## Tasklist
 - [x] review
 - [x] adjust for for comments

/cc @danpat

### DIFF
--- a/src/engine/plugins/tile.cpp
+++ b/src/engine/plugins/tile.cpp
@@ -601,50 +601,24 @@ Status TilePlugin::HandleRequest(const std::shared_ptr<datafacade::BaseDataFacad
     // "pre-loop" over all the edges to create the lookup tables.  Once we have those, we
     // can then encode the features, and we'll know the indexes that feature properties
     // need to refer to.
-    std::vector<EdgeWeight> forward_weight_vector, reverse_weight_vector;
-    std::vector<std::uint8_t> forward_datasource_vector, reverse_datasource_vector;
     for (const auto &edge : edges)
     {
-        int forward_weight = 0, reverse_weight = 0;
-        std::uint8_t forward_datasource = 0;
-        std::uint8_t reverse_datasource = 0;
-        // TODO this approach of writing at least an empty vector for any segment is probably stupid
-        // (inefficient)
 
-        forward_weight_vector =
-            facade->GetUncompressedForwardWeights(edge.packed_geometry_id);
-        reverse_weight_vector =
-            facade->GetUncompressedReverseWeights(edge.packed_geometry_id);
-
-        forward_weight = forward_weight_vector[edge.fwd_segment_position];
-
-        BOOST_ASSERT(edge.fwd_segment_position < reverse_weight_vector.size());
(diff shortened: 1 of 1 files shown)


src/engine/plugins/tile.cpp
```diff
@@ -601,50 +601,24 @@ Status TilePlugin::HandleRequest(const std::shared_ptr<datafacade::BaseDataFacad
...
-        reverse_weight =
-            reverse_weight_vector[reverse_weight_vector.size() - edge.fwd_segment_position - 1];
-
-        forward_datasource_vector.clear();
-        forward_datasource_vector = facade->GetUncompressedForwardDatasources(edge.packed_geometry_id);
-        forward_datasource = forward_datasource_vector[edge.fwd_segment_position];
-
-        reverse_datasource_vector.clear();
-        reverse_datasource_vector =
+        auto forward_datasource_vector = facade->GetUncompressedForwardDatasources(edge.packed_geometry_id);
```
daniel-j-h: Here and below: we only need a read-only view? => const

src/engine/plugins/tile.cpp
```diff
@@ -673,59 +647,32 @@ Status TilePlugin::HandleRequest(const std::shared_ptr<datafacade::BaseDataFacad
...
-                    int reverse_weight = 0;
+                    const double length = osrm::util::coordinate_calculation::haversineDistance(a, b);
 
-                    std::uint8_t forward_datasource = 0;
-                    std::uint8_t reverse_datasource = 0;
-                    forward_weight_vector.clear();
-                    forward_weight_vector =
-                        facade->GetUncompressedForwardWeights(edge.packed_geometry_id);
-                    forward_weight = forward_weight_vector[edge.fwd_segment_position];
+                    auto forward_weight_vector = facade->GetUncompressedForwardWeights(edge.packed_geometry_id);
```
daniel-j-h: Same here

src/engine/plugins/tile.cpp
```diff
@@ -673,59 +647,32 @@ Status TilePlugin::HandleRequest(const std::shared_ptr<datafacade::BaseDataFacad
...
-                                                  edge.fwd_segment_position - 1];
-
-                    // Keep track of the highest datasource seen so that we don't write unnecessary
-                    // data to the layer attribute values
-                    max_datasource_id = std::max(max_datasource_id, forward_datasource);
-                    max_datasource_id = std::max(max_datasource_id, reverse_datasource);
+                    if (name_offsets.find(name) == name_offsets.end())
+                    {
+                        names.push_back(name);
+                        name_offsets[name] = names.size() - 1;
```
daniel-j-h: Hm do we need the name further down here? If not I would re-order those two lines and `std::move` the already allocated name into the vector. (push back has an rvalue overload)

TheMarex:
@daniel-j-h addressed your comments. Let me know if that works for you.

daniel-j-h:
:+1:

PR #3071, opened 2016-10-17
Docs: Rotaries can have exits as well

freenerd:
# Issue

Adding missing docs for rotaries and exits.

## Tasklist
 - [x] review and merge


MoKob:
This is also the case for roundabout-turns. While a turn is given in the form of a direction, it also provides the exit number (might be interesting for visual purposes/additional safety).

PR #3073, opened 2016-10-17
More Low Speed Service Tags

daniel-j-h:
Follow up on https://github.com/Project-OSRM/osrm-backend/commit/f540485cb085391448f66eccb68ef5dfc79b3aaf.

http://taginfo.openstreetmap.org/keys/service#values

## Tasklist
 - [x] review
 - [x] adjust for for comments


TheMarex:
Looks good to me.

Issue #2989, opened 2016-10-03
Deduplicate code between InternalDataFacade and osrm-datastore


TheMarex:
In osrm-datastore there is a lot of code for loading all the `.osrm.*` files into memory. This code replicates a lot of the code that we wrote for the `InternalDataFacade`.

We should try to deduplicate this as much as possible, because having duplicate code is error prone. The easiest way I see to do factor this out into a shared header in `inlcude/storage`. Something like `include/storage/io.hpp` that implements loading this data to somewhere in memory.

A trivial example for the new interface for `.osrm.properties` in `io.hpp`:

``` cpp
size_t GetProfilePropertiesSize(boost::filestream::ifstream &input_stream);
void ReadProfileProperties(boost::filestream::ifstream &input_stream, ProfileProperties* buffer, size_t size);
```

Both `osrm-datastore` and `InternalDataFacade` can use this to load data: Either into internal or shared memory.

I recommend doing this one file at a time.

TheMarex:
Straight forward:

- [ ] `.hsgr` file `InternalDataFacde::LoadGraph()` (there is already a function `readHSGRFileFromStream`) that can be adapted
- [ ] `.nodes` `InternalDataFacade::LoadNodeAndEdgeInformation`
- [ ] `.edges` `InternalDataFacade::LoadNodeAndEdgeInformation`
- [ ] `.datasource_names` `InternalDataFacade::LoadDatasourceInfo`
- [ ] `.datasource_indexes` `InternalDataFacade::LoadDatasourceInfo`
- [ ] `.timestamp` `InternalDataFacade::LoadTimestamp`
- [ ] `.profile_properties` `InternalDataFacade::LoadProfileProperties`
- [ ] `.ramIndex` `InternalDataFacde::LoadRTree()`

Need more investigation:

- [ ] `.tld` `InternalDataFacade::LoadLaneDescriptions`
- [ ] `.tls` `InternalDataFacde::LoadLaneTupleIdPairs`
- [ ] `.names` `InternalDataFacade::LoadStreetNames` (has own de-serialization code: in `range_table.hpp` implemented as `operator>>`)
- [ ] `.core` `InternalDataFacade::LoadCoreInformation` (look at the packing logic needed for `std::vector<bool>` in `storage.cpp`)
- [ ] `.icd` `InternalDataFacade::LoadIntersectionClasses` (has own de-serialization code: in `range_table.hpp`)

Don't touch:

[...]

PR #3074, opened 2016-10-17
Move loading of .hsgr to io.hpp

TheMarex:
# Issue

This is the first part of https://github.com/Project-OSRM/osrm-backend/issues/2989

## Tasklist
 - [x] review
 - [x] adjust for for comments

/cc @karenzshea @chaupow @miccolis @danpat

### DIFF
--- a/include/storage/io.hpp
+++ b/include/storage/io.hpp
@@ -1,0 +1,68 @@
+#ifndef OSRM_STORAGE_IO_HPP_
+#define OSRM_STORAGE_IO_HPP_
+
+#include "util/fingerprint.hpp"
+#include "util/simple_logger.hpp"
+
+#include <boost/filesystem/fstream.hpp>
+
+#include <tuple>
+
+namespace osrm
+{
+namespace storage
+{
+namespace io
+{
+
+#pragma pack(push, 1)
+struct HSGRHeader
+{
+    std::uint32_t checksum;
+    std::uint32_t number_of_nodes;
+    std::uint32_t number_of_edges;
+};
+#pragma pack(pop)
+static_assert(sizeof(HSGRHeader) == 12, "HSGRHeader is not packed");
+
+// Returns the checksum and the number of nodes and number of edges
+inline HSGRHeader readHSGRHeader(boost::filesystem::ifstream &input_stream)
+{
+    const util::FingerPrint fingerprint_valid = util::FingerPrint::GetValid();
+    util::FingerPrint fingerprint_loaded;
+    input_stream.read(reinterpret_cast<char *>(&fingerprint_loaded), sizeof(util::FingerPrint));
+    if (!fingerprint_loaded.TestGraphUtil(fingerprint_valid))
+    {
+        util::SimpleLogger().Write(logWARNING) << ".hsgr was prepared with different build.\n"
+                                                  "Reprocess to get rid of this warning.";
+    }
+
+    HSGRHeader header;
+    input_stream.read(reinterpret_cast<char *>(&header.checksum), sizeof(header.checksum));
+    input_stream.read(reinterpret_cast<char *>(&header.number_of_nodes), sizeof(header.number_of_nodes));
+    input_stream.read(reinterpret_cast<char *>(&header.number_of_edges), sizeof(header.number_of_edges));
+
+    BOOST_ASSERT_MSG(0 != header.number_of_nodes, "number of nodes is zero");
+    // number of edges can be zero, this is the case in a few test fixtures
+
+    return header;
+}
+
+// Needs to be called after getHSGRSize() to get the correct offset in the stream
+// 
+template <typename NodeT, typename EdgeT>
+void readHSGR(boost::filesystem::ifstream &input_stream,
+              NodeT *node_buffer,
+              std::uint32_t number_of_nodes,
+              EdgeT *edge_buffer,
+              std::uint32_t number_of_edges)
+{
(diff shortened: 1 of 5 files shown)


include/storage/io.hpp
```diff
@@ -0,0 +1,58 @@
...
+
+namespace osrm
+{
+namespace storage
+{
+namespace io
+{
+
+// Returns the checksum and the number of nodes and number of edges
+std::tuple<unsigned, std::size_t, std::size_t> readHSGRSize(boost::filesystem::ifstream &input_stream)
```
daniel-j-h: Maybe call it `readHSGRMetadata` (?) - also all of those functions will return a fingerprint and some additional data, right? What about making a namedtuple type instead so call sites don't have to `get<Nth>` all over the place.

    struct { Fingerprint fingerprint; size_t nbNodes; size_t nbEdges; }

or similar.

include/storage/io.hpp
```diff
@@ -0,0 +1,58 @@
...
+{
+    const util::FingerPrint fingerprint_valid = util::FingerPrint::GetValid();
+    util::FingerPrint fingerprint_loaded;
+    input_stream.read(reinterpret_cast<char *>(&fingerprint_loaded), sizeof(util::FingerPrint));
```
daniel-j-h: read can fail after which the stream's bool conversion will fail - `if (!stream) ..`

include/storage/io.hpp
```diff
@@ -0,0 +1,58 @@
...
+    if (!fingerprint_loaded.TestGraphUtil(fingerprint_valid))
+    {
+        util::SimpleLogger().Write(logWARNING) << ".hsgr was prepared with different build.\n"
+                                                  "Reprocess to get rid of this warning.";
+    }
+
+    unsigned check_sum = 0;
+    unsigned number_of_nodes = 0;
+    unsigned number_of_edges = 0;
```
daniel-j-h: Same here: fixed-size for all of those? (Also at write-time then)

include/storage/io.hpp
```diff
@@ -0,0 +1,58 @@
...
+    return std::make_tuple(check_sum, number_of_nodes, number_of_edges);
+}
+
+// Needs to be called after getHSGRSize() to get the correct offset in the stream
+template <typename NodeT, typename EdgeT>
+void readHSGR(boost::filesystem::ifstream &input_stream,
+              NodeT *node_buffer,
+              std::size_t number_of_nodes,
+              EdgeT *edge_buffer,
+              std::size_t number_of_edges)
```
daniel-j-h: Can those raw pointers ever be null?

include/storage/io.hpp
```diff
@@ -0,0 +1,58 @@
...
+    BOOST_ASSERT_MSG(0 != number_of_nodes, "number of nodes is zero");
+    BOOST_ASSERT_MSG(0 != number_of_edges, "number of edges is zero");
+    input_stream.read(reinterpret_cast<char *>(check_sum), sizeof(unsigned));
+    input_stream.read(reinterpret_cast<char *>(&number_of_nodes), sizeof(unsigned));
+    input_stream.read(reinterpret_cast<char *>(&number_of_edges), sizeof(unsigned));
+
+    return std::make_tuple(check_sum, number_of_nodes, number_of_edges);
+}
+
+// Needs to be called after getHSGRSize() to get the correct offset in the stream
```
daniel-j-h: Why do I have to care about this? if I _have to_ do something, why don't you already do it for me?

src/storage/storage.cpp
```diff
@@ -189,35 +190,13 @@ int Storage::Run()
...
-        util::SimpleLogger().Write(logWARNING) << ".hsgr was prepared with different build. "
-                                                  "Reprocess to get rid of this warning.";
-    }
-
-    // load checksum
     unsigned checksum = 0;
-    hsgr_input_stream.read((char *)&checksum, sizeof(unsigned));
+    std::size_t number_of_graph_nodes = 0;
+    std::size_t number_of_graph_edges = 0;
+    std::tie(checksum, number_of_graph_nodes, number_of_graph_edges) = io::readHSGRSize(hsgr_input_stream);
```
daniel-j-h: Depends on tuple order - there are no named product types in C++ (think Python's namedtuple) but at least create a small type and use semantic member names here.

include/storage/io.hpp
TheMarex: Can we make it throw an exception always?

include/storage/io.hpp
MoKob: Also, is there a reason that we have to use raw pointers at all?

TheMarex:
@daniel-j-h please confirm if this works for you now.

include/storage/io.hpp
```diff
@@ -0,0 +1,68 @@
...
+// 
+template <typename NodeT, typename EdgeT>
+void readHSGR(boost::filesystem::ifstream &input_stream,
+              NodeT *node_buffer,
+              std::uint32_t number_of_nodes,
+              EdgeT *edge_buffer,
+              std::uint32_t number_of_edges)
+{
+    input_stream.read(reinterpret_cast<char *>(node_buffer), number_of_nodes * sizeof(NodeT));
+    input_stream.read(reinterpret_cast<char *>(edge_buffer), number_of_edges * sizeof(EdgeT));
```
daniel-j-h: These can still fail leaving the stream corrupted - the call site does not check for this either. But it's a more general problem of this code base..

Issue #2841, opened 2016-09-02
Map Matching: Use Max Speed from Profiles


daniel-j-h:
Currently Map Matching assumes we're matching traces for cars, working with 180 km/h as max speed. This makes no sense for other profiles.

https://github.com/Project-OSRM/osrm-backend/blob/6fac14dbd830e5ff195c0cbd5290a34251b2f0af/include/engine/routing_algorithms/map_matching.hpp#L36

We can fix this by
- storing a new profile property internally https://github.com/Project-OSRM/osrm-backend/blob/master/include/extractor/profile_properties.hpp
- registering the new variable for Lua https://github.com/Project-OSRM/osrm-backend/blob/6fac14dbd830e5ff195c0cbd5290a34251b2f0af/src/extractor/scripting_environment_lua.cpp#L125
- setting the variable on a per-profile level from Lua https://github.com/Project-OSRM/osrm-backend/blob/6fac14dbd830e5ff195c0cbd5290a34251b2f0af/profiles/car.lua#L144

PR #3089, opened 2016-10-18
added max_speed to the profile

ghoshkaj:
# Issue

Map Matching: Use Max Speed from Profiles #2841
## Tasklist
- [x] Add max_speed to the profile
- [x] Fix broken tests
- [x] Add new cucumber tests
- [x] Fix algorithm code

### DIFF
--- a/features/step_definitions/matching.js
+++ b/features/step_definitions/matching.js
@@ -132,10 +132,10 @@ module.exports = function () {
 
                     var testSubMatching = (sub, si) => {
                         var testSubNode = (ni) => {
-                        var node = this.findNodeByName(sub[ni]),
-                            outNode = subMatchings[si][ni];
+                            var node = this.findNodeByName(sub[ni]),
+                                outNode = subMatchings[si][ni];
 
-                        if (this.FuzzyMatch.matchLocation(outNode, node)) {
+                            if (this.FuzzyMatch.matchLocation(outNode, node)) {
                                 encodedResult += sub[ni];
                                 extendedTarget += sub[ni];
                             } else {
--- a/features/step_definitions/matching.js
+++ b/features/step_definitions/matching.js
@@ -4,13 +4,6 @@ var util = require('util');
 var d3 = require('d3-queue');
 var polyline = require('polyline');
 
-// 1) When you have a Trace  -- make sure that they get split into the correct number of parts
-// 2) Make sure that each part of the trace mmatches 
-// legs of each submatch and check that they're there
-// and that they're on the correct path
-
-// "geometries=geojson"
-
 module.exports = function () {
     this.When(/^I match I should get$/, (table, callback) => {
         var got;
@@ -26,7 +19,6 @@ module.exports = function () {
 
                     if (res.body.length) {
                         json = JSON.parse(res.body);
-                        console.log('json', JSON.stringify(json));
                     }
 
                     if (headers.has('status')) {
@@ -80,25 +72,21 @@ module.exports = function () {
                         if (headers.has('turns')) {
                             if (json.matchings.length != 1) throw new Error('*** Checking turns only supported for matchings with one subtrace');
                             turns = this.turnList(json.matchings[0].instructions);
-                            console.log('turns', turns);
                         }
 
                         if (headers.has('route')) {
                             if (json.matchings.length != 1) throw new Error('*** Checking route only supported for matchings with one subtrace');
                             route = this.wayList(json.matchings[0]);
-                            console.log('route', route);
                         }
 
                         if (headers.has('duration')) {
                             if (json.matchings.length != 1) throw new Error('*** Checking duration only supported for matchings with one subtrace');
                             duration = json.matchings[0].duration;
-                            console.log('duration', duration);
                         }
 
                         if (headers.has('annotation')) {
                             if (json.matchings.length != 1) throw new Error('*** Checking annotation only supported for matchings with one subtrace');
                             annotation = this.annotationList(json.matchings[0]);
-                            console.log('annotation', annotation);
                         }
 
                         if (headers.has('geometry')) {
@@ -109,7 +97,6 @@ module.exports = function () {
                         if (headers.has('OSM IDs')) {
                             if (json.matchings.length != 1) throw new Error('*** Checking annotation only supported for matchings with one subtrace');
                             OSMIDs = this.OSMIDList(json.matchings[0]);
-                            console.log('OSMIDs', OSMIDs);
                         }
                     }
 
@@ -154,8 +141,6 @@ module.exports = function () {
                         } else {
                             var sq = d3.queue();
                             var testSubNode = (ni, ncb) => {
-                                console.log('subMatchings[si][ni]', subMatchings[si][ni]);
-                                console.log('sub[ni]', sub[ni]);
                                 var node = this.findNodeByName(sub[ni]),
                                     outNode = subMatchings[si][ni];
                                 
@@ -211,7 +196,6 @@ module.exports = function () {
                 };
 
                 if (row.request) {
-                    console.log('row.request', row.request);
                     got = {};
                     got.request = row.request;
                     this.requestUrl(row.request, afterRequest);
--- a/features/step_definitions/matching.js
+++ b/features/step_definitions/matching.js
@@ -1,3 +1,5 @@
+'use strict';
+
 var util = require('util');
 var d3 = require('d3-queue');
 var polyline = require('polyline');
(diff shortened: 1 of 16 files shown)


ghoshkaj:
@daniel-j-h for review.  I followed the instructions [here](https://github.com/Project-OSRM/osrm-backend/issues/2841#issue-174738576), but am missing some context. Questions below address context problems:

1. Not sure where max_speed is actually used within the lua profiles. There seems to be [other places](https://github.com/Project-OSRM/osrm-backend/blob/6fac14dbd830e5ff195c0cbd5290a34251b2f0af/profiles/car.lua#L98) and [other functions](https://github.com/Project-OSRM/osrm-backend/blob/6fac14dbd830e5ff195c0cbd5290a34251b2f0af/profiles/car.lua#L181),  where max_speeds are defined already. So I'm a bit confused about where to incorpoerate `properties.max_speed` into the profile code, if we need to at all.
2. The values for max_speed in walking and bicycle are average speeds from google for those modes of commuting. Is that an okay value to use?
3. How does incorporating profile speeds affect https://github.com/Project-OSRM/osrm-backend/blob/6fac14dbd830e5ff195c0cbd5290a34251b2f0af/include/engine/routing_algorithms/map_matching.hpp#L36 at all?

daniel-j-h:
> Not sure where max_speed is actually used within the lua profiles. There seems to be other places and other functions, where max_speeds are defined already. So I'm a bit confused about where to incorpoerate properties.max_speed into the profile code, if we need to at all.

You want to add a global `max_speed_for_map_matching` property to the profiles which won't ever get used _inside_ the Lua profiles. Instead you have to pass it on to the `ProfileProperties` type for the Map Matching plugin to make use of it.

> The values for max_speed in walking and bicycle are average speeds from google for those modes of commuting. Is that an okay value to use?

No. Those values will be used by the Map Matching plugin (this plugin snaps noisy GPS traces to the road network, see e.g. [this example](http://bl.ocks.org/anonymous/raw/3434747370929f977db4ba56a91a96cb/)) to filter out errors in the data. For car there is already a value in the plugin (check issue and ref I linked there). You have to come up with values for bike and walk for which we filter out traces going faster.



> How does incorporating profile speeds affect https://github.com/Project-OSRM/osrm-backend/blob/6fac14dbd830e5ff195c0cbd5290a34251b2f0af/include/engine/routing_algorithms/map_matching.hpp#L36 at all?

You have to remove this line. And then read in the speed from the `ProfileProperties` which you set based in the Lua profile.

src/extractor/scripting_environment_lua.cpp
```diff
@@ -130,6 +130,9 @@ void LuaScriptingEnvironment::InitContext(LuaScriptingContext &context)
              .property("u_turn_penalty",
                        &ProfileProperties::GetUturnPenalty,
                        &ProfileProperties::SetUturnPenalty)
+             .property("max_speed",
+                       &ProfileProperties::MaxSpeed,
+                       &ProfileProperties::MaxSpeed)
```
ghoshkaj: Yeah, sowry, my bad.

include/extractor/profile_properties.hpp
```diff
@@ -12,7 +12,7 @@ struct ProfileProperties
 {
     ProfileProperties()
         : traffic_signal_penalty(0), u_turn_penalty(0), continue_straight_at_waypoint(true),
-          use_turn_restrictions(false), left_hand_driving(false)
+          use_turn_restrictions(false), left_hand_driving(false), max_speed(0)
```
MoKob: didn't we have discussions on this and that a change here would be changing the API? I am thinking of initialiser lists breaking, if we add an additional member.

daniel-j-h: Care to elaborate?

We don't expose the ProfilesProperties type - we bind it for Lua, but if we set a default value here users don't have to set one in the profiles. Should be compatible from what I can see?

MoKob: That was just a question, since I was unsure how this is exposed. I remembered problems coming up with a change @danpat did some while back. If we don't expose Profile Properties anywhere but in lua, thats totally fine.

include/extractor/profile_properties.hpp
```diff
@@ -11,8 +11,8 @@ namespace extractor
 struct ProfileProperties
 {
     ProfileProperties()
-        : traffic_signal_penalty(0), u_turn_penalty(0), continue_straight_at_waypoint(true),
-          use_turn_restrictions(false), left_hand_driving(false)
+        : traffic_signal_penalty(0), u_turn_penalty(0), max_speed_for_map_matching(0),
```
MoKob: is 0 what we want here? our default is currently 180km/h

ghoshkaj:
@daniel-j-h would you please review? My problem is that I can't get the tests to fail.

ghoshkaj:
Or @danpat would you please review this?

include/engine/routing_algorithms/map_matching.hpp
```diff
@@ -98,7 +100,11 @@ class MapMatching final : public BasicRoutingInterface<DataFacadeT, MapMatching<
         const auto max_distance_delta = [&] {
             if (use_timestamps)
             {
-                return median_sample_time * MAX_SPEED;
+                std::cout << "facade.GetMapMatchingMaxSpeed()" << std::endl;
+                std::cout << facade.GetMapMatchingMaxSpeed() << std::endl;
+                std::cout << "median_sample_time" << std::endl;
+                std::cout << median_sample_time << std::endl;
```
daniel-j-h: These debug statements needs to go before merging into master

include/extractor/profile_properties.hpp
```diff
@@ -11,8 +11,8 @@ namespace extractor
 struct ProfileProperties
 {
     ProfileProperties()
-        : traffic_signal_penalty(0), u_turn_penalty(0), continue_straight_at_waypoint(true),
-          use_turn_restrictions(false), left_hand_driving(false)
+        : traffic_signal_penalty(0), u_turn_penalty(0), max_speed_for_map_matching(180/3.6),
```
daniel-j-h: Maybe add a 

    const constexpr auto DEFAULT_MAX_SPEED = ..;

at the top of the struct and then use it here

include/extractor/profile_properties.hpp
```diff
@@ -30,10 +30,18 @@ struct ProfileProperties
         traffic_signal_penalty = boost::numeric_cast<int>(traffic_signal_penalty_ * 10.);
     }
 
+    double GetMaxSpeedForMapMatching() const { return max_speed_for_map_matching; } // /3.6 converts kmph to m/s
+
+    void SetMaxSpeedForMapMatching(const double max_speed_for_map_matching_) 
+    {
+        max_speed_for_map_matching = boost::numeric_cast<int>(max_speed_for_map_matching_);  // *3.6 converts  m/s to kmph
```
daniel-j-h: Neither here not a couple of lines above is a 3.6 multiplication - comments off?

daniel-j-h: Why converting to int - the member attribute is a double, too

profiles/bicycle.lua
```diff
@@ -93,8 +93,9 @@ surface_speeds = {
 
 -- these need to be global because they are accesed externaly
 properties.traffic_signal_penalty        = 2
-properties.use_turn_restrictions         = false
 properties.u_turn_penalty                = 20
+properties.max_speed_for_map_matching    = 30/3.6 -- 15.5 km/h is the average speed
```
daniel-j-h: Just to make it clear here: the max speed is like an upper bound over which traces are getting filtered out - you don't want to set an average speed here, otherwise you will throw away half of all traces. Max speed should be an upper bound of what makes sense per profile. (Think of a bike trace with 200 km/h _probably_ makes no sense :P)

include/extractor/profile_properties.hpp
ghoshkaj: Good thinking! Done!

TheMarex:
Squashed and rebase on master.

features/step_definitions/matching.js
```diff
@@ -121,41 +134,46 @@ module.exports = function () {
...
+                            sq.defer(testSubNode, i);
                         }
+
+                        sq.awaitAll(scb);
+                        
                     };
 
+                    if (subMatchings.length != row.matchings.split(',').length) {
+                        ok = false;
+                        throw new Error('*** table matchings and api response are not the same');
```
TheMarex: Instead of using `throw` here it should call the callback `cb(new Erro(...));`.

features/step_definitions/matching.js
```diff
@@ -121,41 +134,46 @@ module.exports = function () {
...
-
-                            var testSubNode = (ni, ncb) => {
-                                var node = this.findNodeByName(sub[ni]),
-                                    outNode = subMatchings[si][ni];
-
-                                if (this.FuzzyMatch.matchLocation(outNode, node)) {
-                                    encodedResult += sub[ni];
-                                    extendedTarget += sub[ni];
-                                } else {
+                        var sq = d3.queue();
```
TheMarex: This function is actually does not need to be asynchronous. Taking a callback `scb` is superfluous.

features/step_definitions/matching.js
```diff
@@ -121,41 +134,46 @@ module.exports = function () {
...
+                        var testSubNode = (ni, ncb) => {
```
TheMarex: Same for this helper function. Nothing that gets called is asynchronous. You can omit the callback and instead return the result directly.

features/step_definitions/matching.js
```diff
@@ -121,41 +134,46 @@ module.exports = function () {
...
-                                sq.defer(testSubNode, i);
+                                extendedTarget += util.format('%s [%d,%d]', node.lat, node.lon);
+                                ok = false;
                             }
+                            ncb();
+                        };
 
-                            sq.awaitAll(scb);
+                        for (var i=0; i<sub.length; i++) {
+                            sq.defer(testSubNode, i);
```
TheMarex: Instead of using a queue here you call also use the synchronous version: `testSubNode(i)`.

features/step_definitions/matching.js
```diff
@@ -121,41 +134,46 @@ module.exports = function () {
...
+                        ok = false;
+                        throw new Error('*** table matchings and api response are not the same');
+                    }
+
                     row.matchings.split(',').forEach((sub, si) => {
                         q.defer(testSubMatching, sub, si);
                     });
 
-                    q.awaitAll(() => {
+                    q.awaitAll((error) => {
```
TheMarex: Same here, the queue and async calls are not necessary. A simple loop and synchronous functions will do the same.

PR #3091, opened 2016-10-18
Removes Gnu Date Usage

daniel-j-h:
There is no -Iseconds in non-Gnu date binaries. I'm sorry.

\U+1Fstallman-emoji

### DIFF
--- a/scripts/timer.sh
+++ b/scripts/timer.sh
@@ -7,6 +7,6 @@ START=$(date "+%s.%N")
 /bin/bash -c "$CMD"
 END=$(date "+%s.%N")
 TIME="$(echo "$END - $START" | bc)s"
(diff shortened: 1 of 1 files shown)


TheMarex:
Just for context, where is this a problem?

daniel-j-h:
Noticed this today in Travis OSX builds.

The real issue was a timeout, but I looked into the logs and turns out this never worked on OSX.

danpat:
@TheMarex example error from this: https://travis-ci.org/Project-OSRM/osrm-backend/jobs/168484795#L1056

The script has been failing to output the timing information on OSX for a while now - but it hasn't been causing builds to fail.

TheMarex:
I see. :+1: Good to merge.

Issue #3070, opened 2016-10-17
Parallelize Turn Penalty Updates


daniel-j-h:
We provide two ways for ingesting real-time'ish data: a csv file for 1/ updating edge weights and 2/ a csv file for updating turn penalties. Both files are getting ingested by `osrm-contract` - specifically this happens in the [contractor.cpp](https://github.com/Project-OSRM/osrm-backend/blob/8ed6bb8a1b2c21819822da714b81a4f7b72bde58/src/contractor/contractor.cpp) file.

In https://github.com/Project-OSRM/osrm-backend/pull/2412 we made 1/ faster by doing more in parallel and using a flat-map + binary search approach instead of a hash table.

Now the same has to be done for loading turn penalties.

Tasks:
- [x] I can't find docs for loading turn penalties: get yourself familiar with how updating turn penalties works and write docs for the Wiki (maybe Traffic page)
- [ ] Measure / profile where we spent the most time, what should be prioritized
- [ ] Follow the contractor's code path for updating edge speeds: updating turn penalties has to be adapted to this more or less 1:1
- [ ] parallelize / flat-map turn penalty updating similar to edge weight updating
- [ ] Measure speedup

cc @karenzshea

karenzshea:
Also included in this will be a refactor to use boost's [flat_map](http://www.boost.org/doc/libs/1_58_0/doc/html/boost/container/flat_map.html)

PR #3095, opened 2016-10-18
Parse turn penalty files in parallel

karenzshea:
# Issue

Hits https://github.com/Project-OSRM/osrm-backend/issues/3070. We already load edge weight update files in parallel, so some code has been borrowed to load penalties in parallel as well. The `find()` function has been adapted to work on either the speed source or turn penalty flat maps.

I explored using boost's flat_map here as well, but it turned out to be less flexible in terms of sorting as a homemade flat_map. Boost's flat map handles deduplication on insertion, and can be sorted with a custom sort operator like we have, which sorts by `from, to, speed source`. However, to then use the flat_map's `find()` function, the find can only work on the same keys that the flat_map was sorted by, but the way we use `find`, we only know the `from, to` values.

## Tasklist
 - [ ] Write docs for the wiki on how turn penalty loading works
-  [ ] Test for speedup
 - [ ] review
 - [ ] adjust for for comments

## Requirements / Relations
cc @daniel-j-h

### DIFF
--- a/src/contractor/contractor.cpp
+++ b/src/contractor/contractor.cpp
@@ -432,6 +432,29 @@ parse_turn_penalty_lookup_from_csv_files(const std::vector<std::string> &turn_pe
 
     tbb::parallel_for(std::size_t{0}, turn_penalty_filenames.size(), parse_turn_penalty_file);
 
+    // With flattened map-ish view of all the files, sort and unique them on from,to,source
+    // The greater '>' is used here since we want to give files later on higher precedence
+    const auto sort_by = [](const TurnPenaltySource &lhs, const TurnPenaltySource &rhs) {
+        return std::tie(lhs.segment.from, lhs.segment.via, lhs.segment.to, lhs.penalty_source.source) >
+               std::tie(rhs.segment.from, rhs.segment.via, rhs.segment.to, rhs.penalty_source.source);
+    };
+
+    std::stable_sort(begin(map), end(map), sort_by);
+
+    // Unique only on from,to to take the source precedence into account and remove duplicates
+    const auto unique_by = [](const TurnPenaltySource &lhs, const TurnPenaltySource &rhs) {
+        return std::tie(lhs.segment.from, lhs.segment.via, lhs.segment.to) ==
+               std::tie(rhs.segment.from, rhs.segment.via, rhs.segment.to);
+    };
+
+    const auto it = std::unique(begin(map), end(map), unique_by);
+
+    map.erase(it, end(map));
+
+    util::SimpleLogger().Write() << "In total loaded " << turn_penalty_filenames.size()
+                                 << " turn penalty file(s) with a total of " << map.size()
+                                 << " unique values";
+
     return map;
 }
 } // anon ns
--- a/src/contractor/contractor.cpp
+++ b/src/contractor/contractor.cpp
@@ -210,6 +210,10 @@ namespace
 struct Segment final
 {
     OSMNodeID from, to;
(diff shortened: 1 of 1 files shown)


src/contractor/contractor.cpp
```diff
@@ -222,36 +226,52 @@ struct SegmentSpeedSource final
 {
     Segment segment;
     SpeedSource speed_source;
+    bool operator<(const SegmentSpeedSource &other)
+    {
+        return std::tie(segment.from, segment.to) > std::tie(other.segment.from, other.segment.to);
```
daniel-j-h: maybe add a comment here - looks like you're lying to the user < / >

src/contractor/contractor.cpp
```diff
@@ -222,36 +226,52 @@ struct SegmentSpeedSource final
...
+    double penalty;
+    std::uint8_t source;
+};
+struct TurnPenaltySource final
+{
+    Turn segment;
+    PenaltySource penalty_source;
+    bool operator<(const TurnPenaltySource &other)
+    {
+        return std::tie(segment.from, segment.via, segment.to) >
```
daniel-j-h: same

src/contractor/contractor.cpp
```diff
@@ -222,36 +226,52 @@ struct SegmentSpeedSource final
...
+               std::tie(other.segment.from, other.segment.via, other.segment.to);
+    }
+};
+using TurnPenaltySourceFlatMap = std::vector<TurnPenaltySource>;
 using SegmentSpeedSourceFlatMap = std::vector<SegmentSpeedSource>;
 
 // Binary Search over a flattened key,val Segment storage
-SegmentSpeedSourceFlatMap::iterator find(SegmentSpeedSourceFlatMap &map, const Segment &key)
+template <typename FlatMap, typename SegmentKey>
+typename FlatMap::iterator find(FlatMap &map, const SegmentKey &key)
```
daniel-j-h: You can just say `auto` as return value and the compiler will pick up the right type

karenzshea:
### Process for measuring speedup
- create 40 turn penalty files between sizes `83k` and `2.9M`, minimal duplicated values as possible
- Add `TIMER_START/STOP` code to penalty parsing code
- build OSRM on master OR on this branch
- run extraction. run contraction with as many `--turn-penalty-file` flags as files, save timer reported seconds spent parsing
- repeat three times

### Results
Average of 5 runs on master branch: 0.292 sec
Average of 5 runs on parallel branch: 0.1346978

karenzshea:
### benchmark
Had to rerun benchmarks as the unique/sorting steps got lost during my first rebase. Same setup as above but with 22 files, ~82mb total

**Master** average of 1.0488088 secs
**Parallel** average of 0.6647624 secs

Also I added a little bit of background to the Traffic wiki about how to include and process turn penalties.

src/contractor/contractor.cpp
```diff
@@ -616,8 +680,8 @@ EdgeID Contractor::LoadEdgeExpandedGraph(
                 const double segment_length = util::coordinate_calculation::greatCircleDistance(
                     util::Coordinate{u->lon, u->lat}, util::Coordinate{v->lon, v->lat});
 
-                auto forward_speed_iter =
-                    find(segment_speed_lookup, Segment{u->node_id, v->node_id});
+                auto forward_speed_iter = find(
+                    segment_speed_lookup, SegmentSpeedSource{{u->node_id, v->node_id}, {0, 0}});
```
daniel-j-h: Usually you'd hide this in the `find` function only taking a Segment.

Issue #3080, opened 2016-10-18
Create a toolkit for combining steps


MoKob:
Every route coming from OSRM that includes step undergoes some post-processing.

During this post-processing we can decide to collapse multiple steps into a single turn instruction.
In that process, we call the functions `elongate` ([here](https://github.com/Project-OSRM/osrm-backend/blob/master/src/engine/guidance/post_processing.cpp#L562-L593)) and `forwardStepSignage` ([here](https://github.com/Project-OSRM/osrm-backend/blob/master/src/engine/guidance/post_processing.cpp#L51-L59)).

Since our RouteSteps contain more and more data, we should move these functions out into a general toolkit to combine steps.

I imagine a function that takes as input two steps `A` and `B` as well as a combination scenario. This scenario should describe which one of the two will be the location of the turn and which meta-data to preserve.

E.g. in

```
 a - b - c - e
     |   |
     f   d
```

When turning from `d` via `c` and `b` onto `a`, we might end up with two steps. The first indicating a turn and the second indicating a new name (since the intersection b-c can belong to any of the roads).

In this situation we would like to keep the location of the first turn (`d-c` onto `c-b`) but want to have the name of `c-b to`a-b`which will in the end return`turn left onto `a-b`.

To improve readability of post-processing, having a clear toolkit function for this would help a lot.

MoKob:
As a first step, https://github.com/Project-OSRM/osrm-backend/pull/3261 allows to test for turn locations to ensure the correct turn location is announced.

TheMarex:
`elongate` and similar functions should be made superfluous by not generating `RouteStep` objects for these intermediate instructions in the first place.

PR #3097, opened 2016-10-19
Failing Testcase for Collapsing Away Steps with Destinations

daniel-j-h:
This test case is currently failing: we `forwardStepSignage` the destination signs but we really want to tell the user about the sign she sees _at that maneuver_ and not a sign that will come up later.

@MoKob what's your take on forwarding _all_ signage?

(I added this test mainly because the question about lane+destinations came up)

### DIFF
--- a/features/guidance/anticipate-lanes.feature
+++ b/features/guidance/anticipate-lanes.feature
@@ -297,19 +297,19 @@ Feature: Turn Lane Guidance
                """
 
            And the ways
-               | nodes | turn:lanes:forward                     | name |
-               | ab    |                                        | main |
-               | bc    | left\|through\|through\|through\|right | main |
-               | cd    | left\|through\|right                   | main |
-               | de    |                                        | main |
-               | cf    |                                        | off  |
-               | ch    |                                        | off  |
-               | dg    |                                        | off  |
-               | di    |                                        | off  |
+               | nodes | turn:lanes:forward                     | name | destination | oneway |
+               | ab    |                                        | main | One         | yes    |
+               | bc    | left\|through\|through\|through\|right | main | Two         | yes    |
+               | cd    | left\|through\|right                   | main | Three       | yes    |
+               | de    |                                        | main | Four        | yes    |
+               | cf    |                                        | off  |             | yes    |
+               | ch    |                                        | off  |             | yes    |
+               | dg    |                                        | off  |             | yes    |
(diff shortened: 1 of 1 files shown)


MoKob:
This looks like a perfect example for https://github.com/Project-OSRM/osrm-backend/issues/3080.

Right now, not all items are considered correctly. Destinations is obviously one of these cases.

daniel-j-h:
cc @pratikyadav for context.

MoKob:
Just cherry-picked this into my post-processing branch. Lets keep this open, until post processing refactor actually gets merged and includes this test.

MoKob:
```
    [0]: 17 0  2 (lon:1, lat:0.99964)  Duration: 5.6 Distance: 79.9595383891 Geometry: 0 3 Exit: 0 Mode: 1
    Intersections: 1 [(Lanes: 0 255 [18446744073709551615,0] bearings: 90, entry:  true)] name[8]: main Ref:  Pronunciation: Destination: One
    [1]: 16 4  0 (lon:1.000719, lat:0.99964)  Duration: 2.8 Distance: 39.9241645086 Geometry: 2 4 Exit: 0 Mode: 1
    Intersections: 1 [(Lanes: 3 1 [3,1] bearings: 30 90 150 270, entry:  true true true false)] name[16]: main Ref:  Pronunciation: Destination: Three
    [2]: 16 4  0 (lon:1.001078, lat:0.99964)  Duration: 2.8 Distance: 40.0353738805 Geometry: 3 5 Exit: 0 Mode: 1
    Intersections: 1 [(Lanes: 1 1 [3,1] bearings: 30 90 150 270, entry:  true true true false)] name[12]: main Ref:  Pronunciation: Destination: Four
    [3]: 17 0  1 (lon:1.001438, lat:0.99964)  Duration: 0 Distance: 0 Geometry: 4 5 Exit: 0 Mode: 1
    Intersections: 1 [(Lanes: 0 255 [0,18446744073709551615] bearings: 270, entry:  true)] name[12]: main Ref:  Pronunciation: Destination: Four
```

When I run this test, the steps getting past to processing are the above. It seems that `two`/`three` are not even in the correct locations to begin with.

MoKob:
Capturing from chat:

we have to check if such a modelling as we see here is reasonable. We see a switch of destination along a road and without a `turn`. The destinations announced are the ones of the actual turns. At `b`, there is no turn so we don't announce `two` as a destination.

Issue #3085, opened 2016-10-18
Roundabout Handler Seems to Break / Assertion on Demo Server


MoKob:
We are seeing an assertion on the demo server:

```
[assert] src/engine/guidance/post_processing.cpp:265
in: void osrm::engine::guidance::{anonymous}::closeOffRoundabout(bool, std::vector<osrm::engine::guidance::RouteStep>&, std::size_t): leavesRoundabout(steps[1].maneuver.instruction) || steps[1].maneuver.instruction.type == TurnType::StayOnRoundabout || steps[1].maneuver.instruction.type == TurnType::Suppressed || steps[1].maneuver.instruction.type == TurnType::NoTurn
```
triggering https://github.com/Project-OSRM/osrm-backend/blob/master/src/engine/guidance/post_processing.cpp#L262-L265.

This route seems to trigger the issue: `https://router.project-osrm.org/route/v1/driving/20.027888,50.079578999999995;19.94622,50.068037999999994?alternatives=false&overview=full&steps=true`

Needs investigation to find the roundabout in question.

MoKob:
Turned out that this was only a broken assertion. Fix in #3098

daniel-j-h:
Context for @TheMarex - https://github.com/Project-OSRM/osrm-backend/pull/3098 isn't merged yet, that's why you're still seeing this assertion.

We should resolve this asap.

daniel-j-h:
This just landed in master.

Issue #2843, opened 2016-09-02
Lanes for Roundabouts


daniel-j-h:
In https://github.com/Project-OSRM/osrm-backend/pull/2693 we removed all lane information from roundabouts (read the pull request for details).

Now @MoKob wants to bring back lanes in roundabouts. This ticket is for tracking progress.

PR #3098, opened 2016-10-19
use-lane can occur in roundabouts as well

MoKob:
# Issue

Resolves https://github.com/Project-OSRM/osrm-backend/issues/3085

## Tasklist
 - [ ] review
 - [ ] adjust for for comments

## Requirements / Relations
Should get cherry-picked into 5.4.1 for a 5.4.2 release (as already indicated by changelog)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,6 +27,7 @@
   - Changes from 5.4.1
     - Bugfixes
       - #3032 Fixed a bug that could result in emitting `invalid` as an instruction type on sliproads with mode changes
+      - #3085 Fixed an outdated assertion that could throw without a cause for concern
 
 # 5.4.1
   - Changes from 5.4.0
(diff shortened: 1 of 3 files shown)


TheMarex:
Is this related to https://github.com/Project-OSRM/osrm-backend/issues/2843 ?

MoKob:
Kind of related. Use-Lanes is a local view that is calculated at a point where we don't know about roundabouts anymore. They are calculated no matter what we do in the end.

It's lanes on roundabouts, though. But #2843 is describing the output for our steps.

daniel-j-h:
For the record, this triggers in

https://github.com/Project-OSRM/osrm-backend/blob/865c6590fa93f8e6cbfe1cfc803904ddfe847bfe/include/engine/api/route_api.hpp#L146

but lanes are only removed from roundabouts (for now) later on in

https://github.com/Project-OSRM/osrm-backend/blob/865c6590fa93f8e6cbfe1cfc803904ddfe847bfe/include/engine/api/route_api.hpp#L153

Looks good, merging this.

Issue #3083, opened 2016-10-18
Ascii-Art into tests for better readability


MoKob:
With the improved cucumber syntax, we introduced the possibility to add ASCII art to our cucumber features (https://github.com/Project-OSRM/osrm-backend/tree/master/features/guidance).

We should make use of this to better illustrate the different tests.

**CAREFUL: Be sure only to replace, and don't add any spaces. Otherwise turn angles can be modified and thresholds can be missed**

We have many files that can be checked for their visual appearance and can be improved:
- [ ] advanced-lanes.feature
- [ ] continue.feature
- [ ] motorway.feature
- [ ] rotary.feature
- [ ] suppressed.feature
- [ ]  ~~anticipate-lanes.feature~~ already done
- [ ] dedicated-turn-roads.feature
- [ ] new-name.feature
- [ ] roundabout-bike.feature
- [ ] trimming.feature
- [ ] bicycle-sliproads.feature
- [ ] destination-signs.feature
- [ ] notification.feature
- [ ] roundabout-turn-bike.feature
- [ ] turn-lanes.feature
- [ ] bridges_and_tunnels.feature
- [ ] end-of-road.feature
- [ ] perception.feature
- [ ] roundabout-turn.feature
- [ ] turn.feature
- [ ] bugs.feature
- [ ] fork.feature
- [ ] post-processing.feature
- [ ] roundabout.feature
- [ ] collapse-detail.feature
- [ ] intersections.feature
- [ ] ramp.feature
- [ ] staggered-intersections.feature
- [ ] collapse.feature
- [ ] merge.feature
- [ ] rotary-bike.feature
- [ ] suffix-changes.feature

The goal here is turn a thing like

```
"""
a           b
"""
```

into 

```
"""
a - - - - - b
"""
```

Due to popular demand, I recommend a restriction to the following set of characters:

`,.'*<>^/\|-_+{}:;=`

emiltin:
Might be useful with a small guideline (ie examples) on what's the recommended style to use, so we don't end up with too many weird ascii chars.

danpat:
> too many weird ascii chars.

Do we support UTF-8?  There are so many possibilities:

⑂⑃↱↲↳↴↵↶↷↰⇄⇇⇈ 🚏

emiltin:
[...]

PR #3101, opened 2016-10-19
ascii art tests

ghoshkaj:
# Issue

Ascii-Art into tests for better readability #3083
## Tasklist
- [x] Draw ASCII art
- [x] Check to see they reflect tables properly
- [x] review
- [x] adjust for for comments

### DIFF
--- a/features/guidance/dedicated-turn-roads.feature
+++ b/features/guidance/dedicated-turn-roads.feature
@@ -9,13 +9,13 @@ Feature: Slipways and Dedicated Turn Lanes
         Given the node map
             """
                     e
-            a b . . c d
-                 `h .
-                   `
-                  1 `
-                    .
-                    f
-                    .
+            a b-----c d
+               `--h |
+                   ||
+                  1||
+                   ||
+                   `f
+                    |
                     g
             """
 
@@ -41,11 +41,11 @@ Feature: Slipways and Dedicated Turn Lanes
             """
                     e
             a-b-----c-------------------------d
-                 `h |
+               `--h |
                    ||
                   1||
-                   `|
-                    f
+                   ||
+                   `f
                     |
                     g
             """
@@ -72,18 +72,18 @@ Feature: Slipways and Dedicated Turn Lanes
         Given the node map
             """
                     e
-            a b     c d
-                  h
-
-
-
-
-
-
-
+            a-b----]c[-d
+               `--h |
+                   \|
+                    |
+                    |
+                    |
+                    |
+                    |
+                    |
                     f
-
-
+                    |
+                    |
                     g
             """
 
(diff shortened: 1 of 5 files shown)


TheMarex:
@ghoshkaj what about merging this in as-is because this is already a good first step?

features/guidance/dedicated-turn-roads.feature
```diff
@@ -71,19 +71,19 @@ Feature: Slipways and Dedicated Turn Lanes
...
+            a-b-----c-d |
+               `--h     |
+                   \    /
+                    |  /
+                    | /
+                    |/
+                    |
+                    |
+                    |
                     f
```
MoKob: while technically correct (e does not connect to `c` due to a bridge, the road looks strange with this angles.

how about adding `] c [` style chars here? to indicate elevation?

features/guidance/dedicated-turn-roads.feature
```diff
@@ -100,15 +100,15 @@ Feature: Slipways and Dedicated Turn Lanes
...
             """
-                    e
-            a b     c d
-                  h
-
-
-
+                    e---.
+            a-b-----c-d |
+               `--h     |
```
MoKob: same here (see above)

features/guidance/dedicated-turn-roads.feature
```diff
@@ -349,31 +349,30 @@ Feature: Slipways and Dedicated Turn Lanes
...
+                          \                               |
+                           \                              |
+                            \                             |
+                             \                            |
+                              \                           |
+                               \                          |
+                                \                          \
+                                 \                          \
+                                  \                          |
+                                   \                      1 /
```
MoKob: no need to go around `1`. `1` is not part of the road, but a waypoint on it.

ghoshkaj:
@daniel-j-h yes I'll get this rebased and in a state ready to merge asap.

features/guidance/dedicated-turn-roads.feature
ghoshkaj: sounds good!

features/guidance/dedicated-turn-roads.feature
ghoshkaj: 👍

features/guidance/dedicated-turn-roads.feature
ghoshkaj: 👍

ghoshkaj:
Could someone review this please? Then I'll merge!

MoKob:
Looking at it now @ghoshkaj

features/guidance/dedicated-turn-roads.feature
MoKob: This should not have a detour for e--f. They should go straight under `c`. Sorry for not being clear here
A better description of the case would be
```
                    e
                   \_/
a-b - - - - - - - - c - d
                   /-\
                    |
```

Note that changing the location of `e` does not influence the case (a lot) since only one segment gets a bit longer. turn angles remain the same. It is not a change that is save in general, but here it is.
Also below.

TheMarex:
@MoKob @ghoshkaj can I mark this as ready to merge or are there still open issues?

ghoshkaj:
I'm done with it if @MoKob is :+1: with it.

PR #3110, opened 2016-10-19
fix changelog to not include 5.5.1 ahead of time

MoKob:
# Issue

Making the changelog correct again. 5.5.0 is still in the works.


TheMarex:
I think it is fine to push doc/changelog fixes with `[skip ci] ...`. Looks good to me.

PR #3075, opened 2016-10-18
Removes constant speed offset from calibration two years ago, resolves #3023.

daniel-j-h:
For #3023.

## Tasklist
 - [x] Why are unrelated turn lane and name tests failing? Investigate!
 - [ ] review
 - [ ] adjust for for comments

Our fine-tuned profiles are better in modelling real speed by now. This
constant offset is no longer needed. We still scale maxspeed, though.

Issue #3121, opened 2016-10-20
Guidance Preprocessing invalidly uses `.distance` for distance


MoKob:
Some functions (turn discovery and sliproad handler) use `.distance` of our node-based-edges in their computation to find out about the length of segments.

This field is only named distance, though, and actually refers to the routing weight. Doing so, we are actually dependent on the speed of the road, which is undesirable.
After all, a sliproad remains a sliproad, even if there is a traffic jam around it.

We see failures due to this in https://github.com/Project-OSRM/osrm-backend/pull/3075.

daniel-j-h:
I was just about to close this, but over in https://github.com/Project-OSRM/osrm-backend/pull/3075 the test you added is failing:

https://github.com/Project-OSRM/osrm-backend/blob/a3719696fbea3e165bcccd6f44327d42a46a0dad/features/guidance/dedicated-turn-roads.feature#L64

with

    depart,on ramp left,arrive

(also https://github.com/Project-OSRM/osrm-backend/pull/3075 accidentally got closed by Github since the commit message in the pull request for this issue says it's resolving it)

daniel-j-h:
For the record: this is the test which should confirm we're actually making a turn instead of a merge based on the maxspeed changes.

MoKob:
The issue in #3075 that remains is not due to this issue. The actual issue is solved over in https://github.com/Project-OSRM/osrm-backend/pull/3143. I also pushed an update to the scenario you see errors in.

PR #3122, opened 2016-10-20
actually calculate distance instead of using `.distance`

MoKob:
# Issue

resolves, https://github.com/Project-OSRM/osrm-backend/issues/3121

## Tasklist
 - [x] review
 - [x] adjust for for comments

## Requirements / Relations
Should fix https://github.com/Project-OSRM/osrm-backend/pull/3075 problems


TheMarex:
Looks good to me. :+1:

daniel-j-h:
Ouch - thanks @MoKob for fixing this!

What about changing the name for the `.distance` member attribute.

TheMarex:
@daniel-j-h already done in the #1614

Issue #3062, opened 2016-10-14
Travis cache still growing


TheMarex:
@daniel-j-h reported that our cache breached the 2GB limit again. That means there must be something that does not get cleaned up. We can investigate this by sshing into a travis box.

daniel-j-h:
For the record the Travis cache for master grew to 3 GB again - timespan: ~6 days.

daniel-j-h:
@TheMarex is there a reason for keeping this open?

Caches for master seem to be stable at roughly 2 GB - which is more or less what we expect with 8 builds and a ccache limit of 8 * 256 MB + caching Cucumber datasets. I'm surprised it's not higher actually.

TheMarex:
No, if this works now that is great. :+1:

PR #3132, opened 2016-10-20
Limits the object cache size to 256M

daniel-j-h:
For https://github.com/Project-OSRM/osrm-backend/issues/3062. Trying if that helps.

We need a bit time to see how this will help: idea is to merge this, clear the Travis caches multiple times today (since multiple builds in in-flight and will restore the caches), and then see how this goes.


daniel-j-h:
Works on Travis - merging.

PR #3137, opened 2016-10-20
Supports confidence value for updating segment speeds, resolves #3129.

### LINKED ISSUE #3129: Add confidence value to the osrm-contract speed values
Currently osrm-contract expects a speed file to follow:

```
from_node_id,to_node_id,speed_kmh[,optional comment]
```

We want to change this to:

```
from_node_id,to_node_id,speed_kmh,confidence[,optional comment]
```

This would allow for weighting the input speed value by confidence when overwriting the old value.

The update function [here](https://github.com/Project-OSRM/osrm-backend/blob/master/src/contractor/contractor.cpp#L789) and [here](https://github.com/Project-OSRM/osrm-backend/blob/master/src/contractor/contractor.cpp#L92) would need to change to something similar to:

`new_speed = (1-confidence) * old_speed + confidence * pancake_winner_speed;`

/cc @danpat @benjamintd @lily-chai

daniel-j-h:
For #3129: support confidence based updates.

cc @TheMarex 
## Tasklist
- [x] confidence -> smoothing factor
- [x] throw on out of range 0,1
- [ ] cucumber scenarios for different confidence values
- [ ] review
- [ ] adjust for for comments

### DIFF
--- a/features/car/traffic_speeds.feature
+++ b/features/car/traffic_speeds.feature
@@ -26,14 +26,15 @@ Feature: Traffic - speeds
         Given the profile "testbot"
         Given the extract extra arguments "--generate-edge-lookup"
         Given the contract extra arguments "--segment-speed-file {speeds_file}"
+        # from,to,speed,smoothing factor
         Given the speed file
         """
-        1,2,0
-        2,1,0
-        2,3,27
-        3,2,27
-        1,4,27
-        4,1,27
+        1,2,0,1
+        2,1,0,1
+        2,3,27,1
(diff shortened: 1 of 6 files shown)


daniel-j-h:
Strictly speaking we're changing behavior here which could affect clients. But then again the wildcard support was never documented or intended for users to know about see 

https://github.com/Project-OSRM/osrm-backend/wiki/Traffic

daniel-j-h:
Stale. We didn't prioritized this and it never landed.

PR #3140, opened 2016-10-21
Deduplicate code for loading profile properties

karenzshea:
# Issue

Hits one item on https://github.com/Project-OSRM/osrm-backend/issues/2989

I've moved the code for loading the `osrm.profile_properties` file into shared memory into the shared header `storage/io.hpp` and used it in datastore and the internal datafacade.

## Tasklist
 - [ ] add regression / cucumber cases?
 - [ ] review
 - [ ] adjust for for comments

## Requirements / Relations
cc @chaupow @TheMarex @danpat

### DIFF
--- a/include/storage/io.hpp
+++ b/include/storage/io.hpp
@@ -15,6 +15,16 @@ namespace storage
 namespace io
 {
 
+inline std::size_t readPropertiesSize() { return 1; }
+
+template <typename PropertiesT>
+inline void readProperties(boost::filesystem::ifstream &properties_stream,
+                           PropertiesT properties[],
+                           std::size_t PropertiesSize)
+{
+    properties_stream.read(properties, PropertiesSize);
+}
+
 #pragma pack(push, 1)
 struct HSGRHeader
 {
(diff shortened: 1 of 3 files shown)


include/storage/io.hpp
```diff
@@ -15,6 +15,16 @@ namespace storage
...
 {
 
+inline std::size_t readPropertiesSize() { return 1; }
+
+template <typename PropertiesT>
+inline void readProperties(boost::filesystem::ifstream &properties_stream,
+                           PropertiesT properties[],
+                           std::size_t PropertiesSize)
+{
+    properties_stream.read(properties, PropertiesSize);
```
daniel-j-h: I don't think ProfileProperties is one byte - the `1` in the SetBlockSize function before were not bytes.

karenzshea: We're planning to rename the generic `getTypeSize()` functions to something like `getTypeCount()` to be more accurate

karenzshea:
Superseded by a combined PR in https://github.com/Project-OSRM/osrm-backend/pull/3148

PR #3143, opened 2016-10-21
fix turn scenarios

MoKob:
# Issue

Some of our tests use invalid turn restrictions. This PR updates all turn restrictions in the guidance part of the code to work with OSRM. In OSRM, we cannot start a turn restriction in the middle of a way (e.g. `abc` and `bd` we can't specify a restriction `abc` via `b` to `bd`, since both `ab` and `cb` could be seen as `in` for the restriction).

## Tasklist
 - [x] review
 - [x] adjust for for comments

## Requirements / Relations

### DIFF
--- a/features/guidance/collapse.feature
+++ b/features/guidance/collapse.feature
@@ -635,12 +635,14 @@ Feature: Collapse
             | bd    | road  | yes    | primary   |
             | bc    | road  | yes    | primary   |
             | de    | road  | yes    | primary   |
-            | fdcg  | cross | no     | secondary |
+            | fd    | cross | no     | secondary |
+            | dc    | cross | no     | secondary |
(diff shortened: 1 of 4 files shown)


TheMarex:
> In OSRM, we cannot start a turn restriction in the middle of a way

Are those turn restrictions still correctly formed in terms of the OSM guidelines or is this just a limitation of our implementation?

MoKob:
Well, OSM does not specify exactly how a turn restriction is specified. But since `no_right_turn` is not actually required to check `right`, I'd say that it is invalid OSM.
Otherwise
```
abc
 d
```
with no-right-turn would prohibit all turns if `abc` via `b` to `d` is used.

TheMarex:
Looks good to merge, but needs a rebase.

Issue #3033, opened 2016-10-10
Segfault with foot/bicycle profile in 5.3.0 upwards


woodpeck:
I observed a segmentation violation when trying to osrm-extract Europe for foot.lua or bicycle.lua with all OSRM versions starting with commit 2b466b2fb29c416149b4d881c151349218a63e1e.

The issue is detailed in https://lists.openstreetmap.org/pipermail/osrm-talk/2016-October/001334.html, I'm giving a tl;dr here:
- The problem is always reported as "double free or corruption (fasttop)" and leads to a segfault in osrm-extract on Ubuntu 16.04.
- The problem affects all OSRM versions from commit 2b466b2fb29c416149b4d881c151349218a63e1e on (this means v5.2.7 is the last-known-good release). I read through that commit and couldn't immediately see anything that looks broken.
- The problem does not occur with the "car" profile but it does occur with the "bicycle" and "foot" profiles.
- The problem does not occur with data extracts smaller than ~ 2 GB but it does occur with larger ones. Sadly that made it impossible for me to find a very small input file that would provoke the crash, and any
  "valgrind" debugging was out of the question.

It appears that @oxidase confirmed the problem.

oxidase:
@woodpeck the problem at https://github.com/Project-OSRM/osrm-backend/blob/master/src/extractor/extraction_containers.cpp#L87-L90 that can be a data-dependent issue. At the moment i am running bisect for edges of the europe foot profile. The current range is 12637448 edges with ~ 6 hours per trial, so i would need ~ 6 days to find a hint.

oxidase:
@woodpeck i have a proposal in fix/3033 . Please could you try the branch and say whether it fixes segfaults for your case. I have found a pretty complicated set of preconditions for the segfault in my case.

daniel-j-h:
Ping - @woodpeck any updates on your side? If this solves your issue feel free to close this ticket.

woodpeck:
Sorry I was slow with this, but have now successfully managed to run osrm-extract under the same conditions that previously caused the crash and it worked. Thank you for following up on this!

PR #3145, opened 2016-10-21
Make thread-locked access to stxxl containers

oxidase:
# Issue

From stxxl FAQ: "you should not share a data structure between
threads (without implementing proper locking yourself)."
The access to name_char_data can be implicitly parallelized
if _GLIBCXX_PARALLEL is defined and invalidate local-thread iterators, but
in CmpEdgeByInternalSourceTargetAndName there is no guaranty to have valid iterators.

Preconditions for the UB:
- defined _GLIBCXX_PARALLEL and fopenmp  
- __gnu_parallel::__get_max_threads() > 1
- all_edges_list > __gnu_parallel::_Settings::get().sort_minimal_n 
- number of touched pages in name_char_data > STXXLNameCharData::PagerType::n_pages
- unlucky execution sequence that is almost sure for europe-latest

To review:
- drop access to name_char_data, but this will lead to a random edge name selection
- make thread-safe iterators, but i can not find better solution than keep a reference to a mutex, because mutex is non-copyable and non-movable, but Comparable concept requires copyable predicate.

## Tasklist
 - [x] review
 - [x] adjust for for comments

## Requirements / Relations
 #3033

### DIFF
--- a/src/extractor/extraction_containers.cpp
+++ b/src/extractor/extraction_containers.cpp
@@ -20,6 +20,7 @@
 
 #include <chrono>
 #include <limits>
+#include <mutex>
 
 namespace
 {
@@ -67,10 +68,13 @@ struct CmpEdgeByInternalSourceTargetAndName
         if (lhs.result.source != rhs.result.source)
             return lhs.result.source < rhs.result.source;
 
+        if (lhs.result.source == SPECIAL_NODEID)
+            return false;
+
         if (lhs.result.target != rhs.result.target)
             return lhs.result.target < rhs.result.target;
 
-        if (lhs.result.source == SPECIAL_NODEID)
+        if (lhs.result.target == SPECIAL_NODEID)
             return false;
 
         if (lhs.result.name_id == rhs.result.name_id)
@@ -82,6 +86,7 @@ struct CmpEdgeByInternalSourceTargetAndName
         if (rhs.result.name_id == EMPTY_NAMEID)
             return true;
 
+        std::lock_guard<std::mutex> lock(mutex);
         BOOST_ASSERT(!name_offsets.empty() && name_offsets.back() == name_data.size());
         const oe::ExtractionContainers::STXXLNameCharData::const_iterator data = name_data.begin();
         return std::lexicographical_compare(data + name_offsets[lhs.result.name_id],
@@ -93,6 +98,7 @@ struct CmpEdgeByInternalSourceTargetAndName
     value_type max_value() { return value_type::max_internal_value(); }
     value_type min_value() { return value_type::min_internal_value(); }
 
+    std::mutex& mutex;
     const oe::ExtractionContainers::STXXLNameCharData &name_data;
     const oe::ExtractionContainers::STXXLNameOffsets &name_offsets;
 };
@@ -433,9 +439,10 @@ void ExtractionContainers::PrepareEdges(ScriptingEnvironment &scripting_environm
     // Sort edges by start.
     std::cout << "[extractor] Sorting edges by renumbered start ... " << std::flush;
     TIMER_START(sort_edges_by_renumbered_start);
+    std::mutex name_data_mutex;
     stxxl::sort(all_edges_list.begin(),
                 all_edges_list.end(),
-                CmpEdgeByInternalSourceTargetAndName{name_char_data, name_offsets},
+                CmpEdgeByInternalSourceTargetAndName{name_data_mutex, name_char_data, name_offsets},
                 stxxl_memory);
     TIMER_STOP(sort_edges_by_renumbered_start);
     std::cout << "ok, after " << TIMER_SEC(sort_edges_by_renumbered_start) << "s" << std::endl;
--- a/include/extractor/internal_extractor_edge.hpp
+++ b/include/extractor/internal_extractor_edge.hpp
@@ -113,7 +113,7 @@ struct InternalExtractorEdge
     {
         return InternalExtractorEdge(MAX_OSM_NODEID,
                                      MAX_OSM_NODEID,
-                                     0,
+                                     SPECIAL_NODEID,
                                      WeightData(),
                                      false,
                                      false,


src/extractor/extraction_containers.cpp
```diff
@@ -433,9 +439,10 @@ void ExtractionContainers::PrepareEdges(ScriptingEnvironment &scripting_environm
     // Sort edges by start.
     std::cout << "[extractor] Sorting edges by renumbered start ... " << std::flush;
     TIMER_START(sort_edges_by_renumbered_start);
+    std::mutex name_data_mutex;
     stxxl::sort(all_edges_list.begin(),
                 all_edges_list.end(),
-                CmpEdgeByInternalSourceTargetAndName{name_char_data, name_offsets},
+                CmpEdgeByInternalSourceTargetAndName{name_data_mutex, name_char_data, name_offsets},
```
TheMarex: So `stxxl::sort` is parallelized? Interesting, didn't know that. Might be worth a comment, since on the calling code there is no indication.

TheMarex:
Using the mutex is fine. The critical section will only be entered for every few comparisons, so I don't expect a serious performance impact.

src/extractor/extraction_containers.cpp
daniel-j-h: => https://github.com/stxxl/stxxl/blob/master/include/stxxl/bits/algo/sort.h

oxidase:
@TheMarex just checked in germany-latesst for car profile 64 hits out of 729750295  calls or 8.77012e-06%, so should be quite rare

bvbever:
When I run in 5.4.0 The extractor gives me for processing Europe the following error


`[info] Parsing finished after 4496.31 seconds`
`[info] Raw input contains 1614199431 nodes, 200970520 ways, and 2761204 relations`
`[extractor] Sorting used nodes        ... ok, after 106.832s`
`[extractor] Erasing duplicate nodes   ... ok, after 208.667s`
`[extractor] Sorting all nodes         ... ok, after 1030.95s`
`[extractor] Building node id map      ... ok, after 1199.18s`
`[extractor] setting number of nodes   ... ok`
`[extractor] Confirming/Writing used nodes     ... ok, after 959.825s`
`[info] Processed 341628702 nodes`
`[extractor] Sorting edges by start    ... ok, after 683.963s`
`[extractor] Setting start coords      ... ok, after 1927.18s`
`[extractor] Sorting edges by target   ... ok, after 657.574s`
`[extractor] Computing edge weights    ... ok, after 2164.88s`
`[extractor] Sorting edges by renumbered start ... [assert] /home/routeyou/osrm-5.4.1/src/extractor/extraction_containers.cpp:85`
`in: bool {anonymous}::CmpEdgeByInternalSourceTargetAndName::operator()(const value_type&, const value_type&) const: !name_offsets.empty() && name_offsets.back() == name_data.size()
`terminate called without an active exception`
`Aborted`

Has this something to do with this ? 
When I run it on small datasets like belgium, I do not get this error.

`

oxidase:
@bvbever your output looks like the issue,  please check  with 7753845f5c99a843ca84283eb64ea8cd647fc255 that was recently merged to master

bvbever:
@oxidase ok, I run it yesterday with the master version and got this error: 

`[info] Parsing finished after 4500.8 seconds`
`[info] Raw input contains 1614199431 nodes, 200970520 ways, and 2761204 relations`
`[extractor] Sorting used nodes        ... ok, after 106.922s`
`[extractor] Erasing duplicate nodes   ... ok, after 207.922s`
`[extractor] Sorting all nodes         ... ok, after 1031.87s`
`[extractor] Building node id map      ... ok, after 1188.95s`
`[extractor] setting number of nodes   ... ok`
`[extractor] Confirming/Writing used nodes     ... ok, after 961.004s`
`[info] Processed 341628702 nodes`
`[extractor] Sorting edges by start    ... ok, after 684.085s`
`[extractor] Setting start coords      ... ok, after 1925.2s`
`[extractor] Sorting edges by target   ... ok, after 656.908s`
`[extractor] Computing edge weights    ... ok, after 2170.69s`
`[extractor] Sorting edges by renumbered start ... [STXXL-ERRMSG] READ request submitted for a BID with a pending WRITE request`
`[STXXL-ERRMSG] READ request submitted for a BID with a pending WRITE request`
`[assert] /home/routeyou/osrm-5.4.1/src/extractor/extraction_containers.cpp:85`
`in: bool {anonymous}::CmpEdgeByInternalSourceTargetAndName::operator()(const value_type&, const value_type&) const: !name_offsets.empty() && name_offsets.back() == name_data.size()`
`terminate called without an active exception`
`Aborted`

oxidase:
@bvbever i see now that it is another issue, but i can not reproduce it locally. 
I could imagine two reasons: 
* access to non-flushed stxxl vectors
* offset value overflow 

please can you try branch https://github.com/Project-OSRM/osrm-backend/tree/test/assertion_in_compare  that contains be76ab3f8558dcbdd4c3bbe7f6cb1ba3364d940c  and post output result from "Sorting edges by renumbered start"  like
```
[extractor] Sorting edges by renumbered start ... name_offsets.size() = 148537, name_offsets.back() = 586949, name_data.size() = 586949
```

bvbever:
@oxidase I will give you the result today.

bvbever:
@oxidase 
the result is
[extractor] Sorting edges by renumbered start ... name_offsets.size() = 159707417, name_offsets.back() = 738862466, name_data.size() = 738862466

oxidase:
@bvbever if it is only one output line with sizes and extraction passes till the end, I would assume that stxxl vectors must be flushed before PrepareData call.

bvbever:
@oxidase , ok, Is this something that needs to be done while processing with OSRM ? Or that I have to do ones (and how) ?

oxidase:
@bvbever i have created a separate PR  #3285 because it is not related to the PR issue.

Issue #3088, opened 2016-10-18
Windows builds are broken


TheMarex:
```
C:\projects\osrm\osrm-deps\boost\include\boost-1_59\boost/mpl/eval_if.hpp(41): error C2504: 'boost::result_of<const UnaryFunc (unsigned int &)>': base class undefined [C:\projects\osrm\build\ENGINE.vcxproj]
```

This was broken by #3015.

oxidase:
@TheMarex  If #include <boost/range/adaptor/transformed.hpp> that is in assemble_leg.hpp
is included after
#include "engine/data_watchdog.hpp"
#include "storage/shared_barriers.hpp"
that include
#include <boost/detail/workaround.hpp>
#include <boost/interprocess/detail/workaround.hpp>
than something is going wrong, probably with forced inlines.

I can not reproduce the issue with 1.60 http://rextester.com/live/SZDL55018
I would suggest just to update boost in osrm-deps-win-x64-14.0.7z, and  there is still  a link issue with libboost_zlib-vc140-mt-1_59.lib

oxidase:
@TheMarex just figured out the problem: in https://mapbox.s3.amazonaws.com/windows-builds/windows-build-deps/osrm-deps-win-x64-14.0.7z in file osrm-deps/boost/include/boost-1_59/boost/thread/lockable_traits.hpp contains `#define BOOST_NO_SFINAE` that breaks Boost.Range
It is a local modification, https://github.com/boostorg/thread/blob/develop/include/boost/thread/lockable_traits.hpp does not contain the line

BergWerkGIS:
Yep, can/will fix. 😄 

Deps package gets created via [`windows-builds/scripts/build_osrm_deps.bat`](https://github.com/mapbox/windows-builds/blob/master/scripts/build_osrm_deps.bat)

BergWerkGIS:
Updated `osrm-deps-win-x64-14.0.7z` to use `boost@1.61` without any modifications.

Now there's a different set of errors:
https://ci.appveyor.com/project/TheMarex/osrm-backend/build/1.0.3762

compared to the previous run:
https://ci.appveyor.com/project/TheMarex/osrm-backend/build/1.0.3761

oxidase:
@BergWerkGIS The problem is in missing commit https://github.com/boostorg/config/commit/ea80234d5f1fff82573f7180f43b6600572fecc4
[...]

PR #3146, opened 2016-10-21
Try to fix msvc build error "base class undefined"

oxidase:
# Issue

Try to fix msvc build error "base class undefined":
if `#include <boost/range/adaptor/transformed.hpp>` that is in assemble_leg.hpp
is included after
`#include "engine/data_watchdog.hpp"`
`#include "storage/shared_barriers.hpp"`
that include
`#include <boost/detail/workaround.hpp>`
`#include <boost/interprocess/detail/workaround.hpp>`
than something is going wrong, probably with forced inlines.

I can not reproduce the issue with 1.60 http://rextester.com/live/SZDL55018

Also fixed msvc related boost linking, because engine requires explicit linking of zlib, chrono and date_time libraries.

## Tasklist
 - [x] review
 - [x] adjust for for comments

Related issue #3088

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -67,7 +67,7 @@ add_custom_target(FingerPrintConfigure ALL ${CMAKE_COMMAND}
   COMMENT "Configuring revision fingerprint"
   VERBATIM)
 
-set(BOOST_COMPONENTS date_time filesystem iostreams program_options regex system thread unit_test_framework)
+set(BOOST_COMPONENTS date_time chrono filesystem iostreams program_options regex system thread unit_test_framework)
 
 configure_file(
   ${CMAKE_CURRENT_SOURCE_DIR}/include/util/version.hpp.in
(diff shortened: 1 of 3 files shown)


include/engine/engine.hpp
```diff
@@ -20,6 +18,9 @@
 #include "engine/status.hpp"
 #include "util/json_container.hpp"
 
+#include "engine/data_watchdog.hpp"
```
TheMarex: Why were these headers moved?

include/engine/data_watchdog.hpp
```diff
@@ -8,7 +8,6 @@
 #include "storage/shared_memory.hpp"
 
 #include <boost/interprocess/sync/named_upgradable_mutex.hpp>
-#include <boost/thread/lock_types.hpp>
```
TheMarex: Was this superfluous?

CMakeLists.txt
```diff
@@ -337,23 +337,22 @@ include_directories(SYSTEM ${OSRM_INCLUDE_PATHS})
...
 set(BOOST_BASE_LIBRARIES
    ${Boost_DATE_TIME_LIBRARY}
+   ${Boost_CHRONO_LIBRARY}
    ${Boost_FILESYSTEM_LIBRARY}
    ${Boost_IOSTREAMS_LIBRARY}
    ${Boost_THREAD_LIBRARY}
    ${Boost_SYSTEM_LIBRARY})
 
 set(BOOST_ENGINE_LIBRARIES
-   ${Boost_FILESYSTEM_LIBRARY}
```
TheMarex: So we never had a dependency on thread, system and iostreams in `ENGINE`?

TheMarex:
Thanks for looking into this! :+1:

include/engine/data_watchdog.hpp
oxidase: no, it is still needed. just a left over from bisectioning

include/engine/engine.hpp
oxidase: This is actually the workaround: interprocess headers must be included after Boost.Range. It is somehow related to inclusion of /usr/include/boost/interprocess/detail/workaround.hpp

If someone can update boost in osrm-deps-win-x64-14.0.7z  to 1.61 or 1.62 probably it will not be required.

daniel-j-h: cc @BergWerkGIS

PR #3112, opened 2016-10-19
Move loading of .nodes and .edges to io.hpp

chaupow:
# Issue

Working on data loading refactoring captured here: 
https://github.com/Project-OSRM/osrm-backend/issues/2989

This issue is working on `.nodes` and `.edges` files

## Tasklist
 - [ ]  wait for comment on the TODO in `io.hpp`
 - [x] wait for https://github.com/Project-OSRM/osrm-backend/pull/3074 to be merged
 - [x] wait for https://github.com/Project-OSRM/osrm-backend/pull/3090 to be merged
 - [ ] review
 - [ ] adjust for comments

/cc @karenzshea @TheMarex  @miccolis @danpat

PR #3148, opened 2016-10-21
Refactor file loading and move into io.hpp

chaupow:
# Issue

Working on data loading refactoring captured here: 
https://github.com/Project-OSRM/osrm-backend/issues/2989

This issue is working on 

* `.nodes` and `.edges` files
* `.datasource_names` and `.datasource_indexes` files
* `.ramIndex` files
* `.profile_properties`

## Tasklist
 - [x] Decide on plain pointer `*` or `[]` bracket coding style
 - [x] review
 - [x] adjust for for comments

### DIFF
--- a/include/engine/datafacade/internal_datafacade.hpp
+++ b/include/engine/datafacade/internal_datafacade.hpp
@@ -123,8 +123,8 @@ class InternalDataFacade final : public BaseDataFacade
         {
             throw util::exception("Could not open " + properties_path.string() + " for reading.");
         }
-        auto PropertiesSize = storage::io::readPropertiesSize();
-        storage::io::readProperties(in_stream, &m_profile_properties, PropertiesSize);
+        const auto properties_size = storage::io::readPropertiesCount();
+        storage::io::readProperties(in_stream, &m_profile_properties, properties_size);
     }
 
     void LoadLaneTupleIdPairs(const boost::filesystem::path &lane_data_path)
@@ -151,7 +151,7 @@ class InternalDataFacade final : public BaseDataFacade
             throw util::exception("Could not open " + timestamp_path.string() + " for reading.");
         }
 
-        auto timestamp_size = storage::io::readTimestampSize(timestamp_stream);
+        const auto timestamp_size = storage::io::readNumberOfBytes(timestamp_stream);
         m_timestamp.resize(timestamp_size);
         storage::io::readTimestamp(timestamp_stream, &m_timestamp.front(), timestamp_size);
     }
@@ -164,7 +164,7 @@ class InternalDataFacade final : public BaseDataFacade
             throw util::exception("Could not open " + hsgr_path.string() + " for reading.");
         }
 
-        auto header = storage::io::readHSGRHeader(hsgr_input_stream);
+        const auto header = storage::io::readHSGRHeader(hsgr_input_stream);
         m_check_sum = header.checksum;
 
         util::ShM<QueryGraph::NodeArrayEntry, false>::vector node_list(header.number_of_nodes);
@@ -190,10 +190,10 @@ class InternalDataFacade final : public BaseDataFacade
             throw util::exception("Could not open " + nodes_file_path.string() + " for reading.");
         }
 
-        std::uint32_t number_of_coordinates = storage::io::readNodesSize(nodes_input_stream);
+        const auto number_of_coordinates = storage::io::readElementCount(nodes_input_stream);
         m_coordinate_list.resize(number_of_coordinates);
         m_osmnodeid_list.reserve(number_of_coordinates);
-        storage::io::readNodesData(
+        storage::io::readNodes(
             nodes_input_stream, m_coordinate_list.data(), m_osmnodeid_list, number_of_coordinates);
 
         boost::filesystem::ifstream edges_input_stream(edges_file_path, std::ios::binary);
@@ -201,7 +201,7 @@ class InternalDataFacade final : public BaseDataFacade
         {
             throw util::exception("Could not open " + edges_file_path.string() + " for reading.");
         }
-        auto number_of_edges = storage::io::readEdgesSize(edges_input_stream);
+        const auto number_of_edges = storage::io::readElementCount(edges_input_stream);
         m_via_geometry_list.resize(number_of_edges);
         m_name_ID_list.resize(number_of_edges);
         m_turn_instruction_list.resize(number_of_edges);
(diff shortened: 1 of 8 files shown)


chaupow:
@MoKob regarding your comment here https://github.com/Project-OSRM/osrm-backend/pull/3112#pullrequestreview-5028128.
There's some discussion whether we want the `*` or `[]` coding style. to some of us the asterisk feels more intuitive. to some `[]` is more intuitive. but we don't have strong opinions so we can do either style.

your final comment?

include/storage/io.hpp
```diff
@@ -39,44 +60,161 @@ inline HSGRHeader readHSGRHeader(boost::filesystem::ifstream &input_stream)
...
+        BOOST_ASSERT(coordinate_list[i].IsValid());
+    }
+}
+
+
+
+// Reads datasource_indexes
+// Needs to be called after readElementCount() to get the correct offset in the stream
+inline void readDatasourceIndexes(boost::filesystem::ifstream &datasource_indexes_input_stream,
+                                  uint8_t datasource_buffer[],
```
daniel-j-h: Why uint8*?

include/storage/io.hpp
```diff
@@ -39,44 +60,161 @@ inline HSGRHeader readHSGRHeader(boost::filesystem::ifstream &input_stream)
...
+{
+    std::vector<char> names;
+    std::vector<std::uint32_t> offsets;
+    std::vector<std::uint32_t> lengths;
+};
+inline DatasourceNamesData
+readDatasourceNames(boost::filesystem::ifstream &datasource_names_input_stream)
+{
+    DatasourceNamesData datasource_names_data;
+    if (datasource_names_input_stream)
```
daniel-j-h: Why the check here? It will only be false if the stream is bad.

include/storage/io.hpp
```diff
@@ -39,44 +60,161 @@ inline HSGRHeader readHSGRHeader(boost::filesystem::ifstream &input_stream)
...
+                      std::back_inserter(datasource_names_data.names));
+        }
+    }
+    return datasource_names_data;
+}
+
+
+template <typename RTreeNodeT>
+void readRamIndex(boost::filesystem::ifstream &ram_index_input_stream,
+                 RTreeNodeT rtree_buffer[],
```
daniel-j-h: Why templates?

include/util/static_rtree.hpp
```diff
@@ -356,14 +357,11 @@ class StaticRTree
         }
         boost::filesystem::ifstream tree_node_file(node_file, std::ios::binary);
 
-        std::uint32_t tree_size = 0;
-        tree_node_file.read((char *)&tree_size, sizeof(std::uint32_t));
+
+        auto tree_size = storage::io::readElementCount<std::uint32_t>(tree_node_file);
```
daniel-j-h: const

src/storage/storage.cpp
```diff
@@ -232,8 +230,7 @@ Storage::ReturnCode Storage::Run(int max_wait)
         throw util::exception("Could not open " + config.edges_data_path.string() +
                               " for reading.");
     }
-    unsigned number_of_original_edges = 0;
-    edges_input_stream.read((char *)&number_of_original_edges, sizeof(unsigned));
+    auto number_of_original_edges = io::readElementCount<std::uint32_t>(edges_input_stream);
```
daniel-j-h: const

src/storage/storage.cpp
```diff
@@ -269,20 +266,21 @@ Storage::ReturnCode Storage::Run(int max_wait)
     // load rsearch tree size
     boost::filesystem::ifstream tree_node_file(config.ram_index_path, std::ios::binary);
 
-    uint32_t tree_size = 0;
-    tree_node_file.read((char *)&tree_size, sizeof(uint32_t));
+    auto tree_size = io::readElementCount<std::uint32_t>(tree_node_file);
```
daniel-j-h: const

src/storage/storage.cpp
```diff
@@ -269,20 +266,21 @@ Storage::ReturnCode Storage::Run(int max_wait)
...
     shared_layout_ptr->SetBlockSize<RTreeNode>(SharedDataLayout::R_SEARCH_TREE, tree_size);
 
-    // load profile properties
-    shared_layout_ptr->SetBlockSize<extractor::ProfileProperties>(SharedDataLayout::PROPERTIES, 1);
+    // allocate space in shared memory for profile properties
+    std::size_t PropertiesSize = io::readPropertiesSize();
```
daniel-j-h: const

daniel-j-h: everywhere in this pr - also below..

src/storage/storage.cpp
```diff
@@ -348,46 +345,38 @@ Storage::ReturnCode Storage::Run(int max_wait)
     std::uint64_t number_of_compressed_datasources = 0;
     if (geometry_datasource_input_stream)
     {
-        geometry_datasource_input_stream.read(
-            reinterpret_cast<char *>(&number_of_compressed_datasources),
-            sizeof(number_of_compressed_datasources));
+        number_of_compressed_datasources =
+            io::readElementCount<std::uint64_t>(geometry_datasource_input_stream);
```
daniel-j-h: Why is this element count 64 bit but the other ones above 32 bit? This is so easy to get wrong. In case of a mismatch we probably will crash and burn hard.

src/storage/storage.cpp
```diff
@@ -348,46 +345,38 @@ Storage::ReturnCode Storage::Run(int max_wait)
...
     shared_layout_ptr->SetBlockSize<char>(SharedDataLayout::DATASOURCE_NAME_DATA,
-                                          m_datasource_name_data.size());
-    shared_layout_ptr->SetBlockSize<std::size_t>(SharedDataLayout::DATASOURCE_NAME_OFFSETS,
-                                                 m_datasource_name_offsets.size());
-    shared_layout_ptr->SetBlockSize<std::size_t>(SharedDataLayout::DATASOURCE_NAME_LENGTHS,
-                                                 m_datasource_name_lengths.size());
+                                          datasource_names_data.names.size());
+    shared_layout_ptr->SetBlockSize<std::uint32_t>(SharedDataLayout::DATASOURCE_NAME_OFFSETS,
+                                                   datasource_names_data.offsets.size());
+    shared_layout_ptr->SetBlockSize<std::uint32_t>(SharedDataLayout::DATASOURCE_NAME_LENGTHS,
```
daniel-j-h: Why are those 32 bit now but were size_t before?

src/storage/storage.cpp
```diff
@@ -575,19 +564,16 @@ Storage::ReturnCode Storage::Run(int max_wait)
...
+    io::readEdges(edges_input_stream,
+                  via_geometry_ptr,
+                  name_id_ptr,
+                  turn_instructions_ptr,
+                  lane_data_id_ptr,
+                  travel_mode_ptr,
+                  entry_class_id_ptr,
+                  pre_turn_bearing_ptr,
+                  post_turn_bearing_ptr,
+                  number_of_original_edges);
```
daniel-j-h: If you accidentally mix up one of these 10 parameters it will still compile but hopefully crash hard at runtime..

include/storage/io.hpp
chaupow: no further action needed

chaupow:
@daniel-j-h adjusted to comments. would be great if you could take a look at it again.

include/engine/datafacade/internal_datafacade.hpp
```diff
@@ -182,27 +181,28 @@ class InternalDataFacade final : public BaseDataFacade
...
+        if (!nodes_input_stream)
+        {
+            throw util::exception("Could not open " + nodes_file_path.string() + " for reading.");
+        }
 
-        extractor::QueryNode current_node;
-        unsigned number_of_coordinates = 0;
-        nodes_input_stream.read((char *)&number_of_coordinates, sizeof(unsigned));
+        const auto number_of_coordinates = storage::io::readElementCount(nodes_input_stream);
+        boost::filesystem::ifstream nodes_input_streamtest(nodes_file_path, std::ios::binary);
```
daniel-j-h: What is this?

include/storage/io.hpp
```diff
@@ -15,17 +20,38 @@ namespace storage
...
+    std::uint64_t number_of_elements = 0;
+    input_stream.read((char *)&number_of_elements, sizeof(std::uint64_t));
+    return number_of_elements;
+}
+
+// To make function calls consistent, this function returns the fixed number of properties
+inline std::size_t readPropertiesCount() { return 1; }
+
+// Returns the number of bytes in a file
+inline std::size_t readNumberOfBytes(boost::filesystem::ifstream &timestamp_input_stream)
```
daniel-j-h: This only works for timestamp streams? Then probably should have "timestamp" in the function name.

PR #3149, opened 2016-10-21
Transparently Tidy Traces in Map Matching, resolves #2840.

### LINKED ISSUE #2840: Map Matching: Clean Up Traces / Blob Removal / Simplification
At the moment Map Matching requires the user to tidy up traces. For example blob removal (noisy points close to each other while waiting at a traffic signal) has to be done, and some more simplification to get "clean" traces.

If the user fails to do this, results are unexpected at best and completely off at worst.

This pre-processing can be done e.g. with
https://github.com/mapbox/geojson-tidy
but it's really an implementation detail and we should do this transparently for the user.

Users running into this issue:
- https://github.com/Project-OSRM/osrm-backend/issues/2768

daniel-j-h:
For https://github.com/Project-OSRM/osrm-backend/issues/2840. The Map Matching plugin currently has issues with:
- high frequency traces and (performance)
- blobs, think noise at traffic signals (correctness)

This changeset implements trace-tidying transparently for the user.

We hopefully will see both performance gains as well as better matches!

Tasks
- [x] implement tidy logic
- [x] unit tests for tidy logic
- [ ] hook up to map matching
- [ ] review
- [ ] adjust for comments
- [ ] adapt http.md v5 spec telling user about transparent tidying

### DIFF
--- a/src/engine/plugins/match.cpp
+++ b/src/engine/plugins/match.cpp
@@ -3,6 +3,7 @@
 
 #include "engine/api/match_api.hpp"
 #include "engine/api/match_parameters.hpp"
+#include "engine/api/match_parameters_tidy.hpp"
 #include "engine/map_matching/bayes_classifier.hpp"
 #include "util/coordinate_calculation.hpp"
 #include "util/integer_range.hpp"
@@ -132,7 +133,8 @@ Status MatchPlugin::HandleRequest(const std::shared_ptr<datafacade::BaseDataFaca
 
     if (!time_increases_monotonically)
     {
-        return Error("InvalidValue", "Timestamps need to be monotonically increasing.", json_result);
+        return Error(
+            "InvalidValue", "Timestamps need to be monotonically increasing.", json_result);
     }
 
     // assuming radius is the standard deviation of a normal distribution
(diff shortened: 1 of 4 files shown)


src/engine/plugins/match.cpp
```diff
@@ -123,6 +124,8 @@ Status MatchPlugin::HandleRequest(const std::shared_ptr<datafacade::BaseDataFaca
         return Error("InvalidValue", "Invalid coordinate value.", json_result);
     }
 
+    // TODO: check is_sorted(ts, greater_equal), throw if not
```
daniel-j-h: Done separately in https://github.com/Project-OSRM/osrm-backend/pull/3152 - should be merged into master asap.

daniel-j-h:
I just re-wrote this to directly work on `MatchParameters` since the Match plugin is quite heavily coupled to the parameters. Idea is to make a copy, tidy it, pass it to the matching implementation. The adjust based on the mapping and the original parameters.

daniel-j-h:
Unless this PR gets prioritized and finished users have to tidy / blob-remove their traces before sending them to the Match service. Otherwise map matching will produce really bad results.

This can be done on client / api side via https://github.com/mapbox/geojson-tidy, too.

 cc @karenzshea @willwhite @danpat

TheMarex:
Waiting for this would however mean waiting for the `5.6.0` release. I feel this is an important improvement, but also that we can ship this incrementally. Maybe it will be worth even a second blog post?

danpat:
I agree, we shouldn't hold the 5.6.0 release for this, but I'd be happy to see it land sooner rather than later.

daniel-j-h:
This was merged with https://github.com/Project-OSRM/osrm-backend/pull/3815 and landed in master. Closing here.

PR #3151, opened 2016-10-22
Prototype export to normalized graph

TheMarex:
# Issue

Explores exporting our compressed graph for other consumers.


TheMarex:
Stale.

PR #3152, opened 2016-10-22
Prevents the Matching getting confused when users are time-travelling.

daniel-j-h:
Error code when timestamps in traces are not monotonically increasing.

What about time-syncs on devices (think hard ntpd re-syncs)? I'm fine with adding the no-jumps-into-the- past pre-condition, heck even my [Dovecot shut down hard](http://wiki.dovecot.org/TimeMovedBackwards) when this occurred (happened to me before).
 
## Tasklist
 - [x] Add a note to Wiki
 - [x] review
 - [x] adjust for comments

### DIFF
--- a/src/engine/plugins/match.cpp
+++ b/src/engine/plugins/match.cpp
@@ -12,6 +12,8 @@
 #include <cstdlib>
 
 #include <algorithm>
+#include <functional>
+#include <iterator>
 #include <memory>
 #include <string>
 #include <vector>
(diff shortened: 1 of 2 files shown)


src/engine/plugins/match.cpp
```diff
@@ -123,6 +125,16 @@ Status MatchPlugin::HandleRequest(const std::shared_ptr<datafacade::BaseDataFaca
...
     }
 
+    // Check for same or increasing timestamps. Impl. note: Incontrast to `sort(first,
+    // last, less_equal)` checking `greater` in reverse meets irreflexive requirements.
+    const auto time_increases_monotonically = std::is_sorted(
+        parameters.timestamps.rbegin(), parameters.timestamps.rend(), std::greater<>{});
+
+    if (!time_increases_monotonically)
+    {
+        return Error("InvalidValue", "Invalid timestamp value - jumps into the past.", json_result);
```
TheMarex: Maybe an error message like "Timestamp values need to increase in time" would work better?

daniel-j-h: Done.

PR #3157, opened 2016-10-24
Documents InvalidValue response code, resolves #3150

### LINKED ISSUE #3150: InvalidValue is not in the v5 spec
Plugins `viaroute`, `match`, `trip` may return `InvalidValue` response code.

This return code is not documented in the v5 spec at

https://github.com/Project-OSRM/osrm-backend/blob/master/docs/http.md

daniel-j-h:
For #3150 - `InvalidValue` can be returned when parameters are not valid. Ouch.


TheMarex:
@daniel-j-h you can do doc changes with `[skip ci]` in the commit message to tell travis not to bother. :wink:

daniel-j-h:
Ah good point, thanks for reminding!

Issue #3156, opened 2016-10-24
Extract fails to load Sweden


rinigus:
Running extract on Sweden latest (http://download.geofabrik.de/europe/sweden-latest.osm.pbf) leads to a core dump:

/usr/local/osrm/bin/osrm-extract -p ../code/osrm-backend/profiles/car.lua sweden-latest.osm.pbf                                                            1 :(
[info] Using script ../code/osrm-backend/profiles/car.lua
[info] Input file: sweden-latest.osm.pbf
[info] Profile: car.lua
[info] Threads: 4
[STXXL-MSG] STXXL v1.4.1 (prerelease/Debug)
[STXXL-ERRMSG] Warning: no config file found.
[STXXL-ERRMSG] Using default disk configuration.
[STXXL-MSG] Disk '/var/tmp/stxxl' is allocated, space: 1000 MiB, I/O implementation: syscall autogrow delete_on_exit queue=0 devid=0
[info] Parsing in progress..
[info] input file generated by Osmium (http://wiki.openstreetmap.org/wiki/Osmium)
[info] timestamp: 2016-10-23T19:29:02Z
[info] Found 3 turn restriction tags:
[info]   motorcar
[info]   motor_vehicle
[info]   vehicle
terminate called after throwing an instance of 'osrm::util::exception'
  what():  ERROR occurred in profile script:
No matching overload found, candidates:
void <unknown>(RoadClassification&,unsigned char)
[1]    32740 abort (core dumped)  /usr/local/osrm/bin/osrm-extract -p ../code/osrm-backend/profiles/car.lua 
/usr/local/osrm/bin/osrm-extract -p ../code/osrm-backend/profiles/car.lua   117.61s user 2.77s system 318% cpu 37.795 total

I'm running the current version from Github with the latest commit being d05b85c2272df7f84cbcf10971868545ed05838f. Platform: Ubuntu 16.04.1 LTS

daniel-j-h:
This was fixed in https://github.com/Project-OSRM/osrm-backend/pull/3160 - please try latest master.

rinigus:
Thank you - that helped!

PR #3160, opened 2016-10-24
Don't fail on incorrect lane tags '1; 2'

oxidase:
# Issue

Some lane tags can be incorrect like http://www.openstreetmap.org/way/30350672 and lc will be nil, that does not match num_lanes setter signature.

## Tasklist
 - [x] review
 - [ ] adjust for comments

## Requirements / Relations
#3156

### DIFF
--- a/profiles/lib/guidance.lua
+++ b/profiles/lib/guidance.lua
@@ -63,7 +63,7 @@ function Guidance.set_classification (highway, result, input_way)
     local lane_count = input_way:get_value_by_key("lanes")
     if lane_count and lane_count ~= "" then
         local lc = tonumber(lane_count)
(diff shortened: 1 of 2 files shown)


profiles/lib/guidance.lua
```diff
@@ -63,7 +63,7 @@ function Guidance.set_classification (highway, result, input_way)
     local lane_count = input_way:get_value_by_key("lanes")
     if lane_count and lane_count ~= "" then
         local lc = tonumber(lane_count)
-        if lane_count ~= nil then
+        if lc ~= nil then
```
daniel-j-h: Ouch. Looks good, thanks for catching this!

PR #3130, opened 2016-10-20
cmake fixes

springmeyer:
# Issue

Before this PR the `Libs.private` from libosrm.pc looked like (For the upcoming masonize PR):

```
[...snip...]
-L -lpthread -L-L/home/travis/build/Project-OSRM/osrm-backend/mason_packages/linux-x86_64/tbb/43_20150316 -l tbb -L -lrt -L -lz
```

Notice the `-L-L` and the orphaned `-L`.

And looked like this for `master`:

```
[...snip...]
Libs.private:  -L/usr/lib -ltbb -L/usr/lib -ltbbmalloc -L -lrt -L -lgcov -L/usr/lib/x86_64-linux-gnu -lz
```

Has orphaned `-L`.

Seen via `pkg-config libosrm --debug`

This PR also refactors to call find_package for boost in one place (helps prepare for upcoming mason PR)

PR #3162, opened 2016-10-25
Revert "cmake fixes"

springmeyer:
Project-OSRM/osrm-backend#3130 breaks node-osrm and needs more testing. Preparing to revert for now.


springmeyer:
Closing this - no need to revert at the fix turned out to be simple and will land with #3163

Issue #3072, opened 2016-10-17
Consolidate SharedDataFacade and InternalDataFacade


danpat:
The SharedDataFacade allocates one large memory block and loads data into it.

The InternalDataFacade allocates memory for individual data structures.

Beyond that, their logic is basically the same, they both provide access to large amounts of data in RAM.

We can modify the InternalDataFacade to use the same approach as the SharedDataFacade - allocate a large block of RAM and load data into it.  This would allow us to consolidate a lot of duplicated logic.

Related file-loading de-duplication work has already started here:  https://github.com/Project-OSRM/osrm-backend/issues/2989

danpat:
Plan for implementing this:

  [ ] - update the SharedDatafacade to use http://theboostcpplibraries.com/boost.multiarray instead of util::ShM<>::vector.
  [ ] - update the InternalDatafacade to do the same
  [ ] - refactor facades to share common logic
  [ ] - move internal-vs-shm logic into separate module
  [ ] - Replace SharedDatafacade/InternalDatafacade with MemoryBlockDatafacade

danpat:
This was done with #3165

Issue #2242, opened 2016-04-09
New OSRM data format


TheMarex:
Currently our output on a global planet extract looks like this:

```
  25G latest.osrm
    4 latest.osrm.core
    8 latest.osrm.datasource_indexes
   12 latest.osrm.datasource_names
 7.7G latest.osrm.ebg
 7.7G latest.osrm.edges
 980M latest.osrm.enw
  21G latest.osrm.fileIndex
  11G latest.osrm.geometry
  18G latest.osrm.hsgr
 980M latest.osrm.level
 119M latest.osrm.names
 8.8G latest.osrm.nodes
   12 latest.osrm.properties
 1.3G latest.osrm.ramIndex
  15M latest.osrm.restrictions
   20 latest.osrm.timestamp
```

Yes this are 16 files. Its about time that we reconcile this. This issues should capture our requirements around a new data format. Of the top of my head:
- Only one final file (I expect we will need at least on temporary file for `osrm-contract`)
- Platform independent on-disk storage
- Needs to be extremely fast to read and write
- Versioned
- Documented
- Resides in its own subsystem (no random `std::fstream` calls allover the place)
### Platform dependence

We are kind of fortunate in the sense that we don't need to support complex nested data types. Almost all data we use is just a big array of something. Usually 32bit integers, no floating points and no pointers (thankfully).
So we might get away with a lot less complex solutions (no alignment problems).
So what we want to make sure is that we get the following right:
- datatype size (only use types that have an explicit size like `std::uint64_t`)
- alignment through sticking to primitive types

Existing solutions here:
- Protobuf (high adoption, already a dependency, slow, schema)
- Cap'n'Proto (schema, claims to be zero overhead of x64)
- Boost::serialization (slow as hell, no schema)
- cereal http://uscilab.github.io/cereal/ (something like https://github.com/USCiLab/cereal/blob/master/include/cereal/archives/portable_binary.hpp should be what we need)

Reading through Protobuf and Cap'n'Proto, they don't strike me as immediate fits. Both rely on schema based generators. I would expect we could capture the very tightly scoped functionality from above in a single header-only library.
### References
- https://isocpp.org/wiki/faq/serialization#serialize-binary-format good read in general
- https://capnproto.org/cxx.html

/cc @daniel-j-h @danpat @miccolis

TheMarex:
[...]

PR #3165, opened 2016-10-25
Consolidate InternalDataFacade and SharedDataFacade

danpat:
This PR targets https://github.com/Project-OSRM/osrm-backend/issues/3072

The main changes are:

  - both datafacades now share the same "one big memory block" logic, the only difference is how the memory blocks are obtained/retained
  - all file-to-memory logic now resides in `storage.cpp`, there is _Only One Way To Do It_<sup>TM</sup>

What this PR does not do:
  - no changes to file loading error handling
  - no changes to file formats

These should be addressed separately.

The main goal here is to remove the similar-but-slightly-different code between the InternalDatafacade and the SharedDataFacade.

The InternalDataFacade now allocates a large memory block and accesses it using the same logic that the SharedDatafacade used to do with the IPC SHM block.

## Tasklist
- [x] review
- [x] adjust for comments


danpat:
@TheMarex @daniel-j-h @MoKob @jakepruitt care to take a look over this?

The logic seems sound, but, as usual, there is a lot of historical cruft left.  My preference is to get this merged, then work on other problems (file loading logic ugliness, error handling) as separate pieces of work.

include/engine/datafacade/memory_datafacade_base.hpp
```diff
@@ -47,10 +44,10 @@ namespace engine
 namespace datafacade
 {
 
-class SharedDataFacade final : public BaseDataFacade
+class MemoryDataFacadeBase : public BaseDataFacade
```
MoKob: I find this name highly impractical. `memory` at least to me is anything you can store and read data from. So `Internal Memory`, usually referring to private RAM, `Shared Memory` to shared Ram or `External Memory` which collects all mounted devices.

Since All Datafacades (shared/internal) implement this interface, why do we need to put it between BaseDataFacade and the two users? Why don't we simply add the functionality in BaseDataFacade and avoid this additional layer.

Is there ever a use-case where we would have a BaseDataFacade without a MemoryDataFacade functionality?

include/engine/datafacade/memory_datafacade_base.hpp
```diff
@@ -158,325 +150,265 @@ class SharedDataFacade final : public BaseDataFacade
...
     }
 
-    void LoadNames()
+    void InitNames()
     {
-        auto offsets_ptr = data_layout->GetBlockPtr<unsigned>(
-            shared_memory, storage::SharedDataLayout::NAME_OFFSETS);
-        auto blocks_ptr = data_layout->GetBlockPtr<IndexBlock>(
-            shared_memory, storage::SharedDataLayout::NAME_BLOCKS);
+        auto offsets_ptr =
```
MoKob: these should be possible to be `auto *const` (all over this PR)

src/storage/storage.cpp
```diff
@@ -184,12 +184,78 @@ Storage::ReturnCode Storage::Run(int max_wait)
     }
 
     // Allocate a memory layout in shared memory
-    auto layout_memory = makeSharedMemory(layout_region, sizeof(SharedDataLayout), true);
-    auto shared_layout_ptr = new (layout_memory->Ptr()) SharedDataLayout();
+    auto layout_memory = makeSharedMemory(layout_region, sizeof(DataLayout), true);
```
MoKob: The following set-up routines are local to the SharedMemoryFacade, are they? Could we potentially encapsulate this in the Facade/Layout or provide a factory? It seems dangerous to me to require this kind of knowledge outside of the respective facade.

src/storage/storage.cpp
```diff
@@ -433,394 +492,515 @@ Storage::ReturnCode Storage::Run(int max_wait)
...
     }
-    NodeID *geometries_node_id_list_ptr = shared_layout_ptr->GetBlockPtr<NodeID, true>(
-        shared_memory_ptr, SharedDataLayout::GEOMETRIES_NODE_LIST);
-
-    geometry_input_stream.read((char *)&temporary_value, sizeof(unsigned));
-    BOOST_ASSERT(temporary_value ==
-                 shared_layout_ptr->num_entries[SharedDataLayout::GEOMETRIES_NODE_LIST]);
 
-    if (shared_layout_ptr->GetBlockSize(SharedDataLayout::GEOMETRIES_NODE_LIST) > 0)
+    // ram index file name
```
MoKob: kind of unhelpful without further context

src/storage/storage.cpp
```diff
@@ -433,394 +492,515 @@ Storage::ReturnCode Storage::Run(int max_wait)
...
+        /* To be replaced by io loading code */
+        boost::filesystem::ifstream name_stream(config.names_data_path, std::ios::binary);
+        unsigned name_blocks = 0;
+        name_stream.read((char *)&name_blocks, sizeof(unsigned));
+        unsigned number_of_chars = 0;
+        name_stream.read((char *)&number_of_chars, sizeof(unsigned));
+        /* END to be replaced */
+
+        // Loading street names
+        unsigned *name_offsets_ptr =
```
MoKob: not only the auto ptrs but also this ptr should be possible to convert to `const unsigned *`

src/storage/storage.cpp
```diff
@@ -433,394 +492,515 @@ Storage::ReturnCode Storage::Run(int max_wait)
...
-    uint8_t *datasources_list_ptr = shared_layout_ptr->GetBlockPtr<uint8_t, true>(
-        shared_memory_ptr, SharedDataLayout::DATASOURCES_LIST);
-    if (shared_layout_ptr->GetBlockSize(SharedDataLayout::DATASOURCES_LIST) > 0)
+    // Turn lane data
     {
-        io::readDatasourceIndexes(geometry_datasource_input_stream,
-                                  datasources_list_ptr,
-                                  number_of_compressed_datasources);
-    }
+        /* TODO: remove this block, replace with io.hpp ref */
```
MoKob: possibly link to the ticket for these?

include/engine/datafacade/memory_datafacade_base.hpp
TheMarex: The maps.me code for example adds an out-of-memory implementation that would break this paradigm. Seems overkill for now, but having a fully abstract base-class makes sense (also for mocking, see `MockDataFacade`).

include/engine/datafacade/internal_memory_datafacade.hpp
```diff
@@ -0,0 +1,56 @@
...
+        // Allocate the memory block, then load data from files into it
+        internal_memory.reset(new char[internal_layout->GetSizeOfLayout()]);
+        storage.LoadData(internal_layout.get(), internal_memory.get());
+
+        // Set the common base-class pointers
+        data_layout = internal_layout.get();
+        memory_block = internal_memory.get();
+
+        // Adjust all the private m_* members to point to the right places
+        Init();
```
TheMarex: Abbreviated names for functions is bad style. Please consider using `Initialize` or to be more exact `InitializePrivateData` or something.

include/engine/datafacade/memory_datafacade_base.hpp
```diff
@@ -121,35 +113,35 @@ class SharedDataFacade final : public BaseDataFacade
     std::shared_ptr<util::RangeTable<16, true>> m_bearing_ranges_table;
     util::ShM<DiscreteBearing, true>::vector m_bearing_values_table;
 
-    void LoadChecksum()
+    void InitChecksum()
```
TheMarex: Same here with all the `Init*` names.

include/storage/shared_datatype.hpp
```diff
@@ -171,6 +172,20 @@ struct SharedDataLayout
 
         return ptr;
     }
+    friend std::ostream &operator<<(std::ostream &os, const DataLayout &layout)
```
TheMarex: Could we try to keep the debugging code separate from the implementation? @MoKob did something very useful like this by adding a `debug.hpp` header. This makes it a lot easier to spot debug code in a code review.

src/storage/storage.cpp
```diff
@@ -433,394 +492,515 @@ Storage::ReturnCode Storage::Run(int max_wait)
...
+        if (!datasource_names_input_stream)
+        {
+            throw util::exception("Could not open " + config.datasource_names_path.string() +
+                                  " for reading.");
+        }
+
+        const io::DatasourceNamesData datasource_names_data =
+            io::readDatasourceNames(datasource_names_input_stream);
+
+        /* END TODO */
```
TheMarex: left-over from debugging?

src/storage/storage.cpp
```diff
@@ -433,394 +492,515 @@ Storage::ReturnCode Storage::Run(int max_wait)
...
+                      layout_ptr->num_entries[DataLayout::COORDINATE_LIST]);
     }
 
-    if (!bearing_class_table.empty())
+    // store timestamp
     {
-        auto bearing_class_ptr = shared_layout_ptr->GetBlockPtr<DiscreteBearing, true>(
-            shared_memory_ptr, SharedDataLayout::BEARING_VALUES);
-        std::copy(bearing_class_table.begin(), bearing_class_table.end(), bearing_class_ptr);
+        /* TODO: replace this */
```
TheMarex: What is the problem here?

include/engine/datafacade/internal_memory_datafacade.hpp
```diff
@@ -0,0 +1,56 @@
...
+ * shared memory, so this class, and the SharedDatafacade both
+ * share a common base.
+ * This class holds a unique_ptr to the memory block, so it
+ * is auto-freed upon destruction.
+ */
+class InternalDataFacade final : public BigRAMBlockDataFacadeBase
+{
+
+  private:
+    std::unique_ptr<char[]> internal_memory;
```
daniel-j-h: Same here char, uint8_t

include/engine/datafacade/internal_memory_datafacade.hpp
```diff
@@ -0,0 +1,56 @@
...
+    std::unique_ptr<storage::DataLayout> internal_layout;
```
daniel-j-h: Why wrapped in a unique ptr?

include/engine/datafacade/internal_memory_datafacade.hpp
```diff
@@ -0,0 +1,56 @@
...
+    explicit InternalDataFacade(const storage::StorageConfig &config)
+    {
+        storage::Storage storage(config);
+
+        // Calculate the layout/size of the memory block
+        internal_layout = std::make_unique<storage::DataLayout>();
+        storage.LoadLayout(internal_layout.get());
+
+        // Allocate the memory block, then load data from files into it
+        internal_memory.reset(new char[internal_layout->GetSizeOfLayout()]);
```
daniel-j-h: Hm can't you just

    mem = make_unique<T>(42);

instead of this (I think I saw the pattern above, too)

include/engine/datafacade/shared_memory_datafacade.hpp
```diff
@@ -0,0 +1,95 @@
...
+    storage::SharedDataType layout_region;
+    storage::SharedDataType data_region;
+    unsigned shared_timestamp;
+
+    SharedDataFacade() {}
+
+  public:
+    // this function handle the deallocation of the shared memory it we can prove it will not be
+    // used anymore
+    virtual ~SharedDataFacade()
```
daniel-j-h: I think some functions below can throw which is a no-go here in the dtor.

Either try-catch them and do "something" about it, or add `noexcept(true)` to the dtor declaration which at least shuts down hard as soon as an exception is getting thrown.

include/engine/datafacade/shared_memory_datafacade.hpp
```diff
@@ -0,0 +1,95 @@
...
+                                             << shared_timestamp;
+
+        BOOST_ASSERT(storage::SharedMemory::RegionExists(layout_region));
+        m_layout_memory = storage::makeSharedMemory(layout_region);
+
+        data_layout = static_cast<storage::DataLayout *>(m_layout_memory->Ptr());
+
+        BOOST_ASSERT(storage::SharedMemory::RegionExists(data_region));
+        m_large_memory = storage::makeSharedMemory(data_region);
+        memory_block = (char *)(m_large_memory->Ptr());
```
daniel-j-h: reinterpret_cast

include/storage/io.hpp
```diff
@@ -20,11 +24,134 @@ namespace storage
...
+            throw util::exception("Error opening " + filename + ":" + std::strerror(errno));
+
+        if (check_fingerprint && !readAndCheckFingerprint())
+        {
+            throw util::exception("Fingerprint mismatch in " + filename);
+        }
+    }
+
+    /* Read count objects of type T into pointer dest */
+    template <typename T> void readInto(T *dest, const std::size_t count)
```
daniel-j-h: Why a T*? Can dest ever be null? Yes -> check for it. No -> Make it a ref.

include/storage/io.hpp
```diff
@@ -20,11 +24,134 @@ namespace storage
...
+
+    template <typename T> void skip(const std::size_t element_count)
+    {
+        boost::iostreams::seek(input_stream, element_count * sizeof(T), BOOST_IOS::cur);
+    }
+
+    /*******************************************/
+
+    std::uint32_t readElementCount32() { return readOne<std::uint32_t>(); }
+    std::uint64_t readElementCount64() { return readOne<std::uint64_t>(); }
```
daniel-j-h: Why the destinction?

include/storage/io.hpp
```diff
@@ -20,11 +24,134 @@ namespace storage
...
+    std::vector<std::string> readLines()
+    {
+        std::vector<std::string> result;
+        std::string thisline;
+        try
+        {
+            while (std::getline(input_stream, thisline))
+            {
+                std::clog << "Read " << thisline << std::endl;
+                result.push_back(thisline);
```
daniel-j-h: move

include/storage/io.hpp
```diff
@@ -20,11 +24,134 @@ namespace storage
...
+    }
+
+    /* Read count objects of type T into pointer dest */
+    template <typename T> void readInto(T *dest, const std::size_t count)
+    {
+        static_assert(std::is_trivially_copyable<T>::value,
+                      "bytewise reading requires trivially copyable type");
+        if (count == 0)
+            return;
+        input_stream.read(reinterpret_cast<char *>(dest), count * sizeof(T));
```
daniel-j-h: Why not capture the return stream& and check for (!stream)

include/storage/io.hpp
```diff
@@ -20,11 +24,134 @@ namespace storage
...
+        }
+        catch (const std::ios_base::failure &e)
+        {
+            // EOF is OK here, everything else, re-throw
+            if (!input_stream.eof())
+                throw;
+        }
+        return result;
+    }
+};
```
daniel-j-h: Why does File have to do more than it has to? For example why does it do Fingerprint checking and reading?

include/storage/io.hpp
```diff
@@ -20,11 +24,134 @@ namespace storage
...
-inline std::uint64_t readElementCount(boost::filesystem::ifstream &input_stream)
+inline std::uint64_t readElementCount64(boost::filesystem::ifstream &input_stream)
 {
     std::uint64_t number_of_elements = 0;
-    input_stream.read((char *)&number_of_elements, sizeof(std::uint64_t));
+    input_stream.read(reinterpret_cast<char *>(&number_of_elements), sizeof(number_of_elements));
+    return number_of_elements;
+}
+
+inline std::uint64_t readElementCount32(boost::filesystem::ifstream &input_stream)
```
daniel-j-h: Count32 returns a uin64_t?

include/storage/io.hpp
```diff
@@ -20,11 +24,134 @@ namespace storage
...
+            if (!input_stream.eof())
+                throw;
+        }
+        return result;
+    }
+};
+
 // Reads the count of elements that is written in the file header and returns the number
-inline std::uint64_t readElementCount(boost::filesystem::ifstream &input_stream)
+inline std::uint64_t readElementCount64(boost::filesystem::ifstream &input_stream)
```
daniel-j-h: Shouldn't these take a File now?

include/storage/io.hpp
```diff
@@ -20,11 +24,134 @@ namespace storage
 namespace io
 {
 
+class File
```
daniel-j-h: Why not making a type-level distinction between an InFile and an OutFile?

include/storage/io.hpp
```diff
@@ -52,23 +179,20 @@ static_assert(sizeof(HSGRHeader) == 20, "HSGRHeader is not packed");
...
 
     HSGRHeader header;
-    input_stream.read(reinterpret_cast<char *>(&header.checksum), sizeof(header.checksum));
-    input_stream.read(reinterpret_cast<char *>(&header.number_of_nodes),
-                      sizeof(header.number_of_nodes));
-    input_stream.read(reinterpret_cast<char *>(&header.number_of_edges),
-                      sizeof(header.number_of_edges));
+    input_file.readInto(header.checksum);
+    input_file.readInto(header.number_of_nodes);
+    input_file.readInto(header.number_of_edges);
```
daniel-j-h: We had a policy-based writer once abstracting over all of this (never got merged..)
https://github.com/daniel-j-h/osrm-backend/blob/39383855dd6958a0de3b9981b2780b4a37ce4b56/io/osrm_writer.hpp

Why do we have to know about the file's layout here?

include/storage/storage.hpp
```diff
@@ -52,6 +53,9 @@ class Storage
 
     ReturnCode Run(int max_wait);
 
+    void LoadLayout(DataLayout *layout);
+    void LoadData(DataLayout *layout_ptr, char *memory_ptr);
```
daniel-j-h: Can the raw pointers every be null? Can these be const?

include/util/guidance/turn_lanes.hpp
```diff
@@ -27,6 +28,11 @@ template <> struct hash<::osrm::util::guidance::LaneTuple>
 {
     inline std::size_t operator()(const ::osrm::util::guidance::LaneTuple &bearing_class) const;
 };
+template <> struct hash<::osrm::util::guidance::LaneTupleIdPair>
+{
+    inline std::size_t
+    operator()(const ::osrm::util::guidance::LaneTupleIdPair &bearing_class) const;
+};
```
daniel-j-h: Only a declaration? This looks like dead code? You seem to implement the hash_value function below?

include/util/io.hpp
```diff
@@ -66,17 +56,14 @@ bool serializeVector(std::ostream &stream, const std::vector<simple_type> &data)
 template <typename simple_type>
 bool deserializeVector(const std::string &filename, std::vector<simple_type> &data)
 {
-    std::ifstream stream(filename, std::ios::binary);
 
-    if (!readAndCheckFingerprint(stream))
-        return false;
+    storage::io::File file(filename, true);
```
daniel-j-h: Meh, can you please remove the defaulted boolean and add an enum class at least

`enum class CheckFingerprint { Yes, No };`

no idea what `true` means for a File in this context without taking a look at the File header.

include/util/io.hpp
```diff
@@ -66,17 +56,14 @@ bool serializeVector(std::ostream &stream, const std::vector<simple_type> &data)
...
 
-    std::uint64_t count = 0;
-    stream.read(reinterpret_cast<char *>(&count), sizeof(count));
+    const auto count = file.readElementCount64();
     data.resize(count);
     if (count)
-        stream.read(reinterpret_cast<char *>(&data[0]), sizeof(simple_type) * count);
-    return static_cast<bool>(stream);
+        file.readInto(data.data(), count);
+    return true;
```
daniel-j-h: If the function now throws on error, it should return `void` and not `bool`.

`bool fn();` makes it seem like the call site has to check for the return value, when it really does not.
Instead exceptions will be thrown and _that_ has to be accounted for.

include/util/io.hpp
```diff
@@ -199,28 +186,24 @@ inline bool serializeFlags(const boost::filesystem::path &path, const std::vecto
...
-        flag_stream.read(reinterpret_cast<char *>(&chunk), sizeof(chunk));
+        flag_file.readInto(chunk);
         std::bitset<32> chunk_bits(chunk);
         for (std::size_t bit = 0; bit < 32 && bit_position < number_of_bits; ++bit, ++bit_position)
             flags[bit_position] = chunk_bits[bit];
     }
     SimpleLogger().Write() << "Read " << number_of_bits << " bits in " << chunks
                            << " Chunks from disk.";
-    return static_cast<bool>(flag_stream);
+    return true;
```
daniel-j-h: Same here and in related functions.

src/storage/storage.cpp
```diff
@@ -184,643 +182,670 @@ Storage::ReturnCode Storage::Run(int max_wait)
...
+    {
+        const auto file_index_path_ptr =
+            layout_ptr->GetBlockPtr<char, true>(memory_ptr, DataLayout::FILE_INDEX_PATH);
+        // make sure we have 0 ending
+        std::fill(file_index_path_ptr,
+                  file_index_path_ptr + layout_ptr->GetBlockSize(DataLayout::FILE_INDEX_PATH),
+                  0);
+        auto absolute_file_index_path = boost::filesystem::absolute(config.file_index_path);
+        std::copy(absolute_file_index_path.string().begin(),
+                  absolute_file_index_path.string().end(),
```
daniel-j-h: This is Undefined Behavior  - string() returns a temporary so .begin() and .end() point to different variables.

src/storage/storage.cpp
```diff
@@ -184,643 +182,670 @@ Storage::ReturnCode Storage::Run(int max_wait)
...
+        io::readEdges(edges_input_file,
+                      via_geometry_ptr,
+                      name_id_ptr,
+                      turn_instructions_ptr,
+                      lane_data_id_ptr,
+                      travel_mode_ptr,
+                      entry_class_id_ptr,
+                      pre_turn_bearing_ptr,
+                      post_turn_bearing_ptr,
+                      number_of_original_edges);
```
daniel-j-h: This is fugly.

src/storage/storage.cpp
```diff
@@ -184,643 +182,670 @@ Storage::ReturnCode Storage::Run(int max_wait)
...
-    for (auto i = 0u; i < number_of_core_markers; ++i)
     {
-        BOOST_ASSERT(unpacked_core_markers[i] == 0 || unpacked_core_markers[i] == 1);
+        io::File geometry_datasource_file(config.datasource_indexes_path);
+        const auto number_of_compressed_datasources = geometry_datasource_file.readElementCount64();
 
-        if (unpacked_core_markers[i] == 1)
+        // load datasource information (if it exists)
+        const auto datasources_list_ptr =
+            layout_ptr->GetBlockPtr<uint8_t, true>(memory_ptr, DataLayout::DATASOURCES_LIST);
```
daniel-j-h: Related: can we change the `bool` here to an enum? What is `true`? (I know shm but still this is semantically meaningless in this context)

src/storage/storage.cpp
```diff
@@ -184,643 +182,670 @@ Storage::ReturnCode Storage::Run(int max_wait)
...
-                 hsgr_header.number_of_edges);
-    hsgr_input_stream.close();
+        // load datasource name information (if it exists)
+        const auto datasource_name_data_ptr =
+            layout_ptr->GetBlockPtr<char, true>(memory_ptr, DataLayout::DATASOURCE_NAME_DATA);
+        if (layout_ptr->GetBlockSize(DataLayout::DATASOURCE_NAME_DATA) > 0)
+        {
+            std::copy(datasource_names_data.names.begin(),
+                      datasource_names_data.names.end(),
+                      datasource_name_data_ptr);
```
daniel-j-h: Please check all copy's to make sure you assert for end - begin == size of the rout range.

Maybe even copy_n.

include/engine/datafacade/bigramblock_datafacade_base.hpp
```diff
@@ -158,325 +158,265 @@ class SharedDataFacade final : public BaseDataFacade
...
-            shared_memory, storage::SharedDataLayout::TRAVEL_MODE);
+            memory_block, storage::DataLayout::TRAVEL_MODE);
         util::ShM<extractor::TravelMode, true>::vector travel_mode_list(
-            travel_mode_list_ptr, data_layout->num_entries[storage::SharedDataLayout::TRAVEL_MODE]);
+            travel_mode_list_ptr, data_layout->num_entries[storage::DataLayout::TRAVEL_MODE]);
         m_travel_mode_list = std::move(travel_mode_list);
 
-        const auto lane_data_id_ptr = data_layout->GetBlockPtr<LaneDataID>(
-            shared_memory, storage::SharedDataLayout::LANE_DATA_ID);
+        const auto lane_data_id_ptr =
```
MoKob: all of these following blocks should really be abstracted over, reducing the duplicate behaviour and reducing possible chances for errors.

include/engine/datafacade/bigramblock_datafacade_base.hpp
```diff
@@ -158,325 +158,265 @@ class SharedDataFacade final : public BaseDataFacade
...
-        LoadGeometries();
-        LoadTimestamp();
-        LoadViaNodeList();
-        LoadNames();
-        LoadTurnLaneDescriptions();
-        LoadCoreInformation();
-        LoadProfileProperties();
-        LoadRTree();
-        LoadIntersectionClasses();
+    void Init(storage::DataLayout *data_layout, char *memory_block)
```
MoKob: All `Initialize`.

Also, since we are essentially doing a huge `shared pointer` style implementation. How about naming it `reset`?

include/engine/datafacade/bigramblock_datafacade_base.hpp
```diff
@@ -537,7 +477,7 @@ class SharedDataFacade final : public BaseDataFacade
 
     OSMNodeID GetOSMNodeIDOfNode(const unsigned id) const override final
     {
-        return m_osmnodeid_list.at(id);
+        return m_osmnodeid_list[id];
```
MoKob: for consistency, could we change the `.at` on all following locations as well?

include/engine/datafacade/internal_memory_datafacade.hpp
```diff
@@ -0,0 +1,52 @@
+#ifndef INTERNAL_DATAFACADE_HPP
```
MoKob: IMHO: to avoid name clashes with any other possible libs, we should keep to `OSRM_ENGINE_DATAFACADE_INTERNAL_DATAFACADE_HPP` (full path header guards).

include/engine/datafacade/bigramblock_datafacade_base.hpp
```diff
@@ -1000,4 +940,4 @@ class SharedDataFacade final : public BaseDataFacade
 }
```
MoKob: /* namespace ... */

include/engine/datafacade/internal_memory_datafacade.hpp
```diff
@@ -0,0 +1,52 @@
...
+ * shared memory, so this class, and the SharedDatafacade both
+ * share a common base.
+ * This class holds a unique_ptr to the memory block, so it
+ * is auto-freed upon destruction.
+ */
+class InternalDataFacade final : public BigRAMBlockDataFacadeBase
+{
+
+  private:
+    std::unique_ptr<char[]> internal_memory;
```
MoKob: so the only difference between the `internal` data facade and the `shared datafacade` is where `internal_memory` / `shared_memory` is stored?
From my understanding this would be a property of the storage, and the `internal` data should in that case be resident to an `internal storage` and not be part of the facade. The facade describes how to access the data, the storage how it is stored.

Previously I already stated that I don't get the separation of both. After all, the datafacade is what we need to access the storage data. So why are we decoupling the data and its access patterns?

include/engine/datafacade/shared_memory_datafacade.hpp
```diff
@@ -0,0 +1,94 @@
...
+            boost::interprocess::defer_lock);
+
+        // if this returns false this is still in use
+        if (exclusive_lock.try_lock())
+        {
+            // Now check if this is still the newest dataset
+            const boost::interprocess::sharable_lock<boost::interprocess::named_upgradable_mutex>
+                lock(shared_barriers->current_regions_mutex);
+
+            auto shared_regions = storage::makeSharedMemory(storage::CURRENT_REGIONS);
```
MoKob: can this be const?

include/storage/io.hpp
MoKob: Also should be `ReadInto`

MoKob: General issue in the file (Naming convention requests large case)

MoKob: Ref is not possible due to reading multiple entries, but can we use a iterator range and check for overflowing the destination?

include/storage/io.hpp
```diff
@@ -20,11 +24,134 @@ namespace storage
...
+        {
+            throw util::exception("Error reading from " + filename + ": " + std::strerror(errno));
+        }
+        else if (bytes_read < expected_bytes)
+        {
+            throw util::exception("Error reading from " + filename + ": Unexpected end of file");
+        }
+    }
+
+    template <typename T> void readInto(T &target) { readInto(&target, 1); }
```
MoKob: Why read into here? Can we rely on `readOne` only?

include/storage/io.hpp
```diff
@@ -20,11 +24,134 @@ namespace storage
...
+
+    std::vector<std::string> readLines()
+    {
+        std::vector<std::string> result;
+        std::string thisline;
+        try
+        {
+            while (std::getline(input_stream, thisline))
+            {
+                std::clog << "Read " << thisline << std::endl;
```
MoKob: left over debug output?

include/storage/shared_datatype.hpp
```diff
@@ -7,6 +7,7 @@
 #include <cstdint>
 
 #include <array>
+#include <iostream>
```
MoKob: left over from debug?

include/util/static_rtree.hpp
```diff
@@ -346,18 +346,9 @@ class StaticRTree
...
-        }
-        if (boost::filesystem::file_size(node_file) == 0)
-        {
-            throw exception("ram index file is empty");
-        }
-        boost::filesystem::ifstream tree_node_file(node_file, std::ios::binary);
+        storage::io::File tree_node_file(node_file);
 
-        const auto tree_size = storage::io::readElementCount(tree_node_file);
+        const auto tree_size = tree_node_file.readElementCount64();
```
MoKob: can we unify the size to be 64 bit in general? Not even offering a 32 bit version?

src/storage/storage.cpp
```diff
@@ -184,643 +182,670 @@ Storage::ReturnCode Storage::Run(int max_wait)
...
-        shared_memory_ptr, SharedDataLayout::FILE_INDEX_PATH);
-    // make sure we have 0 ending
-    std::fill(file_index_path_ptr,
-              file_index_path_ptr +
-                  shared_layout_ptr->GetBlockSize(SharedDataLayout::FILE_INDEX_PATH),
-              0);
-    std::copy(absolute_file_index_path.string().begin(),
-              absolute_file_index_path.string().end(),
-              file_index_path_ptr);
+        BOOST_ASSERT(name_blocks_count * sizeof(unsigned) ==
```
MoKob: can there ever be padding involved? (<=)

src/storage/storage.cpp
```diff
@@ -184,643 +182,670 @@ Storage::ReturnCode Storage::Run(int max_wait)
...
-        shared_memory_ptr, SharedDataLayout::LANE_DATA_ID);
-
-    extractor::guidance::TurnInstruction *turn_instructions_ptr =
-        shared_layout_ptr->GetBlockPtr<extractor::guidance::TurnInstruction, true>(
-            shared_memory_ptr, SharedDataLayout::TURN_INSTRUCTION);
-
-    EntryClassID *entry_class_id_ptr = shared_layout_ptr->GetBlockPtr<EntryClassID, true>(
-        shared_memory_ptr, SharedDataLayout::ENTRY_CLASSID);
+        // Loading street names
+        const auto name_offsets_ptr =
```
MoKob: this repeated behaviour should be abstracted away.

src/storage/storage.cpp
```diff
@@ -184,643 +182,670 @@ Storage::ReturnCode Storage::Run(int max_wait)
...
 
-    // load compressed geometry
-    unsigned temporary_value;
-    unsigned *geometries_index_ptr = shared_layout_ptr->GetBlockPtr<unsigned, true>(
-        shared_memory_ptr, SharedDataLayout::GEOMETRIES_INDEX);
-    geometry_input_stream.seekg(0, geometry_input_stream.beg);
-    geometry_input_stream.read((char *)&temporary_value, sizeof(unsigned));
-    BOOST_ASSERT(temporary_value ==
-                 shared_layout_ptr->num_entries[SharedDataLayout::GEOMETRIES_INDEX]);
+        // For some reason, this value (the same as )
```
MoKob: ?!?

src/storage/storage.cpp
```diff
@@ -184,643 +182,670 @@ Storage::ReturnCode Storage::Run(int max_wait)
...
-    }
-    NodeID *geometries_node_id_list_ptr = shared_layout_ptr->GetBlockPtr<NodeID, true>(
-        shared_memory_ptr, SharedDataLayout::GEOMETRIES_NODE_LIST);
+        const auto name_char_ptr =
+            layout_ptr->GetBlockPtr<char, true>(memory_ptr, DataLayout::NAME_CHAR_LIST);
 
-    geometry_input_stream.read((char *)&temporary_value, sizeof(unsigned));
-    BOOST_ASSERT(temporary_value ==
-                 shared_layout_ptr->num_entries[SharedDataLayout::GEOMETRIES_NODE_LIST]);
+        BOOST_ASSERT_MSG(layout_ptr->AlignBlockSize(temp_count) ==
```
MoKob: is this the only aligned block?

src/storage/storage.cpp
MoKob: can we add some `i/o iterator` to the file and do a ranged based for loop for stuff like this?

src/storage/storage.cpp
```diff
@@ -184,643 +182,670 @@ Storage::ReturnCode Storage::Run(int max_wait)
...
+        nodes_file.skip<std::uint64_t>(1); // node_count
+        const auto coordinates_ptr = layout_ptr->GetBlockPtr<util::Coordinate, true>(
+            memory_ptr, DataLayout::COORDINATE_LIST);
+        const auto osmnodeid_ptr =
+            layout_ptr->GetBlockPtr<std::uint64_t, true>(memory_ptr, DataLayout::OSM_NODE_ID_LIST);
+        util::PackedVector<OSMNodeID, true> osmnodeid_list;
+
+        osmnodeid_list.reset(osmnodeid_ptr, layout_ptr->num_entries[DataLayout::OSM_NODE_ID_LIST]);
+
+        io::readNodes(nodes_file,
```
MoKob: just like above. would be nicer with an iterator abstraction

src/storage/storage.cpp
```diff
@@ -184,643 +182,670 @@ Storage::ReturnCode Storage::Run(int max_wait)
...
+                      coordinates_ptr,
+                      osmnodeid_list,
+                      layout_ptr->num_entries[DataLayout::COORDINATE_LIST]);
     }
 
-    if (shared_layout_ptr->GetBlockSize(SharedDataLayout::BEARING_OFFSETS) > 0)
+    // store timestamp
```
MoKob: store while reading?

src/storage/storage.cpp
```diff
@@ -184,643 +182,670 @@ Storage::ReturnCode Storage::Run(int max_wait)
...
+        const auto timestamp_size = timestamp_file.size();
+
+        const auto timestamp_ptr =
+            layout_ptr->GetBlockPtr<char, true>(memory_ptr, DataLayout::TIMESTAMP);
+        BOOST_ASSERT(timestamp_size == layout_ptr->num_entries[DataLayout::TIMESTAMP]);
+        timestamp_file.readInto(timestamp_ptr, timestamp_size);
     }
 
-    if (shared_layout_ptr->GetBlockSize(SharedDataLayout::BEARING_BLOCKS) > 0)
+    // store search tree portion of rtree
```
MoKob: again, store?

src/storage/storage.cpp
```diff
@@ -184,643 +182,670 @@ Storage::ReturnCode Storage::Run(int max_wait)
...
     {
-        auto bearing_class_ptr = shared_layout_ptr->GetBlockPtr<DiscreteBearing, true>(
-            shared_memory_ptr, SharedDataLayout::BEARING_VALUES);
-        std::copy(bearing_class_table.begin(), bearing_class_table.end(), bearing_class_ptr);
+        io::File core_marker_file(config.core_data_path);
+        const auto number_of_core_markers = core_marker_file.readElementCount32();
+
+        // load core markers
+        std::vector<char> unpacked_core_markers(number_of_core_markers);
+        core_marker_file.readInto(unpacked_core_markers.data(), number_of_core_markers);
```
MoKob: would love if we could move away from these `read into the output` with pointers and use proper iterators( e.g. back inserters).

include/storage/io.hpp
danpat: Can't move here - we're re-using the same string object on each iteration of the loop.

include/storage/io.hpp
danpat: This is the first cut at getting a bunch of the file reading logic into one place.  I expect this class to see further re-factoring, but for now, it's simply code-shuffling.

src/storage/storage.cpp
danpat: This fugliness has simply been moved from elsewhere - we'll tidy this up as part of https://github.com/Project-OSRM/osrm-backend/issues/2242

Issue #3107, opened 2016-10-19
Revisit Merging of Segregated Roads/Joining Road Adjustments


MoKob:
We should refactor the code in the intersection generator [here](https://github.com/Project-OSRM/osrm-backend/blob/3f0f0e306b3c2ee5eada6512c3ce3ab00b1b0460/src/extractor/guidance/intersection_generator.cpp#L184-L480). 

When https://github.com/Project-OSRM/osrm-backend/pull/2779 gets merged, the angle offsets chosen here might not be as relevant anymore.
## Tasks
- [x] create map that shows locations where merging triggers (see https://github.com/Project-OSRM/osrm-backend/pull/3170)
- [ ] compare turn angles at these locations
- [ ] revisit merging criteria (canMerge)
- [ ] remove own angle computations (rely on intersection-classification/coordinate extractor)
- [ ] general refactor

In a related task, we should also take a look at:
[merging at destination](https://github.com/Project-OSRM/osrm-backend/blob/3f0f0e306b3c2ee5eada6512c3ce3ab00b1b0460/src/extractor/guidance/intersection_generator.cpp#L502-L565).

MoKob:
Issues found in looking at the debug map:

Hohenzollernbrücke should not get merged OSM: http://www.openstreetmap.org/#map=18/52.48767/13.30016&layers=C

Gustav Müller Platz is a false merge: http://www.openstreetmap.org/#map=18/52.48337/13.36184&layers=C

Lietzenburger Straße is a false positive
http://www.openstreetmap.org/#map=18/52.49950/13.33916&layers=C

PR #3170, opened 2016-10-25
add a geojson debugger that allows creating features for easy inspection

MoKob:
# Issue

This PR enables GeoJson logging, printing out FeatureCollections to a logfile.
The logger can be initialised at the top defining

`util::ScopedGeojsonLoggerGuard<util::NodeIdVectorToLineString> geojson_guard(
        "debug.geojson", data-for-conversion);`

and used via 
`util::GeojsonLogger<util::NodeIdVectorToLineString>::Write(list_of_node_ids);`
later on.

These collections can later be used for easier inspection of changed features in guidance/other debugging.
## Tasklist
- [x] write up usage docs, with example, in https://github.com/Project-OSRM/osrm-backend/wiki/Debugging or ??
- [x] review
- [x] adjust for comments
## Requirements / Relations

First step on the way to https://github.com/Project-OSRM/osrm-backend/issues/3107

### DIFF
--- a/include/util/geojson_debug_logger.hpp
+++ b/include/util/geojson_debug_logger.hpp
@@ -1,0 +1,185 @@
+#ifndef OSRM_GEOJSON_DEBUG_LOGGER_HPP
+#define OSRM_GEOJSON_DEBUG_LOGGER_HPP
+
+#include <fstream>
+#include <mutex>
+#include <string>
+
+#include "util/json_container.hpp"
+#include "util/json_renderer.hpp"
+#include "util/simple_logger.hpp"
+
+namespace osrm
+{
+namespace util
+{
+
+// in case we want to do scenario-based logging, we can specify a dedicated logging scenario to be
+// able to use multiple files for the same converter
+enum class LoggingScenario
+{
+    eDefault = 0
+};
+
+// forward declaration to become friends
+template <class geojson_conversion_policy, LoggingScenario scenario> class ScopedGeojsonLoggerGuard;
+
+// a geojson logger requires a conversion policy to transfer arbitrary input into viable geojson.
+// features of the same kind are stored in the same geojson file
+template <class geojson_conversion_policy, LoggingScenario scenario = LoggingScenario::eDefault>
+class GeojsonLogger
+{
+    // become friends with the guard
+    friend class ScopedGeojsonLoggerGuard<geojson_conversion_policy, scenario>;
+
+    // having these private enforces the guard to be used to initialise/close/write
+  private:
+    // cannot lock, is tooling for locked function
+    static void output(bool first, const util::json::Object &object)
+    {
+        if (!first)
+            ofs << ",\n\t\t";
+        // objects are simply forwarded
+        util::json::render(ofs, object);
+    }
+
+    // cannot lock, is tooling for locked function
+    static void output(bool first, const util::json::Array &array)
+    {
+        for (const auto object : array.values)
+        {
+            if (!first)
+                ofs << ",\n\t\t";
+
+            util::json::render(ofs, object.get<util::json::Object>());
+
+            first = false;
+        }
+    }
+
+    // writes a single feature into the Geojson file
+    template <typename... Args> static bool Write(Args &&... args)
+    {
+        // make sure to syncronize logging output, our writing should be sequential
+        std::lock_guard<std::mutex> guard(lock);
+
+        // if there is no logfile, we cannot write
+        if (!ofs.is_open() || (nullptr == policy))
+        {
+            // this can only happend between two guards when concurrent writing occurs
+            return false;
+        }
+
+        // use our policy to convert the arguments into geojson, this can be done in parallel
+        const auto json_object = (*policy)(std::forward<Args>(args)...);
+
+        // different features are separated by `,`
+        // since we are not building a full json collection of features, we have to do some of the
+        // bookeeping ourselves. This prevens us from doing one huge json object, if we are
+        // processing larger results
+        output(first, json_object);
+        first = false;
+
+        return static_cast<bool>(ofs);
+    }
+
+    // Opening a logger, we initialize a geojson feature collection
+    // to be later filled with additional data
+    static bool Open(const std::string &logfile)
+    {
+        // if a file is open, close it off. When this function returns, there should be an open
+        // logfile. However, there is a brief period between closing and openen where we could miss
+        // out on log output. Such a sad life
+        if (ofs.is_open())
+        {
+            util::SimpleLogger().Write(logWARNING)
+                << "Overwriting " << logfile
+                << ". Is this desired behaviour? If this message occurs more than once rethink the "
+                   "location of your Logger Guard.";
+            Close();
+        }
+
+        // make sure to syncronize logging output, cannot be locked earlier, since Close also locks
+        // and we don't want deadlocks
+        std::lock_guard<std::mutex> guard(lock);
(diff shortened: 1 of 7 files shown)


include/util/geojson_debug_logger.hpp
```diff
@@ -0,0 +1,154 @@
...
+bool GeojsonLogger<geojson_conversion_policy, scenario>::first;
+
+template <class geojson_conversion_policy, LoggingScenario scenario>
+std::mutex GeojsonLogger<geojson_conversion_policy, scenario>::lock;
+
+template <class geojson_conversion_policy, LoggingScenario scenario>
+std::ofstream GeojsonLogger<geojson_conversion_policy, scenario>::ofs;
+
+template <class geojson_conversion_policy, LoggingScenario scenario>
+geojson_conversion_policy* GeojsonLogger<geojson_conversion_policy, scenario>::policy;
```
daniel-j-h: Hmm don't we have to define them in a single impl. file - otherwise we would validate the one-definition-rule when multiple files include this header?

MoKob:
Documentation to add to the wiki on merge:

# How to use GeoJson-Debugging

This is a short guide to describe usage of our GeoJson debug logging mechanism.

## Outputting into a single file
Geojson debugging requires a few simple steps to output data into a feature collection.

- Create a Scoped Guard that lives through the process and provide it with all required datastructures
- At the location of the output, simply call Write with your own parameters.

A guard (ScopedGeojsonLoggerGuard) requires a logging policy. Per default we provide a way of printing out node-ids as coordinates.

The initialisation to do so looks like this:
`util::ScopedGeojsonLoggerGuard<util::NodeIdVectorToLineString> geojson_guard( "debug.geojson", data-for-conversion);`

The field `data-for-conversion` can be an arbitrary long set of features and needs to match the parameters used for constructing our policy (in this case `util::NodeIdVectorToLineString`).

The policy itself offers a `operator()` accepting a `vector` of `NodeID`.

For outputting data into our file (debug.geojson), we simply need to call the matching logging routine: `util::GeojsonLogger<util::NodeIdVectorToLineString>::Write(list_of_node_ids);`.

Note how both `ScopedGeojsonLoggerGuard` and `GeojsonLogger` are using the same template paramters.


## Limitations
GeoJson debugging requires a single GeoJsonGuard (ScopedGeojsonLoggerGuard) for each desired output file.
For each set of template parameters, only the most recent guard will actually produce output.

`util::ScopedGeojsonLoggerGuard<util::NodeIdVectorToLineString> geojson_guard( "debug.geojson", data-for-conversion);`

`util::ScopedGeojsonLoggerGuard<util::NodeIdVectorToLineString> geojson_guard( "debug-2.geojson", data-for-conversion);`

Will not provide a way to write into two files, but only `debug-2` will actually contain features.

We cannot nest-these calls.

If we want to use the same policy for multiple files, we need to use different template parameters both for the logger and the guard.

`util::ScopedGeojsonLoggerGuard<util::NodeIdVectorToLineString,0> geojson_guard( "debug.geojson", data-for-conversion);`

`util::ScopedGeojsonLoggerGuard<util::NodeIdVectorToLineString,1> geojson_guard( "debug-2.geojson", data-for-conversion);`

as well as,
 
`util::GeojsonLogger<util::NodeIdVectorToLineString,0>::Write(list_of_node_ids);`

`util::GeojsonLogger<util::NodeIdVectorToLineString,1>::Write(list_of_node_ids);`

MoKob:
screen shot 2016-10-25 at 21 07 09

Example in Studio using the IntersectionPrinter policy

daniel-j-h:
Here is some feedback from using it today:

1/ at the moment we have to do

```cpp
ScopedGeojsonLoggerGuard<NodeIdVectorToLineString> sentry( "debug.geojson", node_coordinate_vector);

GeojsonLogger<NodeIdVectorToLineString>::Write(list_of_node_ids);
```

having to repeat the type twice. What about creating a static `Write` function in the scope guard delegating to the logger's `Write` function? Then users don't have to know about the logger at all. 

Usage would be:

```cpp
ScopedGeojsonLoggerGuard<NodeIdVectorToLineString> sentry("debug.geojson", node_coordinate_vector);

sentry.Write(list_of_node_ids);
```

2/ setting attributes (GeoJSON properties) would be a great addition: at the moment we have to serialize out different files. Think: being able to tag nodes with a `Color` type / enum class and then having a converter for converting `Color` to GeoJSON properties.

daniel-j-h:
3/ if you screw up scoping (e.g. if you put the guard into a handler's `operator()`) the out file will silently get overwritten, making it appear as if only 0 / 1 intersection matched. Either throw if there is already a out file or append some sort of number to the file name and write to that instead. I'd really go for the first option, though.

MoKob:
@daniel-j-h throwing is kind of undesired, since it will also throw if you simply run your program twice without cleaning up in between. I'll think about what we could do to prevent this without unwanted side-effects.

include/util/geojson_debug_logger.hpp
```diff
@@ -0,0 +1,185 @@
...
+        if (ofs.is_open())
+        {
+            ofs << "\n\t]\n}";
+            ofs.close();
+        }
+
+        return static_cast<bool>(ofs);
+    }
+
+    static void SetPolicy(geojson_conversion_policy *new_policy) { policy = new_policy; }
```
daniel-j-h: Should this always overwrite an already existing policy? Or should we 
- assert !policy
- assert new_poliy

include/util/geojson_debug_logger.hpp
```diff
@@ -0,0 +1,185 @@
...
+    static geojson_conversion_policy *policy;
+};
+
+// make sure to do opening and closing of our guard
+template <class geojson_conversion_policy, LoggingScenario scenario = LoggingScenario::eDefault>
+class ScopedGeojsonLoggerGuard
+{
+  public:
+    template <typename... Args>
+    ScopedGeojsonLoggerGuard(const std::string logfile, Args &&... args)
```
daniel-j-h: const string & ?

include/util/geojson_debug_policy_toolkit.hpp
```diff
@@ -0,0 +1,112 @@
...
+    num_colors
+};
+
+const constexpr char *geojson_debug_predifined_colors[GeojsonStyleColors::num_colors] = {
+    "#FF4848", "#800080", "#5757FF", "#1FCB4A", "#FFE920", "#29AFD6", "#B05F3C", "#FE67EB"};
+
+const constexpr double geojson_predefined_sizes[GeojsonStyleSize::num_styles] = {
+    2.0, 3.5, 5.0, 6.5, 8};
+
+util::json::Object makeStyle(const GeojsonStyleSize size_type,
```
daniel-j-h: inline

src/util/geojson_debug_policies.cpp
```diff
@@ -0,0 +1,69 @@
...
+                   std::back_inserter(coordinates.values),
+                   NodeIdToCoordinate(node_coordinates));
+
+    return makeFeature("MultiPoint", std::move(coordinates), properties);
+}
+
+//----------------------------------------------------------------
+util::json::Object CoordinateVectorToMultiPoint::
+operator()(const std::vector<util::Coordinate> &input_coordinates,
+           boost::optional<json::Object> properties) const
```
daniel-j-h: if a read-only view is good enough for an optional you should be able to make it a const optional<T> &

applies to all optional usages in this pr

daniel-j-h:
Looks good.

PR #3172, opened 2016-10-25
[WIP] Refactor LUA profiles

emiltin:
# Issue

The purpose of this PR is to:
- Make the LUA code manageable and enable code reuse between profiles.
- Establish a common framework/flow for how to process tags, ie. highway, oneway, surface, maxspeed, etc. while still enabling full customisation. The framework should work across profiles, but allow full customisation.
- Make it easier to debug LUA code by making it easy to run the profile code from the command line, without the C++ binaries. Ie. provide input tags and see the result.

To achieve this:

All profile code has been reorganised into a class hierarchy. Of course LUA doesn't really have classes, but we can emulate them. The current hierarchy is as follows:

```
Base ----- Car
      \----Bicycle
       \---Foot
```

The Base class established a framework for how tags are processed, and call various methods to handle things like oneway, surface, etc. Sub-classes overwrite various methods when needed.

Processing a way basically consists of:

```
prepare_processing()
pre_process_combined()
process_each_direction()
post_process_combined()
output()
```

The move to class hierarchy makes it possible to pull in processing from other modes when necessary. Because you can push your bike, correct bicycle routing really requires full foot handling as well. It has always been a problem to get robust foot handling in the bike profile. Now the bike profile delegates this to the Foot class.

The file profiles/debug.lua is an initial demonstration of how the profile code can be run directly form the command line to easily check how a specific set of tags are processed. To support this, a log of messages can be written during processing, to make it easier to understand how the processing reaches a certain result.

However, this LUA debugging is currently broken, because there are a few C++ methods that the LUA code relies on, related to guidance.

Utility methods have been added for handling common tag processing like looking for forward/backward variants of a specific tag, and searching through the hierarchy of access like, eg. acces/vehicle/motorcar/car. Utility stuff like parsing max speed and durations have been moved to methods of the base Mode class, to make it easier for subclasses to modify the behaviour if needed, and to make it easier to integrate more closely between the various steps when required, but this approach could be discussed.

Each mode processing class have it's own set of configurations. The helper function deep_merge() makes it easy for subclasses to modify/add the config.

Each direction (forward/backward) of a way is processed separately, ie. with the same code called twice with different direction objects. This avoids a lot of code duplication where code handle first forward, then backward. It also makes it easier to make sure that forward/backward variants are handled in all relevant cases.

Running all tests after doing 'rm -r test/cache' shows no significant difference in processing time between this branch and master, but we need to benchmark with realistic data sizes.
## Tasklist
- [ ] discuss the overall approach/architecture.
- [ ] benchmark with real-world data set.
- [ ] clean up integration of turn lanes and classification (current solution is quick and dirty)
- [ ] decide whether we want this merged.
- [ ] iron on remaining test cases.
- [ ] review
- [ ] adjust for comments
## Requirements / Relations

Need to sync with recent changes to profiles.
Supersedes #2032.


emiltin:
I don't understand exactly what the C++ trimLaneString() and applyAccessTokens() methods do, they are called from guidance.lua. Is there a particular reason these two methods are implemented in C++?

MoKob:
@emiltin the split routines for strings in lua are just ugly. It was way simpler to implement the logic we need here in OSRM. If you can replicate the same behaviour in lua, I'd be fine with it.

emiltin:
@MoKob Ok.
What's the status on moving profiles to JS - is this PR still relevant?

MoKob:
Due to JS limitations regarding I/O (e.g. querying databases), we can't move off of `lua` in the near future.
From a short conversation with @daniel-j-h, we consider this work very relevant still (sorry for the lack of communication in the recent days).

From my point of view, the class based approach looks reasonable. I can't fully grasp yet whether it could complicate any handling, but I guess that feeling is mostly based on paranoia. @emiltin can you think of anything in the profiles that could get limited in its capabilities due to this refactor?

emiltin:
Well, the current profiles still work, since the C++ interface is the same. So if you want/need you can bypass the class framework entirely and work on the 'bare metal', ie. I don't think there's anything this refactor your limit.

I think a few of the lua lib files have been changed. If we wan't perfect backwards compatibility then we should avoid that.

Preliminary benchmarking shows that extracting is slower. But according to some quick LUA profiling, around 50% of time is spend in tag(), which is simply a wrapper for the C++ utility get_value_by_key(). So it seems it should be possible to improve the performance.

MoKob:
Can you clarify `extracting is slower`? What kind of time differences are we talking about here?

For the libraries, I feel that the compatibility guarantees end at the `OSRM : lua` boundary. The capability of lua is just to wide of an area to let anything in the profiles define our backwards compatibility. Otherwise we couldn't change any constants/penalties/arrays without being a breaking change.

emiltin:
Currently about 60% slower. The goal should be that it's not significantly slower. Here's a few timings of osrm-extract on a 19MB .pbf file:

```
> ~/code/osrm-backend/sandbox$ (rewrite/profiles) ls -l luxembourg.osm.pbf 
-rw-r--r--  1 emil  staff  20300968 Nov  2 14:24 luxembourg.osm.pbf

> ~/code/osrm-backend/sandbox$ (master) time ../build/osrm-extract luxembourg.osm.pbf -p ../profiles/car.lua
real	0m13.007s
user	0m25.620s
sys	0m1.738s

real	0m15.148s
user	0m25.603s
sys	0m2.366s

real	0m15.000s
user	0m25.433s
sys	0m2.307s

> ~/code/osrm-backend/sandbox$ (rewrite/profiles) time ../build/osrm-extract luxembourg.osm.pbf -p ../profiles/car.lua
real	0m24.849s
user	0m58.527s
sys	0m2.597s

real	0m23.984s
user	0m58.634s
sys	0m2.520s

real	0m24.056s
user	0m59.159s
sys	0m2.595s
```

I'm trying to use use luaprofiler http://luaprofiler.luaforge.net. Is there a way to get the current thread number from LUA?

MoKob:
Given how long the planet already takes, we would definitely have to invest here. This slowdown would be hard to swallow. Based on the readability alone, I already feel this is a huge step into the right direction.

emiltin:
down from 60% to around 40% increase in processing time after some easy improvements.

TheMarex:
On thing that comes to mind is the heavy use of sub-classing and OOP paradigms. Utilizing inheritance based schemes here is not really necessary, since we never actually use polymorphism. In this case it is much faster to only rely on free helper functions.

This would also make the profiles easier to read. I have a really hard time following down the abstractions to see what is actually going on. Instead of hiding the invocations in abstracted function calls, I think it makes much more sense to show the actual call to helper functions.

My advice would be to get rid of OOP scheme and solely rely on encapsulation of functionality in helper functions, achieving better performance and readability.

emiltin:
One of the main goal with this rewrite was to avoid all the error prone duplication of complex logic across profiles, by establishing a common processing flow. For example, in which order do you handle access, oneway, way type, speed, etc? How do you handle the two different directions and directional tag variants? Solving this in each profile causes a lot of bugs and maintenance problems.

In my view a natural way to solve this is with a base class that sets up a a common processing flow with sub classes (modes) overriding relevant methods. I think this is harder to achieve only with helper functions?

Classes also make it easier to reuse processing elements between modes. For example, the bike profile uses the walk profile (class) to correctly determine where you can push the bike, without duplicating the code.

When it comes to performance, I think moving some of the tag utility methods to C++ would provide a bigger gain. Methods like find_directional_tag() which are called a lot currently take up a a significant percentage of processing time.

Regarding readability, the fact is that the tag processing IS complex. In my view classes with small methods is a useful way to handle this. Of course it means you get longer call chains. As long as each method has a clear purpose I don't think it makes the code harder to read.

TheMarex:
> Classes also make it easier to reuse processing elements between modes. 

This is completely independent of using classes. Let me give an example:

```
local foot = require('foot');

function way_function(result):
    set_accessibility(result, access_tags)
    if result.not_accessible() then
        set_accessibility(result, foot.access_tags)
    end
end
```

This is how this could look. Now let's chase that precedence down in the current code:

1. `way_function` just defers to an opaque function https://github.com/Project-OSRM/osrm-backend/pull/3172/files#diff-d9b22a7d341f39bc1a582ce3e278e0b9R31
2. Chasing down that class leads us to `lib/bicycle.lua` which doesn't have a `process` function
3. Opening it's base class we find `processing` which defers to `prepare_processing`, `pre_process_combined` and `process_each_direction`, each of which can be overridden by a sub-class
4. Let's check out `prepare_processing`. Indeed it is overridden by the Bicycle class. First line calls back to `Base` version of `prepare_processing`. And there is a call to the foot profile that sets a temporary `self.tmp.foot`
5. Now let's try to find out where `self.tmp.foot` comes into play and it is not referenced anymore in that class?
6. At this point I literally have up to try to chase down that behavior.

You have to do the above step for anything you want to look up. You can hopefully see why I have a hard time following the code.

Procedural style is exactly what the profiles are meant to capture. It is undeniable that this needs a refactor into sub-functions that can be reused between profiles, but hiding that away behind a long call chain through abstractions is completely overkill.

> When it comes to performance, I think moving some of the tag utility methods to C++ would provide a bigger gain. 

There is a real overhead of context switches between lua and back to native land. Apart from the maintenance overhead of moving more to C++-land (which is exactly what the profiles try to avoid). I don't think this will yield any significant speedups.

There is a lot of incremental improvements we can do before we launch into a head-on refactor that does not make the code easier to understand. To get this into mainline I would suggest you break up your refactor into smaller changes, that we can actually review independently and track down regression.

emiltin:
Alright I will close this PR then and we can work on more incremental refactoring.

> There is a real overhead of context switches between lua and back to native land. Apart from the maintenance overhead of moving more to C++-land (which is exactly what the profiles try to avoid). I don't think this will yield any significant speedups.

You're probably right. Maybe this insight could be used to speed up the current code. The C++ method get_value_by_key() is called a large number of times, so maybe a speed up could be achieved by making it possible to fetch a number of tag in one go.

Issue #3176, opened 2016-10-25
Map matching reports wrong OSM ids in annotation


TheMarex:
The annotation returned by the map matching will return completely wrong ids. There will be duplicates, the ids will be in the wrong order and all of the above combined. See the request below for an example:

Example request on Berlin:

http://127.0.0.1:5000/match/v1/driving/13.407810,52.531632;13.408201,52.531508;13.408470,52.531394;13.408711,52.531332;13.408754,52.531316;13.408684,52.531319;13.408792,52.531286;13.408733,52.531303;13.408792,52.531322;13.409065,52.531208;13.409500,52.531081;13.409832,52.530989?annotations=true

```
{
    "code": "Ok",
    "matchings": [
        {
            "confidence": 0.94712,
            "distance": 171.7,
            "duration": 25.5,
            "geometry": "uas_IyuypAXkARw@Pk@@ICJFUCJ@ERy@ZsA",
            "legs": [
                {
                    "annotation": {
                        "datasources": [
                            0
                        ],
                        "nodes": [
                            2712265859,
                            26753080
                        ],
                        "duration": [
                            3.7
                        ],
                        "distance": [
                            29.771361
                        ]
                    },
                    "duration": 4.5,
                    "summary": "",
                    "distance": 29.8,
                    "steps": []
                },
                {
                    "annotation": {
                        "datasources": [
                            0
                        ],
                        "nodes": [
                            2712265859,
                            26753080
                        ],
                        "duration": [
                            0.4
                        ],
                        "distance": [
                            22.140556
                        ]
                    },
                    "duration": 3.3,
                    "summary": "",
                    "distance": 22.1,
                    "steps": []
                },
                {
                    "annotation": {
                        "datasources": [
                            0,
                            0
                        ],
                        "nodes": [
                            2712265859,
                            2712265859,
                            2712265859
                        ],
                        "duration": [
                            0.5,
                            2.3
                        ],
                        "distance": [
                            3.064004,
                            14.451721
                        ]
                    },
                    "duration": 2.6,
                    "summary": "",
                    "distance": 17.5,
                    "steps": []
                },
                {
                    "annotation": {
                        "datasources": [
                            0
                        ],
                        "nodes": [
                            26753078,
                            2712265859
                        ],
                        "duration": [
                            1.8
                        ],
                        "distance": [
                            3.41058
                        ]
                    },
                    "duration": 0.5,
                    "summary": "",
                    "distance": 3.4,
[...]

PR #3177, opened 2016-10-25
Fix annotations node ids

TheMarex:
# Issue

Addresses issue #3176
## Tasklist
- [x] review
- [x] adjust for comments

This is waiting on a fix that will unbreak the matching tests.

### DIFF
--- a/features/testbot/matching.feature
+++ b/features/testbot/matching.feature
@@ -235,3 +235,66 @@ Feature: Basic Map Matching
         When I match I should get
             | trace | timestamps | matchings |
             | abcd  | 0 1 2 3    | abcd      |
+
+    # Regression test 1 for issue 3176
+    Scenario: Testbot - multiuple segments: properly expose OSM IDs
+        Given the query options
+            | annotations | true    |
+
+        Given the node map
+            """
+            a-1-b--c--d--e--f-2-g
+            """
+
+        And the nodes
+            | node | id |
+            | a    | 1  |
+            | b    | 2  |
+            | c    | 3  |
+            | d    | 4  |
+            | e    | 5  |
+            | f    | 6  |
+            | g    | 7  |
+
+        And the ways
+            | nodes | oneway |
+            | ab    | no     |
+            | bc    | no     |
+            | cd    | no     |
+            | de    | no     |
+            | ef    | no     |
+            | fg    | no     |
+
+        When I match I should get
+            | trace | OSM IDs       |
+            | 12    | 1,2,3,4,5,6,7 |
+            | 21    | 7,6,5,4,3,2,1 |
+
+    # Regression test 2 for issue 3176
+    Scenario: Testbot - same edge: properly expose OSM IDs
+        Given the query options
+            | annotations | true    |
+
+        Given the node map
+            """
+            a-1-b--c--d--e-2-f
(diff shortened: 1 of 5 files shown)


features/testbot/matching.feature
```diff
@@ -169,3 +169,47 @@ Feature: Basic Map Matching
         When I match I should get
             | trace | matchings | geometry                                   |
             | abd   | abd       | 1,1,1.000089,1,1.000089,1,1.000089,0.99991 |
+
+
+    # Regression test 1 for issue 3176
+    Scenario: Testbot - multiuple segments: properly expose OSM IDs
```
MoKob: multiple

features/testbot/matching.feature
```diff
@@ -169,3 +169,47 @@ Feature: Basic Map Matching
...
+            | ab    | no     |
+            | bc    | no     |
+            | cd    | no     |
+            | de    | no     |
+            | ef    | no     |
+            | fg    | no     |
+
+        When I match I should get
+            | trace | OSM IDs       |
+            | 12    | 1,2,3,4,5,6,7 |
```
MoKob: can the order in which nodes are generated from the above docstrings ever change? Or should we use the addtional `id` table to ensure fixed ids?

TheMarex: Technically we don't guarantee this, but we already rely on this implementation detail at multiple places. I feel this is fine for a first step and we can ticket adding an explicit table.

MoKob: I kind of disagree here since I don't feel like continuing bad behaviour, simply since it's done bad elsewhere as well,  is a valid argument.

Adding a table for
```
        And the nodes
            | node | id |
            | a    | 1  |
            | b    | 2  |
            | c    | 3  |
            | d    | 4  |
            | e    | 5  |
            | f    | 6  |
            | g    | 7  |
```
does not seem to be worth ticketing for later to me.

danpat: If we're going to depend on the node ID values in the tests, we should make sure they're stable.  I suspect the results are deterministic right now, but the behaviour is implied.

Looks like there is *some* logic to generate consistent IDs here:
https://github.com/Project-OSRM/osrm-backend/blob/master/features/support/data.js#L43

But the logic seems incomplete.

@TheMarex if you don't have time to clear this up, I can look at adding it to this PR today.  I think we should fix this at the same time.

TheMarex: Making our tests more consistent is an important goal, but it does not take precedence over fixing the bug for me.

This works a regression test and will not be flunky since I know the ids will be generated that way. This is more about documenting it to the outside world and ensuring this will always work in the future even if we change the behavior of the support code. These are separate issues from having a working regression test.

For reference consistency is ensured over [this](https://github.com/Project-OSRM/osrm-backend/blob/master/features/step_definitions/data.js#L54-L64). Nodes are processed from left to right, not the code you linked @danpat.

TheMarex: Actually we have that. Just added it.

daniel-j-h: Hrm could you add docs for this? Also what about bringing the Cucumber docs to the repo. instead of having it (potentially outdated?) in the Wiki?

Issue #3179, opened 2016-10-26
Demo Server is Hitting Assertion in Extraction


MoKob:
```[assert] /mnt/ebs/osrm/2016-10-25-181942/src/extractor/guidance/turn_handler.cpp:374
in: osrm::extractor::guidance::Intersection osrm::extractor::guidance::TurnHandler::assignRightTurns(EdgeID, osrm::extractor::guidance::Intersection, std::size_t) const: first_direction != third_direction```

Extracting the planet, we see above assertion on current master.

PR #3181, opened 2016-10-26
fix breaking the sorting order by node adjustments

MoKob:
# Issue

Resolves https://github.com/Project-OSRM/osrm-backend/issues/3179.

It turns out that we were able to destroy our sorting within the temporary intersection structure during road adjustments.

## Tasklist
 - [x] review
 - [ ] adjust for comments

## Requirements / Relations
 none

### DIFF
--- a/src/extractor/guidance/intersection_generator.cpp
+++ b/src/extractor/guidance/intersection_generator.cpp
@@ -578,17 +578,38 @@ Intersection IntersectionGenerator::AdjustForJoiningRoads(const NodeID node_at_i
         if (range.size() <= 1)
             continue;
 
+        // the order does not matter
+        const auto get_offset = [](const ConnectedRoad &lhs, const ConnectedRoad &rhs) {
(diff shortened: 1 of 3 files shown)


src/extractor/guidance/intersection_generator.cpp
```diff
@@ -597,10 +605,16 @@ Intersection IntersectionGenerator::AdjustForJoiningRoads(const NodeID node_at_i
                           next_intersection_along_road[0].turn.angle,
                           next_intersection_along_road[next_intersection_along_road.size() - 1]
                               .turn.angle);
+
+            const auto max_offset =
+                angularDeviation(road.turn.angle, intersection[index - 1].turn.angle);
+            const auto corrected_offset = (offset + MAXIMAL_ALLOWED_NO_TURN_DEVIATION > max_offset)
+                                              ? 0.5 * max_offset
+                                              : offset;
```
daniel-j-h: Want to abstract that duplicate code into a function?

PR #3199, opened 2016-10-28
Refactors the Sliproad Handler

daniel-j-h:
Refactors the Sliproad Handler.

cc @MoKob

### DIFF
--- a/features/guidance/advanced-lanes.feature
+++ b/features/guidance/advanced-lanes.feature
@@ -5,13 +5,17 @@ Feature: Turn Lane Guidance
         Given the profile "car"
         Given a grid size of 3 meters
 
+
     @sliproads
     Scenario: Separate Turn Lanes
         Given the node map
             """
                           e
-            a     b       c g
-                          d
+                          .
+            a ... b ..... c . g
+                    `     .
+                     `... d
+                          .
                           f
             """
 
@@ -41,8 +45,10 @@ Feature: Turn Lane Guidance
         Given the node map
             """
                           e
-            a     b       c g
-                          d
+            a . . b . . . c g
+                    `     .
+                       `  .
+                        ` d
                           f
             """
 
@@ -66,22 +72,23 @@ Feature: Turn Lane Guidance
             | a,e       | in,cross,cross       | depart,turn left,arrive         | ,left:true straight:false right:false, |
             | a,g       | in,straight,straight | depart,new name straight,arrive | ,left:false straight:true right:false, |
             | a,f       | in,cross,cross       | depart,turn right,arrive        | ,left:false straight:false right:true, |
+
 
     @sliproads
     Scenario: Separate Turn Lanes Next to other turns
         Given the node map
             """
-                          e
-            a - - b.-.- - c-g
-                  |    ' 'd
-                  |       f
-                  |
-                  |
-                  |
-                  |
-                  |
-                  |
-            i - - h - - - j
+                        . e
+            a . . b . . . c g
+                  .   `   .
+                  .     ` .
+                  .       d
+                  .       f
+                  .
+                  .
+                  .
+                  .
+            i . . h . . . j
             """
 
         And the ways
@@ -108,6 +115,7 @@ Feature: Turn Lane Guidance
             | a,f       | in,cross,cross       | depart,turn right,arrive            | ,left:false straight:false right:true, |
             | a,j       | in,turn,other,other  | depart,turn right,turn left,arrive  | ,,left:true right:false,               |
             | a,i       | in,turn,other,other  | depart,turn right,turn right,arrive | ,,left:false right:true,               |
+
 
     @todo @2654 @none
     #https://github.com/Project-OSRM/osrm-backend/issues/2645
@@ -230,6 +238,7 @@ Feature: Turn Lane Guidance
             | a,h       | ghough,market,market | depart,turn slight right,arrive    | ,none:false straight:false straight:true straight:true,  |
             | a,j       | ghough,market,market | depart,turn left,arrive            | ,none:true straight:false straight:false straight:false, |
             | a,f       | ghough,ghough,ghough | depart,continue slight left,arrive | ,none:true straight:true straight:false straight:false,  |
+
 
     Scenario: Check sliproad handler loop's exit condition, Issue #2896
       # http://www.openstreetmap.org/way/198481519
--- a/features/guidance/collapse.feature
+++ b/features/guidance/collapse.feature
@@ -701,12 +701,15 @@ Feature: Collapse
         Given the node map
             """
                       g
-
+                      .
                       c
-            a     b
-                        d
+            a . . b .'
+                      ` d.
                         f e
             """
+            # as it is right now we don't classify this as a sliproad,
+            # check collapse-detail.feature for a similar test case
+            # which removes the fork here due to it being a Sliproad.
 
         And the ways
             | nodes | name  | oneway | highway   |
@@ -725,10 +728,9 @@ Feature: Collapse
 
         When I route I should get
             | waypoints | route                 | turns                                      |
-            | a,g       | road,cross,cross      | depart,turn left,arrive                    |
-            | a,e       | road,road,road        | depart,continue straight,arrive            |
-            # We should discuss whether the next item should be collapsed to depart,turn right,arrive.
-            | a,f       | road,road,cross,cross | depart,continue straight,turn right,arrive |
+            | a,g       | road,cross,cross      | depart,fork left,arrive                    |
+            | a,e       | road,road,road        | depart,fork slight right,arrive            |
+            | a,f       | road,road,cross,cross | depart,fork slight right,turn right,arrive |
 
     Scenario: On-Off on Highway
         Given the node map
--- a/include/extractor/guidance/sliproad_handler.hpp
+++ b/include/extractor/guidance/sliproad_handler.hpp
@@ -11,6 +11,8 @@
 
 #include <vector>
 
+#include <boost/optional.hpp>
+
 namespace osrm
 {
 namespace extractor
@@ -20,7 +22,7 @@ namespace guidance
 
 // Intersection handlers deal with all issues related to intersections.
 // They assign appropriate turn operations to the TurnOperations.
-class SliproadHandler : public IntersectionHandler
+class SliproadHandler final : public IntersectionHandler
 {
   public:
     SliproadHandler(const IntersectionGenerator &intersection_generator,
@@ -40,6 +42,37 @@ class SliproadHandler : public IntersectionHandler
     Intersection operator()(const NodeID nid,
                             const EdgeID via_eid,
                             Intersection intersection) const override final;
+
+  private:
+    boost::optional<std::size_t> getObviousIndexWithSliproads(const EdgeID from,
+                                                              const Intersection &intersection,
+                                                              const NodeID at) const;
+
+    // Next intersection from `start` onto `onto` is too far away for a Siproad scenario
+    bool nextIntersectionIsTooFarAway(const NodeID start, const EdgeID onto) const;
+
+    // Through street: does a road continue with from's name at the intersection
+    bool isThroughStreet(const EdgeID from, const IntersectionView &intersection) const;
+
+    // Does the road from `current` to `next` continue
+    bool roadContinues(const EdgeID current, const EdgeID next) const;
+
+    // Is the area under the triangle a valid Sliproad triangle
+    bool isValidSliproadArea(const double max_area, const NodeID, const NodeID, const NodeID) const;
+
+    // Is the Sliproad a link the both roads it shortcuts must not be links
+    bool isValidSliproadLink(const IntersectionViewData &sliproad,
+                             const IntersectionViewData &first,
+                             const IntersectionViewData &second) const;
+
+    // Could a Sliproad reach this intersection?
+    static bool canBeTargetOfSliproad(const IntersectionView &intersection);
+
+    // Scales a threshold based on the underlying road classification.
+    // Example: a 100 m threshold for a highway if different on living streets.
+    // The return value is guaranteed to not be larger than `threshold`.
+    static double scaledThresholdByRoadClass(const double max_threshold,
+                                             const RoadClassification &classification);
 };
 
 } // namespace guidance
--- a/src/extractor/guidance/intersection.cpp
+++ b/src/extractor/guidance/intersection.cpp
@@ -1,13 +1,7 @@
 #include "extractor/guidance/intersection.hpp"
 
-#include <boost/range/adaptor/transformed.hpp>
-#include <boost/range/algorithm/find_if.hpp>
-
-#include <boost/assert.hpp>
-
-#include <algorithm>
-#include <functional>
 #include <limits>
+#include <string>
 
 using osrm::util::angularDeviation;
 
@@ -69,34 +63,6 @@ std::string toString(const ConnectedRoad &road)
               std::to_string(static_cast<std::int32_t>(road.instruction.direction_modifier)) + " " +
               std::to_string(static_cast<std::int32_t>(road.lane_data_id));
     return result;
-}
-
-bool Intersection::valid() const
-{
-    return !empty() &&
-           std::is_sorted(begin(), end(), std::mem_fn(&ConnectedRoad::compareByAngle)) &&
-           operator[](0).angle < std::numeric_limits<double>::epsilon();
-}
-
-std::uint8_t
-Intersection::getHighestConnectedLaneCount(const util::NodeBasedDynamicGraph &graph) const
-{
-    BOOST_ASSERT(valid()); // non empty()
-
-    const std::function<std::uint8_t(const ConnectedRoad &)> to_lane_count =
-        [&](const ConnectedRoad &road) {
-            return graph.GetEdgeData(road.eid).road_classification.GetNumberOfLanes();
-        };
-
-    std::uint8_t max_lanes = 0;
-    const auto extract_maximal_value = [&max_lanes](std::uint8_t value) {
-        max_lanes = std::max(max_lanes, value);
-        return false;
-    };
-
-    const auto view = *this | boost::adaptors::transformed(to_lane_count);
-    boost::range::find_if(view, extract_maximal_value);
-    return max_lanes;
 }
 
 } // namespace guidance
--- a/src/extractor/guidance/motorway_handler.cpp
+++ b/src/extractor/guidance/motorway_handler.cpp
@@ -193,7 +193,7 @@ Intersection MotorwayHandler::fromMotorway(const EdgeID via_eid, Intersection in
             assignFork(via_eid, intersection[3], intersection[2], intersection[1]);
         }
 
-        else if (countValid(intersection) > 0) // check whether turns exist at all
+        else if (intersection.countEnterable() > 0) // check whether turns exist at all
         {
             // FALLBACK, this should hopefully never be reached
             return fallback(std::move(intersection));
@@ -338,7 +338,7 @@ Intersection MotorwayHandler::fromMotorway(const EdgeID via_eid, Intersection in
 
 Intersection MotorwayHandler::fromRamp(const EdgeID via_eid, Intersection intersection) const
 {
-    auto num_valid_turns = countValid(intersection);
+    auto num_valid_turns = intersection.countEnterable();
     // ramp straight into a motorway/ramp
     if (intersection.size() == 2 && num_valid_turns == 1)
     {
--- a/src/extractor/guidance/sliproad_handler.cpp
+++ b/src/extractor/guidance/sliproad_handler.cpp
@@ -4,11 +4,14 @@
 #include "util/coordinate_calculation.hpp"
 #include "util/guidance/name_announcements.hpp"
 
+#include <cmath>
+
+#include <algorithm>
+#include <iterator>
 #include <limits>
 
 #include <boost/assert.hpp>
 
-using EdgeData = osrm::util::NodeBasedDynamicGraph::EdgeData;
 using osrm::extractor::guidance::getTurnDirection;
 using osrm::util::angularDeviation;
 
@@ -32,7 +35,8 @@ SliproadHandler::SliproadHandler(const IntersectionGenerator &intersection_gener
 {
 }
 
-// included for interface reasons only
+// The intersection has to connect a Sliproad, see the example scenario below:
+// Intersection at `d`: Sliproad `bd` connecting `cd` and the road starting at `d`.
 bool SliproadHandler::canProcess(const NodeID /*nid*/,
                                  const EdgeID /*via_eid*/,
                                  const Intersection &intersection) const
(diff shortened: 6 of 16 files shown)


src/extractor/guidance/sliproad_handler.cpp
```diff
@@ -34,221 +36,152 @@ SliproadHandler::SliproadHandler(const IntersectionGenerator &intersection_gener
...
-    if (obvious_turn_index == 0)
-        return intersection;
+    const auto source_edge_data = node_based_graph.GetEdgeData(source_edge_id);
 
-    const auto &next_road = intersection[obvious_turn_index];
+    // The road leading to the intersection (bc) has to continue from our source
+    {
+        auto same_road_category =
+            next_road_edge_data.road_classification == source_edge_data.road_classification;
+        auto same_name = next_road_edge_data.name_id != EMPTY_NAMEID &&
```
MoKob: should we use `requiresNameAnnounced` here as well?

daniel-j-h: Yeah this will be done in https://github.com/Project-OSRM/osrm-backend/pull/2945 - I just haven't rebased and finished it yet.

There will be some conflicts, for sure.

daniel-j-h:
Squashed and rebased against master - starting to sample real-world situations now looking for both false positives and negatives.

daniel-j-h:
Here's what we currently classify as sliproad scenarios in Berlin:

https://api.mapbox.com/styles/v1/danieljh/ciuyc3mrp01c92jl852h1ntu5.html?title=true&access_token=pk.eyJ1IjoiZGFuaWVsamgiLCJhIjoiTnNYb25JSSJ9.vYOnsuu1zeKcGW2nj0uJZw#12.17/52.5081/13.3670

MoKob:
Glad we have this visualisation, there quite a few strange ones in there

daniel-j-h:
Now with restricting the Sliproad classification to ramps and links:

https://api.mapbox.com/styles/v1/danieljh/ciuznn1mg01ga2iqqpdi9rsr4.html?title=true&access_token=pk.eyJ1IjoiZGFuaWVsamgiLCJhIjoiTnNYb25JSSJ9.vYOnsuu1zeKcGW2nj0uJZw#12.17/52.5081/13.3670

daniel-j-h:
Latest visualization is here:

https://api.mapbox.com/styles/v1/danieljh/civ2i4gf000402ino5qir8u1r.html?title=true&access_token=pk.eyJ1IjoiZGFuaWVsamgiLCJhIjoiTnNYb25JSSJ9.vYOnsuu1zeKcGW2nj0uJZw#11.07/52.5107/13.4348

Pink are old(er) commits, blue is current state:
- taking link / ramp into account
- fixes for discarding Forks
- narrow angle for turn off the Sliproad and back onto the street

Next checks to try out:
- [ ] turn off of the Sliproad `ade` has to be obvious (implies entry_allowed)
- [ ] non-sliproad `cde` has to be straight-ish (~15 degree) for turn at intersection `c` where sliproad joins
- [ ] test curvature, sliproad `ad` has to snuggle against roads it shortcuts `abc` `cde` (check ccw / cw, fix start `a`, end `d`)

General fixes that need to land:
- [ ] Graph walker for skipping traffic lights on all roads involved
- [ ] Graph walker for finding next intersection / distance check

---------------------------------------------------------------------------------------------------------

Some naming conventions used here

```
       .
       e
       .
       .
 a ... b .... c .   < `next` is intersection at `c`
       `      .
         `    .
           `  .
              d     < `target_intersection` is intersection at `d`
              .
              e


          ^ `sliproad` is `bd`
       ^ `intersection` is intersection at `b`
```

src/extractor/guidance/sliproad_handler.cpp
```diff
@@ -135,13 +135,13 @@ operator()(const NodeID /*nid*/, const EdgeID source_edge_id, Intersection inter
     if (intersection_node_id == next->node)
         return intersection;
 
-    std::unordered_set<NameID> target_road_names;
+    std::vector<NameID> target_road_name_ids;
+    target_road_name_ids.reserve(next->intersection.size());
```
daniel-j-h: @MoKob I checked the sizes on Berlin, turns out it's really small.
We're fine with reserving here and then simply looping below.

Ticketed the `string_ref` idea in https://github.com/Project-OSRM/osrm-backend/issues/3265.

daniel-j-h:
Latest snapshot:

- [Berlin](https://api.mapbox.com/styles/v1/danieljh/civ2i4gf000402ino5qir8u1r.html?title=true&access_token=pk.eyJ1IjoiZGFuaWVsamgiLCJhIjoiTnNYb25JSSJ9.vYOnsuu1zeKcGW2nj0uJZw#10.85/52.5121/13.4983)
- [Bay Area](https://api.mapbox.com/styles/v1/danieljh/civ2i4gf000402ino5qir8u1r.html?title=true&access_token=pk.eyJ1IjoiZGFuaWVsamgiLCJhIjoiTnNYb25JSSJ9.vYOnsuu1zeKcGW2nj0uJZw#10.29/37.7060/-122.2671)

There are some ideas left to explore as noted in https://github.com/Project-OSRM/osrm-backend/pull/3199#issuecomment-258197682.

daniel-j-h:
Here is a current snapshot with additional heuristics implemented:

https://api.mapbox.com/styles/v1/danieljh/civ2i4gf000402ino5qir8u1r.html?title=true&access_token=pk.eyJ1IjoiZGFuaWVsamgiLCJhIjoiTnNYb25JSSJ9.vYOnsuu1zeKcGW2nj0uJZw#10.411588333960331/52.51237533716511/13.416338076773059/0

@MoKob - Red before, Green at the moment.
Also check out SF for how it looks currently.

MoKob:
This looks way better, I've found a few issues though:

- [ ] should be possible to filter out given the size of the intersection at sliproad end (here 4). I'd expect a sliproad to have intersection size 3. Did you check for this? Or did you find it to be different?
https://api.mapbox.com/styles/v1/danieljh/civ2i4gf000402ino5qir8u1r.html?title=true&access_token=pk.eyJ1IjoiZGFuaWVsamgiLCJhIjoiTnNYb25JSSJ9.vYOnsuu1zeKcGW2nj0uJZw#17.47/52.51539/13.59450

- [ ] we could possibly limit the minimum street category, or at least make low street-categories only consider very low distances
https://api.mapbox.com/styles/v1/danieljh/civ2i4gf000402ino5qir8u1r.html?title=true&access_token=pk.eyJ1IjoiZGFuaWVsamgiLCJhIjoiTnNYb25JSSJ9.vYOnsuu1zeKcGW2nj0uJZw#17.56/52.61139/13.48635
https://api.mapbox.com/styles/v1/danieljh/civ2i4gf000402ino5qir8u1r.html?title=true&access_token=pk.eyJ1IjoiZGFuaWVsamgiLCJhIjoiTnNYb25JSSJ9.vYOnsuu1zeKcGW2nj0uJZw#17.85/52.56741/13.21783

- [ ] the edge is the sliproad, right here the road does not allow the proposed turn, even without the turn restriction. This could be easily checked by looking at the `reversed` flag. A road that connects to the sliproad turn cannot be reversed. https://api.mapbox.com/styles/v1/danieljh/civ2i4gf000402ino5qir8u1r.html?title=true&access_token=pk.eyJ1IjoiZGFuaWVsamgiLCJhIjoiTnNYb25JSSJ9.vYOnsuu1zeKcGW2nj0uJZw#19.16/52.42785/13.40512 looks like the turn to the `left` does only allow a u-turn here. 

- [ ] the sliproad should not be the end of a different edge:
https://api.mapbox.com/styles/v1/danieljh/civ2i4gf000402ino5qir8u1r.html?title=true&access_token=pk.eyJ1IjoiZGFuaWVsamgiLCJhIjoiTnNYb25JSSJ9.vYOnsuu1zeKcGW2nj0uJZw#17.66/52.37530/13.39308

features/guidance/advanced-lanes.feature
```diff
@@ -10,8 +10,11 @@ Feature: Turn Lane Guidance
         Given the node map
             """
                           e
-            a     b       c g
-                          d
+                          .
+            a ... b ..... c . g
+                    `     .
```
MoKob: modification of test - intended? this increases distance between `c` and `d`

features/guidance/advanced-lanes.feature
```diff
@@ -41,8 +44,10 @@ Feature: Turn Lane Guidance
         Given the node map
             """
                           e
-            a     b       c g
-                          d
+            a . . b . . . c g
+                    `     .
+                       `  .
```
MoKob: even more so here. Please check all test modifications whether they are intended or not

features/guidance/collapse.feature
```diff
@@ -725,10 +725,10 @@ Feature: Collapse
 
         When I route I should get
             | waypoints | route                 | turns                                      |
-            | a,g       | road,cross,cross      | depart,turn left,arrive                    |
-            | a,e       | road,road,road        | depart,continue straight,arrive            |
+            | a,g       | road,cross,cross      | depart,fork left,arrive                    |
+            | a,e       | road,road,road        | depart,fork slight right,arrive            |
```
MoKob: this test here was supposed to collapse the turns `bcg` and `bde`,`bdf` which now does not happen anymore.

This change modifies the purpose of the test.

include/engine/api/route_api.hpp
```diff
@@ -152,6 +153,8 @@ class RouteAPI : public BaseAPI
                  */
 
                 guidance::trimShortSegments(steps, leg_geometry);
+                // TODO: remove before merging
+                util::guidance::print(steps);
```
MoKob: don't forget :)

include/extractor/guidance/sliproad_handler.hpp
```diff
@@ -9,10 +9,15 @@
 #include "util/name_table.hpp"
 #include "util/node_based_graph.hpp"
 
+#include "util/geojson_debug_logger.hpp"
+#include "util/geojson_debug_policies.hpp"
```
MoKob: remove before merge

include/extractor/guidance/sliproad_handler.hpp
```diff
@@ -42,6 +47,50 @@ class SliproadHandler : public IntersectionHandler
     Intersection operator()(const NodeID nid,
                             const EdgeID via_eid,
                             Intersection intersection) const override final;
+
+  private:
+    util::ScopedGeojsonLoggerGuard<util::NodeIdVectorToLineString> geojson_lines;
+    util::ScopedGeojsonLoggerGuard<util::NodeIdVectorToMultiPoint> geojson_points;
```
MoKob: remove before merge

include/extractor/guidance/sliproad_handler.hpp
```diff
@@ -42,6 +47,50 @@ class SliproadHandler : public IntersectionHandler
...
+
+    struct IntersectionAndNode final
+    {
+        Intersection intersection;
+        NodeID node;
+    };
+
+    // Returns a potential next intersection along a road skipping over traffic lights.
+    boost::optional<IntersectionAndNode> getNextIntersection(const NodeID at,
```
MoKob: should probably not be a part of the sliproad handler but at least of the base class, allowing all handlers to find next intersections

src/extractor/guidance/intersection.cpp
```diff
@@ -112,6 +117,45 @@ Intersection::getHighestConnectedLaneCount(const util::NodeBasedDynamicGraph &gr
     const auto view = *this | boost::adaptors::transformed(to_lane_count);
     boost::range::find_if(view, extract_maximal_value);
     return max_lanes;
+
+    return max_lanes;
```
MoKob: duplicated return?

src/extractor/guidance/intersection.cpp
```diff
@@ -112,6 +117,45 @@ Intersection::getHighestConnectedLaneCount(const util::NodeBasedDynamicGraph &gr
...
+}
+
+// Intersections are sorted roads: [0] being the UTurn road, then from sharp right to sharp left.
+
+ConnectedRoad &Intersection::getUTurnRoad()
+{
+    BOOST_ASSERT(valid());
```
MoKob: while technically reasonable, checking the validity at every such iterator could have a serious performance impact on debug builds/release with assertion builds.

I would say that we should rely on the generators to assert validity to not run into problems with debug builds/assertion builds on the demo server.

src/extractor/guidance/sliproad_handler.cpp
```diff
@@ -29,273 +32,720 @@ SliproadHandler::SliproadHandler(const IntersectionGenerator &intersection_gener
                           node_info_list,
                           name_table,
                           street_name_suffix_table,
-                          intersection_generator)
+                          intersection_generator),
+      geojson_lines{"sliproads-lines.geojson", node_info_list},
+      geojson_points{"sliproads-points.geojson", node_info_list}
```
MoKob: remove before merge

src/extractor/guidance/sliproad_handler.cpp
```diff
@@ -29,273 +32,720 @@ SliproadHandler::SliproadHandler(const IntersectionGenerator &intersection_gener
...
+    BOOST_ASSERT(intersection.size() > 2);
+
+    // Potential splitting / start of a Sliproad (b)
     auto intersection_node_id = node_based_graph.GetTarget(source_edge_id);
 
-    // if there is no turn, there is no sliproad
-    if (intersection.size() <= 2)
-        return intersection;
+    // Road index prefering non-sliproads (bc)
+    auto obvious = getObviousIndexWithSliproads(source_edge_id, intersection, intersection_node_id);
```
MoKob: since this returns an optional, I feel that `getObviousIndex` could be misleading as a name. We do not have a naming convention here, so far as I know. But how about `checkIfHasObviousIndexWithSliproads`?

src/extractor/guidance/sliproad_handler.cpp
```diff
@@ -29,273 +32,720 @@ SliproadHandler::SliproadHandler(const IntersectionGenerator &intersection_gener
...
-            const auto couldBeSliproad = [](const Intersection &intersection) {
-                if (intersection.size() != 3)
-                    return false;
-                if ((intersection[1].entry_allowed && intersection[2].entry_allowed) ||
-                    intersection[0].entry_allowed)
-                    return false;
-                return true;
-            };
+    // Potential non-sliproad road (bc), leading to the intersection (c) the Sliproad (bd) shortcuts
+    const auto &next_road = intersection[*obvious];
```
MoKob: could probably also benefit from an improvement in naming over my original poor choice

src/extractor/guidance/sliproad_handler.cpp
```diff
@@ -29,273 +32,720 @@ SliproadHandler::SliproadHandler(const IntersectionGenerator &intersection_gener
...
-                return 1;
-            else
-                return 0;
-        }
-        else
-            return 0;
-    }();
+    // The intersection `c` has to be reachable
+    if (!next_road.entry_allowed)
+    {
```
MoKob: Are we using `{}` for single line `if` branches, or not? I feel that this is often inconsistent in our code-base. Just 5 lines above you use the other style. I'd prefer not having the additional braces here.

src/extractor/guidance/sliproad_handler.cpp
```diff
@@ -29,273 +32,720 @@ SliproadHandler::SliproadHandler(const IntersectionGenerator &intersection_gener
...
+        return intersection;
+    }
 
-    if (obvious_turn_index == 0)
+    // The road leading to the intersection (bc) has to continue from our source
+    if (!roadContinues(source_edge_id, next_road.eid))
+    {
```
MoKob: and here

src/extractor/guidance/sliproad_handler.cpp
```diff
@@ -29,273 +32,720 @@ SliproadHandler::SliproadHandler(const IntersectionGenerator &intersection_gener
...
+        // Narrow turn angle for road (bd) and guard against data issues (overlapping roads)
+        auto is_narrow = angularDeviation(road.angle, STRAIGHT_ANGLE) <= 2 * NARROW_TURN_ANGLE;
+        auto same_angle = angularDeviation(next_road.angle, road.angle) //
+                          <= std::numeric_limits<double>::epsilon();
 
-    bool hasNarrow =
-        std::find_if(intersection.begin(), intersection.end(), linkTest) != intersection.end();
-    if (!hasNarrow)
-        return intersection;
+        if (!is_narrow || same_angle)
```
MoKob: you've probably spent some time here, but what about an intersection that models a turn when the road turns?
```
           .   d
a - b
           `   c
```

Why would this not be a sliproad when

```
            . d
        .
a - b - - - - c
```

could be?

src/extractor/guidance/sliproad_handler.cpp
```diff
@@ -29,273 +32,720 @@ SliproadHandler::SliproadHandler(const IntersectionGenerator &intersection_gener
...
+            return false;
+
+        // Prevent from starting in or going onto a roundabout
+        auto onto_roundabout = hasRoundaboutType(road.instruction);
```
MoKob: this check should probably be done earlier, since it is a simple check on the instruction and does not require looking for an obvious index (which costs more)

src/extractor/guidance/sliproad_handler.cpp
```diff
@@ -29,273 +32,720 @@ SliproadHandler::SliproadHandler(const IntersectionGenerator &intersection_gener
...
+            return false;
 
-        const auto same_name = !util::guidance::requiresNameAnnounced(
-            road_edge_data.name_id, source_edge_data.name_id, name_table, street_name_suffix_table);
+        const auto road_data = node_based_graph.GetEdgeData(road.eid);
 
-        return road_edge_data.road_classification == source_edge_data.road_classification &&
-               road_edge_data.name_id != EMPTY_NAMEID && source_edge_data.name_id != EMPTY_NAMEID &&
-               same_name && road.entry_allowed;
+        auto is_roundabout = road_data.roundabout;
```
MoKob: duplicated check? Within roundabouts, you should have only size two connections.
Outside of roundabouts/on exits you would see a roundabout type instruction.

Also, If it should be necessary, I'd prefer doing a single function style check for `isRoundabout` instead of two separate checks.

src/extractor/guidance/sliproad_handler.cpp
```diff
@@ -29,273 +32,720 @@ SliproadHandler::SliproadHandler(const IntersectionGenerator &intersection_gener
...
-    // Threshold check, if the intersection is too far away, don't bother continuing
-    const auto coordinate_extractor = intersection_generator.GetCoordinateExtractor();
-    const auto next_road_length = util::coordinate_calculation::getLength(
-        coordinate_extractor.GetForwardCoordinatesAlongRoad(
-            node_based_graph.GetTarget(source_edge_id), next_road.eid),
-        &util::coordinate_calculation::haversineDistance);
-    if (next_road_length > MAX_SLIPROAD_THRESHOLD)
+    // If the intersection is too far away, don't bother continuing
+    if (nextIntersectionIsTooFarAway(intersection_node_id, next_road.eid))
     {
```
MoKob: brace style

src/extractor/guidance/sliproad_handler.cpp
```diff
@@ -29,273 +32,720 @@ SliproadHandler::SliproadHandler(const IntersectionGenerator &intersection_gener
...
 
-    // Threshold check, if the intersection is too far away, don't bother continuing
-    const auto coordinate_extractor = intersection_generator.GetCoordinateExtractor();
-    const auto next_road_length = util::coordinate_calculation::getLength(
-        coordinate_extractor.GetForwardCoordinatesAlongRoad(
-            node_based_graph.GetTarget(source_edge_id), next_road.eid),
-        &util::coordinate_calculation::haversineDistance);
-    if (next_road_length > MAX_SLIPROAD_THRESHOLD)
+    // If the intersection is too far away, don't bother continuing
+    if (nextIntersectionIsTooFarAway(intersection_node_id, next_road.eid))
```
MoKob: is this check valid? Looks like you are finding the next intersection below. You need to know about the full distance for this check, don't you? Or are you calculating the path to the next intersection twice? (In that case, it could be considered performance bug)

src/extractor/guidance/sliproad_handler.cpp
MoKob: Yeah. They are small. I just remember someone telling me to used `unordered_sets` for these ;)

src/extractor/guidance/sliproad_handler.cpp
```diff
@@ -29,273 +32,720 @@ SliproadHandler::SliproadHandler(const IntersectionGenerator &intersection_gener
...
-    if (intersection_node_id == next_intersection_node)
+    if (intersection_node_id == next->node)
         return intersection;
 
-    std::unordered_set<NameID> target_road_names;
+    std::vector<NameID> target_road_name_ids;
+    target_road_name_ids.reserve(next->intersection.size());
 
-    for (const auto &road : next_road_next_intersection)
+    for (const auto &road : next->intersection)
```
MoKob: can we potentially change `next` into `path_along_main_road` or something similar?

src/extractor/guidance/sliproad_handler.cpp
```diff
@@ -29,273 +32,720 @@ SliproadHandler::SliproadHandler(const IntersectionGenerator &intersection_gener
...
     {
         const auto &target_data = node_based_graph.GetEdgeData(road.eid);
-        target_road_names.insert(target_data.name_id);
+        target_road_name_ids.push_back(target_data.name_id);
     }
 
-    for (auto &road : intersection)
+    // TODO: remove geojson debugging
+    NodeID b, c, d;
```
MoKob: Please never use `b`,`c`,d` or other single-char variable names. Finding where they are used if you try to follow the code is near impossible. If you need these to make your cases simple, I'd prefer adding a lookup table at the top of the examples. (e.g. node_at_intersection (c), ...)

src/extractor/guidance/sliproad_handler.cpp
```diff
@@ -29,273 +32,720 @@ SliproadHandler::SliproadHandler(const IntersectionGenerator &intersection_gener
...
+    b = intersection_node_id;
+    c = next->node;
+
+    auto sliproad_found = false;
+
+    // Check all roads for Sliproads and assign appropriate TurnType
+    for (std::size_t road_index = 0, last = intersection.size(); road_index < last; ++road_index)
     {
-        if (linkTest(road))
+        auto &sliproad = intersection[road_index]; // this is what we're checking and assigning to
```
MoKob: `possible_sliproad` or `sliproad_candidate`?

src/extractor/guidance/sliproad_handler.cpp
```diff
@@ -29,273 +32,720 @@ SliproadHandler::SliproadHandler(const IntersectionGenerator &intersection_gener
...
+        //
+        //
+        //          ^ `sliproad` is `bd`
+        //       ^ `intersection` is intersection at `b`
+
+        if (!is_potential_link(sliproad))
+            continue;
+
+        // b -> d edge id
+        EdgeID sliproad_edge = sliproad.eid;
```
MoKob: `auto`?

src/extractor/guidance/sliproad_handler.cpp
```diff
@@ -29,273 +32,720 @@ SliproadHandler::SliproadHandler(const IntersectionGenerator &intersection_gener
...
+
+        // TODO: remove debugging
+        d = node_based_graph.GetTarget(sliproad_edge);
```
MoKob: should be something like `node_at_end_of_sliproad`

src/extractor/guidance/sliproad_handler.cpp
```diff
@@ -29,273 +32,720 @@ SliproadHandler::SliproadHandler(const IntersectionGenerator &intersection_gener
...
+
+        const auto target_intersection = [&](NodeID node) {
+            auto intersection = intersection_generator(node, sliproad_edge);
```
MoKob: don't you have logic for skipping over these? Also: there is a high potential for multiple traffic lights. The intersection_generator as well offers `GetActualNextIntersection` (could use a better name) which does exactly what you do here, a bit more efficiently. Here you are generating coordinates for all traffic lights, but you don't care about these.

src/extractor/guidance/sliproad_handler.cpp
```diff
@@ -29,273 +32,720 @@ SliproadHandler::SliproadHandler(const IntersectionGenerator &intersection_gener
...
+            if (intersection.size() == 2)
+            {
+                node = node_based_graph.GetTarget(sliproad_edge);
+                sliproad_edge = intersection[1].eid;
+                intersection = intersection_generator(node, sliproad_edge);
+            }
+            return intersection;
+        }(intersection_node_id);
+
+        const NodeID sliproad_edge_target = node_based_graph.GetTarget(sliproad_edge);
```
MoKob: you now have `d` and `sliproad_edge_target` both from `sliproad_edge`. Following these without better names can become difficult. I'd prefer `first_node_along_sliproad_candidate` and `node_at_next_intersection_along_sliproad`

src/extractor/guidance/sliproad_handler.cpp
```diff
@@ -29,273 +32,720 @@ SliproadHandler::SliproadHandler(const IntersectionGenerator &intersection_gener
...
+
+        // Distinct triangle nodes `bcd`
+        if (intersection_node_id == next->node || intersection_node_id == sliproad_edge_target ||
+            next->node == sliproad_edge_target)
```
MoKob: how does `next->node` interact with `d` and `sliproad_edge_target`?
Wouldn't `intersection_node_id == sliproad_edge_target` imply a self-loop? 😕

src/extractor/guidance/sliproad_handler.cpp
```diff
@@ -29,273 +32,720 @@ SliproadHandler::SliproadHandler(const IntersectionGenerator &intersection_gener
...
+
+        // If the sliproad candidate is a through street, we cannot handle it as a sliproad.
+        if (isThroughStreet(sliproad_edge, target_intersection))
+        {
+            continue;
+        }
+
+        // The turn off of the Sliproad has to be obvious and a narrow turn
+        {
+            const auto index = findObviousTurn(sliproad_edge, target_intersection);
```
MoKob: I'd go as far as to say it has to be the `only` turn available. Did you find any coincidences where larger intersections would be desirable?

src/extractor/guidance/sliproad_handler.cpp
```diff
@@ -29,273 +32,720 @@ SliproadHandler::SliproadHandler(const IntersectionGenerator &intersection_gener
...
+        //
+        // a ... b .... c .   a ... b .... c .   a ... b .... c .
+        //       `      .           `  .   .             .    .
+        //         `    .                . .           .      .
+        //           `  .                 ..             .    .
+        //              d                  d                . d
+        //
+        //                    Sliproad           Not a Sliproad
+        {
+            // Intersection is orderd: 0 is UTurn, then from sharp right to sharp left.
```
MoKob: some of these checks are quite expensive. Have you checked whether you can re-order some of them so that cheaper ones come first to exit early when possible?

Can you potentially move some of these into dedicated functions (detectors) and annotate them in `O(n)`. That could make finding a reasonable order easier.

src/extractor/guidance/sliproad_handler.cpp
```diff
@@ -29,273 +32,720 @@ SliproadHandler::SliproadHandler(const IntersectionGenerator &intersection_gener
...
+        // Check for area under triangle `bdc`.
+        //
+        // a ... b .... c .
+        //       `      .
+        //         `    .
+        //           `  .
+        //              d
+        //
+        const auto area_threshold = std::pow(
+            scaledThresholdByRoadClass(MAX_SLIPROAD_THRESHOLD, sliproad_data.road_classification),
```
MoKob: assuming the length of the sliproad scales linearly, I'd expect `scaledThreshold()/2` here to have a more consistent relation between sliproadLength (which is the threshold, right?) and triangle are.

src/extractor/guidance/sliproad_handler.cpp
```diff
@@ -29,273 +32,720 @@ SliproadHandler::SliproadHandler(const IntersectionGenerator &intersection_gener
...
+            continue;
+        }
+
+        // Check all roads at `d` if one is connected to `c`, is so `bd` is Sliproad.
+        for (const auto &candidate_road : target_intersection)
+        {
+            const auto &candidate_data = node_based_graph.GetEdgeData(candidate_road.eid);
+
+            // Name mismatch: check roads at `c` and `d` for same name
+            const auto name_mismatch = [&](const NameID road_name_id) {
```
MoKob: are you considering `name_a` -> `name_b` announcement does not necessarily require `name_b -> name_a` announcement.
Checking for `EmptyNameID` is missing here and should probably be here.

src/extractor/guidance/sliproad_handler.cpp
```diff
@@ -29,273 +32,720 @@ SliproadHandler::SliproadHandler(const IntersectionGenerator &intersection_gener
...
+            {
+                sliproad.instruction.type = TurnType::Sliproad;
+                sliproad_found = true;
+                break;
+            }
+            else
             {
-                const auto &candidate_data = node_based_graph.GetEdgeData(candidate_road.eid);
-                if (target_road_names.count(candidate_data.name_id) > 0)
+                const auto skip_traffic_light_intersection = intersection_generator(
```
MoKob: should probably also use `GetActualNextIntersection` from the intersection Generator

src/extractor/guidance/sliproad_handler.cpp
```diff
@@ -29,273 +32,720 @@ SliproadHandler::SliproadHandler(const IntersectionGenerator &intersection_gener
...
-            intersection[obvious_turn_index].instruction.type = TurnType::Suppressed;
+            intersection[*obvious].instruction.type = TurnType::Suppressed;
         }
     }
 
+    if (sliproad_found)
+    {
+        geojson_points.Write(std::vector<NodeID>{b, c, d});
+        geojson_lines.Write(std::vector<NodeID>{b, d});
+    }
```
MoKob: remove before merge

src/extractor/guidance/sliproad_handler.cpp
```diff
@@ -29,273 +32,720 @@ SliproadHandler::SliproadHandler(const IntersectionGenerator &intersection_gener
...
+//
+//  a ... tl ... b .. c
+//               .
+//               .
+//               d
+//
+//  ^ at
+//     ^ road
+boost::optional<SliproadHandler::IntersectionAndNode>
+SliproadHandler::getNextIntersection(const NodeID at, const ConnectedRoad &road) const
```
MoKob: can potentially be removed in favour of the intersection generator function. Feel free to adjust that part to also return node/intersection.

src/extractor/guidance/sliproad_handler.cpp
```diff
@@ -29,273 +32,720 @@ SliproadHandler::SliproadHandler(const IntersectionGenerator &intersection_gener
...
+
+    // Otherwise check if the road is forking into two and one of them is a Sliproad;
+    // then the non-Sliproad is the obvious one.
+    if (intersection.size() != 3)
+        return boost::none;
+
+    const auto forking = intersection[1].instruction.type == TurnType::Fork &&
+                         intersection[2].instruction.type == TurnType::Fork;
+
+    if (!forking)
```
MoKob: am I reading this correctly? You rely on forks to have a obvious index? That sounds strange to me.

src/extractor/guidance/sliproad_handler.cpp
```diff
@@ -29,273 +32,720 @@ SliproadHandler::SliproadHandler(const IntersectionGenerator &intersection_gener
...
+
+    if (canBeTargetOfSliproad(second->intersection) && at != first->node)
+        return boost::make_optional(std::size_t{1});
+
+    return boost::none;
+}
+
+bool SliproadHandler::nextIntersectionIsTooFarAway(const NodeID start, const EdgeID onto) const
+{
+    BOOST_ASSERT(start != SPECIAL_NODEID);
```
MoKob: many of these functions you use compute the next intersection. Given that this implies computing a new intersection, I'd prefer if we can reduce that check to a single computation. After all that is all I am trying right now to improve performance again.

src/extractor/guidance/sliproad_handler.cpp
```diff
@@ -29,273 +32,720 @@ SliproadHandler::SliproadHandler(const IntersectionGenerator &intersection_gener
...
+                return boost::none;
+        }
+    } const selector;
+
+    (void)graph_walker.TraverseRoad(start, onto, accumulator, selector);
+
+    return accumulator.too_far_away;
+}
+
+bool SliproadHandler::isThroughStreet(const EdgeID from, const Intersection &intersection) const
```
MoKob: don't we have such a check somewhere else already? If not, I'd move it to the BaseClass

src/extractor/guidance/sliproad_handler.cpp
```diff
@@ -29,273 +32,720 @@ SliproadHandler::SliproadHandler(const IntersectionGenerator &intersection_gener
...
+                                                      name_table,
+                                                      street_name_suffix_table); //
+    };
+
+    return std::find_if(first, last, same_name) != last;
+}
+
+bool SliproadHandler::roadContinues(const EdgeID current, const EdgeID next) const
+{
+    const auto current_data = node_based_graph.GetEdgeData(current);
```
MoKob: references?

src/extractor/guidance/sliproad_handler.cpp
```diff
@@ -29,273 +32,720 @@ SliproadHandler::SliproadHandler(const IntersectionGenerator &intersection_gener
...
+                                                            next_data.name_id,
+                                                            name_table,
+                                                            street_name_suffix_table); //
+
+    const auto continues = same_road_category && same_travel_mode && same_name;
+    return continues;
+}
+
+bool SliproadHandler::isValidSliproadArea(const double max_area,
+                                          const NodeID a,
```
MoKob: please no single char variables :(

src/extractor/guidance/sliproad_handler.cpp
```diff
@@ -29,273 +32,720 @@ SliproadHandler::SliproadHandler(const IntersectionGenerator &intersection_gener
...
+    const constexpr auto MIN_SLIPROAD_AREA = 3.;
+
+    if (area < MIN_SLIPROAD_AREA || area > max_area)
+        return false;
+
+    return true;
+}
+
+bool SliproadHandler::isValidSliproadLink(const ConnectedRoad &sliproad,
+                                          const ConnectedRoad &first,
```
MoKob: first/second don't really provide an idea of what other roads you are passing in.

src/extractor/guidance/sliproad_handler.cpp
```diff
@@ -29,273 +32,720 @@ SliproadHandler::SliproadHandler(const IntersectionGenerator &intersection_gener
...
+    case RoadPriorityClass::MAIN_RESIDENTIAL:
+        factor = 0.4;
+        break;
+    case RoadPriorityClass::SIDE_RESIDENTIAL:
+        factor = 0.3;
+        break;
+    case RoadPriorityClass::LINK_ROAD:
+        factor = 0.3;
+        break;
+    case RoadPriorityClass::CONNECTIVITY:
```
MoKob: You explicitly forbid Connectivity Roads above. They are never going to be sliproads. You could probably assert this here?

src/extractor/guidance/sliproad_handler.cpp
```diff
@@ -29,273 +32,720 @@ SliproadHandler::SliproadHandler(const IntersectionGenerator &intersection_gener
...
+        factor = 0.1;
```
MoKob: are these in relation to the differences we usually assume in speed? So do all of them relate to `speed_class/speed_motorway`?

features/guidance/collapse.feature
MoKob: The scenario in question still provides a now different test. Can you create a copy of this test with similar setup that again identifies a sliproad here? The feature in collapse is supposed to collapse, which is not happening now.
The test as it is here can be moved to a different feature set.

src/extractor/guidance/sliproad_handler.cpp
MoKob: I've seen quite a selection of intersections that turn right before merging into a road. So I'd say we should try and detect these.
I'd say that the main road has to be narrow and the sliproad can't turn far from it.

src/extractor/guidance/sliproad_handler.cpp
MoKob: Does this really change so much? It's a bit confusing to be, but ok to keep. It's just that we are looking at a triangle area but filter by the square. So a longer sliproad allows stronger growing areas, which would result in scenarios that scale up linearly to work for one length and fail for the other.

e.g.
```
a - b - c
        d
```
could fail for a smaller grid size and work for a larger one.
Or is my math off here?

src/extractor/guidance/sliproad_handler.cpp
MoKob: Fork should always assign both roads

src/extractor/guidance/sliproad_handler.cpp
MoKob: hm. ok. might be a good item to add a refactor ticket for to consolidate these

daniel-j-h:
After rebasing onto your changes from today @MoKob the Sliproad handler still works on Intersection but the turn handler now only knows how to find the next IntersectionView along a road which again creates incompatibilities.

Going from an IntersectionView -> Intersection happens in turn analysis, but this is where we call the Sliproad handler to begin with. And the Sliproad handler has to work on Intersections since it potentially modifies the turn type.

:(

include/extractor/guidance/sliproad_handler.hpp
```diff
@@ -42,6 +47,40 @@ class SliproadHandler : public IntersectionHandler
...
+
+    boost::optional<std::size_t> getObviousIndexWithSliproads(const EdgeID from,
+                                                              const Intersection &intersection,
+                                                              const NodeID at) const;
+
+    // Next intersection from `start` onto `onto` is too far away for a Siproad scenario
+    bool nextIntersectionIsTooFarAway(const NodeID start, const EdgeID onto) const;
+
+    // Through street: does a road continue with from's name at the intersection
+    bool isThroughStreet(const EdgeID from, const IntersectionView &intersection) const;
```
MoKob: can you add a ticket to the guidance refactor project for removing this duplication with the intersection handler?

src/extractor/guidance/motorway_handler.cpp
```diff
@@ -490,7 +490,7 @@ Intersection MotorwayHandler::fromRamp(const EdgeID via_eid, Intersection inters
     else
     { // FALLBACK, hopefully this should never been reached
         util::Log(logDEBUG) << "Reached fallback on motorway ramp with " << intersection.size()
```
MoKob: heads up: these loggings have just been removed, since we don't need to check them anymore. Only extreme cases that we don't really care about. We don't need to clutter our logs for these

src/extractor/guidance/sliproad_handler.cpp
```diff
@@ -2,15 +2,18 @@
...
+#include <algorithm>
+#include <iterator>
 #include <limits>
 #include <utility>
 
 #include <boost/assert.hpp>
 
-using EdgeData = osrm::util::NodeBasedDynamicGraph::EdgeData;
-using osrm::util::guidance::getTurnDirection;
 using osrm::util::guidance::angularDeviation;
```
MoKob: also: moved (I believe to `#include "util/bearings.hpp"`)

src/extractor/guidance/sliproad_handler.cpp
```diff
@@ -29,273 +32,730 @@ SliproadHandler::SliproadHandler(const IntersectionGenerator &intersection_gener
...
-            const auto couldBeSliproad = [](const IntersectionView &intersection) {
-                if (intersection.size() != 3)
-                    return false;
-                if ((intersection[1].entry_allowed && intersection[2].entry_allowed) ||
-                    intersection[0].entry_allowed)
-                    return false;
-                return true;
-            };
+    // The intersection `c` has to be reachable
+    if (!main_road.entry_allowed)
```
MoKob: is this not already part of `obvious`? If not, shouldn't it be?

src/extractor/guidance/sliproad_handler.cpp
```diff
@@ -29,273 +32,730 @@ SliproadHandler::SliproadHandler(const IntersectionGenerator &intersection_gener
...
+        }
+
+        const auto road_data = node_based_graph.GetEdgeData(road.eid);
 
-    const auto source_edge_data = node_based_graph.GetEdgeData(source_edge_id);
-    // check whether the continue road is valid
-    const auto check_valid = [this, source_edge_data](const ConnectedRoad &road) {
-        const auto road_edge_data = node_based_graph.GetEdgeData(road.eid);
-        // Test to see if the source edge and the one we're looking at are the same road
+        auto is_roundabout = road_data.roundabout;
```
MoKob: doing roundabout checks early is usually a good idea. Make sure they are up as far as possible in the processing order, since they are relatively cheap compared to walking the graph.

src/extractor/guidance/sliproad_handler.cpp
```diff
@@ -29,273 +32,730 @@ SliproadHandler::SliproadHandler(const IntersectionGenerator &intersection_gener
...
+                return main_road_intersection->intersection.getRightmostRoad();
+
+            BOOST_ASSERT_MSG(false, "Sliproad is neither a left nor right of obvious main road");
+            return main_road_intersection->intersection.getLeftmostRoad();
+        }();
+
+        const auto &crossing_road_data = node_based_graph.GetEdgeData(crossing_road.eid);
+
+        // The crossing road at the main road's intersection must not be incoming-only
+        if (crossing_road_data.reversed)
```
MoKob: I am asking myself if we could get away with checking left/right of obvious and not iterating over all roads here.

src/extractor/guidance/sliproad_handler.cpp
```diff
@@ -29,273 +32,730 @@ SliproadHandler::SliproadHandler(const IntersectionGenerator &intersection_gener
...
+            {
+                node = node_based_graph.GetTarget(sliproad_edge);
+                sliproad_edge = intersection[1].eid;
+                intersection = intersection_generator(node, sliproad_edge);
+            }
+            return intersection;
+        }(intersection_node_id);
+
+        // Constrain the Sliproad's target to sliproad, outgoing, incoming from main intersection
+        if (target_intersection.size() != 3)
```
MoKob: looks like a prime candidate to check early (node_based_graph.getOutDegree(target_node) == 3)

src/extractor/guidance/sliproad_handler.cpp
```diff
@@ -29,273 +32,730 @@ SliproadHandler::SliproadHandler(const IntersectionGenerator &intersection_gener
...
+    {
+        geojson_points.Write(std::vector<NodeID>{b, c, d});
+        geojson_lines.Write(std::vector<NodeID>{b, d});
+    }
+
     return intersection;
 }
 
+// Implementation details
+
```
MoKob: the function calls above are only showing brief comments, could you maybe add some more specific documentation down here?

src/extractor/guidance/sliproad_handler.cpp
daniel-j-h: Can you please re-check the code it's using the graph walker now - which gives me back the intersection on which this check is getting performed on.

src/extractor/guidance/sliproad_handler.cpp
daniel-j-h: The interface (fun decls in the header) should have all documentation the callsite has to know about. I think.

daniel-j-h:
Rebased, squashed, adapted to review remarks.

Ticketed some possible performance improvements for all intersection handlers in https://github.com/Project-OSRM/osrm-backend/issues/3416.

Issue #3164, opened 2016-10-25
Enabling more compiler warnings by default


springmeyer:
We can help prevent common bugs by enabling more compiler warnings and adding `-Werror` to treat warnings as errors (at least on CI systems with consistent compilers).

I'd like to start by proposing adding `-Wsign-compare` and `-Wconversion` and fixing all the warnings they trigger as a first step.

/cc @chaupow @MoKob

springmeyer:
Here is the current output of warnings with `-Wsign-compare -Wconversion` when building with clang++-3.8: https://gist.github.com/springmeyer/1dd8dfa83a2dfce7d2e159502373db5f

springmeyer:
Many warnings for `-Wsign-compare -Wconversion` will come from third party headers (like boost). We want to ensure osrm-backend builds without warnings and suppress those from third party code. To do this I propose:

```c++

#include <osrm/warning_supress_start.hpp>
#include <third party include>
#include <osrm/warning_supress_end.hpp>
```

Where `osrm/warning_supress_start.hpp` looks like:

```c++
#pragma GCC diagnostic push
#pragma GCC diagnostic ignored "-Wunknown-pragmas" // clang+gcc
#pragma GCC diagnostic ignored "-Wpragmas" // gcc
#pragma GCC diagnostic ignored "-W#pragma-messages"
#pragma GCC diagnostic ignored "-Wsign-compare"
#pragma GCC diagnostic ignored "-Wconversion"
```

And `osrm/warning_supress_end.hpp` looks like:

```c++
#pragma GCC diagnostic pop
```

As we add more default warnings, we can then ensure they are suppressed by adding them to one place: `osrm/warning_supress_start.hpp`

chaupow:
I would also suggest adding `-Wshadow` as we have lots of `typedef` in the code that are potentially shadowing each other

springmeyer:
> Note: the alternative method of suppressing warnings from third party or system headers is to ensure that the include paths for them are referenced not with -I/path/to but with -isystem /path/to (https://gcc.gnu.org/onlinedocs/cpp/System-Headers.html). 

@chaupow and I took a look at the cmake build last week with an eye towards this issue. We found that all paths to third party dependencies are added with `-isystem`. So, this is actually working great for compilers that support `-isystem` (unix only) and there should be no need to use the `#include <osrm/warning_suppress_start.hpp>` method I propose above.

[...]

PR #3209, opened 2016-10-29
Enable warning flag -Wshorten-64-to-32

chaupow:
# Issue

This PR works towards #3164 and starts with solving all warnings that come with enabling `-Wshorten-64-to-32`

cc @springmeyer 
## Tasklist
- [ ] KILL ALL WARNINGS
- [ ] add `-Wshorten-64-to-32` and `-Werror`(?) to CMake File 
- [ ] rebase with master
- [ ] review
- [ ] adjust for comments

### DIFF
--- a/src/extractor/compressed_edge_container.cpp
+++ b/src/extractor/compressed_edge_container.cpp
@@ -4,6 +4,7 @@
 #include <boost/assert.hpp>
 #include <boost/filesystem.hpp>
 #include <boost/filesystem/fstream.hpp>
+#include <boost/numeric/conversion/cast.hpp>
 
 #include <limits>
 #include <string>
@@ -76,8 +77,8 @@ unsigned CompressedEdgeContainer::GetZippedPositionForReverseID(const EdgeID edg
 void CompressedEdgeContainer::SerializeInternalVector(const std::string &path) const
 {
     boost::filesystem::fstream geometry_out_stream(path, std::ios::binary | std::ios::out);
-    const unsigned compressed_geometry_indices = m_compressed_geometry_index.size() + 1;
-    const unsigned compressed_geometries = m_compressed_geometry_nodes.size();
+    const std::size_t compressed_geometry_indices = m_compressed_geometry_index.size() + 1;
+    const std::size_t compressed_geometries = m_compressed_geometry_nodes.size();
     BOOST_ASSERT(std::numeric_limits<unsigned>::max() != compressed_geometry_indices);
     geometry_out_stream.write((char *)&compressed_geometry_indices, sizeof(unsigned));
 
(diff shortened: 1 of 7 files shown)


src/extractor/compressed_edge_container.cpp
```diff
@@ -76,8 +77,8 @@ unsigned CompressedEdgeContainer::GetZippedPositionForReverseID(const EdgeID edg
 void CompressedEdgeContainer::SerializeInternalVector(const std::string &path) const
 {
     boost::filesystem::fstream geometry_out_stream(path, std::ios::binary | std::ios::out);
-    const unsigned compressed_geometry_indices = m_compressed_geometry_index.size() + 1;
-    const unsigned compressed_geometries = m_compressed_geometry_nodes.size();
+    const std::size_t compressed_geometry_indices = m_compressed_geometry_index.size() + 1;
+    const std::size_t compressed_geometries = m_compressed_geometry_nodes.size();
```
daniel-j-h: For all sizes written to disk it makes sense to use fixed-size integral types.

daniel-j-h:
`Werror` is problematic for multiple reasons, one is that we support a broad range of compilers of which some are strict and some are stricter. For example, gcc 4.9 issues warnings you won't see on Travis with more recent compilers (since they're wrong). Making these builds fail hard is not an option.

springmeyer:
@daniel-j-h I agree with you that `-Werror` would be problematic now and is not realistic short-term, but I think it should be a goal. Both `mbgl` and `v8` use `-Werror` by default because they are able to build with minimal warnings across a large range of compilers.

springmeyer:
interesting read in this context of this work: http://blog.reverberate.org/2012/12/testing-for-integer-overflow-in-c-and-c.html

daniel-j-h:
What's the state of this? Asking since @MoKob did a PR today fixing the `int` vs `size_t` differences, too. Want to rebase against master, squash changes and then already merge this into master?

springmeyer:
See also https://github.com/Project-OSRM/osrm-backend/issues/3483

daniel-j-h:
Ping - what's the state of this? We're having this PR around now for three month; it's not up to date or rebased on master and the contractor size\_t issues already fixed by @MoKob. If it's not a priority at the moment please at least rebase and let's get the changes in - otherwise I'm inclined to close this.

daniel-j-h:
Same with https://github.com/Project-OSRM/osrm-backend/issues/3247 - if it's a part of this ticket and / or stale please update accordingly.

chaupow:
@springmeyer I don't have to time to work on this atm 😞 
Should we just close this for now?

springmeyer:
Let's close. I propose the next action be that we re-issue a new PR that actually enables this warning. Then we can fix the warnings in followup PRs (and look back at this one for hints).

springmeyer:
See also #4495

PR #3210, opened 2016-10-31
Refactor/graph traverser - pending #3170

MoKob:
# Issue

A first step in direction of https://github.com/Project-OSRM/osrm-backend/issues/3107, enabling a clearer traversal of the node-based-graph.

The graph-hopper abstracts away the logic needed to traverse the node-based-graph for different settings, finding next/previous intersections.

## Tasklist
 - [x] merge https://github.com/Project-OSRM/osrm-backend/pull/3170
 - [x] review
 - [x] adjust for comments

## Requirements / Relations
this PR is based on https://github.com/Project-OSRM/osrm-backend/pull/3170

### DIFF
--- a/include/extractor/guidance/intersection.hpp
+++ b/include/extractor/guidance/intersection.hpp
@@ -5,6 +5,7 @@
 #include <vector>
 
 #include "extractor/guidance/turn_instruction.hpp"
+#include "util/guidance/toolkit.hpp"
 #include "util/typedefs.hpp" // EdgeID
 
 namespace osrm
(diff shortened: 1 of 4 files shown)


include/extractor/guidance/graph_hopper.hpp
```diff
@@ -0,0 +1,143 @@
...
+{
+
+// forward declaration to allow interaction between the intersection generator and the graph hopper
+class IntersectionGenerator;
+
+/*
+ * The graph hopper is a utility that lets you find certain intersections with a node based graph,
+ * accumulating information along the way
+ */
+class GraphHopper
```
daniel-j-h: Not sure you really want to call it that - there is a routing engine out there with that name, might be confusing for people stumbling over it.

include/extractor/guidance/graph_hopper.hpp
```diff
@@ -0,0 +1,143 @@
...
+                                                                     const selector_type &selector)
+{
+    // since graph hopping is used in many ways, we don't generate an adjusted intersection
+    // (otherwise we could end up in infinite recursion if we call the graph hopper during the
+    // adjustment itself). Relying only on `GetConnectedRoads` (which itself does no graph hopping),
+    // we prevent this from happening.
+    const auto stop_node_id = current_node_id;
+    // we wan't to put out the last valid entries. To do so, we need to update within the following
+    // while loop.
+    for (std::size_t safety_hop_limit = 0; safety_hop_limit < 1000; ++safety_hop_limit)
```
daniel-j-h: Would you be so kind to explain the limit to me? Why do we need one, for cycles? In other locations we're explicitly storing nodes for cycle detection.

include/extractor/guidance/intersection.hpp
```diff
@@ -59,7 +60,73 @@ struct ConnectedRoad final
 // small helper function to print the content of a connected road
 std::string toString(const ConnectedRoad &road);
 
-typedef std::vector<ConnectedRoad> Intersection;
+struct Intersection
```
daniel-j-h: Nice - thank you for starting a refactor here! You could probably make your life easier by doing

```cpp
struct Intersection final : vector<Road> {
  using Base = vector<Road>;

  using Base::size;
  ..
};
```

include/extractor/guidance/graph_hopper.hpp
MoKob: if you look upwards, we are also doing this here (stop_node_id). I just wanted to ensure we don't get caught in endless loops again (due to some random graph issues) and prevent any while loops in the code. If the graph traversal should ever be looking at more than 10 intersections, we are probably doing something wrong anyhow. Having this limit in here simply assures that we will get notice but the graph will not simply hang forever.

daniel-j-h:
Looks good.

MoKob:
Going to add some additional documentation and merge afterwards.

PR #3211, opened 2016-10-31
Refactor/merging segregated roads

MoKob:
# Issue
Resolves https://github.com/Project-OSRM/osrm-backend/issues/3107

The PR makes the merging of segregated roads much clearer to follow and less error prone. There are still quite a few situations that we could detect better, some due to data issues (e.g. no lanes provided). Nevertheless I'd prefer to close this one off and continue tuning work as a separate part.

## Tasklist
 - [x] requires https://github.com/Project-OSRM/osrm-backend/pull/3170 merged
 - [x] requires https://github.com/Project-OSRM/osrm-backend/pull/3210 merged
 - [x] requires https://github.com/Project-OSRM/osrm-backend/pull/3272 merged
 - [x] requires https://github.com/Project-OSRM/osrm-backend/pull/3277 merged
 - [x] needs detector to exclude turn-restriction based merges (if it's not a oneway street, don't merge)
 - [x] refactor to use less circular dependencies
 - [x] fix link road detection
 - [x] fix performance problems
 - [x] investigate failing tests
 - [x] remove geojson output
 - [x] review
 - [x] adjust for comments

## Requirements / Relations
Based on the above mentioned PRs

## Review Walkthrough:

- Cucumber Tests: Need to include `lanes` on many occasions to correctly compute the width between parallel roads. This becomes necessary, since by now the detection only allows a maximal width to be located between two roads to be considered as one. If it is a larger width in between, we don't consider the roads as one.

- include/engine/guidance/assemble_steps.hpp: naming/formatting changes only

- include/extractor/geojson_debug_policies.hpp: new base class for intersection printer to allow printing of shapes (debug tool only)

- include/extractor/guidance/intersection.hpp some additional functionality for shapes. `EnableShapeOps` is new and should be checked. Rest is formatting/naming

- include/extractor/guidance/intersection_generator.hpp: simply switched `std::pair` for a named struct
	- influences: include/extractor/guidance/intersection_handler.hpp (only change)

	
- include/extractor/guidance/intersection_normalizer.hpp/cpp: main point of entry for merging. It does not detect merges anymore but hands off to the **new** `mergable_road_detector`

- include/extractor/guidance/mergable_road_detector.hpp: complete rewrite of merging discovery and main point of this PR. Uses a series of positive/negative detectors. Can possibly be optimised for performance, but seems fine right now. I'll ticket the further optimisations and the fine tuning.

- include/extractor/guidance/node_based_graph_walker.hpp: new graph walker to be used in detecting mergable roads

- include/util/bearing.hpp reduced naming duplications: `util::bearing::reverseBearing` -> `util::bearing::reverse` and moved two functions in here (angleBetween and restrictAngleToValidRange)

- include/util/coordinate_calculation.hpp/.cpp: tools to detect mergable roads. main focus: detection of parallel roads. Moved leastSquareRegression here (and rewrote it, due to some errors). Main point: rotate for near vertical point sets


MoKob:
Currently seeing some slowdown (10%). Need to see what I can do to reduce that as much as possible

danpat:
@MoKob Would you mind writing up a short summary of what's in this change?  A brief "here's what you're looking at" would really help speed up comprehension when reviewing, I'm struggling a bit to sort out what's basic refactoring, and what's a logic change that needs extra attention.

MoKob:
@danpat well. It's mostly a complete rewrite. But I will see that I put a small review guide together.

MoKob:
Current timings on Germany:

Current Master:
Generating edges: 273.463s

Refactored:
Generating edges: 332.128s

Slowdown factor of 1.2. Seems like I have to do a lot of digging first to keep this at bay.

MoKob:
After improving the distinction detection (not using merge as often), the performance problems seem to be non-existent anymore.

src/extractor/extractor_callbacks.cpp
```diff
@@ -139,8 +139,6 @@ void ExtractorCallbacks::ProcessWay(const osmium::Way &input_way, const Extracti
         return;
     }
 
-    // FIXME this need to be moved into the profiles
```
MoKob: this `FIXME` was left over, point it was referring to was implemented.

MoKob:
@danpat I've added a small overview to the opening part of this PR. Let me know if you need further guidance here

MoKob:
Update on the timing: 284.715s for Germany, which is an acceptable slowdown in my view for this initial part. Further improvements can be done in the already ticketed https://github.com/Project-OSRM/osrm-backend/issues/3473

include/extractor/guidance/mergable_road_detector.hpp
```diff
@@ -0,0 +1,149 @@
...
+     * we don't wan't to have this information within our list of intersections/possible turn
+     * locations.
+     *
+     *     ___________
+     * ---<___________>-----
+     *
+     *
+     * Would feel just like a single straight road to a driver and should be represented as such in
+     * our engine.*/
+    bool ConnectAgain(const NodeID intersection_node,
```
daniel-j-h: Hm want to rename this to `IsTrafficIsland` then?

include/extractor/guidance/mergable_road_detector.hpp
```diff
@@ -0,0 +1,149 @@
...
+                      const MergableRoadData &rhs) const;
+
+    /* A negative detector, preventing a merge, trying to detect link roads between two main roads.
+     *
+     *  d - - - - - - - - e - f
+     *             . / '
+     *  a - - - b - - - - - - c
+     *
+     * The detector wants to prevent merges that are connected to `b-e` */
+    bool IsLinkRoad(const NodeID intersection_node, const MergableRoadData &road) const;
```
daniel-j-h: Hm same here, this could maybe be of use for the sliproad handler (?) - not sure if actionable, you decide

include/extractor/guidance/mergable_road_detector.hpp
```diff
@@ -0,0 +1,149 @@
...
+
+    const util::NodeBasedDynamicGraph &node_based_graph;
+    const std::vector<QueryNode> &node_coordinates;
+    const IntersectionGenerator &intersection_generator;
+    const CoordinateExtractor &coordinate_extractor;
+
+    // name detection
+    const util::NameTable &name_table;
+    const SuffixTable &street_name_suffix_table;
```
daniel-j-h: For all of these, if we only ever use them by ref in this header: you could forward declare them and only in the cpp impl. file include the headers. Not sure if it's worth it (especially with lto). Also maybe we should think of a `_fwd.hpp` header file all handler header could include, since most of the types used should be the same - what do you think?

include/extractor/guidance/node_based_graph_walker.hpp
```diff
@@ -105,10 +105,35 @@ struct SelectRoadByNameOnlyChoiceAndStraightness
      */
     boost::optional<EdgeID> operator()(const NodeID nid,
                                        const EdgeID via_edge_id,
-                                       const Intersection &intersection,
+                                       const IntersectionView &intersection,
+                                       const util::NodeBasedDynamicGraph &node_based_graph) const;
+
+    const NameID desired_name_id;
+    const bool requires_entry;
```
daniel-j-h: Hrm here and below: the selectors / accumulators are structs with all member attributes public. But from what I can tell here, these are completely useless as the callsite sets them in the ctor.

What about at least making internals private and public what the user should be able to access?

include/extractor/guidance/road_classification.hpp
```diff
@@ -1,6 +1,7 @@
 #ifndef OSRM_EXTRACTOR_CLASSIFICATION_DATA_HPP_
 #define OSRM_EXTRACTOR_CLASSIFICATION_DATA_HPP_
 
+#include <cstdlib>
```
daniel-j-h: Hm cstdlib is already included below, messed up rebase?

include/util/bearing.hpp
```diff
@@ -4,6 +4,7 @@
 #include <algorithm>
 #include <boost/assert.hpp>
 #include <cmath>
+#include <cmath>
```
daniel-j-h: Same here

include/util/coordinate_calculation.hpp
```diff
@@ -95,6 +95,23 @@ inline std::pair<double, FloatCoordinate> projectPointOnSegment(const FloatCoord
...
+// find the closest distance between a coordinate and a set of coordinates
+// O(|coordinates|)
+double findClosestDistance(const Coordinate coordinate, const std::vector<Coordinate> &coordinates);
+
+// find the closes distance between two sets of coordinates
+// O(|lhs| * |rhs|)
+double findClosestDistance(const std::vector<Coordinate> &lhs, const std::vector<Coordinate> &rhs);
+
+// checks if two sets of coordinates describe a parallel set of ways
+bool areParallel(const std::vector<Coordinate> &lhs, const std::vector<Coordinate> &rhs);
```
daniel-j-h: Hm these interfaces are maybe a bit to strict - do we want to hardcode vector<Coord> here, or could we use Iterators? (I just want to avoid situations we already have in the code base where we create vectors, call a function, and then throw them away only)

src/extractor/geojson_debug_policies.cpp
```diff
@@ -18,26 +18,28 @@ IntersectionPrinter::IntersectionPrinter(
...
     coordinates.push_back(node_coordinates[intersection_node]);
 
-    const auto road_to_coordinate = [&](const extractor::guidance::ConnectedRoad &connected_road) {
-        const constexpr auto FORWARD = false;
-        const auto to_node = node_based_graph.GetTarget(connected_road.eid);
-        return coordinate_extractor.GetCoordinateAlongRoad(
-            intersection_node, connected_road.eid, FORWARD, to_node, intersection_lanes);
-    };
+    const auto road_to_coordinate =
+        [&](const extractor::guidance::IntersectionShapeData &connected_road) {
```
daniel-j-h: const auto& road

src/extractor/geojson_debug_policies.cpp
```diff
@@ -65,5 +67,28 @@ operator()(const NodeID intersection_node,
...
+           const extractor::guidance::Intersection &intersection,
+           const boost::optional<util::json::Object> &node_style,
+           const boost::optional<util::json::Object> &way_style) const
+{
+    guidance::IntersectionShape intersection_shape;
+    intersection_shape.resize(intersection.size());
+    std::transform(intersection.begin(),
+                   intersection.end(),
+                   intersection_shape.begin(),
+                   [](const auto &road) { return guidance::IntersectionShapeData(road); });
```
daniel-j-h: Hmmm let's see if I get this right:
- intersection if of type Intersection, which is a vector of ConnectedRoads
- ConnectedRoad is a IntersectionViewData
- IntersectionViewData is a IntersectionShapeData

you upcast through three layers here slicing away the ConnectedRoad's structure leaving you with a IntersectionShapeData (?)

Hmm strongly couples the implementation here with the intersection representation a bit ..

src/extractor/guidance/intersection_handler.cpp
```diff
@@ -382,19 +386,17 @@ IntersectionHandler::getNextIntersection(const NodeID at, const EdgeID via) cons
...
-    EdgeID edge = SPECIAL_EDGEID;
-
-    std::tie(node, edge) = intersection_generator.SkipDegreeTwoNodes(at, via);
-    auto intersection = intersection_generator(node, edge);
-
+    const auto intersection_parameters = intersection_generator.SkipDegreeTwoNodes(at, via);
     // This should never happen, guard against nevertheless
-    if (node == SPECIAL_NODEID || edge == SPECIAL_EDGEID)
+    if (intersection_parameters.nid == SPECIAL_NODEID ||
+        intersection_parameters.via_eid == SPECIAL_EDGEID)
```
daniel-j-h: I'm wondering when this could ever happen ..

src/extractor/guidance/intersection_normalizer.cpp
```diff
@@ -21,178 +20,125 @@ IntersectionNormalizer::IntersectionNormalizer(
...
+        {
+            if (merged_shape_and_merges.second.end() !=
+                std::find_if(merged_shape_and_merges.second.begin(),
+                             merged_shape_and_merges.second.end(),
+                             [road](const auto &val) {
+                                 return road.eid == val.first || road.eid == val.second;
+                             }))
+                reduced_shape.push_back(road);
+        }
+    }
```
daniel-j-h: Hmm the copy because of the move is unfortunate here ..

src/extractor/guidance/intersection_normalizer.cpp
```diff
@@ -21,178 +20,125 @@ IntersectionNormalizer::IntersectionNormalizer(
...
 {
 }
 
 std::pair<IntersectionShape, std::vector<std::pair<EdgeID, EdgeID>>> IntersectionNormalizer::
 operator()(const NodeID node_at_intersection, IntersectionShape intersection) const
 {
+    const auto intersection_copy = intersection;
     auto merged_shape_and_merges =
         MergeSegregatedRoads(node_at_intersection, std::move(intersection));
+    if (!merged_shape_and_merges.second.empty())
```
daniel-j-h: This is what I mean with the pairs above: what is .second here? What is .first below?

src/extractor/guidance/intersection_normalizer.cpp
```diff
@@ -21,178 +20,125 @@ IntersectionNormalizer::IntersectionNormalizer(
...
-    else
-        return false;
+    const auto is_distinct = [&]() {
+        return mergable_road_detector.IsDistinctFrom(
+                   intersection[second_index_in_ccw],
+                   intersection[(second_index_in_ccw + 1) % intersection.size()]) &&
+               mergable_road_detector.IsDistinctFrom(
+                   intersection[(fist_index_in_ccw + intersection.size() - 1) %
+                                intersection.size()],
+                   intersection[fist_index_in_ccw]);
```
daniel-j-h: Want to simplify that expression a bit?

src/extractor/guidance/mergable_road_detector.cpp
```diff
@@ -0,0 +1,427 @@
...
+{
+namespace extractor
+{
+namespace guidance
+{
+
+namespace
+{
+const auto ASSUMED_LANE_WIDTH = 3.25;
+const auto MERGABLE_ANGLE_DIFFERENCE = 95.0;
```
daniel-j-h: constexpr

daniel-j-h: Also should this be in constants.hpp?

src/extractor/guidance/mergable_road_detector.cpp
```diff
@@ -0,0 +1,427 @@
...
+
+// check a connected road for equality of a name
+inline auto makeCheckRoadForName(const NameID name_id,
+                                 const util::NodeBasedDynamicGraph &node_based_graph)
+{
+    return [name_id, &node_based_graph](const MergableRoadDetector::MergableRoadData &road) {
+        // since we filter here, we don't want any other name than the one we are looking for
+        return name_id != node_based_graph.GetEdgeData(road.eid).name_id;
```
daniel-j-h: Hm so this is doing the name_id check and not the name change heuristic check - why?

src/extractor/guidance/mergable_road_detector.cpp
```diff
@@ -0,0 +1,427 @@
...
+    // Don't merge link roads
+    if (IsLinkRoad(intersection_node, lhs) || IsLinkRoad(intersection_node, rhs))
+        return false;
+
+    // check if we simply split up prior to an intersection
+    if (IsNarrowTriangle(intersection_node, lhs, rhs))
+        return true;
+
+    // finally check if two roads describe the direction
+    return HaveSameDirection(intersection_node, lhs, rhs);
```
daniel-j-h: I assume you ordered the checks by expensiveness?

src/extractor/guidance/mergable_road_detector.cpp
```diff
@@ -0,0 +1,427 @@
...
+    return intersection_node == node_based_graph.GetTarget(connection.via_eid);
+}
+
+bool MergableRoadDetector::IsNarrowTriangle(const NodeID intersection_node,
+                                            const MergableRoadData &lhs,
+                                            const MergableRoadData &rhs) const
+{
+    // selection data to the right and left
+    IntersectionFinderAccumulator left_accumulator(5, intersection_generator),
+        right_accumulator(5, intersection_generator);
```
daniel-j-h: Why 5?

src/extractor/guidance/mergable_road_detector.cpp
```diff
@@ -0,0 +1,427 @@
...
+
+    /* Standard following the straightmost road
+     * Since both items have the same id, we can `select` based on any setup
+     */
+    SelectStraightmostRoadByNameAndOnlyChoice selector(
+        node_based_graph.GetEdgeData(lhs.eid).name_id, lhs.bearing, false);
```
daniel-j-h: /\*requires_entry=\*/false

src/extractor/guidance/mergable_road_detector.cpp
```diff
@@ -0,0 +1,427 @@
...
+        return std::max<std::uint8_t>(
+            node_based_graph.GetEdgeData(road.eid).road_classification.GetNumberOfLanes(), 1);
+    };
+
+    // the width we can bridge at the intersection
+    const auto assumed_road_width = (num_lanes(lhs) + num_lanes(rhs)) * ASSUMED_LANE_WIDTH;
+    const auto distance_between_triangle_corners = util::coordinate_calculation::haversineDistance(
+        node_coordinates[node_based_graph.GetTarget(left_accumulator.via_edge_id)],
+        node_coordinates[node_based_graph.GetTarget(right_accumulator.via_edge_id)]);
+    if (distance_between_triangle_corners > (assumed_road_width + 10))
```
daniel-j-h: Why 10?

daniel-j-h: Also feel free to create more variables, I feel like a lot of your function calls are endlessly long since you subscript, call functions etc.

src/extractor/guidance/mergable_road_detector.cpp
```diff
@@ -0,0 +1,427 @@
...
+    if (angularDeviation(lhs.bearing, rhs.bearing) > 95)
+        return false;
+
+    // Find a coordinate following a road that is far away
+    NodeBasedGraphWalker graph_walker(node_based_graph, intersection_generator);
+    const auto getCoordinatesAlongWay = [&](const EdgeID edge_id, const double max_length) {
+        LengthLimitedCoordinateAccumulator accumulator(
+            coordinate_extractor, node_based_graph, max_length);
+        SelectStraightmostRoadByNameAndOnlyChoice selector(
+            node_based_graph.GetEdgeData(edge_id).name_id, lhs.bearing, false);
```
daniel-j-h: Hm here and bove I had to check the selector what the boolean false is all about .. what do you think about making as much enums out of bools as possible or at least on the callsite doing something like `/*pred*/=false`?

src/extractor/guidance/mergable_road_detector.cpp
```diff
@@ -0,0 +1,427 @@
...
+        graph_walker.TraverseRoad(intersection_node, edge_id, accumulator, selector);
+
+        return std::make_pair(accumulator.accumulated_length, accumulator.coordinates);
+    };
+
+    std::vector<util::Coordinate> coordinates_to_the_left, coordinates_to_the_right;
+    double distance_traversed_to_the_left, distance_traversed_to_the_right;
+
+    const double constexpr distance_to_extract = 100;
```
daniel-j-h: Why 100? Why 40 below? Why 5 below? Want to symbolize at least?

src/extractor/guidance/mergable_road_detector.cpp
```diff
@@ -0,0 +1,427 @@
...
+
+    coordinates_to_the_right = coordinate_extractor.SampleCoordinates(
+        std::move(coordinates_to_the_right), distance_to_extract, 5);
+
+    /* extract the number of lanes for a road
+     * restricts a vector to the last two thirds of the length
+     */
+    const auto prune = [](auto &data_vector) {
+        BOOST_ASSERT(data_vector.size() >= 3);
+        data_vector.erase(data_vector.begin(), data_vector.begin() + data_vector.size() / 3);
```
daniel-j-h: Maybe simplify via .resize(size / 3)

src/extractor/guidance/mergable_road_detector.cpp
```diff
@@ -0,0 +1,427 @@
...
+    const auto distance_between_roads = util::coordinate_calculation::findClosestDistance(
+        coordinates_to_the_left[coordinates_to_the_left.size() / 2], coordinates_to_the_right);
+
+    const auto combined_road_width =
+        0.5 *
+        (std::max<int>(
+             1, node_based_graph.GetEdgeData(lhs.eid).road_classification.GetNumberOfLanes()) +
+         std::max<int>(
+             1, node_based_graph.GetEdgeData(rhs.eid).road_classification.GetNumberOfLanes())) *
+        ASSUMED_LANE_WIDTH;
```
daniel-j-h: Can you simplify expressions like this? It's starting to look like Lisp with our narrow column limit :)

src/extractor/guidance/mergable_road_detector.cpp
```diff
@@ -0,0 +1,427 @@
...
+        const auto required_name_id = node_based_graph.GetEdgeData(range.front()).name_id;
+
+        const auto has_required_name = [this, required_name_id](const auto edge_id) {
+            return node_based_graph.GetEdgeData(edge_id).name_id == required_name_id;
+        };
+
+        /* the beautiful way would be:
+         * return range.end() == std::find_if_not(range.begin(), range.end(), has_required_name);
+         * but that does not work due to range concepts
+         */
```
daniel-j-h: Want to TODO / FIXME this?

daniel-j-h: Wait, the iteration version should work, no? I thought only the boost::range version does not because of our range's limitations?

src/extractor/guidance/mergable_road_detector.cpp
```diff
@@ -0,0 +1,427 @@
...
+        return false;
+
+    const auto distance_between_candidates = util::coordinate_calculation::haversineDistance(
+        node_coordinates[intersection_node], node_coordinates[left_candidate]);
+
+    const auto both_split_join = degree_three_connect_in && degree_three_connect_out;
+
+    // allow longer separations if both are joining directly
+    return both_split_join ? (distance_between_candidates < 30)
+                           : (distance_between_candidates < 15);
```
daniel-j-h: Why 30, why 15?

src/extractor/guidance/mergable_road_detector.cpp
```diff
@@ -0,0 +1,427 @@
...
+}
+
+bool MergableRoadDetector::IsLinkRoad(const NodeID intersection_node,
+                                      const MergableRoadData &road) const
+{
+    const auto next_intersection_parameters =
+        intersection_generator.SkipDegreeTwoNodes(intersection_node, road.eid);
+    const auto next_intersection_along_road = intersection_generator.GetConnectedRoads(
+        next_intersection_parameters.nid, next_intersection_parameters.via_eid);
+    const auto extract_name = [this](const MergableRoadData &road) {
```
daniel-j-h: Want to call this extract_name_id and adapt vars below?

src/extractor/guidance/mergable_road_detector.cpp
```diff
@@ -0,0 +1,427 @@
...
+     * incoming road.
+     */
+    if (angularDeviation(angularDeviation(next_road_along_path->angle, 180),
+                         angularDeviation(opposite_of_next_road_along_path->angle, 0)) <
+        FUZZY_ANGLE_DIFFERENCE)
+        return false;
+
+    // near straight road that continues
+    return angularDeviation(opposite_of_next_road_along_path->angle, next_road_along_path->angle) >=
+               160 &&
```
daniel-j-h: Don't we have almost-straight'ish constants for these or why the 160 here?

src/extractor/guidance/node_based_graph_walker.cpp
```diff
@@ -48,6 +51,7 @@ void LengthLimitedCoordinateAccumulator::update(const NodeID from_node,
     coordinates.insert(coordinates.end(), current_coordinates.begin(), current_coordinates.end());
 
     accumulated_length += length;
+    accumulated_length = std::min(accumulated_length, max_length);
```
daniel-j-h: Hm not sure this should be done in the walker - the user knows the max length, what's probably more interesting is the actual accumulated distance, no? Otherwise we make decisions for the call site so to speak.

src/extractor/guidance/turn_lane_handler.cpp
```diff
@@ -6,6 +6,7 @@
 #include "util/bearing.hpp"
 #include "util/log.hpp"
 #include "util/typedefs.hpp"
+#include "util/bearing.hpp"
```
daniel-j-h: Remove - look three lines up.

src/util/coordinate_calculation.cpp
```diff
@@ -315,49 +318,237 @@ bool isCCW(const Coordinate first_coordinate,
...
+                 coordinates.begin(), coordinates.end(), 0., make_variance(mean_lon, extract_lon)) /
+             (coordinates.size() - 1));
+
+    // if we don't change longitude, return the vertical line as is
+    if (std::abs(sample_variance_lon) <
+        std::numeric_limits<decltype(sample_variance_lon)>::epsilon())
+        return {coordinates.front(), coordinates.back()};
+
+    const auto sample_variance_lat =
+        sqrt(std::accumulate(
```
daniel-j-h: std::

src/util/coordinate_calculation.cpp
```diff
@@ -315,49 +318,237 @@ bool isCCW(const Coordinate first_coordinate,
...
+        return [extraction_function, mean](const double sum_so_far, const Coordinate coordinate) {
+            const auto difference = extraction_function(coordinate) - mean;
+            return sum_so_far + difference * difference;
+        };
+    };
+
+    // using the unbiased version, we divide by num_samples - 1 (see
+    // http://mathworld.wolfram.com/SampleVariance.html)
+    const auto sample_variance_lon =
+        sqrt(std::accumulate(
```
daniel-j-h: std::

src/util/coordinate_calculation.cpp
```diff
@@ -315,49 +318,237 @@ bool isCCW(const Coordinate first_coordinate,
...
+            util::FloatLatitude{sin_alpha * lon + cos_alpha * lat}};
+}
+
+Coordinate difference(const Coordinate lhs, const Coordinate rhs)
+{
+    const auto lon_diff_int = static_cast<int>(lhs.lon) - static_cast<int>(rhs.lon);
+    const auto lat_diff_int = static_cast<int>(lhs.lat) - static_cast<int>(rhs.lat);
+    return {util::FixedLongitude{lon_diff_int}, util::FixedLatitude{lat_diff_int}};
+}
+
```
daniel-j-h: What do you think of having some unit tests for all of these?

features/guidance/collapse.feature
```diff
@@ -143,37 +143,33 @@ Feature: Collapse
...
+               g h
+            c - b - a
+            d - e - f
+               j i
+               | |
+               | |
+               | |
+               | |
+               | |
+               k l
```
MoKob: This all depends on which detector is used. Before I only checked the angles between the splitting roads. It turns out that there are a lot of situations, though, with a lot of space between the two segments. If there is a big park in between, I don't think we should completely omit the fact that the road separates into a side left/right of the park. 

To pass the width test between the roads and to make the test check the original behaviour, it was needed to adjust the width between roads here and in some other cases.
The number of lanes is used to calculate the road width. Thats why I had to add it in some places. (think half of the number of lanes times 3 -> 6 roads = 9 m left/right of the road).

It is, of course, not a well distributed feature. Using the lane width shifts the problem of merging/not merging into the data-space rather than the OSRM space. We cannot detect everything validly, so I feel using the data as a method to tune the detection is the best way to go.

features/guidance/collapse.feature
```diff
@@ -143,37 +143,33 @@ Feature: Collapse
...
-            | ngbhm | primary | second | yes    |
-            | liejk | primary | second | yes    |
+            | nodes | highway | name   | oneway | lanes |
+            | ab    | primary | first  | yes    |       |
+            | bc    | primary | first  | yes    |       |
+            | de    | primary | first  | yes    |       |
+            | ef    | primary | first  | yes    |       |
+            | be    | primary | first  | no     |       |
+            | ngbhm | primary | second | yes    | 5     |
+            | liejk | primary | second | yes    | 5     |
```
MoKob: see above. 2.5 * 2 *3 = 15 meters so we got 5 meters to spare between the roads.

features/guidance/collapse.feature
```diff
@@ -353,13 +353,13 @@ Feature: Collapse
...
-            | d,g       | first,second,second | depart,turn right,arrive     |
-            | g,f       | second,first,first  | depart,turn right,arrive     |
-            | g,c       | second,first,first  | depart,turn left,arrive      |
+            | waypoints | route               | turns                          |
+            | d,c       | first,first,first   | depart,continue uturn,arrive   |
+            | a,f       | first,first,first   | depart,continue uturn,arrive   |
+            | a,g       | first,second,second | depart,turn left,arrive        |
+            | d,g       | first,second,second | depart,turn right,arrive       |
+            | g,f       | second,first,first  | depart,turn right,arrive       |
+            | g,c       | second,first,first  | depart,end of road left,arrive |
```
MoKob: the turn is at `b`. This currently oscillates quite heavily due to https://github.com/Project-OSRM/osrm-backend/issues/3365. It all depends on whether the turn at the end is recognised as obvious/not. Since I have fixed some minor errors in angle detection, this now again is recognised as a turn at the end of the road (at `b`)

features/guidance/collapse.feature
```diff
@@ -425,18 +425,28 @@ Feature: Collapse
...
-
         And the ways
-            | nodes | highway | name    | oneway |
-            | kabhj | primary | inroad  | yes    |
-            | bc    | primary | inroad  | no     |
-            | cd    | primary | bridge  | no     |
-            | defg  | primary | outroad | no     |
-            | fi    | primary | cross   | no     |
+            | nodes | highway | name    | oneway | lanes |
+            | kabhj | primary | inroad  | yes    | 4     |
```
MoKob: If the lane tag is not present, the road will not be recognised as being the same. We will not merge the two into one. This changes turn angles / obvious detection. It can end up as a `slight turn` instead of straight, for example.

features/guidance/turn-lanes.feature
```diff
@@ -599,22 +599,21 @@ Feature: Turn Lane Guidance
...
 
         And the ways
-            | nodes | name     | turn:lanes:forward                 | oneway | highway   |
-            | abc   | road     | left\|left\|left\|through\|through | yes    | primary   |
-            | cde   | road     |                                    | yes    | primary   |
-            | hc    | cross    |                                    | yes    | secondary |
-            | cg    | straight |                                    | no     | tertiary  |
-            | cf    | left     |                                    | yes    | primary   |
+            | nodes | name     | turn:lanes:forward                 | oneway | highway   | lanes |
+            | abc   | road     | left\|left\|left\|through\|through | yes    | primary   | 5     |
```
MoKob: So far we do not do that. I've tested it on Berlin, and Berlin was adding 3 tags. It seems that when lanes are tagged for indications, they often come with number of lanes as well. If it turns out to be way different on the planet, we could think about it.

features/guidance/turn-angles.feature
```diff
@@ -768,7 +768,31 @@ Feature: Simple Turns
...
+
+               .             .
+
+              .             .
+
+             .             .
+
+            .             .
+
+            g             h
```
MoKob: Also necessary to adjust to the new detectors. Since parallel roads require a certain length (here the grid size is really small), I had to adjust the length of the roads to still have the same test (here with a merged road going to the bottom).

include/extractor/guidance/intersection.hpp
```diff
@@ -108,11 +115,60 @@ struct ConnectedRoad final : IntersectionViewData
 };
 
 // small helper function to print the content of a connected road
+std::string toString(const IntersectionShapeData &shape);
+std::string toString(const IntersectionViewData &view);
```
MoKob: here I print all data that is part of the item, they cannot be templated

include/extractor/geojson_debug_policies.hpp
```diff
@@ -37,6 +37,21 @@ struct IntersectionPrinter
...
+                        const std::vector<extractor::QueryNode> &node_coordinates,
+                        const extractor::guidance::CoordinateExtractor &coordinate_extractor);
+
+    // renders the used coordinate locations for all entries/as well as the resulting
+    // intersection-classification
+    util::json::Array operator()(const NodeID intersection_node,
+                                 const extractor::guidance::Intersection &intersection,
+                                 const boost::optional<util::json::Object> &node_style = {},
+                                 const boost::optional<util::json::Object> &way_style = {}) const;
+};
```
MoKob: I tried to not make it a template to reduce the number of items that we define in headers. Right now the derived class transforms what it is given into an intersectionShape and calls its base class operator. There is no code duplication for rendering, but I can transform it into a template just as well.

Whats your opinion here? I don't feel strongly one way or the other.

include/extractor/guidance/intersection.hpp
```diff
@@ -108,11 +115,60 @@ struct ConnectedRoad final : IntersectionViewData
...
+{
+    // same as closest turn, but for bearings
+    auto findClosestBearing(double bearing) const
+    {
+        auto comp = makeCompareShapeDataByBearing(bearing);
+        return std::min_element(self()->begin(), self()->end(), comp);
+    }
+
+    // search a given eid in the intersection
+    auto findEid(const EdgeID eid) const
```
MoKob: Ticketed at https://github.com/Project-OSRM/osrm-backend/issues/3491

include/extractor/guidance/intersection.hpp
```diff
@@ -108,11 +115,60 @@ struct ConnectedRoad final : IntersectionViewData
...
+    {
+        return std::find_if(
+            self()->begin(), self()->end(), [eid](const auto &road) { return road.eid == eid; });
+    }
+
+    // find the maximum value based on a conversion operator
+    template <typename UnaryPredicate> auto findMaximum(const UnaryPredicate converter) const
+    {
+        BOOST_ASSERT(!self()->empty());
```
MoKob: well. Technically no intersection should ever be empty. Still, I am accessing `front()` here directly and don't simply iterator over all elements, returning possibly and `end()`.  If you check below, I don't assert for `!empty()` there, since that function is only iterating.
If we do an access to an element directly, I feel the additional assertion is justified.

Whats your take here?

include/extractor/guidance/intersection.hpp
```diff
@@ -108,11 +115,60 @@ struct ConnectedRoad final : IntersectionViewData
...
+        auto initial = converter(self()->front());
 
-using IntersectionShape = std::vector<IntersectionShapeData>;
+        const auto extract_maximal_value = [&initial, converter](const auto &road) {
+            initial = std::max(initial, converter(road));
+            return false;
+        };
+
+        boost::range::find_if(*self(), extract_maximal_value);
```
MoKob: changed other usages to boost ranges, only one place (that skips the u-turn edge) is left at `std` functions.

include/extractor/guidance/intersection.hpp
```diff
@@ -204,18 +241,38 @@ template <typename Self> struct EnableIntersectionOps
...
+            self()->begin(), self()->end(), [angle, &filter](const auto &lhs, const auto &rhs) {
+                const auto filtered_lhs = filter(lhs), filtered_rhs = filter(rhs);
+                const auto deviation_lhs = util::angularDeviation(lhs.angle, angle),
+                           deviation_rhs = util::angularDeviation(rhs.angle, angle);
+                return std::tie(filtered_lhs, deviation_lhs) <
+                       std::tie(filtered_rhs, deviation_rhs);
+            });
+
+        // make sure only to return valid elements
+        return filter(*candidate) ? self()->end() : candidate;
```
MoKob: you are correct, I've added the assertion here.

include/extractor/guidance/intersection_handler.hpp
```diff
@@ -508,15 +508,15 @@ std::size_t IntersectionHandler::findObviousTurn(const EdgeID via_edge,
             // even reverse the direction. Since we don't want to compute actual turns but simply
             // try to find whether there is a turn going to the opposite direction of our obvious
             // turn, this should be alright.
-            NodeID new_node;
-            const auto previous_intersection = [&]() {
-                EdgeID turn_edge;
-                std::tie(new_node, turn_edge) = intersection_generator.SkipDegreeTwoNodes(
+            const auto previous_intersection = [&]() -> IntersectionView {
+                const auto parameters = intersection_generator.SkipDegreeTwoNodes(
```
MoKob: Hm. I could potentially think of implementing a specialised graph walker. The skipping itself does not accumulate and can only handle degree two intersections. It's more efficient than the graph walking process that needs to generate intersections/bearings. What would be your take here? I'd also like to unify it, but the only thing I could think of is to split the skipping from the intersection_generator. It is tightly coupled to its usages to intersection generation, though.

include/extractor/guidance/mergable_road_detector.hpp
```diff
@@ -0,0 +1,149 @@
...
+                      const MergableRoadData &lhs,
+                      const MergableRoadData &rhs) const;
+
+    /* check if a road cannot influence the merging of the other. This is necessary to prevent
+     * situations with more than two roads that could participate in a merge */
+    bool IsDistinctFrom(const MergableRoadData &lhs, const MergableRoadData &rhs) const;
+
+  private:
+    // check if two name ids can be seen as identical (in presence of refs/others)
+    bool NameIDsAreCompatible(const NameID lhs, const NameID rhs) const;
```
MoKob: This is the name change heuristic in both directions. Clarified in comment. It uses the name change heuristic. Also changed name to `HaveIdenticalNames`

include/extractor/guidance/mergable_road_detector.hpp
```diff
@@ -0,0 +1,149 @@
...
+
+    /* When it comes to merging roads, we need to find out if two ways actually represent the
+     * same road. This check tries to identify roads which are the same road in opposite directions
+     */
+    bool RoadDataIsCompatible(const util::NodeBasedEdgeData &lhs_edge_data,
```
MoKob: Changed to `RoadDataSupportsMerge`

include/extractor/guidance/mergable_road_detector.hpp
```diff
@@ -0,0 +1,149 @@
...
+                              const util::NodeBasedEdgeData &rhs_edge_data) const;
+
+    /* Detect traffic loops.
+     * Since OSRM cannot handle loop edges, we cannot directly see a connection between a node and
+     * itself. We need to skip at least a single node in between. */
+    bool IsTrafficLoop(const NodeID intersection_node, const MergableRoadData &road) const;
```
MoKob: Hm. This could probably be used at some other location. I'd prefer to wait for other refactors when it comes up, though.

include/extractor/guidance/mergable_road_detector.hpp
MoKob: Well All of these are kind of specifically tailored to what we can merge into a single road. If these detectors trigger, all other handlers will not even see the respective roads (since they are merged into other ones). It might be that we can re-use them, for right now I would let them remain here. The worst thing that might happen is that we have to make this handler selection here accessible by making all private functions public.

src/extractor/geojson_debug_policies.cpp
MoKob: Isn't the `isA` relation already a strong coupling? I can do the full thing as a templated printer as well, if you would feel more comfortable with it. (see discussion above).

src/extractor/guidance/intersection_handler.cpp
MoKob: It shouldn't. It could be, that we change `SkipDegreeTwo` to return invalid IDs, in case we run into a dead end in the graph. So I feel it is safer to not assume too much about what happens when finding the next intersection

src/extractor/guidance/intersection_normalizer.cpp
MoKob: it's less unfortunate than left over debugging code :D

src/extractor/guidance/node_based_graph_walker.cpp
MoKob: hm. the coordinates are trimmed to the requested length as well. It essentially gives the information is the distance could be walked completely.

include/extractor/guidance/intersection.hpp
```diff
@@ -108,11 +115,60 @@ struct ConnectedRoad final : IntersectionViewData
 };
 
 // small helper function to print the content of a connected road
+std::string toString(const IntersectionShapeData &shape);
+std::string toString(const IntersectionViewData &view);
 std::string toString(const ConnectedRoad &road);
 
 // Intersections are sorted roads: [0] being the UTurn road, then from sharp right to sharp left.
+// common operations shared amongst all intersection types
+template <typename Self> struct EnableShapeOps
```
MoKob: Branched out into https://github.com/Project-OSRM/osrm-backend/issues/3493

include/extractor/guidance/intersection_generator.hpp
```diff
@@ -22,6 +22,13 @@ namespace extractor
 {
 namespace guidance
 {
+
+struct IntersectionGenerationParameters
+{
+    NodeID nid;
+    EdgeID via_eid;
+};
```
MoKob: Hm. I don't feel that it is reasonable to create an intermediate struct every time to call a handler. Whats your take?

include/extractor/guidance/mergable_road_detector.hpp
MoKob: Moved to forward declarations

src/extractor/guidance/mergable_road_detector.cpp
MoKob: At least I tried... It's not yet completely clear. There is a dedicated ticket for this, though: https://github.com/Project-OSRM/osrm-backend/issues/3473

MoKob:
Thanks again @daniel-j-h for the comments. I've addressed/branched out/commented all of them. Can you have a final look to see if I managed to address all of your concerns correctly.

include/extractor/guidance/intersection.hpp
daniel-j-h: Can we put a single assertion at locations where we generate `Intersection`s / `IntersectionView`s instead of doing it for every function call then? Roads should be immutable or do we ever remove / add roads?

include/extractor/guidance/intersection_generator.hpp
daniel-j-h: Hm yeah I think my main issue is with how we handle nodes and edges, and compositions of them. They're completely disconnected from the type system. That is, I can
- pass an edge id where a node id is required (they're both weak uint32_t aliases)
- pass a node id for a completely different intersection not corresponding to the via edge id

when in reality node ids should be typed and be distinguishable to edge ids at compile time. And starting node and via edge should probably, too. Then all our guidance code could work on those.

But I guess this is way out of scope for this pull request ..

include/extractor/guidance/intersection_handler.hpp
daniel-j-h: Ah I see -- the graph walker does too much here and we don't want to pay the incurred costs. How would a specialized graph walker look like: simply not generating intermediate `Intersection`s for skipping? Could this decision be a policy for the graph walker? And are there other use-cases for not generating intersections while walking the graph? (If not maybe just leave it as is)

src/extractor/geojson_debug_policies.cpp
daniel-j-h: Yeah you are right. We expose internals here and I fear this may all break down once we decide we need to change the `Intersection` / `IntersectionView` types again, though.

include/extractor/guidance/intersection_handler.hpp
MoKob: Hm. I haven't thought about it much, yet. But I'd say this is a different task, not related to this refactoring quest. We have two tickets regarding the performance already. I've added this as an additional task to https://github.com/Project-OSRM/osrm-backend/issues/3473.

PR #3119, opened 2016-10-20
Masonize

springmeyer:
This is a work in progress branch addressing #2790.

## Tasklist
 - [x] Further work to get building on linux
 - [x] Further work to get building on osx
 - [x] depends on #3130
 - [x] review
 - [x] adjust for for comments

PR #3212, opened 2016-11-01
FindOsmium invoke is required for MSVC build dependencies

oxidase:
# Issue

Commit 87d09f7 breaks msvc builds, because some osmium dependencies were missing.
PR is a partial revert of the commit

## Tasklist
 - [x] review
 - [ ] adjust for comments

## Requirements / Relations
Related issues #3088 and commit 87d09f7

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -457,7 +457,10 @@ endif()
 find_package(ZLIB REQUIRED)
 add_dependency_includes(${ZLIB_INCLUDE_DIRS})
 
-include_directories(SYSTEM ${CMAKE_CURRENT_SOURCE_DIR}/third_party/libosmium/include)
+list(APPEND CMAKE_MODULE_PATH "${CMAKE_CURRENT_SOURCE_DIR}/third_party/libosmium/cmake")
+set(OSMIUM_INCLUDE_DIR "${CMAKE_CURRENT_SOURCE_DIR}/third_party/libosmium/include")
+find_package(Osmium REQUIRED COMPONENTS io)
+include_directories(SYSTEM ${OSMIUM_INCLUDE_DIR})
 
 if(NOT WIN32 AND NOT Boost_USE_STATIC_LIBS)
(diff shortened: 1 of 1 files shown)


daniel-j-h:
This should always be required not only for MSVC builds, right? The libosmium CMake module does more than just setting the include path.

oxidase:
yes, it is also needed for Osmium_USE.* settings

TheMarex:
How does this interact with a system-installed libosmium version? Will the bundled version be preferred by the FindOsmium module?

oxidase:
At the moment there is no osmium configuration, just an unconditional addition of include directory. 

In the revert cmake  sets the cached variable to `third_party/libosmium/include` that is preferable, but it is possible to change to systems osmium as `-DOsmium_DEBUG=TRUE -DOSMIUM_INCLUDE_DIR=/usr/include`

springmeyer:
Sorry about breaking windows with #3119. Originally I had the `find_package(Osmium` still in use for the non ENABLE_MASON branch, but after PR review consolidated in one place.

I will followup now with another PR that fixes the ENABLE_MASON case, because `find_package(Osmium` will not work. This is because `find_package(Osmium` needs expat and bzip2 at system locations.

Currently ENABLE_MASON builds are failings with:

```
Make Error at mason_packages/linux-x86_64/cmake/3.6.2/share/cmake-3.6/Modules/FindPackageHandleStandardArgs.cmake:148 (message):
  Could NOT find Osmium (missing: EXPAT_FOUND BZIP2_FOUND)
  mason_packages/linux-x86_64/cmake/3.6.2/share/cmake-3.6/Modules/FindPackageHandleStandardArgs.cmake:388 (_FPHSA_FAILURE_MESSAGE)
  third_party/libosmium/cmake/FindOsmium.cmake:262 (find_package_handle_standard_args)
  CMakeLists.txt:468 (find_package)
```

I will fix this shortly as part of other work related to #3234

Issue #3081, opened 2016-10-18
Refactor Obvious Turn Discovery


MoKob:
The code to find which of the possible turns at an intersection could be considered `obvious` has grown a lot and needs to be refactored (see [here](https://github.com/Project-OSRM/osrm-backend/blob/9af00b19256c51de3d700024357199287e6b8ad3/src/extractor/guidance/intersection_handler.cpp#L377-L711))

I advice being very careful, though, since the discovery of obvious turns affects close to every turn scenario and is a very central component of the guidance process. Be advised that things will fall apart when you change it.

The main goal of obvious turn discovery is, as stated in the name, to decide which turn is obvious.
At a first glance, if you are on a road and there is a road with the same name, staying on it should be obvious.

But you don't always connect roads with same names, still passing straight over an intersection should be considered obvious.

At the same time, you can have intersections where the same road is actually not the obvious one to follow (e.g. [here](http://www.openstreetmap.org/#map=18/52.54578/13.33237)).

In general, we need to consider a lot of different factors like turn angles, road classes, names, other valid turns.... Therefore, the complexity has grown to where it is now. It has grown, however, and could be reduced and cleaned-up considerably.

daniel-j-h:
- [ ] `findObviousTurn` should return an `boost::optional<size_t>` to indicate if an obvious turn was found `boost::make_optional(index)` or not `boost::none`. Call sites then do `if (obv) use(*obv);`.

PR #3217, opened 2016-11-01
Refactor find obvious turn handling

karenzshea:
# Issue

https://github.com/Project-OSRM/osrm-backend/issues/3081

I made a pass at this code to hopefully make it cleaner and easier to follow, but more or less tried to preserve the same logic and handling via `best` and `best_continue` candidates.

## Tasklist
 - [x] Make a refactor pass at the second half of the function
-  [x] Investigate all of the cucumber test failures
-  [ ] ~~Return `boost::optional` instead of `0` for no obvious turn found~~ https://github.com/Project-OSRM/osrm-backend/issues/3337
-  [x] Comments
 - [x] review
 - [x] adjust for comments

## Requirements / Relations
cc @MoKob

### DIFF
--- a/features/guidance/collapse.feature
+++ b/features/guidance/collapse.feature
@@ -947,16 +947,17 @@ Feature: Collapse
 
     #http://www.openstreetmap.org/#map=19/52.48778/13.30024
     Scenario: Hohenzollerdammbrücke
+        Given a grid size of 10 meters
         Given the node map
             """
                   q          s
                   p          o
-                  ..       . .
-                .     .  .      .
+                  ..        ..
+                 .    .   .    .
             j - i - - - h - - - g - f
                   > k <   > l <
             a - b - - - c - - - d - e
-                .     .  .      .
+                 .    .  .     .
                   ..        ..
                   m          n
                   t          r
@@ -1013,13 +1014,13 @@ Feature: Collapse
             | restriction | ph       | hi     | h        | no_right_turn |
 
         When I route I should get
-            | waypoints | route          | turns                       |
-            | a,e       | hohe,hohe      | depart,arrive               |
-            | a,s       | hohe,a100,a100 | depart,on ramp left,arrive  |
-            | a,t       | hohe,a100,a100 | depart,on ramp right,arrive |
-            | a,j       |                |                             |
-            | f,j       | hohe,hohe      | depart,arrive               |
-            | a,t       | hohe,a100,a100 | depart,on ramp right,arrive |
-            | f,e       |                |                             |
-            | q,j       | a100,hohe,hohe | depart,turn right,arrive    |
-            | q,e       | a100,a100,hohe | depart,continue left,arrive |
+            | waypoints | route                 | turns                       | locations |
+            | a,e       | hohe,hohe             | depart,arrive               | a,e       |
+            | a,s       | hohe,a100,a100        | depart,on ramp left,arrive  | a,b,s     |
+            | a,t       | hohe,a100,a100        | depart,on ramp right,arrive | a,b,t     |
+            | a,j       |                       |                             |           |
+            | f,j       | hohe,hohe             | depart,arrive               | f,j       |
+            | a,t       | hohe,a100,a100        | depart,on ramp right,arrive | a,b,t     |
+            | f,e       |                       |                             |           |
+            | q,j       | a100,hohe,hohe        | depart,turn right,arrive    | q,p,j     |
+            | q,e       | a100,hohebruecke,hohe | depart,turn left,arrive     | q,p,e     |
--- a/features/guidance/low-priority.feature
+++ b/features/guidance/low-priority.feature
@@ -1,0 +1,119 @@
+@routing  @guidance
+Feature: Exceptions for routing onto low-priority roads
+
+    Background:
+        Given the profile "car"
+        Given a grid size of 10 meters
+
+    Scenario: Straight onto low-priority: same name
+        Given the node map
+            """
+                c
+
+            a   b d
+
+                e
+            """
+
+        And the ways
+            | nodes  | highway     | name    |
+            | abd    | residential | road    |
+            | eb     | service     | service |
+            | bc     | service     | service |
+
+       When I route I should get
+            | waypoints | route            | turns         |
+            | c,e       | service,service  | depart,arrive |
+            | e,c       | service,service  | depart,arrive |
+
+    Scenario: Straight onto low-priority: onto and from unnamed
+        Given the node map
+            """
+                c
+
+            a   b d
+
+                e
+            """
+
+        And the ways
+            | nodes  | highway     | name    |
+            | abd    | residential | road    |
+            | eb     | service     |         |
+            | bc     | service     |         |
+
+       When I route I should get
+            | waypoints | route | turns         |
+            | e,c       | ,     | depart,arrive |
+            | c,e       | ,     | depart,arrive |
+
+    Scenario: Straight onto low-priority: unnamed
+        Given the node map
+            """
+                c
+
+            a   b d
+
+                e
+            """
+
+        And the ways
+            | nodes  | highway     | name    |
+            | abd    | residential | road    |
+            | eb     | service     | service |
+            | bc     | service     |         |
+
+       When I route I should get
+            | waypoints | route             | turns                       |
+            | e,c       | service,          | depart,arrive               |
+            | c,e       | ,service,service  | depart,turn straight,arrive |
+
+    Scenario: Straight onto low-priority
+        Given the node map
+            """
+            a b c
+            """
+
+        And the ways
+            | nodes  | highway     | name    |
+            | ab     | residential | road    |
+            | bc     | service     | service |
+
+       When I route I should get
+            | waypoints | route                | turns                           |
+            | a,c       | road,service,service | depart,new name straight,arrive |
+
+    Scenario: Straight onto low-priority, with driveway
+        Given the node map
+            """
+                  f
+            a b   c
+            """
+
+        And the ways
+            | nodes  | highway     | name  |
+            | ab     | residential | road  |
+            | bc     | service     | road  |
+            | bf     | driveway    |       |
+
+       When I route I should get
+            | waypoints | route      | turns         |
+            | a,c       | road,road  | depart,arrive |
+
+    Scenario: Straight onto low-priority, with driveway
+        Given the node map
+            """
+                  f
+            a b   c
+            """
+
+        And the ways
+            | nodes  | highway     | name  |
+            | ab     | residential | road  |
+            | bc     | service     |       |
+            | bf     | driveway    |       |
+
+       When I route I should get
+            | waypoints | route      | turns                           |
+            | a,c       | road,      | depart,arrive                   |
+            | c,a       | ,road,road | depart,new name straight,arrive |
--- a/src/extractor/guidance/intersection_handler.cpp
+++ b/src/extractor/guidance/intersection_handler.cpp
@@ -1,5 +1,5 @@
-#include "extractor/guidance/intersection_handler.hpp"
 #include "extractor/guidance/constants.hpp"
+#include "extractor/guidance/intersection_handler.hpp"
 #include "extractor/guidance/toolkit.hpp"
 
 #include "util/coordinate_calculation.hpp"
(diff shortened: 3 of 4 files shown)


karenzshea:
Test failures so far look like they're of a few kinds:
- `turn straight` given over a `new name straight`
- `continue straight` added where previously not given
- `turn slight {left|right}` added where previously not given

The latter two I imagine are cases where an intersection is not marked with an obvious turn when it should be. The first case I'm less sure about but might be related.

karenzshea:
OK - I've fixed the majority of the test failures noted above. New set of failures among 9 tests:
- `new name` becomes turn, or `turn` becomes `new name`
- `fork` is collapsed away
- `fork` becomes `turn`

Regarding the fork cases, I need to look into how a fork is determined and how it uses obvious turn. Where I'm seeing a turn instruction instead of a fork I feel like with either a turn or a fork, we'd want to declare no obvious turn.

MoKob:
@karenzshea a few remarks given the failing cases:

collapse.feature:364 - this can be seen either way. I said if the continue road does not differ by more than 90 degrees from the straightmost turn, the continue/best can probably not be easily distinguished. (the problem here are lane sizes). If you add number of lanes into that comparison, we could probably just keep the `new name straight`.

continue.feature:8/62 - same here

fork.feature:254 - this one is really broken. You seem to not consider if the road is really distinct to the left/right anymore. At least the two turns here are very close and you seemed to simply have chosen the straight one.

perception.feature:133 - also distinction: here you need to compare how strong the deviation from going straight is. You have two turns (left, slight right) but both don't differ much in their total deviation from straight.

ramp.feature:112 - I defined that entering a motorway is always less obvious than staying on a street (since its an obvious change, big signs...). This is not considered anymore as it seems.

roundabout.feature:335 - change should be fine, different but valid interpretation of the turn. Feel free to adjust the case

turn-angles.feature:684 / turn-lanes.feature:578 - this is an example where pure selection by `continue first`, `turns later` does not work. Continue is actually turning around and going straight suddenly does not look obvious anymore. Any segregated intersection can switch names and has to be treated just like a four way intersection. Merging roads tries to hide this from the code, still there are situations were merging fails/cannot be done safely and you don't seem to consider these.

karenzshea:
@MoKob rebased onto master :+1: 

Can you elaborate more on how this failing perception.features:133 is supposed to work? It looks like this test illustrates a roundabout. Do we want to emit an obvious turn in cases where the turn is out of the roundabout? It seems like an obvious turn is not being returned, but that a `new name` instruction is being returned over a `turn` one.

collapse.feature:931 I still need to dig into the this, this is a segregated intersection that I don't fully understand yet. Seems like something is not being properly collapsed down the line. What's strangest to me is that an extra `slight right` is being emitted, though I see no slight rights to be made in this scenario.

MoKob:
@karenzshea the `perception.feature:133` is a road around a place, but not a roundabout. I have to say though, that I find it curious that no obvious turn should be involved here. Emitting `NewName` over `Turn` is practically the only way I see in the code that would end up outputting a new name here. What you have is when coming to an intersection is a turn of ~60 degree to the left and a turn of about ~45-50 to the right. Both have the same name. when driving left to right (or back) and the exit has a different name from top to bottom (and back). These two are failing. It seems that you prefer the ever so much slighter turn here, which is not that obvious to me.

For the `collapse.feature:931`, the slight turn is the additional turn at the top from the segment `qp` to `pi`. The original setting detects that turn as obvious and suppresses it. If you are comparing the sattelite images at the linked location and the scenario itself, it seems that the scenario might not be modelled closely enough to the real situation. It seems the segments between the turns might be too long. You could move the scenario to a smaller grid size (the current collapse.feature has 20 meters, which is quite wide) and adjust the angles to better fit the original scenario.

features/guidance/collapse.feature
```diff
@@ -995,13 +996,13 @@ Feature: Collapse
...
+            | waypoints | route                 | turns                       |
+            | a,e       | hohe,hohe             | depart,arrive               |
+            | a,s       | hohe,a100,a100        | depart,on ramp left,arrive  |
+            | a,t       | hohe,a100,a100        | depart,on ramp right,arrive |
+            | a,j       |                       |                             |
+            | f,j       | hohe,hohe             | depart,arrive               |
+            | a,t       | hohe,a100,a100        | depart,on ramp right,arrive |
+            | f,e       |                       |                             |
+            | q,j       | a100,hohe,hohe        | depart,turn right,arrive    |
+            | q,e       | a100,hohebruecke,hohe | depart,turn left,arrive     |
```
karenzshea: @MoKob can you confirm for me that this is an OK change, from a `continue left` to a `turn left`? A `continue` would be appropriate if the street being turned onto were named the same as the approach street, but this isn't the case here.

features/guidance/turn-angles.feature
```diff
@@ -742,6 +742,7 @@ Feature: Simple Turns
 
     #http://www.openstreetmap.org/#map=19/52.53293/13.32956
     Scenario: Curved Exit from Curved Road
+        Given a grid size of 5 meters
         Given the node map
```
karenzshea: @MoKob Can you approve this change as well? The issue I was seeing with a grid size of 1m as it was before is that it wasn't accurately representing the linked OSM intersection.

While debugging, I got around to logging the calculated angle deviation for each intersection, and somehow the angle for `dhij` was `19.9583`, and the angle for `defg` was `27.2692`, which I believe misrepresents the scenario we're trying to model. This explains why the minor `Siemens` street was always being considered the obvious turn.

As a result, when the 90 degree turn onto the residential `Siemens` street is properly modelled the instructions returned also needed to be updated as proper turns rather than slight turns.

karenzshea:
After fixing the other test failures, the two types I'm left with now are kind of tricky. Still seeing the same issue with `perception.feature:133` where a `NewName` is emitted over a `turn`.

The other test failures, which are all new, seem to be the same issue: previously suppressed `continue straight`s are not suppressed. As far as I can tell though, the correct obvious turns for continuing straight are still emitted from the obvious turn finder. Need to dig on this when I have a clear mind again.

features/guidance/collapse.feature
MoKob: With the feature in, can you add `locations` to the test and check the turn locations? For the respective change, locations should be `q,p,e`

src/extractor/guidance/intersection_handler.cpp
```diff
@@ -383,131 +394,147 @@ std::size_t IntersectionHandler::findObviousTurn(const EdgeID via_edge,
...
+        out_ways.push_back(out);
+    }
 
-        const bool other_is_better_choice_by_priority =
-            best != 0 && obviousByRoadClass(in_data.road_classification,
-                                            current_best_class,
-                                            out_data.road_classification);
+    const auto End = end(out_ways);
+    // dead end, or otherwise no exitable ways
+    if (num_valid_entry == 0)
```
MoKob: should be plural

src/extractor/guidance/intersection_handler.cpp
```diff
@@ -383,131 +394,147 @@ std::size_t IntersectionHandler::findObviousTurn(const EdgeID via_edge,
...
-            best = i;
-        }
+    // sort out ways by what they share with the in way
+    // entry allowed, lowest deviation to greatest deviation, name, classification and priority
+    const auto sort_by = [](const out_way &lhs, const out_way &rhs) {
+        auto left_tie =
+            std::tie(lhs.same_classification, lhs.same_or_higher_priority);
+        auto right_tie =
+            std::tie(rhs.same_classification, rhs.same_or_higher_priority);
+        return lhs.entry_allowed > rhs.entry_allowed ||
```
MoKob: I don't really like comparing `bool` with `>` which is kind of unintuitive.

Apart from this, why the explicit check for some, the `tie` based check for the others. Should be possible to make all checks compatible with `<` (e.g. `left_tie > right_tie` -> `right_tie < left_tie`) and finally simply compare `smaller factors < larger_factors`

src/extractor/guidance/intersection_handler.cpp
```diff
@@ -383,131 +394,147 @@ std::size_t IntersectionHandler::findObviousTurn(const EdgeID via_edge,
...
+               (lhs.entry_allowed == rhs.entry_allowed &&
+                (lhs.deviation_from_straight < rhs.deviation_from_straight ||
+                 (lhs.deviation_from_straight == rhs.deviation_from_straight &&
+                  left_tie > right_tie)));
+    };
+    std::stable_sort(begin(out_ways), end(out_ways), sort_by);
+
+    // best candidate is the first way sorted to have the smallest deviation from
+    // straight and is same or higher priority class
+    const auto sameOrHigher = [&](const out_way &way) {
```
MoKob: not necessarily true. `motorway` to `motorway_link` in presence of a `service` will also be obvious, even though the priority drops. And this does not even have to be an explicit low priority road. Also `motorway->primary` (within cities this is a real possibility) next to a tertiary road.

src/extractor/guidance/intersection_handler.cpp
```diff
@@ -383,131 +394,147 @@ std::size_t IntersectionHandler::findObviousTurn(const EdgeID via_edge,
...
+        return way.entry_allowed && way.same_or_higher_priority;
+    };
+
+    const auto notLowPriority = [&](const out_way &way) {
+        return way.entry_allowed &&
+               !node_based_graph.GetEdgeData(way.road->eid)
+                    .road_classification.IsLowPriorityRoadClass();
+    };
+    auto best_candidate_it = std::find_if(begin(out_ways), End, sameOrHigher);
```
MoKob: isn't this was the sorting is for? To make sure the best candidate is at the front?

src/extractor/guidance/intersection_handler.cpp
```diff
@@ -383,131 +394,147 @@ std::size_t IntersectionHandler::findObviousTurn(const EdgeID via_edge,
...
+    // No non-low priority roads? Declare no obvious turn
     if (best == 0)
         return 0;
 
+    // get a count of number of ways from that intersection that qualify to have
+    // continue instruction because they share a name with the approaching way
     const std::pair<std::int64_t, std::int64_t> num_continue_names = [&]() {
         std::int64_t count = 0, count_valid = 0;
-        if (in_data.name_id != EMPTY_NAMEID)
+        std::size_t i, last = out_ways.size();
```
MoKob: `last` is somewhat a synonym for the element itself. Can we stick with size. Also don't use `i` as index, ever. Searching in the code for usage will be made impossible.

Also, this could be an std::algorithm `find_if` returning false and keeping count.
```
[&count,&count_valid]( ... way)
{
    if( same_name_id )
       ++count;
    if( entry_allowed)
       ++valid;
    return false;
}
```

Or (since it is small), directly do two algorithms using `count_if` with dedicated lamdas.

src/extractor/guidance/intersection_handler.cpp
```diff
@@ -383,131 +394,147 @@ std::size_t IntersectionHandler::findObviousTurn(const EdgeID via_edge,
...
+    // no out ways share the same name as the approach street
+    if (num_continue_names.first == 0)
+    {
+        best_continue = 0;
+    }
+    else
+    {
+        // continue roads exist
+        // best_continue is the first sorted way that shares the same name as the in way
+        const auto sameName = [&](const out_way &lhs) {
```
MoKob: sameName `doesn't` capture the full usage of this lamda

src/extractor/guidance/intersection_handler.cpp
karenzshea: Issue here is that because I'm comparing `bool`s, which we do want sorted with `>` but we want deviation to be sorted from `<`

MoKob: But you could still do:

`std::tie(a0,b1,a2) < std::tie(b0,a1,b2)` to achieve `a0 < a1`, if equal `a1 > a2`, if equal `a2 < b2`, right? After all `a0 < b0` is equivalent to `b0 > a0`. So all `>` and `<` can be switched by choosing the comparison side.

src/extractor/guidance/intersection_handler.cpp
karenzshea: the lambda only checks if lhs has the same name as the approach way now.

src/extractor/guidance/intersection_handler.cpp
```diff
@@ -12,6 +12,16 @@
 using EdgeData = osrm::util::NodeBasedDynamicGraph::EdgeData;
 using osrm::util::guidance::getTurnDirection;
 
+struct out_way
```
TheMarex: This doesn't adhere to our naming convention for structs which is CamelCase.

src/extractor/guidance/intersection_handler.cpp
```diff
@@ -383,136 +393,140 @@ std::size_t IntersectionHandler::findObviousTurn(const EdgeID via_edge,
...
-                       const auto same_name =
-                           !util::guidance::requiresNameAnnounced(in_data.name_id,
-                                                                  road_data.name_id,
-                                                                  name_table,
-                                                                  street_name_suffix_table);
-
-                       return same_name &&
-                              angularDeviation(road.angle, STRAIGHT_ANGLE) < NARROW_TURN_ANGLE;
-                   }) == num_continue_names.first;
+        return std::count_if(begin(out_ways), end(out_ways), [&](const out_way &road) {
```
TheMarex: This does not need to be wrapped in a lambda. You can write `const bool all_continues_are_narrow = std::count_if(...) == num_continue_names.first`.

src/extractor/guidance/intersection_handler.cpp
```diff
@@ -12,6 +12,16 @@
 using EdgeData = osrm::util::NodeBasedDynamicGraph::EdgeData;
 using osrm::util::guidance::getTurnDirection;
 
+struct out_way
+{
+    bool same_name_id = false;
+    bool same_classification = false;
+    bool same_or_higher_priority = false;
+    double deviation_from_straight = 180;
+    int index = 0;
```
MoKob: `index` in itself is a non-descriptive name. Index into what?

src/extractor/guidance/intersection_handler.cpp
```diff
@@ -383,136 +393,140 @@ std::size_t IntersectionHandler::findObviousTurn(const EdgeID via_edge,
...
-        const auto continue_class =
-            node_based_graph.GetEdgeData(intersection[best_continue].eid).road_classification;
+    // create sortable vector of only entry allowed out_ways
+    // get a count of number of ways from that intersection that qualify to have
+    // continue instruction because they share a name with the approaching way
+    std::int64_t continue_count = 0, continue_count_valid = 0;
+    for (std::size_t i = 1, last = intersection.size(); i < last; ++i)
+    {
+        out_way out;
+        const auto out_node = node_based_graph.GetEdgeData(intersection[i].eid);
```
MoKob: this is not a node, this is edge data? Nodes are usually entries in the graph. The chosen name here is very confusing.

src/extractor/guidance/intersection_handler.cpp
```diff
@@ -383,136 +393,140 @@ std::size_t IntersectionHandler::findObviousTurn(const EdgeID via_edge,
...
 
-        const auto same_name = !util::guidance::requiresNameAnnounced(
-            in_data.name_id, out_data.name_id, name_table, street_name_suffix_table);
+        out.road = &intersection[i];
```
MoKob: why do we access the address of the road here? isn't it good enough to store the `index`?
the value here is accessed on two locations. Once to check for entry (which is unnecessary right now, you only keep allowed entries, but see my other two comments on still keeping them) and to check for the edge ID.

I'd rather store the two values directly (even less memory required), and avoid keeping these pointers.

src/extractor/guidance/intersection_handler.cpp
```diff
@@ -383,136 +393,140 @@ std::size_t IntersectionHandler::findObviousTurn(const EdgeID via_edge,
...
-                          (deviation < best_continue_deviation &&
-                           out_data.road_classification == continue_class) ||
-                          (continue_class != in_classification &&
-                           out_data.road_classification == continue_class)))
+        if (!out.road->entry_allowed)
         {
-            best_continue_deviation = deviation;
-            best_continue = i;
+            ++continue_count_valid;
+            continue;
```
MoKob: don't we need to initialise the struct completely? Jumping off early here leaves `index, deviation from straight`... uninitialised, doesn't it?

src/extractor/guidance/intersection_handler.cpp
```diff
@@ -383,136 +393,140 @@ std::size_t IntersectionHandler::findObviousTurn(const EdgeID via_edge,
...
-                                                 in_classification != continue_class) ||
-                          (deviation < best_continue_deviation &&
-                           out_data.road_classification == continue_class) ||
-                          (continue_class != in_classification &&
-                           out_data.road_classification == continue_class)))
+        if (!out.road->entry_allowed)
         {
-            best_continue_deviation = deviation;
-            best_continue = i;
+            ++continue_count_valid;
```
MoKob: isn't this continue count `invalid` then?

src/extractor/guidance/intersection_handler.cpp
```diff
@@ -383,136 +393,140 @@ std::size_t IntersectionHandler::findObviousTurn(const EdgeID via_edge,
...
+            continue;
         }
 
-        const auto current_best_class =
-            node_based_graph.GetEdgeData(intersection[best_continue].eid).road_classification;
+        out.index = i;
+        out.deviation_from_straight = angularDeviation(intersection[i].angle, STRAIGHT_ANGLE);
+        out.same_classification = out_node.road_classification == in_way.road_classification;
+        out.same_or_higher_priority =
+            out_node.road_classification.GetPriority() <= in_way.road_classification.GetPriority();
```
MoKob: completely ignoring invalid roads is not an option, in my opinion.

in
```
a - b - c
d `
```
and `c-b` a `oneway` I wouldn't feel that it is obvious to turn onto `d`. I'd be assured if the navigation system actually told me to take a sharp turn here. When I come from `a` to `d`

src/extractor/guidance/intersection_handler.cpp
```diff
@@ -383,136 +393,140 @@ std::size_t IntersectionHandler::findObviousTurn(const EdgeID via_edge,
...
+            ++continue_count_valid;
+            continue;
         }
 
-        const auto current_best_class =
-            node_based_graph.GetEdgeData(intersection[best_continue].eid).road_classification;
+        out.index = i;
+        out.deviation_from_straight = angularDeviation(intersection[i].angle, STRAIGHT_ANGLE);
+        out.same_classification = out_node.road_classification == in_way.road_classification;
+        out.same_or_higher_priority =
```
MoKob: couldn't all of these predefined comparisons be part of the sorting operator? Why do we have to precompute them? This makes look-up a bit more confusing.

src/extractor/guidance/intersection_handler.cpp
```diff
@@ -383,136 +393,140 @@ std::size_t IntersectionHandler::findObviousTurn(const EdgeID via_edge,
...
+            out_node.road_classification.GetPriority() <= in_way.road_classification.GetPriority();
 
-        // don't prefer low priority classes
-        if (best != 0 && out_data.road_classification.IsLowPriorityRoadClass() &&
-            !current_best_class.IsLowPriorityRoadClass())
-            continue;
+        out_ways.push_back(out);
+    }
+    const std::pair<std::int64_t, std::int64_t> num_continue_names =
+        std::make_pair(continue_count, continue_count_valid);
```
MoKob: see above: this is `continue_count_invalid`. Also, what is the benefit in storing this in a pair? `num_continue_name.first/.second` is far less intuitive then a named variable.

src/extractor/guidance/intersection_handler.cpp
```diff
@@ -383,136 +393,140 @@ std::size_t IntersectionHandler::findObviousTurn(const EdgeID via_edge,
...
+    // dead end, or otherwise no exitable ways
+    if (out_ways.size() == 0)
+        return 0;
 
-        const bool other_is_better_choice_by_priority =
-            best != 0 && obviousByRoadClass(in_data.road_classification,
-                                            current_best_class,
-                                            out_data.road_classification);
+    // sort out ways by what they share with the in way
+    const auto sort_by = [](const out_way &lhs, const out_way &rhs) {
```
MoKob: what would `std::sort(out_ways.begin(),out_was.end(),sort_by)` mean? Can we choose a descriptive name here?

src/extractor/guidance/intersection_handler.cpp
```diff
@@ -383,136 +393,140 @@ std::size_t IntersectionHandler::findObviousTurn(const EdgeID via_edge,
...
+        return left_tie > right_tie;
+    };
+    std::stable_sort(begin(out_ways), End, sort_by);
+
+    // the ideal obvious turn candidate is the first way sorted to have the smallest deviation from
+    // straight and is same or higher priority class
+    const auto sameOrHigher = [&](const out_way &way) {
+        return way.same_or_higher_priority == true;
+    };
+    auto best_candidate_it = std::find_if(begin(out_ways), End, sameOrHigher);
```
MoKob: the sorting itself should make sure that the first candidate has `same` or higher priority, and if not that the first one in the list is the one to choose? Why do we need two additional find-ifs, if we compare by so many criteria above?

As long as you use `sameOrHigher` and `notLowPriority` as first comparators in the sort, your sorting should end up having these at the front.

src/extractor/guidance/intersection_handler.cpp
```diff
@@ -383,136 +393,140 @@ std::size_t IntersectionHandler::findObviousTurn(const EdgeID via_edge,
...
+    const std::pair<std::int64_t, std::int64_t> num_continue_names =
+        std::make_pair(continue_count, continue_count_valid);
 
-        const bool is_better_choice_by_priority =
-            best == 0 || obviousByRoadClass(in_data.road_classification,
-                                            out_data.road_classification,
-                                            current_best_class);
+    const auto End = end(out_ways);
+    // dead end, or otherwise no exitable ways
+    if (out_ways.size() == 0)
```
MoKob: if you include invalid entries (see comment above), this check can be done right at the start on `intersection`

src/extractor/guidance/intersection_handler.cpp
```diff
@@ -383,136 +393,140 @@ std::size_t IntersectionHandler::findObviousTurn(const EdgeID via_edge,
...
-                                            out_data.road_classification);
+    // sort out ways by what they share with the in way
+    const auto sort_by = [](const out_way &lhs, const out_way &rhs) {
+        auto left_tie = std::tie(
+            rhs.deviation_from_straight, lhs.same_classification, lhs.same_or_higher_priority);
+        auto right_tie = std::tie(
+            lhs.deviation_from_straight, rhs.same_classification, rhs.same_or_higher_priority);
+        return left_tie > right_tie;
+    };
+    std::stable_sort(begin(out_ways), End, sort_by);
```
MoKob: by doing stable sort, you prefer right turns over left turns in the following `find_if` calls. Intended? Can this have side-effects?

src/extractor/guidance/intersection_handler.cpp
```diff
@@ -383,136 +393,140 @@ std::size_t IntersectionHandler::findObviousTurn(const EdgeID via_edge,
...
+    const auto notLowPriority = [&](const out_way &way) {
+        return !node_based_graph.GetEdgeData(way.road->eid)
+                    .road_classification.IsLowPriorityRoadClass();
+    };
+    if (best_candidate_it == End)
+        best_candidate_it = std::find_if(begin(out_ways), End, notLowPriority);
+
+    if (best_candidate_it != End)
+    {
+        best = best_candidate_it->index;
```
MoKob: I know I choose these names _back then_, but they really suck. Can we improve them in this refactor?

src/extractor/guidance/intersection_handler.cpp
```diff
@@ -383,136 +393,140 @@ std::size_t IntersectionHandler::findObviousTurn(const EdgeID via_edge,
...
+        best_deviation = best_candidate_it->deviation_from_straight;
+    }
 
-        if ((!other_is_better_choice_by_priority && deviation < best_deviation) ||
-            is_better_choice_by_priority)
+    // double check if the sorted lowest deviation from straight would still be better choice
+    if (static_cast<uint32_t>(out_ways[0].index) != best)
+    {
+        const out_way first = out_ways[0];
```
MoKob: reference

src/extractor/guidance/intersection_handler.cpp
```diff
@@ -383,136 +393,140 @@ std::size_t IntersectionHandler::findObviousTurn(const EdgeID via_edge,
...
+        const auto not_ramp_class = !node_based_graph.GetEdgeData(intersection[first.index].eid)
+                                         .road_classification.IsRampClass();
+        const auto not_link_class = !node_based_graph.GetEdgeData(intersection[first.index].eid)
+                                         .road_classification.IsLinkClass();
+        if (deviation_diff && notLowPriority(first) && not_ramp_class && not_link_class)
         {
-            best_deviation = deviation;
-            best = i;
+            best = first.index;
+            best_deviation = first.deviation_from_straight;
```
MoKob: what would this set-up end up preferring in:

```

                  .      d
a - - - b - - - - c

with the ways
`abd, motorway`
and `bc` tertiary.
```
From what I read this would end up substituting the perfectly good continuing road with a `continue slight left` and offer only a `new name` for going from `a` to `c`.

For example here, the turn to the right might actually be small enough as an angle. (Note: they also all share the same name. A nice present from the road planners in Germany)

screen shot 2016-11-14 at 09 45 05

src/extractor/guidance/intersection_handler.cpp
```diff
@@ -383,136 +393,140 @@ std::size_t IntersectionHandler::findObviousTurn(const EdgeID via_edge,
...
-        {
-            for (std::size_t i = 1; i < intersection.size(); ++i)
-            {
-                const auto &road = intersection[i];
-                const auto &road_data = node_based_graph.GetEdgeData(road.eid);
-
-                const auto same_name = !util::guidance::requiresNameAnnounced(
-                    in_data.name_id, road_data.name_id, name_table, street_name_suffix_table);
+    // best_continue is the first sorted way that shares the same name as the in way
+    const auto sameName = [&](const out_way &lhs) { return lhs.same_name_id == true; };
```
MoKob: from all these checks, I get the impression that the only sorting you require above is actually the `angularDifference`?

src/extractor/guidance/intersection_handler.cpp
```diff
@@ -383,136 +393,140 @@ std::size_t IntersectionHandler::findObviousTurn(const EdgeID via_edge,
...
+    const auto best_continue_it = std::find_if(begin(out_ways), End, sameName);
+    if (best_continue_it != End)
+    {
+        best_continue = best_continue_it->index;
+        best_continue_deviation = best_continue_it->deviation_from_straight;
+    }
```
MoKob: shouldn't we be done elsewise? why don't return best here instantly in an `else` branch?

src/extractor/guidance/intersection_handler.cpp
```diff
@@ -383,136 +393,140 @@ std::size_t IntersectionHandler::findObviousTurn(const EdgeID via_edge,
...
-                {
-                    ++count;
-                    if (road.entry_allowed)
-                        ++count_valid;
-                }
-            }
-        }
-        return std::make_pair(count, count_valid);
-    }();
+    const auto &best_data = node_based_graph.GetEdgeData(intersection[best].eid);
```
MoKob: again, I would prefer deprecating the crappy names I chose back then.

src/extractor/guidance/intersection_handler.cpp
```diff
@@ -383,136 +393,140 @@ std::size_t IntersectionHandler::findObviousTurn(const EdgeID via_edge,
     if (intersection.size() == 2)
         return 1;
```
MoKob: wasn't the plan to substitute these angle and `0` is `no obvious candidate` by `boost::optional`?

karenzshea: ooh, forgot about this. I'll add this to the op.

src/extractor/guidance/intersection_handler.cpp
karenzshea: I was following what was here before, and didn't really change this part but is definitely doable.

src/extractor/guidance/intersection_handler.cpp
```diff
@@ -400,108 +393,105 @@ std::size_t IntersectionHandler::findObviousTurn(const EdgeID via_edge,
...
     {
-        out_way out;
-        const auto out_node = node_based_graph.GetEdgeData(intersection[i].eid);
+        const auto out_data = node_based_graph.GetEdgeData(intersection[i].eid);
 
-        out.road = &intersection[i];
-        out.same_name_id = !util::guidance::requiresNameAnnounced(
-            in_way.name_id, out_node.name_id, name_table, street_name_suffix_table);
-        if (out.same_name_id)
+        bool out_same_name_id = !util::guidance::requiresNameAnnounced(
```
MoKob: can be put into a lambda `const auto is_continue_road = [&in_way,this](const ConnectedRoad &road){ const auto &out_data = ...; return !requires...; }`

After you can call `std::count_if(intersection.begin(),intersection.end(),is_continue_road)` to count the continues.

The `count valid` part can probably be moved to an additional check in `intersection` (just as findClosestTurn) since it can be used at other parts of the code as well.

src/extractor/guidance/intersection_handler.cpp
```diff
@@ -400,108 +393,105 @@ std::size_t IntersectionHandler::findObviousTurn(const EdgeID via_edge,
...
-        return 0;
-
-    // sort out ways by what they share with the in way
-    const auto sort_by = [](const out_way &lhs, const out_way &rhs) {
-        auto left_tie = std::tie(
-            rhs.deviation_from_straight, lhs.same_classification, lhs.same_or_higher_priority);
-        auto right_tie = std::tie(
-            lhs.deviation_from_straight, rhs.same_classification, rhs.same_or_higher_priority);
-        return left_tie > right_tie;
+    const auto SameName = [&](const EdgeData &way_data) {
```
MoKob: this is a difference from what you checked before, see my comment above.

src/extractor/guidance/intersection_handler.cpp
```diff
@@ -400,108 +393,105 @@ std::size_t IntersectionHandler::findObviousTurn(const EdgeID via_edge,
...
+        return way_data.name_id == in_way.name_id;
+    };
+    auto sameOrHigherPriority = [&](const auto &way_data) {
```
MoKob: capture defaults are sometimes nice, but if it's only a single value you need, I'd prefer capturing explicitly (`[&in_way]`).

Also `way_data` vs `in_way`, both should be ways or data, right?

src/extractor/guidance/intersection_handler.cpp
```diff
@@ -400,108 +393,105 @@ std::size_t IntersectionHandler::findObviousTurn(const EdgeID via_edge,
...
+    auto IsLowPriority = [&](const auto &way_data) {
+        return way_data.road_classification.IsLowPriorityRoadClass();
+    };
+    const auto RoadCompare = [&](const ConnectedRoad &lhs, const ConnectedRoad &rhs) {
+        const EdgeData &lhs_data = node_based_graph.GetEdgeData(lhs.eid);
+        const EdgeData &rhs_data = node_based_graph.GetEdgeData(rhs.eid);
+        const auto lhs_deviation = angularDeviation(lhs.angle, STRAIGHT_ANGLE);
+        const auto rhs_deviation = angularDeviation(rhs.angle, STRAIGHT_ANGLE);
+
+        bool rhs_same_or_priority = sameOrHigherPriority(rhs_data);
```
MoKob: `const`, also missing `higher`

src/extractor/guidance/intersection_handler.cpp
```diff
@@ -400,108 +393,105 @@ std::size_t IntersectionHandler::findObviousTurn(const EdgeID via_edge,
...
+        bool rhs_low_priority = IsLowPriority(rhs_data);
+        bool lhs_same_or_priority = sameOrHigherPriority(lhs_data);
+        bool lhs_low_priority = IsLowPriority(lhs_data);
+        auto left_tie = std::tie(rhs.entry_allowed, rhs_same_or_priority, lhs_low_priority, lhs_deviation);
```
MoKob: deviation is now the least important criterion. That needs to be in the front, right? Same order as before:
`lhs.deviation_from_straight, rhs.same_classification, rhs.same_or_higher_priority`. Only keeping the `entry_allowed` in front is correct, since you filtered these before.

src/extractor/guidance/intersection_handler.cpp
```diff
@@ -400,108 +393,105 @@ std::size_t IntersectionHandler::findObviousTurn(const EdgeID via_edge,
...
-    };
-    if (best_candidate_it == End)
-        best_candidate_it = std::find_if(begin(out_ways), End, notLowPriority);
-
-    if (best_candidate_it != End)
+    BOOST_ASSERT(best_it != end(intersection));
+    best = std::distance(begin(intersection), best_it);
+    best_deviation = angularDeviation(intersection[best].angle, STRAIGHT_ANGLE);
+    const auto &best_data = node_based_graph.GetEdgeData(intersection[best].eid);
+    if (IsLowPriority(best_data))
```
MoKob: this check could be done right after the min_element, possibly combining both into a dedicated template tool function and making `best_it` const?

``` 
(pseuodo code)
template<iterator_type,comparator,filter>
iterator_type minimum_valid_element(iterator_type begin,iterator_type end,comparator comp, filter fil)
{
     const auto res = std::min_element(begin,end,comp);
     return fil(*res) ? end : fil;
};
```

Also: eliminating `low_priority` completely, what would you do in any of the normal turn scenarios, where all roads are service.

Needs some additional test cases with service.

src/extractor/guidance/intersection_handler.cpp
```diff
@@ -400,108 +393,105 @@ std::size_t IntersectionHandler::findObviousTurn(const EdgeID via_edge,
...
+            best_deviation = angularDeviation(intersection[best].angle, STRAIGHT_ANGLE);
         }
     }
 
     // No non-low priority roads? Declare no obvious turn
     if (best == 0)
         return 0;
 
-    // best_continue is the first sorted way that shares the same name as the in way
-    const auto sameName = [&](const out_way &lhs) { return lhs.same_name_id == true; };
```
MoKob: this check probably breaks your tests now: see two comments above

src/extractor/guidance/intersection_handler.cpp
```diff
@@ -393,15 +393,18 @@ std::size_t IntersectionHandler::findObviousTurn(const EdgeID via_edge,
 
     const EdgeData &in_way = node_based_graph.GetEdgeData(via_edge);
 
+    const auto IsContinueRoad = [&](const EdgeData &way_data) {
+        return !util::guidance::requiresNameAnnounced(
+            in_way.name_id, way_data.name_id, name_table, street_name_suffix_table);
+    };
```
daniel-j-h: I'm doing the same over in my Sliproad Refactoring

https://github.com/Project-OSRM/osrm-backend/blob/eae226a9b1685ceba39589856afb24b266fbed9c/src/extractor/guidance/sliproad_handler.cpp#L589-L606

Maybe we should put it into the IntersectionHandler base class then we can both use it.

src/extractor/guidance/intersection_handler.cpp
```diff
@@ -383,107 +383,109 @@ std::size_t IntersectionHandler::findObviousTurn(const EdgeID via_edge,
...
+                                 lhs_same_or_higher_priority,
+                                 rhs_low_priority,
+                                 rhs_deviation,
+                                 lhs_same_classification);
+        auto right_tie = std::tie(rhs.entry_allowed,
+                                  rhs_same_or_higher_priority,
+                                  lhs_low_priority,
+                                  lhs_deviation,
+                                  rhs_same_classification);
         return left_tie > right_tie;
```
MoKob: what is really confusing me here is that you use `min_element` and still compare `>` here, this makes it really hard to compare what you want here.

Right now you get:

 - entry allowed before not entry allowed
 - same or higher priority before different priorities
 - all non-low priority roads before low priority roads
 - smaller deviation
 - all same deviations preferred from non same.

In the PR before you changed this, you did:

- entry allowed (due to the creation of out ways)
- deviation
- same classification (note that the find_if for same_or_higher (https://github.com/Project-OSRM/osrm-backend/pull/3217/commits/0cee3486e5de11ca2c2f191e6233392171f1317f#diff-b10c978a328e386ebdc0aa4db955388eL455) will always succeed on same_classification (contains priority)
- same or higher priority

if that did not succeed (since no same or higher priority), you selected the first one without low priority:
- entry allowed (due to the creation of out ways)
- deviation
- not low priority

While this still does not handle obviousness on service roads correctly (needs additional tests), the check worked for all other settings. Your new sorting criterion is different from your old one, as you can see from this analysis.

src/extractor/guidance/intersection_handler.cpp
karenzshea: https://github.com/Project-OSRM/osrm-backend/issues/3337

src/extractor/guidance/intersection_handler.cpp
```diff
@@ -383,201 +383,224 @@ std::size_t IntersectionHandler::findObviousTurn(const EdgeID via_edge,
...
+            best_option_data.road_classification)
     {
-        // if the best angle is going straight but the road is turning, we don't name anything
-        // obvious
         return 0;
     }
 
+    // get a count of number of ways from that intersection that qualify to have
+    // continue instruction because they share a name with the approaching way
+    std::int64_t continue_count =
```
MoKob: This (and below) could probably be `const auto`. I've done a small search, it seems not to be written anymore later on.

MoKob:
I have seen one tiny thing. Apart from this, the PR looks fine to me and improves readability a lot. Thank you for all the effort here.

features/guidance/low-priority.feature
```diff
@@ -0,0 +1,98 @@
...
+        And the ways
+            | nodes  | highway     | name    |
+            | abd    | residential | road    |
+            | eb     | service     | service |
+            | bc     | service     |         |
+
+       When I route I should get
+            | waypoints | route             | turns                       |
+            | e,c       | service,          | depart,arrive               |
+            | c,e       | ,service,service  | depart,turn straight,arrive |
```
MoKob: just for clarification: this is due to the name not being the same anymore? I'd probably still see the exact same shape here as obvious, but having this turn in here or not is much more a personal preference. So feel free to keep it this way.

Issue #3037, opened 2016-10-11
Trimmed geometry in match results


gardster:
Last segment coordinates of a matching plugin response are trimmed in the geometry section, but can be found in a steps section inside a detailed response.
For example:

```
https://router.project-osrm.org/match/v1/car/polyline(%7DjmwFz~ubMyCa@%60@yDJqE)?annotations=false&geometries=geojson&overview=simplified&radiuses=10;10;10;10&steps=true
```

Has geometry:

```
    "geometry": {
        "coordinates": [
          [
            -74.009755,
            40.705973
          ],
          [
            -74.009502,
            40.706379
          ],
          [
            -74.00929,
            40.706619
          ],
          [
            -74.008269,
            40.707722
          ],
          [
            -74.007459,
            40.707219
          ],
          [
            -74.006781,
            40.706724
          ],
          [
            -74.007452,
            40.706051
          ],
          [
            -74.008389,
            40.706598
          ],
          [
            -74.008996,
            40.706953
          ],
          [
            -74.008269,
            40.707722
          ],
          [
            -74.007459,
            40.707219
          ],
          [
            -74.006781,
            40.706724
          ],
          [
            -74.00713,
            40.706375
          ]
        ],
        "type": "LineString"
      }
```

But last waypoint is:

```
{
      "waypoint_index": 3,
      "location": [
        -74.007354,
        40.706518
      ],
      "name": "Cedar Street",
      "hint": "dyZGgP___38ULgIAHAAAAMIAAAAAAAAAAAAAAPysJQn6rCUJtOUAAMa8lvvWIW0CeryW-5IhbQIAAAEBrTIIxg==",
      "matchings_index": 0
    }
```

And there is a step with trimmed coordinates:

```
        {
              "intersections": [
                {
                  "out": 2,
                  "location": [
                    -74.00713,
                    40.706375
                  ],
                  "bearings": [
                    45,
                    210,
                    315
                  ],
                  "entry": [
                    false,
                    true,
                    true
                  ],
                  "in": 0
                }
              ],
              "geometry": {
                "coordinates": [
                  [
                    -74.00713,
                    40.706375
                  ],
                  [
                    -74.007354,
                    40.706518
                  ]
                ],
                "type": "LineString"
              },
              "maneuver": {
                "bearing_after": 310,
                "type": "turn",
                "modifier": "right",
                "bearing_before": 217,
                "location": [
                  -74.00713,
                  40.706375
                ]
              },
[...]

PR #3218, opened 2016-11-01
Fix omitting the last coordinate when simplifiying route

TheMarex:
# Issue

Addresses the first part of #3037.

## Tasklist
 - [x] review
 - [x] adjust for comments

### DIFF
--- a/src/engine/guidance/assemble_overview.cpp
+++ b/src/engine/guidance/assemble_overview.cpp
@@ -43,37 +43,11 @@ unsigned calculateOverviewZoomLevel(const std::vector<LegGeometry> &leg_geometri
     return util::viewport::getFittedZoom(south_west, north_east);
 }
 
-std::vector<util::Coordinate> simplifyGeometry(const std::vector<LegGeometry> &leg_geometries,
-                                               const unsigned zoom_level)
-{
-    std::vector<util::Coordinate> overview_geometry;
-    auto leg_index = 0UL;
-    for (const auto &geometry : leg_geometries)
-    {
-        auto simplified_geometry =
-            douglasPeucker(geometry.locations.begin(), geometry.locations.end(), zoom_level);
-        // not the last leg
-        if (leg_index < leg_geometries.size() - 1)
-        {
-            simplified_geometry.pop_back();
-        }
-        overview_geometry.insert(
-            overview_geometry.end(), simplified_geometry.begin(), simplified_geometry.end());
-    }
-    return overview_geometry;
-}
 }
 
 std::vector<util::Coordinate> assembleOverview(const std::vector<LegGeometry> &leg_geometries,
                                                const bool use_simplification)
 {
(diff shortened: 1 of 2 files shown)


src/engine/guidance/assemble_overview.cpp
```diff
@@ -85,16 +59,37 @@ std::vector<util::Coordinate> assembleOverview(const std::vector<LegGeometry> &l
...
+            auto end = simplified.cend();
+            insert_without_overlap(begin, end);
+        }
+    }
+    else
+    {
+        for (const auto &geometry : leg_geometries)
+        {
+            auto begin = geometry.locations.begin();
+            auto end = geometry.locations.end();
```
MoKob: why `.begin()` and `.end()` here, while using `.cbegin()`/`cend()` right above?

src/engine/guidance/assemble_overview.cpp
```diff
@@ -85,16 +59,37 @@ std::vector<util::Coordinate> assembleOverview(const std::vector<LegGeometry> &l
...
+            insert_without_overlap(begin, end);
```
MoKob: within this full if branch, we simply do the same for either simplified geometry or for the normal geometry, right?

I feel this could be written nice with `boost::adaptors` selecting either a `forwardingOperator` (for the cases which don't need to be simplified) or a `simplificationOperator` (for the simplification type).

like
```
auto simplify_if_desired = [](select operator and initialise);

boost::accumulate(boost::transformed(leg_geometries) | simplify_if_desired);
```

TheMarex: Problem with that is that `douglasPeucker` returns a rvalue and we would need to add a copy for `leg_geometries` to get the same interface (we can't move, since we are not allowed to modify `leg_geometries`).

I would prefer to keep the current interface since it is pretty obvious what it does and works well.

MoKob: I would prefer if we could find a way without code duplications, but nothing I insist on. 👍 if you want to keep it like this.

src/engine/guidance/assemble_overview.cpp
```diff
@@ -43,37 +43,11 @@ unsigned calculateOverviewZoomLevel(const std::vector<LegGeometry> &leg_geometri
...
                                                const bool use_simplification)
 {
-    if (use_simplification)
-    {
-        const auto zoom_level = std::min(18u, calculateOverviewZoomLevel(leg_geometries));
-        return simplifyGeometry(leg_geometries, zoom_level);
-    }
-    BOOST_ASSERT(!use_simplification);
-
     auto overview_size =
```
TheMarex: This only computes the upper bound which is still true after simplification. Before the code was never run for the simplified version, which is actually more of a bug (possibly multiple allocations even though we have a good idea of the size).

Issue #2485, opened 2016-05-28
Retrieve polylines with a precision of 6


Florian77711:
I have noticed that in the current HTTP API (release v5.1.0) the polyline returned by e.g. the route service has a precision of 1e5.
I'm still running a server (version 4.7.1) which provides the polyline with a precision of 6.
Is there a parameter or any chance to get the polyline returned with a precision of 1e6 in the new version, or is the precision 'lost' at the extract / contract process?

I haven't found anything in the Wiki, and after serching the repository, the line 15 of the following file seems to make the precision hardcoded:
https://github.com/Project-OSRM/osrm-backend/blob/807aa71a7c80b714b7a89db9c99c15720572ba25/include/engine/polyline_compressor.hpp

danpat:
@Florian77711 We changed the default precision to 5 to match the default precision in Google's `polyline.js` library.  Although it can be changed to 6, most people weren't, and we were fielding a constant stream of questions about it.  We decided to switch to 5 just to stop the flood of problems.

One way to approach this might be to support `polyline6` as a separate type.  I can understand the desire for 1e6 precision, but I can't think of many use-cases for anything beyond that (or less than 1e5).

@TheMarex thoughts?  This could simplify (and speed up) some of our internal API code as well....

TheMarex:
I don't see a problem with offering `polyline6` as well.

thrownblown:
For what its worth, changing the include/engine/polyline_compressor.hpp back to 1e6 was a trivial change.
I just finished rebuilding and reparsing my planet files and everything works as expected. If anyone else is having this issue here's my fix:
https://github.com/thrownblown/osrm-backend/commit/244f0b07810607b2cd2da6270bc8ac0b7a92e75b

danpat:
This landed with PR #3220 , a new `geometries=polyline6` option is available.

thrownblown:
Rad!

*Alex Farioletti*
*415.312.1674*
*tcbcourier.com <http://tcbcourier.com> *

PR #3220, opened 2016-11-01
polyline precision 6 positions

ghoshkaj:
# Issue

Retrieve polylines with a precision of 6 #2485

## Tasklist
 - [x] review
 - [x] adjust for comments

### DIFF
--- a/features/testbot/matching.feature
+++ b/features/testbot/matching.feature
@@ -177,7 +177,7 @@ Feature: Basic Map Matching
             | trace | matchings | geometry                                         |
             | efbc  | efbc      | 1,0.99964,1.000359,0.99964,1.000359,1,1.000718,1 |
 
-    Scenario: Testbot - Geometry details
+    Scenario: Testbot - Geometry details using geojson
         Given the query options
             | overview   | full     |
             | geometries | geojson  |
@@ -197,7 +197,47 @@ Feature: Basic Map Matching
             | trace | matchings | geometry                                   |
             | abd   | abd       | 1,1,1.000089,1,1.000089,1,1.000089,0.99991 |
 
-    Scenario: Testbot - Speed greater than speed threshhold, should split -- returns trace as abcd but should be split into ab,cd
+    Scenario: Testbot - Geometry details using polyline
+        Given the query options
+            | overview   | full      |
+            | geometries | polyline  |
+
+        Given the node map
+            """
+            a b c
+              d
+            """
+
+        And the ways
+            | nodes | oneway |
+            | abc   | no     |
+            | bd    | no     |
+
+        When I match I should get
+            | trace | matchings | geometry                                |
+            | abd   | abd       | 1,1,1,1.00009,1,1.00009,0.99991,1.00009 |
+
+    Scenario: Testbot - Geometry details using polyline6
+        Given the query options
+            | overview   | full       |
+            | geometries | polyline6  |
+
+        Given the node map
+            """
+            a b c
+              d
+            """
+
+        And the ways
+            | nodes | oneway |
+            | abc   | no     |
+            | bd    | no     |
+
+        When I match I should get
+            | trace | matchings | geometry                                   |
+            | abd   | abd       | 1,1,1,1.000089,1,1.000089,0.99991,1.000089 |
+
+    Scenario: Testbot - Speed greater than speed threshhold
         Given a grid size of 10 meters
         Given the query options
             | geometries | geojson  |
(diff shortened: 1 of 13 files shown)


include/engine/api/json_factory.hpp
```diff
@@ -42,9 +42,9 @@ std::string modeToString(const extractor::TravelMode mode);
 
 } // namespace detail
 
-template <typename ForwardIter> util::json::String makePolyline(ForwardIter begin, ForwardIter end)
+template <typename ForwardIter> util::json::String makePolyline(ForwardIter begin, ForwardIter end, double polylinePrecision = 1e5)
```
TheMarex: The polylinePrecision is a factor we do know at compile time (either 5 or 6 depending on the callsite). This function is rather performance critical, specially for long geometries. This leads me to suggest that we make it a template parameter like:

`template <typename ForwardIter, unsigned PRECISION=1e5>  makePolyline`.

include/engine/polyline_compressor.hpp
```diff
@@ -10,21 +10,15 @@ namespace osrm
...
-constexpr double POLYLINE_PRECISION = 1e5;
-constexpr double COORDINATE_TO_POLYLINE = POLYLINE_PRECISION / COORDINATE_PRECISION;
-constexpr double POLYLINE_TO_COORDINATE = COORDINATE_PRECISION / POLYLINE_PRECISION;
-}
 
 using CoordVectorForwardIter = std::vector<util::Coordinate>::const_iterator;
 // Encodes geometry into polyline format.
 // See: https://developers.google.com/maps/documentation/utilities/polylinealgorithm
-std::string encodePolyline(CoordVectorForwardIter begin, CoordVectorForwardIter end);
+std::string encodePolyline(CoordVectorForwardIter begin, CoordVectorForwardIter end, double polylinePrecision = 1e5);
```
TheMarex: Same goes for here, a template parameter would help the optimizer to pre-compute some arithmetic values for us. Example:
```
template<unsigned PRECISION=1e5>
std::string encodePolyline(CoordVectorForwardIter begin, CoordVectorForwardIter end)
{
   /* implementation goes here */
}
```

ghoshkaj:
@danpat success this works! 

I potentially still need to add tests in the following files:
osrm-backend/unit_tests/engine/geometry_string.cpp
osrm-backend/unit_tests/server/parameters_parser.cpp
osrm-backend/unit_tests/server/url_parser.cpp

Also, should I add the ability to decode polylines that have precision 6 as well? I've left decodePolyline alone. Not sure if it's something we need for the API to handle?

ghoshkaj:
@TheMarex the templating is done! \o/

include/engine/polyline_compressor.hpp
```diff
@@ -13,14 +15,42 @@ namespace engine
 namespace detail
 {
 constexpr double POLYLINE_PRECISION = 1e5;
```
danpat: This variable should be dead now ?

features/testbot/matching.feature
```diff
@@ -197,13 +197,50 @@ Feature: Basic Map Matching
             | trace | matchings | geometry                                   |
             | abd   | abd       | 1,1,1.000089,1,1.000089,1,1.000089,0.99991 |
 
-    Scenario: Testbot - Speed greater than speed threshhold, should split -- returns trace as abcd but should be split into ab,cd
-        Given a grid size of 10 meters
+    Scenario: Testbot - Geometry details in polyline
```
danpat: Hmm, looks like either this branch needs to be rebased on master, or you've accidentally removed some test code.

include/engine/polyline_compressor.hpp
ghoshkaj: This exists for use by the decodePolyline function. I haven't changed decode function to make it customizable. Should I? I couldn't figure out if I should because the only place that decodePolyline is used is here: https://github.com/Project-OSRM/osrm-backend/blob/master/include/server/api/base_parameters_grammar.hpp#L127 and I'm not sure what that does.

ghoshkaj: Okay, yeah, I'll just update the variable name and adjust the docs to say that we're only dealing with precision 6 for the encoding and not the decoding 😄

ghoshkaj:
@danpat okay here, I fixed my mess up. I think this is it. If you're okay with this and it turns green, then I'll merge it in. I've run the unit  tests, the linting and the cucumber tests locally and they all pass and behave as expected. Just waiting to see if Travis likes it.

danpat:
@ghoshkaj Can you update `docs/httpd.md` to mention the new response option?

danpat:
LGTM!

Issue #3180, opened 2016-10-26
Location from NodeEdge in profile.lua


fijemax:
Hello,

I need to access at  "location"  from each "node" in way_function in profile.lua script. We can't access since we're not using libosnmium's NodeLocationsForWays cache.
I would like to know if we could access again to location from a way in Lua script in a future version of OSRM ?

daniel-j-h:
See the comment here:

https://github.com/Project-OSRM/osrm-backend/blob/ceddfada3dfae23b04829ff78fb4fbe0d3629cca/src/extractor/scripting_environment_lua.cpp#L200-L203

OSRM was started before there even was `libosmium` and the extractor is using `libosmium` now manually reading chunks in parallel. If we want locations in there (which we _do_ need at some point for example for country awareness or general point-in-polygon tests without the need for external I/O) we have to plug in the location cache somehow.

This is definitely on the todo-list - if you want to have a go at it we're more than glad to help you there.

fijemax:
Hi,

Thank you for the answer. This confirms what I understood by reading the code. I try to plug a cache "NodeLocationsForWays" into LuaScriptingEnvironment without success for the moment.

daniel-j-h:
Here's the extractor's entry point:

https://github.com/Project-OSRM/osrm-backend/blob/01a2c6647251facf7fdd86d6801a8cd004570f12/src/extractor/extractor.cpp#L109

You have to plug in a location cache in there, filling locations in the Node processing step. Then register the appropriate functions in the scripting environment.

This is where the parsing happens in parallel:

https://github.com/Project-OSRM/osrm-backend/blob/5e167b8745954dece49dd7502a2f5d247d0d1c1e/src/extractor/scripting_environment_lua.cpp#L262

Because you first have to cache all locations and only then want to process ways I think you need to break the parallel for loop to no longer go go over all objects at once.

fijemax:
I try to plug in a cache but when i try to update the location on the nodeRef i have an exception who say : what():  id 3655224911 not found
I use a NodeLocationsForWays. It is filled juste after the LuaScriptingEnvironment creation. And passed as ref in LuaScriptingContext.

daniel-j-h:
The `osmium` tool already has a `add-locations-to-ways` command which reads a pbf caching all locations and then attaching those locations to the corresponding ways.

Running this on your `.pbf` file before using it with osrm _should_ allow you to access a way locations already without any changes needed.

[...]

PR #3223, opened 2016-11-02
Add static cache into scripting_environment.

fijemax:
#3180 

This is just a test, the code is a little dirty, sorry.
The cache is successfully filled. But when i try to access the static cache into the lua binding I have this exceotion : 
```
terminate called after throwing an instance of 'tbb::captured_exception'
  what():  id 3655224911 not found
```

### DIFF
--- a/include/extractor/scripting_environment.hpp
+++ b/include/extractor/scripting_environment.hpp
@@ -15,6 +15,11 @@
 #include <string>
 #include <vector>
 
+#include <osmium/index/map/sparse_file_array.hpp>
+#include <osmium/index/map/dense_file_array.hpp>
+#include <osmium/handler/node_locations_for_ways.hpp>
+#include <osmium/dynamic_handler.hpp>
+
 namespace osmium
 {
 class Node;
@@ -35,6 +40,13 @@ namespace extractor
 class RestrictionParser;
 struct ExtractionNode;
 struct ExtractionWay;
+
+using index_sparse_type = osmium::index::map::SparseFileArray<osmium::unsigned_object_id_type, osmium::Location>;
+using index_dense_type = osmium::index::map::DenseFileArray<osmium::unsigned_object_id_type, osmium::Location>;
+
+// The location handler always depends on the index type
+using location_handler_sparse_type = osmium::handler::NodeLocationsForWays<index_sparse_type>;
+using location_handler_dense_type = osmium::handler::NodeLocationsForWays<index_dense_type>;
 
 /**
  * Abstract class that handles processing osmium ways, nodes and relation objects by applying
@@ -65,8 +77,14 @@ class ScriptingEnvironment
                     tbb::concurrent_vector<std::pair<std::size_t, ExtractionWay>> &resulting_ways,
                     tbb::concurrent_vector<boost::optional<InputRestrictionContainer>>
                         &resulting_restrictions) = 0;
+
+    static location_handler_sparse_type *location_sparse_handler;
+    static location_handler_dense_type *location_dense_handler;
+    static void setSparseCache(location_handler_sparse_type &cache) {
+        location_sparse_handler = &cache;};
+    static void setDenseCache(location_handler_dense_type &cache) {
+        location_dense_handler = &cache;};
 };
 }
 }
-
 #endif /* SCRIPTING_ENVIRONMENT_HPP */
--- a/src/extractor/extractor.cpp
+++ b/src/extractor/extractor.cpp
@@ -14,7 +14,7 @@
 #include "util/name_table.hpp"
 #include "util/range_table.hpp"
 #include "util/simple_logger.hpp"
-#include "util/timing_util.hpp"
+
 
 #include "extractor/compressed_edge_container.hpp"
 #include "extractor/restriction_map.hpp"
@@ -59,6 +59,7 @@ namespace extractor
 
 namespace
 {
+
 std::tuple<std::vector<std::uint32_t>, std::vector<guidance::TurnLaneType::Mask>>
 transformTurnLaneMapIntoArrays(const guidance::LaneDescriptionMap &turn_lane_map)
 {
@@ -114,8 +115,43 @@ int Extractor::run(ScriptingEnvironment &scripting_environment)
     const unsigned recommended_num_threads = tbb::task_scheduler_init::default_num_threads();
     const auto number_of_threads = std::min(recommended_num_threads, config.requested_num_threads);
     tbb::task_scheduler_init init(number_of_threads);
-
     {
+        // Node Location Cache initialize
+        std::unique_ptr<index_sparse_type> index_sparse;
+        std::unique_ptr<index_dense_type> index_dense;
+        std::unique_ptr<location_handler_sparse_type> location_handler_sparse;
+        std::unique_ptr<location_handler_dense_type> location_handler_dense;
+
+        std::string cacheName(config.cache_file_path);
+        cacheName += "nodes.cache";
+        int fd = open(cacheName.c_str(), O_RDWR | O_CREAT, 0666);
+        if (fd == -1)
+        {
+            throw std::runtime_error(strerror(errno));
+        }
+
+        if("NONE" != config.cache_file_type)
+        {
+
+
+            if("SPARSE" == config.cache_file_type)
+            {
+                std::cout << "SPARSE" << std::endl;
+                index_sparse.reset(new index_sparse_type{fd});
+                location_handler_sparse.reset(new location_handler_sparse_type(*index_sparse.get()));
+                location_handler_sparse->ignore_errors();
+                scripting_environment.setSparseCache(*location_handler_sparse);
+            }
+            else
+            {
+                std::cout << "DENSE" << std::endl;
+                index_dense.reset(new index_dense_type{fd});
+                location_handler_dense.reset(new location_handler_dense_type(*index_dense.get()));
+                location_handler_dense->ignore_errors();
+                scripting_environment.setDenseCache(*location_handler_dense);
+            }
+        }
+
         util::SimpleLogger().Write() << "Input file: " << config.input_path.filename().string();
         if (!config.profile_path.empty())
         {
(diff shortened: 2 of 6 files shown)


include/extractor/scripting_environment_lua.hpp
```diff
@@ -13,12 +13,24 @@
...
+
 struct lua_State;
 
 namespace osrm
 {
 namespace extractor
 {
+  
+using index_neg_type = osmium::index::map::Dummy<osmium::unsigned_object_id_type, osmium::Location>;
+using index_type = osmium::index::map::SparseFileArray<osmium::unsigned_object_id_type, osmium::Location>;
```
daniel-j-h: Eventually we want a sparse and a dense cache as in

```cpp
using SparseIndex = osmium::index::map::SparseMemArray<osmium::unsigned_object_id_type, osmium::Location>;
using DenseIndex = osmium::index::map::DenseMemArray<osmium::unsigned_object_id_type, osmium::Location>;
```

and either let the user decide which one to use or dispatch based on the base map .pbf's size: dense for planet, sparse for everything considerably smaller.

include/extractor/scripting_environment_lua.hpp
```diff
@@ -33,6 +45,8 @@ struct LuaScriptingContext final
     bool has_node_function;
     bool has_way_function;
     bool has_segment_function;
+
+    location_handler_type * location_handler;
```
daniel-j-h: I'm not sure we want to store the location cache in the Lua scripting environment. From a conceptual standpoint, all scripting environments need the cache. Also we're currently storing the coordinates manually anyway in the extractor. A good place to put this therefore would be the extractor.

daniel-j-h: Here is where we store nodes during extraction in the stxxl containers and then serialize out to disk / discard:

- https://github.com/Project-OSRM/osrm-backend/blob/refactor-sliproad-handler/include/extractor/extraction_containers.hpp
- [ExternalMemoryNode](https://github.com/Project-OSRM/osrm-backend/blob/388d84a89ed2eff812e8f68bc4444f76e01abb20/include/extractor/external_memory_node.hpp)
- which is a [QueryNode](https://github.com/Project-OSRM/osrm-backend/blob/master/include/extractor/query_node.hpp)

(Related: let's rename QueryNode to something more semantically correct https://github.com/Project-OSRM/osrm-backend/issues/3238)

src/extractor/scripting_environment_lua.cpp
```diff
@@ -18,18 +19,70 @@
...
+    const osmium::io::File input_file(file_name);
+    osmium::io::Reader cacheReader(input_file, osmium::osm_entity_bits::node);
+    int fd = open("nodes.cache", O_RDWR | O_CREAT, 0666);
+    if (fd == -1)
+    {
+	throw std::runtime_error(strerror(errno));
+    }
+
+    index_type index{fd};
+    index_neg_type index_neg;
```
daniel-j-h: Your indices go out of scope after the `setCache` function returns. That's the reason once yo run the [extractor here](https://github.com/Project-OSRM/osrm-backend/pull/3223/files#diff-38493d025685d69c5109f547b722a4ebR164) your locations are no longer valid.

src/extractor/scripting_environment_lua.cpp
```diff
@@ -18,18 +19,70 @@
...
+
+    LuaScriptingEnvironment::location_handler = new location_handler_type(index, index_neg);
```
daniel-j-h: This leaks memory.

fijemax:
Hi @daniel-j-h  and thank you for the feedback.
The new version work correctly but i have some questions.
I feed the cache during the "extractor" extracts objects from buffer but i assume node are stored on the top of pbf, can I do this way ?
I do static cache pointer into the scripting_environment otherwise the LUA can't access to it into bind function.

daniel-j-h:
Yes, nodes come before ways in the osm pbf. That's the reason you can do

    osmium:apply(reader, locationCache, myHandler);

and get locations in the `myHandler`s and `way` callback.

As said, the extractor currently parses buffers manually and in parallel without taking the object type into account (for parsing, later on we dispatch on it and feed it to the appropriate `Process*` functions). You have to change that.

fijemax:
I am not sure to understand, I should change all the process so use the
` osmium:apply(reader, locationCache, myHandler);` function ?

or i should just move this code : 
 ```
              // Feed the cache, we assume the nodes are stored first, into data source.
              if(iter->type() == osmium::item_type::node)
                  location_handler.node(static_cast<const osmium::Node &>(*iter));
```
in extraction_containers callback, for feeding nodes into cache here ?

**Comment by daniel-j-h** [comment]
No we can't use the `apply` with a handler, since osmium parses in parallel but then serializes the results back before feeding them to the callback handler. That's the reason we're parsing buffers in parallel ourself, since we don't want the serialization constraints.

**Comment by fijemax** [comment]
Hi @daniel-j-h,
I refactored my code, and add cache options.
What do you think about the new cache solution ?
This version work fine for me.

File: include/extractor/scripting_environment.hpp [hunk 5]
```diff
@@ -36,6 +41,13 @@ class RestrictionParser;
 struct ExtractionNode;
 struct ExtractionWay;
 
+using index_sparse_type = osmium::index::map::SparseFileArray<osmium::unsigned_object_id_type, osmium::Location>;
+using index_dense_type = osmium::index::map::DenseFileArray<osmium::unsigned_object_id_type, osmium::Location>;
```
daniel-j-h: Is there a reason you want the cache to live on disk? I would use these instead:

```c++
using SparseIndex = osmium::index::map::SparseMemArray<osmium::unsigned_object_id_type, osmium::Location>;
using DenseIndex = osmium::index::map::DenseMemArray<osmium::unsigned_object_id_type, osmium::Location>;
```

See https://github.com/Project-OSRM/osrm-backend/pull/3223#discussion_r86320537

include/extractor/scripting_environment.hpp
```diff
@@ -65,8 +77,14 @@ class ScriptingEnvironment
                     tbb::concurrent_vector<std::pair<std::size_t, ExtractionWay>> &resulting_ways,
                     tbb::concurrent_vector<boost::optional<InputRestrictionContainer>>
                         &resulting_restrictions) = 0;
+
+    static location_handler_sparse_type *location_sparse_handler;
+    static location_handler_dense_type *location_dense_handler;
+    static void setSparseCache(location_handler_sparse_type &cache) {
+        location_sparse_handler = &cache;};
+    static void setDenseCache(location_handler_dense_type &cache) {
+        location_dense_handler = &cache;};
```
daniel-j-h: Why are the pointers static? Why do you set pointers here?

See https://github.com/Project-OSRM/osrm-backend/pull/3223#discussion_r86321880

src/extractor/extractor.cpp
```diff
@@ -114,8 +115,43 @@ int Extractor::run(ScriptingEnvironment &scripting_environment)
...
+            }
+            else
+            {
+                std::cout << "DENSE" << std::endl;
+                index_dense.reset(new index_dense_type{fd});
+                location_handler_dense.reset(new location_handler_dense_type(*index_dense.get()));
+                location_handler_dense->ignore_errors();
+                scripting_environment.setDenseCache(*location_handler_dense);
+            }
+        }
```
daniel-j-h: The indices should be default constructible, so there is no need for the unique ptr. Also it would be great to not having to parse strings here. These should be typed options set in the config.

    Index ind;

    if (config.use_sparse)
      ind = makeIndex();

src/extractor/scripting_environment_lua.cpp
```diff
@@ -202,6 +205,20 @@ void LuaScriptingEnvironment::InitContext(LuaScriptingContext &context)
...
+                     osmium::Location location = LuaScriptingEnvironment::location_sparse_handler->get_node_location(nref.ref());
+                     return location;
+                 }
+                 else if(location_sparse_handler)
+                 {
+                     osmium::Location location = LuaScriptingEnvironment::location_sparse_handler->get_node_location(nref.ref());
+                     return location;
+                 }
+                 else
+                     return nref.location();
```
daniel-j-h: This will always crash (no cache), right?

src/tools/extract.cpp
```diff
@@ -3,6 +3,7 @@
 #include "extractor/scripting_environment_lua.hpp"
 #include "util/simple_logger.hpp"
 #include "util/version.hpp"
+#include <../test/t/geom/test_tile_data.hpp>
```
daniel-j-h: Remove

src/extractor/scripting_environment_lua.cpp
fijemax: Yes, but what should I do when the LUA script tries to access the cache that is not generated?
I propose to catch location exception and propagate with complement information about how generate cache.

include/extractor/scripting_environment.hpp
fijemax: Cache is so big, what do SparseMerArray and DenseMemArray when cache is biger than ram memory ?

include/extractor/scripting_environment.hpp
fijemax: Cache is static because i can't access into luabind anonymous function ?
```
.def("location", +[](const osmium::NodeRef& nref){
                 if(location_sparse_handler)
                 {
                     osmium::Location location = LuaScriptingEnvironment::location_sparse_handler->get_node_location(nref.ref());
                     return location;
                 }
                 else if(location_sparse_handler)
                 {
                     osmium::Location location = LuaScriptingEnvironment::location_sparse_handler->get_node_location(nref.ref());
                     return location;
                 }
                 else'
                     return nref.location();
                     return location;
                 }
                 else
                     return nref.location();
})
```

include/extractor/scripting_environment.hpp
daniel-j-h: You let the underlying system take care of swapping (which heavily degrades performance).

include/extractor/scripting_environment.hpp
daniel-j-h: If you can't bind an anonymous lambda with a capture what about making a free standing private member function helper and register that with Luabind.

fijemax: How can access to extractor Cache into static helper function ?

fijemax: I don't know how to do if the cache isn't referenced by a static pointer somewhere.

daniel-j-h: You could do the following:
- store the cache in the extractor
- pass it into the scripting environment (which both fills it and also can use it)
- re-use the cache in the extraction containers where we're storing lon, lat coordinates anyway

The last point would make this pull request feasible to integrate since otherwise we duplicate the data which is _a lot_ on planet scale.

fijemax: Ok, i think i understand.
instead of create a libosmium cache I should use the 
`stxxl::vector<ExternalMemoryNode>;`
who store data into the extraction_containers.
And split the `for` loop reader into two loop, to parse in first time Node and after that other geometry.

daniel-j-h:
A slightly different behavior is getting explored in https://github.com/Project-OSRM/osrm-backend/pull/3316 by @fijemax - closing here.

Issue #2029, opened 2016-02-29
Clean libosrm from concepts and dependencies external to library's design


daniel-j-h:
In the strict sense libosrm does not need to know about responses being encoded as JSON or hints being encoded in Base64.

JSON is only needed for the HTTP use-case we have e.g. with node-osrm. Why not build node-osrm completely on top of libosrm? A libosrm user neither needs JSON nor the variant dependency it comes with.

The same goes for the hint. The requirement for Base64 encoding comes from the HTTP use-case and is not inherent in libosrm. Using libosrm I am able to store a Hint-typed object just fine.

Feel free to discuss.

TheMarex:
> JSON is only needed for the HTTP use-case we have e.g. with node-osrm. Why not build node-osrm completely on top of libosrm?

I don't know what you are getting here on. We need a way to represent the response. We either come up with a C++ structure that encapsulates our response, but in that case we then need to write an additional translation layer that transforms the C++ API structutre to the JSON API structure. This is unneeded overhead and we optimize for our use case - direct mapping to JS structures.

> The requirement for Base64 encoding comes from the HTTP use-case and is not inherent in libosrm.

Yes this is why `RouteParameters` contains `Hint` objects and not strings. If you want o move the serialization/deserialization outside of the `hint.hpp`  header feel free to move it to the server completely.

daniel-j-h:
Bumping this issue in light of our recent findings: assembling the JSON object / variant is heavy on the allocator. Quick fix would be to try a pool allocator or link in e.g. jemalloc if it's found on the user's system. What are our long-term plans for fixing this?

TheMarex:
@daniel-j-h I did an experiment with a custom `Array` type for coordinates (e.g. the geojson stuff which is a real bottle neck) that was basically `std::array<2, json::Value>` and didn't observe dramatic speedups. Which makes me think the actual heap allocation is not the main problem here. Though trying jemalloc is worth a try.

Long term I would like to explore the option of dropping the internal JSON for `libosrm` and expose the internal `RouteStep`, `RouteLeg` etc objects. Main motivation would be the possibility of exposing `std::vector<Coordinate>` for the geometry and not needing to handle format encoding in OSRM. This has some consequences on where certain code needs to live.

1. JSON serialization would need to live in `node-osrm`. If we want to keep the C++ version of `osrm-routed`, it needs to be duplicated there.

[...]

PR #3228, opened 2016-11-02
Don't expose JSON in libosrm but C++ object

TheMarex:
# Issue

Explores https://github.com/Project-OSRM/osrm-backend/issues/2029. Goal is to get some benchmarking numbers to get a compelling argument here.

## Tasklist
- [ ] Port a hacked node-osrm
- [ ] Run benchmarks


ssuluh:
+1, this will be great since my project using libosrm directly.

include/engine/guidance/result.hpp
```diff
@@ -0,0 +1,22 @@
...
+{
+namespace guidance
+{
+class Waypoint;
+class Route;
+
+struct Result
+{
+    std::vector<Waypoint> waypoints;
+    std::vector<Route> routes;
```
daniel-j-h: Just had a look over this changeset - what you're doing here is not allowed: you're storing incomplete types in stdlib containers. The container have to know the `value_type`'s size though.

include/engine/guidance/route.hpp
```diff
@@ -1,17 +1,30 @@
...
 namespace guidance
 {
 
+struct RouteLeg;
+
 struct Route
 {
     double duration;
     double distance;
+    std::vector<RouteLeg> legs;
```
daniel-j-h: Same here

TheMarex:
Outdated.

PR #3230, opened 2016-11-02
Fixes Cucumber's Unknown Node Error Reporting

daniel-j-h:
`Error(format('%s'), x)` => `Error(format('%s', x))` - Js does not care. We do.

Before: `unknown node %s`
After: `unknown node t`

https://nodejs.org/api/util.html#util_util_format_format_args


TheMarex:
Looks good to me. The failing travis test is one of our flukes and has nothing to do with this.

Issue #3234, opened 2016-11-02
Mason Integration Issues


daniel-j-h:
The following tests were made on completely fresh environments, e.g.

    docker run -it ubuntu:16.04 /bin/bash

##### Ubuntu 16.04

I expected these packages to be required for mason:

    apt-get install git cmake curl

but it seems like I also need

    libstdc++-5-dev
    pkg-config

with those, I get

```
CMake Error at third_party/mason/mason.cmake:187 (message):
  [Mason] Package at mason_packages/linux-x86_64/expat/2.1.1 has version '',
  but required '2.1.1'
```

##### Ubuntu 15.04

zlib seems to be missing as dependency

```
CMake Error at /usr/share/cmake-2.8/Modules/FindPackageHandleStandardArgs.cmake:108 (message):
  Could NOT find ZLIB (missing: ZLIB_LIBRARY ZLIB_INCLUDE_DIR)
```

with zlib from apt, the stdlib is too old so `libstdc++5-dev` is needed, too.

springmeyer:
Working on fixes in #3235. The problems were:

 - The `expat` package had an empty `mason.ini` so metadata could not be correctly queried. This is fixed by upgrading to a new `expat` package. Tracking the underlying issue at https://github.com/mapbox/mason/issues/274.

 - `pkg-config` is no longer required for mason integration after the `expat` fix. However it is required of course for the build of the example. But that has always been that way. Now `pkg-config` is no longer a prereq for `ENABLE_MASON`.

Other commenets:

 - Yes, `libstdc++-5-dev` is a requirement. We can loosen this if desirable by rebuilding boost, tbb, and luabind against an older libstdc++. Under discussion at https://github.com/mapbox/mason/issues/267.

 - `zlib` is depended upon via `find_package` in the same way as non mason builds. So this is by design right now and to ensure we link against the system zlib. This is precautionary and forward looking to allow the ENABLE_MASON built `libosrm.a` to be linked to the the node.js binaries provided by https://nodejs.org/en/download/current/ which also link system zlib. Said another way: It would be easy to link to a mason-provided zlib.a, but this might not work for node-osrm once we integrate it into osrm-backend. And most systems have zlib installed already, so this is a minor annoyance. On debian systems you'll need `apt-get install zlib1g-dev`.

I will address the libstdc++ and zlib needs in the docs: #3207

springmeyer:
Closing, solved by #3235

PR #3235, opened 2016-11-02
Mason improvements

springmeyer:
This makes improvements to the `-DENABLE_MASON` support.

 - Brings in fixed expat package
 - Checks for `curl` which is currently a required dependency of mason. Follow https://github.com/mapbox/mason/issues/274 for news of when it may not be needed.
 - Avoids calling `find_package(Osmium ...)` since this has the unwanted side effect of bringing in system libs for expat and bzip2 (when we need to link ENABLE_MASON builds against mason-provded expat and bzip2): https://github.com/Project-OSRM/osrm-backend/pull/3212#issuecomment-257990620

# Issue

Targets #3234 and https://github.com/Project-OSRM/osrm-backend/pull/3212#issuecomment-257990620.

## Tasklist
 - [x] review
 - [x] adjust for comments

## Requirements / Relations

Fixing the expat package requires a change to mason. The `update_dependencies.sh` depends on github releases to update deps. So we have a chicken n egg situation were we can't update mason until tagged. And mason can't be tagged until this PR is proven to work. So I've updated mason directly. Not sure yet how this will play with subtree merge.


springmeyer:
Working on travis. Ready for review. On travis their is one timeout failure that is unrelated to this PR.

TheMarex:
> So we have a chicken n egg situation were we can't update mason until tagged.

I don't see how tagging a new mason version depends on this PR?

> So I've updated mason directly. Not sure yet how this will play with subtree merge.

Please don't do that. Not going through the merge script causes some real head-aches every time we try that again down the read. My suggestion would be to do a mason release, revert this local change and then use the script.

CMakeLists.txt
```diff
@@ -400,6 +406,8 @@ if(ENABLE_MASON)
     set(ENV{CCACHE_CPP2} "true")
   endif()
 
+  include_directories(SYSTEM ${CMAKE_CURRENT_SOURCE_DIR}/third_party/libosmium/include)
```
TheMarex: Can we add a note here that `find_package` is not compatible with `mason_use` hence we need to manually define the include path here?

TheMarex:
@springmeyer is there anything that is blocking another mason release?

CMakeLists.txt
springmeyer: Done in https://github.com/Project-OSRM/osrm-backend/commit/4751b8af3c8680f3bb35e4c0d09d4d5170edd397

springmeyer:
> > So we have a chicken n egg situation were we can't update mason until tagged.

> I don't see how tagging a new mason version depends on this PR?

Sorry I was not very clear: My feeling is that it depends on this PR working. Now that it is working/passing, there is nothing blocking a mason release.

> > So I've updated mason directly. Not sure yet how this will play with subtree merge.

> Please don't do that. Not going through the merge script causes some real head-aches every time we try that again down the read. My suggestion would be to do a mason release, revert this local change and then use the script.

Will do 👍 

> @springmeyer is there anything that is blocking another mason release?

No. Will tag shortly and then update this PR.

springmeyer:
Adjustments have been made after review. This is ready to merge once green.

TheMarex:
Looks good to me. :+1:

PR #3237, opened 2016-11-03
Refactor/intersections

MoKob:
# Issue

Clean Up Intersections / add functionality

## Tasklist
 - [x] review
 - [x] adjust for comments

### DIFF
--- a/include/extractor/guidance/intersection.hpp
+++ b/include/extractor/guidance/intersection.hpp
@@ -6,6 +6,7 @@
 
 #include "extractor/guidance/turn_instruction.hpp"
 #include "util/guidance/toolkit.hpp"
+#include "util/node_based_graph.hpp"
 #include "util/typedefs.hpp" // EdgeID
 
 namespace osrm
@@ -18,7 +19,7 @@ namespace guidance
 // Every Turn Operation describes a way of switching onto a segment, indicated by an EdgeID. The
 // associated turn is described by an angle and an instruction that is used to announce it.
 // The Turn Operation indicates what is exposed to the outside of the turn analysis.
-struct TurnOperation final
+struct TurnOperation
 {
     EdgeID eid;
     double angle;
@@ -48,13 +49,23 @@ struct TurnOperation final
 // aaaaaaaa
 //
 // We would perceive a->c as a sharp turn, a->b as a slight turn, and b->c as a slight turn.
-struct ConnectedRoad final
+struct ConnectedRoad final : public TurnOperation
 {
+    using Base = TurnOperation;
+
     ConnectedRoad(const TurnOperation turn, const bool entry_allowed = false);
 
     // a turn may be relevant to good instructions, even if we cannot enter the road
     bool entry_allowed;
-    TurnOperation turn;
+
+    // used to sort the set of connected roads (we require sorting throughout turn handling)
+    bool compareByAngle(const ConnectedRoad &other) const;
+
+    // make a left turn into an equivalent right turn and vice versa
+    void mirror();
+
+    OSRM_ATTR_WARN_UNUSED
+    ConnectedRoad getMirroredCopy() const;
 };
 
 // small helper function to print the content of a connected road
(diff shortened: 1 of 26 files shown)


src/engine/guidance/assemble_overview.cpp
```diff
@@ -77,7 +77,8 @@ std::vector<util::Coordinate> assembleOverview(const std::vector<LegGeometry> &l
         const auto zoom_level = std::min(18u, calculateOverviewZoomLevel(leg_geometries));
         for (const auto &geometry : leg_geometries)
         {
-            const auto simplified = douglasPeucker(geometry.locations.begin(), geometry.locations.end(), zoom_level);
+            const auto simplified =
+                douglasPeucker(geometry.locations.begin(), geometry.locations.end(), zoom_level);
```
daniel-j-h: Same here, why is this re-formatting the source?

src/extractor/guidance/intersection.cpp
```diff
@@ -9,44 +16,97 @@ namespace guidance
...
+                                                                    DirectionModifier::SlightLeft,
+                                                                    DirectionModifier::Straight,
+                                                                    DirectionModifier::SlightRight,
+                                                                    DirectionModifier::Right,
+                                                                    DirectionModifier::SharpRight};
+
+    if (angularDeviation(angle, 0) > std::numeric_limits<double>::epsilon())
+    {
+        angle = 360 - angle;
+        instruction.direction_modifier = mirrored_modifiers[instruction.direction_modifier];
```
daniel-j-h: I know this is just moving some code around, but can we assert for out of bounds here?

src/extractor/guidance/intersection.cpp
```diff
@@ -9,44 +16,97 @@ namespace guidance
...
+            return util::guidance::angularDeviation(lhs.angle, angle) <
+                   util::guidance::angularDeviation(rhs.angle, angle);
+        });
+}
+
+bool Intersection::valid() const
+{
+    return !empty() &&
+           std::is_sorted(begin(), end(), std::mem_fn(&ConnectedRoad::compareByAngle)) &&
+           at(0).angle < std::numeric_limits<double>::epsilon();
```
daniel-j-h: This can be [0].angle

include/extractor/guidance/intersection.hpp
```diff
@@ -64,25 +75,24 @@ struct Intersection final : public std::vector<ConnectedRoad>
...
+    /*
+     * Check validity of the intersection object. We assume a few basic properties every set of
+     * connected roads should follow throughout guidance pre-processing. This utility function
+     * allows checking intersections for validity
+     */
+    bool valid() const;
+
+    // given all possible turns, which is the highest connected number of lanes per turn. This value
+    // is used, for example, during generation of intersections.
+    std::uint8_t getHighestConnectedLaneCount(const util::NodeBasedDynamicGraph &) const;
```
daniel-j-h: What about functionality for checking / getting the UTurn ConnectedRoad?

And / Or all roads except the UTurn road? Maybe

    auto ThinkAboutAGoodNameForNonUTurnRoadsBegin() { return begin() + 1; }
    auto ThinkAboutAGoodNameForNonUTurnRoadsEnd() { return end; }

with assert on !empty, const overloads, make_iterator_range overloads.

include/extractor/guidance/intersection.hpp
```diff
@@ -48,13 +49,23 @@ struct TurnOperation final
...
-    TurnOperation turn;
+
+    // used to sort the set of connected roads (we require sorting throughout turn handling)
+    bool compareByAngle(const ConnectedRoad &other) const;
+
+    // make a left turn into an equivalent right turn and vice versa
+    void mirror();
+
+    OSRM_ATTR_WARN_UNUSED
+    ConnectedRoad getMirroredCopy() const;
```
daniel-j-h: At the moment a ConnectedRoad does not know if it is the UTurn road or not - this is implicit in the parent Intersection ordering. Should there be a boolean flag here indicating this instead? Otherwise we always need the Intersection object when ConnectedRoads (think first, last iterators) would be good enough.

MoKob:
@karenzshea @chaupow heads up, this will most likely break your refactor prs. Intersections now don't offer the `.turn` anymore. You simply have to remove that part from any access you are doing on intersections and directly access the contained values. `.turn.angle` has to become `.angle`

daniel-j-h:
A tip here:

    git checkout --theirs path/to/conflicting/file

and then

    s/turn.angle/angle/g
    s/turn.eid/eid/g
    s/turn.intersction/intersection/g

makes rebasing less painfull.

PR #3246, opened 2016-11-04
add assertions / tests for elongation calls

MoKob:
# Issue

In investigating the already fixed segfault again, I added a few additional assertions that might help future debugging.

## Tasklist
 - [x] review
 - [x] adjust for comments

## Requirements / Relations

### DIFF
--- a/src/engine/guidance/post_processing.cpp
+++ b/src/engine/guidance/post_processing.cpp
@@ -477,6 +477,8 @@ void collapseUTurn(std::vector<RouteStep> &steps,
         invalidateStep(steps[step_index]);
         if (u_turn_with_name_change)
         {
+            BOOST_ASSERT_MSG(compatible(steps[one_back_index], steps[next_step_index]),
+                             "Compatibility should be transitive");
             steps[one_back_index] =
                 elongate(std::move(steps[one_back_index]), steps[next_step_index]);
             invalidateStep(steps[next_step_index]); // will be skipped due to the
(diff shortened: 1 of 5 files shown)


src/engine/guidance/post_processing.cpp
```diff
@@ -1620,8 +1624,11 @@ std::vector<RouteStep> collapseUseLane(std::vector<RouteStep> steps)
         if (step.maneuver.instruction.type == TurnType::UseLane && canCollapseUseLane(step))
         {
             const auto previous = getPreviousIndex(step_index, steps);
-            steps[previous] = elongate(std::move(steps[previous]), steps[step_index]);
-            invalidateStep(steps[step_index]);
+            if( compatible(steps[previous],step) )
```
daniel-j-h: clang-format

MoKob:
@daniel-j-h any further requests here?

src/engine/guidance/post_processing.cpp
```diff
@@ -1037,6 +1039,9 @@ std::vector<RouteStep> collapseTurns(std::vector<RouteStep> steps)
     // a series of turns is only possible to collapse if its only name changes and suppressed turns.
     const auto canCollapseAll = [&steps](std::size_t index, const std::size_t end_index) {
         BOOST_ASSERT(end_index <= steps.size());
+        if (!compatible(steps[index], steps[index + 1]))
+            return false;
+        ++index;
```
daniel-j-h: Wait wait wait - this not only adds a test but changes behavior. Why the + 1 here?

MoKob: Before we didn't check the very first `collapse` for compatibility.
What was tested before was for `a,b,c,d,e` you would check `b,c,d,e` since we are looping over all instructions unequal to `TurnType::Suppressed`.
Now we check `a,b` before doing the normal loop. This `+index` is compensated for by an additional `-1` for calling the function here.

MoKob: The respective call site: https://github.com/Project-OSRM/osrm-backend/pull/3246/files/31ae8492dbecf266d8d4413348f03805ac621ef1#diff-99adcb8bccf8e94a176bdf92509ff070L1139

Issue #3248, opened 2016-11-04
Turn data in debug tiles is incorrect


danpat:
There are several issues with the turn data in the debug tiles:

<img width="563" alt="screen shot 2016-11-04 at 2 08 51 pm" src="https://cloud.githubusercontent.com/assets/1892250/20020715/4d22d984-a298-11e6-8553-d9e44ccb4552.png">

  - some angles are off
  - many turns overlapping
  - several turns with a heading of 0

danpat:
Fixed by #3249

Issue #3064, opened 2016-10-15
Debug tiles aren't always fetching all visible geometry near tile edges


danpat:
Even with a query buffer, sometimes the debug tiles have results like this:

<img width="360" alt="screen shot 2016-10-14 at 9 34 54 pm" src="https://cloud.githubusercontent.com/assets/1892250/19407141/2f5eaaba-9256-11e6-9ef2-0c73c91dd56e.png">

Looks like there might be a bug in the rtree box search.  We seem to not be capturing all segments that intersect with our tile bounding box.

danpat:
Possibly related issue:  turns are showing up in Z17 tile, but not on Z18 tile:

Z17:
<img width="383" alt="screen shot 2016-10-14 at 10 40 41 pm" src="https://cloud.githubusercontent.com/assets/1892250/19407432/5bcd3aa4-925f-11e6-95b2-ada222d74ab4.png">

Z18:
<img width="387" alt="screen shot 2016-10-14 at 10 40 33 pm" src="https://cloud.githubusercontent.com/assets/1892250/19407433/5e3adb70-925f-11e6-9939-555d7b59d593.png">

danpat:
Ah, a clue.  I dumped some of the RTree bounding box search coordinates, this is what comes back:

<img width="670" alt="screen shot 2016-11-04 at 4 08 43 pm" src="https://cloud.githubusercontent.com/assets/1892250/20023995/13bc6ef6-a2a9-11e6-840e-04b1f558fd47.png">

We have a nice overlap on the X axis, but a gap on the Y.  That gap is causing the missing bits of geometry.

danpat:
Fixed by #3249

PR #3249, opened 2016-11-04
Fix debug tile turn data

danpat:
This PR addresses two tickets:

https://github.com/Project-OSRM/osrm-backend/issues/3248 - turn angles are wrong, and some turns aren't showing up.

This was caused by changes to the geometry encoding - the plugin to be updated to find the reverse geometry properly in the new GetForward/GetReverse world.

https://github.com/Project-OSRM/osrm-backend/issues/3064 - losing geometry near tile edges.  This was caused by a basic math error when adding a buffer to the bounding box - I was effectively shrinking the bounding box vertically, rather than growing it.

With this PR, we now get:
screen shot 2016-11-04 at 2 54 46 pm

## Tasklist
 - [x] review
 - [x] adjust for comments

### DIFF
--- a/include/util/web_mercator.hpp
+++ b/include/util/web_mercator.hpp
@@ -141,12 +141,13 @@ inline FloatCoordinate toWGS84(const FloatCoordinate &mercator_coordinate)
 
 // Converts a WMS tile coordinate (z,x,y) into a wgs bounding box
 inline void xyzToWGS84(
-    const int x, const int y, const int z, double &minx, double &miny, double &maxx, double &maxy)
+    const int x, const int y, const int z, double &minx, double &miny, double &maxx, double &maxy,
+    int mercator_buffer = 0)
 {
-    minx = x * TILE_SIZE;
-    miny = (y + 1.0) * TILE_SIZE;
-    maxx = (x + 1.0) * TILE_SIZE;
-    maxy = y * TILE_SIZE;
+    minx = x * TILE_SIZE - mercator_buffer;
+    miny = (y + 1.0) * TILE_SIZE - mercator_buffer;
+    maxx = (x + 1.0) * TILE_SIZE + mercator_buffer;
+    maxy = y * TILE_SIZE + mercator_buffer;
     // 2^z * TILE_SIZE
     const double shift = (1u << static_cast<unsigned>(z)) * TILE_SIZE;
     pixelToDegree(shift, minx, miny);
--- a/include/util/web_mercator.hpp
+++ b/include/util/web_mercator.hpp
@@ -140,14 +140,19 @@ inline FloatCoordinate toWGS84(const FloatCoordinate &mercator_coordinate)
 }
 
 // Converts a WMS tile coordinate (z,x,y) into a wgs bounding box
-inline void xyzToWGS84(
-    const int x, const int y, const int z, double &minx, double &miny, double &maxx, double &maxy,
-    int mercator_buffer = 0)
+inline void xyzToWGS84(const int x,
+                       const int y,
+                       const int z,
+                       double &minx,
+                       double &miny,
+                       double &maxx,
+                       double &maxy,
+                       int mercator_buffer = 0)
 {
     minx = x * TILE_SIZE - mercator_buffer;
-    miny = (y + 1.0) * TILE_SIZE - mercator_buffer;
+    miny = (y + 1.0) * TILE_SIZE + mercator_buffer;
     maxx = (x + 1.0) * TILE_SIZE + mercator_buffer;
-    maxy = y * TILE_SIZE + mercator_buffer;
+    maxy = y * TILE_SIZE - mercator_buffer;
     // 2^z * TILE_SIZE
     const double shift = (1u << static_cast<unsigned>(z)) * TILE_SIZE;
     pixelToDegree(shift, minx, miny);
--- a/src/engine/plugins/tile.cpp
+++ b/src/engine/plugins/tile.cpp
@@ -387,259 +387,218 @@ Status TilePlugin::HandleRequest(const std::shared_ptr<datafacade::BaseDataFacad
         // are all pointed in the same direction.
         struct EdgeBasedNodeInfo
         {
-            unsigned source_intersection; // node-based-node ID
-            unsigned target_intersection; // node-based-node ID
-            bool is_geometry_forward;     // Is the geometry forward or reverse?
+            bool is_geometry_forward; // Is the geometry forward or reverse?
             unsigned packed_geometry_id;
         };
         // Lookup table for edge-based-nodes
         std::unordered_map<NodeID, EdgeBasedNodeInfo> edge_based_node_info;
-        std::unordered_map<NodeID, std::vector<EdgeID>> outgoing_edges;
-        std::unordered_map<NodeID, std::vector<EdgeID>> incoming_edges;
 
-        // Now, loop over all the road segments we saw, and build up a mini
-        // graph for just the network that's visible.
+        struct SegmentData
+        {
+            NodeID target_node;
+            EdgeID edge_based_node_id;
+        };
+
+        std::unordered_map<NodeID, std::vector<SegmentData>> directed_graph;
+
+        // Build an adjacency list for all the road segments visible in
+        // the tile
         for (const auto &edge : edges)
         {
-            // Note: edge.u is the node-based node ID of the source intersection
-            //       edge.v is the node-based node ID of the target intersection
-            //       both these values can be directly looked up for coordinates
-
-            // If forward travel is enabled on this road section, and we haven't seen this
-            // edge-based-node
-            // before
-            if (edge.forward_segment_id.enabled &&
-                edge_based_node_info.count(edge.forward_segment_id.id) == 0)
+            if (edge.forward_segment_id.enabled)
             {
-                // Add this edge-based-nodeid as an outgoing from the source intersection
-                const auto outgoing_itr_u = outgoing_edges.find(edge.u);
-                if (outgoing_itr_u != outgoing_edges.end())
+                if (directed_graph.count(edge.u) == 0)
                 {
-                    outgoing_itr_u->second.push_back(edge.forward_segment_id.id);
+                    directed_graph[edge.u] = {{edge.v, edge.forward_segment_id.id}};
                 }
                 else
                 {
-                    outgoing_edges[edge.u] = {edge.forward_segment_id.id};
+                    directed_graph[edge.u].push_back({edge.v, edge.forward_segment_id.id});
                 }
-
-                // Add this edge-based-nodeid as an incoming to the target intersection
-                const auto incoming_itr_v = incoming_edges.find(edge.v);
-                if (incoming_itr_v != incoming_edges.end())
-                {
-                    incoming_itr_v->second.push_back(edge.forward_segment_id.id);
-                }
-                else
-                {
-                    incoming_edges[edge.v] = {edge.forward_segment_id.id};
-                }
-
-                edge_based_node_info[edge.forward_segment_id.id] = {
-                    edge.u, edge.v, true, edge.packed_geometry_id};
+                edge_based_node_info[edge.forward_segment_id.id] = {true, edge.packed_geometry_id};
             }
-            // Same as previous block, but everything flipped
-            if (edge.reverse_segment_id.enabled &&
-                edge_based_node_info.count(edge.reverse_segment_id.id) == 0)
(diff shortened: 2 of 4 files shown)


include/util/web_mercator.hpp
```diff
@@ -140,13 +140,19 @@ inline FloatCoordinate toWGS84(const FloatCoordinate &mercator_coordinate)
...
 
 // Converts a WMS tile coordinate (z,x,y) into a wgs bounding box
-inline void xyzToWGS84(
-    const int x, const int y, const int z, double &minx, double &miny, double &maxx, double &maxy)
-{
-    minx = x * TILE_SIZE;
-    miny = (y + 1.0) * TILE_SIZE;
-    maxx = (x + 1.0) * TILE_SIZE;
-    maxy = y * TILE_SIZE;
+inline void xyzToWGS84(const int x,
```
MoKob: wasn't this formatted or are you using a different version than the rest?
At least this looks like it is only caused by the additional mercator_buffer. Might be all correct, though.

src/engine/plugins/tile.cpp
```diff
@@ -380,226 +387,218 @@ Status TilePlugin::HandleRequest(const std::shared_ptr<datafacade::BaseDataFacad
...
                 {
-                    outgoing_edges[edge.u] = {edge.forward_segment_id.id};
-                }
-
-                // Add this edge-based-nodeid as an incoming to the target intersection
-                const auto incoming_itr_v = incoming_edges.find(edge.v);
-                if (incoming_itr_v != incoming_edges.end())
-                {
-                    incoming_itr_v->second.push_back(edge.forward_segment_id.id);
+                    directed_graph[edge.u] = {{edge.v, edge.forward_segment_id.id}};
```
MoKob: is this distinction `if/else` necessary? The `[]` operator should create the initial vector on access.

src/engine/plugins/tile.cpp
```diff
@@ -380,226 +387,218 @@ Status TilePlugin::HandleRequest(const std::shared_ptr<datafacade::BaseDataFacad
...
                 }
                 else
                 {
-                    incoming_edges[edge.v] = {edge.forward_segment_id.id};
+                    directed_graph[edge.u].push_back({edge.v, edge.forward_segment_id.id});
                 }
-
-                edge_based_node_info[edge.forward_segment_id.id] = {
-                    edge.u, edge.v, edge.packed_geometry_id};
+                edge_based_node_info[edge.forward_segment_id.id] = {true, edge.packed_geometry_id};
```
MoKob: any chance this can overwrite different information?

src/engine/plugins/tile.cpp
```diff
@@ -380,226 +387,218 @@ Status TilePlugin::HandleRequest(const std::shared_ptr<datafacade::BaseDataFacad
...
                 {
-                    incoming_edges[edge.u] = {edge.reverse_segment_id.id};
+                    directed_graph[edge.v].push_back({edge.u, edge.reverse_segment_id.id});
                 }
-
-                // Save info about this edge-based-node, note reversal from forward
-                // block above.
-                edge_based_node_info[edge.reverse_segment_id.id] = {
-                    edge.v, edge.u, edge.packed_geometry_id};
+                edge_based_node_info[edge.reverse_segment_id.id] = {false, edge.packed_geometry_id};
```
MoKob: same here, any chance this can interact badly with above?

src/engine/plugins/tile.cpp
```diff
@@ -380,226 +387,218 @@ Status TilePlugin::HandleRequest(const std::shared_ptr<datafacade::BaseDataFacad
...
+        //   Look at the outgoing edges from A
+        //   If the outgoing edge has a different edge_based_node_id
+        //     This is a turn.  Find it in the CH
+        //     Get the edge data
+        //     Subtract the road weights from the edge weight
+        //     Calculate angles, add to tile.
         std::vector<contractor::QueryEdge::EdgeData> unpacked_shortcut;
-        std::vector<EdgeWeight> forward_weight_vector;
-        for (const auto &source_ebn : edge_based_node_info)
+        std::vector<EdgeWeight> first_weight_vector;
```
MoKob: is there a way we can choose more meaningful names here? firstnode/firstedge seem a bit strange for me, if we actually loop over many of them.

src/engine/plugins/tile.cpp
```diff
@@ -380,226 +387,218 @@ Status TilePlugin::HandleRequest(const std::shared_ptr<datafacade::BaseDataFacad
...
+                                                &unpacked_shortcut]() {
+                                const auto inner_data = facade->GetEdgeData(smaller_edge_id);
+                                if (inner_data.shortcut)
+                                {
+                                    unpacked_shortcut.clear();
+                                    UnpackEdgeToEdges(*facade,
+                                                      firstedge.edge_based_node_id,
+                                                      secondedge.edge_based_node_id,
+                                                      unpacked_shortcut);
+                                    return unpacked_shortcut.front();
```
MoKob: if we only care about the very first node, it might give some performance if we implement a `left sided` unpack that does not recurse in both directions. Should switch from `O(n)` to `O(log(n))` on average, assuming a nice and balanced shortcut creation. (Worst case will stay `O(n)`)

src/engine/plugins/tile.cpp
```diff
@@ -380,226 +387,218 @@ Status TilePlugin::HandleRequest(const std::shared_ptr<datafacade::BaseDataFacad
...
+                            }
+
+                            // Add the turn angle value to the value lookup table for the vector
+                            // tile.
+                            const auto turn_angle_index = use_point_int_value(turn_angle);
+                            // And, same for the actual turn cost value - it goes in the lookup
+                            // table,
+                            // not directly on the feature itself.
+                            const auto turn_cost_index = use_point_float_value(
+                                turn_cost / 10.0); // Note conversion to float here
```
MoKob: nitpicking: conversion to double, then restriction to float ;)

src/engine/plugins/tile.cpp
```diff
@@ -380,226 +387,218 @@ Status TilePlugin::HandleRequest(const std::shared_ptr<datafacade::BaseDataFacad
...
-                    all_turn_data.emplace_back(
-                        TurnData{coord_via, angle_in_index, turn_angle_index, turn_cost_index});
                 }
             }
         }
     }
 
-    // Vector tiles encode feature properties as indexes into a lookup table.  So, we need to
+    // Vector tiles encode feature properties as indexes into a lookup table.  So, we need
+    // to
```
MoKob: formatting issue? Definitely very ugly breaks here

src/engine/plugins/tile.cpp
```diff
@@ -703,8 +709,10 @@ Status TilePlugin::HandleRequest(const std::shared_ptr<datafacade::BaseDataFacad
                                                                     const std::size_t name_idx,
                                                                     std::int32_t &start_x,
                                                                     std::int32_t &start_y) {
-                        // Here, we save the two attributes for our feature: the speed and the
-                        // is_small boolean.  We only serve up speeds from 0-139, so all we do is
+                        // Here, we save the two attributes for our feature: the speed and
+                        // the
```
MoKob: formatting, see above

src/engine/plugins/tile.cpp
```diff
@@ -800,7 +808,8 @@ Status TilePlugin::HandleRequest(const std::shared_ptr<datafacade::BaseDataFacad
             }
 
             // Field id 3 is the "keys" attribute
-            // We need two "key" fields, these are referred to with 0 and 1 (their array indexes)
+            // We need two "key" fields, these are referred to with 0 and 1 (their array
+            // indexes)
```
MoKob: formatting

src/engine/plugins/tile.cpp
```diff
@@ -380,226 +387,218 @@ Status TilePlugin::HandleRequest(const std::shared_ptr<datafacade::BaseDataFacad
...
 
-        // Now, loop over all the road segments we saw, and build up a mini
-        // graph for just the network that's visible.
+        struct SegmentData
+        {
+            NodeID target_node;
+            EdgeID edge_based_node_id;
+        };
+
+        std::unordered_map<NodeID, std::vector<SegmentData>> directed_graph;
```
daniel-j-h: Possible to .reserve here (potentially optimistically)?

src/engine/plugins/tile.cpp
```diff
@@ -380,226 +387,218 @@ Status TilePlugin::HandleRequest(const std::shared_ptr<datafacade::BaseDataFacad
...
+                            // table,
+                            // not directly on the feature itself.
+                            const auto turn_cost_index = use_point_float_value(
+                                turn_cost / 10.0); // Note conversion to float here
+
+                            // Save everything we need to later add all the points to the tile.
+                            // We need the coordinate of the intersection, the angle in, the turn
+                            // angle and the turn cost.
+                            all_turn_data.emplace_back(TurnData{
+                                coord_via, angle_in_index, turn_angle_index, turn_cost_index});
```
daniel-j-h: `all_turn_data` contains `TurnData`, correct? If so, `emplace_back` should perfectly forward arguments to its ctor.

Either change the `emplace_back` with the `&&`-taking `push_back` overload here or remove the explicit `TurnData`.

include/util/web_mercator.hpp
danpat: Either it wasn't formatted, or the additional parameter added enough width that it pushed it into the vertical layout mode.  Formatting was done with scripts/format.sh as a final pass.

src/engine/plugins/tile.cpp
```diff
@@ -400,206 +379,197 @@ Status TilePlugin::HandleRequest(const std::shared_ptr<datafacade::BaseDataFacad
...
         {
             if (edge.forward_segment_id.enabled)
             {
-                if (directed_graph.count(edge.u) == 0)
+                // operator[] will construct an empty vector at [edge.u] if there is no value.
+                directed_graph[edge.u].push_back({edge.v, edge.forward_segment_id.id});
+                if (edge_based_node_info.count(edge.forward_segment_id.id) == 0)
                 {
-                    directed_graph[edge.u] = {{edge.v, edge.forward_segment_id.id}};
+                    edge_based_node_info[edge.forward_segment_id.id] = {true,
```
MoKob: isn't this if-switch just the same as above (construction on access)?

src/engine/plugins/tile.cpp
```diff
@@ -400,206 +379,197 @@ Status TilePlugin::HandleRequest(const std::shared_ptr<datafacade::BaseDataFacad
...
+                        edge_based_node_info[edge.forward_segment_id.id].packed_geometry_id ==
+                        edge.packed_geometry_id);
                 }
-                edge_based_node_info[edge.forward_segment_id.id] = {true, edge.packed_geometry_id};
             }
             if (edge.reverse_segment_id.enabled)
             {
-                if (directed_graph.count(edge.v) == 0)
+                directed_graph[edge.v].push_back({edge.u, edge.reverse_segment_id.id});
+                if (edge_based_node_info.count(edge.reverse_segment_id.id) == 0)
```
MoKob: same here, check for count is unneccessary, if reverse/forward segments are unique. This repeats my question from the last review if these can ever interact?

src/engine/plugins/tile.cpp
```diff
@@ -400,206 +379,197 @@ Status TilePlugin::HandleRequest(const std::shared_ptr<datafacade::BaseDataFacad
...
+                        const auto node_from = startnode.first;
+                        const auto node_via = approachedge.target_node;
+                        const auto node_to = exit_edge.target_node;
+
+                        const auto coord_from = facade->GetCoordinateOfNode(node_from);
+                        const auto coord_via = facade->GetCoordinateOfNode(node_via);
+                        const auto coord_to = facade->GetCoordinateOfNode(node_to);
+
+                        // Calculate the bearing that we approach the intersection at
+                        const auto angle_in = static_cast<int>(
```
MoKob: just to be clear: this will be out of sync with the coordinates used in the data. Why don't you use bearing-classes and turn-angles from the edges?
(compare route step generation)

danpat: These turn angles are for map rendering - they need to align with the exact visual geometry so that icons can be placed next to the lines on the map.

Exposing the guidance angles as part of the tiles could be useful, but I think we should do that later if we decide we need them.  I think if we tried to use them for a visual layer, it might be difficult to determine which turn arrows

src/engine/plugins/tile.cpp
danpat: Done.

Issue #3251, opened 2016-11-05
URL decoder does not decode percent-encoded characters


TheMarex:
Our `URLParser` does not actually first decoded `%HEX` encoded characters before doing the real parsing. This for example breaks polylines like in this request:

```
https://router.project-osrm.org/match/v1/car/polyline(%7DjmwFz~ubMyCa@%60@yDJqE)?annotations=false&geometries=geojson&overview=simplified&radiuses=10;10;10;10&steps=true
```

(notice `%7D` and `%60`).

Second part of https://github.com/Project-OSRM/osrm-backend/issues/3037

daniel-j-h:
Isn't this a matter of bringing in node-osrm and then switching to a proper dev HTTPD?

Or do you want to implement a fully fledged [url-encoding](https://en.wikipedia.org/wiki/Percent-encoding) decoder in C++ which will be ripped out in the coming month anyway?

TheMarex:
Maybe we can just bundle something like https://github.com/larroy/urlcpp which should get us there quite fast.

TheMarex:
Addressed by #3252. Thanks @oxidase

PR #3252, opened 2016-11-06
Added percent-encoding to URL parser

oxidase:
# Issue

Added percent-encoding to URL parser, so _query_ part of parse accepts %HEX and converts the triple to char value.

## Tasklist
 - [x] review
 - [ ] adjust for comments

## Requirements / Relations
Issue #3251
https://tools.ietf.org/html/rfc1738


daniel-j-h:
Looks good :+1: - the question from the issue remains:
I think we should replace the dev server asap with a httpd going through the node bindings.

Issue #3254, opened 2016-11-07
No Roundabout Instructions for Routing over junction=roundabout


daniel-j-h:
It seems like our Roundabout handling is broken for more or less all roundabouts I just sampled:

![r0](https://cloud.githubusercontent.com/assets/527241/20060233/71fc7d14-a4fa-11e6-9ef5-c3407f9ff0ba.png)

![r1](https://cloud.githubusercontent.com/assets/527241/20060231/71f7717a-a4fa-11e6-9d76-410286df7903.png)

![r2](https://cloud.githubusercontent.com/assets/527241/20060232/71f84adc-a4fa-11e6-8452-3c6e118041ec.png)

- http://map.project-osrm.org/?z=17&center=52.513084%2C13.322436&loc=52.512826%2C13.324378&loc=52.511174%2C13.322918&hl=en&alt=0
- http://map.project-osrm.org/?z=17&center=52.513090%2C13.322436&loc=52.512826%2C13.324378&loc=52.512349%2C13.317361&hl=en&alt=0
- http://map.project-osrm.org/?z=17&center=52.513090%2C13.322436&loc=52.512826%2C13.324378&loc=52.514236%2C13.316728&hl=en&alt=0

[The `junction=roundabout` tag is on the way.](http://www.openstreetmap.org/way/139880220#map=19/52.51255/13.32209&layers=D)

Example request without roundabout instructions:

    http 'http://router.project-osrm.org/route/v1/driving/13.324378,52.512826;13.322918,52.511174?steps=true'

Here is a simple scenario:

- http://map.project-osrm.org/?z=17&center=52.514886%2C13.274285&loc=52.517272%2C13.271645&loc=52.515725%2C13.273104&hl=en&alt=0

Also try starting and ending in this roundabout.

daniel-j-h:
This was resolved by https://github.com/Project-OSRM/osrm-backend/pull/3255.

PR #3255, opened 2016-11-07
fix collapsing of roundabout instructions

MoKob:
# Issue

https://github.com/Project-OSRM/osrm-backend/issues/3254

## Tasklist
 - [x] review
 - [x] adjust for comments

## Requirements / Relations
none

### DIFF
--- a/src/engine/guidance/post_processing.cpp
+++ b/src/engine/guidance/post_processing.cpp
@@ -498,6 +498,10 @@ void collapseTurnAt(std::vector<RouteStep> &steps,
     BOOST_ASSERT(one_back_index < steps.size());
     const auto &current_step = steps[step_index];
     const auto &one_back_step = steps[one_back_index];
+    // Don't collapse roundabouts
+    if (entersRoundabout(current_step.maneuver.instruction) ||
+        entersRoundabout(one_back_step.maneuver.instruction))
+        return;
 
     // This function assumes driving on the right hand side of the streat
     BOOST_ASSERT(!one_back_step.intersections.empty() && !current_step.intersections.empty());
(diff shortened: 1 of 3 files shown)


src/engine/guidance/post_processing.cpp
```diff
@@ -812,6 +816,11 @@ void collapseTurnAt(std::vector<RouteStep> &steps,
 //
 bool isStaggeredIntersection(const RouteStep &previous, const RouteStep &current)
 {
+    //don't touch roundabouts
+    if (entersRoundabout(previous.maneuver.instruction) ||
+        entersRoundabout(current.maneuver.instruction))
+        return false;
+
```
daniel-j-h: I think we should completely disable the Staggered Intersection logic within roundabouts (or what about mode changes, or road classes for which it makes no sense, .. ?) - it really only targets the very short zig-zags in rural areas as described in https://github.com/Project-OSRM/osrm-backend/pull/2738. What's your take on this? (Not blocking this PR, though)

src/engine/guidance/post_processing.cpp
```diff
@@ -498,6 +498,10 @@ void collapseTurnAt(std::vector<RouteStep> &steps,
     BOOST_ASSERT(one_back_index < steps.size());
     const auto &current_step = steps[step_index];
     const auto &one_back_step = steps[one_back_index];
+    // Don't collapse roundabouts
+    if (entersRoundabout(current_step.maneuver.instruction) ||
+        entersRoundabout(one_back_step.maneuver.instruction))
+        return;
```
daniel-j-h: Hmm do you know where we introduced the regression and why our tests did not catch it? Also why only the roundabout enter instruction - are the silent roundabout instructions already collapsed into the multi-hop roundabout instruction here?

MoKob: No test so far checked for this situation. So our tests couldn't catch it. It's likely that the regression came in with the changes of turn angles, since I modified postprocessing a bit as well.

What are you calling a `multi-hop roundabout instruction`? During turn processing, we only have `enter roundabout` left. All other roundabout instructions should have been removed by now.

src/engine/guidance/post_processing.cpp
MoKob: The check is doing exactly that, it disables the staggered intersections for roundabouts? It makes sure that wherever it is checked, staggered intersections don't interact with roundabouts. So what's your question here?

src/engine/guidance/post_processing.cpp
daniel-j-h: > What are you calling a multi-hop roundabout instruction? During turn processing, we only have enter roundabout left. All other roundabout instructions should have been removed by now.

Yep that's what I meant. Thanks.

MoKob:
Failing tests are the ticketed timeouts again :(

PR #3262, opened 2016-11-08
Remove {highway: type} fallback from bike / walk profile, fixes #3231

### LINKED ISSUE #3231: Remove {highway: type} fallback from default bike / walk profile
See https://github.com/Project-OSRM/osrm-text-instructions/issues/62: there is no reason to ever return an encoding not intended for humans in the response. We removed it from the car profile long ago, and although at the moment we focus on car routing, bike and walk should return reasonable responses.

- [x] remove https://github.com/Project-OSRM/osrm-backend/blob/4ba8ccfcce50c1d1c6e91df34b85ecc98a116ec2/profiles/foot.lua#L157
- [x] remove https://github.com/Project-OSRM/osrm-backend/blob/4ba8ccfcce50c1d1c6e91df34b85ecc98a116ec2/profiles/bicycle.lua#L221

daniel-j-h:
For #3231: there should be no reason our _default_ profiles expose internals to the user.

## Tasklist
 - [x] review
 - [x] adjust for comments

References:
- https://github.com/Project-OSRM/osrm-text-instructions/issues/62#issuecomment-257889721
- https://github.com/Project-OSRM/osrm-text-instructions/pull/63

### DIFF
--- a/features/foot/ref.feature
+++ b/features/foot/ref.feature
@@ -15,8 +15,8 @@ Feature: Foot - Way ref
             | ab    | Utopia Drive | E7  |
 
         When I route I should get
-            | from | to | route                               |
-            | a    | b  | Utopia Drive,Utopia Drive           |
+            | from | to | route                     | ref   |
+            | a    | b  | Utopia Drive,Utopia Drive | E7,E7 |
 
     Scenario: Foot - Way with only ref
         Given the node map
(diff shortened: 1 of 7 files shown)


TheMarex:
Looks good to me. :+1:

features/foot/ref.feature
```diff
@@ -29,8 +29,8 @@ Feature: Foot - Way ref
             | ab    |      | E7  |
 
         When I route I should get
-            | from | to | route                               |
-            | a    | b  | {highway:primary},{highway:primary} |
+            | from | to | route |
```
MoKob: shouldn't this test (since it has refs) check for refs?

daniel-j-h: :+1:

daniel-j-h:
This was merged into master and the branch then deleted.
Seems like Github saw this as first deleting the branch and then merging.

PR #3264, opened 2016-11-08
Refactor fork handling in guidance

chaupow:
# Issue

This PR works towards #3099

## Tasklist
 - [x] refactor `TurnHandler::findFork()` code
 - [x] refactor `TurnHandler::isObviousOfTwo()`
 - [ ] revise definition of forks and `TurnHandler::findFork()` 
 - [ ] refactor `IntersectionHandler::assignFork()`
 - [ ] review
 - [ ] adjust for comments

## Tasklist unrelated to this PR (but arose during this work)
 - [ ] rename `struct Intersection` in `route_step.hpp`
 - [ ] revise `BasicTurnType`
 - [ ] move `isObviousByRoadClass` in `turn_handler.hpp` to a dedicated .compatible() in road_class (road class refactor)
 - [ ] refactor `struct Fork` to use a `begin` and `end` iterator instead of `right` and `left`

## Code Review Todolist
 - [x] deduplicate [leftmost / rightmost code](https://github.com/Project-OSRM/osrm-backend/blob/58537a1fe4696182a048471ca06e45758f7906f7/src/extractor/guidance/turn_handler.cpp#L530-L572)
 - [x] add `OSRM_ATTR_WARN_UNUSED` to `setTurnTypes` in `turn_handler.hpp`
 - [x] refactor `findFork()` return type
 - [x] capsule [this](https://github.com/Project-OSRM/osrm-backend/blob/58537a1fe4696182a048471ca06e45758f7906f7/src/extractor/guidance/turn_handler.cpp#L488-L497) in own function
 - [x] revert [this condition](https://github.com/Project-OSRM/osrm-backend/blob/58537a1fe4696182a048471ca06e45758f7906f7/src/extractor/guidance/turn_handler.cpp#L500)
- [x] Make member function: [1](https://github.com/Project-OSRM/osrm-backend/blob/58537a1fe4696182a048471ca06e45758f7906f7/src/extractor/guidance/turn_handler.cpp#L600) [2](https://github.com/Project-OSRM/osrm-backend/blob/58537a1fe4696182a048471ca06e45758f7906f7/src/extractor/guidance/turn_handler.cpp#L620) [3](https://github.com/Project-OSRM/osrm-backend/blob/58537a1fe4696182a048471ca06e45758f7906f7/src/extractor/guidance/turn_handler.cpp#L638).

### DIFF
--- a/include/extractor/guidance/intersection.hpp
+++ b/include/extractor/guidance/intersection.hpp
@@ -300,7 +300,7 @@ struct Intersection final : std::vector<ConnectedRoad>,         //
     std::uint8_t getHighestConnectedLaneCount(const util::NodeBasedDynamicGraph &) const;
 
     // check if all roads between first and last allow entry
-    bool hasValidEntries(std::size_t first, std::size_t last) const;
+    bool hasValidEntries(Intersection::iterator first, Intersection::iterator last) const;
 
     Base::iterator findClosestTurn(double angle);
     Base::const_iterator findClosestTurn(double angle) const;
--- a/include/extractor/guidance/intersection.hpp
+++ b/include/extractor/guidance/intersection.hpp
@@ -26,6 +26,14 @@ namespace guidance
 {
 
 // the shape of an intersection only knows about edge IDs and bearings
+// It a way of switching onto a segment, indicated by an EdgeID. The
+// associated turn is described by an angle and an instruction that is used to announce it.
+// The Turn Operation indicates what is exposed to the outside of the turn analysis.
+//
+// `angle`      is given counter-clockwise:
+//              0 = uturn, 90 = right, 180 = straight, 270 = left
+// `bearing`    is the direction in clockwise angle from true north after taking the turn:
+//              0 = heading north, 90 = east, 180 = south, 270 = west
 struct IntersectionShapeData
 {
     EdgeID eid;
@@ -271,11 +279,31 @@ struct IntersectionView final : std::vector<IntersectionViewData>,      //
     using Base = std::vector<IntersectionViewData>;
 };
 
+// `Intersection` is a relative view of an intersection by an incoming edge.
+// `Intersection` are streets at an intersection ordered from from sharp right counter-clockwise to
+// sharp left where `intersection[0]` is _always_ a u-turn
 struct Intersection final : std::vector<ConnectedRoad>,         //
                             EnableShapeOps<Intersection>,       //
                             EnableIntersectionOps<Intersection> //
 {
     using Base = std::vector<ConnectedRoad>;
+
+    /*
+     * Check validity of the intersection object. We assume a few basic properties every set of
+     * connected roads should follow throughout guidance pre-processing. This utility function
+     * allows checking intersections for validity
+     */
+    bool valid() const;
+
+    // given all possible turns, which is the highest connected number of lanes per turn. This value
+    // is used, for example, during generation of intersections.
+    std::uint8_t getHighestConnectedLaneCount(const util::NodeBasedDynamicGraph &) const;
+
+    // check if all roads between first and last allow entry
+    bool hasValidEntries(std::size_t first, std::size_t last) const;
+
+    Base::iterator findClosestTurn(double angle);
+    Base::const_iterator findClosestTurn(double angle) const;
 };
 
 } // namespace guidance
--- a/include/extractor/guidance/intersection.hpp
+++ b/include/extractor/guidance/intersection.hpp
@@ -9,6 +9,7 @@
 #include <vector>
 
 #include "util/bearing.hpp"
+#include "util/log.hpp"
 #include "util/node_based_graph.hpp"
 #include "util/typedefs.hpp" // EdgeID
 
@@ -26,12 +27,6 @@ namespace guidance
 {
 
 // the shape of an intersection only knows about edge IDs and bearings
-// It a way of switching onto a segment, indicated by an EdgeID. The
-// associated turn is described by an angle and an instruction that is used to announce it.
-// The Turn Operation indicates what is exposed to the outside of the turn analysis.
-//
-// `angle`      is given counter-clockwise:
-//              0 = uturn, 90 = right, 180 = straight, 270 = left
 // `bearing`    is the direction in clockwise angle from true north after taking the turn:
 //              0 = heading north, 90 = east, 180 = south, 270 = west
 struct IntersectionShapeData
@@ -188,6 +183,12 @@ template <typename Self> struct EnableIntersectionOps
         auto comp = makeCompareAngularDeviation(angle);
         return boost::range::min_element(*self(), comp);
     }
+    // returns a non-const_interator
+    auto findClosestTurn(double angle)
+    {
+        auto comp = makeCompareAngularDeviation(angle);
+        return std::min_element(self()->begin(), self()->end(), comp);
+    }
 
     /* Check validity of the intersection object. We assume a few basic properties every set of
      * connected roads should follow throughout guidance pre-processing. This utility function
@@ -267,6 +268,18 @@ template <typename Self> struct EnableIntersectionOps
         return filter(*candidate) ? self()->end() : candidate;
     }
 
+    // check if all roads between begin and end allow entry
+    template <typename InputIt>
+    bool hasAllValidEntries(const InputIt begin, const InputIt end) const
+    {
+        static_assert(
+            std::is_base_of<std::input_iterator_tag,
+                            typename std::iterator_traits<InputIt>::iterator_category>::value,
+            "hasAllValidEntries() only accepts input iterators");
+        return std::all_of(
+            begin, end, [](const IntersectionViewData &road) { return road.entry_allowed; });
+    }
+
   private:
     auto self() { return static_cast<Self *>(this); }
     auto self() const { return static_cast<const Self *>(this); }
@@ -282,28 +295,32 @@ struct IntersectionView final : std::vector<IntersectionViewData>,      //
 // `Intersection` is a relative view of an intersection by an incoming edge.
 // `Intersection` are streets at an intersection ordered from from sharp right counter-clockwise to
 // sharp left where `intersection[0]` is _always_ a u-turn
+
+// An intersection is an ordered list of connected roads ordered from from sharp right
+// counter-clockwise to sharp left where `intersection[0]` is always a u-turn
+//
+//                                           |
+//                                           |
+//                                     (intersec[3])
+//                                           |
+//                                           |
+//                                           |
+//  nid ---(via_eid/intersec[0])--- nbg.GetTarget(via)  ---(intersec[2])---
+//                                           |
+//                                           |
+//                                           |
+//                                     (intersec[1])
+//                                           |
+//                                           |
+//
+// intersec := intersection
+// nbh := node_based_graph
+//
 struct Intersection final : std::vector<ConnectedRoad>,         //
                             EnableShapeOps<Intersection>,       //
                             EnableIntersectionOps<Intersection> //
 {
     using Base = std::vector<ConnectedRoad>;
-
-    /*
-     * Check validity of the intersection object. We assume a few basic properties every set of
-     * connected roads should follow throughout guidance pre-processing. This utility function
-     * allows checking intersections for validity
-     */
-    bool valid() const;
-
-    // given all possible turns, which is the highest connected number of lanes per turn. This value
-    // is used, for example, during generation of intersections.
-    std::uint8_t getHighestConnectedLaneCount(const util::NodeBasedDynamicGraph &) const;
-
-    // check if all roads between first and last allow entry
-    bool hasValidEntries(Intersection::iterator first, Intersection::iterator last) const;
-
-    Base::iterator findClosestTurn(double angle);
-    Base::const_iterator findClosestTurn(double angle) const;
 };
 
 } // namespace guidance
--- a/include/extractor/guidance/intersection_handler.hpp
+++ b/include/extractor/guidance/intersection_handler.hpp
@@ -42,34 +42,10 @@ class IntersectionHandler
     virtual ~IntersectionHandler() = default;
 
     // check whether the handler can actually handle the intersection
-    //
-    // note that `intersection` is a ordered list of connected roads ordered from from sharp right
-    // counter-clockwise to sharp left where `intersection[0]` is always a u-turn
-    // #IntersectionExplanation@intersection_handler.hpp
-    //
-    //                                           |
-    //                                           |
-    //                                     (intersec[3])
-    //                                           |
-    //                                           |
-    //                                           |
-    //  nid ---(via_eid/intersec[0])--- nbg.GetTarget(via)  ---(intersec[2])---
-    //                                           |
-    //                                           |
-    //                                           |
-    //                                     (intersec[1])
-    //                                           |
-    //                                           |
-    //
-    // intersec := intersection
-    // nbh := node_based_graph
-    //
-    // #IntersectionExplanation@intersection_handler.hpp
     virtual bool
     canProcess(const NodeID nid, const EdgeID via_eid, const Intersection &intersection) const = 0;
 
-    // handle and process the intersection with parameters as described in
-    // #IntersectionExplanation@intersection_handler.hpp
+    // handle and process the intersection
     virtual Intersection
     operator()(const NodeID nid, const EdgeID via_eid, Intersection intersection) const = 0;
 
@@ -80,8 +56,6 @@ class IntersectionHandler
     const SuffixTable &street_name_suffix_table;
     const IntersectionGenerator &intersection_generator;
     const NodeBasedGraphWalker graph_walker; // for skipping traffic signal, distances etc.
-
-    std::size_t countValid(const Intersection &intersection) const;
 
     // Decide on a basic turn types
     TurnType::Enum findBasicTurnType(const EdgeID via_edge, const ConnectedRoad &candidate) const;
--- a/include/extractor/guidance/intersection_handler.hpp
+++ b/include/extractor/guidance/intersection_handler.hpp
@@ -42,10 +42,34 @@ class IntersectionHandler
     virtual ~IntersectionHandler() = default;
 
     // check whether the handler can actually handle the intersection
+    //
+    // note that `intersection` is a ordered list of connected roads ordered from from sharp right
+    // counter-clockwise to sharp left where `intersection[0]` is always a u-turn
+    // #IntersectionExplanation@intersection_handler.hpp
+    //
+    //                                           |
+    //                                           |
+    //                                     (intersec[3])
+    //                                           |
+    //                                           |
+    //                                           |
+    //  nid ---(via_eid/intersec[0])--- nbg.GetTarget(via)  ---(intersec[2])---
+    //                                           |
+    //                                           |
+    //                                           |
+    //                                     (intersec[1])
+    //                                           |
+    //                                           |
+    //
+    // intersec := intersection
+    // nbh := node_based_graph
+    //
+    // #IntersectionExplanation@intersection_handler.hpp
     virtual bool
     canProcess(const NodeID nid, const EdgeID via_eid, const Intersection &intersection) const = 0;
 
-    // process the intersection
+    // handle and process the intersection with parameters as described in
+    // #IntersectionExplanation@intersection_handler.hpp
     virtual Intersection
     operator()(const NodeID nid, const EdgeID via_eid, Intersection intersection) const = 0;
 
@@ -56,6 +80,8 @@ class IntersectionHandler
     const SuffixTable &street_name_suffix_table;
     const IntersectionGenerator &intersection_generator;
     const NodeBasedGraphWalker graph_walker; // for skipping traffic signal, distances etc.
+
+    std::size_t countValid(const Intersection &intersection) const;
 
     // Decide on a basic turn types
     TurnType::Enum findBasicTurnType(const EdgeID via_edge, const ConnectedRoad &candidate) const;
--- a/include/extractor/guidance/turn_handler.hpp
+++ b/include/extractor/guidance/turn_handler.hpp
@@ -50,11 +50,10 @@ class TurnHandler : public IntersectionHandler
   private:
     struct Fork
     {
-        const std::size_t right;
-        const std::size_t left;
+        Intersection::iterator right;
+        Intersection::iterator left;
         const std::size_t size;
-
-        Fork(const std::size_t right, const std::size_t left);
+        Fork(Intersection::iterator right, Intersection::iterator left);
     };
 
     bool isObviousOfTwo(const EdgeID via_edge,
@@ -63,15 +62,13 @@ class TurnHandler : public IntersectionHandler
 
     bool hasObvious(const EdgeID &via_edge,
                     const Intersection &intersection,
-                    const std::size_t right,
-                    const std::size_t left) const;
+                    const Fork) const;
 
     boost::optional<Fork>
-    findLeftAndRightmostForkCandidates(const Intersection &intersection) const;
+    findLeftAndRightmostForkCandidates(Intersection &intersection) const;
 
     bool isCompatibleByRoadClass(const Intersection &intersection,
-                                 const std::size_t right,
-                                 const std::size_t left) const;
+                                 const Fork fork) const;
 
     // Dead end.
     OSRM_ATTR_WARN_UNUSED
@@ -93,7 +90,7 @@ class TurnHandler : public IntersectionHandler
     handleDistinctConflict(const EdgeID via_edge, ConnectedRoad &left, ConnectedRoad &right) const;
 
     // Classification
-    boost::optional<Fork> findFork(const EdgeID via_edge, const Intersection &intersection) const;
+    boost::optional<Fork> findFork(const EdgeID via_edge, Intersection &intersection) const;
 
     OSRM_ATTR_WARN_UNUSED
     Intersection assignLeftTurns(const EdgeID via_edge,
--- a/include/extractor/guidance/turn_handler.hpp
+++ b/include/extractor/guidance/turn_handler.hpp
@@ -10,6 +10,9 @@
 #include "util/name_table.hpp"
 #include "util/node_based_graph.hpp"
 
+#include <boost/optional.hpp>
+
+
 #include <cstddef>
 #include <utility>
 #include <vector>
@@ -45,6 +48,15 @@ class TurnHandler : public IntersectionHandler
                             Intersection intersection) const override final;
 
   private:
+    struct Fork
+    {
+        const std::size_t right;
+        const std::size_t left;
+        const std::size_t size;
+
+        Fork(const std::size_t right, const std::size_t left);
+    };
+
     bool isObviousOfTwo(const EdgeID via_edge,
                         const ConnectedRoad &road,
                         const ConnectedRoad &other) const;
@@ -54,9 +66,7 @@ class TurnHandler : public IntersectionHandler
                     const std::size_t right,
                     const std::size_t left) const;
 
-    std::tuple<std::size_t, double> findClosestToStraight(const Intersection &intersection) const;
-
-    std::pair<std::size_t, std::size_t>
+    boost::optional<Fork>
     findLeftAndRightmostForkCandidates(const Intersection &intersection) const;
 
     bool isCompatibleByRoadClass(const Intersection &intersection,
@@ -83,8 +93,7 @@ class TurnHandler : public IntersectionHandler
     handleDistinctConflict(const EdgeID via_edge, ConnectedRoad &left, ConnectedRoad &right) const;
 
     // Classification
-    std::pair<std::size_t, std::size_t> findFork(const EdgeID via_edge,
-                                                 const Intersection &intersection) const;
+    boost::optional<Fork> findFork(const EdgeID via_edge, const Intersection &intersection) const;
 
     OSRM_ATTR_WARN_UNUSED
     Intersection assignLeftTurns(const EdgeID via_edge,
--- a/include/extractor/guidance/turn_handler.hpp
+++ b/include/extractor/guidance/turn_handler.hpp
@@ -48,6 +48,21 @@ class TurnHandler : public IntersectionHandler
     bool isObviousOfTwo(const EdgeID via_edge,
                         const ConnectedRoad &road,
                         const ConnectedRoad &other) const;
+
+    bool hasObvious(const EdgeID &via_edge,
+                    const Intersection &intersection,
+                    const std::size_t right,
+                    const std::size_t left) const;
+
+    std::tuple<std::size_t, double> findClosestToStraight(const Intersection &intersection) const;
+
+    std::pair<std::size_t, std::size_t>
+    findLeftAndRightmostForkCandidates(const Intersection &intersection) const;
+
+    bool isCompatibleByRoadClass(const Intersection &intersection,
+                                 const std::size_t right,
+                                 const std::size_t left) const;
+
     // Dead end.
     OSRM_ATTR_WARN_UNUSED
     Intersection handleOneWayTurn(Intersection intersection) const;
--- a/include/extractor/guidance/turn_handler.hpp
+++ b/include/extractor/guidance/turn_handler.hpp
@@ -12,7 +12,6 @@
 
 #include <boost/optional.hpp>
 
-
 #include <cstddef>
 #include <utility>
 #include <vector>
@@ -50,25 +49,21 @@ class TurnHandler : public IntersectionHandler
   private:
     struct Fork
     {
-        Intersection::iterator right;
-        Intersection::iterator left;
+        const Intersection::iterator right;
+        const Intersection::iterator left;
         const std::size_t size;
-        Fork(Intersection::iterator right, Intersection::iterator left);
+        Fork(const Intersection::iterator right, const Intersection::iterator left);
     };
 
     bool isObviousOfTwo(const EdgeID via_edge,
                         const ConnectedRoad &road,
                         const ConnectedRoad &other) const;
 
-    bool hasObvious(const EdgeID &via_edge,
-                    const Intersection &intersection,
-                    const Fork) const;
+    bool hasObvious(const EdgeID &via_edge, const Fork &fork) const;
 
-    boost::optional<Fork>
-    findLeftAndRightmostForkCandidates(Intersection &intersection) const;
+    boost::optional<Fork> findForkCandidatesByGeometry(Intersection &intersection) const;
 
-    bool isCompatibleByRoadClass(const Intersection &intersection,
-                                 const Fork fork) const;
+    bool isCompatibleByRoadClass(const Intersection &intersection, const Fork fork) const;
 
     // Dead end.
     OSRM_ATTR_WARN_UNUSED
--- a/src/extractor/extractor.cpp
+++ b/src/extractor/extractor.cpp
@@ -55,6 +55,11 @@
 #include <unordered_map>
 #include <vector>
 
+// // uncomment for @fork_geoprinting
+// #include "util/geojson_debug_logger.hpp"
+// #include "util/geojson_debug_policies.hpp"
+
+
 namespace osrm
 {
 namespace extractor
@@ -491,6 +496,11 @@ Extractor::BuildEdgeExpandedGraph(ScriptingEnvironment &scripting_environment,
         turn_lane_offsets,
         turn_lane_masks,
         turn_lane_map);
+
+
+    // // uncomment for @fork_geoprinting
+    // util::ScopedGeojsonLoggerGuard<util::NodeIdVectorToLineString> line_logger_guard("fork-lines.geojson", internal_to_external_node_map);
+    // util::ScopedGeojsonLoggerGuard<util::NodeIdVectorToMultiPoint> point_logger_guard("fork-points.geojson", internal_to_external_node_map);
 
     edge_based_graph_factory.Run(scripting_environment,
                                  config.edge_output_path,
--- a/src/extractor/extractor.cpp
+++ b/src/extractor/extractor.cpp
@@ -55,11 +55,6 @@
 #include <unordered_map>
 #include <vector>
 
-// // uncomment for @fork_geoprinting
-// #include "util/geojson_debug_logger.hpp"
-// #include "util/geojson_debug_policies.hpp"
-
-
 namespace osrm
 {
 namespace extractor
@@ -496,11 +491,6 @@ Extractor::BuildEdgeExpandedGraph(ScriptingEnvironment &scripting_environment,
         turn_lane_offsets,
         turn_lane_masks,
         turn_lane_map);
-
-
-    // // uncomment for @fork_geoprinting
-    // util::ScopedGeojsonLoggerGuard<util::NodeIdVectorToLineString> line_logger_guard("fork-lines.geojson", internal_to_external_node_map);
-    // util::ScopedGeojsonLoggerGuard<util::NodeIdVectorToMultiPoint> point_logger_guard("fork-points.geojson", internal_to_external_node_map);
 
     edge_based_graph_factory.Run(scripting_environment,
                                  config.edge_output_path,
--- a/src/extractor/guidance/intersection.cpp
+++ b/src/extractor/guidance/intersection.cpp
@@ -87,58 +87,6 @@ std::string toString(const ConnectedRoad &road)
     return result;
 }
 
-Intersection::Base::iterator Intersection::findClosestTurn(double angle)
-{
-    // use the const operator to avoid code duplication
-    return begin() +
-           std::distance(cbegin(), static_cast<const Intersection *>(this)->findClosestTurn(angle));
-}
-
-Intersection::Base::const_iterator Intersection::findClosestTurn(double angle) const
-{
-    return std::min_element(
-        begin(), end(), [angle](const ConnectedRoad &lhs, const ConnectedRoad &rhs) {
-            return angularDeviation(lhs.angle, angle) < angularDeviation(rhs.angle, angle);
-        });
-}
-bool Intersection::valid() const
-{
-    return !empty() &&
-           std::is_sorted(begin(), end(), std::mem_fn(&ConnectedRoad::compareByAngle)) &&
-           operator[](0).angle < std::numeric_limits<double>::epsilon();
-}
-
-std::uint8_t
-Intersection::getHighestConnectedLaneCount(const util::NodeBasedDynamicGraph &graph) const
-{
-    BOOST_ASSERT(valid()); // non empty()
-
-    const std::function<std::uint8_t(const ConnectedRoad &)> to_lane_count =
-        [&](const ConnectedRoad &road) {
-            return graph.GetEdgeData(road.eid).road_classification.GetNumberOfLanes();
-        };
-
-    std::uint8_t max_lanes = 0;
-    const auto extract_maximal_value = [&max_lanes](std::uint8_t value) {
-        max_lanes = std::max(max_lanes, value);
-        return false;
-    };
-
-    const auto view = *this | boost::adaptors::transformed(to_lane_count);
-    boost::range::find_if(view, extract_maximal_value);
-    return max_lanes;
-}
-
-// check if all entries in the given range allow entry
-bool Intersection::hasValidEntries(Intersection::iterator first, Intersection::iterator last) const
-{
-    BOOST_ASSERT(last < end());
-
-    return all_of(first, last + 1, [&](const ConnectedRoad &road) {
-        return road.entry_allowed;
-    });
-}
-
 } // namespace guidance
 } // namespace extractor
 } // namespace osrm
--- a/src/extractor/guidance/intersection.cpp
+++ b/src/extractor/guidance/intersection.cpp
@@ -85,6 +85,65 @@ std::string toString(const ConnectedRoad &road)
     return result;
 }
 
+Intersection::Base::iterator Intersection::findClosestTurn(double angle)
+{
+    // use the const operator to avoid code duplication
+    return begin() +
+           std::distance(cbegin(), static_cast<const Intersection *>(this)->findClosestTurn(angle));
+}
+
+Intersection::Base::const_iterator Intersection::findClosestTurn(double angle) const
+{
+    return std::min_element(
+        begin(), end(), [angle](const ConnectedRoad &lhs, const ConnectedRoad &rhs) {
+            return angularDeviation(lhs.angle, angle) < angularDeviation(rhs.angle, angle);
+        });
+}
+bool Intersection::valid() const
+{
+    return !empty() &&
+           std::is_sorted(begin(), end(), std::mem_fn(&ConnectedRoad::compareByAngle)) &&
+           operator[](0).angle < std::numeric_limits<double>::epsilon();
+}
+
+std::uint8_t
+Intersection::getHighestConnectedLaneCount(const util::NodeBasedDynamicGraph &graph) const
+{
+    BOOST_ASSERT(valid()); // non empty()
+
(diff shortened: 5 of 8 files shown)


include/extractor/guidance/intersection.hpp
```diff
@@ -71,7 +76,10 @@ struct ConnectedRoad final : public TurnOperation
 // small helper function to print the content of a connected road
 std::string toString(const ConnectedRoad &road);
 
-struct Intersection final : public std::vector<ConnectedRoad>
+// `ConnectedRoads` is a relative view of an intersection by an incoming edge.
+// `ConnectedRoads` are streets at an intersection ordered from from sharp right counter-clockwise to
+// sharp left where `connected_roads[0]` is _always_ a u-turn
+struct ConnectedRoads final : public std::vector<ConnectedRoad>
```
daniel-j-h: Why the change in the API here? This will break _all_ users throughout osrm-backend.

Also semantically it it an intersection represented as ordered roads.

include/extractor/guidance/intersection_handler.hpp
```diff
@@ -34,13 +34,42 @@ class IntersectionHandler
...
-
-    // process the intersection
-    virtual Intersection
-    operator()(const NodeID nid, const EdgeID via_eid, Intersection intersection) const = 0;
+    // check whether the handler can actually handle
+    // the intersection
+    //     _at_ `node_based_graph.GetTarget(via_edge)`
+    //     _from_ `NodeID`
+    //     _over_ `via_edge`
+    //     _from where_ there are other `connected_roads`
```
daniel-j-h: What is `_at_` and `_from where_` all about?

src/extractor/guidance/intersection_handler.cpp
```diff
@@ -38,13 +38,17 @@ IntersectionHandler::IntersectionHandler(const util::NodeBasedDynamicGraph &node
...
 {
     return std::count_if(intersection.begin(), intersection.end(), [](const ConnectedRoad &road) {
         return road.entry_allowed;
     });
 }
 
+// Checks whether a turn from `via_edge` onto `road` is a turn
+// - on a ramp
+// - a continue
+// - or a normal turn
```
daniel-j-h: What does this mean wrt. our turn instructions?

https://github.com/Project-OSRM/osrm-backend/blob/master/include/extractor/guidance/turn_instruction.hpp#L33

src/extractor/guidance/turn_handler.cpp
```diff
@@ -459,124 +476,187 @@ Intersection TurnHandler::assignRightTurns(const EdgeID via_edge,
...
+            }
+            else if (angle_between_prev_road_and_right <= NARROW_TURN_ANGLE && angle_between_right_road_and_straight <= GROUP_ANGLE)
+            {
+                --right;
+            }
+            else
+            {
+                break;
+            }
+        }
```
daniel-j-h: I think you can express this and above more clearly using a adjacent-ish stdlib algorithm

src/extractor/guidance/turn_handler.cpp
```diff
@@ -459,124 +476,187 @@ Intersection TurnHandler::assignRightTurns(const EdgeID via_edge,
...
+        // then there are no two roads that are close enough to be considered a fork
         if (left == right)
+        {
             return std::make_pair(std::size_t{0}, std::size_t{0});
+        }
+
 
-        const bool valid_indices = 0 < right && right < left;
+        BOOST_ASSERT(0 < right);
+        BOOST_ASSERT(right < left);
```
daniel-j-h: Change from a check to an assert: does this always hold?

src/extractor/guidance/turn_handler.cpp
```diff
@@ -459,124 +476,187 @@ Intersection TurnHandler::assignRightTurns(const EdgeID via_edge,
...
         const bool separated_at_right_side =
-            right > 0 &&
-            angularDeviation(intersection[right].angle, intersection[right - 1].angle) >=
-                GROUP_ANGLE;
+            angularDeviation(connected_roads[right].angle, connected_roads[right - 1].angle) >=
+            GROUP_ANGLE;
+
+        // there cannot be a fork of more than three streets
+        std::size_t size_of_fork = left - right + 1;
+        const bool not_more_than_three_old = (left - right) <= 2;
```
daniel-j-h: Dead code

daniel-j-h:
Looking at your refactored [findFork](https://github.com/Project-OSRM/osrm-backend/blob/58537a1fe4696182a048471ca06e45758f7906f7/src/extractor/guidance/turn_handler.cpp#L479-L661): 

- What is the return type? A `std::pair<std::size_t, std::size_t>` expresses no semantic meaning. Looks like those are indices into the Intersection array. What if `findFork` fails, what's the return type then? What about the following: either return an iterator pair, or create a local `struct Fork { .. };` and change to `optional<Fork> findFork(..);`.
- Does `findFork` has to work with all settings or can you put a pre-condition in to only handle specific scenarios, e.g. the Intersection array has to have a size of at least 3 (1 uturn + 2 for fork).
- What about refactoring [this](https://github.com/Project-OSRM/osrm-backend/blob/58537a1fe4696182a048471ca06e45758f7906f7/src/extractor/guidance/turn_handler.cpp#L488-L497) into its own `findClosestToStraight` (feel free to change all namings mentioned here) function and then using e.g. a min stdlib function.
- Looks like we only handle [this branch](https://github.com/Project-OSRM/osrm-backend/blob/58537a1fe4696182a048471ca06e45758f7906f7/src/extractor/guidance/turn_handler.cpp#L500). What about inverting the condition and returning early if it's not met. Otherwise we need to carry around an extra level of indentation. Related: please [clang-format-3.8 your code](https://github.com/Project-OSRM/osrm-backend/blob/58537a1fe4696182a048471ca06e45758f7906f7/src/extractor/guidance/turn_handler.cpp#L533).
- The [leftmost / rightmost code](https://github.com/Project-OSRM/osrm-backend/blob/58537a1fe4696182a048471ca06e45758f7906f7/src/extractor/guidance/turn_handler.cpp#L530-L572) here is duplicated. Should be able to 1/ put it into its own function and 2/ write once, run on fwd iterators and then again on reverse iterators. You can probably use adjacent find from the stdlib for this.
- [Why?](https://github.com/Project-OSRM/osrm-backend/blob/58537a1fe4696182a048471ca06e45758f7906f7/src/extractor/guidance/turn_handler.cpp#L595)
- Make member function: [1](https://github.com/Project-OSRM/osrm-backend/blob/58537a1fe4696182a048471ca06e45758f7906f7/src/extractor/guidance/turn_handler.cpp#L600) [2](https://github.com/Project-OSRM/osrm-backend/blob/58537a1fe4696182a048471ca06e45758f7906f7/src/extractor/guidance/turn_handler.cpp#L620) [3](https://github.com/Project-OSRM/osrm-backend/blob/58537a1fe4696182a048471ca06e45758f7906f7/src/extractor/guidance/turn_handler.cpp#L638).
- Use the GeoJSON printer and check what you classify as fork e.g. on Berlin.
- Then check [this heuristic](https://github.com/Project-OSRM/osrm-backend/blob/58537a1fe4696182a048471ca06e45758f7906f7/src/extractor/guidance/turn_handler.cpp#L654).

include/extractor/geojson_debug_policies.hpp
```diff
@@ -27,7 +27,7 @@ struct IntersectionPrinter
     // renders the used coordinate locations for all entries/as well as the resulting
     // intersection-classification
     util::json::Array operator()(const NodeID intersection_node,
-                                 const extractor::guidance::Intersection &intersection,
+                                 const extractor::guidance::ConnectedRoads &intersection,
```
MoKob: can we split unrelated changes like these into a dedicated PR?

include/extractor/guidance/turn_analysis.hpp
```diff
@@ -46,12 +46,13 @@ class TurnAnalysis
...
-    Intersection
-    assignTurnTypes(const NodeID from_node, const EdgeID via_eid, Intersection intersection) const;
+    ConnectedRoads getIntersection(const NodeID from_node, const EdgeID via_eid) const;
+    ConnectedRoads assignTurnTypes(const NodeID from_node,
+                                   const EdgeID via_eid,
+                                   ConnectedRoads intersection) const;
 
     std::vector<TurnOperation>
-    transformIntersectionIntoTurns(const Intersection &intersection) const;
+    transformIntersectionIntoTurns(const ConnectedRoads &intersection) const;
```
MoKob: if you rename it, than `intersection` in the name here does not make any sense.

include/extractor/guidance/turn_analysis.hpp
```diff
@@ -64,8 +65,8 @@ class TurnAnalysis
     const SliproadHandler sliproad_handler;
 
     // Utility function, setting basic turn types. Prepares for normal turn handling.
-    Intersection
-    setTurnTypes(const NodeID from, const EdgeID via_edge, Intersection intersection) const;
+    ConnectedRoads
```
MoKob: we should add `OSRM_ATTR_WARN_UNUSED` here.

include/extractor/guidance/intersection_generator.hpp
```diff
@@ -90,8 +90,8 @@ class IntersectionGenerator
     // The treatment results in a straight turn angle of 180º rather than a turn angle of approx
     // 160
     OSRM_ATTR_WARN_UNUSED
-    Intersection MergeSegregatedRoads(const NodeID intersection_node,
-                                      Intersection intersection) const;
+    ConnectedRoads MergeSegregatedRoads(const NodeID intersection_node,
+                                        ConnectedRoads intersection) const;
```
MoKob: if it's important to have it named `ConnectedRoads` to you, then we should also change the name `intersection` into `connected_roads` to not loose sync between type and name. Right now (`Intersection intersection`) connects the variable to its type

include/extractor/guidance/intersection_handler.hpp
```diff
@@ -34,13 +34,42 @@ class IntersectionHandler
...
+    //
+    // note that `connected_roads` are ordered from from sharp right counter-clockwise to sharp left
+    // where `connected_roads[0]` is _always_ a u-turn
+    // #ConnectedRoadsExplanation@intersection_handler.hpp
+    //
+    //
+    //                                                 |
+    //                                                 |
+    //                                         (connected_roads[i])
```
MoKob: comments out of sync with name below (connected_roads vs intersection)

include/extractor/guidance/turn_classification.hpp
```diff
@@ -25,7 +25,7 @@ namespace guidance
 {
 
 std::pair<util::guidance::EntryClass, util::guidance::BearingClass>
-classifyIntersection(Intersection intersection);
+classifyIntersection(ConnectedRoads intersection);
```
MoKob: Same here, if we name it roads, the connection between `intersection` and `connected roads` is not clear.

Sure, a `vector<ConnectedRoad>` is a list of `ConnectedRoads`, but so is `std::set<ConnectedRoad>` or `std::list<ConnectedRoad>`. I don't really like removing the additional information that is inherently given by `having a specifically sorted list of Roads`. Especially with the added functionality.

It's a dedicated class, not simply a plain collection.

chaupow:
Regarding `Intersection` vs `ConnectedRoads`:

The main reason for the change is that there is another `struct Intersection` in `route_step.hpp` and I thought that might be confusing.

As the discussed `Intersection`is basically a vector of `<ConnectedRoad>`, @themarex and me thought that `ConnectedRoads` makes sense and is also inline with the naming conventions with some other vectors and `connected_roads[0]` or `connected_roads.size()` is pretty intuitive to me.

However, if you think we should rename it to something more similar to `Intersection` I'm open to it and I’ll undo the `Intersection` changes.

MoKob:
Capturing from chat:

in short: yes, we should to stuff about `intersection` vs `intersection`. But I feel renaming `intersection` in `engine` should be the way to go, since it's only a output part

[12:21]  
by adding more functionality to `intersection` in preprocessing (e.g. `findClosestTurn`), we are moving away from a pure collection

[12:22]  
`ConnectedRoads` is a really bad choice for me, since it would be valid for a `set`, a `vector`, a `map`, a `list`, basically any collection type

[12:24]  
`intersection.size()` vs `connected_roads.size()` does not improve readability to me. The `size` of an intersection conveys some meaning, size of `connected roads` is (especially since `s` vs no `s` is easily lost) way less meaningful to me

[12:25]  
you could easily confuse `.size()` with the `size` of a road (e.g. width)

also you have the sorting criterium which implies meaning. With the addition of `.valid()`, it moves away form a pure collection to an ordered list of items. So there is more meaning to it rather than just a random collection of entries

In total: We are renaming the `route-step` intersection. Name up for discussion.

include/extractor/guidance/intersection.hpp
chaupow: reverted

include/extractor/guidance/intersection_handler.hpp
chaupow: updated comment to make it less confusing

src/extractor/guidance/intersection_handler.cpp
chaupow: moved to unrelated todo

src/extractor/guidance/turn_handler.cpp
chaupow: thanks!

include/extractor/geojson_debug_policies.hpp
chaupow: will do.

include/extractor/guidance/turn_analysis.hpp
chaupow: reverted

include/extractor/guidance/intersection_generator.hpp
chaupow: reverted

include/extractor/guidance/intersection_handler.hpp
chaupow: corrected

include/extractor/guidance/turn_classification.hpp
chaupow: reverted

src/extractor/guidance/turn_handler.cpp
chaupow: done. thanks!

TheMarex:
> in short: yes, we should to stuff about `intersection` vs `intersection`. But I feel renaming `intersection` in `engine` should be the way to go, since it's only a output part

Being on the output part is actually the prime reason I would argue against remaining it. We always refer to this object in the docs as `Intersection` object and having parity with docs and internal class names makes it quite easy to orient yourself in the API code. Changing anything that is touched by people besides internal developers always leads to confusion. In context of #3228 this might get even more confusing.

You could maybe use `IntersectionModel` or `InternalIntersection` or something similar to resolve the name conflict.

chaupow:
I did some work on this PR, things happened so far:
- [x] commented code that was necessary to understand how fork handling works
- [x] refactor `findFork()`
- [x] slightly refactor other functions and files that are strongly related to `findFork()`

what these change not do:
- [ ] changes of the _behaviour_ of `findFork()`; refactor was only done regarding code readibility
- [ ] refactor of code outside `findFork()`

With these thigns in mind, would you look at the code again before I continue work here? Thanks a lot, lot, lot! @daniel-j-h @MoKob

src/extractor/guidance/turn_handler.cpp
```diff
@@ -459,14 +522,62 @@ Intersection TurnHandler::assignRightTurns(const EdgeID via_edge,
...
+template <typename Iterator>
+std::size_t findOutermostForkCandidate(Iterator base, Iterator start, Iterator end)
+{
+    const auto outermost_fork_candidate = std::adjacent_find(start, end, isOutermostForkCandidate);
+    auto outermost_fork_candidate_index = std::distance(base, outermost_fork_candidate);
+    // if Iterator is a reverse iterator, the index is different
+    if (outermost_fork_candidate_index < 0)
+    {
+        outermost_fork_candidate_index = std::abs(outermost_fork_candidate_index) - 1;
+    }
```
daniel-j-h: `std::distance` returns "the number of elements between first and last" - when and how can this ever be negative?

src/extractor/guidance/turn_handler.cpp
```diff
@@ -459,14 +522,62 @@ Intersection TurnHandler::assignRightTurns(const EdgeID via_edge,
...
+    {
+        return outermost_fork_candidate_index - 1;
+    }
+}
 
+// @CODE_REVIEW_QUESTION: tuple or struct? I cannot decide
+// returns a tuple {i, angle_dev} of the road at the intersection that is closest to going straight:
+//     `intersection[i]` is the actual road
+//     `angle_dev` is the angle between going straight and taking the road `intersection[i]`
+std::tuple<std::size_t, double>
```
daniel-j-h: a tuple of two is a `std::pair` - but better to return a named `struct { size_t index; double angle; }` here

src/extractor/guidance/turn_handler.cpp
```diff
@@ -459,14 +522,62 @@ Intersection TurnHandler::assignRightTurns(const EdgeID via_edge,
...
+TurnHandler::findClosestToStraight(const Intersection &intersection) const
```
daniel-j-h: What happens if `intersection` is empty? Should this return a `optional<T>` or is `!empty` a pre-condition?

Or return the end iterator? Btw I'm still not sure about the size_t vs iterator is this PR.

src/extractor/guidance/turn_handler.cpp
```diff
@@ -476,108 +587,153 @@ std::pair<std::size_t, std::size_t> TurnHandler::findFork(const EdgeID via_edge,
...
+TurnHandler::findLeftAndRightmostForkCandidates(const Intersection &intersection) const
+{
+    std::size_t best;
+    double best_deviation;
+    std::tie(best, best_deviation) = findClosestToStraight(intersection);
+
+    // no fork found
+    if (best_deviation > NARROW_TURN_ANGLE)
+    {
+        return std::make_pair(std::size_t{0}, std::size_t{0});
```
daniel-j-h: Same here - semantically this is wrong, isn't it? The index 0 is the UTurn road. Shouldn’t this return an `optional<T>` and `T` being a `struct { leftmost, rightmost }`?

src/extractor/guidance/turn_handler.cpp
```diff
@@ -476,108 +587,153 @@ std::pair<std::size_t, std::size_t> TurnHandler::findFork(const EdgeID via_edge,
...
+        node_based_graph.GetEdgeData(intersection[right].eid).road_classification;
+    for (std::size_t index = right + 1; index <= left; ++index)
+    {
+        const auto road_class =
+            node_based_graph.GetEdgeData(intersection[index].eid).road_classification;
+        if (right_class.IsLinkClass() != road_class.IsLinkClass())
+        {
+            return false;
+        }
+    }
```
daniel-j-h: As said, this looks like `adjacent_find(first, last, link_predicate);`

src/extractor/guidance/turn_handler.cpp
```diff
@@ -476,108 +587,153 @@ std::pair<std::size_t, std::size_t> TurnHandler::findFork(const EdgeID via_edge,
...
-                        return false;
-                }
+                return false;
             }
-            return true;
-        }();
+        }
+    }
+    return true;
+}
```
daniel-j-h: Can you express these two nested raw for loops via stdlib algorithms?
The outer one looks like a for each, the inner one like a any-of.

src/extractor/guidance/turn_handler.cpp
```diff
@@ -476,108 +587,153 @@ std::pair<std::size_t, std::size_t> TurnHandler::findFork(const EdgeID via_edge,
...
 
-        // check if all entries in the fork range allow entry
-        const bool only_valid_entries = [&]() {
-            BOOST_ASSERT(right <= left && left < intersection.size());
+// Checks whether a three-way-intersection coming from `via_edge` is a fork
+// with `intersection` as described as in #IntersectionExplanation@intersection_handler.hpp
+TurnHandler::Fork TurnHandler::findFork(const EdgeID via_edge,
+                                        const Intersection &intersection) const
```
daniel-j-h: What happens if `intersection` is empty (the UTurn road only)? Is size = 3 a pre-condition?

src/extractor/guidance/turn_handler.cpp
```diff
@@ -476,108 +587,153 @@ std::pair<std::size_t, std::size_t> TurnHandler::findFork(const EdgeID via_edge,
...
+    std::size_t right;
+    std::size_t left;
+    std::tie(right, left) = findLeftAndRightmostForkCandidates(intersection);
+
+    // if the leftmost and rightmost roads with the conditions above are the same
+    // or if there are more than three fork candidates
+    // they cannot be fork candidates
+    if ((left == right) || ((left - right + 1) > 3))
+    {
+        return Fork::getEmptyFork();
```
daniel-j-h: As said, the default constructor already does that.

src/extractor/guidance/turn_handler.cpp
```diff
@@ -22,6 +22,30 @@ namespace extractor
...
 {
 
+// @CHAU_TODO
+// a wrapper to handle road indices of forks at intersections
+struct TurnHandler::Fork
+{
+    const std::size_t right;
+    const std::size_t left;
+    const std::size_t size;
+    const bool valid;
```
daniel-j-h: data + boolean valid is more or less what `optional<T>` is for - just make it a plain old `struct T { U data; };` and then use `optional<T>` and check with `if (myoptional) use(*myoptional);`

src/extractor/guidance/turn_handler.cpp
```diff
@@ -476,108 +587,153 @@ std::pair<std::size_t, std::size_t> TurnHandler::findFork(const EdgeID via_edge,
...
-        if (valid_indices && separated_at_left_side && separated_at_right_side &&
-            not_more_than_three && !has_obvious && has_compatible_classes && only_valid_entries)
-            return std::make_pair(right, left);
+    // A fork can only happen between edges of similar types where none of the ones is obvious
+    const bool has_compatible_classes = isCompatibleByRoadClass(intersection, right, left);
+
+    // check if all entries in the fork range allow entry
+    const bool only_valid_entries = intersection.hasValidEntries(right, left);
+
+    // TODO check whether 2*NARROW_TURN is too large
```
daniel-j-h: Where did the 2 * narrow turn angle go?

src/extractor/guidance/turn_handler.cpp
MoKob: @daniel-j-h it can be negative if first > last. `std::distance(0,1) == 1, std::distance(1,0) == -1`

src/extractor/guidance/turn_handler.cpp
chaupow: An Intersection always has the u-turn. Finding the closest to straight in such an intersection would be the u-turn so I think the return type should be ok and not optional.

Probably it would be good to have some assertions that Intersections always have the u-turn road but shouldn't be checked in this function in my opinion?

chaupow: > Btw I'm still not sure about the size_t vs iterator is this PR.

src/extractor/guidance/turn_handler.cpp
MoKob: It looks like the comment is out of date. Feel free to remove.

src/extractor/guidance/turn_handler.cpp
chaupow: @daniel-j-h I have to search for the rightmost and the leftmost road and I use the same function but call it with the reverse iterator when looking for the rightmost one.

In that case (using the reverse iterator) the index will be negative and then I will have to do that. I added an explanation in the comment.

include/extractor/guidance/intersection.hpp
```diff
@@ -215,10 +222,51 @@ struct IntersectionView final : std::vector<IntersectionViewData>,      //
...
+
+    // given all possible turns, which is the highest connected number of lanes per turn. This value
+    // is used, for example, during generation of intersections.
+    std::uint8_t getHighestConnectedLaneCount(const util::NodeBasedDynamicGraph &) const;
+
+    // check if all roads between first and last allow entry
+    bool hasValidEntries(Intersection::iterator first, Intersection::iterator last) const;
+
+    Base::iterator findClosestTurn(double angle);
+    Base::const_iterator findClosestTurn(double angle) const;
```
daniel-j-h: All of the functions here should no longer be attached to `Intersection` or `IntersectionView` - but instead check `EnableIntersectionOps<T>`. If you have the time for it, read up about the CRTP pattern.

We need this since even though this creates a inheritance hierarchy

```c++
struct base {};
struct derived : base {};
```

the following types stand in no relationship to each other

```c++
using BaseVec = vector<base>;
using DerivedVec = vector<derived>;
```

using the CRTP pattern we inject the functionality into both collection types.

include/extractor/guidance/turn_handler.hpp
```diff
@@ -45,9 +47,24 @@ class TurnHandler : public IntersectionHandler
                             Intersection intersection) const override final;
 
   private:
+    struct Fork
+    {
+        Intersection::iterator right;
+        Intersection::iterator left;
+        const std::size_t size;
+        Fork(Intersection::iterator right, Intersection::iterator left);
```
daniel-j-h: Should this work on `Intersection` or `IntersectionView` - or both?

Check the data these types hold and decide.

src/extractor/extractor.cpp
```diff
@@ -55,6 +55,10 @@
 #include <unordered_map>
 #include <vector>
 
+// // uncomment for @fork_geoprinting
+// #include "util/geojson_debug_logger.hpp"
+// #include "util/geojson_debug_policies.hpp"
+
```
daniel-j-h: remove before merge

src/extractor/guidance/coordinate_extractor.cpp
```diff
@@ -130,8 +130,7 @@ util::Coordinate CoordinateExtractor::ExtractRepresentativeCoordinate(
     // fallback, mostly necessary for dead ends
     if (intersection_node == to_node)
     {
-        const auto result = ExtractCoordinateAtLength(
-            skipping_inaccuracies_distance, coordinates);
+        const auto result = ExtractCoordinateAtLength(skipping_inaccuracies_distance, coordinates);
```
daniel-j-h: formatted?

src/extractor/guidance/intersection.cpp
```diff
@@ -65,6 +65,56 @@ std::string toString(const ConnectedRoad &road)
...
+}
+
+// check if all entries in the given range allow entry
+bool Intersection::hasValidEntries(Intersection::iterator first, Intersection::iterator last) const
+{
+    BOOST_ASSERT(last < end());
+
+    return all_of(first, last + 1, [&](const ConnectedRoad &road) { return road.entry_allowed; });
+}
+
```
daniel-j-h: This comes from the messed up rebase.

include/extractor/guidance/intersection.hpp
```diff
@@ -25,7 +25,14 @@ namespace extractor
 namespace guidance
 {
 
-// the shape of an intersection only knows about edge IDs and bearings
+// the shape of an intersection only knows about edge IDs and bearings. The
+// associated turn is described by an angle and an instruction that is used to announce it.
+// The Turn Operation indicates what is exposed to the outside of the turn analysis.
+//
+// `angle`      is given counter-clockwise:
+//              0 = uturn, 90 = right, 180 = straight, 270 = left
```
MoKob: the intersection shape itself does not know about where you come from. It does not offer an `angle` property.

include/engine/guidance/route_step.hpp
```diff
@@ -29,7 +29,26 @@ namespace guidance
 // Departure: s --> a --> b. Represents the segment s,a with location being s.
 // Arrive: a --> b --> t. The segment (b,t) is already covered by the previous segment.
 
-// A representation of intermediate intersections
+// A representation of intermediate intersections:
+// An intersection is an ordered list of connected roads ordered from from sharp right
+// counter-clockwise to sharp left where `intersection[0]` is always a u-turn
```
MoKob: the intersection in the output is a different object than the intersection in the preprocessing. Here the concept of `Connected Roads` is unknown. We offer available turn bearings (in all directions, not angles) and allowed entry flags.

include/extractor/guidance/turn_handler.hpp
MoKob: Fork handling is only called on Intersections (in the dedicated handlers), just as the sliproad handler. Right now this is correct. We could think about how we could update this behaviour, but right no I'd say for the scope of this PR having Intersections here is correct.

include/extractor/guidance/turn_handler.hpp
```diff
@@ -45,9 +47,24 @@ class TurnHandler : public IntersectionHandler
...
+    };
+
     bool isObviousOfTwo(const EdgeID via_edge,
                         const ConnectedRoad &road,
                         const ConnectedRoad &other) const;
+
+    bool hasObvious(const EdgeID &via_edge, const Fork) const;
```
MoKob: is this the replacement of the above function? Also Fork could probably benefit from a `reference`, due to its size.

include/extractor/guidance/turn_handler.hpp
```diff
@@ -45,9 +47,24 @@ class TurnHandler : public IntersectionHandler
                             Intersection intersection) const override final;
 
   private:
+    struct Fork
+    {
+        Intersection::iterator right;
+        Intersection::iterator left;
```
MoKob: since we are using iterators, wouldn't it be better to follow the stdlib paradigm for `begin`, `end` here? I'd prefer a begin / end approach at least, but given that `Fork` is wrapped into an optional, I'd be alright with `left`/`right`. However, `left` and `right` requires implicit knowledge about the order. So we could maybe use `ccw_right` and `ccw_left` to indicate the order here as well? This way you know which direction to iterate over.

include/extractor/guidance/turn_handler.hpp
```diff
@@ -45,9 +47,24 @@ class TurnHandler : public IntersectionHandler
...
+
     bool isObviousOfTwo(const EdgeID via_edge,
                         const ConnectedRoad &road,
                         const ConnectedRoad &other) const;
+
+    bool hasObvious(const EdgeID &via_edge, const Fork) const;
+
+    boost::optional<Fork> findLeftAndRightmostForkCandidates(Intersection &intersection) const;
+
+    bool isCompatibleByRoadClass(const Intersection &intersection, const Fork fork) const;
```
MoKob: should probably move to a dedicated `.compatible()` in `road_class`? We can do so in refactoring road classes, though.

include/extractor/guidance/turn_handler.hpp
```diff
@@ -68,8 +85,7 @@ class TurnHandler : public IntersectionHandler
     handleDistinctConflict(const EdgeID via_edge, ConnectedRoad &left, ConnectedRoad &right) const;
 
     // Classification
-    std::pair<std::size_t, std::size_t> findFork(const EdgeID via_edge,
-                                                 const Intersection &intersection) const;
+    boost::optional<Fork> findFork(const EdgeID via_edge, Intersection &intersection) const;
```
MoKob: could you maybe add documentation on what the difference between `findFork` and `findLeftmost...` versions are?

src/extractor/extractor.cpp
```diff
@@ -487,6 +491,12 @@ Extractor::BuildEdgeExpandedGraph(ScriptingEnvironment &scripting_environment,
         turn_lane_masks,
         turn_lane_map);
 
+    // // uncomment for @fork_geoprinting
```
MoKob: same here, also remove

src/extractor/guidance/coordinate_extractor.cpp
```diff
@@ -252,8 +249,8 @@ util::Coordinate CoordinateExtractor::ExtractRepresentativeCoordinate(
         // As a back-up, we have to check for this case
         if (coordinates.front() == coordinates.back())
         {
-            const auto result = ExtractCoordinateAtLength(
-                skipping_inaccuracies_distance, coordinates);
+            const auto result =
+                ExtractCoordinateAtLength(skipping_inaccuracies_distance, coordinates);
```
MoKob: Is your clang-format version different from mine?

src/extractor/guidance/intersection_handler.cpp
```diff
@@ -7,6 +7,10 @@
 #include <algorithm>
 #include <cstddef>
 
+// // uncomment for @fork_geoprinting
```
MoKob: remove before merge

src/extractor/guidance/intersection_handler.cpp
```diff
@@ -158,6 +164,20 @@ TurnInstruction IntersectionHandler::getInstructionForObvious(const std::size_t
     }
 }
 
+// // uncomment for @fork_geoprinting
```
MoKob: remove before merge

src/extractor/guidance/intersection_handler.cpp
```diff
@@ -191,6 +211,12 @@ void IntersectionHandler::assignFork(const EdgeID via_edge,
                 }
                 else
                 {
+                    // // uncomment for @fork_geoprinting
```
MoKob: also remove

src/extractor/guidance/intersection_handler.cpp
```diff
@@ -230,6 +256,12 @@ void IntersectionHandler::assignFork(const EdgeID via_edge,
                     }
                     else
                     {
+                        // // uncomment for @fork_geoprinting
```
MoKob: best check your entire PR for these

src/extractor/guidance/turn_handler.cpp
```diff
@@ -9,19 +9,39 @@
...
-{
-namespace guidance
-{
 namespace
 {
+
+using namespace osrm::extractor::guidance;
+// given two adjacent roads and `road1` being a candidate for a fork,
+// return false, if next road `road2` is also a fork candidate or
+// return true, if `road2` is not a suitable fork candidate and thus, `road1` the outermost fork
```
MoKob: adjacent == any specific order? I'd assume they are still in ccw order? Would be nice to specify this.

src/extractor/guidance/turn_handler.cpp
```diff
@@ -31,6 +51,67 @@ bool isEndOfRoad(const ConnectedRoad &,
...
+{
+    const auto outermost = std::adjacent_find(start, end, isOutermostForkCandidate);
+    if (outermost != end)
+    {
+        return outermost;
+    }
+    // if all roads are part of a fork, set `candidate` to the last road
+    else
+    {
+        return outermost - 1;
```
MoKob: If we would use the `begin`/`end` paradigm I mentioned above, we wouldn't need these distinctions here.

src/extractor/guidance/turn_handler.cpp
```diff
@@ -31,6 +51,67 @@ bool isEndOfRoad(const ConnectedRoad &,
...
+    }
+}
+
+struct StraightestTurnAtIntersection
```
MoKob: adding a dedicated struct here seems wrong to me. the straightmost turn at an instersection can be found with `getClosestTurn(STRAIGHT_ANGLE)` for any intersection and is an iterator into the intersection. It's deviation can easily be computed whenever necessary by `angularDeviation(itr.angle,STRAIGHT_ANGLE)`.

The dedicated struct is overkill in my personal opinion. Is it used often enough to justify the definition?

src/extractor/guidance/turn_handler.cpp
```diff
@@ -31,6 +51,67 @@ bool isEndOfRoad(const ConnectedRoad &,
...
+        : id(id), deviation_from_straight(deviation_from_straight){};
+
+    std::size_t id;
+    double deviation_from_straight;
+};
+
+// returns a tuple {i, angle_dev} of the road at the intersection that is closest to going straight:
+//     `intersection[i]` is the actual road
+//     `angle_dev` is the angle between going straight and taking the road `intersection[i]`
+StraightestTurnAtIntersection findClosestToStraight(const Intersection &intersection)
```
MoKob: can be removed completely and replaced with `findClosestTurn(STRAIGHT_ANGLE)` in `intersection`

src/extractor/guidance/turn_handler.cpp
MoKob: After checking: given that it is used twice, I feel that introducing a dedicated struct here only adds confusion. Having the indirection using an `id` instead of the iterator `findClosestTurn` returns does only add confusion in my view.

include/engine/guidance/route_step.hpp
chaupow: ugh. changed this. don't know how this happened

include/extractor/guidance/turn_handler.hpp
chaupow: good point. I also renamed `findLeftAndRightmostForkCandidates` to `findForkCandidatesByGeometry` which is hopefully a better name because we're just looking at angles and the geometric relations between the roads here.

src/extractor/guidance/turn_handler.cpp
chaupow: moving to new issue

src/extractor/guidance/turn_handler.cpp
chaupow: good catch. thanks!

src/extractor/guidance/turn_handler.cpp
chaupow: moved to new issue

src/extractor/guidance/turn_handler.cpp
chaupow: done

include/extractor/guidance/turn_handler.hpp
chaupow: moving to new issue

include/extractor/guidance/intersection.hpp
```diff
@@ -204,6 +212,15 @@ template <typename Self> struct EnableIntersectionOps
     // Returns the number of roads we can not enter at this intersection, respectively.
     auto countNonEnterable() const { return self()->size() - self()->countEnterable(); }
 
+    // check if all roads between begin and end allow entry
+    template <typename iteratorT>
+    bool hasAllValidEntries(const iteratorT begin, const iteratorT end) const
+    {
+        BOOST_ASSERT(end < self()->end());
+
+        return all_of(begin, end, [](const ConnectedRoad &road) { return road.entry_allowed; });
```
MoKob: this should use type `auto`, since we can also call it on an intersectionView which does not have a connected road.

include/extractor/guidance/intersection.hpp
```diff
@@ -125,6 +127,12 @@ template <typename Self> struct EnableIntersectionOps
         auto comp = makeCompareAngularDeviation(angle);
         return std::min_element(self()->begin(), self()->end(), comp);
     }
+    // returns a non-const_interator
+    auto findClosestTurn(double angle)
+    {
+        auto comp = makeCompareAngularDeviation(angle);
+        return std::min_element(self()->begin(), self()->end(), comp);
+    }
```
daniel-j-h: Non-const overload for the function above?

There is a trick involving a const-cast so you don't have to duplicate code.
Search around for how to implement const / non-const overloads.

Not sure if we need it here ..

include/extractor/guidance/intersection.hpp
```diff
@@ -204,6 +212,13 @@ template <typename Self> struct EnableIntersectionOps
     // Returns the number of roads we can not enter at this intersection, respectively.
     auto countNonEnterable() const { return self()->size() - self()->countEnterable(); }
 
+    // check if all roads between begin and end allow entry
+    template <typename iteratorT>
+    bool hasAllValidEntries(const iteratorT begin, const iteratorT end) const
+    {
+        return all_of(begin, end, [](const ConnectedRoad &road) { return road.entry_allowed; });
+    }
```
daniel-j-h: `std::all_of` and not `all_of` - the reason the unqualified `all_of` works is argument-dependant lookup (ADL, great to know, read up a bit about it): if the iterators are in the std namespace `all_of` is searched in the std namespace, too.

Check `std::all_of`s docs: it requires an `InputIt` (make yourself comfortable with the different iterator [concepts](http://en.cppreference.com/w/cpp/iterator)), which basically means you can run over it once. Because you are using `std::all_of` your iterators now also have to be (at least) `InputIterators`. You _could_ assert that at compile time via

    static_assert(std::is_base_of<std::input_iterator_tag, typename std::iterator_traits<Iter>::iterator_category>::value, "")

but what you usually do is to make the template's type just InputIt, as in

    template <typename InputIt> void fn(InputIt first, InputIt last);

(C++ Concepts will change this in the future and enforce the static assert automatically)

Finally, don't hard code `ConnectedRoad` as parameter here: the reason we're having the `EnableOps` CRTP struct is because you don't know if it's a `ConnectedRoad` or a `ViewData`. Make it generic.

src/extractor/guidance/turn_handler.cpp
```diff
@@ -31,6 +51,37 @@ bool isEndOfRoad(const ConnectedRoad &,
            angularDeviation(possible_right_turn.angle, possible_left_turn.angle) >
                2 * NARROW_TURN_ANGLE;
 }
+
+template <typename Iterator>
+Iterator findOutermostForkCandidate(const Iterator start, const Iterator end)
+{
+    const auto outermost = std::adjacent_find(start, end, isOutermostForkCandidate);
```
daniel-j-h: ForwardIt requirement, as explained above

src/extractor/guidance/turn_handler.cpp
```diff
@@ -31,6 +51,37 @@ bool isEndOfRoad(const ConnectedRoad &,
...
+namespace osrm
+{
+namespace extractor
+{
+namespace guidance
+{
+
+// a wrapper to handle road indices of forks at intersections
+TurnHandler::Fork::Fork(Intersection::iterator right, Intersection::iterator left)
+    : right(right), left(left), size((int)(left - right) + 1)
```
daniel-j-h: Why `int`? And why signed at all?

src/extractor/guidance/turn_handler.cpp
```diff
@@ -87,48 +142,37 @@ Intersection TurnHandler::handleTwoWayTurn(const EdgeID via_edge, Intersection i
...
-
-    const auto &road_data = node_based_graph.GetEdgeData(road.eid);
-
     const auto same_name = !util::guidance::requiresNameAnnounced(
-        in_data.name_id, road_data.name_id, name_table, street_name_suffix_table);
-
-    if (turn_is_perfectly_straight && in_data.name_id != EMPTY_NAMEID &&
-        road_data.name_id != EMPTY_NAMEID && same_name)
+        via_data.name_id, road_data.name_id, name_table, street_name_suffix_table);
+    if (turn_is_perfectly_straight && via_data.name_id != EMPTY_NAMEID && same_name)
```
daniel-j-h: Wait, the empty name id check has to happen _before_ the `requiresNameAnnounced` above

src/extractor/guidance/turn_handler.cpp
```diff
@@ -140,8 +184,24 @@ bool TurnHandler::isObviousOfTwo(const EdgeID via_edge,
     return is_much_narrower_than_other;
 }
 
+bool TurnHandler::hasObvious(const EdgeID &via_edge, const Fork &fork) const
+{
+    for (auto road = fork.right; road < fork.left; ++road)
+    {
+        if (isObviousOfTwo(via_edge, *road, *(road + 1)) ||
+            isObviousOfTwo(via_edge, *(road + 1), *road))
```
daniel-j-h: What is +1 here? Is this always safe to do?

src/extractor/guidance/turn_handler.cpp
```diff
@@ -470,125 +527,133 @@ Intersection TurnHandler::assignRightTurns(const EdgeID via_edge,
     return intersection;
 }
 
-std::pair<std::size_t, std::size_t> TurnHandler::findFork(const EdgeID via_edge,
-                                                          const Intersection &intersection) const
+// finds a fork candidate by just looking at the geometry and angle of an intersection
+boost::optional<TurnHandler::Fork>
+TurnHandler::findForkCandidatesByGeometry(Intersection &intersection) const
```
daniel-j-h: Do you ever modify the intersection? If yes, why? If no => const!

src/extractor/guidance/turn_handler.cpp
```diff
@@ -470,125 +527,133 @@ Intersection TurnHandler::assignRightTurns(const EdgeID via_edge,
...
+                   node_based_graph.GetEdgeData(road.eid).road_classification.IsLinkClass();
+        }))
+    {
+        return false;
+    }
+
+    return std::all_of(fork.right, fork.left + 1, [&](ConnectedRoad &base) {
+        const auto base_class = node_based_graph.GetEdgeData(base.eid).road_classification;
+        // check that there is no turn obvious == check that all turns are non-onvious
+        return std::all_of(fork.right, fork.left + 1, [&](ConnectedRoad &compare) {
```
daniel-j-h: Does the second loop have to exclude the road from the first road? Otherwise you're doing a self-comparison?

src/extractor/guidance/turn_handler.cpp
```diff
@@ -470,125 +527,133 @@ Intersection TurnHandler::assignRightTurns(const EdgeID via_edge,
...
+boost::optional<TurnHandler::Fork> TurnHandler::findFork(const EdgeID via_edge,
+                                                         Intersection &intersection) const
+{
+    const auto fork = findForkCandidatesByGeometry(intersection);
+    if (fork)
+    {
+        // makes sure that the fork is isolated from other neighbouring streets on the left and
+        // right side
+        const auto next =
+            (fork->left + 1) == intersection.end() ? intersection.begin() : (fork->left + 1);
```
daniel-j-h: +1 safe to do?

src/extractor/guidance/turn_handler.cpp
```diff
@@ -470,125 +527,133 @@ Intersection TurnHandler::assignRightTurns(const EdgeID via_edge,
...
         const bool separated_at_left_side =
-            angularDeviation(intersection[left].angle,
-                             intersection[(left + 1) % intersection.size()].angle) >= GROUP_ANGLE;
+            angularDeviation(fork->left->angle, next->angle) >= GROUP_ANGLE;
         const bool separated_at_right_side =
-            right > 0 &&
-            angularDeviation(intersection[right].angle, intersection[right - 1].angle) >=
-                GROUP_ANGLE;
+            angularDeviation(fork->right->angle, (fork->right - 1)->angle) >= GROUP_ANGLE;
```
daniel-j-h: same

src/extractor/guidance/turn_handler.cpp
chaupow: will be solved by https://github.com/Project-OSRM/osrm-backend/issues/3457

include/extractor/guidance/intersection.hpp
chaupow: Thanks 🙇

src/extractor/guidance/turn_handler.cpp
chaupow: Oh, maybe you can help me here cause I was having trouble with this part anyways: If I make intersection const, all iterators I can get are const-iterators. The function is supposed to return a Fork struct with iterators to the beginning and end of the fork end those iterators cannot be const-iterators cause later on I want to modify their instructions. The only solution I got working was to make the intersection non-const, thus resulting in that code.

you have any idea how to fix that?

src/extractor/guidance/turn_handler.cpp
chaupow: if it's a self comparison, the inner loop will evaluate `return [...] || compare.eid == base.eid;`

Issue #3263, opened 2016-11-08
Split computing mode and speed values


TheMarex:
Currently we mix computing the [speed value](https://github.com/Project-OSRM/osrm-backend/blob/master/profiles/car.lua#L355-L366), [setting the mode](https://github.com/Project-OSRM/osrm-backend/blob/master/profiles/car.lua#L494-L499) and indirectly determine the accessibility (by calling `return` in a [few odd places](https://github.com/Project-OSRM/osrm-backend/blob/master/profiles/car.lua#L325-L328)).

As a result we determine if we are going to use a way by two criterion:
1. `mode != inaccessible` (for at least one directions)
2. `speed > 0` (for at least one direction that is accessible)

This is complex to track in the profile, because you can never be sure what is computed.

We can improve readability if we split the huge `way_function` into two parts:

1. Determine the accessibility of the way (by setting the mode)
2. Determine the speed of the road

Example:

```
function way_function(way, result)
  set_way_mode(way, result)

  if result.forward_mode != mode.inaccessible then
    result.forward_speed = get_way_speed(result.forward_mode, way)
    result.forward_speed = get_way_forward_speed(result.forward_mode,
                                                 result.backward_speed,
                                                 way)
  end
  if result.backward_mode != mode.inaccessible then
    result.backward_speed = get_way_speed(result.backward_mode, way)
    result.backward_speed = get_way_backward_speed(result.backward_mode,
                                                   result.backward_speed,
                                                   way)
  end
end
```

PR #3266, opened 2016-11-08
Refactor car profile

emiltin:
# Issue
This is first step in an incremental refactoring of the LUA profiles to make them more manageable.
This PR refactors the car profile into smaller functions.

Sub functions typically require access to tag values. Because calling into C++ is expensive, we don't want sub functions to call the C++ get_value_by_key() function if we already fetched a particular tag.
We also don't want to have to manage and pass lists of tags around between functions.

This PR therefore introduces a very simply local tag cache, which is basically just a LUA key-value table. The way function sets up a simple table with the input/output objects, which is then passed around.

This PR also includes a debug.lua file, which can be used to call the LUA way_function from the command line tool, passing some input tags and printing the result. This makes it easier to debug the profile code. There are a few C++ helper functions, which are mocked with empty LUA functions - debugging certain tag processing will not work correctly unless we reimplement these helper functions in LUA, which should be doable.

All tests are green.

Testing this PR on a 200 MB denmark.osm.pbf file I see no change in performance. 

## Tasklist
 - [x] review
 - [ ] adjust for comments

## Requirements / Relations
Saw this issue was just added: #3263.


profiles/lib/tag_cache.lua
```diff
@@ -0,0 +1,35 @@
+TagCache = {}
+
+function TagCache.get(data, key)
+  local v = data.cache[key]
+  if v then
+    if v == '' then
```
emiltin: what's going on here is just that we always return nil in case the tag value is an empty string. this means we don't need to sprinkle the code with checks for empty strings.

profiles/lib/tag_cache.lua
```diff
@@ -0,0 +1,35 @@
...
+      return nil
+    else
+      return v
+    end
+  else
+    v = data.way:get_value_by_key(key)
+    if v == nil then
+      data.cache[key] = ''
```
emiltin: in case the tag value is nil, then we store an empty string, so that next time the tag is requested, we know that we already fetched it

profiles/lib/tag_cache.lua
```diff
@@ -0,0 +1,35 @@
...
+      if v == '' then
+        return nil
+      else
+        return v
+      end
+    end
+  end
+end
+
+function TagCache.set(data,key,value)
```
emiltin: the existing profile sometimes modifies the locally stored tag value, this functions support that mechanism

emiltin:
Next steps could be:
- So similar initial refactoring of the other profiles
- Move parts that are common across profiles into helper files
- improve the logic and flow, eg. issues like #3263.

profiles/lib/tag_cache.lua
TheMarex: Maybe we can move this mechanism to C++ land. I would like to remove this class because it works around broken binding behavior.

@daniel-j-h @MoKob any opinions here?

profiles/lib/tag_cache.lua
TheMarex: Can me maybe change the profile code to not need this functionality anymore? Modifying state that you would assume is constant (like the tagging of the road) can lead to confusion.

profiles/debug.lua
```diff
@@ -0,0 +1,99 @@
+-- Enable calling our lua profile code directly from the lua command line,
```
TheMarex: This is actually a good idea. We could do a real mock of the C++ interface in lua for testing.

profiles/car.lua
```diff
@@ -287,328 +279,444 @@ function way_function (way, result)
...
+    result = result,
+    cache = {},
+  }
+  
+  -- perform each procesing step sequentially.
+  -- most steps can abort processing, meaning the way
+  -- is not routable  
+  
+  if handle_initial_check(data) == false then return end
+  if handle_default_mode(data) == false then return end--
```
TheMarex: Some lines here end in `--` which seems like left-over code?

profiles/car.lua
```diff
@@ -287,328 +279,444 @@ function way_function (way, result)
...
+function way_function(way, result)
+  
+  -- define a table that we can pass around to helper functions
+  -- so they have access to the input/output objects.
+  
+  -- we also use a table as a local cache of tags, to avoid calling
+  -- into C++ more than once for each tag, without having to
+  -- pass lists of already fetches tags around.
+
+  local data = {
```
TheMarex: This might be more of a style thing, but I think passing these members unwrapped to the helper functions like `handle_initial_check(way)` and `handle_speed(way, result)` make it more explicit what is actually used.

profiles/lib/tag_cache.lua
emiltin: yes that would be possible and probably better

profiles/debug.lua
emiltin: all that's really missing is to implement the couple of c++ helper methods in lua

profiles/car.lua
emiltin: yes, i was marking which functions can't (yet) return false.

emiltin:
Looking more at performance, running osrm-extract is around 4% slower, testing on a 200 MB denmark.osm.pbf file. This is due to the parsing step itself being around 7% slower.

The slow down was introduced in the commit https://github.com/Project-OSRM/osrm-backend/pull/3266/commits/2653e75c0ebf28fffdfb970affcbabc8e109c0a8, which adds a LUA wrapper function for getting tags. The purpose of the wrapper is to have a way of keeping track of which tags have already been fetched from C++, so that the individual way_function handlers don't do expensive round-trips to C++ to fetch a tag if it has already been fetched.

I haven't been able to find a way in LUA to avoid this slow down, while still keeping the code split into smaller functions.

A thing we could try is to add a C++ function for fetching a list of tag. The way_function could then pre-populate the local LUA tag cache with tags that are fetched for most way, like highway, bridge, route, oneway, name, ref, etc. That would reduce the number of roundtrips to C++ caused by fetching tags. Or the list of tags to pre-populate could be a global property, and the tags could be passed to the way_function in the first place.

emiltin:
The C++ function get_value_by_key() was using "" as default value for empty tags when fetching from osmium. By using null as defualt I could simplify LUA code in a number of places, including node_function() and get_access_tag().

The LUA TagCache could also be eliminated. Even though it means some tags will be fetched via C++ more than once, it provided a speed up due to less LUA code in the critical path of getting tag values.

Preliminary benchmark on a small dataset indicates that extraction is now faster than the original code.

emiltin:
Here's the time of the parsing step when extracting a 38MB file using a release build, average over 3 runs:

```
master branch:   13.5 s
refactor branch: 12.6 s
```

Measurements on larger files away a fix for #3303.

Anything more you want to test/fix before this is merged?

emiltin:
Results extracting 407 MB norway.osm.pfb file with 0839f1c, compared to master:

master:
[info] Parsing finished after 173.476 seconds
real	5m6.998s

refactor (0839f1c):
[info] Parsing finished after 169.923 seconds
real	5m3.540s

TheMarex:
@mokob @daniel-j-h @danpat anything blocking this from your side? This looks good to me.

profiles/car.lua
```diff
@@ -606,9 +701,45 @@ function way_function (way, result)
...
+-- main entry point for processsing a way
+function way_function(way, result)
+  -- table of temporary values, e.g. computed access status
+  local data = {}
+  
+  -- perform each procesing step sequentially.
+  -- most steps can abort processing, meaning the way
+  -- is not routable  
+  
+  if handle_initial_check(way,result,data) == false then return end
```
MoKob: this function could probably be not any more ambiguous. What is a initial check? What are we filtering here?

profiles/car.lua
```diff
@@ -606,9 +701,45 @@ function way_function (way, result)
...
+  if handle_movables(way,result) == false then return end
+  if handle_service(way,result) == false then return end
+  if handle_oneway(way,result) == false then return end
+  if handle_speed(way,result,data) == false then return end
+  if handle_turn_lanes(way,result) == false then return end
+  if handle_junctions(way,result) == false then return end
+  if handle_startpoint(way,result) == false then return end
+  if handle_restricted(way,result,data) == false then return end
+  if handle_classification(way,result) == false then return end
+  if handle_names(way,result) == false then return end
```
MoKob: I kind of don't like this set-up here. The functions are all less descriptive than what I would hope for. We should (can be a next step) give at least a comment for each describing the general workflow here. Otherwise I see a huge potential for messed up rebases that, at some point later, come back to bite us.

emiltin: let me try to improve the naming and add some comments. would you prefer the comments here, or at the actual functions?

MoKob: The function itself should be more descriptive. Here I would prefer having a single sentence giving you the general idea (as long as the function doesn't tell you).
E.g. if a function is called `extract_names`, I don't need a comment saying "Extracts road names from the way and assigns them", but `handle_initial_check` could benefit a lot from "Check if the way is admissible for navigation" or similar.

src/extractor/scripting_environment_lua.cpp
```diff
@@ -34,7 +34,12 @@ namespace
 template <class T>
 auto get_value_by_key(T const &object, const char *key) -> decltype(object.get_value_by_key(key))
 {
-    return object.get_value_by_key(key, "");
+    auto v = object.get_value_by_key(key);
```
emiltin: Do I need to adjust the C++ formatting here? Not sure how the tidy workflow works.

profiles/car.lua
emiltin: right

emiltin:
I've improved naming a bit and added some comments.

The intention with this PR was to be a first step in regards to refactoring. There's a lot more that can be done in other PRs, like reshuffling and reconsidering the flow. For example, do we really need to handle turn lanes or hov if we're on a ferry?

profiles/debug.lua
TheMarex: Can me move this to a separate PR just to keep the scope here small?

src/extractor/scripting_environment_lua.cpp
TheMarex: Yeah `{` goes on the next line and we try not use short names like `v` but rather `value`.

daniel-j-h: clang-format-3.8 -style=file -i path/to/file.cpp

We can do this for you before merging, if you don't have the tooling for it installed.

daniel-j-h:
Can we get a changelog entry for that - e.g. there may be downstream consumers which now have to adapt files to ship for profiles.

emiltin:
Added a changelog item under 5.5.0 - I assumed this is the place it goes?

PR #3270, opened 2016-11-09
Use single precision values in vector<float> initializer list

oxidase:
# Issue

appveyor fails with "Element '7': conversion from 'double' to 'float' requires a narrowing conversion"
The PR should fix windows build

## Tasklist
 - [x] review
 - [x] adjust for comments


danpat:
Ah, good catch, thanks!

PR #3272, opened 2016-11-09
separate intersection generation and intersection optimization

MoKob:
# Issue

Supporting PR for https://github.com/Project-OSRM/osrm-backend/pull/3211.
Due to us looking at the graph a lot during the creation of an actual intersection, we currently would need to introduce a lot of forward declarations / circular dependencies.

This PR separates the generation of intersections from their optimisation for guidance, removing these circular dependencies.

## Tasklist
 - [x] review
 - [x] adjust for comments

## Requirements / Relations
 Will support https://github.com/Project-OSRM/osrm-backend/pull/3211


include/extractor/guidance/intersection_optimizer.hpp
```diff
@@ -0,0 +1,101 @@
...
+/*
+ * An intersection is a central part in computing guidance decisions. However the model in OSM and
+ * the view we want to use in guidance are not necessarily the same thing. We have to account for
+ * some models that are chosen explicitly in OSM and that don't actually describe how a human would
+ * experience an intersection.
+ *
+ * For example, if a small pedestrian island is located at a traffic light right in the middle of a
+ * road, OSM tends to model the road as two separate ways. A human would consider these two ways a
+ * single road, though. In this optimizer, we try to account for these subtle differences between
+ * OSM data and human perception to improve our decision base for guidance later on.
```
daniel-j-h: So is this really an optimizer of more a normalizer / canonicalizer?

include/extractor/guidance/intersection_optimizer.hpp
```diff
@@ -0,0 +1,101 @@
...
+    IntersectionOptimizer(const util::NodeBasedDynamicGraph &node_based_graph,
+                          const std::vector<extractor::QueryNode> &node_coordinates,
+                          const util::NameTable &name_table,
+                          const SuffixTable &street_name_suffix_table,
+                          const IntersectionGenerator &intersection_generator);
+
+    // The function takes an intersection an converts it to a `perceived` intersection which closer
+    // represents how a human might experience the intersection
+    OSRM_ATTR_WARN_UNUSED
+    Intersection Process(const NodeID node_at_intersection, Intersection intersection) const;
```
daniel-j-h: What's your stance on operator() e.g. in the handlers and having a named single function here?

src/extractor/guidance/turn_analysis.cpp
```diff
@@ -123,7 +126,9 @@ TurnAnalysis::transformIntersectionIntoTurns(const Intersection &intersection) c
 
 Intersection TurnAnalysis::getIntersection(const NodeID from_nid, const EdgeID via_eid) const
 {
-    return intersection_generator(from_nid, via_eid);
+    auto intersection = intersection_generator(from_nid, via_eid);
+    const auto node_at_intersection = node_based_graph.GetTarget(via_eid);
+    return intersection_optimizer.Process(node_at_intersection, std::move(intersection));
```
daniel-j-h: This is what I meant above: operator() for the generator vs Process for the optimizer

daniel-j-h: Should this pattern be abstracted over? That is the pipeline: generator -> optimizer.

(in a way to not again get us into the issues we wanted to avoid here in the first place)

MoKob:
managed to break slip-road lane handling as it seems. Investigating :(

PR #3276, opened 2016-11-09
Add logic to avoid service roads in the car profile using edge weight

jakepruitt:
# Issue

As part of issue #77, I started testing a prototype of avoiding service roads just to see if alternative routes are chosen that are less fast but more driveable. Would love more ideas on other things we should add to the car/bicycle profile while we're at it.

This PR is against the feature/edge_weights PR #2399.

## Tasklist
 - [ ] Get feedback for more things to avoid
 - [ ] review
 - [ ] adjust for comments

## Requirements / Relations
Will hopefully get merged as part of #2399 

cc/ @danpat @MoKob @TheMarex


features/car/weight.feature
```diff
@@ -0,0 +1,52 @@
+@routing @car @weight
+Feature: Car - weights
+
+    Background: Use specific speeds
+        Given the profile "car"
+
+    Scenario: Only routes down service road when that's the destination
+        Given the node map
+            """
+            a  b  c
```
MoKob: just for the fun of it, could you use ascii chars to directly show the roads?

profiles/car.lua
```diff
@@ -606,6 +606,16 @@ function way_function (way, result)
     end
   end
 
+  -- add driveability penalty to service roads
+  if properties.weight_name == 'duration' and highway == "service" then
+    if result.duration > 0 or result.forward_speed > 0 then
+      result.forward_weight_per_meter = result.forward_speed / 100;
```
MoKob: I am a bit confused. Isn't the weight supposed to be high, the moment we want to avoid a road?
Could you explain to me, why are we dividing by 100 here?

jakepruitt: The `weight_per_meter` name is slightly misleading, this represents a value analogous to `speed`, where higher values mean faster roads, lower values mean slower roads.

I definitely felt the vocabulary pain here, and would totally be open to finding different wording.

features/car/weight.feature
jakepruitt: Totally 😄

profiles/car.lua
MoKob: I believe `slightly` is an understatement :(

jakepruitt:
Per @daniel-j-h's comment at #3325, going to add a running list of other things to avoid:

- `hazard=` roads
- `access=private` or `access=destination`
- `HOV` roads
- anything else in the [blacklist](https://github.com/Project-OSRM/osrm-backend/blob/master/profiles/car.lua#L10) we'd like to route too but not through

TheMarex:
Need to close this PR because it is against the wrong branch.

PR #3284, opened 2016-11-10
Speed up osx builds: use Mason + Release

springmeyer:
As we know, OS X builds on travis are slow. Today [a build took 30 minutes](https://travis-ci.org/Project-OSRM/osrm-backend/jobs/174879851). For this build the longest running steps, sorted by time, are below:

 - 754s (npm test)
 - 235s (brew install)
 - 104s (make osrm-extract --jobs=3)
 - 98s (make --jobs=3)
 - 57s (npm install)
 - 47s (./unit_tests/util-tests)
 - 32s (make tests)
 - 32s (make install)

Of those the longest `npm test` is easy to improve upon. We can build in `Release` mode instead of `Debug` to ensure the tests run much faster. This PR therefore moves the `osx` build to `Release` mode.

Also, of those the second longest is also easy to improve upon. Instead of using `brew` we can use mason deps pulled from s3. This PR therefore moves the `osx` build to use mason.

## Tasklist
 - [x] review
 - [x] adjust for comments

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -26,6 +26,8 @@ env:
    - CCACHE_COMPRESS=1
    - CASHER_TIME_OUT=599 # one second less than 10m to avoid 10m timeout error: https://github.com/Project-OSRM/osrm-backend/issues/2742
    - JOBS=4
+   - CCACHE_VERSION=3.3.1
+   - CMAKE_VERSION=3.6.2
 
 matrix:
   fast_finish: true
@@ -39,7 +41,7 @@ matrix:
       addons: &gcc6
         apt:
           sources: ['ubuntu-toolchain-r-test']
-          packages: ['g++-6', 'libbz2-dev', 'libstxxl-dev', 'libstxxl1', 'libxml2-dev', 'libzip-dev', 'lua5.1', 'liblua5.1-0-dev', 'libtbb-dev', 'libgdal-dev', 'libluabind-dev', 'libboost-all-dev', 'ccache']
+          packages: ['g++-6', 'libbz2-dev', 'libstxxl-dev', 'libstxxl1', 'libxml2-dev', 'libzip-dev', 'lua5.1', 'liblua5.1-0-dev', 'libtbb-dev', 'libgdal-dev', 'libluabind-dev', 'libboost-all-dev']
       env: CCOMPILER='gcc-6' CXXCOMPILER='g++-6' BUILD_TYPE='Debug' TARGET_ARCH='x86_64-asan' ENABLE_COVERAGE=ON ENABLE_SANITIZER=ON BUILD_COMPONENTS=ON
 
     - os: linux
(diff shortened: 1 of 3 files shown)


springmeyer:
Currently hitting:

```
undef: _clock_gettime
Undefined symbols for architecture x86_64:
  "_clock_gettime", referenced from:
      boost::interprocess::spin_wait::yield() in lto.o
ld: symbol(s) not found for architecture x86_64
clang-3.8: error: linker command failed with exit code 1 (use -v to see invocation)
```

danpat:
@springmeyer ugh - could this be related to our OSX min version of 10.10 ?

Looking at SO: http://stackoverflow.com/questions/5167269/clock-gettime-alternative-in-mac-os-x
it seems like clock_gettime re-appeared in OSX 10.12.  What version of OSX SDK were the mason binaries built against?  It might be that if they were built in a 10.12 environment, some conditionals inside boost have included a symbol that we don't want on older platforms.

springmeyer:
@danpat - What I'm seeing is that the symbol problem is in `boost::interprocess`. This `mason` package is header only so there are no binaries involved. I don't know if this is related to our "min version". My guess is that it is a boost interprocess header-detection bug and that
https://github.com/boostorg/interprocess/blob/master/include/boost/interprocess/detail/os_thread_functions.hpp#L343 might need patched for 10.12.

springmeyer:
hmm, or we need to set `-no_weak_imports`. Will plan to digest and test https://github.com/Homebrew/homebrew-core/issues/3727 more next week.

springmeyer:
This branch is now working by using `xcode7.3`. Will rebase shortly. Ticketed the `clock_gettime` problem separately: #3297

springmeyer:
> build took 30 minutes

This PR is now building in less than half the time on OSX: 12 minutes. Here are the new timings:

 - 754s (npm test) -> 371s
 - 235s (brew install) -> 32s (to install mason deps during `cmake ../`)
 - 104s (make osrm-extract --jobs=3) ->42s
 - 98s (make --jobs=3) -> 43s
 - 57s (npm install) -> 23s
 - 47s (./unit_tests/util-tests) ->  16s
 - 32s (make tests) -> 43s
 - 32s (make install) -> 6s

springmeyer:
This is ready for review. A summary is:

  - Now building in `Release` on OSX instead of `Debug` for faster test runs
  - Now avoiding using `brew` to install dependencies to save ~4min
    - This exposed a few things that could be improved for both linux and osx:
      - Putting `ccache` from mason on PATH to avoid needing to install globally (since travis calls ccache directly)
      - Dropping the dependency on md5sum like we did in node-osrm: https://github.com/Project-OSRM/node-osrm/pull/258
      - Since we now vendor mason, this PR moves to using the mason client rather than the custom access to mason packages for `ccache` and `cmake`

TheMarex:
While running in `Debug` is always from advantage because of assertions, I can see that the OSX situation is just unmanageable. Hopefully we can revert back to `Debug` one day when travis is a little faster. Otherwise :+1:

springmeyer:
> Hopefully we can revert back to Debug one day when travis is a little faster

👍  Or should we consider adding a `Release+Assertions` mode?

TheMarex:
Yeah good idea. Would also help to actually test that option on travis. :+1:

springmeyer:
> Yeah good idea. Would also help to actually test that option on travis. 👍

Queued up PR in https://github.com/Project-OSRM/osrm-backend/pull/3305.

PR #3286, opened 2016-11-11
Output newlines for each progress indicator when not on a tty.

danpat:
# Issue

With long-running OSRM processing jobs, it's nice to see how far along things have got.
Currently, some of the very slow steps use the `Percent` class to output a sequence of dots and `N%` markers.

Unfortunately, many piped process perform line buffering, so when we've got a multi-hour long `osrm-contract` process running, we get nothing logged until we hit the 100% marker.

This change adds a newline after every progress marker, when the output is *not* a TTY (i.e. redirected to a file, or otherwise piped).  This causes progress markers to be visible in logfiles.  Behaviour when on a TTY is unchanged.

## Tasklist
 - [x] review
 - [x] adjust for comments

### DIFF
--- a/include/util/percent.hpp
+++ b/include/util/percent.hpp
@@ -4,6 +4,8 @@
 #include <atomic>
 #include <iostream>
 
+#include "util/isatty.hpp"
+
 namespace osrm
 {
 namespace util
@@ -71,6 +73,12 @@ class Percent
             {
                 std::cout << ".";
             }
+
+            // When not on a TTY, print newlines after each progress indicator so
+            // so that progress is visible to line-buffered logging systems
+            if (!IsStdoutATTY())
+                std::cout << std::endl;
+
             std::cout.flush();
         }
     }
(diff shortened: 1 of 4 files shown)


include/util/percent.hpp
```diff
@@ -4,6 +4,14 @@
 #include <atomic>
 #include <iostream>
 
+// For isatty()
+#ifdef WIN32
+#include <io.h>
+#else
```
daniel-j-h: It's not a binary Windows vs rest situation.
I would ifdef this on the POSIX.1-2001 standard and provide an else #error always failing.

include/util/percent.hpp
```diff
@@ -71,6 +79,13 @@ class Percent
             {
                 std::cout << ".";
             }
+#ifdef WIN32
+            if (!_isatty(_fileno(stdout)))
+                std::cout << std::endl;
+#else
+            if (!isatty(fileno(stdout)))
+                std::cout << std::endl;
+#endif
```
daniel-j-h: Instead of doing this here (and in other files in osrm-backend) what about a utility header providing a portable isatty function hiding all the platform specifics.

Issue #3287, opened 2016-11-11
Request processing duration in routed logs.


gardster:
Add information about processing duration in osrm-routed logs.

PR #3288, opened 2016-11-11
Add request duration to routed logs.

gardster:
# Issue

resolves https://github.com/Project-OSRM/osrm-backend/issues/3287

## Tasks
- [x] review
- [x] adjusted for comments


TheMarex:
We have some handy timer macros in `util/timing_utils.hpp` that should be used here. General idea is fine. :+1:

gardster:
Ok. PTAL

Issue #3282, opened 2016-11-10
Text and routing issue


boothym:
Noticed a few issues with this route: https://www.openstreetmap.org/directions?engine=osrm_car&route=56.06440%2C-3.23100%3B56.06360%2C-3.22170#map=17/56.06266/-3.22621

1.  Missing translation error - "en-GB.javascripts.directions.instructions.slight_right_with_exit" - I couldn't find the slight_right_with_exit text anywhere in Project-OSRM.
2.  Surely at the roundabout it should say continue straight on/take the 2nd exit, instead of slight right?
3.  The route diverts from the primary A921 road to use the Craigholm Lane service road, which isn't really suitable as a shortcut. It's a narrow one lane road for access to the backs of housing, see street view: https://goo.gl/maps/DSDiQM4RaQ12 Don't know whether that is an issue with the router, or something that can be fixed with the OSM data.

(Apologies for the issue title, didn't know what to write!)

daniel-j-h:
I saw the same translation error today on someone else's phone with German locale. This please open a ticket in https://github.com/openstreetmap/openstreetmap-website/issues where the translation happens.

This is a Roundabout Intersection, see this ticket for an explanation:
https://github.com/Project-OSRM/osrm-backend/pull/2718

(probably should have a different icon)

Here's your route on the demo server.

http://map.project-osrm.org/?z=17&center=56.063193%2C-3.223672&loc=56.064501%2C-3.231493&loc=56.063244%2C-3.223211&hl=en&alt=0

daniel-j-h:
This was discussed on IRC yesterday; translation issue is ticketed at
https://github.com/openstreetmap/openstreetmap-website/issues/1369

Closing here.

boothym:
Just wondering, how often does the debug map update?

danpat:
It updates at the same time as the routing - the view is dynamically rendered by the routing engine, so it changes as soon as the data does.

boothym:
Ok. I notice on the map that two similar roads have different speeds in the debug map: http://map.project-osrm.org/debug/#17.64/56.06277/-3.22708. This no doubt affecting the routing away from the primary road.
This service road is 26km/h - https://www.openstreetmap.org/way/75549910
While this residential road is 16km/h - https://www.openstreetmap.org/way/75550158
Any idea why there is a differences in speeds? Both have lanes=1 and 20mph speed limit.

daniel-j-h:
[...]

PR #3289, opened 2016-11-11
fix getTurnDirection, widen `straight` range, adjust tests accordingly

MoKob:
# Issue

resolves https://github.com/Project-OSRM/osrm-backend/issues/3282

screen shot 2016-11-11 at 11 34 58

Turned out we had some inconsistent `if` conditions [here](https://github.com/Project-OSRM/osrm-backend/blob/12ded539aaba3f730638ab008db1a3ad3c0c345a/include/util/guidance/toolkit.hpp#L53-L57), resulting in this intersection not being identified as `straight` in the first place.

I widened the range for straight by `5` degree as well.

This required the adjustment of a few tests.

## Tasklist
 - [ ] review
 - [ ] adjust for comments

## Requirements / Relations

### DIFF
--- a/features/guidance/collapse.feature
+++ b/features/guidance/collapse.feature
@@ -724,11 +724,11 @@ Feature: Collapse
             | restriction | bc       | dc     | c        | no_right_turn |
 
         When I route I should get
-            | waypoints | route                 | turns                                          |
-            | a,g       | road,cross,cross      | depart,turn left,arrive                        |
-            | a,e       | road,road,road        | depart,continue slight right,arrive            |
+            | waypoints | route                 | turns                                      |
+            | a,g       | road,cross,cross      | depart,turn left,arrive                    |
+            | a,e       | road,road,road        | depart,continue straight,arrive            |
             # We should discuss whether the next item should be collapsed to depart,turn right,arrive.
-            | a,f       | road,road,cross,cross | depart,continue slight right,turn right,arrive |
+            | a,f       | road,road,cross,cross | depart,continue straight,turn right,arrive |
 
     Scenario: On-Off on Highway
         Given the node map
@@ -755,12 +755,13 @@ Feature: Collapse
     Scenario: Don't collapse going straight if actual turn
         Given the node map
             """
-              c e
-                d   f
-
+                e
+             c  |
+              \ d - - - f
+               \|
                 b
-
-
+                |
+                |
                 a
             """
 
@@ -771,10 +772,10 @@ Feature: Collapse
             | df    | right    | residential |
 
         When I route I should get
-            | waypoints | route                     | turns                                  |
-            | a,c       | main,main                 | depart,arrive                          |
-            | a,e       | main,straight,straight    | depart,turn straight,arrive            |
-            | a,f       | main,straight,right,right | depart,turn straight,turn right,arrive |
+            | waypoints | route                     | turns                                  | locations |
+            | a,c       | main,main                 | depart,arrive                          | a,c       |
+            | a,e       | main,straight,straight    | depart,turn straight,arrive            | a,b,e     |
+            | a,f       | main,straight,right,right | depart,turn straight,turn right,arrive | a,b,d,f   |
 
     Scenario: Entering a segregated road
         Given the node map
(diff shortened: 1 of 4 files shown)


features/guidance/collapse.feature
```diff
@@ -755,12 +755,13 @@ Feature: Collapse
...
+             c  |
+              \ d - - - f
+               \|
                 b
-
-
+                |
+                |
                 a
             """
```
MoKob: this tests (due to the move of c) is also changed for turn-angles. The widening of `straight` in `getTurnDirection` would have altered the purpose of the test otherwise, interacting with https://github.com/Project-OSRM/osrm-backend/blob/12ded539aaba3f730638ab008db1a3ad3c0c345a/src/extractor/guidance/turn_handler.cpp#L192-L195

features/guidance/turn-angles.feature
```diff
@@ -773,7 +773,7 @@ Feature: Simple Turns
 
         When I route I should get
             | waypoints | route                   | turns                               |
-            | a,j       | Siemens,Siemens,Siemens | depart,continue slight right,arrive |
+            | a,j       | Siemens,Siemens,Siemens | depart,continue straight,arrive     |
```
MoKob: this change looks weird at first, but if you consider two roads `(a--b)` and `(i--j)` with half a meter offset in between the two, going straight is actually a reasonable resposne here.

MoKob:
There was some formatting missing in the project again. The second commit only assures formatting rules.

daniel-j-h:
Hm I already closed the ticket this was about to resolve - can you explain the widening by 10 degree here in the context of the scenario above? What is the turn angle here and what did we announce before?

MoKob:
The original turn is 165 degree angle, just on the original borders. I found that these 15 degrees on either side of straight seemed a bit narrow, given the above seen turn angle.

We did false announce `slight right`, simply due to the overlap in the if-conditions. However we probably want to have the `straight` which we can see now.

Even though fixing the if-condition alone did solve the issue, I felt that an additional 5 degrees for going straight wouldn't be too bad, since 20 degree turns are most likely felt to be straight anyhow.

boothym:
On a related note, I see that Mapzen indicates the exit number on a roundabout for the same route. Both Mapzen and Google Maps also tell you which road you are continuing onto - is the something being considered for OSRM?
Mapzen: Enter the roundabout and take the 2nd exit.	
	      Exit the roundabout onto Cromwell Road/A921.
Gmaps: "At the roundabout, continue straight onto Cromwell Rd/A921"

daniel-j-h:
Here is the full response we're emitting in osrm-backend for the route shown in the image above:

https://gist.github.com/daniel-j-h/d31afa7c5d95b13ccfbf91402e32ff0d

what to do with it is up to the client.

If you check this section of the response corresponding to the roundabout:

https://gist.github.com/daniel-j-h/d31afa7c5d95b13ccfbf91402e32ff0d#file-response-json-L123-L136

you will see a RoundaboutTurn instruction:

https://github.com/Project-OSRM/osrm-backend/blob/master/docs/http.md#stepmaneuver
> Describes a turn at a small roundabout that should be treated as normal turn. The modifier indicates the turn direciton. Example instruction: At the roundabout turn left.

and both a "name" and a "ref" field "Cromwell Road", "A921".

Looping @freenerd in here for frontend clarification.

daniel-j-h:
Hm looks like the demo server _does_ indeed report this already

http://map.project-osrm.org/?z=18&center=56.064069%2C-3.230147&loc=56.064259%2C-3.230699&loc=56.063834%2C-3.229584&hl=en&alt=0

@MoKob I think your frontend may be outdated.

MoKob:
@boothym regarding https://github.com/Project-OSRM/osrm-backend/pull/3289#issuecomment-260002969, the road entered is part of the route-step response. Showing it or not is a question of the front-end. If you feel it should be allowed, feel free to start a discussion over at osrm-frontend.

freenerd:
@MoKob @boothym 

screen shot 2016-11-14 at 11 34 18

If the exit road is tagged, then we emit the street name in the instruction. We are also [thinking about](https://github.com/Project-OSRM/osrm-text-instructions/issues/41) adding a `stay on` instruction instead.

Issue #3278, opened 2016-11-10
Skip Expensive Metadata PBF Parsing


daniel-j-h:
> The Reader can take an additional optional read_meta flag. If this is set to false the PBF input will ignore metadata on OSM objects (like version, timestamp, uid, ...) which speeds up file reading by 10 to 20%.

https://github.com/osmcode/libosmium/blob/9c0d4bb41d14da3e1d768a2eb14ca27685e59c14/CHANGELOG.md#added

Reader is used here:

https://github.com/Project-OSRM/osrm-backend/blob/cf35bbeb50824b38bad9154b44899000bc3a5cc6/src/extractor/extractor.cpp#L130

Tasks
- [x] Update libosmium
- [x] Pass flag to Reader

PR #3296, opened 2016-11-11
Osmium 2.10

daniel-j-h:
For https://github.com/Project-OSRM/osrm-backend/issues/3278.


daniel-j-h:
Metadata skipping for reader done in https://github.com/Project-OSRM/osrm-backend/pull/3296/commits/0375af197d865ff19de5507e1c3a61a91c29234b.

Needs to be tested on planet.

PR #3305, opened 2016-11-14
Test with assertions+release on travis

springmeyer:
We currently:

  - We do not test building with assertions on in Release mode on travis
  - We do not test OS X debug builds on travis (since they are very slow - refs #3284) and therefore don't test any OS X builds with assertions.

This PR addresses both of these issues by always building with assertions, in both Debug and Release, on travis.

## Tasklist
 - [x] review
 - [x] adjust for comments


springmeyer:
As expected, this noticeably slows down release builds. They are between 1.6-2x slower:

This branch:

screen shot 2016-11-14 at 10 59 23 am

Recent PR merged to master:

screen shot 2016-11-14 at 11 00 52 am

TheMarex:
Ah I was only thinking about OSX with ENABLE_ASSERTIONS=On because we don't have debug builds for those. Sorry for the misunderstanding.

TheMarex:
Fixed the remaining misunderstanding. Merging.

PR #3315, opened 2016-11-15
Refactor/performance improve

MoKob:
# Issue

addressing performance overheads introduced in 5.5

The performance improvements are a result of a series of optimisations.
 - using a low precision coordinate mode (merged previously) for situations where the pure order of turns is enough
 - improved graph traversal (generate intermediate intersections only if necessary)
 - ordering of detectors in coordinate extraction for early exit on cheap tests
 - caching segment lengths during initial extraction to enable early exits on normalisation
 - skipping certain angle adjustments in post-processing when creating turn lane assignment (optimisation unnecessary here, since matching is fuzzy anyhow)
 - small performance bugs (copy instead of reference)

## Tasklist
 - [x] find out what breaks in the single test -> new interpretation of `obvious` results in not discovering end of road anymore
 - [x] review
 - [x] adjust for comments

## Requirements / Relations
  none

### DIFF
--- a/features/guidance/collapse.feature
+++ b/features/guidance/collapse.feature
@@ -338,10 +338,10 @@ Feature: Collapse
         Given the node map
             """
               a f       g
-
-              b e
-
-
+              | |   . '
+              b-e '
+              / /
+             / /
             c d
             """
 
@@ -353,13 +353,13 @@ Feature: Collapse
             | ge    | primary | second | no     |
 
         When I route I should get
-            | waypoints | route               | turns                          |
-            | d,c       | first,first,first   | depart,continue uturn,arrive   |
-            | a,f       | first,first,first   | depart,continue uturn,arrive   |
-            | a,g       | first,second,second | depart,turn left,arrive        |
-            | d,g       | first,second,second | depart,turn right,arrive       |
-            | g,f       | second,first,first  | depart,turn right,arrive       |
-            | g,c       | second,first,first  | depart,end of road left,arrive |
+            | waypoints | route               | turns                        |
+            | d,c       | first,first,first   | depart,continue uturn,arrive |
+            | a,f       | first,first,first   | depart,continue uturn,arrive |
+            | a,g       | first,second,second | depart,turn left,arrive      |
+            | d,g       | first,second,second | depart,turn right,arrive     |
+            | g,f       | second,first,first  | depart,turn right,arrive     |
+            | g,c       | second,first,first  | depart,turn left,arrive      |
 
     Scenario: Do not collapse turning roads
         Given the node map
--- a/src/extractor/guidance/intersection_generator.cpp
+++ b/src/extractor/guidance/intersection_generator.cpp
@@ -99,6 +99,19 @@ Intersection IntersectionGenerator::GetConnectedRoads(const NodeID from_node,
                          from_node, via_eid, traversed_in_reverse, to_node, intersection_lanes);
     };
 
+    // The first coordinate (the origin) can depend on the number of lanes turning onto,
+    // just as the target coordinate can. Here we compute the corrected coordinate for the
+    // incoming edge
+
+    // to compute the length along the path
+    const auto in_segment_length = [&]() {
+        const auto in_coordinates =
+            coordinate_extractor.GetCoordinatesAlongRoad(from_node, via_eid, INVERT, turn_node);
+        return util::coordinate_calculation::getLength(
+            in_coordinates, util::coordinate_calculation::haversineDistance);
+    }();
+    const auto first_coordinate = extract_coordinate(from_node, via_eid, INVERT, turn_node);
+
     for (const EdgeID onto_edge : node_based_graph.GetAdjacentEdgeRange(turn_node))
     {
         BOOST_ASSERT(onto_edge != SPECIAL_EDGEID);
@@ -117,14 +130,7 @@ Intersection IntersectionGenerator::GetConnectedRoads(const NodeID from_node,
             // the turn is not restricted
             !restriction_map.CheckIfTurnIsRestricted(from_node, turn_node, to_node);
 
-        auto angle = 0.;
-        double bearing = 0.;
-
-        // The first coordinate (the origin) can depend on the number of lanes turning onto,
-        // just as the target coordinate can. Here we compute the corrected coordinate for the
-        // incoming edge.
-        const auto first_coordinate = extract_coordinate(from_node, via_eid, INVERT, turn_node);
-
+        double bearing = 0., out_segment_length = 0., angle = 0.;
         if (from_node == to_node)
         {
             bearing = util::coordinate_calculation::bearing(turn_coordinate, first_coordinate);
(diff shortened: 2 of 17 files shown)


features/guidance/collapse.feature
```diff
@@ -353,13 +353,13 @@ Feature: Collapse
...
-            | d,g       | first,second,second | depart,turn right,arrive       |
-            | g,f       | second,first,first  | depart,turn right,arrive       |
-            | g,c       | second,first,first  | depart,end of road left,arrive |
+            | waypoints | route               | turns                        |
+            | d,c       | first,first,first   | depart,continue uturn,arrive |
+            | a,f       | first,first,first   | depart,continue uturn,arrive |
+            | a,g       | first,second,second | depart,turn left,arrive      |
+            | d,g       | first,second,second | depart,turn right,arrive     |
+            | g,f       | second,first,first  | depart,turn right,arrive     |
+            | g,c       | second,first,first  | depart,turn left,arrive      |
```
MoKob: this is due to https://github.com/Project-OSRM/osrm-backend/issues/3365

MoKob:
The impact of this PR:

Master: Germany
Handled: 95613 of 167868 lanes: 56.9573 %.
[info] Timing statistics for edge-expanded graph:
[info] Renumbering edges: 0.267233s
[info] Generating nodes: 14.4786s
[info] Generating edges: 436.623s

Refactor/Performance:
[info] Renumbering edges: 0.247944s
[info] Generating nodes: 14.6859s
[info] Generating edges: 345.221s

MoKob:
@daniel-j-h could you have a look, this is the first step in improving the performance back to reasonable levels

TheMarex:
Can you put a short summary in one sentence of how the performance was optimized in the description? (or link to an issue) This was not immediately clear to me when looking at the changes and it would help during a review.

The numbers look great though. 30% improvement is significant.

MoKob:
I've added a few remarks to the initial ticket. I hope these explanations help to explain the differences we see in processing time.

MoKob:
From experiments, over at https://github.com/Project-OSRM/osrm-backend/pull/3341, most of the speed-up here comes from doing many items over and over again. Sadly, the refactor for better processing order reduces the speed-up we see from the changes in this PR considerably. The speed-up simply due to this PR will remain at only 5% after merging the better processing order.

Still, I'd say right now we can make use of whatever we can get to speed-up extraction again.

src/extractor/guidance/intersection_generator.cpp
```diff
@@ -99,6 +99,20 @@ Intersection IntersectionGenerator::GetConnectedRoads(const NodeID from_node,
...
+    util::Coordinate first_coordinate;
+    double in_segment_length = 0.;
+    std::tie(first_coordinate, in_segment_length) = [&]() {
+        const auto in_coordinates =
+            coordinate_extractor.GetCoordinatesAlongRoad(from_node, via_eid, INVERT, turn_node);
+        const auto length = util::coordinate_calculation::getLength(
+            in_coordinates, util::coordinate_calculation::haversineDistance);
+        const auto coordinate = extract_coordinate(from_node, via_eid, INVERT, turn_node);
+        return std::make_pair(coordinate, length);
+    }();
```
daniel-j-h: Not sure if the tie+lambda makes the code more readable here.. if you want to scope the code in the lambda, what about

    {   // artificially scoped
        const auto tmp = fn();
        ..
    }

src/extractor/guidance/intersection_generator.cpp
```diff
@@ -150,14 +157,22 @@ Intersection IntersectionGenerator::GetConnectedRoads(const NodeID from_node,
...
-                extract_coordinate(turn_node, onto_edge, !INVERT, to_node);
+            util::Coordinate third_coordinate;
+            std::tie(third_coordinate, out_segment_length) = [&]() {
+                const auto out_coordinates = coordinate_extractor.GetCoordinatesAlongRoad(
+                    turn_node, onto_edge, !INVERT, to_node);
+                const auto length = util::coordinate_calculation::getLength(
+                    out_coordinates, util::coordinate_calculation::haversineDistance);
+                const auto coordinate = extract_coordinate(turn_node, onto_edge, !INVERT, to_node);
+                return std::make_pair(coordinate, length);
+            }();
```
daniel-j-h: same here

src/extractor/guidance/intersection_generator.cpp
```diff
@@ -227,41 +243,39 @@ IntersectionGenerator::GetActualNextIntersection(const NodeID starting_node,
...
+    };
 
-    // using a maximum lookahead, we make sure not to end up in some form of loop
     std::unordered_set<NodeID> visited_nodes;
-    while (visited_nodes.count(node_at_intersection) == 0 &&
-           (result.size() == 2 &&
-            node_based_graph.GetEdgeData(via_edge).IsCompatibleTo(
-                node_based_graph.GetEdgeData(result[1].eid))))
+    while (0 == visited_nodes.count(query_node) &&
+           2 == node_based_graph.GetOutDegree(node_based_graph.GetTarget(query_edge)))
```
daniel-j-h: So this loops until we detect a cycle or encounter a non-artificial node?

src/extractor/guidance/intersection_normalizer.cpp
```diff
@@ -363,11 +363,14 @@ Intersection IntersectionNormalizer::AdjustForJoiningRoads(const NodeID node_at_
     if (intersection.size() <= 1)
         return intersection;
 
-    const util::Coordinate coordinate_at_intersection = node_coordinates[node_at_intersection];
     // never adjust u-turns
     for (std::size_t index = 1; index < intersection.size(); ++index)
     {
         auto &road = intersection[index];
+        // only consider roads that are close
+        if( road.segment_length && *(road.segment_length) > 30 )
```
daniel-j-h: clang-format all files

daniel-j-h: "close" ~ 30m?

src/extractor/guidance/turn_discovery.cpp
```diff
@@ -55,6 +55,13 @@ bool findPreviousIntersection(const NodeID node_v,
     // (looking at the reverse direction).
     const auto node_w = node_based_graph.GetTarget(via_edge);
     const auto u_turn_at_node_w = intersection[0].eid;
+    BOOST_ASSERT(node_based_graph.BeginEdges(node_w) <= u_turn_at_node_w &&
+                 u_turn_at_node_w <= node_based_graph.EndEdges(node_w));
```
daniel-j-h: Care to explain this assertion to me? The intersection edge has to be an element of `node_w`'s edges?

Shouldn't the uturn edge at the intersection be the via edge from above?

MoKob: different directions. one is from the node prior to the intersection (node_v to node_w) and the uturn is the other way around (node_w to node_v)

src/extractor/guidance/intersection_generator.cpp
```diff
@@ -99,6 +99,19 @@ Intersection IntersectionGenerator::GetConnectedRoads(const NodeID from_node,
                          from_node, via_eid, traversed_in_reverse, to_node, intersection_lanes);
     };
 
+    // The first coordinate (the origin) can depend on the number of lanes turning onto,
+    // just as the target coordinate can. Here we compute the corrected coordinate for the
+    // incoming edge
```
daniel-j-h: :+1:

PR #3316, opened 2016-11-15
Add cache into scripting_environment (new behavior).

fijemax:
#3180

@daniel-j-h,
This is my new cache system behavior. (I use the nodes saved in the ExtractionContainers).
I splited the read in two parts. Node reader and all other geometry reader.
Between this two loop, I feed a cache who contain all vector_id of node contains in ExtractionContainers for each osm_id.
I have always a problem to access to the cache in luabind. That is the reason the "ExtractionContainers" reference is static into LuaScriptingEnvironment class. I don't know what is the right way to do this.

One last thing, I will implement a lazy loader to feed the cache only when user need to access at "location" into lua script. What do you think about that ?

### DIFF
--- a/include/extractor/extraction_containers.hpp
+++ b/include/extractor/extraction_containers.hpp
@@ -16,7 +16,7 @@ namespace osrm
 {
 namespace extractor
 {
-
+class ScriptingEnvironment;
 /**
  * Uses external memory containers from stxxl to store all the data that
  * is collected by the extractor callbacks.
@@ -60,6 +60,10 @@ class ExtractionContainers
     STXXLRestrictionsVector restrictions_list;
     STXXLWayIDStartEndVector way_start_end_id_list;
     std::unordered_map<OSMNodeID, NodeID> external_to_internal_node_id_map;
+    // FIXME This map store the vector id for each osm_id contain in "all_nodes_list"
+    // FIXME add a lazy loading to init this cache when user need location on noderef
+    std::unordered_map<OSMNodeID, osmium::Location> nodeCache;
+
     unsigned max_internal_node_id;
 
     ExtractionContainers();
@@ -68,6 +72,9 @@ class ExtractionContainers
                      const std::string &output_file_name,
                      const std::string &restrictions_file_name,
                      const std::string &names_file_name);
+
+    void prepareCache();
+    void clearCache();
 };
 }
 }
--- a/include/extractor/scripting_environment_lua.hpp
+++ b/include/extractor/scripting_environment_lua.hpp
@@ -2,8 +2,8 @@
 #define SCRIPTING_ENVIRONMENT_LUA_HPP
 
 #include "extractor/scripting_environment.hpp"
-
 #include "extractor/raster_source.hpp"
+#include "extractor/extraction_containers.hpp"
 
 #include "util/lua_util.hpp"
 
@@ -67,6 +67,12 @@ class LuaScriptingEnvironment final : public ScriptingEnvironment
                     tbb::concurrent_vector<std::pair<std::size_t, ExtractionWay>> &resulting_ways,
                     tbb::concurrent_vector<boost::optional<InputRestrictionContainer>>
                         &resulting_restrictions) override;
+
+    void setupCache(std::unordered_map<OSMNodeID, osmium::Location> &cache) override;
+    // FIXME This is static because I don't know how to access to this in lua nodeRef class binding.
+    //static ExtractionContainers *extraction_containers;
+    //osmium::Location getLocation(osmium::NodeRef &nref);
+    static std::unordered_map<OSMNodeID, osmium::Location> nodeCache;
 
   private:
     void InitContext(LuaScriptingContext &context);
--- a/src/extractor/extraction_containers.cpp
+++ b/src/extractor/extraction_containers.cpp
@@ -136,6 +136,22 @@ void ExtractionContainers::FlushVectors()
     way_start_end_id_list.flush();
 }
 
+void ExtractionContainers::prepareCache()
+{
+    nodeCache.reserve(all_nodes_list.size());
+    for(size_t i = 0; i < all_nodes_list.size(); i++)
+    {
+        //ExternalMemoryNode data = all_nodes_list[i];
+        osmium::Location location(static_cast<double>(util::toFloating(all_nodes_list[i].lon)), static_cast<double>(util::toFloating(all_nodes_list[i].lat)));
+        nodeCache.insert({OSMNodeID{static_cast<std::uint64_t>(all_nodes_list[i].node_id)}, location});
+    }
+}
+
+void ExtractionContainers::clearCache()
+{
+    nodeCache.clear();
+}
+
 /**
  * Processes the collected data and serializes it.
  * At this point nodes are still referenced by their OSM id.
--- a/src/extractor/extractor.cpp
+++ b/src/extractor/extractor.cpp
@@ -197,6 +197,8 @@ int Extractor::run(ScriptingEnvironment &scripting_environment)
         }
         nodeReader.close();
 
+        extraction_containers.prepareCache();
+
         // WAY AND RELATION PARSING
         osmium::io::Reader wayReader(input_file, osmium::osm_entity_bits::way | osmium::osm_entity_bits::relation, osmium::io::read_meta::no);
         while (const osmium::memory::Buffer buffer = wayReader.read())
@@ -231,6 +233,9 @@ int Extractor::run(ScriptingEnvironment &scripting_environment)
             }
         }
         TIMER_STOP(parsing);
+
+        // Clear cache data
+        extraction_containers.clearCache();
 
         util::SimpleLogger().Write() << "Parsing finished after " << TIMER_SEC(parsing)
                                      << " seconds";
--- a/src/extractor/extractor.cpp
+++ b/src/extractor/extractor.cpp
@@ -198,6 +198,7 @@ int Extractor::run(ScriptingEnvironment &scripting_environment)
         nodeReader.close();
 
         extraction_containers.prepareCache();
+        scripting_environment.setupCache(extraction_containers.nodeCache);
 
         // WAY AND RELATION PARSING
         osmium::io::Reader wayReader(input_file, osmium::osm_entity_bits::way | osmium::osm_entity_bits::relation, osmium::io::read_meta::no);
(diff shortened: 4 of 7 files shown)


danpat:
@fijemax Do you have any before/after timing information for this change?  The two-pass approach is the only way to reliably build the node cache - how much extra time does it add?

fijemax:
hi @danpat,
I just realized this test on andorra data source.
On master, whit two pass approach and two pass approch with node cache feeding.

images

daniel-j-h:
Could you do the same benchmark on larger extracts from
http://www.geofabrik.de/data/download.html
the Andorra dataset is quite small and not very telling.

src/extractor/extraction_containers.cpp
```diff
@@ -136,6 +136,29 @@ void ExtractionContainers::FlushVectors()
     way_start_end_id_list.flush();
 }
 
+void ExtractionContainers::prepareCache()
+{
+    nodeCache.reserve(all_nodes_list.size());
+    for(int i = 0; i < all_nodes_list.size(); i++)
```
daniel-j-h: `int` will overflow for the planet (and even smaller extracts) - make counter `size_t`

src/extractor/extractor.cpp
```diff
@@ -127,8 +127,8 @@ int Extractor::run(ScriptingEnvironment &scripting_environment)
         auto extractor_callbacks = std::make_unique<ExtractorCallbacks>(extraction_containers);
 
         const osmium::io::File input_file(config.input_path.string());
-        osmium::io::Reader reader(input_file, osmium::io::read_meta::no);
-        const osmium::io::Header header = reader.header();
+        osmium::io::Reader nodeReader(input_file, osmium::osm_entity_bits::node);
```
daniel-j-h: `read_metadata::no` still needs to be set, otherwise the reader will parse version, user id etc. on all nodes, ways, etc.

src/extractor/extractor.cpp
```diff
@@ -194,6 +194,33 @@ int Extractor::run(ScriptingEnvironment &scripting_environment)
...
                     result.second);
             }
+        }
+        nodeReader.close();
+
+        // Prepare cache data
+        extraction_containers.prepareCache();
+
+        // WAY AND RELATION PARSING
+        osmium::io::Reader wayReader(input_file, osmium::osm_entity_bits::way | osmium::osm_entity_bits::relation);
```
daniel-j-h: same here, `read_metadata::no` needs to be passed

this flag alone accounts for ~10-20 performance while parsing nodes, ways, etc.

src/extractor/scripting_environment_lua.cpp
```diff
@@ -200,10 +208,8 @@ void LuaScriptingEnvironment::InitContext(LuaScriptingContext &context)
          luabind::class_<osmium::WayNodeList>("WayNodeList").def(luabind::constructor<>()),
          luabind::class_<osmium::NodeRef>("NodeRef")
              .def(luabind::constructor<>())
-             // Dear ambitious reader: registering .location() as in:
-             // .def("location", +[](const osmium::NodeRef& nref){ return nref.location(); })
-             // will crash at runtime, since we're not (yet?) using libosnmium's
-             // NodeLocationsForWays cache
+             .def("location", +[](osmium::NodeRef &nref){
+                 return extraction_containers->getLocationForNode(nref);})
```
daniel-j-h: Instead of a static var, you can make a ScriptingEnvironment-local function (see e.g. lonToDouble at the top) and register it via `&my_function` with Luabind.

fijemax:
@daniel-j-h 
New benchmark on larger extracts : ile-de-france-latest.osrm
capture du 2016-11-17 14-38-54

include/extractor/extraction_containers.hpp
```diff
@@ -42,6 +43,14 @@ class ExtractionContainers
     void WriteEdges(std::ofstream &file_out_stream) const;
     void WriteCharData(const std::string &file_name);
 
+    struct ComparatorGreater
+    {
+        bool operator () (const int & a, const int & b) const
+        { return a > b; }
+        static int max_value()
+        { return std::numeric_limits<int>::min(); }
```
fijemax: This code useless I forgot to remove sry.

include/extractor/extraction_containers.hpp
```diff
@@ -60,6 +69,11 @@ class ExtractionContainers
     STXXLRestrictionsVector restrictions_list;
     STXXLWayIDStartEndVector way_start_end_id_list;
     std::unordered_map<OSMNodeID, NodeID> external_to_internal_node_id_map;
+
+    // FIXME This map store the vector id for each osm_id contain in "all_nodes_list"
+    // FIXME add a lazy loading to init this cache when user need location on noderef
+    std::unordered_map<OSMNodeID, NodeID> nodeCache;
```
fijemax: This code useless I forgot to remove sry.

include/extractor/scripting_environment_lua.hpp
```diff
@@ -68,6 +68,11 @@ class LuaScriptingEnvironment final : public ScriptingEnvironment
                     tbb::concurrent_vector<boost::optional<InputRestrictionContainer>>
                         &resulting_restrictions) override;
 
+    void setExtractionContainers(ExtractionContainers &extractor) override;
+
+    // FIXME This is static because I don't know how to access to this in lua nodeRef class binding.
+    static ExtractionContainers *extraction_containers;
```
daniel-j-h: But why do we need the full extraction containers here.
The ScriptingEnvironment only needs to access the cache, right?

src/extractor/scripting_environment_lua.cpp
```diff
@@ -47,6 +49,11 @@ template <class T> double lonToDouble(T const &object)
     return static_cast<double>(util::toFloating(object.lon));
 }
 
+template <class T> osmium::Location locationForNode(T const &object)
+{
+        return LuaScriptingEnvironment::extraction_containers->getLocationForNode(object);
+}
```
daniel-j-h: The reason this only works with a static var is the locationForNode function needs to be a member function as in

    struct ScriptingEnv {

    Location getLocation(Node n) {
      return cache.at(n);
    }

      Cache cache;
    }

src/extractor/scripting_environment_lua.cpp
```diff
@@ -47,6 +49,11 @@ template <class T> double lonToDouble(T const &object)
     return static_cast<double>(util::toFloating(object.lon));
 }
 
+template <class T> osmium::Location locationForNode(T const &object)
+{
+        return LuaScriptingEnvironment::extraction_containers->getLocationForNode(object);
```
fijemax: I create this function and i pass it to luabind but into i can't acces to extraction_containers if it's not a static var : 
 error: invalid use of non-static data member ‘osrm::extractor::LuaScriptingEnvironment::extraction_containers’

include/extractor/scripting_environment_lua.hpp
```diff
@@ -68,7 +68,14 @@ class LuaScriptingEnvironment final : public ScriptingEnvironment
                     tbb::concurrent_vector<boost::optional<InputRestrictionContainer>>
                         &resulting_restrictions) override;
 
+    void setExtractionContainers(ExtractionContainers &extractor) override;
+
+    // FIXME This is static because I don't know how to access to this in lua nodeRef class binding.
+    osmium::Location getLocation(osmium::NodeRef &nref);
```
fijemax: @daniel-j-h  I try this,
Now ExtractionContainers is non static and code compile.
But when i try to access to location in lua i have this error : 

> maxime@lemurien:~/DEV/osrm/osrm-backend$ build/osrm-extract ../pbf-data/andorra-latest.osm.pbf -t 1
> [info] Using script profile.lua
> [info] Input file: andorra-latest.osm.pbf
> [info] Profile: profile.lua
> [info] Threads: 1
> [STXXL-MSG] STXXL v1.4.1 (prerelease/Debug)
> [STXXL-ERRMSG] Warning: no config file found.
> [STXXL-ERRMSG] Using default disk configuration.
> [STXXL-MSG] Disk '/var/tmp/stxxl' is allocated, space: 1000 MiB, I/O implementation: syscall autogrow delete_on_exit queue=0 devid=0
> [info] Parsing in progress..
> [info] input file generated by Osmium (http://wiki.openstreetmap.org/wiki/Osmium)
> [info] timestamp: 2016-11-06T20:28:03Z
> [info] Found 3 turn restriction tags:
> [info]   motorcar
> [info]   motor_vehicle
> [info]   vehicle
> way id : 	5203906
>   node id  : 	3655224911
> terminate called after throwing an instance of 'osrm::util::exception'
>   what():  ERROR occurred in profile script:
> No matching overload found, candidates:
> Location location(custom [N4osrm9extractor23LuaScriptingEnvironmentE]&,NodeRef&)
> Abandon (core dumped)

fijemax: I think this is because, luabind try to call getLocation function with a NodeRef instance. Am I Right ?
Is that the way you asked me to do it?

fijemax:
Hi @daniel-j-h,
I created a function : 
`void SetupCache(unordered_map<Node, Location> cache);`
to set cache into ScriptingEnvironment.

I have 3 questions : 
 - About cache, it's always a static variable (but is not longer a reference). I really don't know how to access to a this variable if she isn't static, this variable can't stay static ?
 - I would like implement a "lazy loader" to load cache only when user need to access at location into Lua script.  How to you think of that ?
 - To finish if : 
`nodeCache.at(i)`
return a null value, how i can warn Lua that the location don't exist ?

daniel-j-h:
@fijemax if the variable is not static you have to pass an object to Lua on which you can access the variable. Regarding the user accessing the cache in the profiles if there is none: you can just assert that this never happens (=> user error) and shut down / crash hard if the user still does this.

What I'm seeing blocking us from merging this upstream is besides some rough edges the performance decrease due to the two-pass approach: 5% without cache usage and 15% with cache usage is quite a big hit for a feature most users probably don't need.

I'm especially talking about running this on planet scale workloads.

fijemax:
I test a new behavior who doesn't need the two-pass approach, performance decrease 0% without cache and 13% with cache feeding. But if we feed the cache only when lua intent to access at location all other users will not be impacted by this changes.
I would like to show you this evolution, I can push this on this pull request or I open a new pull request ?

daniel-j-h:
Sounds great - yes feel free to open another pull request and maybe close this one then.

If a user doesn't need the location cache she shouldn’t have to pay for it.

fijemax:
A slightly different behavior is getting explored in #3358 by @fijemax - closing here.

PR #3318, opened 2016-11-15
Fix ubsan runtime error: left shift of negative value -9

### LINKED ISSUE #3222: Polyline Compressor Left Shift Undefined Behavior
Ubsan reports we're left-shifting -9 here (undefined behavior) - can be reproduced by running the tests.

https://github.com/Project-OSRM/osrm-backend/blob/c30f43b148ddcfc8c2626dde29f718a0e9e0eb9f/src/engine/polyline_compressor.cpp#L45

oxidase:
# Issue

Fixes #3222, a left shift of signed integer is undefined due to 5.8.2
Now unsigned 2's complement is shifted, inverted and statically casted to int.

Added also a check to avoid another UB in `m_lane_tuple_id_pairs[0]` if m_lane_tuple_id_pairs is empty.

## Tasklist
 - [x] review
 - [x] adjust for comments

### DIFF
--- a/src/engine/polyline_compressor.cpp
+++ b/src/engine/polyline_compressor.cpp
@@ -33,20 +33,16 @@ std::string encode(std::vector<int> &numbers)
     std::string output;
     for (auto &number : numbers)
     {
-        bool isNegative = number < 0;
-
-        if (isNegative)
+        if (number < 0)
         {
             const unsigned binary = std::llabs(number);
             const unsigned twos = (~binary) + 1u;
-            number = twos;
+            const unsigned shl = twos << 1u;
+            number = static_cast<int>(~shl);
         }
-
-        number <<= 1u;
-
-        if (isNegative)
+        else
         {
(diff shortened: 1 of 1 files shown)


daniel-j-h:
Looks good, thanks!

oxidase:
Just rebased to master, so now be6c4c3 is not needed, but there is alignment issues like
```
osrm-backend/include/storage/io.hpp:170:24: runtime error: reference binding to misaligned address 0x7ffdc52db594 for type 'long unsigned int', which requires 8 byte alignment
0x7ffdc52db594: note: pointer points here
  75 09 4c eb 65 00 00 00  00 00 00 00 00 00 00 00  00 00 00 00 00 00 00 00  fb 00 00 00 00 00 00 00
              ^ 
```

danpat:
Other UB ticketed in https://github.com/Project-OSRM/osrm-backend/issues/3326, this change LGTM.

PR #3319, opened 2016-11-16
Make rounding when computing PhantomNode weight symmetric

### LINKED ISSUE #3038: Different durations returned for reverse and forward routes

### LINKED ISSUE #3088: Windows builds are broken

TheMarex:
# Issue

Fixes #3038.

## Tasklist
 - [x] review
 - [x] adjust for comments

### DIFF
--- a/include/engine/geospatial_query.hpp
+++ b/include/engine/geospatial_query.hpp
@@ -400,7 +400,8 @@ template <typename RTreeT, typename DataFacadeT> class GeospatialQuery
         }
         if (data.reverse_segment_id.id != SPECIAL_SEGMENTID)
         {
-            reverse_weight *= 1.0 - ratio;
+            const EdgeWeight difference = reverse_weight * ratio;
+            reverse_weight -= difference;
         }
 
         auto transformed = PhantomNodeWithDistance{PhantomNode{data,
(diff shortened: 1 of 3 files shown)


include/engine/geospatial_query.hpp
```diff
@@ -400,7 +400,8 @@ template <typename RTreeT, typename DataFacadeT> class GeospatialQuery
         }
         if (data.reverse_segment_id.id != SPECIAL_SEGMENTID)
         {
-            reverse_weight *= 1.0 - ratio;
+            EdgeWeight difference = reverse_weight * ratio;
```
MoKob: can be `const`.

MoKob:
Minor nitpick, otherwise fine.

daniel-j-h:
- [ ] needs to adapt Cucumber tests to see the fix in action.

Issue #3256, opened 2016-11-07
Constrain Staggered Intersection Classification


daniel-j-h:
Breaking out of https://github.com/Project-OSRM/osrm-backend/pull/3255#discussion_r86793694.

[Staggered Intersection classification](https://github.com/Project-OSRM/osrm-backend/blob/3e18e31bc916e711f20f7502c5b0f68fbc6c00bb/src/engine/guidance/post_processing.cpp#L806-L846) should be constrained to handle exactly the very short zig-zags in rural areas as described in #2738 - not more, not less.

This means e.g. checking for
- ~~mode changes~~ (resoled by #3324 )
- road classes
- number of lanes to derive road width etc.

Basically all information available in a [RouteStep](https://github.com/Project-OSRM/osrm-backend/blob/3e18e31bc916e711f20f7502c5b0f68fbc6c00bb/include/engine/guidance/route_step.hpp#L57-L74) which makes sense.

To fully solve this, we need to move the check for staggered intersections to its own location (prior to name changes?) instead of tagging it onto other scenarios. Code duplication that might become necessary would be the target of collapse refactoring.

When fully solved, cases like: https://github.com/Project-OSRM/osrm-backend/issues/3376 should be possible to handle.

karenzshea:
Started in https://github.com/Project-OSRM/osrm-backend/tree/constrain-staggering

daniel-j-h:
Hm road class is not available there. What you can check is the [Intersection](https://github.com/Project-OSRM/osrm-backend/blob/3e18e31bc916e711f20f7502c5b0f68fbc6c00bb/include/engine/guidance/route_step.hpp#L73) array in a `RouteStep`: intersections[0] corresponds to the `StepManeuver`, everything else describes intersections on the way. There should be no additional intersections in between the two steps.

TheMarex:
This was merged. :tada:

karenzshea:
No, @MoKob feels that the scope of this ticket is wider than what was addressed in [this pr](https://github.com/Project-OSRM/osrm-backend/issues/3324https://github.com/Project-OSRM/osrm-backend/issues/3324) so we're leaving open for anyone who can tackle the rest.

karenzshea:
Criteria for staggered intersections addressed in the last PR were mode changes and intermediary restrictions. Everything else in the OP is still open.

MoKob:
Moving remaining items to 5.6

TheMarex:
[...]

PR #3324, opened 2016-11-16
Constrain staggered intersection detection by mode change and intermediary intersections

karenzshea:
# Issue

 https://github.com/Project-OSRM/osrm-backend/issues/3256

Ideally would've also been able to return pertinent information about whether road classes have changed between two staggered intersections, but we currently don't have road classification available during post-processing.

## Tasklist
    - [x] more space between roads, currently we are modelling data issues
    - [ ] add asymmetric scenarios
 - [x] investigate interaction of mode changes (currently we only check two of the participating three mode changes (in-segment, mid-segment (three meters), end (after staggered)). We should check all three, not relying on implicit success due to other locations).  
 - [ ] At some point, we should make sure which mode changes we might still want to collapse -> create a dedicated issue
 - [ ] review
 - [ ] adjust for comments

## Requirements / Relations
cc @daniel-j-h @MoKob

### DIFF
--- a/features/guidance/bike-staggered-intersections.feature
+++ b/features/guidance/bike-staggered-intersections.feature
@@ -24,9 +24,9 @@ Feature: Staggered Intersections
             | ihedcj | residential | Cedar Dr | yes    |
 
         When I route I should get
-            | waypoints | route         | turns         |
-            | a,g       | Oak St,Oak St | depart,arrive |
-            | g,a       | Oak St,Oak St | depart,arrive |
+            | waypoints | route                         | turns                              | modes                                |
+            | a,g       | Oak St,Cedar Dr,Oak St,Oak St | depart,turn right,turn left,arrive | cycling,pushing bike,cycling,cycling |
+            | g,a       | Oak St,Oak St                 | depart,arrive                      | cycling,cycling                      |
 
     Scenario: Staggered Intersection - pushing at start
         Given the node map
@@ -46,9 +46,9 @@ Feature: Staggered Intersections
             | ihedcj | residential | Cedar Dr |        |
 
         When I route I should get
-            | waypoints | route         | turns         |
-            | a,g       | Oak St,Oak St | depart,arrive |
-            | g,a       | Oak St,Oak St | depart,arrive |
+            | waypoints | route                         | turns                              | modes                                |
+            | a,g       | Oak St,Cedar Dr,Oak St,Oak St | depart,turn right,turn left,arrive | pushing bike,cycling,cycling,cycling |
+            | g,a       | Oak St,Oak St                 | depart,arrive                      | cycling,cycling                      |
 
     Scenario: Staggered Intersection - pushing at end
         Given the node map
@@ -68,9 +68,9 @@ Feature: Staggered Intersections
             | ihedcj | residential | Cedar Dr |        |
 
         When I route I should get
-            | waypoints | route         | turns         |
-            | a,g       | Oak St,Oak St | depart,arrive |
-            | g,a       | Oak St,Oak St | depart,arrive |
+            | waypoints | route         | turns                                              | modes                                     |
+            | a,g       | Oak St,Cedar Dr,Oak St,Oak St | depart,turn right,turn left,arrive | cycling,cycling,pushing bike,pushing bike |
+            | g,a       | Oak St,Oak St | depart,arrive                                      | cycling,cycling                           |
 
     Scenario: Staggered Intersection - pushing at start and end
         Given the node map
@@ -90,6 +90,6 @@ Feature: Staggered Intersections
             | ihedcj | residential | Cedar Dr |        |
 
         When I route I should get
-            | waypoints | route         | turns         |
-            | a,g       | Oak St,Oak St | depart,arrive |
-            | g,a       | Oak St,Oak St | depart,arrive |
+            | waypoints | route                         | turns                              | modes                                          |
+            | a,g       | Oak St,Cedar Dr,Oak St,Oak St | depart,turn right,turn left,arrive | pushing bike,cycling,pushing bike,pushing bike |
+            | g,a       | Oak St,Oak St                 | depart,arrive                      | cycling,cycling                                |
--- a/features/guidance/bike-staggered-intersections.feature
+++ b/features/guidance/bike-staggered-intersections.feature
@@ -1,0 +1,95 @@
+@routing  @guidance @staggered-intersections
+Feature: Staggered Intersections
+
+    Background:
+        Given the profile "bicycle"
+        Given a grid size of 1 meters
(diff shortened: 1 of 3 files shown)


features/guidance/staggered-intersections.feature
```diff
@@ -13,7 +13,7 @@ Feature: Staggered Intersections
                 j
             a b c
                 d
-                e f g
+                e f  g
```
daniel-j-h: Why the change here?

features/guidance/staggered-intersections.feature
```diff
@@ -98,3 +98,48 @@ Feature: Staggered Intersections
...
+                e  f g
+                h
+            """
+
+        And the ways
+            | nodes  | highway | name     | route |
+            | abc    | primary | to_sea   |       |
+            | ef     |         | to_sea   | ferry |
+            | fg     | primary | road     |       |
+            | jcdeh  | primary | road     |       |
```
daniel-j-h: Only ef is a ferry here - what happens if
- both start (abc) and end (efg) are ferries
- only the middle part is a ferry (cde)

features/guidance/staggered-intersections.feature
```diff
@@ -98,3 +98,48 @@ Feature: Staggered Intersections
...
+
+    Scenario: Staggered Intersection: do not collapse intermediary intersections
+        Given the node map
+            """
+                j
+            a b c
+                e f g
+                k l m
+                i
+            """
```
daniel-j-h: Perfect, this is exactly what we should guard against!

daniel-j-h: Hm also what I just realized, think about this scenario

```
    d e f
a b c
    g h i
```

and a route a,g.

Because we're only working on RouteSteps in post-processing we're detecting a Staggered Intersection here and collapse it away emitting nothing at c. There is no way to get information about the second ambiguous way cdef.

To handle this correctly Staggered Intersections have to be detected in pre-processing and be a separate (internal) instruction. Thoughts @MoKob?

daniel-j-h: Also @karenzshea can you add a test case for this (scenario above, should not be collapsed) and tag it `todo` so it does not fail by default.

karenzshea: :+1:

MoKob: Hm. If both are modelled like this, it would probably be possible to detect via turn-angles. If `gcd` is a through street with additional nodes on top and below, we are not able to detect it.
We could argue, however, that if we should find two different roads withing 6 meters of each other, then we are probably talking about such an extreme edge case that we can ignore it.

features/guidance/staggered-intersections.feature
```diff
@@ -143,3 +143,21 @@ Feature: Staggered Intersections
             | a,m       | Oak St,Cedar Dr,Oak St,Oak St | depart,turn right,turn left,arrive |
             | m,a       | Oak St,Cedar Dr,Oak St,Oak St | depart,turn right,turn left,arrive |
 
+    Scenario: Staggered Intersection:
```
karenzshea: @daniel-j-h does this test model what you had in mind? I can sort of imagine the collapsed silent scenario you were describing, but I can only get it to recognize the end `e` point here as part of the middle `jce` way, and not the branching `efg` way.

MoKob:
I'v added some further tasks to the original part of the PR

features/guidance/staggered-intersections.feature
```diff
@@ -98,3 +98,66 @@ Feature: Staggered Intersections
...
             | a,g       | Oak St,Cedar Dr,Elm St,Elm St | depart,turn right,turn left,arrive |
             | g,a       | Elm St,Cedar Dr,Oak St,Oak St | depart,turn right,turn left,arrive |
+
+    Scenario: Staggered Intersection: do not collapse if a mode change is involved
+        Given the node map
+            """
+                j
+            a b c
+                d
+                e  f g
```
MoKob: These cases capture my fear very well. I would appreciate if you could also do a final one where only one of the sides is `pushing`, the segment in the middle and at the end are `driving`.

My reasoning here is that [here](https://github.com/Project-OSRM/osrm-backend/pull/3324/files#diff-99adcb8bccf8e94a176bdf92509ff070L819), we only have access to two of the contributing modes. We are missing the mode of the segment entering the staggered intersection. In a staggered intersection, three modes are part of the process. The segment coming in (currently untested in our check for a staggered intersection), the middle segment (which is the first of the two turns) and the out segment (the second parameter).

It seems that from the surrounding code, these might actually be considered correctly. But if we ever change the site of the call, we would have an untested regression.

features/guidance/bike-staggered-intersections.feature
```diff
@@ -0,0 +1,139 @@
...
+
+        And the ways
+            | nodes  | highway     | name     | oneway |
+            | abc    | residential | Oak St   |        |
+            | gfe    | residential | Oak St   | yes    |
+            | ihedcj | residential | Cedar Dr |        |
+
+        When I route I should get
+            | waypoints | route         | turns                                              | modes                                     |
+            | a,g       | Oak St,Cedar Dr,Oak St,Oak St | depart,turn right,turn left,arrive | cycling,cycling,pushing bike,pushing bike |
```
MoKob: This test is something we should probably discuss. In my personal view, I don't think that we should treat the same intersection differently, depending on where we come from.
It should either both be a staggered intersection, or none.
Changing mode in a double turn would be fine with me in these situations. But I'd be really confused if i travel to and from and get completely different instructions.

In addition, could you have a look at the alignment of the tables?

daniel-j-h: Capturing from chat: a `RouteStep` holds the specific `TravelMode` in its `mode` member.

If you check the two modes you can selectively allow only specific modes (e.g. bicycle pushing, but not ferries).

karenzshea: I added a check that allows an intersection to be considered staggered even if the previous previous step and the current step have differing modes, as long as the intermediary way does not involve a mode change. However, there is a subsequent mode check that simply compares `previous` and `current` mode that disallows collapsing of these situations. I ticketed Ticketed in https://github.com/Project-OSRM/osrm-backend/issues/3376 as this is beyond the scope of just refactoring `isStaggeredIntersection()`

MoKob:
I am ok with merging this in (given we link both the original issue (https://github.com/Project-OSRM/osrm-backend/issues/3256) / the new one (https://github.com/Project-OSRM/osrm-backend/issues/3376) on the cucumber cases with mode changes) as a first step.

I feel that the original ticket goes further than what is done here (I've updated the description) and should be kept open, even after we merge these changes.

Issue #3299, opened 2016-11-12
Import of planet.osm.pbf failed


georgbachmann:
Hello,

I already did a testrun with a smaller osm.pbf file and now I wanted to see how OSRM performs on the whole world. So I downloaded a pdf file covering the whole world and tried importing. It failed before because it ran out of memory, then out of disk space. I eliminated those issues (by increasing my google-compute-engine machine... now I have quite a monster with more than 50GB RAM and 8 cores) and now it says that it was killed in the middle of processing?!?

```
osrm@instance-1:~/pbf-files$ osrm-extract planet-latest.osm.pbf -p ~/osrm-backend/profiles/bicycle.lua
[info] Using script /home/osrm/osrm-backend/profiles/bicycle.lua
[info] Input file: planet-latest.osm.pbf
[info] Profile: bicycle.lua
[info] Threads: 8
[STXXL-MSG] STXXL v1.4.1 (prerelease/Debug)
[STXXL-ERRMSG] Warning: no config file found.
[STXXL-ERRMSG] Using default disk configuration.
[STXXL-MSG] Disk '/var/tmp/stxxl' is allocated, space: 1000 MiB, I/O implementation: syscall autogrow delete_on_exit queue=0 devid=0
[info] Parsing in progress..
[info] input file generated by planet-dump-ng 1.1.4
[info] timestamp: 2016-10-31T02:00:01Z
[info] Parsing finished after 4317.63 seconds
[info] Raw input contains 3585286779 nodes, 374765439 ways, and 4564745 relations
[extractor] Sorting used nodes        ... ok, after 289.562s
[extractor] Erasing duplicate nodes   ... ok, after 249.858s
[extractor] Sorting all nodes         ... ok, after 2283.89s
[extractor] Building node id map      ... ok, after 1244.16s
[extractor] setting number of nodes   ... ok
[extractor] Confirming/Writing used nodes     ... ok, after 680.895s
[info] Processed 904070651 nodes
[extractor] Sorting edges by start    ... Killed
```

Are there any more logs that would be helpful? In `stxxl.errlog` and `stxxl.log` I can't find any more useful information on what went wrong...

Should OSRM even be able to handle files as large as the whole world?

Cheers,
Georg

daniel-j-h:
I don't think 50 GB is enough for the planet. Check `dmesg` to see if you got killed by the oom killer?

[...]

PR #3327, opened 2016-11-16
Conditionally checks std::is_trivially_copyable only if available.

daniel-j-h:
Master doesn't build for some stdlibs since we merged the [datafacade changeset](https://github.com/Project-OSRM/osrm-backend/pull/3165).

Also a user [reported](https://github.com/Project-OSRM/osrm-backend/issues/3299#issuecomment-260940464) seeing a lot of

> /home/georgbachmann/osrm-backend/include/util/isatty.hpp:13:30: warning: missing terminating ' character
> #error Unknown platform - don't know which header to include for isatty()

warnings coming from the same changeset.

## Tasklist

 - [x] review
 - [x] adjust for comments

### DIFF
--- a/include/storage/io.hpp
+++ b/include/storage/io.hpp
@@ -5,17 +5,18 @@
 #include "extractor/extractor.hpp"
 #include "extractor/original_edge_data.hpp"
 #include "extractor/query_node.hpp"
+#include "util/exception.hpp"
 #include "util/fingerprint.hpp"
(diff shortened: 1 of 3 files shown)


danpat:
Looks good - as long as we keep that check on our main platforms, we'll catch any mis-uses.

daniel-j-h:
Sadly we don't have Travis build e.g. for Ubuntu 15.04 environments on which we ran into this.

Issue #2951, opened 2016-09-26
GCC/ LTO: Internal Compiler Error (lto_varpool_replace_node), different optimization flags


julienfr112:
after compilation of all lib and programs without problem:

cd example;mkdir build;cd build;cmake ..;make

[ 50%] Linking CXX executable osrm-example
lto1: internal compiler error: in lto_varpool_replace_node, at lto/lto-symtab.c:143
Please submit a full bug report,
with preprocessed source if appropriate.
See file:///usr/share/doc/gcc-5/README.Bugs for instructions.
lto-wrapper: fatal error: /usr/bin/c++ returned 1 exit status
compilation terminated.
/usr/bin/ld: error: lto-wrapper failed

ubuntu 16.04, gcc version 5.4.0 20160609 (Ubuntu 5.4.0-6ubuntu1~16.04.2)

daniel-j-h:
Looks like [a bug in GCC's LTO implementation due to using different optimization flag](https://gcc.gnu.org/ml/gcc-bugs/2016-01/msg00667.html).

If I can remember correctly @springmeyer adapted the build system a couple of month ago to now use the same optimization flags everywhere, not sure what's the state of this though.

Here's a workaround for now: build both `libosrm` as well as the example without LTO:

    cmake .. -DCMAKE_BUILD_TYPE=Release -DENABLE_LTO=OFF
    cmake --build .

TheMarex:
Problem is caused by the fact that we export `CMAKE_CXX_FLAGS` as cflags in our `libosrm.pc` file. This includes a ton of internal flags that make GCC crash when used on binaries and not on libraries apparently.

TheMarex:
Fixed by #3330.

PR #3330, opened 2016-11-16
Don't expose internal cflags to pkg-config

TheMarex:
# Issue

This fixes issue #2951.

## Tasklist
 - [x] review
 - [x] adjust for comments

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -25,7 +25,6 @@ env:
    - CCACHE_TEMPDIR=/tmp/.ccache-temp
    - CCACHE_COMPRESS=1
    - CASHER_TIME_OUT=599 # one second less than 10m to avoid 10m timeout error: https://github.com/Project-OSRM/osrm-backend/issues/2742
-   - JOBS=4
    - CCACHE_VERSION=3.3.1
    - CMAKE_VERSION=3.6.2
 
@@ -119,7 +118,15 @@ matrix:
 
 before_install:
   - if [[ ! -z $TARGET_ARCH ]] ; then source ./scripts/travis/before_install.$TARGET_ARCH.sh ; fi
-  - if [[ $(uname -s) == 'Darwin' ]]; then sudo mdutil -i off /; fi;
+  - |
+    if [[ "${TRAVIS_OS_NAME}" == "linux" ]]; then
+      export JOBS=$((`nproc` + 1))
+    fi
+  - |
+    if [[ "${TRAVIS_OS_NAME}" == "osx" ]]; then
+      export JOBS=$((`sysctl -n hw.ncpu` + 1))
+      sudo mdutil -i off /
+    fi
   - source ./scripts/install_node.sh 4
   - npm install
   - ./third_party/mason/mason install ccache ${CCACHE_VERSION}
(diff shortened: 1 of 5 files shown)


TheMarex:
I don't understand why the 32bit build does not work anymore with this, see https://travis-ci.org/Project-OSRM/osrm-backend/jobs/176806721#L1488-L1516

On the one side the cmake for the `osrm-example` seems to detect 32bit but then prints:

> /usr/bin/ld: skipping incompatible /usr/local/lib/libosrm.a when searching for -losrm

oxidase:
@TheMarex the problem was in https://github.com/Project-OSRM/osrm-backend/pull/3330/commits/9abe341c21361b7081c0e174c3aa8b2ea9b33ebd#diff-60ec4631040a17b93814f9da4955c281R36 environment flags were ignored and  x86_64 was used to build example. i also moved flags to .travis to make them visible.

TheMarex:
Had to do some fixes for mason induced weirdness as well. All good now.

@daniel-j-h please review. :bowing_man:

.travis.yml
```diff
@@ -55,17 +54,18 @@ matrix:
...
     # Release Builds
     - os: linux
-      compiler: "mason-release"
+      compiler: "mason-linux-release"
       addons:
         apt:
           sources: ['ubuntu-toolchain-r-test']
           packages: ['libstdc++-5-dev']
-      env: BUILD_TYPE='Release' ENABLE_MASON=ON
+      env: CLANG_VERSION='3.8.1' BUILD_TYPE='Release' ENABLE_MASON=ON
```
daniel-j-h: Why 3.8.1 - 3.9 is last stable

.travis.yml
```diff
@@ -77,7 +77,9 @@ matrix:
 
     - os: linux
       compiler: "gcc-6-release-i686"
-      env: TARGET_ARCH='i686' CCOMPILER='gcc-6' CXXCOMPILER='g++-6' BUILD_TYPE='Release'
+      env: >
```
daniel-j-h: Before reading through the 80 pdf Yaml spec (again), I guess this just allows multi-line env vars?

.travis.yml
```diff
@@ -116,15 +118,23 @@ matrix:
...
-  - if [[ $(uname -s) == 'Darwin' ]]; then sudo mdutil -i off /; fi;
+  - |
+    if [[ "${TRAVIS_OS_NAME}" == "linux" ]]; then
+      export JOBS=$((`nproc` + 1))
+    fi
+  - |
+    if [[ "${TRAVIS_OS_NAME}" == "osx" ]]; then
+      export JOBS=$((`sysctl -n hw.ncpu` + 1))
+      sudo mdutil -i off /
+    fi
```
daniel-j-h: Do you know how the number of procs works on the Travis machines? I think I remember reading somewhere it was reporting the hosts number of procs and should be hardcoded.

CMakeLists.txt
```diff
@@ -240,6 +234,7 @@ if (ENABLE_COVERAGE)
 endif()
 if (ENABLE_SANITIZER)
   set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -fsanitize=address")
+  set(OSRM_CFLAGS "${OSRM_CFLAGS} -fsanitize=address")
```
daniel-j-h: Why do we expose the sanitizer flags to users? Is this needed to build the example?

TheMarex: Yes this is needed if you link a sanitizer enabled libosrm in any library.

.travis.yml
TheMarex: Just keeps the current version we use on master. Upgrading should be a separate PR. This fixes a problem where we didn't install clang because this wasn't set.

.travis.yml
TheMarex: https://travis-ci.org/Project-OSRM/osrm-backend/jobs/177262393#L1117

Seems to work. I think this was only a problem on container machines not GCE.

Issue #3332, opened 2016-11-16
Refactor GraphLoader


ghoshkaj:
Use the new File reader library inside graph_loader.hpp. References https://github.com/Project-OSRM/osrm-backend/issues/3308.

PR #3333, opened 2016-11-16
Refactor graphloader

ghoshkaj:
# Issue

https://github.com/Project-OSRM/osrm-backend/issues/3332

## Tasklist
 - [x] review
 - [x] adjust for comments

### DIFF
--- a/include/util/graph_loader.hpp
+++ b/include/util/graph_loader.hpp
@@ -5,6 +5,7 @@
 #include "extractor/node_based_edge.hpp"
 #include "extractor/query_node.hpp"
 #include "extractor/restriction.hpp"
+#include "storage/io.hpp"
 #include "util/exception.hpp"
 #include "util/fingerprint.hpp"
 #include "util/simple_logger.hpp"
@@ -33,25 +34,14 @@ namespace util
  * The since the restrictions reference nodes using their external node id,
  * we need to renumber it to the new internal id.
 */
-inline unsigned loadRestrictionsFromFile(std::istream &input_stream,
+inline unsigned loadRestrictionsFromFile(storage::io::FileReader &file_reader,
                                          std::vector<extractor::TurnRestriction> &restriction_list)
 {
-    const FingerPrint fingerprint_valid = FingerPrint::GetValid();
-    FingerPrint fingerprint_loaded;
-    unsigned number_of_usable_restrictions = 0;
-    input_stream.read((char *)&fingerprint_loaded, sizeof(FingerPrint));
-    if (!fingerprint_loaded.TestContractor(fingerprint_valid))
-    {
-        SimpleLogger().Write(logWARNING) << ".restrictions was prepared with different build.\n"
-                                            "Reprocess to get rid of this warning.";
-    }
-
-    input_stream.read((char *)&number_of_usable_restrictions, sizeof(unsigned));
+    unsigned number_of_usable_restrictions = file_reader.ReadElementCount32();
     restriction_list.resize(number_of_usable_restrictions);
     if (number_of_usable_restrictions > 0)
     {
-        input_stream.read((char *)restriction_list.data(),
-                          number_of_usable_restrictions * sizeof(extractor::TurnRestriction));
+        file_reader.ReadInto(restriction_list.data(), number_of_usable_restrictions);
     }
 
     return number_of_usable_restrictions;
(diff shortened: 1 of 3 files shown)


ghoshkaj:
@daniel-j-h for review

include/util/graph_loader.hpp
```diff
@@ -32,27 +33,17 @@ namespace util
  * The since the restrictions reference nodes using their external node id,
  * we need to renumber it to the new internal id.
 */
-inline unsigned loadRestrictionsFromFile(std::istream &input_stream,
+inline unsigned loadRestrictionsFromFile(std::string &filename,
```
MoKob: if we pass in a filename, it should be `const`. Any reason for the switch to the filename?
It's less flexible, since we cannot store multiple items in a single file anymore. At some point, we probably want to get rid of all these 1000 files and have many items combined in a file. At that point, we would have to switch back to a stream again.

include/util/graph_loader.hpp
```diff
@@ -32,27 +33,17 @@ namespace util
...
-    input_stream.read((char *)&fingerprint_loaded, sizeof(FingerPrint));
-    if (!fingerprint_loaded.TestContractor(fingerprint_valid))
-    {
-        SimpleLogger().Write(logWARNING) << ".restrictions was prepared with different build.\n"
-                                            "Reprocess to get rid of this warning.";
-    }
-
-    input_stream.read((char *)&number_of_usable_restrictions, sizeof(unsigned));
+    storage::io::FileReader file(filename, storage::io::FileReader::VerifyFingerprint);
+    unsigned number_of_usable_restrictions = file.ReadElementCount32();
```
MoKob: I've requested this before in a different PR. Can we get rid of `32` bit size in the beginning completely? Right now we store many different versions of sizes, hard to keep track of which file has which size.

include/util/graph_loader.hpp
```diff
@@ -62,32 +53,21 @@ inline unsigned loadRestrictionsFromFile(std::istream &input_stream,
  *  - list of traffic lights
  *  - nodes indexed by their internal (non-osm) id
  */
-inline NodeID loadNodesFromFile(std::istream &input_stream,
+inline NodeID loadNodesFromFile(std::string &filename,
```
MoKob: same as above, while not pass a file, instead of a filename?

include/util/graph_loader.hpp
```diff
@@ -62,32 +53,21 @@ inline unsigned loadRestrictionsFromFile(std::istream &input_stream,
...
-    NodeID n;
-    input_stream.read(reinterpret_cast<char *>(&n), sizeof(NodeID));
-    SimpleLogger().Write() << "Importing n = " << n << " nodes ";
-
-    node_array.reserve(n);
+    node_array.reserve(number_of_nodes);
 
     extractor::ExternalMemoryNode current_node;
-    for (NodeID i = 0; i < n; ++i)
+    for (NodeID i = 0; i < number_of_nodes; ++i)
```
MoKob: since we are reading each element anyhow, I would say we should `resize` the `node_array` instead of reserve and pass the elements directly into `ReadInto`.

include/util/graph_loader.hpp
```diff
@@ -103,21 +83,24 @@ inline NodeID loadNodesFromFile(std::istream &input_stream,
...
                                 std::vector<extractor::NodeBasedEdge> &edge_list)
 {
-    EdgeID m;
-    input_stream.read(reinterpret_cast<char *>(&m), sizeof(unsigned));
-    edge_list.resize(m);
-    SimpleLogger().Write() << " and " << m << " edges ";
 
-    input_stream.read((char *)edge_list.data(), m * sizeof(extractor::NodeBasedEdge));
+    storage::io::FileReader file(filename, storage::io::FileReader::VerifyFingerprint);
+    NodeID number_of_nodes = file.ReadElementCount32();
```
MoKob: again, 64 bit just to be sure and have it consistent?

include/util/graph_loader.hpp
```diff
@@ -103,21 +83,24 @@ inline NodeID loadNodesFromFile(std::istream &input_stream,
...
+    file.Skip<extractor::ExternalMemoryNode>(number_of_nodes);
+    EdgeID number_of_edges_filereader = file.ReadElementCount32();
```
MoKob: same here

MoKob: Also add a check after to make sure the number of edges still fits into our EdgeIDs (e.g. via a numeric_cast)

include/util/graph_loader.hpp
ghoshkaj: @MoKob the only reason I'm switching to filename is so that I can use the new fileread library that was merged in with @danpat's [big refactor PR](https://github.com/Project-OSRM/osrm-backend/pull/3165/files#diff-6404e45b65aab921c04657dd4505e125R27) which as far as I can tell only allows fileNames. 

I'll change the the filename into being a const.

include/util/graph_loader.hpp
ghoshkaj: Same as [above](https://github.com/Project-OSRM/osrm-backend/pull/3333#discussion_r88491574).

include/util/graph_loader.hpp
MoKob: Then you could still pass in a file per reference?

include/util/graph_loader.hpp
ghoshkaj: :+1:

ghoshkaj: Okay I tried to do this but not sure how to. I used resize, but then it doesn't work with emplace_back, as I think it started pushing the current_node.lon, lat, etc. onto the end of the space allocated for this vector. So I changed it back to reserve and it worked. Going to look into this further tomorrow.

ghoshkaj:
for review please! @daniel-j-h or @MoKob 😃

include/util/graph_loader.hpp
MoKob: ok. sorry for not explaining this better:

you can either do a combination:

```
vector<data_type> data;
data.reserve(n);
for( i = 0 to n-1 )
   data.push_back(meaningful_stuff);
```

or

```
vector<data_type> data;
data.resize(n);
for( i = 0 to n-1 )
   data[i] = meaningful_stuff;
```

In your case you would do `ReadInto(data[i]);`

ghoshkaj: Okay! So! After discussion with Karen and DanPat, I realize that we can't actually simply read directly from the disk data (from the file) into RAM as is, because the disk data structures is a vector of `ExternalMemoryNode`s which holds 5 pieces of information (lon, lat, id, barrier, traffic_light), while the RAM data structure is a vector of `QueryNode`s which contains 3 pieces of information (lon, lat, id). So we can't do a direct `ReadInto(data[i]);`. I've adopted it as far as I could though... :)

include/util/graph_loader.hpp
```diff
@@ -33,27 +34,15 @@ namespace util
...
-    }
-
-    input_stream.read((char *)&number_of_usable_restrictions, sizeof(unsigned));
+    unsigned number_of_usable_restrictions = file_reader.ReadElementCount32();
     restriction_list.resize(number_of_usable_restrictions);
-    if (number_of_usable_restrictions > 0)
-    {
-        input_stream.read((char *)restriction_list.data(),
-                          number_of_usable_restrictions * sizeof(extractor::TurnRestriction));
+    if (number_of_usable_restrictions > 0) {
```
TheMarex: This needs to be formated with clang-format again, it should place the `{` on the next line.

include/util/graph_loader.hpp
```diff
@@ -64,33 +53,24 @@ inline unsigned loadRestrictionsFromFile(std::istream &input_stream,
  *  - nodes indexed by their internal (non-osm) id
  */
 template <typename BarrierOutIter, typename TrafficSignalsOutIter>
-NodeID loadNodesFromFile(std::istream &input_stream,
+NodeID loadNodesFromFile(storage::io::FileReader& file_reader,
```
TheMarex: Same formating problem here should be ` &file_reader`

src/tools/components.cpp
```diff
@@ -111,7 +107,7 @@ int main(int argc, char *argv[])
     }
 
     std::vector<osrm::tools::TarjanEdge> graph_edge_list;
-    auto number_of_nodes = osrm::tools::loadGraph(argv[1], coordinate_list, graph_edge_list);
+    auto number_of_nodes = osrm::tools::loadGraph(std::string(argv[1]), coordinate_list, graph_edge_list);
```
TheMarex: `const char*` should converted to `std::string` automatically (implicit conversion). You can remove the `std::string()` I think.

MoKob:
From me as well. After the formatting is done (as already pointed out by @TheMarex), it looks fine to me as well.

ghoshkaj:
Merge plz?

PR #3336, opened 2016-11-17
Added a copy of std::align due to missing implementation in gcc < 5

### LINKED ISSUE #3334: std::align breaks master on gcc < 5
https://github.com/Project-OSRM/osrm-backend/pull/3328 uses `std::align` - looks like it's missing from the stdlib shipping with gcc < 5 see https://gcc.gnu.org/bugzilla/show_bug.cgi?id=57350.

cc @oxidase

oxidase:
# Issue

Due to missing implementation std::align in gcc 4.9 https://gcc.gnu.org/bugzilla/show_bug.cgi?id=57350#c8
own align is still needed. 
The method is directly copied from /usr/include/c++/6/memory but has no space computation support.

To check a new job with gcc 4.9 must added in .travis file.

## Tasklist
 - [x] review
 - [x] adjust for comments

## Requirements / Relations
Resolves #3334

References:
https://gcc.gnu.org/bugzilla/show_bug.cgi?id=57350

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -79,6 +79,14 @@ matrix:
       compiler: "gcc-6-release-i686"
       env: TARGET_ARCH='i686' CCOMPILER='gcc-6' CXXCOMPILER='g++-6' BUILD_TYPE='Release'
 
+    - os: linux
+      compiler: "gcc-4.9-release"
+      addons: &gcc49
+        apt:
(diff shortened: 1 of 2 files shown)


oxidase:
Added 4.9 build job. gcc 4.8 is already old, because it uses std::make_unique that is not supported.

PR #3339, opened 2016-11-18
Makes Throughput Numbers Integral Types

daniel-j-h:
Noticed this while having to look at our logs, otherwise you'd get floating point e-notation default printing.

### DIFF
--- a/src/contractor/contractor.cpp
+++ b/src/contractor/contractor.cpp
@@ -6,11 +6,11 @@
 #include "extractor/edge_based_graph_factory.hpp"
 #include "extractor/node_based_edge.hpp"
 
+#include "storage/io.hpp"
 #include "util/exception.hpp"
 #include "util/graph_loader.hpp"
 #include "util/integer_range.hpp"
 #include "util/io.hpp"
-#include "storage/io.hpp"
 #include "util/simple_logger.hpp"
 #include "util/static_graph.hpp"
 #include "util/static_rtree.hpp"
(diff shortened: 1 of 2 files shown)


src/contractor/contractor.cpp
```diff
@@ -6,11 +6,11 @@
 #include "extractor/edge_based_graph_factory.hpp"
 #include "extractor/node_based_edge.hpp"
 
+#include "storage/io.hpp"
```
TheMarex: Where is thus new include coming from? Was it missing before?

src/contractor/contractor.cpp
```diff
@@ -193,10 +193,14 @@ int Contractor::Run()
 
     TIMER_STOP(preparing);
 
+    const auto nodes_per_second =
```
TheMarex: Why not use `std::setprecision(2)` here instead of explicitly rounding?

src/extractor/extractor.cpp
```diff
@@ -289,9 +289,13 @@ int Extractor::run(ScriptingEnvironment &scripting_environment)
 
         WriteEdgeBasedGraph(config.edge_graph_output_path, max_edge_id, edge_based_edge_list);
 
-        util::SimpleLogger().Write()
-            << "Expansion  : " << (number_of_node_based_nodes / TIMER_SEC(expansion))
-            << " nodes/sec and " << ((max_edge_id + 1) / TIMER_SEC(expansion)) << " edges/sec";
+        const auto nodes_per_second =
```
TheMarex: Same here, could just be `std::setprecision`.

src/contractor/contractor.cpp
daniel-j-h:
```c++
{ cout << std::setprecision(2) << (50000000000000ull / 2.); }
2.5e+13
```

daniel-j-h: To be clear here this works with an additional `std::fixed`, too - no reason I chose the `uint64_t` over the format specifiers other that I don't like streams much.

daniel-j-h:
@TheMarex - okay with the reasoning above? Want to approve and merge, or should I change the casts?

TheMarex:
Meh I feel like using actual stream operators to be the less hacky solution, but for this it doesn't really matter. Merge at will.

Issue #3311, opened 2016-11-15
Node Ordering for Spacial and Temporal Locality (Node Based Graph / Node Info List)


daniel-j-h:
At the moment our nodes are re-numbered sequentially from 0 to number of used nodes. There is no locality in node order at all.

The extractor and especially our guidance code makes use of an immense amount of 1/ coordinate lookups through the node info list (keyed by node id) and 2/ graph traversal via the Dynamic Node Based Graph (nodes sorted by source).

At the moment this is random access as random as it can get (read: bad).

We should establish locality in node ordering by re-using our Hilbert space-filling curve. The reason we're not doing a BFS-based ordering is mainly because we'd need to construct a graph first. A space-filling curve based ordering is good enough for a solid approximation.

-------------------------------------------------------------------------------------------------------------------

Here are some notes:

- For every way we loop over its nodes remembering used node ids [in the extractor callbacks](https://github.com/Project-OSRM/osrm-backend/blob/538bbd47d1ee169764b66df3ef67ecfe131cdfec/src/extractor/extractor_callbacks.cpp#L304-L309)

- In the extraction containers we renumber nodes [sequentially](https://github.com/Project-OSRM/osrm-backend/blob/538bbd47d1ee169764b66df3ef67ecfe131cdfec/src/extractor/extraction_containers.cpp#L253) saving a mapping into the [external to internal node id map](https://github.com/Project-OSRM/osrm-backend/blob/538bbd47d1ee169764b66df3ef67ecfe131cdfec/src/extractor/extraction_containers.cpp#L269)

- We [write the nodes to disk](https://github.com/Project-OSRM/osrm-backend/blob/538bbd47d1ee169764b66df3ef67ecfe131cdfec/src/extractor/extraction_containers.cpp#L597) in a contiguous chunk of memory to the beginning of the `.osrm` file prefixed with the number of nodes

- When [building the edge expanded graph](https://github.com/Project-OSRM/osrm-backend/blob/538bbd47d1ee169764b66df3ef67ecfe131cdfec/src/extractor/extractor.cpp#L254) the first thing we're doing is [reading the nodes from disk again into the ](https://github.com/Project-OSRM/osrm-backend/blob/538bbd47d1ee169764b66df3ef67ecfe131cdfec/src/extractor/extractor.cpp#L465-L466) this time unzipping them into nodes, barriers, traffic signals. Related: https://github.com/Project-OSRM/osrm-backend/issues/3238.

[...]

PR #3342, opened 2016-11-18
Sort nodes spatially to speed up later queries.

danpat:
# Issue

Addresses #3311 

This change sorts node IDs on a Hilbert curve, on the assumption that this will improve cache locality during later queries of this data (coordinate data is heavily used during edge-based-graph construction when the guidance code does lots of localize graph exploring).

## Tasklist
 - [ ] review
 - [ ] adjust for comments

### DIFF
--- a/src/extractor/extraction_containers.cpp
+++ b/src/extractor/extraction_containers.cpp
@@ -9,6 +9,7 @@
 #include "util/io.hpp"
 #include "util/simple_logger.hpp"
 #include "util/timing_util.hpp"
+#include "util/hilbert_value.hpp"
 
 #include <boost/assert.hpp>
 #include <boost/filesystem.hpp>
@@ -234,14 +235,11 @@ void ExtractionContainers::PrepareNodes()
     TIMER_START(sorting_nodes);
     stxxl::sort(all_nodes_list.begin(),
                 all_nodes_list.end(),
-                ExternalMemoryNodeSTXXLCompare(),
+                ExternalMemoryNodeWithHilbertSTXXLIDCompare(),
                 stxxl_memory);
     TIMER_STOP(sorting_nodes);
     std::cout << "ok, after " << TIMER_SEC(sorting_nodes) << "s" << std::endl;
 
-    std::cout << "[extractor] Building node id map      ... " << std::flush;
-    TIMER_START(id_map);
-    external_to_internal_node_id_map.reserve(used_node_id_list.size());
     auto node_iter = all_nodes_list.begin();
     auto ref_iter = used_node_id_list.begin();
     const auto all_nodes_list_end = all_nodes_list.end();
@@ -250,8 +248,9 @@ void ExtractionContainers::PrepareNodes()
     // handle > uint32_t actual usable nodes.  This should be OK for a while
     // because we usually route on a *lot* less than 2^32 of the OSM
     // graph nodes.
-    std::uint64_t internal_id = 0;
 
+    std::cout << "[extractor] Calculating Hilbert codes for used nodes      ... " << std::flush;
+    TIMER_START(calculating_hilbert_codes);
     // compute the intersection of nodes that were referenced and nodes we actually have
     while (node_iter != all_nodes_list_end && ref_iter != used_node_id_list_end)
     {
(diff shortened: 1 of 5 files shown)


src/extractor/extraction_containers.cpp
```diff
@@ -292,23 +322,23 @@ void ExtractionContainers::PrepareEdges(ScriptingEnvironment &scripting_environm
...
-    auto node_iterator = all_nodes_list.begin();
+    auto node_iterator = used_node_id_list.begin();
     auto edge_iterator = all_edges_list.begin();
 
     const auto all_edges_list_end = all_edges_list.end();
-    const auto all_nodes_list_end = all_nodes_list.end();
+    const auto used_node_id_list_end = used_node_id_list.end();
 
-    while (edge_iterator != all_edges_list_end && node_iterator != all_nodes_list_end)
+    while (edge_iterator != all_edges_list_end && node_iterator != used_node_id_list_end)
```
danpat: I'm pretty sure this logic is incorrect, because edge creation hangs for larger graphs.  Still needs some work.

danpat:
Note: This is *not ready*, it doesn't work properly yet.

danpat:
So, while this PR is still slightly buggy, the main point was to try to speed up edge-based-edge generation, which is coordinate-usage heavy.

Based on re-processing California a few times, the results are:

Edge-based-graph generation without Hilbert sorting:  30s
Edge-based-graph generation *with* Hilbert sorting: 27s

This doesn't seem like a significant gain.  In addition, the loss of locality during edge generation in the `extraction_containers.cpp` leads to that phase taking significantly longer - 5 seconds vs 800 seconds with the current implementation.  While my changes are probably less than optimal, it doesn't seem possible to avoid a penalty - the current approach is reasonably well optimized for unused-node removal and edge-coordinate updating.

I'm open to ideas here, or a review to spot if I've made some glaring error (note: there are 5 or 6 tests failing still, so *some* graph-shape behaviour has changed, I'm just not exactly sure what without a deeper dive).

Given the negligible speedup, I'm inclined to resign this change to the trash.

/cc @daniel-j-h @TheMarex @MoKob

src/extractor/extraction_containers.cpp
```diff
@@ -371,7 +389,14 @@ void ExtractionContainers::PrepareEdges(ScriptingEnvironment &scripting_environm
...
 
-        if (edge_iterator->result.osm_target_id < node_iterator->node_id)
+        BOOST_ASSERT(edge_iterator->weight_data.speed >= 0);
+        BOOST_ASSERT(edge_iterator->source_coordinate.lat !=
+                     util::FixedLatitude{std::numeric_limits<std::int32_t>::min()});
+        BOOST_ASSERT(edge_iterator->source_coordinate.lon !=
+                     util::FixedLongitude{std::numeric_limits<std::int32_t>::min()});
+
+        auto id_iter = external_to_internal_node_id_map.find(edge_iterator->result.osm_target_id);
+        if (id_iter == external_to_internal_node_id_map.end())
```
danpat: I suspect this lookup is a big source of the slowdown.

src/extractor/extraction_containers.cpp
```diff
@@ -380,25 +405,15 @@ void ExtractionContainers::PrepareEdges(ScriptingEnvironment &scripting_environm
...
-        }
 
-        BOOST_ASSERT(edge_iterator->result.osm_target_id == node_iterator->node_id);
-        BOOST_ASSERT(edge_iterator->weight_data.speed >= 0);
-        BOOST_ASSERT(edge_iterator->source_coordinate.lat !=
-                     util::FixedLatitude{std::numeric_limits<std::int32_t>::min()});
-        BOOST_ASSERT(edge_iterator->source_coordinate.lon !=
-                     util::FixedLongitude{std::numeric_limits<std::int32_t>::min()});
+        const auto node_coord = util::Coordinate(all_nodes_list[id_iter->second].lon,
+                                                 all_nodes_list[id_iter->second].lat);
```
danpat: Also this - previously, coordinates were cache-local in the node iterator, no additional lookup was required.
Because the edge IDs aren't hilbert order sorted, we're essentially doing random access to the `all_nodes_list`.

TheMarex:
I think it is valuable to figure out why the tests are not passing. If we have behavior in our code that relies on a specific node order, that is not good. Looking at the failure there seems to be something odd going on with the geometry.

Looking at the implementation it would make sense to do the following algorithmic changes to skip most of the overhead you are seeing:

1. Compute hilbert value in `ExtractionCallbacks` when creating the ExternalMemoryNode (will not incur a write for very single node in a IO heavy loop)
2. Keep the current logic that first filters the used nodes (to first reduce the data size)
3. Sort the nodes by hilbert value (most expensive step, but data volume will be lower)
4. After the first mapping OSM -> internal was computed use it to compute OSM -> internal -> spatial

The optimization here is to not do the `stxxl::sort` on all nodes, but only on the relevant ones.

I would also recommend running this on a much bigger dataset before we discard this. The graph for California is small enough that there is a reasonable chance to hit the cache even if you use random order.

daniel-j-h:
@danpat can you check your numbers again now with the Hilbert space-filling curve improvements?

daniel-j-h:
Ping - what's the state of this, especially after the Hilbert changes? Still worth investigating?

Issue #3340, opened 2016-11-18
profiles: assigning nil to result strings crashes osrm-extract


emiltin:
ie this crashes:

```lua
result.ref = nil
```

name, ref and pronunciation fields all have this behaviour. allowing asignment of nil should just be trreated as an empty string, as this simplifies profiles and is safer.

emiltin:
assigning nil to boolean fields result.roundabout also causes a crash. it would be better to treat nil as false.

daniel-j-h:
Here's my understanding: you assign `nil` in Lua - Luabind converts this to a `nullptr` in C++-land.

And assigning a `std::string` a `nullptr` is Undefined Behavior.

TheMarex:
> as this simplifies profiles

Can you elaborate on this?

> and is safer.

I don't think this is a safety concern. I expect that luabind just throws an exception, right?

@daniel-j-h what is your feeling w.r.t to this issue? Currently it seems like expected behavior to me.

emiltin:
by simplifying i mean you could safely do e.g:
```lua
result.name = way:get_value_by_key('name')
```

TheMarex:
Fixed by #3344

PR #3344, opened 2016-11-20
Don't crash on assigning nil values to strings in lua

TheMarex:
# Issue

Fixes issue #3340.

## Tasklist
 - [x] review
 - [x] adjust for comments


features/testbot/nil.feature
```diff
@@ -0,0 +1,42 @@
+@routing @testbot @nil
+Feature: Testbot - Check assigning nil values
+    Scenario: Assign nil values to all way strings
+        Given the profile file "testbot" extended with
+        """
+        function way_function (way, result)
```
MoKob: this looks like we could easily create strange interaction with `way_function` also defined in the `testbot` profile. Is this guaranteed to replace the previously defined `way_function` or should we do a full dedicated profile here?

include/extractor/extraction_way.hpp
```diff
@@ -13,6 +13,21 @@ namespace osrm
 {
 namespace extractor
 {
+namespace detail
+{
+    inline void maybeSetString(std::string &str, const char* value)
+    {
+        if (value == nullptr)
```
MoKob: personal preference here, but I'd prefer if the string should be in a valid state before getting into this function. But nothing that has to change.

include/extractor/extraction_way.hpp
```diff
@@ -50,6 +65,20 @@ struct ExtractionWay
     void set_backward_mode(const TravelMode m) { backward_travel_mode = m; }
     TravelMode get_backward_mode() const { return backward_travel_mode; }
 
+    // wrappers to allow assigning nil (nullptr) to string values
+    void SetName(const char* value) { detail::maybeSetString(name, value); }
```
MoKob: is there a way we could also pass the destination to write into from, so we don't have to replicate this behaviour for all string types?

src/extractor/scripting_environment_lua.cpp
```diff
@@ -186,16 +186,16 @@ void LuaScriptingEnvironment::InitContext(LuaScriptingContext &context)
              // .def(luabind::constructor<>())
              .def_readwrite("forward_speed", &ExtractionWay::forward_speed)
              .def_readwrite("backward_speed", &ExtractionWay::backward_speed)
-             .def_readwrite("name", &ExtractionWay::name)
-             .def_readwrite("ref", &ExtractionWay::ref)
-             .def_readwrite("pronunciation", &ExtractionWay::pronunciation)
-             .def_readwrite("destinations", &ExtractionWay::destinations)
+             .property("name", &ExtractionWay::GetName, &ExtractionWay::SetName)
```
MoKob: if we could pass the `read-write`, we could do without the dedicated setters here.

features/testbot/nil.feature
TheMarex: Yes this is well defined in any scripting language I can think of (javascript, python, ruby and lua): Last definition always overrides. Pretty similar to the name shadowing for variables. Only C++ and the likes are a little bit more picky.

include/extractor/extraction_way.hpp
TheMarex: Hrm we need a get/set function for every member. The business logic is already moved to a helper function. I don't see how we could optimize this further.

MoKob: Might be that this is not possible. But before lua was assigning the values itself (below you changed this to a property with set/get methods). I am asking if we could have a function of the style `SetStringValue(std::string &destination, const char* value)` instead of a dedicated one for every value. 
But that is just a proposal. We would need to call this function from `lua` which can be considered just as ugly. Feel free to stick with the solution we have here.

MoKob: On second thought: since we want to have it less error prone in the profiles, having these setters inside of OSRM is the more secure version to do. Otherwise we could still run into problems when someone forgets to use the function I proposed.

PR #3345, opened 2016-11-20
Refactor/profile helpers

emiltin:
# Issue

This PR continues incremental LUA profile refactoring.
directional.lua adds convenient functions for finding the correct tag value in forward/backward directions, searching either a single tag of a sequence of tag. It looks first for the directional tag (e.g. maxspeed:forward) then the common tag (maxspeed). This is a commonly used pattern in profiles. The tag helper is used in the car profile as well as in guidance.lua and destination.lua where it streamlines code.

guidance.lua and direction.lua has been refactored and simplified (all tests still pass) and the number of tags fetched has been reduced - often you can abort after checking the first tags, instead of initially fetching all tags you might need. The files also uses Set to improve readability of value lists.

I don't see any drop in performance.

## Tasklist
 - [x] review
 - [ ] adjust for comments

## Requirements / Relations
 None that I see.


emiltin:
Functions in TagHelper could perhaps be named better?

profiles/car.lua
```diff
@@ -437,26 +437,15 @@ function handle_speed(way,result,data)
     local max_speed = parse_maxspeed( way:get_value_by_key("maxspeed") )
```
MoKob: is this max_speed value here deprecated now?

profiles/debug.lua
```diff
@@ -106,3 +139,9 @@ way_function(way,result)
 pprint(way)
 print("=>")
 pprint(result)
+print("\n")
```
MoKob: leftover debug? or is this still work in progress?

profiles/car.lua
```diff
@@ -732,9 +721,6 @@ function way_function(way, result)
   if handle_startpoint(way,result) == false then return end
   if handle_restricted(way,result,data) == false then return end
 
-  -- handle roundabout flags
-  if handle_roundabouts(way,result) == false then return end
-
   -- check if this way can be routed through, but not used as
   -- origin or destination
   if handle_restricted(way,result,data) == false then return end
```
emiltin: fixed

profiles/car.lua
emiltin: oh forgot to remove that line. maxspeed is handled in handle_maxspeed()

profiles/lib/guidance.lua
```diff
@@ -1,154 +1,146 @@
...
-        end
-    end
-    if psv_backward then
-        bw = tonumber(psv_backward);
-        if( bw == nil ) then
-            bw = 0
-        end
-    end
-    return fw, bw
+  local psv_forward, psv_backward = Directional.get_values_by_key(way,'lanes:psv')
```
daniel-j-h: From what I can tell the `Directional` abstraction just passes through the forward / backward namespaced values.

Which means all call sites have to check for
- reversible oneways and flip the direction, think: oneway=-1, *:forward=x, *:backward=y, and and
- check for oneway (or implied oneway) and discard a direction mis-tagging, think: oneway=1, *:forward=x, *:backward=y

(basically what handle_oneway does)

Can we abstract over all of this?

emiltin: yes checking for reverse oneways could be done in Directional.

i don't think all call sites currently perform this check, yet all tests pass - does that mean we are missing tests that validate this behaviour?

emiltin: to clarify, Directional.get_values_by_key() will first look for e.g. 'maxspeed:forward'. If that's not present it will loook for 'maxspeed'. This behaviour is correct for things like maxspeed, surface, access, etc.
Are there cases where it does not make sense? What about lanes?

daniel-j-h: I think tags `X:forward` and `X` should _always_ be the same (that is along the way's direction) - the forward prefix is only to allow for the additional backward version.

cc @MoKob for experience with lanes.

daniel-j-h: => https://github.com/Project-OSRM/osrm-backend/pull/3372

We don't handle directional overrides, but we should (not only for access but for all tags). I think your `Directional` abstraction is the perfect place to put logic for handling this.

emiltin:
The Directional helper swaps forward/backward tags if oneway=-1. But is this the correct behaviour?

Doesn't oneway=-1 mean your can only move in the reverse direction, and therefore that :backward tags should be considered while :forward tags should be ignored?

I Couldn't find a clear answer in the OSM wiki.

danpat:
@emiltin `:forward` and `:backward` are in the direction of the geometry, see: https://wiki.openstreetmap.org/wiki/Forward_%26_backward,_left_%26_right

They're independent of direction-of-travel.

emiltin:
@danpat right, that's what I thought. @daniel-j-h you previously commented:

> reversible oneways and flip the direction, think: oneway=-1, *:forward=x, *:backward=y, and and

That happened previsouly in guidance.lua i think? I'll remove the swapping then and have a look at adjusting any failing tests.

emiltin:
I removed the forward/backward swapping and fixed the tests "Scenario: Car - a way with all lanes HOV-designated is inaccessible by default (similar to hov=designated)" and "Car - Forwward/backward maxspeed on reverse oneways".

I think this PR is ready for merging, or do you see other things that should be changed? 
Additional profile refactor work can happen in new PRs.

MoKob:
@daniel-j-h I'd be good here. What about you?

@emiltin I'd only prefer if you could squash away some of these many commits

daniel-j-h:
Does the Directional abstraction now correctly handle the problems noted above? (e.g. the failing tests in https://github.com/Project-OSRM/osrm-backend/pull/3372). Other than that looks great - thanks @emiltin for your ongoing efforts to make osrm profiles beautiful!

emiltin:
The Directional helper was handling forward/backward tags correctly, but the car profile was still using the old find_access_tag() helper function which didn't. 

I've change this to use the new Directional.get_values_by_set() method, and added tests.

A question that came up is how to handle the result.is_access_restricted flag. Since now have both forward/backward access status, it's not entirely clear when I should  set is_access_restricted. I choose to set it if either direction is restricted. What's your thoughts on this? 

I've squashed commits as well.

daniel-j-h:
Hrm going back in time I can see the `access_restricted` flag added here:

https://github.com/Project-OSRM/osrm-backend/commit/3f6cc725d635ae8ccded81b8908dbe44069f3e7c

(also check the comments :wave: @emiltin :P)

I don't think we need this actually. cc @TheMarex

emiltin:
anything for me to do regarding the access_restricted flag?

daniel-j-h:
We don't need it at the moment, it's dead code - I'm for removing it from the code base.

@TheMarex @MoKob opinions on that?

TheMarex:
@daniel-j-h sounds good. We might want something similar in the future but I'm not sure the flag would even be the right mechanism. Worst case will be we need to revert.

emiltin:
i removed use of the is_access_restricted flag from the profile.

daniel-j-h:
@emiltin I just removed the `access_restricted` flag completely from the internal logic in https://github.com/Project-OSRM/osrm-backend/pull/3345/commits/1fee6a22b2c3be7c12252cfb825acdb29ef88f60.

Let's wait for Travis on this and then it's ready to go.

emiltin:
rebased on latest master

daniel-j-h:
Hm @emiltin I think you just deleted my commits which removed the `access_restricted` flag internally.

emiltin:
hmm that rebase was a too fast, since i was missing your changes :-/
rebasing your commit 1fee6a2 on master results in a conflict in node_based_edge.hpp:

```c++
<<<<<<< 2d788016d28147e6666e7f6bb373af289b286cb7
      backward(false), roundabout(false), circular(false), access_restricted(false),
      startpoint(true), is_split(false), travel_mode(false),
      lane_description_id(INVALID_LANE_DESCRIPTIONID)
=======
      backward(false), roundabout(false), circular(false), startpoint(true), is_split(false),
      travel_mode(false), lane_description_id(INVALID_LANE_DESCRIPTIONID)
>>>>>>> Removes access_restricted flag internally
```
care to take a look at that?

daniel-j-h:
Can you include my commits 1fee6a2 and ef31f46 (this one just removes the dead access restricted code from the bike and foot profile, too).

The conflicts you're seeing is the member variable `access_restricted` which I removed. There should be no `access_restricted` in the code base afterwards.

emiltin:
pushed your commit ef31f46. but i'm not sure how to resolve the conflict that arises when i rebase on master, as mentioned above.

emiltin:
let me look at it.

emiltin:
alright, should be fixed now. sorry for the hick up, and thanks for removing the flag on the c++ side.

daniel-j-h:
Thanks - and no worries Git can be a beast sometimes :)

PR #3346, opened 2016-11-21
Fixes accidentally registering types twice to Luabind, closes #2811.

### LINKED ISSUE #2811: Luabind "you are trying to register a class twice" Error in Extractor
root@/osrm_log #  ./osrm-extract -p profile.lua gcc-states-latest.osm.pbf 
[info] Using script profile.lua
[info] Input file: gcc-states-latest.osm.pbf
[info] Profile: profile.lua
[info] Threads: 24
[STXXL-MSG] STXXL v1.4.99 (prerelease/Release) (git 1babe452214b4613a2a488d80073f4185c05a0b3) + gnu parallel(20150623)
[STXXL-MSG] Disk '/tmp/stxx' is allocated, space: 50000 MiB, I/O implementation: syscall queue=0 devid=0
[info] Parsing in progress..
osrm-extract: src/class_registry.cpp:152: void luabind::detail::class_registry::add_class(const luabind::type_id&, luabind::detail::class_rep*): Assertion `(m_classes.find(info) == m_classes.end()) && "you are trying to register a class twice"' failed.

daniel-j-h:
For https://github.com/Project-OSRM/osrm-backend/issues/2811. I could reproduce the issues reported by users in a Ubuntu 16.04 container with Luabind (our fork) compiled from source. After a horrible deep dive through Luabind code this is what I found out:
- Luabind uses the `std::type_info` returned by the `typeid` operator to register classes
- Some of our enums are char typedefs since we can't bit-pack enums => `::Enum` hack
- Binding several `*::Enum` to Luabind registers the same underlying type with the same `typeid`
- Luabind registers them via a map's `operator[]` and therefore replaces the older ones

Note how this is fundamentally flawed from the beginning:
- If at all Luabind has to use the `type_info`'s hash code \*
- You better compile Luabind statically [otherwise this house of cards might fall down on you](https://github.com/mapbox/luabind/blob/a70e747c579d82e8e2ab108c2469454231630150/luabind/detail/class_rep.hpp#L160-L161)
- `typeid` for `int8_t`, `uint8_t` may be the same as `signed char`, `unsigned char` respectively as the stdlib may just typedef them

(\*)
> There is no guarantee that the same std::type_info instance will be referred to by all evaluations of the typeid expression on the same type, although std::type_info::hash_code of those type_info objects would be identical, as would be their std::type_index.

The workaround / hack here is to 1/ create new types with different `type_info`s and 2/ keep on binding the `::Enum`'s values since Luabind converts everything to integrals.

## Tasklist
 - [x] review
 - [x] adjust for comments

### DIFF
--- a/src/extractor/scripting_environment_lua.cpp
+++ b/src/extractor/scripting_environment_lua.cpp
@@ -35,9 +35,12 @@ template <class T>
 auto get_value_by_key(T const &object, const char *key) -> decltype(object.get_value_by_key(key))
 {
     auto v = object.get_value_by_key(key);
-    if (v && *v) {             // non-empty string?
+    if (v && *v)
+    { // non-empty string?
         return v;
-    } else {
+    }
+    else
+    {
         return NULL;
     }
 }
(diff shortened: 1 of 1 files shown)


MoKob:
@daniel-j-h I'd probably like to have a chat on why exactly this happens, since I cannot completely grasp the problem, but it looks fine to me.

Issue #3335, opened 2016-11-17
Use Bridleways when explicitly allowed to


daniel-j-h:
As reported in https://github.com/mapbox/mapbox-directions.js/issues/109:

We don't route over `highway=bridleway` + `foot=yes` or `bicycle=yes`.

Example way: https://www.openstreetmap.org/way/32713960#map=19/29.62293/-99.20729&layers=D
Bridleways in OSM: http://wiki.openstreetmap.org/wiki/Tag:highway%3Dbridleway

The OSM Wiki makes it clear there are country-specific regulations.
Together with explicit yes-tags we should route over them.

Tasks:
- [x] add Cucumber tests for routability and speeds with and without explicit yes-tags - `features/{bicycle, foot}access.feature`, https://github.com/Project-OSRM/osrm-backend/pull/3348
- [ ] add support to profiles - `profiles/{bicycle, foot}.lua`
- [ ] check other tags we currently don't route over but should route over when there are explicit yes-tags

http://wiki.openstreetmap.org/wiki/Key:access

MoKob:
@daniel-j-h While @karenzshea has added some tests (that work without modification), I'd be really interested to see what has been broken here.
From the OSM data, the path in question is a single way in OSM data. This includes both the part that is routed over and the part that isn't.
Given that we break them into different parts, I'd be curious to why a single part of that path might end up  routed over, and another one didn't.

Has anyone been able to reproduce the issue in question on an extract? In the example?

daniel-j-h:
Hrm this was not what I expected, sorry for the issue description then.

I can put reproducing this locally with the bike and walk profile on my plate.

daniel-j-h:
For the record testing this on both our cycling and walking stacks works now

![route](https://cloud.githubusercontent.com/assets/527241/20519667/1b10cb12-b0a4-11e6-8a24-4d87629e67a2.png)


even though the way in question was last edited a month ago - the path going east  7 days ago.

daniel-j-h:
Okay I can't reproduce this locally - @MoKob threw in the idea of a broken CH order. Here's what I used to run the pipeline in a loop and detecting changes in duration.

```bash
#!/usr/bin/env bash

set -o errexit
set -o pipefail
set -o nounset

# Run this script in a loop to detect the unroutable bridleway behavior found in #3335.
# Returns non-zero exit code on failure.

[...]

PR #3348, opened 2016-11-21
Route on highway=bridleway when access is explicitly given

karenzshea:
# Issue
https://github.com/Project-OSRM/osrm-backend/issues/3335

## Tasklist
 - [x] review
 - [x] adjust for comments

## Requirements / Relations
cc @daniel-j-h


karenzshea:
@daniel-j-h it seems like this should have always worked, as we have a check in here that will set a default speed for any tag that has a `bicycle=yes` or `foot=yes` tag even if it's an unknown way. I double checked the original broken query and it seems to work now.

daniel-j-h:
Interesting - sorry about my issue description then and thank you for tests.

Have to look into the original issue then :thinking:

PR #3353, opened 2016-11-22
fix update to length cache

MoKob:
# Issue

Minor issue with not updating the length cache correctly. The length cache assumes distances between coordinates. The update (at the changed location) assumes a prefix sum

## Tasklist
 - [x] review
 - [x] adjust for comments

## Requirements / Relations


src/extractor/guidance/coordinate_extractor.cpp
```diff
@@ -261,7 +261,8 @@ CoordinateExtractor::GetCoordinateAlongRoad(const NodeID intersection_node,
         const double offset = 0.5 * considered_lanes * ASSUMED_LANE_WIDTH;
         coordinates = TrimCoordinatesToLength(std::move(coordinates), offset, segment_distances);
         segment_distances.resize(coordinates.size());
-        segment_distances.back() = offset;
+        segment_distances.back() = util::coordinate_calculation::haversineDistance(
```
TheMarex: An assertion that `coordinates.size() >= 2` would help document assumptions here.

Issue #3352, opened 2016-11-22
hilbertCode is not really Hilbert SFC linear coordinate


oxidase:
[hilbertCode](https://github.com/Project-OSRM/osrm-backend/blob/master/src/util/hilbert_value.cpp#L72-L82) returns a space filing curve linearizion, but it is not the Hilbert curve. 

Here is the recurrent motif of the current implementation 
<img src="https://cloud.githubusercontent.com/assets/4421046/20527523/18c3b458-b0c9-11e6-914f-a91cfbfc77f1.png" width="50">, but correct is 
<img src="https://cloud.githubusercontent.com/assets/4421046/20527545/2e4dad42-b0c9-11e6-955b-c506c2a87e92.png" width="50">.

For 5th recursion level of SFC linearizion looks  
<img src="https://cloud.githubusercontent.com/assets/4421046/20527646/71403d54-b0c9-11e6-99d0-6e56ac311415.png" width="160"> instead of 
<img src="https://cloud.githubusercontent.com/assets/4421046/20527656/7cf99316-b0c9-11e6-827e-57f2b69225d6.png" width="160">

It must be clarified  if the non-Hilbert SFC is the correct original decision.

Related PR  #3342

danpat:
I can only think of two reasons:

  1. it's a bug (the function is `hilbertCode` after all).
  2.  the current implementation is an approximation that was fast and gave "good enough" locality for our purposes
  3. There is some subtle reason why an alternate curve shape performs better for OSRM's use-case.

My bet is on (1).  We can't know if it's (2) without talking to Dennis, and we can determine (3) by running some benchmarks.

Next step would be to profile https://github.com/Project-OSRM/osrm-backend/pull/3354 (I did some quick tests and at least in the benchmark scenario, performance was improved with the corrected curve).  If it performs better for every workload we can think of, let's make the change.

If it turns out that there is some magic performance bonus to the alternative motif, we should at least re-name the `hilbertCode` function call :-P

TheMarex:
My money is on 1. This code never has test coverage and the result passes as "space-filling curve" which is really all the StaticRTree needs. Looking forward to the benchmark results, this could be great on big datasets.

oxidase:
Even the Hilbert linearizion is some sort of heuristic and not the optimal packing. 
I have written a small python script that [dumps R-trees leaves](https://gist.github.com/oxidase/66ccac70893a582f6c968cc44b8fb996).
[...]

Issue #2287, opened 2016-04-20
Turn penalties are being used when route starts at an intersection, not just before it 


lbud:
In writing a cucumber test for a node map
```
| a | b | c |
| d |   | e |
| f | g | h |
```

with ways 
```
ab
ad
bc
ce
df
eh
fg
gh
```
where all ways are `highway=primary`, and equidistant, I discovered that routes `ac`, `ch`, `af`, `fh` were returning different speeds, where I would have expected them to all return the same speed.

This bug also manifests itself in the following node map:
```
|   | i |   | j |   |
| k | a | b | c | l |
|   | d |   | e |   |
| m | f | g | h | n |
|   | o |   | p |   |
```
(with the same ways, in addition to `ai`, `ak`, `fm`, `fo`, `hn`, `hp`, `cj`, `cl`, so that all four intersections were full four-way intersections) — in this case again `ac,` `ch`, `hf`, `fa` all return 59 km/h, wherease `ca`, `hc`, `fh`, `af` return 57 km/h.

[This log](https://gist.github.com/lbud/f28c7cb3e4e394f36402483ed73b6f3a) shows numbers logged out of the [edge_based_graph_factory](https://github.com/Project-OSRM/osrm-backend/blob/develop/src/extractor/edge_based_graph_factory.cpp#L353-L369) for the first case. (The first two way names are `name_table.GetNameForID(edge_data1.name_id)` and `name_table.GetNameForID(edge_data2.name_id)`. The turn angle is apparently being calculated from a possible preceding turn, despite the fact that there is 0 distance traveled on the preceding way — no turn is made as the route starts at an intersection, but a turn penalty is being factored into the first instruction nonetheless.

Bit of an edge case, but 🐛 regardless… : ] @MoKob

danpat:
This looks like a more subtle version of https://github.com/Project-OSRM/osrm-backend/issues/2096.

What I suspect is happening is that we're snapping to the beginning of the segment after the turn, rather than the end of the segment before the turn.  This means that the turn penalty gets included in the duration, and this lowers the average speed.

Which segment gets selected comes purely down to leaf-node sort order in the RTree.  The two segments are exactly the same distance away, but our rtree search will only return a single candidate.

We could possibly fix this by simply adding both segments to the search heap.  There was a related question on the OSRM-Talk mailing list a couple of days ago:

https://lists.openstreetmap.org/pipermail/osrm-talk/2016-April/001180.html

[...]

PR #3354, opened 2016-11-22
Refactor Hilbert values computation

oxidase:
# Issue

The current Hilbert codes is an SFC but not the original Hilbert curve #3352. The PR fixes Hilbert code computation. There are 6 failing test cases, that failing due to reordering of edges in the R-tree. For example, features/testbot/compression.feature:7 returns phantom node that uses 3 segments instead of 4, so the small adjustment of 0.11 meters is not added. Other tie-breaking cases look not so trivial and must be clarified first.

## Tasklist
 - [x] refactor implementation of Hilbert linear codes
 - [x] clarify failing features/car/maxspeed.feature:81
 - [x] clarify failing features/car/maxspeed.feature:100
 - [x] clarify failing features/car/summaries.feature:9
 - [x] clarify failing features/car/traffic_turn_penalties.feature:105
 - [x] clarify failing features/testbot/bearing_param.feature:48
 - [x] review
 - [x] adjust for comments

## Requirements / Relations
[Arndt, Jörg. Matters Computational Ideas, Algorithms, Source Code, 2010. p86](http://www.jjj.de/fxt/fxtbook.pdf)
http://home.pipeline.com/~hbaker1/hakmem/topology.html#item115


danpat:
Quick comparison using `rtree-bench` with a small dataset (`monaco-latest.osm`):

Before this change:
```
Running raw RTree queries (1 result) with 10000 coordinates: Took 14.0459 seconds (14045.9ms)  ->  1.40459 ms/query (14045.9ms)
Running raw RTree queries (10 results) with 10000 coordinates: Took 14.3965 seconds (14396.5ms)  ->  1.43965 ms/query (14396.5ms)
```

After:
```
Running raw RTree queries (1 result) with 10000 coordinates: Took 13.4529 seconds (13452.9ms)  ->  1.34529 ms/query (13452.9ms)
Running raw RTree queries (10 results) with 10000 coordinates: Took 13.6612 seconds (13661.2ms)  ->  1.36612 ms/query (13661.2ms)
```

Looks like the corrected curve shape could give us a modest performance boost due to improved cache locality.

oxidase:
Tests features/testbot/compression.feature:7,  failing features/car/maxspeed.feature:81,  failing features/car/maxspeed.feature:100 use exact values for expectations, but  other ordering could return different edges for phantom nodes and due to some round-off errors introduce 0.11 meters segments that correspond to 0.000001 coordinate precision. Expectation in tests are update to the new values that are still ordering dependent.

oxidase:
Tests features/car/summaries.feature:9, features/car/traffic_turn_penalties.feature:105 fail due to 0-length segments in snapped phantom nodes. 
Must be fixed in #2287

oxidase:
Test features/testbot/bearing_param.feature:48 fail due to different components of the source and the target phantom nodes.

It should not be in real cases, so i would suggest to remove `ha` edge, but just commented out the test row

features/testbot/bearing_param.feature
```diff
@@ -75,7 +75,10 @@ Feature: Bearing parameter
             #| 0    | b  | 189 189  | da          | 180                                   |
             | 0    | 1  | 90 270   | ab,bc,cd,cd    | 0->90,90->0,0->270,270->0           |
             | 1    | d  | 10 10    | bc,bc          | 0->0,0->0                           |
-            | 1    | d  | 90 90    | ab,bc,cd,da,da | 0->90,90->0,0->270,270->180,180->0  |
+            # The target point `a` can belong to either `ab` or `ha` edge.
+            # The latter case leads to "Impossible route between points",
+            # because `ab` and `ha` are in different components
+            #| 1    | d  | 90 90    | ab,bc,cd,da,da | 0->90,90->0,0->270,270->180,180->0 |
```
TheMarex: I'm not sure I understand this. We are not trying to snap to `a`, we are trying to snap to a coordinate close to `1` (which is just a coordinate not linked to an OSM node). It should be something in the middle of `ab` (which has bearing 90°).

The target segment for `d` should be unique as well. It is only contained by one segment (`cd`) with 90°. I'm quite confused why the original test specifies `da` as target segment which has bearing `180` and should not be chosen as segment for `d`.

oxidase: The problem is in the target point for `d`. Here the nearest edges to `d`:
```
input_coordinate = (lon:1.000089, lat:1)
   2,1 "cd" distance 0.111226299998 fwd bearing 270.000007845 invalid, bck bearing 90.0000078449 invalid
   1,5 "da" distance 0.111226299998 fwd bearing 180 invalid, bck bearing 360 invalid
   1,0 "dg" distance 0.111226299998 fwd bearing 270.000000777 invalid, bck bearing 90.0000007767 invalid
   4,5 "ha" distance 10.1215933 fwd bearing 89.9999992235 valid, bck bearing 269.999999223 invalid
   5,6 "ab" distance 10.1215933 fwd bearing 90 valid, bck bearing 270 invalid 0.1112263
```
With the new ordering `ha` is before `ab` and `ha` also has bearing 90, so it is a valid phantom node, but `ha` is in another scc. One possibility would be to introduce point `2` close to `d` on the edge `cd` and use instead of `d`

TheMarex: Ah now I understand. Yes we should fix this test case by adding a middle node `2` on `da`. I think I misunderstood the original test, it seems like we want to snap to `da` because cd has bearing 270° (since it is going in the different direction).

PR #3355, opened 2016-11-23
Improved usability for API documentation

danpat:
This changes adds a Markdown to HTML conversion step for our API docs, making them a whole lot easier to read (mostly due to better formatting).

To generate the documentation, do:

```
npm install
npm run build-api-docs
cd build/docs
python -m SimpleHTTPServer
```

Then browse to http://localhost:8000/

**IMPORTANT**: make sure you're using `npm` V3 or higher.  V2 has a bug that will cause it to pull down about 700MB of dependencies via `npm install` (the new `docbox` module depends on a few large projects).  With `npm` V4, the total install size is about 90MB.  A simple `npm install -g npm` should be all that's required to upgrade.

I'd eventually like to publish these docs on the demo server along with each new OSRM build.

I've done a first pass over `http.md` and `libosrm.md`.  The only tangible changes were adjusting section depths and some re-arranging to fit the hierarchy that Docbox was expecting.

The existing `http.md` file is still perfectly usable alone.  In fact, it had a few layout inconsistencies that using Docbox picked up.

## Tasklist
 - [x] Have Travis build docs (need to upgrade Travis to `npm` V3 or higher)
 - [ ] review
 - [ ] adjust for comments


package.json
```diff
@@ -17,11 +18,18 @@
...
+      "brfs"
+    ]
+  },
   "scripts": {
     "lint": "eslint -c ./.eslintrc features/step_definitions/ features/support/",
     "test": "npm run lint && ./node_modules/cucumber/bin/cucumber.js features/ -p verify",
     "clean-test": "rm -rf test/cache",
-    "cucumber": "./node_modules/cucumber/bin/cucumber.js"
+    "cucumber": "./node_modules/cucumber/bin/cucumber.js",
+    "build-api-docs": "rm -rf build/docs && mkdir -p build/docs/tmp/src && cp -r node_modules/docbox/src/* build/docs/tmp/src && cp -r node_modules/docbox/css build/docs/ && cp -r docs/images build/docs && cp node_modules/docbox/index.html build/docs/tmp && cp docs/src/* build/docs/tmp/src/custom && mkdir -p build/docs/tmp/content && cp docs/*.md build/docs/tmp/content && cd build/docs/tmp && NODE_ENV=production browserify src/index.js | uglifyjs -c -m > ../bundle.js && babel src --out-dir lib && node lib/render.js ../index.html && cd .. && rm -rf tmp"
```
TheMarex: hm can we move this to a script in `./script/build_docs.sh` or something? Kind of hard to read.

package.json
```diff
@@ -7,6 +7,7 @@
     "chalk": "^1.1.3",
     "cucumber": "^1.2.1",
     "d3-queue": "^2.0.3",
+    "docbox": "^1.0.2",
```
TheMarex: Can we put this in `devDependencies`?

daniel-j-h:
Changing the sections in the `http.md` will break all existing URLs - still works with canonicalized URLs.

Just noting here.

daniel-j-h:
I just removed the env vars section from the HTTP docs and polished them a bit.

What I noticed during this: it would be great if we link to the response object docs in the request description already. As in "this will return a \[Waypoint\](#waypoint-anchor)".

danpat:
@daniel-j-h I don't think we need to worry too much about API doc links.  Anyone linking directly to `master` is in for trouble anyway, and because we've moved docs into `git` and away from the Wiki, folks that need stability should be linking to immutable versions (either tags or gitshas) anyway.

PR #3357, opened 2016-11-23
Update Variant to latest Release, closes #3356

### LINKED ISSUE #3356: Update Variant to latest Release
We're currently running with the Mapbox Variant release 1.1.0 from February. Update to 1.1.4.

This should get us (among other fixes and features)

- https://github.com/mapbox/variant/pull/128 `.match` which is great for downstream users (cc @jcoupey)
- https://github.com/mapbox/variant/pull/126 (hashable response)

daniel-j-h:
For #3356. Don't rebase the subtree update - this is one of the rare situations where pressing the merge button does The Right Thing (tm).

### DIFF
--- a/include/extractor/guidance/coordinate_extractor.hpp
+++ b/include/extractor/guidance/coordinate_extractor.hpp
@@ -49,6 +49,17 @@ class CoordinateExtractor
     std::vector<util::Coordinate> GetForwardCoordinatesAlongRoad(const NodeID from,
                                                                  const EdgeID turn_edge) const;
 
+    // a less precise way to compute coordinates along a route. Due to the heavy interaction of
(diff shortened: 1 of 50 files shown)


daniel-j-h:
@TheMarex any hints on how to resolve this? We could merge the variant headers into a single one and manually include this - but it's far from beautiful.

TheMarex:
@daniel-j-h hrm I have no idea how to solve this elegantly.

We could switch back to submodules but this would also give us all the downsides. (e.g. every user not only needs to clone the repo but also remember to run `git submodule init`, `git submodule update` etc) and in combination with branches I have found this to break often.

You could also try to convince `variant` that it doesn't need to bundle mason since it is only used for travis.

TheMarex:
@danpat merged this by hand using the release tarball.

PR #3358, opened 2016-11-23
 Add cache into scripting_environment (new behavior best performance).

fijemax:
#3180

@daniel-j-h,
This is the last behavior I was talking to you in the previous PR.
Without cache feeding, performances increases 0%  and with cache feeding, increases 12-13%.
But i will feed the cache only the first time user need to access at location in lua script, as this way only users using location in luascript will be affected by this changes.

### DIFF
--- a/include/extractor/extraction_containers.hpp
+++ b/include/extractor/extraction_containers.hpp
@@ -16,7 +16,7 @@ namespace osrm
 {
 namespace extractor
 {
-
+class ScriptingEnvironment;
 /**
  * Uses external memory containers from stxxl to store all the data that
  * is collected by the extractor callbacks.
@@ -60,6 +60,10 @@ class ExtractionContainers
     STXXLRestrictionsVector restrictions_list;
     STXXLWayIDStartEndVector way_start_end_id_list;
     std::unordered_map<OSMNodeID, NodeID> external_to_internal_node_id_map;
+    // FIXME This map store the vector id for each osm_id contain in "all_nodes_list"
+    // FIXME add a lazy loading to init this cache when user need location on noderef
+    std::unordered_map<OSMNodeID, osmium::Location> nodeCache;
+
     unsigned max_internal_node_id;
 
     ExtractionContainers();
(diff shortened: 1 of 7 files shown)


daniel-j-h:
There's an exception getting thrown, you can see it e.g. here:

https://travis-ci.org/Project-OSRM/osrm-backend/jobs/178341632#L1548

```
[info] Found 3 turn restriction tags:
[info]   motorcar
[info]   motor_vehicle
[info]   vehicle
terminate called after throwing an instance of 'osrm::util::exception'
terminate called recursively
terminate called recursively
```

daniel-j-h:
I'm also not sure if you always want to use a unordered map for the location cache - if I can remember correctly libosmium has both sparse and dense indices: sparse for small extracts, dense for planet size extracts. The difference is the sparse indices are more or less hash tables, the dense one contiguous vectors.

If someone wants to run this on the planet a unordered map might explode in size. Even a vector of all locations in the planet should be in the tens of GBs.

daniel-j-h:
Here's something @joto just told me in a conversation about your issue: the `osmium` tool already has a `add-locations-to-ways` command which reads a pbf caching all locations and then attaching those locations to the corresponding ways.

Running this on your `.pbf` file before using it with osrm _should_ allow you to access a way locations already without any changes needed.

http://osmcode.org/osmium-tool/

http://docs.osmcode.org/osmium/latest/osmium-add-locations-to-ways.html

fijemax:
Thank you for the solution.
I tested this way, it's work fine however performance is worse than the cache solution.
I need to generate europe, I haven't yet the results of the generation time with this solution on it.
I'll write here the result time.

daniel-j-h:
@fijemax any updates on this? Is the `osmium-tool` solution good enough for you?

frodrigo:
We will use the osmium-tool solution, even it's 3 times slower than the caching in OSRM. Nevertheless we will make a pull request to improve thing with missing location for nodes.

fijemax:
Ok, I close it.

Issue #3359, opened 2016-11-24
Failure to correctly unpack paths for roads that allow u-turns at both ends


MoKob:
Right now we mostly forbid u-turns. In cases where they are explicitly allowed, we fail to unpack paths correctly.

The following scenario:

```
     Scenario: U-Turn In Bearings
        Given the node map
            """
            a 1 b
            """

        And the ways
            | nodes |
            | ab    |

        When I route I should get
            | waypoints | bearings   | route    | turns                    |
            | 1,a       | 90,2 270,2 | ab,ab,ab | depart,turn uturn,arrive |
            | 1,b       | 270,2 90,2 | ab,ab,ab | depart,turn uturn,arrive |
```

Fails to succeed currently.

Based on which direction the shortcut is first added, only one of the two paths will be found correctly.

Right now we see

```
1) Scenario: U-Turn In Bearings - features/testbot/via.feature:331
   Step: When I route I should get - features/testbot/via.feature:341
   Step Definition: features/step_definitions/routing.js:4
   Message:
     Error: Tables were not identical:
         waypoints |     bearings |     route |     turns
     |     1,a |     90,2 270,2 |     ab,ab,ab |     depart,turn uturn,arrive | 
     | (-) 1,b | (-) 270,2 90,2 | (-) ab,ab,ab | (-) depart,turn uturn,arrive | 
     | (+) 1,b | (+) 270,2 90,2 | (+) ab,ab | (+) depart,arrive | 
```

MoKob:
A test case has been added [here](https://github.com/Project-OSRM/osrm-backend/blob/3721f8c8872b17f0b1d00f8ccffe78dfbca21c94/features/testbot/via.feature#L331-L345). Issue up for the taking.

danpat:
This problem was fixed at some point in the last 12 months.  I removed the `@todo` tag in https://github.com/Project-OSRM/osrm-backend/pull/4712, as the tests now pass.

oxidase:
@danpat the issues was fixed by 40d02978858502ed5f08fa7aa69bb9afc72b2515 the problem was in path unpacking.

PR #3360, opened 2016-11-24
fix failure to unpack route with u-turns at both ends

MoKob:
# Issue

https://github.com/Project-OSRM/osrm-backend/issues/3359

## Tasklist
 - [x] review
 - [x] adjust for comments

## Requirements / Relations
 none

### DIFF
--- a/features/testbot/via.feature
+++ b/features/testbot/via.feature
@@ -327,3 +327,19 @@ Feature: Via points
         When I route I should get
             | waypoints | route |
             | a,b,e     |       |
+
+    @todo @3359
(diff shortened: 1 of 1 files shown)


MoKob:
@daniel-j-h same here as over in https://github.com/Project-OSRM/osrm-backend/pull/3364. Would love to add this given the open issue and no-one probably feeling inclined towards picking this up as long as a PR is open for it.

daniel-j-h:
Looks good to merge as long as we keep the original issue open :+1:

Issue #3361, opened 2016-11-24
Handle junction=circular for Not-a-Roundabouts


daniel-j-h:
We have multiple roundabout'ish situations to handle:

- Proper Roundabouts. Think: "at the roundabout take the nth exit".
- Roundabout Intersections. These are roundabouts with up to four ways and turn angles which makes the turns obvious. Think: "at the roundabout turn left".
- Rotaries. These are large and named roundabouts. Think: "at Meinplatz take the nth exit".

There is one additional edge-case which we want to handle now: **Not-a-Roundabout**. There are situations where one of the entering road has right of way, which disqualifies the scenario for being classified as a roundabout. The Wiki has a [section on these Not-a-Roundabouts](http://wiki.openstreetmap.org/wiki/Tag:junction%3Droundabout#Not_a_roundabout):

> In all these cases, the key “junction” is not necessary at all. You need no special tagging.
Most of these Not-a-Roundabouts have a note on them, telling users not to
tag them as junction=roundabout as this is a common mistake. Here are two
examples in Berlin:

- [Bersarinplatz](http://www.openstreetmap.org/way/288767318#map=19/52.51849/13.45355&layers=D
)
- [Kottbusser Tor](http://www.openstreetmap.org/way/159155852#map=19/52.49898/13.41881&layers=D)
- [Strausberger Platz](http://www.openstreetmap.org/way/30825182#map=19/52.51850/13.42896&layers=D)
- [Grosser Stern](http://www.openstreetmap.org/way/4413638#map=18/52.51456/13.35041&layers=D)

The routing engine is not able to issue special instructions since there is no tag to parse. And
parsing note tags for the occurrence of roundabout is not a solution.

---------------------------------------------------------------------------------------------------------------------

After some discussion on the [tagging mailinglist](https://lists.openstreetmap.org/pipermail/tagging/2016-November/030520.html
) we now have a `junction=circular` tag for exactly this. And by now all the roundabouts above are already tagged with it!

Here is its Wiki page for now: https://wiki.openstreetmap.org/wiki/Tag:junction%3Dcircular

Let's handle it!

I'll implement this asap so we can see the change in behavior on the demo server in 1-2 days.

polarbearing:
The funny thing is that in #3362 I find the previous announcements more intuitive than the new ones. As you realise already, the small roads can spoil your exit counting: the service lane here, and the small residential roads at Kotti. 

[...]

PR #3362, opened 2016-11-24
Adds Support for junction=circular Not-a-Roundabouts

MoKob:
Adds support for @daniel-j-h 's child: `junction = circular` https://github.com/Project-OSRM/osrm-backend/issues/3361

## Tasklist
 - [x] pull in changes from https://github.com/Project-OSRM/osrm-backend/tree/junction-circular
 - [x] review
 - [x] adjust for comments
 - [x] requires https://github.com/Project-OSRM/osrm-backend/pull/3363 merged

From chat:
Circular junctions are not behaving as roundabouts. There is a lot of confusion going on in the OSM community, however, often tagging them as roundabouts. A circular junction, or traffic circle/rotary (see e.g. [Wikipedia](https://en.wikipedia.org/wiki/Traffic_circle)) is a junction that looks just like a roundabout, but does not have to follow the usual right of way situations associated with a roundabout.

This PR emits only `Rotary` typed instructions for the flag `circular` opposed to simply treating it as a roundabout. Roundabout-types could be considered dangerous, as they imply a certain right of way behaviour which is not necessarily associated with the intersection.

## Relations
Bases on https://github.com/Project-OSRM/osrm-backend/pull/3363


daniel-j-h:
Please merge my branch https://github.com/Project-OSRM/osrm-backend/tree/junction-circular into this for profile level support and tests. Beware the `junction=circular` are not yet in latest Geofabrik exports (as of writing this). I had to manually export XML files from OSM. Some before / after images:

b-0
b-1

and

s-0
s-1

this one is questionable from a user interface standpoint alone - we don't report or show lanes in roundabouts / circular roundabouts at the moment at all.

TheMarex:
Can you explain why it is necessary to have an internal flag `circular` that is distinct from `roundabout`? Form what I can see we emit a `roundabout` instruction in both cases?

MoKob:
The reasoning here is that a `circular` instruction does not follow `roundabout` priorities. It is a named circular junction. The rotary is exactly that (while it can follow roundabout rules as well). We should not emit `roundabout` though.

The difference is that the roundabout instruction could imply specific `right of way` implications that are not correct (e.g. due to stop signs). I want to avoid emitting `at the roundabout turn right`. We should only emit `at grosser stern take the 4th exit`.

profiles/foot.lua
```diff
@@ -160,6 +160,9 @@ function way_function (way, result)
   if "roundabout" == junction then
     result.roundabout = true
   end
+  if "circular" == junction then
+    result.circular = false
```
daniel-j-h: `false`?

unit_tests/extractor/graph_compressor.cpp
```diff
@@ -16,6 +16,30 @@ using namespace osrm::extractor;
...
+            SPECIAL_EDGEID,
+            0,
+            false,
+            false,
+            false,
+            false,
+            true,
+            TRAVEL_MODE_INACCESSIBLE,
+            INVALID_LANE_DESCRIPTIONID};
+}
```
daniel-j-h: :+1:

profiles/foot.lua
MoKob: well... I'd argue that foot should neither announce roundabout nor circular. After all: what is an exit here? Sidewalk?

daniel-j-h: But we do for rotaries right? Above we set the roundabout flag, which means large names roundabouts will be announced as rotaries already.

MoKob: yes... but I'd rather set the `roundabout` flag to false as well.

daniel-j-h: Okay, then maybe set both to false and add a comment. Otherwise this PR looks fine.

(Strictly speaking post-processing should handle this, right? as in emitting different instructions based on the duration between steps)

PR #3363, opened 2016-11-24
handle service roads next to roundabouts - ambiguous exit... :(

MoKob:
# Issue

Turns out parking isles in exit counting is a really bad idea. However, we create ambiguities here.

## Tasklist
 - [x] review
 - [x] adjust for comments

## Requirements / Relations
 none

### DIFF
--- a/include/extractor/guidance/road_classification.hpp
+++ b/include/extractor/guidance/road_classification.hpp
@@ -65,7 +65,7 @@ class RoadClassification
   public:
     // default construction
     RoadClassification()
-        : motorway_class(0), link_class(0), may_be_ignored(1),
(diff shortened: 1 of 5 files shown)


include/extractor/guidance/road_classification.hpp
```diff
@@ -65,7 +65,7 @@ class RoadClassification
   public:
     // default construction
     RoadClassification()
-        : motorway_class(0), link_class(0), may_be_ignored(1),
+        : motorway_class(0), link_class(0), may_be_ignored(0),
```
daniel-j-h: Huh, why the change here?

PR #3364, opened 2016-11-24
add test for falsely classified obvious turn

MoKob:
# Issue

https://github.com/Project-OSRM/osrm-backend/issues/3365

This PR adds only a test-case and is up for the taking.

## Tasklist
 - [x] review
 - [x] adjust for comments

## Requirements / Relations
 none


MoKob:
@daniel-j-h I would love to clean up some of my open PRs. Given that I just marked the additional case as TODO/added a label for the respective issue. Can you have a look and if you don't have further remarks we could merge this in now to be solved later.

daniel-j-h:
Same here - as long as we keep the original issue open this is ready to merge :+1:

PR #3367, opened 2016-11-26
Rework RPATH and building as subproject

TheMarex:
# Issue

This would make logic in node-osrm obsolete that tries to do this after the fact.

## Tasklist
 - [x] review
 - [x] adjust for comments

## Requirements / Relations


CMakeLists.txt
```diff
@@ -640,9 +653,28 @@ function(JOIN VALUES GLUE OUTPUT)
...
-# Set up variables, then write to pkgconfig file
-JOIN("${OSRM_DEFINES}" " " OSRM_DEFINES_STRING)
-JOIN("-I${OSRM_INCLUDE_PATHS}" " -I" OSRM_INCLUDE_PATHS_STRING)
+JOIN("${OSRM_DEFINES}" " " TMP_OSRM_DEFINES)
+set(LibOSRM_CXXFLAGS "${OSRM_CXXFLAGS} ${TMP_OSRM_DEFINES}")
+
+if(BUILD_AS_SUBPROJECT)
+  set(LibOSRM_CXXFLAGS "${LibOSRM_CXXFLAGS}" PARENT_SCOPE)
+  set(LibOSRM_INCLUDE_DIR "${CMAKE_CURRENT_SOURCE_DIR}/include" PARENT_SCOPE)
+  set(LibOSRM_LIBRARY_DIR "${CMAKE_CURRENT_BINARY_DIR}" PARENT_SCOPE)
```
DerDakon: This should already be globally available as OSRM_BINARY_DIR (created by the project() call).

CMakeLists.txt
```diff
@@ -640,9 +653,28 @@ function(JOIN VALUES GLUE OUTPUT)
...
+  set(LibOSRM_LIBRARIES "osrm" PARENT_SCOPE)
+  set(LibOSRM_DEPENDENT_LIBRARIES "${ENGINE_LIBRARIES}" PARENT_SCOPE)
+  set(LibOSRM_INCLUDE_DIRS "${CMAKE_CURRENT_SOURCE_DIR}/include;${CMAKE_CURRENT_SOURCE_DIR}/include/osrm;${CMAKE_CURRENT_SOURCE_DIR}/third_party;${DEPENDENCIES_INCLUDE_DIRS}" PARENT_SCOPE)
+  set(LibOSRM_LIBRARY_DIRS "${LibOSRM_LIBRARY_DIR}" PARENT_SCOPE)
+endif()
+
+# pkgconfig defines
+set(PKGCONFIG_OSRM_CXXFLAGS "${LibOSRM_CXXFALGS}")
```
DerDakon: not "FALGS", FLAGS.

CMakeLists.txt
```diff
@@ -640,9 +653,28 @@ function(JOIN VALUES GLUE OUTPUT)
...
+set(PKGCONFIG_LIBRARY_DIR "${CMAKE_INSTALL_PREFIX}/lib")
```
DerDakon: I suggest using the GNUInstallDirs module, which will automatically handle e.g. lib64, too.

CMakeLists.txt
```diff
@@ -640,9 +653,28 @@ function(JOIN VALUES GLUE OUTPUT)
...
+JOIN("${OSRM_DEFINES}" " " TMP_OSRM_DEFINES)
+set(LibOSRM_CXXFLAGS "${OSRM_CXXFLAGS} ${TMP_OSRM_DEFINES}")
+
+if(BUILD_AS_SUBPROJECT)
+  set(LibOSRM_CXXFLAGS "${LibOSRM_CXXFLAGS}" PARENT_SCOPE)
+  set(LibOSRM_INCLUDE_DIR "${CMAKE_CURRENT_SOURCE_DIR}/include" PARENT_SCOPE)
+  set(LibOSRM_LIBRARY_DIR "${CMAKE_CURRENT_BINARY_DIR}" PARENT_SCOPE)
+  set(LibOSRM_LIBRARIES "osrm" PARENT_SCOPE)
+  set(LibOSRM_DEPENDENT_LIBRARIES "${ENGINE_LIBRARIES}" PARENT_SCOPE)
+  set(LibOSRM_INCLUDE_DIRS "${CMAKE_CURRENT_SOURCE_DIR}/include;${CMAKE_CURRENT_SOURCE_DIR}/include/osrm;${CMAKE_CURRENT_SOURCE_DIR}/third_party;${DEPENDENCIES_INCLUDE_DIRS}" PARENT_SCOPE)
```
daniel-j-h: Is there a difference between `;`-delimiting manually here and calling CMake's `append`? Does this work on Linux, macOS, Windows?

TheMarex: `list(APPEND ...)` didn't work with `PARENT_SCOPE`. Need to investigate if there is a better way to do this. Seems like just using space delimiters and `"` around each string should work as well.

DerDakon: 'list(APPEND foo bar)', 'set(foo "${foo};bar")' and 'set(foo ${foo} bar)' all do the same. The list() version has the advantage that it will "just work" when the items contain whitespace, which something you need to pay attention to with set(). set() has the advantage of PARENT_SCOPE.

TheMarex:
@DerDakon thanks for your feedback. :+1: Merging this.

PR #3372, opened 2016-11-29
Adds failing tests for directional access overrides, discovered in #3345

daniel-j-h:
We don't handle directional overrides.

This adds an initial failing (!) test for the situation outlined by @emiltin in https://github.com/Project-OSRM/osrm-backend/pull/3345#discussion_r89841432.

### DIFF
--- a/features/car/access.feature
+++ b/features/car/access.feature
@@ -220,3 +220,15 @@ Feature: Car - Restricted access
         Then routability should be
             | highway | toll        | bothw |
             | primary | yes         |       |
+
+    Scenario: Car - directional access tags
+        Then routability should be
+            | highway | access | access:forward | access:backward | forw | backw |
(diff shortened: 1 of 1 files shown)


daniel-j-h:
With https://github.com/Project-OSRM/osrm-backend/issues/3345 merged these tests can go in as regression tests. They pass now.

emiltin:
i think this might already be covered by https://github.com/Project-OSRM/osrm-backend/blob/efb97f91b9ab86c991d8990458610012791f0059/features/car/access.feature#L33?

daniel-j-h:
They were failing before see https://github.com/Project-OSRM/osrm-backend/pull/3345#discussion_r89841432 - better to have some regression tests in place :)

emiltin:
the existing tests also failed, they were just marked with @todo. the test check the same thing: access:forward and access:backward.

PR #3373, opened 2016-11-29
Adding nodes's and ways's version access in lua binding.

fijemax:
Adding nodes's and ways's version access in lua binding for users who need this information.

## Tasklist
 - [x] Add node version access.
 - [x] Add way version access.
 - [x] requires https://github.com/Project-OSRM/osrm-backend/issues/3452 to be resolved first

Signed-off-by: FILLAU Jean-Maxime <jean-maxime.fillau@mapotempo.com>

### DIFF
--- a/include/extractor/extractor_config.hpp
+++ b/include/extractor/extractor_config.hpp
@@ -102,6 +102,8 @@ struct ExtractorConfig
     bool generate_edge_lookup;
     std::string edge_penalty_path;
     std::string edge_segment_lookup_path;
+
+    bool use_metadata;
 };
 }
 }
--- a/src/extractor/extractor.cpp
+++ b/src/extractor/extractor.cpp
@@ -130,7 +130,11 @@ int Extractor::run(ScriptingEnvironment &scripting_environment)
         auto extractor_callbacks = std::make_unique<ExtractorCallbacks>(extraction_containers);
 
         const osmium::io::File input_file(config.input_path.string());
-        osmium::io::Reader reader(input_file, osmium::io::read_meta::no);
+
+        osmium::io::Reader reader(
+            input_file,
+            (config.use_metadata ? osmium::io::read_meta::yes : osmium::io::read_meta::no));
+
         const osmium::io::Header header = reader.header();
 
         unsigned number_of_nodes = 0;
--- a/src/extractor/scripting_environment_lua.cpp
+++ b/src/extractor/scripting_environment_lua.cpp
@@ -189,7 +189,9 @@ void Sol2ScriptingEnvironment::InitContext(LuaScriptingContext &context)
         "id",
         &osmium::Way::id,
         "get_nodes",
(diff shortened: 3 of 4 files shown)


daniel-j-h:
Are you sure the version tag is still available since we're using the `read_meta::no` flag for osmium:

https://github.com/Project-OSRM/osrm-backend/blob/fe5cc55b0e98204b7e56037fba7c2f8b2adbd9b2/src/extractor/extractor.cpp#L132

fijemax:
Hi @daniel-j-h 
You are right, can I add an option at extractor to parse meta ?

daniel-j-h:
The reason we added this flag is a pbf parsing speed up by 10-20%.

https://github.com/osmcode/libosmium/releases/tag/v2.10.0

frodrigo:
We do computation from way geometries in lua to adjust speed. As computation are time consuming, we want to cache the result for next run of osrm-extract. But for caching the computation result we need to have the way version. So in our case reading metata will result in speed up.

If we code parsing meta data of pbf as as option, and add it to this pull request, do you merge it ?

daniel-j-h:
Yes one way to achieve this would be to add a new command line option e.g. `osrm-extract --with-osm-metadata` and keep the metadata reader flag disabled by default unless explicitly specified.

TheMarex:
Adding a flag would work for me, but what happens when this function is called from a profile and there is no meta data? Does it crash?

fijemax:
I tested this use case, if this function is called but there isn't meta data function return 0.

fijemax:
Hi @daniel-j-h, I added the command line option `osrm-extract --with-osm-metadata` for enable meta data reading. @TheMarex  if the flag is disable and function is called from a profile zero is return.

TheMarex:
@fijemax Looks good. Added this to the 5.6 milestone since we have feature freeze for 5.5.

TheMarex:
This needs a rebase, current conflicts in master.

fijemax:
Hi @heMarex,
I can test node version since Sol2 integration.
I open an issue #3452

TheMarex:
This is blocked by #3452?

daniel-j-h:
Yes - theoretically it should work, as it worked with Luabind before.

But we want to be sure #3452 is resolved first.

fijemax:
Hi,
@daniel-j-h @TheMarex
The #3452 is solved.
I rebased and solved conflict of this PR.

daniel-j-h:
Perfect, thanks - sorry it took so long.

Could you format your changes to `scripting_environment_lua.cpp` using the `scripts/format.sh` script (or clang-format-3.8 manually)? Then this is ready to go.

https://travis-ci.org/Project-OSRM/osrm-backend/jobs/188805267#L409-L410

fijemax:
it's done.

daniel-j-h:
Thank you, this just landed!

Issue #2120, opened 2016-03-18
LTO_AVAILABLE check failing for clang


springmeyer:
Currently the `-flto` configure check is failing for clang++ linux builds on travis: https://travis-ci.org/Project-OSRM/osrm-backend/jobs/116873025#L1200

The way LTO works with clang on linux is:
- `-flto` needs to be passed to CXXFLAGS
- `-flto` needs to be passed to LDFLAGS
- The gold linker needs to either be `/usr/bin/ld` or you need to pass `-B/path/to/custom/bin` if the gold linker lives at `/path/to/custom/bin/ld`
- libLLVMgold.so needs to be installed `apt-get install llvm-3.7-dev`

The reason that the configure check is failing in our builds right now is that `-flto` is not being passed to the linking flags for the [check_cxx_compiler_flag check here](https://github.com/Project-OSRM/osrm-backend/blob/b1c84b598f7969eecccf890effd99dc57ed7141a/CMakeLists.txt#L118). And therefore the linker is not put in LTO mode for that configure check and the `.o` file with LTO support cannot be read. The error in the cmake logs is: `CMakeFiles/cmTryCompileExec3478428907.dir/src.cxx.o: file not recognized: File format not recognized`

One workaround I've found is to do `export LDFLAGS="-flto"` which then triggers the `check_cxx_compiler_flag` to use `-flto` in the linker flags and then the binary produced is valid and can be run. But ideally some better workaround could be found, however my knowledge of cmake is too limited to begin to know what to do.

/cc @daniel-j-h for help.

daniel-j-h:
Good that you're raising this issue! Over the last weekend I compiled a freshly releases LLVM 3.8 (and then all of osrm-backend's dependencies with Clang/libc++ for ABI compatibility), and only after reading this issue checked for lto in the compilation log --- and indeed the CMake buildsystem did not activate it! Let's fix this once and for all! Here's a quick brain dump for Clang/lto:

I played around with `ld.gold` three month ago (but back then in combination with GCC). Updating `ld` so that it points to `ld.gold`seems like the best bet as in:
https://gist.github.com/daniel-j-h/ed3bf9e5cb458b012b5a

For Clang/lto I _think_ we also need to use the lto-aware binutils as in:
http://llvm.org/docs/GoldPlugin.html#quickstart-for-using-lto-with-autotooled-projects

Setting `env CXXFLAGS='-flto' LDFLAGS='-flto'` is fine when the user knows what to do.
[...]

PR #3377, opened 2016-11-29
Fix -flto support with clang++

springmeyer:
# Issue

Targeting #2120.

The gcc -flto support works currently by:

  - Telling cmake about gcc-ar and gcc-ranlib (https://github.com/Project-OSRM/osrm-backend/blob/ef087f963d2f50ce7bc7b7019c4cf7170cf8005e/CMakeLists.txt#L221-L223)
  - Gcc internally using its `lto-wrapper` for linking

For clang++ and -flto to work we need to:

 - 1) Add the `-flto` flag in both compile flags and link flags manually (since no equivalent `lto-wrapper` exists for linking with clang++)
 - 2) Use recent binutils for gold linker that understands clang++ lto format
 - 3) Provide the absolute paths to `llvm-ar` and `llvm-ranlib`

This PR builds **i** into the CMakeLists.txt and **ii** + **iii** into the `.travis.yml`.

This PR also:

  - Starts passing `-flto` and related flags to the pkg config file. Without this the example.cpp will not link because it is not compiled with `-flto` and the lto-enabled `libosrm.a` cannot be used.
  - Changes the warning about ENABLE_MASON possibly not working with lto on some systems to only display on linux (since OS X and windows are not impacted)

## Tasklist
 - [x] review
 - [x] adjust for comments

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -143,7 +143,6 @@ before_install:
       # and therefore don't need to us ld.gold or llvm tools for linking
       # for debug builds
       if [[ ${BUILD_TYPE} == 'Release' ]]; then
-        export CMAKE_EXTRA="-DCMAKE_AR=$(which llvm-ar) -DCMAKE_RANLIB=$(which llvm-ranlib)"
         ./third_party/mason/mason install binutils 2.27
         export PATH=$(./third_party/mason/mason prefix binutils 2.27)/bin:${PATH}
       fi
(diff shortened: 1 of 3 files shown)


TheMarex:
Grabbing this to cmakefy this a little bit more.

TheMarex:
@springmeyer I moved some of this logic to cmake to make it work consistent with the gcc binutils. Otherwise this looks good to me. :+1:

springmeyer:
@TheMarex great, thanks for the cmake help! I think before merging this we should test node-osrm against it. My hope is that the `-flto` flags will be picked up by node-osrm and it should link correctly without changes. But we should make sure. I'll kick off a node-osrm build against this branch once travis is green with your changes.

springmeyer:
confirmed that node-osrm master is working against this branch and the build flags show `-flto` properly coming through from osrm-backend's pkg-config file: https://travis-ci.org/Project-OSRM/node-osrm/jobs/180491933. I think this is ready to merge.

TheMarex:
@springmeyer :bow: thanks for you help. Merging.

Issue #3013, opened 2016-10-06
Rip out Luabind for good


daniel-j-h:
Until [Javascript-based profiles](https://github.com/Project-OSRM/osrm-backend/pull/2687) lands (and then probably even longer due to I/O constraints) we still have to support Lua. In https://github.com/Project-OSRM/osrm-backend/pull/2647 we gained the profile interface abstraction which limits the changes needed to the Lua scripting environment.

In https://github.com/Project-OSRM/osrm-backend/pull/3011#issuecomment-251924968 @TheMarex hit yet another Luabind issue which I'm just adding on top of the Lua / Luabind pile listed in https://github.com/Project-OSRM/osrm-backend/issues/1974#issuecomment-189187428. Luabind is not maintained, is the main source of [user problem we're seeing](https://github.com/Project-OSRM/osrm-backend/issues?q=is%3Aissue+label%3ACompilation+is%3Aclosed) and package manager's ship a version from 2010, which some heavily patch some don't.

This ticket is about ripping out Luabind to get rid of the constant issues we're seeing.

The idea is to drop in the modern and idiomatic single header [sol2](https://github.com/ThePhD/sol2) project which would allow to completely rip out Luabind and no longer having to work around Luabind's restrictions (e.g. enum classes).

Documentation: http://sol2.readthedocs.io/en/latest/

Tasks
- [x] build small self-contained example and try out 1/ registering objects, enums, enum classes writing from Lua into C++ objects 2/ check if Lua function exists from C++ (I _think_ these are the two main use-cases we're having)
- [x] check how sol2 behaves with Lua51, Lua52
- [x] drop `sol2.hpp` amalgamation single header into third_party
- [x] change scripting env. implementation

Issue #1777, opened 2015-11-17
Move processing loop to lua


TheMarex:
Currently we have a context switch for parsing each way. We could move the loop code to lua and save a lot of context switches.

PR #3382, opened 2016-11-30
Replace Luabind with Sol2

karenzshea:
# Issue

https://github.com/Project-OSRM/osrm-backend/issues/3013

To quote @daniel-j-h 

> We're running on Luabind and it's a constant pain not only for us but also for users. As soon as we manage to port our code we will switch over to sol2 and never look back.

## Tasklist
 - [x] Investigate test failures
 - [x] Figure out if returning an empty string instead of `null` is backwards incompatible
 - [x] Port forward any recent changes to scripting_environment_lua.cpp
 - [x] Check feature parity wrt. what we bind to Lua. Go over old / new impl. and check off.
 - [x] Add to update-dependencies script
 - [x] Check compatibility with lua versions
 - [ ] update relevant [Wiki pages](https://github.com/Project-OSRM/osrm-backend/wiki) I have a list of pages to update once pr has been merged
 - [x] review
 - [x] adjust for comments

## Requirements / Relations
This is a pre-MLD task.


src/extractor/scripting_environment_sol2.cpp
```diff
@@ -0,0 +1,447 @@
...
+        "lat", &latToDouble<ExternalMemoryNode>);
+
+    context.state.new_usertype<util::Coordinate>("Coordinate",
+             "lon", &lonToDouble<util::Coordinate>,
+             "lat", &latToDouble<util::Coordinate>);
+
+    context.state.new_type<RasterDatum>("RasterDatum",
+            "datum", &RasterDatum::datum,
+            "invalid_data", &RasterDatum::get_invalid);
+    */
```
daniel-j-h: Same here before we can merge this.

src/extractor/scripting_environment_sol2.cpp
```diff
@@ -20,6 +20,29 @@
...
+    }
+
+    const osmium::Node &node;
+};
+
+struct LuaWay
+{
+    auto get_value_by_key(const char *key, const char *default_value = nullptr) const
+    {
+        return way.get_value_by_key(key, default_value);
```
emiltin: note the recent change to return null instead of empty tags

CMakeLists.txt
```diff
@@ -91,6 +91,7 @@ endif()
 
 include_directories(BEFORE ${CMAKE_CURRENT_BINARY_DIR}/include/)
 include_directories(BEFORE ${CMAKE_CURRENT_SOURCE_DIR}/include/)
+include_directories(SYSTEM ${CMAKE_CURRENT_SOURCE_DIR}/third_party/)
```
daniel-j-h: also below: can you rebase on master, all access restricted tags should be gone by now - if not please ping me

karenzshea: Already rebased, which is how I picked up on the missing sol directory includes. I'll update how the sol header is saved and update in the cmakelists now.

include/extractor/extraction_way.hpp
```diff
@@ -46,6 +46,7 @@ struct ExtractionWay
         roundabout = false;
         circular = false;
         is_startpoint = true;
+        ignore_in_grid = false;
```
MoKob: Could this be left-over code? I wanted to check for it but could not find it in the remaining parts of the PR.

karenzshea: I added this field. Sol was erroring in places where we set this property in [bicycle/foot profiles](https://github.com/Project-OSRM/osrm-backend/blob/0e6863aec1eb1d60d9318cb166a66cd0c20f23e3/profiles/bicycle.lua#L242) but hadn't registered a corresponding field in the bindings. I'm not sure how luabind handled this before.

MoKob: I did some digging. Seems the profiles are out of date. That flag was removed in https://github.com/Project-OSRM/osrm-backend/commit/faa880d60a162728f8a9dd06c87373b8924d0837

MoKob: Capturing from chat: Next actions:

- [ ] remove dead flag (ignore_in_grid) from foot/bicycle profile
- [ ] remove bindings for (ignore_in_grid)

karenzshea:
I checked compatibility with Lua 5.1 and 5.2 by building osrm with both versions and running tests. They check out OK; if there's any further testing I should do, lmk @MoKob @daniel-j-h

karenzshea:
@daniel-j-h @TheMarex review?

CMakeLists.txt
```diff
@@ -282,15 +282,15 @@ endif()
 
 # Configuring compilers
 if(${CMAKE_CXX_COMPILER_ID} STREQUAL "Clang")
-  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wall -Wextra -pedantic -Wuninitialized -Wunreachable-code -Wstrict-overflow=2 -U_FORTIFY_SOURCE -D_FORTIFY_SOURCE=2 -fPIC -fcolor-diagnostics")
+  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wall -Wextra -pedantic -Wuninitialized -Wunreachable-code -Wstrict-overflow=2 -U_FORTIFY_SOURCE -D_FORTIFY_SOURCE=2 -fPIC -fcolor-diagnostics -ftemplate-depth=1024")
```
TheMarex: 😿

include/extractor/scripting_environment_sol2.hpp
```diff
@@ -1,33 +1,30 @@
-#ifndef SCRIPTING_ENVIRONMENT_LUA_HPP
-#define SCRIPTING_ENVIRONMENT_LUA_HPP
-
-#include "extractor/scripting_environment.hpp"
+#ifndef SCRIPTING_ENVIRONMENT_SOL2_HPP
```
TheMarex: Hm why did we rename this file? Sure the library that we use for executing lua changed but this is still for executing lua code.

src/extractor/scripting_environment_sol2.cpp
```diff
@@ -0,0 +1,473 @@
...
+
+    sol::function segment_function = context.state["segment_function"];
+
+    if (segment_function.valid())
+    {
+        segment_function(source, target, distance, weight);
+    }
+}
+
+void Sol2ScriptingContext::processNode(const osmium::Node &node, ExtractionNode &result)
```
TheMarex: Naming convention would be `ProcessNode`.

src/extractor/scripting_environment_sol2.cpp
```diff
@@ -0,0 +1,473 @@
...
+{
+    BOOST_ASSERT(state.lua_state() != nullptr);
+
+    sol::function node_function = state["node_function"];
+
+    node_function(node, result);
+}
+
+void Sol2ScriptingContext::processWay(const osmium::Way &way, ExtractionWay &result)
```
TheMarex: Naming convention would be `ProcessWay`.

CMakeLists.txt
```diff
@@ -473,6 +465,33 @@ else()
     find_package(GDAL)
   endif()
 
+  FIND_PACKAGE(Lua 5.2 EXACT)
```
TheMarex: I think we can now simplify all this logic down to `find_package(Lua 5.1)` which will look for lua 5.1+. Sol seems to have a good support for all lua versions, unlike luabind.

include/extractor/extraction_way.hpp
```diff
@@ -96,6 +96,7 @@ struct ExtractionWay
     bool roundabout;
     bool circular;
     bool is_startpoint;
+    bool ignore_in_grid;
```
TheMarex: This seems like an accidental change.

CMakeLists.txt
daniel-j-h: clang has a default limit of 256, gcc has 1024 - this is fine :+1:

CMakeLists.txt
daniel-j-h: The check here was because of the Lua 5.3 memory issue and not specifically for Luabind - until we confirm the Lua 5.3 memory issues are gone simply by using sol we should keep it.

include/extractor/extraction_way.hpp
daniel-j-h: yes this has to go - not only here but also in the profiles and the bindings, see https://github.com/Project-OSRM/osrm-backend/pull/3382#pullrequestreview-11385147

daniel-j-h:
Looks like your rebase got a few issues - it's `util/log.hpp` on master

> /home/travis/build/Project-OSRM/osrm-backend/src/extractor/scripting_environment_sol2.cpp:14:34: fatal error: util/simple_logger.hpp: No such file or directory
>
> #include "util/simple_logger.hpp"

daniel-j-h:
Windows build are failing:

> C:\projects\osrm\third_party\sol2\sol2/sol.hpp(771): fatal error C1083: Cannot open include file: 'lua.hpp': No such file or directory [C:\projects\osrm\build\EXTRACTOR.vcxproj]

Seems like the C++ `lua.hpp` wrapper can't be found on Windows. cc @BergWerkGIS

karenzshea:
Timing the switch:
- ran extraction the germany geofabrik extract
- r3.8xlarge machine, 32 cores

| branch | extraction time | parsing time |
| --- | --- | --- |
| bye-luabind | real    15m41.392s, user    39m24.131s, sys     16m7.835s | 130.264 |
| master | real    16m25.099s, user    58m34.169s, sys     28m56.980s | 196.322s |

daniel-j-h:
These are great numbers and will lift some pressure from the guidance code!

(Unrelated to this PR: the next step would be to instead of context switching for every node or way, we switch on a nodes-chunk and ways-chunk basis => https://github.com/Project-OSRM/osrm-backend/issues/1777)

BergWerkGIS:
@daniel-j-h where is `lua.hpp` coming from on other platforms?
On Windows we clone https://github.com/LuaDist/lua
The only one I could find in this repo is https://github.com/LuaDist/lua/blob/lua-5.1/etc/lua.hpp
But we don't include it - would it be the one we need?

daniel-j-h:
Yes this looks like the file we need. It's just setting up the includes for the C name mangling.

BergWerkGIS:
@daniel-j-h could you try again - I've updated and rebuilt that binary deps package.

`lua.hpp` is now in `osrm-deps\libs\include`:

image

daniel-j-h:
Rebased against master and force-pushed. Let's see if Appveyor is happy with it now. cc @karenzshea

daniel-j-h:
@BergWerkGIS thanks, this resolved it. Ready for merge now! :tada:

daniel-j-h:
It's gone, Luabind is gone - I can't believe it! :)

Updating the Wiki pages now.

PR #3321, opened 2016-11-16
Improve I/O error handling

danpat:
# Issue

This PR is to address #3047.  It adds some basic wrappers around our file I/O which include consistent error handling and useful exception messages on error.

No file formats are changed.

## Tasklist
 - [x] add regression / cucumber cases (see docs/testing.md)
 - [x] review
 - [x] adjust for comments

PR #3383, opened 2016-11-30
FileWriter and some FileReader refactoring

duizendnegen:
# Issue

`FileWriter` counterpart to `FileReader` and one example implementation. 

## Tasklist
 - [x] review
 - [x] adjust for comments

## Requirements / Relations
- https://github.com/Project-OSRM/osrm-backend/pull/3321 replaces all `iostream` with `FileReader` 
- Part of https://github.com/Project-OSRM/osrm-backend/issues/2242, the `File` abstraction

### DIFF
--- a/include/storage/io.hpp
+++ b/include/storage/io.hpp
@@ -8,7 +8,6 @@
 #include <boost/filesystem/fstream.hpp>
 #include <boost/iostreams/seek.hpp>
 
-#include <cerrno>
 #include <cstring>
 #include <tuple>
 #include <type_traits>
@@ -23,7 +22,7 @@ namespace io
 class FileReader
 {
   private:
-    const std::string filename;
+    const boost::filesystem::path filepath;
     boost::filesystem::ifstream input_stream;
 
   public:
@@ -47,16 +46,16 @@ class FileReader
     {
     }
 
-    FileReader(const boost::filesystem::path &filename_, const FingerprintFlag flag)
-        : filename(filename_.string())
+    FileReader(const boost::filesystem::path &filepath_, const FingerprintFlag flag)
+        : filepath(filepath_)
     {
-        input_stream.open(filename_, std::ios::binary);
+        input_stream.open(filepath, std::ios::binary);
         if (!input_stream)
-            throw util::exception("Error opening " + filename + ":" + std::strerror(errno));
+            throw util::exception("Error opening " + filepath.string());
 
         if (flag == VerifyFingerprint && !ReadAndCheckFingerprint())
         {
-            throw util::exception("Fingerprint mismatch in " + filename);
+            throw util::exception("Fingerprint mismatch in " + filepath.string());
         }
     }
 
@@ -77,10 +76,10 @@ class FileReader
         {
             if (result.eof())
             {
-                throw util::exception("Error reading from " + filename +
+                throw util::exception("Error reading from " + filepath.string() +
                                       ": Unexpected end of file");
             }
-            throw util::exception("Error reading from " + filename + ": " + std::strerror(errno));
+            throw util::exception("Error reading from " + filepath.string());
         }
     }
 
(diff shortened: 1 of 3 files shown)


duizendnegen:
I'd be happy to get input as for how tests should be conducted for this specific part of infrastructure - iirc @daniel-j-h mentioned that all Cucumber tests cover the IO layer implicitly anyway, and I couldn't find additional existing unit tests for the `FileReader`.

include/storage/io.hpp
```diff
@@ -144,6 +144,80 @@ class FileReader
...
+	};
+
+
+	FileWriter(const std::string &filename, const FingerprintFlag flag)
+		: FileWriter(boost::filesystem::path(filename), flag)
+	{
+	}
+
+	FileWriter(const boost::filesystem::path &filename_, const FingerprintFlag flag)
+		: filename(filename_.string())
```
daniel-j-h: Why don't we store a boost::fs::path to begin with? :)

include/storage/io.hpp
```diff
@@ -144,6 +144,80 @@ class FileReader
...
+	{
+		output_stream.open(filename_, std::ios::binary);
+		if (!output_stream)
+			throw util::exception("Error opening " + filename + ":" + std::strerror(errno));
```
daniel-j-h: Would it make sense to have separate IO exceptions? (Check the Gist in https://github.com/Project-OSRM/osrm-backend/issues/2242#issuecomment-261223565).

(btw all those commends apply to the FileReader, too cc @ghoshkaj @danpat)

include/storage/io.hpp
```diff
@@ -144,6 +144,80 @@ class FileReader
...
+			"bytewise writing requires trivially copyable type");
+#endif
+
+		if (count == 0)
+			return true;
+
+		const auto &result = output_stream.write(reinterpret_cast<char *>(src), count * sizeof(T));
+		if (!result)
+		{
+			throw util::exception("Error writing to " + filename + ": " + std::strerror(errno));
```
daniel-j-h: Hrm does an error here really set errno?

include/util/io.hpp
```diff
@@ -54,7 +54,8 @@ template <typename simple_type>
 bool serializeVectorIntoAdjacencyArray(const std::string &filename,
                                        const std::vector<std::vector<simple_type>> &data)
 {
-    std::ofstream out_stream(filename, std::ios::binary);
+	storage::io::FileWriter file(filename, storage::io::FileWriter::HasNoFingerprint);
```
daniel-j-h: There's a clang-format script at scripts/format.sh - once your done with changes it would be great to run this.

(if you don't have clang-format we can do this for you before merging, too)

include/storage/io.hpp
duizendnegen: Yea I generally agree, however, I'd suggest opening another issue for that - at the moment `exception` (https://github.com/Project-OSRM/osrm-backend/blob/master/include/util/exception.hpp) is marked `final`, we'd be better of differentiating exceptions across the board in a refactor.

include/storage/io.hpp
duizendnegen: Probably not, will remove.

include/storage/io.hpp
duizendnegen: Sure, will do so across both `FileReader` and `FileWriter`.

duizendnegen:
@daniel-j-h for further review

daniel-j-h:
Travis fails because of variant include path. I think @danpat fixed it yesterday.

Try squashing, rebasing, force pushing.

include/storage/io.hpp
```diff
@@ -170,6 +171,79 @@ class FileReader
...
+#endif
+
+        if (count == 0)
+            return true;
+
+        const auto &result = output_stream.write(reinterpret_cast<char *>(src), count * sizeof(T));
+        if (!result)
+        {
+            throw util::exception("Error writing to " + filepath.string() + ": " +
+                                  std::strerror(errno));
```
daniel-j-h: I still can't find the documentation for o(f)streams setting errno.

daniel-j-h:
cc @danpat @ghoshkaj please check in the context of your IO endeavour and then merge at will.

include/storage/io.hpp
duizendnegen: Yea I dropped it from the `FileWriter` but forgot to remove it from the `FileReader`, will do so.

duizendnegen:
Rebased, squashed, let's do this! 🎉

include/storage/io.hpp
duizendnegen: FYI it does set `errno`, at least, based on the tests it does.

oxidase: @duizendnegen it is an ad-hoc platform-dependent result that may not correspond to a stream method call. 

@daniel-j-h errno shows the thread-local result of the last syscall and without knowing stream internals it can be unrelated and [will not be fixed in gcc](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=71950).

oxidase: @duizendnegen it is an ad-hoc platform-dependent result that may not correspond to a stream method call. 

@daniel-j-h errno shows the thread-local result of the last syscall and without knowing stream internals it can be incorrect and [will not be fixed in gcc](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=71950).

oxidase: @duizendnegen it is an ad-hoc platform-dependent result that may not correspond to a stream method call. 

@daniel-j-h errno shows the thread-local result of the last syscall and without knowing stream internals it can be unrelated and [will not be fixed in gcc](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=71950).

duizendnegen: Thanks for the pointer @oxidase, I updated the PR again to reflect these learnings.

TheMarex:
Looks good to me 👍 merging.

Issue #3046, opened 2016-10-12
Exceptions during parallel task invocation are hidden


danpat:
When this error occurs:
https://github.com/Project-OSRM/osrm-backend/blob/77931ffe2405b579c0e9c8cd8772957aef75a114/src/contractor/contractor.cpp#L712

The exception is being hidden because this code is being executed under `tbb::parallel_invoke`.

We should take some care to modify this so that we can at least extract the original error message.

Exception handling performed by `tbb::parallel_invoke` is described here:
https://software.intel.com/en-us/node/506317

danpat:
Fixed by https://github.com/Project-OSRM/osrm-backend/issues/3385

PR #3385, opened 2016-12-01
Clean up exception handling and make logging a bit more consistent.

danpat:
This PR was initially to address #3046, but it has grown a bit in scope.  Once you peel the corner back.....

It bundles a few changes:

  - deprecates `SimpleLogger().Write(XXX)` in favour of the simpler `Log(XXX)` interface
  - adds a new `logERROR` level to differentiate between actual errors and warnings.  This should make logfiles easier to grep for errors.
  - proper unwrapping of exceptions thrown by a few of the error-prone `tbb::*` parallel workers
  - a new utility macro `SOURCE_REF` that is intended to be used along with throwing exceptions - it contains the filename and line number where the exception is thrown, to help making log analysis easier.  Existing `util::exceptions` have been updated to include this info
  - throws exceptions in two cases where we were simply performing `return 1` previously
  - remove use of `std::cerr/cout` in favour of `util::Log()` in most locations - this makes output more consistent and makes logging easier to find in the code
  - full run of `clang-format`

The overall intent of this change is to:
  - make logging code a little less verbose
  - catch some errors (around tbb::parallel_invoke) that we haven't been logging properly
  - include more useful information in exceptions that get thrown

There is a new logger called `util::UnbufferedLog()` that is used to support `doing something ..... done` style logging.  It flushes the final newline when the logging object is destroyed, so adapting existing to to use this involved creating some scopes and indenting some chunks of code.

Overall, this PR is mostly a big search/replace job, with a few extra scoping blocks added.

## Tasklist
 - [x] review
 - [ ] adjust for comments


include/util/exception.hpp
```diff
@@ -40,8 +41,14 @@ namespace util
 class exception final : public std::exception
 {
   public:
-    explicit exception(const char *message) : message(message) {}
-    explicit exception(std::string message) : message(std::move(message)) {}
+    explicit exception(const char *message_) : message(message_)
+    {
+        util::SimpleLogger().Write(logERROR) << message;
```
MoKob: I am not sure here if this is such a good idea. @daniel-j-h, whats your take? The logger is a synchronised entity and from what I can tell not exception save. So throwing an exception while throwing an exception looks like opening the doors of hell to me.

daniel-j-h: Yeah throwing while there's an exception active is a bad idea. That's the reason your dtor must not throw, too.

src/contractor/contractor.cpp
```diff
@@ -329,7 +332,10 @@ parse_segment_lookup_from_csv_files(const std::vector<std::string> &segment_spee
                           speed); //
 
                 if (!ok || it != last)
-                    throw util::exception{"Segment speed file " + filename + " malformed"};
+                {
+                    const std::string message{"Segment speed file " + filename + " malformed on line "+std::to_string(line_number)};
```
daniel-j-h: format

src/contractor/contractor.cpp
```diff
@@ -756,7 +768,8 @@ EdgeID Contractor::LoadEdgeExpandedGraph(
         std::ofstream geometry_stream(geometry_filename, std::ios::binary);
         if (!geometry_stream)
         {
-            throw util::exception("Failed to open " + geometry_filename + " for writing");
+            const std::string message{"Failed to open " + geometry_filename + " for writing"};
+            throw util::exception(std::move(message));
```
daniel-j-h: Huh, why does this even compile? You shouldn't be able to move from a const object.

daniel-j-h: Or better: the move will cast this to a `const string&&` which does not allow for the string's move ctor to trigger.

TheMarex:
I like the new ERROR vs. WARNING formatting however why did catching `tbb::capture_exception` not work in this case?

https://www.threadingbuildingblocks.org/docs/help/tbb_userguide/Exceptions_and_Cancellation.html

Catching this at the function scope and re-throwing as `osrm::exception` should work, no?

danpat:
@TheMarex The problem is that `tbb::exception` is *supposed* to contain the original exception, but doesn't always seem to.  This ticket originated from a production problem where `e.what()` for the `tbb::exception` was a blank.

My intent is to try to get the error logged as soon as we notice it.

src/util/log.cpp
```diff
@@ -75,21 +75,29 @@ SimpleLogger::~SimpleLogger()
         switch (level)
         {
         case logWARNING:
-            std::cerr << (is_terminal ? RED : "") << os.str() << (is_terminal ? COL_RESET : "")
-                      << std::endl;
+        case logERROR:
+            std::cerr << buffer.str();
```
MoKob: shouldn't all of these write to `stream`?

TheMarex:
Looks good from my side. 👍

src/util/log.cpp
danpat: @MoKob ah, good point - there was some implied behaviour here.  The base class assumes std::cout/cerr are the final destination, but subclasses can override that.

I've clarified the logic.  `Log` still finishes up with cerr/cout, but subclasses that override the stream will use that instead.

PR #3389, opened 2016-12-01
Try clang 3.9 lto

springmeyer:
experiemental, don't merge.


springmeyer:
Closing. Conclusion is that clang++ 3.8.1 and clang++ 4.0.0 both produce LTO that work with binutils 2.27 and latest binutils. But clang++ 3.9.0 is broken. We will hope that 3.9.1, when released will support lto again in a way that does not segfault binutils. Tracking using clang++ 3.8.1 in #3377

PR #3391, opened 2016-12-01
Use correct newlines during Windows test runs.

danpat:
Windows builds were failing a test case because of bad newline assumptions.

## Tasklist
 - [ ] review
 - [ ] adjust for comments


danpat:
Replaced by #3390 (which actually works).

PR #3393, opened 2016-12-02
Enforce iteration order so that tiles are encoded consistently

### LINKED ISSUE #3343: Size of tile depends on architecture
Looking at the node-osrm tests, they fail because the size of a tile is no consistent over linux vs. OSX. This should not be the case since we are using protobuf, which should provide platform independent encoding.

https://travis-ci.org/Project-OSRM/node-osrm/jobs/177262797#L1688

/cc @danpat

danpat:
# Issue

This fixes #3343.  Because PBF uses variable-length encoding, differences in feature order (due to sorting differences on different platforms) leads to differently encoded files, even though they're otherwise equivalent.

This PR enforces the sort order for tile layers, features and properties, which causes tiles to be encoded the same on all platforms.

## Tasklist
 - [x] review
 - [x] adjust for comments

### DIFF
--- a/src/engine/plugins/tile.cpp
+++ b/src/engine/plugins/tile.cpp
@@ -357,6 +357,26 @@ Status TilePlugin::HandleRequest(const std::shared_ptr<datafacade::BaseDataFacad
         return offset;
     };
 
+    // In order to ensure consistent tile encoding, we need to process
+    // all edges in the same order.  Differences in OSX/Linux/Windows
+    // sorting methods mean that GetEdgesInBox doesn't return the same
+    // ordered array on all platforms.
+    // GetEdgesInBox is marked `const`, so we can't sort the array itself,
+    // instead we create an array of indexes and sort that instead.
+    std::vector<std::size_t> sorted_edge_indexes(edges.size(), 0);
+    std::iota(sorted_edge_indexes.begin(), sorted_edge_indexes.end(), 0); // fill with 1,2,3,...N
+
+    // Now, sort that array based on the edges list, using the u/v node IDs
+    // as the sort condition
+    std::sort(sorted_edge_indexes.begin(),
+              sorted_edge_indexes.end(),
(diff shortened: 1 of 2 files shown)


src/engine/plugins/tile.cpp
```diff
@@ -357,6 +357,27 @@ Status TilePlugin::HandleRequest(const std::shared_ptr<datafacade::BaseDataFacad
...
 
+    // In order to ensure consistent tile encoding, we need to process
+    // all edges in the same order.  Differences in OSX/Linux/Windows
+    // sorting methods mean that GetEdgesInBox doesn't return the same
+    // ordered array on all platforms.
+    // GetEdgesInBox is marked `const`, so we can't sort the array itself,
+    // instead we create an array of indexes and sort that instead.
+    std::vector<std::size_t> sorted_edge_indexes(edges.size(), 0);
+    for (std::size_t i = 0; i < sorted_edge_indexes.size(); i++)
+        sorted_edge_indexes[i] = i;
```
daniel-j-h: std::iota

daniel-j-h:
:+1:

danpat:
TIL - `iota`.  Also note to self, it's not `itoa`, I was very confused there for a minute.

springmeyer:
@danpat - did you try using `std::stable_sort` instead of `sort`? refs https://github.com/mapbox/mapnik-vector-tile/issues/107#issuecomment-105683106

danpat:
@springmeyer I thought about it, but didn't dive too deep.  The inconsistently sorted data is coming out of OSRM's RTree (that's where the real platform-dependent ordering is introduced), but that's some pretty hot code, so introducing `std::stable_sort` there seemed like a bad idea - there are many other users of the rtree that don't care about this level of ordering.

I opted to contain the performance hit/ordering to the tile plugin, and because the problem ordering is in data fed from outside the plugin, `std::stable_sort` isn't able to consistentify those results.

PR #3394, opened 2016-12-02
Cleaning Up included headers/Exposed Functions

MoKob:
# Issue

Inclusions in some files are getting out of hand, introducing circular dependencies and similar problems. This PR aims at cleaning up a bit.

This PR removes the dedicated toolkits (which introduced a series of circular dependencies) and moves their functionality into dedicated headers or close to the data.

A side effect: 7 seconds faster build time on my machine vs master (2:04 vs 2:11). Small, but I'll take it.

I've also identified a few issues with duplicated code, removing distance calculations (side effect: a single additional call to the `cos` function).

## Tasklist
 - [x] review
 - [x] adjust for comments

## Requirements / Relations


src/extractor/guidance/intersection_generator.cpp
```diff
@@ -58,8 +56,15 @@ IntersectionGenerator::ComputeIntersectionShape(const NodeID node_at_center_of_i
...
-        getLaneCountAtIntersection(node_at_center_of_intersection, node_based_graph);
+    const auto intersection_lanes = [&]() {
+        std::uint8_t lanes = 0;
+        for (const EdgeID onto_edge :
+             node_based_graph.GetAdjacentEdgeRange(node_at_center_of_intersection))
+            lanes = std::max(
+                lanes,
+                node_based_graph.GetEdgeData(onto_edge).road_classification.GetNumberOfLanes());
+        return lanes;
+    }();
```
daniel-j-h: this is how clang-format formats this? blargh

src/extractor/guidance/intersection_generator.cpp
```diff
@@ -58,8 +53,15 @@ IntersectionGenerator::ComputeIntersectionShape(const NodeID node_at_center_of_i
     const util::Coordinate turn_coordinate = node_info_list[node_at_center_of_intersection];
 
     // number of lanes at the intersection changes how far we look down the road
-    const auto intersection_lanes =
-        getLaneCountAtIntersection(node_at_center_of_intersection, node_based_graph);
+    const auto intersection_lanes = [&]() {
+        std::uint8_t lanes = 0;
```
TheMarex: `number_of_lanes` would be a more precise naming.

src/extractor/guidance/intersection_generator.cpp
TheMarex: What about using `std::accumulate()` here instead of a manual `for` loop?

MoKob:
Turns out we are missing some modifications that have been done to the toolkits in between. Investigating.

MoKob:
Never mind, turned out to be outdated binaries on my side

PR #3395, opened 2016-12-02
Abstracts away over differences in IntersectionView and Intersection.

daniel-j-h:
CRTP-away the differences in `IntersectionView` and `Intersection` for re-usable intersection features.

## Tasklist
 - [x] review
 - [x] adjust for comments

Usage:

```c++
struct MyIntersection : EnableIntersectionOps<MyIntersection> {

};
```

Done.

We require `MyIntersection` having at least the member attributes from
`IntersectionViewData` but don't enforce a inheritance hierarchy.


MoKob:
Beautiful :)

Issue #3396, opened 2016-12-02
assertion while osrm-extract stage.


gardster:
I got an assertion message
```
[info] generating edge-expanded edges
[assert] /osrm-backend/src/extractor/guidance/coordinate_extractor.cpp:1087
in: double osrm::extractor::guidance::CoordinateExtractor::ComputeInterpolationFactor(double, double, double) const: distance_to_first < desired_distance
```
when I built master and called the osrm-extract for map with NYC inside.

I called osrm-extract with parameters: ```-p osrm-backend/profile.lua --generate-edge-lookup```

gardster:
extracting with HEAD~50 version helped me.

MoKob:
I've just fixed two errors in the coordinate extraction over in #3397 . Using the new york extract of today it works fine on my end, when it did not before.

TheMarex:
@MoKob on the demo server we just hit:

```
in: double osrm::extractor::guidance::CoordinateExtractor::ComputeInterpolationFactor(double, double, double) const: distance_to_second >= desired_distance
[assert] /mnt/ebs/osrm/2016-12-02-173001/src/extractor/guidance/coordinate_extractor.cpp:1090
```

Is this the same bug? The line number is different. Happend around 0:36 UTC not sure which gitsha.

gardster:
I recontract the map and now it works on master. Thank you for the quick fix!

MoKob:
From the assertion alone it's impossible to tell if that is the same error or not. Could be, but I can't tell. I've fixed two errors leading to calls with invalid arguments, but I can't tell if the above error is due to the these.

PR #3397, opened 2016-12-02
fix extraction of coordinate along a way

MoKob:
# Issue
resolves https://github.com/Project-OSRM/osrm-backend/issues/3396

## Tasklist
 - [x] review
 - [x] adjust for comments

## Requirements / Relations
none


danpat:
@MoKob the fix looks good.  I'll ticket adding some unit tests for this specifically as part of future refactoring - it's a perfect candidate for some self-contained tests, but the `private` visibility makes it impossible with the current layout.

PR #3399, opened 2016-12-04
Make absolute fixtures path

oxidase:
# Issue

If the build directory is not in CMAKE_SOURCE_DIR than some tests that use relative paths may fail.
PR adds compile definition `FIXTURES_DIR="${CMAKE_SOURCE_DIR}/unit_tests/fixtures"`  to `extractor-tests`.

Also PR makes `rm $(DATA_NAME).*` non-failing if `$(DATA_NAME).*` files do not exist.

## Tasklist
 - [x] review
 - [x] adjust for comments

 - [x] New exceptions, logging, errors - are messages distinct enough to track down in the code if they get thrown in production on non-debug builds?
~~- [ ] Relevant docs updated~~
~~- [ ] Changelog entry if required~~
~~- [ ] Impact on the API surface~~
~~- [ ] If HTTP/libosrm.o is backward compatible features, bump the minor version~~
~~- [ ] File format changes require at minor release~~
~~- [ ] If old clients can't use the API after changes, bump the major version~~

### DIFF
--- a/test/data/Makefile
+++ b/test/data/Makefile
@@ -15,7 +15,7 @@ PROFILE:=$(PROFILE_ROOT)/car.lua
 all: $(DATA_NAME).osrm.hsgr
 
 clean:
-	rm $(DATA_NAME).*
+	-rm $(DATA_NAME).*
 
 $(DATA_NAME).osm.pbf:
 	wget $(DATA_URL) -O $(DATA_NAME).osm.pbf
--- a/unit_tests/CMakeLists.txt
+++ b/unit_tests/CMakeLists.txt
@@ -48,6 +48,8 @@ if(NOT WIN32 AND NOT Boost_USE_STATIC_LIBS)
   add_definitions(-DBOOST_TEST_DYN_LINK)
 endif()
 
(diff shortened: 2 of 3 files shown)


test/data/Makefile
```diff
@@ -15,7 +15,7 @@ PROFILE:=$(PROFILE_ROOT)/car.lua
 all: $(DATA_NAME).osrm.hsgr
 
 clean:
-	rm $(DATA_NAME).*
+	-rm $(DATA_NAME).*
```
TheMarex: What does the `-` before `rm` do?

unit_tests/CMakeLists.txt
```diff
@@ -48,6 +48,8 @@ if(NOT WIN32 AND NOT Boost_USE_STATIC_LIBS)
   add_definitions(-DBOOST_TEST_DYN_LINK)
 endif()
 
+target_compile_definitions(extractor-tests PRIVATE COMPILE_DEFINITIONS FIXTURES_DIR="${CMAKE_SOURCE_DIR}/unit_tests/fixtures")
```
TheMarex: Can we prefix this define with `OSRM_`?

test/data/Makefile
TheMarex: Ah cool didn't know about that functionality. 👍

unit_tests/CMakeLists.txt
oxidase: Ok

Issue #3076, opened 2016-10-18
Normalize turn penalty value


TheMarex:
After a discussion with @nickcordella we should change the way we save turn penalties per edge. Basically we need to make sure that the weighted average of turn penalties equals to zero. The weighting would be happening based on how many of the turns take the left/right/straight turn. Since we don't know that in advance, we would weight every turn equally (hence yielding the unweighted average).

Currently we are introducing a constant offset to every edge that screws our average time estimation. This is a problem exactly at [this location](https://github.com/Project-OSRM/osrm-backend/blob/master/src/contractor/contractor.cpp#L837) when we mix it with an unbiased turn penalty source. That would make all edges that ingest them much faster, since they are missing the offset. Furthermore there is a scaling problem with edges that are only covered by lua turn penalty data.

The proposed fix is to update fix [this line](https://github.com/Project-OSRM/osrm-backend/blob/master/src/extractor/edge_based_graph_factory.cpp#L552) to really only save the normalized turn penalty.

/cc @danpat

danpat:
Hmm.  We'll need to loop over each intersection twice to achieve this.  The first iteration will be to retrieve all the turn penalties from Lua.  Once we've done that, we can calculate the mean, then we can write out the adjusted penalties.

Taking this task.

TheMarex:
@danpat grabbed this task from you.

danpat:
After some discussion with @TheMarex and @nickcordella, pushing this task to 5.6.0, we're not clear on the value yet, and it's preferable not to backtrack on a feature if we can at all avoid it.

PR #3402, opened 2016-12-05
Normalize turn penalties

TheMarex:
# Issue

Addresses #3076.

## Tasklist
 - [x] review
 - [x] adjust for comments

 - [x] ~~Relevant docs updated~~
 - [x] ~~Changelog entry if required~~
 - [x] ~~Impact on the API surface~~
   - [x] ~~If HTTP/libosrm.o is backward compatible features, bump the minor version~~
   - [x] ~~File format changes require at minor release~~
   - [x] ~~If old clients can't use the API after changes, bump the major version~~

/cc @danpat

### DIFF
--- a/src/extractor/edge_based_graph_factory.cpp
+++ b/src/extractor/edge_based_graph_factory.cpp
@@ -19,6 +19,7 @@
 #include <boost/assert.hpp>
 #include <boost/numeric/conversion/cast.hpp>
 
+#include <numeric>
 #include <algorithm>
 #include <cmath>
 #include <fstream>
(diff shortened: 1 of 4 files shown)


src/extractor/edge_based_graph_factory.cpp
```diff
@@ -449,8 +450,47 @@ void EdgeBasedGraphFactory::GenerateEdgeExpandedEdges(
...
+                    return turn_penalty;
+                };
+
+            std::vector<int32_t> turn_penalties(intersection.size());
+            std::transform(intersection.begin(),
+                           intersection.end(),
+                           turn_penalties.begin(),
+                           turn_to_penalty);
+            const int64_t total_turn_penalties = std::accumulate(turn_penalties.begin(), turn_penalties.end(), 0);
+            const int32_t average_turn_penalty = total_turn_penalties / intersection.size();
```
MoKob: Could this `floor` here imply problems for us?

src/extractor/edge_based_graph_factory.cpp
```diff
@@ -449,8 +450,47 @@ void EdgeBasedGraphFactory::GenerateEdgeExpandedEdges(
...
+
+            for (const auto turn_index : util::irange<std::size_t>(0, intersection.size()))
             {
+                const auto &turn = intersection[turn_index];
+                const auto turn_penalty = turn_penalties[turn_index];
+
                 // only keep valid turns
                 if (!turn.entry_allowed)
```
MoKob: what are the implications of having `invalid` turns in the set we compute `turn_penalties` on? Can invalid instructions (I am thinking of the currently miss-used `DirectionModifier::UTurn` for invalid turns) modify the normalisation for the worse?

src/extractor/edge_based_graph_factory.cpp
MoKob: Should this only divide by the number of valid turns? Or how do we define the turn penalty here?

TheMarex: Good catch, fixed this by using a filter on intersection now, since we are not interested in invalid turns at that point anymore.

MoKob:
Looks good to me

MoKob:
Maybe squash before merge, but otherwise it looks good to go.

TheMarex:
I broke something while fixing this. Needs another debugging run. :-/

TheMarex:
Arg I just realized this is not that simple. The edge weights and the turn penalty need to sum up the value in edge based edge.

So we would need to re-distribute the bias the we exclude from the turn penalty to the single segments that make up an edge-based-node.

TheMarex:
Worked around this problem by saving the "bias": The difference between the penalty + segment weights, which can't be updated.

TheMarex:
@danpat major problem with this implementation is that we now save an additional value and that we break the data format to the previous PR.

I'm not sure this is the right method to implement this, about this would unblock people from experimenting with this feature more.

TheMarex:
This is stale and need a general discussion how we can change the modeling of turn penalties without breaking the current API.

Issue #3401, opened 2016-12-05
Demo server assertion hit


TheMarex:
```
terminate called without an active exception	

in: osrm::util::Coordinate osrm::extractor::guidance::CoordinateExtractor::ExtractRepresentativeCoordinate(NodeID, EdgeID, bool, NodeID, uint8_t, std::vector<osrm::util::Coordinate>) const: is_valid_result(result)

[assert] /mnt/ebs/osrm/2016-12-04-173001/src/extractor/guidance/coordinate_extractor.cpp:368
```

Got this on Sun, 04 Dec 2016 19:50:05 GM.

/cc @MoKob

danpat:
@TheMarex I'm pretty sure that https://github.com/Project-OSRM/osrm-backend/pull/3397 hadn't yet been merged when the last demo server build kicked off - @MoKob's fix probably wasn't part of that code.

A new build should start tonight using the latest code, we should get confirmation that it's fixed in a couple of days.

TheMarex:
@danpat if you look at the date of the checkout `2016-12-04-17300` it is from Sunday, three days after this was merged.

danpat:
@TheMarex oh, good point, I didn't notice that.

That raises another question - what happened to the extraction run after `2016-11-30-173001`?  Was `2016-12-04-173001` really the next time we tried?

PR #3403, opened 2016-12-05
fix errors in coordinate extractor due to duplicated coordinates

MoKob:
# Issue

resolves https://github.com/Project-OSRM/osrm-backend/issues/3401

## Tasklist
 - [x] review
 - [x] adjust for comments

 - [x] New exceptions, logging, errors - are messages distinct enough to track down in the code if they get thrown in production on non-debug builds?
 - [x] Relevant docs updated
 - [x] Impact on the API surface
   - [x] If HTTP/libosrm.o is backward compatible features, bump the minor version
   - [x] File format changes require at minor release
   - [x] If old clients can't use the API after changes, bump the major version

## Requirements / Relations
none

### DIFF
--- a/features/guidance/turn-angles.feature
+++ b/features/guidance/turn-angles.feature
@@ -317,30 +317,30 @@ Feature: Simple Turns
     Scenario: Curved Turn At Cross
         Given the node map
             """
-                                                h
-                                                |
-                                                |
-                                                |
-                                                |
-                                                |      . b - - - - - - - - - - - - - - - - - - a
-                                                |    .
-                                                |  .
-                                                | c
-                                                |
-                                                |
-                                                |
-                                                d
-                                                |
-                                                |
-                                                |
-                                              e |
-                                             .  |
-                                          .     |
-            g - - - - - - - - - - - - - f       |
-                                                |
-                                                |
-                                                |
-                                                i
+                                                          h
+                                                          |
+                                                          |
+                                                          |
+                                                          |
+                                                          |      . b - - - - - - - - - - - - - - - - - - - - - - - a
+                                                          |    .
+                                                          |  .
+                                                          | c
+                                                          |
+                                                          |
+                                                          |
+                                                          d
+                                                          |
+                                                          |
+                                                          |
+                                                        e |
+                                                       .  |
+                                                    .     |
+            g - - - - - - - - - - - - - - - - - - f       |
+                                                          |
+                                                          |
+                                                          |
+                                                          i
             """
 
         And the ways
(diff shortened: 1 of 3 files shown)


features/guidance/turn-angles.feature
```diff
@@ -317,30 +317,30 @@ Feature: Simple Turns
...
-                                                |
-                                                |
-                                                |
-                                                i
+                                                          h
+                                                          |
+                                                          |
+                                                          |
+                                                          |
+                                                          |      . b - - - - - - - - - - - - - - - - - - - - - - - a
```
MoKob: with the fixed curve detection, this test wasn't modelled long enough. Had to change it to test the same information as before

features/guidance/turn-angles.feature
```diff
@@ -627,12 +627,14 @@ Feature: Simple Turns
...
 
+
     #http://www.openstreetmap.org/#map=18/52.56251/13.32650
+    @todo
     Scenario: Curved Turn on Separated Directions
         Given the node map
             """
                 e                               d
-                f - - - - - - - - - - - - - - - c - - - - - - - - - - - j
+                f                               c - - - - - - - - - - - j
```
MoKob: this was never in here

features/guidance/turn-angles.feature
```diff
@@ -680,6 +682,58 @@ Feature: Simple Turns
             | j,o       | Kapweg,Kapweg,Kapweg | depart,continue uturn,arrive |
             | a,i       | Kurt,Kurt,Kurt       | depart,continue uturn,arrive |
 
+    #http://www.openstreetmap.org/#map=18/52.56251/13.32650
```
MoKob: same scenario as above, but now without the additional short segment before the turn

src/extractor/guidance/coordinate_extractor.cpp
```diff
@@ -352,40 +370,25 @@ util::Coordinate CoordinateExtractor::ExtractRepresentativeCoordinate(
          * We distinguish between turns that simply model the initial way of getting onto the
          * destination lanes and the ones that performa a larger turn.
          */
-        const double offset =
-            std::min(0.5 * considered_lanes * ASSUMED_LANE_WIDTH, 0.2 * segment_distances.back());
-        coordinates = TrimCoordinatesToLength(std::move(coordinates), offset, segment_distances);
+        coordinates =
+            TrimCoordinatesToLength(std::move(coordinates),
+                                    2 * distance_to_skip_over_due_to_coordinate_inaccuracies,
```
TheMarex: What about `distance_epsilon`, this name maybe a little bit long, heh.

src/extractor/guidance/coordinate_extractor.cpp
```diff
@@ -709,16 +715,17 @@ bool CoordinateExtractor::IsCurve(const std::vector<util::Coordinate> &coordinat
             return std::make_tuple(
                 true, 1, get_deviation(coordinates.front(), coordinates.back(), coordinates[1]));
 
-        const auto maximum_itr =
+        const auto one_past_maximum_itr =
```
TheMarex: `itr` -> `iter`

src/extractor/guidance/coordinate_extractor.cpp
danpat: I dunno, I actually think this is helpful naming - in 6 months from now, when someone else is looking at this code, the intent is pretty clear.

Makes it easy to search for too :-)

src/extractor/guidance/coordinate_extractor.cpp
```diff
@@ -337,6 +337,24 @@ util::Coordinate CoordinateExtractor::ExtractRepresentativeCoordinate(
         return result;
     }
 
+    if (IsDirectOffset(coordinates,
```
danpat: Would you be able to add a comment that describes why `IsDirectOffset` needs to happen before `IsCurve`?  That's not obvious to me, and it seems like an important part of this change.

src/extractor/guidance/coordinate_extractor.cpp
chaupow: :D or name it `distance_epsilon` and put the explanation in a comment.

gotta agree with danpat tho. its intent is very clear.

Issue #3404, opened 2016-12-05
Segmentation fault on demo server


TheMarex:
The query below will crash the demo server:

```
https://router.project-osrm.org/table/v1/driving/polyline(egs_Iq_aqAppHzbHulFzeMe\`EuvKpnCglA)?radiuses=20;20;20;20;20
```

oxidase:
```
[debug] req: /table/v1/driving/polyline(egs_Iq_aqAppHzbHulFzeMe\`EuvKpnCglA)?radiuses=20;20;20;20;20
[warn] [server error] code: basic_string::at: __n (which is 35) >= this->size() (which is 35), uri: /table/v1/driving/polyline(egs_Iq_aqAppHzbHulFzeMe%5C%60EuvKpnCglA)?radiuses=20;20;20;20;20
```

danpat:
Fixed by https://github.com/Project-OSRM/osrm-backend/pull/3406

PR #3406, opened 2016-12-06
Decoding should not fail for incomplete polyline strings (#3404)

oxidase:
# Issue

#3404 

The problem was in an assumption that a polyline string always have pairs (dlat, dlng).
Moved index checks before reading polyline symbols.
 
There is no end-of-number checks, so "incorrect" coordinates can be used if the last symbol has value >=0x20.

## Tasklist
 - [x] review
 - [x] adjust for comments

 - [ ] Relevant docs updated
 - ~~[ ] Impact on the API surface~~

## Requirements / Relations

Possible fails in
1) correct lattitude, longitude is missing
2) no end-of-number (0 5th bit) marker in the last character

### DIFF
--- a/src/engine/polyline_compressor.cpp
+++ b/src/engine/polyline_compressor.cpp
@@ -2,6 +2,7 @@
 
 #include <algorithm>
 #include <boost/assert.hpp>
+#include <climits>
 #include <cmath>
 #include <cstddef>
 #include <cstdlib>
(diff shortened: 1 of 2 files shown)


src/engine/polyline_compressor.cpp
```diff
@@ -61,24 +61,25 @@ std::vector<util::Coordinate> decodePolyline(const std::string &geometry_string)
 
     while (index < len)
     {
-        int b, shift = 0, result = 0;
-        do
+        int b = 0x20, shift = 0, result = 0;
```
MoKob: would it be possible to describe the logic behind this magical number in a comment around here?

src/engine/polyline_compressor.cpp
```diff
@@ -61,24 +61,25 @@ std::vector<util::Coordinate> decodePolyline(const std::string &geometry_string)
...
+        while (b >= 0x20 && index < len)
```
MoKob: can you explain if the switch here does change anything in the order/count of execution? The first loop looks like it should be executed exactly as often as before. I am totally fine with the change, I just wonder if there is a reason other than style?

src/engine/polyline_compressor.cpp
```diff
@@ -61,24 +61,25 @@ std::vector<util::Coordinate> decodePolyline(const std::string &geometry_string)
...
-        } while (b >= 0x20 && index < len);
+        }
         int dlat = ((result & 1) != 0 ? ~(result >> 1) : (result >> 1));
         lat += dlat;
 
+        b = 0x20;
         shift = 0;
         result = 0;
-        do
+        while (b >= 0x20 && index < len)
```
MoKob: so is this the actual error, executing the second part, if we don't have enough chars?

src/engine/polyline_compressor.cpp
```diff
@@ -61,24 +61,25 @@ std::vector<util::Coordinate> decodePolyline(const std::string &geometry_string)
...
         {
             b = geometry_string.at(index++) - 63;
```
MoKob: while we are touching this, do you think we could also refactor the loop to use lamdas? Both while loops do the same, right?

unit_tests/engine/geometry_string.cpp
```diff
@@ -22,14 +25,11 @@ BOOST_AUTO_TEST_CASE(decode)
...
-    util::Coordinate coord4(util::FloatLongitude{10.3}, util::FloatLatitude{10.03});
-    util::Coordinate coord5(util::FloatLongitude{10.4}, util::FloatLatitude{10.04});
-
-    // Put the test coordinates into the vector for comparison
-    std::vector<util::Coordinate> cmp_coords = {coord1, coord2, coord3, coord4, coord5};
+    std::vector<util::Coordinate> cmp_coords = {{Lng{10.0}, Lat{10.00}},
+                                                {Lng{10.1}, Lat{10.01}},
+                                                {Lng{10.2}, Lat{10.02}},
+                                                {Lng{10.3}, Lat{10.03}},
+                                                {Lng{10.4}, Lat{10.04}}};
```
MoKob: @TheMarex I believe you pointed out that you feel negatively about having abbreviations in the code. Would this fall into the definition, or would it be alright in the unit tests here? I would be fine with having them here.

unit_tests/engine/geometry_string.cpp
```diff
@@ -69,21 +68,75 @@ BOOST_AUTO_TEST_CASE(encode6)
...
 
     const auto encodedPolyline = encodePolyline<1000000>(cmp_coords.begin(), cmp_coords.end());
 
     BOOST_CHECK_EQUAL(encodedPolyline, polyline);
 }
 
+BOOST_AUTO_TEST_CASE(polyline_sign_check)
```
MoKob: should we maybe start documenting these tests on what they guard against? So what would be a possible error we could see here?

src/engine/polyline_compressor.cpp
oxidase: no, it just to have either `do/while` or `while` loop style

src/engine/polyline_compressor.cpp
oxidase: Ok, changed to named constants

src/engine/polyline_compressor.cpp
oxidase: moved to lambda

unit_tests/engine/geometry_string.cpp
oxidase: reverted, anyway clang format makes own formatting

unit_tests/engine/geometry_string.cpp
oxidase: The test checks a negative sign branch in `result = ((result & 1) == 1) ? ~(result >> 1) : (result >> 1);`

MoKob:
Looks great!

Issue #1623, opened 2015-08-21
[Trip Plugin] Computing round trips with fixed start and end point


chaupow:
It should be possible to set a fixed start and end point when computing roundtrip.

i.e.
A round trip that starts at location A and ends at location B and on the way other locations C, D, E, F, ... should be visited is needed. Currently the trip plugin does not support this. It should.

Implementation detail:

It should work when the distance between A and B is set to zero in the distance table.
This is not planned before osrm v4.8.0

jcoupey:
Depending on your heuristic behaviour, setting a cost of 0 from B to A won't necessary be sufficient to ensure that B will be placed just before A in the TSP solution. 
But also adding an artificially huge cost from any other point to A should do the trick, this is how I implemented it myself.

JMGGarcia:
Hey! Is there any plan of implementing this in the future? I think it's an important feature that should exist.

daniel-j-h:
No plans for the near future; ticket is open to track the feature.

chaupow:
@ghoshkaj and me drafted an idea to implement this. Here's a write down of our plan as well as a failed attempt to have a formal proof but I think this could count as an unformal proof:

**Given:**

* _n_ points {A, B, C, D, ....} 
* a distance table with distances between all of these points

**Wanted:** 

A shortest trip that 

* visits all points {A, B, C, D, ...}
* starts at A and ends at B

**Claim:**

* We merge the two nodes A and B to a single node BA 
* We create a new distance table with 
    * any distances from BA to another node being the distance from A to another node: 
`dist(BA, X) = dist(A, X)`
    * any distances to BA from another node being the distance from the other node to B: 
`dist(X, BA) = dist(X, B)`
* Having an algorithm _ALG_ that correctly computes the traveling salesman for the points 
{BA, C, D, ...}, we can find a solution for our problem by splitting BA in the solution
    * e.g. if a round trip BA → C → D → BA is a shortest roundtrip, the actual route is A → C → D → B → A

**Explanation**

Any solution to our problem starts at A and ends at B. As we look for a roundtrip, the trip from B to A always exist. Thus, the travel B → A is fixed.
![asdf](https://cloud.githubusercontent.com/assets/804027/20921170/37d3c420-bba3-11e6-89fd-315b8d71e98e.jpg)
Thus, it only has to be decided in which order all the other nodes have to be visited. And algorithm that minimizes the distance of the route visiting all nodes and also going from A to other nodes and 
[...]

PR #3408, opened 2016-12-06
Trip with Fixed Start and End points (TFSE)

ghoshkaj:
# Issue

https://github.com/Project-OSRM/osrm-backend/issues/1623

## Tasklist
 - [x] Work with Chau to figure out algorithm for roundtrip with fixed start and end points 
(We're going to implement @chaupow's algorithm described [here](https://github.com/Project-OSRM/osrm-backend/issues/1623#issuecomment-26510917))
 - [x] Figure out API design 
(@chaupow and @TheMarex and I have settled on: 
`coordinates`: list of coordinates
`source`: the index of the start coordinate
`destination`: the index of the end coordinate )
- [x] Write failing unit tests for new parameters
- [x] Make unit tests pass
- [x] Write failing cuke tests
- [x] Write code
- [ ] Review
- [ ] Adjust for comments

 - [x] Relevant docs updated
 - [x] Changelog entry if required
 - [ ] Impact on the API surface

## Requirements / Relations

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -30,6 +30,8 @@
       - libOSRM now creates an own watcher thread then used in shared memory mode to listen for data updates
     - Tools:
       - Added osrm-extract-conditionals tool for checking conditional values in OSM data
+    - Trip Plugin
+      - Added a new feature that finds the optimal route given a list of waypoints, a source and a destination. This does not return a roundtrip and instead returns a one way optimal route from the fixed source to the destination points.
 
 # 5.5.1
   - Changes from 5.5.0
--- a/docs/http.md
+++ b/docs/http.md
@@ -175,7 +175,7 @@ In addition to the [general options](#general-options) the following options are
 |annotations |`true`, `false` (default)                    |Returns additional metadata for each coordinate along the route geometry.      |
 |geometries  |`polyline` (default), `polyline6`, `geojson` |Returned route geometry format (influences overview and per step)              |
 |overview    |`simplified` (default), `full`, `false`      |Add overview geometry either full, simplified according to highest zoom level it could be display on, or not at all.|
-|continue\_straight |`default` (default), `true`, `false`   |Forces the route to keep going straight at waypoints constraining uturns there even if it would be faster. Default value depends on the profile. |
+|continue\_straight |`default` (default), `true`, `false` |Forces the route to keep going straight at waypoints constraining uturns there even if it would be faster. Default value depends on the profile. |
 
 \* Please note that even if an alternative route is requested, a result cannot be guaranteed.
 
@@ -311,27 +311,62 @@ All other properties might be undefined.
 
 ### Trip service
 
-The trip plugin solves the Traveling Salesman Problem using a greedy heuristic (farthest-insertion algorithm).
-The returned path does not have to be the fastest path, as TSP is NP-hard it is only an approximation.
-Note that if the input coordinates can not be joined by a single trip (e.g. the coordinates are on several disconnected islands)
-multiple trips for each connected component are returned.
+The trip plugin solves the Traveling Salesman Problem using a greedy heuristic (farthest-insertion algorithm) for 10 or more waypoints and uses brute force for less than 10 waypoints.
+The returned path does not have to be the fastest path. As TSP is NP-hard it only returns an approximation.
+Note that all input coordinates have to be connected for the trip service to work. 
 
 ```endpoint
-GET /trip/v1/{profile}/{coordinates}?steps={true|false}&geometries={polyline|polyline6|geojson}&overview={simplified|full|false}&annotations={true|false}'
+GET /trip/v1/{profile}/{coordinates}?roundtrip={true|false}&source{any|first}&destination{any|last}&steps={true|false}&geometries={polyline|polyline6|geojson}&overview={simplified|full|false}&annotations={true|false}'
 ```
 
 In addition to the [general options](#general-options) the following options are supported for this service:
 
 |Option      |Values                                          |Description                                                                |
 |------------|------------------------------------------------|---------------------------------------------------------------------------|
+|roundtrip   |`true` (default), `false`                       |Return route is a roundtrip                                                |
+|source      |`any` (default), `first`                        |Return route starts at `any` or `first` coordinate                         |
+|destination |`any` (default), `last`                         |Return route ends at `any` or `last` coordinate                            |
 |steps       |`true`, `false` (default)                       |Return route instructions for each trip                                    |
 |annotations |`true`, `false` (default)                       |Returns additional metadata for each coordinate along the route geometry.  |
 |geometries  |`polyline` (default), `polyline6`, `geojson`    |Returned route geometry format (influences overview and per step)          |
 |overview    |`simplified` (default), `full`, `false`         |Add overview geometry either full, simplified according to highest zoom level it could be display on, or not at all.|
 
-**Response**
+**Fixing Start and End Points**
 
-- `code` if the request was successful `Ok` otherwise see the service dependent and general status codes.
+It is possible to explicitely set the start or end coordinate of the trip. 
+When source is set to `first`, the first coordinate is used as start coordinate of the trip in the output. When destination is set to `last`, the last coordinate will be used as destination of the trip in the returned output. If you specify `any`, any of the coordinates can be used as the first or last coordinate in the output.
+
+However, if `source=any&destination=any` the returned round-trip will still start at the first input coordinate by default.
+
+Currently, not all combinations of `roundtrip`, `source` and `destination` are supported.
+Right now, the following combinations are possible:
+
+| roundtrip | source | destination | supported |
+| :-- | :-- | :-- | :-- |
+| true | first | last | **yes** | 
+| true | first | any | **yes** |
+| true | any | last | **yes** |
+| true | any | any | **yes** |
+| false | first | last | **yes** |
+| false | first | any | no |
+| false | any | last | no |
+| false | any | any | no |
+
+#### Example Requests
+
+```curl
+# Round trip in Berlin with three stops:
+curl 'http://router.project-osrm.org/trip/v1/driving/13.388860,52.517037;13.397634,52.529407;13.428555,52.523219'
+```
+
+```curl
+# Round trip in Berlin with four stops, starting at the first stop, ending at the last:
+curl 'http://router.project-osrm.org/trip/v1/driving/13.388860,52.517037;13.397634,52.529407;13.428555,52.523219;13.418555,52.523215?source=first&destination=last'
+```
+
+#### Response
+
+- `code`: if the request was successful `Ok` otherwise see the service dependent and general status codes.
 - `waypoints`: Array of `Waypoint` objects representing all waypoints in input order. Each `Waypoint` object has the following additional properties:
   - `trips_index`: Index to `trips` of the sub-trip the point was matched to.
   - `waypoint_index`: Index of the point in the trip.
@@ -341,7 +376,8 @@ In case of error the following `code`s are supported in addition to the general 
 
 | Type              | Description         |
 |-------------------|---------------------|
-| `NoTrips`         | No trips found.     |
+| `NoTrips`         | No trips found because input coordinates are not connected.|
+| `NotImplemented`  | This request is not supported |
 
 All other properties might be undefined.
 
--- a/features/step_definitions/trip.js
+++ b/features/step_definitions/trip.js
@@ -24,7 +24,6 @@ module.exports = function () {
                             } else if (row[k]) {
                                 params[match[1]] = [row[k]];
                             }
-
                             got[k] = row[k];
                         }
                     }
@@ -35,11 +34,11 @@ module.exports = function () {
                     }
 
                     if (headers.has('status')) {
-                        got.status = json.status.toString();
+                        got.status = json.code;
                     }
 
                     if (headers.has('message')) {
-                        got.message = json.status_message;
+                        got.message = json.message;
                     }
 
                     if (headers.has('geometry')) {
@@ -59,6 +58,7 @@ module.exports = function () {
 
                     var subTrips;
                     var trip_durations;
+                    var trip_distance;
                     if (res.statusCode === 200) {
                         if (headers.has('trips')) {
                             subTrips = json.trips.filter(t => !!t).map(t => t.legs).map(tl => Array.prototype.concat.apply([], tl.map((sl, i) => {
@@ -74,13 +74,19 @@ module.exports = function () {
                             })));
                             trip_durations = all_durations.map( a => a.reduce(add, 0));
                         }
+                        if(headers.has('distance')) {
+                            var all_distance = json.trips.filter(t => !!t).map(t => t.legs).map(tl => Array.prototype.concat.apply([], tl.map(sl => {
+                                return sl.distance;
+                            })));
+                            trip_distance = all_distance.map( a => a.reduce(add, 0));
+                        }
                     }
 
                     var ok = true,
                         encodedResult = '',
                         extendedTarget = '';
 
-                    row.trips.split(',').forEach((sub, si) => {
+                    if (json.trips) row.trips.split(',').forEach((sub, si) => {
                         if (si >= subTrips.length) {
                             ok = false;
                         } else {
@@ -105,10 +111,10 @@ module.exports = function () {
                         got.via_points = row.via_points;
                     } else {
                         got.trips = encodedResult;
-                        got.trips = extendedTarget;
                     }
 
                     got.durations = trip_durations;
+                    got.distance = trip_distance;
 
                     for (var key in row) {
                         if (this.FuzzyMatch.match(got[key], row[key])) {
@@ -144,6 +150,19 @@ module.exports = function () {
                             waypoints.push(node);
                         });
                         got = { waypoints: row.waypoints };
+
+                        if (row.source) {
+                            params.source = got.source = row.source;
+                        } 
+                        
+                        if (row.destination) {
+                            params.destination = got.destination = row.destination;
+                        } 
+
+                        if (row.hasOwnProperty('roundtrip')) { //roundtrip is a boolean so row.roundtrip alone doesn't work as a check here
+                            params.roundtrip = got.roundtrip = row.roundtrip;
+                        }
+
                         this.requestTrip(waypoints, params, afterRequest);
                     } else {
                         throw new Error('*** no waypoints');
--- a/features/testbot/trip.feature
+++ b/features/testbot/trip.feature
@@ -5,7 +5,25 @@ Feature: Basic trip planning
         Given the profile "testbot"
         Given a grid size of 10 meters
 
-    Scenario: Testbot - Trip planning with less than 10 nodes
+    Scenario: Testbot - Trip: Roundtrip with one waypoint
+        Given the node map
+            """
+            a b
+            c d
+            """
+
+        And the ways
+            | nodes |
+            | ab    |
+            | bc    |
+            | cb    |
+            | da    |
+
+        When I plan a trip I should get
+            | waypoints | trips  |
+            | a         | aa     |
+
+    Scenario: Testbot - Trip: Roundtrip with waypoints (less than 10)
         Given the node map
             """
             a b
@@ -24,7 +42,7 @@ Feature: Basic trip planning
             | a,b,c,d   | abcda  | 7.6       |
             | d,b,c,a   | dbcad  | 7.6       |
 
-    Scenario: Testbot - Trip planning with more than 10 nodes
+    Scenario: Testbot - Trip: Roundtrip waypoints (more than 10)
         Given the node map
             """
             a b c d
@@ -37,7 +55,6 @@ Feature: Basic trip planning
             | nodes |
             | ab    |
             | bc    |
-            | cb    |
             | de    |
             | ef    |
             | fg    |
@@ -48,12 +65,85 @@ Feature: Basic trip planning
             | kl    |
             | la    |
 
+        When I plan a trip I should get
+            | waypoints               | trips         |
+            | a,b,c,d,e,f,g,h,i,j,k,l | alkjihgfedcba |
+
+    Scenario: Testbot - Trip: Roundtrip FS waypoints (more than 10)
+        Given the node map
+            """
+            a b c d
+            l     e
+            k     f
+            j i h g
+            """
+
+        And the ways
+            | nodes |
+            | ab    |
+            | bc    |
+            | de    |
+            | ef    |
+            | fg    |
+            | gh    |
+            | hi    |
+            | ij    |
+            | jk    |
+            | kl    |
+            | la    |
+
+        When I plan a trip I should get
+            | waypoints               | source | trips         |
+            | a,b,c,d,e,f,g,h,i,j,k,l | first  | alkjihgfedcba |
+
+    Scenario: Testbot - Trip: Roundtrip FE waypoints (more than 10)
+        Given the query options
+            | source | last  |
+        Given the node map
+            """
+            a b c d
+            l     e
+            k     f
+            j i h g
+            """
+
+        And the ways
+            | nodes |
+            | ab    |
+            | bc    |
+            | de    |
+            | ef    |
+            | fg    |
+            | gh    |
+            | hi    |
+            | ij    |
+            | jk    |
+            | kl    |
+            | la    |
 
         When I plan a trip I should get
             | waypoints               | trips         |
-            | a,b,c,d,e,f,g,h,i,j,k,l | cbalkjihgfedc |
+            | a,b,c,d,e,f,g,h,i,j,k,l | lkjihgfedcbal |
 
-    Scenario: Testbot - Trip planning with multiple scc
+    Scenario: Testbot - Trip: Unroutable roundtrip with waypoints (less than 10)
+        Given the node map
+            """
+            a b
+
+            d c
+            """
+
+         And the ways
+            | nodes |
+            | ab    |
+            | dc    |
+
+         When I plan a trip I should get
+            |  waypoints    | status         | message                                       |
+            |  a,b,c,d      | NoTrips        | No trip visiting all destinations possible.  |
+
+
+    Scenario: Testbot - Trip: Unroutable roundtrip with waypoints (more than 10)
         Given the node map
             """
             a b c d
@@ -62,7 +152,7 @@ Feature: Basic trip planning
             j i h g
 
             q m n
-              p o
+            p o
             """
 
         And the ways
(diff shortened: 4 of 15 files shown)


danpat:
Capturing from chat today:

The complexity seems quite high with the approach of creating the tfse_ table, performing the FA/brute search, then mapping the results back to the original node IDs.

An alternative could be just to munge certain values in the distance table.  If `A` is our start point, and `D` is our designated endpoint, we can do something like set all "X->A" cell values to `INT_MAX`, and all `D->X` values to `INT_MAX`.  This should be basically equivalent to the `tfse_table_`, but without needing to perform that step.

Of course, with any heuristic, it's possible that we'd calculate a result that includes an `INT_MAX` edge, and this could cause all kinds of integer overflow problems.  To make the result correct, you can update the brute force and FA algorithms to fast-fail a candidate route if it touches an `INT_MAX` edge - i.e. `INT_MAX` is treated as a special value that means "no connection".

@chaupow Thoughts on the above?  The creation of the `tfse_table_` makes sense, but it adds some complexity that I think we could avoid.  Can you see anything fundamentally wrong with the approach above?

chaupow:
> @chaupow Thoughts on the above? The creation of the tfse_table_ makes sense, but it adds some complexity that I think we could avoid. Can you see anything fundamentally wrong with the approach above?

If I'm not wrong, this should work as well and result in the same effect of forcing a round trip that visits A after D. Also, Julien from vroom does the same https://github.com/Project-OSRM/osrm-backend/issues/1623#issuecomment-267534420

I'm 👍

ghoshkaj:
This is the table that @danpat and I came up with from our conversation. Putting it here for the record as I work through the cucumber test scenarios that use this and so that everyone is on the same page about new table.

original table:
```  
  a  b  c  d  e
a 0  15 36 34 30 
b 15 0  25 30 34 
c 36 25 0  18 32 
d 34 30 18 0  15 
e 30 34 32 15 0 
```
manipulated table:  
```
  a      b      c      d      e
a 0      15     10000  34     30 
b 10000  0      25     30     34 
c 0      10000  10000  10000  10000 
d 10000  30     18     0      15 
e 10000  34     32     15     0
```

ghoshkaj:
@danpat @chaupow and anyone else on directions, review please?

src/engine/plugins/trip.cpp
```diff
@@ -187,8 +204,90 @@ Status TripPlugin::HandleRequest(const std::shared_ptr<const datafacade::BaseDat
...
+    std::vector<EdgeWeight> tfse_table_ = result_table.GetTable();
+
+    if (parameters.source > -1 && parameters.destination > -1)
+    {
+        for (std::size_t f_counter = 0; f_counter < tfse_table_.size(); f_counter++)
+        {
+            // parameters.source column
+            if (f_counter % result_table.GetNumberOfNodes() == (unsigned long)parameters.source)
+            {
+                tfse_table_[f_counter] = std::numeric_limits<EdgeWeight>::max();
```
danpat: Given that the FI and BruteForce searches are looking for `INVALID_EDGE_WEIGHT`, that's what you should set here, not `::max()`.  Practically, they're the same thing, but if the `INvALID_EDGE_WEIGHT` value changes for some reason, the `if()` logic later on will break.

src/engine/plugins/trip.cpp
```diff
@@ -187,8 +204,90 @@ Status TripPlugin::HandleRequest(const std::shared_ptr<const datafacade::BaseDat
...
+            }
+
+            // parameters.destination row
+            if (f_counter >= (result_table.GetNumberOfNodes() * parameters.destination) &&
+                f_counter < result_table.GetNumberOfNodes() * parameters.destination +
+                                result_table.GetNumberOfNodes())
+            {
+                tfse_table_[f_counter] = std::numeric_limits<EdgeWeight>::max();
```
danpat: Same here.

src/engine/plugins/trip.cpp
```diff
@@ -187,8 +204,90 @@ Status TripPlugin::HandleRequest(const std::shared_ptr<const datafacade::BaseDat
...
+            }
+        }
+
+        tfse_table_[parameters.source * result_table.GetNumberOfNodes() + parameters.source] = 0;
+        tfse_table_[parameters.destination * result_table.GetNumberOfNodes() + parameters.source] =
+            0;
+        tfse_table_[parameters.destination * result_table.GetNumberOfNodes() +
+                    parameters.destination] = 0;
+        tfse_table_[parameters.source * result_table.GetNumberOfNodes() + parameters.destination] =
+            std::numeric_limits<EdgeWeight>::max();
```
danpat: And here.

danpat:
@ghoshkaj For routes >10 waypoints, we will need to make the same logic change to the FarthestInsertion algorithm that got made to BruteForce.  A test for that scenario would be great too.

include/engine/api/base_parameters.hpp
```diff
@@ -96,6 +96,10 @@ struct BaseParameters
                            });
     }
 };
+inline std::ostream &operator<<(std::ostream &out, const BaseParameters &)
+{
+    return out << "BaseParameters()";
```
danpat: Debugging code?

CHANGELOG.md
```diff
@@ -27,6 +27,8 @@
       - libOSRM now creates an own watcher thread then used in shared memory mode to listen for data updates
     - Tools:
       - Added osrm-extract-conditionals tool for checking conditional values in OSM data
+    - Trip Plugin
+      - Added a new feature that finds the optimal route given a list of location, a start and destination locations. This does not return a roundtrip, but instead gives the optimal route starting at the start and destination locations.
```
danpat: "... optimal route starting at the start and destination locations" - > "... optimal route given fixed start and end points"

docs/http.md
```diff
@@ -329,6 +330,30 @@ In addition to the [general options](#general-options) the following options are
 |geometries  |`polyline` (default), `polyline6`, `geojson`    |Returned route geometry format (influences overview and per step)          |
 |overview    |`simplified` (default), `full`, `false`         |Add overview geometry either full, simplified according to highest zoom level it could be display on, or not at all.|
 
+A second feature of the trip plugin returns the optimal route from a source to a destination while visiting all the other input locations in the middle.
```
danpat: "input locations" -> "waypoints"

docs/http.md
```diff
@@ -329,6 +330,30 @@ In addition to the [general options](#general-options) the following options are
 |geometries  |`polyline` (default), `polyline6`, `geojson`    |Returned route geometry format (influences overview and per step)          |
 |overview    |`simplified` (default), `full`, `false`         |Add overview geometry either full, simplified according to highest zoom level it could be display on, or not at all.|
 
+A second feature of the trip plugin returns the optimal route from a source to a destination while visiting all the other input locations in the middle. 
+As with the roundtrip service, for locations 10 nodes of less, the brute force algorithm is used, while for more than 10 nodes, the farthest insertion heuristic is used.
```
danpat: "10 nodes or less", not "of"

docs/http.md
```diff
@@ -329,6 +330,30 @@ In addition to the [general options](#general-options) the following options are
...
+
+**Example:**
+
+```
+sources=0&destinations=5
+```
+
+|Element     |Values                       |
+|------------|-----------------------------|
+|index       |`0 <= integer < #locations`  |
```
danpat: either `coordinates` or `waypoints`, not `locations`

features/testbot/trip.feature
```diff
@@ -48,12 +47,130 @@ Feature: Basic trip planning
...
+            | bc    |
+
+        When I plan a trip I should get
+            |  waypoints  | status        | message                                                                                  |
+            |  a,b,c      | InvalidInputs | Source or destination indices are greater number of coordinates provided. |
+
+
+    Scenario: Testbot - Trip planning with illegal source
+        Given the query options
+            | source        | 10     |
```
danpat: `source` and `destination` should be columns in the "When I plan a trip I should get" columns, not a separate "Given", allowing per-query adjustment.  This will allow some consolidation of these test cases.

docs/http.md
```diff
@@ -311,7 +311,7 @@ All other properties might be undefined.
 
 ### Trip service
 
-The trip plugin solves the Traveling Salesman Problem using a greedy heuristic (farthest-insertion algorithm).
+The trip plugin solves the Traveling Salesman Problem using a greedy heuristic (farthest-insertion algorithm) for more than ten nodes and uses brute force for 10 nodes or less.
```
daniel-j-h: That's not true: < 10 is brute force

https://github.com/Project-OSRM/osrm-backend/blob/b3ef27d104787af8c82ee11d51769a2164bce9b2/src/engine/plugins/trip.cpp#L209

docs/http.md
```diff
@@ -329,6 +329,30 @@ In addition to the [general options](#general-options) the following options are
 |geometries  |`polyline` (default), `polyline6`, `geojson`    |Returned route geometry format (influences overview and per step)          |
 |overview    |`simplified` (default), `full`, `false`         |Add overview geometry either full, simplified according to highest zoom level it could be display on, or not at all.|
 
+A second feature of the trip plugin returns the route from a source to a destination while visiting all the other waypoints in the middle. 
+As with the roundtrip service, for fewer than 10 waypoints, the brute force algorithm is used, while for more than 10 waypoints the farthest insertion heuristic is used.
```
daniel-j-h: Same here:
- < 10 brute force
- \>= 10 farthest insertion

docs/http.md
```diff
@@ -329,6 +329,30 @@ In addition to the [general options](#general-options) the following options are
...
+
+|Option      |Values                                          |Description                                                                                                                              |
+|------------|------------------------------------------------|-----------------------------------------------------------------------------------------------------------------------------------------|
+|source      |`{index of coordinate in coordinates provided}` |Will error if destionation is not provided as well. Returns trip from source to destination routing through all other locations provided.|
```
daniel-j-h: destination - grammar

features/step_definitions/trip.js
```diff
@@ -35,11 +34,12 @@ module.exports = function () {
                     }
 
                     if (headers.has('status')) {
-                        got.status = json.status.toString();
+                        got.status = json.code;
```
daniel-j-h: Why the change from the status message to the error code here?

features/testbot/trip.feature
```diff
@@ -48,12 +47,130 @@ Feature: Basic trip planning
...
+            """
+
+        And the ways
+            | nodes |
+            | ab    |
+            | bc    |
+
+        When I plan a trip I should get
+            |  waypoints  | status        | message                                                                                  |
+            |  a,b,c      | InvalidInputs | Source or destination indices are greater number of coordinates provided. |
```
daniel-j-h: All these tests should be unit tests - check the unit tests dir and the library tests there, e.g. for route.

include/engine/api/trip_parameters.hpp
```diff
@@ -47,7 +47,27 @@ namespace api
  */
 struct TripParameters : public RouteParameters
 {
-    // bool IsValid() const; Falls back to base class
```
daniel-j-h: remove

include/engine/api/trip_parameters.hpp
```diff
@@ -47,7 +47,27 @@ namespace api
...
+    TripParameters()
+        : RouteParameters(false,
+                          false,
+                          false,
+                          RouteParameters::GeometriesType::Polyline,
+                          RouteParameters::OverviewType::Simplified,
+                          {}),
```
daniel-j-h: Why setting parameters here for RouteParameters?

Can't this be `TripParameters() = default;` and setting `source = -1` and `destination = -1` below?

include/engine/api/trip_parameters.hpp
```diff
@@ -47,7 +47,27 @@ namespace api
...
+    }
+
+    template <typename... Args>
+    TripParameters(int source_, int destination_, Args &&... args_)
+        : RouteParameters{std::forward<Args>(args_)...}, source{source_}, destination{destination_}
+    {
+    }
+
+    int source;
+    int destination;
```
daniel-j-h: Source and destination are indices - they should really be `std::size_t` and maybe even wrapper in an `optional<T>`

include/engine/api/trip_parameters.hpp
```diff
@@ -47,7 +47,27 @@ namespace api
...
+
+    bool IsValid() const { return RouteParameters::IsValid() && source >= -1 && destination >= -1; }
```
daniel-j-h: You check >= -1 but what about < coordinates size?

include/engine/trip/trip_brute_force.hpp
```diff
@@ -31,7 +31,20 @@ EdgeWeight ReturnDistance(const util::DistTableWrapper<EdgeWeight> &dist_table,
...
+        // Fixed start and end trip needs to escape this path if a max edge weight is found on the
+        // dist table
+        if (dist_table(location_order[i], location_order[(i + 1) % component_size]) ==
+            INVALID_EDGE_WEIGHT)
+        {
+            return INVALID_EDGE_WEIGHT;
+        }
+        else
+        {
+            route_dist += dist_table(location_order[i], location_order[(i + 1) % component_size]);
```
daniel-j-h: Call dist_table once outside - then store result and make the check

include/engine/trip/trip_brute_force.hpp
```diff
@@ -31,7 +31,20 @@ EdgeWeight ReturnDistance(const util::DistTableWrapper<EdgeWeight> &dist_table,
     std::size_t i = 0;
     while (i < location_order.size() && (route_dist < min_route_dist))
     {
-        route_dist += dist_table(location_order[i], location_order[(i + 1) % component_size]);
+        // Check for path that touches INVALID_EDGE_WEIGHT (which is MAX_32INT)
+        // Fixed start and end trip needs to escape this path if a max edge weight is found on the
+        // dist table
```
daniel-j-h: Hum.. why?

include/engine/trip/trip_farthest_insertion.hpp
```diff
@@ -50,8 +50,8 @@ GetShortestRoundTrip(const NodeID new_loc,
         const auto dist_to = dist_table(new_loc, *to_node);
         const auto trip_dist = dist_from + dist_to - dist_table(*from_node, *to_node);
 
-        BOOST_ASSERT_MSG(dist_from != INVALID_EDGE_WEIGHT, "distance has invalid edge weight");
-        BOOST_ASSERT_MSG(dist_to != INVALID_EDGE_WEIGHT, "distance has invalid edge weight");
+        if (dist_from == INVALID_EDGE_WEIGHT || dist_to == INVALID_EDGE_WEIGHT)
+            continue;
```
daniel-j-h: Same here, what's going on here?

include/util/dist_table_wrapper.hpp
```diff
@@ -46,6 +46,30 @@ template <typename T> class DistTableWrapper
         return table_[index];
     }
 
+    void InvalidateRoute(NodeID from, NodeID to)
```
daniel-j-h: The wrapper holds `T`s so you probably should accept `T`s to index the vector.

src/engine/plugins/trip.cpp
```diff
@@ -160,6 +176,14 @@ Status TripPlugin::HandleRequest(const std::shared_ptr<const datafacade::BaseDat
         return Error("InvalidValue", "Invalid coordinate value.", json_result);
     }
 
+    if (parameters.source >= static_cast<int>(parameters.coordinates.size()) ||
+        parameters.destination >= static_cast<int>(parameters.coordinates.size()))
+    {
+        return Error("InvalidInputs",
```
daniel-j-h: Not a documented v5 error code.

src/engine/plugins/trip.cpp
```diff
@@ -187,8 +211,76 @@ Status TripPlugin::HandleRequest(const std::shared_ptr<const datafacade::BaseDat
...
+    // d 34 30 18 0  15
+    // e 30 34 32 15 0
+
+    // New Table
+    //   a        b         c        d         e
+    // a 0        15        10000    34        30
+    // b 10000    0         25       30        34
+    // c 0        10000     0        10000     10000
+    // d 10000    30        18       0         15
+    // e 10000    34        32       15        0
```
daniel-j-h: So what does it do?

src/engine/plugins/trip.cpp
```diff
@@ -187,8 +211,76 @@ Status TripPlugin::HandleRequest(const std::shared_ptr<const datafacade::BaseDat
...
+
+    if (parameters.source > -1 && parameters.destination > -1)
+    {
+        // parameters.source column
+        for (NodeID i = 0; i < result_table.GetNumberOfNodes(); i++) {
+            if (i == (NodeID)parameters.source) continue;
```
daniel-j-h: format all files you're touching

unit_tests/server/parameters_parser.cpp
```diff
@@ -439,9 +439,19 @@ BOOST_AUTO_TEST_CASE(valid_trip_urls)
...
+    TripParameters reference_2{};
+    reference_2.coordinates = coords_1;
+    reference_2.source = 1;
+    reference_2.destination = 2;
+    auto result_2 = parseParameters<TripParameters>("1,2;3,4?source=1&destination=2");
+    BOOST_CHECK(result_2);
+    BOOST_CHECK_EQUAL(reference_2.source, result_2->source);
+    BOOST_CHECK_EQUAL(reference_2.destination, result_2->destination);
+    CHECK_EQUAL_RANGE(reference_2.radiuses, result_2->radiuses);
+    CHECK_EQUAL_RANGE(reference_2.coordinates, result_2->coordinates);
```
daniel-j-h: More tests please. For parsing but also library tests for responses.

include/util/dist_table_wrapper.hpp
```diff
@@ -46,6 +46,30 @@ template <typename T> class DistTableWrapper
...
+        BOOST_ASSERT_MSG(to < number_of_nodes_, "to ID is out of bound");
+
+        const auto index = from * number_of_nodes_ + to;
+
+        BOOST_ASSERT_MSG(index < table_.size(), "index is out of bound");
+
+        table_[index] = INVALID_EDGE_WEIGHT;
+    }
+
+    void ShortcutRoute(NodeID from, NodeID to)
```
chaupow: I'm not totally convinced that `ShortcutRoute ` and `InvalidateRoute ` are good names. If you want to keep them more general as they are now, can you add a comment why we do this here? so something like
`force the roundtrip to start at source and end at destination by virtually squashing them together`

or you can maybe name them `squashSourceAndDestination` and `separateDestination`

src/engine/plugins/trip.cpp
```diff
@@ -187,8 +211,76 @@ Status TripPlugin::HandleRequest(const std::shared_ptr<const datafacade::BaseDat
...
+
+        result_table.ShortcutRoute(parameters.destination, parameters.source);
+        result_table.InvalidateRoute(parameters.source, parameters.destination);
+    }
+
     // get scc components
-    SCC_Component scc = SplitUnaccessibleLocations(number_of_locations, result_table);
+    SCC_Component scc = SplitUnaccessibleLocations(result_table.GetNumberOfNodes(), result_table);
+
+    if (parameters.source > -1 && parameters.destination > -1) // check if fixed start and end
```
chaupow: later you're checking this to set  `bool roundtrip`. you could also set the boolean already here and use it.

include/engine/api/trip_parameters.hpp
ghoshkaj: I think I _did_ remove this :)

features/step_definitions/trip.js
ghoshkaj: The status.toString() code was copy pasted from another cucumber test but wasn't really being used in the trip context (it was broken code essentially with no way to trigger it to check), so I changed it so that it does work.

include/util/dist_table_wrapper.hpp
ghoshkaj: The DistTableWrapper is a wrapper around a vector that holds <T>, in the trip case the <T> is an EdgeWeight. This method here simply accepts NodeIDs to access the correct position of the necessary EdgeWeight in the table vector of EdgeWeights. So keeping the function as-is should be fine in this case.

src/engine/plugins/trip.cpp
```diff
@@ -117,24 +117,40 @@ SCC_Component SplitUnaccessibleLocations(const std::size_t number_of_locations,
...
+
+    if (roundtrip)
+    {
+        BOOST_ASSERT(min_route.segment_end_coordinates.size() == trip.size());
+    }
+    else
+    {
+        // trip comes out to be something like 0 1 4 3 2, so the sizes don't
+        BOOST_ASSERT(min_route.segment_end_coordinates.size() == trip.size() - 1);
+    }
```
ghoshkaj: I did this, but it got a bit confusing for `&&` / `||` noobs like me to keep the logic straight. So I reverted back to the if statements for clarity and simplicity's sake.

daniel-j-h:
Tasks for `node-osrm` ticketed here: https://github.com/Project-OSRM/node-osrm/issues/294.

ghoshkaj:
To Do (@chaupow if you want to get started on any of these)
- Figure out what to do for scc.component > 1 and fix error handling accordingly
- Add `roundtrip` parameter parser test(s)
- Add the `source={fixed|any}` and `destination={fixed|any}` parameters
- Add parameter parser tests, unit tests and cucumber tests for the scenarios involving fixed start and fixed end parameters

features/testbot/trip.feature
```diff
@@ -48,12 +65,85 @@ Feature: Basic trip planning
...
+            | fg    |
+            | gh    |
+            | hi    |
+            | ij    |
+            | jk    |
+            | kl    |
+            | la    |
+
+        When I plan a trip I should get
+            | waypoints               | trips         | source |
```
TheMarex: Why is `source` a result column? This should be a query parameter like in https://github.com/Project-OSRM/osrm-backend/pull/3408/files#diff-19534b2b92b1d620a3a6390c0f0b378dR100 no?

features/testbot/trip.feature
```diff
@@ -85,13 +175,106 @@ Feature: Basic trip planning
...
+            """
+
+         And the ways
+            | nodes |
+            | ab    |
+            | dc    |
+
+         When I plan a trip I should get
+            |  waypoints    | source  | destination | roundtrip | status           | message                                       |
+            |  a,b,c,d      | first   | last        | false     | NoTrips          | No trip visiting all destinations possible.   |
```
TheMarex: Same here: `source` and `destination` should be query parameters not part of the result table.

features/testbot/trip.feature
```diff
@@ -85,13 +175,106 @@ Feature: Basic trip planning
...
+            | be    |
+            | cd    |
+            | ce    |
+            | de    |
 
         When I plan a trip I should get
-            | waypoints                       | trips               |
-            | a,b,c,d,e,f,g,h,i,j,k,l,m,n,o,p | defghijklabcd,mnopm |
+            |  waypoints  | source | destination |roundtrip | trips  | durations         | distance               |
+            |  a,b,d,e,c  | first  | last        | false    | abedc  | 8.200000000000001 | 81.6                   |
```
TheMarex: Same here.

features/testbot/trip.feature
```diff
@@ -85,13 +175,106 @@ Feature: Basic trip planning
...
+            | de    |
+            | ef    |
+            | fg    |
+            | gh    |
+            | hi    |
+            | ij    |
+            | jk    |
+
+        When I plan a trip I should get
+            |  waypoints              | source | destination | roundtrip |  trips       | durations  | distance  |
```
TheMarex: Same here.

include/engine/trip/trip_brute_force.hpp
```diff
@@ -25,62 +26,77 @@ namespace trip
...
 {
     EdgeWeight route_dist = 0;
     std::size_t i = 0;
     while (i < location_order.size() && (route_dist < min_route_dist))
     {
-        route_dist += dist_table(location_order[i], location_order[(i + 1) % component_size]);
-        BOOST_ASSERT_MSG(dist_table(location_order[i], location_order[(i + 1) % component_size]) !=
-                             INVALID_EDGE_WEIGHT,
-                         "invalid route found");
+
```
TheMarex: A better naming for `i` would could be `current_index` and for `next_index = (current_index + 1) % number_of_locations`.

include/engine/trip/trip_farthest_insertion.hpp
```diff
@@ -158,57 +153,26 @@ std::vector<NodeID> FarthestInsertionTrip(const NodeIDIterator &start,
...
-                    max_from = *x;
-                    max_to = *y;
-                }
-            }
-        }
-    }
+    // find the pair of location with the biggest distance and make the pair the initial start
+    // trip. Skipping over the very first element (0,0), we make sure not to end up with the
+    // same start/end in the special case where all entries are the same.
+    const auto index = std::distance(
```
TheMarex: `index` should be named more precisely: `weight_index`.

include/engine/trip/trip_farthest_insertion.hpp
```diff
@@ -158,57 +153,26 @@ std::vector<NodeID> FarthestInsertionTrip(const NodeIDIterator &start,
...
+        std::begin(dist_table), std::max_element(std::begin(dist_table) + 1, std::end(dist_table)));
+    NodeID max_from = index / number_of_locations;
```
TheMarex: Needs a comment for computing the row from the `weight_index`.

include/engine/trip/trip_farthest_insertion.hpp
```diff
@@ -158,57 +153,26 @@ std::vector<NodeID> FarthestInsertionTrip(const NodeIDIterator &start,
...
+    NodeID max_to = index % number_of_locations;
```
TheMarex: Needs a comment for computing the column from the `weight_index`.

include/util/dist_table_wrapper.hpp
TheMarex: Should these functions be part of the `DistTableWrapper`? This class seems very generic and the functions are only relevant in the context of the `trip` plugin.

Another way to do it would be to have two functions `invalidateRoute(DistanceTableWrapper &dist, NodeID from, NodeID to)` and `shortcutRoute(DistanceTableWrapper &dist, NodeID from, NodeID to)` in the trip plugin.

include/util/typedefs.hpp
```diff
@@ -60,6 +60,8 @@ using NameID = std::uint32_t;
 using EdgeWeight = std::int32_t;
 using TurnPenalty = std::int16_t; // turn penalty in 100ms units
 
+static const std::size_t INVALID_SIZE_T = std::numeric_limits<std::size_t>::max();
```
TheMarex: `INVALID_INDEX` would probably be a more fitting name, since that is what it is used to denote in the using code.

src/engine/plugins/trip.cpp
```diff
@@ -28,129 +27,151 @@ namespace engine
...
+        source_id = 0;
+    }
+    if (parameters.destination == api::TripParameters::DestinationType::Last)
+    {
+        BOOST_ASSERT(number_of_locations > 0);
+        destination_id = number_of_locations - 1;
+    }
+    bool fixed_start = (source_id == 0);
+    bool fixed_end = (destination_id == number_of_locations - 1);
+    bool fixed_start_and_end = fixed_start && fixed_end;
```
TheMarex: It might be less confusing to not pre-compute `fixed_start_and_end`, because in the code below it was not clear if this is somehow different then `fixed_start && fixed_end`.

src/engine/plugins/trip.cpp
```diff
@@ -187,56 +214,48 @@ Status TripPlugin::HandleRequest(const std::shared_ptr<const datafacade::BaseDat
...
-        return Error("NoTrips", "Cannot find trips", json_result);
+        trip = trip::FarthestInsertionTrip(number_of_locations, result_table);
     }
 
-    // compute all round trip routes
-    std::vector<InternalRouteResult> routes;
-    routes.reserve(trips.size());
-    for (const auto &trip : trips)
+    // rotate result such that roundtrip starts at node with index 0
+    if (!(fixed_end && !fixed_start_and_end))
```
TheMarex: Same as `!fixed_end || fixed_start_and_end` == `!fixed_end && fixed_start || !fixed_end && fixed_end` == `fixed_start && !fixed_end`

src/engine/plugins/trip.cpp
```diff
@@ -187,56 +214,48 @@ Status TripPlugin::HandleRequest(const std::shared_ptr<const datafacade::BaseDat
...
     {
-        routes.push_back(ComputeRoute(facade, snapped_phantoms, trip));
+        auto desired_start_index = std::find(std::begin(trip), std::end(trip), 0);
+        BOOST_ASSERT(desired_start_index != std::end(trip));
+        std::rotate(std::begin(trip), desired_start_index, std::end(trip));
     }
+    else if (fixed_end && !fixed_start_and_end && parameters.roundtrip)
```
TheMarex: Same as `fixed_end && !fixed_start && parameters.roundtrip`

features/testbot/trip.feature
danpat: @TheMarex I asked @ghoshkaj to move `source` and `dest` into the Results table - they're query parameters in the same way that `waypoints` are.

Having these in the `When` table allows multiple combinations for the same scenario - putting them in the `Given the query options` clause limits us to one set per scenario.

danpat: Note that there is precedence for this style in the `testbot/matching.feature` where we supply timestamps.

chaupow: @TheMarex @danpat: and now... - fight! 😄
kidding. I actually agree with @danpat but in most of these cases we only test one parameter combination so the argument is just semi-valid.

If you both don't mind and there are no further opinions on these, I'd leave it as it is just because it's less work.

include/util/typedefs.hpp
chaupow: HAHAHA. True. Thanks. invalid_size_t is wrong indeed

features/testbot/trip.feature
ghoshkaj: Leaving this as is due to comment above.

features/testbot/trip.feature
ghoshkaj: Leaving this as is due to comment above.

features/testbot/trip.feature
ghoshkaj: Leaving this as is due to comment above.

src/engine/plugins/trip.cpp
ghoshkaj: `!(fixed_end && !fixed_start_and_end` == `!fixed_end || fixed_start_and_end` == `!fixed_end || (fixed_start && fixed_end)` == `(!fixed_end || fixed_start) && (!fixed_end || fixed_end)` == `!fixed_end || fixed_start` == `fixed_start`

ghoshkaj: No `!fixed_end || fixed_start_and_end == !fixed_end || (fixed_start && fixed_end)` is wrong from my comment above. Ugh it's so late, I can't keep the logic straight :/ 

In any case, using `fixed_start && !fixed_end` from Patrick's comment makes a test fail :/

ghoshkaj: While using `fixed_start` lets them all pass.

oxidase: hm,  can it be simplified? There are only 16 2-variable boolean functions

ghoshkaj: Okay no matter how many times I work it out, I always get to `!fixed_end || fixed_start` and this lets the tests pass happily so this is what I'll keep it as.

ghoshkaj: Realistically, what this case is taking care of are the following scenarios:
`!fixed_end` OR
`fixed_start` OR
`fixed_start && fixed_end`
so I'll either write all these cases out or comment with this.

oxidase: `!fixed_end || fixed_start` is the same as `!fixed_end OR
fixed_start OR fixed_start && fixed_end`, because `fixed_start OR fixed_start && fixed_end` is equal to `fixed_start`

Issue #1000, opened 2014-04-30
OS X compile error: call to 'abs' is ambiguous


springmeyer:
This is happening on Travis's OS X machines. Full log is [here](https://travis-ci.org/DennisOSRM/node-osrm/jobs/24120890). Details on XCode for this machine is [here](https://gist.github.com/springmeyer/d5be545a84f832ec1938).

One odd thing is that `c++11` should not be enabled and `libc++` should not be being linked  because `-DOSXLIBSTD="libstdc++"` is being passed to cmake, but the configure output says:

```
-- Activating -std=c++11 flag for >= OS X 10.9
-- linking against libc++
```

Error is:

```
[ 22%] Building CXX object CMakeFiles/OSRM.dir/Algorithms/DouglasPeucker.cpp.o
/Users/travis/build/DennisOSRM/node-osrm/Project-OSRM/Algorithms/DouglasPeucker.cpp:102:37: error: 
      call to 'abs' is ambiguous
            const double distance = std::abs(temp_dist);
                                    ^~~~~~~~
/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/../lib/c++/v1/cmath:660:1: note: 
      candidate function
abs(float __x) _NOEXCEPT {return fabsf(__x);}
^
/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/../lib/c++/v1/cmath:664:1: note: 
      candidate function
abs(double __x) _NOEXCEPT {return fabs(__x);}
^
/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/../lib/c++/v1/cmath:668:1: note: 
      candidate function
abs(long double __x) _NOEXCEPT {return fabsl(__x);}
^
1 error generated.
make[2]: *** [CMakeFiles/OSRM.dir/Algorithms/DouglasPeucker.cpp.o] Error 1
```

entangledloops:
Headsup: the fabs solution works for clang, but doesn't work with at least Visual Studio 2013 (which will now report ambiguous call for fabs).

PR #3409, opened 2016-12-06
Add headers consistency check

oxidase:
# Issue

Added a simple check for headers consistency. It creates a set of cpp files that only include hpp files and `make check-headers` will show incorrect  header files. It will not work for files with duplicated names.

## Tasklist
 - [x] review
 - [x] adjust for comments

 - [ ] Relevant docs updated
 - [ ] Impact on the API surface

## Requirements / Relations

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -735,3 +735,21 @@ if (ENABLE_FUZZING)
 
   add_subdirectory(fuzz)
 endif ()
+
+
+## add headers sanity check target that includes all headers independently
+set(check_headers_dir "${PROJECT_BINARY_DIR}/check-headers")
+file(GLOB_RECURSE headers_to_check
+  ${PROJECT_BINARY_DIR}/*.hpp
+  ${PROJECT_SOURCE_DIR}/include/*.hpp)
+foreach(header ${headers_to_check})
+  get_filename_component(filename ${header} NAME_WE)
+  set(filename "${check_headers_dir}/${filename}.cpp")
+  if (NOT EXISTS ${filename})
+    file(WRITE ${filename} "#include \"${header}\"\n")
+  endif()
+  list(APPEND sources ${filename})
+endforeach()
+add_library(check-headers STATIC EXCLUDE_FROM_ALL ${sources})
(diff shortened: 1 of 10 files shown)


TheMarex:
@oxidase good idea. 👍 What about adding this to one of the Travis jobs, too?

oxidase:
Rebased and added the check to the `gcc-6-release-i686` travis job.

/cc @MoKob

CMakeLists.txt
```diff
@@ -735,3 +736,21 @@ if (ENABLE_FUZZING)
...
+    ${PROJECT_SOURCE_DIR}/include/*.hpp)
+  foreach(header ${headers_to_check})
+    get_filename_component(filename ${header} NAME_WE)
+    set(filename "${check_headers_dir}/${filename}.cpp")
+    if (NOT EXISTS ${filename})
+      file(WRITE ${filename} "#include \"${header}\"\n")
+    endif()
+    list(APPEND sources ${filename})
+  endforeach()
+  add_library(check-headers STATIC EXCLUDE_FROM_ALL ${sources})
```
TheMarex: This is excluded by default from the `ALL` target, why do we need an additional option?

oxidase: removed, makes no sense. Just wanted to double check that it will not be included in `all`

oxidase:
@TheMarex removed ENABLE_HEADERS_CHECK option and added a conditional check in .travis.yml

oxidase:
Updated PR:
- EXCLUDE_FROM_ALL is still used otherwise in OSX build `check-header` target is added to `all` by default
- Appveyor also builds `check-header` project as a part of msvc solution.
- for build fail https://travis-ci.org/Project-OSRM/osrm-backend/jobs/183724043 added `cstdlib` https://github.com/Project-OSRM/osrm-backend/pull/3409/commits/b1aced327923c14d848b4a83f64fdbe3d29df3e9#diff-4ff77436d81ab7e8c0c026734a2d79cdR6 (similar, but not the same issue https://github.com/Project-OSRM/osrm-backend/issues/1000)

Issue #3414, opened 2016-12-07
Support for conditional restriction and/or speed limits


oxidase:
To handle conditional tags like [restriction:conditional](http://www.openstreetmap.org/relation/5443569#map=18/37.34131/-121.88777) support for opening hours must be added into osrm-extract/osrm-contract

Tasks:
 - [x] qi [opening_hours grammar](http://wiki.openstreetmap.org/wiki/Key:opening_hours/specification) https://github.com/Project-OSRM/osrm-backend/pull/3415
 - [x]  data type discussion/implementation for time ranges
 - [x] grammar extension with semantic actions
 - [ ] added handling of restrictions https://github.com/Project-OSRM/osrm-backend/blob/master/src/extractor/restriction_parser.cpp#L58
 - [ ] added weights updates in osrm-contract
  

/cc @daniel-j-h

daniel-j-h:
To clarify here: we don't strictly have to implement this in osrm-backend. It could be a little executable reading a pbf file and generating two csv files for every 5 minutes: one file for speed updates, one file for turn penalty updates.

We then can use these csv files in combination with the Traffic feature. Check the format here:
https://github.com/Project-OSRM/osrm-backend/wiki/Traffic

daniel-j-h:
We also have to convert the local time in OSM tags to UTC. For this I used

http://efele.net/maps/tz/world/

    ogr2ogr -f CSV output.csv world/tz_us_mask.shp -lco GEOMETRY=AS_WKT

before. It's all quite hacky but should work .. I think internally we might have timezone handling already, maybe ping @danpat in chat.

danpat:
@oxidase This is great.  Let me sketch out how I think we could use this so that people could support conditional restrictions.

The basic approach would be:
  1. Generate "traffic data files" that set certain road speeds to 0 when conditions require it.
  2. Use the "traffic data" features of `osrm-contract` to load these simulated traffic files.

The limitations are:
  1. Data size - it's not feasible to re-contract the whole planet every 5 minutes to keep up with short-term closures.
  2. Time-dependent routing - we can only mark closures "now" - if you request a 45 minute route, we will consider the conditions "now" everywhere, even if part of the route may close during the trip.

`osrm-contract` lets us replace road segment weights by [loading CSV files](https://github.com/Project-OSRM/osrm-backend/wiki/Traffic).  We can load many files, and they're loaded in the order supplied on the command line - if conflicting data appears, it's last-one-in-wins.

[...]

PR #3415, opened 2016-12-07
Added opening_hours grammar

oxidase:
# Issue

#3414
To handle conditional tags like restriction:conditional support for opening hours must be added into osrm-extract and/or osrm-contract

## Tasklist
  - [x]  qi opening_hours grammar
  - [x] grammar extension with semantic actions
 ~~- [ ] added handling of restrictions https://github.com/Project-OSRM/osrm-backend/blob/master/src/extractor/restriction_parser.cpp#L58~~
 ~~- [ ] weights updates~~
 - [x] review
 - [x] adjust for comments

## Requirements / Relations
http://wiki.openstreetmap.org/wiki/Key:opening_hours/specification

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -625,6 +625,14 @@ if(BUILD_TOOLS)
   target_link_libraries(osrm-io-benchmark ${BOOST_BASE_LIBRARIES})
 
   install(TARGETS osrm-io-benchmark DESTINATION bin)
+
+  find_package(Shapefile) # package libshp-dev
+  if(Shapefile_FOUND)
+    add_executable(osrm-extract-conditionals src/tools/extract-conditionals.cpp $<TARGET_OBJECTS:UTIL>)
+    target_include_directories(osrm-extract-conditionals PRIVATE ${LIBSHAPEFILE_INCLUDE_DIR})
+    target_link_libraries(osrm-extract-conditionals ${OSMIUM_LIBRARIES} ${BOOST_BASE_LIBRARIES} ${Boost_PROGRAM_OPTIONS_LIBRARY} ${LIBSHAPEFILE_LIBRARY})
+    install(TARGETS osrm-extract-conditionals DESTINATION bin)
+  endif()
 endif()
 
 if (ENABLE_ASSERTIONS)
--- a/cmake/FindShapefile.cmake
+++ b/cmake/FindShapefile.cmake
@@ -1,0 +1,21 @@
+# - Try to find Shapefile C Library
+#   http://shapelib.maptools.org/
+#
+# Exports:
+#  Shapefile_FOUND
+#  LIBSHAPEFILE_INCLUDE_DIR
+#  LIBSHAPEFILE_LIBRARY
+# Hints:
+#  LIBSHAPEFILE_LIBRARY_DIR
+
+find_path(LIBSHAPEFILE_INCLUDE_DIR
+          shapefil.h)
+
+find_library(LIBSHAPEFILE_LIBRARY
+             NAMES shp
+             HINTS "${LIBSHAPEFILE_LIBRARY_DIR}")
+
+include(FindPackageHandleStandardArgs)
+find_package_handle_standard_args(Shapefile DEFAULT_MSG
+                                  LIBSHAPEFILE_LIBRARY LIBSHAPEFILE_INCLUDE_DIR)
+mark_as_advanced(LIBSHAPEFILE_INCLUDE_DIR LIBSHAPEFILE_LIBRARY)
--- a/include/util/conditional_restrictions.hpp
+++ b/include/util/conditional_restrictions.hpp
@@ -1,0 +1,109 @@
+#ifndef OSRM_CONDITIONAL_RESTRICTIONS_HPP
+#define OSRM_CONDITIONAL_RESTRICTIONS_HPP
+
+#include <boost/fusion/include/adapt_struct.hpp>
+#include <boost/spirit/include/phoenix.hpp>
+#include <boost/spirit/include/qi.hpp>
+
+namespace osrm
+{
+namespace util
+{
+
+// Helper functions for OSM conditional restrictions
+// http://wiki.openstreetmap.org/wiki/Conditional_restrictions
+// Consitional restrictions is a vector of ConditionalRestriction
+// with a restriction value and a condition string
+struct ConditionalRestriction
+{
+    std::string value;
+    std::string condition;
+};
+
+#ifndef NDEBUG
+// Debug output stream operators for use with BOOST_SPIRIT_DEBUG
+inline std::ostream &operator<<(std::ostream &stream, const ConditionalRestriction &restriction)
+{
+    return stream << restriction.value << "=" << restriction.condition;
+}
+#endif
+}
+}
+
+BOOST_FUSION_ADAPT_STRUCT(osrm::util::ConditionalRestriction,
+                          (std::string, value)(std::string, condition))
+
+namespace osrm
+{
+namespace util
+{
+namespace detail
+{
+
+namespace
+{
+namespace ph = boost::phoenix;
+namespace qi = boost::spirit::qi;
+}
+
+template <typename Iterator, typename Skipper = qi::blank_type>
+struct conditional_restrictions_grammar
+    : qi::grammar<Iterator, Skipper, std::vector<ConditionalRestriction>()>
+{
+    // http://wiki.openstreetmap.org/wiki/Conditional_restrictions
+    conditional_restrictions_grammar() : conditional_restrictions_grammar::base_type(restrictions)
+    {
+        using qi::_1;
+        using qi::_val;
+        using qi::lit;
+
+        // clang-format off
+
+        restrictions
+            = restriction % ';'
+            ;
+
+        restriction
+            = value >> '@' >> condition
+            ;
+
+        value
+            = +(qi::char_ - '@')
+            ;
+
+        condition
+            = *qi::blank
+            >> (lit('(') >> qi::as_string[qi::no_skip[*~lit(')')]][_val = _1] >> lit(')')
+                | qi::as_string[qi::no_skip[*~lit(';')]][_val = _1]
+               )
+            ;
+
+        // clang-format on
+
+        BOOST_SPIRIT_DEBUG_NODES((restrictions)(restriction)(value)(condition));
+    }
+
+    qi::rule<Iterator, Skipper, std::vector<ConditionalRestriction>()> restrictions;
+    qi::rule<Iterator, Skipper, ConditionalRestriction()> restriction;
+    qi::rule<Iterator, Skipper, std::string()> value, condition;
+};
+}
+
+inline std::vector<ConditionalRestriction> ParseConditionalRestrictions(const std::string &str)
+{
+    auto it(str.begin()), end(str.end());
+    const detail::conditional_restrictions_grammar<decltype(it)> static grammar;
+
+    std::vector<ConditionalRestriction> result;
+    bool ok = boost::spirit::qi::phrase_parse(it, end, grammar, boost::spirit::qi::blank, result);
+
+    if (!ok || it != end)
+        return std::vector<ConditionalRestriction>();
+
+    return result;
+}
+
+} // util
+} // osrm
+
+#endif // OSRM_CONDITIONAL_RESTRICTIONS_HPP
--- a/include/util/opening_hours.hpp
+++ b/include/util/opening_hours.hpp
@@ -374,15 +374,16 @@ struct opening_hours_grammar : qi::grammar<Iterator, Skipper, std::vector<Openin
             ;
 
         // Selectors
-        selector_sequence = (wide_range_selectors(_a) || small_range_selectors(_a))[_val = _a];
+        selector_sequence = (wide_range_selectors(_a) >> small_range_selectors(_a))[_val = _a];
 
         wide_range_selectors
-            = (monthday_selector(_r1)
-               || year_selector(_r1)
-               || week_selector(_r1) // TODO week_selector
-              ) >> -lit(':');
+            = (-monthday_selector(_r1)
+               >> -year_selector(_r1)
+               >> -week_selector(_r1) // TODO week_selector
+              ) >> -lit(':')
+            ;
 
-        small_range_selectors = (weekday_selector(_r1) >> (&~lit(',') | eoi)) || time_selector(_r1);
+        small_range_selectors = -(weekday_selector(_r1) >> (&~lit(',') | eoi)) >> -time_selector(_r1);
 
         // Time selector
         time_selector = (timespan % ',')[ph::bind(&OpeningHours::times, _r1) = _1];
@@ -458,7 +459,7 @@ struct opening_hours_grammar : qi::grammar<Iterator, Skipper, std::vector<Openin
         date_offset = (plus_or_minus >> wday) | day_offset;
 
         date_from
-            = ((-year[_a = _1] >> (month[_b = _1] || (daynum[_c = _1] >> (&~lit(':') | eoi))))
+            = ((-year[_a = _1] >> ((month[_b = _1] >> -daynum[_c = _1]) | daynum[_c = _1]))
                | variable_date)
             [_val = ph::construct<OpeningHours::Monthday>(_a, _b, _c)]
             ;
@@ -504,7 +505,10 @@ struct opening_hours_grammar : qi::grammar<Iterator, Skipper, std::vector<Openin
             ("Sa", 6)
             ;
 
-        daynum = uint2_p[_pass = bind([](unsigned x) { return 01 <= x && x <= 31; }, _1), _val = _1];
+        daynum
+            = uint2_p[_pass = bind([](unsigned x) { return 01 <= x && x <= 31; }, _1), _val = _1]
+            >> (&~lit(':') | eoi)
+            ;
 
         weeknum = uint2_p[_pass = bind([](unsigned x) { return 01 <= x && x <= 53; }, _1), _val = _1];
 
--- a/include/util/opening_hours.hpp
+++ b/include/util/opening_hours.hpp
@@ -1,0 +1,637 @@
+#ifndef OSRM_OPENING_HOURS_HPP
+#define OSRM_OPENING_HOURS_HPP
+
+#include <boost/date_time/gregorian/gregorian.hpp>
+#include <boost/spirit/include/phoenix.hpp>
+#include <boost/spirit/include/qi.hpp>
+
+#include <boost/io/ios_state.hpp>
+
+#include <algorithm>
+#include <cctype>
+#include <iomanip>
+#include <iterator>
+#include <limits>
+#include <string>
+
+namespace osrm
+{
+namespace util
+{
+
+// Helper classes for "opening hours" format http://wiki.openstreetmap.org/wiki/Key:opening_hours
+// Supported simplified features in CheckOpeningHours:
+// - Year/Month/Day ranges
+// - Weekday ranges
+// - Time ranges
+// Not supported:
+// - Week numbers
+// - Holidays, events, variables dates
+// - Day offsets and periodic ranges
+struct OpeningHours
+{
+    enum Modifier
+    {
+        unknown,
+        open,
+        closed,
+        off,
+        is24_7
+    };
+
+    struct Time
+    {
+        enum Event : unsigned char
+        {
+            invalid,
+            none,
+            dawn,
+            sunrise,
+            sunset,
+            dusk
+        };
+
+        Event event;
+        std::int32_t minutes;
+
+        Time() : event(invalid), minutes(0) {}
+        Time(Event event) : event(event), minutes(0) {}
+        Time(char hour, char min) : event(none), minutes(hour * 60 + min) {}
+        Time(Event event, bool positive, const Time &offset)
+            : event(event), minutes(positive ? offset.minutes : -offset.minutes)
+        {
+        }
+    };
+
+    struct TimeSpan
+    {
+        Time from, to;
+        TimeSpan() = default;
+        TimeSpan(const Time &from_, const Time &to_) : from(from_), to(to_)
+        {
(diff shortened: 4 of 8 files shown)


include/extractor/opening_hours.hpp
```diff
@@ -0,0 +1,461 @@
...
+
+    OpeningHours() : modifier(open) {}
+
+    std::vector<TimeSpan> times;
+    std::vector<WeekdayRange> weekdays;
+    std::vector<MonthdayRange> monthdays;
+    Modifier modifier;
+};
+
+std::ostream &operator<<(std::ostream &stream, const OpeningHours::Modifier value)
```
daniel-j-h: I think this has to be `inline`

src/tools/extract-conditionals.cpp
```diff
@@ -0,0 +1,279 @@
...
+#include <shapefil.h>
+
+#include <fstream>
+#include <iostream>
+#include <unordered_map>
+
+#include <cstdlib>
+#include <ctime>
+
+// better place in third_party/libosmium/include/osmium/tags/regex_filter.hpp
```
daniel-j-h: Maybe if you ping @joto and be nice about it he might include it in libosmium ;)

include/extractor/opening_hours.hpp
oxidase: sure, i will do it

oxidase:
Just pushed `osrm-extract-conditionals` tool that has two commands:
- `cond-dump` to save conditional restrictions in CSV format
- `cond-check` to check dumped conditional restrictions and produce another CSV file with filtered out inactive conditions. The command requires [tz_world shapefile](http://efele.net/maps/tz/world/)

Some statistics:
- california-latest.osm.pbf - 363 restrictions, `cond-dump` processing time 0m12.458s, 7 incorrect restriction values (`no_left_turn (Mo-Fr 07:00-09:00, 14;00-18:00)`, ...),   41 incorrect opening hours conditions (`(Mo-Fr, 07:00`, `Mo-Fr 07:00-09:00. 16:00-18:00`)
- germany-latest.osm.pbf - 15 restrictions, `cond-dump` processing time 0m50.262s, 4 unsupported conditions (`length > 10 m`, ..), 2 incorrect opening hours conditions (`Mo-Fr 7:00-13:30`)

On travis `osrm-extract-conditionals` is not yet tested. I will update .travis.yml to install `libshp-dev` package and  add some cucumber tests.

TheMarex:
Things todo here:

- [ ] Rebase (only conflict in the changelog)
- [ ] Evaluate if the number of restrictions is plausible or if the parser is too restrictive. @daniel-j-h is skeptical if these numbers make sense.

daniel-j-h:
For the record I was saying there are over 70k [`maxspeed:conditional`](http://taginfo.openstreetmap.org/keys/maxspeed%3Aconditional) in the data and a MVP should include those for "traffic updates". Sorry if this wasn't clear.

TheMarex:
> For the record I was saying there are over 70k maxspeed:conditional in the data and a MVP should include those for "traffic updates".

How are we going to ingest those? Remember the speed we derive from `maxspeed` in the profile can be influence by a number of factors. We would need to add this to the profiles to really get time-dependent `speed` values.

daniel-j-h:
Good point - in general (not only for conditional maxspeeds) we would have to pass them again through the profiles. Hm this is where it gets kinda ugly and should probably be integrated with extract.

oxidase:
Updated branch to use segment midpoints and `:forward:` and `:backward:` terms.
`[:<transportation mode>]` conditions and non-opening-hours conditions will require support in profiles,
so ignored for now.

@TheMarex the current implementation is a PoC to check OSM data and conditional restrictions. If `maxspeed` value must be mapped and the mapping is defined in profile than the `dump-speed` command must be moved into `osrm-extract` but  it will require a profile extension to handle multiple conditions like `80 @ (22:00-06:00); 120 @ (20:00-22:00); 80 @ (weight > 7); 70 @ "wet"`. I would skip it for now, as it would require quite big changes in `way_function`.

TheMarex:
@oxidase yes agreed this is not scope of this PR. I was more asking in terms of priority of integrating condition max speeds, as we have no way right now to use them.

@daniel-j-h can you review this? I think this is fine to merge and try an integration into our pipeline.

CMakeLists.txt
```diff
@@ -648,6 +648,14 @@ if(BUILD_TOOLS)
...
   install(TARGETS osrm-unlock-all DESTINATION bin)
   install(TARGETS osrm-springclean DESTINATION bin)
+
+  find_package(Shapefile)
+  if(Shapefile_FOUND)
+    add_executable(osrm-extract-conditionals src/tools/extract-conditionals.cpp $<TARGET_OBJECTS:UTIL>)
+    target_include_directories(osrm-extract-conditionals PRIVATE ${Shapefile_INCLUDE_DIR})
+    target_link_libraries(osrm-extract-conditionals ${OSMIUM_LIBRARIES} ${BOOST_BASE_LIBRARIES} ${Boost_PROGRAM_OPTIONS_LIBRARY} ${Shapefile_LIBRARY})
+    install(TARGETS osrm-extract-conditionals DESTINATION bin)
+  endif()
```
daniel-j-h: Want to add a not for where users can get this lib from, say for the most recent Ubuntu releases? We probably should have a Wiki section on all our tools and what is required to build them, how they can be used etc.

cmake/FindShapefile.cmake
```diff
@@ -0,0 +1,17 @@
...
+#   http://shapelib.maptools.org/
+#
+# Once done, this will define
+#
+#  Shapefile_FOUND - system has Shapefile
+#  Shapefile_INCLUDE_DIRS - the Shapefile include directories
+#  Shapefile_LIBRARIES - link these to use Shapefile
+
+find_path(Shapefile_INCLUDE_DIR NAMES shapefil.h)
+find_library(Shapefile_LIBRARY NAMES shp)
```
daniel-j-h: HINTS library dir here?

cmake/FindShapefile.cmake
```diff
@@ -0,0 +1,17 @@
...
+
+set(Shapefile_FOUND TRUE)
+if ("${Shapefile_INCLUDE_DIR}" STREQUAL "Shapefile_INCLUDE_DIR-NOTFOUND" OR
+    "${Shapefile_LIBRARY}" STREQUAL "Shapefile_LIBRARY-NOTFOUND")
+  set(Shapefile_FOUND FALSE)
+endif()
```
daniel-j-h: I think the canonical way to do this is to dispatch to CMake's FindPackageHandleStandardArgs module, e.g.

https://github.com/daniel-j-h/cku/blob/master/cmake/FindCapstone.cmake

include/util/conditional_restrictions.hpp
```diff
@@ -0,0 +1,118 @@
...
+{
+    std::string value;
+    std::string condition;
+};
+
+// TODO: remove debug
+// inline std::ostream &operator<<(std::ostream &stream, const ConditionalRestriction &restriction)
+// {
+//     return stream << restriction.value << "=" << restriction.condition;
+// }
```
daniel-j-h: Feel free to keep it - but I would uncomment it.

Also seeing that you adapt the struct, fusion should give you this I/O operator for free

http://www.boost.org/doc/libs/1_63_0/libs/fusion/doc/html/fusion/sequence/operator.html

include/util/conditional_restrictions.hpp
```diff
@@ -0,0 +1,118 @@
...
+namespace util
+{
+namespace detail
+{
+
+namespace
+{
+namespace ph = boost::phoenix;
+namespace qi = boost::spirit::qi;
+}
```
daniel-j-h: :+1: Good idea!

include/util/conditional_restrictions.hpp
```diff
@@ -0,0 +1,118 @@
...
+    bool ok = boost::spirit::qi::phrase_parse(it, end, grammar, boost::spirit::qi::blank, result);
+
+    // // TODO: remove debug
+    // if (!ok || it != end)
+    //     std::cout << "failed at " << std::string(it, end) << "\n";
+    // else
+    //     std::cout << "passed " << str << "\n";
+
+    // for (auto x : result)
+    //     std::cout << "    " << x << "\n";
```
daniel-j-h: Want to keep this debug code behind some ifdef, or remove it completely?

include/util/conditional_restrictions.hpp
```diff
@@ -0,0 +1,118 @@
+#ifndef CONDITIONAL_RESTRICTIONS_HPP
+#define CONDITIONAL_RESTRICTIONS_HPP
```
daniel-j-h: Can you prefix all header guards with OSRM_

include/util/opening_hours.hpp
```diff
@@ -0,0 +1,645 @@
...
+
+// Helper classes for "opening hours" format http://wiki.openstreetmap.org/wiki/Key:opening_hours
+// Supported simplified features in CheckOpeningHours:
+// - Year/Month/Day ranges
+// - Weekday ranges
+// - Time ranges
+// Not supported:
+// - Week numbers
+// - Holidays, events, variables dates
+// - Day offsets and periodic ranges
```
daniel-j-h: Good call on these features :+1:

include/util/opening_hours.hpp
```diff
@@ -0,0 +1,645 @@
...
+        {
+            if (to.minutes < from.minutes)
+                to.minutes += 24 * 60;
+        }
+
+        bool IsInRange(const struct tm &time, bool &use_curr_day, bool &use_next_day) const
+        {
+            // TODO: events are not handled
+            if (from.event != OpeningHours::Time::none || to.event != OpeningHours::Time::none)
+                return false;
```
daniel-j-h: I guess this is fine for now - tagging conditional attributes with `sunrise` etc. is the mappers fault :+1:

include/util/opening_hours.hpp
```diff
@@ -0,0 +1,645 @@
...
+        int weekdays, overnight_weekdays;
+        WeekdayRange() = default;
+        WeekdayRange(unsigned char from, unsigned char to)
+        {
+            // weekdays mask for [from, to], e.g [2, 5] -> 0111100, [5, 2] -> 1100111,
+            //  [3, 3] -> 0001000, [0,6] -> 1111111, [6,0] -> 1000001, [4, 3] -> 1111111
+            weekdays = (from <= to) ? ((1 << (to - from + 1)) - 1) << from
+                                    : ~(((1 << (from - to - 1)) - 1) << (to + 1));
+            weekdays &= 0x7f;
+            overnight_weekdays = (weekdays << 1) | (weekdays & 0x40 ? 1 : 0);
```
daniel-j-h: Ah, nice idea!

include/util/opening_hours.hpp
```diff
@@ -0,0 +1,645 @@
...
+                            weekdays.end(),
+                            [&time, use_curr_day, use_next_day](const auto &x) {
+                                return x.IsInRange(time, use_curr_day, use_next_day);
+                            })) &&
+               (monthdays.empty() ||
+                std::any_of(monthdays.begin(),
+                            monthdays.end(),
+                            [&time, use_curr_day, use_next_day](const auto &x) {
+                                return x.IsInRange(time, use_curr_day, use_next_day);
+                            }));
```
daniel-j-h: Hm want to simplify this a bit?

include/util/opening_hours.hpp
```diff
@@ -0,0 +1,645 @@
...
+            ("Mar", 3)
+            ("Apr", 4)
+            ("May", 5)
+            ("Jun", 6)
+            ("Jul", 7)
+            ("Aug", 8)
+            ("Sep", 9)
+            ("Oct", 10)
+            ("Nov", 11)
+            ("Dec", 12)
```
daniel-j-h: Re. month and weekday - I hope they're language agnostic, e.g. german mappers also use "Dec" and not "Dez". Reading the spec this should be the case.

include/util/opening_hours.hpp
```diff
@@ -0,0 +1,645 @@
...
+            return true;
+
+        if (opening_hours.IsInRange(time))
+        {
+            is_open = opening_hours.modifier == OpeningHours::open;
+        }
+    }
+
+    return is_open;
+}
```
daniel-j-h: This is nicely abstracted :+1:

src/tools/extract-conditionals.cpp
```diff
@@ -0,0 +1,991 @@
...
+
+template <> struct match_key<std::regex>
+{
+    bool operator()(const std::regex &rule_key, const char *tag_key)
+    {
+        return std::regex_match(tag_key, rule_key);
+    }
+}; // struct match_key<std::regex>
+}
+}
```
daniel-j-h: I think we can update libosmium in a independent pr, merge into master, rebase and then remove this.

https://github.com/osmcode/libosmium/releases

Check scripts/update_dependencies.

src/tools/extract-conditionals.cpp
```diff
@@ -0,0 +1,991 @@
...
+    po::store(po::command_line_parser(arguments).options(options).positional(pos).run(), vm);
+
+    if (vm.count("help"))
+    {
+        std::cout << boost::filesystem::path(executable).filename().string() << " " << command_name
+                  << " [<args>]\n\n"
+                  << options;
+        std::exit(EXIT_SUCCESS);
+    }
+    boost::program_options::notify(vm);
```
daniel-j-h: Can you explain to me the po store/count/notify interaction?

src/tools/extract-conditionals.cpp
```diff
@@ -0,0 +1,991 @@
...
+
+    // Prepare output stream
+    std::streambuf *buf = std::cout.rdbuf();
+    std::ofstream of;
+    if (!csv_filename.empty())
+    {
+        of.open(csv_filename, std::ios::binary);
+        buf = of.rdbuf();
+    }
+    std::ostream stream(buf);
```
daniel-j-h: Hm why the ofstream/streambuf/ostream dance?

src/tools/extract-conditionals.cpp
```diff
@@ -0,0 +1,991 @@
...
```
daniel-j-h: Same here

src/tools/extract-conditionals.cpp
```diff
@@ -0,0 +1,991 @@
...
+
+    // Lambda function that returns local time in the tzname time zone
+    // Thread safety: MT-Unsafe const:env
+    std::unordered_map<std::string, struct tm> local_time_memo;
+    auto get_local_time_in_tz = [utc_time, &local_time_memo](const char *tzname) {
+        auto it = local_time_memo.find(tzname);
+        if (it == local_time_memo.end())
+        {
+            struct tm timeinfo;
+            setenv("TZ", tzname, 1);
```
daniel-j-h: Why?

src/tools/extract-conditionals.cpp
```diff
@@ -0,0 +1,991 @@
...
+
+            // Get time zone name and emplace polygon and local time for the UTC input
+            const auto tzname = DBFReadStringAttribute(dbfhandle, shape, tzid);
+            local_times.emplace_back(local_time_t{polygon, get_local_time_in_tz(tzname)});
+
+            // std::cout << boost::geometry::dsv(boost::geometry::return_envelope<box_t>(polygon))
+            //           << " " << tzname << " " << asctime(&local_times.back().second);
+        }
+
+        SHPDestroyObject(object);
```
daniel-j-h: Hm want to bind this to the object's scope? (RAII)

src/tools/extract-conditionals.cpp
```diff
@@ -0,0 +1,991 @@
...
+    }
+
+    DBFClose(dbfhandle);
+    SHPClose(shphandle);
```
daniel-j-h: Same here

I think there is a Boost at-scope-exit with lambdas

src/tools/extract-conditionals.cpp
```diff
@@ -0,0 +1,991 @@
...
+        const auto &restriction = value.restriction;
+
+        // Get local time of the restriction
+        const auto &local_time = get_local_time(point_t{location.lon(), location.lat()});
+
+        // TODO: check restriction type [:<transportation mode>][:<direction>]
+        // http://wiki.openstreetmap.org/wiki/Conditional_restrictions#Tagging
+
+        // TODO: parsing will fail for combined conditions, e.g. Sa-Su AND weight>7
+        // http://wiki.openstreetmap.org/wiki/Conditional_restrictions#Combined_conditions:_AND
```
daniel-j-h: Also we don't yet handle the `except` tag, do we? Check the osrm restriction parser we have restriction exceptions such as except motorcar.

src/tools/extract-conditionals.cpp
```diff
@@ -0,0 +1,991 @@
...
+
+    // Parse program argument
+    auto arguments = ParseGlobalArguments(argc, argv);
+    BOOST_ASSERT(!arguments.empty());
+
+    std::unordered_map<std::string, int (*)(const char *, const std::vector<std::string> &)>
+        commands = {{"cond-dump", &RestrictionsDumpCommand},
+                    {"cond-check", &RestrictionsCheckCommand},
+                    {"speed-dump", &SpeedLimitsDumpCommand},
+                    {"speed-check", &SpeedLimitsCheckCommand}};
```
daniel-j-h: Beautiful registering the functionality with the users commands here :+1:

CMakeLists.txt
oxidase: added comment " # package libshp-dev"

cmake/FindShapefile.cmake
oxidase: adopted to Shp: package exports now Shapefile_FOUND, LIBSHAPEFILE_INCLUDE_DIR, LIBSHAPEFILE_LIBRARY and  uses hint LIBSHAPEFILE_LIBRARY_DIR

include/util/conditional_restrictions.hpp
oxidase: I will keep it as is  in 
```
#ifndef NDEBUG
// Debug output stream operators for use with BOOST_SPIRIT_DEBUG
...
#endif
```

It is only needed for `BOOST_SPIRIT_DEBUG` that is extremely helpful for complex cases

include/util/conditional_restrictions.hpp
oxidase: Here i will remove completely debug code

include/util/opening_hours.hpp
oxidase: Yeah, i was thinking about it: first it needs a location and altitude :boom: to compute :sun_with_face: and second it is mainly relevant to opening hours in some public parks like ":ghost: @ (dusk-dawn)", so it is almost irrelevant to traffic, but can be added later as a map injection in the constructor that maps key to the event local time

include/util/opening_hours.hpp
oxidase: I can add human-readable comments, because it uses logical shortcuts, otherwise it will be like
```
if (!bla)
{
   return false;
}
```

include/util/opening_hours.hpp
oxidase: idk, here http://wiki.openstreetmap.org/wiki/Key:opening_hours/specification#month only English abbreviations in grammar. It is not a technical problem to add German translations, but think it is better to keep the grammar accordingly to the spec.

src/tools/extract-conditionals.cpp
oxidase: removed

src/tools/extract-conditionals.cpp
oxidase: * `po::store`  stores parsed command line arguments into a variable map `vm`
* `vm.count("help")` returns number of keys in the variable map, so if --help or -h was in arguments `count` will return 1
* `po::notify` propagates data from the variable map `vm` to handlers in `po::options_description`, eg  if `--input arg`  was specified then "arg" will be handled by `boost::program_options::value<std::string>(&osm_filename)` that is basically a value copy, but handler can be more complicated than a copy.

Nice thing that options can come from arguments line and from configuration file like
```
variables_map vm;
store(parse_command_line(argc, argv, desc), vm);
store(parse_config_file("example.cfg", desc), vm);
notify(vm);
```

src/tools/extract-conditionals.cpp
oxidase: It is only to have possibility for piping text streams. I would like keep it, but don't know how to make it simpler without macros

src/tools/extract-conditionals.cpp
oxidase: This comes from POSIX interface  http://man7.org/linux/man-pages/man3/tzset.3.html "The tzset() function initializes the tzname variable from the TZ environment variable." 

Will look how to safely change `tzname` without TZ envar modification.

Hard solution: parse timezone /usr/share/zoneinfo/ but it will be just non-portable reimplementation

src/tools/extract-conditionals.cpp
oxidase: It would require profile options and filtering tags with given restrictions. I would keep it as is for now, because also profiles are need to compute correct speed values. Probably it can be merged into `osrm-extract` with tuen restrictions and speed value computations.

src/tools/extract-conditionals.cpp
oxidase: Just checked: tzinfo is read at https://sourceware.org/git/?p=glibc.git;a=blob;f=time/tzset.c;h=eb420699c42a49b4763eb1303e033ca448c4a1db;hb=HEAD#l441 but `tz` comes from https://sourceware.org/git/?p=glibc.git;a=blob;f=time/tzset.c;h=eb420699c42a49b4763eb1303e033ca448c4a1db;hb=HEAD#l407

Looks like the only way is to use `TZ` variable.

daniel-j-h: Wow, thanks for the explanation. A bit ugly but :+1:

daniel-j-h:
Ready to go from my side :+1:

PR #3417, opened 2016-12-07
Add assertion to geoprinter

chaupow:
# Issue

If someone tries to print geojson output with the geojson printer and there is no logger open, an assertion will now notify the developer

## Tasklist
 - [ ] review
 - [ ] adjust for comments

@MoKob

### DIFF
--- a/include/util/geojson_debug_logger.hpp
+++ b/include/util/geojson_debug_logger.hpp
@@ -63,11 +63,11 @@ class GeojsonLogger
         // make sure to syncronize logging output, our writing should be sequential
         std::lock_guard<std::mutex> guard(lock);
 
-        // if there is no logfile, we cannot write
+        // if there is no logfile, we cannot write (possible reason: the guard might be out of scope
+        // (e.g. if it is anonymous))
         if (!ofs.is_open() || (nullptr == policy))
         {
-            // this can only happend between two guards when concurrent writing occurs
-            return false;
+            throw util::exception("Trying to use the geojson printer without an open logger.");
         }
 
         // use our policy to convert the arguments into geojson, this can be done in parallel


include/util/geojson_debug_logger.hpp
```diff
@@ -63,12 +63,8 @@ class GeojsonLogger
...
         std::lock_guard<std::mutex> guard(lock);
 
-        // if there is no logfile, we cannot write
-        if (!ofs.is_open() || (nullptr == policy))
-        {
-            // this can only happend between two guards when concurrent writing occurs
-            return false;
-        }
+        // if there is no logfile, we cannot write (possible reason: the guard might be out of scope (e.g. if it is anonymous))
+        BOOST_ASSERT_MSG(!ofs.is_open() || (nullptr == policy), "Trying to use the geojson printer without an open logger.")
```
daniel-j-h: Why catching this only in debug mode? If this is a hard error we should throw.

chaupow: Are we actually planning to use the geojson printer for anything else but debugging? If it is purely a debugging tool we might not want to throw? Or should we throw?

MoKob: It is a debugging tool, but we can still use it in release mode (e.g. when processing bigger datasets). It should error, no matter what. It should never be in a final code.

chaupow: alright, I'll change it :)

include/util/geojson_debug_logger.hpp
```diff
@@ -63,11 +63,10 @@ class GeojsonLogger
...
         std::lock_guard<std::mutex> guard(lock);
 
-        // if there is no logfile, we cannot write
-        if (!ofs.is_open() || (nullptr == policy))
+        // if there is no logfile, we cannot write (possible reason: the guard might be out of scope (e.g. if it is anonymous))
+        if(!ofs.is_open() || (nullptr == policy))
         {
-            // this can only happend between two guards when concurrent writing occurs
-            return false;
+            throw util::exception("Trying to use the geojson printer without an open logger.");
```
daniel-j-h: :+1:

include/util/geojson_debug_logger.hpp
```diff
@@ -63,11 +63,10 @@ class GeojsonLogger
         // make sure to syncronize logging output, our writing should be sequential
         std::lock_guard<std::mutex> guard(lock);
 
-        // if there is no logfile, we cannot write
-        if (!ofs.is_open() || (nullptr == policy))
+        // if there is no logfile, we cannot write (possible reason: the guard might be out of scope (e.g. if it is anonymous))
```
daniel-j-h: don't forget to format

Issue #3225, opened 2016-11-02
Don't announce turns while on a ferry


MoKob:
<img width="855" alt="screen shot 2016-11-02 at 11 18 28" src="https://cloud.githubusercontent.com/assets/12932279/19924922/3e852cc0-a0ee-11e6-811c-e4762d789640.png">

Right now we may announce turns while on ferries. Since it's unlikely that someone can switch boats mid-ride, we should disable any turn announcements between ferry-segments.

PR #2705, opened 2016-07-26
Refine Roundabout Handling: Turbo-Roundabouts

daniel-j-h:
## Modified

I found some errors in the tests added here and corrected the `removeLanesFromRoundabout` feature. As long as we don't support these correctly, roundabouts will no longer be assigned lanes and, as a result, we don't emit strange instructions within the roundabouts.

## Original PR Item:

These kind of roundabouts came up during Lane Handling for roundabouts (https://github.com/Project-OSRM/osrm-backend/pull/2693).

They're called Turbo-roundabouts or Turbine-roundabouts and are very
popular e.g. in Germany and the UK.

Seems like our roundabout handler sometimes is getting confused.
- https://www.mapillary.com/app/?lat=48.78199533&lon=8.194456&z=17&lng=8.1936397&pKey=ayxfZQt3BkkBi11cJZ5o9Q&focus=map
- https://www.mapillary.com/app/?lat=50.18011193834988&lon=8.194456&z=17&lng=8.4751722501247&focus=map&pKey=YJy_ZngowTLsX6sc9yTq_g
- https://www.mapillary.com/app/?lat=47.5772012447751&lon=8.194456&z=16.873861849026316&lng=7.7974465839462255&focus=map

Trying to figure out why, and codifying some scenarios for cucumber.

PR #3420, opened 2016-12-07
Suppress turns emitted while on ferry

karenzshea:
# Issue

Fixes https://github.com/Project-OSRM/osrm-backend/issues/3225

## Tasklist
 - [x] review
 - [x] adjust for comments

## Requirements / Relations
cc @MoKob @daniel-j-h

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,9 @@
-# 5.5.2
-  - Changes from 5.5.1
+# 5.6.0
+  - Changes from 5.5
     - Bugfixes
       - Fix #3475 removed an invalid `exit` field from the `arrive` maneuver
+    - Guidance
+      - No longer emitting turns on ferries, if a ferry should use multiple docking locations
 
 # 5.5.1
   - Changes from 5.5.0
--- a/features/guidance/collapse-ferry.feature
+++ b/features/guidance/collapse-ferry.feature
@@ -20,17 +20,17 @@ Feature: Collapse
             """
 
         And the ways
-            | nodes | highway | route | name |
-            | jacbk | primary |       | land |
-            | ad    |         | ferry | sea  |
-            | bd    |         | ferry | sea  |
-            | cd    |         | ferry | sea  |
-            | de    |         | ferry | sea  |
-            | ef    | primary |       | land |
+            | nodes | highway | route | name            |
+            | jacbk | primary |       | land            |
+            | ad    |         | ferry | sea             |
+            | bd    |         | ferry | sea             |
+            | cd    |         | ferry | sea             |
+            | de    |         | ferry | sea             |
+            | ef    | primary |       | pennydog-island |
 
         When I route I should get
-            | waypoints | route              | turns                                             | modes                         |
-            | f,j       | land,sea,land,land | depart,notification right,end of road left,arrive | driving,ferry,driving,driving |
+            | waypoints | route                         | turns                                      | modes                         | locations |
+            | f,j       | pennydog-island,sea,land,land | depart,notification right,turn left,arrive | driving,ferry,driving,driving | f,e,b,j   |
 
     Scenario: Switching Ferry in a Harbour
         Given the node map
@@ -49,16 +49,45 @@ Feature: Collapse
             """
 
         And the ways
-            | nodes | highway | route | name                  |
-            | ea    | primary |       | melee-island          |
-            | ab    |         | ferry | melee-island-ferry    |
-            | cf    | primary |       | pennydog-island       |
-            | bd    | primary |       | landmass              |
-            | bc    | primary | ferry | pennydog-island-ferry |
+            | nodes | highway | route | name                |
+            | ea    | primary |       | melee-island        |
+            | ab    |         | ferry | melee-island-ferry  |
+            | cf    | primary |       | monkey-island       |
+            | bd    | primary |       | scabb-island        |
+            | bc    | primary | ferry | monkey-island-ferry |
 
         When I route I should get
-            | waypoints | route                                                                                 | turns                                                                | modes                               |
-            | e,f       | melee-island,melee-island-ferry,pennydog-island-ferry,pennydog-island,pennydog-island | depart,notification straight,turn right,notification straight,arrive | driving,ferry,ferry,driving,driving |
+            | waypoints | route                                                                           | turns                                                                | modes                               |
+            | e,f       | melee-island,melee-island-ferry,monkey-island-ferry,monkey-island,monkey-island | depart,notification straight,turn right,notification straight,arrive | driving,ferry,ferry,driving,driving |
+
+
+    Scenario: End of Road Ferries
+        Given the node map
+            """
+            a - b ~ ~ ~ ~ c ~ ~ ~ ~ ~ d - e
+                          ~
+                          ~
+                          ~
+                          ~
+                          ~
+                          ~
+                          ~
+                          f
+                          |
+                          g
+            """
+
+        And the ways
+            | nodes | highway | route | name        |
+            | ab    | primary |       | land-left   |
+            | de    | primary |       | land-right  |
+            | gf    | primary |       | land-bottom |
+            | bcd   |         | ferry | ferry       |
+            | fc    |         | ferry | ferry       |
+
+        When I route I should get
+            | waypoints | route                                   | turns                                                     |
+            | g,e       | land-bottom,ferry,land-right,land-right | depart,notification straight,notification straight,arrive |
 
     Scenario: Fork Ferries
         Given the node map
@@ -85,6 +114,7 @@ Feature: Collapse
             | cb    |         | ferry | ferry       |
             | cd    |         | ferry | ferry       |
             | fc    |         | ferry | ferry       |
+
 
         When I route I should get
             | waypoints | route                                   | turns                                                  |
--- a/features/guidance/collapse-ferry.feature
+++ b/features/guidance/collapse-ferry.feature
@@ -1,0 +1,91 @@
+@routing  @guidance @collapsing
+Feature: Collapse
+
+    Background:
+        Given the profile "car"
+        Given a grid size of 20 meters
+
+    Scenario: Collapse Steps While On Ferry
+        Given the node map
+            """
+            j----a---c---b----k
+                  ~   ~  ~
+                    ~  ~ ~
+                       ~~~
+                         d
+                          ~
+                           ~
+                            ~
+                             e --- f
+            """
+
+        And the ways
+            | nodes | highway | route | name |
+            | jacbk | primary |       | land |
+            | ad    |         | ferry | sea  |
+            | bd    |         | ferry | sea  |
+            | cd    |         | ferry | sea  |
+            | de    |         | ferry | sea  |
+            | ef    | primary |       | land |
+
+        When I route I should get
+            | waypoints | route              | turns                                             | modes                         |
+            | f,j       | land,sea,land,land | depart,notification right,end of road left,arrive | driving,ferry,driving,driving |
+
+    Scenario: Switching Ferry in a Harbour
+        Given the node map
+            """
+                          d
+                          |
+                          |
+                          |
+            e - a ~ ~ ~ ~ b
+                          ~
+                          ~
+                          ~
+                          c
+                          |
+                          f
+            """
+
+        And the ways
+            | nodes | highway | route | name                  |
+            | ea    | primary |       | melee-island          |
+            | ab    |         | ferry | melee-island-ferry    |
+            | cf    | primary |       | pennydog-island       |
+            | bd    | primary |       | landmass              |
+            | bc    | primary | ferry | pennydog-island-ferry |
+
+        When I route I should get
+            | waypoints | route                                                                                 | turns                                                                | modes                               |
+            | e,f       | melee-island,melee-island-ferry,pennydog-island-ferry,pennydog-island,pennydog-island | depart,notification straight,turn right,notification straight,arrive | driving,ferry,ferry,driving,driving |
+
+    Scenario: Fork Ferries
+        Given the node map
+            """
+            a - b         d - e
+                 ~       ~
+                  ~     ~
+                   ~   ~
+                    ~ ~
+                     c
+                     ~
+                     ~
+                     ~
+                     f
+                     |
+                     g
+            """
+
+        And the ways
+            | nodes | highway | route | name        |
+            | ab    | primary |       | land-left   |
+            | de    | primary |       | land-right  |
+            | gf    | primary |       | land-bottom |
+            | cb    |         | ferry | ferry       |
+            | cd    |         | ferry | ferry       |
+            | fc    |         | ferry | ferry       |
+
+        When I route I should get
+            | waypoints | route                                   | turns                                                  |
+            | g,e       | land-bottom,ferry,land-right,land-right | depart,notification straight,notification right,arrive |
--- a/include/extractor/guidance/suppress_mode_handler.hpp
+++ b/include/extractor/guidance/suppress_mode_handler.hpp
@@ -15,6 +15,9 @@ namespace extractor
 namespace guidance
 {
 
+// Suppresses instructions for certain modes.
+// Think: ferry route. This handler suppresses all instructions while on the ferry route.
+// We don't want to announce "Turn Right", "Turn Left" while on ferries, as one example.
 class SuppressModeHandler final : public IntersectionHandler
 {
   public:
@@ -26,15 +29,13 @@ class SuppressModeHandler final : public IntersectionHandler
 
     ~SuppressModeHandler() override final = default;
 
-    bool canProcess(const NodeID /*nid*/,
+    bool canProcess(const NodeID nid,
                     const EdgeID via_eid,
                     const Intersection &intersection) const override final;
 
     Intersection operator()(const NodeID nid,
                             const EdgeID via_eid,
                             Intersection intersection) const override final;
-
-    using SuppressModeListT = std::array<osrm::extractor::TravelMode, 2>;
 };
 
 } // namespace osrm
--- a/include/extractor/guidance/suppress_mode_handler.hpp
+++ b/include/extractor/guidance/suppress_mode_handler.hpp
@@ -1,0 +1,44 @@
+#ifndef OSRM_EXTRACTOR_GUIDANCE_SUPPRESS_MODE_HANDLER_HPP_
+#define OSRM_EXTRACTOR_GUIDANCE_SUPPRESS_MODE_HANDLER_HPP_
+
+#include "extractor/guidance/constants.hpp"
+#include "extractor/guidance/intersection.hpp"
+#include "extractor/guidance/intersection_generator.hpp"
+#include "extractor/guidance/intersection_handler.hpp"
+#include "extractor/travel_mode.hpp"
+#include "util/node_based_graph.hpp"
+
+namespace osrm
+{
+namespace extractor
+{
+namespace guidance
+{
+
+class SuppressModeHandler final : public IntersectionHandler
+{
+  public:
+    SuppressModeHandler(const IntersectionGenerator &intersection_generator,
+                        const util::NodeBasedDynamicGraph &node_based_graph,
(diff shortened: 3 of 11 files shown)


emiltin:
the same should probably apply to other types of routes: trains, movable bridges, etc.

TheMarex:
Hm could we handle this in pre-processing instead of post-processing? Turning from a `starting_point = false` to another `starting_point = false` node should be a `NoTurn`.

/cc @MoKob

MoKob:
@TheMarex you are correct. If we assume for ferries always to be `staring_point == false`. But we don't event need this assumption. We know the travel modes. So we can check `TravelMode==Ferry` in Preprocessing and not emit any instruction.

daniel-j-h:
Sorry my mistake here. I implemented lane removal from roundabouts in post-processing and was under the impression @karenzshea could do the same quick fix there, too.

Now as @MoKob today told me in the context of https://github.com/Project-OSRM/osrm-backend/pull/2705 is that `suppressed` + `use lane` could still happen. The easier and correct solution is to not emit them in pre-processing only.

@karenzshea take a look at the changes in https://github.com/Project-OSRM/osrm-backend and how @MoKob handles not collapsing inside roundabouts and not assigning lanes there.

Maybe this is a good start.

src/extractor/guidance/intersection_handler.cpp
```diff
@@ -51,6 +51,11 @@ TurnType::Enum IntersectionHandler::findBasicTurnType(const EdgeID via_edge,
     const auto &in_data = node_based_graph.GetEdgeData(via_edge);
     const auto &out_data = node_based_graph.GetEdgeData(road.eid);
 
+    // suppress instructions on edges that the can't control anyway
+    if (!in_data.startpoint && !out_data.startpoint)
```
MoKob: In my view, we shouldn't abuse is startpoint here (and below). We want to filter out `Ferry` turns. As already stated in the discussion, we should use the TravelMode Ferry to do so. Someone might decide that starting on other roads is invalid, but we know that jumping ship is not possible. If someone disables motorways as start points, we might still want to give instructions.

MoKob:
Just a heads up. I've added a small test

MoKob:
To explain the reasoning behind my test-case:

I have the feeling that the two locations currently modified in this PR might not catch all ferry cases. In my view it would be easier to do a dedicated ferry handler in the beginning that checks for `allFerry` and immediately returns all NoTurns.

src/extractor/guidance/intersection_handler.cpp
TheMarex: I disagree here. This would mean that you need to add explicit handling based on mode for every mode you want to support. That means the probability that we need to find and adjust this piece of code in the future is very high, which I think is undesirable and duplicates logic from the profile. 

Using `startingpoint` uses a concept that we already expose in the profiles: Flagging ways that should be threaded as a single 'unit' from a user perspective. The name might be good or bad but this is effectively what this does. I don't see any scenario where we would reasonably emit an instruction between edges of this type. Any division between steps is done by the routing engine for technical reasons not user intend.

I feel this will be easier to use from a user perspective and offer less surprises then explicitly needing to cycle back every time a user finds a use-case for having `ferry`-like edges.

MoKob: The problem I see here is that what we snap to and what we consider as connections in forms of a turn are two entirely different concepts to me. Ferries are special in terms of that they often share ways in OSM data but don't allow switching from one item to another in between.

If we want to specify that snapping to is equivalent to does not allow turning, then you would be right. But as long as we don't make a distinction here, using `startingpoint` would not capture the issue we see in ferries. 

The concept we need are ways that share segments but are distinct (right now we don't handle these correctly in all cases, since we can right now actually switch ferries in between). Ferries will need specific handling anyhow. The main question for me here is: do we really want to tie snapping to turn instructions?

/cc @daniel-j-h @danpat for additional oppinions

include/extractor/travel_mode.hpp
```diff
@@ -53,4 +55,7 @@ const constexpr osrm::extractor::TravelMode TRAVEL_MODE_RIVER_UP = 10;
 const constexpr osrm::extractor::TravelMode TRAVEL_MODE_RIVER_DOWN = 11;
 const constexpr osrm::extractor::TravelMode TRAVEL_MODE_ROUTE = 12;
 
+// travel modes for which navigation should be suppressed
+const std::array<osrm::extractor::TravelMode, 2> constexpr SUPPRESS_MODE_LIST = {{TRAVEL_MODE_TRAIN, TRAVEL_MODE_FERRY}};
```
daniel-j-h: The problem with using std::array here is the size has to be known at compile time. If you now add an additional mode you have to go through the code base and change all array<T, N> to array<T, N + 1>.

TheMarex: Yeah this could be solved using a type alias though. `using SupressionModeArray = std::array<osrm::extractor::TravelMode, 2>;`

features/guidance/collapse-ferry.feature
```diff
@@ -0,0 +1,90 @@
+@routing  @guidance @collapsing
+Feature: Collapse
+
+    Background:
+        Given the profile "car"
+        Given a grid size of 20 meters
+
+    Scenario: Collapse Steps While On Ferry
```
MoKob: could you maybe directly add the `ascii-art` to make the tests more readable to begin with? (here and on the other tests)

features/guidance/collapse-ferry.feature
```diff
@@ -0,0 +1,90 @@
...
+            | bd    |         | ferry | sea  |
+            | cd    |         | ferry | sea  |
+            | de    |         | ferry | sea  |
+            | ef    | primary |       | land |
+
+        When I route I should get
+            | waypoints | route              | turns                                      | modes                         |
+            | f,j       | land,sea,land,land | depart,notification right,turn left,arrive | driving,ferry,driving,driving |
+
+    Scenario: Collapse Steps While On Ferry
```
MoKob: I believe we could remove this test. It looks like the only difference here to the test above is an additional coordinate on the ferry route, which should not influence any turn decisions.

features/guidance/collapse-ferry.feature
```diff
@@ -0,0 +1,90 @@
...
+        And the ways
+            | nodes | highway | route | name                  |
+            | ea    | primary |       | melee-island          |
+            | ab    |         | ferry | melee-island-ferry    |
+            | cf    | primary |       | pennydog-island       |
+            | bd    | primary |       | landmass              |
+            | bc    | primary | ferry | pennydog-island-ferry |
+
+        When I route I should get
+            | waypoints | route                                                                           | turns                                                                | modes                               |
```
MoKob: misaligned, also I miss my monkey island reference...

include/extractor/guidance/intersection_handler.hpp
```diff
@@ -57,6 +57,15 @@ class IntersectionHandler
     const IntersectionGenerator &intersection_generator;
     const NodeBasedGraphWalker graph_walker; // for skipping traffic signal, distances etc.
 
+    // Decide whether or not navigation should be suppressed on an edge based on mode type
+    bool SuppressModeNavigation(const TravelMode &in_mode,
+                                const TravelMode &out_mode,
+                                const SuppressModeListT &SUPPRESS_MODE_LIST) const;
+
+    bool SuppressModeNavigation(const EdgeID &via_edge,
```
MoKob: via_edge/connected are unintuitive to me. how about `from_edge` and `to_edge`? or `in_edge` and `out_edge`?

src/extractor/guidance/intersection_handler.cpp
```diff
@@ -37,13 +38,41 @@ IntersectionHandler::IntersectionHandler(const util::NodeBasedDynamicGraph &node
...
+    return IntersectionHandler::SuppressModeNavigation(in_mode, out_mode, SUPPRESS_MODE_LIST);
+}
+
+bool IntersectionHandler::SuppressModeNavigation(const TravelMode &in_mode,
+                                                 const TravelMode &out_mode,
+                                                 const SuppressModeListT &SUPPRESS_MODE_LIST) const
+{
+    const auto suppress_in_mode = std::find(begin(SUPPRESS_MODE_LIST), end(SUPPRESS_MODE_LIST), in_mode);
+    if (suppress_in_mode != end(SUPPRESS_MODE_LIST))
+    {
```
MoKob: minor nitpick: the check for `in_mode == out_mode` is cheaper than the algorithm call. So doing it the other way around would be faster:
```
if( in_mode != out_mode )
 return false;

const auto suppress_in_mode = std::find(begin(SUPPRESS_MODE_LIST), end(SUPPRESS_MODE_LIST), in_mode);

return (suppress_in_mode != SUPPRESS_MODE_LIST.end();
```

In addition: using `end` like this requires the inclusion of `iterator` and a `using std::end`.
In this case, it's probably easier to simply call `SUPPRESS_MODE_LIST.end()`.

src/extractor/guidance/intersection_handler.cpp
```diff
@@ -37,13 +38,41 @@ IntersectionHandler::IntersectionHandler(const util::NodeBasedDynamicGraph &node
 {
 }
 
+bool IntersectionHandler::SuppressModeNavigation(const EdgeID &via_edge,
+                                                 const EdgeID &connected,
+                                                 const SuppressModeListT &SUPPRESS_MODE_LIST) const
```
MoKob: I believe it would be cleaner to not pass the mode list as a parameter.
Since you already define it globally in `travel_mode.hpp`, you are essentially shadowing the name.

Shadowing describes a situation where you use the same name (here SUPPRESS_MODE_LIST) for two variables that are available at the same time. (https://en.wikipedia.org/wiki/Variable_shadowing). Simply accessing the global variable would be cleaner, in my opinion, since I don't see a reason where we would require greater flexibility at some point.

src/extractor/guidance/intersection_handler.cpp
```diff
@@ -37,13 +38,41 @@ IntersectionHandler::IntersectionHandler(const util::NodeBasedDynamicGraph &node
...
 TurnType::Enum IntersectionHandler::findBasicTurnType(const EdgeID via_edge,
                                                       const ConnectedRoad &road) const
 {
 
     const auto &in_data = node_based_graph.GetEdgeData(via_edge);
     const auto &out_data = node_based_graph.GetEdgeData(road.eid);
 
+    const auto in_mode = node_based_graph.GetEdgeData(via_edge).travel_mode;
+    const auto out_mode = node_based_graph.GetEdgeData(road.eid).travel_mode;
+    if (SuppressModeNavigation(in_mode, out_mode, SUPPRESS_MODE_LIST))
```
MoKob: As per https://github.com/Project-OSRM/osrm-backend/pull/3420#issuecomment-265997097, I've added a test. Using the three fallbacks is not sufficient to suppress all turns in ferries. I've added a test that still emits a turn on a ferry.

To fully solve this, I'd propose a dedicated handler (see previous comment). The handler should check all pairs of roads for same travel mode (e.g std::all_of(intersection,SuppressModeNavigation(first_mode,road_mode)) to check if it is applicable and simply assign `NoTurn` to all turns in its `operator()`.

features/guidance/collapse-ferry.feature
karenzshea: all the tests are rendered in ASCII now

src/extractor/guidance/suppress_mode_handler.cpp
```diff
@@ -0,0 +1,62 @@
...
+    const auto in_mode = node_based_graph.GetEdgeData(intersection[0].eid).travel_mode;
+    BOOST_ASSERT(std::find(begin(SUPPRESS_MODE_LIST), end(SUPPRESS_MODE_LIST), in_mode) != SUPPRESS_MODE_LIST.end());
+
+    // if all out ways share the same mode as the approach way, suppress all the turns
+    const auto all_share_mode =
+        std::all_of(begin(intersection) + 1, end(intersection), [this, &in_mode](ConnectedRoad way) {
+            return node_based_graph.GetEdgeData(way.eid).travel_mode == in_mode;
+        });
+
+    if (all_share_mode)
```
karenzshea: Originally I'd been selectively suppressing only instructions onto ways that involved modes on the mode blacklist, but this failed a test added by @MoKob wherein there is an intersection with a new ferry line and a land-based way.

https://github.com/Project-OSRM/osrm-backend/pull/3420/files#diff-0bf26f655163cdd60bedcd61e763028dR35

The test is written such that the instruction at the stopover point is expected to be suppressed. 

My thinking on this is that without a `destination` based instruction when the _first_ ferry line is boarded, it's not exactly useful to emit an instruction at the "stopover" if there really is a stopover at all or just a change of direction? Do we have a way of knowing like in the test above whether or not a switch to a new ferry involves user action? I actually would kind of assume that switching to a new ferry line necessitates user action, but I defer to @MoKob who wrote the original test.

danpat: Ah, ok.  I guess there's a larger discussion to be had around route relations, ferries/trains changing name, stopovers, etc, but I think the logic here handles the egregious cases that the original ticket identified.  👍  on this logic for now.

include/extractor/guidance/suppress_mode_handler.hpp
```diff
@@ -0,0 +1,42 @@
...
+#include "util/node_based_graph.hpp"
+
+namespace osrm
+{
+namespace extractor
+{
+namespace guidance
+{
+
+class SuppressModeHandler : public IntersectionHandler
```
daniel-j-h: I think you can make it `SuppressModeHandler final : ` - all handler could be final except the base class `IntersectionHandler`

include/extractor/travel_mode.hpp
```diff
@@ -53,4 +56,7 @@ const constexpr osrm::extractor::TravelMode TRAVEL_MODE_RIVER_UP = 10;
 const constexpr osrm::extractor::TravelMode TRAVEL_MODE_RIVER_DOWN = 11;
 const constexpr osrm::extractor::TravelMode TRAVEL_MODE_ROUTE = 12;
 
+// travel modes for which navigation should be suppressed
+const osrm::extractor::SuppressModeListT constexpr SUPPRESS_MODE_LIST = {{TRAVEL_MODE_TRAIN, TRAVEL_MODE_FERRY}};
+
```
daniel-j-h: Should this be in the SuppressModeHandler then?

src/extractor/guidance/intersection_handler.cpp
```diff
@@ -44,6 +45,9 @@ TurnType::Enum IntersectionHandler::findBasicTurnType(const EdgeID via_edge,
     const auto &in_data = node_based_graph.GetEdgeData(via_edge);
     const auto &out_data = node_based_graph.GetEdgeData(road.eid);
 
+    const auto in_mode = node_based_graph.GetEdgeData(via_edge).travel_mode;
+    const auto out_mode = node_based_graph.GetEdgeData(road.eid).travel_mode;
```
daniel-j-h: Hm? Dead code?

src/extractor/guidance/suppress_mode_handler.cpp
```diff
@@ -0,0 +1,62 @@
...
+    const auto in_mode = node_based_graph.GetEdgeData(via_eid).travel_mode;
+    const auto suppress_in_mode =
+        std::find(begin(SUPPRESS_MODE_LIST), end(SUPPRESS_MODE_LIST), in_mode);
+    return suppress_in_mode != end(SUPPRESS_MODE_LIST);
+}
+
+Intersection SuppressModeHandler::
+operator()(const NodeID /*nid*/, const EdgeID source_edge_id, Intersection intersection) const
+{
+    const auto in_mode = node_based_graph.GetEdgeData(intersection[0].eid).travel_mode;
```
daniel-j-h: This is the travel mode for the UTurn road?

src/extractor/guidance/suppress_mode_handler.cpp
```diff
@@ -0,0 +1,62 @@
...
+    BOOST_ASSERT(std::find(begin(SUPPRESS_MODE_LIST), end(SUPPRESS_MODE_LIST), in_mode) != SUPPRESS_MODE_LIST.end());
+
+    // if all out ways share the same mode as the approach way, suppress all the turns
+    const auto all_share_mode =
+        std::all_of(begin(intersection) + 1, end(intersection), [this, &in_mode](ConnectedRoad way) {
```
daniel-j-h: const auto& road

src/extractor/guidance/suppress_mode_handler.cpp
```diff
@@ -0,0 +1,62 @@
...
+            return node_based_graph.GetEdgeData(way.eid).travel_mode == in_mode;
+        });
+
+    if (all_share_mode)
+    {
+        std::for_each(begin(intersection) + 1, end(intersection), [&](ConnectedRoad &out_way) {
+            out_way.instruction = {TurnType::Suppressed, getTurnDirection(out_way.angle)};
+        });
+    }
```
daniel-j-h: If you

    auto first = begin(intersection) + 1;
    auto last = end(intersection);

above you can re-use them here

src/extractor/guidance/turn_analysis.cpp
```diff
@@ -143,6 +148,13 @@ Intersection TurnAnalysis::AssignTurnTypes(const NodeID node_prior_to_intersecti
                 road.instruction.type = TurnType::OffRamp;
         });
     }
+    // Suppress turns on ways between mode types that do not need guidance
+    if (suppress_mode_handler.canProcess(
+            node_prior_to_intersection, entering_via_edge, intersection))
+    {
+        intersection = suppress_mode_handler(
+            node_prior_to_intersection, entering_via_edge, std::move(intersection));
+    }
```
daniel-j-h: Your handler should come first, since it could influence others.

daniel-j-h: Also if your handler fires you have to disable all other handlers. Otherwise it could be that e.g. your handler changes instructions to NoTurn and now the Sliproad Handler accidentally is getting confused with the NoTurn and classifies this as a Sliproad.

Make the handler come first and if it triggers skip all other handler.

include/extractor/travel_mode.hpp
karenzshea: ohh good thinking, this doesn't need to be a global anymore

src/extractor/guidance/suppress_mode_handler.cpp
karenzshea: :+1:

daniel-j-h:
Mason macOS build is failing due to `std::abs` issue already stumbled upon in https://github.com/Project-OSRM/osrm-backend/issues/1000.

@springmeyer you looked into this back then, any ideas why it is failing only now on this branch and e.g. not on master? The file failing to compile is the same as on master after all.

Build is

https://travis-ci.org/Project-OSRM/osrm-backend/builds/183957985

and error is at

https://travis-ci.org/Project-OSRM/osrm-backend/jobs/183957988#L1179

oxidase:
@daniel-j-h just rebase is needed. It was "fixed" by https://github.com/Project-OSRM/osrm-backend/pull/3409/commits/b1aced327923c14d848b4a83f64fdbe3d29df3e9#diff-4ff77436d81ab7e8c0c026734a2d79cdR6
in #3409

daniel-j-h:
Rebased and refactored a bit. @MoKob could you have a final look over this.

Note: needs to be squashed before merging.

CHANGELOG.md
```diff
@@ -1,7 +1,9 @@
-# 5.5.2
-  - Changes from 5.5.1
+# 5.6.0
+  - Changes from 5.5
     - Bugfixes
       - Fix #3475 removed an invalid `exit` field from the `arrive` maneuver
+    - Guidance
+      - No longer emitting turns on ferries, if a ferry should use multiple docking locations
```
daniel-j-h: Hm @MoKob, I just wanted to add one but it looks like you already added a changelog entry here - what's missing then?

MoKob: yeah. I just added one, after I said it in chat. Sorry for not updating you. Other than that, if you are OK with my last changes, it should be good to go.

PR #3426, opened 2016-12-08
Makes GeoJSON conforming to spec, closes #3425.

### LINKED ISSUE #3425: Point should not be "coordinates": [[lon, lat]], but [lon, lat] when setting "geometry=geojson"
Hello,
At present, a query with`geometry=geojson` results in:

```
$ curl "http://localhost:5000/route/v1/driving/133.540336,33.558187;133.540937,33.55841?steps=true&geometries=geojson" | jq -c '.routes[0].legs[].steps[].geometry | select(.type == "Point")'

{"type":"Point","coordinates":[[133.540906,33.558277]]}
```

But a correct geojson should be the following:
```
{"type": "Point", "coordinates": [133.540906, 33.558277]}
(or {"type": "MultiPoint", "coordinates": [[133.540906, 33.558277]]}).
```

daniel-j-h:
For #3425. Read your comment to late @TheMarex - this makes the response GeoJSON spec conforming. Want me to implement always-linestring instead?

## Tasklist

 - [x] review
 - [x] adjust for comments

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,6 +29,7 @@
       - fixed a bug that could result in crashes when leaving a ferry directly onto a motorway ramp
       - fixed a bug in the tile plugin that resulted in discovering invalid edges for connections
       - improved error messages when missing files during traffic updates (#3114)
+      - For single coordinate geometries the GeoJSON `Point` encoding was broken. We now always emit `LineString`s even in the one-coordinate-case (backwards compatible) (#3425)
     - Debug Tiles
       - Added support for turn penalties
     - Internals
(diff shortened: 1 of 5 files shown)


daniel-j-h:
Last commit changes from making `Point` conforming to the spec to always returning a `LineString`.

Reading the spec: https://github.com/Project-OSRM/osrm-backend/blob/8ae981e14238c3236cbb1acad6b6d6260d306f86/docs/http.md#routestep-object

> GeoJSON LineString or GeoJSON Point if it is only one coordinate (not wrapped by a GeoJSON feature)

Users will still notice this when they have code detecting a single coordinate. Not sure how strict our API guarantees are here.

- [ ] Update `http.md` docs
- [ ] Squash before merge

TheMarex:
@daniel-j-h seeing that we need to roll another RC anyway: Could you add a changelog entry so we could just roll this all up in RC4?

Issue #3429, opened 2016-12-08
Assert during map-matching path unpacking


danpat:
It seems that some combinations of traffic/turn cost data can lead to the following assertion being thrown:

```
[assert] /Users/danpat/mapbox/osrm-backend/include/engine/routing_algorithms/routing_base.hpp:223
in: void osrm::engine::routing_algorithms::BasicRoutingInterface<osrm::engine::datafacade::BaseDataFacade, osrm::engine::routing_algorithms::MapMatching<osrm::engine::datafacade::BaseDataFacade> >::UnpackPath(const DataFacadeT &, RandomIter, RandomIter, const osrm::engine::PhantomNodes &, std::vector<PathData> &) const [DataFacadeT = osrm::engine::datafacade::BaseDataFacade, Derived = osrm::engine::routing_algorithms::MapMatching<osrm::engine::datafacade::BaseDataFacade>, RandomIter = std::__1::__wrap_iter<const unsigned int *>]: *std::prev(packed_path_end) == phantom_node_pair.target_phantom.forward_segment_id.id || *std::prev(packed_path_end) == phantom_node_pair.target_phantom.reverse_segment_id.id
libc++abi.dylib: terminating
```

The assertion is triggered with this request:

http://localhost:5000/match/v1/driving/-122.505588,37.872619;-122.505702,37.872678?annotations=true&geometries=geojson

which represents this:

![screen shot 2016-12-08 at 1 12 31 pm](https://cloud.githubusercontent.com/assets/1892250/21030855/857b21ec-bd55-11e6-8f82-d6aec02abca2.png)

With no traffic data/turn penalty data applied, it works as expected.  The problem thus seems to be either an edge-case caused by the data, or something we're doing wrong when applying traffic data.

After chatting with @TheMarex, a suspect is https://github.com/Project-OSRM/osrm-backend/blob/master/include/contractor/graph_contractor.hpp#L826

danpat:
On the assumption that @TheMarex was onto something, I dug into `contractor.hpp`.  It looks like the CH algorithm depends on the data in the `.enw` file - but those node weights are not being updated if traffic data is applied.

MoKob:
I've tracked down the problem to the following issue:

due to the values put into the engine, we end up with negative initial segments that before could only happen if you looked at a route from `s` to `s`. Since we now don't have that guarantee anymore, we find invalid loops where none should exist, resulting in the assertion to trigger.

[...]

Issue #2016, opened 2016-02-25
error with self-loops


ivantrouvetout:
I think there is a problem with U-turns
http://map.project-osrm.org/?z=18&center=46.301289%2C4.870076&loc=46.302107%2C4.869497&loc=46.301803%2C4.870875&loc=46.301892%2C4.870575&loc=46.300331%2C4.869229&hl=en&ly=&alt=&df=&srv=

MoKob:
This is only partly related to u-turns. Since your query does not allow them at the intermediate vertices, the route is computed without u-turns at via nodes.
However, the routing fails to find the (very long) loop in this case.
I've updated the ticket accordingly.

MoKob:
reopening as the basic problem is not yet fixed

TheMarex:
Looking at the route again, I don't see a faster alternative. Was this fixed? As @MoKob already outlined this is the expected behavior as uturns in the middle of the street are not allowed.

MoKob:
The route here is found randomly based on the contraction order. The general problem should still exist. I will have a look tomorrow and see whether I can find a reason that results in the omission of some required self-loops

MoKob:
After some investigation we found the issue to be the following:

A prerequisite for this issue is the request of a route of the form `s -- v -- t` with `v` and `t` on the same segment (in the reported case, this was the two via-nodes, the final destination is not required to trigger it).

Next to this, the optimal path has to be a self-loop (required due to the forbidden `u-turn` in the middle of the street) -- that is it has to start and end at the same road segment.

OSRM assumes a self-loop to be only necessary for one-way streets or if the direct connection is slower than the loop around (think very slow forrest road, for example).
In the described case, the forbidden `u-turn` translates the normal road into a implied oneway street.
This results in the connection to be missed.

If, in addition, the street is higher up in the hierarchy than connecting other roads, we will not be able to find a path at all.

Potentially we could make sure to at least find a path if we allow `u-turns` with a very high penalty.

Due to the specific edge-case, we are postponing this issue to at least `5.1`

PR #3430, opened 2016-12-08
Update node weights if traffic data is applied

danpat:
# Issue

This is an attempt at fixing #3429.  Our extractor saves the edge-based-node weights into the `.enw` file.  However, if traffic data is applied, those node weights are invalidated, yet we still use them during graph contraction.

This PR moves the location that the `.enw` file is loaded and updates node weights with new values if traffic data is supplied.

With this change, can no longer reproduce the problem described in #3429

@MoKob @TheMarex can you sanity check this please?

TODO: an isolated test case - I have so far only reproduced this problem with large sets of internal data, I haven't narrowed it down to something I document as a cucumber test.

## Tasklist
 - [x] review
 - [x] adjust for comments

### DIFF
--- a/src/contractor/contractor.cpp
+++ b/src/contractor/contractor.cpp
@@ -136,12 +136,24 @@ int Contractor::Run()
 
     TIMER_START(preparing);
 
+    util::Log() << "Reading node weights.";
+    std::vector<EdgeWeight> node_weights;
+    std::string node_file_name = config.osrm_input_path.string() + ".enw";
+
+    {
+        storage::io::FileReader node_file(node_file_name,
+                                          storage::io::FileReader::VerifyFingerprint);
+        node_file.DeserializeVector(node_weights);
+    }
+    util::Log() << "Done reading node weights.";
+
     util::Log() << "Loading edge-expanded graph representation";
 
     util::DeallocatingVector<extractor::EdgeBasedEdge> edge_based_edge_list;
 
     EdgeID max_edge_id = LoadEdgeExpandedGraph(config.edge_based_graph_path,
                                                edge_based_edge_list,
+                                               node_weights,
                                                config.edge_segment_lookup_path,
                                                config.edge_penalty_path,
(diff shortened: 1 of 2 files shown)


src/contractor/contractor.cpp
```diff
@@ -914,6 +916,8 @@ EdgeID Contractor::LoadEdgeExpandedGraph(
                 previous_osm_node_id = segmentblocks[i].this_osm_node_id;
             }
 
+            node_weights[inbuffer.source] = new_weight;
```
danpat: This is the only bit that's actually *new* here.

MoKob:
@danpat I have serious doubts if the fix you did is actually the cause of the problem, but I cannot fully understand whats going on, yet. The code using the node weights is required to find self-loops. So starting on a segment, ending on a segment and driving a loop in between (e.g. bearings front/back in the wrong direction) or if you consider a longer slow road with a fast route going around.

The fix you did might, by chance, actually prevent the issue from surfacing, but the only difference we should see on traffic stacks is whether or not we find a traffic loop. As long as a direct path exists, the (which should be a given in map matching), the effect should not be visible at all.
The worst thing that should happen, if the node weights are incorrect, is that we add an additional edge (if the edges around get way faster) or not find a path (when the node weight does not capture the slowness of the way).

MoKob:
Aside from potential other problems. This PR can be merged still to update the node weights to be more accurate.

TheMarex:
@danpat I think it is okay to skip on the test-case here as in it's current form it is mostly a performance fix for the loop pruning heuristic in the CH code.

MoKob:
Technically it's also a question of loop correctness (which is broken to begin with, some other issues here). So I'd be ok with skipping the test as well. (see https://github.com/Project-OSRM/osrm-backend/issues/2016)

TheMarex:
Rebased and squashed, waiting on travis then merging.

TheMarex:
Damn forgot to merge this before doing the next RC.

PR #3434, opened 2016-12-12
fix invalid assertion in coordinate_extractor

MoKob:
# Issue

Discovered while reviewing https://github.com/Project-OSRM/osrm-backend/pull/3264. Invalid assertion.

## Tasklist
 - [x] review
 - [x] adjust for comments

## Requirements / Relations
 none

### DIFF
--- a/src/extractor/guidance/coordinate_extractor.cpp
+++ b/src/extractor/guidance/coordinate_extractor.cpp
@@ -132,7 +132,7 @@ util::Coordinate CoordinateExtractor::ExtractRepresentativeCoordinate(
     {
         const auto result = ExtractCoordinateAtLength(
             skipping_inaccuracies_distance, coordinates);
-        BOOST_ASSERT(is_valid_result(coordinates.back()));
+        BOOST_ASSERT(is_valid_result(result));
         return result;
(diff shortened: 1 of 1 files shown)


src/extractor/guidance/coordinate_extractor.cpp
```diff
@@ -132,7 +132,7 @@ util::Coordinate CoordinateExtractor::ExtractRepresentativeCoordinate(
     {
         const auto result = ExtractCoordinateAtLength(
             skipping_inaccuracies_distance, coordinates);
-        BOOST_ASSERT(is_valid_result(coordinates.back()));
+        BOOST_ASSERT(is_valid_result(result));
```
daniel-j-h: This asserts the result is neither the to nor the intersection node? The assertion before therefore could trigger at any time?

Should this be backported to 5.5?

MoKob: Essentially it could trigger on any dead end. Yes. Since intersection_node == to_node, it will always trigger if this case is reached.

TheMarex:
Yeah this should go into 5.5

Issue #3435, opened 2016-12-12
Car profile go over steps


frodrigo:
When way are steps (highway=steps) and access tag whitelist (eg access=permissive) the router profile can use the steps.

http://map.project-osrm.org/?z=18&center=48.841233%2C2.318432&loc=48.841316%2C2.318099&loc=48.841030%2C2.318748&hl=en&alt=0

daniel-j-h:
Way we route over: https://www.openstreetmap.org/way/30019448#map=19/48.84148/2.31917&layers=D

and nearby. Urgh this is bad. I'll dig in and investigate.

Thanks for reporting!

daniel-j-h:
Here's the underlying issue: it's the `access=` tag overriding the allowed transportation mode. This is in line with the OSM Wiki page on access tags:

https://wiki.openstreetmap.org/wiki/Key:access#Transport_mode_restrictions

> Use the access=* key to describe a general access restriction that **applies to all transport modes**. Note, that, for example, adding **access=yes to highway=footway** changes default restrictions (which usually are foot=yes and vehicle=no for highway=footway) to yes, highway=footway + access=yes means "**road, which is open for all pedestrians and vehicles**". Be very careful when adding general permitting tags access=yes and access=permissive, think about adding precise correct tags with concrete transport modes. If you want, for example, distingush footway with open access from footway with closed access, use tags like foot=yes and foot=private instead of access=yes and access=private.

Which means `access=permissive` + `highway=*` allows for _all transportation modes_ including cars to route over it if we strictly follow the definition in the Wiki. On the other hand sampling the data in Berlin this would also mean all `highway=steps` with an access tag we allow are now routable for cars:

![map](https://cloud.githubusercontent.com/assets/527241/21097520/22b5d7c4-c065-11e6-9b7b-7b4c249cb075.png)

Check the query: http://overpass-turbo.eu/s/kBa
> highway=steps and access~"yes|permissive|designated|destination"

Most if not all of them should not be routed over with a car.
If we strictly follow the Wiki these are mis-tagged.

On the other hand decoupling access tags from highway tags is confusing at best.

cc @systemed let's continue the discussion from IRC here.

frodrigo:
The question is not if it's allowed or not (according to the OSM tags). But since it's steps, you can't go through due to physical issues because we are using a car.

daniel-j-h:
Interestingly enough here is the Wiki page for computing access restrictions:

[...]

PR #3436, opened 2016-12-12
Adds `access=permissive` routability scenarios, see #3435.

daniel-j-h:
cc @MoKob

### DIFF
--- a/features/car/access.feature
+++ b/features/car/access.feature
@@ -220,3 +220,11 @@ Feature: Car - Restricted access
         Then routability should be
             | highway | toll        | bothw |
             | primary | yes         |       |
+
+    Scenario: Car - Access combinations
+        Then routability should be
+            | highway    | access     | bothw |
+            | primary    | permissive | x     |
+            | steps      | permissive | x     |
+            | footway    | permissive | x     |
+            | garbagetag | permissive | x     |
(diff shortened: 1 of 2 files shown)


features/car/access.feature
```diff
@@ -220,3 +220,11 @@ Feature: Car - Restricted access
...
             | highway | toll        | bothw |
             | primary | yes         |       |
+
+    Scenario: Car - Access combinations
+        Then routability should be
+            | highway    | access     | bothw |
+            | primary    | permissive | x     |
+            | steps      | permissive | x     |
+            | footway    | permissive | x     |
+            | garbagetag | permissive | x     |
```
daniel-j-h: Looked into this - this scenario is passing at the moment.

See #3435 for a discussion if it should pass or if the highway tag should take precedence.

MoKob:
@daniel-j-h is this still an issue? can we simply merge this, or do you think this should produce different results. I can't really tell from the discussion in #3435

daniel-j-h:
The Wiki says one thing the real world data says another. #3435 is not resolved. My opinion is routing cars over steps makes no sense. Feel free to close this PR - we can revive it once we resolved #3435.

TheMarex:
Replaced by https://github.com/Project-OSRM/osrm-backend/pull/3668

PR #3437, opened 2016-12-12
cache minimal adjacent edge weights to speed up contraction

MoKob:
# Issue

Contraction of large graphs is quite compute intensive. I've tried to see if the contraction can be reduced in time.

The result is pretty simple: We require a bit more complexity and more memory (one EdgeWeight entry for all nodes) during contraction.

This comes along with a reduction of about 5% in contraction time.
Germany:
 - master: 660s
 - this branch: 630s

In my view, while a nice experiment, the reduction in processing time is probably not worth the additional memory overhead/the added complexity.

Opinions?

## Tasklist
 - [ ] discuss if reasonable enough
 - [ ] review
 - [ ] adjust for comments

## Requirements / Relations
none

### DIFF
--- a/include/contractor/graph_contractor.hpp
+++ b/include/contractor/graph_contractor.hpp
@@ -133,6 +133,137 @@ class GraphContractor
         EnumerableThreadData data;
     };
 
+    EdgeWeight GetMinimalAdjacentWeight(const NodeID nid) const
+    {
+        const auto range = contractor_graph->GetAdjacentEdgeRange(nid);
+        if (range.empty())
+            return std::numeric_limits<EdgeWeight>::max() / 2;
+
+        const auto edge_weight = [this](const EdgeID edge_id) -> EdgeWeight {
+            return contractor_graph->GetEdgeData(edge_id).weight;
+        };
+
+        // cannot be std::min_elment, since edge_range does not fit the concept
+        auto minimum = edge_weight(range.front());
+        for (auto eid : range)
+            minimum = std::min(minimum, edge_weight(eid));
+        return minimum;
+    }
+
+    std::vector<EdgeWeight> BuildMinimalAdjacentWeightCache() const
+    {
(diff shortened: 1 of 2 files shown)


include/contractor/graph_contractor.hpp
```diff
@@ -133,6 +133,137 @@ class GraphContractor
...
+            [this, &minimal_adjacent_weight_cache](const tbb::blocked_range<NodeID> &range) {
+                for (int x = range.begin(), end = range.end(); x != end; ++x)
+                    minimal_adjacent_weight_cache[x] = GetMinimalAdjacentWeight(x);
+            });
+
+        return minimal_adjacent_weight_cache;
+    }
+
+    /* Flush all data from the contraction to disc and reorder stuff for better locality */
+    void FlushDataAndRebuildContractorGraph(ThreadDataContainer &thread_data_list,
```
TheMarex: Independently from whether this approach will work or not, I think this refactor makes sense.

TheMarex:
Can you get some precise measurements on the memory overhead? The server logs should contain the number of edge-based-nodes for the planet which should give you the size of the vector and a rough estimate.

Other then that, this would need some measuring in our actual setup. Things todo there are:
- Create a node pre-release from this branch:
   - Adapt the `package.json` to point to this branch and change the version suffix to `-minweight.1`.
   - `git commit --allow-empty "[publish binary] 5.5.0-minweight.1"`
   - `npm publish --tag test` (the `--tag` is important otherwise this is pushed out as a production version)
 - Launch a new staging stack of our pipeline and take measurements (initial create steps and traffic update time)

Code complexity does not increase too much with this additional cache, so I'm not worried there.

MoKob:
The refactor has made it into master. I've thought quite a bit about this, but I feel that the additional memory/complexity (even though not really complex, but it would hinder further refactoring) is just not worth it right now.

Closing.

Issue #3438, opened 2016-12-12
Node Ranges in GraphContractor should use sane types


MoKob:
While probably not critical right now, the graph contractor uses a wide range of integer types for contraction. Right now, we iterate over `int` in most cases.

These cases should all refer to `NodeID` which is the integer type we currently support.

One example is [here](https://github.com/Project-OSRM/osrm-backend/blob/8c7f744b1ad1c66b5c98a1572581e9ffb856605f/include/contractor/graph_contractor.hpp#L281-L283). If the number of edge based nodes is larger than what fits in an int, we don't initialise all nodes.

PR #3439, opened 2016-12-12
use correct type in graph contractor to avoid overflows in nodes

MoKob:
# Issue

With the ever growing number of nodes, we should prematurely fix https://github.com/Project-OSRM/osrm-backend/issues/3438 before it becomes a problem.

## Tasklist
 - [x] review
 - [x] adjust for comments

## Requirements / Relations
none


daniel-j-h:
Looks good - @TheMarex do you know why some of them were `size_t` and some of them were `int`?

TheMarex:
@daniel-j-h no this was (half)-automatically ported from OpenMP to TBB at the time. I think the original version just used `int`.

Issue #2445, opened 2016-05-23
Log peak STXXL/memory usage


danpat:
Because OSRM is so resource heavy (disk, RAM), we should log resource usage as an aid for server planning.

Specifically:
- [x] Track and log peak real RAM allocation
- [x] Track and log peak STXXL usage

daniel-j-h:
@danpat hint: dump [capacity() / raw_capacity()](https://github.com/stxxl/stxxl/blob/5b9663e6b769748f3b3d3a9a779b4b89e24d7a27/include/stxxl/bits/containers/vector.h#L1037-L1047) for `stxxl::vector`.

daniel-j-h:
This was done in https://github.com/Project-OSRM/osrm-backend/pull/3440. Closing here.

@karenzshea @miccolis might be a good idea to get statistics and alarms in place for these numbers; example how it looks like:

```
[info] STXXL: peak bytes used: 150994944
[info] STXXL: total disk allocated: 1048576000
[info] RAM: peak bytes used: 1011208192
```

PR #3440, opened 2016-12-12
Output resource usage stats at end of extract/contract jobs

danpat:
# Issue

This PR addresses #2445.  At the end of `osrm-extract` and `osrm-contract`, there are now some additional log lines:

```
[info] STXXL: peak bytes used: 150994944
[info] STXXL: total disk allocated: 1048576000
[info] RAM: peak bytes used: 1011208192
```

These should be helpful for monitoring production servers, and not using machines that are any larger than necessary.

If using STXXL < 1.4, a warning message will be issued instead (STXXL < 1.4 lacks the required functions).  Peak RAM measurement isn't implemented for Windows and results in another warning.

Comments appreciated.

## Tasklist
 - [x] review
 - [x] adjust for comments

### DIFF
--- a/include/util/meminfo.hpp
+++ b/include/util/meminfo.hpp
@@ -1,0 +1,43 @@
+#ifndef MEMINFO_HPP
+#define MEMINFO_HPP
+
+#include "util/log.hpp"
+
+#include <stxxl/mng>
+#ifndef _WIN32
+#include <sys/resource.h>
+#endif
+
+namespace osrm
+{
+namespace util
+{
+inline void DumpMemoryStats()
+{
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+# 5.6.0
+  - Changes from 5.5.0
+    - osrm-extract/osrm-contract will now output peak memory (STXXL/RAM) usage upon completion
```
MoKob: maybe we should drop a note here, that it does not work on windows, pre 1.4?

daniel-j-h: Also Linux Kernel requirement

> ru_maxrss (since Linux 2.6.32)

PR #3442, opened 2016-12-13
Works around Unreachable Warning for Debug Build

daniel-j-h:
cc @MoKob


MoKob:
Thanks for fixing this.

Issue #3443, opened 2016-12-13
Unexpected Behavious in Small Components


MoKob:
The combination of small SCCs and small Components can introduce quite unintuitive routing behaviour.

See [1](https://www.mapbox.com/osrm-frontend-v2/?z=16&center=50.229610%2C12.878140&loc=50.226755%2C12.878600&loc=50.227054%2C12.878444&hl=en&ly=&alt=&df=&srv=&profile=mapbox.driving), [2](https://www.mapbox.com/osrm-frontend-v2/?z=18&center=50.225725%2C12.877160&loc=50.225733%2C12.876974&loc=50.225726%2C12.877339&hl=en&ly=&alt=&df=&srv=&profile=mapbox.driving), [3](https://www.mapbox.com/osrm-frontend-v2/?z=18&center=45.465035%2C9.185680&loc=45.465522%2C9.185718&loc=45.465308%2C9.186024&hl=en&ly=&alt=&df=&srv=&profile=mapbox.driving).

As long as we can find a route from between the requested points, we shouldn't prematurely snap to large components.

TheMarex:
The problem with that is, this approach will not work for other plugins, generating a different snapping behavior and unexpected results. The query for a route from A to B should have the same snapping behavior as doing a distance table query with A and B as coordinates. We have consumers like vroom depending on this.

Let's say you would want to do this with distance tables, then you first create a distance table with `PhantomNode`s from the small components. Some entries will be empty because there is no route, others will have values. When do you decide to switch to the bigger component entries?

Previous iterations on this problem have already explored placing both small and big component where added to the search heaps. This will have the result that paths from A->B and B->C will not necessarily be equal to A->C because if A and B are in small components the search could return paths for A->B and A->C where A is actually not snapped to the same road segments. This lead to major confusions for distance tables.

What we ideally want to have is some sort of value/data structure that tells us immediately if a path A->B exists. The SCC tells us if there is a path from A->B and B->A. The CC only if there is a path from A->B or B->A. So far I couldn't find something that would actually yield this directed property. (Hub Labels maybe, but it kind of defeats the purpose)

Another option would be to remove all small/big component checks and always use the nearest neighbor. That would remove some server-side ugliness, but would mean a lot more failed routes. Customers care a lot about routes not failing, even if the snapping is weird.

We could also not include small components in the RTree. This would mean that small islands are not routable which is also undesirable (and would actually have the same result as the linked examples).

Open for suggestions here.

[...]

PR #3444, opened 2016-12-13
allow routing between small components if path exists

MoKob:
# Issue

First step on the way to https://github.com/Project-OSRM/osrm-backend/issues/3443. This PR only handles direct shortest paths and does not handle via-routes.

## Tasklist
 - [ ] review
 - [ ] adjust for comments

## Requirements / Relations
 none

### DIFF
--- a/src/engine/plugins/viaroute.cpp
+++ b/src/engine/plugins/viaroute.cpp
@@ -82,21 +82,46 @@ Status ViaRoutePlugin::HandleRequest(const std::shared_ptr<datafacade::BaseDataF
                 !continue_straight_at_waypoint;
         }
     };
-    util::for_each_pair(snapped_phantoms, build_phantom_pairs);
 
-    if (1 == raw_route.segment_end_coordinates.size())
+    if (phantom_node_pairs.size() == 2)
     {
-        if (route_parameters.alternatives && facade->GetCoreSize() == 0)
+        // first try the route with the original input, if a route should exist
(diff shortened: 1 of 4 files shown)


src/engine/plugins/viaroute.cpp
```diff
@@ -82,21 +82,46 @@ Status ViaRoutePlugin::HandleRequest(const std::shared_ptr<datafacade::BaseDataF
                 !continue_straight_at_waypoint;
         }
     };
-    util::for_each_pair(snapped_phantoms, build_phantom_pairs);
 
-    if (1 == raw_route.segment_end_coordinates.size())
+    if (phantom_node_pairs.size() == 2)
```
oxidase: :+1: for removing Yoda style

MoKob: Yeah. kind of doesn't help next to `const` functions. Can save you from `if( var = 2 )` errors, though.

daniel-j-h: I think this is on by default (`-Wall`):

`-Wparentheses`
> Warn if parentheses are omitted in certain contexts, such as when there is an assignment in a context where a truth value is expected, or when operators are nested whose precedence people often get confused about.

TheMarex:
The main problem I see here is consistency in snapping behavior. Sure this will work for two endpoints but how does this generalize to three or four? How does this work with distance matrices? Keeping the discussion in #3443.

I think we should wait until we have a solution that generalizes before moving on changing behavior in one plugin only.

Issue #581, opened 2013-02-06
Routing over a simple barrier=gate node with no further access-tags should be avoided


Schiki:
See http://osrm.at/2iC

Regards Schiki

daniel-j-h:
We route over `barrier=gate` without further access restrictions in the _default profiles_ by now. 
Otherwise this would break routability for many users.

PR #3445, opened 2016-12-13
Adds barrier=gate regression tests, see #581

daniel-j-h:
See https://github.com/Project-OSRM/osrm-backend/issues/581 and the recent discussion in https://github.com/Project-OSRM/osrm-backend/issues/3435.

These tests all pass even though [`barrier=*`](http://wiki.openstreetmap.org/wiki/Key:barrier) implies `access=no` per Wiki.

But we still route e.g. over `barrier=gate` without any further access tags.

### DIFF
--- a/features/car/access.feature
+++ b/features/car/access.feature
@@ -221,6 +221,7 @@ Feature: Car - Restricted access
             | highway | toll        | bothw |
             | primary | yes         |       |
 
+
     Scenario: Car - directional access tags
         Then routability should be
             | highway | access | access:forward | access:backward | forw | backw |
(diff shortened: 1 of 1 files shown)


features/car/access.feature
```diff
@@ -220,3 +220,15 @@ Feature: Car - Restricted access
         Then routability should be
             | highway | toll        | bothw |
             | primary | yes         |       |
+
+
+     Scenario: Car - barrier=gate should be routed over unless explicitely forbidden
+        Then routability should be
+            | node/barrier | access     | bothw |
+            | gate         |            | x     |
```
daniel-j-h: This is the questionable default.

emiltin:
well. i've seen barrier=gate more than once in places where access is indeed possible, like https://www.openstreetmap.org/edit?node=1227080468#map=19/55.69242/12.55155

features/car/access.feature
MoKob: So shouldn't it be the other way around? If you want to add the test, it should be as a `todo` and have the correct default value. Otherwise it might imply to someone that we want it to be this way.

TheMarex:
This needs a rebase, current conflicts in master.

features/car/access.feature
MoKob: 👍 misunderstood the `questionable default` part here. Thought you where referring to what you added as `strange`.

PR #3451, opened 2016-12-15
Update mason to v0.3.0 and change update script

TheMarex:
# Issue

Base work for #2687 to get v8 packages.

## Tasklist
 - [ ] review
 - [ ] adjust for comments


daniel-j-h:
v8 packaging requires a new mason release?
Also what's the idea behind the changes in the dependency script?

TheMarex:
> v8 packaging requires a new mason release?

Yes because we only pull in tagged releases of dependencies and mason keeps its package registry in the repository itself. Adding any package means a version bump.

>  Also what's the idea behind the changes in the dependency script?

This enables us to update dependencies selectively and recover from "oh this merge went wrong, let me just remove the third_party/mason folder" more easily.

PR #3453, opened 2016-12-15
Adding the location validity access for lua scripting.

fijemax:
## Tasklist
 - [x] Add location validity status


daniel-j-h:
:+1: makes sense for having conditional profiles based on if there are locations on the way or not.

PR #3454, opened 2016-12-15
 Enables Optional Support for Stacktraces

daniel-j-h:
Header-only stacktrace library which is currently up for review to be
included in Boost. First try to include this as optional feature.

At the moment we register segfault and assertion handlers.

- http://lists.boost.org/boost-users/2016/12/86979.php
- https://github.com/apolukhin/stacktrace
- http://apolukhin.github.io/stacktrace/boost_stacktrace/getting_started.html

Tested on my Linux machine and produces nice stacktraces for both assertions as well as segfaults.

Needs some more testing for macOS (maybe a signal handler and support for Windows, but we mostly don't care there) and on our infrastructure.

## Tasklist
 - [ ] review
 - [ ] adjust for comments

## Requirements / Relations

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -24,6 +24,7 @@ option(ENABLE_SANITIZER "Use memory sanitizer for Debug build" OFF)
 option(ENABLE_LTO "Use LTO if available" ON)
 option(ENABLE_FUZZING "Fuzz testing using LLVM's libFuzzer" OFF)
 option(ENABLE_GOLD_LINKER "Use GNU gold linker if available" ON)
+option(ENABLE_STACKTRACE "Enable verbose stacktraces for assertions and crashes, requires libdl" OFF)
 
 if(ENABLE_MASON)
   # versions in use
@@ -525,6 +526,22 @@ else()
 
 endif()
 
+if(ENABLE_STACKTRACE)
+  if(WIN32)
+    # For Windows support we have to write a signal handler (see util/stacktrace.hpp) and link with DbgHelp
+    message(FATAL_ERROR "Building with stacktrace support is not implemented on Windows")
+  endif()
+
+  if (Boost_VERSION VERSION_LESS 105600)
+    message(FATAL_ERROR "Building with stacktrace support requires Boost 1.56 or newer")
+  endif()
+
+  add_definitions(-DOSRM_HAS_STACKTRACE)
+  include_directories(SYSTEM ${CMAKE_CURRENT_SOURCE_DIR}/third_party/stacktrace/include)
+  set(MAYBE_STACKTRACE_LIBRARIES -ldl)
+  message(STATUS "Enabling stacktraces for assertions and crashes")
+endif()
+
 # prefix compilation with ccache by default if available and on clang or gcc
 if(ENABLE_CCACHE AND (${CMAKE_CXX_COMPILER_ID} STREQUAL "Clang" OR ${CMAKE_CXX_COMPILER_ID} STREQUAL "GNU"))
   find_program(CCACHE_FOUND ccache)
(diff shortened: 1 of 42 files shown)


CMakeLists.txt
```diff
@@ -166,6 +168,13 @@ if(ENABLE_GOLD_LINKER)
     endif()
 endif()
 
+if(ENABLE_STACKTRACE)
+  # For Windows support we have to write a signal handler (see util/stacktrace.hpp) and link with DbgHelp
+  add_definitions(-DOSRM_HAS_STACKTRACE)
+  set(MAYBE_STACKTRACE_LIBRARIES -ldl)
```
oxidase: `-ldl` also needed in tools targets, or  added to `BOOST_BASE_LIBRARIES`

src/extractor/scripting_environment_lua.cpp
```diff
@@ -21,6 +21,7 @@
 #include <memory>
 #include <sstream>
 
+
```
oxidase: leftover?

include/util/stacktrace.hpp
```diff
@@ -0,0 +1,54 @@
...
+            std::cerr << "Backtrace:\n" << boost::stacktrace::stacktrace() << "\n";
+
+        std::cerr << std::endl;
+        std::exit(EXIT_FAILURE);
+    };
+
+    if (::sigaction(SIGSEGV, &sa, nullptr) == -1)
+        throw std::system_error{errno, std::system_category()};
+
+    if (::sigaction(SIGABRT, &sa, nullptr) == -1)
```
oxidase: SIGBUS also interesting, but it should not be visible without comands that requires strict alignments like  AVX

include/util/stacktrace.hpp
```diff
@@ -0,0 +1,54 @@
...
+inline void InstallStacktraceHandler()
+{
+    struct ::sigaction sa;
+    sa.sa_flags = SA_SIGINFO;
+    ::sigemptyset(&sa.sa_mask);
+
+    sa.sa_sigaction = [](int, ::siginfo_t *si, void *) {
+        boost::stacktrace::stacktrace bt;
+
+        std::cerr << "Signal at address " << si->si_addr << " from " << si->si_call_addr << ".\n";
```
oxidase: Somehow i get here `Signal at address 0 from 0.`  for a dummy SIGSEGV, but don't know why at the moment.
Also printing `si_signo` will be helpful.

daniel-j-h:
Yeah, line info is not in the binary. Unless you compile a `RelWithDebInfo` CMake build then you will get

> osrm::extractor::Extractor::run(osrm::extractor::ScriptingEnvironment&) at /tmp/osrm-backend/src/extractor/extractor.cpp:117

for a manual `::raise(SIGSEGV);` there.

CMakeLists.txt
```diff
@@ -166,6 +168,13 @@ if(ENABLE_GOLD_LINKER)
     endif()
 endif()
 
+if(ENABLE_STACKTRACE)
```
daniel-j-h: :+1:

include/util/stacktrace.hpp
daniel-j-h: Hm I get

> Signal at address 0x3e80000297e from 0x3e80000297e.

both with release as well as debug build.

I will add signo :+1:

daniel-j-h:
What's your take on this? I quickly glues this together as a experimental proof of concept. Should we include it or completely drop this approach? Would love to get some opinions.

oxidase:
@daniel-j-h I find the idea to show a callstack nice, but find using internal tools is somewhat limited. In development phase gdb-like debugging still will be needed. In staging/production phase `/proc/sys/kernel/core_pattern` approach is more powerful. Two main points from my point of view:
* core file can be piped into a script using '|'  symbol (similar to https://github.com/mapbox/logbt/blob/master/bin/logbt#L170 but without waiting). So any kind of processing is possible, not only printing
* gdb can show stack traces of all threads with frame arguments and local variables

I would suggest not to duplicate `logbt` or similar functionality in the OSRM codebase.

PR #3456, opened 2016-12-15
Lets the user specify us dropping hints in response, resolves #1789.

### LINKED ISSUE #1789: Add option to disable hint_data output.

daniel-j-h:
For #1789.

Adds an `?emit_hints=false` option which lets us skip generating and
emitting hints for Waypoints. This can be used to decrease the response
size when the user does not need hints anyway.

We should think about making `false` the default in v6, and let advanced
users be explicit about the hint feature.

## Tasklist
 - [x] figure out a good name for this option, then document in http.md api spec
 - [x] review
 - [x] adjust for comments

cc @TheMarex

### DIFF
--- a/include/server/api/base_parameters_grammar.hpp
+++ b/include/server/api/base_parameters_grammar.hpp
@@ -141,11 +141,18 @@ struct BaseParametersGrammar : boost::spirit::qi::grammar<Iterator, Signature>
                          add_hint, qi::_r1, qi::_1)] %
                          ';';
 
+        generate_hints_rule =
+            qi::lit("generate_hints=") >
+            qi::bool_[ph::bind(&engine::api::BaseParameters::generate_hints, qi::_r1) = qi::_1];
+
         bearings_rule =
             qi::lit("bearings=") >
             (-(qi::short_ > ',' > qi::short_))[ph::bind(add_bearing, qi::_r1, qi::_1)] % ';';
 
-        base_rule = radiuses_rule(qi::_r1) | hints_rule(qi::_r1) | bearings_rule(qi::_r1);
+        base_rule = radiuses_rule(qi::_r1)   //
+                    | hints_rule(qi::_r1)    //
+                    | bearings_rule(qi::_r1) //
+                    | generate_hints_rule(qi::_r1);
     }
 
   protected:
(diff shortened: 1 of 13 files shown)


TheMarex:
>  figure out a good name for this option, then document in http.md api spec

`generate_hints` maybe?

> We should think about making false the default in v6, and let advanced
users be explicit about the hint feature.

:+1: Can you ticket this and put it in the v6 API changes ticket?

emiltin:
what about simply hints=false?

daniel-j-h:
`hints` is already taken - it's the place where users can send hints previously received.

Check the shiny new docs for the `hints` parameter:
http://project-osrm.org/docs/v5.5.0/api/#general-options

include/server/api/base_parameters_grammar.hpp
```diff
@@ -141,11 +141,18 @@ struct BaseParametersGrammar : boost::spirit::qi::grammar<Iterator, Signature>
...
+        emit_hints_rule =
+            qi::lit("emit_hints=") >
+            qi::bool_[ph::bind(&engine::api::BaseParameters::emit_hints, qi::_r1) = qi::_1];
+
         bearings_rule =
             qi::lit("bearings=") >
             (-(qi::short_ > ',' > qi::short_))[ph::bind(add_bearing, qi::_r1, qi::_1)] % ';';
 
-        base_rule = radiuses_rule(qi::_r1) | hints_rule(qi::_r1) | bearings_rule(qi::_r1);
+        base_rule = radiuses_rule(qi::_r1)   //
```
oxidase: I have recently found `// clang-format off` `// clang-format on` and used here https://github.com/Project-OSRM/osrm-backend/blob/opening_hours/include/util/opening_hours.hpp#L302

Would it be ok to allow suppress clang format for qi rules  and use qi-usual formating like
```
    emit_hints_rule
        = qi::lit("emit_hints=")
        > qi::bool_[ph::bind(&engine::api::BaseParameters::emit_hints, qi::_r1) = qi::_1]
        ;
```
?

daniel-j-h: Ah yeah makes sense probably not only for Spirit grammars but e.g. also for program option parser and other similar code. Want to ticket this or make a separate pull request for it?

daniel-j-h:
Adjusted to comments - @oxidase and @themarex could you have a final look again.

TheMarex:
Flagging here: This will need an adaption in `node-osrm`.

TheMarex:
There is one test still failing https://travis-ci.org/Project-OSRM/osrm-backend/jobs/184548957#L5796

daniel-j-h:
Ah sorry, missed updating the expected parsing failure for the `emit_hints` -> `generate_hints` name change. Updating and then merging this.

Issue #3265, opened 2016-11-08
Avoid copying string data in data facades


daniel-j-h:
Our data facades provide means to access [name, ref, pronunciation and destination](https://github.com/Project-OSRM/osrm-backend/blob/ef2261661c91405902abf55db4765818e956b674/include/engine/datafacade/datafacade_base.hpp#L162-L168) string data.

The fact that we're using `std::string` in the facade API here is unfortunate:

internally we store string data as a huge chunk of memory anyway. [Here](https://github.com/Project-OSRM/osrm-backend/blob/ef2261661c91405902abf55db4765818e956b674/include/engine/datafacade/internal_datafacade.hpp#L650-L695) we have to construct a `std::string` out of the correct sub-part of that chunk of memory, copying the char data into a string.

What we really want is a view onto that huge chunk of memory, which should then behave as a string.

Enter [string_ref](http://www.boost.org/doc/libs/1_61_0/libs/utility/doc/html/string_ref.html).

`string_ref` is a non-owning view on top of an arbitrary chunk of memory for representing string data without the need for allocations.

The proposed change here would be to switch out all `std::string` usages in the facades and replace it with non-owning `string_ref` views into our internal chunk of memory. Then only at the `libosrm` boundary we have to create `std::string`s.

(The same should be possible for other types in the facade API - this ticket is only about strings for now)

daniel-j-h:
cc @danpat since you're re-working the facades currently.

oxidase:
@daniel-j-h string_ref is deprecated. It was removed once and reverted back in https://github.com/boostorg/utility/commit/0b492bee9cca261066846216f0ba36602e2216b8 
it is better to use https://github.com/boostorg/utility/blob/develop/include/boost/utility/string_view.hpp and than switch directly to std::string_view with c++17

daniel-j-h:
`string_view` was introduced in mid 2015 from what I can tell.
Which means not even Boost 1.55 has it yet :/

oxidase:
yeah, it will require Boost 1.61

Issue #3458, opened 2016-12-15
ShMemIterator should be RandomAccessIterator


daniel-j-h:
The ShM wrapper has either a vector or a shared mem block as underlying storage. It makes no sense for its `begin()` and `end()` functions to return an `InputIterator` which you can't use for random access, and can't e.g. subtract from another.

https://github.com/Project-OSRM/osrm-backend/blob/9832825abd4f55403ae3fcaa006ad3f9b503687a/include/util/shared_memory_vector_wrapper.hpp#L21-L47

@danpat you played with ripping the wrapper out once and for all, what was the progress there? As it is now, the iterator interface is completely useless and even dangerous: e.g. `std::distance` dispatches on the iterator tag and would loop over all elements for counting: O(n) instead of O(1).

danpat:
I played with `boost::iterator_range` which seems to work as a pretty-much drop-in replacement for *most* cases where we've got our `shared_memory_vector_wrapper` right now.  However, I pulled that https://github.com/Project-OSRM/osrm-backend/pull/3165 in order to not clutter the change with too many things.

We should re-visit this - using `boost::iterator_range` was trivial in most cases.  There were a few sticking points, but the bulk was mostly a search/replace exercise.

danpat:
FWIW: `boost::array_ref` is *not* what we want here, the API there seems to be intended for a different use case, and is missing copy-constructors that we need.

Issue #2859, opened 2016-09-05
Make integer_range a model of the range concept


daniel-j-h:
Currently

```
boost::copy(irange(0,2), out);
```

does not work since the `integer_range` returned by `irange` does not seem to satisfy the requirements to be considered a range. Investigate.

Note: Boost 1.55's `irange` has an issue with unsigneds so we can't just swap it in. See:

https://github.com/Project-OSRM/osrm-backend/blob/7a523713c7e0cc7c4822419ba41d3b7191abf49c/include/util/integer_range.hpp

daniel-j-h:
Relevant Thread: http://lists.boost.org/boost-users/2016/09/86532.php

daniel-j-h:
This landed in https://github.com/Project-OSRM/osrm-backend/issues/3507. @MoKob if you could do a pass over the code - you probably know where we could have used algorithms but were blocked by this.

PR #3459, opened 2016-12-15
Implements Zero-Copy String Views onto Contiguous Memory.

daniel-j-h:
For https://github.com/Project-OSRM/osrm-backend/issues/3265. Work in progress. This is harder than I thought, mostly because call sites often mutate the string which would mean we have to make a copy.

Also cleans up some type aliases on the way.

## Tasklist
 - [x] re-write call sites using string views
 - [x] rip out old-style string interface
 - [x] review
 - [x] adjust for comments

References:

- http://www.boost.org/doc/libs/1_61_0/libs/utility/doc/html/string_ref.html
- http://en.cppreference.com/w/cpp/string/basic_string_view

### DIFF
--- a/include/engine/datafacade/contiguous_internalmem_datafacade.hpp
+++ b/include/engine/datafacade/contiguous_internalmem_datafacade.hpp
@@ -76,7 +76,7 @@ class ContiguousInternalMemoryDataFacade : public BaseDataFacade
     util::ShM<util::Coordinate, true>::vector m_coordinate_list;
     util::PackedVector<OSMNodeID, true> m_osmnodeid_list;
     util::ShM<GeometryID, true>::vector m_via_geometry_list;
-    util::ShM<unsigned, true>::vector m_name_ID_list;
+    util::ShM<NameID, true>::vector m_name_ID_list;
     util::ShM<LaneDataID, true>::vector m_lane_data_id;
     util::ShM<extractor::guidance::TurnInstruction, true>::vector m_turn_instruction_list;
     util::ShM<extractor::TravelMode, true>::vector m_travel_mode_list;
@@ -89,7 +89,7 @@ class ContiguousInternalMemoryDataFacade : public BaseDataFacade
     util::ShM<EdgeWeight, true>::vector m_geometry_fwd_weight_list;
     util::ShM<EdgeWeight, true>::vector m_geometry_rev_weight_list;
     util::ShM<bool, true>::vector m_is_core_node;
-    util::ShM<uint8_t, true>::vector m_datasource_list;
+    util::ShM<DatasourceID, true>::vector m_datasource_list;
     util::ShM<std::uint32_t, true>::vector m_lane_description_offsets;
     util::ShM<extractor::guidance::TurnLaneType::Mask, true>::vector m_lane_description_masks;
 
@@ -231,8 +231,8 @@ class ContiguousInternalMemoryDataFacade : public BaseDataFacade
         m_turn_instruction_list = std::move(turn_instruction_list);
 
         const auto name_id_list_ptr =
-            data_layout.GetBlockPtr<unsigned>(memory_block, storage::DataLayout::NAME_ID_LIST);
-        util::ShM<unsigned, true>::vector name_id_list(
+            data_layout.GetBlockPtr<NameID>(memory_block, storage::DataLayout::NAME_ID_LIST);
+        util::ShM<NameID, true>::vector name_id_list(
             name_id_list_ptr, data_layout.num_entries[storage::DataLayout::NAME_ID_LIST]);
         m_name_ID_list = std::move(name_id_list);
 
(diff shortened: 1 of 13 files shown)


include/engine/datafacade/contiguous_internalmem_datafacade_base.hpp
```diff
@@ -729,7 +729,30 @@ class ContiguousInternalMemoryDataFacadeBase : public BaseDataFacade
...
+        if (range.begin() == range.end())
+        {
+            return {};
+        }
+
+        auto first = m_names_char_list.begin() + range.front();
+        auto last = m_names_char_list.begin() + range.back() + 1;
+        // These iterators are useless: they're InputIterators onto a contiguous block of memory.
+        // Deref to get to the first element, then Addressof to get the memory address of the it.
+        auto len = &*last - &*first;
```
daniel-j-h: Ticketed in https://github.com/Project-OSRM/osrm-backend/issues/3458

include/util/integer_range.hpp
```diff
@@ -39,6 +39,7 @@ template <typename Integer> class range
 
     // Iterator functions
     bool operator!=(const range &) const noexcept { return iter < last; }
+    bool operator==(const range &other) const noexcept { return !operator!=(other); }
```
daniel-j-h: Ticketed a long time ago: https://github.com/Project-OSRM/osrm-backend/issues/2859 this is biting us down the road..

daniel-j-h:
Status: call sites needs to be adapted. Some of them mutate the strings, some of them make comparisons with / lookups against sets of strings - this all needs to be re-written. Can be done incrementally now.

TheMarex:
> Some of them mutate the strings

Where would that be? Can you ticket those location, that could be code smell.

daniel-j-h:
For example

https://github.com/Project-OSRM/osrm-backend/blob/b1f6797aab453e141931af2389b8e16687e47fd6/include/util/guidance/name_announcements.hpp#L29-L40

lower case prefix and suffix, creating new strings out of the strings we get from the underlying storage.

daniel-j-h:
I had some time to revisit this - implemented all string view features, there's one call site left which needs to be adapted: the [mutations here](https://github.com/Project-OSRM/osrm-backend/blob/521492b3eb696e7284486a7f9d42e480ad3ce6ec/include/util/guidance/name_announcements.hpp#L36-L37) and up the call chain. The logic needs to be changed to work on immutable string views only. Squashed and rebased onto master.

TheMarex:
@daniel-j-h could we fix this for the moment by doing a copy to a string inside the function? If the algorithm is adjusted to work on read-only data we can remove this without changing the callsite.

daniel-j-h:
@TheMarex rebased against master and updated to recent changes. I ripped out all old string datafacade functions now for good. Two issues I'd like to flag here:
- mutation in the suffix function, making a copy of the strings
- in order to fully benefit from zero-copy we should expose string views to the outside (api)

include/engine/datafacade/contiguous_internalmem_datafacade.hpp
```diff
@@ -856,19 +859,19 @@ class ContiguousInternalMemoryDataFacade : public BaseDataFacade
...
-        std::copy(m_datasource_name_data.begin() + m_datasource_name_offsets[datasource_name_id],
-                  m_datasource_name_data.begin() + m_datasource_name_offsets[datasource_name_id] +
-                      m_datasource_name_lengths[datasource_name_id],
-                  std::back_inserter(result));
+        auto first = m_datasource_name_data.begin() + m_datasource_name_offsets[id];
+        auto last = m_datasource_name_data.begin() + m_datasource_name_offsets[id] +
+                    m_datasource_name_lengths[id];
+        // These iterators are useless: they're InputIterators onto a contiguous block of memory.
+        // Deref to get to the first element, then Addressof to get the memory address of the it.
+        const std::size_t len = &*last - &*first;
```
TheMarex: Why not use `std::distance`?

include/util/string_view.hpp
```diff
@@ -0,0 +1,34 @@
...
+
+} // namespace util
+} // namespace osrm
+
+// Specializing hash<> for user-defined type in std namespace, this is standard conforming.
+namespace std
+{
+template <> struct hash<::osrm::util::StringView> final
+{
+    std::size_t operator()(::osrm::util::StringView v) const noexcept
```
TheMarex: Why is `v` taken by value here? Is there a downside to `const&`?

src/extractor/suffix_table.cpp
```diff
@@ -11,14 +13,23 @@ namespace extractor
...
+    auto to_view = [](const auto &s) { return util::StringView{s}; };
+    std::transform(begin(suffixes), end(suffixes), into, to_view);
 }
 
 bool SuffixTable::isSuffix(const std::string &possible_suffix) const
 {
+    return isSuffix(util::StringView{possible_suffix});
+}
+
+bool SuffixTable::isSuffix(util::StringView possible_suffix) const
```
TheMarex: Same here: StringView passed by value.

include/engine/datafacade/contiguous_internalmem_datafacade.hpp
```diff
@@ -716,56 +716,59 @@ class ContiguousInternalMemoryDataFacade : public BaseDataFacade
...
     unsigned GetCheckSum() const override final { return m_check_sum; }
 
-    unsigned GetNameIndexFromEdgeID(const unsigned id) const override final
+    NameID GetNameIndexFromEdgeID(const EdgeID id) const override final
     {
         return m_name_ID_list.at(id);
     }
 
-    std::string GetNameForID(const unsigned name_id) const override final
+    virtual StringView GetNameForID(const NameID id) const override final
```
TheMarex: Most functions here omit the `virtual` since it is not necessary.

daniel-j-h: Yeah will adapt it to be consistent with the rest

include/engine/datafacade/contiguous_internalmem_datafacade.hpp
daniel-j-h: Because these are pointers. Always.
Also `operator-` is guaranteed to work for random access iterators anyway.

include/util/string_view.hpp
daniel-j-h: `StringView` is basically `struct StringView { const char* first; const char* last; }` - better to pass by value. The `StringView` is immutable by default, too, so there is no real benefit of passing it by `const`.

(If you're asking yourself, why immutable by default, consider `StringView s{"hello"}; s[0] = 'w'; // oops`)

TheMarex: Thanks for the clarification. 👍

PR #3460, opened 2016-12-15
Makes Units in Profiles Explicit

daniel-j-h:
Here's the deal. Lua is untyped. But at least we can be explicit about our units.

Why? Because time was once in deci-seconds. And now is in seconds.
And it's all very confusing and un-checked if you get things wrong.

@emiltin with my limited Lua knowledge this is the only way I could think of pulling it off. Is there a way to "fake types" so to speak? Maybe even checked via asserts?

It would be great to be explicit at least for time, speeds, etc. - not sure if it makes sense to have a `Key()`, `Value` and `Tag` "type" for our sets.

### DIFF
--- a/profiles/car.lua
+++ b/profiles/car.lua
@@ -6,6 +6,7 @@ local get_turn_lanes = require("lib/guidance").get_turn_lanes
 local Set = require('lib/set')
 local Sequence = require('lib/sequence')
 local Directional = require('lib/directional')
+local Units = require('lib/units')
 
 -- Begin of globals
 barrier_whitelist = Set {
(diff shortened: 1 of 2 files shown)


emiltin:
hmm. another way would to indicate the type would be:

properties.u_turn_penalty_seconds   = 20

daniel-j-h:
Yeah sure. But this isn't checked, too and requires us to change all internal vars.

emiltin:
right. but if we keep the vars as they are we're really just setting an integer, in which case i don't see a way of checking the type/unit.

daniel-j-h:
What we could do for variables passing the Lua/C++ bridge: on the C++ side add typed wrapper, think

```c++
struct Seconds { int64_t value; };
struct KmH { int64_t value; };
```

then bind those transparently to Lua such you have to

```lua
profiles.max_speed_match = KmH(180);
```

emiltin:
looks good

daniel-j-h:
Closing as not really actionable..

PR #3462, opened 2016-12-16
Fix changing shared memory in multi-process setup

### LINKED ISSUE #3455: Error while running a lot of osrm-routed processes with frequent data updates
Test setup:
- 4 different datasets prepared with a core factor of 0.8
- 20 osrm-routed instances connected to shared memory with two threads each
- continuously sending batches of ca. 10 long-ish running requests
- Every second cycle in a new dataset

Observation:
```
[error] Error while attempting to allocate shared memory: No such file or directory, code 7
[warn] [server error] code: No such file or directory (at include/storage/shared_memory.hpp:257)
```

This is blocking using `cluster` with our node bindings. Probable cause is a race condition in the data update part that I missed during my rewrite.

/cc  @danpat @miccolis

TheMarex:
# Issue

Fixes #3455.

This change fixes two bugs:

1. A dead-lock that occurs between osrm-datastore and libosrm when an
   old dataset is free during a data update. This happened because the
   mutexes where acquired in a different order.

2. A region is deleted eventhough it is still in use. This happens when
   libosrm gets overtaken by osrm-datastore, so the new dataset is in
   the same region the old one was.

## Tasklist
 - [ ] review
 - [ ] adjust for comments

### DIFF
--- a/include/engine/datafacade/shared_memory_datafacade.hpp
+++ b/include/engine/datafacade/shared_memory_datafacade.hpp
@@ -38,29 +38,33 @@ class SharedMemoryDataFacade : public ContiguousInternalMemoryDataFacadeBase
     // used anymore.  We crash hard here if something goes wrong (noexcept).
     virtual ~SharedMemoryDataFacade() noexcept
     {
+        // Now check if this is still the newest dataset
+        boost::interprocess::sharable_lock<boost::interprocess::named_upgradable_mutex>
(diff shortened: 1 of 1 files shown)


danpat:
Logic looks good.  Tested this locally by using the script from @oxidase and hammering `osrm-routed` with requests from a bunch of workers.  Couldn't reproduce a hang or deadlock.

LGTM, merging and I'll cherry pick this into the 5.5 branch and release 5.5.1.

Issue #3063, opened 2016-10-14
Rip out Fingerprint logic


TheMarex:
Since we enforce versioning now and the checks are not even enforced over all files we generate, I think we should remove this legacy code.

daniel-j-h:
Context: as a basic check for data file version mismatched there is a fingerprint mechanism in place which works as follows: we generate a hash over the source files (think some .h, .cpp) files and write it as first couple of bytes to the data files we dump out (i.e. *.osrm* files).

Then when reading these files in again, we read in the first couple of bytes (the fingerprint) and make sure they're equal to the current version's fingerprint: this is a simple check to compare two hashes.

Searching for `FingerPrint::GetValid` should reveal all places where we check the fingerprint, e.g. see

https://github.com/Project-OSRM/osrm-backend/blob/8ed6bb8a1b2c21819822da714b81a4f7b72bde58/include/util/graph_loader.hpp#L38-L46

In [CMakeLists.txt](https://github.com/Project-OSRM/osrm-backend/blob/8ed6bb8a1b2c21819822da714b81a4f7b72bde58/CMakeLists.txt#L63-L68) we let CMake generate this fingerprint hashing some .h, .cpp files (check cmake/FingerPrint-Config.cmake).

Here is the custom Fingerprint Type:
- https://github.com/Project-OSRM/osrm-backend/blob/master/include/util/fingerprint.hpp
- https://github.com/Project-OSRM/osrm-backend/blob/master/include/util/fingerprint_impl.hpp.in

TheMarex:
Since we refactored this now into a nicer abstract file handling I think we should really move in 5.6.0 on simplifying this a lot. Currently it is impossible to tell if a change will create a new fingerprint or not. This already bit me during deployments of RC4 to RC5: I knew the file format was compatible but the Fingerprint changed anyway.

This could bite us even more if we release a patch release that changes the fingerprint accidentally.

Since we already employ semantic versioning and encode data file compatibility over that we should do replace the Fingerprint with the version number check (equal major and minor version).

/cc @daniel-j-h @danpat

danpat:
I think there's a path to keep the fingerprint - we just need to be smarter about what's included in its fingerprint calculation.  It's getting easier as we refactor the I/O code.

We could do something like the following:

  1. Copy `src/storage/storage.cpp` to a temp file
  2. Run `gcc -E storage.cpp.tmp` (this works with clang and gcc)
[...]

PR #3467, opened 2016-12-17
Replace fingerprint with semver-based scheme.

danpat:
This change modifies the fingerprint logic to be based on the OSRM version, rather than a generated MD5SUM.  It attempts to address the discussion about file fingerprint handling in https://github.com/Project-OSRM/osrm-backend/issues/3063.

The overview of this change is:

  - the fingerprint is no longer generated from MD5sums, it only uses the `OSRM_VERSION_*` fields
  - fingerprints are considered compatible between minor releases
  - error messages when incompatible files are found are now much more explicit
  - all existing users of the fingerprint validation logic now go through the same checks
  - the fingerprint magic number is now `OSRN` instead of `OSRM`

The fingerprint is now an 8 byte value, made up of:

  The magic number `OSRN` as 4 bytes
  The major version as a byte
  The minor version as a byte
  The patch version as a byte
  An 8-bit checksum of the previous bytes to ensure the header is not damaged.

The version values are the OSRM version values, I am not proposing a separate versioning scheme for the files.

In places where we use it, fingerprint comparison is now configured to fail where:

  The magic number doesn't match
  The major version doesn't match
  The minor version doesn't match
  The checksum is invalid.

and a helpful error message is logged explaining what failed.

The patch version is effectively ignored - our versioning rule is that "patch versions are file-format compatible, minor versions are not".  It's still present so that the checksum value is meaningful, and examination of the file data can retrieve the exact OSRM version used to generate the data.

This change only affects *existing* files with fingerprints, but has not been applied to currently-unfingerprinted files.  In the spirit of keeping changes fairly self-contained, I've opened a separate ticket https://github.com/Project-OSRM/osrm-backend/issues/3520 for rolling out fingerprint checks on all files.

## Tasklist
 - [x] Figure out final semver rule handling
 - [x] review
 - [x] adjust for comments


include/util/fingerprint_impl.hpp
```diff
@@ -0,0 +1,105 @@
...
+FingerPrint FingerPrint::GetValid()
+{
+    FingerPrint fingerprint;
+
+    // 4 chars, 'O','S','R','M'
+    fingerprint.magic_number = 1297240911;
+    fingerprint.major_version = static_cast<char>(atoi(OSRM_VERSION_MAJOR));
+    fingerprint.minor_version = static_cast<char>(atoi(OSRM_VERSION_MINOR));
+    fingerprint.patch_version = static_cast<char>(atoi(OSRM_VERSION_PATCH));
+    fingerprint.crc8 = Compute_CRC8(reinterpret_cast<const std::uint8_t *>(&fingerprint), 7);
```
daniel-j-h: Hard coding 7 is so prone to errors when someone modifies the members, or even reorders them.
Please change this, e.g. do 

    Bytes bytes:
    bytes.append(magic)
    bytes.append(major)
    ..

    ret = crc(bytes);

include/util/fingerprint_impl.hpp
```diff
@@ -0,0 +1,105 @@
...
+    return IsMajorVersionSAME(other) && other.minor_version == minor_version;
+}
+bool FingerPrint::IsPatchVersionSAME(const FingerPrint &other) const
+{
+    return IsMinorVersionSAME(other) && other.patch_version == patch_version;
+}
+
+bool FingerPrint::IsChecksumValid() const
+{
+    return crc8 == Compute_CRC8(reinterpret_cast<const std::uint8_t *>(this), 7);
```
daniel-j-h: Same here

src/storage/storage.cpp
```diff
@@ -308,7 +308,7 @@ void Storage::PopulateLayout(DataLayout &layout)
     }
 
     {
-        io::FileReader hsgr_file(config.hsgr_data_path, io::FileReader::HasNoFingerprint);
+        io::FileReader hsgr_file(config.hsgr_data_path, io::FileReader::VerifyFingerprint);
```
daniel-j-h: Why the change here and below? is there a fingerprint in the hsgr file or not?

daniel-j-h:
Why do we need the CRC checksum? Errors in writing should fail hard and loud anyway.
We don't do checksums over e.g. all nodes, too.

daniel-j-h:
Travis is hitting some compilation issues.

TheMarex:
> If someone attempts to use files generated with OSRM 5.4.0 with OSRM 5.5.0, what should happen?

We should throw an error.

> If someone attempts to use files generated with OSRM 5.5.0 with OSRM 5.4.0, what should happen?

We should throw an error.

The only thing that we really want to grantee is that if MINOR.MAJOR matches, the data files are compatible.

danpat:
@daniel-j-h I put the CRC is there to verify that the fingerprint is a valid new-style fingerprint, particularly because the magic number is the same as the old-style fingerprint.

But your question makes me think - the magic number should probably change.  It's a defacto "fingerprint schema version", and without changing it, I added some potential ambiguity if the bytes from an old file version just happened to line up.

I'll update it to `OSRN` instead of `OSRM`.  This will mean we can identify new-style fingerprints without the checksum.  I'll keep the padding byte there as `0x00` so that it's a nice round 8 bytes, which should pack nicely in most places.

src/storage/storage.cpp
danpat: Yeah, it does.  If you take a look at the changes I made to `include/storage/serialization.hpp`, you'll see that the fingerprint reading logic was duplicated there.  I removed that copied logic, so now the `.hsgr` loading can use the same fingerprint validation logic that everyone else does.

Slowly but surely removing all the mess.....

unit_tests/util/io.cpp
```diff
@@ -57,16 +60,18 @@ BOOST_AUTO_TEST_CASE(file_too_small)
 
         osrm::util::serializeVector(IO_TOO_SMALL_FILE, v);
 
-        std::ofstream f(IO_TOO_SMALL_FILE);
-        f.seekp(0, std::ios_base::beg);
-        std::uint64_t garbage = 0xDEADBEEFCAFEFACE;
-        f.write(reinterpret_cast<char *>(&garbage), sizeof(garbage));
+        std::fstream f(IO_TOO_SMALL_FILE);
```
danpat: Note - this test was broken.  These changes aren't strictly part of this fingerprint change, but it did expose that this test didn't work, so I've bundled the fix in with this PR.

src/util/fingerprint.cpp
```diff
@@ -1,2 +1,61 @@
...
+    FingerPrint fingerprint;
+
+    // 4 chars, 'O','S','R','N' - note the N instead of M, v1 of the fingerprint
+    // used M, so we add one and use N to indicate the newer fingerprint magic number.
+    // Bump this value if the fingerprint format ever changes.
+    fingerprint.magic_number = {{'O', 'S', 'R', 'N'}};
+    // fingerprint.magic_number[1] = 'S';
+    // fingerprint.magic_number[2] = 'R';
+    // fingerprint.magic_number[3] = 'N';
+    fingerprint.major_version = static_cast<std::uint8_t>(atoi(OSRM_VERSION_MAJOR));
```
oxidase: It is better to fix directly in `include/util/version.hpp.in`
```
#define OSRM_VERSION_MAJOR @OSRM_VERSION_MAJOR@
#define OSRM_VERSION_MINOR @OSRM_VERSION_MINOR@
#define OSRM_VERSION_PATCH @OSRM_VERSION_PATCH@

#define OSRM_VERSION_(A,B,C) "v" #A "." #B "." #C
#define OSRM_VERSION OSRM_VERSION_(OSRM_VERSION_MAJOR, OSRM_VERSION_MINOR, OSRM_VERSION_PATCH)
```
and here avoid str->int->char conversion

src/util/fingerprint.cpp
```diff
@@ -1,2 +1,61 @@
...
+    fingerprint.unused = 0x00;
+
+    return fingerprint;
+}
+
+int FingerPrint::GetMajorVersion() const { return major_version; }
+int FingerPrint::GetMinorVersion() const { return minor_version; }
+int FingerPrint::GetPatchVersion() const { return patch_version; }
+
+bool FingerPrint::IsMagicNumberSame(const FingerPrint &other) const
```
oxidase: I think better to have member operators, like `operator==`, `operator<` and `operator<<` to hide details about magic and versioning.  It will be easy later modify internals to have e.g.  non-numeric release version. 
`operator<` will allow version range checks, although at the moment there is no use case for it.

include/storage/io.hpp
```diff
@@ -119,10 +120,30 @@ class FileReader
...
+
+        // Check magic number
+        if (!valid.IsMagicNumberSame(fingerprint))
+        {
+            util::Log(logERROR) << "Fingerprint magic number does not match in "
+                                << filepath.string();
+            return false;
+        }
+
+        // This check also validates the major version
```
oxidase: This is a non-obvious implementation detail -> better to change a member name

src/util/fingerprint.cpp
danpat: 👍  great idea, done.

src/util/fingerprint.cpp
danpat: I've modified the functions, now there is just `IsValid()` and `IsDataCompatible(other)`.

Adding operators makes sense, but I think we can leave them off until we have a use-case for them.  The `IsDataCompatible()` is really all the `FingerPrint` class needs to do right now.

include/storage/io.hpp
danpat: Yeah, you're right.  I've dropped this approach, there is now just a single `IsDataCompatible()` function, which hopefully has a clearer intent.

danpat:
@oxidase I've adjusted based on your comments and rebased.  Let me know if you think it needs anything else.  If not, I'll squash and merge it tomorrow.

oxidase:
LGTM! I pushed a commit to have an additional macro level indirection -> should fix failing version tests. Also added a small polish -> implicit casts of numerical literals should be ok in that place.

Issue #3452, opened 2016-12-15
Problem since luabind replace by Sol2 


fijemax:
Since that luabind was replaced by Sol2 I can't iter on way's nodes.

My lua code : 

```
  print("way id : ", way:id());
  for node in way:get_nodes() do
   print("  node id  : ", node:id())
  end
```

The error exception : 
```
way id : 	5203906
TBB Warning: Exact exception propagation is requested by application but the linked library is built without support for it
terminate called after throwing an instance of 'tbb::captured_exception'
  what():  lua: error: profile.lua:677: attempt to call a userdata value
Abandon (core dumped)
```

daniel-j-h:
Thanks,

@karenzshea can you have a look? This could be related to the notes [here](http://sol2.readthedocs.io/en/latest/api/containers.html).

@fijemax can you tell us the Lua version you are using? 5.1 by chance?

fijemax:
@daniel-j-h 
Lua 5.1.5  Copyright (C) 1994-2012 Lua.org, PUC-Rio

fijemax:
I can solved the problem by updating lua ?

daniel-j-h:
Maybe; you can give Lua 5.2 a try.

But that's still an issue for us since we support both Lua 5.1 _and_ Lua 5.2.

Lua 5.3 has some garbage collection issues which make it consume way more memory during extraction.

danpat:
I can reproduce this, and unfortunately, simply switching to Lua 5.2 doesn't fix it.  Needs some more digging, but it looks like we're not binding everything we need to with Sol2.

oxidase:
@fijemax in your example `for node in way:get_nodes() do` will not work with #3468 but `for _, node in ipairs(way:get_nodes()) do` should be ok. I don't see a way to make an iterator with a state in sol.

fijemax:
Hi,
I tested with `for _, node in ipairs(way:get_nodes()) do`
this don't work no more.

oxidase:
@fijemax the fix is not yet merged to `master` but located  the branch `sol-metafunctions`.
It must be clarified first how to deal with iterators for user defined types in sol2.

ThePhD:
[...]

PR #3468, opened 2016-12-17
Explicitly copy way.nodes() into vector to have proper table meta functions

oxidase:
# Issue

For #3452 added index, length and call metafunctions to WayNodeList, so it is possible to  access  node references in lua as 
```
  local nodes = way:get_nodes()
  for index = 1, #nodes do
     print (nodes[index]:id())
  end
```
or 
```
  for i,node in ipairs(way:get_nodes()()) do
     print (i, node:id())
  end
```

The return value of `call` is a table, so `for node in way:get_nodes() do` does not work. It should be checked what `call` should return to have an "iterable" object.

EDIT: the original commit defines explicitly meta functions for `WayNodeList`.  Also `__pairs` and `__ipairs` functions can not be explicitly defined for a usertype. Just copying `NodeRef` to vector solves the issue by using implicitly  `container_usertype_metatable`.

## Tasklist
 - [x] review
 - [x] adjust for comments

### DIFF
--- a/features/options/extract/lua.feature
+++ b/features/options/extract/lua.feature
@@ -1,0 +1,28 @@
+@extract
+Feature: osrm-extract lua ways:get_nodes()
+
+    Background:
+        Given the node map
+            """
+            a b
+            """
+        And the ways
+            | nodes |
+            | ab    |
+        And the data has been saved to disk
+
+    Scenario: osrm-extract - Passing base file
+        Given the profile file "testbot" extended with
+        """
+        function way_function(way, result)
+          for _, node in ipairs(way:get_nodes()) do
+            print('node id ' .. node:id())
+          end
+          result.forward_mode = mode.driving
+          result.forward_speed = 1
+        end
+        """
+        When I run "osrm-extract --profile {profile_file} {osm_file}"
+        Then it should exit successfully
+        And stdout should contain "node id 1"
+        And stdout should contain "node id 2"
--- a/src/extractor/scripting_environment_lua.cpp
+++ b/src/extractor/scripting_environment_lua.cpp
@@ -74,8 +74,6 @@ template <class T> double lonToDouble(T const &object)
     return static_cast<double>(util::toFloating(object.lon));
 }
 
-auto get_nodes_for_way(const osmium::Way &way) -> decltype(way.nodes()) { return way.nodes(); }
-
 Sol2ScriptingEnvironment::Sol2ScriptingEnvironment(const std::string &file_name)
     : file_name(file_name)
 {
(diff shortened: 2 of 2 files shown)


daniel-j-h:
This looks like a good workaround to let users still do what they want to.

Still we break profiles here by not providing a 100% replacement with Luabind .. hrm.

oxidase:
I guess to have the old behavior `WayNodeList` should return bogus `IteratorX`
```
        sol::meta_function::call,
        [](const osmium::WayNodeList &list) { return IteratorX{list.begin(), list.end()}; }
```
but i did not find to make working. https://github.com/Project-OSRM/osrm-backend/blob/master/third_party/sol2/sol2/sol.hpp#L11334 is how it works with sol `__pairs`/`__ipairs`

TheMarex:
Reading the issues around this, I think only supporting stateful iterators for Lua 5.2+ is fine. There is a workaround for old lua versions, and really we don't want to be slowed down by Lua 5.1 that much.

@oxidase can you rebase and merge?

oxidase:
@TheMarex rebased. But just one note - the change will not allow `for node in way:get_nodes() do` with a stateful iterator. Possible are `for _, node in ipairs(way:get_nodes()) do` with a state-less iterator or `for i = 1, #way:get_nodes() do` with an index.

I simply would rely on sol2 implementation: if the lua `table` interface for `std::vector` will implement `sol::meta_function::call`  than `for node in way:get_nodes() do` will work without changes in OSRM code.

MoKob:
@daniel-j-h @oxidase from the conversation this looks like it is ready to merge, any objections here?

oxidase:
@MoKob PR makes a copy of `NodeRefList` as a `vector<NodeRef>` that is not good. It should be checked first if a correct metatable for `NodeRefList` can be created  automatically.

MoKob:
As long as this check is pending, I am re-tagging this as work in progress to avoid confusion. Feel free to directly move to `ready-to-merge` when checks should turn out fine.

oxidase:
updated PR to wrap a `NodeRefList` const reference by `sol::as_table`. So copying is avoided, but life time of `way:get_nodes()` returned value must be in scope of `way`. 

Although the original issue #3452 is not completely resolved, because the returned `WayNodeList` value has `sol` `table` interface that supports only stateless iterators, i think PR can be merged now.

oxidase:
updated PR due to windows build fail: `decltype(way.nodes())` is resolved by sol to `sol.osmium::WayNodeList *` lua type that has no table metafunctions. With in-place lambda type is correctly resolved.

Issue #3236, opened 2016-11-03
Encapsulate Functionality in RouteStep


daniel-j-h:
Member functions for `RouteStep`: at the moment our types are dumb structs, functionality is provided all over the place where it is needed (e.g. see post processing .h/.cpp).

First candidate to refactor this is `RouteStep`: think adding `invalidate`, `elongate`, lane handling etc. to `RouteStep`.

There should be no need for the call site having to know in which of our current three toolkit headers lane support for a `RouteStep` is implemented in. Or is it implemented in postprocessing.h/.cpp? Or lane-pocessing.h/.cpp?

This is just confusion and making it hard to get into the code base.

MoKob:
This was implemented in #3474. Further functionality can be added.

PR #3474, opened 2016-12-20
Refactor/guidance/route step

MoKob:
# Issue

https://github.com/Project-OSRM/osrm-backend/issues/3236

## Tasklist
 - [x] review
 - [x] adjust for comments

For ease of review, the different items are kept in separate commits. @daniel-j-h do you have other items in mind with as well for #3236 so far? I couldn't see other items that operate plainly on a route step and are not tightly coupled to collapsing (which I feel would best be handled when refactoring post-processing).

## Requirements / Relations
 none

### DIFF
--- a/include/engine/guidance/route_step.hpp
+++ b/include/engine/guidance/route_step.hpp
@@ -30,7 +30,7 @@ namespace guidance
 // Arrive: a --> b --> t. The segment (b,t) is already covered by the previous segment.
 
 // A representation of intermediate intersections
-struct Intersection
+struct IntermediateIntersection
 {
     static const constexpr std::size_t NO_INDEX = std::numeric_limits<std::size_t>::max();
     util::Coordinate location;
@@ -44,13 +44,13 @@ struct Intersection
     extractor::guidance::TurnLaneDescription lane_description;
 };
 
-inline Intersection getInvalidIntersection()
+inline IntermediateIntersection getInvalidIntersection()
 {
     return {util::Coordinate{util::FloatLongitude{0.0}, util::FloatLatitude{0.0}},
             {},
             {},
-            Intersection::NO_INDEX,
-            Intersection::NO_INDEX,
+            IntermediateIntersection::NO_INDEX,
+            IntermediateIntersection::NO_INDEX,
             util::guidance::LaneTuple(),
             {}};
 }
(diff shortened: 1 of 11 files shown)


include/engine/guidance/route_step.hpp
```diff
@@ -30,7 +30,7 @@ namespace guidance
 // Arrive: a --> b --> t. The segment (b,t) is already covered by the previous segment.
 
 // A representation of intermediate intersections
-struct Intersection
+struct IntermediateIntersection
```
daniel-j-h: Thanks for this clarification! :+1:

include/engine/guidance/route_step.hpp
```diff
@@ -71,55 +71,117 @@ struct RouteStep
...
+        mode = TRAVEL_MODE_INACCESSIBLE;
+        maneuver = getInvalidStepManeuver();
+        geometry_begin = 0;
+        geometry_end = 0;
+        intersections.clear();
+        intersections.push_back(getInvalidIntersection());
+    }
+
+    // Elongate by another step in front
+    RouteStep &MakePreceededBy(const RouteStep &preceeding_step)
```
daniel-j-h: Hm the name is a bit confusing to me to be honest.

Especially because this is `Make` prefixed and elongate (or others) below are not. They return a Step& too.

include/engine/guidance/route_step.hpp
```diff
@@ -71,55 +71,117 @@ struct RouteStep
...
+        geometry_end = following_step.geometry_end;
+        intersections.insert(intersections.end(),
+                             following_step.intersections.begin(),
+                             following_step.intersections.end());
+
+        return *this;
+    }
+
+    // Elongate without prior knowledge of in front, or in back.
+    RouteStep &MergeWith(const RouteStep &by_step)
```
daniel-j-h: If we have this functionality, why do we even expose the front / back functions above? Can't we always hide them from the user, preventing misuse?

include/engine/guidance/route_step.hpp
```diff
@@ -71,55 +71,117 @@ struct RouteStep
...
-            TRAVEL_MODE_INACCESSIBLE,
-            getInvalidStepManeuver(),
-            0,
-            0,
-            {getInvalidIntersection()}};
-}
-}
-}
-}
+} // namespace guidance
```
daniel-j-h: In general, what we could do is instead of

```c++
struct S {
  void fn()
  {
    // really long
    // and detailed impl
  }
};
```

do

```c++
struct S {
  // Does X because of Y. Return Z.
  void fn();
};

// Impl.

void S::fn()
{
  // impl.
}
```

src/engine/guidance/lane_processing.cpp
```diff
@@ -166,7 +166,7 @@ std::vector<RouteStep> anticipateLaneChange(std::vector<RouteStep> steps,
             // step as invalid, scheduled for later removal.
             if (collapsable(previous, current))
             {
-                previous = elongate(previous, current);
+                previous.ElongateBy(current);
```
daniel-j-h: All functions above mutate a RouteStep in-place. For example here we also mutate in-place. What's the reason for returning *this by ref? So we can chain calls?

daniel-j-h: It's fine, I just wanted to clarify this :+1:

PR #3476, opened 2016-12-21
Revert default search radius

TheMarex:
# Issue

The changes to the default search radius degraded the performance of map matching on non-car road networks dramatically.

## Tasklist
 - [x] review
 - [ ] adjust for comments


danpat:
👍  @kerrick-lyft the improved car-based matching radius really messes with matching on non-car graphs (i.e. the radius is way too big with busier graphs and performance tanks).

We're going to roll back this change then have a think about how we can re-integrate it without busting other profiles.

kerrick-lyft:
The issue is mainly with networks that are very dense (high number of segments per unit area)? Is there any way to make this configurable in the Lua file and use the new radius formula for car and the old formula for every other profile?

Issue #3475, opened 2016-12-20
Route ending inside Roundabout / Rotary / Circular Junction should not provide Exit Number


daniel-j-h:
For the following request:

```
http 'http://router.project-osrm.org/route/v1/driving/13.451997,52.519655;13.453708,52.518588?steps=true' | jq '.routes[0].legs[].steps'
```

arriving **inside** a roundabout / rotary / circular junction still emits a `exit` property (in this case 3), which then the frontend translates into this ugly instruction, for which the exit part is semantically wrong:

![map](https://cloud.githubusercontent.com/assets/527241/21356726/f372a842-c6d2-11e6-8e3d-7c12521e3a1f.png)

The three instructions are of type `depart`, `rotary`, `arrive`. Here is the maneuver with exit 3 attached:

```json
"maneuver": {
  "exit": 3,
  "type": "arrive",
  "bearing_before": 11,
  "location": [
    13.45371,
    52.518587
  ],
  "bearing_after": 0
},
```

The v5 spec sazs the following:

> roundabout | traverse roundabout, **has additional field exit with NR if the roundabout is left**. the modifier specifies the direction of entering the roundabout

> rotary | a traffic circle. While very similar to a larger version of a roundabout, it does not necessarily follow roundabout rules for right of way. **It can offer rotary_name/rotary_pronunciation in addition to the exit parameter**.

http://project-osrm.org/docs/v5.5.1/api/#stepmaneuver-object

@MoKob what's your take on this? You said we should reset the exit field? Why do we even emit it when arriving inside the roundabout / rotary / circular junction?

MoKob:
@daniel-j-h this is due to the `enter + exit` -> `roundabout` processing. When we find the instruction, we necessarily don't know about the exit. So I believe somewhere in the processing chain there could still be a way to output an exit here (which is useless).
We should definitely not emit it, which needs to be fixed in roundabout-postprocessing (weren't you going to take that on anyhow?)

daniel-j-h:
Good to know you also think we should remove the exits in this case. I can only find https://github.com/Project-OSRM/osrm-backend/issues/3108 for roundabout pre-processing - is this the ticket you mean? And yes I want to have a look at roundabouts anyway. There are some other things I wanted to get done before, such as the string view thingy.

PR #3478, opened 2016-12-21
remove exit tag from arrive step

MoKob:
# Issue

resolves https://github.com/Project-OSRM/osrm-backend/issues/3475. It turns out that the front-end tries to make more out of invalid data we give out than there is. The issue here was that when we end in a roundabout, we don't emit any exit number on the roundabout instruction itself.
Due to the internal counting, there was a left-over `exit` on the arrive step.

Cutting a test for this is kind of difficult, I'd propose simply to assert that `arrive`/`depart` are trivial (in this case don't sport an `exit` field) and notice it this way, if we introduce a regression.

## Tasklist
 ~~- [ ] add regression / cucumber cases (see docs/testing.md)~~
 - [x] review
 - [x] adjust for comments

## Requirements / Relations

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,8 @@
+# 5.5.2
+  - Changes from 5.5.1
+    - Bugfixes
+      - Fix #3475 removed an invalid `exit` field from the `arrive` maneuver
+
 # 5.5.1
   - Changes from 5.5.0
     - API:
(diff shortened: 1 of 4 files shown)


daniel-j-h:
Do we still emit the `exit: 0` in this situation then?

MoKob:
No. The exit tag in steps is only given out when its unequal to zero. We should probably refactor this to an optional at some point.

MoKob:
@daniel-j-h given the rarity, I don't think it's a pressing bug.

Issue #3477, opened 2016-12-21
`osrm-datastore` should release locks on crashes, if possible


danpat:
We encountered a situation where `osrm-datastore` crashed due to an out-of-memory condition.

Because `osrm-datastore` uses interprocess locks (`boost::interprocess::named_sharable_mutex`), these persist after the process dies.  If we crash without releasing the lock, it hangs around, causing later invocations believing a process is still running, leading to an incorrect report of `A data update is in progress`.

The `osrm-datastore` tool should try really hard to clean up on exit.  We should probably go as far as:

  - http://www.cplusplus.com/reference/exception/set_terminate/
  - adding signal handlers

Basically, we should always unlock, for everything except `SIGKILL`.

TheMarex:
This was fixed 🎉

PR #3479, opened 2016-12-21
Make a hard reset of named barrier mutexes on signal

oxidase:
# Issue

For #3477 added a simple signal handler to cleanup the shared barrier mutexes. It will not prevent stale locks after OOM kills. `set_terminate` is not really needed, because SIGABRT will be sent to the process by default `terminate` handler.

## Tasklist
 - [x] review
 - [x] adjust for comments

## Requirements / Relations

### DIFF
--- a/src/tools/store.cpp
+++ b/src/tools/store.cpp
@@ -1,3 +1,4 @@
+#include "storage/shared_barriers.hpp"
 #include "storage/storage.hpp"
 #include "util/exception.hpp"
 #include "util/log.hpp"
@@ -6,6 +7,9 @@
 
 #include <boost/filesystem.hpp>
 #include <boost/program_options.hpp>
+
+#include <csignal>
+#include <cstdlib>
 
 using namespace osrm;
 
@@ -87,8 +91,20 @@ bool generateDataStoreOptions(const int argc,
     return true;
 }
 
+[[ noreturn ]] void CleanupSharedBarriers(int signum)
+{ // Here the lock state of named mutexes is unknown, make a hard cleanup
+    osrm::storage::SharedBarriers::resetCurrentRegions();
+    std::_Exit(128 + signum);
+}
+
 int main(const int argc, const char *argv[]) try
 {
+    int signals[] = {SIGTERM, SIGSEGV, SIGINT, SIGILL, SIGABRT, SIGFPE};
+    for (auto sig : signals)
+    {
+        std::signal(sig, CleanupSharedBarriers);
+    }
+
     util::LogPolicy::GetInstance().Unmute();
 
     boost::filesystem::path base_path;


daniel-j-h:
The overall approach looks good - some question re. signals, as I don't grasp the edge cases around them fully:
- What happens if a signal comes in while our lock-release signal handler is running?
- `signal` is conforming to the C standard so _should_ be portable - can we test this on macOS and Windows before merging
- Should the signal handler be `[[noreturn]]` `noexcept`?
- Should the signal handler call `_Exit` to shut down hard? (\*)

(\*)



https://github.com/adishavit/Terminators

oxidase:
- it is implementation dependent: signal handler can reset to DFL or block other signals, the best would be to use `sigaction` where behavior can be tuned, but it is platform-dependent. 
- yes, `signal` is defined in §18.10
- noreturn  with _exit is ok, noexcept is also ok, but here https://github.com/boostorg/interprocess/blob/develop/include/boost/interprocess/sync/named_sharable_mutex.hpp#L140 just a comment :)
- good point! i will add "_exit(128 + signum);" and noreturn

src/tools/store.cpp
```diff
@@ -87,8 +90,22 @@ bool generateDataStoreOptions(const int argc,
     return true;
 }
 
+[[ noreturn ]] void CleanupSharedBarriers(int signum)
+{ // Here the lock state of named mutexes is unknown, make a hard cleanup
+    osrm::storage::SharedBarriers::resetCurrentRegions();
+    osrm::storage::SharedBarriers::resetRegions1();
+    osrm::storage::SharedBarriers::resetRegions2();
+    _exit(128 + signum);
```
oxidase: :+1: updated

daniel-j-h:
Looks good - let's wait for Appveyor / Travis and then this should be ready to merge. Would maybe love to test this in action, at least n macOS. Maybe @danpat can you try to force this behavior.

src/tools/store.cpp
```diff
@@ -87,8 +91,22 @@ bool generateDataStoreOptions(const int argc,
     return true;
 }
 
+[[ noreturn ]] void CleanupSharedBarriers(int signum)
+{ // Here the lock state of named mutexes is unknown, make a hard cleanup
+    osrm::storage::SharedBarriers::resetCurrentRegions();
+    osrm::storage::SharedBarriers::resetRegions1();
```
TheMarex: This is potentially dangerous:

These locks govern the shared mutexes for the data region. They should only be reset if osrm-datastore actually had a lock on them.

If these locks are just removed while a libosrm instance is running this can lead to crashes on the next data update or exceptions because locks will try to unlock mutexes that are in an unlocked state.

oxidase: right, this is dangerous, but reaching this point with a custom or default handler is a bad sign for the whole IPC because locks will be in unknown state. I will remove reseting regions, so it can be controlled by `--max-wait` and reset at https://github.com/Project-OSRM/osrm-backend/blob/master/src/storage/storage.cpp#L158-L164

Issue #3480, opened 2016-12-21
Support more platforms with ENABLE_MASON=ON


springmeyer:
OS X >= 10.10 and Ubuntu >= 12.04 are the two main platforms that ENABLE_MASON=ON builds work well on and are well tested ([per wiki documentation](https://github.com/Project-OSRM/osrm-backend/issues/3207#issuecomment-268555489)).

This ticket is to track adding support for more platforms. This is low urgency, but nice to have.

 - [ ] Systems with libstdc++ older than 5.x and no ability to upgrade will hit https://github.com/Project-OSRM/osrm-backend/issues/3207#issuecomment-258020247. (Will be unblocked by https://github.com/mapbox/mason/issues/267)
 - [ ] Arch linux only works currently by [turning off LTO and Gold linking](https://github.com/Project-OSRM/osrm-backend/pull/3119#issuecomment-256790635). The task is to figure out a workaround to enable LTO and ENABLE_MASON on arch.

springmeyer:
> This is low urgency, but nice to have.

Therefore I'm considering this ticket as not actionable and not valuable open. Closing since there is no active work planned to address ancient libstdc++ or time spent to debug arch linux binutils+clang+lto issues (which might be solvable but don't have an obvious solution).

PR #3481, opened 2016-12-21
Remove the LTO warning for ENABLE_MASON builds

springmeyer:
This PR removes a warning that is only known to be helpful for arch users that enable LTO (#3480) and warns unnecessary for ubuntu versions built per the wiki (https://github.com/Project-OSRM/osrm-backend/wiki/Building-on-Ubuntu). I think the noise of this warning is just not worth it, so this removes it.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -272,11 +272,6 @@ if(CMAKE_BUILD_TYPE MATCHES Release OR CMAKE_BUILD_TYPE MATCHES MinRelSize OR CM
   endif()
 endif()
 
-if(UNIX AND NOT APPLE AND ENABLE_MASON AND (LTO_WORKS OR ENABLE_GOLD_LINKER))
-  message(WARNING "ENABLE_MASON and ENABLE_LTO/ENABLE_GOLD_LINKER may not work on all linux systems currently")
-  message(WARNING "For more details see: https://github.com/Project-OSRM/osrm-backend/issues/3202")
-endif()
-
 set(MAYBE_COVERAGE_LIBRARIES "")
 if (ENABLE_COVERAGE)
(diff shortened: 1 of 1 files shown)


daniel-j-h:
Arch Linux is the only platform _we know about_ not compiling with mason and lto because @TheMarex is running it, right? Why not leave the warning in then, until we fix it for all systems?

TheMarex:
When setting on source compilation on an Ubuntu 14.04 box using the ubuntu testing toolchain and clang-3.8 from the llvm PPA I had similar problems.

I can try to reproduce this in a docker container, but there is definitely something weird going on with different binuitils/clang/libstdc++ combinations.

daniel-j-h:
With `docker run -it ubuntu:16.04 /bin/bash` and following the Mason Wiki docs the Clang lto plugin included with mason (3.9.0) segfaults by default - you have to disable LTO manually: https://github.com/Project-OSRM/osrm-backend/issues/3501#issuecomment-269992386

springmeyer:
> the Clang lto plugin included with mason (3.9.0) segfaults by default

This is a clang bug and not an issue with OSRM or mason as noted at https://github.com/Project-OSRM/osrm-backend/issues/3501#issuecomment-270508561.

My preference would be to disable LTO by default, then keep this warning in place. Thoughts?

springmeyer:
> When setting on source compilation on an Ubuntu 14.04 box using the ubuntu testing toolchain and clang-3.8 from the llvm PPA I had similar problems.

@TheMarex please ticket when you have more details.

daniel-j-h:
LTO is off by default with https://github.com/Project-OSRM/osrm-backend/issues/3524 - let's keep this warning. Users will only see it if they enable lto and are using mason.

PR #3485, opened 2016-12-21
Don't retain SharedDataFacade between queries

TheMarex:
# Issue

This addresses an issue we discovered while using shared memory: Segments only really get deleted when every process has detached.

This has the downside that we will re-create the SharedDataFacade often. I see a lot of syscalls going out with this change and I think the "cache hit rate" seems fairly low.

If anyone has some ideas around optimizations that would be great.

## Tasklist
 - [x] review
 - [ ] adjust for comments

/cc @oxidase @danpat for some 👀

### DIFF
--- a/include/engine/data_watchdog.hpp
+++ b/include/engine/data_watchdog.hpp
@@ -22,11 +22,6 @@ namespace engine
 // This class monitors the shared memory region that contains the pointers to
 // the data and layout regions that should be used. This region is updated
 // once a new dataset arrives.
-//
-// TODO: This also needs a shared memory reader lock with other clients and
-// possibly osrm-datastore since updating the CURRENT_REGIONS data is not atomic.
-// Currently we enfore this by waiting that all queries have finished before
-// osrm-datastore writes to this section.
 class DataWatchdog
 {
   public:
@@ -57,7 +52,9 @@ class DataWatchdog
         const auto shared_timestamp =
             static_cast<const storage::SharedDataTimestamp *>(shared_regions->Ptr());
 
-        const auto get_locked_facade = [this, shared_timestamp]() {
+
+        const auto get_locked_facade = [this, shared_timestamp](
+            const std::shared_ptr<datafacade::SharedMemoryDataFacade> &facade) {
             if (current_timestamp.region == storage::REGION_1)
             {
                 return std::make_pair(RegionsLock(shared_barriers->region_1_mutex), facade);
@@ -76,8 +73,11 @@ class DataWatchdog
 
             if (shared_timestamp->timestamp == current_timestamp.timestamp)
             {
-                BOOST_ASSERT(shared_timestamp->region == current_timestamp.region);
-                return get_locked_facade();
+                if (auto facade = cached_facade.lock())
+                {
+                    BOOST_ASSERT(shared_timestamp->region == current_timestamp.region);
+                    return get_locked_facade(facade);
+                }
             }
         }
 
@@ -89,20 +89,25 @@ class DataWatchdog
         // in that case we don't modify anything
         if (shared_timestamp->timestamp == current_timestamp.timestamp)
         {
-            BOOST_ASSERT(shared_timestamp->region == current_timestamp.region);
-
-            return get_locked_facade();
+            if (auto facade = cached_facade.lock())
+            {
+                BOOST_ASSERT(shared_timestamp->region == current_timestamp.region);
+                return get_locked_facade(facade);
+            }
         }
 
         // this thread has won and can update the data
         boost::upgrade_to_unique_lock<boost::upgrade_mutex> unique_facade_lock(facade_lock);
 
         current_timestamp = *shared_timestamp;
-        facade = std::make_shared<datafacade::SharedMemoryDataFacade>(shared_barriers,
-                                                                      current_timestamp.region,
-                                                                      current_timestamp.timestamp);
 
-        return get_locked_facade();
+        auto new_facade =
+            std::make_shared<datafacade::SharedMemoryDataFacade>(shared_barriers,
+                                                                 current_timestamp.region,
+                                                                 current_timestamp.timestamp);
+        cached_facade = new_facade;
+
+        return get_locked_facade(new_facade);
     }
 
   private:
@@ -114,7 +119,7 @@ class DataWatchdog
     std::unique_ptr<storage::SharedMemory> shared_regions;
 
     mutable boost::shared_mutex facade_mutex;
-    std::shared_ptr<datafacade::SharedMemoryDataFacade> facade;
+    std::weak_ptr<datafacade::SharedMemoryDataFacade> cached_facade;
     storage::SharedDataTimestamp current_timestamp;
 };
 }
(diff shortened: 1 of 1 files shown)


oxidase:
Checked with `strace` syscalls for 100 requests in master
```
% time     seconds  usecs/call     calls    errors syscall
------ ----------- ----------- --------- --------- ----------------
 66.13    0.075017          45      1652       143 futex
 33.83    0.038377         127       302           epoll_wait
  0.03    0.000031           2        17           read
  0.01    0.000012           0       221           write
  0.00    0.000000           0        41        17 open
  0.00    0.000000           0       125           close
  0.00    0.000000           0       121         3 stat
  0.00    0.000000           0        30           fstat
  0.00    0.000000           0         1           lseek
  0.00    0.000000           0        53           mmap
  0.00    0.000000           0        63           mprotect
  0.00    0.000000           0         7           munmap
  0.00    0.000000           0         3           brk
  0.00    0.000000           0         2           rt_sigaction
  0.00    0.000000           0         4           rt_sigprocmask
  0.00    0.000000           0       529           ioctl
  0.00    0.000000           0        10        10 access
  0.00    0.000000           0       477           madvise
  0.00    0.000000           0         6           shmget
  0.00    0.000000           0         3           shmat
  0.00    0.000000           0         9           shmctl
  0.00    0.000000           0         3           socket
  0.00    0.000000           0         1         1 connect
  0.00    0.000000           0       100           accept
  0.00    0.000000           0         1           sendto
  0.00    0.000000           0       100           sendmsg
  0.00    0.000000           0       103           recvmsg
  0.00    0.000000           0       100           shutdown
  0.00    0.000000           0         2           bind
  0.00    0.000000           0         1           listen
  0.00    0.000000           0         2           getsockname
  0.00    0.000000           0       100           getpeername
  0.00    0.000000           0         2           setsockopt
  0.00    0.000000           0         5           clone
  0.00    0.000000           0         1           execve
  0.00    0.000000           0         1           getrlimit
  0.00    0.000000           0         1           statfs
  0.00    0.000000           0         1         1 mlockall
  0.00    0.000000           0         1           arch_prctl
  0.00    0.000000           0         1           set_tid_address
  0.00    0.000000           0       304           epoll_ctl
  0.00    0.000000           0         6           set_robust_list
  0.00    0.000000           0         1           timerfd_create
  0.00    0.000000           0         1           eventfd2
  0.00    0.000000           0         1           epoll_create1
------ ----------- ----------- --------- --------- ----------------
100.00    0.113437                  4515       175 total
```
and branch
```
% time     seconds  usecs/call     calls    errors syscall
------ ----------- ----------- --------- --------- ----------------
 66.98    0.118288          72      1639       131 futex
 32.99    0.058270         192       304           epoll_wait
  0.02    0.000034           0       140        17 open
  0.01    0.000012           0      1408         3 stat
  0.00    0.000000           0        17           read
  0.00    0.000000           0       613           write
  0.00    0.000000           0       225           close
  0.00    0.000000           0       129           fstat
  0.00    0.000000           0         1           lseek
  0.00    0.000000           0       152           mmap
  0.00    0.000000           0        73           mprotect
  0.00    0.000000           0       105           munmap
  0.00    0.000000           0         3           brk
  0.00    0.000000           0         2           rt_sigaction
  0.00    0.000000           0         4           rt_sigprocmask
  0.00    0.000000           0      1321           ioctl
  0.00    0.000000           0        10        10 access
  0.00    0.000000           0       523           madvise
  0.00    0.000000           0       402           shmget
  0.00    0.000000           0       201           shmat
  0.00    0.000000           0       603           shmctl
  0.00    0.000000           0         3           socket
  0.00    0.000000           0         1         1 connect
  0.00    0.000000           0       100           accept
  0.00    0.000000           0         1           sendto
  0.00    0.000000           0       100           sendmsg
  0.00    0.000000           0       103           recvmsg
  0.00    0.000000           0       100           shutdown
  0.00    0.000000           0         2           bind
  0.00    0.000000           0         1           listen
  0.00    0.000000           0         2           getsockname
  0.00    0.000000           0       100           getpeername
  0.00    0.000000           0         2           setsockopt
  0.00    0.000000           0         5           clone
  0.00    0.000000           0         1           execve
  0.00    0.000000           0       200           shmdt
  0.00    0.000000           0         1           getrlimit
  0.00    0.000000           0         1           statfs
  0.00    0.000000           0         1         1 mlockall
  0.00    0.000000           0         1           arch_prctl
  0.00    0.000000           0         1           set_tid_address
  0.00    0.000000           0       304           epoll_ctl
  0.00    0.000000           0         6           set_robust_list
  0.00    0.000000           0         1           timerfd_create
  0.00    0.000000           0         1           eventfd2
  0.00    0.000000           0         1           epoll_create1
------ ----------- ----------- --------- --------- ----------------
100.00    0.176604                  8914       163 total
```
For sequential requests chaching is not working, but now shmdt is called 200 times, the rest of increase is also shmem and ipc related.

include/engine/data_watchdog.hpp
```diff
@@ -92,22 +91,27 @@ class DataWatchdog
...
 
-            return get_locked_facade();
+                return get_locked_facade(facade);
+            }
         }
 
         // this thread has won and can update the data
         boost::upgrade_to_unique_lock<boost::upgrade_mutex> unique_facade_lock(facade_lock);
 
         current_timestamp = *shared_timestamp;
```
oxidase: Should be check `if (shared_timestamp->timestamp == current_timestamp.timestamp)` also added here before `current_timestamp` assignment? This will prevent situation when two threads will try acquire a unique lock, and the second will reload facade with the same data

TheMarex:
@danpat @oxidase how do you both feel about shipping this as a bug fix? (maybe 5.5.3)

This of course means decreased performance, but it might be better then relying on hacks to work around this problem.

I have some ideas how we could optimize this solution a little bit more:
- https://linux.die.net/man/2/shmat allows for specifying a shared memory address: If we keep these fixed for the same dataset re-computing all the addresses for our SharedDataFacade internal datastructures in unnecessary.
- we can also cache the `key_t` that goes into `shmat` which would require a syscall to compute
- ideally we would only need to two these things for each query:
  - acquire a reader lock
  - call shmat
  - call shmdet
  - release reader lock

oxidase:
@TheMarex for me looks good as a bug fix, now stale memory blocks can appear only if a request hangs, but it will be a general problem with timeouts. About optimization: I think the third point is the most important -> ~16 `futex` calls and 1.8ms per request - it is too much. But it is not related to the change: in comparison `master` vs `branch` number of `futex` syscalls is approximately the same.

oxidase:
Another possible optimization is to use one memory block for the layout header and data. It will reduce by factor 2 number of shmat/shmdt calls. It will require populate layout at https://github.com/Project-OSRM/osrm-backend/blob/master/src/storage/storage.cpp#L192 into local memory, allocate shared memory for sizeof(DataLayout)+GetSizeOfLayout() bytes, copy local layout to shared memory (just 888 bytes) and modify computation of shared_memory_ptr as shmem_ptr+sizeof(DataLayout).

TheMarex:
@oxidase that is a good idea. Going to ticket. Needing to always attach two shared memory regions is kind of weird anyway.

MoKob:
I've rebased this onto current master to resolve conflicts. @oxidase, since you have had a look before. Can you double check if my rebase was correct and click this nice litte rebase-merge button if you agree with the rebase? Lets clean up some of the PRs that are waiting for merge.

Issue #3483, opened 2016-12-21
Fixing potential integer overflow reports from `-fsanitize=integer`


springmeyer:
Running `-fsanitize=integer` with clang++ 3.8.1 on OS X flags a few potential spots where we may be overflowing unexpectedly:

- [x] Deallocating Vector

```bash
include/util/deallocating_vector.hpp:150:42: runtime error: unsigned integer overflow: 0 - 5299 cannot be represented in type 'unsigned long'
include/util/deallocating_vector.hpp:136:55: runtime error: unsigned integer overflow: 1647 + 18446744073709551615 cannot be represented in type 'unsigned long'
include/util/deallocating_vector.hpp:101:42: runtime error: unsigned integer overflow: 0 - 13087 cannot be represented in type 'unsigned long'
include/util/deallocating_vector.hpp:87:55: runtime error: unsigned integer overflow: 227 + 18446744073709551615 cannot be represented in type 'unsigned long'
```

- [x] CRC32

```bash
include/contractor/crc32_processor.hpp:69:17: runtime error: unsigned integer overflow: 0 - 1 cannot be represented in type 'unsigned int'
include/contractor/crc32_processor.hpp:78:17: runtime error: unsigned integer overflow: 0 - 1 cannot be represented in type 'unsigned int'
```

- [x] Shared Datatype

```bash
include/storage/shared_datatype.hpp:141:54: runtime error: negation of 4 cannot be represented in type 'std::size_t' (aka 'unsigned long')
```

- [x] Packed Vector

```bash
include/util/packed_vector.hpp:48:49: runtime error: unsigned integer overflow: 2112 - 2145 cannot be represented in type 'unsigned long'
```

- [x] Coordinate Calculation

```bash
src/util/coordinate_calculation.cpp:27:25: runtime error: unsigned integer overflow: 18446744073709530423 * 18446744073709530423 cannot be represented in type 'unsigned long long'
src/util/coordinate_calculation.cpp:27:15: runtime error: unsigned integer overflow: 18446744073709549253 * 18446744073709549253 cannot be represented in type 'unsigned long long'
src/util/coordinate_calculation.cpp:27:15: runtime error: unsigned integer overflow: 18446744073709451616 * 18446744073709451616 cannot be represented in type 'unsigned long long'
src/util/coordinate_calculation.cpp:27:25: runtime error: unsigned integer overflow: 18446744073709540491 * 18446744073709540491 cannot be represented in type 'unsigned long long'
```

[...]

PR #3486, opened 2016-12-22
Makes Types in Squared Dist Calculation Explicit, see #3483.

daniel-j-h:
For #3483.

This was introduced back in https://github.com/Project-OSRM/osrm-backend/commit/68ee4eab6154847bb89bc2bb73002b2b5bb9e82e. The math works out nicely (and errors there would have been visible e.g. in the rtree unit tests). This changeset makes types explicit, no longer relying on the unsigned wraparound behavior.

cc @springmeyer

### DIFF
--- a/src/util/coordinate_calculation.cpp
+++ b/src/util/coordinate_calculation.cpp
@@ -21,10 +21,15 @@ namespace coordinate_calculation
 // Does not project the coordinates!
 std::uint64_t squaredEuclideanDistance(const Coordinate lhs, const Coordinate rhs)
 {
-    const std::uint64_t dx = static_cast<std::int32_t>(lhs.lon - rhs.lon);
-    const std::uint64_t dy = static_cast<std::int32_t>(lhs.lat - rhs.lat);
+    std::int64_t d_lon = static_cast<std::int32_t>(lhs.lon - rhs.lon);
+    std::int64_t d_lat = static_cast<std::int32_t>(lhs.lat - rhs.lat);
 
-    return dx * dx + dy * dy;
+    std::int64_t sq_lon = d_lon * d_lon;
+    std::int64_t sq_lat = d_lat * d_lat;
+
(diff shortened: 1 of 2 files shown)


src/util/coordinate_calculation.cpp
```diff
@@ -21,10 +21,15 @@ namespace coordinate_calculation
...
-    const std::uint64_t dx = static_cast<std::int32_t>(lhs.lon - rhs.lon);
-    const std::uint64_t dy = static_cast<std::int32_t>(lhs.lat - rhs.lat);
+    std::int64_t d_lon = static_cast<std::int32_t>(lhs.lon - rhs.lon);
+    std::int64_t d_lat = static_cast<std::int32_t>(lhs.lat - rhs.lat);
 
-    return dx * dx + dy * dy;
+    std::int64_t sq_lon = d_lon * d_lon;
+    std::int64_t sq_lat = d_lat * d_lat;
+
+    std::uint64_t result = sq_lon + sq_lat;
```
oxidase: static_cast<std::uint64_t> ?

src/util/coordinate_calculation.cpp
```diff
@@ -21,10 +21,15 @@ namespace coordinate_calculation
 // Does not project the coordinates!
 std::uint64_t squaredEuclideanDistance(const Coordinate lhs, const Coordinate rhs)
 {
-    const std::uint64_t dx = static_cast<std::int32_t>(lhs.lon - rhs.lon);
-    const std::uint64_t dy = static_cast<std::int32_t>(lhs.lat - rhs.lat);
+    std::int64_t d_lon = static_cast<std::int32_t>(lhs.lon - rhs.lon);
```
TheMarex: Urg this was a stupid bug. Good catch.

daniel-j-h: The math works out due to unsigned wraparound - so functionality wise we're fine.

The change set just makes it more explicit.

Issue #3384, opened 2016-12-01
Version flag for profiles


TheMarex:
Currently when we want to make changes to the profiles that are not backward compatible we would do a major version bump of `osrm-backend`.

However that also means that profiles that were not ported properly will fail (maybe even silently) if they are not adjusted to API changes.

To make this explicit and open up the possibility for supporting multiple profile API version we should add a version flag to `ProfileProperties` so we could load the appropriate version of the scripting context.

daniel-j-h:
Interesting. 

For a start can't we simply expose the OSRM version we set in CMaketo Lua and _also_ set a required version in the Lua profiles? Then the profiles could check their version against the OSRM version currently running.

This would catch profile mismatches and would even allow the profiles to specify required version bounds. Think of profiles specifying `> 5.4 && < 5.5` and the OSRM version bound via Luabind then reporting `5.4.3`.

TheMarex:
Hrm you think that will be less error prone then doing another sequential version like we do for the HTTP API? I was thinking about something like:

```lua
properties.version = 1 -- (default value for everything below 5.5)
```
When we load the scripting context we use that version number to load the appropriate bindings. The default value makes sure that at least every current profile in existence will be backward compatible and needs to be upgrade explicitly once we change the profile API.

We could throw a deprecation warning if your profile does not expose the current version, but still support older versions.

TheMarex:
For the record we settled on a global `api_version = 1` in every profile file.

daniel-j-h:
This landed with https://github.com/Project-OSRM/osrm-backend/pull/3487.

PR #3487, opened 2016-12-22
Profile version

TheMarex:
# Issue

Add explicit versioning to the profiles. Addresses #3384 

This is currently unused but could be utilized to activate other API features only when explicitly bumping the version. This will be used to make #2399 backwards compatible.

## Tasklist
 - [x] review
 - [x] adjust for comments

## Requirements / Relations

Requirement for #2399

### DIFF
--- a/profiles/car.lua
+++ b/profiles/car.lua
@@ -1,3 +1,5 @@
+api_version = 0
+
 -- Car profile
 local find_access_tag = require("lib/access").find_access_tag
 local get_destination = require("lib/destination").get_destination
--- a/src/extractor/scripting_environment_lua.cpp
+++ b/src/extractor/scripting_environment_lua.cpp
@@ -306,6 +306,20 @@ void Sol2ScriptingEnvironment::InitContext(LuaScriptingContext &context)
     context.has_node_function = node_function.valid();
     context.has_way_function = way_function.valid();
     context.has_segment_function = segment_function.valid();
+    auto maybe_version = context.state.get<sol::optional<int>>("api_version");
+    if (maybe_version)
+     {
+         context.api_version = *maybe_version;
+     }
+
+    if (context.api_version < SUPPORTED_MIN_API_VERSION ||
+        context.api_version > SUPPORTED_MAX_API_VERSION )
+    {
+        throw util::exception("Invalid profile API version " + std::to_string(context.api_version) +
+                              " only versions from " + std::to_string(SUPPORTED_MIN_API_VERSION) +
+                              " to " + std::to_string(SUPPORTED_MAX_API_VERSION) +
+                              " are supported." + SOURCE_REF);
+    }
 }
 
 const ProfileProperties &Sol2ScriptingEnvironment::GetProfileProperties()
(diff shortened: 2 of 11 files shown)


profiles/car.lua
```diff
@@ -194,6 +194,7 @@ properties.max_speed_for_map_matching      = 180/3.6 -- 180kmph -> m/s
 properties.use_turn_restrictions           = true
 properties.continue_straight_at_waypoint   = true
 properties.left_hand_driving               = false
+properties.version                         = 0
```
daniel-j-h: Hrm should we call it `profile_api_version` or something like that? Not sure if version alone is telling for users.

TheMarex: Actually now that you mention it, maybe we should not even put it in `ProfileProperties`.

Do you think a global variable like `profile_api_version = 0` would be too obscure?

TheMarex: My thinking here is that we actually don't need the profile version information outside of the scripting context.

daniel-j-h: Hm maybe at the top of the profile, like CMake does it :+1:

oxidase:
@TheMarex maybe not related to the PR, how to do testing in future? Should tests be replicated for every profile version (actually bad idea), just maintain the actual profile version (no tests for deprecated profile versions) or some tests added to check only profile functions?

TheMarex:
@oxidase that is a good question. I the most feasible solution would be to only have full tests for the current version, but specific tests for functionality changes between versions.

For example in the context of the #2399 PR we should have all our tests run on `api_version = 1` (e.g. the new version). But should have a test that `api_version = 0`-style `turn_function` and `segment_function` still work.

TheMarex:
Moved it to a global variable and enforced correct version over a check.

oxidase:
@TheMarex just [a simple test](https://gist.github.com/oxidase/cfe6287d069007415905b73397fda86c) to make codecov green

TheMarex:
@oxidase the test looks good. 👍 you can just push to the branch, that's fine.

oxidase:
@TheMarex pushed. `Not-defined API version` test expects that osrm-extract is not failing if api-version is not defined. api-version 0 will be used as default initialized.

emiltin:
should version be in the format 1.0.0?

TheMarex:
@emiltin I was think using the same scheme as in the HTTP API, which also uses simple increments. Minor/patch versions are not that useful for single components because you typically only care about breaking changes.

src/extractor/scripting_environment_lua.cpp
```diff
@@ -306,6 +306,20 @@ void Sol2ScriptingEnvironment::InitContext(LuaScriptingContext &context)
     context.has_node_function = node_function.valid();
     context.has_way_function = way_function.valid();
     context.has_segment_function = segment_function.valid();
+    auto maybe_version = context.state.get<sol::optional<int>>("api_version");
+    if (maybe_version)
+     {
```
daniel-j-h: Looks like the formatting is off here ..

Issue #3489, opened 2016-12-22
Only one shared memory region per dataset


TheMarex:
Per https://github.com/Project-OSRM/osrm-backend/pull/3485#issuecomment-268821528 this would reduce the overhead of attaching shared memory regions. There is no real reason why we have two right now.

They are created/destroyed together and semantically belong together.

That would mean the shared memory layout and the data need to be packed in the same memory block.

/cc @oxidase

PR #3490, opened 2016-12-23
Join LAYOUT and DATA shared memory blocks into REGION

oxidase:
# Issue

PR for #3489
Joined LAYOUT and DATA regions and changed to REGION. Variables and enums are renamed to singular `region`. 

## Tasklist
 - [x] review
 - [x] adjust for comments

## Requirements / Relations
Should reduce number of syscalls by factor 2 in #3485

### DIFF
--- a/include/engine/datafacade/shared_memory_datafacade.hpp
+++ b/include/engine/datafacade/shared_memory_datafacade.hpp
@@ -24,10 +24,8 @@ class SharedMemoryDataFacade : public ContiguousInternalMemoryDataFacadeBase
 {
 
   protected:
-    std::unique_ptr<storage::SharedMemory> m_layout_memory;
     std::unique_ptr<storage::SharedMemory> m_large_memory;
     std::shared_ptr<storage::SharedBarriers> shared_barriers;
-    storage::SharedDataType layout_region;
     storage::SharedDataType data_region;
     unsigned shared_timestamp;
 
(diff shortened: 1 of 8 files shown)


include/engine/datafacade/shared_memory_datafacade.hpp
```diff
@@ -40,58 +38,49 @@ class SharedMemoryDataFacade : public ContiguousInternalMemoryDataFacadeBase
...
-        BOOST_ASSERT(storage::SharedMemory::RegionExists(layout_region));
-        m_layout_memory = storage::makeSharedMemory(layout_region);
-
         BOOST_ASSERT(storage::SharedMemory::RegionExists(data_region));
         m_large_memory = storage::makeSharedMemory(data_region);
 
-        InitializeInternalPointers(*reinterpret_cast<storage::DataLayout *>(m_layout_memory->Ptr()),
-                                   reinterpret_cast<char *>(m_large_memory->Ptr()));
+        InitializeInternalPointers(*reinterpret_cast<storage::DataLayout *>(m_large_memory->Ptr()),
+                                   reinterpret_cast<char *>(m_large_memory->Ptr()) + sizeof(storage::DataLayout));
```
danpat: Is the memory alignment still OK here?  I think the answer is yes, as `sizeof(storage::DataLayout)` should include padding, right?

oxidase: Padding depends on packing, but it does not matter here.

Alignment will be adjusted at https://github.com/Project-OSRM/osrm-backend/blob/master/include/storage/shared_datatype.hpp#L145 internally in the data block.
It will work because GetSizeOfLayout pessimistically assumes misaligned data and request ~500 bytes overhead to adjust pointers

PR #3496, opened 2016-12-24
normalize file writes

duizendnegen:
# Issue

Normalizing file input / file writes, to remove ofstream usage across the project.

## Tasklist
 - [x] replace all ofstreams
 - [x] Binarize `*.osrm.datasource_names`
 - [ ] Convert all Read/WriteElementCount32 to its 64-bit variant
 - [x] Check whether all files have now `storage/io.hpp` in their includes
 - [x] review
 - [x] adjust for comments

## Requirements / Relations
https://github.com/Project-OSRM/osrm-backend/issues/2242


daniel-j-h:
Nice, thanks for starting this quest! cc @danpat @ghoshkaj 

And a happy holiday break to you!

duizendnegen:
Same to y'all!

TheMarex:
Nice to see progress here. 👍 Haha, sorry had to change the title marker to a label, just a pet peeve of mine. Happy Holidays!

duizendnegen:
In `static_rtree.hpp` we still have an unnormalized file read (deserialize):

```cpp
m_leaves_region.open(leaf_file);
std::size_t num_leaves = m_leaves_region.size() / sizeof(LeafNode);
m_leaves.reset(reinterpret_cast<const LeafNode *>(m_leaves_region.data()), num_leaves);
```

duizendnegen:
Just a heads up, I'll be away from Jan 2nd until Jan 15th, so this ticket might look stale for some time, I will be returning to it though.

src/contractor/contractor.cpp
```diff
@@ -1065,17 +1045,15 @@ Contractor::WriteContractedGraph(unsigned max_node_id,
...
     // serialize number of edges
-    hsgr_output_stream.write((char *)&contracted_edge_count, sizeof(std::uint64_t));
+	hsgr_output_file.WriteElementCount64(contracted_edge_count);
     // serialize all nodes
     if (node_array_size > 0)
     {
-        hsgr_output_stream.write((char *)&node_array[0],
-                                 sizeof(util::StaticGraph<EdgeData>::NodeArrayEntry) *
-                                     node_array_size);
+		hsgr_output_file.WriteFrom(&node_array, node_array_size);
```
daniel-j-h: This seems to trigger the trivially copyable static assertion on Travis

https://travis-ci.org/Project-OSRM/osrm-backend/jobs/187703561#L2237-L2241

daniel-j-h:
@danpat @ghoshkaj you took over the I/O refactor - care to have a look here and spec out what is left.

duizendnegen:
From my side: there's quite a lot of stuff left still (I haven't been able to touch this since I was away). There's some immediate problems I don't clearly now how to address (e.g. writing nodes in batches, possibly to not consume too much memory, then winding back and writing the amount of nodes written) but I should be able to make more progress in the coming month, at least with some trivial cases.

daniel-j-h:
Started to use the `FileWriter` today for dumping out the compressed nodebased graph for `osrm-partition`, and there is one issue I stumbled upon already: the `WriteElementCount` functions throwing when the `write` call failed. But then they in addition return a `bool` if the stream is good. This is weird because now call sites have to check for both: exceptions and if the call returned true. Maybe something to refactor.

duizendnegen:
Agree - I'll refactor to using `void` instead probably

daniel-j-h:
Referencing https://github.com/Project-OSRM/osrm-backend/issues/3621 here - I did this for writing the compressed node based graph. Having a async. write feature in place and using it would be a great addition. Probably out of scope here - just wanted to cross-reference it.

src/extractor/extraction_containers.cpp
```diff
@@ -191,26 +190,25 @@ void ExtractionContainers::WriteCharData(const std::string &file_name)
...
-    }
+		if (buffer_len >= WRITE_BLOCK_BUFFER_SIZE)
+		{
+			file_out.WriteFrom(write_buffer, WRITE_BLOCK_BUFFER_SIZE);
+			buffer_len = 0;
+		}
+	}
 
-    file_stream.write(write_buffer, buffer_len);
+	file_out.WriteFrom(write_buffer, buffer_len);
```
duizendnegen: Discovered that this wasn't the way stxxl works anyway

duizendnegen:
@daniel-j-h ready for review.
A few notes:
- not _all_ `ofstream`s are replaced. Maybe we can pair program for an hour to judge whether the remaining calls are essential for cross-platform (some of them write GeoJSON text files & don't seem to be passed across `osrm-*` binaries. At least one - on `NameTable::IndexedData` I could use some help with.)
- Converting all Read/WriteElementCount32 to its 64-bit variant is not done yet, I will leave this to a second PR.
- I can't get clang-format to do what I want. I format it locally, send it in, Travis clang rules protest. I'd like to learn how to do this, so I can finish of my own PR.

daniel-j-h:
Make sure you're using clang-format-3.8 - earlier and more recent version seem to be _slightly_ incompatible

    clang-format-3.8 -i -style=file *.cpp *.hpp

(or better check `scripts/format.sh`)

include/util/range_table.hpp
```diff
@@ -139,6 +139,18 @@ template <unsigned BLOCK_SIZE, bool USE_SHARED_MEMORY> class RangeTable
         sum_lengths = lengths_prefix_sum;
     }
 
+    void WriteARangeTable(osrm::storage::io::FileWriter &filewriter)
```
daniel-j-h: Hm the naming here and below are a bit weird. The type is already `RangeTable`, so `Read` or `Write` is good enough.

include/util/range_table.hpp
```diff
@@ -139,6 +139,18 @@ template <unsigned BLOCK_SIZE, bool USE_SHARED_MEMORY> class RangeTable
...
+    {
+        unsigned number_of_blocks = diff_blocks.size();
+
+        filewriter.WriteElementCount32(number_of_blocks);
```
daniel-j-h: :+1: for iteratively removing all those four byte counts - there is no reason for saving four bytes but being prone to overflows and _some_ files having eight byte counts and _some_ having four byte counts.

src/contractor/contractor.cpp
```diff
@@ -713,47 +711,27 @@ Contractor::LoadEdgeExpandedGraph(const ContractorConfig &config,
...
+
+        geometry_file.WriteOne(number_of_indices);
+        geometry_file.WriteFrom(&geometry_indices[0], number_of_indices);
+
+        geometry_file.WriteOne(number_of_compressed_geometries);
+        geometry_file.WriteFrom(&geometry_node_list[0], number_of_compressed_geometries);
+        geometry_file.WriteFrom(&geometry_fwd_weight_list[0], number_of_compressed_geometries);
+        geometry_file.WriteFrom(&geometry_rev_weight_list[0], number_of_compressed_geometries);
+        geometry_file.WriteFrom(&geometry_fwd_duration_list[0], number_of_compressed_geometries);
+        geometry_file.WriteFrom(&geometry_rev_duration_list[0], number_of_compressed_geometries);
```
daniel-j-h: Hm seeing this, I _think_ you can just use `.data()` instead of the weird `&vec[0]` trick.

src/contractor/contractor.cpp
```diff
@@ -967,12 +944,12 @@ void Contractor::WriteCoreNodeMarker(std::vector<bool> &&in_is_core_node) const
...
-                                                          std::ios::binary);
-    unsigned size = unpacked_bool_flags.size();
-    core_marker_output_stream.write((char *)&size, sizeof(unsigned));
-    core_marker_output_stream.write((char *)unpacked_bool_flags.data(),
-                                    sizeof(char) * unpacked_bool_flags.size());
+    storage::io::FileWriter core_marker_output_file(config.core_output_path,
+                                                    storage::io::FileWriter::HasNoFingerprint);
+
+    const auto count = unpacked_bool_flags.size();
+    core_marker_output_file.WriteElementCount32(count);
```
daniel-j-h: Hm count should be a size_t and here we write only a four byte count - as you said, all of these should be eight bytes

src/extractor/edge_based_graph_factory.cpp
```diff
@@ -337,20 +338,18 @@ void EdgeBasedGraphFactory::GenerateEdgeExpandedEdges(
...
+                                          storage::io::FileWriter::HasNoFingerprint);
+
+    storage::io::FileWriter turn_penalties_index_file(turn_penalties_index_filename,
+                                          storage::io::FileWriter::HasNoFingerprint);
 
     // Writes a dummy value at the front that is updated later with the total length
     const std::uint64_t length_prefix_empty_space{0};
-    edge_data_file.write(reinterpret_cast<const char *>(&length_prefix_empty_space),
-                         sizeof(length_prefix_empty_space));
+    edge_data_file.WriteElementCount64(0);
```
daniel-j-h: Hm do we need the length_prefix_empty_space above any longer?

src/extractor/extraction_containers.cpp
```diff
@@ -567,41 +558,30 @@ void ExtractionContainers::WriteEdges(std::ofstream &file_out_stream) const
 
             // IMPORTANT: here, we're using slicing to only write the data from the base
             // class of NodeBasedEdgeWithOSM
-            NodeBasedEdge tmp = edge.result;
-            file_out_stream.write((char *)&tmp, sizeof(NodeBasedEdge));
-            used_edges_counter++;
+            normal_edges.push_back(edge.result);
```
daniel-j-h: Hm can we `.reserve` memory for the vector above?

src/extractor/extraction_containers.cpp
```diff
@@ -567,41 +558,30 @@ void ExtractionContainers::WriteEdges(std::ofstream &file_out_stream) const
...
         }
 
-        if (used_edges_counter > std::numeric_limits<unsigned>::max())
+        if (normal_edges.size() > std::numeric_limits<unsigned>::max())
```
daniel-j-h: Hm the unsigned here really should be a uint32_t

src/extractor/extraction_containers.cpp
duizendnegen: afaik we don't know how large the vector would be because of the conditional. Do you want to reserve for the size of `all_edges_list` and potentially overallocating?

duizendnegen: Alternatively we could use the approach of writing 0 as total number of elements, writing elements one by one, skipping back to the beginning and writing the appropriate count.

daniel-j-h: Yes if can make reasonable guesses :+1:

daniel-j-h: Ah re. second comment - not sure if there are implications wrt. performance, needs to be tested

duizendnegen: Added the `reserve` part. Writing each individual NodeBasedEdge is how it was done before so performance would be as before.

duizendnegen:
@daniel-j-h ready to merge?

duizendnegen:
Sure - could we talk on IRC / videoconf one of the coming days re/ remaining reads/writes?

I'll tag remaining items in #2242

PR #3497, opened 2016-12-26
Make DataWatchdog spawn its own thread and listen for updates

TheMarex:
## Issue 

This makes DataWatchdog spawn its own thread that checks for data
updates. This addresses an issue where libOSRM instances would not
notice a data update and thus never dettach from shared memory regions.

## Tasklist
 - [ ] Investigate behavior when query segfaults
 - [ ] Use timed mutexes in osrm-datastore for aquiring exclusive locks
 - [ ] Review
 - [ ] Adjust for comments

## Requirements / Relations

This is an alternative to #3485

### DIFF
--- a/include/engine/data_watchdog.hpp
+++ b/include/engine/data_watchdog.hpp
@@ -13,6 +13,7 @@
 #include <boost/thread/shared_mutex.hpp>
 
 #include <memory>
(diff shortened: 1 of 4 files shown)


TheMarex:
Replaces by #3528.

PR #3498, opened 2016-12-27
Windows tests

oxidase:
# Issue

library and cucumber tests are not running on AppVeyor. The PR fixes features js support code for win32 and adds checks in appveyor.yml and appveyor-build.bat to run `npm test` and `library-tests`.

## Tasklist
 - [x] review
 - [x] adjust for comments

### DIFF
--- a/features/support/cache.js
+++ b/features/support/cache.js
@@ -141,7 +141,7 @@ module.exports = function() {
 
     // converts the scenario titles in file prefixes
     this.getScenarioID = (scenario) => {
-        let name = scenario.getName().toLowerCase().replace(/[\/\-'=,\(\)]/g, '').replace(/\s/g, '_').replace(/__/g, '_').replace(/\.\./g, '.');
+        let name = scenario.getName().toLowerCase().replace(/[\/\-'=,\(\):\*#]/g, '').replace(/\s/g, '_').replace(/__/g, '_').replace(/\.\./g, '.');
         return util.format('%d_%s', scenario.getLine(), name);
     };
 
--- a/features/support/env.js
+++ b/features/support/env.js
@@ -42,44 +42,33 @@ module.exports = function () {
         this.OSRM_PORT = process.env.OSRM_PORT && parseInt(process.env.OSRM_PORT) || 5000;
         this.HOST = 'http://127.0.0.1:' + this.OSRM_PORT;
 
-        // TODO make sure this works on win
-        if (process.platform.match(/indows.*/)) {
+        if (process.platform.match(/^win.*/)) {
             this.TERMSIGNAL = 9;
             this.EXE = '.exe';
-            this.LIB = '.dll';
-            this.QQ = '"';
         } else {
             this.TERMSIGNAL = 'SIGTERM';
             this.EXE = '';
+        }
 
-            // heuristically detect .so/.a suffix
-            this.LIB = null;
-
+        // heuristically detect .so/.a/.dll/.lib suffix
+        this.LIB = ['lib%s.a', 'lib%s.so', '%s.dll', '%s.lib'].find((format) => {
             try {
-                const dot_a = util.format('%s/libosrm%s', this.BIN_PATH, '.a');
-                fs.accessSync(dot_a, fs.F_OK);
-                this.LIB = '.a';
-            } catch(e) { /*nop*/ }
+                const lib = this.BIN_PATH + '/' + util.format(format, 'osrm');
+                fs.accessSync(lib, fs.F_OK);
+            } catch(e) { return false; }
+            return true;
+        });
 
-            try {
-                const dot_so = util.format('%s/libosrm%s', this.BIN_PATH, '.so');
-                fs.accessSync(dot_so, fs.F_OK);
-                this.LIB = '.so';
-            } catch(e) { /*nop*/ }
-
-            if (!this.LIB) {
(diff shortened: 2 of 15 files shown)


MoKob:
In principle, this looks good to me.
Might be that I don't know enough about some of the syntaxes used, but some of these changes (e.g. like https://github.com/Project-OSRM/osrm-backend/pull/3498/files#diff-80cc062f7a9c64a60e9ea8310db0675aR40) don't look obvious to me why they needed to be done.

Could you maybe go through the PR and add commentary to the code (if possible on a reasonable level) on changes like that? I am referring to tiny changes like the one I linked that seemed to work before but for windows+linux somehow need to be different.

features/support/cache.js
```diff
@@ -141,7 +141,7 @@ module.exports = function() {
 
     // converts the scenario titles in file prefixes
     this.getScenarioID = (scenario) => {
-        let name = scenario.getName().toLowerCase().replace(/[\/\-'=,\(\)]/g, '').replace(/\s/g, '_').replace(/__/g, '_').replace(/\.\./g, '.');
+        let name = scenario.getName().toLowerCase().replace(/[\/\-'=,\(\):\*#]/g, '').replace(/\s/g, '_').replace(/__/g, '_').replace(/\.\./g, '.');
```
oxidase: `:*#` are incorrect symbols in filenames

features/support/env.js
```diff
@@ -42,44 +42,33 @@ module.exports = function () {
         this.OSRM_PORT = process.env.OSRM_PORT && parseInt(process.env.OSRM_PORT) || 5000;
         this.HOST = 'http://127.0.0.1:' + this.OSRM_PORT;
 
-        // TODO make sure this works on win
-        if (process.platform.match(/indows.*/)) {
+        if (process.platform.match(/^win.*/)) {
```
oxidase: `win32` is a platform name https://nodejs.org/api/process.html#process_process_platform

features/support/env.js
```diff
@@ -42,44 +42,33 @@ module.exports = function () {
...
-            this.LIB = null;
-
-            try {
-                const dot_a = util.format('%s/libosrm%s', this.BIN_PATH, '.a');
-                fs.accessSync(dot_a, fs.F_OK);
-                this.LIB = '.a';
-            } catch(e) { /*nop*/ }
-
+        // heuristically detect .so/.a/.dll/.lib suffix
+        this.LIB = ['lib%s.a', 'lib%s.so', '%s.dll', '%s.lib'].find((format) => {
```
oxidase: here instead of library extensions a library platform-dependent format string is used  -> windows libraries have no `lib` prefix

features/support/env.js
```diff
@@ -108,7 +97,7 @@ module.exports = function () {
         var verify = (binPath, cb) => {
             fs.exists(binPath, (exists) => {
                 if (!exists) return cb(new Error(util.format('%s is missing. Build failed?', binPath)));
-                var helpPath = util.format('%s --help > /dev/null 2>&1', binPath);
+                var helpPath = util.format('%s --help', binPath);
```
oxidase: `/dev/null` is an incorrect device, but stdout is simply ignored in `child_process.exec` callback

features/support/run.js
```diff
@@ -36,7 +37,7 @@ module.exports = function () {
     };
 
     this.runBin = (bin, options, env, callback) => {
-        let cmd = util.format('%s%s/%s%s%s', this.QQ, this.BIN_PATH, bin, this.EXE, this.QQ);
+        let cmd = path.resolve(util.format('%s/%s%s', this.BIN_PATH, bin, this.EXE));
```
oxidase: double quoted commands in `execFile` do not work for me on windows. I see the only reason to have double quotes around command is spaces in the path, but in http://stackoverflow.com/questions/16395612/unable-to-execute-child-process-exec-when-path-has-spaces stated that spaces must be escaped with backslashes. I did not check exactly this case, but think that `path.resolve` returns correctly escaped absolute path

package.json
```diff
@@ -25,7 +25,7 @@
   },
   "scripts": {
     "lint": "eslint -c ./.eslintrc features/step_definitions/ features/support/",
-    "test": "npm run lint && ./node_modules/cucumber/bin/cucumber.js features/ -p verify",
+    "test": "npm run lint && node ./node_modules/cucumber/bin/cucumber.js features/ -p verify",
```
oxidase: `node` is needed, because windows ignores shebang `#!/usr/bin/env node`

MoKob:
Great. Thank you for the explanations. Looks good to go for me.

oxidase:
rebased to master. For double quotes and spaces just manually checked:
```
C:\work\osrm test>node
> const path = require('path');
undefined
> const child_process = require('child_process');
undefined
> let cmd = path.resolve(util.format('%s/%s%s', 'build/Release', 'osrm-extract', '.exe'));
undefined
> cmd;
'C:\\work\\osrm test\\build\\Release\\osrm-extract.exe'
> let child = child_process.execFile(cmd, ['--version'], {maxBuffer: 1024 * 1024 * 1000, env: []}, (err, out) => { console.log(err, out);});
undefined
> null 'v5.5.0\r\n'
```
So whitespaces in paths are correctly treated by `child_process.execFile` and `path.resolve`.

PR #3507, opened 2017-01-02
Make osrm::util::range a model of SinglePassRangeConcept

oxidase:
# Issue

The reason of boost concepts checks fails in boost algorithms is a mix of two concepts `SinglePassRangeConcept` and `SinglePassIteratorConcept` in `osrm::utill::range`. Namely, `begin` and `end` must return at least a single pass iterator. Also https://github.com/Project-OSRM/osrm-backend/blob/master/include/util/integer_range.hpp#L41 is a workaround mix of iterators and ranges that works for range-based for loops, but incorrect for algorithms.

The PR is a backport of the fix https://github.com/boostorg/range/commit/9e6bdc13ba94af4e150afae547557a2fbbfe3bf0 and the custom implementation must be removed after switch to Boost >=1.56

## Tasklist
 - [x] review
 - [x] adjust for comments

## Requirements / Relations
- Issue #2859 
- http://www.boost.org/doc/libs/1_63_0/libs/range/doc/html/range/concepts/single_pass_range.html

### DIFF
--- a/include/util/integer_range.hpp
+++ b/include/util/integer_range.hpp
@@ -2,6 +2,7 @@
 #define INTEGER_RANGE_HPP
 
 #include <boost/assert.hpp>
+#include <boost/iterator/iterator_facade.hpp>
 
 #include <type_traits>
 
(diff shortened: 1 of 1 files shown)


include/util/integer_range.hpp
```diff
@@ -12,35 +13,72 @@ namespace osrm
...
 
-    // Iterable functions
-    const range &begin() const noexcept { return *this; }
-    const range &end() const noexcept { return *this; }
-    Integer front() const noexcept { return iter; }
-    Integer back() const noexcept { return last - 1; }
+    range(Integer begin, Integer end) : iter(begin), last(end) {}
+
+    iterator begin() const noexcept { return iter; }
+    iterator end() const noexcept { return last; }
```
daniel-j-h: Do the `const` overloads have to return `const_iterator`s?

daniel-j-h:
cc @MoKob - you probably know best the locations where this was blocking using algorithms in guidance. As soon as this changeset lands we should check and refactor those.

include/util/integer_range.hpp
oxidase: This part corresponds https://github.com/boostorg/range/blob/develop/include/boost/range/iterator_range_core.hpp#L202-L204 and https://github.com/boostorg/range/blob/develop/include/boost/range/iterator_range_core.hpp#L221-L229 so iterator constness is ignored and i just used boost iterators interface.

I think the method `const` qualifier should be treated as a `begin` call does not change the range, so the second `begin` call will return the same value. An  iterator constness will be relevant in the `dereference` method, but `boost::iterator_facade` defines `reference` as `Integer`, so in an integer range no big difference between const and non-const iterators.

PR #3509, opened 2017-01-02
Reworks the Readme Landing Page

daniel-j-h:
Dealing with the fallout of our default installation instruction method (Mason) being broken I not only reworked the Wiki but now also the readme landing page.

Here's a preview: https://github.com/Project-OSRM/osrm-backend/tree/quick-start

Would love some feedback and ideas how to make it friendlier to beginners and more appealing to first-time users.

## Tasklist
 - [x] review
 - [x] adjust for comments

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,64 +1,87 @@
-## About
+## Open Source Routing Machine
 
-The Open Source Routing Machine is a high performance routing engine written in C++11 designed to run on OpenStreetMap data.
+| Linux / macOS | Windows | Code Coverage |
+| ------------- | ------- | ------------- |
+| [![Travis](https://travis-ci.org/Project-OSRM/osrm-backend.png?branch=master)](https://travis-ci.org/Project-OSRM/osrm-backend) | [![AppVeyor](https://ci.appveyor.com/api/projects/status/4iuo3s9gxprmcjjh)](https://ci.appveyor.com/project/DennisOSRM/osrm-backend) | [![Codecov](https://codecov.io/gh/Project-OSRM/osrm-backend/branch/master/graph/badge.svg)](https://codecov.io/gh/Project-OSRM/osrm-backend) |
 
-## Current build status
+High performance routing engine written in C++14 designed to run on OpenStreetMap data.
 
-| build config | status |
-|:-------------|:-------|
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1,64 +1,82 @@
-## About
+## Open Source Routing Machine
 
-The Open Source Routing Machine is a high performance routing engine written in C++11 designed to run on OpenStreetMap data.
+[![Travis](https://travis-ci.org/Project-OSRM/osrm-backend.png?branch=master)](https://travis-ci.org/Project-OSRM/osrm-backend)
```
MoKob: I have to say that some information regarding the systems used should be here. Before we had the table with linux/windows. I wouldn't have been able to tell if the builds are passing on mac, linux, windows based on the buttons alone before I worked here. So I feel having information here would benefit the readability.
What is your take? Is this clear from the buttons alone?

README.md
```diff
@@ -1,64 +1,82 @@
...
-| Coverage     | [![codecov](https://codecov.io/gh/Project-OSRM/osrm-backend/branch/master/graph/badge.svg)](https://codecov.io/gh/Project-OSRM/osrm-backend) |
+The following services are available via HTTP API, C++ library interface and NodeJs wrapper:
+- Nearest - Snaps coordinates to the street network and returns the nearest matches
+- Route - Finds the fastest route between coordinates
+- Table - Computes the duration of the fastest route between all pairs of supplied coordinates
+- Match - Snaps noisy GPS traces to the road network in the most plausible way
+- Trip - Solves the Traveling Salesman Problem using a greedy heuristic
+- Tile - Generates Mapbox Vector Tiles with internal routing metadata
+
+To quickly try OSRM use our [free and daily updated demo server](http://map.project-osrm.org).
```
MoKob: given that that thing hasn't been updated for more than 3 weeks now, I would feel better if we don't claim `daily` here.

README.md
daniel-j-h: I felt like the table was too clunky and putting the system names in front of the banners I was not able to align them properly with markdown. Should be fine with our other projects
- https://github.com/mapbox/mapbox-gl-js/blob/master/README.md
- https://github.com/Turfjs/turf/blob/master/README.md

PR #3510, opened 2017-01-02
Removes Profile Symlink

daniel-j-h:
Removes the `./profiles.lua` symlink from the root dir.


MoKob:
Hm. We should probably think about notifying people specifically about this when we do this. Might be some bash scripts could rely on this being linked to the car by default.

daniel-j-h:
I will add a changelog entry.

PR #3511, opened 2017-01-03
Removes mlock for osrm-routed, closes #3484

### LINKED ISSUE #3484: Remove mlock usage
Currently we use [`mlock`](https://linux.die.net/man/2/mlock) to lock all of our process memory into RAM. This seems sensible except for this part:

> MCL_FUTURE
> Lock all pages which will become mapped into the address space of the process in the future. These could be for instance new pages required by a growing heap and stack as well as **new memory mapped files** or shared memory regions. 

This basically means that for everyone running `osrm-routed` we will lock our external memory storage in RAM. This also explains the delay of about 20 seconds it takes for the first query to be served after `osrm-routed` comes online: It loads the whole StaticRTree into memory.

Note: This only affect `osrm-routed` not `libosrm`. (we do this for `osrm-datastore` as well, maybe we should also overthink that)

/cc @danpat

daniel-j-h:
For https://github.com/Project-OSRM/osrm-backend/issues/3484.

This changeset removes mlock-ing the virtual address space to RAM in osrm-routed.

The same is still present [here](https://github.com/Project-OSRM/osrm-backend/blob/e6ff17ab2ac37518b9da108191d3aad2dce5d1b5/src/storage/storage.cpp#L114-L121) for the datastore.

## Tasklist
 - [x] review
 - [x] adjust for comments

### DIFF
--- a/src/tools/routed.cpp
+++ b/src/tools/routed.cpp
@@ -10,10 +10,6 @@
 #include <boost/filesystem.hpp>
 #include <boost/program_options.hpp>
 
-#ifdef __linux__
-#include <sys/mman.h>
-#endif
-
 #include <cstdlib>
 
 #include <signal.h>
@@ -269,25 +265,6 @@ int main(int argc, const char *argv[]) try
         return EXIT_FAILURE;
     }
 
-#ifdef __linux__
-    struct MemoryLocker final
-    {
-        explicit MemoryLocker(bool should_lock) : should_lock(should_lock)
-        {
-            if (should_lock && -1 == mlockall(MCL_CURRENT | MCL_FUTURE))
-            {
-                could_lock = false;
-                util::Log(logWARNING) << "memory could not be locked to RAM";
-            }
-        }
-        ~MemoryLocker()
-        {
-            if (should_lock && could_lock)
-                (void)munlockall();
-        }
-        bool should_lock = false, could_lock = true;
-    } memory_locker(config.use_shared_memory);
-#endif
     util::Log() << "starting up engines, " << OSRM_VERSION;
 
     if (config.use_shared_memory)
@@ -373,7 +350,7 @@ catch (const std::bad_alloc &e)
     util::Log(logWARNING) << "Please provide more memory or consider using a larger swapfile";
     return EXIT_FAILURE;
 }
-catch (const std::exception& e)
+catch (const std::exception &e)
 {
     std::cerr << "[exception] " << e.what();
     return EXIT_FAILURE;


oxidase:
LGTM, but what the reason to keep https://github.com/Project-OSRM/osrm-backend/blob/e6ff17ab2ac37518b9da108191d3aad2dce5d1b5/src/storage/storage.cpp#L114-L121? This code chunk makes loaded shared memory regions locked and was a partial precondition for hard OOM kills.

daniel-j-h:
I think the reasoning for keeping the mlock for shared memory was to force the data we need for response generation to sit in ram and never to be paged out so we can guarantee for response latencies in production use. For osrm-routed we don't really care. Maybe @TheMarex can clarify here.

oxidase:
Yeah, the PR definitely needs some benchmarking with a "high-load" scenario. I would suggest to wait for @TheMarex and clarify 

> (we do this for osrm-datastore as well, maybe we should also overthink that)

TheMarex:
`osrm-routed` uses `mlock` which (IMHO) is a bug that stems from misunderstanding how it works and interacts with mmap. `libosrm` never used mlock to begin with and this is what we use in production. So this PR is fine. 👍 

`osrm-datastore` uses `mlock` to force the OS to not swap out data to disk before we do the switch. This can be a problem when we do the switch-over: Some queries might hit a page that is on disk an could be slowed down by that.

However I think this was actually never tested if it is necessary for performance to do that. My gut tells me it is very unlikely that the OS will swap out data to disk if it has enough memory available.

Not messing with the OS settings might gain us the ability to have (a little bit) less then 2x the required memory on instances that do frequent data switches. This will of course mean that during the switch-over we will see swapping and page faults, but maybe it will still be fine in terms of SLA and would potentially unlock using smaller instances on AWS.

/cc @miccolis

oxidase:
@TheMarex [from](http://man7.org/linux/man-pages/man2/mlock.2.html) "mlockall locks all pages mapped into the address space of the calling process.", so memory is locked for life-time of `osrm-datastore` that just loads data. Shared memory blocks are locked via `SHM_LOCK` flag by `shmctl`.
I think it is better to remove `mlockall` with `MCL_FUTURE` also from `osrm-datastore` as it can be a source of 

>       If MCL_FUTURE has been specified, then a later system call (e.g.,
       mmap(2), sbrk(2), malloc(3)), may fail if it would cause the number
       of locked bytes to exceed the permitted maximum (see below).  In the
       same circumstances, stack growth may likewise fail: the kernel will
       deny stack expansion and deliver a SIGSEGV signal to the process.

TheMarex:
@oxidase ok, then let's settle on removing mlock everywhere.

TheMarex:
Merging this and PRing removing mlock from osrm-datastore separetely.

daniel-j-h:
:+1: please PR or ticket otherwise we will forget about it.

PR #3513, opened 2017-01-03
Mason clang format

oxidase:
# Issue

Resolves #3512: added `./scripts/format.sh` to `mason-linux-release` job. The script call is moved from `after_success` section where exit code is ignored to `before_install` to fail earlier. Also the mason script path is parametrized in MASON envar.

## Tasklist
 - [x] install clang-tidy via mason (https://github.com/mapbox/mason/tree/master/scripts/clang-format)
 - [x] make sure the Clang build on Travis detects it and runs it (fails the build). Example of a failing job https://travis-ci.org/oxidase/osrm-backend/jobs/188654943
 - [x] format the code base once and then have the check enabled for all prs
 - [x] review
 - [x] adjust for comments

PR #3514, opened 2017-01-04
Added info to CONTRIBUTING.md about automatic format checks

oxidase:
# Issue

Addition to #3513

## Tasklist
 - [x] review
 - [x] adjust for comments

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -7,6 +7,28 @@ You can add a :+1: emoji to the issue if you want to express interest in this.
 # Developer
 
 We use `clang-format` version `3.8` to consistently format the code base. There is a helper script under `scripts/format.sh`.
+The format is automatically checked by the `mason-linux-release` job of a Travis CI build.
+To save development time a local hook `.git/hooks/pre-push`
+```
+#!/bin/sh
+
+remote="$1"
+if [ x"$remote" = xorigin  ] ; then
+    if [ $(git rev-parse --abbrev-ref HEAD) = master ] ; then
+        echo "Rejected push to $remote/master" ; exit 1
+    fi
+
+    ./scripts/format.sh
+    if [ $? -ne 0 ] ; then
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -7,6 +7,17 @@ You can add a :+1: emoji to the issue if you want to express interest in this.
...
 
 We use `clang-format` version `3.8` to consistently format the code base. There is a helper script under `scripts/format.sh`.
+The format is automatically checked by the `mason-linux-release` job of a Travis CI build.
+To save development time a local hook `.git/hooks/pre-push`
+```
+#!/bin/sh
+
+remote="$1"
+if [ x"$remote" = xorigin ] ; then
+    ./scripts/format.sh || (echo "Unstaged format changes" ; exit 1)
```
MoKob: really makes your life easier

daniel-j-h:
Hm should we add the hook script into the scripts directory and add a note here on how to symlink to it? Or would that result in issues when we change the hook script and users jump around commits via git?

oxidase:
@daniel-j-h i don't think it is ok from the security side. git hooks are too private and I would copy-paste either from CONTRIBUTING.md or from the file.

Just thinking about a small extension: `scripts/format.sh` changes all files, so i would like to add `git rev-list $branch --not --remotes=origin`  to change only pushed files.

oxidase:
Ok, just after some attempts a :japanese_ogre: line
```
files=$(git diff --name-only $(git rev-list --reverse HEAD --not --remotes=$remote | head -1)~ HEAD | grep -e ".[ch]pp$")
```
to get the list of cpp/hpp files that will be pushed. Also `scripts/format.sh` needs changes to format only a list of files. 

I will add just a warning comment line that `scripts/format.sh` formats all local cpp/hpp. If you feel that formating of only pushed files is good, please create an improvement ticket.

PR #3517, opened 2017-01-04
Consider number of lanes to cross, resolves #3025.

### LINKED ISSUE #3025: Consider number of lanes to cross for Lane Anticipation
At the moment Lane Anticipation [has a 15s threshold](https://github.com/Project-OSRM/osrm-backend/blob/9eb7fc03cea177ed2a67452bad316053b6a9aecd/include/engine/guidance/lane_processing.hpp#L16-L23) for two subsequent turns. Turns further apart than 15s are not considered for anticipation.

The idea is to change from a static 15s threshold only to considering the number of lanes to cross in addition: having to change from the leftmost lane across three lanes to the rightmost lane should be given a higher threshold for anticipation as simply changing one lane.

daniel-j-h:
For #3025.

Lane Anticipation currently triggers on quick steps with lanes. This
changeset makes the "quick" part more dynamic by taking lanes left and
right of the turn into account. The reasoning for this is as follows.

The user can drive on the leftmost or rightmost lane and has to cross
all lanes left or right of the turn, respecitvely.

We scale our threshold appropriately, which now means the threshold
describes the duration the user needs for crossing _a single lane_.

Note: this is a heuristic and assumes the worst case. Which in my
opinion is fine to do since triggering Lane Anticipation in complex
scenarios is desirable.

## Tasklist
 - [x] review
 - [x] adjust for comments

### DIFF
--- a/src/engine/guidance/lane_processing.cpp
+++ b/src/engine/guidance/lane_processing.cpp
@@ -4,6 +4,7 @@
 #include "extractor/guidance/turn_instruction.hpp"
 #include "engine/guidance/post_processing.hpp"
 
+#include <algorithm>
 #include <iterator>
 #include <unordered_set>
 #include <utility>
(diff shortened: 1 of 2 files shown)


src/engine/guidance/lane_processing.cpp
```diff
@@ -27,9 +28,19 @@ std::vector<RouteStep> anticipateLaneChange(std::vector<RouteStep> steps,
...
         const auto has_lanes = step.intersections.front().lanes.lanes_in_turn > 0;
-        return has_lanes && is_quick;
+        if (!has_lanes)
+            return false;
+
+        // The more unused lanes to the left and right of the turn there are, the higher
+        // the chance the user is driving on one of those and has to cross lanes.
+        // Scale threshold for these cases to be adaptive to the situation's complexity.
+        const auto num_irrelevant = step.NumLanesToTheRight() + step.NumLanesToTheLeft();
+        const auto scale = std::max(1, num_irrelevant);
```
daniel-j-h: Do we want to scale linearly with the number of lanes?
Does crossing n lanes take n times as long as crossing 1 lane?

src/engine/guidance/lane_processing.cpp
```diff
@@ -27,9 +28,19 @@ std::vector<RouteStep> anticipateLaneChange(std::vector<RouteStep> steps,
...
-        const auto is_quick = step.duration < min_duration_needed_for_lane_change;
         const auto has_lanes = step.intersections.front().lanes.lanes_in_turn > 0;
-        return has_lanes && is_quick;
+        if (!has_lanes)
+            return false;
+
+        // The more unused lanes to the left and right of the turn there are, the higher
+        // the chance the user is driving on one of those and has to cross lanes.
+        // Scale threshold for these cases to be adaptive to the situation's complexity.
+        const auto num_irrelevant = step.NumLanesToTheRight() + step.NumLanesToTheLeft();
```
MoKob: Can you explain your intention here?
From what I understand you are summing up all lanes left/right of the turn? Shouldn't you only care about the ones you actually have to cross? These should be either the ones to the left, or the ones to the right.
So what would happen if you are on the second lane from the left on a 16 lane highway and you want to turn left. Would you get 15 times the threshold, or only once?

daniel-j-h: Yes you are right - but we don't know the lane source and target since we have no mapping.

Hm I assume a more conservative approach would be `max(lanes right, lanes left)`?

MoKob: I guess the max should be fine. After all, the worst case is that we are a bit over eager in restricting lanes. It will not hurt too much, if we do it to much. I don't think that these cases might be very common after all.

daniel-j-h: Changed :+1: Thanks for your feedback!

PR #3519, opened 2017-01-04
Guards against no lanes for lanes left and right of turn, resolves #3518

### LINKED ISSUE #3518: LanesToTheLeft / Right returns garbage when lanes are not present
https://github.com/Project-OSRM/osrm-backend/blob/487d8bc0ba8bcc7d131d1531f6629d2a1006a273/include/engine/guidance/route_step.hpp#L182-L184

In case the total lane count is zero we subtract the invalid lane id marker, wrap around and return garbage values. Instead we should return zero in those cases.

daniel-j-h:
For #3518.

In case the total lane count is zero we subtract the invalid lane id marker, wrap around and return garbage values. Instead we should return zero for lanes left / right of turn.

## Tasklist
 - [x] review
 - [x] adjust for comments

### DIFF
--- a/include/engine/guidance/route_step.hpp
+++ b/include/engine/guidance/route_step.hpp
@@ -90,12 +90,19 @@ struct RouteStep
     // copy all strings from origin into the step, apart from rotary names
     RouteStep &AdaptStepSignage(const RouteStep &origin);
 
-    LaneID NumLanesToTheRight() const;
+    // Lane utilities for the step's turn.
+    // A step may have lanes left or right of the turn: think left turn with lanes going straight.
+    // Note: Lanes for intersections along the way are available via intersections[n].lanes.
 
+    bool HasLanesAtTurn() const;
+
+    LaneID NumLanesInTurn() const;
+    LaneID NumLanesInTotal() const;
+
+    LaneID NumLanesToTheRight() const;
     LaneID NumLanesToTheLeft() const;
 
     auto LanesToTheLeft() const;
-
     auto LanesToTheRight() const;
 };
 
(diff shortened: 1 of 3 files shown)


include/engine/guidance/route_step.hpp
```diff
@@ -90,12 +90,19 @@ struct RouteStep
     // copy all strings from origin into the step, apart from rotary names
     RouteStep &AdaptStepSignage(const RouteStep &origin);
 
-    LaneID NumLanesToTheRight() const;
+    // Lane utilities for the step's turn.
+    // A step may have lanes left or right of the turn: think left turn with lanes going straight.
+    // Note: Lanes for intersections along the way are available via intersections[n].lanes.
+
+    bool hasLanesAtTurn() const;
```
danpat: Capitalization?

include/engine/guidance/route_step.hpp
```diff
@@ -172,16 +179,32 @@ inline RouteStep &RouteStep::AdaptStepSignage(const RouteStep &origin)
...
 
 inline LaneID RouteStep::NumLanesToTheLeft() const
 {
-    LaneID const total = intersections.front().lane_description.size();
-    return total - (intersections.front().lanes.lanes_in_turn +
-                    intersections.front().lanes.first_lane_from_the_right);
+    if (!hasLanesAtTurn())
+        return 0;
+
+    return NumLanesInTotal() - (NumLanesInTurn() + NumLanesToTheRight());
```
danpat: Math always makes me nervous, can you add an assert here that this is > 0?

include/engine/guidance/route_step.hpp
daniel-j-h: Urgh yeah will do

include/engine/guidance/route_step.hpp
daniel-j-h: Hm it's a uint8_t so it will never be negative - I could add an assert for lhs > rhs - not sure if this could ever happen (.lane_description out of sync with .lanes), so good to assert from my side. cc @MoKob

include/engine/guidance/route_step.hpp
```diff
@@ -172,16 +179,32 @@ inline RouteStep &RouteStep::AdaptStepSignage(const RouteStep &origin)
     return *this;
 }
 
+inline bool RouteStep::hasLanesAtTurn() const { return NumLanesInTotal() != 0; }
```
MoKob: danpat mentioned it already. Maybe do a full sweep on these for capitalisation.

include/engine/guidance/route_step.hpp
MoKob: Hm. I guess if we want to assert this, we should probably do this where the data is set, not every time it is used. Have you checked the initialisation for whether they are not already asserted to be fine?

include/util/guidance/turn_lanes.hpp
```diff
@@ -69,7 +69,7 @@ class LaneTuple
     bool operator!=(const LaneTuple other) const;
 
     LaneID lanes_in_turn;
-    LaneID first_lane_from_the_right;
+    LaneID first_lane_from_the_right; // is INVALID_LANEID when no lanes present
```
MoKob: Just a thought: the OSRM convention was to use these special-ids. Whats your take on using optionals for these? Would be a slight performance hit, but I guess it would be a good price to pay for getting rid of these where possible.

daniel-j-h: Would be great to have optionals for all of these - yes.
Since this is exactly the use-case optionals were designed for.

daniel-j-h:
Added assertions to step assembly to make sure the lane description and lane tuple is in sync. Also changed naming. Care to take a look again?

MoKob:
capture from chat: 👍

Issue #3208, opened 2016-10-28
Make libosrm algorithm aware


TheMarex:
We want to specialize the query-side of OSRM to use different algorithms. Primarily that means the following classes need to know about what algorithm we want to use:
1. Everything in `engine/routing_algorithms/*` (different implementation for each base algorithm)
2. Everything in `engine/data_facades/*` (loads different datafiles for each algorithm)
3. Everything in `engine/plugins/*` (needs to know which data facade to pass to which algorithm and whether that algorithm exists for the specified algorithm)

There are two ways we can approach this: Runtime polymorphism or compile-time polymorphism. Sine this abstraction touches a lot of our performance critical functions (`RouteStep` for example) compile-time polymorphism is strongly preferred.

One pattern to do this is policies or some other derived form (skeleton class that uses template functions), template specialization, etc.

A first step might be to extend `RoutingBase` by a template parameter for an algorithm tag like `osrm::routing_algorithm::CH` and make the current implementation the template specialization for `RoutingBase<CH>`.

/cc @daniel-j-h @danpat @MoKob on ideas to implement this nicely.

daniel-j-h:
Policy-based I/O writer we never merged: https://github.com/daniel-j-h/osrm-backend/blob/39383855dd6958a0de3b9981b2780b4a37ce4b56/io/osrm_writer.hpp as an example.

@MoKob is currently playing with a similar approach for geojson logger with arbitrary properties.

MoKob:
## My Feedback on Template Approach

In total I would love to see some measurements to back up [this approach](https://github.com/Project-OSRM/osrm-backend/commit/8b57aa37ce531a836901af7ef75ca9eea63511be) to get rid of the virtual functions we currently use in the project. I feel that using templates simply to get rid of virtual function calls adds a lot of complexity and maintenance cost over a few virtual function calls. Sure, the templates don't require the virtual functions but code duplication can blow up our code, also leading to slowdown. 
Compared to the overhead of creating our JSON response, both seem to be completely negligible to me. Our algorithm currently requires at most a few milliseconds to produce results, the construction of the JSON and the Geometry costs a lot more. I think we are close to a 100 ms?

[...]

PR #3522, opened 2017-01-05
remove CRTP leftovers

MoKob:
# Issue

Cleaning Up the Search, one step at a time.

## Tasklist
 - [x] review
 - [x] adjust for comments

## Requirements / Relations
none

### DIFF
--- a/include/engine/routing_algorithms/routing_base.hpp
+++ b/include/engine/routing_algorithms/routing_base.hpp
@@ -29,7 +29,7 @@ namespace engine
 namespace routing_algorithms
 {
 
-template <class DataFacadeT, class Derived> class BasicRoutingInterface
+template <class DataFacadeT> class BasicRoutingInterface
 {
   private:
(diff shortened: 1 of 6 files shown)


include/engine/routing_algorithms/routing_base.hpp
```diff
@@ -29,7 +29,7 @@ namespace engine
 namespace routing_algorithms
 {
 
-template <class DataFacadeT, class Derived> class BasicRoutingInterface
+template <class DataFacadeT> class BasicRoutingInterface
```
daniel-j-h: I think this was used in a CRTP pattern before, but I'm not completely sure about this. cc @TheMarex 

From what I can see it's unused currently and can go :+1:

TheMarex:
Same problem as with #3525, we are looking at possibly more templatisation not less.

TheMarex:
With #3581 in place this does not conflict anymore with #3208. I rebased this on master. Should be good to go once travis is 🍏

Issue #3501, opened 2016-12-28
Linker error for osrm-datastore target when using mason


Gubio:
Ubuntu 16.04
t2.micro AWS
swap 1gb

```
-- Enabling mason
-- Building on a 64 bit system
-- Using GNU gold as linker.
-- No build type specified, defaulting to Release
-- Configuring OSRM in release mode
-- Configuring release mode optimizations
-- LTO working
-- Using GCC specific binutils for LTO:
--  /usr/bin/gcc-ar
--  /usr/bin/gcc-ranlib
CMake Warning at CMakeLists.txt:276 (message):
  ENABLE_MASON and ENABLE_LTO/ENABLE_GOLD_LINKER may not work on all linux
  systems currently


CMake Warning at CMakeLists.txt:277 (message):
  For more details see:
  https://github.com/Project-OSRM/osrm-backend/issues/3202


-- Setting linker optimizations
-- Installing dependencies via mason
-- OpenMP support found. Linking just in case for stxxl
-- Configuring done
-- Generating done
-- Build files have been written to: /var/app/osrm-backend/build
[ 30%] Built target EXTRACTOR
[ 30%] Configuring revision fingerprint
-- Fingerprint unchanged, not regenerating
[ 30%] Built target FingerPrintConfigure
[ 42%] Built target UTIL
[ 46%] Built target osrm_extract
[ 65%] Built target ENGINE
[ 69%] Built target STORAGE
[ 73%] Built target osrm
[ 73%] Built target osrm_store
[ 73%] Linking CXX executable osrm-datastore
../mason_packages/linux-x86_64/boost_libprogram_options/1.61.0/lib/libboost_program_options.a(cmdline.o):libs/program_options/src/cmdline.cpp:function boost::program_options::detail::cmdline::run(): error: undefined reference to 'boost::program_options::basic_option<char>::~basic_option()'
../mason_packages/linux-x86_64/boost_libprogram_options/1.61.0/lib/libboost_program_options.a(cmdline.o):libs/program_options/src/cmdline.cpp:function boost::program_options::detail::cmdline::run(): error: undefined reference to 'boost::program_options::basic_option<char>::basic_option(boost::program_options::basic_option<char> const&)'
[...]

Issue #3441, opened 2016-12-13
Unable to compile with LTO on Ubuntu 14.04 (with ubuntu-toolchain-r ppa)


daniel-j-h:
Splitting off from https://github.com/Project-OSRM/osrm-backend/issues/3269#issuecomment-266561278 - this is a separate issue. @karenzshea reported seeing:

```
[ 56%] Linking CXX executable osrm-datastore
/tmp/ccfdAtPm.ltrans20.ltrans.o:<artificial>:function main: error: undefined reference to 'osrm::storage::StorageConfig::StorageConfig(boost::filesystem::path const&)'
/tmp/ccfdAtPm.ltrans20.ltrans.o:<artificial>:function main: error: undefined reference to 'osrm::storage::StorageConfig::IsValid() const'
/tmp/ccfdAtPm.ltrans20.ltrans.o:<artificial>:function main: error: undefined reference to 'osrm::storage::Storage::Storage(osrm::storage::StorageConfig)'
/tmp/ccfdAtPm.ltrans20.ltrans.o:<artificial>:function main: error: undefined reference to 'osrm::storage::Storage::Run(int)'
collect2: error: ld returned 1 exit status
make[2]: *** [osrm-datastore] Error 1
make[1]: *** [CMakeFiles/osrm-datastore.dir/all] Error 2
make[1]: *** Waiting for unfinished jobs....
```

```
-- The C compiler identification is GNU 4.9.4
-- The CXX compiler identification is GNU 4.9.4
-- Check for working C compiler: /usr/bin/gcc-4.9
-- Check for working C compiler: /usr/bin/gcc-4.9 -- works
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Detecting C compile features
-- Detecting C compile features - done
-- Check for working CXX compiler: /usr/bin/g++-4.9
-- Check for working CXX compiler: /usr/bin/g++-4.9 -- works
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Detecting CXX compile features
-- Detecting CXX compile features - done
-- Building on a 64 bit system
-- Using GNU gold as linker.
-- Disabling gc-sections on gold binutils < 2.26, see: https://sourceware.org/bugzilla/show_bug.cgi?id=17639
-- No build type specified, defaulting to Release
-- Configuring OSRM in release mode
-- Configuring release mode optimizations
-- Performing Test LTO_AVAILABLE
[...]

PR #3524, opened 2017-01-05
Disables link-time optimization unless -DENABLE_LTO=ON is passed.

daniel-j-h:
This disables the `-flto` LTO flag by default since we're seeing
segfaults in compiler lto plugins, binutils and linker errors again and
again for various clang / gcc / binutils combinations.

You can pass `-DNEBALE_LTO` to `cmake` in order to re-enable LTO.

LTO situation in short:
- LTO does not work at all for gcc<4.9
- With gcc>=4.9 the "slim" LTO format is getting used dumping IR
- Older binutils need LTO plugins which know how to read this IR
- Recent binutils handle this format all by themselves
- LLVM is more or less the same with some Clang versions segfaulting

If you need the performance benefit of LTO, make sure your compiler and
binutils are up to date and see for yourself if LTO builds work for you.

References:
- https://gcc.gnu.org/wiki/LinkTimeOptimizationFAQ
- https://github.com/Project-OSRM/osrm-backend/pull/3481#issuecomment-270618997
- https://github.com/Project-OSRM/osrm-backend/issues/3501
- https://github.com/Project-OSRM/osrm-backend/issues/3441

(and a ton of other LTO tickets if you search for them)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@
       - No longer emitting turns on ferries, if a ferry should use multiple docking locations
     - Profiles
       - Removed the `./profile.lua -> ./profiles/car.lua` symlink. Use specific profiles from the `profiles` directory.
+    - Infrastructure
+      - Disabled link-time optimized (LTO) builds by default. Enable by passing `-DENABLE_LTO=ON` to `cmake` if you need the performance and know what you are doing.
 
 # 5.5.1
   - Changes from 5.5.0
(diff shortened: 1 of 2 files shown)


daniel-j-h:
Which downstream projects relied on this default and have to be updated now?

TheMarex:
@daniel-j-h `node-osrm` needs to be updated to pass the changed default.

PR #3525, opened 2017-01-05
Reduce Unnecessary Templatisation

MoKob:
# Issue

Remove templates from routing algorithms to allow better maintenance/further refactor to create an actual routing interface.

When we want to switch out routing algorithms, having unnecessary templates will only increase maintenance cost/code complexity.

Right now all our routing algorithms are templated on the datafacade which, in turn, is  using virtual functions. By templating onto a reference of a base datafacade (e.g. [here](https://github.com/Project-OSRM/osrm-backend/blob/master/include/engine/plugins/viaroute.hpp#L32-L34)), we don't gain anything but code complexity. This PR removes this unnecessary template architecture and starts on separating interface and implementation for better readability.

To the reviewers: this is nearly completely moving code around. The only non-moving change is changing these https://github.com/Project-OSRM/osrm-backend/blob/dc81f581a0a9e231547dab9e24483dd273ba8c80/include/engine/routing_algorithms/alternative_path.hpp#L24-L26 into `constexpr` as a drive-by refactor.

## Tasklist
 - [x] review
 - [x] adjust for comments

## Requirements / Relations
Based on https://github.com/Project-OSRM/osrm-backend/pull/3522

### DIFF
--- a/include/engine/plugins/match.hpp
+++ b/include/engine/plugins/match.hpp
@@ -39,8 +39,8 @@ class MatchPlugin : public BasePlugin
 
   private:
     mutable SearchEngineData heaps;
-    mutable routing_algorithms::MapMatching<datafacade::BaseDataFacade> map_matching;
-    mutable routing_algorithms::ShortestPathRouting<datafacade::BaseDataFacade> shortest_path;
+    mutable routing_algorithms::MapMatching map_matching;
+    mutable routing_algorithms::ShortestPathRouting shortest_path;
     const int max_locations_map_matching;
 };
 }
--- a/include/engine/plugins/table.hpp
+++ b/include/engine/plugins/table.hpp
@@ -26,7 +26,7 @@ class TablePlugin final : public BasePlugin
 
   private:
     mutable SearchEngineData heaps;
-    mutable routing_algorithms::ManyToManyRouting<datafacade::BaseDataFacade> distance_table;
+    mutable routing_algorithms::ManyToManyRouting distance_table;
     const int max_locations_distance_table;
 };
 }
--- a/include/engine/plugins/trip.hpp
+++ b/include/engine/plugins/trip.hpp
@@ -30,11 +30,11 @@ class TripPlugin final : public BasePlugin
 {
   private:
     mutable SearchEngineData heaps;
-    mutable routing_algorithms::ShortestPathRouting<datafacade::BaseDataFacade> shortest_path;
-    mutable routing_algorithms::ManyToManyRouting<datafacade::BaseDataFacade> duration_table;
+    mutable routing_algorithms::ShortestPathRouting shortest_path;
+    mutable routing_algorithms::ManyToManyRouting duration_table;
     const int max_locations_trip;
 
-    InternalRouteResult ComputeRoute(const datafacade::BaseDataFacade &facade,
+    InternalRouteResult ComputeRoute(const std::shared_ptr<const datafacade::BaseDataFacade> facade,
                                      const std::vector<PhantomNode> &phantom_node_list,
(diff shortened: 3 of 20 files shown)


TheMarex:
This would need to be reverted in order to implement https://github.com/Project-OSRM/osrm-backend/issues/3208 since we need templatisation of `DataFacade` to implement compile-time switches on loading different data for MLD / CH.

While templating is not needed now I don't feel it makes sense to merge this and then reverse it again.

TheMarex:
> Can you elaborate why you see templates as only possible way of implementing algorithms? 

It is not the only way to solve this, but the compile time option using templates will be the fastest. This might make compile times slower, but I feel we can deal with this by moving code that does not depend on the datafacade or the search algorithm outside of the abstraction that would depend on it.

I strongly prefer compile time options here as this is the only part of our code base where performance really matters. This will actually also open up the possibility of to circumvent passing the abstract BaseDatafacade down to the user code and really exploit CRTP, as was intended here but never implemented fully.

MoKob:
Right now the compile time option does not save us anything. We are templating a base interface, making the look-up still depending on the actual virtual function call. So either we get rid of base classes for anything involved and template to the correct option directly, or we directly use the base class.

Right now we have

```
class Base
{
virtual functionality() = 0;
};

class A : public Base
{
functionality();
};

class B : public Base
{
functionality();
};
```

And use
```
template<class impl>
class Algo
{
void call(shared_ptr<impl> ptr);
};

Algo<Base> algo;
algo.call(make_shared(new B));
```

This setup does include a completely unnecessary template, as we would get get exact the same behaviour with 
```
class Algo
{
void call(shared_ptr<Base> ptr);
};

Algo algo;
algo.call(make_shared(new B));
```

This change is what I implemented and I don't see why we should keep this unnecessary template around.

MoKob:
Since you obviously have a concrete plan on how everything will have to be implemented, could you share that so we can all discuss on how this will end up to be?

I don't think that **We will have to pull a template parameter for the algorithm up to at least the Engine object giving us template <typename AlgorithmT> Engine**, since we will either have to make everything depend on switches or duplicate all the code anyhow.

No matter if we want to offer different executables or a single one that can switch at runtime. The only item that needs to do the switch is the Engine which can either be templated or be, again, implemented as a common base class for common operations and used directly without virtual functions with a CHEngine and a MLDEngine.

```
if(use_mld)
 MLDEngine(src,tgt)
if(use_ch)
 CHEngine(src,tgt)
```

No need to have
```
if(use_mld)
 Engine<MLD>(src,tgt)
if(use_ch)
 Engine<CH>(src,tgt)
```
The algorithms are so different, that we don't share much. And what we share can still be put into a common base class to be called from MLDEngine and CHEngine.

The same holds for the datafacade, as long as we want to be able to have it being switched at runtime, we need the common base class for everything used outside of the algorithms, since the interface has to match and the functionality is switched at runtime.
All depending on the algorithms can be implemented in the different datafacades for both MLD and CH, again saving us templates.

Everything outside of the algorithms itself should not depend on the algorithm used, apart from maybe loading the data from somewhere else (already done via virtual functions).

In total, we would still rely on no template at all, if we clearly separate what is belonging to what. Templated functionality that is shared can still be encapsulated in the common base type. Pulling everything through the entire routing engine only increases maintenance cost for possibly no benefit in speed at all.

TheMarex:
Can we hold on this for a bit? I want to explore this a little bit before we continue any discussions in this direction.

TheMarex:
With #3581 in place this does not conflict anymore with #3208. I rebased this on master. Should be good to go once #3522 is merged.

PR #3526, opened 2017-01-05
Upgrades LLVM 3.8.1 -> LLVM 3.9.1 on Travis

daniel-j-h:
LLVM 3.9.1 is now packaged in mason: https://github.com/mapbox/mason/issues/308
Upgrade our Clang compiler to the latest stable release.

Keeps clang-format at 3.8.1 for consistency.

Let's wait on Travis and see how it goes.


TheMarex:
@daniel-j-h does clang 3.9.1 work for you locally? I remember that 3.9 segfaulted for me.

daniel-j-h:
Yeah last weekend I build LLVM 3.9.1 locally and it lets me build master just fine. Check Travis, the mason build runs through, too.

That said, we're getting fscked by type deduction and debug symbols:
https://travis-ci.org/Project-OSRM/osrm-backend/jobs/189152203#L2071
> error: debug information for auto is not yet supported

I can see upstream tickets marked resolved fixed: https://llvm.org/bugs/show_bug.cgi?id=16876.
This is a hard error and not only a warning making our Clang 3.9.1 debug builds unusable. Argh!

daniel-j-h:
Forget about this - we actually have to update mason we're vendoring. LLVM 3.9.1 is quite new in mason and out vendored version does not yet include it. The Travis build fails to use Clang 3.9.1 from mason silently and falls back to the system's Clang 3.5. I'll update mason..

daniel-j-h:
@springmeyer it just came to me that we're blocked on this by a mason release:

- We need a new mason release which includes LLVM 3.9.1 we can vendor
- Then we need to vendor this release into third_party/mason
- Only then can we make use of Clang 3.9.1

Sidenote: the mason release build on Travis also silently fails and falls back to the system's Clang 3.5. The build passes since in release mode we're not hitting the debug symbol issue outlined above.

daniel-j-h:
All Travis builds work now with LLVM 3.9.1. The mason subtree update can not be squashed.

PR #3528, opened 2017-01-05
Add own update thread and removed query mutexes

TheMarex:
# Issue

This makes DataWatchdog spawn its own thread that checks for data
updates. This addresses an issue where libOSRM instances would not
notice a data update and thus never dettach from shared memory regions.

Replaces #3497 by using the implementation from @oxidase and adjusting some functionality.

## Tasklist
 - [x] add changelog entry that we are now launching our own thread
 - [x] review
 - [x] adjust for comments

## Requirements / Relations

Replaces #3497

### DIFF
--- a/include/engine/data_watchdog.hpp
+++ b/include/engine/data_watchdog.hpp
@@ -13,6 +13,7 @@
 #include <boost/thread/shared_mutex.hpp>
 
 #include <memory>
+#include <thread>
 
 namespace osrm
 {
@@ -26,10 +27,17 @@ class DataWatchdog
 {
   public:
     DataWatchdog()
-        : shared_barriers{std::make_shared<storage::SharedBarriers>()},
-          shared_regions(storage::makeSharedMemory(storage::CURRENT_REGION)),
-          current_timestamp{storage::REGION_NONE, 0}
+        : active(true)
+        , timestamp(0)
     {
+        watcher = std::thread(&DataWatchdog::Run, this);
+    }
+
+    ~DataWatchdog()
+    {
+        active = false;
+        barrier.region_condition.notify_all();
+        watcher.join();
     }
 
     // Tries to connect to the shared memory containing the regions table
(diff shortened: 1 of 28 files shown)


TheMarex:
> barrier.region_condition.notify_all();

Yeah I looked into that as well and so far I couldn't find any other solution then:

- waiting with a fixed timeout
- using `notify_all` to trigger wake-ups

I think just waking up the threads is as good as a solution as we will get.

> there is no way to send reliable signal from osrm-routed to js testing code.

Yes that is an unfortunate side-effect of the new data update model. The only way I would see to get this to work reliably is to make every instance of `libosrm` register somewhere and `osrm-datastore` would wait until all instances have confirmed that the data was switched. Technically it would be enough if we could wait all processes have detached from the old shared memory region.

Looking at http://man7.org/linux/man-pages/man2/shmctl.2.html we could maybe add a busy loop at the end of `osrm-datastore` that reads out the number of processes attached to the old shared memory region and then only terminates if the number reaches `1` (for `osrm-datastore`).

TheMarex:
I'm currently exploring the busy waiting/checking for number of attached processes road. Seems to be straight forward to integrate into our `SharedMemory` type as a `WaitForRemove()` function.

TheMarex:
@oxidase I think the current iteration should avoid any problems that need delays using the approach I outlined above. Can you give it another review if the changes work for you?

src/storage/storage.cpp
```diff
@@ -56,57 +49,19 @@ using QueryGraph = util::StaticGraph<contractor::QueryEdge::EdgeData>;
...
-    // hard unlock in case of any exception.
-    catch (boost::interprocess::lock_exception &)
+    if (!datastore_lock.TryLock())
     {
-        barriers.current_region_mutex.unlock_upgradable();
-        // make sure we exit here because this is bad
-        throw;
+        util::Log(logWARNING) << "Could not aquire current region lock after " << max_wait
+                              << " seconds. Claiming the lock by force.";
+        datastore_lock.ForceLock();
```
oxidase: This line can be a source of many problems, because it will make state of `CURRENT_REGION` undefined and anything could happen after executing `ForceLock`. I would suggest to use non-timed mutex, and add mutex removal to `CleanupSharedBarriers` to prevent making locked mutexes after signals.

include/engine/data_watchdog.hpp
```diff
@@ -38,102 +52,45 @@ class DataWatchdog
...
-            }
-            else
-            {
-                BOOST_ASSERT(current_timestamp.region == storage::REGION_2);
-                return std::make_pair(RegionsLock(shared_barriers->region_2_mutex), facade);
-            }
-        };
+    void Run()
+    {
+        auto shared_memory = makeSharedMemory(storage::CURRENT_REGION);
```
oxidase: Here `barrier.region_mutex` must be locked for access to `CURRENT_REGION`.  It must be moved at line 71

include/engine/data_watchdog.hpp
```diff
@@ -26,10 +27,23 @@ class DataWatchdog
...
   public:
     DataWatchdog()
-        : shared_barriers{std::make_shared<storage::SharedBarriers>()},
-          shared_regions(storage::makeSharedMemory(storage::CURRENT_REGION)),
-          current_timestamp{storage::REGION_NONE, 0}
+        : active(true)
+        , timestamp(0)
     {
+        watcher = std::thread(&DataWatchdog::Run, this);
+        while (!facade)
```
oxidase: :+1: but it is a heavy spinlock, and better to move into `engine.cpp` as
```
while (!watchdog.GetDataFacade())
  {
      sleep_for(10us);
  }
```

TheMarex: I would like the class to be in a usable state before the constructor finishes.
Having a watchdog that can potentially return an empty `shared_ptr` will crash the user-code to crash.

You think it is okay to trade a heavy constructor for ease-of-use?

include/engine/data_watchdog.hpp
TheMarex: Why is the lock necessary here? All we do is to attach to a pre-defined shared memory location (that needs to exist prior to `libosrm` starting up) and then get a pointer `current` to the data contained in it. I don't see any actual reading before  `while (active && timestamp == current->timestamp)` right?

TheMarex:
The reason why I wanted re-try logic for every lock is the following scenario, maybe you have some ideas on it:

_Given:_ A production server where we run `osrm-datastore` in fixed intervals (say every 10 minutes).
_Scenario:_ `osrm-datastore` crashes for whatever reason (maybe we have a bad download of a dataset, maybe a real bug).
_Goal:_ Try to recover from this scenario on the next update cycle without human intervention.

Without re-try logic we would probably see the datastore lock in a locked stated without any data update actually happening (the updating `osrm-datastore` process is long dead).

My idea was to set `--max-wait` to half the update interval (say 5mins), so we can be reasonably sure something went wrong. This of course yields the problems that you outlined above if there are several concurrent `osrm-datastore` processes running.

Ideal would be if we could use robust locks from http://pubs.opengroup.org/onlinepubs/9699919799/functions/pthread_mutex_lock.html that get automatically released by the kernel once the owner process dies. The major problem is that `boost::interprocess` has no support for this lock type, so I was working around that. I investigated extending the boost locks, but we would need a custom solution until the next boost release anyway.

src/storage/storage.cpp
```diff
@@ -118,113 +73,103 @@ Storage::ReturnCode Storage::Run(int max_wait)
...
-                BOOST_ASSERT(regions_layout.old_data_region == REGION_2);
-                barriers.resetRegion2();
-            }
+        storage::SharedMemory::Remove(next_region);
+        util::UnbufferedLog() << "ok.";
 
-            return ReturnCode::Retry;
-        }
+        util::UnbufferedLog() << "Waiting for clients to detach... ";
+        next_shared_memory->WaitForDetach();
```
TheMarex: > Also it is safe to remove the block and without waiting create the new one with the same shmid: the old block id will be changed to 0 (it makes the block private) and will be removed only when shm_nattch = 0 (no user of the block)

Hm yes the scenario I was worried about is having three data regions on a machine that can only fit two in main memory. At least with the `mlock` code that would have caused an instant crash due to lack of memory.

Do you have any idea how this would look without the `mlock` call? Is this still a problem?

include/engine/data_watchdog.hpp
oxidase: To have non-empty facade it must loaded in contructor and watcher thread must be started after the initial facade is loaded.

include/engine/data_watchdog.hpp
oxidase: Sure, did not get the change correctly, only data access must be protected

include/engine/data_watchdog.hpp
TheMarex: Right, that is actually a better idea. 👍

oxidase:
boost::named_mutex internally is a semaphore with name, and there is no robust analog of a semaphore

Just looked around and found a solution with [lock files](http://stackoverflow.com/a/2058642). The idea that kernel will release lock if process died. It will be a **per process lock** and as `osrm-datastore` is a single threaded it should be fine. Some experiments needed to check if `boost/interprocess/sync/file_lock.hpp` is an alternative

src/storage/storage.cpp
oxidase: Sorry, i don't understand how `mlock` is related here. The problem what i see is that `osrm-datastore` is waiting for an event `all routed requests using this memory block are terminated` that may never happen. On other side it is not a problem to remove outdated block that can be still in use. This will unassign ftok key ID of that block. Than it is possible allocate a new block and assign the key ID to the newly allocated block. 

For sure an outdated block will not be deallocated if any request still holds facade that was attached to the block. But it should solved either by restarting `osrm-routed` or by introducing a request timeout like 5 minutes.

TheMarex:
@oxidase Good find. Yes that seems like a workable solution, I will try this and report back.

oxidase:
it uses fcntl https://github.com/boostorg/interprocess/blob/develop/include/boost/interprocess/detail/os_file_functions.hpp#L520-L528 and exactly what is needed:
```
#include <boost/interprocess/sync/file_lock.hpp>
#include <boost/interprocess/sync/scoped_lock.hpp>
#include <iostream>

int main()
{
    boost::interprocess::file_lock file_lock("/tmp/lock");
    boost::interprocess::scoped_lock<boost::interprocess::file_lock> lock(file_lock);
    std::cout << "locked " << getpid() <<"\n";
    //while(1);
    sleep(10);
}
```

TheMarex:
@oxidase the file lock works beautifully. Good find. :+1:

src/storage/storage.cpp
TheMarex: The scenario I was worried about was overallocating shared memory if we exit before this line https://github.com/Project-OSRM/osrm-backend/blob/optimize/no-query-mutex/src/storage/storage.cpp#L173, have a long running query that keeps the region in memory long enough for the next run of `osrm-datastore` to hit a pre-existing shared memory region that is still in use.

Of course this is a very contrived example. I think you might be right that we can just mark it as removed in that very unlikely case.

TheMarex:
@oxidase ready for final review, I think I should have addressed all concerns.

src/tools/store.cpp
```diff
@@ -86,14 +87,21 @@ bool generateDataStoreOptions(const int argc,
         return false;
     }
 
+    if (option_variables.count("remove-locks"))
+    {
+        osrm::storage::SharedBarriers::remove();
+        boost::interprocess::named_mutex::remove("osrm-datastore");
```
oxidase: not relevant anymore

TheMarex:
Urg it seems I broke the windows build. let me just fix that as well.

oxidase:
@TheMarex i have pushed some fixes for windows tests. a small sleep is still required for windows and to avoid false-positives i changed default loading method on windows to "directly". Also i changed signal to kill `osrm-routed` from `SIGTERM` to `SIGINT` to simulate "keyboard interrupt" and removed  unprotected flag `this.child.down` because it can be overwritten by the next spawned  process.

There is still a problem with hangs on windows, but it is unrelated to the PR: if process is stopped by a signal, the code starting from https://github.com/Project-OSRM/osrm-backend/blob/master/src/tools/routed.cpp#L326 is not executed and `osrm-datastore` hangs after three `osrm-routed` starts and kills. I can not find the reason why it happens and why any first three runs after `osrm-unlock-all` are ok. If the server is stopped manually by `Ctrl-C` and the finalization code is executed than hangs could not be reproduced. It can be handled as another issue to fix signals handling on windows, but imho it is really low priority issue unless someone will run production code on a windows server.

TheMarex:
> There is still a problem with hangs on windows

Agreed this is not important right now. Maybe ticket it so we know it is there.

I'm currently investigating why `shmctl` on MacOSX does seem to throw an error.

TheMarex:
Okay I debugged this on OSX and we are getting an `EINVAL` error even though the shmid is correct. I'm really not sure why OSX is being so difficult, but for now I went the same route as for Windows and just added a wait.

PR #3529, opened 2017-01-06
Fixes various diagnostics

daniel-j-h:
Found with LLVM 3.9.1 and then fixed the most severe categories.

### DIFF
--- a/include/util/exception_utils.hpp
+++ b/include/util/exception_utils.hpp
@@ -5,11 +5,11 @@
 // Helper macros, don't use these ones
 // STRIP the OSRM_PROJECT_DIR from the front of a filename.  Expected to come
 // from CMake's CURRENT_SOURCE_DIR, which doesn't have a trailing /, hence the +1
-#define _PROJECT_RELATIVE_PATH(x) std::string(x).substr(strlen(OSRM_PROJECT_DIR) + 1)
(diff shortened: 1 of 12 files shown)


include/util/exception_utils.hpp
```diff
@@ -5,11 +5,11 @@
 // Helper macros, don't use these ones
 // STRIP the OSRM_PROJECT_DIR from the front of a filename.  Expected to come
 // from CMake's CURRENT_SOURCE_DIR, which doesn't have a trailing /, hence the +1
-#define _PROJECT_RELATIVE_PATH(x) std::string(x).substr(strlen(OSRM_PROJECT_DIR) + 1)
```
daniel-j-h: _Uppercase is reserved

include/util/geojson_debug_logger.hpp
```diff
@@ -47,7 +47,7 @@ class GeojsonLogger
     // cannot lock, is tooling for locked function
     static void output(bool first, const util::json::Array &array)
     {
-        for (const auto object : array.values)
+        for (const auto &object : array.values)
```
daniel-j-h: No longer copying potentially expensive variants around - this is the logger only so not really an issue

daniel-j-h:
Hm maybe only with selected `-W` categories. What I did was running with `-Weverything`, then grouping all diagnostics by category, manually picking the most severe categories and fixing those. There are tons and tons of warning categories we do not need to act on, e.g. all c++11 compat diagnostics.

Issue #3508, opened 2017-01-02
More information about map matching process in answers.


gardster:
I suggest to add some additional functions in the current matching engine to support incremental matching.
Current matching API is quite good, but it describes input as a separate complete track. There are not information about states during the call. And last track points may have bad confidence, because several routes are possible with same probability. There are not reliable methods to obtain the last point with good confidence and drop several points from the end.

I suggest to add additional parameters to matching API:

Point probability flag. This flag will add point probability information inside a tracepoints (or any other section) for the every input point.

I think this change will allow to use matching widely.

TheMarex:
> Point probability flag.

This is an interesting idea. I'm just not sure which probability we should expose here?

1. Expose the state emission probability:
    - how probable is it to see the GPS location given the matched street segment
    - this value is independent of any context
2. Expose the transition probability:
    - how probable is it that the street segment occurs after the previous street segment in a trace
    - does only account for context but not the actual matching point -> segment
3. Expose the join transition + emission probability
   - accounts for both

The main problem with just exposing these values is that the absolute values doesn't tell you much: They should always be looked on in relation to the other candidates for the same point. So we would need to come up with a mapping that normalizes this to an easily understandable range `0 .. 1` like we do for the `confidence` value of the whole trace.

An idea would be:

1. For every trace point save `transition + emission` log-probability of every candidate
2. When returning the trace compute the relative probability of the selected candidate.  `p_selected = e^logp_selected` and `p_scale = e^(logp_max - logp_min)` resulting in `p_confidence = p_selected / p_scale` which should normalize it to a usable value in the `0 .. 1` range.

gardster:
Thank you for explanation. Probability really seems really hard to expose. But for incremental matching most interesting think is how many alternative routes left inside the Viterbi algorithm. Because even they have good probabilities, if they don't participate in the Viterbi algorithm they will never be reached. This approach I suggest in my PR #3531 
More probabilistic functions requires more information about probabilities and links inside the Viterbi network (by intuition probability of the each inner tracepoint depends on score of the whole path in the Viterbi "network" and on score of all alternatives paths). It will take a lot of time to calculate all this scores.
What do you think if we add just alternative routes count field to the final output?

TheMarex:
@gardster interesting idea! This could also be useful for optimizations that look for "fixed points" in the map matching: Timestamps for which only one viable candidate is used by the next timestamp. At points like this we can forget all previous state and potentially save some memory/runtime later on.

[...]

PR #3531, opened 2017-01-06
Matching alternatives count output

gardster:
# Issue

https://github.com/Project-OSRM/osrm-backend/issues/3508

Possible way to determine degree of "finality" of the solution is how many ways that points to final HMM state lefts in the current layer. If there is only one candidate, that has link with final candidates set, than we can't change this candidate in future, even we obtain additional points after the last one. So we can save all matching results for points without alternatives and match only a part of the track when we obtain additional points. 

## Tasklist
 - [x] review
 - [x] adjust for comments

### DIFF
--- a/docs/http.md
+++ b/docs/http.md
@@ -298,7 +298,7 @@ The area to search is chosen such that the correct candidate should be considere
   Each `Waypoint` object has the following additional properties:
   - `matchings_index`: Index to the `Route` object in `matchings` the sub-trace was matched to.
   - `waypoint_index`: Index of the waypoint inside the matched route.
-  - `alternatives_count`: number of routes leading to the destination from this trace point. 1 means there are no other routes reaching destination. Greater values mean that there are different routes available and different route can be selected if you provide more coordinates.
+  - `alternatives_count`: number of alternative routes leading to the destination from this trace point. 0 means there are no other routes reaching destination. Greater values mean that there are different routes available and different route can be selected if you provide more coordinates.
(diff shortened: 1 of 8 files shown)


include/engine/routing_algorithms/map_matching.hpp
```diff
@@ -370,12 +371,8 @@ class MapMatching final : public BasicRoutingInterface<DataFacadeT, MapMatching<
             std::deque<std::pair<std::size_t, std::size_t>> reconstructed_indices;
             while (parent_timestamp_index > sub_matching_begin)
             {
-                if (model.breakage[parent_timestamp_index])
```
TheMarex: This seems unlreated, why was this removed?

gardster: It can lead to infinite loop. We don't change `parent_timestamp_index` or `sub_matching_begin` here, but continue to the while statement.

TheMarex:
I don't have time for an in-depth review right now, but I would like to see this in 5.7.0.

One thing I think is confusing is that the number of alternatives includes the currently selected route. "no alternatives" -> alternatives=0 would be the more intuitive interpretation. You probably need to port this to the `src/engine/routing_algorithms/map_matching.cpp` file since we shuffled code around.

gardster:
Ok. I port a code to the current master and decrease alternatives_count value.

TheMarex:
Finally had time to look at this again: Looks great. 👍 Merging.

PR #3537, opened 2017-01-09
Logs the current thread's id in assertions and for requests, resolves #3534

### LINKED ISSUE #3534: osrm-routed: link assertion error messages to the requests that triggered them
When `osrm-routed` hits an assertion, it exits with a message.  We *do* log requests before they're serviced, but in a busy multi-core environment, it's often difficult to tie the error message to a particular request without a bunch of time-consuming trial and error.

We should adjust the log messages so that when assertions are thrown, there is an easy way to quickly find the request that was being serviced by the thread that asserts.

One simple approach would be to log the thread ID in both request log messages, and the assertion error message.

daniel-j-h:
For #3534: prints tids for assertions and requests.
This lets us connect requests with assertions in a multi-threaded context.

Output will be

> [req][1267] get cheese
> [assert][1267] we're out of cheese

## Tasklist
 - [x] review
 - [x] adjust for comments

### DIFF
--- a/src/server/request_handler.cpp
+++ b/src/server/request_handler.cpp
@@ -25,6 +25,7 @@
 #include <iostream>
 #include <iterator>
 #include <string>
+#include <thread>
 
 namespace osrm
 {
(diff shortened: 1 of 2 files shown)


daniel-j-h:
Good to go here? @danpat?

MoKob:
Just tested it locally:
```
[debug] [req][0x70000f2b7000] /route/v1/driving/13.388860,52.517037;13.397634,52.529407?overview=false
[assert][0x70000f2b7000] /Users/mokob/git/engine/project-osrm/osrm-backend/src/engine/plugins/viaroute.cpp:34
in: osrm::engine::Status osrm::engine::plugins::ViaRoutePlugin::HandleRequest(const std::shared_ptr<datafacade::BaseDataFacade>, const api::RouteParameters &, util::json::Object &) const: !route_parameters.IsValid()
```

Looks good to go to me.

PR #3538, opened 2017-01-09
Provides OSRM_ASSERT_WITH_LOC macros for asserting with location, resolves #3533

### LINKED ISSUE #3533: Assertions should log coordinates wherever possible
We've got lots of assertions all over the codebase, which is great.

What's not so great is that if we hit one when dealing with planet-sized data, it's often very difficult to track down what the cause was.

Many errors are location specific - they are caused due to specific network shape or properties.

We should create a wrapper to `BOOST_ASSERT`, called `OSRM_ASSERT_WITH_LOCATION`, where we pass in location data.  If the assertion is thrown, we should log the coordinates in addition to the failed assertion information.  This would (hopefully) make it much faster to track down datasets to reproduce the problem.

This is not always going to be possible, but it would massively speedup debugging for location-specific assertion failures.

daniel-j-h:
For #3533.

Assertions now log a clickable location url:

> [Location] http://www.openstreetmap.org/?mlat=52.5396&mlon=13.2954#map=19/52.5396/13.2954
> [assert] /tmp/osrm-backend/src/extractor/guidance/sliproad_handler.cpp:322 in: virtual osrm::extractor::guidance::Intersection osrm::extractor::guidance::SliproadHandler::operator()(NodeID, EdgeID, osrm::extractor::guidance::Intersection) const: false

## Tasklist
 - [x] review
 - [x] adjust for comments

### DIFF
--- a/include/util/assert.hpp
+++ b/include/util/assert.hpp
@@ -1,0 +1,45 @@
+#ifndef OSRM_UTIL_ASSERT_HPP
+#define OSRM_UTIL_ASSERT_HPP
+
+#include <iostream>
+#include <string>
+
+#include "util/coordinate.hpp"
+
+#include <boost/assert.hpp>
+
+// Enhances BOOST_ASSERT / BOOST_ASSERT_MSG with convenience location printing
+// - OSRM_ASSERT(cond, coordinate)
+// - OSRM_ASSERT_MSG(cond, coordinate, msg)
+
+#ifdef BOOST_ENABLE_ASSERT_HANDLER
+
+#define OSRM_ASSERT_MSG(cond, loc, msg)                                                            \
+    do                                                                                             \
+    {                                                                                              \
+        if (!static_cast<bool>(cond))                                                              \
+        {                                                                                          \
+            ::osrm::util::FloatCoordinate c_(loc);                                                 \
+            std::cerr << "[Location] "                                                             \
+                      << "http://www.openstreetmap.org/?mlat=" << c_.lat << "&mlon=" << c_.lon     \
+                      << "#map=19/" << c_.lat << "/" << c_.lon << '\n';                            \
+        }                                                                                          \
+        BOOST_ASSERT_MSG(cond, msg);                                                               \
+    } while (0)
+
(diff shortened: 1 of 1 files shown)


daniel-j-h:
Callsites need to be adapted where possible.

oxidase:
@daniel-j-h would it make sense to catch assertion exceptions, log request part of URL and return error code in json instead of hard aborts?

daniel-j-h:
Hm not sure about that - catching assertions from deep down the engine and going on in routed can be potentially problematic (think: corrupting internal state). Waiting on @danpat for what he had in mind anyway.

oxidase:
@daniel-j-h 
 > corrupting internal state
Hrm, in the engine should not be an internal mutable state (think scaling), so failing one request should not influence others. At the moment an asserted request leads to the complete server process restart.

To have immutable data a compile-time guarantee of constness is required, eg datafacades must have the const qualifier. Anyway it deserves another PR that could add handling of asserts after finishing ongoing refactoring

daniel-j-h:
Ah I see now what you mean - can you ticket this up so we don't forget about it.

include/util/assert.hpp
```diff
@@ -0,0 +1,45 @@
...
+
+#include <boost/assert.hpp>
+
+// Enhances BOOST_ASSERT / BOOST_ASSERT_MSG with convenience location printing
+// - OSRM_ASSERT_WITH_LOC(cond, coordinate)
+// - OSRM_ASSERT_WITH_LOC_MSG(cond, coordinate, msg)
+
+#ifdef BOOST_ENABLE_ASSERT_HANDLER
+
+#define OSRM_ASSERT_WITH_LOC_MSG(cond, loc, msg)                                                   \
```
TheMarex: Minor nitpick `OSRM_ASSERT` is a little bit shorter to type then `OSRM_ASSERT_WITH_LOC` and since we don't have `OSRM_ASSERT_WITHOUT_LOC` I don't think we need that extra piece of information. :-)

daniel-j-h:
People - start using this! ;)

PR #3543, opened 2017-01-09
Supports destination:street, resolves #3541

### LINKED ISSUE #3541: Support destination:street tag
Support the `destination:street=` tag (and check if there are others we are mapping but missing out on).

http://taginfo.openstreetmap.org/search?q=destination

cc @pratikyadav @MoKob

daniel-j-h:
For #3541. We seem to map `destination:street` where appropriate. 
This changeset adds support for it.

cc @pratikyadav

## Tasklist
 - [x] review
 - [x] adjust for comments


daniel-j-h:
Good to go here?

profiles/lib/destination.lua
```diff
@@ -18,11 +18,18 @@ end
...
   dest = Destination.get_directional_tag(way, is_forward, 'destination')
+  street = Destination.get_directional_tag(way, is_forward, 'destination:street')
   if ref and dest then
     return ref .. ': ' .. dest
+  elseif ref then
+    return ref
+  elseif dest then
+    return dest
+  elseif street then
+    return street
```
MoKob: capturing from desk: this could be revisited to use the former syntax `return ref or dest or street or ' '`, as long as we do the check for street anyhow.

PR #3545, opened 2017-01-09
Contractor loading edges refactoring

oxidase:
# Issue

The PR is aimed to reduce code duplication in `Contractor::LoadEdgeExpandedGraph`

## Tasklist
 - [x] Make CSV parsing function generalized
 - [x] Decouple R-Tree updates and edges loading
 - [x] review
 - [x] adjust for comments

## Requirements / Relations
Related issues  #3446 #2399

### DIFF
--- a/src/contractor/contractor.cpp
+++ b/src/contractor/contractor.cpp
@@ -22,9 +22,13 @@
 #include <boost/assert.hpp>
 #include <boost/filesystem/fstream.hpp>
 #include <boost/functional/hash.hpp>
+#include <boost/fusion/adapted/std_pair.hpp>
+#include <boost/fusion/include/adapt_adt.hpp>
 #include <boost/interprocess/file_mapping.hpp>
 #include <boost/interprocess/mapped_region.hpp>
+#include <boost/spirit/include/phoenix.hpp>
 #include <boost/spirit/include/qi.hpp>
+#include <boost/spirit/include/support_line_pos_iterator.hpp>
 
 #include <tbb/blocked_range.h>
 #include <tbb/concurrent_unordered_map.h>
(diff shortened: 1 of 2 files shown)


src/contractor/contractor.cpp
```diff
@@ -45,37 +49,162 @@
...
+    std::uint8_t source;
 };
 
-template <> struct hash<std::tuple<OSMNodeID, OSMNodeID, OSMNodeID>>
+struct Turn final
 {
-    std::size_t operator()(const std::tuple<OSMNodeID, OSMNodeID, OSMNodeID> &k) const noexcept
+    std::uint64_t from, via, to;
+    Turn() : from(0), via(0), to(0) {}
+    Turn(const std::uint64_t &from, const std::uint64_t &via, const std::uint64_t &to)
```
daniel-j-h: Here and above - wondering why you're passing primitive types by ref

src/contractor/contractor.cpp
```diff
@@ -45,37 +49,162 @@
...
 {
     static_assert(sizeof(T) % alignof(T) == 0, "pointer can not be used as an array pointer");
     return reinterpret_cast<uintptr_t>(pointer) % alignof(T) == 0;
 }
+} // anon ns
+
+BOOST_FUSION_ADAPT_STRUCT(Segment, (std::uint64_t, from)(std::uint64_t, to))
+BOOST_FUSION_ADAPT_STRUCT(SpeedSource, (unsigned, speed))
+BOOST_FUSION_ADAPT_STRUCT(Turn, (std::uint64_t, from)(std::uint64_t, via)(std::uint64_t, to))
+BOOST_FUSION_ADAPT_STRUCT(PenaltySource, (double, penalty))
```
daniel-j-h: Hrm we require and old Boost version without the type deduction macro, right? What about manually deducing the types here via C++11+ features so we're not duplicating the types used in the impl. here?

src/contractor/contractor.cpp
```diff
@@ -45,37 +49,162 @@
...
+// Function parses a list of CSV files using "key,value,comment" grammar.
+// Key and Value structures must be a model of Random Access Sequence.
+// Also the Value structure must have source member that will be filled
+// with the index of the csv_filenames array.
+// Function returns a lambda function that maps input Key to boost::optional<Value>
+template <typename Key, typename Value>
+auto ParseCSV(
+    const std::vector<std::string> &csv_filenames,
+    const qi::rule<boost::spirit::line_pos_iterator<boost::spirit::istream_iterator>, Key()> &key,
+    const qi::rule<boost::spirit::line_pos_iterator<boost::spirit::istream_iterator>, Value()>
```
daniel-j-h: Urgh maybe typedef those?

src/contractor/contractor.cpp
```diff
@@ -45,37 +49,162 @@
...
+BOOST_FUSION_ADAPT_STRUCT(Turn, (std::uint64_t, from)(std::uint64_t, via)(std::uint64_t, to))
+BOOST_FUSION_ADAPT_STRUCT(PenaltySource, (double, penalty))
+
+// Function parses a list of CSV files using "key,value,comment" grammar.
+// Key and Value structures must be a model of Random Access Sequence.
+// Also the Value structure must have source member that will be filled
+// with the index of the csv_filenames array.
+// Function returns a lambda function that maps input Key to boost::optional<Value>
+template <typename Key, typename Value>
+auto ParseCSV(
```
daniel-j-h: With the return type type deduction, can you at least make a comment what this function returns.

src/contractor/contractor.cpp
```diff
@@ -45,37 +49,162 @@
...
+        std::vector<std::pair<Key, Value>> lookup;
+        tbb::parallel_for(
+            std::size_t{0},
+            csv_filenames.size(),
+            [&](const std::size_t idx) {
+                const auto file_id = idx + 1; // starts at one, zero means we assigned the weight
+                const auto &filename = csv_filenames[idx];
+
+                std::ifstream input_stream(filename, std::ios::binary);
+                input_stream.unsetf(std::ios::skipws);
```
daniel-j-h: Is there a reason for formatted input here? Are istreambuf_iterators good enough?

src/contractor/contractor.cpp
```diff
@@ -45,37 +49,162 @@
...
+        osrm::util::Log() << "In total loaded " << csv_filenames.size()
+                          << " file(s) with a total of " << lookup.size() << " unique values";
+
+        return [lookup](const Key &key) {
+            const auto it = std::lower_bound(
+                std::begin(lookup), std::end(lookup), key, [](const auto &lhs, const auto &rhs) {
+                    return lhs.first < rhs;
+                });
+            return it != std::end(lookup) && !(key < it->first) ? boost::optional<Value>(it->second)
+                                                                : boost::optional<Value>();
```
daniel-j-h: :+1:

src/contractor/contractor.cpp
oxidase: Here I could wrap return type into `std::function` but it is too much verbose. Please let me know if a line  above  "Function returns a lambda function that maps input Key to boost::optional<Value>" is ok.

src/contractor/contractor.cpp
oxidase: here `skipws` must be *unset*, otherwise `ifstream` eats all spaces and eols, so `line_pos_iterator` can not correctly count line numbers.

src/contractor/contractor.cpp
oxidase: yeah, this is bad looking, because input grammar parameters should know about internals to have correct  `Iterator` type at the compile time. I will move the function into a functor object `CSVParser` and make typedefs `Iterator` and `Rule`, and `operator(const ... &filenames, const Rule& key, const Rule &value)`

src/contractor/contractor.cpp
oxidase: no any reason, will remove `&`

src/contractor/contractor.cpp
oxidase: I don't fully get it: do you mean use `auto` keyword as in last example http://www.boost.org/doc/libs/1_63_0/libs/fusion/doc/html/fusion/adapted/adapt_struct.html ?

daniel-j-h: Somethine like

    ADAPT_STRUCT(Member, decltype(Type::Member))

oxidase:
As discussed with @danpat no turn penalty datasourse IDs are required in near future. So instead of decoupling R-tree updates and edge loading, only small refactoring is done that will give smaller patch for #2399.

oxidase:
rebased and adjusted to comments

PR #3554, opened 2017-01-11
Fixes default profile symlink not being available

daniel-j-h:
Adaption for b8e13d9b - people are running into this.

(Should probably not be a default in the first place or rel. to the base binary ..)

## Tasklist
 - [x] review
 - [ ] adjust for comments


TheMarex:
Cool, was this the only place where we mentioned that?

daniel-j-h:
Yes

PR #3558, opened 2017-01-11
Add abstraction layer to support multiple algorithms

TheMarex:
# Issue

This implements #3208 by using a template tag to automatically generate a class hierarchy for an algorithm and only use actual virtual function calls in a minimum of places.

This design is the result of the following prototype https://gist.github.com/TheMarex/8ba7594cab6082f5c39f08e79b0152db that compares a manually written OOM scheme against template based code.

## Compile time

    ccache -C
    cmake ..
    time make osrm # cold build

`master`

   make osrm  104.80s user 7.58s system 99% cpu 1:52.99 total

`refactor/algorithm-aware`

   make osrm  121.77s user 8.10s system 99% cpu 2:10.49 total

## Binary size

    ls -l libosrm.a

`master`

    6052202

`refactor/algorithm-aware`

    6590768

## Performance

Measuring the 10000 random requests on California the result as basically the same:

`master`

median: 0.462845 p95: 11.4442 p05: 0.322627
median: 0.459129 p95: 11.2884 p05: 0.320714
median: 0.498818 p95: 11.7293 p05: 0.339625

`refactor/algorithm-aware`

median: 0.462679 p95: 11.3626 p05: 0.319035
median: 0.462412 p95: 11.2776 p05: 0.321605
median: 0.456968 p95: 11.3829 p05: 0.32047

## Tasklist
 - [x] Finish up #3528 
 - [x] micro-benchmarks for osrm-routed
 - [x] review
 - [x] adjust for comments

## Requirements / Relations

Depends on #3528 to remove the query locks for `SharedMemoryDataFacade`.


oxidase:
I have temporarily added MLD data to the data facade in 1b71375094f4077306bae6fc2ebd03abe765d4be can cherry-picked here or a separate PR is required for MLD data facade?

TheMarex:
@oxidase I would prefer two separate PRs. This PR is not supposed to change/add any functionality to make it easier to find regressions.

TheMarex:
I think I found a solution to abstract away the algorithms for the plugins so we don't need them in the header file. Will try this first before we start the review process.

src/osrm/osrm.cpp
```diff
@@ -15,7 +16,18 @@ namespace osrm
...
-OSRM::OSRM(engine::EngineConfig &config) : engine_(std::make_unique<engine::Engine>(config)) {}
+OSRM::OSRM(engine::EngineConfig &config)
+{
+    if (engine::Engine<engine::algorithm::CoreCH>::CheckCompability(config))
+    {
+        engine_ = std::make_unique<engine::Engine<engine::algorithm::CoreCH>>(config);
+    }
+    else
+    {
+        // Will fail hard if there is no CH data
```
oxidase: Better to check compatibility for `CH` and throw exception if not, otherwise fail can be somewhere deep in routing code

src/osrm/osrm.cpp
```diff
@@ -15,7 +16,18 @@ namespace osrm
 
 // Pimpl idiom
 
-OSRM::OSRM(engine::EngineConfig &config) : engine_(std::make_unique<engine::Engine>(config)) {}
+OSRM::OSRM(engine::EngineConfig &config)
+{
+    if (engine::Engine<engine::algorithm::CoreCH>::CheckCompability(config))
```
oxidase: At the moment this makes a priority of check CoreCH -> CH.
Is it ok to add a command line option in this PR to check requested algorithm  `config.use_core_ch && engine::Engine<engine::algorithm::CoreCH>::CheckCompability(config)`?
It will unlock possibility to use both CH and MLD in one dataset.

src/osrm/osrm.cpp
```diff
@@ -15,7 +16,18 @@ namespace osrm
...
+    {
+        engine_ = std::make_unique<engine::Engine<engine::algorithm::CoreCH>>(config);
+    }
```
oxidase: A log line about loaded data would nice here

include/engine/data_watchdog.hpp
```diff
@@ -82,7 +83,7 @@ class DataWatchdog
     std::thread watcher;
```
oxidase: One step further is to make `storage::SharedDataTimestamp` parameterized with `AlgorithmT`.
It will give possibility later update only one specific data portion in `/dev/shm/osrm-region-ch`,  `/dev/shm/osrm-region-core-ch` or ` `/dev/shm/osrm-region-mld``

include/engine/datafacade/algorithm_datafacade.hpp
```diff
@@ -0,0 +1,67 @@
...
+#include "util/integer_range.hpp"
+
+namespace osrm
+{
+namespace engine
+{
+namespace datafacade
+{
+
+using EdgeRange = util::range<EdgeID>;
```
daniel-j-h: Seeing this, I'd really love our graphs moving away from the manual node and edge iteration and provide a proper iterator / range based interface. Would be great to be able to say

    for (auto node in graph.Nodes())
        for (auto edge in graph.Edges(node))
            fn(edge);

The partition graph already has something similar to this:

https://github.com/Project-OSRM/osrm-backend/blob/6f34d70907dadb225d8a2e8bc70827275e991fc6/include/partition/partition_graph.hpp

(It's out of context here but it just came to mind while looking at the EdgeRange)

include/engine/datafacade/contiguous_internalmem_datafacade.hpp
```diff
@@ -44,33 +46,169 @@ namespace engine
 namespace datafacade
 {
 
+template <typename AlgorithmT> class ContiguousInternalMemoryAlgorithmDataFacade;
+
+template <>
+class ContiguousInternalMemoryAlgorithmDataFacade<algorithm::CH>
+    : public datafacade::AlgorithmDataFacade<algorithm::CH>
```
daniel-j-h: In

    class X<Algo> : Y<Algo>

does Algo always have to be the same?

include/engine/datafacade/contiguous_internalmem_datafacade.hpp
```diff
@@ -44,33 +46,169 @@ namespace engine
...
+    }
+
+    void InitializeInternalPointers(storage::DataLayout &data_layout, char *memory_block)
+    {
+        InitializeCoreInformationPointer(data_layout, memory_block);
+    }
+
+    bool IsCoreNode(const NodeID id) const override final
+    {
+        if (m_is_core_node.size() > 0)
```
daniel-j-h: When can this happen? Using CoreCH but then having no core nodes? When core factor = 0?

include/engine/datafacade/contiguous_internalmem_datafacade.hpp
```diff
@@ -1008,6 +1058,36 @@ class ContiguousInternalMemoryDataFacade : public BaseDataFacade
                     m_lane_description_offsets[lane_description_id + 1]);
     }
 };
+
+template <typename AlgorithmT> class ContiguousInternalMemoryDataFacade;
+
+template <>
+class ContiguousInternalMemoryDataFacade<algorithm::CH>
+    : public ContiguousInternalMemoryDataFacadeBase,
+      public ContiguousInternalMemoryAlgorithmDataFacade<algorithm::CH>
```
daniel-j-h: Ah, so here we're connecting base and algo-specific facades?

include/engine/datafacade_provider.hpp
```diff
@@ -0,0 +1,56 @@
...
+};
+
+template <typename AlgorithmT> class ImmutableProvider final : public DataFacadeProvider<AlgorithmT>
+{
+    using FacadeT = datafacade::ContiguousInternalMemoryDataFacade<AlgorithmT>;
+
+  public:
+    ImmutableProvider(const storage::StorageConfig &config)
+        : immutable_data_facade(std::make_shared<FacadeT>(
+              std::make_shared<datafacade::ProcessMemoryAllocator>(config)))
```
daniel-j-h: Urgh this is a shared ptr to a shared ptr to an allocator?

include/engine/engine.hpp
```diff
@@ -30,38 +32,162 @@ namespace osrm
...
+    virtual Status Table(const api::TableParameters &parameters,
+                         util::json::Object &result) const = 0;
+    virtual Status Nearest(const api::NearestParameters &parameters,
+                           util::json::Object &result) const = 0;
+    virtual Status Trip(const api::TripParameters &parameters,
+                        util::json::Object &result) const = 0;
+    virtual Status Match(const api::MatchParameters &parameters,
+                         util::json::Object &result) const = 0;
+    virtual Status Tile(const api::TileParameters &parameters, std::string &result) const = 0;
+};
```
daniel-j-h: Why the split again? For osrm <-> Engine pimpl? Can't you fwd. declare Engine, the unique_ptr will happily store a fwd. declared type - then in the impl. include the header.

include/engine/engine.hpp
```diff
@@ -30,38 +32,162 @@ namespace osrm
 namespace engine
 {
 
-class Engine final
+class EngineInterface
+{
+  public:
+    virtual ~EngineInterface(){};
```
daniel-j-h: virtual ~E() = default;

include/engine/engine.hpp
```diff
@@ -30,38 +32,162 @@ namespace osrm
...
+                std::make_unique<ImmutableProvider<AlgorithmT>>(config.storage_config);
+        }
+    }
 
     Engine(Engine &&) noexcept = delete;
     Engine &operator=(Engine &&) noexcept = delete;
 
     Engine(const Engine &) = delete;
     Engine &operator=(const Engine &) = delete;
+    virtual ~Engine(){};
```
daniel-j-h: Why virtual? Why an empty dtor at all?

include/engine/engine.hpp
```diff
@@ -30,38 +32,162 @@ namespace osrm
...
+
+    Status Route(const api::RouteParameters &params,
```
daniel-j-h: Here and below: it's always the same: get facade, init algos, plugin.HandleRequest - can we abstract over it?

include/engine/engine.hpp
```diff
@@ -30,38 +32,162 @@ namespace osrm
...
+        boost::interprocess::scoped_lock<mutex_type> current_region_lock(barrier.get_mutex());
+
+        auto mem = storage::makeSharedMemory(barrier.data().region);
+        auto layout = reinterpret_cast<storage::DataLayout *>(mem->Ptr());
+        return layout->GetBlockSize(storage::DataLayout::CH_GRAPH_NODE_LIST) > 4 &&
+               layout->GetBlockSize(storage::DataLayout::CH_GRAPH_EDGE_LIST) > 4;
+    }
+    else
+    {
+        std::ifstream in(config.storage_config.hsgr_data_path.string().c_str());
```
daniel-j-h: if(!in) throw

include/engine/engine.hpp
```diff
@@ -30,38 +32,162 @@ namespace osrm
...
+        boost::interprocess::scoped_lock<mutex_type> current_region_lock(barrier.get_mutex());
+
+        auto mem = storage::makeSharedMemory(barrier.data().region);
+        auto layout = reinterpret_cast<storage::DataLayout *>(mem->Ptr());
+        std::cout << layout->GetBlockSize(storage::DataLayout::CH_CORE_MARKER) << std::endl;
+        return layout->GetBlockSize(storage::DataLayout::CH_CORE_MARKER) > 4;
+    }
+    else
+    {
+        std::ifstream in(config.storage_config.core_data_path.string().c_str());
```
daniel-j-h: if(!in) throw

include/engine/routing_algorithms.hpp
```diff
@@ -0,0 +1,163 @@
...
+    virtual bool HasMapMatching() const = 0;
+    virtual bool HasManyToManySearch() const = 0;
+    virtual bool HasGetTileTurns() const = 0;
+};
+
+namespace detail
+{
+
+struct NotImplementedException : public std::runtime_error
+{
```
daniel-j-h: You should inherit ctors, too
```
struct E : std::runtime_error {
  using Base = std::runtime_error;
  using Base::Base;
};
```

daniel-j-h: Also is this unused right now?

include/engine/routing_algorithms.hpp
```diff
@@ -0,0 +1,163 @@
...
+template <typename AlgorithmT> class RoutingAlgorithms final : public RoutingAlgorithmsInterface
+{
+  public:
+    RoutingAlgorithms(SearchEngineData &heaps,
+                      const datafacade::ContiguousInternalMemoryDataFacade<AlgorithmT> &facade)
+        : heaps(heaps), facade(facade)
+    {
+    }
+
+    virtual ~RoutingAlgorithms() = default;
```
daniel-j-h: Why? The type is final (?)

include/engine/routing_algorithms/direct_shortest_path.hpp
```diff
@@ -19,30 +15,31 @@ namespace engine
...
 {
 
+template <typename AlgorithmT>
+InternalRouteResult
+directShortestPathSearch(SearchEngineData &,
+                         const datafacade::ContiguousInternalMemoryDataFacade<AlgorithmT> &,
+                         const std::vector<PhantomNodes> &)
+{
+    throw util::exception(std::string("directShortestPathSearch is not implemented for ") +
+                          typeid(AlgorithmT).name());
```
daniel-j-h: Urgh I wouldn't use rtti - if you really need a string repr. I would simply add another trait to get a const char* literal

include/engine/routing_algorithms/routing_base.hpp
```diff
@@ -31,403 +33,641 @@ namespace engine
...
-class BasicRoutingInterface
+static constexpr bool FORWARD_DIRECTION = true;
+static constexpr bool REVERSE_DIRECTION = false;
+
+// Stalling
+template <bool DIRECTION, typename HeapT>
+bool stallAtNode(const datafacade::ContiguousInternalMemoryDataFacade<algorithm::CH> &facade,
+                 const NodeID node,
+                 const EdgeWeight weight,
+                 HeapT &query_heap)
```
daniel-j-h: Hm can this be a const &?

src/engine/plugins/tile.cpp
```diff
@@ -247,33 +230,58 @@ FixedPoint coordinatesToTilePoint(const util::Coordinate point, const BBox &tile
...
+std::vector<std::size_t> getEdgeIndex(const std::vector<RTreeLeaf> &edges)
+{
+    // In order to ensure consistent tile encoding, we need to process
+    // all edges in the same order.  Differences in OSX/Linux/Windows
+    // sorting methods mean that GetEdgesInBox doesn't return the same
+    // ordered array on all platforms.
+    // GetEdgesInBox is marked `const`, so we can't sort the array itself,
+    // instead we create an array of indexes and sort that instead.
+    std::vector<std::size_t> sorted_edge_indexes(edges.size(), 0);
+    std::iota(sorted_edge_indexes.begin(), sorted_edge_indexes.end(), 0); // fill with 1,2,3,...N
```
daniel-j-h: comment is wrong, starts with 0 :P

src/osrm/osrm.cpp
TheMarex: I was planning on adding a flag `use_mld` for when we generalize this for MLD. I don't think it makes sense to use vanilla CH data when you could be using CoreCH: It just makes things a lot slower by not using the extra data that is there.

src/osrm/osrm.cpp
TheMarex: `Engine` will throw a detailed exception in case there are files/shared memory regions missing.

src/osrm/osrm.cpp
TheMarex: Added a log line in `Engine` that print shared vs. internal memory + algorithm used.

include/engine/data_watchdog.hpp
TheMarex: I agree, this would be sensible in the future. I would rather add this in another PR however.

include/engine/algorithm.hpp
```diff
@@ -0,0 +1,87 @@
...
+struct CH final
+{
+};
+// Contraction Hiearchy with core
+struct CoreCH final
+{
+};
+}
+
+namespace algorithm_trais
```
TheMarex: Hrm, I tried a few tricks here and all of them failed. This is hard to do for free functions with overloads and non-const ref parameters. :-/ Keeping the manual switches.

include/engine/datafacade/algorithm_datafacade.hpp
TheMarex: Yeah there are a lot of places we can just switch over, mostly legacy code.

include/engine/datafacade/contiguous_internalmem_datafacade.hpp
TheMarex: True forget to optimize this away, we needed this for plain CH.

include/engine/datafacade/contiguous_internalmem_datafacade.hpp
TheMarex: Yes. Honestly there is no point in keeping the abstract interfaces anymore. I will remove them in a different PR later.

include/engine/datafacade_provider.hpp
TheMarex: Yeah but we only do functions calls on this during initialization, not performance critical.

include/engine/datafacade/contiguous_internalmem_datafacade.hpp
TheMarex: Yes this makes it possible to use both the shared functions and the algorithm specific functions on the same object instead of needing to keep track of which is which.

include/engine/engine.hpp
TheMarex: Because we inherit from the `EngineInterface` object. We store the Engine in OSRM as a `std::unique_ptr<EngineInterface>`. When the pointer gets destructed we need to clean up OSRMs state.

include/engine/engine.hpp
TheMarex: It is more about not needing to do explicit runtime checks for deciding which implementation of `Engine<>` to use. `OSRM` need to select the algorithm at runtime right now so we can't just directly store a `std::unique_ptr<Engine<AlgorihtmT>>`.

So we store `std::unique_ptr<EngineInteface>` in `OSRM` and then are able to do `Status OSRM::Route(..) { return pimp->Route(); }`.

include/engine/engine.hpp
TheMarex: Hrm any good ideas for an alternative? I tried to come up with something better but it would either break the abstraction of `RoutingAlgorithms` (like a `GetFacade()` function) or just as ugly:
```
std::shared_ptr<datafacade::ContiguousDataFacade> facade;
RoutingAlgorithms algos;
std::tie(facade, algos) = GetInstance();
pluging->HandleRequest(facade, algos, ...);
```

include/engine/routing_algorithms.hpp
TheMarex: Nope was dead code.

include/engine/routing_algorithms.hpp
TheMarex: Hrm the sanitizer complained about this, could also make it virtual in the base class I guess.

include/engine/routing_algorithms/routing_base.hpp
TheMarex: Nope heaps are mutable, because Dijkstra.

TheMarex:
@daniel-j-h can you check if I addressed your comments and approve?

include/engine/routing_algorithms.hpp
daniel-j-h: Yeah virtual destructor is required in base class for polymorphic delete - you can even make it

    virtual ~Base() = default;

include/engine/engine.hpp
daniel-j-h: maybe a function abstracting over all of this?

    template <typename Plugin, typename Algo>
    Status Run(FacadeProvider fp, Heaps heaps, Plugin plugin, Params params, Result& ret) {
      facade = fp.get();
      algo = RoutingAlgorithms<Algo>{heaps, facade};
      return plugin.HandleRequests(facade, algo, params, result);
    }

(`HandleRequest` is not virtual in base so we need the plugin template here)

and then for each Service function do

    return Run<Table, AlgorithmT>(facade_provider, heaps, table_plugin, params, result);

Hm we could be even fancier here, all we do is a type-level loop over Service and ServiceParams

    for_each({{Table, TableParams}, {Route, RouteParams}, [](){..})

Issue #3413, opened 2016-12-07
Refactor bicycle profile


emiltin:
Should be done along the line of what was done with the car profile. The bicycle profile is more complex because it must handle pushing of bikes.

The approach will be to do things as similar as possible to the car profile, so common steps can later be factored out into helpers.

I've started on this issue.

daniel-j-h:
This landed with https://github.com/Project-OSRM/osrm-backend/pull/3654 :sparkles:

PR #3561, opened 2017-01-12
Added missing backward_speed for cycleways

oxidase:
# Issue

PR #2399 contains a fix for cycleway `backward_speed` values, but it must be merged before to have proper commit history for bisecting.

## Tasklist
 - [x] review
 - [x] adjust for comments

## Requirements / Relations
#2399

### DIFF
--- a/profiles/bicycle.lua
+++ b/profiles/bicycle.lua
@@ -369,10 +369,13 @@ function way_function (way, result)
   -- cycleways
   if cycleway and cycleway_tags[cycleway] then
     result.forward_speed = bicycle_speeds["cycleway"]
+    result.backward_speed = bicycle_speeds["cycleway"]
   elseif cycleway_left and cycleway_tags[cycleway_left] then
     result.forward_speed = bicycle_speeds["cycleway"]
+    result.backward_speed = bicycle_speeds["cycleway"]
   elseif cycleway_right and cycleway_tags[cycleway_right] then
     result.forward_speed = bicycle_speeds["cycleway"]
+    result.backward_speed = bicycle_speeds["cycleway"]
   end
 
   -- dismount


daniel-j-h:
Ouch, good catch :+1: cc @emiltin

daniel-j-h:
Backported into 5.5.4 otherwise cycleways are broken on 5.5.

emiltin:
if only cycleway_left or cycleway_right is set, then only forward or backward speed should be affected, not both?

danpat:
Ah crap, I forgot to hit "Comment" when I looked at this on Thursday.  @emiltin is correct - this change isn't quite right, it sets the forward/backward speed in cases where it shouldn't.

The root problem was that the travel mode is being set, but no speed.  This is definitely a bug, but I suspect that this change isn't quite right.

The existing logic around cycleways is a bit difficult to follow.  I think this change fixes one problem, but introduces another :-(

emiltin:
It would be good to first ensure that we have cucumber tests covering these cases. We do have tests for cycleway_left/right, but I think they only cover access, not speed.

Looking at the existing code I wonder if this is really correct:
```lua
   elseif cycleway_left and cycleway_tags[cycleway_left] then
      result.forward_speed = bicycle_speeds["cycleway"]
```

I think cycleway_left shuold modify backward_speed. Yes the existing code is hard to follow. As mentioned over in #3413 I've started on a refactor of the bike profile, along the lines of the car profile. But's it's a lot more complex that the car profile and I'm a bit low on time at the moment, so not done yet.

danpat:
After discussing with @TheMarex, this PR should be OK for side-effects.  We need to set both the mode *and* speed for an edge to be included.

This PR fixes a case where we set a mode without setting the speed, but setting the speed where the mode remains invalid should have no effect.

PR #3562, opened 2017-01-12
Use internal logger instead of std stream

oxidase:
# Issue

After f8499957fa1d5c696c0a91c78e4daefeae8427ea  incorrectly used std::cerr can lead to test fails that check output of stderr. The internal logger correctly adds EOL.

## Tasklist
 - [x] review
 - [x] adjust for comments

### DIFF
--- a/src/tools/contract.cpp
+++ b/src/tools/contract.cpp
@@ -179,8 +179,10 @@ catch (const std::bad_alloc &e)
     util::Log(logERROR) << "Please provide more memory or consider using a larger swapfile";
     return EXIT_FAILURE;
 }
+#ifdef _WIN32
 catch (const std::exception &e)
 {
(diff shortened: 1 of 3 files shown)


danpat:
@oxidase ah, I missed https://github.com/Project-OSRM/osrm-backend/pull/3498

The choice to not catch `std::exception` was deliberate - we were using it as a mechanism to automatically generate a corefile, making debugging of exceptions in production easier.

If you can think of a better mechanism for that though, I'm open to suggestions.

oxidase:
@danpat use unhandled exceptions for core dumps is ok. But it is platform-dependent: on windows a message box with an error 0xC0000409 pops up and `npm test` fails, because of no exception text is in stderr. I placed `#ifdef _WIN32` around the catch block.

Issue #3550, opened 2017-01-10
Improve osrm-contract performance - caching


torshie:
My Linux server has 200G memory, when contracting a large road network, osrm-contract uses only about 200M memory, most data are put on disk.

The performance is relatively poor (compared with a simpler CH algorithm I implemented before). I haven't profiled osrm-contract, the poor performance is likely caused by stxxl, IMO.

Add an option for end-users, so the users can disable stxxl for better performance if large amount of memory is available.

I can provide some help in my spare time if needed, but guidelines maybe needed, because I'm not familiar with the code base.

daniel-j-h:
Care to switch in a custom stxxl config with the memory fileio and report back?

http://stxxl.sourceforge.net/tags/master/install_config.html

> memory : keeps all data in RAM, for quicker testing

TheMarex:
The out-of-memory storage should be even less then the process memory storage. `osrm-contract` only writes to disk exactly once at about 70% of the contraction phase to rebuild the graph and remove nodes that are already contracted. 200MB for `osrm-contract` is very unusual - for road networks like Europe/North-America we are talking about 30GB or more.

All graph data structures actually needed for contraction are kept in main memory. Are you sure you are talking about `osrm-contract` and not `osrm-extract`?

torshie:
@daniel-j-h I did a quit & dirty hack: removed all stxxl dependencies in CMakeLists.txt, and replaced stxxl::vector with std::vector. Then recompiled osrm-contract, I didn't see any improvements. I seems my guess was wrong, stxxl wasn't the bottleneck, I'll do some profiling then report back later.

@TheMarex Yes, I'm talking about osrm-contract. I was contracting road network of a big city, so it's much smaller than Europe/North-America. 200M is reasonable in my case. 

BTW, according to my personal CH implementing experience, 30GB to contract a road network of a country is surprisingly small. My simpler CH algorithm needed about 150G+ memory to contract a graph of 54M edges + 20M nodes, time cost was 2.5 hours. My CH didn't build the edge expanded graph, so it's not exactly an apples to apples comparison

daniel-j-h:
Thanks for following up and clarifying the issue - I'm interested to see what you can find. There is an open issue (https://github.com/Project-OSRM/osrm-backend/issues/3331) with contract being slow in some specific cases.

[...]

PR #3564, opened 2017-01-13
refactor contractor code for readability

MoKob:
# Issue
Working towards testing the performance cache described in https://github.com/Project-OSRM/osrm-backend/issues/3550#issuecomment-272201095. Right now we have a single heap per thread, sized by the graph.

Edit: Turned out that the cache does just not work for us. We would need immense amounts of memory that just outweigh the benefits.
We are looking at a quadratic number of caches (src/forbidden), which is simply just too large to keep around.

The refactor in itself is something that I feel still is valuable, since it increases the readability of the code a bit.

 - [x] review
 - [x] adjust for comments

## Requirements / Relations
 none

### DIFF
--- a/include/contractor/contractor_dijkstra.hpp
+++ b/include/contractor/contractor_dijkstra.hpp
@@ -1,0 +1,50 @@
+#ifndef OSRM_CONTRACTOR_DIJKSTRA_HPP
+#define OSRM_CONTRACTOR_DIJKSTRA_HPP
+
+#include "contractor/contractor_graph.hpp"
+#include "contractor/contractor_heap.hpp"
+#include "util/typedefs.hpp"
+
+#include <cstddef>
+
+namespace osrm
+{
+namespace contractor
+{
+
+// allow access to the heap itself, add Dijkstra functionality on top
+class ContractorDijkstra
+{
+  public:
+    ContractorDijkstra(std::size_t heap_size);
+
+    // search the graph up
+    void Run(const unsigned number_of_targets,
+             const int node_limit,
+             const int weight_limit,
+             const NodeID forbidden_node,
+             const ContractorGraph &graph);
+
+    // adaption of the heap interface
+    void Clear();
+    bool WasInserted(const NodeID node) const;
+    void Insert(const NodeID node,
+                const ContractorHeap::WeightType weight,
+                const ContractorHeap::DataType &data);
+
+    // cannot be const due to node-hash access in the binary heap :(
+    ContractorHeap::WeightType GetKey(const NodeID node);
+
+  private:
+    void RelaxNode(const NodeID node,
+                   const int node_weight,
+                   const NodeID forbidden_node,
+                   const ContractorGraph &graph);
+
+    ContractorHeap heap;
+};
+
+} // namespace contractor
+} // namespace osrm
+
+#endif // OSRM_CONTRACTOR_DIJKSTRA_HPP
--- a/include/contractor/graph_contractor_adaptors.hpp
+++ b/include/contractor/graph_contractor_adaptors.hpp
@@ -1,0 +1,59 @@
+#ifndef OSRM_CONTRACTOR_GRAPH_CONTRACTION_ADAPTORS_HPP_
+#define OSRM_CONTRACTOR_GRAPH_CONTRACTION_ADAPTORS_HPP_
+
+#include "contractor/contractor_graph.hpp"
+#include "util/log.hpp"
+
+#include <vector>
+
+namespace osrm
+{
+namespace contractor
+{
+
+// Make sure to move in the input edge list!
+template <typename InputEdgeContainer>
+std::vector<ContractorEdge> adaptToContractorInput(InputEdgeContainer input_edge_list)
+{
+    std::vector<ContractorEdge> edges;
+    edges.reserve(input_edge_list.size() * 2);
+
+    for (const auto &input_edge : input_edge_list)
+    {
+#ifndef NDEBUG
+        const unsigned int constexpr DAY_IN_DECI_SECONDS = 24 * 60 * 60 * 10;
+        if (static_cast<unsigned int>(std::max(input_edge.weight, 1)) > DAY_IN_DECI_SECONDS)
+        {
+            util::Log(logWARNING) << "Edge weight large -> "
+                                  << static_cast<unsigned int>(std::max(input_edge.weight, 1))
+                                  << " : " << static_cast<unsigned int>(input_edge.source) << " -> "
+                                  << static_cast<unsigned int>(input_edge.target);
+        }
+#endif
+        edges.emplace_back(input_edge.source,
+                           input_edge.target,
+                           static_cast<unsigned int>(std::max(input_edge.weight, 1)),
+                           1,
+                           input_edge.edge_id,
+                           false,
+                           input_edge.forward ? true : false,
+                           input_edge.backward ? true : false);
+
+        edges.emplace_back(input_edge.target,
+                           input_edge.source,
+                           static_cast<unsigned int>(std::max(input_edge.weight, 1)),
+                           1,
+                           input_edge.edge_id,
+                           false,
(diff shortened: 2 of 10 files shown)


torshie:
We don't need quadratic number of caches, just linear to number of nodes (with large constant factor though):

For node `A`, we need to save all 1000 nodes (`N[0]...N[999]`) settled during the Dijkstra search (a normal Dijkstra search without any forbidden node), for each settled node `N[i]` only two integers are required: 1) the node ID `N[i].id`, 2) the offset `N[i].previous` where `N[N[i].previous]` is the "previous node" on the shortest path from `A` to node `N[i]`, so the cache size is 8kB per node. `N[0]...N[999]` are sorted by `N[i].id`.

The cache can be used to find out whether the shortest path from `A` to node `X` passes a specific "forbidden node" `Y` very quickly: 

- do a binary search, and find the offset `i` where `N[i].id == X.id`
- let `i = N[i].previous`
- do the previous step repeatedly until either `N[i].id == Y.id` or `N[i].id == A.id`

I suggest do a test with only 100 or fewer settled nodes per Dijkstra search on planet sized network, because it did work on country sized network.

Small Dijkstra search space only has significant impact on shortcut accuracy when the graph is dense. For 100 settled nodes, only when the average degree is 10+. 10+ average degree usually means 95% of nodes are already contracted.

You can check the online course I mentioned previously in #3550 http://ad-wiki.informatik.uni-freiburg.de/teaching/EfficientRoutePlanningSS2012 (lecture 6 & 7). The professor Dr. Hannah Bast is an expert on this topic.

MoKob:
@torshie your assumption is true, if you only consider road networks. However, OSRM is a more general tool. Using your way of storing searches is only possible if you assume few paths share the same length.

No consider someone using a walk profile without turn penalties and a grid city like Manhatten or pretty much any other city in the us. Consider roads with sidewalks and connections in between. Walking speed along any of these will be the same, turning penalties on a board walk may or may not make sense, but usually I would assume them to be the same (omitting traffic crossing).

For simplicity lets stay with the grid model:

```
a - - b - - c
|     |     |
d - - e - - f
|     |     |
g - - h - - i
```

And lets start contracting `e`. Whether we find the path using `e` to get to `b` or `h` is essentially random, since the path `dab` and `dgh` are just as valid and (given the complexity of the search) it is not simple to decide in general which of the equal paths to keep.
Lets assume we find the path using `e`, we now add additional shortcuts to all the surrounding nodes:

```
a - - b - - c
|  /     \  |
d - - - - - f
|  \     /  |
g - - h - - i
```

These new connections now have to be considered in any other setting. For example if we contract `a`, we would add a connection to `h` and so on... The danger of this snowballing based on our profiles and turn penalties is a just too large. It is unlikely on a road network, but entirely possible in many set-ups that we need to handle.

I know Professor Bast personally and have worked with her, her assumption is certainly a sane way to go if you consider very specific networks only. As long as you want to offer customisable profiles on an OSM database, we need to consider what people can allow. And in case of many similar routes, the random element of shortcuts can not be neglected. Searching without blocking the node can easily result in random generation of unnecessary shortcuts and is not an option in my view, as long as we don't ensure in the engine that these will not happen often.
For example, using Berlin on our walk profile, we can already see an increase in shortcuts from 3988461 to 4305589, when using a search like you propose. This is an increase of 8 percent, and will only get worse on larger graphs.

Now even if we did assume this would not happen, for some of our graphs we would require an additional 661256722 (graph) * 100 (search space) * 8 (byte per node) / 1024/1024/1024 = 492GB of memory for the cache, which is just not possible. When we are not able to do it on a one-search-per-node basis, the amount of nodes we can cache is very small. Even though we usually only use a very small number of nodes in our searches (Berlin usually uses well less than a hundred), the benefit will only become apparent in later stages of the graph. Since all caches can be re-used at most once (and are extended to find less shortcuts), I still don't think that caching is the way to go for us.

You are welcome to submit a pull request though, if you feel that you can guarantee that the changes that you mentioned don't impact profiles as badly as my initial experiments show. For now, a 10 percent increase in shortcuts on Berlin is just something that strikes me a real danger if you consider the world. Berlin is usually a very simple set-up to begin with and seeing increases lie this on a graph that small, I would expect far larger increases.

src/contractor/contractor.cpp
```diff
@@ -307,7 +308,7 @@ int Contractor::Run()
 
     util::Log() << "Loading edge-expanded graph representation";
 
-    util::DeallocatingVector<extractor::EdgeBasedEdge> edge_based_edge_list;
+    std::vector<extractor::EdgeBasedEdge> edge_based_edge_list;
```
TheMarex: Any particular reason why you switched from `DeallocatingVector` to `std::vector`? My last status on this was that `DeallocatingVector` saved some memory, but I never evaluated if this is still true.

include/contractor/graph_contractor_adaptors.hpp
```diff
@@ -0,0 +1,58 @@
...
+#include <vector>
+
+namespace osrm
+{
+namespace contractor
+{
+
+// Make sure to move in the input edge list!
+template <typename InputEdgeContainer>
+std::vector<ContractorEdge> ContractorEdgeInputAdaptor(InputEdgeContainer input_edge_list)
```
TheMarex: We use lower-case first letter for free functions. Should probably be named `toContractorEdge`, or at least something that is not a noun which is usually reserved for classes.

include/contractor/graph_contractor_adaptors.hpp
```diff
@@ -0,0 +1,58 @@
...
+{
+    std::vector<ContractorEdge> edges;
+    edges.reserve(input_edge_list.size() * 2);
+
+    for (const auto &input_edge : input_edge_list)
+    {
+#ifndef NDEBUG
+        if (static_cast<unsigned int>(std::max(input_edge.weight, 1)) > 24 * 60 * 60 * 10)
```
TheMarex: Not scope of this PR just a though: Can we replace this by a real overflow check?

src/contractor/contractor.cpp
```diff
@@ -334,12 +335,17 @@ int Contractor::Run()
...
 
     util::DeallocatingVector<QueryEdge> contracted_edge_list;
-    ContractGraph(max_edge_id,
-                  edge_based_edge_list,
-                  contracted_edge_list,
-                  std::move(node_weights),
-                  is_core_node,
-                  node_levels);
+    { // own scope to not keep the contractor around
+        GraphContractor graph_contractor(
```
TheMarex: Not part of the PR scope: What about making the GraphContractor into a function? Having this `Constructor(), Run(), Get()` paradigmatic always seemed kinda weird to me here, since you actually can't call `Run()` more then once.

include/contractor/contractor_dijkstra.hpp
```diff
@@ -0,0 +1,39 @@
...
+
+#include <cstddef>
+
+namespace osrm
+{
+namespace contractor
+{
+
+// allow access to the heap itself, add Dijkstra functionality on top
+class ContractorDijkstra : public ContractorHeap
```
TheMarex: This is kind of a inheritance misuse IMHO. An algorithm and a data structure don't share the same concept. It would be cleaner to make this a member and wrap access if needed.

src/contractor/contractor.cpp
MoKob: This is not true anymore, since the vector is moved into the function (==destroyed for the outside world) and the scope takes care of the deallocation.
The only reason to have a `DeallocatingVector` is locations that don't deallocate by scope.

include/contractor/graph_contractor_adaptors.hpp
MoKob: What kind of check would you have in mind? The entire edge-weight check is against a more or less random value to me. It's edges that take more than a day in time, but its not technically an overflow. The only thing that I could see is improving the logic to give a meaning to `24*60*60*10` in form of a `constexpr`.

PR #3565, opened 2017-01-13
Minor fixes

seemoritz:
I provide fixes for some minor issues, including a bug in DeallocatingVector::back, a possible integer overflow in GraphContractor::ContractorEdgeData, the adding of duplicate edges in GraphContractor::Run and a bug in StorageConfig::IsValid.
I also cleaned up some misleading if-statements and added a few comments, where I found the code unclear.

### DIFF
--- a/include/contractor/graph_contractor.hpp
+++ b/include/contractor/graph_contractor.hpp
@@ -402,7 +402,7 @@ class GraphContractor
 
         unsigned current_level = 0;
         bool flushed_contractor = false;
-        while (number_of_nodes > 2 &&
+        while (remaining_nodes.size() > 1 &&
                number_of_contracted_nodes < static_cast<NodeID>(number_of_nodes * core_factor))
         {
             if (!flushed_contractor && (number_of_contracted_nodes >
@@ -513,16 +513,20 @@ class GraphContractor
                 {
                     const EdgeID current_edge_ID =
                         contractor_graph->FindEdge(edge.source, edge.target);
-                    if (current_edge_ID < contractor_graph->EndEdges(edge.source))
+                    if (current_edge_ID != SPECIAL_EDGEID)
                     {
                         ContractorGraph::EdgeData &current_data =
                             contractor_graph->GetEdgeData(current_edge_ID);
                         if (current_data.shortcut && edge.data.forward == current_data.forward &&
-                            edge.data.backward == current_data.backward &&
-                            edge.data.weight < current_data.weight)
+                            edge.data.backward == current_data.backward)
                         {
-                            // found a duplicate edge with smaller weight, update it.
-                            current_data = edge.data;
+                            // we found a duplicate edge
+                            // if weight of new one is smaller, update old one
+                            if (edge.data.weight < current_data.weight)
+                            {
+                                current_data = edge.data;
+                            }
+                            // don't insert duplicates
                             continue;
                         }
                     }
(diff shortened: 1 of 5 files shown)


include/contractor/graph_contractor.hpp
```diff
@@ -47,7 +47,7 @@ class GraphContractor
                            bool shortcut,
                            bool forward,
                            bool backward)
-            : weight(weight), id(id), originalEdges(std::min((unsigned)1 << 28, original_edges)),
+            : weight(weight), id(id), originalEdges(std::min((1u << 28) - 1u, original_edges)),
```
daniel-j-h: cc @MoKob who is refactoring this code right now anyway - from a first sight it looks like you're changing semantic here - why?

include/contractor/graph_contractor.hpp
```diff
@@ -336,15 +336,15 @@ class GraphContractor
...
         // a lot of data. It is *important* for the last loop iterations
         // (which have a very small dataset) that it is devisible.
-        const constexpr size_t IndependentGrainSize = 1;
-        const constexpr size_t ContractGrainSize = 1;
-        const constexpr size_t NeighboursGrainSize = 1;
-        const constexpr size_t DeleteGrainSize = 1;
+        constexpr size_t IndependentGrainSize = 1;
+        constexpr size_t ContractGrainSize = 1;
+        constexpr size_t NeighboursGrainSize = 1;
+        constexpr size_t DeleteGrainSize = 1;
```
daniel-j-h: Why removing const here? constexpr is not const by default in C++14 +

src/storage/storage_config.cpp
```diff
@@ -32,7 +32,7 @@ bool StorageConfig::IsValid() const
                                                       core_data_path,
                                                       geometries_path,
                                                       timestamp_path,
-                                                      datasource_indexes_path,
+                                                      datasource_names_path,
```
daniel-j-h: Ouch - fortunately it's only the check for missing files

include/contractor/graph_contractor.hpp
seemoritz: Oh, my bad. I was sure that constexpr implies const, and found it more readable.
It's one commit where I did this change, should I revert it?

include/contractor/graph_contractor.hpp
seemoritz: This std::min seems to be for prevention of overflow, since originalEdges became a 28 bit field.
But if you assign 1u<<28 you will acually assign 0 ...

include/util/deallocating_vector.hpp
```diff
@@ -356,8 +356,8 @@ class DeallocatingVector
 
     ElementT &back() const
     {
-        const std::size_t _bucket = current_size / ELEMENTS_PER_BLOCK;
-        const std::size_t _index = current_size % ELEMENTS_PER_BLOCK;
+        const std::size_t _bucket = (current_size - 1) / ELEMENTS_PER_BLOCK;
```
oxidase: :+1: but better `return operator[](current_size - 1);`

include/contractor/graph_contractor.hpp
seemoritz: Hmm ... I'm still a little confused.
What I found on the subject is this:
https://isocpp.org/wiki/faq/cpp14-language#extended-constexpr

Where it says:
C++14 also removes the C++11 rule that constexpr member functions are implicitly const.

But that does not apply to local variables.

MoKob:
This will have to be adjusted to the refactored version, but from what I can tell, these changes look sane.

Issue #3457, opened 2016-12-15
Refactor `struct Fork`


chaupow:
In `turn_handler.hpp` is a struct Fork that represents a fork at an intersection.

```
struct Fork
{
    Intersection::iterator right;
    Intersection::iterator left;
    const std::size_t size;
    Fork(Intersection::iterator right, Intersection::iterator left);
};
```

Currently all the `Connected Roads` between right and left (including left) are part of the road. 

Change right and left to `begin` and `end` and end being the iterator to the first road that is _not_ part of the road to make it more stdlib compatible.

(see https://github.com/Project-OSRM/osrm-backend/pull/3264)

PR #3568, opened 2017-01-16
Refactor `struct Fork` in turn_handler

chaupow:
# Issue

Works towards #3457 to refactor the `struct Fork` that represents a Fork:
Before it was:
```
struct Fork
{
    const Intersection::iterator right;
    const Intersection::iterator left;
    const std::size_t size;
};
```
where `*right` and `*left` returned the `ConnectedRoad` that are the left or the right roads in a fork. Now the struct is as follows:
```
struct Fork
{
    const Intersection::iterator begin;
    const Intersection::iterator end;
    const std::size_t size;
};
```
where `*begin` returns the right road in a fork and `*(end-1)` returns the left road in a fork.

Reason: `begin` and `end` are more intuitive when talking about iterators and more std::lib conform.

I kind of think this is a sensible change. Sometimes not sure whether this is over-engineering or really needed though? Also, most of the times a Fork only consists of left and right in which case and begin and end iterator is not that intuitive? But using iterators it is often not necessary to differentiate Forks with two or three roads.

@MoKob @daniel-j-h 🙇‍♀️  

## Tasklist
 - [ ] review
 - [ ] adjust for comments

### DIFF
--- a/src/extractor/guidance/turn_handler.cpp
+++ b/src/extractor/guidance/turn_handler.cpp
@@ -52,12 +52,12 @@ bool isEndOfRoad(const ConnectedRoad &,
 }
 
 template <typename InputIt>
-InputIt findOutermostForkCandidate(const InputIt start, const InputIt end)
+InputIt findOutermostForkCandidate(const InputIt begin, const InputIt end)
 {
     static_assert(std::is_base_of<std::input_iterator_tag,
                                   typename std::iterator_traits<InputIt>::iterator_category>::value,
                   "findOutermostForkCandidate() only accepts input iterators");
-    const auto outermost = std::adjacent_find(start, end, isOutermostForkCandidate);
+    const auto outermost = std::adjacent_find(begin, end, isOutermostForkCandidate);
     if (outermost != end)
     {
         return outermost;
@@ -78,12 +78,36 @@ namespace guidance
 {
 
 // a wrapper to handle road indices of forks at intersections
-TurnHandler::Fork::Fork(const Intersection::iterator right, const Intersection::iterator left)
-    : right(right), left(left), size((left - right) + 1)
+TurnHandler::Fork::Fork(const Intersection::iterator intersection_base,
+                        const Intersection::iterator begin,
+                        const Intersection::iterator end)
+    : intersection_base(intersection_base), begin(begin), end(end), size(std::distance(begin, end))
 {
-    BOOST_ASSERT(right < left);
-    BOOST_ASSERT(size >= 2);
-    BOOST_ASSERT(size <= 3);
+    BOOST_ASSERT(begin < end);
+    BOOST_ASSERT(size == 2 || size == 3);
+}
+
+ConnectedRoad &TurnHandler::Fork::getRight() { return *begin; }
+ConnectedRoad &TurnHandler::Fork::getLeft() { return *(end - 1); }
+ConnectedRoad &TurnHandler::Fork::getMiddle()
+{
+    BOOST_ASSERT(size == 3);
+    return *(begin + 1);
+}
+ConnectedRoad &TurnHandler::Fork::getRight() const { return *begin; }
+ConnectedRoad &TurnHandler::Fork::getLeft() const { return *(end - 1); }
+ConnectedRoad &TurnHandler::Fork::getMiddle() const
+{
+    BOOST_ASSERT(size == 3);
+    return *(begin + 1);
+}
+std::size_t TurnHandler::Fork::getRightIndex() const
+{
+    return std::distance(intersection_base, begin);
+}
+std::size_t TurnHandler::Fork::getLeftIndex() const
+{
+    return std::distance(intersection_base, end) - 1;
 }
 
 TurnHandler::TurnHandler(const util::NodeBasedDynamicGraph &node_based_graph,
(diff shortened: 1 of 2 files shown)


src/extractor/guidance/turn_handler.cpp
```diff
@@ -78,14 +78,32 @@ namespace guidance
 {
 
 // a wrapper to handle road indices of forks at intersections
-TurnHandler::Fork::Fork(const Intersection::iterator right, const Intersection::iterator left)
-    : right(right), left(left), size((left - right) + 1)
+TurnHandler::Fork::Fork(const Intersection::iterator intersection_base,
+                        const Intersection::iterator begin,
+                        const Intersection::iterator end)
+    : intersection_base(intersection_base), begin(begin), end(end), size(end - begin)
```
MoKob: should be fine, but I'd consider using `std::distance(begin,end)` here. Not really needed, but cleaner for different iterator types.

src/extractor/guidance/turn_handler.cpp
```diff
@@ -78,14 +78,32 @@ namespace guidance
...
+}
+ConnectedRoad &TurnHandler::Fork::getRight() const { return *begin; }
+ConnectedRoad &TurnHandler::Fork::getLeft() const { return *(end - 1); }
+ConnectedRoad &TurnHandler::Fork::getMiddle() const
+{
+    BOOST_ASSERT(size == 3);
+    return *(begin + 1);
+}
+std::size_t TurnHandler::Fork::getRightIndex() const { return begin - intersection_base; }
+std::size_t TurnHandler::Fork::getLeftIndex() const { return end - intersection_base; }
```
MoKob: I haven't checked the usage yet. From the name I would assume that you want `end - 1 - intersection_base`. Is this correct here, or actually a typo?

src/extractor/guidance/turn_handler.cpp
```diff
@@ -191,15 +209,12 @@ bool TurnHandler::isObviousOfTwo(const EdgeID via_edge,
...
-    {
-        if (isObviousOfTwo(via_edge, *road, *(road + 1)) ||
-            isObviousOfTwo(via_edge, *(road + 1), *road))
-        {
-            return true;
-        }
-    }
-    return false;
+    auto obvious_road = std::adjacent_find(
+        fork.begin, fork.end, [&](const ConnectedRoad &a, const ConnectedRoad &b) {
```
MoKob: beautiful. Is is possible to even use `auto` here for the parameters, then we would be even more flexible in the future.

src/extractor/guidance/turn_handler.cpp
```diff
@@ -632,12 +644,11 @@ boost::optional<TurnHandler::Fork> TurnHandler::findFork(const EdgeID via_edge,
...
         // right side
-        const auto next =
-            (fork->left + 1) == intersection.end() ? intersection.begin() : (fork->left + 1);
+        const auto next = fork->end == intersection.end() ? intersection.begin() : (fork->end);
         const bool separated_at_left_side =
-            angularDeviation(fork->left->angle, next->angle) >= GROUP_ANGLE;
+            angularDeviation(fork->getLeft().angle, next->angle) >= GROUP_ANGLE;
         const bool separated_at_right_side =
-            angularDeviation(fork->right->angle, (fork->right - 1)->angle) >= GROUP_ANGLE;
+            angularDeviation(fork->getRight().angle, (fork->begin - 1)->angle) >= GROUP_ANGLE;
```
MoKob: are we sure that `begin` is safe to subtract from here? Should be due to the u-turn, right?

src/extractor/guidance/turn_handler.cpp
```diff
@@ -78,14 +78,32 @@ namespace guidance
...
 {
-    BOOST_ASSERT(right < left);
-    BOOST_ASSERT(size >= 2);
-    BOOST_ASSERT(size <= 3);
+    BOOST_ASSERT(begin < end);
+    BOOST_ASSERT(size == 2 || size == 3);
 }
 
+ConnectedRoad &TurnHandler::Fork::getRight() { return *begin; }
+ConnectedRoad &TurnHandler::Fork::getLeft() { return *(end - 1); }
```
chaupow: The Fork constructor makes sure this doesn't happen (`BOOST_ASSERT(begin < end);`)

src/extractor/guidance/turn_handler.cpp
```diff
@@ -78,14 +78,32 @@ namespace guidance
...
+ConnectedRoad &TurnHandler::Fork::getMiddle()
+{
+    BOOST_ASSERT(size == 3);
+    return *(begin + 1);
+}
+ConnectedRoad &TurnHandler::Fork::getRight() const { return *begin; }
+ConnectedRoad &TurnHandler::Fork::getLeft() const { return *(end - 1); }
```
chaupow: The Fork constructor makes sure this doesn't happen (`BOOST_ASSERT(begin < end);`)

src/extractor/guidance/turn_handler.cpp
chaupow: Should be safe due to u-turn. Assertion added.

src/extractor/guidance/turn_handler.cpp
chaupow: Changing to `auto` parameters, code is not compilable on all compilers 😞 
Error message says `void value not ignored as it ought to be`.

Changed back to ConnectedRoad params

daniel-j-h: Hm strange, this is the case with `[&](const auto& l, const auto& r) { }`? Which compilers fail here?

MoKob: I'd be curious about this as well, especially as `void` seems a bit strange.

chaupow:
@daniel-j-h @MoKob this is the build that failed:
https://travis-ci.org/Project-OSRM/osrm-backend/builds/192404012

src/extractor/guidance/turn_handler.cpp
daniel-j-h: Capturing (heh) from chat: you have to capture `this` for being able to call a member function.

All member function calls `memfn` are implicitely `this->memfn()`.

Issue #1602, opened 2015-08-18
Highlighting small components at inaccessible nodes


planemad:
Came across an unroutable junction where all the oneway roads were leading out of it. 

<img width="626" alt="screenshot 2015-08-18 15 31 44" src="https://cloud.githubusercontent.com/assets/126868/9327898/d492d958-45c1-11e5-94cd-c4c8650d489d.png">
http://osrm.at/f2j

It took some time to figure out where the issue was since the small component layer had no segment to highlight. It would really help to highlight nodes as well with which we can easily spot such cases and those where there are barriers with access restrictions.

cc @TheMarex

daniel-j-h:
https://github.com/Project-OSRM/osrm-backend/pull/3570 adds GeoJSON from / to / type properties to the components output. You can use these to style your map. Closing as resolved. If you need more properties or a different feature set, feel free to re-open or open a new ticket.

PR #3570, opened 2017-01-16
Rewrites the osrm-components tool to dump GeoJSON, resolves #2176 #1738 #1602

### LINKED ISSUE #2176: Remove osrm-components
We added support for small/big component flags in the `tile` plugins. That means we don't have to generate that with a separate tool but can serve it at runtime from the demo server.
- [x] Fix style on `map.project-osrm.org/debug` to display small components
- [ ] Add link to go from route to debug map (`osrm-frontend`)
- [ ] Remove `osrm-components`

daniel-j-h:
Resolves #2176 #1738 #1602.
Unblocks `node-osrm` from bundling `osrm-components` https://github.com/Project-OSRM/node-osrm/issues/283.

On Monaco, unstyled raw dump: http://bl.ocks.org/d/9d6265b064c0f29a71503e0f6527769d

Would love some input here: at the moment I expose OSM ids as properties for the `LineString`'s from and to node. Is this helpful and are these use-cases where osm node ids help here? Otherwise I would remove it, since it bloats the file quite a bit.

cc @MoKob @geohacker @planemad @srividyacb @oini

## Tasklist
 - [x] review
 - [x] adjust for comments
 - [x] decide on node osm id properties
 - [x] check on planet how large the new geojson file will get (~2 GB, ~100 MB compressed)
 - [x] adapt demo server integration
 - [x] notify Fred from Geofabrik and figure out new integration with them

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -435,10 +435,6 @@ if(ENABLE_MASON)
   set(CMAKE_SHARED_LINKER_FLAGS "${CMAKE_SHARED_LINKER_FLAGS} ${LINKER_FLAGS}")
   set(CMAKE_MODULE_LINKER_FLAGS "${CMAKE_MODULE_LINKER_FLAGS} ${LINKER_FLAGS}")
 
-  if(BUILD_COMPONENTS)
-    message(FATAL_ERROR "BUILD_COMPONENTS is not supported with ENABLE_MASON")
-  endif()
-
   # current mason packages target -D_GLIBCXX_USE_CXX11_ABI=0
   set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -D_GLIBCXX_USE_CXX11_ABI=0")
 
@@ -468,10 +464,6 @@ else()
 
   find_package(BZip2 REQUIRED)
   add_dependency_includes(${BZIP2_INCLUDE_DIR})
-
-  if(BUILD_COMPONENTS)
-    find_package(GDAL)
-  endif()
 
   FIND_PACKAGE(Lua 5.2 EXACT)
   IF (LUA_FOUND)
@@ -622,15 +614,9 @@ target_link_libraries(osrm_extract ${EXTRACTOR_LIBRARIES})
 target_link_libraries(osrm_store ${STORAGE_LIBRARIES})
 
 if(BUILD_COMPONENTS)
-  if(GDAL_FOUND)
-    add_executable(osrm-components src/tools/components.cpp $<TARGET_OBJECTS:UTIL>)
-    target_link_libraries(osrm-components ${TBB_LIBRARIES})
-    include_directories(SYSTEM ${GDAL_INCLUDE_DIR})
-    target_link_libraries(osrm-components ${GDAL_LIBRARIES} ${BOOST_BASE_LIBRARIES})
-    install(TARGETS osrm-components DESTINATION bin)
-  else()
-    message(WARNING "libgdal and/or development headers not found")
-  endif()
+  add_executable(osrm-components src/tools/components.cpp $<TARGET_OBJECTS:UTIL>)
+  target_link_libraries(osrm-components ${TBB_LIBRARIES} ${BOOST_BASE_LIBRARIES})
+  install(TARGETS osrm-components DESTINATION bin)
 endif()
 
 if(BUILD_TOOLS)
--- a/src/tools/components.cpp
+++ b/src/tools/components.cpp
@@ -1,8 +1,7 @@
 #include "extractor/tarjan_scc.hpp"
+#include "util/coordinate.hpp"
 #include "util/coordinate_calculation.hpp"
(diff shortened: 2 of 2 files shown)


daniel-j-h:
The previous behavior (and what I replicated so far) is to highlight all ways inside small components.

What I want to do in addition is to highlight segments where

    component(source) != component(target)

and one of the components being a small one. This can be used to more easily find the edges going into or out of the small components (if such edges exist at all).

daniel-j-h:
Now adds `"type": "inner"`, `"type": "border"` to let users style edges completely inside the small component vs edges crossing the small component border. Should be helpful for gated areas etc.

http://bl.ocks.org/d/63117690b40e48e0105b437dc72aaed5

srividyacb:
@daniel-j-h Looking at the Monaco Dump of small-components, the id stored is `node-id` present in the `LineString`. What are `from` and `to` node-ids? 

Is there a way we can keep the `way-id` ? This will be easy if we wish to open these features from a map to any of the editors.

daniel-j-h:
We don't store way ids in any form, all we have at that point are osm node ids.

At the moment, the `LineString`s have properties:

```json
"from": 1,
"to": 2,
"type": "border"
```

- `from` is the way segment's starting osm node id
- `to` is the way segments's ending osm node id
- `type` is either `border` for a way segment crossing components or `inner` for a way segment starting and ending inside the same component

We can talk about the properties here or how to represent the way segments in GeoJSON. I see no way to add osm way ids though.

daniel-j-h:
Here's an example, styled based on the `inner` vs `border` type. Not really that helpful for monaco, but could help for gated areas, airports and so on.

- inner (inside the same component): orange
- border (crossing two components): teal

map


https://api.mapbox.com/styles/v1/danieljh/ciy2rfz1b00gx2sl5v4nvsfkr.html?title=true&access_token=pk.eyJ1IjoiZGFuaWVsamgiLCJhIjoiTnNYb25JSSJ9.vYOnsuu1zeKcGW2nj0uJZw#13.8/43.738447/7.430143/0

daniel-j-h:
@geohacker @oini @srividyacb - could you check if these modifications to the `osrm-component` tool suit your needs. If that's fine I'd like to merge it and bundle it with `node-osrm` for the next release. I also just ran the tool on the planet and it produces a ~2 GB GeoJSON file, compresses down to ~100 MB.

cc @woodpeck for context on ticket

woodpeck:
For OSMI integration, if you can create and transmit the GeoJSON, I'd probably simply convert to shape on my end and dump into OSMI as before. If creating and uploading the file is an extra burden for you then I could probably also run that somewhere on OSMI.

daniel-j-h:
@woodpeck not sure if you are aware of it but we are trying out the following for a more realistic small components layer: use a custom profile and enable routing over tags such as `access=destination` and others which are routing engine limitations but not data issues.

The small components layer then shows data issues not influenced by the absence of features in the routing engine. Here is this experimental connectivity profile: https://github.com/Project-OSRM/osrm-backend/pull/3551

The demo server will still continue to run the regular car profile since we don't want the connectivity profile for routing. I'm not sure yet how we should handle this split: we would need to keep the connectivity profile in sync and run it regularly for small components. If you think you can run this on your servers that would be great (osrm-extract + osrm-components is all you need). I just fear this might accidentally go out of sync. What's your take here?

woodpeck:
Hm, I'm not too deep into OSRM development but are there many things that actually need to be kept in sync? I'd guess that tuning on the standard profiles will mostly be changing a speed here, changing the handling of gates there, and so on; if the connectivity profile just looks at the bare road layout then I can't see many things that actually would need syncing?

daniel-j-h:
With syncing I mostly mean you running the latest OSRM release and us keeping the profile changes here https://github.com/Project-OSRM/osrm-backend/pull/3551 in sync with master.

The profile changes shouldn't be too hard to keep in sync with master. I don't know how often you want and can update your OSRM instance. We provide Docker images and pre-built binaries via npm, so installation does not require compiling from source.

What I really want to avoid is having 1/ a stale OSRM installation creating this layer and 2/ a stale components layer which we then show to the OpenStreetMap community.

woodpeck:
Updating the software frequently should be the easier bit; as for updating the data, I'd aim at daily but we'd have to see how it turns out, performance-wise...

TheMarex:
@daniel-j-h from my side this is ready to merge, anything else left?

daniel-j-h:
api-osrm-routed needs to be adapted, otherwise we break the next cycling phase - will take a look now.

daniel-j-h:
Demo server integration adapted, updated and cycled anew.

daniel-j-h:
@woodpeck the demo server will now no longer upload the car small components layer - please set up your components pipeline to create the layer from the connectivity profile (see https://github.com/Project-OSRM/osrm-backend/pull/3551) which does not show routing engine issues.

Issue #3516, opened 2017-01-04
Rounding problem in matching tracepoints location


gardster:
For the matching request

http://router.project-osrm.org/match/v1/car/-73.99039000003845,40.75171000001269;-73.99025999998277,40.75148000000408?radiuses=26;10&annotations=true&steps=true&geometries=geojson&overview=full

I got in geometry:

>"geometry": {
>        "coordinates": [
>          [
>            -73.990134,
>            40.75165 _**8**_
>          ],
>          [
>            -73.990263,
>            40.751481
>          ]
>        ]

and in tracepoints:

>tracepoints": [
>    {
>      "waypoint_index": 0,
>      "location": [
>        -73.990134,
>        40.75165 _**7**_
>      ],
>      "name": "West 35 Street",
>      "hint": "QAZGgP___38oKRIAfwEAAH8BAAAaAAAAAAAAANxENADf5gAACgCX-ynSbQIK_5b7XtJtAgEAAQHEwdXo",
>      "matchings_index": 0
>    }

Must not these coordinates be similar? 
When I have several legs in matching answer, how do I get tracepoints this leg lies between?

TheMarex:
> Must not these coordinates be similar?

Yes they should be the same. This looks like some weird rounding bug to me that needs investigating.

> When I have several legs in matching answer, how do I get tracepoints this leg lies between?

You will need to build a mapping `leg idx -> tracepoints idx` which you can do over iterating over `tracepoints` and do something like `to_start_index[tracepoint.matchings_idx][tracepoint.waypoint_index] = idx`,
`to_target_index[tracepoint.matchings_idx][tracepoint.waypoint_index-1] = idx` (except for `tracepoint.waypoint_index == 0` which is of course only a start point).

TheMarex:
This was fixed by #3578.

PR #3576, opened 2017-01-17
Fix for #3516

oxidase:
# Issue

Analysis for #3516:
Coordinates in the response are different because they come from different sources at https://github.com/Project-OSRM/osrm-backend/blob/master/src/engine/plugins/match.cpp#L219 and there is no guarantee that they can be equal.

* `geometry.coordinates[0]` -> [-73.990134, 40.751658] is a location of https://www.openstreetmap.org/node/42439952 and comes from https://github.com/Project-OSRM/osrm-backend/blob/master/include/engine/guidance/assemble_geometry.hpp#L64 even if `source_node` is {... location = {-73.990134, 40.751657}, input_location = {-73.99039, 40.75171}, ...}.

* `tracepoints[0].location` -> [-73.990134, 40.751657] is a point in `candidates_lists = GetPhantomNodesInRange...` that is a projection of `query_location = {-73.99039, 40.75171}` onto `segment_source = {-73.989687, 40.752288}`, `segment_target = {-73.990134, 40.751658}` at https://github.com/Project-OSRM/osrm-backend/blob/master/include/engine/geospatial_query.hpp#L363-L367

Itself it is a rounding problem in `perpendicularDistance`, but would it make sense to use `source_node.location` as the first geometry point?

## Tasklist
 - [ ] fix a problem
 - [ ] review
 - [ ] adjust for comments

## Requirements / Relations

### DIFF
--- a/unit_tests/engine/guidance_assembly.cpp
+++ b/unit_tests/engine/guidance_assembly.cpp
@@ -33,44 +33,45 @@ BOOST_AUTO_TEST_CASE(trim_short_segments)
                                            {}};
 
     // Check that duplicated coordinate in the end is removed
-    std::vector<RouteStep> steps = {{324,
-                                     "Central Park West",
-                                     "",
-                                     "",
-                                     "",
-                                     "",
-                                     "",
-                                     0.2,
-                                     1.9076601161280742,
-                                     TRAVEL_MODE_DRIVING,
-                                     {{FloatLongitude{-73.981492}, FloatLatitude{40.768258}},
-                                      329,
-                                      348,
-                                      {TurnType::ExitRotary, DirectionModifier::Straight},
-                                      WaypointType::Depart,
-                                      0},
-                                     0,
-                                     3,
-                                     {intersection1}},
-                                    {324,
-                                     "Central Park West",
-                                     "",
-                                     "",
-                                     "",
-                                     "",
-                                     "",
-                                     0,
-                                     0,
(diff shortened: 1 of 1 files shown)


TheMarex:
> but would it make sense to use source_node.location as the first geometry point?

We should already do that because of this line:
https://github.com/Project-OSRM/osrm-backend/blob/master/include/engine/guidance/assemble_geometry.hpp#L45

Any idea why this is not the case here? Maybe the post-processing is playing a trick on us?

oxidase:
yes, `geometry = {locations = {{-73.990134, 40.751657}, {-73.990134, 40.751658}, {-73.990263, 40.751481}}, segment_offsets = {0, 1, 2}, segment_distances = {0.1112262997322674, 22.488323336387797}, osm_node_ids = {1329098617, 42439952, 42437654}, annotations = ...}` is returned and at https://github.com/Project-OSRM/osrm-backend/blob/master/src/engine/guidance/post_processing.cpp#L1229 the first point is removed because both `zero_length_step` and `duplicated_coordinate` are true.

If the first point must be a projection of the input location then `trimShortSegments` must be adjusted.

oxidase:
@MoKob i have updated the regression test for the issue, please could you check if expectations at https://github.com/Project-OSRM/osrm-backend/blob/fix/3516/unit_tests/engine/guidance_assembly.cpp#L198-L205 make sense?

oxidase:
Stale PR: #3775 and #3516 are closed now

Issue #3368, opened 2016-11-27
Different rounding in coordinate transformations in osmium wrt to osrm


oxidase:
Some tests expectation are ad hoc and have significant differences if input data is slightly translated, for example
```
    Scenario: Car - Check roundoff errors
        Then routability should be

            | highway | maxspeed | forw    | backw   |
            | primary |          | 52 km/h | 52 km/h |
            | primary | 60       | 47 km/h | 47 km/h  |
            | primary | 60       | 47 km/h | 47 km/h  |
```
fails in last row. The difference is due to rounding up in osmium at https://github.com/Project-OSRM/osrm-backend/blob/master/third_party/libosmium/include/osmium/osm/location.hpp#L183
but osrm uses truncated coordinates at https://github.com/Project-OSRM/osrm-backend/blob/master/include/util/coordinate.hpp#L77. 

In the above test

| OSM/request        | osmium reads | osmium rounds to |    node lon |   phantom input |
|---|---|---|---|---|
| 1.926039974358575  |192603997    |19260400 |         19260400 |   1926039 |

so  for phantom input 1926039 weight is 1499 and speed is 3.6*1999.8/149.9=48.0272181454 -> 48,
but for phantom input 1926040 weight is 1500 and speed is 3.6*1999.8/150=47.9952 -> 47
so initial relative error of 2e-7 propagates to 1/48 in the result speed.

TheMarex:
Good find. We should probably use `std::round` [here](https://github.com/Project-OSRM/osrm-backend/blob/master/include/util/coordinate.hpp#L60-L79).

TheMarex:
To fully fix these incompatibilities we need to switch from 1e6 to 1e7 coordinate precision. Since we expose the `Coordinate` objects to our users, that could be a breaking change, hence the 6.0 milestone.

PR #3578, opened 2017-01-17
Use round for float to fixed coordinate translations

oxidase:
# Issue

There is no obvious reason to use 1e6 precision and also libosmium uses 1e7 precision with rounding towards infinity that in some special cases may lead to unexpected results within OSRM like #3516 

The PR change internal precision to 1e7 and fixes hard-coded expectations in tests.

For #3516 both latitudes  40.751658 and 40.751657 will be changed to 40.751659 as the rounded value of the input 40.7516588. Even both results are equal now the PR is not a solution for #3516
 
## Tasklist
 - [x] review
 - [x] adjust for comments

## Requirements / Relations
#3516, #3368


include/util/coordinate.hpp
```diff
@@ -40,7 +40,7 @@ SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 namespace osrm
 {
 
-constexpr const double COORDINATE_PRECISION = 1e6;
+constexpr const double COORDINATE_PRECISION = 1e7;
```
daniel-j-h: I think this could break existing user-code since our public interface include these private headers (which is unfortunate on its own):

https://github.com/Project-OSRM/osrm-backend/blob/1ae69fe40d68a15ca978f0a4535d75523b292467/include/osrm/coordinate.hpp#L31

oxidase: Hrm, this is the main change of the PR.
I will revert the line but will keep rounding coordinates to make conversions correct in https://github.com/Project-OSRM/osrm-backend/blob/1e7/unit_tests/util/coordinate_calculation.cpp#L391-L406

TheMarex: What makes you think this could break user code? Wouldn't you see the actual encoding as internal state?

oxidase:
Reverted /1e6/1e7/ change, because it could be an interface breaking change. The only effective change is `round` function in float to fixed coordinate transformations. Also updated tests towards stronger expectations ranges and adjusted input points in some tests, because snapping behavior is changed by PR.

include/util/coordinate.hpp
daniel-j-h: We still include internal headers in our public interface - this exposes internals such as the coordinate precision, too. Check this:

https://github.com/Project-OSRM/osrm-backend/blob/fce3bb180c56d0fc70d439c8e943cd6a3003363b/include/util/coordinate.hpp#L40-L43

osrm namespace, exposed via the `osrm/coordinate.hpp` hack here:

https://github.com/Project-OSRM/osrm-backend/blob/1ae69fe40d68a15ca978f0a4535d75523b292467/include/osrm/coordinate.hpp#L31

This is ticketed since some time now in https://github.com/Project-OSRM/osrm-backend/issues/2031.
If we're fine with this kind of breakage, I'm happy to include the changeset.

oxidase: @daniel-j-h it is better to cleanup public/private headers first. Anyway i think better to split PR into two commits:
* change to rounding coordinates to reduce error in coordinate projection with transformations float-point -> fixed-point -> float-point -> fixed-point. The added regression test expects `to(from(x)) = x`, but only for some specific locations.
* change to 1e7 precision
The first change is in the PR, but the latter change is more complicated to handle because it hardly influences computations in guidance, i hope in better direction, but all failing tests must be analyzed as a separate PR

Issue #3577, opened 2017-01-17
Move `osrm-unlock-all` and `osrm-springclean` functionality to `osrm-datastore`


oxidase:
After #3528 was merged the locking mechanism has been changed and `osrm-unlock-all` and `osrm-springclean` can be moved into `osrm-datastore` to have write access to shared memory blocks at one place and do not duplicate it in different tools.

Also [wiki page](https://github.com/Project-OSRM/osrm-backend/wiki/Configuring-and-using-Shared-Memory) must be updated with a section how to remove locks, eg, https://github.com/Project-OSRM/osrm-backend/blob/master/src/storage/storage.cpp#L62-L71 will not be unlocked by `osrm-unlock-all` and the lock file must be removed manually. This situation could be sign of a very severe  `osrm-datastore` internal problem.

Moved from PR:

- [x]    once this lands please close Project-OSRM/node-osrm#284
- [x]    https://github.com/Project-OSRM/osrm-backend/wiki/Configuring-and-using-Shared-Memory could need some polishing and we should have docs in place for these ipc features
- [x]    Travis mason build fails, probably a blip - try restarting it

oxidase:
Wiki page updated https://github.com/Project-OSRM/osrm-backend/wiki/Configuring-and-using-Shared-Memory#locking-mechanism-and-cleaning-locks and resolved via 6f543fb

Issue #3574, opened 2017-01-17
Unlocking Shared Fails after osrm-routed crashes


MoKob:
While running local experiments on a crashing request, I am not able to re-instate OSRM into a sane state.

```
./build/osrm-datastore data/berlin.osrm
[warn] Data update in progress, waiting until it finishes... 

./build/osrm-unlock-all 
[info] Releasing all locks

./build/osrm-springclean 
[info] Releasing all locks
[info] ATTENTION! BE CAREFUL!
[info] ----------------------
[info] This tool may put osrm-routed into an undefined state!
[info] Type 'Y' to acknowledge that you know what your are doing.
[info] 

Do you want to purge all shared memory allocated by osrm-datastore? [type 'Y' to confirm]
Y
[info] spring-cleaning all shared memory regions

./build/osrm-datastore data/berlin.osrm
[warn] Data update in progress, waiting until it finishes... 
```

It seems that unlocking shared memory is not possible with our toolchain anymore.

Edit:
a previous run of `osrm-springclean` showed:
```
[debug] deallocating prev memory 28049409
[debug] deallocating prev memory 27983872
[debug] deallocating prev memory 65538
```

Edit 2:
Even using `max-wait` does not work:

```
./osrm-datastore --max-wait 1 ../data/berlin.osrm
[warn] Data update in progress, waiting until it finishes... 
```
hangs for minutes.

Edit 3:
I got to this state running `osrm-datastore OSRM-FILE`, ran `osrm-routed -s` and had it crash multiple times. After I cannot run any queries anymore against osrm.

oxidase:
@MoKob please could you check `./osrm-datastore --remove-locks` that is similar to `./osrm-unlock-all`.
`--max-wait` options now is used only to get write access to `CURRENT_REGION` block

MoKob:
@oxidase thank you for the tip. I will see if I can reproduce it again, since I kind of needed my PC and had to restart.

Regardless of the outcome, I feel that we should opt to have a clear way of disabling locks. Right now I can see 4 ways to possibly remove locks, if everything hangs. This is kind of confusing and does not make it easy to find the source of the problem.

oxidase:
@MoKob i have created #3577 to reduce confusion in tools.

Please check
```
I got to this state running osrm-datastore OSRM-FILE, ran osrm-routed -s and had it crash multiple times. After I cannot run any queries anymore against osrm.
```
[...]

Issue #3466, opened 2016-12-16
Add --force flag to osrm-datastore


TheMarex:
Currently we bail out if we detect any left-over locks when running `osrm-datastore`. That becomes inconvenient when `osrm-datastore` or `libosrm` fails to release a lock for whatever reason.

We should add a `--force` flag that reset any external state so we safely recover:

- Remove all shared memory regions owned by OSRM
- Remove all shared memory locks

This combines the functionality from `osrm-springclean` and `osrm-unlock-all` and would make them obsolete.

/cc @miccolis

danpat:
@TheMarex hmm, if we want to remove `osrm-springclean`, we might want something like:

```
osrm-datastore --cleanup --force
  - basically osrm-springclean, overrides locks

osrm-datastore --force datafile.osrm
  - loads new data, overrides locks, removes old shared memory regions

osrm-datastore datafile.osrm
  - respects locks
```

I can't think of many situations where you would practically want to remove the shared memory regions without replacing them with different data, but it could be useful.

TheMarex:
Obsolete with new locking code.

PR #3579, opened 2017-01-18
Move IPC tools functionality into osrm-datastore

oxidase:
# Issue

After #3528 is landed, IPC tools can be out of sync with osrm-datastore because of locking mechanism and lock names were changed, ref #3577  #3574.

PR moves cleaning functionality into `osrm-datastore` and can be used as 
* `osrm-datastore --remove-locks` similar to `osrm-unlock-all`
* `osrm-datastore  --spring-clean` similar to `osrm-springclean && osrm-unlock-all`

Also can added an info option that prints some `ipcs` info (`SHM_INFO` and `SHM_STAT` http://man7.org/linux/man-pages/man2/shmctl.2.html) for regions.

## Tasklist
 - [x] review
 - [x] adjust for comments


daniel-j-h:
Looks good, a few things here:
- [ ] once this lands please close https://github.com/Project-OSRM/node-osrm/issues/284
- [ ] https://github.com/Project-OSRM/osrm-backend/wiki/Configuring-and-using-Shared-Memory could need some polishing and we should have docs in place for these ipc features
- [ ] Travis mason build fails, probably a blip - try restarting it

daniel-j-h:
Also: related https://github.com/Project-OSRM/osrm-backend/issues/3466

Issue #3572, opened 2017-01-17
Building fails on centos 7


hadi332:
hi guys.

I have a problem building osrm on my centos 7 server. I previously have built it successfully on ubuntu 14.04.
I do similar steps explained in https://github.com/Project-OSRM/osrm-backend/wiki/Building-OSRM and run the following command in order to build the package:
`cmake .. -DCMAKE_CC_COMPILER=/usr/local/bin/gcc -DCMAKE_CXX_COMPILER=/usr/local/bin/g++ -DSTXXL_LIBRARY=/usr/lib/libstxxl.a`

However the making process fails with the following message:
`-- Building on a 64 bit system
-- Using GNU gold as linker.
-- Disabling gc-sections on gold binutils < 2.26, see: https://sourceware.org/bugzilla/show_bug.cgi?id=17639
-- No build type specified, defaulting to Release
-- Configuring OSRM in release mode
-- Configuring release mode optimizations
-- Setting linker optimizations
CMake Error at /usr/share/cmake/Modules/FindBoost.cmake:1138 (message):
  Unable to find the requested Boost libraries.

  Boost version: 1.53.0

  Boost include path: /usr/include

  Detected version of Boost is too old.  Requested version was 1.54 (or
  newer).
Call Stack (most recent call first):
  CMakeLists.txt:451 (find_package)

-- Found Intel TBB
-- TBB interface version: 6103
-- Looking for STXXL...
-- Found STXXL: /usr/lib/libstxxl.a
-- Could NOT find Lua (missing:  LUA_INCLUDE_DIR) (Required is exact version "5.2")
-- Could NOT find Lua (missing:  LUA_INCLUDE_DIR) (Required is exact version "5.1")
-- Using Lua 5.3.3
-- Looking for protozero
-- Looking for protozero - found
-- OpenMP support found. Linking just in case for stxxl
-- Configuring incomplete, errors occurred!
See also "/home/hadi/osrm-backend/build/CMakeFiles/CMakeOutput.log".
See also "/home/hadi/osrm-backend/build/CMakeFiles/CMakeError.log".
`

And this what is written in CMakeError.log:
`Determining if the pthread_create exist failed with the following output:
Change Dir: /home/hadi/osrm-backend/build/CMakeFiles/CMakeTmp

Run Build Command:/usr/bin/gmake "cmTryCompileExec2975653299/fast"
[...]

PR #3580, opened 2017-01-18
Fix Lua static library linking

oxidase:
# Issue

If a Lua library is compiled from source as liblua.a linking of dl library is required
Also updated Lua profile generation: in 5.3 `loadstring` function has been renamed to `load`

## Tasklist
 - [x] review
 - [x] adjust for comments

## Requirements / Relations

Was reported in #3572

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -497,7 +497,7 @@ else()
     ENDIF()
   ENDIF()
 
-  set(USED_LUA_LIBRARIES ${LUA_LIBRARY})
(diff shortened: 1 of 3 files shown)


daniel-j-h:
Care to upstream the `FindLua.cmake` changes with CMake? We vendor it since we have to support older CMake versions, too. But this really should get fixed directly with CMake. At least in the long term. What's your take here?

oxidase:
@daniel-j-h https://gitlab.kitware.com/cmake/cmake/merge_requests/413

danpat:
LGTM, works fine on OSX.

PR #3581, opened 2017-01-18
refactor datafaces to allocator scheme

MoKob:
# Issue

Refactors the datafacades as a step towards better integration of https://github.com/Project-OSRM/osrm-backend/issues/3208.

Right now we have two separate datafacades that share the same logic. The main difference is the allocation scheme used, which is currently implemented as two derived classes from a common base.

Instead of creating different classes for the allocation schemes, this PR provides a common scheme using an allocator class to provide access to both memory variants.

In cases where we don't want to support any more datafacades at runtime, we can now use the common class directly instead of relying on virtual functions. This will allow for faster access around the full infrastructure, as long as we don't require access to further datafacades.

In short, we change

```
class FacadeBase {};
class SharedFacade : FacadeBase
{
     SharedFacade(SharedAllocInfo)
     {
          FacadeBase::Init(SharedAllocInfo.GetMemory());
     }
};
class ProcessFacade: FacadeBase
{
     ProcessFacade(ProcessAllocInfo)
     {
          FacadeBase::Init(ProcessAllocInfo.GetMemory());
     }
};

FacadeBase = make_shared(ProcessFacade);
//  or
FacadeBase = make_shared(SharedFacade);
```

into 

```
class Facade(AllocatorBase *allocator);
class AllocatorBase{ virtual GetMemory(); };
class SharedAllocator
{
GetMemory(){ // get shared memory}
}
class ProcessAllocator
{
GetMemory(){ // get internal memory}
}

Facade facade(make_shared(ProcessAllocator));
//or
Facade facade(make_shared(SharedAllocator));
```

This switch requires a single virtual function call for the allocator over a virtual function call on every FacadeBase access.
If we stay on these two Facades, we can move all our code to the use of Facade, instead of the common Base class. 

## Tasklist
 - [x] review
 - [x] adjust for comments


danpat:
I'm 👍 on this refactor - it's the logical next step after my earlier changes, and a much better separation of concerns in the code.

My earlier changes *removed* the `InternalDataFacade` which had a completely different (heap allocated) memory layout.  So far, there's been no evidence that it added any value.

For MLD, we may have multiple datafacades, but I don't see why we wouldn't be able to continue with this pattern with Allocators.

PR #3583, opened 2017-01-18
Refactor ShMemIterator, resolves #3458

### LINKED ISSUE #3458: ShMemIterator should be RandomAccessIterator

oxidase:
# Issue

For #3458, change ShMemIterator to a model of Random Access Traversal Iterator concept with `boost::iterator_facade` and reverse iterator with `boost::reverse_iterator`

## Tasklist
 - [x] review
 - [x] adjust for comments

### DIFF
--- a/include/util/shared_memory_vector_wrapper.hpp
+++ b/include/util/shared_memory_vector_wrapper.hpp
@@ -1,10 +1,13 @@
 #ifndef SHARED_MEMORY_VECTOR_WRAPPER_HPP
 #define SHARED_MEMORY_VECTOR_WRAPPER_HPP
 
-#include <boost/assert.hpp>
-
 #include "util/log.hpp"
 
+#include <boost/assert.hpp>
+#include <boost/iterator/iterator_facade.hpp>
+#include <boost/iterator/reverse_iterator.hpp>
+
+#include <climits>
 #include <cstddef>
 
 #include <algorithm>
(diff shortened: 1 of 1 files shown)


danpat:
@oxidase Thanks for cleaning this up.

A little while back, I played with removing `ShM<T>::vector` completely, replacing it with `boost::iterator_range<>`, like so:

```
util::ShM<util::Coordinate, true>::vector m_coordinate_list;
```

becomes:

```
using CoordList = boost::iterator_range<util::Coordinate>;
CoordList m_coordinate_list;
```

and data assignment inside the datafacades becomes:

```
const auto coordinate_list_ptr = data_layout.GetBlockPtr<util::Coordinate>(
            memory_block, storage::DataLayout::COORDINATE_LIST);
m_coordinate_list = CoordList(coordinate_list_ptr, coordinate_list_ptr + data_layout.num_entries[storage::DataLayout::COORDINATE_LIST]);
```

I'm 👍 on this PR, it's a welcome cleanup, but long term, I we should consider removing `ShM::vector` (and the support classes that are only used there) - this class isn't special enough to warrant maintaining our own.

Issue #2844, opened 2016-09-02
Handle up to 255 full Unicode Codepoints (not 255 bytes)


daniel-j-h:
Issue for test case added in https://github.com/Project-OSRM/osrm-backend/pull/2821.

The OSM spec says "up to 255 full unicode chars", so we should be able
to handle 255 x 🐼 --- but we don't.

The internal limit is on 255 _bytes_ in the `std::string`, but there can
be e.g. up to four bytes in a utf8 encoding.

References:
- http://wiki.openstreetmap.org/wiki/API_v0.6#Tags
- http://www.fileformat.info/info/unicode/char/1f43c/browsertest.htm
- https://en.wikipedia.org/wiki/UTF-8#Description
- https://github.com/Project-OSRM/osrm-backend/blob/807aa71a7c80b714b7a89db9c99c15720572ba25/include/util/range_table.hpp#L28

MoKob:
The feature to check for this is in (https://github.com/Project-OSRM/osrm-backend/pull/2821), now we just need to handle it.

daniel-j-h:
I looked into the range table being the limiting factor for how we ultimately store strings. @MoKob pointed out the prefix sum being the main issue here. Playing with it for a bit and switching from a `char` to a larger type now lets the intersection classification I/O code fail.

```
terminate called after throwing an instance of 'osrm::util::exception'
  what():  Failed to read bearing values from /tmp/osrm-backend/test/cache/ec565977fc21e9bcff967d61a3bd71e09a5b7460_c499d2c4587859ff929c81688cdef37d5dbc70bd_c12f16d2fcd912fb00db7ff4f55d4106f9d0cc57.osrm.icd
```

MoKob:
@daniel-j-h this is also reading in a range table. So if you break the range table, you have to adjust the code for reading the intersection_classification in as well.

TheMarex:
The reason why the range table operates on bytes is that this enables vectorized computation of the prefix-sum. You would need to check not only the memory impact (for using 16bit) but also the performance impact on the range table.

The specification might say 255 Unicode chars, but what do we actually hit on a planet extract?

As a side note there is a wealth of research for efficiently encoding delta-int values like these. Most of them are based on varint encodings - we might give this a spin and drop the hard coded length limit altogether. For example Folly has an implementation for GroupVarint https://github.com/facebook/folly/blob/master/folly/docs/GroupVarint.md but there are numerous other open source libs for this.

[...]

PR #3585, opened 2017-01-19
Added indexed array data type with variable and fixed group blocks

oxidase:
# Issue

One way to solve #2844 without additional memory is to use variable group codes. The current implementation of `VariableGroupBlock` uses 2 bits in the descriptor word with the following meaning:
* 0 - zero size string, no additional space is used
* 1 - size of string 1..255 code units and 1 additional byte is used
* 2 - size of string 256..65535 code units and 2 additional bytes are used
* 3 - size of string 65536..16777215 code units and 3 additional bytes are used
Because now empty strings use only 2 bits instead of 1 byte the name data index size will be reduced if empty strings are at least 25%.

Index can be reduced even more using:
* 0 - zero size string, no additional space is used
* 1 - size of string 1..15 code units and 4 bits additionaly  used
* 2 - size of string 16..255 code units and 1 additional byte is used
* 3 - size of string 256..4096 code units and 12 additional bits are used

`FixedGroupBlock` is the original implementation with additional 1 byte per string for length 0..255,
The only difference that array with string length is now placed in data array as 16 bytes prefix and decriptor contains only 32 bit global offset of the block. 

## Tasklist
 - [x] Use `IndexedData` instead of `RangeTable` to save names data
 - [x] Check nibble sizes wrt performance
 - [x] review
 - [x] adjust for comments

## Requirements / Relations
https://arxiv.org/abs/1209.2137


oxidase:
added usage of `IndexedData`. Checked performance with callgrind with O3, it is ~2 slower wrt to fixed block.
Also there is no more SSE optimization possible because blocks have `char` alignment. 

func | calls | fixed | variable | "instructions" per call for variable blocks 
--- | --- | --- | --- | ---
GetNameForID(unsigned int) const | 8 512 | 632 846 | 1 253 107 | 145 
GetDestinationsForID(unsigned int) const | 8 512 | 632 861 | 1 260 120 | 146 
GetPronunciationForID(unsigned int) const | 8 512 | 632 883 | 1 263 050 | 146 
GetRefForID(unsigned int) const | 8 512 | 632 898 | 1 267 281 | 146

If performance will be a problem it is possible to switch back to the fixed block implementation and try to use SSE optimization. Also possible to load the index directly into memory, but it will increase memory consumption

TheMarex:
@oxidase question w.r.t to the table: It shows the number of instructions for `fixed` and `variable` over all calls right?

We can estimate the impact this can have on the performance by ball parking the maximum number of `RouteStep` objects in a response. I would imagine 200 would be a save estimate. How long do 4x 200 calls do these functions take? Anything below 10ms is negligible (these routes are typically in the order of 200-300ms and huge).

oxidase:
@TheMarex the numbers are total cost of events " Instruction Read" per function. Roughly for 0.5ns per instruction read the execution time is 0.5ns*1253107/8512*4*200=58us vs 30us

oxidase:
just changing 
```
        data += byte_length >= 1 ? static_cast<unsigned char>(*length++) : 0;
        data += byte_length >= 2 ? static_cast<unsigned char>(*length++) << 8 : 0;
        data += byte_length == 3 ? static_cast<unsigned char>(*length++) << 16 : 0;
```
to https://github.com/Project-OSRM/osrm-backend/blob/46b6e144c9e8060fb2047566c97b2600599abef4/include/util/indexed_data.hpp#L52-L70
improves performance by 10%, so performance drop wrt fixed block ~50% and time estimation is 77us
variable

danpat:
I'm totally OK with a 77us time penalty for fetching 200 names in return for supporting 4096 bytes of data per name (🐼🐼 hooray).

@oxidase for the OSM name data (for say, Berlin?), how much additional data is required?  I assume it's quite small...

oxidase:
@danpat size of osrm.names files will be ~10% less because of more tight index deltas packing
```
berlin-latest 214432 -> 194990
germany-latest 10352768 -> 9506086
```

TheMarex:
Rebased on master.

include/util/indexed_data.hpp
```diff
@@ -0,0 +1,366 @@
...
+};
+
+template <typename GroupBlock> struct IndexedData
+{
+    static constexpr std::uint32_t BLOCK_SIZE = GroupBlock::BLOCK_SIZE;
+
+    typedef std::uint32_t BlocksNumberType;
+    typedef std::uint64_t DataSizeType;
+
+    typedef typename GroupBlock::BlockReference BlockReference;
```
TheMarex: nit pick: I always find the new using syntax easier to read: `using BlockReference = typename GroupBlock::BlockReference;`

src/util/name_table.cpp
```diff
@@ -1,64 +1,58 @@
...
-
-#include <boost/filesystem/fstream.hpp>
-
 namespace osrm
 {
 namespace util
 {
 
-NameTable::NameTable(const std::string &filename)
+NameTable::NameTable(const std::string &file_name)
```
TheMarex: The name of this class is a little misnormer now, since there is not RangeTable implementation anymore. Maybe `NameIndex`?

include/util/indexed_data.hpp
```diff
@@ -0,0 +1,366 @@
...
+
+        if (local_index < BLOCK_SIZE)
+        {
+            last = first;
+            var_advance(last, var_lengths, descriptor & 0x3);
+        }
+    }
+};
+
+template <int N, typename T = std::string> struct FixedGroupBlock
```
TheMarex: This implementation is mostly used for testing or do you have some ideas where we could utilize it in the future?

src/util/name_table.cpp
oxidase: It is still a table with block index and variable part of the index that  prepends variable data. I would keep `NameTable` name here

include/util/indexed_data.hpp
oxidase: It can be useful in some cases. Originally i thought to use `FixedGroupBlock` for `util::RangeTable<> bearing_class_range_table(bearing_counts)`, but did not manage to finish it.

TheMarex: Cool, we should ticket this. 👍

Issue #3422, opened 2016-12-07
Refactor foot profile


emiltin:
Along the lines of the recent car refactoring.

daniel-j-h:
Referencing same for bike: https://github.com/Project-OSRM/osrm-backend/issues/3413

daniel-j-h:
This landed with https://github.com/Project-OSRM/osrm-backend/pull/3597 :sparkles:  - Thank you!

PR #3597, opened 2017-01-23
refactor foot profile (as was done for car)

emiltin:
# Issue
Refactors the foot profile into smaller methods, as was previously done for the car profile: #3422.

Next step will be to factor out common methods from car & foot profile to helper files. Most of the methods used in the two profiles are now the same.

## Tasklist
 - [ ] review
 - [ ] adjust for comments

## Requirements / Relations

### DIFF
--- a/profiles/debug.lua
+++ b/profiles/debug.lua
@@ -43,14 +43,16 @@ mode = {
 
 -- input tags, normally extracted from OSM data
 local way = {
-  highway = 'primary',
-  name = 'Main Street',
+  --highway = 'footway',
+  amenity = 'parking',
+  name = 'Hyper Drive',
+  --foot = 'yes',
+  --oneway = 'yes',
   --access = 'no'
-  ["access:forward"] = 'no'
+  --["access:forward"] = 'no'
   --width = '3',
   --maxspeed = '30',
   --['maxspeed:advisory'] = '25',
-  --oneway = '-1',
   --service = 'alley',
   --['oneway:bicycle'] = 'yes',
   --junction = 'roundabout',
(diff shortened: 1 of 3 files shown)


profiles/lib/directional.lua
```diff
@@ -56,4 +56,37 @@ function Directional.get_values_by_set(way,data,keys)
   return forward, backward
 end
 
+-- look through a sequence of keys combined with a prefix
+-- e.g. for the sequence [motorcar,motor_vehicle,vehicle] and the prefix 'oneway' search for:
+-- oneway:motorcar
+-- oneway:motor_vehicle
+-- oneway:vehicle
+
+function Directional.get_value_by_prefixed_sequence(way,seq,prefix)
```
emiltin: these helpers are not directly related to directional (forward/backward) tags. either they could be moved to a separate file, or we could rename the Directional helper to something more general.

profiles/debug.lua
```diff
@@ -43,14 +43,16 @@ mode = {
 
 -- input tags, normally extracted from OSM data
 local way = {
-  highway = 'primary',
```
emiltin: i will clean up the debug.lua file, so it's just the general lua debugging tool, and then do my local work in a git ignore file. no need to have these kind of development changes in the commits going forward

daniel-j-h: :+1:

daniel-j-h:
Sweet! Thank you for your contiguous improvements to the profiles!

> Next step will be to factor our common methods from car & foot profile to helper files. Most of the methods used in the two profiles are now the same.

Yep, sounds good. The generic functionality should be split up and be more modular.

profiles/debug.lua
emiltin: was done in a separate PR

emiltin:
Anything else I can do here?

PR #3599, opened 2017-01-23
clean up lua debugging

emiltin:
# Issue

Cleans up the LUA debug helper. File was split into lib/profile_debugger.lua, and an example file debug_example.lua. The ideas is that you copy debug_example.lua to a file which you gitignore (debug.lua has already been gitignored).

## Tasklist
 - [x] review
 - [x] adjust for comments

## Requirements / Relations
None.


profiles/lib/profile_debugger.lua
```diff
@@ -0,0 +1,138 @@
...
+  return true
+end
+
+function parseDuration(str)
+  return 1
+end
+
+function canonicalizeStringList(str)
+  return str
+end
```
daniel-j-h: I'm only now seeing it but this means every time we bind a C++ function to Lua we must not forget to add a stub here too..

profiles/debug_example.lua
```diff
@@ -0,0 +1,50 @@
...
+  highway = 'primary',
+  name = 'Magnolia Boulevard',
+  ["access:forward"] = 'no'
+}
+
+-- output will go here
+local result = {}
+
+-- call the way function
+Debug.way_function(way,result)
```
daniel-j-h: cc @TheMarex @oxidase this could also be a way to take e.g. conditional maxspeeds into account: feed them only through the profiles at this location and dump out their speed to a speed update csv file.

profiles/lib/profile_debugger.lua
emiltin: or implement it in LUA in the first place, it performance is not a problem

emiltin:
Next step?

daniel-j-h:
Can you rebase against master, then this should be good to go.

emiltin:
rebased on master

daniel-j-h:
Thank you @emiltin ! :)

PR #3602, opened 2017-01-23
Remove decltype(Value::source), fixes xcode 7.3 compilation

oxidase:
# Issue

For https://github.com/Project-OSRM/node-osrm/issues/289 `decltype(Value::source)` changed to `std::numeric_limits<std::uint8_t>::max()` 

## Tasklist
 - [x] review
 - [x] adjust for comments

## Requirements / Relations


daniel-j-h:
The reason we didn't catch this in `osrm-backend` is we do not have macOS Debug builds on Travis.

Issue #3586, opened 2017-01-19
Calculating Flows Recursively


MoKob:
To generate partitions for our MLD implementation, we need to generate a set of partitions `P1,..,PL` with `L` equal to the number of levels and for all border vertices `P_i \in P_{i+1}`.

For the general algorithm, we can refer to http://sommer.jp/roadseparator.htm

Terminology

 - `partition`: mapping V -> [1,k] for every vertex in the graph
 - `bisection`: partition with `k == 2`
 - `cell`: the set of v in V with the same mapping
 - `level`: in a recursive scheme, this is the depth of the recursion

## Recursive Bisection with Inertial Flow
We have had some experience with the inertial flow algorithm which @daniel-j-h implemented. The basic idea of the algorithm is to select nodes in geographic regions to the left/right side of the map and compute a minimal border in between. This create a bisection of the graph.

We reinterpret every partition as a new local graph and recursively apply the interial flow algorithm again.
Recursively cutting the partitions generated in a level l, we can generate level `l+1` from it, adding an additional cut in every `cell`.

![recursive-bisection](https://cloud.githubusercontent.com/assets/12932279/22111770/d4e6796c-de60-11e6-9662-2d3e4dd46f52.png)

To implement this intertial flow partitioning scheme, we require an algorithm to compute the minimum cut between two sets of vertices (the `source` and the `sink`).

### Finding Spatial Orders

One tool that we require for the inertial flow algorithm is to compute spatial orders.

@daniel-j-h already has a lot of experience here. http://daniel-j-h.github.io/post/selection-algorithms-for-partitioning/ is a nice source to refer to for how to implement this.

Multiple of these spatial orders will have to be compared for their partitioning result on every level. It might be possible to re-use these orders by doing some stable partitioning on them, accoring to our selection on cells.

For starters, I think we should be fine with doing local partitionings on every level, since we are bounded by `n log^2(n)` here, so the overhead should be negligible in relation to the overhead for the flow algorithm itself. 

## Tool: Dinics algorithm
One important constraint we can apply is that we don't care about weights or in terms of flow algorithms about capacity. We only care about the minimum number of border vertices.

[...]

Issue #3205, opened 2016-10-28
Implement Inertial Flow Partitioner


daniel-j-h:
Once we have a dummy partitioner (ticketed in https://github.com/Project-OSRM/osrm-backend/issues/3204) we unblock further steps in the pipeline. Eventually we want to replace the dummy partitioner with a proper partitioning approach: Inertial Flow.

> On Balanced Separators in Road Networks
> Aaron Schild and Christian Sommer
> SEA 2015 - 14th International Symposium on Experimental Algorithms (pp. 286-297)
> doi: 10.1007/978-3-319-20086

See: http://sommer.jp/roadseparator.htm

We already successfully prototyped a internal routing engine with that - some hints are over at

http://daniel-j-h.github.io/post/selection-algorithms-for-partitioning/

our multiple candidate cuts and parallelization should be doable in osrm, too.

Tasks:
- [x] Dinic's Max Flow / Min Cut
- [x] Inertial Flow

daniel-j-h:
The Inertial Flow Partitioner and the dummy partitioner outlined in https://github.com/Project-OSRM/osrm-backend/issues/3204 will share the same interface - only the implementation will change. Let's spec out how this interface will look like (and answer open questions).

Eventually we need a partition on the edge based graph for the customization to make use of it (the call site is outlined in https://github.com/Project-OSRM/osrm-backend/issues/3552). The Intertial Flow partitioner needs an embedding, which we already have for the node based graph. This and the fact that @MoKob already had some ideas how to transfer a partition on the node based graph to the edge based graph makes it a viable option to partition the node based graph already.

- [ ] @MoKob would you be so kind to post the details for how to transform a partition on the node based graph to the edge based graph - where should this happen, what are the limitations, how can this be done.

In case we want to make the partitioner a new tool `osrm-partition` reading in the node based graph and writing out a partition for the node based graph (or already for the edge based graph?) ticket https://github.com/Project-OSRM/osrm-backend/issues/2055 is to mention here: at the moment we write out the node based graph in `osrm-extract` and then read it back in immediately with `osrm-contract`.

- [ ] Clarify if the ticket https://github.com/Project-OSRM/osrm-backend/issues/2055 is still relevant: do we want to write out the node based graph to decouple `osrm-extract` from `osrm-partition` now?
[...]

PR #3603, opened 2017-01-23
Basic Toolchain for Partitioning

MoKob:
Builds basic infrastructure outline in https://github.com/Project-OSRM/osrm-backend/issues/3586, defining interfaces and connections between the different parts. Foundation for https://github.com/Project-OSRM/osrm-backend/issues/3205.

Tasks
- [x] Immutable (Sub-) Graph View
- [x] Spatial Order
- [x] Dinic's Max-Flow / Min-Cut
- [x] Inertial Flow
- [x] Optimized Inertial Flow, rotating slope, in parallel
- [x] Recursive Bisection
- [x] Apply to Compressed Node Based Graph from `osrm-extract`
- [x] Run Tarjan for big / small component detection
- [x] Parallelize recursive bisection, outlined in https://github.com/Project-OSRM/osrm-backend/issues/3586
- [x] Do bisection in parallel on all big components
- [x] Fake first level for big components, aggregate all small components
- [x] Annotation mechanism for combining bisection ids -> levels
- [x] Unit tests for all the things ^
  - [x] partition graph
  - [x] graph view
  - [x] cut algorithm
  - [x] scc integration
  - [x] recursive bisection
- [x] fix windows builds
- [x] Check why sanitizers trigger for TBB (probably false positive)

## Once done
 - [ ] review
 - [ ] adjust for comments


include/partition/bisection_graph.hpp
```diff
@@ -0,0 +1,99 @@
...
+
+        // count all edges with this source
+        while (edge_itr != edges.end() && edge_itr->source == node_id)
+            ++edge_itr;
+
+        // go to the next node
+        ++node_id;
+
+        return {static_cast<std::size_t>(std::distance(edges.begin(), edges_of_node)), coordinate};
+    };
```
daniel-j-h: Hm can we clean this one up - the stateful lambda with init captures and transform is a bit ugly

include/partition/inertial_flow.hpp
```diff
@@ -0,0 +1,25 @@
...
+{
+namespace partition
+{
+
+class InertialFlow
+{
+  public:
+    InertialFlow(const GraphView &view);
+
+    std::vector<bool> ComputePartition(const double balance, const double source_sink_rate);
```
daniel-j-h: Should we use dynamic bitset here

http://www.boost.org/doc/libs/1_63_0/libs/dynamic_bitset/dynamic_bitset.html

(search for vector bool in the docs there)

daniel-j-h: What do you mean with balance and source sink rate here?

If balance is 0.25 we take 0.25 * nodes as source and 0.25 * nodes as sinks, right?

daniel-j-h: Also seeing you implemented the skeleton as stateful class holding a graph view ref. I suppose you want to abstract the try-n-cuts-take-best inside `ComputePartition` (?)

include/partition/partition_config.hpp
```diff
@@ -0,0 +1,74 @@
...
+THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND
+ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED
+WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
+DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE FOR
+ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES
+(INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
+LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON
+ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+(INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
```
daniel-j-h: License can go

include/partition/inertial_flow.hpp
MoKob: The `interface` here is just a placeholder for me to imagine what would happen:

- balance would be a guideline for the cuts to be chosen. Any flow that is computed gives a large number of possible cuts (e.g. `a-b-c-d-e-f-g-h` can be split on any intermediate node with an identical cut size). The balance would guide the flow algorithm which of the possible minimal flows to report / what kind of effort to spend on balancing the cuts in size
- the source/sink-rate would be for the initialisation of the flow algorithm. The inertial flow picks X percent of the nodes as source/sink, right? So that would be the X.
- For the statefulness: we can easily make this not stateful and make the flow algorithm do the `best out of n` part here. That was just to get something working quickly to iterate on.

include/partition/graph_view.hpp
```diff
@@ -49,7 +49,7 @@ class EdgeIDIterator : public boost::iterator_facade<EdgeIDIterator,
     void decrement() { --position; }
     void advance(difference_type offset) { position += offset; }
     bool equal(const EdgeIDIterator &other) const { return position == other.position; }
-    const reference dereference() const { return position; }
+    reference dereference() const { return position; }
```
daniel-j-h: Hm why the change here?

MoKob: since `reference` is already `const` due to https://github.com/Project-OSRM/osrm-backend/pull/3603/files/50b03a93f9472a3e726fc82d6d452ecb1561054f..3bde4426d491540190d403b889f5a8fcf702164a#diff-d79e8547c1b4e3f383ca1879d7db6bebL35 and my compiler complained about the double `const`

src/tools/extract.cpp
```diff
@@ -46,6 +46,11 @@ return_code parseArguments(int argc, char *argv[], extractor::ExtractorConfig &e
             ->implicit_value(true)
             ->default_value(false),
         "Generate a lookup table for internal edge-expanded-edge IDs to OSM node pairs")(
+        "dump-partition-graph",
+        boost::program_options::value<bool>(&extractor_config.dump_compressed_node_based_graph)
+            ->implicit_value(true)
+            ->default_value(false),
+        "Generate a partitionable graph file (.cnbg) for use with osrm-partition")(
```
daniel-j-h: Hm on the demo server with Boost 1.55 I _have to_ say `--dump-partition-graph true` even though `implicit_value` above should only require `dump-partition-graph`. No idea if this is an Boost issue...

http://www.boost.org/doc/libs/1_55_0/doc/html/boost/program_options/typed_value.html#idp101668472-bb

danpat: Can we just do `-DENABLE_MASON=ON` on the demoserver and pretend this is ancient history?

daniel-j-h: It works specifying it explicitly.

daniel-j-h:
Two blocking issues:

The first one is gcc 4.9 failing to compile TBB's `parallel_do` call:
- https://travis-ci.org/Project-OSRM/osrm-backend/jobs/198773611#L2249
- https://github.com/01org/tbb/blob/4.4/include/tbb/parallel_do.h#L134-L142
- https://gcc.gnu.org/bugzilla/show_bug.cgi?id=51048 (looks related, fixed in gcc 5)

The second one is a memory leak coming from TBB:
- https://travis-ci.org/Project-OSRM/osrm-backend/jobs/198817534#L5916-L5922

Note: TBB creates some static data structures which often provoke false positives in sanitizers / valgrind.

danpat:
@daniel-j-h What platforms only have GCC 4.9?  MLD is a major new feature, we could consider making GCC 5 the minimum.

daniel-j-h:
Ubuntu 15.10 by default for example (can use ppa), CentOS (we don't really care.. sorry).
I'm fine with upping the requirements here if it unblocks us.

@oxidase maybe you have ideas here. If you check the `parallel_do` implementation, the `parallel_do_feeder<Item>` is implemented with NVI pattern: `parallel_do_feeder_impl` inherits and implements `internal_add` (pure virtual). I can't see why this should not be legal. Except compiler bug.

Also there is no way for us to work around this issue, since it's coming directly from within the internal TBB code. The options we:
- Upgrade to >=gcc5
- Do not use `parallel_do` for bisection tree parallelization (urgh)
- Implement `parallel_do` ourselves (blargh), maybe vendor `parallel_do` with gcc49 fixes
- Report gcc49 bug with TBB and wait until they fix it in a new release

I can reproduce the issue in _our code_ locally. But I can't reproduce it in a small self-contained snippet.

```c++
#include <tbb/parallel_do.h>
#include <utility>
#include <vector>

struct node_t {
  int v;
};

int main() {
  std::vector<node_t> v{node_t{0}, node_t{1}};

  using feeder_t = tbb::parallel_do_feeder<node_t>;

  tbb::parallel_do(begin(v), end(v), [&](const node_t& n, feeder_t& feeder) {
    if (n.v < 2) {
      node_t more{n.v + 1};
      feeder.add(std::move(more));
    }
  });
}
```

    g++ -Wall -Wextra -pedantic -std=c++14 repro.cc -ltbb

 I expected this to trigger - but it doesn't.

daniel-j-h:
Here's a workaround @oxidase told me: just make gcc47 less std conformant: `-fpermissive`. Tada.

.travis.yml
```diff
@@ -42,7 +42,7 @@ matrix:
         apt:
           sources: ['ubuntu-toolchain-r-test']
           packages: ['g++-6', 'libbz2-dev', 'libstxxl-dev', 'libstxxl1', 'libxml2-dev', 'libzip-dev', 'lua5.1', 'liblua5.1-0-dev', 'libtbb-dev', 'libgdal-dev', 'libluabind-dev', 'libboost-all-dev']
-      env: CCOMPILER='gcc-6' CXXCOMPILER='g++-6' BUILD_TYPE='Debug' TARGET_ARCH='x86_64-asan' ENABLE_COVERAGE=ON ENABLE_SANITIZER=ON BUILD_COMPONENTS=ON
+      env: CCOMPILER='gcc-6' CXXCOMPILER='g++-6' BUILD_TYPE='Debug' TARGET_ARCH='x86_64-asan' ENABLE_COVERAGE=ON ENABLE_SANITIZER=OFF BUILD_COMPONENTS=ON
```
MoKob: This is a problem that needs to be solved, we see (hopefully) false positives in TBB. @daniel-j-h can you investigate here why we see these, if they are actually false positives or if we use TBB wrong?

include/util/static_graph.hpp
```diff
@@ -18,54 +20,87 @@ namespace osrm
...
+    using NodeArrayEntry = NodeT;
+    using EdgeArrayEntry = EdgeT;
 
     EdgeRange GetAdjacentEdgeRange(const NodeID node) const
     {
         return irange(BeginEdges(node), EndEdges(node));
     }
 
-    template <typename ContainerT> StaticGraph(const int nodes, const ContainerT &graph)
+    template <typename ContainerT> FlexibleStaticGraph(const int nodes, const ContainerT &graph)
```
MoKob: we could think about reverting these changes completely, since the static graph is no longer used in partitioning and we don't need the adjustments anymore. @daniel-j-h since you added the traits, whats your take?

daniel-j-h: Feel free to revert the changes. I'd like to keep the traits, though. They're still valid for the static graph and should even be incorporated into the dynamic graph. There should be no need of an edge data.

.travis.yml
daniel-j-h: Yes, will do when I have the time for it. Added to the todo list above.

oxidase: What you can do is `-fsanitize-blacklist=...` and put TBB into the blacklist. We also can add UBSAN with blaclisted TBB, because it has some issues with undefined behavior.

MoKob: that would be the last resort. Before I would like to be sure that the errors we see (we do a lot of tbb stuff, it just shows now) are actually false positives and not us screwing up.

daniel-j-h:
Hey @MoKob - heads up, I just implemented the changes you requested on my branch, then rebased and merged in it here. Should unblock @oxidase for now.

MoKob:
@daniel-j-h how do you feel about merging this now-ish and ticket the remaining items +

 - [ ] unit test edge-based-graph loader

for later? Would love to get this PR off my plate. If travis should work out now that we actually fixed the leaks in unit-tests, we don't even need the additional mason build anymore.

Ok if we bring this into `review`?

daniel-j-h:
Yeah you can bring it into review. That said, I think these are the additional open tasks:

- [ ] Ebg loader is ugly and should probably refactored with what the contractor does (if possible)
- [ ] Annotation tool should actually do the annotation (?)
- [ ] Border cases needs to be resolved in a smarter way (part I)
- [ ] We need to write out the new transformed edge based graph for customize (part II)

MoKob:
Tracking further issues in https://github.com/Project-OSRM/osrm-backend/issues/3731

oxidase:
Rebased the branch to master, adjusted to new StaticGraph interface tests and scc code, fixed some warnings and return of incorrect edges vector.

TheMarex:
Grabbing this PR for rebasing and fixing PR comments.

PR #3604, opened 2017-01-23
Updates Boost to 1.63 for Mason Build

daniel-j-h:
See if Travis likes this now.

Should fix a few issues upstream (such as https://github.com/boostorg/lexical_cast/issues/21), but more importantly we need it in order to see if node-osrm builds works it again (see https://github.com/Project-OSRM/node-osrm/pull/290).


springmeyer:
This is green now. I fixed the issue upstream in mason and restart the failing OS X build, which is now passing: https://travis-ci.org/Project-OSRM/osrm-backend/jobs/194535913.

Issue #3605, opened 2017-01-24
Incorrectly supressing turns onto oneways


danpat:
It looks like we're missing a right-turn instruction in some circumstances:

![screen shot 2017-01-23 at 4 33 29 pm](https://cloud.githubusercontent.com/assets/1892250/22228997/be6fa664-e189-11e6-9a1e-6dac33ddb3d9.png)

http://map.project-osrm.org/?z=18&center=45.522030%2C-73.594408&loc=45.521879%2C-73.595143&loc=45.521755%2C-73.593668&hl=en&alt=0

compare to this, where the start point is moved slightly south (so the route passes an extra intersection):

![screen shot 2017-01-23 at 4 33 43 pm](https://cloud.githubusercontent.com/assets/1892250/22228993/bcc8725a-e189-11e6-933e-6b271c26025a.png)

http://map.project-osrm.org/?z=18&center=45.520391%2C-73.595647&loc=45.522071%2C-73.594928&loc=45.521755%2C-73.593668&hl=en&alt=0

At a first glance, this looks like it *might* be related to obvious turn handling.

danpat:
It's not just right turns: http://map.project-osrm.org/?z=18&center=45.521402%2C-73.595175&loc=45.522466%2C-73.593974&loc=45.521932%2C-73.594006&hl=en&alt=0

![screen shot 2017-01-23 at 5 10 23 pm](https://cloud.githubusercontent.com/assets/1892250/22229805/e0f9ba12-e18e-11e6-88c4-a14bb3c52e3f.png)

danpat:
Ah - it looks to be occurring only in combination with end-of-road detection:

Example failure:
![screen shot 2017-01-23 at 5 14 06 pm](https://cloud.githubusercontent.com/assets/1892250/22229917/81d5bb84-e18f-11e6-9001-4bb10d7e6389.png)

but one block away, with no T-intersection for the turn:
![screen shot 2017-01-23 at 5 14 21 pm](https://cloud.githubusercontent.com/assets/1892250/22229913/7964fc12-e18f-11e6-981c-0de7d0d689f1.png)
(http://map.project-osrm.org/?z=18&center=51.042413%2C-114.082718&loc=51.043569%2C-114.081398&loc=51.042908%2C-114.082101&hl=en&alt=0)

PR #3607, opened 2017-01-24
3605 - fix error in treating end-of-road as choiceless

MoKob:
# Issue

Resolves https://github.com/Project-OSRM/osrm-backend/issues/3605

## Tasklist
 - [x] review
 - [x] adjust for comments

## Requirements / Relations
none

### DIFF
--- a/src/engine/guidance/post_processing.cpp
+++ b/src/engine/guidance/post_processing.cpp
@@ -60,7 +60,7 @@ inline bool choiceless(const RouteStep &step, const RouteStep &previous)
                                                    step.intersections.front().entry.end(),
                                                    true);
 
-    return is_without_choice;
+    return is_without_choice && step.maneuver.instruction.type != TurnType::EndOfRoad;
 }
 
 // List of types that can be collapsed, if all other restrictions pass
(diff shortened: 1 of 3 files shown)


MoKob:
screen shot 2017-01-24 at 11 05 40

src/engine/guidance/post_processing.cpp
```diff
@@ -60,7 +60,8 @@ inline bool choiceless(const RouteStep &step, const RouteStep &previous)
                                                    step.intersections.front().entry.end(),
                                                    true);
 
-    return is_without_choice;
+    return is_without_choice && step.maneuver.instruction.type != TurnType::EndOfRoad;
+    ;
```
daniel-j-h: accidental ;

src/engine/guidance/post_processing.cpp
```diff
@@ -60,7 +60,8 @@ inline bool choiceless(const RouteStep &step, const RouteStep &previous)
                                                    step.intersections.front().entry.end(),
                                                    true);
 
-    return is_without_choice;
+    return is_without_choice && step.maneuver.instruction.type != TurnType::EndOfRoad;
```
daniel-j-h: Shouldn't this be good enough to fix the issue we're seeing? What's going in below?

src/engine/guidance/post_processing.cpp
MoKob: No, on roads without a name, the `isLinkroad` can trigger, even though it is a longer road. The image linked above will still not show the turn, even though it might seem just the same, without the additional modifications. The two tests added reflect these two cases.

MoKob:
The tests added will ensure that this is fixed just the same in the refactor. So far it is not backported, but I will add it when I rebase the next time.

Issue #3595, opened 2017-01-22
Matching plugin: add a no u-turn properties


kaligrafy:
Would it be possible to add a no_uturn_in_matching flag properties in the lua profile so matching algorithm doesn't create uturns even when angle is large?

For now, I just commented lines 37 to 49 in /src/engine/plugins/match.cpp and it works well, but a parameter would be awesome.

Thanks!

daniel-j-h:
Not sure what the general idea behind this was cc @TheMarex - but it is possible to add such a parameter for sure. Here's how you can do it:

- [ ] adapt [filterCandidates](https://github.com/Project-OSRM/osrm-backend/blob/a5d0707dd0f081c6b4b98fdc36201567a2cfbbfe/src/engine/plugins/match.cpp#L30) to take a boolean
- [ ] on the [callsite](https://github.com/Project-OSRM/osrm-backend/blob/a5d0707dd0f081c6b4b98fdc36201567a2cfbbfe/src/engine/plugins/match.cpp#L170) pass a boolean based on the `MatchParameters` which `HandleRequests` gets
- [ ] add a parameter to [MatchParameters](https://github.com/Project-OSRM/osrm-backend/blob/a5d0707dd0f081c6b4b98fdc36201567a2cfbbfe/include/engine/api/match_parameters.hpp#L51) (in a backwards compatible way)
- [ ] add a parameter to the query parser: [MatchParametersGrammer](https://github.com/Project-OSRM/osrm-backend/blob/a5d0707dd0f081c6b4b98fdc36201567a2cfbbfe/include/server/api/match_parameter_grammar.hpp#L25)
- [ ] add tests to the [match grammar unit tests](https://github.com/Project-OSRM/osrm-backend/blob/a5d0707dd0f081c6b4b98fdc36201567a2cfbbfe/unit_tests/server/parameters_parser.cpp#L352)

Feel free to start this and open a PR early on where we can help you in case you're stuck :)

kaligrafy:
Thanks a lot!

In fact, the matching algorithm is quite useful to calculate shortest path when waypoints are approximate such as bus stops we need to serve in a 50m radius around a station location (this helps us determine the best precise location for each stop in a route during design phase). But as you know, allowing u-turns for buses is completely unrealistic. I'll prepare the PR. Thanks!

daniel-j-h:
Wow that's a cool use-case! Thanks for doing a PR!

kaligrafy:
[...]

PR #3608, opened 2017-01-24
Add allow_uturn parameter to match api

kaligrafy:
Related to #3595


include/engine/api/match_parameters.hpp
```diff
@@ -61,12 +62,16 @@ struct MatchParameters : public RouteParameters
     }
 
     template <typename... Args>
-    MatchParameters(std::vector<unsigned> timestamps_, Args... args_)
-        : RouteParameters{std::forward<Args>(args_)...}, timestamps{std::move(timestamps_)}
+    MatchParameters(
+        std::vector<unsigned> timestamps_,
+        const boost::optional<bool> allow_uturn_,
+        Args... args_)
+        : RouteParameters{std::forward<Args>(args_)...}, timestamps{std::move(timestamps_), allow_uturn{allow_uturn_}}
```
daniel-j-h: This needs to be done in a backwards compatible way: this is libosrm's public interface which we can't change without bumping the major version number.

You could add a constructor overload taking the new parameter and keeping the old one as is. Then add a note (and once we merged this PR open a ticket) for removing this backwards compatible workaround in v6. And tag it with the milestone v6.

include/engine/api/match_parameters.hpp
```diff
@@ -61,12 +62,16 @@ struct MatchParameters : public RouteParameters
...
     {
     }
 
     std::vector<unsigned> timestamps;
+    boost::optional<bool> allow_uturn = true;
```
daniel-j-h: Should this be an optional or a boolean with true being the default? Why the optional here?

include/server/api/match_parameter_grammar.hpp
```diff
@@ -32,6 +32,11 @@ struct MatchParametersGrammar final : public RouteParametersGrammar<Iterator, Si
             qi::lit("timestamps=") >
             (qi::uint_ %
              ';')[ph::bind(&engine::api::MatchParameters::timestamps, qi::_r1) = qi::_1];
+        
+        allow_uturn_rule =
+            qi::lit("allow_uturn=") >
+            (qi::uint_ %
+             ';')[ph::bind(&engine::api::MatchParameters::allow_uturn, qi::_r1) = qi::_1];
```
daniel-j-h: This needs a changelog entry, docs/http.md entry and more unit tests.

TheMarex:
Could we make this consistent with the `route` plugin? We already have a parameter called `continue_straight` which enables overriding the default u-turn behavior.

https://github.com/Project-OSRM/osrm-backend/blob/master/docs/http.md#route-service

The behavior should be as follows:
- `continue_straight` not specified by the user: auto detection of u-turns in the trace by map matching (current behavior)
- `continue_straight=false` same behavior as above
- `continue_straight=true` (aka `allow_uturns=false`) disable uturns detection in map matching plugin

kaligrafy:
If I understand correctly, the continue_straight parameter is already available in the match parameters since it is inherited, isn't it? So we just need to send it to the filterCandidates method?

daniel-j-h:
@kaligrafy Yes.

kaligrafy:
I don't know why, but the continue_straight is not passed from the route parameters parser. I guest it is because they are not include in the base rule in osrm-backend/include/server/api/route_parameters_grammar.hpp? Can somedy just tell me how to make sure it propagates to the match parameters parser?

oxidase:
@kaligrafy yes, it is just a `route` parameter. small change should fix it 
[a.txt](https://github.com/Project-OSRM/osrm-backend/files/730223/a.txt)

daniel-j-h:
The match grammar uses the route grammar as base for parsing

https://github.com/Project-OSRM/osrm-backend/blob/445225bf6dc719ef4f91fe6ab7e6fa2f4941fef7/include/server/api/match_parameter_grammar.hpp#L25

and the route grammar parses continue straight

https://github.com/Project-OSRM/osrm-backend/blob/445225bf6dc719ef4f91fe6ab7e6fa2f4941fef7/include/server/api/route_parameters_grammar.hpp#L34-L37

You can access continue straight via `parameters.continue_straight` here:

https://github.com/Project-OSRM/osrm-backend/blob/445225bf6dc719ef4f91fe6ab7e6fa2f4941fef7/src/engine/plugins/match.cpp#L116

I just tried it locally on master and this compiles just fine.

(Note: continue straight is a [optional](https://github.com/Project-OSRM/osrm-backend/blob/445225bf6dc719ef4f91fe6ab7e6fa2f4941fef7/include/engine/api/route_parameters.hpp#L107) you can check `if (opt) use_boolean(*opt)`)

kaligrafy:
It compiles but using it in a request triggers a query error.

Envoyé de mon iPhone

daniel-j-h:
Yes sorry for that: use [the patch](https://github.com/Project-OSRM/osrm-backend/files/730223/a.txt) @oxidase provided for now. We opened a ticket to keep parameters and grammars in sync in the future: https://github.com/Project-OSRM/osrm-backend/issues/3614

TheMarex:
I think it is fine if you copy the `continue_straight` rule from the route plugin to the match plugin.

kaligrafy:
I'll check that as soon as possible.

daniel-j-h:
What's the status here? Do you need help running this out @kaligrafy? Or where are you blocked?

kaligrafy:
No, I just have to finish a big project at the job and I have no time left for this right now. I'll check as soon as possible!

daniel-j-h:
👋 is this still on your todo list?

If not feel free to close this pull request, the issue already captures the task.

daniel-j-h:
Any updates here or can this be closed?

kaligrafy:
Yes, I prefer that somebody fix this since I do not have the time to do so right now.

Thanks!

PR #3612, opened 2017-01-25
Spatial Order + Rotating Slope for #3603

daniel-j-h:
For #3603.

Spatial order + trying cuts in parallel rotating the slope and taking best one.

### DIFF
--- a/include/partition/graph_view.hpp
+++ b/include/partition/graph_view.hpp
@@ -81,6 +81,9 @@ class GraphView
 
     NodeID GetTarget(const EdgeID eid) const;
 
+    const BisectionNode &GetNode(const NodeID nid) const;
+    const BisectionEdge &GetEdge(const EdgeID eid) const;
+
   private:
     const BisectionGraph &bisection_graph;
     const RecursiveBisectionState &bisection_state;
--- a/include/partition/inertial_flow.hpp
+++ b/include/partition/inertial_flow.hpp
@@ -2,6 +2,8 @@
 #define OSRM_PARTITION_INERTIAL_FLOW_HPP_
 
 #include "partition/graph_view.hpp"
+
+#include <unordered_set>
 #include <vector>
 
 namespace osrm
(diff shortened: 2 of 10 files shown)


include/partition/inertial_flow.hpp
```diff
@@ -2,20 +2,56 @@
...
+    // The node ids refer to nodes in the GraphView.
+    struct SpatialOrder
+    {
+        std::unordered_set<NodeID> sources;
+        std::unordered_set<NodeID> sinks;
+    };
+
+    // Creates a spatial order of n * sources "first" and n * sink "last" node ids.
+    // The slope determines the spatial order for sorting node coordinates.
+    SpatialOrder makeSpatialOrder(double ratio, double slope) const;
```
daniel-j-h: Capturing some trade-offs and ideas from chat here:
- Our typical `ratio` will be around .25 which means we have to hold .25\*nodes + .25\*nodes for sources and sinks, respectively, in memory. This can get quite large for planet sized graphs.
- We could re-order the nodes in-place (in the GraphView) but at the moment it is immutable and we want to keep it that way, so that we can run partitioning in parallel. We could `lock(); reorder(); unlock()` in the GraphView directly but this makes our abstractions ugly.

include/partition/inertial_flow.hpp
```diff
@@ -2,20 +2,56 @@
...
+};
+
+inline MinCut Dinic(const GraphView &graph,
+                    const std::unordered_set<NodeID> &sources,
+                    const std::unordered_set<NodeID> &sinks)
+{
+    return MinCut{10};
+}
+
+//
```
daniel-j-h: Please rip out and adapt when merging. This is really just a dummy impl. for me to plug in.

include/partition/graph_view.hpp
```diff
@@ -79,6 +79,9 @@ class GraphView
     EdgeIterator EdgeBegin(const NodeID nid) const;
     EdgeIterator EdgeEnd(const NodeID nid) const;
 
+    const BisectionNode &GetNode(const NodeID nid) const;
+    const BisectionEdge &GetEdge(const EdgeID eid) const;
```
daniel-j-h: I started adapting the graph we're holding on to in the graph view - we probably need more functionality here anyway, for the flow algorithm to operate on.

include/partition/inertial_flow.hpp
```diff
@@ -2,20 +2,56 @@
...
+
+struct MinCut
+{
+    std::int32_t num_edges;
+    // Should actually contain cut edges and/or lhs and rhs nodes
+};
+
+inline MinCut Dinic(const GraphView &graph,
+                    const std::unordered_set<NodeID> &sources,
+                    const std::unordered_set<NodeID> &sinks)
```
daniel-j-h: Can we quickly talk design for the flow algorithm here: should the flow algorithm do the sources / sink tracking internally, or should we contract sources and sinks (note: immutable graph), or should we put a flag into the bisection graph nodes?

src/partition/inertial_flow.cpp
```diff
@@ -7,16 +20,105 @@ namespace partition
...
+        }
+
+        NodeID nid;
+        util::Coordinate coordinate;
+    };
+
+    using Embedding = std::vector<NodeWithCoordinate>;
+
+    Embedding embedding;
+    embedding.reserve(view.NumberOfNodes());
```
daniel-j-h: This is what I was talking about above: with a planet sized graph we store node ids and coordinates here for the spatial order re-ordering to take place and then return the 0.25*n "first" and "last" node ids.

MoKob:
Merged into graph view branch. Had to do some minor adjustments to make it work together. But we know have parallel partitioning using Dinic algorithm.

PR #3613, opened 2017-01-25
[Isochrone plugin] Add start edge if it is a bidirectional to avoid the formation of several groups.

fijemax:
To prevent this case : 
http://bl.ocks.org/anonymous/raw/94631c4ea04286efc1fb6f03c69d3744/
i add the start edge if it is bidirectional.


danpat:
@fijemax I re-based my branch on `master`, can you re-based yours so we can merge this fix in?

fijemax:
Hi @danpat ,
My pull request is wrong, I will close it.
I will explain that in #2189.

PR #3615, opened 2017-01-25
move common code from car and foot profiles to helper file

emiltin:
# Issue

Factors out common code from car and foot profile, and moves it to the helper file profiles/lib/handlers.lua.

Different handlers take care of aspects like oneways, speed, access tags, etc. A profile can choose which handlers to call. Handlers don't have to be used and can also be combine with custom code if desired.

Points:

- car and foot profile now rely only on these handlers, and do not require any custom code. Difference is only in the settings, and the list of handlers called.
- all profile settings are now located in a single table, named 'profile', which is passed to the handlers.
- Handlers.run() is a convenience function for calling a list of handlers. All it does it call each and abort if a handler returns false.
- handle_prefetch() (previously called initial_routability_check) prefetches a list of tags and stores them in lua, so they can be reused later. if none of the listed tags are present, the way is considered non-routable, and processing is aborted.
- it might be better to separate prefetch into prefetch and abort_if_obviosly_not_routable.
- extracting data for copenhagen is about 5% slower for me than master branch. most of this seems to be due to the changes to handle_prefetch. This handler is called for every single way (since it's where the initial check for non-routability happen) so even a small change in performance here has a significant effect.
- the Directional helper was renamed Tags, because it now included other useful ways to search/parse tags, which are not directly related to forward/backward tags.

All tests pass.

## Tasklist
 - [x] investigate/fix slow down
 - [x] review
 - [ ] adjust for comments

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,8 @@
       - `properties.uturn_penalty` is deprecated. Set it in the `turn_function`. The turn type is exposed as `ExtractionTurn::direction_modifier`.
       - `properties.traffic_light_penalty` is deprecated. Traffic light penalties now need to be set over in the turn function.
          Each turn with a traffic light is marked with `ExtractionTurn::has_traffic_light = true`.
+      - Renamed the helper file `profiles/lib/directional.lua` to `profiles/lib/tags.lua` since it now provides more general tags parsing utility functions.
+      - The car and foot profiles now depend on the helper file `profiles/lib/handlers.lua`.
     - Infrastructure
       - Disabled link-time optimized (LTO) builds by default. Enable by passing `-DENABLE_LTO=ON` to `cmake` if you need the performance and know what you are doing.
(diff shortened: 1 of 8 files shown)


emiltin:
rebased on master.
extraction is about 5% slower than the master branch, but it also handles additonal edge case like oneway:vehicle in the car profile. i added a test to reflect these oneway cases.

emiltin:
want me to add a note to the changelog? would that go under 5.6.0?

the file lib/directional.lua was renamed to lib/tags.lua, maybe that change in dependency should be noted as well?

emiltin:
i removed changes to debug_example.lua. it was just noise and not related to this PR.

daniel-j-h:
@emiltin yes please add a changelog under a 5.6 section. Master contains the latest code which will become the next stable release (5.6).

We want to merge this in asap since it's blocking https://github.com/Project-OSRM/osrm-backend/pull/3632 (which triggers an assertion the demo server).

emiltin:
added changelog notes and rebased on master

PR #3620, opened 2017-01-27
Implements Compressed Node Based Graph (De-)Serialization

daniel-j-h:
For https://github.com/Project-OSRM/osrm-backend/pull/3603. Asynchronously dumping the compressed node based graph to disk with

    osrm-extract --dump-partition-graph

Then reading it back in with `osrm-partition`.

- [ ] Once merged into master document the new `.osrm.cnbg` in our Wiki file listing


MoKob:
Merged into partition branch

PR #3623, opened 2017-01-27
Update http.md

ghoshkaj:
# Issue

Change the word `field` to `property` for accuracy.

### DIFF
--- a/docs/http.md
+++ b/docs/http.md
@@ -373,7 +373,7 @@ Vector tiles contain two layers:
 
 `speeds` layer:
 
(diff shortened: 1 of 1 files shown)


danpat:
@ghoshkaj hmm, I don't think we've ever done a field-vs-property editing pass over these docs - I can see lots of uses of both.

Want to normalize everything to `property` in this PR?

ghoshkaj:
@danpat okay I think I've got all the fields, and they make sense to turn to properties.

Issue #3619, opened 2017-01-27
assertion/deadlock when starting second `osrm-routed` consumer of shared memory block on OSX


danpat:
When compiled with `-DCMAKE_BUILD_TYPE=Debug`, the following steps trigger an assertion from inside boost in one process, and a deadlock in another, on OSX:

```
#!/bin/sh

./osrm-datastore berlin.osrm
./osrm-routed -s &
./osrm-routed -s  # assertion here
```

The assertion that's thrown is:

```
$ ./osrm-routed -s
[info] starting up engines, v5.5.0
[info] Loading from shared memory
[info] Threads: 8
[info] IP address: 0.0.0.0
[info] IP port: 5000
[info] http 1.1 compression handled by zlib version 1.2.5
[info] Listening on: 0.0.0.0:5000
[debug] opening 76939265 from id 0
[debug] Loading new data for region REGION_2
[debug] opening 34471938 from id 2
[info] set checksum: 4127811523
[debug] opening 76939265 from id 0
[info] running and waiting for re[assert][q0x700000081000u] e/Users/danpat/mapbox/osrm-backend/mason_packages/headers/boost/1.63.0/include/boost/interprocess/sync/posix/condition.hpps:t175s
in: oid boost::interprocess::ipcdetail::posix_condition::do_wait(boost::interprocess::ipcdetail::posix_mutex &)[: 0res == 0m

libc++abi.dylib: terminating
Abort trap: 6
```

In addition, when attempting to kill the first `osrm-routed` process with CTRL+C, it fails to exit, hanging with this output:

```
^C[info] initiating shutdown
[info] stopping threads
[info] freeing objects
```

When run under `Release` mode, the assertion inside BOOST does not trigger, and everything looks normal (both `osrm-routed -s` processes continuously swap data and answer queries).

Note: the same behaviour is seen with `node-osrm` and instantiating two `OSRM()` objects.

danpat:
Output from LLDB at the assertion point:

```
(lldb) n
Process 35936 stopped
* thread #2: tid = 0x1ae45b, 0x00000001002201cf osrm-routed`boost::interprocess::ipcdetail::posix_condition::do_wait(this=0x0000000101297050, mut=0x0000000101297010) + 47 at condition.hpp:175, stop reason = step over
[...]

PR #3627, opened 2017-01-30
Make shared memory region mapping at fixed address for OSX

oxidase:
# Issue

For #3619 changed `named_mutex` and `named_condition` that are mapped with 0 (no preference) user space address to `interprocess_mutex` and `interprocess_condition` that are tightly bound and located in  `shared_memory_object` that is mapped to fixed address for OSX to prevent `EINVAL` error in `pthread_cond_wait`  call. Also `CURRENT_REGION` data is moved into `SharedBarrier::internal_data`.

[Fixed memory address](https://github.com/Project-OSRM/osrm-backend/blob/fix/osx-fixed-address-mapping/include/storage/shared_barrier.hpp#L124) is arbitrary but it should not overlap with the OSX process map: it is close and below `STACK GUARD            0000700000000000-0000700000001000 [    4K] ---/rwx SM=NUL  stack guard for thread 1` .

It must be adjusted to some value to avoid possible `boost::interprocess_exception::library_error` errors.

/cc @TheMarex @danpat 

## Tasklist
 - [x] review
 - [x] adjust for comments

## Requirements / Relations


oxidase:
Even with tests ok, PR should be checked directly with OSX. My try on travis https://travis-ci.org/oxidase/node-osrm/builds/196566346#L1171-L1176 failed.

/cc @danpat

oxidase:
Figured out the difference between `osrm-backend` tests that pass and `node-osrm` tests that fail for OSX:
*  `node-osrm` tests instantiate multiple `Engine` instances
* `SharedBarrier` uses address 00000caffee00000 as a hint
* the first block is mapped to 00000caffee00000
* the second block is mapped 00000caffee01000
* exception is thrown due to the failing condition at https://github.com/boostorg/interprocess/blob/develop/include/boost/interprocess/mapped_region.hpp#L717

Using `MAP_FIXED` in `mmap`  fixes the exception but brings a segmentation fault, because the first destructor unmaps the memory at https://github.com/boostorg/interprocess/blob/develop/include/boost/interprocess/mapped_region.hpp#L841 and the second destructor call access the unmapped block.

Will try to reuse XSI memory via makeSharedRegion.

oxidase:
XSI memory attachment also fails at [`shmat` call](https://github.com/boostorg/interprocess/blob/develop/include/boost/interprocess/mapped_region.hpp#L627), because it is not allowed to map two memory blocks to one memory address.

At the moment i see only one way how to solve a resource contention problem: use a singleton watchdog.
I don't like this solution, because life-time of the watchdog now is undefined and depends on the initialization order, but have no better solution how to make interprocess communication in OSX with multiple process consumers and multiple engines per process.  

The following build `node-osrm` build https://travis-ci.org/oxidase/node-osrm/builds/196851065 should be green before merging.

include/storage/shared_barrier.hpp
```diff
@@ -0,0 +1,127 @@
...
+
+    static void Remove() { bi::shared_memory_object::remove(block_name); }
+
+  private:
+    bi::shared_memory_object shmem;
+    bi::mapped_region region;
+
+    static constexpr const char *const block_name = "osrm-region";
+    static constexpr std::size_t block_size = sizeof(internal_data);
+#if defined(__APPLE__)
```
TheMarex: We should add a comment here why we need this behavior on OSX. Maybe with link to the issue.

src/engine/engine.cpp
```diff
@@ -22,27 +18,28 @@
 
 namespace
 {
+
+auto GetWatchdogDataFacade()
```
TheMarex: I agree a singleton is an ugly solution here, however it has at least the upside that we only have one update thread per process.

Thinking ahead a little bit, I'm not sure how we could extend this to multiple shared-memory regions (to have multiple datasets in shared memory at the same time).

Maybe we could circumvent the need for a central control region with message based IPC for updates and a control file for persistent storage. I'm going to open an issue about this, maybe we can come up with more ideas. For now this is fine. 👍

oxidase:
added a comment line.

daniel-j-h:
Do we have a upstream ticket for this yet? Should eventually be fixed in Boost directly.

oxidase:
No, there is no boost ticket for this. I don't think it is a boost issue anymore, but OSX known behavour. I even don't know how to fix it in general way in boost: the current change is a workaround that introduces the fixed address mapping and the watchdog singleton. On other side there is no problem with a single consumer process, only if more than one processes are running. 

Singletons are really bad beasts and must be removed soon. Anyway with multiple data updates, like independent MLD and CH data, the current approach over a control block "condvar/mutex" is badly scalable.

Issue #3563, opened 2017-01-12
Allow annotations parameter to accept specific fields


karenzshea:
The `annotations` parameter currently accepts a boolean and returns all the fields `duration, distance, nodes, datasources`. It could be useful to accept requests for only specific fields, i.e. accept `annotations=duration` and only return a `{ duration: 10, 5, 20 }` in the returned annotations object.

Leaving `annotations=true|false` to return all fields will ensure the changes are only additive.

daniel-j-h:
The use-case is to trim down the response size? Because we have the data at hand anyway. I guess you want to accept a list of those enums or a boolean as in `annotations=duration,distance`.

Let's spec out what needs to be done for this:
- [ ] For [RouteParameters](https://github.com/Project-OSRM/osrm-backend/blob/fd57c5b48bd289ecf15ffa7ea703f0461536f784/include/engine/api/route_parameters.hpp#L56) we need to extend the boolean annotations flag to an enum with all the fields you want to include. We need to keep the old constructor and boolean for backwards compatibility. Check the `GeometriesType` enum a few lines above
- [ ] For [RouteParametersGrammar](https://github.com/Project-OSRM/osrm-backend/blob/fd57c5b48bd289ecf15ffa7ea703f0461536f784/include/server/api/route_parameters_grammar.hpp#L57-L58) which parses the user input we need to extend the rules from a boolean to a boolean or a list of enum values (check the `.add()` a few lines above and the `GeometriesType` enum for how to do this)
- [ ] For the [route api](https://github.com/Project-OSRM/osrm-backend/blob/fd57c5b48bd289ecf15ffa7ea703f0461536f784/include/engine/api/route_api.hpp#L218-L253) we can then conditionally add specific annotation fields

Want to run this out @karenzshea?

daniel-j-h:
Note: `speed` along the way would be great to have, too.

danpat:
@daniel-j-h WIP here https://github.com/mapbox/api-directions/pull/896

TheMarex:
Breaking out the speed addition to https://github.com/Project-OSRM/osrm-backend/issues/3664

PR #3628, opened 2017-01-30
Refactor annotations parameter to accept explicit values

karenzshea:
# Issue

Closes https://github.com/Project-OSRM/osrm-backend/issues/3563

## Tasklist
-  [ ] ~~Adapt annotations parsing for map matching route~~ matching inherits from route
 - [x] Abstract annotation insertion to steps into a function
 - [x] add cucumber cases (see docs/testing.md)
 - [x] review
 - [x] adjust for comments

### DIFF
--- a/features/step_definitions/matching.js
+++ b/features/step_definitions/matching.js
@@ -83,7 +83,10 @@ module.exports = function () {
                             duration = json.matchings[0].duration;
                         }
 
-                        if (headers.has('annotation')) {
+                        // annotation response values are requested by 'a:{type_name}'
+                        var found = false;
+                        headers.forEach((h) => { if (h.match(/^a:/)) found = true; });
+                        if (found) {
                             if (json.matchings.length != 1) throw new Error('*** Checking annotation only supported for matchings with one subtrace');
                             annotation = this.annotationList(json.matchings[0]);
                         }
@@ -91,11 +94,6 @@ module.exports = function () {
                         if (headers.has('geometry')) {
                             if (json.matchings.length != 1) throw new Error('*** Checking geometry only supported for matchings with one subtrace');
                             geometry = json.matchings[0].geometry;
-                        }
-
-                        if (headers.has('OSM IDs')) {
-                            if (json.matchings.length != 1) throw new Error('*** Checking annotation only supported for matchings with one subtrace');
-                            OSMIDs = this.OSMIDList(json.matchings[0]);
                         }
                     }
 
@@ -111,9 +109,19 @@ module.exports = function () {
                         got.duration = duration.toString();
                     }
 
-                    if (headers.has('annotation')) {
-                        got.annotation = annotation.toString();
-                    }
+                    // if header matches 'a:*', parse out the values for *
+                    // and return in that header
+                    headers.forEach((k) => {
+                        let whitelist = ['duration', 'distance', 'datasources', 'nodes', 'weight'];
+                        if (k.match(/^a:/)) {
+                            let a_type = k.slice(2);
+                            if (whitelist.indexOf(a_type) == -1)
+                                return cb(new Error('Unrecognized annotation field:' + a_type));
+                            if (!annotation[a_type])
+                                return cb(new Error('Annotation not found in response: ' + a_type));
+                            got[k] = annotation[a_type];
+                        }
+                    });
 
                     if (headers.has('geometry')) {
                         if (this.queryParams['geometries'] === 'polyline') {
--- a/features/support/shared_steps.js
+++ b/features/support/shared_steps.js
@@ -133,9 +133,19 @@ module.exports = function () {
                                 got.locations = (locations || '').trim();
                             }
 
-                            if (headers.has('annotation')){
-                                got.annotation = (annotation || '').trim();
-                            }
+                            // if header matches 'a:*', parse out the values for *
+                            // and return in that header
+                            headers.forEach((k) => {
+                                let whitelist = ['duration', 'distance', 'datasources', 'nodes', 'weight'];
+                                if (k.match(/^a:/)) {
+                                    let a_type = k.slice(2);
+                                    if (whitelist.indexOf(a_type) == -1)
+                                        return cb(new Error('Unrecognized annotation field', a_type));
+                                    if (!annotation[a_type])
+                                        return cb(new Error('Annotation not found in response', a_type));
+                                    got[k] = annotation[a_type];
+                                }
+                            });
 
                             var putValue = (key, value) => {
                                 if (headers.has(key)) got[key] = instructions ? value : '';
--- a/features/testbot/matching.feature
+++ b/features/testbot/matching.feature
@@ -142,7 +142,7 @@ Feature: Basic Map Matching
 
         When I match I should get
             | trace | matchings | annotation    |
-            | ach   | ach       | 1:1,0:1:1:2:1 |
+            | ach   | ach       | 1,1,0,1,1,2,1 |
 
     Scenario: Testbot - Duration details
         Given the query options
@@ -167,9 +167,9 @@ Feature: Basic Map Matching
         And the contract extra arguments "--segment-speed-file {speeds_file}"
 
         When I match I should get
-            | trace | matchings | annotation                                                                                                      |
-            | abeh  | abeh      | 1:1:10.008842:1,0:0:0:0,1:1:10.008842:0,1:1:10.008842:0,1:1:10.008842:0,0:0:0:0,2:2:19.906475:0,1:1:10.008842:0 |
-            | abci  | abci      | 1:1:10.008842:1,0:0:0:0,1:1:10.008842:0,0:0:0:0,1:1:10.010367:0                                                 |
+            | trace | matchings | annotation                                                                                                                      |
+            | abeh  | abeh      | 1:10.008842:1:1:1,0:0:0:2:0,1:10.008842:1:2:0,1:10.008842:1:3:0,1:10.008842:1:4:0,0:0:0:4:0,2:19.906475:2:5:0,1:10.008842:1:6:0 |
+            | abci  | abci      | 1:10.008842:1:1:1,0:0:0:2:0,1:10.008842:1:2:0,0:0:0:2:0,1:10.010367:1:3:0                                                       |
 
         # The following is the same as the above, but separated for readability (line length)
         When I match I should get
@@ -388,4 +388,4 @@ Feature: Basic Map Matching
         When I match I should get
             | trace | annotation   |
             | ac    | 2:2,2:2      |
-            | ca    | 2:2:0,2:2:0  |
+            | ca    | 2:2,2:2,0:0  |
--- a/features/testbot/matching.feature
+++ b/features/testbot/matching.feature
@@ -141,12 +141,12 @@ Feature: Basic Map Matching
         And the contract extra arguments "--segment-speed-file {speeds_file}"
 
         When I match I should get
-            | trace | matchings | annotation    |
-            | ach   | ach       | 1,1,0,1,1,2,1 |
+            | trace | matchings | a:duration       |
+            | ach   | ach       | 1:1,0:1:1:2:1    |
 
     Scenario: Testbot - Duration details
         Given the query options
-            | annotations | true    |
+            | annotations | duration,nodes |
 
         Given the node map
             """
@@ -167,15 +167,15 @@ Feature: Basic Map Matching
         And the contract extra arguments "--segment-speed-file {speeds_file}"
 
         When I match I should get
-            | trace | matchings | annotation                                                                                                                      |
-            | abeh  | abeh      | 1:10.008842:1:1:1,0:0:0:2:0,1:10.008842:1:2:0,1:10.008842:1:3:0,1:10.008842:1:4:0,0:0:0:4:0,2:19.906475:2:5:0,1:10.008842:1:6:0 |
-            | abci  | abci      | 1:10.008842:1:1:1,0:0:0:2:0,1:10.008842:1:2:0,0:0:0:2:0,1:10.010367:1:3:0                                                       |
+            | trace | matchings | a:duration      |
+            | abeh  | abeh      | 1:0,1:1:1,0:2:1 |
+            | abci  | abci      | 1:0,1,0:1       |
 
         # The following is the same as the above, but separated for readability (line length)
         When I match I should get
-            | trace | matchings | OSM IDs               |
-            | abeh  | abeh      | 1,2,3,2,3,4,5,4,5,6,7 |
-            | abci  | abci      | 1,2,3,2,3,2,3,8       |
+            | trace | matchings | a:nodes               |
+            | abeh  | abeh      | 1:2:3,2:3:4:5,4:5:6:7 |
+            | abci  | abci      | 1:2:3,2:3,2:3:8       |
 
     Scenario: Testbot - Regression test for #3037
         Given the query options
@@ -303,7 +303,7 @@ Feature: Basic Map Matching
             | abcd  | 0 1 2 3    | abcd      |
 
     # Regression test 1 for issue 3176
-    Scenario: Testbot - multiuple segments: properly expose OSM IDs
+    Scenario: Testbot - multiple segments: properly expose OSM IDs
         Given the query options
             | annotations | true    |
 
@@ -332,9 +332,9 @@ Feature: Basic Map Matching
             | fg    | no     |
 
         When I match I should get
-            | trace | OSM IDs       |
-            | 12    | 1,2,3,4,5,6,7 |
-            | 21    | 7,6,5,4,3,2,1 |
+            | trace | a:nodes       |
+            | 12    | 1:2:3:4:5:6:7 |
+            | 21    | 7:6:5:4:3:2:1 |
 
     # Regression test 2 for issue 3176
     Scenario: Testbot - same edge: properly expose OSM IDs
@@ -360,32 +360,6 @@ Feature: Basic Map Matching
             | abcdef  | no     |
 
         When I match I should get
-            | trace | OSM IDs     |
-            | 12    | 1,2,3,4,5,6 |
-            | 21    | 6,5,4,3,2,1 |
-
-    Scenario: Testbot - return annotations based on parameter
-        Given the query options
-            | annotations | duration,weight |
-
-        Given the node map
-            """
-            a - b
-                |
-                c
-            """
-
-        And the nodes
-            | node | id |
-            | a    | 1  |
-            | b    | 2  |
-            | c    | 3  |
-
-        And the ways
-            | nodes | oneway |
-            | abc   | no     |
-
-        When I match I should get
-            | trace | annotation   |
-            | ac    | 2:2,2:2      |
-            | ca    | 2:2,2:2,0:0  |
+            | trace | a:nodes     |
+            | 12    | 1:2:3:4:5:6 |
+            | 21    | 6:5:4:3:2:1 |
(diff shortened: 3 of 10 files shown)


features/testbot/matching.feature
```diff
@@ -141,9 +167,9 @@ Feature: Basic Map Matching
         And the contract extra arguments "--segment-speed-file {speeds_file}"
 
         When I match I should get
-            | trace | matchings | annotation                                                                                                      |
-            | abeh  | abeh      | 1:1:10.008842:1,0:0:0:0,1:1:10.008842:0,1:1:10.008842:0,1:1:10.008842:0,0:0:0:0,2:2:19.906475:0,1:1:10.008842:0 |
-            | abci  | abci      | 1:1:10.008842:1,0:0:0:0,1:1:10.008842:0,0:0:0:0,1:1:10.010367:0                                                 |
+            | trace | matchings | annotation                                                                                                                      |
+            | abeh  | abeh      | 1:10.008842:1:1:1,0:0:0:2:0,1:10.008842:1:2:0,1:10.008842:1:3:0,1:10.008842:1:4:0,0:0:0:4:0,2:19.906475:2:5:0,1:10.008842:1:6:0 |
+            | abci  | abci      | 1:10.008842:1:1:1,0:0:0:2:0,1:10.008842:1:2:0,0:0:0:2:0,1:10.010367:1:3:0                                                       |
```
daniel-j-h: What are the changes here?

features/testbot/matching.feature
```diff
@@ -118,6 +118,32 @@ Feature: Basic Map Matching
...
+        And the speed file
+        """
+        1,2,36
+        """
+
+        And the contract extra arguments "--segment-speed-file {speeds_file}"
+
+        When I match I should get
+            | trace | matchings | annotation    |
+            | ach   | ach       | 1,1,0,1,1,2,1 |
```
daniel-j-h: The annotations here are durations, right? How do I read those in the context of the node map? Why is there a duration 0 in there?

features/testbot/matching.feature
```diff
@@ -337,3 +363,29 @@ Feature: Basic Map Matching
...
+            a - b
+                |
+                c
+            """
+
+        And the nodes
+            | node | id |
+            | a    | 1  |
+            | b    | 2  |
+            | c    | 3  |
```
daniel-j-h: Why is this needed?

include/engine/api/route_api.hpp
```diff
@@ -215,43 +215,71 @@ class RouteAPI : public BaseAPI
...
-                              [this, &nodes](const OSMNodeID &node_id) {
-                                  nodes.values.push_back(static_cast<std::uint64_t>(node_id));
-                              });
                 util::json::Object annotation;
-                annotation.values["distance"] = std::move(distances);
-                annotation.values["duration"] = std::move(durations);
-                annotation.values["weight"] = std::move(weights);
-                annotation.values["nodes"] = std::move(nodes);
-                annotation.values["datasources"] = std::move(datasources);
+                if (parameters.annotations_type & RouteParameters::AnnotationsType::Duration)
```
daniel-j-h: Hm either you do the bit fiddling here or what about abstracting it away? The code here could work no matter how the `AnnotationsType` is implemented.

include/engine/api/route_api.hpp
```diff
@@ -215,43 +215,71 @@ class RouteAPI : public BaseAPI
...
+                {
+                    util::json::Array nodes;
+                    nodes.values.reserve(leg_geometry.osm_node_ids.size());
+                    std::for_each(leg_geometry.osm_node_ids.begin(),
+                                  leg_geometry.osm_node_ids.end(),
+                                  [this, &nodes](const OSMNodeID &node_id) {
+                                      nodes.values.push_back(static_cast<std::uint64_t>(node_id));
+                                  });
+                    annotation.values["nodes"] = std::move(nodes);
+                }
```
daniel-j-h: Hm the ~100 lines above are all doing the same:
- check if annotation x was requested
- create `json::Array` for many x
- extract `step.x` from `RouteStep` and add into array
- add array to annotations response as `"x" : [...]`

I think you could abstract over this and make it extensible for the future at the same time.

What about abstracting this behind a function which takes
- a `AnnotationsType`
- a lambda taking a `RouteStep` and returning annotations items out of it (e.g. `step.duration` for `AnnotationsType::Duration`
- a string describing the annotations type for the json object's key as in `"key": [...]` above

Example callsite:

```c++
makeAnnotation(AnnotationsType::Duration, "duration", [](auto& step){ return step.duration });
```

which then does all of the steps above behind the scenes and wires the parts together.

include/engine/api/route_parameters.hpp
```diff
@@ -67,6 +67,16 @@ struct RouteParameters : public BaseParameters
         Full,
         False
     };
+    enum class AnnotationsType : int
```
daniel-j-h: Why the explicit `: int` here? Should be `int` by default.

include/engine/api/route_parameters.hpp
```diff
@@ -94,20 +104,54 @@ struct RouteParameters : public BaseParameters
...
+    RouteParameters(const bool steps_,
+                    const bool alternatives_,
+                    const AnnotationsType annotations_,
+                    const GeometriesType geometries_,
+                    const OverviewType overview_,
+                    const boost::optional<bool> continue_straight_,
+                    Args... args_)
+        : BaseParameters{std::forward<Args>(args_)...}, steps{steps_}, alternatives{alternatives_},
+          annotations_type{annotations_}, annotations{true}, geometries{geometries_},
+          overview{overview_}, continue_straight{continue_straight_}
```
daniel-j-h: Nice :+1: This should keep the API stable for the time being. We may want to ticket the removal of all those hacks for v6. I think we already have tickets for these kinds of hacks.

include/server/api/route_parameters_grammar.hpp
```diff
@@ -50,17 +57,25 @@ struct RouteParametersGrammar : public BaseParametersGrammar<Iterator, Signature
             "full", engine::api::RouteParameters::OverviewType::Full)(
             "false", engine::api::RouteParameters::OverviewType::False);
 
+        annotations_type.add("true", engine::api::RouteParameters::AnnotationsType::All)(
+            "false", engine::api::RouteParameters::AnnotationsType::None)(
+            "duration", engine::api::RouteParameters::AnnotationsType::Duration)(
+            "nodes", engine::api::RouteParameters::AnnotationsType::Nodes)(
+            "distance", engine::api::RouteParameters::AnnotationsType::Distance)(
+            "weight", engine::api::RouteParameters::AnnotationsType::Weight)(
+            "datasources", engine::api::RouteParameters::AnnotationsType::Datasources);
```
daniel-j-h: Hm we need a mapping `AnnotationsType` -> `string` here, too. The same is needed for the response generation (as commented above). Maybe we should add this mapping already next to the `AnnotationsType` enum class definition. This would guarantee that we're in sync with name and enum item.

include/server/api/route_parameters_grammar.hpp
```diff
@@ -50,17 +57,25 @@ struct RouteParametersGrammar : public BaseParametersGrammar<Iterator, Signature
...
-            (qi::lit("annotations=") >
-             qi::bool_[ph::bind(&engine::api::RouteParameters::annotations, qi::_r1) = qi::_1]) |
             (qi::lit("geometries=") >
              geometries_type[ph::bind(&engine::api::RouteParameters::geometries, qi::_r1) =
                                  qi::_1]) |
             (qi::lit("overview=") >
-             overview_type[ph::bind(&engine::api::RouteParameters::overview, qi::_r1) = qi::_1]);
+             overview_type[ph::bind(&engine::api::RouteParameters::overview, qi::_r1) = qi::_1]) |
+            (qi::lit("annotations=") >
+             annotations_type[ph::bind(add_annotation, qi::_r1, qi::_1)] % ',');
```
daniel-j-h: Is an empty list allowed by this? (I don't know => make a unit test)

unit_tests/server/parameters_parser.cpp
```diff
@@ -321,6 +325,37 @@ BOOST_AUTO_TEST_CASE(valid_route_urls)
...
+        parseParameters<RouteParameters>("1,2;3,4?geometries=polyline&"
+                                         "overview=simplified&annotations=duration,weight,nodes");
+    BOOST_CHECK(result_15);
+    BOOST_CHECK_EQUAL(reference_15.geometries, result_15->geometries);
+    BOOST_CHECK_EQUAL(
+        static_cast<bool>(result_2->annotations_type & (RouteParameters::AnnotationsType::Duration |
+                                                        RouteParameters::AnnotationsType::Weight |
+                                                        RouteParameters::AnnotationsType::Nodes)),
+        true);
+    BOOST_CHECK_EQUAL(result_15->annotations, true);
```
daniel-j-h: Order should not matter - care to create a test expressing this?
Also edge cases such as
- empty list
- all possibilities
- all possibilities including All, None and combinations

include/engine/api/route_api.hpp
karenzshea: :+1: refactoring this part is already in the OP as the todo item `Abstract annotation insertion to steps into a function`

I was going to take a template based approach to the function, but I can consider your suggestion as possible solution as well.

include/server/api/route_parameters_grammar.hpp
oxidase: using karma or a map value->string? would it make sense to make json output completely in karma?

include/engine/api/route_parameters.hpp
```diff
@@ -67,6 +67,16 @@ struct RouteParameters : public BaseParameters
         Full,
         False
     };
+    enum class AnnotationsType : int
+    {
+        None = 0,
+        Duration = 1 << 1,
```
oxidase: here it can start from `1<<0`, but better to use explicit  numerals 0x01, 0x02, ...

include/server/api/route_parameters_grammar.hpp
daniel-j-h: a value->string mapping would be good enough for now - our json handling is all based on a variant, not sure how karma fits in there

include/server/api/route_parameters_grammar.hpp
karenzshea: Added! to the parameters_parser unit test

features/support/shared_steps.js
```diff
@@ -133,9 +133,19 @@ module.exports = function () {
...
+                                let whitelist = ['duration', 'distance', 'datasources', 'nodes', 'weight'];
+                                if (k.match(/^a:/)) {
+                                    let a_type = k.slice(2);
+                                    if (whitelist.indexOf(a_type) == -1)
+                                        return cb(new Error('Unrecognized annotation field', a_type));
+                                    if (!annotation[a_type])
+                                        return cb(new Error('Annotation not found in response', a_type));
+                                    got[k] = annotation[a_type];
+                                }
+                            });
```
daniel-j-h: Hm this is dup. code with above

include/engine/api/route_api.hpp
```diff
@@ -79,6 +79,19 @@ class RouteAPI : public BaseAPI
...
     }
 
+    template <typename GetFn>
+    util::json::Array GetAnnotations(const guidance::LegGeometry &leg, GetFn Get) const
+    {
+        util::json::Array annotations_store;
+        annotations_store.values.reserve(leg.annotations.size());
+        std::for_each(leg.annotations.begin(),
+                      leg.annotations.end(),
+                      [Get, &annotations_store](const guidance::LegGeometry::Annotation &step) {
```
daniel-j-h: I think you can simply say `const auto& step` here

include/engine/api/route_api.hpp
```diff
@@ -215,43 +228,54 @@ class RouteAPI : public BaseAPI
...
-                annotation.values["weight"] = std::move(weights);
-                annotation.values["nodes"] = std::move(nodes);
-                annotation.values["datasources"] = std::move(datasources);
+
+                if (parameters.annotations_type & RouteParameters::AnnotationsType::Duration)
+                {
+                    annotation.values["duration"] =
+                        GetAnnotations(leg_geometry,
+                                       std::bind(&guidance::LegGeometry::Annotation::duration,
+                                                 std::placeholders::_1));
```
daniel-j-h: Instead of bind you can simply use a lambda, which is probably easier to read and does not have all the weird quirks bind has (see https://www.youtube.com/watch?v=zt7ThwVfap0). E.g.

```cpp
[](const auto& v) { return v.distance };
```

features/step_definitions/matching.js
```diff
@@ -111,9 +109,19 @@ module.exports = function () {
                         got.duration = duration.toString();
                     }
 
-                    if (headers.has('annotation')) {
-                        got.annotation = annotation.toString();
-                    }
+                    // if header matches 'a:*', parse out the values for *
+                    // and return in that header
+                    headers.forEach((k, v) => {
```
oxidase: :+1: but lint complains about unused `v` variable

features/support/shared_steps.js
```diff
@@ -133,9 +133,19 @@ module.exports = function () {
                                 got.locations = (locations || '').trim();
                             }
 
-                            if (headers.has('annotation')){
-                                got.annotation = (annotation || '').trim();
-                            }
+                            // if header matches 'a:*', parse out the values for *
+                            // and return in that header
+                            headers.forEach((k, v) => {
```
oxidase: and here `v` is unused

unit_tests/server/parameters_parser.cpp
```diff
@@ -128,6 +132,9 @@ BOOST_AUTO_TEST_CASE(valid_route_urls)
     CHECK_EQUAL_RANGE(reference_2.radiuses, result_2->radiuses);
     CHECK_EQUAL_RANGE(reference_2.coordinates, result_2->coordinates);
     CHECK_EQUAL_RANGE(reference_2.hints, result_2->hints);
+    BOOST_CHECK_EQUAL(
+        static_cast<bool>(result_2->annotations_type & RouteParameters::AnnotationsType::All),
+        true);
```
daniel-j-h: Hm maybe add `operator==` overload then? We already have `|` and `&`.

unit_tests/server/parameters_parser.cpp
```diff
@@ -321,6 +328,58 @@ BOOST_AUTO_TEST_CASE(valid_route_urls)
...
+    RouteParameters reference_17{};
+    reference_17.annotations_type = RouteParameters::AnnotationsType::All;
+    reference_17.coordinates = coords_1;
+    auto result_17 = parseParameters<RouteParameters>(
+        "1,2;3,4?overview=simplified&annotations=duration,weight,nodes,datasources,distance");
+    BOOST_CHECK(result_17);
+    BOOST_CHECK_EQUAL(reference_17.geometries, result_17->geometries);
+    BOOST_CHECK_EQUAL(
+        static_cast<bool>(result_2->annotations_type & RouteParameters::AnnotationsType::All),
+        true);
```
daniel-j-h: Same for all of these tests, `type == All`

include/engine/api/route_parameters.hpp
```diff
@@ -94,20 +104,54 @@ struct RouteParameters : public BaseParameters
...
+                    Args... args_)
+        : BaseParameters{std::forward<Args>(args_)...}, steps{steps_}, alternatives{alternatives_},
+          annotations_type{annotations_}, annotations{true}, geometries{geometries_},
+          overview{overview_}, continue_straight{continue_straight_}
     {
     }
 
     bool steps = false;
     bool alternatives = false;
     bool annotations = false;
```
oxidase: this is now a duplication, better
```
    bool annotations() const { return annotations_type != AnnotationsType::None; };
```

But it is not used in code base anymore, the best would be delete it and update unit tests.

include/server/api/route_parameters_grammar.hpp
```diff
@@ -42,6 +42,13 @@ struct RouteParametersGrammar : public BaseParametersGrammar<Iterator, Signature
 
     RouteParametersGrammar(qi::rule<Iterator, Signature> &root_rule_) : BaseGrammar(root_rule_)
     {
+        const auto add_annotation = [](engine::api::RouteParameters &route_parameters,
```
oxidase: this lambda is not needed with `operator |=` and in-place phoenix bind (i added related comments)

include/engine/api/route_parameters.hpp
```diff
@@ -94,20 +104,54 @@ struct RouteParameters : public BaseParameters
...
+    template <typename... Args>
+    RouteParameters(const bool steps_,
+                    const bool alternatives_,
+                    const AnnotationsType annotations_,
+                    const GeometriesType geometries_,
+                    const OverviewType overview_,
+                    const boost::optional<bool> continue_straight_,
+                    Args... args_)
+        : BaseParameters{std::forward<Args>(args_)...}, steps{steps_}, alternatives{alternatives_},
+          annotations_type{annotations_}, annotations{true}, geometries{geometries_},
```
oxidase: here `annotations` is `true` even if `annotations_` is `None`

include/server/api/route_parameters_grammar.hpp
oxidase: ... here in-place functor
```
annotations_type[ph::bind(&engine::api::RouteParameters::annotations_type, qi::_r1) |= qi::_1] % ',');
```

include/engine/api/route_parameters.hpp
```diff
@@ -94,20 +104,54 @@ struct RouteParameters : public BaseParameters
...
+}
+
+inline RouteParameters::AnnotationsType operator|(RouteParameters::AnnotationsType lhs,
+                                                  RouteParameters::AnnotationsType rhs)
+{
+    return (RouteParameters::AnnotationsType)(
+        static_cast<std::underlying_type_t<RouteParameters::AnnotationsType>>(lhs) |
+        static_cast<std::underlying_type_t<RouteParameters::AnnotationsType>>(rhs));
+}
+
```
oxidase: ... and here compound operator
```
inline RouteParameters::AnnotationsType& operator|=(RouteParameters::AnnotationsType& lhs,
                                                    RouteParameters::AnnotationsType rhs)
{
    return lhs = lhs | rhs;
}
```

include/engine/api/route_parameters.hpp
daniel-j-h: We can't do that - the annotations member attribute is in our public API - we can't change it to a function :(

include/server/api/route_parameters_grammar.hpp
oxidase: due to https://github.com/Project-OSRM/osrm-backend/pull/3628#discussion_r99120126 both members must be updated, so `add_annotation` must remain

include/server/api/route_parameters_grammar.hpp
```diff
@@ -42,6 +42,13 @@ struct RouteParametersGrammar : public BaseParametersGrammar<Iterator, Signature
...
+                                       engine::api::RouteParameters::AnnotationsType &route_param) {
+            route_parameters.annotations_type = route_parameters.annotations_type | route_param;
+            route_parameters.annotations = route_parameters.annotations_type != engine::api::RouteParameters::AnnotationsType::None;
+            return route_parameters;
```
oxidase: returned value is not used and `return` can be removed

include/engine/api/route_parameters.hpp
karenzshea: https://github.com/Project-OSRM/osrm-backend/issues/3644

TheMarex:
This is still missing the `speed` parameter. Capturing this in https://github.com/Project-OSRM/osrm-backend/issues/3664 to be fixed in another PR.

Otherwise ready to merge.

PR #3630, opened 2017-01-30
Simpler cucumber error output

danpat:
We don't need to log a stack trace when we encounter a table diff (unexpected test results) during a `cucumberjs` run.  This PR logs the table differences in the normal way, but removes the stack trace.

Test-run errors change from this:
screen shot 2017-01-30 at 3 11 10 pm

to this:
screen shot 2017-01-30 at 3 11 36 pm

which is much easier on the eye.  The stack trace wasn't helpful for the actual errors anyway, it was a trace inside the cucumberjs framework.

## Tasklist
 - [x] review
 - [x] adjust for comments

### DIFF
--- a/features/lib/table_diff.js
+++ b/features/lib/table_diff.js
@@ -50,5 +50,5 @@ module.exports = function (expected, actual) {
         s.push(rowString);
     });
 
-    return s.join('\n') + '\nTODO this is a temp workaround waiting for https://github.com/cucumber/cucumber-js/issues/534';
+    return s.join('\n');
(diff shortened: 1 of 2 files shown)


emiltin:
if there's an error in the lua script, will it be shown?

TheMarex:
@emiltin errors in the lua script should surface in the `test/log/{feature_subdir}/{name}.feature/{line}_{scenario_name}.log` and make `osrm-extract` exit with a non-zero error code.

Issue #3629, opened 2017-01-30
osrm-extract triggers on demo server


MoKob:
seeing

`[assert][140281425536896] /mnt/ebs/osrm/2017-01-30-173002/src/extractor/extractor_callbacks.cpp:149` (triggering [here](https://github.com/Project-OSRM/osrm-backend/blob/master/src/extractor/extractor_callbacks.cpp#L149))

on the demo server. This seems to be most likely to be connected to the merge of https://github.com/Project-OSRM/osrm-backend/pull/2399 on https://github.com/Project-OSRM/osrm-backend/issues/77 since it introduces the assertion [here](https://github.com/Project-OSRM/osrm-backend/commit/279f8aabfb67bda62b51cc32542ea18a812e4823#diff-8713a422b8d487cb0e1670bdfcc7fde0R149).

/cc @oxidase @TheMarex

oxidase:
assertion occurs at http://www.openstreetmap.org/node/181295699 The current car profile allows to use the street http://map.project-osrm.org/?z=18&center=53.077581%2C8.808975&loc=53.077405%2C8.808852&loc=53.077689%2C8.809131&hl=en&alt=0

PR #3632, opened 2017-01-31
Split if in ProcessWay and adjsuted profiles to comform to old behavior.

TheMarex:
# Issue

Addresses #3629 by making if backward compatible.

## Tasklist
 - [x] review
 - [x] adjust for comments

## Requirements / Relations

### DIFF
--- a/features/bicycle/access.feature
+++ b/features/bicycle/access.feature
@@ -182,3 +182,9 @@ Feature: Bike - Access tags on ways
             | bridleway |            | yes  |         | x     |
             | bridleway | designated |      |         |       |
             | bridleway |            |      |         |       |
+
+    Scenario: Bike - Tram with oneway when access is implicit
+        Then routability should be
+            | highway     | railway | oneway | bothw |
+            | residential | tram    | yes    | x     |
+
--- a/features/car/access.feature
+++ b/features/car/access.feature
@@ -245,3 +245,14 @@ Feature: Car - Restricted access
             | gate         | designated | x     |
             | gate         | private    |       |
             | gate         | garbagetag | x     |
+
+    Scenario: Car - a way with conditional access
+        Then routability should be
+            | highway    | vehicle:forward | vehicle:backward:conditional | forw | backw |
+            | pedestrian | yes             | delivery @ (20:00-11:00)     | x    |       |
+
+    Scenario: Car - a way with a list of tags
+        Then routability should be
+            | highway | motor_vehicle            | motor_vehicle:forward | motor_vehicle:backward | forw | backw |
+            | footway |                          |                       | destination            |      | x     |
+            | track   | destination;agricultural | destination           |                        | x    | x     |
--- a/features/foot/access.feature
+++ b/features/foot/access.feature
@@ -103,3 +103,8 @@ Feature: Foot - Access tags on ways
             | bridleway |            | yes     |      |       |
             | bridleway | designated |         |      |       |
             | bridleway |            |         |      |       |
+
+    Scenario: Foot - a way with missing :forward tag
+        Then routability should be
+            | highway  | bicycle:backward | foot:backward | forw | backw |
+            | cycleway | designated       | designated    |      | x     |
--- a/profiles/bicycle.lua
+++ b/profiles/bicycle.lua
@@ -259,10 +259,10 @@ function way_function (way, result)
     result.forward_speed = platform_speeds[public_transport]
     result.backward_speed = platform_speeds[public_transport]
   elseif use_public_transport and railway and railway_speeds[railway] then
-      result.forward_mode = mode.train
-      result.backward_mode = mode.train
-     -- railways
-    if access and access_tag_whitelist[access] then
+    result.forward_mode = mode.train
+    result.backward_mode = mode.train
+    -- railways
+    if (not access and highway) or access_tag_whitelist[access] then
       result.forward_speed = railway_speeds[railway]
       result.backward_speed = railway_speeds[railway]
     end
--- a/profiles/foot.lua
+++ b/profiles/foot.lua
@@ -21,7 +21,7 @@ local profile = {
   oneway_handling         = 'specific',     -- respect 'oneway:foot' but not 'oneway'
   traffic_light_penalty   = 2,
   u_turn_penalty          = 2,
-  
+
   barrier_whitelist = Set {
     'cycle_barrier',
     'bollard',
@@ -34,19 +34,19 @@ local profile = {
     'no',
     'block'
   },
-  
+
   access_tag_whitelist = Set {
-    'yes', 
-    'foot', 
-    'permissive', 
-    'designated'  
+    'yes',
+    'foot',
+    'permissive',
+    'designated'
   },
-  
+
   access_tag_blacklist = Set {
-    'no', 
-    'private', 
-    'agricultural', 
-    'forestry', 
+    'no',
+    'private',
+    'agricultural',
+    'forestry',
     'delivery'
   },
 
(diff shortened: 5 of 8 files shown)


oxidase:
another assertion on http://www.openstreetmap.org/way/4621857

oxidase:
Updated lua profiles to set  {forward,backward}_access correctly for the following ways:
* http://www.openstreetmap.org/way/4282676 has `vehicle:backward:conditional` that is not handled in `Directional.get_values_by_set`  that  returns `nil` value. Added a check to return "yes" in similar cases.

* http://www.openstreetmap.org/way/4621857 has `motor_vehicle:backward` but has no `motor_vehicle` or `motor_vehicle:forward` tags. Similar case to the first one 

* http://www.openstreetmap.org/way/4898368 has `motor_vehicle` = `destination;agricultural` and 
`motor_vehicle:forward` = `destination`. `data.backward_access` is `destination;agricultural` and is not in the black list or the white list. Changed to use the complement of the black list as a  white list to set `speed_profile["default"]`

* http://www.openstreetmap.org/way/24455356 has `railway=tram` but no access tag. bicycle profile sets driving mode to `mode.train` but does not set `railway_speeds[railway]`. Changed to use `mode.inaccessible` in similar cases.

The adjusted behavior is similar to api_version=0, but it should be checked if such behavior is correct.

emiltin:
mind looking at (and perhaps merge) #3615 before merging this? if factors out common code from the car and foot profile to a common helper file. in case you like the approach, it would be easier to fix it first and change the {forward,backward}_access profile behaviour in the helper file.

oxidase:
@emiltin yes, #3615 must be merged before this PR.

 I have checked car, bicycle and foot profiles on the latest planet file, no further assertions found.
/cc @TheMarex

features/bicycle/access.feature
```diff
@@ -182,3 +182,9 @@ Feature: Bike - Access tags on ways
             | bridleway |            | yes  |         | x     |
             | bridleway | designated |      |         |       |
             | bridleway |            |      |         |       |
+
+    Scenario: Bike - Tram with oneway when access is implicit
+        Then routability should be
+            | highway     | railway | oneway | forw | backw |
+            | residential | tram    | yes    |      | x     |
```
TheMarex: Shouldn't `oneway=yes` only make `forw` accessible not `backw`?

profiles/lib/directional.lua
```diff
@@ -53,6 +53,13 @@ function Directional.get_values_by_set(way,data,keys)
     end
   end
 
+  -- allow access in undefined in one direction and defined in opposite one
```
TheMarex: I think this implies usage in the access restriction context. This function make no assumptions about tag values however. This would need to be handled in the caller code.

features/bicycle/access.feature
oxidase: Here is pretty hidden bug in c++ code that was solved by #2399 

* lua bicycle profile returns `result.forward_mode = mode.train (5), result.forward_speed = -1, result.backward_mode = mode.pushing_bike (6), result.backward_speed = 6`

* [the check](https://github.com/Project-OSRM/osrm-backend/blob/445225bf6dc719ef4f91fe6ab7e6fa2f4941fef7/src/extractor/extractor_callbacks.cpp#L121-L122) fails and `forward_weight_data` is default constructed with `forward_weight_data = {0.0, WeightType::INVALID}`

* [the next check](https://github.com/Project-OSRM/osrm-backend/blob/445225bf6dc719ef4f91fe6ab7e6fa2f4941fef7/src/extractor/extractor_callbacks.cpp#L135-L136) pass, because `forward_weight_data = {6, WeightType::SPEED}`

* [here](https://github.com/Project-OSRM/osrm-backend/blob/445225bf6dc719ef4f91fe6ab7e6fa2f4941fef7/src/extractor/extractor_callbacks.cpp#L291-L297) `split_edge = false` because `parsed_way.forward_speed = -1`

* [here](https://github.com/Project-OSRM/osrm-backend/blob/445225bf6dc719ef4f91fe6ab7e6fa2f4941fef7/src/extractor/extractor_callbacks.cpp#L312) `is_opposite_way = false` because `parsed_way.forward_travel_mode` is `TRAVEL_MODE_TRAIN`

* [here](https://github.com/Project-OSRM/osrm-backend/blob/445225bf6dc719ef4f91fe6ab7e6fa2f4941fef7/src/extractor/extractor_callbacks.cpp#L346-L347) `forward_only` is false because neither `split_edge` nor `TRAVEL_MODE_INACCESSIBLE == parsed_way.backward_travel_mode` are true

* [adding edges](https://github.com/Project-OSRM/osrm-backend/blob/445225bf6dc719ef4f91fe6ab7e6fa2f4941fef7/src/extractor/extractor_callbacks.cpp#L353-L365)
```
osm_source_id = 1, osm_target_id= 2, weight_data = {0.0, WeightType::INVALID}, forward = 1, backward= 1, is_split = 0, travel_mode = 5
osm_source_id = 2, osm_target_id= 3, weight_data = {0.0, WeightType::INVALID}, forward = 1, backward= 1, is_split = 0, travel_mode = 5
osm_source_id = 3, osm_target_id= 4, weight_data = {0.0, WeightType::INVALID}, forward = 1, backward= 1, is_split = 0, travel_mode = 5
osm_source_id = 4, osm_target_id= 5, weight_data = {0.0, WeightType::INVALID}, forward = 1, backward= 1, is_split = 0, travel_mode = 5
```

* now because of [**bug here**](https://github.com/Project-OSRM/osrm-backend/blob/445225bf6dc719ef4f91fe6ab7e6fa2f4941fef7/src/extractor/extraction_containers.cpp#L447-L448) the return value for `data={0.0, WeightType::INVALID}` is -1

* now -1 weight is corrected [here](https://github.com/Project-OSRM/osrm-backend/blob/445225bf6dc719ef4f91fe6ab7e6fa2f4941fef7/src/extractor/extraction_containers.cpp#L454)
to 1

As a result 2 compressed edges with weight 4 and mode `TRAVEL_MODE_TRAIN` that is not the same as desired by profile and allows to drive in both directions.

profiles/lib/directional.lua
oxidase: Ok, it can be moved out to handle `nil` values as `no`

oxidase:
@emiltin i have changed the fallback value 'yes' to 'no'. This requires only test changes in https://github.com/Project-OSRM/osrm-backend/blob/1ea88c21b747a8a80c1d786c100dafd2ef0cb88a/features/car/access.feature#L53-L60 where `access` and `motorcar` are not specified.

features/car/access.feature
```diff
@@ -50,14 +50,14 @@ Feature: Car - Restricted access
             | no     |                | yes             |          |                  |                   |      | x     |
             | no     |                | no              | yes      |                  |                   | x    | x     |
             | no     |                | no              | no       |                  | yes               |      | x     |
-            |        | no             |                 |          | no               |                   |      | x     |
-            |        |                | no              |          |                  | no                | x    |       |
+            |        | no             |                 |          | no               |                   |      |       |
```
emiltin: i don't think changes to this test are correct. if you have no access tags, access is implied by a valid way type. for this line in the test we have:

highway=primary (this is set by "routability should be" tests is no highway tag is specified)
access:forward=no
motorcar:forward=no

this means backward access is implied by the way type, while forward access is denied by motorcar:forward (and also by access:forward).

same goes for the other changes to this test.

oxidase: thanks for clarification! i changed behavior in this case: if access in one direction is specified but in other direction is not and can not be derived from the way type then set mode to `inaccessible`. This keeps old behavior for the test.

oxidase:
@emiltin i have updated the branch. i am currently running extraction for the planet file with `foot.lua` and found some missing tables in the `profile` table. I committed empty tables in 3abbdac.

emiltin:
thanks. i fixed that aready in #46ac9d, but seems it disappeared during the refactoring of common code.

TheMarex:
This looks good to me. 👍 Merging.

PR #3635, opened 2017-01-31
Don't use bool return values on successful write, instead throw

duizendnegen:
# Issue

As mentioned in #3496 by @daniel-j-h - both catching errors and checking a bool for success values is confusing, the `FileWriter` now only throws.

### DIFF
--- a/include/util/io.hpp
+++ b/include/util/io.hpp
@@ -71,15 +71,8 @@ bool serializeVectorIntoAdjacencyArray(const std::string &filename,
     for (auto const &vec : data)
         all_data.insert(all_data.end(), vec.begin(), vec.end());
 
-    if (!file.SerializeVector(offsets))
-    {
-        return false;
-    }
-
-    if (!file.SerializeVector(all_data))
-    {
-        return false;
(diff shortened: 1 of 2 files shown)


include/util/io.hpp
```diff
@@ -71,15 +71,8 @@ bool serializeVectorIntoAdjacencyArray(const std::string &filename,
...
-    {
-        return false;
-    }
-
-    if (!file.SerializeVector(all_data))
-    {
-        return false;
-    }
+    file.SerializeVector(offsets);
+    file.SerializeVector(all_data);
```
daniel-j-h: Nice! See below, the `util/io.hpp` seems to have the same issues - sometimes returning `bool` but at the same time low-level I/O functions might throw.

duizendnegen:
Seems like Travis is on its feet again - is there anything to do with regards to the codecov results? (we merely reduced complexity, but it had a side effect on coverage)

TheMarex:
@duizendnegen I think this is fine. The diff wasn't probably covered to begin with.

PR #3636, opened 2017-02-01
Add logic to penalize certain roads in the car profile

TheMarex:
Continuation of https://github.com/Project-OSRM/osrm-backend/pull/3276 against the right branch.

# Issue

In the past we have used penalties on the speed to make routing over certain types of roads less appealing to the routing engine. However this screws with the ETAs when you actually need to go over those penalized roads.

This PR changes this speed penalties to weight penalties which was unlocked by fixing #77. 

This means certain tests that test for applying the penalties in the car profile will fail now. They need to be adapted to test if the speed is unchanged while the rate/weight is adjusted for the penalty.

## Tasklist
-  [x] port hov handler into penalty handler to heavily penalize HOV tagged ways
-  [x] use min applicable penalty rather than a multiplied aggregate penalty value
-  [x] Add support for new cucumber test headers: `rate` and `weight`
-  [x] Update tests to test for more important values, e.g. weight rather than speed
 - [x] review
 - [x] adjust for comments

### DIFF
--- a/features/car/service.feature
+++ b/features/car/service.feature
@@ -4,11 +4,11 @@ Feature: Car - Surfaces
     Background:
         Given the profile "car"
 
-    Scenario: Car - Surface should reduce speed
+    Scenario: Car - Ways tagged service should reduce speed
         Then routability should be
-            | highway  | service           | forw       | backw       |
-            | service  | alley             | 5 km/h +-1 | 5 km/h +-1  |
-            | service  | emergency_access  |            |             |
-            | service  | driveway          | 5 km/h +-1 | 5 km/h +-1  |
-            | service  | drive-through     | 5 km/h +-1 | 5 km/h +-1  |
-            | service  | parking           | 5 km/h +-1 | 5 km/h +-1  |
+            | highway  | service           | forw        | backw        | forw_rate  |
+            | service  | alley             | 15 km/h +-1 | 15 km/h +-1  | 12.05 km/h |
+            | service  | emergency_access  |             |              |            |
+            | service  | driveway          | 15 km/h +-1 | 15 km/h +-1  | 12.05 km/h |
+            | service  | drive-through     | 15 km/h +-1 | 15 km/h +-1  | 12.05 km/h |
+            | service  | parking           | 15 km/h +-1 | 15 km/h +-1  | 12.05 km/h |
--- a/features/car/service.feature
+++ b/features/car/service.feature
@@ -7,8 +7,8 @@ Feature: Car - Surfaces
     Scenario: Car - Ways tagged service should reduce speed
         Then routability should be
             | highway  | service           | forw        | backw        | forw_rate  |
-            | service  | alley             | 15 km/h +-1 | 15 km/h +-1  | 12.05 km/h |
+            | service  | alley             | 15 km/h +-1 | 15 km/h +-1  | 2          |
             | service  | emergency_access  |             |              |            |
-            | service  | driveway          | 15 km/h +-1 | 15 km/h +-1  | 12.05 km/h |
-            | service  | drive-through     | 15 km/h +-1 | 15 km/h +-1  | 12.05 km/h |
-            | service  | parking           | 15 km/h +-1 | 15 km/h +-1  | 12.05 km/h |
+            | service  | driveway          | 15 km/h +-1 | 15 km/h +-1  | 2          |
+            | service  | drive-through     | 15 km/h +-1 | 15 km/h +-1  | 2          |
+            | service  | parking           | 15 km/h +-1 | 15 km/h +-1  | 2          |
--- a/features/step_definitions/routability.js
+++ b/features/step_definitions/routability.js
@@ -6,9 +6,10 @@ module.exports = function () {
     this.Then(/^routability should be$/, (table, callback) => {
         this.buildWaysFromTable(table, () => {
             var directions = ['forw','backw','bothw'],
+                testedHeaders = ['forw','backw','bothw','forw_rate','backw_rate','bothw_rate'],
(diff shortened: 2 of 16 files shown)


profiles/lib/handlers.lua
```diff
@@ -266,45 +256,58 @@ function Handlers.handle_surface(way,result,data,profile)
...
-      penalized_speed = profile.service_speeds[service]
-    elseif width <= 3 or (lanes <= 1 and is_bidirectional) then
-      penalized_speed = result.backward_speed / 2
+  local sideroad_penalty = 1.0
+  data.sideroad = way:get_value_by_key("side_road")
+  if "yes" == data.sideroad or "rotary" == data.sideroad then
+    sideroad_penalty = side_road_multipler;
+  end
+
+  local penalty = service_penalty * width_penalty * alternating_penalty * sideroad_penalty
```
emiltin: i haven't worked with weights yet, but for speed reductions my experience is that multiplying penalites doesn't give good results. it's better to use mininimum. the problem is at a modest reduction for all parameters results in a a very low speed, effectually causing the way to be unroutable.

karenzshea: :+1: Updated to use the minimum of all set penalties

unit_tests/library/tile.cpp
```diff
@@ -32,7 +32,7 @@ BOOST_AUTO_TEST_CASE(test_tile)
     const auto rc = osrm.Tile(params, result);
     BOOST_CHECK(rc == Status::Ok);
 
-    BOOST_CHECK_EQUAL(result.size(), 114033);
+    BOOST_CHECK_EQUAL(result.size(), 113824);
```
daniel-j-h: Why does this change?

features/step_definitions/routability.js
```diff
@@ -31,6 +59,8 @@ module.exports = function () {
                                 usingShortcut = row[direction];
                             }
 
+                            // TODO split out accessible/not accessible value from forw/backw headers
+                            // rename forw/backw to forw/backw_speed
```
daniel-j-h: Can you ticket this

features/step_definitions/routability.js
```diff
@@ -6,19 +6,47 @@ module.exports = function () {
...
+                            switch (true) {
+                            case '' === want:
+                                outputRow[rate] = result[direction].status ?
+                                    result[direction].status.toString() : '';
+                                break;
+                            case /^\d+$/.test(want):
+                                if (result[direction].rate) {
+                                    outputRow[rate] = !isNaN(result[direction].rate) ?
+                                        result[direction].rate.toString() :
+                                        result[direction].rate.toString() || '';
```
daniel-j-h: Looks strange, is line 39 required?

profiles/lib/handlers.lua
```diff
@@ -278,45 +265,69 @@ function Handlers.handle_surface(way,result,data,profile)
...
-      penalized_speed = profile.service_speeds[service]
-    elseif width <= 3 or (lanes <= 1 and is_bidirectional) then
-      penalized_speed = result.backward_speed / 2
+  local sideroad_penalty = 1.0
+  data.sideroad = way:get_value_by_key("side_road")
+  if "yes" == data.sideroad or "rotary" == data.sideroad then
+    sideroad_penalty = side_road_multiplier;
+  end
+
+  local penalty = math.min(service_penalty, width_penalty, alternating_penalty, sideroad_penalty, hov_penalty)
```
daniel-j-h: Not sure if this is semantically correct - e.g. a narrow alternating oneway should have a higher penalty than a narrow street, right?

TheMarex: One of Emil's concerns was that this usually leads to roads becoming unroutable instead of penalized.

emiltin: it should be based on considering the nature of the inputs.
regarding speed: if a way is narrow and has a bad surface, it's probably not going to be slower than if it's just narrow. on the other hand, alternating directions means you're going to spend time standing still, so it might make sense to multiple that factor, e.g speed = speed*min(width,surface)*alternating

features/car/service.feature
```diff
@@ -4,11 +4,11 @@ Feature: Car - Surfaces
...
         Then routability should be
-            | highway  | service           | forw       | backw       |
-            | service  | alley             | 5 km/h +-1 | 5 km/h +-1  |
-            | service  | emergency_access  |            |             |
-            | service  | driveway          | 5 km/h +-1 | 5 km/h +-1  |
-            | service  | drive-through     | 5 km/h +-1 | 5 km/h +-1  |
-            | service  | parking           | 5 km/h +-1 | 5 km/h +-1  |
+            | highway  | service           | forw        | backw        | forw_rate  |
+            | service  | alley             | 15 km/h +-1 | 15 km/h +-1  | 2          |
+            | service  | emergency_access  |             |              |            |
```
danpat: Now that we have the ability to penalize access-limited ways, should `emergency_access` just become like all the others (i.e. allow routing if a pin is dropped) ?

On second thought, I guess not - cut-off roads on freeways that are emergency-only should never be routed over, and because they're short, even heavily penalizing them is not sufficient to prevent access.  We would need to implement turn-based penalties for that to work.

So, nevermind, but I've written all this stuff now, so I'm going to post it anyway for perpetuity.

unit_tests/library/tile.cpp
TheMarex: I think it's because the duration values encoded in the vector tiles changed. It uses a delta encoding so the size is heavily data dependent.

features/step_definitions/routability.js
TheMarex: https://github.com/Project-OSRM/osrm-backend/issues/3686

Issue #3638, opened 2017-02-01
weight_name set to distance results in osrm::util::exception


nocallaghan:
Issue extracting pbf when using the new weight_name set to distance. 

```
osrm-extract british-isles-latest.osm.pbf 
[info] Using script profiles/car.lua
[info] Input file: british-isles-latest.osm.pbf
[info] Profile: car.lua
[info] Threads: 32
[info] Parsing in progress..
[STXXL-MSG] STXXL v1.4.1 (prerelease/Debug)
[STXXL-MSG] Disk '/tmp/stxxl' is allocated, space: 47683 MiB, I/O implementation: syscall queue=0 devid=0
[info] input file generated by osmium/1.5.0
[info] timestamp: n/a
[info] Found 3 turn restriction tags:
[info]   motorcar
[info]   motor_vehicle
[info]   vehicle
[info] Parsing finished after 49.0357 seconds
[info] Raw input contains 103929949 nodes, 12702569 ways, and 222996 relations
terminate called after throwing an instance of 'osrm::util::exception'
  what():  There are no edges remaining after parsing. (at src/extractor/extractor.cpp:231)
Aborted (core dumped)
```

Server: Ubuntu 16.04.1 LTS (GNU/Linux 4.4.0-53-generic x86_64)
Lua 5.2.4  Copyright (C) 1994-2015 Lua.org, PUC-Rio
OSRM is latest from master
The only change made to car.lua was to set weight_name to distance.  

If I reset car.lua to the default option (duration), the extraction completes no problem.

oxidase:
@nocallaghan the assertion is correct because if `weight_name` is not "duration" then weights for edges must be specified. This is not implemented in the current master, so all edges will have 0 weight and rejected. 

Edge weights can be set in `way_function` via `forward_rate` and `backward_rate` properties. The properties have meaning similar to speed, but have units of meters per a weight unit instead of meters per second. The edge weight is computed as the edge distance divided by the rate, so using rate = 1 will result to weight = edge distance. 

PR #3640 adds "distance" weight to `way_function`.

nocallaghan:
Thank you @oxidase I'll try latest master out, I see #3640 was merged. Thanks for your other comment re: stxxl

nocallaghan:
@oxidase that PR worked great, closing this. Thanks again

PR #3640, opened 2017-02-01
Added 'distance' weight to car profile

oxidase:
# Issue

For #3638 added usage of `forward_rate` and `backward_rate` in `car.lua` profile. Also fixes weight assembly :bug: in `assembleLeg`

## Tasklist
 - [x] review
 - [x] adjust for comments

### DIFF
--- a/include/engine/guidance/assemble_leg.hpp
+++ b/include/engine/guidance/assemble_leg.hpp
@@ -131,6 +131,8 @@ inline RouteLeg assembleLeg(const datafacade::BaseDataFacade &facade,
 {
     const auto target_duration =
         (target_traversed_in_reverse ? target_node.reverse_duration : target_node.forward_duration);
+    const auto target_weight =
+        (target_traversed_in_reverse ? target_node.reverse_weight : target_node.forward_weight);
 
     auto distance = std::accumulate(
         leg_geometry.segment_distances.begin(), leg_geometry.segment_distances.end(), 0.);
@@ -165,6 +167,7 @@ inline RouteLeg assembleLeg(const datafacade::BaseDataFacade &facade,
     // On local segments, the target duration is already part of the duration, however.
 
     duration = duration + target_duration;
+    weight = weight + target_weight;
     if (route_data.empty())
     {
         duration -= (target_traversed_in_reverse ? source_node.reverse_duration
@@ -204,7 +207,7 @@ inline RouteLeg assembleLeg(const datafacade::BaseDataFacade &facade,
 
     return RouteLeg{std::round(distance * 10.) / 10.,
                     duration / 10.,
-                    duration / facade.GetWeightMultiplier(),
+                    weight / facade.GetWeightMultiplier(),
                     summary,
                     {}};
 }
--- a/profiles/lib/handlers.lua
+++ b/profiles/lib/handlers.lua
@@ -66,7 +66,7 @@ end
 -- determine if this way can be used as a start/end point for routing
 function Handlers.handle_startpoint(way,result,data,profile)
   -- only allow this road as start point if it not a ferry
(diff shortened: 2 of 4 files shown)


emiltin:
can we do this in the appropriate handlers, so that we don't duplicate code between profiles?

oxidase:
@emiltin moved to `Handlers.handle_weights`

emiltin:
nice. should weights only be set in forward/backward directions if the way is  routable in the corresponding directions?

profiles/lib/handlers.lua
```diff
@@ -399,15 +399,22 @@ function Handlers.handle_oneway(way,result,data,profile)
   end
 end
 
+function Handlers.handle_weights(way,result,data,profile)
+  if properties.weight_name == 'distance' then
+    -- set weight rates to 1 for the distance weight, edge weights are distance / rate
+    result.forward_rate = 1
```
TheMarex: As @emiltin pointed out this needs a check for either `{foward,backward}_speed > 0` or `{forward,backward}_mode != mode.inaccessible`.

include/engine/guidance/assemble_leg.hpp
```diff
@@ -204,7 +207,7 @@ inline RouteLeg assembleLeg(const datafacade::BaseDataFacade &facade,
 
     return RouteLeg{std::round(distance * 10.) / 10.,
                     duration / 10.,
-                    duration / facade.GetWeightMultiplier(),
+                    weight / facade.GetWeightMultiplier(),
```
TheMarex: Urg, good catch.

profiles/lib/handlers.lua
oxidase: Good point!  But if `{forward,backward}_mode != mode.inaccessible` then assertion `parsed_way.duration > 0 || parsed_way.forward_speed > 0` in `ProcessWay` will fail if speed is 0,
so the condition "`{forward,backward}_mode != mode.inaccessible` and `{foward,backward}_speed > 0`" is asserted to be true and if `{forward,backward}_mode == mode.inaccessible`  the weight or rate is not used to compute the edge weight.

What i missed is an additional assertion in `ProcessWay` if `parsed_way.forward_travel_mode != TRAVEL_MODE_INACCESSIBLE` and `!fallback_to_duration` then must be asserted `parsed_way.weight > 0 || parsed_way.forward_rate > 0`. At the moment such edges will have default initialized weights that is 0 for `ValueByEdge` and `ValueByMeter`.

oxidase:
@emiltin for `{forward,backward}_mode != mode.inaccessible` 0 speed values will result in an assertion [here](https://github.com/Project-OSRM/osrm-backend/blob/8f28d317cf14aa80a3fe3b732d3b85743e75803b/src/extractor/extractor_callbacks.cpp#L133) or [here](https://github.com/Project-OSRM/osrm-backend/blob/8f28d317cf14aa80a3fe3b732d3b85743e75803b/src/extractor/extractor_callbacks.cpp#L150)
Also using 0 or negative rate to mark inaccessible edges if mode is not mode.inaccessible will be incorrect, i  added assertions to prevent such situations.

profiles/lib/handlers.lua
emiltin: but why are forward and backward rates still being set, without checking whether the way is accessible (mode != 0 and speed > 0)? if the way is not accessible, i don't think the rate should be set.

oxidase: agree, i added checks and also set `result.weight` explicitly to 0, that means "use rate values to compute edge weights"

TheMarex:
Travis had a hiccup here, restarted the build. Otherwise this looks good to merge. 👍

PR #3645, opened 2017-02-02
Implements Mapping for NodeBasedGraph -> EdgeBasedGraph Translation

daniel-j-h:
For https://github.com/Project-OSRM/osrm-backend/pull/3603. 

Implements a mapping to translate from the node based graph to the edge based graph. This allows us to use our node based graph partition for the edge based graph.

Based on the outline in: https://github.com/Project-OSRM/osrm-backend/issues/3205#issuecomment-275169844 - @TheMarex care to have a look?

### DIFF
--- a/include/extractor/edge_based_graph_factory.hpp
+++ b/include/extractor/edge_based_graph_factory.hpp
@@ -7,18 +7,16 @@
 #include "extractor/edge_based_edge.hpp"
 #include "extractor/edge_based_node.hpp"
 #include "extractor/extraction_turn.hpp"
+#include "extractor/guidance/turn_analysis.hpp"
+#include "extractor/guidance/turn_instruction.hpp"
+#include "extractor/guidance/turn_lane_types.hpp"
 #include "extractor/original_edge_data.hpp"
 #include "extractor/profile_properties.hpp"
 #include "extractor/query_node.hpp"
 #include "extractor/restriction_map.hpp"
-
-#include "extractor/guidance/turn_analysis.hpp"
-#include "extractor/guidance/turn_instruction.hpp"
-#include "extractor/guidance/turn_lane_types.hpp"
+#include "util/deallocating_vector.hpp"
 #include "util/guidance/bearing_class.hpp"
 #include "util/guidance/entry_class.hpp"
-
-#include "util/deallocating_vector.hpp"
 #include "util/name_table.hpp"
 #include "util/node_based_graph.hpp"
 #include "util/typedefs.hpp"
@@ -35,6 +33,7 @@
 #include <vector>
 
 #include <boost/filesystem/fstream.hpp>
+#include <boost/optional.hpp>
 
 namespace osrm
 {
@@ -86,7 +85,9 @@ struct TurnIndexBlock
 #pragma pack(pop)
 static_assert(std::is_trivial<TurnIndexBlock>::value, "TurnIndexBlock is not trivial");
 static_assert(sizeof(TurnIndexBlock) == 24, "TurnIndexBlock is not packed correctly");
-}
+} // ns lookup
+
+struct NodeBasedGraphToEdgeBasedGraphMappingWriter; // fwd. decl
 
 class EdgeBasedGraphFactory
 {
@@ -113,7 +114,9 @@ class EdgeBasedGraphFactory
              const std::string &turn_weight_penalties_filename,
              const std::string &turn_duration_penalties_filename,
              const std::string &turn_penalties_index_filename,
-             const bool generate_edge_lookup);
+             const bool generate_edge_lookup,
+             const bool generate_nbg_ebg_mapping,
+             const std::string &nbg_ebg_mapping_path);
 
     // The following get access functions destroy the content in the factory
     void GetEdgeBasedEdges(util::DeallocatingVector<EdgeBasedEdge> &edges);
@@ -172,7 +175,10 @@ class EdgeBasedGraphFactory
     guidance::LaneDescriptionMap &lane_description_map;
 
     unsigned RenumberEdges();
-    void GenerateEdgeExpandedNodes();
+
+    void GenerateEdgeExpandedNodes(const bool generate_nbg_ebg_mapping,
+                                   const std::string &nbg_ebg_mapping_path);
+
     void GenerateEdgeExpandedEdges(ScriptingEnvironment &scripting_environment,
                                    const std::string &original_edge_data_filename,
                                    const std::string &turn_lane_data_filename,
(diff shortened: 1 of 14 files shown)


src/partition/partitioner.cpp
```diff
@@ -57,13 +60,74 @@ struct CompressedNodeBasedGraph
...
+        auto tails_it = tails.find(edge_based_node);
+        if (tails_it != end(tails))
+            return tails_it->second;
+
+        BOOST_ASSERT(false);
+        return NodeBasedNodes{SPECIAL_NODEID, SPECIAL_NODEID};
+    }
+
+  private:
+    std::unordered_map<EdgeID, NodeBasedNodes> heads;
```
TheMarex: The edge_ids are continuous if you merge `heads` and `tails` into one array. If the information about head/tail is needed, saving that in a continuous `std::vector<bool>` should be more efficient.

src/partition/partitioner.cpp
```diff
@@ -57,13 +60,74 @@ struct CompressedNodeBasedGraph
...
+        for (std::uint64_t i{0}; i < num_mappings; ++i)
+        {
+
+            const auto u = reader.ReadOne<NodeID>();    // node based graph `from` nodes
+            const auto v = reader.ReadOne<NodeID>();    // node based graph `to` nodes
+            const auto head = reader.ReadOne<EdgeID>(); // edge based graph forward nodes
+            const auto tail = reader.ReadOne<EdgeID>(); // edge based graph backward nodes
+
+            heads.insert({head, {u, v}});
+            tails.insert({tail, {u, v}});
```
TheMarex: For tail it should be `v, u`.

include/extractor/edge_based_graph_factory.hpp
```diff
@@ -182,7 +188,15 @@ class EdgeBasedGraphFactory
                                    const std::string &turn_penalties_index_filename,
                                    const bool generate_edge_lookup);
 
-    void InsertEdgeBasedNode(const NodeID u, const NodeID v);
+    // Mapping betweenn the node based graph u,v nodes and the edge based graph head,tail edge ids.
+    // Required in the osrm-partition tool to translate from a nbg partition to a ebg partition.
+    struct Mapping
+    {
+        NodeID u, v;
+        EdgeID head, tail;
```
TheMarex: `head` and `tail` can be a little bit confusing here because they typically reference the start/target for an `arc` (edge based edge). However there is no arc from `tail` to `head` and both can be heads and tails for different arcs (there are typically arcs both from and to `head` and `tail`).

`{forward,backward}_edge_based_node` might be more precise but maybe you have another idea for a name.

daniel-j-h:
@TheMarex care to explain what's going on [in the contractor](https://github.com/Project-OSRM/osrm-backend/blob/c7640903c36a3ea7dbe062e8d98e5e1284507501/src/contractor/graph_contractor.cpp#L13-L84) in detail. First try is [here](https://github.com/Project-OSRM/osrm-backend/pull/3645/commits/6cee83c86efbb28ff43edf9b43c3ccb65d8d245f) and is not yet handling all the edge cases the contractor does since I don't understand all of them.

src/extractor/edge_based_graph_factory.cpp
```diff
@@ -286,15 +299,20 @@ void EdgeBasedGraphFactory::GenerateEdgeExpandedNodes()
 
                 BOOST_ASSERT(node_u < node_v);
 
+                boost::optional<Mapping> mapping;
+
                 // if we found a non-forward edge reverse and try again
                 if (edge_data.edge_id == SPECIAL_NODEID)
                 {
-                    InsertEdgeBasedNode(node_v, node_u);
+                    mapping = InsertEdgeBasedNode(node_v, node_u);
```
MoKob: hm. This `mapping` is kind of hanging in the air for me. @daniel-j-h do you see a way we could do this in a less `hacky` approach? Interleaving it with the EdgeBasedGraph Generation seems a bit messy.

src/extractor/edge_based_graph_factory.cpp
```diff
@@ -172,6 +173,8 @@ void EdgeBasedGraphFactory::InsertEdgeBasedNode(const NodeID node_u, const NodeI
     }
 
     BOOST_ASSERT(current_edge_source_coordinate_id == node_v);
+
+    return Mapping{node_u, node_v, forward_data.edge_id, reverse_data.edge_id};
```
MoKob: see below. I feel this could become kind of messy. Its a mapping tailored to MLD but integrated into our general data structures.

include/partition/edge_based_graph_reader.hpp
```diff
@@ -0,0 +1,202 @@
...
+#include <iterator>
+#include <memory>
+#include <vector>
+
+namespace osrm
+{
+namespace partition
+{
+
+struct EdgeBasedGraphEdgeData
```
MoKob: is this a duplication of the edge based graph edge? Without looking at the rest of the code, I am a bit surprised to see this here. Given the bit restrictions, I strongly would prefer having a single fixed location for this.

daniel-j-h: Well I started out only exposing _some_ member attributes. Then had to add some for the edge based graph preparation step. Then added the boundary arc flag (which we're currently not using). And finally added all extractor edge members in addition, since we have to serialize out the full edge based graph after modification anyway, right?

src/extractor/edge_based_graph_factory.cpp
daniel-j-h: I know and the whole implementation of it just felt wrong as well. I see no other way to hook into the edge based graph factory, though. If you have better ideas, I'm open for changes here.

zhudelong:
It seems that the mapping does not consider the new feature, namely multi-via-way restriction. This leads to a large partition cell and huge memory consumption. This bug should be fixed.

mjjbell:
@zhudelong are you able to provide details of your setup so that I can try and reproduce the issue?

zhudelong:
@mjjbell 
https://github.com/Project-OSRM/osrm-backend/blob/f7478ba80f04ee6777556661f93edff5bef85c90/src/partitioner/partitioner.cpp#L103

Can you check the EBG partition result after this mapping process? You can sort the EBG cells according to the cell size, and check whether there exist large cells and whether the largest cell owns a partition id of SPECIAL_NODEID.

If there exists a cell that holds SPECIAL_NODEID, it means some EBG nodes are not mapped during the above process.
According to my test, this issue is caused by the generation of the cnbg_to_ebg file, in which the duplicated nodes are not considered.

mjjbell:
Thanks for the explanation @zhudelong, I see the problem now. I will create a new ticket to track this issue.

PR #3646, opened 2017-02-03
Reduce memory in stxxl

oxidase:
# Issue

Issue #77 introduced a regression in memory consumption in [`InternalExtractorEdge`](https://github.com/Project-OSRM/osrm-backend/blob/9d2628b74f9ba7742955cd178c42a1bd09a55523/include/extractor/internal_extractor_edge.hpp#L23-L41). This PR aimed to reduce memory consumption by using float type for the value and avoid the variant type. The float type can be used with assumption that initial edge weights and duration values from profiles have  24 significant bits (about 7 decimal digits). This is pretty safe assumption, because these values are used as 31 significant bits integers and summing up without precision loss 128 maximum possible values will lead to signed integer overflow. To route for over 128 edges without risk of the integer overflow  edge weights should have less than 24 significant bits.
The variant type introduces 8 bytes overhead by `std::size_t type_index` and results to sizeof(mapbox::util::variant<osrm::extractor::detail::ValueByEdge, osrm::extractor::detail::ValueByMeter>)
= 16 bytes per `WeightData` or 32 bytes per edge. Because there are only two types `ValueByEdge` that is always non-negative and  `ValueByMeter` that is always strictly positive the sign bit can be used to encode the value type in a single value without values domains overlap.

Structure sizes:
* before merging #2399 was 
```
sizeof(osrm::extractor::InternalExtractorEdge::WeightData) = 16
sizeof(osrm::extractor::NodeBasedEdgeWithOSM) = 40
sizeof(osrm::extractor::InternalExtractorEdge) = 64
ptype osrm::extractor::InternalExtractorEdge
type = struct osrm::extractor::InternalExtractorEdge {
    osrm::extractor::NodeBasedEdgeWithOSM result;
    osrm::extractor::InternalExtractorEdge::WeightData weight_data;
    osrm::util::Coordinate source_coordinate;
```
* after merging #2399
```
sizeof(osrm::extractor::InternalExtractorEdge::WeightData) = 16
sizeof(osrm::extractor::NodeBasedEdgeWithOSM) = 48
sizeof(osrm::extractor::InternalExtractorEdge) = 88
ptype osrm::extractor::InternalExtractorEdge
type = struct osrm::extractor::InternalExtractorEdge {
    osrm::extractor::NodeBasedEdgeWithOSM result;
    WeightData weight_data;
    WeightData duration_data;
    osrm::util::Coordinate source_coordinate;
```
* with the PR
```
sizeof(osrm::extractor::InternalExtractorEdge::WeightData) = 4
sizeof(osrm::extractor::NodeBasedEdgeWithOSM) = 44
sizeof(osrm::extractor::InternalExtractorEdge) = 60
ptype osrm::extractor::InternalExtractorEdge
type = struct osrm::extractor::InternalExtractorEdge {
    osrm::extractor::NodeBasedEdgeWithOSM result;
    WeightData weight_data;
    WeightData duration_data;
    osrm::util::Coordinate source_coordinate;
```

So PR saves 28 bytes per edge wrt to the current master (>11GB) and 4 bytes wrt master before #2399 merge (1.6GB) for raw 392418286 input ways.

## Tasklist
 - [x] review
 - [x] adjust for comments

## Requirements / Relations
The PR is rebased on top of #3629, so only e6c884d commit is relevant for review.

### DIFF
--- a/include/extractor/internal_extractor_edge.hpp
+++ b/include/extractor/internal_extractor_edge.hpp
@@ -20,23 +20,37 @@ namespace extractor
 
 namespace detail
 {
-// these are used for duration based mode of transportations like ferries
-OSRM_STRONG_TYPEDEF(double, ValueByEdge)
-OSRM_STRONG_TYPEDEF(double, ValueByMeter)
-using ByEdgeOrByMeterValue = mapbox::util::variant<detail::ValueByEdge, detail::ValueByMeter>;
-
-struct ToValueByEdge
+// Use a single float to pack two positive values:
+//  * by_edge: + sign, value >= 0, value used as the edge weight
+//  * by_meter: - sign, value > 0, value used as a denominator in distance / value
+struct ByEdgeOrByMeterValue
 {
-    ToValueByEdge(double distance_) : distance(distance_) {}
-
-    ValueByEdge operator()(const ValueByMeter by_meter) const
+    struct ValueByEdge
     {
-        return ValueByEdge{distance / static_cast<double>(by_meter)};
+    } static const by_edge;
+    struct ValueByMeter
+    {
+    } static const by_meter;
+
+    ByEdgeOrByMeterValue() : value(0.f) {}
+
+    ByEdgeOrByMeterValue(ValueByEdge, double input)
+    {
+        BOOST_ASSERT(input >= 0.f);
+        value = static_cast<value_type>(input);
     }
 
-    ValueByEdge operator()(const ValueByEdge by_edge) const { return by_edge; }
+    ByEdgeOrByMeterValue(ValueByMeter, double input)
+    {
+        BOOST_ASSERT(input > 0.f);
+        value = -static_cast<value_type>(input);
+    }
 
-    double distance;
+    double operator()(double distance) { return value >= 0 ? value : -distance / value; }
+
+  private:
+    using value_type = float;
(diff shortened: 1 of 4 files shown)


include/extractor/internal_extractor_edge.hpp
```diff
@@ -20,23 +20,37 @@ namespace extractor
...
-// these are used for duration based mode of transportations like ferries
-OSRM_STRONG_TYPEDEF(double, ValueByEdge)
-OSRM_STRONG_TYPEDEF(double, ValueByMeter)
-using ByEdgeOrByMeterValue = mapbox::util::variant<detail::ValueByEdge, detail::ValueByMeter>;
-
-struct ToValueByEdge
+// Use a single float to pack two positive values:
+//  * by_edge: + sign, value >= 0, value used as the edge weight
+//  * by_meter: - sign, value > 0, value used as a denominator in distance / value
+struct ByEdgeOrByMeterValue
```
TheMarex: Elegant solution. Precision should be more than enough.

oxidase:
rebased. with the change i see only two lines `[STXXL-ERRMSG] External memory block allocation error: 2145386496 bytes requested, 1163919360 bytes free. Trying to extend the external memory space...`
on the planet size and extraction + contraction time is 14 hours

Issue #3633, opened 2017-01-31
Investigate alternatives to central shared memory control block


TheMarex:
Currently we have a central control block that contains the current shared memory region. We use a conditional variable to signal all `libosrm` instances that the data was updated. This has a major limitation on OSX which expects the _address_ of the mutex that guards the conditional variable to be the same over all client (even in different processes where the shared memory might be mapped to completely different regions!).

#3627 fixes this issue by introducing a fixed address mapping for the control block which also contains the mutex. But this has the downside that our DataWatchdog is now a singleton since we can only ever have one per process, otherwise the shared memory mapping conflicts for each instance of DataWatchdog (since it is fixed to be at the same address on OSX).

This should be addressed down the line by not relying on a central control block at all anymore. For this we should investigate alternatives such as message based IPC mechanism to propagate updates and discovery of shared memory segments over files.

Future design constraints should include the possibility of having multiple (named) datasets in shared memory.

/cc @oxidase @danpat

oxidase:
Tried solutions:
- spinning checks: 
  - :-1: introduces delays in data updates
  - :+1: simplicity
* two semaphores with counting waiting threads
  - :-1: non-robust against crashed or killed processes
  - :+1: fixed memory usage; data update delays depend only on the context switch
* semaphores queue
  - :-1: limited number of waiters due to fixed queue size
  - :+1: data update delays depend only on the context switch; one semaphore is used by only one waiting process, if the processes is  killed the associated semaphore will be signaled and later can be reused.

Closed via #3648 with a queue of semaphores for non-Linux systems the hard limit is 128 Engine instances in OS. It is allowed to use shared memory control block in multiple multi-threaded processes.

Issue #1221, opened 2014-10-14
Make sharedmemory named so you can run multiple osrm-routed instances


emiltin:
loading data into separate named stores with osrm-datastore would support:
- running multiple profiles on the same production machine (car, bike, foot..)
- isolating tests, to make sure if doesn't affect already running processes

TheMarex:
Not sure about the complexity involved here. I get the point that is is bad to assume you only have on `osrm-routed` instance running.

What Dennis is getting at is a possible OSRMv2 where you can store multiple profiles in memory (different algorithm). But this is quite a while way..

emiltin:
the use case is providing different routing profiles on the same server, like car and bike, and making the required server setup simpler.

woodbri:
+1 on this feature

bvbever:
For the moment, I have 6 different routing profiles on the same server. 
When I do an update, I restart the corresponding routing what results in a downtime of several minutes.
So I like this feature ;-)

arcadas:
+1

pfannkuchengesicht:
Hey, so I was wondering if there are any estimations as to when this feature might be available? 
I'm currently working on a project that integrates OSRM into PostgreSQL and as of now it's a bit limiting.
Anyhow, any work-around available for now? One way would probably be to load the files directly, but I'm quite sure that would really slow things down. Any suggestion would be really appreciated!

danpat:
@pfannkuchengesicht Nobody is working on it at the moment AFAIK.  Patches are always welcome though!

TheMarex:
This was implemented in #4982.

PR #3648, opened 2017-02-03
Use shared monitor with two interprocess_semaphore's

oxidase:
# Issue

One possible solution for #3633 is to use interprocess_semaphore's to model a conditional variable.

PR works for linux, windows, and OSX, but hits https://svn.boost.org/trac/boost/ticket/12617 boost bug on OS X 10.11. it can be fixed by the patch.
<details>
diff --git a/include/boost/interprocess/detail/os_thread_functions.hpp b/include/boost/interprocess/detail/os_thread_functions.hpp
index 291f426..bd50fd9 100644

```diff
--- a/include/boost/interprocess/detail/os_thread_functions.hpp
+++ b/include/boost/interprocess/detail/os_thread_functions.hpp
@@ -53,7 +53,11 @@
 #     include <sys/sysctl.h>
 #  endif
 //According to the article "C/C++ tip: How to measure elapsed real time for benchmarking"
-#  if defined(CLOCK_MONOTONIC_PRECISE)   //BSD
+
+#  if (defined(macintosh) || defined(__APPLE__) || defined(__APPLE_CC__))
+#     include <mach/mach_time.h>  // mach_absolute_time, mach_timebase_info_data_t
+#     define BOOST_INTERPROCESS_MATCH_ABSOLUTE_TIME
+#  elif defined(CLOCK_MONOTONIC_PRECISE) //BSD
 #     define BOOST_INTERPROCESS_CLOCK_MONOTONIC CLOCK_MONOTONIC_PRECISE
 #  elif defined(CLOCK_MONOTONIC_RAW)     //Linux
 #     define BOOST_INTERPROCESS_CLOCK_MONOTONIC CLOCK_MONOTONIC_RAW
@@ -61,9 +65,9 @@
 #     define BOOST_INTERPROCESS_CLOCK_MONOTONIC CLOCK_HIGHRES
 #  elif defined(CLOCK_MONOTONIC)         //POSIX (AIX, BSD, Linux, Solaris)
 #     define BOOST_INTERPROCESS_CLOCK_MONOTONIC CLOCK_MONOTONIC
-#  elif !defined(CLOCK_MONOTONIC) && (defined(macintosh) || defined(__APPLE__) || defined(__APPLE_CC__))
-#     include <mach/mach_time.h>  // mach_absolute_time, mach_timebase_info_data_t
-#     define BOOST_INTERPROCESS_MATCH_ABSOLUTE_TIME
+// #  elif !defined(CLOCK_MONOTONIC) && (defined(macintosh) || defined(__APPLE__) || defined(__APPLE_CC__))
+// #     include <mach/mach_time.h>  // mach_absolute_time, mach_timebase_info_data_t
+// #     define BOOST_INTERPROCESS_MATCH_ABSOLUTE_TIME
 #  else
 #     error "No high resolution steady clock in your system, please provide a patch"
 #  endif
```
</details>

/cc @danpat, @TheMarex 

## Tasklist
 - [x] review
 - [x] adjust for comments

Related:
#1221, #3558


oxidase:
at the moment the solution can lead to a deadlock, because if the waiter is killed the `waiters` number  will not be decremented

oxidase:
After some attempts to implement a conditional variable via semaphores and waiters counter that can handled killed waiters, just get to similar as https://github.com/boostorg/interprocess/blob/develop/include/boost/interprocess/sync/spin/condition.hpp 

My suggest is not to use NIH implementation and fallback  to the spinlock-based conditional variable via `BOOST_INTERPROCESS_FORCE_GENERIC_EMULATION` on OSX.

oxidase:
checked boost interprocess conditional variables on windows and osx:  *non-robust against application kills*. will fallback on windows and osx to a solution with one semaphore that generates spurious wake ups for every killed `osrm-routed` process.

oxidase:
`node-osrm` tests also pass on OS X 10.11.6

include/storage/shared_monitor.hpp
```diff
@@ -0,0 +1,175 @@
...
+    mutex_type &get_mutex() const { return internal().mutex; }
+
+#if USE_BOOST_INTERPROCESS_CONDITION
+    template <typename Lock> void wait(Lock &lock) { internal().condition.wait(lock); }
+
+    void notify_all() { internal().condition.notify_all(); }
+#else
+    template <typename Lock> void wait(Lock &lock)
+    {
+        if (((internal().head + 1) & (buffer_size - 1)) == (internal().tail & (buffer_size - 1)))
```
TheMarex: What is the advantage of using `& (buffer_size - 1)` instead of `% buffer_size`? I'm more familiar with the latter when it comes to wrap-a-round indexing.

Wrapping the ring buffer in smaller helper functions to hide the index "magic" might improve readability.

include/storage/shared_monitor.hpp
```diff
@@ -0,0 +1,175 @@
...
+        semaphore->~interprocess_semaphore();
+    }
+
+    void notify_all()
+    {
+        bi::scoped_lock<mutex_type> lock(internal().mutex);
+        while (internal().tail != internal().head)
+        {
+            auto index = (internal().tail++) & (buffer_size - 1);
+            auto semaphore = reinterpret_cast<bi::interprocess_semaphore *>(
```
TheMarex: Same here, getting the semaphore could be hidden behind a nicer interface.

include/storage/shared_monitor.hpp
```diff
@@ -0,0 +1,175 @@
...
+    static constexpr int buffer_size = 256;
+    static constexpr int internal_size = 4 * 4096;
+
+    struct InternalData
+    {
+        InternalData() : head(0), tail(0){};
+
+        std::size_t head, tail;
+        mutex_type mutex;
+        char buffer[buffer_size * sizeof(bi::interprocess_semaphore)];
```
TheMarex: This could warrant a little bit of an explanation of what we are going with these semaphore: Implementing a conditional variable using a queue of semaphores.

oxidase:
@TheMarex i have moved queue functionality into the internal data structure. It is still has to be checked on OSX

TheMarex:
@oxidase Looks good to me. 👍 Seems to work on travis, we should make @danpat let it run on his computer though. 😉

oxidase:
Checked `node-osrm` with the master branch -> tests with multiple Engines pass! Please don't forget to run `osrm-datastore --spring-clean` after building master :smiley_cat:

PR #3652, opened 2017-02-04
Isochrone Plugin v2

danpat:
This PR adds an `isochrone` plugin, which can retrieve travel-time [isochrones](https://en.wikipedia.org/wiki/Isochrone_map) from the OSRM routing graph.

This is an alternative implementation to https://github.com/Project-OSRM/osrm-backend/pull/2477 that does not need to load the node-based-graph into memory - instead, it derives a temporary edge-based-graph from the CH.  It's not super fast, but it works :-)

It functions as follows:

  1. Calculates a maximum bounding box for the isochrone using the `range` x MAX_SPEED_METERS_PER_SECOND (200km/h for this implementation)
  2. Fetches everything in that box from the RTree
  3. Identifies all the distinct edge-based-nodes returned from the RTree
  4. Uses the start/end node-based-node IDs to identify potentially directly connected edge-based-nodes
  5. Builds a temporary edge-based-graph by testing the potentially directly connected edge-based-nodes in the CH
  6. Performs a Dijkstra search on the temporary edge-based-graph, terminating when all possible paths up to `range` have been explored
  7.  Returns the result

Usage is:

```
/isochrone/v1/driving/{lon},{lat}?range={seconds}
```

A 300 second isochrone on the `car` profile takes about 350ms to return.

Currently, the plugin returns a GeoJSON `FeatureCollection` for every edge-based-node geometry within the `range`.

## TODO:

 - [ ] add a `type` parameter, and support:
    - [ ] `exact` - return a custom JSON structure for the full isochrone tree
    - [ ] `points` - return a big GeoJSON `FeatureCollection` of `Point` features with distance/duration for each point, for use with third-party topography tools, like [gdal_contour](http://www.gdal.org/gdal_contour.html)
    - [ ] `polygon` - return the boundary polygon for the isochrone (TODO: what type? convex? concave? other params?)
 - [ ] add flag to limit isochrone max range (big isochrones get really expensive)
 - [ ] review
 - [ ] adjust for comments

### DIFF
--- a/src/engine/plugins/isochrone.cpp
+++ b/src/engine/plugins/isochrone.cpp
@@ -1,0 +1,334 @@
+#include "engine/plugins/isochrone.hpp"
+#include "engine/edge_unpacker.hpp"
+#include "engine/plugins/plugin_base.hpp"
+#include "util/timing_util.hpp"
+#include "util/coordinate_calculation.hpp"
+
+#include <queue>
+#include <iomanip>
+#include <cmath>
+
+namespace osrm
+{
+namespace engine
+{
+namespace plugins
+{
+
+Status IsochronePlugin::HandleRequest(const std::shared_ptr<datafacade::BaseDataFacade> facade,
+                                      const api::IsochroneParameters &parameters,
+                                      std::string &pbf_buffer) const
+{
+
+    const auto range = parameters.range;
+
+    util::Coordinate startcoord{util::FloatLongitude{parameters.lon},
+                                util::FloatLatitude{parameters.lat}};
+
+    const auto MAX_SPEED_METERS_PER_SECOND = 200 / 3.6;
+    const auto MAX_TRAVEL_DISTANCE_METERS = MAX_SPEED_METERS_PER_SECOND * parameters.range;
+
+    const auto latitude_range = MAX_TRAVEL_DISTANCE_METERS /
+                                (util::coordinate_calculation::detail::EARTH_RADIUS * M_PI * 2) *
+                                360;
+
+    const auto longitude_range =
+        360 * MAX_TRAVEL_DISTANCE_METERS /
+        ((util::coordinate_calculation::detail::EARTH_RADIUS * M_PI * 2) *
+         cos(parameters.lat * util::coordinate_calculation::detail::DEGREE_TO_RAD));
+
+    util::Coordinate southwest{
+        util::FloatLongitude{static_cast<double>(parameters.lon - longitude_range)},
+        util::FloatLatitude{static_cast<double>(parameters.lat - latitude_range)}};
+    util::Coordinate northeast{
+        util::FloatLongitude{static_cast<double>(parameters.lon + longitude_range)},
+        util::FloatLatitude{static_cast<double>(parameters.lat + latitude_range)}};
+
+    util::SimpleLogger().Write(logINFO) << "sw " << southwest;
+    util::SimpleLogger().Write(logINFO) << "ne " << northeast;
+
+    TIMER_START(GET_EDGES_TIMER);
+    const auto edges = facade->GetEdgesInBox(southwest, northeast);
+    TIMER_STOP(GET_EDGES_TIMER);
+    util::SimpleLogger().Write(logINFO) << "Fetch RTree " << TIMER_MSEC(GET_EDGES_TIMER);
+
+    auto startpoints = facade->NearestPhantomNodes(startcoord, 1);
+    // Struct to hold info on all the EdgeBasedNodes that are visible in our tile
+    // When we create these, we insure that (source, target) and packed_geometry_id
+    // are all pointed in the same direction.
+    struct EdgeBasedNodeInfo
+    {
+        bool is_geometry_forward; // Is the geometry forward or reverse?
+        unsigned packed_geometry_id;
+    };
+    // Lookup table for edge-based-nodes
+    std::unordered_map<NodeID, EdgeBasedNodeInfo> edge_based_node_info;
+
+    struct SegmentData
+    {
+        NodeID target_node;
+        EdgeID edge_based_node_id;
+    };
+
+    std::unordered_map<NodeID, std::vector<SegmentData>> node_based_graph;
+    // Reserve enough space for unique edge-based-nodes on every edge.
+    // Only a tile with all unique edges will use this much, but
+    // it saves us a bunch of re-allocations during iteration.
+    node_based_graph.reserve(edges.size() * 2);
+
+    TIMER_START(CONSTRUCT_LOOKUP);
+    // Build an adjacency list for all the road segments visible in
+    // the tile
+    for (const auto &edge : edges)
+    {
+        if (edge.forward_segment_id.enabled)
+        {
+            // operator[] will construct an empty vector at [edge.u] if there is no value.
+            node_based_graph[edge.u].push_back({edge.v, edge.forward_segment_id.id});
+            if (edge_based_node_info.count(edge.forward_segment_id.id) == 0)
+            {
+                edge_based_node_info[edge.forward_segment_id.id] = {true, edge.packed_geometry_id};
+            }
+            else
+            {
+                BOOST_ASSERT(edge_based_node_info[edge.forward_segment_id.id].is_geometry_forward ==
+                             true);
+                BOOST_ASSERT(edge_based_node_info[edge.forward_segment_id.id].packed_geometry_id ==
+                             edge.packed_geometry_id);
+            }
+        }
+        if (edge.reverse_segment_id.enabled)
+        {
+            node_based_graph[edge.v].push_back({edge.u, edge.reverse_segment_id.id});
+            if (edge_based_node_info.count(edge.reverse_segment_id.id) == 0)
+            {
+                edge_based_node_info[edge.reverse_segment_id.id] = {false, edge.packed_geometry_id};
+            }
+            else
+            {
+                BOOST_ASSERT(edge_based_node_info[edge.reverse_segment_id.id].is_geometry_forward ==
+                             false);
+                BOOST_ASSERT(edge_based_node_info[edge.reverse_segment_id.id].packed_geometry_id ==
+                             edge.packed_geometry_id);
+            }
+        }
+    }
+    TIMER_STOP(CONSTRUCT_LOOKUP);
+    util::SimpleLogger().Write(logINFO) << "Create lookup table " << TIMER_MSEC(CONSTRUCT_LOOKUP);
+
+    // Given a turn:
+    //     u---v
+    //         |
+    //         w
+    //  uv is the "approach"
+    //  vw is the "exit"
+    std::vector<contractor::QueryEdge::EdgeData> unpacked_shortcut;
+    std::vector<EdgeWeight> approach_weight_vector;
+
+    // Make sure we traverse the startnodes in a consistent order
+    // to ensure identical PBF encoding on all platforms.
+    std::vector<NodeID> sorted_startnodes;
+    sorted_startnodes.reserve(node_based_graph.size());
+    for (const auto &startnode : node_based_graph)
+        sorted_startnodes.push_back(startnode.first);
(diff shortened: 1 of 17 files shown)


src/engine/plugins/isochrone.cpp
```diff
@@ -0,0 +1,330 @@
...
+                }
+            }
+        }
+    }
+
+    // Phase 2 - scan nodes in descending CH order, updating d(v) where we can
+    TIMER_START(TABLE_TIMER);
+    std::sort(phantoms.begin(), phantoms.end(), [](const auto &a, const auto &b) {
+        const auto &a_id =
+            a.forward_segment_id.enabled ? a.forward_segment_id.id : a.reverse_segment_id.id;
```
TheMarex: Sadly the node ID is not a proxy for contraction level in our CH implementation: We keep the node ID constant starting in the `EdgeBasedGraphFactory`, because we need to save the ID in the RTree which happens before contraction.

The way you could fix this would be to load the `.osrm.level` file which contains exactly that (we use it for speeding up the contraction on successive contractions).

danpat: Welp, that probably explains the missing edges in the test isochrones I've been making.

I'll see if I can get correct results by loading the `.level` file, then we can discuss how we might optionally load it.

danpat: Interestingly, the PHAST paper here: https://www.microsoft.com/en-us/research/wp-content/uploads/2010/09/phastTR.pdf

states:

> The second subphase scans all vertices in G↓ in descending rank order.*
> * - For correctness, any reverse topological order will do.

Dirk23:
is it posible to set a max speed like this [isochrone server](https://github.com/Mapotempo/osrm-isochrone-server) is able to?

src/engine/plugins/isochrone.cpp
Fkawala: @danpat I'm curious about what you have in mind for the sorting, would you care to share some details ?

mloskot:
Here in `ml/5.10-ch_isochrone_poc` branch is my first attempted to rebase-merge this PR on top of `5.10` branch (current release): https://github.com/mloskot/osrm-backend/tree/ml/5.10-ch_isochrone_poc
which contains single commit resolving all the conflicts (mloskot/osrm-backend@f3d2b19f8ed5a32e3be32146c7da0fa015ceadd9).

This allows to build Isochrone plugin with GCC 7 and VS2017.
I haven't managed to try it out which success though (see below).

**TODO:** I'd appreciate if someone reviewed the merge and checked if it didn't loose any crucial parts.

------

I briefely tried to run the isochrone plugin and I'm hitting wall, perhaps unrelated to the plugin itself.
The plugin uses the `QueryHeap` and in `osrm::util::QueryHeap` class, definitions like `HeapHandle{};` yield object and things become crooked later, in `boost/heap/detail/mutable_heap.hpp`,  comparisons like `iterator == rhs.iterator` faili, even with access violations.

Basically, this `std::list` iterator is/becomes singular, not properly initialised.
But, the promise is that `std::list<T>::iterator it;` will construct non-singular object with null node pointer.

To be investigated...

adomas-mazeikis:
@mloskot Have you had any success fixing the issue with QueryHeap?

mloskot:
@adomas-mazeikis Not yet, but the issue seems MSVC-specific annoyance and iterator debug levels interfering with testing execution of the plugin. I have asked for advice on boost-users: [[heap] Singular handle issue with MSVC iterator debug facilities](https://lists.boost.org/boost-users/2017/08/87787.php).

------

Another quick weird observations I'm experiencing trying out the plugin is around the "Phase 2" step:

* accessing these `a_id` and `b_id` references in `a_id < b_id` give me read access violation
```
    std::sort(phantoms.begin(), phantoms.end(), [](const auto &a, const auto &b) {
        const auto &a_id =
            a.forward_segment_id.enabled ? a.forward_segment_id.id : a.reverse_segment_id.id;
        const auto &b_id =
            b.forward_segment_id.enabled ? b.forward_segment_id.id : b.reverse_segment_id.id;

        return a_id < b_id;
    });
```
* switching to plain values works around the problem
```
  std::sort(phantoms.begin(), phantoms.end(), [](const auto &a, const auto &b) {
         auto a_id = a.forward_segment_id.enabled ? a.forward_segment_id.id : a.reverse_segment_id.id;
         auto b_id = b.forward_segment_id.enabled ? b.forward_segment_id.id : b.reverse_segment_id.id;

        return a_id < b_id;
    });
```

It is uncleared why that is happening really.
The `id` is pretty much an `unsigned int`, and I there is no concurrency issue that could have invalidated references in the lambda, since `std::sort` is sequential. Unless I'm missing something :)

------

If I run the plugin using optimised build configuration (ie. iterator debug facilities are disabled, so the heap issue is gone) with the `std::sort` changed as above, I get `"type":"FeatureCollection"` result with collection of `Point` geometries.  However, I haven't checked correctness of the result. 

(Tomorrow, I'll gone offline for 1-3 weeks, I will continue as soon as I'm back.)

src/engine/plugins/isochrone.cpp
```diff
@@ -0,0 +1,330 @@
...
+IsochronePlugin::IsochronePlugin() {}
+
+Status
+IsochronePlugin::HandleRequest(const std::shared_ptr<const datafacade::BaseDataFacade> facade,
+                               const api::IsochroneParameters &parameters,
+                               std::string &pbf_buffer) const
+{
+
+    // const auto range = parameters.range;
+    const auto range = 300;
```
mloskot: Is `range` intentionally `signed int`?

Compare with [unsigned IsochroneParameters::range](https://github.com/Project-OSRM/osrm-backend/pull/3652/files#diff-284c45c35e4acbf777ab511606d198e2R53) declaration.

The type difference actually does affect the isochrones result (see my next comment).

src/engine/plugins/isochrone.cpp
```diff
@@ -0,0 +1,330 @@
...
+            if (data.forward)
+            {
+                const NodeID to = facade->GetTarget(edge);
+                const EdgeWeight edge_weight = data.weight;
+
+                BOOST_ASSERT_MSG(edge_weight > 0, "edge_weight invalid");
+                const EdgeWeight to_weight = weight + edge_weight;
+
+                // No need to search further than our maximum radius
+                if (weight > range * 10)
```
mloskot: TL;TR: Is it intentional here to mismatch signed/unsigned types and rely on `signed int` to `unsigned int` implicit conversion (eg. to filter nodes with negative weights)?

In my previous https://github.com/Project-OSRM/osrm-backend/pull/3652#discussion_r132565108 I asked about signed/unsigned mismatch of `range`.
This `weight > range * 10` test is where the difference matters:
* `weight` is `signed int`
* if `range` is `unsigned int` (eg. pulled as `const auto range = parameters.range;` earlier)
then result of comparison test is different than, I assume, expected.

For example, if `weight` value is negative, it is implicitly converted to `unsigned int`, according to the usual arithmetic conversions pattern (C++/5/9), and overflow flips the result loosing some nodes.

danpat: Never ascribe to cleverness that which can be explained by oversight.

I probably just didn't notice :-)

Note that this plugin has a serious problem with the reverse graph exploration currently - see the comments by @TheMarex above.

mloskot: @danpat for a moment, I assumed the obscure aims for actual utilisation of the C++ dark fu to benefit :-)  One of those "failed by `auto`" trips.
I will correct in my 5.10/master-based branches of the plugin.

src/engine/plugins/isochrone.cpp
```diff
@@ -0,0 +1,330 @@
...
+        }
+    }
+
+    // Phase 2 - scan nodes in descending CH order, updating d(v) where we can
+    TIMER_START(TABLE_TIMER);
+    std::sort(phantoms.begin(), phantoms.end(), [](const auto &a, const auto &b) {
+        const auto &a_id =
+            a.forward_segment_id.enabled ? a.forward_segment_id.id : a.reverse_segment_id.id;
+        const auto &b_id =
+            b.forward_segment_id.enabled ? b.forward_segment_id.id : b.reverse_segment_id.id;
```
mloskot: This initialisation of (const) reference from bit field is not allowed by MSVC.

It's a grey area. While C++ std does not allow to initialise pointer and non-const reference from a bit field, it does allow it for const references, but via binding a copy-constructed temporary. Technically, there is no gain in defining `auto&` vs plain `auto` value here.

mloskot:
@adomas-mazeikis `QueryHeap` w/ MSVC issue solved in #4396

CraigglesO:
Are there any updates on this project? I played with it a little. Some weird issues if anyone else uses this PR:
* Setup the map inside the osrm-backend folder, otherwise you get weird errors when using osrm-routed.
* for some reason on linux, this PR cannot find stxxl, just direct install it `sudo apt-get install libstxxl-dev`

Thanks for your work though, it's very interesting.

mloskot:
@ CraigglesO 
> Are there any updates on this project? 

I'm not aware of any. I stopped evaluating this PR in Aug 2017

This PR is far behind the current master (which, n.b., moves foward with light speed) and it would be nice to rebase it, so it's easier to try it out again. I may give it a go some time during next weeks.

> *  Setup the map inside the osrm-backend folder

What do you mean exactly? What map? Setup how?

> * for some reason on linux, this PR cannot find stxxl, just direct install it sudo apt-get install libstxxl-dev

Current master no longer requires STXXL, it's an optional dependency now. Again, rebase would be nice.

geography:
Hi all, any news about this plugin?

mloskot:
On 2 February 2018 at 16:56, Emanuele Notarnicola <notifications@github.com>
wrote:

> Hi all, any news about this plugin?
>

How about reading the very recent comments from 2-3 days ago?

CraigglesO:
LMAO. hey man. Thanks for the response. Sorry for not replying sooner. Got sidetracked and github is pretty gnarly about not showing old notifications.

for setup inside osrm-backend I mean that if you run `osrm-extract` with a profile setup outside of the osrm-backend folder (right after git clone and cmake....) then you run into issues. However, if you use `osrm-extract` on a pbf file INSIDE the osrm-backend folder no errors come up. It's just a weird bug or I am just doing something wrong.

Good to hear about stxxl though.

Thanks for your work its inspiring. I wrote an implementation that uses a 1080ti GTX and it takes about 1s for 15min car, 7s for 30min car and about 22-30s for 1hr car. However, if your not in a city, 1hr car takes less then a second BAHAHAA. It's kind of funny. I was having a lot of trouble wrapping my head around all of OSRM so I wrote a basic r-tree for all the nodes and edges and keep all of them on RAM over storage for now. LOTS of improvement to go.

tgrushka:
I would really, really like to see this functionality implemented. It seems multiple individuals have tried to create this plugin, and it was never accepted. Right now, I'm generating isochrones using a grid and durations table. This is not ideal, either in terms of efficiency or accuracy. With the grid, there is always a trade-off between accuracy and calculation time. Taking 8 minutes to generate isochrones for a 1000x1000 grid is just too long, when using this R-Tree method would seem to be much more efficient and effective. Yet, where is it after all this time? I've searched and searched this repository and documentation. Is this being implemented or planned for this project? How can the efforts of those who have spent a lot of time on this already be rewarded?

mloskot:
@tgrushka Try Valhalla, https://github.com/valhalla/valhalla-docs/blob/master/isochrone/api-reference.md

danpat:
Sadly, this was a side project for me, and I have not had time to revisit it.

If you really need isochrones, I second @mloskot suggestion of using Valhalla.

Another option, which uses the same approach as Valhalla, is https://github.com/mapbox/mapbox-isochrone by @peterqliu - it uses the table/matrix plugin to sample, and calculates a concave polygon around the points (Valhalla does this internally too).

This PR was about generating a precise isochrone.  Maybe I'll get back to it one day....

DennisOSRM:

PR #3654, opened 2017-02-05
Refactor/bicycle profile

emiltin:
# Issue

First step in refactoring the bike profile (#3413), similar to what was done with the car and foot profile.

Settings have been moved to a 'profile' table, and code handling straightforward stuff have been replaced by calls to the standard handlers.

The bike profile is significantly more complex than the foot and car profile, because it must handle pushing of bikes. This means it must also handle pedestrian routing. As a next step, perhaps we can find a way to delegate that part to the foot profile.

I modified the hande_startpoint handler because the bike profile allows starting on ways with two modes: cycling and pushing_bike. The handler checks the allowed start mode from the 'allowed_start_modes' set in the profile table, or if not found, compares to the default mode for the profile. 

## Tasklist
 - [x] review
 - [x] adjust for comments

### DIFF
--- a/profiles/bicycle.lua
+++ b/profiles/bicycle.lua
@@ -1,96 +1,11 @@
 api_version = 1
 
 -- Bicycle profile
-
 local find_access_tag = require("lib/access").find_access_tag
+local Set = require('lib/set')
+local Sequence = require('lib/sequence')
 local limit = require("lib/maxspeed").limit
 local set_classification = require("lib/guidance").set_classification
-
--- Begin of globals
-barrier_whitelist = { [""] = true, ["cycle_barrier"] = true, ["bollard"] = true, ["entrance"] = true, ["cattle_grid"] = true, ["border_control"] = true, ["toll_booth"] = true, ["sally_port"] = true, ["gate"] = true, ["no"] = true, ["block"] = true }
-access_tag_whitelist = { ["yes"] = true, ["permissive"] = true, ["designated"] = true }
-access_tag_blacklist = { ["no"] = true, ["private"] = true, ["agricultural"] = true, ["forestry"] = true, ["delivery"] = true }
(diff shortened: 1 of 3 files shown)


oxidase:
@emiltin there are missing tables in `profile`, please cherry-pick in branch b0686b2

emiltin:
ah yes, thanks

emiltin:
i think it's better to make tracktype and smoothness tables optional?

oxidase:
@emiltin i think tables must be required if handler is used, otherwise it is better to split the handler into two parts and don't use in bicycle and foot profiles the optional handler

emiltin:
yeah my conclusion too

emiltin:
hmm, why can't i chery-pick the commit?

> git fetch origin
> git cherry-pick d3da40f
fatal: bad revision 'd3da40f'

emiltin:
seems i need to use the commit ac63ab649

TheMarex:
Rebasing on `master` to include cedeb15adee8acd3061a01d6f3b5f26e1f879e1f

profiles/bicycle.lua
```diff
@@ -1,96 +1,12 @@
 api_version = 1
 
 -- Bicycle profile
-
 local find_access_tag = require("lib/access").find_access_tag
+local Set = require('lib/set')
+local Sequence = require('lib/sequence')
+local Handlers = require("lib/handlers")
+local next = next       -- bind to local for speed
```
TheMarex: Can you elaborate in this `next` change?

emiltin: i read looking up global functions can be slow in lua. by assigning to a local it should be faster. the reason i bother in this case is that the function is used in the initial check for routabilitity, which is performed on every single way. haven't measured the effect though.

Issue #3653, opened 2017-02-04
Address sanitizer reports undefined behavior after #3646


springmeyer:
After #3646 with clang++ 3.9.1 and `-fsanitize=address,undefined` on OS X I'm seeing:

```
backend/src/extractor/extraction_containers.cpp:47:41: runtime error: member call on misaligned address 0x00011001d15c for type 'OSMNodeID', which requires 8 byte alignment
0x00011001d15c: note: pointer points here
  fc 00 00 00 c6 9f 98 50  00 00 00 00 12 a0 98 50  00 00 00 00 55 55 55 c0  55 55 55 c0 00 00 00 80
              ^ 
SUMMARY: AddressSanitizer: undefined-behavior /Users/dane/projects/osrm-backend/src/extractor/extraction_containers.cpp:47:41 in 
```

Full details:

<details>
Running osrm-extract...
../../scripts/timer.sh "osrm-extract" /Users/dane/projects/osrm-backend/build-osrm/osrm-extract monaco.osm.pbf -p ../../profiles/car.lua
[info] Using script ../../profiles/car.lua
[info] Input file: monaco.osm.pbf
[info] Profile: car.lua
[info] Threads: 4
[info] Parsing in progress..
[STXXL-MSG] STXXL v1.4.1 (prerelease/Release)
[STXXL-ERRMSG] Warning: no config file found.
[STXXL-ERRMSG] Using default disk configuration.
[STXXL-MSG] Disk '/var/tmp/stxxl' is allocated, space: 1000 MiB, I/O implementation: syscall autogrow delete_on_exit queue=0 devid=0
[STXXL-ERRMSG] stxxl::aligned_alloc: disabling realloc()
[info] input file generated by osmconvert 0.7T
[info] timestamp: 2016-03-05T00:26:02Z
[info] Found 3 turn restriction tags:
[info]   motorcar
[info]   motor_vehicle
[info]   vehicle
[info] Parsing finished after 1.25342 seconds
[info] Raw input contains 180118 nodes, 27377 ways, and 212 relations
[info] Sorting used nodes        ... ok, after 0.007841s
[info] Erasing duplicate nodes   ... ok, after 0.006665s
[info] Sorting all nodes         ... ok, after 0.207063s
[info] Building node id map      ... ok, after 0.052193s
[info] setting number of nodes   ... ok
[info] Confirming/Writing used nodes     ... ok, after 0.03648s
[info] Processed 15481 nodes
[...]

PR #3656, opened 2017-02-05
Fix of an assertion in bicycle profile

oxidase:
# Issue

In bicycle profile there is a possibility to have some accessible edges with undefined speed. The problem is that union of `access_tag_whitelist` and `access_tag_blacklist` sets is not the universal set of OSM access tags. 

PR changes `access_tag_whitelist` to a complement of `access_tag_blacklist`. Another possibility would be to list all access tags in `access_tag_whitelist` and `access_tag_blacklist`.

/cc @emiltin 

## Tasklist
 - [x] review
 - [x] adjust for comments

## Requirements / Relations
PR is rebased on top of #3654 and #3653 to check bicycle.lua on the planet size OSM network.

### DIFF
--- a/profiles/bicycle.lua
+++ b/profiles/bicycle.lua
@@ -338,10 +338,8 @@ function way_function (way, result)
     result.forward_mode = mode.train
     result.backward_mode = mode.train
     -- railways
-    if (not access and data.highway) or (access and not profile.access_tag_blacklist[access]) then
-      result.forward_speed = profile.railway_speeds[railway]
-      result.backward_speed = profile.railway_speeds[railway]
-    end
+    result.forward_speed = profile.railway_speeds[railway]
+    result.backward_speed = profile.railway_speeds[railway]
   elseif amenity and profile.amenity_speeds[amenity] then
     -- parking areas
     result.forward_speed = profile.amenity_speeds[amenity]
@@ -350,7 +348,7 @@ function way_function (way, result)
     -- regular ways
     result.forward_speed = profile.bicycle_speeds[data.highway]
     result.backward_speed = profile.bicycle_speeds[data.highway]
-  elseif access and not profile.access_tag_blacklist[access]  then
+  elseif access and profile.access_tag_whitelist[access]  then
     -- unknown way, but valid access tag
     result.forward_speed = default_speed
     result.backward_speed = default_speed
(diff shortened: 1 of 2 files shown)


emiltin:
is there are cucumber test that shows the problem? is it this one? https://github.com/Project-OSRM/osrm-backend/blob/52bf640e6d140b41ec6cbb0262e7097ed0a03516/features/bicycle/access.feature#L186

oxidase:
yes, the test fails without change due to assertion, because `psv` is not in the white or black list. 

Previously it was unnoticed because of missing `throw` statement and such ways had weights value 1 because of [pulling up -1 to 1 here](https://github.com/Project-OSRM/osrm-backend/blob/445225bf6dc719ef4f91fe6ab7e6fa2f4941fef7/src/extractor/extraction_containers.cpp#L454). [here](https://github.com/Project-OSRM/osrm-backend/pull/3632#discussion_r98888543) is the explanation of the weight data flow before merging #2399.

profiles/bicycle.lua
```diff
@@ -338,7 +338,7 @@ function way_function (way, result)
     result.forward_mode = mode.train
     result.backward_mode = mode.train
     -- railways
-    if (not access and data.highway) or profile.access_tag_whitelist[access] then
+    if (not access and data.highway) or (access and not profile.access_tag_blacklist[access]) then
```
TheMarex: Looking at https://github.com/Project-OSRM/osrm-backend/blob/cb32971a664ea97760173e06dbfd927779efc659/profiles/bicycle.lua#L286 the second clause should always be true since for blacklisted ways we don't get this far. I think the fix would be to not check access tags (which are meant for bikes, not trams) at all.

TheMarex: Looking at https://github.com/Project-OSRM/osrm-backend/blob/cb32971a664ea97760173e06dbfd927779efc659/profiles/bicycle.lua#L286 the second clause should always be true since for blacklisted ways we don't get this far. I think the fix would be to not check access tags (which are meant for bikes, not trams) at all.

I can grab this.

oxidase: :+1: but the second clause is not similar to the upper condition. Here condition is *access and not in the blacklist* and upper one is *access and in the blacklist*. it will be ok to fix it above. The original problem is that  the set {¬(whitelist ∪ blacklist)} is non-empty and edges with access tags from the set have undetermined speeds and travel modes but this was ignored before.

Issue #3131, opened 2016-10-20
Test LUA profiles directly


emiltin:
A lot of the Cucumber tests are really just testing the tag parsing that happens in the LUA profiles. It would be simpler and faster to have tests that tests the lua code directly, rather than having to go through constructing an .osm file, extracting, contracting and querying.

Testing would simply load the LUA file, provide and input way/node and check the output.

One issue is that there are a few C++ methods that are called from LUA. Testing LUA code would be simpler if this could be avoided.

daniel-j-h:
Related: having pure-Lua profiles would also allow for https://github.com/Project-OSRM/osrm-backend/issues/3864

PR #3658, opened 2017-02-05
[WIP] direct testing of LUA profiles

emiltin:
# Issue
Begins to address #3131, with a LUA script for testing profiles directly. No osrm binaries or OSM files are involved.

The way it works is simple. You call the script form the command line, passing the name of the profile, and some input tags. The profile processes the way and the result table is printed. Example:

```
~/code/osrm-backend/profiles$  lua5.1 ./lib/testing.lua car maxspeed 30 highway primary oneway yes name 'Linea Lane'
{ 
  backward_mode = 0,
  backward_speed = 24,
  destinations = '',
  forward_mode = 1,
  forward_speed = 24,
  is_startpoint = true,
  name = 'Linea Lane',
  road_classification = { 
    may_be_ignored = false,
    road_priority_class = 4 
  } 
}
```

With a bit of cucumber js support code we could then write cucumber test like this:
```gherkin
@tags @car
Feature: Car - Speed tags

  Background:
    Given the profile "car"

  Scenario: Car - parsing of maxspeed tags    
    Given the way
      | highway  | primary    |
      | name     | Linea Lane |
      | maxspeed | 30         |
    When I process the way 
    Then I should get
      | name           | Linea Lane |
      | forward_mode   | driving    |
      | forward_speed  | 24         |
      | backward_mode  | 0          |
      | backward_speed | 0          |
      | is_startpoint  | true       |
```

I think these tests would be very fast to run, since there are no OSM files or binaries to run.
 
There a a few c++ helper functions that profiles can call, these would have to be rewritten in lua to test things that uses these helpers.

## Tasklist
 - [ ] Is this a good idea?
 - [ ] Reimplement C++ helper functions in LUA
 - [ ] Add cucumber.js support code
 - [ ] review
 - [ ] adjust for comments

### DIFF
--- a/features/lib/table_diff.js
+++ b/features/lib/table_diff.js
@@ -17,73 +17,107 @@ String.prototype.padRight = function(char, length) {
 
 module.exports = function (expected, actual) {
     let headers = expected.raw()[0];
-    let diff = [];
+    let expectedRows = expected.hashes();
     let tableError = false;
+    let statusRows = [];
+    let columnStatus = {}
 
-    expected.hashes().forEach((expectedRow, i) => {        
+    expectedRows.forEach((expectedRow, i) => {
         var rowError = false;
-
+        statusRows[i] = {};
+        var statusRow = statusRows[i];
         for (var key in expectedRow) {
             var actualRow = actual[i]
             var row
             if (unescapeStr(expectedRow[key]) != actualRow[key]) {
-                
-                row = Object.assign({}, expectedRow, {diff_status: 'expected'})
-                diff.push(row);
-                
-                row = Object.assign({}, actualRow, {diff_status: 'actual'})
-                diff.push(row);
-                
+                statusRow[key] = false;
                 tableError = true;
-                rowError = true;
-                break;
+                columnStatus[key] = false;
             }
         }
-        if( !rowError ) diff.push(expectedRow);
+    });
+    
+    if (!tableError) return null;
+
+
+    // determine column widths
+    var widths = {};
+    var wantStr = '(-) ';
+    var gotStr  = '(+) ';
+    var okStr   = '    ';
+
+    headers.forEach( (key) => {
+        widths[key] = key.length;
     });
 
-    if (!tableError) return null;
-    
-    
-    // insert a hash of headers where key=value so we can treat
-    // the header row like other rows in the processing below
-    var header_hash = {}
-    headers.forEach( (key,i) => {
-        header_hash[key] = key;
-    });
-    diff.unshift( header_hash )
-    
-    // determine column widths
-    var widths = [];
-    diff.forEach((row) => {
+    expectedRows.forEach((row,i) => {
         var cells = []
-        headers.forEach( (key,i) => {
-            var s = row[key]
-            var length = s.length;
-            if(widths[i]==null || length > widths[i])
-              widths[i] = length;
+        headers.forEach( (key) => {
+            var content = row[key]
+            var length = content.length;
+            if(widths[key]==null || length > widths[key])
+              widths[key] = length;
        });
     });
 
+
     // format
-    var lines = ['Tables were not identical:'];
-    diff.forEach((row) => {
-        var cells = []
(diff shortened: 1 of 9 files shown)


daniel-j-h:
> Reimplement C++ helper functions in LUA

Hm this is my main problem with your approach: 1/ is it feasible for all helper functions and 2/ then why not completely do it in Lua :)

emiltin:
as far as i know it's only these 3 functions? 

- function durationIsValid
- function parseDuration
- canonicalizeStringList

emiltin:
what do you mean by "why not completely do it in Lua"?

daniel-j-h:
I meant: can we implement these functions completely in Lua. Then your approach would be much easier.

emiltin:
i'm not sure exactly what these functions do. durationIsValid and parseDuration sounds like they might be simply. not sure what canonicalizeStringList do?

daniel-j-h:
@emiltin See https://github.com/Project-OSRM/osrm-backend/blob/master/include/extractor/extraction_helper_functions.hpp#L121-L135

emiltin:
thanks. doesn't seem too complicated

emiltin:
another question is: do we expect new helpers which can only bo implemented on the c++ side?

daniel-j-h:
Not right now, no.

emiltin:
Another thing this would enable is easier testing across multiple profiles.

A lot of behaviour is similar across profiles. Now that we have factored out common code to handlers, it's easier to implement such behaviour across profiles. But testing the behaviour still requires writing similar tests for all the profiles.

A fast way to check the tag parsing makes it practical do do something like:

```gherkin
Scenario: Parsing of basic tags    
  Given the way
    | highway  | primary    |
    | name     | Linea Lane |
  When I process the way 
  Then I should get
    | profile | name       | forward_mode | backward_mode | forward_speed | backward_mode |
    | car     | Linea Lane | driving      | driving       | 50            | 50            |
    | bike    | Linea Lane | cycling      | cycling       | 20            | 20            |
    | foot    | Linea Lane | walking      | walking       | 5             | 5             |
```

emiltin:
Possible cucumber syntax for single profile testing. The part before '>' are input tags. The part after '>' is the expected output from the profile.

```gherkin
    When processing way tags I should get  
      | highway  | oneway | > | forward_mode | backward_mode | forward_speed | backward_speed |
      | cycleway |        |   | cycling      | cycling       | 15            | 15             |
      | cycleway | yes    |   | cycling      |               | 15            |                |
      | primary  |        |   | cycling      | cycling       | 15            | 15             |
      | primary  | yes    |   | cycling      |               | 15            |                |
```

emiltin:
It's now possible to run scenarios like this:

```gherkin
When processing way tags I should get  
  | highway  | oneway | > | forward_mode | backward_mode |
  | cycleway |        |   | 2            | 2             |
  | cycleway | yes    |   | 2            | 6             |
  | footway  |        |   | 6            | 6             |
  | footway  | yes    |   | 6            | 6             |
```

Running these tests are fast (and no caching required), but not super fast, because the LUA interpreter and profile is currently being loaded for every row.

A great speedup could probably we achieved by using a simple LUA TCP server, which loads the interpreter and profiles once and can be called with input tags and returns the profile response.

Also need a portable way to get the name of the lua interpreter: lua, lua5.1, lua5.2...?

emiltin:
added a lua tcp server which you can launch and keep in the background. it receives tags encoded as json, run them through the profile way_functions and returns the output values encoded as json. this could be used to very quickly run large number of tag tests from cucumber.

however it requires the 'luasocket' luamodule, which is a bit akward to install for lua 5.2:
https://stackoverflow.com/questions/39760619/lua-cannot-find-installed-luarocks-on-ubuntu

emiltin:
raster support in profiles depend on c++ helper functions in the lua class "sources", which is not practical to implement in lua. so test that depend on these helpers cannot be testing in a pure-lua environment.

DennisOSRM:

Issue #3650, opened 2017-02-03
Trip plugin should return a route that begins at the first coordinate unless otherwise specified


ghoshkaj:

chaupow:
Is this a duplicate of #1939 ?
If so, let's close one :)

danpat:
Yup, dupe, good catch @chaupow :-)

Issue #3634, opened 2017-01-31
Make trip only return a trip if all waypoints can be included


TheMarex:
Currently we run an SCC analysis on every distance table we generate for the trip plugin and only compute trips for each each component.

This can have unexpected results when users accidentally include waypoints from other continents. In that case the trip plugin will return multiple trips, one for each component. This behavior is notably different then for example the `route` plugin which would generate an error in that case.

The most extreme case is when each waypoint is in an own component leading to 0 duration trips that only contain one waypoint each.

To make this more consistent with the `/route` plugin I would suggest just throwing an `NoTrip` error if a single trip with all waypoints is impossible.

To make migration easier we should add a parameter to `TripParameters` e.g. `bool TripParameters::split_trip = true`. That way we would only need to switch the default from `true` -> `false` for the next API iteration.

/cc @freenerd @chaupow @ghoshkaj

chaupow:
> To make migration easier we should add a parameter to TripParameters e.g. bool TripParameters::split_trip = true. That way we would only need to switch the default from true -> false for the next API iteration.

Can you explain a bit? Does that mean, that we should still return multiple trips if all waypoints can't be reached?

Implementation-wise, the easiest solution would be to throw `NoTrip` if there are two points that can't be reached. 

But that would break the current API right? How does `TripParameters::split_trip = true` help here? If we will set `true -> false` it will be an API break anyways, right? Can't we throw it every time then?

TheMarex:
Capturing from voice: The current behavior is more a bug then intended behavior. We are going to fix it in the next release, without adding a query parameter for it.

Issue #3067, opened 2016-10-17
Make Trip Features uniquely valid


MoKob:
Currently, the outcome of our trip tests in `features/testbot/trips` is random. Since all tests use roundtrips of the same length, the success / failure of the tests is purely random.

Trips can fail in any random situation, making adjustments to the fixtures necessary. We should replace all tests with checks that have unique results.

chaupow:
should be solved when fixing https://github.com/Project-OSRM/osrm-backend/issues/1939

daniel-j-h:
What's blocking here? Is this resolved by now?

MoKob:
If the first coordinate is always chosen as the result, this should be resolved.

Issue #1939, opened 2016-01-29
Start point for round trip should be first coordinate


DEHWall:
Am I right in assuming that the first and last via_points should be the Start and End (being the same) points of a requested \trip url as follows.

Is there a known issue / fix available? 

It also affects the order of the route_instructions.

http://osrm:5000/trip?loc=**51.853842,-3.129779**&loc=51.75983,-3.36673&loc=51.63623,-3.95339&loc=51.64771,-4.03711&loc=51.5827,-3.00176&loc=51.48148,-2.52805&loc=51.37876,-2.43739&loc=51.43524,-2.6002&loc=51.56916,-1.80141&loc=51.5523,-1.7864&loc=51.63753,-1.83719&loc=51.90335,-2.07979&loc=**51.853842,-3.129779**
{
 "status_message": "Found trips",
 "status": 200,
 "trips": [
 ....
   "via_points": [
    [
     51.636239,-3.953348],[
     51.582627,-3.001746],[
     51.481475,-2.528061],[
     51.435281,-2.600249],[
     51.37858,-2.437094],[
     51.552301,-1.7864],[
     51.569181,-1.801448],[
     51.63735,-1.837459],[
     51.903344,-2.079795],[
    *\* 51.854382,-3.12917],[    <-------
     51.854382,-3.12917],[    <-----**
     51.759672,-3.366476],[
     51.647716,-4.037114],[
     51.636239,-3.953348]],
   "permutation": [
    2,4,5,7,6,9,8,10,11,**0,12**,1,3],

Many thanks
Dave

TheMarex:
@DEHWall omit the last location entry, all trips are planned as round-trips automatically.

DEHWall:
Still does not do as expected - am I misunderstanding the functionality here?

Given that the starting point is 51.853842,-3.129779 

SelectedStops:
0	51.853842,-3.129779 
1	51.78524,-3.87921
2	51.66592,-3.30539 
3	51.78099,-3.24361 
4	51.6192,-3.93939 

http://osrm:5000/trip?loc=51.853842,-3.129779&loc=51.78524,-3.87921&loc=51.66592,-3.30539&loc=51.78099,-3.24361&loc=51.6192,-3.93939

Nearest (replot points osrm/nearest)
0	51.854382,-3.12917
1	51.785243,-3.879195
2	51.665925,-3.305397
3	51.780753,-3.243968
4	51.61924,-3.939364

{
 "status_message": "Found trips",
[...]

PR #3660, opened 2017-02-06
Fix trip response such that it starts at the first coordinate

chaupow:
# Issue

Currently, the response for roundtrips will return a roundtrip starting at an arbitrary location. 
This means that 
* if a roundtrip between `n0, n1, n2, n3, n4, n5` is requested
* the response could start anywhere e.g. `n4, n2, n1, n3, n0, n5, n4`

This PR will solve this such that the roundtrip starts (and ends) at the first location given in the request:

* if a roundtrip between `n0, n1, n2, n3, n4, n5` is requested
* the response will start at n0 e.g. `n0, n5, n4, n2, n1, n3, n0`

This will solve
#1939
#3067

and partially #3650

## Tasklist
 - [ ] Remove or add assertion of finding zero index depending on the answer [here](https://github.com/Project-OSRM/osrm-backend/issues/3634#issuecomment-277690675)
 - [ ] update relevant [Wiki pages](https://github.com/Project-OSRM/osrm-backend/wiki) & API docs
 - [ ] review
 - [ ] adjust for comments

## Requirements / Relations

### DIFF
--- a/features/testbot/trip.feature
+++ b/features/testbot/trip.feature
@@ -49,7 +49,7 @@ Feature: Basic trip planning
 
         When I plan a trip I should get
             | waypoints               | trips         |
-            | a,b,c,d,e,f,g,h,i,j,k,l | cbalkjihgfedc |
+            | a,b,c,d,e,f,g,h,i,j,k,l | alkjihgfedcba |
 
 
     Scenario: Testbot - Trip planning with less than 10 waypoints tfse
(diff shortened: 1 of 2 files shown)


chaupow:
Edit: The PR looks messy because it uses #3408 as a base (to keep merge conflicts low). To see relevat changes, just look at the last commit or compare to the other branch:
https://github.com/Project-OSRM/osrm-backend/pull/3660/commits/5f2d431f9b089ff0aa4dba773a62ba654fc4990f

chaupow:
Currently, it changes the roundtrip such that it starts at `n0`, but as we currently have the problem of having several roundtrips if the nodes are not connected, I am not sure what the expected behaviour should be:

E.g. requesting `n0, n1, n2, n3, n4, n5` can result in two response round-trips:

* a roundtrip of `n0, n1, n2, n3`
* a roundtrip of `n4, n5`

The expected behaviour for the first case is kind of clear (starting at `n0`?). What is the expected behaviour for the second roundtrip? The node with the smallest index? Might be related to the discussion of several round trips [here](https://github.com/Project-OSRM/osrm-backend/issues/3634#issuecomment-277690675)

@ghoshkaj @TheMarex @danpat ideas? Personally I'd love to not return several trips at all with coming 6.0, but... this fix might be wished to be merged earlier than 6.0

PR #3662, opened 2017-02-06
Removes distance field from v5 Waypoint object docs

daniel-j-h:
Waypoint objects have
- name
- location
- hint

fields. The only API adding `distance` to waypoints is the Nearest
plugin. And it documents this behavior independently.

The docs are wrong here from what I can tell. See:
- https://github.com/Project-OSRM/osrm-backend/blob/5.5/src/engine/api/json_factory.cpp#L290-L297
- https://github.com/Project-OSRM/osrm-backend/blob/5.5/include/engine/api/nearest_api.hpp#L37-L44

### DIFF
--- a/docs/http.md
+++ b/docs/http.md
@@ -721,7 +721,6 @@ Object used to describe waypoint on a route.
 
 - `name` Name of the street the coordinate snapped to
 - `location` Array that contains the `[longitude, latitude]` pair of the snapped coordinate
-- `distance` The distance of the snapped point from the original
 - `hint` Unique internal identifier of the segment (ephemeral, not constant over data updates)
    This can be used on subsequent request to significantly speed up the query and to connect multiple services.
    E.g. you can use the `hint` value obtained by the `nearest` query as `hint` values for `route` inputs.
--- a/docs/http.md
+++ b/docs/http.md
@@ -638,12 +638,8 @@ step.
 | `turn`                 | `modifier` indicates the change in direction accomplished through the turn                                                |
 | `depart`/`arrive`      | `modifier` indicates the position of departure point and arrival point in relation to the current direction of travel      |
 
-- `exit` An optional `integer` indicating number of the exit to take. The property exists for the following `type` property:
-
-| `type`                 | Description                                                                                                               |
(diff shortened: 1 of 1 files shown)


daniel-j-h:
curl 'http://router.project-osrm.org/route/v1/driving/13.4301,52.5109;13.4265,52.5080;13.4194,52.5072' | jq ''

TheMarex:
Not sure why we only have it in the `nearest` plugin but, we should fix the docs however.

docs/http.md
```diff
@@ -638,12 +638,8 @@ step.
 | `turn`                 | `modifier` indicates the change in direction accomplished through the turn                                                |
 | `depart`/`arrive`      | `modifier` indicates the position of departure point and arrival point in relation to the current direction of travel      |
 
-- `exit` An optional `integer` indicating number of the exit to take. The property exists for the following `type` property:
-
-| `type`                 | Description                                                                                                               |
-|------------------------|---------------------------------------------------------------------------------------------------------------------------|
-| `roundabout`/`rotary`         | Number of the roundabout exit to take. If exit is `undefined` the destination is on the roundabout.                       |
-| else                   | Indicates the number of intersections passed until the turn. Example instruction: `at the fourth intersection, turn left` |
+- `exit` An optional `integer` indicating number of the exit to take. The property exists for the `roundabout` / `rotary` property:
```
TheMarex: Where is this change coming from? I don't think it is correct?

daniel-j-h: Separate commit. It is correct, exit counting only ever happens for roundabouts
- https://github.com/Project-OSRM/osrm-backend/blob/f3ebba7a79a3e499584ccd11d1f9381f16a2c0f4/src/engine/api/json_factory.cpp#L197-L198
- https://github.com/Project-OSRM/osrm-backend/blob/f3ebba7a79a3e499584ccd11d1f9381f16a2c0f4/src/engine/guidance/post_processing.cpp#L153

TheMarex: Wait so we broke the API an nobody noticed? Well, guess it is too late now.

PR #3675, opened 2017-02-08
[Trip] Support more parameter combinations of fixed start/end when requesting roundtrips

chaupow:
# Issue
This PR implements all supported parameter combinations tagged as `Shouldn't be too hard to extend to`

|  | Roundtrip | NoRoundtrip |
| :-- | :-- | :-- |
| Fixed Start&End (FSE) |Shouldn't be too hard to extend to|Already Implemented|
| Fixed Start (FS)  |Shouldn't be too hard to extend to| Won't do this for now|
| Fixed End (FE)  |Shouldn't be too hard to extend to|Won't do this for now|
| Nothing fixed  | Already Implemented |Won't do this for now||

## Tasklist
 - [ ] review
 - [ ] adjust for comments

## Requirements / Relations
based on #3408 
based on #3675

### DIFF
--- a/features/testbot/trip.feature
+++ b/features/testbot/trip.feature
@@ -5,7 +5,7 @@ Feature: Basic trip planning
         Given the profile "testbot"
         Given a grid size of 10 meters
 
-    Scenario: Testbot - Roundtrip with waypoints <10
+    Scenario: Testbot - Trip: Roundtrip with waypoints <10
         Given the node map
             """
             a b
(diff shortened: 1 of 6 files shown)


features/testbot/trip.feature
```diff
@@ -51,6 +51,64 @@ Feature: Basic trip planning
             | waypoints               | trips         |
             | a,b,c,d,e,f,g,h,i,j,k,l | alkjihgfedcba |
 
+    Scenario: Testbot - Roundtrip FS waypoints >10
+        Given the query options
+            | source | first  |
```
ghoshkaj: Move these to the "When I plan a trip I should get" portion of the test

Issue #3641, opened 2017-02-02
Assertion thrown during map matching when traffic data loaded


danpat:
We're hitting the following assertion:

<details>
  <summary>"bt full" output from the asserting thread (click to expand)</summary>
  <pre>
#0  0x00007ffff69a5c37 in raise () from /lib/x86_64-linux-gnu/libc.so.6
No symbol table info available.
#1  0x00007ffff69a9028 in abort () from /lib/x86_64-linux-gnu/libc.so.6
No symbol table info available.
#2  0x00007ffff72bbbbd in __gnu_cxx::__verbose_terminate_handler() () from /usr/lib/x86_64-linux-gnu/libstdc++.so.6
No symbol table info available.
#3  0x00007ffff72b9b96 in ?? () from /usr/lib/x86_64-linux-gnu/libstdc++.so.6
No symbol table info available.
#4  0x00007ffff72b9be1 in std::terminate() () from /usr/lib/x86_64-linux-gnu/libstdc++.so.6
No symbol table info available.
#5  0x0000000000df8b96 in (anonymous namespace)::assertion_failed_msg_helper (
    expr=0xf53447 <.L.str.153> "forward_core_heap.GetData(middle).parent == middle && reverse_core_heap.GetData(middle).parent == middle", msg=0xf46345 <.L.str.91> "",
    function=0xf531a4 <.L__PRETTY_FUNCTION__._ZNK4osrm6engine18routing_algorithms21BasicRoutingInterfaceINS0_10datafacade14BaseDataFacadeENS1_11MapMatchingIS4_EEE14SearchWithCoreERKS4_RNS_4util10BinaryHeapIjjiNS0_8HeapDataENSA_19UnorderedMapStorageIjiEEEESG_SG_SG_RiRSt6vectorIjSaIjEEbbi> "void osrm::engine::routing_algorithms::BasicRoutingInterface<osrm::engine::datafacade::BaseDataFacade, osrm::engine::routing_algorithms::MapMatching<osrm::engine::datafacade::BaseDataFacade> >::Search"...,
    file=0xf52ceb <.L.str.149> "/home/travis/build/Project-OSRM/node-osrm/deps/osrm-backend-Debug/include/engine/routing_algorithms/routing_base.hpp", line=718)
    at /home/travis/build/Project-OSRM/node-osrm/deps/osrm-backend-Debug/src/util/assert.cpp:14
No locals.
[...]

PR #3676, opened 2017-02-08
Add assertion to check weights consistency in contractor after update

oxidase:
# Issue

Issues #3641 and #3429 show an assertion in `osrm-routed` but the the reason is in `osrm-contract`  when summed up segment weights of the edge are greater than edge-based edge weight. Without the check it can be observed only in `osrm-routed` that appears as negative weights in the initial edges due to snapping of phantoms on segments with inconsistent weights.

The check is disabled for turn updates as it currently allowed to use negative turn weight updates that can lead to the same issue. The use-case should clarified  and feature tests are adjusted.

EDIT: the check also needed as a seatbelt for #77 `turn_function` weight updates that can also bring negative weights as in the test `Step weights -- segment_function and turn_function with weight precision`

## Tasklist
 - [x] review
 - [x] adjust for comments

/cc @TheMarex @danpat

### DIFF
--- a/features/testbot/weight.feature
+++ b/features/testbot/weight.feature
@@ -215,7 +215,7 @@ Feature: Weight tests
         end
         function turn_function (turn)
           print (turn.angle)
-          turn.weight = turn.angle / 100
+          turn.weight = 2 + turn.angle / 100
           turn.duration = turn.angle
         end
         """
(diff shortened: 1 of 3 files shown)


features/testbot/weight.feature
```diff
@@ -215,7 +215,7 @@ Feature: Weight tests
         end
         function turn_function (turn)
           print (turn.angle)
-          turn.weight = turn.angle / 100
+          turn.weight = 2 + turn.angle / 100
```
TheMarex: Urg yeah this can be negative since angle is +- 180

TheMarex:
https://travis-ci.org/Project-OSRM/osrm-backend/jobs/199628678#L4541-L4554

@danpat can you take a look at this OSX failure?

oxidase:
I have checked the failing case: new assertion `BOOST_ASSERT_MSG(to_weight >= 0, "to weight is invalid");` is failing only on OSX, because weight = -200, edge_weight = 167 for the first snapped point, but still don't understand why it happens on OSX, on linux weight = 200.

oxidase:
The test `features/testbot/traffic_turn_penalties.feature:52` fails on osx and passes on linux, but there is no system dependency, just different order of edges after the contraction. On both system results are wrong due to negative values, but on linux it is hidden by snapping a shortcut edge with weight 167+200+200=567, so inserted to the heap value is 567-200=367. On osx the snapped edge is forward non-contracted one and the inserted  value into heap is 167-200=-33 that breaks the assertion. 

I have removed assertion from the commit and created  #3683

TheMarex:
@oxidase thanks for digging here and confirming this is an "expected" bug. I think this would be fine to merge now.

PR #3678, opened 2017-02-09
Replace "highway" with "traffic_signals" in bicycle.lua

mloskot:
where the latter was intended.

------

Corrects copy/paste/typo-like bug.

### DIFF
--- a/profiles/bicycle.lua
+++ b/profiles/bicycle.lua
@@ -222,7 +222,7 @@ function node_function (node, result)
   end
 
   -- check if node is a traffic light
-  local tag = node:get_value_by_key("highway")
+  local tag = node:get_value_by_key("traffic_signals")
   if tag and "traffic_signals" == tag then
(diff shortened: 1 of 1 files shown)


profiles/bicycle.lua
```diff
@@ -222,7 +222,7 @@ function node_function (node, result)
   end
 
   -- check if node is a traffic light
-  local tag = node:get_value_by_key("highway")
+  local tag = node:get_value_by_key("traffic_signals")
   if tag and "traffic_signals" == tag then
```
emiltin: seems odd to then compare to the value "traffic_signals"?

TheMarex:
http://wiki.openstreetmap.org/wiki/Tag:highway=traffic_signals

The key is `highway` the value is `traffic_signals`.

mloskot:
@TheMarex OSM/OSRM newbie error. Sorry for time wasting. Thanks

PR #3688, opened 2017-02-10
Routing Engine / SCC Engine are disconnected

MoKob:
# Issue

Resolves https://github.com/Project-OSRM/osrm-backend/issues/3689

It turns out that the `FindComponents` function added u-turn functionality for all edges which caused the disconnect between the routing graph and the graph used for SCC calculation.

Turns out this is even deeper: In the edge based graph edges can be of different components, even though they are the same road.

e.g. `a-b` with u-turns disabled on both ends will have a component for `a-b` and a component for `b-a`. So either we allow turning around on all segments, or we would require a component ID for all forward/backward edges and adjust snapping accordingly.

## Tasklist
 - [ ] find and fix the actual issue
 - [ ] review
 - [ ] adjust for comments

## Requirements / Relations

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
       - Fix #3515 adjusted number of `nodes` in `annotation`
       - Fix #3605 Fixed a bug that could lead to turns at the end of the road to be suppressed
       - Fix #2844 handle up to 16777215 code units in OSM names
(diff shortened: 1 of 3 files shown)


MoKob:
Superseeded by https://github.com/Project-OSRM/osrm-backend/issues/4236

PR #3690, opened 2017-02-10
Don't bundle all Mason packages in third_party

TheMarex:
# Issue

This removed the `third_party` bundling of mason and only ships the `mason.cmake`.

## Tasklist
 - [x] review
 - [x] adjust for comments


TheMarex:
There is an interesting bootstrapping problem here since we use mason to install cmake but our `mason.sh` shells out to cmake. 😆 

Maybe we should install the cmake binaries manually again.

Issue #3684, opened 2017-02-10
Refactor `annotations` paramater


chaupow:
Crossposting from https://github.com/Project-OSRM/node-osrm/issues/297:

**tl;dr: node-osrm doesn't pass all tests anymore and its probably due to a bug in #3628**

<img width="1343" alt="screenshot 2017-02-10 11 35 09" src="https://cloud.githubusercontent.com/assets/804027/22823731/21f0ded4-ef85-11e6-908b-d2c30c9575d0.png">
https://travis-ci.org/Project-OSRM/node-osrm/builds

tests are failing due to response not having `annotations` even though the request is set to `annotations=true`

this is a blocker for 5.6

/cc @danpat @TheMarex

chaupow:
In oder to implement https://github.com/Project-OSRM/osrm-backend/issues/3563
and to not add a breaking API change, we currently have two parameter member variables:
```
bool annotations
AnnotationsType annotations_type
```

this makes node-osrm have to handle two parameters. it should be made such that there's only one parameter:
```
AnnotationsType annotations
```

and still making sure `annotations=true|false` doesnt fail.

daniel-j-h:
@chaupow this is already ticketed for v6 in https://github.com/Project-OSRM/osrm-backend/issues/3644.

Until then, node-osrm handles Javascript. Therefore for node-osrm you can handle

    annotations=true

keeping the old API and at the same time support

    annotations='speed'

and support the new API in Javascript.

chaupow:
> @chaupow this is already ticketed for v6 in #3644.

awesome. missed this. so this can be closed as the bug is fixable in `node-osrm`only.

Issue #3644, opened 2017-02-02
Remove redundant bool member for annotations param in RouteParameters


karenzshea:
Given the work in https://github.com/Project-OSRM/osrm-backend/pull/3628 which makes the `annotations` parameter for route and matching services accept an enum of individual values, including `All` and `None`, the boolean based annotations parameter can be removed from the interface. Needs to happen in v6 though because it is breaking.

daniel-j-h:
And we should think about encapsulation. At least for our public API.
Otherwise we will hit this in the future again.

PR #3692, opened 2017-02-10
Parse old annotations parameter true as new annotations_type all

karenzshea:
# Issue

Closes https://github.com/Project-OSRM/osrm-backend/issues/3684

## Tasklist
 - [x] review
 - [x] adjust for comments

## Requirements / Relations
Sort of fixes https://github.com/Project-OSRM/node-osrm/issues/297 but doesn't really matter, this pr mostly preserves backwards compatibility

### DIFF
--- a/include/engine/api/route_api.hpp
+++ b/include/engine/api/route_api.hpp
@@ -228,8 +228,16 @@ class RouteAPI : public BaseAPI
 
         std::vector<util::json::Object> annotations;
 
-        if (parameters.annotations_type != RouteParameters::AnnotationsType::None)
+        if (parameters.annotations_type != RouteParameters::AnnotationsType::None ||
+            parameters.annotations == true)
         {
+            auto requested_annotations = parameters.annotations_type;
+            if ((parameters.annotations == true) &&
+                (parameters.annotations_type == RouteParameters::AnnotationsType::None))
+            {
+                requested_annotations = RouteParameters::AnnotationsType::All;
+            }
+
             for (const auto idx : util::irange<std::size_t>(0UL, leg_geometries.size()))
             {
                 auto &leg_geometry = leg_geometries[idx];
@@ -243,7 +251,8 @@ class RouteAPI : public BaseAPI
                         });
                 }
 
-                if (parameters.annotations_type & RouteParameters::AnnotationsType::Duration)
+                // AnnotationsType uses bit flags, & operator checks if a property is set
+                if (ReqAnnotations & RouteParameters::AnnotationsType::Duration)
                 {
                     annotation.values["duration"] = GetAnnotations(
                         leg_geometry, [](const guidance::LegGeometry::Annotation &anno) {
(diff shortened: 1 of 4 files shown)


include/engine/api/route_api.hpp
```diff
@@ -228,8 +228,11 @@ class RouteAPI : public BaseAPI
 
         std::vector<util::json::Object> annotations;
 
-        if (parameters.annotations_type != RouteParameters::AnnotationsType::None)
+        if (parameters.annotations_type != RouteParameters::AnnotationsType::None || parameters.annotations == true)
         {
+            if (parameters.annotations == true)
+                parameters.annotations_type = RouteParameters::AnnotationsType::All;
```
oxidase: This will break new functionality with a list of values. It will always fallback to `All`

oxidase: This will break new functionality with a list of values. It will always fallback to `All`

include/engine/api/route_api.hpp
```diff
@@ -228,42 +228,50 @@ class RouteAPI : public BaseAPI
 
         std::vector<util::json::Object> annotations;
 
-        if (parameters.annotations_type != RouteParameters::AnnotationsType::None)
+        if (parameters.annotations_type != RouteParameters::AnnotationsType::None ||
+            parameters.annotations == true)
         {
+            auto ReqAnnotations = parameters.annotations_type;
+            if ((parameters.annotations == true) & (parameters.annotations_type &
+                RouteParameters::AnnotationsType::None))
```
chaupow: i think it should be `&&` instead of `&` in the first case because it is a boolean comparison?
```
 if ((parameters.annotations == true) && (parameters.annotations_type &
     RouteParameters::AnnotationsType::None))
```

Also I'm not sure what `(parameters.annotations_type & RouteParameters::AnnotationsType::None)` is doing. Is 
`(parameters.annotations_type == RouteParameters::AnnotationsType::None)` not more intuitive and does the same? Not sure though cause I don't know much about bit-magic

include/engine/api/route_api.hpp
```diff
@@ -228,42 +228,50 @@ class RouteAPI : public BaseAPI
 
         std::vector<util::json::Object> annotations;
 
-        if (parameters.annotations_type != RouteParameters::AnnotationsType::None)
+        if (parameters.annotations_type != RouteParameters::AnnotationsType::None ||
+            parameters.annotations == true)
```
karenzshea: I think there are _technically_ cases where only checking for `parameters.annotations` could be wrong. Say with a constructor where `annotations` defaults to `false` and then annotations_type is manually set.

include/engine/api/route_api.hpp
```diff
@@ -228,42 +228,50 @@ class RouteAPI : public BaseAPI
...
+                ReqAnnotations = RouteParameters::AnnotationsType::All;
+            }
+
             for (const auto idx : util::irange<std::size_t>(0UL, leg_geometries.size()))
             {
                 auto &leg_geometry = leg_geometries[idx];
                 util::json::Object annotation;
 
-                if (parameters.annotations_type & RouteParameters::AnnotationsType::Duration)
+                if (ReqAnnotations & RouteParameters::AnnotationsType::Duration)
```
chaupow: maybe we should add a comment explaining to the future that this basically means
"if Duration bit was set"

chaupow:
looks good overall 👍 
Have some questions about the first `if`-conditions but otherwise good to go I think.

Also:
 - [ ]  needs to be clangformatted to make travis happy.

include/engine/api/route_api.hpp
chaupow: Ah I get your point. Like, it should be good for how we use it now but better safe than sorry?

karenzshea: Seems reasonable to me to keep both checks, especially since this PR is supposed to be fixing the usage case in node-osrm that I didn't think of.

include/engine/api/route_api.hpp
```diff
@@ -228,42 +228,50 @@ class RouteAPI : public BaseAPI
 
         std::vector<util::json::Object> annotations;
 
-        if (parameters.annotations_type != RouteParameters::AnnotationsType::None)
+        if (parameters.annotations_type != RouteParameters::AnnotationsType::None ||
+            parameters.annotations == true)
         {
+            auto ReqAnnotations = parameters.annotations_type;
```
TheMarex: local variables use `lower_case` names. Maybe `annotations_type` would be a good name here:

```
auto annotations_type = parameters.annotations_type;
```

include/engine/api/route_api.hpp
TheMarex: `(parameters.annotations_type & RouteParameters::AnnotationsType::None` will always be false, needs to be `==` instead of `&`.

include/engine/api/route_api.hpp
chaupow: ok, just noticed: definitely keep this, it'll make `node-osrm` prettier :D

include/engine/api/route_api.hpp
```diff
@@ -228,42 +228,51 @@ class RouteAPI : public BaseAPI
...
-        if (parameters.annotations_type != RouteParameters::AnnotationsType::None)
+        if (parameters.annotations_type != RouteParameters::AnnotationsType::None ||
+            parameters.annotations == true)
         {
+            auto requested_annotations = parameters.annotations_type;
+            if ((parameters.annotations == true) &&
+                (parameters.annotations_type == RouteParameters::AnnotationsType::None))
+            {
+                requested_annotations = RouteParameters::AnnotationsType::All;
+            }
```
TheMarex: If this code block is moved before line 231 you can change the first if to also only check for the `::None` case. That reduces the points in the code that need to know about the work-a-round.

karenzshea:
Also to note: this problem will go away when we remove redundant bool parameters in v6 https://github.com/Project-OSRM/osrm-backend/issues/3644

Issue #3379, opened 2016-11-30
Lane indications suggest turning across potentially straight-through lanes


danpat:
http://map.project-osrm.org/?z=18&center=37.427114%2C-121.908921&loc=37.427438%2C-121.909849&loc=37.426428%2C-121.907891&hl=en&alt=0

![screen shot 2016-11-29 at 4 33 08 pm](https://cloud.githubusercontent.com/assets/1892250/20734950/8d045614-b651-11e6-8597-13d13edfd9f9.png)

The response for this route has:

![screen shot 2016-11-29 at 4 30 50 pm](https://cloud.githubusercontent.com/assets/1892250/20734878/3a9b7876-b651-11e6-8762-74b90f29a15c.png)

It seems to me like `valid: true` for the middle lane is not correct - this indicates you can turn right across traffic a potentially straight-through lane to the right.

The actual tagging for the `way` leading up to this intersection is:

`turn:lanes=left|none|none`

Our API response is indicating that it's OK to make the right-turn from the middle lane - that seems wrong here.  We should only have `valid:true` on the lane if it's either the right-most, *or* it has `right` as a lane indication.

/cc @daniel-j-h

PR #3693, opened 2017-02-10
Fix/3379

MoKob:
# Issue

Resolves https://github.com/Project-OSRM/osrm-backend/issues/3379

## Tasklist
 - [x] review
 - [x] adjust for comments

### DIFF
--- a/src/extractor/guidance/turn_lane_augmentation.cpp
+++ b/src/extractor/guidance/turn_lane_augmentation.cpp
@@ -107,25 +107,25 @@ LaneDataVector augmentMultiple(const std::size_t none_index,
         return std::make_pair(std::size_t{0}, std::size_t{0});
     }();
 
+    const auto intersection_range_first = intersection.begin() + range.first;
+    const auto intersection_range_end = intersection.begin() + range.second;
     const auto allowed_in_range =
-        std::count_if(intersection.begin() + range.first,
-                      intersection.begin() + range.second,
-                      [](const auto &road) { return road.entry_allowed; });
+        std::count_if(intersection_range_first, intersection_range_end, [](const auto &road) {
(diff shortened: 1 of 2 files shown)


src/extractor/guidance/turn_lane_augmentation.cpp
```diff
@@ -107,20 +106,65 @@ LaneDataVector augmentMultiple(const std::size_t none_index,
         util::Log(logWARNING) << "Failed lane assignment. Reached bad situation.";
         return std::make_pair(std::size_t{0}, std::size_t{0});
     }();
-    for (auto intersection_index = range.first; intersection_index < range.second;
-         ++intersection_index)
+
+    const auto allowed_in_range =
```
daniel-j-h: In the following, we need intersection.begin() + range.first/second a lot. Want to make temp. vars for them?

src/extractor/guidance/turn_lane_augmentation.cpp
```diff
@@ -107,20 +106,65 @@ LaneDataVector augmentMultiple(const std::size_t none_index,
...
         {
-            // FIXME this probably can be only a subset of these turns here?
-            lane_data.push_back(
-                {tag_by_modifier[intersection[intersection_index].instruction.direction_modifier],
-                 lane_data[none_index].from,
-                 lane_data[none_index].to,
-                 false});
+            if (intersection[intersection_index].entry_allowed)
+            {
+                // FIXME this probably can be only a subset of these turns here?
```
daniel-j-h: Care to elaborate?

MoKob: leftover from before. I feared this bug could happen, and it came true. Removing fixme

Issue #1392, opened 2015-02-24
Use Pearce SCC algorithm for component analysis


DennisOSRM:
Will be faster and needs much less memory than Tarjans canonical algorithm.

TheMarex:
We are talking about [this algorithm](http://damclubijmuiden.nl.server22.firstfind.nl/performance-rating/scc-10.1.1.102.1707-David-J-Pearce.pdf) right?

DennisOSRM:
yep

DennisOSRM:
Pearces algorithm is surprisingly simple. Implementing a recursive variant is a matter of copy-and-pasting the pseudo code from the paper. The more challenging aspect is implementing an iterative variant that doesn't use the system stack.

DennisOSRM:
Note to self: push children in reverse order onto stack in iterative variant to preserve DFS order of the traversal.

PR #3701, opened 2017-02-13
Implement Pearce SCC algorithm

oxidase:
# Issue

For #1392 Pearce SCC algorithm is similar to Tarjan SCC, but requires slightly less memory, but in worst case it uses the same amount of memory as Tarjans SCC. Time performance is nearly the same. 

Don't know if it is useful to use the Pearce SCC algorithm instead of the current one.

## Tasklist
 - [ ] review
 - [ ] adjust for comments

## Requirements / Relations

### DIFF
--- a/include/extractor/pearce_scc.hpp
+++ b/include/extractor/pearce_scc.hpp
@@ -1,0 +1,161 @@
+#ifndef PEARCE_SCC_HPP
+#define PEARCE_SCC_HPP
+
+#include "extractor/node_based_edge.hpp"
+#include "extractor/query_node.hpp"
+#include "util/deallocating_vector.hpp"
+#include "util/percent.hpp"
+#include "util/typedefs.hpp"
+
+#include "util/integer_range.hpp"
+#include "util/log.hpp"
+#include "util/std_hash.hpp"
+#include "util/timing_util.hpp"
+
+#include "osrm/coordinate.hpp"
+#include <boost/assert.hpp>
+#include <cstdint>
+
+#include <algorithm>
+#include <climits>
+#include <memory>
+#include <stack>
+#include <vector>
+
+namespace osrm
+{
+namespace extractor
+{
+
+template <typename GraphT> class PearceSCC
+{
+    std::vector<NodeID> components_index;
+    std::vector<std::size_t> component_size_vector;
+    const GraphT &m_graph;
+
+    struct StackFrame
+    {
+        NodeID v;
+        typename GraphT::EdgeRange edges;
+        bool root;
+    };
+
+  public:
+    PearceSCC(const GraphT &graph)
+        : components_index(graph.GetNumberOfNodes(), SPECIAL_NODEID), m_graph(graph)
+    {
+    }
+
+    void Run()
+    {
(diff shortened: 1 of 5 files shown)


include/extractor/pearce_scc.hpp
```diff
@@ -0,0 +1,154 @@
...
+{
+
+template <typename GraphT> class PearceSCC
+{
+    std::vector<NodeID> components_index;
+    std::vector<std::size_t> component_size_vector;
+    std::shared_ptr<const GraphT> m_graph;
+
+  public:
+    PearceSCC(std::shared_ptr<const GraphT> graph)
```
daniel-j-h: I don't understand why we need to take shared ownership over a graph here.

A `const GraphT&` should be good enough? The callsite can unwrap if needs be.

include/extractor/pearce_scc.hpp
```diff
@@ -0,0 +1,154 @@
...
+    {
+    }
+
+    void Run()
+    {
+        TIMER_START(SCC_RUN);
+        const NodeID max_node_id = m_graph->GetNumberOfNodes();
+
+        // recursion stack contains local variables: v, (w, end(w)), root
+        std::stack<std::tuple<NodeID, typename GraphT::EdgeRange, bool>> recursion_stack;
```
daniel-j-h: Maybe change from `tuple` and `get<n>` below to dedicated named local struct.

unit_tests/util/scc.cpp
```diff
@@ -0,0 +1,73 @@
...
+                       {3, 2},
+                       {3, 7},
+                       {4, 0},
+                       {4, 5},
+                       {5, 6},
+                       {6, 5},
+                       {7, 3},
+                       {7, 6}}};
+    // osrm::extractor::TarjanSCC<Graph> scc(std::make_shared<const Graph>(graph));
+    osrm::extractor::PearceSCC<Graph> scc(std::make_shared<const Graph>(graph));
```
daniel-j-h: No need for shared ownership.

daniel-j-h:
Can you benchmark this? Would be cool to see memory and runtime on the planet.

include/extractor/pearce_scc.hpp
oxidase: it is not needed, just copied constructor arguments from `TarjanSCC`

daniel-j-h:
We're currently seeing the partitioner taking roughly 40 GB for the planet.
And the Tarjan run before using above 50 GB on its own.

Might make sense to try this one out.

oxidase:
Just checked: the current implementation has no memory benefit over the original implementation :cold_sweat:. The problem is in `StackFrame` has a flag `root`  that is 32 bits and `edges` range that is 64 bits that can be reduced. The paper promises only memory reduction from `v(2+5w)` to `v(1+3w)` where `w` is 32 bits. Will try to fit to `v(1+3w)` estimation

daniel-j-h:
@oxidase even though it would be a nice to have feature, I would not prioritize this.

TheMarex:
Didn't pay off. :crying_cat_face:

Issue #3571, opened 2017-01-16
Lane Anticipation Seems a bit Over Eager at places


MoKob:
Lane Anticipation can apparently get out of hand in some situations. I suspect that our recently introduced thresholds might just be too large (see https://github.com/Project-OSRM/osrm-backend/pull/3517).

What we see right now is the following situation:

http://map.project-osrm.org/?z=12&center=34.218474%2C-118.428154&loc=34.201013%2C-118.452873&loc=34.201297%2C-118.375883&hl=en&alt=0 shows a lot of instructions for lanes:

<img width="1256" alt="screen shot 2017-01-16 at 18 02 14" src="https://cloud.githubusercontent.com/assets/12932279/21992250/03ad0100-dc16-11e6-8f13-8b9022fe6d7a.png">

Shortening the route a bit:

http://map.project-osrm.org/?z=14&center=34.206301%2C-118.410087&loc=34.201013%2C-118.452873&loc=34.201013%2C-118.404465&hl=en&alt=0

shows no instructions but `depart` and `arrive`

<img width="1110" alt="screen shot 2017-01-16 at 18 03 33" src="https://cloud.githubusercontent.com/assets/12932279/21992300/2fda6060-dc16-11e6-919e-8d45606ec264.png">

I would assume that we could handle this by allowing by not always considering the worst but by gradually allowing more lanes. E.g. if you have four lanes at every intersection, you don't need to consider all four when looking 10 intersections back.

/cc @daniel-j-h

PR #3495, opened 2016-12-23
Refactor post-processing and add test for turn location

MoKob:
# Issue

Works towards https://github.com/Project-OSRM/osrm-backend/issues/3158 and https://github.com/Project-OSRM/osrm-backend/issues/3293 

## Tasklist
 - [x] remove all debug output
 - [x] clean-up scenarios
 - [x] review
 - [x] adjust for comments

## Requirements / Relations
 Link any requirements here. Other pull requests this PR is based on?

### Guide:

- The refactor fixes an issue that prevented new-names from being suppressed when they are shorter than 100 meters. This was in before, but didn't work apparently. Many tests are adjusted for this in one of two ways:
  - added a given a grid size of 200 meters (possible if tests only rely on shape, not on lengths)
  - removed additional road names from `route` field
- I removed collapse considerations from `lane processing`. Lane processing should not handle collapse functionality. Collapse should run after lane processing and consider lanes to decide if collapse is possible.
- many tests are given the `locations` check as well, especially for collapse scenarios
- to avoid thresholds that are too large, many cases require lanes to check for road widths now

PR #3705, opened 2017-02-13
reduce verbosity of use-lane in combination with lane-anticipation

MoKob:
# Issue

Resolves https://github.com/Project-OSRM/osrm-backend/issues/3571.

## Tasklist
 - [x] review
 - [x] adjust for comments

## Requirements / Relations
none

Changes the results from the issue to:

screen shot 2017-02-13 at 16 37 37

The remaining use-lane is a result of the following use-lane which is 2 of 2 lanes but the 2 leftmost. The rightmost lanes becomes a turn right, which is now a reason to call out early for keeping left.

### DIFF
--- a/include/engine/guidance/lane_processing.hpp
+++ b/include/engine/guidance/lane_processing.hpp
@@ -20,7 +20,7 @@ namespace guidance
 // as separate maneuvers.
 OSRM_ATTR_WARN_UNUSED
 std::vector<RouteStep> anticipateLaneChange(std::vector<RouteStep> steps,
-                                            const double min_duration_needed_for_lane_change = 15);
+                                            const double min_duration_needed_for_lane_change = 10);
 
 } // namespace guidance
 } // namespace engine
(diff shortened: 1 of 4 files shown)


include/engine/guidance/lane_processing.hpp
```diff
@@ -20,7 +20,7 @@ namespace guidance
 // as separate maneuvers.
 OSRM_ATTR_WARN_UNUSED
 std::vector<RouteStep> anticipateLaneChange(std::vector<RouteStep> steps,
-                                            const double min_duration_needed_for_lane_change = 15);
+                                            const double min_duration_needed_for_lane_change = 10);
```
MoKob: original was just a bit long, we could think about going even lower

include/engine/guidance/post_processing.hpp
```diff
@@ -28,6 +28,7 @@ std::vector<RouteStep> collapseTurns(std::vector<RouteStep> steps);
 
 // A check whether two instructions can be treated as one. This is only the case for very short
 // maneuvers that can, in some form, be seen as one. Lookahead of one step.
+// This is only a pre-check and does not necessarily allow collapsing turns!!!
```
daniel-j-h: Maybe we should change then name - `collapsable` really sounds like checking for, well, collapsibility. :)

MoKob: well. its a precondition, I can see what we can do here. In my post-processing refactor, this should be already better, though.
One function over there might suffer from the same miss-readability. Will change https://github.com/Project-OSRM/osrm-backend/pull/3495/files#diff-2a61b83618a06f087b5d5f03384829ceR58, since this function here is not going to survive for long.

PR #3712, opened 2017-02-14
Allow routing on toll roads by default

### LINKED ISSUE #3711: Default car profile should route over toll roads
The OSRM demo server runs the default `car.lua` profile.  If we leave the default as "avoid tolls", this leads to some weird routing in high-profile areas with single-direction tolled access, like the Golden Gate Bridge and other major bridges in the San Francisco area:

screen shot 2017-02-14 at 11 42 29 am

For purely practical reasons, the default `car.lua` profile should allow toll access, otherwise people complain about how the demoserver works.

We've hit this before: https://github.com/Project-OSRM/osrm-backend/issues/2771#issuecomment-244352809

but we didn't have a regression test to catch the problem if it re-surfaced, which it now has.

danpat:
This fixes #3711 and re-enables toll-road routing by default in the `car.lua` profile.  It also adds a regression test so we don't do this again.

## Tasklist
 - [x] review
 - [x] adjust for comments

### DIFF
--- a/features/car/access.feature
+++ b/features/car/access.feature
@@ -224,12 +224,10 @@ Feature: Car - Restricted access
             | primary | no          | x     |
             | primary | snowmobile  | x     |
 
-     # To test this we need issue #2781
-     @todo
-     Scenario: Car - only toll=yes ways are ignored by default
+     Scenario: Car - toll=yes ways are enabled by default
         Then routability should be
             | highway | toll        | bothw |
-            | primary | yes         |       |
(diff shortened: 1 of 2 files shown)


daniel-j-h:
Urgh how did this make it back in again - and Cucumber tests being tagged as todo only :(

Yes, needs to go in asap!

TheMarex:
Urg good catch. :+1:

danpat:
Thanks to @oxidase for spotting this :-)

Issue #156, opened 2012-03-01
Respect access=destination


DennisOSRM:
Feature request: Don't route through streets with access=destination tag set, if start and/or target do not lie on them.

derTobsch:
This Feature would be great! :+1: We used a version which pa5cal made for as at synyx with the fix in it. But i would be very good if this would be a real feature in osrm :)

frodrigo:
The hack is still available somewhere ?

pa5cal:
@frodrigo the hack was to modify the mentioned EdgeBasedGraphFactory.cpp in line 316-317 with:
```
if(!edgeData1.isAccessRestricted  &&  edgeData2.isAccessRestricted)  { 
    distance += 10*1200;  
}
```

marcioaguiar:
Hi, @DennisOSRM 

Can we document the special cases so we can come up with a proper solution?

Thanks!

danpat:
@marcioaguiar The proper solution is to implement https://github.com/Project-OSRM/osrm-backend/issues/77, which would allows us to put arbitrary penalties on `access=destination` without modifying ETA values.  There are a bunch of other related use cases that this feature will be helpful for.

marcioaguiar:
Great! Thanks for the feedback, @danpat

derTobsch:
Now in version 5.1.0 ? :-/ It feels like this is one of the most missing feature...maybe we can support you to get this done faster?

danpat:
@derTobsch A patch that implements #77 would make this possible :smile:

emiltin:
would you accept such a patch? it will increase ram usage

danpat:
@emiltin Maybe someone will come up with a clever way to do it that doesn't use extra memory :-)

But seriously, if someone writes the patch, we can certainly make it a compile-time option, so it has no effect unless it's enabled.

yuryleb:
Still [routing](http://map.project-osrm.org/?z=17&center=59.969201%2C30.313787&loc=59.969452%2C30.313989&loc=59.970060%2C30.309218&hl=en&alt=0) through [vehicle=destination](https://www.openstreetmap.org/way/172747175) :worried:
![osrm routes through vehicle destination](https://cloud.githubusercontent.com/assets/4529411/22245349/e7e68cdc-e248-11e6-8f10-1ca9e438706e.png)

daniel-j-h:
Yes, OSRM does not respect the destination access _in the default profiles and on the demo server_ see

[...]

PR #3721, opened 2017-02-16
Penalize turns from non-local access only edges to local access only edges

karenzshea:
# Issue

Closes https://github.com/Project-OSRM/osrm-backend/issues/156

Instead of completely removing ways tagged with local access only or routing over them with impunity, this pr adds a heavy penalty on turn maneuvers from non-local access only edges to local access only edges. This allows routing onto a local access only, but only if the destination of the route ends on the way.

"Local access only" is defined in the `local_access_tag_list` set in the car lua profile. The default profile includes ways tagged like
- delivery
- private
- destination
- hov

## Tasklist
 - [x] review
 - [x] adjust for comments

### DIFF
--- a/include/extractor/extraction_way.hpp
+++ b/include/extractor/extraction_way.hpp
@@ -58,6 +58,8 @@ struct ExtractionWay
         turn_lanes_forward.clear();
         turn_lanes_backward.clear();
         road_classification = guidance::RoadClassification();
+        backward_restricted = false;
+        forward_restricted = false;
     }
 
     // These accessors exists because it's not possible to take the address of a bitfield,
@@ -106,6 +108,8 @@ struct ExtractionWay
     bool roundabout;
     bool circular;
     bool is_startpoint;
+    bool backward_restricted;
+    bool forward_restricted;
     TravelMode forward_travel_mode : 4;
     TravelMode backward_travel_mode : 4;
     guidance::RoadClassification road_classification;
--- a/profiles/car.lua
+++ b/profiles/car.lua
@@ -34,6 +34,9 @@ local profile = {
   speed_reduction            = 0.8,
   traffic_light_penalty      = 2,
   u_turn_penalty             = 20,
+  restricted_penalty         = 3000,
+  -- Note^: abstract value but in seconds correlates approximately to 50 min
+  -- meaning that a route through a local access way is > 50 min faster than around
 
   -- Note: this biases right-side driving.
   -- Should be inverted for left-driving countries.
@@ -63,18 +66,21 @@ local profile = {
     'vehicle',
     'permissive',
     'designated',
-    'destination',
-    'hov' -- we might filter hov out later depending on the avoid settings or add penalties
+    'hov'
   },
 
   access_tag_blacklist = Set {
     'no',
-    'private',
     'agricultural',
     'forestry',
     'emergency',
-    'psv',
-    'delivery'
+    'psv'
+  },
+
+  restricted_access_tag_list = Set {
+    'private',
+    'delivery',
+    'destination'
   },
 
   access_tags_hierarchy = Sequence {
@@ -331,7 +337,7 @@ function way_function(way, result)
     -- handle service road restrictions
     'handle_service',
 
-    -- check high occupancy vehicle restrictions
+    -- handle hov
     'handle_hov',
 
     -- compute speed taking into account way type, maxspeed tags, etc.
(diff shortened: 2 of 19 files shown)


danpat:
There's one scenario that I can think of that's very similar, but not covered by this:

Nodes: http://wiki.openstreetmap.org/wiki/Tag:barrier=toll_booth
Ways: http://wiki.openstreetmap.org/wiki/Key:toll

I think this PR could handle the `toll=yes` tagging on ways that should be added, and hooked into the `avoid` mechanism in the profile so we can turn it on/off easily (and it should default to not avoiding tolls).

Some highways *only* have `barrier=toll_booth` though - the ways are missing the toll tags.  It would be nice to try to handle these too, but that's a harder problem because it's a node tag, and maybe belongs in a separate PR.

profiles/car.lua
```diff
@@ -62,19 +62,22 @@ local profile = {
...
 
   access_tag_blacklist = Set {
     'no',
-    'private',
     'agricultural',
     'forestry',
     'emergency',
-    'psv',
-    'delivery'
+    'psv'
```
daniel-j-h: Is there a reason for having the blacklist at all now? If we never route onto the local access ways we're fine in the sense of "blacklisting" them from routing. But if users somehow end up on those ways we still want to be able to route them off of them?

profiles/car.lua
```diff
@@ -358,6 +358,7 @@ function way_function(way, result)
   Handlers.run(handlers,way,result,data,profile)
 end
 
+-- Called during edge-based-graph creation
```
daniel-j-h: Hm maybe remove impl. detail comment here, I think this might just confuse users who have no idea about the C++ side of things.

profiles/car.lua
```diff
@@ -386,5 +387,11 @@ function turn_function (turn)
     else
        turn.weight = turn.duration
     end
+    if properties.weight_name == 'routability' then
+        -- penalize turns from non-local access only segments onto local access only tags
+        if not turn.source_local_access_only and turn.target_local_access_only then
+            turn.weight = turn.weight + 3000
+        end
+    end
```
daniel-j-h: Can we symbolize both routability as well as 3000. Also why 3000?

unit_tests/library/tile.cpp
```diff
@@ -205,7 +205,7 @@ BOOST_AUTO_TEST_CASE(test_tile)
     }
 
     BOOST_CHECK_EQUAL(number_of_turn_keys, 3);
-    BOOST_CHECK_EQUAL(number_of_turns_found, 732);
+    BOOST_CHECK_EQUAL(number_of_turns_found, 763);
```
daniel-j-h: Should we remove exact checks here - every other pull request has to adapt these numbers

daniel-j-h:
Btw `SIGABRT` in some Cucumber tests on Travis is something to look into.

unit_tests/library/tile.cpp
karenzshea: yeah this is annoying

profiles/car.lua
karenzshea: This is a good point. As someone who has definitely driven onto roads I wasn't supposed to be on, routing out of these is probably a better way to handle them, rather than pretending they don't exist in the road network.

karenzshea: I'll ticket this and do it in a separate PR because it will involve changing all tests that test access of any kind.

include/extractor/extraction_way.hpp
```diff
@@ -106,6 +108,8 @@ struct ExtractionWay
     bool roundabout;
     bool circular;
     bool is_startpoint;
+    bool backward_local_access_only;
```
TheMarex: Using the term `restricted` instead of `local_access_only` might be more in line with our usual vocabulary for this. My suggestion would be `{backward,forward}_restricted`

profiles/car.lua
```diff
@@ -34,6 +34,9 @@ local profile = {
   speed_reduction            = 0.8,
   traffic_light_penalty      = 2,
   u_turn_penalty             = 20,
+  local_access_penalty       = 3000,
```
TheMarex: Possibly rename to `restriced_penalty`

include/extractor/extraction_way.hpp
TheMarex: As per slack, renamed.

PR #3724, opened 2017-02-16
Add class for storing weight of cells

TheMarex:
# Issue

This class in an integral part of the MLD approach because it stores the weights of the overlay cells and indexes them by node ids. 

## Tasklist
 - [x] add unit tests
 - [x] review
 - [x] adjust for comments

### DIFF
--- a/include/util/cell_storage.hpp
+++ b/include/util/cell_storage.hpp
@@ -1,0 +1,321 @@
+#ifndef OSRM_UTIL_CELL_STORAGE_HPP
+#define OSRM_UTIL_CELL_STORAGE_HPP
+
+#include "util/assert.hpp"
+#include "util/for_each_range.hpp"
+#include "util/multi_level_partition.hpp"
+#include "util/typedefs.hpp"
+
+#include <boost/range/iterator_range.hpp>
+#include <tbb/parallel_sort.h>
+
+#include <algorithm>
+#include <numeric>
+#include <utility>
+#include <vector>
+
+namespace osrm
+{
+namespace util
+{
+
+class CellStorage
+{
+  public:
+    using WeightOffset = std::uint32_t;
+    using BoundaryOffset = std::uint32_t;
+    using BoundarySize = std::uint32_t;
+    using SourceIndex = std::uint32_t;
+    using DestinationIndex = std::uint32_t;
+
+    static constexpr auto INVALID_WEIGHT_OFFSET = std::numeric_limits<WeightOffset>::max();
+    static constexpr auto INVALID_BOUNDARY_OFFSET = std::numeric_limits<BoundaryOffset>::max();
+
+  private:
+    struct CellData
+    {
+        WeightOffset weight_offset = INVALID_WEIGHT_OFFSET;
+        BoundaryOffset source_boundary_offset = INVALID_BOUNDARY_OFFSET;
+        BoundaryOffset destination_boundary_offset = INVALID_BOUNDARY_OFFSET;
+        BoundarySize num_source_nodes = 0;
+        BoundarySize num_destination_nodes = 0;
+    };
+
+    // Implementation of the cell view. We need a template parameter here
+    // because we need to derive a read-only and read-write view from this.
+    template <typename WeightValueT> class CellImpl
+    {
+      private:
+        using WeightPtrT = WeightValueT *;
+        using WeightRefT = WeightValueT &;
+        BoundarySize num_source_nodes;
+        BoundarySize num_destination_nodes;
+
+        WeightPtrT const weights;
+        const NodeID *const source_boundary;
+        const NodeID *const destination_boundary;
+
+        using RowIterator = WeightPtrT;
+        // Possibly replace with
+        // http://www.boost.org/doc/libs/1_55_0/libs/range/doc/html/range/reference/adaptors/reference/strided.html
+        class ColumnIterator : public std::iterator<std::random_access_iterator_tag, EdgeWeight>
+        {
+          public:
+            explicit ColumnIterator(WeightPtrT begin, std::size_t row_length)
+                : current(begin), stride(row_length)
+            {
+                BOOST_ASSERT(begin != nullptr);
+            }
+
+            WeightRefT operator*() const { return *current; }
+
+            ColumnIterator &operator++()
+            {
+                current += stride;
+                return *this;
+            }
+
+            ColumnIterator &operator+=(int amount)
+            {
+                current += stride * amount;
+                return *this;
+            }
+
+            bool operator==(const ColumnIterator &other) const { return current == other.current; }
+
+            bool operator!=(const ColumnIterator &other) const { return current != other.current; }
+
+            std::int64_t operator-(const ColumnIterator &other) const
+            {
+                return (current - other.current) / stride;
+            }
+
+          private:
+            WeightPtrT current;
+            std::size_t stride;
+        };
+
+        std::size_t GetRow(NodeID node) const
+        {
+            auto iter = std::find(source_boundary, source_boundary + num_source_nodes, node);
+            BOOST_ASSERT(iter != source_boundary + num_source_nodes);
+            return iter - source_boundary;
+        }
+
+        std::size_t GetColumn(NodeID node) const
+        {
+            auto iter =
+                std::find(destination_boundary, destination_boundary + num_destination_nodes, node);
+            BOOST_ASSERT(iter != destination_boundary + num_destination_nodes);
+            return iter - destination_boundary;
+        }
+
+      public:
+        auto GetOutWeight(NodeID node) const
+        {
+            auto row = GetRow(node);
+            auto begin = weights + num_destination_nodes * row;
+            auto end = begin + num_destination_nodes;
+            return boost::make_iterator_range(begin, end);
+        }
+
+        auto GetInWeight(NodeID node) const
+        {
+            auto column = GetColumn(node);
+            auto begin = ColumnIterator{weights + column, num_destination_nodes};
+            auto end = ColumnIterator{weights + column + num_source_nodes * num_destination_nodes,
+                                      num_destination_nodes};
+            return boost::make_iterator_range(begin, end);
+        }
+
+        auto GetSourceNodes() const
+        {
+            return boost::make_iterator_range(source_boundary, source_boundary + num_source_nodes);
+        }
+
+        auto GetDestinationNodes() const
+        {
+            return boost::make_iterator_range(destination_boundary,
+                                              destination_boundary + num_destination_nodes);
+        }
+
+        CellImpl(const CellData &data,
+                 WeightPtrT const all_weight,
+                 const NodeID *const all_sources,
+                 const NodeID *const all_destinations)
+            : num_source_nodes{data.num_source_nodes},
+              num_destination_nodes{data.num_destination_nodes},
+              weights{all_weight + data.weight_offset},
+              source_boundary{all_sources + data.source_boundary_offset},
+              destination_boundary{all_destinations + data.destination_boundary_offset}
+        {
+            BOOST_ASSERT(all_weight != nullptr);
+            BOOST_ASSERT(all_sources != nullptr);
+            BOOST_ASSERT(all_destinations != nullptr);
+        }
+    };
+
+    std::size_t LevelIDToIndex(LevelID level) const { return level - 1; }
+
+  public:
+    using Cell = CellImpl<EdgeWeight>;
+    using ConstCell = CellImpl<const EdgeWeight>;
+
+    template <typename GraphT>
+    CellStorage(const MultiLevelPartition &partition, const GraphT &base_graph)
+    {
+        // pre-allocate storge for CellData so we can have random access to it by cell id
+        unsigned number_of_cells = 0;
+        for (LevelID level = 1u; level < partition.GetNumberOfLevels(); ++level)
+        {
+            level_to_cell_offset.push_back(number_of_cells);
+            number_of_cells += partition.GetNumberOfCells(level);
+        }
+        level_to_cell_offset.push_back(number_of_cells);
+        cells.resize(number_of_cells);
+
+        std::vector<std::pair<CellID, NodeID>> level_source_boundary;
+        std::vector<std::pair<CellID, NodeID>> level_destination_boundary;
+
+        for (LevelID level = 1u; level < partition.GetNumberOfLevels(); ++level)
+        {
+            auto level_offset = level_to_cell_offset[LevelIDToIndex(level)];
+
+            level_source_boundary.clear();
+            level_destination_boundary.clear();
+
+            for (auto node = 0u; node < base_graph.GetNumberOfNodes(); ++node)
+            {
+                const CellID cell_id = partition.GetCell(level, node);
+                bool is_source_node = false;
+                bool is_destination_node = false;
+                bool is_boundary_node = false;
+
+                for (auto edge : base_graph.GetAdjacentEdgeRange(node))
+                {
+                    auto other = base_graph.GetTarget(edge);
+                    const auto &data = base_graph.GetEdgeData(edge);
+
+                    is_boundary_node |= partition.GetCell(level, other) != cell_id;
+                    is_source_node |= partition.GetCell(level, other) == cell_id && data.forward;
+                    is_destination_node |=
+                        partition.GetCell(level, other) == cell_id && data.backward;
+                }
+
+                if (is_boundary_node)
+                {
+                    if (is_source_node)
+                        level_source_boundary.emplace_back(cell_id, node);
+                    if (is_destination_node)
+                        level_destination_boundary.emplace_back(cell_id, node);
+                    // a partition that contains boundary nodes that have no arcs going into
+                    // the cells or coming out of it is invalid. These nodes should be reassigned
+                    // to a different cell.
+                    BOOST_ASSERT_MSG(
+                        is_source_node || is_destination_node,
+                        "Node needs to either have incoming or outgoing edges in cell");
+                }
+            }
+
+            tbb::parallel_sort(level_source_boundary.begin(), level_source_boundary.end());
+            tbb::parallel_sort(level_destination_boundary.begin(),
+                               level_destination_boundary.end());
+
+            const auto insert_cell_boundary = [this, level_offset](auto &boundary,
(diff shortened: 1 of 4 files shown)


include/util/cell_storage.hpp
```diff
@@ -0,0 +1,295 @@
...
+        BoundarySize num_destination_nodes = 0;
+    };
+
+    // Implementation of the cell view. We need a template parameter here
+    // because we need to derive a read-only and read-write view from this.
+    template <typename WeightPtrT> class CellImpl
+    {
+      private:
+        using WeightValueT = EdgeWeight;
+        using WeightRefT = decltype(*WeightPtrT());
```
oxidase: this is misleading, i would suggest to use explicit type traits `std::remove_pointer` and `std::add_lvalue_reference` 
But better to change `WeightPtrT` to `WeightType` and than make  pointer or reference types out of `WeightType`

include/util/cell_storage.hpp
```diff
@@ -0,0 +1,295 @@
...
+            int operator-(const ColumnIterator &other) const { return (current - other.current) / offset; }
+
+          private:
+            WeightPtrT current;
+            std::size_t offset;
+        };
+
+        std::size_t GetRow(NodeID node) const
+        {
+            return std::find(source_boundary, source_boundary + num_source_nodes, node) -
```
oxidase: shall it be asserted that `node` is in `source_boundary`?

include/util/cell_storage.hpp
```diff
@@ -0,0 +1,295 @@
...
+        WeightPtrT const weights;
+        const NodeID *const source_boundary;
+        const NodeID *const destination_boundary;
+
+        using RowIterator = WeightPtrT;
+        class ColumnIterator : public std::iterator<std::random_access_iterator_tag, EdgeWeight>
+        {
+          public:
+            explicit ColumnIterator(WeightPtrT begin, std::size_t row_length)
+                : current(begin), offset(row_length)
```
oxidase: `offset` -> `stride`. i would suggest to make later a cache-friendly storage

include/util/cell_storage.hpp
```diff
@@ -0,0 +1,295 @@
...
+
+            for (auto node = 0u; node < base_graph.GetNumberOfNodes(); ++node)
+            {
+                const CellID cell_id = partition.GetCell(level, node);
+                bool is_source_node = false;
+                bool is_destination_node = false;
+                bool is_boundary_node = false;
+
+                for (auto edge = base_graph.BeginEdges(node); edge < base_graph.EndEdges(node);
+                     ++edge)
```
oxidase: just minor: `for (auto edge : base_graph.GetAdjacentEdgeRange(node))`

include/util/cell_storage.hpp
```diff
@@ -0,0 +1,295 @@
...
+        WeightOffset weight_offset = INVALID_WEIGHT_OFFSET;
+        BoundaryOffset source_boundary_offset = INVALID_BOUNDARY_OFFSET;
+        BoundaryOffset destination_boundary_offset = INVALID_BOUNDARY_OFFSET;
+        BoundarySize num_source_nodes = 0;
+        BoundarySize num_destination_nodes = 0;
+    };
+
+    // Implementation of the cell view. We need a template parameter here
+    // because we need to derive a read-only and read-write view from this.
+    template <typename WeightPtrT> class CellImpl
```
daniel-j-h: Below: assert for `ptr` in ctor / functions, or take a ref

daniel-j-h: Also if you only want this to work with pointers and not e.g. random access iterators

    static_assert(is_pointer<T>:value, "");

include/util/cell_storage.hpp
```diff
@@ -0,0 +1,295 @@
...
+        using WeightRefT = decltype(*WeightPtrT());
+        BoundarySize num_source_nodes;
+        BoundarySize num_destination_nodes;
+
+        WeightPtrT const weights;
+        const NodeID *const source_boundary;
+        const NodeID *const destination_boundary;
+
+        using RowIterator = WeightPtrT;
+        class ColumnIterator : public std::iterator<std::random_access_iterator_tag, EdgeWeight>
```
daniel-j-h: I would just use boost's iterator facade to implement this random access iterator; makes away with a lot boilerplate and actually provides things like required typedefs. Check the osrm code base we are already using it. And see

http://www.boost.org/doc/libs/1_55_0/libs/iterator/doc/iterator_facade.html

daniel-j-h: On second thought, shouldn't a column iterator be the same as a strided iterator over your flattened matrix?

http://www.boost.org/doc/libs/1_55_0/libs/range/doc/html/range/reference/adaptors/reference/strided.html

include/util/cell_storage.hpp
```diff
@@ -0,0 +1,295 @@
...
+        }
+        std::size_t GetColumn(NodeID node) const
+        {
+            return std::find(
+                       destination_boundary, destination_boundary + num_destination_nodes, node) -
+                   destination_boundary;
+        }
+
+      public:
+        std::pair<RowIterator, RowIterator> GetOutWeight(NodeID node) const
```
daniel-j-h: Here and below: for all the pairs of iterators, what about either using

    boost:make_iterator_range(first, last)

or creating a small struct with `begin` and `end` functions so we can iterate over the return values and use std algorithms directly.

include/util/cell_storage.hpp
```diff
@@ -0,0 +1,295 @@
...
+        auto GetDestinationNodes() const
+        {
+            return std::make_pair(destination_boundary,
+                                  destination_boundary + num_destination_nodes);
+        }
+
+        CellImpl(const CellData &data,
+                 WeightPtrT const all_weight,
+                 const NodeID *const all_sources,
+                 const NodeID *const all_destinations)
```
daniel-j-h: Hm if you really want to user pointers at least assert them not being nullptrs. The interface is a bit ugly..

include/util/cell_storage.hpp
```diff
@@ -0,0 +1,295 @@
...
+            : num_source_nodes{data.num_source_nodes},
+              num_destination_nodes{data.num_destination_nodes},
+              weights{all_weight + data.weight_offset},
+              source_boundary{all_sources + data.source_boundary_offset},
+              destination_boundary{all_destinations + data.destination_boundary_offset}
+        {
+        }
+    };
+
+    inline std::size_t LevelIDToIndex(LevelID level) const { return level-1; }
```
daniel-j-h: These are all inline by default, this won't help at all

include/util/cell_storage.hpp
```diff
@@ -0,0 +1,295 @@
...
+
+  public:
+    using Cell = CellImpl<EdgeWeight *>;
+    using ConstCell = CellImpl<const EdgeWeight *>;
+
+    template <typename GraphT>
+    CellStorage(const MultiLevelPartition &partition, const GraphT &base_graph)
```
daniel-j-h: Realistically speaking, we want to use only the static graph here (?)

include/util/cell_storage.hpp
```diff
@@ -0,0 +1,295 @@
...
+        }
+        level_to_cell_offset.push_back(number_of_cells);
+        cells.resize(number_of_cells);
+
+        for (LevelID level = 1u; level < partition.GetNumberOfLevels(); ++level)
+        {
+            auto level_offset = level_to_cell_offset[LevelIDToIndex(level)];
+
+            std::vector<std::pair<CellID, NodeID>> level_source_boundary;
+            std::vector<std::pair<CellID, NodeID>> level_destination_boundary;
```
daniel-j-h: Hm this creates vectors and destroys them after each iteration again. Should be fine with only a couple of levels, but still you could hoist them outside and `.clear()` between each iteration instead.

include/util/cell_storage.hpp
```diff
@@ -0,0 +1,295 @@
...
+                    // the cells or coming out of it is invalid. These nodes should be reassigned
+                    // to a different cell.
+                    BOOST_ASSERT_MSG(
+                        is_source_node || is_destination_node,
+                        "Node needs to either have incoming or outgoing edges in cell");
+                }
+            }
+
+            std::sort(level_source_boundary.begin(), level_source_boundary.end());
+            std::sort(level_destination_boundary.begin(), level_destination_boundary.end());
```
daniel-j-h: tbb sort?

include/util/for_each_range.hpp
```diff
@@ -0,0 +1,27 @@
...
+{
+
+template <typename Iter, typename Func> void for_each_range(Iter begin, Iter end, Func f)
+{
+    auto iter = begin;
+    while (iter != end)
+    {
+        const auto key = iter->first;
+        auto begin_range = iter;
+        while (iter != end && iter->first == key)
```
daniel-j-h: Not sure if this should be a generic util function then at all. Even now it depends on the sub ranges being represented in pairs.

daniel-j-h: Also isn't this simply

    for_each(f, l, [](auto sub_f, auto sub_l) { for_each(sub_f, sub_f, functor); };

at the moment

include/util/cell_storage.hpp
```diff
@@ -0,0 +1,295 @@
...
+
+        // Set weight offsets and calculate total storage size
+        WeightOffset weight_offset = 0;
+        for (auto &cell : cells)
+        {
+            cell.weight_offset = weight_offset;
+            weight_offset += cell.num_source_nodes * cell.num_destination_nodes;
+        }
+
+        weights.resize(weight_offset+1, INVALID_EDGE_WEIGHT);
```
daniel-j-h: clang-format

include/util/cell_storage.hpp
```diff
@@ -0,0 +1,295 @@
...
+        auto offset = level_to_cell_offset[LevelIDToIndex(level)];
+        return ConstCell{
+            cells[offset + id], &weights[0], &source_boundary[0], &destination_boundary[0]};
+    }
+
+    Cell GetCell(LevelID level, CellID id)
+    {
+        auto offset = level_to_cell_offset[LevelIDToIndex(level)];
+        return Cell{cells[offset + id], &weights[0], &source_boundary[0], &destination_boundary[0]};
+    }
```
daniel-j-h: Would love some assertions in the file - especially since you subscript tons of vectors.

include/util/multi_level_partition.hpp
```diff
@@ -0,0 +1,33 @@
...
+namespace osrm
+{
+namespace util
+{
+
+using LevelID = std::uint8_t;
+using CellID = std::uint32_t;
+
+// Mock interface, can be removed when we have an actual implementation
+class MultiLevelPartition
```
daniel-j-h: What is the idea behind making this a abstract base class?

Can't this be a concept/trait like interface with concrete implementations?
We don't need runtime polymorphism here, do we?

include/util/cell_storage.hpp
```diff
@@ -0,0 +1,295 @@
...
+                bool is_boundary_node = false;
+
+                for (auto edge = base_graph.BeginEdges(node); edge < base_graph.EndEdges(node);
+                     ++edge)
+                {
+                    auto other = base_graph.GetTarget(edge);
+                    const auto &data = base_graph.GetEdgeData(edge);
+
+                    is_boundary_node |= partition.GetCell(level, other) != cell_id;
+                    is_source_node |= partition.GetCell(level, other) == cell_id && data.forward;
```
TheMarex: What we are checking here is whether this node will have a row /column in the cell that could contain anything other than `INVALID_EDGE_WEIGHT`. This is only the case if there is any path starting from 0 in the cell, that could eventually lead to another border node.

There is no path in the blue cell, leading from 1 to 0. So if you would add 0 as a destination that would be a column that all weights are set to `INVALID_EDGE_WEIGHT`. However we there is an edge going from 0 to 1 so there could be a path to another border vertex (also `1` in this case).

include/util/cell_storage.hpp
TheMarex: DynamicGraph would work as well, considering we might run this after the graph transformation in the partitioner without creating a new graph. Maybe we should add a graph-trait to solidify the interface?

include/util/multi_level_partition.hpp
TheMarex: No runtime polymorphism will be removed, see comment. This is just during implementation. I didn't want to couple building this class and the actual MultiLevelPartition.

include/util/cell_storage.hpp
daniel-j-h: The partitioner branch already has some graph traits
https://github.com/Project-OSRM/osrm-backend/blob/5ce90558ce535f02ffdcd589ffc515d0010d4220/include/util/graph_traits.hpp
mainly to allow to to SFINAE out edge `.data` functions and providing a `void` specialization.

You can add your own traits for all the functions we require in customization and then static assert for them, for sure.

include/util/cell_storage.hpp
TheMarex: Added a comment to investigate this. I need to read the source code to verify whether this will have similar performance.

TheMarex:
@oxidase @daniel-j-h can you take a look again if I addressed your concerns appropriately? Currently fighting a little bit of travis weirdness but it should be good to go after that.

include/util/cell_storage.hpp
```diff
@@ -0,0 +1,314 @@
...
+        using RowIterator = WeightPtrT;
+        // Possibly replace with
+        // http://www.boost.org/doc/libs/1_55_0/libs/range/doc/html/range/reference/adaptors/reference/strided.html
+        class ColumnIterator : public std::iterator<std::random_access_iterator_tag, EdgeWeight>
+        {
+          public:
+            explicit ColumnIterator(WeightPtrT begin, std::size_t row_length)
+                : current(begin), stride(row_length)
+            {
+            }
```
daniel-j-h: assert(begin)

include/util/cell_storage.hpp
```diff
@@ -0,0 +1,314 @@
...
+            {
+                current += stride * amount;
+                return *this;
+            }
+
+            bool operator==(const ColumnIterator &other) const { return current == other.current; }
+
+            bool operator!=(const ColumnIterator &other) const { return current != other.current; }
+
+            int operator-(const ColumnIterator &other) const
```
daniel-j-h: Why int as a return type here?

include/util/cell_storage.hpp
```diff
@@ -0,0 +1,314 @@
...
+        for (LevelID level = 1u; level < partition.GetNumberOfLevels(); ++level)
+        {
+            level_to_cell_offset.push_back(number_of_cells);
+            number_of_cells += partition.GetNumberOfCells(level);
+        }
+        level_to_cell_offset.push_back(number_of_cells);
+        cells.resize(number_of_cells);
+
+        std::vector<std::pair<CellID, NodeID>> level_source_boundary;
+        std::vector<std::pair<CellID, NodeID>> level_destination_boundary;
```
daniel-j-h: What about making this proper types instead of using `.first` and `.second` below

    struct LevelNode { CellID level; NodeID node; };

    std::vector<LevelNode> sources, sinks;

Then we can say `.level` and `.node` instead.

include/util/cell_storage.hpp
```diff
@@ -0,0 +1,314 @@
...
+                    is_destination_node |=
+                        partition.GetCell(level, other) == cell_id && data.backward;
+                }
+
+                if (is_boundary_node)
+                {
+                    if (is_source_node)
+                        level_source_boundary.emplace_back(cell_id, node);
+                    if (is_destination_node)
+                        level_destination_boundary.emplace_back(cell_id, node);
```
daniel-j-h: Hm can we guestimate the number of sources and sinks here and `.reserve` above?

include/util/cell_storage.hpp
```diff
@@ -0,0 +1,314 @@
...
+                    BOOST_ASSERT(level_offset + cell_id < cells.size());
+                    auto &cell = cells[level_offset + cell_id];
+                    cell.num_destination_nodes = std::distance(begin, end);
+                    cell.destination_boundary_offset = destination_boundary.size();
+
+                    std::transform(begin,
+                                   end,
+                                   std::back_inserter(destination_boundary),
+                                   [](const auto &cell_and_node) { return cell_and_node.second; });
+                });
```
daniel-j-h: This and above shared a large amount of code - what about abstracting over it?

include/util/cell_storage.hpp
```diff
@@ -0,0 +1,314 @@
...
+    Cell GetCell(LevelID level, CellID id)
+    {
+        const auto level_index = LevelIDToIndex(level);
+        BOOST_ASSERT(level_index < level_to_cell_offset.size());
+        const auto offset = level_to_cell_offset[level_index];
+        const auto cell_index = offset + id;
+        BOOST_ASSERT(cell_index < cells.size());
+        return Cell{
+            cells[cell_index], weights.data(), source_boundary.data(), destination_boundary.data()};
+    }
```
daniel-j-h: Hm const/non-const has the same impl. - what about const_cast'ing and implementing one wrt. the other?
Check Meyer's Effective C++ for how this can be done.

TheMarex: The problem I tried to solve here is around read-only views and read-write views (we need both). Even if a `Cell` itself can be `const` it might return non-const iterators to the backing store. All of the methods of `Cell` are always `const`, the view cannot be changed. I don't see how we could use `const` overloading there tor return appropriate iterators in this case?

include/util/cell_storage.hpp
TheMarex: Right this should probably be a signed 64bit integer. 👍

include/util/cell_storage.hpp
TheMarex: It is hard to estimate and depends heavily on the quality of the partition. I think 5% of all nodes is a good estimate for level 1, and every level above that will be much below this anyway.

include/util/cell_storage.hpp
TheMarex: For temporary types like this, I prefer using pair/tuple since they automatically derive all of the operators needed. I hope variable names such as `cell_and_node` compensate for that.

oxidase: why not in local name scope `using LevelNode = std::pair<CellID, NodeID>;`?

Issue #3727, opened 2017-02-17
Add an option to the lua scripts that will allow to ignore all conditional oneway roads. 


gardster:
There are many conditional oneway roads inside OSM data. http://overpass-turbo.eu/s/mO8
Now OSRM looks only at `oneway` tag. It is normal for routing, but when I use it for matching, it fails to process tracks in case if `oneway=yes` and `oneway:conditional=no@(some options)`. So I want to have an additional option on the extract stage to ignore all conditional restrictions, that will choose weakest restriction. `oneway=no` for oneway roads.

daniel-j-h:
cc @oxidase who worked on a tool to dump and ingest time-dependent conditional restrictions.

daniel-j-h:
This landed with https://github.com/Project-OSRM/osrm-backend/pull/3728 :sparkles:

PR #3728, opened 2017-02-17
Lua option to ignore conditional oneway restrictions.

gardster:
An option to add conditional oneway roads to the road graph as two sided roads.

# Issue
https://github.com/Project-OSRM/osrm-backend/issues/3727

## Tasklist
 - [x] review
 - [x] adjust for comments


gardster:
Ok. I added a comment.

PR #3730, opened 2017-02-17
Add class to manage partition information

### LINKED ISSUE #3552: Add interface for retrieving partition information
We need a class for #3350 that will be the interface to the query algorithm and the customize algorithm:

When talking about node and edge ID this will refer to the IDs of the Edge-Based-Graph stored in the `.ebg` file.

Some pseudo code of a possible interface:

```cpp
class MultiLevelPartition
{
// Returns the ID of the cell of `node` at `level`
CellID GetCellID(NodeID node, LevelID level);

// Returns the ID of the level in which `node` is relevant for our query
// That is the lowest level in which node is not in the same cell as start or target
LevelID GetQuerryLevelID(NodeID start, NodeID target, NodeID node);

std::size_t GetNumberOfLevels() const = 0;

std::size_t GetNumberOfCells(LevelID level) const = 0;
};
```

The data for this class will be provided by the partitioner outlined in #3204.

`CellID` can be either continuous or not. The latter enables compact packing into a `32bit` integer, the former makes the data structure to look up the actual cells more memory efficient.

TheMarex:
# Issue

Fixes #3552 by adding a class for representing the partition information that node has in a hierarchical partition.

## Tasklist
 - [x] Add sub-cell iterator
 - [x] Finish unit tests
 - [ ] review
 - [ ] adjust for comments

## Requirements / Relations

Based on #3724.

### DIFF
--- a/include/util/multi_level_partition.hpp
+++ b/include/util/multi_level_partition.hpp
@@ -1,16 +1,53 @@
 #ifndef OSRM_UTIL_MULTI_LEVEL_PARTITION_HPP
 #define OSRM_UTIL_MULTI_LEVEL_PARTITION_HPP
 
+#include "util/exception.hpp"
+#include "util/for_each_pair.hpp"
 #include "util/typedefs.hpp"
+
+#include <algorithm>
+#include <array>
+#include <climits>
+#include <cmath>
 #include <cstdint>
+#include <numeric>
+#include <vector>
+
+#include <boost/range/adaptor/reversed.hpp>
 
 namespace osrm
 {
 namespace util
 {
+namespace detail
+{
+// get the msb of an integer
+// return 0 for integers without msb
+template <typename T> std::size_t highestMSB(T value)
+{
+    static_assert(std::is_integral<T>::value, "Integer required.");
+    std::size_t msb = 0;
+    while (value > 0)
+    {
+        value >>= 1u;
+        msb++;
+    }
+    return msb;
+}
+
+#if (defined(__clang__) || defined(__GNUC__) || defined(__GNUG__)) && __x86_64__
+inline std::size_t highestMSB(std::uint64_t v)
+{
+    BOOST_ASSERT(v > 0);
+    return 63UL - __builtin_clzl(v);
+}
+#endif
+}
 
 using LevelID = std::uint8_t;
 using CellID = std::uint32_t;
+
+static constexpr CellID INVALID_CELL_ID = std::numeric_limits<CellID>::max();
 
 // Mock interface, can be removed when we have an actual implementation
 class MultiLevelPartition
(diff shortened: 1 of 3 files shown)


unit_tests/util/multi_level_partition.cpp
```diff
@@ -0,0 +1,31 @@
...
+
+BOOST_AUTO_TEST_CASE(mutable_cell_storage)
+{
+    // node:                0  1  2  3  4  5  6  7  8  9 10 11
+    std::vector<CellID> l1{{0, 0, 1, 1, 2, 2, 3, 3, 4, 4, 5, 5}};
+    std::vector<CellID> l2{{0, 0, 0, 0, 1, 1, 1, 1, 2, 2, 3, 3}};
+    std::vector<CellID> l3{{0, 0, 0, 0, 1, 1, 1, 1, 1, 1, 1, 1}};
+    std::vector<CellID> l4{{0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0}};
+    PackedMultiLevelPartition mlp{{l1, l2, l3, l4}, {6, 4, 2, 1}};
+
```
oxidase: Please add expectations check, the following fails
```
    BOOST_CHECK_EQUAL(mlp.GetCell(1, 10), 5);
    BOOST_CHECK_EQUAL(mlp.GetCell(2, 10), 3);
    BOOST_CHECK_EQUAL(mlp.GetCell(3, 10), 1);
    BOOST_CHECK_EQUAL(mlp.GetCell(4, 10), 0);
```
but may be my expectations are wrong

include/util/multi_level_partition.hpp
```diff
@@ -0,0 +1,250 @@
...
+    // Returns the level at which a `node` is relevant for a query from start to target
+    virtual LevelID GetQueryLevel(NodeID start, NodeID target, NodeID node) const = 0;
+
+    virtual std::size_t GetNumberOfLevels() const = 0;
+
+    virtual std::size_t GetNumberOfCells(LevelID level) const = 0;
+
+};
+
+class PackedMultiLevelPartition final : public MultiLevelPartition
```
oxidase: Would it make sense to check also a fixed length multi-level partitioning based on bisection IDs directly from the partitioner? Basically the question is about how and where translate bisection ids to `std::vector<std::vector<CellID>> partitions`

include/util/multi_level_partition.hpp
```diff
@@ -0,0 +1,250 @@
...
+
+namespace osrm
+{
+namespace util
+{
+namespace detail
+{
+// get the msb of an integer
+// return 0 for integers without msb
+template <typename T> std::size_t highestMSB(T value)
```
oxidase: This can be optimized via lookup tables or SWAR O(lg(N)) as here
https://graphics.stanford.edu/~seander/bithacks.html#IntegerLog
i am for
```


unsigned int v;	         // 32-bit value to find the log2 of 
register unsigned int r; // result of log2(v) will go here
register unsigned int shift;

r =     (v > 0xFFFF) << 4; v >>= r;
shift = (v > 0xFF  ) << 3; v >>= shift; r |= shift;
shift = (v > 0xF   ) << 2; v >>= shift; r |= shift;
shift = (v > 0x3   ) << 1; v >>= shift; r |= shift;
                                        r |= (v >> 1);
```

TheMarex: Adapted this for 64bit. Improves about 2x on the generic implementation.

unit_tests/util/multi_level_partition.cpp
TheMarex: Added tests, let me know if they work for you.

include/util/multi_level_partition.hpp
TheMarex: Replaced this with a single instruction, which improves 30x on the loop implementation. Yay!

include/util/multi_level_partition.hpp
```diff
@@ -1,31 +1,309 @@
...
+            cell_to_children.push_back(GetCell(level_idx + 1, permutation.back()) + 1);
+        }
+    }
+
+    std::vector<PartitionID> partition;
+    std::vector<std::size_t> level_offsets;
+    std::vector<PartitionID> level_masks;
+    std::vector<std::size_t> level_to_children_offset;
+    std::vector<CellID> cell_to_children;
+    std::array<LevelID, NUM_PARTITION_BITS> bit_to_level;
```
daniel-j-h: One Travis build is failing here - probably can be fixed with `#include <array>`
https://travis-ci.org/Project-OSRM/osrm-backend/jobs/203890071#L1928

TheMarex:
@oxidase @daniel-j-h can I mark this as reviewed?

daniel-j-h:
Yes from my side.

include/util/multi_level_partition.hpp
```diff
@@ -1,31 +1,310 @@
...
+                    last_parent_id = parent_partition[node];
+                }
+            }
+            // insert sentinel for the last cell
+            cell_to_children.push_back(GetCell(level_idx + 1, permutation.back()) + 1);
+        }
+    }
+
+    std::vector<PartitionID> partition;
+    std::vector<std::size_t> level_offsets;
```
oxidase: Here better to use fixed-size `uint8_t` with assumption of no more 255 bits per level, atm it is 63

include/util/multi_level_partition.hpp
```diff
@@ -1,31 +1,310 @@
...
+    std::vector<PartitionID> level_masks;
+    std::vector<std::size_t> level_to_children_offset;
```
oxidase: This vector must be serialized and better to use fixed-size `uint32_t`, i don't think it is possible to have totally 4bln children.

include/util/multi_level_partition.hpp
```diff
@@ -1,31 +1,310 @@
...
+        offsets.reserve(level_to_num_cells.size());
+
+        auto sum_bits = 0;
+        for (auto num_cells : level_to_num_cells)
+        {
+            // bits needed to number all contained vertexes
+            auto bits = static_cast<std::uint64_t>(std::ceil(std::log2(num_cells + 1)));
+            offsets.push_back(sum_bits);
+            sum_bits += bits;
+            BOOST_ASSERT(sum_bits < 64);
```
oxidase: Here can be non-strict condition `<=`, also `64` -> `NUM_PARTITION_BITS`

i think here is better to throw an exception with the text "`sum_bits` are required on level `offsets.size() + 1`"  on line 167. In release mode missing assertion leads to hardly detectable segmentation faults.

include/util/multi_level_partition.hpp
```diff
@@ -1,31 +1,310 @@
...
+    {
+        std::vector<PartitionID> masks;
+        masks.reserve(level_offsets.size());
+
+        util::for_each_pair(level_offsets.begin(),
+                            level_offsets.end(),
+                            [&](const auto offset, const auto next_offset) {
+                                // create mask that has `bits` ones at its LSBs.
+                                // 000011
+                                BOOST_ASSERT(offset < sizeof(PartitionID) * 8);
```
oxidase: It can be removed if `makeLevelOffsets` guarantees `offset <= NUM_PARTITION_BITS`

include/util/multi_level_partition.hpp
```diff
@@ -1,31 +1,310 @@
...
+
     virtual std::size_t GetNumberOfLevels() const = 0;
 
     virtual std::size_t GetNumberOfCells(LevelID level) const = 0;
 };
+
+class PackedMultiLevelPartition final : public MultiLevelPartition
+{
+    using PartitionID = std::uint64_t;
+    static const constexpr std::size_t NUM_PARTITION_BITS = sizeof(PartitionID) * 8;
```
oxidase: 8 -> CHAR_BITS

unit_tests/util/multi_level_partition.cpp
oxidase: It works!

TheMarex:
@oxidase addressed your PR comment, please have a look and merge if it works for you. 🙇

oxidase:
@TheMarex i have rebased the branch, adjusted `MockMLP` and changed [condition](https://github.com/Project-OSRM/osrm-backend/blob/634c7f640984bf5403e0cbeb57f587045f8809cd/include/util/multi_level_partition.hpp#L172) to throw if more than 64 bits are required (with `>=` only 63 bits can be used). If it is ok for you i will merge PR

TheMarex:
@oxidase thank you. 👍 Merging.

Issue #3350, opened 2016-11-21
Implement new tool `osrm-customize`


danpat:
OSRM is going to grow a new speedup approach - MLD.  https://github.com/Project-OSRM/osrm-backend/projects/3

We will need a new pre-processing tool - `osrm-customize`.  This tool will:

  - take the edge-based-graph
  - take the partition data generated by https://github.com/Project-OSRM/osrm-backend/issues/3205
  - import traffic data into the edge-based-graph
  - calculate an MLD overlay
  - serialize it to disk

Basically, this tool will be an alternative to `osrm-contract`, using a different speedup algorithm.

Once the data is serialized to disk, `osrm-routed` and `libosrm` will need to be adapted to load this file and [use an alternative routing algorithm](https://github.com/Project-OSRM/osrm-backend/issues/3208) to route on the different data structure.

james-gould:
Forgive me if this is the wrong place to ask but what source of traffic data will be used? Or isn't there a definite one yet?

daniel-j-h:
There is no free and open traffic provider out there yet from what we can tell - and OSM does not contain real-time traffic data. This does not mean companies don't have such datasets available :)

james-gould:
Waze might be something to look into, not sure of their position on open source contributions though :)

TheMarex:
Output of customize should be a file `.overlay` that can be ingested by #3351 containing the weights and border vertices for the overlay graph.

TheMarex:
This has landed with #3738.

PR #3738, opened 2017-02-20
MLD customization: fill weights in CellStorage

oxidase:
# Issue

PR partially addresses #3350: filling weights in CellStorage.

## Tasklist
 - [x] add a simple one-to-many Dijkstra cell customizer
 - [x] add `osrm-customize` tool
 - [x] review
 - [x] adjust for comments

## Requirements / Relations
Required PRs: #3724, #3730

### DIFF
--- a/include/customizer/cell_customizer.hpp
+++ b/include/customizer/cell_customizer.hpp
@@ -10,7 +10,7 @@
 
 namespace osrm
 {
-namespace customizer
+namespace customize
 {
 
 class CellCustomizer
--- a/include/customizer/cell_customizer.hpp
+++ b/include/customizer/cell_customizer.hpp
@@ -1,0 +1,148 @@
+#ifndef OSRM_CELLS_CUSTOMIZER_HPP
+#define OSRM_CELLS_CUSTOMIZER_HPP
+
+#include "partition/cell_storage.hpp"
+#include "partition/multi_level_partition.hpp"
+#include "util/binary_heap.hpp"
+
+#include <boost/thread/tss.hpp>
+#include <unordered_set>
+
+namespace osrm
+{
+namespace customizer
+{
+
+class CellCustomizer
+{
+
+  public:
+    CellCustomizer(const partition::MultiLevelPartition &partition) : partition(partition) {}
+
+    template <typename GraphT>
+    void Customize(const GraphT &graph,
+                   partition::CellStorage &cells,
+                   partition::LevelID level,
+                   partition::CellID id)
+    {
+        auto cell = cells.GetCell(level, id);
+        auto destinations = cell.GetDestinationNodes();
+
+        // for each source do forward search
+        for (auto source : cell.GetSourceNodes())
+        {
+            std::unordered_set<NodeID> destinations_set(destinations.begin(), destinations.end());
+            Heap heap(graph.GetNumberOfNodes());
+            heap.Insert(source, 0, {});
+
+            // explore search space
+            while (!heap.Empty() && !destinations_set.empty())
+            {
+                const NodeID node = heap.DeleteMin();
+                const EdgeWeight weight = heap.GetKey(node);
+
+                if (level == 1)
+                    RelaxNode<true>(graph, cells, heap, level, id, node, weight);
+                else
+                    RelaxNode<false>(graph, cells, heap, level, id, node, weight);
+
+                destinations_set.erase(node);
+            }
+
+            // fill a map of destination nodes to placeholder pointers
+            auto destination_iter = destinations.begin();
+            for (auto &weight : cell.GetOutWeight(source))
+            {
+                BOOST_ASSERT(destination_iter != destinations.end());
+                const auto destination = *destination_iter++;
+                weight =
+                    heap.WasInserted(destination) ? heap.GetKey(destination) : INVALID_EDGE_WEIGHT;
+            }
+        }
+    }
+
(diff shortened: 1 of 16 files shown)


include/customizer/cell_customizer.hpp
```diff
@@ -0,0 +1,148 @@
...
+                   osrm::util::LevelID level,
+                   osrm::util::CellID id)
+    {
+        auto cell = cells.GetCell(level, id);
+        auto destinations = cell.GetDestinationNodes();
+
+        // for each source do forward search
+        for (auto source : cell.GetSourceNodes())
+        {
+            std::unordered_set<NodeID> destinations_set(destinations.begin(), destinations.end());
```
TheMarex: Destination nodes should be unique, why do we need the unordered_set here?

oxidase: Here it is just a data structure with fast erase operation to stop search if all destination nodes are reached.

oxidase:
Rebased onto `mld/facade` branch. Added only unit tests,  proper routing testing would require an MLD routing plugin.

include/customizer/cell_customizer.hpp
```diff
@@ -0,0 +1,148 @@
...
+
+#include <boost/thread/tss.hpp>
+#include <unordered_set>
+
+namespace osrm
+{
+namespace customizer
+{
+
+class CellCustomizer
```
TheMarex: This class does not maintain any state right now. Do you think we could just split it into free functions?

include/partition/cell_storage.hpp
```diff
@@ -117,33 +123,27 @@ template <bool UseShareMemory> class CellStorageImpl
             std::size_t stride;
         };
 
-        std::size_t GetRow(NodeID node) const
```
TheMarex: What is the idea here behind inlining these functions below and removing the interface?

unit_tests/util/cell_customization.cpp
```diff
@@ -0,0 +1,277 @@
+#include <boost/numeric/conversion/cast.hpp>
+#include <boost/test/unit_test.hpp>
+
+#include "customizer/cell_customizer.hpp"
+#include "partition/cell_storage.hpp"
+#include "partition/multi_level_partition.hpp"
+#include "util/static_graph.hpp"
+
+#define REQUIRE_SIZE_RANGE(range, ref) BOOST_REQUIRE_EQUAL(range.size(), ref)
+#define CHECK_EQUAL_RANGE(range, ...)                                                              \
```
TheMarex: This is cool, could we move this into a common header? We could reuse this for the `CellStorage` tests.

include/customizer/cell_storage.hpp
```diff
@@ -186,24 +186,24 @@ template <bool UseShareMemory> class CellStorageImpl
     CellStorageImpl() {}
 
     template <typename GraphT, typename = std::enable_if<!UseShareMemory>>
-    CellStorageImpl(const MultiLevelPartition &partition, const GraphT &base_graph)
+    CellStorageImpl(const partition::MultiLevelPartition &partition, const GraphT &base_graph)
     {
         // pre-allocate storge for CellData so we can have random access to it by cell id
         unsigned number_of_cells = 0;
-        for (LevelID level = 1u; level < partition.GetNumberOfLevels(); ++level)
+        for (partition::LevelID level = 1u; level < partition.GetNumberOfLevels(); ++level)
```
TheMarex: Should we move `LevelID` and `CellID` to the global OSRM namespace?

include/partition/multi_level_partition.hpp
```diff
@@ -33,14 +33,13 @@ using MultiLevelPartitionView = detail::MultiLevelPartitionImpl<true>;
...
 {
 template <bool UseShareMemory>
+void read(const boost::filesystem::path &file,
+          detail::MultiLevelPartitionImpl<UseShareMemory> &mlp);
+template <bool UseShareMemory>
 void write(const boost::filesystem::path &file,
            const detail::MultiLevelPartitionImpl<UseShareMemory> &mlp);
 }
 
-namespace detail
```
TheMarex: Good catch, thank you.

src/partition/partitioner.cpp
```diff
@@ -174,14 +174,8 @@ int Partitioner::Run(const PartitionConfig &config)
     TIMER_STOP(packed_mlp);
     util::Log() << "MultiLevelPartition constructed in " << TIMER_SEC(packed_mlp) << " seconds";
 
-    TIMER_START(cell_storage);
```
TheMarex: Eventually we might want to move this back as generating this data structure takes some time (for large datasets) and we don't want it in the hot-path of `osrm-customize` runs.

Thinking long-term making `osrm-customize` eventually digest `.osrm.cells` will will enable only updating cells selectively and using an earlier customization for the rest.

oxidase: Sure, just forgot about heavy constructor. I will add a reader function for `CellStorage`

include/partition/cell_storage.hpp
oxidase: `GetOutWeight` and `GetInWeight` will return an empty range if `node`  is not a source or destination node. Row or column value in this case should be an invalid value and introduce an undefined behavior, so i moved both `find` calls and iterator checks to `GetOutWeight` and `GetInWeight`

Issue #3588, opened 2017-01-19
OSRM uses man_made=bridge lines to route?


yuryleb:
Please look at very strange route of your demo server: http://map.project-osrm.org/?z=18&center=59.923917%2C30.390326&loc=59.923916%2C30.390327&loc=59.924363%2C30.388628&hl=en&alt=0
![route through man_made bridge line](https://cloud.githubusercontent.com/assets/4529411/22125968/432a7d44-deaf-11e6-9bf8-ed909963c00f.png)

As I could see, it proposes to make U-turn on non-highway line (man_made=bridge)??? Or demo server uses obsolete/wrong OSM data in this area (Saint Petersburg, Russia) and should not be used there?

BTW destination is not "Проспект Обуховской Обороны", this way is "площадь Александра Невского" in OSM.

danpat:
This looks like bad OSM tagging.  `man_made=bridge` should be on the outline, `bridge=movable` should be on the road.

http://wiki.openstreetmap.org/wiki/Key:bridge
http://wiki.openstreetmap.org/wiki/Key:man_made

Here you can see how we route around the bridge outline:

![screen shot 2017-01-19 at 1 44 04 pm](https://cloud.githubusercontent.com/assets/1892250/22126663/753cd500-de4d-11e6-8bd2-678e69612a71.png)

![screen shot 2017-01-19 at 1 44 14 pm](https://cloud.githubusercontent.com/assets/1892250/22126667/7ba49ff4-de4d-11e6-97c3-8e457bb63cc1.png)

yuryleb:
Maybe (although it's quite strange to distribute bridge structure properties between different entities), but nevertheless routing through non-highway looks very disputable. Especially for car profile.

danpat:
Hmm, there is a more detailed description here:

https://wiki.openstreetmap.org/wiki/Tag:man_made%3Dbridge

The wiki states that `bridge=movable` should be applied to the area, so the tagging here actually does look valid.

Our `car` profile only looks for the presence of the `bridge=` tag, it doesn't require it in combination with a `highway=` tag - that might be a mistake.

http://keepright.at/ has checks for bridge tags:

> bridge-tags: Someone draws a bridge and forgets to tag it as highway... A bridge should always have one of the 'major' tags that its direct neighbour ways have. Currently these are: highway, railway, cycleway, waterway, footway, piste, aerialway, pipeline, building, via_ferrata

[...]

PR #3739, opened 2017-02-21
Set highway/route tags required but bridge optional in car profile

yuryleb:
This commit fixes issue #3588 "OSRM uses man_made=bridge lines to
route?" with exclusion of 'bridge' tag analisys from all profiles to
disable routing on bridges contours (with no highway tags)

### DIFF
--- a/features/car/bridge.feature
+++ b/features/car/bridge.feature
@@ -15,7 +15,7 @@ Feature: Car - Handle driving
         And the ways
             | nodes | highway | bridge  | bicycle |
             | abc   | primary |         |         |
-            | cde   |         | movable | yes     |
+            | cde   | primary | movable | yes     |
             | efg   | primary |         |         |
 
         When I route I should get
@@ -40,7 +40,7 @@ Feature: Car - Handle driving
         And the ways
             | nodes | highway | bridge  |
             | abc   | primary |         |
-            | cde   |         | movable |
+            | cde   | primary | movable |
             | efg   | primary |         |
 
         When I route I should get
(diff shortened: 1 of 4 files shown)


daniel-j-h:
You can't remove the bridge data we need it for moveable bridges and their capacity and duration tags

https://github.com/Project-OSRM/osrm-backend/blob/master/profiles/lib/handlers.lua#L123-L143

Also can you adapt and / or add Cucumber tests to see the change in behavior.

yuryleb:
It's very sad actually. I tried to rollback my changes and modify just profiles/lib/handlers.lua:
`-- handling movable bridges
function Handlers.handle_movables(way,result,data,profile)
  local bridge = data.bridge
  if bridge then
    local bridge_speed = profile.bridge_speeds[bridge]
    if bridge_speed and bridge_speed > 0 then
      local capacity_car = way:get_value_by_key("capacity:car")
      if capacity_car ~= 0 then
        result.forward_mode = profile.default_mode
        result.backward_mode = profile.default_mode
        local duration  = way:get_value_by_key("duration")
        if duration and durationIsValid(duration) then
          result.duration = math.max( parseDuration(duration), 1 )
        else
          result.forward_speed = bridge_speed
          result.backward_speed = bridge_speed
        end
      end
    else
      -- bridge type is not listed in bridge_speeds table => don't route
      result.forward_mode = mode.inaccessible
      result.backward_mode = mode.inaccessible
      return false
    end
  end
end`
That's to check if bridge tag value is listed in bridge_speed type and disable routing if not there. But unfortunately this has no effect. Do you see anything wrong here?

yuryleb:
Found what's wrong: [the bridge](https://www.openstreetmap.org/way/117802273#map=15/59.9256/30.3954) I tried to bypass is movable too

yuryleb:
I applied my changes to car profile only and updated all related tests. Should I do anything else to get this PR merged?

Issue #3742, opened 2017-02-22
OSRM matching returns not valid JSON


gardster:
Some requests, like:

[...]

PR #3744, opened 2017-02-22
Clamp annotation speed values

karenzshea:
# Issue

Closes https://github.com/Project-OSRM/osrm-backend/issues/3742

This clamps `speed` annotations when that value comes out as `infinite` (i.e. when duration is `0` for whatever reason) so that the JSON parsing upstream can handle the values properly.

I'm having trouble recreating the conditions for this in a cucumber test, likely because `0` durations are the result of some slippery phantom node snapping conditions at the beginning and ends of routes.

## Tasklist
 - [x] review
 - [x] adjust for comments

## Requirements / Relations
Needed for 5.6 release

### DIFF
--- a/include/engine/api/route_api.hpp
+++ b/include/engine/api/route_api.hpp
@@ -19,6 +19,7 @@
 
 #include "util/coordinate.hpp"
 #include "util/integer_range.hpp"
+#include "util/json_util.hpp"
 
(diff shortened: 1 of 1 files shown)


emiltin:
note that you can set node locations direction as lat/lon in cucumber using a 'and the nodes' table

PR #3758, opened 2017-02-28
Makes Cucumber Output Pretty

daniel-j-h:
Turns out I've never seen the pretty cucumber tests scroll by on my machine! The default format should be `pretty` but apparently it is not. I checked to make sure I'm using the bundled cucumber version.

https://github.com/cucumber/cucumber-js/blob/master/docs/cli.md#formats

This changeset hard-coded the pretty flag.

What I was seeing before for passing tests: dots.

cc @willwhite @MoKob 

```
Node Version v4.4.7
.....................................................................................................................................................................................................................
.....................................................................................................................................................................................................................
.....................................................................................................................................................................................................................
.....................................................................................................................................................................................................................
.....................................................................................................................................................................................................................
.....................................................................................................................................................................................................................
.....................................................................................................................................................................................................................
.....................................................................................................................................................................................................................
.....................................................................................................................................................................................................................
.....................................................................................................................................................................................................................
.....................................................................................................................................................................................................................
.....................................................................................................................................................................................................................
.....................................................................................................................................................................................................................
.....................................................................................................................................................................................................................
.....................................................................................................................................................................................................................
.....................................................................................................................................................................................................................
.....................................................................................................................................................................................................................
.....................................................................................................................................................................................................................
.....................................................................................................................................................................................................................
.....................................................................................................................................................................................................................
.....................................................................................................................................................................................................................
.....................................................................................................................................................................................................................
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -25,7 +25,7 @@
   },
   "scripts": {
     "lint": "eslint -c ./.eslintrc features/step_definitions/ features/support/",
-    "test": "npm run lint && node ./node_modules/cucumber/bin/cucumber.js features/ -p verify",
+    "test": "npm run lint && node ./node_modules/cucumber/bin/cucumber.js features/ -p verify --format=pretty",
     "clean-test": "rm -rf test/cache",
     "cucumber": "./node_modules/cucumber/bin/cucumber.js",
     "build-api-docs": "./scripts/build_api_docs.sh"


emiltin:
the format was chosen to avoid overly long log files at the CI server. if a test fails, it will show it the full/prettty format. when running tests locally you can use the default profile which uses the pretty format.

by the way, the "-p verify" profile is defined in cucumber.js and uses --format progress. in case we want to use the pretty format at the CI server, i think we should not combine -p verify and with -f pretty which is confusing, but just instead just use a different profile (like default)

emiltin:
in other words, when i run test locally i just do this and get the pretty format:
```
./node_modules/cucumber/bin/cucumber.js features/
```

oxidase:
another argument for dots: full output slows down tests. it is acceptable for local terminals, but running tests remotely via ssh is will be terrible. As a compromise could be a second command for npm with `--format=pretty`

daniel-j-h:
Good points - will close here.

But why is it some people seem to have the pretty format on by default (as advertised)? I know @MoKob at least does have the pretty format.

emiltin:
what command do you use to run tests locally?

daniel-j-h:
`npm test`

and

`cucumber path/to/specific.feature`

emiltin:
you could try adding "--format pretty" to the default profile in cucumber.js. if that helps. it doesn't specify any format, maybe that's why it ends up different for different people?

Issue #3754, opened 2017-02-27
restricted access roads not penalized if restriction begins non at an intersection


tyrasd:
Example:
* http://map.project-osrm.org/?z=18&center=46.462718%2C11.255539&loc=46.463673%2C11.253616&loc=46.462166%2C11.256207&hl=en&alt=0 – route correctly doesn't go through access=destination road)
* http://map.project-osrm.org/?z=18&center=46.462718%2C11.255539&loc=46.462166%2C11.256207&loc=46.463673%2C11.253616&hl=en&alt=0 – in the opposite direction route goues through access=destination road

Here's the respective osm way: https://www.openstreetmap.org/way/33182436

[Apparently](http://map.project-osrm.org/debug/#18/46.46271/11.25536) the large turn restriction value onto an access=destination road is not set, if it begins on a node where only two ways meet (i.e. not a real road intersection):

![selection_018](https://cloud.githubusercontent.com/assets/1927298/23369964/682c1a62-fd13-11e6-8bf6-64b6f9d1ab1a.png)

daniel-j-h:
You are right there should be a turn penalty from

https://www.openstreetmap.org/way/33182435

onto

https://www.openstreetmap.org/way/77531178#map=19/46.46232/11.25647

but it seems like there is not.

Here is where we dispatch to the profile's turn function and in the following add up penalties

https://github.com/Project-OSRM/osrm-backend/blob/18a50d357f508026e88b41a396fd92c30340eb0b/src/extractor/edge_based_graph_factory.cpp#L531

@karenzshea can you take a look here.

TheMarex:
The problem is this line https://github.com/Project-OSRM/osrm-backend/blob/master/profiles/car.lua#L375

For degree 2 nodes the turn type will be `NoTurn` and https://github.com/Project-OSRM/osrm-backend/blob/master/profiles/car.lua#L396-L401 will not be executed, hence not adding a turn penalty.

- [ ] https://github.com/Project-OSRM/osrm-backend/blob/master/profiles/car.lua#L396-L401 needs to be moved outside the check
- [ ] Regression test needs to be written
- [ ] Cherry-pick to 5.6 branch
- [ ] add changelog entry

TheMarex:
This was fixed by #3763.

PR #3763, opened 2017-03-01
add restricted penalty on NoTurn turns

karenzshea:
# Issue

Fix for https://github.com/Project-OSRM/osrm-backend/issues/3754, where routing still went through a restricted way when the restricted way started at a degree 2 node intersection.

## Tasklist
 - [x] review
 - [x] adjust for comments

## Requirements / Relations
Needs to be cherrypicked into a 5.6.1 branch.


daniel-j-h:
Also an unrelated test seems to fail on two Travis builds now
- https://travis-ci.org/Project-OSRM/osrm-backend/jobs/206720567#L4797
- https://travis-ci.org/Project-OSRM/osrm-backend/jobs/206720566#L5827

PR #3765, opened 2017-03-01
Integrate MLD data into DataFacade

TheMarex:
# Issue

For https://github.com/Project-OSRM/osrm-backend/pull/3738 we need to pull all the data structures up to the server to integrate into our algorithms.
@oxidase I partially cherry picked https://github.com/Project-OSRM/osrm-backend/pull/3738/commits/8204c961caff93afa0908c4c37ae3b690fad9bba and https://github.com/Project-OSRM/osrm-backend/pull/3738/commits/8204c961caff93afa0908c4c37ae3b690fad9bba and squashed them.

This also adds and algorithm to go from BisectionIDs to CellIDs with cell sizes `s` that are guaranteed `max_cell_size > s > max_cell_size * balance` where `balance` is the partition parameter.

## Tasklist
 - [x] Merge #3558 
 - [x] Rebase this branch on master
 - [x] Rebase #3738 on this branch
 - [x] review
 - [x] adjust for comments

## Requirements / Relations

Based on #3558.


oxidase:
@TheMarex i have rebased the branch to master and directly pushed some fixes.

include/util/static_graph.hpp
```diff
@@ -134,8 +134,11 @@ template <typename EdgeDataT, bool UseSharedMemory = false> class StaticGraph
     StaticGraph(typename ShM<NodeArrayEntry, UseSharedMemory>::vector &nodes,
                 typename ShM<EdgeArrayEntry, UseSharedMemory>::vector &edges)
     {
+        BOOST_ASSERT(!nodes.empty());
+
         number_of_nodes = static_cast<decltype(number_of_nodes)>(nodes.size() - 1);
-        number_of_edges = static_cast<decltype(number_of_edges)>(edges.size());
+        number_of_edges = static_cast<decltype(number_of_edges)>(nodes.back().first_edge);
+        BOOST_ASSERT(number_of_edges <= edges.size());
```
TheMarex: Why would we have more entries in the edge vector then we could reach through a node?

include/partition/edge_based_graph_reader.hpp
```diff
@@ -38,6 +38,107 @@ struct EdgeBasedGraphEdge : EdgeBasedGraph::InputEdge
...
+                              edge.data.duration,
+                              edge.data.backward,
+                              edge.data.forward);
+    }
+
+    return directed;
+}
+
+std::vector<EdgeBasedGraphEdge>
+PrepareEdgesForUsageInGraph(std::vector<extractor::EdgeBasedEdge> edges)
```
TheMarex: We use lower-case first letter for free functions.

include/partition/edge_based_graph_reader.hpp
```diff
@@ -38,6 +38,107 @@ struct EdgeBasedGraphEdge : EdgeBasedGraph::InputEdge
     using Base::Base;
 };
 
+// Bidirectional (s,t) to (s,t) and (t,s)
+std::vector<extractor::EdgeBasedEdge>
+SplitBidirectionalEdges(const std::vector<extractor::EdgeBasedEdge> &edges)
```
TheMarex: Needs to be inline or moved to `.cpp`.

include/util/static_graph.hpp
oxidase: The problem that the preallocated size in the memory block is `2 * #edges` to include duplicated edges, so shmem vector size is `2 * #edges`, but actual number of edges is the sentinel value after graph is loaded.  Some edges will be merged [here](https://github.com/Project-OSRM/osrm-backend/blob/ee076e61564a5770be6a518f1ecbe7c6daf67246/include/partition/edge_based_graph_reader.hpp#L154-L161)

src/storage/storage.cpp
```diff
@@ -840,6 +956,86 @@ void Storage::PopulateData(const DataLayout &layout, char *memory_ptr)
...
+        if (boost::filesystem::exists(config.edge_based_graph_path))
+        {
+            io::FileReader reader(config.edge_based_graph_path, io::FileReader::VerifyFingerprint);
+
+            const auto number_of_edges = reader.ReadElementCount64();
+            const auto number_of_nodes = reader.ReadOne<EdgeID>() + 1;
+            std::vector<extractor::EdgeBasedEdge> original_edges(number_of_edges);
+            reader.ReadInto(original_edges);
+
+            original_edges = partition::splitBidirectionalEdges(std::move(original_edges));
```
TheMarex: Missed this part. For the future we would want to not do this on load-time but before the graph hit's disk in `osrm-extract`. This would avoid over-allocating. But since this will need changes to how `osrm-contract` works (currently iterating over the edge list), it is fine to add a `TODO` here. Can you open a ticket to capture this?

include/util/static_graph.hpp
TheMarex: Okay, we should address this in a separate PR that also refactor functionality in `osrm-contract`.

src/storage/storage.cpp
oxidase: https://github.com/Project-OSRM/osrm-backend/issues/3783

TheMarex:
@oxidase do you think we should merge this in so it will be easier to base branches of it?

oxidase:
@TheMarex i have added a fix for MLD data reading,  otherwise LGTM!

PR #3768, opened 2017-03-02
Merges node-osrm into repo. 

daniel-j-h:
Merging node-osrm NodeJS bindings into the repository. Can be built with

    cmake .. -DENABLE_NODE_BINDINGS=On

Structured as follows
- `src/bindings/node` - recursive CMakeLists.txt and binding impl.
- `include/bindings/node` - binding headers

What's still missing
- [x] copy tbb libs to lib/bindings or use static tbb libs only for the node bindings (?)
- [x] binary publishing / caching, node-pre-gyp can we do this without the need for a Makefile?
- [x] tests
- [x] docs
- [x] cmake depends on `npm install` for node-cmake - should cmake do this on the fly?
- [x] hook up to travis
- [x] merge tests, only keep Berlin dataset, discard Monaco and adapt all tests
- [x] let travis build and publish binaries


src/bindings/node/node_osrm.cpp
```diff
@@ -0,0 +1,507 @@
+#include "osrm/engine_config.hpp"
```
daniel-j-h: This was me being sneaky and opening up the door for more bindings :)

Also give https://github.com/junegunn/fzf a try (or ctrl+p)

package.json
```diff
@@ -12,7 +12,11 @@
     "polyline": "^0.2.0",
     "request": "^2.69.0",
     "rimraf": "^2.5.4",
-    "xmlbuilder": "^4.2.1"
+    "xmlbuilder": "^4.2.1",
+
+    "nan": "^2.1.0",
+    "node-cmake": "^1.2.1",
+    "node-pre-gyp": "~0.6.30"
```
daniel-j-h: It does, a few lines below https://github.com/Project-OSRM/osrm-backend/blob/merge-node-osrm/package.json#L37-L38

test/bindings/node/data/Makefile
```diff
@@ -0,0 +1,32 @@
+OSRM_RELEASE:=$(shell node -e "console.log(require('../../../../package.json').osrm_release)")
```
daniel-j-h: Yes, requires changing all coordinates from locations in Bonaco to Berlin - I agree having multiple datasets is ugly.

src/bindings/node/node_osrm.cpp
TheMarex: Yeah but you can also have `src/python` if you want that. I don't think we will have namespace conflicts 😉

TheMarex:
> cmake depends on npm install for node-cmake - should cmake do this on the fly?

@daniel-j-h Hm, would it be possible to bundle the cmake script directly? That is what the v2 docs recommend as well: https://github.com/cjntaylor/node-cmake

daniel-j-h:
Yes, can do. Vendoring in the cmake dir.

daniel-j-h:
Merged node-osrm tests into backend; the Monaco dataset is gone now. Important: I kept the s3 files for backwards compat. and uploaded the Berlin ones and made them public.

Had to adapt all tests, but now it works. The Tile plugin is especially fragile in that regard, and caused some pain. In the end I got it working (e.g. with range checks for sizes).

The tile in Berlin does not seem to have strings in it for street names - any insights here @danpat? Should all tiles have the street names in them for streets they display?

Refs.
- http://map.project-osrm.org/debug/#18.14/52.52337/13.40302
- http://map.project-osrm.org/?z=18&center=52.523087%2C13.402623&hl=en&alt=0

daniel-j-h:
Having LTO builds enabled for this repo.'s release / production builds is a blocker for merging this PR: https://github.com/Project-OSRM/osrm-backend/pull/3770

Also this is mostly done with the exception of
- the Tile plugin behavior outlined above
- publishing binaries via node-pre-gyp, I'm missing some context how to set this up and test the full pipeline without publishing artifacts to npm while doing so. cc @TheMarex

TheMarex:
> publishing binaries via node-pre-gyp, I'm missing some context how to set this up and test the full pipeline without publishing artifacts to npm while doing so. cc @TheMarex

node-pre-gyp does not publish packages to npm, that is a manual step you would need to do. To just test binary builds try the following:

1. Make sure https://github.com/Project-OSRM/node-osrm/blob/master/scripts/publish.sh is run at the end of the travis build
2. Change `version` in the package.json to something new like `5.7.0-daniel`
3. Commit this with `[publish binary] Test version of 5.7.0 for Daniel`
4. ????
5. Profit.

danpat:
@daniel-j-h yes - there should be street name attributes for any visible street on the tile:

https://github.com/Project-OSRM/osrm-backend/pull/2488

We're not rendering those labels on the debug frontend though - are you sure they're missing from the tiles?

TheMarex:
What is blocking here? I would like to get this merged in time to integrate the algorithm selection in `node-osrm` and add new MLD toolchain.

daniel-j-h:
Here's an update:
- gcc debug builds are timing out on travis, looks unrelated but is a pita
- clang 4 linker error https://travis-ci.org/Project-OSRM/osrm-backend/jobs/212117117#L4487-L4496, also looks unrelated
- Check headers failing for nan https://travis-ci.org/Project-OSRM/osrm-backend/jobs/212117115#L2521

Otherwise building the binaries is hooked up now via node-pre-gyp.

TheMarex:
> Check headers failing for nan https://travis-ci.org/Project-OSRM/osrm-backend/jobs/212117115#L2521

You can either exclude the bindings from the header check or make the compile unit for it a fake node module.

> gcc debug builds are timing out on travis, looks unrelated but is a pita
> clang 4 linker error https://travis-ci.org/Project-OSRM/osrm-backend/jobs/212117117#L4487-L4496, also looks unrelated

Did you try rebasing on master?

daniel-j-h:
Yes it is rebased on master.

TheMarex:
> clang 4 linker error https://travis-ci.org/Project-OSRM/osrm-backend/jobs/212117117#L4487-L4496, also looks unrelated

Just realized these linker issues occur because we try to link mason binaries on a 32bit system:

https://travis-ci.org/Project-OSRM/osrm-backend/jobs/212117117#L4154

Why is mason running on that build in the first place?

daniel-j-h:
Good catch - seems like the build and publish scripts were triggering a second build. I guarded them now behind the `ENABLE_NODE_BINDINGS` env var. Let's see how this works.

.travis.yml
```diff
@@ -173,7 +173,7 @@ install:
   - export OSRM_BUILD_DIR="$(pwd)/build-osrm"
   - mkdir ${OSRM_BUILD_DIR} && pushd ${OSRM_BUILD_DIR}
   - export CC=${CCOMPILER} CXX=${CXXCOMPILER}
-  - cmake .. -DCMAKE_BUILD_TYPE=${BUILD_TYPE} -DENABLE_MASON=${ENABLE_MASON:-OFF} -DENABLE_ASSERTIONS=${ENABLE_ASSERTIONS:-OFF} -DBUILD_SHARED_LIBS=${BUILD_SHARED_LIBS:-OFF} -DENABLE_COVERAGE=${ENABLE_COVERAGE:-OFF} -DENABLE_SANITIZER=${ENABLE_SANITIZER:-OFF} -DBUILD_TOOLS=ON -DENABLE_CCACHE=ON
+  - cmake .. -DCMAKE_BUILD_TYPE=${BUILD_TYPE} -DENABLE_MASON=${ENABLE_MASON:-OFF} -DENABLE_ASSERTIONS=${ENABLE_ASSERTIONS:-OFF} -DBUILD_SHARED_LIBS=${BUILD_SHARED_LIBS:-OFF} -DENABLE_COVERAGE=${ENABLE_COVERAGE:-OFF} -DENABLE_SANITIZER=${ENABLE_SANITIZER:-OFF} -DBUILD_TOOLS=ON -DENABLE_CCACHE=ON -DENABLE_NODE_BINDINGS=${ENABLE_NODE_BINDINGS}
```
TheMarex: Needs to be `-DENABLE_NODE_BINDINGS=${ENABLE_NODE_BINDINGS:-OFF}`.

TheMarex:
Fixed the windows install and data preparation and disabled the ASAN option for the GCC build for now. 

We need to switch the dataset back to Monaco, Berlin is too big for CI (contracting it takes ages in debug mode and makes our builds time out).

Hopefully this will work and we can merge this to unblock #3846.

daniel-j-h:
Travis green! Beware, we are not yet running the NodeJs tests (tests need to be adapted to Monaco now that I changed it back from the Berlin dataset), and we have to enable coverage again.

oxidase:
```
[00:12:32] > osrm@5.7.0 install C:\projects\osrm
[00:12:32] > node-pre-gyp install --fallback-to-build=false
[00:12:32] 
[00:12:32] node-pre-gyp ERR! install error 
[00:12:32] node-pre-gyp ERR! stack Error: 403 status code downloading tarball https://mapbox-node-binary.s3.amazonaws.com/osrm/v5.7.0/Release/node-v48-win32-ia32.tar.gz
```

is there any chance to get for win32? Otherwise it should be commented in `appveyor.yml`

TheMarex:
Hrm I'm not sure why `npm` even tries to run `node-pre-gyp` on Windows. I thought this was disabled with https://github.com/Project-OSRM/osrm-backend/pull/3768/commits/decf222a4f820c2c4cf1f174aec2879a6ab8f8f8#diff-180360612c6b8c4ed830919bbb4dd459R26

TheMarex:
Remaining issue with the tests ticketed here https://github.com/Project-OSRM/osrm-backend/issues/3851 Merging to unblock #3834

Issue #3773, opened 2017-03-03
Implement own access tag default


TheMarex:
With the introduction of `access=private` the access tag handling that is defined in the OSM wiki became absolutely unmanageable. There are just too many footpaths leading into the nothingness that are tagged to be accessible for all private vehicles.

We should add an explicit table for all highway types that states what we deem accessible by default for a certain mode. Example:

|highway | motor_vehicle | foot | bicycle |
|------------|---------------------|-------|-----------|
| footway |         no           | yes  | no        |
| primary |        yes           | no   |  yes     |
| cycleway|        no           | no   | yes      |

An `access={accessibility}` tag without any further specification would only affect the modes that we deem routable by default for that type. For example setting `access=private` would result in:

|highway | motor_vehicle | foot | bicycle |
|------------|---------------------|-------|-----------|
| footway |         no           | private  | no        |
| primary |        private           | no   |  private     |
| cycleway|        no           | no   | private      |

If we encounter an explicit `access={mode}` tag we will use that to override the default value. Same with `{mode}={accessibility}`.

/cc @daniel-j-h @karenzshea @danpat @emiltin

PR #3774, opened 2017-03-03
Only allow restricted access for road of certain highway type

TheMarex:
# Issue

This resolves the issue that we were routing on a lot of footpaths and service roads that were tagged as private. This creates some problems when the access tag actually means to allow access there, we would need #3773 for that.

## Tasklist
 - [x] review
 - [x] adjust for comments


TheMarex:
As per chat @daniel-j-h gave 👍

Issue #3745, opened 2017-02-23
Penalize turns onto ways with restricted access: fees, customers


daniel-j-h:
In https://github.com/Project-OSRM/osrm-backend/pull/3721 @karenzshea implemented a penalty for turns onto restricted ways (think destination routing, hov ways, etc.). This makes sure we prefer alternative routes but we can still route you there _if you really want to go there_.

There is currently a discussion going on about zoos and other areas where customers have acces and / or you have to pay a fee for entry.

http://lists.openstreetmap.org/pipermail/talk-de/2017-February/113797.html

It would be great to support those tags (and maybe more?) in the same way.

- https://wiki.openstreetmap.org/wiki/Tag:access%3Dcustomers
- http://wiki.openstreetmap.org/wiki/Key:fee

cc @karenzshea

daniel-j-h:
@bhousel just mentioned we should do the same for [`highway=track`](http://wiki.openstreetmap.org/wiki/Tag:highway%3Dtrack).

daniel-j-h:
Same with the [toll=*](http://wiki.openstreetmap.org/wiki/Key:toll) tag (and related namespaced tags). Note there are "except" tags. Also fee tags.

TheMarex:
> highway=track

There are some implications on data size that I think are worthwhile for ignoring this road-class for now.

tyrasd:
[`highway=living_street`](https://wiki.openstreetmap.org/wiki/Tag:highway%3Dliving_street#National_specialities) does imply such a restriction in some countries, e.g. Austria.

daniel-j-h:
Adding [`amenity=weighbridge`](http://wiki.openstreetmap.org/wiki/Tag:amenity%3Dweighbridge) and similar weigh station tags.

MichalPP:
is there a way to make `profile.restricted_penalty` different for different source and target ways? do I have a complete list of tags on source and target ways?

eg penalize left furn from primary & lanes > 1 to anything is duration+30

(looks like no: https://github.com/Project-OSRM/osrm-backend/blob/master/docs/profiles.md )

daniel-j-h:
@MichalPP 

- [Here](https://github.com/Project-OSRM/osrm-backend/blob/1f69df0d892c8aa9d7694819f517aff0c01c7be4/profiles/car.lua#L418-L421) we apply the turn penalty in the turn function
[...]

PR #3777, opened 2017-03-04
Fix spelling of access=customer_s_ and add to the restricted list

TheMarex:
# Issue

Fixes the following routing bug in Tiergarten Berlin:

broken

## Tasklist
 - [ ] review
 - [ ] adjust for comments

### DIFF
--- a/profiles/car.lua
+++ b/profiles/car.lua
@@ -75,7 +75,7 @@ local profile = {
     'forestry',
     'emergency',
     'psv',
-    'customer',
+    'customers',
     'private',
     'delivery',
     'destination'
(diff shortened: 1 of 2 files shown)


emiltin:
why is customers in both the blacklist and list of restricted tags?
should this be added to bike and foot profiles?

TheMarex:
@emiltin the blacklist entry makes sure highways that use these tags need to be on `restricted_highway_whitelist` or will get removed. We added this specifically for access restricted non-car roads, that we would have deemed routable (e.g. like in this case a `foodpath`).

profiles/car.lua
```diff
@@ -84,7 +84,8 @@ local profile = {
   restricted_access_tag_list = Set {
     'private',
     'delivery',
-    'destination'
+    'destination',
+    'customers',
```
daniel-j-h: Related: https://github.com/Project-OSRM/osrm-backend/issues/3745

We should probably just add the fee tag there, too.

TheMarex: Yup we should, but would do this in a separate PR. 👍

PR #3784, opened 2017-03-07
Add MSVC flag /bigobj to CMAKE_CXX_FLAGS

mloskot:
Fix fatal error [C1128](https://msdn.microsoft.com/en-us/library/8578y171.aspx): number of sections exceeded object file format limit while compiling `src/extractor/scripting_environment_lua.cpp`

------

I noticed the issue when building OSRM with MSVC in Debug configuration only.
Is it preferred to modify the `CMAKE_CXX_FLAGS_DEBUG` only?


oxidase:
Thanks! LGTM

PR #3782, opened 2017-03-07
Add graph with efficient border edge / internal edge scan

TheMarex:
# Issue

Addresses #3779 by implementing a graph data structure based on a sorted StaticGraph.

## Tasklist
 - [x] add regression / cucumber cases (see docs/testing.md)
 - [x] review
 - [x] adjust for comments

## Requirements / Relations
Based on #3765 
Blocked by implicit edge sorting https://github.com/Project-OSRM/osrm-backend/issues/3737

Issue #3737, opened 2017-02-20
Refactor segment update code


TheMarex:
Currently we are doing updates of our edge weight in `osrm-contract` in the following two phases:

1. Parse all the CSV files and build a global lookup table of speed overrides.
2. Update each individual segment in the compressed geometry.
3. Update all edge-based-edges using the `.segment_lookup` file and `.turn_duration_penalty`, `.turn_weight_penalty`.

However since we originally implemented this there have been some structural changes that would allow us to drop using the `.segment_lookup` file.

1. Parse all the CSV files and build a global lookup table of speed overrides
2. Convert the speed/rate value to a duration/weight value using the segment distance (calculate from the coordinates)
3. Update the compressed geometry, sum up all per-segment values and populate hash map `GeometryID -> {weight,duration}`
4. Iterate over all edge based edges, use `EdgeBasedEdge::id` to lookup `GeometryID` from `OriginalEdgeData`
5. Use `GeometryID` to check for new duration/weight value, combine value with the turn penalty. If there is no new value for the `GeometryID` but we only updated the turn penalty we need to re-calculate the sum of all weight/duration values for this `GeometryID`.

Related to #3349.

/cc @danpat @oxidase @karenzshea

TheMarex:
We now only use `data.turn_id` for the indexing in the update.

PR #3785, opened 2017-03-07
Split update code into own module and refactor it

TheMarex:
# Issue

Splits out the update code into an own module in preparation to be used in the customizer as well as in the contractor. ~Aim is to implement https://github.com/Project-OSRM/osrm-backend/issues/3737~

Update: Implementing issue #3737 would require a few invasive changes that I don't want to include in this PR since it moves a lot of code. Splitting this off in a separate PR.

## Tasklist
 - [x] Split update code from osrm-contract
 - [x] Clean up updater
 - [ ] ~Implement update method based on #3737~
 - [ ] review
 - [ ] adjust for comments

## Requirements / Relations
~This is need for https://github.com/Project-OSRM/osrm-backend/pull/3782 because the current update code assumes the edges are in the same order as in the `osrm-extract` code.~


include/extractor/compressed_edge_container.hpp
```diff
@@ -58,21 +59,19 @@ class CompressedEdgeContainer
     NodeID GetLastEdgeTargetID(const EdgeID edge_id) const;
     NodeID GetLastEdgeSourceID(const EdgeID edge_id) const;
 
+    // Invalidates the internal storage
+    SegmentDataContainer ToSegmentData();
```
daniel-j-h: Urgh this is a bit weird - can you do the following in the function impl.

    void fn() {
      static bool dead = false;
      assert !dead;
      dead = true;
      return move(data);
    }

include/extractor/io.hpp
```diff
@@ -0,0 +1,62 @@
...
+
+namespace osrm
+{
+namespace extractor
+{
+namespace io
+{
+
+template <>
+void read(const boost::filesystem::path &path, SegmentDataContainer &segment_data)
```
daniel-j-h: Why is this a template (specialization)?

include/extractor/io.hpp
```diff
@@ -0,0 +1,62 @@
...
+    writer.WriteFrom(segment_data.index.data(), segment_data.index.size());
+    // FIMXE remove unnecessary senitel
+    writer.WriteElementCount32(segment_data.nodes.size());
+
+    writer.WriteElementCount32(segment_data.nodes.size());
+    writer.WriteFrom(segment_data.nodes.data(), segment_data.nodes.size());
+    writer.WriteFrom(segment_data.fwd_weights.data(), segment_data.fwd_weights.size());
+    writer.WriteFrom(segment_data.rev_weights.data(), segment_data.rev_weights.size());
+    writer.WriteFrom(segment_data.fwd_durations.data(), segment_data.fwd_durations.size());
+    writer.WriteFrom(segment_data.rev_durations.data(), segment_data.rev_durations.size());
```
daniel-j-h: Hm here and above - we should probably provide a range-based read/write function for contiguous containers. Pointer and size is such a common pattern, this all should be

    reader.ReadInto(myvec);
    writer.WriteFrom(myvec);

include/extractor/segment_data_container.hpp
```diff
@@ -0,0 +1,157 @@
...
+    auto& ForwardWeight(const DirectionalGeometryID id, const SegmentOffset offset)
+    {
+        return fwd_weights[index[id] + 1 + offset];
+    }
+    auto& ReverseWeight(const DirectionalGeometryID id, const SegmentOffset offset)
+    {
+        return rev_weights[index[id] + offset];
+    }
+    // TODO we only need this for the datasource file since it breaks this
+    // abstraction, but uses this index
```
daniel-j-h: Want to ticket al TODOs?

include/extractor/segment_data_container.hpp
```diff
@@ -0,0 +1,157 @@
...
+    Vector<NodeID> nodes;
+    Vector<EdgeWeight> fwd_weights;
+    Vector<EdgeWeight> rev_weights;
+    Vector<EdgeWeight> fwd_durations;
+    Vector<EdgeWeight> rev_durations;
+};
+}
+
+using SegmentDataView = detail::SegmentDataContainerImpl<true>;
+using SegmentDataContainer = detail::SegmentDataContainerImpl<false>;
```
daniel-j-h: Hm generating two types for this is a bit weird but I can see how it has to be done for the shm vec...

daniel-j-h: Btw what we should do (not only here) is to stick to enums and specialize based on the enums instead of true / false.

    SegmentDataContainerImpl<false> // ?? what is false? I have to look this up
    SegmentDataContainerImpl<DoNotUseSharedMemory> // ah!

include/updater/csv_source.hpp
```diff
@@ -0,0 +1,19 @@
...
+
+namespace osrm
+{
+namespace updater
+{
+namespace csv
+{
+    SegmentLookupTable readSegmentValues(const std::vector<std::string> &paths);
+    TurnLookupTable readTurnValues(const std::vector<std::string> &paths);
+}
```
daniel-j-h: Nice this allows us to easily rip out csvs and implement something else.

_hint hint_ @karenzshea

include/updater/source.hpp
```diff
@@ -0,0 +1,96 @@
...
+    }
+    bool operator==(const Segment &rhs) const
+    {
+        return std::tie(from, to) == std::tie(rhs.from, rhs.to);
+    }
+};
+
+struct SpeedSource final
+{
+    SpeedSource() : speed(0), weight(std::numeric_limits<double>::quiet_NaN()) {}
```
daniel-j-h: Why NaN here? How does this work with the rest of the pipeline?

include/updater/source.hpp
```diff
@@ -0,0 +1,96 @@
...
+    }
+    bool operator==(const Turn &rhs) const
+    {
+        return std::tie(from, via, to) == std::tie(rhs.from, rhs.via, rhs.to);
+    }
+};
+
+struct PenaltySource final
+{
+    PenaltySource() : duration(0.), weight(std::numeric_limits<double>::quiet_NaN()) {}
```
daniel-j-h: Same

include/updater/updater.hpp
```diff
@@ -0,0 +1,29 @@
...
+#include <vector>
+
+namespace osrm
+{
+namespace updater
+{
+class Updater
+{
+public:
+    Updater(const UpdaterConfig &config) :config(config) {}
```
daniel-j-h: clang-format (?)

include/updater/updater.hpp
```diff
@@ -0,0 +1,29 @@
...
+
+    EdgeID LoadAndUpdateEdgeExpandedGraph(
+        std::vector<extractor::EdgeBasedEdge> &edge_based_edge_list,
+        std::vector<EdgeWeight> &node_weights) const;
```
daniel-j-h: Hm should this be split into loading (can be re-used) and updating?

daniel-j-h: Also from looking at the interface only what is the return edge id supposed to mean? max id I guess but it's not clear from the decl.

include/updater/updater.hpp
```diff
@@ -0,0 +1,29 @@
...
+
+private:
+    UpdaterConfig config;
```
daniel-j-h: Storing by const ref? Otherwise take by value and move in. Because now a ocpy is always made. If you take a value and move in users can move in, too. No copy is made.

include/util/shared_memory_vector_wrapper.hpp
```diff
@@ -34,6 +34,7 @@ class ShMemIterator
     typedef typename base_t::reference reference;
     typedef std::random_access_iterator_tag iterator_category;
 
+    explicit ShMemIterator() : m_value(nullptr) {}
```
daniel-j-h: Hm why the default ctor now?

src/extractor/compressed_edge_container.cpp
```diff
@@ -364,5 +326,11 @@ NodeID CompressedEdgeContainer::GetLastEdgeSourceID(const EdgeID edge_id) const
     BOOST_ASSERT(bucket.size() >= 2);
     return bucket[bucket.size() - 2].node_id;
 }
+
+SegmentDataContainer CompressedEdgeContainer::ToSegmentData()
+{
+    return std::move(segment_data);
```
daniel-j-h: assert not dead

src/updater/csv_source.cpp
```diff
@@ -0,0 +1,47 @@
...
+BOOST_FUSION_ADAPT_STRUCT(osrm::updater::SpeedSource,
+                          (decltype(osrm::updater::SpeedSource::speed),
+                           speed)(decltype(osrm::updater::SpeedSource::weight), weight))
+BOOST_FUSION_ADAPT_STRUCT(osrm::updater::Turn,
+                          (decltype(osrm::updater::Turn::from),
+                           from)(decltype(osrm::updater::Turn::via),
+                                 via)(decltype(osrm::updater::Turn::to), to))
+BOOST_FUSION_ADAPT_STRUCT(osrm::updater::PenaltySource,
+                          (decltype(osrm::updater::PenaltySource::duration),
+                           duration)(decltype(osrm::updater::PenaltySource::weight), weight))
```
daniel-j-h: urgh format seems to be broken here, what about

    // clang-format off
    manually format
    // clang-format on

include/extractor/compressed_edge_container.hpp
TheMarex: Hrm I can out it in a `unique_ptr`, I guess that is what it would be for.

include/extractor/io.hpp
TheMarex: Because `SegmentDataContainer = detail::SegmentDataContainer<UseSharedMemory=false>`. This could extended to allow the shared memory version, I'm not sure this makes sense though.

include/extractor/segment_data_container.hpp
TheMarex: Captured in https://github.com/Project-OSRM/osrm-backend/issues/3797

include/extractor/segment_data_container.hpp
TheMarex: I agree, we should do this in a sweep that changes it everywhere consistently.

https://github.com/Project-OSRM/osrm-backend/issues/3799

include/updater/updater.hpp
TheMarex: Yeah I'm not done with the refactor, there will be more splits here. Just wanted to get the big move into `master`. 👍

include/util/shared_memory_vector_wrapper.hpp
TheMarex: We need this for the `boost::adapters::reverse` range.

include/updater/source.hpp
TheMarex: This is from the original code. `NaN` signifies "no value set" and means it will compute the weight from the duration value.

include/extractor/io.hpp
TheMarex: Actually we already had that for reading, added it for writing too.

danpat: We already have `FileWriter::SerializeVector()` and `FileReader::DeserializeVector()` those should probably be used here.  Maybe not the best names, but they're already used elsewhere, so we should stay consistent until we have a better interface.

include/updater/updater_config.hpp
```diff
@@ -0,0 +1,81 @@
+/*
+
+Copyright (c) 2016, Project OSRM contributors
```
oxidase: Just curious about license boilerplate, would be enough just to reference [LICENSE.txt](http://softwareengineering.stackexchange.com/a/19653)?

include/extractor/io.hpp
```diff
@@ -0,0 +1,61 @@
...
+
+template <>
+void write(const boost::filesystem::path &path, const SegmentDataContainer &segment_data)
+{
+    const auto fingerprint = storage::io::FileWriter::HasNoFingerprint;
+    storage::io::FileWriter writer{path, fingerprint};
+
+    // FIXME this _should_ just be size and the senitel below need to be removed
+    writer.WriteElementCount32(segment_data.index.size() + 1);
+    writer.WriteFrom(segment_data.index.data(), segment_data.index.size());
```
oxidase: `writer.WriteFrom(segment_data.index);`

include/extractor/io.hpp
```diff
@@ -0,0 +1,61 @@
...
+    // FIMXE remove unnecessary senitel
+    writer.WriteElementCount32(segment_data.nodes.size());
```
oxidase: why not remove it now? it will require only `segment_data.index.push_back(num_entries) // add sentinel` on L25

include/updater/updater_config.hpp
daniel-j-h: I think we removed it everywhere except for public headers we install. But even then it's questionable..


Also it's 2017 already, who wants to s/2016/2017/g ? :D

src/updater/updater.cpp
```diff
@@ -0,0 +1,640 @@
...
+    const double duration = distance_in_meters / speed_in_ms;
+    return std::max<EdgeWeight>(1, static_cast<EdgeWeight>(std::round(duration * 10.)));
+}
+
+inline EdgeWeight ConvertToWeight(double weight, double weight_multiplier, EdgeWeight duration)
+{
+    if (std::isfinite(weight))
+        return std::round(weight * weight_multiplier);
+
+    return duration == MAXIMAL_EDGE_DURATION ? INVALID_EDGE_WEIGHT
```
TheMarex: We have the problem however that we depend on this functionality right now on our traffic processing. This is not a real problem for us right, since our weights are duration based still anyway.

We should re-think the whole approach on how we ingest weight/duration data though. It is much too easy to screw this up. Can you ticket this?

include/extractor/io.hpp
TheMarex: Right, originally I didn't want to break the data format but I'm doing this with #3797 anyway.

TheMarex: Actually I found the real reason the code was added at the time: The way `CompressedEdgeContainer` constructs the index was broken, so somehow a fix for this landed in the IO layer. Fixed now. 👍

TheMarex:
@daniel-j-h @oxidase pushed some fixes to address your PR comments. Please let me know if that works for you and merge this once travis gives it a 🍏 .

src/updater/updater.cpp
oxidase: #3805

PR #3786, opened 2017-03-08
Allow build with Osmium from external location

mloskot:
If `OSMIUM_INCLUDE_DIR` is specified in CMake command line, use Osmium from that location, instead of `/third_party`.

------

A non-intrusive change in CMake configuration to let override Osmium location.


daniel-j-h:
Good idea to allow for flexibility on the user's side here.

The Tile plugin makes use of Protozero which itself is bundled with recent libosmium versions. Strictly speaking we have to bundle Protozero ourselves - but for some reason we don't. Old libosmium versions up to v2.2 won't do. Just in case you run into this.

See http://osmcode.org/libosmium/manual.html#building-libosmium

mloskot:
Thanks for the review and useful notes.

Issue #3776, opened 2017-03-03
make install not deploying all headers


mloskot:
I'm building the current master on Linux, with success.

The `make install`, however, does not deploy all headers:

```
$ ls /usr/local/include/osrm/
bearing.hpp     engine_config.hpp   match_parameters.hpp    partition             storage_config.hpp    util
contractor      exception.hpp       nearest_parameters.hpp  route_parameters.hpp  table_parameters.hpp
coordinate.hpp  extractor           osrm_fwd.hpp            status.hpp            tile_parameters.hpp
engine          json_container.hpp  osrm.hpp                storage               trip_parameters.hpp

$ ls /usr/local/include/osrm/extractor
extractor_config.hpp  extractor.hpp  travel_mode.hpp

$ grep \#include /usr/local/include/osrm/extractor/extractor.hpp 
#include "extractor/edge_based_edge.hpp"
#include "extractor/edge_based_graph_factory.hpp"
#include "extractor/extractor_config.hpp"
#include "extractor/graph_compressor.hpp"
#include "util/guidance/bearing_class.hpp"
#include "util/guidance/entry_class.hpp"
#include "util/guidance/turn_lanes.hpp"
#include "util/typedefs.hpp"
```

As you can see, `/usr/local/include/osrm/extractor/extractor.hpp` includes at least two headers which are missing:
```
#include "extractor/edge_based_edge.hpp"
#include "extractor/edge_based_graph_factory.hpp"
```

There might be more.

TheMarex:
Urg sorry the `libosrm_extractor.{a,so}` libraries were experimental, we currenly only use them to link against out internal binary tool `osrm-extract`. Maybe you can just use those?

If you need this as a library, the best course of action would be to migrate this behind a pimpl-facade like we do we `libosrm.{so,a}`. https://github.com/Project-OSRM/osrm-backend/blob/master/src/osrm/osrm.cpp

Basically:
- Add new files `src/osrm/extractor.cpp`, `include/osrm/extractor.hpp`
- `extractor.hpp` basicalliy just contains one function `osrm::extract(ExtractorConfig &config)`
- `extractor.cpp` implements that function by creating an `Extractor` object and calling `Extractor::run()`

You woul need to do the same with `osrm_contract.so` for the full toolchan.

mloskot:
@TheMarex Gotcha. Thanks for the hints.

I'll leave it to you to decide if this issue should be closed.

mloskot:
[...]

PR #3787, opened 2017-03-08
Add basic facades for Extractor and Contractor run methods

mloskot:
Based on idea suggested in comments to #3776, simplifies use of extractor and contractor as libraries.

------

Questions:

1. ~~Do I need to add tests for those trivial facades?~~
   * I've added two new tests: `extract.cpp` and `contract.cpp` with primitive tests verifying the `osrm::{extract|contract}` functions are callable. Placeholders easy to extend with more advanced use cases, if needed.
   * I've added two corresponding targets `library-{contract|extract}-tests`

2. ~~Do I need to add try-catch around `run` calls to ensure nothrow policy and plain exit codes are returned?~~
   * Answered in comments below and addressed in the PR updates.


TheMarex:
> Do I need to add try-catch around run calls to ensure nothrow policy and plain exit codes are returned?

I just checked: You can safely ignore the exit codes: We will always `return 0` there. The `int` type seems to be a left-over from a refactor. You can safely change that to `void`. All errors should be reported by exceptions.

Thanks for finding the rough spots here! 👍 

If you are up for it: There is a define that we add for the library tests now: https://github.com/Project-OSRM/osrm-backend/blob/master/unit_tests/CMakeLists.txt#L67

You could use for example `OSRM_TEST_DIR "/monaco.om.pbf"` to actually run the `extract` function without error.

Another idea for a refactor: Make `osrm-extract` and `osrm-contract` in `src/tools/` use your new interface. That way it is ensured it will keep working. :-)

mloskot:
> You can safely ignore the exit codes: (...) All errors should be reported by exceptions.

OK, I've just updated the PR.

> If you are up for it: There is a define that we add for the library tests now: https://github.com/Project-OSRM/osrm-backend/blob/master/unit_tests/CMakeLists.txt#L67

Yup, `CMakeLists.txt` update sets `OSRM_TEST_DATA_DIR` for the new test targets too.

> You could use for example OSRM_TEST_DIR "/monaco.om.pbf" to actually run the extract function without error.

That's my plan to add at least one test for successful run.

> Another idea for a refactor: Make osrm-extract and osrm-contract in src/tools/ use your new interface. 

Good idea, but I'd prefer to submit it as a separate PR (if/once this one is accepted).

mloskot:
>> You could use for example OSRM_TEST_DIR "/monaco.om.pbf" to actually run the extract function without error.
> That's my plan to add at least one test for successful run.

Added `test_extract_with_valid_config` case.

mloskot:
Cheers!

Issue #3783, opened 2017-03-07
Write preprocessed edge-based graph in osrm-extract or osrm-partition


oxidase:
Added in #3765 edge-based-graph edges loading requires [graph pre-processing](https://github.com/Project-OSRM/osrm-backend/pull/3765#discussion_r104666675) using functions `partition::splitBidirectionalEdges` and `partition::prepareEdgesForUsageInGraph`.
Loading of an EBG file should be  adjusted in `osrm-contractor`, `osrm-partitioner`, `osrm-customizer` and `osrm-storage`. 

The preprocessing should be moved into `Extractor::WriteEdgeBasedGraph` or in `osrm-partitioner`:
* saving in `Extractor::WriteEdgeBasedGraph` has downside that `osrm-contractor` uses only forward edges
* saving in `osrm-partitioner` has downside that two EBG files must be created, one for `osrm-contractor` and another for MLD tools

daniel-j-h:
Related: check out the graph contractor doing the same here

https://github.com/Project-OSRM/osrm-backend/blob/master/src/contractor/graph_contractor.cpp

and before we split edges here

https://github.com/Project-OSRM/osrm-backend/blob/master/include/contractor/graph_contractor_adaptors.hpp

getting called here on the edge list

https://github.com/Project-OSRM/osrm-backend/blob/264cec12e9344599ec4edfe77d4724858e2391c9/src/contractor/contractor.cpp#L395

TheMarex:
Currently we handle this by writing out an own graph `mldgr` after `osrm-customize`. This works toolchain wise, but adds unnecessary overhead for IO.

I have been thinking a little bit on how to address this and I think the most efficient way would be to save the graph as adjacency array (StaticGraph like encoding) at the end of `osrm-extract`. There are a few things this needs to be able to works:

1. We need unified graph loading/serialization for `osrm-extract`, `osrm-contract`, `osrm-partition` and `osrm-customize`, `osrm-routed`. All need the same graph but with slight variations on the data structure:
    - `extractor` only needs to write, does not actually use any queries on the EBG (yet)
    - `contractor` needs to load, update the weight and then _modify_ the graph. So we need a way to load the graph from the file and create a `DynamicGraph` for `GraphContractor` in an efficient way. `GraphContractor` creates a list of edges again and then creates a graph again that goes on disk: this logic should be refactored because creating a graph from a list of edges is already handled by `StaticGraph` and `DynamicGraph`.
[...]

Issue #3764, opened 2017-03-01
Replace BinaryHeap implementation


TheMarex:
Our current binary heap implementation is tuned towards not using a lot of memory but not being very efficient for each heap operation.

This is fine with CH since the search space is so small it hardly matters.This was already not the case for `CoreCH` anymore, but okay for our use-case.

With MLD we will increase the heap operations dramatically, so we will need to fine-tune this. A few ideas:

1. Eat the cost of saving 64bit per node per thread (potentially multiple GB) for the heaps to maintain NodeID -> HeapIndex lookup
2. Investigate 4-ary-Heap instead of binary heap
3. Investigate if the current heap architecture of saving all data makes sense, or if we should go for explicit parent and cost vectors.

daniel-j-h:
Related: https://github.com/Project-OSRM/osrm-backend/issues/1765 want to close this, then?

TheMarex:
@oxidase profiles our usage patterns and found excessive use of `std::unordered_map<>::find()` for the NodeID -> HeapIndex translation. There are a few ways to migrate this:

1. Use an array storage for the lookup (a lot of memory)
2. Use a hash-map that is not pointer based (but linear-probing) like https://github.com/sparsehash/sparsehash

daniel-j-h:
For 2 - libosmium should already bundle various hash map impls.

TheMarex:
Random datapoint: Just switching from `UnorderedMapStorage` to `ArrayStorage` gives a speedup of `2.02` on MLD queries.

oxidase:
Fixed via #3931

Issue #3759, opened 2017-02-28
Implement MLD path unpacking


TheMarex:
Breaking out of https://github.com/Project-OSRM/osrm-backend/issues/3351:

High-Level Unpacking algorithm:
- Top-Down: For a pair of border vertices `i` and `j` at level `l` find shortest path at level `l-1` only using the cell matrices of `l-1`.
- Add resulting edges of the shortest path to a stack
- Unpack pair from the queue until we hit level 0 at which we need to do a search on the base graph
- Assemble resulting route on the base graph and return result as list of Edge-Based-Edge ids

Implementation brain dump:

- Vector of NodeIDs containing the final path
- Stack that contains `std::tuple<LevelID, NodeID, NodeID>`
- Initialization: Put unpacked path on stack in reverse order as edge pairs. The level of each edge is `min(query_level(s, t, edge_start_node), query_level(s, t, edge_target_node))`
- Take an arc from the stack check if its a boundary arc:
   - If yes it is already unpacked and the target node  to the final path.
   - If no it is a clique arc, we need to do the actual clique edge unpacking work.

Unpacking an edge:
  - Goal: Replace the clique edge on level `l` in cell `C` through a list of clique edges and border edges on level `l-1`.
  - For that we need to execute a bidirectional dijkstra search in the graph implied by the clique arcs of all subcells of `C` and their boundary edges
   - For that we need a custom dijkstra implementation like here: https://github.com/Project-OSRM/osrm-backend/pull/3738/files#diff-756eb99f3ffc32006906e2d9c2455968R93
   - Add all edges of the found path back to the stack in reverse order

/cc @oxidase

oxidase:
I have implemented path unpacking for customization checks in mld/customization branch for 1-to-1 compare with respect to CH algorithm. At the moment performance comparison is not possible because contractor in the branch skips contraction. 
Here some visual representation of MLD search spaces (inserted ebg nodes, green->cyan is the forward heap, read->violet is the reverse heap, yellow node is the middle point):
* Munich -> Stuttgart https://gist.github.com/oxidase/c0a7e4657c32660dcc87d5357020badb
  MLD routing 0.038239 seconds, unpacking 0.007025 seconds
* Munich -> Ingolstadt https://gist.github.com/oxidase/5203ef1840f68d46a313cc6515841e5a
  MLD routing 0.0173 seconds, unpacking 0.006534 seconds

There are at least two possibilities for performance improvement:
[...]

PR #3788, opened 2017-03-08
MLD shortest direct path search

oxidase:
# Issue

PR solves #3759 by adding MLD `search` and `routeStep` functions.

Some preliminary results for germany-latest:
* CH contraction: 764.415 sec
* CoreCH95%: 92.8731 sec
* MLD customization for 128 max-cell size: 32.8147 seconds

Query times for  bayern-latest  München -> Würzburg :
* CH 2-2.4ms
* CoreCH95% 49-54ms
* MLD 11-13ms
[Here](https://gist.github.com/oxidase/92f4229f9ee26634a1d601a97d51083f) is the search space for MLD. GeoJSON annotations show level and number of visited/inserted nodes into heap, greenish is the forward heap, reddish is the reverse one.

There are some discussion points:
* `osrm::engine::routing_algorithms::unpackPath` is hardcoded to CH unpacking and it is not so trivial to decouple unpacking procedure. Some possibilities:
  * pass an unpacking function by reference (it will require complete function template arguments resolution on the caller side)
  * create a new `annotatePath` function that will take a list of unpacked edges

* different member names in `EdgeData` like `QueryEdge::EdgeData::id`, `EdgeBasedEdge::EdgeData::edge_id` -> it must be unified for duck-typing (`turn_id` ?)

* atm `routing_base.[ch]pp` files are CH-aware, should it be split into algorithm-aware `ch_routing_base.[ch]pp`, `mld_routing_base.[ch]pp` and algorithm-insensitive `routing_base.[ch]pp`?

* no tests: would it make sense to run cucumber tests in different jobs with `-a CH` and `-a MLD` or tests must be for both CH and MLD?

## Tasklist
 - [x] clarify open points
 - [x] adjust implementation
 - [x] add regression / cucumber cases?
 - [x] review
 - [x] adjust for comments

## Requirements / Relations
* #3738 is required
* #3783 must be merged before the PR
* #3764 would be nice to have it before merging

### DIFF
--- a/include/engine/routing_algorithms/routing_base_ch.hpp
+++ b/include/engine/routing_algorithms/routing_base_ch.hpp
@@ -1,0 +1,475 @@
+#ifndef OSRM_ENGINE_ROUTING_BASE_CH_HPP
+#define OSRM_ENGINE_ROUTING_BASE_CH_HPP
+
+#include "extractor/guidance/turn_instruction.hpp"
+
+#include "engine/algorithm.hpp"
+#include "engine/datafacade/contiguous_internalmem_datafacade.hpp"
+#include "engine/internal_route_result.hpp"
+#include "engine/routing_algorithms/routing_base.hpp"
+#include "engine/search_engine_data.hpp"
+
+#include "util/coordinate_calculation.hpp"
+#include "util/guidance/turn_bearing.hpp"
+#include "util/typedefs.hpp"
+
+#include <boost/assert.hpp>
+
+#include <cstddef>
+#include <cstdint>
+
+#include <algorithm>
+#include <functional>
+#include <iterator>
+#include <memory>
+#include <numeric>
+#include <stack>
+#include <utility>
+#include <vector>
+
+namespace osrm
+{
+namespace engine
+{
+
+namespace routing_algorithms
+{
+
+namespace ch
+{
+
+// Stalling
+template <bool DIRECTION, typename HeapT>
+bool stallAtNode(const datafacade::ContiguousInternalMemoryDataFacade<algorithm::CH> &facade,
+                 const NodeID node,
+                 const EdgeWeight weight,
+                 const HeapT &query_heap)
+{
+    for (auto edge : facade.GetAdjacentEdgeRange(node))
+    {
+        const auto &data = facade.GetEdgeData(edge);
+        if (DIRECTION == REVERSE_DIRECTION ? data.forward : data.backward)
+        {
+            const NodeID to = facade.GetTarget(edge);
+            const EdgeWeight edge_weight = data.weight;
+            BOOST_ASSERT_MSG(edge_weight > 0, "edge_weight invalid");
+            if (query_heap.WasInserted(to))
+            {
+                if (query_heap.GetKey(to) + edge_weight < weight)
+                {
+                    return true;
+                }
+            }
+        }
+    }
+    return false;
+}
+
+template <bool DIRECTION>
+void relaxOutgoingEdges(const datafacade::ContiguousInternalMemoryDataFacade<algorithm::CH> &facade,
+                        const NodeID node,
+                        const EdgeWeight weight,
+                        SearchEngineData::QueryHeap &heap)
+{
+    for (const auto edge : facade.GetAdjacentEdgeRange(node))
+    {
+        const auto &data = facade.GetEdgeData(edge);
+        if (DIRECTION == FORWARD_DIRECTION ? data.forward : data.backward)
+        {
+            const NodeID to = facade.GetTarget(edge);
+            const EdgeWeight edge_weight = data.weight;
+
+            BOOST_ASSERT_MSG(edge_weight > 0, "edge_weight invalid");
+            const EdgeWeight to_weight = weight + edge_weight;
+
+            // New Node discovered -> Add to Heap + Node Info Storage
+            if (!heap.WasInserted(to))
+            {
+                heap.Insert(to, to_weight, node);
+            }
+            // Found a shorter Path -> Update weight
+            else if (to_weight < heap.GetKey(to))
+            {
+                // new parent
+                heap.GetData(to).parent = node;
+                heap.DecreaseKey(to, to_weight);
+            }
+        }
+    }
+}
+
+/*
+min_edge_offset is needed in case we use multiple
+nodes as start/target nodes with different (even negative) offsets.
+In that case the termination criterion is not correct
+anymore.
+
+Example:
+forward heap: a(-100), b(0),
+reverse heap: c(0), d(100)
+
+a --- d
+  \ /
+  / \
+b --- c
+
+This is equivalent to running a bi-directional Dijkstra on the following graph:
+
+    a --- d
+   /  \ /  \
+  y    x    z
+   \  / \  /
+    b --- c
+
+The graph is constructed by inserting nodes y and z that are connected to the initial nodes
+using edges (y, a) with weight -100, (y, b) with weight 0 and,
+(d, z) with weight 100, (c, z) with weight 0 corresponding.
+Since we are dealing with a graph that contains _negative_ edges,
+we need to add an offset to the termination criterion.
+*/
+static constexpr bool ENABLE_STALLING = true;
+static constexpr bool DISABLE_STALLING = false;
+template <bool DIRECTION, bool STALLING = ENABLE_STALLING>
+void routingStep(const datafacade::ContiguousInternalMemoryDataFacade<algorithm::CH> &facade,
+                 SearchEngineData::QueryHeap &forward_heap,
+                 SearchEngineData::QueryHeap &reverse_heap,
+                 NodeID &middle_node_id,
+                 EdgeWeight &upper_bound,
+                 EdgeWeight min_edge_offset,
+                 const bool force_loop_forward,
+                 const bool force_loop_reverse)
+{
+    const NodeID node = forward_heap.DeleteMin();
+    const EdgeWeight weight = forward_heap.GetKey(node);
+
+    if (reverse_heap.WasInserted(node))
+    {
+        const EdgeWeight new_weight = reverse_heap.GetKey(node) + weight;
+        if (new_weight < upper_bound)
+        {
+            // if loops are forced, they are so at the source
+            if ((force_loop_forward && forward_heap.GetData(node).parent == node) ||
+                (force_loop_reverse && reverse_heap.GetData(node).parent == node) ||
+                // in this case we are looking at a bi-directional way where the source
+                // and target phantom are on the same edge based node
+                new_weight < 0)
+            {
+                // check whether there is a loop present at the node
+                for (const auto edge : facade.GetAdjacentEdgeRange(node))
+                {
+                    const auto &data = facade.GetEdgeData(edge);
+                    if (DIRECTION == FORWARD_DIRECTION ? data.forward : data.backward)
+                    {
+                        const NodeID to = facade.GetTarget(edge);
+                        if (to == node)
+                        {
+                            const EdgeWeight edge_weight = data.weight;
+                            const EdgeWeight loop_weight = new_weight + edge_weight;
+                            if (loop_weight >= 0 && loop_weight < upper_bound)
+                            {
+                                middle_node_id = node;
+                                upper_bound = loop_weight;
+                            }
+                        }
+                    }
+                }
+            }
+            else
+            {
+                BOOST_ASSERT(new_weight >= 0);
+
+                middle_node_id = node;
+                upper_bound = new_weight;
+            }
+        }
+    }
+
+    // make sure we don't terminate too early if we initialize the weight
+    // for the nodes in the forward heap with the forward/reverse offset
+    BOOST_ASSERT(min_edge_offset <= 0);
+    if (weight + min_edge_offset > upper_bound)
+    {
+        forward_heap.DeleteAll();
+        return;
+    }
+
+    // Stalling
+    if (STALLING && stallAtNode<DIRECTION>(facade, node, weight, forward_heap))
+    {
+        return;
+    }
+
+    relaxOutgoingEdges<DIRECTION>(facade, node, weight, forward_heap);
+}
+
+template <bool UseDuration>
+EdgeWeight
+getLoopWeight(const datafacade::ContiguousInternalMemoryDataFacade<algorithm::CH> &facade,
+              NodeID node)
+{
+    EdgeWeight loop_weight = UseDuration ? MAXIMAL_EDGE_DURATION : INVALID_EDGE_WEIGHT;
+    for (auto edge : facade.GetAdjacentEdgeRange(node))
+    {
+        const auto &data = facade.GetEdgeData(edge);
+        if (data.forward)
+        {
+            const NodeID to = facade.GetTarget(edge);
+            if (to == node)
+            {
+                const auto value = UseDuration ? data.duration : data.weight;
+                loop_weight = std::min(loop_weight, value);
+            }
+        }
+    }
+    return loop_weight;
+}
+
+/**
+ * Given a sequence of connected `NodeID`s in the CH graph, performs a depth-first unpacking of
+ * the shortcut
+ * edges.  For every "original" edge found, it calls the `callback` with the two NodeIDs for the
+ * edge, and the EdgeData
+ * for that edge.
+ *
+ * The primary purpose of this unpacking is to expand a path through the CH into the original
+ * route through the
+ * pre-contracted graph.
+ *
+ * Because of the depth-first-search, the `callback` will effectively be called in sequence for
+ * the original route
+ * from beginning to end.
+ *
+ * @param packed_path_begin iterator pointing to the start of the NodeID list
+ * @param packed_path_end iterator pointing to the end of the NodeID list
+ * @param callback void(const std::pair<NodeID, NodeID>, const EdgeID &) called for each
+ * original edge found.
+ */
+template <typename BidirectionalIterator, typename Callback>
+void unpackPath(const datafacade::ContiguousInternalMemoryDataFacade<algorithm::CH> &facade,
+                BidirectionalIterator packed_path_begin,
+                BidirectionalIterator packed_path_end,
+                Callback &&callback)
+{
+    // make sure we have at least something to unpack
+    if (packed_path_begin == packed_path_end)
+        return;
+
+    std::stack<std::pair<NodeID, NodeID>> recursion_stack;
+
+    // We have to push the path in reverse order onto the stack because it's LIFO.
+    for (auto current = std::prev(packed_path_end); current != packed_path_begin;
+         current = std::prev(current))
+    {
+        recursion_stack.emplace(*std::prev(current), *current);
+    }
+
+    std::pair<NodeID, NodeID> edge;
+    while (!recursion_stack.empty())
+    {
+        edge = recursion_stack.top();
+        recursion_stack.pop();
+
+        // Look for an edge on the forward CH graph (.forward)
+        EdgeID smaller_edge_id = facade.FindSmallestEdge(
+            edge.first, edge.second, [](const auto &data) { return data.forward; });
+
+        // If we didn't find one there, the we might be looking at a part of the path that
+        // was found using the backward search.  Here, we flip the node order (.second, .first)
+        // and only consider edges with the `.backward` flag.
+        if (SPECIAL_EDGEID == smaller_edge_id)
+        {
+            smaller_edge_id = facade.FindSmallestEdge(
+                edge.second, edge.first, [](const auto &data) { return data.backward; });
+        }
+
+        // If we didn't find anything *still*, then something is broken and someone has
+        // called this function with bad values.
+        BOOST_ASSERT_MSG(smaller_edge_id != SPECIAL_EDGEID, "Invalid smaller edge ID");
+
+        const auto &data = facade.GetEdgeData(smaller_edge_id);
+        BOOST_ASSERT_MSG(data.weight != std::numeric_limits<EdgeWeight>::max(),
+                         "edge weight invalid");
+
+        // If the edge is a shortcut, we need to add the two halfs to the stack.
+        if (data.shortcut)
+        { // unpack
+            const NodeID middle_node_id = data.turn_id;
+            // Note the order here - we're adding these to a stack, so we
+            // want the first->middle to get visited before middle->second
+            recursion_stack.emplace(middle_node_id, edge.second);
+            recursion_stack.emplace(edge.first, middle_node_id);
+        }
+        else
+        {
+            // We found an original edge, call our callback.
+            std::forward<Callback>(callback)(edge, smaller_edge_id);
+        }
+    }
+}
+
+template <typename RandomIter, typename FacadeT>
+void unpackPath(const FacadeT &facade,
+                RandomIter packed_path_begin,
+                RandomIter packed_path_end,
+                const PhantomNodes &phantom_nodes,
+                std::vector<PathData> &unpacked_path)
+{
+    const auto nodes_number = std::distance(packed_path_begin, packed_path_end);
+    BOOST_ASSERT(nodes_number > 0);
+
+    std::vector<EdgeID> unpacked_edges;
+
+    auto source_node = *packed_path_begin, target_node = *packed_path_begin;
+    if (nodes_number > 1)
+    {
+        target_node = *std::prev(packed_path_end);
+        unpacked_edges.reserve(std::distance(packed_path_begin, packed_path_end));
+        unpackPath(
+            facade,
+            packed_path_begin,
+            packed_path_end,
+            [&facade, &unpacked_edges](std::pair<NodeID, NodeID> & /* edge */,
+                                       const auto &edge_id) { unpacked_edges.push_back(edge_id); });
+    }
+
+    annotatePath(facade, source_node, target_node, unpacked_edges, phantom_nodes, unpacked_path);
+}
+
+/**
+ * Unpacks a single edge (NodeID->NodeID) from the CH graph down to it's original non-shortcut
+ * route.
+ * @param from the node the CH edge starts at
+ * @param to the node the CH edge finishes at
+ * @param unpacked_path the sequence of original NodeIDs that make up the expanded CH edge
+ */
+void unpackEdge(const datafacade::ContiguousInternalMemoryDataFacade<algorithm::CH> &facade,
+                const NodeID from,
+                const NodeID to,
+                std::vector<NodeID> &unpacked_path);
+
+void retrievePackedPathFromHeap(const SearchEngineData::QueryHeap &forward_heap,
+                                const SearchEngineData::QueryHeap &reverse_heap,
+                                const NodeID middle_node_id,
+                                std::vector<NodeID> &packed_path);
+
+void retrievePackedPathFromSingleHeap(const SearchEngineData::QueryHeap &search_heap,
+                                      const NodeID middle_node_id,
+                                      std::vector<NodeID> &packed_path);
+
+// assumes that heaps are already setup correctly.
+// ATTENTION: This only works if no additional offset is supplied next to the Phantom Node
+// Offsets.
+// In case additional offsets are supplied, you might have to force a loop first.
+// A forced loop might be necessary, if source and target are on the same segment.
+// If this is the case and the offsets of the respective direction are larger for the source
+// than the target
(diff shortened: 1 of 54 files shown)


TheMarex:
> create a new annotatePath function that will take a list of unpacked edges

Yes please. I have been meaning to do this refactor for ages. ❤️ 

> different member names in EdgeData like QueryEdge::EdgeData::id, EdgeBasedEdge::EdgeData::edge_id -> it must be unified for duck-typing (turn_id ?)

Yes, `turn_id` sounds good.

> atm routing_base.[ch]pp files are CH-aware, should it be split into algorithm-aware ch_routing_base.[ch]pp, mld_routing_base.[ch]pp and algorithm-insensitive routing_base.[ch]pp

Yeah there are two options here:
- Sub-folders with matching name-spaces `routing_algorithms/ch/shortest_path.{c,h}pp` and then alias headers in `routing_algorithms/shortest_path.hpp` which pull functions form `routing_algorithms::ch` in the `routing_algorithms` namespace so the function overloading in `engine/routing_algorithms.hpp` works.
- Just post-fixes to the file names like you suggested. `shortest_path_ch.{c,h}pp`

> no tests: would it make sense to run cucumber tests in different jobs with -a CH and -a MLD or tests must be for both CH and MLD?

Running the complete tool-chain two times on each job seems a little overkill. Maybe we can add an own travis job that runs the MLD toolchain with the tests? There are definitely some that won't work on MLD since we haven't implemented the algorithms yet. We could tag these with `@todo_mld` in cucumber to get the tests passing.

oxidase:
Adjusted branch to comments and added `mld`  cucumber profile that excludes tests with the tags `@match`, `@alternative`, `@traffic`, `@matrix`, `@trip`, `@via`.
Running 907 scenarios with `-a MLD` takes only 24 seconds for linux and 2 minutes for OSX builds. I would like to keep running `osrm-extract`, `osrm-contract`, `osrm-partition` and `osrm-customize` in all jobs to have regressions checks in the MLD toolchain.

include/engine/search_engine_data.hpp
```diff
@@ -37,13 +53,23 @@ struct SearchEngineData
...
+    using MultiLayerDijkstraHeapPtr = boost::thread_specific_ptr<MultiLayerDijkstraHeap>;
+
     static SearchEngineHeapPtr forward_heap_1;
     static SearchEngineHeapPtr reverse_heap_1;
     static SearchEngineHeapPtr forward_heap_2;
     static SearchEngineHeapPtr reverse_heap_2;
     static SearchEngineHeapPtr forward_heap_3;
     static SearchEngineHeapPtr reverse_heap_3;
     static ManyToManyHeapPtr many_to_many_heap;
+    static MultiLayerDijkstraHeapPtr mld_forward_heap;
```
TheMarex: We should work towards making MLD use the same heaps that the CH variants use.

If this is generally not possible, we could also type-tag the `SearchEngineData` and use the CH heap data for `SeachEngineData<CH>` and the MLD versions for `SearchEngineData<MLD>`.

Only algorithm dependent part that I think would really be needed would be `bool HeapData::from_clique_arc`.

include/engine/search_engine_data.hpp
```diff
@@ -23,6 +24,21 @@ struct ManyToManyHeapData : HeapData
     ManyToManyHeapData(NodeID p, EdgeWeight duration) : HeapData(p), duration(duration) {}
 };
 
+struct MultiLayerDijkstraHeapData : HeapData
```
TheMarex: We can type-tag this if we need to have it algorithm dependent. (see below)

include/engine/routing_algorithms/routing_base_mld.hpp
```diff
@@ -0,0 +1,283 @@
...
+{
+namespace routing_algorithms
+{
+namespace mld
+{
+
+template <bool DIRECTION>
+void routingStep(const datafacade::ContiguousInternalMemoryDataFacade<algorithm::MLD> &facade,
+                 const partition::MultiLevelPartitionView &partition,
+                 const partition::CellStorageView &cells,
```
TheMarex: We can add an closure here that makes it easy to compute the query level for a node:

```
struct QueryLevel
{
   const MultiLevelPartition &mlp;
   const PhantomNode &source;
   const PhantomNode &target;
   LevelID operator()(NodeID node)
  {
    auto forward_level = mlp.GetQueryLevel(source.foward_node, target.forward_node, node);
    auto backward_level = mlp.GetQueryLevel(source.backward_node, target.backward_node, node);
    return std::min(forward_level, backward_level);
  }
};
```

If we can fix out partition to ensure that the level 1 cell for `forward_node` and `backward_node` is always the same, we can optimize to only compute one.

include/engine/routing_algorithms/routing_base_ch.hpp
```diff
@@ -0,0 +1,458 @@
...
+        else
+        {
+            // We found an original edge, call our callback.
+            std::forward<Callback>(callback)(edge, smaller_edge_id);
+        }
+    }
+}
+
+template <typename RandomIter, typename FacadeT>
+void unpackPath(const FacadeT &facade,
```
oxidase: `unpackPath` is only needed for CH and CoreCH. Same unpacking, but different facade types

include/partition/edge_based_graph_reader.hpp
```diff
@@ -114,16 +114,16 @@ prepareEdgesForUsageInGraph(std::vector<extractor::EdgeBasedEdge> edges)
...
-        {
-            if ((int)forward_edge.data.weight != INVALID_EDGE_WEIGHT)
-            {
-                forward_edge.data.backward = true;
-                graph_edges.push_back(forward_edge);
-            }
-        }
-        else
+        // // merge edges (s,t) and (t,s) into bidirectional edge
+        // if (forward_edge.data.weight == reverse_edge.data.weight)
```
oxidase: The problem is that some backward edges will be missing in GetAdjacentEdgeRange

include/partition/cell_storage.hpp
```diff
@@ -331,8 +331,10 @@ template <bool UseShareMemory> class CellStorageImpl
         const auto offset = level_to_cell_offset[level_index];
         const auto cell_index = offset + id;
         BOOST_ASSERT(cell_index < cells.size());
-        return ConstCell{
-            cells[cell_index], weights.data(), source_boundary.data(), destination_boundary.data()};
+        return ConstCell{cells[cell_index],
+                         weights.data(),
+                         source_boundary.empty() ? nullptr : source_boundary.data(),
```
oxidase: All tests that have two  nodes like "a b" will have no source and boundary nodes on all levels. `.data()` or referencing `&..[0]` will an undefined behavior without explicit check for `.empty()`

include/engine/routing_algorithms/routing_base_mld.hpp
```diff
@@ -0,0 +1,283 @@
...
+    unpacked_path.reserve(packed_path.size());
+    for (auto &packed_edge : packed_path)
+    {
+        LevelID level;
+        NodeID source, target;
+        EdgeID edge_id;
+        std::tie(level, source, target, edge_id) = packed_edge;
+        if (edge_id != SPECIAL_EDGEID)
+        { // a base graph edge
+            unpacked_path.push_back(edge_id);
```
oxidase: I would keep `edge_id` in heap data, because a flag `is_overlay` is  needed to distinguish edge types. Here `is_overlay` flag  is `edge_id == SPECIAL_EDGEID`.

include/engine/search_engine_data.hpp
oxidase: I have tested performance of creating a new heap and clearing, it is almost the same. 
It should be changed in #3816

include/engine/search_engine_data.hpp
oxidase: yes, `HeapData::from_clique_arc` is only needed, but it will 4 bytes with alignment, so i placed complete edge ID. In CH the same flag is hidden in `QueryEdge` but i don't think it is good idea to cut one bit from node ID in `EdgeBasedEdge``

oxidase:
I have adjusted the branch to review comments, but two points still can be discussed and  i have no better ideas (i  pushed two additional commits into a reference branch to mld/routing-reference):
1. upper bound in f8cb6ba is similar to CH one, but it is too pessimistic because of the problem sketched here for 1D graph when one point is close to a high level border
upper_bound_problem
The highest difference level is 2, so forward and reverse heap will meet first at the left point, but the result will not be the shortest path from `s` to `t`. So the upper bound must be very pessimistic to find the correct middle point on the right side. Because of the different search levels the reverse heap will scan level 2 for nothing. In my original implementation 581c2a7 two upper bounds for the forward and reverse heaps are updated separately, so in the example the reverse heap will stop scanning and "wait" until the forward heap will get to the level 2 (it is not exactly the condition but the upper bound for the reverse heap can be shifted only when the forward heap will reach the second level).
Search spaces for both upper bounds
  -  one upper bound f8cb6ba https://gist.github.com/oxidase/57b4c8a1dd7482a73a02d6d6f4a663a0
  -  two separate upper bounds 581c2a7 https://gist.github.com/oxidase/b371c80479dc4598f399dddc6ef25d5b
2. keep the level value  in the heap data f6ec5d2. I tried to remove the level field, but found that i can not compute the level of an overlay edge from the node ids, because `GetHighestDifferentLevel(source, target)` for any overlay edge `source -> target` is 0. This leads to a problem: after going on a boundary edge to the level 1 and taking a shortcut on the level 1, the next possible level will be again 0, so MLD just falls back to the Dijkstra algorithm on the base graph. Atm, i don't know how to avoid storing the level value in the node data.

@TheMarex please could you check these two points?

oxidase:
Tested locally on german-latest: customization is ~33seconds

```
[info] Cells statistics per level
[info] Level 1 #cells 70419 #nodes 13943644,   source nodes: average 755984 (5.42171%) invalid 35 (0.00025101%),   destination nodes: average 1073630 (7.69978%) invalid 71 (0.000509193%)
[info] Level 2 #cells 5131 #nodes 13943644,   source nodes: average 121072 (0.868295%) invalid 14 (0.000100404%),   destination nodes: average 164219 (1.17773%) invalid 9 (6.45455e-05%)
[info] Level 3 #cells 315 #nodes 13943644,   source nodes: average 21721 (0.155777%) invalid 0 (0%),   destination nodes: average 28961 (0.2077%) invalid 1 (7.17173e-06%)
[info] Level 4 #cells 10 #nodes 13943644,   source nodes: average 1853 (0.0132892%) invalid 0 (0%),   destination nodes: average 2428 (0.017413%) invalid 0 (0%)
```

route [Munich->Hamburg](http://map.project-osrm.org/?z=7&center=50.812877%2C12.293701&loc=48.137132%2C11.575395&loc=53.547244%2C9.981766&hl=en&alt=0) ~30ms
```
[info] 15-03-2017 01:35:37 30.2905ms 127.0.0.1 - HTTPie/0.9.2 200 /route/v1/driving/11.578946,48.138177;9.981766,53.547244
[info] 15-03-2017 01:35:38 31.4429ms 127.0.0.1 - HTTPie/0.9.2 200 /route/v1/driving/11.578946,48.138177;9.981766,53.547244
[info] 15-03-2017 01:35:39 29.9721ms 127.0.0.1 - HTTPie/0.9.2 200 /route/v1/driving/11.578946,48.138177;9.981766,53.547244
[info] 15-03-2017 01:35:41 30.7037ms 127.0.0.1 - HTTPie/0.9.2 200 /route/v1/driving/11.578946,48.138177;9.981766,53.547244
[info] 15-03-2017 01:35:42 30.4965ms 127.0.0.1 - HTTPie/0.9.2 200 /route/v1/driving/11.578946,48.138177;9.981766,53.547244
[info] 15-03-2017 01:35:43 30.7287ms 127.0.0.1 - HTTPie/0.9.2 200 /route/v1/driving/11.578946,48.138177;9.981766,53.547244
[info] 15-03-2017 01:35:44 30.4513ms 127.0.0.1 - HTTPie/0.9.2 200 /route/v1/driving/11.578946,48.138177;9.981766,53.547244
```

Search space https://gist.github.com/oxidase/3015ec44f4570090a84790bba60a266d

TheMarex:
Questions were resolved as per slack: Above example uses the wrong query level for the backward heap.

oxidase:
@TheMarex i have pushed the current state:
* removed `level` from the heap data and added `get_query_heap` closure

* change to the proper upper bound. I could not use `WasInserted(node) && WasRemoved(node)` in the upper bound condition, because it changes snapping behavior of start points with negative weights. 
This change should be also in CH-dependent functions, algorithm-independent function `annotatePath` and tests expectations.

* removing `edge_id` in 223b65f requires to use `FindEdge` for every path edge, this adds a performance regression wrt to 7907cf3. Here is a query timing summary in milliseconds  for  [the test path](http://map.project-osrm.org/?z=10&center=48.479749%2C11.703186&loc=48.137132%2C11.575395&loc=48.762809%2C11.425202&hl=en&alt=0)
```
`7907cf3`
   Min. 1st Qu.  Median    Mean 3rd Qu.    Max.
  3.507   3.631   3.683   3.789   3.769   5.678

`223b65f`
   Min. 1st Qu.  Median    Mean 3rd Qu.    Max.
  4.464   4.614   4.679   4.863   4.821   7.626
```
compare
So i would keep `edge_id` in MLD heaps.

Still TODO: split MLD `search` function into  unconstrained path search and path unpacking that is constrained by level and parent cell id.

/cc @TheMarex

TheMarex:
> So i would keep edge_id in MLD heaps.

👍 In tfhat case a Algorithm-specific specialization of `SearchEngineData` would make sense.

> change to the proper upper bound. I could not use WasInserted(node) && WasRemoved(node) in the upper bound condition, because it changes snapping behavior of start points with negative weights

Yes we need to initialize the `forward_heap` to have a non-zero `forward_heap.MinKey()`. The easiest way to do this consistently for MLD and CH is to do a few initialization rounds. Something like this:

```
// heap is not fully initialized yet we need to resolve negative
// initial offsets first for Dijkstra to work
while (!forward_heap.empty() && forward_heap.MinKey() < 0)
{
     routeStep<true>(forward_heap, ....)
}

// real query with termination start here:
while (!forward_heap.empty() && )
{
     routeStep<true>(forward_heap, ....)

     routeStep<false>(reverse_heap, ....)

     if (forward_heap.MinKey() + reverse_heap.MinKey() > 0)
       // terminate
}
```

oxidase:
@TheMarex i think i addressed all open points. Just two notes:
* I split unrestricted and restricted  searches via argument packs https://github.com/Project-OSRM/osrm-backend/blob/mld/routing/include/engine/routing_algorithms/routing_base_mld.hpp#L24-L44
At first i tried to use closure `[sublevel] (NodeID) { return sublevel; }` for the fixed level restriction, but this introduces indirect function calls that can not be optimized out by compiler. With argument packs compiler uses a function pointer only for the unrestricted search, because the closure function should capture phantom nodes. For the restricted search a simple integer compare will be used. This gives improvement from 4ms to 3.6ms for the test path.

* I was wrong calling "never-go-down" [condition](https://github.com/Project-OSRM/osrm-backend/blob/mld/routing/include/engine/routing_algorithms/routing_base_mld.hpp#L138). Now it is simpler, but it is still needed to avoid going inside into cells on lower sublevels. Without this condition internal edges on the level 0 that can be ignored will be added into heap.

oxidase:
@TheMarex negative values should not be inserted into heaps or weight values after delete must be increased from negative values to 0 for starting points otherwise `.GetKey(start)` will return negative values.
I think the first option is better - it will allow to change weights to unsigned values.
It should be like:
```
heap.Insert(start_forward, 0 {start_forward}); heap.DeleteMin();
heap.Insert(start_reverse, 0 {start_reverse}); heap.DeleteMin();
for (edge : GetAdjacentEdgeRange(start_forward)) {
    assert(edge_data.weight - GetForwardWeightPlusOffset() >= 0);
    heap.Insert(to, edge_data.weight - GetForwardWeightPlusOffset(), {start_forward});
}
for (edge : GetAdjacentEdgeRange(start_reverse)) {
    assert(edge_data.weight - GetReverseWeightPlusOffset() >= 0);
    heap.Insert(to, edge_data.weight - GetReverseWeightPlusOffset(), {start_reverse});
}
```

This must be fixed in algorithm-independent part. One problem that I have found: 144 scenarios fail because of different snapping in 5a4ceaff9a47c8422f2f0c11d9bdd215b17d26d5. I have impression that the current snapping behavior is  used implicitly:  at https://github.com/oxidase/osrm-backend/blob/5a4ceaff9a47c8422f2f0c11d9bdd215b17d26d5/include/engine/guidance/assemble_steps.hpp#L227-L228 it is possible to get negative weights and durations if the first edge is changed.

TheMarex:
@oxidase yes good point, negative source key would still bite us. Your refactor will also allow use to address https://github.com/Project-OSRM/osrm-backend/issues/3429#issuecomment-267756917 eventually I think. This would give us fixing two bugs with one fix. 😄 

That code you linked only deals with the corner case of having two phantom nodes on the same segment. The routing should terminate immediately with an empty path (since source == target). Not sure how this could be influenced by these routing changes.

oxidase:
Checked on Europe size:
* partitioning
```
[info] Computing recursive bisection
[info] Loaded compressed node based graph: 101166490 edges, 215440564 nodes
[info]  running partition: 128 1.2 0.25 10 1000 # max_cell_size balance boundary cuts small_component_size
[info] large component [0]=40047027
[info] large component [1566047]=12077
[info] large component [8522990]=7732
[info] large component [10727121]=7385
[info] large component [12838659]=8000
[info] large component [13241465]=1277
[info] large component [19526101]=5302
[info] large component [22262219]=1619
[info] large component [29259452]=3320
[info] SCC run took: 14.1289s
[info] Full bisection done in 1472.02s
[info] Loaded node based graph to edge based graph mapping
[info] Loaded edge based graph for mapping partition ids: 358189760 edges, 94965162 nodes
[info] Edge-based-graph annotation:
[info]   level 1 #cells 477500 bit size 19
[info]   level 2 #cells 35067 bit size 16
[info]   level 3 #cells 2187 bit size 12
[info]   level 4 #cells 74 bit size 7
[info] MultiLevelPartition constructed in 84.7451 seconds
[warn] Node needs to either have incoming or outgoing edges in cell. Number of unconnected nodes is 236
[info] CellStorage constructed in 17.1767 seconds
[info] MLD data writing took 1.03853 seconds
[info] Bisection took 1760.56 seconds.
[info] RAM: peak bytes used: 40976625664
```
* customization **~ 2 min** :tada:
```
[info] Opening europe-latest.osrm.ebg
[info] Reading 179249752 edges from the edge based graph
[info] In total loaded 0 file(s) with a total of 0 unique values
[info] In total loaded 0 file(s) with a total of 0 unique values
[info] Done reading edges
[info] Loaded edge based graph for mapping partition ids: 358189760 edges, 94965162 nodes
[info] Loading partition data took 64.6892 seconds
[info] Cells customization took 47.9601 seconds
[info] MLD customization writing took 3.95046 seconds
[info] Graph writing took 4.476 seconds
[info] Cells statistics per level
[info] Level 1 #cells 477500 #nodes 94965162,   source nodes: average 5479064 (5.76955%) invalid 150 (0.000157953%),   destination nodes: average 7869387 (8.2866%) invalid 431 (0.000453851%)
[info] Level 2 #cells 35067 #nodes 94965162,   source nodes: average 854503 (0.899807%) invalid 27 (2.84315e-05%),   destination nodes: average 1188264 (1.25126%) invalid 81 (8.52944e-05%)
[info] Level 3 #cells 2187 #nodes 94965162,   source nodes: average 136718 (0.143966%) invalid 5 (5.26509e-06%),   destination nodes: average 187010 (0.196925%) invalid 18 (1.89543e-05%)
[info] Level 4 #cells 74 #nodes 94965162,   source nodes: average 12111 (0.0127531%) invalid 0 (0%),   destination nodes: average 16461 (0.0173337%) invalid 2 (2.10604e-06%)
[info] RAM: peak bytes used: 17938042880
```
* queries
Munich -> Hamburg
```
[info] 16-03-2017 15:33:45 40.7062ms 127.0.0.1 - curl/7.35.0 200 /route/v1/driving/11.578946,48.138177;9.981766,53.547244
[info] 16-03-2017 15:33:47 39.0631ms 127.0.0.1 - curl/7.35.0 200 /route/v1/driving/11.578946,48.138177;9.981766,53.547244
```
Lisbon -> Kiev
```
[info] 16-03-2017 15:38:24 162.832ms 127.0.0.1 - curl/7.35.0 200 /route/v1/driving/-9.136506,38.707793;30.524050,50.450107
[info] 16-03-2017 15:38:27 160.95ms 127.0.0.1 - curl/7.35.0 200 /route/v1/driving/-9.136506,38.707793;30.524050,50.450107
```

Athens -> London
```
[info] 16-03-2017 15:43:22 126.19ms 127.0.0.1 - curl/7.35.0 200 /route/v1/driving/23.727984,37.984149;-0.127647,51.507322
[info] 16-03-2017 15:43:23 126.622ms 127.0.0.1 - curl/7.35.0 200 /route/v1/driving/23.727984,37.984149;-0.127647,51.507322
[info] 16-03-2017 15:43:25 126.884ms 127.0.0.1 - curl/7.35.0 200 /route/v1/driving/23.727984,37.984149;-0.127647,51.507322
```

oxidase:
To compare wrt contraction with core 0.95
```
[info] Input file: europe-latest.osrm
[info] Threads: 32
[info] Reading node weights.
[info] Done reading node weights.
[info] Loading edge-expanded graph representation
[info] Opening europe-latest.osrm.ebg
[info] Reading 179249752 edges from the edge based graph
[info] In total loaded 0 file(s) with a total of 0 unique values
[info] In total loaded 0 file(s) with a total of 0 unique values
[info] Done reading edges
[STXXL-MSG] STXXL v1.3.1 (release)
[STXXL-MSG] 1 disks are allocated, total space: 250000 MiB
[info] merged 309744 edges out of 358499504
[info] contractor finished initalization
[info] initializing elimination PQ ...ok
[info] preprocessing 94965162 nodes ...
[info] . 10% . 20% . 30% . 40% . 50% . 60%  [flush 60147171 nodes] . 70% . 80% . 90% 	.[info] [core] 4693662 nodes 223066107 edges.

[info] Getting edges of minimized graph . 10% . 20% . 30% . 40% . 50% . 60% . 70% . 80% . 90% . 100%
[info] Contraction took 1015.5 sec
[info] Serializing compacted graph of 437397367 edges
[info] Building node array
[info] Serializing node array
[info] Writing CRC32: 2378896905
[info] Building edge array
[info] Preprocessing : 1059.6 seconds
[info] Contraction: 93515 nodes/sec and 430721 edges/sec
[info] finished preprocessing
[info] STXXL: memory summary not available, needs STXXL 1.4 or higher
[info] RAM: peak bytes used: 43861553152
```

Munich -> Hamburg
```
[info] 16-03-2017 16:17:43 447.272ms 127.0.0.1 - curl/7.35.0 200 /route/v1/driving/11.578946,48.138177;9.981766,53.547244
[info] 16-03-2017 16:17:45 428.854ms 127.0.0.1 - curl/7.35.0 200 /route/v1/driving/11.578946,48.138177;9.981766,53.547244
[info] 16-03-2017 16:17:49 436.507ms 127.0.0.1 - curl/7.35.0 200 /route/v1/driving/11.578946,48.138177;9.981766,53.547244
[info] 16-03-2017 16:17:51 451.617ms 127.0.0.1 - curl/7.35.0 200 /route/v1/driving/11.578946,48.138177;9.981766,53.547244
[info] 16-03-2017 16:17:53 454.651ms 127.0.0.1 - curl/7.35.0 200 /route/v1/driving/11.578946,48.138177;9.981766,53.547244
```

Lisbon -> Kiev
```
[info] 16-03-2017 16:18:18 3278.79ms 127.0.0.1 - curl/7.35.0 200 /route/v1/driving/-9.136506,38.707793;30.524050,50.450107
[info] 16-03-2017 16:18:24 3051.81ms 127.0.0.1 - curl/7.35.0 200 /route/v1/driving/-9.136506,38.707793;30.524050,50.450107
[info] 16-03-2017 16:18:29 3128.72ms 127.0.0.1 - curl/7.35.0 200 /route/v1/driving/-9.136506,38.707793;30.524050,50.450107
[info] 16-03-2017 16:18:34 3194.31ms 127.0.0.1 - curl/7.35.0 200 /route/v1/driving/-9.136506,38.707793;30.524050,50.450107
```

Athens -> London
```
[info] 16-03-2017 16:18:45 2892.4ms 127.0.0.1 - curl/7.35.0 200 /route/v1/driving/23.727984,37.984149;-0.127647,51.507322
[info] 16-03-2017 16:18:50 2828.1ms 127.0.0.1 - curl/7.35.0 200 /route/v1/driving/23.727984,37.984149;-0.127647,51.507322
[info] 16-03-2017 16:18:54 2821.27ms 127.0.0.1 - curl/7.35.0 200 /route/v1/driving/23.727984,37.984149;-0.127647,51.507322
[info] 16-03-2017 16:18:58 2832.04ms 127.0.0.1 - curl/7.35.0 200 /route/v1/driving/23.727984,37.984149;-0.127647,51.507322
[info] 16-03-2017 16:19:02 2665.64ms 127.0.0.1 - curl/7.35.0 200 /route/v1/driving/23.727984,37.984149;-0.127647,51.507322
```

File sizes:
```
19362940 europe-latest.osm.pbf
14013292 europe-latest.osrm
 7205296 europe-latest.osrm.hsgr
 7098264 europe-latest.osrm.fileIndex
 5967680 europe-latest.osrm.mldgr
 5858192 europe-latest.osrm.geometry
 3500976 europe-latest.osrm.ebg
 3366264 europe-latest.osrm.nodes
 2800784 europe-latest.osrm.edges
 2473500 europe-latest.osrm.cnbg
  842444 europe-latest.osrm.icd
  790368 europe-latest.osrm.cnbg_to_ebg
  742068 europe-latest.osrm.partition
  731864 europe-latest.osrm.cells
  370960 europe-latest.osrm.enw
  370960 europe-latest.osrm.level
  350104 europe-latest.osrm.turn_duration_penalties
  350104 europe-latest.osrm.turn_weight_penalties
   92740 europe-latest.osrm.core
   63692 europe-latest.osrm.names
   11508 europe-latest.osrm.restrictions
    7260 europe-latest.osrm.ramIndex
      68 europe-latest.osrm.datasource_names
      20 europe-latest.osrm.tls
      12 europe-latest.osrm.tld
       4 europe-latest.osrm.properties
```

oxidase:
Full CH preprocessing takes 1415.92 seconds and query responses are
```
[info] 16-03-2017 17:31:21 6.67244ms 127.0.0.1 - curl/7.35.0 200 /route/v1/driving/11.578946,48.138177;9.981766,53.547244
[info] 16-03-2017 17:31:23 5.79274ms 127.0.0.1 - curl/7.35.0 200 /route/v1/driving/11.578946,48.138177;9.981766,53.547244
[info] 16-03-2017 17:31:24 5.73531ms 127.0.0.1 - curl/7.35.0 200 /route/v1/driving/11.578946,48.138177;9.981766,53.547244
[info] 16-03-2017 17:31:28 7.30056ms 127.0.0.1 - curl/7.35.0 200 /route/v1/driving/11.578946,48.138177;9.981766,53.547244
```
```
[info] 16-03-2017 17:31:47 34.731ms 127.0.0.1 - curl/7.35.0 200 /route/v1/driving/-9.136506,38.707793;30.524050,50.450107
[info] 16-03-2017 17:31:49 31.7682ms 127.0.0.1 - curl/7.35.0 200 /route/v1/driving/-9.136506,38.707793;30.524050,50.450107
[info] 16-03-2017 17:31:53 31.9014ms 127.0.0.1 - curl/7.35.0 200 /route/v1/driving/-9.136506,38.707793;30.524050,50.450107
[info] 16-03-2017 17:31:54 46.9425ms 127.0.0.1 - curl/7.35.0 200 /route/v1/driving/-9.136506,38.707793;30.524050,50.450107
```

```
[info] 16-03-2017 17:32:15 24.2817ms 127.0.0.1 - curl/7.35.0 200 /route/v1/driving/23.727984,37.984149;-0.127647,51.507322
[info] 16-03-2017 17:32:16 24.5297ms 127.0.0.1 - curl/7.35.0 200 /route/v1/driving/23.727984,37.984149;-0.127647,51.507322
[info] 16-03-2017 17:32:17 24.5493ms 127.0.0.1 - curl/7.35.0 200 /route/v1/driving/23.727984,37.984149;-0.127647,51.507322
```

So MLD ~ 5-6 times slower than full CH

TheMarex:
@oxidase looks good to me know. Merge at will. 🎉

oxidase:
Rebased onto refactor/split-updater. Must be merge after #3809

PR #3789, opened 2017-03-08
Ensure Extractor scheduler is requested with non-Zero number of threads

mloskot:
By default, `ExtractorConfig::requested_num_threads` is Zero, which is an invalid value for tbb task scheduler.
If user does not specify value greater than Zero, then use `tbb::task_scheduler_init::automatic`.


src/extractor/extractor.cpp
```diff
@@ -118,7 +118,7 @@ int Extractor::run(ScriptingEnvironment &scripting_environment)
     TIMER_START(extracting);
 
     const unsigned recommended_num_threads = tbb::task_scheduler_init::default_num_threads();
-    const auto number_of_threads = std::min(recommended_num_threads, config.requested_num_threads);
+    const auto number_of_threads = std::min(recommended_num_threads, std::max(config.requested_num_threads, 1u));
```
daniel-j-h: Did you check all tools, I think we're doing it in most if not all of them

mloskot: Nope, I didn't.

One more thing I noticed was that `Extractor::run` initialises the `tbb::task_scheduler_init` based on the config, while `Contractor::Run` does not (seems like leaving it as user's responsibility). I'm unsure, but looks like overlooked interface unification or perhaps by-design. I know too little of the OSRM, didn't want to draw too far-reaching conclusions :)

mloskot:
I've switched from `1u` to `tbb::task_scheduler_init::automatic`.

src/extractor/extractor.cpp
daniel-j-h: Yes probably overlooked or for some other historic reasons.

As I said, all the tools and the library interfaces should be consistent when possible.

mloskot: > As I said, all the tools and the library interfaces should be consistent when possible.

What does it mean for this PR? Rejection or Need Work? If the latter, what changes you suggest?

daniel-j-h: No this PR is fine, it just would be great to consolidate the initialization across all tools and interfaces (extract, contract, ..)

mloskot: Right. Looks like a job for a `{Extractor|Contractor|*Config` factory.

PR #3790, opened 2017-03-09
Compute correct speed values in tile plugin

oxidase:
# Issue

In #2399  tile plugin was not adjusted and returned weight-based speed values that look misleading if profile weight is not duration. 
PR fixes speed to be `distance / duration`, adds `weight` feature and fixes turn durations.

/cc @TheMarex 

## Tasklist
 - [x] review
 - [x] adjust for comments

## Requirements / Relations

### DIFF
--- a/src/engine/plugins/tile.cpp
+++ b/src/engine/plugins/tile.cpp
@@ -421,6 +421,8 @@ void encodeVectorTile(const datafacade::ContiguousInternalMemoryDataFacadeBase &
             for (const auto &edge_index : sorted_edge_indexes)
             {
                 const auto &edge = edges[edge_index];
+
+                // Weight values
                 const auto forward_weight_vector =
                     facade.GetUncompressedForwardWeights(edge.packed_geometry_id);
                 const auto reverse_weight_vector =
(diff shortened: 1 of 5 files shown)


src/engine/plugins/tile.cpp
```diff
@@ -568,14 +596,15 @@ void encodeVectorTile(const datafacade::ContiguousInternalMemoryDataFacadeBase &
 
                         // Calculate the speed for this line
                         std::uint32_t speed_kmh =
-                            static_cast<std::uint32_t>(round(length / reverse_weight * 10 * 3.6));
+                            static_cast<std::uint32_t>(round(length / reverse_duration * 10 * 3.6));
```
TheMarex: If statement above needs to be adjusted as well, it check for zero division.

src/engine/plugins/tile.cpp
```diff
@@ -544,14 +571,15 @@ void encodeVectorTile(const datafacade::ContiguousInternalMemoryDataFacadeBase &
 
                         // Calculate the speed for this line
                         std::uint32_t speed_kmh =
-                            static_cast<std::uint32_t>(round(length / forward_weight * 10 * 3.6));
+                            static_cast<std::uint32_t>(round(length / forward_duration * 10 * 3.6));
```
TheMarex: If statement above needs to be adjusted as well, it check for zero division.

PR #3804, opened 2017-03-10
Adds Dockerfile for Docker CI

daniel-j-h:
This adds a Dockerfile to the repository for building the current source tree. Once we have this in place we can use the Docker CI system to build and publish images automatically for every push / tag.

Docker CI is here https://hub.docker.com/r/osrm/osrm-backend-ci/

These images should be around 10 - 20 MB and can be used directly e.g. on the demo server.
Or by our users. Or for reproducing issues.

---

Also removes our unused Docker scripts. If you want to replicate the Trusty environment just do

    docker run -it ubuntu:trusty /bin/bash

and go from there.

---

cc @miccolis @danpat


daniel-j-h:
Success! We have our first fully automated osrm docker image
- https://hub.docker.com/r/osrm/osrm-backend-ci/builds/bv39enxwtquvxnc2vw839xs/
- https://hub.docker.com/r/osrm/osrm-backend-ci/tags/

TheMarex:
Great, how are these build triggered?

danpat:
@TheMarex dockerhub has a commit hook in the osrm-backend repo.

daniel-j-h:
Automated Docker images are getting published to

https://hub.docker.com/r/osrm/osrm-backend-ci/

at the moment.

We want to switch to osrm/osrm-backend once we made sure publishing works.

Ping me if you want access to the Docker Hub orga.

danpat:
Main downside is that this is just `Release` mode binaries, right?  I guess if anyone wants debug variants, it's easy enough to build your own images.

daniel-j-h:
I looked a bit into it and I think there is a way to do multiple builds:
- [Configure](https://docs.docker.com/docker-hub/builds/#create-an-automated-build) multiple automated build tags for the same branch (e.g. `latest-release`, `latest-debug`).
- In the Dockerfile dispatch based upon the [`DOCKER_TAG`](https://docs.docker.com/docker-cloud/builds/advanced/) and parse the release / debug part.

It's a bit ugly since we have to encode all options we want to pass in the tag.
But at least for release and debug builds this should do the trick.

Issue #3298, opened 2016-11-12
Prepare list of deprecated API parameters/values for `6.0` release


TheMarex:
We deprecated some API parameters/values over the 5.x live-cycle and need to prepare for the 6.x cleanup.

Off the top of my head the following things are marked for removal:

- [ ] `StepManeuver.bearing_before` and `StepManeuver.bearing_after` (replaced by `RouteStep.intersections`)
- [ ] `StepManeuver.location` (replaced by `RouteStep.intersections`)
- [ ] `StepManeuver.type = "ramp"` (replaced by `on ramp` and `off ramp`)
- [ ] Using `bearings=` without `radiuses=` per https://github.com/Project-OSRM/osrm-backend/issues/2983
- [ ] Specifying a `bearings` range per https://github.com/Project-OSRM/osrm-backend/issues/2983

Any more things that come to mind @MoKob @daniel-j-h @danpat @freenerd?

daniel-j-h:
libosrm API:
- https://github.com/Project-OSRM/osrm-backend/pull/2941
- https://github.com/Project-OSRM/osrm-backend/pull/2991
- potentially moving away from the JSON variant

MoKob:
- [ ] roundabout instruction -> enter roundabout + exit roundabout
- [ ] direction_modifier u-turn -> invalid + new turn type u-turn

TheMarex:
@MoKob can you link the tickets for those here? I'm missing some context here.

MoKob:
Also:
 - [ ] valid per indication https://github.com/Project-OSRM/osrm-backend/issues/2935

daniel-j-h:
- [ ] Switch from always generating hint to hints on demand https://github.com/Project-OSRM/osrm-backend/pull/3456

karenzshea:
- [ ] Remove redundant boolean based parameter for annotations in RouteParameters https://github.com/Project-OSRM/osrm-backend/issues/3644

freenerd:
- [ ] Remove geojson responses and go polyline only https://github.com/Project-OSRM/osrm-backend/issues/3752

TheMarex:
- [ ] Remove `--generate-edge-lookup` parameter from `osrm-{contract,customize}` as it does nothing now

daniel-j-h:
[...]

PR #3809, opened 2017-03-11
Refactor update routines

TheMarex:
# Issue

This continue https://github.com/Project-OSRM/osrm-backend/pull/3785 to make the update code easier to read and simplify logic.

On my test dataset the update speed was slightly faster then before, but the CSV parsing dominates almost all processing time.

## Tasklist
 - [x] review
 - [x] Update wiki of changed files
 - [x] adjust for comments

## Relations

Implements https://github.com/Project-OSRM/osrm-backend/issues/3737


src/updater/updater.cpp
```diff
@@ -149,446 +95,568 @@ void CheckWeightsConsistency(
...
+            // In this case we want it oriented from in forward directions
+            auto rev_weights_range =
+                boost::adaptors::reverse(segment_data.GetReverseWeights(geometry_id));
+            auto rev_durations_range =
+                boost::adaptors::reverse(segment_data.GetReverseDurations(geometry_id));
+            auto rev_datasources_range =
+                boost::adaptors::reverse(segment_data.GetReverseDatasources(geometry_id));
+            bool rev_was_updated = false;
+
+            for (auto segment_offset = 0UL; segment_offset < fwd_weights_range.size();
```
daniel-j-h: Check all 0UL's - either replace with 0ull or better size_t instead of auto.

0ul is unsigned long; longs are only guaranteed to have 32 bits (in reality: Linux 64 bit, Windows 32 bit).

http://en.cppreference.com/w/cpp/language/types

src/updater/updater.cpp
```diff
@@ -149,446 +95,568 @@ void CheckWeightsConsistency(
...
-        tbb::enumerable_thread_specific<counters_type> segment_speeds_counters(
-            counters_type(num_counters, 0));
-        const constexpr auto LUA_SOURCE = 0;
-
-        tbb::parallel_for_each(first, last, [&](const LeafNode &current_node) {
-            auto &counters = segment_speeds_counters.local();
-            for (size_t i = 0; i < current_node.object_count; i++)
+        for (DirectionalGeometryID geometry_id = 0;
+             geometry_id < segment_data.GetNumberOfGeometries();
+             geometry_id++)
```
daniel-j-h: `for (auto it = 0; it < fn(); ++it)`

computes `fn()` on every loop, why not

`for (auto it = 0, last = fn(); it < last; ++it)`

or

`irange(it, fn());`

src/updater/updater.cpp
```diff
@@ -149,446 +95,568 @@ void CheckWeightsConsistency(
...
+    extractor::Datasources sources;
+    DatasourceID source = 0;
+    sources.SetSourceName(source++, "lua profile");
+
+    // Only write the filename, without path or extension.
+    // This prevents information leakage, and keeps names short
+    // for rendering in the debug tiles.
+    for (auto const &name : config.segment_speed_lookup_paths)
+    {
+        sources.SetSourceName(source++, boost::filesystem::path(name).stem().string());
```
daniel-j-h: I know you didn't change this but could you split the `source++`.
Pre / post increment magic just makes it harder to get into this code.

src/updater/updater.cpp
```diff
@@ -149,446 +95,568 @@ void CheckWeightsConsistency(
...
-            else
-            {
-                // segments_speeds_counters has 0 as LUA, segment_speed_filenames not, thus we need
-                // to susbstract 1 to avoid off-by-one error
-                util::Log() << "Used " << merged_counters[i] << " speeds from "
-                            << config.segment_speed_lookup_paths[i - 1];
-            }
+            util::Log(logWARNING) << "Negative turn penalty at " << turn_index.from_id << ", "
+                                  << turn_index.via_id << ", " << turn_index.to_id
+                                  << ": turn penalty " << turn_weight_penalty;
```
daniel-j-h: Care to explain to me what happens on negative penalties? Seems like we store them above anyway and here only log their occurrence. We're crashing somewhere up the toolchain when this happens currently, no?

src/updater/updater.cpp
```diff
@@ -149,446 +95,568 @@ void CheckWeightsConsistency(
...
+Updater::LoadAndUpdateEdgeExpandedGraph(std::vector<extractor::EdgeBasedEdge> &edge_based_edge_list,
+                                        std::vector<EdgeWeight> &node_weights) const
+{
+    TIMER_START(load_edges);
 
-    const auto save_datastore_names = [&] {
-        extractor::Datasources sources;
-        DatasourceID source = 0;
-        sources.SetSourceName(source++, "lua profile");
+    auto max_edge_id = 0;
```
daniel-j-h: EdgeID

src/updater/updater.cpp
```diff
@@ -149,446 +95,568 @@ void CheckWeightsConsistency(
...
-        for (auto const &name : config.segment_speed_lookup_paths)
-        {
-            sources.SetSourceName(source++, boost::filesystem::path(name).stem().string());
-        }
+    {
+        storage::io::FileReader reader(config.edge_based_graph_path,
+                                       storage::io::FileReader::VerifyFingerprint);
+        auto num_edges = reader.ReadElementCount64();
+        edge_based_edge_list.resize(num_edges);
+        max_edge_id = reader.ReadOne<EdgeID>();
```
daniel-j-h: Huh interesting that this is not an error / warning at all - you're reading a u32 into a i32 here.

src/updater/updater.cpp
```diff
@@ -149,446 +95,568 @@ void CheckWeightsConsistency(
...
+                              SOURCE_REF);
 
+    extractor::ProfileProperties profile_properties;
+    std::vector<extractor::OriginalEdgeData> edge_data;
+    extractor::SegmentDataContainer segment_data;
     std::vector<TurnPenalty> turn_weight_penalties;
     std::vector<TurnPenalty> turn_duration_penalties;
+    if (update_edge_weights || update_turn_penalties)
+    {
+        const auto load_segment_data = [&]() {
```
daniel-j-h: fyi you can leave out the () in nullary lambdas

src/updater/updater.cpp
```diff
@@ -149,446 +95,568 @@ void CheckWeightsConsistency(
...
+        auto updated_turn_penalties = updateTurnPenalties(config,
+                                                          profile_properties,
+                                                          turn_penalty_lookup,
+                                                          turn_weight_penalties,
+                                                          turn_duration_penalties);
+        // we need to re-compute all edges that have updated turn penalties.
+        // this marks it for re-computation
+        std::transform(updated_turn_penalties.begin(),
+                       updated_turn_penalties.end(),
+                       std::back_inserter(updated_segments),
```
daniel-j-h: I would be cautious with back_inserter (and to some extend push_back), if you do a structure preserving transform and call the inserter for every element anyway, the better alternative is to simply resize and assign.

Here's some anecdotal evidence https://github.com/mattgodbolt/zindex/commit/27d229afc5021f208467c3526b5a223ecf44b6e4

(Not only or specifically here, it just caught my eye)

src/updater/updater.cpp
```diff
@@ -149,446 +95,568 @@ void CheckWeightsConsistency(
...
-
-            // Update the node-weight cache. This is the weight of the edge-based-node only,
-            // it doesn't include the turn. We may visit the same node multiple times, but
+            const auto weights = segment_data.GetForwardWeights(geometry_id.id);
+            for (const auto weight : weights)
+            {
+                if (weight == INVALID_EDGE_WEIGHT)
+                {
+                    new_weight = INVALID_EDGE_WEIGHT;
+                    break;
```
daniel-j-h: What's the deal here - if one weight is invalid are 1/ all invalid 2/ some invalid 3/ only one invalid?

src/updater/updater.cpp
```diff
@@ -149,446 +95,568 @@ void CheckWeightsConsistency(
...
+                }
+                new_weight += weight;
+            }
+            const auto durations = segment_data.GetForwardDurations(geometry_id.id);
+            new_duration = std::accumulate(durations.begin(), durations.end(), 0);
```
daniel-j-h: typed init

src/updater/updater.cpp
```diff
@@ -149,446 +95,568 @@ void CheckWeightsConsistency(
...
+        }
+        else
+        {
+            const auto weights = segment_data.GetReverseWeights(geometry_id.id);
+            for (const auto weight : weights)
+            {
+                if (weight == INVALID_EDGE_WEIGHT)
+                {
+                    new_weight = INVALID_EDGE_WEIGHT;
+                    break;
```
daniel-j-h: same

src/updater/updater.cpp
```diff
@@ -149,446 +95,568 @@ void CheckWeightsConsistency(
...
+                }
+                new_weight += weight;
+            }
+            const auto durations = segment_data.GetReverseDurations(geometry_id.id);
+            new_duration = std::accumulate(durations.begin(), durations.end(), 0);
```
daniel-j-h: same

src/updater/updater.cpp
```diff
@@ -149,446 +95,568 @@ void CheckWeightsConsistency(
...
+            edge.data.weight = new_weight + turn_weight_penalty;
+            edge.data.duration = new_duration + turn_duration_penalty;
         }
+    };
 
-        edge_based_edge_list.emplace_back(std::move(inbuffer));
+    if (updated_segments.size() > 0)
+    {
+        tbb::parallel_for(tbb::blocked_range<std::size_t>(0, edge_based_edge_list.size()),
+                          [&](const tbb::blocked_range<std::size_t> &range) {
```
daniel-j-h: auto

src/updater/updater.cpp
```diff
@@ -149,446 +95,568 @@ void CheckWeightsConsistency(
...
+            }
+            const auto durations = segment_data.GetReverseDurations(geometry_id.id);
+            new_duration = std::accumulate(durations.begin(), durations.end(), 0);
+        }
+        return std::make_tuple(new_weight, new_duration);
+    };
+
+    std::vector<WeightAndDuration> accumulated_segment_data(updated_segments.size());
+    tbb::parallel_for(tbb::blocked_range<std::size_t>(0, updated_segments.size()),
+                      [&](const tbb::blocked_range<std::size_t> &range) {
```
daniel-j-h: auto

include/contractor/graph_contractor_adaptors.hpp
```diff
@@ -20,6 +20,9 @@ std::vector<ContractorEdge> adaptToContractorInput(InputEdgeContainer input_edge
 
     for (const auto &input_edge : input_edge_list)
     {
+        if (input_edge.data.weight == INVALID_EDGE_WEIGHT)
+            continue;
```
TheMarex: No previously we filtered these edges out the edge loading routine. We we loading one struct at a time, so we could just skip them.

include/partition/edge_based_graph_reader.hpp
```diff
@@ -30,6 +30,9 @@ splitBidirectionalEdges(const std::vector<extractor::EdgeBasedEdge> &edges)
 
     for (const auto &edge : edges)
     {
+        if (edge.data.weight == INVALID_EDGE_WEIGHT)
+            continue;
```
TheMarex: Se above.

src/updater/updater.cpp
```diff
@@ -149,446 +95,568 @@ void CheckWeightsConsistency(
...
+    if (config.log_edge_updates_factor > 0)
+    {
+        // copy the old data so we can compare later
+        segment_data_backup = std::make_unique<extractor::SegmentDataContainer>(segment_data);
+    }
 
-    auto mmap_file = [](const std::string &filename, boost::interprocess::mode_t mode) {
-        using boost::interprocess::file_mapping;
-        using boost::interprocess::mapped_region;
+    tbb::concurrent_vector<GeometryID> updated_segments;
```
TheMarex: Interesting, I will try benchmarking this. Main problem I had so far is that the whole update code has so much variance (+- 200ms) it makes micro-benchmarks hard. :-/

features/support/shared_steps.js
```diff
@@ -152,9 +152,9 @@ module.exports = function () {
                                     let a_type = k.slice(2);
                                     if (whitelist.indexOf(a_type) == -1)
                                         return cb(new Error('Unrecognized annotation field', a_type));
-                                    if (!annotation[a_type])
+                                    if (annotation && !annotation[a_type])
                                         return cb(new Error('Annotation not found in response', a_type));
-                                    got[k] = annotation[a_type];
+                                    got[k] = annotation && annotation[a_type] || '';
```
TheMarex: Noticed this while debugging a failing test. Cucumber crashes instead of just failing the test case.

include/extractor/segment_data_container.hpp
```diff
@@ -63,30 +63,65 @@ template <bool UseShareMemory> class SegmentDataContainerImpl
     {
     }
 
-    // TODO these random access functions can be removed after we implemented #3737
-    auto &ForwardDuration(const DirectionalGeometryID id, const SegmentOffset offset)
+    auto GetForwardGeometry(const DirectionalGeometryID id)
     {
-        return fwd_durations[index[id] + 1 + offset];
+        const auto begin = nodes.begin() + index.at(id);
+        const auto end = nodes.begin() + index.at(id + 1);
```
TheMarex: Good catch, don't remember doing this change. oO

src/tools/extract.cpp
```diff
@@ -57,13 +52,17 @@ return_code parseArguments(int argc, char *argv[], extractor::ExtractorConfig &e
...
     // shown to the user
     boost::program_options::options_description hidden_options("Hidden options");
     hidden_options.add_options()(
         "input,i",
         boost::program_options::value<boost::filesystem::path>(&extractor_config.input_path),
-        "Input file in .osm, .osm.bz2 or .osm.pbf format");
+        "Input file in .osm, .osm.bz2 or .osm.pbf format")(
+        "generate-edge-lookup",
+        boost::program_options::value<bool>(&dummy)->implicit_value(true)->default_value(false),
+        "Not used anymore");
```
TheMarex: https://github.com/Project-OSRM/osrm-backend/issues/3298#issuecomment-287195377

src/updater/updater.cpp
TheMarex: Replaced it with `irange` to side-step the issue.

src/updater/updater.cpp
TheMarex: So the gist is that we have allowed negative turn penalties for several reason (mostly around the fact that it's easier to define them w.r.t. to the avg speed of the segment), but they are not a good idea and can cause bugs. This warning makes sure we have a paper trail in the logs if someone feeds them in regardless.

src/updater/updater.cpp
TheMarex: Yeah if one segment on the path is blocked we can't fully traverse it anymore, hence we filter it out completely.
If we should really remove the whole edge is up to debate but this is consistent with old behavior.

TheMarex:
@daniel-j-h should have addressed all your comments, let me know if this works for you.

src/updater/updater.cpp
```diff
@@ -149,446 +95,569 @@ void CheckWeightsConsistency(
...
+                segment_lengths.push_back(util::coordinate_calculation::greatCircleDistance(
+                    util::Coordinate{internal_to_external_node_map[u]},
+                    util::Coordinate{internal_to_external_node_map[v]}));
+            });
+
+            auto fwd_weights_range = segment_data.GetForwardWeights(geometry_id);
+            auto fwd_durations_range = segment_data.GetForwardDurations(geometry_id);
+            auto fwd_datasources_range = segment_data.GetForwardDatasources(geometry_id);
+            bool fwd_was_updated = false;
+            for (const auto segment_offset : util::irange(0, fwd_weights_range.size()))
```
oxidase: SFINAE fails for (int, size_t) -> `util::irange<std::size_t>(0, fwd_weights_range.size()))`

src/updater/updater.cpp
```diff
@@ -149,446 +95,569 @@ void CheckWeightsConsistency(
...
+            // In this case we want it oriented from in forward directions
+            auto rev_weights_range =
+                boost::adaptors::reverse(segment_data.GetReverseWeights(geometry_id));
+            auto rev_durations_range =
+                boost::adaptors::reverse(segment_data.GetReverseDurations(geometry_id));
+            auto rev_datasources_range =
+                boost::adaptors::reverse(segment_data.GetReverseDatasources(geometry_id));
+            bool rev_was_updated = false;
+
+            for (const auto segment_offset : util::irange(0, rev_weights_range.size()))
```
oxidase: and here

Issue #3778, opened 2017-03-06
Add the option to remove the trace split check in the match method


gardster:
The match method has difficulties at tunnels processing, because there are no points during the tunnel. 
http://bl.ocks.org/d/35c59aaffa581712ad7e89db9b3c8872
This is happening because of the trace split check. To fix this at client side I need to add the interpolated point to have points often than 'median_sample_time * MAX_BROKEN_STATES'. Or filter the outer track to have 'median_sample_time' big enough. Both methods are tricky and seems like a black magic. I propose to add the special parameter to the request, that will say that I'm sure in my track and it doesn't require additional filtration or processing. So OSRM will not split my track against my will.

daniel-j-h:
This landed with https://github.com/Project-OSRM/osrm-backend/pull/3815. Closing here.

PR #3815, opened 2017-03-13
Tidying prarameter for the map matching plugin.

gardster:
# Issue

#3778

Adds special parameter to turn off track splitting. I think this parameter will be useful also in other tidying functions, like in #3149 
Default value does not change current map matching behavior.

## Tasklist
 - [x] review
 - [x] adjust for comments

### DIFF
--- a/docs/http.md
+++ b/docs/http.md
@@ -280,7 +280,7 @@ In addition to the [general options](#general-options) the following options are
 |timestamps  |`{timestamp};{timestamp}[;{timestamp} ...]`     |Timestamps for the input locations in seconds since UNIX epoch. Timestamps need to be monotonically increasing. |
 |radiuses    |`{radius};{radius}[;{radius} ...]`              |Standard deviation of GPS precision used for map matching. If applicable use GPS accuracy.|
 |gaps        |`split` (default), `ignore`                     |Allows the input track splitting based on huge timestamp gaps between points.             |
-|tidying     |`true`, `false` (default)                       |Allows the input track modification to obtain better matching quality for noisy tracks.   |
+|tidy        |`true`, `false` (default)                       |Allows the input track modification to obtain better matching quality for noisy tracks.   |
 
 |Parameter   |Values                             |
 |------------|-----------------------------------|
--- a/docs/http.md
+++ b/docs/http.md
@@ -274,7 +274,7 @@ In addition to the [general options](#general-options) the following options are
 |Option      |Values                                          |Description                                                                               |
 |------------|------------------------------------------------|------------------------------------------------------------------------------------------|
 |steps       |`true`, `false` (default)                       |Return route steps for each route                                                         |
-|tidying     |`true` (default), `false`                       |Allows the input track modification to obtain better matching quality for noisy tracks.   |
+|preprocess  |`full`, `simple` (default), `false`             |Allows the input track modification to obtain better matching quality for noisy tracks.   |
 |geometries  |`polyline` (default), `polyline6`, `geojson`    |Returned route geometry format (influences overview and per step)                         |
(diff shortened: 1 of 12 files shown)


daniel-j-h:
Could you add a changelog entry. Travis is failing because of clang-format check.


Also there is a trace tidying implementation for map matching [here](https://github.com/Project-OSRM/osrm-backend/pull/3149) which needs to be hooked up to the plugin. It does exactly the opposite of what you are doing here: it removes too many trace points from the trace for blob removal. Not sure if we should call this parameter `tidying` and have it on by default. Can we unify both PRs?

gardster:
Ok. I cherry-picked #3149 functions and added it to the matching logic. So now this parameter called `preprocess` and has 3 options: 

- `false` - without any preprocessing, or split identification;
- `simple` - current matching behavior, when the plugin checks timestamps gaps. I think that this option is good by default, because it keeps current behavior;
- `full` option - for #3149 logic.

At what release note I need to add a changelog entry?

danpat:
@gardster Just put it in the top section - we'll merge this into master, then we usually edit the CHANGELOG when it comes time to branch the next release.

gardster:
@danpat I added a changelog entry.

#3149 as is fail the CI tests. Maybe we ommit this functionality from current PR (but keep all flag options), and include track tidying when it will be ready?

TheMarex:
API wise I don't think `tidy` is the right word for what you are trying to disable. Nor is this really connected to @daniel-j-h PR, but thanks for integrating it anyway! Proposal:

- `gaps=split,ignore` (`ignore` enables the functionality you implemented)
- `tidy=true,false` (implements what @daniel-j-h has implemented)

However the build is failing to compile right now, so I'll mark this as WIP.

gardster:
@TheMarex @daniel-j-h  I split options to `gaps` and `tidy`, fix build failure and add some additional logic to handle tidied coordinates. Please, take another look to this PR.

docs/http.md
```diff
@@ -279,6 +279,8 @@ In addition to the [general options](#general-options) the following options are
 |overview    |`simplified` (default), `full`, `false`         |Add overview geometry either full, simplified according to highest zoom level it could be display on, or not at all.|
 |timestamps  |`{timestamp};{timestamp}[;{timestamp} ...]`     |Timestamps for the input locations in seconds since UNIX epoch. Timestamps need to be monotonically increasing. |
 |radiuses    |`{radius};{radius}[;{radius} ...]`              |Standard deviation of GPS precision used for map matching. If applicable use GPS accuracy.|
+|gaps        |`split` (default), `ignore`                     |Allows the input track splitting based on huge timestamp gaps between points.             |
+|tidying     |`true`, `false` (default)                       |Allows the input track modification to obtain better matching quality for noisy tracks.   |
```
TheMarex: Can you rename the option to `tidy` (verb)? Thanks!

include/engine/api/match_parameters.hpp
```diff
@@ -50,23 +50,37 @@ namespace api
...
+                    GapsType gaps_processing_,
+                    bool use_tidying_,
+                    Args... args_)
+        : RouteParameters{std::forward<Args>(args_)...}, timestamps{std::move(timestamps_)},
+          gaps_processing(gaps_processing_), use_tidying(use_tidying_)
     {
     }
 
     std::vector<unsigned> timestamps;
+    GapsType gaps_processing;
```
TheMarex: Just for consistency I would try to name these like the query option: `gaps` and `tidy`.

gardster:
Fixed

TheMarex:
@gardster thanks for getting this in. 👍

PR #3816, opened 2017-03-13
Use boost d-ary heap for MLD routing

oxidase:
# Issue

For #3764 can be used `boost::heap::d_ary_heap` as a 4-ary heap. The PR adds `SearchHeap` that internally uses `boost::heap::d_ary_heap` with `boost::heap::arity<4>`. In general on a small dataset like `bayern-latest` the difference between 2-ary and 4-ary heaps is not statistically relevant, but the difference wrt to `BinaryHeap` is for 1000 queries `http://localhost:5000/route/v1/driving/11.575395,48.137132;9.932966,49.792450`
```
	Welch Two Sample t-test

data:  before$V1 and after$V1
t = 13.672, df = 1274.7, p-value < 2.2e-16
alternative hypothesis: true difference in means is not equal to 0
95 percent confidence interval:
 0.001313073 0.001753019
sample estimates:
  mean of x   mean of y 
0.010305615 0.008772569 
```

Data summary of MLD routing  with `BinaryHeap`  (red on histogram) is
```
    Min.  1st Qu.   Median     Mean  3rd Qu.     Max. 
0.007715 0.008152 0.008602 0.010310 0.011940 0.039360 
```
and  for the new `SearchHeap` (blue on histogram) is
```
    Min.  1st Qu.   Median     Mean  3rd Qu.     Max. 
0.007814 0.008083 0.008232 0.008773 0.008688 0.014980 
```

heaps

## Tasklist
 - [ ] review
 - [ ] adjust for comments

## Requirements / Relations

### DIFF
--- a/include/contractor/contractor_heap.hpp
+++ b/include/contractor/contractor_heap.hpp
@@ -1,7 +1,7 @@
 #ifndef OSRM_CONTRACTOR_CONTRACTOR_HEAP_HPP_
 #define OSRM_CONTRACTOR_CONTRACTOR_HEAP_HPP_
 
-#include "util/binary_heap.hpp"
+#include "util/query_heap.hpp"
 #include "util/typedefs.hpp"
 #include "util/xor_fast_hash_storage.hpp"
 
@@ -18,11 +18,11 @@ struct ContractorHeapData
     bool target = false;
 };
 
-using ContractorHeap = util::BinaryHeap<NodeID,
-                                        NodeID,
-                                        EdgeWeight,
-                                        ContractorHeapData,
-                                        util::XORFastHashStorage<NodeID, NodeID>>;
+using ContractorHeap = util::QueryHeap<NodeID,
+                                       NodeID,
+                                       EdgeWeight,
+                                       ContractorHeapData,
+                                       util::XORFastHashStorage<NodeID, NodeID>>;
 
 } // namespace contractor
 } // namespace osrm
--- a/include/customizer/cell_customizer.hpp
+++ b/include/customizer/cell_customizer.hpp
@@ -3,7 +3,7 @@
 
 #include "partition/cell_storage.hpp"
 #include "partition/multi_level_partition.hpp"
-#include "util/binary_heap.hpp"
+#include "util/query_heap.hpp"
 
 #include <tbb/enumerable_thread_specific.h>
 
@@ -24,7 +24,7 @@ class CellCustomizer
 
   public:
     using Heap =
-        util::BinaryHeap<NodeID, NodeID, EdgeWeight, HeapData, util::ArrayStorage<NodeID, int>>;
+        util::QueryHeap<NodeID, NodeID, EdgeWeight, HeapData, util::ArrayStorage<NodeID, int>>;
     using HeapPtr = tbb::enumerable_thread_specific<Heap>;
 
     CellCustomizer(const partition::MultiLevelPartition &partition) : partition(partition) {}
--- a/include/engine/search_engine_data.hpp
+++ b/include/engine/search_engine_data.hpp
@@ -4,7 +4,7 @@
 #include <boost/thread/tss.hpp>
 
 #include "engine/algorithm.hpp"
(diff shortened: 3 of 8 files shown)


include/util/search_heap.hpp
```diff
@@ -0,0 +1,131 @@
...
+        std::for_each(data.begin(), data.end(), [](auto &node) { node.handle = HeapHandle(); });
+        heap.clear();
+    }
+
+    void DecreaseKey(NodeID node, WeightType weight)
+    {
+        BOOST_ASSERT(WasInserted(node));
+        auto reference = index.find(node)->second;
+        reference->weight = weight;
+        heap.increase(reference->handle, std::make_pair(weight, reference));
```
daniel-j-h: because it implements a max heap?

daniel-j-h:
:+1: 

What about our existing heap usages e.g. in the CH code?
I think we have some Map and Array storage for heaps in use somewhere.

include/util/search_heap.hpp
oxidase: yes, `greater` instead of `less` and `increase` instead of `decrease`

oxidase:
It is possible to use boost heaps, but i don't want to make a performance regression for CH algorithms, as performance depends on memory access patterns. Basically for CH-routing only `UnorderedMapStorage` is used, but in contractor also `XORFastHashStorage` is used.

oxidase:
Rebased to the current master and checked performance with #3857 The current  implementation of the binary heap is nearly the same as `boost::heap::d_ary_heap`
compare_heaps

@TheMarex as the 4-ary heap does not bring significant performance improvements, is it still needed? I see only benefit in using generalized library code instead of a custom version

I will continue with checking variants for [`DataIndex`](https://github.com/Project-OSRM/osrm-backend/blob/ca3a0342de0eaae9d22ee0ff42bd552e61d3e4b4/include/util/search_heap.hpp#L24)

TheMarex:
@oxidase I suspect the current runtime is dominated by access to the `DataIndex` hence we don't see any improvements.

A comparison with `ArrayIndex` might be worthwhile. However if we already pay the price for a full array of size `|V|` for each heap we might actually go a step further and remove the indirection altogether.

Instead of having a separate index for translating global `NodeID` to a heap internal (continuous) ID and then use that for indexing the data, we could have a fixed pre-alloced array for that which is indexed by `NodeID` directly.

That said this would be a bit of work again and would yield an implementation with more memory usage. I'm fine with tabling again, since this is definitely to the catagory `quick win` anymore.

oxidase:
@TheMarex `ArrayIndex` is somewhat not suitable already on Germany-size graphs, because almost all time spent in `memset_avx2_erms` by clearing heaps. I have reused generation arrays and here some results 
compare_hashes

for the MLD algorithm and germany-latest OSM file
* `std::unordered_map`
* [`tsl::hopscotch_map`](https://github.com/Tessil/hopscotch-map)
* [`google::sparse_map`](https://github.com/sparsehash/sparsehash)
* `generative_array`
* CH-timing for reference

I think it is ready for review, but it brings almost no value, so it is ok to reject the PR as not needed.

oxidase:
@TheMarex after lengthy performance investigation i think it is better  to close PR without merging, because it is  brings no performance improvement but increases memory consumption in places like in master `const Key key = static_cast<Key>(heap.size());` with `Key = uint32_t` and in branch `reference->handle = heap.push(std::make_pair(weight, reference));` with `handle = void *`

What i can do:
* remove `heap[0] ` element
* make  `CheckHeap` optional

oxidase:
@TheMarex please could you check the PR? Summary:
* removed `heap[0] element`
* removed `CheckHeap`
* heap changed to `boost::heap::d_ary_heap`, but any mutable heap structure can be used
* renamed `BinaryHeap` to `QueryHeap`

I did not change the index storage as it is performance critical and could not find a better  solution than the current one with `XORFastHashStorage`. The performance with 4-ary heap is  the same as with `BinaryHeap` compare_hashes3

PR #3822, opened 2017-03-14
Downgrade log levels for SHM detach messages

danpat:
# Issue

We're logging a warning when `osrm-datastore` asks clients to detach from shared memory segments during a data update on non-Linux platforms.

```
[warn] Shared memory support for non-Linux systems does not wait for clients to dettach. Going to sleep for 50ms.
```

This isn't really a warning, it's expected and known behaviour.  This PR downgrades these messages to `logDEBUG` so that there aren't any scary warnings generated for `Release` builds under normal operating conditions.

## Tasklist
 - [x] review
 - [x] adjust for comments

### DIFF
--- a/include/storage/shared_memory.hpp
+++ b/include/storage/shared_memory.hpp
@@ -151,7 +151,7 @@ class SharedMemory
 #else
     void WaitForDetach()
     {
-        util::Log(logWARNING)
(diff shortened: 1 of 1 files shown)


TheMarex:
@danpat this needs a `clang-format` then it's good to go.

Issue #3818, opened 2017-03-13
Remove unconnected boundary nodes


TheMarex:
Below an algorithm to remove all unconnected boundary nodes for every cell, once we have a `CellID` assignment over every level:

- Starting with the highest level `l`: For every partition array (`NodeID` -> `CellID` translation) on the edge-based-graph:
  - For every `node` in the graph:
     - Check all `target` nodes:
         - If `target` is in a different cell, save `target` as `witness` that `node` is a boundary node.
         - If `target` is from the same cell: Check if it is a forward/backward edge. If forward, `node` is a source node, if `backward` then `node` is a target node.
    - If we detect a `node` on level `l` that is a boundary node, but not marked as source or target:
       - assign `node` to the same cells as the `witness` on all levels ` <= l`

/cc @oxidase

TheMarex:
Landed with #3826

Issue #3780, opened 2017-03-06
Enforce forward/backward node to be in the same level 1 cell


TheMarex:
We need to enforce that the forward/backward node of every edge-based node are in the same level 1 cell. We need this to compute the correct query level for every node during the search using:

```
min(highestDifferentLevel(s, node), highestDifferentLevel(t, node))
```
If forward and backward node are in the same level 1 cell then `highestDifferentLevel(s, node)` will be the same for every start node `s` in the heap.

Placing the forward/backward node in the same heap is important for the search to automatically select the correct direction in which to leave a street.

This needs some insurances in the partition code and the id translation.

/cc @oxidase

TheMarex:
I tried this approach in #3826 and it creates problems around unconnected boundary nodes: Frequently either the forward or backward will be an unconnected boundary node and we can't resolve this easily. For now we are going to handle this by allowing arbitrary start/target cells in the query, which makes computing the query level a little bit more expensive. We are not worries about that though since it is a fairly optimized operation (xor and builtin).

PR #3826, opened 2017-03-15
Remove unconnected boundary nodes from partition

TheMarex:
# Issue

Addresses ~#3780~ and #3818 and refactors assigning IDs from node-based edges to edge-based edges.

We now chose the `BisectionID` of the side that would induce the least amount of boundary edges (since we are assigning `BisectionID`s and not `CellID`s this is only heuristic, since two different `BisectionID`s can later be contained in the same cell).

We also remove all unconnected boundary nodes.

## Tasklist
 - [x] review
 - [x] adjust for comments


TheMarex:
@oxidase ah sorry my bad, renamed the PR title. Going to squash the commits before merging, the reverts are just there for transparency.

oxidase:
@TheMarex it is really hard to compare heuristics, because non-covered edge cases could give different results. I checked only two cases:
*  `./osrm-partition bayern-latest.osrm` on master and the branch: number of unconnected nodes goes from 0,2,3,4 to 18,19 (numbers are different between runs, because partitioning is not stable). 
* `./osrm-partition germany-latest.osrm` number of unconnected nodes goes from 43,45 to 89,94. 

These numbers are really small (~10^-6 part of all nodes), so the difference is almost insignificant. 

I think it is better to use heuristic as simple as possible or even better avoid any kind by partitioning the edge-based graph, but don't know if it must be prioritized because of  so minor win.

TheMarex:
@oxidase I will try your test datasets. This method should guarantee exactly zero unconnected boundary nodes.

TheMarex:
@oxidase fixed a bug in the code that breaks the partition, which was causing this weirdness.

Did some investigation around the increased border nodes: Picking the minimum border edge `BisectionID` does actually decrease quality. I tried a few combinations and surprisingly picking `u` for the forward and `v` for the backward node works the best.

My hypothesis is this is because this comes closest to a partitioning on the actual embedding. (forward node would be represented by the start of the edge and the backward node by the target)

PR #3829, opened 2017-03-16
Builds Release / Debug Docker image; set via Docker CI

daniel-j-h:
For https://github.com/Project-OSRM/osrm-backend/pull/3804#issuecomment-286996352. Lets see if this works.


daniel-j-h:
Seems like `$DOCKER_TAG` is empty when building.

Issue #2414, opened 2016-05-17
Stable GCC 6.1 + Intel TBB 4.4/20160413 crashing osrm-extract


daniel-j-h:
```
Thread 15 "osrm-extract" received signal SIGSEGV, Segmentation fault.
[Switching to Thread 0x7fffd4608700 (LWP 18123)]
tbb::internal::generic_scheduler::allocate_task (this=this@entry=0x7ffff7fb3e00, number_of_bytes=number_of_bytes@entry=8, 
    parent=parent@entry=0x0, context=context@entry=0x7ffff5f2f880 <tbb::internal::the_dummy_context>)
    at /home/daniel/gcc-dev-6.1/sources/tbb44_20160413oss/./src/tbb/scheduler.cpp:315
315                 my_free_list = t->prefix().next;
(gdb) bt
#0  tbb::internal::generic_scheduler::allocate_task (this=this@entry=0x7ffff7fb3e00, number_of_bytes=number_of_bytes@entry=8, 
    parent=parent@entry=0x0, context=context@entry=0x7ffff5f2f880 <tbb::internal::the_dummy_context>)
    at /home/daniel/gcc-dev-6.1/sources/tbb44_20160413oss/./src/tbb/scheduler.cpp:315
#1  0x00007ffff5d1771b in tbb::internal::generic_scheduler::generic_scheduler (this=this@entry=0x7ffff7fb3e00, m=...)
    at /home/daniel/gcc-dev-6.1/sources/tbb44_20160413oss/./src/tbb/scheduler.cpp:100
#2  0x00007ffff5d191b8 in tbb::internal::custom_scheduler<tbb::internal::IntelSchedulerTraits>::custom_scheduler (m=..., this=0x7ffff7fb3e00)
    at /home/daniel/gcc-dev-6.1/sources/tbb44_20160413oss/./src/tbb/custom_scheduler.h:59
#3  tbb::internal::custom_scheduler<tbb::internal::IntelSchedulerTraits>::allocate_scheduler (m=...)
    at /home/daniel/gcc-dev-6.1/sources/tbb44_20160413oss/./src/tbb/custom_scheduler.h:115
#4  0x00007ffff5d17fc5 in tbb::internal::allocate_scheduler (m=...)
    at /home/daniel/gcc-dev-6.1/sources/tbb44_20160413oss/./src/tbb/scheduler.cpp:42
#5  tbb::internal::generic_scheduler::create_worker (m=..., index=index@entry=1)
[...]

PR #3833, opened 2017-03-17
Configure custom TBB builds via TBB_INSTALL_DIR

oxidase:
# Issue

Use TBB_INSTALL_DIR as a prefix for locally installed TBB.

TBB [2017_U5](https://github.com/01org/tbb/archive/2017_U5.tar.gz) can be used for  #2414 until recent version appear in a local distributive upstream

## Tasklist
 - [x] review
 - [x] adjust for comments

### DIFF
--- a/cmake/FindTBB.cmake
+++ b/cmake/FindTBB.cmake
@@ -57,7 +57,7 @@
 if (WIN32)
     # has em64t/vc8 em64t/vc9
     # has ia32/vc7.1 ia32/vc8 ia32/vc9
-    set(_TBB_DEFAULT_INSTALL_DIR "C:/Program Files/Intel/TBB" "C:/Program Files (x86)/Intel/TBB")
+    set(_TBB_DEFAULT_INSTALL_DIR "C:/Program Files/Intel/TBB")
     set(_TBB_LIB_NAME "tbb")
     set(_TBB_LIB_MALLOC_NAME "${_TBB_LIB_NAME}malloc")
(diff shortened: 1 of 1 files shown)


daniel-j-h:
Looks good!

@TheMarex do you know where we got the FindTBB module from? This should probably be fixed upstream (if we care).

TheMarex:
@daniel-j-h my memory is hazy but I think it included a google search and some c&p from a random project. Unless we are trying to upstream this to `cmake` I don't think there is a real upstream.

oxidase:
@daniel-j-h there is no `FindTBB.cmake` in cmake repository. Mainly c&p [549 times](https://github.com/search?utf8=%E2%9C%93&q=filename%3Afindtbb.cmake+%22Hannes+Hofmann%22&type=Code&ref=searchresults) out of [1310 matches](https://github.com/search?utf8=%E2%9C%93&q=filename%3Afindtbb.cmake+&type=Code&ref=searchresults)

TheMarex:
This seems to break TBB on travis. Other then that 👍

oxidase:
@TheMarex the original problem is at https://github.com/Project-OSRM/osrm-backend/blob/e166246eb887ab492e0871d310edd4362329cc67/cmake/FindTBB.cmake#L105 and later https://github.com/Project-OSRM/osrm-backend/blob/e166246eb887ab492e0871d310edd4362329cc67/cmake/FindTBB.cmake#L154

Basically it is wrong because it must be a one-value variable but not list.
Will change to "/usr" for linux and "C:/Program Files/Intel/TBB" for windows

PR #3834, opened 2017-03-17
Expose algorithms option in node bindings and integrate new toolchain

TheMarex:
# Issue

Integrates algorithm flags into node bindings and bundles the new toolchain for MLD. This addresses https://github.com/Project-OSRM/osrm-backend/issues/3846.

## Tasklist

 - [x] Merge #3768
 - [x] review
 - [x] adjust for comments

## Requirements / Relations

Based on #3768.


.travis.yml
```diff
@@ -7,21 +7,29 @@ git:
...
+  yarn: true
   ccache: true
   apt: true
   directories:
     - test/cache
 
 env:
   global:
+   - secure: "hk+32aXXF5t1ApaM2Wjqooz3dx1si907L87WRMkO47WlpJmUUU/Ye+MJk9sViH8MdhOcceocVAmdYl5/WFWOIbDWNlBya9QvXDZyIu2KIre/0QyOCTZbrsif8paBXKIO5O/R4OTvIZ8rvWZsadBdmAT9GSbDhih6FzqXAEgeIYQ="
+   - secure: "VE+cFkseFwW4jK6XwkP0yW3h4DixPJ8+Eb3yKcchGZ5iIJxlZ/8i1vKHYxadgPRwSYwPSB14tF70xj2OmiT2keGzZUfphmPXinBaLEhYk+Bde+GZZkoSl5ND109I/LcyNr0nG9dDgtV6pkvFchgchpyP9JnVOOS0+crEZlAz0RE="
```
daniel-j-h: Where did you get those from and how are they getting used? for node pre gyp publishing / aws credentials?

.travis.yml
```diff
@@ -7,21 +7,29 @@ git:
...
   email: false
 
 branches:
   only:
     - master
+    # enable building tags
+    - /^v\d+\.\d+(\.\d+)?(-\S*)?$/
 
 cache:
+  yarn: true
```
daniel-j-h: Ah I guess this is https://blog.travis-ci.com/2016-11-21-travis-ci-now-supports-yarn - is yarn worth it locally?

.travis.yml
```diff
@@ -190,7 +208,8 @@ install:
   - cmake .. -DCMAKE_BUILD_TYPE=${BUILD_TYPE}
   - make --jobs=${JOBS}
   - popd
-  - npm run build-api-docs
+  # building docs only works with npm3+ not with yarn or npm2
+  #- yarn run docs
```
daniel-j-h: Why? And can't we just use npm here, then?

docs/releasing.md
```diff
@@ -32,21 +32,25 @@ We may introduce forward-compatible changes: query parameters and response prope
...
-8. Proceede with the `node-osrm` release as [outlined in the repository](https://github.com/Project-OSRM/node-osrm/blob/master/docs/releasing.md).
+2. Make sure `CHANGELOG.md` is up to date.
+3. Make sure the OSRM version in `CMakeLists.txt` is up to date
+4. Make sure the `package.json` is up to date.
+5. Make sure all tests are passing (e.g. Travis CI gives you a :thumbs_up:)
+6. Use an annotated tag to mark the release: `git tag vx.y.z -a` Body of the tag description should be the changelog entries.
+7. Use `npm run build-api-docs` to generate the API documentation.  Copy `build/docs/*` to `https://github.com/Project-OSRM/project-osrm.github.com` in the `docs/vN.N.N/api` directory
+8. Push tags and commits: `git push; git push --tags`
 9. If not a release-candidate: Write a mailing-list post to osrm-talk@openstreetmap.org to announce the release
+10. Wait until the travis build has been complated and check if the node binaries were published by doing:
```
daniel-j-h: completed

include/nodejs/node_osrm_support.hpp
```diff
@@ -138,6 +139,34 @@ inline engine_config_ptr argumentsToEngineConfig(const Nan::FunctionCallbackInfo
...
+        {
+            Nan::ThrowError("algorithm option must be one of 'CH', 'CoreCH', or 'MLD'.");
+            return engine_config_ptr();
+        }
+    }
+    else if (!algorithm->IsUndefined())
+    {
+        Nan::ThrowError("algorithm option must be a string and one of 'CH', 'CoreCH', or 'MLD'.");
+        return engine_config_ptr();
+    }
```
daniel-j-h: Hm, using the maybe returned from Get is fine this way? as in, this is the same as

    maybe = Get();
    if (!maybe) err;

?

test/nodejs/route.js
```diff
@@ -14,6 +16,30 @@ test('route: routes Berlin', function(assert) {
...
+    assert.plan(5);
+    var osrm = new OSRM({path: berlin_corech_path, algorithm: 'CoreCH'});
+    osrm.route({coordinates: [[13.43864,52.51993],[13.415852,52.513191]]}, function(err, route) {
+        assert.ifError(err);
+        assert.ok(route.waypoints);
+        assert.ok(route.routes);
+        assert.ok(route.routes.length);
+        assert.ok(route.routes[0].geometry);
+    });
+});
```
daniel-j-h: Urgh please merge into master asap, I'm already working on changing the tests to monaco

.travis.yml
```diff
@@ -138,11 +146,12 @@ before_install:
     if [[ "${TRAVIS_OS_NAME}" == "osx" ]]; then
       export JOBS=$((`sysctl -n hw.ncpu` + 1))
       sudo mdutil -i off /
+      npm install -g yarn
```
daniel-j-h: We need to update the yarn.lock file every time we update package.config now?

.travis.yml
TheMarex: Yeah factor 3-5 faster then npm.

.travis.yml
TheMarex: Yup they set `node_pre_gyp_{secretKey,KeyId}`.

.travis.yml
TheMarex: No it is updated automatically by yarn when you add a package using `yarn add package`. You will have to commit it though.

.travis.yml
TheMarex: This needs fixes in the docs tool. I don't see the point of building the docs for every build job. It almost takes a minute and would block us from using a sensible package manager like yarn.

include/nodejs/node_osrm_support.hpp
TheMarex: Hrm the docs say that the result from `Object::Get` is a `MaybeLocal<>` but that type does not have a `operator->`. For a `Local<>` calling `operator->` is not safe since it would dereference a `nullptr`. So yes this would need a check.

package.json
```diff
@@ -30,19 +17,16 @@
...
     "install": "node-pre-gyp install --fallback-to-build=false"
   },
   "repository": {
     "type": "git",
     "url": "https://github.com/Project-OSRM/osrm-backend.git"
   },
-  "author": "",
-  "license": "ISC",
+  "author": "Project OSRM Team",
+  "license": "BSD 2-Clause",
```
oxidase: guess `ISC` was wrong originally?

yarn.lock
```diff
@@ -0,0 +1,6179 @@
+# THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.
```
oxidase: Is it really necessary to commit the auto-generated file? Guess it is better to add it to `.gitignore` otherwise will be always local changes and conflicts.

TheMarex: Yes this basically ensure that `yarn` resolves package versions the same for everyone installing osrm. One advantage of the `yarn` approach with the `lock` file is to enable caching, e.g. on Travis.

package.json
TheMarex: Yes, this is just the default license and nobody changed it.

TheMarex:
@daniel-j-h @oxidase hopefully I addressed all concerns. Let me know if this is good to merge.

daniel-j-h:
:+1:

PR #3841, opened 2017-03-20
Conditional turn restriction support

karenzshea:
# Issue

https://github.com/Project-OSRM/osrm-backend/issues/3414

Initial pass at this will only support conditional turns not conditional speed limits.

So far, this branch contains updates for
- conditional turn parsing by including an extra filter in osmium key parsing
- writing found conditional turns to disk (and removing unconditional turn restrictions from being written to disk)
- restriction file deserialization during update

## Tasklist
 - [x] Update edge based graph factory to index turns by internal node ids
-  [x] Expose `internal_to_external_node_map` to `UpdateTurnPenalties` function
-  [x] Lookup turn penalties and turn restrictions based on internal node ids
-  [x] Add extra check in `UpdateTurnPenalties` for restricted turns
-  [x] Store conditional turn conditions in a compact format, rather than strings
-  [x] Add turn restriction validation (is this conditional turn restriction relevant _now_) and update turn penalties accordingly
 - [x] write support code in cucumber for conditional turns
 - [x] store in-memory conditional turn restrictions during contraction as a hash map for faster lookups
 - [x] add cucumber cases
 - [x] review
 - [x] adjust for comments

## Requirements / Relations
After this first pass is done, clean up tasks:
- investigate how often applications of turn restrictions that expire immediately OR non-application of turn restrictions that will be shortly thereafter valid occur
- normalize time range conditions to UTC during extraction

cc @TheMarex @oxidase


include/conditionals/extract-conditionals.hpp
```diff
@@ -0,0 +1,401 @@
...
+    std::string condition;
+};
+
+struct LocatedConditionalSpeedLimit
+{
+    osmium::Location location;
+    ConditionalSpeedLimit speed_limit;
+};
+
+auto ParseGlobalArguments(int argc, char *argv[]);
```
TheMarex: These should probably not be exposed globally, because they are just there to parse CLI parameters. This is important for the conditionals tool, but not for anything just using the functions.

karenzshea:
@oxidase @daniel-j-h If either of you have a time to look at what's going on with inclusion of `Shapefile` into cmake outside of just the `BUILD_TOOLS` building, I would be grateful! :pray: 

In my last commit I made a flimsy pass at pulling out the `find_package(Shapefile)` results and exposing the paths to `UPDATER_LIBRARIES` and `CONTRACTOR_LIBRARIES` but I'm almost certainly missing something. Linking errors abound to all shapefile provided functions. Currently `shapefil.h` is only being included into the `src/util/timezones.cpp` file.

CMakeLists.txt
```diff
@@ -522,8 +522,8 @@ else()
   ENDIF()
 
   find_package(Shapefile) # package libshp-dev
-  if(SHAPEFILE_FOUND)
-    add_dependency_includes(${LIBSHAPEFILE_LIBRARY})
+  if(NOT SHAPEFILE_FOUND)
+    message(FATAL_ERROR "Shapefile library (libshp-dev) was not found")
```
daniel-j-h: this is the same as find_package(P REQUIRED) (?)

CMakeLists.txt
```diff
@@ -682,7 +684,7 @@ target_link_libraries(osrm_store ${STORAGE_LIBRARIES})
 
 # BUILD_COMPONENTS
 add_executable(osrm-components src/tools/components.cpp $<TARGET_OBJECTS:UTIL>)
-target_link_libraries(osrm-components ${TBB_LIBRARIES} ${BOOST_BASE_LIBRARIES})
+target_link_libraries(osrm-components ${TBB_LIBRARIES} ${BOOST_BASE_LIBRARIES} ${LIBSHAPEFILE_LIBRARY})
```
daniel-j-h: why does components need the shapefile lib? transitive dep?

CMakeLists.txt
```diff
@@ -602,12 +602,12 @@ set(BOOST_ENGINE_LIBRARIES
...
 
 # Binaries
-target_link_libraries(osrm-datastore osrm_store ${Boost_PROGRAM_OPTIONS_LIBRARY})
+target_link_libraries(osrm-datastore osrm_store ${Boost_PROGRAM_OPTIONS_LIBRARY} ${LIBSHAPEFILE_LIBRARY})
 target_link_libraries(osrm-extract osrm_extract ${Boost_PROGRAM_OPTIONS_LIBRARY})
 target_link_libraries(osrm-partition osrm_partition ${Boost_PROGRAM_OPTIONS_LIBRARY})
 target_link_libraries(osrm-customize osrm_customize ${Boost_PROGRAM_OPTIONS_LIBRARY})
 target_link_libraries(osrm-contract osrm_contract ${Boost_PROGRAM_OPTIONS_LIBRARY})
-target_link_libraries(osrm-routed osrm ${Boost_PROGRAM_OPTIONS_LIBRARY} ${OPTIONAL_SOCKET_LIBS} ${ZLIB_LIBRARY})
+target_link_libraries(osrm-routed osrm ${Boost_PROGRAM_OPTIONS_LIBRARY} ${LIBSHAPEFILE_LIBRARY} ${OPTIONAL_SOCKET_LIBS} ${ZLIB_LIBRARY})
```
daniel-j-h: same here, why does routed need shapefile lib? and datastore above?

oxidase:
@karenzshea i have pushed changes to `CMakeLists.txt`, but actually i don't like to have Shapelib as a requirement. You can make optional by changing `FATAL_ERROR` to `WARNING`,  define compilation flag `-DSHAPELIB_FOUND` and use it in the code later.

Also there is a problem with `timezones.cpp` in the "util lib" that is not really a library but a set of object files`$<TARGET_OBJECTS:UTIL>`. It requires to add everywhere `${LIBSHAPEFILE_LIBRARY}`. I would suggest to move `LoadLocalTimesRTree` into a header file. It will make configuration with `SHAPELIB_FOUND` easier.

You can kick out my commit, it makes all build jobs red because libshp-dev is not installed. Only mason build should be green.

CMakeLists.txt
oxidase: yes, but i don't think it must be a requirement. better to have a warning and conditional compilation without time-based conditionals

CMakeLists.txt
oxidase: util lib is not a library, so it is needed  everywhere  with `$<TARGET_OBJECTS:UTIL>`

it must be changed for sure

CMakeLists.txt
```diff
@@ -630,11 +635,15 @@ set(CUSTOMIZER_LIBRARIES
     ${MAYBE_RT_LIBRARY}
     ${MAYBE_COVERAGE_LIBRARIES})
 set(UPDATER_LIBRARIES
+    ${BZIP2_LIBRARIES}
```
TheMarex: BZIP2 and EXPAT should only be needed in `EXTRACTOR` after the header refactor.

CMakeLists.txt
```diff
@@ -630,11 +635,15 @@ set(CUSTOMIZER_LIBRARIES
...
     ${BOOST_BASE_LIBRARIES}
     ${CMAKE_THREAD_LIBS_INIT}
+    ${EXPAT_LIBRARIES}
     ${TBB_LIBRARIES}
+    ${LIBSHAPEFILE_LIBRARY}
```
TheMarex: This should be moved to `UTIL_LIBRARIES`

CMakeLists.txt
TheMarex: Yes, this should come from `UTIL_LIBARIES` eventually.

CMakeLists.txt
TheMarex: Same, this is a transitive dependency and should come from the `UTIL_LIBRARIES` variable.

CMakeLists.txt
```diff
@@ -521,6 +521,11 @@ else()
     ENDIF()
   ENDIF()
 
+  find_package(Shapefile) # package libshp-dev
+  if(NOT SHAPEFILE_FOUND)
```
TheMarex: We want to make the shapefile dependency optional. For this we need to ensure a few things:

1. We don't try to link against a missing library when it's not there
2. The code that uses shapefile is not compiled when the library is not found.

We get 1. by making the linker flags for shapefiles optional:

```
find_package(Shapefile)
if (SHAPEFILE_FOUND)
  set(MAYBE_SHAPEFILE_LIBRARIES "${SHAPEFILE_LIBRARIES}")
  target_include_directories(UTIL PRIVATE ${LIBSHAPEFILE_INCLUDE_DIR}) 
else()
  set(MAYBE_SHAPEFILE_LIBRARIES "")
endif()
```

This creates a variable that is just empty in case there is no shapefile library found, and only add the header include in case we find the library.

To get 2. we need to tell the C++ code whether it is allowed to use anything that references shapefiles. The typical technique to do that is use defines like `#define FOO_BAR`. Luckily we can pass these options straight to the compiler like:

```
g++ -D FOO_BAR test.cpp -o test
```
Conveniently cmake can generate those flags for us. [Here](https://github.com/Project-OSRM/osrm-backend/blob/master/unit_tests/CMakeLists.txt#L93) is an example of how we do this to pass C++ the location of the test directory.

In our case that would be something like:

```
target_compile_definitions(UTIL PRIVATE COMPILE_DEFINITIONS ENABLE_SHAPEFILE)
```

include/conditionals/extract-conditionals.hpp
```diff
@@ -0,0 +1,384 @@
+#include "util/conditional_restrictions.hpp"
```
daniel-j-h: include guard missing

include/extractor/edge_based_graph_factory.hpp
```diff
@@ -50,13 +50,13 @@ namespace lookup
...
-    OSMNodeID via_id;
-    OSMNodeID to_id;
+    NodeID from_id;
+    NodeID via_id;
+    NodeID to_id;
 };
 #pragma pack(pop)
 static_assert(std::is_trivial<TurnIndexBlock>::value, "TurnIndexBlock is not trivial");
-static_assert(sizeof(TurnIndexBlock) == 24, "TurnIndexBlock is not packed correctly");
+static_assert(sizeof(TurnIndexBlock) < 24, "TurnIndexBlock is not packed correctly");
```
daniel-j-h: Why the change here? Byte size checks here should always check for equality.

include/extractor/extractor.hpp
```diff
@@ -56,14 +56,18 @@ class Extractor
   private:
     ExtractorConfig config;
 
+    std::vector<TurnRestriction> ParseOSMData(ScriptingEnvironment &scripting_environment,
+                                              const unsigned number_of_threads);
```
daniel-j-h: I think we can pull out the number of threads from the type's extractor config, no need to pass it in here

https://github.com/Project-OSRM/osrm-backend/blob/e568b600f0c4ad5a2cd9b8f97868ef9b1b7fc6b2/include/extractor/extractor_config.hpp#L106

include/extractor/io.hpp
```diff
@@ -96,6 +99,51 @@ inline void write(const boost::filesystem::path &path, const SegmentDataContaine
...
     writer.WriteFrom(segment_data.datasources);
 }
+
+// read/write for conditional turn restrictions file
+inline void read(const boost::filesystem::path &path, std::vector<InputRestrictionContainer> &restrictions)
+{
+    const auto fingerprint = storage::io::FileReader::VerifyFingerprint;
+    storage::io::FileReader reader{path, fingerprint};
+
+    auto num_indices = reader.ReadElementCount32();
```
daniel-j-h: I would make all counts 64 bits in your new code already; see also https://github.com/Project-OSRM/osrm-backend/pull/3848

include/extractor/io.hpp
```diff
@@ -96,6 +99,51 @@ inline void write(const boost::filesystem::path &path, const SegmentDataContaine
...
+        reader.ReadInto(is_only);
+        for (auto &c : res.restriction.condition)
+        {
+            reader.ReadInto(c.modifier);
+            reader.DeserializeVector(c.times);
+            reader.DeserializeVector(c.weekdays);
+            reader.DeserializeVector(c.monthdays);
+        }
+        res.restriction.flags.is_only = is_only;
+        restrictions.push_back(res);
```
daniel-j-h: Above the code resizes the `restrictions` vector to size `num_indices`, but here you `push_back` into it.

I think what we want is either
- To `.reserve(num_indices)` above (tells the vector we expect roughly `num_incices` elements) and then `push_back` here or
- To `.resize(num_indices)` above and assign to `restrictions[i]` here

Think of `reserve` as a hint for vector to allocate memory more efficiently. In contrast, `resize(n)` adds `n` elements to the vector (default constructed).

daniel-j-h: You can `std::move(res)` here to move the restriction container into the vector. At the moment we make a copy of res and then destroy res immediately after. Similar, you could put the `InputRestrictionContainer` above outside of the loop, otherwise we create an instance of it for every iteration.

include/extractor/io.hpp
```diff
@@ -96,6 +99,51 @@ inline void write(const boost::filesystem::path &path, const SegmentDataContaine
...
+    restrictions.resize(num_indices);
+    while (num_indices > 0)
+    {
+        InputRestrictionContainer res;
+        bool is_only;
+        reader.ReadInto(res.restriction.via);
+        reader.ReadInto(res.restriction.from);
+        reader.ReadInto(res.restriction.to);
+        reader.ReadInto(is_only);
+        for (auto &c : res.restriction.condition)
```
daniel-j-h: We never read into `res.restriction.condition` (?)

include/storage/io.hpp
```diff
@@ -104,6 +104,13 @@ class FileReader
         }
     }
 
+    void ReadInto(std::string &target)
+    {
+        auto len = ReadElementCount32();
```
daniel-j-h: Hm.. I would say 64 bit here for consistency. But this would also mean we require 8 bytes per string. These strings are osm strings, no? So it should be bound by 255 * 4 bytes (max. 4 bytes in utf-8 coded code point), and 16 bit should suffice?

include/util/timezones.hpp
```diff
@@ -0,0 +1,16 @@
+#include <boost/geometry.hpp>
```
daniel-j-h: include guard

include/util/timezones.hpp
```diff
@@ -0,0 +1,16 @@
...
+using point_t = boost::geometry::model::
+    point<int32_t, 2, boost::geometry::cs::spherical_equatorial<boost::geometry::degree>>;
+using polygon_t = boost::geometry::model::polygon<point_t>;
+using box_t = boost::geometry::model::box<point_t>;
+using rtree_t =
+    boost::geometry::index::rtree<std::pair<box_t, size_t>, boost::geometry::index::rstar<8>>;
+using local_time_t = std::pair<polygon_t, struct tm>;
+
+std::function<struct tm(const point_t &)> LoadLocalTimesRTree(const std::string &tz_shapes_filename,
+                                                              std::time_t utc_time);
```
daniel-j-h: dup. with timezoner class above?

include/extractor/extractor.hpp
karenzshea: Looks like a separate parameter is used here because at the call site extractor takes the min of a `recommended_num_threads` and the config provided num of threads. 

Do we want to change this?

include/conditionals/extract-conditionals.hpp
karenzshea: Added!

include/extractor/io.hpp
karenzshea: :pray:

include/storage/io.hpp
karenzshea: I'm using this particular method for osm strings, but it's not unlikely that this method could be reused in the future for strings other than osm strings. Would a 16 bit size still work?

include/extractor/extractor.hpp
daniel-j-h: Ah okay then we're fine here!

include/storage/io.hpp
daniel-j-h: Depends how large the strings will be. 

```c++
numeric_limits<uint8_t>::max(); // -> 255
numeric_limits<uint16_t>::max(); // -> 65535
numeric_limits<uint32_t>::max(); // -> 4294967295
```

OSM guarantees for a max of 256*4 bytes per string, so a `uint16_t` would be fine here. Other strings not coming from osm might need more.

src/extractor/restriction_parser.cpp
```diff
@@ -160,7 +173,42 @@ RestrictionParser::TryParse(const osmium::Relation &relation) const
...
+            {
+                std::vector<util::OpeningHours> parsed = util::ParseOpeningHours(p.condition);
+                if (!parsed.empty())
+                    restriction_container.restriction.condition = std::move(parsed);
+            }
+        }
+    }
+
+    // push back a copy of turn restriction
+    parsed_restrictions.push_back(boost::make_optional(restriction_container));
```
daniel-j-h: I was just about to say, can we `make_optional(move(resr_cont))`, and then checked optional's source:

http://www.boost.org/doc/libs/1_63_0/boost/optional/optional.hpp

Looks like the overload for `make_optional(Args&& ...)` is missing. It's in C++17's stdlib, though:

http://en.cppreference.com/w/cpp/utility/optional/make_optional

daniel-j-h: okay, bug, ticketed boostorg/optional 30

if you want to avoid a copy here you can do

    optional<YourType> tmp{std::move(v)};
    r.push_back(std::move(tmp));

src/extractor/restriction_parser.cpp
```diff
@@ -160,7 +173,42 @@ RestrictionParser::TryParse(const osmium::Relation &relation) const
...
+
+    return std::move(parsed_restrictions);
```
daniel-j-h: The C++ guarantees for function-local variables to be moved out by default. If you add a move you penalize further optimization.

By which I mean

    T fn() {
      T tmp;
      ..
      return tmp; // will be moved out if possible
    }

`return move(tmp)` would be bad for optimization here. Also there are situations in which you have to move:


    struct S {
      T gn() {
        return tmp; // will _not_ move tmp out, as tmp is not a function-local variable
      }

      T tmp;
    };

(will crash if you cann gn() more than once, since a moved-from object is in a valid but unspecified state)

There are some articles out there, search for penalizing move, RVO, e.g.

https://www.ibm.com/developerworks/community/blogs/5894415f-be62-4bc0-81c5-3956e82276f3/entry/RVO_V_S_std_move?lang=en

src/extractor/scripting_environment_lua.cpp
```diff
@@ -525,8 +526,16 @@ void Sol2ScriptingEnvironment::ProcessElements(
...
-                    resulting_restrictions.push_back(restriction_parser.TryParse(
-                        static_cast<const osmium::Relation &>(*entity)));
+                    result_res.clear();
+                    result_res = restriction_parser.TryParse(
+                        static_cast<const osmium::Relation &>(*entity));
+                    if (result_res.size() > 0)
+                    {
+                        for (boost::optional<InputRestrictionContainer> r : result_res)
+                        {
+                            resulting_restrictions.push_back(r);
```
daniel-j-h: for (T t : res)

will make a copy for every r in res; which is fine if you want to make a `move(r)` here

src/updater/updater.cpp
```diff
@@ -44,6 +46,25 @@
...
+    }
+};
+
+template <typename T1, typename T2> struct hash<std::tuple<T1, T2>>
+{
+    size_t operator()(const std::tuple<T1, T2> &t) const
+    {
+        return hash_val(std::get<0>(t), std::get<1>(t));
+    }
+};
```
daniel-j-h: Where does hash_val come from? Also can you just use `boost::hash` they properly do the hash mixing part

http://www.boost.org/doc/libs/1_63_0/doc/html/hash.html

src/updater/updater.cpp
```diff
@@ -408,27 +474,98 @@ updateTurnPenalties(const UpdaterConfig &config,
...
+    BOOST_ASSERT(is_aligned<extractor::lookup::TurnIndexBlock>(turn_index_blocks));
+
+    std::vector<std::uint64_t> updated_turns;
+    if (conditional_turns.size() == 0)
+        return updated_turns;
+
+    // TODO make this into a function
+    LookupTable<std::tuple<NodeID, NodeID>, NodeID> is_only_lookup;
+    std::unordered_set<std::tuple<NodeID, NodeID, NodeID>,
+                       std::hash<std::tuple<NodeID, NodeID, NodeID>>>
```
daniel-j-h: Here you'd need to change to `boost::hash` and it should magically just work

include/extractor/serialization.hpp
```diff
@@ -81,6 +85,54 @@ inline void write(storage::io::FileWriter &writer,
...
+        for (uint64_t i = 0; i < num_conditions; i++)
+        {
+            reader.ReadInto(restriction.condition[i].modifier);
+            storage::serialization::read(reader, restriction.condition[i].times);
+            storage::serialization::read(reader, restriction.condition[i].weekdays);
+            storage::serialization::read(reader, restriction.condition[i].monthdays);
+        }
+        restriction.flags.is_only = is_only;
+
+        restrictions.push_back(restriction);
```
karenzshea: Arghhh sorry comments like these are very helpful :pray: You already explained this to me once though, I was using an `std::move` before and then I must have rebased it away. Thanks for pointing out again!

src/extractor/extractor.cpp
```diff
@@ -317,6 +191,140 @@ int Extractor::run(ScriptingEnvironment &scripting_environment)
...
+    if (timestamp.empty())
+    {
+        timestamp = "n/a";
+    }
+    util::Log() << "timestamp: " << timestamp;
+
+    storage::io::FileWriter timestamp_file(config.timestamp_file_name,
+                                           storage::io::FileWriter::HasNoFingerprint);
+
+    timestamp_file.WriteFrom(timestamp.c_str(), timestamp.length());
```
karenzshea: Hmm, true it seems like this is not being used. This should be ticketed and done in a separate PR I think.

src/extractor/restriction_parser.cpp
```diff
@@ -160,7 +173,42 @@ RestrictionParser::TryParse(const osmium::Relation &relation) const
...
+                                            << fi_begin->value() << "\" at the turn "
+                                            << restriction_container.restriction.from.way << " -> "
+                                            << restriction_container.restriction.via.node << " -> "
+                                            << restriction_container.restriction.to.way;
+                continue;
+            }
+
+            for (auto &p : parsed)
+            {
+                std::vector<util::OpeningHours> parsed = util::ParseOpeningHours(p.condition);
```
karenzshea: :0 The utility functions I'm using here first parse the conditional restrictions on a tag into a vector of string-based types for the condition, then this `ParseOpeningHours` function parses the strings into a vector of enum-based types for better storage. 

I'll add it to the cleanup task lists to merge these steps into one.

src/extractor/restriction_parser.cpp
karenzshea: I can do this, but requires removing `TryParse`'s constness as a function

src/extractor/restriction_parser.cpp
oxidase: Conditional restrictions are not always based on the opening hours grammar. it is better to keep both steps separated, otherwise it will require a back refactoring if eg vehicle types must be supported like "weight > 5t"

karenzshea: Oo good point @oxidase forgot about the possibility of supporting conditional speed limits as well.

src/extractor/extractor.cpp
karenzshea: https://github.com/Project-OSRM/osrm-backend/issues/3961

src/updater/updater.cpp
```diff
@@ -409,27 +435,141 @@ updateTurnPenalties(const UpdaterConfig &config,
...
+    auto max_edge_id = Updater::LoadAndUpdateEdgeExpandedGraph(
+        edge_based_edge_list, node_weights, node_coordinates, osm_node_ids);
     return std::make_tuple(max_edge_id + 1, std::move(edge_based_edge_list));
 }
 
 EdgeID
 Updater::LoadAndUpdateEdgeExpandedGraph(std::vector<extractor::EdgeBasedEdge> &edge_based_edge_list,
-                                        std::vector<EdgeWeight> &node_weights) const
+                                        std::vector<EdgeWeight> &node_weights,
+                                        std::vector<util::Coordinate> node_coordinates,
```
TheMarex: `const std::vector<util::Coordinate>&`, `const extractor::PackedOSMIDs&` otherwise these vectors will be copied.

karenzshea:
I found a case that still needs to be handled:
A restriction is tagged on a relation where one of the restriction's members references an invalid internal id. During extraction, restrictions whose via node is a node on a from/to way that is not the start or target node are deemed invalid restrictions, but they are not dropped from the restrictions list.

This technically is an issue in master as well, but is an edge case and difficult to hit. For now I've added a test case that triggers it and marks it as @todo

src/updater/updater.cpp
karenzshea: This doesn't work, boost::hash implementation for tuples doesn't work across all the systems we test on https://github.com/Project-OSRM/osrm-backend/issues/3962#issuecomment-297096071

features/car/conditional_restrictions.feature
```diff
@@ -0,0 +1,447 @@
...
+    @no_turning @conditionals
+    Scenario: Car - Restriction would be on, but the restriction was badly tagged
+        Given the extract extra arguments "--parse-conditional-restrictions=1"
+                                            # time stamp for 10am on Tues, 02 May 2017 GMT
+        Given the contract extra arguments "--time-zone-file=test/data/tz_world.shp --parse-conditionals-from-now=1493719200"
+        Given the customize extra arguments "--time-zone-file=test/data/tz_world.shp --parse-conditionals-from-now=1493719200"
+
+        Given the node map
+            """
+              n
```
TheMarex: Some dots and lines here could be nice to signify which nodes form a street.

.travis.yml
```diff
@@ -325,7 +325,7 @@ install:
              -DENABLE_CCACHE=ON \
              -DCMAKE_INSTALL_PREFIX=${OSRM_INSTALL_DIR}
   - echo "travis_fold:start:MAKE"
-  - make --jobs=${JOBS}
+  - make --jobs=${JOBS} VERBOSE=1
```
daniel-j-h: don't forget to remove this before merging

features/car/conditional_restrictions.feature
```diff
@@ -0,0 +1,447 @@
...
+            | ej    | no     |
+            | jp    | no     |
+
+        And the relations
+            | type        | way:from | way:to | node:via | restriction:conditional               | except   |
+            | restriction | ej       | nj     | j        | only_right_turn @ (Mo-Su 08:00-12:00) | motorcar |
+            | restriction | jp       | nj     | j        | only_left_turn @ (Mo-Su 08:00-12:00)  | bus      |
+
+        When I route I should get
+            | from | to | route          |
```
TheMarex: By adding a column with a `#` header, we could describe what each row is supposed to test.

CMakeLists.txt
```diff
@@ -448,8 +448,9 @@ if(ENABLE_MASON)
   set(TBB_LIBRARIES ${MASON_PACKAGE_tbb_LDFLAGS})
 
   mason_use(libshp2 VERSION ${MASON_LIBSHP_VERSION})
-  set(LIBSHAPEFILE_INCLUDE_DIR ${MASON_PACKAGE_libshp2_INCLUDE_DIRS})
-  set(LIBSHAPEFILE_LIBRARY ${MASON_PACKAGE_libshp2_LDFLAGS})
+  target_include_directories(UTIL PRIVATE ${MASON_PACKAGE_libshp2_INCLUDE_DIRS})
+  set(MAYBE_SHAPEFILE ${MASON_PACKAGE_libshp2_LDFLAGS})
```
daniel-j-h: I don't understand this - I thought `MAYBE_SHAPEFILE` contains the (optional) shapefile lib, e.g. as in `libshapefile.so`? Why do we add linker flags here, too?

features/step_definitions/data.js
```diff
@@ -38,7 +38,7 @@ module.exports = function () {
         callback();
     });
 
-    this.Given(/^the origin ([-+]?[0-9]*\.?[0-9]+),([-+]?[0-9]*\.?[0-9]+)$/, (lat, lon, callback) => {
+    this.Given(/^the origin ([-+]?[0-9]*\.?[0-9]+),([-+]?[0-9]*\.?[0-9]+)$/, (lon, lat, callback) => {
```
TheMarex: Classic.

CMakeLists.txt
```diff
@@ -524,6 +525,15 @@ else()
     ENDIF()
   ENDIF()
 
+  find_package(Shapefile) # optional package libshp-dev
+  if (Shapefile_FOUND)
+    set(MAYBE_SHAPEFILE "${LIBSHAPEFILE_LIBRARY}")
+    target_include_directories(UTIL PRIVATE ${LIBSHAPEFILE_INCLUDE_DIR})
+    target_compile_definitions(UTIL PRIVATE COMPILE_DEFINITIONS ENABLE_SHAPEFILE)
+  else()
```
daniel-j-h: Compare to above: here we only add the library to `MAYBE_SHAPEFILE`.

include/extractor/serialization.hpp
```diff
@@ -94,6 +98,54 @@ inline void write(storage::io::FileWriter &writer,
     storage::serialization::write(writer, node_data_container.name_ids);
     storage::serialization::write(writer, node_data_container.travel_modes);
 }
+
+// read/write for conditional turn restrictions file
+inline void read(const boost::filesystem::path &path, std::vector<TurnRestriction> &restrictions)
```
TheMarex: Instead of taking a path this should take an instance of `FileReader` like the other functions in this file.

You should also add a function `readRestrictions` in `files.hpp` that takes a path and then calls this function.

We do this to abstract away the notion of files from parsing the data.

include/storage/io.hpp
```diff
@@ -108,6 +108,13 @@ class FileReader
         }
     }
 
+    void ReadInto(std::string &target)
```
TheMarex: We want to keep these classes as small as possible, hence I would suggest adding a function to `storage/io.hpp` like `read(FileReader, std::string)` and `write(FileWrite, const std::string &)`.

include/storage/io.hpp
```diff
@@ -225,6 +232,13 @@ class FileWriter
         WriteFrom(src.data(), src.size());
     }
 
+    void WriteFrom(const std::string &src)
```
TheMarex: Same as above, should be moved to `storage/io.hpp`

include/updater/updater.hpp
```diff
@@ -2,9 +2,20 @@
...
 
 #include "extractor/edge_based_edge.hpp"
+#include "extractor/packed_osm_ids.hpp"
+#include "extractor/query_node.hpp"
+#include "extractor/restriction.hpp"
+#include "util/coordinate.hpp"
+#include "util/packed_vector.hpp"
 
+#include <boost/geometry.hpp>
+#include <boost/geometry/index/rtree.hpp>
```
TheMarex: There is no direct dependency on the RTree in the header, it should be fine to remove these and only add includes to `updater.cpp`. All added includes should be removed and moved to `updater.cpp` if it causes compile errors.

This reduces dependencies of files only including `updater.hpp`.

include/util/extract-conditionals.hpp
```diff
@@ -0,0 +1,386 @@
+#include "util/conditional_restrictions.hpp"
```
TheMarex: This file does not have an include guard and is not included besides from `src/tools/extract-conditionals.cpp`. It needs include guard and should be moved to `include/tools`.

include/util/timezones.hpp
```diff
@@ -0,0 +1,44 @@
...
+// rtree_t is a lookup R-tree that maps a geographic point to an index in a local_time_t vector
+using point_t = boost::geometry::model::
+    point<double, 2, boost::geometry::cs::spherical_equatorial<boost::geometry::degree>>;
+using polygon_t = boost::geometry::model::polygon<point_t>;
+using box_t = boost::geometry::model::box<point_t>;
+using rtree_t =
+    boost::geometry::index::rtree<std::pair<box_t, size_t>, boost::geometry::index::rstar<8>>;
+using local_time_t = std::pair<polygon_t, struct tm>;
+
+std::function<struct tm(const point_t &)> LoadLocalTimesRTree(const std::string &tz_shapes_filename,
```
TheMarex: This `pre-decleration` + all the `using` above (except the `point_t`) should be moved to `timezoner.cpp`

src/contractor/contractor.cpp
```diff
@@ -7,6 +7,7 @@
 #include "extractor/compressed_edge_container.hpp"
 #include "extractor/edge_based_graph_factory.hpp"
 #include "extractor/node_based_edge.hpp"
+#include "extractor/packed_osm_ids.hpp"
```
TheMarex: Unnecessary include.

src/extractor/extractor.cpp
```diff
@@ -115,142 +115,15 @@ transformTurnLaneMapIntoArrays(const guidance::LaneDescriptionMap &turn_lane_map
...
-            {
-                extractor_callbacks->ProcessRestriction(result);
-            }
-        }
-        TIMER_STOP(parsing);
-        util::Log() << "Parsing finished after " << TIMER_SEC(parsing) << " seconds";
-
-        util::Log() << "Raw input contains " << number_of_nodes << " nodes, " << number_of_ways
-                    << " ways, and " << number_of_relations << " relations";
+        auto turn_restrictions = ParseOSMData(scripting_environment, number_of_threads);
```
TheMarex: Nice refactor. You could now also remove the scoping `{ }` since we only have the function scope now.

src/extractor/restriction_parser.cpp
```diff
@@ -54,9 +56,10 @@ RestrictionParser::RestrictionParser(ScriptingEnvironment &scripting_environment
  * in the corresponding profile. We use it for both namespacing restrictions, as in
  * restriction:motorcar as well as whitelisting if its in except:motorcar.
  */
-boost::optional<InputRestrictionContainer>
+std::vector<boost::optional<InputRestrictionContainer>>
```
TheMarex: We can get rid of the `boost::optional` here. The `optional` was only used because a function needs to return something even if the parsing fails. In this case we can just return an empty vector. 🎉

src/extractor/scripting_environment_lua.cpp
```diff
@@ -498,6 +498,7 @@ void Sol2ScriptingEnvironment::ProcessElements(
         [&](const tbb::blocked_range<std::size_t> &range) {
             ExtractionNode result_node;
             ExtractionWay result_way;
+            std::vector<boost::optional<InputRestrictionContainer>> result_res;
```
TheMarex: Remove the `optional`.

src/extractor/scripting_environment_lua.cpp
```diff
@@ -525,8 +526,16 @@ void Sol2ScriptingEnvironment::ProcessElements(
                     resulting_ways.push_back(std::make_pair(x, std::move(result_way)));
                     break;
                 case osmium::item_type::relation:
-                    resulting_restrictions.push_back(restriction_parser.TryParse(
-                        static_cast<const osmium::Relation &>(*entity)));
+                    result_res.clear();
+                    result_res =
+                        restriction_parser.TryParse(static_cast<const osmium::Relation &>(*entity));
+                    if (result_res.size() > 0)
```
TheMarex: Size check is not necessary, the for loop would just not do any iterations in that case.

src/updater/updater.cpp
```diff
@@ -135,12 +156,10 @@ tbb::concurrent_vector<GeometryID>
 updateSegmentData(const UpdaterConfig &config,
                   const extractor::ProfileProperties &profile_properties,
                   const SegmentLookupTable &segment_speed_lookup,
-                  extractor::SegmentDataContainer &segment_data)
+                  extractor::SegmentDataContainer &segment_data,
+                  std::vector<util::Coordinate> coordinates,
```
TheMarex: This copies all the coordinates and OSM ids. Should be `const std::vector<util::Coordinate>&` and `const extractor::PackedOSMIDs&`.

src/updater/updater.cpp
```diff
@@ -413,20 +439,128 @@ updateTurnPenalties(const UpdaterConfig &config,
 
         if (turn_weight_penalty < 0)
         {
-            util::Log(logWARNING) << "Negative turn penalty at " << turn_index.from_id << ", "
-                                  << turn_index.via_id << ", " << turn_index.to_id
+            util::Log(logWARNING) << "Negative turn penalty at " << internal_turn.from_id << ", "
```
TheMarex: You could use `osm_turn` instead of `internal_turn` here because it is easier for the user to see OSM ids instead of internal IDs.

src/updater/updater.cpp
```diff
@@ -413,20 +439,128 @@ updateTurnPenalties(const UpdaterConfig &config,
...
+bool IsRestrictionValid(const Timezoner &tz_handler,
+                        const extractor::TurnRestriction &turn,
+                        const std::vector<util::Coordinate> &coordinates,
+                        const extractor::PackedOSMIDs &osm_node_ids)
+{
+    // get restriction's lon/lat coords
+    const auto via_node = osm_node_ids[turn.via.node];
+    const auto from_node = osm_node_ids[turn.from.node];
+    const auto to_node = osm_node_ids[turn.to.node];
+    const auto &lon = static_cast<double>(toFloating(coordinates[turn.via.node].lon));
```
TheMarex: no `&` for `lon` and `lat`. For `double` (8 bytes) it is faster to copy, since a memory address is also 8 bytes.

src/updater/updater.cpp
```diff
@@ -413,20 +439,128 @@ updateTurnPenalties(const UpdaterConfig &config,
...
+    // Get local time of the restriction
+    const auto &local_time = tz_handler.GetLocalTime(point_t{lon, lat});
+
+    // TODO: check restriction type [:<transportation mode>][:<direction>]
+    // http://wiki.openstreetmap.org/wiki/Conditional_restrictions#Tagging
+
+    // TODO: parsing will fail for combined conditions, e.g. Sa-Su AND weight>7
+    // http://wiki.openstreetmap.org/wiki/Conditional_restrictions#Combined_conditions:_AND
+
+    if (condition.empty())
```
TheMarex: Move this at the beginning of the function for an early exit. In that case we don't actually need to do any of the operations above.

src/updater/updater.cpp
```diff
@@ -413,20 +439,128 @@ updateTurnPenalties(const UpdaterConfig &config,
...
+    {
+        osrm::util::Log(logWARNING) << "Condition parsing failed for the turn " << from_node
+                                    << " -> " << via_node << " -> " << to_node;
+        return false;
+    }
+
+    if (osrm::util::CheckOpeningHours(condition, local_time))
+    {
+        return true;
+        // output_stream << restriction.from << "," << restriction.via << "," << restriction.to
```
TheMarex: Left-over debugging code.

src/updater/updater.cpp
```diff
@@ -413,20 +439,128 @@ updateTurnPenalties(const UpdaterConfig &config,
...
+        //              << "," << restriction_value << "\n";
+    }
+    return false;
+}
+
+std::vector<std::uint64_t>
+updateConditionalTurns(const UpdaterConfig &config,
+                       std::vector<TurnPenalty> &turn_weight_penalties,
+                       const std::vector<extractor::TurnRestriction> &conditional_turns,
+                       std::vector<util::Coordinate> coordinates,
```
TheMarex: This has an accidental copy again. :-)

src/updater/updater.cpp
```diff
@@ -413,20 +439,128 @@ updateTurnPenalties(const UpdaterConfig &config,
...
+    {
+        const extractor::lookup::TurnIndexBlock internal_turn = turn_index_blocks[edge_index];
+
+        const auto is_no_tuple =
+            std::make_tuple(internal_turn.from_id, internal_turn.via_id, internal_turn.to_id);
+        const auto is_only_tuple = std::make_tuple(internal_turn.from_id, internal_turn.via_id);
+        // turn has a no_* restriction
+        if (is_no_set.find(is_no_tuple) != is_no_set.end())
+        {
+            util::Log() << "Conditional penalty set on edge: " << edge_index;
```
TheMarex: This will be a very noisy log statement. I would remove this for production.

src/updater/updater.cpp
```diff
@@ -413,20 +439,128 @@ updateTurnPenalties(const UpdaterConfig &config,
...
+            updated_turns.push_back(edge_index);
+        }
+        // turn has an only_* restriction
+        else if (is_only_lookup(is_only_tuple))
+        {
+            // with only_* restrictions, the turn on which the restriction is tagged is valid
+            if (*is_only_lookup(is_only_tuple) == internal_turn.to_id)
+                continue;
+
+            util::Log() << "Conditional penalty set on edge: " << edge_index;
```
TheMarex: Same here, noisy log statement.

src/updater/updater.cpp
```diff
@@ -543,6 +694,40 @@ Updater::LoadAndUpdateEdgeExpandedGraph(std::vector<extractor::EdgeBasedEdge> &e
                        });
     }
 
+    if (SupportsShapefiles())
```
TheMarex: Merge the if-statements with an `&&` operator, to make this more readable.

src/updater/updater.cpp
```diff
@@ -543,6 +694,40 @@ Updater::LoadAndUpdateEdgeExpandedGraph(std::vector<extractor::EdgeBasedEdge> &e
...
+    {
+        if (update_conditional_turns)
+        {
+            Timezoner time_zone_handler;
```
TheMarex: You can make this `const` if you use the lambda trick:
```
const Timezoner zoner = [&]() {
 if (config.value_now) { return Timezoner(); }
 else { return Timezoner(); }
}();
```

CMakeLists.txt
karenzshea: This should be pointing to `${MASON_PACKAGE_libshp2_STATIC_LIBS}`

src/updater/updater.cpp
karenzshea: o nice

src/updater/updater.cpp
karenzshea: I could make these debug statements?

include/util/extract-conditionals.hpp
karenzshea: Somehow I ended up with an old copy of the same file in `include/conditionals/` that did have an include guard. I've deleted that one and added the guard to this file.

src/updater/updater.cpp
TheMarex: Or that, if you prefer.

src/updater/updater.cpp
kkaefer: you can also leave out the `()` of the lambda

joto: Or, if it is something simple, like in this case, use `?:` operator:

    const auto time_zone_handler = config.valid_now <= 0
        ? Timezoner(config.tz_file_path)
        : Timezoner(config.tz_file_path, config.valid_now);

which is not quite as ugly as a lambda.

src/updater/updater.cpp
karenzshea: Comes from our own implementation of `std_hash`. Boost::hash implementation for tuples doesn't work across all the systems we test on https://github.com/Project-OSRM/osrm-backend/issues/3962#issuecomment-297096071

src/util/timezones.cpp
```diff
@@ -0,0 +1,152 @@
...
+        rtree.query(boost::geometry::index::intersects(point), std::back_inserter(result));
+        for (const auto v : result)
+        {
+            const auto index = v.second;
+            if (boost::geometry::within(point, local_times[index].first))
+                return local_times[index].second;
+        }
+        return tm{0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0};
+    };
+#endif
```
TheMarex: For Visual-Studio this needs to return an empty value to compile:
```
#else
return std::function<struct tm(const point_t &)>{};
#endif
```

TheMarex:
It seems like we are running out of memory now on AppVeyor. Not sure if we can actually fix this, digging the build config.

oxidase:
@karenzshea i have pushed a fix for msvs. please don't split tests into conditionals and non-conditionals: if there is no TZ information [`tm`](https://github.com/Project-OSRM/osrm-backend/blob/c7a84e6eef3c298c156ef015b0e3e7effa271e56/src/util/timezones.cpp#L139) must fall back to the input UTC. This can be used to setup feature tests with conditionals without TZ shape files in both linux and windows builds

CMakeLists.txt
```diff
@@ -337,6 +337,7 @@ elseif(${CMAKE_CXX_COMPILER_ID} STREQUAL "MSVC")
   # using Visual Studio C++
   set(BOOST_COMPONENTS ${BOOST_COMPONENTS} zlib)
   set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} /bigobj") # avoid compiler error C1128 from scripting_environment_lua.cpp
+  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} /Zm200")    # avoid compiler error C1060 due to Boost.Spirit on AppVeyor
```
oxidase: O, this is leftover from msvc checks,  please remove the line

TheMarex:
@oxidase @karenzshea I really disagree about removing testing with a shapefile here. There should be no silent fallback to UTC, a shapefile should be a hard requirement for conditional turns.

If shipping the original shapefile is a problem we need to ship our own. Splitting the testing is ugly but can't be avoided with optional dependencies.

oxidase:
@TheMarex i don't see any problem in including dummy TZ shapefile, but keeping 50Mb binary data is not good anyway because just shapefile loading in debug takes more than 5 seconds. 
I added a dummy tz file with two "timzone" polygins generated from https://gist.github.com/oxidase/3122c705c9533f73fc63b01b22ad3ea8

mloskot:
Does this introduce requirement of GeoJSON timezone file during contraction step of the processing flow?


Does this look like a valid way to disable the updater step during contraction?
```
osrm::ContractorConfig config;
config.updater_config.valid_now = 0; // disable updating of conditional turn restriction (requires timezone file)
...
osrm::contract(config);
```

karenzshea:
@mloskot This introduced an optional requirement of a GeoJSON timezone file.

Whether or not conditional turn restriction updating is applied during contraction/customization depends onthe [presence of the `.osrm.restrictions` file](https://github.com/Project-OSRM/osrm-backend/blob/master/src/updater/updater.cpp#L546-L547), which contains the extracted conditional restrictions.

You can disable conditional turn restriction application by disabling the `--parse-conditional-restrictions` flag during `osrm-extract`, or removing the `.osrm.restrictions` file.

PR #3840, opened 2017-03-20
Add read-only max_turn_weight in ProfileProperties

oxidase:
# Issue

`max_turn_weight` can be useful in profiles to avoid some turns by setting maximum possible turn weight.

## Tasklist
 - [x] add regression / cucumber cases (see docs/testing.md)
 - [x] review
 - [x] adjust for comments

## Requirements / Relations
 Link any requirements here. Other pull requests this PR is based on?

PR #3842, opened 2017-03-20
Document all options available to Lua profiles

systemed:
Recent improvements to OSRM have led to a lot of changes to the Lua scripting interface, and refactoring of the default profiles makes them slightly more abstract and difficult to grok for those writing profiles from scratch.

This is therefore an attempt to fully document the attributes that can be read and set from Lua profiles. I've done it from my own understanding and from reading the source, but I may well have missed things or mistaken them!

### DIFF
--- a/docs/profiles.md
+++ b/docs/profiles.md
@@ -25,6 +25,17 @@ As you scroll down the file you'll see local variables, and then local functions
 
 `way_function` and `node_function` are the important functions which are called when extracting OpenStreetMap data with `osrm-extract`.
 
+The following global properties can be set in your profile:
+
+Attribute                     | Type     | Notes
+------------------------------|----------|----------------------------------------------------------------------------
+weight_name                   | String   | Name used in output for the routing weight property (default 'duration')
+weight_precision              | Unsigned | Decimal precision of edge weights (default 1)
+left_hand_driving             | Boolean  | Are vehicles assumed to drive on the left? (used in guidance)
+use_turn_restrictions         | Boolean  | Are turn instructions followed?
+continue_straight_at_waypoint | Boolean  | Must the route continue straight on at a via point, or are U-turns allowed?
+max_speed_for_map_matching    | Float    | Maximum vehicle speed to be assumed in matching (in m/s)
+
 ## way_function
 
 Given an OpenStreetMap way, the way_function will either return nothing (meaning we are not going to route over this way at all), or it will set up a result hash to be returned. The most important thing it will do is set the value of `result.forward_speed` and `result.backward_speed` as a suitable integer value representing the speed for traversing the way.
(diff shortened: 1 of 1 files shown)


docs/profiles.md
```diff
@@ -33,15 +44,75 @@ All other calculations stem from that, including the returned timings in driving
...
+backward_speed                          | Float    |  "   "
+forward_mode                            | Enum     | Mode of travel (e.g. car, ferry). Defined in include/extractor/travel_mode.hpp
+backward_mode                           | Enum     |  "   "
+duration                                | Float    | Alternative setter for duration of the whole way in both directions
+weight                                  | Float    | Alternative setter for weight of the whole way in both directions
+turn_lanes_forward                      | String   | Directions for individual lanes (normalised OSM turn:lanes value)
+turn_lanes_backward                     | String   |  "   "
+forward_restricted                      | Boolean  | Is this a restricted access road? (e.g. private, deliveries only)
+backward_restricted                     | Boolean  |  "   "
+is_startpoint                           | Boolean  | Can a journey start on this way?
```
daniel-j-h: Maybe add a note that it is used for snapping / to filter out what is in the rtree

docs/profiles.md
```diff
@@ -33,15 +44,75 @@ All other calculations stem from that, including the returned timings in driving
...
+backward_rate                           | Float    |  "   "
+forward_speed                           | Float    | Speed on this way in km/h
+backward_speed                          | Float    |  "   "
+forward_mode                            | Enum     | Mode of travel (e.g. car, ferry). Defined in include/extractor/travel_mode.hpp
+backward_mode                           | Enum     |  "   "
+duration                                | Float    | Alternative setter for duration of the whole way in both directions
+weight                                  | Float    | Alternative setter for weight of the whole way in both directions
+turn_lanes_forward                      | String   | Directions for individual lanes (normalised OSM turn:lanes value)
+turn_lanes_backward                     | String   |  "   "
+forward_restricted                      | Boolean  | Is this a restricted access road? (e.g. private, deliveries only)
```
daniel-j-h: Maybe add a note re. what restricted actually means https://www.openstreetmap.org/user/happygo/diary/40564

docs/profiles.md
```diff
@@ -25,6 +25,17 @@ As you scroll down the file you'll see local variables, and then local functions
...
+
+Attribute                     | Type     | Notes
+------------------------------|----------|----------------------------------------------------------------------------
+weight_name                   | String   | Name used in output for the routing weight property (default 'duration')
+weight_precision              | Unsigned | Decimal precision of edge weights (default 1)
+left_hand_driving             | Boolean  | Are vehicles assumed to drive on the left? (used in guidance)
+use_turn_restrictions         | Boolean  | Are turn instructions followed?
+continue_straight_at_waypoint | Boolean  | Must the route continue straight on at a via point, or are U-turns allowed?
+max_speed_for_map_matching    | Float    | Maximum vehicle speed to be assumed in matching (in m/s)
+
```
oxidase: Please could you add here a new property introduced in #3840 ?
`max_turn_weight | Float |  Maximum turn penalty weight |`

docs/profiles.md
```diff
@@ -33,15 +45,75 @@ All other calculations stem from that, including the returned timings in driving
 
 Using the power of the scripting language you wouldn't typically see something as simple as a `result.forward_speed = 20` line within the way_function. Instead a way_function will examine the tagging (e.g. `way:get_value_by_key("highway")` and many others), process this information in various ways, calling other local functions, referencing the global variables and look-up hashes, before arriving at the result.
 
-## Guidance
+The following attributes can be set on the result in way_function:
+
+Attribute                               | Type     | Notes
+----------------------------------------|----------|--------------------------------------------------------------------------
+forward_rate                            | Float    | Routing weight on this way per metre
```
TheMarex: The unit for `rate` is actually weirder. Speed is `meter per seconds`, this is a fake speed quantity so the unit is `meter per {weight}`.

docs/profiles.md
```diff
@@ -25,6 +25,18 @@ As you scroll down the file you'll see local variables, and then local functions
...
+weight_precision              | Unsigned | Decimal precision of edge weights (default 1)
+left_hand_driving             | Boolean  | Are vehicles assumed to drive on the left? (used in guidance)
+use_turn_restrictions         | Boolean  | Are turn instructions followed?
+continue_straight_at_waypoint | Boolean  | Must the route continue straight on at a via point, or are U-turns allowed?
+max_speed_for_map_matching    | Float    | Maximum vehicle speed to be assumed in matching (in m/s)
+max_turn_weight               | Float    | Maximum turn penalty weight
+
 ## way_function
 
 Given an OpenStreetMap way, the way_function will either return nothing (meaning we are not going to route over this way at all), or it will set up a result hash to be returned. The most important thing it will do is set the value of `result.forward_speed` and `result.backward_speed` as a suitable integer value representing the speed for traversing the way.
```
emiltin: it's just as important to set forward/backward mode

docs/profiles.md
systemed: Ok, got it. How would you define `{weight}`?

docs/profiles.md
systemed: Yes, agreed. I've not looked at the narrative text (which I think was originally written by @harry-wood) in this PR; this is just to list out the parameters.

harry-wood: Yeah. I remember at the time I noted the need for more detail on what the global variables mean, so this looks like a very useful improvement to the docs

Incidentally from a narrative point of view, it might read more flowingly if you insert this table two paragraphs earlier no? Well doesn't really matter.

Issue #3838, opened 2017-03-20
Quick-fix I/O error handling for OSRM 5.6


danpat:
OSRM 5.7 includes a [nice refactor](https://github.com/Project-OSRM/osrm-backend/pull/3496) of the I/O writing code, with much better error handling.

However, OSRM 5.6 doesn't have this yet, and well, sometimes the demoserver behaves badly if it runs out of disk space and `osrm-extract` doesn't notice.

We should add a quick-fix to the 5.6 branch where we simply do `if (!stream.good()) throw ...` after each file write.  This code would not be backported to `master`, it's just to quickly improve the behaviour of the stable release of OSRM while we wait for 5.7 to land (hopefully in a few weeks).

TheMarex:
There is also `stream.exceptions ( ifstream::failbit | ifstream::badbit )`

daniel-j-h:
Landed with https://github.com/Project-OSRM/osrm-backend/issues/3843.

PR #3843, opened 2017-03-20
Quick-n-dirty error handling for file writes on 5.6 branch

ghoshkaj:
# Issue
https://github.com/Project-OSRM/osrm-backend/issues/3838 related to failures seen in production environments.

## Tasklist
 - [ ] throw errors to all file writes

### DIFF
--- a/include/util/indexed_data.hpp
+++ b/include/util/indexed_data.hpp
@@ -2,6 +2,7 @@
 #define OSRM_INDEXED_DATA_HPP
 
 #include "util/exception.hpp"
+#include "util/exception_utils.hpp"
 #include "util/string_view.hpp"
 
 #include <boost/assert.hpp>
@@ -107,6 +108,11 @@ template <int N, typename T = std::string> struct VariableGroupBlock
 
         out.write((const char *)&refernce, sizeof(refernce));
 
+        if (!out)
+        {
+            throw util::exception(std::string("Error writing block reference: ") + SOURCE_REF);
+        }
+
         return prefix_length;
     }
 
@@ -128,6 +134,11 @@ template <int N, typename T = std::string> struct VariableGroupBlock
                 continue;
 
             out.write((const char *)&data_length, byte_length);
+
+            if (!out)
+            {
+                throw util::exception(std::string("Error writing block prefix: ") + SOURCE_REF);
+            }
         }
     }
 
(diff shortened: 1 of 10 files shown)


include/storage/io.hpp
```diff
@@ -253,7 +253,7 @@ class FileWriter
         const auto &result =
             output_stream.write(reinterpret_cast<const char *>(src), count * sizeof(T));
 
-        if (!result)
+        if (!result || !output_stream.good())
```
daniel-j-h: you can check `if (!stream)` everywhere

include/util/io.hpp
```diff
@@ -36,6 +41,10 @@ bool serializeVector(std::ostream &stream, const std::vector<simple_type> &data)
     stream.write(reinterpret_cast<const char *>(&count), sizeof(count));
     if (!data.empty())
```
danpat: Missed one for the write on L41.

danpat:
@ghoshkaj can you make the error text a bit clearer?  These messages are going to show up in production logs, so clear messages to whoever gets an alarm is always super helpful.  Something like:

  `Writing to .ebg (edge-based-graph) file failed`.

perhaps....

include/util/io.hpp
ghoshkaj: No it's there. It's just on L44. It's referring to the same stream :)

danpat:
@ghoshkaj If you can add a CHANGELOG.md ("Added error handling for file writes") entry, then I think this is basically ready to go for a 5.6.4 patch release.

include/util/indexed_data.hpp
```diff
@@ -107,6 +108,11 @@ template <int N, typename T = std::string> struct VariableGroupBlock
 
         out.write((const char *)&refernce, sizeof(refernce));
 
+        if (!out.good())
+        {
+            throw util::exception(std::string("Error writing block reference: ") + SOURCE_REF);
+        }
```
daniel-j-h: re. `stream.good()` vs `stream.fail()` - if you check the table at the bottom 

http://en.cppreference.com/w/cpp/io/basic_ios/good

you will see that `fail` _is not_ the negation of `good`. It behaves differently for the eof bit. `.good` returns false if the eofbit is set, `.fail` returns false there too (!). I would keep this in mind and simply always check streams via `if (!stream)`.

ghoshkaj: Aha, okay, good an explanation of why one is better. Thanks @daniel-j-h !! 😄

PR #3852, opened 2017-03-22
Give sudo:false a try again

TheMarex:
# Issue

Try to get builds on `sudo:false` running again since they have trusty now.

## Tasklist
 - [ ] review
 - [ ] adjust for comments


TheMarex:
Main problem I encountered here is that container builds start a minute faster but usually take a lot longer. Not actionable right now.

PR #3857, opened 2017-03-26
Added script for performance checks

oxidase:
# Issue

The script is inspired by Frederik Ramm's  presentation ["Routing Engines für OpenStreetMap"](https://www.fossgis-konferenz.de/2017/programm/event.php?id=5195), where statistics is collected for multiple routes and plotted regression `query_time ~ distance`. 

The script runs random queries for CH, CoreCH and MLD algorithms and saves input coordinates,  query times, output distances and weights. Some results for Germany with 10000 queries
compare_ch_mld_corech
and only CH and MLD
compare_ch_mld 

The script also shows that MLD incorrectly returns weights for two paths out of 10000:
* CH weight 22727.5, MLD weight 22820.9
http://map.project-osrm.org/?z=6&loc=52.30485%2C6.544397&loc=52.28878%2C14.46445&hl=en&alt=0

* CH weight 22907.7, MLD weight 23009.3
http://map.project-osrm.org/?z=6&loc=52.9349%2C6.822492&loc=52.38906%2C14.45874&hl=en&alt=0

## Tasklist
 - [x] Check failing paths 
 - [ ] review
 - [ ] adjust for comments

## Requirements / Relations
Can be used in #3816 for performance checks

### DIFF
--- a/scripts/osrm-runner.js
+++ b/scripts/osrm-runner.js
@@ -1,5 +1,6 @@
 #!/usr/bin/env node
 
+const fs = require('fs');
 const http = require('http');
 const process = require('process');
 const cla = require('command-line-args');
@@ -52,11 +53,7 @@ function generate_queries(options, query_points, coordinates_number) {
     {
         let points = query_points.slice(chunk, chunk + coordinates_number);
         let query = options.path.replace(/{}/g, x =>  points.pop().join(','));
-        queries.push({
-            hostname: options.server.hostname,
-            port: options.server.port,
-            path: query
-        });
+        queries.push(query);
     }
     return queries;
 }
@@ -86,8 +83,9 @@ const optionsList = [
     {name: 'max-sockets', alias: 'm', type: Number, defaultValue: 1,
      description: 'how many concurrent sockets the agent can have open per origin, default 1', typeLabel: '[underline]{number}'},
     {name: 'number', alias: 'n', type: Number, defaultValue: 10,
-     description: 'number of query points, default 10', typeLabel: '[underline]{number}'}
-]
+     description: 'number of query points, default 10', typeLabel: '[underline]{number}'},
+    {name: 'queries-files', alias: 'q', type: String,
+     description: 'CSV file with queries in the first row', typeLabel: '[underline]{file}'}];
 const options = cla(optionsList);
 if (options.help) {
     const banner = '╔═╗╔═╗╦═╗╔╦╗      \n║ ║╚═╗╠╦╝║║║      \n╚═╝╚═╝╩╚═╩ ╩      \n┬─┐┬ ┬┌┐┌┌┐┌┌─┐┬─┐\n├┬┘│ │││││││├┤ ├┬┘\n┴└─└─┘┘└┘┘└┘└─┘┴└─';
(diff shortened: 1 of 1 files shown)


daniel-j-h:
:wave: @woodpeck

oxidase:
Changed query time to curl's `time_total` because logged query time does not look reasonable. Unfortunately it is only 1ms precise. Here are plots for 30000 queries
compare_ch_mld_corech
and the lower part
compare_ch_mld

Line fits are:
*  CH   6.2 ms +0.006 ms/km 
*  MLD  7 ms +0.030 ms/km

daniel-j-h:
Talk is [here](https://www.youtube.com/watch?v=rpKU-8J4HAo&index=17&list=PLTli5-lbeoibyuVe_GXqZjYqNT-P83zEp&spfreload=10).

@oxidase good to merge here? These scripts will only gather the numbers, no?
Want to add your scripts for generating plots, too?

woodpeck:
I still need to write up details about the methodology used in my FOSSGIS talk. Essentially I picked 100,000 routes randomly, making sure that both the start and end point were exactly on a road of class "residential" or highere (in order not to disadvantage routers with bad nearest-edge finding too much). Then I ran a Perl script that pushed the 100,000 routes via HTTP to localhost. My talk (non-permanent link) is here: http://www.remote.org/frederik/tmp/routingtalk

oxidase:
@woodpeck thanks for sharing slides and sampling method!

@daniel-j-h i just tried another python script to have more accurate query time data, but think it was really bad idea due to large noise like on the plot
compare_germany
and adding :snake:, because the script mainly duplicates testing on monaco-latest OSM data, where it is a mixture of make, bash and js.

I don't think that results interpretation must be in the script, basically the plot above is just 3 lines in R
```
library(ggplot2)
library(quantreg)
data<-read.csv('germany-latest.csv')
df<-rbind(data.frame(query_time=data$ch_query_time, weight=data$ch_weight, distance=data$ch_distance/1000, algorithm=as.factor('CH')),
          data.frame(query_time=data$mld_query_time, weight=data$mld_weight, distance=data$mld_distance/1000, algorithm=as.factor('MLD')))
ggplot(df, aes(distance, query_time, colour=algorithm)) + coord_cartesian(ylim = c(0, 0.1)) + labs(x="Distance, km", y="Query time, sec") + geom_point(alpha=.2,size=1) + geom_smooth(formula=y~poly(x, 1), method='rq', se = FALSE, alpha=1, size=2)
```

I would like to rewrite the script purely in js to have and use it directly in `test/data` and for black-box performance data  in  #3816

TheMarex:
@oxidase might be worthwhile to write this directly on top of the node bindings to remove any HTTP noise:

```js
const OSRM = require('osrm');
const d3 = require('d3-queue');

function benchmark(algorithm, callback) {
     let osrm = new OSRM({shared_memory: true, algorithm: algorithm});
     let q = d3.queue(1);
     let measure = (query, cb) => {
       // start timer
       osrm.route(query, (err, result) {
          // stop timer
          return cb();
       });
     };

     for (query) { q.defer(measure, query); }
     q.awaitAll(callback);
}
```

oxidase:
Adjusted node.js script to run osrm queries. High query time were due to infinite number of sockets used to run queries and resolved by setting [`maxSockets`](https://github.com/Project-OSRM/osrm-backend/blob/48373e6e1a1198de4f94a63678ea4336b9f17d4a/scripts/osrm-runner.js#L108) to 1 by default.

Results for `maxSockets=1`
germany_mld

Results for `maxSockets=100`
germany_inf

oxidase:
The PR is a precondition for #3816 and was merged in 9fb9d90

PR #3858, opened 2017-03-26
Revert "Integrate internal cell scan into customizer"

oxidase:
# Issue

This reverts commit b085add97362764ee8fb54c8f97d0ecb3ae150d2.

Fixes missing overlay shortcuts if a shortcut contains border edges of higher levels than the current customization level. 

Customization uses both internal and border edges on all levels.

Missing overlay edges lead to non-optimal paths like in #3857
mld
instead of
ch

/cc @TheMarex 

## Tasklist
 - [ ] review
 - [ ] adjust for comments

## Requirements / Relations

### DIFF
--- a/include/customizer/cell_customizer.hpp
+++ b/include/customizer/cell_customizer.hpp
@@ -116,11 +116,11 @@ class CellCustomizer
         }
 
         // Relax base graph edges if a sub-cell border edge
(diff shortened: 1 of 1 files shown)


oxidase:
Should be fixed in #3866

PR #3860, opened 2017-03-27
Documents the user having to be in the docker group for running our images

daniel-j-h:
Some users at fossgis were confused why they could not connect to the docker daemon. Except with sudo which then causes other problems when mixing root and the local user.

Documents the need to be in the docker group - at least on Ubuntu / Debian this seems to be required.

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,18 +41,26 @@ The easiest and quickest way to setup your own routing engine backend is to use 
 
 ### Using Docker
 
-We base the Docker images on Alpine Linux and make sure they are as lightweight as possible (around 10-15 MB).
-In the following, replace `X.Y.Z` with the current stable release version.
+We base [our Docker images](https://hub.docker.com/r/osrm/osrm-backend/) on Alpine Linux and make sure they are as lightweight as possible.
 
 ```
 wget http://download.geofabrik.de/europe/germany/berlin-latest.osm.pbf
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -54,6 +54,15 @@ docker run -t -i -p 5000:5000 -v $(pwd):/data osrm/osrm-backend:vX.Y.Z osrm-rout
 curl http://127.0.0.1:5000/route/v1/driving/13.388860,52.517037;13.385983,52.496891?steps=true
 ```
 
+In case Docker complains about not being able to connect to the Docker daemon make sure you are in the `docker` group.
+
+```
+sudo usermod -aG docker <username>
```
oxidase: `<username>` -> `$USER`

README.md
```diff
@@ -41,7 +41,7 @@ The easiest and quickest way to setup your own routing engine backend is to use
 
 ### Using Docker
 
-We base the Docker images on Alpine Linux and make sure they are as lightweight as possible (around 10-15 MB).
+We base [our Docker images](https://hub.docker.com/r/osrm/osrm-backend/) on Alpine Linux and make sure they are as lightweight as possible (around 10-15 MB).
 In the following, replace `X.Y.Z` with the current stable release version.
```
oxidase: `vX.Y.Z` was also confusing, better to drop here any tags and use always implicitly default tag like `latest` or current stable tag. If someone knows about OSRM versions, it is not a problem to ask for a version tag or look at the OSRM docker page

daniel-j-h:
Good points - changed!

PR #3863, opened 2017-03-27
fixes speed on cycleway+oneway

### LINKED ISSUE #3853: Incorrect travel speed on cycleway travelling against a oneway
When using the bicycle profile, this way:

http://www.openstreetmap.org/way/19937708

is assigned a travel speed of 15km/h in both directions. The forward travel mode is `cycling` but the reverse mode is set to `push bike`.

The reverse travel speed should be set to the `push bike` speed of 6km/h.

Here's an example route that has been chosen based on the fast reverse travel speed on Rue La Roche:

screen shot 2017-03-22 at 1 49 23 pm

But the JSON response indicates that it should be `push bike` along Rue La Roche:

screen shot 2017-03-22 at 1 52 13 pm

So it looks like the profile is setting the mode correctly, but the speed value isn't being set to match :-(

This is causing calculated routes to have large sections of "push bike" because the travel speeds for pushing is not penalized (in the example route above, about 50% of the 2.4km route is pushing).

emiltin:
# Issue

Fixes #3853. It's a step in the right direction, however there are other problems with the existing code, which haven't yet been refactored in the latest round of refactoring. I'm looking into that.

## Tasklist

 - [x] review
 - [x] adjust for comments

### DIFF
--- a/features/bicycle/cycleway.feature
+++ b/features/bicycle/cycleway.feature
@@ -79,3 +79,19 @@ Feature: Bike - Cycle tracks/lanes
             | residential | lane     | yes    | x    | x     |
             | footway     | lane     | yes    | x    | x     |
             | cycleway    | lane     | yes    | x    | x     |
+
+    Scenario: Bike - Cycleway on oneways, modes
+        Then routability should be
+            | highway     | cycleway | oneway | forw         | backw        |
+            | motorway    | track    | yes    | cycling      |              |
+            | residential | track    | yes    | cycling      | pushing bike |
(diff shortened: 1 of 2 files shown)


daniel-j-h:
:+1: Thanks!

> however there are other problems with the existing code

Care to elaborate?

emiltin:
the way that pushing bikes is handled is not very robust. there's a lot of cases it doesn't handle correct. a better way would be to use the foot profile to determine whether you can walk, ie. push a bike.

PR #3866, opened 2017-03-27
Fix internal edges for nodes not in the boundary

TheMarex:
# Issue

@oxidase did the leg work for this in #3858 already. For the edge case of nodes that are not in the boundary and have a very high ID, `BeginBorderEdges` is broken.

## Tasklist
 - [x] review
 - [x] adjust for comments

## Requirements / Relations
Fixes the problem in #3858.

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -1,7 +1,6 @@
 coverage:
 
   ignore:
-    - unit_tests/.*
     - third_party/.*
 
 comment: off
--- a/include/customizer/cell_customizer.hpp
+++ b/include/customizer/cell_customizer.hpp
@@ -39,9 +39,9 @@ class CellCustomizer
                 const EdgeWeight weight = heap.GetKey(node);
 
                 if (level == 1)
-                    RelaxNode<true>(graph, cells, heap, level, id, node, weight);
+                    RelaxNode<true>(graph, cells, heap, level, node, weight);
                 else
-                    RelaxNode<false>(graph, cells, heap, level, id, node, weight);
+                    RelaxNode<false>(graph, cells, heap, level, node, weight);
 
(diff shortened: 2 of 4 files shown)


oxidase:
I also enabled coverage reports for unit tests. It is needed to get correct coverage for code in header files that is covered only in unit tests. 

This PR has the test `check_last_internal_edge` but coverage of the line 115 will not be included in the report because it is ignored by `./unit_tests/*`:
```
./unit_tests/partition-tests -t */check_last_internal_edge
gcov ./unit_tests/CMakeFiles/partition-tests.dir/partition/multi_level_graph.cpp.gcno 
grep -A 18 "EdgeID BeginBorderEdges" multi_level_graph.hpp.gcov 
grep -A 18 "EdgeID BeginBorderEdges" multi_level_graph.hpp.gcov 
       45:  106:    EdgeID BeginBorderEdges(const LevelID level, const NodeID node) const
        -:  107:    {
       45:  108:        auto index = node * GetNumberOfLevels();
        -:  109:        // `node_to_edge_offset` has size `max_border_node_id + 1`
        -:  110:        // which can be smaller then the total number of nodes.
        -:  111:        // this will save memory in case we sort the border nodes first
       45:  112:        if (index >= node_to_edge_offset.size() - 1)
        -:  113:        {
        -:  114:            // On level 0 all edges are border edges
       45:  115:            if (level == 0)
       15:  116:                return SuperT::BeginEdges(node);
        -:  117:            else
       30:  118:                return SuperT::EndEdges(node);
        -:  119:        }
        -:  120:        else
        -:  121:        {
    #####:  122:            return SuperT::BeginEdges(node) + node_to_edge_offset[index + level];
        -:  123:        }
        -:  124:    }
```

PR #3867, opened 2017-03-28
Fix Win tests by copying data explicitly to test/data

duizendnegen:
# Issue

Local Win builds weren't working anymore due to two issues:
- XCOPY can't handle multiple files
- files didn't get copied to `./test/data` and stayed in `./build instead`

## Tasklist
 - [ ] review
 - [ ] adjust for comments


daniel-j-h:
Thanks, looks like AppVeyor is happy again!

Issue #3869, opened 2017-03-28
Raster feature broken: segment_function requires split edges


daniel-j-h:
In https://github.com/Project-OSRM/osrm-backend/issues/3861 @OnlyMate tried to use our experimental raster / elevation feature and hit a problem with the segment function not properly getting called with the forward and backward edge.

The problem is non-split edges. For a-b we only split it into a forward a->b and a backward b->a if the edges are "compatible" (think same speed). Otherwise we only keep a single a-b:

https://github.com/Project-OSRM/osrm-backend/blob/827a595b6c603239468eea29286fe4225b29dd4c/src/extractor/extractor_callbacks.cpp#L325-L330

https://github.com/Project-OSRM/osrm-backend/blob/827a595b6c603239468eea29286fe4225b29dd4c/src/extractor/extractor_callbacks.cpp#L75-L83

This reduces storage requirements substantially but now when we call the segment function later on there is no forward and backward edge - only a single one.

The raster / elevation feature needs all non-split edges to work correctly for forward and backward edge.

Workaround: set `forward_speed = backward_speed * 1.01` to force splitting.

cc @lbud

oxidase:
I think #3625 and  #3775 have the same origin

MichalPP:
I looks like segment_function() calls each segment twice, but with the same source and target (judging by printing source and target lat/lon which are always twice the same, they should be reversed in the second call).

setting different speed/rate did not help in my case.

OnlyMate:
@MichalPP : I wrote something in [https://github.com/Project-OSRM/osrm-backend/issues/3861#issuecomment-289767152](url). Is this the same behavior?

Setting different speed/rate did not helped in my case too. Or I don't know how to set it correctly. :)

MichalPP:
@OnlyMate yes, it's the same behaviour

oxidase:
@MichalPP looks like a bug as no flipping of segment end-points for reversed edges

OnlyMate:
I've got a very ugly solution for me. I need this feature for _slope sensitive routing_. 
I'm using the **bicycle profile** from **OSRM** and added this: 

```lua
function source_function ()
  raster_source = sources:load(
    "/data/swiss.asc",
    5.9559,           -- lon_min
    10.492099999214,  -- lon_max
    45.818,           -- lat_min
    47.808400001144,  -- lat_max
    2388,             -- nrows
    5443              -- ncols
  )
end

local index = 0
[...]

PR #3873, opened 2017-03-29
Reduce memory consumption during extraction and flip backward segment coordinates

### LINKED ISSUE #3861: Speed up/down gradient
I've got a question about elevation data in **OSRM**.
Is there a way to speed up downhill and speed down uphill? 

And another question: 
For example I got a uphill way. Then I have a _positive_ gradient. If I go this way now back, I've got a _negative_ gradient. Is there a way that I can define in the _segment_function_ of the lua-profile the uphill and downhill weight/duration? 

bildschirmfoto 2017-03-27 um 10 30 53

I want that the weight/duration of the two sides on the way are independent. 
I'm working with the master.

My primary goal is, that I got different estimated travel time depends on elevation data.

oxidase:
# Issue

PR solves two issues:
* unnecessarily duplication of backward segments if `in_forward_direction` is true and `split_edge` is false. This will reduce memory consumption during extraction, but will prohibit workaround as in #3869
* fixes #3861 by flipping source and target coordinates for backward edges

For #3869 i would suggest to add and additional flag `split_edge` in `way_function`'s `result` argument and use it at https://github.com/Project-OSRM/osrm-backend/blob/fix/segment-end-points/src/extractor/extractor_callbacks.cpp#L326-L330

The current workaround to split edges to use slightly different weights
```
  result.forward_speed = 15
  result.backward_speed = 15.01
```
## Tasklist
 - [x] review
 - [x] adjust for comments

### DIFF
--- a/src/extractor/extraction_containers.cpp
+++ b/src/extractor/extraction_containers.cpp
@@ -388,21 +388,24 @@ void ExtractionContainers::PrepareEdges(ScriptingEnvironment &scripting_environm
             BOOST_ASSERT(edge_iterator->source_coordinate.lon !=
                          util::FixedLongitude{std::numeric_limits<std::int32_t>::min()});
 
-            const util::Coordinate target_coord{node_iterator->lon, node_iterator->lat};
-            const double distance = util::coordinate_calculation::greatCircleDistance(
-                edge_iterator->source_coordinate, target_coord);
+            util::Coordinate source_coord(edge_iterator->source_coordinate);
+            util::Coordinate target_coord{node_iterator->lon, node_iterator->lat};
 
-            auto weight = edge_iterator->weight_data(distance);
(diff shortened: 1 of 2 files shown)


daniel-j-h:
Wow, that was a quick fix!

We always duplicated the backwards segment or did we introduce this only recently?

I'm :+1: with your suggestion to add an additional flag to the `ExtractionWay` and use it for splitting.

oxidase:
yeah, it can be a post-issue77 bug ;)

Ok, i will add the flag, basically there are  some other bugs with missing backward edges

TheMarex:
Urg this will save like 50% of disk space. Good catch! Can we maybe backport this to 5.6 as well?

MichalPP:
@oxidase works very well, thank you.

TheMarex:
Merging an cherry-picking to 5.6

Issue #3799, opened 2017-03-09
Use Enum instead of `bool` for shared memory templatisation


TheMarex:
Right now multiple classes use the following pattern:

```
namespace detail {
template<bool UseSharedMemory>
class FooContainer{};
}

using FooView = detail::Container<false>;
using FooContainer = detail::Container<true>;
```
Instead of templating for `bool` we should make it a descriptive enum like:

```
enum class MemorySetting : std::uint8_t
{
     SharedMemory,
     InternalMemory
};
```

danpat:
Is this just for comprehensibility?

daniel-j-h:
Yes.

duizendnegen:
I've taken on this task, will push something soon-ish.

duizendnegen:
Can be closed now.

PR #3875, opened 2017-03-29
shared memory enum

duizendnegen:
# Issue

#3799 - use `enum` over `bool` for shared mem.

## Tasklist
 - [x] review
 - [x] adjust for comments


unit_tests/util/static_rtree.cpp~RF1bf682.TMP
```diff
@@ -0,0 +1,481 @@
+#include "util/static_rtree.hpp"
```
oxidase: not needed

include/engine/datafacade/contiguous_internalmem_datafacade.hpp
```diff
@@ -154,7 +155,7 @@ class ContiguousInternalMemoryAlgorithmDataFacade<algorithm::CoreCH>
     : public datafacade::AlgorithmDataFacade<algorithm::CoreCH>
 {
   private:
-    util::ShM<bool, true>::vector m_is_core_node;
+    util::ShM<unsigned, osrm::storage::MemorySetting::SharedMemory>::vector m_is_core_node;
```
oxidase: i think the problem is here because `unsigned` could have different sizes.

Could you clarify why `bool` was changed to `unsigned`?

duizendnegen: Sure, compilation kept failing on my computer here until I changed to `unsigned`. I'll look into re-introducing `bool` and figuring out another way to proceed.

danpat: @duizendnegen Note that the `util::ShM<bool>::vector` class has a special implementation:

https://github.com/Project-OSRM/osrm-backend/blob/master/include/util/shared_memory_vector_wrapper.hpp#L133

where it packs bool values inside individual bits in a large `char` buffer.  This is significantly more compact (probably around 32x) than using `unsigned`.

duizendnegen: That helps @danpat - that probably explains why this part was acting up! I really appreciate getting your input on these issues - OSRM is a fantastic piece of software but sometimes slightly complex :)

duizendnegen:
:ship: it!

include/engine/datafacade/contiguous_internalmem_datafacade.hpp
```diff
@@ -398,28 +421,29 @@ class ContiguousInternalMemoryDataFacadeBase : public BaseDataFacade
     {
         auto offsets_ptr = data_layout.GetBlockPtr<std::uint32_t>(
             memory_block, storage::DataLayout::LANE_DESCRIPTION_OFFSETS);
-        util::ShM<std::uint32_t, true>::vector offsets(
+        util::ShM<std::uint32_t, osrm::storage::MemorySetting::SharedMemory>::vector offsets(
```
TheMarex: Should use `util::SharedMemoryVector` directly.

include/engine/datafacade/contiguous_internalmem_datafacade.hpp
```diff
@@ -398,28 +421,29 @@ class ContiguousInternalMemoryDataFacadeBase : public BaseDataFacade
...
             offsets_ptr, data_layout.num_entries[storage::DataLayout::LANE_DESCRIPTION_OFFSETS]);
         m_lane_description_offsets = std::move(offsets);
 
         auto masks_ptr = data_layout.GetBlockPtr<extractor::guidance::TurnLaneType::Mask>(
             memory_block, storage::DataLayout::LANE_DESCRIPTION_MASKS);
 
-        util::ShM<extractor::guidance::TurnLaneType::Mask, true>::vector masks(
-            masks_ptr, data_layout.num_entries[storage::DataLayout::LANE_DESCRIPTION_MASKS]);
+        util::ShM<extractor::guidance::TurnLaneType::Mask,
```
TheMarex: Same here.

include/engine/datafacade/contiguous_internalmem_datafacade.hpp
```diff
@@ -969,15 +998,17 @@ class ContiguousInternalMemoryAlgorithmDataFacade<algorithm::MLD>
...
+                                                         weight_entries_count);
+            util::SharedMemoryVector<NodeID> source_boundary(mld_source_boundary_ptr,
+                                                             source_boundary_entries_count);
+            util::SharedMemoryVector<NodeID> destination_boundary(
                 mld_destination_boundary_ptr, destination_boundary_entries_count);
-            util::ShM<partition::CellStorageView::CellData, true>::vector cells(
-                mld_cells_ptr, cells_entries_counts);
-            util::ShM<std::uint64_t, true>::vector level_offsets(mld_cell_level_offsets_ptr,
-                                                                 cell_level_offsets_entries_count);
+            util::ShM<partition::CellStorageView::CellData,
```
TheMarex: Should use `SharedMemoryVector`.

include/engine/datafacade/contiguous_internalmem_datafacade.hpp
```diff
@@ -997,13 +1028,13 @@ class ContiguousInternalMemoryAlgorithmDataFacade<algorithm::MLD>
...
-        util::ShM<GraphNode, true>::vector node_list(
+        util::SharedMemoryVector<GraphNode> node_list(
             graph_nodes_ptr, data_layout.num_entries[storage::DataLayout::MLD_GRAPH_NODE_LIST]);
-        util::ShM<GraphEdge, true>::vector edge_list(
+        util::SharedMemoryVector<GraphEdge> edge_list(
             graph_edges_ptr, data_layout.num_entries[storage::DataLayout::MLD_GRAPH_EDGE_LIST]);
-        util::ShM<QueryGraph::EdgeOffset, true>::vector node_to_offset(
-            graph_node_to_offset_ptr,
-            data_layout.num_entries[storage::DataLayout::MLD_GRAPH_NODE_TO_OFFSET]);
+        util::ShM<QueryGraph::EdgeOffset, osrm::storage::MemorySetting::SharedMemory>::vector
```
TheMarex: Same here.

include/engine/datafacade/contiguous_internalmem_datafacade.hpp
```diff
@@ -326,61 +341,69 @@ class ContiguousInternalMemoryDataFacadeBase : public BaseDataFacade
...
+        util::SharedMemoryVector<LaneDataID> lane_data_id(
             lane_data_id_ptr, data_layout.num_entries[storage::DataLayout::LANE_DATA_ID]);
         m_lane_data_id = std::move(lane_data_id);
 
         const auto lane_tupel_id_pair_ptr =
             data_layout.GetBlockPtr<util::guidance::LaneTupleIdPair>(
                 memory_block, storage::DataLayout::TURN_LANE_DATA);
-        util::ShM<util::guidance::LaneTupleIdPair, true>::vector lane_tupel_id_pair(
-            lane_tupel_id_pair_ptr, data_layout.num_entries[storage::DataLayout::TURN_LANE_DATA]);
+        util::ShM<util::guidance::LaneTupleIdPair,
```
TheMarex: Should use `SharedMemoryVector`

include/engine/datafacade/contiguous_internalmem_datafacade.hpp
```diff
@@ -326,61 +341,69 @@ class ContiguousInternalMemoryDataFacadeBase : public BaseDataFacade
 
         const auto travel_mode_list_ptr = data_layout.GetBlockPtr<extractor::TravelMode>(
             memory_block, storage::DataLayout::TRAVEL_MODE);
-        util::ShM<extractor::TravelMode, true>::vector travel_mode_list(
-            travel_mode_list_ptr, data_layout.num_entries[storage::DataLayout::TRAVEL_MODE]);
+        util::ShM<extractor::TravelMode, osrm::storage::MemorySetting::SharedMemory>::vector
```
TheMarex: Should use `SharedMemoryVector`

include/engine/datafacade/contiguous_internalmem_datafacade.hpp
```diff
@@ -211,46 +214,58 @@ class ContiguousInternalMemoryDataFacadeBase : public BaseDataFacade
...
-    util::ShM<BearingClassID, true>::vector m_bearing_class_id_table;
+    util::SharedMemoryVector<BearingClassID> m_bearing_class_id_table;
     // entry class IDs
-    util::ShM<EntryClassID, true>::vector m_entry_class_id_list;
+    util::SharedMemoryVector<EntryClassID> m_entry_class_id_list;
 
     // the look-up table for entry classes. An entry class lists the possibility of entry for all
     // available turns. Such a class id is stored with every edge.
-    util::ShM<util::guidance::EntryClass, true>::vector m_entry_class_table;
+    util::ShM<util::guidance::EntryClass, osrm::storage::MemorySetting::SharedMemory>::vector
```
TheMarex: Should use `SharedMemoryVector`

include/engine/datafacade/contiguous_internalmem_datafacade.hpp
```diff
@@ -211,46 +214,58 @@ class ContiguousInternalMemoryDataFacadeBase : public BaseDataFacade
...
     extractor::SegmentDataView segment_data;
 
-    util::ShM<char, true>::vector m_datasource_name_data;
-    util::ShM<std::size_t, true>::vector m_datasource_name_offsets;
-    util::ShM<std::size_t, true>::vector m_datasource_name_lengths;
-    util::ShM<util::guidance::LaneTupleIdPair, true>::vector m_lane_tupel_id_pairs;
+    util::SharedMemoryVector<char> m_datasource_name_data;
+    util::ShM<std::size_t, osrm::storage::MemorySetting::SharedMemory>::vector
+        m_datasource_name_offsets;
+    util::ShM<std::size_t, osrm::storage::MemorySetting::SharedMemory>::vector
```
TheMarex: Should use `SharedMemoryVector`.

include/engine/datafacade/contiguous_internalmem_datafacade.hpp
```diff
@@ -211,46 +214,58 @@ class ContiguousInternalMemoryDataFacadeBase : public BaseDataFacade
...
+        m_datasource_name_lengths;
+    util::ShM<util::guidance::LaneTupleIdPair, osrm::storage::MemorySetting::SharedMemory>::vector
```
TheMarex: Should use `SharedMemoryVector`.

include/engine/datafacade/contiguous_internalmem_datafacade.hpp
```diff
@@ -211,46 +214,58 @@ class ContiguousInternalMemoryDataFacadeBase : public BaseDataFacade
...
+    util::SharedMemoryVector<TurnPenalty> m_turn_weight_penalties;
+    util::SharedMemoryVector<TurnPenalty> m_turn_duration_penalties;
     extractor::SegmentDataView segment_data;
 
-    util::ShM<char, true>::vector m_datasource_name_data;
-    util::ShM<std::size_t, true>::vector m_datasource_name_offsets;
-    util::ShM<std::size_t, true>::vector m_datasource_name_lengths;
-    util::ShM<util::guidance::LaneTupleIdPair, true>::vector m_lane_tupel_id_pairs;
+    util::SharedMemoryVector<char> m_datasource_name_data;
+    util::ShM<std::size_t, osrm::storage::MemorySetting::SharedMemory>::vector
```
TheMarex: Should use `SharedMemoryVector`.

include/engine/datafacade/contiguous_internalmem_datafacade.hpp
```diff
@@ -211,46 +214,58 @@ class ContiguousInternalMemoryDataFacadeBase : public BaseDataFacade
...
-    util::ShM<bool, true>::vector m_is_core_node;
-    util::ShM<std::uint32_t, true>::vector m_lane_description_offsets;
-    util::ShM<extractor::guidance::TurnLaneType::Mask, true>::vector m_lane_description_masks;
-    util::ShM<TurnPenalty, true>::vector m_turn_weight_penalties;
-    util::ShM<TurnPenalty, true>::vector m_turn_duration_penalties;
+    util::SharedMemoryVector<unsigned> m_name_begin_indices;
+    util::SharedMemoryVector<bool> m_is_core_node;
+    util::ShM<std::uint32_t, osrm::storage::MemorySetting::SharedMemory>::vector
+        m_lane_description_offsets;
+    util::ShM<extractor::guidance::TurnLaneType::Mask,
```
TheMarex: Should use `SharedMemoryVector`.

include/engine/datafacade/contiguous_internalmem_datafacade.hpp
```diff
@@ -211,46 +214,58 @@ class ContiguousInternalMemoryDataFacadeBase : public BaseDataFacade
...
-    util::ShM<extractor::TravelMode, true>::vector m_travel_mode_list;
-    util::ShM<util::guidance::TurnBearing, true>::vector m_pre_turn_bearing;
-    util::ShM<util::guidance::TurnBearing, true>::vector m_post_turn_bearing;
+    util::ShM<util::Coordinate, osrm::storage::MemorySetting::SharedMemory>::vector
+        m_coordinate_list;
+    util::PackedVector<OSMNodeID, osrm::storage::MemorySetting::SharedMemory> m_osmnodeid_list;
+    util::SharedMemoryVector<GeometryID> m_via_geometry_list;
+    util::SharedMemoryVector<NameID> m_name_ID_list;
+    util::SharedMemoryVector<LaneDataID> m_lane_data_id;
+    util::ShM<extractor::guidance::TurnInstruction,
```
TheMarex: Should use `SharedMemoryVector`.

include/engine/datafacade/contiguous_internalmem_datafacade.hpp
```diff
@@ -211,46 +214,58 @@ class ContiguousInternalMemoryDataFacadeBase : public BaseDataFacade
...
+              osrm::storage::MemorySetting::SharedMemory>::vector m_turn_instruction_list;
+    util::ShM<extractor::TravelMode, osrm::storage::MemorySetting::SharedMemory>::vector
```
TheMarex: Should use `SharedMemoryVector`.

include/engine/datafacade/contiguous_internalmem_datafacade.hpp
```diff
@@ -211,46 +214,58 @@ class ContiguousInternalMemoryDataFacadeBase : public BaseDataFacade
...
+        m_travel_mode_list;
+    util::ShM<util::guidance::TurnBearing, osrm::storage::MemorySetting::SharedMemory>::vector
```
TheMarex: Should use `SharedMemoryVector`.

include/engine/datafacade/contiguous_internalmem_datafacade.hpp
```diff
@@ -211,46 +214,58 @@ class ContiguousInternalMemoryDataFacadeBase : public BaseDataFacade
...
+        m_pre_turn_bearing;
+    util::ShM<util::guidance::TurnBearing, osrm::storage::MemorySetting::SharedMemory>::vector
```
TheMarex: Should use `SharedMemoryVector`.

include/engine/datafacade/contiguous_internalmem_datafacade.hpp
```diff
@@ -211,46 +214,58 @@ class ContiguousInternalMemoryDataFacadeBase : public BaseDataFacade
...
     util::NameTable m_names_table;
-    util::ShM<unsigned, true>::vector m_name_begin_indices;
-    util::ShM<bool, true>::vector m_is_core_node;
-    util::ShM<std::uint32_t, true>::vector m_lane_description_offsets;
-    util::ShM<extractor::guidance::TurnLaneType::Mask, true>::vector m_lane_description_masks;
-    util::ShM<TurnPenalty, true>::vector m_turn_weight_penalties;
-    util::ShM<TurnPenalty, true>::vector m_turn_duration_penalties;
+    util::SharedMemoryVector<unsigned> m_name_begin_indices;
+    util::SharedMemoryVector<bool> m_is_core_node;
+    util::ShM<std::uint32_t, osrm::storage::MemorySetting::SharedMemory>::vector
```
TheMarex: Should use `SharedMemoryVector`.

include/engine/datafacade/contiguous_internalmem_datafacade.hpp
```diff
@@ -211,46 +214,58 @@ class ContiguousInternalMemoryDataFacadeBase : public BaseDataFacade
...
-    util::ShM<util::Coordinate, true>::vector m_coordinate_list;
-    util::PackedVector<OSMNodeID, true> m_osmnodeid_list;
-    util::ShM<GeometryID, true>::vector m_via_geometry_list;
-    util::ShM<NameID, true>::vector m_name_ID_list;
-    util::ShM<LaneDataID, true>::vector m_lane_data_id;
-    util::ShM<extractor::guidance::TurnInstruction, true>::vector m_turn_instruction_list;
-    util::ShM<extractor::TravelMode, true>::vector m_travel_mode_list;
-    util::ShM<util::guidance::TurnBearing, true>::vector m_pre_turn_bearing;
-    util::ShM<util::guidance::TurnBearing, true>::vector m_post_turn_bearing;
+    util::ShM<util::Coordinate, osrm::storage::MemorySetting::SharedMemory>::vector
```
TheMarex: Should use `SharedMemoryVector`.

include/engine/datafacade/contiguous_internalmem_datafacade.hpp
```diff
@@ -163,7 +164,7 @@ class ContiguousInternalMemoryAlgorithmDataFacade<algorithm::CoreCH>
     {
         auto core_marker_ptr =
             data_layout.GetBlockPtr<unsigned>(memory_block, storage::DataLayout::CH_CORE_MARKER);
-        util::ShM<bool, true>::vector is_core_node(
+        util::ShM<bool, osrm::storage::MemorySetting::SharedMemory>::vector is_core_node(
```
TheMarex: Should use `SharedMemoryVector`.

include/engine/datafacade/contiguous_internalmem_datafacade.hpp
```diff
@@ -154,7 +155,7 @@ class ContiguousInternalMemoryAlgorithmDataFacade<algorithm::CoreCH>
     : public datafacade::AlgorithmDataFacade<algorithm::CoreCH>
 {
   private:
-    util::ShM<bool, true>::vector m_is_core_node;
+    util::ShM<bool, osrm::storage::MemorySetting::SharedMemory>::vector m_is_core_node;
```
TheMarex: Should use `SharedMemoryVector`.

duizendnegen:
Thanks @TheMarex - good catch. My regex didn't capture all I wanted it seems.

Renamed `SharedMemoryVector` to `vector_view`. What should we then call the `InternalMemoryVector`? Just `vector`? Or leave as-is?

include/util/shared_memory_vector_wrapper.hpp
```diff
@@ -173,12 +175,19 @@ void swap(SharedMemoryWrapper<DataT> &lhs, SharedMemoryWrapper<DataT> &rhs) noex
...
+    using vector =
+        typename std::conditional<MemorySetting == osrm::storage::MemorySetting::SharedMemory,
+                                  SharedMemoryWrapper<DataT>,
+                                  std::vector<DataT>>::type;
 };
+
+template <typename DataT>
+using vector_view = typename util::ShM<DataT, osrm::storage::MemorySetting::SharedMemory>::vector;
+template <typename DataT>
+using InternalMemoryVector =
```
TheMarex: I don't think there is a need for this type alias. This is nothing else then an `std::vector`. If we want to use that explicitly we can just write `std::vector`.

include/util/shared_memory_vector_wrapper.hpp
```diff
@@ -173,12 +175,19 @@ void swap(SharedMemoryWrapper<DataT> &lhs, SharedMemoryWrapper<DataT> &rhs) noex
...
-                                             SharedMemoryWrapper<DataT>,
-                                             std::vector<DataT>>::type;
+    using vector =
+        typename std::conditional<MemorySetting == osrm::storage::MemorySetting::SharedMemory,
+                                  SharedMemoryWrapper<DataT>,
+                                  std::vector<DataT>>::type;
 };
+
+template <typename DataT>
+using vector_view = typename util::ShM<DataT, osrm::storage::MemorySetting::SharedMemory>::vector;
```
TheMarex: Instead of creating a type alias, you can just rename `SharedMemoryWrapper` to `vector_view`.

TheMarex:
What about renaming `osrm::storage::MemorySetting` to `osrm::storage::Ownership` and using the values `Container`, `View`.

include/util/shared_memory_vector_wrapper.hpp
duizendnegen: Not sure, as `SharedMemoryWrapper` is not always a vector (it has a member which is `vector`).

include/extractor/segment_data_container.hpp
```diff
@@ -22,24 +24,24 @@ class CompressedEdgeContainer;
 
 namespace detail
 {
-template <bool UseShareMemory> class SegmentDataContainerImpl;
+template <osrm::storage::Ownership Ownership> class SegmentDataContainerImpl;
```
TheMarex: Ah missed that one: You can omit the `osrm::` for all of these, since we are always in that namespace.

include/storage/shared_memory.hpp
```diff
@@ -30,6 +30,12 @@ namespace osrm
 namespace storage
 {
 
+enum class Ownership
```
TheMarex: I would move this to an own header because otherwise all classes have a dependency on quite a bit of boost interprocess headers that are not needed.

TheMarex:
There is a small `sed` job that could improve readability and maybe a header split. Otherwise this could go in. 👍

include/storage/shared_memory.hpp
duizendnegen: Isn't it the case that `shared_memory.hpp` will be included one way or the other anyway? Or are all cpp files compiled separately? (just for my understanding)
i.e. `edge_based_graph` now doesn't depend on `shared_memory` directly, but depends on `static_graph`, which depends on `shared_memory_vector_wrapper`, ... and that one inevitably points to the actual `shared_memory` header file.

TheMarex: Yeah eventually all headers end up in each cpp file, but usually relying on indirect header includes is considered bad style.

There is actually nothing in `vector_view` that really is specific to shared memory. That whole area in the code base needs some proper naming refresh (we use the same code for process internal memory and shared memory segments...). There are only very few parts of the code that actually need to know about the shared memory stuff, but I guess this is pending a refactor. 😉

PR #3887, opened 2017-03-30
NodeJs based Http Server

### LINKED ISSUE #3832: Reimplement `osrm-routed` in NodeJS
Right now we are shipping our own implementation of:

- HTTP request parser
- async HTTP server
- URL parser (including a bunch of grammars for every API)

We can get rid of most of the `server/` folder by moving the code to NodeJS which already ships with production quality modules for all of this. After #3768 lands we would not induce a weird cyclic dependency anymore and this becomes feasible.

We should be able to replace most of the `boost::spirit` code with https://pegjs.org/.

Having this as node-express based tool would also enable us to easily integrate `osrm-text-instructions` and simplify hooking middle ware like authentication services into this.

daniel-j-h:
Resolves #3832.

## Tasklist
 - [ ] Url parses, grammars
 - [ ] Hook up node-osrm bindings
 - [ ] Switch Cucumber tests to using this
 - [ ] Battle test this
 - [ ] review
 - [ ] adjust for comments

### DIFF
--- a/package.json
+++ b/package.json
@@ -4,19 +4,26 @@
   "private": true,
   "description": "The Open Source Routing Machine is a high performance routing engine written in C++14 designed to run on OpenStreetMap data.",
   "dependencies": {
+    "@types/cors": "^2.8.1",
+    "@types/express": "^4.0.35",
+    "@types/morgan": "^1.7.32",
+    "@types/node": "^7.0.12",
     "chalk": "^1.1.3",
+    "cors": "^2.8.3",
     "cucumber": "^1.2.1",
     "d3-queue": "^2.0.3",
+    "express": "^4.15.2",
     "mkdirp": "^0.5.1",
+    "morgan": "^1.8.1",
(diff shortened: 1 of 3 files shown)


daniel-j-h:
Stale. This is not specced out and feedback wrt. consolidating our internal server never happened.

emiltin:
hmm. still sounds like a good idea, if the recommended way to run osm in production is node, not osrm-routed?

daniel-j-h:
Yes and that's why the issue is still open. I just closed this pull request since nothing was moving on that front and I don't want to keep it around in that state.

Switching osrm-routed to a Node.js based server and away from the crufty C++ HTTP server is still on the todo list just not a priority for us atm since we already do have a functioning HTTP server on top of node-osrm. Would require more coordination with our internal API (cc @freenerd @miccolis).

emiltin:
i see, makes sense

Issue #3844, opened 2017-03-20
Port matching plugin to MLD


TheMarex:
We should port the map matching plugin to MLD. This should be straight forward since we only need to swap out the `ch::getNetworkDistance` function.

TheMarex:
Was merged with #3888.

PR #3888, opened 2017-03-31
MLD map matching plugin

oxidase:
# Issue

Implementation of MLD map matching plugin #3844 

Still has to be decided how to resolve algorithm-dependent functions like [`getNetworkDistance`](https://github.com/Project-OSRM/osrm-backend/blob/mld/map-matching/src/engine/routing_algorithms/map_matching.cpp#L223)

Two possibilities :
* function overloading, all functions must be in one namespace
* argument dependent lookup, all functions are in different namespaces

/cc @TheMarex 

## Tasklist
 - [ ] Clarify usage ADL/overloading
 - [ ] Finish implementation
 - [ ] check tests
 - [ ] review
 - [ ] adjust for comments

## Requirements / Relations

### DIFF
--- a/src/engine/routing_algorithms/map_matching.cpp
+++ b/src/engine/routing_algorithms/map_matching.cpp
@@ -419,14 +419,15 @@ mapMatchingImpl(SearchEngineData &engine_working_data,
     return sub_matchings;
 }
 
-SubMatchingList
-mapMatching(SearchEngineData &engine_working_data,
-            const datafacade::ContiguousInternalMemoryDataFacade<algorithm::CH> &facade,
-            const CandidateLists &candidates_list,
-            const std::vector<util::Coordinate> &trace_coordinates,
-            const std::vector<unsigned> &trace_timestamps,
-            const std::vector<boost::optional<double>> &trace_gps_precision,
-            const bool use_tidying)
+namespace ch
+{
+SubMatchingList mapMatching(SearchEngineData &engine_working_data,
+                            const datafacade::ContiguousInternalMemoryDataFacade<Algorithm> &facade,
+                            const CandidateLists &candidates_list,
+                            const std::vector<util::Coordinate> &trace_coordinates,
+                            const std::vector<unsigned> &trace_timestamps,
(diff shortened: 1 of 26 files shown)


oxidase:
At the moment PR is blocked by unimplemented [ShortestPathSearch](https://github.com/Project-OSRM/osrm-backend/blob/mld/map-matching/src/engine/plugins/match.cpp#L236-L237)
also results can be different because https://github.com/Project-OSRM/osrm-backend/blob/mld/map-matching/src/engine/routing_algorithms/map_matching.cpp#L202-L203 is not used in MLD

src/engine/routing_algorithms/map_matching.cpp
```diff
@@ -419,42 +420,32 @@ mapMatchingImpl(SearchEngineData &engine_working_data,
     return sub_matchings;
 }
 
-SubMatchingList
+template SubMatchingList
```
TheMarex: huh why is there a `template` without `<>`?

unit_tests/engine/base64.cpp
```diff
@@ -43,7 +43,7 @@ BOOST_AUTO_TEST_CASE(hint_encoding_decoding_roundtrip)
 
     const Coordinate coordinate;
     const PhantomNode phantom;
-    const osrm::test::MockDataFacade<osrm::engine::algorithm::CH> facade{};
+    const osrm::test::MockDataFacade<osrm::engine::datafacade::CH> facade{};
```
TheMarex: I think it might be a little bit confusing if we use the `datafacade` internal name here. Doing something like:

`using CH = osrm::engine::ch::algorithm` would be more obvious when reading this code.

src/engine/routing_algorithms/map_matching.cpp
oxidase: Here is "14.7.2 Explicit instantiation" with syntax `extern_opt template declaration` and not a template specialization like `template<> `

oxidase:
Rebased onto master after #3895 merge. Ready for the second review round

PR #3892, opened 2017-04-02
Refactors serialisation code to avoid duplicate data loading logic

TheMarex:
# Issue

Currently we duplicate the logic to load files between `storage.cpp` and the `io.hpp` layers.
This refactor splits the `io.hpp` logic into two parts:

- `serialisation.hpp` That writes/reads a data structure to a `File{Reader,Writer}`
- `files.hpp` that handles reading the actual files `.osrm.*` and uses the right (de)serialization logic.
- `storage.cpp` now uses the file loading logic to load data to the non-owning version of each data structure.

## Tasklist
 - [x] Fix geometries loading
 - [x] Port `MultiLevelGraph`, `MultiLevelPartition`, `CellStorage` to use the file loaders.
 - [x] review
 - [x] adjust for comments

## Requirements / Relations
Should go in after #3875.


duizendnegen:
This is fantastic - great new part to also start cross-platform IO from! Really excited to see this.

include/contractor/files.hpp
```diff
@@ -0,0 +1,46 @@
...
+
+namespace osrm
+{
+namespace contractor
+{
+namespace files
+{
+
+// reads .osrm.hsgr file
+template <bool UseSharedMemory>
```
daniel-j-h: Needs to be synchronized with the enum switch pull request

include/contractor/query_graph.hpp
```diff
@@ -0,0 +1,27 @@
+#ifndef OSRM_CONTRACTOR_QUERY_GRAPH_HPP
+#define OSRM_CONTRACTOR_QUERY_GRAPH_HPP
+
+#include "contractor/query_edge.hpp"
+
+#include "util/static_graph.hpp"
+#include "util/typedefs.hpp"
+
+#include <tuple>
```
daniel-j-h: Why tuple, typedefs?

include/extractor/files.hpp
```diff
@@ -0,0 +1,160 @@
...
+
+namespace osrm
+{
+namespace extractor
+{
+namespace files
+{
+
+// reads .osrm.nodes
+template <bool UseShareMemory>
```
daniel-j-h: Same, also below

include/partition/multi_level_partition.hpp
```diff
@@ -314,11 +323,18 @@ template <bool UseShareMemory> class MultiLevelPartitionImpl final
         }
     }
 
-    //! this is always owned by this class because it is so small
-    LevelData level_data;
+    LevelDataPtr level_data;
```
daniel-j-h: Hm I think if you `= {};` here you don't need the ctors below

include/util/deallocating_vector.hpp
```diff
@@ -239,6 +239,19 @@ class DeallocatingVector
...
+        bucket_list = std::move(other.bucket_list);
+        current_size = std::move(other.current_size);
+    }
+
+    DeallocatingVector &operator=(DeallocatingVector &&other)
+    {
+        bucket_list = std::move(other.bucket_list);
+        current_size = std::move(other.current_size);
+        return *this;
+    }
```
daniel-j-h: I don't understand the move ctor and move assign op here. The dealloc. vec has a size_t and a vector member attribute. Why don't we initialize the size_t here? (The move ops are basically what swap is supposed to do).

https://github.com/Project-OSRM/osrm-backend/blob/86a31f1080c7fe458030b341a84deab3ec826b5f/include/util/deallocating_vector.hpp#L227-L228

Hm, `= default();`-ing both should have the same effect and be correct, no?

include/util/shared_memory_vector_wrapper.hpp
```diff
@@ -80,6 +81,18 @@ template <typename DataT> class SharedMemoryWrapper
...
+    // operations on this object
+    void resize(std::size_t size) const
+    {
+        if (m_size != size)
+        {
+            throw util::exception("Invalid resize " + std::to_string(size) +
+                                  " on immutable vector view of size " + std::to_string(m_size) +
+                                  ".");
+        }
+    }
```
daniel-j-h: Why do we provide the function then anyway?

include/util/shared_memory_vector_wrapper.hpp
```diff
@@ -126,20 +139,21 @@ template <typename DataT> class SharedMemoryWrapper
...
     std::size_t m_size;
 
   public:
-    SharedMemoryWrapper() : m_ptr(nullptr), m_size(0) {}
+    using value_type = bool;
 
-    SharedMemoryWrapper(unsigned *ptr, std::size_t size) : m_ptr(ptr), m_size(size) {}
+    vector_view() : m_ptr(nullptr), m_size(0) {}
+
+    vector_view(unsigned *ptr, std::size_t size) : m_ptr(ptr), m_size(size) {}
```
daniel-j-h: Here and in reset: should the unsigned ptrs be const? the vector view is not able to mutate after all. Also the member m_ptr should probably be a `const unsigned *` (read-only data, but ptr can be swapped out, e.g. inreset)

include/contractor/files.hpp
oxidase: `UseSharedMemory` is non relevant detail here, better to make `template<typename QueryGrahp>` in this PR

include/extractor/files.hpp
```diff
@@ -0,0 +1,160 @@
...
+inline void
+writeNodes(const boost::filesystem::path &path,
+           const typename util::ShM<util::Coordinate, UseShareMemory>::vector &coordinates,
+           const util::detail::PackedVector<OSMNodeID, UseShareMemory> &osm_node_ids)
+{
+    const auto fingerprint = storage::io::FileWriter::GenerateFingerprint;
+    storage::io::FileWriter writer{path, fingerprint};
+
+    writer.SerializeVector(coordinates);
```
oxidase: From outside it looks strange to have both `writer.SerializeVector` and `util::serialization::write(writer, ...)` Would it make sense to unify interface as `util::serialization::write(writer, vector)` ?

Or even with syntactic sugar `writer << vector` ?

src/storage/storage.cpp
```diff
@@ -557,151 +553,146 @@ void Storage::PopulateData(const DataLayout &layout, char *memory_ptr)
...
+        util::vector_view<util::Coordinate> coordinates(
+            coordinates_ptr, layout.num_entries[DataLayout::COORDINATE_LIST]);
+        util::PackedVectorView<OSMNodeID> osm_node_ids;
+        osm_node_ids.reset(osmnodeid_ptr, layout.num_entries[DataLayout::OSM_NODE_ID_LIST]);
 
-        serialization::readNodes(nodes_file,
-                                 coordinates_ptr,
-                                 osmnodeid_list,
-                                 layout.num_entries[DataLayout::COORDINATE_LIST]);
+        extractor::files::readNodes<true>(config.nodes_data_path, coordinates, osm_node_ids);
```
oxidase: `true` can be derived from arguments, but better to use template arguments for `coordinates` and `osm_node_ids`

include/extractor/files.hpp
TheMarex: Yeah changing `SerializeVector` to the new `serialization::{read,write}`interface is on my TODO list.

include/partition/multi_level_partition.hpp
TheMarex: Is {} the same as `make_unique<T>()`?

daniel-j-h: same as unique_ptr<T>'s default ctor

include/util/shared_memory_vector_wrapper.hpp
TheMarex: For comparability with the `std::vector` interface code. With that we can use the same code for de-serializing a vector and a vector view (resize is NOP on a vector that is already of correct size and a NOP here too).

include/util/deallocating_vector.hpp
TheMarex: I don't quite understand what you mean. The `DeallocationVector` had an implicitly deleted move constructor.

I had the problem that if you pass a `DeallocationVector` by value, it generated double free errors: If you copy `bucket_list` which is a vector of pointer, there was no deep-copy. The bucket's where owned by two `DeallocationVector` instances and caused problems.

I fixed this by adding a move constructor that safely allows to transfer ownership of buckets. The copy constructor should be deleted though, I don't think there is much sense in allowing deep copies here.

include/partition/multi_level_partition.hpp
TheMarex: Hrm this doesn't work. `= {}` seems to create an empty unique ptr. I need a unique ptr with a real value in it.

include/contractor/query_graph.hpp
TheMarex: Good catch, I think this is unnecessary.

include/util/shared_memory_vector_wrapper.hpp
TheMarex: Actually we do need write access to views now. That allows us to unify shared memory loading code and generic de-serialization code.

include/util/shared_memory_vector_wrapper.hpp
TheMarex: Never mind, I refactored this away now.

TheMarex:
@oxidase @daniel-j-h I think all PR comments are addressed now, please check if this works for you.

include/util/deallocating_vector.hpp
daniel-j-h: 1/ Check the move ctor / assign op [here](https://github.com/Project-OSRM/osrm-backend/blob/f650aa9df2cee6e99c994e2964d4c84c0a589e04/include/util/deallocating_vector.hpp#L246-L257): aren't we missing to copy over the `current_size` member?

2/ what I meant was:

    DeallocatingVector(DeallocatingVector &&) = default;
    DeallocatingVector &operator=(DeallocatingVector &&) = default;

Isn't this what we are trying to do manually here?

TheMarex:
@oxidase 👍 fixed and waiting for travis before merging.

PR #3696, opened 2017-02-12
Make edge weights and durations strongly defined

oxidase:
# Issue

For #3601 PR defines `EdgeWeight`, `TurnPenalty`, `EdgeDuration`, `TurnDuration` as strong types. Also fixes #3677 by defining 30 bits size for duration that roughly corresponds to 621 days in deciseconds. 

## Tasklist
 - [ ] Fix TODOs in the PR
 - [ ] Fix fallback to durations in traffic speed updates
 - [ ] Refine explicit `static_cast` by defining operators
 - [ ] review
 - [ ] adjust for comments

Issue #3601, opened 2017-01-23
Make edge weight a strongly defined type


oxidase:
Working on #2399 found different weight types 
* `unsigned` https://github.com/Project-OSRM/osrm-backend/blob/a5d0707dd0f081c6b4b98fdc36201567a2cfbbfe/include/contractor/contractor_graph.hpp#L29
*  `std::int32_t` https://github.com/Project-OSRM/osrm-backend/blob/a5d0707dd0f081c6b4b98fdc36201567a2cfbbfe/src/engine/routing_algorithms/routing_base.cpp#L23
* `int` https://github.com/Project-OSRM/osrm-backend/blob/a5d0707dd0f081c6b4b98fdc36201567a2cfbbfe/src/engine/routing_algorithms/many_to_many.cpp#L131

PR #2399 unifies weight types as `EdgeWeight` but compile-time type checking is required to check all places. 
It would also good to define strongly typed `EdgeDuration`, `TurnWeight` and `TurnDuration` to have compile-type checks and avoid problems by mixing edge/turn weights and durations.

daniel-j-h:
Yes, yes please. I poked @danpat often enough some time ago that we ended up building and incorporating

https://github.com/Project-OSRM/osrm-backend/blob/master/include/util/strong_typedef.hpp#L43
https://github.com/Project-OSRM/osrm-backend/blob/master/include/util/typedefs.hpp#L40-L44

which is similar to the one in Boost, except that it does not allow implicit conversions.

danpat:
@oxidase yeah, you want to take a look and see if `OSRM_STRONG_TYPEDEF` is usable here?  Check the `OSMNodeID` type as an example (also, include/util/coordinate.hpp where we use it to prevent lon/lat swap mistakes).

It *should* be fully optimized away under `-O3`, and will give the compile-time checking you're after.

TheMarex:
This has no clear priority right now. We should have a look at this when we got some downtime for refactors.

PR #3895, opened 2017-04-03
MLD shortest path plugin

oxidase:
# Issue

Implementation of the MLD shortest path plugin. It uses algorithm-aware Heaps 

There are still problems that must be fixed:
* some tests with self-loops(?) fail
* [`phantom_nodes`](https://github.com/Project-OSRM/osrm-backend/blob/784c5f05922b0573fc84f74c1b3b55e9a9915ccc/include/engine/routing_algorithms/routing_base_ch.hpp#L439) added to `search` functions parameters. It can be fixed via algorithm specific parameter packs.
* `SearchEngineData` for MLD needs ["fake" core heaps](https://github.com/Project-OSRM/osrm-backend/blob/784c5f05922b0573fc84f74c1b3b55e9a9915ccc/src/engine/search_engine_data.cpp#L162-L164). It can be solved only by moving access to core heaps into `corech` namespace. it requires refactoring of heaps usage.

## Tasklist
 - [x] all tests pass for MLD via tests
 - [x] fixed `{ch,corech,mld}::search` interface
 - [x] refactor `SearchEngineData` usage
 - [x] review
 - [x] adjust for comments

## Requirements / Relations
The PR is the pre-condition for #3888

### DIFF
--- a/include/engine/routing_algorithms/routing_base_ch.hpp
+++ b/include/engine/routing_algorithms/routing_base_ch.hpp
@@ -365,10 +365,11 @@ inline void search(const datafacade::ContiguousInternalMemoryDataFacade<Algorith
                    SearchEngineData<Algorithm>::QueryHeap &reverse_heap,
                    SearchEngineData<Algorithm>::QueryHeap &,
                    SearchEngineData<Algorithm>::QueryHeap &,
-                   std::int32_t &weight,
+                   EdgeWeight &weight,
                    std::vector<NodeID> &packed_leg,
                    const bool force_loop_forward,
                    const bool force_loop_reverse,
+                   const PhantomNodes & /*phantom_nodes*/,
                    const int duration_upper_bound = INVALID_EDGE_WEIGHT)
 {
     search(facade,
@@ -380,10 +381,6 @@ inline void search(const datafacade::ContiguousInternalMemoryDataFacade<Algorith
            force_loop_reverse,
            duration_upper_bound);
 }
-
-bool needsLoopForward(const PhantomNode &source_phantom, const PhantomNode &target_phantom);
-
-bool needsLoopBackwards(const PhantomNode &source_phantom, const PhantomNode &target_phantom);
 
 double getPathDistance(const datafacade::ContiguousInternalMemoryDataFacade<ch::Algorithm> &facade,
                        const std::vector<NodeID> &packed_path,
@@ -437,6 +434,7 @@ void search(const datafacade::ContiguousInternalMemoryDataFacade<corech::Algorit
             std::vector<NodeID> &packed_leg,
             const bool force_loop_forward,
             const bool force_loop_reverse,
+            const PhantomNodes &phantom_nodes,
             int duration_upper_bound = INVALID_EDGE_WEIGHT);
 
 // Requires the heaps for be empty
@@ -451,6 +449,17 @@ getNetworkDistance(const datafacade::ContiguousInternalMemoryDataFacade<corech::
                    const PhantomNode &source_phantom,
                    const PhantomNode &target_phantom,
                    int duration_upper_bound = INVALID_EDGE_WEIGHT);
+
+template <typename RandomIter, typename FacadeT>
+void unpackPath(const FacadeT &facade,
+                RandomIter packed_path_begin,
+                RandomIter packed_path_end,
+                const PhantomNodes &phantom_nodes,
+                std::vector<PathData> &unpacked_path)
+{
+    return ch::unpackPath(facade, packed_path_begin, packed_path_end, phantom_nodes, unpacked_path);
+}
+
 } // namespace corech
 
 } // namespace routing_algorithms
--- a/include/engine/routing_algorithms/routing_base_ch.hpp
+++ b/include/engine/routing_algorithms/routing_base_ch.hpp
@@ -23,7 +23,7 @@ namespace ch
 
 // Stalling
 template <bool DIRECTION, typename HeapT>
-bool stallAtNode(const datafacade::ContiguousInternalMemoryDataFacade<algorithm::CH> &facade,
+bool stallAtNode(const datafacade::ContiguousInternalMemoryDataFacade<Algorithm> &facade,
                  const NodeID node,
                  const EdgeWeight weight,
                  const HeapT &query_heap)
@@ -49,7 +49,7 @@ bool stallAtNode(const datafacade::ContiguousInternalMemoryDataFacade<algorithm:
 }
 
 template <bool DIRECTION>
-void relaxOutgoingEdges(const datafacade::ContiguousInternalMemoryDataFacade<algorithm::CH> &facade,
+void relaxOutgoingEdges(const datafacade::ContiguousInternalMemoryDataFacade<Algorithm> &facade,
                         const NodeID node,
                         const EdgeWeight weight,
                         SearchEngineData::QueryHeap &heap)
@@ -113,7 +113,7 @@ we need to add an offset to the termination criterion.
 static constexpr bool ENABLE_STALLING = true;
 static constexpr bool DISABLE_STALLING = false;
 template <bool DIRECTION, bool STALLING = ENABLE_STALLING>
-void routingStep(const datafacade::ContiguousInternalMemoryDataFacade<algorithm::CH> &facade,
+void routingStep(const datafacade::ContiguousInternalMemoryDataFacade<Algorithm> &facade,
                  SearchEngineData::QueryHeap &forward_heap,
                  SearchEngineData::QueryHeap &reverse_heap,
                  NodeID &middle_node_id,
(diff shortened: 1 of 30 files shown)


oxidase:
MLD requires additional tie breaking ac4f1b2 for the test `features/testbot/mode.feature:153`.
The MLD EBG of the test is
mode feature 153

The green point is the snapped source phantom `id:1, offset 0`. The red points are snapped target phantoms `id:1, offset:900` and `id:0, offset:0` 

The first MLD path has `total_weight_to_forward=900` and `packed_leg_to_forward_begin` contains nodes `(1,0)`. The second MLD result has also `total_weight_to_reverse=900` and `packed_leg_to_reverse_begin` is empty.

The problem is that in the first case guidance string is `river upstream,river upstream,river upstream,river downstream,river downstream`
that incorrectly contains a zero weight u-turn.
In the second case guidance is `river upstream,river upstream,river upstream,river upstream`  without a u-turn.

In case of the CH algorithm the forward path is not existing, because CH EBG has no edge-based edges 2 and 3.

src/engine/routing_algorithms/routing_base_ch.cpp
```diff
@@ -139,7 +139,124 @@ void search(const datafacade::ContiguousInternalMemoryDataFacade<algorithm::CH>
     }
 }
 
-// assumes that heaps are already setup correctly.
+double getPathDistance(const datafacade::ContiguousInternalMemoryDataFacade<Algorithm> &facade,
```
TheMarex: Why does this function need to be duplicated? Is there a way to share the implementation with the MLD variation?

src/engine/routing_algorithms/shortest_path.cpp
```diff
@@ -453,56 +457,55 @@ shortestPathSearchImpl(SearchEngineData &engine_working_data,
                  total_weight_to_reverse != INVALID_EDGE_WEIGHT);
 
     // We make sure the fastest route is always in packed_legs_to_forward
-    if (total_weight_to_forward > total_weight_to_reverse)
+    if (total_weight_to_forward < total_weight_to_reverse ||
+        (total_weight_to_forward == total_weight_to_reverse &&
+         total_packed_path_to_forward.size() < total_packed_path_to_reverse.size()))
```
TheMarex: Can be written as `std::tie(total_weight_to_forward, total_packed_path_to_forward.size()) < std::tie(total_weight_to_reverse, total_packed_path_to_reverse.size())`

src/engine/search_engine_data.cpp
```diff
@@ -93,25 +110,97 @@ void SearchEngineData::InitializeOrClearManyToManyThreadLocalStorage(const unsig
...
+template SearchEngineData<CoreCH>::SearchEngineHeapPtr SearchEngineData<CoreCH>::forward_heap_1;
+template SearchEngineData<CoreCH>::SearchEngineHeapPtr SearchEngineData<CoreCH>::reverse_heap_1;
+template SearchEngineData<CoreCH>::SearchEngineHeapPtr SearchEngineData<CoreCH>::forward_heap_2;
+template SearchEngineData<CoreCH>::SearchEngineHeapPtr SearchEngineData<CoreCH>::reverse_heap_2;
+template SearchEngineData<CoreCH>::SearchEngineHeapPtr SearchEngineData<CoreCH>::forward_heap_3;
+template SearchEngineData<CoreCH>::SearchEngineHeapPtr SearchEngineData<CoreCH>::reverse_heap_3;
+template SearchEngineData<CoreCH>::ManyToManyHeapPtr SearchEngineData<CoreCH>::many_to_many_heap;
+
+template void
+SearchEngineData<CoreCH>::InitializeOrClearFirstThreadLocalStorage(unsigned number_of_nodes);
```
TheMarex: Hrm we might want to simplify the interface here a little bit. There is no reason to ever get a reference to one of the heaps without clearing it first. So what about:

```
std::array<SearchEngineHeapPtr, MAX_HEAPS> heaps;

SearchHeap& GetHeap(HeapID id, unsigned number_of_nodes) {
     // init or clear code
     return *heaps[id];
}
```

That would also enable us to make all `{forward,backward}_heap_{1,2,3}` private and hide the thread-specific pointer. Also there is no reason why we have a distinction between forward and backward heaps.

TheMarex: Also no need for this to be static, we have exactly one instance of `SearchEngineData`.

include/engine/routing_algorithms/routing_base_mld.hpp
```diff
@@ -279,6 +330,63 @@ search(const datafacade::ContiguousInternalMemoryDataFacade<algorithm::MLD> &fac
...
+                   SearchEngineData<Algorithm>::QueryHeap &forward_heap,
+                   SearchEngineData<Algorithm>::QueryHeap &reverse_heap,
+                   SearchEngineData<Algorithm>::QueryHeap &,
+                   SearchEngineData<Algorithm>::QueryHeap &,
+                   EdgeWeight &weight,
+                   std::vector<NodeID> &packed_leg,
+                   const bool force_loop_forward,
+                   const bool force_loop_reverse,
+                   const PhantomNodes &phantom_nodes,
+                   const int duration_upper_bound = INVALID_EDGE_WEIGHT)
```
daniel-j-h: same

include/engine/search_engine_data.hpp
```diff
@@ -45,33 +38,49 @@ struct SearchEngineData
...
+};
 
-    void InitializeOrClearSecondThreadLocalStorage(const unsigned number_of_nodes);
+template <> struct SearchEngineData<routing_algorithms::mld::Algorithm>
+{
+    using QueryHeap = util::BinaryHeap<NodeID,
+                                       NodeID,
+                                       EdgeWeight,
+                                       MultiLayerDijkstraHeapData,
+                                       util::UnorderedMapStorage<NodeID, int>>;
```
daniel-j-h: int?

src/engine/routing_algorithms/routing_base.cpp
```diff
@@ -0,0 +1,28 @@
...
+               target_phantom.GetForwardWeightPlusOffset();
+}
+
+bool needsLoopBackwards(const PhantomNode &source_phantom, const PhantomNode &target_phantom)
+{
+    return source_phantom.reverse_segment_id.enabled && target_phantom.reverse_segment_id.enabled &&
+           source_phantom.reverse_segment_id.id == target_phantom.reverse_segment_id.id &&
+           source_phantom.GetReverseWeightPlusOffset() >
+               target_phantom.GetReverseWeightPlusOffset();
+}
```
daniel-j-h: Could you explain these two to me in the context of this pr?

src/engine/routing_algorithms/routing_base_ch.cpp
oxidase: yes, it is move one scope level up in the map matching branch. i would keep it as only CH-related in this PR

src/engine/routing_algorithms/routing_base.cpp
oxidase: MLD should use the loops forcing in the same way as CH. So i just moved functions in common `routing_algorithm` namespace and used forcing loops flags to prevent too early exit in MLD.

include/engine/routing_algorithms/routing_base_ch.hpp
```diff
@@ -382,6 +382,40 @@ inline void search(const datafacade::ContiguousInternalMemoryDataFacade<algorith
...
+// Requires the heaps for be empty
+// If heaps should be adjusted to be initialized outside of this function,
+// the addition of force_loop parameters might be required
+double
+getNetworkDistance(const datafacade::ContiguousInternalMemoryDataFacade<ch::Algorithm> &facade,
+                   SearchEngineData<Algorithm>::QueryHeap &forward_heap,
+                   SearchEngineData<Algorithm>::QueryHeap &reverse_heap,
+                   const PhantomNode &source_phantom,
+                   const PhantomNode &target_phantom,
+                   int duration_upper_bound = INVALID_EDGE_WEIGHT);
```
oxidase: the reason why i did not change is at https://github.com/Project-OSRM/osrm-backend/blob/e568b600f0c4ad5a2cd9b8f97868ef9b1b7fc6b2/src/engine/routing_algorithms/map_matching.cpp#L201-L202  
basically the value is neither weight not duration. It should fixed in #3696

src/engine/routing_algorithms/shortest_path.cpp
oxidase: not really, `std::tie` requres non-const references as
```
template< class... Types >
tuple<Types&...> tie( Types&... args );
```

oxidase:
@TheMarex @daniel-j-h please could you take a look before merge?

I did not change in this PR:
* `int` -> `EdgeWeight` it is not used in MLD and i think it is better to  fix it  in #3601 with guarantee that all all such cases are covered.

* Heaps are broken on the architecture level, i  would also dedicate a separate PR to fix it. The main issue is that input phantom nodes must be  placed into heaps before `search` so any plugin code should call `Clear`  before inserting phantom nodes. This requires to add "core" heap into MLD heaps structure and some [workarounds](https://github.com/Project-OSRM/osrm-backend/blob/59350f0b16288b6ebad6f7bf4cc5315e53ddf3bb/include/engine/routing_algorithms/routing_base_ch.hpp#L362-L384). Also separate static members are not so good. Refactoring of this issue would require big code change in the engine code base.

TheMarex:
@oxidase fine by me. If you could ticket the remaining issues that would be great. 👍

PR #3897, opened 2017-04-04
Don't scan back clique arcs in customization

TheMarex:
# Issue

Fix a performance problem in the customization that gives a factor ~3 speedup on Bayern.

## Tasklist
 - [x] review
 - [x] adjust for comments

### DIFF
--- a/include/util/binary_heap.hpp
+++ b/include/util/binary_heap.hpp
@@ -19,15 +19,21 @@ namespace util
 template <typename NodeID, typename Key> class GenerationArrayStorage
 {
     using GenerationCounter = std::uint16_t;
-  public:
-    explicit GenerationArrayStorage(std::size_t size) : positions(size, 0), generation(1), generations(size, 0) {}
 
-    Key &operator[](NodeID node) {
+  public:
+    explicit GenerationArrayStorage(std::size_t size)
+        : positions(size, 0), generation(1), generations(size, 0)
+    {
+    }
+
+    Key &operator[](NodeID node)
+    {
         generation[node] = generation;
         return positions[node];
     }
 
-    Key peek_index(const NodeID node) const {
+    Key peek_index(const NodeID node) const
+    {
         if (generations[node] < generation)
         {
             return std::numeric_limits<Key>::max();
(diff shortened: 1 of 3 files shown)


include/util/binary_heap.hpp
```diff
@@ -16,6 +16,40 @@ namespace osrm
...
+  public:
+    explicit GenerationArrayStorage(std::size_t size) : positions(size, 0), generation(0), generations(size, 0) {}
+
+    Key &operator[](NodeID node) {
+        generation[node] = generation;
+        return positions[node];
+    }
+
+    Key peek_index(const NodeID node) const {
+        if (generations[node] < generation)
```
oxidase: If the initial generation is 0 and `Clear` was not called before then  `generations[node] < generation` will be false, so `peak_index` will return 0 for nodes that were not touched. It is different wrt `UnorderedMapStorage` and different in other way wrt `ArrayStorage` 

Also i guess  if `ArrayStorage` will be removed then `heap[0].weight = std::numeric_limits<Weight>::min();` can be also removed from `BinaryHeap`

include/util/binary_heap.hpp
```diff
@@ -16,6 +16,40 @@ namespace osrm
...
+        {
+            return std::numeric_limits<Key>::max();
+        }
+        return positions[node];
+    }
+
+    void Clear() {
+        generation++;
+        if (generation == std::numeric_limits<GenerationCounter>::max())
```
oxidase: Here must be check for a wrap-around as `if (generation == 0)`

include/util/binary_heap.hpp
```diff
@@ -16,6 +16,40 @@ namespace osrm
...
+        {
+            std::fill(generations.begin(), generations.end(), 0);
+            std::fill(positions.begin(), positions.end(), 0);
```
oxidase: this is not needed if `generation` will be set to 1 here and in the constructor

include/util/binary_heap.hpp
TheMarex: The main problem with the generation counted approach is, that right now we don't really need it. Since we have the indirection over `inserted_nodes` in the `BinaryHeap` actually clearing the vector is not necessary (even for `ArrayStorage`). 

To make this really effective we would need to rip out the `inserted_nodes` array, replace it with a generation counted arrays of `Weight` and `Data`. Though this has again the memory impact of at least 400MB per thread (For ~20 0Million nodes). Might be interesting for the customization and contractor though.

TheMarex:
@oxidase I think all PR comments are addressed now, please check if this works for you.

Issue #3902, opened 2017-04-05
Port ManyToMany plugin to MLD


TheMarex:
We want to be able to efficiently compute many to many queries even on MLD. There are two implementations possible here:

1. Port the current search bucket based approach to MLD:
    - Probably slow since the search space will be huge
    - No additional data needed

2. Implement GRASP:
   - Needs a special partition scheme to be fast
   - Possibly needs parallelization to be efficient
   - Needs a lot of memory for the naive implementation
   - Very fast queries possible (1:100k in 150ms sequential)

oxidase:
Closing here as the first point is implemented. GRASP speed up is worth in case of one-to-all queries (1:10k and more) but it will also increase memory consumption due to introduction a downward graph G↓.

Feel free to open a research ticket about GRASP memory/performance trade-offs with respect to the current implementation.

PR #3903, opened 2017-04-05
Add GRASP implementation for many to many

TheMarex:
# Issue

This implements the GRASP approach of https://github.com/Project-OSRM/osrm-backend/issues/3902. This is mostly a prototype right now, because this approach has a few unknowns w.r.t to memory consumption <> speed trade-off.

## Tasklist
 - [x] Load data into server
 - [x] Implement query algorithm
 - [ ] Reduce memory consumption to something manageable 
 - [ ] review
 - [ ] adjust for comments

## Requirements / Relations

This PR is based on #3892

### DIFF
--- a/include/customizer/cell_customizer.hpp
+++ b/include/customizer/cell_customizer.hpp
@@ -16,12 +16,14 @@ namespace customizer
 
 class CellCustomizer
 {
-  private:
+  protected:
     struct HeapData
     {
         bool from_clique;
         EdgeDuration duration;
     };
+    struct TerminateEarly;
+    struct TerminateOnExhaution;
 
   public:
     using Heap =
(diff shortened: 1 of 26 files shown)


include/customizer/cell_customizer.hpp
```diff
@@ -66,28 +92,30 @@ class CellCustomizer
                 weight =
                     heap.WasInserted(destination) ? heap.GetKey(destination) : INVALID_EDGE_WEIGHT;
             }
+
```
oxidase: The part above can be moved into a callback on line 36

include/customizer/cell_customizer.hpp
```diff
@@ -33,16 +33,42 @@ class CellCustomizer
...
+                  [this](const GraphT &graph,
+                         Heap &heap,
+                         partition::CellStorage &cells,
+                         LevelID level,
+                         CellID id) { Customize(graph, heap, cells, level, id); });
+    }
+
+  protected:
+    template <typename GraphT, typename RowFn>
+    void Customize(const GraphT &graph,
```
oxidase: There are too many `Customize` methods now. It is better to rename like `CustomizeAllCells`, `CustomizeCell`, `CustomizeRow`

unit_tests/customizer/grasp_customization.cpp
```diff
@@ -0,0 +1,150 @@
...
+            auto source = grasp_storage.GetSource(edge);
+            auto weight = grasp_storage.GetEdgeData(edge).weight;
+            down_edges.push_back(MockEdge{source, node, weight});
+        }
+    }
+    std::sort(down_edges.begin(), down_edges.end());
+
+    std::vector<MockEdge> down_edges_reference = {{0, 5, 1},
+                                                  {0, 6, 2},
+                                                  {0, 7, 3},
```
oxidase: I  think here must be 2 instead of 3, because if edge `5->7` is added the path `0->5->7` is shorter than `0->1->6->7` in paper.

unit_tests/customizer/grasp_customization.cpp
```diff
@@ -0,0 +1,150 @@
...
+                                                  {1, 5, 1},
+                                                  {1, 6, 1},
+                                                  {1, 7, 2},
+                                                  {2, 5, INVALID_EDGE_WEIGHT},
+                                                  {2, 6, 2},
```
oxidase: `2->6` does not exist if the graph is directed

if the graph has undirected edges on lines 53,54 as `true,true` then path is ok, but also `2->5` will have length 2

unit_tests/customizer/grasp_customization.cpp
```diff
@@ -0,0 +1,150 @@
...
+                                                  {2, 7, 1},
+                                                  {3, 8, 1},
+                                                  {3, 9, 2},
```
oxidase: `3->9` does not exist

unit_tests/customizer/grasp_customization.cpp
```diff
@@ -0,0 +1,150 @@
...
+                                                  {4, 8, 2},
```
oxidase: `4->8` does not exist

unit_tests/customizer/grasp_customization.cpp
```diff
@@ -0,0 +1,150 @@
...
+                                   {0, 5, 1},
+                                   {1, 4, 1},
+                                   {1, 5, 1},
+                                   {1, 6, 1},
+                                   {2, 4, 1},
+                                   {2, 7, 1},
+                                   {2, 12, 1},
+                                   {3, 8, 1},
+                                   {4, 9, 1},
+                                   {5, 7, 1},
```
oxidase: The directed graph with the edge  is m and the corresponding downwards graph is 
s

I would suggest to add another test without the edge and with bidirectional edges, so the graph will
nm and the downwards graphs 
ns

TheMarex:
One major problem that I again forgot is that we only have `weight` data for every clique arc in the overlay. Returning fill duration matrices would be blocked by multi-metric support. Will work with `weight` values for now to play with the query algorithm.

oxidase:
@TheMarex it is better to completely abstract "weight" values later by decoupling from edge data structures and after that duration-, routability- or  distance-based  matrix computations will be for free. But it is really big refactoring step that can be binded with data size reduction.

TheMarex:
GRASP didn't work well for us, currently this is just a simple one-to-all.

Issue #3520, opened 2017-01-05
Put fingerprints/headers on all datafiles


danpat:
https://github.com/Project-OSRM/osrm-backend/pull/3467 adds semver-like versioning to our files.

We should add this 8 byte header to all datafiles, and check it when we read things back.

duizendnegen:
I'm interested in maybe picking this up along with creating a serialization layer for the different modules - might work well combined.

Or, were you thinking in terms of adding `VerifyFingerprint` & `GenerateFingerprint` to every `FileReader`/`FileWriter` call? That'd be very easy to get done, can also pick that up.

danpat:
@duizendnegen the second option - no reason to not have a header on every file that gets verified IMO.

duizendnegen:
This one can be closed, #3951 tracks the remainder.

PR #3904, opened 2017-04-05
Unify fingerprinting & Count64

duizendnegen:
# Issue

#3520

## Tasklist
 - [x] convert serialization to use fingerprint generation and validation
 - [x] replace Count32 with Count64
 - [x] review
 - [x] adjust for comments

## Requirements / Relations
This should go in after #3892 (the other way around is too much hassle)

### DIFF
--- a/include/engine/engine.hpp
+++ b/include/engine/engine.hpp
@@ -185,7 +185,7 @@ bool Engine<routing_algorithms::corech::Algorithm>::CheckCompability(const Engin
 
         auto mem = storage::makeSharedMemory(barrier.data().region);
         auto layout = reinterpret_cast<storage::DataLayout *>(mem->Ptr());
-        return layout->GetBlockSize(storage::DataLayout::CH_CORE_MARKER) > 16;
+        return layout->GetBlockSize(storage::DataLayout::CH_CORE_MARKER) > sizeof(std::uint64_t) + sizeof(util::FingerPrint);
     }
     else
     {
--- a/include/engine/engine.hpp
+++ b/include/engine/engine.hpp
@@ -185,7 +185,8 @@ bool Engine<routing_algorithms::corech::Algorithm>::CheckCompability(const Engin
 
         auto mem = storage::makeSharedMemory(barrier.data().region);
         auto layout = reinterpret_cast<storage::DataLayout *>(mem->Ptr());
-        return layout->GetBlockSize(storage::DataLayout::CH_CORE_MARKER) > sizeof(std::uint64_t) + sizeof(util::FingerPrint);
+        return layout->GetBlockSize(storage::DataLayout::CH_CORE_MARKER) >
+               sizeof(std::uint64_t) + sizeof(util::FingerPrint);
     }
     else
     {
--- a/include/engine/engine.hpp
+++ b/include/engine/engine.hpp
@@ -22,6 +22,7 @@
 #include "engine/status.hpp"
 #include "util/exception.hpp"
 #include "util/exception_utils.hpp"
+#include "util/fingerprint.hpp"
 #include "util/json_container.hpp"
 
 #include <memory>
@@ -184,7 +185,7 @@ bool Engine<routing_algorithms::corech::Algorithm>::CheckCompability(const Engin
 
         auto mem = storage::makeSharedMemory(barrier.data().region);
         auto layout = reinterpret_cast<storage::DataLayout *>(mem->Ptr());
-        return layout->GetBlockSize(storage::DataLayout::CH_CORE_MARKER) > 4;
+        return layout->GetBlockSize(storage::DataLayout::CH_CORE_MARKER) > 16;
     }
     else
     {
@@ -194,8 +195,8 @@ bool Engine<routing_algorithms::corech::Algorithm>::CheckCompability(const Engin
 
         in.seekg(0, std::ios::end);
         std::size_t size = in.tellg();
-        // An empty core files is only the 4 byte size header.
-        return size > sizeof(std::uint32_t);
+        // An empty core files is only the 8 byte size header plus the 8 byte Fingerprint.
+        return size > sizeof(std::uint64_t) + sizeof(util::FingerPrint);
     }
 }
 }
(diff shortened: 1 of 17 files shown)


src/storage/storage.cpp
```diff
@@ -331,7 +331,7 @@ void Storage::PopulateLayout(DataLayout &layout)
 
     // load geometries sizes
     {
-        io::FileReader reader(config.geometries_path, io::FileReader::HasNoFingerprint);
+        io::FileReader reader(config.geometries_path, io::FileReader::VerifyFingerprint);
 
         const auto number_of_geometries_indices = reader.ReadVectorSize<unsigned>();
```
duizendnegen: What is the logic behind 'ReadVectorSize' reading `CountElement64`, then skipping `T` (in this case `unsigned`)? Naming doesn't cover what's actually happening here.

Ideally I'd also take this through a `files` layer, any input as to how?

include/engine/engine.hpp
```diff
@@ -184,7 +185,7 @@ bool Engine<routing_algorithms::corech::Algorithm>::CheckCompability(const Engin
 
         auto mem = storage::makeSharedMemory(barrier.data().region);
         auto layout = reinterpret_cast<storage::DataLayout *>(mem->Ptr());
-        return layout->GetBlockSize(storage::DataLayout::CH_CORE_MARKER) > 4;
+        return layout->GetBlockSize(storage::DataLayout::CH_CORE_MARKER) > 16;
```
duizendnegen: This (and underlying `sizeof(std::uint64_t) + sizeof(util::FingerPrint);`) can really bite... Any suggestions for a more understandable naming, to get rid of these magic numbers?

The least we should do is add some comments to both sections.

duizendnegen:
Would like to unify file names between `storage_config` and `extract_config` - any objections?

danpat:
@duizendnegen no objections - we have way too much repetition

src/updater/updater.cpp
```diff
@@ -120,10 +120,16 @@ auto mmapFile(const std::string &filename, boost::interprocess::mode_t mode)
     using boost::interprocess::file_mapping;
     using boost::interprocess::mapped_region;
 
+    {
+        storage::io::FileReader file(filename, storage::io::FileReader::VerifyFingerprint);
```
duizendnegen: Opening this file with the `FileReader` just for verifying the fingerprint seems a bit clumsy. On Windows this fails in a way since a subsequent call from `file_mapping` results in `Access denied`.

Is there a way to read just the Fingerprint from the `mapped_region`? Should we implement an explicit `Close` function for the `FileReader`/`FileWriter`, or do an implicit `fclose` on the destructor?

duizendnegen:
Ready for review / feedback. I'm tackling config normalization in a separate PR.

src/contractor/contractor.cpp
```diff
@@ -103,22 +103,14 @@ int Contractor::Run()
 
 void Contractor::ReadNodeLevels(std::vector<float> &node_levels) const
 {
-    storage::io::FileReader order_file(config.level_output_path,
-                                       storage::io::FileReader::HasNoFingerprint);
-
-    const auto level_size = order_file.ReadElementCount32();
-    node_levels.resize(level_size);
-    order_file.ReadInto(node_levels);
+    files::readLevels(config.level_output_path, node_levels);
```
TheMarex: Thanks for the refactor! I think we can remove `Contractor::ReadNodeLevels` entirely now and directly call `files::readLevels`.

src/contractor/contractor.cpp
```diff
@@ -103,22 +103,14 @@ int Contractor::Run()
...
 
 void Contractor::WriteNodeLevels(std::vector<float> &&in_node_levels) const
 {
     std::vector<float> node_levels(std::move(in_node_levels));
 
-    storage::io::FileWriter writer(config.level_output_path,
-                                   storage::io::FileWriter::HasNoFingerprint);
-
-    storage::serialization::write(writer, node_levels);
+    files::writeLevels(config.level_output_path, node_levels);
```
TheMarex: Same as above.

src/storage/storage.cpp
TheMarex: We need to extract the file size in order to pre-allocate the shared memory block. For this we need to extract the size of every block/vector. 

I couldn't come up with a good abstraction for this yet. But yeah having a dedicated function in `files::` would be good.

src/updater/updater.cpp
```diff
@@ -120,10 +120,16 @@ auto mmapFile(const std::string &filename, boost::interprocess::mode_t mode)
...
+    }
+
     try
     {
         const file_mapping mapping{filename.c_str(), mode};
-        mapped_region region{mapping, mode};
+
+        // map region started at an offset of util::FingerPrint size
+        mapped_region region{mapping, mode, 8};
```
TheMarex: instead of `8` what about using `sizeof(Fingerprint)`?

TheMarex:
Windows tests seem to fail during updating. I suspect the `8` bytes for the fingerprint are not working cross platform?

duizendnegen:
@TheMarex it's not about the 8 bytes part - see https://github.com/Project-OSRM/osrm-backend/pull/3904#discussion_r110520810 - it's opening and not explicitly closing the file, then re-opening (somehow). To get things to work I can leave out fingerprint checking .turn_penalties_index, or figure out a different way to read the fingerprint, or close the file explicitly somewhere. Open for suggestions.

src/updater/updater.cpp
TheMarex: > or do an implicit fclose on the destructor

Wait we don't close file handles? This is definitely a bug that should be fixed.

/cc @danpat I think these might be the reason we saw the file handle exhaustion.

TheMarex: Okay just looked at the [implementation](https://github.com/Project-OSRM/osrm-backend/blob/master/include/storage/io.hpp#L28): We just wrap `ifstream` which closes on destruction, which still begs the question of why this files here.

danpat: Yes, it should be close-on-destroy.  The errors on appveyor seem to be around `mmap` calls though - perhaps we're not unmapping at some point?  Multiple mappings should work though.....

duizendnegen: I was wrong anyway, commented out the file reader fingerprint verification and errors still occur. I'll skip fingerprinting these mmap'ed files for now (just one file) to get this to land and open a ticket.

duizendnegen:
Skipped the problematic `.turn_penalties_index` for another PR. Ready to merge from my side.

PR #3909, opened 2017-04-06
Adapts bindings to node-cmake v2, resolves #3879

### LINKED ISSUE #3879: Update NodeJs buildsystem integration node-cmake to v2
There has been a major change in node-cmake and after bundling node-osrm we're still on v1:

https://github.com/Project-OSRM/osrm-backend/blob/7a1a209168b5865c7e5f989028a4716dc58e96ea/package.json#L17-L19

Here's the pull request over in node-osrm upgrading to v2 - from December:

https://github.com/Project-OSRM/node-osrm/pull/280

I don't know why we never merged it, but we should upgrade this now.

daniel-j-h:
For #3879. I worked around upstream issues noted in #3879.

Two questions here:
- I vendored the `NodeJS.cmake` module again. We could call `../node_modules/.bin/ncmake update` in the `package.json` scripts field but at the moment we recommend users `npm install`-ing via `--ignore-scripts`. Therefore no scripts will be executed. Once my patch for the Debian and Ubuntu issue is merged upstream, should we simply append `node_modules/node-cmake/` to the CMake search path instead of vendoring?
- Why is nan, node-pre-gyp and node-cmake in the `dependencies` section in our `package.json`? Aren't those dev dependencies only? It's unclear to me how we differentiate between the two at the moment.


daniel-j-h:
In addition the CMake target is now called `node_osrm` instead of `node-osrm` (because `-` is not allowed ..). As long as the `name` field in the `package.json` is `node-osrm` we should be fine here, though.

TheMarex:
@daniel-j-h a test one can use to decide whether a dependency belongs in `devDependencies` or `dependencies` is doing:

`npm install --only=prod` (only installs `dependencies`)

If this works and produces a functional module all dependencies are captured correctly. I think the split is there so the dependency tree can be trimmed a little (you won't get dev packages from packages that you depend on).

Now to the issue at hand: `node-pre-gyp` is needed to download the binaries (if existent) and `node-cmake` to fallback to source build. `nan` is needed for the source build.

Now the problem is that we don't implement the fallback to source build. `mapbox-gl-native` does something like `node-pre-gyp install --fallback-to-build-from-source=false || make node`.

I'm not sure about `ncmake` syntax but maybe something like `|| ncmake build -DENABLE_MASON=On -DENABLE_NODE_BINDINGS=On` would work for us?

daniel-j-h:
Ah thanks for the explanations re. fallback. I tried to implement the `|| ncmake` fallback, but at the moment there is no way to pass options to cmake through ncmake (=> https://github.com/cjntaylor/node-cmake/issues/25).

Issue #3854, opened 2017-03-23
Instructions to run Cucumber test locally doesn't work for me


emiltin:
Trying to follow the instructions at https://github.com/Project-OSRM/osrm-backend/wiki/Cucumber-Test-Suite#running-tests-locally on Ubuntu 16.04 fails for me.

I've installed nvm and node as described:

```shell
emil@emil-OptiPlex-7010:~/code/osrm-backend$ nvm run node --version
Running node v7.7.4 (npm v4.1.2)
v7.7.4
```
But running npm install fails:
```
emil@emil-OptiPlex-7010:~/code/osrm-backend$ npm install

> osrm@5.7.0 preinstall /home/emil/code/osrm-backend
> npm install node-pre-gyp

osrm@5.7.0 /home/emil/code/osrm-backend
└── node-pre-gyp@0.6.34 

npm WARN optional SKIPPING OPTIONAL DEPENDENCY: fsevents@^1.0.0 (node_modules/chokidar/node_modules/fsevents):
npm WARN notsup SKIPPING OPTIONAL DEPENDENCY: Unsupported platform for fsevents@1.1.1: wanted {"os":"darwin","arch":"any"} (current: {"os":"linux","arch":"x64"})

> osrm@5.7.0 install /home/emil/code/osrm-backend
> node-pre-gyp install --fallback-to-build=false

node-pre-gyp ERR! install error 
node-pre-gyp ERR! stack Error: 403 status code downloading tarball https://mapbox-node-binary.s3.amazonaws.com/osrm/v5.7.0/Release/node-v51-linux-x64.tar.gz
node-pre-gyp ERR! stack     at Request.<anonymous> (/home/emil/code/osrm-backend/node_modules/node-pre-gyp/lib/install.js:120:27)
node-pre-gyp ERR! stack     at emitOne (events.js:101:20)
node-pre-gyp ERR! stack     at Request.emit (events.js:191:7)
node-pre-gyp ERR! stack     at Request.onRequestResponse (/home/emil/code/osrm-backend/node_modules/request/request.js:1074:10)
node-pre-gyp ERR! stack     at emitOne (events.js:96:13)
node-pre-gyp ERR! stack     at ClientRequest.emit (events.js:191:7)
node-pre-gyp ERR! stack     at HTTPParser.parserOnIncomingClient [as onIncoming] (_http_client.js:522:21)
[...]

PR #3914, opened 2017-04-07
Add script to fall back to source build

TheMarex:
# Issue

This issue addresses https://github.com/Project-OSRM/osrm-backend/issues/3854 by fixing the `install` target.

## Tasklist
 - [x] review
 - [x] adjust for comments

### DIFF
--- a/scripts/node_install.sh
+++ b/scripts/node_install.sh
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/bin/bash
 
 set -e
 
--- a/scripts/node_install.sh
+++ b/scripts/node_install.sh
@@ -1,0 +1,28 @@
+#!/bin/sh
+
+set -e
+
+CMAKE_OPTIONS=${CMAKE_OPTIONS:-"-DCMAKE_BUILD_TYPE=Release -DENABLE_NODE_BINDINGS=On -DENABLE_MASON=On"}
+
+if [[ ! -f $(which cmake) ]]; then
+    echo "Needs cmake to build from source"
+    exit 1
+fi
+if [[ ! -f $(which clang) ]]; then
+    echo "Needs clang to build from source"
+    exit 1
+fi
+if [[ ! -f $(which clang++) ]]; then
+    echo "Needs clang++ to build from source"
+    exit 1
+fi
+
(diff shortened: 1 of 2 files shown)


daniel-j-h:
Where is this `node_install.sh` you are referencing here?
https://github.com/Project-OSRM/osrm-backend/tree/nodejs/fix-install/scripts

scripts/node_install.sh
```diff
@@ -0,0 +1,28 @@
+#!/bin/sh
+
+set -e
+
+CMAKE_OPTIONS=${CMAKE_OPTIONS:-"-DCMAKE_BUILD_TYPE=Release -DENABLE_NODE_BINDINGS=On -DENABLE_MASON=On"}
```
daniel-j-h: Why do we force a mason build? Because it's the only way to be sure users have all the deps.?

scripts/node_install.sh
```diff
@@ -0,0 +1,28 @@
...
+    exit 1
+fi
+if [[ ! -f $(which clang) ]]; then
+    echo "Needs clang to build from source"
+    exit 1
+fi
+if [[ ! -f $(which clang++) ]]; then
+    echo "Needs clang++ to build from source"
+    exit 1
+fi
```
daniel-j-h: If we use mason anyway, can't we use clang 4 from mason here, too?

scripts/node_install.sh
```diff
@@ -0,0 +1,28 @@
...
+
+if [[ -d build ]]; then
+    echo "Detected existing build directory, skipping compiling."
+    exit 0
+else
+    mkdir -p build
+    pushd build
+    CXX=clang++ CC=clang cmake .. $CMAKE_OPTIONS
```
daniel-j-h: The `cmake --build .` step is missing on purpose?

scripts/node_install.sh
TheMarex: I mainly did this to be compatible with `node-osrm`. I don't feel strongly about this, vanilla build would be fine for me as well.

TheMarex:
@daniel-j-h should have addressed all your comments. Let me know if this works for you.

Issue #3878, opened 2017-03-29
Expand build matrix for more node versions


TheMarex:
Currently we only build the node binaries for the `Release` target and node `4`. We should expand the build matrix to capture the same configurations as `node-osrm`:

- Linux Node 4 `Release`
- Linux Node 6 `Release`
- Linux Node 4 `Debug`
- Linux Node 6 `Debug`
- OSX Node 4 `Release`
- OSX Node 6 `Release`
- OSX Node 4 `Debug`
- OSX Node 6 `Debug`

This is a lot of build jobs to add to our already huge matrix, so we should find a way to only enable them conditionally. Options:

- Add each configuration and `exit 0` for non-releases
- Try to get Travis to support `branches` in `matrix.include`: https://github.com/travis-ci/travis-ci/issues/2778
- Use Travis API to run jobs via CLI only when we need to.

daniel-j-h:
Targeting 5.7 here - we need proper releases at least for the LTS node versions.

Especially given that Node v4 goes from active LTS into maintenance mode starting April.

https://github.com/nodejs/LTS#lts-schedule

daniel-j-h:
This landed with https://github.com/Project-OSRM/osrm-backend/pull/3915.

PR #3915, opened 2017-04-07
Enable full node build matrix using sudo:false

TheMarex:
# Issue

This PR addresses https://github.com/Project-OSRM/osrm-backend/issues/3878 by adding a build matrix using `sudo:false` jobs that skip running the cucumber tests.

## Tasklist
 - [x] review
 - [x] adjust for comments

## Requirements / Relations
Based on https://github.com/Project-OSRM/osrm-backend/pull/3914


.travis.yml
```diff
@@ -132,20 +132,72 @@ matrix:
...
   - if [[ ! -z $TARGET_ARCH ]] ; then source ./scripts/travis/before_install.$TARGET_ARCH.sh ; fi
   - |
-    if [[ "${TRAVIS_OS_NAME}" == "linux" ]]; then
-      export JOBS=$((`nproc` + 1))
+    if [[ -z $JOBS ]]; then
+      if [[ "${TRAVIS_OS_NAME}" == "linux" ]]; then
+        export JOBS=$((`nproc` + 1))
+      elif [[ "${TRAVIS_OS_NAME}" == "osx" ]]; then
+        export JOBS=$((`sysctl -n hw.ncpu` + 1))
+      fi
```
daniel-j-h: Can we hardcode jobs to 2 or 3, otherwise this will bite us on containerized envs.

https://docs.travis-ci.com/user/ci-environment/

TheMarex: Yes there is already an override for container builds. Note the `-z $JOBS`

TheMarex:
@daniel-j-h I would like to avoid looping over build configurations somewhere.

Using containers instead of `sudo:true` instances hopefully causes no congestion in practice, as the build time on `node-osrm` was usually below 4 minutes per job (compared to 20 minutes for our longest job right now). 

I'm not sure we should expand this to OSX builds though. Capacity there is notoriously limited and only constraining this to node 4, node 6 in release should be fine IMHO.

scripts/timer.js
```diff
@@ -0,0 +1,16 @@
+#!/usr/bin/env node
+
+var exec = require('child_process').exec;
+var fs = require('fs');
+
+var name = process.argv[2];
+var cmd = process.argv.slice(3).join(' ');
+var start = +new Date();
```
danpat: I would prefer `Date.now()` for readability instead of depending on type conversion using `+`.  Don't make me think so hard :-P

danpat:
I'm 👍 on not building debug binaries for OSX - for debugging, I (and most OSX users I know) will build locally if we need debug.  Linux debug builds are useful for replicating production problems, but I agree, we can avoid the OSX queue on Travis here.

It's simple enough to turn on later if we need it.

PR #3919, opened 2017-04-07
make turn penalty if weight is not duration

MichalPP:
make turn penalty if weight is not duration

# Issue

in the bike profile, the turn penalty does not occur when weight is other than duration

## Tasklist
 - [ ] review

### DIFF
--- a/profiles/bicycle.lua
+++ b/profiles/bicycle.lua
@@ -544,5 +544,10 @@ function turn_function(turn)
       if not turn.source_restricted and turn.target_restricted then
           turn.weight = turn.weight + 3000
       end
+    if normalized_angle >= 0.0 then
+      turn.weight = turn.weight + normalized_angle * normalized_angle * profile.turn_penalty / profile.turn_bias
+    else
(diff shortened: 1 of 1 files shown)


MichalPP:
@emiltin , what do you think?

MichalPP:
@emiltin, @danpat what do you think? it's just a post #77 profile bug

profiles/bicycle.lua
```diff
@@ -544,5 +544,10 @@ function turn_function(turn)
       if not turn.source_restricted and turn.target_restricted then
           turn.weight = turn.weight + 3000
       end
+    if normalized_angle >= 0.0 then
```
TheMarex: What about just setting `turn.weight = turn.duration` on line 543 and removing this code? I think this was a c&p bug.

MichalPP: ok, that is a lot more simpler.

TheMarex:
Thanks for taking a stab at this! 👍  Replaced by #4044.

Issue #2865, opened 2016-09-06
OSRM tools should have a way to list the required files


danpat:
When deploying OSRM to a scalable infrastructure, it's necessary to move copies of all the generated files around.  The list of files changes frequently as new features are added to OSRM.

However, for optimal performance, not everything needs to be moved - `osrm-routed/osrm-datastore`, for example, don't need the `.ebg` file.

In addition, some files are optional, for example the `.edge_penalties` and `.edge_segment_lookup` files.

Each tool should have a `--list-inputs` option that lists all the file extensions it expects, and whether they're optional or not.  This can then be trivially parsed by wrapper scripts to ensure that the correct file list.  Example:

```
$ ./osrm-datastore --list-inputs
required .nodes
required .icd
required .fileIndex
required .names
required .ramIndex
optional .core
```

Each line would be prefixed with `required` or `optional` - if required files are not available, the wrapper process should error.  If `optional` files are missing, the wrapper process can continue without error, but note that some features may not be available (such as core search).

duizendnegen:
I'll be tackling this in #3920 as well.

MathiasVersichele:
I see this is still open. Can anyone direct me to which files are needed for the map-matching service ?

PR #3920, opened 2017-04-08
Normalize config files

duizendnegen:
# Issue

Normalizing the usage & file names across config files by referring to a single config file.

## Tasklist
 - [x] contractor_config
 - [x] updater_config
 - [x] customizer_config
 - [x] partition_config
 - [x] extractor_config
 - [x] review
 - [x] adjust for comments
 - [x] engine_config
 - [x] transform to use `GetPath`


duizendnegen:
Currently pointed config files to use storage_config. This is slightly confusing as `IsValid` doesn't make sense now in all cases.

Alternatively we can create an `io_config`, either as a reference or as a parent class. Welcoming suggestions.

TheMarex:
What is the motivation for using `StorageConfig` in all modules? It is meant as the configuration for the `storage` component, what is the relation with the other modules?

duizendnegen:
No motivation - I misunderstood it at the start and then just got it to a working situation. What's your opinion of using `io_config` as a parent class (option A) or as a reference / variable (option B)?

duizendnegen:
Ready for review!

include/storage/io_config.hpp
```diff
@@ -0,0 +1,82 @@
+#ifndef IO_CONFIG_HPP
```
daniel-j-h: Should probably be prefix with `OSRM_`

include/storage/io_config.hpp
```diff
@@ -0,0 +1,82 @@
...
+
+#include <boost/filesystem/path.hpp>
+
+namespace osrm
+{
+namespace storage
+{
+struct IOConfig
+{
+    IOConfig() = default;
```
daniel-j-h: Why is the default constructor needed? A default constructed `IOConfig` can't do anything, no?

include/storage/io_config.hpp
```diff
@@ -0,0 +1,82 @@
...
+        mld_partition_path = {osrm_path.string() + ".partition"};
+        mld_storage_path = {osrm_path.string() + ".cells"};
+        mld_graph_path = {osrm_path.string() + ".mldgr"};
+        level_path = {osrm_path.string() + ".level"};
+        node_path = {osrm_path.string() + ".enw"};
+        edge_based_graph_path = {osrm_path.string() + ".ebg"};
+        compressed_node_based_graph_path = {osrm_path.string() + ".cnbg"};
+        cnbg_ebg_mapping_path = {osrm_path.string() + ".cnbg_to_ebg"};
+        restriction_path = {osrm_path.string() + ".restrictions"};
+        intersection_class_data_path = {osrm_path.string() + ".icd"};
```
daniel-j-h: Hm is there a way we can make it so that we get a warning if we forget to initialize a path below?

Should this e.g. all be done in the constructor, as in
```c++
struct io {
  io(path base)
    : a(base + '.a'),
      b(base + '.b') { }

  path a;
  path b;
};
```
?

daniel-j-h:
Also needs a rebase onto master before we can merge 🎄

include/storage/io_config.hpp
duizendnegen: Yes, I would prefer to drop this as a constructor, but in practice it doesn't work: https://github.com/Project-OSRM/osrm-backend/blob/bf6698f4cc7608a75d9ac72f18248356b6ed162c/src/tools/extract.cpp#L27 is how command line arguments are parsed, they require an instantiated version of the Config files. I started without that and without the `UseDefaultOutputNames` function, but I couldn't get it to work like that.

include/storage/io_config.hpp
duizendnegen: See above - yes, should be the case, but didn't work out in practice.

daniel-j-h: Ok I see. Thanks for explaining! :)

duizendnegen:
Rebased, addressed comments :+1:

daniel-j-h:
You might be thinking of what's ticketed here https://github.com/Project-OSRM/osrm-backend/issues/2865

TheMarex:
Yeah that would be one of the applications of this, but I'm more concerned about the code readability.

duizendnegen:
Hmm... I see what you mean. How would you then get the `.osrm.edges` filename from `config` to the actual code that uses it? You can't do `config.edges_path` anymore then, right? Or what would the `IOConfig` constructor do?

I just wanted to normalize the naming a bit :) it seems a bit silly to spend a lot more time on this, I thought normalized file names would help figuring out the flow of file names in the code - but I would rather get this done & focus on the cross-platform part again.

TheMarex:
@duizendnegen maybe something like `boost::filesystem::path IOConfig::GetPath(std::string name)`:

```
// Throws an error if IOConfig was not constructed with `.osrm.edges` as required or optional file, or if
// the path does not exist.
 auto path = config.GetPath(".osrm.edges");
```

duizendnegen:
I like that @TheMarex - @daniel-j-h how do you feel about this? Just don't want to spend time refactoring the entire config and finding out that we end up not being aligned :)

duizendnegen:
Ping @daniel-j-h

daniel-j-h:
👍 then you need to hide all the impl. specific member attributes and expose only those functions publicly.

duizendnegen:
Note to self: `.nodes` and `.icd` are actually unused.

duizendnegen:
@TheMarex @daniel-j-h implemented as proposed, ready for review again.
It needs eyes on the `_config.hpp` files to validate that the right files are configured as input/optional/output.

daniel-j-h:
Some Traivs builds seem to fail because of clang-format, some because you throw an exception with a string literal.

duizendnegen:
Yep - I'm on it, the exception gets thrown because some I/O files are not properly configured too, I think.

duizendnegen:
Tests (mostly) pass, `Weighting based on speed file weights, ETA based on file durations` fails in two builds (but passes in others & works locally). Any help pushing this one?

daniel-j-h:
I just re-triggered the failing tests on Traivs - at least two of them were blips with the network connection dropping. The failing test you were seeing is probably related to https://github.com/Project-OSRM/osrm-backend/issues/4110.

duizendnegen:
No tests fail now 🥇 just code cov - OK to merge?

duizendnegen:
Ping @daniel-j-h - any review feedback?

duizendnegen:
Rebased, would be cool if we could merge before more conflicts arise 🙈

daniel-j-h:
@duizendnegen 👋 on my todo list, will hopefully come to it tomorrow morning.

Sorry I was out on vacation.

duizendnegen:
@daniel-j-h :+1: thanks! Rebased again.

duizendnegen:
don't know what happened with the 1 failing build in Travis; was passing before I rebased, error is not too informative

CMakeLists.txt
```diff
@@ -718,11 +718,11 @@ file(GLOB VariantGlob third_party/variant/include/mapbox/*.hpp)
 file(GLOB LibraryGlob include/osrm/*.hpp)
 file(GLOB ParametersGlob include/engine/api/*_parameters.hpp)
 set(EngineHeader include/engine/status.hpp include/engine/engine_config.hpp include/engine/hint.hpp include/engine/bearing.hpp include/engine/approach.hpp include/engine/phantom_node.hpp)
-set(UtilHeader include/util/coordinate.hpp include/util/json_container.hpp include/util/typedefs.hpp include/util/alias.hpp include/util/exception.hpp)
-set(ExtractorHeader include/extractor/extractor.hpp include/extractor/extractor_config.hpp include/extractor/travel_mode.hpp)
+set(UtilHeader include/util/coordinate.hpp include/util/json_container.hpp include/util/typedefs.hpp include/util/alias.hpp include/util/exception.hpp include/util/log.hpp)
```
daniel-j-h: Hm this makes `util/log.hpp` part of our public api - is this really needed? And if so can we prefix the include guards with `OSRM_` here: https://github.com/Project-OSRM/osrm-backend/blob/master/include/util/log.hpp#L1-L2

example/CMakeLists.txt
```diff
@@ -31,6 +31,6 @@ add_executable(osrm-example example.cpp)
 
 find_package(LibOSRM REQUIRED)
 
-target_link_libraries(osrm-example ${LibOSRM_LIBRARIES} ${LibOSRM_DEPENDENT_LIBRARIES})
+target_link_libraries(osrm-example ${LibOSRM_LIBRARIES} ${LibOSRM_DEPENDENT_LIBRARIES} ${UtilGlob})
```
daniel-j-h: Why does the example now need the util lib? Shouldn't all symbols needed for linking against osrm be included in the `libosrm` library already?

include/contractor/contractor_config.hpp
```diff
@@ -39,30 +40,28 @@ namespace osrm
 namespace contractor
 {
 
-struct ContractorConfig
+struct ContractorConfig final : storage::IOConfig
 {
-    ContractorConfig() : requested_num_threads(0) {}
+    ContractorConfig()
+        : requested_num_threads(0), IOConfig(
```
daniel-j-h: there are some compiler warnings here (and at similar places below) because we first initialize the num. of threads and then the parent object - you might want to change the order here

include/storage/io_config.hpp
```diff
@@ -0,0 +1,100 @@
...
+
+namespace osrm
+{
+namespace storage
+{
+struct IOConfig
+{
+    IOConfig(std::vector<boost::filesystem::path> _required_input_files,
+             std::vector<boost::filesystem::path> _optional_input_files,
+             std::vector<boost::filesystem::path> _output_files)
```
daniel-j-h: Hmm I _think_ `_lowercase` was reserved, that's why the common idiom is to make it `lowercase_` - not entirely sure though

include/storage/io_config.hpp
```diff
@@ -0,0 +1,100 @@
...
+        {
+            const auto pos = path.find(ext);
+            if (pos != std::string::npos)
+            {
+                path.replace(pos, ext.size(), "");
+                break;
+            }
+        }
+
+        base_path = {path};
```
daniel-j-h: I think you can replace the last 15 lines with

    trim_right_if(path, is_any_of("all your extensions"));

http://www.boost.org/doc/libs/1_63_0/doc/html/string_algo/usage.html#idp113215984

include/storage/io_config.hpp
duizendnegen: OK, will change

include/contractor/contractor_config.hpp
duizendnegen: OK, will change

example/CMakeLists.txt
duizendnegen: Will check, don't remember exactly. I recall this: `osrm-example` uses `io_config` which uses `util/log`.

CMakeLists.txt
duizendnegen: Because we expose `storage_config` which is `io_config` which uses `util/log`. Welcoming feedback or improvements.

include/storage/io_config.hpp
duizendnegen: Thanks, will investigate!

duizendnegen: `is_any_of` doesn't work in this case, only for single chars.

duizendnegen:
Addressed comments, still need to follow ups on
- [ ] file name changes
- [ ] `is_any_of` feedback
- [ ] exposing `log.hpp`

duizendnegen:
Ping! Looking for some feedback on the open comments above.

daniel-j-h:
The failing Travis job timed out; I just restarted it.

include/util/log.hpp
```diff
@@ -1,5 +1,5 @@
-#ifndef LOG_HPP
-#define LOG_HPP
+#ifndef OSRM_LOG_HPP
+#define OSRM_LOG_HPP
```
daniel-j-h: thanks for prefixing!

daniel-j-h:
Maybe you can hide the `log.hpp` file from the public by splitting the config into header (not using log here) and implementation (using log)? The `is_any_of` is fine. I will ping @oxidase for the file changes, I think we added a new file and renamed one.

oxidase:
@daniel-j-h in v5.8.0 `.nodes` and `.ebg` files were restructured and renamed to `.nbg_nodes` and `.ebg_nodes`

duizendnegen:
Rebased, moved log usage in `io_config` to implementation, went over the file restructuring & renaming.
From the changelog: `.osrm.nodes file was renamed to .nbg_nodes and .ebg_nodes was added` - this is covered.

duizendnegen:
From IRC: ```Hey _1009 there's a small backwards incompatible change in your current IOConfig PR - https://github.com/Project-OSRM/osrm-backend/pull/3920/files#diff-9b0fcadd9f362547a7dd4f631c4ec3fcR33 breaks the API. What you can do is make it protected and call the function in all sub-class constructors.```

duizendnegen:
Ping @daniel-j-h

example/example.cpp
```diff
@@ -30,7 +30,7 @@ int main(int argc, const char *argv[])
 
     // Configure based on a .osrm base path, and no datasets in shared mem from osrm-datastore
     EngineConfig config;
-    config.storage_config = {argv[1]};
+    config.storage_config.UseDefaultOutputNames(argv[1]);
```
daniel-j-h: Hm user coder here needs to change this is still breaking the api

include/nodejs/node_osrm_support.hpp
```diff
@@ -97,8 +97,9 @@ inline engine_config_ptr argumentsToEngineConfig(const Nan::FunctionCallbackInfo
 
     if (args[0]->IsString())
     {
-        engine_config->storage_config = osrm::StorageConfig(
-            *v8::String::Utf8Value(Nan::To<v8::String>(args[0]).ToLocalChecked()));
+        std::string base_path(*v8::String::Utf8Value(args[0]->ToString()));
+        engine_config->storage_config = osrm::StorageConfig();
+        engine_config->storage_config.UseDefaultOutputNames(base_path);
```
daniel-j-h: Same here

include/nodejs/node_osrm_support.hpp
```diff
@@ -121,8 +122,9 @@ inline engine_config_ptr argumentsToEngineConfig(const Nan::FunctionCallbackInfo
 
     if (!path->IsUndefined())
     {
-        engine_config->storage_config =
-            osrm::StorageConfig(*v8::String::Utf8Value(Nan::To<v8::String>(path).ToLocalChecked()));
+        std::string base_path(*v8::String::Utf8Value(path->ToString()));
+        engine_config->storage_config = osrm::StorageConfig();
+        engine_config->storage_config.UseDefaultOutputNames(base_path);
```
daniel-j-h: And here

src/benchmarks/match.cpp
```diff
@@ -30,7 +30,7 @@ int main(int argc, const char *argv[]) try
 
     // Configure based on a .osrm base path, and no datasets in shared mem from osrm-datastore
     EngineConfig config;
-    config.storage_config = {argv[1]};
+    config.storage_config.UseDefaultOutputNames(argv[1]);
```
daniel-j-h: And here

daniel-j-h:
Ah I've just seen you added a constructor calling `UseDefaultOutputNames`. Can you switch all the public code to using that and make the `UseDefaultOutputNames` protected so only sub-classes can call it.

duizendnegen:
But in previous versions most subclasses had `UseDefaultOutputNames` type of constructs, right...? This is why the initialisation from command line worked. Are you proposing making it protected (for `storage_config`) and public in `customizer_config`, `extractor_config`, ...?

daniel-j-h:
Yeah the user should not see the `UseDefaultOutputNames` function on the publicly exposed configs. And especially should not have to adapt her code to call this function.

duizendnegen:
Addressed comments (waiting for builds to pass, might have to do some clang-formatting, feel free to leave additional comments)

daniel-j-h:
Yep Travis complains about clang-formating `include/storage/io_config.hpp` - otherwise looks good!

duizendnegen:
:+1: formatted the file, now waiting for 📗 🍏

duizendnegen:
One test seemed to have timed out, could you restart?

daniel-j-h:
Thanks for this - it's making some I/O parts nicer for sure! I just merged it into master.

✨

PR #3923, opened 2017-04-09
Buffer turn data on disk to save memory

TheMarex:
# Issue

Solve #3992 by not caching data in memory.

## Tasklist
 - [x] Measure slow down: 5% (on Bayern)
 - [x] review
 - [x] adjust for comments

## Requirements / Relations

### DIFF
--- a/include/storage/serialization.hpp
+++ b/include/storage/serialization.hpp
@@ -1,8 +1,8 @@
 #ifndef OSRM_STORAGE_SERIALIZATION_HPP
 #define OSRM_STORAGE_SERIALIZATION_HPP
 
-#include "util/vector_view.hpp"
 #include "util/integer_range.hpp"
+#include "util/vector_view.hpp"
 
 #include "storage/io.hpp"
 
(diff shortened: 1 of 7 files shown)


include/util/vector_view.hpp
```diff
@@ -175,9 +177,15 @@ template <typename DataT> void swap(vector_view<DataT> &lhs, vector_view<DataT>
 }
 
 template <typename DataT, storage::Ownership Ownership>
+using InternalOrExternalVector =
+    typename std::conditional<Ownership == storage::Ownership::External,
+                              stxxl::vector<DataT>,
```
oxidase: `stxxl::vector` will not allow to use parallelization.

TheMarex: Yes. We need to reinvestigate this when we try to parallelize that loop.

include/storage/serialization.hpp
```diff
@@ -1,16 +1,39 @@
...
+    }
+}
+
+template <typename T>
+inline void write(storage::io::FileWriter &writer, const stxxl::vector<T> &vec)
+{
+    writer.WriteOne(vec.size());
+    for (auto idx : util::irange<std::size_t>(0, vec.size()))
+    {
+        writer.WriteOne<T>(vec[idx]);
```
daniel-j-h: I don't understand why we don't provide an I/O overload for everything providing begin/end or size/data.

danpat: We do - `serializeVector()`

PR #3925, opened 2017-04-10
Implement distance pruning for MLD as well

TheMarex:
# Issue

Implements distance based pruning like for CH. This potentially fixes the latency increase that I encountered for map matching.

## Tasklist
 - [x] review
 - [x] adjust for comments

### DIFF
--- a/src/engine/routing_algorithms/map_matching.cpp
+++ b/src/engine/routing_algorithms/map_matching.cpp
@@ -198,9 +198,9 @@ SubMatchingList mapMatching(SearchEngineData<Algorithm> &engine_working_data,
 
             const auto haversine_distance = util::coordinate_calculation::haversineDistance(
                 prev_coordinate, current_coordinate);
-            // assumes minumum of 0.1 m/s
-            const int duration_upper_bound =
-                ((haversine_distance + max_distance_delta) * 0.25) * 10;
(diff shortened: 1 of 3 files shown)


src/engine/routing_algorithms/map_matching.cpp
```diff
@@ -198,9 +198,9 @@ SubMatchingList mapMatching(SearchEngineData<Algorithm> &engine_working_data,
 
             const auto haversine_distance = util::coordinate_calculation::haversineDistance(
                 prev_coordinate, current_coordinate);
-            // assumes minumum of 0.1 m/s
-            const int duration_upper_bound =
-                ((haversine_distance + max_distance_delta) * 0.25) * 10;
+            // assumes minumum of 4 m/s
+            const EdgeWeight weight_upper_bound =
+                ((haversine_distance + max_distance_delta) / 4.) * 10;
```
oxidase: But how the upper bound will work for non-duration weights? Here must be a "duration"-like value
If weight is ok then  here instead of 10 must be `weight_multiplier`.

TheMarex: Good catch with the multiplier. 👍 

For non-duration based weights this will not work, but in that case I would expect other parts of the map matching to already break (e.g. the parameters for the probability distributions).

Issue #3823, opened 2017-03-14
Segment speed file has a `weight` column instead of a `rate` column


TheMarex:
I noticed this while working on the update code: The current format we expect for speed files is currently:

```
from,to,speed,weight
```

However what we really want is `rate` (e.g. a speed-like) value instead of `weight` which is an absolute value like `duration`. In my mind this is an API bug: Somehow I did not catch this in the review process.

My proposal would be to change this to:

```
from,to,speed,rate
```

and backport a fix to 5.6.

@danpat @oxidase opinions?

oxidase:
:+1: i would suggest also to make it a required field, related to #3805

danpat:
👍 Agreed, this should be `rate`, using `weight` makes things awfully fragile when matching generated data against the map, `rate` is much more forgiving.

TheMarex:
We forgot about this in the `5.6` series, we should fix it now.

TheMarex:
This was fixed by #3926

PR #3926, opened 2017-04-10
Change traffic CSV field value from weight to rate

oxidase:
# Issue

PR changes for #3823 traffic CSV field value from weight to rate and makes it  a required field.

If the weight name is 'duration' than the rate value can be computed as speed / 3.6.

## Tasklist
 - [ ] review
 - [ ] adjust for comments


src/updater/csv_source.cpp
```diff
@@ -34,7 +34,7 @@ namespace csv
 SegmentLookupTable readSegmentValues(const std::vector<std::string> &paths)
 {
     CSVFilesParser<Segment, SpeedSource> parser(
-        1, qi::ulong_long >> ',' >> qi::ulong_long, qi::uint_ >> -(',' >> qi::double_));
+        1, qi::ulong_long >> ',' >> qi::ulong_long, qi::double_ >> ',' >> qi::double_);
```
TheMarex: Making the weight column non-optional would cause some serious problems downstream. What about showing a warning if someone uses a three column CSV but still supporting the fallback method?

src/updater/updater.cpp
```diff
@@ -62,18 +62,34 @@ inline EdgeWeight convertToDuration(double speed_in_kmh, double distance_in_mete
...
+inline EdgeWeight convertToWeight(const SpeedSource &value,
+                                  double distance_in_meters,
+                                  const extractor::ProfileProperties &profile_properties)
 {
-    if (std::isfinite(weight))
-        return std::round(weight * weight_multiplier);
+    double rate = value.rate;
+    if (!std::isfinite(rate))
+    {
+        if (!profile_properties.fallback_to_duration)
```
TheMarex: Instead of printing this for every line what about adding a counter, and only printing this warning one with a count like "used rate fallback to speed for X entries"? This might spam the logs, given we have CSV files with a lot of entries.

daniel-j-h:
Could you update the Traffic wiki with the new column and a description of it's usage.

Issue #3896, opened 2017-04-03
Use 20bit for segment weights and durations


TheMarex:
Currently we use 32bit for every edge segment (that means pair of OSM node ids, that would represent a street segment) to store the weight an duration. That is very generous considering these segments are usually only a few hundred meters long at the most.

In a 16bit unsigned integer we can store 65535 deci-seconds, or 1.8 hours. That should be plenty of time for such short segments. Of course we should measure this, but I fill this is a safe limit for any reasonable setting.

kostadin24:
18 hours are plenty of time if you travel with car, but what about route with walk? If we have 5kph walk speed 18 hours make 90km distance. It is not too big distance if we have contracted ways.

TheMarex:
This would only be for the very short segments that we need to compute start / target offsets from (like starting in the middle of the street instead of one of the endpoints). There is no contraction happening on those.

TheMarex:
Of course this would have been too simple. Let me introduce you to the reason why we can't have nice things: https://www.openstreetmap.org/way/441977614#map=11/47.4648/9.3998 13km of uninterrupted ferry travel.

We might be able to use a var-int or packed storage instead.

TheMarex:
Evaluated our options here for saving file sizes. Packing everything into 20bits is the least-effort solution and will yield 40% size reduction. Encoding speeds in km/h would allow us to get a 75% size reduction, but introduces some problems with units (km/h uses most of the 8bits, while m/s would only use 6bits which reduces precision).

# Use 20bits fixed size

- 20/32 =~ 40% reduction in size
- Needs fixes for PackedVector for random access
- Implementation overhead: low

# Encode as speed + distance

- don't encode duration as speed in km/h 0..255
- encoding as m/s would need a range of 180/3.6 ~= 50
- problem with weight: rate is encoded in something like m/s not km/h (does not fit in 0..255 range)
- Potentially 1 byte per value (4 bytes for forward/backward speed/rate)
- 75% reduction in size (if we don't encode the distance explicitly)
- Updates faster (no conversion speed -> duration)
- More overhead in query (PhantomNode generation needs to compute distance from coordinates, multiply with speed to compute the offset)

# Varint encoding

- We can't have a shared index for each range -> 4 bytes per GeometryID per foward/backward weight/duration for the index
- Varint encodings use at least 8bits per entry, but we would only see gains over speed-based encoding if we can go below 8 bits per entry
[...]

PR #3928, opened 2017-04-11
Make PackedVector generic on the number of bits

TheMarex:
# Issue

This makes the packed vector generic on number of bit is uses per entry. The idea is to use this for https://github.com/Project-OSRM/osrm-backend/issues/3896

## Tasklist
 - [x] review
 - [x] adjust for comments

### DIFF
--- a/include/util/packed_vector.hpp
+++ b/include/util/packed_vector.hpp
@@ -33,10 +33,12 @@ namespace detail
 {
 template <typename T, std::size_t Bits, storage::Ownership Ownership> class PackedVector
 {
+    static_assert(std::is_integral<T>, "T must be an integral type.");
     static_assert(sizeof(T) <= sizeof(std::uint64_t), "Maximum size of type T is 8 bytes");
-    static_assert(Bits <= sizeof(std::uint64_t)*CHAR_BIT, "Maximum number of bits it 64.");
+    static_assert(Bits > 0, "Minimum number of bits is 0.");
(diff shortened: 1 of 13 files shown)


include/util/packed_vector.hpp
```diff
@@ -16,33 +16,28 @@ namespace util
...
- * configure BITSIZE and ELEMSIZE
- */
-template <typename T, storage::Ownership Ownership> class PackedVector
+template <typename T, std::size_t Bits, storage::Ownership Ownership> class PackedVector
 {
-    static const constexpr std::size_t BITSIZE = 33;
-    static const constexpr std::size_t ELEMSIZE = 64;
-    static const constexpr std::size_t PACKSIZE = BITSIZE * ELEMSIZE;
+    static_assert(sizeof(T) <= sizeof(std::uint64_t), "Maximum size of type T is 8 bytes");
+    static_assert(Bits <= sizeof(std::uint64_t)*CHAR_BIT, "Maximum number of bits it 64.");
```
daniel-j-h: Hm should probably also assert for is_arithmetic (?)

daniel-j-h: Bits > 0
i*s* 64
clang format

TheMarex: Bits is of type `std::size_t` that should always be true right?

daniel-j-h: Ah sorry I meant on T

PR #3929, opened 2017-04-11
Limit SegmentWeight / SegmentDuration fewer bits

TheMarex:
# Issue

This PR explores #3896. As already described in the issue 16bit does not work because of ferries. 20bit seems fine (1747 hours per segment edge).

## Tasklist
 - [ ] ~~Use packed vector to get memory savings~~ Needs https://github.com/Project-OSRM/osrm-backend/issues/3952
 - [x] review
 - [x] adjust for comments

## Requirements / Relations

~~Requires https://github.com/Project-OSRM/osrm-backend/pull/3928~~


daniel-j-h:
Does clipping trigger on the planet? Can you add a detailed warning in there if it happens.

TheMarex:
> Does clipping trigger on the planet? Can you add a detailed warning in there if it happens.

This PR is primarily there to explore this. We should see a warning in the server logs. Unless there is a straight light ferry connection that takes 72 days (8735 km at 5km/h), I'm optimistic this will not cause problems.

include/util/typedefs.hpp
```diff
@@ -58,6 +58,9 @@ using NodeID = std::uint32_t;
 using EdgeID = std::uint32_t;
 using NameID = std::uint32_t;
 using EdgeWeight = std::int32_t;
+using EdgeDuration = std::int32_t;
+using SegmentWeight = std::int32_t;
```
oxidase: here and below can be used unsigned value. 

also unpacking should extend sign if type is signed. it is not used atm but better to add a static assert that the base type of the packed vector is unsigned

src/extractor/compressed_edge_container.cpp
```diff
@@ -134,7 +135,21 @@ void CompressedEdgeContainer::CompressEdge(const EdgeID edge_id_1,
     // weight1 is the distance to the (currently) last coordinate in the bucket
     if (edge_bucket_list1.empty())
     {
-        edge_bucket_list1.emplace_back(OnewayCompressedEdge{via_node_id, weight1, duration1});
+        if (weight1 >= INVALID_SEGMENT_WEIGHT)
```
oxidase: better `, clip(weight1), ` with `clip = [&clipped_weights] (EdgeWeight) -> SegmentWeight ` closure to reuse in both places

TheMarex:
@oxidase could you take a look again if these changes work for you?

Issue #3908, opened 2017-04-06
Reactor heaps interface


oxidase:
PR  #3895 introduces a workaround for MLD heaps by adding ["core" heaps](https://github.com/Project-OSRM/osrm-backend/blob/59350f0b16288b6ebad6f7bf4cc5315e53ddf3bb/include/engine/search_engine_data.hpp#L78-L80) that are only needed in the shortest plugin for [`Clear` calls](https://github.com/Project-OSRM/osrm-backend/blob/59350f0b16288b6ebad6f7bf4cc5315e53ddf3bb/src/engine/routing_algorithms/shortest_path.cpp#L147-L148)

The implementation of algorithm-aware heaps should not be exposed and used in plugins code with the following steps:
* heaps implementation details must be hidden in `routing_algorithms::{ch,corech,mld}` namespaces, ATM input phantom nodes are placed into heaps and passed into `search` functions.  `search` functions must get phantom nodes as input and a reference to heaps structure. 
* `search` functions should internally clear heaps and place phantom nodes into heaps
* cleanup of algorithm-dependent heaps
* remove static heaps members and make heaps structure members life-time in scope of engines life-time

daniel-j-h:
This landed with https://github.com/Project-OSRM/osrm-backend/pull/3931.

PR #3931, opened 2017-04-11
Refactor heaps

oxidase:
# Issue

PR solves a regression introduced in with the MLD shortest path plugin by duplicating CH and CoreCH heaps that can be used simultaneously and potentially can double memory usage in heaps. 

PR was aimed to solve #3908, but hiding heaps into `search` functions would require a significant change in internal `engine` interfaces and could lead to a performance drop. The only solved point is a removal of core heaps from functions arguments and adding `SearchEngineData` in functions arguments.

The non-obvious part is [`HeapAlgorithm` derived type](https://github.com/Project-OSRM/osrm-backend/blob/ba233629e692e163a5f5574898c04207607a34d5/include/engine/routing_algorithms.hpp#L60-L67) that solves the regression and restores pre-MLD `SearchEngineData` for CH and CoreCH alorithms.

The current state state of the algorithms split is:
* algorithmic parts is placed in `ch`, `corech` and `mld` namespaces.
* data facades can be instantiated for `Algorithm` = {`ch`, `corech`, `mld`}
* `SearchEngineData` can be instantiated for `Algorithm` = {`ch`, `mld`}. Unfortunately this requires an additional template parameter in places like https://github.com/Project-OSRM/osrm-backend/blob/41740c2afc3a2899b09629107faf2b0006d455ff/include/engine/routing_algorithms/map_matching.hpp#L25
Allowed combinations of template parameters are

| `Algorithm` |  `SearchEngineData` |
|:-----------:|:-----------------:|
| `ch::Algorithm` | `ch::Algorithm` |
| `corech::Algorithm` | `ch::Algorithm` |
| `mld::Algorithm` | `mld::Algorithm` |

## Tasklist
 - [x] review
 - [x] adjust for comments

### DIFF
--- a/include/engine/routing_algorithms.hpp
+++ b/include/engine/routing_algorithms.hpp
@@ -57,7 +57,16 @@ class RoutingAlgorithmsInterface
 template <typename Algorithm> class RoutingAlgorithms final : public RoutingAlgorithmsInterface
 {
   public:
-    RoutingAlgorithms(SearchEngineData<Algorithm> &heaps,
+    // HeapAlgorithm is an explicit heap algorithm type to be used with Algorithm
+    // - CH algorithms use CH heap
+    // - CoreCH algorithms use CH heap
+    // - MLD algorithms use MLD heap
+    using HeapAlgorithm = typename std::conditional<
+        std::is_same<Algorithm, routing_algorithms::corech::Algorithm>::value,
+        routing_algorithms::ch::Algorithm,
+        Algorithm>::type;
+
+    RoutingAlgorithms(SearchEngineData<HeapAlgorithm> &heaps,
                       const datafacade::ContiguousInternalMemoryDataFacade<Algorithm> &facade)
         : heaps(heaps), facade(facade)
     {
@@ -122,7 +131,8 @@ template <typename Algorithm> class RoutingAlgorithms final : public RoutingAlgo
     }
 
   private:
-    SearchEngineData<Algorithm> &heaps;
+    SearchEngineData<HeapAlgorithm> &heaps;
+
     // Owned by shared-ptr passed to the query
     const datafacade::ContiguousInternalMemoryDataFacade<Algorithm> &facade;
 };
--- a/include/engine/routing_algorithms.hpp
+++ b/include/engine/routing_algorithms.hpp
@@ -57,16 +57,7 @@ class RoutingAlgorithmsInterface
 template <typename Algorithm> class RoutingAlgorithms final : public RoutingAlgorithmsInterface
 {
   public:
-    // HeapAlgorithm is an explicit heap algorithm type to be used with Algorithm
-    // - CH algorithms use CH heap
-    // - CoreCH algorithms use CH heap
-    // - MLD algorithms use MLD heap
-    using HeapAlgorithm = typename std::conditional<
-        std::is_same<Algorithm, routing_algorithms::corech::Algorithm>::value,
-        routing_algorithms::ch::Algorithm,
-        Algorithm>::type;
-
-    RoutingAlgorithms(SearchEngineData<HeapAlgorithm> &heaps,
+    RoutingAlgorithms(SearchEngineData<Algorithm> &heaps,
                       const datafacade::ContiguousInternalMemoryDataFacade<Algorithm> &facade)
         : heaps(heaps), facade(facade)
     {
@@ -131,7 +122,7 @@ template <typename Algorithm> class RoutingAlgorithms final : public RoutingAlgo
     }
 
   private:
-    SearchEngineData<HeapAlgorithm> &heaps;
+    SearchEngineData<Algorithm> &heaps;
 
     // Owned by shared-ptr passed to the query
     const datafacade::ContiguousInternalMemoryDataFacade<Algorithm> &facade;
--- a/include/engine/routing_algorithms/routing_base_ch.hpp
+++ b/include/engine/routing_algorithms/routing_base_ch.hpp
@@ -388,8 +388,8 @@ namespace corech
 // a force loop, if the heaps have been initialized with positive offsets.
 void search(SearchEngineData<Algorithm> &engine_working_data,
             const datafacade::ContiguousInternalMemoryDataFacade<corech::Algorithm> &facade,
-            SearchEngineData<Algorithm>::QueryHeap &forward_heap,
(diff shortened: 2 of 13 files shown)


include/engine/routing_algorithms.hpp
```diff
@@ -57,7 +57,16 @@ class RoutingAlgorithmsInterface
...
   public:
-    RoutingAlgorithms(SearchEngineData<Algorithm> &heaps,
+    // HeapAlgorithm is an explicit heap algorithm type to be used with Algorithm
+    // - CH algorithms use CH heap
+    // - CoreCH algorithms use CH heap
+    // - MLD algorithms use MLD heap
+    using HeapAlgorithm = typename std::conditional<
+        std::is_same<Algorithm, routing_algorithms::corech::Algorithm>::value,
+        routing_algorithms::ch::Algorithm,
+        Algorithm>::type;
```
daniel-j-h: Should we explicitly allow combinations and put a static assert here for everything else?

src/engine/routing_algorithms/routing_base_ch.cpp
```diff
@@ -71,13 +71,15 @@ void retrievePackedPathFromSingleHeap(const SearchEngineData<Algorithm>::QueryHe
 // && source_phantom.GetForwardWeightPlusOffset() > target_phantom.GetForwardWeightPlusOffset())
 // requires
 // a force loop, if the heaps have been initialized with positive offsets.
-void search(const datafacade::ContiguousInternalMemoryDataFacade<Algorithm> &facade,
+void search(SearchEngineData<Algorithm> & /*engine_working_data*/,
+            const datafacade::ContiguousInternalMemoryDataFacade<Algorithm> &facade,
```
daniel-j-h: Hm so here we still pass the facade _and_ the search engine data now?

include/engine/routing_algorithms.hpp
oxidase: Good idea! But i don't know what exactly to check: `HeapAlgorithm` is a derived type and not a parameter, so static asserts with `HeapAlgorithm` will just check `std::conditional` implementation. Also function definitions have explicit types, so checks on that side will be useless.

I would like to have a human-readable compiler message, at the moment incorrect types will lead to messages like " bla   **has not been declared.**  bla   **note: suggested alternatives:** bla bla bla "

src/engine/routing_algorithms/routing_base_ch.cpp
oxidase: yes, `SearchEngineData<Algorithm>` contains  mutable heaps, `facade` contains constant data.

It is the same convention as before: `SearchEngineData<Algorithm> &engine_working_data` just hides `heap_2` from the parameters list.

include/engine/routing_algorithms.hpp
TheMarex: Would `class SearchEngineData<CoreCH> : public SearchEngineData<CH> {}` work as well? That would enable us to remove the additional template parameter since we could use `SearchEngineData<Algorithm>`.

include/engine/routing_algorithms/routing_base_ch.hpp
```diff
@@ -350,63 +350,29 @@ void retrievePackedPathFromSingleHeap(const SearchEngineData<Algorithm>::QueryHe
 // && source_phantom.GetForwardWeightPlusOffset() > target_phantom.GetForwardWeightPlusOffset())
 // requires
 // a force loop, if the heaps have been initialized with positive offsets.
-void search(const datafacade::ContiguousInternalMemoryDataFacade<Algorithm> &facade,
+void search(SearchEngineData<Algorithm> &engine_working_data,
```
TheMarex: Passing `engine_working_data` as another parameter complicates the interface more, and makes it even less unit test-table. However I see why we need this right now.

Ideally I would like to see something like:
```cpp
std::pair<EdgeWeight, std::vector<NodeID>>
search(GraphT& graph,
       HeapT&,
       HeapT&,
       std::array<PhantomNode, N1>& sources,
       std::array<PhantoNode, N2>& targets);
```
Which would enable us to write unit tests for these functions (no `facade` and `engine_working_data` dependency). We should aim for this in `5.8`.

include/engine/search_engine_data.hpp
```diff
@@ -75,12 +79,8 @@ template <> struct SearchEngineData<routing_algorithms::mld::Algorithm>
 
     static SearchEngineHeapPtr forward_heap_1;
     static SearchEngineHeapPtr reverse_heap_1;
-    static SearchEngineHeapPtr forward_heap_2;
```
TheMarex: Just a heads up that @daniel-j-h will probably need at least two heaps for the alternative plugin down the road.

oxidase: I have added dummy heaps for another reason and before @daniel-j-h asked me why i added them, so at the moment it is a technical debt.

Good place to re-introduce them would be an alternative search implementation PR

include/engine/routing_algorithms/routing_base_ch.hpp
oxidase: The first point is about `SearchEngineData`:  some algorithms may need internally additional heaps and there is no possibility to tell how to get heaps without `SearchEngineData`. The proposal  is possible if [`SearchEngineData<Algorithm>`](https://github.com/Project-OSRM/osrm-backend/blob/e56d01b2e9deb8c22a501305e2e587ccd993353f/include/engine/routing_algorithms.hpp#L125) is moved in places where it is needed. `faced` argument exposes `Algorithm` and `SearchEngineData<Algorithm>` can be derived directly in `search` functions. It is doable in this PR.

The second point about `std::array`: there some places that already use [`std::vector`](https://github.com/Project-OSRM/osrm-backend/blob/e56d01b2e9deb8c22a501305e2e587ccd993353f/src/engine/routing_algorithms/routing_base_ch.cpp#L239-L240), so it is not clear how to generalize different approaches to pass initial points as arguments. It is ok to pass initial nodes directly in heaps, so the interface can be simpler
```
std::pair<EdgeWeight, std::vector<NodeID>>
search(GraphT& graph,
       HeapT&,
       HeapT&,
       Args... );
```

TheMarex:
Waiting on Travis to confirm this is 🍏 then we can move to merge + tag 5.7.0 RC1.

PR #3936, opened 2017-04-12
Use total angle for turn instruction if entry step has large distance

### LINKED ISSUE #3933: Collapse Scenario Failure
The following route ([link](http://map.project-osrm.org/?hl=de&z=13&loc=-33.964473845748486,151.1381596326828&loc=-33.964874269496946,151.13710280042142)) exhibits an unwanted collapsing behaviour.

The intersection in question looks like this:

map

The instruction shown is a `turn straight` which is obviously not the case and should be a `turn left`.
The instructions leading up to the turn are:

`Suppressed` and `turn slight left`. The instruction turning onto the road itself, however, is a `merge slight right`, as visible in [this request](https://router.project-osrm.org/route/v1/driving/151.137382,-33.964527;151.137135,-33.964919?overview=false&alternatives=true&steps=true&hints=;).

I see the following possible reasons, since (straight + left == ok [link](https://github.com/Project-OSRM/osrm-backend/blob/d6ac924b945546ce5727ffefd2a78c4536fcc035/src/engine/guidance/collapse_turns.cpp#L68-L85)).

The most likely error is that merge isn't correctly considered to compute the [angle itself](https://github.com/Project-OSRM/osrm-backend/blob/d6ac924b945546ce5727ffefd2a78c4536fcc035/src/engine/guidance/collapse_turns.cpp#L50-L51).

Somewhat in the same league would be the the [scenario operator](https://github.com/Project-OSRM/osrm-backend/blob/d6ac924b945546ce5727ffefd2a78c4536fcc035/src/engine/guidance/collapse_turns.cpp#L175-L250).

Less likely would be the scenarios not to trigger  [here](https://github.com/Project-OSRM/osrm-backend/blob/d6ac924b945546ce5727ffefd2a78c4536fcc035/src/engine/guidance/collapse_turns.cpp#L382-L383), [here](https://github.com/Project-OSRM/osrm-backend/blob/d6ac924b945546ce5727ffefd2a78c4536fcc035/src/engine/guidance/collapse_turns.cpp#L401-L408), or [here](https://github.com/Project-OSRM/osrm-backend/blob/d6ac924b945546ce5727ffefd2a78c4536fcc035/src/engine/guidance/collapse_turns.cpp#L409-L419).

A final possibility I see is that some other scenario (like [here](https://github.com/Project-OSRM/osrm-backend/blob/d6ac924b945546ce5727ffefd2a78c4536fcc035/src/engine/guidance/collapse_turns.cpp#L382-L389)) might be screwing with the detection.

oxidase:
# Issue

Fixes #3933. There is no obvious reason to not use total angles for entry steps with distance > 60 meters.

## Tasklist
 - [x] review
 - [x] adjust for comments

### DIFF
--- a/features/guidance/collapse.feature
+++ b/features/guidance/collapse.feature
@@ -1019,3 +1019,39 @@ Feature: Collapse
 			|       a,g | road,cross,cross	    | depart,fork left,arrive 			         | a,b,g 	 |
      		|       a,e | road,road,road 		| depart,fork slight right,arrive 			 | a,b,e 	 |
      		|       a,f | road,road,cross,cross | depart,fork slight right,turn right,arrive | a,b,d,f 	 |
+
+
+    # http://www.openstreetmap.org/way/92415447 #3933
+    Scenario: Use total angle for turn instruction if entry step has large distance
+        #   """
+        #    kf-_     a
+        #     |  - b
+        #    |  c
+        #    |d
+        #    e
+        #    |
+        #    i
+        #    """
+
+        And the node locations
+          | node |         lat |         lon |        #id |
+          | a    | -33.9644254 | 151.1378673 |   33226063 |
+          | b    | -33.9644373 | 151.1377172 | 1072787030 |
+          | c    | -33.9644791 | 151.1374452 | 4222903609 |
+          | d    | -33.9645661 | 151.1372654 | 4222903610 |
+          | e    | -33.9646986 | 151.1371539 | 4222903611 |
+          | f    |  -33.964386 | 151.1372133 | 1072786875 |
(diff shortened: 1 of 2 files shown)


MoKob:
@oxidase thank you for looking into this. One thing to watch out with changes like these: the change can affect a lot of other scenarios. If you can, maybe create a few routes and see where the changes affect them. The test scenarios can only do so much.

oxidase:
@TheMarex the test case  is very specific: it must be a sliproad with large enough entry step distance and  near straightforward entry angle. Without proper generalization in feature tests this change can affect guidance quality via `AdjustToCombinedTurnStrategy`, and probably it will arise again as wrong turn instructions based on the total turn angle.

PR #3937, opened 2017-04-12
Debug travis builds with different node version

TheMarex:
# Issue

Travis currently fails to select the right node version.

## Tasklist
 - [x] review
 - [x] adjust for comments

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -142,7 +142,7 @@ matrix:
         apt:
           sources: ['ubuntu-toolchain-r-test']
           packages: ['libstdc++-5-dev']
-      env: CLANG_VERSION='4.0.0' BUILD_TYPE='Release' ENABLE_MASON=ON ENABLE_LTO=ON ENABLE_NODE_BINDINGS=ON PUBLISH_NODE_BINDINGS=On JOBS=4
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -36,6 +37,7 @@ env:
    - CCACHE_VERSION=3.3.1
    - CMAKE_VERSION=3.7.2
    - MASON="$(pwd)/scripts/mason.sh"
+   - NODE="4"
```
daniel-j-h: I don't understand how this interacts with the `node_js` values set above. Is this just to have a default, and will the values above overwrite node 4 here?

.travis.yml
```diff
@@ -230,6 +228,10 @@ matrix:
         - ./scripts/travis/publish.sh
 
 before_install:
+  - source $NVM_DIR/nvm.sh
+  - nvm install $NODE
+  - nvm use $NODE
+  - node --version
```
daniel-j-h: nvm is installed on travis by default I assume?

.travis.yml
TheMarex: I need the values above to install `yarn` for me. I think it also causes node 4 and 6 to be in the nvm cache. That seems to work at least for non OSX builds which need an explicit `nvm install` step.

.travis.yml
TheMarex: Only if you set the global `node_js` though I think.

Issue #3378, opened 2016-11-29
Investigate STR packing for RTree


danpat:
The StaticRTree currently uses Hilbert-code based node ordering.  This gives us reasonable, but not optimal leaf-node layout.

Per https://github.com/Project-OSRM/osrm-backend/issues/3352#issuecomment-262579307, we should investigate whether an alternative packing would give us a performance boost - currently, there is some overlap in our RTree leafnodes, leading to unnecessary disk reads.

daniel-j-h:
Refs:

> Leutenegger, Scott T.; Edgington, Jeffrey M.; Lopez, Mario A. (1997).
> STR: A Simple and Efficient Algorithm for R-Tree Packing

> Garcia, Yvan J.; Lopez, Mario A.; Leutenegger, Scott T. (1997).
> A Greedy Algorithm for Bulk Loading R-trees 

For the record Boost.Geometry's R-Tree provides a range ctor for a STR-packing construction. But we also had some issues with this R-Tree at the time we tried it in a prototype with Boost 1.55 / 1.56 I think.

danpat:
Also, some possible improvements in a newer approach: http://www.cs.umd.edu/~hjs/pubs/AlborziIPL07.pdf

Note that Boost's packing constructor uses a mix of STR and TGS:  https://github.com/boostorg/geometry/blob/develop/include/boost/geometry/index/detail/rtree/pack_create.hpp#L110

PR #3941, opened 2017-04-13
OMT packing for rtree.

danpat:
# Issue

This PR is to address https://github.com/Project-OSRM/osrm-backend/issues/3378.  Our current RTree uses a Hilbert sort to group leaf nodes.  For road network data, this gives us a less-than-optimal packing with quite a lot of overlap for leaf nodes.  We can speed up coordinate snapping by using a better packing method.

This PR implements the [Overlap Minimizing Top-down bulk-loading](http://ftp.informatik.rwth-aachen.de/Publications/CEUR-WS/Vol-74/files/FORUM_18.pdf) in our rtree, which produces an RTree with far less overlap than the current Hilbert Sort approach.

As a bonus, OMT makes it easy to back the TreeNode data into a breadth-first-search linear order, which means we don't need fixed-sized child arrays in the TreeNode struct, range pointers are sufficient.  This greatly reduces the size of the `.ramIndex` files.

## Tasklist
 - [ ] compare to Boost's packing approach, which is an [altered STR method](https://github.com/boostorg/geometry/blob/develop/include/boost/geometry/index/detail/rtree/pack_create.hpp#L110)
 - [ ] benchmarking
 - [ ] review
 - [ ] adjust for comments

### DIFF
--- a/features/car/bridge.feature
+++ b/features/car/bridge.feature
@@ -33,9 +33,9 @@ Feature: Car - Handle driving
     Scenario: Car - Control test without durations, osrm uses movable bridge speed to calculate duration
         Given the node map
             """
-            a b c
+            a b1c
                 d
-                e f g
+                e2f g
             """
 
         And the ways
@@ -48,8 +48,8 @@ Feature: Car - Handle driving
             | from | to | route           | modes                           | speed   | time     |
             | a    | g  | abc,cde,efg,efg | driving,driving,driving,driving | 13 km/h | 340s +-1 |
             | b    | f  | abc,cde,efg,efg | driving,driving,driving,driving | 9 km/h  | 318s +-1 |
-            | c    | e  | cde,cde         | driving,driving                 | 5 km/h  | 295s +-1 |
-            | e    | c  | cde,cde         | driving,driving                 | 5 km/h  | 295s +-1 |
+            | 1    | 2  | abc,cde,efg     | driving,driving,driving         | 7 km/h  | 306s +-1 |
+            | 2    | 1  | efg,cde,abc     | driving,driving,driving         | 7 km/h  | 306s +-1 |
 
     Scenario: Car - Properly handle durations
         Given the node map
--- a/include/util/static_rtree.hpp
+++ b/include/util/static_rtree.hpp
@@ -55,7 +55,7 @@ namespace util
 // are computed, this means the internal distance metric doesn not represent meters!
 template <class EdgeDataT,
           storage::Ownership Ownership = storage::Ownership::Container,
-          std::uint32_t BRANCHING_FACTOR = 63>
+          std::uint32_t BRANCHING_FACTOR = 64>
 class StaticRTree
 {
     template <typename T> using Vector = ViewOrVector<T, Ownership>;
@@ -268,7 +268,7 @@ class StaticRTree
                     // Calculate the bounding-box for this leaf node based on the
                     // endpoints of the EdgeDataT values contained within
                     std::for_each(edges.begin() + current_tile.left_edge_index,
-                                  edges.begin() + current_tile.right_edge_index,
+                                  edges.begin() + current_tile.left_edge_index + number_of_edges,
                                   [this, &leaf_node](const EdgeDataT &edge) {
                                       Coordinate projected_u{web_mercator::fromWGS84(
                                           Coordinate{m_coordinate_list[edge.u]})};
@@ -596,32 +596,24 @@ class StaticRTree
                 const auto &current_candidate =
                     CandidateSegment{current_query_node.fixed_projected_coordinate, edge_data};
 
-                /*
-                                std::clog << "Selecting item at " <<
-                   current_query_node.squared_min_dist
-                                          << std::endl;
-                                          */
-                // to allow returns of no-results if too restrictive filtering, this needs
-                // to be
-                // done here even though performance would indicate that we want to stop
-                // after
+                // to allow returns of no-results if too restrictive filtering, this needs to be
+                // done here even though performance would indicate that we want to stop after
                 // adding the first candidate
                 if (terminate(results.size(), current_candidate))
                 {
-                    // std::clog << "Terminating" << std::endl;
                     break;
                 }
 
+                // Check if either directino of the segment passes our filter.  If not
+                // continue to the next candidate
                 auto use_segment = filter(current_candidate);
                 if (!use_segment.first && !use_segment.second)
                 {
-                    // std::clog << "Can't use" << std::endl;
                     continue;
                 }
                 edge_data.forward_segment_id.enabled &= use_segment.first;
                 edge_data.reverse_segment_id.enabled &= use_segment.second;
 
-                // std::clog << "Pushing result" << std::endl;
                 // store phantom node in result vector
                 results.push_back(std::move(edge_data));
             }
--- a/include/util/static_rtree.hpp
+++ b/include/util/static_rtree.hpp
@@ -385,10 +385,12 @@ class StaticRTree
 
                 // Now, for each column we generated, sort those by latitude, so the data
                 // is ready for the subsequent tile-by-tile processing in the next loop.
-                tbb::parallel_do(tiles, [&](const std::pair<EdgeIndex, EdgeIndex> tile) {
-                    tbb::parallel_sort(
-                        edges.begin() + tile.first, edges.begin() + tile.second, latitude_compare);
-                });
+                tbb::parallel_do(
+                    tiles.begin(), tiles.end(), [&](const std::pair<EdgeIndex, EdgeIndex> tile) {
+                        tbb::parallel_sort(edges.begin() + tile.first,
+                                           edges.begin() + tile.second,
+                                           latitude_compare);
+                    });
             }
 
             // Because we used a queue above, the m_search_tree vector is sorted in
--- a/include/util/static_rtree.hpp
+++ b/include/util/static_rtree.hpp
@@ -6,7 +6,6 @@
 #include "util/coordinate_calculation.hpp"
 #include "util/deallocating_vector.hpp"
 #include "util/exception.hpp"
-#include "util/hilbert_value.hpp"
 #include "util/integer_range.hpp"
 #include "util/rectangle.hpp"
 #include "util/typedefs.hpp"
@@ -23,11 +22,13 @@
 #include <boost/format.hpp>
 #include <boost/iostreams/device/mapped_file.hpp>
 
+#include <tbb/parallel_do.h>
 #include <tbb/parallel_for.h>
 #include <tbb/parallel_sort.h>
 
 #include <algorithm>
 #include <array>
+#include <cmath>
 #include <limits>
 #include <memory>
 #include <queue>
@@ -54,8 +55,7 @@ namespace util
 // are computed, this means the internal distance metric doesn not represent meters!
 template <class EdgeDataT,
           storage::Ownership Ownership = storage::Ownership::Container,
-          std::uint32_t BRANCHING_FACTOR = 128,
-          std::uint32_t LEAF_PAGE_SIZE = 4096>
+          std::uint32_t BRANCHING_FACTOR = 63>
 class StaticRTree
 {
     template <typename T> using Vector = ViewOrVector<T, Ownership>;
@@ -64,11 +64,6 @@ class StaticRTree
     using Rectangle = RectangleInt2D;
     using EdgeData = EdgeDataT;
     using CoordinateList = Vector<util::Coordinate>;
-
-    static_assert(LEAF_PAGE_SIZE >= sizeof(uint32_t) + sizeof(EdgeDataT), "page size is too small");
-    static_assert(((LEAF_PAGE_SIZE - 1) & LEAF_PAGE_SIZE) == 0, "page size is not a power of 2");
-    static constexpr std::uint32_t LEAF_NODE_SIZE =
-        (LEAF_PAGE_SIZE - sizeof(uint32_t) - sizeof(Rectangle)) / sizeof(EdgeDataT);
 
     struct CandidateSegment
     {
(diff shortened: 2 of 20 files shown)


danpat:
Just to confirm that this avenue is worth pursuing, I did some quick profiling on the existing RTree using `src/benchmarks/static_rtree.cpp`.

This code runs two test scenarios - each takes about 3.5 seconds on my machine, single threaded.

Below shows the breakdown of one of the scenarios.  The key observation is that we're spending about 2.84 seconds out of 3.49 seconds (81%) of our time in the `ExploreLeafNode` method.  Leaf-node exploration is exactly what STR should reduce.

screen shot 2017-04-13 at 10 35 51 pm

danpat:
Also benchmarked the table plugin for a `/table` request with 1000 nodes:

screen shot 2017-04-13 at 11 33 55 pm

At this size, we're spending 13% of our time in the RTree search.

For a 100 coordinate `/table`, it's 15%:

screen shot 2017-04-13 at 11 36 09 pm

And for 500 coordinates:

screen shot 2017-04-13 at 11 37 39 pm

(Note for the curious: these results are from the Timing component of `Instruments.app` on OSX (part of XCode))

/cc @daniel-j-h

include/util/static_rtree.hpp
```diff
@@ -166,15 +167,245 @@ class StaticRTree
...
     explicit StaticRTree(const std::vector<EdgeDataT> &input_data_vector,
                          const std::string &tree_node_filename,
                          const std::string &leaf_node_filename,
-                         const Vector<Coordinate> &coordinate_list)
+                         const Vector<Coordinate> &coordinate_list,
+                         const PackingMethod packing_method = Hilbert)
         : m_coordinate_list(coordinate_list)
     {
+        util::Log() << "Starting rtree";
+        switch (packing_method)
```
oxidase: This is good for the algorithm selection at run-time, but packing method can be generalized out and selected at compile-time using a packing policy

include/util/static_rtree.hpp
```diff
@@ -166,15 +167,245 @@ class StaticRTree
...
+                                  [this](const EdgeDataT &a, const EdgeDataT &b) {
+                                      auto a_centroid = coordinate_calculation::centroid(
+                                          m_coordinate_list[a.u], m_coordinate_list[a.v]);
+                                      auto b_centroid = coordinate_calculation::centroid(
+                                          m_coordinate_list[b.u], m_coordinate_list[b.v]);
+                                      return a_centroid.lat < b_centroid.lat;
+                                  });
+                          });
+
+        // open leaf file
```
oxidase: Code below is a duplication. Better to avoid this by dropping `HilbertInputElement` (it will require computation of Hilbert values in comparator, but reduce memory consumption :sunglasses:) and use a generalized `PackingPolicy` for soring.

include/util/static_rtree.hpp
```diff
@@ -166,15 +167,245 @@ class StaticRTree
...
+                                                                   m_coordinate_list[b.v]);
+                return a_centroid.lon < b_centroid.lon;
+
+            });
+
+        // Now, break into even-sized vertical slices, and sort each slice by latitude
+        // TODO: make sure this isn't 0
+        const std::size_t leaf_K = std::sqrt(copy.size() / LEAF_NODE_SIZE);
+        util::Log() << "leaf_K is " << leaf_K;
+        tbb::parallel_for(tbb::blocked_range<uint64_t>(0, copy.size(), leaf_K),
```
oxidase: I would expect [floor(sqrt(15809/127))] = 11 vertical stripes for monaco but the result is
str

The problem that `leaf_K` is used as `grainsize` by `tbb::blocked_range` and is not what expected.

danpat:
Hokay, assuming I haven't messed this up:

Before:

```
$ ls -l san-francisco_california.osrm.*Index
-rw-r--r--  1 danpat  staff  14045184 Apr 22 13:07 san-francisco_california.osrm.fileIndex
-rw-r--r--  1 danpat  staff     14912 Apr 22 13:07 san-francisco_california.osrm.ramIndex
$ ./src/benchmarks/rtree-bench san-francisco_california.osrm.ramIndex san-francisco_california.osrm.fileIndex san-francisco_california.osrm.nodes
Running raw RTree queries (1 result) with 10000 coordinates: Took 2.67895 seconds (2678.95ms)  ->  0.267895 ms/query (2678.95ms)
Running raw RTree queries (10 results) with 10000 coordinates: Took 3.53938 seconds (3539.38ms)  ->  0.353938 ms/query (3539.38ms)
```

After:

```
$ ls -l san-francisco_california.osrm.*Index
-rw-r--r--  1 danpat  staff  33554432 Apr 25 01:20 san-francisco_california.osrm.fileIndex
-rw-r--r--  1 danpat  staff      3160 Apr 25 01:20 san-francisco_california.osrm.ramIndex
$ ./src/benchmarks/rtree-bench san-francisco_california.osrm.ramIndex san-francisco_california.osrm.fileIndex san-francisco_california.osrm.nodes
Running raw RTree queries (1 result) with 10000 coordinates: Took 1.03206 seconds (1032.06ms)  ->  0.103206 ms/query (1032.06ms)
Running raw RTree queries (10 results) with 10000 coordinates: Took 1.83538 seconds (1835.38ms)  ->  0.183538 ms/query (1835.38ms)
```

The `.ramIndex` file is only 21% of the previous size - this is a legit space saving and means we can use a much smaller `BRANCHING_FACTOR` (and probably where much of the performance increase comes from, but not all).

Speed wise, it looks like queries only take 38% of the time they previously did.  Not a bad speedup at all.

I'm not quite sure why the `.fileIndex` has doubled - I suspect I've mis-calculated the `LEAF_NODE_PAGE_SIZE` and there is a lot of dead space.  Should be fixable.

include/util/static_rtree.hpp
```diff
@@ -65,11 +79,20 @@ class StaticRTree
     using EdgeData = EdgeDataT;
     using CoordinateList = Vector<util::Coordinate>;
 
+    static constexpr auto UNALIGNED_LEAF_PAGE_SIZE =
+        BRANCHING_FACTOR * sizeof(EdgeDataT) + sizeof(Rectangle) + sizeof(std::uint32_t);
```
TheMarex: This yields a page size of `516` bytes, which rounds out to 1024. Do you need to use `BRANCHING_FACTOR` for the leaves too?

include/util/static_rtree.hpp
```diff
@@ -166,171 +170,329 @@ class StaticRTree
     StaticRTree(const StaticRTree &) = delete;
     StaticRTree &operator=(const StaticRTree &) = delete;
 
-    // Construct a packed Hilbert-R-Tree with Kamel-Faloutsos algorithm [1]
+    // In-place partial sort-by-group
+    template <typename Iterator, typename Compare>
+    void grouped_partial_sort(Iterator left, Iterator right, std::size_t n, Compare compare)
```
oxidase: Unused sorting?

oxidase:
i have spotted in OMT leafs
omt that one leaf has large size and contains one distant node 
omt1
Probably there is a problem with ceiling?

I have pushed the script that dumps r-trees at https://github.com/oxidase/home/blob/master/notes/topics/osrm/dump_rtree.py

TheMarex:
Given the problem with leaf node packing maybe we can get around the concept of a `LeafNode` completely and only store a continuous array of `EdgeDataT` on disk:

1. Instead of saving the bounding box of each `TreeNode` we save the bounding boxes of each child: `std::array<Rectangle, BRANCHING_FACTOR> child_minimum_bounding_rectangle`
2. `first_child_index` already has the dual use of either pointing to a `LeafNode` or another `TreeNode`. Instead we could also make it point at an entry of `vector_view<EdgeDataT>`
3. The query doesn't need to change much, because we already need to fetch every childs MBR here https://github.com/Project-OSRM/osrm-backend/blob/master/include/util/static_rtree.hpp#L577-L583

This should also cache-optimize the query since processing a `TreeNode` is now memory local (child bounding boxes and IDs are already stored in the node).

include/util/static_rtree.hpp
danpat: Yes, that was my next thought.  I'm going to remove the `LeafNode` object and just use  `TreeNode` all the way down (with an `is_leaf`) flag.  This will allow for variable size leaf nodes.

My hope is that the performance gains from the better packing and smaller node sizes will outweigh the benefits of the improved disk cache interaction.

src/util/coordinate_calculation.cpp
```diff
@@ -34,6 +34,13 @@ std::uint64_t squaredEuclideanDistance(const Coordinate lhs, const Coordinate rh
 
 double haversineDistance(const Coordinate coordinate_1, const Coordinate coordinate_2)
```
oxidase: I would keep the original function as `haversineDistance` and the new one rename to `greatCircleDistanceFCC`,
Also `greatCircleDistance` must have the same interface and can be used in  a proxy function like `calculateDistance() { return greatCircleDistanceFCC(); } `.  It is really helpful to have a fast flip between "exact" and approximated distance values.

features/car/bridge.feature
```diff
@@ -48,8 +48,8 @@ Feature: Car - Handle driving
             | from | to | route           | modes                           | speed   | time     |
             | a    | g  | abc,cde,efg,efg | driving,driving,driving,driving | 13 km/h | 340s +-1 |
             | b    | f  | abc,cde,efg,efg | driving,driving,driving,driving | 9 km/h  | 318s +-1 |
-            | c    | e  | cde,cde         | driving,driving                 | 5 km/h  | 295s +-1 |
-            | e    | c  | cde,cde         | driving,driving                 | 5 km/h  | 295s +-1 |
+            | 1    | 2  | abc,cde,efg     | driving,driving,driving         | 7 km/h  | 306s +-1 |
```
oxidase: Would it be better if tests specify also source and  target bearings to avoid changes in expected values?

include/util/static_rtree.hpp
```diff
@@ -158,189 +127,329 @@ class StaticRTree
...
-            tbb::blocked_range<uint64_t>(0, element_count),
-            [&input_data_vector, &input_wrapper_vector, this](
-                const tbb::blocked_range<uint64_t> &range) {
-                for (uint64_t element_counter = range.begin(), end = range.end();
-                     element_counter != end;
-                     ++element_counter)
-                {
-                    WrappedInputElement &current_wrapper = input_wrapper_vector[element_counter];
-                    current_wrapper.m_array_index = element_counter;
+        auto leaves = const_cast<std::vector<EdgeDataT> &>(input_data_vector);
```
TheMarex: Why the const cast?

include/util/static_rtree.hpp
```diff
@@ -158,189 +127,329 @@ class StaticRTree
...
-            });
+        while (!queue.empty())
+        {
+            auto r = queue.front();
+            queue.pop();
 
-        // open leaf file
-        boost::filesystem::ofstream leaf_node_file(leaf_node_filename, std::ios::binary);
+            auto N = r.right - r.left + 1;
+            auto M = BRANCHING_FACTOR;
```
TheMarex: These variables could be named more intuitively like 'num_children'.

include/util/static_rtree.hpp
```diff
@@ -158,189 +127,329 @@ class StaticRTree
...
 
-        // sort the hilbert-value representatives
-        tbb::parallel_sort(input_wrapper_vector.begin(), input_wrapper_vector.end());
-        std::vector<TreeNode> tree_nodes_in_level;
+            // We're processing a leaf
+            if (N <= M)
+            {
+                TreeNode t;
```
TheMarex: Instead of one-letter variable `t`, `node` would be more readable.

include/util/static_rtree.hpp
```diff
@@ -158,189 +127,329 @@ class StaticRTree
...
-                        web_mercator::latToY(toFloating(current_centroid.lat)))};
+        // position of the last leaf node written to diskcountindex
+        std::size_t saved_edges_count = 0;
 
-                    current_wrapper.m_hilbert_value = GetHilbertCode(current_centroid);
-                }
-            });
+        while (!queue.empty())
+        {
+            auto r = queue.front();
```
TheMarex: `r` -> `range` would help readability.

include/util/static_rtree.hpp
```diff
@@ -158,189 +127,329 @@ class StaticRTree
...
+                        grouped_partial_sort(
+                            leaves.begin() + r.left, leaves.begin() + r.right, N,
+               longitude_compare);
+                            */
 
-            tree_nodes_in_level.emplace_back(current_node);
-        }
-        leaf_node_file.flush();
-        leaf_node_file.close();
+            std::sort(leaves.begin() + r.left, leaves.begin() + r.right, longitude_compare);
```
TheMarex: This can be replaced by `tbb::parallel_sort`

include/util/static_rtree.hpp
```diff
@@ -158,189 +127,329 @@ class StaticRTree
...
-                // pack BRANCHING_FACTOR elements into tree_nodes each
-                for (std::uint32_t current_child_node_index = 0;
-                     current_child_node_index < BRANCHING_FACTOR;
-                     ++current_child_node_index)
+                auto right2 = std::min(i + N1 - 1, r.right);
+                /*
+                grouped_partial_sort(
+                    leaves.begin() + i, leaves.begin() + right2, N2, latitude_compare);
+                    */
+                std::sort(leaves.begin() + i, leaves.begin() + right2, latitude_compare);
```
TheMarex: Can be replaced with `tbb::parallel_sort`

include/util/static_rtree.hpp
```diff
@@ -158,189 +127,329 @@ class StaticRTree
...
+            }
 
-                // append the leaf node to the current tree node
-                current_node.child_count += 1;
-                current_node.children[leaf_index] =
-                    TreeIndex{node_index * BRANCHING_FACTOR + leaf_index, true};
-                current_node.minimum_bounding_rectangle.MergeBoundingBoxes(
-                    current_leaf.minimum_bounding_rectangle);
+            std::size_t N2 = std::ceil(static_cast<double>(N) / M);
+            std::size_t N1 = N2 * std::ceil(std::sqrt(M));
```
TheMarex: `N1` and `N2` make the following code rather hard to follow. `N2` is the number of children, right?

src/benchmarks/static_rtree.cpp
```diff
@@ -34,14 +36,9 @@ std::vector<util::Coordinate> loadCoordinates(const boost::filesystem::path &nod
...
 
-    extractor::QueryNode current_node;
-    auto coordinate_count = nodes_path_file_reader.ReadElementCount64();
-    std::vector<util::Coordinate> coords(coordinate_count);
-    for (unsigned i = 0; i < coordinate_count; ++i)
-    {
-        nodes_path_file_reader.ReadInto(&current_node, 1);
-        coords[i] = util::Coordinate(current_node.lon, current_node.lat);
-    }
+    std::vector<util::Coordinate> coords;
```
TheMarex: This could actually reuse the `extractor::files::readNodes` function.

daniel-j-h:
@danpat can we get a small rundown of what's the problem here or why your approach didn't work out? I'd love to get some knowledge out of this pull request wrt. rtree mechanics.

danpat:
@daniel-j-h the downside to OMT was the overall data size.  There was a performance gain, but it came at the expense of a fairly large increase in memory usage.  Because OMT uses a top-down packing approach, it was difficult to tweak the branching factor to get the right performance/size balance.  The OMT `.ramIndex` file was turning out 10x or larger.

Our current approach where the lowest level has many more branches than the majority of the tree seems to perform better overall, and requires less space (more branches at the leaf nodes = fewer nodes in the tree altogether = smaller trees).

The OMT construction method is also slow - about 5x slower than the Hilbert Sort.

I just opened up #4056 that ports some of the data structure changes from this PR into our existing Hilbert Sort structure - it speeds things up by >30% for a reasonable increase in RAM usage.

At this stage, I don't think OMT actually buys us anything.

Closing in favour of #4056

PR #3959, opened 2017-04-19
Apply traffic light penalty also for non-turns

### LINKED ISSUE #3958: No traffic_light penalties applied for degree=2 nodes
I think our turn_function is not applying penalties to nodes that it should (specifically, traffic lights):

https://github.com/Project-OSRM/osrm-backend/blob/e96545be2ebebee67791dff9afaed1c6f05462b2/profiles/car.lua#L395

For nodes like this:

http://www.openstreetmap.org/node/1675470014#map=18/53.48624/-2.33026

we tag the turn type as `no_turn` - this means that we apply no penalty in the `turn_function`.  We're missing traffic light penalties in these scenarios.

c0d9ee5c-243a-11e7-8f10-a4406c4ef293

TheMarex:
# Issue

Fixes #3958 by applying the traffic light penalty also to non-turns.

## Tasklist
 - [x] add regression
 - [x] review
 - [x] adjust for comments

## Requirements / Relations

#3958

### DIFF
--- a/features/car/traffic_light_penalties.feature
+++ b/features/car/traffic_light_penalties.feature
@@ -1,0 +1,39 @@
+@routing @car @traffic_light
+Feature: Car - Handle traffic lights
+
+    Background:
+        Given the profile "car"
+
+    Scenario: Car - Encounters a traffic light
+        Given the node map
+            """
+            a-1-b-2-c
+
+            d-3-e-4-f
+
+            g-h-i   k-l-m
+              |       |
+              j       n
+
+            """
+
+        And the ways
+            | nodes | highway |
+            | abc   | primary |
+            | def   | primary |
+            | ghi   | primary |
+            | klm   | primary |
+            | hj    | primary |
+            | ln    | primary |
+
+        And the nodes
+            | node | highway         |
+            | e    | traffic_signals |
+            | l    | traffic_signals |
+
+        When I route I should get
+            | from | to | time   | # |
+            | 1    | 2  |  11.1s | no turn with no traffic light |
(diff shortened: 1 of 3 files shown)


features/car/traffic_light_penalties.feature
```diff
@@ -0,0 +1,27 @@
...
+Feature: Car - Handle traffic lights
+
+    Background:
+        Given the profile "car"
+
+    Scenario: Car - Encounters a traffic light
+        Given the node map
+            """
+            a b c 1 e 2 f
+                d
```
daniel-j-h: Can you in addition add a test for a traffic light in the profile's turn branch?

profiles/car.lua
```diff
@@ -392,6 +392,10 @@ function turn_function (turn)
   local turn_penalty = profile.turn_penalty
   local turn_bias = profile.turn_bias
 
+  if turn.has_traffic_light then
+      turn.duration = turn.duration + profile.traffic_light_penalty
+  end
```
daniel-j-h: Argh - great we can get this into 5.7 still :+1:

TheMarex:
@daniel-j-h can you take a final look? Actually found a problem again.

features/car/traffic_light_penalties.feature
```diff
@@ -0,0 +1,39 @@
...
+        Given the node map
+            """
+            a-1-b-2-c
+
+            d-3-e-4-f
+
+            g-h-i   k-l-m
+              |       |
+              j       n
+
```
daniel-j-h: Thanks for adding.

Issue #3952, opened 2017-04-18
Add random access to PackedVector


TheMarex:
Currently we can't use the `PackedVector` for `SegmentWeight` and `SegmentDuration` in https://github.com/Project-OSRM/osrm-backend/pull/3929 because we need random write access on the values during updating weights e.g.

https://github.com/Project-OSRM/osrm-backend/blob/master/src/updater/updater.cpp#L208-L212
https://github.com/Project-OSRM/osrm-backend/blob/master/src/updater/updater.cpp#L241-L245

This would mean we need a different `PackedVector` implementation. If we assume that every entry will have less then `64bit` we could pre-compute 64 pairs of bit masks.

We can then compute the result with something like:

```
result = (lower_mask[index % 64] & block[index % num_entries]) >> lower_offset[index % 64] |
         (upper_mask[index % 64] & block[index % num_entries + 1]) >> upper_offset[index % 64]
```

PR #3966, opened 2017-04-21
Switch from macro based StrongTypedef to template version

TheMarex:
# Issue

This enables us to use lazy instantiation of member functions. Namely we an now have a `operator|` that does not cause compile errors for `From = double` (only when one actually tries to use it).

## Tasklist
 - [x] review
 - [x] adjust for comments

Needed for #3952


include/util/alias.hpp
```diff
@@ -0,0 +1,112 @@
...
+(INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
+LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON
+ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+(INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+*/
+
+#ifndef ALIAS_HPP
+#define ALIAS_HPP
```
daniel-j-h: OSRM_

include/util/alias.hpp
```diff
@@ -0,0 +1,112 @@
...
+    return stream << inst.__value;
+}
+
+template <typename T>
+struct is_valid_alias
+    : std::integral_constant<bool,
+                             std::is_trivial<T>::value && std::is_standard_layout<T>::value &&
+                                 std::is_pod<T>::value>
+{
+};
```
daniel-j-h: Hmm why not e.g. (not tested)

    template <typename T, typename Alias>
    struct is_alias_for : is_same<typename Alias::From, T> { };

    template <typename T, typename Alias>
    void assert_alias_for( static_assert(is_alias_for<T, Alias>, ""); );

Then we can check e.g.

    assert_alias_for<int32_t, OsmId>();

include/util/alias.hpp
```diff
@@ -0,0 +1,112 @@
...
+#include <type_traits>
+
+namespace osrm
+{
+
+template <typename From, typename Tag> struct Alias;
+template <typename From, typename Tag>
+inline std::ostream &operator<<(std::ostream &stream, const Alias<From, Tag> &inst);
+
+template <typename From, typename Tag> struct Alias final
```
daniel-j-h: If you make `Alias` a bit smarter and special case for primitive types vs. non-primitive types, you can inherit from non-primitive types and :sparkles: we have proper newtype semantics e.g. for vectors:

https://github.com/mapbox/node-or-tools/blob/master/src/types.h#L13-L23

compiler will complain if you mix things up, conversion is still possible to its base type

include/engine/routing_algorithms/routing_base.hpp
```diff
@@ -320,17 +320,19 @@ double getPathDistance(const datafacade::ContiguousInternalMemoryDataFacade<Algo
...
 
         const double current_lat =
-            static_cast<double>(toFloating(current_coordinate.lat)) * DEGREE_TO_RAD;
+            static_cast<double>(util::toFloating(current_coordinate.lat)) * DEGREE_TO_RAD;
         const double current_lon =
-            static_cast<double>(toFloating(current_coordinate.lon)) * DEGREE_TO_RAD;
+            static_cast<double>(util::toFloating(current_coordinate.lon)) * DEGREE_TO_RAD;
         const double current_cos = std::cos(current_lat);
 
         const double sin_dlon = std::sin((prev_lon - current_lon) / 2.0);
```
oxidase: if types now have proper `operator-` and `operator/(double)` then here degrees be subtracted first and divided by 2 and than converted to double radians

include/util/alias.hpp
```diff
@@ -0,0 +1,112 @@
...
+    }
+    Alias operator-(const Alias rhs_) const
+    {
+        return Alias{__value - static_cast<const From>(rhs_)};
+    }
+    Alias operator*(const Alias rhs_) const
+    {
+        return Alias{__value * static_cast<const From>(rhs_)};
+    }
+    Alias operator/(const Alias rhs_) const
```
oxidase: here better to add also `Alias operator/(const double denom) const` for scaling

include/util/alias.hpp
```diff
@@ -0,0 +1,112 @@
...
+inline std::ostream &operator<<(std::ostream &stream, const Alias<From, Tag> &inst)
+{
+    return stream << inst.__value;
+}
+
+template <typename T>
+struct is_valid_alias
+    : std::integral_constant<bool,
+                             std::is_trivial<T>::value && std::is_standard_layout<T>::value &&
+                                 std::is_pod<T>::value>
```
oxidase: i guess `is_pod` in c++11 terms is enough http://en.cppreference.com/w/cpp/types/is_pod

include/util/alias.hpp
TheMarex: The trait is more supposed to check if it really is just a type alias and will be optimized away. Will just check `is_pod` instead.

TheMarex:
@danpat checked the assembly, seems to boil down to virtually the same instructions in both cases.

Issue #3979, opened 2017-04-26
Intersection array incorrect when departure point is inside a roundabout


danpat:
It looks like we're incorrectly specifying either the `exit` value or missing some `intersections[]` data in a response when a route begins on a roundabout.

Example: The following route that begins on a roundabout:

http://map.project-osrm.org/?z=18&center=49.079871%2C6.111738&loc=49.079650%2C6.111606&loc=49.080061%2C6.111934&hl=en&alt=0

The second step in this response looks like this:
```
{
   "name" : "",
   "intersections" : [
      {
         "bearings" : [ 75, 165, 330 ],
         "location" : [ 6.111552, 49.079856 ],
         "in" : 1,
         "out" : 0,
         "entry" : [ true, false, true ]
      }
   ],
   "duration" : 11.8,
   "weight" : 11.8,
   "maneuver" : {
      "bearing_after" : 68,
      "modifier" : "right",
      "type" : "roundabout",
      "location" : [ 6.111552, 49.079856 ],
      "bearing_before" : 348,
      "exit" : 1
   },
   "distance" : 51.3,
   "geometry" : "_{pjHathd@KEKAKBQoA",
   "mode" : "driving"
}
```

Note that `exit: 1` in the `maneuver` is describing `intersections[0]`.

If you move the marker slightly south so the route does *not* begin on the roundabout, the second step now looks like this:

http://map.project-osrm.org/?z=18&center=49.079873%2C6.111738&loc=49.079428%2C6.111515&loc=49.080061%2C6.111934&hl=en&alt=0

```
{
   "name" : ""
   "intersections" : [
      {
         "bearings" : [ 60, 165, 255 ],
         "location" : [ 6.111398, 49.079596 ],
         "in" : 1,
         "out" : 0,
         "entry" : [ true, false, false ]
      },
      {
         "bearings" : [ 75, 165, 330 ],
         "location" : [ 6.111552, 49.079856 ],
         "in" : 1,
         "out" : 0
         "entry" : [ true, false, true ],
      }
   ],
   "duration" : 12.6,
   "weight" : 12.6,
   "maneuver" : {
      "bearing_after" : 54
      "modifier" : "right",
      "type" : "roundabout",
      "location" : [ 6.111398, 49.079596 ],
      "bearing_before" : 338,
      "exit" : 1,
  },
   "distance" : 64.7,
   "mode" : "driving",
   "geometry" : "ozpjHgshd@EMIKKEKAKBQoA",
[...]

PR #3980, opened 2017-04-27
Fix incorrect exit turn invalidation

### LINKED ISSUE #3977: Incorrect turn rejection in invalidateExitAgainstDirection
[`invalidateExitAgainstDirection`](https://github.com/Project-OSRM/osrm-backend/blob/0e39aa9488bff97cffda1d12fb18cdeb537e0c03/src/extractor/guidance/roundabout_handler.cpp#L136-L139) rejects turns that go against roundabout direction from a roundabout exit to an entry edge (red sectors in 
roundabout)

This  condition works if the entry edge is outside of the roundabout (entry-edge   ..  roundabout entry ... roundabout exit ... ) but fails if the entry edge is inside the roundabout - the right use case in the figure  (entry-edge   ..  roundabout exit ... roundabout entry ... ) .

Failing request http://map.project-osrm.org/?z=18&center=38.917065%2C-76.978100&loc=38.917111%2C-76.979465&loc=38.917182%2C-76.976729&hl=de&alt=0

/cc @MoKob

oxidase:
# Issue

Fixes #3977 by checking an additional case if a roundabout entry is before a roundabout exit in counter-clockwise order. The check works for both CCW and CW roundabouts.

Fixed guidance instructions in #3977 
fixed_exit
and  https://github.com/Project-OSRM/osrm-backend/blob/d441b0975834686f92c9544e89fd7a93ff4c30e9/features/guidance/roundabout.feature#L766-L771 may be related to #3979

## Tasklist
 - [x] review
 - [x] adjust for comments

### DIFF
--- a/features/guidance/roundabout.feature
+++ b/features/guidance/roundabout.feature
@@ -222,6 +222,43 @@ Feature: Basic Roundabout
            | j,f       | jkl,def,def | depart,roundabout-exit-2,arrive |
            | j,c       | jkl,abc,abc | depart,roundabout-exit-3,arrive |
 
+     Scenario: Mixed Entry and Exit - clockwise order
+        Given the node map
+           """
+             c   a
+           j   b   f
+             k   e
+           l   h   d
+             g   i
+           """
+
+        And the ways
+           | nodes | junction   | oneway |
+           | abc   |            | yes    |
+           | def   |            | yes    |
+           | ghi   |            | yes    |
+           | jkl   |            | yes    |
+           | behkb | roundabout | yes    |
+
+        When I route I should get
+           | waypoints | route       | turns                           |
+           | a,c       | abc,abc,abc | depart,roundabout-exit-4,arrive |
+           | a,l       | abc,jkl,jkl | depart,roundabout-exit-3,arrive |
+           | a,i       | abc,ghi,ghi | depart,roundabout-exit-2,arrive |
+           | a,f       | abc,def,def | depart,roundabout-exit-1,arrive |
+           | d,f       | def,def,def | depart,roundabout-exit-4,arrive |
+           | d,c       | def,abc,abc | depart,roundabout-exit-3,arrive |
+           | d,l       | def,jkl,jkl | depart,roundabout-exit-2,arrive |
+           | d,i       | def,ghi,ghi | depart,roundabout-exit-1,arrive |
+           | g,i       | ghi,ghi,ghi | depart,roundabout-exit-4,arrive |
+           | g,f       | ghi,def,def | depart,roundabout-exit-3,arrive |
+           | g,c       | ghi,abc,abc | depart,roundabout-exit-2,arrive |
+           | g,l       | ghi,jkl,jkl | depart,roundabout-exit-1,arrive |
+           | j,l       | jkl,jkl,jkl | depart,roundabout-exit-4,arrive |
+           | j,i       | jkl,ghi,ghi | depart,roundabout-exit-3,arrive |
+           | j,f       | jkl,def,def | depart,roundabout-exit-2,arrive |
+           | j,c       | jkl,abc,abc | depart,roundabout-exit-1,arrive |
+
     Scenario: Mixed Entry and Exit - segregated roads, different names
         Given the node map
            """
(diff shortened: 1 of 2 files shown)


features/guidance/roundabout.feature
```diff
@@ -707,3 +744,27 @@ Feature: Basic Roundabout
...
+
+        And the ways
+           | nodes | junction   | oneway |
+           | abcda | roundabout | yes    |
+           | edf   |            |        |
+           | gch   |            | yes    |
+
+        When I route I should get
+           | waypoints | bearings | route       | turns                           |
+           | e,f       | 90 90    | edf,edf,edf | depart,roundabout-exit-1,arrive |
```
MoKob: This is not necessarily related to the PR at hand, but I am unsure whether we wan't to count this as an exit. Do you see a possibility that this could be changed to a `TurnType::Suppressed`? It would require an adjustment to the roundabout handler.
Probably a topic for a different PR, though.

src/extractor/guidance/roundabout_handler.cpp
```diff
@@ -111,32 +111,57 @@ void RoundaboutHandler::invalidateExitAgainstDirection(const NodeID from_nid,
...
-    for (std::size_t cnt = 0, idx = lhs ? intersection.size() - 1 : 0; cnt < intersection.size();
-         ++cnt, idx += step)
+    std::cout << "invalidateExitAgainstDirection\n";
+
+    // Find range in which exits that must be invalidated (shaded areas):
+    // exit..end   exit..end  begin..exit for ↺ roundabouts
+    // ⭦           ⭦  ⭩         ⭦
+    //  ⭡⭧          ⭡⭩ ▒       ▒⭨⭡
+    //  ⭡⭦          ⭡⭨▒▒       ▒▒⭡⭨
+    // ⭧▒▒⭦        ⭧▒▒▒▒      ▒▒⭧
```
MoKob: Nice!

features/guidance/roundabout.feature
oxidase: I don't think that the instruction should be suppressed, but with the PR it will be announced as "Enter the roundabout and take the 1st exit onto" that is not correct in my opinion and must be announced as a usual intersection, but it will require fix for counting.  Sure it is a topic for another PR after fixing #3979. I will add a note to check the test in a fix for #3979

MoKob: Sure. What kind of instruction to announce can be discussed. In my option it should end up as something like `cross through the roundabout` or simply nothing, if the turn is going straight. But that kind of depends on the scenario.

src/extractor/guidance/roundabout_handler.cpp
TheMarex: It would be great if we could stay on ASCII based characters since in some browsers/OS this diff looks like this:

screenshot from 2017-05-02 09-37-17

src/extractor/guidance/roundabout_handler.cpp
```diff
@@ -111,32 +111,57 @@ void RoundaboutHandler::invalidateExitAgainstDirection(const NodeID from_nid,
     if (in_edge_data.roundabout || in_edge_data.circular)
         return;
 
-    bool past_roundabout_angle = false;
-    const bool lhs = profile_properties.left_hand_driving;
-    const int step = lhs ? -1 : 1;
-    for (std::size_t cnt = 0, idx = lhs ? intersection.size() - 1 : 0; cnt < intersection.size();
-         ++cnt, idx += step)
+    std::cout << "invalidateExitAgainstDirection\n";
```
TheMarex: Leftover debugging code.

src/extractor/guidance/roundabout_handler.cpp
oxidase: It looks like a font issue, because boxes are supported by the font. Anyway i changed to the markdeep style, so rendered 
image

daniel-j-h:
All good here, Travis failing is because of spurious connection issues
```
CMake Error at cmake/mason.cmake:80 (message):
  [Mason] Failed to download
  https://mason-binaries.s3.amazonaws.com/linux-x86_64/binutils/2.27.tar.gz:
  curl: (56) SSL read: error:00000000:lib(0):func(0):reason(0), errno 104
```

Issue #3978, opened 2017-04-26
Problem building osrm-backend with osrm-node on Debian Jessie


angelocavallet:
Hello, first sorry for my awful English, I'm Brazilian.

Since I read that running osrm-routed on production environment is not recommended, I have been trying to install osrm module for nodejs with the new node-osrm merge into osrm-backend but I'm getting some problems with building this on Debian Jessie

Linux  3.16.0-4-amd64 # 1 SMP Debian 3.16.36-1+deb8u2 (2016-10-19) x86_64 GNU/Linux

g++ (Debian 4.9.2-10) 4.9.2
gcc (Debian 4.9.2-10) 4.9.2

./usr/lib/x86_64-linux-gnu/libstdc++.so.5.0.7
./usr/lib/x86_64-linux-gnu/libstdc++.so.6
./usr/lib/x86_64-linux-gnu/libstdc++.so.5
./usr/lib/x86_64-linux-gnu/libstdc++.so.6.0.20

So, I have to update cmake from default Debian Jessie 3.0.2 to 3.1+ (more specifically 3.7.2) because the building error messages told me to do that.

After update cmake, I tried to build osrm-backend again.

```
mkdir -p build
cd build
cmake .. -DCMAKE_BUILD_TYPE=Release -DENABLE_NODE_BINDINGS=On -DENABLE_MASON=On #this is recommended on CHANGELOG to build node-osrm with osrm-backend
```
which output the following text:

```
root@srvopenstreetmap:/home/osm/osrm/osrm-backend/build# cmake .. -DCMAKE_BUILD_TYPE=Release -DENABLE_NODE_BINDINGS=On -DENABLE_MASON=On
-- Enabling mason
-- The C compiler identification is GNU 4.9.2
-- The CXX compiler identification is GNU 4.9.2
-- Check for working C compiler: /usr/bin/cc
-- Check for working C compiler: /usr/bin/cc -- works
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Detecting C compile features
-- Detecting C compile features - done
-- Check for working CXX compiler: /usr/bin/c++
-- Check for working CXX compiler: /usr/bin/c++ -- works
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Detecting CXX compile features
-- Detecting CXX compile features - done
-- Building on a 64 bit system
-- Using GNU gold as linker.
-- Configuring OSRM in release mode
[...]

PR #3984, opened 2017-04-28
Lets loader resolve symbols at startup

daniel-j-h:
At the moment the loader will lazily resolve symbols at runtime when the
first function call happens. This can lead to situations where the libs
are not compatible but we are still able to run - until unresolveable
symbols are encountered.

This changeset tells the loader to resolve all symbols at startup,
failing hard immediately if symbols are unresolveable.

> now
>
> When generating an executable or shared library, mark it to tell the
> dynamic linker to resolve all symbols when the program is started,
> or when the shared library is linked to using dlopen, instead of
> deferring function call resolution to the point when the function is
> first called.

https://linux.die.net/man/1/ld

See https://github.com/Project-OSRM/osrm-backend/issues/3978

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -359,7 +359,7 @@ if("${LINKER_VERSION}" MATCHES "GNU gold" OR "${LINKER_VERSION}" MATCHES "GNU ld
     set(LINKER_FLAGS "${LINKER_FLAGS} -Wl,--gc-sections")
   endif()
   # Default linker optimization flags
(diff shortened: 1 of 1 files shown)


daniel-j-h:
Hmm but these env vars have to be set by the user - which we can not control.

We can control the `-z,now` linker option, though.

oxidase:
Is there a possibility to run `after_install` stage? Otherwise i don't understand the use case. Wrong ABI version in general should not appear for native and cross-compiled builds.

daniel-j-h:
Capturing from chat - this should really happen via npm after `npm install osrm` fetched the binary artifacts and before the user runs the binaries. I'm closing here.

PR #3986, opened 2017-04-28
Make gitsha available in docker image

freenerd:
# Issue

It would be great to retain the gitsha so that when you get a `latest` image from dockerhub you know exactly which code as used.

While working on this I also changed when `COPY . /src` is executed, which makes local development with docker easier, since OS packages remain cached.

## Tasklist

- [ ] Make sure this docker file actually works locally
 - [ ] 👀 @danpat 
 - [ ] merge

/cc @daniel-j-h

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -27,6 +27,8 @@ RUN NPROC=$(grep -c ^processor /proc/cpuinfo 2>/dev/null || 1) && \
     \
     echo "Building OSRM ${DOCKER_TAG}" && \
     cd /src && \
+    git show --format="%H" | head -n1 > /opt/OSRM_GITSHA && \
+    echo "Building OSRM gitsha ${cat /opt/OSRM_GITSHA}" && \
     mkdir build && \
     cd build && \
     cmake -DCMAKE_BUILD_TYPE=${BUILD_TYPE} -DENABLE_LTO=On .. && \


docker/Dockerfile
```diff
@@ -2,39 +2,43 @@ FROM alpine:3.5
...
+
+RUN echo "@testing http://dl-cdn.alpinelinux.org/alpine/edge/testing" >> /etc/apk/repositories && \
     apk update && \
     apk upgrade && \
-    apk add git cmake wget make libc-dev gcc g++ bzip2-dev boost-dev zlib-dev expat-dev lua5.1-dev libtbb@testing libtbb-dev@testing && \
-    \
-    echo "Building libstxxl" && \
+    apk add git cmake wget make libc-dev gcc g++ bzip2-dev boost-dev zlib-dev expat-dev lua5.1-dev libtbb@testing libtbb-dev@testing
+
+RUN echo "Building libstxxl" && \
```
daniel-j-h: This already uses the NPROC var, but you only export it below

freenerd:
The reshuffling of the docker file didn't really work, i broke that out into an own branch https://github.com/Project-OSRM/osrm-backend/compare/reshuffle-dockerfile

TheMarex:
Since this is now only a trivial change, this looks good to merge.

PR #3989, opened 2017-04-28
Fixes util.format calls in Cucumber support code

daniel-j-h:
```
fn(util.format("%s"), 'foo')

=>

fn(util.format("%s", 'foo'))
```


daniel-j-h:
50 years after printf and we have learned nothing ;)

Issue #3987, opened 2017-04-28
Missing turn left instruction


oxidase:
Turn left instruction is missing in http://map.project-osrm.org/?z=17&center=52.833252%2C10.484980&loc=52.832640%2C10.483832&loc=52.835453%2C10.486858&hl=en&alt=0

PR #3990, opened 2017-04-28
Fixes  obvious turn detection no longer taking entry allowed into account

daniel-j-h:
For #3987.

This adds a failing test case for obvious turn at the end of the road.

Note how the skewed angles are required, the test works with:

```
                         d
                         .
            a . b  .  .  c
                    '    .
                      '  e
                         .
                         f
```

Obvious turn detection for end of road scenarios in [TurnHandler](https://github.com/Project-OSRM/osrm-backend/blob/0e39aa9488bff97cffda1d12fb18cdeb537e0c03/src/extractor/guidance/turn_handler.cpp#L226-L228) most likely needs a fix for this.

cc @karenzshea @oxidase

### DIFF
--- a/features/guidance/advanced-lanes.feature
+++ b/features/guidance/advanced-lanes.feature
@@ -35,10 +35,10 @@ Feature: Turn Lane Guidance
             | restriction | bc       | cd     | c        | no_right_turn |
 
        When I route I should get
-            | waypoints | route                | turns                           | lanes                                  |
-            | a,e       | in,cross,cross       | depart,turn left,arrive         | ,left:true straight:false right:false, |
-            | a,g       | in,straight,straight | depart,new name straight,arrive | ,left:false straight:true right:false, |
-            | a,f       | in,cross,cross       | depart,turn right,arrive        | ,left:false straight:false right:true, |
+            | waypoints | route                | turns                           | lanes                      |
+            | a,e       | in,cross,cross       | depart,turn left,arrive         | ,left:true straight:false, |
+            | a,g       | in,straight,straight | depart,new name straight,arrive | ,left:false straight:true, |
+            | a,f       | in,cross,cross       | depart,continue right,arrive    | ,,                         |
 
     @sliproads
     Scenario: Separate Turn Lanes
--- a/features/guidance/collapse.feature
+++ b/features/guidance/collapse.feature
@@ -359,7 +359,7 @@ Feature: Collapse
             | a,g       | first,second,second | depart,turn left,arrive        | a,b,g     |
             | d,g       | first,second,second | depart,turn right,arrive       | d,e,g     |
             | g,f       | second,first,first  | depart,turn right,arrive       | g,e,f     |
-            | g,c       | second,first,first  | depart,end of road left,arrive | g,b,c     |
+            | g,c       | second,first,first  | depart,turn left,arrive        | g,e,c     |
 
     Scenario: Do not collapse turning roads
         Given the node map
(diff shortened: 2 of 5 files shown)


daniel-j-h:
I just pushed a commit fixing the behavior - it has some side-effects on tests some of which I can't explain or can only explain as the tests were modelling broken behavior before.

Please have a careful look @MoKob.

features/guidance/collapse.feature
```diff
@@ -359,7 +359,7 @@ Feature: Collapse
             | a,g       | first,second,second | depart,turn left,arrive        | a,b,g     |
             | d,g       | first,second,second | depart,turn right,arrive       | d,e,g     |
             | g,f       | second,first,first  | depart,turn right,arrive       | g,e,f     |
-            | g,c       | second,first,first  | depart,end of road left,arrive | g,b,c     |
```
MoKob: this looks like a collapse scenario not triggering anymore. the turn here should stay at `e`.

features/guidance/advanced-lanes.feature
```diff
@@ -35,10 +35,10 @@ Feature: Turn Lane Guidance
...
 
        When I route I should get
-            | waypoints | route                | turns                           | lanes                                  |
-            | a,e       | in,cross,cross       | depart,turn left,arrive         | ,left:true straight:false right:false, |
-            | a,g       | in,straight,straight | depart,new name straight,arrive | ,left:false straight:true right:false, |
-            | a,f       | in,cross,cross       | depart,turn right,arrive        | ,left:false straight:false right:true, |
+            | waypoints | route                | turns                           | lanes                      |
+            | a,e       | in,cross,cross       | depart,turn left,arrive         | ,left:true straight:false, |
+            | a,g       | in,straight,straight | depart,new name straight,arrive | ,left:false straight:true, |
+            | a,f       | in,cross,cross       | depart,continue right,arrive    | ,,                         |
```
MoKob: this looks like a bad interaction with the sliproad handler, due to the missing lanes (the combination we see here is only happening on sliproads)

features/guidance/collapse.feature
```diff
@@ -690,9 +690,9 @@ Feature: Collapse
             | restriction | bc       | dc     | c        | no_right_turn |
 
         When I route I should get
-            | waypoints | route            | turns                   | locations |
-            | a,g       | road,cross,cross | depart,turn left,arrive | a,b,g     |
-            | a,e       | road,road        | depart,arrive           | a,e       |
+            | waypoints | route                 | turns                                        | locations |
+            | a,g       | road,road,cross,cross | depart,continue slight left,turn left,arrive | a,b,c,g   |
```
MoKob: again, probably a bad interaction with collapse. It depends on which scenario is relevant here.

MoKob: Turns out the changes visible here no longer are a problem. Seems to have been fixed upstream.

include/extractor/guidance/intersection_handler.hpp
```diff
@@ -410,14 +410,11 @@ std::size_t IntersectionHandler::findObviousTurn(const EdgeID via_edge,
...
             const auto adjusted_distinction_ratio = [&]() {
-                // not allowed competitors are easily distinct
-                if (!intersection[index].entry_allowed)
-                    return 0.7 * DISTINCTION_RATIO;
-                // a bit less obvious are road classes
-                else if (in_way_data.road_classification == best_option_data.road_classification &&
-                         best_option_data.road_classification.GetPriority() <
-                             node_based_graph.GetEdgeData(intersection[index].eid)
-                                 .road_classification.GetPriority())
+                // obviousness by road classes
```
MoKob: Did you print out intersections where this was used? Seems fine in the test-suite, but I am a bit scared of unwanted side-effects. Could be a chance to utilise debug capabilities on a local network. Otherwise I think that this PR is fine. Nice work on the tests, btw.

daniel-j-h: I checked at the location where the problem occured - but didn't go through searching for unwanted side-effects

MoKob: Well. Feeling lucky? :trollface: 

I'd say it should probably work out. The effects on the test-suite are checked, so we should be fine. We might have to revisit if stuff comes up.

PR #3992, opened 2017-05-02
Fixes Table not checking for valid phantom nodes

daniel-j-h:
We failed to check if we could actually find phantom nodes for all
coordinates in the table plugin, leading to corrupt internal state.

```
curl 'http://localhost:5000/table/v1/car/7.4151,43.7305;7.4222,43.7368?radiuses=0;'
```

```
[assert][140505627227904] /tmp/osrm-backend/include/engine/routing_algorithms/routing_base.hpp:68
in: void osrm::engine::routing_algorithms::insertNodesInHeap(osrm::engine::SearchEngineData<osrm::engine::routing_algorithms::ch::Algorithm>::ManyToManyQueryHeap&, const osrm::engine::PhantomNode&) [with bool DIRECTION = false; osrm::engine::SearchEngineData<osrm::engine::routing_algorithms::ch::Algorithm>::ManyToManyQueryHeap = osrm::util::BinaryHeap<unsigned int, unsigned int, int, osrm::engine::ManyToManyHeapData, osrm::util::UnorderedMapStorage<unsigned int, int> >]: phantom_node.IsValid()
```

### DIFF
--- a/src/engine/plugins/table.cpp
+++ b/src/engine/plugins/table.cpp
@@ -67,7 +67,17 @@ Status TablePlugin::HandleRequest(const datafacade::ContiguousInternalMemoryData
         return Error("TooBig", "Too many table coordinates", result);
     }
 
-    auto snapped_phantoms = SnapPhantomNodes(GetPhantomNodes(facade, params));
+    auto phantom_nodes = GetPhantomNodes(facade, params);
+
+    if (phantom_nodes.size() != params.coordinates.size())
+    {
+        return Error("NoSegment",
+                     std::string("Could not find a matching segment for coordinate ") +
+                         std::to_string(phantom_nodes.size()),
+                     result);
+    }
+
+    auto snapped_phantoms = SnapPhantomNodes(phantom_nodes);
     auto result_table =
         algorithms.ManyToManySearch(snapped_phantoms, params.sources, params.destinations);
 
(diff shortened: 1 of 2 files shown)


src/engine/plugins/table.cpp
```diff
@@ -67,7 +67,14 @@ Status TablePlugin::HandleRequest(const datafacade::ContiguousInternalMemoryData
         return Error("TooBig", "Too many table coordinates", result);
     }
 
-    auto snapped_phantoms = SnapPhantomNodes(GetPhantomNodes(facade, params));
+    auto phantom_nodes = GetPhantomNodes(facade, params);
+
+    if (phantom_nodes.size() != params.coordinates.size())
```
TheMarex: You can use the size of `params.coordinate` to sprint a more helpful error message (e.g. "Could not find a matching segment for the i-th coordinate").

daniel-j-h: Good idea, done

daniel-j-h:
Hm Cucumber does not handle situations where `code != Ok` - ticketed in https://github.com/Project-OSRM/osrm-backend/issues/3993. For now I just added a unit test explicitly checking for `code == NoSegment`.

What we probably want is being able to set expected error codes, as in

```cucumber
Scenario: Testbot - Travel time matrix with radiuses option, unable to snap
    Given the node map
        """
        1 a b
        """

    And the ways
        | nodes |
        | ab    |

    And the query options
        | radiuses | 0; |

    And the response code should be "NoSegment"
```

daniel-j-h:
Cherry picked into 5.7.

PR #3994, opened 2017-05-02
Move geometry ids, name_ids and travel_modes to EdgeBasedNodeData

oxidase:
# Issue

Added `.nodes_data` as a temporary solution until  #3446 in which `EdgeBasedNode` data can be split between R-tree and `EdgeBasedNodesData` parts, so the file can be renamed to `.nodes`

So far reduction is not so good: `germany-latest.osrm.edges` 415239224->285476973=-123.8 MiB and the new file `germany-latest.osrm.nodes_data` has 131555480=125.5 MiB. So data size increased by 1.7 MiB because of an additional NodeID vector in  `.edges` file.

What can be improved 
* with #3952 `travel_modes` size can be reduced by factor 2
* check if forward/reverse edges are duplicated in `.nodes_data` file

/cc @TheMarex @danpat 

## Tasklist
 - [x] review
 - [x] adjust for comments

## Requirements / Relations
#3952


src/extractor/edge_based_graph_factory.cpp
```diff
@@ -291,8 +297,47 @@ std::vector<NBGToEBG> EdgeBasedGraphFactory::GenerateEdgeExpandedNodes()
...
+                m_compressed_edge_container.HasZippedEntryForReverseID(edge_id);
+            if (is_encoded_forwards || is_encoded_backwards)
+            {
+                auto geometry_id =
+                    is_encoded_forwards
+                        ? m_compressed_edge_container.GetZippedPositionForForwardID(edge_id)
+                        : m_compressed_edge_container.GetZippedPositionForReverseID(edge_id);
+
+                BOOST_ASSERT(index.find(edge_id) == index.end());
+                index.insert({edge_id, ebg_node_id++});
```
TheMarex: Instead of introducing an own numbering, we can use the (future) `NodeID` from the search graph here. These IDs are computed [here](https://github.com/Project-OSRM/osrm-backend/blob/master/src/extractor/edge_based_graph_factory.cpp#L206-L236). They can be accessed using `GetEdgeData(edge_id).edge_id`.

src/extractor/edge_based_graph_factory.cpp
```diff
@@ -452,39 +497,15 @@ void EdgeBasedGraphFactory::GenerateEdgeExpandedEdges(
...
-                            turn.lane_data_id,
-                            entry_class_id,
-                            edge_data1.travel_mode,
-                            util::guidance::TurnBearing(intersection[0].bearing),
-                            util::guidance::TurnBearing(turn.bearing));
-                    }
+                    BOOST_ASSERT(nbg_edgeid_to_ebg_nodeid.find(incoming_edge) !=
+                                 nbg_edgeid_to_ebg_nodeid.end());
+                    turn_data_container.push_back(
+                        nbg_edgeid_to_ebg_nodeid.find(incoming_edge)->second,
```
TheMarex: This data is unnecessary if we use the search graph node IDs here.

src/extractor/edge_based_graph_factory.cpp
oxidase: Thanks for hint! I have completely missed https://github.com/Project-OSRM/osrm-backend/blob/master/src/extractor/edge_based_graph_factory.cpp#L228

oxidase:
@TheMarex thanks for help! I have removed the index from TurnData. [One place](https://github.com/Project-OSRM/osrm-backend/blob/a2fac1bb01e35a9f64231152a3d8b6f06d11d9f9/src/extractor/edge_based_graph_factory.cpp#L292-L329) needs a refactoring together with #3446

oxidase:
There is a new temporary file `.nodes_data` that must be renamed to `.nodes` in the next refactoring stage. 
Data sizes for `germany-latest.osm.pbf`
```
-rw-r--r-- 1 miha miha  181667289 May  3 17:51 germany-latest.osrm.edges
-rw-r--r-- 1 miha miha  126112703 May  3 17:42 germany-latest.osrm.nodes_data
```

So reduction is from 415239224 (396 MiB)  to 181667289+126112703 (293MiB) or 26%

include/engine/routing_algorithms/routing_base_mld.hpp
```diff
@@ -326,27 +331,18 @@ inline void search(SearchEngineData<Algorithm> &engine_working_data,
...
-    if (weight != INVALID_EDGE_WEIGHT)
-    {
-        packed_leg.push_back(source_node);
-        std::transform(unpacked_edges.begin(),
-                       unpacked_edges.end(),
-                       std::back_inserter(packed_leg),
-                       [&facade](const auto edge) { return facade.GetTarget(edge); });
-    }
+    // TODO: change search calling interface to use unpacked_edges result
+    std::tie(weight, packed_leg, std::ignore) = mld::search(engine_working_data,
```
TheMarex: The name `packed_leg` is actually wrong here, right? The search returns `unpacked_nodes` at this position.

include/engine/routing_algorithms/routing_base_mld.hpp
```diff
@@ -326,27 +331,18 @@ inline void search(SearchEngineData<Algorithm> &engine_working_data,
...
+                                                            facade,
+                                                            forward_heap,
+                                                            reverse_heap,
+                                                            force_loop_forward,
+                                                            force_loop_reverse,
+                                                            weight_upper_bound,
+                                                            phantom_nodes);
 }
 
+// TODO: remove CH-related stub
```
TheMarex: For the sake of the abstraction: Could we make `search()` return an unpacked path, and split out the actual unpacking logic to this function? (not part of this PR, just noticed this)

src/extractor/edge_based_graph_factory.cpp
```diff
@@ -208,6 +208,9 @@ void EdgeBasedGraphFactory::Run(ScriptingEnvironment &scripting_environment,
 /// Returns the number of edge based nodes.
 unsigned EdgeBasedGraphFactory::RenumberEdges()
 {
+    // heuristic: node-based graph node is a simple intersection with four edges (edge-based nodes)
+    m_edge_based_node_weights.reserve(4 * m_node_based_graph->GetNumberOfNodes());
```
TheMarex: `m_node_based_graph->GetNumberOfEdges()` could be used instead too.

TheMarex:
Hrm there seems to be a failure on 32bit https://travis-ci.org/Project-OSRM/osrm-backend/jobs/228378349#L2287 not sure why it works in the other jobs.

include/engine/routing_algorithms/routing_base_mld.hpp
oxidase: yes, MLD already returns an unpacked path. CH algorithms previously used an in-place lambda  callback in the recursive unpacking. It was removed because of quite heavy explicit type specification of the callback function. Now as both algorithms return unpacked nodes and edges for `annotatePath` it can cleaned up.

src/extractor/edge_based_graph_factory.cpp
oxidase: just checked: after edges zipping  `GetNumberOfEdges` overestimates the  number valid edges, so i would keep 4 * #nodes to avoid  regression in memory consumption

PR #3995, opened 2017-05-02
Use rectified linear unit to prevent negative duration values

oxidase:
# Issue

In quite special cases duration values can be negative due to different round-offs in weight and duration values. In the [test case](https://github.com/Project-OSRM/osrm-backend/blob/0d3da8628916efae7dd9ffb439b8616d95d1a4b2/features/testbot/projection.feature#L68) for points 2 and 4 [phantom snapping](https://github.com/Project-OSRM/osrm-backend/blob/9334cc463d4e8b42101d96680b6f416c4fc5d27c/include/engine/geospatial_query.hpp#L408-L418) has ratio values 0.325 and 0.315, so forward weights are 0.325 * 55 = 17.875 and 0.315 * 55 = 17.325, but forward durations  0.325 * 28 = 9.1 and 0.315 * 28 = 8.82. The weight difference is 0, this leads to an empty path with phantom nodes in the incorrect order, so the computes duration is 8-9=-1 deci-second.

Clipping durations at 0 is not really a solution, but without the clipping negative values for zero weight paths must accepted as zero durations. 

## Tasklist
 - [ ] review
 - [ ] adjust for comments

## Requirements / Relations


oxidase:
@TheMarex added an assertion `target_duration >= source_duration || weight == 0`, but think OSRM should not return negative durations even if it may hide other errors.

TheMarex:
@oxidase agreed. Ready to merge from my side.

PR #3996, opened 2017-05-02
Replace PackedVector with an implementation that supports random access

### LINKED ISSUE #3952: Add random access to PackedVector

TheMarex:
# Issue

This fixes #3952 and unlocks using it in `SegmentDataContainer`.

The resulting vector type has the following performance characteristics:

```
[info] random write: std::vector 37537.3 ms, util::packed_vector 82940.7 ms. 2.20955
[info] random read: std::vector 55340.9 ms, util::packed_vector 70225.4 ms. 1.26896
```

20% overhead for reads and 120% overhead for writes, given the intended usage I think that should be fine. This needs a benchmark after the integration.

## Tasklist
 - [x] Add benchmark to measure against plain `std::vector<>`
 - [x] review
 - [x] adjust for comments

## Requirements / Relations

### DIFF
--- a/include/util/packed_vector.hpp
+++ b/include/util/packed_vector.hpp
@@ -1,12 +1,17 @@
 #ifndef PACKED_VECTOR_HPP
 #define PACKED_VECTOR_HPP
 
+#include "util/integer_range.hpp"
 #include "util/typedefs.hpp"
 #include "util/vector_view.hpp"
 
 #include "storage/io_fwd.hpp"
 #include "storage/shared_memory_ownership.hpp"
 
+#include <boost/iterator/iterator_facade.hpp>
+#include <boost/iterator/reverse_iterator.hpp>
+
+#include <array>
 #include <cmath>
 #include <vector>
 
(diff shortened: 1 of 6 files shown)


TheMarex:
Currently seeing errors like:

```
/usr/include/boost/iterator/iterator_facade.hpp: In instantiation of ‘typename boost::iterators::detail::enable_if_interoperable<Derived1, Derived2, typename boost::mpl::apply2<boost::iterators::detail::always_bool2, Derived1, Derived2>::type>::type boost::iterators::operator!=(const boost::iterators::iterator_facade<Derived1, V1, TC1, Reference1, Difference1>&, const boost::iterators::iterator_facade<Derived2, V2, TC2, Reference2, Difference2>&) [with Derived1 = osrm::util::detail::PackedVector<unsigned int, 33ul, (osrm::storage::Ownership)0>; V1 = unsigned int; TC1 = boost::iterators::random_access_traversal_tag; Reference1 = osrm::util::detail::PackedVector<unsigned int, 33ul, (osrm::storage::Ownership)0>::reference; Difference1 = long int; Derived2 = osrm::util::detail::PackedVector<unsigned int, 33ul, (osrm::storage::Ownership)0>; V2 = unsigned int; TC2 = boost::iterators::random_access_traversal_tag; Reference2 = osrm::util::detail::PackedVector<unsigned int, 33ul, (osrm::storage::Ownership)0>::reference; Difference2 = long int; typename boost::iterators::detail::enable_if_interoperable<Derived1, Derived2, typename boost::mpl::apply2<boost::iterators::detail::always_bool2, Derived1, Derived2>::type>::type = bool]’:
/home/patrick/Code/osrm-backend/unit_tests/util/packed_vector.cpp:74:23:   required from here
/usr/include/boost/iterator/iterator_facade.hpp:900:3: error: invalid static_cast from type ‘const boost::iterators::iterator_facade<osrm::util::detail::PackedVector<unsigned int, 33ul, (osrm::storage::Ownership)0>, unsigned int, boost::iterators::random_access_traversal_tag, osrm::util::detail::PackedVector<unsigned int, 33ul, (osrm::storage::Ownership)0>::reference, long int>*’ to type ‘const osrm::util::detail::PackedVector<unsigned int, 33ul, (osrm::storage::Ownership)0>*’
   BOOST_ITERATOR_FACADE_RELATION(!=, return !, equal)
   ^
/usr/include/boost/iterator/iterator_facade.hpp:900:3: error: invalid static_cast from type ‘const boost::iterators::iterator_facade<osrm::util::detail::PackedVector<unsigned int, 33ul, (osrm::storage::Ownership)0>, unsigned int, boost::iterators::random_access_traversal_tag, osrm::util::detail::PackedVector<unsigned int, 33ul, (osrm::storage::Ownership)0>::reference, long int>*’ to type ‘const osrm::util::detail::PackedVector<unsigned int, 33ul, (osrm::storage::Ownership)0>*’
   BOOST_ITERATOR_FACADE_RELATION(!=, return !, equal)
```

While compiling. Seems like the boost iterator facade is not happy. @oxidase have you maybe seen this error before?

oxidase:
@TheMarex  problem is in the `Derived` parameter 
```
    class iterator_impl
        : public boost::iterator_facade<iterator_impl<DataT>, DataT, boost::random_access_traversal_tag, reference>
    {
        typedef boost::iterator_facade<iterator_impl<DataT>, DataT, boost::random_access_traversal_tag> base_t;
```
but some other problems remain

oxidase:
@TheMarex i have pushed compilation fixes, but `packed_vector_iterator_test` has a malloc fail starting from `PackedVector<std::uint32_t, 33> packed_vec(70);`

include/util/packed_vector.hpp
```diff
@@ -31,188 +35,353 @@ inline void write(storage::io::FileWriter &writer,
...
+template <typename WordT, typename T>
+inline WordT set_lower_value(WordT word, WordT mask, std::uint8_t offset, T value)
+{
+    return (word & ~mask) | (static_cast<WordT>(value) << offset);
+}
+
+template <typename WordT, typename T>
+inline WordT set_upper_value(WordT word, WordT mask, std::uint8_t offset, T value)
+{
+    return (word & ~mask) | (static_cast<WordT>(value) >> offset);
```
oxidase: Here can be a problem due to an arithmetic shift  if `WordT` is signed: cast to the derived unsigned type?

include/util/packed_vector.hpp
```diff
@@ -31,188 +35,353 @@ inline void write(storage::io::FileWriter &writer,
...
-    /**
-     * Returns the size of the packed vector datastructure with `elements` packed elements (the size
-     * of
-     * its underlying uint64 vector)
-     */
-    inline static std::size_t elements_to_blocks(std::size_t elements)
-    {
-        return std::ceil(static_cast<double>(elements) * Bits / ELEMSIZE);
-    }
+    static constexpr std::size_t BLOCK_ELEMENTS = WORD_BITS;
```
oxidase: Here i don't understand why `BLOCK_ELEMENTS` must be `WORD_BITS` that is 8*8?

TheMarex: This is to make the total number of bits in a block devisable by the number of bits in a word. For example if every record has 22bits then 22*64bits => 22 words per block. Technically we could also try to compute the smallest common multiple of the number of bits in a record and the number of bits in a word, but this was simpler.

include/util/packed_vector.hpp
```diff
@@ -31,188 +35,353 @@ inline void write(storage::io::FileWriter &writer,
...
+                                               upper_mask[internal_index.element],
+                                               upper_offset[internal_index.element],
+                                               value);
+    }
+
+    void initialize_mask()
     {
-        return vec.back();
+        // TODO port to constexptr function
+        const std::uint64_t mask = (1UL << Bits) - 1;
```
TheMarex: Does it need to be `1ULL`?

include/util/packed_vector.hpp
```diff
@@ -31,188 +35,353 @@ inline void write(storage::io::FileWriter &writer,
...
+        auto offset = 0;
+        for (auto element_index : util::irange<std::uint8_t>(0, BLOCK_ELEMENTS))
+        {
+            auto local_offset = offset % 64;
```
TheMarex: Yes this should be `BLOCK_ELEMENTS`

include/util/packed_vector.hpp
TheMarex: I'm going to add a static assertion to check that `WordT` is an unsigned type, if that works for you.

oxidase: :+1: for signed types a sign extension must be implemented, don't know if it is needed now

include/util/packed_vector.hpp
oxidase: [coliru does not support -m32](http://coliru.stacked-crooked.com/a/48abfe596ba3c184), but with `1ul` -m32 prints `1023`, `1ull` is ok, but maybe `std::uint64_t{1} << Bits`? so it can be asserted that `Bits < sizeof(std::uint64_t)` to avoid UB

include/util/packed_vector.hpp
oxidase: :+1:  the power of named constants

TheMarex:
@oxidase could you please have a look again if this new version works for you?

TheMarex:
There are some bugs in the iterator code that make them silently fail (`for (auto value : packed_vector)` does nothing for example)

include/util/packed_vector.hpp
```diff
@@ -31,188 +36,458 @@ inline void write(storage::io::FileWriter &writer,
...
-    signed cursor = -1;
+    auto end() const { return const_iterator(this, num_elements); }
 
-    template <bool enabled = (Ownership == storage::Ownership::View)>
-    void replace_last_elem(typename std::enable_if<enabled, std::uint64_t>::type last_elem)
+    auto cbegin() const { return const_iterator(this, 0); }
+
+    auto cend() const { return const_iterator(this, num_elements); }
+
+    auto rbegin() const { return reverse_iterator(end()); }
```
oxidase: const is not needed here and below

TheMarex:
@oxidase thank you for the test case and the review! This needs to be fixed before merging, on it.

TheMarex:
Blocked by failing doc builds.

PR #3988, opened 2017-04-28
Update the node docs

TheMarex:
# Issue

This fixes a few issues:

- https://github.com/Project-OSRM/osrm-backend/issues/3981
- There were updates to the `API.md` that were not reflected in the `node_osrm.cpp` source
- Adds better docs for all parameters of the constructor

## Tasklist
 - [x] review
 - [x] adjust for comments

## Requirements / Relations
 Link any requirements here. Other pull requests this PR is based on?

PR #3997, opened 2017-05-03
Updates Node.js docs: bearing supports `null` as dont-care

daniel-j-h:
We support `bearings=[(val0,rng0), null, (val1,rng1)]`, too.


TheMarex:
Should go in after https://github.com/Project-OSRM/osrm-backend/pull/3988 which I forgot to merge. This changes how we document these options to a per-method level to conform to `documention.js` style.

TheMarex:
This was fixed with #4000.

Issue #3998, opened 2017-05-03
Node.js bindings do not support `unlimited` as advertised


daniel-j-h:
Our Node.js binding docs state

https://github.com/Project-OSRM/osrm-backend/blob/master/docs/nodejs/api.md#general-options

    radiuses | null or double >= 0 or unlimited (default)

but we only support `null` and `Number` [here](https://github.com/Project-OSRM/osrm-backend/blob/master/include/nodejs/node_osrm_support.hpp#L415-L455).

daniel-j-h:
We resolved this.

PR #3999, opened 2017-05-03
Implements `unlimited` as dont-care for radiuses option in Node.js bindings

daniel-j-h:
For #3998. Brings Node.js bindings closer to what we advertise in the v5 API ..

Second commit checking for `node_modules` since without `npm install --ignore-scripts` error messages when compiling the Node.js bindings are atrocious.


TheMarex:
Given that using `null` is already the established use, I would be inclined to just fix the docs and not increase the API surface area. Seem like this was a mistake from porting them over from the HTTP docs.

daniel-j-h:
Yep sounds reasonable 👍

PR #4008, opened 2017-05-04
Add penalties to alleys on the bicycle profile

### LINKED ISSUE #3971: Bike profiles route through alleys in Montreal - this is less than desirable
In Montréal, our bike profile likes to send people down alleys:

screen shot 2017-04-24 at 10 56 09 am

however, at least in Montréal, routing via `highway=service, service=alley` class roads highly undesirable:

http://www.openstreetmap.org/way/21062511

There are two reasons we should consider avoiding alleys for bikes:

  1. The road surface could be a bit iffy
  2. You typically need to cross a through-street at the end of each alley - this is a lot less preferable than traveling with the flow of traffic.

We should consider making two changes:

  1. Slow down travel speeds in alleys - they're currently set to 15km/h, just like all other roads.
  2. Add some entry/exit penalties - either via https://github.com/Project-OSRM/osrm-backend/pull/3721, or https://github.com/Project-OSRM/osrm-backend/pull/2912 / https://github.com/Project-OSRM/osrm-backend/pull/2399

Related old ticket: https://github.com/Project-OSRM/osrm-backend/issues/592

TheMarex:
# Issue

Fixes #3971 by applying penalties to alleys.

## Tasklist
 - [x] review
 - [x] adjust for comments

### DIFF
--- a/features/bicycle/alley.feature
+++ b/features/bicycle/alley.feature
@@ -1,0 +1,35 @@
+@routing @bicycle @alley
+Feature: Bicycle - Route around alleys
+
+    Background:
+        Given the profile file
+        """
+        require 'bicycle'
+        properties.weight_name = 'cyclability'
+        """
+
+    Scenario: Bicycle - Avoid taking alleys
+        Given the query options
+            | annotations | nodes |
+
+        Given the node map
+            """
+            a-----b-----c
+            |     :     |
+            d.....e.....f
+            """
+
+        And the ways
+            | nodes | highway     | service  |
+            | abc   | residential |          |
+            | def   | service     | alley    |
+            | ad    | residential |          |
+            | be    | service     | alley    |
+            | cf    | residential |          |
+
+        When I route I should get
+            | from | to | a:nodes    | weight | #                                |
+            | a    | f  | 1:2:3:6    | 192    | Avoids d,e,f                     |
+            | a    | e  | 1:2:5      | 168    | Take the alley b,e if neccessary |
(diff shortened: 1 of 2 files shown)


features/bicycle/alley.feature
```diff
@@ -0,0 +1,35 @@
...
+            | nodes | highway     | service  |
+            | abc   | residential |          |
+            | def   | service     | alley    |
+            | ad    | residential |          |
+            | be    | service     | alley    |
+            | cf    | residential |          |
+
+        When I route I should get
+            | from | to | a:nodes    | weight | #                                |
+            | a    | f  | 1:2:3:6    | 192    | Avoids d,e,f                     |
```
emiltin: the node ordering here depends on the cucumber support code, and probably should not be used directly in tests? or is the ordering well-defined and documented?

daniel-j-h: @emiltin this is using the `anotations` feature for OSM IDs, not sure what you mean with ordering is dependent on support code? It's the ids along the route, no?

emiltin: the wiki at https://github.com/Project-OSRM/osrm-backend/wiki/Cucumber-Test-Suite#nodes states that ids are assigned starting at 1, in left-to-right, top-to-bottom order. it's that is ensured there is no problem.


|

TheMarex: We changed this to assign IDs based on node name, e.g. `a` will always be `1`. I used ids instead of steps here to circumvent side effects cause by guidance step collapsing.

Issue #3974, opened 2017-04-25
Benchmark boost CSV parsing


danpat:
Working on another internal project, we noticed that using `boost::spirit` to parse CSV data was not super performant.  Ripping out the `boost::spirit` code and replacing it with https://github.com/ben-strasser/fast-cpp-csv-parser gave a huge speedup.

We should re-visit the CSV parsing for traffic updates - `fast-cpp-csv-parser` was about 10x faster for linear parsing.

TheMarex:
We had this originally https://github.com/Project-OSRM/osrm-backend/pull/1764/files and ripped it out because it was slower then just using `boost::spirit`. So one of these benchmarks must have been wrong.

oxidase:
@TheMarex there is a possible performance regression in using `boost::spirit::istream_iterator` that is visible on >1GB files. It should be checked and reverted.

oxidase:
Closed via #4010

PR #4010, opened 2017-05-05
Use CSV mmaped files

oxidase:
# Issue

Benchmarking of #3974 for ~3GB CSV files shows that ~3 minutes loading time was spent in the stream `std::ifstream` with iterators wrapped into `boost::spirit::line_pos_iterator` that are used used by qi as `boost::spirit::multi_pass` iterators. This PR changes `std::ifstream` to `boost::iostreams::mapped_file_source` and reduces loading time for benchmark to ~35-40 seconds with ~10 seconds used for parsing.

/cc @danpat 

## Tasklist
 - [x] review
 - [x] adjust for comments

### DIFF
--- a/include/updater/csv_file_parser.hpp
+++ b/include/updater/csv_file_parser.hpp
@@ -10,6 +10,7 @@
 #include <tbb/parallel_for.h>
 #include <tbb/spin_mutex.h>
 
+#include <boost/exception/diagnostic_information.hpp>
 #include <boost/iostreams/device/mapped_file.hpp>
 #include <boost/spirit/include/phoenix.hpp>
 #include <boost/spirit/include/qi.hpp>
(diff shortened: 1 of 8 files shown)


TheMarex:
Wow awesome, so this brings a speedup of about 5x? Wow. That will seriously cut down our processing times.

oxidase:
@TheMarex yes, it changes `InputIterator` wrapped in into a multi-pass iterator by spirit to `RandomAccessIterator` that makes faster rollbacks.

include/updater/csv_file_parser.hpp
```diff
@@ -10,34 +10,28 @@
...
 // Functor to parse a list of CSV files using "key,value,comment" grammar.
 // Key and Value structures must be a model of Random Access Sequence.
 // Also the Value structure must have source member that will be filled
 // with the corresponding file index in the CSV filenames vector.
 template <typename Key, typename Value> struct CSVFilesParser
 {
-    using Iterator = boost::spirit::line_pos_iterator<boost::spirit::istream_iterator>;
-    using KeyRule = qi::rule<Iterator, Key()>;
-    using ValueRule = qi::rule<Iterator, Value()>;
+    using Iterator = boost::iostreams::mapped_file_source::iterator;
```
TheMarex: So interesting that this little bit of abstraction is so slow.

Issue #4011, opened 2017-05-05
Bearing Constraints not Respected when using Continue Straight


daniel-j-h:
In the following I set bearing constraints of `270,90;;` for three waypoints to force the route's start going west. The response starts off to the east, though:

![bearing](https://cloud.githubusercontent.com/assets/527241/25740437/e4f250aa-3186-11e7-8e9a-597724a405f5.png)

[Map](http://bl.ocks.org/anonymous/raw/58e61915e597e3db2d77d2b3773edf42/)

Request:

    curl 'http://router.project-osrm.org/route/v1/driving/-6.808948516845703,62.0000383831454;-6.802253723144531,62.00414819760298;-6.808948516845703,62.0000383831454?bearings=270,90;;&continue_straight=false&steps=true&geometries=geojson'

Waypoints are:
- (-6.808948516845703,62.0000383831454)
- (-6.802253723144531,62.00414819760298)
- (-6.808948516845703,62.0000383831454)

daniel-j-h:
Looks like `continue_straight=false` is to blame here.
Same request with `continue_straight=true` (the default):

![bearing2](https://cloud.githubusercontent.com/assets/527241/25741286/4e322f6a-318a-11e7-8ed8-d7451a9c7a26.png)

[Map](http://bl.ocks.org/d/0388ff1222489f734268980eb8f52357)

daniel-j-h:
Investigation with @MoKob lead to [this](https://github.com/Project-OSRM/osrm-backend/blob/35d7b7ceaf53863bbd6cb49bccaeef4e171b1985/src/engine/plugins/viaroute.cpp#L78-L102) location where we unconditionally set the enable flag (because `!continue_straight` is `true` and we `|=` it).

Removing the two conditionals reveals the expected behavior:
http://bl.ocks.org/anonymous/raw/be471f5881fc6c5c41966296f46cd83f/

@MoKob has a fix and some test cases at hand shortly.

---

This needs to go in asap - looks like `continue_straight` was kinda broken since it made it in.

PR #4013, opened 2017-05-05
fix continue_straight interaction with bearing specification

MoKob:
# Issue

Resolves https://github.com/Project-OSRM/osrm-backend/issues/4011

It seems that we incorrectly disabled segments via the continue_straight parameter.
Continue_straight is a via-only parameter and can only be evaluated during the query.
From my understanding, we should not disable start/end locations in the viaplugin.

The added tests show the fixed behaviour when using bearings with continue_straight.

## Tasklist
 - [x] review
 - [x] adjust for comments

### DIFF
--- a/features/testbot/via.feature
+++ b/features/testbot/via.feature
@@ -352,3 +352,64 @@ Feature: Via points
             | waypoints | bearings   | route    | turns                    |
             | 1,a       | 90,2 270,2 | ab,ab,ab | depart,turn uturn,arrive |
             | 1,b       | 270,2 90,2 | ab,ab,ab | depart,turn uturn,arrive |
+
+    Scenario: Continue Straight in presence of Bearings
+        Given the node map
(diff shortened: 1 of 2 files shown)


features/testbot/via.feature
```diff
@@ -352,3 +352,64 @@ Feature: Via points
...
+            | ah    | yes    |
+            | bg    | yes    |
+            | cf    | yes    |
+            | di    | yes    |
+            | dj    | yes    |
+
+        When I route I should get
+            | waypoints | bearings               | route                                   |
+            | 1,2,3     | 270,90 180,180 180,180 | ab,cdea,cdea,bc,bc,bc,ab,cdea,cdea,cdea |
+
```
daniel-j-h: 👍

daniel-j-h:
Cherry picked into 5.7.

PR #4019, opened 2017-05-05
Negative and invalid weights after speed updates

oxidase:
# Issue

PR should fix negative weights after traffic updates that some aborts and incorrect results.

Test `Check restricted way 5` fails in following steps:
* [reverse_weight_vector = {2147483647, 1500}](https://github.com/Project-OSRM/osrm-backend/blob/9d1a032f14ec09e8a6e152d056462ed392efed4a/include/engine/geospatial_query.hpp#L379)
* [reverse_heap.inserted_nodes = std::vector of length 1, capacity 1 = {..., node = 1, weight = 2147483647, data = {parent = 1, from_clique_arc = false}}}](https://github.com/Project-OSRM/osrm-backend/blob/9d1a032f14ec09e8a6e152d056462ed392efed4a/src/engine/routing_algorithms/direct_shortest_path.cpp#L130)
* [ reverse_heap.Empty()=false, reverse_heap.MinKey() = 214748364](https://github.com/Project-OSRM/osrm-backend/blob/9d1a032f14ec09e8a6e152d056462ed392efed4a/include/engine/routing_algorithms/routing_base_mld.hpp#L197)

/cc @TheMarex @danpat

## Tasklist
 - [x] decision about how to fix: don't snap edges with segments INVALID_EDGE_WEIGHT or filter snapped nodes
 - [x] fix
 - [ ] review
 - [ ] adjust for comments

## Requirements / Relations

### DIFF
--- a/features/step_definitions/routability.js
+++ b/features/step_definitions/routability.js
@@ -126,6 +126,7 @@ module.exports = function () {
 
                 r.query = this.query;
                 r.json = JSON.parse(body);
+                r.code = r.json.code;
                 r.status = res.statusCode === 200 ? 'x' : null;
                 if (r.status) {
                     r.route = this.wayList(r.json.routes[0]);
--- a/include/engine/phantom_node.hpp
+++ b/include/engine/phantom_node.hpp
@@ -55,9 +55,9 @@ struct PhantomNode
           forward_duration_offset(0), reverse_duration_offset(0),
           packed_geometry_id(SPECIAL_GEOMETRYID), component{INVALID_COMPONENTID, false},
           fwd_segment_position(0), forward_travel_mode(TRAVEL_MODE_INACCESSIBLE),
-          backward_travel_mode(TRAVEL_MODE_INACCESSIBLE), is_forward_valid_source(false),
-          is_forward_valid_target(false), is_reverse_valid_source(false),
-          is_reverse_valid_target(false)
+          backward_travel_mode(TRAVEL_MODE_INACCESSIBLE), is_valid_forward_source(false),
+          is_valid_forward_target(false), is_valid_reverse_source(false),
+          is_valid_reverse_target(false)
     {
     }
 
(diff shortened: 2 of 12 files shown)


oxidase:
From discussion with @TheMarex:
1/ As a quick solution source phantom nodes that contain at least one segment with `INVALID_EDGE_WEIGHT` and target phantom nodes that contain at least one segment from 0 to `fwd_segment_position + 1` with `INVALID_EDGE_WEIGHT` are disabled. This prevents using weight values with integer overflows and in the scope of the pull request. Also now `search` functions  must check  for an empty input heaps and return an empty route if at least one heap is empty.

2/ The fix will return empty routes in the following correct situations (`xxx` is a segment with `INVALID_EDGE_WEIGHT`, `s` is a starting point on a `ab` edge, `t` is a target node):
    a) forward node `a----xxx-----s------b`  because `-as` value is used as weight 
    b) reverse node  `a--------s---xxx---b`  because `-bs` value is used as weight
    c) special case  `a---xxx---s---t---xxx---b`

Fixing the above cases requires refactoring [`insertNodesInHeap`](https://github.com/Project-OSRM/osrm-backend/blob/0e81d749ff66b400d079d9bbde2a7d88b46fbd9b/include/engine/routing_algorithms/routing_base.hpp#L73-L86) to avoid using restricted segments. 
Fixing case c) requires refactoring a special case handling in [`assembleSteps`](https://github.com/Project-OSRM/osrm-backend/blob/0e81d749ff66b400d079d9bbde2a7d88b46fbd9b/include/engine/guidance/assemble_steps.hpp#L214-L226).

features/step_definitions/routability.js
```diff
@@ -126,6 +126,7 @@ module.exports = function () {
 
                 r.query = this.query;
                 r.json = JSON.parse(body);
+                r.code = r.json.code;
```
TheMarex: Are these changes from another commit that is already in master>

oxidase: The changes are in separate commit, but must be in previous PR:  just missed `trip` and `routability`  `code` values

include/engine/phantom_node.hpp
```diff
@@ -134,6 +105,23 @@ struct PhantomNode
 
     bool IsValid() const { return location.IsValid() && (name_id != INVALID_NAMEID); }
 
+    bool IsForwardValidSource() const
```
TheMarex: The naming is a little bit weird. In English `IsValidForwardSource` would make more sense, same for the other functions.

include/engine/routing_algorithms/routing_base.hpp
```diff
@@ -41,41 +41,23 @@ bool needsLoopForward(const PhantomNode &source_phantom, const PhantomNode &targ
...
-
 template <bool DIRECTION>
 void insertNodesInHeap(SearchEngineData<ch::Algorithm>::ManyToManyQueryHeap &heap,
                        const PhantomNode &phantom_node)
 {
     BOOST_ASSERT(phantom_node.IsValid());
 
     const auto weight_sign = DIRECTION == FORWARD_DIRECTION ? -1 : 1;
-    if (phantom_node.forward_segment_id.enabled)
+    if ((DIRECTION == FORWARD_DIRECTION && phantom_node.IsForwardValidSource()) ||
```
TheMarex: `DIRECTION` is a little bit of a mis-normer here. In the `ManyToMany` plugin we insert each source and target in forward and backward direction.

It would make more sense to have two functions `insertSourceInHeap` and `insertTargetInHeap`.

src/engine/routing_algorithms/routing_base_ch.cpp
```diff
@@ -155,31 +161,7 @@ double getNetworkDistance(SearchEngineData<Algorithm> &engine_working_data,
...
-                            target_phantom.GetForwardWeightPlusOffset(),
-                            target_phantom.forward_segment_id.id);
-    }
-    if (target_phantom.reverse_segment_id.enabled)
-    {
-        reverse_heap.Insert(target_phantom.reverse_segment_id.id,
-                            target_phantom.GetReverseWeightPlusOffset(),
-                            target_phantom.reverse_segment_id.id);
-    }
+    insertNodesInHeaps(forward_heap, reverse_heap, {source_phantom, target_phantom});
```
TheMarex: Good catch! 👍

PR #4022, opened 2017-05-09
Add response code  to result values

### LINKED ISSUE #3993: Cucumber tests should check and handle response error codes
At the moment Cucumber never checks for the router's `code` and `message` response property and assumes routing just worked fine (`code=Ok`). What we instead should do is:
- make sure `code == 'Ok'` and error out if not
- handle `code != 'Ok'` and let the user provide expected error codes

For example, having a Cucumber phrase: `And the response code should be "NoSegment"`.

---

Currently the support code just crashes hard because [we assume we always get a result](https://github.com/Project-OSRM/osrm-backend/blob/2021c30805258cbab4fcec4ff1249c1bc213f1f5/features/step_definitions/distance_matrix.js#L48), e.g. for table:

```
    TypeError: Cannot read property 'map' of undefined
         at /tmp/osrm-backend/features/step_definitions/distance_matrix.js:51:47
         at /tmp/osrm-backend/features/support/http.js:50:20
         at Object.<anonymous> (/tmp/osrm-backend/node_modules/node-timeout/lib/timeout.js:36:17)
         at Request._callback (/tmp/osrm-backend/features/support/http.js:39:24)
         at Request.self.callback (/tmp/osrm-backend/node_modules/request/request.js:188:22)
         at emitTwo (events.js:87:13)
         at Request.emit (events.js:172:7)
         at Request.<anonymous> (/tmp/osrm-backend/node_modules/request/request.js:1171:10)
         at emitOne (events.js:77:13)
         at Request.emit (events.js:169:7)
```

oxidase:
# Issue

PR adds `json.code` to result values that can be checked as 
```
        When I route I should get
          | from | to | bearings | code    |
          |    1 |  2 | 270 270  | NoRoute |
```

Fixes  #3993

/cc @daniel-j-h 

## Tasklist
 - [x] review
 - [x] adjust for comments

### DIFF
--- a/features/step_definitions/matching.js
+++ b/features/step_definitions/matching.js
@@ -16,9 +16,12 @@ module.exports = function () {
 
                     var headers = new Set(table.raw()[0]);
 
+                    got.code = 'unknown';
                     if (res.body.length) {
                         json = JSON.parse(res.body);
+                        got.code = json.code;
                     }
+
 
                     if (headers.has('status')) {
                         got.status = json.status.toString();
@@ -33,7 +36,7 @@ module.exports = function () {
                         got['#'] = row['#'];
                     }
 
-                    var subMatchings = [],
+                    var subMatchings = [''],
                         turns = '',
                         route = '',
                         duration = '',
--- a/features/support/shared_steps.js
+++ b/features/support/shared_steps.js
@@ -39,6 +39,8 @@ module.exports = function () {
 
                         let json = JSON.parse(body);
 
+                        got.code = json.code;
+
                         let hasRoute = json.code === 'Ok';
 
                         if (hasRoute) {


daniel-j-h:
👍 thanks - could you update the cucumber docs with the changes

oxidase:
@daniel-j-h https://github.com/Project-OSRM/osrm-backend/wiki/Cucumber-Test-Suite#response-codes 
i did not add messages as they are not really interesting to check.

daniel-j-h:
Cherry picked into 5.7 because the 5.7 branch was failing due to https://github.com/Project-OSRM/osrm-backend/commit/321d1988a05ef2147a71a8153d81e929c8c561b5 already making use of `code`.

PR #4025, opened 2017-05-10
Refactor osrm-extract and osrm-contract to use osrm lib interface

mloskot:
Simplifies implementation of the tools slightly.
Ensures `osrm::extract` and `osrm::contract` keep working.

Follows up PR #3787 (see https://github.com/Project-OSRM/osrm-backend/pull/3787#issuecomment-285090810)

### DIFF
--- a/src/tools/contract.cpp
+++ b/src/tools/contract.cpp
@@ -1,7 +1,6 @@
-#include "contractor/contractor.hpp"
-#include "contractor/contractor_config.hpp"
-#include "extractor/profile_properties.hpp"
 #include "storage/io.hpp"
+#include "osrm/contractor.hpp"
+#include "osrm/contractor_config.hpp"
 #include "util/log.hpp"
 #include "util/version.hpp"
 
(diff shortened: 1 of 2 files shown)


daniel-j-h:
👍

I guess we should do the same for customize and partition then, too? (Or better all our libs <-> binaries)

mloskot:
@daniel-j-h Good idea. I'll see if I can issue separate PRs with relevant refactoring in the other tools too.

BTW, Is this failure in https://travis-ci.org/Project-OSRM/osrm-backend/jobs/230803351 related to my changes?

TheMarex:
@mloskot Looks like you need to rerun `./scripts/format.sh` since the logs say that `src/tools/contract.cpp` is not formatted properly.

mloskot:
@TheMarex Cheers. I missed that looking just at the end of the build log.

I re-formatted the files using clang-format 5.x (master) - ATM I'm on VS2017, so I use [clang-format plugin for Visual Studio](http://llvm.org/builds/) snapshot. Despite OSRM seems to require 3.8.0, reformatting with 5.x seems fine.

TheMarex:
Yeah I think `clang-fromat >= 3.8` is generally fine. We had a few problems with `3.7` and `3.8` interacting, that is why we pinned it.

danpat:
Just FYI, I've seen a couple of differences come up with clang-format 5.x :-(  I've pinned my local copy to 3.8.

PR #4029, opened 2017-05-10
Use packed vector for segment weights/durations

TheMarex:
# Issue

Implements #3896 using #3996. 

For Australia that means the `.geometry` files goes from `223809105` to `172337249` (23% reduction.)

## Tasklist
 - [x] Merge https://github.com/Project-OSRM/osrm-backend/pull/3996
 - [x] run on Australia because of the long roads: No clipped segments.
 - [x] benchmark speed update: No measurable difference between `master` and this branch
 - [x] review
 - [x] adjust for comments

## Requirements / Relations

Based on https://github.com/Project-OSRM/osrm-backend/pull/3996


features/testbot/ferry.feature
```diff
@@ -30,15 +30,15 @@ Feature: Testbot - Handle ferry routes
             | fg    |         | ferry | 0:10     |
             | jk    |         | ferry | 1:00     |
             | no    |         | ferry | 24:00    |
-            | rs    |         | ferry | 96:00    |
+            | rs    |         | ferry | 72:00    |
```
oxidase: One segment with clipping like  `| vw    |         | ferry | 96:00    |`  is still needed

src/updater/updater.cpp
```diff
@@ -180,13 +188,22 @@ updateSegmentData(const UpdaterConfig &config,
             ++fallbacks_to_duration;
             rate = value.speed / 3.6;
         }
+        std::cout << "rate: " << rate << std::endl;
```
oxidase: leftover

features/testbot/traffic_speeds.feature
```diff
@@ -99,15 +99,15 @@ Feature: Traffic - speeds
...
-          | g    | a  | ab,ab       | 1 km/h  | 10010.37,0                  | 1:1           |
+          | from | to | route       | speed   | weights                  | a:datasources |
+          | a    | b  | ab,ab       | 1 km/h  | 20020.73,0               | 1:0           |
+          | a    | c  | ab,bc,bc    | 2 km/h  | 20020.73,741.51,0        | 1:1:0         |
+          | b    | c  | bc,bc       | 27 km/h | 741.51,0                 | 1:0           |
+          | a    | d  | ab,eb,de,de | 2 km/h  | 20020.73,378.17,400.41,0 | 1:0:0         |
+          | d    | c  | dc,dc       | 36 km/h | 956.8,0                  | 0             |
+          | g    | b  | ab,ab       | 1 km/h  | 10010.37,0               | 1:0           |
+          | a    | g  | ab,ab       | 1 km/h  | 10010.36,0               | 1             |
+          | g    | a  | ab,ab       | 1 km/h  | 10010.36,0               | 1:1           |
```
oxidase: Also here one test row with a clipped value update would be nice to have

features/testbot/ferry.feature
TheMarex: Actually I discovered I did my math wrong: We still support 96 hours, this was failing for unrelated reasons before.

TheMarex:
There is one test failing only on coverage builds, but it seems repeatable and related to the changes made in this PR. So far no luck reproducing it locally.

TheMarex:
Found the problem but I don't think it's related to changes made in the PR:

This failure occurs https://travis-ci.org/Project-OSRM/osrm-backend/jobs/233362124#L2296 if the new duration is the same as the old duration. If you call `osrm-extract` it will do an in-place update of the `.geometry` file. If you do this twice the old value will be the same as the new value.

Given that the feature in question is not that important and the fix for this is non-trivial, I'll disable this and ticket for further investigation.

Issue #4024, opened 2017-05-10
Master segfaults in osrm-extract on planet


daniel-j-h:
On the demo server running `osrm-extract from master segfaults on the planet.

> Segmentation fault

Without further details. The kernel shows a protection violation:

> kernel: [38657.099205] traps: osrm-extract[21443] general protection ip:5b327d sp:7ffccdd89a90 error:0 in osrm-extract[400000+310000]

Instruction pointer was `0x5b327d` when we got killed.

I could get my hands on the binaries leading to the segfault:

```asm
(gdb) info symbol 0x5b327d
osrm::extractor::guidance::RoundaboutHandler::invalidateExitAgainstDirection(unsigned int, unsigned int, osrm::extractor::guidance::Intersection&) const + 365 in section .text

(gdb) x/i 0x5b327d
   0x5b327d <osrm::extractor::guidance::RoundaboutHandler::invalidateExitAgainstDirection(unsigned int, unsigned int, osrm::extractor::guidance::Intersection&) const+365>:
    movzbl 0x10(%rax),%eax
```

https://github.com/Project-OSRM/osrm-backend/blob/a5bfcd876f601861958f0b4955f35292b45373cf/src/extractor/guidance/roundabout_handler.cpp#L106-L173

Introduced in https://github.com/Project-OSRM/osrm-backend/pull/3980

Unfortunately I could not find a core dump so I don't have a location or a concrete scenario ready.

cc @freenerd @oxidase

oxidase:
The failing case is http://www.openstreetmap.org/way/149545759 , extracted osm 
[map.osm.txt](https://github.com/Project-OSRM/osrm-backend/files/992711/map.osm.txt)

daniel-j-h:
@oxidase fixed this in https://github.com/Project-OSRM/osrm-backend/pull/4031 🙇  - closing as resolved.

PR #4031, opened 2017-05-11
Fix for overlapping ways in invalidateExitAgainstDirection

oxidase:
# Issue

PR fixes invalidateExitAgainstDirection in case of overlapping ways, issue #4024 

## Tasklist
 - [x]  fix the issue
 - [x] review
 - [x] adjust for comments

## Requirements / Relations

### DIFF
--- a/src/extractor/guidance/roundabout_handler.cpp
+++ b/src/extractor/guidance/roundabout_handler.cpp
@@ -1,6 +1,7 @@
 #include "extractor/guidance/roundabout_handler.hpp"
 #include "extractor/guidance/constants.hpp"
 
+#include "util/assert.hpp"
 #include "util/bearing.hpp"
 #include "util/coordinate_calculation.hpp"
 #include "util/guidance/name_announcements.hpp"
@@ -157,6 +158,8 @@ void RoundaboutHandler::invalidateExitAgainstDirection(const NodeID from_nid,
             }
         }
     }
+
+    OSRM_ASSERT(invalidate_from <= invalidate_to, coordinates[from_nid]);
 
     // Exiting roundabouts at an entry point is technically a data-modelling issue.
     // This workaround handles cases in which an exit precedes and entry. The resulting
--- a/features/guidance/roundabout.feature
+++ b/features/guidance/roundabout.feature
@@ -768,3 +768,28 @@ Feature: Basic Roundabout
            | g,f       | 45 90    | gch,edf,edf | depart,roundabout-exit-2,arrive |
            | g,h       | 45 135   | gch,gch,gch | depart,roundabout-exit-1,arrive |
            | e,e       | 90 270   | edf,edf,edf | depart,roundabout-exit-3,arrive |
+
+    Scenario: CCW and CW roundabouts with overlaps
+        Given the node map
+            """
+            a   d          g   h
+
+            b   c          j   i
+            f   e          k   l
+            """
+
+        And the ways
+            | nodes | highway  | junction   |
+            | abcda | tertiary | roundabout |
+            | ed    | tertiary |            |
+            | af    | tertiary |            |
+            | ghijg | tertiary | roundabout |
+            | kg    | tertiary |            |
+            | hl    | tertiary |            |
+
+        When I route I should get
+            | from | to | route    | turns                           | distance |
+            | e    | f  | ed,af,af | depart,roundabout-exit-1,arrive | 80.1m    |
+            | f    | e  | af,ed,ed | depart,roundabout-exit-1,arrive | 120.1m   |
+            | k    | l  | kg,hl,hl | depart,roundabout-exit-1,arrive | 80.1m    |
+            | l    | k  | hl,kg,kg | depart,roundabout-exit-1,arrive | 120.1m   |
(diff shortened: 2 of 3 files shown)


src/extractor/guidance/roundabout_handler.cpp
```diff
@@ -131,43 +132,59 @@ void RoundaboutHandler::invalidateExitAgainstDirection(const NodeID from_nid,
...
-                }
-            }
+                intersection_exit = index;
             else
-            {
-                roundabout_entry_first = true;
-            }
+                intersection_entry = index;
+        }
+        else if (road == size || intersection[road].angle == intersection[index].angle)
```
MoKob: I am unsure whether I understand this correctly. You are choosing the last road of a series of roads with same angles that is not a roundabout?

So on `(0:true), (0:false), (0:false)`, which one would you end up choosing (angle : is_roundabout). Shouldn't you compare to the `via_eid` somehow? Or is this mainly about not crashing?

Wouldn't the order of these roads be more or less random?

I feel that for a correct fix for the overarching problem, we would have to do some tie-breaking mechanisms in the road generation to ensure that these angles are always sorted correctly.

src/extractor/guidance/roundabout_handler.cpp
```diff
@@ -131,43 +132,59 @@ void RoundaboutHandler::invalidateExitAgainstDirection(const NodeID from_nid,
...
+        {
+            road = index;
         }
     }
 
+    OSRM_ASSERT(road != size, coordinates[from_nid]);
+    OSRM_ASSERT(intersection_entry != size, coordinates[from_nid]);
+    OSRM_ASSERT(intersection_exit != size, coordinates[from_nid]);
+    if (road == size || intersection_entry == size || intersection_exit == size)
+    { // ignore wrong intersections
```
MoKob: Right above you assert that all of these are unequal to size. Is this simply an additional save-guard, or am I missing something?

src/extractor/guidance/roundabout_handler.cpp
oxidase: yes, `road` is the last one in series of non-roundabout edges it is random for more than one. 
I think tie-breaking  with the  assumption that intersection.front() corresponds to `via_eid` is better and can be useful in other places.

src/extractor/guidance/roundabout_handler.cpp
oxidase: it is just a safe-guard, but from the above comment it is not a proper solution, and with tie-breaking it is not required

oxidase:
@MoKob the `entering_via_edge` now is swapped with the front value. The PR change is near #4038 assertion,  but the issue does not look related to the PR

daniel-j-h:
@oxidase @MoKob - what's the state here can this be rebased onto master and then can go in to prevent master from segfaulting during extraction?

daniel-j-h:
Also @oxidase you tagged this for 5.7.1 but it should not be in the 5.7 branch only in master, no?

MoKob:
@daniel-j-h personally I would like to document a tie-breaking requirement for equal angles.

The example here has two roads at `0` angle. Swapping them, as done here, assures that the first road in the intersection will represent the opposite of `via_eid`, but it also does not solve the complete problem.
The order of these roads can be important, especially when it comes to roundabout counting. If the reverse of `via_eid` is to the right of the other road, we are entering the roundabout normally. Otherwise we are maybe crossing through the roundabout.
The swap that is implemented right now switches randomly between crossing and entering.
I am fine with merging this PR as is to unblock the demo-server processing, but I would strongly prefer to ticket this fix for further improvement.

oxidase:
@daniel-j-h it is only in master, i have removed 5.7.1 tag

@MoKob to unblock master it is better to revert 2402d60429584aaa1ade632c796c3136667311ad 19494984eb6b0f15ab05d5c11dac60937d07e7e9 and  do not merge an incomplete solution.

Ordering by angles is not enough, one possibility would be to additionally order by signed curvatures at the intersection point if angles are equal (a tangentially touching smaller circle can be inside and outside of the bigger one).

oxidase:
@MoKob i pushed another try to fix the issue. Now order is preserved, but `intersection_view` is rotated to place `entering_via_edge` at the front of `intersection_view` and  this is the expectation now for [the assertion](https://github.com/Project-OSRM/osrm-backend/blob/5ddb0ddcbc9b4596bec29a7d6f44de6478da36a4/src/extractor/guidance/intersection_generator.cpp#L383-L385). The change requires of setting 360 degrees for entries that are moved to the back of  `intersection_view`, but it must be correct for modulo 360 arithmetic.

Anyway the ordering must be fixed later, for example,  magic roundabouts are really interesting edge cases http://map.project-osrm.org/?z=18&center=51.562870%2C-1.770682&loc=51.562625%2C-1.769931&loc=51.563429%2C-1.773047&hl=en&alt=0

MoKob:
@oxidase I don't know if any case exists that assumes angle < 360. Did you change to 359.9? 360.0 could theoretically be asserted for elsewhere. Did you do a `grep` for 360? Could be worth a shot. Otherwise I think this is the better solution 👍

oxidase:
@MoKob i have checked OSM planet with car.profile and it finished without assertions. The angle modified in PR is used only to compute angles differences mod 360, so using 360 degrees is ok to preserve ordering.

PR #4033, opened 2017-05-11
Use foot profile in bike profile

emiltin:
Continues the refactoring of the bike profile, to a point where the way_function is now simply a list of handlers to call.

The main change is that all pushing of bikes is is now handled by calling out to the foot profile. This greatly simplifies the bike profile and makes the handling more complete, since the foot profile handles many more edge cases than the rudimentary foot handling that was present in the bike profile.

First the normal bike handling is performed. If one or both directions is not routable my bike, the foot profile is called. Finally the result is merged, ie. if you can't bike in a particular direction, the foot result in that direction is copied to the result.

Also has some additional improvements, like providing functions directly to the Handlers.run(), making it possible to interleave your own custom functions.

The configs in the profile for speeds is improve and now handles multiple tags together.

Added a number of tests, particularly related to pushing of bikes.

There are a few tests failing, related to ferry times for cars changing slightly. Not sure why.

I'm interested in feedback on the approach of using the foot profile in the bike profile.

### DIFF
--- a/features/bicycle/access.feature
+++ b/features/bicycle/access.feature
@@ -7,126 +7,216 @@ Feature: Bike - Access tags on ways
 
     Scenario: Bike - Access tag hierarchy on ways
         Then routability should be
-            | highway | access | vehicle | bicycle | bothw   |
-            | primary |        |         |         | cycling |
-            | primary | yes    |         |         | cycling |
-            | primary | no     |         |         |         |
-            | primary |        | yes     |         | cycling |
-            | primary |        | no      |         |         |
-            | primary | no     | yes     |         | cycling |
-            | primary | yes    | no      |         |         |
-            | primary |        |         | yes     | cycling |
-            | primary |        |         | no      |         |
-            | primary | no     |         | yes     | cycling |
-            | primary | yes    |         | no      |         |
-            | primary |        | no      | yes     | cycling |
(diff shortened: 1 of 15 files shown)


TheMarex:
Approach looks good to me. Thank you for taking a stab at this. 👍 

Only thing I would consider is a better module interface for each file. Maybe more something like:

```lua
return {
["way_function"] = way_function,
["node_function"] = node_function,
["turn_function"] = turn_function,
["profile"] = profile
}
```

For every profile file? One thing to consider is that for some cucumber tests we "patch" profile files like:

```
require 'bicycle'
properties.weight_name = 'cyclability'
```

I guess this would break?

emiltin:
patching from tests works fine, since it's referring to the main profile loaded.
it would be nice to move the properties into the profile table, so we just have one table. i guess this would be considered a breaking change, since profiles needs to be updated?

TheMarex:
@emiltin I'm going to split off some of your new `bicylce` tests in a new PR so it will be easier to compare against master (and backport the fixes to them for `5.7`)

emiltin:
good idea. i'm splitting out the module interface improvements.

emiltin:
module and api improvements is will be handled in #4072. going to close this PR and open a new that builds on #4072.

Issue #3905, opened 2017-04-05
Reasonable Alternative Routes for MLD


daniel-j-h:
Provide reasonably looking alternative routes (many, if possible) with reasonable query times (< 100ms) on MLD. The main use-cases are alternatives directly for the end-users but also for having alternatives in the context of traffic.

Notes:
- The task here is to find alternatives for a single metric. Once we have multi-metric it could make sense to return fastest, shortest, most efficient, .. routes (out of scope).
- We are also not doing the Penalty Method here (penalizing edges to simulate congestion) - it can be somewhat done via the traffic feature externally already (out of scope).

## Resources

- [Alternative Route Techniques](http://algo2.iti.kit.edu/documents/Dissertation_Kobitzsch_Moritz.pdf)
- [Candidate Sets for Alternative Routes in Road Networks](https://algo2.iti.kit.edu/download/s-csarrn-12.pdf)
- [Algorithmen für Routenplanung: Alternativrouten](http://i11www.iti.uni-karlsruhe.de/_media/teaching/sommer2016/routenplanung/chap2-alternative.pdf) (requires learning German first)

## General Idea

Here's the general idea for the so called Via-Method for finding alternative routes.

- Start search from `s` and continue "for a while" when `t` was found. Save all vertices.
- Start search from `t` and continue "for a while" when `s` was found. Save all vertices.
- Intersect both vertex sets: these are the candidate vertices.
- For all candidate vertices `c` a (potentially arbitrarily bad) alternative route is `(s, c, t)`.
- Make route request with candidate `c` as via: `(s, c, t)` .
- Apply heuristic to evaluate alternative route based on stretch, overlap, how reasonable it is.

Problem: the candidate set is huge; need to be pruned for reasonable candidates.

## MLD Specific Pruning

The partition and the level structure we generate out of it can be used to shrink the candidate sets.

- Prune candidate vertices based on partition border vertices: all routes and therefore also alternative routes have to go through border vertices. Only consider `(s, c, t)` alternative routes where candidate `c` is a border vertex. \*
- Add meta data to all border vertices \* e.g. a flag if the vertex is on highway. Candidate vertices should be "important" in the road network to prevent unreasonable detours. Only consider `(s, c, t)` alternative routes where candidate `c` is on a highway.

\* Let's assume at a fixed level for now. Later on we might want to descend.

## Look out for

There are some constraints we need to operate with and ideas how we can make it work.

[...]

PR #4035, opened 2017-05-11
Refactors Existing Alternatives Architecture

daniel-j-h:
Splitting off some required alternatives refactoring work to already target master. Functionality-wise nothing changes - but it lifts some restrictions to move forward in my mld alternatives work:

https://github.com/Project-OSRM/osrm-backend/issues/3905

Note: the API needs to know that our alternatives impl. is using the via-method and therefore does not support queries with via nodes (only s,t queries) - would require some more re-structuring. I made a hard cut there.

### DIFF
--- a/include/engine/api/route_api.hpp
+++ b/include/engine/api/route_api.hpp
@@ -41,27 +41,33 @@ class RouteAPI : public BaseAPI
     {
     }
 
+    void MakeResponse(const InternalManyRoutesResult &raw_routes,
+                      util::json::Object &response) const
+    {
+        BOOST_ASSERT(!raw_routes.routes.empty());
+
+        util::json::Array jsRoutes;
+
+        for (const auto &route : raw_routes.routes)
+        {
+            if (!route.is_valid())
+                continue;
+
+            jsRoutes.values.push_back(MakeRoute(route.segment_end_coordinates,
+                                                route.unpacked_path_segments,
+                                                route.source_traversed_in_reverse,
+                                                route.target_traversed_in_reverse));
+        }
+
+        response.values["waypoints"] =
+            BaseAPI::MakeWaypoints(raw_routes.routes[0].segment_end_coordinates);
+        response.values["routes"] = std::move(jsRoutes);
+        response.values["code"] = "Ok";
+    }
+
     void MakeResponse(const InternalRouteResult &raw_route, util::json::Object &response) const
     {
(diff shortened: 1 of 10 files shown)


include/engine/api/route_api.hpp
```diff
@@ -41,29 +41,35 @@ class RouteAPI : public BaseAPI
...
+
+        response.values["waypoints"] =
+            BaseAPI::MakeWaypoints(raw_routes.routes[0].segment_end_coordinates);
+        response.values["routes"] = std::move(jsRoutes);
         response.values["code"] = "Ok";
     }
 
+    void MakeResponse(const InternalRouteResult &raw_route, util::json::Object &response) const
+    {
+        return MakeResponse(raw_route, response);
```
danpat: This function calls itself infinitely.

Should this be `return MakeResponse(InternalManyRouteResult(raw_route),response);` ?

daniel-j-h: Hm the InternalRouteResult should just silently convert to a InternalManyRouteResult, at least that was the plan. Not sure you're right because then basically no test would have run through. I will make a pr to change it to your suggestion - it's much clearer. Thanks for flagging!

daniel-j-h: https://github.com/Project-OSRM/osrm-backend/pull/4057

danpat: I noticed because clang issues a warning 😀

daniel-j-h: Your local Clang or Clang 4 on Travis? Which warning is it? Can we enable it and `-Werror` on Travis? I want to catch these things asap and it looks like this was a serious issue.

danpat:
```
/Users/danpat/mapbox/osrm-backend/include/engine/api/route_api.hpp:69:5: warning: all paths through this function will call itself [-Winfinite-recursion]
    {
    ^
```

I'm using `clang` from my local XCode install:

```
$ clang --version
Apple LLVM version 8.1.0 (clang-802.0.42)
Target: x86_64-apple-darwin16.5.0
Thread model: posix
InstalledDir: /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin
```

I have no idea how it maps to the upstream version, but it's from XCode 8.3.2.

PR #4036, opened 2017-05-12
Refactor nodes file

### LINKED ISSUE #4028: Reduce data duplication in fileIndex files
As a part of #3955 some [duplicated data](https://github.com/Project-OSRM/osrm-backend/blob/e2b7e8a4daa5a0f916fe78ab533f2790ab916c35/src/extractor/edge_based_graph_factory.cpp#L148-L158) in `.fileIndex` file can be removed now:

* packed_geometry_id
* forward_data.name_id
* forward_data.travel_mode
* reverse_data.travel_mode

Also refactoring related to a proper fix of #3625

oxidase:
# Issue

PR fixes #4028

## Tasklist
 - [x] Cleanup `fileIndex`, `nodes` and `nodes_data` files
 - ~~[ ] Use Geometry views to prevent vector copies~~
 - [x] Check size reduction for `germany-latest.osm.pbf`
 - [x] review
 - [x] adjust for comments

## Requirements / Relations
PR partially duplicates #3446


oxidase:
Changed sizes germany-latest.osrm.fileIndex 817152000 -> 614072320 ~25% and germany-latest.osrm.ramIndex 836852-> 629372 ~25%

oxidase:
@TheMarex [`u` and `v`](https://github.com/Project-OSRM/osrm-backend/blob/b76ba89ad369b0d65e9a0762efb5d7d4445dd710/include/extractor/edge_based_node.hpp#L42-L43) can also be removed, but it will require indirection `forward_segment_id` -> `geometry_id` -> `geometry` -> `geometry[fwd_segment_position]` -> `coordinates`. So phantom nodes snapping will require a [copy](https://github.com/Project-OSRM/osrm-backend/blob/b76ba89ad369b0d65e9a0762efb5d7d4445dd710/include/engine/datafacade/contiguous_internalmem_datafacade.hpp#L561) of a geometry vector. The  copying of vectors can be avoided by returning `boost::iterator_range` but this will expose internal iterator types in a datafacade, so i would like to skip it in this PR. Also removing `reverse_segment_id` will require an additional index and/or changes in snapping phantom nodes.

With the current implementation size of `osrm::extractor::EdgeBasedNode` changes from 32 to 20 bytes.

oxidase:
Size reduction for `germany-latest.osm.pbf` is 26%:
* `fileIndex` 817152000->511221760   -37%
* `ramIndex` 836852->524036   -37%
* `nodes_data` 126112613->182162657 +44%

oxidase:
Runtime memory consumption for CH+MLD `germany-latest.osm.pbf` data slightly increased  from 
```
[info] Allocating shared memory of 3731803922 bytes
```
to
```
[info] Allocating shared memory of 3787552858 bytes
```
because part of `fileIndex` data moved into a shared memory block

TheMarex:
@oxidase is that a measurement against `master` or against `5.7.0`?

oxidase:
@TheMarex no, it is `master` against the branch

include/extractor/edge_based_graph_factory.hpp
```diff
@@ -93,7 +93,8 @@ class EdgeBasedGraphFactory
 
     // The following get access functions destroy the content in the factory
     void GetEdgeBasedEdges(util::DeallocatingVector<EdgeBasedEdge> &edges);
-    void GetEdgeBasedNodes(std::vector<EdgeBasedNode> &nodes);
+    void GetEdgeBasedNodes(EdgeBasedNodeDataContainer &ebg_node_data_container);
+    void GetNodeBasedEdges(std::vector<EdgeBasedNode> &nodes);
```
TheMarex: The naming here is a little bit confusing. This function returns a `EdgeBasedNode` but is named `GetNodeBasedEdge`.  I think we should rename that class, it does not represent a single `EdgeBasedNode` but actually a single segment. Maybe `ExternalSegmentData`?

include/extractor/node_data_container.hpp
```diff
@@ -65,6 +68,13 @@ template <storage::Ownership Ownership> class EdgeBasedNodeDataContainerImpl
         travel_modes[node_id] = travel_mode;
     }
 
+    // Used by EdgeBasedGraphFactory to fill data structure
+    template <typename = std::enable_if<Ownership == storage::Ownership::Container>>
+    void SetData(NodeID node_id, ComponentID component_id)
```
TheMarex: Maybe a less generic name: `SetComponentID`.

src/extractor/extractor.cpp
```diff
@@ -359,18 +362,6 @@ void Extractor::FindComponents(unsigned max_edge_id,
         }
     }
 
-    // connect forward and backward nodes of each edge
```
oxidase: @TheMarex just forgot to mention this place that was introduced in c80c2233c5a537696e194c55b6f4f9cf162b4875. i don't get idea why graph must be extended with potentially non-exisiting edges. So if there is a use case for it must be  reverted before merge

oxidase:
@TheMarex as discussed per chat `EdgeBasedNode` is renamed to `EdgeBasedNodeSegment`

src/extractor/extractor.cpp
TheMarex: This code is necessary to enforce that the component ID of the forward and backward node is the same (the implicit assumption is that we are always about to do a u-turn on the spot). There are of course cases where this is generally not true, but we can't handle these right now anyway (for that we would need to split the PhantomNode into forward/backward and have two component IDs).

Issue #2201, opened 2016-04-03
Keep waypoints on the left/right


danpat:
If you're a taxi driver planning a route, it might be preferable to approach a waypoint on the side of the road that deposits/picks up your passenger without needing to cross the road, and then continue routing you without issuing a u-turn.

This is partly implemented by the "allow u-turns at waypoints" parameter in the Lua profiles.

We could implement a query-time parameter called "keep waypoints on the left/right".  When constructing phantom nodes, we would eliminate edges that do not put the foot segment on the specified side.  The behaviour would be similar to the bearing filter - only the relevant edges would be seeded on the query heap.

It should be a query-time parameter to allow for left/right side driving, allowing localized apps to specify the desired side.

frodrigo:
This feature is also interesting in start/end point and on matrix computation, since matrix is just route segments for TSP optimization. Sometimes I recompute only segment of route, but still interested on waypoints on the left/right.

jperaltamalvar:
This feature would be useful for buses, since they only have doors on one side, either on the left (british ones) or on the right (others), as well as some kind of trucks, like [this one](https://stoysnetcdn.com/brud/brud2761/brud2761.jpg).

In Network Analyst this is the `CurbApproach `parameter.

emiltin:
Also relevant for bike and foot routing, since crossing bigger roads is often only possible at intersections.

TheMarex:
This was addressed with the `approaches` parameter.

PR #4039, opened 2017-05-12
Parameter for keep waypoints on the left/right

fijemax:
# Description : 
This PR adds a parameter for each point to chose the start/stop side.
Three option can be used : 
  - DEFAULT to start/stop at the drive side
  - OPPOSITE to start/stop at the opposite of default (cross the road)
  - BOTH each side can be used (old behavior).

The default depending at the country.

# Issue
This pull request is linking to this issue : 
https://github.com/Project-OSRM/osrm-backend/issues/2201

## Tasklist
 - [x] Adding sides parameter into base parameters, it can take the values DEFAULT, OPPOSITE or BOTH.
 - [x] Adding url parser for "sides" parameter, url values are "d" for DEFAULT, "o" for OPPOSITE and "b" for BOTH, example : "sides=s;s".
 - [x] Checking parameters, if "sides" parameter is used, the number of parameter is the same as number of location.
 - [x] Create a phantom to start/stop at side drive or opposite drive side.
 - [x] Retrieve the country's drive side for apply the DEFAULT.

**For the moment the country's drive side is set to right by default, we can find this parameter in the lua file.
How can access this parameter in the osrm-routed ?**

Example for DEFAULT on start and stop route point value : 
Here start and stop 
capture d ecran de 2017-05-12 16-07-54
capture d ecran de 2017-05-12 16-07-32

Another example : 
capture d ecran de 2017-05-12 18-00-44 capture d ecran de 2017-05-12 18-00-51

**For the one-way roads :** 
The algorythm do not use the "side" parameter for the phantom node option because the driver hasn't the choice.
capture d ecran de 2017-05-12 16-11-29
capture d ecran de 2017-05-12 16-11-37


fijemax:
@frodrigo

TheMarex:
Thank you for taking a stab at this! 🎉 I just skimmed the diff quickly some things that I noticed:

1. No test coverage, this should have a cucumber test
2. No URL parser test, this should have a unit-test
3. NodeJS bindings are not extended yet.

I can put a real review on my TODO for later today, I need to think a little bit about parameter naming and such.

fijemax:
Hi @TheMarex 
I I extended the NodeJS  for the sides parameter.
I added the cucumber test.

features/testbot/side_param.feature
```diff
@@ -0,0 +1,207 @@
...
+            """
+
+        And the ways
+            | nodes |
+            | ab    |
+            | bc    |
+
+        When I route I should get
+            | from | to | sides | route    |
+            | s    | e  | b o   | ab,bc,bc |
```
oxidase: is it possible to add a turn instructions check? to see explicitly a u-turn

include/engine/geospatial_query.hpp
```diff
@@ -208,16 +211,50 @@ template <typename RTreeT, typename DataFacadeT> class GeospatialQuery
...
+                    //                       |
+                    //                       |
+                    //                       |
+                    //                       |
+                    //                       |
+                    //                       |
+                    //                       |
+                    //                 input_coordinate
+
+                    bool input_coordinate_is_at_right = !util::coordinate_calculation::isCCW(
```
oxidase: `util::coordinate_calculation::isCCW(coordinates[segment.data.u], input_coordinate, coordinates[segment.data.v]);` to correspond the figure above

fijemax:
I just added the nodeJS binding.

include/server/api/base_parameters_grammar.hpp
```diff
@@ -149,10 +159,13 @@ struct BaseParametersGrammar : boost::spirit::qi::grammar<Iterator, Signature>
             qi::lit("bearings=") >
             (-(qi::short_ > ',' > qi::short_))[ph::bind(add_bearing, qi::_r1, qi::_1)] % ';';
 
+        sides_rule = qi::lit("sides=") >
```
fijemax: I tryed to change by this code :
```
        side_rule = qi::char_("a-zA-Z")[qi::_val = engine::Side{engine::BOTH}];
        struct StartSideValues: qi::symbols<char, engine::Side>
        {
            StartSideValues() {add("b", engine::Side{engine::BOTH})("d", engine::Side{engine::DEFAULT})("o", engine::Side{engine::OPPOSITE})("", engine::Side{engine::OPPOSITE}); }
        };
        StartSideValues values;
        sides_rule = qi::lit("sides=") >
                        (-(values | side_rule) %
                         ';')[ph::bind(&engine::api::BaseParameters::sides, qi::_r1) = qi::_1];
```
But it provokes segfault.
I am trying to analyse with gdb but it's a really complicated backtrace.

oxidase: @fijemax it crashes because grammars are local variables in the constructor.

What i suggest is to use in the way as in commit https://github.com/oxidase/osrm-backend/commit/301a1cca6c5a6f3ee229213448577d9753e337d5 It contains more restrictive grammar [bod|eps]

Please add also some unit tests for parsing incorrect URLs as suggested by @TheMarex

fijemax:
Hi @oxidase, thanks for parser example.
I refactored the code 
 - Simple enumerator for Side
 - Change gi::lit to qi::symbol for side parser
I added URL tests in the unit_tests

unit_tests/server/parameters_parser.cpp
```diff
@@ -46,7 +46,7 @@ BOOST_AUTO_TEST_CASE(invalid_route_urls)
     BOOST_CHECK_EQUAL(testInvalidOptions<RouteParameters>("1,2;3,4?overview=false&radiuses=foo"),
                       32UL);
     BOOST_CHECK_EQUAL(testInvalidOptions<RouteParameters>("1,2;3,4?overview=false&sides=foo"),
-                      30UL);
+                      29UL);
```
oxidase: :+1:

include/engine/side.hpp
```diff
@@ -0,0 +1,47 @@
...
+ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+(INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+*/
+
+#ifndef OSRM_ENGINE_SIDE_HPP
+#define OSRM_ENGINE_SIDE_HPP
+
+#include <string>
```
oxidase: It is not needed

include/server/api/base_parameters_grammar.hpp
```diff
@@ -149,10 +149,17 @@ struct BaseParametersGrammar : boost::spirit::qi::grammar<Iterator, Signature>
             qi::lit("bearings=") >
             (-(qi::short_ > ',' > qi::short_))[ph::bind(add_bearing, qi::_r1, qi::_1)] % ';';
 
+        side_type.add("d", engine::Side::DEFAULT)("o", engine::Side::OPPOSITE)("b", engine::Side::BOTH);
+        sides_rule = qi::lit("sides=") >
+                        (-(side_type | qi::attr(engine::Side::BOTH)) %
+                         ';')[ph::bind(&engine::api::BaseParameters::sides, qi::_r1) = qi::_1];
+                     //(-qi::symbols[qi::char_("aA-zZ")])[ph::bind(add_side, qi::_r1, qi::_1)] % ';';
```
oxidase: not needed

include/server/api/base_parameters_grammar.hpp
```diff
@@ -149,10 +149,17 @@ struct BaseParametersGrammar : boost::spirit::qi::grammar<Iterator, Signature>
             qi::lit("bearings=") >
             (-(qi::short_ > ',' > qi::short_))[ph::bind(add_bearing, qi::_r1, qi::_1)] % ';';
 
+        side_type.add("d", engine::Side::DEFAULT)("o", engine::Side::OPPOSITE)("b", engine::Side::BOTH);
+        sides_rule = qi::lit("sides=") >
+                        (-(side_type | qi::attr(engine::Side::BOTH)) %
```
oxidase: Here must be either `-` for `vector<optional<Side>>` or `qi::attr(engine::Side::BOTH)`  for `vector<Side>` 
`-` will place empty optionals for empty values, but `attr(BOTH)` will place `BOTH`.

`(-side_type %';')[...]` is enough for vector of optionals

TheMarex:
Thanks for adding the tests! Some API features need a quick discussion:

1. Is there a real use-case for `opposite`? I feel the actual feature here is the ability to force the curb approach.
2. Parameter name. I feel something like `approach={curb,unrestriced}` might be more descriptive.
3. Do we need the ability to configure this on per-waypoint basis? The scenarios I could come up with (taxi, garbage can pickup) would only need this to be a global flag that applies to all waypoints.

frodrigo:
> 1. Is there a real use-case for opposite? I feel the actual feature here is the ability to force the curb approach.

Yes we build it for a use case where we need the curb approach. We do not have use case the opposite. But why implement half an option as we can easily make it for both sides? Nevertheless we can remove opposite if you want.

> 3. Do we need the ability to configure this on per-waypoint basis? The scenarios I could come up with (taxi, garbage can pickup) would only need this to be a global flag that applies to all waypoints.

We need it per waypoint. E.g. on garbage pickup, the pickup point must be done with the curb approach, but at the depot the side does not matter.

We also plan to add to this PR the support of the approach to other plugins. Especially on the table plugin (approach per waypoint make even more sens on it).

TheMarex:
> But why implement half an option as we can easily make it for both sides?

One of the main concerns with adding stuff is usually API-surface area. The fewer options we have, the less likely it is that we need to do a breaking API change. Would be cool if we could limit this to only `unrestricted` and `curb`.

> We need it per waypoint. E.g. on garbage pickup, the pickup point must be done with the curb approach, but at the depot the side does not matter.

Okay thanks for the explanation. /cc @daniel-j-h because he has been doing a lot of VRP stuff lately and this might be useful for him too?

> We also plan to add to this PR the support of the approach to other plugins.

🤔 I think that should already work if the right method was extended, all plugins rely on `GetPhantomNodes()` which should operate on the `BaseParameter` (hence works for all plugins, similar to how we handle `bearings` and `radiuses`)

---------------------------

I would like to propose the following API:

### HTTP

- `approaches={approach}[;{approach}; ...]` where `approach` is a `String` of either `curb` or `unrestricted` (default: `unrestricted`). Needs to have the same length as the number of coordinates given.

Example: `?approaches=curb;unrestricted;curb;curb;curb;unrestricted` or`?approaches=curb;;curb;curb;curb;`

### Node JS

- `approaches` **Array** of **String**s that can either be `curb` or `unrestricted`. Needs to have the same length as the number of coordinates given. (default: `null`, which means `unrestricted`)

Example:  `{approaches: ["curb", "unrestricted", "curb", "curb"]}` or `{approaches: ["curb", null, "curb", "curb"]}`

fijemax:
Hi @TheMarex ,
Thank you for the review, I going to make the changes as respecting your API descriptions.
Also, I will don't know how to retrieve the country's drive side for apply the `curb` option, this is actualy possible ?
For the moment I apply right side by default for `curb` approach.

TheMarex:
> Also, I will don't know how to retrieve the country's drive side for apply the curb option, is this actually possible?

Ah this is part of `ProfileProperties::left_hand_driving`. To use this you would need to extend `ContiguousInternalMemoryDataFacadeBase` and `BaseDataFacade` with a new function like `bool IsLeftHandDriving() { return m_profile_properties.left_hand_driving; }`.

It can be configured in the profile via `properties.left_hand_driving = true`.

daniel-j-h:
> We need it per waypoint. E.g. on garbage pickup, the pickup point must be done with the curb approach, but at the depot the side does not matter.
> 
> Okay thanks for the explanation. /cc @daniel-j-h because he has been doing a lot of VRP stuff lately and this might be useful for him too?

Yeah I can totally see how this is useful. You can somewhat fake it already setting bearings - but then you need to do some pre-processing / set the bearing constraints manually.

emiltin:
note that ProfileProperties::left_hand_driving is a global property set once by the profile. it doesn't take into account what country the current request (and and waypoint) is located in. so if you're working with a g planet dataset, the api will be wrong in some countries, and right in others.

fijemax:
@oxidase 
I refactored the API and propagate the param for every phantom nodes search function.
The support for left hand driving is ok.
Cucumber tests are ok and url unit tests too.

TheMarex:
@fijemax looks like the only thing that is failing right now is the formating check. Can you try:

```
./scripts/format.sh
```

That requires `clang-format` 3.8.

include/engine/approach.hpp
```diff
@@ -0,0 +1,44 @@
...
+
+#ifndef OSRM_ENGINE_APPROACH_HPP
+#define OSRM_ENGINE_APPROACH_HPP
+
+namespace osrm
+{
+namespace engine
+{
+
+enum Approach
```
TheMarex: `enum class Approach : std::uint8_t` would be better here, for one to remove implicit conversions and to limit the storage requirements to one byte per coordinate.

include/engine/datafacade/contiguous_internalmem_datafacade.hpp
```diff
@@ -650,108 +651,117 @@ class ContiguousInternalMemoryDataFacadeBase : public BaseDataFacade
...
+        return m_geospatial_query->NearestPhantomNodesInRange(input_coordinate, max_distance, approach);
     }
 
     std::vector<PhantomNodeWithDistance>
     NearestPhantomNodesInRange(const util::Coordinate input_coordinate,
                                const float max_distance,
                                const int bearing,
-                               const int bearing_range) const override final
+                               const int bearing_range,
+                               const engine::Approach approach) const override final
```
TheMarex: `engine::` is unnecessary.

include/engine/datafacade/contiguous_internalmem_datafacade.hpp
```diff
@@ -650,108 +651,117 @@ class ContiguousInternalMemoryDataFacadeBase : public BaseDataFacade
...
         return m_geospatial_query->NearestPhantomNodesInRange(
-            input_coordinate, max_distance, bearing, bearing_range);
+            input_coordinate, max_distance, bearing, bearing_range, approach);
     }
 
     std::vector<PhantomNodeWithDistance>
     NearestPhantomNodes(const util::Coordinate input_coordinate,
-                        const unsigned max_results) const override final
+                        const unsigned max_results,
+                        const engine::Approach approach) const override final
```
TheMarex: `engine::`

include/engine/datafacade/contiguous_internalmem_datafacade.hpp
```diff
@@ -650,108 +651,117 @@ class ContiguousInternalMemoryDataFacadeBase : public BaseDataFacade
...
-        return m_geospatial_query->NearestPhantomNodes(input_coordinate, max_results);
+        return m_geospatial_query->NearestPhantomNodes(input_coordinate, max_results, approach);
     }
 
     std::vector<PhantomNodeWithDistance>
     NearestPhantomNodes(const util::Coordinate input_coordinate,
                         const unsigned max_results,
-                        const double max_distance) const override final
+                        const double max_distance,
+                        const engine::Approach approach) const override final
```
TheMarex: `engine::`

include/engine/datafacade/contiguous_internalmem_datafacade.hpp
```diff
@@ -650,108 +651,117 @@ class ContiguousInternalMemoryDataFacadeBase : public BaseDataFacade
...
+        return m_geospatial_query->NearestPhantomNodes(input_coordinate, max_results, max_distance, approach);
     }
 
     std::vector<PhantomNodeWithDistance>
     NearestPhantomNodes(const util::Coordinate input_coordinate,
                         const unsigned max_results,
                         const int bearing,
-                        const int bearing_range) const override final
+                        const int bearing_range,
+                        const engine::Approach approach) const override final
```
TheMarex: `engine::`

include/engine/datafacade/contiguous_internalmem_datafacade.hpp
```diff
@@ -650,108 +651,117 @@ class ContiguousInternalMemoryDataFacadeBase : public BaseDataFacade
...
     }
 
     std::vector<PhantomNodeWithDistance>
     NearestPhantomNodes(const util::Coordinate input_coordinate,
                         const unsigned max_results,
                         const double max_distance,
                         const int bearing,
-                        const int bearing_range) const override final
+                        const int bearing_range,
+                        const engine::Approach approach) const override final
```
TheMarex: `engine::`

include/engine/datafacade/contiguous_internalmem_datafacade.hpp
```diff
@@ -650,108 +651,117 @@ class ContiguousInternalMemoryDataFacadeBase : public BaseDataFacade
...
         return m_geospatial_query->NearestPhantomNodes(
-            input_coordinate, max_results, max_distance, bearing, bearing_range);
+            input_coordinate, max_results, max_distance, bearing, bearing_range, approach);
     }
 
-    std::pair<PhantomNode, PhantomNode> NearestPhantomNodeWithAlternativeFromBigComponent(
-        const util::Coordinate input_coordinate) const override final
+    std::pair<PhantomNode, PhantomNode>
+    NearestPhantomNodeWithAlternativeFromBigComponent(const util::Coordinate input_coordinate,
+                                                      const engine::Approach approach) const override final
```
TheMarex: `engine::`

include/engine/datafacade/contiguous_internalmem_datafacade.hpp
```diff
@@ -650,108 +651,117 @@ class ContiguousInternalMemoryDataFacadeBase : public BaseDataFacade
...
         BOOST_ASSERT(m_geospatial_query.get());
 
         return m_geospatial_query->NearestPhantomNodeWithAlternativeFromBigComponent(
-            input_coordinate);
+            input_coordinate, approach);
     }
 
     std::pair<PhantomNode, PhantomNode> NearestPhantomNodeWithAlternativeFromBigComponent(
-        const util::Coordinate input_coordinate, const double max_distance) const override final
+        const util::Coordinate input_coordinate, const double max_distance, const engine::Approach approach) const override final
```
TheMarex: `engine::`

include/engine/datafacade/contiguous_internalmem_datafacade.hpp
```diff
@@ -650,108 +651,117 @@ class ContiguousInternalMemoryDataFacadeBase : public BaseDataFacade
...
-            input_coordinate, max_distance, bearing, bearing_range);
+            input_coordinate, max_distance, bearing, bearing_range, approach);
     }
 
     std::pair<PhantomNode, PhantomNode>
     NearestPhantomNodeWithAlternativeFromBigComponent(const util::Coordinate input_coordinate,
                                                       const int bearing,
-                                                      const int bearing_range) const override final
+                                                      const int bearing_range,
+                                                      const engine::Approach approach) const override final
```
TheMarex: `engine::`

include/engine/datafacade/datafacade_base.hpp
```diff
@@ -93,44 +94,54 @@ class BaseDataFacade
...
                                const float max_distance,
                                const int bearing,
-                               const int bearing_range) const = 0;
+                               const int bearing_range,
+                               const engine::Approach approach) const = 0;
     virtual std::vector<PhantomNodeWithDistance>
     NearestPhantomNodesInRange(const util::Coordinate input_coordinate,
-                               const float max_distance) const = 0;
+                               const float max_distance,
+                               const engine::Approach approach) const = 0;
```
TheMarex: `engine::`

include/engine/datafacade/datafacade_base.hpp
```diff
@@ -93,44 +94,54 @@ class BaseDataFacade
...
 
     virtual std::vector<PhantomNodeWithDistance>
     NearestPhantomNodes(const util::Coordinate input_coordinate,
                         const unsigned max_results,
                         const double max_distance,
                         const int bearing,
-                        const int bearing_range) const = 0;
+                        const int bearing_range,
+                        const engine::Approach approach) const = 0;
```
TheMarex: `engine::`

include/engine/datafacade/datafacade_base.hpp
```diff
@@ -93,44 +94,54 @@ class BaseDataFacade
...
     virtual std::vector<PhantomNodeWithDistance>
     NearestPhantomNodes(const util::Coordinate input_coordinate,
                         const unsigned max_results,
                         const int bearing,
-                        const int bearing_range) const = 0;
+                        const int bearing_range,
+                        const engine::Approach approach) const = 0;
```
TheMarex: `engine::`

include/engine/datafacade/datafacade_base.hpp
```diff
@@ -93,44 +94,54 @@ class BaseDataFacade
...
     virtual std::vector<PhantomNodeWithDistance>
     NearestPhantomNodes(const util::Coordinate input_coordinate,
-                        const unsigned max_results) const = 0;
+                        const unsigned max_results,
+                        const engine::Approach approach) const = 0;
```
TheMarex: `engine::`

include/engine/datafacade/datafacade_base.hpp
```diff
@@ -93,44 +94,54 @@ class BaseDataFacade
...
     virtual std::vector<PhantomNodeWithDistance>
     NearestPhantomNodes(const util::Coordinate input_coordinate,
                         const unsigned max_results,
-                        const double max_distance) const = 0;
+                        const double max_distance,
+                        const engine::Approach approach) const = 0;
```
TheMarex: `engine::`

include/engine/datafacade/datafacade_base.hpp
```diff
@@ -93,44 +94,54 @@ class BaseDataFacade
...
-    virtual std::pair<PhantomNode, PhantomNode> NearestPhantomNodeWithAlternativeFromBigComponent(
-        const util::Coordinate input_coordinate) const = 0;
     virtual std::pair<PhantomNode, PhantomNode>
     NearestPhantomNodeWithAlternativeFromBigComponent(const util::Coordinate input_coordinate,
-                                                      const double max_distance) const = 0;
+                                                      const engine::Approach approach) const = 0;
+    virtual std::pair<PhantomNode, PhantomNode>
+    NearestPhantomNodeWithAlternativeFromBigComponent(const util::Coordinate input_coordinate,
+                                                      const double max_distance,
+                                                      const engine::Approach approach) const = 0;
```
TheMarex: `engine::`

include/engine/datafacade/datafacade_base.hpp
```diff
@@ -93,44 +94,54 @@ class BaseDataFacade
...
     virtual std::pair<PhantomNode, PhantomNode>
     NearestPhantomNodeWithAlternativeFromBigComponent(const util::Coordinate input_coordinate,
                                                       const double max_distance,
                                                       const int bearing,
-                                                      const int bearing_range) const = 0;
+                                                      const int bearing_range,
+                                                      const engine::Approach approach) const = 0;
```
TheMarex: `engine::`

include/engine/datafacade/datafacade_base.hpp
```diff
@@ -93,44 +94,54 @@ class BaseDataFacade
...
     virtual std::pair<PhantomNode, PhantomNode>
     NearestPhantomNodeWithAlternativeFromBigComponent(const util::Coordinate input_coordinate,
                                                       const int bearing,
-                                                      const int bearing_range) const = 0;
+                                                      const int bearing_range,
+                                                      const engine::Approach approach) const = 0;
```
TheMarex: `engine::`

include/engine/geospatial_query.hpp
```diff
@@ -50,14 +53,17 @@ template <typename RTreeT, typename DataFacadeT> class GeospatialQuery
     // Does not filter by small/big component!
     std::vector<PhantomNodeWithDistance>
     NearestPhantomNodesInRange(const util::Coordinate input_coordinate,
-                               const double max_distance) const
+                               const double max_distance,
+                               const engine::Approach approach) const
```
TheMarex: `engine::`

include/engine/geospatial_query.hpp
```diff
@@ -69,13 +75,17 @@ template <typename RTreeT, typename DataFacadeT> class GeospatialQuery
     NearestPhantomNodesInRange(const util::Coordinate input_coordinate,
                                const double max_distance,
                                const int bearing,
-                               const int bearing_range) const
+                               const int bearing_range,
+                               const engine::Approach approach) const
```
TheMarex: `engine::`

include/engine/geospatial_query.hpp
```diff
@@ -91,13 +101,15 @@ template <typename RTreeT, typename DataFacadeT> class GeospatialQuery
     NearestPhantomNodes(const util::Coordinate input_coordinate,
                         const unsigned max_results,
                         const int bearing,
-                        const int bearing_range) const
+                        const int bearing_range,
+                        const engine::Approach approach) const
```
TheMarex: `engine::`

include/engine/geospatial_query.hpp
```diff
@@ -134,11 +148,15 @@ template <typename RTreeT, typename DataFacadeT> class GeospatialQuery
     // Returns max_results nearest PhantomNodes.
     // Does not filter by small/big component!
     std::vector<PhantomNodeWithDistance>
-    NearestPhantomNodes(const util::Coordinate input_coordinate, const unsigned max_results) const
+    NearestPhantomNodes(const util::Coordinate input_coordinate,
+                        const unsigned max_results,
+                        const engine::Approach approach) const
```
TheMarex: `engine::`

include/engine/geospatial_query.hpp
```diff
@@ -151,11 +169,14 @@ template <typename RTreeT, typename DataFacadeT> class GeospatialQuery
     std::vector<PhantomNodeWithDistance>
     NearestPhantomNodes(const util::Coordinate input_coordinate,
                         const unsigned max_results,
-                        const double max_distance) const
+                        const double max_distance,
+                        const engine::Approach approach) const
```
TheMarex: `engine::`

include/engine/geospatial_query.hpp
```diff
@@ -15,6 +16,8 @@
 #include <memory>
 #include <vector>
 
+#include <util/log.hpp>
```
TheMarex: not needed anymore?

include/engine/geospatial_query.hpp
```diff
@@ -545,6 +578,32 @@ template <typename RTreeT, typename DataFacadeT> class GeospatialQuery
         return datafacade.GetComponentID(data.forward_segment_id.id).is_tiny;
     }
 
+    std::pair<bool, bool> approach_usage(const util::Coordinate &input_coordinate,
```
TheMarex: Rename to `ApproachUsage` for camel-case. A better name could be `CheckApproach`.

include/nodejs/node_osrm_support.hpp
```diff
@@ -297,6 +298,63 @@ inline bool argumentsToParameter(const Nan::FunctionCallbackInfo<v8::Value> &arg
...
+
+        for (uint32_t i = 0; i < approaches_array->Length(); ++i)
+        {
+            v8::Local<v8::Value> approach_raw = approaches_array->Get(i);
+            if (approach_raw.IsEmpty())
+                return false;
+
+            if (approach_raw->IsNull())
+            {
+                params->bearings.emplace_back();
```
TheMarex: Copy&Paste error, should be `params->approaches`.

include/nodejs/node_osrm_support.hpp
```diff
@@ -297,6 +298,63 @@ inline bool argumentsToParameter(const Nan::FunctionCallbackInfo<v8::Value> &arg
...
 
+    if (obj->Has(Nan::New("approaches").ToLocalChecked()))
+    {
+        v8::Local<v8::Value> approaches = obj->Get(Nan::New("approaches").ToLocalChecked());
+        if (approaches.IsEmpty())
+            return false;
+
+        if (!approaches->IsArray())
+        {
+            Nan::ThrowError("Approaches must be an array of arrays of numbers");
```
TheMarex: Error message is wrong: Should be `arrays of strings`

include/nodejs/node_osrm_support.hpp
```diff
@@ -297,6 +298,63 @@ inline bool argumentsToParameter(const Nan::FunctionCallbackInfo<v8::Value> &arg
...
+                {
+                    params->approaches.push_back(osrm::Approach::CURB);
+                }
+                else if (approach_str == "unrestricted")
+                {
+                    params->approaches.push_back(osrm::Approach::UNRESTRICTED);
+                }
+                else
+                {
+                    Nan::ThrowError("'approach' param must be one of [curb, unrestricted]");
```
TheMarex: `approach` -> `approaches`

fijemax:
Hi @TheMarex 
Change was done.
I don't konw how testing the nodejs binding ?

TheMarex:
@fijemax for some examples look at `test/nodejs/route.js`. What you basically want is to create a request that uses your parameter and try a few permutations that would create an error.

fijemax:
@TheMarex  When I run nodejs `test/nodes/index.js` or `nodejs test/nodes/route.js` I have the following error : 
`Error: Cannot find module './binding/node_osrm.node'`

fijemax:
It's ok, I hadn't activate ENABLE_NODE_BINDINGS during cmake compilation.

TheMarex:
@fijemax awesome, this looks good to merge for me. 🎉

daniel-j-h:
Very cool - looking forward to having this in a stable release!

What I just now realized after this was merged is we need
- [ ] a changelog entry (changelog.md)
- [ ] docs for the http api (docs/http.md)
- [ ] docs for the node bindings (docs/nodejs/api.md)

could you run this out or ticket and tag it for the next release @fijemax so we don't forget about it.

(also feel free to write a osm diary / blog post about this if you have the time for it, I can totally see people being interested in this feature :) I wrote a bit [about bearings](https://www.openstreetmap.org/user/daniel-j-h/diary) - this one would fit right in)

fijemax:
@daniel-j-h I just created this PR #4104.

fijemax:
Hi @daniel-j-h, i wrote [this](https://www.openstreetmap.org/user/Jean-Maxime%20FILLAU/diary/41741) about the new feature.

daniel-j-h:
Sweet, thank you!

asaveljevs:
> 1. Is there a real use-case for opposite? I feel the actual feature here is the ability to force the curb approach.

We have tried to describe our use case for opposite in #5616, which is needed for solving a VRP problem.

geocept:
Is there a reason why this great feature is not in the stable release of OSRM?? I mean 6,5 years have passed now and many many people are waiting for it. Would be great if someone would bring it to stable release.

frodrigo:
> Is there a reason why this great feature is not in the stable release of OSRM?? I mean 6,5 years have passed now and many many people are waiting for it. Would be great if someone would bring it to stable release.

It was released in 5.8.0
https://github.com/Project-OSRM/osrm-backend/blob/master/CHANGELOG.md#580

PR #4040, opened 2017-05-12
Document default values of global properties set in profile

mloskot:
Should spare users time spent on digging through the source code.

BTW, It would be good to mark attributes with mandatory/optional, what you think?

p.s. Sorry for missing `[ci skip]` in commit message.

### DIFF
--- a/docs/profiles.md
+++ b/docs/profiles.md
@@ -31,12 +31,12 @@ Attribute                     | Type     | Notes
 ------------------------------|----------|----------------------------------------------------------------------------
 weight_name                   | String   | Name used in output for the routing weight property (default 'duration')
 weight_precision              | Unsigned | Decimal precision of edge weights (default 1)
-left_hand_driving             | Boolean  | Are vehicles assumed to drive on the left? (used in guidance)
-use_turn_restrictions         | Boolean  | Are turn instructions followed?
-continue_straight_at_waypoint | Boolean  | Must the route continue straight on at a via point, or are U-turns allowed?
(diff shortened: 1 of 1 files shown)


docs/profiles.md
```diff
@@ -31,12 +31,12 @@ Attribute                     | Type     | Notes
 ------------------------------|----------|----------------------------------------------------------------------------
 weight_name                   | String   | Name used in output for the routing weight property (default 'duration')
 weight_precision              | Unsigned | Decimal precision of edge weights (default 1)
-left_hand_driving             | Boolean  | Are vehicles assumed to drive on the left? (used in guidance)
-use_turn_restrictions         | Boolean  | Are turn instructions followed?
-continue_straight_at_waypoint | Boolean  | Must the route continue straight on at a via point, or are U-turns allowed?
+left_hand_driving             | Boolean  | Are vehicles assumed to drive on the left? (used in guidance, default false)
```
TheMarex: Nitpick: Values should be enclosed in code tags `false`.

mloskot: @TheMarex Do you mean boolean values specifically or any?

I followed the status quo, namely these are not code-tagged
* weight_name: default 'duration'
* weight_precision: default 1

TheMarex: Good catch, these should be enclosed too. It's just easier to differentiate what a value and what a name is with `code` tags. 😉

mloskot: Refined in #4052

PR #4044, opened 2017-05-15
Fix bicycle turn penalties

TheMarex:
# Issue

Adds turn penalties for the cyclablity option of bicycles. Replaces https://github.com/Project-OSRM/osrm-backend/pull/3919. Thanks to @MichalPP for the bug report!

## Tasklist
 - [x] Add regression test
 - [x] review
 - [x] adjust for comments

## Requirements / Relations
https://github.com/Project-OSRM/osrm-backend/pull/3919.

### DIFF
--- a/features/bicycle/turn_penalty.feature
+++ b/features/bicycle/turn_penalty.feature
@@ -34,3 +34,39 @@ Feature: Turn Penalties
             | s    | e  | sj,je,je | 53s +-1 | 483m +-1 |
             | s    | f  | sj,jf,jf | 50s +-1 | 400m +-1 |
             | s    | g  | sj,jg,jg | 63s +-1 | 483m +-1 |
+
+    Scenario: Bicycle - Turn penalties on cyclability
+        Given the profile file
+        """
+        require 'bicycle'
+        properties.weight_name = 'cyclability'
+        """
+
+        Given the node map
+            """
+            a--b-----c
+               |
+               |
+               d
+
+            e--------f-----------g
+                  /
+                /
+              /
+            h
+            """
+
+        And the ways
+            | nodes | highway     |
+            | abc   | residential |
+            | bd    | residential |
+            | efg   | residential |
+            | fh    | residential |
+
+        When I route I should get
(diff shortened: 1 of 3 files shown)


features/bicycle/turn_penalty.feature
```diff
@@ -34,3 +34,39 @@ Feature: Turn Penalties
...
+            | bd    | residential |
+            | efg   | residential |
+            | fh    | residential |
+
+        When I route I should get
+            | from | to | distance  | weight | #                                         |
+            | a    | c  | 900m +- 1 | 216    | Going straight has no penalties           |
+            | a    | d  | 900m +- 1 | 220.2  | Turning right had penalties               |
+            | e    | g  | 2100m +- 4| 503.9  | Going straght has no penalties            |
+            | e    | h  | 2100m +- 4| 515.1  | Turn sharp right has even higher penalties|
```
emiltin: it's kind of hard to compare the penalties when two different geometries are used (abc/bd vs. efg/fh), and the legs have different length.

profiles/bicycle.lua
```diff
@@ -544,6 +544,7 @@ function turn_function(turn)
      turn.duration = turn.duration + profile.traffic_light_penalty
   end
   if properties.weight_name == 'cyclability' then
+      turn.weight = turn.duration
```
emiltin: i find this a bit hard to understand. if you don't set the weight, i guess the duration is used? so what happens when we set weight=duration?

TheMarex: There is no fallback here since the weight is assigned a few lines below. The main problem is that the turn penalty was only used for the duration, but we didn't also use it as basis for the turn weight (effectively disabling it).

features/bicycle/turn_penalty.feature
TheMarex: Yeah I tried to get a comparison by always adding the straight "turn" as well. In the first case the penalty is about `220 - 216 ~ 4` in the second case its `515 - 514 ~ 11`.

Issue #4027, opened 2017-05-10
Better safety penalties for the bike profile


TheMarex:
Currently our safety penalty code is rather simplistic: Every major street is assumed to be more dangerous then a local street.

Instead we should also base the penalties on which kind of cycleway is used.

- `cycleway=lane` No penalty even for major street
- `cycleway=shared_lane` bonus on small road, only slight penalty on major road
- `highway=cycleway` should get a slight bonus

/cc @almithani @danpat @emiltin

danpat:
We should also make the safety penalty graduated.  `highway=primary`, `lanes=6` should have a higher penalty than `highway=secondary`, `lanes=2`.

If there's a way to have on-street parking tagged, we should consider that too, especially if there is no cycling lane.

emiltin:
might be useful to also consider maxspeed. in general the safety will depend on the speed cars go by - which is affected by a number of tags, like highway, maxspeed, maxspeed:forward/backward, surface, etc.

cycleway=track is better than nothing.

we also need to consider all the annoying directional cases, e.g. cycleway:left, cycleway:right and cycleway=opposite_track, etc.

parked cars makes cycling safe, if the parked cars are between you and the traffic.

no busses/trucks usually makes the road safer for cyclists.

this said, i think it worth considering how much we stress safety if the profile is still supposed to compute fastest bike route.

almithani:
I think that giving a bonus to all cycleways would be a huge improvement, even without looking at # of lanes, traffic speed or types of traffic.

- `cycleway=track` and `highway=cycleway` (aka separated bike lanes) are the most "desirable" from a cyclist perspective, and since they are separated from traffic, the traffic speed and # of lanes doesn't really matter that much.
- `cycleway=lane` (aka marked bike lanes) are slightly less desirable than the above, but still WAY more desirable than a regular road
- `cycleway=shared_lane` (aka sharrows or designated cycling streets) are less desirable than both the above, but still preferred over any regular road

RE: safety vs speed - if it helps the discussion, we are getting feedback right now that the current routes that use major roads for cycle directions are not really seen as "bike routes".  In fact, the feedback says that bike routes should strongly prefer cycling features such as bike lanes and tracks.

For this reason, I think a huge improvement could be made by adjusting the bonuses and penalties on the cycleway tags.

emiltin:
[...]

PR #4046, opened 2017-05-15
Don't apply safety penalty for major streets with cycleways

TheMarex:
# Issue

First step towards #4027 

## Tasklist
 - [x] Fix tests
 - [x] review
 - [x] adjust for comments

## Requirements / Relations

### DIFF
--- a/profiles/bicycle.lua
+++ b/profiles/bicycle.lua
@@ -34,8 +34,8 @@ local profile = {
   turn_bias                 = 1.4,
 
   -- reduce the driving speed by 30% for unsafe roads
-  -- local safety_penalty            = 0.7,
-  safety_penalty            = 1.0,
+  -- only used for cyclability metric
+  safety_penalty            = 0.7,
   use_public_transport      = true,
 
   allowed_start_modes = Set {
--- a/profiles/bicycle.lua
+++ b/profiles/bicycle.lua
@@ -98,11 +98,11 @@ local profile = {
   -- only used for cyclability metric
   unsafe_highway_list = {
     primary = 0.7,
-    secondary = 0.7,
-    tertiary = 0.7,
+    secondary = 0.75,
+    tertiary = 0.8,
     primary_link = 0.7,
-    secondary_link = 0.7,
-    tertiary_link = 0.7,
+    secondary_link = 0.75,
+    tertiary_link = 0.8,
   },
 
   service_penalties = {
--- a/profiles/bicycle.lua
+++ b/profiles/bicycle.lua
@@ -94,7 +94,8 @@ local profile = {
   	'opposite_track',
   	'share_busway',
   	'sharrow',
-  	'shared'
+  	'shared',
+    'shared_lane'
   },
 
   unsafe_highway_list = Set {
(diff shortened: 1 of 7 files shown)


TheMarex:
Main problems I found stem from the handling of `opposite` vs. `opposite_lane` in combination with `cycleway:left` and `cycleway:right`.

- `cycleway{:left,:right,}=opposite_{lane,track}` only reverses a oneway, the directional tag does not matter for traffic flow. No effect on non-oneway streets. Unclear how it interacts with `oneway:bicycle`.
- `cycleway=opposite` makes a oneway street bi-directional for bicycles

This has ramifications on how we would figure out if there is an actual bicycle lane going in your direction of travel which is important for this PR.

emiltin:
i don't think cycleway:left/right=opposite_* is corect tagging. cycleway=opposite_* indicates that you can cycle in the opposite direction on the normal oneway direction. http://wiki.openstreetmap.org/wiki/Tag:cycleway%3Dopposite

emiltin:
related #4060

oxidase:
@TheMarex @emiltin please could you review the rebased PR. In addition to safety penaltie i have changed `cycleway=opposite` handling that allows no cycling in both directions as ["cyclists are permitted to travel in both directions on a road which is one-way for normal traffic"](https://wiki.openstreetmap.org/wiki/Tag:cycleway%3Dopposite)

TheMarex:
@oxidase there is still on broken edge-case with implied oneways: https://travis-ci.org/Project-OSRM/osrm-backend/jobs/248819214#L2658

profiles/bicycle.lua
```diff
@@ -94,7 +94,8 @@ local profile = {
   	'opposite_track',
   	'share_busway',
   	'sharrow',
-  	'shared'
+  	'shared',
+    'shared_lane'
```
TheMarex: Hrm this seems like it's missing a space.

emiltin:
i think we need to graduate the penalties for primary, secondary and tertiary. at least in copenhagen, there are many major bike routes placed on tertiary or secondary road. (but they have cycle lanes or tracks most of the way.)

in the cucumber test i see only cycleway=lane tested, but cycleway=track is missing. track (painted lane) is important too, if should be somewhere in between no cycleway and a lane (raised curb).

profiles/bicycle.lua
```diff
@@ -411,12 +409,10 @@ function way_function (way, result)
   if cycleway_right and profile.cycleway_tags[cycleway_right] then
     has_cycleway_right = true
   end
-  if cycleway and string.find(cycleway, "opposite") == 1 then
-    if reverse then
-      has_cycleway_right = true
-  else
-      has_cycleway_left = true
-    end
+  if cycleway and string.find(cycleway, "opposite") == 1 and not implied_oneway then
```
emiltin: i don't think it's correct to disregard contraflow on implied oneways. for example, a motorway implies oneway. cycleway=oppositve means that you can cycle in the backward direction, (since that's where the cycleway is), but not forward.

oxidase: @emiltin i have removed the corresponding commit

oxidase:
@emiltin i modified  safety penalties  https://github.com/Project-OSRM/osrm-backend/blob/de7421c562258f83c90fabd864f03ef534c4b373/profiles/bicycle.lua#L100-L105 
Do you have good numbers for `primary`, `secondary`, and `tertiary` types? Atm i would keep 0.7 for all unsafe road types as was before.

emiltin:
all i can suggest is something like 0.7, 0.75, 0.8

Issue #3835, opened 2017-03-17
Implement path unpacking cache


TheMarex:
Path unpacking takes up a considerable amount of the query time for MLD (and CH actually). A simple technique to mitigate this is to create an unpacking cache. Generally this cache would store an `EdgeID` or `source, target` pair to unpacked path in a LRU manner.

There are a lot of knobs one can tune to make this perform best some of the basic considerations:

1. What do we store in the cache: Just `EdgeIDs` that we need to turn into `PathData` again to full lists of `PathData`
2. What is a good cache size and how do we determine it
3. What kind of edges should be cached: For MLD we have a lot of choices here (e.g. only store clique arcs from certain levels), for CH there is no good mechanism to decide when a shortcut should be cached and when not.
4. Should we have a thread-local cache or a shared cache

This would hook into `SearchEngineData` (since it needs read-write and thread-safe access) and would need to be cleared on every data update.

/cc @oxidase

DennisOSRM:
this is indeed a low hanging fruit. i think i remember measurable speedups after adding this in a different but similar implementation.

DennisOSRM:
I have implemented a two-tier browse-resistant LRU cache for path unpacking as an experiment. 

The numbers generally look good. When running queries against one can see the cache become effective quickly. 

More details to share at a later time. A first experiment compares hot cache vs no cache query performance. The numbers are reflecting time given in milliseconds. Speedup is the ratio of the two corresponding numbers. 

|         | Min  | Max  | Average  | Median |
|---------|------|------|----------|--------|
| Master  | 0.21 | 8.95 | 3.13     | 3.12   |
| Cache   | 0.22 | 7.76 | 2.46     | 2.48   |
| Speedup | —    | 1.15 | 1.27     | 1.26   |

PR #4047, opened 2017-05-15
Implements Reasonable Alternative Routes for MLD

daniel-j-h:
For https://github.com/Project-OSRM/osrm-backend/issues/3905. Implements reasonable alternatives using the via-method. See the ticket for details.

- [x] Measurements: Average query speed with 1,2,3,4 alternatives
- [x] Review
- [x] Adjust for comments

### DIFF
--- a/docs/profiles.md
+++ b/docs/profiles.md
@@ -27,16 +27,16 @@ As you scroll down the file you'll see local variables, and then local functions
 
 The following global properties can be set in your profile:
 
-Attribute                     | Type     | Notes
-------------------------------|----------|----------------------------------------------------------------------------
-weight_name                   | String   | Name used in output for the routing weight property (default `'duration'`)
-weight_precision              | Unsigned | Decimal precision of edge weights (default `1`)
-left_hand_driving             | Boolean  | Are vehicles assumed to drive on the left? (used in guidance, default `false`)
-use_turn_restrictions         | Boolean  | Are turn instructions followed? (default `false`)
-continue_straight_at_waypoint | Boolean  | Must the route continue straight on at a via point, or are U-turns allowed? (default `true`)
-max_speed_for_map_matching    | Float    | Maximum vehicle speed to be assumed in matching (in m/s)
-max_turn_weight               | Float    | Maximum turn penalty weight
-force_split_edges             | Boolean  | True value forces a split of forward and backward edges of extracted ways and guarantees that `segment_function` will be called for all segments (default `false`)
+Attribute                            | Type     | Notes
+-------------------------------------|----------|----------------------------------------------------------------------------
+weight_name                          | String   | Name used in output for the routing weight property (default `'duration'`)
+weight_precision                     | Unsigned | Decimal precision of edge weights (default `1`)
+left_hand_driving                    | Boolean  | Are vehicles assumed to drive on the left? (used in guidance, default `false`)
+use_turn_restrictions                | Boolean  | Are turn instructions followed? (default `false`)
+continue_straight_at_waypoint        | Boolean  | Must the route continue straight on at a via point, or are U-turns allowed? (default `true`)
+max_speed_for_map_matching           | Float    | Maximum vehicle speed to be assumed in matching (in m/s)
+max_turn_weight                      | Float    | Maximum turn penalty weight
+force_split_edges                    | Boolean  | True value forces a split of forward and backward edges of extracted ways and guarantees that `segment_function` will be called for all segments (default `false`)
 
 ## way_function
 
--- a/include/engine/api/route_parameters.hpp
+++ b/include/engine/api/route_parameters.hpp
@@ -89,8 +89,9 @@ struct RouteParameters : public BaseParameters
                     const boost::optional<bool> continue_straight_,
                     Args... args_)
         : BaseParameters{std::forward<Args>(args_)...}, steps{steps_}, alternatives{alternatives_},
-          annotations{false}, annotations_type{AnnotationsType::None}, geometries{geometries_},
-          overview{overview_}, continue_straight{continue_straight_}
+          number_of_alternatives{alternatives_ ? 1u : 0u}, annotations{false},
+          annotations_type{AnnotationsType::None}, geometries{geometries_}, overview{overview_},
+          continue_straight{continue_straight_}
     // Once we perfectly-forward `args` (see #2990) this constructor can delegate to the one below.
     {
     }
@@ -105,7 +106,7 @@ struct RouteParameters : public BaseParameters
                     const boost::optional<bool> continue_straight_,
                     Args... args_)
         : BaseParameters{std::forward<Args>(args_)...}, steps{steps_}, alternatives{alternatives_},
-          annotations{annotations_},
+          number_of_alternatives{alternatives_ ? 1u : 0u}, annotations{annotations_},
           annotations_type{annotations_ ? AnnotationsType::All : AnnotationsType::None},
           geometries{geometries_}, overview{overview_}, continue_straight{continue_straight_}
     {
@@ -121,6 +122,7 @@ struct RouteParameters : public BaseParameters
                     const boost::optional<bool> continue_straight_,
                     Args... args_)
         : BaseParameters{std::forward<Args>(args_)...}, steps{steps_}, alternatives{alternatives_},
+          number_of_alternatives{alternatives_ ? 1u : 0u},
           annotations{annotations_ == AnnotationsType::None ? false : true},
           annotations_type{annotations_}, geometries{geometries_}, overview{overview_},
           continue_straight{continue_straight_}
@@ -128,14 +130,21 @@ struct RouteParameters : public BaseParameters
     }
 
     bool steps = false;
+    // TODO: in v6 we should remove the boolean and only keep the number parameter; for compat.
     bool alternatives = false;
+    unsigned number_of_alternatives = 0;
     bool annotations = false;
     AnnotationsType annotations_type = AnnotationsType::None;
     GeometriesType geometries = GeometriesType::Polyline;
     OverviewType overview = OverviewType::Simplified;
     boost::optional<bool> continue_straight;
 
-    bool IsValid() const { return coordinates.size() >= 2 && BaseParameters::IsValid(); }
+    bool IsValid() const
+    {
+        const auto coordinates_ok = coordinates.size() >= 2;
+        const auto base_params_ok = BaseParameters::IsValid();
+        return coordinates_ok && base_params_ok;
+    }
 };
 
 inline bool operator&(RouteParameters::AnnotationsType lhs, RouteParameters::AnnotationsType rhs)
--- a/include/engine/internal_route_result.hpp
+++ b/include/engine/internal_route_result.hpp
@@ -61,6 +61,18 @@ struct InternalRouteResult
     {
         return (leg != unpacked_path_segments.size() - 1);
     }
+
+    // Note: includes duration for turns, except for at start and end node.
+    EdgeWeight duration() const
+    {
+        EdgeWeight ret{0};
+
+        for (const auto &leg : unpacked_path_segments)
+            for (const auto &segment : leg)
+                ret += segment.duration_until_turn;
+
+        return ret;
+    }
 };
 
 struct InternalManyRoutesResult
--- a/include/engine/routing_algorithms/routing_base_mld.hpp
+++ b/include/engine/routing_algorithms/routing_base_mld.hpp
@@ -10,6 +10,12 @@
 
 #include <boost/assert.hpp>
 
+#include <algorithm>
+#include <iterator>
+#include <limits>
+#include <tuple>
+#include <vector>
+
 namespace osrm
 {
 namespace engine
@@ -24,7 +30,7 @@ namespace
 // Unrestricted search (Args is const PhantomNodes &):
 //   * use partition.GetQueryLevel to find the node query level based on source and target phantoms
 //   * allow to traverse all cells
-inline LevelID getNodeQureyLevel(const partition::MultiLevelPartitionView &partition,
+inline LevelID getNodeQueryLevel(const partition::MultiLevelPartitionView &partition,
                                  NodeID node,
                                  const PhantomNodes &phantom_nodes)
 {
@@ -49,7 +55,7 @@ inline bool checkParentCellRestriction(CellID, const PhantomNodes &) { return tr
 //   * use the fixed level for queries
 //   * check if the node cell is the same as the specified parent onr
 inline LevelID
-getNodeQureyLevel(const partition::MultiLevelPartitionView &, NodeID, LevelID level, CellID)
+getNodeQueryLevel(const partition::MultiLevelPartitionView &, NodeID, LevelID level, CellID)
 {
     return level;
 }
@@ -58,6 +64,70 @@ inline bool checkParentCellRestriction(CellID cell, LevelID, CellID parent)
 {
     return cell == parent;
 }
+}
+
+// Heaps only record for each node its predecessor ("parent") on the shortest path.
+// For re-constructing the actual path we need to trace back all parent "pointers".
+// In contrast to the CH code MLD needs to know the edges (with clique arc property).
+
+using PackedEdge = std::tuple</*from*/ NodeID, /*to*/ NodeID, /*from_clique_arc*/ bool>;
+using PackedPath = std::vector<PackedEdge>;
+
+template <bool DIRECTION, typename OutIter>
+inline void retrievePackedPathFromSingleHeap(const SearchEngineData<Algorithm>::QueryHeap &heap,
+                                             const NodeID middle,
+                                             OutIter out)
+{
+    NodeID current = middle;
+    NodeID parent = heap.GetData(current).parent;
+
+    while (current != parent)
+    {
+        const auto &data = heap.GetData(current);
+
+        if (DIRECTION == FORWARD_DIRECTION)
+        {
+            *out = std::make_tuple(parent, current, data.from_clique_arc);
+            ++out;
+        }
+        else if (DIRECTION == REVERSE_DIRECTION)
+        {
+            *out = std::make_tuple(current, parent, data.from_clique_arc);
+            ++out;
+        }
+
+        current = parent;
+        parent = heap.GetData(parent).parent;
+    }
+}
+
+template <bool DIRECTION>
+inline PackedPath
+retrievePackedPathFromSingleHeap(const SearchEngineData<Algorithm>::QueryHeap &heap,
+                                 const NodeID middle)
+{
+    PackedPath packed_path;
+    retrievePackedPathFromSingleHeap<DIRECTION>(heap, middle, std::back_inserter(packed_path));
+    return packed_path;
+}
+
+// Trace path from middle to start in the forward search space (in reverse)
+// and from middle to end in the reverse search space. Middle connects paths.
+
+inline PackedPath
+retrievePackedPathFromHeap(const SearchEngineData<Algorithm>::QueryHeap &forward_heap,
+                           const SearchEngineData<Algorithm>::QueryHeap &reverse_heap,
+                           const NodeID middle)
+{
+    // Retrieve start -> middle. Is in reverse order since tracing back starts from middle.
+    auto packed_path = retrievePackedPathFromSingleHeap<FORWARD_DIRECTION>(forward_heap, middle);
+    std::reverse(begin(packed_path), end(packed_path));
+
+    // Retrieve middle -> end. Is already in correct order, tracing starts from middle.
+    auto into = std::back_inserter(packed_path);
+    retrievePackedPathFromSingleHeap<REVERSE_DIRECTION>(reverse_heap, middle, into);
+
+    return packed_path;
 }
 
 template <bool DIRECTION, typename... Args>
@@ -96,7 +166,7 @@ void routingStep(const datafacade::ContiguousInternalMemoryDataFacade<Algorithm>
         }
     }
 
-    const auto level = getNodeQureyLevel(partition, node, args...);
+    const auto level = getNodeQueryLevel(partition, node, args...);
 
     if (level >= 1 && !forward_heap.GetData(node).from_clique_arc)
     {
@@ -112,6 +182,7 @@ void routingStep(const datafacade::ContiguousInternalMemoryDataFacade<Algorithm>
                 if (shortcut_weight != INVALID_EDGE_WEIGHT && node != to)
                 {
                     const EdgeWeight to_weight = weight + shortcut_weight;
+                    BOOST_ASSERT(to_weight >= weight);
                     if (!forward_heap.WasInserted(to))
                     {
                         forward_heap.Insert(to, to_weight, {node, true});
@@ -137,6 +208,7 @@ void routingStep(const datafacade::ContiguousInternalMemoryDataFacade<Algorithm>
                 if (shortcut_weight != INVALID_EDGE_WEIGHT && node != to)
                 {
                     const EdgeWeight to_weight = weight + shortcut_weight;
+                    BOOST_ASSERT(to_weight >= weight);
                     if (!forward_heap.WasInserted(to))
                     {
                         forward_heap.Insert(to, to_weight, {node, true});
@@ -179,16 +251,24 @@ void routingStep(const datafacade::ContiguousInternalMemoryDataFacade<Algorithm>
     }
 }
 
+// With (s, middle, t) we trace back the paths middle -> s and middle -> t.
+// This gives us a packed path (node ids) from the base graph around s and t,
+// and overlay node ids otherwise. We then have to unpack the overlay clique
+// edges by recursively descending unpacking the path down to the base graph.
+
+using UnpackedNodes = std::vector<NodeID>;
+using UnpackedEdges = std::vector<EdgeID>;
+using UnpackedPath = std::tuple<EdgeWeight, UnpackedNodes, UnpackedEdges>;
+
 template <typename... Args>
-std::tuple<EdgeWeight, std::vector<NodeID>, std::vector<EdgeID>>
-search(SearchEngineData<Algorithm> &engine_working_data,
-       const datafacade::ContiguousInternalMemoryDataFacade<Algorithm> &facade,
-       SearchEngineData<Algorithm>::QueryHeap &forward_heap,
-       SearchEngineData<Algorithm>::QueryHeap &reverse_heap,
-       const bool force_loop_forward,
-       const bool force_loop_reverse,
-       EdgeWeight weight_upper_bound,
-       Args... args)
+UnpackedPath search(SearchEngineData<Algorithm> &engine_working_data,
+                    const datafacade::ContiguousInternalMemoryDataFacade<Algorithm> &facade,
+                    SearchEngineData<Algorithm>::QueryHeap &forward_heap,
+                    SearchEngineData<Algorithm>::QueryHeap &reverse_heap,
+                    const bool force_loop_forward,
+                    const bool force_loop_reverse,
+                    EdgeWeight weight_upper_bound,
+                    Args... args)
 {
     if (forward_heap.Empty() || reverse_heap.Empty())
     {
@@ -242,27 +322,12 @@ search(SearchEngineData<Algorithm> &engine_working_data,
         return std::make_tuple(INVALID_EDGE_WEIGHT, std::vector<NodeID>(), std::vector<EdgeID>());
     }
 
-    // Get packed path as edges {from node ID, to node ID, edge ID}
-    std::vector<std::tuple<NodeID, NodeID, bool>> packed_path;
-    NodeID current_node = middle, parent_node = forward_heap.GetData(middle).parent;
-    while (parent_node != current_node)
-    {
-        const auto &data = forward_heap.GetData(current_node);
-        packed_path.push_back(std::make_tuple(parent_node, current_node, data.from_clique_arc));
-        current_node = parent_node;
-        parent_node = forward_heap.GetData(parent_node).parent;
-    }
-    std::reverse(std::begin(packed_path), std::end(packed_path));
-    const NodeID source_node = current_node;
+    // Get packed path as edges {from node ID, to node ID, from_clique_arc}
+    auto packed_path = retrievePackedPathFromHeap(forward_heap, reverse_heap, middle);
 
-    current_node = middle, parent_node = reverse_heap.GetData(middle).parent;
-    while (parent_node != current_node)
-    {
-        const auto &data = reverse_heap.GetData(current_node);
-        packed_path.push_back(std::make_tuple(current_node, parent_node, data.from_clique_arc));
-        current_node = parent_node;
-        parent_node = reverse_heap.GetData(parent_node).parent;
-    }
+    // Beware the edge case when start, middle, end are all the same.
+    // In this case we return a single node, no edges. We also don't unpack.
+    const NodeID source_node = !packed_path.empty() ? std::get<0>(packed_path.front()) : middle;
 
     // Unpack path
     std::vector<NodeID> unpacked_nodes;
@@ -271,6 +336,7 @@ search(SearchEngineData<Algorithm> &engine_working_data,
     unpacked_edges.reserve(packed_path.size());
 
     unpacked_nodes.push_back(source_node);
+
     for (auto const &packed_edge : packed_path)
     {
         NodeID source, target;
@@ -283,7 +349,7 @@ search(SearchEngineData<Algorithm> &engine_working_data,
         }
         else
         { // an overlay graph edge
-            LevelID level = getNodeQureyLevel(partition, source, args...);
+            LevelID level = getNodeQueryLevel(partition, source, args...);
             CellID parent_cell_id = partition.GetCell(level, source);
             BOOST_ASSERT(parent_cell_id == partition.GetCell(level, target));
 
--- a/include/extractor/profile_properties.hpp
+++ b/include/extractor/profile_properties.hpp
@@ -8,6 +8,7 @@
 #include <algorithm>
 #include <boost/assert.hpp>
 #include <boost/numeric/conversion/cast.hpp>
+#include <cstdint>
 
 namespace osrm
 {
@@ -109,7 +110,6 @@ struct ProfileProperties
     std::array<char[MAX_CLASS_NAME_LENGTH + 1], MAX_CLASS_INDEX + 1> class_names;
     unsigned weight_precision = 1;
     bool force_split_edges = false;
-
     bool call_tagless_node_function = true;
 };
 }
--- a/profiles/car.lua
+++ b/profiles/car.lua
@@ -9,16 +9,16 @@ local Handlers = require("lib/handlers")
 local next = next       -- bind to local for speed
 
 -- set profile properties
-properties.max_speed_for_map_matching      = 180/3.6 -- 180kmph -> m/s
-properties.use_turn_restrictions           = true
-properties.continue_straight_at_waypoint   = true
-properties.left_hand_driving               = false
+properties.max_speed_for_map_matching           = 180/3.6 -- 180kmph -> m/s
+properties.use_turn_restrictions                = true
+properties.continue_straight_at_waypoint        = true
+properties.left_hand_driving                    = false
 -- For routing based on duration, but weighted for preferring certain roads
-properties.weight_name                     = 'routability'
+properties.weight_name                          = 'routability'
 -- For shortest duration without penalties for accessibility
---properties.weight_name                     = 'duration'
+--properties.weight_name                        = 'duration'
 -- For shortest distance without penalties for accessibility
---properties.weight_name                     = 'distance'
+--properties.weight_name                        = 'distance'
 
 -- Set to true if you need to call the node_function for every node.
 -- Generally can be left as false to avoid unnecessary Lua calls
--- a/src/engine/routing_algorithms/alternative_path_mld.cpp
+++ b/src/engine/routing_algorithms/alternative_path_mld.cpp
@@ -1,0 +1,936 @@
+#include "engine/routing_algorithms/alternative_path.hpp"
+#include "engine/routing_algorithms/routing_base_mld.hpp"
+
+#include "util/static_assert.hpp"
+
+#include <boost/assert.hpp>
+
+#include <algorithm>
+#include <iterator>
+#include <memory>
+#include <type_traits>
+#include <unordered_set>
+#include <utility>
+#include <vector>
+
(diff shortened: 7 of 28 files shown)


daniel-j-h:
Tasks (not in order; also I will split off into sub-tasks as I see fit):
- [x] Make pipeline recognize alternatives feature on MLD
- [x] Lift restriction to only ever have zero or one alternative (done separately in https://github.com/Project-OSRM/osrm-backend/pull/4035 already in master)
- [x] Let search spaces overlap and intersect nodes as via-candidates
- [ ] Pre-filter and rank candidates with heuristics (requires some 🤔)
- [ ] Unpack paths and rank unpacked alternatives
- [ ] Get alternatives into the API response
- [ ] Get multiple alternatives to show up in the frontend

daniel-j-h:
We have the full set of all possible alternatives available now. From the search space overlap, candidate generation based on boundary nodes and base graph nodes at the bottom, path reconstruction, path unpacking in the backend all through the pipeline to the response. Screenshot from local frontend:

map

    make -C test/data
    ./osrm-routed --algorithm mld ../test/data/mld/monaco.osrm

---

The next task will be filtering via-candidates as soon as possible (for performance reasons) based on
- stretch - how much is the alternative longer than the primary route
- sharing - how many ways does the alternative share with the primary route
- local optimality - sub-paths up to certain length are themselves shortest paths

---

For the stretch-test @MoKob just brought up a good point I haven't thought about.

For restricted areas (think: `access=destination`) we apply a "high" weight penalty to the turn onto the restricted way (idea was blogged about [in this diary post](https://www.openstreetmap.org/user/happygo/diary/40564)).

[Here](https://github.com/Project-OSRM/osrm-backend/blob/403db7cc84f767bf77cd9cab98353b93a507b17f/profiles/car.lua#L414-L417) is where we apply the penalty weight and [here](https://github.com/Project-OSRM/osrm-backend/blob/403db7cc84f767bf77cd9cab98353b93a507b17f/profiles/car.lua#L33-L35) is where it is defined.

In the case of restricted areas on the paths the stretch-test based on raw weights suddenly makes no sense anymore. For example a route that is longer in terms of weight can be of almost the same duration but passing a restricted area instead. In the case when start and / or target are _inside_ a restricted area both our primary route as well as all alternatives have at least the start and / or target penalty weight on them. We probably also want to filter via candidates which are inside a restricted area to begin with.

If we could access the high turn penalty from the C++ side we could simply subtract it from the route's weights. But we can't since the restricted penalty is only set in the Lua profile and is not bound in our scripting environment. Binding it would be a breaking profile API change. Not binding it means I can't do reasonable stretch-tests on weights - at least not for the case when restricted areas are involved.

danpat:
For those following along - the "stretch test" means "how much longer is the alternative than the shortest/fastest/least-weight path".  A stretch factor of 1.2 means a path is 20% longer than the shortest ("longer" is in terms of the metric being used for shortest-path routing, in our case, `weight`).

@daniel-j-h Given that we're applying these weights in order to *avoid* these restricted ways, how does this cause problems when using weight to calculate stretch?  If an alternative enters a restricted way, it will have a high weight, and the stretch will be large, possibly causing it to be pruned.  Is this not desirable behaviour?  If we're using weights to avoid private roads, why would we want alternatives that *use* private roads?

It seems to me that we want all our other metrics (avoid private roads, etc) to hold for alternatives as well, and using weight to calculate stretch would generally achieve this.  Am I missing something?

MoKob:
@danpat when travelling a-b, there are two valid locations for private roads (begin/end, if you are starting/ending on it). In our case, we can get intermediate vertices on private roads, if we aren't careful.

Right now we add 50 mins per entry/exit. Now if you have a long path that has a privat path as a shortcut, you could end up taking that part (since you only find a candidate in that area and can't see whether it is private or not). If the road around the (e.g.) private bridge is long enough, suddenly we allow paying that penalty in the middle.

Its all about avoiding private areas, not allowing them. But stretch alone is not guaranteed to work out. It will in most cases, since military bases or other private roads should hopefully not be the easiest path of reaching an area, but in any remote areas it could well be the case. It can, of course, already happen with the shortest path itself. By allowing stretch, we reduce the effective penalty of the already rather small penalty when it comes to remote areas.

daniel-j-h:
For the record this is already a problem in the current impl. (see [here](https://github.com/Project-OSRM/osrm-backend/blob/5.7/src/engine/routing_algorithms/alternative_path.cpp#L709-L735)) and not specific to this changeset.

daniel-j-h:
@danpat here's an example from Berlin where I filter candidate routes by at most `1.05` times longer than the shortest path. And the shortest path ending in a restricted area. Which means the shortest path has a high weight applied to it - we suddenly filter out a lot less candidate paths then we should.

stretch

danpat:
I see, so because the penalty is so large in comparison to the total route weight, and because it's mandatory to take the penalty turn to access the endpoint, doing a simple stretch comparison of all alternatives makes them appear to be quite similar (the weight penalty dominates the total route weight).

daniel-j-h:
I just implemented a rough sharing-heuristic (how many ways does the alternative share with the primary route) making use of the cell structure we have at our hands from the partitioner.

In contrast to the techniques described in comments above this heuristic already needs the packed paths from the heaps. I then compute the similarity based on the cell ids for the nodes on the packed path.

sharing0
sharing1

At the moment I'm still comparing `sharing(shortest path, alternative)` only - what we eventually want for returning multiple alternatives is comparing the sharing between _all routes_.

---

There are a few open questions I don't have answers to yet:
- On which level does it make sense to compute cell sharing on? Should sharing be a linear combination based on level and cell sharing on each level?
- Do we need to scale sharing with edge weights? Is sharing based on cells only good enough?

daniel-j-h:
Comparing sharing between all combination of routes is implemented by now. The simple approach of taking the pre-filtered via candidates ranking them by weight and then filtering based on the cell structure for the packed paths on the first level already gives good looking results:

alternatives

([click for video](https://gfycat.com/JollyOffbeatFlounder))

---

What's next is local optimality, investigating cell structure sharing on multiple levels, experimenting with parameters and trying to solve the open questions from above.

oxidase:
@daniel-j-h i have pushed a fix that wraps the lambda into a copyable `std::function` fell free to kick out if it is nok.
Related issue https://github.com/osmcode/osmium-tool/issues/7

daniel-j-h:
Update: I just pushed a first pass at local optimality. The overall problem we try to solve is as follows: we could pick a via node for which the alternative path would make an unreasonable detour. Think of picking a via node which leads the the path leaving the highway and then immediately going onto it again.

From the tests I did so far _without local optimality_ I have the impression that picking via nodes using the mld search stepping through cell boundary nodes and filtering via the cell structure for sharing already prevents a lot of unreasonable local detours. Similarly, an idea worth exploring is only picking via nodes on "important" roads and discarding via nodes on "unimportant" roads for further filtering.

Here's a rough sketch for local optimality: with the primary route found we pick via nodes in the forward and reverse search space overlap. For these via nodes we then do a stretch- and sharing-test as described above. To guarantee for local optimality around the via nodes we take a look at the two sub-paths around a via node and use the forward and reverse heaps to check if they are optimal.

local-optimality

Have a look at the alternative route above the shortest path here. Around the area we pick via nodes from it totally looks like an unreasonable detour. But the local optimality filter is already on so we're sure it's a shortest sub-path. We can confirm this by visualizing the sub-path ourselves:

local-optimality2

Success!

daniel-j-h:
I started to implement filtering on unpacked paths. Not sure how many paths I can unpack in the time budget yet - I guess it's in the low two digit area (needs benchmarking).

alt

([click for webm video](https://gfycat.com/ThoughtfulAgonizingBettong))

---

Next tasks here are: running out filtering and ranking on unpacked paths. Cleaning up what I implemented during the last days. Playing around with all the tune-able knobs that made it in. Resolving open questions raised in this ticket or flagging them for future improvements.

src/engine/routing_algorithms/alternative_path_mld.cpp
```diff
@@ -0,0 +1,786 @@
...
+    UnpackedNodes nodes;
+    UnpackedEdges edges;
+};
+
+// Filters candidates which are on not unique.
+// Returns an iterator to the uniquified range's new end.
+// Note: mutates the range in-place invalidating iterators.
+template <typename RandIt> RandIt filterViaCandidatesByUniqueNodeIds(RandIt first, RandIt last)
+{
+    util::static_assert_iter_category<RandIt, std::random_access_iterator_tag>();
```
MoKob: any reasons for this check? std::unique should already complain if the iterators don't match, right? If they don't work with sort/unique, we can't use them here. Other restrictions are not necessary, or am I missing a benefit.

The same with WeightedViaNode. What if I decide to do a wrapper and want to still use the function. These asserts do not really add benefit but only reduce, flexibility to me.

src/engine/routing_algorithms/alternative_path_mld.cpp
```diff
@@ -0,0 +1,786 @@
...
+    (void)last;
+    (void)facade;
+
+    return last;
+}
+
+// Filters candidates with much higher weight than the primary route. Mutates range in-place.
+// Returns an iterator to the filtered range's new end.
+template <typename RandIt>
+RandIt filterViaCandidatesByStretch(RandIt first, RandIt last, EdgeWeight weight)
```
MoKob: this kinda indicates to me that the stretch-test would be OK after this test. We test a global stretch, but we want to test the local stretch in the end. The total weight is only a very bad indicator:

Lets take `a --- b --- c  --- d` and `a --- b --- e --- d`, `|abcd| = 1`, |abed| = (1+0.25)`.

If `|bcd| == 0.8 * |abcd|`, the local stretch of `bed` vs `bcd` is `((1+0.25) - 0.2) / 0.8 = 1.31`, for  `|bcd| == 0.5 * |abcd|` you already get a stretch of `1.5`.  In total your stretch gets scaled here by the amount of sharing, which is an undesired effect.

src/engine/routing_algorithms/alternative_path_mld.cpp
```diff
@@ -0,0 +1,786 @@
...
+
+    const auto over_weight_limit = [=](const auto via) {
+        return via.weight > stretch_weight_limit;
+    };
+
+    return std::remove_if(first, last, over_weight_limit);
+}
+
+// The packed paths' sharing in [0, 1] for no sharing and equality, respectively.
+inline double normalizedPackedPathSharing(const Partition &partition,
```
MoKob: how do you feel about making clear in the function name, that all of these are only heuristics on the packed path?

src/engine/routing_algorithms/alternative_path_mld.cpp
```diff
@@ -0,0 +1,786 @@
...
+                                          RandIt last)
+{
+    util::static_assert_iter_category<RandIt, std::random_access_iterator_tag>();
+    util::static_assert_iter_value<RandIt, WeightedViaNodePackedPath>();
+
+    // Check sub-path optimality on alternative path crossing the via node candidate.
+    //
+    //   s - - - v - - - t  our packed path made up of (from, to) edges
+    //        |--|--|       sub-paths "left" and "right" of v based on threshold
+    //        f  v  l       nodes in [v,f] and in [v, l] have to match predecessor in heaps
```
MoKob: without knowing what I explained to you, I would have had difficulties following this explanation

src/engine/routing_algorithms/alternative_path_mld.cpp
```diff
@@ -0,0 +1,786 @@
...
+        // we may not have seen the node's parent in the dual heap.
+        if (!forward_heap.WasInserted(parent_in_reverse))
+            return true;
+        if (!reverse_heap.WasInserted(parent_in_forward))
+            return true;
+
+        const auto parent_parent_in_forward = forward_heap.GetData(parent_in_reverse).parent;
+        const auto parent_parent_in_reverse = reverse_heap.GetData(parent_in_forward).parent;
+
+        const auto plateaux = parent_parent_in_forward == via && parent_parent_in_reverse == via;
```
MoKob: this plateaux check should consider length of the paths to ensure that this is not a minimum plateaux.

src/engine/routing_algorithms/alternative_path_mld.cpp
```diff
@@ -0,0 +1,786 @@
...
+
+        return !plateaux;
+    };
+
+    return std::remove_if(first, last, is_not_locally_optimal);
+}
+
+// The unpacked paths' sharing in [0, 1] for no sharing and equality, respectively.
+inline double normalizedUnpackedPathSharing(const UnpackedNodes &lhs, const UnpackedNodes &rhs)
```
MoKob: As you already said to me, all of these should probably use weights, normalised to not include access penalties

src/engine/routing_algorithms/alternative_path_mld.cpp
```diff
@@ -0,0 +1,786 @@
...
+
+    const bool force_loop_forward = needsLoopForward(phantom_node_pair);
+    const bool force_loop_backward = needsLoopBackwards(phantom_node_pair);
+
+    EdgeWeight forward_heap_min = forward_heap.MinKey();
+    EdgeWeight reverse_heap_min = reverse_heap.MinKey();
+
+    while (forward_heap.Size() + reverse_heap.Size() > 0)
+    {
+        if (shortest_path_weight != INVALID_EDGE_WEIGHT)
```
MoKob: Do you think we could make this process less brittle by not assigning/updating/re-using weights? Overlap weight is assigned here for a single check, before it is cleared to Invalid, updated in the routing step and again invalidated. This can get somewhat confusing.

src/engine/routing_algorithms/alternative_path_mld.cpp
```diff
@@ -0,0 +1,786 @@
...
+
+            if (overlap_weight != INVALID_EDGE_WEIGHT && overlap_via != SPECIAL_NODEID)
+            {
+                if (shortest_path_via != SPECIAL_NODEID)
+                {
+                    candidate_vias.push_back(WeightedViaNode{overlap_via, overlap_weight});
+                }
+                else
+                {
+                    shortest_path_via = overlap_via;
```
MoKob: `shortest_path_via` and `shortest_path_weight` should stay in sync, right? Why do we keep track of them at all during the algorithm and not simply use the minimum-weight via candidate?

src/engine/routing_algorithms/alternative_path_mld.cpp
```diff
@@ -0,0 +1,786 @@
...
+    std::cout << ">>> shortest path weight: " << shortest_path_weight << std::endl;
+    std::cout << ">>> number of candidates: " << candidate_vias.size() << std::endl;
+
+    // Filters via candidate nodes with heuristics
+
+    // Note: filter pipeline below only makes range smaller; no need to erase items
+    // from the vector when we can mutate in-place and for filtering adjust iterators.
+    auto it = end(candidate_vias);
+
+    it = filterViaCandidatesByUniqueNodeIds(begin(candidate_vias), it);
```
MoKob: depending on the size, we could potentially do better with `n_th` element rankings, just as an idea for later optimisations.

src/engine/routing_algorithms/alternative_path_mld.cpp
```diff
@@ -0,0 +1,786 @@
...
+        {
+            alternative_paths_last =
+                filterPackedPathsByCellSharing(weighted_packed_paths[j],
+                                               partition,
+                                               begin(weighted_packed_paths) + i,
+                                               alternative_paths_last);
+        }
+    }
+
+    number_of_filtered_alternative_paths = std::min(
```
MoKob: we should be careful for long routes. We could end up having multiple very similar alternatives. The ranking should prefer routes that are different over routes that are similar and have low stretch to avoid unpacking only very similar routes.
Maybe select the the top `k` candidates from multiple criteria?

src/engine/routing_algorithms/alternative_path_mld.cpp
```diff
@@ -0,0 +1,786 @@
...
+// Apply heuristic to evaluate alternative route based on stretch, overlap, how reasonable it is.
+//
+// For MLD specifically we can pull off some tricks to make evaluating alternatives fast:
+//   Only consider (s, c, t) with c border vertex: re-use MLD search steps.
+//   Add meta data to border vertices: consider (s, c, t) only when c is e.g. on a highway.
+//   Prune based on vertex cell id
+//
+// https://github.com/Project-OSRM/osrm-backend/issues/3905
+InternalManyRoutesResult
+alternativePathSearch(SearchEngineData<Algorithm> &search_engine_data,
```
MoKob: This is becoming a 👹  of a function. Do you think it could be split into multiple sub-functions?

src/engine/routing_algorithms/alternative_path_mld.cpp
```diff
@@ -0,0 +1,786 @@
...
+    }
+
+    number_of_filtered_alternative_paths = std::min(
+        static_cast<std::size_t>(kMaxAlternativesToUnpack),
+        static_cast<std::size_t>(alternative_paths_last - (begin(weighted_packed_paths) + 1)));
+
+    // ^ refactor
+
+    const auto paths_first = begin(weighted_packed_paths);
+    const auto paths_last = begin(weighted_packed_paths) + 1 + number_of_filtered_alternative_paths;
```
MoKob: Personal preference: last is accessible, end isn't. You are using the `begin/end` paradigm here, right?

src/engine/routing_algorithms/alternative_path_mld.cpp
```diff
@@ -0,0 +1,786 @@
...
+        }
+    }
+
+    number_of_filtered_alternative_paths = std::min(
+        static_cast<std::size_t>(kMaxAlternativesToUnpack),
+        static_cast<std::size_t>(alternative_paths_last - (begin(weighted_packed_paths) + 1)));
+
+    // ^ refactor
+
+    const auto paths_first = begin(weighted_packed_paths);
```
MoKob: is the first excluding the shortest path?

src/engine/routing_algorithms/alternative_path_mld.cpp
```diff
@@ -0,0 +1,786 @@
...
+    for (std::size_t i = 1; i < unpacked_paths.size(); ++i)
+    {
+        for (std::size_t j = 0; j < i; ++j)
+        {
+            unpacked_paths_last = filterUnpackedPathsBySharing(
+                unpacked_paths[i], begin(unpacked_paths) + i, unpacked_paths_last);
+        }
+    }
+
+    // Todo: rank a last time by weight and sharing here, then just take first n
```
MoKob: what about stretch, which depends on the sharing (as outlined above).

Also: alpha/gamma/epsilon should depend on the shape of the detour to allow more tuning.

src/engine/routing_algorithms/alternative_path_mld.cpp
```diff
@@ -0,0 +1,786 @@
...
+
+// Implementation details
+namespace
+{
+
+// Alternative paths candidate via nodes are taken from overlapping search spaces.
+// Overlapping by a third guarantees us taking candidate nodes "from the middle".
+const constexpr auto kSearchSpaceOverlapFactor = 1.66;
+// Maximum number of alternative paths to return.
+const constexpr auto kMaxAlternatives = 3;
```
daniel-j-h: Not sure if it should be a query-time parameter but I can definitely see the point in exposing some knobs here in the profiles.

src/engine/routing_algorithms/alternative_path_mld.cpp
daniel-j-h: The static asserts I use here and below are for erroring out early and with a nice line such as

    RandIt is not a random access iterator

or

    The value RandIt points to is not a type T

if you mix up the types here you will get a huge amount of garbage errors from within the stdlib. 

The sort call requires at least a `.nodes` property on the types. But writing traits for these things is a pita. This will be solved once we have Concepts (in C++3* maybe ..).

src/engine/routing_algorithms/alternative_path_mld.cpp
daniel-j-h: No `weighted_packed_paths.front()` is the shortest path, pushed in line 600 above

daniel-j-h:
Updating with the current status after I got pulled into some other projects.

Tasks left:
- [ ] Take way length into account when checking for local optimality. At the moment we do a single hop in both directions which might not be good enough in some edge-cases where partition cells meet. We also compute local optimality based on packed paths, because unpacking destroys our heaps. Should we improve on this?
- [ ] Scale sharing with weights. Do this for cell sharing and / or unpacked path sharing only. Maybe think about doing this on durations because of the penalty weight problem.
- [ ] Make parameters tune-able during query-time. At the moment we expose two knobs in the profiles only: number of alternatives and number of candidate paths to unpack.

Open Questions:
- We use weight penalties and the penalty for turning onto a restricted way is huge (think: `access=destination`). This skews the weight-based stretch test when the route e.g. ends in an area only accessible by a restricted way - see https://github.com/Project-OSRM/osrm-backend/pull/4047#issuecomment-302409619
- We compute cell sharing as a proxy for the route's way sharing. At the moment we use cells on the first level which _seems_ to work well in practice from what I've seen so far. Should sharing be a linear combination based on all levels? - see https://github.com/Project-OSRM/osrm-backend/pull/4047#issuecomment-303123687
- Do we need an unpacking cache to either boost performance or allow for more candidates which in turn leads to higher quality? - see https://github.com/Project-OSRM/osrm-backend/issues/3835
- Do we need a stack allocator for the handful of small stdlib vectors we create? Or are a few memory allocations per candidate path acceptable performance-wise?
- Should we filter via candidate nodes by road importance heuristic? We already have the road classification data in pre-processing from guidance. We'd need to serialize it out and then read it back in again in the facades.

docs/profiles.md
```diff
@@ -27,16 +27,18 @@ As you scroll down the file you'll see local variables, and then local functions
...
+weight_name                          | String   | Name used in output for the routing weight property (default `'duration'`)
+weight_precision                     | Unsigned | Decimal precision of edge weights (default `1`)
+left_hand_driving                    | Boolean  | Are vehicles assumed to drive on the left? (used in guidance, default `false`)
+use_turn_restrictions                | Boolean  | Are turn instructions followed? (default `false`)
+continue_straight_at_waypoint        | Boolean  | Must the route continue straight on at a via point, or are U-turns allowed? (default `true`)
+max_speed_for_map_matching           | Float    | Maximum vehicle speed to be assumed in matching (in m/s)
+max_turn_weight                      | Float    | Maximum turn penalty weight
+force_split_edges                    | Boolean  | True value forces a split of forward and backward edges of extracted ways and guarantees that `segment_function` will be called for all segments (default `false`)
+max_number_of_alternatives           | Unsigned | Maximum number of alternative routes to return. Note: there is no guarantee for a minimum. (default `3`)
```
TheMarex: I don't think we should expose these settings in the `ProfileProperties`. The `ProfielProperties` typically store setting that determine how the pre-processing is run and have implications on the query later on.

`max_number_of_alternatives` should be handled similarly to our `max_number_of_coordinates` options.

docs/profiles.md
```diff
@@ -27,16 +27,18 @@ As you scroll down the file you'll see local variables, and then local functions
...
+max_number_of_alternatives_to_unpack | Unsigned | Maximum number of alternative paths to unpack and run high-quality checks on. Set lower for higher performance but also lower chance to find alternative routes. (default `15`)
```
TheMarex: This should not even be user-configurable IMHO. People that need this level of fine tuning can be expected to patch and recompile it themselves. This should be an internal constant somewhere in the alternatives plugin.

include/engine/api/route_parameters.hpp
```diff
@@ -121,21 +122,29 @@ struct RouteParameters : public BaseParameters
...
         : BaseParameters{std::forward<Args>(args_)...}, steps{steps_}, alternatives{alternatives_},
+          number_of_alternatives{alternatives_ ? 1u : 0u},
           annotations{annotations_ == AnnotationsType::None ? false : true},
           annotations_type{annotations_}, geometries{geometries_}, overview{overview_},
           continue_straight{continue_straight_}
     {
     }
 
     bool steps = false;
+    // TODO: in v6 we should remove the boolean and only keep the number parameter; for compat.
```
TheMarex: Please add this to the 6.0 ticket that tracks API changes to be done in 6.0.

include/engine/datafacade/datafacade_base.hpp
```diff
@@ -164,6 +164,9 @@ class BaseDataFacade
 
     virtual double GetMapMatchingMaxSpeed() const = 0;
 
+    virtual std::size_t GetMaxNumberOfAlternatives() const = 0;
```
TheMarex: See above, should not be exposed in the datafacade. Adding this comment as a reminder to remove the interface.

include/engine/internal_route_result.hpp
```diff
@@ -59,6 +59,20 @@ struct InternalRouteResult
...
+    // Note: this is the accumulated duration without turn penalties factored in.
+    // We already have the duration attached to edges and it may be good enough
+    // for heuristics based on way durations (e.g. used in the mld alternatives).
+    EdgeWeight duration_without_turns() const
+    {
+        EdgeWeight ret{0};
+
+        for (const auto &leg : unpacked_path_segments)
+            for (const auto &segment : leg)
+                ret += segment.duration_until_turn;
```
TheMarex: This will include the turn penalty. We only ever explicitly de-compose it for the start and end node.

include/engine/routing_algorithms/routing_base_mld.hpp
```diff
@@ -242,27 +322,12 @@ search(SearchEngineData<Algorithm> &engine_working_data,
...
-    while (parent_node != current_node)
-    {
-        const auto &data = forward_heap.GetData(current_node);
-        packed_path.push_back(std::make_tuple(parent_node, current_node, data.from_clique_arc));
-        current_node = parent_node;
-        parent_node = forward_heap.GetData(parent_node).parent;
-    }
-    std::reverse(std::begin(packed_path), std::end(packed_path));
-    const NodeID source_node = current_node;
+    // Get packed path as edges {from node ID, to node ID, from_clique_arc}
```
TheMarex: Nice refactor!

include/extractor/profile_properties.hpp
```diff
@@ -88,8 +89,12 @@ struct ProfileProperties
     char weight_name[MAX_WEIGHT_NAME_LENGTH + 1];
     unsigned weight_precision = 1;
     bool force_split_edges = false;
-
     bool call_tagless_node_function = true;
+
+    // Maximum number of alternative paths to return.
+    std::size_t max_number_of_alternatives = 3;
```
TheMarex: Reminder: These properties should be moved.

include/util/packed_vector.hpp
```diff
@@ -108,7 +108,7 @@ template <typename T, std::size_t Bits, storage::Ownership Ownership> class Pack
     // C++14 does not allow operator[] to be constexpr, this is fixed in C++17.
     static /* constexpr */ std::array<WordT, BLOCK_ELEMENTS> initialize_lower_mask()
     {
-        std::array<WordT, BLOCK_ELEMENTS> lower_mask{};
+        std::array<WordT, BLOCK_ELEMENTS> lower_mask;
```
TheMarex: These changes will go away after a rebase, right?

profiles/car.lua
```diff
@@ -9,16 +9,18 @@ local Handlers = require("lib/handlers")
 local next = next       -- bind to local for speed
 
 -- set profile properties
-properties.max_speed_for_map_matching      = 180/3.6 -- 180kmph -> m/s
-properties.use_turn_restrictions           = true
-properties.continue_straight_at_waypoint   = true
-properties.left_hand_driving               = false
+properties.max_number_of_alternatives           = 3
```
TheMarex: Reminder: These properties should be moved.

src/extractor/scripting_environment_lua.cpp
```diff
@@ -251,7 +251,11 @@ void Sol2ScriptingEnvironment::InitContext(LuaScriptingContext &context)
         "force_split_edges",
         &ProfileProperties::force_split_edges,
         "call_tagless_node_function",
-        &ProfileProperties::call_tagless_node_function);
+        &ProfileProperties::call_tagless_node_function,
```
TheMarex: Reminder: These properties should be moved.

src/engine/routing_algorithms/alternative_path_mld.cpp
```diff
@@ -0,0 +1,936 @@
...
+    BOOST_ASSERT(difference >= 0.);
+    BOOST_ASSERT(difference <= 1.);
+
+    return 1. - difference;
+}
+
+// Filters packed paths with similar cells compared to the primary route. Mutates range in-place.
+// Returns an iterator to the filtered range's new end.
+template <typename RandIt>
+RandIt filterPackedPathsByCellSharingHeuristic(const WeightedViaNodePackedPath &path,
```
TheMarex: Instead of `path` this could be `primary_path` or `shortest_path`, right?

src/engine/routing_algorithms/alternative_path_mld.cpp
```diff
@@ -0,0 +1,936 @@
...
+        BOOST_ASSERT(main_heap.WasInserted(node));
+        auto const parent_node = main_heap.GetData(node).parent;
+        return side_heap.WasInserted(parent_node) && side_heap.GetData(parent_node).parent == node;
+    };
+
+    // a plateaux is defined as a segment in which the search tree from s and the search tree from t
+    // overlap. An edge is part of such a plateaux around `v` if
+    // v == parent_in_reverse_search(parent_in_forward_search(v)).
+    // Here we calculate the last node no the plateaux in either direction
+    const auto plateaux_end = [&](auto node, auto const &main_heap, auto const &side_heap) {
```
TheMarex: What is the meaning of `main_heap` and `side_heap` here? Is this the `forward_heap` and `backward_head`?

src/engine/routing_algorithms/alternative_path_mld.cpp
```diff
@@ -0,0 +1,936 @@
...
+
+    // Filters via candidate nodes with heuristics
+
+    // Note: filter pipeline below only makes range smaller; no need to erase items
+    // from the vector when we can mutate in-place and for filtering adjust iterators.
+    auto it = end(candidate_vias);
+
+    it = filterViaCandidatesByUniqueNodeIds(begin(candidate_vias), it);
+    it = filterViaCandidatesByRoadImportanceHeuristic(begin(candidate_vias), it, facade);
+    it = filterViaCandidatesByStretchHeuristic(begin(candidate_vias), it, shortest_path_weight);
```
TheMarex: Nitpick: all of these are `heuristics`, so the it doesn't add information to the function name. Removing that could make the function name a little bit less unwieldy.

src/engine/routing_algorithms/alternative_path_mld.cpp
```diff
@@ -0,0 +1,936 @@
...
+    // Only now that we annotated the alternatives do we have its actual duration. Note: the
+    // weight we used so far abstracts over duration and includes (potentially high) penalties.
+    // Note: see assembleGeometry for distance calculation
+
+    // Shortest Path
+    routes.push_back(unpacked_path_to_route(*unpacked_paths_first));
+
+    for (auto it = std::next(unpacked_paths_first); it != unpacked_paths_last; ++it)
+    {
+        const auto shortest_path_duration = routes.front().duration_without_turns();
```
TheMarex: This should be moved outside of the loop, otherwise this will be recalculated on every run, right?

src/engine/routing_algorithms/alternative_path_mld.cpp
```diff
@@ -0,0 +1,936 @@
...
+    const auto unpacked_path_to_route = [&](const WeightedViaNodeUnpackedPath &path) {
+        return extractRoute(facade, path.via.weight, phantom_node_pair, path.nodes, path.edges);
+    };
+
+    // Only now that we annotated the alternatives do we have its actual duration. Note: the
+    // weight we used so far abstracts over duration and includes (potentially high) penalties.
+    // Note: see assembleGeometry for distance calculation
+
+    // Shortest Path
+    routes.push_back(unpacked_path_to_route(*unpacked_paths_first));
```
TheMarex: The final unpacking + filtering step could be decomposed again as above. First unpack all candidates, then call an inplace filter function.

src/engine/routing_algorithms/alternative_path_mld.cpp
```diff
@@ -0,0 +1,936 @@
...
+                      search_engine_data,
+                      facade,
+                      phantom_node_pair);
+
+    //
+    // Filter and rank a second time. This time instead of being fast and doing
+    // heuristics on the packed path only we now have the detailed unpacked path.
+    //
+
+    auto unpacked_paths_last = end(unpacked_paths);
```
TheMarex: This should be refactored into an own function that works on a range of unpacked paths.

src/engine/routing_algorithms/alternative_path_mld.cpp
```diff
@@ -0,0 +1,936 @@
...
+        filterPackedPathsByLocalOptimalityHeuristic(weighted_packed_paths[0],
+                                                    forward_heap, // paths for s, via
+                                                    reverse_heap, // paths for via, t
+                                                    begin(weighted_packed_paths) + 1,
+                                                    alternative_paths_last);
+
+    const auto number_of_alternative_paths =
+        alternative_paths_last - (begin(weighted_packed_paths) + 1);
+
+    // Todo: refactor - this is ugly af
```
TheMarex: :-) you said it best yourself.

src/engine/routing_algorithms/alternative_path_mld.cpp
```diff
@@ -0,0 +1,936 @@
...
+    it = filterViaCandidatesByRoadImportanceHeuristic(begin(candidate_vias), it, facade);
+    it = filterViaCandidatesByStretchHeuristic(begin(candidate_vias), it, shortest_path_weight);
+
+    // Pre-rank by weight; sharing filtering below then discards by similarity.
+    std::sort(begin(candidate_vias), it, [](const auto lhs, const auto rhs) {
+        return lhs.weight < rhs.weight;
+    });
+
+    // Filtered and ranked candidate range
+    const auto first = begin(candidate_vias);
```
TheMarex: Maybe `via_condidates_first` might avoid name collisions.

TheMarex:
@daniel-j-h could you please also do some preliminary measurements on how much this increases query time? A few test routes on `bayern` should be enough to get a feeling here.

src/engine/routing_algorithms/alternative_path_mld.cpp
```diff
@@ -0,0 +1,936 @@
...
+    // We only have generic weights here and no durations without unpacking.
+    // We also have restricted way penalties which are huge and will screw scaling here.
+    //
+    // Users can pass us generic weights not based on durations; we can't do anything about
+    // it here other than either generating too many or no alternatives in these cases.
+    //
+    // We scale the weights with a step function based on some rough guestimates, so that
+    // they match tens of minutes, in the low hours, tens of hours, etc.
+
+    const constexpr auto minutes = 60.;
```
oxidase: Should it be `60. * weight_multiplier` here? Also it will work only for time-based weights

daniel-j-h:
Note: in https://github.com/Project-OSRM/osrm-backend/commit/dcdd0cf3c29547511fbf86b5d2e000168f59bdbd Michael is adding durations to the overlay. Which means we can already get durations from the heaps (`heap.GetData(n).duration`) for pre-filtering without expensive unpacking.

src/engine/routing_algorithms/alternative_path_mld.cpp
```diff
@@ -147,15 +147,19 @@ double scaledAtMostLongerByFactorBasedOnDuration(EdgeWeight duration)
...
 {
     util::static_assert_iter_category<RandIt, std::random_access_iterator_tag>();
     util::static_assert_iter_value<RandIt, WeightedViaNode>();
 
     // Assumes weight roughly corresponds to duration-ish. If this is not the case e.g.
     // because users are setting weight to be distance in the profiles, then we might
     // either generate more candidates than we have to or not enough. But is okay.
-    const auto scaledAtMostLongerBy = scaledAtMostLongerByFactorBasedOnDuration(weight);
+    const auto scaledAtMostLongerBy =
+        scaledAtMostLongerByFactorBasedOnDuration(weight / weight_multiplier);
```
daniel-j-h: @oxidase do you think this is good enough for now until we have durations in the overlay?

include/engine/api/route_parameters.hpp
daniel-j-h: 👍 https://github.com/Project-OSRM/osrm-backend/issues/3298

docs/profiles.md
daniel-j-h: Please check if this works you https://github.com/Project-OSRM/osrm-backend/pull/4047/commits/0cbca290e1edae294465e8ad8b49fbc3e1af488b

include/engine/internal_route_result.hpp
daniel-j-h: Hrm sorry I got confused by the `duration_until_turn` naming, will change and add a comment 👍

daniel-j-h:
@oxidase just helped me out running his plotting scripts for the osrm-runner bench runs on Bavaria. Seem's like our alternatives on mld are a bit slow in the current state.

Here's ch `alternatives=true` versus mld `alternatives=true`:

plot0

And here is ch `alternatives=true` versus mld `alternatives=5` (note the change on y axis):

plot1

I think we need to do two things here:
- Go in with a profiler and figure out which check is expensive and if we can make it faster while still providing high-quality alternatives. Especially figure out why the slowdown is not linear.
- See if unpacking twice as many candidate paths as requested alternatives is way too high. For five requested alternatives we unpack up to ten candidate paths in the final round of checks. We may have to implement the unpacking cache (see https://github.com/Project-OSRM/osrm-backend/issues/3835) and / or a stack allocator (see https://github.com/Project-OSRM/osrm-backend/issues/4213).

daniel-j-h:
Another datapoint with higher partition cell sizes,

    osrm-partition bayern-latest.osrm --max-cell-sizes 4096,131072,2097152,67108864

map

daniel-j-h:
Oops, I accidentally deleted your comment about the 400ms outlier @MoKob on bavaria - to be clear here this is on purpose using a `4096` level 1 cell size to see how the behavior changes and not about production performance numbers. Here are some additional cell sizes:

Our default partition config values: `128 4096 65536 2097152`
bench-128 4096 65536 2097152


Making the top most level smaller: `128 4096 65536 524288`
bench-128 4096 65536 524288

And introducing an additional level: `128 2048 32768 524288 2097152`:
bench-128 2048 32768 524288 2097152

PR #4049, opened 2017-05-16
remove penalty_table from car.lua

woodpeck:
remove penalty_table as it is apparently not used anywhere.


TheMarex:
Good catch, this was left-over from a profile restructuring. 👍 Thanks!

PR #4052, opened 2017-05-16
Mark way_function properties as mandatory

mloskot:
Format values and function names as code (refines #4040)

-----

@TheMarex More tidying-up


TheMarex:
Awesome thanks!

Issue #4038, opened 2017-05-12
Segfault on „Generating edge-expanded edges“


timojuez:
Hi people,

I try to osrm-extract the osm-file and get a segmentation fault. What can I do please?

Thank you,

Timo

Debug output:
Starting program: /home/timo/Dokumente/studium/aktuelles Semester/bewegungsprofile/osrm/osrm-backend/build/osrm-extract Dresden.osm -p ../profiles/foot.lua
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/lib/i386-linux-gnu/libthread_db.so.1".
[info] Using script ../profiles/foot.lua
[info] Input file: Dresden.osm
[info] Profile: foot.lua
[info] Threads: 2
[New Thread 0xb592db40 (LWP 27316)]
[New Thread 0xb4efeb40 (LWP 27317)]
[info] Parsing in progress..
[STXXL-MSG] STXXL v1.4.1 (prerelease/Debug)
[STXXL-ERRMSG] Warning: no config file found.
[STXXL-ERRMSG] Using default disk configuration.
[STXXL-MSG] Disk '/var/tmp/stxxl' is allocated, space: 1000 MiB, I/O implementation: syscall autogrow delete_on_exit queue=0 devid=0
[info] input file generated by osmconvert 0.8.4
[info] timestamp: n/a
[New Thread 0x9a25eb40 (LWP 27318)]
[Thread 0xb592db40 (LWP 27316) exited]
[Thread 0xb4efeb40 (LWP 27317) exited]
[info] Parsing finished after 35.7135 seconds
[info] Raw input contains 2587351 nodes, 445129 ways, and 0 relations
[New Thread 0xb592db40 (LWP 27335)]
[info] Sorting used nodes        ... [New Thread 0xa742bb40 (LWP 27336)]
ok, after 0.577053s
[info] Erasing duplicate nodes   ... ok, after 0.468767s
[info] Sorting all nodes         ... ok, after 3.29639s
[info] Building node id map      ... ok, after 1.90283s
[info] setting number of nodes   ... ok
[info] Confirming/Writing used nodes     ... ok, after 1.51424s
[info] Processed 557704 nodes
[info] Sorting edges by start    ... ok, after 1.99433s
[info] Setting start coords      ... ok, after 3.16936s
[info] Sorting edges by target   ... ok, after 2.2304s
[info] Computing edge weights    ... ok, after 4.27239s
[...]

PR #4053, opened 2017-05-16
Fix precision loss in angleBetween function

oxidase:
# Issue

In 32 bit mode  `angleBetween(257.78421507794314, 77.784215077943117)` returns 2.8421709430404007e-14 that causes an assertion #4038 
`angleBetween(77.784215077943117, 257.78421507794314)` returns 0 as expected.

In 64-bit mode both calls return 0

## Tasklist
 - [x] review
 - [x] adjust for comments

## Requirements / Relations

### DIFF
--- a/include/util/bearing.hpp
+++ b/include/util/bearing.hpp
@@ -120,15 +120,7 @@ inline double reverse(const double bearing)
 // entry_bearing of 0.
 inline double angleBetween(const double entry_bearing, const double exit_bearing)
 {
-    // transform bearing from cw into ccw order
-    const double offset = 360 - entry_bearing;
-    const double rotated_exit = [](double bearing, const double offset) {
-        bearing += offset;
-        return bearing > 360 ? bearing - 360 : bearing;
-    }(exit_bearing, offset);
-
-    const auto angle = 540 - rotated_exit;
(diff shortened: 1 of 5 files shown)


include/util/bearing.hpp
```diff
@@ -120,15 +120,7 @@ inline double reverse(const double bearing)
...
-    // transform bearing from cw into ccw order
-    const double offset = 360 - entry_bearing;
-    const double rotated_exit = [](double bearing, const double offset) {
-        bearing += offset;
-        return bearing > 360 ? bearing - 360 : bearing;
-    }(exit_bearing, offset);
-
-    const auto angle = 540 - rotated_exit;
-    return angle >= 360 ? angle - 360 : angle;
+    return std::fmod(entry_bearing - exit_bearing + 540., 360.);
```
TheMarex: Nice refactor!

oxidase:
@TheMarex basically the main change is to use
```
	movsd	.LC0(%rip), %xmm1
	movsd	.LC1(%rip), %xmm0
	call	fmod@PLT
```
so the result is corrected wrt arch-dependent ulps by fmod, but downside that it is less performant than
`while  (0 <= angle && angle <= 360.) angle -= 360.`

it is not related to 5.7.1, but can be merged in

PR #4054, opened 2017-05-16
Removes unused PUBLISH_NODE_BINDINGS var from .travis.yml

daniel-j-h:
I didn't get that we're no longer using this env var - sorry. cc @TheMarex @danpat


daniel-j-h:
- https://github.com/Project-OSRM/osrm-backend/blob/4294295242870d662abba9a2db754c4042995767/.travis.yml#L286
- https://github.com/Project-OSRM/osrm-backend/blob/4294295242870d662abba9a2db754c4042995767/scripts/travis/publish.sh#L10

PR #4056, opened 2017-05-17
R-Tree tuning

danpat:
This PR refactors the R-Tree data layout to improve performance, reduce data sizes (a little bit) and simplify file structures.

TL;DR - 20% faster queries at the expense of 3.8x more `.ramIndex` data.  Simpler file structures.  1% less R-Tree data overall (data moved from disk to RAM).

The main changes here are:

  - The `LeafNode` structure is removed.  The `.fileIndex` file now contains a flat array of `EdgeDataT` objects, (still) sorted by Hilbert Code on their centroid.  We still choose the number of `EdgeDataT` objects to pack into each leaf node based on a power-of-2 block size.
  - The `TreeNode` structure now only contains the bounding rectangle for each node in the R-Tree.

Tree navigation is now achieved by calculation, rather than storing pointers within the tree structure.  A small new data structure `m_tree_level_sizes` is stored in the `.ramIndex` file.  It contains the number of `TreeNodes` in each level of the tree, starting from the root (which is always size 1).

This is sufficient information, along with knowledge of BRANCHING_FACTOR and LEAF_NODE_SIZE, to calculate the position of the child of any node in the tree.  The `ExploreTreeNode`, `ExploreLeafNode` and `SearchInBox` functions have been updated to use this approach (the calculation logic resides in the `child_indexes(TreeIndex parent)` member function of the `StaticRTree` class.

The net result of these changes is:

  1. Improved performance - primarily due to the leaf node bounding rectangles being accessible in RAM, avoiding I/O for the lowest level bounding box check.
  2. Slightly less disk - about 0.5% to 1% of the total R-Tree data was eliminated.
  3. More RAM usage - the `.ramIndex` is now approximately 3.8x larger

Here is a spreadsheet that summarizes the size/performance differences after these changes are applied:

screen shot 2017-05-24 at 3 16 12 pm

## Tasklist
 - [x] review
 - [x] adjust for comments


include/util/rectangle.hpp
```diff
@@ -179,9 +179,12 @@ struct RectangleInt2D
...
 {
-    out << std::setprecision(12) << "(" << toFloating(rect.min_lon) << ","
-        << toFloating(rect.max_lon) << "," << toFloating(rect.min_lat) << ","
-        << toFloating(rect.max_lat) << ")";
+    out << std::setprecision(12) << "[[" << toFloating(rect.min_lon) << ","
+        << toFloating(rect.min_lat) << "],[" << toFloating(rect.min_lon) << ","
+        << toFloating(rect.max_lat) << "],[" << toFloating(rect.max_lon) << ","
+        << toFloating(rect.max_lat) << "],[" << toFloating(rect.max_lon) << ","
+        << toFloating(rect.min_lat) << "],[" << toFloating(rect.min_lon) << ","
+        << toFloating(rect.min_lat) << "]";
```
oxidase: a closing square bracket is missing or an opening one is not needed

include/util/static_rtree.hpp
```diff
@@ -49,71 +49,150 @@ namespace osrm
...
+ * Static RTree for serving nearest neighbour queries
+ * // All coordinates are pojected first to Web Mercator before the bounding boxes
+ * // are computed, this means the internal distance metric doesn not represent meters!
+ */
+
 template <class EdgeDataT,
           storage::Ownership Ownership = storage::Ownership::Container,
-          std::uint32_t BRANCHING_FACTOR = 128,
+          std::uint32_t BRANCHING_FACTOR = 64,
           std::uint32_t LEAF_PAGE_SIZE = 4096>
```
oxidase: It is unused and can be removed, also L37-45 can be removed

include/util/static_rtree.hpp
```diff
@@ -49,71 +49,150 @@ namespace osrm
...
     using Rectangle = RectangleInt2D;
     using EdgeData = EdgeDataT;
     using CoordinateList = Vector<util::Coordinate>;
 
-    static_assert(LEAF_PAGE_SIZE >= sizeof(uint32_t) + sizeof(EdgeDataT), "page size is too small");
+    static_assert(LEAF_PAGE_SIZE >= sizeof(EdgeDataT), "page size is too small");
     static_assert(((LEAF_PAGE_SIZE - 1) & LEAF_PAGE_SIZE) == 0, "page size is not a power of 2");
-    static constexpr std::uint32_t LEAF_NODE_SIZE =
-        (LEAF_PAGE_SIZE - sizeof(uint32_t) - sizeof(Rectangle)) / sizeof(EdgeDataT);
+    static constexpr std::uint32_t LEAF_NODE_SIZE = (LEAF_PAGE_SIZE / sizeof(EdgeDataT));
```
oxidase: `LEAF_NODE_SIZE` can be a template parameter similar to `BRANCHING_FACTOR` without computing from `LEAF_PAGE_SIZE`

include/util/static_rtree.hpp
```diff
@@ -49,71 +49,150 @@ namespace osrm
...
+     */
     struct TreeIndex
     {
-        TreeIndex() : index(0), is_leaf(false) {}
-        TreeIndex(std::size_t index, bool is_leaf) : index(index), is_leaf(is_leaf) {}
-        std::uint32_t index : 31;
-        std::uint32_t is_leaf : 1;
+        TreeIndex() : level(0), offset(0) {}
+        TreeIndex(std::size_t level_, std::size_t offset_) : level(level_), offset(offset_) {}
+        std::size_t level;  // Which level of the tree is this node in
```
TheMarex: Going smaller here would help, the size of `TreeIndex` is important for the performance of the queue: The smaller this struct is the less data will be shuffled around. Going `size_t` -> `uint32` should work for both. (maybe even 16 bits?)

include/util/static_rtree.hpp
```diff
@@ -125,13 +204,13 @@ class StaticRTree
...
             : squared_min_dist(squared_min_dist), tree_index(tree_index),
-              segment_index(std::numeric_limits<std::uint32_t>::max())
+              segment_index(std::numeric_limits<std::uint64_t>::max())
         {
         }
 
         QueryCandidate(std::uint64_t squared_min_dist,
                        TreeIndex tree_index,
-                       std::uint32_t segment_index,
+                       std::uint64_t segment_index,
```
TheMarex: Why 64bit?

include/util/static_rtree.hpp
```diff
@@ -140,27 +219,44 @@ class StaticRTree
...
         TreeIndex tree_index;
-        std::uint32_t segment_index;
+        std::uint64_t segment_index;
         Coordinate fixed_projected_coordinate;
     };
 
-    Vector<TreeNode> m_search_tree;
+    // We use a const view type when we don't own the data, otherwise
+    // we use a mutable type (probably becase we're building the tree)
+    using treeviewtype = typename std::conditional<Ownership == storage::Ownership::View,
```
TheMarex: `CamelCase` for our types.

include/util/static_rtree.hpp
```diff
@@ -254,97 +357,116 @@ class StaticRTree
...
-            while (processed_tree_nodes_in_level < tree_nodes_in_level.size())
+            auto previous_level_start_pos = m_search_tree.size() - nodes_in_previous_level;
+
+            // We can calculate how many nodes will be in this level, we divide by
+            // BRANCHING_FACTOR
+            // and round up
+            std::uint32_t nodes_in_current_level =
+                std::ceil(static_cast<double>(nodes_in_previous_level) / BRANCHING_FACTOR);
+
+            for (std::uint32_t current_node_idx = 0; current_node_idx < nodes_in_current_level;
```
TheMarex: Could be refactored using `util::irange`

include/util/static_rtree.hpp
```diff
@@ -254,97 +357,116 @@ class StaticRTree
...
+                auto first_child_index =
+                    current_node_idx * BRANCHING_FACTOR + previous_level_start_pos;
+                auto last_child_index =
+                    first_child_index +
+                    std::min(BRANCHING_FACTOR,
+                             nodes_in_previous_level - current_node_idx * BRANCHING_FACTOR);
+
+                // Calculate the bounding box for BRANCHING_FACTOR nodes in the previous
+                // level, then save that box as a new TreeNode in the new level.
+                for (auto child_node_idx = first_child_index; child_node_idx < last_child_index;
```
TheMarex: `util::irange`

include/util/static_rtree.hpp
```diff
@@ -254,97 +357,116 @@ class StaticRTree
...
+        // later on.  The workflow here is:
+        // The initial order of tree nodes in the m_search_tree array is roughly:
+        // 6789 345 12 0   (each block here is a level of the tree)
+        // Then we reverse it and get:
+        // 0 21 543 9876
+        // Now the loop below reverses each level to give us the final result
+        // 0 12 345 6789
+        // This ordering keeps the position math easy to understand during later
+        // searches
+        for (std::size_t i = 0; i < m_tree_level_sizes.size(); ++i)
```
TheMarex: `util::irange`

danpat:
@TheMarex I've addressed most of the review points.  I'm out for the next few days, please feel free to merge this as-is, or make whatever remaining modifications are necessary to get it into `master` for 5.8.

include/util/static_rtree.hpp
danpat: The reason I left this in is so that the leaf-node groups of EdgeDataT objects are still clustered in page-sized blocks, to minimize disk reads.  I guess we could make your suggested change and then just do some empirical tests to tune, but to be honest, the existing value seemed to perform best in my local tests, so I just left it alone.

PR #4057, opened 2017-05-17
Lets MakeResponse(Route) explicitly call MakeResponse(ManyRoutes) in RouteAPI

daniel-j-h:
See https://github.com/Project-OSRM/osrm-backend/pull/4035#discussion_r116890340

### DIFF
--- a/include/engine/api/route_api.hpp
+++ b/include/engine/api/route_api.hpp
@@ -65,11 +65,6 @@ class RouteAPI : public BaseAPI
         response.values["code"] = "Ok";
     }
 
-    void MakeResponse(const InternalRouteResult &raw_route, util::json::Object &response) const
-    {
-        return MakeResponse(raw_route, response);
-    }
-
   protected:
     template <typename ForwardIter>
     util::json::Value MakeGeometry(ForwardIter begin, ForwardIter end) const


daniel-j-h:
Small self-contained example:

- clang https://godbolt.org/g/CJ6Pa7
- gcc https://godbolt.org/g/fIl2S5

oxidase:
nice, it hits [the note](https://stackoverflow.com/questions/3592557/optimizing-away-a-while1-in-c0x)

> [ Note: This is intended to allow compiler transformations such as removal of empty loops, even when
termination cannot be proven. — end note ]

danpat:
Yeah.  I think we avoided the problem through the following mechanism:

  1. Compiler optimized away `MakeResponse(Route)`
  2. Calls that would've matched `MakeResponse(Route)` were instead redirected to `MakeResponse(ManyRoutes)` because it can cast `Route`->`ManyRoutes`

While this works, it's definitely a bit black-magicy, removing this weird function is 👍

daniel-j-h:
Follow up from chat: AppVeyor build was 🍏 looks like a spurious error:
https://ci.appveyor.com/project/DennisOSRM/osrm-backend/build/1.0.8186

Also: after meging this into master Travis also had a spurious failure:
https://travis-ci.org/Project-OSRM/osrm-backend/jobs/233549233

where we were unable to download Clang, silently continued the build, used the system's pre-installed Clang 3.5 and then errored during compilation.

```
[Mason] Downloading package https://mason-binaries.s3.amazonaws.com/linux-x86_64/clang++/4.0.0.tar.gz...
CMake Error at cmake/mason.cmake:80 (message):
[Mason] Failed to download  https://mason-binaries.s3.amazonaws.com/linux-x86_64/clang++/4.0.0.tar.gz:
curl: (56) SSL read: error:00000000:lib(0):func(0):reason(0), errno 104
```

PR #4059, opened 2017-05-17
Avoid boost::numeric_cast on trusted coordinates

danpat:
# Issue

During profiling for https://github.com/Project-OSRM/osrm-backend/pull/4056, I found that `toFixed()` was surprisingly high on the list of things that take up CPU time during route calculation.

This change creates a new longitude/latitude type that's flagged (via naming) as "unsafe" - instances of these unsafe types are passed to different versions of the `toFixed()` function which continue to perform `boost::numeric_cast` checks.  Not-`Unsafe*` variables are routed to simpler versions of `toFixed()` that don't perform overflow checking.

Data supplied by external sources: (1) user supplied coordinates in queries, and (2) longitude/latitude values read from `.osm` files during pre-processing are created as `Unsafe(Longitude|Latitude)` types, which causes them to receive an extra `boost::numeric_cast` check.

Data that comes from internally, already vetted sources (such as `GetCoordinateOfNode()` on the `datafacade`) is routed to the simpler `toFixed()` function which doesn't perform a `boost::numeric_cast`.

Our `rtree-bench` does a lot of lookups of already safe coordinates.  Here's the performance before this change:

```
./src/benchmarks/rtree-bench moldova-latest.osrm.ramIndex moldova-latest.osrm.fileIndex moldova-latest.osrm.nodes
Running raw RTree queries (1 result) with 10000 coordinates: Took 0.551909 seconds (551.91ms)  ->  0.055191 ms/query (551.91ms)
Running raw RTree queries (10 results) with 10000 coordinates: Took 0.595937 seconds (595.938ms)  ->  0.0595938 ms/query (595.938ms)
```

and after:

```
./src/benchmarks/rtree-bench moldova-latest.osrm.ramIndex moldova-latest.osrm.fileIndex moldova-latest.osrm.nodes
Running raw RTree queries (1 result) with 10000 coordinates: Took 0.45309 seconds (453.09ms)  ->  0.045309 ms/query (453.09ms)
Running raw RTree queries (10 results) with 10000 coordinates: Took 0.506266 seconds (506.266ms)  ->  0.0506266 ms/query (506.266ms)
```

## Tasklist
 - [X] Quantify performance gain (coordinate lookups are 19% faster)
 - [x] review
 - [x] adjust for comments

### DIFF
--- a/include/server/api/base_parameters_grammar.hpp
+++ b/include/server/api/base_parameters_grammar.hpp
@@ -111,15 +111,15 @@ struct BaseParametersGrammar : boost::spirit::qi::grammar<Iterator, Signature>
                                                 qi::_1,
                                                 qi::_2)];
 
-        location_rule =
-            (double_ > qi::lit(',') >
-             double_)[qi::_val = ph::bind(
-                          [](double lon, double lat) {
-                              return util::Coordinate(util::toFixed(util::FloatLongitude{lon}),
-                                                      util::toFixed(util::FloatLatitude{lat}));
-                          },
-                          qi::_1,
-                          qi::_2)];
+        location_rule = (double_ > qi::lit(',') >
+                         double_)[qi::_val = ph::bind(
+                                      [](double lon, double lat) {
+                                          return util::Coordinate(
+                                              util::toFixed(util::UnsafeFloatLongitude{lon}),
+                                              util::toFixed(util::UnsafeFloatLatitude{lat}));
+                                      },
+                                      qi::_1,
+                                      qi::_2)];
 
         polyline_rule = qi::as_string[qi::lit("polyline(") > +polyline_chars > ')']
                                      [qi::_val = ph::bind(
--- a/include/util/coordinate.hpp
+++ b/include/util/coordinate.hpp
@@ -53,16 +53,29 @@ struct latitude
 struct longitude
 {
 };
+struct unsafelatitude
+{
+};
+struct unsafelongitude
+{
+};
 }
 
+// Internal lon/lat types - assumed to be range safe
 using FixedLatitude = Alias<std::int32_t, tag::latitude>;
 using FixedLongitude = Alias<std::int32_t, tag::longitude>;
 using FloatLatitude = Alias<double, tag::latitude>;
 using FloatLongitude = Alias<double, tag::longitude>;
+// Types used for external input data - conversion functions perform extra
+// range checks on these (toFixed/toFloat, etc)
+using UnsafeFloatLatitude = Alias<double, tag::unsafelatitude>;
+using UnsafeFloatLongitude = Alias<double, tag::unsafelongitude>;
 static_assert(std::is_pod<FixedLatitude>(), "FixedLatitude is not a valid alias");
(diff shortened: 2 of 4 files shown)


daniel-j-h:
Idea is great 👍 - not sure we need more types here, though.
What about `unsafeTo*` functions, we need to be explicit anyway.

oxidase:
i think it is safe to directly switch from numeric_cast to static_cast, because int32_t is safely promoted to double. But i would do: 
```
     const auto floating = static_cast<double>(longitude) / COORDINATE_PRECISION;
```
to avoid assumptions about `COORDINATE_PRECISION` type

danpat:
@daniel-j-h Either approach would work of course, but I feel like a distinct type is better, because of it will prevent accidental passing of an unsafe value to a method that doesn't handle unsafe data properly.

include/server/api/base_parameters_grammar.hpp
```diff
@@ -111,15 +111,15 @@ struct BaseParametersGrammar : boost::spirit::qi::grammar<Iterator, Signature>
...
-                              return util::Coordinate(util::toFixed(util::FloatLongitude{lon}),
-                                                      util::toFixed(util::FloatLatitude{lat}));
-                          },
-                          qi::_1,
-                          qi::_2)];
+        location_rule = (double_ > qi::lit(',') >
+                         double_)[qi::_val = ph::bind(
+                                      [](double lon, double lat) {
+                                          return util::Coordinate(
+                                              util::toFixed(util::UnsafeFloatLongitude{lon}),
```
oxidase: Here numeric_cast also not really needed, because it :
* checks overflows
* checks underflows
* makes rounding  accordingly to a rounding policy 

double to double conversion with default rounding policy would be similar to static_cast. 

One thing that is allowed atm  is [parsing of NaNs](https://github.com/boostorg/spirit/blob/abe2a14a7fd281e0e6a6a32ab3cde31f277fbb64/include/boost/spirit/home/qi/numeric/detail/real_impl.hpp#L207-L208)
but it can be rejected by providing `RealPolicy` that fail at `parse_nan` and `parse_inf`

daniel-j-h: Here's an example for a `RealPolicy` we're already using

https://github.com/Project-OSRM/osrm-backend/blob/b1f6797aab453e141931af2389b8e16687e47fd6/include/server/api/base_parameters_grammar.hpp#L30-L68

see for context https://github.com/Project-OSRM/osrm-backend/issues/2009

oxidase: :+1: just did not look above, qi should provide safe double values in place

danpat: https://github.com/Project-OSRM/osrm-backend/blob/safe_coordinate_conversion/unit_tests/util/coordinate_calculation.cpp#L136

Also, we need to catch the user supplying double vales that are large. `toFixed` multiplies the user value by `COORDINATE_PRECISION`, so that can overflow if the user gives a value grater than max(double)/COORDINATE_PRECISION.

So this should remain as I wrote it, right?  Also, this usage is only called once per user coordinate, so doesn't contribute much to total query time. Other usages were called thousands of time per query, they were the real target of this PR.

PR #4063, opened 2017-05-18
Fixes penalty for turning onto restricted ways

daniel-j-h:
Potentially overflowing the `int16_t` turn weight type.
Removing custom turn penalty of `3000` - simply using the max turn weight.

See: https://github.com/Project-OSRM/osrm-backend/pull/4047#issuecomment-302409619


daniel-j-h:
Cherry picked into 5.7.

PR #4066, opened 2017-05-19
Set testbot profile and adjust expectations

### LINKED ISSUE #4062: features/testbot/zero-speed-updates.features use the bicycle profile
no profile is specified, and the test therefore defaults to bicycle

oxidase:
# Issue

Fixes #4062. Tests in 9358aa11287bd450468cc7067085f9e519dec9bc  implicitly use the bicycle profile.

For testbot profile:
* expectation in `Routing on restricted oneway` is modified because  segments are onway.
* some tests hit now incorrect assertion for empty heaps. This handled now in `search` functions after 9358aa11287bd450468cc7067085f9e519dec9bc

The first commit already landed in master :man_facepalming: but the second one is required.

## Tasklist
 - [x] review
 - [x] adjust for comments

## Requirements / Relations

### DIFF
--- a/features/testbot/zero-speed-updates.feature
+++ b/features/testbot/zero-speed-updates.feature
@@ -111,8 +111,8 @@ Feature: Check zero speed updates
             """
 
         When I route I should get
-          | from | to | bearings | code    |
-          |    1 |  2 | 270 270  | NoRoute |
+          | from | to | bearings | code      |
+          |    1 |  2 | 270 270  | NoSegment |
 
 
     Scenario: Via routing on restricted oneway
--- a/src/engine/routing_algorithms/shortest_path.cpp
+++ b/src/engine/routing_algorithms/shortest_path.cpp
@@ -131,8 +131,6 @@ void search(SearchEngineData<Algorithm> &engine_working_data,
                                     source_phantom.GetReverseWeightPlusOffset(),
                                 source_phantom.reverse_segment_id.id);
         }
-        BOOST_ASSERT(forward_heap.Size() > 0);
-        BOOST_ASSERT(reverse_heap.Size() > 0);
 
         search(engine_working_data,
                facade,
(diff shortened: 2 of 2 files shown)


daniel-j-h:
Care to elaborate why the assertions are wrong now or where it is handled now?

features/testbot/zero-speed-updates.feature
```diff
@@ -111,8 +111,8 @@ Feature: Check zero speed updates
             """
 
         When I route I should get
-          | from | to | bearings | code    |
-          |    1 |  2 | 270 270  | NoRoute |
```
TheMarex: The error changed because in the `bicycle` profile oneways are actually traversable in both directions (if you push your bike), right?

src/engine/routing_algorithms/shortest_path.cpp
```diff
@@ -166,8 +164,7 @@ void search(SearchEngineData<Algorithm> &engine_working_data,
                                     source_phantom.GetReverseWeightPlusOffset(),
                                 source_phantom.reverse_segment_id.id);
         }
-        BOOST_ASSERT(forward_heap.Size() > 0);
```
TheMarex: This can happen because we could get `PhantomNodes` that are not admissible as source/target right?

oxidase:
Some nodes that contain segments with `INVALID_EDGE_WEIGHT` after speed updates can not be used for routing due to integer overflows in weight, so heaps can be empty and the check is now [here](https://github.com/Project-OSRM/osrm-backend/blob/5c61f00ffa18a676de91c7869c831d73dd3d6162/src/engine/routing_algorithms/routing_base_ch.cpp#L85) instead of assertions

features/testbot/zero-speed-updates.feature
oxidase: yes, just missed the point that was testing on bicycle profile that allows  routing in reverse direction, for testbot profile it is not allowed

PR #4068, opened 2017-05-19
Fix failing bicycle test cases from refactor

TheMarex:
# Issue

These need fixes on master before we can merge #4033. This makes it easier to test the refactor and enable backporting these fixes to 5.7.

## Tasklist
 - [ ] review
 - [ ] adjust for comments

## Requirements / Relations

Broken out of #4033

### DIFF
--- a/features/bicycle/access.feature
+++ b/features/bicycle/access.feature
@@ -10,80 +10,80 @@ Feature: Bike - Access tags on ways
             | access | vehicle | bicycle  | bothw        |
             |        |         |          | cycling      |
             |        |         | yes      | cycling      |
-            |        |         | no       |              |
+            |        |         | no       | pushing bike |
             |        |         | dismount | pushing bike |
             |        | yes     |          | cycling      |
             |        | yes     | yes      | cycling      |
-            |        | yes     | no       |              |
+            |        | yes     | no       | pushing bike |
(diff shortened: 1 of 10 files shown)


features/bicycle/access.feature
```diff
@@ -10,80 +10,80 @@ Feature: Bike - Access tags on ways
...
+            |        |      | no       | pushing bike |                          |
+            |        |      | dismount | pushing bike |                          |
+            |        | yes  |          | cycling      |                          |
+            |        | yes  | yes      | cycling      |                          |
+            |        | yes  | no       | pushing bike |                          |
+            |        | yes  | dismount | pushing bike |                          |
+            |        | no   |          | cycling      |                          |
+            |        | no   | yes      | cycling      |                          |
+            |        | no   | no       |              |                          |
+            |        | no   | dismount | pushing bike | questionable combination |
```
emiltin: i think foot=no should prevent pushing of bikes

daniel-j-h:
What's the status here? The new profile version landed - @emiltin want to run the bike tests out?

TheMarex:
This PR is outdated but I still think we should take a fresh look at the bicycle profiles and the tests. There are some weird rules with oneway streets that need to be revisited, but I don't have the bandwidth for this right now.

Issue #3683, opened 2017-02-10
Usage of negative turn penalties


oxidase:
Some tests in OSRM make turn penalties updates with negative value. This introduces incorrect behavior in snapping start points and may lead to quite non-obvious problems like #3647 and #3641. 
Practically, negative turn penalties would make turns more preferable instead of going straight with 0 turn penalty. This can be achieved also by using a positive bias value and making all turn weight penalties positive. Duration values are not used as key values in heaps and can be negative.

 - [ ] clarify domain of turn penalties
- [ ] review tests features/car/traffic_turn_penalties.feature and features/testbot/traffic_turn_penalties.feature 
- [ ] adjust updates https://github.com/Project-OSRM/osrm-backend/blob/005124fe1c5c2e71a17531023b2437685b08ff6e/src/contractor/contractor.cpp#L832-L841 
- [ ] add non-negativity check of weights at https://github.com/Project-OSRM/osrm-backend/blob/005124fe1c5c2e71a17531023b2437685b08ff6e/src/contractor/contractor.cpp#L460-L467
- [ ] add at https://github.com/Project-OSRM/osrm-backend/blob/005124fe1c5c2e71a17531023b2437685b08ff6e/src/engine/routing_algorithms/routing_base.cpp#L112 assertion that `to_weight >= 0`

/cc @danpat

danpat:
I think that we should allow for negative turn penalty values.

Consider a simple intersection where most of the traffic is *not* taking the straight maneuver - 25% of the traffic volume goes straight, 75% makes the right-hand turn:

![screen shot 2017-02-13 at 4 09 07 pm](https://cloud.githubusercontent.com/assets/1892250/22909076/cc047560-f206-11e6-9145-f20e4a241cae.png)


Let's say we have samples of all cars passing the "Average speed" measurement point.  It's likely that the average speed is biased towards cars making the turning maneuver.

In this case, it would make sense for the 25% straight through traffic to receive a "negative turn penalty" - it's faster than average to proceed straight.

Of course, the validity of this idea depends on these biases existing.  If the speed approaching the intersection is the average of the vehicles making the fastest maneuver, then all penalties can be positive for the same effect, and the fastest maneuver would have a penalty of 0.

PR #4072, opened 2017-05-22
profiles api v2

emiltin:
improves the profile API and bumps the version from 1 to 2.

## Tasklist
- [x] move properties into profile table
- [x] improve version handling to avoid calling api functions when loading file
- [x] avoid the need for get_restrictions() in profiles
- [x] avoid the need for get_name_suffix_list() in profiles
- [x] reconcile constants.max_turn_weight and contants.precisions
- [x] improve module interface so profiles can include each other easily
- [x] stop using globals, instead pass profile as an arg to way/etc functions
- [x] rename way_functions to process_way (some for node, segment, turn)
- [x] pass functions instead of strings to Handlers.run()
- [x] initialize raster sources in setup() instead of in a separate callback
- [x] rename Handlers to WayHandlers, to prepare for NodeHandlers, etc
- [x] reorder arguments to Handlers functions to match way function.
- [x] group fields in the profile table accessed from c++ under profile.properties
- [x] rename sources to raster in the api
- [x] update profile documentation
- [x] update relevant Wiki pages
- [x] review
- [x] adjust for comments

## Requirements / Relations
This is a breaking change of the profile API.

### DIFF
--- a/docs/profiles.md
+++ b/docs/profiles.md
@@ -1,31 +1,97 @@
-OSRM supports "profiles". Configurations representing different routing behaviours for (typically) different transport modes. A profile describes whether or not we route along a particular type of way, or over a particular node in the OpenStreetMap data, and also how quickly we'll be travelling when we do. This feeds into the way the routing graph is created and thus influences the output routes.
+# OSRM profiles
+OSRM supports "profiles". Profiles representing routing behavior for different transport modes like car, bike and foot. You can also create profiles for variations like a fastest/shortest car profile or fastest/safest/greenest bicycles profile.
+
+ A profile describes whether or not it's possible to route along a particular type of way, whether we can pass a particular node, and  how quickly we'll be traveling when we do. This feeds into the way the routing graph is created and thus influences the output routes.
 
 ## Available profiles
+Out-of-the-box OSRM comes with profiles for car, bicycle and foot. You can easily modify these or create new ones if you like.
 
-Out-of-the-box OSRM comes with several different profiles, including car, bicycle and foot.
+Profiles have a 'lua' extension, and are places in 'profiles' directory.
 
-Profile configuration files have a 'lua' extension, and are found under the 'profiles' subdirectory.
-Alternatively commands will take a lua profile specified with an explicit -p param, for example:
+When running OSRM preprocessing commands you specify the profile with the --profile (or the shorthand -p) option, for example:
 
-`osrm-extract -p ../profiles/car.lua planet-latest.osm.pbf`
+`osrm-extract --profile ../profiles/car.lua planet-latest.osm.pbf`
 
-And then **you will need to extract and contract again** (A change to the profile will typically affect the extract step as well as the contract step. See [Processing Flow](https://github.com/Project-OSRM/osrm-backend/wiki/Processing-Flow))
+## Processing flow
+It's important to understand that profiles are used when preprocessing the OSM data, NOT at query time when routes are computed.
 
-## lua scripts?
+This means that after modifying a profile **you will need to extract, contract and reload the data again** and to see changes in the routing results. See [Processing Flow](https://github.com/Project-OSRM/osrm-backend/wiki/Processing-Flow) for more.
 
-Profiles are not just configuration files. They are scripts written in the "lua" scripting language ( http://www.lua.org )  The reason for this, is that OpenStreetMap data is not sufficiently straightforward, to simply define tag mappings. Lua scripting offers a powerful way of coping with the complexity of different node,way,relation,tag combinations found within OpenStreetMap data.
+## Profiles are written in LUA
+Profiles are not just configuration files. They are scripts written in the [LUA scripting language](http://www.lua.org). The reason for this is that OpenStreetMap data is complex, and it's not possible to simply define tag mappings. LUA scripting offers a powerful way to handle all the possible tag combinations found in OpenStreetMap nodes and ways.
 
-## Basic structure of a profile
+## Basic structure of profiles
+A profile will process every node and way in the OSM input data to determine what ways are routable in which direction, at what speed, etc.
 
-You can understand these lua scripts enough to make interesting modifications, without needing to get to grips with how they work completely.
+A profile will typically:
 
-Towards the top of the file, a profile (such as [car.lua](../profiles/car.lua)) will typically define various configurations as global variables. A lot of these are look-up hashes of one sort or another.
+- Define api version (required)
+- Require library files (optional)
+- Define setup function (required)
+- Define process functions (some are required)
+- Return functions table (required)
 
-As you scroll down the file you'll see local variables, and then local functions, and finally...
+A profile can also define various local functions it needs.
 
-`way_function` and `node_function` are the important functions which are called when extracting OpenStreetMap data with `osrm-extract`.
+Looking at [car.lua](../profiles/car.lua) as an example, at the top of the file the api version is defined and then required library files are included.
 
-The following global properties can be set in your profile:
+Then follows the `setup` functions, which is called once when the profile is loaded. It returns a big hash table of configurations, specifying things like what speed to use for different way types. The configurations are used later in the various processing functions. Many adjustments can be done just be modifying this configuration table.
+
+The setup function is also where you can do other setup, like loading elevation data source if you want to consider that when processing ways.
+
+Then comes the `process_node` and `process_way` functions, which are called for each OSM node and way when extracting OpenStreetMap data with `osrm-extract`.
+
+The `process_turn` function processes every possible turn in the network, and sets a penalty depending on the angle and turn of the movement.
+
+Profiles can also define a `process_segment` function to handle differences in speed along an OSM way, for example to handle elevation. As you can see, this is not currently used in the car profile.
+
+At the end of the file, a table if returned with references to the setup and processing functions the profile has defined.
+
+## Understanding speed, weight and rate 
+When computing a route from A to B there can be different measure of what is the best route. That's why there's a need for different profiles.
+
+Because speeds very on different types of roads, the shortest and the fastest route are typically different. But there are many other possible preferences. For example a user might prefer a bicycle route that follow parks or other green areas, even though both duration and distance are a bit longer.
+
+To handle this, OSRM doesn't simply choose the ways with the highest speed. Instead it uses the concept of `weight` and `rate`. The rate is an abstract measure that you can assign to  ways as you like to make some ways preferable to others. Routing will prefer ways with high rate.
+
+The weight of a way normally computed as length / rate. The weight can be thought of as the resistance or cost when passing the way. Routing will prefer ways with low weight.
+
+You can also set the weight of a way to a fixed value, In this case it's not calculated based on the length or rate, and the rate is ignored.
+
+You should set the speed to you best estimate of the actual speed that will be used on a particular way. This will result in the best estimated travel times.
+
+If you want to prefer certain ways due to other factors than the speed, adjust the rate accordingly. If you adjust the speed, the time time estimation will be skewed.
+
+If you set the same rate on all ways, the result will be shortest path routing.
+If you set rate = speed on all ways, the result will be fastest path routing.
+If you want to prioritize certain street, increase the rate on these.
+
+## Elements
+### api_version
+A profile should set api_version at the top of your profile. This is done to ensure that older profiles are still supported when the api changes. If api_version is not defined, 0 will be assumed. The current api version is 2.
+
+### Library files
+The folder [profiles/lib/](../profiles/lib/) contains LUA library files for handling many common processing tasks.
+
+File              | Notes
+------------------|------------------------------
+way_handlers.lua  | Functions for processing way tags
+tags.lua          | Functions for general parsing of OSM tags
+set.lua           | Defines the Set helper for handling sets of values
+sequence.lua      | Defines the Sequence helper for handling sequences of values
+access.lua        | Function for finding relevant access tags
+destination.lua   | Function for finding relevant destination tags
+destination.lua   | Function for determining maximum speed
+guidance.lua      | Function for processing guidance attributes
+
+They all return a table of functions when you use `require` to load them. You can either store this table and reference it's functions later, of if you need only a single you can store that directly.
+
+### setup()
+The `setup` function is called once when the profile is loaded and must return a table of configurations. It's also where you can do other global setup, like loading data sources that are used during processing.
+
+Note that processing of data is parallelized and several unconnected LUA interpreters will be running at the same time. The `setup` function will be called once for each. Each LUA iinterpreter will have it's own set of globals.
+
+The following global properties can be set under `properties` in the hash you return in the `setup` function: 
 
 Attribute                            | Type     | Notes
 -------------------------------------|----------|----------------------------------------------------------------------------
@@ -36,17 +102,40 @@ use_turn_restrictions                | Boolean  | Are turn instructions followed
 continue_straight_at_waypoint        | Boolean  | Must the route continue straight on at a via point, or are U-turns allowed? (default `true`)
 max_speed_for_map_matching           | Float    | Maximum vehicle speed to be assumed in matching (in m/s)
 max_turn_weight                      | Float    | Maximum turn penalty weight
-force_split_edges                    | Boolean  | True value forces a split of forward and backward edges of extracted ways and guarantees that `segment_function` will be called for all segments (default `false`)
+force_split_edges                    | Boolean  | True value forces a split of forward and backward edges of extracted ways and guarantees that `process_segment` will be called for all segments (default `false`)
 
-## way_function
+### process_node(profile, node, result)
+Process an OSM node to determine whether this node is a barrier or can be passed and whether passing it incurs a delay.
 
-Given an OpenStreetMap way, the `way_function` will either return nothing (meaning we are not going to route over this way at all), or it will set up a result hash to be returned. The most important thing it will do is set the value of `result.forward_speed` and `result.backward_speed` as a suitable integer value representing the speed for traversing the way.
+Argument | Description
+---------|-------------------------------------------------------
+profile  | The configuration table you returned in `setup`.
+node     | The input node to process (read-only).
+result   | The output that you will modify.
 
-All other calculations stem from that, including the returned timings in driving directions, but also, less directly, it feeds into the actual routing decisions the engine will take (a way with a slow traversal speed, may be less favoured than a way with fast traversal speed, but it depends how long it is, and... what it connects to in the rest of the network graph)
+The following attributes can be set on `result`:
 
-Using the power of the scripting language you wouldn't typically see something as simple as a `result.forward_speed = 20` line within the `way_function`. Instead a `way_function` will examine the tagging (e.g. `way:get_value_by_key("highway")` and many others), process this information in various ways, calling other local functions, referencing the global variables and look-up hashes, before arriving at the result.
+Attribute       | Type    | Notes
+----------------|---------|---------------------------------------------------------
+barrier         | Boolean | Is it an impassable barrier?
+traffic_lights  | Boolean | Is it a traffic light (incurs delay in `process_turn`)?
 
-The following attributes can be set on the result in `way_function`:
+## process_way(profile, way, result)
+Given an OpenStreetMap way, the `process_way` function will either return nothing (meaning we are not going to route over this way at all), or it will set up a result hash.
+
+Argument | Description
+---------|-------------------------------------------------------
+profile  | The configuration table you returned in `setup`.
+node     | The input way to process (read-only).
+result   | The output that you will modify.
+
+Importantly it will set `result.forward_mode` and `result.backward_mode` to indicate the travel mode in each direction, as well as set `result.forward_speed` and `result.backward_speed` to integer values representing the speed for traversing the way.
+
+It will also set a number of other attributes on `result`.
+
+Using the power of the scripting language you wouldn't typically see something as simple as a `result.forward_speed = 20` line within the `process_way` function. Instead `process_way` will examine the tag set on the way, process this information in various ways, calling other local functions and referencing the configuration in `profile`, etc, before arriving at the result.
+
+The following attributes can be set on the result in `process_way`:
 
 Attribute                               | Type     | Notes
 ----------------------------------------|----------|--------------------------------------------------------------------------
@@ -60,7 +149,7 @@ forward_classes                         | Table    | Mark this way as being of a
 backward_classes                        | Table    |  "   "
 duration                                | Float    | Alternative setter for duration of the whole way in both directions
 weight                                  | Float    | Alternative setter for weight of the whole way in both directions
-turn_lanes_forward                      | String   | Directions for individual lanes (normalised OSM `turn:lanes` value)
+turn_lanes_forward                      | String   | Directions for individual lanes (normalized OSM `turn:lanes` value)
 turn_lanes_backward                     | String   |  "   "
 forward_restricted                      | Boolean  | Is this a restricted access road? (e.g. private, or deliveries only; used to enable high turn penalty, so that way is only chosen for start/end of route)
 backward_restricted                     | Boolean  |  "   "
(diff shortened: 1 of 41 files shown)


oxidase:
@emiltin `max_turn_weight` is a read-only property that depends on `TurnPenalty` type and `weight_precision`. It is used to clip turn penalties in Lua to prevent overflows in conversions `double` to `TurnPenalty`. It can be set together with `context.properties.weight_precision` as

```
    if( weight_precision != sol::nullopt) {
      context.properties.weight_precision = weight_precision.value();
      context.properties.weight_multiplier = std::pow(10., context.properties.weight_precision);
      context.properties.max_turn_penalty = std::numeric_limits<TurnPenalty>::max() / weight_multiplier;
   }
```

emiltin:
ok. but is it really needed in the lua profiles then? can the capping be done on the c++ side?

oxidase:
it is needed in profiles as the maximal turn penalty that is used in transit routing restrictions. The turn restriction value can be clipped in c++ by removing assertions, but it should be made explicit with warnings

emiltin:
can the profile use a big constant as the max weight, and then let the c++ side cap it according to the current precision?

oxidase:
@emiltin yes, it is possible /cc @TheMarex

emiltin:
all tests are green. profiles can now use constants.max_turn_weight if needed. it's defined as a constants with the value -1. this assumes that negative turn weights are not valid, is that true?

this PR is a breaking change changes the profile interface a bit.

emiltin:
rebased

emiltin:
travis is failing with link errors:
/home/travis/build/Project-OSRM/osrm-backend/src/extractor/scripting_environment_lua.cpp:406: error: undefined reference to 'osrm::extractor::Sol2ScriptingEnvironment::MAX_TURN_WEIGHT'

it's compiling for me locally with gcc on linux. not sure why it's failing on travis?

include/extractor/scripting_environment_lua.hpp
```diff
@@ -46,6 +46,7 @@ class Sol2ScriptingEnvironment final : public ScriptingEnvironment
   public:
     static const constexpr int SUPPORTED_MIN_API_VERSION = 0;
     static const constexpr int SUPPORTED_MAX_API_VERSION = 1;
+    static const constexpr int MAX_TURN_WEIGHT = -1;
```
oxidase: Negative turn restrictions are correct values #3683

src/extractor/scripting_environment_lua.cpp
```diff
@@ -598,6 +615,9 @@ void Sol2ScriptingEnvironment::ProcessTurn(ExtractionTurn &turn)
         {
             turn_function(turn);
 
+            if( turn.weight == MAX_TURN_WEIGHT )
+              turn.weight = context.properties.GetMaxTurnWeight();
```
oxidase: It is better to set here `std::min(turn.weight, context.properties.GetMaxTurnWeight())` to avoid `MAX_TURN_WEIGHT` and print later number of clamped values

include/extractor/scripting_environment_lua.hpp
emiltin: ok, then of course we cannot use -1 as a special value

src/extractor/scripting_environment_lua.cpp
emiltin: you mean unconditionally cap the value? ie:
```c++
turn.weight = std::min(turn.weight, context.properties.GetMaxTurnWeight());`
```
what value should the profile use then to indicate max weight?

oxidase: the only purpose of the turn penalty maximum weight is to prevent int32_t overflows on c++ side. If turn penalties are modified on c++ then side the maximum value is not needed in Lua, but number of modifications must be reported as a warning.

emiltin:
@oxidase adjusted according to your suggestions. let's see what travis thinks.

oxidase:
@emiltin i have pushed some commits directly to the branch to check  changes on travis. The main change is in 62a0819 where `max_turn_weight` is moved to be a member of `profiles` properties. If it s not ok, you can kick out of the branch these commits.

oxidase:
@emiltin @TheMarex if the profile version is increased then 6e056e1d00c578fc55d54ec5acd402386fea0934 must be kicked out

emiltin:
thanks @oxidase for the commits.
@TheMarex yes, backward compatibility needs to be in place, that's not handled yet.

emiltin:
i think it's best to increase the api version to 2, and then incorporate other relevant changes to the api. it can then be merged at an appropriate osrm version change.

emiltin:
library files like handlers.lua are tied to a specific version of the api, for example because it accceses properties.weight_name.
this means that if you update osrm you also need to update the lua library. it also means that tests for older api version cannot use the library files, since the library files work with the latest api version.

emiltin:
for version 2 of the api might want to improve the way versioning is handled: #4133

emiltin:
bumped profile api to v2, while still maintaining compatibility with v0 and v1.

emiltin:
@oxidase i would prefer to keep max_turn_weight in constants rather than move it to profile, since it's a read-only value.

emiltin:
a single test fails on travis:

1) Scenario: Weighting based on speed file weights, ETA based on file durations - features/testbot/traffic_speeds.feature:86

this particular test amends the testprofile:

```
Scenario: Weighting based on speed file weights, ETA based on file durations
        Given the profile file "testbot" extended with
        """
        profile.traffic_signal_penalty = 0
        profile.u_turn_penalty = 0
        profile.weight_precision = 2
        """
```
not sure why that would work locally, but fail on travis.

oxidase:
@emiltin `max_turn_weight` is not really a constant but depends on `weight_precision`. I think it can be freely removed and turn penalties on c++ side can be silently clipped with `std::numeric_limits<TurnPenalty>::max()` value.

emiltin:
we can choose to make it constant with the value of std::numeric_limits<TurnPenalty>::max() and then clip the value the profile assigns to std::numeric_limits<TurnPenalty>::max() / std::pow(10., weight_precision)?

emiltin:
the profiles need some constant to assign when they want the maximum turn weight.

emiltin:
improved version handling (#4133), by moving all initialization of profile settings into a initialize() function.
to support modifying settings from cucumber tests (which is already used a lot) and from other profile (in cases where a profile includes another profile), a specialize() function is called if present, after the initialize() function.

danpat:
@emiltin just a heads up - in https://github.com/Project-OSRM/osrm-backend/commit/cd8fb82215e62653ffda645b8d58662182270b32 I added a new profile property, just want to flag it so you don't miss it accidentally on your next rebase.

emiltin:
@danpat thanks, will include it in the next rebase.

emiltin:
squashed and rebased on master

emiltin:
should we use semantic versioning of the profile api, e.g "v1.0.0", "v1.1", instead of just an integer?
i'm not sure the changes in this PR justifies moving to from version 1 to version 2.

TheMarex:
@emiltin the version flag is only used to work around API-breaking changes. We already version OSRM itself semantically so I don't see much use of having a separate semantic profile version.

Since this breaks the API version as far as I can tell, I feel a major version bump is fine.

emiltin:
ok let's keep using an int for the version then

emiltin:
windows build is failing. i'm having a hard time reading the error messages.
but could it be related to this line?
```c++
(void)key; // avoid warning about unused parameter
```

emiltin:
i think this is ready for a new review. are there other changes to the profile api that you want in v2?

TheMarex:
@emiltin looks good I thinkt global profile is now quite awkward to use. What about making it a parameter to the `way_function` and `node_function`?

Also it would be great if we could keep the `ProfileProperties` as one object within in the `profile` object.

```lua

function setup(profile)
  profile.properties.left_handed_driving = false
  profile.properties.weight_name = "cyclability"
end

function node_function(profile, node, result)
  -- sutff
end

function way_function(profile, way, result)
  -- sutff
end

function turn_function(profile, turn)
  -- sutff
end
```

I feel this is a little bit cleaner then our old approach of global objects, which we now want to deprecate.

emiltin:
yes, i would prefer passing profile as an argument and avoid using globals. it's close to how the functions in handlers.lua work. e.g:

```lua
function Handlers.handle_penalties(way,result,data,profile)
[...]
end
```

(where data is a table of fetched/cached tags)

emiltin:
yes we can group the properties inside the profile table. but they are not so special anymore, since they are not bindings to c++, just plain lua values that are read from c++. so not sure what the value of grouping them would be?

TheMarex:
> so not sure what the value of grouping them would be?

They are still saved in a C++ struct that will be available in the server, I feel it's easier to document that when they are grouped together.

oxidase:
@emiltin what do you think about profiles as instances of lua classes with possibility of specialization via inheritance? 

I could imaging a base car profile and possible derived specializations for trucks or any other special car type

emiltin:
@oxidase that's where i started, see the old PR #3172 which reorganized lua code into a class hierachy. (( (lua does not have real classes, however it's possible to emulate the behaviour.)

 @TheMarex felt this was overkill, so we decided on refactoring in smaller steps.

emiltin:
it's still a valid question how we make it easy to create variants of profiles without having to duplicate (and maintain) all the code in the profile. for example, we use a profile for cargo bikes. it should be exactly similar to the bike profile, except it should never pass steps or bicycle barriers, the speed is lower and a few other things like stronger avoidance of cobblestones.

with this PR, most or all of these things can be handled by using the new specialize function, which is simply a way to modify the profile table: you can require a profile and then define the specialize function and modify the settings when it's called. 

i would also like profiles to return the profile table, and include a list of the processing functions in the profile table. this would make it easier to require/adjust/reuse profiles. 

the idea of having a base profile and then specialized does sounds a bit like a class hierachy, but there's probably not much need to instantiate objects, since we process a single way, node or turn at a time.

emiltin:
the 'profile' table is not stored as a global anymore in lua profiles. instead it's returned by the initialize() function, and pass to way/node/segment/turn functions.

emiltin:
to improve the module interface, i propose that profiles return a table with references to the profile functions. this will make it easier for profiles to require each other and just modify where needed.

e.g:
```lua
-- tandem profile
local functions = require('bicycle')

function initialize(profile)
  profile.default_speed = 10
  return profile
end

functions.initialize:append(initialize)
return functions
```

functions.initialize would be a Sequence, that you can easily append to. initialize() will receive the profile table as produced by the previous initialize() function. the c++ would the call each function in the init sequence.

emiltin:
improved profile mode interface, as suggested above.

emiltin:
also added a lib/profile_v2.lua file, which includes all the common helpers, and sets api_version to 2.

TheMarex:
> functions.initialize would be a Sequence, that you can easily append to. initialize() will receive the profile table as produced by the previous initialize() function. the c++ would the call each function in the init sequence.

I feel it's a little bit more readable if we try to minimize the magic here. Same behavior can be achieved just by calling the function explicitly in the including functions `initialize` function.

```lua
-- tandem profile
local bicycle = require('bicycle')

function initialize(profile)
  bicycle.initialize(profile)
  profile.default_speed = 10
  return profile
end

return {
  ["initialize"]: initialize,
  ["way_function"]: bicycle.way_function,
  ["node_function"]: bicycle.node_function,
  ["turn_function"]: bicycle.turn_function,
}
```

emiltin:
alright, i'll change that.

emiltin:
what do you think about changing the name of the way_function (and similar) to something like process_way? having 'function' in a function name doesn't seem very descriptive.

TheMarex:
Yeah renaming the functions is a good idea. :+1:

emiltin:
could loading of raster data be done in initialize(), getting rid of source_function? or is there a particular reason to do it in a separate function?

TheMarex:
@emiltin excellent idea. 👍

emiltin:
@TheMarex how you you like this style of extending a profile:

```lua
functions = require('testbot')
functions.setup_testbot = functions.setup    -- store

-- modify setup function
functions.setup = function()
  local profile = functions.setup_testbot()
  profile.weight_name = 'steps'
  return profile
end

-- modify way function
functions.way = function(profile, way, result)
  result.forward_mode = mode.driving
  result.backward_mode = mode.driving
  result.weight = 1
  result.duration = 1
end

-- return modified table
return functions
```

it assigns functions directly to the function table, which gets rid of this block:

```
return {
  ["initialize"]: initialize,
  ["way_function"]: way_function,
  ["segment_function"]: testbot.segment,
  ["node_function"]: testbot.node,
  ["turn_function"]: testbot.turn,
}
```

which gets cumbersome, especially when having to repeat it in cucumber tests that otherwise modify only small parts of an existing profile. building the full table each time is also a bit error prone, since you need to remember to include all the functions, not just those you modified.

profiles/lib/function_table.lua
```diff
@@ -2,10 +2,20 @@
...
-    initialize = Sequence { initialize },
+    setup = setup,
     way = way_function,
     node = node_function,
     segment = segment_function,
     turn = turn_function,
   }
 end
+
+function modify_functions(original,modified)
```
emiltin: including this was not intended. going to remove

TheMarex:
@emiltin I think both are fine. Personally I prefer the more explicit version because it's easier to trace back where they are coming from.

emiltin:
it might be useful for a profile extending a parent profile to adjust the way_function. it could be done as with setup(), i.e. first call the parent way_function, then do additional processing.

another option: most profiles already uses Handlers.run() to run a sequence of functions. if the sequence is defined in the profile table then the sub profile can adjust it in setup(), for example by appending or removing handlers.

btw, this mechanism is similar to my previous implementation of doing setup: provide a sequence of  functions and call each in turn.

emiltin:
grouping properties inside the profile table requires that the user remember which fields are inside  profile.properties and which are not when accessing them from e.g. way_function:

```
profile.properties.u_turn_penalty     -- this is a property
profile.turn_penalty                  -- but this is not
```

this feels confusing. i would prefer keeping the properties as regular fields in the table so everything is accesses in the same way via profile.<field> or group them according to more meaningful categories than "stored in c++"/"not stored in c++".

emiltin:
I think this is ready for a new review. All tests green.

TheMarex:
> or group them according to more meaningful categories than "stored in c++"/"not stored in c++".

It's not really about C++ more like surfacing that these attributes are going to be available during the query. This abstraction makes the code easier to understand, otherwise I would always need to read the scripting code to know which values are put where. I feel it worthwhile to keep this abstraction around.

emiltin:
i though most of the attributes in ProfileProperties was only used during preprocesssing:

weight_name
traffic_signal_penalty
u_turn_penalty
max_speed_for_map_matching
continue_straight_at_waypoint
use_turn_restrictions
left_hand_driving
weight_precision
force_split_edges

emiltin:
i've grouped properties in the profile table. let me know what you think.

emiltin:
rebased.
should i care about codecov failing? travis and appveyer are both green.

emiltin:
the raster helpers are currently called via the global "sources", e.g sources:load(...). what about moving helpers like that into a general "api" scope, like api.sources:load(...)?

profiles/lib/profile_v2.lua
```diff
@@ -0,0 +1,9 @@
+-- profile api version 2
```
TheMarex: Is there a functional reason for having this as an own file, or was this just a refactor?

I would be cool if we could make the dependencies explicit in each profile.

src/extractor/scripting_environment_lua.cpp
```diff
@@ -542,48 +648,62 @@ void Sol2ScriptingEnvironment::ProcessElements(
...
+            strings.push_back(pair.second.as<std::string>());
+        };
     }
-
-    return restrictions;
+    return strings;
 }
 
-void Sol2ScriptingEnvironment::SetupSources()
+std::vector<std::string> Sol2ScriptingEnvironment::GetNameSuffixList()
```
TheMarex: This is cool, I think there is somewhere a binding fo `vector:Add` that we now probably only need for v1 and below.

include/extractor/scripting_environment_lua.hpp
```diff
@@ -49,8 +52,8 @@ struct LuaScriptingContext final
 class Sol2ScriptingEnvironment final : public ScriptingEnvironment
 {
   public:
-    static const constexpr int SUPPORTED_MIN_API_VERSION = 0;
-    static const constexpr int SUPPORTED_MAX_API_VERSION = 1;
+    static const constexpr int SUPPORTED_MIN_API_VERSION = 1;
```
TheMarex: Does that mean profiles with version `0` won't work anymore? That would be consider a breaking change, and I would like to avoid this.

src/extractor/scripting_environment_lua.cpp
```diff
@@ -459,14 +452,127 @@ void Sol2ScriptingEnvironment::InitContext(LuaScriptingContext &context)
...
+            throw util::exception("Profile setup() must return a table.");
+        else
+            context.profile_table = profile_table.value();
+
+        // store functions
+        context.turn_function = function_table.value()["process_turn"];
+        context.node_function = function_table.value()["process_node"];
+        context.way_function = function_table.value()["process_way"];
+        context.segment_function = function_table.value()["process_segment"];
+        context.source_function = context.state["source_function"];
```
TheMarex: So the source function is still used for v2 as well instead of `setup`?

profiles/lib/profile_v2.lua
emiltin: just a refactor to make it simpler to create a profile. all the profiles include almost the same files- can be revered if you like.

src/extractor/scripting_environment_lua.cpp
emiltin: no, this line should be removed

include/extractor/scripting_environment_lua.hpp
emiltin: there was never any code for handling version 0. m but we can allow it and make the api behave as with version 1 if you like.

emiltin: to clarify, setting version=0 was treated the same way as if no version was set in the profile file. i agree we should keep supporting both cases: no version set and version specifically set to 0. of course also version set to 1.

src/extractor/scripting_environment_lua.cpp
emiltin: actually, we still need the source_function for v1 profiles. seems we're missing a test for source_function in v1 profiles.

emiltin:
maybe we should maintain profiles/compatibiltiy/v0, profiles/compatibiltiy/v1 etc. to test backward compatibility? for example we're missing a test for the v1 way of loading raster sources and rasterbot.lua has been modified to work with the v2 api. instead of writing entire profiles in the cucumber tests, it might be better to keep a profile that uses all aspects of a specific api?

emiltin:
restored support for version=0 and added tests in features/options/profiles/version0.feature. profiles without a version defined are also supported, version 0 will be assumed.  negative version and version higher that 2 will return an error.

emiltin:
is it correctly understood that when the weight is just the inverse of the rate? and that setting output.weight in way_function is just a shorthand for setting both forward and backward rates?

should be change output.weigh to output.rate?

emiltin:
seems that weight = distance/rate. i'm trying to explain this in doc/profiles.md

## Understanding speed, weight, rate and distance
When computing a route from A to B there can be different measure of what is the best route. Because speeds very on differnt types of roads, the shortest and the fastest route are typically different. But there are many other possible preferences. For example a user might prefer a bicycle route that follow parks or other green areas, even though both duration and distance are a bit longer.

To handle this, OSRM doesn't simply choose the ways with the highest speed. Intead it uses the concept of `weight` and `rate`. The rate is an abstract meassure that you can assign to to ways as you like to make some ways preferable to others. Routing will prefer ways with high rate.

The weight of a way is computed as distance / rate. The weight can be though of as the resistance or cost when passing a way. Routing will prefer ways with low weight.

You should set the speed to you best estimante of the actual speed that will be used on a particual way. This will result in the best estimated travel times.

If you want to prefer certain ways due to other factors than the speed, adjust the rate accordingly. If you adjust the speed, the time time estimation will be skewed.

If you set the same rate on all ways, the result will be shortest path routing.
If you set rate = speed on all ways, the result will be fastest path routing.
If you want to prioritize e.g. smaller street, increase the rate on these.

emiltin:
looking at how raster source data is used i'm wondering about this bit:

```lua
  local sourceData = sources:query(profile.raster_source, segment.source.lon, segment.source.lat)
  local invalid = sourceData.invalid_data()
  if sourceData.datum ~= invalid then
    -- use value in targetData.datum
```

wouldn't it be easier if you could just do:
```lua
  local datum = sources:query(profile.raster_source, segment.source.lon, segment.source.lat)
  if datum ~= nil then
    -- use datum
```

and maybe we should rename source to something like raster_data? sources seems like a very generic word.

TheMarex:
> is it correctly understood that the weight is just the inverse of the rate? and that setting output.weight in way_function is just a shorthand for setting both forward and backward rates?

`weight` is like the `duration` value: It is fixed for the whole way. `rate` on the other hand is like the `speed` values, it depends on the geometric length on the way. The formula is `weight = length / rate` (like `duration = length / speed`).

> looking at how raster source data is used i'm wondering about this bit:

I think this is due to the fact that the C++ code will return an integer always. Probably needs a shim in the bindings to make this return `nil` in lua-land.

> and maybe we should rename source to something like raster_data? sources seems like a very generic word.

Hrm yeah what about `raster`?

emiltin:
makes sense. however i'm confused by your statement that rate depends on the geometric length on the way. isn't it the other way around: weight depend on the length of the way, while rate does not?



.

TheMarex:
@emiltin yes, I meant you need the `geometric length` to convert `rate` to the actual value used by OSRM.

emiltin:
alright. if you set the weight, are the rates then ignored?

emiltin:
why is there no forward_weight and backward_weight?

emiltin:
update docs/profiles.md. might need a few adjustments regarding weight and rate.

TheMarex:
@emiltin just because we don't have `forward_duration` and `backward_duration`. I don't see a problem with adding it, it was just done for consistency I think. If you are up for it, that should be a simple change to `ExtractionWay` and `ExtractorCallbacks::ProcessWay` to change it to directional versions.

emiltin:
updated changelog and rebased on master.

docs/profiles.md
```diff
@@ -1,31 +1,97 @@
...
 
-You can understand these lua scripts enough to make interesting modifications, without needing to get to grips with how they work completely.
+To handle this, OSRM doesn't simply choose the ways with the highest speed. Intead it uses the concept of `weight` and `rate`. The rate is an abstract meassure that you can assign to  ways as you like to make some ways preferable to others. Routing will prefer ways with high rate.
 
-Towards the top of the file, a profile (such as [car.lua](../profiles/car.lua)) will typically define various configurations as global variables. A lot of these are look-up hashes of one sort or another.
+The weight of a way normally computed as length / rate. The weight can be though of as the resistance or cost when passing the way. Routing will prefer ways with low weight.
 
-As you scroll down the file you'll see local variables, and then local functions, and finally...
+You can also set the weight og a way to a fixed value, In this case it's not calculated based on the length or rate, and the rate is ignored.
```
TheMarex: typo `the weight of a way to a fixed value`

docs/profiles.md
```diff
@@ -1,31 +1,97 @@
...
 
-## Basic structure of a profile
+Because speeds very on differnt types of roads, the shortest and the fastest route are typically different. But there are many other possible preferences. For example a user might prefer a bicycle route that follow parks or other green areas, even though both duration and distance are a bit longer.
 
-You can understand these lua scripts enough to make interesting modifications, without needing to get to grips with how they work completely.
+To handle this, OSRM doesn't simply choose the ways with the highest speed. Intead it uses the concept of `weight` and `rate`. The rate is an abstract meassure that you can assign to  ways as you like to make some ways preferable to others. Routing will prefer ways with high rate.
 
-Towards the top of the file, a profile (such as [car.lua](../profiles/car.lua)) will typically define various configurations as global variables. A lot of these are look-up hashes of one sort or another.
+The weight of a way normally computed as length / rate. The weight can be though of as the resistance or cost when passing the way. Routing will prefer ways with low weight.
```
TheMarex: typo `can be thought of`

emiltin:
i ran the changelog through a spell checker (US english) and corrected a number of errors.

emiltin:
should this wiki page https://github.com/Project-OSRM/osrm-backend/blob/master/docs/profiles.md be updated, or replaced with a reference to the file in the repo?

TheMarex:
@emiltin I think a link to the repo-docs would be best.

emiltin:
ah, the main wiki page already uses a link to the repo file.

PR #4076, opened 2017-05-23
don't collapse highway ramps into uturns (#4074)

### LINKED ISSUE #4074: Name -> Ref change considered obvious
When changing from a road with name `X` onto a road that is only offering a reference `Y` (e.g. each highway ramp onto a major motorway name only with a reference), the name change is considered obvious.
This behaviour leads to classifying roads as u-turns invalidly.

One example can be found here: http://0.0.0.0:9966/?z=18&center=40.585056%2C-122.349908&loc=40.585645%2C-122.349780&loc=40.584944%2C-122.352542&hl=en&alt=0

screen shot 2017-05-23 at 11 15 41 2

There are two aspects to this, one is that a `ramp` should never be part of a u-turn.
The other is that name changes like this are, in my opinion, not obvious and should constitute a full new name.

MoKob:
# Issue

resolves #4074

## Tasklist
 - [ ] review
 - [ ] adjust for comments

### DIFF
--- a/src/engine/guidance/collapse_scenario_detection.cpp
+++ b/src/engine/guidance/collapse_scenario_detection.cpp
@@ -147,6 +147,12 @@ bool isUTurn(const RouteStepIterator step_prior_to_intersection,
             step_prior_to_intersection, step_entering_intersection, step_leaving_intersection))
         return false;
 
+    // uturns only allowed on turns
+    if (!hasTurnType(*step_entering_intersection, TurnType::Turn) &&
+        !hasTurnType(*step_entering_intersection, TurnType::Continue) &&
+        !hasTurnType(*step_entering_intersection, TurnType::EndOfRoad))
+        return false;
+
     // the most basic condition for a uturn is that we actually turn around
     const bool takes_u_turn = bearingsAreReversed(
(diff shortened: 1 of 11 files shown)


src/engine/guidance/collapse_scenario_detection.cpp
```diff
@@ -147,6 +147,9 @@ bool isUTurn(const RouteStepIterator step_prior_to_intersection,
             step_prior_to_intersection, step_entering_intersection, step_leaving_intersection))
         return false;
 
+    if (hasTurnType(*step_entering_intersection), TurnType::OnRamp)
```
daniel-j-h: Is `OnRamp` the only type we need to care about here? Why this one specifically - because it fixes the particular edge case? What about e.g. `OffRamp`?

MoKob: Of course we want `OffRamp` as well. Thank you for pointing it out. It might be that other types are also of interest, although none come to mind right now. Added `OffRamp`.

MoKob:
Added the whitelist approach. Will rebase to get Changelog unconflicted

Issue #4078, opened 2017-05-23
Left instead of Right


MoKob:
There is a copy/paste issue in handling forks, assigning left instead of right to a turn (located [here](https://github.com/Project-OSRM/osrm-backend/blob/a63b43c259d49fc4915810eb4bfdd86bc78bce93/src/extractor/guidance/intersection_handler.cpp#L281))

![screen shot 2017-05-23 at 12 32 22 2](https://cloud.githubusercontent.com/assets/12932279/26353212/93fe0fb2-3fbf-11e7-8dc1-f3e81be60a28.png)

PR #4079, opened 2017-05-23
fix invalid turn angle in forks

MoKob:
# Issue

resolves https://github.com/Project-OSRM/osrm-backend/issues/4078

## Tasklist
 - [x] review
 - [x] adjust for comments

### DIFF
--- a/src/extractor/guidance/turn_handler.cpp
+++ b/src/extractor/guidance/turn_handler.cpp
@@ -705,9 +705,8 @@ void TurnHandler::handleDistinctConflict(const EdgeID via_edge,
         const auto left_classification = node_based_graph.GetEdgeData(left.eid).road_classification;
         const auto right_classification =
             node_based_graph.GetEdgeData(right.eid).road_classification;
-        if (canBeSeenAsFork(left_classification, right_classification))
-            assignFork(via_edge, left, right);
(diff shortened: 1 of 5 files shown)


src/extractor/guidance/turn_handler.cpp
```diff
@@ -705,9 +705,8 @@ void TurnHandler::handleDistinctConflict(const EdgeID via_edge,
         const auto left_classification = node_based_graph.GetEdgeData(left.eid).road_classification;
         const auto right_classification =
             node_based_graph.GetEdgeData(right.eid).road_classification;
-        if (canBeSeenAsFork(left_classification, right_classification))
-            assignFork(via_edge, left, right);
-        else if (left_classification.GetPriority() > right_classification.GetPriority())
+
```
daniel-j-h: The c&p issue above I get but why do we need the change here removing the `canBeSeenAsFork` branch?

MoKob: Fork does left/right assignment which does not really work in all cases. I wasn't able to reproduce a case for it, but since we are assigning all right/ all left turns here, assigning 'fork left/fork right` doesn't do so well.

MoKob:
Travis hanging here, in CI its completely passed.

Issue #4084, opened 2017-05-24
Matching plugin suspicious stops matching the trace


gardster:
[...]

PR #4085, opened 2017-05-24
Matching split breakage ignoring fix.

gardster:
# Issue

Issue #4084


TheMarex:
To be clear this would allow splitting even if `gaps=ignore` is set which seems in contrast to what the parameter was intended to do?

gardster:
This allow splitting if something went wrong and track already has MAX_BROKEN_STATES coordinates without snapping.  But splitting based on the huge timestamp difference will be ignored.

Issue #4069, opened 2017-05-19
Fix weights of start nodes and check related issues


oxidase:
At the moment there are some issues that looks different but have a common problem:

* https://github.com/Project-OSRM/osrm-backend/issues/3953
* https://github.com/Project-OSRM/osrm-backend/pull/4019 (it is not real fix, but a workaround to prevent UB)
* https://github.com/Project-OSRM/osrm-backend/issues/3429
* probably https://github.com/Project-OSRM/osrm-backend/issues/3647
* https://github.com/Project-OSRM/osrm-backend/issues/2016
* https://github.com/Project-OSRM/osrm-backend/issues/3969
* https://github.com/Project-OSRM/osrm-backend/blob/5c61f00ffa18a676de91c7869c831d73dd3d6162/include/engine/guidance/assemble_geometry.hpp#L101-L103 as part of https://github.com/Project-OSRM/osrm-backend/issues/3606

The problem is in computation weights for the starting edge-based-node and weights for its neighboors that are computed as `-weight(as) + [weight(ab) + turn_penalty(adc)]` for segments like `a---s---b-------c`. The reason for this is the total [`EdgeBasedEdge::EdgeDate::weight`](https://github.com/Project-OSRM/osrm-backend/blob/5c61f00ffa18a676de91c7869c831d73dd3d6162/src/extractor/edge_based_graph_factory.cpp#L503-L504), so it is not possible to compute directly `weight(sb) + turn_penalty(adc)`.  Another problem occurs if at least one segment in `as` has `INVALID_EDGE_WEIGHT` value, so `weight(as)` can not be computed and there is no possibility to compute route even if `weight(sb)` has a valid value.

The solution can be as discussed with @TheMarex:
*  ~~adding `TurnPenalty` and `TurnDuration` to [`EdgeBasedEdge::EdgeData`](https://github.com/Project-OSRM/osrm-backend/blob/5c61f00ffa18a676de91c7869c831d73dd3d6162/include/extractor/edge_based_edge.hpp#L47-L50) (+4 bytes per turn)~~ We already have this https://github.com/Project-OSRM/osrm-backend/blob/master/include/engine/datafacade/contiguous_internalmem_datafacade.hpp#L624-L634
* only filtering out edges with **all** `INVALID_EDGE_WEIGHT` segments (atm edges with **at least one** `INVALID_EDGE_WEIGHT` segment are filtered out of the base graph)
[...]

PR #4086, opened 2017-05-24
Refactor insertion of source and target phantom nodes

oxidase:
# Issue

The  PR is a part of #4069

## Tasklist
 - [ ] review
 - [ ] adjust for comments

## Requirements / Relations

### DIFF
--- a/include/engine/routing_algorithms/routing_init.hpp
+++ b/include/engine/routing_algorithms/routing_init.hpp
@@ -59,119 +59,272 @@ sumSegmentValues(Iterator first, const Iterator last, float first_ratio, float l
 }
 
 template <typename Facade, typename Heap>
+void insertForwardSourceNode(const Facade &facade,
+                             Heap &heap,
+                             const PhantomNode &node,
+                             const EdgeWeight initial_weight)
+{ // Insert source node in forward direction into a forward heap
+    // For the weights vector of the phantom node and fwd_segment_position = 4:
+    //   0---1---2---3---(4-s==)5===6===end
+    //      first_segment ^              ^ last_segment
+    // source.fwd_segment_ratio = distance(4→s) / distance(4→5)
+    const auto node_id = node.forward_segment_id.id;
+    const auto geometry_index = facade.GetGeometryIndex(node_id);
+    const auto weights = facade.GetUncompressedForwardWeights(geometry_index.id);
+    const auto first_segment = weights.begin() + node.fwd_segment_position;
+    const auto last_segment = weights.end();
+    if (areSegmentsValid(first_segment, last_segment))
+    {
+        heap.InsertVisited(node_id, initial_weight, node_id);
+        auto weight = sumSegmentValues(first_segment, last_segment, node.fwd_segment_ratio, 1.);
+        detail::relaxSourceEdges(facade, heap, node_id, initial_weight + weight);
+    }
+}
+
+template <typename Facade, typename Heap>
+void insertReverseSourceNode(const Facade &facade,
+                             Heap &heap,
+                             const PhantomNode &node,
+                             const EdgeWeight initial_weight)
+{ // Insert source node in reverse direction into a forward heap
+    // For the weights vector of the phantom node and fwd_segment_position = 4:
+    // end===6===5===4===3(=s--2)---1---0
+    //  ^ last_segment         ^ first_segment
+    // source.fwd_segment_ratio = distance(s→3) / distance(2→3)
+    const auto node_id = node.reverse_segment_id.id;
+    const auto geometry_index = facade.GetGeometryIndex(node_id);
+    const auto weights = facade.GetUncompressedReverseWeights(geometry_index.id);
+    const auto first_segment = weights.end() - node.fwd_segment_position - 1;
+    const auto last_segment = weights.end();
+    if (areSegmentsValid(first_segment, last_segment))
+    {
+        heap.InsertVisited(node_id, initial_weight, node_id);
+        auto weight = sumSegmentValues(first_segment, last_segment, 1 - node.fwd_segment_ratio, 1.);
+        detail::relaxSourceEdges(facade, heap, node_id, initial_weight + weight);
+    }
+}
+
+template <typename Facade, typename Heap>
+void insertForwardTargetNode(const Facade &facade,
+                             Heap &heap,
+                             const PhantomNode &node,
+                             const EdgeWeight initial_weight)
+{ // Insert target node in forward direction into a backward heap
+    // For the weights vector of the phantom node and fwd_segment_position = 4:
+    //   0===1===2===3===(4=t--)5---6---end
+    //   ^ first_segment        ^ last_segment
+    // target.fwd_segment_ratio = distance(4→t) / distance(4→5)
+    const auto node_id = node.forward_segment_id.id;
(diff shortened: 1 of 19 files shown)


include/engine/routing_algorithms/routing_init.hpp
```diff
@@ -0,0 +1,334 @@
...
+    else if (cross_reverse)
+    {
+        return getReverseSingleNodePath(facade, source, target);
+    }
+
+    return std::pair<NodeID, EdgeWeight>{SPECIAL_NODEID, INVALID_EDGE_WEIGHT};
+}
+
+template <typename Facade, typename Heap>
+auto insertNodesInHeapsToTargetForward(const Facade &facade,
```
TheMarex: `TargetForward` -> `ForwardTarget` reads better.

include/engine/routing_algorithms/routing_init.hpp
```diff
@@ -0,0 +1,334 @@
...
+    if (cross_forward)
+    {
+        return getForwardSingleNodePath(facade, source, target);
+    }
+
+    return std::pair<NodeID, EdgeWeight>{SPECIAL_NODEID, INVALID_EDGE_WEIGHT};
+}
+
+template <typename Facade, typename Heap>
+auto insertNodesInHeapsToTargetReverse(const Facade &facade,
```
TheMarex: `TargetReverse` -> `ReverseTarget`

include/util/query_heap.hpp
```diff
@@ -158,6 +158,13 @@ class QueryHeap
         node_index[node] = index;
     }
 
+    void InsertVisited(NodeID node, Weight weight, const Data &data)
```
TheMarex: What is the difference between `Insert()` and this function?

src/engine/routing_algorithms/routing_base_mld.cpp
```diff
@@ -0,0 +1,54 @@
...
+namespace osrm
+{
+namespace engine
+{
+namespace routing_algorithms
+{
+namespace mld
+{
+
+double getNetworkDistance(SearchEngineData<Algorithm> &engine_working_data,
```
TheMarex: Good idea to move this. 👍

src/updater/updater.cpp
```diff
@@ -868,7 +868,7 @@ Updater::LoadAndUpdateEdgeExpandedGraph(std::vector<extractor::EdgeBasedEdge> &e
                           });
     }
 
-    if (update_turn_penalties)
+    if (update_turn_penalties || update_conditional_turns)
```
TheMarex: Interesting, is this a bug that is in `master` too?

include/util/query_heap.hpp
oxidase: `InsertVisited` inserts node only into data vector and omits insertion into heap. It is needed only for source nodes, because source nodes weights are not real weights that must be 0, but  weights from projected source phantoms to ends of edge-based nodes. The special `relaxSourceEdges` function uses these weight values to relax source nodes.

include/engine/routing_algorithms/routing_init.hpp
```diff
@@ -0,0 +1,334 @@
...
+        const auto &edge_data = facade.GetEdgeData(edge);
+        if (edge_data.forward)
+        {
+            const auto to = facade.GetTarget(edge);
+            const auto to_weight = weight + getEdgeInternalWeight(facade, node, edge);
+            if (!heap.WasInserted(to))
+            {
+                heap.Insert(to, to_weight, node);
+            }
+            else if (!heap.WasRemoved(to) && to_weight < heap.GetKey(to))
```
oxidase: Here node `to` can be removed if it is a source node (inserted via `InsertedVisited`) and has not a weight, so `heap.GetKey(to)` returns a weight from the projected point to the end of node.

For example, for an ebg node `a------->b` let the snapped node is `a`, forward node is `0` and reverse node is `1`. Let `weight(ab)` is 100, so node `0` will be inserted with weight 100, the next  node `1` will be inserted with weight 0. If there is a forward u-turn edge from `1` to `0` with weight smaller than `weight(ab)` then  `to_weight < heap.GetKey(to)` is true and weight 100 will be changed to some wrong one.

oxidase:
@TheMarex to debug some test cases, i added `geojson` and `svg` printers for `gdb`. Here is an example for an edge loop `6->6` [roundabout-turn.feature_140.pdf](https://github.com/Project-OSRM/osrm-backend/files/1055383/roundabout-turn.feature_140.pdf) in https://github.com/Project-OSRM/osrm-backend/blob/84268d3b7de24f33e3c6be89f536a335f21134ad/features/guidance/roundabout-turn.feature#L162 that is unpacked to `6->2->5->9->6`

oxidase:
@TheMarex with almost finished refactoring I realized that the PR makes almost no improvement but makes handling of edge cases more complicated.
As described #4069 PR should modify the following:

1) Change weights to an unsigned type
2) Allow to route from partially blocked ebg nodes

The first goal is possible but it requires explicit checks for source nodes in `search` functions, because  weight's sign is **used implicitly  as a flag to distinguish source nodes**:

- in relaxation of edges
- in computation of a candidate path weight when the meeting node is the source one

Making all weights non-negative requires [explicit checks for source nodes](https://github.com/Project-OSRM/osrm-backend/blob/0843386d4b9c9cb1eb79f70f0f3f6b96265946a5/include/engine/routing_algorithms/routing_base_ch.hpp#L122-L130) and [checks that the `to` node](https://github.com/Project-OSRM/osrm-backend/blob/0843386d4b9c9cb1eb79f70f0f3f6b96265946a5/include/engine/routing_algorithms/routing_base_mld.hpp#L200) was not removed, because previously `to_weight < forward_heap.GetKey(to)` was always false for source nodes and now it is not the case.

For the second goal I added a special [relaxation of edges from source nodes](https://github.com/Project-OSRM/osrm-backend/blob/0843386d4b9c9cb1eb79f70f0f3f6b96265946a5/include/engine/routing_algorithms/routing_init.hpp#L17-L37). The problem here is that after the contraction some forward edges can be removed and depending on the contraction order it is possible to get different  unpredictable routing results. I would suggest to rediscuss routing results for nodes with invalid  segments as it touches snapping functions [`HasValidEdge`](https://github.com/Project-OSRM/osrm-backend/blob/0843386d4b9c9cb1eb79f70f0f3f6b96265946a5/include/engine/geospatial_query.hpp#L565-L593) and [`MakePhantomNode`](https://github.com/Project-OSRM/osrm-backend/blob/0843386d4b9c9cb1eb79f70f0f3f6b96265946a5/include/engine/geospatial_query.hpp#L484-L492) and SCC information because filtering out invalid edges also invalidates components information in the R-tree.

I have finished refactoring as planned but results are not good to be merged. I would suggest to keep negative weight values and fix/recheck related issues without the refactoring. One thing that i would like to merge is a gdb printer command `svg facade` that saves facade graphs as svg images like
alternative_loop_38-fullpage

danpat:
Sounds like this PR is never going to land - @oxidase should we close it?

oxidase:
@danpat yes, PR is stale now, but the related issue ticket is still valid

danpat:
👍  Too bad this didn't work out as an approach, I know it was a lot of work.

Issue #3779, opened 2017-03-06
Graph data structure for MLD


TheMarex:
We need a data structure that makes the following operations fast:

1. Scanning all boundary arcs of a node at a certain level
2. Scanning all arcs to nodes of the same cell on level 1
3. Identify when a node is a boundary node on a certain level

For this I propose the following data structure that basically just relies on a `StaticGraph` but enforces a node ordering that is specific to the partition.

1. Sort all nodes by the following criterion:
     1. The highest level for which it is a boundary node (or never)
     2. The cell in which it is contained on that level
     3. Node ID (even better: BFS ID, but we can do that later)

2. Node IDs of boundary nodes at a certain level will be consecutive. That way we can easily implement a fast map: Node ID -> Boundary ID (row/column) for each cell using a vector for each level, without too much memory overhead.

3. We will get a lot of cache efficiency because the search on the level 0 base graph will only be within one part of the adjacency array.

We can create this in the partitioner just ~after we have determined the `MultiLevelPartition` and~ before we will create the `CellStorage`. `CellStorage` will then depend on the graph having a certain ID ordering to facilitate fast lookups.

EDIT: We need to do this before the `MultiLevelPartition` because we store data reference by node ID.

/cc @oxidase

TheMarex:
Will try to sketch this data structure out today.

TheMarex:
Resorting the edges makes us dependent on https://github.com/Project-OSRM/osrm-backend/issues/3737 since the current format of the lookup file assumes edges are always sorted in the way we emit them in the `EdgeBasedGraphFactory`.

TheMarex:
Thinking a little bit more on how to do this, I think it can be done bypassing some of the structural problems here:

1. What we want to re-number are the `EdgeBasedNode` ids. These are referenced in two files after executing `osrm-extract`:
    - `.fileIndex` stores a reference to the `forward`/`backward` node in the search graph.
    - `.ebg` stores edges as `(source, target, data)`

2. We already read the `.ebg` file in `osrm-partition` anyway, we could also renumber every edge and write it out again.

3. Loading the `.fileIndex` file in `osrm-partition` and renumbering it is kind of ugly but works.

4. There should be no downside from doing something like `osrm-extract` then run `osrm-partition` (modified the `.ebg` file) `osrm-contract` (reads the `.ebg` file modied by `osrm-partition`). Best case this will also improve cache locality for CH.

[...]

PR #4089, opened 2017-05-24
Renumber graph nodes after partitioning them

TheMarex:
# Issue

This implements the second part of https://github.com/Project-OSRM/osrm-backend/issues/3779

## Tasklist

 - [x] Merge #4056 
 - [x] renumber `.fileIndex`
 - [x] renumber `.ebg_node`
 - [x] review
 - [x] adjust for comments

## Requirements / Relations

This PR is rebased on #4056.

### DIFF
--- a/include/partition/edge_based_graph_reader.hpp
+++ b/include/partition/edge_based_graph_reader.hpp
@@ -4,10 +4,14 @@
 #include "partition/edge_based_graph.hpp"
 
 #include "extractor/edge_based_edge.hpp"
+#include "extractor/files.hpp"
 #include "storage/io.hpp"
 #include "util/coordinate.hpp"
 #include "util/dynamic_graph.hpp"
 #include "util/typedefs.hpp"
+
+#include <tbb/parallel_reduce.h>
+#include <tbb/parallel_sort.h>
 
 #include <cstdint>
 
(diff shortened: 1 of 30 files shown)


TheMarex:
Some tests are failing, need to investigate those.

include/partition/edge_based_graph_reader.hpp
```diff
@@ -125,55 +128,51 @@ std::vector<OutputEdgeT> prepareEdgesForUsageInGraph(std::vector<extractor::Edge
...
-    }
+    NodeID max_turn_id = 0;
+    auto range = tbb::blocked_range<NodeID>(0, edge_based_graph.GetNumberOfNodes());
+    tbb::parallel_for(range, [&](const auto range) {
+        for (auto node = range.begin(); node < range.end(); ++node)
+        {
+            for (auto edge : edge_based_graph.GetAdjacentEdgeRange(node))
+            {
+                const auto &data = edge_based_graph.GetEdgeData(edge);
+                max_turn_id = std::max(max_turn_id, data.turn_id);
```
oxidase: Here can be a data race. It can fixed by 
* `max_turn_id` is atomic (bad)
* use thread local values 
* use `tbb::parallel_reduce`

src/partition/renumber.cpp
```diff
@@ -0,0 +1,65 @@
...
+    {
+        const auto &partition = partitions[level - 1];
+        for (auto node : util::irange<NodeID>(0, graph.GetNumberOfNodes()))
+        {
+            for (auto edge : graph.GetAdjacentEdgeRange(node))
+            {
+                auto target = graph.GetTarget(edge);
+                if (partition[node] != partition[target])
+                {
+                    border_level[node] = level;
```
oxidase: but why not `border_level[node] = std::max(border_level[node], level);`?
or level is always non-decreasing -> assertion ` border_level[node] <= level;`

src/partition/renumber.cpp
```diff
@@ -0,0 +1,65 @@
...
+
+std::vector<std::uint32_t> makePermutation(const DynamicEdgeBasedGraph &graph,
+                                           const std::vector<Partition> &partitions)
+{
+    std::vector<std::uint32_t> ordering(graph.GetNumberOfNodes());
+    std::iota(ordering.begin(), ordering.end(), 0);
+
+    for (const auto &partition : partitions)
+    {
+        std::stable_sort(
```
oxidase: please add in words what is this sort is doing and

src/partition/renumber.cpp
```diff
@@ -0,0 +1,65 @@
...
+            ordering.begin(), ordering.end(), [&partition](const auto lhs, const auto rhs) {
+                return partition[lhs] < partition[rhs];
+            });
+    }
+
+    auto border_level = getHighestBorderLevel(graph, partitions);
+    std::stable_sort(
```
oxidase: .. what this sort adjusts after the first ordering.

include/util/dynamic_graph.hpp
```diff
@@ -309,6 +332,36 @@ template <typename EdgeDataT> class DynamicGraph
...
+        {
+            auto new_first_edge = new_edge_index;
+            // move all filled edges
+            for (auto edge : GetAdjacentEdgeRange(node))
+            {
+                edge_list[edge].target = old_to_new_node[edge_list[edge].target];
+                old_to_new_edge[edge] = new_edge_index++;
+            }
+            // and all adjacent empty edges
+            for (auto edge = EndEdges(node); edge < number_of_edges && isDummy(edge); edge++)
```
oxidase: just don't understand why dummy edges are needed? If it is only to have correct results of `GetAdjacentEdgeRange` and `EndEdges` then it is better to make an iteration over `node_array` directly and remove empty edges.

oxidase:
@TheMarex tests fail for CH algorithms only in cases where MLD ordering is really used, probably some CH data also has to be reordered or data preprocessing must be split between CH and MLD steps

TheMarex:
> tests fail for CH algorithms only in cases where MLD ordering is really used, probably some CH data also has to be reordered or data preprocessing must be split between CH and MLD steps

Ah good point forgot that running contract before partition is also an option. I'll add some code that detects and `.hsgr` file and reorders it as well if present, to guarantee consistency.

include/util/dynamic_graph.hpp
TheMarex: I mainly wanted the renumber operation to not interfere with the structure of the dynamic graph. Removing the dummy edges would compactify the structure but would have non-obvious side-effects like for example `InsertEdge` would now take `O(n)` because it needs to resize+copy the edge array and relocate the source node edges.

We might want an additional function `shrink_to_fit` or something like that, but I don't think we have good use for it right now.

include/partition/edge_based_graph_reader.hpp
TheMarex: Urg good catch, that was a stupid mistake on my side.

TheMarex:
Did some measurements today:

The memory access pattern for MLD improves significantly: It is concentrated around the low ids which are the border nodes. For CH we see some improved concentration as well. For MLD this translates into speedups, for CH there is no impact as far as I can tell.

### Memory acces

Some plots that show the node id space on the x-axis and the time-slice on the y-axis. Basically these are snapshots of the accessed graph nodes very 10000 operations.

## CH

Before:
memory_access_ch
After:
memory_access_ch_renumbered

## MLD

Before:
memory_access_mld
After:
memory_access_mld_renumbered

### Runtime

Left before renumbering, right after renumbering. These are for random routes on Bayern, measuring the whole HTTP request time with `osrm-runner`.

## CH
time_ch

## MLD
time_mld

TheMarex:
The only failing test is https://github.com/Project-OSRM/osrm-backend/blob/master/features/testbot/zero-speed-updates.feature#L28 one of the zero-speed segment tests. Curiously it now returns a match. Might be related to the fact that the order in which `osrm-contract` and `osrm-partition` are run was changed.

oxidase:
@TheMarex ~22% speedup just from renumbering looks really awesome! I have just few remarks:
* histogram values can be visualized in log-scale otherwise it is almost pure blue pattern
* it is better to check speedup and access patterns on large-range >1000km qureies
* `zero-speed-updates.feature` tests really depends on snapping and renumbering changes nodes order, so either a oneway flag or bearing must be added to the test. Or simply add `@todo` and "FIXME: #4086"

TheMarex:
@oxidase investigated the failing test case:

1. We start on the reverse node (valid source and target since we only change the weight of `2,3` not `3,2`)
2. We do a u-turn on the forward node (both forward and reverse search)
3. Search terminates since both forward + reverse search have met

This yields a path that includes an invalid weight.

I don't get how this test case was ever passing. There are only two nodes in the graph, renumbering should not have any effect on the graph structure. Also the rtree snapping is unique in this case since we don't place the coordinates on the nodes but in the middle of the segments.

oxidase:
@TheMarex idk atm why the response is `Ok` instead of `NoMatch`, basically these tests must be revisited in #4086 

In the general a problem may occur if source node is a meeting node than a negative weight value of the source node  is used to compute [`new_weight`](https://github.com/Project-OSRM/osrm-backend/blob/master/include/engine/routing_algorithms/routing_base_ch.hpp#L130):
- If `new_weight < 0` the upper bound is ignored
- `new_weight >= 0` than incorrect path can be selected if`reverse_heap.GetKey(node)` >= `weight`

oxidase:
@TheMarex is the fix of the failing test here https://github.com/Project-OSRM/osrm-backend/blob/5c7288a9eb8b2557f4c429692150d66cb38c0e20/include/partition/edge_based_graph_reader.hpp#L174-L175 ? It seems the fix is needed also for #4086

TheMarex:
@oxidase #4086 should not be impacted because this bug only manifested when you wrote out the `.ebg` in `osrm-partition` again. Basically it was a mismatch between the format that `osrm-extract` produces and what `osrm-partition` wrote out after renumbering.

PR #4093, opened 2017-05-25
Misc warning fixes

danpat:
# Issue

These commits fix various compiler warnings that have crept in (under `clang`).

  - v8 includes should be included with `-isystem` to avoid warnings from code that is not ours
  - unused variables don't need to be named
  - one function that overrode a virtual didn't have `override` at the implementation
  - unused struct padding doesn't need to be named or initialized

## Tasklist
 - [x] review
 - [x] adjust for comments

### DIFF
--- a/cmake/NodeJS.cmake
+++ b/cmake/NodeJS.cmake
@@ -595,7 +595,7 @@ function(add_nodejs_module NAME)
         PUBLIC ${NODEJS_DEFINITIONS}
     )
     # This properly defines includes for the module
-    target_include_directories(${NAME} PUBLIC ${NODEJS_INCLUDE_DIRS})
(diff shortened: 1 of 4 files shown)


cmake/NodeJS.cmake
```diff
@@ -595,7 +595,7 @@ function(add_nodejs_module NAME)
         PUBLIC ${NODEJS_DEFINITIONS}
     )
     # This properly defines includes for the module
-    target_include_directories(${NAME} PUBLIC ${NODEJS_INCLUDE_DIRS})
+    target_include_directories(${NAME} SYSTEM PUBLIC ${NODEJS_INCLUDE_DIRS})
```
daniel-j-h: This is upstream code vendored from node-cmake (because they had an issue, should be fixed now with https://github.com/cjntaylor/node-cmake/pull/24) - the next update will remove your changes here again

Issue #3967, opened 2017-04-21
Refactor extract-conditionals tool libraries to read geojson timezone files


karenzshea:
Including `libshp-dev` and `libshp2` dependencies have proven to be difficult for the current setup.

### Blockers
- Ubuntu 14.04 Trusty ships with `libshp-dev` and `libshp1`, but is missing implementation for several functions, e.g. `DBFOpen`, `SHPReadObject`, `SHPClose` etc
- Upgrading to Xenial is a big lift for OSRM builds and users. I don't think we can justify the switch for shapefile support
- Using/making a ppa for `libshp2` would also be a comparably big lift 

### instead
For now we should switch to using geojson based version of the timezone files and switch to using a library like rapidjson to do the file loading.

- See if geojson files are small enough to roll into osrm-backend
- Replace usage of shapefile in `LoadLocalTimesRTree` to use rapidjson for loading geojson into rtree

cc @oxidase

daniel-j-h:
Looks like want to bundle rapidjson in the repo. now. A few questions wrt. that:

Is rapidjson packaged in Linux distributions we care about and can we use that?

If not, is it worth it to bundle rapidjson just for parsing geojson shapefiles? Or can we convert the shapefile polygones (?) into e.g. WKT we can then read with boost geometry's [`read_wkt`](http://www.boost.org/doc/libs/1_64_0/libs/geometry/doc/html/geometry/reference/io/wkt/read_wkt.html)?

PR #4094, opened 2017-05-26
Replace timezone shapefile parsing with geojson parsing

karenzshea:
# Issue

Closes https://github.com/Project-OSRM/osrm-backend/issues/3967

Shapefile dependency is finicky and possessive. This implements timezone parsing with rapidjson instead! Thus expecting timezone files to be provided as geojson files...

## Tasklist
 - [x] Remove code relating to shapefile that can be removed
 - [x] add unit tests for geojson parsing
 - [x] review
 - [ ] adjust for comments

## Requirements / Relations
5.8


daniel-j-h:
Looks like you're bundling rapidjson in the repo. now. A few questions wrt. that:

Is rapidjson packaged in Linux distributions we care about and can we use that?

If not, is it worth it to bundle rapidjson just for parsing geojson shapefiles? Or can we convert the shapefile polygones (?) into e.g. WKT we can then read with boost geometry's [`read_wkt`](http://www.boost.org/doc/libs/1_64_0/libs/geometry/doc/html/geometry/reference/io/wkt/read_wkt.html)?

src/util/timezones.cpp
```diff
@@ -16,6 +16,8 @@
 #include <string>
 #include <unordered_map>
 
+#include <time.h>
```
karenzshea: should we also #ifdef this line?

oxidase: it is ok both `localtime` versions are declared `time.h` in http://pubs.opengroup.org/onlinepubs/009695399/functions/localtime.html
somewhere the file was already included in non win32 builds

src/util/timezones.cpp
```diff
@@ -81,50 +88,75 @@ void Timezoner::LoadLocalTimesRTree(const std::string &tz_shapes_filename, std::
...
             polygons.emplace_back(boost::geometry::return_envelope<box_t>(polygon),
                                   local_times.size());
 
             // Get time zone name and emplace polygon and local time for the UTC input
-            const auto tzname = DBFReadStringAttribute(dbfhandle, shape, tzid);
-            local_times.emplace_back(local_time_t{polygon, get_local_time_in_tz(tzname)});
-
-            // std::cout << boost::geometry::dsv(boost::geometry::return_envelope<box_t>(polygon))
-            //           << " " << tzname << " " << asctime(&local_times.back().second);
+            const auto tzname =
```
oxidase: const auto &

CMakeLists.txt
```diff
@@ -447,11 +447,6 @@ if(ENABLE_MASON)
   add_dependency_includes(${MASON_PACKAGE_tbb_INCLUDE_DIRS})
   set(TBB_LIBRARIES ${MASON_PACKAGE_tbb_LDFLAGS})
 
-  mason_use(libshp2 VERSION ${MASON_LIBSHP_VERSION})
```
TheMarex: Yay 🎉 death to all shapefiles.

CMakeLists.txt
```diff
@@ -698,10 +685,11 @@ install(TARGETS osrm-components DESTINATION bin)
 if(BUILD_TOOLS)
   message(STATUS "Activating OSRM internal tools")
   add_executable(osrm-io-benchmark src/tools/io-benchmark.cpp $<TARGET_OBJECTS:UTIL>)
-  target_link_libraries(osrm-io-benchmark ${BOOST_BASE_LIBRARIES} ${MAYBE_SHAPEFILE})
+  target_link_libraries(osrm-io-benchmark ${BOOST_BASE_LIBRARIES})
 
   install(TARGETS osrm-io-benchmark DESTINATION bin)
 
+  find_package(Shapefile)
```
TheMarex: Could we integrate the geojson parser in the tool too?

include/util/geojson_validation.hpp
```diff
@@ -0,0 +1,84 @@
...
+#include "util/log.hpp"
+
+#include "rapidjson/document.h"
+
+namespace osrm
+{
+namespace util
+{
+
+inline void ValidateCoordinate(const rapidjson::Value &coordinate)
```
TheMarex: free functions should start with a small first letter: `validateCoordinate`.

include/util/geojson_validation.hpp
```diff
@@ -0,0 +1,84 @@
...
+    {
+        for (rapidjson::SizeType i = 0; i < coordinate.Size(); i++)
+        {
+            if (!coordinate[i].IsNumber() && !coordinate[i].IsDouble())
+                throw osrm::util::exception("Feature geometry has a non-number coordinate.");
+        }
+    }
+}
+
+inline void ValidateFeature(const rapidjson::Value &feature)
```
TheMarex: Nice! Resilient code. Function name should be `validateFeature` as above.

TheMarex:
As per chat with @karenzshea: Fixed the remaining PR issue to get this merged in time for the pre-release of 5.8.0.

Issue #3968, opened 2017-04-21
File write/out-of-disk problems - don't throw exception to main


danpat:
Currently, the exception handling strategy for `osrm-extract` et. al. is to let exceptions propagate all the way up the stack and trigger the C++ `std::terminate()` built-in.

By default, `std::terminate()` logs the exception to `stdout`, then calls `std::abort()`, which in turns kills the process with `SIGABRT`.  [`SIGABRT`](http://man7.org/linux/man-pages/man7/signal.7.html) is one of the signals that will trigger a core dump, which is handy for debugging some exceptions.

However, we're currently throwing some exceptions for which we could be doing normal error handling.  Specifically, when our `FileWriter` encounters writing problems due to out-of-disk conditions, there's no reason to have such ugly error handling - we could throw a specific exception, then handle it gracefully at `main()`, returning a specific error code and making integration with large-scale production infrastructure easier.

The specific action here:  change our tooling throw a specific exception for file write problems, catch it in main, log an error, and return a specific error code, rather than just letting the exception trigger `std::terminate()`.

danpat:
Hmm.  I'm not seeing a good way to change this behaviour without breaking our ABI.

We currently document that things like `osrm::extractor::extract(..)` will throw `osrm::util::exception` and `osmium::io_error`, and it has a `void` return type.

Given that, I'm tempted to just leave this ticket for a later release when we can change the ABI.

@TheMarex thoughts?

TheMarex:
Given that I get alarms daily concerning this I would naturally like to fix a this in `5.8.0`.

I don't think going `void` -> `int` is a breaking API change.

danpat:
A little bit of research revealed that while it's not defined by a standard, no C++ compilers use the return type as part of the mangled symbol name.  It generally doesn't make any sense to do so anyway, because it would indicate support for  ambiguous function overloading.

We should be ABI and source-code safe to change the return type of these externally visible functions.  I'll go ahead and implement that, and document the return codes.

TheMarex:
Fixed by #4096.

PR #4096, opened 2017-05-27
Catch well-known exceptions and return defined error codes

danpat:
# Issue

This PR is to address https://github.com/Project-OSRM/osrm-backend/issues/3968.  Currently, if any of the OSRM tools (`osrm-extract`, `osrm-contract`, etc) come across an error, we throw `osrm::util::exception`.  These don't get caught, so they trigger the default C++ [terminate handler](http://en.cppreference.com/w/cpp/error/terminate).  The default handler calls `abort()` on the current process.  In turn, this kills the process by sending it `SIGABRT`.

While this all sort of works, it's pretty messy, and doesn't leave glue code much room to figure out what went wrong, and possibly fix it.

This PR modifies the exception heirarchy, and creates a class of catchable-but-non-recoverable exceptions.  The tools have been modified to have more graceful handling of these new exceptions.

The practical outcome of this is that:

  - file permission errors
  - missing file problems
  - corrupt data problems
  - incorrect file problems
  - truncated file problems

will now print a clean error message and exit with a non-zero error code.  Users of `libosrm` will get a `osrm::RuntimeError` that can be caught for cleaner logging purposes, instead of the catch-all `osrm::exception` (although `osrm::RuntimeError` is a subclass of `osrm::exception`, so users already catching `osrm::exception` will not need any modification).

The new `osrm::RuntimeError` contains a more detailed error message, and a code value (similar to `std::runtime_error`).  The tools can use the code value as the `main()` return code, allowing management scripts to perform custom error recovery depending on the error value.

`osrm::exception` remains for "other" exceptions, where we don't want neat-and-tidy handling in various `src/tools/*.cpp` classes, and throwing `osrm::exception` will continue to trigger the C++ terminate handler.

## Tasklist
 - [ ] review
 - [ ] adjust for comments

### DIFF
--- a/include/storage/io.hpp
+++ b/include/storage/io.hpp
@@ -1,6 +1,8 @@
 #ifndef OSRM_STORAGE_IO_HPP_
 #define OSRM_STORAGE_IO_HPP_
 
+#include "osrm/error_codes.hpp"
+
 #include "util/exception.hpp"
 #include "util/exception_utils.hpp"
 #include "util/fingerprint.hpp"
@@ -10,6 +12,8 @@
 #include <boost/filesystem/fstream.hpp>
 #include <boost/iostreams/seek.hpp>
 
+#include <cerrno>
+#include <cstring>
 #include <cstring>
 #include <tuple>
 #include <type_traits>
@@ -49,12 +53,16 @@ class FileReader
         : filepath(filepath_), fingerprint(flag)
     {
         input_stream.open(filepath, std::ios::binary);
+
+        // Note: filepath.string() is wrapped in std::string() because it can
+        // return char * on some platforms, which makes the + operator not work
         if (!input_stream)
-            throw util::exception("Error opening " + filepath.string());
+            throw util::RuntimeError(
+                filepath.string(), ErrorCode::FileOpenError, SOURCE_REF, std::strerror(errno));
 
         if (flag == VerifyFingerprint && !ReadAndCheckFingerprint())
         {
-            throw util::exception("Fingerprint mismatch in " + filepath_.string() + SOURCE_REF);
+            throw util::RuntimeError(filepath.string(), ErrorCode::InvalidFingerprint, SOURCE_REF);
         }
     }
 
@@ -66,7 +74,11 @@ class FileReader
 
         if (file_size == boost::filesystem::ifstream::pos_type(-1))
         {
-            throw util::exception("File size for " + filepath.string() + " failed " + SOURCE_REF);
+            throw util::RuntimeError("Unable to determine file size for " +
+                                         std::string(filepath.string()),
+                                     ErrorCode::FileIOError,
+                                     SOURCE_REF,
+                                     std::strerror(errno));
         }
 
         // restore the current position
@@ -101,10 +113,11 @@ class FileReader
         {
             if (result.eof())
             {
-                throw util::exception("Error reading from " + filepath.string() +
-                                      ": Unexpected end of file " + SOURCE_REF);
+                throw util::RuntimeError(
+                    filepath.string(), ErrorCode::UnexpectedEndOfFile, SOURCE_REF);
             }
-            throw util::exception("Error reading from " + filepath.string() + " " + SOURCE_REF);
+            throw util::RuntimeError(
+                filepath.string(), ErrorCode::FileReadError, SOURCE_REF, std::strerror(errno));
         }
     }
 
@@ -145,23 +158,19 @@ class FileReader
 
         if (!loaded_fingerprint.IsValid())
         {
-            util::Log(logERROR) << "Fingerprint magic number or checksum is invalid in "
-                                << filepath.string();
-            return false;
+            throw util::RuntimeError(filepath.string(), ErrorCode::InvalidFingerprint, SOURCE_REF);
         }
 
         if (!expected_fingerprint.IsDataCompatible(loaded_fingerprint))
         {
-            util::Log(logERROR) << filepath.string()
-                                << " is not compatible with this version of OSRM";
-
-            util::Log(logERROR) << "It was prepared with OSRM "
-                                << loaded_fingerprint.GetMajorVersion() << "."
-                                << loaded_fingerprint.GetMinorVersion() << "."
-                                << loaded_fingerprint.GetPatchVersion() << " but you are running "
-                                << OSRM_VERSION;
-            util::Log(logERROR) << "Data is only compatible between minor releases.";
-            return false;
+            const std::string fileversion =
+                std::to_string(loaded_fingerprint.GetMajorVersion()) + "." +
+                std::to_string(loaded_fingerprint.GetMinorVersion()) + "." +
+                std::to_string(loaded_fingerprint.GetPatchVersion());
+            throw util::RuntimeError(std::string(filepath.string()) + " prepared with OSRM " +
+                                         fileversion + " but this is " + OSRM_VERSION,
+                                     ErrorCode::IncompatibleFileVersion,
+                                     SOURCE_REF);
         }
 
         return true;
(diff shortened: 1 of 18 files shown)


include/osrm/errorcodes.hpp
```diff
@@ -0,0 +1,33 @@
...
+    NoError = 0u,
+    InvalidFingerprint,
+    IncompatibleFileVersion,
+    MissingFile,
+    FileOpenError,
+    FileReadError,
+    FileWriteError,
+    FileIOError,
+    UnexpectedEndOfFile,
+    __ENDMARKER__ // Leave this value at the end
```
daniel-j-h: __Uppercase* is reserved for the stdlib

src/osrm/errorcodes.cpp
```diff
@@ -0,0 +1,46 @@
...
+    "No Error",                                       // NoError
+    "Fingerpring did not match the expected value",   // InvalidFingerprint
+    "File is incompatible with this version of OSRM", // IncompatibleFileVersion
+    "File is missing",                                // MissingFile
+    "Problem opening file",                           // FileOpenError
+    "Problem reading from file",                      // FileReadError
+    "Problem writing to file",                        // FileWriteError
+    "I/O error occurred",                             // FileIOError
+    "Unexpected end of file",                         // UnexpectedEndOfFile
+};
```
daniel-j-h: Why duplicated descriptions?

TheMarex:
I thought the goal here was to make `osrm-extract`/`osrm-contract`/`osrm-partition`/`osrm-customize` return `4` in case of file errors?

I agree with Daniel that this should just be a custom exception + try/catch + translation at the highest level.

danpat:
@daniel-j-h My main concern was the use of the `final` keyword on `osrm::exception` - I had assumed that changing that would trigger an non-backward compatible ABI change on our exported symbols.

Documentation on how `final` interacts with the ABI is sparse, the best I found was https://stackoverflow.com/a/11941854, which indicates that `final` has no effect on the ABI.

I'll assume this is correct and re-do this using an exception heirarchy.  That was how I implemented it initially, until I scared myself with ABI breakage risk.  I do not want this change to trigger a major version bump.

daniel-j-h:
Sadly there is no ABI compatibility with C++ :/

We can talk about bringing in https://github.com/daniel-j-h/libosrmc if you'd like to :)

include/osrm/errorcodes.hpp
```diff
@@ -0,0 +1,32 @@
...
+{
+
+/**
+ * Various error codes that can be returned by OSRM internal functions.
+ * Note: often, these translate into return codes from `int main()` functions.
+ *       Thus, do not change the order - if adding new codes, append them to the
+ *       end (but leave _ENDMARKER_ in the last position), so the code values do
+ *       not change for users that are checking for certain values.
+ */
+enum ErrorCode
```
daniel-j-h: Hmm not sure why you need to `static_cast<int>` the enum below this should be backed by `int` by default

[`Unscoped Enums/1`](http://en.cppreference.com/w/cpp/language/enum)

include/storage/io.hpp
```diff
@@ -192,7 +199,11 @@ class FileWriter
     {
         output_stream.open(filepath, std::ios::binary);
         if (!output_stream)
-            throw util::exception("Error opening " + filepath.string());
+        {
+            throw util::FileWriteException(std::string(filepath.string()) + ": " +
+                                               std::string(std::strerror(errno)),
```
daniel-j-h: here and everywhere in this changeset - I don't think streams give you the guarantee to populate C's global `errno` var

include/util/exception.hpp
```diff
@@ -54,6 +56,219 @@ class exception final : public std::exception
     virtual void anchor() const;
     const std::string message;
 };
+
+/**
+ * Indicates a class of error that occurred that was caused by some kind of
+ * external input (common examples are out of disk space, file permission errors,
+ * user supplied bad data, etc).
+ */
+class runtime_error : public std::runtime_error
```
daniel-j-h: Why do we inherit from `std::runtime_error` when we store a message and sourceref in this object in addition to the base's `what` string?

include/util/exception.hpp
```diff
@@ -54,6 +56,219 @@ class exception final : public std::exception
...
+{
+    using Base = std::runtime_error;
+
+  protected:
+    explicit runtime_error(const char *message,
+                           const std::string &sourceref,
+                           const std::string &code_description)
+        : Base(message), sourceref(sourceref)
+    {
+        messagebuf = code_description + ": " + Base::what() + " (at " + sourceref + ")";
```
daniel-j-h: Why `Base::what()` here? Won't this be always empty since we're in the progress of creating an exception here?

include/util/exception.hpp
```diff
@@ -54,6 +56,219 @@ class exception final : public std::exception
...
+    // This function exists to 'anchor' the class, and stop the compiler from
+    // copying vtable and RTTI info into every object file that includes
+    // this header. (Caught by -Wweak-vtables under Clang.)
+    virtual void anchor() const;
+    const std::string sourceref; // Source location where the exception was raised
+    std::string messagebuf;
+};
+
+// TODO: the classes below  are dying for templatization, but I don't know of a good pattern to
+// instantiate a template with string literals that isn't hella ugly.
```
daniel-j-h: There isn't if you don't want to do workarounds such as

https://gist.github.com/daniel-j-h/6bc186ff832647bb128895534fce61bb

include/util/exception.hpp
daniel-j-h: Also should be `RuntimeError`

include/util/exception.hpp
```diff
@@ -54,6 +56,219 @@ class exception final : public std::exception
...
+    using Base::Base;
+
+  public:
+    UnknownAlgorithmException(const std::string &message, const std::string &sourceref)
+        : Base(message, sourceref, GetCodeDescription())
+    {
+    }
+    ErrorCode GetCode() const override { return ErrorCode::UnknownAlgorithm; }
+    std::string GetCodeDescription() const override { return "Unrecognized algorithm"; }
+};
```
daniel-j-h: The stdlib does smth similar with `std::system_error` and `std::error_code`

include/storage/io.hpp
danpat: Yes, I'm aware it's not guaranteed.  Including this value is a calculated risk.  I'll change the wording to indicate that it's not necessarily the correct value, but my tests show that it is correct in the cases we care about, on current `libstdc++` implementations.

Something like `Error opening filename.osrm: Possibly caused by Permission Denied`.

include/osrm/extractor.hpp
```diff
@@ -28,6 +28,8 @@ SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 #ifndef OSRM_EXTRACTOR_HPP
 #define OSRM_EXTRACTOR_HPP
 
+#include "osrm/errorcodes.hpp"
```
TheMarex: The file should follow the convention of `under_score`: `error_codes.hpp`

include/util/exception.hpp
```diff
@@ -54,6 +57,61 @@ class exception final : public std::exception
...
+{
+    using Base = exception;
+    using Base::Base;
+
+  public:
+    explicit RuntimeError(const std::string &message,
+                          const ErrorCode code_,
+                          const std::string &sourceref,
+                          const char *possiblecause = nullptr)
+        : Base(std::string(ErrorDescriptions[code_]) + ": " + std::string(message) +
```
TheMarex: For readability you could wrap the string construction in an own function.

include/osrm/errorcodes.hpp
```diff
@@ -0,0 +1,37 @@
...
+/**
+ * Various error codes that can be returned by OSRM internal functions.
+ * Note: often, these translate into return codes from `int main()` functions.
+ *       Thus, do not change the order - if adding new codes, append them to the
+ *       end, so the code values do not change for users that are checking for
+ *       certain values.
+ */
+enum ErrorCode
+{
+    InvalidFingerprint = 1,
```
TheMarex: This should start at 2 to not collide with the generic `EXIT_FAILURE`.

Issue #4097, opened 2017-05-27
Truncated `.fileIndex` file fails siliently.


TheMarex:
Currently hit the case of a `.fileIndex` file that was truncated during RTree creation but, we did not error.

@danpat pointed out he fixed this recently in #4056, so we should backport that fix.

TheMarex:
Closed by #4098.

PR #4098, opened 2017-05-28
Add error checking for `.ramIndex` and `.names`

danpat:
# Issue

This PR is to address https://github.com/Project-OSRM/osrm-backend/issues/4097 where we were failing to check for errors while writing out the `.fileIndex` file in the RTree.

The LeafNode fix in this PR does *not* need to be ported to `master`, it is already included in https://github.com/Project-OSRM/osrm-backend/pull/4056 which should be ready to merge soon.

The second commit for the `.names` file should be backported to master.

In both cases, I've simply converted from using `::ofstream()` for direct file writing to using our wrapper class `storage::io::FileWriter()` which performs error checking for writes.  This will prevent write errors from being silently ignored.

I scanned the codebase for any other un-checked IO behaviour - I only found the `.names` data, everything else looks good.

## Tasklist
 - [x] review
 - [x] adjust for comments

### DIFF
--- a/include/storage/io.hpp
+++ b/include/storage/io.hpp
@@ -227,7 +227,7 @@ class FileWriter
 
     template <typename T> void WriteFrom(const T &src) { WriteFrom(&src, 1); }
 
-    template <typename T> void WriteOne(const T tmp) { WriteFrom(tmp); }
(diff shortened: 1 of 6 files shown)


TheMarex:
For documentation purposes I would like to merge this both to `master` and `5.7` and rebase #4097 on top of it.

PR #4105, opened 2017-05-30
Uses node-cmake's CMake module directly, resolves #3912

### LINKED ISSUE #3912: No longer vendor a custom version of NodeJS.cmake
In https://github.com/Project-OSRM/osrm-backend/pull/3909 we upgraded to node-cmake v2.

Unfortunately there is an issue with Debian and Ubuntu systems requiring us to vendor the NodeJS.cmake CMake module at `cmake/NodeJS.cmake`.

Once this issue is resolved upstream (https://github.com/cjntaylor/node-cmake/pull/24) no longer vendor it but let CMake pick it up directly.

daniel-j-h:
No need for vendoring.


package.json
```diff
@@ -4,8 +4,8 @@
   "private": false,
   "description": "The Open Source Routing Machine is a high performance routing engine written in C++14 designed to run on OpenStreetMap data.",
   "dependencies": {
-    "nan": "^2.1.0",
-    "node-cmake": "^2.2.1",
+    "nan": "^2.6.2",
+    "node-cmake": "^2.3.2",
     "node-pre-gyp": "^0.6.34"
```
oxidase: This will require an update of `yarn.lock`

daniel-j-h:
Thanks for the yarn heads-up - I finally made the switch to `yarn` locally now :P I just synced the lockfile; I also added a yarn check to Travis bailing out if we're out of sync. Can you check if this works for you.

oxidase:
@daniel-j-h thanks, it works!

Issue #4067, opened 2017-05-19
tests in features/bicycle/turn_penaties.features uses testbot profile


emiltin:
the test should therefore be moved to feature/testbot/

Issue #4061, opened 2017-05-18
via test is in the front folder


emiltin:
the test features/testbot/via.feature:228 uses the car profile.
the test should be placed in feature/car/via.feature

PR #4106, opened 2017-05-30
Puts tests into corresponding profile subdir

daniel-j-h:
@emiltin for #4067 #4061

### DIFF
--- a/features/car/approach.feature
+++ b/features/car/approach.feature
@@ -1,11 +1,11 @@
-@routing @approach_param @testbot
+@routing @approach
 Feature: Approach parameter
 
     Background:
         Given the profile "car"
         And a grid size of 10 meters
 
-    Scenario: Testbot - Start End same approach, option unrestricted for Start and End
+    Scenario: Start End same approach, option unrestricted for Start and End
         Given the node map
             """
                s        e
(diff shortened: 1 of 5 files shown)


features/testbot/turn_penalty.feature
```diff
@@ -0,0 +1,36 @@
+@routing @testbot @turn_penalty
+Feature: Turn Penalties
+
+    Background:
+        Given the profile "turnbot"
+        Given a grid size of 200 meters
+
+
+    Scenario: Bike - turns should incur a delay that depend on the angle
```
emiltin: The "Bike" label here is incorrect

features/car/approach.feature
```diff
@@ -1,11 +1,11 @@
-@routing @approach_param @testbot
+@routing @approach
 Feature: Approach parameter
 
     Background:
         Given the profile "car"
         And a grid size of 10 meters
 
-    Scenario: Testbot - Start End same approach, option unrestricted for Start and End
+    Scenario: Start End same approach, option unrestricted for Start and End
```
emiltin: Previously the standard has been include the profile name at the beginning of the senario name. Dpn't have a strong opinion about it but might want to be consistent.

daniel-j-h:
Thanks for the review - should be good to go now.

Issue #4107, opened 2017-05-30
Roads crossing roundabouts


oxidase:
PR #3977 changed invalidation of roads that cross roundabouts, but instructions at the intersections will be 
`Enter the roundabout and take the 1st exit` as on https://cloud.githubusercontent.com/assets/4421046/25481703/b0a32ac4-2b4e-11e7-9afd-87eb30a2d518.png

This case can be analyzed in [preprocessing](https://github.com/Project-OSRM/osrm-backend/blob/b7ee7b17c237c9ded669736e731ffd49f3f09239/src/extractor/guidance/roundabout_handler.cpp#L527-L528) and instead of `EnterAndExit{Roundabout,Rotary,RoundaboutIntersection}` turns that fallback to `Enter{Roundabout,Rotary,RoundaboutIntersection}` in post-processing  generate a `Continue` turn.

Fixing the issue assumes expectations changes in [`Roundabout with a crossing road`](
https://github.com/Project-OSRM/osrm-backend/blob/b7ee7b17c237c9ded669736e731ffd49f3f09239/features/guidance/roundabout.feature#L842-L843)

Issue #4100, opened 2017-05-29
On Master: invalid maneuvers in responses (crashing demo server)


daniel-j-h:
The demo server currently sees hundreds of assertions per minute

```
[assert][140005675415296] /mnt/ebs/osrm/2017-05-26-173001/src/engine/api/json_factory.cpp:186
in: osrm::util::json::Object osrm::engine::api::json::makeStepManeuver(const osrm::engine::guidance::StepManeuver&): maneuver_type != "invalid"
unexpected invalid maneuver type
```

Maybe related to
- https://github.com/Project-OSRM/osrm-backend/pull/4077
- https://github.com/Project-OSRM/osrm-backend/pull/4073
- https://github.com/Project-OSRM/osrm-backend/issues/4030

daniel-j-h:
Flagging: some of those commits are in the `5.7` branch. Not sure yet if we're affected though.

cc @TheMarex @oxidase

TheMarex:
The above request works fine on `5.7`:

```json
[...]

PR #4108, opened 2017-05-30
Fix collapsing turns after crossing roundabout

oxidase:
# Issue

Fixes `collapseTurnInstructions` that ignores all instructions between an open roundabout and end of `steps`. The check is pretty conservative:

* finds a step that either `leavesRoundabout`, `entersRoundabout` or the last one
* if the found step is `leavesRoundabout` then previous behavior is conserved
* if the found step is not `leavesRoundabout` and `maneuver.exit == 1` then it is assumed that the exit step  is the current one

**Implicit assumption:** it is not possible to enter twice into the same roundabout, otherwise instructions that touch or cross roundabouts are indistinguishable.

## Tasklist
 - [x] review
 - [x] adjust for comments

## Requirements / Relations
Related issue #4100
Related discussion  #4107

### DIFF
--- a/features/guidance/roundabout.feature
+++ b/features/guidance/roundabout.feature
@@ -843,6 +843,6 @@ Feature: Basic Roundabout
 
 
         When I route I should get
-            | from | to | route                           | turns                                                                           | distance |
-            | e    | k  | ebds,ebds,ds,ufghl,gi,jhik,jhik | depart,rotary-exit-1,rotary-exit-1,rstur-exit-2,invalid right,turn right,arrive | 189.1m   |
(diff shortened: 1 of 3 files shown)


oxidase:
PR is not finished: `/route/v1/driving/-52.57666,-24.65344;-51.88763380050659,-23.432359372979395?overview=full&alternatives=true&steps=false&hints=;` still asserts at the same place

oxidase:
Removed skipped roundabouts as there is no `leavesRoundabout` steps after `postProcess` call. Also switched the demo server to the branch to prevent assertions until the PR is merged to master

PR #4111, opened 2017-05-31
Uses correct coefficients for local optimality and stretch test in ch alternatives

daniel-j-h:
The alpha constant is for the local optimality T-Test threshold.

Before we used epsilon for the T-Test threshold, but the epsilon
constant is meant to be used for the stretch test(s) only.

This changeset fixes the local optimality T-Test and uses the
epsilon constant for the two stretch tests:
- We test the stretch for the total route against epsilon and
- We test the detour against the epsilon now, too

We can discuss if the second stretch test should actually use
epsilon, too, or a adapted value of it - but definitely not alpha.

---

👴 🚗 ⚡️ <details>This is a Back To The Future reference - I hope you got it</details>

Seems like the t-test come in through [this](https://github.com/Project-OSRM/osrm-backend/commit/5ebc4b392f44f14b3deb025acef06ed050ce23b2) commit in 2012 and it introduced the epsilon constant.
The constant was then later on re-used for the stretch test.

### DIFF
--- a/features/testbot/alternative_loop.feature
+++ b/features/testbot/alternative_loop.feature
@@ -55,8 +55,8 @@ Feature: Alternative route
 
         When I route I should get
             | from | to | route    | alternative       |
-            | b    | c  | bc,bc    |                   |
-            #| c    | b  | bc,bc    |                   | # alternative path depends on phantom snapping order
+            | b    | c  | bc,bc    | ab,ae,ef,fd,cd,cd |
+            #| c    | b  | bc,bc    | cd,fd,ef,ae,ab,ab | # alternative path depends on phantom snapping order
             | 1    | c  | ab,bc,bc | ab,ae,ef,fd,cd,cd |
             #| c    | 1  | bc,ab    | cd,fd,ef,ae,ab    | # alternative path depends on phantom snapping order
             | 2 | c | bc,bc          |                   |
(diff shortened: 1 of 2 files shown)


TheMarex:
Could you run some random queries on `master` and this branch to show the effect of the change?

This should be easy to test with `./scripts/osrm-runner.js`. Below an example for `bayern-latest.osm.pbf` from Geofabrik.

```
/scripts/osrm-runner.js -p "/route/v1/driving/{};{}?alternatives=true" -b 9.371337890625,47.70976154266637,13.7109375,50.499452103967734 -n 1000 -f "$.routes.length"
```

The last column is the number of routes.

daniel-j-h:
Yes will do - the 0.25 is probably also too high. I already made the PR to highlight the issue.

daniel-j-h:
Looks like we find slightly more alternatives with this branch. I just followed your recommendation and used the osrm-runner script on Bavaria (with 10k queries):

```
$ python3 diag.py
master 14024
branch 15301
```

    node scripts/osrm-runner.js -s localhost:5000 -b 10.162,47.680,12.096,50.233 -p '/route/v1/driving/{};{}?alternatives=true' -f '$.routes.length' -n 10000 > queries-master.txt
    node scripts/osrm-runner.js -s localhost:5000 -b 10.162,47.680,12.096,50.233 -p '/route/v1/driving/{};{}?alternatives=true' -f '$.routes.length' -n 10000 -q queries-master.txt > queries-branch.txt

```
$ cat total.py
import csv

with open('queries-master.txt') as fp:
    reader = csv.reader(fp)
    total = sum([int(line[4]) for line in reader])
    print('master', total)

with open('queries-branch.txt') as fp:
    reader = csv.reader(fp)
    total = sum([int(line[4]) for line in reader])
    print('branch', total)
```

Issue #4122, opened 2017-06-02
Check for Algorithm Specific Data Files Upfront


daniel-j-h:
When running the new pipeline

    osrm-extract -p ../profiles/car.lua map.pbf
    osrm-partition map.osrm
    osrm-customize map.osrm
    osrm-routed map.osrm

but forgetting to specify the mld algorithm (via `osrm-routed --algorithm MLD map.osrm`) we start up the engine anyway only showing a warning about missing hsgr and core files as we always do in the mld pipeline.

We then crash hard internally on queries e.g. with

```
[assert][140497935607744] /tmp/osrm-backend/include/util/static_graph.hpp:144
in: osrm::util::StaticGraph<EdgeDataT, Ownership>::StaticGraph(osrm::util::StaticGraph<EdgeDataT, Ownership>::Vector<osrm::util::static_graph_details::NodeArrayEntry>, osrm::util::StaticGraph<EdgeDataT, Ownership>::Vector<osrm::util::static_graph_details::EdgeArrayEntry<EdgeDataT> >) [with EdgeDataT = osrm::contractor::QueryEdge::EdgeData; osrm::storage::Ownership Ownership = (osrm::storage::Ownership)1; osrm::util::StaticGraph<EdgeDataT, Ownership>::Vector<osrm::util::static_graph_details::NodeArrayEntry> = osrm::util::vector_view<osrm::util::static_graph_details::NodeArrayEntry>; osrm::util::StaticGraph<EdgeDataT, Ownership>::Vector<osrm::util::static_graph_details::EdgeArrayEntry<EdgeDataT> > = osrm::util::vector_view<osrm::util::static_graph_details::EdgeArrayEntry<osrm::contractor::QueryEdge::EdgeData> >]: !node_array.empty()
```

of with a segfault if assertions are not enabled.

---

`osrm-routed` should error out early if
- mld is selected and mld data files are missing or
- ch is selected (default) and ch data files are missing

or better: this should be checked in `libosrm` already and not be tied to `osrm-routed`.

TheMarex:
The that was supposed to test this has no implementation for MLD it seems:

https://github.com/Project-OSRM/osrm-backend/blob/master/include/engine/engine.hpp#L146-L202

TheMarex:
Fixed with #4127

PR #4127, opened 2017-06-05
Incompatible dataset abort

danpat:
# Issue

This PR addresses https://github.com/Project-OSRM/osrm-backend/issues/4122 by causing the `OSRM::OSRM()` constructor to throw an exception if the dataset that is being loaded is not compatible with the algorithm chosen.  Prior to this change, the failure wasn't seen until the first query attempted to access the loaded dataset.  It is generally preferable to be [fail-fast](https://en.wikipedia.org/wiki/Fail-fast).

## Tasklist
 - [x] review
 - [x] adjust for comments

### DIFF
--- a/src/tools/routed.cpp
+++ b/src/tools/routed.cpp
@@ -223,32 +223,16 @@ int main(int argc, const char *argv[]) try
     {
         config.storage_config = storage::StorageConfig(base_path);
     }
+    if (!config.use_shared_memory && !config.storage_config.IsValid())
+    {
+        util::Log(logERROR) << "Required files are missing, cannot continue";
+        return EXIT_FAILURE;
+    }
     if (!config.IsValid())
     {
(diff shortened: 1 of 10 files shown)


src/tools/routed.cpp
```diff
@@ -221,33 +221,17 @@ int main(int argc, const char *argv[]) try
...
+        util::Log(logERROR) << "Required files are missing, cannot continue";
+        return EXIT_FAILURE;
+    }
     if (!config.IsValid())
     {
         if (base_path.empty() != config.use_shared_memory)
         {
             util::Log(logWARNING) << "Path settings and shared memory conflicts.";
         }
-        else
```
TheMarex: This pushes the failure higher up to the `Storage` code?

TheMarex:
@danpat ⚠️ going to rebase to resolve conflic on master.

Issue #3883, opened 2017-03-30
Parallize generation of edge-expanded edges


TheMarex:
Currently [this section](https://github.com/Project-OSRM/osrm-backend/blob/master/src/extractor/edge_based_graph_factory.cpp#L316-L646) takes about 160 minutes on the OSM Planet on a fairly beefy server. The main problem is that this loop is 100% synchronous and also does IO. By parallelizing this we could cut down processing time by almost ~2.5 hours (160 min -> 10min on our 16 core server).

The main problems that need to be solved for parallelizing are the following:

- `turn_id` is created on-demand here https://github.com/Project-OSRM/osrm-backend/blob/master/src/extractor/edge_based_graph_factory.cpp#L545
- and implicitly used [here](https://github.com/Project-OSRM/osrm-backend/blob/master/src/extractor/edge_based_graph_factory.cpp#L559-L562) and [here](https://github.com/Project-OSRM/osrm-backend/blob/master/src/extractor/edge_based_graph_factory.cpp#L490-L517) by order
- we can't compute `turn_id` in advance because the number of turns is not known (this is the function that evaluates turn restrictions) and the turn restriction evaluation is conflated in the guidance code.

MichalPP:
could this issue be connected to #3814 ?

TheMarex:
@MichalPP only if your are concerned with the `turn_function` which is the only part of the lua profile that is invoked there.

oxidase:
@TheMarex `turn_id` should not be a big problem, because of a serial bottleneck due to data flushing. It can be done in batches as:
```
for (auto node_at_center_of_intersection = 0; node_at_center_of_intersection < GetNumberOfNodes; node_at_center_of_intersection +=batch_size)
```
1. **parallel** generate tuples `(incoming_edge, turn.eid, OriginalEdgeData)`
2. **serial** flatten TLS vectors
    a. flush `OriginalEdgeData` to the file
    b. assign `turn_id` as tuples `(turn_id, incoming_edge, turn.eid)`
    c. resize `turn_weight_penalties`, `turn_duration_penalties` and `m_edge_based_edge_list`
3. **parallel** generate edge-based edges for every `turn_id, incoming_edge, turn.eid`

[...]

PR #4131, opened 2017-06-07
Parallelize creation of edge-expanded edges

danpat:
# Issue

This PR is to address https://github.com/Project-OSRM/osrm-backend/issues/3883.

Currently, creation of the edge-expanded graph is single-threaded.  This PR changes that so that we analyze and generate all turns using all available CPU resources.

The crux of this change is:

  - it uses `tbb::blocked_range` to break the nodes in the node-based-graph into groups
  - within each group, we perform the same logic as before, saving results to thread-local buffers
  - at the end of each block, there's a global lock where we copy results serially into the primary buffers
  - as a final step after all blocks are complete, `turn_id` values are assigned to members of `m_edge_based_edge_list[].data.turn_id`

The global lock introduces a contention point that's currently unavoidable - ordering of the various data structures must remain consistent, so items have to be appended to the main lists at the same time.  However, most of the work is in the turn analysis, and that can operate on independent worker threads.

Because of the size of the `lookup::TurnIndexBlock`, it's not practical to buffer that data in RAM during edge generation, it needs to be written to disk as processing proceeds.  This precludes the use of `tbb::enumerable_thread_specific`, as we can't guarantee that it will iterate in the same order that we write the `TurnIndexBlock` data to disk during the processing run.

## Tasklist
 - [ ] review
 - [ ] adjust for comments

### DIFF
--- a/src/extractor/edge_based_graph_factory.cpp
+++ b/src/extractor/edge_based_graph_factory.cpp
@@ -30,6 +30,11 @@
 #include <string>
 #include <unordered_map>
 
+#include <tbb/blocked_range.h>
+#include <tbb/parallel_for.h>
+#include <tbb/pipeline.h>
+#include <tbb/task_scheduler_init.h>
+
 namespace osrm
 {
 namespace extractor
@@ -319,6 +324,7 @@ void EdgeBasedGraphFactory::GenerateEdgeExpandedEdges(
     const std::string &turn_duration_penalties_filename,
     const std::string &turn_penalties_index_filename)
 {
+
     util::Log() << "Generating edge-expanded edges ";
 
     std::size_t node_based_edge_counter = 0;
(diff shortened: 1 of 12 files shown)


danpat:
So far, the approach I've taken here works, but it's not super efficient.  Here are the results:

**Current `master`**
```
1496877991.572789 [info] Generating edge-expanded edges
..
1496878185.306065 [info] Created 116 entry classes and 20326 Bearing Classes
Total time = 193 seconds
```

**This branch with 2 threads:**
```
1496879229.241765 [info] Generating edge-expanded edges
...
1496879365.356916 [info] Created 116 entry classes and 20326 Bearing Classes
Total time = 136 seconds
```

**This branch with 4 threads:**
```
1496878804.228027 [info] Generating edge-expanded edges
...
1496878892.847690 [info] Created 116 entry classes and 20325 Bearing Classes

Total time = 88 seconds
```

So, the speedup is there, but it's less than NUM_CPUS x.  The three mutexes in play are definitely a source of contention.  In addition to the locking, the `entry_class_hash`, `bearing_class_hash` are small enough that they might be sources of [false sharing](https://en.wikipedia.org/wiki/False_sharing), which would further hurt performance.

danpat:
I repeated these tests on a bigger box with more CPU to see how it scales:

**Master**
```
master c3.4xlarge - single threaded
1496882067.392343 [info] Generating edge-expanded edges
...
1496882247.595650 [info] Created 116 entry classes and 20326 Bearing Classes

Total time = 180 seconds
```

**This branch with 16 threads:**
```
1496882718.509253 [info] Generating edge-expanded edges
...
1496882742.521072 [info] Created 116 entry classes and 20325 Bearing Classes

time = 24 seconds
```

I wasn't expecting it to continue scaling, but it appears that the mutexes are a relatively small part of the workload.

There's still a lot of data copying and locking happening - we can probably 2x these improvements if those can be reduced or removed.

danpat:
Looks like I missed a shared datastructure that needed protection.  After https://github.com/Project-OSRM/osrm-backend/pull/4131/commits/ccb6bc02b534800a7d55dc4a23e18b6cb2a2c670, the same test is:

**This branch with 16 threads:**
```
1496883687.653823 [info] Generating edge-expanded edges
...
1496883746.253245 [info] Created 116 entry classes and 20326 Bearing Classes

time = 58 seconds
```

src/extractor/edge_based_graph_factory.cpp
```diff
@@ -368,181 +372,267 @@ void EdgeBasedGraphFactory::GenerateEdgeExpandedEdges(
...
-            {
-                const NodeID node_along_road_entering =
-                    m_node_based_graph->GetTarget(outgoing_edge);
-
-                const auto incoming_edge = m_node_based_graph->FindEdge(
-                    node_along_road_entering, node_at_center_of_intersection);
-
-                if (m_node_based_graph->GetEdgeData(incoming_edge).reversed)
-                    continue;
+        tbb::mutex merge_mutex; // To control merging of all the data into the main data structures
```
TheMarex: Any particular reason you went with the `tbb` version here instead of the `std` variants of these?

src/extractor/edge_based_graph_factory.cpp
```diff
@@ -368,181 +372,267 @@ void EdgeBasedGraphFactory::GenerateEdgeExpandedEdges(
...
+        tbb::concurrent_vector<BearingClassID> threadsafe_bearing_class_by_node_based_node;
+
+        threadsafe_bearing_class_by_node_based_node.resize(
+            bearing_class_by_node_based_node.size(), std::numeric_limits<std::uint32_t>::max());
+
+        tbb::parallel_for(
+            tbb::blocked_range<NodeID>(0, m_node_based_graph->GetNumberOfNodes()),
+            [&](const tbb::blocked_range<NodeID> &intersection_node_range) {
+
+                // We capture the thread-local work in these objects, then flush
```
TheMarex: TBB has a nice pattern for the thread-local accumulation and serial merge: [`tbb::enumerable_thread_local`](https://software.intel.com/en-us/node/506244).

The `vector`s would be moved outside of the `parallel_for` loop as `tbb::enumerable_thread_specific` and instead of having a mutex in the `parallel_for` loop we could simply use a serial for loop after this block to iterate over all thread-local buffer and merge them.

src/extractor/edge_based_graph_factory.cpp
```diff
@@ -368,181 +372,267 @@ void EdgeBasedGraphFactory::GenerateEdgeExpandedEdges(
...
-                }(turn_classification.first);
-
-                const auto bearing_class_id =
-                    [&](const util::guidance::BearingClass bearing_class) {
-                        if (0 == bearing_class_hash.count(bearing_class))
                         {
-                            const auto id = static_cast<std::uint32_t>(bearing_class_hash.size());
-                            bearing_class_hash[bearing_class] = id;
-                            return id;
+                            tbb::mutex::scoped_lock lock(turn_lane_mutex);
```
TheMarex: Hrm why does this need a mutex? As far as I can see these should all be thread-local right?

src/extractor/edge_based_graph_factory.cpp
```diff
@@ -368,181 +372,267 @@ void EdgeBasedGraphFactory::GenerateEdgeExpandedEdges(
...
+                                node_along_road_entering, incoming_edge, std::move(intersection));
                         }
-                        else
+
+                        // the entry class depends on the turn, so we have to classify the
+                        // interesction for
+                        // every edge
+                        const auto turn_classification = classifyIntersection(intersection);
+
+                        const auto entry_class_id = [&](
```
TheMarex: I wonder if we could work around this explicit synchronization with a worker thread + queue + `std::future`.

Another idea would be a reader-writer lock maybe? Under the assumption that insertions will become very rare later, we would only need a shared-lock in most cases.

src/extractor/edge_based_graph_factory.cpp
```diff
@@ -368,181 +372,267 @@ void EdgeBasedGraphFactory::GenerateEdgeExpandedEdges(
...
-                const auto incoming_edge = m_node_based_graph->FindEdge(
-                    node_along_road_entering, node_at_center_of_intersection);
-
-                if (m_node_based_graph->GetEdgeData(incoming_edge).reversed)
-                    continue;
+        tbb::mutex merge_mutex; // To control merging of all the data into the main data structures
+        tbb::mutex bearing_hash_mutex; // To control entry into the two hash maps
+        tbb::mutex entry_hash_mutex;   // To control entry into the two hash maps
+        tbb::mutex turn_lane_mutex;    // To control entry into the two hash maps
+        tbb::concurrent_vector<BearingClassID> threadsafe_bearing_class_by_node_based_node;
```
TheMarex: @daniel-j-h told me that `tbb::concurrent_vector` is dog-slow. Why not the same accumulate + merge pattern for this as well?

src/extractor/edge_based_graph_factory.cpp
danpat: the `classifyIntersection` thingy has some internal state, it's not thread-safe :-(  Similar to the bearing class, it captures the `.size()` to assign ID values, so that operation needs to be atomic.

I'll look into refactoring it a bit to see if I can reduce the amount of logic that needs to be wrapped in the lock.

src/extractor/edge_based_graph_factory.cpp
danpat: I looked at that, but there's a problem with the `std::vector<lookup::TurnIndexBlock> local_turn_indexes;` - this data isn't buffered, it's written directly to disk.

It's not exactly clear why - it's not significantly bigger than any of the other data, the only reason I can think that it's directly written is that it's not used again, but other data is.

It's also not clear whether the iteration would work out the same if we have multiple `tbb::enumerable_thread_local` objects - the docs don't mention *how* the serial iteration later on proceeds.  There seems to be some hints (https://software.intel.com/en-us/node/506244) that it's somehow governed by the thread ID, but the ordering isn't specified anywhere I can find.  I'm a bit hesitant to depend on that behaviour if it's undocumented, even if it works.

src/extractor/edge_based_graph_factory.cpp
danpat: Done - this wasn't necessary.  The `bearing_class_by_node_based_node` vector was pre-allocated and we never access the same element twice - I've changed it to simply update the regular vector during the main loop, it should be safe.

src/extractor/edge_based_graph_factory.cpp
danpat: No, I was buried in `tbb::` docs, so it was an easy thing to type - I've switched this to `std::mutex`.

src/extractor/edge_based_graph_factory.cpp
danpat: This has been refactored into a `ConcurrentIDMap` class - there were 4 instances of this exact "allocate and ID and add it to a cache" behaviour.

src/extractor/edge_based_graph_factory.cpp
danpat: The reader/writer lock worked pretty well and seems quite performant - if I remove the locking, there is very little noticeable speedup (but a lot of data corruption).

danpat:
With the implementation of the ConcurrentIDMap wrapper, processing time for the `mexico-latest.osm.pbf` extract is back down to around 25 seconds.

src/extractor/edge_based_graph_factory.cpp
```diff
@@ -368,181 +375,237 @@ void EdgeBasedGraphFactory::GenerateEdgeExpandedEdges(
...
-            // edge. Therefore, we have to search all connected edges for edges entering `b`
-            for (const EdgeID outgoing_edge :
-                 m_node_based_graph->GetAdjacentEdgeRange(node_at_center_of_intersection))
-            {
-                const NodeID node_along_road_entering =
-                    m_node_based_graph->GetTarget(outgoing_edge);
+        std::mutex merge_mutex; // To control merging of all the data into the main data structures
+        std::uint64_t nodes_completed = 0;
+
+        tbb::parallel_for(
```
TheMarex: After using the `tbb::parallel_pipeline` in #4137 I think it would be very easy to adapt it for this use case. The biggest advantage over using that over a manual `parallel_for` is that we can force processing to be in-order. That would remove the non-determinism we introduced here.

danpat:
It looks like the "do everything in parallel" approach broke some downstream assumptions around ordering - we were seeing some intermittent test failures that looked like they were being triggered when we didn't populate the various data structures in serial 0-to-N order.

I've taken @TheMarex suggestion and re-implemented using the `tbb::parallel_pipeline` approach - there's a serial generator for intersection IDs, parallel processing of each intersection, then a serial serialization block that's tied to the ID generation, guaranteeing that we maintain the same order as the original code.

Processing times aren't *quite* as good as the don't-care-about-ordering approach, but it should fix any ordering-related problems we were seeing.  It's still a significant improvement over `master`:

```
16 threads pipeline mexico c3.8xlarge
1497425738.511850 [info] Generating edge-expanded edges
...
1497425776.218879 [info] Created 116 entry classes and 20326 Bearing Classes
= 37 sec

16 threads master
1497427760.351970 [info] Generating edge-expanded edges
...
1497427940.139016 [info] Created 116 entry classes and 20325 Bearing Classes
= 179 sec
```

It looks like we have a few test failures happening, I'm looking into those now.

src/extractor/edge_based_graph_factory.cpp
```diff
@@ -368,181 +377,279 @@ void EdgeBasedGraphFactory::GenerateEdgeExpandedEdges(
...
         // possible turns along these intersections.
-        for (const auto node_at_center_of_intersection :
-             util::irange(0u, m_node_based_graph->GetNumberOfNodes()))
+        std::mutex merge_mutex; // To control merging of all the data into the main data structures
+        std::uint64_t nodes_completed = 0;
+
+        const NodeID node_count = m_node_based_graph->GetNumberOfNodes();
+        NodeID current_node = 0;
+
+        tbb::filter_t<void, NodeID> intersection_ids(tbb::filter::serial_in_order,
```
TheMarex: Just a quick idea here: Maybe it's faster to work on `NodeID` chunks? TBB has the `tbb::blocked_range<NodeID,  BLOCKSIZE>` abstraction for that. Could reduce the overhead a little bit, but I'm not sure.

danpat: Good call.  I implemented this and it brings processing time down to 25 seconds for the `mexico-latest.osm.pbf` extract I've been testing - basically the same as the non-ordered original approach, but less buggy :-)

PR #4137, opened 2017-06-08
Parallize scripting on osmium::Buffer granularity

### LINKED ISSUE #3447: Change parsing parallelization to work on osmium::Buffer objects
Currently we read one buffer object [here](https://github.com/Project-OSRM/osrm-backend/blob/master/src/extractor/extractor.cpp#L172) and then split it apart [here](https://github.com/Project-OSRM/osrm-backend/blob/master/src/extractor/scripting_environment_lua.cpp#L307-L312).

This should be changed to operate on a Buffer-per-thread to save the cost of splitting the buffer apart and possibly unlock using the libosmium handler interface.

/cc @daniel-j-h @joto

TheMarex:
Fixes #3447 and reduces parsing time by about 15%.

## Tasklist
 - [x] review
 - [x] adjust for comments

### DIFF
--- a/src/extractor/extractor.cpp
+++ b/src/extractor/extractor.cpp
@@ -33,12 +33,13 @@
 
 #include <boost/filesystem.hpp>
 #include <boost/filesystem/fstream.hpp>
+#include <boost/iterator/function_input_iterator.hpp>
 #include <boost/optional/optional.hpp>
 #include <boost/scope_exit.hpp>
 
 #include <osmium/io/any_input.hpp>
 
-#include <tbb/concurrent_vector.h>
+#include <tbb/pipeline.h>
 #include <tbb/task_scheduler_init.h>
 
 #include <cstdlib>
@@ -252,11 +253,6 @@ std::vector<TurnRestriction> Extractor::ParseOSMData(ScriptingEnvironment &scrip
 
     timestamp_file.WriteFrom(timestamp.c_str(), timestamp.length());
 
-    // initialize vectors holding parsed objects
-    tbb::concurrent_vector<std::pair<std::size_t, ExtractionNode>> resulting_nodes;
-    tbb::concurrent_vector<std::pair<std::size_t, ExtractionWay>> resulting_ways;
-    tbb::concurrent_vector<boost::optional<InputRestrictionContainer>> resulting_restrictions;
-
     std::vector<std::string> restrictions = scripting_environment.GetRestrictions();
     // setup restriction parser
     const RestrictionParser restriction_parser(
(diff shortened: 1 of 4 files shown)


daniel-j-h:
Travis builds are failing,

https://www.threadingbuildingblocks.org/docs/help/hh_goto.htm?index.htm#reference/algorithms/parallel_for_each_func.html

Looks like the `parallel_for_each`'s body argument needs to be copy-constructible (at least in old TBB versions) and a `osmium::Buffer` is not.

TheMarex:
@daniel-j-h ideally I would like to use an iterator that reads the `Buffer` objects on demand from the reader (instead of doing explicit batches), so I could pass them directly into `tbb::parallel_for_each`. 32 is simply chosen because it's bigger then the usual number of threads.

However there is no interface for this in `libosmium`, it only allows to iterate over single items. @joto maybe you have a better idea on how to process Buffers in parallel without explicit batches?

src/extractor/extractor.cpp
```diff
@@ -252,58 +254,60 @@ std::vector<TurnRestriction> Extractor::ParseOSMData(ScriptingEnvironment &scrip
...
+                std::vector<std::pair<const osmium::Way &, ExtractionWay>> resulting_ways;
+                std::vector<boost::optional<InputRestrictionContainer>> resulting_restrictions;
+
+                scripting_environment.ProcessElements(buffer,
+                                                      restriction_parser,
+                                                      resulting_nodes,
+                                                      resulting_ways,
+                                                      resulting_restrictions);
+
+                // serial insertion
```
oxidase: I think better is to decouple `ProcesElements` and `Process{Node,Way,Restriction}` like in [`flatten2d` example](https://www.threadingbuildingblocks.org/docs/help/reference/thread_local_storage/flattened2d_cls.html)

The idea is to use `enumerable_thread_specific` vectors for `resulting_nodes`, `resulting_ways`, `resulting_restrictions`  and finish parallel processing here.
And below use `parallel_do` with 3 jobs over flattened vectors with `ProcessNode`, `ProcessWay`, `ProcessRestrictions` calls. It should work without explicit locks, but `ProcessNode`, `ProcessWay`, `ProcessRestrictions` should not use any shared mutable state.

src/extractor/extractor.cpp
```diff
@@ -252,58 +254,60 @@ std::vector<TurnRestriction> Extractor::ParseOSMData(ScriptingEnvironment &scrip
...
         config.parse_conditionals,
         restrictions);
 
-    // create a vector of iterators into the buffer
-    for (std::vector<osmium::memory::Buffer::const_iterator> osm_elements;
-         const osmium::memory::Buffer buffer = reader.read();
-         osm_elements.clear())
+    std::mutex process_mutex;
+
+    constexpr std::size_t BATCH_SIZE = 32;
```
oxidase: This value can be parameterized with number of threads as `k * #threads`

joto:
The way osmium and, more specifically, the reader and buffers work, the use I had in mind when multithreading was to loop in the main thread and give each buffer to a thread:

    while (auto buffer = reader.read()) {
        process_buffer(std::move(buffer));
    }

`process_buffer` can either start a new thread or queue the buffer for a worker thread from a pool to get and handle it. The buffer can only be moved, not copied. That is intentional, because it always makes clear who is responsible for it. After the thread is done with the buffer, it can simply be discarded. There is no guarantee about the contents or size of the buffer, it could even be empty. But in normal use they usually have 8000 objects in them. So on average this should work pretty well.

TheMarex:
@oxidase can you take a look again? I ported everything to TBB pipeline which gives the same speedup but removes the ugliness.

TheMarex:
@oxidase yeah, keeping the order has the nice side-effect that our processing becomes more deterministic. Utilization looks good, so I don't think the ordering constraint slows it down much.

Issue #4021, opened 2017-05-08
Expose Map Matching `gaps` and `tidy` parameters in Node.js bindings


daniel-j-h:
With [v5.7](https://github.com/Project-OSRM/osrm-backend/blob/master/CHANGELOG.md#570) there are two new map matching options (impl. was done in https://github.com/Project-OSRM/osrm-backend/pull/3815):
- option gaps=split|ignore to enable/disbale track splitting
- option tidy=true|false to simplify traces automatically

neither is exposed in the Node.js bindings.

Tasks:

- [ ] parse flags and set options [here](https://github.com/Project-OSRM/osrm-backend/blob/2f02384d22a7e5caac9e1918961d114158fab532/include/nodejs/node_osrm_support.hpp#L968-L1022)
- [ ] write docs in `docs/nodejs/api.md`

oxidase:
Closed via  #4145

PR #4145, opened 2017-06-11
Expose Map Matching `gaps` and `tidy` parameters in Node.js bindings

stepankuzmin:
Expose Map Matching `gaps` and `tidy` parameters in Node.js bindings according to #4021

 - [x]  parse flags and set options in node_osrm_support.hpp
 - [x]   write docs in docs/nodejs/api.md


daniel-j-h:
Thanks for implementing this!

Could you add some tests for the new map matching flags [here](https://github.com/Project-OSRM/osrm-backend/blob/master/test/nodejs/match.js) just that we got the bindings covered.

stepankuzmin:
Thanks for the reply @daniel-j-h!

I've added tests in 1410e02d2788bd49c51df57c698e5650f2835739. Is it enough for coverage?

daniel-j-h:
Great, thanks! ✨

PR #4147, opened 2017-06-12
Lua scripting speedups

danpat:
# Issue

After a bit of conversation with @joto, it looks like there are some easy speedups we can make with our Lua handling code to make `osrm-exract` run a lot faster.

  1) Don't make copies of `sol::function` objects for every call - this is simply unnecessary, and using thread-local cached copies gives us a 12-13% speedup very easily.

  2) Make calling the node_function optional.  If the profile property `call_tagless_node_function` is set to `false`, then the `node_function` won't get called if there are no tags on a node (you only need to call this in situations where you're doing coordinate lookups for nodes for some reason.  The default profiles don't do this).  Because there are more nodes than anything else, and most of them have no tags, this creates a speedup of 269 sec -> 135 seconds testing `us-west-latest.osm.pbf`.
  This change should be backwards compatible.  For those with old scripts that don't include the new `call_tagless_node_function` profile property, it will default to the old behaviour of calling the `node_function` for every node.  Users can add this property to their existing scripts to gain access to the speedup.

## Tasklist
 - [ ] review
 - [ ] adjust for comments

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@
     - Features
       - Added conditional restriction support with `parse-conditional-restrictions=true|false` to osrm-extract. This option saves conditional turn restrictions to the .restrictions file for parsing by contract later. Added `parse-conditionals-from-now=utc time stamp` and `--time-zone-file=/path/to/file`  to osrm-contract
       - Command-line tools (osrm-extract, osrm-contract, osrm-routed, etc) now return error codes and legible error messages for common problem scenarios, rather than ugly C++ crashes
+      - Speed up pre-processing by only running the Lua `node_function` for nodes that have tags.  Cuts OSM file parsing time in half.
     - Files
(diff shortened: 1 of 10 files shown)


danpat:
I did some additional testing with `us-west-latest.osm.pbf`.  If I check for `building` tags on ways in C++, it cuts a further 11 seconds (8%) off the OSM parsing phase of `osrm-extract`.  However, this seems a bit too special-casey and could possibly bite us on say, the `foot` profile.  I'll ticket this for later optimization (checking tag white/blacklists in C++), but I think this PR represents the best bang-for-buck for now.

danpat:
Note: processing `planet.osm.pbf` on an `r3.8xlarge` AWS instance currently spends 2700-3000 seconds doing OSM parsing/Lua script stuff.  This change cuts that roughly in half, saving about 20-25 minutes when processing the whole planet.

PR #4151, opened 2017-06-13
Added driveway handler

oxidase:
# Issue

The PR fixes steps collapsing for intersection of a main road with service roads, e.g. removes ['Turn left'](http://map.project-osrm.org/?z=18&center=52.150979%2C9.932730&loc=52.150469%2C9.931110&loc=52.151229%2C9.933132&hl=en&alt=0) instruction 
before
->
after

/cc @MoKob 

## Tasklist
 - [x] review
 - [x] adjust for comments

## Requirements / Relations

### DIFF
--- a/include/extractor/guidance/driveway_handler.hpp
+++ b/include/extractor/guidance/driveway_handler.hpp
@@ -1,0 +1,41 @@
+#ifndef OSRM_EXTRACTOR_GUIDANCE_DRIVEWAY_HANDLER_HPP
+#define OSRM_EXTRACTOR_GUIDANCE_DRIVEWAY_HANDLER_HPP
+
+#include "extractor/guidance/intersection_handler.hpp"
+
+namespace osrm
+{
+namespace extractor
+{
+namespace guidance
+{
+
+// Intersection handlers deal with all issues related to intersections.
+// They assign appropriate turn operations to the TurnOperations.
+class DrivewayHandler final : public IntersectionHandler
+{
+  public:
+    DrivewayHandler(const IntersectionGenerator &intersection_generator,
+                    const util::NodeBasedDynamicGraph &node_based_graph,
+                    const std::vector<util::Coordinate> &coordinates,
+                    const util::NameTable &name_table,
+                    const SuffixTable &street_name_suffix_table);
+
+    ~DrivewayHandler() override final = default;
+
+    // check whether the handler can actually handle the intersection
+    bool canProcess(const NodeID nid,
+                    const EdgeID via_eid,
+                    const Intersection &intersection) const override final;
+
+    // process the intersection
+    Intersection operator()(const NodeID nid,
+                            const EdgeID via_eid,
+                            Intersection intersection) const override final;
+};
+
+} // namespace guidance
+} // namespace extractor
+} // namespace osrm
+
+#endif /* OSRM_EXTRACTOR_GUIDANCE_DRIVEWAY_HANDLER_HPP */
--- a/include/extractor/guidance/driveway_handler.hpp
+++ b/include/extractor/guidance/driveway_handler.hpp
@@ -11,7 +11,6 @@ namespace guidance
 {
 
 // Intersection handlers deal with all issues related to intersections.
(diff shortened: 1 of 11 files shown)


include/extractor/guidance/driveway_handler.hpp
```diff
@@ -0,0 +1,41 @@
...
+
+namespace osrm
+{
+namespace extractor
+{
+namespace guidance
+{
+
+// Intersection handlers deal with all issues related to intersections.
+// They assign appropriate turn operations to the TurnOperations.
```
daniel-j-h: What is `TurnOperations`?

src/extractor/guidance/driveway_handler.cpp
```diff
@@ -0,0 +1,75 @@
...
+//       d
+//       .
+//  a--->b--->c
+//       .
+//       e
+bool DrivewayHandler::canProcess(const NodeID /*nid*/,
+                                 const EdgeID /*via_eid*/,
+                                 const Intersection &intersection) const
+{
+    const auto from_eid = intersection.front().eid;
```
daniel-j-h: should this be `getUTurnRoad().eid` instead of `front().eid`?

src/extractor/guidance/driveway_handler.cpp
```diff
@@ -0,0 +1,75 @@
...
+
+    if (intersection.size() <= 2 ||
+        node_based_graph.GetEdgeData(from_eid).road_classification.IsLowPriorityRoadClass())
+        return false;
```
daniel-j-h: So this handler only handles situations where the main road is a high-priority road and there is a low-priority road crossing?

src/extractor/guidance/driveway_handler.cpp
```diff
@@ -0,0 +1,75 @@
...
+        std::find_if(intersection.begin() + 1, intersection.end(), [this](const auto &road) {
+            return !node_based_graph.GetEdgeData(road.eid)
+                        .road_classification.IsLowPriorityRoadClass();
+        });
+
+    (void)nid;
+    OSRM_ASSERT(road != intersection.end(), coordinates[nid]);
+
+    road->instruction.type =
+        isSameName(source_edge_id, road->eid) ? TurnType::NoTurn : TurnType::NewName;
```
daniel-j-h: Hm this overwrites the road's instruction type - can there ever be a situation where we don't want to do this? For example what if this is a enter roundabout instruction?

src/extractor/guidance/sliproad_handler.cpp
```diff
@@ -473,17 +473,7 @@ operator()(const NodeID /*nid*/, const EdgeID source_edge_id, Intersection inter
...
-
-        const auto same_name = source_edge_data.name_id != EMPTY_NAMEID && //
-                               main_road_data.name_id != EMPTY_NAMEID &&   //
-                               !util::guidance::requiresNameAnnounced(source_edge_data.name_id,
-                                                                      main_road_data.name_id,
-                                                                      name_table,
-                                                                      street_name_suffix_table); //
-
-        if (same_name)
+        if (isSameName(source_edge_id, main_road.eid))
```
daniel-j-h: Thanks for refactoring this! :)

include/extractor/guidance/driveway_handler.hpp
oxidase: idk, it is c&p and looks like a stale comment in 4 places

src/extractor/guidance/driveway_handler.cpp
oxidase: yes

src/extractor/guidance/driveway_handler.cpp
oxidase: it should not be the case, but better to check, i will add an assertion

oxidase: I have added `OSRM_ASSERT(road->instruction.type == TurnType::Turn, coordinates[nid]);` and checked the car profile on planet :white_check_mark:

PR #4155, opened 2017-06-14
Add gdb printer for facade and guidance steps

oxidase:
# Issue

For visualization of edge-based graphs a facade gdb printer can be useful. For some arbitrary osm snippet a gdb command:
```
svg facade 10000x10000 
```
creates 10000x10000  SVG image that contains information about edge-based nodes, segments and edges map-fullpage

The command also supports bounding-box filtering:
```
svg facade 4800x3600 9.93,52.153;9.94,52.154
```
map1-fullpage

For guidance debugging another command `geojson steps`  creates an GeoJSON line query for http://geojson.io/ and annotates it with `RouteStep` data.
Eg., for #4151 the annotated path before is https://gist.github.com/oxidase/b40671805fb4941ba61747ef9a13d5f0 and after https://gist.github.com/oxidase/cfaaa09296f30e0620dd1c2c48926819

/cc @MoKob 

## Tasklist
 - [ ] review
 - [ ] adjust for comments

### DIFF
--- a/scripts/gdb_printers.py
+++ b/scripts/gdb_printers.py
@@ -4,6 +4,19 @@ import gdb.printing
 # https://sourceware.org/gdb/onlinedocs/gdb/Writing-a-Pretty_002dPrinter.html
 
 COORDINATE_PRECISION = 1e6
+coord2float = lambda x: int(x) / COORDINATE_PRECISION
+lonlat = lambda x: (coord2float(x['lon']['__value']), coord2float(x['lat']['__value']))
+
+def call(this, method, *args):
+    """Call this.method(args)"""
+    command = '(*({})({})).{}({})'.format(this.type.target().pointer(), this.address, method, ','.join((str(x) for x in args)))
+    return gdb.parse_and_eval(command)
+
+def iterate(v):
+    s, e = v['_M_impl']['_M_start'], v['_M_impl']['_M_finish']
+    while s != e:
+        yield s.dereference()
+        s +=1
 
 class CoordinatePrinter:
     """Print a CoordinatePrinter object."""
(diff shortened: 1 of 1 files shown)


scripts/gdb_printers.py
```diff
@@ -4,6 +4,8 @@
 # https://sourceware.org/gdb/onlinedocs/gdb/Writing-a-Pretty_002dPrinter.html
 
 COORDINATE_PRECISION = 1e6
+coord2float = lambda x: int(x) / COORDINATE_PRECISION
+lonlat = lambda x: (coord2float(x['lon']['__value']), coord2float(x['lat']['__value']))
```
MoKob: could we potentially use this within the `CoordinatePrinter` as well?

scripts/gdb_printers.py
```diff
@@ -59,5 +61,259 @@ def build_pretty_printer():
...
+        except gdb.error as e:
+            print('error: ' % (e.args[0] if len(e.args)>0 else 'unspecified'))
+            return
+
+GeojsonPrinter()
+
+
+
+class SVGPrinter (gdb.Command):
+    """Display features on geojson.io."""
```
MoKob: looks like a copy paste error?

scripts/gdb_printers.py
```diff
@@ -59,5 +61,259 @@ def build_pretty_printer():
...
+  </defs>
+""" + svg + '\n</svg></html>'
+        fd, name = tempfile.mkstemp('.html')
+        os.write(fd, svg.encode('utf-8'))
+        os.close(fd)
+        print ('Saved to ' + name)
+        webbrowser.open('file://' + name)
+
+    @staticmethod
+    def Facade(facade, width, height, bbox):
```
MoKob: Do you think we could maybe improve readability by splitting this into separate functions?

scripts/gdb_printers.py
```diff
@@ -59,5 +61,259 @@ def build_pretty_printer():
     pp.add_printer('TurnLaneData', '::TurnLaneData$', TurnLaneDataPrinter)
     return pp
 
-#gdb.pretty_printers = [filter(lambda x: x.name != 'OSRM', gdb.pretty_printers)]
```
MoKob: I don't really understand the purpose of this line, so I am not sure why you removed it. Could you give some context?

oxidase: it is needed to reload pretty printers, otherwise gdb module throws `RuntimeError: pretty-printer already registered: OSRM`. I will keep it commented, because it is needed only during script development

scripts/gdb_printers.py
oxidase: :+1: also added  some details here

Issue #4153, opened 2017-06-14
Debug tiles don't expose turn data for MLD algorithm


danpat:
When the dataset is using MLD, the `tile` plugin only exposes edge information, not turn data.

The logic for calculating turn information (angles, weights, durations) needs to be adapted for MLD and integrated into the tile plugin.

/cc @MoKob

PR #4157, opened 2017-06-14
add support for visualising turn penalties in MLD Debug tiles

MoKob:
# Issue

Resolves https://github.com/Project-OSRM/osrm-backend/issues/4153.

Adds support to the debug tiles generated with the MLD plugin. Opposed to the CH plugin, finding the edges that contain turn information is a bit simpler (we do not need to consider shortcuts but can call FindEdge directly on the base layer of MLD).

The PR refactors parts of the CH process to generate tiles to allow re-usage in the MLD plugin.
We simply need to be able to swap out how the edges are located.

## Tasklist
 - [x] refactor so that main function calls do not duplicate code
 - [x] review
 - [x] adjust for comments

/cc @danpat 

Tiles generated with CH:

screen shot 2017-06-14 at 16 52 00 2

Tiles generated with MLD:

screen shot 2017-06-14 at 16 58 33 2

### DIFF
--- a/src/engine/routing_algorithms/tile_turns.cpp
+++ b/src/engine/routing_algorithms/tile_turns.cpp
@@ -7,31 +7,33 @@ namespace engine
 namespace routing_algorithms
 {
 
-std::vector<TurnData>
-getTileTurns(const datafacade::ContiguousInternalMemoryDataFacade<ch::Algorithm> &facade,
-             const std::vector<RTreeLeaf> &edges,
-             const std::vector<std::size_t> &sorted_edge_indexes)
+namespace
 {
-    std::vector<TurnData> all_turn_data;
+// Struct to hold info on all the EdgeBasedNodes that are visible in our tile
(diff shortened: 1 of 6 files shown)


src/engine/routing_algorithms/tile_turns.cpp
```diff
@@ -41,8 +43,12 @@ getTileTurns(const datafacade::ContiguousInternalMemoryDataFacade<ch::Algorithm>
         return facade.GetGeometryIndex(edge.forward_segment_id.id).id;
     };
 
-    // Build an adjacency list for all the road segments visible in
-    // the tile
+    // To build a tile, we can only rely on the r-tree to quickly find all data visible within the
+    // tile
```
danpat: Comment formatting is a bit ugly here, looks like clang-format auto-wrapped.  Can you clean this up?

danpat:
Looks good, thanks @MoKob

PR #4160, opened 2017-06-15
Implements Intersection Validation Handler, resolves #4146

### LINKED ISSUE #4146: Provide an IntersectionValidationHandler
Our intersection handlers (and derived handlers such as the Sliproad handler) run in the `osrm-extract` tool for each internal intersection type. We should create a new (optional) handler running last which reports issues with the intersections and checks their integrity.

Issues we should detect and report (feel free to add more):
- Sharp turns onto ramps
- Uturns (or two quick-turns) with a high speed
- ..

Here's the intersection handler:

https://github.com/Project-OSRM/osrm-backend/blob/9d308172949416701f375535c4a94c9291c4f283/include/extractor/guidance/intersection_handler.hpp#L29-L117

cc @joto @danpat

daniel-j-h:
For #4146.

First check for detecting sharp turns onto ramps.

On Berlin it found some issues e.g. a missing turn restriction [here](https://www.openstreetmap.org/edit#map=19/52.47756/13.30795):

https://www.mapillary.com/app/?lat=52.47755740000002&lng=13.307866699972237&z=17&focus=photo&pKey=peJilZTC-GvMNaUwVtifkQ&x=0.48982839346581003&y=0.5297959150533438&zoom=0

map0

Tasks:
- [ ] Make pass optional via command line flag
- [ ] Make output machine-readable
- [ ] Run on planet (or larger dataset) and sample some locations, check for accuracy

cc @joto


danpat:
@daniel-j-h we should make "possibly bad OSM data" a formal output from `osrm-extract`/`osrm-contract/partition/customize`, and publish it for each demo server data build.

I'm thinking something like:

```
osrm-extract -p car.lua --validation=problems.geojson australia.osm.pbf
```

I'm suggesting GeoJSON format because it's then easy to render complex features on a problem map.  In the case above, you could plot not just the intersection location, but the segments before and after the turn angle, making it easy to identify on a visual map.

include/extractor/guidance/validation_handler.hpp
```diff
@@ -0,0 +1,86 @@
...
+  private:
+    void checkForSharpTurnsOntoRamps(const NodeID,
+                                     const EdgeID via_eid,
+                                     const Intersection &intersection) const
+    {
+        for (std::size_t i = 1; i < intersection.size(); ++i)
+        {
+            const auto road = intersection[i];
+
+            if (!(road.instruction.type == TurnType::OnRamp) || !road.entry_allowed)
```
MoKob: @daniel-j-h what is your opinion on interaction with normal on-ramps here? I would have thought that many on-ramps / off ramps exist that are not necessarily slight turns.
Does the output produce a lot of information here? I would have restricted the check for turns between ramps (which is not possible to check based on the turn type).

What do you think, should we give a check comparing ramp-classes for sharp turns a try?

emiltin:
how does this work for other modes that cars? i suppose these type of checks would be different for e.g. car, bicycle and foot?

include/extractor/guidance/validation_handler.hpp
daniel-j-h: For Berlin I get tens of reports so it's certainly not too bad.

I don't quite understand what you mean by

> check comparing ramp-classes for sharp turns a try

daniel-j-h:
@emiltin I haven't checked other profiles than the car profile. The main idea here is to find e.g. missing turn restrictions and invalid data modelling in osm for driving.

For the walk and bike profile some of the validation checks have to be adapted for sure.

include/extractor/guidance/validation_handler.hpp
daniel-j-h: Capturing from chat: we could refine the check / add a new one to check if we have a sharp turn and the `RoadClassification` type is-ramp for turning from a ramp-type street onto a ramp-type street.

daniel-j-h:
Here's the output for the planet - roughly 30k intersections were flagged:

https://gist.github.com/daniel-j-h/b1d0299b72b1d0784a20e28144e64557

---

Looks like this simple validation pass already finds some major issues in the road network, e.g. a turn restriction missing here https://www.openstreetmap.org/edit?editor=id#map=19/38.41233/-81.79695

daniel-j-h:
[map2](https://api.mapbox.com/styles/v1/danieljh/cj42l3yw05epa2sqvellka881.html?access_token=pk.eyJ1IjoiZGFuaWVsamgiLCJhIjoiTnNYb25JSSJ9.vYOnsuu1zeKcGW2nj0uJZw#5.97/51.108/10.748)

the redder the sharper the turn angle onto the ramp

_* click for interactive map_

daniel-j-h:
Here are some examples from sampling - the first situation _could_ actually be correct as there seems to be a stop sign at the intersection coming from east (not modeled in osm).

This opens up the question if announcing valid (!) sharp turns onto ramps as "take the ramp" makes sense or if these should really be "sharp turn right".

---

- [Router](http://map.project-osrm.org/?z=18&center=38.871805%2C-77.160869&loc=38.871590%2C-77.164664&loc=38.872133%2C-77.160394&hl=en&alt=0)
- [Network](https://www.openstreetmap.org/edit#map=19/38.87242/-77.15894)

map3


---

- [Router](http://map.project-osrm.org/?z=18&center=38.787131%2C-76.999126&loc=38.786572%2C-76.998163&loc=38.787484%2C-76.998442&hl=en&alt=0)
- [Network](http://www.openstreetmap.org/edit#map=18/38.78718/-76.99903)

map4

---

I see these sharp turn onto ramp cases especially bad when the user fails to take a ramp and re-routing kicks in. Then situations as shows in this ticket are more likely to occur resulting in major fsck-ups.

include/extractor/guidance/validation_handler.hpp
```diff
@@ -0,0 +1,84 @@
...
+        for (std::size_t i = 1; i < intersection.size(); ++i)
+        {
+            const auto road = intersection[i];
+
+            if (road.instruction.type != TurnType::OnRamp || !road.entry_allowed)
+            {
+                continue;
+            }
+
+            if (road.angle <= 2 * NARROW_TURN_ANGLE)
```
MoKob: This should probably also check for road.angle >= 360-2*NARROW_TURN_ANGLE

daniel-j-h: note: I should switch to `angularDeviation`

daniel-j-h:
I just ran this branch on the planet and here are the results:

map0

- < 45 degree #e55e5e (red)
- < 65 degree #f9886c (orange)
- < 85 degree #fbb03b (yellow)

[Click for map](https://api.mapbox.com/styles/v1/danieljh/cj4zptszf0xgl2srvbsseiwjw.html?access_token=pk.eyJ1IjoiZGFuaWVsamgiLCJhIjoiTnNYb25JSSJ9.vYOnsuu1zeKcGW2nj0uJZw#6.18/50.678/8.984)

daniel-j-h:
We should not only log the intersection node but instead log the two ways where a sharp turn is possible as line strings; I think we already have a GeoJSON printer which should be able to do so. cc @srividyacb

daniel-j-h:
To reduce the number of false positives we probably should run this handler with a profile that discards all service roads. As an example see this location where a sharp turn from a driveway service road gets detected:

https://api.mapbox.com/styles/v1/danieljh/cj4zptszf0xgl2srvbsseiwjw.html?access_token=pk.eyJ1IjoiZGFuaWVsamgiLCJhIjoiTnNYb25JSSJ9.vYOnsuu1zeKcGW2nj0uJZw#19.59/37.38036/-121.99635

http://map.project-osrm.org/debug/#19.38/37.38026/-121.99622

daniel-j-h:
- [ ] We classify ramps for highway tags only up to trunk. It looks like we should expand this check and look out e.g. for a turn from a primary onto a primary link.

Here's one example where a sharp turn from a primary onto a primary link is possible:

http://www.openstreetmap.org/node/30391780

http://map.project-osrm.org/?z=18&center=52.294286%2C9.820879&loc=52.295357%2C9.819825&loc=52.292582%2C9.822818&hl=en&alt=0

https://www.mapillary.com/app/?lat=52.294130555555554&lng=9.819397222222221&z=17&pKey=C_EzKJZ2Eq-PRN08b9xdog&focus=photo

daniel-j-h:
Added some logic to dump the osm ids for the from, via, to node based graph nodes.

Added a check for sharp turns from fast roads (think: primary, secondary, tertiary). Disabled by default for now: sampling some locations in Berlin showed false positives where there are e.g. traffic signals.

- [ ] Check if restricting the check to only primary roads and to only turns from primary onto primary helps

Currently deploying this branch on a staging demo server stack; we should have results by tomorrow.

daniel-j-h:
Up-to-date results on the planet are [here](https://api.mapbox.com/styles/v1/danieljh/cj69ndyvf2iyu2rmuckadahxs.html?access_token=pk.eyJ1IjoiZGFuaWVsamgiLCJhIjoiTnNYb25JSSJ9.vYOnsuu1zeKcGW2nj0uJZw#17.34/37.78784/-122.39325):

[map1](https://api.mapbox.com/styles/v1/danieljh/cj69ndyvf2iyu2rmuckadahxs.html?access_token=pk.eyJ1IjoiZGFuaWVsamgiLCJhIjoiTnNYb25JSSJ9.vYOnsuu1zeKcGW2nj0uJZw#17.34/37.78784/-122.39325)

daniel-j-h:
Referencing OpenStreetMap diary this is discussed in: [english](https://www.openstreetmap.org/user/daniel-j-h/diary/41777), [german](https://www.openstreetmap.org/user/daniel-j-h/diary/41776)

willwhite:
Added this visualization to https://www.mapbox.com/get-directions/ which should help test the restrictions in the context of the detection.

I'm seeing a few false positives like [this one](https://www.mapbox.com/get-directions/#20.07/37.64674/-122.40551?coordinates=-122.405594137673,37.646798869078836;-122.40551299815073,37.646744235050704) but overall this is finding a lot of good stuff like https://www.mapbox.com/get-directions/#17.76/37.63073/-122.43371?coordinates=-122.433855531552,37.630766120768385;-122.43371315498953,37.63072602946407.

Thoughts on adding dual-carriageway transitions, [like this one](https://hey.mapbox.com/directions/#18.39/37.72306/-122.45251?coordinates=-122.45258170030795,37.72317172845237;-122.45251227848881,37.72306352140487), to the criteria we evaluate?

What's left before this merges?

include/extractor/guidance/validation_handler.hpp
```diff
@@ -177,6 +179,81 @@ class ValidationHandler final : public IntersectionHandler
...
+        // Detect situations where dual-carriageway fans-in from two ways into one and
+        // the Uturn from a dual-carriageway onto the dual-carriageway should be possible.
+        //
+        // . .
+        //     > .
+        // . .
+        //
+
+        // Todo: are there dual-carriageway intersections of larger size?
+        if (intersection.size() != 3)
```
MoKob: This prevents checks for any intersections. It only finds turns around traffic islands. Most of these will split next to intersections -> intersection.size() > 3

include/extractor/guidance/validation_handler.hpp
```diff
@@ -177,6 +179,81 @@ class ValidationHandler final : public IntersectionHandler
...
+            return;
+        }
+
+        if (!via_road.entry_allowed || !leftmost_road.entry_allowed ||
+            !rightmost_road.entry_allowed)
+        {
+            return;
+        }
+
+        const auto left_turn_angle = osrm::util::angularDeviation(0, leftmost_road.angle);
```
MoKob: angle is a bit confusing here. In addition: 2* NARROW could be a bit much (80 degrees).
OSRM will already try to correct for merging roads/offsets. So you should get away here with a far lower angle deviation.

include/extractor/guidance/validation_handler.hpp
```diff
@@ -177,6 +179,81 @@ class ValidationHandler final : public IntersectionHandler
...
+
+        const auto &via_road = intersection.getUTurnRoad();
+        const auto &leftmost_road = intersection.getLeftmostRoad();
+        const auto &rightmost_road = intersection.getRightmostRoad();
+
+        const auto &via_road_data = node_based_graph.GetEdgeData(via_eid);
+        const auto &leftmost_road_data = node_based_graph.GetEdgeData(leftmost_road.eid);
+        const auto &rightmost_road_data = node_based_graph.GetEdgeData(rightmost_road.eid);
+
+        if (via_road_data.name_id == EMPTY_NAMEID || leftmost_road_data.name_id == EMPTY_NAMEID ||
```
MoKob: this Entire part feels like code-duplication. Wouldn't this offer itself to check for `isSharpTurnOntoDual` and simply compare via/left, via/right for themselves instead of carrying all the left/right things?

include/extractor/guidance/validation_handler.hpp
```diff
@@ -177,6 +179,81 @@ class ValidationHandler final : public IntersectionHandler
...
+                                                   rightmost_road_data.name_id,
+                                                   name_table,
+                                                   street_name_suffix_table);
+
+        if (!same_name_left || !same_name_right)
+        {
+            return;
+        }
+
+        if (!via_road.entry_allowed || !leftmost_road.entry_allowed ||
```
MoKob: a check as simple as this should be first to ensure we don not check costly names if we do not accept it anyhow.
Also: you are coming from via, so you are obviously on a turn that is allowed.
What you would care for is that you cannot turn back onto via (check if oneway -> intersection[0]->reversed == true).

You should also check if `left`/`right` are a oneway.

emiltin:
A danish OSM user found a number of places where oneways going to/from roundabouts where reversed: https://agol.dk/osm/fejlgal/#0.

It seems like it should be possible tofind these errors with the osrm validation handlers, by looking at the angle of the turn between the enter/exit way and the roundabout. For valid sitautions the angle is usually (always?) below 90 degrees.  If the oneway direction of the entry/exit way is reversed, it would be above 90 degree.

Looking manually at 10-20 roundabouts in Denmark I could not find any examples where the angle is above 90 degrees.

But as far as I understand it, the validation handler currently only looks at ramps?

daniel-j-h:
By now it's no longer ramps only, it detects sharp roundabout and dual-carriageway turns, too, for example.

emiltin:
Ok I see. Is there an updated map showing the porblems identified by these validations?

daniel-j-h:
If you want to see up-to-date results the best way is to check out this branch and enable / disable checks here (you want the `checkForSharpTurnsInRoundabouts` check):

https://github.com/Project-OSRM/osrm-backend/blob/7ae0270e52412e8a23364d69e14d975f234920de/include/extractor/guidance/validation_handler.hpp#L95-L98

emiltin:
I don't see that check checkForSharpTurnsInRoundabouts in the list, so I would add it?
Is the processing flow documented somewhere? It outputs to a file or as vector tiles?

daniel-j-h:
Yes add it to the list, it's [here](https://github.com/Project-OSRM/osrm-backend/blob/7ae0270e52412e8a23364d69e14d975f234920de/include/extractor/guidance/validation_handler.hpp#L371).

Not really documented it's really more experimental; that's why the branch isn't in master. Output will be lines with osm ids, locations and angles on stdout.

daniel-j-h:
[Here](https://api.mapbox.com/styles/v1/danieljh/cj7vtp9jx29bg2st3h3dxa5st.html?access_token=pk.eyJ1IjoiZGFuaWVsamgiLCJhIjoiTnNYb25JSSJ9.vYOnsuu1zeKcGW2nj0uJZw#8.7/37.7688/-122.0152) is the most recent map I have; does not yet include the roundabout turns, though.

TheMarex:
This PR is quite outdated and I think the last state was that we will not add linters directly to OSRM.

PR #4161, opened 2017-06-15
don't collapse u-turns into combined turns

MoKob:
# Issue

Prevents issues that arise from combining u-turns into other turns. U-Turns are always a dedicated maneuver and shouldn't be collapsed.

## Tasklist

- [x] add regression / cucumber cases (see docs/testing.md) (current test does not fail yet)
- [x] review
- [x] adjust for comments

### DIFF
--- a/src/engine/guidance/collapse_scenario_detection.cpp
+++ b/src/engine/guidance/collapse_scenario_detection.cpp
@@ -61,6 +61,22 @@ bool isShortAndUndisturbed(const RouteStep &step)
     return is_short && noIntermediaryIntersections(step);
 }
 
+// On dual carriageways, we might want to use u-turns in combination with new-name instructions.
+// Otherwise a u-turn should never be part of a collapsing instructions.
+bool noBadUTurnCombination(const RouteStepIterator first, const RouteStepIterator second)
+{
+    auto has_uturn = hasModifier(*first, DirectionModifier::UTurn) ||
+                     hasModifier(*second, DirectionModifier::UTurn);
+
+    auto const from_name_change_into_uturn =
+        hasTurnType(*first, TurnType::NewName) && hasModifier(*second, DirectionModifier::UTurn);
+
+    auto const uturn_into_name_change =
+        hasTurnType(*second, TurnType::NewName) && hasModifier(*first, DirectionModifier::UTurn);
+
+    return !has_uturn || from_name_change_into_uturn || uturn_into_name_change;
+}
+
 } // namespace
 
 bool basicCollapsePreconditions(const RouteStepIterator first, const RouteStepIterator second)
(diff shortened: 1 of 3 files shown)


src/engine/guidance/collapse_scenario_detection.cpp
```diff
@@ -81,8 +95,11 @@ bool basicCollapsePreconditions(const RouteStepIterator first,
                                      hasRoundaboutType(second->maneuver.instruction) ||
                                      hasRoundaboutType(third->maneuver.instruction);
 
+    const auto contains_bad_uturn =
+        !isValidUturnCombination(first, second) && !isValidUturnCombination(second, third);
+
     // require modes to match up
-    return !has_roundabout_type && haveSameMode(*first, *second, *third);
+    return !has_roundabout_type && haveSameMode(*first, *second, *third) && !contains_bad_uturn;
```
daniel-j-h: General note here: should `basicCollapsePreconditions(first, second, third)` run `basicCollapsePreconditions(first, second)` on its `first, second` and `second, third` argument here? If not, why?

src/engine/guidance/collapse_scenario_detection.cpp
```diff
@@ -61,6 +61,17 @@ bool isShortAndUndisturbed(const RouteStep &step)
...
 
+bool isValidUturnCombination(const RouteStepIterator first, const RouteStepIterator second)
+{
+    auto has_uturn = hasModifier(*first, DirectionModifier::UTurn) ||
+                     hasModifier(*second, DirectionModifier::UTurn);
+    auto const from_name_change_into_uturn =
+        hasTurnType(*first, TurnType::NewName) && hasModifier(*second, DirectionModifier::UTurn);
+    auto const uturn_into_name_change =
+        hasTurnType(*second, TurnType::NewName) && hasModifier(*first, DirectionModifier::UTurn);
+    return !has_uturn || from_name_change_into_uturn || uturn_into_name_change;
```
daniel-j-h: What do we actually check here? What is the semantic of a "valid uturn combination"? Why is not having uturn modifiers a valid uturn combination for example? And why does it have to include name changes?

My impression was we only want to check if two steps are uturns and then _not_ collapse them. Care to explain why this is not the case?

MoKob: Forbidding uturns in general does end up suppressing a lot of u-turns at intersections that are dual carriage ways in both directions. We need to allow name-changes right after / before to handle these correctly.

So we could think about changing the name here. What we want to be collapsable is the last three statements: Uturns are only part of a collapse if a name-change is involved. Otherwise they cannot be part of a multi-hop instruction.

src/engine/guidance/collapse_scenario_detection.cpp
MoKob: Technically we do this. The combination of both is what we do here. The only difference is the involvement of waypoint types, which can happen for three-hop scenarios (without changing the waypoint itself)

src/engine/guidance/collapse_scenario_detection.cpp
daniel-j-h: Hm yeah either change the name or at least add a comment, this was not clear to me at all.

Issue #4018, opened 2017-05-05
Debug tiles are missing rate attribute on edges


danpat:
Now that we support #77 , we need to be able to visualize values for route debugging.

The debug tiles expose the following attributes for road segments:

  - speed
  - is_small
  - datasource
  - weight
  - duration
  - name

Missing from this list is `rate`, which is `length / weight`.  We have `weight`, but we have no normalize value (like `speed) that makes it easy to see where penalized weights are in play.

The task here is to add the `rate` attribute - it should be a simple duplication of how the `speed` is calculated and exposed.

daniel-j-h:
Related: https://github.com/Project-OSRM/osrm-backend/issues/3793

daniel-j-h:
@danpat https://github.com/Project-OSRM/osrm-backend/issues/4162 added rate to the debug tiles; anything left here that you want to add? Can you update your comment or close here please.

danpat:
Ah, thanks for the reminder.  This is done.

PR #4162, opened 2017-06-15
Include 'rate' property (reciprocal of weight) on debug tile edges.

danpat:
# Issue

This PR is to address https://github.com/Project-OSRM/osrm-backend/issues/4018 where we're not exposing the reciprocal of the `weight` property (analogous to `speed` if doing time-based routing).

This PR exposes a new property called `rate` for each edge.  Similar to `speed`, it's encoded to 1 decimal place of precision, and the value is in `meters per unit(weight)`, where `weight` is the internal unit-less cost of edge traversal.

**ALSO**, this PR exposes the turn weight property for the turn layer.  Given that actual routing occurs using weights, it'd be helpful to expose the `weight` value for turns in addition to the `cost` (duration) (which is only used for ETA calculation).

## Tasklist
 - [ ] review
 - [ ] adjust for comments

### DIFF
--- a/docs/http.md
+++ b/docs/http.md
@@ -419,8 +419,10 @@ Vector tiles contain two layers:
 | `speed`      | `integer` | the speed on that road segment, in km/h  |
 | `is_small`   | `boolean` | whether this segment belongs to a small (< 1000 node) [strongly connected component](https://en.wikipedia.org/wiki/Strongly_connected_component) |
 | `datasource` | `string`  | the source for the speed value (normally `lua profile` unless you're using the [traffic update feature](https://github.com/Project-OSRM/osrm-backend/wiki/Traffic), in which case it contains the stem of the filename that supplied the speed value for this segment |
-| `duration`   | `float`   | how long this segment takes to traverse, in seconds |
+| `duration`   | `float`   | how long this segment takes to traverse, in seconds.  This value is to calculate the total route ETA. |
+| `weight  `   | `integer` | how long this segment takes to traverse, in units (may differ from `duration` when artificial biasing is applied in the Lua profiles).  ACTUAL ROUTING USES THIS VALUE. |
 | `name`       | `string`  | the name of the road this segment belongs to |
+| `rate`       | `float`   | the value of `length/weight` - analagous to `speed`, but using the `weight` value rather than `duration`, rounded to the nearest integer |
 
 `turns` layer:
 
@@ -429,6 +431,7 @@ Vector tiles contain two layers:
 | `bearing_in` | `integer` | the absolute bearing that approaches the intersection.  -180 to +180, 0 = North, 90 = East |
 | `turn_angle` | `integer` | the angle of the turn, relative to the `bearing_in`.  -180 to +180, 0 = straight ahead, 90 = 90-degrees to the right |
 | `cost`       | `float`   | the time we think it takes to make that turn, in seconds.  May be negative, depending on how the data model is constructed (some turns get a "bonus"). |
+| `weight`     | `float`   | the weight we think it takes to make that turn.  May be negative, depending on how the data model is constructed (some turns get a "bonus"). ACTUAL ROUTING USES THIS VALUE |
 
 
 ## Result objects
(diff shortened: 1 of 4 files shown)


src/engine/plugins/tile.cpp
```diff
@@ -428,6 +428,12 @@ void encodeVectorTile(const datafacade::ContiguousInternalMemoryDataFacadeBase &
                 const auto &edge = edges[edge_index];
                 const auto geometry_id = get_geometry_id(edge);
 
+                // Get coordinates for start/end nodes of segment (NodeIDs u and v)
+                const auto a = facade.GetCoordinateOfNode(edge.u);
+                const auto b = facade.GetCoordinateOfNode(edge.v);
+                // Calculate the length in meters
+                const double length = osrm::util::coordinate_calculation::haversineDistance(a, b);
```
MoKob: when trying to refactor this I thought about this line a bit. It adds some overhead for calculating the distances. Whats is you opinion here @danpat on the potential slowdown? I feel it might be alright, but I am not entirely sure. (This length calculation is already happening further down as well)

src/engine/plugins/tile.cpp
```diff
@@ -439,6 +445,14 @@ void encodeVectorTile(const datafacade::ContiguousInternalMemoryDataFacadeBase &
                 use_line_value(forward_weight);
                 use_line_value(reverse_weight);
 
+                std::uint32_t forward_rate =
+                    static_cast<std::uint32_t>(round(length / forward_weight * 10.));
+                std::uint32_t reverse_rate =
+                    static_cast<std::uint32_t>(round(length / reverse_weight * 10.));
+
+                use_line_value(forward_rate);
```
MoKob: I like storing it as int as well here. My first intuition was to store it as double.

src/engine/plugins/tile.cpp
```diff
@@ -518,6 +532,7 @@ void encodeVectorTile(const datafacade::ContiguousInternalMemoryDataFacadeBase &
                                                    &max_datasource_id,
                                                    &used_line_ints](const FixedLine &tile_line,
                                                                     const std::uint32_t speed_kmh,
+                                                                    const std::uint32_t rate,
```
MoKob: how do you feel about making clearer that all of these are offsets? The first time I though these were supposed to be actual values (like the speed is, since we store all 127 possible values).

src/engine/plugins/tile.cpp
danpat: I'm not super-concerned about performance for things like this - unless there's a really easy way to refactor it, I'm inclined to just eat the cost.  If https://github.com/mapbox/vector-tile ever grows encoding ability, we could do a big refactor and clean up stuff like that.  Until then, I'm totally OK with there being some technical debt here.

docs/http.md
```diff
@@ -429,6 +431,7 @@ Vector tiles contain two layers:
 | `bearing_in` | `integer` | the absolute bearing that approaches the intersection.  -180 to +180, 0 = North, 90 = East |
 | `turn_angle` | `integer` | the angle of the turn, relative to the `bearing_in`.  -180 to +180, 0 = straight ahead, 90 = 90-degrees to the right |
 | `cost`       | `float`   | the time we think it takes to make that turn, in seconds.  May be negative, depending on how the data model is constructed (some turns get a "bonus"). |
+| `weight`     | `float`   | the weight we think it takes to make that turn.  May be negative, depending on how the data model is constructed (some turns get a "bonus"). ACTUAL ROUTING USES THIS VALUE |
```
MoKob: Just a guess: should these also contain `rate` or don't we need it here?

src/engine/plugins/tile.cpp
danpat: Yeah - Mapbox GL does some ugly formatting on floats https://github.com/mapbox/mapbox-gl-js/pull/3358, so some of the reasoning for using ints vs doubles in this tile encoder is driven by rendering ugliness :-(

docs/http.md
danpat: There's no rate on point features, the turn penalty values are simply additive numbers.  The `rate` only applies on edges so that we have an easy way to visually compare (e.g. via colour) edges of differing lengths.

src/engine/plugins/tile.cpp
danpat: Yeah, that's a vector tile encoding gotcha - I agree, it's a bit unintuitive if you haven't read the spec.  I'll rename these, now is as good a time as any.

Issue #4163, opened 2017-06-15
Matching plugin splits tracks with the gaps=ignore option


gardster:
The match plugin splits good and physically possible track if there is a huge gap in coordinates inside this track.
![error](https://user-images.githubusercontent.com/1862559/27185528-4fb78e00-51ee-11e7-8c8e-50cba12907b8.png)

ankitsoni9989:
Hi,
I have seen a few times that match service does return multiple matchings even when `gaps=ignore` is supplied as input, what could be the possible reason/s for that? I have attached an example:

[...]

PR #4164, opened 2017-06-15
Time information for maximum distance in matching

gardster:
# Issue
#4163 
Allowed maximum distance were calculated only by median time sample of the whole track. So distance can be impossible if the source track has gaps in its sequence. I suggest to calculate this distance for each coordinate pair separately.

### DIFF
--- a/src/engine/routing_algorithms/map_matching.cpp
+++ b/src/engine/routing_algorithms/map_matching.cpp
@@ -145,7 +145,8 @@ SubMatchingList mapMatching(SearchEngineData<Algorithm> &engine_working_data,
     for (auto t = initial_timestamp + 1; t < candidates_list.size(); ++t)
     {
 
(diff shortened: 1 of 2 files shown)


src/engine/routing_algorithms/map_matching.cpp
```diff
@@ -155,13 +145,25 @@ SubMatchingList mapMatching(SearchEngineData<Algorithm> &engine_working_data,
     for (auto t = initial_timestamp + 1; t < candidates_list.size(); ++t)
     {
 
+        const auto step_time =
```
TheMarex: This will cause a SIGSEV if `trace_timestamps` is empty (`use_timestamps == false`)

gardster: Oh. Sorry for this error. Fixed.

Issue #3899, opened 2017-04-04
Unify shared memory loading code and de-serialization


TheMarex:
#3892 already lay the groundwork for this but I wan to spec out the plan in more detail here:

We now have an interface for `serialization`/`deserialization` that looks something like this:

```
serialization::read(FileReader &reader, SomeDataStructue &data_structure);
serialization::write(FileWriter &writer, SomeDataStructue &data_structure);
```

We want to hide the fact that `SomeDataStructure` might not own it's backing memory (e.g. by using `util::vector_view` instead of `std::vector`) from the de-serialization code.

This is straight forward to do as can be demonstrated here for some classes where I already ported the new model:
- `TurnDataContainer` https://github.com/Project-OSRM/osrm-backend/blob/dce7d5672b5d88d3d4c738ced8ef32da47769885/src/storage/storage.cpp#L574-L625
- `SegmentDataContainer` https://github.com/Project-OSRM/osrm-backend/blob/dce7d5672b5d88d3d4c738ced8ef32da47769885/src/storage/storage.cpp#L628-L679
- `MultiLevelPartition` https://github.com/Project-OSRM/osrm-backend/blob/dce7d5672b5d88d3d4c738ced8ef32da47769885/src/storage/storage.cpp#L877-L899

This code is starting to look very similar to what we do in `ContingiousInternalMemDataFacade`:

https://github.com/Project-OSRM/osrm-backend/blob/dce7d5672b5d88d3d4c738ced8ef32da47769885/include/engine/datafacade/contiguous_internalmem_datafacade.hpp#L421-L473

In fact it is the same code save for the additional call to `extractor::files::readSegmentData`. So there will be a nice opportunity of a refactor:

```
SegmentDataView makeSegmentDataView(const DataLayout &)
{
         // .... code from data facade goes here ....
}
```

To fully implement this we need to hit a few milestones:

1. All file reading/writing must be fully ported to the `include/{component}/files.hpp` and `include/{component}/serialization.hpp` schema.
2. The data facades need to be refactored to put all object initialization into factory functions like above
3. `storage.cpp` and the datafacades need to be refactored to use said factory functions

/cc @daniel-j-h @duizendnegen @danpat

duizendnegen:
:+1:

Do the `component`s that you talk about map directly to files? Will serialization & deserialization logic be contained within the same factory? I see some wins there as well with regards to understandability.

[...]

PR #4166, opened 2017-06-15
Refactor file writing in OSRM contract

TheMarex:
# Issue

This works towards #3899 by refactoring the file code in the contractor.

## Tasklist
 - [x] review
 - [x] adjust for comments

### DIFF
--- a/include/contractor/graph_contractor.hpp
+++ b/include/contractor/graph_contractor.hpp
@@ -91,8 +91,8 @@ class GraphContractor
 
     GraphContractor(int nodes,
                     std::vector<ContractorEdge> edges,
-                    std::vector<float> &&node_levels_,
-                    std::vector<EdgeWeight> &&node_weights_);
+                    std::vector<float> node_levels_,
+                    std::vector<EdgeWeight> node_weights_);
 
     /* Flush all data from the contraction to disc and reorder stuff for better locality */
     void FlushDataAndRebuildContractorGraph(ThreadDataContainer &thread_data_list,
@@ -101,12 +101,14 @@ class GraphContractor
 
     void Run(double core_factor = 1.0);
 
-    void GetCoreMarker(std::vector<bool> &out_is_core_node);
+    std::vector<bool> GetCoreMarker();
 
-    void GetNodeLevels(std::vector<float> &out_node_levels);
+    std::vector<float> GetNodeLevels();
 
-    template <class Edge> inline void GetEdges(util::DeallocatingVector<Edge> &edges)
+    template <class Edge> inline util::DeallocatingVector<Edge> GetEdges()
     {
+        util::DeallocatingVector<Edge> edges;
+
         util::UnbufferedLog log;
         log << "Getting edges of minimized graph ";
         util::Percent p(log, contractor_graph->GetNumberOfNodes());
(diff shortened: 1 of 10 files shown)


TheMarex:
Urg some tetss are failing need to investigate.

include/contractor/graph_contractor.hpp
```diff
@@ -101,12 +101,14 @@ class GraphContractor
 
     void Run(double core_factor = 1.0);
 
-    void GetCoreMarker(std::vector<bool> &out_is_core_node);
+    std::vector<bool> GetCoreMarker();
```
MoKob: Thanks for getting rid of the `please fill my vector` way. I feel this way is much better. How do you feel about adding some extra commentary here? As far as I read the code, both vectors get destroyed when accessing here, right?

include/util/deallocating_vector.hpp
```diff
@@ -258,8 +258,12 @@ template <typename ElementT> class DeallocatingVector
     DeallocatingVector(DeallocatingVector &other) = delete;
     DeallocatingVector &operator=(DeallocatingVector &other) = delete;
     // moving is fine
-    DeallocatingVector(DeallocatingVector &&other) = default;
-    DeallocatingVector &operator=(DeallocatingVector &&other) = default;
+    DeallocatingVector(DeallocatingVector &&other) { swap(other); }
+    DeallocatingVector &operator=(DeallocatingVector &&other)
```
MoKob: I know where this came from, still I am a bit confused on why this change was necessary. Was it just unexplainable behaviour, or did you find some reason for the problem?

include/util/vector_view.hpp
```diff
@@ -143,8 +143,33 @@ template <> class vector_view<bool>
     unsigned *m_ptr;
     std::size_t m_size;
 
+    static constexpr std::size_t UNSIGNED_BITS = CHAR_BIT * sizeof(unsigned);
```
MoKob: 👍

include/util/vector_view.hpp
```diff
@@ -143,8 +143,33 @@ template <> class vector_view<bool>
...
+
   public:
     using value_type = bool;
+    struct reference
```
MoKob: Nitpick/Questions: I am not entirely sure on the substructures naming convention, so feel free to ignore: shouldn't this be capital case?

include/util/vector_view.hpp
```diff
@@ -166,6 +191,13 @@ template <> class vector_view<bool>
 
     bool operator[](const unsigned index) const { return at(index); }
 
+    reference operator[](const unsigned index)
+    {
+        const std::size_t bucket = index / UNSIGNED_BITS;
```
MoKob: Opposed to `at`, this one is not asserting the index anymore. Could you add an assert like in `at`?

include/util/deallocating_vector.hpp
TheMarex: It basically caused a memory leak because `buckets = std::move(other.buckets)` (the auto generated code) would not free the buckets.

include/util/vector_view.hpp
TheMarex: This needs to be lower-case to be compatible with the vector naming of this type.

TheMarex:
Addressed the PR comments, just waiting on 💚 before merge.

Issue #2772, opened 2016-08-16
Make Profiles Country Aware


MoKob:
We have many issues that relate to specific defaults in different countries.
E.g. highways to allow pedestrians (India), `trunk`roads allowing bicycles (https://github.com/Project-OSRM/osrm-backend/issues/639), or the side of road we are driving on (https://github.com/Project-OSRM/osrm-backend/issues/2269).

Also default speed-values differ on a by country basis.
So allow a country-based set-up, we would need to pass country information along with every way and choose from a set of different default values.

Main issues to adjust in my view would be:
 - [ ] access whitelist ([e.g.](https://github.com/Project-OSRM/osrm-backend/blob/087080a6c3204669275280a68005e538e08ee4d9/profiles/car.lua#L9))
 - [ ] speed values ([e.g.](https://github.com/Project-OSRM/osrm-backend/blob/087080a6c3204669275280a68005e538e08ee4d9/profiles/car.lua#L19-L39)
 - [ ] table which side on the road is driven on based on country (adjust stuff like: [this](https://github.com/Project-OSRM/osrm-backend/blob/087080a6c3204669275280a68005e538e08ee4d9/profiles/car.lua#L148)
 - [ ] adjust [turn penalty](https://github.com/Project-OSRM/osrm-backend/blob/087080a6c3204669275280a68005e538e08ee4d9/profiles/car.lua#L525-L533) based on the road side

MoKob:
https://github.com/Project-OSRM/osrm-backend/issues/1735 is another issue related to this. U-turns allowed in general, but not reasonable based on speeds.

MoKob:
Access to certain tracks might also be based on countries? https://github.com/Project-OSRM/osrm-backend/issues/576

TheMarex:
For a discussion about a possible profile level implementation:

https://github.com/Project-OSRM/osrm-backend/issues/333

andrewharvey:
Would this include uturn restrictions. eg. in NSW uturns cannot be made a traffic lights unless a sign says you can [1]. Where no turn restriction exists in OSM should OSRM follow the region default restricting uturns?

[1] http://www.rms.nsw.gov.au/roads/safety-rules/road-rules/intersections.html

daniel-j-h:
Yes @andrewharvey such a behavior would be desirable: but there's currently no code for parsing such signs for enabling uturns. To pull this off I think you need to

[...]

PR #4167, opened 2017-06-15
Country-aware way function

danpat:
This is an attempt to address https://github.com/Project-OSRM/osrm-backend/issues/3180 and https://github.com/Project-OSRM/osrm-backend/issues/2772 where we want to know what country that a `Way` belongs to when we're processing its tags.

This PR implements two changes:

  1. The `:lat()` and `:lon()` functions will work on nodes inside the `way_function` in Lua
  2. A GeoJSON file(s) with `Polygon` features can be used to perform lookups prior to the `way_function` being called, passing in additional properties related to which polygons a way is inside.

### :lat() and :lon() function support

If the Lua profile property:

```lua
properties.enable_way_coordinates          = true
```

is set, then `Node` objects accessible via `way:get_nodes()` will have coordinate information populated properly.  The cost of using this is about 20-25GB of RAM for the node location cache. 

The `osmium::NodeRef` object has variables ready to hold lon/lat data, they just need to be set.  The `osmium::handler::NodeLocationsForWays` handler does just that - it pulls previously seen nodes out of a cache and copies the lon/lat values into the empty fields on the `osmium::Way` object.

The following will now work inside the `way_function`:

```lua
  local nodes = way:get_nodes()
  for index,node in ipairs(nodes) do
      print(node:lon(), node:lat())
  end
```

### GeoJSON polygon lookup

Assuming node locations are enabled for ways, you can then, given a GeoJSON file like this:

```json
{
  "type": "FeatureCollection",
  "features": [
    { "type": "Feature",
      "properties": {
        "country": "null island",
        "timezone": "UTC+99"
      },
      "geometry": {
        "type": "Polygon",
        "coordinates": [
          [
          [ -2, 2 ],
          [ 2, 2 ],
          [ 2, -2 ],
          [ -2, -2 ],
          [ -2, 2 ]
          ]
        ]
      }
    }
  ]
}
```

The Lua script can now support an additional lookup:

```lua
properties.regions = {
  { file = "example.geojson", properties = { "country", "timezone" } }
}
```

If the first coordinate of a way is inside a polygon, then the properties of that polygon are passed to the `way_function` as a table:

```lua
function way_function(way, result, regions)

  if #regions > 0 then
    print("Region data detected on way "..way:id())
    for key,value in ipairs(regions) do
      print(key, value)
    end
  else
    print("No region data detected on way "..way:id())
  end
...
```

Will print something like:

```
Region data detected on way 12378634
country=null island
timezone=UTC+99
```

Notes:
  - there is a point-in-polygon test made for every way - this may slow down extraction significantly if the polygons are very high resolution (lots of edges to test), or there are many overlapping
  - polygons that overlap and match will have properties merged.  Conflicting property names from polygons will be selected in no guaranteed order.  Non-conflicting property names will be merged cleanly.
  - multiple files can be loaded - non-conflicting properties will be merged OK.

In general, simple polygons with multiple properties is probably the best way to use this feature.  While you *could* do something like:

```
properties.regions = {
  { file = "data/world.geo.json/countries.geo.json", property = "name" },
  { file = "timezones.geojson", property = 'TZID' }
}
```

better performance will occur when there are fewer polygons and each polygon has all the attributes you need when a hit occurs.  Avoiding polygon overlaps is probably also desirable.

## Tasklist
 - [ ] review
 - [ ] adjust for comments


joto:
@danpat I did extensive benchmarks on the different implementations that osmium provides for the location index. Some information about this is in the docs: http://osmcode.org/osmium-concepts/#indexes . If you want to compare your implemenation, there are [benchmark scripts available](https://github.com/osmcode/libosmium/blob/master/benchmarks/osmium_benchmark_index_map.cpp).

The short of it is: For the planet or continent-sized extracts, the dense_mmap_array is best with negligable runtime cost (<20s) and low memory overhead. (This btw. also means that this code doesn't have to be run in multiple threads or so.)

Usually libosmium-based programs allow the user to set the type of index they want to use depending on the type of input they have. This is done using a MapFactory. You can see the an example [in the tests](https://github.com/osmcode/libosmium/blob/master/test/t/index/test_id_to_location.cpp#L183-L204).

Using just one of all these implementations for *all* datasets is far from optimal, the overhead of all the implementations for the "wrong" dataset is huge. It has been on my todo list for a long time to provide a index, that will optimize itself depending on the input data, but I never got around to it.

src/extractor/extractor.cpp
```diff
@@ -336,9 +324,37 @@ std::vector<TurnRestriction> Extractor::ParseOSMData(ScriptingEnvironment &scrip
...
+        using osmium_index_type =
+            osrm::util::SparsePPMemMap<osmium::unsigned_object_id_type, osmium::Location>;
+        using osmium_location_handler_type =
+            osmium::handler::NodeLocationsForWays<osmium_index_type>;
+
+        osmium_index_type location_cache;
+        osmium_location_handler_type location_handler(location_cache);
+
+        // TODO: I don't think that the location cache is thread-safe,
+        //       we probably need to synchronize insertion of nodes
```
joto: Just do it in the main thread. The overhead will be much smaller than compared to the synchronization overhead.

joto:
One other thing: Some index types will need a "sorting" step after all node locations are added and before the first way is read. This is needed for instance to sort a vector so that binary lookups can be done. The NodeLocationsForWays handler does this internally whenever it sees the first way. So if your data is not sorted this can fail in subtle ways or be slow.

In any case you need a synchronisation point after you read the last node and before you read the first way for any of this to work, because any way could reference any node and you need to be sure you have the node in the location store already. So you should check that you do get first all nodes and then all ways. Also the indexes are not necessarily thread safe, so it is better to do the whole thing in the main thread. The overhead is so small, it will not matter.

danpat:
@joto Ok, I can add a test.  We talked about relying on the node->way->relation ordering in OSM files, do you think it'd be reasonable to throw an error for an unsorted file and ask the user to sort it before input?

Our current approach doesn't depend on the ordering, but it also doesn't support location information for ways.  It seems like it would be a reasonable requirement to have that nodes must come before ways, and it's an error if it's not that way.

danpat:
Looking at the TBB docs, it looks like the `tbb::filter_t<...>(tbb::filter::serial_in_order,...)` thingy should guarantee that there is only one object being processed at a time.  Only the `tbb::filter::parallel` filters will be executed with multiple threads, so as long as the `NodeLocationsForWays` handler is accessed inside a `tbb::filter::serial_in_order` filter, there should be no thread safety issues.

joto:
@danpat Yes, it is totally reasonable to only support sorted files. Unsorted files are used seldomly, I have seen them only in cases where people converted non-OSM data into OSM formats and didn't know they were supposed to sort those files.

Not sure about the `tbb` stuff, but you should not think in "objects" when splitting up the work load, but in "buffers" to keep any kind of threading overhead down.

danpat:
@emiltin Patrick pointed out that the changes I've made to the profile properties here and the new `way_function` signature might conflict with your work on the v2 profiles.  Any comments on how this PR could best fit in?

emiltin:
hi @danpat, yes the PR #4072 contains a number of improvements to the profile API and bumps the version nr to 2.

amongst other things it moves the properties structure into the profile table, removes the usage of globals and instead passes the profile table as an argument to way/node/etc_functions and moves all profile initialization to a setup() function.

there are some conflicts with this PR. unless there's an urgent need to to merge this PR, i think it would be best to base it on #4072. at least if the plan is to merge #4072 in the foreseeable future.

enable_way_coordinates would be a flag in profile.properties, and regions would live in profile.regions. both would be set in setup(): https://github.com/Project-OSRM/osrm-backend/blob/refactor/profile_properties/profiles/car.lua#L4.

note that the ProfileProperties usertype is not used to bind lua fields to c++ settings anymore. instead the values are simply read from c++. so this part would have to change a bit https://github.com/Project-OSRM/osrm-backend/blob/5e894973fd27c3d2a922a846072e4d9736880f31/src/extractor/scripting_environment_lua.cpp#L258. the usertype is still defined, but it's only to support older api versions.

daniel-j-h:
What's the state of this - where is this PR blocked and what is left to do here?

oxidase:
@danpat https://github.com/Project-OSRM/osrm-backend/pull/4415 just landed in master with slightly different interface `way:get_location_tag('name')` to avoid copying properties from the rtree to Lua if tags are not needed in profiles. The additional `way:get_location_tag` method will appear with API 3 and does not change Lua callback functions.

Wiki page https://github.com/Project-OSRM/osrm-backend/wiki/Using-location-dependent-data-in-profiles

Please reopen or open a new issue if something is missing in #4415

PR #4172, opened 2017-06-19
Add layer to debug tiles to expose all OSM nodes

danpat:
# Issue

In order to help improve debugging, this PR adds a new layer `osmnodes` to the debug vector tiles.  Each node visible in the tile is encoded as a point feature, and the `id` field is set to the OSM node ID value.  There are no additional properties available at this stage.

The intent is to make it quick and easy to link to openstreetmap.org/overpass-turbo.eu and other services that enable analysis of OSM data.

@themarex I know this is late to the game, but it'd be great to include this in the 5.8 release if possible.  It's a purely additive change.

## Tasklist
 - [x] review
 - [x] adjust for comments

### DIFF
--- a/src/engine/plugins/tile.cpp
+++ b/src/engine/plugins/tile.cpp
@@ -831,6 +831,51 @@ void encodeVectorTile(const datafacade::ContiguousInternalMemoryDataFacadeBase &
                 values_writer.add_float(util::vector_tile::VARIANT_TYPE_FLOAT, value);
             }
         }
+
+        // OSM Node tile layer
+        {
+            protozero::pbf_writer point_layer_writer(tile_writer, util::vector_tile::LAYER_TAG);
+            point_layer_writer.add_uint32(util::vector_tile::VERSION_TAG, 2);       // version
+            point_layer_writer.add_string(util::vector_tile::NAME_TAG, "osmnodes"); // name
+            point_layer_writer.add_uint32(util::vector_tile::EXTENT_TAG,
+                                          util::vector_tile::EXTENT); // extent
+
+            std::vector<NodeID> internal_nodes;
+            internal_nodes.reserve(edges.size() * 2);
+            for (const auto &edge : edges)
+            {
+                internal_nodes.push_back(edge.u);
+                internal_nodes.push_back(edge.v);
+            }
+            std::sort(internal_nodes.begin(), internal_nodes.end());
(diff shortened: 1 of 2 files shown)


src/engine/plugins/tile.cpp
```diff
@@ -831,6 +831,53 @@ void encodeVectorTile(const datafacade::ContiguousInternalMemoryDataFacadeBase &
...
+        // OSM Node tile layer
+        {
+            protozero::pbf_writer point_layer_writer(tile_writer, util::vector_tile::LAYER_TAG);
+            point_layer_writer.add_uint32(util::vector_tile::VERSION_TAG, 2);       // version
+            point_layer_writer.add_string(util::vector_tile::NAME_TAG, "osmnodes"); // name
+            point_layer_writer.add_uint32(util::vector_tile::EXTENT_TAG,
+                                          util::vector_tile::EXTENT); // extent
+
+            std::unordered_set<NodeID> internal_node_set;
+            for (const auto &edge : edges)
```
MoKob: Not really a problem, but I'd prefer one of a few other methods:

- std::vector + std::sort + std::unique

- std::set

Right now we get the slower unordered set for de-duplication, create a vector and sort after. Using a set directly would give you easy access to the sorted data. So you wouldn't go through multiple datasources here to get a sorted and unique set of entries.

The entire `internal_node_set` + `internal_nodes` process could be simplified to creating a set directly and loop over the set instead of the `internal_nodes` vector.

The `sort` + `unique` part would require twice as much memory for the initial vector, would would avoid creating the additional data structure.

unit_tests/library/tile.cpp
```diff
@@ -212,6 +213,58 @@ template <typename algorithm> void test_tile(algorithm &osrm)
 
     BOOST_CHECK_EQUAL(number_of_turn_keys, 4);
     BOOST_CHECK(number_of_turns_found > 700);
+
+    tile_message.next();
```
MoKob: This test is getting longer and longer, I would feel better if we split this into separate sub-functions to assert on. BOOST_CHECK should work from within a function call, right?

daniel-j-h:
This will be great to have - did you check how much bigger the tiles will become and how much longer it will take to generate and transmit tiles? The debug frontend is already a bit slow can you quickly check what's the impact here?

TheMarex:
@daniel-j-h our test tile in the node bindings seems to have increased by about 50% w.r.t `5.7`.

PR #4182, opened 2017-06-21
Fix/4060

### LINKED ISSUE #4060: cycling allowed against oneways if cycleway:right=*
the cycling profile incorrectly allows cycling against oneways if cycleway:right={track|lane}

```gherkin
    Scenario: Bike - Left/right cycleways on oneways
        Then routability should be
            | highway     | oneway | cycleway:left  | cycleway:right | forw    | backw        |
            | primary     | yes    |                |                | cycling | pushing bike |
            | primary     | yes    |                | track          | cycling | pushing bike |
            | primary     | yes    | track          |                | cycling | cycling      |
            | primary     | yes    | track          | track          | cycling | cycling      |

```
The second case above fails on master.

The bug is fixed in the PR #4033.

emiltin:
# Issue
Fixes #4060, incorrect cycling against oneways if cycleway:right=*.
The code  needs a bigger refactoring like in #4033, but for now this fixes the problem.

I adjusted two tests, since I think that cycleway=* should allow cycling on a footway.

## Tasklist

 - [x] review
 - [x] adjust for comments

### DIFF
--- a/features/bicycle/cycleway.feature
+++ b/features/bicycle/cycleway.feature
@@ -82,11 +82,11 @@ Feature: Bike - Cycle tracks/lanes
 
     Scenario: Bike - Cycleway on oneways, modes
         Then routability should be
-            | highway     | cycleway | oneway | forw         | backw        |
-            | motorway    | track    | yes    | cycling      |              |
-            | residential | track    | yes    | cycling      | pushing bike |
-            | cycleway    | track    | yes    | cycling      | pushing bike |
-            | footway     | track    | yes    | pushing bike | pushing bike |
+            | highway     | cycleway | oneway | forw    | backw        |
+            | motorway    | track    | yes    | cycling |              |
+            | residential | track    | yes    | cycling | pushing bike |
+            | cycleway    | track    | yes    | cycling | pushing bike |
+            | footway     | track    | yes    | cycling | pushing bike |
 
     Scenario: Bike - Cycleway on oneways, speeds
         Then routability should be
--- a/features/bicycle/cycleway.feature
+++ b/features/bicycle/cycleway.feature
@@ -90,8 +90,8 @@ Feature: Bike - Cycle tracks/lanes
 
     Scenario: Bike - Cycleway on oneways, speeds
         Then routability should be
-            | highway     | cycleway | oneway | forw       | backw      |
-            | motorway    | track    | yes    | 15 km/h    |            |
-            | residential | track    | yes    | 15 km/h    | 6 km/h     |
-            | cycleway    | track    | yes    | 15 km/h    | 6 km/h     |
-            | footway     | track    | yes    | 6 km/h +-1 | 6 km/h +-1 |
+            | highway     | cycleway | oneway | forw    | backw      |
+            | motorway    | track    | yes    | 15 km/h |            |
+            | residential | track    | yes    | 15 km/h | 6 km/h +-1 |
+            | cycleway    | track    | yes    | 15 km/h | 6 km/h +-1 |
+            | footway     | track    | yes    | 15 km/h | 6 km/h +-1 |
--- a/profiles/bicycle.lua
+++ b/profiles/bicycle.lua
@@ -309,7 +309,7 @@ function way_function (way, result)
   local maxspeed_backward = parse_maxspeed(way:get_value_by_key( "maxspeed:backward"))
   local barrier = way:get_value_by_key("barrier")
   local oneway = way:get_value_by_key("oneway")
-  local onewayClass = way:get_value_by_key("oneway:bicycle")
+  local oneway_bicycle = way:get_value_by_key("oneway:bicycle")
   local cycleway = way:get_value_by_key("cycleway")
   local cycleway_left = way:get_value_by_key("cycleway:left")
   local cycleway_right = way:get_value_by_key("cycleway:right")
@@ -321,6 +321,8 @@ function way_function (way, result)
   local bicycle = way:get_value_by_key("bicycle")
 
 
+  local way_type_allows_pushing = false
+
   -- speed
   local bridge_speed = profile.bridge_speeds[bridge]
   if (bridge_speed and bridge_speed > 0) then
@@ -330,6 +332,7 @@ function way_function (way, result)
     end
     result.forward_speed = bridge_speed
     result.backward_speed = bridge_speed
+    way_type_allows_pushing = true
   elseif profile.route_speeds[route] then
     -- ferries (doesn't cover routes tagged using relations)
     result.forward_mode = mode.ferry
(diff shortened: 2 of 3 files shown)


features/bicycle/cycleway.feature
```diff
@@ -82,16 +82,16 @@ Feature: Bike - Cycle tracks/lanes
...
-            | highway     | cycleway | oneway | forw       | backw      |
-            | motorway    | track    | yes    | 15 km/h    |            |
-            | residential | track    | yes    | 15 km/h    | 6 km/h     |
-            | cycleway    | track    | yes    | 15 km/h    | 6 km/h     |
-            | footway     | track    | yes    | 6 km/h +-1 | 6 km/h +-1 |
+            | highway     | cycleway | oneway | forw    | backw      |
+            | motorway    | track    | yes    | 15 km/h |            |
+            | residential | track    | yes    | 15 km/h | 6 km/h +-1 |
+            | cycleway    | track    | yes    | 15 km/h | 6 km/h +-1 |
+            | footway     | track    | yes    | 15 km/h | 5 km/h +-1 |
```
MoKob: @emiltin do you have any insight why this backward speed needed to change?

profiles/bicycle.lua
```diff
@@ -358,123 +362,137 @@ function way_function (way, result)
...
+    allowed_way_type = true
   end
 
-  -- direction
+  -- oneway
   local impliedOneway = false
   if junction == "roundabout" or junction == "circular" or data.highway == "motorway" then
     impliedOneway = true
   end
+  local onewayClassSet = false
```
MoKob: I know that this is already a bit inconsistent, but it looks like we also use `_` mainly instead of camelCase for variables. I'd prefer if you changed these new variables to that naming scheme.

profiles/bicycle.lua
```diff
@@ -320,6 +320,7 @@ function way_function (way, result)
   local foot_backward = way:get_value_by_key("foot:backward")
   local bicycle = way:get_value_by_key("bicycle")
 
+  local allowed_way_type = false
```
MoKob: from the name alone I wouldn't know if that `allowed` is allowed for cycling, pushing or something different. This is cycling, right? If so, I don't understand why we allow railway platforms in the same way we do with parking areas and so on. This is, I guess, inferred from the speeds? Still I feel the name is a bit confusing in that context. I don't know how I would name it to make it clear. Maybe add some commentary?

profiles/bicycle.lua
```diff
@@ -358,123 +362,137 @@ function way_function (way, result)
...
-      result.backward_speed = profile.bicycle_speeds["cycleway"]
-      result.backward_mode = mode.inaccessible
-    end
+    -- prevent other cases
   elseif oneway == "-1" then
     result.forward_mode = mode.inaccessible
-  elseif oneway == "yes" or oneway == "1" or oneway == "true" or impliedOneway then
+    reverse = true
+  elseif impliedOneway then
+    print('set implied')
```
MoKob: these look like left-over debugging code. Could you remove this and below?

features/bicycle/cycleway.feature
emiltin: i wondered about the difference between 5 and 6 km/h myself, but didn't immediately find an answer. the main point of the test is to verify that it's pedestrian speed.

profiles/bicycle.lua
emiltin: i agree

profiles/bicycle.lua
emiltin: yeah sorry about that

features/bicycle/cycleway.feature
MoKob: Ok. I wouldn't consider this a reason to hold off on this PR. Still a bit confusing.

profiles/bicycle.lua
emiltin: what about pushing_possible, since that's why it's needed: if tags like oneway means we can't bike, we need to check if we can push the bike. instead of rechecking all the way types we see if pushing_possible is set.

emiltin: or maybe way_type_allows_pushing

MoKob: I would feel much better about that name

emiltin:
adressed review comments

features/bicycle/cycleway.feature
emiltin: i think it's because each line in the test is placed on a different lon/lat. this means you get tiny differences in distances and thus some might round to 5 km/h and some to 6 km/h.

i you change the tests to to the same tags in all there cases, you still get 5 km/h for some and 6 km/h for others:

```cucumber
     Tables were not identical:
     | highway     | cycleway | oneway | forw    |     backw  |
     | motorway    | track    | yes    | 15 km/h |            |
     | residential | track    | yes    | 15 km/h | (-) 1 km/h |
     | residential | track    | yes    | 15 km/h | (+) 6 km/h |
     | residential | track    | yes    | 15 km/h | (-) 1 km/h |
     | residential | track    | yes    | 15 km/h | (+) 6 km/h |
     | residential | track    | yes    | 15 km/h | (-) 1 km/h |
     | residential | track    | yes    | 15 km/h | (+) 5 km/h |
```

MoKob:
Hm. Getting confused by all the travis failures. Most of them seem to be on completely unrelated items (e.g. car profile tests), some interact with bicycle profiles.

emiltin:
all test pass locally for me. pushing a new version to see if travis errors are spurious.

emiltin:
seems this is an assertion triggering in debug build if result.forward/backward_speed < 0 and the mode is set to inaccessible:

[assert][140624685709248] /home/emil/code/osrm-backend/src/extractor/extractor_callbacks.cpp:145
in: void osrm::extractor::ExtractorCallbacks::ProcessWay(const osmium::Way&, const osrm::extractor::ExtractionWay&): parsed_way.duration > 0 || parsed_way.forward_speed > 0

this surpises me a bit; i thought leaving the speed at the default -1 was fine and would mean simply cause the way direction to be ignored.

profiles/bicycle.lua
```diff
@@ -485,9 +489,21 @@ function way_function (way, result)
...
 
 
+
   -- maxspeed
   limit( result, maxspeed, maxspeed_forward, maxspeed_backward )
 
+
+  -- not routable if no speed assigned
+  -- this avoid assertions in debug builds
+  if result.forward_speed <= 0 then
```
MoKob: this added check seems kind of wrong to me. I feel we should not by default and only allow if a positive speed is set. How do you feel about this?
It seems that many other locations set the mode in combination with the speed (just from my first impression on searching for `forward_mode`)

emiltin: i fully agree, it seems very akward. i'm a bit puzzled that this problem didn't appear earlier, i would think setting the mode and not the speed would happen in other cases too, but appearently not.

not sure i understand your suggestion. but we would remove the assertion and instead silently ignore ways that have neither speed not duration.

emiltin:
all tests green. do you want to merge as it is now, or change the assert mechanism when speed <= 0?

MoKob:
I'd be alright with the current state. Just would prefer squashing to meaningful commit sets.

@TheMarex any reason to hold off on merging stuff right now?

emiltin:
squashed, keeping "adjust tests, cycleway allows cycling on footway" as a separate commit

emiltin:
any timeline for merging this?

TheMarex:
@emiltin ah sorry, didn't have time to look at this sooner. Looks good!

emiltin:
thanks for the merge

PR #4190, opened 2017-06-22
Refactor extractor file reading

TheMarex:
# Issue

This also works towards https://github.com/Project-OSRM/osrm-backend/issues/3899 by refactoring reading and writing of some extractor files.

## Tasklist
 - [x] review
 - [x] adjust for comments


include/util/concurrent_id_map.hpp
```diff
@@ -27,6 +27,23 @@ struct ConcurrentIDMap
...
+        data = std::move(other.data);
+    }
+    ConcurrentIDMap &operator=(ConcurrentIDMap &&other)
+    {
+        ScopedWriterLock other_lock{other.mutex};
+        ScopedWriterLock lock{mutex};
+
+        data = std::move(other.data);
+        return *this;
+    }
```
daniel-j-h: to be on the safe side I would add a check for `this != &other` otherwise a deadlock occurs in the code above

include/engine/routing_algorithms/routing_base.hpp
```diff
@@ -233,7 +233,7 @@ void annotatePath(const FacadeT &facade,
                      extractor::guidance::TurnInstruction::NO_TURN(),
                      {{0, INVALID_LANEID}, INVALID_LANE_DESCRIPTIONID},
                      facade.GetTravelMode(target_node_id),
-                     INVALID_ENTRY_CLASSID,
+                     {},
```
MoKob: These defines are something @daniel-j-h always pushed me to do and I quite like them by now. Do you think that it would be possible to preserve some naming here, instead of anonymous braces. We could of course use comments as well.

PR #4191, opened 2017-06-22
Fix turn lanes data propagation after traffic signs

### LINKED ISSUE #4189: Don't suggest u-turns from lanes that might cross straight-through traffic
This route indicates that u-turn maneuvers are valid from all lanes, which is wrong and dangerous:

http://map.project-osrm.org/?z=18&center=52.149544%2C9.938298&loc=52.150074%2C9.938040&loc=52.149857%2C9.938255&hl=en&alt=0

(Note: were not showing lane info for u-turns on the `osrm-frontend` GUI, so the data isn't visible, see below)
screen shot 2017-06-21 at 2 30 45 pm

screen shot 2017-06-21 at 2 31 43 pm

The OSM data for these ways only has `lanes=2`, there are no `turn:lanes` tags.  We should not suggest turning across potentially `straight` lanes in any situation.  If there are no `turn:lanes` data, we should issue only edge-hugging turn maneuvers.

oxidase:
# Issue

Fix #4189 

EDIT: the origin of missing turning lanes announcement  is in missing left turn in OSM.
Also the lane data was not propagating over  traffic signs and this PR fixes it.

## Tasklist
 - [x] review
 - [x] adjust for comments

## Requirements / Relations

### DIFF
--- a/features/guidance/turn-lanes.feature
+++ b/features/guidance/turn-lanes.feature
@@ -1242,5 +1242,5 @@ Feature: Turn Lane Guidance
             | mdhk  | road2 | 2     |                              | yes    |
 
         When I route I should get
-            | waypoints | route             | turns                        | lanes | locations |
-            | a,f       | road1,road1,road1 | depart,continue uturn,arrive | ,,    | a,d,f     |
+            | waypoints | route             | turns                        | lanes                                                | locations |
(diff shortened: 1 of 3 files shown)


oxidase:
@daniel-j-h no, there is no such threshold: if two edges can be combined and some edges have no lane information then lane data will be propagated even if a traffic light is between edges.

Issue #4065, opened 2017-05-18
Edge-weights update warning cucumber test is flunky


TheMarex:
Repeated execution of `features/options/contract/edge-weight-updates-over-factor.feature` will cause `https://travis-ci.org/Project-OSRM/osrm-backend/jobs/233362124#L2296` as the test will modify the `.geometry` file in-place.

This will cause the check to fail since the old value is already the high value.

oxidase:
Closed via #4199

PR #4199, opened 2017-06-26
Use thread-safe lock-free assignment in PackedVector::set_value

### LINKED ISSUE #4110: Traffic speed test depends on non-deterministic snapping behavior
The following test depends on random behavior and how snapping works out:

https://github.com/Project-OSRM/osrm-backend/blob/f3558260b6c0d73bcaf40171a63ccacd9a698a35/features/testbot/traffic_speeds.feature#L86-L117

There is no way to `g` (see the [network a bit above](https://github.com/Project-OSRM/osrm-backend/blob/f3558260b6c0d73bcaf40171a63ccacd9a698a35/features/testbot/traffic_speeds.feature#L4-L31)) and @MoKob pointed out snapping at `b` could be `eb` or `ab`. The test failure is currently on the [`mld/alternatives` branch](https://github.com/Project-OSRM/osrm-backend/pull/4047) but it's not the cause for it (it's not hitting the alternatives code path at all).

Here's the test diff:

https://travis-ci.org/Project-OSRM/osrm-backend/jobs/237850494#L2198

cc @danpat for non-deterministic snapping behavior

oxidase:
# Issue

PR fixes #4110 and #4065 that is caused by data races in `PackedVector::set_value` 
Lock-free `compare_exchange_weak` is implemented in <boost/atomic/detail/operations.hpp> as it is not possible to use `atomic_compare_exchange_weak` from <stdatomic.h> within C++

EDIT: the current approach requires to add Boost.Atomics as a new dependency, so  other approaches that do not introduce a global vector lock are welcome

## Tasklist
 - [x] review
 - [x] adjust for comments

### DIFF
--- a/include/util/packed_vector.hpp
+++ b/include/util/packed_vector.hpp
@@ -10,6 +10,7 @@
 
 #include <boost/iterator/iterator_facade.hpp>
 #include <boost/iterator/reverse_iterator.hpp>
+#include <tbb/atomic.h>
 
 #include <array>
 #include <cmath>
@@ -348,7 +349,7 @@ template <typename T, std::size_t Bits, storage::Ownership Ownership> class Pack
         fill(initial_value);
     }
 
-    PackedVector(util::ViewOrVector<std::uint64_t, Ownership> vec_, std::size_t num_elements)
+    PackedVector(util::ViewOrVector<WordT, Ownership> vec_, std::size_t num_elements)
         : vec(std::move(vec_)), num_elements(num_elements)
     {
         initialize();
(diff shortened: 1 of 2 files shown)


include/util/packed_vector.hpp
```diff
@@ -497,20 +498,38 @@ template <typename T, std::size_t Bits, storage::Ownership Ownership> class Pack
...
-        lower_word = set_lower_value<WordT, T>(lower_word,
-                                               lower_mask[internal_index.element],
-                                               lower_offset[internal_index.element],
-                                               value);
-        upper_word = set_upper_value<WordT, T>(upper_word,
-                                               upper_mask[internal_index.element],
-                                               upper_offset[internal_index.element],
-                                               value);
+        // Lock-free update of the lower word
+        WordT local_lower_word, new_lower_word;
```
daniel-j-h: WordT is a TBB typedef? Also is there a reason for not using the stdlib equivalents?

oxidase: `WordT` is an internal type alias. Please could you tell more about stdlib equivalents? So far i see only possible ways:
- use sequential update
- use an internal packed vector lock  -> makes packed vector non-movable
- use boost.atomic -> requires new dependency
- use boost.interprocess atomics implementation -> outdated and only 32 bit version
- use glib atomic's -> requires new dependency
- use gcc __sync_bool_compare_and_swap and msvc _InterlockedCompareExchange64 -> possible, but requires proper testing
- wait for https://isocpp.org/blog/2014/05/n4013 `as_atomic`
- use c11 _Atomic -> not possible
- use TBB atomic's -> uses TBB internal's

daniel-j-h: Urgh okay I see why you chose this solution now - thanks for the explanation! 🙇 
Could you add this as a comment to the code please.

oxidase: @daniel-j-h i added a comment line about TBB atomics usage, but placed the options list in the git commit message

PR #4200, opened 2017-06-27
Makes suffix list aware of 45-degree directions

daniel-j-h:
Teaches the suffix list to recognize e.g. "Northeast -> Northwest" changes. Should fix:

http://map.project-osrm.org/?z=18&center=38.903614%2C-77.008139&loc=38.904927%2C-77.009128&loc=38.904981%2C-77.008967&hl=en&alt=0

(see test below)

@willwhite @danpat

### DIFF
--- a/features/guidance/suffix-changes.feature
+++ b/features/guidance/suffix-changes.feature
@@ -128,3 +128,29 @@ Feature: Suppress New Names on dedicated Suffices
             | waypoints | route   | turns         |
             | a,c       | 42,42 S | depart,arrive |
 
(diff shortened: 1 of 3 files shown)


daniel-j-h:
Capturing from chat: suffix list already triggers, it's just that the streets are too wide (derived from the `lanes=4` tag) so that we don't emit a u-turn instruction.

Closing here - we might need to tune when the heuristics trigger here instead. Then we can use the test with a grid size of e.g. 20-25 meters.

PR #4202, opened 2017-06-27
Matrix plugin with MLD overlays

oxidase:
# Issue

The PR implements `manyToManySearch` with MLD overlays

## Tasklist
 - [x] add matrix plugin to `MLD` algorithms
 - [x] implement `relaxOutgoingEdges` with MLD overlays
 - [x] check performance of `unordered_multimap`
 - [x] review
 - [x] adjust for comments


src/engine/routing_algorithms/many_to_many.cpp
```diff
@@ -33,8 +33,7 @@ struct NodeBucket
     }
 };
 
-// FIXME This should be replaced by an std::unordered_multimap, though this needs benchmarking
-using SearchSpaceWithBuckets = std::unordered_map<NodeID, std::vector<NodeBucket>>;
+using SearchSpaceWithBuckets = std::unordered_multimap<NodeID, NodeBucket>;
```
oxidase: 😱 https://bugs.llvm.org/show_bug.cgi?id=16747
```
clang-3.3 with libc++ from 2013-01-31: 
Duration for 123456 integer inserts: 21836ms

gcc-4.6 with GLIBCXX_3.4.16:
Duration for 123456 integer inserts: 7ms
```

thanks for pointing! But i will definitely  re-check with the current gcc/clang versions

oxidase:
Comparison results of matrix plugin with CH vs MLD (used bayern-latest as it is large enough for time checks and not too large for 16GB RAM):

* dependency of response time in ms wrt matrix size for MLD with 5 levels matrix_bayern 

* dependency of response time in ms wrt number of MLD level for 25x25 matrix levels_bayern

The ratio `time(MLD)/time(CH)` does not depend on the matrix size and depends on MLD level sizes and graph size and is in range 7-20 for berlin-, bayern-, and germany-sized graphs.

src/engine/routing_algorithms/many_to_many.cpp
```diff
@@ -72,10 +104,116 @@ void relaxOutgoingEdges(const datafacade::ContiguousInternalMemoryDataFacade<Alg
...
+    const EdgeWeight weight,
+    const EdgeDuration duration,
+    typename SearchEngineData<mld::Algorithm>::ManyToManyQueryHeap &query_heap)
+{
+    const auto &partition = facade.GetMultiLevelPartition();
+    const auto &cells = facade.GetCellStorage();
+
+    const auto &node_data = query_heap.GetData(node);
+    const auto level =
+        std::max(node_data.level, partition.GetHighestDifferentLevel(node_data.parent, node));
```
oxidase: @TheMarex thanks for pointing to the edge case! I added a similar test case to check the current implementation. In your example node `e` will not be updated in forward and will have weight 21 that corresponds a path without crossing the cell border. In the reverse pass if the starting node is `e` the meeting point will be `v`  with weight 5 and the shortest path for `s->e` will be `s->a->b->d->v->e` with weight 5+1. Using level `partition.GetHighestDifferentLevel(node, source)` instead of the current one looks good to me, I will check it on Monday.

oxidase:
@TheMarex yeah, the test had ambiguous shortest paths that diverge at node `m` 
```
abcda [ 1, 1 ] depart undefined
cm [ 1.001798, 0.998202 ] turn right
mnopm [ 1.003596, 0.994606 ] new name slight right
mnopm [ 1.003596, 0.992808 ] turn left
mnopm [ 1.005394, 0.992808 ] arrive undefined
```
and
```
abcda [ 1, 1 ] depart undefined
cm [ 1.001798, 0.998202 ] turn right
mnopm [ 1.003596, 0.994606 ] turn left
mnopm [ 1.005394, 0.992808 ] arrive undefined
```

I added traffic signals at nodes `i` and `n`  for tiebreaking.

Issue #4203, opened 2017-06-27
Expose user-defined class information


TheMarex:
Breaking this out from https://github.com/Project-OSRM/osrm-backend/issues/4006 we need a way to set a user-defined class per edge and use this in the server. This will enable features like displaying when a route is using a toll road, or a private road.

### Lua profile

Instead of hard-coding specific classes for #4006 we want to keep this more general. This requires us to have an interface in the profiles to mark ways with `true`/`false` if they fall into that class.

```lua
function way_function(way, result)
   result.classes['highway'] = way:get_value_by_key('highway') == 'motorway'
   result.classes['ferry'] =  way:get_value_by_key('route') == 'ferry'
end
```

### HTTP/node

We can expose this as a new property `classes` in the `RouteStep` object, that is only there if there is class information for that step.

- `classes`: An array of strings signifying the classes of the road as specified in the profile.

Example:

```json
{
   "geometry" : "{lu_IypwpAVrAvAdI",
   "mode" : "driving",
   "classes": ["toll", "highway"],
   "duration" : 15.6,
   "weight" : 15.6,
   "intersections" : [
      {  "bearings" : [ 10, 92, 184, 270 ],
         "lanes" : [
            { "indications" : [ "left", "straight" ],
               "valid" : "false" },
            { "valid" : "true",
               "indications" : [ "right" ] }
         ],
         "out" : 2,
         "in" : 3,
         "entry" : [ "true", "true", "true", "false" ],
         "location" : [ 13.39677, 52.54366 ]
      },
      {  "out" : 1,
         "lanes" : [
            { "indications" : [ "straight" ],
               "valid" : "true" },
            { "indications" : [ "right" ],
               "valid" : "false" }
         ],
         "bearings" : [ 60, 240, 330 ],
         "in" : 0,
         "entry" : [ "false", "true", "true" ],
         "location" : [ 13.394718, 52.543096 ]
      }
   ],
   "name" : "Lortzingstraße",
   "distance" : 152.3,
   "maneuver" : {
      "modifier" : "right",
      "type" : "turn"
   }
}
```

danpat:
We should probably add something to `annotations` as well, in case someone wants the full route details:

```
annotations : { "classes" : [ ["highway","ferry"], ["highway"],["highway"],["highway"],["ferry"],["ferry"],["highway","ferry"] ] }
```

This would allow for correct rendering of various route sections if desired.

emiltin:
[...]

Issue #4006, opened 2017-05-04
Excluding specific roads at query-time


TheMarex:
This issue is a planning ticket for upcoming features in the 5.9 series.

# Exclude Overlays

## Goal

Enable users to exclude certain classes of roads at query time. The typical use case for this is excluding highways or ferries.

### Lua profile

Instead of hard-coding specific classes we want to keep this more general. This requires us to have an interface in the profiles to:

1. Specify the existence of a exclude flag combinations and what classes they cover
2. Mark ways with `true`/`false` if they fall into that exclude category. #4203

This is a sketch of what the interface could look like:

```lua
profile {
-- defines combination of classes that should supported for the exclude flags
"excludable": Sequence { Set { "motorway" }, Set { "ferry" }, Set { "motorway", "ferry" }},
"classes": Set {"motorway", "ferry", "restricted", "toll"}
}

function way_function(way, result)
   result.forward_classes['motorway'] = way:get_value_by_key('highway') == 'motorway'
   result.forward_classes['ferry'] =  way:get_value_by_key('route') == 'ferry'
end
```

### HTTP/node

We need a new query parameter to select which avoid layer (if any) to use.

- `exclude` Array of **String** Name of the classes to exclude. (Default: `none`)

Example:
`http://router.project-osrm.org/-100.063476,37.230328;-100.898437,35.817813?exclude=highway,ferry`

```json
{
   "coordinates": [[-100.063476,37.230328], [-100.898437,35.817813]],
   "exclude": ["highway", "ferry"]
}
```

## Implementation changes

1. While parsing the OSM ways we need to keep track of all referenced exclude classes and save all ways that are marked with `true`.

2. Flags need to be propagated up `EdgeBadedGraphFactory` and should be indexed by `EdgeBasedNode` (e.g. a node in the turn based search graph)

3. The algorithm dependent tools need to be extended:
     - `osrm-customize` will run the customization for every exclude layer (and with no exclude flags at all).
     - `osrm-contract` will create an own (filtered) graph for every exclude layer and contract it

5. We need to modify the engine to accommodate multiple overlays:
   - When `exclude` is passed the `RTree` should not snap on nodes with a matching exclude flag
   - `CellStorage` needs to be able to store multiple weight vectors (one for each exclude overlay)
   - We need a way to go from `string` to ID of the exclude layer in the `CellStorage`
   - We need an interface to get exclude flags for the base graph
   - The search function needs to be able to take advantage of optional exclude flags

[...]

PR #4204, opened 2017-06-27
Allow users to specify a class for each way

TheMarex:
This PR implements #4203 as a first step toward #4006.  We now allow users to mark ways with a specific class in the profile and expose this information in JSON response.

## Tasklist
 - [x] move `ClassData` to own header
 - [x] add tests for new JSON response property
 - [x] emit `Notification` instructions on class changes. 
 - [x] review
 - [x] adjust for comments


TheMarex:
@MoKob I would like to emit a `Notification` instruction when the `classes` of a way changes (similar to `mode` or `name`).

However I would only like to only do this for additions, for example:

- `classes: []` -> `classes: ["toll"]`
- `classes: ["highway"]` -> `classes: ["toll", "highway"]`

But not:

- `classes: ["toll"]` -> `classes: []` 
- `classes: ["toll", "highway"]` -> `classes: ["highway"]`

Does this approach make sense to you? Where would be the best position to hook into the guidance processing code to check for this?

MoKob:
@TheMarex I assume that you would be alright with not changing anything if a turn instruction is already emitted and only care about what happens if there is otherwise no instruction.

In that case, it should suffice if you go into `getInstructionForObvious` ([here](https://github.com/Project-OSRM/osrm-backend/blob/929e5a4de6cab989d5fe7352d6dcb874da0a8014/src/extractor/guidance/intersection_handler.cpp#L72)). All cases that think they might not require an announcement fall back to that function. Emitting your class change notification there should be the way to go.

In terms of the announcements, I feel that the classes are essentially a `travel_mode`. Could you elaborate your thought process in why you want to avoid notifications on `["toll"] -> []` for example? Wouldn't it be nice to know for an application that the segment of toll road is now over? E.g. if there is clear rules on how tolls are calculated based on length, we could ensure that a program knows about the end of the toll road.

I feel that in most cases the question will be void to begin with, since I assume that most of these changes will correspond to a `turn` of some kind anyhow. But if you ask my personal opinion, I would try and go with the same paradigm as the mode change, even if only for consistency.

TheMarex:
> Could you elaborate your thought process in why you want to avoid notifications

My main concerns was an increase in useless instruction that could be caused by this. For example a notification for `classes: ["restricted"]` -> `classes: []` seems silly. In terms of announcement "You are leaving a private road" doesn't really serve a purpose.

But you might be right, we probably already emit turn instructions in these cases anyway. We can start with behavior that is consistent with `mode` and circle back if there are problems.

MoKob:
@TheMarex I agree with the `useless` instruction part. One thing we might want to consider is where these `classes` are going to be exposed. Right now `mode` is associated with the entire RouteStep. As such, it needs to be announced when the mode changes. Is `classes` going to be an element on a per route step basis, or more a per `segment` (between intersections) element? If so, we might be able to omit it.

One side-effect we need to consider: how are classes going to be treated in collapsing? This might require a bit more thought, since it can affect how we can/where we give information about classes.

TheMarex:
Adding a notification instruction for all `classes` changes seems to dramatically increase the number of steps. For `10000` random requests on Bavaria already `2329` increase by at least one instruction or more.

I pulled the top offenders that increased the most and need investigation:

```
7.0,/route/v1/driving/13.607033773391777,47.965865035758775;11.207857196369844,47.95188612117001?steps=true
6.0,/route/v1/driving/13.478278091102993,48.43346245627276;11.399417314499374,47.88150602145781?steps=true
6.0,/route/v1/driving/13.296331636464345,48.45421744573064;11.586640589667354,48.146216415246215?steps=true
5.0,/route/v1/driving/13.704482824358024,47.930048523478575;11.486108038231748,48.11885578834559?steps=true
5.0,/route/v1/driving/13.672956595498533,47.99473229474607;13.30127680094705,47.899599218602816?steps=true
5.0,/route/v1/driving/13.588708303917088,48.13659419325602;13.312375268487244,47.883052210293855?steps=true
5.0,/route/v1/driving/13.585853501583358,47.99072141014606;9.611134280601732,48.418556180425156?steps=true
5.0,/route/v1/driving/13.562688493981625,47.96666856304219;9.696558358013476,48.48802221531398?steps=true
5.0,/route/v1/driving/13.559146820249737,48.202380427451004;9.474619241162543,47.88725799020104?steps=true
5.0,/route/v1/driving/13.468867525451458,48.33802959815377;11.505817593883307,48.20859636001675?steps=true
```

src/extractor/extractor_callbacks.cpp
```diff
@@ -175,6 +175,38 @@ void ExtractorCallbacks::ProcessWay(const osmium::Way &input_way, const Extracti
...
+        auto iter = classes_map.find(class_name);
+        if (iter == classes_map.end())
+        {
+            if (classes_map.size() > MAX_CLASS_INDEX)
+            {
+                throw util::exception("Maximum number of classes if " +
+                                      std::to_string(MAX_CLASS_INDEX + 1));
+            }
+            ClassData class_mask = 1u << classes_map.size();
+            classes_map[class_name] = class_mask;
```
oxidase: Here is a data race in `classes_map`

CHANGELOG.md
```diff
@@ -5,8 +5,11 @@
         - Plugins supported: `table`
     - API:
       - Support for exits numbers and names. New member `exits` in `RouteStep`, based on `junction:ref` on ways
+      - `RouteStep` now has new parameter `classes` that can be set in the profile on each way.
```
daniel-j-h: If I understand the classes feature correctly the user can set a class per way in the profile (way function) and we will then return it in the response `RouteStep`. Our `RouteStep` does not correspond to ways, so all the classes field tells the user is the road class _for this step_ but not for roads before or after it. Especially with the toll use-case I don't understand how this is useful then. Shouldn't classes be more like an annotation-feature then similar to e.g. osm ids?

daniel-j-h: (unrelated to that maybe add to the changeset that this is about _road classes_ and make an example)

include/engine/datafacade/contiguous_internalmem_datafacade.hpp
```diff
@@ -783,6 +789,22 @@ class ContiguousInternalMemoryDataFacadeBase : public BaseDataFacade
...
+    std::vector<std::string> GetClasses(const extractor::ClassData class_data) const override final
+    {
+        auto indexes = extractor::getClassIndexes(class_data);
+        std::vector<std::string> classes(indexes.size());
+        std::transform(indexes.begin(), indexes.end(), classes.begin(), [this](const auto index) {
+            return m_profile_properties->GetClassName(index);
+        });
+
+        return classes;
+    }
```
daniel-j-h: Hm I remember us switching to zero-copy string views - should we do the same here? See a few lines above the `StringView`s.

(for the response we still need to make a copy, this is only about _where_ we make the copy)

src/extractor/extractor_callbacks.cpp
```diff
@@ -175,6 +175,38 @@ void ExtractorCallbacks::ProcessWay(const osmium::Way &input_way, const Extracti
...
+            return iter->second;
+        }
+    };
+    const auto classesToMask = [&](const auto &classes) {
+        ClassData mask = 0;
+        for (const auto &pair : classes)
+        {
+            if (pair.second)
+            {
+                mask |= classStringToMask(pair.first);
```
oxidase: Another possibility is to use functional-like chain of transformations http://coliru.stacked-crooked.com/a/0c38afd1f996d167

include/extractor/class_data.hpp
```diff
@@ -0,0 +1,20 @@
...
+
+#include <cstdint>
+
+namespace osrm
+{
+namespace extractor
+{
+
+using ClassData = std::uint8_t;
+static const std::uint8_t MAX_CLASS_INDEX = 8 - 1;
```
daniel-j-h: If we constraint it to 8 bits to keep the class data small we should tell users this in the changelog/docs/profile docs.

(Btw overflowing this by setting a higher number from Lua should just wrap around and have the usual unsigned behavior, no? Should we check that in our Lua-exposed setter function?)

include/extractor/extractor_callbacks.hpp
```diff
@@ -84,6 +88,9 @@ class ExtractorCallbacks
 
     // destroys the internal laneDescriptionMap
     guidance::LaneDescriptionMap &&moveOutLaneDescriptionMap();
+
+    // destroys classes_map
+    ClassesMap &&moveOutClassMap();
```
daniel-j-h: I don't understand the `R&& fn();` return value here. Shouldn't this be `R fn();` and internally we should probably set a `is_dead` boolean to allow only for calling it once?

The move-out here and with the lane map above is a bit ugly, maybe there should be a free-standing function `ClassesMap makeClassesMap(ExtractorCallbacks&);` to better show intent? The call-once issue would still persist then, though.

include/extractor/profile_properties.hpp
```diff
@@ -86,6 +105,8 @@ struct ProfileProperties
     bool fallback_to_duration;
     //! stores the name of the weight (e.g. 'duration', 'distance', 'safety')
     char weight_name[MAX_WEIGHT_NAME_LENGTH + 1];
+    //! stores the names of each class
+    std::array<char[MAX_CLASS_NAME_LENGTH + 1], MAX_CLASS_INDEX + 1> class_names;
```
daniel-j-h: Hm why not a vector of strings? Serialization and deserialization should be independent of how we store it here I think?

include/util/bit_range.hpp
```diff
@@ -0,0 +1,95 @@
...
+
+namespace osrm
+{
+namespace util
+{
+
+namespace detail
+{
+template <typename T> std::size_t countOnes(T value)
+{
```
daniel-j-h: Can you static assert `is_unsigned` here

include/util/bit_range.hpp
```diff
@@ -0,0 +1,95 @@
...
+inline std::size_t countOnes(std::uint8_t value)
+{
+    return __builtin_popcount(std::uint32_t{value});
+}
+inline std::size_t countOnes(std::uint16_t value)
+{
+    return __builtin_popcount(std::uint32_t{value});
+}
+inline std::size_t countOnes(std::uint32_t value) { return __builtin_popcount(value); }
+inline std::size_t countOnes(std::uint64_t value) { return __builtin_popcountl(value); }
```
daniel-j-h: I think you need `popcountll` (note the `ll` suffix) for 64 bit
- https://gcc.gnu.org/onlinedocs/gcc/Other-Builtins.html
- http://en.cppreference.com/w/cpp/language/types#Properties

include/util/bit_range.hpp
```diff
@@ -0,0 +1,95 @@
...
+#endif
+}
+
+template <typename DataT>
+class BitIterator : public boost::iterator_facade<BitIterator<DataT>,
```
daniel-j-h: Maybe boost's dynamic bitset could work here instead of our own impl.?

src/extractor/extractor_callbacks.cpp
```diff
@@ -175,6 +175,38 @@ void ExtractorCallbacks::ProcessWay(const osmium::Way &input_way, const Extracti
         }
     }
 
+    const auto classStringToMask = [this](const std::string &class_name) {
+        auto iter = classes_map.find(class_name);
+        if (iter == classes_map.end())
```
daniel-j-h: Can we use the string map for deduplicating both strings as well as the classes here? Or better: I don't understand why we need another deduplication map.

src/extractor/extractor_callbacks.cpp
```diff
@@ -175,6 +175,38 @@ void ExtractorCallbacks::ProcessWay(const osmium::Way &input_way, const Extracti
...
+        ClassData mask = 0;
+        for (const auto &pair : classes)
+        {
+            if (pair.second)
+            {
+                mask |= classStringToMask(pair.first);
+            }
+        }
+        return mask;
+    };
```
daniel-j-h: Hm this is a bit hard to follow with the auto and first/second working on basically arbitrary types - can you make it a bit more explicit?

src/extractor/guidance/intersection_handler.cpp
```diff
@@ -139,22 +144,22 @@ TurnInstruction IntersectionHandler::getInstructionForObvious(const std::size_t
...
+            if (needs_notification)
                 return {TurnType::Notification, getTurnDirection(road.angle)};
             else
                 return {num_roads == 2 ? TurnType::NoTurn : TurnType::Suppressed,
                         getTurnDirection(road.angle)};
         }
     }
     BOOST_ASSERT(type == TurnType::Continue);
-    if (in_mode != out_mode)
+    if (needs_notification)
```
daniel-j-h: Ah so for _all_ road class changes we issue a notification instruction 👍

oxidase:
@TheMarex 7 additional notifications hit A94/B12 https://www.merkur.de/lokales/erding/gegner-wird-eine-geister-autobahn-3028059.html 
```
"notification, straight, [13.220145,48.328256]"
"notification, straight, [13.171839,48.307238]"
"notification, straight, [12.840767,48.244172]"
"notification, straight, [12.364108,48.247667]"
"notification, straight, [11.627202,48.138513]"
"notification, straight, [11.501258,48.102291]"
"notification, straight, [11.361197,48.000942]"
```

TheMarex:
I changed the behavior to only announce when the new classes are not a subset of the old classes (we only remove classes). This changed the number of added notifications dramatically down to `904` from `2399`. The rational here is that in these cases we usually don't need announcements for the user.

I spot checked a few of the routes that saw a large increase. The maximum was 3 added instructions on very long routes that go over the spot that Michael found.

src/extractor/extractor_callbacks.cpp
TheMarex: Can you elaborate? This code is not executed multi-threaded.

CHANGELOG.md
TheMarex: This is not about road classes? This flag is user-defined, it doesn't expose our road-class hierarchy. The classes are mostly for two things:

1/ Announce to the user when we enter a new class (e.g. private road or toll road)
2/ Begin able to mark a step as using toll roads/private roads whatever in the UI

Main reason I don't want is as annotation is that at the moment we just use data that we already need for the avoid flags anyway (edge-based-node level), for annotations we need segment-level which means a lot more data.

include/extractor/class_data.hpp
TheMarex: We would throw in the extractor callbacks when we detect more then 8 flags. There is a changelog entry for this.

include/engine/datafacade/contiguous_internalmem_datafacade.hpp
TheMarex: Hrm the problem is that immediately store these in the `RouteStep` object, which should already own all the contained data, since we might want to expose this object outside of OSRM in the future. (like we already do for all the string data)

 I think I'm going to pass on this optimization since it won't actually help right now. :-/

include/extractor/extractor_callbacks.hpp
TheMarex: After thinking about this they should not be owned by `ExtractorCallbacks` in the first place. Going to change that.

include/extractor/profile_properties.hpp
TheMarex: This struct needs to have a fixed size because we store it as a single block in shared memory.

include/util/bit_range.hpp
TheMarex: Added a comment in the code for investigating this when needed.

src/extractor/extractor_callbacks.cpp
TheMarex: We need this for indexing: The number of elements in the classes map gives us a continuous ID per class.

MoKob:
@TheMarex if the classes change indeed that often, going to the removal version seems ok to me. Will the full classes be visible somewhere in the output, though? (out of pure interest)

src/extractor/extractor_callbacks.cpp
oxidase: You are right, just missed that the function is called in  `serial_in_order` filter.

TheMarex:
@mokob yes as a property of the `RouteStep`.

CHANGELOG.md
daniel-j-h: > This is not about road classes? This flag is user-defined, it doesn't expose our road-class hierarchy.

Ah sorry I think I have a different understanding for "road classes" than our road class hierarchy. Thanks for the explanations, all clear here.

include/extractor/extractor_callbacks.hpp
daniel-j-h: Thanks for refactoring this!

daniel-j-h:
> I changed the behavior to only announce when the new classes are not a subset of the old classes (we only remove classes). This changed the number of added notifications dramatically down to 904 from 2399. The rational here is that in these cases we usually don't need announcements for the user.

Sounds reasonable from a performance perspective - but now users can no longer detect when e.g. going off of a toll road, no?

TheMarex:
@daniel-j-h yes there can be cases where this would hide this changes. We could add special logic to make certain flags "notifiable" in this case, but I would first like to establish a need for this.

Maybe we need annotation-level toll data anyway (different pricing, etc) so this might not actually be in the critical path of that specific use-case. This should be good enough though for navigation that wants to announce that there is a toll road coming up. (which is a good warning, because you can pay $$$ for this)

emiltin:
very nice feature.

the function Handlers.handle_classes seems to be a bit car-focused with toll and motorways. perhaps it should be more generic or moved to the car profile?

PR #4206, opened 2017-06-28
added check for sharp angles between ramps

### LINKED ISSUE #4146: Provide an IntersectionValidationHandler

ghoshkaj:
# Issue

[Provide an IntersectionValidationHandler](https://github.com/Project-OSRM/osrm-backend/issues/4146)

## Tasklist
 - [x] Write the check + cucumber test for a small testing subject. Don't commit the cuke test
 - [x] Run on the original part of the world where we saw this problem: http://map.project-osrm.org/?z=18&center=52.309541%2C9.871645&loc=52.309089%2C9.872187&loc=52.310581%2C9.870207&hl=en&alt=0.  (Download the OSM pbf of this, use JOSM to remove the fixed turn restriction that Moritz added, and run the extractor on this pbf. <-- We actually found another sharp turn between ramps across from the reported one!)
 - [x] Run on whole of Germany to fine-tune the check (whether we should whittle down by different road classification priorities)
 - [x] Review
 - [x] Adjust for comments
 - [x] Merge

Optionally:
 - [x] Make a pretty map and OSM diary about it!

## Requirements / Relations
[Implements Intersection Validation Handler, resolves #4146](https://github.com/Project-OSRM/osrm-backend/pull/4160)

cc/ @MoKob @daniel-j-h

### DIFF
--- a/include/extractor/guidance/validation_handler.hpp
+++ b/include/extractor/guidance/validation_handler.hpp
@@ -46,7 +46,8 @@ class ValidationHandler final : public IntersectionHandler
                             const EdgeID via_eid,
                             Intersection intersection) const override final
     {
-        checkForSharpTurnsOntoRamps(nid, via_eid, intersection);
+        // checkForSharpTurnsOntoRamps(nid, via_eid, intersection);
+        checkForSharpTurnsBetweenRamps(nid, via_eid, intersection);
 
         return intersection;
     }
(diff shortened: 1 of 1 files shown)


ghoshkaj:
Here's a map!
https://api.mapbox.com/styles/v1/ghoshkaj/cj4o8yuai4zuf2sk8y70pzvq6.html?title=true&access_token=pk.eyJ1IjoiZ2hvc2hrYWoiLCJhIjoiY2oxNmphOXExMDJseTJ5bzZjcG9tamN0NSJ9.WaB4LiJml37WV0OlblGUVg#5.12/51.357/7.411
angle<100 = green
200<angel<300 = red
angle > 300 = purple

include/extractor/guidance/validation_handler.hpp
```diff
@@ -46,7 +46,8 @@ class ValidationHandler final : public IntersectionHandler
                             const EdgeID via_eid,
                             Intersection intersection) const override final
     {
-        checkForSharpTurnsOntoRamps(nid, via_eid, intersection);
+        // checkForSharpTurnsOntoRamps(nid, via_eid, intersection);
```
MoKob: this should probably be left active, now that we know our check works. @daniel-j-h any thoughts on applying categories here to distinguish between the handlers results? Or simply keep for now, as both consider ramps.

daniel-j-h: eventually the validation handler should store deduplicated

    (intersection, category)

tuples and on destruction report them as geojson with properties attached to the features.

For now I think this is fine.

ghoshkaj: @daniel-j-h so for now, I'll just uncomment your check and we'll let both of them run one after the other correct?

daniel-j-h: Yes

Issue #3400, opened 2016-12-04
Include Exit Numbers and expose them in RouteStep


roman-yepishev:
Based on #228 routing engine supposedly can provide information on exit number, however currently the engine tells "take the ramp on the right" or "keep right at the fork".

I was unable to find any references for motorway_junction in the code, so I assume the functionality to relay the exit numbers from OSM is not present.

Is it something OSRM plans to include?

daniel-j-h:
We handle exit / destination tags for the `destination=` key on ways:

https://wiki.openstreetmap.org/wiki/Key:destination

We don't handle `exit_to=` tags on nodes:

https://wiki.openstreetmap.org/wiki/Key:exit_to

mostly because the node tag is a historic mistake used primarily for rendering purpose and hard to work around on our side. Closing as duplicate of https://github.com/Project-OSRM/osrm-backend/issues/3317.

roman-yepishev:
Hm, how to get OSRM to provide the exit _number_?AFAIR ref= on the highway=motorway_junction is the exit number and destination tag is not for setting the exit number.On Dec 4, 2016 15:52, "Daniel J. H." <notifications@github.com> wrote:We handle exit / destination tags for the destination= key on ways:
https://wiki.openstreetmap.org/wiki/Key:destination
We don't handle exit_to= tags on nodes:
https://wiki.openstreetmap.org/wiki/Key:exit_to
mostly because the node tag is a historic mistake used primarily for rendering purpose and hard to work around on our side. Closing as duplicate of #3317.

danpat:
@roman-yepishev you can't right now, we don't extract the necessary tags, nor do we support an "exit number" attribute in the API response.  We only support destination tags so far.

@daniel-j-h I think we should consider re-opening this ticket as a separate enhancement to #3317.  Exit numbers and destinations aren't the same thing, we might actually want both in the response for highway ramps.  At least in the US, exits are numbered and navigation like "Take exit 33 onto the I95 towards Spokane" (Take exit `exit_to` onto the `ref/name` towards `destination:ref`).

http://wiki.openstreetmap.org/wiki/Proposed_features/junction_details

Near me, they seem to be tagged on junction nodes with `highway=motor_junction` and `ref=NNN`.  Example:

https://www.mapillary.com/app/?lat=47.637103&lng=-117.473553&z=17&pKey=FeCWcVDAWrVk4HfK_F4vcw&focus=photo

Note the "Exit 277" sign.

[...]

PR #4215, opened 2017-06-29
Implements Exit Numbers + Names

daniel-j-h:
For https://github.com/Project-OSRM/osrm-backend/issues/3400: implements exit numbers and exit names using the osm `junction:ref` way tag for now. Exposes a `exits` property in the response's `RouteStep`. 

- [x] Measure increase in memory consumption
- [x] Review
- [x] Adjust for comments

cc @miccolis @willwhite @danpat


include/util/guidance/name_announcements.hpp
```diff
@@ -129,16 +131,34 @@ inline bool requiresNameAnnounced(const std::string &from_name,
...
+    //         ` e . . f
+    //
+    // could assign the exit number to `be` when exiting `abcd` instead of the full ramp.
+    //
+    // Issuing a new-name instruction here would result in the turn assuming the short segment to be
+    // irrelevant and remove the exit number in a collapse scenario. We don't want to issue any
+    // instruction from be-ef, since we only loose the exit number. So we want to make sure that we
+    // don't just loose an exit number, when exits change
+    const auto exits_change = from_exits != to_exits;
+    const auto looses_exit = (names_are_equal && !from_exits.empty() && to_exits.empty());
```
TheMarex: I'm unclear on what the overall effect here is supposed to be, please clarify on which these are true:

1/ We emit a new-name instruction at `be->ef`
2/ We don't emit a new-name instruciton at `be->ef`
3/ We just ensure these steps won't be collapsed
4/ We just ensure the exit number will be copied to the new step

MoKob: The overall effect is 2), not announcing a new name. If we would announce a new name, the engine would see a short segment after which a new name is given and use the second name for the ramp instruction.
By not announcing a new name, we ensure that the name with the exit number is used for the ramp.

MoKob:
Some insight into index change: on Germany, my index increases from 9 MB to 9.2 MB.

CHANGELOG.md
```diff
@@ -5,6 +5,9 @@
       - new parameter `approaches` for `route`, `table`, `trip` and `nearest` requests.  This parameter keep waypoints on the curb side.
         'approaches' accepts both 'curb' and 'unrestricted' values.
         Note : the curb side depend on the `ProfileProperties::left_hand_driving`, it's a global property set once by the profile. If you are working with a planet dataset, the api will be wrong in some countries, and right in others.
+      - Support for exits numbers and names. New member `exits` in `RouteStep`, based on `junction:ref` on ways
```
TheMarex: Just noticed that this was added on the `5.8` entry, not a `5.9` one.

TheMarex:
Urg I think travis updated eslint on the OSX image, I think the failure is unrelated. Needs a fix on `master`.

PR #4218, opened 2017-06-30
Fix instruction for turning on through street that is not straight

### LINKED ISSUE #4205: Wrong modifier in StepManeuver
Sometimes the modifier in the StepManeuver is wrong. The bearings in the StepManeuver are correct. Here is an example:

wrong-turn

Link: http://map.project-osrm.org/?z=18&center=51.975672%2C7.617731&loc=51.975817%2C7.618520&loc=51.975754%2C7.616926&hl=en&alt=0

MoKob:
# Issue

Resolves #4205 

Turning onto a through street often feels like going straight. We currently consider all angles though, which is not what is intended. This PR changes the requirements to consider only narrow turns as going straight, improving the behaviour on merges.

## Tasklist
 - [x] review
 - [x] adjust for comments

### DIFF
--- a/features/guidance/turn-angles.feature
+++ b/features/guidance/turn-angles.feature
@@ -1371,3 +1371,75 @@ Feature: Simple Turns
         When I route I should get
             | waypoints | route       |
             | g,e       | abcde,abcde |
+
+    Scenario: small ticks in roads
+        Given the node map
+            """
+                          e
+                           `
+                            `
+                             `
+                              `
+                               d
+                                c
+            a - - - - - - - b``
+                                  `
+
+                                     `
+
+                                        `
+
+                                           `
+
+                                              `
+
+                                                  `
+
+                                                      `
+
+                                                          `
+
+                                                              `
+
+                                                                  `
+
+                                                                      `
+
+                                                                          `
+
+                                                                              `
+
+                                                                                  `
+
+                                                                                      `
+
+                                                                                          `
+
+                                                                                              `
+
+                                                                                                 `
+                                                                                                   `
+
+                                                                                                       `
+
+
+                                                                                                            `
+
+
+                                                                                                                `
+
+
+                                                                                                                    f
+
+            """
+
+        And the ways
+            | nodes | oneway | name |
+            | abc   | yes    | fww  |
+            | fcde  | no     | jahn |
+
+        When I route I should get
+            | waypoints | route         | turns                       |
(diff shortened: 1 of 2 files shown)


features/guidance/turn-angles.feature
```diff
@@ -1371,3 +1371,74 @@ Feature: Simple Turns
         When I route I should get
             | waypoints | route       |
             | g,e       | abcde,abcde |
+
+    Scenario: small ticks in roads
```
daniel-j-h: care to add a osm link to this scenario

MoKob:
This PR fixes my new test, but still emits turn straight at the location mentioned in #4205, which is a bit over-eager. Needs some fine-tuning

src/extractor/guidance/intersection_handler.cpp
```diff
@@ -129,12 +131,12 @@ TurnInstruction IntersectionHandler::getInstructionForObvious(const std::size_t
...
                         coordinates[node_based_graph.GetTarget(road.eid)]);
-                    return {
-                        TurnType::Turn,
-                        (angularDeviation(road.angle, STRAIGHT_ANGLE) < FUZZY_ANGLE_DIFFERENCE ||
-                         distance > 2 * MAX_COLLAPSE_DISTANCE)
-                            ? DirectionModifier::Straight
-                            : getTurnDirection(road.angle)};
+
+                    return {TurnType::Turn,
+                            (angularDeviation(road.angle, STRAIGHT_ANGLE) < NARROW_TURN_ANGLE &&
```
daniel-j-h: we return a straight below 40 degree instead of only below 25 degree now? isn't the issue exactly the other way around: we return a straight but the user wants a turn? (from the original wording in the issue it's a bit unclear to me what the user expects to be honest)

MoKob:
fix

Taking everything back, got confused in the tabs. Local version is now correct

src/extractor/guidance/intersection_handler.cpp
MoKob: The issue was, that we did return 'DirectionModifier::Straight' on ALL angles before. Now we do it to at most 40 degrees

MoKob: (big difference here: `||` -> `&&`

TheMarex:
I this relevant for 5.8? If so please tag with `5.8.2`.

PR #4220, opened 2017-06-30
readme example curl command needs quotes

harry-wood:
We need to add quotes around the URL in the curl command (at least in the shell
I'm using). Otherwise it truncates the URL at ';'


TheMarex:
Heh, yeah good catch. Thanks!

Issue #4152, opened 2017-06-14
more lua speedup?


emiltin:
inspired by #4147, what about the profile providing a list of tags, and only calling the node function if at least one of the tags are present? the node functions typically look at a limited number of tags. e.g. the car profile:

```lua
function node_function (node, result)
  -- parse access and barrier tags
  local access = find_access_tag(node, profile.access_tags_hierarchy)
  if access then
    if profile.access_tag_blacklist[access] and not profile.restricted_access_tag_list[access] then
      result.barrier = true
    end
  else
    local barrier = node:get_value_by_key("barrier")
    if barrier then
      --  make an exception for rising bollard barriers
      local bollard = node:get_value_by_key("bollard")
      local rising_bollard = bollard and "rising" == bollard

      if not profile.barrier_whitelist[barrier] and not rising_bollard then
        result.barrier = true
      end
    end
  end

  -- check if node is a traffic light
  local tag = node:get_value_by_key("highway")
  if "traffic_signals" == tag then
    result.traffic_lights = true
  end
end
```

which only looks at the tags: access, vehicle, motor-vehicle,vehicle, barrier, bollard, highway
if none of these tags are present, there's no need to call the node function.

daniel-j-h:
@emiltin yep the idea came up during discussion, too. Basically the idea is to pull out a list of tags from Lua into C++ and then do the fast tag comparison on the C++ side already. @danpat and @joto found the approach in #4147 to be sufficient for a first try and easy to implement.

danpat:
There are definitely some more speedups to be had with this general approach, but the improvements will be less significant than the tagless node filtering.  I noted some other experiments I did here https://github.com/Project-OSRM/osrm-backend/pull/4147#issuecomment-308047549

Basically, for the planet we have 4.6 billion nodes, only 180 million of which have any tags.  That's ~4.3 `node_functions` we can avoid.  In comparison, doing tag checking on the remaining nodes or ways will reduce the call count by a few million, a much less significant saving.  I did some hard-coded testing and I saw ~8-10% speedups by filtering out way tags like `buildings` or checking for `access/highway` on nodes.

It's worth keeping this ticket though, I think it's generally a decent further optimization.

emiltin:
yes, it would be a smaller gain

PR #4240, opened 2017-07-05
Process relevant nodes only

emiltin:
# Issue

During extraction, processing of nodes takes a lot of time, because there are so many. With this PR, the LUA node function is only called if at least one of a list of relevant tags is present. The list is set by the LUA profile.

The list of relevant tags is typically small. For car it's:
```lua
    node_tags_requiring_processing = {
      'barrier',
      'highway',
      'motorcar',
      'motor_vehicle',
      'vehicle',
      'access'
    }
```

Testing on data for Denmark with the car profile, it cuts the parsing step from about 24s to 16s, or around 30%. The total time of running osrm-extract is reduced from about 135s to 117s, or around 15%.

This PR is based on the profiles v2 branch, and should therefore NOT be merged into the current master.

### DIFF
--- a/profiles/bicycle.lua
+++ b/profiles/bicycle.lua
@@ -194,6 +194,15 @@ function setup()
     avoid = Set {
       'impassable',
       'construction'
+    },
+
+    node_tags_requiring_processing = {
+      'barrier',
+      'highway',
+      'crossing',
+      'bicycle',
+      'vehicle',
+      'access'
     }
   }
 end
--- a/src/extractor/scripting_environment_lua.cpp
+++ b/src/extractor/scripting_environment_lua.cpp
@@ -495,6 +495,16 @@ void Sol2ScriptingEnvironment::InitContext(LuaScriptingContext &context)
         context.has_way_function = context.way_function.valid();
         context.has_segment_function = context.segment_function.valid();
 
+        // fetch list of tags requiring node processing
+        sol::table node_tags_table = context.profile_table["node_tags_requiring_processing"];
+        if (node_tags_table.valid())
+        {
+            for (auto &&pair : node_tags_table)
+            {
+                context.node_tags_requiring_processing.push_back(pair.second.as<std::string>());
+            };
+        }
+
         // read properties from 'profile.properties' table
         sol::table properties = context.profile_table["properties"];
         if (properties.valid())
@@ -812,8 +822,28 @@ void LuaScriptingContext::ProcessNode(const osmium::Node &node, ExtractionNode &
     switch (api_version)
     {
     case 2:
-        node_function(profile_table, node, result);
+    {
+        bool process = false;
+
+        if (node_tags_requiring_processing.empty())
+            process = true;
+        else
+        {
+            for (auto &&tag : node_tags_requiring_processing)
+            {
+                auto v = node.get_value_by_key(tag.c_str());
+                if (v && *v)
+                {
+                    process = true;
+                    break;
+                }
+            }
+        }
+
+        if (process)
+            node_function(profile_table, node, result);
         break;
+    }
     case 1:
     case 0:
         node_function(node, result);
--- a/features/options/profiles/version2.feature
+++ b/features/options/profiles/version2.feature
@@ -85,3 +85,159 @@ Feature: Profile API version 2
            | a    | b  | ac,cb,cb | 19.2s |
            | a    | d  | ac,cd,cd | 19.2s |
            | a    | e  | ac,ce    | 20s   |
+
+    Scenario: Process only nodes with relevant tags
+        Given the profile file
+            """
+            functions = require('lib/profile_v2')
+
+            function setup()
+              return {
+                node_tags_requiring_processing = {
+                  'barrier'
+                }
+              }
+            end
+
+            function process_node(profile, node, result)
+              print ('process_node ' .. node:id())
+            end
+
+            function process_way(profile, way, result)
+              result.name = way:get_value_by_key('name')
+              result.weight = 10
+              result.forward_mode = mode.driving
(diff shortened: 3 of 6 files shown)


emiltin:
The process_call_tagless_node flag is not needed anymore. If the list of relevant tags is empty, the node function is always called.

emiltin:
A smal drawback is that the profile must define the list of relevant tag, otherwise all nodes are processed. We could keep the process_call_tagless_node flag to avoid that - unless it's set we don't process nodes without tags.

emiltin:
a further optimization might be to keep the checked tags and pass then to the node processing function, so the LUA code don't have to fetch them again one at a time.

emiltin:
implements #4152

emiltin:
rebased on #4072.

emiltin:
any thoughts on this PR?

features/options/contract/edge-weight-updates-over-factor.feature
```diff
@@ -33,15 +33,25 @@ Feature: osrm-contract command line option: edge-weight-updates-over-factor
 
 
     Scenario: Logging using weigts as durations for non-duration profile
-        Given the profile file "testbot" extended with
+        Given the profile file
```
MoKob: Is this a necessary change with the new API? I would hate to maintain these in addition to the testbot profile, if only a specific entry should change. In my view we should maintain as much as we can within the main profile and only adapt what is absolutely necessary.

features/guidance/anticipate-lanes.feature
```diff
@@ -582,9 +582,9 @@ Feature: Turn Lane Guidance
 
     @anticipate
     Scenario: No Lanes for Roundabouts, see #2626
-        Given the profile file "car" extended with
+        Given the profile file "car" initialized with
         """
-        properties.left_hand_driving = true
+        profile.left_hand_driving = true
```
emiltin: @TheMarex requested that properties be grouped under profile.properties. see discussion at https://github.com/Project-OSRM/osrm-backend/pull/4072#issuecomment-309457857

features/options/contract/edge-weight-updates-over-factor.feature
emiltin: this change is part of the already merged profile v2

src/extractor/scripting_environment_lua.cpp
```diff
@@ -812,8 +822,28 @@ void LuaScriptingContext::ProcessNode(const osmium::Node &node, ExtractionNode &
     switch (api_version)
     {
     case 2:
-        node_function(profile_table, node, result);
+    {
+        bool process = false;
+
+        if (node_tags_requiring_processing.empty())
+            process = true;
```
emiltin: yes, empty means process all tags.

features/options/contract/edge-weight-updates-over-factor.feature
MoKob: @emiltin could you rebased in that case to make reviewing easier?

features/guidance/anticipate-lanes.feature
MoKob: Ok

features/options/contract/edge-weight-updates-over-factor.feature
emiltin: will do

emiltin:
rebased on master

profiles/bicycle.lua
```diff
@@ -188,6 +188,15 @@ function setup()
...
       'construction'
+    },
+
+    node_tags_requiring_processing = {
+      'barrier',
+      'highway',
+      'crossing',
+      'bicycle',
+      'vehicle',
+      'access'
```
emiltin: a possibility would be to do test-run where you build a list of all tags looked at by the profile. 

or maybe it would be possible to keep all node-relevant tags in a single parent table (with sub tables as needed), so you can collect the total set of tags automatically? 

but in practice i think this is a small problem, since the process_node function is typically rather simple and is rarely modified.

emiltin:
@MoKob regarding your concern "can we verify somehow that all our tests actually work with both profile versions?".

i don't see it as two profile versions. it's just an optimisation that prevents unnecessary calls to process_node. as long as the list of tags is correct, the change should not affect profile behaviour or test result, only performance.

the tests in https://github.com/Project-OSRM/osrm-backend/pull/4240/files#diff-0eb7c47949e04b2d23425b733646c5a2R89 tests the new behaviour - a node is only processed if it has at least one tags in the list present.

TheMarex:
👋 Sorry didn't really had time to look at this until now. A few notes:

We might switch to multi-pass processing. (Parse the OSM PBF file multiple types) This might mean we will switch to splitting the `process_{node,way}` function into two parts:

- `filter_node` and `filter_way` that just determine if the node/way is necessary/drivable
- `process_node` and `process_way` (maybe also `process_relation` ?) that determine all the values associated with the entity

I'm unsure how the above proposal would fit into that. The performance increase is nice, but for larger datasets the bulk of the time is spend in later parts of `osrm-extract` (computing weights for edges, graph compression, parts of the EBGF). The nice thing about profile stuff is that it is highly parallelized so usually inefficiencies don't increase data turn-around much.

emiltin:
i guess the way it would fit would be to avoid unnecessary calls to filter_node

emiltin:
i only benchmarked this with data for denmark. i'll let you decide if you want to benchmark with larger datasets to see if the approach is worthwhile. the change in code is quite small, and existing profiles would still work. only drawback is that you need to update the list if you modify the process_node function or an profile settings that's used in the function.

TheMarex:
Outdated.

Issue #4221, opened 2017-06-30
Wild routing oddity


joovel:
Here's a brief script to capture this odd result I'm getting from a routing:

```
mkdir -p data/osrm
pushd data/osrm
wget https://s3.amazonaws.com/metro-extracts.mapzen.com/los-angeles_california.osm.pbf
popd
docker run --tty --volume `pwd`/data/osrm:/data osrm/osrm-backend:v5.8.0 osrm-extract -p /opt/foot.lua /data/los-angeles_california.osm.pbf
docker run --tty --volume `pwd`/data/osrm:/data osrm/osrm-backend:v5.8.0 osrm-contract /data/los-angeles_california.osrm
docker run --detach -v `pwd`/data/osrm:/data -p 45000:5000 osrm/osrm-backend:v5.8.0 osrm-routed /data/los-angeles_california.osrm
```

Then I issue this curl:
```
curl 'http://localhost:45000/route/v1/walking/-117.70203,33.46642;-117.691793,33.460241'
```

The result is a crazy route taking the user over two ferries (I think) and then back down the PCH. The polyline shows the rough route.
```
{
  "code": "Ok",
  "routes": [
    {
      "geometry": "klgkEvtkmUj@|p@yxMdiMq_BxvEalFbcIy_BzuAwzAboKg{DjeIe_DlrEa}IppIor@vhBolAhh@ev@nkNre@fz@v`BcuCr|@RxkBp`AbzgAjsYoySudzBav@_@",
      "legs": [
        {
          "steps": [],
          "distance": 172320.3,
          "duration": 53331.8,
          "summary": "",
          "weight": 53331.8
        }
      ],
      "distance": 172320.3,
      "duration": 53331.8,
      "weight_name": "duration",
      "weight": 53331.8
    }
  ],
  "waypoints": [
    {
      "hint": "JNEJgB02F4CCAAAAfwAAAAAAAABHAQAAggAAAH8AAAAAAAAARwEAAAEAAABnAvz4Xqj-AXIC_Pg0qP4BAAAPAEFVQIU=",
      "name": "Pacific Coast Highway",
      "location": [
        -117.702041,
        33.466462
      ]
    },
    {
      "hint": "HJYOgB-WDoBYAAAAfgAAAP8BAAAAAAAAWAAAAH4AAAD_AQAAAAAAAAEAAAAJK_z47o_-AW8q_PgRkP4BBgAPAEFVQIU=",
      "name": "",
      "location": [
[...]

PR #4242, opened 2017-07-06
handle startpoint in bicycle profile, add startpoint tests for bike, car and foot

emiltin:
# Issue

The bicycle profile is currently not handling startpoints. This means routes can start and stop in the middle of ferries and trains. This PR fixes this by adding the 'handle_startpoint' handler.

Also adds cuke tests for bike, car and foot.

## Tasklist
 - [x] review
 - [x] adjust for comments

### DIFF
--- a/features/bicycle/ferry.feature
+++ b/features/bicycle/ferry.feature
@@ -53,8 +53,7 @@ Feature: Bike - Handle ferry routes
     Scenario: Bike - Ferry duration, multiple nodes
         Given the node map
             """
-            x         y
-              a b c d
+            x a b c d y
             """
 
         And the ways
@@ -64,6 +63,6 @@ Feature: Bike - Handle ferry routes
             | abcd  |         | ferry | yes     | 1:00     |
 
         When I route I should get
-            | from | to | route     | time      |
-            | a    | d  | abcd,abcd | 3600s +-5 |
-            | d    | a  | abcd,abcd | 3600s +-5 |
+            | from | to | route     | time  |
+            | a    | d  | abcd,abcd | 3600s |
+            | d    | a  | abcd,abcd | 3600s |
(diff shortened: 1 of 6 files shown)


profiles/bicycle.lua
```diff
@@ -543,7 +543,9 @@ function way_function (way, result)
 
     -- handle various other flags
     'handle_roundabouts',
-    --'handle_startpoint',
```
emiltin: this was the problem

daniel-j-h: Urgh, good catch! Looks like the foot profile already has the handler enabled.

emiltin: yes also car

features/bicycle/ferry.feature
```diff
@@ -64,6 +64,6 @@ Feature: Bike - Handle ferry routes
             | abcd  |         | ferry | yes     | 1:00     |
 
         When I route I should get
-            | from | to | route     | time  |
-            | a    | d  | abcd,abcd | 3600s |
-            | d    | a  | abcd,abcd | 3600s |
+            | from | to | route     | time      |
+            | a    | d  | abcd,abcd | 3600s +-5 |
```
oxidase: Here is not an accuracy issue, but the expectation change: now turn durations are added to `time`. I think it is better to change the test to
```
     | x    | y  | xa,abcd,yd,yd | 3672s |
     | y    | x  | yd,abcd,xa,xa | 3669.8s |
```

emiltin: ok

emiltin:
i don't think this is related to  #4221, since that is with the foot profile. this PR fixes a bug in the bike profile.

features/bicycle/ferry.feature
emiltin: i changed the geometry so there are no turns, the the result is 3600s.

Issue #4245, opened 2017-07-06
Load graph before opening a socket in the osrm-routed


gardster:
Amazon ELB balancer can't determine osrm-routed start on redeploy.

When we use a tcp health check it determines osrm-routed opened port and sends requests while osrm-routed loads a road graph.
When we use a http health check on a large graph it kills it after 120 seconds of loading data without answer.

Also it looks strange, that port open, but requests will not processed for some time.

TheMarex:
Fixed by issue #4246

PR #4246, opened 2017-07-06
Load graph before creating a server

gardster:
# Issue #4245 

Minimize time when osrm-routed opens network port and can't handle requests.


TheMarex:
Thanks for the fix, this was definitely a bug!

PR #4249, opened 2017-07-07
Remove bool flags from ExternalMemoryNode

TheMarex:
The two `bool` flags in `ExternalMemoryNode` cause an enormous 8 byte (on a 24byte struct) alignment padding. I feel @oxidase pointed this out at some point but we somehow dropped this again.

By de-composing this we are reducing our STXXL usage by 24% (measured on `bayern-latest`).

## Tasklist
 - [x] review
 - [x] adjust for comments


include/util/graph_loader.hpp
```diff
@@ -50,26 +49,26 @@ NodeID loadNodesFromFile(storage::io::FileReader &file_reader,
     coordinates.resize(number_of_nodes);
     osm_node_ids.reserve(number_of_nodes);
 
-    extractor::ExternalMemoryNode current_node;
+    extractor::QueryNode current_node;
     for (NodeID i = 0; i < number_of_nodes; ++i)
```
danpat: Hmm, it's somewhat unrelated, but I wonder if we could get some speedup here by reading blocks of `current_node` and looping over the blocks.  Reading these objects one at a time almost certainly is creating a lot of syscall overhead......

src/extractor/extraction_containers.cpp
```diff
@@ -625,6 +635,40 @@ void ExtractionContainers::WriteNodes(storage::io::FileWriter &file_out) const
...
 
+    {
+        util::UnbufferedLog log;
+        log << "Writing barrier nodes     ... ";
+        TIMER_START(write_nodes);
+        std::vector<NodeID> internal_barrier_nodes;
+        for (const auto id : barrier_nodes)
+        {
+            auto iter = external_to_internal_node_id_map.find(id);
+            if (iter != external_to_internal_node_id_map.end())
```
danpat: Hmm, is this test necessary?  Every `id` should exist in the map, we can probably just `BOOST_ASSERT()` the comparison here and assume `iter->second` is good.

src/extractor/extraction_containers.cpp
```diff
@@ -625,6 +635,40 @@ void ExtractionContainers::WriteNodes(storage::io::FileWriter &file_out) const
...
+    {
+        util::UnbufferedLog log;
+        log << "Writing traffic light nodes     ... ";
+        TIMER_START(write_nodes);
+        std::vector<NodeID> internal_traffic_lights;
+        for (const auto id : traffic_lights)
+        {
+            auto iter = external_to_internal_node_id_map.find(id);
+            if (iter != external_to_internal_node_id_map.end())
+            {
```
danpat: Same as above, this should always be true and it's a bad thing if it's not.

include/util/graph_loader.hpp
```diff
@@ -50,26 +49,26 @@ NodeID loadNodesFromFile(storage::io::FileReader &file_reader,
...
-        }
-
-        if (current_node.traffic_lights)
-        {
-            *traffic_signals = i;
-            ++traffic_signals;
-        }
+    auto num_barriers = file_reader.ReadElementCount64();
+    for (auto index = 0UL; index < num_barriers; ++index)
+    {
```
danpat: As above - might be a bit more performant to read larger blocks.  That said, there probably aren't very many of these in the grand scheme of things, so maybe not worth the extra logic.

danpat:
@TheMarex Basically LGTM.  The only thing that really stands out is that the barrier_nodes and traffic_lights are no longer sorted the same as the QueryNode data.  I don't think it matters though, as that data ends up in an `unordered_set`.

src/extractor/extraction_containers.cpp
TheMarex: Yes. `external_to_internal_node_id_map` only has entries for nodes that we actually use in the road network. However we do parse all nodes.

src/extractor/extraction_containers.cpp
TheMarex: See above.

PR #4251, opened 2017-07-07
Makes tile size checks strict, resolves #4177

### LINKED ISSUE #4177: Stricter vtile test
The test at https://github.com/Project-OSRM/osrm-backend/blob/99b02cedfbfc6fb24f6b07dedd6df37e885690a9/test/nodejs/tile.js#L11 is not strict. We should ideally assert the exact length to be able to detect unexpected changes.

daniel-j-h:
For https://github.com/Project-OSRM/osrm-backend/issues/4177#issuecomment-313541371. Tile size increased because we added more attributes (e.g. in https://github.com/Project-OSRM/osrm-backend/pull/4172 and https://github.com/Project-OSRM/osrm-backend/pull/4162).


test/nodejs/constants.js
```diff
@@ -10,7 +10,7 @@ exports.three_test_coordinates = [[7.41337, 43.72956],
 
 exports.two_test_coordinates = exports.three_test_coordinates.slice(0, 2)
 
-exports.test_tile = {'at': [17059, 11948, 15], 'size': 114000};
+exports.test_tile = {'at': [17059, 11948, 15], 'size': 162422};
```
MoKob: I might be seeing ghosts here, but didn't we have inconsistent tile sizes on different machines once? /cc @danpat

MoKob: never mind. conversation in issue resolves this

PR #4252, opened 2017-07-07
Switch from stxxl::vector to std::vector

oxidase:
# Issue

Looking at the current CPU utilization during extraction for Bavaria
before it is possible to see recent improvements for edge-based graph edges generation :tada:  (now 4 cores are used at the end of the graph instead of 1). But the main bottleneck from seconds 40 to 180 with 0.15 average CPU utilization is still here. The main reason is STXXL pages cache that makes `stxxl::sort` very inefficient wrt to parallelization. 

I propose to switch to `boost::container::vector` ~~that has a mmap-based external storage allocator~~ and piggybacks on kswapd page replacement. With the change extraction time goes down from 234 to 78 seconds after

Open questions:
* is there any significant reason to keep `stxxl::vector` and not use `boost::container::vector` or `std::vector`?
* i don't know how to limit memory usage internally in a process similar to [`stxxl_memory`](https://github.com/Project-OSRM/osrm-backend/blob/243526169e00a7002968537fb7bb6d73bf973429/include/extractor/extraction_containers.hpp#L32-L38) It is pretty important to have memory usage limit to avoid system freezes.

/cc @TheMarex @danpat 

## Tasklist
 - [x] Discuss open questions and reject PR or continue
 - [x] Finish implementation
 - [x] review
 - [ ] adjust for comments

## Requirements / Relations


oxidase:
With 1GB memory limit via CG mechanism CPU utilization looks similar to stxxl
cpu_cg
 and RSS in MB is
rss_cg

TheMarex:
Why `boost::container::vector` instead of `std::vector`? Apart from this we can basically get rid of most of the sort/merge steps in `ExtractionContainers` now thanks to sweet sweet random memory access.
However this might be a good first steps because it's basically for free.

Especially computing the edge weights should be parallelized since this potentially calls into lua (`segment_function`), I think that would get rid of the flat section in the middle.

oxidase:
@TheMarex i thought originally to try different `mmap`-based allocators, and as pointed by @daniel-j-h there is no major difference  between  `boost::container::vector` and `std::vector` with default allocators. Sure it is better to use the default vector implementation. 

Plateau  in the middle is the edge-based graph nodes generation and parts with zero user space CPU usage are disk IO.

This change is quite significant, because it changes STXXL-based caching to system-dependent one and probably must be discussed first with non-Linux OSRM users or may be STXXL usage must be  optional.

But results are really good: for germany-size ~8 minutes extraction with user space CPU
cpu_germany
and RSS in MB
rss_germany

oxidase:
@TheMarex rebased to master
cpu_germany
and
rss_germany

TheMarex:
@oxidase do you know where the peak in the middle is coming from? What is the peak memory consumption with regard to `master`?

oxidase:
@TheMarex slowly going up to 6GB is OSM parsing, the next jump to 7.5GB is `Building node id map      ... ok, after 3.65991s`, and jump to 8.2GB is in [`WriteEdges`](https://github.com/Project-OSRM/osrm-backend/blob/445cc4f7b592623ec838b3b90c448e0e577d3295/src/extractor/extraction_containers.cpp#L558)

danpat:
Hmm, the `external_to_internal_node_map` is an `std::unordered_map` -> we can probably save a bit by using https://github.com/greg7mdp/sparsepp instead.

TheMarex:
This is a first good step I would fell comfortable shipping with `5.9` since it doesn't change much structurally but improves pre-processing times dramatically.

Things we should ticket after merging this:
- Try replacing unordered_map with something more efficient like sorted list of tuples
- Rewrite the `PrepareNodes` `PrepareWays` function since they assume only single-pass iteration which makes them slow
- Use `tbb::parallel_pipeline` for processing each entity

oxidase:
@TheMarex i have removed `external_to_internal_node_id_map`, so now for germany-latest only ~6GB of RAM are needed
rss_germany

src/extractor/extraction_containers.cpp
```diff
@@ -163,6 +199,7 @@ void ExtractionContainers::PrepareData(ScriptingEnvironment &scripting_environme
     PrepareNodes();
     WriteNodes(file_out);
     PrepareEdges(scripting_environment);
+    NodeVector().swap(all_nodes_list);
```
TheMarex:
```
all_nodes_list.clear();
all_nodes_list.shrink_to_fit();
```

Makes the intend a little bit more clear here.

.travis.yml
```diff
@@ -72,7 +72,7 @@ matrix:
       addons: &gcc6
         apt:
           sources: ['ubuntu-toolchain-r-test']
-          packages: ['g++-6', 'libbz2-dev', 'libstxxl-dev', 'libstxxl1', 'libxml2-dev', 'libzip-dev', 'lua5.1', 'liblua5.1-0-dev', 'libtbb-dev', 'libgdal-dev', 'libluabind-dev', 'libboost-all-dev']
+          packages: ['g++-6', 'libbz2-dev', 'libxml2-dev', 'libzip-dev', 'lua5.1', 'liblua5.1-0-dev', 'libtbb-dev', 'libgdal-dev', 'libboost-all-dev']
```
TheMarex: I'm curious why Travis passes with this, since we still have a dependency on STXXL [here](https://github.com/Project-OSRM/osrm-backend/blob/master/src/extractor/edge_based_graph_factory.cpp#L337) see https://github.com/Project-OSRM/osrm-backend/blob/master/include/extractor/turn_data_container.hpp#L117 with https://github.com/Project-OSRM/osrm-backend/blob/master/include/util/vector_view.hpp#L212-L216

include/contractor/graph_contractor.hpp
```diff
@@ -401,7 +409,7 @@ class GraphContractor
     bool Bias(const NodeID a, const NodeID b) const;
 
     std::shared_ptr<ContractorGraph> contractor_graph;
-    stxxl::vector<QueryEdge> external_edge_list;
+    ExternalVector<QueryEdge> external_edge_list;
```
TheMarex: Did you check if this increases the peak memory consumption for `osrm-contract`?

src/extractor/extraction_containers.cpp
danpat: The clear/shrink-to-fit pattern doesn't guarantee that the reserved memory is released.  The swap method is the recommended approach to ensure the memory reserve is released - see http://www.cplusplus.com/reference/vector/vector/clear/

I think a comment here would be good though, this is a c++ smell 😷

.travis.yml
oxidase: `InternalOrExternalVector` does not depend on type and is `std::vector` if STXXL is not found

src/extractor/extraction_containers.cpp
oxidase: this will not compile  because `stxxl::vector` does not support `shrink_to_fit`. 
I have to check this line actually there is an invalid free issue with stxxl and i guess because of that line.

include/contractor/graph_contractor.hpp
oxidase: Not yet, but before merging some numbers of RSS usage for planet size are needed.

oxidase:
@TheMarex now PR removes STXXL usage from the first part of extraction and makes STXXL usage optional in EBGF and contractor. The optional STXXL usage by default is enabled, so PR preserves previous memory usage.

PR #4255, opened 2017-07-07
Via way restrictions

### LINKED ISSUE #2681: Turn-Restrictions with via ways
Outcome of discussion with @MoKob:

Tagging scheme of osm restrictions using ways http://wiki.openstreetmap.org/wiki/Relation:restriction#Prohibitory_restriction
OSRM does not handle restrictions ways with via role and
ignores via.way restrictions in RestrictionParser parser
at https://github.com/Project-OSRM/osrm-backend/blob/master/src/extractor/restriction_parser.cpp#L168-L172

For example, in germany-latest there are 1404 such relations with 1331 (95%) one-via-way relations
[relations.txt](https://github.com/Project-OSRM/osrm-backend/files/375808/relations.txt)

Possible approaches and their respective drawbacks:
 1) modify graph creation
    - hidden turns within the new edges
    - parallel edges -> new edges non start
 2) additional contraction
    - contract relation vertices first (breaks contraction order)
    - pass restrictions along to the contraction
    - only allows to handle simple isolated intersections first

For future work:
- longer restrictions (more than one via ways)
- non-isolated restrictions that can have overlapping ways

Related issues:
#483 relation http://www.openstreetmap.org/relation/2339761 has a via node restriction and will not be affected by the change
#2634 https://www.openstreetmap.org/relation/6187972 has one via way restriction and must be fixed with the first step
#1828 looks like an implementation of the first approach

/cc @MoKob

MoKob:
# Issue

 - refactors turn restrictions to follow strict types/structures.
 - adds support for via-way turn restrictions
 - first step towards #2681

## Tasklist
 - [x] add changelog entry indicating breaking changes to data-files
 - [x] review
 - [x] adjust for comments

## Performance:

Extraction of Germany:

```
	Command being timed: "./build/osrm-extract /data/europe/germany/germany-latest.osm.pbf"
	User time (seconds): 1820.21
	System time (seconds): 120.63
	Percent of CPU this job got: 289%
	Elapsed (wall clock) time (h:mm:ss or m:ss): 11:09.29
	Average shared text size (kbytes): 0
	Average unshared data size (kbytes): 0
	Average stack size (kbytes): 0
	Average total size (kbytes): 0
	Maximum resident set size (kbytes): 29249273856
	Average resident set size (kbytes): 0
	Major (requiring I/O) page faults: 558
	Minor (reclaiming a frame) page faults: 7944112
	Voluntary context switches: 1733
	Involuntary context switches: 6452780
```

Contraction:
```
	Command being timed: "./build/osrm-contract /data/europe/germany/germany-latest.osrm"
	User time (seconds): 5743.62
	System time (seconds): 61.59
	Percent of CPU this job got: 350%
	Elapsed (wall clock) time (h:mm:ss or m:ss): 27:37.11
	Average shared text size (kbytes): 0
	Average unshared data size (kbytes): 0
	Average stack size (kbytes): 0
	Average total size (kbytes): 0
	Maximum resident set size (kbytes): 12026462208
	Average resident set size (kbytes): 0
	Major (requiring I/O) page faults: 237
	Minor (reclaiming a frame) page faults: 2390626
	Voluntary context switches: 2106
	Involuntary context switches: 3609162
```

Partition
```
[info] RAM: peak bytes used: 2839437312
	Command being timed: "./build/osrm-partition /data/europe/germany/germany-latest.osrm"
	User time (seconds): 2416.88
	System time (seconds): 48.81
	Percent of CPU this job got: 312%
	Elapsed (wall clock) time (h:mm:ss or m:ss): 13:09.87
	Average shared text size (kbytes): 0
	Average unshared data size (kbytes): 0
	Average stack size (kbytes): 0
	Average total size (kbytes): 0
	Maximum resident set size (kbytes): 11357749248
	Average resident set size (kbytes): 0
	Major (requiring I/O) page faults: 133
	Minor (reclaiming a frame) page faults: 3211197
	Voluntary context switches: 33
	Involuntary context switches: 2509986
```

Customize

```
	Command being timed: "./build/osrm-customize /data/europe/germany/germany-latest.osrm"
	User time (seconds): 92.72
	System time (seconds): 4.02
	Percent of CPU this job got: 267%
	Elapsed (wall clock) time (h:mm:ss or m:ss): 0:36.15
	Average shared text size (kbytes): 0
	Average unshared data size (kbytes): 0
	Average stack size (kbytes): 0
	Average total size (kbytes): 0
	Maximum resident set size (kbytes): 11178278912
	Average resident set size (kbytes): 0
	Major (requiring I/O) page faults: 153
	Minor (reclaiming a frame) page faults: 1078677
	Voluntary context switches: 18
	Involuntary context switches: 68336
```

## Master

Extraction
```
	Command being timed: "./build/osrm-extract /data/europe/germany/germany-latest.osm.pbf"
	User time (seconds): 1803.88
	System time (seconds): 127.52
	Percent of CPU this job got: 285%
	Elapsed (wall clock) time (h:mm:ss or m:ss): 11:16.82
	Average shared text size (kbytes): 0
	Average unshared data size (kbytes): 0
	Average stack size (kbytes): 0
	Average total size (kbytes): 0
	Maximum resident set size (kbytes): 29302489088
	Average resident set size (kbytes): 0
	Major (requiring I/O) page faults: 525
	Minor (reclaiming a frame) page faults: 9716627
	Voluntary context switches: 2248
	Involuntary context switches: 4430620
```

Contraction
```
	Command being timed: "./build/osrm-contract /data/europe/germany/germany-latest.osrm"
	User time (seconds): 5772.61
	System time (seconds): 68.48
	Percent of CPU this job got: 353%
	Elapsed (wall clock) time (h:mm:ss or m:ss): 27:31.05
	Average shared text size (kbytes): 0
	Average unshared data size (kbytes): 0
	Average stack size (kbytes): 0
	Average total size (kbytes): 0
	Maximum resident set size (kbytes): 12026134528
	Average resident set size (kbytes): 0
	Major (requiring I/O) page faults: 232
	Minor (reclaiming a frame) page faults: 2396365
	Voluntary context switches: 3366
	Involuntary context switches: 4783880
```

Partition
```
	Command being timed: "./build/osrm-partition /data/europe/germany/germany-latest.osrm"
	User time (seconds): 2257.08
	System time (seconds): 46.40
	Percent of CPU this job got: 332%
	Elapsed (wall clock) time (h:mm:ss or m:ss): 11:32.09
	Average shared text size (kbytes): 0
	Average unshared data size (kbytes): 0
	Average stack size (kbytes): 0
	Average total size (kbytes): 0
	Maximum resident set size (kbytes): 11366449152
	Average resident set size (kbytes): 0
	Major (requiring I/O) page faults: 133
	Minor (reclaiming a frame) page faults: 3355731
	Voluntary context switches: 33
	Involuntary context switches: 1645742
```

Customize

```
	Command being timed: "./build/osrm-customize /data/europe/germany/germany-latest.osrm"
	User time (seconds): 98.63
	System time (seconds): 5.48
	Percent of CPU this job got: 242%
	Elapsed (wall clock) time (h:mm:ss or m:ss): 0:42.99
	Average shared text size (kbytes): 0
	Average unshared data size (kbytes): 0
	Average stack size (kbytes): 0
	Average total size (kbytes): 0
	Maximum resident set size (kbytes): 11177803776
	Average resident set size (kbytes): 0
	Major (requiring I/O) page faults: 155
	Minor (reclaiming a frame) page faults: 1078662
	Voluntary context switches: 15
	Involuntary context switches: 172589
```

### DIFF
--- a/features/car/restrictions.feature
+++ b/features/car/restrictions.feature
@@ -537,7 +537,7 @@ Feature: Car - Turn restrictions
             | from | to | route               |
             | a    | h  | abc,cde,efc,cgh,cgh |
 
-    @restriction
+    @restriction-way
     Scenario: Car - prohibit turn
         Given the node map
             """
@@ -563,9 +563,11 @@ Feature: Car - Turn restrictions
             | restriction | ab       | be      | de     | no_right_turn |
 
         When I route I should get
-            | from | to | route       |
-            | a    | d  | ab,be,de,de |
-            | a    | f  | ab,be,ef,ef |
+            | from | to | route             | turns                                                               | locations   |
+            | a    | d  | ab,be,ef,ef,de,de | depart,turn right,turn left,continue uturn,new name straight,arrive | a,b,e,f,e,d |
+            | a    | f  | ab,be,ef,ef       | depart,turn right,turn left,arrive                                  | a,b,e,f     |
+            | c    | d  | bc,be,de,de       | depart,turn left,turn right,arrive                                  | c,b,e,d     |
+            | c    | f  | bc,be,ef,ef       | depart,turn left,turn left,arrive                                   | c,b,e,f     |
 
     @restriction @overlap
     Scenario: Car - prohibit turn
@@ -597,11 +599,11 @@ Feature: Car - Turn restrictions
             | restriction | bc       | be      | ef     | no_left_turn  |
 
         When I route I should get
-            | from | to | route       |
-            | a    | d  | ab,be,de,de |
-            | a    | f  | ab,be,ef,ef |
-            | c    | d  | bc,be,de,de |
-            | c    | f  | bc,be,ef,ef |
+            | from | to | route             |
+            | a    | d  | ab,be,ef,ef,de,de |
+            | a    | f  | ab,be,ef,ef       |
+            | c    | d  | bc,be,de,de       |
+            | c    | f  | bc,be,de,de,ef,ef |
 
     @restriction-way @overlap
     Scenario: Two times same way
@@ -618,9 +620,25 @@ Feature: Car - Turn restrictions
                 |   |
                 |   |
                 |   |
-            a - b - c - f
-                |   | \ |
-                i - d - e
+                |   |
+                |   |
+                |   |
+                |   |
+                |   |
+                |   |
+                |   |
+                |   |
+                |   |
+                |   |
+                |   |
+                |   |
+                |   |
+                |   |
+                |   |
+                |   |
+            a - b - c - - - - - - - - - - - - - - - - - - - f
+                |   | \                                     /
+                i - d - e - - - - - - - - - - - - - - - - -
             """
 
         And the ways
@@ -641,7 +659,7 @@ Feature: Car - Turn restrictions
 
        When I route I should get
             | from | to | route                |
-            | a    | i  | ab,bc,cd,fedib,fedib |
+            | a    | i  | ab,bc,cf,fedib,fedib |
 
 
     @restriction-way @overlap
@@ -682,8 +700,8 @@ Feature: Car - Turn restrictions
 
         When I route I should get
             | from | to | route                  |
-            | a    | j  | left,third,right,right |
-            | f    | e  | right,first,left,left  |
+            | a    | j  | left,first,right,right |
+            | f    | e  | right,third,left,left  |
 
     @restriction
     Scenario: Car - allow only turn
@@ -711,9 +729,11 @@ Feature: Car - Turn restrictions
             | restriction | ab       | be      | ef     | only_left_on |
 
         When I route I should get
-            | from | to | route       |
-            | a    | d  | ab,be,de,de |
-
+            | from | to | route       | turns | locations |
+            | a    | d  | ab,be,ef,ef,de,de | depart,turn right,turn left,continue uturn,new name straight,arrive | a,b,e,f,e,d |
+            | a    | f  | ab,be,ef,ef       | depart,turn right,turn left,arrive                                  | a,b,e,f     |
+            | c    | d  | bc,be,de,de       | depart,turn left,turn right,arrive                                  | c,b,e,d     |
+            | c    | f  | bc,be,ef,ef       | depart,turn left,turn left,arrive                                   | c,b,e,f     |
 
     @restriction
     Scenario: Car - allow only turn
@@ -810,14 +830,15 @@ Feature: Car - Turn restrictions
             | restriction | gf       | fb,bc   | cd     | only_u_turn    |
 
         When I route I should get
-            | from | to | route             |
-            | a    | d  | abcd,abcd         |
-            | a    | e  | abcd,ce,ce        |
-            | a    | f  | abcd,hfb,hfb      |
-            | g    | e  | gf,hfb,abcd,ce,ce |
-            | g    | d  | gf,hfb,abcd,abcd  |
-            | h    | e  | hfb,abcd,ce,ce    |
-            | h    | d  | hfb,abcd,abcd     |
+            | from | to | route                | turns                                            | locations |
+            | a    | d  | abcd,ce,ce,abcd,abcd | depart,turn left,continue uturn,turn left,arrive | a,c,e,c,d |
+            | a    | e  | abcd,ce,ce           | depart,turn left,arrive                          | a,c,e     |
+            | a    | f  | abcd,hfb,hfb         | depart,turn right,arrive                         | a,b,f     |
+            | g    | e  | gf,hfb,abcd,ce,ce    | depart,turn right,turn right,turn left,arrive    | g,f,b,c,e |
+            | g    | d  | gf,hfb,abcd,abcd     | depart,turn right,turn right,arrive              | g,f,b,d   |
+            | h    | e  | hfb,abcd,ce,ce       | depart,end of road right,turn left,arrive        | h,b,c,e   |
+            | h    | d  | hfb,abcd,abcd        | depart,end of road right,arrive                  | h,b,d     |
+
 
     @restriction
     Scenario: Car - prohibit turn, traffic lights
@@ -855,7 +876,8 @@ Feature: Car - Turn restrictions
 
 
         When I route I should get
-            | from | to | route       |
-            | h    | j  | ab,be,de,de |
-            | h    | f  | ab,be,ef,ef |
-
+            | from | to | route             | turns                                                               | locations   |
+            | a    | d  | ab,be,ef,ef,de,de | depart,turn right,turn left,continue uturn,new name straight,arrive | a,b,e,f,e,d |
+            | a    | f  | ab,be,ef,ef       | depart,turn right,turn left,arrive                                  | a,b,e,f     |
+            | c    | d  | bc,be,de,de       | depart,turn left,turn right,arrive                                  | c,b,e,d     |
+            | c    | f  | bc,be,ef,ef       | depart,turn left,turn left,arrive                                   | c,b,e,f     |
--- a/features/car/restrictions.feature
+++ b/features/car/restrictions.feature
@@ -582,7 +582,6 @@ Feature: Car - Turn restrictions
             |   d
             |
             a
-
             """
 
         And the ways
@@ -881,3 +880,112 @@ Feature: Car - Turn restrictions
             | a    | f  | ab,be,ef,ef       | depart,turn right,turn left,arrive                                  | a,b,e,f     |
             | c    | d  | bc,be,de,de       | depart,turn left,turn right,arrive                                  | c,b,e,d     |
             | c    | f  | bc,be,ef,ef       | depart,turn left,turn left,arrive                                   | c,b,e,f     |
+
+
+      @restriction @overlap @geometry
+      Scenario: Geometry
+        Given the node map
+            """
+            c
+            |
+            |   f
+            |   |
+            b-g-e
+            |   |
+            |   d
+            |
+            a
+            """
+
+        And the ways
+            | nodes |
+            | ab    |
+            | bc    |
+            | bge   |
+            | de    |
+            | ef    |
+
+        And the relations
+            | type        | way:from | way:via | way:to | restriction   |
+            | restriction | ab       | bge     | de     | no_right_turn |
+            | restriction | bc       | bge     | ef     | no_left_turn  |
+
+        When I route I should get
+            | from | to | route              |
+            | a    | d  | ab,bge,ef,ef,de,de |
+            | a    | f  | ab,bge,ef,ef       |
+            | c    | d  | bc,bge,de,de       |
+            | c    | f  | bc,bge,de,de,ef,ef |
+
+      @restriction @overlap @geometry @traffic-signals
+      Scenario: Geometry
+        Given the node map
+            """
+            c
+            |
+            |   f
+            |   |
+            b-g-e
+            |   |
+            |   d
+            |
+            a
+            """
+
+        And the ways
+            | nodes |
+            | ab    |
+            | bc    |
+            | bge   |
+            | de    |
+            | ef    |
+
+        And the nodes
+            | node | highway         |
+            | g    | traffic_signals |
+
+        And the relations
+            | type        | way:from | way:via | way:to | restriction   |
+            | restriction | ab       | bge     | de     | no_right_turn |
+            | restriction | bc       | bge     | ef     | no_left_turn  |
+
+        # this case is currently not handling the via-way restrictions and we need support for looking across traffic signals.
+        # It is mainly included to show limitations and to prove that we don't crash hard here
+        When I route I should get
+            | from | to | route        |
+            | a    | d  | ab,bge,de,de |
+            | a    | f  | ab,bge,ef,ef |
+            | c    | d  | bc,bge,de,de |
+            | c    | f  | bc,bge,ef,ef |
+
+      # don't crash hard on invalid restrictions
+      @restriction @invalid
+      Scenario: Geometry
+        Given the node map
+            """
+            c
+            |
+            |   f
+            |   |
+            b---e
+            |   |
+            |   d
+            |
+            a
+            """
+
+        And the ways
+            | nodes | oneway |
+            | ab    |        |
+            | bc    |        |
+            | be    | yes    |
+            | de    |        |
+            | ef    |        |
+
+        And the relations
+            | type        | way:from | way:via | way:to | restriction   |
+            | restriction | de       | be      | ab     | no_left_turn  |
+
+        When I route I should get
+            | from | to | route       |
+            | a    | f  | ab,be,ef,ef |
--- a/features/car/restrictions.feature
+++ b/features/car/restrictions.feature
@@ -506,3 +506,356 @@ Feature: Car - Turn restrictions
             | s    | n  | sj,nj,nj |
             | s    | e  | sj,ej,ej |
 
+    @restriction @compression
+    Scenario: Restriction On Compressed Geometry
+        Given the node map
+            """
+                        i
+                        |
+                    f - e
+                    |   |
+            a - b - c - d
+                    |
+                    g
+                    |
+                    h
+            """
+
+        And the ways
+            | nodes |
+            | abc   |
+            | cde   |
+            | efc   |
+            | cgh   |
+            | ei    |
+
+        And the relations
+            | type        | way:from | node:via | way:to | restriction   |
+            | restriction | abc      | c        | cgh    | no_right_turn |
+
+        When I route I should get
+            | from | to | route               |
+            | a    | h  | abc,cde,efc,cgh,cgh |
+
+    @restriction
+    Scenario: Car - prohibit turn
+        Given the node map
+            """
+            c
+            |
+            |   f
+            |   |
+            b---e
+            |   |
+            a   d
+            """
+
+        And the ways
+            | nodes |
+            | ab    |
+            | bc    |
+            | be    |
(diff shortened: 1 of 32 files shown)


features/car/restrictions.feature
```diff
@@ -506,3 +506,135 @@ Feature: Car - Turn restrictions
...
+        And the ways
+            | nodes |
+            | ab    |
+            | bc    |
+            | be    |
+            | de    |
+            | ef    |
+
+        And the relations
+            | type        | way:from | way:via | way:from | restriction   |
```
daniel-j-h: Here and below: shouldn't it be `way:to` instead of `way:from` twice?

include/extractor/restriction.hpp
```diff
@@ -5,111 +5,267 @@
...
+    boost::variant<InputWayRestriction, InputNodeRestriction> node_or_way;
+    restriction_details::Bits flags;
+
+    OSMEdgeID_weak from() const
     {
-        via.node = node;
-        from.node = SPECIAL_NODEID;
-        to.node = SPECIAL_NODEID;
+        return flags.way_type ? boost::get<InputWayRestriction>(node_or_way).from
+                              : boost::get<InputNodeRestriction>(node_or_way).from;
```
daniel-j-h: I don't understand this construct - above we're using a variant but we're also storing flags? The variant already contains a flag and knows what actual type it is - I think you can use a visitor or a checked get call here.

include/extractor/restriction.hpp
```diff
@@ -5,111 +5,267 @@
...
+};
+
+// Wrapper for turn restrictions that gives more information on its type / handles the switch
+// between node/way/multi-way restrictions
+struct TurnRestriction
 {
-    using value_type = InputRestrictionContainer;
-    bool operator()(const InputRestrictionContainer &a, const InputRestrictionContainer &b) const
+    boost::variant<NodeRestriction, WayRestriction> node_or_way;
+    restriction_details::Bits flags;
```
daniel-j-h: Same question here

features/car/restrictions.feature
MoKob: Of course.

include/extractor/restriction.hpp
MoKob: This is an open change I have to do. I was originally using the normal structs that were in the data and only then switched to a true variant to avoid reinterpreting data over multiple types. Need to update that.

MoKob:
Demo Server:

screen shot 2017-07-13 at 15 19 57 2

This PR:

screen shot 2017-07-13 at 15 19 59 2

Relation in question: https://www.openstreetmap.org/relation/6495959#map=19/37.77034/-122.51128

MoKob:
Impact so far on Baden Wuerttemberg:

extraction stays more or less the same speed:

without:
real	1m36.738s
user	4m12.325s
sys	0m17.995s

with:
real	1m35.268s
user	4m18.018s
sys	0m17.866s

Contraction measurements fluctuate quite a bit but do not show a consistent slowdown over current master branch.

src/extractor/restriction_compressor.cpp
```diff
@@ -0,0 +1,104 @@
...
+
+void RestrictionCompressor::Compress(const NodeID from, const NodeID via, const NodeID to)
+{
+    const auto get_value = [](const auto pair) { return pair.second; };
+
+    // extract all head ptrs and move them from via to from.
+    auto all_heads_range = heads.equal_range(via);
+    std::vector<NodeRestriction *> head_ptrs;
+    std::transform(
+        all_heads_range.first, all_heads_range.second, std::back_inserter(head_ptrs), get_value);
```
TheMarex: Lambda `get_value` should be inlined here and `boost::transform` can work directly on a range like that.

src/extractor/restriction_compressor.cpp
```diff
@@ -0,0 +1,104 @@
...
+        }
+        // ____ | to - p.from | via - p.via | from - p.to | ____
+        else
+        {
+            BOOST_ASSERT(ptr->via == from);
+            ptr->from = to;
+        }
+    };
+
+    std::for_each(head_ptrs.begin(), head_ptrs.end(), update_head);
```
TheMarex: Simple for loop would be great.

src/extractor/restriction_compressor.cpp
```diff
@@ -0,0 +1,104 @@
...
+namespace osrm
+{
+namespace extractor
+{
+
+RestrictionCompressor::RestrictionCompressor(std::vector<TurnRestriction> &restrictions)
+{
+    // add a node restriction ptr to the heads/tails maps, needs to be a reference!
+    auto index = [&](auto &element) {
+        heads.insert(std::make_pair(element.from, &element));
```
TheMarex: Terminology wise, the `head` is typically the target of an edge, here it is the `from` of the restriction. Maybe `restriction_starts` and `restrictions_targets` would be more obvious?

src/extractor/restriction_compressor.cpp
```diff
@@ -0,0 +1,104 @@
...
+        }
+    };
+
+    std::for_each(head_ptrs.begin(), head_ptrs.end(), update_head);
+
+    const auto reinsert_head = [&](auto ptr) { heads.insert(std::make_pair(ptr->from, ptr)); };
+
+    // update the ptrs in our mapping
+    heads.erase(via);
+    std::for_each(head_ptrs.begin(), head_ptrs.end(), reinsert_head);
```
TheMarex: This would again me much easier to read without having to find `reinsert_head` again and just iterating with a `for` loop.

src/extractor/restriction_filter.cpp
```diff
@@ -0,0 +1,88 @@
...
+            return !is_valid_node(restriction.AsNodeRestriction());
+        }
+        else
+        {
+            BOOST_ASSERT(restriction.Type() == RestrictionType::WAY_RESTRICTION);
+            return !is_valid_way(restriction.AsWayRestriction());
+        }
+    };
+
+    restrictions.erase(std::remove_if(restrictions.begin(), restrictions.end(), is_invalid),
```
TheMarex: Decomposing these two functions by storing the new end returned by `remove_if` would make this less arcane looking. I would recommend to inline the `is_invalid` in the `remove_if` as well.

src/extractor/scripting_environment_lua.cpp
```diff
@@ -645,14 +644,15 @@ void Sol2ScriptingEnvironment::ProcessElements(
                 static_cast<const osmium::Way &>(*entity), std::move(result_way)));
             break;
         case osmium::item_type::relation:
-            result_res.clear();
-            result_res =
+        {
+            auto result_res =
```
TheMarex: Good catch, I wonder why `result_res` was a vector in the first place. 👍

src/extractor/way_restriction_map.cpp
```diff
@@ -0,0 +1,199 @@
...
+
+            // so far we can only handle restrictions that are not interrupted
+            if (way.in_restriction.via == way.out_restriction.from &&
+                way.in_restriction.to == way.out_restriction.via)
+                restriction_data.push_back(turn_restriction);
+        }
+    };
+    std::for_each(turn_restrictions.begin(), turn_restrictions.end(), extract_restrictions);
+
+    const auto as_duplicated_node =
```
TheMarex: If this function gets inlined into `by_duplicated_node` we could simply use `std::tie(way.in_restriction.via, way.out_restriction.via, way.in_restriction.from)` and avoid the copy over the tuple.

src/extractor/way_restriction_map.cpp
```diff
@@ -0,0 +1,199 @@
...
+    std::size_t index = 0, duplication_id = 0;
+    // map all way restrictions into access containers
+    const auto prepare_way_restriction = [this, &index, &duplication_id, as_duplicated_node](
+        const auto &restriction) {
+        const auto &way = restriction.AsWayRestriction();
+        restriction_starts.insert(
+            std::make_pair(std::make_pair(way.in_restriction.from, way.in_restriction.via), index));
+        ++index;
+    };
+    std::for_each(restriction_data.begin(), restriction_data.end(), prepare_way_restriction);
```
TheMarex: This should be a for loop for sure as it keep external state. Also `as_duplicate_node` is bound but not used, and I don't see a reference to `duplication_id`.

src/extractor/way_restriction_map.cpp
```diff
@@ -0,0 +1,199 @@
...
+
+    auto const add_offset_on_new_groups = [&](auto const &lhs, auto const &rhs) {
+        BOOST_ASSERT(rhs == restriction_data[offset]);
+        // add a new lower bound for rhs
+        if (as_duplicated_node(lhs) != as_duplicated_node(rhs))
+            duplicated_node_groups.push_back(offset);
+        ++offset;
+        return false; // continue until the end
+    };
+    std::adjacent_find(restriction_data.begin(), restriction_data.end(), add_offset_on_new_groups);
```
TheMarex: We have `util::for_each_pair` for this, which makes the intend a little bit clearer.

src/extractor/way_restriction_map.cpp
```diff
@@ -0,0 +1,199 @@
...
+
+    const auto &way = itr->AsWayRestriction();
+
+    return way.out_restriction.from == from && way.out_restriction.via == to;
+}
+
+std::size_t WayRestrictionMap::AsDuplicatedNodeID(const std::size_t restriction_id) const
+{
+    return std::distance(duplicated_node_groups.begin(),
+                         std::upper_bound(duplicated_node_groups.begin(),
```
TheMarex: This should be split to two expression for readability, the `1` is almost sneaking away at the end.

src/extractor/way_restriction_map.cpp
```diff
@@ -0,0 +1,199 @@
...
+    return util::irange<std::size_t>(
+        AsDuplicatedNodeID(as_restriction_id(duplicated_node_range_itr.first)),
+        AsDuplicatedNodeID(as_restriction_id(duplicated_node_range_itr.second)));
+}
+
+bool WayRestrictionMap::IsRestricted(std::size_t duplicated_node, const NodeID to) const
+{
+    // loop over all restrictions associated with the node. Mark as restricted based on
+    // is_only/restricted targets
+    for (std::size_t restriction_index = duplicated_node_groups[duplicated_node];
```
TheMarex: `util::irange` can be used for this.

features/car/restrictions.feature
```diff
@@ -506,3 +506,486 @@ Feature: Car - Turn restrictions
...
+                |   |
+                |   |
+                |   |
+                |   |
+                |   |
+                |   |
+            a - b - c - - - - - - - - - - - - - - - - - - - f
+                |   | \                                     /
+                i - d - e - - - - - - - - - - - - - - - - -
+            """
```
MoKob: The correct turn needs to be undesirable, even without the turn restrictions. Otherwise having multiple non-working restrictions could be hidden by the fact that the non-restricted path is better, based on turn penalties alone.

features/car/restrictions.feature
```diff
@@ -506,3 +506,486 @@ Feature: Car - Turn restrictions
...
+            | ef    | ef   |
+
+        And the relations
+            | type        | way:from | way:via | way:to | restriction   |
+            | restriction | hgab     | be      | jide   | no_right_turn |
+
+        And the nodes
+            | node | highway         |
+            | g    | traffic_signals |
+            | i    | traffic_signals |
```
MoKob: Barriers are still not collapsed, so it wouldn't make a different.

features/car/restrictions.feature
```diff
@@ -506,3 +506,486 @@ Feature: Car - Turn restrictions
...
+            | node | highway         |
+            | g    | traffic_signals |
+
+        And the relations
+            | type        | way:from | way:via | way:to | restriction   |
+            | restriction | ab       | bge     | de     | no_right_turn |
+            | restriction | bc       | bge     | ef     | no_left_turn  |
+
+        # this case is currently not handling the via-way restrictions and we need support for looking across traffic signals.
+        # It is mainly included to show limitations and to prove that we don't crash hard here
```
MoKob: This is already handled in #4327

include/extractor/restriction_compressor.hpp
```diff
@@ -0,0 +1,46 @@
...
+
+    // account for the compression of `from-via-to` into `from-to`
+    void Compress(const NodeID from, const NodeID via, const NodeID to);
+
+  private:
+    // a turn restriction is given as `from head via node to tail`. Edges ending at `head` being
+    // contracted move the head pointer to their respective head. Edges starting at tail move the
+    // tail values to their respective tails. Way turn restrictions are represented by two
+    // node-restrictions, so we can focus on them alone
+    boost::unordered_multimap<NodeID, NodeRestriction *> heads;
```
MoKob: https://stackoverflow.com/questions/11519648/unordered-multimapequal-range-slow

features/car/restrictions.feature
```diff
@@ -506,3 +506,486 @@ Feature: Car - Turn restrictions
...
+            | nodes |
+            | abc   |
+            | cde   |
+            | efc   |
+            | cgh   |
+            | ei    |
+
+        And the relations
+            | type        | way:from | node:via | way:to | restriction   |
+            | restriction | abc      | c        | cgh    | no_right_turn |
```
MoKob: We should just ignore them. I would consider these a mapping error to begin with and not worry about them.
Internally

`a - b - c` and `ab` via `b` to `bc` should turn out as `ac` via `b` to `ac` after compression and be ignored as inconsistent.

features/car/restrictions.feature
```diff
@@ -506,3 +506,486 @@ Feature: Car - Turn restrictions
...
+            | nodes |
+            | ab    |
+            | bc    |
+            | be    |
+            | de    |
+            | ef    |
+
+        And the relations
+            | type        | way:from | way:via | way:to | restriction  |
+            | restriction | ab       | be      | ef     | only_left_on |
```
MoKob: I am not sure whether I follow, but yes. We evaluate `only_*` without checking the `postfix`. Just as we did for via-nodes.

include/extractor/edge_based_graph_factory.hpp
```diff
@@ -152,14 +157,16 @@ class EdgeBasedGraphFactory
 
     unsigned RenumberEdges();
 
-    std::vector<NBGToEBG> GenerateEdgeExpandedNodes();
+    std::vector<NBGToEBG> GenerateEdgeExpandedNodes(const WayRestrictionMap &way_restriction_map);
```
MoKob: Will document here. You are correct. EdgeExpandedNodes has to generate all our duplicated nodes as well. Therefor it requires access to the way-restriction-map.

include/extractor/edge_based_graph_factory.hpp
```diff
@@ -152,14 +157,16 @@ class EdgeBasedGraphFactory
...
 
     void GenerateEdgeExpandedEdges(ScriptingEnvironment &scripting_environment,
                                    const std::string &original_edge_data_filename,
                                    const std::string &turn_lane_data_filename,
                                    const std::string &turn_weight_penalties_filename,
                                    const std::string &turn_duration_penalties_filename,
-                                   const std::string &turn_penalties_index_filename);
+                                   const std::string &turn_penalties_index_filename,
+                                   const RestrictionMap &restriction_map,
+                                   const WayRestrictionMap &way_restriction_map);
```
MoKob: Before they were stored as members and used within. We require the maps to check which turns are allowed. The intersection-generation / handling of via-way restrictions is required when looking at the edge-based-edges to find out which of the turns are valid (`entry_allowed` flag).

I just moved this to a reference parameter to get rid of some pointer usage we did beforehand.

include/extractor/extraction_containers.hpp
```diff
@@ -49,9 +47,15 @@ class ExtractionContainers
     NameCharData name_char_data;
     NameOffsets name_offsets;
     // an adjacency array containing all turn lane masks
-    RestrictionsVector restrictions_list;
     WayIDStartEndVector way_start_end_id_list;
+
     unsigned max_internal_node_id;
+
+    // list of restrictions before we transform them into the output types
+    std::vector<InputConditionalTurnRestriction> restrictions_list;
```
MoKob: Input restrictions are generated during parsing and store OSM IDs. Only after we performed our re-mapping, we can assign the actual restrictions to be used later on.

include/extractor/serialization.hpp
```diff
@@ -135,41 +135,67 @@ inline void write(storage::io::FileWriter &writer,
...
+    else
+    {
+        BOOST_ASSERT(restriction.Type() == RestrictionType::NODE_RESTRICTION);
+        write(writer, boost::get<NodeRestriction>(restriction.node_or_way));
+    }
+}
+
+inline void write(storage::io::FileWriter &writer, const ConditionalTurnRestriction &restriction)
+{
+    write(writer, static_cast<TurnRestriction>(restriction));
```
MoKob: Without the down-cast, we would end up in recursion, since `write` is available for the `ConditionalTurnRestriction` as well, wouldn't we?

TheMarex: Good point, I missed that! But maybe `static_cast<const TurnRestriction&>(restriction)` would avoid the copy.

include/extractor/restriction_compressor.hpp
daniel-j-h: I think it was https://bugs.llvm.org/show_bug.cgi?id=21275

MoKob:
I've addressed the reviews. Could you re-check?

PR #4256, opened 2017-07-07
Fixes Node.js bindings always enabling alternatives when using boolean overload

daniel-j-h:
When using the Node.js bindings and using the boolean overload for requesting alternatives (`alternatives: true`) we set the boolean libosrm RouteParam parameter to the user specified value and fix the number parameter to one:

https://github.com/Project-OSRM/osrm-backend/blob/5ede5577d169efee0a6bc78fb15c2de406de74bf/include/nodejs/node_osrm_support.hpp#L723-L724

In https://github.com/Project-OSRM/osrm-backend/blob/master/src/engine/plugins/viaroute.cpp#L101-L106 we try to be smart in case the user gets the boolean and the number value out of sync.

Turns out the combination of both is a bad idea.


daniel-j-h:
Cherry-picked into 5.9.

PR #4258, opened 2017-07-08
Discards construction and proposed ways, resolves #4230

### LINKED ISSUE #4230: Car profile should discard construction highways
We check for `avoid.construction` here

https://github.com/Project-OSRM/osrm-backend/blob/master/profiles/lib/handlers.lua#L475

but do not have it in the car profile's avoid sequence here

https://github.com/Project-OSRM/osrm-backend/blob/master/profiles/car.lua#L107-L114

we have it in the bike profile, though.

daniel-j-h:
For #4230: discards `highway=construction`, `highway=proposed`, `construction=`, `proposed=`.

http://wiki.openstreetmap.org/wiki/Key:construction
http://wiki.openstreetmap.org/wiki/Key:proposed
http://wiki.openstreetmap.org/wiki/Key:highway

I see no reason we ever want to route over these tags. Please correct me if I'm wrong or if there are any edge cases I haven't thought of. Better to be safe than sorry here.

cc @emiltin @1ec5

### DIFF
--- a/features/car/construction.feature
+++ b/features/car/construction.feature
@@ -1,0 +1,16 @@
+@routing @car @construction
+Feature: Car - all construction tags the OpenStreetMap community could think of and then some
+
+    Background:
+        Given the profile "car"
+
+    Scenario: Various ways to tag construction and proposed roads
+        Then routability should be
+            | highway      | construction | proposed | bothw |
+            | primary      |              |          | x     |
+            | construction |              |          |       |
+            | proposed     |              |          |       |
+            | primary      |          yes |          |       |
+            | primary      |              |     yes  |       |
+            | primary      |           no |          | x     |
+            | primary      |     widening |          | x     |
--- a/profiles/bicycle.lua
+++ b/profiles/bicycle.lua
@@ -189,7 +189,8 @@ local profile = {
 
   avoid = Set {
     'impassable',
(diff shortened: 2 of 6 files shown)


emiltin:
only thing i can think of is construction=no... is that a thing?

features/car/construction.feature
```diff
@@ -0,0 +1,14 @@
...
+    Background:
+        Given the profile "car"
+
+    Scenario: Various ways to tag construction and proposed roads
+        Then routability should be
+            | highway      | construction | proposed | bothw |
+            | primary      |              |          | x     |
+            | construction |              |          |       |
+            | proposed     |              |          |       |
+            | primary      | primary      |          |       |
```
emiltin: construction=primary is unothodox. usually it's construction=yes

1ec5:
> only thing i can think of is construction=no... is that a thing?

Yes, there are a could tags like [`construction=widening`](https://taginfo.openstreetmap.org/tags/construction=widening) and [`construction=no`](https://taginfo.openstreetmap.org/tags/construction=no) where the road remains passable, but they occur at a much lower rate than `construction=yes`.

daniel-j-h:
Added negative tags and taginfo items. Cherry-picked into 5.9 branch.

Issue #4260, opened 2017-07-10
Deprecate `UseLane` instruction


daniel-j-h:
We're currently emitting a `UseLane` instruction when the user has to take specific lanes but otherwise keep going straight (think: lane anticipation triggers).

Now that we have the `intersections` array in the `RouteStep` we can deprecate the `UseLane` instruction and have our clients figure out lane details on their own.

Tasks:
- [ ] remove `UseLane`
- [ ] adapt `docs/http.md`
- [ ] add changelog entry
- [ ] ticket removal for v6

daniel-j-h:
Blocked by: https://github.com/Project-OSRM/osrm-backend/issues/4261 distance and duration in the intersections array. Otherwise clients only have the `(lon, lat)` for deciding if and when to announce an intermediate intersection.

MoKob:
@daniel-j-h can you elaborate why you consider #4261 as blocking? I would say that our current process isn't optimal. But I would say that #4268 should be mergable. Adding information to the intersections would only be a bonus to make it easier to distinguish for the client.

Actual durations will be determined on device, anyhow. For that reason, I would consider #4261 as an additional feature, but not as blocking.

Also: capturing from #4267: We should think about doing the same for new-name and put the name and its relation into the intersection. This would allow full visual representation and reduce verbosity of the engine.

daniel-j-h:
How can clients determine the distance and duration between intersections along the route if we remove `UseLane`? All clients have is the coordinate for the intermediate intersection - but no way of knowing when to announce it ahead of time, no? cc @1ec5 for context.

MoKob:
Well. Let me turn that question around. Without constant re-routing, how can we determine the duration? We can do it with distance, but distance does not necessarily describe the best way to announce a turn.
So if we agree that the main announcement should be the turn and `use lane` should only be used as a visual cue, we only need to see whether we passed a coordinate, which should be possible to deduce. The device needs to handle this anyhow, right now.

1ec5:
> How can clients determine the distance and duration between intersections along the route if we remove `UseLane`? All clients have is the coordinate for the intermediate intersection - but no way of knowing when to announce it ahead of time, no?

Correct, a client currently has to assume that the user will travel at the current rate (i.e., the instantaneous speed from a GPS reading) from the current location to the end of the step.

MoKob:
[...]

PR #4268, opened 2017-07-10
deprecate use-lane

MoKob:
# Issue

Resolves https://github.com/Project-OSRM/osrm-backend/issues/4260

## Tasklist
 - [x] review
 - [x] adjust for comments

### DIFF
--- a/docs/http.md
+++ b/docs/http.md
@@ -657,7 +657,7 @@ step.
 | `off ramp`       | take a ramp to exit a highway (direction given my `modifier`)  |
 | `fork`           | take the left/right side at a fork depending on `modifier`   |
 | `end of road`    | road ends in a T intersection turn in direction of `modifier`|
-| `use lane`       | going straight on a specific lane -- deprecated, use the intersections information |
+| `use lane`       | **Deprecated** replaced by lanes on all intersection entries |
 | `continue`       | Turn in direction of `modifier` to stay on the same road     |
 | `roundabout`     | traverse roundabout, has additional property `exit` with NR if the roundabout is left. The modifier specifies the direction of entering the roundabout. |
 | `rotary`         | a traffic circle. While very similar to a larger version of a roundabout, it does not necessarily follow roundabout rules for right of way. It can offer `rotary_name` and/or `rotary_pronunciation` parameters (located in the RouteStep object) in addition to the `exit` parameter (located on the StepManeuver object).  |
--- a/docs/http.md
+++ b/docs/http.md
@@ -657,7 +657,7 @@ step.
 | `off ramp`       | take a ramp to exit a highway (direction given my `modifier`)  |
 | `fork`           | take the left/right side at a fork depending on `modifier`   |
 | `end of road`    | road ends in a T intersection turn in direction of `modifier`|
-| `use lane`       | going straight on a specific lane                            |
+| `use lane`       | going straight on a specific lane -- deprecated, use the intersections information |
 | `continue`       | Turn in direction of `modifier` to stay on the same road     |
 | `roundabout`     | traverse roundabout, has additional property `exit` with NR if the roundabout is left. The modifier specifies the direction of entering the roundabout. |
 | `rotary`         | a traffic circle. While very similar to a larger version of a roundabout, it does not necessarily follow roundabout rules for right of way. It can offer `rotary_name` and/or `rotary_pronunciation` parameters (located in the RouteStep object) in addition to the `exit` parameter (located on the StepManeuver object).  |
--- a/features/guidance/anticipate-lanes.feature
+++ b/features/guidance/anticipate-lanes.feature
@@ -3,7 +3,7 @@ Feature: Turn Lane Guidance
 
     Background:
         Given the profile "car"
-        Given a grid size of 20 meters
+        Given a grid size of 100 meters
 
     @anticipate
     Scenario: Anticipate Lane Change for subsequent multi-lane intersections
@@ -371,9 +371,9 @@ Feature: Turn Lane Guidance
        Scenario: Anticipate Lanes for through with turn before / after
            Given the node map
                """
-               c       g       l
-               b – d – e – h - i
-               a       f       j
+               c   g   l
+               b d e h i
+               a   f   j
                """
 
            And the ways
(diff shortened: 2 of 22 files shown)


features/support/route.js
```diff
@@ -171,6 +171,22 @@ module.exports = function () {
                                                               ('out' in s.intersections[0] ? s.intersections[0].bearings[s.intersections[0].out] : 0));
     };
 
+    this.lanesList = (instructions) => {
+        return this.extractInstructionList(instructions, s => {
+            return s.intersections.map( i => {
+                if(i.lanes == undefined )
```
TheMarex: `if(i.lanes)` + inverting this if would be the more javascripty way (`undefined` is a `falsey` value).

src/engine/guidance/post_processing.cpp
```diff
@@ -608,7 +608,7 @@ std::vector<RouteStep> buildIntersections(std::vector<RouteStep> steps)
     {
         auto &step = steps[step_index];
         const auto instruction = step.maneuver.instruction;
-        if (instruction.type == TurnType::Suppressed)
+        if (instruction.type == TurnType::Suppressed || instruction.type == TurnType::UseLane)
```
TheMarex: That means we still create `UseLane` instructions but threat them like `Suppressed`? Can we maybe change this in the pre-processing so that we will just emit `Suppress` instructions in these cases?

features/support/route.js
MoKob: Could you be more specific, I don't believe I understand what you ask for (my javascript foo is weak :( )

src/engine/guidance/post_processing.cpp
MoKob: There is a slight difference in how we treat these instructions in multiple processing steps. I can try, but I am unsure how it would interfere with other items (e.g. lane anticipation and or collapsing). Will report back after trying.

features/support/route.js
TheMarex: Ah just invert the if, like `if (i.lanes) { /* do sensible stuff */ } else { return ""; }`

TheMarex:
@MoKob synced with our upstream users: Deprecating `UseLane` is 👍

src/engine/guidance/post_processing.cpp
MoKob: I've removed the usage completely (except for a single location which exposes it to the scripting environment to not break profiles). At least the tests pass alright, so I would be inclined towards the complete rip-out you suggested.

docs/http.md
```diff
@@ -657,7 +657,7 @@ step.
 | `off ramp`       | take a ramp to exit a highway (direction given my `modifier`)  |
 | `fork`           | take the left/right side at a fork depending on `modifier`   |
 | `end of road`    | road ends in a T intersection turn in direction of `modifier`|
-| `use lane`       | going straight on a specific lane                            |
+| `use lane`       | going straight on a specific lane -- deprecated, use the intersections information |
```
daniel-j-h: Can you be consistent with the deprecation warning for ramp here

features/guidance/anticipate-lanes.feature
```diff
@@ -151,9 +151,9 @@ Feature: Turn Lane Guidance
             | cj    |                    | 1     | motorway_link | yes    | xbcj |
 
        When I route I should get
-            | waypoints | route        | turns                           | lanes                          |
-            | a,i       | ab,ci,ci     | depart,turn slight right,arrive | ,none:false slight right:true, |
-            | a,j       | ab,xbcj      | depart,arrive                   | ,                              |
+            | waypoints | route        | turns                           | lanes                           |
+            | a,i       | ab,ci,ci     | depart,turn slight right,arrive | ;,none:false slight right:true, |
+            | a,j       | ab,xbcj      | depart,arrive                   | ;;none:true slight right:false, |
```
daniel-j-h: Why do lanes change here? Lanes should not be affected by the UseLane removal, no?

daniel-j-h: Same below

include/engine/api/route_api.hpp
```diff
@@ -174,14 +174,13 @@ class RouteAPI : public BaseAPI
                 guidance::trimShortSegments(steps, leg_geometry);
                 leg.steps = guidance::postProcess(std::move(steps));
                 leg.steps = guidance::collapseTurnInstructions(std::move(leg.steps));
+                leg.steps = guidance::anticipateLaneChange(std::move(leg.steps));
```
daniel-j-h: Why did we move lane anticipation up the processing chain? Wasn't there a reason we had it down there?

include/extractor/guidance/turn_instruction.hpp
```diff
@@ -49,7 +49,7 @@ const constexpr Enum EnterRotary = 12;                 // Enter a rotary
 const constexpr Enum EnterAndExitRotary = 13;          // Touching a rotary
 const constexpr Enum EnterRoundaboutIntersection = 14; // Entering a small Roundabout
 const constexpr Enum EnterAndExitRoundaboutIntersection = 15; // Touching a roundabout
-const constexpr Enum UseLane = 16; // No Turn, but you need to stay on a given lane!
+// depreacted: const constexpr Enum UseLane = 16; // No Turn, but you need to stay on a given lane!
```
daniel-j-h: I would love to have it in the code base but a deprecated attribute attached to it. Then we will get compiler warnings if we ever touch it but still have the value 16 taken by it.

src/engine/guidance/lane_processing.cpp
```diff
@@ -96,7 +96,7 @@ std::vector<RouteStep> anticipateLaneChange(std::vector<RouteStep> steps,
 
             // only prevent use lanes due to making all turns. don't make turns during curvy
             // segments
-            if (previous_inst.type == TurnType::UseLane)
+            if (previous_inst.type == TurnType::Suppressed)
```
daniel-j-h: Checking for Suppressed will trigger more often then we actually want to, no?

MoKob: yes, it will trigger more often, but that is fine. Time_to_constrained will only increase, if we have suppressed instructions here. After all a suppressed turn is not influencing our ability to make the turn, right? Just as use-lane. So we only increase our chance to `can_make_all_turns`.

include/engine/api/route_api.hpp
MoKob: That was because of buildIntersection now removing use-lane. But with use-lane gone we can stay on the safe side and move it back down (perhaps). I am not 100% sure about how it affects the `can make all turns` combination below

features/guidance/anticipate-lanes.feature
MoKob: lanes do not change. We only show all lanes now. Before we were only showing lanes directly at the turns and suppressed use-lane as long as all lanes were allowed (as is the case here).

TheMarex:
@daniel-j-h @MoKob can you sort out this PR and https://github.com/Project-OSRM/osrm-backend/pull/4296 tomorrow morning, so I could maybe a another RC in the evening? Would be great if you could cherry-pick this to the `5.9` branch. Thanks!

Issue #4158, opened 2017-06-14
Datafacades assume data layout is in a contiguous memory block


danpat:
Currently, it is only possible to implement alternative datafacades using the all-in-one-block structure described by the ContiguousInternalMemoryDatafacade.

If someone would want to implement an alternative storage approach - e.g. keeping data in individual files and `mmap()`-ing them, then our current inheritance structure makes this impossible - the algorithm-specific classes assume they're based on the ContiguousInternalMemoryDatafacade rather than the BaseDatafacade.

We should refactor this so that alternative storage can be implemented without needing to re-write the algorithm-specific subclasses.

TheMarex:
I talked with @oxidase about this yesterday a little bit an we concluded the best option here would be a compile time switch, since we don't need the flexibility of switching this at run-time.

There are two methods to do this:

1. Go full template. This is undesirable on multiple levels and chances are we would never use this functionality.
2. Compile-time switch using a type alias. For external embedders of OSRM this should work just as well.

PR #4269, opened 2017-07-10
Some algorithm datafacade interface refactoring

deniskoronchik:
Now it doesn't depend on cell and partiition data storage implementation

# Issue

Currently interface of algorithm datafacade depends on implementation of CellStorage and PartitionView. It doesn't allow to implement another way to store this data for MLD. This pull requests implement data structure independent interface for MLD algorithm datafacade.

## Tasklist
 - [x] ADD OWN TASKS HERE
 - [ ] review
 - [ ] adjust for comments

## Requirements / Relations

### DIFF
--- a/include/engine/datafacade/contiguous_internalmem_datafacade.hpp
+++ b/include/engine/datafacade/contiguous_internalmem_datafacade.hpp
@@ -1070,12 +1070,46 @@ template <> class ContiguousInternalMemoryAlgorithmDataFacade<MLD> : public Algo
         InitializeInternalPointers(allocator->GetLayout(), allocator->GetMemory());
     }
 
-    const partition::MultiLevelPartitionView &GetMultiLevelPartition() const override
+    LevelID GetQueryLevel(NodeID source, NodeID target, NodeID node) const override
     {
-        return mld_partition;
+        return mld_partition.GetQueryLevel(source, target, node);
     }
 
-    const partition::CellStorageView &GetCellStorage() const override { return mld_cell_storage; }
+    LevelID GetHighestDifferentLevel(NodeID first, NodeID second) const override
+    {
+        return mld_partition.GetHighestDifferentLevel(first, second);
+    }
+
+    void ForEachSourceNodes(LevelID level, NodeID node, NodeFnT const &fn) const override
+    {
+        const auto &cell = mld_cell_storage.GetCell(level, mld_partition.GetCell(level, node));
+        auto source = cell.GetSourceNodes().begin();
+        auto durations = cell.GetInDuration(node).begin();
+
+        BOOST_ASSERT(cell.GetInDuration(node).size() == cell.GetInWeight(node).size());
+
+        for (auto shortcut_weight : cell.GetInWeight(node))
+            fn(*source++, shortcut_weight, *durations++);
+    }
+
+    void ForEachDestinationNodes(LevelID level, NodeID node, NodeFnT const &fn) const override
+    {
+        const auto &cell = mld_cell_storage.GetCell(level, mld_partition.GetCell(level, node));
+        auto destination = cell.GetDestinationNodes().begin();
+        auto durations = cell.GetOutDuration(node).begin();
+
+        BOOST_ASSERT(cell.GetOutDuration(node).size() == cell.GetOutWeight(node).size());
+
+        for (auto shortcut_weight : cell.GetOutWeight(node))
+            fn(*destination++, shortcut_weight, *durations++);
+    }
+
+    CellID GetPartitionCell(LevelID level, NodeID node) const override
+    {
+        return mld_partition.GetCell(level, node);
+    }
+
+    virtual uint8_t GetNumberOfLevels() const override { return mld_partition.GetNumberOfLevels(); }
 
     // search graph access
     unsigned GetNumberOfNodes() const override final { return query_graph.GetNumberOfNodes(); }
--- a/include/engine/routing_algorithms/routing_base_mld.hpp
+++ b/include/engine/routing_algorithms/routing_base_mld.hpp
@@ -30,13 +30,14 @@ namespace
 // Unrestricted search (Args is const PhantomNodes &):
 //   * use partition.GetQueryLevel to find the node query level based on source and target phantoms
 //   * allow to traverse all cells
-inline LevelID getNodeQueryLevel(const partition::MultiLevelPartitionView &partition,
-                                 NodeID node,
-                                 const PhantomNodes &phantom_nodes)
+inline LevelID
+getNodeQueryLevel(const datafacade::ContiguousInternalMemoryDataFacade<Algorithm> &facade,
+                  NodeID node,
+                  const PhantomNodes &phantom_nodes)
 {
-    auto level = [&partition, node](const SegmentID &source, const SegmentID &target) {
+    auto level = [&facade, node](const SegmentID &source, const SegmentID &target) {
         if (source.enabled && target.enabled)
-            return partition.GetQueryLevel(source.id, target.id, node);
+            return facade.GetQueryLevel(source.id, target.id, node);
         return INVALID_LEVEL_ID;
     };
     return std::min(std::min(level(phantom_nodes.source_phantom.forward_segment_id,
@@ -54,8 +55,10 @@ inline bool checkParentCellRestriction(CellID, const PhantomNodes &) { return tr
 // Restricted search (Args is LevelID, CellID):
 //   * use the fixed level for queries
 //   * check if the node cell is the same as the specified parent onr
-inline LevelID
-getNodeQueryLevel(const partition::MultiLevelPartitionView &, NodeID, LevelID level, CellID)
+inline LevelID getNodeQueryLevel(const datafacade::ContiguousInternalMemoryDataFacade<Algorithm> &,
+                                 NodeID,
+                                 LevelID level,
+                                 CellID)
 {
     return level;
 }
@@ -140,9 +143,6 @@ void routingStep(const datafacade::ContiguousInternalMemoryDataFacade<Algorithm>
                  const bool force_loop_reverse,
                  Args... args)
 {
-    const auto &partition = facade.GetMultiLevelPartition();
-    const auto &cells = facade.GetCellStorage();
-
     const auto node = forward_heap.DeleteMin();
     const auto weight = forward_heap.GetKey(node);
 
@@ -166,61 +166,53 @@ void routingStep(const datafacade::ContiguousInternalMemoryDataFacade<Algorithm>
         }
     }
 
-    const auto level = getNodeQueryLevel(partition, node, args...);
+    const auto level = getNodeQueryLevel(facade, node, args...);
 
     if (level >= 1 && !forward_heap.GetData(node).from_clique_arc)
     {
         if (DIRECTION == FORWARD_DIRECTION)
         {
             // Shortcuts in forward direction
-            const auto &cell = cells.GetCell(level, partition.GetCell(level, node));
-            auto destination = cell.GetDestinationNodes().begin();
-            for (auto shortcut_weight : cell.GetOutWeight(node))
-            {
-                BOOST_ASSERT(destination != cell.GetDestinationNodes().end());
-                const NodeID to = *destination;
-                if (shortcut_weight != INVALID_EDGE_WEIGHT && node != to)
-                {
-                    const EdgeWeight to_weight = weight + shortcut_weight;
-                    BOOST_ASSERT(to_weight >= weight);
-                    if (!forward_heap.WasInserted(to))
+            facade.ForEachDestinationNodes(
+                level, node, [&](NodeID destination, EdgeWeight shortcut_weight, EdgeDuration) {
+                    const NodeID to = destination;
+                    if (shortcut_weight != INVALID_EDGE_WEIGHT && node != to)
                     {
-                        forward_heap.Insert(to, to_weight, {node, true});
+                        const EdgeWeight to_weight = weight + shortcut_weight;
+                        BOOST_ASSERT(to_weight >= weight);
+                        if (!forward_heap.WasInserted(to))
+                        {
+                            forward_heap.Insert(to, to_weight, {node, true});
+                        }
+                        else if (to_weight < forward_heap.GetKey(to))
+                        {
+                            forward_heap.GetData(to) = {node, true};
+                            forward_heap.DecreaseKey(to, to_weight);
+                        }
                     }
-                    else if (to_weight < forward_heap.GetKey(to))
-                    {
-                        forward_heap.GetData(to) = {node, true};
-                        forward_heap.DecreaseKey(to, to_weight);
-                    }
-                }
-                ++destination;
-            }
+                });
         }
         else
         {
             // Shortcuts in backward direction
-            const auto &cell = cells.GetCell(level, partition.GetCell(level, node));
-            auto source = cell.GetSourceNodes().begin();
-            for (auto shortcut_weight : cell.GetInWeight(node))
-            {
-                BOOST_ASSERT(source != cell.GetSourceNodes().end());
-                const NodeID to = *source;
-                if (shortcut_weight != INVALID_EDGE_WEIGHT && node != to)
-                {
-                    const EdgeWeight to_weight = weight + shortcut_weight;
-                    BOOST_ASSERT(to_weight >= weight);
-                    if (!forward_heap.WasInserted(to))
+            facade.ForEachSourceNodes(
+                level, node, [&](NodeID source, EdgeWeight shortcut_weight, EdgeDuration) {
+                    const NodeID to = source;
+                    if (shortcut_weight != INVALID_EDGE_WEIGHT && node != to)
                     {
-                        forward_heap.Insert(to, to_weight, {node, true});
+                        const EdgeWeight to_weight = weight + shortcut_weight;
+                        BOOST_ASSERT(to_weight >= weight);
+                        if (!forward_heap.WasInserted(to))
+                        {
+                            forward_heap.Insert(to, to_weight, {node, true});
+                        }
+                        else if (to_weight < forward_heap.GetKey(to))
+                        {
+                            forward_heap.GetData(to) = {node, true};
+                            forward_heap.DecreaseKey(to, to_weight);
+                        }
                     }
-                    else if (to_weight < forward_heap.GetKey(to))
-                    {
-                        forward_heap.GetData(to) = {node, true};
-                        forward_heap.DecreaseKey(to, to_weight);
-                    }
-                }
-                ++source;
-            }
+                });
         }
     }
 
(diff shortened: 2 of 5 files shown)


include/engine/datafacade/contiguous_internalmem_datafacade.hpp
```diff
@@ -1070,12 +1070,44 @@ template <> class ContiguousInternalMemoryAlgorithmDataFacade<MLD> : public Algo
...
+        return mld_partition.GetHighestDifferentLevel(first, second);
+    }
+
+    void ForEachSourceNodes(LevelID level, NodeID node, NodeFnT const &fn) const override
+    {
+        const auto &cell = mld_cell_storage.GetCell(level, mld_partition.GetCell(level, node));
+        auto source = cell.GetSourceNodes().begin();
+        auto durations = cell.GetInDuration(node).begin();
+
+        BOOST_ASSERT(cell.GetSourceNodes().size() == cell.GetInDuration(node).size());
```
vng: Add assert for GetInWeight().
Please, separate asserts with blank lines before and after.

include/engine/datafacade/contiguous_internalmem_datafacade.hpp
```diff
@@ -1070,12 +1070,44 @@ template <> class ContiguousInternalMemoryAlgorithmDataFacade<MLD> : public Algo
...
+        for (auto shortcut_weight : cell.GetInWeight(node))
+            fn(*source++, shortcut_weight, *durations++);
+    }
+
+    void ForEachDestinationNodes(LevelID level, NodeID node, NodeFnT const &fn) const override
+    {
+        const auto &cell = mld_cell_storage.GetCell(level, mld_partition.GetCell(level, node));
+        auto destination = cell.GetDestinationNodes().begin();
+        auto durations = cell.GetOutDuration(node).begin();
+        BOOST_ASSERT(cell.GetDestinationNodes().size() == cell.GetOutDuration(node).size());
```
vng: Add assert for GetOutWeight().
Please, separate asserts with blank lines before and after.

include/engine/routing_algorithms/routing_base_mld.hpp
```diff
@@ -30,13 +30,14 @@ namespace
 // Unrestricted search (Args is const PhantomNodes &):
 //   * use partition.GetQueryLevel to find the node query level based on source and target phantoms
 //   * allow to traverse all cells
-inline LevelID getNodeQueryLevel(const partition::MultiLevelPartitionView &partition,
-                                 NodeID node,
-                                 const PhantomNodes &phantom_nodes)
+inline LevelID
+getNodeQueryLevel(const datafacade::ContiguousInternalMemoryDataFacade<Algorithm> &facade,
```
vng: Please, write it "inline LevelID getNodeQueryLevel(...) in one string as before.

include/engine/routing_algorithms/routing_base_mld.hpp
```diff
@@ -166,61 +166,53 @@ void routingStep(const datafacade::ContiguousInternalMemoryDataFacade<Algorithm>
...
-                {
-                    const EdgeWeight to_weight = weight + shortcut_weight;
-                    BOOST_ASSERT(to_weight >= weight);
-                    if (!forward_heap.WasInserted(to))
-                    {
-                        forward_heap.Insert(to, to_weight, {node, true});
-                    }
-                    else if (to_weight < forward_heap.GetKey(to))
+            facade.ForEachDestinationNodes(
+                level, node, [&](NodeID destination, EdgeWeight shortcut_weight, EdgeDuration) {
```
vng: Better according to code style is:
```
facade.ForEachDestinationNodes(level, node,
     [&](NodeID destination, EdgeWeight shortcut_weight, EdgeDuration)
     {
```

include/engine/routing_algorithms/routing_base_mld.hpp
```diff
@@ -166,61 +166,53 @@ void routingStep(const datafacade::ContiguousInternalMemoryDataFacade<Algorithm>
...
-                {
-                    const EdgeWeight to_weight = weight + shortcut_weight;
-                    BOOST_ASSERT(to_weight >= weight);
-                    if (!forward_heap.WasInserted(to))
-                    {
-                        forward_heap.Insert(to, to_weight, {node, true});
-                    }
-                    else if (to_weight < forward_heap.GetKey(to))
+            facade.ForEachSourceNodes(
+                level, node, [&](NodeID source, EdgeWeight shortcut_weight, EdgeDuration) {
```
vng: Better according to code style is:
```
facade.ForEachDestinationNodes(level, node,
     [&](NodeID destination, EdgeWeight shortcut_weight, EdgeDuration)
     {
```

include/engine/routing_algorithms/routing_base_mld.hpp
vng: Leave the previous formatting and replace ContiguousInternalMemoryDataFacade with AlgorithmDataFacade

include/engine/routing_algorithms/routing_base_mld.hpp
```diff
@@ -54,8 +55,10 @@ inline bool checkParentCellRestriction(CellID, const PhantomNodes &) { return tr
 // Restricted search (Args is LevelID, CellID):
 //   * use the fixed level for queries
 //   * check if the node cell is the same as the specified parent onr
-inline LevelID
-getNodeQueryLevel(const partition::MultiLevelPartitionView &, NodeID, LevelID level, CellID)
+inline LevelID getNodeQueryLevel(const datafacade::ContiguousInternalMemoryDataFacade<Algorithm> &,
```
vng: Leave the previous formatting and replace ContiguousInternalMemoryDataFacade with AlgorithmDataFacade

src/engine/routing_algorithms/alternative_path_mld.cpp
```diff
@@ -199,7 +199,9 @@ filterViaCandidatesByViaNotOnPath(const WeightedViaNodePackedPath &path, RandIt
 // Filters packed paths with similar cells between each other. Mutates range in-place.
 // Returns an iterator to the filtered range's new end.
 template <typename RandIt>
-RandIt filterPackedPathsByCellSharing(RandIt first, RandIt last, const Partition &partition)
+RandIt filterPackedPathsByCellSharing(RandIt first,
+                                      RandIt last,
+                                      const datafacade::AlgorithmDataFacade<Algorithm> &facade)
```
vng: first, last can fit into one string.
Algorithm - where is it defined?

src/engine/routing_algorithms/many_to_many.cpp
```diff
@@ -129,62 +126,53 @@ void relaxOutgoingEdges(
...
-                {
-                    const auto to_weight = weight + shortcut_weight;
-                    const auto to_duration = duration + shortcut_durations.front();
-                    if (!query_heap.WasInserted(to))
+            facade.ForEachDestinationNodes(
+                level,
+                node,
+                [&](NodeID destination,
+                    EdgeWeight shortcut_weight,
+                    EdgeDuration shortcut_duration) {
```
vng: Better according to code style is:
```
facade.ForEachDestinationNodes(level, node,
     [&](NodeID destination, EdgeWeight shortcut_weight, EdgeDuration)
     {
```

src/engine/routing_algorithms/many_to_many.cpp
```diff
@@ -129,62 +126,53 @@ void relaxOutgoingEdges(
...
-                    const auto to_duration = duration + shortcut_durations.front();
-                    if (!query_heap.WasInserted(to))
-                    {
-                        query_heap.Insert(to, to_weight, {node, true, to_duration});
-                    }
-                    else if (to_weight < query_heap.GetKey(to))
+            facade.ForEachSourceNodes(
+                level,
+                node,
+                [&](NodeID source, EdgeWeight shortcut_weight, EdgeDuration shortcut_duration) {
```
vng: Better according to code style is:
```
facade.ForEachDestinationNodes(level, node,
     [&](NodeID destination, EdgeWeight shortcut_weight, EdgeDuration)
     {
```

include/engine/routing_algorithms/routing_base_mld.hpp
deniskoronchik: https://github.com/Project-OSRM/osrm-backend/blob/master/include/engine/algorithm.hpp#L30

include/engine/datafacade/contiguous_internalmem_datafacade.hpp
```diff
@@ -1070,12 +1070,48 @@ template <> class ContiguousInternalMemoryAlgorithmDataFacade<MLD> : public Algo
...
+            fn(*source++, shortcut_weight, *durations++);
+    }
+
+    void ForEachDestinationNodes(LevelID level, NodeID node, NodeFnT const &fn) const override
+    {
+        const auto &cell = mld_cell_storage.GetCell(level, mld_partition.GetCell(level, node));
+        auto destination = cell.GetDestinationNodes().begin();
+        auto durations = cell.GetOutDuration(node).begin();
+
+        BOOST_ASSERT(cell.GetDestinationNodes().size() == cell.GetOutDuration(node).size());
```
deniskoronchik: There is a very strange behaviour, that can be an error in OSRM. Sometimes nodes count not equal to durations count. For example in local tests, sometimes i got 20 nodes and 0 durations and weights.
This is really strange in 90-95% case it equal, why?

vng:
Update: Removed all code style comments because of automatic code formatting.

include/engine/datafacade/contiguous_internalmem_datafacade.hpp
TheMarex: > sometimes i got 20 nodes and 0 durations and weights

This would indicate that the node was not a source/destination node respectively. Calling `cell.GetOutDuration(node)` on a node that is not actually a source node for the cell will return an empty range. Same for destination nodes and `cell.GetInDuration(node)` if `node` is not a destination node.

deniskoronchik: Still don't know why number of nodes not equal to number of durations and weight. But duration and weight number should be equal according to a code.

deniskoronchik: I we fixed that issue now, to work as it was in original OSRM. Wait for testing result. Local tests work perfectly.

danpat:
Main ticket for this: https://github.com/Project-OSRM/osrm-backend/issues/4158

deniskoronchik:
@danpat I'll prepare next pull request to fix that issue, it would pass datafacade base interfaces into algorithms

include/engine/datafacade/contiguous_internalmem_datafacade.hpp
```diff
@@ -1070,12 +1070,46 @@ template <> class ContiguousInternalMemoryAlgorithmDataFacade<MLD> : public Algo
...
+        return mld_partition.GetHighestDifferentLevel(first, second);
+    }
+
+    void ForEachSourceNodes(LevelID level, NodeID node, NodeFnT const &fn) const override
+    {
+        const auto &cell = mld_cell_storage.GetCell(level, mld_partition.GetCell(level, node));
+        auto source = cell.GetSourceNodes().begin();
+        auto durations = cell.GetInDuration(node).begin();
+
+        BOOST_ASSERT(cell.GetInDuration(node).size() == cell.GetInWeight(node).size());
```
vng: Add BOOST_ASSERT(cell.GetSourceNodes().size() >= cell.GetInWeight(node).size());

include/engine/datafacade/contiguous_internalmem_datafacade.hpp
```diff
@@ -1070,12 +1070,46 @@ template <> class ContiguousInternalMemoryAlgorithmDataFacade<MLD> : public Algo
...
+            fn(*source++, shortcut_weight, *durations++);
+    }
+
+    void ForEachDestinationNodes(LevelID level, NodeID node, NodeFnT const &fn) const override
+    {
+        const auto &cell = mld_cell_storage.GetCell(level, mld_partition.GetCell(level, node));
+        auto destination = cell.GetDestinationNodes().begin();
+        auto durations = cell.GetOutDuration(node).begin();
+
+        BOOST_ASSERT(cell.GetOutDuration(node).size() == cell.GetOutWeight(node).size());
```
vng: Add BOOST_ASSERT(cell. GetDestinationNodes().size() >= cell. GetOutWeight(node).size());

include/engine/routing_algorithms/routing_base_mld.hpp
```diff
@@ -166,61 +166,53 @@ void routingStep(const datafacade::ContiguousInternalMemoryDataFacade<Algorithm>
...
-                    const EdgeWeight to_weight = weight + shortcut_weight;
-                    BOOST_ASSERT(to_weight >= weight);
-                    if (!forward_heap.WasInserted(to))
-                    {
-                        forward_heap.Insert(to, to_weight, {node, true});
-                    }
-                    else if (to_weight < forward_heap.GetKey(to))
+            facade.ForEachDestinationNodes(
+                level, node, [&](NodeID destination, EdgeWeight shortcut_weight, EdgeDuration) {
+                    const NodeID to = destination;
```
vng: No need in this var. Just rename "NodeID destination" to "NodeID to".

include/engine/routing_algorithms/routing_base_mld.hpp
```diff
@@ -166,61 +166,53 @@ void routingStep(const datafacade::ContiguousInternalMemoryDataFacade<Algorithm>
...
-                    const EdgeWeight to_weight = weight + shortcut_weight;
-                    BOOST_ASSERT(to_weight >= weight);
-                    if (!forward_heap.WasInserted(to))
-                    {
-                        forward_heap.Insert(to, to_weight, {node, true});
-                    }
-                    else if (to_weight < forward_heap.GetKey(to))
+            facade.ForEachSourceNodes(
+                level, node, [&](NodeID source, EdgeWeight shortcut_weight, EdgeDuration) {
+                    const NodeID to = source;
```
vng: No need in this var. Just rename "NodeID source" to "NodeID to".

src/engine/routing_algorithms/many_to_many.cpp
```diff
@@ -129,62 +126,53 @@ void relaxOutgoingEdges(
...
-                    const auto to_weight = weight + shortcut_weight;
-                    const auto to_duration = duration + shortcut_durations.front();
-                    if (!query_heap.WasInserted(to))
+            facade.ForEachDestinationNodes(
+                level,
+                node,
+                [&](NodeID destination,
+                    EdgeWeight shortcut_weight,
+                    EdgeDuration shortcut_duration) {
+                    const NodeID to = destination;
```
vng: No need in this var. Just rename "NodeID destination" to "NodeID to".

src/engine/routing_algorithms/many_to_many.cpp
```diff
@@ -129,62 +126,53 @@ void relaxOutgoingEdges(
...
-                    if (!query_heap.WasInserted(to))
-                    {
-                        query_heap.Insert(to, to_weight, {node, true, to_duration});
-                    }
-                    else if (to_weight < query_heap.GetKey(to))
+            facade.ForEachSourceNodes(
+                level,
+                node,
+                [&](NodeID source, EdgeWeight shortcut_weight, EdgeDuration shortcut_duration) {
+                    const NodeID to = source;
```
vng: No need in this var. Just rename "NodeID destination" to "NodeID to".

include/engine/datafacade/contiguous_internalmem_datafacade.hpp
```diff
@@ -1070,12 +1070,46 @@ template <> class ContiguousInternalMemoryAlgorithmDataFacade<MLD> : public Algo
...
+        return mld_partition.GetQueryLevel(source, target, node);
     }
 
-    const partition::CellStorageView &GetCellStorage() const override { return mld_cell_storage; }
+    LevelID GetHighestDifferentLevel(NodeID first, NodeID second) const override
+    {
+        return mld_partition.GetHighestDifferentLevel(first, second);
+    }
+
+    void ForEachSourceNodes(LevelID level, NodeID node, NodeFnT const &fn) const override
```
TheMarex: Functions like these are great for the convenience but I'm a little bit worried about introducing interfaces that use `std::function` objects.

I would be great if we could fine an interface that doesn't need this and still makes it possible to swap out enough so that you can implement your external algorithms.

/cc @oxidase might have some ideas as well.

oxidase: I see two problems:
* instead of 3 virtual function calls there will be `node degree` indirect function calls that will not be optimized
* point-to-point path search does need duration values, so it also brings unnecessary `cell.GetInDuration(node)` call overhead per node

It is better to directly return iterator ranges and wrap into a zip iterator using `boost::combine` or  as we stick with boost 1.55 a `zip` function
```
template <typename... T>
auto zip(T&&... containers) -> boost::iterator_range<boost::zip_iterator<decltype(boost::make_tuple(std::begin(containers)...))>>
{
    auto zip_begin = boost::make_zip_iterator(boost::make_tuple(std::begin(containers)...));
    auto zip_end = boost::make_zip_iterator(boost::make_tuple(std::end(containers)...));
    return boost::make_iterator_range(zip_begin, zip_end);
}
``` 
see http://coliru.stacked-crooked.com/a/f177dc86f9b1321d from https://stackoverflow.com/questions/8511035/sequence-zip-function-for-c11

deniskoronchik: @oxidase What if we have no plane structure of data? Then we need to implement abstract iterator interface, that would have ~equal call parfomance

vng: To use iterator concept, we need to get concrete iterator type in base facade interface. But it also limits us in using different structures for storing in/out nodes, weights, durations.

Do you have any benchmark tests to check the performance?

oxidase: @deniskoronchik sorry, i don't understand the use case, but what do you mean by the plain structure of data? The callback function signature is fixed to `void(NodeID, EdgeWeight, EdgeDuration)>` at compile time, but is it really needed to change a callback function at runtime and what would a use case for this?

As a solution i would suggest to add simple proxy-functions to facade (this requires explicit types for return values of function  and extend facade with `ForEachSourceNodes`/`ForEachDestinationNodes` to have possibility for runtime selection of a callback.

deniskoronchik: @oxidase This functions made as a replacement of 
```cpp
const partition::MultiLevelPartitionView &GetMultiLevelPartition() const
const partition::CellStorageView &GetCellStorage() const
```
Because there are no way to make another implementation of `MultiLevelPartitionView` and `CellStorageView`.

Common case of usage was: https://github.com/Project-OSRM/osrm-backend/blob/master/include/engine/routing_algorithms/routing_base_mld.hpp#L176-L197

We need to encode this data in runtime. We can't store this data in memory.

oxidase: @deniskoronchik yeah, it is mainly a discussion about abstract vs generic facade.

@vng there is no direct MLD benchmarks. `scripts/osrm-runner.js` allows to run random queries and repeat them to make pairwise tests

deniskoronchik:
Will prepare pull request for a new approach

Issue #4270, opened 2017-07-10
Exits: Use comma instead of semicolon as separator


freenerd:
(Found by @1ec5 in https://github.com/mapbox/MapboxDirections.swift/pull/149)

To stay consistent with multiple names and destinations, exits https://github.com/Project-OSRM/osrm-backend/pull/4215 should use comma `,` as separator instead of `;` semicolon.

daniel-j-h:
We need arrays for these in a new API badly (https://github.com/Project-OSRM/osrm-backend/issues/3329). I don't think there can be multiple names - and if so we use them as is; destination is the only place where we use a comma and that's more of a bug than a feature:

https://github.com/Project-OSRM/osrm-backend/blob/5b5a907023a08a10daef649139fea08259ace610/profiles/lib/destination.lua#L11

https://github.com/Project-OSRM/osrm-backend/blob/df3c39cef51acd8d55287d51cd37bbad39b6f68c/include/extractor/extraction_helper_functions.hpp#L121-L135

https://github.com/Project-OSRM/osrm-backend/blob/5b5a907023a08a10daef649139fea08259ace610/profiles/lib/handlers.lua#L30-L53



---

What we can do now is either gsub `;` to `,` and have an inconsistency between

    destination, exits

and

    name, ref, pronunciation

or not to change it to a comma and consider the comma in `destinations` a bug that we can't fix at the moment because of API compatibility.

freenerd:
Agreed with using arrays so much!

The inconsistency with destination vs. others slipped my mind. Let's stick with the semicolon since it's closer to OSM.

Fixed downstream https://github.com/Project-OSRM/osrm-text-instructions/pull/126

PR #4271, opened 2017-07-10
Canonicalizes exits string list handling in the profiles

daniel-j-h:
Looking over https://github.com/Project-OSRM/osrm-backend/issues/4270 I found some inconsistencies in canonicalizing our string lists. We do the same for ref and should be consistent here.

- [x] Review
- [x] Resolved feedback comments

### DIFF
--- a/features/guidance/exit-numbers-names.feature
+++ b/features/guidance/exit-numbers-names.feature
@@ -41,8 +41,8 @@ Feature: Exit Numbers and Names
             | ef     | motorway_link | ExitRamp |              |
 
        When I route I should get
-            | waypoints | route                      | turns                               | exits   |
-            | a,f       | MainRoad,ExitRamp,ExitRamp | depart,off ramp slight right,arrive | ,10;12, |
+            | waypoints | route                      | turns                               | exits    |
+            | a,f       | MainRoad,ExitRamp,ExitRamp | depart,off ramp slight right,arrive | ,10; 12, |
 
 
     Scenario: Exit number on the ways after the motorway junction, multiple exits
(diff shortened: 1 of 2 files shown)


features/guidance/exit-numbers-names.feature
```diff
@@ -41,8 +41,8 @@ Feature: Exit Numbers and Names
             | ef     | motorway_link | ExitRamp |              |
 
        When I route I should get
-            | waypoints | route                      | turns                               | exits   |
-            | a,f       | MainRoad,ExitRamp,ExitRamp | depart,off ramp slight right,arrive | ,10;12, |
+            | waypoints | route                      | turns                               | exits    |
+            | a,f       | MainRoad,ExitRamp,ExitRamp | depart,off ramp slight right,arrive | ,10; 12, |
```
MoKob: from the discussions around https://github.com/Project-OSRM/osrm-backend/issues/4270 I presume we will stay with `;`. So I guess this is just what we should do in terms of string handling and consistency.

TheMarex:
What about changing the canonicalization to `foo;bar;bla`? I don't see a problem with removing a trailing space but adding one could lead to odd formatting.

daniel-j-h:
I'm a bit scared of changing the response formatting too much here since clients already rely on it.

We also only ever introduce a single space (at most):

https://github.com/Project-OSRM/osrm-backend/blob/df3c39cef51acd8d55287d51cd37bbad39b6f68c/include/extractor/extraction_helper_functions.hpp#L121-L135

TheMarex:
Alright I think this is ugly but I get your point. Cherry-picking to 5.9 branch.

Issue #4274, opened 2017-07-10
Demote Trunk Roads to non-motorway type


MoKob:
[OSM](http://wiki.openstreetmap.org/wiki/Key:highway) lists trunk as the second-most important road in a country, not necessarily a divided highway.

Right now, the car profile and OSRM in general assume trunk to be just like a motorway. Trunk roads are mapped within many cities, which is why the guidance handlers currently work around normal intersections by checking connections to normal roads.

This is in conflict with some motorways starting directly connected to the road network. To allow assigning correct turns to motorways starting within a city, we should reconsider the treatment of trunks as motorways.
This aligns well with what I see in the tests, where many tests use off/on ramps on situations that just really do not look like they should be considered as such.

This would imply
 - changing the car profile in terms of the default motorway-like tags
 - updating the intersection handlers to reconsider ramp assignments
 - assign ramps even when directly entering a motorway

![screen shot 2017-07-10 at 16 59 12 2](https://user-images.githubusercontent.com/12932279/28025023-87f7f428-6592-11e7-82a1-50ef978588e2.png)

Resolving this issue would turn the above pictured `slight left` into `ramp slight left`.

MoKob:
The main problem here is that trunk roads are different on a by country basis. In Germany we need trunks to be treated like ramps, in US cities, trunk roads are most often not even closely related to ramps. If we want to treat both ways, we would require country awareness #4167

MoKob:
A case where this triggers in SF: http://map.project-osrm.org/?z=18&center=37.800890%2C-122.424393&loc=37.801082%2C-122.425954&loc=37.800561%2C-122.424479&hl=en&alt=0

PR #4275, opened 2017-07-10
demote trunk from divided highways to normal roads, anounce ramps whe…

MoKob:
…n entering a motorway directly

# Issue

resolves https://github.com/Project-OSRM/osrm-backend/issues/4274

## Tasklist
 - [ ] review
 - [ ] adjust for comments

### DIFF
--- a/src/extractor/guidance/motorway_handler.cpp
+++ b/src/extractor/guidance/motorway_handler.cpp
@@ -62,7 +62,7 @@ bool MotorwayHandler::canProcess(const NodeID,
     for (const auto &road : intersection)
     {
         // not merging or forking?
-        if (road.entry_allowed && angularDeviation(road.angle, STRAIGHT_ANGLE) > 60)
+        if (road.entry_allowed && angularDeviation(road.angle, STRAIGHT_ANGLE) > 91)
             return false;
         else if (isMotorwayClass(road.eid, node_based_graph))
         {
@@ -74,7 +74,9 @@ bool MotorwayHandler::canProcess(const NodeID,
     }
 
     if (has_normal_roads)
+    {
         return false;
+    }
 
     return has_motorway || isMotorwayClass(via_eid, node_based_graph);
 }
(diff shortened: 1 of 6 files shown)


TheMarex:
How does this affect scenarios that use a `trunk_link` like [this example](http://map.project-osrm.org/?z=17&center=49.016236%2C8.350881&loc=49.014061%2C8.351712&loc=49.018593%2C8.348193&hl=en&alt=0)? Do we emit a `on ramp` in this case, or would this become a `turn right`?

MoKob:
Sadly these would be labeled a simple turn. The problem is that we need a way to find out about these dual-carriage ways. We might be able to assign the motorway tag based on available references/country level.
While the linked example would surely be best described as a ramp, we have many other cases just like [here](https://www.openstreetmap.org/way/8922554) [osrm](http://map.project-osrm.org/?z=18&center=37.800640%2C-122.425488&loc=37.800968%2C-122.426577&loc=37.800052%2C-122.424302&hl=en&alt=0) which surely aren't ramps :(

TheMarex:
I think the OSM tagging in the highway `101` case is actually wrong. This segment is clearly part of the normal traffic flow of the highway. For example [this](https://www.openstreetmap.org/way/23876482) segment also just connects two `trunk` roads but is also tagged `trunk`.

We should first get a good understanding of what the trade-off here is. We might be improving instructions inside some US cities but seems like we could vastly decrease our guidance quality in other areas. Maybe the answer here can be to insist on better tagging on OSM side.

MoKob:
I think you are right and we should investigate further. It might be that we just need another level of country awareness, based on tagging schemes. I am tagging this as experimental and feature request for now.

src/extractor/guidance/motorway_handler.cpp
```diff
@@ -62,7 +62,7 @@ bool MotorwayHandler::canProcess(const NodeID,
     for (const auto &road : intersection)
     {
         // not merging or forking?
-        if (road.entry_allowed && angularDeviation(road.angle, STRAIGHT_ANGLE) > 60)
+        if (road.entry_allowed && angularDeviation(road.angle, STRAIGHT_ANGLE) > 91)
```
daniel-j-h: Why 91?!

Issue #4278, opened 2017-07-12
Unnecessary Turn due to misdetection in motorway handler


MoKob:
[osrm link](http://map.project-osrm.org/?z=18&center=38.893335%2C-77.055058&loc=38.893425%2C-77.054806&loc=38.893146%2C-77.055241&hl=en&alt=0)

![screen shot 2017-07-12 at 11 13 17 2](https://user-images.githubusercontent.com/12932279/28110578-251cc970-66f3-11e7-8c8b-d0713c5ef7f5.png)

This turn present in OSRM right now should probably not be considered a turn. It is a side-effect of two effects in the data.
For one, on empty names we detect changing references as the same name (/cc @daniel-j-h).
The other part is, that we do not consider the `entry_allowed` flag [here](https://github.com/Project-OSRM/osrm-backend/blob/8508834e50171665f91608b45f0c0fd191273e2e/src/extractor/guidance/motorway_handler.cpp#L118-L132).

MoKob:
This was resolved by #4279

PR #4279, opened 2017-07-12
Fix/continue on motorway

MoKob:
# Issue

Works towards https://github.com/Project-OSRM/osrm-backend/issues/4278

## Tasklist
 - [x] review
 - [x] adjust for comments


TheMarex:
Should this go out with `5.9`?

MoKob:
@TheMarex i wouldn't consider it too bad if it was in 5.9, as it is occurring on a prominent intersection in SF

PR #4280, opened 2017-07-12
Log warning for intersections with more then 32 streets

oxidase:
# Issue

The PR doubles size of `EntryClass` to prevent assertions 
```
[assert][140167020267392] /mnt/persist/api-osrm-routed-no-stxxl/2017-07-10-140458/osrm/src/util/guidance/entry_class.cpp:14
in: void osrm::util::guidance::EntryClass::activate(uint32_t): index < 8 * sizeof(FlagBaseType)
```
for nodes like http://www.openstreetmap.org/node/27547986#map=18/47.90601/106.92157

## Tasklist
 - [x] review
 - [x] adjust for comments

### DIFF
--- a/include/extractor/extractor.hpp
+++ b/include/extractor/extractor.hpp
@@ -85,12 +85,6 @@ class Extractor
                        std::vector<util::Coordinate> &coordinates,
                        extractor::PackedOSMIDs &osm_node_ids);
 
-    void WriteIntersectionClassificationData(
-        const std::string &output_file_name,
-        const std::vector<std::uint32_t> &node_based_intersection_classes,
-        const std::vector<util::guidance::BearingClass> &bearing_classes,
-        const std::vector<util::guidance::EntryClass> &entry_classes) const;
-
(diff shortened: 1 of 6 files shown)


MoKob:
@oxidase This PR leaves me wondering:

 - what is the memory impact of this? Should be small, hopefully?
 - do we maybe want to switch that assertion to an error message (just like restrictions and invalid ways)? After all, these only happen in case of data errors, right? We do not really care what we output int these situations
 - if 32 is enough for everything else, what is your feeling about staying with 32 bit and messaging in case of larger nodes?

oxidase:
@MoKob expected memory increase is about 10-20KB, but agree initial PR is not correct!
Just looked at intersections for foot profile: the largest correct intersection has only 14 roads http://www.openstreetmap.org/way/51844724#map=19/52.04310/6.08507&layers=D

I will make a hard limit of 32 classes to prevent UB and log a warning about non-activated roads

PR #4287, opened 2017-07-13
Exposes EngineConfig system-wide limits in Node.js bindings, resolves #4226

### LINKED ISSUE #4226: Expose EngineConfig Limits to Node.js
We do not expose a way to set engine config limits via our Node.js bindings.

Here is where we parse the Js object for engine config:

https://github.com/Project-OSRM/osrm-backend/blob/e2e279bc852cd024aef1ad445a81bad2884902cd/include/nodejs/node_osrm_support.hpp#L81-L180

And here are the limits (unlimited by default) we fail to expose:

https://github.com/Project-OSRM/osrm-backend/blob/e2e279bc852cd024aef1ad445a81bad2884902cd/include/engine/engine_config.hpp#L85-L89

daniel-j-h:
For #4226. Exposes `EngineConfig` limits.

Sidenote: at some point we should re-write our Node.js bindings and auto-generate the v8 parsing code ..

### DIFF
--- a/include/nodejs/node_osrm_support.hpp
+++ b/include/nodejs/node_osrm_support.hpp
@@ -176,6 +176,64 @@ inline engine_config_ptr argumentsToEngineConfig(const Nan::FunctionCallbackInfo
         return engine_config_ptr();
     }
 
+    // Set EngineConfig system-wide limits on construction, if requested
+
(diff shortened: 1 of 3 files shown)


TheMarex:
@daniel-j-h we could try to use https://github.com/pmed/v8pp which works with node addons as well. Not sure how it interacts with NAN?

PR #4289, opened 2017-07-13
Implement usage of abstract interfaces in all routing algorithms

deniskoronchik:
# Issue

Use abstract interfaces in algorithms: #4158 

## Tasklist
 - [ ] review
 - [ ] adjust for comments

## Requirements / Relations

### DIFF
--- a/include/engine/datafacade/algorithm_datafacade.hpp
+++ b/include/engine/datafacade/algorithm_datafacade.hpp
@@ -61,7 +61,7 @@ template <> class AlgorithmDataFacade<CH>
                                     const std::function<bool(EdgeData)> filter) const = 0;
 };
 
-template <> class AlgorithmDataFacade<CoreCH>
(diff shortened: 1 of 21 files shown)


TheMarex:
Thanks for this. However we have consciously avoided passing the base-class everywhere because we think we don't need run-time polymorphism.

I implemented a compile-time approach in https://github.com/Project-OSRM/osrm-backend/pull/4291 I hope this unblocks you.

vng:
Well, #4291 looks good if we will have an opportunity to override some specific behavior in DataFacade<SomeOfflineMLD> data storage. And hope that it will be possible to override any specific algorithm components, leaving the main algorithm functions as-is.

vng:
UPD: I didn't realize where "DataFacade" comes from ... Is it the replacement of ``` template <AlgorithmT> class AlgorithmDataFacade ``` ?

deniskoronchik:
Will prepare new approach pull request

Issue #4150, opened 2017-06-13
How to build OSRM in Ubuntu without sudo priviledge?


humasak:
hi all,

I want to build OSRM in ubuntu server in my local directory. Actually, I do not have sudo privilege. is there anyone have the experience to build OSRM in Ubuntu without sudo privilege? May please share the steps? thank you in advanced.

daniel-j-h:
You don't need `sudo` for building osrm from source. Just follow the steps [here](https://github.com/Project-OSRM/osrm-backend#building-from-source) and leave out the global installation step. Then you can just use the osrm binaries in the build directories.

Or tell CMake to install to a local directory:

    cmake .. -DCMAKE_BUILD_TYPE=Release -DCMAKE_INSTALL_PREFIX=/home/user/osrm

In case you also don't have the dependencies around you need to build them locally before, too.

humasak:
thanks @daniel-j-h 

I already tried. But, when I tried to execute command: 
cmake .. -DCMAKE_BUILD_TYPE=Release -DCMAKE_INSTALL_PREFIX=/home/user/osrm

I got this error: 
-- Building on a 64 bit system
CMake Error at CMakeLists.txt:101 (message):
  GCC>=4.9 required.  In case you are on Ubuntu upgrade via
  ppa:ubuntu-toolchain-r/test
-- Configuring incomplete, errors occurred!

I already installed  GCC 5.2.0 locally. But. i do not know how to solve this error. please help. tks

danpat:
Try doing:

```
export CXX=g++-5
export CC=gcc-5
cmake .. -DCMAKE_BUILD_TYPE=Release -DCMAKE_INSTALL_PREFIX=/home/user/osrm
```

(here, `g++-5` is the name of your GCC 5.2 executable, and it should be on your `PATH` somewhere)

humasak:
@danpat I already tried it, but it still gives the same error.

actually, my .bashrc contain these:
export PATH=~/gcc-5.2.0/bin:$PATH
export LD_LIBRARY_PATH=~/gcc-5.2.0/lib:$LD_LIBRARY_PATH
export LD_LIBRARY_PATH=~/gcc-5.2.0/lib/../lib64:$LD_LIBRARY_PATH

I do not know what it still gives the same error. need help...

danpat:
[...]

PR #4291, opened 2017-07-13
Add abstraction to change the data facade at compile time

TheMarex:
# Issue

Implements issue #4150 using a compile time approach.

## Tasklist
 - [x] review
 - [x] adjust for comments

### DIFF
--- a/include/engine/datafacade_provider.hpp
+++ b/include/engine/datafacade_provider.hpp
@@ -2,6 +2,7 @@
 #define OSRM_ENGINE_DATAFACADE_PROVIDER_HPP
 
 #include "engine/data_watchdog.hpp"
+#include "engine/datafacade.hpp"
 #include "engine/datafacade/contiguous_internalmem_datafacade.hpp"
 #include "engine/datafacade/process_memory_allocator.hpp"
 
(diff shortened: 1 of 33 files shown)


include/engine/datafacade_provider.hpp
```diff
@@ -2,55 +2,85 @@
...
 
-    std::shared_ptr<const FacadeT> Get() const override final { return immutable_data_facade; }
+    std::shared_ptr<const Facade> Get() const override final { return immutable_data_facade; }
 
   private:
-    std::shared_ptr<const FacadeT> immutable_data_facade;
+    std::shared_ptr<const Facade> immutable_data_facade;
+};
+
+template <typename AlgorithmT, template <typename A> class FacadeT>
```
oxidase: it is better to have one templated method here as
```
template <typename AlgorithmT, template <typename A> class FacadeT>
class WatchingProvider : public DataFacadeProvider<AlgorithmT, FacadeT>
{
    DataWatchdog<AlgorithmT> watchdog;
    using Facade = typename DataFacadeProvider<AlgorithmT, FacadeT>::Facade;

  public:
    std::shared_ptr<const Facade> Get() const override final
    {
        return watchdog.Get();
    }
};
```

For some reason clang uses the primary template and gcc specialized, but it is better to remove the ambiguity

oxidase:
@TheMarex just looking around for clang bugs, found a way how to keep the primary template. Looks like clang fails at https://github.com/Project-OSRM/osrm-backend/blob/8c0d61682bd22ae9b5c39b2f102125cb31e0cc89/include/engine/datafacade_provider.hpp#L77-L82

Changing `DataFacade` to explicit type `datafacade::ContiguousInternalMemoryDataFacade` fixes the issue. 

But i would prefer a compilation error if provider will be used with incompatible facade over a runtime exception with message `Not implemented`.

include/engine/datafacade_provider.hpp
```diff
@@ -2,54 +2,69 @@
...
-    using FacadeT = datafacade::ContiguousInternalMemoryDataFacade<AlgorithmT>;
     DataWatchdog<AlgorithmT> watchdog;
 
   public:
-    std::shared_ptr<const FacadeT> Get() const override final
+    using Facade = typename DataFacadeProvider<AlgorithmT, FacadeT>::Facade;
+
+    std::shared_ptr<const Facade> Get() const override final
     {
         // We need a singleton here because multiple instances of DataWatchdog
```
oxidase: This comment is really outdated and confusing, please could you remove it?

Issue #4290, opened 2017-07-13
Don't emit `Notification` instructions for class changes


TheMarex:
As per voice with @bsudekum we decided to not emit `Notification` instructions on `classes` changes.

I would this propose the following change:

Move the `classes` from `RouteStep` to the `Intersection`:

```json
{
    "location":[13.394718,52.543096],
    "in":0,
    "out":2,
    "bearings":[60,150,240,330],
    "entry":["false","true","true","true"],
    "lanes":{
        "indications": ["left", "straight"],
        "valid": "false"
    },
    "classes": ["toll", "highway"]
}
```

daniel-j-h:
Ah, so we will always have the classes available and the client can detect 1/ a new class was added and 2/ an existing class was removed - see https://github.com/Project-OSRM/osrm-backend/pull/4204#issuecomment-313340990.

daniel-j-h:
This landed with https://github.com/Project-OSRM/osrm-backend/issues/4292.

PR #4292, opened 2017-07-14
Move classes to intersection object and don't emit notifications

TheMarex:
# Issue

Implements #4290 by including the class of the outgoing road in the intersection object.

## Tasklist
 - [x] review
 - [x] adjust for comments

### DIFF
--- a/features/car/classes.feature
+++ b/features/car/classes.feature
@@ -17,13 +17,13 @@ Feature: Car - Mode flag
             | cd    | primary |       |
 
         When I route I should get
-            | from | to | route       | turns                                              | classes      |
-            | a    | d  | ab,bc,cd,cd | depart,notification right,notification left,arrive | ,ferry,,     |
-            | d    | a  | cd,bc,ab,ab | depart,notification right,notification left,arrive | ,ferry,,     |
-            | c    | a  | bc,ab,ab    | depart,notification left,arrive                    | ferry,,      |
-            | d    | b  | cd,bc,bc    | depart,notification right,arrive                   | ,ferry,ferry |
-            | a    | c  | ab,bc,bc    | depart,notification right,arrive                   | ,ferry,ferry |
-            | b    | d  | bc,cd,cd    | depart,notification left,arrive                    | ferry,,      |
+            | from | to | route       | turns                                              | classes                  |
+            | a    | d  | ab,bc,cd,cd | depart,notification right,notification left,arrive | [()],[(ferry)],[()],[()] |
+            | d    | a  | cd,bc,ab,ab | depart,notification right,notification left,arrive | [()],[(ferry)],[()],[()] |
+            | c    | a  | bc,ab,ab    | depart,notification left,arrive                    | [(ferry)],[()],[()]      |
+            | d    | b  | cd,bc,bc    | depart,notification right,arrive                   | [()],[(ferry)],[()]      |
+            | a    | c  | ab,bc,bc    | depart,notification right,arrive                   | [()],[(ferry)],[()]      |
+            | b    | d  | bc,cd,cd    | depart,notification left,arrive                    | [(ferry)],[()],[()]      |
 
 
     Scenario: Car - We tag motorways with a class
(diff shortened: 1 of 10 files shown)


features/car/classes.feature
```diff
@@ -17,13 +17,13 @@ Feature: Car - Mode flag
...
         When I route I should get
-            | from | to | route       | turns                                              | classes      |
-            | a    | d  | ab,bc,cd,cd | depart,notification right,notification left,arrive | ,ferry,,     |
-            | d    | a  | cd,bc,ab,ab | depart,notification right,notification left,arrive | ,ferry,,     |
-            | c    | a  | bc,ab,ab    | depart,notification left,arrive                    | ferry,,      |
-            | d    | b  | cd,bc,bc    | depart,notification right,arrive                   | ,ferry,ferry |
-            | a    | c  | ab,bc,bc    | depart,notification right,arrive                   | ,ferry,ferry |
-            | b    | d  | bc,cd,cd    | depart,notification left,arrive                    | ferry,,      |
+            | from | to | route       | turns                                              | classes                  |
+            | a    | d  | ab,bc,cd,cd | depart,notification right,notification left,arrive | [[]],[[ferry]],[[]],[[]] |
```
MoKob: I like this `[]` approach. We should think about doing this on all outputs for intersections.

include/engine/guidance/assemble_steps.hpp
```diff
@@ -119,6 +120,8 @@ inline std::vector<RouteStep> assembleSteps(const datafacade::BaseDataFacade &fa
                 const auto exits = facade.GetExitsForID(step_name_id);
                 const auto distance = leg_geometry.segment_distances[segment_index];
                 auto classes = facade.GetClasses(path_point.classes);
+                // intersections contain the classes of exiting road
+                intersection.classes = std::move(classes);
```
MoKob: just a personal preference: I'd prefer not assigning a name to `classes` if we only use it to transfer the data. i.e. `intersection.classes = face.GetClasses()`.

include/engine/guidance/assemble_steps.hpp
```diff
@@ -196,6 +198,8 @@ inline std::vector<RouteStep> assembleSteps(const datafacade::BaseDataFacade &fa
         const auto distance = leg_geometry.segment_distances[segment_index];
         const EdgeWeight duration = segment_duration + target_duration;
         const EdgeWeight weight = segment_weight + target_weight;
+        // intersections contain the classes of exiting road
+        intersection.classes = std::move(target_classes);
```
MoKob: same here.

MoKob:
darn. pressed that button to early: the approach here seems fine, I am unsure about some stuff for collapsing, though. I am going to collect a few links and post here.

features/car/classes.feature
```diff
@@ -96,6 +96,6 @@ Feature: Car - Mode flag
             | cd    | primary |          |
 
         When I route I should get
-            | from | to | route    | turns                            | classes             |
-            | a    | d  | ab,bc,cd | depart,notification right,arrive | toll,motorway;toll, |
+            | from | to | route    | turns         | classes                          |
+            | a    | d  | ab,cd    | depart,arrive | [[toll],[motorway,toll],[]],[[]] |
```
oxidase: Just minor:  spaces after comma increase readability, 
```
[[toll],[motorway,toll],[]],[[]]
[[toll],[motorway,toll],[]], [[]]
[[toll], [motorway,toll], []], [[]]
```

TheMarex:
@MoKob added a test for the case you mentioned, seems to be working as expected.

I don't really understand what is going on at the position that you linked. Could you please outline a test case you would want to see for testing correct behavior?

features/car/classes.feature
TheMarex: Changed to round braces for a similar effect. 👍

MoKob:
@TheMarex I went through the entire code. It seems all locations correctly transfer the intersections. So it is more a question of what we want in the output. If it is alright that the intersections don't line up with announcements, we should be fine.

E.g. if you collapse a turn, at some points we transfer signage (names/destinations...), since the second part is the actual turn. It looks like we keep all intersections intact. So the classes refer to the exact location the intersection is in the data, which would be fine with me. We wouldn't return a motorway class in the exact step we turn onto a motorway in some cases, or if we turn onto toll road with a sliproad, the turn itself will not feature `toll` but only the second intersection will.

In total I would consider the `keep it where it is` method preferable, since we don't mess with the data and are less likely to produce errors.

PR #4293, opened 2017-07-14
Only enables -fopenmp in case the user wants stxxl

daniel-j-h:
stxxl is optional now (enabled by default) - only set the `-fopenmp` flag if there is a need for it


TheMarex:
@daniel-j-h can you cherry pick this to the `5.9` branch after merging?

daniel-j-h:
Yep, done.

PR #4296, opened 2017-07-17
Trigger lane anticipation based on distance, see discussion in #4260

daniel-j-h:
See discussion in https://github.com/Project-OSRM/osrm-backend/pull/4268: the concern was raised that our traffic speeds change and we don't want to be speed-dependent here. Triggering lane anticipation based on distance is not perfect either, but there's not much we can do without congestion / flow along the route.

The changeset here assums the user needs 200 meters for a single lane change; in case there are more lanes to the left or to the right we scale up the threshold linearly (think: crossing two lanes: 400 m, three lanes 600 m and so on).

Note: tests adapted to distance and then shrinked some scenarios down in cell sizes to make them realistic again.

@MoKob this targets your `deprecate/use-lane` - please merge into it.

cc @willwhite

### DIFF
--- a/features/guidance/anticipate-lanes.feature
+++ b/features/guidance/anticipate-lanes.feature
@@ -3,7 +3,7 @@ Feature: Turn Lane Guidance
 
     Background:
         Given the profile "car"
-        Given a grid size of 20 meters
+        Given a grid size of 100 meters
 
     @anticipate
     Scenario: Anticipate Lane Change for subsequent multi-lane intersections
@@ -371,9 +371,9 @@ Feature: Turn Lane Guidance
        Scenario: Anticipate Lanes for through with turn before / after
            Given the node map
                """
-               c       g       l
-               b – d – e – h - i
-               a       f       j
+               c   g   l
+               b d e h i
+               a   f   j
                """
 
            And the ways
@@ -625,38 +625,9 @@ Feature: Turn Lane Guidance
             x – b   d – y
                 |   |
                 |   |
-                |   |
-                |   |
-                |   |
-                |   |
-                |   |
-                |   |
-                |   |
-                |   |
                  | |
                  | |
-                 | |
-                 | |
-                 | |
-                 | |
-                 | |
-                 | |
-                 | |
-                 | |
-                 | |
-                 | |
-                  |
-                  |
-                  |
-                  |
-                  |
-                  |
-                  |
-                  |
-                  |
-                  |
-                  |
-                  |
+                 \ /
                   c
             """
 
@@ -792,13 +763,6 @@ Feature: Turn Lane Guidance
                 |
                 |
                 |
-                |
-                |
-                |
-                |
-                |
-                |
-                |
                 c
                 |
             e – d – y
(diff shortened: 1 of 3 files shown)


features/guidance/anticipate-lanes.feature
```diff
@@ -623,40 +623,7 @@ Feature: Turn Lane Guidance
...
-                  |
-                  |
-                  |
-                  |
-                  |
-                  |
-                  |
-                  |
-                  |
+                 \ /
```
MoKob: can you give a bit more context for the test-changes? The reduction in distance is quite severe in a those two. Do we do much less lane anticipation now?

daniel-j-h: I increased the grid size above; and with a grid size this high these scenarios no longer model reality.

MoKob: Well, here we still reduce from 640 meters to 200 meters, below from 200 to 150 meters

TheMarex:
@daniel-j-h @MoKob Is this good to merge into `deprecate/use-lane`?

emiltin:
for guidance to make sense for other modes that car, such distance should not be hardcoded, but instead defined by the profile.

daniel-j-h:
Good point - our guidance targets car navigation at the moment; not sure how bike lanes are tagged and if lane anticipation even makes sense for it :P

emiltin:
lane guidance is obviously more imporant for motorvehicle. but you might want different settings for cars, trucks, etc.

it's a good question whether guidance is needed for bikes. cyclists usually get by without lane guidance. on the other hand we do have lane markings for bikes in copenhagen. a few examples:

https://www.google.dk/maps/@55.6819133,12.5576559,3a,75y,312.99h,62.24t/data=!3m6!1e1!3m4!1s7Huqfa6_KHCUZolDjl9vFg!2e0!7i13312!8i6656

https://www.google.dk/maps/@55.6815094,12.5606516,3a,60y,254.24h,68.72t/data=!3m6!1e1!3m4!1szIDDB8bfbNX2rurFBe92pQ!2e0!7i13312!8i6656

https://www.google.dk/maps/@55.7134175,12.5597584,3a,60y,78.63h,77.83t/data=!3m6!1e1!3m4!1srJuVhsRExlWdEPrXpqyubg!2e0!7i13312!8i6656

https://www.google.dk/maps/@55.6910425,12.5707972,3a,46.8y,159.17h,85.87t/data=!3m6!1e1!3m4!1sOnlexNdqiP93lloicPTUHg!2e0!7i13312!8i6656

emiltin:
in general, bike navigation is often complicated. check out the bike lanes in this intersection: https://www.google.dk/maps/@55.7130524,12.5600254,39a,59.8y,359.82h,31.74t/data=!3m1!1e3.

here you need to keep right if you want to turn left (blue painted lanes)
https://www.google.dk/maps/@55.6810592,12.5605579,105a,35y,299.37h/data=!3m1!1e3

emiltin:
to you question above i have not seen bike lanes tagged in osm yet. looking at http://wiki.openstreetmap.org/wiki/Lanes it not even clear to me if you can combine `:lanes` with a mode specifier. can you do turn:bicycle:lanes=left|straight?

emiltin:
i can add that we now have lane directions for bikes in many places intersections, e.g: https://thenounproject.com, however it's not really linked to lanes painted on the road. maybe an idea for out planners...

Issue #4298, opened 2017-07-17
No ability to disable logging output


mloskot:
There is no way to disable the logger (`osrm::util::Log class`) output sent to stdout/stderr.

This is quite inconvenient if OSRM is used by a client as library.

mloskot:
Logging verbosity can now controlled.
Implemented by #4451

PR #4299, opened 2017-07-17
Enable turning off logging output

mloskot:
# Issue

#4298

This PR addresses the issue allowing to control logging output:

* Adds CMake option: `ENABLE_LOGGING` (default ON; if OFF, then `osrm::util::LogPolicy` forces to mute logs).
* Adds null/no-op stream buffer pluggable into output stream used with/passed to instance of `osrm::util::Log`.

at runtime (limited, eg.  `Extractor::run` unmutes logger) or compile time (overcomes the previous limitation mentioned, forces muted logger).

## Tasklist
 - [ ] review

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -19,6 +19,7 @@ option(ENABLE_CCACHE "Speed up incremental rebuilds via ccache" ON)
 option(BUILD_TOOLS "Build OSRM tools" OFF)
 option(BUILD_PACKAGE "Build OSRM package" OFF)
 option(ENABLE_ASSERTIONS "Use assertions in release mode" OFF)
+option(ENABLE_LOGGING "Output log messages" ON)
 option(ENABLE_COVERAGE "Build with coverage instrumentalisation" OFF)
 option(ENABLE_SANITIZER "Use memory sanitizer for Debug build" OFF)
 option(ENABLE_STXXL "Use STXXL library" ON)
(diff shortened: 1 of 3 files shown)


oxidase:
@mloskot thanks for PR! I would suggest to make a proper runtime verbosity command line option  like `--verbosity,-v` with arguments `{NONE,ERROR,WARNING,INFO,DEBUG}` in the following steps:
* reorder [log levels](https://github.com/Project-OSRM/osrm-backend/blob/8c0d61682bd22ae9b5c39b2f102125cb31e0cc89/include/util/log.hpp#L8-L14) and add  `NONE` level
* propagation of command line option to `LogPolicy`  in addition to [`m_is_mute`](https://github.com/Project-OSRM/osrm-backend/blob/8c0d61682bd22ae9b5c39b2f102125cb31e0cc89/include/util/log.hpp#L37)
* add a check [`&& level <= LogPolicy::GetInstance().GetLevel()`](https://github.com/Project-OSRM/osrm-backend/blob/8c0d61682bd22ae9b5c39b2f102125cb31e0cc89/src/util/log.cpp#L80)

mloskot:
@oxidase With this first stab at the idea, I've aimed with the minimal changes required and I'm probing the interest.

I agree with your suggestion to make it properly. I will try to work out necessary changes and update the PR (or open new one).

TheMarex:
This was replaced by https://github.com/Project-OSRM/osrm-backend/pull/4451

Issue #4302, opened 2017-07-18
Error on missing `tzdata` dependency when conditional restriction options are used


karenzshea:
Trying to use the conditional restriction functionality in OSRM without the `tzdata` dependency installed results in some very ambiguous behaviour. The default fallback on linux machines seems to be to interpret everything as UTC, meaning that the functionality still works, but not at all as expected.
cc @TheMarex @oxidase

oxidase:
We have update system requirements with a phrase about time zones data. To have proper time-dependent conditional tags `tzdata` is required.

PR #4303, opened 2017-07-18
Throw if Timezoner can not convert a sample timestamp to local time

oxidase:
# Issue

Conditional turn restriction are wrong if `localtime` fallsback to UTC #4302. 
The PR check a sample timestamp conversion to EST timzone and throws an exception if conversion fails.
The check must be at runtime as `tzdata` is not required to be installed on a build server.

/cc @karenzshea 

## Tasklist
 - [ ] review
 - [ ] adjust for comments

### DIFF
--- a/src/util/timezones.cpp
+++ b/src/util/timezones.cpp
@@ -26,6 +26,24 @@ namespace osrm
 namespace updater
 {
 
+namespace
+{
+inline struct tm getLocalTimeinfo(const char *tzname, const std::time_t utc_time)
+{ // Thread safety: MT-Unsafe const:env
+    struct tm timeinfo;
+#if defined(_WIN32)
+    _putenv_s("TZ", tzname);
+    _tzset();
+    localtime_s(&timeinfo, &utc_time);
+#else
+    setenv("TZ", tzname, 1);
+    tzset();
+    localtime_r(&utc_time, &timeinfo);
+#endif
+    return timeinfo;
+}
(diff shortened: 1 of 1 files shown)


src/util/timezones.cpp
```diff
@@ -39,6 +57,13 @@ Timezoner::Timezoner(const char geojson[], std::time_t utc_time_now)
                                     std::to_string(code) + " malformed at offset " +
                                     std::to_string(offset));
     }
+
+    // Run-time check for time zones data: 1500000000 ↔ 'Thu Jul 13 22:40:00 EDT 2017'
+    if (getLocalTimeinfo("EST+5EDT", 1500000000).tm_hour != 22)
```
karenzshea: Could the check also just see if the returned time is the same as the given one?

oxidase: @karenzshea Rethinking about the check: better to move it out of OSRM codebase and document only  `tzdata` requirement and making check as
`[ $(TZ=EST+5EDT date -d @1500000000 +%H) == 22 ]  || echo "tzdata is not installed"`
once before the first time updater run.

Or as a full check
`[ $(TZ=EST+5EDT date -d @1500000000 +%Y%m%d%H%M%S) == 20170713224000 ] || echo "tzdata is not installed"`

oxidase:
Closed as tzdata check is a part of OS requirements

PR #4304, opened 2017-07-18
Update third party dependencies

TheMarex:
# Issue

Primary reason for this is a new sol2 version that supports `sol::nested`.

## Tasklist
 - [x] review
 - [x] adjust for comments


TheMarex:
Seems like the update of `Sol2` is causing `SIGSEGV` on multiple `mason + clang` builds. I'll try to reproduce locally.

daniel-j-h:
Seeing [`sol::nested`](http://sol2.readthedocs.io/en/latest/api/nested.html) - can you give some context where this is needed?

TheMarex:
Added an older version of sol `2.17.5` which should not have the global state problem. Hopefully this will just work.

@daniel-j-h `sol::nested` would be useful for doing the following thing:

```lua
local set_of_sets = Set {
 Set { "foo", "bar"},
 Set { "bla" }
}
```

And mapping this to a C++ `std::vector<std::vector<std::string>>`.

PR #4306, opened 2017-07-19
Makes construction=minor ways routable again, see #4258

daniel-j-h:
In #4258 we removed construction tags except for `construction=widening` and `construction=no`.

Turns out we missed `construction=minor` in the list of 3804 unique construction values.

This affects 1194 ways on the planet.

### DIFF
--- a/profiles/foot.lua
+++ b/profiles/foot.lua
@@ -57,6 +57,8 @@ function setup()
 
     restricted_highway_whitelist = Set { },
 
+    construction_whitelist = Set {},
+
     access_tags_hierarchy = Sequence {
       'foot',
       'access'
--- a/profiles/foot.lua
+++ b/profiles/foot.lua
@@ -137,7 +137,7 @@ function setup()
   }
 end
 
-function process_node (profile, node, result)
+function process_node(profile, node, result)
   -- parse access and barrier tags
   local access = find_access_tag(node, profile.access_tags_hierarchy)
   if access then
--- a/profiles/lib/way_handlers.lua
+++ b/profiles/lib/way_handlers.lua
@@ -521,7 +521,7 @@ function WayHandlers.blocked_ways(profile,way,result,data)
     local construction = way:get_value_by_key('construction')
 
     -- Of course there are negative tags to handle, too
-    if construction and construction ~= 'no' and construction ~= 'widening' and construction ~= 'minor' then
+    if construction and not profile.construction_whitelist[construction] then
       return false
     end
   end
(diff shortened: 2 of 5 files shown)


profiles/lib/way_handlers.lua
```diff
@@ -521,7 +521,7 @@ function WayHandlers.blocked_ways(profile,way,result,data)
     local construction = way:get_value_by_key('construction')
 
     -- Of course there are negative tags to handle, too
-    if construction and construction ~= 'no' and construction ~= 'widening' then
+    if construction and construction ~= 'no' and construction ~= 'widening' and construction ~= 'minor' then
```
MoKob: how about doing an explicit check for the tags we wan't to block from routing? Otherwise we might just end up excluding another strange tag we might not yet even know about/that is introduced in the future.

daniel-j-h: In general this is a good idea in practice osm allows for

> The value of the construction=\* key should be one of those defined bellow and indicates the type of feature being built (you can use any value from from highway=\*, railway=\* and building=\*).

so unless we want to hard-code all of those values I think we're good for now.

MoKob: Also from the OSM wiki as a side-note. 

```
Unclassified road under construction (alternatively tag construction=minor)
```

MoKob: We could check for `https://github.com/Project-OSRM/osrm-backend/blob/master/profiles/car.lua#L143-L156` or some similar list, right? That would avoid hardcoding more strings there.

MoKob: Capturing from chat: I was under the misconception that this tagging scheme would try to parse `highway=construction and construction=highway_type`. I am fine with this change for the special cases to handle deprecated tags.

TheMarex: What would be cool however is a whitelist in the profile table. Then you could do:

```lua
if construction and not profile.construction_whitelist[construction] then
  ...
end
```

daniel-j-h:
Updated - please have a look again.

profiles/foot.lua
```diff
@@ -57,6 +57,8 @@ function setup()
 
     restricted_highway_whitelist = Set { },
 
+    construction_whitelist = Set {},
```
emiltin: probably not needed, since avoid construction is not activated?

daniel-j-h:
Ah, thanks for cherry picking and running this out 🙇

PR #4310, opened 2017-07-20
fix profile debugging related to way classes

emiltin:
# Issue

profile debugger was broken, because forward_classes and backward_classes was not initialized with empty tables.


daniel-j-h:
👍 thanks for fixing, I will cherry-pick this into the 5.9 branch, otherwise debugging will be broken with 5.9 and the default profiles.

daniel-j-h:
Also do you think we could test the debugger code path on Travis so catch issues like this early on in the future?

emiltin:
yes we could add cucumber support code to run profile debugger case

PR #4314, opened 2017-07-20
refactor the bike profile, other fixes

emiltin:
# Issue
the bike profile still has som long functions with old code that's hard to read. this PR is an efford to clean it up.

## Tasklist
 - [x] use foot profile to determine where you can push bikes
 - [x] add tests to check improved push handling
 - [x] add missing suffix list in bike profile
 - [ ] use existing WayHandlers when possible
 - [ ] factor remaining code out into manageable functions 
 - [ ] review
 - [ ] adjust for comments

Since this refactor touched the WayHandlers and thus other profiles, it includes a few other improvements as well:

- control what tag combinations imply oneway by setting profile.implied_oneways
- tag helper to support the above
- add missing test that circular is oneway in the car profile
- add missing support for public_transport=platform in foot profile
- add helper functions in WayHandlers for initializing and merging way result tables (used when using the foot profile in the bike profile)

the test in features/guidance/bicycle-sliproads.feature b/features/guidance/bicycle-sliproads.feature was failing, i believe this was due to an incorrect oneway direction of the slipway eb, so i changed this. is this correct?

-

### DIFF
--- a/features/bicycle/cycleway.feature
+++ b/features/bicycle/cycleway.feature
@@ -85,13 +85,21 @@ Feature: Bike - Cycle tracks/lanes
             | highway     | cycleway | oneway | forw    | backw        |
             | motorway    | track    | yes    | cycling |              |
             | residential | track    | yes    | cycling | pushing bike |
-            | cycleway    | track    | yes    | cycling | pushing bike |
+            | cycleway    | track    | yes    | cycling |              |
             | footway     | track    | yes    | cycling | pushing bike |
 
     Scenario: Bike - Cycleway on oneways, speeds
         Then routability should be
             | highway     | cycleway | oneway | forw    | backw      |
             | motorway    | track    | yes    | 15 km/h |            |
-            | residential | track    | yes    | 15 km/h | 6 km/h +-1 |
-            | cycleway    | track    | yes    | 15 km/h | 6 km/h +-1 |
-            | footway     | track    | yes    | 15 km/h | 6 km/h +-1 |
+            | residential | track    | yes    | 15 km/h | 5 km/h +-1 |
+            | cycleway    | track    | yes    | 15 km/h |            |
+            | footway     | track    | yes    | 15 km/h | 5 km/h +-1 |
+
+    Scenario: Bike - Directional sideways and oneway
+        Then routability should be
+            | highway     | oneway | cycleway:right | cycleway:left | forw         | backw        |
+            | primary     | yes    | track          |               | cycling      | pushing bike |
+            | primary     | yes    |                | track         | cycling      | cycling      |
+            | primary     | -1     | track          |               | cycling      | cycling      |
+            | primary     | -1     |                | track         | pushing bike | cycling      |
--- a/features/bicycle/cycleway.feature
+++ b/features/bicycle/cycleway.feature
@@ -7,78 +7,77 @@ Feature: Bike - Cycle tracks/lanes
 
     Scenario: Bike - Cycle tracks/lanes should enable biking
         Then routability should be
-            | highway     | cycleway     | forw | backw |
-            | motorway    |              |      |       |
-            | motorway    | track        | x    |       |
-            | motorway    | lane         | x    |       |
-            | motorway    | shared       | x    |       |
-            | motorway    | share_busway | x    |       |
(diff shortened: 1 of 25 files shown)


emiltin:
seem i removed the test for leisure=track by accident

features/bicycle/cycleway.feature
```diff
@@ -85,13 +85,21 @@ Feature: Bike - Cycle tracks/lanes
             | highway     | cycleway | oneway | forw    | backw        |
             | motorway    | track    | yes    | cycling |              |
             | residential | track    | yes    | cycling | pushing bike |
-            | cycleway    | track    | yes    | cycling | pushing bike |
+            | cycleway    | track    | yes    | cycling |              |
```
emiltin: i don't think a oneway cycleway should allow pushing bikes in the reveres direction

features/bicycle/way.feature
```diff
@@ -9,31 +9,31 @@ Feature: Bike - Accessability of different way types
...
+            | living_street  | cycling      |
+            | road           | cycling      |
+            | track          | cycling      |
+            | path           | cycling      |
+            | footway        | pushing bike |
+            | pedestrian     | pushing bike |
+            | steps          | pushing bike |
+            | cycleway       | cycling      |
+            | bridleway      |              |
+            | pier           | pushing bike |
```
emiltin: pushing bikes on piers was not support before

features/car/oneway.feature
```diff
@@ -40,6 +42,9 @@ Feature: Car - Oneway streets
             | motorway        | roundabout | x    |       |                       |
             | motorway_link   | roundabout | x    |       |                       |
             | primary         | roundabout | x    |       |                       |
+            | motorway        | circular   | x    |       |                       |
+            | motorway_link   | circular   | x    |       |                       |
+            | primary         | circular   | x    |       |                       |
```
emiltin: test for circular was missing

features/guidance/bicycle-sliproads.feature
```diff
@@ -24,9 +24,9 @@ Feature: Bike - Mode flag
         And the ways
             | nodes | highway   | name   | oneway:bicycle |
             | abcd  | cycleway  | street |                |
-            | eb    | path      |        | yes            |
+            | be    | path      |        | yes            |
```
emiltin: was there a reason the slipway was oneone in the opposite direction? due to small changes in the speed on pushing bikes, the test was now failing. swtiching the oneway direction made the tests pass again

profiles/lib/way_handlers.lua
```diff
@@ -11,6 +11,35 @@ local Tags = require('lib/tags')
...
 WayHandlers = {}
 
+function WayHandlers.new_result()
+  return {
+    road_classification = {},
+    forward_speed = -1,
+    backward_speed = -1,
+    forward_mode = 0,
+    backward_mode = 0,
+    duration = 0
```
emiltin: hmm forward_classes = {} and backward_classes= = {} is missing here i see

emiltin:
this PR supercedes #4033

emiltin:
need to work more on how to best integrate foot processing into the bike profile. running the full foot processing chain might not be the best option. a lot of things is already handled, but more importantly, the result of the bike processing sometimes has to influence the foot processing. e.g. bicycle=no means you cannot push a bike.

emiltin:
here's a new approach to better handling of pushing of bikes in the bike profile. instead of manually checking for walking (which is error prone), or running the entire foot profile (which is a waste of processing and has other issues), the bike profile now loads the foot profile settings table and uses it to tall calls the required processing steps:

https://github.com/Project-OSRM/osrm-backend/blob/refactor/bike_profile_cleanup/profiles/bicycle.lua#L380

this seems to work well.

however this PR has otherwise become a bit unruly with changes to too many files, so probably need simplification. also a few car tests are failing.

emiltin:
a change that me moved to a different PR is a more unified handling of routes (ferry, train, shuttle_train, etc), with better handling of which routes require specific access tags:

https://github.com/Project-OSRM/osrm-backend/blob/refactor/bike_profile_cleanup/profiles/bicycle.lua#L152

emiltin:
going to close this PR and split it into smaller ones

PR #4315, opened 2017-07-20
Implement excluding certain classes

TheMarex:
# Issue

This PR implements https://github.com/Project-OSRM/osrm-backend/issues/4006

## Tasklist
 - [x] Parse and store avoid combinations in `osrm-extract`
 - [x] Implement `avoid` parameter for `RouteParameters` (C++, HTTP)
 - [x] Split out metric from `CellStorage` and allow storage of multiple metrics per file
 - [x] Compute separate overlays for each avoid flag
 - [x] Add `Datafacade` factory that setups facades for each avoid flag and selects on query
 - [x] Extend query logic to check if the avoid mask matches each outgoing node before relaxing
 - [x] Don't snap to avoided nodes.
 - [x] Write tests
 - [x] Proper error messages for unknown avoid class names: Needs change in the profile API
 - [x] Throw error for algorithms that don't implement avoid flags
 - [x] ~~Debug Boost::Spirit sanitizer error~~ Gone?
 - [x] Measure memory file size impact
 - [x] Write documentation 
 - [ ] Review
 - [ ] Adjust for comments

## Measurements

|Customization/seconds |`master`|This branch (3 exclude combinations)|Increase|
|---------|------------|-----------|-------|
|Denmark| 1.59 |  6.84 |  x4.3 |
|Germany| 12.89  | 54.178 | x4.2  |

|Total `osrm-routed` data/MB|`master`|This branch (3 exclude combinations)|Increase|
|-------------|------------|-----------|----------|
|Denmark|  338.98 |  430.64 |  x1.27 |
|Germany|  3529.83 | 4017.39 |  x1.13  |

| Average Query/ms | `master` | This branch | This branch `exclude=motorway` |
|----------------------|-------------|------------------|-----------------------------------------------|
| Denmark          |   4.95            | 5.69 (x1.20) |   N/A                                       |
| Germany          |   12.37   | 14.20 (x1.14) | 21.82 (x1.76)                               |

### Excluding motorways on Berlin:

No avoidance
with motorway

`exclude=motorway`
without motorway

### Ferries in Denmark

`exclude=ferry`
avoid ferry


emiltin:
i'm not sure i understand the combinations.
when processing ways, the profile marks the classes that a way belongs to, say highway and ferry. if you then want to avoid both ferries and highways, it seems you need to have 3 combinations: {'highway'} and {'ferry'} but also {'ferry','highway'}?

TheMarex:
> i'm not sure i understand the combinations.

Instead of automatically generating data to support all possible combinations (which would be exponential in the number of classes), we allow users to specify which combinations should be allowed.

For example it might make sense to support `ferry, motorway` + `toll` + `ferry + toll` but not `ferry, motorway, toll`. This enables us to be very selective here since every combination that can be avoided has a real impact on memory consumption.

emiltin:
do you mean that the profile lists combinations of classes that can be avoided at query time?

TheMarex:
> do you mean that the profile lists combinations of classes that can be avoided at query time?

Yes. `Set {"motorway", "ferry"}` => `avoid=motorway,ferry`.

emiltin:
would it avoid ways with either the motorway or the ferry class, or only ways with both?

profiles/car.lua
```diff
@@ -101,6 +101,17 @@ function setup()
...
+    -- classes to support for exclude flags
+    excludable = Sequence {
+        Set {'toll'},
+        Set {'motorway'},
+        Set {'ferry'}
+    },
+
     avoid = Set {
       'area',
       -- 'toll',    -- uncomment this to avoid tolls
```
oxidase: this is confusing with `toll` in `excludable` now

profiles/testbot.lua
```diff
@@ -19,6 +19,14 @@ function setup()
       use_turn_restrictions         = true
     },
 
+    classes = {[0] = "motorway", [1] = "toll"},
```
oxidase: is there any benefit to use keys here and below? `{ "motorway", "toll"}` should be ok

profiles/testbot.lua
```diff
@@ -19,6 +19,14 @@ function setup()
...
+
+    excludable = {
+        [1] = {["motorway"] = true},
+        [2] = {["toll"] = true},
+        [3] = {["motorway"] = true, ["toll"] = true}
```
oxidase: is `= true` really needed? I think also it is better to use plane tables as
```
    excludable = {
        {'motorway'},
        {'toll'},
        {'motorway', 'toll'}
    },
```

It will require only change `first` -> `second` at 
https://github.com/Project-OSRM/osrm-backend/blob/936539ca8a8810e9cf8ebbbd7c7e788200b87b22/src/extractor/scripting_environment_lua.cpp#L722

emiltin: our lua helpers Set and Sequence can also be used

TheMarex: Turns out the `[1] = "foo", [2] = "bar"` syntax is the same as `{'foo', 'bar'}`. In the other case I would like to make these Sets so you can have duplicate keys in there and order does not matter. I could have used the wrapper but didn't want to include another module in `testbot.lua` just for one line.

TheMarex:
@oxidase Thanks for the review! I hope I addressed all PR comments, can you take a look again?

afeezaziz:
Sorry to ask but is this exclude class function can be extended to exclude OSM node also? For example, there might be an opportunity for us to avoid route vehicles in flood prone areas.

danpat:
@afeezaziz No, this feature can't be used that way.  However, if you can identify the nodes where you want to prevent access (either by tags on the node, or by the ID), you can add logic to the `process_node` function in `car.lua`, and set `result.barrier = true` for that node - this will prevent routes passing over that point.

It won't enable query-time selection like the `exclude` feature however.

apfz:
@danpat is there currently a way in OSRM to avoid certain areas (perhaps through a polygon area) on query time?

danpat:
@apfz No, this is currently not possible.

BastiR9:
Any idea, when this feature will be available?

turing95:
Major cities in the world all have some sort of Low Emission Zones, where vehicles can't drive through. If a routing engine does not consider them,or it doesn't provide a way to exclude some areas, it's not useful.

frodrigo:
> Major cities in the world all have some sort of Low Emission Zones, where vehicles can't drive through. If a routing engine does not consider them,or it doesn't provide a way to exclude some areas, it's not useful.

It is not available as is by default. But can be done by flagging ways on this area at data processing step. You have to make a custom profile.

PR #4319, opened 2017-07-21
Changes to reuse algorithms with custom data facade.

vng:
This changes are needed for full support of custom (offline) data facade. According to our latest agreement, we use data facade template specialization approach.

TODO: Need to factor out alternative, direct, many-to-many algorithms to hpp files to use them in offline project. Also need to make this and depend functions as template<Algorithm>.

To discuss: Actually, the approach with defining OSRM_EXTERNAL_MEMORY is ok, but I think that may be a better solution. As I did with SearchDataEngine - it should also be specialized for the offline purpose. It's enough for template algorithms to have only forward declarations:

```
template <Algorithm> class SearchEngineData;
template <Algorithm> class DataFacade;
...
template <Algorithm> shortestPath(SearchEngineData<Algorithm> & heap, DataFacade<Algorithm> & facade, ...) 
{
 ... 
}
```

See the last commit for more details.

@TheMarex @oxidase


include/engine/plugins/match.hpp
```diff
@@ -2,12 +2,10 @@
 #define MATCH_HPP
 
 #include "engine/api/match_parameters.hpp"
+#include "engine/datafacade/contiguous_internalmem_datafacade.hpp"
```
vng: Ok.

include/engine/datafacade/contiguous_internalmem_datafacade.hpp
```diff
@@ -1129,6 +1129,27 @@ class ContiguousInternalMemoryDataFacade<MLD> final
     {
     }
 };
+
+template <typename Callback>
+void ForEachSourceNodes(ContiguousInternalMemoryDataFacade<MLD> const & facade,
```
vng: It was introduced to be overloaded for the offline CompressedDataFacade. Moving it to routing_base_mld.hpp will cause using DataFacade, but it depends with preprocessor option. Can make a separate header for this functions.

include/engine/phantom_node.hpp
```diff
@@ -167,8 +167,8 @@ struct PhantomNode
     util::Coordinate location;
     util::Coordinate input_location;
     unsigned short fwd_segment_position;
+
     // is phantom node valid to be used as source or target
-  private:
```
vng: To have access to is_valid_* flags. Current ctor is not convenient for us to set this flags. If it's important, we can leave them private, but add some modify functions.

include/engine/engine.hpp
```diff
@@ -18,6 +18,8 @@
 #include "engine/plugins/trip.hpp"
 #include "engine/plugins/viaroute.hpp"
 #include "engine/routing_algorithms.hpp"
+#include "engine/routing_algorithms/routing_base_ch.hpp"
```
vng: Ok.

include/engine/routing_algorithms/routing_base_mld.hpp
```diff
@@ -73,8 +74,8 @@ inline bool checkParentCellRestriction(CellID cell, LevelID, CellID parent)
 using PackedEdge = std::tuple</*from*/ NodeID, /*to*/ NodeID, /*from_clique_arc*/ bool>;
 using PackedPath = std::vector<PackedEdge>;
 
-template <bool DIRECTION, typename OutIter>
-inline void retrievePackedPathFromSingleHeap(const SearchEngineData<Algorithm>::QueryHeap &heap,
+template <bool DIRECTION, typename Algorithm, typename OutIter>
+inline void retrievePackedPathFromSingleHeap(const typename SearchEngineData<Algorithm>::QueryHeap &heap,
```
vng: Yes. But also, we need to skip include tss with that way. I thought that SearchEngineData template specialization will be better ...

include/engine/routing_algorithms/shortest_path.hpp
```diff
@@ -2,22 +2,464 @@
...
+#include <memory>
+
 namespace osrm
 {
 namespace engine
 {
 namespace routing_algorithms
 {
 
+namespace
```
vng: Hm. With this way, we need to put all external/offline source code into osrm repo, but it's not allowed for us. Or make any other code hacks ...

include/engine/search_engine_data.hpp
```diff
@@ -70,27 +72,6 @@ struct SearchEngineData<routing_algorithms::corech::Algorithm>
 {
 };
 
-struct MultiLayerDijkstraHeapData
```
vng: Yes, the reason is to specialize SearchEngineData.

include/partition/multi_level_partition.hpp
```diff
@@ -141,6 +141,10 @@ template <storage::Ownership Ownership> class MultiLevelPartitionImpl final
     friend void serialization::write<Ownership>(storage::io::FileWriter &writer,
                                                 const MultiLevelPartitionImpl &mlp);
 
+    // Used in compression tests only.
+    auto const & GetPartitionVector() const { return partition; }
+    auto const & GetLevelData() const { return *level_data; }
```
vng: Ok.

vng:
On the current step, we need to resolve 2 main problems:
- use template specialization of SearchEngineData or use simple conditional compilation using OSRM_EXTERNAL. For my point of view, template specialization is a better solution as we did for data facade. Probably, we can make some other QueryHeap that is better for offline usage with mobile or embedded devices with RAM limitations.
- move direction algorithms into hpp files or invent some other code hacks to inject with special data facade.

include/engine/datafacade/contiguous_internalmem_datafacade.hpp
TheMarex: You mean you will get a compile error because `routing_base_mld.hpp` does not include `contiguous_internalmem_datafacade.hpp` hence this will reference an unknown type?

Does including `contiguous_internalmem_datafacade.hpp` cause errors itself?

include/engine/phantom_node.hpp
TheMarex: Can you explain more why you could want to modify these?

include/engine/routing_algorithms/routing_base_mld.hpp
TheMarex: If this work better for you we can also move this to a compatibility header (`utils/thread_specific_ptr.hpp`) or something.

include/engine/routing_algorithms/shortest_path.hpp
TheMarex: I'm proposing we add this block to mainline OSRM so you can inject a header without having to modify OSRM code. I don't understand what you are trying to say.

TheMarex:
> use template specialization of SearchEngineData or use simple conditional compilation using OSRM_EXTERNAL. For my point of view, template specialization is a better solution as we did for data facade. Probably, we can make some other QueryHeap that is better for offline usage with mobile or embedded devices with RAM limitations.

I'm fine with specializing it but I think its overkill to do it just because of a compatibility issue that `boost::thread_specific_ptr` has with your platform.

> move direction algorithms into hpp files or invent some other code hacks to inject with special data facade.

Right now I'm very much favoring the approach that requires minimal changes to our mainline code base. We are in an ugly limbo right now because you are using internal interfaces that we don't actually want people to use. If the external code would be in mainline we would not need to do most of this changes because we could properly integrate it in the right files.

Since the long term goal will be to unify mainline and the external code I think these hacks will be closer to the final solution than big interface changes. The goal of this PR is to make your life easier while maintaining not to achieve architectural harmony.

vng:
Ok, will not argue with it. The only thing with shortest_path.cpp
Is it ok to make such kind of changes there?

```
#ifdef OSRM_EXTERNAL
// From the external private repo.
#include "../../offline/compressed_datafacade.hpp"
#endif

...

#ifdef OSRM_EXTERNAL
template InternalRouteResult
shortestPathSearch(SearchEngineData<mld::Algorithm> &engine_working_data,
                   const DataFacade<offline::mld::Algorithm> &facade,
                   const std::vector<PhantomNodes> &phantom_nodes_vector,
                   const boost::optional<bool> continue_straight_at_waypoint);
#endif
```
And approx the same stuff in shortest_path.hpp.
I don't see any other way to inject our code, not moving function's definitions into hpp file.

Also, please, check my other comments on your's in this PR.
@TheMarex ?

vng:
Year, this code can be placed in datafacade.hpp. The question was is it ok to have include reference from other repo?

PR #4321, opened 2017-07-22
Handles distinction of no-route vs invalid-route in mld alternatives

daniel-j-h:
The viaroute plugin always expects a route to be there potentially
with invalid edge weight to represent no-route-found. By switching
to the many-route-result for the mld alternatives algorithm we might
return an empty many-route-result invalidating the post-condition.

At some point we should think about making the interface less prone
to errors here. Reproduce e.g. with

    wget http://download.geofabrik.de/europe/netherlands-latest.osm.pbf
    osmium extract --bbox 6.24298095703125,53.016435582533816,6.88568115234375,53.363665164191865 netherlands-latest.osm.pbf -o groningen.osm.pbf

    ./osrm-extract -p ../profiles/car.lua groningen.osm.pbf
    ./osrm-partition groningen.osrm
    ./osrm-customize groningen.osrm
    ./osrm-routed --algorithm mld groningen.osrm

    curl -s 'localhost:5000/route/v1/driving/6.565710,53.219939;6.566032,53.219994?alternatives=true'

Before: assertion / crash, after: no-route status code

cc @TheMarex @oxidase

### DIFF
--- a/src/engine/plugins/viaroute.cpp
+++ b/src/engine/plugins/viaroute.cpp
@@ -120,9 +120,12 @@ Status ViaRoutePlugin::HandleRequest(const RoutingAlgorithmsInterface &algorithm
         routes = algorithms.ShortestPathSearch(start_end_nodes, route_parameters.continue_straight);
     }
 
+    // The post condition for all path searches is we have at least one route in our result.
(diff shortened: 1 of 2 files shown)


daniel-j-h:
Cherry-picked into 5.9.

PR #4327, opened 2017-07-24
Via way traffic signals

MoKob:
# Issue

Second step towards https://github.com/Project-OSRM/osrm-backend/issues/2681.

This PR removes the restriction that we cannot compress `traffic_signals` or anything similar that offers a node penalty.

The basic idea behind node-penalties is that we add a `turn`. So `abc` with a traffic signal at `b` becomes `ab` `bc` with a `turn` of type `NoTurn` in between and a turn penalty that matches the expected delay of the traffic signal.

This way of modelling node-penalties complicates multiple features in OSRM:
 - turn lanes
 - sliproad handling
 - via way restrictions
 - turn angle computations

and possibly even more.

This PR models node-penalties as segments within the graph. `abc` becomes `abbc` with the node-penalty applied to the segment `bb`. This way we don't create artificial turns that influence a lot of the codebase.

## Tasklist
 - [x] clean-up
    - [x] remove debug
    ~-[]strip lookahead code from turn lanes / sliproads / turn-angles...~ -> should move to a dedicated ticket and kept out of this PR
 - [x] review
 - [x] adjust for comments

## Requirements / Relations
 Requires #4255

### DIFF
--- a/src/extractor/graph_compressor.cpp
+++ b/src/extractor/graph_compressor.cpp
@@ -1,6 +1,8 @@
 #include "extractor/graph_compressor.hpp"
 
 #include "extractor/compressed_edge_container.hpp"
+#include "extractor/extraction_turn.hpp"
+#include "extractor/guidance/intersection.hpp"
 #include "extractor/restriction.hpp"
 #include "extractor/restriction_compressor.hpp"
 
@@ -19,7 +21,8 @@ namespace extractor
 {
 
 void GraphCompressor::Compress(const std::unordered_set<NodeID> &barrier_nodes,
-                               const std::unordered_set<NodeID> &traffic_lights,
+                               const std::unordered_set<NodeID> &traffic_signals,
+                               ScriptingEnvironment &scripting_environment,
                                std::vector<TurnRestriction> &turn_restrictions,
                                util::NodeBasedDynamicGraph &graph,
                                CompressedEdgeContainer &geometry_compressor)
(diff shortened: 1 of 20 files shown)


src/extractor/graph_compressor.cpp
```diff
@@ -217,6 +240,14 @@ void GraphCompressor::Compress(const std::unordered_set<NodeID> &barrier_nodes,
                 graph.GetEdgeData(forward_e1).duration += forward_duration2;
                 graph.GetEdgeData(reverse_e1).duration += reverse_duration2;
 
+                if (node_weight_penalty && node_duration_penalty)
+                {
+                    graph.GetEdgeData(forward_e1).weight += *node_weight_penalty;
+                    graph.GetEdgeData(reverse_e1).weight += *node_weight_penalty;
+                    graph.GetEdgeData(forward_e1).duration += *node_duration_penalty;
```
oxidase: From here additional duration penalties will be included into speed annotations.

src/extractor/graph_compressor.cpp
```diff
@@ -217,6 +240,14 @@ void GraphCompressor::Compress(const std::unordered_set<NodeID> &barrier_nodes,
                 graph.GetEdgeData(forward_e1).duration += forward_duration2;
                 graph.GetEdgeData(reverse_e1).duration += reverse_duration2;
 
+                if (node_weight_penalty && node_duration_penalty)
+                {
+                    graph.GetEdgeData(forward_e1).weight += *node_weight_penalty;
```
oxidase: Additional weight penalties will be lost after update of any segment of the edge-based graph node, so it is possible to get inconsistent weights values.

oxidase:
@MoKob i have added a test with traffic updates. Removal artificial traffic node works perfect with updates!

One point is a correctness of CSV files: a line `x,x,y` can potentially change values of `x,x` artificial segment. It is possible to check after CSV parsing at https://github.com/Project-OSRM/osrm-backend/blob/a186a06ba02afcdaa35a335bc18218464c51d2ed/src/updater/csv_source.cpp#L39-L40

PR #4329, opened 2017-07-24
Don't include turn durations when calculating the speed of a segment.

danpat:
# Issue

This is a proposed fix for https://github.com/Project-OSRM/osrm-backend/issues/4330 where we're not returning the same values with `annotations=speed` that we load via the traffic data mechanism, because we're including turn durations in segment durations for the last segment in any leg.

This change subtracts the turn duration from the duration annotation, and corrects both the `duration` and the `speed` annotation values to align with any traffic data loaded.  Turn duration data is now not included in any duration-based annotations (it is still included in the `weight` annotation, per [this line](https://github.com/Project-OSRM/osrm-backend/blob/6c0ab24100bb0917875eedba1f3c3a64ae158799/include/engine/routing_algorithms/routing_base.hpp#L206).

## Tasklist
 - [x] review
 - [x] adjust for comments

### DIFF
--- a/docs/http.md
+++ b/docs/http.md
@@ -539,10 +539,10 @@ Annotation of the whole route leg with fine-grained information about each segme
 **Properties**
 
 - `distance`: The distance, in metres, between each pair of coordinates
-- `duration`: The duration between each pair of coordinates, in seconds
+- `duration`: The duration between each pair of coordinates, in seconds.  Does not include the duration of any turns.
 - `datasources`: The index of the datasource for the speed between each pair of coordinates. `0` is the default profile, other values are supplied via `--segment-speed-file` to `osrm-contract`
 - `nodes`: The OSM node ID for each coordinate along the route, excluding the first/last user-supplied coordinates
-- `weight`: The weights between each pair of coordinates
+- `weight`: The weights between each pair of coordinates.  Does not include any turn costs.
 - `speed`: Convenience field, calculation of `distance / duration` rounded to one decimal place
 
 #### Example
--- a/include/engine/guidance/assemble_geometry.hpp
+++ b/include/engine/guidance/assemble_geometry.hpp
@@ -78,11 +78,19 @@ inline LegGeometry assembleGeometry(const datafacade::BaseDataFacade &facade,
         }
 
         prev_coordinate = coordinate;
-        geometry.annotations.emplace_back(
-            LegGeometry::Annotation{current_distance,
-                                    path_point.duration_until_turn / 10.,
-                                    path_point.weight_until_turn / facade.GetWeightMultiplier(),
-                                    path_point.datasource_id});
+        geometry.annotations.emplace_back(LegGeometry::Annotation{
+            current_distance,
+            // NOTE: we want annotations to include only the duration/weight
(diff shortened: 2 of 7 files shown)


TheMarex:
Conceptually it makes more sense to make the whole annotation inconsistent with the `RouteLeg` `duration`/`weight` and just not include the turn penalties at all, only the per segment weights.

Documenting this as segment values that do not include turn penalties seems fine to me. I don't think there is a sensible application that would relay on `RouteLeg.duration` == `sum(annotations.durations)`.

danpat:
@TheMarex OK, that sounds reasonable.  I've made that change.

Note that the leg duration calculation depends on the `duration_to_turn` including the turn duration [here](https://github.com/Project-OSRM/osrm-backend/blob/6c0ab24100bb0917875eedba1f3c3a64ae158799/include/engine/internal_route_result.hpp#L66), so it's not as simple as just removing [this line](https://github.com/Project-OSRM/osrm-backend/blob/6c0ab24100bb0917875eedba1f3c3a64ae158799/include/engine/routing_algorithms/routing_base.hpp#L205).

include/engine/guidance/assemble_geometry.hpp
```diff
@@ -78,11 +78,11 @@ inline LegGeometry assembleGeometry(const datafacade::BaseDataFacade &facade,
...
 
         prev_coordinate = coordinate;
-        geometry.annotations.emplace_back(
-            LegGeometry::Annotation{current_distance,
-                                    path_point.duration_until_turn / 10.,
-                                    path_point.weight_until_turn / facade.GetWeightMultiplier(),
-                                    path_point.datasource_id});
+        geometry.annotations.emplace_back(LegGeometry::Annotation{
+            current_distance,
+            (path_point.duration_until_turn - path_point.duration_of_turn) / 10.,
```
daniel-j-h: For a major change this is a bit hidden - can we make this more prominent and document why we need to do this here

include/engine/guidance/leg_geometry.hpp
```diff
@@ -38,8 +38,12 @@ struct LegGeometry
     struct Annotation
     {
         double distance; // distance in meters
-        double duration; // duration in seconds
-        double weight;   // weight value
+
+        // Total duration of a segment, in seconds, NOT including
+        // the turn penalty if the segment preceeds a turn
+        double duration;
+        double weight; // weight value, including the turn weight
```
daniel-j-h: I can see how it makes sense to not report the turn duration in the annotation - but why do we then report the turn weight in the annotations? Can and should we at least make the duration and weight annotation _both_ not report the turn duration and weight, respectively?

docs/http.md
```diff
@@ -539,10 +539,10 @@ Annotation of the whole route leg with fine-grained information about each segme
 **Properties**
 
 - `distance`: The distance, in metres, between each pair of coordinates
-- `duration`: The duration between each pair of coordinates, in seconds
+- `duration`: The duration between each pair of coordinates, in seconds.  Does not include the duration of any turns.
```
TheMarex: I would phrase this in a positive way: `Does only include the duration of each segment, not the duration of the turn penalty.`

Issue #4331, opened 2017-07-25
Turn angle calculation needs to respect initial road order


MoKob:
<img width="840" alt="screen shot 2017-07-25 at 09 03 42" src="https://user-images.githubusercontent.com/12932279/28559771-42a92a54-7118-11e7-8ca4-9ded103bc7ab.png">

When a fork turns, simply using the turn angle to compute the direction modifier can result in misleading directions. Forks should be hard limited to `left` and `right` to distinguish between the two possibilities, ignoring the actual turn angle.

[osrm link](http://map.project-osrm.org/?z=17&center=37.810161%2C-122.197827&loc=37.811361%2C-122.198964&loc=37.810615%2C-122.198267&hl=en&alt=0)

MoKob:
Turns out, that the fork handler already assigns (slight) left/right only. There seems to be a mixup in turn-angle computation here.

MoKob:
My findings show that the main problem at the given intersection is that multiple version of turn angle calculation interact badly with each other given the very small angle where the roads actually turn.

To fix this, we need to ensure that the initial order of turn angles is kept when changing computation from looking at the very first coordinate to some more advanced modelling.

miccolis:
Is there any lingering work here?  Appears that with #4332 merged this should be closed.

MoKob:
To capture the state here: #4332 was only a symptomatic fix. @oxidase has prepared a correct fix in #4443 which is currently up for review.

PR #4332, opened 2017-07-25
shift circle angle a bit

MoKob:
# Issue

This PR is tuning the constants for turns in sharp curves just a tiny bit to circumvent a mix-up in fork directions at a specific intersection (see https://github.com/Project-OSRM/osrm-backend/issues/4331).

While this does not solve the underlying issue, it eliminates a wrong instruction at a prominent place.

## Tasklist
        - I failed for more than an hour to get a test in cucumber that is exactly right. Given how specific and fragile the test would be, I currently cannot justify more time to create one.
 - [ ] review
 - [ ] adjust for comments

## Requirements / Relations


src/extractor/guidance/coordinate_extractor.cpp
```diff
@@ -122,15 +122,6 @@ util::Coordinate CoordinateExtractor::ExtractRepresentativeCoordinate(
     // coordinate set to add a small level of fault tolerance
     const constexpr double skipping_inaccuracies_distance = 2;
 
-    // fallback, mostly necessary for dead ends
```
MoKob: this is a duplication branch with above that cannot be reached.

daniel-j-h:
- I don't know how we come up with these constants
- They should probably be symbolized?

MoKob:
@daniel-j-h by educated guessing...

MoKob:
Capturing from chat:

This PR is all about choosing a small segment from the curve to represent the turn. The segment we had before moved one unit down the path.
A curved turn mostly feels straight, since the actual turn does not take a curve but the road after does.
Before we used the second 2-4 meters, now 4-6.
This is not a general fix but more a symptomatic change. The issue #4331 remains open.

PR #4333, opened 2017-07-25
handle throughabouts -- do not announce going through

### LINKED ISSUE #3142: Handle Throughabouts
We currently already have some tests (e.g. [here](https://github.com/Project-OSRM/osrm-backend/blob/46994e87be1a546446ef22d0b6aaba9a69894f5e/features/guidance/roundabout.feature#L284-L323)) for throughabouts, but we don't consider them.

A throughabout is a roundabout with an additional through lane. See the following links for reference.
http://wiki.openstreetmap.org/wiki/Throughabout
https://en.wikipedia.org/wiki/Roundabout#Hamburger_roundabout

Right now we give very strange instructions for these (e.g. exit roundabout at first exit) which is just misleading. 
We should (since we can only go through) potentially add a specific instruction (Throughabout) or simply use roundabout turn (possible conflicts in direction modifiers).

For wording, I imagine something like `At the roundabout, use the through lane`.

### LINKED ISSUE #4107: Roads crossing roundabouts

MoKob:
# Issue

This PR resolves #3142 and resolves #4107. For a quick fix, this PR is walking the road of the most silent instructions possible. So for going through a throughabout, we will not see any instruction in the usual case.
This might result in confusion for strange maneuvers, but I would consider most of these to ne separate issues.

Compare screenshots for examples:
screen shot 2017-07-25 at 16 55 41 2
[osm-link](https://www.openstreetmap.org/#map=19/38.91721/-76.97792)

screen shot 2017-07-25 at 17 08 27 2
going through the roundabout doesn't announce a maneuver. This might be overdoing the `silent` part. Should we go for a separate instruction here as in `Use the throughlane`

screen shot 2017-07-25 at 17 10 01 2
Exits like these make counting exits really confusing. They could use some improvement.

screen shot 2017-07-25 at 17 11 06 2

Usually I would expect a normal `exit` to ne on the outside of the roundabout. It would be nice to clarify this through lane here.

/cc @1ec5  

## Tasklist
 - [x] review
 - [x] adjust for comments
 - [x] ticket further improvements
 - [ ] add changelog entry

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@
       - Pass functions instead of strings to `WayHandlers.run()`, so it's possible to mix in your own functions.
       - Reorders arguments to `WayHandlers` functions to match `process_way()`.
       - Profiles must return a hash of profile functions. This makes it easier for profiles to include each other.
+      - Guidance: add support for throughabouts
 
 # 5.9.1
     - Infrastructure
--- a/features/guidance/roundabout-turn.feature
+++ b/features/guidance/roundabout-turn.feature
@@ -567,5 +567,5 @@ Feature: Basic Roundabout
             | ab    | residential   | in      |            |        |
 
         When I route I should get
-            | waypoints | turns                                                    | route                      |
-            | a,f       | depart,turn right,roundabout turn straight exit-1,arrive | in,through,through,through |
+            | waypoints | turns                    | route              |
+            | a,f       | depart,turn right,arrive | in,through,through |
--- a/features/guidance/roundabout.feature
+++ b/features/guidance/roundabout.feature
@@ -763,7 +763,7 @@ Feature: Basic Roundabout
 
         When I route I should get
            | waypoints | bearings | route       | turns                           |
-           | e,f       | 90 90    | edf,edf,edf | depart,roundabout-exit-1,arrive |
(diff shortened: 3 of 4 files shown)


MoKob:
Further considerations ticketed in https://github.com/Project-OSRM/osrm-backend/issues/4337

Issue #4340, opened 2017-07-26
Continue instruction on certain motorways


MoKob:
<img width="773" alt="image" src="https://user-images.githubusercontent.com/12932279/28615716-5eed43de-71fa-11e7-8dfa-31d945fba082.png">

On some motorway exits, ramps are not modelled as additional ramps but actually split off of the main lanes. On these, an instruction can be needed to indicate the driver needs to switch from the potentially rightmost lane to some of the left lanes to stay on the motorway.

We should investigate in how far lane information can be used to find out about these motorways.

PR #4341, opened 2017-07-26
Emit continue instruction on motorway when lanes split

MoKob:
# Issue

resolves https://github.com/Project-OSRM/osrm-backend/issues/4340

## Impact

A list of locations triggering this changed instruction:

[Germany](http://bl.ocks.org/d/fcff660c16780abccc91f1116befd36b)

[California](http://bl.ocks.org/d/129027f0048e79a426cd9160778fe4e6)

## Tasklist
 - [ ] evaluate above maps to check whether announced turns are or aren't desirable
 - [ ] review
 - [ ] adjust for comments

### DIFF
--- a/features/guidance/motorway.feature
+++ b/features/guidance/motorway.feature
@@ -297,5 +297,5 @@ Feature: Motorway Guidance
 
         When I route I should get
             | waypoints | route | turns                               |
-            | a,c       | ,     | depart,arrive                       |
+            | a,c       | ,,    | depart,continue straight,arrive     |
             | a,f       | ,,    | depart,off ramp slight right,arrive |
--- a/features/guidance/motorway.feature
+++ b/features/guidance/motorway.feature
@@ -281,3 +281,21 @@ Feature: Motorway Guidance
             | waypoints | route | turns         |
             | a,d       | ,     | depart,arrive |
             | b,d       | ,     | depart,arrive |
+
+     Scenario: Multi Lane highway splitting
+        Given the node map
+            """
+            a - - - - b - - - - - c
+                          e - - - f
+            """
+
+        And the ways
+            | nodes | name | ref    | highway       | oneway | lanes |
+            | ab    |      | US 101 | motorway      | yes    | 5     |
+            | bc    |      | US 101 | motorway      | yes    | 3     |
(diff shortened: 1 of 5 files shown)


MoKob:
@willwhite any update here? You said you would have a look into whether you'd be considering the `continue` instructions at the above maps a helpful addition

willwhite:
These results are super interesting. The difference between the US and Germany is striking. A common pattern in Germany appears to be for a few lanes to be added to a highway and then those lanes become an exit ramp and get peeled off the highway.

I think is why our hit rate in Germany is so high. These feel like false positives because, while you are loosing 2 out of 5 lanes, those 2 lanes were only just added for the explicit purpose of the exit. Without a way to compensate for that, we might just be detecting off-ramps with >2 lanes.

I don't think this is quite ready, but it will take more research to determine what would improve it. Let's pause on this for now?

willwhite:
Here's an example of where we have users requesting a "keep left" instruction:

- [Missing "Keep left to stay on I-395 S"](https://www.mapbox.com/get-directions/#18.32/38.8736/-77.04446?coordinates=-77.044047,38.874044;-77.044579,38.873429)

Does the heuristic capture this case in DC?

I wonder if we can expand this scope to include more than just motorways and capture examples like these as well:

- [Missing "Continue straight to stay on 10th St" at intersection with Bryant](https://www.mapbox.com/get-directions/#17.44/37.76992/-122.40952?coordinates=-122.410124,37.770634;-122.409264,37.769952)
- [Missing "Turn left to stay on Otis St" at intersection with Gough St](https://hey.mapbox.com/directions/#16.96/37.77174/-122.42042?coordinates=-122.419369,37.772543;-122.420226,37.770783)

MoKob:
A possible improvement here would be to detect how long lived a lane is. The idea is that if a lane only exists for a short time, it can split off without the driver requiring a maneuver (the driver would have to switch lanes intentionally to exit, instead of accidentally being routed off the road).

A first idea would be so see if there is a newly introduced lane (without any intersection) before the split:

undisturbed-new-lane

Before we look into how this information could be processed/stored, we should see if the detection is promising. This would require a map that works on the graph compressor to find roads that introduce lanes without disturbance and combines this information with the split-detection currently seen in this PR.

PR #4343, opened 2017-07-26
For external data facade.

vng:
This changes are needed for full support of custom (offline) data facade.
Replaces PR #4319
Made as minimal changes as possible.

@TheMarex @oxidase

### DIFF
--- a/src/engine/routing_algorithms/shortest_path.cpp
+++ b/src/engine/routing_algorithms/shortest_path.cpp
@@ -1,6 +1,11 @@
 #include "engine/routing_algorithms/shortest_path.hpp"
+
+#ifdef OSRM_EXTERNAL_MEMORY
+#include "routing/routing_base_offline.hpp"
+#else
 #include "engine/routing_algorithms/routing_base_ch.hpp"
 #include "engine/routing_algorithms/routing_base_mld.hpp"
+#endif
 
 #include <boost/assert.hpp>
 #include <boost/optional.hpp>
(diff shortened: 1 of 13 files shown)


src/engine/routing_algorithms/shortest_path.cpp
```diff
@@ -1,6 +1,11 @@
 #include "engine/routing_algorithms/shortest_path.hpp"
+
+#ifdef OSRM_EXTERNAL_MEMORY
+#include "routing/routing_base_offline.hpp"
```
oxidase: This is not really good to mix different code bases.
@TheMarex  I understand the point about placing code into cpp file, but we still need to move templated version into an impl header, remove references to offline code here and keep template instantiations in a cpp file to avoid rebuild.
So offline code could instantiate own implementations using private impl headers. Otherwise we again make wrong dependencies into OSRM code and block changes in separate projects.

oxidase:
after a chat with @TheMarex we could split into 3 parts:
- public header with declaration
- private header with implementation
- cpp with instantiations

vng:
Ok. We can wait while you will prepare main code base to inject out changes. Just let us know when it will be completed. Also, please, take into account that current approach with datafacade.hpp also assumes including "offline/compressed_datafacade.hpp" from the private repo. Otherwise, we together need to follow the approach described in #4319 with forward declaration of data facade classes.

@TheMarex @oxidase

oxidase:
@vng please check #4383 I am not allowed to commit into private repositories, so i cherry picked some commits, moved the shortest path implementation into `shortest_path_impl.hpp` and added a test that shows how an external facade can be used with the shotrest path plugin without modification of OSRM code base.

TheMarex:
Most of this should have gone in with #4383. Thanks for taking a stab at this!

PR #4353, opened 2017-07-28
Fixed getting entry bearing for maneuver.

vng:
To get entry bearing for maneuver, we need to reverse 'in' bearing for intersection, how it's made in other functions.

### DIFF
--- a/src/engine/guidance/collapse_scenario_detection.cpp
+++ b/src/engine/guidance/collapse_scenario_detection.cpp
@@ -17,7 +17,7 @@ namespace
 {
 
 // check bearings for u-turns.
-// since bearings are wrapped around at 0 (we only support 0,360), we need to do some minor math to
+// since bearings are wrapped around at 0 (we only support 0,360), we need to do some minor math to
 // check if bearings `a` and `b` go in opposite directions. In general we accept some minor
 // deviations for u-turns.
 bool bearingsAreReversed(const double bearing_in, const double bearing_out)
@@ -123,7 +123,7 @@ bool isStaggeredIntersection(const RouteStepIterator step_prior_to_intersection,
 
     const auto angle = [](const RouteStep &step) {
         const auto &intersection = step.intersections.front();
-        const auto entry_bearing = intersection.bearings[intersection.in];
+        const auto entry_bearing = util::bearing::reverse(intersection.bearings[intersection.in]);
         const auto exit_bearing = intersection.bearings[intersection.out];
         return util::bearing::angleBetween(entry_bearing, exit_bearing);
     };


src/engine/guidance/collapse_scenario_detection.cpp
```diff
@@ -123,7 +123,7 @@ bool isStaggeredIntersection(const RouteStepIterator step_prior_to_intersection,
 
     const auto angle = [](const RouteStep &step) {
         const auto &intersection = step.intersections.front();
-        const auto entry_bearing = intersection.bearings[intersection.in];
+        const auto entry_bearing = util::bearing::reverse(intersection.bearings[intersection.in]);
```
MoKob: Can you elaborate a bit on how you verified that this is necessary?

Off the top of my head and without having checked again:

The bearings for an intersection could look like `[0,90,180,270]` while `in = 0, out = 2`.

This would represent a straight traversal on the north-south axis (note that the bearings are stored from the center of the intersection, so in is already reversed).

The angle between in and out in this representation should be 180 degree, which would be the case without reversing the bearing.
(see `is_right: > 45 < 135`).
With reversing, both bearings would be identical and there would be a turn angle of zero.

I might be totally off here, though.

vng: 'in' bearing in the intersection struct is stored opposite to the driving direction. You can ensure about it in the assembleSteps function [here](https://github.com/Project-OSRM/osrm-backend/blob/master/include/engine/guidance/assemble_steps.hpp#L183)

'in' bearing from intersection also reversed [here](https://github.com/Project-OSRM/osrm-backend/blob/master/src/engine/guidance/collapse_turns.cpp#L38) and [here](https://github.com/Project-OSRM/osrm-backend/blob/master/src/engine/guidance/post_processing.cpp#L227) while passing to bearing::angleBetween.

But now I'm confused with the bearings::angleBetween function:
```
bearings::angleBetween(0, 0) == bearings::angleBetween(180, 180) == 180;
bearings::angleBetween(0, 180) == bearings::angleBetween(180, 0) == 0;
```
Seems like it's made specially for the opposite driving direction 'in' bearing. In that case, the logic in collapse_turns.cpp or post_processing.cpp should also be reviewed.

daniel-j-h: @MoKob @vng what are next actions here - can we ticket reviewing the post-processing bearing calculations to work on as a next step and move forward here?

MoKob: From the values @vng posted, the change should be fine. It is void, since `is_right` and `is_left` are used in combination (right + left), but it might prevent future bugs.

PR #4357, opened 2017-07-31
Fix a single weekday grammar callback

oxidase:
# Issue

The PR adds a constructor callback for a single weekday condition like `Su 00:00-24:00`

## Tasklist
 - [ ] review
 - [ ] adjust for comments

## Requirements / Relations

### DIFF
--- a/src/util/opening_hours.cpp
+++ b/src/util/opening_hours.cpp
@@ -218,9 +218,9 @@ struct opening_hours_grammar : qi::grammar<Iterator, Skipper, std::vector<Openin
         weekday_sequence = (weekday_range % ',')[ph::bind(&OpeningHours::weekdays, _r1) = _1];
 
         weekday_range
-            = wday[_a = _1, _b = _1]
-            >> -(('-' >> wday[_b = _1])
-                 | ('[' >> (nth_entry % ',') >> ']' >> -day_offset))
+            = (wday[_a = _1, _b = _1]
+               >> -(('-' >> wday[_b = _1])
+                    | ('[' >> (nth_entry % ',') >> ']' >> -day_offset)))
             [_val = ph::construct<OpeningHours::WeekdayRange>(_a, _b)]
             ;
 
--- a/include/util/opening_hours.hpp
+++ b/include/util/opening_hours.hpp
@@ -12,6 +12,7 @@ namespace util
 {
 
 // Helper classes for "opening hours" format http://wiki.openstreetmap.org/wiki/Key:opening_hours
+// Grammar https://wiki.openstreetmap.org/wiki/Key:opening_hours/specification
 // Supported simplified features in CheckOpeningHours:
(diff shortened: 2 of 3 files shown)


src/util/opening_hours.cpp
```diff
@@ -218,9 +218,9 @@ struct opening_hours_grammar : qi::grammar<Iterator, Skipper, std::vector<Openin
         weekday_sequence = (weekday_range % ',')[ph::bind(&OpeningHours::weekdays, _r1) = _1];
 
         weekday_range
-            = wday[_a = _1, _b = _1]
-            >> -(('-' >> wday[_b = _1])
-                 | ('[' >> (nth_entry % ',') >> ']' >> -day_offset))
+            = (wday[_a = _1, _b = _1]
```
MoKob: @oxidase I have a very hard time understanding this magic syntax. Given that these changes are not that obvious, do you think that we could add some commentary to make it easier to follow what is supposed to happen here?

oxidase: @MoKob i can explain what happened in this case: without parenthesis the semantic action `[_val = ph::construct<OpeningHours::WeekdayRange>(_a, _b)]` that constructs  a `WeekdayRange` was invoked only in a case if both the first and the second part of the rule succeed.

In case `Mo-Fr` local rule variable `_a` will be `Mo` and `_b` will be `Fr`. So the actions will be called as `ph::construct<OpeningHours::WeekdayRange>(1, 5)`.

If parsing of the second rule failed but the first part succeeded then  both `_a` and `_b` will be initialized to let say `Sa`, but construction of a `ph::construct<OpeningHours::WeekdayRange>(6, 6)` will not be called. So `WeekdayRange` will be a default initialized object with an empty range.

Adding parenthesis around both rules makes invocation of `ph::construct<OpeningHours::WeekdayRange>(_a, _b)` in both cases: 
- if the first  part succeeds, like 'Mo' -> `ph::construct<OpeningHours::WeekdayRange>(1,1)`
- if the first   and the second optional part  succeeds, like 'Tu-Fr' -> `ph::construct<OpeningHours::WeekdayRange>(2,5)`

Idk if such lengthy comments would make sense, but I can make a step-by-step intro how to add semantic to unimplemented  parts like `nth_entry` or `day_offset`, For example, with `nth_entry` it is possible to define `Oct Su[1]` (the first Sunday in October) like in the Oktoberfest example.

MoKob: @oxidase thank you for this explanation. This makes a lot more sense to me now.
I believe simply having a bit more context on what kind of format the grammar parses might already help here (`Parsing Mo-Fr into ph::construct<>(1,5)` or `Sa` into `ph:construct<6,6>`).  I agree that the longer explanation here might be a bit much, but it is already accessible now via git-blame, so I guess we are golden here.

Issue #4360, opened 2017-07-31
std::bad_alloc using conditional restrictions


oxidase:
For extract `osmium extract california-latest.osm.pbf -b -123.046875,36.5625,-120.9375,38.671875 -o sf.osm.pbf`
and using
```
./osrm-extract -p ../profiles/car.lua sf.osm.pbf --parse-conditional-restrictions && ./osrm-partition sf.osrm  && ./osrm-customize sf.osrm --time-zone-file timezones.geojson --parse-conditionals-from-now=1501920732
```

the current master exits with the error message
```
[error] [exception] std::bad_alloc
[error] Please provide more memory or consider using a larger swapfile
```

The first bad commit 2e9a7d9c1

PR #4362, opened 2017-08-01
fix serialisation of conditional turn restrictions

MoKob:
# Issue

resolves https://github.com/Project-OSRM/osrm-backend/issues/4360

## Tasklist
 - [ ] review
 - [ ] adjust for comments


include/extractor/serialization.hpp
```diff
@@ -184,6 +186,8 @@ inline void read(storage::io::FileReader &reader, TurnRestriction &restriction)
 inline void write(storage::io::FileWriter &writer, const TurnRestriction &restriction)
 {
     writer.WriteOne(restriction.is_only);
+    const int restriction_type = restriction.Type();
```
oxidase: it is better to use a bitsize-fixed types for serialization, like std::uint32_t

PR #4363, opened 2017-08-01
Add regression test for interaction between via-way and via-node

MoKob:
# Issue

Adding a regression test to ensure node and via relations do not cancel each other out.

## Tasklist
 - [ ] review
 - [ ] adjust for comments

### DIFF
--- a/features/car/restrictions.feature
+++ b/features/car/restrictions.feature
@@ -991,3 +991,40 @@ Feature: Car - Turn restrictions
         When I route I should get
             | from | to | route       |
             | a    | f  | ab,be,ef,ef |
+
+
+    @restriction @overlap @geometry
+    Scenario: Duplicated restriction
+        Given the node map
+            """
+            c
+            |
+            |   f
+            |   |
+            b-g-e
+            |   |
+            |   d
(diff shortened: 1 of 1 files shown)


features/car/restrictions.feature
```diff
@@ -991,3 +991,40 @@ Feature: Car - Turn restrictions
...
+            | restriction | bge      | ef     | e        | no_left_turn |
+
+        And the relations
+            | type        | way:from | way:via | way:to | restriction   |
+            | restriction | ab       | bge     | de     | no_right_turn |
+            | restriction | bc       | bge     | ef     | no_left_turn  |
+
+        When I route I should get
+            | from | to | route              |
+            | a    | d  | ab,bc,bc,bge,de,de |
```
oxidase: Would it make sense to add an additional check
```
| a    | f  | ab,bc,bc,bge,de,de,ef,ef |
```
to cover `no_left_turn` restrictions?

MoKob: since we do not distinguish between left/right but only check for `no`, this wouldn't make any difference.

PR #4367, opened 2017-08-01
Remove restricted turns from edge-based graph

### LINKED ISSUE #4366: Turn restrictions does not block turns
Turn restrictions In the updater implementation do not block turns, but set relatively high turn penalties. In some use cases, like blocking areas with turn restrictions,  such restrictions will have no effect on routing, because all restrictions will have equal additional penalty. 

To make such use cases possible at
https://github.com/Project-OSRM/osrm-backend/blob/84ce97a31e177dc7b26aa60c915745ce121f2253/src/updater/updater.cpp#L806 edges that have `turn_weight_penalty = INVALID_TURN_PENALTY` must be removed from the graph.

oxidase:
# Issue

Fixes #4366

## Tasklist
 - [ ] review
 - [ ] adjust for comments

## Requirements / Relations

### DIFF
--- a/features/options/profiles/version1.feature
+++ b/features/options/profiles/version1.feature
@@ -15,7 +15,7 @@ Feature: Profile API version 1
           properties.weight_name                     = 'test_version1'
           properties.weight_precision                = 2
 
-          assert(properties.max_turn_weight == 327.67)
+          assert(properties.max_turn_weight == 327.66)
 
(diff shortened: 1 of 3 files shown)


oxidase:
From communication with  @MoKob: we should not break connectivity of the road graph by restrictions and must be a possibility to route into a restricted area with high penalties.

PR #4372, opened 2017-08-04
Fix/warnings tests

MoKob:
# Issue

make master succeed in testing again

## Tasklist
 - [ ] review
 - [ ] adjust for comments

### DIFF
--- a/features/guidance/turn-angles.feature
+++ b/features/guidance/turn-angles.feature
@@ -1189,7 +1189,7 @@ Feature: Simple Turns
 
         When I route I should get
             | waypoints | route             | turns                           |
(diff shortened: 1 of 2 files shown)


features/guidance/turn-angles.feature
```diff
@@ -1189,7 +1189,7 @@ Feature: Simple Turns
 
         When I route I should get
             | waypoints | route             | turns                           |
-            | a,c       | rose,trift        | depart,arrive                   |
+            | a,c       | rose,trift,trift  | depart,turn slight left,arrive  |
```
MoKob: this changed due to traffic signals now being collapsed. Needed to fix #4332 test now

PR #4375, opened 2017-08-07
fix bug in conflict resolution

### LINKED ISSUE #4374: Conflict handling results in invalid turn directions
screen shot 2017-08-07 at 13 16 11 2
[osrm-link](http://map.project-osrm.org/?z=18&center=37.769479%2C-122.410596&loc=37.768544%2C-122.410762&loc=37.769667%2C-122.411336&hl=en&alt=0)

screen shot 2017-08-07 at 13 16 06 2
[osrm-link](http://map.project-osrm.org/?z=18&center=37.769479%2C-122.410596&loc=37.768544%2C-122.410762&loc=37.769739%2C-122.413133&hl=en&alt=0)

OSRM tries to assign distinct direction modifiers to turns. This was necessary to distinguish between two turns that are going into the same direction. However, it turns out that this is only introducing more problems than it solves. We need to revisit what kind of modifications we want to allow.

MoKob:
# Issue

resolves #4374

### DIFF
--- a/src/extractor/guidance/turn_handler.cpp
+++ b/src/extractor/guidance/turn_handler.cpp
@@ -391,6 +391,7 @@ Intersection TurnHandler::handleComplexTurn(const EdgeID via_edge, Intersection 
     {
         assignTrivialTurns(via_edge, intersection, 1, intersection.size());
     }
+
     return intersection;
 }
 
@@ -744,7 +745,7 @@ void TurnHandler::handleDistinctConflict(const EdgeID via_edge,
         right.instruction = {right_type, DirectionModifier::Right};
         return;
     }
-    // Two Right Turns
+    // Two Left Turns
     if (angularDeviation(left.angle, 270) < MAXIMAL_ALLOWED_NO_TURN_DEVIATION)
     {
         // Keep left perfect, shift right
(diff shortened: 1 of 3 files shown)


src/extractor/guidance/turn_handler.cpp
```diff
@@ -773,17 +774,18 @@ void TurnHandler::handleDistinctConflict(const EdgeID via_edge,
         return;
     }
 
-    if (getTurnDirection(left.angle) == DirectionModifier::Right)
+    // turn to the right
+    if (getTurnDirection(left.angle) <= 180)
```
MoKob: this is the main bug here. By checking for the modifier, we do not use right turns when the angles involved are slight/sharp

src/extractor/guidance/turn_handler.cpp
```diff
@@ -773,17 +774,18 @@ void TurnHandler::handleDistinctConflict(const EdgeID via_edge,
...
     {
         if (angularDeviation(left.angle, 85) >= angularDeviation(right.angle, 85))
         {
-            left.instruction = {left_type, DirectionModifier::Right};
-            right.instruction = {right_type, DirectionModifier::SharpRight};
+            left.instruction = {left_type, DirectionModifier::SlightRight};
```
MoKob: I notices this bug in the process of fixing the above.
`Left` is farther away from a 90 degree turn than `right`. So left cannot be a stronger turn than `right`.
These two were mixed up.

src/extractor/guidance/turn_handler.cpp
daniel-j-h: Urgh this compiled because of the enum not being a strong typedef, right?

Issue #4273, opened 2017-07-10
Pre-compute segregated intersections


MoKob:
![roads](https://user-images.githubusercontent.com/12932279/28023950-a6fb116e-658f-11e7-97f3-b02da0825ba9.png)

In post-processing, we can only access the information along the shortest path. This information is limited in what we can access, especially with regards to the roads connected to the path itself.
For specific situations, we need to know about all the ways associated with the intersection of two turns.

In the above scenario, I would consider the two situations different based on the assigned names.
In I), we see a segregated intersection and we would consider the maneuvers shown to the right.

If we only change a minor detail, having a road `a` and  road `b` running next to each other, I would consider all of the turns at the intersection as two distinct maneuvers.
The distinction in post-processing between those situations is not possible, though.

Therefore I would like to propose a change in the paradigm of post-processing: Introduce an additional pre-requisite which shows that a turn is part of a complex intersection.
We could compute in pre-processing and for every turn if it is part of a complex maneuver (essentially identifying segregated intersections in pre-processing).
This would allow us to see locations that would actually partake in a collapse scenario (which we could evaluate on a map, rather than on routes coming in one at a time) and construct scenarios that get access to a lot more information, since we can actually traverse the full graph at that time.

Technically, we could implement a dedicated post-processing for MLD, but as long as we want to support multiple algorithms, we need to assume the graph as untraversable.

This would require:

 - [ ] detect segregated intersections in turn analysis
 - [ ] add a flag to instructions indicating whether they are part of maneuvers
 - [ ] update post-processing to require said flag on instructions (where appropriate)

The immediate benefits I see in this approach:
 - more resilient post-processing (we know where it happens)
 - simpler scenarios in post-processing (we did the leg-work already)
 - more advanced detection (e.g. like this double turn) [osrm](http://map.project-osrm.org/?z=18&center=38.903403%2C-77.019412&loc=38.905140%2C-77.015748&loc=38.904347%2C-77.015941&hl=en&alt=0)

![screen shot 2017-07-10 at 16 59 12 2](https://user-images.githubusercontent.com/12932279/28024497-29ebe8ea-6591-11e7-8c37-d5286cfa5524.png)

TheMarex:
What would be the difference between dedicated flag and making the first left turn `Suppressed`?

vng:
[...]

PR #4380, opened 2017-08-07
Use turn tag to prevent suppression of designated turns, #3365

### LINKED ISSUE #3365: Obvious Turns Fail to Recognise Turning Roads in segregated Intersections
Our current obvious turn discovery fails in cases like this:
screen shot 2016-11-24 at 16 41 24

Here the road keeps the same reference, but follows a series of different roads.

I've added a test over at https://github.com/Project-OSRM/osrm-backend/pull/3364. 

Currently, the turn to the left at the intersection is considered obvious, the one to the right is announced as a turn.

In my opinion, the turn to the left and the turn to the right should both not be classified obvious.

/cc @karenzshea

oxidase:
# Issue

Fixes #3365 by using [`turn`](https://wiki.openstreetmap.org/wiki/Key:turn?uselang=en-US) tag.

At the moment the new flag `turn` is used only to prevent suppression of `Continue` instruction if ways have tags `turn=left` or `turn=right`

/cc @MoKob 

## Tasklist
 - [ ] check places with modified instructions
 - [ ] review
 - [ ] adjust for comments

## Requirements / Relations


profiles/lib/way_handlers.lua
```diff
@@ -82,6 +82,14 @@ function WayHandlers.startpoint(profile,way,result,data)
   end
 end
 
+-- determine if this way has the indicated direction in which a way will lead
+function WayHandlers.turn(profile,way,result,data)
+  local turn = way:get_value_by_key("turn")
+  if turn == "left" or turn == "right" then
+    result.is_turn = true
+  end
+end
```
daniel-j-h: Needs an update in the taginfo.json (and maybe also the changelog)

daniel-j-h:
Hm from the osm wiki the `turn=*` tag is not clear to me - `turn:lanes=*` are for turn lanes, the `turn=*` tag is for a single ways indicated direction? I can also only see a few thousand of these in the wild:
 
https://taginfo.openstreetmap.org/keys/turn#values

Could you clarify the use-case here in this scenario where turn lanes are tagged initially:

https://www.openstreetmap.org/edit#map=20/49.48756/8.47643

MoKob:
@oxidase I don't feel that using barely supported tags is the way to go here. It feels like the *easy way out*. Are there other adjustments we could consider to prevent the suppression of an instruction? I feel that #4273 would be a better way to detect this and properly collapse it.

TheMarex:
Hrm I've looked at the usages around Germany and this tags seems primarily used for tagging sliproads. Maybe this would be a nice addition for the sliproad handler to augment automatic detection?

oxidase:
@TheMarex i agree here with @MoKob as we need to fix #4273 first and recheck the particular intersection. I  think it can be closed for now and it is always possible to restore the branch if it will be useful in the future.

PR #4384, opened 2017-08-08
Conditional via ways

MoKob:
# Issue

Add support for conditional via-way restriction.

 - requires refactor of updater
 - breaks fileformat of osrm.restrictions

## Tasklist
 - [x] add more regression / cucumber cases (see docs/testing.md)
 - [x] support overlap of conditional-node/conditional via-way
 - [ ] review
 - [ ] adjust for comments


include/extractor/restriction_index.hpp
```diff
@@ -0,0 +1,101 @@
...
+
+    auto Restrictions(NodeID first, NodeID second) const
+    {
+        return restriction_hash.equal_range(std::make_pair(first, second));
+    };
+
+    auto Size() const { return restriction_hash.size(); }
+
+  private:
+    boost::unordered_multimap<std::pair<NodeID, NodeID>, restriction_type *> restriction_hash;
```
oxidase: Is there any preference for `boost::unordered_multimap` over `std::unordered_multimap`

daniel-j-h: probably https://bugs.llvm.org//show_bug.cgi?id=21275 again

PR #4385, opened 2017-08-08
Add workaround for Visual C++ issue with std::array in debug

mloskot:
If Visual C++ `_ITERATOR_DEBUG_LEVEL > 0`  (eg. while building Debug configuration), then accessing `std::array<char[N], M>` elements via reference to const causes compilation error:

```
...\msvc\14.10.25017\include\array(181): error C2440: 'return': cannot convert from 'const char *' to 'const char (&)[256]'
```

Alternative workaround is to remove `const` qualifier from the `GetClassName` method.

------

TBH, I haven't checked in details why the iterator debug facilities are being **annoying**.
I have reported it as an issue 
https://developercommunity.visualstudio.com/content/problem/90761/iterator-debug-level-causes-problems-with-stdarray.html

### DIFF
--- a/include/extractor/profile_properties.hpp
+++ b/include/extractor/profile_properties.hpp
@@ -83,7 +83,8 @@ struct ProfileProperties
     std::string GetClassName(std::size_t index) const
     {
         BOOST_ASSERT(index <= MAX_CLASS_INDEX);
-        return std::string(class_names[index]);
+        const auto &name_it = std::begin(class_names) + index;
+        return std::string(*name_it);
     }
 
     double GetWeightMultiplier() const { return std::pow(10., weight_precision); }


include/extractor/profile_properties.hpp
```diff
@@ -83,7 +83,16 @@ struct ProfileProperties
     std::string GetClassName(std::size_t index) const
     {
         BOOST_ASSERT(index <= MAX_CLASS_INDEX);
+#if defined(_MSC_VER) && defined(_ITERATOR_DEBUG_LEVEL) && (_ITERATOR_DEBUG_LEVEL > 0)
+        // Work around issue caused by Visual C++ iterator debug level facilities:
+        // ...\msvc\14.10.25017\include\array(181): error C2440: 'return': cannot convert from 'const char *' to 'const char (&)[256]'
+        // Alternative is to remove const qualifier from the method.
+        using NameArray = std::array<char[MAX_CLASS_NAME_LENGTH + 1], MAX_CLASS_INDEX + 1>;
+        auto class_names_ncv = const_cast<NameArray&>(class_names);
```
daniel-j-h: Hm should this be `auto&` or do you want a copy on purpose?

mloskot: I aimed for `auto&`, of course. Good catch, thanks. Fixed.

oxidase:
@mloskot can we wait with merge until the issue will be answered?

oxidase:
@mloskot just tried another workaround that does not change constness http://rextester.com/DVUWVV1940
```
        return std::string(*(std::begin(class_names) + index));
```

so type of `*x` is `char const [256]`.

mloskot:
Or array::data() with pointer arithmetic would do too. 

Yes I think it's sensible to wait for response from Microsoft, unless the new workaround are preferred anyway.

mloskot:
Microsoft has [confirmed the issue and fixed it in latest VS2017 Preview](https://developercommunity.visualstudio.com/content/problem/90761/iterator-debug-level-causes-problems-with-stdarray.html).

Meanwhile, I have changed the workaround to the version suggested by @oxidase

PR #4396, opened 2017-08-10
Switch to reliable non-existent handle for QueryHeap items

mloskot:
# Issue

Default-constructed objects of (`boost::heap`) `handle_type` are singular, including the wrapped `handle_type::iterator`.

Apparently, MSVC iterator debug facilities strictly require that one singular instance is compared to another singular instance.

It is not possible to get check-comparabe iterators of non-singular and singular instances as owning container will always mismatch.

Initially investigated at Boost ML (https://lists.boost.org/boost-users/2017/08/87787.php), eventually received confirmation via SO (https://stackoverflow.com/a/45622940/151641)

## Tasklist
 - [x] review
 - [x] adjust for comments

### DIFF
--- a/include/util/query_heap.hpp
+++ b/include/util/query_heap.hpp
@@ -180,7 +180,16 @@ class QueryHeap
     {
         BOOST_ASSERT(WasInserted(node));
         const Key index = node_index.peek_index(node);
-        return inserted_nodes[index].handle == HeapHandle{};
+
+        // Use end iterator as a reliable "non-existent" handle.
+        // Default-constructed handles are singular and
+        // can only be checked-compared to another singular instance.
+        // Behaviour investigated at https://lists.boost.org/boost-users/2017/08/87787.php,
+        // eventually confirmation at https://stackoverflow.com/a/45622940/151641.
+        // Corrected in https://github.com/Project-OSRM/osrm-backend/pull/4396
+        auto const end_it = const_cast<HeapContainer &>(heap).end();  // non-const iterator
+        auto const none_handle = heap.s_handle_from_iterator(end_it); // from non-const iterator
+        return inserted_nodes[index].handle == none_handle;
     }
 
     bool WasInserted(const NodeID node) const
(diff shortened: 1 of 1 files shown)


include/util/query_heap.hpp
```diff
@@ -180,7 +180,9 @@ class QueryHeap
     {
         BOOST_ASSERT(WasInserted(node));
         const Key index = node_index.peek_index(node);
-        return inserted_nodes[index].handle == HeapHandle{};
+        auto const end_it = const_cast<HeapContainer&>(heap).end(); // non-const iterator
```
oxidase: interestingly this method is used only in `DecreaseKey` in assertion. 
I guess  it can be removed here and in the tests,
and in `DecreaseKey` 
`BOOST_ASSERT(reference.handle != HeapContainer::s_handle_from_iterator(heap.end())); // was not removed`

mloskot:
I'll be more than happy to improve the PR but after I'm back from short trip in a week or two. No access to any compiler enabled computer ATM.

include/util/query_heap.hpp
mloskot: @daniel-j-h & @TheMarex shall I update the PR as you suggest or simply remove the method completely as @oxidase suggests?

mloskot:
while waiting for response to my question in https://github.com/Project-OSRM/osrm-backend/pull/4396#discussion_r135484926, I did apply:


* added relevant comment as per @daniel-j-h request.
* run `./scripts/format.sh` using clang-format 3.8 fixing a few inconsistencies, as per @TheMarex request

PR updated with both changes.

include/util/query_heap.hpp
TheMarex: @mloskot I would like to keep the interface change out of this PR. :-)

TheMarex:
@mloskot looks good to me, I restarted the travis build. I think the failure was unrelated. 👍

include/util/query_heap.hpp
mloskot: @TheMarex Makes sense.

mloskot:
@TheMarex Thanks

PR #4398, opened 2017-08-11
Silence Name Changes

MoKob:
# Issue

There have been a few discussions behind the scenes on this. `NewNames` currently mixes a concept I would refer to as `cue` with our maneuvers. This PR eliminates announcements of `New name` as maneuver.

For anyone trying to show names along the way, the route annotator can be used to retrieve the same information, without any heuristics on our side to reduce maneuver spam.

## Tasklist
 - [ ] review
 - [ ] adjust for comments

/cc @willwhite


miccolis:
@mcwhittemore I want to get your eyes on here too, but it sounds like everyone would rather delegate this work to the downstream consumers for now & revisit the larger question of how to represent non-manuevers in OSRM's API when we start looking at a new version.

mcwhittemore:
@miccolis - I tend to agree that "new name" as a maneuver muddles the maneuver concept. That said, I agree that this can be handled by down stream consumers for now.

We've talked a few times about what guidance looks like in OSRM moving forwards. I've thought of those changes as next version changes and I'd put this kind of functionality into that convo.

TheMarex:
I'm going to remove this from the `5.13.0` milestone because we are blocked here.

PR #4407, opened 2017-08-14
do not change fork directions when combining turns

MoKob:
# Issue

Forks can end up showing `fork-straight` when using a combined turn angle. Instead of changing the fork angle to be more or less to the `right|left`, we should keep unmodified directions for forks when combining turns.

## Tasklist
 - [x] add regression / cucumber cases (see docs/testing.md) (updated)
 - [x] review
 - [x] adjust for comments

### DIFF
--- a/features/guidance/perception.feature
+++ b/features/guidance/perception.feature
@@ -97,5 +97,7 @@ Feature: Simple Turns
             | ei    | left  | yes    |
 
         When I route I should get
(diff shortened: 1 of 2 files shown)


MoKob:
Added commentary (duplicated) to the locations

daniel-j-h:
Re-started Travis job, probably just a spurious failure. Let's see.

PR #4409, opened 2017-08-14
Route by supporting physical limitation of height and weight

frodrigo:
In short: test for maxwidth and maxheight.

Define a vehicle size using height and width.
Then validate the sizes to set mode.inaccessible if the vehicle can't fit on the way.

This PR only check for maxwidth and maxheight and not for maxweight and maxlength. Because the first two are physical limitations, the last are more legal issues.

This PR is extracted from a soft constrained truck profile I'm building. I treat maxweight and maxlength as soft legal limitation by adjusting weight.
If you don't want this in the core I can submit it to the contrib repo.

### DIFF
--- a/profiles/lib/way_handlers.lua
+++ b/profiles/lib/way_handlers.lua
@@ -8,6 +8,7 @@ local get_turn_lanes = require("lib/guidance").get_turn_lanes
 local set_classification = require("lib/guidance").set_classification
 local get_destination = require("lib/destination").get_destination
 local Tags = require('lib/tags')
+local Measure = require("lib/measure")
 
 WayHandlers = {}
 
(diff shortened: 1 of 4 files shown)


frodrigo:
cc @emiltin

frodrigo:
The test fails with strange issue. I misunderstand something or there is a bug.

TheMarex:
Looks interesting, thanks for taking a stab at this. We would definitely be interested in having this in the core. A minimum car width of 2.2 seem kind of large though?

Instead of removing the edges from the graph, you could also just mark them as `restricted`, which will avoid using them unless there is no other way or you are starting on a narrow street.

daniel-j-h:
Ping - what's the state of this @frodrigo?

profiles/lib/way_handlers.lua
```diff
@@ -429,6 +429,66 @@ function WayHandlers.parse_maxspeed(source,profile)
...
+      n = n * 12
+      local m = tonumber(inches:match("%d+"))
+      if m then
+        n = n + m
+      end
+      n = n * 2.54 / 100
+    end
+    return n
+  end
+end
```
deniskoronchik: We did something like this: https://github.com/vng/osrm-backend/blob/trucks/profiles/lib/measure.lua

frodrigo:
Still an important feature to me. Now I have other thing to do and I plan to work on OSRM in two weeks.

profiles/lib/way_handlers.lua
frodrigo: Nice. I planned to do a lib like this to read units.

frodrigo:
I redo it. Using @deniskoronchik commit for measure lib.

Now vehicle size is:
- vehicle_height = 2.5 (2.5m is the height of van)
- vehicle_width = 1.9 (American full-size car should be 2.0m)

Narrow ways are now 2.2m width, but open for change this.

TheMarex:
@frodrigo would be cool if you could rebase this. This looks like it would be ready to merge and an interesting addition.

Using the new default values, are there any roads that will be excluded, that we include now?

TimMcCauley:
@TheMarex @frodrigo great feature, thank you. I have one question: how does this feature behave in terms of passing dynamic values of the vehicle to the request directly? Could I for instance compute a route with width=2.2 and in the next request use width=2.5 ? Or would this mean that the graphs have to be preprocessed again beforehand?

frodrigo:
@TimMcCauley You have to recompute the graph.

frodrigo:
@TheMarex my last push is not good. I repush an old commit. I do somthing bad in my rebase. I think I lost the last commit with the fixed vehicle size and using the lib for parsing the unit.

TimMcCauley:
@frodrigo hm, this would take a quite a while I guess, depending on the size of the region, right?
 Would this also be possible via MLD which OSRM has implemented in the customizing step on the fly?

frodrigo:
@TheMarex Ok now it's done. My teammate have the right commits :-]

TimMcCauley:
@TheMarex could you possibly comment on my previous question - Is currently anything around dynamic weighting with OSRM possible or in the pipeline?

TheMarex:
@TimMcCauley we have a feature that is called `exclude`. You could add your own class [here](https://github.com/Project-OSRM/osrm-backend/blob/master/profiles/car.lua#L102-L104) for narrow streets and add it to the profile [here](https://github.com/Project-OSRM/osrm-backend/blob/master/profiles/car.lua#L107-L111). You can then specify this class in the query to exclude streets with this class e.g. pass `exclude=narrow`. Of course this only works with discrete steps.

I'm not sure what level of control you need but you could also do multiple categories like `narrow_below_2m`, `narrow_below_3m` etc.

HendrikLeuschner:
@TheMarex could you elaborate how this can be used in conjunction with conditional restrictions from #3841 ? Thanks a lot.

TimMcCauley:
@TheMarex - thank you for explaining!!

PR #4411, opened 2017-08-15
make end of road non obvious

MoKob:
# Issue

When reaching a oneway street at the end of the road, we currently do not emit a turn instruction. Even though there is no choice, it can make the user feel uncertain whether he is doing the right thing. E.g. if a road is modelled as dual carriage-way and has no obvious separation, turning right instead of left might be the clearer choice, an announcement of this turn will avoid confusion.

This PR introduces exactly this behaviour, announcing turns onto oneway through streets at the end of the road.

It affects quite a few intersections, as can be seen on [this map](https://www.mapbox.com/studio/styles/mokob/cj1yu7w72001f2srwyfyviz9p/edit/) -- contains values for Germany and California.

I've looked at many of the turns and from what I can tell, the selection shows a promising set.
All of these locations show a turn where we previously wouldn't have announced anything and now would announce `turn/continue right`.

## Tasklist
 - [x] review
 - [x] adjust for comments

### DIFF
--- a/include/extractor/guidance/intersection_handler.hpp
+++ b/include/extractor/guidance/intersection_handler.hpp
@@ -327,6 +327,42 @@ std::size_t IntersectionHandler::findObviousTurn(const EdgeID via_edge,
         return false;
     }();
 
+    // check whether we turn onto a oneway through street. These typically happen at the end of
+    // roads and might not seem obvious, since it isn't always as visible that you cannot turn
+    // left/right. To be on the safe side, we announce these as non-obvious
+    const auto turns_onto_through_street = [&](const auto &road) {
+        // find edge opposite to the one we are checking (in-road)
+        const auto in_through_candidate =
+            intersection.FindClosestBearing(util::bearing::reverse(road.bearing));
+
+        const auto &in_data = node_based_graph.GetEdgeData(in_through_candidate->eid);
+        const auto &out_data = node_based_graph.GetEdgeData(road.eid);
+
+        // by asking for the same class, we ensure that we do not overrule obvious by road-class
+        // decisions
+        const auto same_class = in_data.road_classification == out_data.road_classification;
+
+        // only if the entry is allowed for one of the two, but not the other, we need to check.
+        // Otherwise other handlers do it better
+        const bool is_oneway = !in_through_candidate->entry_allowed && road.entry_allowed;
+
+        const bool not_roundabout =
+            !(in_data.roundabout || in_data.circular || out_data.roundabout || out_data.circular);
+
+        // for the purpose of this check, we do not care about low-priority roads (parking lots,
+        // mostly). Since we postulate both classes to be the same, checking one of the two is
+        // enough
+        const bool not_low_priority = !in_data.road_classification.IsLowPriorityRoadClass();
+
+        const auto in_deviation = angularDeviation(in_through_candidate->angle, STRAIGHT_ANGLE);
+        const auto out_deviaiton = angularDeviation(road.angle, STRAIGHT_ANGLE);
+        // in case the deviation isn't considerably lower for the road we are turning onto,
+        // consider it non-obvious. The threshold here requires a slight (60) vs sharp (120)
+        // degree variation, at lest (120/60 == 2)
+        return is_oneway && same_class && not_roundabout && not_low_priority &&
+               (in_deviation / (std::max(out_deviaiton, 0.5)) <= 2);
+    };
+
     if (best_over_best_continue)
     {
         // Find left/right deviation
@@ -366,8 +402,7 @@ std::size_t IntersectionHandler::findObviousTurn(const EdgeID via_edge,
             angularDeviation(intersection[right_index].angle, STRAIGHT_ANGLE);
 
         // return best_option candidate if it is nearly straight and distinct from the nearest other
-        // out
-        // way
+        // out way
         if (best_option_deviation < MAXIMAL_ALLOWED_NO_TURN_DEVIATION &&
             std::min(left_deviation, right_deviation) > FUZZY_ANGLE_DIFFERENCE)
             return best_option;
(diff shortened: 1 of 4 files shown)


include/extractor/guidance/intersection_handler.hpp
```diff
@@ -327,6 +327,42 @@ std::size_t IntersectionHandler::findObviousTurn(const EdgeID via_edge,
         return false;
     }();
 
+    // check whether we turn onto a oneway through street. These typically happen at the end of
```
daniel-j-h: This is getting a bit out of hand with all the inline lambdas etc.

Do you think we could refactor the intersection handler a bit to make it easier to follow the logic here?

include/extractor/guidance/intersection_handler.hpp
```diff
@@ -327,6 +327,42 @@ std::size_t IntersectionHandler::findObviousTurn(const EdgeID via_edge,
...
+    // roads and might not seem obvious, since it isn't always as visible that you cannot turn
+    // left/right. To be on the safe side, we announce these as non-obvious
+    const auto turns_onto_through_street = [&](const typename IntersectionType::value_type &road) {
```
daniel-j-h: arg can be `const auto&` (?)

include/extractor/guidance/intersection_handler.hpp
```diff
@@ -327,6 +327,42 @@ std::size_t IntersectionHandler::findObviousTurn(const EdgeID via_edge,
...
+        const auto in_through_candidate =
+            intersection.FindClosestBearing(util::bearing::reverse(road.bearing));
```
daniel-j-h: Why reverse?

include/extractor/guidance/intersection_handler.hpp
```diff
@@ -327,6 +327,42 @@ std::size_t IntersectionHandler::findObviousTurn(const EdgeID via_edge,
...
+
+        const auto &in_data = node_based_graph.GetEdgeData(in_through_candidate->eid);
+        const auto &out_data = node_based_graph.GetEdgeData(road.eid);
+
+        // by asking for the same class, we ensure that we do not overrule obvious by road-class
+        // decisions
+        const auto same_class = in_data.road_classification == out_data.road_classification;
+
+        const auto in_deviation = angularDeviation(in_through_candidate->angle, STRAIGHT_ANGLE);
+        const auto out_deviaiton = angularDeviation(road.angle, STRAIGHT_ANGLE);
```
daniel-j-h: We seem to only use these further down, care to move them down to where we need them?

include/extractor/guidance/intersection_handler.hpp
```diff
@@ -327,6 +327,42 @@ std::size_t IntersectionHandler::findObviousTurn(const EdgeID via_edge,
...
+
+        // only if the entry is allowed for one of the two, but not the other, we need to check.
+        // Otherwise other handlers do it better
+        const bool is_oneway = !in_through_candidate->entry_allowed && road.entry_allowed;
+
+        const bool not_roundabout =
+            !(in_data.roundabout || in_data.circular || out_data.roundabout || out_data.circular);
```
daniel-j-h: Care to remind me about the difference between checking for the roundabout instruction and the roundabout data here?

include/extractor/guidance/intersection_handler.hpp
```diff
@@ -327,6 +327,42 @@ std::size_t IntersectionHandler::findObviousTurn(const EdgeID via_edge,
...
+    const auto turns_onto_through_street = [&](const typename IntersectionType::value_type &road) {
+        const auto in_through_candidate =
+            intersection.FindClosestBearing(util::bearing::reverse(road.bearing));
+
+        const auto &in_data = node_based_graph.GetEdgeData(in_through_candidate->eid);
+        const auto &out_data = node_based_graph.GetEdgeData(road.eid);
+
+        // by asking for the same class, we ensure that we do not overrule obvious by road-class
+        // decisions
+        const auto same_class = in_data.road_classification == out_data.road_classification;
```
daniel-j-h: The implication here is a change in road class is always obvious already? Is this always the case? Where is it determined?

include/extractor/guidance/intersection_handler.hpp
```diff
@@ -327,6 +327,42 @@ std::size_t IntersectionHandler::findObviousTurn(const EdgeID via_edge,
...
+        // for the purpose of this check, we do not care about low-priority roads (parking lots,
+        // mostly). Since we postulate both classes to be the same, checking one of the two is
+        // enough
+        const bool not_low_priority = !in_data.road_classification.IsLowPriorityRoadClass();
+
+        // in case the deviation isn't considerably lower for the road we are turning onto,
+        // consider it non-obvious. The threshold here requires a slight (60) vs sharp (120)
+        // degree variation, at lest (120/60 == 2)
+        return is_oneway && same_class && not_roundabout && not_low_priority &&
+               (in_deviation / out_deviaiton <= 2);
```
daniel-j-h: Can out deviation ever be zero?

include/extractor/guidance/intersection_handler.hpp
MoKob: we could, but that is not the scope of this PR

include/extractor/guidance/intersection_handler.hpp
MoKob: not necessarily, but we check for a `through` street. A street that changes class in between is probably not just a normal road but a different scenario

include/extractor/guidance/intersection_handler.hpp
MoKob: roundabout instructions can only be checked after they have been assigned. We are in the process of assigning turn-types, so they cannot be checked.

include/extractor/guidance/intersection_handler.hpp
MoKob: good catch!

MoKob:
Capturing from voice: @daniel-j-h says 👍  on merge.

PR #4415, opened 2017-08-16
Location-dependent left hand driving flag

oxidase:
# Issue

PR adds location-dependent tags and adds parsing `driving_side` tag.
Location-dependent data is taken from a GeoJSON file (file name is an argument of a new extractor option  `--location-dependent-data`) and provided as an optional argument `location_data` of `function process_way(profile, way, result, location_data)`. At the moment to use location-dependent data OSM files must be preprocessed with `osmium add-locations-to-ways`.

Left-hand driving flag is added to node-based-graph edges, edge-based-graph nodes and used in guidance pre-processing and in geospatial queries.

Some points to be discussed:
1)  [`MultiPolygon`](https://tools.ietf.org/html/rfc7946#section-3.1.7) support?
2) Use a nodes locations cache. This will require an implementation equivalent to `add-locations-to-ways` before parsing ways #4167
3) Merging conflicting tags at different locations of a way or multiple-defined at one node. At the moment a way is represented by a single node node and no dedicated merging is implemented. 
4) `is_left_hand_driving` flag is indexed by EBG node ids, but it can be indexed by geometry ids. This will require an additional file that is indexed by geometry ids, but contains geometry non-segments data.
5) `vector<bool>` is stored as 8-bit per 1-bit value, is it a new issue?

## Tasklist
 - [x] check with real OSM data
 - [x] check edge cases near signs similar https://wiki.openstreetmap.org/wiki/File:Fari-Wechsel.jpg
 - [x] review
 - [x] adjust for comments

## Requirements / Relations
Partially supersedes #4167


emiltin:
do i understand it correctly, that this adds a general ability to use location-based data when preprocessing osm data, not just data about left/right side driving?

can the features in the geojson file overlap?

oxidase:
@emiltin yes, it is not limited to left-hand driving tags and GeoJSON can have any data that will be forwarded into `parse_way` function, but I did not checked yet other tags. The next step will be  to check `driving_side` together with some other tags.

Polygons with non-intersecting tags sets can overlap. If some overlapping polygons have the same tag then  `last-value-wins` merging in the rtree indexing order  will be used. Basically it just a https://github.com/Project-OSRM/osrm-backend/blob/de0a73d3c8256bbab6dfd356837b535f011f1783/src/extractor/location_dependent_data.cpp#L133 so no dedicated conflict resolution yet.

emiltin:
ok ,thanks for clarifying. this seems very useful. it will cover a different set of uses related to location data than the raster-based approach already available.

TheMarex:
> vector<bool> is stored as 8-bit per 1-bit value, is it a new issue?

You mean stored like that on disk? As far as I remember there were some issues around using out default memory-dump based serialization. If you see a fix for that, go for it. 👍 

> Merging conflicting tags at different locations of a way or multiple-defined at one node.

This is mainly an issue while crossing country polygon borders?

oxidase:
@TheMarex i added a vector<bool> serializer

> This is mainly an issue while crossing country polygon borders?

yes, maybe for driving side flag this would make no sense (an edge case https://www.openstreetmap.org/export#map=17/22.52030/114.07047), but in many cases ways can cross polygons with the same tags, but different values.

> Do you think it would be possible to specify this from within the profile? Do we want to allow for multiple of those sources? This could work similar to how we use raster-data, in this case it would just be polygon based data.

yes, it is possible to specify in profiles, but data must loaded before creating contexts, otherwise many copies of rtrees will be multiplied for every thread-local context.

Multiple of sources are also possible, in principle we can index data in a single rtree, but it will bring  the merging question to discussion,  i would go first towards a single GeoJSON file with MultiPolygon support and `Sol2ScriptingEnvironment` as the sole owner of a constant  single copy of data.

TheMarex:
> yes, it is possible to specify in profiles, but data must loaded before creating contexts,

We could call the initialization in the `setup` functions.  In `setup` we would call a function in C++ land that would check if the specific GeoJSON file has already been loaded. Data would be owned by `SolScriptingEnvironment` (not `SolScriptingContext`) so one global R-Tree per file and many threads access. I take it lookups in the RTree would be thread-safe?

src/extractor/scripting_environment_lua.cpp
```diff
@@ -224,7 +225,7 @@ void Sol2ScriptingEnvironment::InitContext(LuaScriptingContext &context)
                                                 "interpolate",
                                                 &RasterContainer::GetRasterInterpolateFromSource);
 
-    context.state.new_usertype<ProfileProperties>(
+    auto registration_ProfileProperties = context.state.new_usertype<ProfileProperties>(
```
TheMarex: Why do we need to save the return type of `new_usertype` here?

unit_tests/util/serialization.cpp
```diff
@@ -0,0 +1,38 @@
+#include "storage/serialization.hpp"
```
TheMarex: Good idea. We should add tests for the other serialization methods too at some point.

features/car/side_bias.feature
```diff
@@ -76,3 +74,45 @@ Feature: Testbot - side bias
...
+        Given the profile "car"
+        And the node map
+            """
+            a   b   c
+
+                d
+            """
+        And the ways with locations
+            | nodes | driving_side |
+            | ab    | right        |
```
oxidase: there some [residential roads](https://www.openstreetmap.org/way/134124277), [service roads](https://www.openstreetmap.org/way/34420205), [private ways]( https://www.openstreetmap.org/way/384394384)  or [u-turns](https://www.openstreetmap.org/way/364692295). Yes, we can add in future such handling, atm tags on ways have higher priority than default values.

include/extractor/extraction_way.hpp
```diff
@@ -114,6 +115,7 @@ struct ExtractionWay
     bool is_startpoint : 1;
     bool forward_restricted : 1;
     bool backward_restricted : 1;
+    bool is_left_hand_driving : 1;
```
oxidase: VCS have other rules for padding, but here it should be similar to gcc packing

include/extractor/profile_properties.hpp
```diff
@@ -103,7 +103,7 @@ struct ProfileProperties
     bool continue_straight_at_waypoint;
     //! flag used for restriction parser (e.g. used for the walk profile)
     bool use_turn_restrictions;
-    bool left_hand_driving;
+    bool left_hand_driving; // DEPRECATED: property value is local to edges from API version 2
```
oxidase: removed

src/extractor/location_dependent_data.cpp
```diff
@@ -0,0 +1,147 @@
...
+    table_setter(sol::table &table, const std::string &key) : table(table), key(key) {}
+    template <typename T> void operator()(const T &value) const { table.set(key, value); }
+    void operator()(const boost::blank &) const { /* ignore */}
+
+    sol::table &table;
+    const std::string &key;
+};
+}
+
+sol::table LocationDependentData::operator()(sol::state &state, const osmium::Way &way) const
```
oxidase: i would like to keep it as a shortcut to avoid creating `unordered_map`. It can be easily adjusted when it will be clear how conflicting tags must be merged

include/extractor/extraction_way.hpp
oxidase: after some hours of debugging i think i will keep bool type here, because of some strange gcc behavior here: with `std::uint8_t` setting `is_startpoint` also sets `forward_restricted` and `backward_restricted`. it can be either ub or gcc optimization issue

oxidase:
@TheMarex please can you check the last changes? 

I have tried to use real data with OSM administrative boundaries and results are very disappointing: 
265 `boost::geometry::within(point, polygons[v.second].first)` calls increase OSM parsing from 30ms to 835 with just Thailand single border line.

oxidase:
@TheMarex i have ported point-in-polygon check from osmium and added some unit tests to check correctness. The performance results are better than using `boost:geometry::within` but the price is additional memory usage to store segment bands. For `berlin-latest` with 690932 ways anf the Germany border parsing time changed from 5.2 to 40 seconds, that is ~ 100-200 microseconds per lookup that is still bad.

There are still some space for improvement because "bands" solution is a naive implementation of an interval tree with fixed-size bands for OSM Germany border with latitudes from 47.2701 to 55.0992 with a band width 0.00056389 degrees the histogram of segments is
germany-bands13884

so a check performance will depend on node locations and at the southern border it will be 1000 times slower than at the northern one. I think it is ok to merge under assumption of simple polygons for the left-hand side driving checks, and performance can be improved later by using similar check with an interval tree.

Another huge performance improvement  is to avoid using `unordered_map` merging because 54% out 59% of `LocationDependentData::operator()` time for `berlin-latest` extraction is spent at https://github.com/Project-OSRM/osrm-backend/blob/116d7d2b2932f7ed4c237b2039afd1d521aa7b77/src/extractor/location_dependent_data.cpp#L204
But i don't know atm the best way how to deal with the issue: may be create Lua tables in every context in advance, don't merge tables in c++ and provide an optional table of tags tables (it is possible  to add also relation tags or any other "extension" tags into that table). So it is mainly a Lua interface question.

src/extractor/location_dependent_data.cpp
```diff
@@ -0,0 +1,284 @@
...
+            const auto &polygons = geometry["coordinates"].GetArray();
+            for (rapidjson::SizeType ipolygon = 0; ipolygon < polygons.Size(); ++ipolygon)
+            {
+                index_polygon(polygons[ipolygon].GetArray(), properties_index);
+            }
+        }
+    }
+
+    // Create R-tree for bounding boxes of collected polygons
+    rtree = rtree_t(bounding_boxes);
```
TheMarex: Hrm I don't quite understand how this works with multiple input files: It seems to me on every call of this functions the rtree gets overwritten and all previous data is lost?

src/extractor/scripting_environment_lua.cpp
```diff
@@ -882,18 +889,50 @@ void LuaScriptingContext::ProcessNode(const osmium::Node &node, ExtractionNode &
...
     {
     case 2:
-        way_function(profile_table, way, result);
+        if (location_dependent_data.empty())
+        {
+            way_function(profile_table, way, result);
+        }
+        else
+        {
+            way_function(profile_table, way, result, toLua(state, location_dependent_data(way)));
```
TheMarex: Does this add any overhead when not giving any location dependent data as input (e.g. no geojson file)?

src/extractor/location_dependent_data.cpp
```diff
@@ -0,0 +1,284 @@
...
+{
+    properties_t result;
+
+    auto merger = [this, &result](const rtree_t::value_type &rtree_entry) {
+        const auto &polygon_properties = properties[polygons[rtree_entry.second].second];
+        result.insert(polygon_properties.begin(), polygon_properties.end());
+    };
+
+    // Search the R-tree and collect a Lua table of tags that correspond to the location
+    rtree.query(boost::geometry::index::intersects(point) &&
```
TheMarex: Instead of saving each polygon in one bounding box you could try a quad tree approximation:
- Divide the bounding box into 4 equal parts
- For every new box check if its fully container, fully outside or intersects the boundary
- If a box intersects the boundary: Continue dividing that box.
- Stop when all boxes are either fully in or fully out, or you reached a minimum box size (like 10x10m)

Now your query can terminate immediately when it hits a fully in/fully out bounding box. You only need to do the expensive point-in-polygon check on the points that hit the boundary boxes.

src/extractor/location_dependent_data.cpp
oxidase: :man_facepalming: will be fixed

src/extractor/scripting_environment_lua.cpp
oxidase: empty data should handled in `then` branch when rtree is empty, and no optional arguments will be passed to the lua function

oxidase:
@TheMarex pls, could you review the current state of the PR. I would keep the current implementation of the point-in-polygon check unless it will present a real bottleneck. Local processing time of germany-latest with Germany border is 474.374 seconds, without location dependent data but with locations cache is 455.142 seconds, and on master 402.891 seconds. So 12% increase is due to locations caching and 5% increase due to lookups with 100% hits. Also the location cache (I used FlexMap from osmium) would require additional memory to map node ids to locations.

@MoKob i found two places with comments [ `needs to be adjusted for left side driving`](https://github.com/Project-OSRM/osrm-backend/blob/1350460835d5cac94506a98909e51106772bac3b/src/extractor/guidance/turn_lane_matcher.cpp#L74) and [`Needs to be fixed for left side driving`](https://github.com/Project-OSRM/osrm-backend/blob/1350460835d5cac94506a98909e51106772bac3b/src/extractor/guidance/turn_lane_matcher.cpp#L98)in guidance. Please could you clarify what exactly must be adjusted?

TheMarex:
>  5% increase due to lookups 

That seems fine with me. What changed since the original measurements on Berlin (` parsing time changed from 5.2 to 40 seconds`)?

oxidase:
@TheMarex just checked the mean difference is 0.3 seconds or 5%
```
> t.test(d[d$V2=='no',1],d[d$V2=='yes',1])

	Welch Two Sample t-test

data:  d[d$V2 == "no", 1] and d[d$V2 == "yes", 1]
t = -4.4597, df = 197.97, p-value = 1.374e-05
alternative hypothesis: true difference in means is not equal to 0
95 percent confidence interval:
 -0.4449259 -0.1720891
sample estimates:
mean of x mean of y 
 6.723882  7.032389 
```
berlin

Issue #4420, opened 2017-08-17
Suffix tables do not detect all suffixes/prefixes


MoKob:
The following test case 

https://github.com/Project-OSRM/osrm-backend/blob/master/features/guidance/new-name.feature#L280-L293

and a few more exhibit prefix/suffix changes our name-change heuristic currently does not identify as such.

This can lead to unnecessary announcements, problems in discovering obvious turns and many more implications.

PR #4421, opened 2017-08-17
fix detection of suffix/prefix changes for name-changes

MoKob:
# Issue

Resolves https://github.com/Project-OSRM/osrm-backend/issues/4420

## Tasklist
 - [x] review
 - [ ] adjust for comments

### DIFF
--- a/include/util/guidance/name_announcements.hpp
+++ b/include/util/guidance/name_announcements.hpp
@@ -10,6 +10,7 @@
 
 #include <algorithm>
 #include <string>
+#include <tuple>
 #include <utility>
 #include <vector>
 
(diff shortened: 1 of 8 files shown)


include/util/guidance/name_announcements.hpp
```diff
@@ -64,51 +129,24 @@ inline bool requiresNameAnnounced(const std::string &from_name,
...
+    const auto checkForPrefixOrSuffixChange =
+        [](const std::string &first, const std::string &second, const SuffixTable &suffix_table) {
+            std::string first_prefix, first_suffix, second_prefix, second_suffix;
+            std::tie(first_prefix, first_suffix, second_prefix, second_suffix) =
+                decompose(first, second);
 
-        const auto first_prefix_and_suffixes = getPrefixAndSuffix(first);
-        const auto second_prefix_and_suffixes = getPrefixAndSuffix(second);
+            const auto checkTable = [&](const std::string &str) {
+                // workaround for cucumber tests:
```
oxidase: :scream_cat:

include/util/guidance/name_announcements.hpp
```diff
@@ -25,17 +26,81 @@ namespace guidance
...
+                if (dp[i][j] > best)
+                {
+                    best = dp[i][j];
+                    best_pos = i + 1;
+                }
+            }
+        }
+    }
+    // the best position marks the end of the string
+    return lhs.substr(best_pos - best, best);
```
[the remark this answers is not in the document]
MoKob: I don't think this is an issue, as the substring still needs to match our pre-set database of suffixes/prefixes. Unless incomplete suffixes are stored within that set, we cannot match against what is left.

include/util/guidance/name_announcements.hpp
MoKob: well... it's a bad side-effect of having all the different roads that are just two letters and include `N/E/S/W` in their lettering :(. In the end, this check prevents matching suffixes that are just 1 of two letters. This could be relevant in `1N` `1S`, if that should exist anywhere. In general, it is more helpful on cucumber tests, where we would need to ensure that no name is actually just two letters (which is mostly the case right now)

PR #4427, opened 2017-08-21
Implement exclude flags for CH

TheMarex:
# Issue

This issue aims to implement the second part of #4006 . For this we are modifying the contraction to keep a core and repeat the contraction of that core for every exclude flag combination.

## Tasklist
 - [x] Measure slow-down of non-exclude CH
 - [x] Contract core for every exclude combination
 - [x] Merge separate 
 - [x] review
 - [x] adjust for comments

## Measurements

This branch uses 4 exclude combinations by default (`none`, `motorway`, `ferry`, `toll`)

### Bayern

| Branch | Query time/msec | Contract time/sec |
|-----------|-------------------------|--------------------------|
| `master` | 2.93 | 204 |
| This branch | 3.03 | 390 

### Germany

| Branch | Query time/msec | Contract time/sec | `osrm-contract` RAM/MB | `osrm-routed` RAM/MB | .hsgr size/MB |
|-----------|-------------------------|--------------------------|--------------|---------------------|--------------|
| `master` | 4.18 | 939.018 | 2930.68 | 3440 | 998.72 |
| This branch | 4.71 (x1.12), 8.12 (x1.94, `exclude=motorway`) | 2245.07 (x2.39) | 5390.30 (x1.83) |  4173 (x1.21) | 1277.08 (x1.28) |


include/util/filtered_graph.hpp
```diff
@@ -0,0 +1,165 @@
...
+    void Renumber(const std::vector<NodeID> &old_to_new_node)
+    {
+        graph.Renumber(old_to_new_node);
+        // FIXME the edge filter needs to be renumbered with a different permutation
+        // util::inplacePermutation(edge_filter.begin(), edge_filter.end(), old_to_new_node);
+    }
+
+  private:
+    Graph graph;
+    Vector<bool> edge_filter;
```
deniskoronchik: I think that memory usage can be minimized 8x times (depending on compiler and platform). Just using of bit vector. You need 1 bit to store bool value. Math to access it very simple:
```cpp
bool isFilter(EdgeID eid)
{
    return edge_filter[eid / 8] & (1 << (eid % 8));
}
```

instead `eid / 8` can be used `eid >> 3`. So it would be fast and use `EdgeNumber / 8` bytes of memory

oxidase: @deniskoronchik internally [`Vector<bool> `](https://github.com/Project-OSRM/osrm-backend/blob/993faba545e90a4ae0bd9027119f45b9635176c5/include/util/vector_view.hpp#L142-L206)
```
(gdb) ptype QueryGraph::edge_filter
type = class osrm::util::vector_view<bool> {
  private:
    unsigned int *m_ptr;
    std::size_t m_size;
    static const std::size_t UNSIGNED_BITS;
...
}
```

ATM it is stored in a file as 8 bits per value, but there is a not-yet merged commit 69d53ad6db64207f565dda7c920b9dec5834d5cc to reduce file sizes

src/contractor/graph_contractor.cpp
```diff
@@ -568,28 +547,216 @@ bool GraphContractor::IsNodeIndependent(const std::vector<float> &priorities,
...
+    util::UnbufferedLog log;
+    util::Percent p(log, remaining_nodes.size());
+
+    const util::XORFastHash<> hash;
+
+    unsigned current_level = 0;
+    std::size_t next_renumbering = number_of_nodes * 0.35;
+    while (remaining_nodes.size() > number_of_core_nodes)
+    {
+        if (remaining_nodes.size() < next_renumbering)
```
oxidase: I have bisected failing test on windows and the different ordering comes from this line in 883a8ca501e0359da90e7917346320764de54179 
```
if (remaining_nodes.size() <= next_renumbering)
```
will make reordering at the same time as before because `<` is not equivalent to `>` after integer truncation

src/contractor/graph_contractor.cpp
```diff
@@ -568,28 +547,216 @@ bool GraphContractor::IsNodeIndependent(const std::vector<float> &priorities,
...
+        // contract independent nodes
+        tbb::parallel_for(
+            tbb::blocked_range<NodeID>(
+                begin_independent_nodes_idx, end_independent_nodes_idx, ContractGrainSize),
+            [&](const auto &range) {
+                ContractorThreadData *data = thread_data_list.GetThreadData();
+                for (auto position = range.begin(), end = range.end(); position != end; ++position)
+                {
+                    const NodeID node = remaining_nodes[position].id;
+                    node_data.is_core[node] = false;
```
oxidase: here is a ub due to N4687 26.2.2 /2

src/contractor/graph_contractor.cpp
TheMarex: Hrm interesting. That line only governs when we do a cache-optimizing renumbering. I think this might create an edge case for nodes with same priority. For those nodes we use a bias function https://github.com/Project-OSRM/osrm-backend/pull/4427/files#diff-e32f0fbbf989930fbb6323b16171f940R124 that decides which one wins the tie. The bias function depends on the node id. If we renumber that tie might change since the node ID is different.

This bug was present in the [old code](https://github.com/Project-OSRM/osrm-backend/blob/master/src/contractor/graph_contractor.cpp#L116) as well. This should be solvable by always using the original node ID for the bias function.

src/contractor/graph_contractor.cpp
TheMarex: Gah yeah forgot vector<bool> is not thread save. Okay yeah that explains a lot of weird bugs I have been seeing.

src/contractor/graph_contractor.cpp
oxidase: what wonders me is that it appears only on :windows: 
:+1: on proper tie breaking

TheMarex: Hrm okay changing proper tie breaking doesn't seem to make a difference. This bug also affects OSX. I will try to compile OSRM on Windows, it seems to be reproducible in a consistent manner there?

PR #4436, opened 2017-08-23
Add osm nodes to `nearest` response

DavidAudrain:
# Issue

The goal is to provide OSM node ids in `nearest` response as requested by #2548.

## TODO list
 - [x] update http docs
 - [x] review
 - [ ] adjust for comments

## Requirements / Relations
 Related to PR #2581

### DIFF
--- a/include/engine/api/nearest_api.hpp
+++ b/include/engine/api/nearest_api.hpp
@@ -42,10 +42,12 @@ class NearestAPI final : public BaseAPI
                            auto waypoint = MakeWaypoint(phantom_node);
                            waypoint.values["distance"] = phantom_with_distance.distance;
 
+                           
+
                            util::json::Array nodes;
 
-                           std::uint64_t from_node = static_cast<std::uint64_t>(SPECIAL_OSM_NODEID);
-                           std::uint64_t to_node = static_cast<std::uint64_t>(SPECIAL_OSM_NODEID);
+                           std::uint64_t from_node = 0;
+                           std::uint64_t to_node = 0;
 
                            std::vector<NodeID> forward_geometry;
                            if (phantom_node.forward_segment_id.enabled )
--- a/include/engine/api/nearest_api.hpp
+++ b/include/engine/api/nearest_api.hpp
@@ -42,15 +42,13 @@ class NearestAPI final : public BaseAPI
                            auto waypoint = MakeWaypoint(phantom_node);
                            waypoint.values["distance"] = phantom_with_distance.distance;
 
-                           
-
                            util::json::Array nodes;
 
                            std::uint64_t from_node = 0;
                            std::uint64_t to_node = 0;
 
                            std::vector<NodeID> forward_geometry;
-                           if (phantom_node.forward_segment_id.enabled )
+                           if (phantom_node.forward_segment_id.enabled)
                            {
                                auto segment_id = phantom_node.forward_segment_id.id;
                                const auto geometry_id = facade.GetGeometryIndex(segment_id).id;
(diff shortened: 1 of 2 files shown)


include/engine/api/nearest_api.hpp
```diff
@@ -38,8 +38,50 @@ class NearestAPI final : public BaseAPI
...
+                           {
+                               nodes.values.push_back(to_node);
+                               nodes.values.push_back(from_node);
+                           }
+                           else
+                           {
+                               nodes.values.push_back(from_node);
+                               nodes.values.push_back(to_node);
+                           }
+                           waypoint.values["nodes"] = std::move(nodes);
```
daniel-j-h: Could you assert we never return the special osm id sentinel set above here please

DavidAudrain: The current implementation may return `SPECIAL_OSM_NODEID` . This happens while doing a scan of Monaco area.
Would you suggest using an another special osm id value? or not returning such value in the response?

daniel-j-h: Yeah we should not expose the osrm internal special node id to the outside. We should probably just ail with an error code in this edge case.

TheMarex: Hrm, would be interesting to see how this could happen. From my understanding there is a guarantee that at least one of `forward` and `backward` is enabled. The code in `nearest.cpp` should already bail out with `NoSegement` if we didn't found candidates.

DavidAudrain: Most of edge case are one way case. In that case `reverse_segment_id.enabled` is false. 

I wonder if it would not be better to use `forward_segment_id` only and get the previous node like this
```
auto osm_node_id = facade.GetOSMNodeIDOfNode(geometry[phantom_node.fwd_segment_position - 1]);
```
As @danpat commented in #2581 , the edge case would be when `fwd_segment_position` is zero.  I guess this would require to find segments linked to `forward_segment_id` to access to their geometry. 
Do you think this would be correct?

DavidAudrain: Any feedback on my comment?

TheMarex:
@DavidAudrain sorry I completely forgot about this PR. Anyway I've looked at your changes again and this iteration looks great. However our CI is complaining about code formatting.

Could you rerun `./scripts/format.sh`? You will need `clang-format` version 3.8. Otherwise this looks good to merge to me.

DavidAudrain:
Thank you @TheMarex for your feedback, I ran `./scripts/format.sh` , it should be ok now.

TheMarex:
Great, going to include this in `5.14`.

daniel-j-h:
Ah just noticed, this should probably have a changelog entry and should show up in the release announcements.

DavidAudrain:
@daniel-j-h I've added an entry to the changelog on the same branch. Should I do another PR to submit this modification?

daniel-j-h:
Either that or @TheMarex please include it in the final release changelog if you tag a release the next days.

DraXus:
This is handy, thanks! However there's no mention in the documentation about this new attribute: https://github.com/Project-OSRM/osrm-backend/blob/faff2c774d09a7227c77ae4fa40d22d54bb00b45/docs/http.md#nearest-service

Is there anything left on this pull request? Otherwise, I wouldn't mind to update the docs.

TheMarex:
> Is there anything left on this pull request? Otherwise, I wouldn't mind to update the docs.

Sorry about that, a PR would be much appreciated?

Issue #482, opened 2012-10-18
support relations in lua profiles


emiltin:
for our bike router we need to prioritize certain ways that are part of routes; green routes, regional routes, etc. the routes  are marked in OSM using relations.

maybe the way function could get a list of relations that the way is part of?

this issue is a follow up on #300 by @karme.

eleu:
+1
(any update on this one?)

Skippern:
Actually remaining on a route instead of going off to a different route and later return to the same route would partly solve my #1414 issue, where routing prefer leaving a national main highway and pass through urban areas and later return to the same national main highway later.

emiltin:
hi @TheMarex what's the status on relation parsing?

TheMarex:
Didn't take a look at it yet. Also not really on my todo list for the foreseeable future. Happy to help if anyone wants to give it a stab.

emiltin:
i  implementation a working solution, but alas with, osmiom that's probably outdated

systemed:
@emiltin's solution is really good - I'm using it at cycle.travel. (But that's why I'm still running off 0.3.10!)

emiltin:
it's in the branch: https://github.com/Project-OSRM/osrm-backend/tree/experimental/route_relations.

code changes can be seen here: https://github.com/Project-OSRM/osrm-backend/commit/46d19ac8b4a7a387d6e6b965754ebe4a7c6b0d0b

emiltin:
The approach is to build a (wayid->relation) map when parsing the data. Later when processing each way, the wayid is used to lookup and pass to LUA all releations that the way is part of.From LUA you can then interate the relations and read the tags and roles of eac..

The approach adds little processing overhead, since few ways are part of relations and gives LUA full access to the route data.

emiltin:
@TheMarex would you want to merge the above solution if it was ported to current develop branch?

TheMarex:
@emiltin I'll take a look at the old code. Porting this should be some effort as we moved to libosmium.

emiltin:
yes unfortunately. but the main approach still holds i think.

emiltin:
much of the old code deals with the parsing of the pbf file. this could now be handled by osmium. the lua code could be ported almost as-is i think. what needs updating is the handling of the map between ways and relations.

TheMarex:
[...]

Issue #3304, opened 2016-11-14
Split destinations into name, ref, ref:to, ...


freenerd:
We should do the same name-ref splitting as in https://github.com/Project-OSRM/osrm-backend/issues/2704 for destinations. Right now these are string concatenated but we've seen the need to separate these again (e.g. https://github.com/Project-OSRM/osrm-text-instructions/issues/51) Naming of fields and response is up for debate.

[Example route that would be impacted](http://map.project-osrm.org/?z=18&center=38.938968%2C-77.133164&loc=38.938307%2C-77.133669&loc=38.939542%2C-77.132596&hl=en&alt=0)

```
# OSM Tags
destination: "Chain Bridge;McLean"
destination:ref: "VA 123"

# OSRM Response Now
destinations: "VA 123: Chain Bridge, McLean"

# OSRM Response Fixed
destinations: {
  names: "Chain Bridge, McLean",
  refs: "VA 123"
}
```

freenerd:
Note that MapboxDirections.swift is currently doing this separation [by splitting the string](https://github.com/mapbox/MapboxDirections.swift/pull/91/files#diff-e7d9704608a58f2271345e5450a30008R417).

TheMarex:
Generally I think this is a good idea. Can someone give me some context why we do the `;` -> `,` thing?

On the API side I would like to encode it as:

```
destinations: "Chain Bridge, McLean",
destination_refs: "VA 123"
```

Just because of the object overhead and it feels more consistent with `name`, `ref` and `pronunciation`.

daniel-j-h:
The question still stands if these responses should be arrays instead. Otherwise we're pushing the delimiter-separated list parsing just one level deeper.

E.g. instead of

    destination_refs: "VA 123; VA 456"

should this be

    destination_refs: [
      "VA 123",
      "VA 45"
    ]

The object overhead seems to be the only reason for not doing this.

1ec5:
This issue makes it difficult for a client to determine whether to assign a highway shield to a given step. For example, [this way](https://www.openstreetmap.org/way/28357255) is tagged:

```
destination:ref = CA 87 South
destination:ref:to = I 280
```

with no `destination` tag, because the on-ramp’s signage shows an I-280 shield instead of a textual destination name. A route step that travels along this way has a `destination` of `CA 87 South`.

[...]

PR #4438, opened 2017-08-24
Work on support relations in lua profiles

deniskoronchik:
# Issue

#482 

## Tasklist
 - [x] Add `process_relation` function into lua
 - [x] Add parameter relation_data to `process_way` and `process_node`
 - [x] Process relations logic (C++)
 - [x] Change lua API to 3, an support this
 - [x] Automation tests of new LUA features
 - [x] Update documentation


emiltin:
very nice to see work in relation support :-)

src/extractor/scripting_environment_lua.cpp
```diff
@@ -284,7 +361,7 @@ void Sol2ScriptingEnvironment::InitContext(LuaScriptingContext &context)
                                              "id",
                                              &osmium::Node::id,
                                              "version",
-                                             &osmium::Way::version);
+                                             &osmium::Node::version);
```
TheMarex: Good catch!

deniskoronchik:
This version have some problems:
* old restriction mechanism still using (need to replace it in future);
* can pass just string values to `process_relation` result (maybe can be changed with boost::any)

daniel-j-h:
Great to see this moving forward! As a first use-case we should pass on cardinal directions from the route relation onto it's containing way members.

For now we're good with concatenating the route's cardinal direction (North, South, ..) onto the way's ref property.

In future versions (probably v6) we might want to reconsider this and properly split it off into own properties: https://github.com/Project-OSRM/osrm-backend/issues/3304.

src/extractor/scripting_environment_lua.cpp
```diff
@@ -276,6 +288,71 @@ void Sol2ScriptingEnvironment::InitContext(LuaScriptingContext &context)
...
+            role = member.role();
+            item_type = member.type();
+            id = member.ref();
+        }
+
+        const char* GetRole() const { return role.c_str(); }
+        osmium::item_type GetItemType() const { return item_type; }
+        osmium::object_id_type GetId() const { return id; }
+
+        util::OsmIDTyped ref() const { return util::OsmIDTyped(id, std::uint8_t(item_type)); }
```
oxidase: conversion of osmium::item_type to `std::uint8_t` must be a hidden in `OsmIDTyped`

src/extractor/scripting_environment_lua.cpp
```diff
@@ -276,6 +288,71 @@ void Sol2ScriptingEnvironment::InitContext(LuaScriptingContext &context)
...
+        "type",
+        &util::OsmIDTyped::GetType
+    );
+
+    context.state.new_usertype<RelationMemberWrap>(
+        "RelationMember",
+        "role",
+        &RelationMemberWrap::GetRole,
+        "item_type",
+        &RelationMemberWrap::GetItemType
```
oxidase: `id` is not exposed to lua

src/extractor/scripting_environment_lua.cpp
```diff
@@ -276,6 +288,71 @@ void Sol2ScriptingEnvironment::InitContext(LuaScriptingContext &context)
...
+        }
+
+        void init(const osmium::RelationMember & member)
+        {
+            role = member.role();
+            item_type = member.type();
+            id = member.ref();
+        }
+
+        const char* GetRole() const { return role.c_str(); }
```
oxidase: you don't need accessor function with pubic data members  , it is better to pass references to sol and set r/o access as `sol::readonly( &RelationMemberWrap::role )`

src/extractor/scripting_environment_lua.cpp
```diff
@@ -276,6 +288,71 @@ void Sol2ScriptingEnvironment::InitContext(LuaScriptingContext &context)
...
+                                                 "get_value_by_key",
+                                                 &get_value_by_key<osmium::Relation>,
+                                                 "id",
+                                                 &osmium::Relation::id,
+                                                 "version",
+                                                 &osmium::Relation::version,
+                                                 "members", [](const osmium::Relation &rel)
+    {
+        std::vector<RelationMemberWrap> members(rel.members().size());
+        size_t i = 0;
```
oxidase: `std::copy(...)` of rel.members() to memeber?

src/extractor/scripting_environment_lua.cpp
```diff
@@ -276,6 +288,71 @@ void Sol2ScriptingEnvironment::InitContext(LuaScriptingContext &context)
         "version",
         &osmium::Way::version);
 
+    struct RelationMemberWrap
```
oxidase: Why not  to expose `osmium::RelationMember` directly to avoid copying role strings?

third_party/libosmium/test/t/io/data-cr.opl
```diff
@@ -1 +1 @@
-n1 v1 dV c1 t2014-01-01T00:00:00Z i1 utest T x1.02 y1.02
```
oxidase: This must be reverted

docs/profiles.md
```diff
@@ -130,7 +130,7 @@ Attribute       | Type    | Notes
 barrier         | Boolean | Is it an impassable barrier?
 traffic_lights  | Boolean | Is it a traffic light (incurs delay in `process_turn`)?
 
-## process_way(profile, way, result)
+### process_way(profile, way, result)
```
oxidase: form the code `process_way` has the 4th `relations` argument but here it is not mentioned

Should it be may be optional ? I feel we need some sort of sync here with location-dependent tags, that I would like to see optional. So we can add an additional positional argument that will hold a table with optional data items.
This will save from bumping api version every time when a new piece of data will be added to lua callback functions.

include/util/osm_id_typed.hpp
```diff
@@ -0,0 +1,40 @@
...
+#define OSM_ID_TYPED_HPP
+
+#include "typedefs.hpp"
+
+namespace osrm
+{
+namespace util
+{
+
+class OsmIDTyped
```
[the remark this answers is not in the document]
deniskoronchik: I've thinked about this. And decide that it's not normal for OSRM (haven't seen union in project). If it good, i can do that.

src/extractor/extractor.cpp
```diff
@@ -391,13 +394,44 @@ Extractor::ParseOSMData(ScriptingEnvironment &scripting_environment,
...
+                relations.AddRelation(result.second);
+            }
+
             for (const auto &result : parsed_buffer->resulting_restrictions)
             {
                 extractor_callbacks->ProcessRestriction(result);
             }
         });
 
+    /* Main trick that we can use the same pipeline. It just receive relation objects
```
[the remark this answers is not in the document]
deniskoronchik: yeah, that also was an idea. But there are some code that request reader before: https://github.com/Project-OSRM/osrm-backend/pull/4438/commits/ed5e00f05a714c89ed98c96719bed410ea849e3f#diff-edf14d92e413123a2b940ab0b8491ffbR298

src/extractor/scripting_environment_lua.cpp
deniskoronchik: Because it has deleted a lot of constructors and operators. So it doesn't compiles with sol.
I've tried for example:
```cpp
context.state.new_usertype<osmium::Relation>("Relation",
                                                 "get_value_by_key",
                                                 &get_value_by_key<osmium::Relation>,
                                                 "id",
                                                 &osmium::Relation::id,
                                                 "version",
                                                 &osmium::Relation::version,
                                                 "members",
                                                 [](const osmium::Relation &rel) {
                                                     return sol::as_table(rel.members());
                                                 });
```

src/extractor/scripting_environment_lua.cpp
deniskoronchik: You talk about store `std::uint8_t` instead of `osmium::item_type` or something else?

third_party/libosmium/test/t/io/data-cr.opl
deniskoronchik: Can't do that until will be fixed in master. It create this one automatically with git on any rebase.

src/extractor/scripting_environment_lua.cpp
deniskoronchik: done

src/extractor/scripting_environment_lua.cpp
deniskoronchik: done

docs/profiles.md
```diff
@@ -177,10 +177,40 @@ road_classification.road_priority_class | Enum     | Guidance: order in priority
...
+### process_relation(profile, relation, result)
+
+Supported since API **version 3**.
+
+Given an OpenStreetMap relation, the `process_relation` function should setup values into result structure.
+
+Argument | Description
+---------|-------------------------------------------------------
+profile  | The configuration table you returned in `setup`.
+node     | The input relation to process (read-only).
```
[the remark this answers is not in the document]
deniskoronchik: done

include/extractor/extraction_relation.hpp
```diff
@@ -0,0 +1,85 @@
...
+#include <vector>
+
+namespace osrm
+{
+namespace extractor
+{
+namespace detail
+{
+
+inline const char * checkedString(const char * str)
```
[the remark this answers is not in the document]
deniskoronchik: done

deniskoronchik:
@oxidase have test features of `api_version` 3: https://github.com/Project-OSRM/osrm-backend/pull/4438/files#diff-c86e1caf4e4a3c911f176b8f8b1c0d87

Or need more and if yes, then where to place them?

oxidase:
@deniskoronchik test looks good, somehow i missed it during review

include/util/osm_id_typed.hpp
```diff
@@ -0,0 +1,38 @@
...
+        BOOST_ASSERT(id_ < (std::uint64_t(1) << 56));
+    }
+
+    bool operator==(const OsmIDTyped &other) { return (id == other.id && type == other.type); }
+    bool operator!=(const OsmIDTyped &other) { return (id != other.id || type != other.type); }
+
+    inline HashType Hash() const { return (std::uint64_t(id) | std::uint64_t(type) << 56); }
+
+    std::uint64_t GetID() const { return id; }
+    std::uint8_t GetType() const { return type; }
```
danpat: I think it would be good to do what osmium does here and make `type` an enum.  We should have an explicit mapping from the `osmium::item_type` enum values to our own internal values.

src/extractor/scripting_environment_lua.cpp
```diff
@@ -276,6 +336,61 @@ void Sol2ScriptingEnvironment::InitContext(LuaScriptingContext &context)
...
+            role = member.role();
+            item_type = member.type();
+            id = member.ref();
+        }
+
+        const char *GetRole() const { return role.c_str(); }
+        osmium::item_type GetItemType() const { return item_type; }
+        osmium::object_id_type GetId() const { return id; }
+
+        util::OsmIDTyped ref() const { return util::OsmIDTyped(id, std::uint8_t(item_type)); }
```
danpat: This is the other side to my comment on `OsmIDTyped` - I think we should be very explicit about what values we store here.  Something like:

```
switch(item_type) {
  case osmium::item_type.node:
      return util::OsmIDTyped(id, osrm::extractor::OSMIDType.NODE);
  case osmium::item_type.way:
      return util::OsmIDTyped(id, osrm::extractor::OSMIDType.WAY);
  default:
     throw util::exception("some error message");
}
```

deniskoronchik: Will do our enum with conversion

1ec5:
> Great to see this moving forward! As a first use-case we should pass on cardinal directions from the route relation onto it's containing way members.
> 
> For now we're good with concatenating the route's cardinal direction (North, South, ..) onto the way's ref property.
> 
> In future versions (probably v6) we might want to reconsider this and properly split it off into own properties: #3304.

Note that this approach will produce incorrect refs for many [route concurrencies](https://wiki.openstreetmap.org/wiki/Highway_Directions_In_The_United_States#Concurrent_routes) in the U.S: https://github.com/Project-OSRM/osrm-text-instructions/issues/119#issuecomment-328053084.

src/extractor/scripting_environment_lua.cpp
```diff
@@ -31,13 +32,64 @@ template <> struct is_container<osmium::Node> : std::false_type
...
+{
+};
 }
 
 namespace osrm
 {
 namespace extractor
 {
 
+class RelationsContainerWrap
```
oxidase: The wrapper is not needed and can be removed. Also `context.state.new_usertype<RelationsContainerWrap>( ...` can be removed. Relations can be forwarded to lua as

```
node_function(profile_table, node, result, sol::as_table(relations));
...
way_function(profile_table, way, result, sol::as_table(relations));sol::as_table(relations)
```

src/extractor/scripting_environment_lua.cpp
```diff
@@ -366,6 +499,45 @@ void Sol2ScriptingEnvironment::InitContext(LuaScriptingContext &context)
...
+    struct ExtractionRelationData
+    {
+        explicit ExtractionRelationData(ExtractionRelation::AttributesMap &attrs_) : attrs(attrs_)
+        {
+        }
+
+        ExtractionRelation::AttributesMap &attrs;
+    };
+
+    context.state.new_usertype<RelationsContainerWrap>(
```
oxidase: Not needed

src/extractor/scripting_environment_lua.cpp
```diff
@@ -366,6 +499,45 @@ void Sol2ScriptingEnvironment::InitContext(LuaScriptingContext &context)
...
+        "RelationsContainerWrap",
+        sol::meta_function::index,
+        [](const RelationsContainerWrap &rel, size_t index) { return rel.GetAttributes(index); });
+
+    context.state.new_usertype<ExtractionRelationData>(
```
oxidase: not needed

src/extractor/scripting_environment_lua.cpp
```diff
@@ -366,6 +499,45 @@ void Sol2ScriptingEnvironment::InitContext(LuaScriptingContext &context)
...
+        [](ExtractionRelationData &data, const std::string &key, sol::stack_object object) {
+            return data.attrs[key] = object.as<std::string>();
+        },
+        sol::meta_function::index,
+        [](ExtractionRelationData &data, const std::string &key) { return data.attrs[key]; });
+
+    context.state.new_usertype<ExtractionRelation>(
+        "ExtractionRelation",
+        sol::meta_function::new_index,
+        [](ExtractionRelation &rel, const RelationMemberWrap &member) {
```
oxidase: add `-> ExtractionRelation::AttributesMap&`

src/extractor/scripting_environment_lua.cpp
```diff
@@ -366,6 +499,45 @@ void Sol2ScriptingEnvironment::InitContext(LuaScriptingContext &context)
...
+            return ExtractionRelationData(rel.GetMember(member.ref()));
+        },
+        sol::meta_function::index,
+        [](ExtractionRelation &rel, const RelationMemberWrap &member) {
```
oxidase: add `add `-> ExtractionRelation::AttributesMap&``

src/extractor/scripting_environment_lua.cpp
oxidase: The wrapper can be removed `return sol::as_table(rel.members());` should work well

oxidase:
@deniskoronchik @TheMarex i have removed Lua 5.1 support because https://github.com/ThePhD/sol2/issues/302 makes relations wrapping too complicated and requires copying of members for every node/way.

ATM windows build failed, because it requires Lua 5.1 -> 5.2 update. 
I will try to build dependencies locally and update osrm-deps-win-x64-14.0.7z

Issue #4432, opened 2017-08-23
macOS binaries for node v6 missing


daniel-j-h:
From https://github.com/Project-OSRM/osrm-backend/issues/4431: it seems like we do not build binaries for macOS Node.js v6. Here's what we published for v5.11.0 (vXY are Node.js ABI version numbers corresponding to Node 4 (v46) and Node 6 (v48)):

```
node-v46-darwin-x64.tar.gz
node-v46-linux-x64.tar.gz
node-v48-linux-x64.tar.gz
```

Here's our macOS Travis build:

https://github.com/Project-OSRM/osrm-backend/blob/e7abe37b1093f14d6b1a03d3a29706d3c97aa8a0/.travis.yml#L143-L149

danpat:
Ugh - the OSX build is currently by far the slowest, *and* we have to wait the longest for the build to start.  Adding another OSX build configuration would make things even slower.

I don't see any way around it though, we should add the Node v6 configuration.

Jma353:
Any follow-up on this?  Would love to have a working configuration :)

danpat:
@Jma353 I've just opened a PR to do this, but it'll only take effect once we push the next OSRM release out.

In the meantime, you can always build locally.  Try the following:

```
git clone https://github.com/Project-OSRM/osrm-backend.git
cd osrm-backend
mkdir build
cd build
cmake -DENABLE_MASON=ON -DENABLE_NODE_BINDINGS=ON ..
make
cd ..
npm link
```

This should build everything locally, and link the build libraries into your global `node_modules`.  You should then be able to `npm install osrm` from other projects locally, and it'll use your local build.

daniel-j-h:
This got fixed in https://github.com/Project-OSRM/osrm-backend/pull/4440 - closing here.

PR #4440, opened 2017-08-25
Enable OSX binary publishing for both NodeJS 4 and NodeJS 6

danpat:
# Issue

This adds a Travis configuration to build on OSX with NodeJS 6.  Previously, we were only building there with NodeJS 4, and only publishing the NodeJS 4 module.  Users on OSX with NodeJS 6 were unable to do `npm install osrm` and use the pre-built binaries.

https://github.com/Project-OSRM/osrm-backend/issues/4432

## Tasklist
 - [x] review
 - [ ] adjust for comments

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -144,7 +144,15 @@ matrix:
       osx_image: xcode8.2
       compiler: "mason-osx-release"
       # we use the xcode provides clang and don't install our own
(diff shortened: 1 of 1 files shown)


daniel-j-h:
Do we need https://github.com/Project-OSRM/osrm-backend/blob/ef8124651822445a3d81d2116b67095448fefe18/.travis.yml#L10-L12 when we explicitly install Node.js versions manually by now?

Feel free to merge in - next release should pick it up.

danpat:
@daniel-j-h good catch, I've removed that clause, it had no effect that I can see.

PR #4441, opened 2017-08-25
[EXPERIMENTAL] Supply additional name hint to guide coordinate snapping

danpat:
# Issue

This PR adds a name hint to help guide coordinate snapping.  If supplied as part of the request, coordinates will only snap to roads that are "similar" (to be defined) to the supplied name hint.

The new parameter is of the form `name_hint=main st;destination st`.  I'm still toying with the actual filtering/selection logic, suggestions welcome.

## Why?
Geocoders don't always return a coordinate that's ideal for routing purposes.  It's not always closest to the intended destination road.

A good example is this address:

http://www.openstreetmap.org/node/596865687

which you can search for with Nominatim as "4850 Edgeware Road".  The data is sourced from the San Diego GIS department and published as public domain data.

Unfortunately, OSRM snaps this coordinate to the alley behind the house, which is closer to the geocoded coordinate than Edgeware Road.

## Tasklist
 - [ ] Make matches case insensitive
 - [ ] Consider stripping prefix/suffix
 - [ ] Limit r-tree search when no nearby names match to avoid runaway searches
 - [ ] Avoid re-checking segments from the same named road that don't match
 - [ ] Finalize selection logic
 - [ ] Consider API/parameter naming
 - [ ] Implement for all `Nearest*` methods
 - [ ] Make hints optional for each coordinate (i.e. blank = no name filtering)
 - [ ] Benchmark
 - [ ] review
 - [ ] adjust for comments

### DIFF
--- a/features/step_definitions/routability.js
+++ b/features/step_definitions/routability.js
@@ -121,7 +121,7 @@ module.exports = function () {
 
             r.which = dir;
 
-            this.requestRoute((dir === 'forw' ? [a, b] : [b, a]), [], [], this.queryParams, (err, res, body) => {
+            this.requestRoute((dir === 'forw' ? [a, b] : [b, a]), [], [], [], this.queryParams, (err, res, body) => {
                 if (err) return callback(err);
 
                 r.query = this.query;
--- a/features/support/route.js
+++ b/features/support/route.js
@@ -46,9 +46,10 @@ module.exports = function () {
         return waypoints.map(w => [w.lon, w.lat].map(ensureDecimal).join(','));
     };
 
-    this.requestRoute = (waypoints, bearings, approaches, userParams, callback) => {
+    this.requestRoute = (waypoints, bearings, approaches, namehints, userParams, callback) => {
         if (bearings.length && bearings.length !== waypoints.length) throw new Error('*** number of bearings does not equal the number of waypoints');
         if (approaches.length && approaches.length !== waypoints.length) throw new Error('*** number of approaches does not equal the number of waypoints');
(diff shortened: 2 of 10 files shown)


emiltin:
some form of edit distance? https://en.wikipedia.org/wiki/Edit_distance

Komzpa:
Please keep in mind ex-USSR block architecture, when there are a lot of unnamed passages inside.

http://map.project-osrm.org/?z=17&center=53.888261%2C27.686223&loc=53.891189%2C27.688218&loc=53.888223%2C27.684109&hl=en&alt=0


For instance, this:
http://map.project-osrm.org/?z=17&center=53.863227%2C27.611625&loc=53.866245%2C27.609962&loc=53.861613%2C27.610348&hl=en&alt=0

Sould not become this:
http://map.project-osrm.org/?z=18&center=53.863439%2C27.608535&loc=53.864372%2C27.611293&loc=53.863391%2C27.608932&hl=en&alt=0

daniel-j-h:
@danpat is this still actionable or can we close?

emiltin:
curious why this was abandoned?

daniel-j-h:
Scope creep - we're not building a geocoder ;)

emiltin:
ok makes sense!

gardster:
I'm really want to have any instrument to handle cases like https://www.openstreetmap.org/way/265282556
http://map.project-osrm.org/?z=18&center=40.758225%2C-73.996798&loc=40.757676%2C-73.998080&loc=40.760293%2C-73.994390&hl=en&alt=0 Maybe you will use any other feature, like osm_id of the way or the exact street name (that I will have from the external geocoder).

ericrwolfe:
@gardster great example. Addresses near tunnels and bridges are a huge problem right now.

PR #4442, opened 2017-08-25
Add support for loading arbitary OSM files in tests

danpat:
# Issue

Creating cucumber test scenarios can be cumbersome.  We often come across situations in OSM where various bits of our processing, guidance, or routing break down.

This PR enables loading arbitary OSM files into cucumber scenarios with:

```
Given the input file XXXXX.osm
```

and using hand-crafted coordinates as the `from` and `to` values when routing:

```
When I route I should get
    | from                | to                 | turns         |
    | -77.016381,38.91594 | -77.01524,38.91591 | depart,arrive |
```

This should make it simpler to import discovered weird cases from OSM, and capture a store of real-world map models to test our engine against.

**NOTE** the example test I've included in this PR so far is not really useful.  I've included it as an example of how this works, but it should be removed (or corrected) before this PR is merged.

## Tasklist
 - [ ] review
 - [ ] adjust for comments

### DIFF
--- a/features/real/roundabouts.feature
+++ b/features/real/roundabouts.feature
@@ -1,0 +1,12 @@
+@routing @car @bugs
+Feature: Real test scenarios from OSM
+
+    Background:
+        Given the profile "car"
+
+    Scenario: Routing across an untagged roundabout
+        Given the input file test/data/extracts/untagged_roundabout.osm
+
+        When I route I should get
+            | from                | to                 | turns                    | route                                                      |
+            | -77.016381,38.91594 | 38.91591,-77.01524 | depart,turn right,arrive | 3rd Street Northwest,T Street Northwest,T Street Northwest |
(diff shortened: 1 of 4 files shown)


emiltin:
i understand the intention, however i a bit concerned that:
1) it makes tests harder to inspect and understand. a separate osm editor is required to view the test data
2) real-world examples are often complex, including tags and ways that might or might not affect the test results, making it hard to tests specific aspects of the routing behaviour.
3) it will increase the size of tests a lot

oxidase:
@danpat the main problem here is that tests no more minimal representative examples and introduce dependency on other code parts due to non-relevant tags. I think overhead of debugging such tests if they fail due to unrelated code change is much larger than cleaning up OSM and using minimum required tags/coordinates to reproduce the issue.

daniel-j-h:
@danpat is this still actionable or can we close?

PR #4443, opened 2017-08-25
Respect initial bearings order

### LINKED ISSUE #4331: Turn angle calculation needs to respect initial road order

oxidase:
# Issue

PR fixes #4331.

For the sample intersection

| way | initial bearing | extracted bearing |
|---|---|---|
| [394198761](https://www.openstreetmap.org/way/394198761) | 325.895 | 322.542 |
| [118426709](https://www.openstreetmap.org/way/118426709) | 162.921 | 163.646 |
| [6323556](https://www.openstreetmap.org/way/6323556) | 151.491 | 170.53  |

PR adds ordering by initial bearings and adjustment of bearings to keep the original order.
For the above example, bearings ordered in the initial order are `322.542, 170.53, 163.646` and after the adjustment are `322.542,170.53,171.03`.

The adjustment angle is at most 0.5 degree or a half-angle between the road bearing and the base bearing.

## Tasklist
 - [ ] review
 - [ ] adjust for comments

### DIFF
--- a/src/extractor/guidance/intersection_generator.cpp
+++ b/src/extractor/guidance/intersection_generator.cpp
@@ -54,11 +54,18 @@ IntersectionGenerator::ComputeIntersectionShape(const NodeID node_at_center_of_i
                                                 const boost::optional<NodeID> sorting_base,
                                                 const bool use_low_precision_angles) const
 {
-    IntersectionShape intersection;
-    // reserve enough items (+ the possibly missing u-turn edge)
     const auto intersection_degree = node_based_graph.GetOutDegree(node_at_center_of_intersection);
-    intersection.reserve(intersection_degree);
     const util::Coordinate turn_coordinate = coordinates[node_at_center_of_intersection];
+
+    // compute bearings in a relatively small circle to prevent wrong roads order with true bearings
+    struct RoadWithInitialBearing
+    {
+        double bearing;
+        IntersectionShapeData road;
+    };
+    std::vector<RoadWithInitialBearing> initial_roads_ordering;
+    // reserve enough items (+ the possibly missing u-turn edge)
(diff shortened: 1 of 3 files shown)


src/extractor/guidance/intersection_generator.cpp
```diff
@@ -123,31 +135,71 @@ IntersectionGenerator::ComputeIntersectionShape(const NodeID node_at_center_of_i
...
             }
-            return util::bearing::reverse(intersection.begin()->bearing);
+            return util::bearing::reverse(initial_roads_ordering.begin()->bearing);
         }();
-        std::sort(intersection.begin(),
-                  intersection.end(),
-                  makeCompareShapeDataAngleToBearing(base_bearing));
+
+        // sort roads with respect to the initial bearings, tie-breaker for equal initial bearings
+        // is to order CCW road before CW one
```
MoKob: @oxidase personally I do not understand what wold be a cw road/ccw road. Could you clarify here a bit?

oxidase:
@MoKob i have updated comment, we have to keep roads in the clockwise order with respect to "final" bearings if initial bearings are equal. This is the original behavior, that potentially should not be observed in real data, but in some tests where roads can overlap
```
features/guidance/perception.feature:76
features/guidance/roundabout-turn.feature:247
features/guidance/roundabout.feature:770
```

PR #4452, opened 2017-08-28
Use user-provided osmium::thread::Pool instance

mloskot:
# Issue

Advantages:
* Allows to use `Pool` configured with user-preferred number of threads.
* If libosmium used in a shared library, this helps to prevent potential deadlock during `Pool` shutdown at (after) shared library unloading.

See https://github.com/osmcode/libosmium/pull/214

## Tasklist

 - [x] review
 - [x] adjust for comments

## Requirements / Relations

Requires libosmium **2.13.0**

### DIFF
--- a/src/extractor/extractor.cpp
+++ b/src/extractor/extractor.cpp
@@ -40,6 +40,7 @@
 #include <boost/scope_exit.hpp>
 
 #include <osmium/io/any_input.hpp>
+#include <osmium/thread/pool.hpp>
 
 #include <tbb/pipeline.h>
 #include <tbb/task_scheduler_init.h>
@@ -286,9 +287,11 @@ Extractor::ParseOSMData(ScriptingEnvironment &scripting_environment,
     util::Log() << "Threads: " << number_of_threads;
 
     const osmium::io::File input_file(config.input_path.string());
-
+    osmium::thread::Pool pool(number_of_threads);
     osmium::io::Reader reader(
-        input_file, (config.use_metadata ? osmium::io::read_meta::yes : osmium::io::read_meta::no));
+        input_file,
+        pool,
+        (config.use_metadata ? osmium::io::read_meta::yes : osmium::io::read_meta::no));
 
     const osmium::io::Header header = reader.header();
 


daniel-j-h:
There's an env var for setting the libosmium number of threads:

http://osmcode.org/libosmium/manual.html#run-time-configuration

How does being explicit here interact with the env var?

mloskot:
Explicit number of threads argument overrides the env var.

However, this PR proposes to use system-specific default, that is `Pool pool(tbb::task_scheduler_init::default_num_threads())` which is equivalent to not setting the `OSMIUM_POOL_THREADS`.

-----

BTW, the 2.13.0 did not change how `osmium::thread::Pool` calculates number of threads, it still does:

```
static constexpr int default_num_threads = 0;

/**
    * Create thread pool with the given number of threads. If
    * num_threads is 0, the number of threads is read from
    * the environment variable OSMIUM_POOL_THREADS. The default
    * value in that case is -2.
    *
    * If the number of threads is a negative number, it will be
    * set to the actual number of cores on the system plus the
    * given number, ie it will leave a number of cores unused.
    *
    * In all cases the minimum number of threads in the pool is 1.
    *
    * If max_queue_size is 0, the queue size is read from
    * the environment variable OSMIUM_MAX_WORK_QUEUE_SIZE.
    */
explicit Pool(int num_threads = default_num_threads, ...)
```

If, however, one sets eg. `OSMIUM_POOL_THREADS=8` then this will not beTBB defa

TheMarex:
I think this will be fine from our deployment side, since we already set both the envar and the cli-parameter. Actually this might be less confusing then having to set both.

TheMarex:
> /home/travis/build/Project-OSRM/osrm-backend/src/extractor/extractor.cpp:290:48: error: no matching function for call to ‘osmium::thread::Pool::Pool(const unsigned int&)’

This is what I get from CI. Is there a libosmium mismatch?

mloskot:
@TheMarex This PR assumes libosmium update to 2.13.0 or later

TheMarex:
@mloskot okay updated `libosium`, should work after a rebase.

Issue #4239, opened 2017-07-05
Pass bearing information for roads from one point.


gardster:
There is no bearing information in the steps maneuver section when track matched in a single point.

https://router.project-osrm.org/match/v1/driving/-73.866287,40.652139;-73.866312,40.652192?steps=true

But there is a `hint` information in the answer. So the OSRM has edge and this edge has a geometry. I want to see bearing based on this geometry.

gardster:
Fixed

PR #4453, opened 2017-08-29
Keep bearing information in one point track

gardster:
# Issue

#4239

There is no possibilities to obtain bearing information on a `assemleSteps` phase if track geometry consists from a single point (several copies of the same point). So I save bearing information inside a `PhantomNode` structure and use it if there is no other sources.


TheMarex:
I'm fine with adding this. 👍 I'm just curious: Why do you need this?

gardster:
I'm using the OSRM for realtime map matching. Sometimes monitored vehicles stayed at one place for a long time. This type of a GPS track matches into a single point and has not bearing information in the answer. So all vehicles of such type will have bearing to the north in the user interface (we must draw em anyway). So I want to draw them along the nearest road. Unfortunately the nearest plugin has not bearing information too, so I couldn't workaroud this behaviour with a second API request.

TheMarex:
@gardster Adding bearing to `nearest` would also be interesting btw.

PR #4454, opened 2017-08-29
Parallelize ManyToMany plugin

oxidase:
# Issue

Backward and forward searches in manyToMany plugin are independent and can be easily parallelized.
Here is timing results in milliseconds with 4 cores (2 real + 2 HT) for 25x25 random table requests for DE-sized extract:

par_ser

## Tasklist
 - [x] review
 - [ ] adjust for comments

## Requirements / Relations

### DIFF
--- a/include/engine/engine.hpp
+++ b/include/engine/engine.hpp
@@ -24,6 +24,8 @@
 #include "util/fingerprint.hpp"
 #include "util/json_container.hpp"
 
+#include <tbb/task_scheduler_init.h>
+
 #include <memory>
 #include <string>
 
@@ -53,7 +55,8 @@ template <typename Algorithm> class Engine final : public EngineInterface
 {
   public:
     explicit Engine(const EngineConfig &config)
-        : route_plugin(config.max_locations_viaroute, config.max_alternatives), //
+        : task_scheduler(config.use_threads_number),
+          route_plugin(config.max_locations_viaroute, config.max_alternatives), //
           table_plugin(config.max_locations_distance_table),                    //
           nearest_plugin(config.max_results_nearest),                           //
           trip_plugin(config.max_locations_trip),                               //
@@ -125,6 +128,7 @@ template <typename Algorithm> class Engine final : public EngineInterface
     }
     std::unique_ptr<DataFacadeProvider<Algorithm>> facade_provider;
     mutable SearchEngineData<Algorithm> heaps;
+    tbb::task_scheduler_init task_scheduler;
 
     const plugins::ViaRoutePlugin route_plugin;
     const plugins::TablePlugin table_plugin;
--- a/include/engine/engine_config.hpp
+++ b/include/engine/engine_config.hpp
@@ -90,6 +90,8 @@ struct EngineConfig final
     int max_alternatives = 3; // set an arbitrary upper bound; can be adjusted by user
     bool use_shared_memory = true;
     Algorithm algorithm = Algorithm::CH;
+    int use_threads_number = 1;
+    std::string verbosity;
 };
 }
 }
--- a/src/engine/routing_algorithms/many_to_many.cpp
+++ b/src/engine/routing_algorithms/many_to_many.cpp
@@ -301,27 +301,38 @@ template <typename Algorithm>
 std::vector<EdgeWeight> manyToManySearch(SearchEngineData<Algorithm> &engine_working_data,
                                          const DataFacade<Algorithm> &facade,
                                          const std::vector<PhantomNode> &phantom_nodes,
-                                         const std::vector<std::size_t> &source_indices,
-                                         const std::vector<std::size_t> &target_indices)
+                                         std::vector<std::size_t> source_indices,
+                                         std::vector<std::size_t> target_indices)
 {
-    const auto number_of_sources =
-        source_indices.empty() ? phantom_nodes.size() : source_indices.size();
-    const auto number_of_targets =
-        target_indices.empty() ? phantom_nodes.size() : target_indices.size();
+    if (source_indices.empty())
+    {
+        source_indices.resize(phantom_nodes.size());
+        std::iota(source_indices.begin(), source_indices.end(), 0);
+    }
+    if (target_indices.empty())
+    {
+        target_indices.resize(phantom_nodes.size());
+        std::iota(target_indices.begin(), target_indices.end(), 0);
+    }
+
+    const auto number_of_sources = source_indices.size();
+    const auto number_of_targets = target_indices.size();
     const auto number_of_entries = number_of_sources * number_of_targets;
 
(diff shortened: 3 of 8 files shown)


include/engine/engine.hpp
```diff
@@ -125,6 +127,7 @@ template <typename Algorithm> class Engine final : public EngineInterface
     }
     std::unique_ptr<DataFacadeProvider<Algorithm>> facade_provider;
     mutable SearchEngineData<Algorithm> heaps;
+    tbb::task_scheduler_init task_scheduler;
```
daniel-j-h: Needs to be constructed with num threads otherwise default ctor will use available cpus to infer number - which will be wrong e.g. in a Docker container: 

https://www.threadingbuildingblocks.org/docs/help/hh_goto.htm?index.htm#reference/task_scheduler/task_scheduler_init_cls.html

oxidase: @daniel-j-h i added `use_threads_number` parameter that can be set in node bindings and in command line arguments of `osrm-routed`. `osrm-routed` now has two separate thread pools: server and internal tbb, so number of threads is actually doubled.

daniel-j-h: It it? I thought the benefit of using the dynamically linked libtbb was to avoid exactly these issues?

include/engine/engine_config.hpp
```diff
@@ -90,6 +90,7 @@ struct EngineConfig final
     int max_alternatives = 3; // set an arbitrary upper bound; can be adjusted by user
     bool use_shared_memory = true;
     Algorithm algorithm = Algorithm::CH;
+    int use_threads_number = 1;
```
TheMarex: Nitpick: `use_threads_number` is kind of a weird phrasing. `number_of_threads` or just `threads`.

We should also document a value that would make TBB use the default number of threads (I think `-1`?).

src/tools/routed.cpp
```diff
@@ -71,16 +81,8 @@ inline unsigned generateServerProgramOptions(const int argc,
...
-                                             bool &use_shared_memory,
-                                             std::string &algorithm,
                                              bool &trial,
-                                             int &max_locations_trip,
-                                             int &max_locations_viaroute,
-                                             int &max_locations_distance_table,
-                                             int &max_locations_map_matching,
-                                             int &max_results_nearest,
-                                             int &max_alternatives)
+                                             EngineConfig &config)
```
TheMarex: Nice! 👍

src/engine/routing_algorithms/many_to_many.cpp
```diff
@@ -229,49 +246,47 @@ void forwardRoutingStep(const DataFacade<Algorithm> &facade,
...
-    const EdgeWeight source_duration = query_heap.GetData(node).duration;
+    const auto node = query_heap.DeleteMin();
+    const auto source_weight = query_heap.GetKey(node);
+    const auto source_duration = query_heap.GetData(node).duration;
 
     // check if each encountered node has an entry
-    const auto bucket_iterator = search_space_with_buckets.find(node);
-    // iterate bucket if there exists one
-    if (bucket_iterator != search_space_with_buckets.end())
+    const auto &bucket_list = std::equal_range(search_space_with_buckets.begin(),
```
TheMarex: Did you check the performance impact of this over using a unordered_map? `2log(N)` might be fine though.

oxidase:
@daniel-j-j let me show what i mean by "double number of threads". In my local run when i stop in `manyToManySearch` `osrm-routed` has the following threads

```
 Id   Target Id         Frame 
  1    Thread 0x7ffff7f96740 (LWP 11822) "osrm-routed" do_sigwait (sig=0x7fffffffc430, set=<optimized out>) at ../sysdeps/unix/sysv/linux/sigwait.c:64
  2    Thread 0x7ffff070e700 (LWP 12263) "osrm-routed" 0x00007ffff6ecd9dd in pthread_join (threadid=140737217296128, thread_return=0x0) at pthread_join.c:90
  3    Thread 0x7fffefd7f700 (LWP 12264) "osrm-routed" 0x00007ffff5f1a373 in epoll_wait () at ../sysdeps/unix/syscall-template.S:84
* 4    Thread 0x7fffef3f0700 (LWP 12265) "osrm-routed" osrm::engine::routing_algorithms::manyToManySearch<osrm::engine::routing_algorithms::ch::Algorithm> (engine_working_data=..., facade=..., phantom_nodes=std::vector of length 25, capacity 25 = {...}, source_indices=std::vector of length 0, capacity 0, target_indices=std::vector of length 0, capacity 0) at /home/miha/mapbox/osrm-backend/src/engine/routing_algorithms/many_to_many.cpp:322
  5    Thread 0x7fffeea61700 (LWP 12266) "osrm-routed" pthread_cond_wait@@GLIBC_2.3.2 () at ../sysdeps/unix/sysv/linux/x86_64/pthread_cond_wait.S:185
  6    Thread 0x7fffee0d2700 (LWP 12267) "osrm-routed" pthread_cond_wait@@GLIBC_2.3.2 () at ../sysdeps/unix/sysv/linux/x86_64/pthread_cond_wait.S:185
  7    Thread 0x7fffed743700 (LWP 12773) "osrm-routed" syscall () at ../sysdeps/unix/sysv/linux/x86_64/syscall.S:38
  8    Thread 0x7fffecf41700 (LWP 12774) "osrm-routed" syscall () at ../sysdeps/unix/sysv/linux/x86_64/syscall.S:38
  9    Thread 0x7fffed342700 (LWP 12775) "osrm-routed" syscall () at ../sysdeps/unix/sysv/linux/x86_64/syscall.S:38
```


The first thread is the main one and waits for a signal at https://github.com/Project-OSRM/osrm-backend/blob/b36122511cf04e119e3e6ee1c77e937e96b6f33a/src/tools/routed.cpp#L303

The second thread is a server thread that starts at https://github.com/Project-OSRM/osrm-backend/blob/b36122511cf04e119e3e6ee1c77e937e96b6f33a/src/tools/routed.cpp#L288 and waits at https://github.com/Project-OSRM/osrm-backend/blob/b36122511cf04e119e3e6ee1c77e937e96b6f33a/include/server/server.hpp#L82

Threads 3-6 are asio services started at https://github.com/Project-OSRM/osrm-backend/blob/b36122511cf04e119e3e6ee1c77e937e96b6f33a/include/server/server.hpp#L76

Threads 7-8 are started by TBB during the first call of https://github.com/Project-OSRM/osrm-backend/blob/b36122511cf04e119e3e6ee1c77e937e96b6f33a/src/engine/routing_algorithms/many_to_many.cpp#L344

Without the PR `osrm-routed` uses 6 threads in my particular case, with the PR 6 + 3 threads in TBB pool.

Also as we checked yesterday there is no static TBB distributions, so it should be safe to assume a unique TBB data singleton.

Also changing a map of vectors to a ordered vector leads to change of average query time for germany latest from 568.7661ms to 496.6084ms

PR #4460, opened 2017-08-31
Adds turn instructions to the turns layer in debug tiles.

danpat:
# Issue

This change adds two new properties to the `turns` layer in the `tile` plugin - `type` and `modifier`, which mirror the values we see in those properties in the StepManeuver object in our API response.

Showing these on the debug tiles is handy for debugging instruction collapsing/supression - the ties show all maneuvers (including the internal types), so when debugging situations where we're missing instructions, or not collapsing, it's easy to dig into the actual values on the routing graph.

## TODO:
  - [x] string encoding for internal type values

## Tasklist
 - [x] review
 - [x] adjust for comments


src/engine/plugins/tile.cpp
```diff
@@ -37,6 +39,41 @@ constexpr const static int MIN_ZOOM_FOR_TURNS = 15;
...
+                                                    "(noturn)", // internal types only below here
+                                                    "(suppressed)",
+                                                    "(enter roundabout at exit)",
+                                                    "(exit roundabout)",
+                                                    "(enter rotary at exit)",
+                                                    "(exit rotary)",
+                                                    "(enter roundabout intersection at exit)",
+                                                    "(exit roundabout intersection)",
+                                                    "(stay on roundabout)",
+                                                    "(sliproad)"};
```
daniel-j-h: Can we move this to instructionTypeToString

src/engine/plugins/tile.cpp
```diff
@@ -731,21 +791,39 @@ void encodeVectorTile(const DataFacadeBase &facade,
...
             // we need to pre-encode all values here because we need the full offsets later
             // for encoding the actual features.
-            std::vector<std::tuple<util::Coordinate, unsigned, unsigned, unsigned, unsigned>>
+            std::vector<std::tuple<util::Coordinate,
+                                   unsigned,
+                                   unsigned,
+                                   unsigned,
+                                   unsigned,
+                                   unsigned,
+                                   unsigned>>
```
daniel-j-h: This is getting a bit unwieldy - how can we refactor this?

daniel-j-h:
Tile tests need adaption.

PR #4466, opened 2017-08-31
Preserve edge weights when blank rate provided in traffic data files

danpat:
# Issue

This change makes it possible to preserve existing `weight` values on edges, while updating the `duration`.

If traffic data CSV files include a blank 4th column:

```
nodeA,nodeB,speed        <- (1) weight will be calculated from speed
nodeA,nodeB,speed,rate   <- (2) weight will be calculated from rate
nodeA,nodeB,speed,       <- (3) weight will be preserved
```

Currently, we support (1) and (2).  This change adds (3), which makes it easy to update the durations on edges without affecting the routing weights.  This is useful, for example, if you're using a distance-based profile, but want to update speed values to track accurate ETAs.

## Tasklist
 - [x] review
 - [x] adjust for comments

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,9 @@
       - New property for profile table: `excludable` that can be used to configure which classes are excludable at query time.
       - New optional property for profile table: `classes` that allows you to specify which classes you expect to be used.
         We recommend this for better error messages around classes, otherwise the possible class names are infered automatically.
+    - Traffic:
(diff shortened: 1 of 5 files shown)


danpat:
@oxidase I thought about that, but...we currently support ignoring fields after the 3rd comma - I feel like this would be a breaking change.

Probably worth removing support in OSRM v6, but for now, just in case someone is using it, I think it should remain.

PR #4473, opened 2017-09-04
Uses parsed len(turn:lanes) to fixup number of lanes, resolves #4472

### LINKED ISSUE #4472: Use len(turn:lanes) for number of lanes in road classification
Our profiles [parse](https://github.com/Project-OSRM/osrm-backend/blob/0fc1aa271188699b996fbb3221c808f3e731b941/profiles/lib/guidance.lua#L81-L107) the `lanes=number` tag but do not use `turn:lanes` to count number of lanes. If a way has `turn:lanes` but not `lanes` we do not set `RoadClassification::num_lanes`.

We should parse `turn:lanes` (and forward, backward, namespaced tags), count the `|` delimiters and set `num_lanes` accordingly.

Affects:
- coordinate extraction in roundabout handler
- traffic island detection and road width estimation in mergeable road detector
- bearing calculation in intersection generation

daniel-j-h:
For https://github.com/Project-OSRM/osrm-backend/issues/4472: when there is a `turn:lanes=` tag present but not a `lanes=n` tag we failed to just use `len(turn:lanes)` for our internal number of lanes calculations.


third_party/libosmium/test/t/io/data-cr.opl
```diff
@@ -1 +1 @@
-n1 v1 dV c1 t2014-01-01T00:00:00Z i1 utest T x1.02 y1.02
+n1 v1 dV c1 t2014-01-01T00:00:00Z i1 utest T x1.02 y1.02
```
MoKob: @daniel-j-h is this an unrelated change?

oxidase: @mokob it is fixed already in upstream master.

Issue #4434, opened 2017-08-23
Derive refs from route relations


1ec5:
Many `ref`s for RouteSteps in the U.S. and Canada should pair the route number (ref) with a cardinal direction, where available. As detailed in Project-OSRM/osrm-text-instructions#119, the cardinal direction is [specified](https://wiki.openstreetmap.org/wiki/Highway_Directions_In_The_United_States) either in the `direction` tag of a route relation that contains the way or in the way’s `role` in the route relation.

Beyond simply appending a cardinal direction to the existing `ref` property, this work gives us an opportunity to implement relation parsing (#482), as well as a more structured response format for `ref` and `destination` (#3329) that enables clients to robustly choose highway shield icons (mapbox/mapbox-navigation-ios#334).

For the Mapbox Navigation SDKs, what we want to display is `US 101 North` (or <img src="https://upload.wikimedia.org/wikipedia/commons/thumb/7/79/US_101.svg/64px-US_101.svg.png" width="24" alt="US 101"> North). The way is tagged `ref=US 101`, while the relation separates this into [`network=US:US`](https://wiki.openstreetmap.org/wiki/Key:network#Highway_routes) `ref=101` `direction=north` (or `north` `role`s). [The mapping between `network` and way `ref` prefixes](https://wiki.openstreetmap.org/wiki/United_States_roads_tagging/Routes) is not especially straightforward:

* way `ref=I 75` = relation `network=US:I` `ref=75`
* way `ref=US 101` = relation `network=US:US` `ref=101`
* way `ref=US 101 Alternate` = relation `network=US:US:Alternate` `ref=101`
* way `ref=SR 4` = relation `network=US:OH` `ref=4` (for example)

For the `ref` and `destination` properties on the route step, we need to either choose between way `ref`s and relation memberships or figure out some way to match relation memberships up with way `ref`s.

In the U.S., at least, we’re unlikely to lose much information for notable routes by cutting over to route relations and ignoring way `ref`s entirely. The unambiguous network value would be a boon for enabling clients to choose route shields (mapbox/mapbox-navigation-ios#334), but it would be less suitable for voice or textual display. For that, either the server or the client would need to map the unambiguous `network` values like `US:I` to human-readable prefixes like `I`. We could build up a lookup table pretty easily. Another challenge is that way maintains an unordered list of relation memberships.

[...]

PR #4478, opened 2017-09-05
Process route relation in car.lua

deniskoronchik:
# Issue

## Tasklist
 - [ ] review
 - [ ] adjust for comments

## Improvements
- [ ] Support more separators in common tokenizer
- [ ] Add more scores for 65M refs (like for a numbers)
- [x] Improve code: return ref match result in specified order from match funciton (replace two containers, that used right now)

## Requirements / Relations
Based on PR: #4438 (merge after)

Support multiref's: https://github.com/Project-OSRM/osrm-text-instructions/issues/119#issuecomment-328053084

## Decsription
It appends route direction value to ref with prefix `$`. 
As example `ref` = `I75 $north`
So `$north`, `$south`, `$west`, `$east` can be processed by UI in useful way

## Problems
Also there is a very strange bug for way http://www.openstreetmap.org/way/19026367
I'm pretty sure, that return correct ref for it in LUA:
```sh
SR 16; SR 83  # source ref
west                # role in one relation
south               # role in second relation
SR 83	south  # matched ref to role
SR 16	west   # matched ref to role
SR 83 $south; SR 16 $west  # final result.ref
```

But in JSON it took value: `SR 83 $west; SR 16 $east`

**Update**: problem wasn't reproduced. It looks like confusion in way id's in LUA and osrm-frontend route during test.

### DIFF
--- a/profiles/car.lua
+++ b/profiles/car.lua
@@ -396,10 +396,14 @@ function process_way(profile, way, result, relations)
   -- now process relations data
   local matched_refs = nil;
   if result.ref then
-    matched_refs = Relations.MatchToRef(relations, result.ref)
+    local match_res = Relations.match_to_ref(relations, result.ref)
+
+    order = match_res['order']
+    matched_refs = match_res['match']
 
     local ref = ''
-    for k, v in pairs(matched_refs) do
+    for _, k in pairs(order) do
+      local v = matched_refs[k]
       if ref ~= '' then
         ref = ref .. '; '
       end
--- a/profiles/car.lua
+++ b/profiles/car.lua
@@ -8,6 +8,7 @@ Handlers = require("lib/way_handlers")
 Relations = require("lib/relations")
 find_access_tag = require("lib/access").find_access_tag
 limit = require("lib/maxspeed").limit
+Utils = require("lib/utils")
 
 function setup()
   local use_left_hand_driving = false
@@ -393,21 +394,54 @@ function process_way(profile, way, result, relations)
   WayHandlers.run(profile,way,result,data,handlers)
 
   -- now process relations data
-  local data = Relations.Merge(relations)
-  local direction = data["route_direction"]
-  if direction then
-    if result.ref then
-      if result.ref:len() > 0 then
-        result.ref = result.ref .. ' ' .. direction
+  local matched_refs = nil;
+  if result.ref then
+    matched_refs = Relations.MatchToRef(relations, result.ref)
+
+    local ref = ''
+    for k ,v in pairs(matched_refs) do
+      if ref ~= '' then
+        ref = ref .. '; '
+      end
+
+      if v then
+        ref = ref .. k .. ' $' .. v
       else
-        result.ref = direction
+        print(ref)
+        ref = ref .. k
       end
     end
+
+    result.ref = ref
+    -- count = 0
+    -- for k, v in pairs(matched_refs) do
+    --   count = count + 1
+    -- end
+    -- if count > 1 then
+    --   print('---', way:id())
+    --   for k, v in pairs(matched_refs) do
+    --     print(k, v)
+    --   end
+    -- end
   end
+
+
 end
 
 function process_relation(profile, relation, result)
   local t = relation:get_value_by_key("type")
+
+  function add_extra_data(m)
+    local name = relation:get_value_by_key("name")
+    if name then
+      result[m]["route_name"] = name
+    end
+
+    local ref = relation:get_value_by_key("ref")
+    if ref then
+      result[m]["route_ref"] = ref
+    end
+  end
 
   if t == "route" then
     local route = relation:get_value_by_key("route")
@@ -417,6 +451,7 @@ function process_relation(profile, relation, result)
         local role = string.lower(m:role())
         if role == "north" or role == "south" or role == "west" or role == "east" then
           result[m]["route_direction"] = role
+          add_extra_data(m)
         end
       end
     end
@@ -428,6 +463,7 @@ function process_relation(profile, relation, result)
         for _, m in ipairs(relation:members()) do
           if m:role() == "forward" then
             result[m]["route_direction"] = direction
+            add_extra_data(m)
           end
         end
       end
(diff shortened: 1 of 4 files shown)


profiles/lib/relations.lua
```diff
@@ -17,4 +19,82 @@ function Relations.Merge(relations)
   return result
 end
 
+-- match ref values to relations data
+function Relations.MatchToRef(relations, ref)
```
emiltin: usually we use match_to_ref() instead of CamelCase for lua functions.

deniskoronchik: Thanks, will fix it soon

profiles/lib/utils.lua
```diff
@@ -0,0 +1,43 @@
...
+
+  return result
+end
+
+-- same as Utils.StringListTokens, but with many possible separators:
+-- ',' | ';' | ' '| '(' | ')'
+function Utils.tokenize_common(str)
+  result = {}
+  local idx = 0
+  for s in str.gmatch(str, "%S+") do
```
1ec5: If I’m not mistaken, this code splits a string by whitespace; it doesn’t seem to split by other separators like `;` or `(`. Note that an item in `ref` may contain parentheses, for example [`ref=A1(M)` on a way](http://www.openstreetmap.org/way/280766464) or [on a relation](http://www.openstreetmap.org/relation/103623).

profiles/lib/relations.lua
```diff
@@ -0,0 +1,100 @@
...
+  function calculate_scores(refs, tag_value)
+    local tag_tokens = Set(Utils.tokenize_common(tag_value))
+    local result = {}
+    for i, r in ipairs(refs) do
+      local ref_tokens = Utils.tokenize_common(r)
+      local score = 0
+
+      for _, t in ipairs(ref_tokens) do
+        if tag_tokens[t] then
+          if Utils.is_number(t) then
```
1ec5: It’s common for routes to have alphabetic suffixes, for example [`ref=I 35W` on a way](http://www.openstreetmap.org/way/50715899) or [`ref=35W`](http://www.openstreetmap.org/relation/2145877) on a (northbound) route relation.

profiles/lib/relations.lua
```diff
@@ -0,0 +1,100 @@
...
+            score = score + 2
+          else
+            score = score + 1
```
1ec5: 👍

In Project-OSRM/osrm-backend#4434, I proposed trying to match a prefix in the way’s `ref` to a component of the relation‘s `network` as an additional score to factor in. Via [this Overpass query](http://overpass-turbo.eu/s/rHB), I found [one way](http://www.openstreetmap.org/way/122502126) that’s simultaneously part of `PA 88 South` and `PA 88 Truck North`.

The state-by-state queries in https://github.com/Project-OSRM/osrm-text-instructions/issues/119#issuecomment-328053084 can only find relations tagged with directions in roles; they’d need to be reworked to handle `direction` tags on relations. But given that I’ve only found one problematic way in the U.S. so far, matching `networks` is probably overkill, even if it’d be more robust.

profiles/lib/utils.lua
deniskoronchik: Yes, you are correct. This is a first version, need to figure out all cases and change this match string

profiles/lib/relations.lua
deniskoronchik: Have you any idea how to support it?

profiles/lib/utils.lua
```diff
@@ -0,0 +1,43 @@
...
+
+-- split string 'a; b; c' to table with values ['a', 'b', 'c']
+-- so it use just one separator ';'
+function Utils.string_list_tokens(str)
+  result = {}
+  local idx = 0
+  for s in str.gmatch(str, "([^;]*)") do
+    if s ~= nil and s ~= '' then
+      idx = idx + 1
+      table.insert(result, idx, (s:gsub("^%s*(.-)%s*$", "%1")))
```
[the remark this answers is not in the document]
deniskoronchik: Does it will save order of tokens in that case? It's critical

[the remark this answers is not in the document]
1ec5: Yes, I’m pretty sure it would have the same behavior as this code, since `idx` always matches `#result`.

profiles/car.lua
```diff
@@ -390,6 +392,78 @@ function process_way(profile, way, result)
...
+
+  function add_extra_data(m)
+    local name = relation:get_value_by_key("name")
+    if name then
+      result[m]["route_name"] = name
+    end
+
+    local ref = relation:get_value_by_key("ref")
+    if ref then
+      result[m]["route_ref"] = ref
```
1ec5: [`route_ref`](https://taginfo.openstreetmap.org/keys/route_ref) happens to be quite common in OpenStreetMap already. Is it safe to overload this tag? If not, perhaps we could prefix it, like `osrm:route_ref`.

profiles/lib/relations.lua
1ec5: How about setting `osrm:routes` to a nested table? For example, [this way](http://www.openstreetmap.org/way/122502126) would have the equivalent of:

```lua
result[m]["osrm:routes"] = {
  {
    network = "US:PA",
    ref = "88",
    direction = "South",
  },
  {
    network = "US:PA:Truck",
    ref = "88",
    direction = "North",
  },
}
```

profiles/car.lua
deniskoronchik: This code just add data with key `reoute_ref` into relation data, that would be passed into process_way function. So it can by any we want, because it doesn't intersect or merge with OSM tags. This are two separated structures (tags, and relations data)

profiles/lib/relations.lua
deniskoronchik: Can't find any example, where current algorithm will work incorrect in this case.

1ec5: After rereading this code, I think the suffix case will be a nonissue. However, for [this way](http://www.openstreetmap.org/way/122502126), wouldn’t the two items in the way `ref` (`PA 88` and `PA 88 Truck`) tie because `88` appears in both items and both relations’ `ref` tags are set to `88`? We could factor in the relations’ `network` tags, sort of like what’s described in https://github.com/Project-OSRM/osrm-backend/pull/4478#discussion_r138947028.

danpat:
After talking with @deniskoronchik, closing this in favour of https://github.com/Project-OSRM/osrm-backend/pull/4524 so that we can do a beta release tag on this branch before merging to `master`.

PR #4481, opened 2017-09-06
Update MLD compatibility file checks

karenzshea:
# Issue

This adds some early aborting on missing MLD files. I added the files to check for based on what is loaded into storage [here](https://github.com/Project-OSRM/osrm-backend/blob/c2dc7e9cd0052e5a5dc7fc4b5de8b858c8e8f571/src/storage/storage.cpp#L479-L556). While we the loading of these files is conditional, if they don't exist the memory blocks are all initialized to 0 (https://github.com/Project-OSRM/osrm-backend/blob/c2dc7e9cd0052e5a5dc7fc4b5de8b858c8e8f571/src/storage/storage.cpp#L557-L575) and anything that tries to access them later will crash.

## Tasklist
 - [ ] review
 - [ ] adjust for comments

## Requirements / Relations
5.12?


daniel-j-h:
What's the state of this? @karenzshea can you rebase on master; also Travis is currently failing. @danpat can you review and merge then?

karenzshea:
On very small graphs where all the nodes fit into one cell, two of the MLD memory blocks may be empty. No need to check if those areas are empty, as long as the cells block isn't empty. 

Once the tests are green this should be ready to merge.

Issue #4465, opened 2017-08-31
Snapping to intersections incurs random turn penalty costs


danpat:
Given a scenario like this:

![screen shot 2017-08-31 at 10 28 35 am](https://user-images.githubusercontent.com/1892250/29936798-57e85fce-8e38-11e7-8358-5435859d13c1.png)

the "nearest" point is the end of one of those three lines.  The distances are all equal, so the line we snap to depends on the search order in the RTree - it's essentially random from the user's perspective.

When we calculate the outgoing edge weights from our phantom nodes, we subtract the sum of the edge segments leading up to the snapped point from the total edge weight.  The problem is that this will leave the turn penalty value as a remainder, even if we snap to the end of the line.

We should special-case this behaviour - if you snap to the end of a line (i.e. an intersection), we should set the outgoing edge weights to 0 to all directly connected outgoing edges.

The problematic code is here:

https://github.com/Project-OSRM/osrm-backend/blob/master/include/engine/geospatial_query.hpp#L458-L464

We calculate the % snapped along the line here:

https://github.com/Project-OSRM/osrm-backend/blob/master/include/engine/geospatial_query.hpp#L479

If that value is 1.0, then the `forward_weight` should be 100% of the edge weight, not just the sum of the proceeding segment weights.

I think it's OK logic that if you snap at 99.99% (e.g 1m back from the intersection), then you pay the full turn cost.

danpat:
Heads up to anyone that tackles this: this will change a lot of our test cases - be prepared to verify/modify many many tests that currently depend on RTree snapping order.

danpat:
A related problem to this: adding waypoints.

If you have a route, and you add a waypoint exactly at an intersection, we will randomly snap to one of the approach roads to the intersection.  The snapped road may not be on the original route, and this will force a potentially large re-routing.  In reality, the waypoint doesn't require any detour, as it's basically exactly on the original route.

MoKob:
@danpat is this related/the same as https://github.com/Project-OSRM/osrm-backend/issues/2287? At first glance, I would say it to be the same problem

mjjbell:
This is now fixed by #5953

PR #4484, opened 2017-09-07
Adds cucumber scenario for u turn restrictions at intersection

daniel-j-h:
@srividyacb here's a cucumber scenario for what we talked about. See how the route does not make a u-turn at `x` but instead goes to the `b` node and makes a u-turn there.


MoKob:
Currently this is blogged by https://github.com/Project-OSRM/osrm-backend/issues/4465

features/car/restrictions.feature
```diff
@@ -115,6 +115,36 @@ Feature: Car - Turn restrictions
...
+        And the relations
+            | type        | way:from | way:to | node:via | restriction |
+            | restriction | ax       | ax     | x        | no_u_turn   |
+            | restriction | bx       | bx     | x        | no_u_turn   |
+            | restriction | cx       | cx     | x        | no_u_turn   |
+            | restriction | dx       | dx     | x        | no_u_turn   |
+
+        When I route I should get
+            | waypoints | route             |
+            | a,x,a     | ax,ax,xb,xb,ax,ax |
```
TheMarex: We could use bearing values here to work around the limitation of snapping to `x` and trigger a real u-turn.

features/car/restrictions.feature
```diff
@@ -115,6 +115,36 @@ Feature: Car - Turn restrictions
...
+            | xd    |
+
+        And the relations
+            | type        | way:from | way:to | node:via | restriction |
+            | restriction | ax       | ax     | x        | no_u_turn   |
+            | restriction | bx       | bx     | x        | no_u_turn   |
+            | restriction | cx       | cx     | x        | no_u_turn   |
+            | restriction | dx       | dx     | x        | no_u_turn   |
+
+        When I route I should get
```
SiarheiFedartsou, line 144: @mjjbell as you worked on PR closing https://github.com/Project-OSRM/osrm-backend/issues/4465 and this PR was blocked by it could you take a look? This test currently failing on master:
```
2022-10-06T14:33:17.0903046Z    Message:
2022-10-06T14:33:17.0903377Z      [31mTables were not identical:
2022-10-06T14:33:17.0903835Z      | waypoints |     route             |
2022-10-06T14:33:17.0904218Z      | a,x,a     | (-) ax,ax,xb,xb,ax,ax |
2022-10-06T14:33:17.0904606Z      | a,x,a     | (+) ax,xb,xb,xb,ax,ax |[39m
```

But it seems in this context `ax,ax,xb,xb,ax,ax` is the same as `ax,xb,xb,xb,ax,ax`, or not? Asking because tbh I don't have enough knowledge yet to now for sure if we can merge it or not(but probably would be great to merge it if it is ready) :)

mjjbell: Short answer: yes

Longer answer:
Adding the `turns` to see what's going on.

```
| a,x,a     | (-) ax,ax,xb,xb,ax,ax | (-) |
| a,x,a     | (+) ax,xb,xb,xb,ax,ax | (+) depart,new name straight,continue uturn,arrive,depart,arrive |
```

So the response is splitting the route into two as  `a->x->b(uturn)->x , x->a`
whereas the test wants `a->x , x->b(uturn)->x->a`

They will have equivalent weight, so it's just whichever OSRM happens to select out of the two.

This wouldn't have worked in the past because the mid journey stop might have been on `cx` or somewhere else that wasn't optimal.

SiarheiFedartsou (approved):

PR #4488, opened 2017-09-08
Add to car profile missing DK:rural maxspeed_table exceptions osm-fr/osmose-backend#174

frodrigo:
From wiki https://wiki.openstreetmap.org/wiki/Speed_limits
The DK:rural maxspeed_table exceptions is missing


danpat:
Good catch, thanks!

Issue #4348, opened 2017-07-27
Review slip-road handlers


MoKob:
## Scenario 1

`primary` via `primary_link` onto secondary.

![screen shot 2017-07-27 at 12 48 34 2](https://user-images.githubusercontent.com/12932279/28667018-f61b34da-72c9-11e7-81af-314f95c9d4ed.png)
[osrm](http://map.project-osrm.org/?z=18&center=38.778487%2C-76.084764&loc=38.778978%2C-76.084764&loc=38.778527%2C-76.087232&hl=en&alt=0)

This feels like a _primary_ candidate for a sliproad. 

 - [x] investigate handler that discards it
 - [x] adjust if possible
 - [x] evaluate other affected locations

## Scenario 2

`tertiary` via `tertiary_link` onto second residential. 

![screen shot 2017-07-27 at 12 51 59 2](https://user-images.githubusercontent.com/12932279/28667141-7f5f4fce-72ca-11e7-80ca-258f2b7895b2.png)

[osrm](http://map.project-osrm.org/?z=18&center=37.803280%2C-122.437145&loc=37.803599%2C-122.436983&loc=37.803075%2C-122.437386&hl=en&alt=0)

Looks very similar to scenario 3, only down two road categories. Actions remain the same:

 - [x] investigate handler that discards it
 - [x] adjust if possible
 - [x] evaluate other affected locations

## Scenario 3

`residential_road` via `residential_road` onto `residential_road`

![screen shot 2017-07-27 at 12 55 54 2](https://user-images.githubusercontent.com/12932279/28667255-1714adf0-72cb-11e7-8608-6b0315007e36.png)

[osrm](http://map.project-osrm.org/?z=18&center=37.804175%2C-122.439741&loc=37.804325%2C-122.439778&loc=37.803994%2C-122.439923&hl=en&alt=0)

This one is tricky, due to no `link` class. Given the matching names and the short segment, we might be able to detect these.

 - [x] investigate handler that discards it
 - [x] adjust if possible
 - [x] evaluate other affected locations

## Scenario 4

`motorway_link` via `motorway_link` onto `primary`

![screen shot 2017-07-27 at 12 58 30 2](https://user-images.githubusercontent.com/12932279/28667494-092d24f0-72cc-11e7-9811-898db1fcfe6f.png)

[...]

PR #4493, opened 2017-09-11
Fix guidance sliproads scenario 3

oxidase:
# Issue

PR fixes scenario 3 of #4348
screenshot from 2017-09-11 20-00-06

by adding [an additional check](https://github.com/Project-OSRM/osrm-backend/blob/de932ea45f2af79ac24726e2191514d98abb8832/src/extractor/guidance/sliproad_handler.cpp#L496-L503) for an intermediate node 

Also PR changes allowed angle from 115 to 140 degrees for non-link sliproad check that updates scenario 2 of #4348

## Tasklist
 - [x] review
 - [x] adjust for comments

## Requirements / Relations

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@
       - MLD uses a unidirectional Dijkstra for 1-to-N and N-to-1 matrices
     - Guidance
       - Fixed some cases of sliproads pre-processing (https://github.com/Project-OSRM/osrm-backend/issues/4348)
+      - don't suppress name announcements via sliproad handler
 
 # 5.12.0
   - Changes from 5.11:
--- a/features/guidance/dedicated-turn-roads.feature
+++ b/features/guidance/dedicated-turn-roads.feature
@@ -1049,5 +1049,5 @@ Feature: Slipways and Dedicated Turn Lanes
 
 
        When I route I should get
-            | waypoints | route              | turns                                      | locations |
-            | s,f       | sab,ae,dbcef,dbcef | depart,turn right,turn slight right,arrive | s,a,e,f   |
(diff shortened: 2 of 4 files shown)


daniel-j-h:
Travis fails: `dk:rural` seems not to be in `taginfo.json`.

I don't know why we didn't see this before.

oxidase:
@daniel-j-h i have created [a map](https://api.mapbox.com/styles/v1/oxidase/cj7i0gg7x4y4f2rqf5n023hyz.html?title=true&access_token=pk.eyJ1Ijoib3hpZGFzZSIsImEiOiJjaXcxNWMxYmswMDRlMnlxOWtlem0yamU3In0.PjtHAtgQDzxhZ1BE484Sqw#17.3/48.176267/11.558609/0) where 
* read nodes are no more sliproads because of a new double announcements check (cases like http://map.project-osrm.org/?z=18&center=39.509808%2C-121.539978&loc=39.510303%2C-121.539626&loc=39.509653%2C-121.540584&hl=en&alt=0 or http://map.project-osrm.org/?z=18&center=32.689455%2C-117.171555&loc=32.689295%2C-117.170949&loc=32.690094%2C-117.171421&hl=en&alt=0)

* blue nodes are new sliproads because of the change angle from 115 to 140 degrees

* green nodes are new sliproads with an intersection in a candidate road

As i see blue nodes is really big change, so i think to remove https://github.com/Project-OSRM/osrm-backend/blob/ad9ef924f403bf842f1d916ee324e9045f6e20ce/src/extractor/guidance/sliproad_handler.cpp#L436 condition but it will require adding `_link` in OSM sliproad ways to keep some "green" sliproad nodes.

PR #4494, opened 2017-09-11
Replace GCC/clang option -std=C++1y with -std=C++14

mloskot:
# Issue

C++1y indicates features from before final C+14 publication.
C++14 contains several further changes.

OSRM declares it's written in C++14. 

## Tasklist
 - [x] review

------

Looks like overlooked update.


oxidase:
Failing travis job will be fixed by 80defb7b1

PR #4426, opened 2017-08-21
Restructure Obvious Turn Handling

MoKob:
# Issue

The decision for whether a turn can be considered obvious or not can have a lot of influence on the quality of the guidance decisions.
This PR refactors / reworks the decision process to make it both easier to follow along and less error prone.

Also resolves https://github.com/Project-OSRM/osrm-backend/issues/3105

## State of the PR / What needs to be done

Currently, we produce logging output with geojson, indicating which roads change in terms of their obvious classification.
Sadly, our test-suite is mainly build on `primary` roads. This results in a lot of unwanted classifications all over the map.
While the test-suite is passing right now, we need to ensure the classification gets tuned correctly. This requires an iterative process of adjusting classificaitons, creating a new map (comparison/geojson output is currently included) and reviewing locations in which obvious turn classification changes.

For every invalid change, a new test has to be added to cucumber and the classification has to be tuned to handle it and all other tests correctly.

## Further cases to investigate / build tests / solve

 - [ ] diagonal links on segregated intersections
    - OSM link: https://www.openstreetmap.org/#map=19/37.61261/-122.40385
screen shot 2017-10-13 at 14 04 19

 - [ ] roads splitting up (sometimes prior to intersections)
    - OSM link: https://www.openstreetmap.org/#map=19/37.63171/-122.46205 x
    - OSM link: https://www.openstreetmap.org/#map=19/37.76407/-122.49642 x
    - OSM link: https://www.openstreetmap.org/#map=19/37.74682/-122.44415
    - OSM link: https://www.openstreetmap.org/#map=19/37.77072/-122.41727 x
    - OSM link: https://www.openstreetmap.org/#map=19/37.26591/-121.84474 x
    - OSM link: https://www.openstreetmap.org/#map=19/37.85108/-122.27078
screen shot 2017-10-13 at 14 05 45
screen shot 2017-10-13 at 14 06 39
screen shot 2017-10-13 at 14 08 31
screen shot 2017-10-13 at 14 11 58
screen shot 2017-10-13 at 14 19 10
screen shot 2017-10-13 at 14 20 23

 - [ ] turn roads
    - OSM link https://www.openstreetmap.org/#map=19/37.76471/-122.49639
screen shot 2017-10-13 at 14 07 18

 - [ ] continuing roads
    - OSM link https://www.openstreetmap.org/#map=19/37.63552/-122.48401
    - OSM link https://www.openstreetmap.org/#map=19/52.47752/13.28864
[...]

PR #4499, opened 2017-09-13
Don't suppress fork instructions

oxidase:
# Issue

In cases of links like [osrm](http://map.project-osrm.org/?z=18&center=37.587649%2C-77.473070&loc=37.588977%2C-77.470720&loc=37.588157%2C-77.473134&hl=en&alt=0) forks detection uses condition of "no obvious direction" https://github.com/Project-OSRM/osrm-backend/blob/4ea3f33376eb9dc3da069b295b7a9c2688a940ad/src/extractor/guidance/turn_handler.cpp#L661-L663 and https://github.com/Project-OSRM/osrm-backend/blob/4ea3f33376eb9dc3da069b295b7a9c2688a940ad/src/extractor/guidance/turn_handler.cpp#L677

But forks have other obvious direction detection as turns. The condition
https://github.com/Project-OSRM/osrm-backend/blob/4ea3f33376eb9dc3da069b295b7a9c2688a940ad/src/extractor/guidance/turn_handler.cpp#L242 uses `obvious_index` from `findObviousTurn`, so if there is no no obvious direction detected by the fork handler, but `findObviousTurn` returns non-obvious direction (obvious_index != 0 ) the fork will be handled as a turn. This behavior is used for non-links highway types, but potentially wrong for links. 

The PR adds a fallback to non-obvious conditional branch for forks links.
So the above route will be
screenshot from 2017-09-13 08-33-54

Map of fork intersections where instructions will be handled by `assignFork`
https://api.mapbox.com/styles/v1/oxidase/cj7ip8d1i5kzz2rq53gd97k7y.html?title=true&access_token=pk.eyJ1Ijoib3hpZGFzZSIsImEiOiJjaXcxNWMxYmswMDRlMnlxOWtlem0yamU3In0.PjtHAtgQDzxhZ1BE484Sqw#17.36/37.41387/-121.8803

/cc @daniel-j-h 

## Tasklist
 - [ ] review
 - [ ] adjust for comments

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,8 @@
       - New function to support relations: `process_relation`. Read more in profiles documentation.
     - Infrastructure:
       - Lua 5.1 support is removed due to lack of support in sol2 https://github.com/ThePhD/sol2/issues/302
+    - Guidance
+      - treat forks from link-typed roads as forks to avoid turn-like instructions suppression
 
 # 5.12.0
     - Guidance
--- a/features/guidance/fork.feature
+++ b/features/guidance/fork.feature
@@ -385,3 +385,21 @@ Feature: Fork Instructions
             | waypoints | route      | turns                    |
             | a,j       | on,xbcj    | depart,arrive            |
             | a,i       | on,off,off | depart,turn right,arrive |
+
+     # https://www.openstreetmap.org/node/619615462
+     Scenario: Fork on links - don't suppress as obvious turns
+        Given the node map
+            """
+            a . . . . . b . . . . . c
+                          `  .   .  d
+            """
+
+        And the ways
+            | nodes | highway       |
+            | abc   | motorway_link |
+            | bd    | motorway_link |
+
+        When I route I should get
+            | waypoints | route     | turns                           |
+            | a,c       | abc,abc   | depart,arrive                   |
+            | a,d       | abc,bd,bd | depart,turn slight right,arrive |
--- a/features/guidance/fork.feature
+++ b/features/guidance/fork.feature
@@ -400,6 +400,6 @@ Feature: Fork Instructions
             | bd    | motorway_link |
 
         When I route I should get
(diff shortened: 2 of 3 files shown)


daniel-j-h:
Hm can you check if the behavior changes with the WIP PR https://github.com/Project-OSRM/osrm-backend/pull/4426. I think you're right we should definitely emit a step here.

oxidase:
@daniel-j-h i have rebased #4426 to master and checked e91124c05ed79510e7f9908e8035c93be572a9e0

The PR makes changes in the second expectation, but not the first one
```
     Tables were not identical:
     | waypoints | route     |     turns                           |
     | a,c       | abc,abc   |     depart,arrive                   |
     | a,d       | abc,bd,bd | (-) depart,turn slight right,arrive |
     | a,d       | abc,bd,bd | (+) depart,turn straight,arrive     |
```

IMHO, it is better to treat forks from links as forks via `assignFork`, otherwise lane information and possibly destinations/refs will be suppressed.

MoKob:
@oxidase did you do a review on where this behaviour triggers? I can imagine that many situations exist in which people would prefer to see no instruction (e.g. two way link with another link splitting) I would love to avoid oscillating between announcements/non announcements here.

oxidase:
@MoKob here is a map with new non-obvious forks https://api.mapbox.com/styles/v1/oxidase/cj7ip8d1i5kzz2rq53gd97k7y.html?title=true&access_token=pk.eyJ1Ijoib3hpZGFzZSIsImEiOiJjaXcxNWMxYmswMDRlMnlxOWtlem0yamU3In0.PjtHAtgQDzxhZ1BE484Sqw#9.21/37.7352/-122.1942

I understand your point, but here we need some objective quality measure of "guidance annoyance" vs "missing instruction" penalties in finding obvious directions, otherwise it will be subjective.

oxidase:
Closing here: PR does not solve the original issue

PR #4516, opened 2017-09-18
ability to force profile when running cuke tests

emiltin:
Version 2 of the profile API makes it easier to build a variation of a profile based on an existing one that you require. For example, you can create a profile for cargo bikes, based on the standard bicycle profile.

In situations like this, it's helpful to be able to run existing test with your new profile.

This PR lets you force the profile used when running cucumber tests by setting the OSRM_PROFILE environment variable, e.g:

```
OSRM_PROFILE=cargo_bike node_modules/cucumber/bin/cucumber.js features/bicycle
```
An annoyance is that the caching mechanism means that changing the OSRM_PROFILE env has no effect unless one of the .feature files are changed as well. I'm a bit unsure how the caching mechanism detects changes in the profile name, and thus how to fix this.

### DIFF
--- a/features/support/env.js
+++ b/features/support/env.js
@@ -44,6 +44,8 @@ module.exports = function () {
 
         this.OSRM_PORT = process.env.OSRM_PORT && parseInt(process.env.OSRM_PORT) || 5000;
         this.HOST = 'http://127.0.0.1:' + this.OSRM_PORT;
+        
+        this.OSRM_PROFILE = process.env.PROFILE
 
         if (this.PLATFORM_WINDOWS) {
             this.TERMSIGNAL = 9;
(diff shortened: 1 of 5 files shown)


features/support/env.js
```diff
@@ -44,6 +44,8 @@ module.exports = function () {
 
         this.OSRM_PORT = process.env.OSRM_PORT && parseInt(process.env.OSRM_PORT) || 5000;
         this.HOST = 'http://127.0.0.1:' + this.OSRM_PORT;
+        
+        this.OSRM_PROFILE = process.env.PROFILE
```
oxidase: should be here `process.env.OSRM_PROFILE`?

emiltin: yes, sorry for the copy paste error

emiltin:
of course most tests are tailored to a specific profile. so if you run the car tests with the bicycle profile many of them will fail. but for situations where you develop variations of profiles it's very useful.

emiltin:
when hashing the feature file, the value of OSRM_PROFILE is now included in the hash computation

emiltin:
the use case i have is storing our profiles in a separate repository from osrm-backend.
for this use case would be useful to allow absolute paths, however it seems a bit tricky due to LUAs way of handling require()

Issue #3329, opened 2016-11-16
Handle names/refs/destinations/destination as Arrays


freenerd:
Moving this discussion out of https://github.com/Project-OSRM/osrm-backend/issues/3304

Right now we are outputting several values for a field in the same way as OSRM, semi-colon separated. It makes sense for downstream users to split these into an array to work better with it. In fact [osrm-text-instructions](https://github.com/Project-OSRM/osrm-text-instructions/blob/46bc8520c019629f48012266b5c1c1fcb15ce388/index.js#L156) and [MapboxDirections.swift](https://github.com/mapbox/MapboxDirections.swift/blob/a241f3a09b236c61c27ffc253fccc619ec9a7f98/MapboxDirections/MBRouteStep.swift#L443) are doing this. We should discuss if this would make sense to move to OSRM right away.

This could happen on during the output serialization only (low effort), but might also make sense to do earlier to aid the `new name` detection (higher effort)

TheMarex:
> This could happen on during the output serialization only (low effort), but might also make sense to do earlier to aid the new name detection (higher effort)

Taking the higher effort route makes sense from a architectural point of view, because we want to encode all data-source specific string-processing in the profiles. What this would mean explicitly is:

1. This would mean an API change in the profiles:
    - `result.destination` (string) -> `result.destinations` (array of strings)
    - `result.destination_ref` (string) -> `result.destination_refs` (array of strings)
    - `result.ref` (string) -> `result.refs` (array of strings)
2. Instead of storing a string `Chain Bridge;McLean` we would store the strings `Chain Brigde` and `McLean`. That means every edge will have multiple string ids associated with it.
3. We need another layer of indirection that translates an id to a list of string ids.  That means another `RangeTable string_ranges` that stores the range of indices into `m_name_table`.

freenerd:
@TheMarex likely the same should happen for `result.name` (string) -> `result.names` (array of strings) then?

TheMarex:
@freenerd `name` is a bit special since we typically only ever have one name per edge (as per OSM wiki).

I'm not sure if including multiple names (e.g. translated ones) would work every well with the array format, since it would be hard to distinguish which one is the localized and which one the translated name.

1ec5:
Per #3329, `exits` should be an array too, because a single shared exit, such as the one in https://github.com/Project-OSRM/osrm-backend/issues/4222#issuecomment-313910706, may have multiple exit numbers.

[...]

PR #4524, opened 2017-09-19
Support cardinal directions from route relations in refs

danpat:
# Issue

This is a re-creation of https://github.com/Project-OSRM/osrm-backend/pull/4478 using a branch in the Project-OSRM project so we can make a beta release tag on the branch and publish binaries for it.

It adds Lua scripting which looks for `north/south/east/west` roles in route relations, and updates the `ref` field on ways to include that direction if refs matching the route relation names are found.

This lets us do things like "Take the ramp onto I95 North" when the word "North" is not actually part of the ref name on the way, but the `role` on the relation says "north".

## Tasklist
 - [x] review
 - [ ] adjust for comments

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
   - Changes from 5.12:
     - Profile:
       - New function to support relations: `process_relation`. Read more in profiles documentation.
+      - Append cardinal directions from route relations to ref fields to improve instructions
       - Support of `distance` weight in foot and bicycle profiles
       - Added `way:get_location_tag(key)` method to get location-dependent tags https://github.com/Project-OSRM/osrm-backend/wiki/Using-location-dependent-data-in-profiles
       - left-side driving mode is specified by a local Boolean flag `is_left_hand_driving` in `ExtractionWay` and `ExtractionTurn`
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,9 +1,9 @@
 # UNRELEASED
   - Changes from 5.12:
     - Profile:
-      - New function to support relations: `process_relation`. Read more in profiles documentation.
       - Append cardinal directions from route relations to ref fields to improve instructions
       - Support of `distance` weight in foot and bicycle profiles
+      - Support of relations processing
       - Added `way:get_location_tag(key)` method to get location-dependent tags https://github.com/Project-OSRM/osrm-backend/wiki/Using-location-dependent-data-in-profiles
       - left-side driving mode is specified by a local Boolean flag `is_left_hand_driving` in `ExtractionWay` and `ExtractionTurn`
     - Infrastructure:
--- a/profiles/car.lua
+++ b/profiles/car.lua
@@ -427,6 +427,8 @@ function process_way(profile, way, result, relations)
       end
     end
 
+    -- print(result.name, ref)
+
     result.ref = ref
   end
 end
--- a/profiles/car.lua
+++ b/profiles/car.lua
@@ -406,7 +406,7 @@ function process_way(profile, way, result, relations)
   local rel_id_list = relations:get_relations(way)
   for i, rel_id in ipairs(rel_id_list) do
     local rel = relations:relation(rel_id)
-    parsed_rel_list[i] = Relations.parse_route_relation(rel, way)
+    parsed_rel_list[i] = Relations.parse_route_relation(rel, way, relations)
   end
 
   -- now process relations data
--- a/profiles/car.lua
+++ b/profiles/car.lua
@@ -404,7 +404,7 @@ function process_way(profile, way, result, relations)
     matched_refs = Relations.MatchToRef(relations, result.ref)
 
     local ref = ''
-    for k ,v in pairs(matched_refs) do
+    for k, v in pairs(matched_refs) do
       if ref ~= '' then
         ref = ref .. '; '
       end
@@ -412,24 +412,12 @@ function process_way(profile, way, result, relations)
       if v then
         ref = ref .. k .. ' $' .. v
       else
-        print(ref)
         ref = ref .. k
       end
     end
 
     result.ref = ref
-    -- count = 0
-    -- for k, v in pairs(matched_refs) do
-    --   count = count + 1
-    -- end
-    -- if count > 1 then
-    --   print('---', way:id())
-    --   for k, v in pairs(matched_refs) do
-    --     print(k, v)
-    --   end
-    -- end
   end
-
 
 end
 
--- a/profiles/car.lua
+++ b/profiles/car.lua
@@ -403,20 +403,16 @@ function process_way(profile, way, result, relations)
   if result.ref then
     local match_res = Relations.match_to_ref(relations, result.ref)
 
-    order = match_res['order']
-    matched_refs = match_res['match']
-
     local ref = ''
-    for _, k in pairs(order) do
-      local v = matched_refs[k]
+    for _, m in pairs(match_res) do
       if ref ~= '' then
         ref = ref .. '; '
       end
 
-      if v then
-        ref = ref .. k .. ' $' .. v
+      if m.dir then
+        ref = ref .. m.ref .. ' $' .. m.dir
       else
-        ref = ref .. k
+        ref = ref .. m.ref
       end
     end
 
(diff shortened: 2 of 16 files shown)


profiles/lib/utils.lua
```diff
@@ -0,0 +1,43 @@
...
+
+  return result
+end
+
+-- same as Utils.StringListTokens, but with many possible separators:
+-- ',' | ';' | ' '| '(' | ')'
+function Utils.tokenize_common(str)
+  result = {}
+  local idx = 0
+  for s in str.gmatch(str, "%S+") do
```
1ec5: https://github.com/Project-OSRM/osrm-backend/pull/4478#discussion_r138776109

> If I’m not mistaken, this code splits a string by whitespace; it doesn’t seem to split by other separators like `;` or `(`. Note that an item in `ref` may contain parentheses, for example [`ref=A1(M)` on a way](http://www.openstreetmap.org/way/280766464) or [on a relation](http://www.openstreetmap.org/relation/103623).

profiles/lib/relations.lua
```diff
@@ -0,0 +1,94 @@
...
+  function calculate_scores(refs, tag_value)
+    local tag_tokens = Set(Utils.tokenize_common(tag_value))
+    local result = {}
+    for i, r in ipairs(refs) do
+      local ref_tokens = Utils.tokenize_common(r)
+      local score = 0
+
+      for _, t in ipairs(ref_tokens) do
+        if tag_tokens[t] then
+          if Utils.is_number(t) then
```
1ec5: https://github.com/Project-OSRM/osrm-backend/pull/4478#discussion_r139229061

> After rereading this code, I think the suffix case will be a nonissue. However, for [this way](http://www.openstreetmap.org/way/122502126), wouldn’t the two items in the way `ref` (`PA 88` and `PA 88 Truck`) tie because `88` appears in both items and both relations’ `ref` tags are set to `88`? We could factor in the relations’ `network` tags, sort of like what’s described in https://github.com/Project-OSRM/osrm-backend/pull/4478#discussion_r138947028.

profiles/lib/relations.lua
```diff
@@ -0,0 +1,94 @@
...
+            score = score + 2
+          else
+            score = score + 1
```
1ec5: https://github.com/Project-OSRM/osrm-backend/pull/4478#discussion_r138778008

> In Project-OSRM/osrm-backend#4434, I proposed trying to match a prefix in the way’s `ref` to a component of the relation‘s `network` as an additional score to factor in. Via [this Overpass query](http://overpass-turbo.eu/s/rHB), I found [one way](http://www.openstreetmap.org/way/122502126) that’s simultaneously part of `PA 88 South` and `PA 88 Truck North`.
> 
> The state-by-state queries in https://github.com/Project-OSRM/osrm-text-instructions/issues/119#issuecomment-328053084 can only find relations tagged with directions in roles; they’d need to be reworked to handle `direction` tags on relations. But given that I’ve only found one problematic way in the U.S. so far, matching `networks` is probably overkill, even if it’d be more robust.

profiles/car.lua
```diff
@@ -390,6 +392,74 @@ function process_way(profile, way, result)
...
+        end
+      end
+    end
+
+    local direction = relation:get_value_by_key('direction')
+    if direction then
+      direction = string.lower(direction)
+      if direction == 'north' or direction == 'south' or direction == 'west' or direction == 'east' then
+        for _, m in ipairs(relation:members()) do
+          if m:role() == 'forward' then
```
1ec5: [This way](http://www.openstreetmap.org/way/261596707) ends up with a `ref` of `CA 17`, not `CA 17 $South` as expected, because its role in the [southbound CA 17 relation](http://www.openstreetmap.org/relation/309738) is empty. It’s pretty common for the role to be omitted on one-way roads (in this case, one side of a divided highway).

1ec5: When the role is `backward`, we should reverse the `direction`, `$north` becoming `$south` etc. The role may be `backward` due to a wrong-way concurrency or simply because the mapper found it more convenient to make the way travel in the direction opposite of that particular route.

profiles/car.lua
```diff
@@ -390,6 +392,74 @@ function process_way(profile, way, result)
...
+    local match_res = Relations.match_to_ref(relations, result.ref)
+
+    local ref = ''
+    for _, m in pairs(match_res) do
+      if ref ~= '' then
+        ref = ref .. '; '
+      end
+
+      if m.dir then
+        ref = ref .. m.ref .. ' $' .. m.dir
```
1ec5: This style of appending a cardinal direction to a route number is specified by the [MUTCD](https://mutcd.fhwa.dot.gov/htm/2009r1r2/part2/part2e.htm#section2E17) (used in the U.S., Canada, and a few other regions), but [`direction` is used all over the world](https://taginfo.openstreetmap.org/keys/direction#map). I’m not sure if it’s used on road route relations in non-MUTCD countries, or if it would be surprising for users to see cardinal directions appended in those countries.

Reusing the raw relation role or `direction` tag in the human-readable `ref` also assumes that the instruction will ultimately be given in English. If the user is viewing a route in Québec in French, they would expect [this way](http://www.openstreetmap.org/way/470714815) to have a ref of `35 Nord; 104 Ouest`, not `35 North; 104 West`. It would make sense for OSRM Text Instructions to be responsible for translating the cardinal direction into the right language, but that would entail some iffy text replacement unless the ref is also provided in a structured format (#3329).

deniskoronchik:
Working on new extraction process for relations. Main idea, that we have any set of allowed relation types in lua profile. And if processing relation has on of this types, then it store into memory into map `rel_data`. It just contains all tags with values and map `OsmIDTyped -> role`.

Also we build 3 additional maps for ways, nodes and relations that are members is any relation.
It looks like: `OsmID -> [RelationID; ...]`.

So it allows to store data just we need to process filtered relations.
In lua we have no function `process_relation`. Instead of it `relations` parameter in `process_way` and `process_node` functions, grant access to any relations for specified object.
Like:
```lua
function process_way(profile, way, result, relations)
    local parent_relations = relations:get_relations(way)
    -- this list will contain id's of all relations, where this way is a member
    -- after that we can access their data from lua and process
    -- also we can get another parent relations for this ones if we need
    -- can get role of this way in each relation
end
```

danpat:
I think we should try to get this PR into OSRM 5.13.0

I've opened a new ticket for forward/backward support https://github.com/Project-OSRM/osrm-backend/issues/4590, I think we can do that as a second step.

@deniskoronchik can you rebase this PR on `master`, and comment out the tests that need forward/backward support?  I think this is basically ready to go, and we can work on #4590 for the next release.

deniskoronchik:
What do you mean?

emiltin:
in tests like this i see "$" in the ref fields:

```gherkin
When I route I should get
  | waypoints | route                | ref                                              |
  | b,a       | westbound,westbound  | I 80 $west,I 80 $west                            |
  | c,d       | eastbound,eastbound  | I 80 $east; CO 93 $east,I 80 $east; CO 93 $east  |
```

deniskoronchik:
Main idea, that `$north`, `$south`, `$west`, `$east` should be replaced to localized values in UI. 
`$` is a prefix, that allows to flag this token as a special symbol (not part of name)

CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
   - Changes from 5.12:
     - Profile:
       - New function to support relations: `process_relation`. Read more in profiles documentation.
+      - Append cardinal directions from route relations to ref fields to improve instructions
       - Support of `distance` weight in foot and bicycle profiles
```
TheMarex: Changelog should contain an entry for the new relation support.

profiles/lib/utils.lua
deniskoronchik: But in this example it should work correctly, because `A1(M)` is a ref in a way and relation.
So `A1(M)` - token that equal in both. Another way if there are any situations for example in way: `A1(M)` in relation: `A1 M`.

We should be careful, because using of any symbol as a separator, can increase number false-positive macthings.

profiles/lib/relations.lua
deniskoronchik: Maybe we can decrease points by `0.1 * N`, where N is a number of not matched tokens?
So in that case we will match to ref, where most of tokens was matched

profiles/lib/relations.lua
deniskoronchik: So where i should to get `networks` tag?

1ec5: The [`network`](https://taginfo.openstreetmap.org/keys/network) tag primarily occurs on route relations, alongside the `direction` tag this PR already uses.

deniskoronchik: Ok, i'll do that soon

Issue #4525, opened 2017-09-20
Lua 5.2 exact required?


peter-evans:
While testing the latest [v5.13.0-cardinal.1](https://github.com/Project-OSRM/osrm-backend/releases/tag/v5.13.0-cardinal.1) release I got the following error.
```
-- Could NOT find Lua: Found unsuitable version "5.2.3", but required is exact version "5.2" (found /usr/lib/x86_64-linux-gnu/liblua5.2.so;/usr/lib/x86_64-linux-gnu/libm.so)
CMake Error at CMakeLists.txt:524 (MESSAGE):
  Lua 5.2 was not found.
```
It seems that due to [this commit](https://github.com/Project-OSRM/osrm-backend/commit/f4f65f62ee460e148356f7d0f5af931c8cfc4630#diff-af3b638bc2a3e6c650974192a53c7291) Lua 5.2 exact is required. Is this intentional?

danpat:
@peter-evans yes, this was intentional, but these are beta releases, so this is not intended to be a final decision.

Some of the Lua interface code is significantly simpler under Lua 5.2, we couldn't see a compelling reason to keep the extra overhead around to maintain Lua 5.1 support - do you have a use-case that requires Lua 5.1, or is an update to 5.2 feasible for you?

peter-evans:
@danpat Sorry, I should have been a bit clearer about the issue. This is not related to support for 5.1 but about how a recent commit to CMakeLists.txt enforces Lua 5.2 **exact**.

Before [this commit](https://github.com/Project-OSRM/osrm-backend/commit/f4f65f62ee460e148356f7d0f5af931c8cfc4630#diff-af3b638bc2a3e6c650974192a53c7291) any minor version of Lua 5.2 could be used. e.g. 5.2.3
```
FIND_PACKAGE(Lua 5.2)
IF (LUA_FOUND)
    MESSAGE(STATUS "Using Lua ${LUA_VERSION_STRING}")
```
After the commit the `EXACT` keyword seems to enforce Lua 5.2(.0) to be present.
```
FIND_PACKAGE(Lua 5.2 EXACT)
IF (LUA_FOUND)
    MESSAGE(STATUS "Using Lua ${LUA_VERSION_STRING}")
```

I understand this is a beta release and not final. This is just testing to check for possible issues.

oxidase:
[...]

PR #4526, opened 2017-09-20
Fix FindLua for names like liblua-5.2.so.0

oxidase:
# Issue

Inspired by talks https://www.slideshare.net/DanielPfeifer1/effective-cmake and http://effective-cmake.com/cmake-talk.pdf and due to recent changes in Lua handlers that require sol2 support with Lua 5.2 and cmake check for exact 5.2.x version #4525, I would suggest to bump cmake to version 3 and refactor later CMake dependencies via CMake 3 features.

## Tasklist
 - [ ] review
 - [ ] adjust for comments

## Requirements / Relations


daniel-j-h:
2.8.11 comes with Ubuntu Trusty by default - if the only reason we want to switch to CMake 3 is the EXACT Lua version check, then should we simply drop the EXACT for now?

oxidase:
@daniel-j-h just found the origin of the configuration fail on an alpine distribution: as here https://pkgs.alpinelinux.org/contents?branch=edge&name=lua5.2-dev&arch=x86_64&repo=main the linker name `liblua.so` is in `/usr/lib/lua5.2` and cmake fails to find the library. I added `pkg_check_modules` that will update library directories for `find_library`

Issue #4531, opened 2017-09-21
`bearing_after` incorrect when we collaspse turns in post-processing to make a uturn


danpat:
This route:

https://www.mapbox.com/get-directions/#16.71/59.33325/18.03711?coordinates=18.040425689034805,59.33337310215188;18.0337997663797,59.33312220932879

performs a u-turn across a divided road.

The `bearing_after` field in the u-turn step seems to be the value from the joining road, not the exit to the actual u-turn:

<img width="398" alt="screen shot 2017-09-21 at 12 39 22 pm" src="https://user-images.githubusercontent.com/1892250/30715101-0f095176-9eca-11e7-9571-7c9562c679da.png">

It looks like we're not copying the `bearing_after` field from the second step when collapsing uturns.

PR #4532, opened 2017-09-21
Fix bearing_after when collapsing turns

danpat:
# Issue

This addresses https://github.com/Project-OSRM/osrm-backend/issues/4531

When we collapse turns in post-processing, it looks like some of the fields remain from the initial turn, when we should instead be copying some data from the second (hidden) maneuver.

This PR addresses the `bearing_after` field, which should be from the second step, not the first.

TODO: only contains a test case so far, needs to be actually fixed.

## Tasklist
 - [ ] review
 - [ ] adjust for comments

/cc @MoKob @oxidase


daniel-j-h:
@danpat what needs to be done here? Is this still actionable? Can you add the test with a todo tag so it does not run by default and then we can keep the ticket open for tracking.

DennisOSRM:

PR #4540, opened 2017-09-25
improve handling of route tags in profiles

emiltin:
WayHandler includes separate handlers for ferries, trains and movable bridges, that had very similar code. This PR replaced them with WayHandler.routes(), which handles all these. It read profile.routes, where you can setup what tag combinations to route on, as well as the speed and mode. You can also control whether specific access is required, e.g. car=yes, before the route is used.

Settings are organized as a hierachy. You can set defaults in the parent struct (profiles.route) and override them for specific tag combinations. If they are not set at all, the profile defaults are used.

The handler also allows you to set the rate. However, this is currently overridden later in WayHandler.penalties. I think the way penalties() and weight() handlers work are not optimal, but that's outside the scope of this PR.

In general it seems the speeds of routes are not set very realistically, e.g. trains are set to 10 km/h. This is a rudiment from the time where rate was not available. We should support rates for routes and set the speed to realistic values. It can be handled in a separate PR.

The foot profile was not handling trains - is that on purpose? If we handle rate, we could it it with realistic speeds and a low rate.

This PR is an extract from #4314, which was getting too big.

### DIFF
--- a/profiles/car.lua
+++ b/profiles/car.lua
@@ -171,13 +171,41 @@ function setup()
       'minor',
     },
 
-    route_speeds = {
-      ferry = 5,
-      shuttle_train = 10
-    },
-
-    bridge_speeds = {
-      movable = 5
+    routes = {
+      access_required = false,
+      keys = {
+        railway = {
+          mode = mode.train,
+          speed = 10,
+          values = Set {
+            'train',
+            'railway',
+            'subway',
+            'light_rail',
+            'monorail',
+            'tram'
+          }
+        },
+        route = {
+          values = {
+            ferry = {
+            speed = 5,
+            mode = mode.ferry
+            },
+            shuttle_train = {
+              mode = mode.train,
+              speed = 10
+           }
+          }
+        },
+        bridge = {
+          values = {
+            movable = {
+              speed = 5
+            }
+          }
+        }
+      }
     },
 
     -- surface/trackype/smoothness
@@ -356,8 +384,7 @@ function process_way(profile, way, result)
     WayHandlers.destinations,
 
     -- check whether we're using a special transport mode
-    WayHandlers.ferries,
-    WayHandlers.movables,
+    WayHandlers.routes,
 
     -- handle service road restrictions
     WayHandlers.service,
--- a/profiles/lib/way_handlers.lua
+++ b/profiles/lib/way_handlers.lua
@@ -108,24 +108,6 @@ function WayHandlers.destinations(profile,way,result,data)
   end
 end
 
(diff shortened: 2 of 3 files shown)


profiles/car.lua
```diff
@@ -171,13 +171,41 @@ function setup()
...
     },
 
-    route_speeds = {
-      ferry = 5,
-      shuttle_train = 10
-    },
-
-    bridge_speeds = {
-      movable = 5
+    routes = {
```
TheMarex: If I'm reading this correctly, we now would allow routing on trains in the car profile? I think using public transportation is rather unexpected and leads to weird/impossible routes (car -> train -> car).

What about keeping the default here of just using ferries/movable bridges?

profiles/lib/way_handlers.lua
```diff
@@ -148,6 +130,76 @@ function WayHandlers.movables(profile,way,result,data)
...
+        local speed
+
+        local duration  = way:get_value_by_key("duration")
+        if duration and durationIsValid(duration) then
+          result.duration = math.max( parseDuration(duration), 1 )
+          result.weight = result.duration
+          speed = 0
+        else
+          speed = select_first_value( {v_speed, k_speed, default_speed } )
+          result.forward_rate = 1
```
TheMarex: I'm not sure what is going on here? The rate is set to `1` but overwritten on line 194.

profiles/car.lua
emiltin: yeah i agree, trains shot not be routable by deafult. only thing we might want by default is shuttle train? https://github.com/Project-OSRM/osrm-backend/blob/4297d7f56babc3302d0bc6a42cc0011a255d0a40/profiles/car.lua#L195

profiles/lib/way_handlers.lua
emiltin: it's a leftover, rate is now handled on line 194. however working on that part i was a bit unsure how to proceed, since rate is handled centrally in the weight handler - in #4543 i suggested further improvements into how rate is used for routes. we could choose to remove the rate support in this PR to keep it ismple?

profiles/car.lua
TheMarex: Yes `shuttle_train` makes sense to me.

profiles/lib/way_handlers.lua
TheMarex: >  we could choose to remove the rate support in this PR to keep it ismple?

Yeah that makes sense to me.

DennisOSRM:

Issue #4528, opened 2017-09-20
1xN shortest routes


oxidase:
1xN (Nx1) searches are not effective with the MLD many-to-many plugin using a single source (destination) node.

On other side MLD 1xN (Nx1) can be implemented as a single forward (backward) search using the `node` level `getNodeQueryLevel` as `min(partition.GetQueryLevel(source, target, node))` for all `target` nodes ( `source` nodes).

/cc @MoKob @TheMarex

oxidase:
Closed via #4547

PR #4547, opened 2017-09-26
MLD one-to-many as unidirectional search

oxidase:
# Issue

PR implements https://github.com/Project-OSRM/osrm-backend/issues/4528 

Also resolves https://github.com/Project-OSRM/osrm-backend/issues/4554

## Tasklist
 - [x] check incorrect duration with `osrm-runner`
 - [x] test performance wrt CH and many-to-many MLD
 - [x] review
 - [x] adjust for comments

## Requirements / Relations

### DIFF
--- a/src/engine/routing_algorithms/many_to_many.cpp
+++ b/src/engine/routing_algorithms/many_to_many.cpp
@@ -465,6 +465,7 @@ manyToManySearch(SearchEngineData<mld::Algorithm> &engine_working_data,
 namespace mld
 {
 
+// Unidirectional multi-layer Dijkstra search for 1-to-N and N-to-1 matrices
 template <bool DIRECTION>
 std::vector<EdgeDuration> oneToManySearch(SearchEngineData<Algorithm> &engine_working_data,
                                           const DataFacade<Algorithm> &facade,
@@ -518,6 +519,10 @@ std::vector<EdgeDuration> oneToManySearch(SearchEngineData<Algorithm> &engine_wo
         }
     }
 
+    // Initialize query heap
+    engine_working_data.InitializeOrClearManyToManyThreadLocalStorage(facade.GetNumberOfNodes());
+    auto &query_heap = *(engine_working_data.many_to_many_heap);
+
     // Check if node is in the destinations list and update weights/durations
     auto update_values = [&](NodeID node, EdgeWeight weight, EdgeDuration duration) {
         auto candidates = target_nodes_index.equal_range(node);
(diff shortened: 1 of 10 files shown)


oxidase:
Some benchmarks with random 1x25 queries for different areas:

* Berlin-sized area 1x25 with CH, 1-to-N MLD and N-to-N MLD algorithms
o2m_sources

* Berlin-sized area 25x1 with CH, 1-to-N MLD and N-to-N MLD algorithms
o2m_destinations

* Bavaria-sized area
o2m_sources_bayern

* Germany-sized area
o2m_sources_germany

src/engine/routing_algorithms/many_to_many.cpp
```diff
@@ -120,27 +121,65 @@ addLoopWeight(const DataFacade<mld::Algorithm> &, const NodeID, EdgeWeight &, Ed
...
+    };
+
+    const auto &source_phantom = phantom_nodes[phantom_index];
+
+    auto result = INVALID_LEVEL_ID;
+    for (const auto &index : phantom_indices)
+    {
+        const auto &target_phantom = phantom_nodes[index];
+        auto min_level = std::min(
+            std::min(level(source_phantom.forward_segment_id, target_phantom.forward_segment_id),
```
MoKob: It seems to me that this could be implemented a bit more performant, since the level between `node` and `source_phantom` does not change over the loop. Depending on how often this is called, it might be a small performance bottle-neck.

Also, we could filter this list for vertices within the same cell to eliminate duplicate checks.

src/engine/routing_algorithms/many_to_many.cpp
```diff
@@ -282,15 +324,15 @@ void forwardRoutingStep(const DataFacade<Algorithm> &facade,
...
+        else if (std::tie(new_weight, new_duration) < std::tie(current_weight, current_duration))
         {
             current_weight = new_weight;
             current_duration = new_duration;
         }
     }
 
     relaxOutgoingEdges<FORWARD_DIRECTION>(
-        facade, node, source_weight, source_duration, query_heap, phantom_node);
+        facade, query_heap, node, source_weight, source_duration, phantom_node);
```
MoKob: any specific reason why you changed the order here? -- fine with the change, but curious

oxidase: no, i will revert this particular change

src/engine/routing_algorithms/many_to_many.cpp
oxidase: ok, i will update here

oxidase:
@MoKoB i have uploaded images with many-to-many plugin. For larger areas does not look as good as  for cities but still of the same order as the many-to-many implementation. But many-to-many plugin is parallelized and one-to-many is serial.

MoKob:
@oxidase I would still say that the times are an improvement. The maximum times might not have changed much, but if I interpret your graphs correctly, the average query time even for Germany seems to be at least halved. The majority of the request seem quicker now than before. Also, can you run the experiments with many-to-many only using a single thread? That would give a better impression, since the total workload should still improve quite a bit, if we only require a single thread now.

oxidase:
@MoKob i re-run benchmarks on germany-latest.osm.pbf with 1000 raundom 1x25 and 25x1 queries with three different bounding boxes for random queries with serial implementation of many-to-many plugin:

* `13.0828,52.3345,13.7614,52.6783` city-sizes area
  - 1x25
o2m_berlin_sources
  - 25x1
o2m_berlin_destinations

* `8.97465,47.2284,14.0903,50.5678` state-sized area
  - 1x25
o2m_bayern_sources
  - 25x1
o2m_bayern_destinations


* `5.86442,47.2654,15.0508,55.1478` country-sized area
  - 1x25
o2m_germany_sources
  - 25x1
o2m_germany_destinations

yes, it is a good improvement for city-sized areas, also for country-sized areas MLD one-to-many implementation will be not slower than parallelized many-to-many implementation and in the worst case query-time will be the same, because one full hierarchy search is still needed in both implementations.

MoKob:
> because one full hierarchy search is still needed in both implementations.

This assumes more or less evenly distributed points, right? For the chance of close points, the one-to-many will automatically use the pruning that is still up for implementation, right? So we would also be much faster on large datasets, if the points do not space out a lot.

oxidase:
yes, i used uniformly distributed points in bounding boxes, so that is why i have 3 different boxes with different areas

MoKob:
Ah, I misunderstood. I though you were running on the extracts. All are on the German road network? So I would say that should cover many of our use-cases.

PR #4551, opened 2017-09-28
Count actual roundabout exits, not just passed steps

danpat:
# Issue

When a node on a roundabout has multiple exits directly connected, we only count it as one exit from the roundabout, which throws off the exit number counting for following exits.

*Usually* this is a result of weird modeling in OSM - multiple exits from the same point on the roundabout is kind of odd, but we found one at http://www.openstreetmap.org/node/49861176, which was where http://www.openstreetmap.org/way/528145111 was connected until @willwhite adjusted it.

This PR looks at the actual exit counts at each step around the roundabout and tallies them up, rather than just counting the number of steps between the entry and the exit.

This could also be a precursor to fixing https://github.com/Project-OSRM/osrm-backend/issues/2533 - one idea I had was possibly not counting exits that also have a mode change.

@MoKob can you eyeball this and see if my logic is sound?  Also open to suggestions if there is a more elegant approach to iterating around the `entry[]` array :-)

TODO: proper left/right driving side lookup from the profile.

## Tasklist
 - [ ] review
 - [ ] adjust for comments

## Requirements / Relations

### DIFF
--- a/features/guidance/roundabout.feature
+++ b/features/guidance/roundabout.feature
@@ -845,3 +845,34 @@ Feature: Basic Roundabout
             | from | to | route                      | turns                                                   | distance |
             | e    | k  | ebds,ufghl,ufghl,jhik,jhik | depart,rstur-exit-2,exit rotary right,turn right,arrive | 189.1m   |
             | 1    | k  | ebds,ufghl,ufghl,jhik,jhik | depart,rstur-exit-2,exit rotary right,turn right,arrive | 159.1m   |
+
+
+    Scenario: count exit numbers properly when multiple exits from a single node
+        Given the node map
+            """
+                 f-----e
+                /       \
+            g--a     l   d-----h
+                \     \ /
+                 b-----c
+                        \\\
+                         ijk
+
+            """
+
+        And the ways
+            | nodes   | highway  | junction   | oneway |
+            | abcdefa | tertiary | roundabout | yes    |
+            | ga      | tertiary |            |        |
+            | ci      | tertiary |            |        |
+            | cj      | tertiary |            |        |
+            | ck      | tertiary |            |        |
+            | cl      | tertiary |            |        |
+            | dh      | tertiary |            |        |
+
+        When I route I should get
+            | from | to | route       | turns                                          |
+            | g    | i  | ga,ci,ci,ci | depart,abcdefa-exit-1,exit rotary right,arrive |
+            | g    | j  | ga,cj,cj,cj | depart,abcdefa-exit-2,exit rotary right,arrive |
+            | g    | k  | ga,ck,ck,ck | depart,abcdefa-exit-3,exit rotary right,arrive |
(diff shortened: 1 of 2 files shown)


src/engine/guidance/post_processing.cpp
```diff
@@ -88,13 +88,66 @@ void processRoundaboutExits(const RouteStepIterator begin, const RouteStepIterat
...
-        return staysOnRoundabout(step.maneuver.instruction) ||
-               leavesRoundabout(step.maneuver.instruction);
-    };
+    const auto exit = std::accumulate(begin, end, 0, [](const long a, const auto &step) {
+        if (staysOnRoundabout(step.maneuver.instruction))
+        {
+            // Count all the possible exits from from the step, minus one, which is the
+            // "continue on the roundabout" exit
+            const auto step_intersection = step.intersections.front();
+            const auto num_exits_from_step = std::count_if(std::begin(step_intersection.entry),
```
MoKob: This logic doesn't work for throughabouts. If a road (bidirectional) passes through a roundabout, the exit will be counted twice.

features/guidance/roundabout.feature
```diff
@@ -845,3 +845,34 @@ Feature: Basic Roundabout
...
+            | ck      | tertiary |            |        |
+            | cl      | tertiary |            |        |
+            | dh      | tertiary |            |        |
+
+        When I route I should get
+            | from | to | route       | turns                                          |
+            | g    | i  | ga,ci,ci,ci | depart,abcdefa-exit-1,exit rotary right,arrive |
+            | g    | j  | ga,cj,cj,cj | depart,abcdefa-exit-2,exit rotary right,arrive |
+            | g    | k  | ga,ck,ck,ck | depart,abcdefa-exit-3,exit rotary right,arrive |
+            | g    | h  | ga,dh,dh,dh | depart,abcdefa-exit-5,exit rotary right,arrive |
```
MoKob: How is `l` decided as being the `fourth` exit. It looks similar to `i` in terms of distance from `g`?

src/engine/guidance/post_processing.cpp
MoKob: Also, this would need to ignore exits that cannot be entered (exit against roundabout direction)

miccolis:
@danpat I though we discussed this and decided the current behavior is reasonable? Shouldn't we look for a case of a roundabout that is modeled like this and looks to be done correctly? ...the Logan circle was pretty much bad modeling.

TheMarex:
Capturing from chat with @danpat: He still thinks this is the right approach but need to revisit when he has time.

danpat:
I chatted with @mokob about this.  Next step is to scan the planet for instances and see if this is even a legit way to model things - there's a nonzero chance that it's never correct to exit a roundabout more than once from a single node.

If it turns out that all instances of this are from poor modeling, this or should change into something that logs warnings during extraction.

DennisOSRM:

PR #4555, opened 2017-09-29
Detect segregated nodes.

vng:
https://github.com/Project-OSRM/osrm-backend/issues/4273

### DIFF
--- a/include/engine/plugins/crossing_check.hpp
+++ b/include/engine/plugins/crossing_check.hpp
@@ -121,21 +121,22 @@ private:
 
       for (auto edge : m_facade.GetAdjacentEdgeRange(e.m_id))
       {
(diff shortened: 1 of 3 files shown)


MoKob:
@vng , from the conversations on the calls, I would suggest doing the detection earlier on in the pipeline. My main hopes in doing this earlier would be the availability of more data and easier traversal of surrounding edges, than what we have in the tile plugin. The data can still be visualised in the tile plugin, if you add a flag. While this will increase memory for now, we could later decide how to path the data through to the engine.

vng:
@MoKob I agree with you. For now it's made in tile processing to have visualization demo. Could you provide me entry point in the pipeline code where I can have access to the graph data?

Issue #4527, opened 2017-09-20
Speedup MLD many-to-may search


oxidase:
As discussed with @MoKob the MLD many-to-many search space can be reduced by stopping the backward search at the top-most level cell [here](https://github.com/Project-OSRM/osrm-backend/blob/fa1a4e8bf6589f8483ff8fe412ee289db3615d52/src/engine/routing_algorithms/many_to_many.cpp#L359-L363) and not explore the complete graph. 

The forward search also can be stopped earlier [here](https://github.com/Project-OSRM/osrm-backend/blob/fa1a4e8bf6589f8483ff8fe412ee289db3615d52/src/engine/routing_algorithms/many_to_many.cpp#L391-L402) if during the backward search a set of boundary nodes at the top most level is collected and all the nodes from the collected set are settled during the forward search.

/cc @MoKob

PR #4558, opened 2017-09-29
M2M: suppress relaxation of edges on top level during backward steps

oxidase:
# Issue

PR implements #4527 

Running similar benchmarks for 1x25, 25x1 and 25x25 uniformly random 1000 queries in 3 areas on one extracted germany-latest.osm.pbf data set  (for 1x25 and 25x1 queries i used manyToMany implementation here and not oneToMany and without parallelization):

* `13.0828,52.3345,13.7614,52.6783` city-sizes area
  - 1x25 queries
 o2m_top_berlin_sources
  - 25x1 queries
 o2m_top_berlin_destinations
  - 25x25 queries 
o2m_mld_berlin_all

* `8.97465,47.2284,14.0903,50.5678` state-sized area
  - 1x25 
o2m_top_bayern_sources

  - 25x1 
o2m_top_bayern_destinations

  - 25x25 
o2m_mld_bayern_all

* `5.86442,47.2654,15.0508,55.1478` country-sized area
  - 1x25 
o2m_top_germany_sources

  - 25x1 
o2m_top_germany_destinations

  - 25x25 
o2m_mld_germany_all

Summary:
* good improvement for `mxN` queries for small `m`
* moderate improvement for `NxN` queries
* no improvement  for `Nx1` queries, but it is can be improved by transposing `Nx1` queries to `1xN` but this will require complete split of CH and MLD code for `manyToMany`

/cc @MoKob 

## Tasklist
 - [ ] review
 - [ ] adjust for comments

## Requirements / Relations


src/engine/routing_algorithms/many_to_many.cpp
```diff
@@ -132,11 +132,16 @@ inline LevelID getNodeQueryLevel(const MultiLevelPartition &partition,
         return INVALID_LEVEL_ID;
     };
 
-    return std::min(highest_diffrent_level(phantom_node.forward_segment_id),
-                    highest_diffrent_level(phantom_node.reverse_segment_id));
+    const auto node_level = std::min(highest_diffrent_level(phantom_node.forward_segment_id),
```
MoKob: Shouldn't a phantom node be in the same cell with all of it's segments? Why do we need both comparisons here?

src/engine/routing_algorithms/many_to_many.cpp
```diff
@@ -179,11 +184,15 @@ void relaxOutgoingEdges(const DataFacade<mld::Algorithm> &facade,
...
-    const auto &metric = facade.GetCellMetric();
 
-    const auto level = getNodeQueryLevel(partition, node, args...);
+    const auto level = getNodeQueryLevel<STOP_AT_TOP_LEVEL>(partition, node, args...);
+
+    // Break outgoing edges relaxation if node at the restricted level
+    if (level == INVALID_LEVEL_ID)
+        return;
 
+    const auto &cells = facade.GetCellStorage();
```
MoKob: These two don't come up later in this PR. Why do we need them now?

src/engine/routing_algorithms/many_to_many.cpp
```diff
@@ -71,7 +71,7 @@ inline bool addLoopWeight(const DataFacade<ch::Algorithm> &facade,
     return false;
 }
 
-template <bool DIRECTION>
+template <bool DIRECTION, bool STOP_AT_TOP_LEVEL = false>
```
MoKob: Is there a way that we could avoid having an unused template paramter here? This parameter does not make sense for CH at all

src/engine/routing_algorithms/many_to_many.cpp
oxidase: it is not guaranteed in EBG after renumbering and re-partitioning of EBG nodes. I can place an assertion to check the condition, but would like to keep atm check for forward and reverse segments.

MoKob: In case we actually do not put the same node into the same partition, I would say the check is the better way to go. 
Is there a reason for not hitting that guarantee, though? Seems to me that the same location should always be in the same cell.

src/engine/routing_algorithms/many_to_many.cpp
oxidase: yes, the parameter is not relevant for CH. 
I would like to split CH and MLD table implementations, this will allow to improve MLD performance for both `few-to-many` and `many-to-few` queries, because atm only former are improved.

src/engine/routing_algorithms/many_to_many.cpp
MoKob: ahh. sorry. Didn't see them in the diff from before. I thought you were adding new unused variables here. fine to keep as is.

src/engine/routing_algorithms/many_to_many.cpp
oxidase: i would rather be careful here, EBG nodes of the same NBG edge can be potentially in different cells.
I would like to keep the condition here as `min(forward level, reverse level)` as in the rest of the code and check the assumption in a separate issue/pr.

MoKob: @oxidase I am not asking you to change the check here, but asking if you know of a reason we do not keep a single nbg edge in a single cell. Separating a single edge logically into two different cells seems like a strange concept to me and something we might need to rethink.

oxidase: i don't quite understand the question "... keep a single nbg edge in a single cell" . The reason for doing that is in using  an NBG partition to make an EBG one with some heuristics  how to place EBG boundary nodes in cells and this potentially could lead to the situation  "Separating a single edge logically into two different cells". This is not really optimal as also can result in "invalid" boundary nodes that neither sources nor destinations, but number of such nodes is vanishingly small. 

I would suggest to open an issue with ideas how to reimplement this, because i still don't understand benefits of doing NBG  instead of EBG partitioning.

MoKob: Opening an issue for this seems fine with me, the only actual issue that needs to change is the introduction of the template parameter below, since it is unrelated to CH.

oxidase:
@MoKob I have split CH and MLD implementations to avoid confusing MLD flags in CH implementation and removed parallelization (#4560). Also i changed MLD implementation to always compute matrix from few to many nodes. 

Here some results for `germany-latest.osm.pbf` extract:

* 2x25 matrix queries with nodes uniformly distributed in  `13.0828,52.3345,13.7614,52.6783` city-sized area
s2m_berlin_sources


* 25x2 matrix queries with nodes uniformly distributed in  `13.0828,52.3345,13.7614,52.6783` city-sized area
s2m_berlin_sources


* 2x25 matrix queries with nodes uniformly distributed in `8.97465,47.2284,14.0903,50.5678` state-sized area
s2m_bayern_sources


* 25x2 matrix queries with nodes uniformly distributed in `8.97465,47.2284,14.0903,50.5678` state-sized area
s2m_bayern_destinations


* 2x25 matrix queries with nodes uniformly distributed in `5.86442,47.2654,15.0508,55.1478` country-sized area
s2m_bayern_sources


* 25x2 matrix queries with nodes uniformly distributed in `5.86442,47.2654,15.0508,55.1478` country-sized area
s2m_berlin_destinations

features/testbot/distance_matrix.feature
```diff
@@ -124,7 +124,7 @@ Feature: Basic Distance Matrix
             | d | 20 | 30  | 0  | 30 |
             | e | 30 | 40  | 10 | 0  |
 
-    Scenario: Testbot - Travel time matrix and with only one source
+    Scenario: Testbot - Rectangular travel time matrix
```
MoKob: should we ensure these are tested over partition-boundaries? I can't see any partition parameters specified here. Since we do dedicated search, it could be helpful to actually go across boundaries/test pruning to not keep values from us

include/engine/routing_algorithms.hpp
```diff
@@ -184,6 +184,32 @@ inline routing_algorithms::SubMatchingList RoutingAlgorithms<Algorithm>::MapMatc
 }
 
 template <typename Algorithm>
+std::vector<EdgeDuration> RoutingAlgorithms<Algorithm>::ManyToManySearch(
+    const std::vector<PhantomNode> &phantom_nodes,
+    const std::vector<std::size_t> &_source_indices,
```
MoKob: as we are copying anyhow, whouldn't it make sense to take by value? This way we could move from the outside, if we don't want to keep the data, also if we don't want to specify them, it would be clearer not to pass in an empty array

include/engine/routing_algorithms/many_to_many.hpp
```diff
@@ -16,24 +16,45 @@ namespace engine
 namespace routing_algorithms
 {
 
+namespace
+{
+struct NodeBucket
```
MoKob: This seems to be an MLD only structure but part of a general many_to_many implementation. Any way wie could connect it with MLD?

oxidase: it is also used in CH. I just moved `NodeBucket` declaration  to a common place

include/engine/routing_algorithms.hpp
oxidase: just looked at `RoutingAlgorithmsInterface` and looks like it is a common pattern to pass const references in interfaces. I will keep the original interface and copy vectors explicitly instead of hiding copy assignment in arguments code.

include/engine/routing_algorithms/many_to_many.hpp
MoKob: It specifically refers to a column in MLD in the commentary? Is that just a misnomor?

oxidase: no, it is always a column index 

At https://github.com/Project-OSRM/osrm-backend/blob/4b8f85115153755ee55b5e72183611d7362f0e6c/src/engine/routing_algorithms/many_to_many_mld.cpp#L404 it is used as a column index in column-major matrix that is in the same time  a transposed matrix if treated as a row-major matrix. I just moved a transpose operation at that line, otherwise an explicit transpose of returned matrix is required at https://github.com/Project-OSRM/osrm-backend/blob/4b8f85115153755ee55b5e72183611d7362f0e6c/src/engine/routing_algorithms/many_to_many_mld.cpp#L556-L557

MoKob: ah. sorry. miss-read this as column-index in the MLD matrix, not the result matrix. All good here.

PR #4562, opened 2017-10-02
Update LICENSE to 2017 and fix typo in filename

JackDanger:


daniel-j-h:
Thanks! I think the public libosrm headers also come with a license prepended e.g.

https://github.com/Project-OSRM/osrm-backend/blob/master/include/osrm/osrm.hpp#L3

would you be so kind and change all of these.

JackDanger:
> would you be so kind and change all of these.

Will do!

JackDanger:
I've update the copyright year everywhere that the copyright was assigned to Project ORSM contributors

PR #4563, opened 2017-10-02
Don't force loops in MLD

oxidase:
# Issue

MLD shortcuts are not suffer from self-loops as CH, so MLD [`routingStep`](https://github.com/Project-OSRM/osrm-backend/blob/3755046c940dfe79e03cd1401572653403ec4adf/include/engine/routing_algorithms/routing_base_mld.hpp#L260-L262) does not check special "single-node" paths. This was removed in 0972ec9115b912a76692b5c59e3d711b1f84af91 

This PR removes using of loops forcing in MLD alternatives code, but keeps naming as is.

Potential problems that are fixed by the PR
screenshot from 2017-10-02 11-56-12

```
curl -s 'localhost:5000/route/v1/driving/-122.454852,37.790617;-122.455153,37.790544?alternatives=true' | jq ".routes[].weight"
194.6
curl -s 'localhost:5000/route/v1/driving/-122.454852,37.790617;-122.455153,37.790544' | jq ".routes[].weight"
3.1
```

In addition alternative feature tests are enabled by the PR.

## Tasklist
 - [ ] review
 - [ ] adjust for comments

## Requirements / Relations

### DIFF
--- a/src/engine/routing_algorithms/alternative_path_mld.cpp
+++ b/src/engine/routing_algorithms/alternative_path_mld.cpp
@@ -456,9 +456,6 @@ void unpackPackedPaths(InputIt first,
     util::static_assert_iter_category<OutIt, std::output_iterator_tag>();
     util::static_assert_iter_value<InputIt, WeightedViaNodePackedPath>();
 
-    const bool force_loop_forward = needsLoopForward(phantom_node_pair);
-    const bool force_loop_backward = needsLoopBackwards(phantom_node_pair);
-
     const Partition &partition = facade.GetMultiLevelPartition();
 
     Heap &forward_heap = *search_engine_data.forward_heap_1;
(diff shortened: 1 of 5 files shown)


src/engine/routing_algorithms/alternative_path_mld.cpp
```diff
@@ -641,8 +635,8 @@ makeCandidateVias(SearchEngineData<Algorithm> &search_engine_data,
                                            forward_heap,
                                            overlap_via,
                                            overlap_weight,
-                                           force_loop_forward,
-                                           force_loop_backward,
+                                           DO_NOT_FORCE_LOOPS,
+                                           DO_NOT_FORCE_LOOPS,
```
daniel-j-h: If we never have to force loops for mld then the code path above here in the routing base mld

https://github.com/Project-OSRM/osrm-backend/pull/4563/files#diff-b397fc7de6b623ef23bec900b2bb5c27R258

can never be triggered and the routingStep interface for mld really should be changed, no?

daniel-j-h: Ah just seeing https://github.com/Project-OSRM/osrm-backend/issues/4564 👍

PR #4565, opened 2017-10-02
Remove unused packed path weights in MLD alternatives

oxidase:
# Issue

PR removes `retrievePackedPathWeightsFromHeap` call and unused `WeightedViaNodePackedPath::path_weights` member.

@daniel-j-h if it is planned to use the function it must be fixed as it does not compute correctly weights of edges from the source and to the destination nodes. The assertion `BOOST_ASSERT(via.weight == edge_weights + phantom_node_weights);` always correct because weight of the the edge to via node is computed as `via.weight - (weight_from + weight_to)` that compensate incorrectness of the first and last edges weights.
 

## Tasklist
 - [ ] review
 - [ ] adjust for comments

## Requirements / Relations

### DIFF
--- a/cucumber.js
+++ b/cucumber.js
@@ -3,5 +3,5 @@ module.exports = {
     verify: '--strict --tags ~@stress --tags ~@mld --tags ~@todo -f progress --require features/support --require features/step_definitions',
     todo: '--strict --tags @todo --require features/support --require features/step_definitions',
     all: '--strict --require features/support --require features/step_definitions',
-    mld: '--strict --tags ~@stress --tags ~@todo --tags ~@alternative --tags ~@ch --require features/support --require features/step_definitions -f progress'
+    mld: '--strict --tags ~@stress --tags ~@todo --tags ~@ch --require features/support --require features/step_definitions -f progress'
 };
--- a/features/testbot/alternative.feature
+++ b/features/testbot/alternative.feature
@@ -12,6 +12,9 @@ Feature: Alternative route
               g h i j
             """
 
+        # enforce multiple cells for filterUnpackedPathsBySharing check
+        And the partition extra arguments "--small-component-size 1 --max-cell-sizes 2,4,8,16"
+
         And the ways
             | nodes |
             | ab    |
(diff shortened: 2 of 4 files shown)


daniel-j-h:
Hm, yes the idea was definitely to use that function when we have the time to re-visit alternative route quality in the future.

Could you explain what's the problem here or how you see a fix would look like? You said the first and last weights are negative by design?

The assumption here was that our total weight is the edge weights plus the phantom node weights.

oxidase:
Ok, than it must be fixed. 

> You said the first and last weights are negative by design?
no, only source node weight is negative.

> The assumption here was that our total weight is the edge weights plus the phantom node weights.
This is correct assumption.

The weights must be computed similar as in `anotatePath`:
* compute separately the weight of the first edge
* compute edge weights till via node from the forward heap
* compute edge weights from via node from the reverse heap
* add the weight from the destination phantom to the last edge

Here is an example of EBG graph
alternative feature_32-fullpage

and here the packed paths from the left  to the right NBG node

The upper path 
```
(gdb) p via
$2 = {node = 18, weight = 1130}
(gdb) p packed_path
$3 = std::vector of length 4, capacity 4 = {std::tuple containing = {[1] = 15, [2] = 20, [3] = false}, std::tuple containing = {[1] = 20, [2] = 18, [3] = false}, std::tuple containing = {[1] = 18, [2] = 12, [3] = true}, std::tuple containing = {[1] = 12, [2] = 1, [3] = false}}
(gdb) p path_weights
$4 = std::vector of length 4, capacity 4 = {283, 200, 200, 447}
```

the bottom path
```
(gdb) p via
$5 = {node = 5, weight = 1166}
(gdb) p packed_path
$6 = std::vector of length 4, capacity 4 = {std::tuple containing = {[1] = 21, [2] = 16, [3] = false}, std::tuple containing = {[1] = 16, [2] = 5, [3] = false}, std::tuple containing = {[1] = 5, [2] = 10, [3] = true}, std::tuple containing = {[1] = 10, [2] = 2, [3] = false}}
(gdb) p path_weights
$7 = std::vector of length 4, capacity 4 = {283, 283, 400, 200}
```

Here path weights `{283, 283, 400, 200}` are incorrect and must be `{283,  200, 400, 283}`

I tried to find place were the hack `via.weight - (weight_from + weight_to)` could appear to add a test, but did not find any usage, so decided to remove :sunglasses:

PR #4583, opened 2017-10-08
Add literal maxspeed value for NO, PL and ZA on maxspeed_table in car profile

frodrigo:


frodrigo:
ZA is for South Africa

PR #4588, opened 2017-10-09
Detect segregated nodes, part 2.

vng:
#4273


src/extractor/extractor.cpp
```diff
@@ -808,5 +818,173 @@ void Extractor::WriteCompressedNodeBasedGraph(const std::string &path,
...
+      {
+        *(++curr) = *next;
+      }
+      else
+      {
+        BOOST_ASSERT(curr->name_id == next->name_id);
+        BOOST_ASSERT(curr->road_class == next->road_class);
+
+        BOOST_ASSERT(curr->direction != next->direction);
+        curr->direction = 2;
```
MoKob: could we split this into two runs to find which ones are bidirecitonal (I guess that is was you are shooting for here?) and then use std::unique?

src/extractor/extractor.cpp
```diff
@@ -808,5 +818,173 @@ void Extractor::WriteCompressedNodeBasedGraph(const std::string &path,
...
+    for (EdgeID edgeID : sourceEdges)
+    {
+      auto & edgeData = graph->GetEdgeData(edgeID);
+
+      if (edgeData.reversed || edgeData.segregated || !processed.insert(edgeID).second)
+        continue;
+
+      NodeID const targetID = graph->GetTarget(edgeID);
+
+      if (osrm::util::coordinate_calculation::haversineDistance(coordinates[sourceID], coordinates[targetID]) > 30.0)
```
MoKob: this 30.0 is a heuristic from post-processing and will need to be revised. It is very restrictive but with many specific cases it just does not work, since road width can take up most of this 30 meters. See above that simply removing this check would not work.

src/extractor/extractor.cpp
```diff
@@ -808,5 +818,173 @@ void Extractor::WriteCompressedNodeBasedGraph(const std::string &path,
...
+
+  preprocessFn(v1);
+  preprocessFn(v2);
+
+  // Internal edge with the name should be connected with any other neibour edge with the same name.
+  if (current.name_id != EMPTY_NAMEID)
+  {
+    auto const findNameFn = [&current] (std::vector<EdgeInfo> const & v)
+    {
+      return std::binary_search(v.begin(), v.end(), current, EdgeInfo::LessName());
```
MoKob: one thing I forgot to mention: name_ids are not necessarily unique for the same name:

`name: lincoln avenue` and `name: lincoln avenue, ref:I55` will result in two different name_ids, even though names remain the same. Adding a reference is a typical feature in OSM. You can check out `requiresNameAnnounced` in util for all name-related checks.

MoKob:
The results look much better already, I am finding many curious cases though:

- why is this kind of intersection not fully internal?
screen shot 2017-10-10 at 14 12 35

- this looks strange
screen shot 2017-10-10 at 14 14 24

- these staggered intersections are just too far, these are two separate turns
screen shot 2017-10-10 at 14 15 28

Most of what we see already looks great, though.

src/extractor/extractor.cpp
```diff
@@ -808,5 +818,173 @@ void Extractor::WriteCompressedNodeBasedGraph(const std::string &path,
...
+      return false;
+  }
+
+  // set_intersection like routine to count equal name pairs.
+  std::vector<std::pair<EdgeInfo const *, EdgeInfo const *>> commons;
+
+  auto i1 = v1.begin();
+  auto i2 = v2.begin();
+
+  while (i1 != v1.end() && i2 != v2.end())
```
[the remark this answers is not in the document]
vng: Left ToDo for detecting connections like a-b (important in U-turn for c), but now there are many false positives like this:
```
              a
              |
c - - c - c - - c
         |
         b
```
where c - c connection is not segregated routing from a - c - b.

src/extractor/extractor.cpp
```diff
@@ -808,5 +818,173 @@ void Extractor::WriteCompressedNodeBasedGraph(const std::string &path,
...
+    {
+      return std::binary_search(v.begin(), v.end(), current, EdgeInfo::LessName());
+    };
+
+    if (!findNameFn(v1) && !findNameFn(v2))
+      return false;
+  }
+
+  // set_intersection like routine to count equal name pairs.
+  std::vector<std::pair<EdgeInfo const *, EdgeInfo const *>> commons;
```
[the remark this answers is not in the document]
vng: Vectors have duplicates: std::set_intersection({a, a, b}, {a, a, c}) == {a}, but
we need {a, a}.

src/extractor/extractor.cpp
```diff
@@ -808,5 +818,173 @@ void Extractor::WriteCompressedNodeBasedGraph(const std::string &path,
...
+    for (auto const & e : edges1)
+    {
+      NodeID const target = graph->GetTarget(e);
+      if (target == node2)
+        continue;
+
+      auto const & data = graph->GetEdgeData(e);
+
+      BOOST_ASSERT(data.name_id != INVALID_NAMEID);
+      info.push_back({target, data.name_id, data.reversed ? 1 : 0, data.classes});
```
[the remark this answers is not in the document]
vng: Well, decided to store only needed data from EdgeData. Can be replaced to store EdgeData or even {EdgeData const *} to minimize memory using and copying.

src/extractor/extractor.cpp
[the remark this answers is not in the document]
MoKob: referencing the stl documentation: Constructs a sorted range beginning at d_first consisting of elements that are found in both sorted ranges [first1, last1) and [first2, last2). If some element is found m times in [first1, last1) and n times in [first2, last2), the first std::min(m, n)

src/extractor/extractor.cpp
vng: So you suggest remove this check, or ... ? I tried to run algorithm without this check and seems like it's ok, but sometimes we had situations like this:
```
a-b-b
|       |
|       |
a      b
|       |
|       |
a-b-b
```

a-a long way marked as segregated.

vng:
Will be good to get the coordinates of all this intersections for future analysis.
2 - looks like tile data is not equal to generation data.

MoKob:
@vng would it be possible to update the example server to reflect current locations in the URL? That would make specifying locations to look at a lot easier.

src/extractor/extractor.cpp
MoKob: No, I don't suggest getting rid of it completely. We need to ensure reasonable treatment, though. I would suggest considering the road-width of crossing roads here as well and to limit the length based on different scenarios.

E.g. if you have a segregated roat with 4 lanes per direction (not uncommon at intersections). You see 4 * 3.2 ~ 13 m per side. So (given that the ways are modelled in the middle of the road), we loose out on 13 meter for that threshold, limiting it to 20 meters. On small roads, we will allow larger differences. This is exactly opposite to what I would deem reasonable. Larger roads often have intermediate segments with traffic islands, while smaller roads have small traffic islands, if any. 
So this threshold should scale based on the local intersection and the involved roads, rather than be the fixed 30 meter from conservative post-processing.

vng:
Done.

MoKob:
I've done a new round of looking at the example server. A few observations:

 - We should (in the future) look at whether we can detect motorway crossings like http://88.99.58.38/segregated.html#16.21/33.74378/-84.38864

 - Cases to look at:
    - http://88.99.58.38/segregated.html#20/33.74458/-84.3981 seems to me from the map alone that these should all be marked?
    - http://88.99.58.38/segregated.html#18.99/33.75812/-84.43095 for a staggered intersection, this is REALLY far
    - same here: http://88.99.58.38/segregated.html#19.52/33.77011/-84.39195
    - http://88.99.58.38/segregated.html#19.43/33.77148/-84.39619 such a merge before an intersection might also be a candidate for internal edges
    - a mean one: http://88.99.58.38/segregated.html#18.79/33.77002/-84.38719 (so much Linden)
    - http://88.99.58.38/segregated.html#20.12/37.74719/-122.41047 the triangle should be marked, not the lower road
    - http://88.99.58.38/segregated.html#19.49/37.74218/-122.39519 similar to the above
    - missing inner segments: http://88.99.58.38/segregated.html#20.33/37.72026/-122.3966
    - http://88.99.58.38/segregated.html#19.39/37.72134/-122.39569 (entire 3rd street, maybe too wide?)
    - http://88.99.58.38/segregated.html#19.4/37.7092/-122.3927 another triangle

Let me know if you want to do a brainstorming session on how we could approach these / what might be the issues.

vng:
@MoKob Thanks, I will investigate your cases. Please, take a look at my last commit here:

https://github.com/Project-OSRM/osrm-backend/pull/4588/commits/9faa1e34b03a4748070b2d36aa22d6cb80e0dc03

How I introduced "segregated" flag and don't break all the current checks and test cases. The only possible way for that is making additional check after all previous guidance::supress/collapse calls.

include/engine/datafacade/datafacade_base.hpp
```diff
@@ -191,6 +191,8 @@ class BaseDataFacade
     virtual util::guidance::EntryClass GetEntryClass(const EdgeID turn_id) const = 0;
 
     virtual bool IsLeftHandDriving(const NodeID id) const = 0;
+
+    virtual bool IsSegregated(const NodeID) const { return false; }
```
MoKob: why not simply pure virtual?

include/engine/guidance/collapse_turns.hpp
```diff
@@ -140,6 +140,8 @@ void combineRouteSteps(RouteStep &step_at_turn_location,
 // alias for suppressing a step, using CombineRouteStep with NoModificationStrategy only
 void suppressStep(RouteStep &step_at_turn_location, RouteStep &step_after_turn_location);
 
+std::vector<RouteStep> suppressSegregated(std::vector<RouteStep> steps);
```
MoKob: I don't really agree with the naming here / use of a dedicated function. Collapse scenarios should make use of the segregated flag instead of checking for length values.

include/engine/guidance/route_step.hpp
```diff
@@ -60,6 +60,7 @@ inline IntermediateIntersection getInvalidIntersection()
 struct RouteStep
 {
     unsigned name_id;
+    bool is_segregated;
```
MoKob: would prefer this being located close to the maneuver, or even being made part of the maneuver. After all it is referring to the turn, not to the step

include/engine/guidance/route_step.hpp
```diff
@@ -162,6 +163,8 @@ inline RouteStep &RouteStep::ElongateBy(const RouteStep &following_step)
                          following_step.intersections.begin(),
                          following_step.intersections.end());
 
+    is_segregated = false;
```
MoKob: what about multi-step intersections. Are we ensuring that all scenarios will be processed in a single step? Shouldn't this be transferring the `is_segregated` flag from the following step?

include/engine/internal_route_result.hpp
```diff
@@ -27,6 +27,8 @@ struct PathData
     NodeID turn_via_node;
     // name of the street that leads to the turn
     unsigned name_id;
+    // segregated edge-based node that leads to the turn
+    bool is_segregated;
```
MoKob: the fact that we now have `is_segregated` in so many more items makes me think that the maneuver would probably be the best place to put it, since we are pulling through maneuvers anyhow.

include/engine/routing_algorithms/routing_base.hpp
```diff
@@ -164,6 +164,9 @@ void annotatePath(const FacadeT &facade,
         const auto turn_id = edge_data.turn_id; // edge-based graph edge index
         const auto node_id = *node_from;        // edge-based graph node index
         const auto name_index = facade.GetNameIndex(node_id);
+        const bool is_segregated = facade.IsSegregated(node_id);
+        if (is_segregated)
```
MoKob: debug output that needs to be removed

include/extractor/node_based_graph_factory.hpp
```diff
@@ -36,12 +36,13 @@ class NodeBasedGraphFactory
...
     // data optimisation and many other aspects. After this step, the edge-based graph factory can
     // turn the graph into the routing graph to be used with the navigation algorithms.
-    NodeBasedGraphFactory(const boost::filesystem::path &input_file,
-                          ScriptingEnvironment &scripting_environment,
-                          std::vector<TurnRestriction> &turn_restrictions,
-                          std::vector<ConditionalTurnRestriction> &conditional_turn_restrictions);
+    NodeBasedGraphFactory(const boost::filesystem::path &input_file);
 
-    auto const &GetGraph() const { return compressed_output_graph; }
+    void CompressAll(ScriptingEnvironment &scripting_environment,
```
MoKob: Any documentation on what this function does would be great, as we can now (as it seems) generate an incomplete state and need to compress the graph before we can process it in the edge-based graph factory.
I would like to avoid settings like these.

Is there a reason we cannot operate on the compressed graph? Any edge that will be compressed will not be segregated anyhow.

src/engine/guidance/collapse_turns.cpp
```diff
@@ -477,6 +478,64 @@ RouteSteps collapseTurnInstructions(RouteSteps steps)
     return steps;
 }
 
+std::vector<RouteStep> suppressSegregated(std::vector<RouteStep> steps)
```
MoKob: this entire function cannot be in here, if we want to merge this and continue working. The segregated flag should supplement turn compression, not do additional work on non-compressed turns.

src/engine/plugins/tile.cpp
```diff
@@ -182,6 +182,60 @@ inline void encodePoint(const FixedPoint &pt, protozero::packed_field_uint32 &ge
     geometry.add_element(protozero::encode_zigzag32(dy));
 }
 
+std::vector<FixedLine> coordinatesToTileLine(const std::vector<util::Coordinate> &points,
```
MoKob: I am confused that we wouldn't have logic for this, @danpat how are the other lines generated?

src/engine/plugins/tile.cpp
```diff
@@ -865,6 +934,66 @@ void encodeVectorTile(const DataFacadeBase &facade,
                 }
             }
         }
+
+        {
+            protozero::pbf_writer line_layer_writer(tile_writer, util::vector_tile::LAYER_TAG);
```
MoKob: wouldn't it be possible to simply add a tag to the already present lines? We add many attributes to lines, why not segregated as well?

src/extractor/extractor.cpp
```diff
@@ -212,10 +212,19 @@ int Extractor::run(ScriptingEnvironment &scripting_environment)
...
-    NodeBasedGraphFactory node_based_graph_factory(config.GetPath(".osrm"),
-                                                   scripting_environment,
-                                                   turn_restrictions,
-                                                   conditional_turn_restrictions);
+    NodeBasedGraphFactory node_based_graph_factory(config.GetPath(".osrm"));
+
+    util::Log() << "Find segregated edges in node-based graph ..." << std::flush;
+    TIMER_START(segregated);
+
+    const size_t segregated_count = FindSegregatedNodes(node_based_graph_factory);
```
MoKob: why on a non-compressed graph?

include/engine/datafacade/datafacade_base.hpp
vng: To minimize possible changes, but ok, will do pure.

include/engine/guidance/collapse_turns.hpp
vng: Please, clarify in more details. Maybe, the naming is not good, but this function checks for segregated flag, but not the length ... Using the dedicated function at this moment is to minimize changes in current code and don't break existing test logic.

include/engine/guidance/route_step.hpp
vng: It make sense if you mean StepManeuver which is a part of RouteStep, right?

include/engine/guidance/route_step.hpp
vng: Yes, this logic will be more complex in future.

include/engine/guidance/route_step.hpp
MoKob: Yes. I was referring to the StepManeuver

include/engine/internal_route_result.hpp
vng: But we need to pass the flag from the annotatePath to the assembleSteps. Did it with the same logic as name_id, classes, etc.

include/engine/routing_algorithms/routing_base.hpp
vng: Ok.

include/extractor/node_based_graph_factory.hpp
vng: Ok, will introduce segregated processing for compressed node-based graph as discussed.

include/engine/datafacade/datafacade_base.hpp
MoKob: We should aim to have this in both versions of the datafacade, so that we have a consistent state over all versions of osrm that can be run.

include/engine/guidance/route_step.hpp
```diff
@@ -162,6 +163,8 @@ inline RouteStep &RouteStep::ElongateBy(const RouteStep &following_step)
                          following_step.intersections.begin(),
                          following_step.intersections.end());
 
+    /// @todo Process is_segregated flag
```
MoKob: what is this todo referring to? If we want to go ahead with a merge, this should be made into an issue and referred to or removed.

include/engine/internal_route_result.hpp
MoKob: You are right. Right now this is consistent. The internal route-result shouldn't be concerned with this data, but for now it is consistent with the name and we can keep it as is. Since we are setting these kind of values for source/target in assemble-steps anyhow, I would like to remove them from InternalRouteResult at some time.

include/extractor/edge_based_node.hpp
```diff
@@ -13,6 +13,7 @@ struct EdgeBasedNode
     GeometryID geometry_id;
     ComponentID component_id;
     AnnotationID annotation_id;
+    bool segregated;
```
MoKob: we should see if we can use a single bit from the annotation-id for this, to not increase the size of an edge-based node.

include/extractor/extractor.hpp
```diff
@@ -100,6 +101,8 @@ class Extractor
     void WriteConditionalRestrictions(
         const std::string &path,
         std::vector<ConditionalTurnRestriction> &conditional_turn_restrictions);
+
+    size_t FindSegregatedNodes(NodeBasedGraphFactory &factory);
```
MoKob: some commentary here would be great to describe the idea of what we want to do here.

include/extractor/node_based_graph_factory.hpp
```diff
@@ -41,7 +41,7 @@ class NodeBasedGraphFactory
                           std::vector<TurnRestriction> &turn_restrictions,
                           std::vector<ConditionalTurnRestriction> &conditional_turn_restrictions);
 
-    auto const &GetGraph() const { return compressed_output_graph; }
+    auto &GetGraph() { return compressed_output_graph; }
```
MoKob: This should remain const, to not break consistency. From what I understand you should be fine adding this back in, after now operating on the compressed graph.

src/engine/plugins/tile.cpp
MoKob: Also: I am a bit confused about why our travis tests fail for systems in a more or less random distribution based on tile size.

src/extractor/extractor.cpp
```diff
@@ -826,5 +835,269 @@ void Extractor::WriteCompressedNodeBasedGraph(const std::string &path,
...
+    {
+        auto const findNameFn = [&current](std::vector<EdgeInfo> const &v) {
+            return std::binary_search(v.begin(), v.end(), current, EdgeInfo::LessName());
+        };
+
+        if (!findNameFn(v1) && !findNameFn(v2))
+            return false;
+    }
+
+    /*
```
MoKob: before a merge, this should probably be removed

src/extractor/extractor.cpp
```diff
@@ -826,5 +835,269 @@ void Extractor::WriteCompressedNodeBasedGraph(const std::string &path,
...
+
+    guidance::RoadPriorityClass::Enum road_priority_class;
+
+    struct LessName
+    {
+        bool operator()(EdgeInfo const &e1, EdgeInfo const &e2) const { return e1.name < e2.name; }
+    };
+};
+
+bool IsSegregated(std::vector<EdgeInfo> v1,
```
MoKob: Could we make all of these `xi` variables into something more descriptive? Personally, I find code far easier to follow if the name of a variable directly tells me what I am looking at.

src/extractor/extractor.cpp
```diff
@@ -826,5 +835,269 @@ void Extractor::WriteCompressedNodeBasedGraph(const std::string &path,
...
+    size_t equal_class_count = 0;
+    for (auto const &e : commons)
+        if (check_equal_class(e))
+            ++equal_class_count;
+
+    if (equal_class_count < 2)
+        return false;
+
+    auto const get_length_threshold = [](EdgeInfo const *e) {
+        switch (e->road_priority_class)
```
MoKob: should this simply be a constexpr lookup table based on the priority classes? Are these meters?

src/extractor/extractor.cpp
```diff
@@ -826,5 +835,269 @@ void Extractor::WriteCompressedNodeBasedGraph(const std::string &path,
...
+        case guidance::RoadPriorityClass::SECONDARY:
+        case guidance::RoadPriorityClass::TERTIARY:
+            return 5.0;
+        default:
+            return 2.5;
+        }
+    };
+
+    double threshold = std::numeric_limits<double>::max();
+    for (auto const &e : commons)
```
MoKob: could you add some commentary for all of these detectors to describe the reasoning/ideas behind them? Why are they chosen in that way, where do the constants come from?

src/extractor/extractor.cpp
```diff
@@ -826,5 +835,269 @@ void Extractor::WriteCompressedNodeBasedGraph(const std::string &path,
...
+    size_t segregated_count = 0;
+
+    for (NodeID sourceID = 0; sourceID < graph.GetNumberOfNodes(); ++sourceID)
+    {
+        auto const sourceEdges = graph.GetAdjacentEdgeRange(sourceID);
+        for (EdgeID edgeID : sourceEdges)
+        {
+            auto &edgeData = graph.GetEdgeData(edgeID);
+
+            if (edgeData.reversed || edgeData.segregated || !processed.insert(edgeID).second)
```
MoKob: have you found yourself processing the same edge-id twice? These should be unique over all Adjacent Edge Ranges

src/engine/plugins/tile.cpp
vng: I'm also confused about randomly failed Monaco tile size tests. Will try to investigate it.

include/extractor/node_based_graph_factory.hpp
vng: Need to have access to non-const EdgeData to set segregated flag.

MoKob: I kind of hate to have non-const access to the graph. This implies that we can do further modifications, and the graph itself should remain the same. Can we potentially do a Decorator pattern here to only add the data to the EdgeBasedNode container, instead of modifying the node-based graph?

MoKob: If you do want to keep it on the edges, the method should probably be called from within the node-based graph factory, to ensure that the graph is in a consistent state after the production step.

src/engine/plugins/tile.cpp
vng: The next coordinatesToTileLine function below, but for the single coordinates pair and one FixedLine as a result.

include/extractor/node_based_graph_factory.hpp
vng: OK, I can build the vector of segregated edge ids and pass it to the BuildEdgeExpandedGraph -> EdgeBasedGraphFactory. In this case we avoid segregated flag in node-based graph edges.

src/engine/plugins/tile.cpp
vng: Actually, I can add segregated flag to the present lines (sections). I thought that separate drawing layer will be a better logic choice rather than modifying current structure. What is your opinion?

include/extractor/edge_based_node.hpp
vng: Sounds good. At this time we have about 8000000 different annotations for the whole world. Should I do it or leave as is for now?
Also, I have thoughts to add segregated flag into Annotation struct, but the logic will be more complex.

src/engine/plugins/tile.cpp
vng: And the separate segregated layer is easier to make optional when retrieving tiles.

include/extractor/edge_based_node.hpp
oxidase: The `segregated` bit can be merged with `AnnotationID` to avoid wasting space of  31 padding bits.

include/extractor/node_based_graph_factory.hpp
oxidase: I am on Moritz's side to keep the node-based graph immutable to prevent possible problems with data races. Using a vector of segregated edges is fine for me. The guidance code has to be refactored later within scope of #4664

oxidase:
The segregated intersection looks good to me, but i am confused here, because don't know the original plans for fixing #4273. In the PR `facade.IsSegregated` flag is only stored in `RouteStep` but never used in response or collapsing logic. Also the test `Scenario: Segregated Intersection, Cross Belonging to Single Street` passes with and without c++ changes, so not clear what kind scenario it captures. If it is [the original scenario](http://map.project-osrm.org/?z=18&center=38.904735%2C-77.015963&loc=38.905140%2C-77.015748&loc=38.904347%2C-77.015941&hl=en&alt=0) then the expectation must be `depart,turn left (onto third ),arrive`

oxidase:
After checking detected segregated intersections, the results look good! There some issues with detection  here http://88.99.58.38/segregated.html#15.69/38.9543/-76.9418   and the detection algorithm must be adjusted to avoid cases like at http://88.99.58.38/segregated.html#19.91/38.95256/-76.94509 that look like staggered intersections, but must be treated as two different intersections for Gallatin St -> 42nd Ave -> Gallatin St.

oxidase:
Looks good to me! Checked locally with debug tiles

PR #4593, opened 2017-10-11
Run lcov before uploading to codecov

oxidase:
# Issue

In #4558 to cover the new code parameters of MLD partitioning for some tests were changed. Unfortunately, this change does not really cover the changeset, but it was not shown in a coverage difference report, that was green.

In this PR:
* explicit `lcov` call
* optimization level changed from `-Og` to `-O0` for coverage builds
* removed `@mld` and `@ch` tags filtering 
* scenario `Testbot - Multi level routing`uses only  2 levels and additional checks 1-to-many, many-to-1, few-to-many and many-to-few
* updated gdb printer to show mld cell ids as background colors

https://codecov.io/gh/Project-OSRM/osrm-backend/src/698de86bec08e122111994ad8554ea6487621167/src/engine/routing_algorithms/many_to_many_mld.cpp#L48 shows that the MLD pruning is covered now.

/cc @MoKob 
 

## Tasklist
 - [x] review
 - [x] adjust for comments

## Requirements / Relations


include/partition/multi_level_partition.hpp
```diff
@@ -115,7 +115,7 @@ template <storage::Ownership Ownership> class MultiLevelPartitionImpl final
 
     std::uint32_t GetNumberOfCells(LevelID level) const
     {
-        return GetCell(level, GetSenitileNode());
+        return GetCell(level, GetSentinelNode());
```
TheMarex: 🙈 thanks!

oxidase:
@TheMarex +3.8% comes from change `-Og` -> `-O0` optimization level. gcov just ignored some optimized simple returns.

PR #4594, opened 2017-10-11
Fix potential source of segfault removing the use of deallocating iterators

TheMarex:
# Issue

`osrm-contract` seems to generate segfaults when merging edges. A likely source of this is the `DeallocatingVector::dbegin()` use in the merging of contracted edges.

Removing the usage of `DeallocatingVector` increases the memory usage by about `3%` for `osrm-contract` due to the additional space needed to maintain the sorting after merging (`permuation` and `ordering` vector).

## Tasklist
 - [x] review
 - [x] adjust for comments


TheMarex:
>  would it make sense to switch to std::vectorin other places?

We need to measure this I think. For example we currently use it in `DyanmicGraph` to avoid copying the whole `edge` vector on relocation.

What we should remove is the `dbegin()`, `dend()` functions, I'll add that to this PR.

Issue #4584, opened 2017-10-08
OSRM contract running for a long time 


wieringen:
I have been running osrm router for over 1.5 years. Every night a cron runs that fetches the geofabrik 
 osm file for the netherlands. I filter out the bicylce networks using osmfilter and feed the result to osrm. Osrm is using the bicycle profile.
Today I updated osrm from 5.6.0 to 5.12.0. What I noticed was a large increase in parsing time. Is this normal?
![cpu usage](http://baijs.com/osrm.png) 

*** Update 20:30 ***
I just killed the process. What normally takes +/- 2 hours is now taking more than 8 and it was still running.

wieringen:
I just killed the process. What normally takes +/- 2 hours is now taking more than 8 and is still running.

TheMarex:
> What I noticed was a large increase in parsing time

No that seems weird. Are you using the default bicycle profile or something custom?

wieringen:
I changed some things:

20c20,21
<       weight_name                   = 'duration',
---
>       --weight_name                   = 'duration',
>       weight_name                   = 'distance',
40a42,43
>       'swing_gate',
>       'bus_trap',
42d44
<       'bus_trap',
50a53
>       'lift_gate',
51a55
>       'yes',
56a61,62
>       'no',
>       'private',
62,63d67
<       'no',
<       'private',
65d68
<       'forestry',
467d469
< 
539c541
<     
---
> 
571a577,582
>   -- for distance based routing we don't want to have penalties based on turn angle
>   if profile.properties.weight_name == 'distance' then
>      turn.weight = 0
>   else
>      turn.weight = turn.duration
>   end

wieringen:
Oh and I'm also using the official docker image: osrm/osrm-backend where I enabled stxxl by creating a file called .stxxl in /opt. Is this still correct?

.stxxl contents:
```
disk=/var/tmp/stxxl,7G,syscall
```

tds4u:
@wieringen: Did you try this? Did this help?

> --weight_name                = 'duration',
 weight_name                   = 'distance',

> if profile.properties.weight_name == 'distance' then
    turn.weight = 0
 else
    turn.weight = turn.duration
 end

wieringen:
Oops sorry maybe I wasn't clear. I have always used those modification in my profile. So I ported them straight away after the upgrade to the v2 profile. Maybe I can try it again with a default profile..

wieringen:
[...]

PR #4596, opened 2017-10-12
Documentation for Multi-Level Dijkstra Pipeline

daniel-j-h:
In https://github.com/Project-OSRM/osrm-backend/issues/4584#issuecomment-335927807 a user was asking for docs about CH vs. MLD. Turns out we never documented the MLD pipeline in our readme or in the wiki. Oops.

This changeset switches our docs to MLD by default: it's production-ready by now and we want users to use the MLD pipeline (except maybe for very large distance matrices cc @jcoupey)

Preview: https://github.com/Project-OSRM/osrm-backend/tree/mld-docs#quick-start

- [x] switch readme to mld
- [x] update https://github.com/Project-OSRM/osrm-backend/wiki/Running-OSRM

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,6 +41,14 @@ Related [Project-OSRM](https://github.com/Project-OSRM) repositories:
 
 The easiest and quickest way to setup your own routing engine is to use Docker images we provide.
 
+There are two pre-processing pipelines available:
+- Contraction Hierarchies (CH)
+- Multi-Level Dijkstra (MLD)
+
+we recommend using MLD by default except for special use-cases such as very large distance matrices where CH is still a better fit for the time being.
+In the following we explain the MLD pipeline.
+If you want to use the CH pipeline instead replace `osrm-partition` and `osrm-customize` with a single `osrm-customize` and change the algorithm option for `osrm-routed`.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -41,6 +41,14 @@ Related [Project-OSRM](https://github.com/Project-OSRM) repositories:
 
 The easiest and quickest way to setup your own routing engine is to use Docker images we provide.
 
+There are two pre-processing pipelines available:
+- Contraction Hierarchies (CH)
+- Multi-Level Dijkstra (MLD)
+
+we recommend using MLD by default except for special use-cases such as very large distance matrices where CH is still a better fit for the time being.
+In the following we explain the MLD pipeline.
+If you want to use the CH pipeline instead replace `osrm-partition` and `osrm-customize` with a single `osrm-customize` and change the algorithm option for `osrm-routed`.
```
MoKob: shouldn't this be `replace with osrm-contract`?

MoKob: Also, should we be explicit about the algorithm option here?

daniel-j-h: Urgh good catch; all done.

Issue #4444, opened 2017-08-25
Deprecate CoreCH algorithm


TheMarex:
Since we now have feature parity with MLD, there is no reason to keep CoreCH around - it is slower and has less features then MLD.

Removing CoreCH will let us remove the following flags for `osrm-contract`:

- `core-factor`
- `use-level-cache`
- `.level` file
- `.core` file

We might be able to remove the code for handling `CoreCH` by automatically falling back to a plain `CH` (and printing warnings).

danpat:
Some of these changes already happened for the 5.12 release - moving this to the 5.13 release so the deprecation can be deliberate.

Issue #4482, opened 2017-09-06
osrm-contract still takes `--level-cache` argument, but it is not functional


danpat:
Adding support for excluding certain edge classes during queries in #4427 dropped support for the `osrm-contract` level cache feature, but did not remove the command-line option.  We should do that, it's currently a confusing no-op.

karenzshea:
Done in https://github.com/Project-OSRM/osrm-backend/pull/4601

PR #4598, opened 2017-10-12
Deprecate CoreCH, fallback to using CH

karenzshea:
# Issue

Closes https://github.com/Project-OSRM/osrm-backend/issues/4444 and https://github.com/Project-OSRM/osrm-backend/issues/4482

## Tasklist
 - [ ] add changelog entry
 - [ ] add regression / cucumber cases (see docs/testing.md) how to test this??
 - [ ] review
 - [ ] adjust for comments

## Requirements / Relations
cc @TheMarex @oxidase


src/engine/routing_algorithms/direct_shortest_path.cpp
```diff
@@ -65,11 +65,6 @@ InternalRouteResult directShortestPathSearch(SearchEngineData<Algorithm> &engine
 }
 
 template InternalRouteResult
-directShortestPathSearch(SearchEngineData<corech::Algorithm> &engine_working_data,
-                         const DataFacade<corech::Algorithm> &facade,
-                         const PhantomNodes &phantom_nodes);
-
-template InternalRouteResult
 directShortestPathSearch(SearchEngineData<ch::Algorithm> &engine_working_data,
```
oxidase: can also be removed but requires a definition change above to
```
template <>
InternalRouteResult directShortestPathSearch(SearchEngineData<ch::Algorithm> &engine_working_data,
                                             const DataFacade<ch::Algorithm> &facade,
                                             const PhantomNodes &phantom_nodes)
```

src/tools/contract.cpp
```diff
@@ -49,7 +49,7 @@ return_code parseArguments(int argc,
         "Number of threads to use")(
         "core,k",
         boost::program_options::value<double>(&contractor_config.core_factor)->default_value(1.0),
-        "Percentage of the graph (in vertices) to contract [0..1]")(
+        "Percentage of the graph (in vertices) to contract [0..1]. Will always be 1.0")(
```
oxidase: would  be `DEPRECATED Percentage of the graph (in vertices) to contract [0..1]` better?

include/contractor/contractor_config.hpp
```diff
@@ -65,6 +65,7 @@ struct ContractorConfig final : storage::IOConfig
 
     unsigned requested_num_threads;
 
+    // Deprecated
```
oxidase: Please could you add also a release version when it is planed to remove, like `// DEPRECATED to be removed in v6.0`

oxidase:
as per chat with @MoKob: all internal usage of `core_factor` can be removed 
* https://github.com/Project-OSRM/osrm-backend/blob/e647547adde8239ab637d1a443263562e1e2d34f/src/contractor/graph_contractor.cpp#L564 and all dependencies removed or adjusted using `core_factor=1`

include/contractor/contractor_config.hpp
karenzshea: I'll open a follow up issue to track removal in v6 too

src/tools/contract.cpp
karenzshea: done in https://github.com/Project-OSRM/osrm-backend/pull/4601/files#diff-097d6382cc3c7256f51a139534b8c76dR52

PR #4601, opened 2017-10-12
Deprecate CoreCH, fallback to using CH

karenzshea:
# Issue

Closes https://github.com/Project-OSRM/osrm-backend/issues/4444 and https://github.com/Project-OSRM/osrm-backend/issues/4482

## Tasklist
 - [ ] add changelog entry
 - [x] add regression / cucumber cases (see docs/testing.md) how to test this??
 - [x] review
 - [x] adjust for comments
 - [ ] Open follow up ticket for fully removing CoreCH options in v6.0
 - [x] Remove core_factor option from exclude graph contraction 

## Requirements / Relations
cc @TheMarex @oxidase sorry for the new pr/noise, I renamed the branch because it was very wrong.


unit_tests/library/tile.cpp
```diff
@@ -846,15 +820,6 @@ BOOST_AUTO_TEST_CASE(test_tile_nodes_ch)
     test_tile_nodes(osrm);
 }
 
-BOOST_AUTO_TEST_CASE(test_tile_nodes_corech)
```
TheMarex: We still need to keep around these tests until we remove the fallback.

unit_tests/library/algorithm.cpp
```diff
@@ -15,14 +15,6 @@ BOOST_AUTO_TEST_CASE(test_incompatible_with_mld)
         osrm::exception);
 }
 
-BOOST_AUTO_TEST_CASE(test_incompatible_with_corech)
```
TheMarex: Same here we still need this test to make sure the fallback works.

src/osrm/osrm.cpp
```diff
@@ -44,26 +43,12 @@ OSRM::OSRM(engine::EngineConfig &config)
     // Now, check that the algorithm requested can be used with the data
     // that's available.
 
-    if (config.algorithm == EngineConfig::Algorithm::CoreCH ||
-        config.algorithm == EngineConfig::Algorithm::CH)
+    if (config.algorithm == EngineConfig::Algorithm::CH)
```
TheMarex: This also needs to check for `config.algorithm == EngineConfig::Algorithm::CoreCH` since we can't remove the `CoreCH` value yet. Doing so we be a breaking change of the library interface.

include/engine/routing_algorithms/direct_shortest_path.hpp
```diff
@@ -15,11 +15,11 @@ namespace engine
 namespace routing_algorithms
 {
 
-/// This is a striped down version of the general shortest path algorithm.
+/// This is a stripped down version of the general shortest path algorithm.
```
TheMarex: 🙈

src/tools/contract.cpp
```diff
@@ -49,21 +49,17 @@ return_code parseArguments(int argc,
...
+                   boost::program_options::value<std::vector<std::string>>(
+                       &contractor_config.updater_config.segment_speed_lookup_paths)
+                       ->composing(),
+                   "Lookup files containing nodeA, nodeB, speed data to adjust edge weights")(
         "turn-penalty-file",
         boost::program_options::value<std::vector<std::string>>(
             &contractor_config.updater_config.turn_penalty_lookup_paths)
             ->composing(),
         "Lookup files containing from_, to_, via_nodes, and turn penalties to adjust turn weights")(
-        "level-cache,o",
```
TheMarex: This flag also need a deprecation warning, no hard removal.

unit_tests/library/algorithm.cpp
karenzshea: Added this test back, but renamed it.

PR #4604, opened 2017-10-12
NodeJS version of osrm-routed

danpat:
# Issue

This PR implements an experimental port of `osrm-routed` to NodeJS using `express` as the URL router.

Currently a WIP - all tests pass, but a few error messages have changed, and working on it has shown that our error scenario coverage is not really very complete.

 - [ ] Fix trip plugin, or delete bad trip test - one test expects to set `source=last`, which according to our docs, is not valid
 - [ ] Discuss error messages - because URL parser is not boost, getting char offset messages is much harder - are we tied to the old error strings?
 - [ ] Result precision - there are some rounding changes in some values - should these be fixed, or are they ok to slip through?
 - [ ] Tidy up/refactor lib/index.js changes
 - [ ] Support all `osrm-routed` command-line options and generate help message
 - [ ] Blocked by https://github.com/Project-OSRM/osrm-backend/issues/4690 (polyline URL support)
- [ ] Support max search radius for a map-matching requests. Issue is [here](https://github.com/Project-OSRM/osrm-backend/issues/4719) 

## Tasklist
 - [ ] ADD OWN TASKS HERE
 - [ ] review
 - [ ] adjust for comments

### DIFF
--- a/features/lib/osrm_loader.js
+++ b/features/lib/osrm_loader.js
@@ -47,7 +47,7 @@ class OSRMBaseLoader{
           if (err) {
             if (retryCount < 10) {
               retryCount++;
-              setTimeout(() => { tryConnect(this.scope.OSRM_PORT, retry); }, 10);
+              setTimeout(() => { tryConnect(this.scope.OSRM_PORT, retry); }, 100);
             } else {
               callback(new Error("Could not connect to osrm-routed after ten retries."));
             }
@@ -75,13 +75,13 @@ class OSRMDirectLoader extends OSRMBaseLoader {
     }
 
     osrmUp (callback) {
-        if (this.osrmIsRunning()) return callback(new Error("osrm-routed already running!"));
+        if (this.osrmIsRunning()) return callback(new Error("osrm-routed.js already running!"));
 
         const command_arguments = util.format('%s -p %d -a %s', this.inputFile, this.scope.OSRM_PORT, this.scope.ROUTING_ALGORITHM);
-        this.child = this.scope.runBin('osrm-routed', command_arguments, this.scope.environment, (err) => {
+        this.child = this.scope.runBin('osrm-routed.js', command_arguments, this.scope.environment, (err) => {
             if (err && err.signal !== 'SIGINT') {
                 this.child = null;
-                throw new Error(util.format('osrm-routed %s: %s', errorReason(err), err.cmd));
+                throw new Error(util.format('osrm-routed.js %s: %s', errorReason(err), err.cmd));
             }
         });
         callback();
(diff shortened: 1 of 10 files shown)


emiltin:
good idea, i think many people miss this  part of the tool chain

features/testbot/status.feature
```diff
@@ -52,22 +52,22 @@ Feature: Status messages
...
+            | route/v1/driving/1                  | 400    | At least two coordinates must be provided |
+            | route/v1/driving/1,1                | 400    | At least two coordinates must be provided |
+            | route/v1/driving/1,1,1              | 400    | At least two coordinates must be provided |
+            | route/v1/driving/x                  | 400    | At least two coordinates must be provided |
+            | route/v1/driving/x,y                | 400    | At least two coordinates must be provided |
+            | route/v1/driving/1,1;               | 400    | Coordinates must be an array of (lon/lat) pairs |
+            | route/v1/driving/1,1;1              | 400    | Coordinates must be an array of (lon/lat) pairs |
+            | route/v1/driving/1,1;1,1,1          | 400    | Coordinates must be an array of (lon/lat) pairs |
+            | route/v1/driving/1,1;x              | 400    | Coordinates must be an array of (lon/lat) pairs |
+            | route/v1/driving/1,1;x,y            | 400    | Lng/Lat coordinates must be valid numbers |
```
daniel-j-h: Why the changes here?

features/lib/osrm_loader.js
```diff
@@ -47,7 +47,7 @@ class OSRMBaseLoader{
           if (err) {
             if (retryCount < 10) {
               retryCount++;
-              setTimeout(() => { tryConnect(this.scope.OSRM_PORT, retry); }, 10);
+              setTimeout(() => { tryConnect(this.scope.OSRM_PORT, retry); }, 100);
```
daniel-j-h: Why the change here? Is 10s no longer enough?

features/testbot/trip.feature
```diff
@@ -96,6 +96,7 @@ Feature: Basic trip planning
             | waypoints               | source | trips         |
             | a,b,c,d,e,f,g,h,i,j,k,l | first  | alkjihgfedcba |
 
+    @todo
```
daniel-j-h: Why disabling this test?

include/nodejs/node_osrm_support.hpp
```diff
@@ -65,7 +65,23 @@ inline void ParseResult(const osrm::Status &result_status, osrm::json::Object &r
...
+        {
+            // Concatenate the code: message so we don't lose details on errors
+            const auto code = code_iter->second.get<osrm::json::String>().value.c_str();
+            const auto message = message_iter->second.get<osrm::json::String>().value.c_str();
+            auto buffer =
+                std::make_unique<char[]>(std::strlen(code) + 2 + std::strlen(message) + 1);
+            std::strcpy(buffer.get(), code);
+            std::strcat(buffer.get(), ": ");
+            std::strcat(buffer.get(), message);
+            throw std::logic_error(buffer.get());
```
daniel-j-h: Please change the last couple of lines to

```c++
const auto code = .. .value;
const auto message = .. .value;
// both are of type std::string now, see https://github.com/Project-OSRM/osrm-backend/blob/031ce72db1bdb724527096703c55741a7c7f572c/include/util/json_container.hpp#L81
throw std::logic_error{code + ": " + message}
```

include/nodejs/node_osrm_support.hpp
```diff
@@ -65,7 +65,23 @@ inline void ParseResult(const osrm::Status &result_status, osrm::json::Object &r
...
+        }
+        else
+        {
+            throw std::logic_error(code_iter->second.get<osrm::json::String>().value.c_str());
```
daniel-j-h: logic_error should have a ctor taking a std::string; the c_str is not needed here

TheMarex:
Not really a big priority anymore given that docker kind of won. Also turns out going through the V8 serialization layer is not for free, so the C++ based HTTP server is actually quite a bit faster.

Issue #4606, opened 2017-10-13
Invalid trigger of collapse scenario - No UTurn


MoKob:
<img width="775" alt="screen shot 2017-10-13 at 16 10 37" src="https://user-images.githubusercontent.com/12932279/31550289-26285d84-b031-11e7-9021-d65d0d09e197.png">

At [this location](https://www.openstreetmap.org/#map=18/53.89755/27.54306) we can see an invalidly triggering collapse into a u-turn.

PR #4608, opened 2017-10-13
Fix/4606

MoKob:
# Issue

Resolves https://github.com/Project-OSRM/osrm-backend/issues/4606

## Tasklist

 - [x] review
 - [ ] adjust for comments


vng:
But how does the test check this case? The code is about turn angels and test case has all right angels, no? And what if h-d link is absent?

MoKob:
@vng the test is about combination of turn angles. If you see the sequence of commits, the first commit offers a different test (which is succeeding at that point)

Issue #4612, opened 2017-10-16
Refactor library interface


TheMarex:
In preparation for a new API version next year we should clean-up our current libosrm interface. The major problem right now is that we expose JSON objects on our C++ interface. This has the downside that it is slow to serialize (lots of `std::unordered_maps` and `std::vector` being created) and also slow to convert to v8 objects since we pay the serialization overhead twice: `Internal C++` -> `JSON` -> `v8 Objects`.

Since we need to retain our current interface for API compatibility I propose the following changes:

1. Refactor code so that we can expose the internal `C++` representation of our API response.
2. Add transformation layer that converts `C++` code to JSON representation (basically what we have with the `json_factory.hpp` right now)
3. Add new libosrm interface that exposes the new `C++` representation long-side the JSON one: `osrm.hpp` + `osrm_v1.hpp` (json) + `osrm_v2.hpp` (C++).
4. Add new `v8_factory.hpp` that can transform `C++` objects to `v8 Objects` directly and make use of it in the node bindings.

PR #4614, opened 2017-10-17
Refactor JSON generation for nearest plugin

TheMarex:
# Issue

This PR is the firs step to doing refactors to implement #4612. As example I ported the `nearest` plugin because it is very atomic and has a simple API.

Important goals of this refactor are:
- Contain JSON-specific part to `json_factory.cpp` / `toJSON` function.
- Add a better way of transporting errors without relying on the JSON object structure (`MaybeResult`)

## Tasklist
- [x] PR refactor for nearest plugin
- [x] PR refactor for table plugin
- [ ] PR refactor for match plugin
- [ ] PR refactor for trip plugin
- [ ] PR refactor for viaroute plugin
- [ ] PR refactor for tile plugin ?
- [ ] Template error types
 - [ ] review
 - [ ] adjust for comments

### DIFF
--- a/include/engine/api/nearest_result.hpp
+++ b/include/engine/api/nearest_result.hpp
@@ -16,7 +16,6 @@ struct NearestResult
 {
     std::vector<Waypoint> waypoints;
 };
-
 }
 }
 }
--- a/include/engine/api/nearest_result.hpp
+++ b/include/engine/api/nearest_result.hpp
@@ -1,0 +1,24 @@
+#ifndef OSRM_ENGINE_API_NEAREST_RESULT_HPP
+#define OSRM_ENGINE_API_NEAREST_RESULT_HPP
+
+#include "engine/api/waypoint.hpp"
+
+#include <vector>
+
+namespace osrm
+{
+namespace engine
+{
+namespace api
+{
+
+struct NearestResult
+{
+    std::vector<Waypoint> waypoints;
+};
+
+}
+}
+}
+
+#endif
--- a/include/engine/api/waypoint.hpp
+++ b/include/engine/api/waypoint.hpp
@@ -1,0 +1,26 @@
+#ifndef OSRM_ENGINE_API_WAYPOINT_HPP
+#define OSRM_ENGINE_API_WAYPOINT_HPP
+
+#include "util/coordinate.hpp"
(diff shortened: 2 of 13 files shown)


include/engine/api/nearest_result.hpp
```diff
@@ -0,0 +1,24 @@
...
+{
+namespace engine
+{
+namespace api
+{
+
+struct NearestResult
+{
+    std::vector<Waypoint> waypoints;
+};
```
daniel-j-h: Because this will become the new interface for libosrm should we (this time) abstract away impl. details behind a proper class we can easily extend down the line? (Not specifically for the Nearest plugin more a general comment)

include/engine/api/waypoint.hpp
```diff
@@ -0,0 +1,22 @@
...
+namespace osrm
+{
+namespace engine
+{
+namespace api
+{
+struct Waypoint
+{
+    double distance;
+    const char *name;
```
daniel-j-h: Why not a string? Or if you want zero-copy onto our datafacades why not a string view/ref?

include/engine/error.hpp
```diff
@@ -0,0 +1,96 @@
...
+
+    operator bool() const {
+        return error.code == ErrorCode::NO_ERROR;
+    }
+
+    operator const Error &() const {
+        return error;
+    }
+
+    operator const ResultT &() const {
```
daniel-j-h: Should these operators be explicit or do you want implicit conversion for all types?

include/engine/error.hpp
```diff
@@ -0,0 +1,96 @@
...
+    std::string message;
+
+    auto throwException() const
+    {
+        throw util::exception(std::string(codeToString(code)) + ":" + message);
+    }
+};
+
+template<typename ResultT>
+struct MaybeResult
```
daniel-j-h: Should the error type be a template and then should the generic Error type above be split e.g. into

    TripError
    MatchError

etc.? As is now we can assign semantically meaningless error codes to plugin responses (e.g. NoTrip to the MatchPlugin response).

TheMarex:
Outdated, this was an attempt at isolating the C++ interface from the JSON response, but the library layer is not really used that much anyway.

Issue #4620, opened 2017-10-17
osrm-extract --parse-conditional-restrictions=0 flag still generates restrictions


mloskot:
I'm using OSRM 5.12.0

In https://github.com/Project-OSRM/osrm-backend/pull/3841#issuecomment-337261750 I asked for clarification about relation between extract and contract steps, and requirement of GeoJSON file with timezones.

@karenzshea explained in https://github.com/Project-OSRM/osrm-backend/pull/3841#issuecomment-337268288 the GeoJSON file is optional and it is required only if `.osrm.restrictions` file is present.
Then, to disable the conditional turn restriction application by running `osrm-extract` with `--parse-conditional-restrictions`  option.

There are a few issues:

1. GeoJSON file is *not* really optional. It is required indirectly though, by presence of some intermediate file, which is required by default now.

2. AFAICT, use of the suggested option has no effect and `.osrm.restrictions` file is still generated that makes the contraction step require GeoJSON file with timezones.

3. The suggested alternative solution by removing the `.osrm.restrictions` file is not a solution, but a hack, especially from OSRM API usage point, not really helpful..

4. Apparently, `--parse-conditional-restrictions` controls corresponding flag in `ExtractorConfig`:

```
osrm::ExtractorConfig config;
config.parse_conditionals = false;
...
osrm::extract(config);
```

and, obviously, that does not have any effect either, as in 2.

Any chance to get that situation clarified?

karenzshea:
I revisited the code handling conditional restrictions and sorry for being misleading in my hasty previous answer.

Thanks for catching that conditional restriction handling is not disabled by default in the extract and contract configurations.

Can you try using the changes in this branch https://github.com/Project-OSRM/osrm-backend/tree/conditionals/disable and let me know if it yields expected behaviour? I've disabled the relevant options by default, and also (bonus) made the `.osrm.restrictions` file optional as I'd previously thought.

PR #4622, opened 2017-10-18
Disable conditional restriction handling by default in configs

karenzshea:
# Issue

Closes https://github.com/Project-OSRM/osrm-backend/issues/4620

## Tasklist
 - [x] review
 - [x] adjust for comments

### DIFF
--- a/include/updater/updater_config.hpp
+++ b/include/updater/updater_config.hpp
@@ -44,22 +44,20 @@ namespace updater
 struct UpdaterConfig final : storage::IOConfig
 {
     UpdaterConfig()
-        : IOConfig(
-              {
-                  ".osrm.ebg",
-                  ".osrm.turn_weight_penalties",
-                  ".osrm.turn_duration_penalties",
-                  ".osrm.turn_penalties_index",
-                  ".osrm.nbg_nodes",
-                  ".osrm.ebg_nodes",
-                  ".osrm.edges",
-                  ".osrm.geometry",
-                  ".osrm.fileIndex",
-                  ".osrm.properties",
-                  ".osrm.restrictions",
-              },
-              {},
-              {".osrm.datasource_names"})
+        : IOConfig({".osrm.ebg",
+                    ".osrm.turn_weight_penalties",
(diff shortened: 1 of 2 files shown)


TheMarex:
This needs a rebase after we merge https://github.com/Project-OSRM/osrm-backend/pull/4623

src/extractor/edge_based_graph_factory.cpp
```diff
@@ -975,18 +975,24 @@ void EdgeBasedGraphFactory::GenerateEdgeExpandedEdges(
...
     auto const indexed_conditionals = IndexConditionals(std::move(conditionals));
+
     {
-        util::Log() << "Writing " << indexed_conditionals.size()
-                    << " conditional turn penalties...";
-        // write conditional turn penalties into the restrictions file
-        storage::io::FileWriter writer(conditional_penalties_filename,
-                                       storage::io::FileWriter::GenerateFingerprint);
-        extractor::serialization::write(writer, indexed_conditionals);
+        if (!indexed_conditionals.empty())
```
TheMarex: The reason why we write out en empty file here is that otherwise the loading code will fail: https://travis-ci.org/Project-OSRM/osrm-backend/jobs/289928057#L2508

The `.restrictions` file can be empty but is not optional.

include/updater/updater_config.hpp
```diff
@@ -44,22 +44,19 @@ namespace updater
...
+                    ".osrm.turn_weight_penalties",
+                    ".osrm.turn_duration_penalties",
+                    ".osrm.turn_penalties_index",
+                    ".osrm.nbg_nodes",
+                    ".osrm.ebg_nodes",
+                    ".osrm.edges",
+                    ".osrm.geometry",
+                    ".osrm.fileIndex",
+                    ".osrm.properties"},
+                   {".osrm.restrictions"},
```
TheMarex: `.restrictions` is actually not optional, but can be empty.

daniel-j-h:
From reading the ticket / conversation here it looks like this is a bug fix which should go into master and a release asap? What needs to be done here?

PR #4623, opened 2017-10-18
Replace use of implicit_value since that broke with boost 1.65

TheMarex:
# Issue

Fixes a problem with 5.13 that cause `osrm-extract` to throw errors even though parameters were correct before boost `1.65`.

Downside is that options that used to take `true/false` now throw an error if you provide them explicitly. (which is odd to begin with)

EDIT: Seems like we actually did that in the tests.

## Tasklist
 - [x] review
 - [x] adjust for comments

### DIFF
--- a/src/tools/extract.cpp
+++ b/src/tools/extract.cpp
@@ -67,10 +67,10 @@ return_code parseArguments(int argc,
                                   &extractor_config.location_dependent_data_paths)
                                   ->composing(),
                               "GeoJSON files with location-dependent data")(
-        "use-locations-cache",
+        "disable-locations-cache",
         boost::program_options::bool_switch(&extractor_config.use_locations_cache)
-            ->implicit_value(true)
-            ->default_value(extractor_config.use_locations_cache),
+            ->implicit_value(false)
+            ->default_value(true),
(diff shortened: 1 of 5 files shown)


src/tools/extract.cpp
```diff
@@ -67,10 +67,10 @@ return_code parseArguments(int argc,
                                   &extractor_config.location_dependent_data_paths)
                                   ->composing(),
                               "GeoJSON files with location-dependent data")(
-        "use-locations-cache",
-        boost::program_options::value<bool>(&extractor_config.use_locations_cache)
-            ->implicit_value(true)
-            ->default_value(extractor_config.use_locations_cache),
+        "disable-locations-cache",
```
danpat: Minor semantics: If we need to change this anyway, can we rename it to `disable-location-cache` (drop the `s` on `locations`)? (would also need to update changelog).

Reason: noun adjuncts in English are *usually* singular, unless there is a lexical restrictions (https://en.wikipedia.org/wiki/Noun_adjunct#English).

TheMarex:
@danpat can you check again if this works for you?

danpat:
My grammar geek has been sated.  Looks good!

PR #4624, opened 2017-10-19
Adds exception for barrier=lift_gate (bike, walk), resolves #4490

### LINKED ISSUE #4490: lift_gate missing from bicycle and foot barrier whitelist
a user reported this route not working, because 'lift_gate' is missing from the barrier whitelist:

https://ibikecph.dk/en#!/xxjk1.5jdq2/xxudm.5j1jd/mode:fast

looking at http://wiki.openstreetmap.org/wiki/Tag:barrier%3Dlift_gate, i'm a bit unsure about whether such gates can be assumed to be passable?

this old PR #1256 added lift_gate to the car profile. should probably be added to both bicycle and foot profiles. there's also an open issue for expanding the whitelist for the foot profile, which also mentions lift_gate: #3890

daniel-j-h:
Let's lift that gate.

### DIFF
--- a/features/bicycle/barrier.feature
+++ b/features/bicycle/barrier.feature
@@ -10,6 +10,7 @@ Feature: Barriers
             |                | x     |
             | bollard        | x     |
             | gate           | x     |
+            | lift_gate      | x     |
             | cycle_barrier  | x     |
             | cattle_grid    | x     |
             | border_control | x     |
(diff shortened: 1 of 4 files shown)


emiltin:
a lift_gate can probably be passed on bicycle/foot, even when it's closed. so i think it's ok to merge this.

it would be nice if barriers like this could incur a delay.

Issue #4585, opened 2017-10-09
Hash Annotation Data


MoKob:
Many road segments in OSM offer the same meta data from the viewpoint of OSRM, since we do not store all data entries that are possible in OSM and require a way to be split.

With https://github.com/Project-OSRM/osrm-backend/pull/4556 we introduced a meta-data store for all edge-based nodes / node-based edges. This store is currently populated based on OSM ways. The moment a way is split into multiple segments, we end up storing the same meta-data multiple times, increasing memory usage.

For example https://www.openstreetmap.org/way/231285502, https://www.openstreetmap.org/way/320630988, and https://www.openstreetmap.org/way/320630989 offer the same annotation.

We can reduce our annotation memory footprint by ensuring that we store unique sets of annotations only once.

TheMarex:
Implemented by #4585, saved about 1% total memory consumption of `osrm-routed`.

PR #4629, opened 2017-10-20
Fix issue #4585

deniskoronchik:
# Issue

#4585 

## Tasklist
 - [x] review
 - [x] adjust for comments

### DIFF
--- a/include/extractor/node_based_edge.hpp
+++ b/include/extractor/node_based_edge.hpp
@@ -71,6 +71,16 @@ struct NodeBasedEdgeAnnotation
             std::tie(name_id, classes, travel_mode, is_left_hand_driving) ==
             std::tie(other.name_id, other.classes, other.travel_mode, other.is_left_hand_driving));
     }
+
+    bool operator<(const NodeBasedEdgeAnnotation &other) const
+    {
+        return (std::tie(name_id, lane_description_id, classes, travel_mode, is_left_hand_driving) <
+                std::tie(other.name_id,
+                         other.lane_description_id,
+                         other.classes,
+                         other.travel_mode,
+                         other.is_left_hand_driving));
+    }
 };
 
 struct NodeBasedEdge
--- a/src/extractor/node_based_graph_factory.cpp
+++ b/src/extractor/node_based_graph_factory.cpp
@@ -4,8 +4,11 @@
 #include "util/graph_loader.hpp"
 
 #include "util/log.hpp"
+#include "util/timing_util.hpp"
 
 #include <boost/assert.hpp>
+
+#include <set>
 
 namespace osrm
 {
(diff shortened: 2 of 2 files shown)


src/extractor/node_based_graph_factory.cpp
```diff
@@ -150,20 +169,42 @@ void NodeBasedGraphFactory::CompressAnnotationData()
         for (EdgeID nbg_edge_id : compressed_output_graph.GetAdjacentEdgeRange(nbg_node_u))
         {
             auto const &edge = compressed_output_graph.GetEdgeData(nbg_edge_id);
-            annotation_mapping[edge.annotation_data] = 0;
+            auto res = unique_annotations.insert(edge.annotation_data);
+            annotation_mapping[edge.annotation_data] = *res.first;
```
MoKob: Since we are looking at the ID's in order, we could directly assign the correct mapping here.

src/extractor/node_based_graph_factory.cpp
```diff
@@ -150,20 +169,42 @@ void NodeBasedGraphFactory::CompressAnnotationData()
...
         }
     }
 
     // now compute a prefix sum on all entries that are 0 to find the new mapping
+    std::unordered_map<AnnotationID, AnnotationID> renumbering;
```
MoKob: and make all of this superfluous

include/extractor/node_based_edge.hpp
```diff
@@ -71,6 +71,23 @@ struct NodeBasedEdgeAnnotation
             std::tie(name_id, classes, travel_mode, is_left_hand_driving) ==
             std::tie(other.name_id, other.classes, other.travel_mode, other.is_left_hand_driving));
     }
+
+    bool operator < (const NodeBasedEdgeAnnotation & other) const
+    {
+        if (name_id != other.name_id)
```
MoKob: this could be `std::tie() < std::tie()`, right?

src/extractor/node_based_graph_factory.cpp
MoKob: No, but we are assigning them based on valid/used entries. Before we were only using a vector to remove duplicated entries. But you are hashing all entries anyhow, so you could directly assign a final value, if you are do full re-mapping anyhow. That way we could remove prefix sum computations and do a single re-ordering step.

deniskoronchik:
Some numbers:
Osm file   | Old annotations | New annotations | %
----------------------------------------------------------------------
ohio         | 468159              | 124906                |26.6802518

src/extractor/node_based_graph_factory.cpp
```diff
@@ -191,7 +212,11 @@ void NodeBasedGraphFactory::CompressAnnotationData()
     const auto old_size = annotation_data.size();
     // remove all remaining elements
     annotation_data.erase(new_end, annotation_data.end());
-    util::Log() << " graoh compression removed " << (old_size - annotation_data.size())
+
+    // reorder data in the same order
+    std::sort(annotation_data.begin(), annotation_data.end());
```
MoKob: Is this sorting necessary, the std::set is sorted. So we are guaranteed to see IDs in order, when we assign cached_ids[id] = new_id. So for all annotation_ids before and after this algorithm, the IDs should be in order, right?

MoKob: Capturing from chat: since the set is ordered by std::tie instead of the ID, it is. But we should potentially replace this sort by sorting the annotation-ids after unique to ensure that we are future-proof.

src/extractor/node_based_graph_factory.cpp
```diff
@@ -174,11 +183,23 @@ void NodeBasedGraphFactory::CompressAnnotationData()
...
+
+            edge.annotation_data = it2->second;
         }
     }
 
+    // mark unused references for remove
+    for (AnnotationID id = 0; id < annotation_data.size(); ++id)
+    {
+        auto const it = unique_annotations.find(id);
+        if (it == unique_annotations.end() || *it != id)
```
oxidase: just curious how `*it != id` is possible here for `AnnotationID`?

src/extractor/node_based_graph_factory.cpp
```diff
@@ -191,7 +212,11 @@ void NodeBasedGraphFactory::CompressAnnotationData()
     const auto old_size = annotation_data.size();
     // remove all remaining elements
     annotation_data.erase(new_end, annotation_data.end());
-    util::Log() << " graoh compression removed " << (old_size - annotation_data.size())
+
+    // reorder data in the same order
+    std::sort(annotation_data.begin(), annotation_data.end());
+
+    util::Log() << " graph compression removed " << (old_size - annotation_data.size())
```
oxidase: Please add here time output for the method scope. for `germany-latest` in release it costs 13.5 seconds, but interesting is timing for a planet-sized extraction.

src/extractor/node_based_graph_factory.cpp
MoKob: This happens, since we don't compare based on ID but based on annotation_data content. So there are multiple IDS (like 1,2,3) which are equal, but we only have a single representation for them (e.g. 3)

deniskoronchik: Set compares by real data. And store any unique id for each equal data. In this line we mark for remove all other duplicates and save just this one (unique)

deniskoronchik:
Possible windows compilation fixed, wait for test results

src/extractor/node_based_graph_factory.cpp
oxidase: ok, but it is just another source of non-determinism: `{1,2,3}` also a posible set

deniskoronchik: yes, and thats not a problem for algorithm, because in that case we also remove just duplicated annotation_data values (with the same data values)

oxidase: as per chat: it is not a problem in the implementation

MoKob: To extend on the reasoning here: the implementation always choses the first ID within the graph as a representative. There is no randomness involved in the choosing process, making this deterministic.

deniskoronchik:
For a whole planet:
```
graph compression removed 69940393 annotations of 77624314 in 163.24 seconds
```

MoKob:
Nice. Half a gigabyte less main memory. This should be even more on walk/bike.

PR #4634, opened 2017-10-23
Contract the exclude-flag sub-graphs as well

TheMarex:
# Issue

Looks like we remove one line too much 🙈  when removing CoreCH.

## Tasklist
 - [ ] review
 - [ ] adjust for comments


oxidase:
Just a note: with running demo server as a part of integration testing such kinds of issues can be discovered earlier.

PR #4636, opened 2017-10-25
Adds cardinal_directions flag to profiles and disables ref-rewriting by default

daniel-j-h:
At the moment we append cardinal directions based on relations to the `ref` field, e.g. `I-80 $north`. This breaks compatibility with all existing clients.

This changeset adds a `cardinal_directions = false` flag to the default car profile disabling cardinal direction re-writing by default.

### DIFF
--- a/features/car/route_relations.feature
+++ b/features/car/route_relations.feature
@@ -4,6 +4,11 @@ Feature: Car - route relations
         Given the profile "car"
 
     Scenario: Assignment using relation membership roles
+        Given the profile file "car" initialized with
+        """
+        profile.cardinal_directions = true
+        """
+
         Given the node map
             """
               a----------------b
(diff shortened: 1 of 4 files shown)


features/car/route_relations.feature
```diff
@@ -184,6 +248,11 @@ Feature: Car - route relations
             | a,b       | eastbound,eastbound  | I 80 $east,I 80 $east |
 
     Scenario: Direction only available via superrelation direction
+        Given the profile file "car" initialized with
+        """
+        profile.cardinal_directions = true
```
MoKob: for completeness, I would like to see a test with explicitly setting `profile.cardinal_directions = false`

MoKob:
Also, @daniel-j-h I cannot explain why that feature flag now breaks coverage for alternatives on both MLD and CH. I would expect the code paths for these parts to remain the same.

daniel-j-h:
The idea to have a switch in the profile will make sense once we have OSRM-text-instruction support.

Then users can enable it when they can enforce clients using a recent OSRM-text-instruction version.

Explicit test added, rebased. Should be ready to merge into master and 5.13.

Issue #4266, opened 2017-07-10
Alternatives: stretch-test based on detour


daniel-j-h:
Splitting out into a ticket from https://github.com/Project-OSRM/osrm-backend/pull/4047 (please read for context).

The alternatives feature on mld generates a lot of via candidates and then filters them based on predicates. The stretch-test checks for the alternatives not to be longer than c * shortest path.

What we need to check, though, is how much longer the _detour_ is compared to the main route the alternative is skipping. A route that is 20% longer than the primary route during the route's last two minutes is pretty bad.

We should compare the weights for `a,via,b` and compare against the the weights for `a,b`:

```
         .  via  .
      .             .
s - a - - - - - - - - b - t
```

This check needs the paths already and should probably happen on packed paths using weights and at the end on the annotated routes using actual duration.

PR #4639, opened 2017-10-26
Implement issue #4266: alternative stretch-test based on detour

deniskoronchik:
# Issue

#4266 

## Tasklist
 - [ ] review
 - [ ] adjust for comments

### DIFF
--- a/src/engine/routing_algorithms/alternative_path_mld.cpp
+++ b/src/engine/routing_algorithms/alternative_path_mld.cpp
@@ -429,16 +429,20 @@ filterAnnotatedRoutesByStretch(RandIt first, RandIt last, const InternalRouteRes
 
     BOOST_ASSERT(shortest_route.is_valid());
 
-    static constexpr double kStretchScale = 1.2f;
-
-    //    const auto shortest_route_duration = shortest_route.duration();
-    //    const auto scaled_at_most_longer_by =
-    //        scaledAtMostLongerByFactorBasedOnDuration(shortest_route_duration);
-    //    const auto stretch_duration_limit = (1. + scaled_at_most_longer_by) *
-    //    shortest_route_duration;
-
+    /* This function compare each alternative route with shortest one to check if
+     * different part of it no longer then shortes one by 25%.
+     * For example we have to ways:
+     *     c d e
+     * a b       f
+     *     k j l
+     *
+     * Shortest: a b c d e f
+     * Alternative: a b k j l f
+     *
+     * So this function ensure, that len(kjl) <= (1 + kAtMostLongerBy) * len(cde)
+     */
     const auto over_duration_limit = [&](const InternalRouteResult &route) {
-        bool passed = true;
+        bool over_limit = false;
         BOOST_ASSERT(route.unpacked_path_segments.size() ==
                      shortest_route.unpacked_path_segments.size());
         for (size_t i = 0; i < shortest_route.unpacked_path_segments.size(); ++i)
@@ -449,35 +453,45 @@ filterAnnotatedRoutesByStretch(RandIt first, RandIt last, const InternalRouteRes
             // find different parts of two segments
             // algorithm use rule, that there are no difference at begin/end of segment
             // also that there are just on different sequence
-            auto s_it_b = short_seg.begin();
-            auto a_it_b = alter_seg.begin();
-            while ((*s_it_b).turn_via_node == (*a_it_b).turn_via_node &&
-                   s_it_b != short_seg.end() && a_it_b != alter_seg.end())
+            auto short_it_begin = short_seg.begin();
+            auto alter_it_begin = alter_seg.begin();
+            while ((*short_it_begin).turn_via_node == (*alter_it_begin).turn_via_node &&
+                   short_it_begin != short_seg.end() && alter_it_begin != alter_seg.end())
             {
-                ++s_it_b;
-                ++a_it_b;
+                ++short_it_begin;
+                ++alter_it_begin;
             }
 
-            if (s_it_b == short_seg.end() || a_it_b == alter_seg.end())
+            if (short_it_begin == short_seg.end() || alter_it_begin == alter_seg.end())
             {
-                passed = false;
+                over_limit = true;
                 break;
             }
 
-            auto s_it_e = short_seg.rbegin();
-            auto a_it_e = alter_seg.rbegin();
-            while ((*s_it_b).turn_via_node == (*a_it_b).turn_via_node &&
-                   s_it_e != short_seg.rend() && a_it_e != alter_seg.rend())
+            if (short_it_begin != short_seg.begin())
+                short_it_begin--;
+            if (alter_it_begin != alter_seg.begin())
+                alter_it_begin--;
+
+            auto short_it_end = short_seg.rbegin();
+            auto alter_it_end = alter_seg.rbegin();
+            while ((*short_it_begin).turn_via_node == (*alter_it_begin).turn_via_node &&
+                   short_it_end != short_seg.rend() && alter_it_end != alter_seg.rend())
             {
-                ++s_it_e;
-                ++a_it_e;
+                ++short_it_end;
+                ++alter_it_end;
             }
 
-            if (s_it_e == short_seg.rend() || a_it_e == alter_seg.rend())
+            if (short_it_end == short_seg.rend() || alter_it_end == alter_seg.rend())
             {
-                passed = false;
+                over_limit = true;
                 break;
             }
+
+            if (short_it_end != short_seg.rbegin())
+                short_it_end--;
+            if (alter_it_end != alter_seg.rbegin())
+                alter_it_end--;
 
             // now calculate duration of found parts
             const auto calcDuration = [](auto it_begin, auto it_end) {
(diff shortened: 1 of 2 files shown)


src/engine/routing_algorithms/alternative_path_mld.cpp
```diff
@@ -429,13 +429,83 @@ filterAnnotatedRoutesByStretch(RandIt first, RandIt last, const InternalRouteRes
...
-    const auto scaled_at_most_longer_by =
-        scaledAtMostLongerByFactorBasedOnDuration(shortest_route_duration);
-    const auto stretch_duration_limit = (1. + scaled_at_most_longer_by) * shortest_route_duration;
+    static constexpr double kStretchScale = 1.2f;
+
+    //    const auto shortest_route_duration = shortest_route.duration();
+    //    const auto scaled_at_most_longer_by =
+    //        scaledAtMostLongerByFactorBasedOnDuration(shortest_route_duration);
+    //    const auto stretch_duration_limit = (1. + scaled_at_most_longer_by) *
+    //    shortest_route_duration;
```
daniel-j-h: Can this be removed?

src/engine/routing_algorithms/alternative_path_mld.cpp
```diff
@@ -429,13 +429,83 @@ filterAnnotatedRoutesByStretch(RandIt first, RandIt last, const InternalRouteRes
...
+
+    const auto over_duration_limit = [&](const InternalRouteResult &route) {
```
daniel-j-h: Could you add a high-level description comment of what the following check does and why we need it

src/engine/routing_algorithms/alternative_path_mld.cpp
```diff
@@ -429,13 +429,83 @@ filterAnnotatedRoutesByStretch(RandIt first, RandIt last, const InternalRouteRes
...
+            }
+
+            if (s_it_b == short_seg.end() || a_it_b == alter_seg.end())
+            {
+                passed = false;
+                break;
+            }
+
+            auto s_it_e = short_seg.rbegin();
+            auto a_it_e = alter_seg.rbegin();
```
daniel-j-h: Here and above: could you either make the variable names more clear or add a comment with the route sketch from the issue

daniel-j-h:
There's also https://github.com/Project-OSRM/osrm-backend/blob/master/features/testbot/alternative.feature failing now because we no longer find an alternative. Feel free to adapt.

src/engine/routing_algorithms/alternative_path_mld.cpp
```diff
@@ -429,13 +429,83 @@ filterAnnotatedRoutesByStretch(RandIt first, RandIt last, const InternalRouteRes
...
+    //    const auto scaled_at_most_longer_by =
+    //        scaledAtMostLongerByFactorBasedOnDuration(shortest_route_duration);
+    //    const auto stretch_duration_limit = (1. + scaled_at_most_longer_by) *
+    //    shortest_route_duration;
+
+    const auto over_duration_limit = [&](const InternalRouteResult &route) {
+        bool passed = true;
+        BOOST_ASSERT(route.unpacked_path_segments.size() ==
+                     shortest_route.unpacked_path_segments.size());
+        for (size_t i = 0; i < shortest_route.unpacked_path_segments.size(); ++i)
```
MoKob: I am a bit confused by this loop. We should only ever have a single path-segment, or the deviation should consider all path segments.
The via-node approach we support only does support a single segment and a single deviation. The moment multiple nodes are selected for the route, we will not provide alternatives. What we do here is somewhat a mix.
The code seems to allow alternatives on a by-segment basis, but removes the entire route if a single segment doesn't offer a reasonable alternative.

I would like to either clarify that we only do single destination alternatives (-> segments.size() == 1) or replace invalid segments with the shortest path in between to allow alternatives on a per-segment basis.

deniskoronchik: This algorithm works fine with [0; N) segments, why we need it to remove it, if it more common, and prevent some future errors

src/engine/routing_algorithms/alternative_path_mld.cpp
```diff
@@ -429,13 +429,83 @@ filterAnnotatedRoutesByStretch(RandIt first, RandIt last, const InternalRouteRes
...
+
+            if (s_it_e == short_seg.rend() || a_it_e == alter_seg.rend())
+            {
+                passed = false;
+                break;
+            }
+
+            // now calculate duration of found parts
+            const auto calcDuration = [](auto it_begin, auto it_end) {
+                double duration = 0.0;
```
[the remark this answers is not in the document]
deniskoronchik: Doesn't need to cast it later, twicely. There are no any problems, that it's already double there

daniel-j-h:
Next actions here:
- [ ] fix test as outlined above
- [ ] fix outstanding comments from above, e.g. `calcDistance` shouldn't be needed
- [ ] use the [osrm-runner.js](https://github.com/Project-OSRM/osrm-backend/blob/master/scripts/osrm-runner.js) script and compare against master: performance as well as how many alternatives you will get on a few different datasets (e.g. bavaria, germany, california)

deniskoronchik:
@daniel-j-h it's very strange, because i've pushed test fix yesterday to my branch, but it doesn't appeared into PR

Will add comments, but algorithm that i've implemented was provided to me by @MoKob. 
Also will prepare stats

deniskoronchik:
@daniel-j-h Have no idea how to use this script to get compare data. Is there any instruction or ready test that can be suitable? Because it seems, that it require a lot of hand made work to prepare input csv file. Maybe there are already done work?

daniel-j-h:
You run the script once and it will create random queries - save these in a txt file. Then you can run repeated tests re-using the same queries you saved. For example something like this should get you started:

```
time node scripts/osrm-runner.js -s localhost:5000 -b 10.162,47.680,12.096,50.233 -p '/route/v1/driving/{};{}?alternatives=true' -n 10000
```

It would be great to compare timings and number of found alternatives for `alternatives={1,2,3,4,5}` for this branch and master.

deniskoronchik:
@daniel-j-h There are some results.

As you wrote, i've used script and run it first 50000 queries, on `ohio` BB fully inside it.
After that i've run it with `-q` param with the same queries but with my commits.
There are time results:

```
MASTER:
real	17m1.428s
user	0m33.936s
sys	0m8.376s

PR:
real	17m6.824s
user	0m35.696s
sys	0m9.244s

```

After that i've made python script to analyze this two csv files. It read each in memory an compare `queries`, `codes` and `values` fields (0, 1, 4 columns in a row) to ensure that there are equal queries.
After ensure, it calculates `min`, `avg`, `max` values for columns(2, 3, 5), where 5 my added column with number of routes in response. There are results (`ttfb`, `elapse` - names of column got from original script):

```
TTFB (min, avg, max):
FIXED: (0.546459, 20.064624352339855, 105.304418)
MASTER: (0.438234, 19.994908907320106, 88.859048)
Elapse (min, avg, max):
FIXED: (0.588163, 20.138843621239733, 105.372058)
MASTER: (0.482674, 20.067620924719886, 88.944774)
Routes num (min, avg, max):
FIXED: (1.0, 1.86986, 2.0)
MASTER: (1.0, 1.83378, 2.0)
```

deniskoronchik:
I think, that approach of alternative routes find should be changed. The main problem right now, that we use 2 directional dijkstra's algorithm, main problem, that candidates for alternative routes got from cache of main route finding. 
If we visualize it, that would be two intersected circles, but they have small intersection in that case we couldn't find very good candidate, because it would be a high speed road far from this one. There are no many cases, where two high speed roads goes side by side.

So I think, that implementing A* algorithm can solve this problem, because it can be implemented with heuristic function, that prioritize better class roads, (`motorway` has more priority, then `secondary`), another part of this function, heuristic distance to a goal.
I that case algorithm will prioritize check of better roads in specified direction. So we will have N branches from A to B using best roads. Another criteria should be, that 2 founded ways should be different for example, by 80% of nodes. In that case algorithm will try to go from A to B by best roads (check them firstly), then will check `secondary` ... `living_street`. 

The main problem to find a good heuristic function, that will include `distance` and `road class`

daniel-j-h:
Yes you need to figure out why your branch seems to find more routes then master - this should never happen if all you do here is filter out _more_ routes. Make sure to fix both the base map and the queries you send.

> I think, that approach of alternative routes find should be changed.

Please read https://github.com/Project-OSRM/osrm-backend/issues/3905 and http://algo2.iti.kit.edu/documents/Dissertation_Kobitzsch_Moritz.pdf section 5.2 "The Via-Node Paradigm" pages 59--65.

The approach we're doing here for MLD is fine - if we need more candidates we can make the search space larger and tune quality [coefficients in the impl.](https://github.com/Project-OSRM/osrm-backend/blob/73f4e1d45a8ec870dcc59868c76f5614b99fa526/src/engine/routing_algorithms/alternative_path_mld.cpp#L37-L52) to trade-off quality vs quantity.

deniskoronchik:
@daniel-j-h There are new resuts:

```
Data length: 50000
Compare request...
Compare code...
Compare values...
TTFB (min, avg, max):
FIXED: (0.464531, 20.616985472960177, 118.720778)
MASTER: (0.438234, 19.994908907320106, 88.859048)
Elapse (min, avg, max):
FIXED: (0.505933, 20.694088594979952, 118.81091)
MASTER: (0.482674, 20.067620924719886, 88.944774)
Routes num (min, avg, max):
FIXED: (1.0, 1.0832, 2.0)
MASTER: (1.0, 1.83378, 2.0)
```

But there are still no good routes, but some of them good. For example:
http://192.168.100.6:9966/?z=9&loc=39.872171865356904%2C-82.15558109339224&loc=40.339825739004816%2C-82.45377897051792&hl=en&alt=0

screenshot from 2017-11-08 18-32-16

Was filtered out in master.

http://192.168.100.6:9966/?z=9&loc=39.907928284114654%2C-82.37755522203265&loc=40.0740015584715%2C-81.39990872688985&hl=en&alt=0

screenshot from 2017-11-08 18-32-52

Wasn't filtered, because diff duration < 25%

There are a list of queries, where `master` filter out result, and this PR didn't:
[out2.txt](https://github.com/Project-OSRM/osrm-backend/files/1454450/out2.txt)

Was ~6000 cases, for now 412 cases

deniskoronchik:
@daniel-j-h Also i've checked for example in case http://192.168.100.6:9966/?z=9&loc=40.76237765397561%2C-82.33968788244843&loc=40.09863185660194%2C-83.13247493222241&hl=en&alt=0 i got for filter just on alternative route. And it's already very bad. With uturn in on way

deniskoronchik:
@daniel-j-h  Why thats going on? 
peek 2017-11-14 14-51

deniskoronchik:
with 
```cpp
const constexpr auto kAtMostLongerBy = 0.5;
```


```
Data length: 50000
Compare request...
Compare code...
Compare values...
TTFB (min, avg, max):
FIXED: (0.554502, 20.55214023020013, 90.156938)
MASTER: (0.438234, 19.994908907320106, 88.859048)
Elapse (min, avg, max):
FIXED: (0.596367, 20.6260494441399, 90.225105)
MASTER: (0.482674, 20.067620924719886, 88.944774)
Routes num (min, avg, max):
FIXED: (1.0, 1.09012, 2.0)
MASTER: (1.0, 1.83378, 2.0)
master-drop: 570
fixed-drop: 37753
no alternatives: 8311
```

DennisOSRM:

Issue #4642, opened 2017-10-27
Empty name_id checks need to be changed to empty name_id string checks


MoKob:
The guidance engine relies on the use of `name_id == EMPTY_NAMEID` in many places where we decide between announcing/not announcing a turn. These checks interact with [requiresNameAnnounced](https://github.com/Project-OSRM/osrm-backend/blob/948025440f886e84a58799c4529dadf712da1f1b/include/util/guidance/name_announcements.hpp#L113), which assumes turning onto a road without name doesn't require an announcement.
The checks for EMPTY_NAMEID precede the call of [requiresNameAnnounced](https://github.com/Project-OSRM/osrm-backend/blob/948025440f886e84a58799c4529dadf712da1f1b/include/util/guidance/name_announcements.hpp#L113) to ensure that roads without names are treated correctly.

With the addition of more and more fields to the field referenced to by `name_id`, a `name_id != EMPTY_NAMEID` can be true, even though the road doesn't have a name or a ref.

We need to revisit all checks that compare for `EMPTY_NAMEID` and ensure that they are working as intended when `destination:ref` (for example) is present on a road without name or ref.

Places to check:

 - [x]  src/engine/guidance/collapse_scenario_detection.cpp
 - [x]  src/extractor/extraction_containers.cpp (most likely fine)
 - [x] src/extractor/guidance/intersection_handler.cpp
 - [x] src/extractor/guidance/mergable_road_detector.cpp
 - [x] src/extractor/guidance/motorway_handler.cpp
 - [x] src/extractor/guidance/roundabout_handler.cpp
 - [x] src/extractor/guidance/sliproad_handler.cpp
 - [x] src/extractor/guidance/turn_handler.cpp

PR #4643, opened 2017-10-27
Fix/wide uturns onto `destination:ref`

MoKob:
# Issue

One step towards fixing side-effects of https://github.com/Project-OSRM/osrm-backend/issues/4642.

## Tasklist
 - [x] review
 - [x] adjust for comments


daniel-j-h:
Formatted; force pushed. Two Travis builds were timing out; restarted.

Issue #4178, opened 2017-06-20
Return all exit angles for roundabouts


danpat:
When a navigation UI wants to render icons for maneuvers, it can use the `intersections` object to render a graphical representation of the upcoming intersection (including valid lane information).

However, roundabouts are problematic.  They're not represented by a single intersection object, we get an intersection object for each roundabout exit that the route uses.  We supply no data for exits that the route does not pass, making rendering of roundabout icons incomplete.

We should try to figure out a place to put additional roundabout metadata.

One option would be to add an `intersection` object to the `StepManeuver` for the roundabout:

```
                     "maneuver" : {
                        "exit" : 2,
                        "bearing_after" : 63,
                        "type" : "roundabout",
                        "modifier" : "straight",
                        "location" : [
                           -77.029316,
                           38.909145
                        ],
                        "bearing_before" : 43,
                        "intersection" :   // <- New intersection object when type=roundabout
                        {
                           "location" : [
                              -77.029316,
                              38.909145
                           ],
                           "entry" : [
                              false,   // Don't use the first exit
                              true,    // The second exit is the one to take
                              false    // Third exit is not on our route at all, but we include it for rendering
                           ],
                           "in" : 1,
                           "out" : 0,
                           "bearings" : [
                              60,
                              225,
                              240
                           ]
                        },
                     },
```

We have precedence for this with the `exit` property.

MoKob:
@danpat I like the idea for rendering all roundabout exits. I don't know whether your proposal  would be considered a breaking change though, since we would change the interpretation of an intersection.

E.g. in your example, taking the third exit would look like entering a roundabout to the left instead of to the right. Also, the entry flags are quite confusing to me, since what I would read from your example is that I would exit the same way I enter (in == 1 == true). But that is most likely me being not fully awake yet :).

How would you feel about adding a full shape to the object? So we wouldn't change the meaning of the intersection itself but add something like you describe as a property?

```
{
  "roundabout_shape": {
	"entry": [false,true,false,true],
	"exit" : [true,false,true,true],
	"bearings": [0,90,180,270]
  }
}
```

This shape would describe the roundabout in full:
 - four roads
 - bearings 0, 180, and 270 allow exiting the roundabout
 - bearings 90, 270 allow entering it (making all roads but 270 oneways)
 - bearings show how to draw arrows
 - based on the maneuver the correct bearings can be highlighted (or we add in/out in our processing)

[...]

PR #4649, opened 2017-11-01
Expose driving side per step

danpat:
# Issue

If you want to draw icons in a UI to represent a route, it can be helpful to know the driving side for a step so you can display the correct image.  

There are two use-cases this is aimed at:

  1. Drawing u-turn arrows.  We currently return `type=turn, maneuver=uturn`, and no other flag to indicate whether the uturn is ↶ or ↷.
  2. Drawing roundabout imagery.  In order to draw an image like the one describe [here](https://github.com/Project-OSRM/osrm-backend/issues/4178#issuecomment-310841761), you need to know whether to paint and count exits clockwise or counterclockwise.  You can currently make a *guess* from the intersection bearings, but it would be better to have an explicit flag.

This PR changes:
  1. Adds a new property to each `RouteStep` called `driving_side` with values of either `left` or `right`
  2. Bundles a GeoJSON file that roughly captures left-hand-side driving countries (I didn't do all the Carribean islands, and the land-border countries are pretty roughly drawn)

Open questions:
  1. Should the `driving_side` be a property of the `RouteStep` ?  A route may cross a border where the driving-side changes, and if there is no maneuver at the side-changing location, there would be no way to indicate it with the current scheme until the next maneuver is reached.  Maybe it's better to make it an `annotation`, or maybe a property of the `intersections` array?  Input appreciated.

This PR basically works - if the current design is acceptable, I'll add test cases and target this for OSRM 5.14.

The `driving_side.geojson` file could do with a bit of work - land borders are pretty rough, and I've missed a bunch of islands.  If anyone wants some straightforward polygon drawing task, this one is pretty straightforward :-)

/cc @oxidase 

## Tasklist
 - [x] update API documentation
 - [x] review
 - [x] adjust for comments


profiles/lib/way_handlers.lua
```diff
@@ -563,7 +563,7 @@ function WayHandlers.driving_side(profile, way, result, data)
    elseif driving_side == 'right' then
       result.is_left_hand_driving = false
    else
-      result.is_left_hand_driving = profile.left_hand_driving
+      result.is_left_hand_driving = profile.properties.left_hand_driving
```
oxidase: Good catch! But  tests preconditions `profile.left_hand_driving = true` must be updated also

danpat: Done!

include/engine/guidance/assemble_steps.hpp
```diff
@@ -204,22 +205,24 @@ inline std::vector<RouteStep> assembleSteps(const datafacade::BaseDataFacade &fa
...
+                      NO_ROTARY_NAME,
+                      duration / 10.,
+                      distance,
+                      weight / weight_multiplier,
+                      target_mode,
+                      maneuver,
+                      leg_geometry.FrontIndex(segment_index),
+                      leg_geometry.BackIndex(segment_index) + 1,
+                      {intersection},
+                      facade.IsLeftHandDriving(target_node_id)}); // TODO: this one is wrong
```
TheMarex: Why is it wrong?

package-lock.json
```diff
@@ -1,8 +1,19 @@
 {
   "name": "osrm",
-  "version": "5.13.0-latest.1",
+  "version": "5.14.0-latest.1",
```
TheMarex: We should not include the `package-lock.json` diff. Seems like this was accidentally committed?

include/engine/guidance/assemble_steps.hpp
danpat: That comment is referring to itself (the comment is wrong) ;-)  Deleting.

Issue #4650, opened 2017-11-01
FindLibOSRM cannot find libosrm


drnextgis:
Try to build custom docker image based on official one and use FindLibOSRM.cmake from [here](https://github.com/Project-OSRM/node-osrm/blob/master/cmake/FindLibOSRM.cmake) to build own OSRM-based application. But CMake cannot find `libosrm` because:

```
$ pkg-config --libs libosrm
Package libosrm was not found in the pkg-config search path.
Perhaps you should add the directory containing `libosrm.pc'
to the PKG_CONFIG_PATH environment variable
Package 'libosrm', required by 'virtual:world', not found
```

It works if specify  `PKG_CONFIG_PATH` manually:

```
PKG_CONFIG_PATH=/usr/local/lib64/pkgconfig pkg-config
 --libs libosrm
-L/usr/local/lib -losrm -flto -Wl,--gc-sections -Wl,-O1 -Wl,--hash-style=gnu -Wl,--sort-common
```

What is the right way to fix this issue?

danpat:
Hmm, looks like the problem is that we're installing the `.pc` file in `CMAKE_INSTALL_LIBDIR/pkgconfig`.  For whatever reason, under the Alpine environment, that resolves to `/usr/local/lib64/pkgconfig`.  However, `pkg-config` on Alpine doesn't have that path in it's search list:

```
# pkg-config --variable pc_path pkg-config
/usr/local/lib/pkgconfig:/usr/local/share/pkgconfig:/usr/lib/pkgconfig:/usr/share/pkgconfig
```

I think the fix is this PR: https://github.com/Project-OSRM/osrm-backend/pull/4652, but I'd like some feedback on that before merging.

daniel-j-h:
#4652 landed in master - @drnextgis could you try and see if it helps?

drnextgis:
I've already tried and all works fine, thanks!

daniel-j-h:
Perfect - happy routing! 🎉 🚗

PR #4652, opened 2017-11-01
Make sure .pc files end up where pkg-config can find them

danpat:
# Issue

Found in https://github.com/Project-OSRM/osrm-backend/issues/4650, on Alpine Linux, we were installing `.pc` files in `/usr/local/lib64` (determined by CMAKE), but `pkg-config` doesn't actually look there for them.

This PR moves the `.pc` file installation to the same `lib` heirarchy location as `libosrm.a`, which is usually actually in a place where `pkg-config` is configured to look.

## Tasklist
 - [x] review
 - [x] adjust for comments


CMakeLists.txt
```diff
@@ -796,7 +796,7 @@ JOIN("-I${DEPENDENCIES_INCLUDE_DIRS}" " -I" PKGCONFIG_OSRM_INCLUDE_FLAGS)
 JOIN("${ENGINE_LIBRARIES}" " " PKGCONFIG_OSRM_DEPENDENT_LIBRARIES)
 
 configure_file(${CMAKE_CURRENT_SOURCE_DIR}/cmake/pkgconfig.in libosrm.pc @ONLY)
-install(FILES ${PROJECT_BINARY_DIR}/libosrm.pc DESTINATION ${CMAKE_INSTALL_LIBDIR}/pkgconfig)
+install(FILES ${PROJECT_BINARY_DIR}/libosrm.pc DESTINATION ${PKGCONFIG_LIBRARY_DIR}/pkgconfig)
```
daniel-j-h: Is the extra `pkgconfig` required?

Looking at

https://github.com/Project-OSRM/osrm-backend/blob/0b6eb85106425a1c1c0f08fe6a8e1936245b85e1/cmake/pkgconfig.in#L3

I don't think so - could you check to make sure we're in sync here.

danpat: @daniel-j-h `.pc` files go into `/usr/local/lib/pkgconfig`, the `PKGCONFIG_LIBRARY_DIR` is where we put `libosrm.a`

Issue #4655, opened 2017-11-01
Remove dependency of guidance code from ProcessTurn


ghoshkaj:
We want to remove guidance code from the EdgeBasedGraphFactory. This is the first step in decoupling the guidance generation and the edge based graph generation.

PR #4656, opened 2017-11-01
Refactor process_turn lua function

ghoshkaj:
# Issue

[Remove dependency of guidance code from ProcessTurn](https://github.com/Project-OSRM/osrm-backend/issues/4655)

## Tasklist
 - [x] review
 - [x] adjust for comments
 - [x] changelog
 - [x] rebase -i and then reorganize commit messages
- [ ] mergey merge


profiles/car.lua
```diff
@@ -420,14 +420,16 @@ function process_turn(profile, turn)
       turn.duration = profile.properties.traffic_light_penalty
   end
 
-  if turn.turn_type ~= turn_type.no_turn then
+  -- if turn.turn_type ~= turn_type.no_turn then
+
+  if turn.number_of_roads > 2 then
```
daniel-j-h: If we expose the `is_u_turn` boolean should we expose a flag here `is_a_turn`?

We don't need the number of roads in this function otherwise, no?

ghoshkaj: @TheMarex pointed out that we don't want to make the assumption of `turn.number_of_roads > 2` become equal to `is_a_turn`.

So let's leave this as is for now 😄

TheMarex:
@ghoshkaj looks like only the formatter is complaining, after a run of `format.sh` this should be good to go!

ghoshkaj:
Related to https://github.com/Project-OSRM/osrm-backend/issues/4664

PR #4661, opened 2017-11-03
Needs some flames

danpat:
# Issue

`osrm-runner` did not have enough flames.  This PR adds some.  You're welcome.

## Tasklist
 - [ ] adjust for comments - note: the only valid adjustment would be to add even more flames


oxidase:
:fire: does not look like screenshot from 2017-11-03 12-01-47

danpat:
Fine, nobody likes my flames.

PR #4665, opened 2017-11-03
Slightly decreases roundabout turn max radius from 25m to 15m

daniel-j-h:
Roundabout turns are meant for situations like the following:

[real-life](https://www.flickr.com/photos/23136132@N04/4483170372)

Quoting @1ec5:

> In general, neighborhood roundabouts should have the same dimensions as mini roundabouts, the only difference being a planter instead of a traversable bump. For example, the City of Sacramento’s [street design standards](https://www.cityofsacramento.org/-/media/Corporate/Files/Public-Works/Publications/Engineering/Development-Engineering/Design-Procedures-Manual/section15-street-design-standards.pdf?la=en#page=18) specify an inscribed circle diameter of 120&nbsp;feet (36&nbsp;meters) for standard single-lane roundabouts. The FHWA [recommends](https://safety.fhwa.dot.gov/intersection/innovative/roundabouts/fhwasa10006/) a diameter of 90–180&nbsp;ft. (27–55&nbsp;m) for standard roundabouts and 45–90&nbsp;ft. (13–27&nbsp;m) for mini roundabouts.

### DIFF
--- a/include/extractor/guidance/constants.hpp
+++ b/include/extractor/guidance/constants.hpp
@@ -26,7 +26,7 @@ const double constexpr DISTINCTION_RATIO = 2;
 const double constexpr MAX_ROUNDABOUT_RADIUS = 15;
 // Unnamed small roundabouts that look like intersections are announced as turns,
 // guard against data issues or such roundabout intersections getting too large.
-const double constexpr MAX_ROUNDABOUT_INTERSECTION_RADIUS = 25;
+const double constexpr MAX_ROUNDABOUT_INTERSECTION_RADIUS = 15;
 
 const double constexpr INCREASES_BY_FOURTY_PERCENT = 1.4;
 


TheMarex:
Next steps here:
- [ ] Generate a map of affected locations
- [ ] Pick a random subset and use Mapillary to verify if we only fixed locations that we wanted
- [ ] Decide if we want to have this as a global change

daniel-j-h:
Super quick comparison from the stats handler we just merged into master - on California base map:

| instruction  | master | this branch |
| ------------- | ------------- | ------------- |
| roundabout turn | 10 | 10 |
| enter roundabout turn | 1951 | 1702 |
| exit roundabout turn | 2007 | 1741 |
| enter and exit roundabout turn | 15 | 15 |

danpat:
Here's the diff:

screen shot 2017-11-06 at 3 53 50 pm

1ec5:
The table above distinguishes between `roundabout turn`, `enter roundabout turn`, and `exit roundabout turn`, whereas my understanding is that “roundabout turns” would be treated as atomic maneuvers. Can you explain the difference?

daniel-j-h:
Yes roundabout turns are atomic from the API's point of view after post-processing. The above table is from pre-processing and includes internal instructions:

https://github.com/Project-OSRM/osrm-backend/blob/9b83649a03ef77519cba3f41440fbeb9d253efc1/include/extractor/guidance/turn_instruction.hpp#L33-L68

Orthogonal to the turn instructions our roundabout handler knows about the different roundabout types:

https://github.com/Project-OSRM/osrm-backend/blob/9b83649a03ef77519cba3f41440fbeb9d253efc1/include/extractor/guidance/roundabout_type.hpp#L10-L16

In post-processing we then do e.g. exit counting for regular roundabouts and rotaries and collapse everything between enter and exit roundabout turn instructions. _Then_ you have your single roundabout turn instruction.

daniel-j-h:
Quick follow up here: currently the split between pre- and post-processing does not allow us to generate statistics for the post-processing part easily (other than querying for routes and see what we get out). 

Ideally we could move _all_ post-processing into pre-processing. Then we'd have maneuvers in our graph already and a clean and encapsulated guidance engine - something to look into in the future.

danpat:
Our debug tiles contain point features for every maneuver.  I generated the debug tile for x=511,y=340,z=10 (London, UK), for both this branch and `master`, and converted to GeoJSON with https://github.com/mapbox/vt2geojson

I then wrote a small script to compare the generated GeoJSON and make a map of the features that changed.

Here's the map:
http://bl.ocks.org/d/c5f88424738c01be146ce1638e5ee9cb

TL;DR - change looks fine.  A few roundabouts that were very close (but under) to the 25m threshold have been converted to proper roundabouts.  None of the changes look bad.

<details>
  <summary>GeoJSON diffing script</summary>

```
#!/usr/bin/env node

var fs = require('fs');

if (process.argv.length < 4) {
  console.log(`Usage: ${process.argv[1]} <before.geojson> <after.geojson>`);
  return;
}

var before = JSON.parse(fs.readFileSync(process.argv[2]));
var after = JSON.parse(fs.readFileSync(process.argv[3]));


var diff = [];

for (var i in before.features) {

    // TODO: do a deep equal here to compare all properties
    if (before.features[i].properties.type !== after.features[i].properties.type ||
        before.features[i].properties.modifier !== after.features[i].properties.modifier) {

        var f = before.features[i];
        f.properties.before_type = f.properties.type;
        f.properties.after_type = after.features[i].properties.type;
        f.properties.before_modifier = f.properties.modifier;
        f.properties.after_modifier = after.features[i].properties.modifier;
        delete f.properties.type;
        delete f.properties.modifier;
        diff.push(f);

    }
}

console.log(JSON.stringify({'type':'FeatureCollection','features':diff}));
```
</details>

Issue #4664, opened 2017-11-03
Separate guidance generation and EdgeBasedGraphFactory


TheMarex:
We want to move the guidance generation out of the EBGF. There are currently the following dependencies on the guidance code:

- Entry classes: At an intersection which roads are accessible. See [`classifyIntersection`](https://github.com/Project-OSRM/osrm-backend/blob/master/src/extractor/guidance/turn_classification.cpp)
- Bearing classes: At an intersection the bearings of roads. See [`classifyIntersection`](https://github.com/Project-OSRM/osrm-backend/blob/master/src/extractor/guidance/turn_classification.cpp)
- Turn lane classification:  Depends on the [`turn_analysis`](https://github.com/Project-OSRM/osrm-backend/blob/master/src/extractor/edge_based_graph_factory.cpp#L427-L431)

The easiest way to move the guidance code out of the EBGF is thus to move all these dependent parts with it, that is the generation of `TurnData`.

For generating `TurnData` we need the following thing:

1. Mapping from an edge-based-graph edge `(u, v, w)` to the node-based-graph edges `(u, v)` and `(v, w)`.
2. Using the `turn_id` from the edge-based-graph data to determine where to write the resulting `TurnData`.

So as result the EBGF will only do the following thing:
- Generate and label the edge-based-nodes
- Generate the mapping between node-based-graph and the edge-based-graph
- Generate the edge-based-edges (weight, duration) and a unique identifier for each turn `turn_id`.

The guidance code outside the EBGF will thus be responsible for:
- Generating the `TurnData` object using the node-based-graph and the edge-based-graph
- Writing the `TurnData` at the right position.

PR #4674, opened 2017-11-09
Split EBG and  guidance pipelines in EBGF

oxidase:
# Issue

The PR is the implementation of #4664. 

## Tasklist
 - [x] Split intersection shape generation stage from `processor_stage`
 - [x] Duplicate EBG and guidance stage in one pipeline
 - [x] Cleanup the EBG stage of guidance code, use only EBG-related processing of `generate_edge`
 - [x] Cleanup the guidance stage of EBG code, use only guidance-related processing of `generate_edge`
 - [x] Split EBG and guidance into two pipelines with duplicated intersection shape generation **Check if needed in this PR**
 - [x] check regression / cucumber cases (see docs/testing.md)
 - [x] review
 - [x] adjust for comments

## Requirements / Relations


oxidase:
@ghoshkaj pushed the current branch state that is working, but almost stuck with the third point `Cleanup the EBG stage of guidance code, use only EBG-related processing of generate_edge` because it is nearly impossible in the current implementation. The problem is that turn and lane analysis code influences intersection connectivity, eg, removing 
https://github.com/Project-OSRM/osrm-backend/blob/e874f65dd61b03bf96bea6b2dd4b288c77960eb1/src/extractor/edge_based_graph_factory.cpp#L716
will prevent running this line of code
https://github.com/Project-OSRM/osrm-backend/blob/e874f65dd61b03bf96bea6b2dd4b288c77960eb1/src/extractor/guidance/turn_lane_matcher.cpp#L231
and EBG will miss a U-turn edge. Ideally intersections shape and connectivity must be analyzed before edges and turns generation, but atm intersection views can be mutated by the turn and lanes handlers during the generation phase. This makes the splitting of code extremely fragile with respect to indexing turns and consequences of the incorrect indexing can be visible only in post-processing.

The drawback of this PR without intersection connectivity pre-analysis is in increasing EBGF processing time without any facilitation in solving current open issues, like multiple via-way restrictions or U-turns  penalization in  segregated intersections.

include/extractor/extractor.hpp
```diff
@@ -101,6 +103,18 @@ class Extractor
...
+    void ProcessGuidanceTurns(
+        const util::NodeBasedDynamicGraph &node_based_graph,
+        const EdgeBasedNodeDataContainer &edge_based_node_container,
+        const std::vector<util::Coordinate> &node_coordinates,
+        const CompressedEdgeContainer &compressed_edge_container,
+        const std::unordered_set<NodeID> &barrier_nodes,
+        const std::vector<TurnRestriction> &turn_restrictions,
+        const std::vector<ConditionalTurnRestriction> &conditional_turn_restrictions,
+        const util::NameTable &name_table,
+        LaneDescriptionMap lane_description_map,
```
TheMarex: Is this passed on purpose by value?

src/updater/updater.cpp
```diff
@@ -112,9 +114,6 @@ void checkWeightsConsistency(
     extractor::EdgeBasedNodeDataContainer node_data;
     extractor::files::readNodeData(config.GetPath(".osrm.ebg_nodes"), node_data);
 
-    extractor::TurnDataContainer turn_data;
-    extractor::files::readTurnData(config.GetPath(".osrm.edges"), turn_data);
```
TheMarex: Oh nice where is this simplification coming from?

src/guidance/guidance_processing.cpp
```diff
@@ -0,0 +1,325 @@
...
+#include <tbb/blocked_range.h>
+#include <tbb/pipeline.h>
+#include <tbb/task_scheduler_init.h>
+
+namespace osrm
+{
+namespace guidance
+{
+
+void processGuidanceTurns(const util::NodeBasedDynamicGraph &node_based_graph,
```
TheMarex: Nitpick on naming here: What we are doing is annotation the turn with guidance information, so for example something like `guidance::annotateTurns` might be sufficient as a name.

include/guidance/turn_data_container.hpp
```diff
@@ -0,0 +1,123 @@
...
+    friend void serialization::read<Ownership>(storage::io::FileReader &reader,
+                                               TurnDataContainerImpl &turn_data_container);
+    friend void serialization::write<Ownership>(storage::io::FileWriter &writer,
+                                                const TurnDataContainerImpl &turn_data_container);
+
+  private:
+    Vector<guidance::TurnInstruction> turn_instructions;
+    Vector<LaneDataID> lane_data_ids;
+    Vector<EntryClassID> entry_class_ids;
+    Vector<util::guidance::TurnBearing> pre_turn_bearings;
```
TheMarex: Just a nitpick: We should probably move `TurnBearing` to the `guidance` module/namespace.

src/updater/updater.cpp
oxidase: it is a left over from some other refactoring step. `turn_data` is loaded, but not used

include/engine/guidance/assemble_geometry.hpp
```diff
@@ -70,7 +70,7 @@ inline LegGeometry assembleGeometry(const datafacade::BaseDataFacade &facade,
         cumulative_distance += current_distance;
 
         // all changes to this check have to be matched with assemble_steps
-        if (path_point.turn_instruction.type != extractor::guidance::TurnType::NoTurn)
+        if (path_point.turn_instruction.type != osrm::guidance::TurnType::NoTurn)
```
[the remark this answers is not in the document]
oxidase: using `guidance::TurnType` leads to an error  "`engine::guidance::TurnType` is undefined type". To avoid ambiguity `osrm::guidance::TurnType` must be used here.

include/extractor/extractor.hpp
oxidase: yes, `lane_description_map` is modified by the turn lane handler. Before `lane_description_map` was a non-const reference, now it is moved into the function scope, modified and serialized  in `ProcessGuidanceTurns`

include/guidance/turn_data_container.hpp
oxidase: moved to `include/guidance/turn_bearing.hpp`

src/extractor/edge_based_graph_factory.cpp
```diff
@@ -589,19 +549,12 @@ void EdgeBasedGraphFactory::GenerateEdgeExpandedEdges(
...
-                                  util::guidance::TurnBearing(turn.bearing)};
-
             // compute weight and duration penalties
             auto is_traffic_light = m_traffic_lights.count(intersection_node);
             ExtractionTurn extracted_turn(
-                turn.angle,
+                turn_angle,
                 m_node_based_graph.GetOutDegree(intersection_node),
-                turn.instruction.IsUTurn(),
+                guidance::getTurnDirection(turn_angle) == guidance::DirectionModifier::UTurn,
```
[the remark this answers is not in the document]
oxidase: just 0 (360) degrees turns.

src/extractor/edge_based_graph_factory.cpp
```diff
@@ -649,29 +602,25 @@ void EdgeBasedGraphFactory::GenerateEdgeExpandedEdges(
...
+        // Edge-based-graph stage
+        //
+        tbb::filter_t<tbb::blocked_range<NodeID>, EdgesPipelineBufferPtr> processor_stage(
             tbb::filter::parallel, [&](const tbb::blocked_range<NodeID> &intersection_node_range) {
 
-                auto buffer = std::make_shared<PipelineBuffer>();
-                buffer->nodes_processed =
-                    intersection_node_range.end() - intersection_node_range.begin();
-
-                // If we get fed a 0-length range for some reason, we can just return right away
```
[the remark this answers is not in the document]
oxidase: yes, `intersection_node_range` size will also be checked in the for loop pre-condition below

src/extractor/edge_based_graph_factory.cpp
```diff
@@ -933,44 +827,23 @@ void EdgeBasedGraphFactory::GenerateEdgeExpandedEdges(
...
-                                               data.turn_duration_penalties.end());
-                turn_data_container.append(data.turn_data_container);
-                turn_indexes_write_buffer.insert(turn_indexes_write_buffer.end(),
-                                                 data.turn_indexes.begin(),
-                                                 data.turn_indexes.end());
+                routing_progress.PrintAddition(buffer->nodes_processed);
 
+                // Copy data from local buffers into global EBG data
+                std::for_each(
+                    buffer->continuous_data.begin(), buffer->continuous_data.end(), transfer_data);
```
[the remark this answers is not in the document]
oxidase: `transfer_data` is use din two places to copy `buffer->continuous_data` and `delayed_data`

Issue #4680, opened 2017-11-14
Add the ability to "stand in place" for the map matching plugin 


gardster:
OSRM can not match frequent GPS tracks. Decimation of tracks became a mandatory requirement before sending points to the match plugin. This is happens because we are looking for the nearest "candidates" to each point from the input. But these "candidates" do not take into account the situation when the GPS signal drifts, but the vehicle did not move at the ground. I propose to add such states into the HMM inside the match plugin. This will allow us to match unfiltered GPS tracks "out of the box".

PR #4681, opened 2017-11-14
Add the ability to "stand in place" for the map matching plugin 

gardster:
# Issue #4680

Adds additional states to the HMM model, that describes that vehicle didn't change it's position.
It can decrease performance in cases of low GPS accuracy and frequent points, but allows to match raw GPS traces. Also this states handle situation with GPS drift on traffic lights. 

Also I add several fixes to exclude movement forward and than backward on extremely close trace points:
- Phantom nodes can be indistinct when difference between them less then 1 EdgeWeight measure.
- I invalidate backward edge in the candidate if we use it for the forward direction in the previous step

Example of map matching with this feature:
http://bl.ocks.org/d/bd91d2937e6fb8d2ebebed1ffe428651

### DIFF
--- a/include/engine/phantom_node.hpp
+++ b/include/engine/phantom_node.hpp
@@ -127,6 +127,19 @@ struct PhantomNode
 
     bool operator==(const PhantomNode &other) const { return location == other.location; }
 
+    bool IsIndistinct(const PhantomNode &other) const
+    {
+        if (forward_segment_id.enabled && other.forward_segment_id.enabled && forward_segment_id.id == other.forward_segment_id.id && forward_weight_offset == other.forward_weight_offset && forward_weight == other.forward_weight)
+        {
+            return true;
+        }
+        if (reverse_segment_id.enabled && other.reverse_segment_id.enabled && reverse_segment_id.id == other.reverse_segment_id.id && reverse_weight_offset == other.reverse_weight_offset && reverse_weight == other.reverse_weight)
+        {
+            return true;
+        }
+        return false;
+    }
+
     template <class OtherT>
     explicit PhantomNode(const OtherT &other,
                          ComponentID component,
--- a/include/engine/phantom_node.hpp
+++ b/include/engine/phantom_node.hpp
@@ -129,11 +129,15 @@ struct PhantomNode
 
     bool IsIndistinct(const PhantomNode &other) const
     {
-        if (forward_segment_id.enabled && other.forward_segment_id.enabled && forward_segment_id.id == other.forward_segment_id.id && forward_weight_offset == other.forward_weight_offset && forward_weight == other.forward_weight)
+        if (forward_segment_id.enabled && other.forward_segment_id.enabled &&
+            forward_segment_id.id == other.forward_segment_id.id &&
+            GetForwardWeightPlusOffset() == other.GetForwardWeightPlusOffset())
         {
             return true;
         }
-        if (reverse_segment_id.enabled && other.reverse_segment_id.enabled && reverse_segment_id.id == other.reverse_segment_id.id && reverse_weight_offset == other.reverse_weight_offset && reverse_weight == other.reverse_weight)
+        if (reverse_segment_id.enabled && other.reverse_segment_id.enabled &&
+            reverse_segment_id.id == other.reverse_segment_id.id &&
+            GetReverseWeightPlusOffset() == other.GetReverseWeightPlusOffset())
         {
             return true;
         }
--- a/src/engine/plugins/match.cpp
+++ b/src/engine/plugins/match.cpp
@@ -33,22 +33,6 @@ void filterCandidates(const std::vector<util::Coordinate> &coordinates,
 {
     for (const auto current_coordinate : util::irange<std::size_t>(0, coordinates.size()))
     {
-        bool allow_uturn = false;
-
-        if (coordinates.size() - 1 > current_coordinate && 0 < current_coordinate)
-        {
-            double turn_angle =
-                util::coordinate_calculation::computeAngle(coordinates[current_coordinate - 1],
-                                                           coordinates[current_coordinate],
-                                                           coordinates[current_coordinate + 1]);
-
-            // sharp turns indicate a possible uturn
-            if (turn_angle <= 90.0 || turn_angle >= 270.0)
-            {
-                allow_uturn = true;
-            }
-        }
-
         auto &candidates = candidates_lists[current_coordinate];
         if (candidates.empty())
         {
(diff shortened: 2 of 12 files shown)


src/engine/routing_algorithms/map_matching.cpp
```diff
@@ -250,6 +268,44 @@ SubMatchingList mapMatching(SearchEngineData<Algorithm> &engine_working_data,
...
+                double precision = DEFAULT_GPS_PRECISION;
+                if (!trace_gps_precision.empty())
+                    precision = (*trace_gps_precision[t]);
+                auto distance = util::coordinate_calculation::haversineDistance(
+                    candidates_list[prev_unbroken_timestamp][s].phantom_node.location,
+                    trace_coordinates[t]);
+
+                // We can use a previous state as one of possible states if it is inside the
+                // precision range.
+                // We don't use this logic at the last state to except using previous point at the
```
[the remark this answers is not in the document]
Komzpa:
```
// We can use a previous state as one of possible states if it is inside the
// precision range, except for the last state in trace
```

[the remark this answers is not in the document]
gardster: fixed

src/engine/routing_algorithms/map_matching.cpp
```diff
@@ -248,8 +263,48 @@ SubMatchingList mapMatching(SearchEngineData<Algorithm> &engine_working_data,
...
+                double precision = DEFAULT_GPS_PRECISION;
+                if (!trace_gps_precision.empty())
+                    precision = (*trace_gps_precision[t]);
+                const auto distance = util::coordinate_calculation::haversineDistance(
+                    candidates_list[prev_unbroken_timestamp][s].phantom_node.location,
+                    trace_coordinates[t]);
+
+                // We can use a previous state as one of possible states if it is inside the
+                // precision range, except for the last state in trace
+                if (distance < 3 * precision && t < candidates_list.size() - 1)
```
Komzpa: is there a constant for 3 available?

src/engine/routing_algorithms/map_matching.cpp
```diff
@@ -248,8 +263,48 @@ SubMatchingList mapMatching(SearchEngineData<Algorithm> &engine_working_data,
...
+                                        return node.phantom_node.IsIndistinct(tmpNode.phantom_node);
+                                    }))
+                        continue;
+
+                    candidates_list[t].push_back(tmpNode);
+                    candidates_list[t].back().phantom_node.input_location = trace_coordinates[t];
+                    candidates_list[t].back().distance = distance;
+                    map_matching::EmissionLogProbability emission_log_probability(precision);
+
+                    // We don't add the transition probability because staying is prefered than
```
Komzpa: if we aren't adding high probability, isn't it going to be a low-probability state, thus not preferred one?

include/engine/phantom_node.hpp
```diff
@@ -127,6 +127,23 @@ struct PhantomNode
 
     bool operator==(const PhantomNode &other) const { return location == other.location; }
 
+    bool IsIndistinct(const PhantomNode &other) const
+    {
+        if (forward_segment_id.enabled && other.forward_segment_id.enabled &&
```
Komzpa: why is only Forward *or* Reverse enough?

TheMarex:
Just a heads up, I'm in the process of reviewing this, but it might take me a while (need to think through an alternative). In general this is a great feature, but I have some concerns for the implementation.

Thanks for working on this! 👍

Komzpa:
Hey @TheMarex, we've deployed this branch locally and it shows a sudden performance degradation. @gardster struggles to find the cause, as in synthetic tests it shows itself well. Any hypothesis to check are welcome :)

oxidase:
@Komzpa could you share some matching requests that show performance degradation?

gardster:
We don't know why it degrade yet. This logic must not be affected in case of our usage because we decimate points to be not closer than 60 meters.

src/engine/routing_algorithms/map_matching.cpp
```diff
@@ -220,13 +221,27 @@ SubMatchingList mapMatching(SearchEngineData<Algorithm> &engine_working_data,
                         continue;
                     }
 
+                    PhantomNode prev_node = prev_unbroken_timestamps_list[s].phantom_node;
```
TheMarex: A phantom node is quite big (~64 bytes), a copy in the hot loop of the code here will degrade performance significantly. I'm not even sure why you need this code: We don't allow u-turns by default, we only enable it when adjacent segments in the trace form a very sharp angle.

src/engine/routing_algorithms/map_matching.cpp
```diff
@@ -248,8 +263,48 @@ SubMatchingList mapMatching(SearchEngineData<Algorithm> &engine_working_data,
...
                         current_pruned[s_prime] = false;
                         model.breakage[t] = false;
+                        current_segment_id[s_prime] = last_id;
                     }
                 }
+                // Here we can add previous states to the current viterbi if we suppose, that user
+                // can stay
+                double precision = DEFAULT_GPS_PRECISION;
+                if (!trace_gps_precision.empty())
+                    precision = (*trace_gps_precision[t]);
```
TheMarex: You can move this evaluation one level up to save on de-referencing the optional GPS trace precision.

src/engine/routing_algorithms/map_matching.cpp
TheMarex: I think we should try to move all of this logic out of the actual Viterbi loop. There is a high chance that the vectors will be modified multiple times per timestamp. That means a reallocation of `std::vector` is triggered and a lot of data is being copied. I suspect this is the major source of the slowdown you see.

To fix this, we can move the logic to the `src/engine/plugins/match.cpp` code that sets up the candidates. Actually this code is working against [this code](https://github.com/Project-OSRM/osrm-backend/blob/master/src/engine/plugins/match.cpp#L58-L71) which filters out these "stay" states.

So an equivalent implementation would be to remove that filter step, since the candidates that the R-Tree returns are already according to your criteria (`distance < 3 * precision`). This should be as fast or faster then the current implementation. There might be a few superfluous calls to `getNetworkDistance()`, but I think we could add a special case check to [that function](https://github.com/Project-OSRM/osrm-backend/blob/master/src/engine/routing_algorithms/routing_base_ch.cpp#L160) for `start == target` that skips some of the overhead of inserting the nodes in the heaps and calculates the distance only use the `PhantomNode`.

include/engine/phantom_node.hpp
gardster: Filtering can decouple forward and reverse edges into different candidates to avoid u-turns. And switch "enable" to false. See [this filter!](https://github.com/Project-OSRM/osrm-backend/blob/master/src/engine/plugins/match.cpp#L90-L99)
In the other hand I suppose that different requests to the r-tree index with same point will return same forward and backward results. So I check forward segment id and backward segment id separately.

gardster:
Thank you for the review and your suggestions, @TheMarex. I extracted the additional PhantomNodes search into the match.cpp file. Please take another look.

gardster:
Unfortunately we faced performance degradation with this PR. Snapping requests become ~2x time slower on the car profile.

src/engine/plugins/match.cpp
```diff
@@ -193,6 +169,36 @@ Status MatchPlugin::HandleRequest(const RoutingAlgorithmsInterface &algorithms,
 
     auto candidates_lists = GetPhantomNodesInRange(facade, tidied.parameters, search_radiuses);
 
+    // Add previous states for stand_in_place behavior
+    for (const auto current_index :
```
TheMarex: I think you could get rid of this step and it would still include `PhantomNodes` from the previous step. They key the changed `IsIndistinct, that not only checks for ID but also for the offset. So if there are segments that are probable for both candidates they should be included in the result from both RTree queries.

src/engine/routing_algorithms/map_matching.cpp
```diff
@@ -220,27 +219,33 @@ SubMatchingList mapMatching(SearchEngineData<Algorithm> &engine_working_data,
...
-
-                    // get distance diff between loc1/2 and locs/s_prime
-                    const auto d_t = std::abs(network_distance - haversine_distance);
-
-                    // very low probability transition -> prune
-                    if (d_t >= max_distance_delta)
+                    PhantomNode prev_node = prev_unbroken_timestamps_list[s].phantom_node;
+                    double network_distance = 0;
+                    // do not calculate transition probability in case without movement
+                    if (!current_timestamps_list[s_prime].phantom_node.IsIndistinct(prev_node))
```
TheMarex: That is a good optimization. 👍

src/engine/plugins/match.cpp
gardster: I think I did not fully understand your idea.
How do we obtain the PhantomNodes representing the previous states?

I am talking about this situation: lets suppose we have a noisy GPS signal ABCD along the single edge of the road graph. 
```
A______B
     C<______D
================
Z    X Y     W   
```
If the entire path is one edge, and the distance between the points is large enough, then `GetPhantomNodesInRange` returns the states of Z Y X W. To visit these states, the algorithm will have to add to the route the movement in the opposite direction Y-X, and in the resulting geometry this section of the road will be processed 3 times. My idea is that we need to add neighboring states. So that the point C can lead to the state Y. Then the resulting sequence on the edge becomes ZYYW. This addition is the responsibility of this code.

TheMarex: > If the entire path is one edge

Ah okay yes that is indeed a problem, I did not think of this edge case - I think we need to rework some of the modeling to make this actually work algorithmically.

Right now we represent states as `(edge_based_node_id, segment_id, weight_offset)`. The problem with this representation is that the `weight_offset` is a continuous number so there are infinitely many states per edge segment that we could encounter. In your example above this would look something like this:

Z : (0, 0, 0.0)
X : (0, 0, 0.45)
Y : (0, 0, 0.5)
W: (0, 0, 0.9)

For the input coordinate `C` the R-Tree will only give us the state `X` since its the nearest point on the edge segment. The main problem here is that we would never consider `Y` to be the "true" origin of `C`, hence leading to a very low transition probability from `B` to `C`.

To work around this problem we could try to make the states discrete. If we have a long edge segment, we could try to split it into small `sub-segments` and only snap the candidate states to these. Then for every candidate we find using the R-Tree query, we don't just add all candidate segments, but also all sub-segments in range.

1. Normal R-Tree query right now will identify all segments in the search radius.
2. For every candidate segment, compute the segment length:
    - If the segment length is below the threshold snap to one of the endpoints.
    - If the segment length is above the threshold, sample points along the segment and insert each as candidate.

By making this discrete, the candidates for segment coordinate `C` would include all candidates for segment `B` that are in query distance.

I need to think about this a little more but I think you found one big modeling problem, you would run into when using very high-precision traces.

DennisOSRM:

Issue #4685, opened 2017-11-14
Use vtzero for generating tiles


joto:
There is a new library for reading and writing vector tiles at https://github.com/mapbox/vtzero. The library is not quite finished yet, but we should start to think about using it in OSRM. I'll create a pull request for this.

PR #4686, opened 2017-11-14
Experimental version of tile plugin using vtzero.

joto:
This uses the new vtzero library (https://github.com/mapbox/vtzero)
instead of using protozero directly. The code becomes much cleaner this way and the generated tiles are smaller. Note that I have rewritten the code but only did a little bit of ad-hoc testing. To run this you'll need the newest release of protozero and vtzero master installed.

# Issue

#4685 

## Tasklist
 - [ ] ADD OWN TASKS HERE
 - [ ] review
 - [ ] adjust for comments

## Requirements / Relations

Might need vendored in vtzero. The protozero vendered in with libosmium will need to be updated. Don't merge until there is an official released version of vtzero.

/cc @springmeyer


springmeyer:
Thanks for getting this going @joto. I agree that this should wait to merge for an official vtzero release. At the same time, would be great if @danpat could find a moment to kick the tires on this to provide any feedback now that might help guide any needed changes in vtzero before official release.

TheMarex:
Thanks this looks great, 300 LOC less 🎉 Looking forward to the vtzero release.

oxidase:
@joto @danpat I have updated tile plugin unit tests. Please could you review the PR?

oxidase:
@TheMarex debug tiles look good
screenshot from 2018-04-24 10-46-04

PR #4693, opened 2017-11-20
Allow single edge paths in MLD alternatives, #4691

### LINKED ISSUE #4691: Undefined behaviour in MLD alternatives
In `alternativePathSearch` method [weighted_packed_paths](https://github.com/Project-OSRM/osrm-backend/blob/a53794f8645df783129fd48bda7a9c1e1df072fd/src/engine/routing_algorithms/alternative_path_mld.cpp#L758) is have empty paths if the middle node is also the starting node, so in debug mode there is an [assertion](https://github.com/Project-OSRM/osrm-backend/blob/a53794f8645df783129fd48bda7a9c1e1df072fd/src/engine/routing_algorithms/alternative_path_mld.cpp#L323) and ub in release mode.

The issue can be reproduced with the OSM export https://www.openstreetmap.org/export#map=15/56.9103/14.5320, updated weights from [crash.txt](https://github.com/Project-OSRM/osrm-backend/files/1480606/crash.txt) file and query `curl 'http://localhost:5000/route/v1/driving/14.5303422%2C56.9087067%3B14.52962589263916%2C56.90761947631836.json?geometries=polyline&alternatives=true&steps=true&overview=full&annotations=true'`

Values in the file are weights but not rates, so to reproduce the issue the line
https://github.com/Project-OSRM/osrm-backend/blob/a53794f8645df783129fd48bda7a9c1e1df072fd/src/updater/updater.cpp#L192
must be changed to
```
        const auto weight = rate;
```

oxidase:
# Issue

The PR  fixes #4691, by allowing alternative routes with `packed.path.empty()`. 
Also an additional check `node != fst.GetData(node).parent` was added to prevent infinite loops if `has_plateaux_at_node(node, fst, snd)` and `node` is the first (last) node of the route.

## Tasklist
 - [x] review
 - [x] adjust for comments

### DIFF
--- a/cucumber.js
+++ b/cucumber.js
@@ -1,6 +1,6 @@
 module.exports = {
-    default: '--strict --tags ~@stress --tags ~@todo --require features/support --require features/step_definitions',
-    verify: '--strict --tags ~@stress --tags ~@todo -f progress --require features/support --require features/step_definitions',
+    default: '--strict --tags ~@stress --tags ~@todo --tags ~@mld-only --require features/support --require features/step_definitions',
+    verify: '--strict --tags ~@stress --tags ~@todo --tags ~@mld-only -f progress --require features/support --require features/step_definitions',
     todo: '--strict --tags @todo --require features/support --require features/step_definitions',
     all: '--strict --require features/support --require features/step_definitions',
     mld: '--strict --tags ~@stress --tags ~@todo --require features/support --require features/step_definitions -f progress'
--- a/features/testbot/alternative_loop.feature
+++ b/features/testbot/alternative_loop.feature
@@ -31,40 +31,32 @@ Feature: Alternative route
             | 5    | 6  | dc,ca,ab,bd,dc,dc |             |
             | 7    | 8  | ca,ab,bd,dc,ca,ca |             |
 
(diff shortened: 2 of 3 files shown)


oxidase:
@daniel-j-h such alternatives are ok for CH, so it is better to keep them also in MLD. This requires  edges vector  emptiness  checks in all filters that use packed paths.

oxidase:
Unfortunately CH alternatives results are not consistent wrt OS and the test fails on OSX and Windows builds

oxidase:
@daniel-j-h i looked at the test fail on windows build. The problem in a loop shortcut that is not handled for CH alternative routes in `computeWeightAndSharingOfViaPath`  so the alternative path weight instead of 7 at https://github.com/Project-OSRM/osrm-backend/blob/a53794f8645df783129fd48bda7a9c1e1df072fd/src/engine/routing_algorithms/alternative_path_ch.cpp#L308 is the weight of the complete loop 6.8+7=13.8 and is filtered out. I did not find the reason why the loop shortcut is added with  an msvc build. Both issues are completely out scope of the PR, so i added `@mld-only` tag to skip the test for CH.

PR #4717, opened 2017-12-04
Refactor segregated intersection classification to right module

TheMarex:
# Issue

Drive-by refactor while investigating the code, just shuffles code around.

## Tasklist
 - [x] review
 - [x] adjust for comments

## Requirements / Relations

### DIFF
--- a/include/extractor/guidance/segregated_intersection_classification.hpp
+++ b/include/extractor/guidance/segregated_intersection_classification.hpp
@@ -20,8 +20,8 @@ namespace guidance
 // - middle edges between two osm ways in one logic road (U-turn)
 // - staggered intersections (X-cross)
 // - square/circle intersections
-std::unordered_set<EdgeID> findSegregatedNodes(const NodeBasedGraphFactory &factory, const util::NameTable& names);
+std::unordered_set<EdgeID> findSegregatedNodes(const NodeBasedGraphFactory &factory,
+                                               const util::NameTable &names);
 }
-
 }
 }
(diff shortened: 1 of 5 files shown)


include/extractor/guidance/segregated_intersection_classification.hpp
```diff
@@ -0,0 +1,27 @@
...
+class NodeBasedGraphFactory;
+
+namespace guidance
+{
+// Find all "segregated" edges, e.g. edges that can be skipped in turn instructions.
+// The main cases are:
+// - middle edges between two osm ways in one logic road (U-turn)
+// - staggered intersections (X-cross)
+// - square/circle intersections
+std::unordered_set<EdgeID> findSegregatedNodes(NodeBasedGraphFactory &factory, const util::NameTable& names);
```
daniel-j-h: Should factory be const? If not why not? Not clear from interface.

src/extractor/guidance/segregated_intersection_classification.cpp
```diff
@@ -0,0 +1,244 @@
...
+    RoadPriorityClass::Enum road_priority_class;
+
+    struct LessName
+    {
+        bool operator()(EdgeInfo const &e1, EdgeInfo const &e2) const { return e1.name < e2.name; }
+    };
+};
+
+bool IsSegregated(std::vector<EdgeInfo> v1,
+                  std::vector<EdgeInfo> v2,
```
daniel-j-h: Not sure why we copy vectors here

include/extractor/guidance/segregated_intersection_classification.hpp
TheMarex: Good catch, turns out the interface of `NodeBasedGraphFactory` did not apply const correctly.

src/extractor/guidance/segregated_intersection_classification.cpp
TheMarex: I think its because they are being sorted. Looks like the call site passes in the vectors by rvalue, so this should not be a problem.

Issue #4719, opened 2017-12-05
Configurable max search radius


ghoshkaj:
Similar to `max_locations` for the `nearest` service, we should be able to configure the max search radius as an engine config parameter.

daniel-j-h:
Tasks:
- [ ] add max radius to the engine config [here](https://github.com/Project-OSRM/osrm-backend/blob/353829a4cc3c060ce5e58bbaf5ffc94b6f5cb265/include/engine/engine_config.hpp#L82-L91)
- [ ] add cmd line option to osrm-routed [here](https://github.com/Project-OSRM/osrm-backend/blob/353829a4cc3c060ce5e58bbaf5ffc94b6f5cb265/src/tools/routed.cpp#L119-L136)
- [ ] expose max radius to node bindings [here](https://github.com/Project-OSRM/osrm-backend/blob/353829a4cc3c060ce5e58bbaf5ffc94b6f5cb265/include/nodejs/node_osrm_support.hpp#L179) and [here](https://github.com/Project-OSRM/osrm-backend/blob/353829a4cc3c060ce5e58bbaf5ffc94b6f5cb265/src/nodejs/node_osrm.cpp#L75-L80)
- [ ] check in plugin like we do with max radius [here](https://github.com/Project-OSRM/osrm-backend/blob/353829a4cc3c060ce5e58bbaf5ffc94b6f5cb265/src/engine/plugins/nearest.cpp#L33-L40)

daniel-j-h:
Oh actually duplicate of https://github.com/Project-OSRM/osrm-backend/issues/4326 😅 Can you close both issues when you're done here?

ghoshkaj:
Eeeks! Yes! Will close both when I'm done!

Issue #4326, opened 2017-07-23
Provide system wide limit for max radius


daniel-j-h:
We should allow users to set a system-wide limit on the radiuses option. In case any of the user-provided radiuses are too large we should return the status code `TooBig`.

Tasks:
- add to [engine config](https://github.com/Project-OSRM/osrm-backend/blob/58811d8f5c244927abd314f0bbf6bd4ce52462d1/include/engine/engine_config.hpp#L85-L90)
- expose in [node bindings](https://github.com/Project-OSRM/osrm-backend/blob/58811d8f5c244927abd314f0bbf6bd4ce52462d1/include/nodejs/node_osrm_support.hpp#L179)
- set from cmdline in [routed](https://github.com/Project-OSRM/osrm-backend/blob/58811d8f5c244927abd314f0bbf6bd4ce52462d1/src/tools/routed.cpp#L112-L129)

cc @TheMarex

PR #4721, opened 2017-12-05
Configurable max search radius

ghoshkaj:
# Issue

https://github.com/Project-OSRM/osrm-backend/issues/4719 and https://github.com/Project-OSRM/osrm-backend/issues/4326

## Tasklist
 - [x] review
 - [x] adjust for comments

## Requirements / Relations
~Currently based off of: https://github.com/Project-OSRM/osrm-backend/pull/4604~ Based off of Master

### DIFF
--- a/include/engine/engine_config.hpp
+++ b/include/engine/engine_config.hpp
@@ -84,6 +84,7 @@ struct EngineConfig final
     int max_locations_viaroute = -1;
     int max_locations_distance_table = -1;
     int max_locations_map_matching = -1;
+    double max_radius_map_matching = -1.0;
     int max_results_nearest = -1;
     int max_alternatives = 3; // set an arbitrary upper bound; can be adjusted by user
     bool use_shared_memory = true;
--- a/include/nodejs/node_osrm_support.hpp
+++ b/include/nodejs/node_osrm_support.hpp
@@ -186,6 +186,8 @@ inline engine_config_ptr argumentsToEngineConfig(const Nan::FunctionCallbackInfo
         params->Get(Nan::New("max_locations_map_matching").ToLocalChecked());
     auto max_results_nearest = params->Get(Nan::New("max_results_nearest").ToLocalChecked());
     auto max_alternatives = params->Get(Nan::New("max_alternatives").ToLocalChecked());
+    auto max_radius_map_matching =
+        params->Get(Nan::New("max_radius_map_matching").ToLocalChecked());
 
     if (!max_locations_trip->IsUndefined() && !max_locations_trip->IsNumber())
     {
@@ -233,6 +235,9 @@ inline engine_config_ptr argumentsToEngineConfig(const Nan::FunctionCallbackInfo
         engine_config->max_results_nearest = static_cast<int>(max_results_nearest->NumberValue());
     if (max_alternatives->IsNumber())
         engine_config->max_alternatives = static_cast<int>(max_alternatives->NumberValue());
+    if (max_radius_map_matching->IsNumber())
+        engine_config->max_radius_map_matching =
+            static_cast<double>(max_radius_map_matching->NumberValue());
 
     return engine_config;
 }
--- a/src/engine/plugins/match.cpp
+++ b/src/engine/plugins/match.cpp
@@ -139,6 +139,17 @@ Status MatchPlugin::HandleRequest(const RoutingAlgorithmsInterface &algorithms,
         return Error("InvalidValue", "Invalid coordinate value.", json_result);
     }
 
+    if (max_radius_map_matching > 0 && std::any_of(parameters.radiuses.begin(),
+                                                   parameters.radiuses.end(),
+                                                   [&](const auto &radius) {
+                                                       if (!radius)
+                                                           return false;
+                                                       return *radius > max_radius_map_matching;
+                                                   }))
+    {
+        return Error("TooBig", "Radius search size is too large for map matching.", json_result);
+    }
+
     // Check for same or increasing timestamps. Impl. note: Incontrast to `sort(first,
     // last, less_equal)` checking `greater` in reverse meets irreflexive requirements.
     const auto time_increases_monotonically = std::is_sorted(
--- a/src/nodejs/node_osrm.cpp
+++ b/src/nodejs/node_osrm.cpp
@@ -76,8 +76,9 @@ NAN_MODULE_INIT(Engine::Init)
  * @param {Number} [options.max_locations_viaroute] Max. locations supported in viaroute query (default: unlimited).
  * @param {Number} [options.max_locations_distance_table] Max. locations supported in distance table query (default: unlimited).
  * @param {Number} [options.max_locations_map_matching] Max. locations supported in map-matching query (default: unlimited).
+ * @param {Number} [options.max_radius_map_matching] Max. radius size supported in map matching query (default: 5).
  * @param {Number} [options.max_results_nearest] Max. results supported in nearest query (default: unlimited).
- * @param {Number} [options.max_alternatives] Max.number of alternatives supported in alternative routes query (default: 3).
+ * @param {Number} [options.max_alternatives] Max. number of alternatives supported in alternative routes query (default: 3).
  *
  * @class OSRM
  *
--- a/src/tools/routed.cpp
+++ b/src/tools/routed.cpp
@@ -133,7 +133,10 @@ inline unsigned generateServerProgramOptions(const int argc,
          "Max. results supported in nearest query") //
         ("max-alternatives",
          value<int>(&config.max_alternatives)->default_value(3),
-         "Max. number of alternatives supported in the MLD route query");
+         "Max. number of alternatives supported in the MLD route query") //
+        ("max-matching-radius",
+         value<double>(&config.max_radius_map_matching)->default_value(5),
+         "Max. radius size supported in map matching query");
 
     // hidden options, will be allowed on command line, but will not be shown to the user
     boost::program_options::options_description hidden_options("Hidden options");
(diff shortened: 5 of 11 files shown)


include/engine/engine_config.hpp
```diff
@@ -84,6 +84,7 @@ struct EngineConfig final
     int max_locations_viaroute = -1;
     int max_locations_distance_table = -1;
     int max_locations_map_matching = -1;
+    boost::optional<double> max_radius_map_matching = -1;
```
daniel-j-h: Should this be a plain double similarly to how other limits above are plain ints? You could initialize with `std::numeric_limits<double>::max()` if you're concerned about the unlimited case.

src/engine/plugins/match.cpp
```diff
@@ -139,6 +139,15 @@ Status MatchPlugin::HandleRequest(const RoutingAlgorithmsInterface &algorithms,
         return Error("InvalidValue", "Invalid coordinate value.", json_result);
     }
 
+    std::for_each(parameters.radiuses.begin(), parameters.radiuses.end(), [&](auto &radius) {
+        // enforce maximum size of radius in radiuses for performance reasons
+        if (max_radius_map_matching > 0.0 && radius > max_radius_map_matching)
```
daniel-j-h: Why the `> 0` check here? How do we encode "unlimited"?

include/nodejs/node_osrm_support.hpp
```diff
@@ -249,6 +256,9 @@ inline engine_config_ptr argumentsToEngineConfig(const Nan::FunctionCallbackInfo
         engine_config->max_results_nearest = static_cast<int>(max_results_nearest->NumberValue());
     if (max_alternatives->IsNumber())
         engine_config->max_alternatives = static_cast<int>(max_alternatives->NumberValue());
+    if (max_radius_map_matching->IsNumber())
+        engine_config->max_radius_map_matching =
+            static_cast<boost::optional<double>>(max_radius_map_matching->NumberValue());
```
daniel-j-h: Not sure the cast is legal.. I think you want to use `boost::make_optional` here (or change to plain double as mentioned above).

src/nodejs/node_osrm.cpp
```diff
@@ -77,7 +77,8 @@ NAN_MODULE_INIT(Engine::Init)
  * @param {Number} [options.max_locations_distance_table] Max. locations supported in distance table query (default: unlimited).
  * @param {Number} [options.max_locations_map_matching] Max. locations supported in map-matching query (default: unlimited).
  * @param {Number} [options.max_results_nearest] Max. results supported in nearest query (default: unlimited).
- * @param {Number} [options.max_alternatives] Max.number of alternatives supported in alternative routes query (default: 3).
+ * @param {Number} [options.max_alternatives] Max. number of alternatives supported in alternative routes query (default: 3).
+ * @param {Number} [options.max_radius_map_matching] Max. radius size supported in map matching query (default: 5).
```
daniel-j-h: Should we group by service? That is move the max map matching radius up to max locations map matching.

src/tools/routed.cpp
```diff
@@ -12,6 +12,7 @@
 #include <boost/algorithm/string/case_conv.hpp>
 #include <boost/any.hpp>
 #include <boost/filesystem.hpp>
+#include <boost/optional/optional_io.hpp>
```
daniel-j-h: Why do we need `optional_io.hpp` here? I think you want the `optional.hpp` header from boost; `optional_io.hpp` is just for printing an optional to stdout.

src/engine/plugins/match.cpp
```diff
@@ -139,6 +139,15 @@ Status MatchPlugin::HandleRequest(const RoutingAlgorithmsInterface &algorithms,
...
+        {
+            return Error(
+                "TooBig", "Radius search size is too large for map matching.", json_result);
+        }
```
daniel-j-h: This does not compile - you're inside the `for_each`'s lambda and are trying to return an error from the lambda here.

What you can do instead of your `for_each` is check with `any_of` if any radius is invalid and if so return an error.

http://en.cppreference.com/w/cpp/algorithm/all_any_none_of

unit_tests/server/parameters_parser.cpp
```diff
@@ -559,6 +557,33 @@ BOOST_AUTO_TEST_CASE(valid_match_urls)
...
+    reference_2.radiuses = {7, 60};
+    auto result_2 = parseParameters<MatchParameters>("1,2;3,4?radiuses=seventy;60");
+    BOOST_CHECK(!result_2);
+    CHECK_EQUAL_RANGE(reference_2.timestamps, result_2->timestamps);
+    CHECK_EQUAL_RANGE(reference_2.bearings, result_2->bearings);
+    BOOST_CHECK(reference_2.radiuses != result_2->radiuses);
+    CHECK_EQUAL_RANGE(reference_2.approaches, result_2->approaches);
+    CHECK_EQUAL_RANGE(reference_2.coordinates, result_2->coordinates);
+}
+
```
daniel-j-h: What are we testing here? You didn't change the grammar or parsing, so I think this test is unnecessary.

Instead what do you think of adding ana dditional limit test here: https://github.com/Project-OSRM/osrm-backend/blob/ad7b99ec5c535ca9ec03a1ea50653e1fb8898cd8/unit_tests/library/limits.cpp#L107-L132

include/engine/engine_config.hpp
```diff
@@ -84,6 +84,7 @@ struct EngineConfig final
     int max_locations_viaroute = -1;
     int max_locations_distance_table = -1;
     int max_locations_map_matching = -1;
+    double max_radius_map_matching = 5;
```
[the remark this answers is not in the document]
ghoshkaj: If radius search is unlimited, doesn't it take a long time to calculate the response for map matching? And isn't that why we're adding this configurability?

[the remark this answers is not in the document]
daniel-j-h: Only if the user requests a large radius. It's the same with limits above, e.g. max locations distance table is -1 (unlimited) here but then the binary sets a limit and allows the OSRM user to change it.

include/nodejs/node_osrm_support.hpp
```diff
@@ -217,6 +219,11 @@ inline engine_config_ptr argumentsToEngineConfig(const Nan::FunctionCallbackInfo
         Nan::ThrowError("max_alternatives must be an integral number");
         return engine_config_ptr();
     }
+    if (!max_radius_map_matching->IsUndefined() && !max_radius_map_matching->IsNumber())
+    {
+        Nan::ThrowError("max_radius_map_matching must be an integral number");
```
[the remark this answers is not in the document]
ghoshkaj: Ohh, you're right. Does `IsNumber()` return true from floating point number too? I can't find the definition of this online or in the code. Perhaps I missed it?

unit_tests/server/parameters_parser.cpp
ghoshkaj: @daniel-j-h ooo I was just trying to check the case of passing in a non-number. It wasn't a well thought out test-case, rather an attempt to get code coverage for [this case](https://github.com/Project-OSRM/osrm-backend/pull/4721/files/ad7b99ec5c535ca9ec03a1ea50653e1fb8898cd8#diff-89873aae5c93116cd0f09f2e1a69c320R222)

ghoshkaj: Let me try with the limit test!!

include/nodejs/node_osrm_support.hpp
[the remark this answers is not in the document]
daniel-j-h: Docs for v8 are at https://v8docs.nodesource.com

JavaScript only has floating point'ish numbers (not [IEEE 754](https://en.wikipedia.org/wiki/IEEE_754) floating points like the `float` and `double` type in C++):

http://www.ecma-international.org/ecma-262/6.0/#sec-number-objects

If you search for `NumberValue` you will find

https://v8docs.nodesource.com/node-8.0/dc/d0a/classv8_1_1_value.html#a8ea9eba2ebb0e2dd99d0c10fcec86d13

returning a `double`.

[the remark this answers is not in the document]
ghoshkaj: Ah it's in Node land, I was searching in C++ land. Thank you for the docs.

include/engine/engine_config.hpp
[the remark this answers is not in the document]
ghoshkaj: Okay, so it's basically up to the user to not request something stupid. We just have to enable it. Got it.

src/engine/plugins/match.cpp
```diff
@@ -139,6 +139,15 @@ Status MatchPlugin::HandleRequest(const RoutingAlgorithmsInterface &algorithms,
         return Error("InvalidValue", "Invalid coordinate value.", json_result);
     }
 
+    if (std::any_of(parameters.radiuses.begin(), parameters.radiuses.end(), [&](auto &radius) {
```
daniel-j-h: Can be `const auto radius&` because we only want a read-only view

ghoshkaj:
@daniel-j-h is this good to be merged in now? 😄

test/nodejs/match.js
```diff
@@ -68,6 +68,7 @@ test('match: match in Monaco with geometry compression', function(assert) {
 
 test('match: match in Monaco with speed annotations options', function(assert) {
     assert.plan(12);
+    // var osrm = new OSRM({path: data_path, max_locations_map_matching: 2);
```
daniel-j-h: Remove comment - or should this test a limit of 2 locations?

ghoshkaj: Ah forgot to push the removal of this comment. Pushing now!

PR #4722, opened 2017-12-06
Bugfix ? Avoid-Set should take precedence over Speeds

quiqua:
# Issue

When specifying road classes in the `Avoid Set` of a profile, they will not be taken into consideration as long as speeds for those classes are defined.

Example:
```
avoid = Set {
  'trunk'
},

speeds = Sequence {
  highway = {
    motorway        = 90,
    motorway_link   = 45,
    trunk           = 85,
    ...
  }
}
```

With this PR, a new `WayHandler` is added to simply discard the road class if it is defined in the set.

Note: The same could be accomplished by just removing the speed entries for the specific roads, so I am not quite sure if this PR would actually fix something.

## Tasklist
 - [x] review
 - [x] adjust for comments

### DIFF
--- a/profiles/car.lua
+++ b/profiles/car.lua
@@ -356,6 +356,7 @@ function process_way(profile, way, result, relations)
     -- routable. this includes things like status=impassable,
     -- toll=yes and oneway=reversible
     WayHandlers.blocked_ways,
+    WayHandlers.avoid_ways,
 
     -- determine access status by checking our hierarchy of
     -- access tags, e.g: motorcar, motor_vehicle, vehicle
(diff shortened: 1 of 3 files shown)


profiles/car.lua
```diff
@@ -356,6 +356,7 @@ function process_way(profile, way, result, relations)
     -- routable. this includes things like status=impassable,
     -- toll=yes and oneway=reversible
     WayHandlers.blocked_ways,
+    WayHandlers.avoid_ways,
```
TheMarex: Good catch. I think we never caught that since the default entries for `avoid` don't have speeds assigned.

Issue #4669, opened 2017-11-08
Collapse legs in map matching response


danpat:
The `match` plugin currently returns a `RouteLeg` object for every pair of matched coordinates.

If you supply a fairly detailed GPS trace (e.g. coordinates every 25m), you end up with a `Route` object with lots of very very short `RouteLegs` - each usually having just a `depart` and a `arrive` `RouteStep`.

We could add a parameter to the `match` plugin, like `waypoints=0,24` that indicates which of supplied coordinates are to be used as waypoints.  Currently, *all* coordinates are used as waypoints, leading to all the legs.

If a trace is split and waypoints are supplied, we should return an error that no route can be found.

TheMarex:
We would need to change how we construct the `InternalRouteResult` objects [here](https://github.com/Project-OSRM/osrm-backend/blob/master/src/engine/plugins/match.cpp#L221).

The easiest implementation just merges the legs after the fact.

ericrwolfe:
> The easiest implementation just merges the legs after the fact.

If we had a generalized solution to merge legs, that might also help implement "via" points.

coxchapman:
>If you supply a fairly detailed GPS trace (e.g. coordinates every 25m),

When do we do this? What's the urgency or need here?

coxchapman:
Capturing some context from chat: this would support refreshing ETAs along a routeline. We would use GPS to match you back to the routeline, and then provide a new ETA.

This could be helpful as part of a wider conversation on ETAs (projecting future traffic, live ETA updates, etc) but we need to focus on time-to-deploy and other tech metrics right now.

danpat:
🤔 `waypoints=0,24` could also work for the `route` plugin - would enable some control of routes without emitting legs for every pair.

TheMarex:
Some more context on why it would be beneficial to do this before generating the actual API object:

In guidance post-processing we collapse several instructions based on the context of the path. The simple example is a round-a-bout. Let's say you have a leg that ends in a roundabout, simply concatenating the instructions of the current and the next leg will give you different instructions and would potentially break the exit counting. The client would need to adjust these values after merging the legs.

coxchapman:
@danpat @TheMarex have we made a decision on the right place and scope to do this?

karenzshea:
It could make sense to implement this right now behind a `waypoints=x,y` parameter, and then in 6.0 make the default behavior to treat the first and last matched coordinates as the waypoints, unless otherwise specified.

[...]

PR #4732, opened 2017-12-11
Implement `waypoints` parameter in match plugin

karenzshea:
# Issue

Closes https://github.com/Project-OSRM/osrm-backend/issues/4669

This adds a parameter to the match service plugin, `&waypoints`, that will allow users to specify which input coordinates should be treated as `waypoints` in the response. Current behavior is to treat all input coordinates as waypoints. Implementation details [here](https://github.com/Project-OSRM/osrm-backend/issues/4669#issuecomment-349706768) 

## Tasklist
 - [x] Write unit tests for parameter
   - [x] CollapseInternalRouteResult
   - [x] valid map matching parameters
   - [x] invalid map matching parameters
   - [x] integration tests map matching
 - [x] Add support for map matching results in cucumber support code
 - [ ] ~Debug why `waypoints` crashes when trace splitting happens~ Don't support trace splitting and waypoints together
 - [x] Add waypoints handling into tracepoint generation
 - [x] Add parameter parsing to match plugin interface
 - [x] Add handling into trace tidying
 - [x] write documentation
 - [ ] review
 - [ ] adjust for comments

## Requirements / Relations
nerp

### DIFF
--- a/include/engine/api/match_parameters.hpp
+++ b/include/engine/api/match_parameters.hpp
@@ -63,25 +63,40 @@ struct MatchParameters : public RouteParameters
                           RouteParameters::GeometriesType::Polyline,
                           RouteParameters::OverviewType::Simplified,
                           {}),
-          gaps(GapsType::Split), tidy(false)
+          gaps(GapsType::Split), tidy(false), waypoints()
     {
     }
 
     template <typename... Args>
     MatchParameters(std::vector<unsigned> timestamps_, GapsType gaps_, bool tidy_, Args... args_)
+        : MatchParameters(std::move(timestamps_), gaps_, tidy_, {}, std::forward<Args>(args_)...)
+    {
+    }
+
+    template <typename... Args>
+    MatchParameters(std::vector<unsigned> timestamps_,
+                    GapsType gaps_,
+                    bool tidy_,
+                    std::vector<std::size_t> waypoints_,
+                    Args... args_)
         : RouteParameters{std::forward<Args>(args_)...}, timestamps{std::move(timestamps_)},
-          gaps(gaps_), tidy(tidy_)
+          gaps(gaps_), tidy(tidy_), waypoints{std::move(waypoints_)}
     {
     }
 
     std::vector<unsigned> timestamps;
     GapsType gaps;
     bool tidy;
+    std::vector<std::size_t> waypoints;
 
     bool IsValid() const
     {
+        const auto valid_waypoints =
+            std::all_of(waypoints.begin(), waypoints.end(), [this](const auto &w) {
+                return w < coordinates.size();
+            });
         return RouteParameters::IsValid() &&
-               (timestamps.empty() || timestamps.size() == coordinates.size());
+               (timestamps.empty() || timestamps.size() == coordinates.size()) && valid_waypoints;
     }
 };
 }
--- a/include/engine/api/match_parameters_tidy.hpp
+++ b/include/engine/api/match_parameters_tidy.hpp
@@ -37,6 +37,9 @@ struct Result
     Mask can_be_removed;
     // Maps the MatchParameter's original items to items which should not be removed.
     Mapping tidied_to_original;
+    // Masking the MatchParameter coordinates for items whose indices were present in the
+    // `waypoints` parameter.
+    Mask was_waypoint;
 };
 
 inline Result keep_all(const MatchParameters &params)
@@ -44,6 +47,15 @@ inline Result keep_all(const MatchParameters &params)
     Result result;
 
     result.can_be_removed.resize(params.coordinates.size(), false);
+    result.was_waypoint.resize(params.coordinates.size(), true);
+    if (!params.waypoints.empty())
+    {
+        for (const auto p : params.waypoints)
+        {
+            result.was_waypoint.set(p, false);
+        }
+        result.was_waypoint.flip();
+    }
     result.tidied_to_original.reserve(params.coordinates.size());
     for (std::size_t current = 0; current < params.coordinates.size(); ++current)
     {
@@ -61,6 +73,8 @@ inline Result keep_all(const MatchParameters &params)
         {
             result.parameters.coordinates.push_back(params.coordinates[i]);
 
+            if (result.was_waypoint[i])
+                result.parameters.waypoints.push_back(result.parameters.coordinates.size() - 1);
             if (!params.hints.empty())
                 result.parameters.hints.push_back(params.hints[i]);
 
@@ -74,6 +88,8 @@ inline Result keep_all(const MatchParameters &params)
                 result.parameters.timestamps.push_back(params.timestamps[i]);
         }
     }
+    if (params.waypoints.empty())
+        result.parameters.waypoints.clear();
 
     return result;
 }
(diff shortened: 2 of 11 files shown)


include/engine/api/match_parameters.hpp
```diff
@@ -63,25 +63,34 @@ struct MatchParameters : public RouteParameters
...
                           RouteParameters::OverviewType::Simplified,
                           {}),
-          gaps(GapsType::Split), tidy(false)
+          gaps(GapsType::Split), tidy(false), waypoints()
     {
     }
 
     template <typename... Args>
-    MatchParameters(std::vector<unsigned> timestamps_, GapsType gaps_, bool tidy_, Args... args_)
+    MatchParameters(std::vector<unsigned> timestamps_,
```
TheMarex: We need to preserve the original constructor here to preserve backwards compatibility.

include/engine/api/match_parameters.hpp
```diff
@@ -63,25 +63,34 @@ struct MatchParameters : public RouteParameters
...
     }
 
     std::vector<unsigned> timestamps;
     GapsType gaps;
     bool tidy;
+    std::vector<std::size_t> waypoints;
 
     bool IsValid() const
     {
+        auto invalid_idx = std::find_if(waypoints.begin(), waypoints.end(), [this](const auto &w) {
```
TheMarex: We can use [`std:all_of`](http://www.cplusplus.com/reference/algorithm/all_of/) instead.

include/engine/api/match_parameters_tidy.hpp
```diff
@@ -157,7 +183,22 @@ inline Result tidy(const MatchParameters &params, Thresholds cfg = {15., 5})
...
         }
+        else
+        {
+            // one of the coordinates meant to be used as a waypoint was marked for removal
+            // update the original waypoint index to the new representative coordinate
+            const auto last_idx = result.parameters.coordinates.size() - 1;
+            if (result.was_waypoint[i] && (result.parameters.waypoints.back() != last_idx))
+            {
+                result.parameters.waypoints.push_back(last_idx);
+                // result.was_waypoint[i] = false;
```
TheMarex: Commented out code should be removed.

include/engine/api/match_parameters_tidy.hpp
```diff
@@ -157,7 +183,22 @@ inline Result tidy(const MatchParameters &params, Thresholds cfg = {15., 5})
...
+            }
+        }
     }
+    // todo: this is not always true... e.g. if the first or last coordinate is duplicated and then
+    // tidied away
+    // but we do want some kind of check like this
+    // BOOST_ASSERT(result.parameters.waypoints.size() == params.waypoints.size());
```
TheMarex: Hrm I don't understand this comment or the assertion. Can we specify this more precisely?

include/engine/internal_route_result.hpp
```diff
@@ -102,6 +102,54 @@ struct InternalManyRoutesResult
...
+    BOOST_ASSERT(is_waypoint.back()); // should always be waypoints
+    // Nothing to collapse! return result as is
+    if (leggy_result.unpacked_path_segments.size() == 1)
+        return leggy_result;
+
+    BOOST_ASSERT(leggy_result.segment_end_coordinates.size() > 1);
+
+    InternalRouteResult collapsed;
+    collapsed.shortest_path_weight = leggy_result.shortest_path_weight;
+    for (std::size_t i = 0; i < leggy_result.unpacked_path_segments.size(); i++)
```
TheMarex: We can use `util::irange<size_t>` here.

include/engine/internal_route_result.hpp
```diff
@@ -102,6 +102,54 @@ struct InternalManyRoutesResult
...
+        }
+        else
+        // no new leg, collapse the next segment into the last leg
+        {
+            auto &last_segment = collapsed.unpacked_path_segments.back();
+            // deduplicate last segment (needs to be checked for empty for the same node query edge
+            // case)
+            if (!last_segment.empty())
+                last_segment.pop_back();
+            // update target phantom node of leg
```
TheMarex: Needs an assertion that `segment_end_coordinates` is not empty.

src/engine/plugins/match.cpp
```diff
@@ -218,6 +228,27 @@ Status MatchPlugin::HandleRequest(const RoutingAlgorithmsInterface &algorithms,
...
+    // Error: Check if user-supplied waypoints can be found in the resulting matches
+    for (const auto waypoint : tidied.parameters.waypoints)
+    {
+        bool found = false;
+        std::for_each(sub_matchings.begin(), sub_matchings.end(), [&](const SubMatching &sm) {
+            auto index = std::find(sm.indices.begin(), sm.indices.end(), waypoint);
+            if (index != sm.indices.end())
+                found = true;
+        });
+        if (!found)
```
TheMarex: Needs brackets around body.

src/engine/plugins/match.cpp
```diff
@@ -218,6 +228,27 @@ Status MatchPlugin::HandleRequest(const RoutingAlgorithmsInterface &algorithms,
...
     }
 
+    // trace was split, we don't support the waypoints parameter across multiple match objects
+    if (sub_matchings.size() > 1 && !parameters.waypoints.empty())
+    {
+        return Error("NoMatch", "Could not match the trace with the given waypoints.", json_result);
+    }
+
+    // Error: Check if user-supplied waypoints can be found in the resulting matches
+    for (const auto waypoint : tidied.parameters.waypoints)
```
TheMarex: The complexity of this check is too high (three nested loops). What about using a `std::set<std::size_t>' that contains the waypoint indices. Then we could loop over all sub-matching indices and remove the index of the waypoint we found.

src/engine/plugins/match.cpp
```diff
@@ -234,12 +265,25 @@ Status MatchPlugin::HandleRequest(const RoutingAlgorithmsInterface &algorithms,
...
+        // we split the phantom nodes anyway and only have bi-directional phantom nodes for
+        // possible uturns
         sub_routes[index] =
             algorithms.ShortestPathSearch(sub_routes[index].segment_end_coordinates, {false});
         BOOST_ASSERT(sub_routes[index].shortest_path_weight != INVALID_EDGE_WEIGHT);
+        if (!tidied.parameters.waypoints.empty())
+        {
+            std::vector<bool> waypoint_legs;
+            waypoint_legs.reserve(sub_matchings[index].nodes.size());
+            for (unsigned i = 0; i < sub_matchings[index].nodes.size(); ++i)
```
TheMarex: I think we can do this in a single sweep over the indices of the submatch:
- Allocate a vector of size #waypoints 
- Iterate over indices and `parameters.waypoints` at the same time (they are both sorted increasing)
- Set `waypoints_legs[i]` to true if we find a match.

include/engine/api/match_parameters_tidy.hpp
karenzshea: Hm, I actually don't think this needs to be assertion anymore, since it's not always true. I think having test coverage over this is good enough for now.

src/engine/plugins/match.cpp
```diff
@@ -229,6 +240,29 @@ Status MatchPlugin::HandleRequest(const RoutingAlgorithmsInterface &algorithms,
...
+    if (sub_matchings.size() > 1 && !parameters.waypoints.empty())
+    {
+        return Error("NoMatch", "Could not match the trace with the given waypoints.", json_result);
+    }
+
+    // Error: Check if user-supplied waypoints can be found in the resulting matches
+    {
+        std::set<std::size_t> tidied_waypoints(tidied.parameters.waypoints.begin(),
+                                               tidied.parameters.waypoints.end());
+        std::for_each(sub_matchings.begin(), sub_matchings.end(), [&](const SubMatching &sm) {
```
TheMarex: Instead of `for_each` you can use `for (const auto& sm : sub_matchings) { for (auto index : sm.indices) {} }` which looks a little cleaner.

PR #4756, opened 2017-12-23
Fix too restrictive no name check

oxidase:
# Issue

8bce06169135da839682c23ac345660d9cfa53ca changes check `unnamed = road_name_id == EMPTY_NAMEID` to `unnamed = name_table.GetNameForID(road_name_id).to_string().empty()`  that is too restrictive and leads to a regression in sliproads handling if the target road has only a reference
like http://map.project-osrm.org/?z=18&center=52.156971%2C9.874869&loc=52.156038%2C9.872396&loc=52.157559%2C9.875057&hl=en&alt=0

PR removes the `unnamed` check. The proper check for empty names and references is in `util::guidance::requiresNameAnnounced`. Also PR removes premature conversions of string views to strings.

## Tasklist
 - [x] review
 - [x] adjust for comments

## Requirements / Relations

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,8 +1,15 @@
 # UNRELEASED
+  - Changes from 5.14.2:
+    - Bugfixes:
+      - FIXED #4754: U-Turn penalties are applied to straight turns.
+      - FIXED #4756: Removed too restrictive road name check in the sliproad handler
+
+# 5.14.2
   - Changes from 5.14.1:
     - Bugfixes:
       - FIXED #4727: Erroring when a old .core file is present.
-      - FIXED #4754: U-Turn penalties are applied to straight turns.
+      - FIXED #4642: Update checks for EMPTY_NAMEID to check for empty name strings
+      - FIXED #4738: Fix potential segmentation fault
     - Node.js Bindings:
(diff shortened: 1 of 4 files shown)


include/util/guidance/name_announcements.hpp
```diff
@@ -26,7 +26,8 @@ namespace guidance
 // Name Change Logic
 // Used both during Extraction as well as during Post-Processing
 
-inline std::string longest_common_substring(const std::string &lhs, const std::string &rhs)
+template <typename StringView>
+inline StringView longest_common_substring(const StringView &lhs, const StringView &rhs)
```
TheMarex: Why do we need a template type for the string view here?

CHANGELOG.md
```diff
@@ -1,7 +1,14 @@
 # UNRELEASED
+  - Changes from 5.14.2:
+    - Bugfixes:
+      - FIXED #4756: Removed too restrictive road name check in the sliproad handler
+
+# 5.14.2
```
TheMarex: Thanks for backporting! 🙇

include/util/guidance/name_announcements.hpp
oxidase: it is not needed. Changed to the explicit type `utill::StringView`

PR #4762, opened 2018-01-02
Use consistent EBG node weights in duplicated via nodes

### LINKED ISSUE #4731: Incorrect weight values for duplicated edge-based nodes.
Seems like it's a bug here:
https://github.com/Project-OSRM/osrm-backend/blob/2e97c78181d294fe9d1d358df323457e6189ecbf/src/extractor/edge_based_graph_factory.cpp#L378

Push back weight to EBN-indexed weight's vector by getting the value from the same vector using NBE-index. This indices are different according to nbe_to_ebn_mapping vector in EdgeBasedGraphFactory.

Why not to do simple:
```
m_edge_based_node_weights.push_back(edge_data.weight);
```
?

oxidase:
# Issue

PR fixes #4731 as  `m_edge_based_node_weights` must be indexed with `nbe_to_ebn_mapping[eid]`.

Additional assertion checks internal consistency of duplicated EBG nodes weights and NBG edges weights 
( EBG node weight can be set to `INVALID_EDGE_WEIGHT` at [`InsertEdgeBasedNode`](
https://github.com/Project-OSRM/osrm-backend/blob/17f25ccbc3687171a91ca2c9422e68e3e035fd07/src/extractor/edge_based_graph_factory.cpp#L134) before expanding via-way turn restrictions).

## Tasklist
 ~- [ ] add regression / cucumber cases (see docs/testing.md)~
 - [x] review
 - [ ] adjust for comments

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
     - Bugfixes:
       - FIXED #4754: U-Turn penalties are applied to straight turns.
       - FIXED #4756: Removed too restrictive road name check in the sliproad handler
+      - FIXED #4731: Use correct weights for edge-based graph duplicated via nodes.
 
 # 5.14.2
   - Changes from 5.14.1:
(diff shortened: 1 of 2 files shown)


vng:
LGTM

PR #4764, opened 2018-01-02
Belarus speed limits

Komzpa:
Speed limits in Belarus currently match the ones in Russia.

https://en.wikipedia.org/wiki/Speed_limits_in_Belarus


oxidase:
:+1: thanks! Please add new tags to `taginfo.json` after https://github.com/Project-OSRM/osrm-backend/blob/a5e0d7011b0357b540810f7f67eef62caed3b43c/taginfo.json#L149 and add a line to unreleased section of ChangeLog file like
```
    - Profile:
      - CHANGED: added Belarus speed limits
```

PR #4774, opened 2018-01-05
Restore use of user-provided osmium::thread::Pool instance

mloskot:
# Issue

`Pool` instance has been removed from `Reader` ctor parameters list in PR #4438, presumably unintentionally.

It is required to prevent potential deadlock during `Pool` shutdown as explained in PR #4452.

## Tasklist
 - [ ] review

## Requirements / Relations

* Regression introduced in  #4438

### DIFF
--- a/src/extractor/extractor.cpp
+++ b/src/extractor/extractor.cpp
@@ -365,7 +365,7 @@ Extractor::ParseOSMData(ScriptingEnvironment &scripting_environment,
     TIMER_START(parsing);
 
     { // Parse OSM header
-        osmium::io::Reader reader(input_file, osmium::osm_entity_bits::nothing);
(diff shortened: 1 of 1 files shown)


oxidase:
Good catch! Please could you add `pool` parameter at two other places https://github.com/Project-OSRM/osrm-backend/blob/master/src/extractor/extractor.cpp#L548 and https://github.com/Project-OSRM/osrm-backend/blob/master/src/extractor/extractor.cpp#L555?

mloskot:
@oxidase Good catch! I missed the `Reader` additions. Done. Thanks.

mloskot:
Updated to fix the clang-format complain (Travis CI build failure)

Issue #4667, opened 2017-11-07
Cannot find route between Britain and Isle of Man


Mekapaedia:
I've been playing around with OSRM to make large trips and routes and it has been working excellently - I can build long multistage routes of 120+ waypoints throughout Eurasia without issue.

However, I have an oddly specific issue that goes beyond my code ([I've checked with the OSRM demo map as well](http://map.project-osrm.org/?z=9&center=53.799029%2C-3.249207&loc=53.405472%2C-2.980539&loc=54.149774%2C-4.477902&hl=en&alt=0)) - OSRM cannot find a route between Britain and the Isle of Man. I thought it might be an issue with the underlying OSM data (no connection) but [Graphhopper-based](https://www.mapquest.com/directions/from/great-britain/liverpool-282414276/to/isle-of-man/douglas-282411021) [services](https://graphhopper.com/maps/?point=Liverpool%2C%20United%20Kingdom&point=Douglas%2C%20Isle%20of%20Man&locale=en-GB&vehicle=car&weighting=fastest&elevation=true&use_miles=false&layer=Omniscale) can find a route and correctly return the transit time. I've also used [the OSM inspector and it does have motorcar=yes](https://www.openstreetmap.org/way/22772385#map=9/53.7966/-3.3646) in all the ferry routes (4 or so) from Britain to the Isle of Man.

I thought since Graphhopper doesn't process turn restrictions it might be this so I turned it off in the profile (I'm using the standard car.lua with slightly higher U-turn cost) but it still doesn't find a path.

I'm also having an issue where there is a bridge on the A591 that OSRM deems impassable, but I think that's something to do with my profile as the demo server doesn't have this issue (not related but not a issue I think).

danpat:
The problem here is that the road the ferry connects to in Douglas is marked as `highway=service` and `access=destination`.

The `car.lua` profile generally eliminates roads with `access=` tags, *unless* they're on the `profile.restricted_highway_whitelist` - and `highway=service` is not.

The reason for this is ... unclear to me.  Anyone happen to remember why we don't have service roads on the `restricted_highway_whitelist` ?

oxidase:
@danpat it is similar to #4632, so  `service` tag was not included in the first commit cc3a4899a2ee3837a80f171ef482568118ce18c4

Mekapaedia:
[...]

PR #4788, opened 2018-01-09
Allow restricted access routing on highway=service roads

karenzshea:
# Issue

Closes https://github.com/Project-OSRM/osrm-backend/issues/4667

There is a [whitelist](https://github.com/Project-OSRM/osrm-backend/blob/a8f3474996af9390eb7166dddd5f4c62bb0ae4c1/profiles/car.lua#L156-L170) of highway types on which we allow restricted access routing, i.e. roads that you can route onto if the route starts/ends on the road, but not as a shortcut through. (https://github.com/Project-OSRM/osrm-backend/pull/3721). 

This PR makes a profile adjustment that adds `highway=service` roads onto the whitelist of roads we will end or start a route on when it's tagged with [a restricted access tag](https://github.com/Project-OSRM/osrm-backend/blob/a8f3474996af9390eb7166dddd5f4c62bb0ae4c1/profiles/car.lua#L82-L87).

Anecdotally speaking this will allow:
- a ferry route to be found between Isle of Man and Liverpool where the end of the ferry leads directly onto a dock/pier road of some kind tagged with both `highway=service` and `access=destination`. 
- routing people who really need help within parking lots that have been tagged with any combination of `highway=service` and `access=destination|customer|private`

## Tasklist
-  [x] Do some manual checks around airport service road tagging
 - [x] review
 - [x] adjust for comments


unit_tests/library/route.cpp
```diff
@@ -69,13 +69,13 @@ BOOST_AUTO_TEST_CASE(test_route_same_coordinates_fixture)
                                                    json::Object{{
                                                        {"location", location},
                                                        {"bearing_before", 0},
-                                                       {"bearing_after", 58},
+                                                       {"bearing_after", 238},
```
karenzshea: This seems like a strange change ... I wouldn't think that a change to access tags in the profiles would result in different bearings in this tests?

karenzshea:
### First change: allowing restricted access routing onto `highway=service` roads
[Westhafenstraße](https://www.openstreetmap.org/way/178594661) is tagged `highway=service` and `access=destination`

image
We allow routing onto it now...

image
... but not through it even if it would be "faster"

### Second change: disallowing restricted access routing onto `highway=service` `access=private` roads
... doesn't work yet

unit_tests/library/route.cpp
TheMarex: Hrm maybe it snaps to a different road now?

karenzshea:
Update ...

We currently do not support routing onto ways with the tag combination `highway=service` `access=private` and want to preserve part of this behavior, while allowing restricted access routing on other combinations of `highway=service` tagged roads.

The behavior we want to preserve is disallowing routes starting from `highway=service` `access=private` roads. This will require extra logic in the way parsing function that adds a boolean like `is_startpoint` or `can_start_from` that will later be used in the edge based graph factory to mark `highway=service` `access=private` roads as invalid start points.

karenzshea:
And disallowing routing onto access=private service roads!
image
image

features/car/destination.feature
```diff
@@ -131,3 +131,56 @@ Feature: Car - Destination only, no passing through
...
+            | cx    | private  | service |
+            | xd    | private  | service |
+            | de    |          | primary |
+
+        When I route I should get
+            | from | to | route     |
+            | a    | x  | acb,xd,xd |
+            | a    | d  | acb,xd,xd |
+            | a    | e  | acb,xd,de |
+            | x    | e  | de,de     |
```
TheMarex: A comment column would help, especially a comment that the last case depends on `x` snapping to `de`

TheMarex:
@karenzshea You seem to have some failing test cases that need to be investigated:

https://ci.appveyor.com/project/DennisOSRM/osrm-backend/build/1.0.9836#L16366

TheMarex:
Travis is just failing because of Mac OS X builds, merging this anyway.

Issue #4795, opened 2018-01-12
Changed waypoint index values for submatchings


oxidase:
Commit 55cc06fd8b6d11cc07097e4fe3dde31cdea9c905 changes `waypoint_index` values of tracepoints in submatchings:

* before
```
curl -s 'localhost:5000/match/v1/driving/-6.806995868682861,62.00034059427156;-6.804485321044921,62.000733464253614;-6.80275797843933,62.001106182373675;-6.8042707443237305,62.00376544418236;-6.803541183471679,62.00369493644869;-6.8034982681274405,62.00335750432094?timestamps=0;1;2;1700;1701;1702' | jq '.tracepoints[] | "matching " + (.matchings_index|tostring) + ", waypoint " + (.waypoint_index|tostring)'
"matching 0, waypoint 0"
"matching 0, waypoint 1"
"matching 0, waypoint 2"
"matching 1, waypoint 0"
"matching 1, waypoint 1"
"matching 1, waypoint 2"
```

* after 
```
curl -s 'localhost:5000/match/v1/driving/-6.806995868682861,62.00034059427156;-6.804485321044921,62.000733464253614;-6.80275797843933,62.001106182373675;-6.8042707443237305,62.00376544418236;-6.803541183471679,62.00369493644869;-6.8034982681274405,62.00335750432094?timestamps=0;1;2;1700;1701;1702' | jq '.tracepoints[] | "matching " + (.matchings_index|tostring) + ", waypoint " + (.waypoint_index|tostring)'
"matching 0, waypoint 0"
"matching 0, waypoint 1"
"matching 0, waypoint 2"
"matching 1, waypoint 3"
"matching 1, waypoint 4"
"matching 1, waypoint 5"
```

/cc @karenzshea

blackvvine:
I have the same problem in my build from commit [55cc06f](https://github.com/Project-OSRM/osrm-backend/commit/55cc06fd8b6d11cc07097e4fe3dde31cdea9c905).

PR #4799, opened 2018-01-15
Fix waypoint_index adjustment

karenzshea:
# Issue

Closes https://github.com/Project-OSRM/osrm-backend/issues/4795

## Tasklist
 - [x] add regression test that checks tracepoint field in api response on split submatch responses
 - [x] review
 - [x] adjust for comments

## Requirements / Relations
5.15.1 release candidate cc @oxidase

### DIFF
--- a/include/engine/api/match_api.hpp
+++ b/include/engine/api/match_api.hpp
@@ -97,6 +97,8 @@ class MatchAPI final : public RouteAPI
             }
         }
 
+        BOOST_ASSERT(parameters.waypoints.empty() || sub_matchings.size() == 1);
+
         std::size_t was_waypoint_idx = 0;
         for (auto trace_index : util::irange<std::size_t>(0UL, parameters.coordinates.size()))
         {
--- a/include/engine/api/match_api.hpp
+++ b/include/engine/api/match_api.hpp
@@ -120,6 +120,7 @@ class MatchAPI final : public RouteAPI
                 sub_matchings[matching_index.sub_matching_index]
                     .alternatives_count[matching_index.point_index];
             // waypoint indices need to be adjusted if route legs were collapsed
+            // waypoint parameter assumes there is only one match object
             if (!parameters.waypoints.empty())
             {
                 if (tidy_result.was_waypoint[trace_index])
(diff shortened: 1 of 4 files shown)


unit_tests/library/match.cpp
```diff
@@ -64,4 +64,53 @@ BOOST_AUTO_TEST_CASE(test_match)
...
+            const auto matchings_index =
+                waypoint_object.values.at("matchings_index").get<json::Number>().value;
+            const auto waypoint_index =
+                waypoint_object.values.at("waypoint_index").get<json::Number>().value;
+            const auto &route_legs = matchings[matchings_index]
+                                         .get<json::Object>()
+                                         .values.at("legs")
+                                         .get<json::Array>()
+                                         .values;
+            BOOST_CHECK_LT(waypoint_index, route_legs.size() + 1);
```
TheMarex: This is a bad check, since it would still pass in the following cases:

- All `waypoint_index` values are the same like `0, 0, 0, 0`
- The order of values of `waypoint_index` is random `0, 2, 1, 4`
- There could be holes like `0, 1, 5, 6`

The check should actually check that the value for `waypoint_index` is really ascending like `0, 1, 2, 3`.

include/engine/api/match_api.hpp
```diff
@@ -115,17 +115,23 @@ class MatchAPI final : public RouteAPI
...
+            // waypoint parameter assumes there is only one match object
+            if (!parameters.waypoints.empty())
             {
-                waypoint.values["waypoint_index"] = was_waypoint_idx;
-                was_waypoint_idx++;
-            }
-            else
-            {
-                waypoint.values["waypoint_index"] = util::json::Null();
+                if (tidy_result.was_waypoint[trace_index])
```
TheMarex: In this case we should have an assertion here that there is only one sub-matching, since that is the assumption that is used here.

oxidase:
@karenzshea @TheMarex i have added a commit with review updates to unblock 5.15.0-rc.2 
Travis CI build https://travis-ci.org/Project-OSRM/osrm-backend/builds/330065569

PR #4812, opened 2018-01-18
add tunnel as a class in lua

ghoshkaj:
# Issue

[Notifications for Tunnels, Toll stations, etc.](https://github.com/Project-OSRM/osrm-backend/issues/3028)

## Tasklist
 - [x] add `tunnel` as a class in `car.lua`
 - [x] add handling for `tunnel` class in `way_handler.lua`
 - [x] CHANGELOG.md entry ([How to write a changelog entry](http://keepachangelog.com/en/1.0.0/#how))
 - [x] add regression / cucumber cases (see docs/testing.md)
 - [x] review
 - [x] adjust for comments

PR #4816, opened 2018-01-22
Set type of trivial intersections where classes change to Suppressed

oxidase:
# Issue

In some cases like tunnels or bridges way classes may change at trivial intersections of degree two.
Maneuvers at such intersections have turn types `NoTurn` and intersections will be removed from  response intersections arrays. 

## Tasklist
 - [x] review
 - [x] adjust for comments

## Requirements / Relations
#4812

### DIFF
--- a/features/car/classes.feature
+++ b/features/car/classes.feature
@@ -119,6 +119,25 @@ Feature: Car - Mode flag
             | from | to | route       | turns                                      | classes                   |
             | a    | d  | ab,bc,cd,cd | depart,new name right,new name left,arrive | [()],[(tunnel)],[()],[()] |
 
+    Scenario: Car - We tag classes without intersections
+        Background:
+            Given a grid size of 200 meters
+
+        Given the node map
+            """
+            a b c d
+            """
+
+        And the ways
+            | nodes | name | tunnel |
+            | ab    | road |        |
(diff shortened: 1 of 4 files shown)


features/car/classes.feature
```diff
@@ -119,6 +119,25 @@ Feature: Car - Mode flag
             | from | to | route       | turns                                      | classes                   |
             | a    | d  | ab,bc,cd,cd | depart,new name right,new name left,arrive | [()],[(tunnel)],[()],[()] |
 
+    Scenario: Car - We tag tunnel with a class
+        Background:
```
ghoshkaj: Should we change the name of the scenario [above](https://github.com/Project-OSRM/osrm-backend/pull/4816/files#diff-4d1491a3addda1b6e121fd5f8961bebcR102)? Does it matter if we keep the same name for both scenarios?

oxidase: :+1: change to `Car - We tag classes without intersections`

ghoshkaj:
Looks good to me! 👍

oxidase:
@TheMarex in SF-area there are 147 new suppressed instructions https://gist.github.com/oxidase/9a78e927b78af1d796639e8dbdaa9acd mainly tunnels and restricted roads.

Issue #4763, opened 2018-01-02
Add support for non-numerical `maxheight` units 


oxidase:
Parsing `maxheight` does not include [non-numerical units](https://wiki.openstreetmap.org/wiki/Key:maxheight#Non-numerical_values).

The profile settings must include the default maximum height that must be used instead of ignoring height restrictions.

frodrigo:
I add it to my TODO list.

SiarheiFedartsou:
Seems to be done in https://github.com/Project-OSRM/osrm-backend/commit/2987292cc0b6ce1b36b1804c0efb07ff62d3625f

PR #4819, opened 2018-01-23
Add support for non-numerical units in car profile for maxhieght #4763

alain-andre:
Fixes the issue #4763


frodrigo:
We hesitated to use `way:get_location_tag('country')`. But we does not have values for lots of countries and 4.5 as default value seam a fair value.

Nevertheless we can look at `way:get_location_tag('country')`, should we have to provide another geojson file ? and keep 4.5 as fallback value ?

TheMarex:
> Nevertheless we can look at way:get_location_tag('country'), should we have to provide another geojson file ? and keep 4.5 as fallback value ?

Yeah, we would need to add another geojson file with country borders. I think @oxidase experimented a while ago with sourcing them directly from OSM but the polygons were too detailed.

Maybe we could go a little bit more coarse here and go with polygons for the EU and the US? I did not find a good source for minimal bridge clearances in Europe, but it seems like the maximum truck height is given between 4m and 4.5m.

frodrigo:
We found this:
https://tunnels.piarc.org/en/operational-safety-requirements-geometry/vertical-clearance
https://tunnels.piarc.org/sites/tunnels/files/public/wysiwyg/import/Chapters%20PIARC%20reports/2001%2005.11.B%20Chap%207%20EN.pdf

4.5 is a good fallback value.
For few other countries we can add polygons.

alain-andre:
We added a `data/maxheight.geojson` containing countries having a maximum height greater than 4.5 based on the pdf @frodrigo found.

This new file contains a `maxheight` property associated to one or more polygons.

For example **Norway** and **Denmark** polygons are in the same feature containing a `maxheight` property of **4.60**

Here are the countries and there maxheight : 

 - Norway : 4.60
 - Denmark : 4.60
 - Austria : 4.70
 - USA : 4.90
 - Spain : 5.00
 - UK : 5.35

PR #4822, opened 2018-01-23
Left driving side

alain-andre:
Addition of left-hand drive countries.


TheMarex:
@alain-andre seems like you accidentally included a commit from your other PR. It would be great if you could split your commit from that. Thanks! 👍

alain-andre:
@TheMarex I have split the commit, sorry for that.

PR #4827, opened 2018-01-24
Fix #4808: network distance comparison

### LINKED ISSUE #4808: `map_matching_delta` may be used inappropriately
I was reviewing the map matching plugin and found a comparison that doesn't make sense to me.

The `max_distance_delta` is set to be the maximum distance that could possibly have been traveled during the current time step:

```
# (paraphrasing)
const auto max_distance_delta = step_time * facade.GetMapMatchingMaxSpeed();
```

And `d_t` is the absolute difference between the great-circle distance and the network distance. 

```
const auto d_t = std::abs(network_distance - haversine_distance);
```

Then we [compare the two](https://github.com/Project-OSRM/osrm-backend/blob/master/src/engine/routing_algorithms/map_matching.cpp#L236) to throw out very low probability transitions:

```
                    // very low probability transition -> prune
                    if (d_t >= max_distance_delta)
                    {
                        continue;
                    }
```

I think this is wrong; if the actual speed was close to the maximum, we might still consider transitions that have a speed 2x the facade's maximum speed.

Instead, maybe compare `max_distance_delta` to `network_distance`? If the network distance implies travel over the maximum speed, then it should be thrown out.

ezheidtmann:
# Issue

Ref. #4808 

This commit fixes the comparison identified in #4808. I also rename `max_distance_delta` to `max_network_distance`.

With this renaming, the initialization of `weight_upper_bound` (touched in this commit) doesn't look quite right. Maybe it should just be `max_network_distance / 4.`, but I'm not sure what that means either.

## Tasklist
 - [x] Confirm that `weight_upper_bound` is still sensible
 - [ ] review
 - [ ] adjust for comments

## Requirements / Relations

### DIFF
--- a/src/engine/routing_algorithms/map_matching.cpp
+++ b/src/engine/routing_algorithms/map_matching.cpp
@@ -156,7 +156,7 @@ SubMatchingList mapMatching(SearchEngineData<Algorithm> &engine_working_data,
             }
         }();
 
-        const auto max_distance_delta = [&] {
+        const auto max_network_distance = [&] {
             if (use_timestamps)
             {
                 return step_time * facade.GetMapMatchingMaxSpeed();
@@ -199,9 +199,9 @@ SubMatchingList mapMatching(SearchEngineData<Algorithm> &engine_working_data,
 
             const auto haversine_distance = util::coordinate_calculation::haversineDistance(
                 prev_coordinate, current_coordinate);
-            // assumes minumum of 4 m/s
+            // assumes minumum of 5 km/h == 1.39 m/s
             const EdgeWeight weight_upper_bound =
-                ((haversine_distance + max_distance_delta) / 4.) * facade.GetWeightMultiplier();
+                (max_network_distance / 1.39) * facade.GetWeightMultiplier();
 
             // compute d_t for this timestamp and the next one
             for (const auto s : util::irange<std::size_t>(0UL, prev_viterbi.size()))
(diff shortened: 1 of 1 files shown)


ezheidtmann:
It looks like the `weight_upper_bound` initialization has some assumptions embedded in it that only make sense for a fastest-trip car profile. I don't know how to compute a more profile-aware value.

src/engine/routing_algorithms/map_matching.cpp
```diff
@@ -201,7 +201,7 @@ SubMatchingList mapMatching(SearchEngineData<Algorithm> &engine_working_data,
                 prev_coordinate, current_coordinate);
             // assumes minumum of 4 m/s
             const EdgeWeight weight_upper_bound =
-                ((haversine_distance + max_distance_delta) / 4.) * facade.GetWeightMultiplier();
+                ((haversine_distance + max_network_distance) / 4.) * facade.GetWeightMultiplier();
```
TheMarex: This would be only `max_network_distance` now, it basically assumes the weight encodes values in `second`.

ezheidtmann:
Gotcha, thanks. Revised commit changes `weight_upper_bound` as well. 

I'm concerned that this change will break some matchings for foot profile (default speed 1.9 m/s), but I suppose I can tackle that in another issue. I don't have a counter-example at hand.

TheMarex:
> I'm concerned that this change will break some matchings for foot profile (default speed 1.9 m/s)

That is a good point. 4 m/s is about 14.4 km/h which is even faster than a fair amount of roads driving with a car (living streets). I'm not sure were the original value came from something closer to 5 km/h would make more sense.

TheMarex:
I investigated our test failures, we seem to set a maximum map matching speed of `30 km/h` for the `testbot.lua` profile. That breaks most test cases, since now the pruning is actually effective.

I'm working on a fix for our test cases.

TheMarex:
Closing this in favor of https://github.com/Project-OSRM/osrm-backend/pull/4856 where I pulled in your commit.

PR #4829, opened 2018-01-26
Only run step collapsing based on original waypoints parameter 

karenzshea:
# Issue

There is a segfault producing bug in the new RouteLeg collapsing code in 5.15.0. The bug appears to only be triggered by map matching requests with the criteria:
- `tidy=true` parameter
- the map matched result has been split into submatches in the Match object

The actual point of failure for a request with the above criteria happens when the RouteLeg collapsing code attempts to collapse the `unpacked_path_segments` of a submatch into an empty vector.

However, this only happens when collapsing is run on a match with multiple submatches because one of the fundamental assumptions of leg collapsing is that the Match result is a continuous match. Fixing the actual issue here involves adding a more robust check that leg collapsing is never run on a match with multiple submatches.

## Tasklist
-  [ ] Debug the return of the `Could not match the trace with the given waypoints.` error when the `waypoints` param has not been passed in
 - [x] review
 - [x] adjust for comments
 - [ ] Make bug fix release

## Requirements / Relations
5.15.1

### DIFF
--- a/src/engine/plugins/match.cpp
+++ b/src/engine/plugins/match.cpp
@@ -162,6 +162,7 @@ Status MatchPlugin::HandleRequest(const RoutingAlgorithmsInterface &algorithms,
             "InvalidValue", "Timestamps need to be monotonically increasing.", json_result);
     }
 
+    bool collapse_legs = false;
     SubMatchingList sub_matchings;
     api::tidy::Result tidied;
     if (parameters.tidy)
@@ -262,6 +263,8 @@ Status MatchPlugin::HandleRequest(const RoutingAlgorithmsInterface &algorithms,
                 "NoMatch", "Requested waypoint parameter could not be matched.", json_result);
         }
     }
+    // we haven't errored yet, only allow leg collapsing if we meet this criteria
+    collapse_legs = sub_matchings.size() == 1 && !parameters.waypoints.empty();
 
     // each sub_route will correspond to a MatchObject
     std::vector<InternalRouteResult> sub_routes(sub_matchings.size());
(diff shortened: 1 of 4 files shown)


src/engine/plugins/match.cpp
```diff
@@ -262,6 +264,8 @@ Status MatchPlugin::HandleRequest(const RoutingAlgorithmsInterface &algorithms,
                 "NoMatch", "Requested waypoint parameter could not be matched.", json_result);
         }
     }
+    // we haven't errored yet, only allow leg collapsing if we meet this criteria
+    collapse_legs = sub_matchings.size() == 1 && !parameters.waypoints.empty();
```
oxidase: Maybe it is better to simplify the condition here to `const auto collapse_legs = !parameters.waypoints.empty()` with the following assertion
```
BOOST_ASSERT(parameters.waypoints.empty() || sub_matchings.size() == 1);
const auto collapse_legs = !parameters.waypoints.empty();
```
similar to https://github.com/Project-OSRM/osrm-backend/blob/279c7510cb440e80177ddac1508ec5d13953b06c/include/engine/api/match_api.hpp#L100?

karenzshea: Makes sense to me — collapsing should only be contingent upon the original input value of `waypoints`, and the error handling here https://github.com/Project-OSRM/osrm-backend/blob/5756e75c30d5f18d0a03920b44a04e70a5b65e26/src/engine/plugins/match.cpp#L244-L248 should handle the case of a split trace and requested waypoints.

PR #4832, opened 2018-01-26
Announce ref changes if road names are empty

### LINKED ISSUE #4830: New name suppression ignores ref changes
@dgearhart reported that, in [this route](http://map.project-osrm.org/?z=15&center=40.268802%2C-76.679592&loc=40.242299%2C-76.708537&loc=40.285722%2C-76.651079&hl=en&alt=0), OSRM doesn’t appear to issue a “continue straight” or “keep left” instruction at [the Ephrata exit](http://www.openstreetmap.org/node/666724455) along U.S. 322 East near Hershey, Pennsylvania.

steps

OSRM assumes that, because the branch to the right is a `motorway_link` as opposed to a `motorway`, this junction is an ordinary highway exit rather than a fork. The change of `ref` from `US 322` to `US 422` at this point should cause OSRM to emit a `new name` instruction, but I suspect that [`osrm::util::guidance::requiresNameAnnounced()`](https://github.com/Project-OSRM/osrm-backend/blob/a5e0d7011b0357b540810f7f67eef62caed3b43c/include/util/guidance/name_announcements.hpp#L108) returns `false` in this case:

* The `name` remains unchanged as an empty string on both the preceding and following `motorway` ways.
* [`obvious_change`](https://github.com/Project-OSRM/osrm-backend/blob/a5e0d7011b0357b540810f7f67eef62caed3b43c/include/util/guidance/name_announcements.hpp#L155-L158) is `true`
* because [`is_suffix_change`](https://github.com/Project-OSRM/osrm-backend/blob/a5e0d7011b0357b540810f7f67eef62caed3b43c/include/util/guidance/name_announcements.hpp#L145) is `true`
* because all four conditions of [`checkForPrefixOrSuffixChange()`](https://github.com/Project-OSRM/osrm-backend/blob/a5e0d7011b0357b540810f7f67eef62caed3b43c/include/util/guidance/name_announcements.hpp#L141-L142) are `true`
* because [`checkTable()`](https://github.com/Project-OSRM/osrm-backend/blob/a5e0d7011b0357b540810f7f67eef62caed3b43c/include/util/guidance/name_announcements.hpp#L137-L139) returns `true` on an empty string, and [`decompose()`](https://github.com/Project-OSRM/osrm-backend/blob/a5e0d7011b0357b540810f7f67eef62caed3b43c/include/util/guidance/name_announcements.hpp#L135) [returns a bunch of empty strings](https://github.com/Project-OSRM/osrm-backend/blob/a5e0d7011b0357b540810f7f67eef62caed3b43c/include/util/guidance/name_announcements.hpp#L82-L85) if [the longest common substring](https://github.com/Project-OSRM/osrm-backend/blob/a5e0d7011b0357b540810f7f67eef62caed3b43c/include/util/guidance/name_announcements.hpp#L66) is empty

/cc @oxidase

oxidase:
# Issue

Fixes #4830
screenshot from 2018-01-26 09-14-36

## Tasklist
 - [x] review
 - [ ] adjust for comments

## Requirements / Relations

### DIFF
--- a/include/util/guidance/name_announcements.hpp
+++ b/include/util/guidance/name_announcements.hpp
@@ -160,7 +160,9 @@ inline bool requiresNameAnnounced(const StringView &from_name,
     const auto needs_announce =
         // " (Ref)" -> "Name " and reverse
         (from_name.empty() && !from_ref.empty() && !to_name.empty() && to_ref.empty()) ||
-        (!from_name.empty() && from_ref.empty() && to_name.empty() && !to_ref.empty());
+        (!from_name.empty() && from_ref.empty() && to_name.empty() && !to_ref.empty()) ||
+        // ... or names are empty but reference changed
+        (names_are_empty && !ref_is_contained);
 
     const auto pronunciation_changes = from_pronunciation != to_pronunciation;
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
   - Changes from 5.15.0:
     - Bugfixes:
       - FIXED: Segfault in map matching when RouteLeg collapsing code is run on a match with multiple submatches
+    - Guidance
+      - CHANGED #4830: Announce reference change if names are empty
     - Profile:
       - FIXED: `highway=service` will now be used for restricted access, `access=private` is still disabled for snapping.
       - ADDED #4775: Exposes more information to the turn function, now being able to set turn weights with highway and access information of the turn as well as other roads at the intersection [#4775](https://github.com/Project-OSRM/osrm-backend/issues/4775)
--- a/features/car/summaries.feature
+++ b/features/car/summaries.feature
@@ -76,7 +76,7 @@ Feature: Basic Routing
 
         When I route I should get
             | waypoints | route  | summary  |
-            | a,c       | ,      | 100, 101 |
+            | a,c       | ,,     | 100, 101 |
 
      Scenario: Single Ref
         Given the node map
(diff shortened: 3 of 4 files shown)


dgearhart:
@oxidase The `Continue straight` looks great! Much better for the user to match reality

include/util/guidance/name_announcements.hpp
```diff
@@ -160,7 +160,9 @@ inline bool requiresNameAnnounced(const StringView &from_name,
     const auto needs_announce =
         // " (Ref)" -> "Name " and reverse
         (from_name.empty() && !from_ref.empty() && !to_name.empty() && to_ref.empty()) ||
-        (!from_name.empty() && from_ref.empty() && to_name.empty() && !to_ref.empty());
+        (!from_name.empty() && from_ref.empty() && to_name.empty() && !to_ref.empty()) ||
+        // ... or names are empty but reference changed
+        (names_are_empty && !ref_is_contained);
```
1ec5: 👍 I was concerned that the `ref_is_contained` check would be insufficient in cases where a route follows an exit and an auxiliary route continues straight. However, the interchanges at either end of [this route](https://www.openstreetmap.org/relation/1933421) (`I 40` versus `I 40 Business`) demonstrate that the suffixed cardinal directions prevent `ref_is_contained` from being true. (In this case, the main route exits as a `motorway`, so there’s a fork instruction anyways, but I’m sure there are cases where the main route exits as a `motorway_link`.)

PR #4844, opened 2018-02-01
Rename namespace partition to partitioner

mloskot:
# Issue

Rename module partition to partitioner.

This cultivates naming used in existing modules like extractor, customizer, etc. - noun vs verb (word partition is both though).

## Tasklist

 - [x] review
 - [x] adjust for comments

## Requirements / Relations

This is part of preparation to submit `osrm::partition` and `osrm::customize` facades to the library API as update related to #3787.

By the way, apart from introducing consistent naming in this PR, it solves problems in future: the original namespace `partition` is confusing with the name of upcoming function `osrm::partition` easily, if not surely, leading to compilation error like: `'osrm::partition': redefinition; previous definition was namespace`


mloskot:
My pleasure. Thanks for accepting 👍

PR #4849, opened 2018-02-02
Upgrade version of compilers and node on Travis CI

TheMarex:
We now build using GCC 7 and Clang 5 and only build binaries for node 8.x.

# Issue

We have not revisited our CI setup in a while. This removes a bunch of outdated jobs and bumps the compiler versions.

## Tasklist

 - [ ] review
 - [ ] adjust for comments


TheMarex:
@oxidase good point, added one for checking GCC 6.

PR #4856, opened 2018-02-06
Fix #4808: network distance comparison

### LINKED ISSUE #4808: `map_matching_delta` may be used inappropriately

TheMarex:
# Issue

Fixes issues #4808.
The distance pruning we previously used was applied to the wrong value and thus very optimistic. This replaces the PR by @ezheidtmann because we will need to adjust some test cases.

## Tasklist

 - [ ] review
 - [ ] benchmark on production logs
 - [ ] adjust for comments

## Requirements / Relations

Replaces https://github.com/Project-OSRM/osrm-backend/pull/4827

### DIFF
--- a/src/engine/routing_algorithms/map_matching.cpp
+++ b/src/engine/routing_algorithms/map_matching.cpp
@@ -156,7 +156,7 @@ SubMatchingList mapMatching(SearchEngineData<Algorithm> &engine_working_data,
             }
         }();
 
-        const auto max_distance_delta = [&] {
+        const auto max_network_distance = [&] {
             if (use_timestamps)
             {
                 return step_time * facade.GetMapMatchingMaxSpeed();
@@ -199,9 +199,9 @@ SubMatchingList mapMatching(SearchEngineData<Algorithm> &engine_working_data,
 
             const auto haversine_distance = util::coordinate_calculation::haversineDistance(
                 prev_coordinate, current_coordinate);
-            // assumes minumum of 4 m/s
+            // assumes minumum of 5 km/h == 1.39 m/s
             const EdgeWeight weight_upper_bound =
-                ((haversine_distance + max_distance_delta) / 4.) * facade.GetWeightMultiplier();
+                (max_network_distance / 1.39) * facade.GetWeightMultiplier();
 
             // compute d_t for this timestamp and the next one
             for (const auto s : util::irange<std::size_t>(0UL, prev_viterbi.size()))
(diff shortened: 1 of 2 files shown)


src/engine/routing_algorithms/map_matching.cpp
```diff
@@ -199,9 +199,9 @@ SubMatchingList mapMatching(SearchEngineData<Algorithm> &engine_working_data,
 
             const auto haversine_distance = util::coordinate_calculation::haversineDistance(
                 prev_coordinate, current_coordinate);
-            // assumes minumum of 4 m/s
+            // assumes minumum of 5 km/h == 1.39 m/s
             const EdgeWeight weight_upper_bound =
-                ((haversine_distance + max_distance_delta) / 4.) * facade.GetWeightMultiplier();
+                (max_network_distance / 1.39) * facade.GetWeightMultiplier();
```
oxidase: Here can be a potential problem if two timestamps are equal and `haversine_distance > 0`, so `step_time`,  `max_network_distance` and `weight_upper_bound` will be 0 and `getNetworkDistance` will fail to find a path, because `weight` must be strictly less than `weight_upper_bound`.

At the moment coordinates in traces with `haversine_distance=0` and `step_time=0` lead to a gap in trace.

TheMarex:
I think I need to go back to the drawing board here. The transition probability function and its interaction with the timestamps does not work. I remembered why I didn't like the original formulation in the paper, because it uses absolute difference of great circle distance and network distance. Using relative values would make much more sense when your sample rate is not fixed.

chaupow:
@TheMarex Do you need a rubber duck and/or a second brain here? If you want to explain someone how Mapmatching works, ping me.

TheMarex:
Punting on this until we implement https://github.com/Project-OSRM/osrm-backend/issues/4864 The problem is that we can't really calculate a good upper bound on the network distance itself right now.

Issue #1947, opened 2016-02-02
Implement mmapDataFacade


danpat:
OSRM currently supports reading data from files into heap memory (`InternalDataFacade`), or pre-loading data into shared memory using IPC shared memory blocks (`SharedDataFacade`+`osrm-datastore`).

We can consolidate the behaviour of both of these by using `mmap`.  Instead of reading files into memory explicitly, we should be able to `mmap` the data files, and immediately begin using them.

There are a few changes that need to be made to get us there:
- [x] Benchmark `mmap`d data access vs heap - what, if any, penalty is there? How does this change when the file we `mmap` is on a ramdisk?
- [x] Identify data structures that can't be `mmap`ed and fix them - basically anything in `osrm-datastore` (`src/storage/storage.cpp`) that isn't just loaded into memory in one big blob. Problem here is `vector<bool>` and its proxy behavior; we need a contiguous container we can `memcpy` to.
- [x] Clone the `SharedDataFacade` and perform similar `.swap` operations against `mmap`ed memory addresses rather than `shm` addresses.
- [ ] Figure out IPC signalling for swapping out `mmap`ed files on-the-fly
- [x] Investigate using `mmap` instead of explicit `read` disk files for leaf nodes in the StaticRTree to boost performance (coordinate lookups represent the largest part of any given routing query because of the I/O in the rtree).
- [ ] Make sure this works on Windows too.

The main goal here is to minimize double-reads of data.  In situations where we are constantly cycling out data sets (in the case of traffic updates), we want to minimize I/O and the number of times any single bit of data gets touched.  By using `mmap` and `tmpfs`, we can emulate the current share-memory behavior, but avoid an extra pass over the data.

For normal `osrm-routed` use, we would essentially get lazy-loading of data - `osrm-routed` would start up faster, but queries would be slower since pages are loaded from disk on demand until data is touched and lives in the filesystem cache. This initial slowness could be avoided by pre-seeding the data files into the filesystem cache or via `MAP_POPULATE` (Linux 2.5.46+), and this could be done in parallel to `osrm-routed` already starting up and answering queries.

/cc @daniel-j-h @TheMarex

TheMarex:
>  Benchmark mmapd data access vs heap - what, if any, penalty is there? How does this change when the file we mmap is on a ramdisk?

Benchmarks revealed a 10% slowdown w.r.t internal memory. We are putting this on ice for the moment.

danpat:
[...]

PR #4881, opened 2018-02-13
Add datafacade allocator backed by mmap

TheMarex:
# Issue

This is a first step towards #1947. We don't mmap the OSRM dataset directly but create a new file that basically stores the memory content as it is now. This is less effective both from a disk storage and startup time perspective, but should give us a good idea how the query performance using mmap without any optimizations.

## Tasklist

 - [x] Try on biggest production dataset (bike?) and measure performance
 - [x] Investigate performance degradation under memory pressure.
 - [x] review
 - [x] adjust for comments

## Requirements / Relations

#1947

### DIFF
--- a/include/util/mmap_file.hpp
+++ b/include/util/mmap_file.hpp
@@ -33,6 +33,32 @@ util::vector_view<T> mmapFile(const boost::filesystem::path &file, RegionT &regi
             SOURCE_REF);
     }
 }
+
+template <typename T, typename RegionT>
+util::vector_view<T>
+mmapFile(const boost::filesystem::path &file, RegionT &region, const std::size_t size)
+{
+    try
+    {
+        // Create a new file with the given size in bytes
+        boost::iostreams::mapped_file_params params;
+        params.path = file.string();
+        params.flags = boost::iostreams::mapped_file::readwrite;
+        params.new_file_size = size;
+        region.open(params);
+
+        std::size_t num_objects = size / sizeof(T);
+        auto data_ptr = region.data();
+        BOOST_ASSERT(reinterpret_cast<uintptr_t>(data_ptr) % alignof(T) == 0);
+        return util::vector_view<T>(reinterpret_cast<T *>(data_ptr), num_objects);
+    }
+    catch (const std::exception &exc)
+    {
+        throw exception(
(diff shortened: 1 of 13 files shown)


TheMarex:
As a first prototype I tried to see what the general overhead is for using `mmap` instead of process internal memory and ran 50000 random queries on Bayern. Blue is `internal` memory `red` is mmap.
As you can see there is already a slight slowdown of just using `mmap` over `internal` memory even without memory pressure.
I ran two experiments once just running `osrm-extract` and `osrm-contract` and one that also ran `osrm-partition` to renumber the nodes in the graph using the partition information.

### No renumbering

image

### With renumbering

image

You can see that without memory pressure both perform about the same. The hope here is that renumbering will improve the locality.

TheMarex:
We ran some experiments using the bicycle profile on North-America. To test the behavior under memory pressure, we first determined the minimal memory usage as 5.8G (first time serving queries completed without an OOM). Shows are the results for random queries which are generally the worst-case. The difference between 20% and 25% is quite large since we start to trash pages at a high rate due to the random access. For a `20%` memory reduction, we would expect a slowdown of factor two in the worst case.

image

TheMarex:
If we use renumbering (running `osrm-partition`) we see a much better performance under memory pressure:

image

For completeness I also run tests using MLD to see how this approach fairs:

image

We can see that MLD is in general more resilient towards memory pressure than CH, but is overall slower.

include/util/mmap_file.hpp
```diff
@@ -33,6 +33,32 @@ util::vector_view<T> mmapFile(const boost::filesystem::path &file, RegionT &regi
...
+template <typename T, typename RegionT>
+util::vector_view<T>
+mmapFile(const boost::filesystem::path &file, RegionT &region, const std::size_t size)
+{
+    try
+    {
+        // Create a new file with the given size in bytes
+        boost::iostreams::mapped_file_params params;
+        params.path = file.c_str();
+        params.mode = std::ios::in | std::ios::out;
```
oxidase: `mode` is deprecated, better to use `params.flags = boost::iostreams::mapped_file::readwrite;`

TheMarex:
@oxidase can you check if this works for you? :bowing_man:

PR #4884, opened 2018-02-13
Add .editorconfig file

mloskot:
Propose basic settings for controlling EOL encoding and space-only indentation of CMake and other scripts.

## Tasklist

 - [x] review
 - [x] adjust for comments

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,30 @@
+# EditorConfig is awesome: http://EditorConfig.org
+#
+# NOTE: Keep settings in sync with the master .clang-format file
+#
+# top-most EditorConfig file
+root = true
+
+# CMake configuration files
+[{CMakeLists.txt,CMakeSettings.json,*.cmake}]
+indent_size = 2
+indent_style = space
+trim_trailing_whitespace = true
+
+# CI configuration files
+[{.travis.yml,appveyor.yml}]
+indent_size = 2
+indent_style = space
+trim_trailing_whitespace = true
+
+# Unix shell scripts
+[*.sh]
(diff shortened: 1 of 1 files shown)


.editorconfig
```diff
@@ -0,0 +1,30 @@
+# EditorConfig is awesome: http://EditorConfig.org
+#
+# NOTE: Keep settings in sync with the master .clang-format file
```
TheMarex: I don't see rules for the `.{h,c}pp` files, is that something that is not usually captured in this config?

mloskot: @TheMarex This config can control encoding/indentation of any text file (see an example https://github.com/nanodbc/nanodbc/blob/master/.editorconfig).

I just didn't know why `.editorconfig` is not used by the project, perhaps that was deliberate decision. I did not want to be too intrusive with rules too many rules. I'm also not sure waht should be the value of `insert_final_newline`, `trim_trailing_whitespace`, etc. for OSRM codebase.

So, I decided to utilise `.editorconfig` for the files which are not handled by `clang-format`. By the way, `.editorconfig` is not a replacement for `clang-format` - it is supporting it during typing, `clang-format` acts a bit later :)

mloskot: @TheMarex Do you wish me to update `.editorconfig` for C++ and other files?

TheMarex: @mloskot no I think this approach is fine. Could you try to re-trigger the travis builds with a new commit? Seems like it wasn't picked up.

mloskot: @TheMarex OK, commit update pushed

PR #4885, opened 2018-02-14
Prefer logical punctuators instead of named operator keywords

mloskot:
Some compilers (eg. MSVC) still require inclusion of `<ciso646>` in order to import named operator keywords.
It is easier to stick with traditional punctuators.

## Tasklist

 - [x] review
 - [x] adjust for comments


mloskot:
Thanks for accepting

Issue #4892, opened 2018-02-16
Bug: Unneccessary u-turn


jguthrie100:
If you get directions from Knutsford, UK to Manchester, UK, then the routing instructs to do a u-turn where it seems none is required..

http://map.project-osrm.org/?z=16&center=53.463443%2C-2.239441&loc=53.307874%2C-2.372051&loc=53.479130%2C-2.244101&hl=en&alt=0

(On Greenheys Lane, near Manchester city centre)

jguthrie100:
Turns out someone had set a set of traffic lights to be "No straight on", which I've now fixed on OSM

oxidase:
@jguthrie100 Thanks for fixing the issue! Closing here as the roads graph was updated.

PR #4893, opened 2018-02-16
Handle motorway forks with links as normal motorway intersections

oxidase:
# Issue

If a motorway fork intersection has links the motorway handler can assign incorrect `Left` modifier
screenshot from 2018-02-16 11-54-20
http://map.project-osrm.org/?z=18&center=53.461582%2C-2.244827&loc=53.461383%2C-2.245116&loc=53.462187%2C-2.245100&hl=en&alt=0

PR changes handling of motorway intersections to a case of a normal motorway passing some ramps if `exiting_motorways` is not equal to `valid_exits`.

## Tasklist

 - [x] review
 - [x] adjust for comments

## Requirements / Relations

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
       - CHANGED #4835: MAXIMAL_ALLOWED_SEPARATION_WIDTH increased to 12 meters
       - CHANGED #4842: Lower priority links from a motorway now are used as motorway links [#4842](https://github.com/Project-OSRM/osrm-backend/pull/4842)
       - CHANGED #4895: Use ramp bifurcations as fork intersections [#4895](https://github.com/Project-OSRM/osrm-backend/issues/4895)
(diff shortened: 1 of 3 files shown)


jguthrie100:
Worth noting as well that the osrm map tile at that location needs to be updated.. https://www.openstreetmap.org/directions?engine=graphhopper_car&route=53.30787%2C-2.37205%3B53.46230%2C-2.24521

oxidase:
@jguthrie100 :wave: I have found the issue while looking at #4892. The data fix for #4892 will be visible after the OSM change will propagate in the data pipeline. But  correct instructions at the intersection will be visible after the next OSRM release.

jguthrie100:
Maybe, but the error is in the same place when using Graphhopper too, which makes me think it's probably something to do with the underlying data (which I've hopefully now fixed in OSM)

screenshot_20180216-173436

ghoshkaj:
It looks like this PR is out of sync with master rn though. Rebase and merge? 😄

PR #4905, opened 2018-02-21
Check required tags of `maneuver` relations

oxidase:
# Issue

Accordingly to https://github.com/Project-OSRM/osrm-backend/wiki/Maneuver-override-tag ways `from`, `to` and node `via` roles are required in relations with `type=maneuver`.

PR adds an additional check for `via_node` ids to  filter out invalid `maneuver` relations.

/cc @karenzshea @TheMarex 

## Tasklist

 - [ ] review
 - [ ] adjust for comments

## Requirements / Relations

### DIFF
--- a/src/extractor/maneuver_override_relation_parser.cpp
+++ b/src/extractor/maneuver_override_relation_parser.cpp
@@ -52,8 +52,10 @@ ManeuverOverrideRelationParser::TryParse(const osmium::Relation &relation) const
     maneuver_override.maneuver = relation.tags().get_value_by_key("maneuver", "");
     maneuver_override.direction = relation.tags().get_value_by_key("direction", "");
 
-    boost::optional<std::uint64_t> from = boost::none, via = boost::none, to = boost::none;
-    std::vector<std::uint64_t> via_ways;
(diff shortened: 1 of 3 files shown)


src/extractor/maneuver_override_relation_parser.cpp
```diff
@@ -103,18 +104,12 @@ ManeuverOverrideRelationParser::TryParse(const osmium::Relation &relation) const
         }
     }
 
-    if (from && (via || via_ways.size() > 0) && to)
+    if (from != SPECIAL_OSM_WAYID && to != SPECIAL_OSM_WAYID && via_node != SPECIAL_OSM_NODEID)
```
TheMarex: This check won't work when there are only via ways right? In that case `via_node` will not be set only `via_ways`.

TheMarex: As per chat, that is intended behavior, because a way is not sufficient to infer the instruction position.

oxidase:
@TheMarex I've added checks for `Required` and `only one member.`, otherwise  a maneuver placement can be ambiguous.

Issue #4909, opened 2018-02-23
Geometry points being dropped when map-matching using waypoints parameter


danpat:
The following test case fails, but it should pass:

```
    Scenario: Regression test - waypoints trimming too much geometry
        Given the node map
            """
            ad
            |
            |
            |
            |
            |e   g
            b--------------c
            f              h
            """

        And the ways
            | nodes |
            | ab    |
            | bc    |

        Given the query options
            | waypoints | 0;4   |

        When I match I should get
            | trace    | geometry                                                     | code |
            | defgh    | 1,1,1,0.99955,1,0.999461,1.000225,0.999461,1.000674,0.999461 | Ok   |
```

Disabling the `waypoints` query option on this test will cause it to pass.

/cc @karenzshea

danpat:
Fixed by #4911, thanks @karenzshea !

PR #4911, opened 2018-02-23
Remove deduplication of unpacked_path_segments in MM collapsing

karenzshea:
# Issue

Fixes the issue in indicated by https://github.com/Project-OSRM/osrm-backend/issues/4909
image

## Bug
Leg collapsing for some traces with high sample rates, resulted in dropped geometries like shown by the red line in the image above, particularly around corners.

## Root cause
The bug was caused by a mishandling of `unpacked_path_segments` collapsing. The scenario arose when two points in a trace were located on the same edge, resulting in a 0 length segment, which, because the leg collapsing was incorrectly removing the last `unpacked_path_segment` value from the previously last collapsed segment, would sometimes result in missing coordinates in the end result.

## Tasklist

 - [x] CHANGELOG.md entry
 - [x] review
 - [x] adjust for comments

## Requirements / Relations

5.16 cc @chaupow

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
 # UNRELEASED
   - Changes from 5.16.0:
+    - Bugfixes: fix deduplication of route steps when waypoints are used [#4909](https://github.com/Project-OSRM/osrm-backend/issues/4909)
 
 # 5.16.0
   - Changes from 5.15.2:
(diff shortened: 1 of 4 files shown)


karenzshea:
Also, thanks to @danpat for teasing out a good test case expressing the bug

PR #4944, opened 2018-03-08
Don't route bikes on oneway roads in the wrong direction

### LINKED ISSUE #4943: Routing bikes on oneway roads in the wrong direction
Sometimes we will route bikes on a oneway street in the wrong direction.

- Example request is `/route/v1/driving/-73.98244857788086,40.755319574776024;-73.999958,40.736072?overview=false&alternatives=true&steps=true` where we route down `6th Ave` in the wrong direction
screenshot 2018-03-08 17 03 23

- Looking at debug tiles shows that we use 15km/h speed in both directions on `6th Ave`
screenshot 2018-03-08 16 55 28

- OpenStreetMap tags this [road](https://www.openstreetmap.org/way/497746843) as 
     - `oneway=yes`
     - `cycleway:left=track` 

### Problem:
- the profile will first correctly identify the oneway and tag it as `mode.inaccessible` [here](https://github.com/Project-OSRM/osrm-backend/blob/master/profiles/bicycle.lua#L351)
- afterwards the profile will check if there are cycleways and if there is a cycleway on the left/right side assume that the road is cyclable backwards/forwards [here](https://github.com/Project-OSRM/osrm-backend/blob/master/profiles/bicycle.lua#L393)
- oneways that have cycleways on the left side to cycle the way **in forward direction** are now handled as if they are cyclable **in backward direction**

**Problem description in other words:**
  ```
  a -------------------> b
  ```
- When there is a cycleway on the left side of a road (left when looking `a -> b`), we assume that we can ride the bike when riding from `b > a`
- Even if `a > b` is a oneway we still think we can do that
- This is not true for 6th in New York: the cycleway is on the left side but it is not for riding `b > a` but `a > b`

### Solution
- don't set roads as cyclable regardless of cycleway if they were marked as inaccessible before.

chaupow:
# Issue

Fixes #4943

## Tasklist

 - [x] review
 - [x] adjust for comments

### DIFF
--- a/features/bicycle/cycleway.feature
+++ b/features/bicycle/cycleway.feature
@@ -51,6 +51,8 @@ Feature: Bike - Cycle tracks/lanes
             | motorway |          | opposite      |                | x    | x     |
             | motorway |          |               | track          | x    |       |
             | motorway |          |               | opposite       | x    | x     |
+            # motorways are implicit oneways and cycleway tracks next to oneways always
+            # follow the oneway direction (unless tagged as opposite)
             | motorway |          | track         | track          | x    |       |
             | motorway |          | opposite      | opposite       | x    | x     |
             | motorway |          | track         | opposite       | x    | x     |
--- a/features/bicycle/cycleway.feature
+++ b/features/bicycle/cycleway.feature
@@ -46,12 +46,12 @@ Feature: Bike - Cycle tracks/lanes
             | primary  |          |               |                | x    | x     |
             | motorway |          |               |                |      |       |
             | motorway | track    |               |                | x    |       |
-            | motorway | opposite |               |                |      | x     |
-            | motorway |          | track         |                |      | x     |
-            | motorway |          | opposite      |                |      | x     |
+            | motorway | opposite |               |                | x    | x     |
+            | motorway |          | track         |                | x    |       |
+            | motorway |          | opposite      |                | x    | x     |
             | motorway |          |               | track          | x    |       |
-            | motorway |          |               | opposite       | x    |       |
-            | motorway |          | track         | track          | x    | x     |
+            | motorway |          |               | opposite       | x    | x     |
+            | motorway |          | track         | track          | x    |       |
             | motorway |          | opposite      | opposite       | x    | x     |
             | motorway |          | track         | opposite       | x    | x     |
             | motorway |          | opposite      | track          | x    | x     |
--- a/features/bicycle/cycleway.feature
+++ b/features/bicycle/cycleway.feature
@@ -92,6 +92,6 @@ Feature: Bike - Cycle tracks/lanes
         Then routability should be
             | highway     | cycleway | oneway | forw    | backw      |
             | motorway    | track    | yes    | 15 km/h |            |
-            | residential | track    | yes    | 15 km/h | 6 km/h +-1 |
-            | cycleway    | track    | yes    | 15 km/h | 6 km/h +-1 |
-            | footway     | track    | yes    | 15 km/h | 6 km/h +-1 |
+            | residential | track    | yes    | 15 km/h | 4 km/h +-1 |
+            | cycleway    | track    | yes    | 15 km/h | 4 km/h +-1 |
+            | footway     | track    | yes    | 15 km/h | 4 km/h +-1 |
--- a/profiles/bicycle.lua
+++ b/profiles/bicycle.lua
@@ -368,35 +368,73 @@ function handle_bicycle_tags(profile,way,result,data)
   end
 
   -- cycleway
-  local has_cycleway_left = false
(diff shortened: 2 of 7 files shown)


chaupow:
Comment to changing the walking-speed:

[Wikipedia](https://en.wikipedia.org/wiki/Walking) says 
> the average human walking speed at crosswalks is about 5.0 

Also pushing the bike is probably slower than the average speed, so I would prefer ad speed of 5 or less.

chaupow:
Comments on failing tests:

Changed [these tests](https://github.com/Project-OSRM/osrm-backend/pull/4944/files#diff-29fca727bdedf0a673183174db4cc2aaL49) to
```
| highway  | cycleway | cycleway:left | cycleway:right | forw | backw |
| motorway | opposite |               |                | x    | x     |
| motorway |          | opposite      |                | x    | x     |
| motorway |          |               | opposite       | x    | x     |
```
because `opposite` is tagged when cycleways can go in both directions

```
| highway  | cycleway | cycleway:left | cycleway:right | forw | backw |
| motorway |          | track         |                | x    |       |
| motorway |          | track         | track          | x    |       |
```
because motorway is an implicit oneway, and a cycleway on the left of a one-way is going in the direction of the one-way road https://github.com/Project-OSRM/osrm-backend/issues/4943#issuecomment-371832446

features/bicycle/cycleway.feature
```diff
@@ -46,12 +46,12 @@ Feature: Bike - Cycle tracks/lanes
...
-            | motorway |          | track         |                |      | x     |
-            | motorway |          | opposite      |                |      | x     |
+            | motorway | opposite |               |                | x    | x     |
+            | motorway |          | track         |                | x    |       |
+            | motorway |          | opposite      |                | x    | x     |
             | motorway |          |               | track          | x    |       |
-            | motorway |          |               | opposite       | x    |       |
-            | motorway |          | track         | track          | x    | x     |
+            | motorway |          |               | opposite       | x    | x     |
+            | motorway |          | track         | track          | x    |       |
```
TheMarex: This particular case should get a comment of why this is different for (implicit) oneways and possibly invalid data.

profiles/bicycle.lua
```diff
@@ -367,35 +368,73 @@ function handle_bicycle_tags(profile,way,result,data)
...
+  if result.forward_mode ~= mode.inaccessible and result.backward_mode ~= mode.inaccessible and not implied_oneway then
+
+    if cycleway and profile.cycleway_tags[cycleway] and string.find(cycleway, "opposite") ~= 1 then
+
+      has_cycleway_backward = true
+      has_cycleway_forward = true
+
+    end
+    if (cycleway_right and profile.cycleway_tags[cycleway_right] and string.find(cycleway_right, "opposite") ~= 1)
+       or (cycleway_left and profile.cycleway_tags[cycleway_left] and string.find(cycleway_left, "opposite") == 1) then
```
TheMarex: Maybe we can simplify this logic a little bit by creating a helper function that handles bi-directional tag lookup like that, similar to `get_forward_backward_by_key` in `tags.lua`.

profiles/bicycle.lua
```diff
@@ -367,35 +368,73 @@ function handle_bicycle_tags(profile,way,result,data)
   end
 
   -- cycleway
-  local has_cycleway_left = false
-  local has_cycleway_right = false
+  local has_cycleway_forward = false
+  local has_cycleway_backward = false
+
+  -- cycleways on normal roads
+  if result.forward_mode ~= mode.inaccessible and result.backward_mode ~= mode.inaccessible and not implied_oneway then
```
TheMarex: We could capture this condition in a variable for documentation. Maybe something like `local is_bidirectional = ....`

profiles/bicycle.lua
```diff
@@ -367,35 +368,73 @@ function handle_bicycle_tags(profile,way,result,data)
...
+       or (cycleway_right and profile.cycleway_tags[cycleway_right] and string.find(cycleway_right, "opposite") == 1) then
+
+      has_cycleway_backward = true
+
+    end
+
   end
 
+  -- cycleways on one-way roads
+  if result.forward_mode == mode.inaccessible or result.backward_mode == mode.inaccessible or implied_oneway then
```
TheMarex: Same here, we could improve readability by capturing this in a variable.

profiles/bicycle.lua
```diff
@@ -576,6 +615,9 @@ function process_turn(profile, turn)
   if profile.properties.weight_name == 'cyclability' then
     turn.weight = turn.duration
   end
+  if turn.source_mode == mode.cycling and turn.target_mode ~= mode.cycling then
```
TheMarex: This is a great idea and will probably increase the usability of this a lot!

PR #4946, opened 2018-03-08
Remove correctly the last segment in annotation

oxidase:
# Issue

Remove the last trimmed segment in annotations array. 

/cc @freenerd @danpat 

## Tasklist

 - [ ] review
 - [ ] adjust for comments

## Requirements / Relations

### DIFF
--- a/features/testbot/traffic_speeds.feature
+++ b/features/testbot/traffic_speeds.feature
@@ -48,10 +48,10 @@ Feature: Traffic - speeds
 
         When I route I should get
           | from | to | route       | speed   | weights              | a:datasources |
-          | a    | b  | ad,de,eb,eb | 30 km/h | 1275.7,400.4,378.2,0 | 1:0:0:0       |
+          | a    | b  | ad,de,eb,eb | 30 km/h | 1275.7,400.4,378.2,0 | 1:0:0         |
           | a    | c  | ad,dc,dc    | 31 km/h | 1275.7,956.8,0       | 1:0           |
-          | b    | c  | bc,bc       | 27 km/h | 741.5,0              | 1:0           |
-          | a    | d  | ad,ad       | 27 km/h | 1275.7,0             | 1:0           |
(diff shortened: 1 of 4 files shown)


features/testbot/traffic_speeds.feature
```diff
@@ -48,10 +48,10 @@ Feature: Traffic - speeds
 
         When I route I should get
           | from | to | route       | speed   | weights              | a:datasources |
-          | a    | b  | ad,de,eb,eb | 30 km/h | 1275.7,400.4,378.2,0 | 1:0:0:0       |
+          | a    | b  | ad,de,eb,eb | 30 km/h | 1275.7,400.4,378.2,0 | 1:0:0         |
```
TheMarex: This seems wrong: `a:datasources` should have the same length as `weights`.

oxidase: @TheMarex the test case already has an inconsistency in expectation values: lines 53-54 have two weight values and two datasources, lines 55-56 have two weight values and one datasource.  The change only fixes  inconsistency in annotation values: now number of annotation weights, durations, data sources, speed values is by one less than number of node IDs. I'll :eyes: how to make consistent step `weights` and annotation `a:weight`, because with `master` number of weights is different for the `dc` case:
```
| from | to | route       | speed   | weights              | a:weight             |
| a    | b  | ad,de,eb,eb | 30 km/h | 1275.7,400.4,378.2,0 | 1275.7:400.4:378.2:0 |
| b    | c  | bc,bc       | 27 km/h | 741.5,0              | 741.5:0              |
| d    | c  | dc,dc       | 36 km/h | 956.8,0              | 956.8                |
```

TheMarex: Oh I see so actually the `weights` are wrong in this scenario?

danpat: I'm 👍 on just fixing one problem at a time here.

Issue #4952, opened 2018-03-12
Refactor mapping of files to memory locations


TheMarex:
Currently the logic to map a section of a file to a location in memory is hand written in [storage.cpp](https://github.com/Project-OSRM/osrm-backend/blob/master/src/storage/storage.cpp#L266-L298) for every file in the function `PopulateLayout`. This is tedious and error prone as it assume this logic needs to be adapted every time we change the structure of the files.

To change this we should capture the logic of figuring out the blocks of memory contained in each file to an own function.

Unblocks mapping separate files in #1947 and makes #4873.

TheMarex:
When thinking how to implement this it became clear very fast that the main blocker for us is that our on-disk format is relatively unstructured and does not allow us to compute the size of memory segments and read/map the data in a very general way.

To address these problems we need to change our file format. Seems like https://github.com/Project-OSRM/osrm-backend/issues/2242 is rearing its ugly head again. Our new constraints would be:

- List content of file and its size
- Content of file `mmap`-able to memory (e.g. no decoding step with memory buffer)
- Writable in a streamable way
- Data should have a human-readable name
- Data should be hierarchical

This already sound very much like a `tar` file and any implementation of it would probably be very close to a `tar`-lite format. However using `tar` files would additionally solve a lot a problems around tooling, they would be easy to inspect, extend and modify. Eventually we don't need to care about how data is packaged, we can accept any split of data.

## Storage

Our investment in using a general abstraction layer with `FileWriter` and `FileReader` seems to pay off: We can swap out the current implementation to write _named_ data. A low-level library to read/write `tar` files we can use [microtar](https://github.com/rxi/microtar) which we can easily wrap/bundle as third-party library.

We can use a "filesystem" like structure to get a human-readable hierarchical representation of data. For example the current `.osrm.mldgr` file could be represented as:

```
/mld/multilevelgraph/node_array
/mld/multilevelgraph/edge_array
/mld/multilevelgraph/node_to_edge_offsets
OSRM_VERSION
```

## Memory

In conjunction with the new on-disk format we can now make `osrm-datastore` a lot more "stupid" as it only needs to care about the following things:

1. Discovering data in files
2. Building an index where to find the data in files
3. Allocating enough in-memory storage (potentially using multiple memory blocks)
4. Building an index to find the data in memory
5. Reading the data to memory

The last three steps would be optional if we go for an `mmap` based approach in the future.

## Data organization

[...]

PR #4960, opened 2018-03-15
Use tar-format to encapsulate data

TheMarex:
# Issue

This PR is another step towards implementing #4952. The goal is to make any `.osrm.*` file a `tar` file internally and address data by (file)name. As example the `.mldgr` file now looks like this:

```
/mld/multilevelgraph/node_array.meta
/mld/multilevelgraph/node_array
/mld/multilevelgraph/edge_array.meta
/mld/multilevelgraph/edge_array
/mld/multilevelgraph/node_to_edge_offset.meta
/mld/multilevelgraph/node_to_edge_offset
/mld/multilevelgraph/connectivity_checksum.meta
/mld/multilevelgraph/connectivity_checksum
```
Where every file contains raw binary data for data structures in OSRM that we can deserialize or read directly to memory. The `.meta` files contain additional information for setting up the data layout, e.g. the number of values for serialized `std::vector`, or the fingerprint.

## Tasklist

 - [x] review
 - [x] adjust for comments


include/storage/serialization.hpp
```diff
@@ -29,160 +33,190 @@ namespace serialization
...
-        writer.WriteOne<unsigned char>(packBits(data, CHAR_BIT * index, CHAR_BIT));
-    }
-    if (count > index)
-        writer.WriteOne<unsigned char>(packBits(data, index, count - index));
+
+    constexpr std::uint64_t WORD_BITS = CHAR_BIT * sizeof(std::uint64_t);
+
+    // FIXME on old boost version the function_input_iterator does not work with lambdas
+    // so we need to wrap it in a function here.
+    const std::function<std::uint64_t()> encode_function = [&]() -> char {
```
oxidase: s/-> char/-> std::uint64_t/

include/storage/serialization.hpp
```diff
@@ -29,160 +33,190 @@ namespace serialization
...
+        auto write_size = std::min<std::size_t>(count - index, WORD_BITS);
+        auto packed = packBits<VectorT, std::uint64_t>(data, index, write_size);
+        index += WORD_BITS;
+        return packed;
+    };
+
+    std::uint64_t number_of_blocks = std::ceil((double)count / WORD_BITS);
```
oxidase: `(count + WORD_BITS - 1) / WORD_BITS` would be better to avoid floating-point arithmetic

TheMarex: Ah nice trick. I know a few places were this would come in useful.

oxidase:
@TheMarex i pushed fixes for msvc to trigger CI builds ("fixes" are reducing c++ syntactic sugar for msvc). Feel free to kick-out or squash the last commit.

TheMarex:
> unknown location(0): fatal error: in "route/test_manual_setting_of_annotations_property": class osrm::util::exception: File "C:/projects/osrm/test/data/ch/monaco.osrm.fileIndex" mapping failed: failed opening file: The process cannot access the file because it is being used by another process.

Some more problems with windows. I suspect this is because I had to change the r-tree to allow writing to an mmap files. Maybe I can work around this.

TheMarex:
Wuhu finally! It compiles. :rocket:

TheMarex:
@mloskot you are right, this doesn't meet the single-file requirement yet.

However, since we now use named identifiers for data, `osrm-datastore` does not really need to care anymore in which files the data is stored, we could figure that out automatically. That would unlock just passing a single file to `osrm-datastore` that is just a bunch of merged tar files. I captured this in https://github.com/Project-OSRM/osrm-backend/issues/4978 for further discussion.

mloskot:
@TheMarex It makes sense, thanks for the clarification.

Issue #4969, opened 2018-03-21
Return iterators instead of vectors in datafacade_base functions


ghoshkaj:
Return iterators instead of vectors in the [datafacade_base](https://github.com/Project-OSRM/osrm-backend/blob/master/include/engine/datafacade/datafacade_base.hpp) function as it will be faster to return the iterators instead of making new copies of the vectors and returning those.

For example in https://github.com/Project-OSRM/osrm-backend/pull/4876, the number of times that [`GetUncompressedForwardDurations` and `GetUncompressedReverseDurations`](https://github.com/Project-OSRM/osrm-backend/blob/master/include/engine/datafacade/datafacade_base.hpp#L82-L83) get called increases by a large number and in order to ensure that it doesn't become a bottleneck, doing the above would be very helpful.

```
virtual std::vector<EdgeWeight> GetUncompressedForwardDurations(const EdgeID id) const = 0;
virtual std::vector<EdgeWeight> GetUncompressedReverseDurations(const EdgeID id) const = 0;
```

PR #4876, opened 2018-02-12
Compute durations for table plugin at runtime

ghoshkaj:
# Issue

This PR implements a way to compute annotations for the table plugin at runtime, specifically, the annotation of `durations` to start off with. It follows the step in the first comment in this Issue: [Compute durations for table plugin at runtime](https://github.com/Project-OSRM/osrm-backend/issues/4798#issue-288608332)

## Tasklist

 - [ ] CHANGELOG.md entry ([How to write a changelog entry](http://keepachangelog.com/en/1.0.0/#how))
 - [ ] update relevant [Wiki pages](https://github.com/Project-OSRM/osrm-backend/wiki)
 - [ ] add tests (see [testing documentation](https://github.com/Project-OSRM/osrm-backend/blob/master/docs/testing.md)
 - [ ] review
 - [ ] adjust for comments
 - [ ] cherry pick to release branch

## Requirements / Relations

 Link any requirements here. Other pull requests this PR is based on?

PR #4972, opened 2018-03-23
Return iterators instead of vectors in datafacade_base functions

ghoshkaj:
# Issue

[Return iterators instead of vectors in datafacade_base functions](https://github.com/Project-OSRM/osrm-backend/issues/4969)

## Tasklist

 - [ ] review
 - [ ] adjust for comments

## Requirements / Relations

Needed for speeding up https://github.com/Project-OSRM/osrm-backend/pull/4876/


ghoshkaj:
@themarex @oxidase currently there are a number of failing unit tests (`library-tests`, `updater tests`, `customizer tests`)  and after this commit [`use any_range in tile_turns`](https://github.com/Project-OSRM/osrm-backend/pull/4972/commits/f85c576bc22a7a08877d258adb033f45b55109d0), a lot of failing cucumber tests: 
I’m wary of continuing to implement this same pattern for:
-  `GetUncompressedForwardGeometry` and `GetUncompressedReverseGeometry`
- `GetUncompressedForwardDatasources` and `GetUncompressedReverseDatasources` 

before solving the failing tests first. So I'll continue solving those tests first and the continue for the two other sets of vectors above.

TheMarex:
@ghoshkaj I can only find compilation errors on Travis: https://travis-ci.org/Project-OSRM/osrm-backend/jobs/358731863#L1535-L1538 what specifically is failing?

oxidase:
@ghoshkaj @TheMarex atm  `boost::any_range` with  `PackedVector` iterators may lead to an undefined behavior that is reproducible with `-O3` optimization level. A stack trace is
```
#0  0x00005555556d61e9 in boost::range_detail::any_iterator<unsigned int, boost::iterators::random_access_traversal_tag, osrm::util::detail::PackedVector<unsigned int, 22ul, (osrm::storage::Ownership)0>::internal_reference, long, boost::any_iterator_buffer<64ul> >::dereference (this=<optimized out>) at /home/miha/mapbox/osrm-backend/mason_packages/headers/boost/1.65.1/include/boost/range/detail/any_iterator.hpp:512
#1  boost::iterators::iterator_core_access::dereference<boost::range_detail::any_iterator<unsigned int, boost::iterators::random_access_traversal_tag, osrm::util::detail::PackedVector<unsigned int, 22ul, (osrm::storage::Ownership)0>::internal_reference, long, boost::any_iterator_buffer<64ul> > > (f=...) at /home/miha/mapbox/osrm-backend/mason_packages/headers/boost/1.65.1/include/boost/iterator/iterator_facade.hpp:549
#2  boost::iterators::detail::iterator_facade_base<boost::range_detail::any_iterator<unsigned int, boost::iterators::random_access_traversal_tag, osrm::util::detail::PackedVector<unsigned int, 22ul, (osrm::storage::Ownership)0>::internal_reference, long, boost::any_iterator_buffer<64ul> >, unsigned int, boost::iterators::random_access_traversal_tag, osrm::util::detail::PackedVector<unsigned int, 22ul, (osrm::storage::Ownership)0>::internal_reference, long, false, false>::operator* (this=0x7fffffffbfa0) at /home/miha/mapbox/osrm-backend/mason_packages/headers/boost/1.65.1/include/boost/iterator/iterator_facade.hpp:655
#3  packed_vector_test::packed_weights_vector::test_method (this=this@entry=0x7fffffffcada) at /home/miha/mapbox/osrm-backend/unit_tests/util/packed_vector.cpp:232
#4  0x00005555556d6a0d in packed_vector_test::packed_weights_vector_invoker () at /home/miha/mapbox/osrm-backend/unit_tests/util/packed_vector.cpp:207
```

oxidase:
@ghoshkaj it can be  a possible solution. `any_range` implements the type-erasure idiom that is used at https://github.com/Project-OSRM/osrm-backend/blob/d5fa8702f78b5e645363c9a75df0d510e0050411/include/engine/routing_algorithms/routing_base.hpp#L138-L141, so code that uses type-specific iterators must be adjusted.

oxidase:
@ghoshkaj the undefined behavior was caused by a regression bug in `boost::any_range`. The "fix" is to use immutable references by adding a `const` modifier to the `Refernce` argument of `boost::any_range`.

References:
https://svn.boost.org/trac10/ticket/10493
https://stackoverflow.com/questions/42427395/boostany-range-with-optimization-level-o2-causes-crash/42427662

oxidase:
@ghoshkaj @TheMarex the undefined behavior is fixed now, but i have concern about `boost::any_range` performance because of dynamic polymorphism used in `boost::any_range`. It would be nice to check unpacking performance for matrices before merging PR.

Similar SO thread https://stackoverflow.com/a/5607311

ghoshkaj:
@oxidase this makes sense. I've done some measurements on a small dataset (berlin-latest) with 5x5 matrix and 5000 requests. On a small scale it doesn't seem to make a slowdown.

But actually I don't know what to compare this against, there isn't an implementation that doesn't use any other type of range :) I could maybe hardcode the forward and reverse range types as we did before as a hack, and compare against that?

unbound cache before vector to range switch | unbounded cache after vector to range switch | lru-cache-100000 line size (each line is 20 bytes) |  master (durations on graph edge)
-- | -- | -- |---
7.38 ms | 5.15 ms | 5.44 ms | 4.52 ms

And I'll run this on a bigger dataset to get more realistic numbers and performance measures but what should I be looking out for in particular?

include/engine/datafacade/datafacade_base.hpp
```diff
@@ -50,6 +50,14 @@ class BaseDataFacade
 {
   public:
     using RTreeLeaf = extractor::EdgeBasedNodeSegment;
+
+    template <typename T>
+    using RangeT = boost::any_range<T, boost::random_access_traversal_tag, const T, std::ptrdiff_t>;
+    using NodesIDRangeT = RangeT<NodeID>;
+    using WeightsRangeT = RangeT<SegmentWeight>;
+    using DurationsRangeT = RangeT<SegmentDuration>;
+    using DatasourceIDRangeT = RangeT<DatasourceID>;
```
TheMarex: Using the type extension `T` kind of breaks convention were we would use that for template parameter names only. We could use `Range`, `NodesIDRange` etc instead.

include/engine/geospatial_query.hpp
```diff
@@ -434,47 +434,49 @@ template <typename RTreeT, typename DataFacadeT> class GeospatialQuery
...
-        {
-            reverse_weight_offset += reverse_weight_vector[i];
-            reverse_duration_offset += reverse_duration_vector[i];
-        }
-        reverse_weight =
-            reverse_weight_vector[reverse_weight_vector.size() - data.fwd_segment_position - 1];
-        reverse_duration =
-            reverse_duration_vector[reverse_duration_vector.size() - data.fwd_segment_position - 1];
+        const auto forward_duration_offset =
+            std::accumulate(forward_durations.begin(),
```
TheMarex: Nice refactor!

include/extractor/segment_data_container.hpp
```diff
@@ -56,7 +56,7 @@ template <storage::Ownership Ownership> class SegmentDataContainerImpl
     using DirectionalGeometryID = std::uint32_t;
     using SegmentOffset = std::uint32_t;
     using SegmentWeightVector = PackedVector<SegmentWeight, SEGMENT_WEIGHT_BITS>;
-    using SegmentDurationVector = PackedVector<SegmentDuration, SEGMENT_DURAITON_BITS>;
+    using SegmentDurationVector = PackedVector<SegmentDuration, SEGMENT_DURATION_BITS>;
```
TheMarex: Good catch!

TheMarex:
I ran 10 000 random queries locally on California. The results are a little bit disappointing, looks like `any_range` really adds an significant overhead that is just as expensive as copying the vector. That indicates we need to re-think using `any_range` and try to expose non-polymorphic types.

What could work is capturing the return type of `Get{Forward,Backward}Uncompressed{Durations,Weights,Geometies}` using `decltype` in an alias.

|  branch | CH (avg. ms)   | MLD (avg. ms) |
|----------|-----------|----------|
| `master` | 14.53 | 22.50 |
| this | 14.59 |  22.78 |
### CH

CH

### MLD

MLD

danpat:
😞 that sadly aligns with the comments in the boost docs:

https://www.boost.org/doc/libs/1_66_0/libs/range/doc/html/range/reference/ranges/any_range.html

> Despite the underlying any_iterator being the fastest available implementation, the performance overhead of any_range is still appreciable due to the cost of virtual function calls required to implement increment, decrement, advance, equal etc. Frequently a better design choice is to convert to a canonical form.

oxidase:
@TheMarex I guess most of time spent in `Get{Forward,Backward}Uncompressed{Durations,Weights}` methods during matrix requests profiling was not directly in copying vectors, but in unpacking values. 

Potentially a single serial copy&unpacking without using internal references could improve performance here.

TheMarex:
I re-ran the benchmark from above locally and as expected there is no speedup for normal routes, but also no real slowdown. This is fine to merge to unblock #4990.

PR #4975, opened 2018-03-26
Refactor shared memory data layout to use named identifiers

TheMarex:
# Issue

This works towards https://github.com/Project-OSRM/osrm-backend/issues/4952 by using the string identifiers everywhere to reference to the data. This makes `DataLayout` quite "stupid" as it does not know about the data anymore and can have a variable number of entries.

## Tasklist

 - [x] Merge #4960 
 - [x] review
 - [x] adjust for comments

## Requirements / Relations

This PR is based on #4960.

### DIFF
--- a/include/storage/shared_datatype.hpp
+++ b/include/storage/shared_datatype.hpp
@@ -2,6 +2,7 @@
 #define SHARED_DATA_TYPE_HPP
 
 #include "storage/block.hpp"
+#include "storage/io_fwd.hpp"
 
 #include "util/exception.hpp"
 #include "util/exception_utils.hpp"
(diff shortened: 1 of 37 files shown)


include/storage/shared_datatype.hpp
```diff
@@ -11,226 +12,152 @@
...
-                                            "MLD_GRAPH_EDGE_LIST",
-                                            "MLD_GRAPH_NODE_TO_OFFSET",
-                                            "MANEUVER_OVERRIDES",
-                                            "MANEUVER_OVERRIDE_NODE_SEQUENCES"};
-
 class DataLayout
 {
   public:
-    enum BlockID
-    {
```
oxidase: :100:

src/storage/storage.cpp
```diff
@@ -402,15 +313,15 @@ void Storage::PopulateData(const DataLayout &layout, char *memory_ptr)
     {
         const auto name_blocks_ptr =
             layout.GetBlockPtr<extractor::NameTableView::IndexedData::BlockReference, true>(
-                memory_ptr, DataLayout::NAME_BLOCKS);
+                memory_ptr, "/common/names/blocks");
```
oxidase: It is good to move initialization into a free function as
```
return util::vector_view<BlockT> (layout.GetBlockPtr<BlockT, true>(memory_ptr, name), layout.GetBlockEntries(name));
```
to avoid duplication of type-name pairs.

include/storage/shared_datatype.hpp
```diff
@@ -11,226 +12,152 @@
...
+inline void write(io::BufferWriter &writer, const DataLayout &layout);
+}
+
+namespace detail
+{
+// Removes the file name if name_prefix is a directory and name is not a file in that directory
+inline std::string trimName(const std::string &name_prefix, const std::string &name)
+{
+    // list directory and
+    if (name_prefix.back() == '/')
```
oxidase: `!name_prefix.empty() && ...`

src/storage/storage.cpp
TheMarex: Oh great idea. :+1:

TheMarex:
@oxidase Could you review this again? I refactored creating the views so it can be shared by `osrm-datastore` and the datafacade. This make it necessary to remove the canary logic.

Issue #4873, opened 2018-02-12
Split shared memory layout into metric/non-metric part


TheMarex:
Current our shared memory is a [singled block](https://github.com/Project-OSRM/osrm-backend/blob/master/include/storage/shared_datatype.hpp) of annotation data and non-metric data. We should split this into one block responsible for persistent data (e.g. names, graph structure, axillary data) and data depending on the metric being used.

Splitting this data into two updatable pieces is the first step towards #4007.

Metric independent:

```
                                            "EDGE_BASED_NODE_DATA",
                                            "ANNOTATION_DATA",
                                            "COORDINATE_LIST",
                                            "OSM_NODE_ID_LIST",
                                            "TURN_INSTRUCTION",
                                            "ENTRY_CLASSID",
                                            "R_SEARCH_TREE",
                                            "R_SEARCH_TREE_LEVELS",
                                            "GEOMETRIES_INDEX",
                                            "GEOMETRIES_NODE_LIST",
                                            "GEOMETRIES_FWD_DURATION_LIST",
                                            "GEOMETRIES_REV_DURATION_LIST",
                                            "GEOMETRIES_FWD_DATASOURCES_LIST",
                                            "GEOMETRIES_REV_DATASOURCES_LIST",
                                            "HSGR_CHECKSUM",
                                            "TIMESTAMP",
                                            "FILE_INDEX_PATH",
                                            "DATASOURCES_NAMES",
                                            "PROPERTIES",
                                            "BEARING_CLASSID",
                                            "BEARING_OFFSETS",
                                            "BEARING_BLOCKS",
                                            "BEARING_VALUES",
                                            "ENTRY_CLASS",
                                            "LANE_DATA_ID",
                                            "PRE_TURN_BEARING",
                                            "POST_TURN_BEARING",
                                            "TURN_LANE_DATA",
                                            "LANE_DESCRIPTION_OFFSETS",
                                            "LANE_DESCRIPTION_MASKS",
                                            "TURN_DURATION_PENALTIES",
                                            "MLD_LEVEL_DATA",
                                            "MLD_PARTITION",
                                            "MLD_CELL_TO_CHILDREN",
                                            "MLD_CELL_SOURCE_BOUNDARY",
                                            "MLD_CELL_DESTINATION_BOUNDARY",
                                            "MLD_CELLS",
                                            "MLD_CELL_LEVEL_OFFSETS",
                                            "MLD_GRAPH_NODE_LIST",
[...]

PR #4982, opened 2018-03-29
Allow multiple named shared memory regions

TheMarex:
# Issue

This PR adds support for having multiple named shared memory regions.
This is going to unlock https://github.com/Project-OSRM/osrm-backend/issues/1221 and https://github.com/Project-OSRM/osrm-backend/issues/4873 by making it easier to have multiple named regions.

## Tasklist

 - [x] review
 - [x] adjust for comments

## Requirements / Relations

Based on #4975.

### DIFF
--- a/include/storage/shared_datatype.hpp
+++ b/include/storage/shared_datatype.hpp
@@ -221,6 +221,18 @@ struct SharedRegionRegister
         }
     }
 
+    template<typename OutIter>
+    void List(OutIter out) const
+    {
+        for (const auto& region : regions)
+        {
+            if (!region.IsEmpty())
+            {
+                *out++ = region.name;
+            }
+        }
+    }
+
     void Deregister(const RegionID key) { regions[key] = SharedRegion{}; }
 
     const auto &GetRegion(const RegionID key) const { return regions[key]; }
--- a/include/storage/shared_datatype.hpp
+++ b/include/storage/shared_datatype.hpp
@@ -6,13 +6,13 @@
 
 #include "util/exception.hpp"
 #include "util/exception_utils.hpp"
-#include "util/log.hpp"
 
 #include <boost/assert.hpp>
 
 #include <array>
 #include <cstdint>
 #include <map>
+#include <numeric>
 #include <unordered_set>
 
 namespace osrm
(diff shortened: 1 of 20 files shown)


include/storage/shared_datatype.hpp
```diff
@@ -160,40 +160,111 @@ class DataLayout
...
+        for (const auto &region : regions)
+        {
+            if (!region.IsEmpty())
+            {
+                *out++ = region.name;
+            }
+        }
+    }
+
+    void Deregister(const RegionID key) { regions[key] = SharedRegion{}; }
```
oxidase: The method is not used and `DataWatchdogImpl` assumes that the shared region pointer won't be changed during an engine process life-time. Is a follow-up issue about deleting named regions needed?

src/osrm/osrm.cpp
```diff
@@ -28,12 +28,6 @@ OSRM::OSRM(engine::EngineConfig &config)
         throw util::exception("Required files are missing, cannot continue.  Have all the "
                               "pre-processing steps been run?");
     }
-    else if (config.use_shared_memory)
```
oxidase: :+1:

include/storage/shared_datatype.hpp
TheMarex: Good catch, I wanted to use this for the spring-clean command, but instead we just overwrite the whole registry.

oxidase:
It is also possible to add `features/options/datastore/datastore.feature` options tests to increase coverage 
```
@datastore @options @help
Feature: osrm-datastore command line options: list

    Background:
        Given the profile "testbot"
        And the node map
            """
            a b
            """
        And the ways
            | nodes |
            | ab    |
        And the data has been contracted

    Scenario: osrm-datastore - Help should be shown when no options are passed
        When I try to run "osrm-datastore --dataset-name test_dataset_42 {processed_file}"
        Then it should exit successfully
        When I try to run "osrm-datastore --list"
        Then it should exit successfully
        And stdout should contain "test_dataset_42/data"
```

TheMarex:
@oxidase thanks for the test case! Changed all locations and added the test. Should be good to merge after Travis passes.

PR #4989, opened 2018-04-01
Shared storage cache

ghoshkaj:
# Issue

Use a shared storage cache to compare against performance of the threadlocal cache in https://github.com/Project-OSRM/osrm-backend/pull/4876. This PR is to test out which is faster and keep the implementation that is faster. Related to cache considerations detailed [here](https://github.com/Project-OSRM/osrm-backend/issues/4798#issuecomment-373790423).

## Tasklist
 - [x] implement singleton pattern to make the cache accessible from global storage
 - [x] ensure that ^ is thread safe (https://github.com/Project-OSRM/osrm-backend/pull/4989/commits/01e65629ff66663106cb1f24471a1cf5ab9d6e6e#diff-3765a18558eecd8c9e22fbf48fb2eccbR96)
 - [x] implement mutexes for all functions in the cache (to lock for every read and every write)
 - [x] swap mutex for ReadWriteLock (only lock on write so that reads can still happen in parallel)
 - [ ] run benchmarks keeping the following in mind:
      - create a profile for the number of threads used: 1, 2 , 8, 16 (the lower count of threads means there will not be as many threads blocking each other, while the higher end may create a deadlock. Getting a profile of these runs gives us a better idea of what the ideal number of threads will be)
     - we want total cache size to be at maximum 500mb extra memory in total so the line-size of this cache can be 26214400. 
     - so, when benching the threadlocal cache, make sure that the line size is divided by the number of threads that you are using,  so that the combined total is not greater than 500mb.
     - run requests 20000 requests for 25x25 matrix and 100x100 matrix on a large dataset
 - [ ] either merge into the base PR or close as not necessary
 - [ ] review
 - [ ] adjust for comments

## Requirements / Relations

https://github.com/Project-OSRM/osrm-backend/pull/4876

### DIFF
--- a/features/testbot/distance_matrix.feature
+++ b/features/testbot/distance_matrix.feature
@@ -23,6 +23,79 @@ Feature: Basic Distance Matrix
             | a | 0  | 10 |
             | b | 10 | 0  |
 
+    Scenario: Testbot - Travel time matrix of minimal network with excludes
+        Given the query options
+            | exclude  | toll        |
+
+        Given the node map
+            """
+            a b
+            c d
+            """
+
+        And the ways
+            | nodes | highway  | toll | #                                        |
+            | ab    | motorway |      | not drivable for exclude=motorway        |
+            | cd    | primary  |      | always drivable                          |
+            | ac    | highway  | yes  | not drivable for exclude=motorway exclude=toll and exclude=motorway,toll |
+            | bd    | highway  | yes  | not drivable for exclude=motorway exclude=toll |
+
+        When I request a travel time matrix I should get
+            |   | a  | b  | c  | d  |
+            | a | 0  | 15 |    |    |
+            | b | 15 | 0  |    |    |
+            | c |    |    | 0  | 10 |
+            | d |    |    | 10 | 0  |
+
+    Scenario: Testbot - Travel time matrix of minimal network with different exclude
+        Given the query options
+            | exclude  | motorway  |
+
+        Given the node map
+            """
+            a b
+            c d
+            """
+
+        And the ways
+            | nodes | highway  | toll | #                                        |
+            | ab    | motorway |      | not drivable for exclude=motorway        |
+            | cd    | primary  |      | always drivable                          |
+            | ac    | highway  | yes  | not drivable for exclude=motorway exclude=toll and exclude=motorway,toll |
+            | bd    | highway  | yes  | not drivable for exclude=motorway exclude=toll |
+
+
+        When I request a travel time matrix I should get
+            |   | a  | b  | c  | d  |
+            | a | 0  | 40 | 15 | 25 |
+            | b | 40 | 0  | 25 | 15 |
+            | c | 15 | 25 | 0  | 10 |
+            | d | 25 | 15 | 10 | 0  |
+
+    Scenario: Testbot - Travel time matrix of minimal network with excludes combination
+        Given the query options
+            | exclude  | motorway,toll  |
+
+        Given the node map
+            """
+            a b
+            c d
+            """
+
+        And the ways
+            | nodes | highway  | toll | #                                        |
+            | ab    | motorway |      | not drivable for exclude=motorway        |
+            | cd    | primary  |      | always drivable                          |
+            | ac    | highway  | yes  | not drivable for exclude=motorway exclude=toll and exclude=motorway,toll |
+            | bd    | highway  | yes  | not drivable for exclude=motorway exclude=toll |
+
+        When I request a travel time matrix I should get
+            |   | a  | b  | c  | d  |
+            | a | 0  | 10 | 0  | 10 |
+            | b | 10 | 0  | 10 | 0  |
+            | c | 0  | 10 | 0  | 10 |
+            | d | 10 | 0  | 10 | 0  |
+
     Scenario: Testbot - Travel time matrix with different way speeds
         Given the node map
             """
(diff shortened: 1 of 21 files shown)


ghoshkaj:
@oxidase is [this declaration](https://github.com/Project-OSRM/osrm-backend/pull/4989/commits/5c16c190045e5ab3a1f516404ad3993d09d66704#diff-abfe271e1786ad59bfe8f6dc65396b88R50) of the cache okay? I'm wondering whether it's threadsafe or not and if it can lead to complications over [here](https://github.com/Project-OSRM/osrm-backend/pull/4989/commits/5c16c190045e5ab3a1f516404ad3993d09d66704#diff-3765a18558eecd8c9e22fbf48fb2eccbR98).

ghoshkaj:
@oxidase I didn't make the cache a singleton since all the accesses to the cache are through the SearchEngine. Instead I got rid of the `boost::thread_specific_ptr` and used a `std::unique_ptr` instead to try and simulate a singleton. However, I'm not sure that it is threadsafe for over here when [we create a new cache for a new dataset (timestamp)](https://github.com/Project-OSRM/osrm-backend/pull/4989/commits/5c16c190045e5ab3a1f516404ad3993d09d66704#diff-3765a18558eecd8c9e22fbf48fb2eccbR98).

ghoshkaj:
After talking with @danpat, I converted the cache pointer into a cache object: https://github.com/Project-OSRM/osrm-backend/pull/4989/commits/f39e37afbb46a5365c4f5c4d6e31a7ab22d61ec2. But tests were still failing non-determiistically.

Then I talked to @miccolis and went with putting a lock in front of the call to clear the function from within the SearchEngine: https://github.com/Project-OSRM/osrm-backend/pull/4989/commits/01e65629ff66663106cb1f24471a1cf5ab9d6e6e. The tests have stopped failing, but this probably makes it slower.

Some conversation came up around whether to make this shared storage cache be inter-thread shared storage versus inter-process shared storage. This implementation is of inter-thread shared storage as of now. Decisions to be made after performance measurements between threadlocal vs inter-thread shared storage. Hopefully we don't have to venture into inter-process shared storage 😬 🤞

TheMarex:
One consideration for using a shared storage is that there can be multiple datasets (usually two) in flight at the same time. This is not a problem for thread-local caching, since a single thread will either only use the new or the old data. For a shared cache we need to keep track of the dataset used to compute every entry in the cache, that means we need to extend the cache line to include the timestamp.

The full timestamp used by OSRM is a `32bit` integer but for the cache it is really only important to be able to differentiate between the datasets still in use. Typically we limit the number of datasets to 2 so in theory `1` bit of information would be enough. For the sake of generality I would suggest we use `1` byte (== 255 possible datasets) to start and optimize this later.

The cache design would need to be changed in the following way:

1. Each cache entry gets an additional value `generation` of one `unsigned char` (== 1 byte)
2. At the beginning of every query we obtain the `generation` for a given `timestamp` from the cache.
3. When inserting a value into the cache the user needs to pass in the generation with the value.
4. When retrieving a value from the cache the user needs to pass in the generation as well.
5. The generation would be used to compute the hash for the key in `unordered_map`.

ghoshkaj:
@TheMarex the steps above are really clever and adding `unsigned char generation` as part of the hash would have been so cool, except that with the vendored version of the LRU cache that I'm using, I don't have access to the hashing function anymore, so I can' t add the `generation` to be part of the unique key signature. Do you have a suggestion for a way around this problem? For now, I've just added the full `timestamp` as part of the key in the line, but that take up an additional 8 bytes to the cache.

TheMarex:
>  LRU cache that I'm using, I don't have access to the hashing function anymore

You can do it the same way you currently handle the full timestamp value by making it part of the key. The generation logic can live in the interface that you build around the LRU cache. Seems like boost automatically generates a hash function for a `std::tuple`, so extending the actual has function is obsolete.

ghoshkaj:
@TheMarex Ooo, actually I could just hash it myself and pass the hash to the library as the key!! Going to do that!

ghoshkaj:
branch-gitsha \ dataset, matrix-size, numReqs | usca-25x25-10k reqs (ms) | usca-25x25-10k reqs (ms) | usca-25x25-10k reqs (ms) | usca-25x25-10k reqs (ms) | usca-100x100-5k reqs (ms) | usca-100x100-5k reqs (ms) |usca-100x100-5k reqs (ms) |  usca-100x100-5k reqs (ms) | berlin-latest -5x5 - 20k reqs (ms) | berlin-latest -5x5 - 20k reqs (ms) | berlin-latest -5x5 - 20k reqs (ms) |  
-- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | --
  | 2 threads | 4 threads | 8 threads | 16 threads | 2 threads | 4 threads | 8 threads | 16 threads | 2 threads | 4 threads | 8 threads | 16 threads
master | 20.556 | 20.807 | 20.750 | 20.755 | 97.250 | 97.154 | 97.017 | 96.763 |   |   |   |  
threadlocal-cache (250 mb) |   |   |   |   |   |   |   |   |   |   |   |  
shared-storage (250mb) |   |   |   |   |   |   |   |   |   |   |   |  
threadlocal-cache (500 mb) | 31.592 | 31.833 | 32.497 | 33.065 | 226.447 | 226.923 | 431.501 | 225.378 |   |   |   |  
shared-storage (500 mb) |   |   |   |   |   |   |   |   |   |   |   |  
threadlocal-cache (1000 mb) |   |   |   |   |   |   |   |   |   |   |   |  
shared-storage (1000 mb) |   |   |   |   |   |   |   |   |   |   |   |  

Running osrm-runner with this command, adjusting lengh of request and number of requests as appropriate:
```
./scripts/osrm-runner.js -n 10000 -b -122.65,37.26,-119.25,39.83  -p '/table/v1/driving/{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{}' -f '$.durations[0]'
```
For threads, using the `- t` parameter on `./osrm-routed`

PR #4990, opened 2018-04-02
Distances in matrix

ghoshkaj:
# Issue

This PR attempts to implement distances in the matrix as in this issue: https://github.com/Project-OSRM/osrm-backend/issues/1353

## Tasklist
 - [ ] implement scaffolding for distances inside many_to_many_ch.hpp to use the unpacking cache
 - [ ] implement distance calculating function in routing_base.hpp (currently doesn't work)
 - [ ] add cucumber tests
 - [ ] add unit tests
 - [ ] pipeline through to node bindings for the request to be made and response
 - [ ] review
 - [ ] adjust for comments

## Requirements / Relations

The unpacking cache in https://github.com/Project-OSRM/osrm-backend/pull/4876 needs to be merged in to master first.


TheMarex:
> the difficulty I'm having is in supplying the second Node ID (the third argument to computeEdgeDistance(const FacadeT &facade, NodeID node_id_1, NodeID node_id_2)). How do I get the correct second NodeID to get the coordinate and perform the HaversineDistance on the pair of coodinates?

We need to do the following things:

1. Call [GetUncompressedForwardGeometry](https://github.com/Project-OSRM/osrm-backend/blob/master/include/engine/datafacade/contiguous_internalmem_datafacade.hpp#L570-L575) using the ID value that you would also call `GetUncompressedForwardDurations` with.
2. Iterate over every pair of adjacent nodes and call [GetCoordinateOfNode](https://github.com/Project-OSRM/osrm-backend/blob/master/include/engine/datafacade) o/contiguous_internalmem_datafacade.hpp#L560) of for both nodes to obtain their coordinate.
3. The calculate the distance between them using `HaversineDistance`.

ghoshkaj:
> 1. Call GetUncompressedForwardGeometry using the ID value that you would also call GetUncompressedForwardDurations with.
> 2. Iterate over every pair of adjacent nodes and call GetCoordinateOfNode o/contiguous_internalmem_datafacade.hpp#L560) of for both nodes to obtain their coordinate.
> 3. The calculate the distance between them using HaversineDistance.

I did this and it's working these are the results I have so far:

```
distance table:
       a        b         c        d
a      0,       199.97,   99.9924, 199.97
b      99.9924, 0,        99.9924, 0
c      0,       99.9772,  0,       99.9772
d      99.9924, 0,        99.9924, 0
```

for this scenario:
```
 Scenario: Testbot - Travel time matrix of minimal network with different exclude
        Given the query options
            | exclude  | motorway  |

        Given the node map
            """
            a b
            c d
            """

        And the ways
            | nodes | highway  | toll | #                                        |
            | ab    | motorway |      | not drivable for exclude=motorway        |
            | cd    | primary  |      | always drivable                          |
            | ac    | highway  | yes  | not drivable for exclude=motorway exclude=toll and exclude=motorway,toll |
            | bd    | highway  | yes  | not drivable for exclude=motorway exclude=toll |


        When I request a travel time matrix I should get
            |   | a  | b  | c  | d  |
            | a | 0  | 40 | 15 | 25 |
            | b | 40 | 0  | 25 | 15 |
            | c | 15 | 25 | 0  | 10 |
            | d | 25 | 15 | 10 | 0  |
```

@TheMarex so I think what's left now is the offset calculations. I looked at this a bit, and I see that we use `source_phantom.GetForwardDuration()` but there is no analogous function or member functions for distance_offset already in the PhantomNode class. I followed the path to see where the duration_offsets are calculated and they are calculated here in [geospatial_query.hpp](https://github.com/Project-OSRM/osrm-backend/blob/2690dd0621327ae3374a00d5fbded0861d72f5b0/include/engine/geospatial_query.hpp#L461). It calls on the `forward_duration_vector` and following this back further the dataface which gets this information from [`segment_data`](https://github.com/Project-OSRM/osrm-backend/blob/2690dd0621327ae3374a00d5fbded0861d72f5b0/include/engine/datafacade/contiguous_internalmem_datafacade.hpp#L584) which is ultimately populated by the [extractor](https://github.com/Project-OSRM/osrm-backend/blob/2690dd0621327ae3374a00d5fbded0861d72f5b0/include/engine/datafacade/contiguous_internalmem_datafacade.hpp#L468).

So to get the distance offsets, is the extractor code next in line to be augmented? Maybe in the [`segment_data_container`](https://github.com/Project-OSRM/osrm-backend/blob/2690dd0621327ae3374a00d5fbded0861d72f5b0/include/extractor/segment_data_container.hpp#L225)? 

Or is this the wrong track? 😄

TheMarex:
> So to get the distance offsets, is the extractor code next in line to be augmented? Maybe in the segment_data_container?

We can replicate [this](https://github.com/Project-OSRM/osrm-backend/blob/2690dd0621327ae3374a00d5fbded0861d72f5b0/include/engine/geospatial_query.hpp#L458-L462) logic and loop over all node ids up until the phantom node segment and then do a similar calculation as the one you implemented for the annotations to sum up the distances.

One particular snag is that we need to add the distance to the phantom node location. Suppose we have a phantom node that is snapped to an edge-based-node that consists of four segments:

```
      x
0---1----2----3----4
```

The whole distance of the edge-based-node would be the haversine distance of the node-based-nodes `0 -> 1 -> 2 -> 3 -> 4`. To account for the fact that the phantom node `(x)`  actually snapped to somewhere between 1 and 2 we need to subtract the distance `0 -> 1 -> x` from the whole distance value. Similar for the backwards offset we need to calculate `x -> 2 -> 3 -> 4`.

ghoshkaj:
@danpat we don't put that part of the calculation in the cache. Instead we do those offset calculations in the main body of the many_to_many function. For instance, for durations, those calculations happen here: https://github.com/Project-OSRM/osrm-backend/pull/4876/files#diff-81e2dc0f0a96c3d1a3fd7e43ee98a7f3R328

All the cache stuff happens 7 lines above, over here: https://github.com/Project-OSRM/osrm-backend/pull/4876/files#diff-81e2dc0f0a96c3d1a3fd7e43ee98a7f3R321

ghoshkaj:
Okay so with the latest commit: https://github.com/Project-OSRM/osrm-backend/pull/4990/commits/ea8cdc7354af7ef06c4b039ad47d485d7a4c8ed1 I'm running into this compile error that I suspect comes from packing problems. 

```
In file included from /Users/kajari/Mapbox/osrm-backend/include/engine/hint.hpp:31:
/Users/kajari/Mapbox/osrm-backend/include/engine/phantom_node.hpp:53:75: error: field 'bearing' will be initialized after field 'forward_distance' [-Werror,-Wreorder]
          is_valid_reverse_source{false}, is_valid_reverse_target{false}, bearing(0),
                                                                          ^
/Users/kajari/Mapbox/osrm-backend/include/engine/phantom_node.hpp:177:61: error: field 'bearing' will be initialized after field 'forward_distance' [-Werror,-Wreorder]
          is_valid_reverse_target{is_valid_reverse_target}, bearing{bearing},
                                                            ^
/Users/kajari/Mapbox/osrm-backend/include/engine/phantom_node.hpp:212:1: error: static_assert failed "PhantomNode has more padding then expected"
static_assert(sizeof(PhantomNode) == 128, "PhantomNode has more padding then expected");
^             ~~~~~~~~~~~~~~~~~~~~~~~~~~
In file included from /Users/kajari/Mapbox/osrm-backend/src/server/service/match_service.cpp:3:
In file included from /Users/kajari/Mapbox/osrm-backend/include/server/api/parameters_parser.hpp:4:
In file included from /Users/kajari/Mapbox/osrm-backend/include/engine/api/base_parameters.hpp:33:
/Users/kajari/Mapbox/osrm-backend/include/engine/hint.hpp:66:1: error: static_assert failed "Hint is bigger than expected"
static_assert(sizeof(Hint) == 128 + 4, "Hint is bigger than expected");
^             ~~~~~~~~~~~~~~~~~~~~~~~
/Users/kajari/Mapbox/osrm-backend/include/engine/hint.hpp:68:1: error: static_assert failed "ENCODED_HINT_SIZE does not match size of Hint"
static_assert(ENCODED_HINT_SIZE / 4 * 3 >= sizeof(Hint),
^             ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
5 errors generated.
```

The datatype that I've added `EdgeDistance` is a double, which is 8 bytes, and I've added 4 new members of this type, so the size of PhantomNode should have increased by 64 bytes. 
So [on this line](https://github.com/Project-OSRM/osrm-backend/pull/4990/commits/ea8cdc7354af7ef06c4b039ad47d485d7a4c8ed1#diff-29eca631a371a1ceb1f4187745f89bf1R212) I changed it to 128, but as you can see, the compiler is not too happy about that either!

Questions: 
1) Is that number not in bytes?
2) What does ENCODED_HINT_SIZE in `constexpr std::size_t ENCODED_HINT_SIZE = 92;` mean?

The other problem is that no matter where in the PhantomNode constructor, I place the new data members, the compiler keeps warning me about the order in which the data members will be initialized in. 

3) I'm pretty sure that I need to take packing into consideration, but not sure how?

ghoshkaj:
Oookay, got the answers:
1) I was doing the maths wrong. It's actually 4 * 8 = 32 btes extra, so it should have been 64 + 32 = 96 bytes.
2) I got the advice to change `EdgeDistance` from a `double` to a `float` as it suffices for the distances we'll encounter (which is about [half the world's circumference](http://map.project-osrm.org/?z=2&center=24.046464%2C98.789063&loc=62.267923%2C162.773438&loc=-39.909736%2C27.421875&hl=en&alt=0) which is 25763.7 km so using a single precision value that encodes `2^24=16777216` ~ 16km is enough. cc @oxidase ). So I've changed `EdgeDistance` to a float and adjust the padding accordingly.
3) The order of the class members not being in the same order as in the constructor causes the compiler to throw up warnings, and the current compiler settings are set to Error on Warnings. 

All these problems are fixed for now!

ghoshkaj:
Okay, so here's the latest, now that I've added put the offset functions into use:
For this scenario:
```
Scenario: Testbot - Travel time matrix of minimal network with different exclude
        Given the query options
            | exclude  | motorway  |

        Given the node map
            """
            a b
            c d
            """

        And the ways
            | nodes | highway  | toll | #                                        |
            | ab    | motorway |      | not drivable for exclude=motorway        |
            | cd    | primary  |      | always drivable                          |
            | ac    | highway  | yes  | not drivable for exclude=motorway exclude=toll and exclude=motorway,toll |
            | bd    | highway  | yes  | not drivable for exclude=motorway exclude=toll |


        When I request a travel time matrix I should get
            |   | a  | b  | c  | d  |
            | a | 0  | 40 | 15 | 25 |
            | b | 40 | 0  | 25 | 15 |
            | c | 15 | 25 | 0  | 10 |
            | d | 25 | 15 | 10 | 0  |
```

I get this table:
```
distance table:
	a 			b 			c 			d
a	0, 			199.97, 	99.9924, 	199.97
b	99.9924, 	0, 			99.9924, 	99.9924
c	0, 			99.9772, 	0, 			99.9772
d	0, 			0, 			0, 			0
```

ghoshkaj:
Distances on CH works! I've built out the pipeline through to the node bindings and have added cucumber and table tests. Here are a few:
```
  ✔ Given the profile "testbot"
  ✔ And the partition extra arguments "--small-component-size 1 --max-cell-sizes 2,4,8,16"
  ✔ Given the node map
      """
      a b
      """
  ✔ And the ways
      | nodes |
      | ab    |
  ✔ When I request a travel time matrix I should get
      |   | a  | b  |
      | a | 0  | 10 |
      | b | 10 | 0  |
  ✔ When I request a travel distance matrix I should get
      |   | a        | b        |
      | a | 0        | 9.997721 |
      | b | 9.997721 | 0        |

  ✔ Given the profile "testbot"
  ✔ And the partition extra arguments "--small-component-size 1 --max-cell-sizes 2,4,8,16"
  ✔ Given the node map
      """
      a b c d
      """
  ✔ And the ways
      | nodes | highway   |
      | ab    | primary   |
      | bc    | secondary |
      | cd    | tertiary  |
  ✔ When I request a travel time matrix I should get
      |   | a  | b  | c  | d  |
      | a | 0  | 10 | 30 | 60 |
      | b | 10 | 0  | 20 | 50 |
      | c | 30 | 20 | 0  | 30 |
      | d | 60 | 50 | 30 | 0  |
  ✔ When I request a travel distance matrix I should get
      |   | a         | b         | c         | d         |
      | a | 0         | 9.997721  | 19.995442 | 29.993164 |
      | b | 0         | 0         | 9.997721  | 19.995442 |
      | c | 9.997721  | 9.997721  | 0         | 9.997721  |
      | d | 29.993164 | 29.993164 | 19.995442 | 0         |
  ✔ When I request a travel time matrix I should get
      |   | a | b  | c  | d  |
      | a | 0 | 10 | 30 | 60 |
  ✔ When I request a travel distance matrix I should get
      |   | a | b        | c         | d         |
      | a | 0 | 9.997721 | 19.995442 | 29.993164 |
  ✔ When I request a travel time matrix I should get
      |   | a  |
      | a | 0  |
      | b | 10 |
      | c | 30 |
      | d | 60 |
  ✔ When I request a travel distance matrix I should get
      |   | a         |
      | a | 0         |
      | b | 0         |
      | c | 9.997721  |
      | d | 29.993164 |
```

@TheMarex if you'd give this a review that would be good.

TheMarex:
@ghoshkaj Great work! Before I can review this, the PR needs to be rebased on `master` and the base branch this PR is merged into need to be chaned to `master`.

ghoshkaj:
In ` ./unit_tests/server-tests` I have a weird failing test case:
```
00:01 $ ./unit_tests/server-tests
Running 13 test cases...
/Users/kajari/Mapbox/osrm-backend/unit_tests/server/parameters_parser.cpp:96: error: in "api_parameters_parser/valid_route_hint": check hint.phantom.input_location == util::Coordinate(util::FloatLongitude{13.38886}, util::FloatLatitude{52.517033}) has failed [{13.38886, 13.38886} != {13.38886, 13.38886}]
```

**Question**: any suggestions as to why this is happening? It could be a type thing, but I haven't changed anything about the test other than the hint string itself, so it should have failed before too if it was a type problem.

For the other `hint` tests that I've currently commented out in the server_tests file, I intend to generate new hints using my new Phantom Node structure (it has two floats members more in it, thus nullifying all the existing hint tests). However, I'm not sure about how to generate phantom nodes with coordinates like the following: 
```
 std::vector<util::Coordinate> coords_1 = {{util::FloatLongitude{1}, util::FloatLatitude{2}},
                                              {util::FloatLongitude{3}, util::FloatLatitude{4}}};
```
**Question**: How were these original hints generated? I can't figure out what osrm file I would run this request against! Perhaps a cucumber scenario? Or a hand-made osm file?

Another failing test is the following, and I'm not sure how to follow the path from the call site to the failing case, but I suspect it has something to do with the vector range changes. @oxidase do you have an idea about this?
```
The command "./unit_tests/engine-tests" exited with 0.
8.95s$ ./unit_tests/util-tests
Running 119 test cases...
[assert][140136739313600] /usr/include/boost/range/iterator_range_core.hpp:354
in: boost::iterator_range<IteratorT>::abstract_value_type boost::iterator_range<IteratorT>::operator()(boost::iterator_range<IteratorT>::difference_type) const [with IteratorT = osrm::util::VectorViewIterator<const unsigned int>; boost::iterator_range<IteratorT>::abstract_value_type = unsigned int; boost::iterator_range<IteratorT>::difference_type = long int]: at >= 0 && at < size()
unknown location(0): fatal error in "bearing_tests": signal: SIGABRT (application abort requested)
/home/travis/build/Project-OSRM/osrm-backend/unit_tests/util/static_rtree.cpp(357): last checkpoint
```

cc @TheMarex

TheMarex:
> Question: any suggestions as to why this is happening? It could be a type thing, but I haven't changed anything about the test other than the hint string itself, so it should have failed before too if it was a type problem.

I think this might be the hint-decoding breaking because it interprets the hint as a raw memory dump of the phantom node, which will just break if you change the phantom node at all. So this error is expected.

> Question: How were these original hints generated? I can't figure out what osrm file I would run this request against! Perhaps a cucumber scenario? Or a hand-made osm file?

These strings were generated by hand calling the `Hint::ToBase64()` method. Looking at the current test scenario I would propose replacing the hard coded string with encoding it on fly, looks like I was too lazy for that when writing the test cases. I'll run out this change to unblock you here.

TheMarex:
Looking at the current implementation we need to adapt two larger things to make this mergable:

1. Computing `distance` values needs to be optional, and should have (little) performance impact of you don't need it. This is important for being able to ship this now, independently of the `durations` refactor.
2. There is now API-level integration yet. We need to implement an `annotations=` parameter for the table plugin.

### Making it optional

We need to find a way how we can still utilize the `duration` annotation on the edge (== fast for the non-distance case) but still support extracting a path if needed to compute the distance value.

For this I propose the following approach:
1. Save both the `weight`, `duration` and `middle_node` in the `NodeBucket`/`ManyToManyHeap` (currently we don't save the middle node).
2. We need to add a parameter to `manyToMany()` that tells us if we need to compute distance values or not.
3. If we don't compute distance values, don't try to extract the packed/unpacked path. It will work similar to the current implementation in `master` that gets the `duration` value from the edges/heap.
4. If a user wants to compute the distance value, we need to proceed like implemented here:
    - Retrieve the path
    - Unpack the path
    - Annotate the path with distance (we get the `duration` anyway)

#### Adding the API parameter

We want to add a parameter called `annotations` that supports the following values:
- `duration` (default)
- `duration,distance` (compute both the duration and the distance and return it)
- `distance` (compute the distance and return it)

For this we need to extend a few objects:
- [`TableParameters`](https://github.com/Project-OSRM/osrm-backend/blob/master/include/engine/api/table_parameters.hpp) needs to get an `annotations` property that work like the one in `RouteParameters` [here](https://github.com/Project-OSRM/osrm-backend/blob/master/include/engine/api/route_parameters.hpp#L143)
- [`TableParameterGrammar`](https://github.com/Project-OSRM/osrm-backend/blob/master/include/server/api/table_parameter_grammar.hpp#L25) needs to be extended to parse the `annotations=....` URL option. For this you can emulate the way `annotations` work in the `RouteParameterGrammar` [here](https://github.com/Project-OSRM/osrm-backend/blob/master/include/server/api/route_parameters_grammar.hpp#L100), [here](https://github.com/Project-OSRM/osrm-backend/blob/master/include/server/api/route_parameters_grammar.hpp#L51-L58), [here](https://github.com/Project-OSRM/osrm-backend/blob/master/include/server/api/route_parameters_grammar.hpp#L68-L71) and [here](https://github.com/Project-OSRM/osrm-backend/blob/master/include/server/api/route_parameters_grammar.hpp#L83-L85)
- The node bindings need to be extended to allow for this value. For this we need to extend [argumentsToTableParameters](https://github.com/Project-OSRM/osrm-backend/blob/master/include/nodejs/node_osrm_support.hpp#L976-L978) to handle annotations like [here](https://github.com/Project-OSRM/osrm-backend/blob/master/include/nodejs/node_osrm_support.hpp#L683-L744)

ghoshkaj:
> I haven't looked into the implementation, but what is the reason for a distance like 29.993164 instead of 30?
@emiltin it was because I hadn't yet rounded the values up. I have done so now. There may still be some need for the fuzzy matching as you suggest above -- I'm hoping that the rounding takes care of that, but will add fuzzy matching if needs me. Thank you for your suggestion of it!

ghoshkaj:
@chaupow if you wanted to take a look -- there are a couple of things that are failing at the moment and I haven't been able to figure out how to get them to pass: 
1) I added these three cucumber scenarios with exclude flags in that begin [on this line](https://github.com/Project-OSRM/osrm-backend/pull/4990/files#diff-f5bde3e2d53de15aff44adf34253fac6R29). They work for the distance table but not the duration table. When I wrote them, I was using the unpack + annotate method for both distances and durations, so they passed.  Would you look into them and verify whether or not the distance matrices should be populated? For the parts of the route that are tagged with exclude flags, I think they shouldn't pass and if that is the case then `master` currently has a problem!

2) There's a failing unit test in the util-tests that can be fixed by[ getting this function to return a valid NodeForwardRange vector](https://github.com/Project-OSRM/osrm-backend/pull/4990/files#diff-525a7f764c6b5a72227d414fe4a4e476R57). You can see my attempts at getting this to pass, but I have been thwarted because it is late at night. 
(Just in case you've forgotten, to run unit tests: `make -j 4 tests` within the build folder, then to run them: `./unit_tests/util-tests` )

3) If you get these working, everything else should be running. If so, could you run some measurements? You can use `osrm-runner` to generate 5000 requests, and check what the average roundtrip speeds are. Do this on `master` for a baseline and then this branch to see how much slower the requests will get if this is merged in! (@TheMarex can show you how!)

TheMarex:
@ghoshkaj I found the problem for the exclude failure (query options were overwritten so `exclude` was not actually passed to the server). However, it seems like there are still failures for the distance test cases?

chaupow:
```
╰─ ./unit_tests/util-tests
Running 119 test cases...

*** No errors detected
```

🎉  I think this is Patrick's work ^

ghoshkaj:
> However, it seems like there are still failures for the distance test cases?
That file used to pass when distances were returned by default, so I fixed it to pass in `annotations` as a request parameter. However, there is something going wrong from node to cpp because the error is now: [`this 'annotations' param is not supported` ](https://travis-ci.org/Project-OSRM/osrm-backend/jobs/365583407#L1720).

That error is generated [here](https://github.com/Project-OSRM/osrm-backend/pull/4990/files#diff-89873aae5c93116cd0f09f2e1a69c320R1101) so something is afoot from node -> osrm.

@TheMarex @chaupow

chaupow:
```
╰─ cucumber features/testbot/distance_matrix.feature

17 scenarios (17 passed)
116 steps (116 passed)
0m16.356s
```

chaupow:
Ok summarizing what I did:

i added two sanity checks:
- [a cucumber test](https://github.com/Project-OSRM/osrm-backend/pull/4990/commits/c25d59654bec47b864ab9aafcc3f5d7e1da78c70) checking the distance
- [an assertion](https://github.com/Project-OSRM/osrm-backend/pull/4990/commits/1b27785fcce1ff4d865fff5e7af791f43108daa5) for forward and reverse geometry

what I tried is [adding the distance of the last node](https://github.com/Project-OSRM/osrm-backend/pull/4990/commits/4e2e8c5c18bc3f5410c14e7f3f7b73067bf58d76)

I haven't tried running more cucumber tests. If adding the distance of the last node actually helps in other tests compared to before, it might be a good start.

Next Pointers:
So while looking at `calculateEBGNodeAnnotations`, here are the things that are suspicious to me (doesn't mean they have to be incorrect but imho worth investigating). Maybe you want to have a look at it @ghoshkaj ?

* https://github.com/Project-OSRM/osrm-backend/blob/distances/include/engine/routing_algorithms/routing_base_ch.hpp#L355-L357

Do we need this?
```
                recursion_stack.emplace(edge);
```
we talked about this too https://github.com/Project-OSRM/osrm-backend/blob/distances/include/engine/routing_algorithms/routing_base_ch.hpp#L372-L384

and it might be because of the cache but do we still use it here? can be remove just for the sake of clarity? doesnt look wrong, but...then we wouldnt need the `bool` in the tuple edge either.

ghoshkaj:
**Things I did**
- Found the reason for some of the test failures -- there were some off by one errors in calculating the [forward_distance_offset](https://github.com/Project-OSRM/osrm-backend/pull/4990/commits/2f7df4f0d310b3c8e95085a0a53d6e97eda2bec6#diff-047e157684260be4ffdea79c2ed2fde5R481) and the [reverse_distance_offset](https://github.com/Project-OSRM/osrm-backend/pull/4990/commits/2f7df4f0d310b3c8e95085a0a53d6e97eda2bec6#diff-047e157684260be4ffdea79c2ed2fde5R508) for the Phantom Nodes.

- While these distance offset functions now work well, they're not very clean, so I tried to [create a templatized function](https://github.com/Project-OSRM/osrm-backend/pull/4990/commits/2f7df4f0d310b3c8e95085a0a53d6e97eda2bec6#diff-047e157684260be4ffdea79c2ed2fde5R31) for them. However, there are several overrides of underlying vectors (VectorView and PackedVector) and I need some help figuring out how to pass in the correct IteratorType (instead of `ForwardIt`). The idea is to be able to have a function signature similar to the [std::accumulate functions above](https://github.com/Project-OSRM/osrm-backend/pull/4990/commits/2f7df4f0d310b3c8e95085a0a53d6e97eda2bec6#diff-047e157684260be4ffdea79c2ed2fde5R468). You can see the errors by uncommenting [the call site](https://github.com/Project-OSRM/osrm-backend/pull/4990/commits/2f7df4f0d310b3c8e95085a0a53d6e97eda2bec6#diff-047e157684260be4ffdea79c2ed2fde5R477).

- [Dropped the profile changes](https://github.com/Project-OSRM/osrm-backend/pull/4990/commits/2f7df4f0d310b3c8e95085a0a53d6e97eda2bec6#diff-047e157684260be4ffdea79c2ed2fde5R31) in the last cuke test because the different weighting is causing a different and longer more roundabout path to be chosen [for a->c in this test resulting in a distance of 799 m or so](https://github.com/Project-OSRM/osrm-backend/pull/4990/commits/2f7df4f0d310b3c8e95085a0a53d6e97eda2bec6#diff-f5bde3e2d53de15aff44adf34253fac6R454).

- Modified the exclude flag scenarios to test the exclude flags more rigourously but changing `motorway` to `primary` [like this](https://github.com/Project-OSRM/osrm-backend/pull/4990/commits/2f7df4f0d310b3c8e95085a0a53d6e97eda2bec6#diff-13c2e9678daa7b95791bf4e370ca0fa8R63).

- Changed the names of the exclude scenarios to better isolate the tests (for single runs).

- Changed the above two things in the scenarios in the [duration_matrix.feature file](https://github.com/Project-OSRM/osrm-backend/pull/4990/commits/2f7df4f0d310b3c8e95085a0a53d6e97eda2bec6#diff-13c2e9678daa7b95791bf4e370ca0fa8) correspondingly.

**Next Steps**
Currently there failing cucumber tests down to 3 failing scenarios. 

- One set are the last two exclude flag scenarios. These also fail in `duration_matrix.feature`. When `toll` is excluded alone, the tests pass, but excluding motorway alone or excluding the combination of `toll` and `motorway` fails. Examine why and fix.

-  Examine and fix the last failing cucumber scenario:  `Testbot - Travel distance matrix based on segment durations`

- Clean the `calculateEBGNodeAnnotations` function of vestiges of keeping values for the cache, as mentioned under `Next Pointers` in [this comment](https://github.com/Project-OSRM/osrm-backend/pull/4990#issuecomment-381206826).

include/engine/geospatial_query.hpp
```diff
@@ -475,27 +475,21 @@ template <typename RTreeT, typename DataFacadeT> class GeospatialQuery
...
+                datafacade.GetCoordinateOfNode(*std::prev(current)),
+                datafacade.GetCoordinateOfNode(*current));
         }
 
         EdgeWeight forward_weight = forward_weights[data.fwd_segment_position];
         EdgeDuration forward_duration = forward_durations[data.fwd_segment_position];
         EdgeDistance forward_distance = util::coordinate_calculation::haversineDistance(
             datafacade.GetCoordinateOfNode(forward_geometry(data.fwd_segment_position)),
-            point_on_segment);
+            datafacade.GetCoordinateOfNode(forward_geometry(data.fwd_segment_position+1)));
```
ghoshkaj: @chaupow what you did here by removing calculating the distance from `data.fwd_segment_position` and `point_on_segment` is what multiplying by the ratio below recreates. `point_on_segment` is the location of the phantom_node.

features/support/shared_steps.js
```diff
@@ -103,6 +103,7 @@ module.exports = function () {
                             time = hasRoute && json.routes[0].duration,
                             weight = hasRoute && json.routes[0].weight;
 
+                        distance = Math.round(distance);
                         if (headers.has('distance')) {
```
ghoshkaj: ghoshkaj [7:46 PM]
@chaupow distances in in the Route are rounded here https://github.com/Project-OSRM/osrm-backend/blob/faff2c774d09a7227c77ae4fa40d22d54bb00b45/include/engine/guidance/assemble_leg.hpp#L214 and also here: https://github.com/Project-OSRM/osrm-backend/blob/faff2c774d09a7227c77ae4fa40d22d54bb00b45/src/engine/api/json_factory.cpp#L188

include/engine/geospatial_query.hpp
chaupow: Ok, I lacked content on what the point on segment is. But if it is the phantom node, I don't think this should be here? Looking at `forward_duration ` it looks like `forward_distance` should be the distance of the whole segment? I might be wrong tho. Can you try fixing this by using `point_on_segment` and not using the ratio then?

ghoshkaj:
Things left to do
- [x] rebase against master and collapsing the many commits
- [x] quick refactor and clean up
- [ ] add changelog and ensure all the documentation is updated. I've done some of the docs, but I haven't done a sweep through to make sure it's all there
- [ ] adjust the cucumber tests to work with decimals
- [ ] quick pre-review
- [ ] run measurements comparing master and PR for durations
- [ ] run measurements on PR for durations+distances

For measurements use `osrm-runner`. Run 5000 requests of 100x100 matrix, piping the requests into a file as well as the number of seconds that the requests took. Find an average of the 5000 request times. Use http://download.geofabrik.de/north-america/us-west.html with a bounding box of: `-124.94, 33.01, -108.95, 47.5` [bounding box visual here](http://tools.geofabrik.de/calc/#type=geofabrik_standard&bbox=-124.935571,33.010914,-108.950218,47.490515&tab=1&proj=EPSG:4326&places=2).

include/engine/api/table_api.hpp
```diff
@@ -116,13 +124,48 @@ class TableAPI final : public BaseAPI
                                {
                                    return util::json::Value(util::json::Null());
                                }
+                               // division by 10 below because the duration up until this
+                               // point in the code is in deciseconds (10s)
+                               // as duration is computed during contraction and stored as int
+                               // rather than float to save on memory usage
+                               // for the extract -> shortcut (CH/MLD) -> route pipeline
```
chaupow: I wonder if the comment is too long? Personally, the information that duration is deciseconds would suffice (so just the two first lines are enough).

Also maybe let's make this shorter 🙂

`division by 10 below because the duration up until this point in the code is in deciseconds (10s)`
->
`division by 10 because the duration is in deciseconds (10s)`

include/engine/api/table_api.hpp
```diff
@@ -65,7 +66,14 @@ class TableAPI final : public BaseAPI
         }
 
         response.values["durations"] =
-            MakeTable(durations, number_of_sources, number_of_destinations);
+            MakeTable(tables.first, number_of_sources, number_of_destinations);
```
chaupow: Let's rename this to `MakeDurationTable` so it is consistent with `MakeDistanceTable`

include/engine/api/table_api.hpp
```diff
@@ -116,13 +124,48 @@ class TableAPI final : public BaseAPI
...
+                           row_end_iterator,
+                           json_row.values.begin(),
+                           [](const EdgeDistance distance) {
+                               if (distance == INVALID_EDGE_DISTANCE)
+                               {
+                                   return util::json::Value(util::json::Null());
+                               }
+                               // std::round(distance * 100.)/100.) below is because we want to
+                               // predictably round to 2 decimal places
+                               // @TODO VERIFY THIS?!?!?!? @CHAUPOW
```
chaupow: Oh, let's remove the todo and change the whole comment to
```
// round to decimal point
```

include/engine/api/trip_parameters.hpp
```diff
@@ -59,6 +59,13 @@ struct TripParameters : public RouteParameters
         Any,
         Last
     };
+    enum class AnnotationsType
+    {
+        None = 0,
+        Duration = 0x01,
+        Distance = 0x02,
+        All = Duration | Distance
+    };
```
chaupow: can this be removed?

include/engine/geospatial_query.hpp
```diff
@@ -457,11 +460,21 @@ template <typename RTreeT, typename DataFacadeT> class GeospatialQuery
...
+            forward_distance_offset += util::coordinate_calculation::haversineDistance(
+                datafacade.GetCoordinateOfNode(*std::prev(current)),
+                datafacade.GetCoordinateOfNode(*current));
+        }
+
         EdgeWeight forward_weight = forward_weights[data.fwd_segment_position];
         EdgeDuration forward_duration = forward_durations[data.fwd_segment_position];
-
-        BOOST_ASSERT(data.fwd_segment_position <
-                     std::distance(forward_durations.begin(), forward_durations.end()));
```
chaupow: uh. did we remove this by accident? i think we should get it back

include/engine/geospatial_query.hpp
```diff
@@ -457,11 +460,21 @@ template <typename RTreeT, typename DataFacadeT> class GeospatialQuery
...
+        EdgeDistance forward_distance_offset = 0;
+        for (auto current = std::next(forward_geometry.begin());
+             current < forward_geometry.begin() + data.fwd_segment_position + 1;
+             ++current)
+        {
+            forward_distance_offset += util::coordinate_calculation::haversineDistance(
+                datafacade.GetCoordinateOfNode(*std::prev(current)),
+                datafacade.GetCoordinateOfNode(*current));
+        }
+
```
chaupow: if you agree, let's change this to

```
for (auto current = forward_geometry.begin();
   current < forward_geometry.begin() + data.fwd_segment_position;
   ++current)
{
  forward_distance_offset += util::coordinate_calculation::haversineDistance(
      datafacade.GetCoordinateOfNode(*current),
      datafacade.GetCoordinateOfNode(*std::next(current)));
}
```

i cant explain but i find it more intuitive. change if you agree :)

include/engine/geospatial_query.hpp
```diff
@@ -473,21 +486,38 @@ template <typename RTreeT, typename DataFacadeT> class GeospatialQuery
...
 
+        EdgeDistance reverse_distance_offset = 0;
+        for (auto current = std::next(forward_geometry.begin());
+             current < forward_geometry.end() - data.fwd_segment_position - 1;
+             ++current)
+        {
+            reverse_distance_offset += util::coordinate_calculation::haversineDistance(
+                datafacade.GetCoordinateOfNode(*std::prev(current)),
+                datafacade.GetCoordinateOfNode(*current));
+        }
```
chaupow: same with other for-loop

include/engine/routing_algorithms/routing_base.hpp
```diff
@@ -405,6 +406,23 @@ InternalRouteResult extractRoute(const DataFacade<AlgorithmT> &facade,
...
 }
 
+template <typename FacadeT> EdgeDistance computeEdgeDistance(const FacadeT &facade, NodeID node_id)
+{
+    const auto geometry_index = facade.GetGeometryIndex(node_id);
+
+    EdgeDistance total_distance = 0.0;
+
+    auto geometry_range = facade.GetUncompressedForwardGeometry(geometry_index.id);
+    for (auto current = std::next(geometry_range.begin()); current != geometry_range.end();
```
chaupow: my personal style nit-pick: change `!=` to `<`

include/engine/routing_algorithms/routing_base.hpp
```diff
@@ -405,6 +406,23 @@ InternalRouteResult extractRoute(const DataFacade<AlgorithmT> &facade,
...
+         ++current)
+    {
+        total_distance += util::coordinate_calculation::haversineDistance(
+            facade.GetCoordinateOfNode(*std::prev(current)), facade.GetCoordinateOfNode(*current));
+    }
```
chaupow: we can also change the for loop just like the others such that it will be 

`total_distance += util::coordinate_calculation::haversineDistance(facade.GetCoordinateOfNode(*current), facade.GetCoordinateOfNode(*std::next(current)));`

include/engine/routing_algorithms/routing_base_ch.hpp
```diff
@@ -288,6 +288,106 @@ void unpackPath(const DataFacade<Algorithm> &facade,
...
+{
+    // Make sure we have at least something to unpack
+    if (packed_path_begin == packed_path_end ||
+        std::distance(packed_path_begin, packed_path_end) <= 1)
+        return 0;
+
+    std::stack<std::tuple<NodeID, NodeID, bool>> recursion_stack;
+    std::stack<EdgeDistance> distance_stack;
+    // We have to push the path in reverse order onto the stack because it's LIFO.
+    for (auto current = std::prev(packed_path_end); current != packed_path_begin;
```
chaupow: change `!=` to `>`

include/engine/routing_algorithms/routing_base_ch.hpp
```diff
@@ -288,6 +288,106 @@ void unpackPath(const DataFacade<Algorithm> &facade,
...
+         current = std::prev(current))
+    {
+        recursion_stack.emplace(*std::prev(current), *current, false);
+    }
+
+    std::tuple<NodeID, NodeID, bool> edge;
```
chaupow: as we don't use the cache currently, we can remove `bool` (and then need to remove it from the following code as well)

src/engine/plugins/table.cpp
```diff
@@ -81,16 +81,23 @@ Status TablePlugin::HandleRequest(const RoutingAlgorithmsInterface &algorithms,
...
     auto snapped_phantoms = SnapPhantomNodes(phantom_nodes);
-    auto result_table =
-        algorithms.ManyToManySearch(snapped_phantoms, params.sources, params.destinations);
 
-    if (result_table.empty())
+    bool requestDistance = false;
+    if (params.annotations & api::TableParameters::AnnotationsType::Distance)
+    {
+        requestDistance = true;
+    }
```
chaupow: we could write this as
```
 bool requestDistance = params.annotations & api::TableParameters::AnnotationsType::Distance
```

src/engine/plugins/trip.cpp
```diff
@@ -216,61 +216,99 @@ Status TripPlugin::HandleRequest(const RoutingAlgorithmsInterface &algorithms,
     BOOST_ASSERT(snapped_phantoms.size() == number_of_locations);
 
     // compute the duration table of all phantom nodes
-    auto result_table = util::DistTableWrapper<EdgeWeight>(
-        algorithms.ManyToManySearch(snapped_phantoms, {}, {}), number_of_locations);
+    auto result_duration_table = util::DistTableWrapper<EdgeWeight>(
+        algorithms.ManyToManySearch(snapped_phantoms, {}, {}, /*requestDistance*/ false).first,
```
chaupow: oh i like this inline comment for input params

src/engine/plugins/trip.cpp
```diff
@@ -216,61 +216,99 @@ Status TripPlugin::HandleRequest(const RoutingAlgorithmsInterface &algorithms,
...
+    // // get an optimized order in which the destinations should be visited
+    // if (number_of_locations < BF_MAX_FEASABLE)
+    // {
+    //     distance_trip = trip::BruteForceTrip(number_of_locations, result_distance_table);
+    // }
+    // else
+    // {
+    //     distance_trip = trip::FarthestInsertionTrip(number_of_locations, result_distance_table);
+    // }
+
```
chaupow: we should clean up 🙂

src/engine/routing_algorithms/many_to_many_ch.cpp
```diff
@@ -144,29 +147,206 @@ void backwardRoutingStep(const DataFacade<Algorithm> &facade,
...
+    // sources  b  [3,  4,  5],
+    //  (rows)  c  [6,  7,  8],
+    //          d  [9, 10, 11]
+    // row_idx * number_of_targets + column_idx
+    // a -> c 0 * 3 + 2 = 2
+    // c -> c 2 * 3 + 2 = 8
+    // d -> c 3 * 3 + 2 = 11
+
+    //   middle_nodes_table = [0 , 1, 2, .........]
+
```
chaupow: umm. let's either clean up or write comments such that it is understandable

src/engine/routing_algorithms/many_to_many_ch.cpp
```diff
@@ -144,29 +147,206 @@ void backwardRoutingStep(const DataFacade<Algorithm> &facade,
...
+                    target_phantom.GetForwardDistance() - source_phantom.GetForwardDistance();
+                distances_table[row_idx * number_of_targets + column_idx] += offset;
+            }
+            else
+            {
+                //       ............       <-- calculateEGBAnnotation returns distance from 0 to 3
+                //         s<--------<--t   <-- GetReverseDistance() returns this offset
+                //       ---.........++++   <-- subtract source offset and add target offset
+                //          .............   <-- want this distance as result
+                // entry 0---1---2---3---   <-- 3 is exit node
```
chaupow: hurray for the asciis 🎉

chaupow:
> I verified that the failing cuke tests were just off by a single decimal point and changed them.

uh, there was one still failing for me. let's hope this is not a weird decimal rounding error that is different on different machines. can you maybe check whether the tests still pass for you?

>  But the formatting is terrible, could you please run your table cleaner plugin over them?

done :)

chaupow:
Good news:

We have almost no slowdown when requesting duration tables in current `master` and this PR:

experiment: run 5 times:
```
node scripts/osrm-runner.js -b -124.94,33.01,-108.95,47.5 -p "/table/v1/driving/{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{}" -f "$.durations[0]" -n 1000
```
So, we run 5000 100x100 duration table requests on us-west.

on master avg time in ms:
```
40.3629
40.4589
40.4351
40.1504
40.281
```

on this branch avg time in ms:
```
40.3629
40.4589
40.4351
40.1504
40.281
```

chaupow:
edit. pls ignore my last comment

chaupow:
Ok. Trying again.

Ran 5000 `100x100 duration matrix` requests in `us-west` using `osrm-runner`:
```
node scripts/osrm-runner.js -b -124.94,33.01,-108.95,47.5 -p "/table/v1/driving/{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{}" -f "$.durations[0]" -n 1000
```

### Average time on current `master`:
`95.16ms`

### Average time on current `PR`:
`98.97ms`

----

I then ran the first 200 requests on current `PR` requesting duration **and** distance tables:

### Average time on current `PR` with `annotations=distances`:
`11.73s`

chaupow:
### More numbers on matrices requests with `annotations=distance`

* `400ms` for 10x10 with distances
* `1.6s` for 20x20 with distances
* `3.8s` for 30x30 with distances
* `11.73s` for 100x100 with distances

chaupow:
### More numbers on matrices requests with `annotations=distance`

I ran more 100x100 matrix requests but limited the max-distance between coordinates to 20km and ran 1000 queries on LA and 1000 queries on Phoenix:

```
node scripts/osrm-runner-distancer.js -b -118.3859252,33.770015,-116.90002,34.229970 -p "/table/v1/driving/{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{}?annotations=distance" -f "$.durations[0]" -n 1000 -d 20000 > la-d20000-n1000.out &&  aws s3 cp la-d20000-n1000.out s3://mapbox-eu-west-1/playground/chaupow/
```
```
node scripts/osrm-runner-distancer.js -b -118.385925,33.770015,-116.9000,34.22997 -p "/table/v1/driving/{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{};{}?annotations=distance" -f "$.durations[0]" -n 1000 -d 20000 > phoenix-d20000-n1000.out &&  aws s3 cp phoenix-d20000-n1000.out s3://mapbox-eu-west-1/playground/chaupow/
```

#### Average query times:
* `3.73s` in LA, 100x100, max distance 20km
* `3.74s` in Phoenix, 100x100, max distance 20km

README.md
```diff
@@ -9,7 +9,7 @@ High performance routing engine written in C++14 designed to run on OpenStreetMa
 The following services are available via HTTP API, C++ library interface and NodeJs wrapper:
 - Nearest - Snaps coordinates to the street network and returns the nearest matches
 - Route - Finds the fastest route between coordinates
-- Table - Computes the duration of the fastest route between all pairs of supplied coordinates
+- Table - Computes the duration of the fastest route between all pairs of supplied coordinates, and optionally returns distances of these routes.
```
chaupow: Maybe just shorten this to 
`- Table - Computes the duration or distances of the fastest route between all pairs of supplied coordinates`

The other services don't list all their options either.

docs/http.md
```diff
@@ -222,20 +222,21 @@ curl 'http://router.project-osrm.org/route/v1/driving/13.388860,52.517037;13.397
 
 ### Table service
 
-Computes the duration of the fastest route between all pairs of supplied coordinates. Also returns the distances between the coordinate pairs. Note that the distances are not the shortest distance between two coordinates, but rather the distances of the fastest routes.
+Computes the duration of the fastest route between all pairs of supplied coordinates. Optinally, also returns the distances between the coordinate pairs. Note that the distances are not the shortest distance between two coordinates, but rather the distances of the fastest routes.
```
chaupow: Typo `Optionally`

docs/http.md
```diff
@@ -222,20 +222,21 @@ curl 'http://router.project-osrm.org/route/v1/driving/13.388860,52.517037;13.397
...
-**Coordinates**
+**Options**
 
 In addition to the [general options](#general-options) the following options are supported for this service:
 
 |Option      |Values                                            |Description                                  |
 |------------|--------------------------------------------------|---------------------------------------------|
 |sources     |`{index};{index}[;{index} ...]` or `all` (default)|Use location with given index as source.     |
 |destinations|`{index};{index}[;{index} ...]` or `all` (default)|Use location with given index as destination.|
+|annotations |`duration` (default), `distance`                  |Return additional table with distances to the response. Whether requested or not, the duration table is always returned.|
```
chaupow: `duration` (default), `distance` **or** `duration,distance` maybe?

docs/http.md
```diff
@@ -253,14 +254,23 @@ sources=0;5;7&destinations=5;1;4;2;3;6
...
-# Returns a asymmetric 3x2 matrix with from the polyline encoded locations `qikdcB}~dpXkkHz`:
+# Returns a asymmetric 3x2 duration matrix with from the polyline encoded locations `qikdcB}~dpXkkHz`:
 curl 'http://router.project-osrm.org/table/v1/driving/polyline(egs_Iq_aqAppHzbHulFzeMe`EuvKpnCglA)?sources=0;1;3&destinations=2;4'
+
+# Returns a 3x3 duration matrix:
+curl 'http://router.project-osrm.org/table/v1/driving/13.388860,52.517037;13.397634,52.529407;13.428555,52.523219&annotations=duration'
+
+# Returns a 3x3 duration matrix and a 3x3 distance matrix:
+curl 'http://router.project-osrm.org/table/v1/driving/13.388860,52.517037;13.397634,52.529407;13.428555,52.523219&annotations=distance'
+
```
chaupow: Ok wait. no. This is confusing. so

`annotation=distance` and `annotation=duration,distance` will  return the same response?

If yes, and this is intended we should make it clearer. this is confusing.
If yes, and this is not 100% thought through can we discuss?

chaupow:
I ran measurements on @danpat's improvement and compared them 🎉 
https://github.com/Project-OSRM/osrm-backend/pull/5041

| What        | `distances` PR           | `faster-distances` PR  |
| ------------- |:-------------| -----|
| 100x100 in `us-west`      | `12s` | `5.6s` |
| 100x100 in la max dist 20km      | `3.7s`      |  `800ms` |

docs/http.md
ghoshkaj: @TheMarex actually maybe you can clarify this. In the comment here: https://github.com/Project-OSRM/osrm-backend/pull/4990#issuecomment-380067808 I took this to mean that we should always return `duration` table whether it was asked for or not in the response. That's what I've implemented and what I've documented here.

include/engine/routing_algorithms/routing_base_ch.hpp
ghoshkaj: we made the decision to leave this in

test/nodejs/table.js
```diff
@@ -6,178 +6,187 @@ var three_test_coordinates = require('./constants').three_test_coordinates;
...
-                    // and finite (not nan, inf etc.)
-                    assert.ok(Number.isFinite(column[j]), 'distance is finite number');
+    test('table: ' + annotation + ' table in Monaco with sources/destinations', function(assert) {
+        assert.plan(7);
+        var osrm = new OSRM(data_path);
+        var options = {
+            coordinates: [three_test_coordinates[0], three_test_coordinates[1]],
+            sources: [0],
+            destinations: [0,1],
+            annotations: [annotation.slice(0,-1)]
```
TheMarex: The fact that our annotations parameters uses singular over plural is one of my biggest API regrets. :see_no_evil:

ghoshkaj: 😢

features/testbot/duration_matrix.feature
```diff
@@ -0,0 +1,492 @@
...
+        And the ways
+            | nodes |
+            | ab    |
+
+        When I request a travel time matrix I should get
+            |   | a  | b  |
+            | a | 0  | 10 |
+            | b | 10 | 0  |
+
+    @ch
```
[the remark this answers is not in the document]
ghoshkaj: These tests fail on the mld profile 😬 . Going to ticket out as a bug!

features/testbot/duration_matrix.feature
```diff
@@ -0,0 +1,492 @@
+@matrix @testbot
```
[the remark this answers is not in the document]
ghoshkaj: 😄

PR #4996, opened 2018-04-03
osrm-routed only on localhost during tests

danpat:
This PR changes the cucumber test suite to start `osrm-routed` only listening on `127.0.0.1` by default.  This achieves:

  1. Your work-in-progress `osrm-routed` with bugs won't be listening on an open public port by default.
  2. If you're like me and subject to a corporate firewall policy on your dev machine, you won't be plagued by annoying popups that ask you to approve the latest `osrm-routed` binary for incoming network connections on a public network interface.

## Tasklist

 - [ ] review
 - [ ] adjust for comments

### DIFF
--- a/features/lib/osrm_loader.js
+++ b/features/lib/osrm_loader.js
@@ -47,7 +47,7 @@ class OSRMBaseLoader{
           if (err) {
             if (retryCount < 10) {
               retryCount++;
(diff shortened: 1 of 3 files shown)


danpat:
Not sure if this is worth a CHANGELOG entry - this is a development environment only fix.

PR #5002, opened 2018-04-05
Split shared memory into updatable and non-updatable section

TheMarex:
# Issue

This is a simple version of #4873 that still works on per-file granularity. To make this more effective we need to split the weights from the MLD graph and refactor the way we update speeds (to work on a graph not an edge list).

## Tasklist

 - [ ] review
 - [ ] adjust for comments

## Requirements / Relations

Based on #4982

### DIFF
--- a/features/options/datastore/datastore.feature
+++ b/features/options/datastore/datastore.feature
@@ -1,0 +1,20 @@
+@datastore @options @help
+Feature: osrm-datastore command line options: list
+
+    Background:
+        Given the profile "testbot"
+        And the node map
+            """
+            a b
+            """
+        And the ways
+            | nodes |
+            | ab    |
+        And the data has been contracted
+
+    Scenario: osrm-datastore - Help should be shown when no options are passed
+        When I try to run "osrm-datastore --dataset-name test_dataset_42 {processed_file}"
+        Then it should exit successfully
+        When I try to run "osrm-datastore --list"
+        Then it should exit successfully
+        And stdout should contain "test_dataset_42/data"
--- a/include/engine/data_watchdog.hpp
+++ b/include/engine/data_watchdog.hpp
@@ -32,20 +32,28 @@ template <typename AlgorithmT, typename FacadeT> class DataWatchdogImpl;
 template <typename AlgorithmT>
 class DataWatchdogImpl<AlgorithmT, datafacade::ContiguousInternalMemoryDataFacade<AlgorithmT>> final
 {
-    using mutex_type = typename storage::SharedMonitor<storage::SharedDataTimestamp>::mutex_type;
+    using mutex_type = typename storage::SharedMonitor<storage::SharedRegionRegister>::mutex_type;
     using Facade = datafacade::ContiguousInternalMemoryDataFacade<AlgorithmT>;
 
   public:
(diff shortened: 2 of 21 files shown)


features/options/datastore/datastore.feature
```diff
@@ -17,4 +17,27 @@ Feature: osrm-datastore command line options: list
         Then it should exit successfully
         When I try to run "osrm-datastore --list"
         Then it should exit successfully
-        And stdout should contain "test_dataset_42/data"
+        And stdout should contain "test_dataset_42/static"
```
chaupow: Maybe we can add `And stdout should contain "test_dataset_42/updatable"` too

include/engine/data_watchdog.hpp
```diff
@@ -43,18 +43,24 @@ class DataWatchdogImpl<AlgorithmT, datafacade::ContiguousInternalMemoryDataFacad
             boost::interprocess::scoped_lock<mutex_type> current_region_lock(barrier.get_mutex());
 
             auto &shared_register = barrier.data();
-            auto region_id = shared_register.Find(dataset_name + "/data");
-            if (region_id == storage::SharedRegionRegister::INVALID_REGION_ID)
+            auto static_region_id = shared_register.Find(dataset_name + "/static");
```
chaupow: we could replace `auto` with `storage::SharedRegionRegister::RegionID` 🤷‍♀️ 
code style nitpick and up to you whether you want to pick it up or not 😂

TheMarex: I usually prefer the short version, since it makes changes to the code easier. For example, when renaming the type `storage::SharedRegionRegister::RegionID` it should not break the code in too many places.

I found that in most cases it is good practice to write your code type-safe but also type-agnostic, meaning the type information is used by the compiler to find errors, but when reading the code you don't really have to care about the types that much. I think the short rule was: "`auto` whenever possible".

chaupow: yea. i can live with both. they have their advantages and disadvantages 👍

emiltin: @TheMarex sounds like Swift :-)

src/storage/storage.cpp
```diff
@@ -263,169 +305,209 @@ void Storage::PopulateLayout(DataLayout &layout)
...
-        auto contracted_metric = make_contracted_metric_view(memory_ptr, layout, metric_prefix);
+        auto contracted_metric = make_contracted_metric_view(index, metric_prefix);
         std::unordered_map<std::string, contractor::ContractedMetricView> metrics = {
             {metric_name, std::move(contracted_metric)}};
 
         std::uint32_t graph_connectivity_checksum = 0;
         contractor::files::readGraph(
             config.GetPath(".osrm.hsgr"), metrics, graph_connectivity_checksum);
-
-        if (turns_connectivity_checksum != graph_connectivity_checksum)
```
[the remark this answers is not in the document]
TheMarex: Good point, I forgot about removing this. Thinking about this, there is probably a way to retain this. :+1:

TheMarex:
I investigated the coverage failures and it seems like the alternatives acting up. :-/

PR #5009, opened 2018-04-09
Simplify toEdges and make it more robust

TheMarex:
# Issue

We have experienced some possible crashes in this code in production when handling big graphs. It could be OOM so this is an attempt to make the allocations more predictable and add better logging.

## Tasklist

 - [x] review
 - [x] adjust for comments
 - [ ] cherry pick to `5.16`

### DIFF
--- a/include/contractor/graph_contractor_adaptors.hpp
+++ b/include/contractor/graph_contractor_adaptors.hpp
@@ -128,23 +128,24 @@ ContractorGraph toContractorGraph(NodeID number_of_nodes, InputEdgeContainer inp
 
 template <class Edge, typename GraphT> inline std::vector<Edge> toEdges(GraphT graph)
 {
-    std::vector<Edge> edges;
-    edges.reserve(graph.GetNumberOfEdges());
+    util::Log() << "Converting contracted graph with " << graph.GetNumberOfEdges()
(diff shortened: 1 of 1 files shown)


include/contractor/graph_contractor_adaptors.hpp
```diff
@@ -128,23 +128,25 @@ ContractorGraph toContractorGraph(NodeID number_of_nodes, InputEdgeContainer inp
...
-    std::vector<Edge> edges;
-    edges.reserve(graph.GetNumberOfEdges());
-
-    util::UnbufferedLog log;
-    log << "Getting edges of minimized graph ";
-    util::Percent p(log, graph.GetNumberOfNodes());
-    const NodeID number_of_nodes = graph.GetNumberOfNodes();
-    if (graph.GetNumberOfNodes())
+    util::Log() << "Converting contracted graph with " << graph.GetNumberOfEdges()
+                << " to edge list (" << (graph.GetNumberOfEdges() * sizeof(Edge) / 1024. / 1024.)
```
oxidase: imho, it is better to show the value in bytes to avoid MiB vs MB confusion
```
" edges to an edges list (allocating "  << graph.GetNumberOfEdges() * sizeof(Edge) << " bytes)";
```

PR #5012, opened 2018-04-10
Fix hard-coded hint strings in unit-tests

TheMarex:
# Issue

Currently we hard-code the hints string which is ugly an break for every change of `PhantomNode`. This doesn't hard code the string but dynamically constructs them.

## Tasklist

 - [x] review
 - [x] adjust for comments


danpat:
@ghoshkaj that weird output looks like a bug in the debug code responsible for printing coordinates:

https://github.com/Project-OSRM/osrm-backend/blob/ea0881553e4016c4b20af8ce3730cf415cc6156e/include/util/debug.hpp#L22-L27

Issue #5015, opened 2018-04-10
Improving the `cyclability` weight in bike profiles


chaupow:
We added the `cyclability` weight to OSRM's default profile for users that prefer safe routes over fast routes.

I would like to improve/work on this profile:

### Flatbush Avenue, NYC
* Flatbush Avenue in NYC is a 6 lane road and not preferred for biking. 
* `osrm-frontend` request: 
   ```
   http://localhost:9966/?z=16&center=40.687212%2C-73.979788&loc=40.690287%2C-73.981826&loc=40.683551%2C-73.978908&hl=en&alt=0
   ```
* current behaviour
   <img width="448" alt="screenshot 2018-04-10 18 12 08" src="https://user-images.githubusercontent.com/804027/38569098-b5c67322-3cea-11e8-80d6-cd2b4b84926a.png">

* desired behaviour is something similar to this
   <img width="414" alt="screenshot 2018-04-10 18 10 39" src="https://user-images.githubusercontent.com/804027/38569105-b9892f72-3cea-11e8-9422-dd5f1cab22d9.png">

### Hamilton Avenue, NYC
* I am pretty sure cycling on Hamilton Avenue is not safe at all
* `osrm-frontend` request:
   ```
   http://localhost:9966/?z=16&center=40.668351%2C-73.995452&loc=40.664526%2C-73.995495&loc=40.670784%2C-73.996933&hl=en&alt=0
   ```
* current behaviour
   <img width="385" alt="screenshot 2018-04-10 18 12 59" src="https://user-images.githubusercontent.com/804027/38569143-d86ff83a-3cea-11e8-967e-b57930d06e95.png">

* desired behaviour is something similar to this   
   <img width="454" alt="screenshot 2018-04-10 18 13 27" src="https://user-images.githubusercontent.com/804027/38569185-ee33afea-3cea-11e8-9f09-394510beff1d.png">

PR #5016, opened 2018-04-10
Revamp `cyclability` weights of the bike profile

chaupow:
# Issue

This PR serves as a playground to discuss the `cyclability` weights in the bike profile.
The related issue is https://github.com/Project-OSRM/osrm-backend/issues/5015

I am aware that this is highly discussionable because there is no clear `right` or `wrong` when it comes to profile preferences. It's possible there will be different opinions, if it does not work out, I can make a customized profile but I think the default `bike` profile will benefit from an overhaul

## Tasklist

 - [x] review
 - [x] adjust for comments

### DIFF
--- a/profiles/bicycle.lua
+++ b/profiles/bicycle.lua
@@ -250,204 +250,237 @@ end
 
 function handle_bicycle_tags(profile,way,result,data)
     -- initial routability check, filters out buildings, boundaries, etc
-  local route = way:get_value_by_key("route")
-  local man_made = way:get_value_by_key("man_made")
-  local railway = way:get_value_by_key("railway")
-  local amenity = way:get_value_by_key("amenity")
-  local public_transport = way:get_value_by_key("public_transport")
-  local bridge = way:get_value_by_key("bridge")
+  data.route = way:get_value_by_key("route")
+  data.man_made = way:get_value_by_key("man_made")
+  data.railway = way:get_value_by_key("railway")
+  data.amenity = way:get_value_by_key("amenity")
+  data.public_transport = way:get_value_by_key("public_transport")
+  data.bridge = way:get_value_by_key("bridge")
 
   if (not data.highway or data.highway == '') and
-  (not route or route == '') and
-  (not profile.use_public_transport or not railway or railway=='') and
-  (not amenity or amenity=='') and
-  (not man_made or man_made=='') and
-  (not public_transport or public_transport=='') and
-  (not bridge or bridge=='')
+  (not data.route or data.route == '') and
+  (not profile.use_public_transport or not data.railway or data.railway=='') and
+  (not data.amenity or data.amenity=='') and
+  (not data.man_made or data.man_made=='') and
+  (not data.public_transport or data.public_transport=='') and
+  (not data.bridge or data.bridge=='')
   then
     return false
   end
 
   -- access
-  local access = find_access_tag(way, profile.access_tags_hierarchy)
-  if access and profile.access_tag_blacklist[access] then
+  data.access = find_access_tag(way, profile.access_tags_hierarchy)
+  if data.access and profile.access_tag_blacklist[data.access] then
     return false
   end
 
   -- other tags
-  local junction = way:get_value_by_key("junction")
-  local maxspeed = parse_maxspeed(way:get_value_by_key ( "maxspeed") )
-  local maxspeed_forward = parse_maxspeed(way:get_value_by_key( "maxspeed:forward"))
-  local maxspeed_backward = parse_maxspeed(way:get_value_by_key( "maxspeed:backward"))
-  local barrier = way:get_value_by_key("barrier")
-  local oneway = way:get_value_by_key("oneway")
-  local oneway_bicycle = way:get_value_by_key("oneway:bicycle")
-  local cycleway = way:get_value_by_key("cycleway")
-  local cycleway_left = way:get_value_by_key("cycleway:left")
-  local cycleway_right = way:get_value_by_key("cycleway:right")
-  local duration = way:get_value_by_key("duration")
-  local service = way:get_value_by_key("service")
-  local foot = way:get_value_by_key("foot")
-  local foot_forward = way:get_value_by_key("foot:forward")
-  local foot_backward = way:get_value_by_key("foot:backward")
-  local bicycle = way:get_value_by_key("bicycle")
+  data.junction = way:get_value_by_key("junction")
+  data.maxspeed = parse_maxspeed(way:get_value_by_key ( "maxspeed") )
+  data.maxspeed_forward = parse_maxspeed(way:get_value_by_key( "maxspeed:forward"))
+  data.maxspeed_backward = parse_maxspeed(way:get_value_by_key( "maxspeed:backward"))
+  data.barrier = way:get_value_by_key("barrier")
+  data.oneway = way:get_value_by_key("oneway")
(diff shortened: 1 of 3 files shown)


chaupow:
## Tuning ideas

I started changing the profile because I think, that the `cyclability` weight should
- penalize `primary` roads more
- penalize `secondary` roads more
- penalize `primary` and `secondary` roads that are oneways more, because that means that these roads are very big and have separated lanes (and thus are even more dangerous for bikes)


## Cyclerating Tags

There are some `cyclerating` tags in the US and Australia:
https://taginfo.openstreetmap.org/keys/cyclerating#values

The ratings go from `1-5` and `5` seems to be the most dangerous, `1` most likeable roads. _There is no OSM wiki page for this tag._

I wonder if we should support this. @1ec5 do you have an opinion on this? ^

## Evaluation

Other than that I will add some screenshots to show the differences this PR introduces.

1ec5:
I’ve never heard of the `cyclerating` tag before. It seems to be [primarily confined](https://taginfo.openstreetmap.org/keys/cyclerating#map) to a few cities in the U.S. and Australia.

systemed:
Please don't support `cyclerating` or any similar tag - such tags are subjective (not [verifiable](https://wiki.openstreetmap.org/wiki/Verifiability)) and don't belong in OSM. Supporting it in a major client like OSRM would encourage people to add inappropriate data.

chaupow:
@systemed @1ec5 thanks for your input! That affirms my approach by using penalties on primary and secondary 😄

emiltin:
In places like Copenhagen with well developed bicycle infrastructure, almost all primary and secondary roads have bicycle tracks with raised curbs and are not considered unsafe. Penalizing them should only happen if there is no bicycle track or lane. Otherwise routes will be considered weird and not very useful.

A track is better than a lane, which is better than nothing.
Consider looking at maxspeed of the cars instead of oneway. There are many oneways that are not very fast for cars.

emiltin:
I think the first step should be to refactor the existing code into more manageble handlers. I worked at this a good while ago, but maybe some of it can still be helpful. specifically it uses the foot profile to check if it's possible to push the bike. the current profile includes code to do this directly, but it's not very robust.

pr:
https://github.com/Project-OSRM/osrm-backend/pull/4314

bicycle.lua:
https://github.com/Project-OSRM/osrm-backend/blob/019992c3c5d17d2c4c916bc89cc513cc06c24257/profiles/bicycle.lua#L351


Othewise, an improvement I would like is the for bicycle profile to parse route relations to e.g. prioritize bicycle routes.

chaupow:
> Penalizing them should only happen if there is no bicycle track or lane. Otherwise routes will be considered weird and not very useful.

I agree! In my PR 
* I changed the `safety_penalty`, making the penalty more aggressive
* which is only applied when the road is considered `is_unsafe ` **and not** `has_cycleway` 

unless there is a bug I think your desired behaviour would be implemented

> There are many oneways that are not very fast for cars.
True, but I wonder if there are many one way `primary` roads that are not fast for cars 🤔 I can look more into the `car.lua` profile, but I think I remember that the `highway` tag is one of the main sources to deduce the speed from.

 > I think the first step should be to refactor the existing code into more manageble handlers. 
Good point 😪  Might safe time in the future. I will look into your PR and see what I can learn & gain :)

chaupow:
Alrighty mighty: 

- ✅ only change behaviour if set to `cyclability`
- ✅ no usage of `cyclerating`
- ✅ adjust penalties for primaries and secondaries and its oneways
- ✅ adjust primaries and secondaries and its oneways
- ✅ don't add penalties of there are cycleways

> I think the first step should be to refactor the existing code into more manageble handlers.

@emiltin I cleaned up the functions a bit, shuffled them around and tried to isolate them in functions. This is not a complete and nice refactor but it is a first step and might make the process easier for the next person who will take on the glorious task of refactoring the bike profile 🙄

chaupow:
I browsed through issues where users reported wrong bike routes and some bike routes that I knew where difficult and sanity checked

- whether the changes improved the routes and
- whether the changes keep the correct behaviour and not mess up routes. 

Some screenshots:

screenshot 2018-04-12 13 37 42
screenshot 2018-04-12 13 37 50
screenshot 2018-04-12 13 37 55
screenshot 2018-04-12 13 38 02
screenshot 2018-04-12 13 38 10
screenshot 2018-04-12 13 38 19
screenshot 2018-04-12 13 38 31
screenshot 2018-04-12 13 38 36

chaupow:
I also noticed that it would be nice to start a collection of reference routes for certain profiles ^

I am just guessing but other people also probably have their own list of reference routes they check when they work on their profiles?

emiltin:
if the "cycleability" metric is really about safety, then maybe it would be clear to call if "safest". The default metric already supposedly is a "cycleable" route?

chaupow:
> if the "cycleability" metric is really about safety, then maybe it would be clear to call if "safest". The default metric already supposedly is a "cycleable" route?

That's is actually a good point. I am trying to find the context of introducing the `cyclability` parameter to understand why we introduced this but I am not able to track down the PR where we added this. there is a comment in `bicycle.lua` that made me think it is mostly about safety but I might be wrong

```
    -- reduce the driving speed by 30% for unsafe roads
    -- only used for cyclability metric
```

I can rename `cyclability` to `bikesafety` or `safety` if nobody objects.

TheMarex:
> I can rename cyclability to bikesafety or safety if nobody objects.

Renaming metrics is out-of-scope for this PR, since that is unrelated to the issue at hand. It could also have unexpected consequences if users are checking for this string.

MichalPP:
btw, profile I use is at https://github.com/FreemapSlovakia/freemap-routing/blob/master/oma-bicycle.lua

I check also for:
* number of lanes (more lanes -> less safe)
* whether buses pass that road -> less safe
* wide segregated footways/paths/pedestrian -> safe and cyclable
* (and elevation)

chaupow:
Awesome thanks @MichalPP 
I just finished route QA on this PR so I am reluctant to do more changes on this PR.

I broke it out to another PR [`Improving the `cyclability` weight in bike profiles II`](https://github.com/Project-OSRM/osrm-backend/issues/5029)

Can you point me to the part in the code that regards bus lanes less safe?

profiles/bicycle.lua
```diff
@@ -373,10 +373,10 @@ function handle_bicycle_tags(profile,way,result,data)
   -- cycleway
   local has_cycleway_forward = false
   local has_cycleway_backward = false
-  local is_oneway = result.forward_mode ~= mode.inaccessible and result.backward_mode ~= mode.inaccessible and not implied_oneway
+  local is_twoway = result.forward_mode ~= mode.inaccessible and result.backward_mode ~= mode.inaccessible and not implied_oneway
```
TheMarex: Good catch!

profiles/bicycle.lua
```diff
@@ -250,204 +250,237 @@ end
...
-  local duration = way:get_value_by_key("duration")
-  local service = way:get_value_by_key("service")
-  local foot = way:get_value_by_key("foot")
-  local foot_forward = way:get_value_by_key("foot:forward")
-  local foot_backward = way:get_value_by_key("foot:backward")
-  local bicycle = way:get_value_by_key("bicycle")
-
-
-  local way_type_allows_pushing = false
+  data.junction = way:get_value_by_key("junction")
```
TheMarex: Oh nice, consistency with `car`! :100:

PR #5021, opened 2018-04-12
Avoid using signed integers for edge IDs

oxidase:
# Issue

PR fixes an integer overflow if number of edges is greater than 2^31 and adds a check for an unsigned integer overflow.

## Tasklist

 - [x] review
 - [ ] adjust for comments

## Requirements / Relations

### DIFF
--- a/include/util/dynamic_graph.hpp
+++ b/include/util/dynamic_graph.hpp
@@ -2,6 +2,8 @@
 #define DYNAMICGRAPH_HPP
 
 #include "util/deallocating_vector.hpp"
+#include "util/exception.hpp"
+#include "util/exception_utils.hpp"
 #include "util/integer_range.hpp"
 #include "util/permutation.hpp"
 #include "util/typedefs.hpp"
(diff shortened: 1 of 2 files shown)


include/util/dynamic_graph.hpp
```diff
@@ -411,14 +413,19 @@ template <typename EdgeDataT> class DynamicGraph
...
-        auto new_edge_index = 0;
+        EdgeID new_edge_index = 0;
         std::vector<EdgeID> old_to_new_edge(edge_list.size(), SPECIAL_EDGEID);
         for (auto node : util::irange<NodeID>(0, number_of_nodes))
         {
             auto new_first_edge = new_edge_index;
             // move all filled edges
             for (auto edge : GetAdjacentEdgeRange(node))
             {
+                if (new_edge_index == std::numeric_limits<EdgeID>::max())
```
danpat: Can this check be done before we start the loop and fail earlier?  We know the size of the edge_list, right?  This renumbering shouldn't create more than there was previously....

oxidase: yes, the check only prevents overflow of `number_of_valid_edges`, but you are right `old_to_new_edge` may be corrupted for dummy edges if `new_edge_index` overflows below

PR #5023, opened 2018-04-12
Remove dummy edges before inplace permutation

oxidase:
# Issue

https://github.com/Project-OSRM/osrm-backend/pull/5021#discussion_r181005061 points that the check `edge_list.size() < std::numeric_limits<EdgeID>::max()` can be placed before the loop, but this check limits number of valid edges by `2^32 - #dummy_edges`. 

PR implements a removal of dummy edges before a `util::inplacePermutation` call:
:heavy_check_mark: number of edges is limited by `2^32`

❌  an additional iteration over the list of edges, but potentially linearized removal can be faster due to caching

/cc @danpat 

## Tasklist

 - [x] review
 - [x] adjust for comments

## Requirements / Relations

### DIFF
--- a/include/util/dynamic_graph.hpp
+++ b/include/util/dynamic_graph.hpp
@@ -413,6 +413,11 @@ template <typename EdgeDataT> class DynamicGraph
         util::inplacePermutation(node_array.begin(), node_array.end(), old_to_new_node);
 
         // Build up edge permutation
+        if (edge_list.size() >= std::numeric_limits<EdgeID>::max())
+        {
+            throw util::exception("There are too many edges, OSRM only supports 2^32" + SOURCE_REF);
+        }
+
         EdgeID new_edge_index = 0;
         std::vector<EdgeID> old_to_new_edge(edge_list.size(), SPECIAL_EDGEID);
         for (auto node : util::irange<NodeID>(0, number_of_nodes))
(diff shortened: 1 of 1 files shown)


TheMarex:
The reason why we don't remove the dummy edges is that it impacts performance significantly during contraction, since we need the over-allocation for the DynamicGraph to be effective. For example on Bayern the contraction on `master` needs `504 sec` but with this branch `549 sec`. If we need to limit the amount of over-allocation we should add a fixed ratio of 30% over-allocation.

oxidase:
@TheMarex thank you  for running performance benchmarks! I think it is enough for now to keep a sanity check to avoid overflows of permutation indexes. 

I don't think we need to limit over-allocation but a check before
https://github.com/Project-OSRM/osrm-backend/blob/4c65210d33a09aa718e123b1e59aa7300ea7b943/src/contractor/graph_contractor.cpp#L737 
that will stop contraction like 
```
                if (edge_list.size() >= std::numeric_limits<EdgeID>::max())
                {
                    util::Log(logWARNING) << "Too many sortcuts, exiting graph contraction";
                    node_data.Renumber(new_to_old_node_id);
                    RenumberGraph(graph, new_to_old_node_id);
                    return std::move(node_data.is_core);
                }
```

Issue #5024, opened 2018-04-12
Exception thrown when osrm-datastore --list called with no shared memory regions exists


karenzshea:
In 5.17 `osrm-datastore` has a new option `--list` that will display shared memory regions and metadata about them. However when run when no shared memory regions have been allocated, we throw an exception.

```
/osrm-datastore --list=1
terminate called after throwing an instance of 'osrm::util::exception'
  what():  No shared memory block 'osrm-region' found, have you forgotten to run osrm-datastore?include/storage/shared_monitor.hpp:83
```

This should be handled more nicely - running list without shmem regions allocated is a reasonable use case.

PR #5025, opened 2018-04-12
Gracefully handle no shared memory regions found

karenzshea:
# Issue

https://github.com/Project-OSRM/osrm-backend/issues/5024

I'm going to forgo adding a test for now - adding a test here would require first clearing all shared memory regions in the environment where osrm-datastore is being run, which is a bit dangerous. Maybe we can consider setting up a very robust isolated testing environment someday that is convenient to run for osrm-datastore.

## Tasklist

 - [x] review
 - [x] adjust for comments

### DIFF
--- a/src/tools/store.cpp
+++ b/src/tools/store.cpp
@@ -27,9 +27,9 @@ void deleteRegion(const storage::SharedRegionRegister::ShmKey key)
 
 void listRegions()
 {
+    osrm::util::Log() << "name\tshm key\ttimestamp\tsize";
     if (!storage::SharedMonitor<storage::SharedRegionRegister>::exists())
     {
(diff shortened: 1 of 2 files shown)


src/tools/store.cpp
```diff
@@ -27,7 +27,11 @@ void deleteRegion(const storage::SharedRegionRegister::ShmKey key)
 
 void listRegions()
 {
-
+    if (!storage::SharedMonitor<storage::SharedRegionRegister>::exists())
+    {
+        osrm::util::Log() << "No shared memory regions found. Try running osrm-datastore";
```
TheMarex: What about just printing an empty list in this case? That would make the output more predictable.

PR #5026, opened 2018-04-12
Add dev dependencies for osrm runner

chaupow:
# Issue

osrm-runner needs some packages as a dependencies which are not included in our `package.json`. adding it now.

## Tasklist

 - [ ] CHANGELOG.md entry ([How to write a changelog entry](http://keepachangelog.com/en/1.0.0/#how))
 - [ ] update relevant [Wiki pages](https://github.com/Project-OSRM/osrm-backend/wiki)
 - [ ] add tests (see [testing documentation](https://github.com/Project-OSRM/osrm-backend/blob/master/docs/testing.md)
 - [ ] review
 - [ ] adjust for comments
 - [ ] cherry pick to release branch

## Requirements / Relations

 Link any requirements here. Other pull requests this PR is based on?

PR #5030, opened 2018-04-16
Fix CSV encoding and help in OSRM runner

TheMarex:
# Issue

When saving JSON data in CSV we did not properly escape it. This replaces custom logic with a module to do it properly.

## Tasklist

 - [x] review
 - [x] adjust for comments

## Requirements / Relations

This PR is based on #5026.

### DIFF
--- a/scripts/osrm-runner.js
+++ b/scripts/osrm-runner.js
@@ -10,6 +10,7 @@ const clu = require('command-line-usage');
 const ansi = require('ansi-escape-sequences');
 const turf = require('turf');
 const jp = require('jsonpath');
+const csv_stringify = require('csv-stringify/lib/sync');
 
 const run_query = (query_options, filters, callback) => {
     let tic = () => 0.;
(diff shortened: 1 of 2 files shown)


scripts/osrm-runner.js
```diff
@@ -123,11 +118,8 @@ queries = queries.map(q => { return {hostname: options.server.hostname, port: op
...
     run_query(query, options.filter, (query, code, ttfb, total, results) => {
-        let str = `"${query}",${code}`;
-        if (ttfb !== undefined) str += `,${ttfb}`;
-        if (total !== undefined) str += `,${total}`;
-        if (typeof results === 'object' && results.length > 0)
-            str += ',' + results.map(x => isNaN(x) ? '"' + JSON.stringify(x).replace(/\n/g, ';').replace(/"/g, "'") + '"' : Number(x)).join(',');
-        console.log(str);
+        let data = results ? JSON.stringify(results[0]) : '';
+        let record = [[query, code, ttfb, total, data]];
+        process.stdout.write(csv_stringify(record));
```
chaupow: looks good and clean!
this changes the behaviour a bit:
if one of the records doesnt exist (e.g. `query`, `code`)
    - the new code will emit a `,` and an empty string
    - the old code does not emit a `,`

i think the new code is better, but just wanted to make sure we are aware that the behaviour changes.

the old code is especially dangerous if `total` would exist but `ttfb` doesn't. then the commas and the positions are off. so new behaviour is better! 👍

Issue #5039, opened 2018-04-19
`invalid` turn type returned


karenzshea:
A request in New Zealand (extract can be fetched from geofabrik), I see a turn type that shouldn't be exposed:

```
$ curl 'localhost:5000/route/v1/car/171.21113,-42.450158;171.12712383270264,-42.58207906066402?steps=true' | jq '.routes[].legs[].steps[].maneuver.type'
```
Returns,
```
"depart"
"exit roundabout"
"invalid"
"exit rotary"
"arrive"
```

In debug mode, [this assertion is triggered](https://github.com/Project-OSRM/osrm-backend/blob/b2aeb476303db43827487d6a9374c496bf203f4d/include/engine/guidance/route_step.hpp#L1580) which assumes that for a step to be collapsed into the one before it, the geometry indices of the two steps should be adjacent.

I managed to narrow down a possible cause for this in the segregated intersection collapsing, so that when I comment out the segregated intersection collapsing in `route_api` and rerun the same request, I see these turn types emitted
```
"depart"
"exit roundabout"
"new name"
"new name"
"new name"
"rotary"
"exit rotary"
"arrive"
```

Inspecting this route though, I'm not sure what should be considered a segregated intersection,
https://gist.github.com/karenzshea/8ff491dbe7155fc8a94cb7701ab2aa89

### Next steps
- [x] Determine which parts of the code are incorrectly collapsing steps
- [x] Is an intersection getting misclassified as segregated?
- [x] Write a test case
- [x] Fix either `invalid` turn type emission, or segregated intersection misclassification

cc @dgearhart

PR #5040, opened 2018-04-19
Do not combine a segregated edge with a roundabout

dgearhart:
In theory, this should fix https://github.com/Project-OSRM/osrm-backend/issues/5039, where we emit an `invalid` instruction if you cross a segregated intersection edge as you enter a roundabout.

However, the test case is proving difficult to recreate.  There's a cucumber test here that in theory models the intersection in #5039, but doesn't seem to be failing.

@karenzshea can you 👁 and see if you can spot why the test case doesn't expose the bug if the fix is removed?

## Update 
@danpat @karenzshea I updated the test to pass properly when the update to `collapse_turns.cpp` was included. Also, the test failed when the update to `collapse_turns.cpp` was not included. Could some please verify the same?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,8 @@
+# 5.17.1
+  - Changes from 5.17.0:
+    - Bugfixes:
+      - FIXED: Do not combine a segregated edge with a roundabout [#5039](https://github.com/Project-OSRM/osrm-backend/issues/5039)
+
 # 5.17.0
   - Changes from 5.16.0:
     - Bugfixes:
--- a/features/testbot/fixed.feature
+++ b/features/testbot/fixed.feature
@@ -43,11 +43,8 @@ Feature: Fixed bugs, kept to check for regressions
             | 1    | 2  | bcd,bcd |
 
     #############################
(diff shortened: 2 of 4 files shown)


oxidase:
@dgearhart the test case looks good and covers a collapsing of the third instruction `{type = EnterRotary (12), direction_modifier = SlightLeft (5)}`  from https://gist.github.com/oxidase/848a9c9bf4f712727c8298e2f00ff5a3 to https://gist.github.com/oxidase/492079460744e005027f13ada94a6cbd

To reproduce the original issue with an `invalid` instruction the test case must include another  roundabout. 

Here is an example route that covers  the original issue and  fails with an assertion 
```
[assert][140301146519296] /home/miha/mapbox/osrm-backend/include/engine/guidance/route_step.hpp:158
in: osrm::engine::guidance::RouteStep& osrm::engine::guidance::RouteStep::ElongateBy(const osrm::engine::guidance::RouteStep&): geometry_end == following_step.geometry_begin + 1
```
without the fix and works correctly with the fix
screenshot from 2018-04-20 11-49-50

PR #5041, opened 2018-04-19
 Use faster method for calculating distances.

danpat:
Our use of the `haversineDistance` function is a bit trigonometry heavy.  Trig functions are relatively expensive, especially when used a lot.

With work on the `distances` branch, we'll be performing a lot of distance calculations within matrix requests.  A 100x100 table with distances without caching will spend over 50% of it's time in `haversineDistnace()`.

This PR switches from using the `haversineDistance` function, to a new `fccApproximateDistance` function, supplied by the https://github.com/mapbox/cheap-ruler-cpp library.

Don't let the name fool you - the `fccApproximateDistance` method is actually *more* precise than Haversine, and significantly faster.

In simple tests, this change reduced a 100x100 table request with `annotations=distances` from 8900ms down to about 4500ms.

Before the change, this was the CPU profile in a table request with distances:
screen shot 2018-04-19 at 4 26 11 pm

after this change:
screen shot 2018-04-19 at 4 21 58 pm

This PR is not ready to merge.  Still to do:

  1. Investigate inlining for `fccApproximateDistance` - some quick test indicated that it could get the 100x100 test request down to about 4050ms.
  2. See if we can achieve the same `cheap_ruler_cache` bucket lookup without the use of `lround()`, which is using 2.5% of the total time in `calculateEBGNAnnotations`.  If we have coords in integer format, simple digit truncation and multiplication should do the job.
  3. Check that we have enough buckets for sufficient ruler precision.  I pulled 10-degree buckets out of the air, but I don't know how much error that will introduce.
  4. Deprecate `haversineDistance` everywhere - there is no need to use it anywhere `fccApproximateDistance` is faster and more accurate.  `greatCircleDistance` may also be removable for the same reason.
  5. Verify test case distances.  I suspect some might change by a tiny fraction.  I haven't looked at any yet.

## Tasklist

 - [ ] review
 - [ ] adjust for comments


src/util/coordinate_calculation.cpp
```diff
@@ -32,6 +58,19 @@ std::uint64_t squaredEuclideanDistance(const Coordinate lhs, const Coordinate rh
...
+// https://www.gpo.gov/fdsys/pkg/CFR-2005-title47-vol4/pdf/CFR-2005-title47-vol4-sec73-208.pdf
+// should be within 0.1% or so of Vincenty method (assuming 19 buckets are enough)
+// Should be more faster and more precise than Haversine
+double fccApproximateDistance(const Coordinate coordinate_1, const Coordinate coordinate_2)
+{
+    const auto lon1 = static_cast<double>(util::toFloating(coordinate_1.lon));
+    const auto lat1 = static_cast<double>(util::toFloating(coordinate_1.lat));
+    const auto lon2 = static_cast<double>(util::toFloating(coordinate_2.lon));
+    const auto lat2 = static_cast<double>(util::toFloating(coordinate_2.lat));
+    return cheap_ruler_cache[std::lround(lat1 / 10) + 9].distance({lon1, lat1}, {lon2, lat2});
```
danpat: Should probably add assertion here that bucket number is within range for our cache.

chaupow:
awesome @danpat 
let me rerun the measurements with this!

emiltin:
sweet

chaupow:
Aaaaaand numbers are out!

| What        | `distances` PR           | `faster-distances` PR  |
| ------------- |:-------------| -----|
| 100x100 in `us-west`      | `12s` | `5.6s` |
| 100x100 in la max dist 20km      | `3.7s`      |  `800ms` |

emiltin:
i like what i see

chaupow:
@danpat if you're ok with it,  I will take it on from here working on your tasklist? 😄

mourner:
Wow, that's a massive speedup! Happy how this is turning out :)

danpat:
@cffk Are you implying a problem with the method itself, or a bug in the implementation?

cffk:
@danpat The problem is the latitude binning.  The distance between two points on latitude 55&deg; uses the latitude scale factor for 60&deg;.  The main term in the latitude scale factor is `cos(latitude)` and `cos(60&deg;)` differs from `cos(55&deg;)` by about 10%.

mourner:
@danpat yeah, rounding difference of 5 degrees is definitely too much. I'd suggest to 1) calculate the cheap ruler cache lazily; 2) bin much more granularly, e.g. 0.01 deg.

chaupow:
@mourner @cffk  

yup, I'm working on this. I think @danpat just used few buckets here to see whether it works at all and see the speed gain. 

I am adding more buckets (currently at 180, with `.5` diff of latitude but still doing Q&A) and the results seem nice. Thanks for flagging this 😄

cffk:
FYI, I submitted a pull request to cheap-ruler-cpp to fix the longitude wrapping bug.

cffk:
A couple of additional comments:

* I submitted another [pull request](https://github.com/mapbox/cheap-ruler-cpp/pull/13) for cheap-ruler-cpp to improve the accuracy and to let you switch to the WGS84 ellipsoid (from Clarke 1866).
* To minimize the errors for a given number of buckets, you would distribute the buckets according to `cos(lat)`.  Since this would slow down selecting the bucket, I don't think there's any way to take advantage of this.

oxidase:
@chaupow `std::experimental::optional` is not supported in MSVC2015. I changed it to `mapbox::optional` in `feature.hpp` and pushed a commit to make a CI check.

src/util/coordinate_calculation.cpp
```diff
@@ -27,26 +27,34 @@ class CheapRulerContainer
...
-        return cheap_ruler_cache[std::min((int)std::floor(abs(lat) / step),
-                                          (int)cheap_ruler_cache.size() - 1)];
+        auto lat = (lat_1 + lat_2) / util::FixedLatitude{2};
+        return getRuler(lat);
+    }
+
+    mapbox::cheap_ruler::CheapRuler &getRuler(const FixedLatitude lat)
+    {
+        BOOST_ASSERT(step > 1);
+        auto bin = (std::abs(static_cast<int>(lat)) - 1) / step;
```
danpat: What if the request is for `lat = 0` here?  This'll lead to `bin = -1`.

features/testbot/distance_matrix.feature
```diff
@@ -17,9 +17,9 @@ Feature: Basic Distance Matrix
             | ab    |
 
         When I request a travel distance matrix I should get
-            |   | a   | b   |
-            | a | 0   | 100 |
-            | b | 100 | 0   |
+            |   | a      | b      |
+            | a | 0      | 100+-1 |
```
emiltin: Elsewhere (like in the "When I route I should get " tests) the syntax used is "100m +-3", rather than "100+-3". For consistency it might be good to follow this syntax.

src/util/coordinate_calculation.cpp
TheMarex: Good catch! Didn't realize rounding direction was not actually defined for `1/2`.

chaupow:
Awesome, thanks a ton @oxidase !

PR #5048, opened 2018-04-23
Tune alternatives parameters for MLD

TheMarex:
# Issue

This PR fixes some bugs in the MLD code that reduced the quality fo alternatives and uses different parameters based on route distance.

### Number of alternatives on `master`

image

image

### Number of alternatives on this branch

image

image

## Tasklist

 - [x] measurements for different route lengths
 - [x] review
 - [x] adjust for comments

### DIFF
--- a/src/engine/routing_algorithms/alternative_path_mld.cpp
+++ b/src/engine/routing_algorithms/alternative_path_mld.cpp
@@ -47,13 +47,13 @@ struct Parameters
     double kAtMostLongerBy = 0.25;
     // Alternative paths similarity requirement (sharing).
     // At least 15% different than the shortest path.
-    double kAtLeastDifferentBy = 0.85;
+    double kAtMostSameBy = 0.85;
     // Alternative paths are still reasonable around the via node candidate (local optimality).
     // At least optimal around 10% sub-paths around the via node candidate.
     double kAtLeastOptimalAroundViaBy = 0.1;
     // Alternative paths similarity requirement (sharing) based on calles.
     // At least 15% different than the shortest path.
-    double kCellsAtLeastDifferentBy = 0.85;
+    double kCellsAtMostSameBy = 0.85;
 };
 
 // Represents a via middle node where forward (from s) and backward (from t)
@@ -140,27 +140,30 @@ Parameters parametersFromRequest(const PhantomNodes &phantom_node_pair)
     if (distance < 10000.)
     {
         parameters.kAlternativesToUnpackFactor = 10.0;
-        parameters.kCellsAtLeastDifferentBy = 1.0;
+        parameters.kCellsAtMostSameBy = 1.0;
         parameters.kAtLeastOptimalAroundViaBy = 0.2;
+        parameters.kAtMostSameBy = 0.50;
     }
     // 20km
     else if (distance < 20000.)
     {
         parameters.kAlternativesToUnpackFactor = 8.0;
-        parameters.kCellsAtLeastDifferentBy = 1.0;
+        parameters.kCellsAtMostSameBy = 1.0;
         parameters.kAtLeastOptimalAroundViaBy = 0.2;
+        parameters.kAtMostSameBy = 0.60;
     }
     // 50km
     else if (distance < 50000.)
     {
         parameters.kAlternativesToUnpackFactor = 6.0;
-        parameters.kCellsAtLeastDifferentBy = 0.95;
+        parameters.kCellsAtMostSameBy = 0.95;
+        parameters.kAtMostSameBy = 0.70;
     }
     // 100km
     else if (distance < 100000.)
     {
         parameters.kAlternativesToUnpackFactor = 4.0;
-        parameters.kCellsAtLeastDifferentBy = 0.75;
+        parameters.kCellsAtMostSameBy = 0.75;
     }
 
     return parameters;
@@ -256,7 +259,7 @@ RandIt filterPackedPathsByCellSharing(RandIt first,
                                       const Parameters &parameters)
 {
     // In this case we don't need to calculate sharing, because it would not filter anything
-    if (parameters.kCellsAtLeastDifferentBy >= 1.0)
+    if (parameters.kCellsAtMostSameBy >= 1.0)
         return last;
 
     util::static_assert_iter_category<RandIt, std::random_access_iterator_tag>();
@@ -311,7 +314,7 @@ RandIt filterPackedPathsByCellSharing(RandIt first,
 
         const auto sharing = 1. - difference;
 
-        if (sharing > parameters.kCellsAtLeastDifferentBy)
+        if (sharing > parameters.kCellsAtMostSameBy)
         {
             return true;
         }
@@ -491,7 +494,7 @@ RandIt filterUnpackedPathsBySharing(RandIt first,
         BOOST_ASSERT(sharing >= 0.);
         BOOST_ASSERT(sharing <= 1.);
 
-        if (sharing > parameters.kAtLeastDifferentBy)
+        if (sharing > parameters.kAtMostSameBy)
         {
             return true;
         }
--- a/src/engine/routing_algorithms/alternative_path_mld.cpp
+++ b/src/engine/routing_algorithms/alternative_path_mld.cpp
@@ -473,21 +473,21 @@ RandIt filterUnpackedPathsBySharing(RandIt first,
             return false;
         }
 
-        EdgeWeight total_weight = 0;
-        const auto add_if_seen = [&](const EdgeWeight weight, const EdgeID edge) {
-            auto edge_weight = facade.GetEdgeData(edge).weight;
-            total_weight += edge_weight;
+        EdgeWeight total_duration = 0;
+        const auto add_if_seen = [&](const EdgeWeight duration, const EdgeID edge) {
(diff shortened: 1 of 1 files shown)


src/engine/routing_algorithms/alternative_path_mld.cpp
```diff
@@ -76,6 +81,94 @@ struct WeightedViaNodeUnpackedPath
...
+    }
+    else if (duration > EdgeWeight(10 * 60 * 60))
+    {
+        return 0.10;
+    }
+
+    // Bigger than 10 minutes but smaller than 10 hours
+    BOOST_ASSERT(duration > 5 * 60 && duration < 10 * 60 * 60);
+
+    return a + b / (duration - d) + c / std::pow(duration - d, 3);
```
daniel-j-h: Ah, so we special case "short" and "long" routes and in between we simply scale the factor?

src/engine/routing_algorithms/alternative_path_mld.cpp
```diff
@@ -76,6 +81,94 @@ struct WeightedViaNodeUnpackedPath
...
+    {
+        return 1.0;
+    }
+    else if (duration > EdgeWeight(10 * 60 * 60))
+    {
+        return 0.10;
+    }
+
+    // Bigger than 10 minutes but smaller than 10 hours
+    BOOST_ASSERT(duration > 5 * 60 && duration < 10 * 60 * 60);
```
daniel-j-h: Not sure what we're asserting here; we have the conditions right above?

daniel-j-h: Also we will hit these assertions if `duration == 5 * 60` or `duration == 10 * 60 * 60`.

Either remote this assertion or handle `>=` and `<=`.

src/engine/routing_algorithms/alternative_path_mld.cpp
```diff
@@ -76,6 +81,94 @@ struct WeightedViaNodeUnpackedPath
...
+
+    return a + b / (duration - d) + c / std::pow(duration - d, 3);
+}
+
+Parameters parametersFromRequest(const PhantomNodes &phantom_node_pair)
```
daniel-j-h: This is a good idea :bow:

src/engine/routing_algorithms/alternative_path_mld.cpp
```diff
@@ -76,6 +81,94 @@ struct WeightedViaNodeUnpackedPath
...
+{
+    Parameters parameters;
+
+    auto distance = util::coordinate_calculation::haversineDistance(
```
daniel-j-h: `const`

src/engine/routing_algorithms/alternative_path_mld.cpp
```diff
@@ -76,6 +81,94 @@ struct WeightedViaNodeUnpackedPath
...
+        phantom_node_pair.source_phantom.location, phantom_node_pair.target_phantom.location);
```
daniel-j-h: Should we estimate parameters based on the haversine distance or can we use the shortest route to estimate alternative parameters?

src/engine/routing_algorithms/alternative_path_mld.cpp
```diff
@@ -76,6 +81,94 @@ struct WeightedViaNodeUnpackedPath
...
+        parameters.kAlternativesToUnpackFactor = 6.0;
+        parameters.kCellsAtMostSameBy = 0.95;
+        parameters.kAtMostSameBy = 0.70;
+    }
+    // 100km
+    else if (distance < 100000.)
+    {
+        parameters.kAlternativesToUnpackFactor = 4.0;
+        parameters.kCellsAtMostSameBy = 0.75;
+    }
```
daniel-j-h: How did you come up with these specific settings? (I know you benchmarked, but this is not captured anywhere and I don't think we should just hard-code these values without explanation or scripts to reproduce them)

src/engine/routing_algorithms/alternative_path_mld.cpp
```diff
@@ -226,14 +288,13 @@ RandIt filterPackedPathsByCellSharing(RandIt first, RandIt last, const Partition
         return last;
 
     std::unordered_set<CellID> cells;
-    cells.reserve(size * (shortest_path.path.size() + 1) * (1. + kAtMostLongerBy));
+    cells.reserve(size * (shortest_path.path.size() + 1) * (1.25));
```
daniel-j-h: Why?

src/engine/routing_algorithms/alternative_path_mld.cpp
```diff
@@ -400,27 +466,35 @@ template <typename RandIt> RandIt filterUnpackedPathsBySharing(RandIt first, Ran
         return last;
 
     std::unordered_set<EdgeID> edges;
-    edges.reserve(size * shortest_path.edges.size() * (1. + kAtMostLongerBy));
+    edges.reserve(size * shortest_path.edges.size() * (1.25));
 
-    edges.insert(begin(shortest_path.edges), begin(shortest_path.edges));
+    edges.insert(begin(shortest_path.edges), end(shortest_path.edges));
```
daniel-j-h: Ouch

src/engine/routing_algorithms/alternative_path_mld.cpp
```diff
@@ -400,27 +466,35 @@ template <typename RandIt> RandIt filterUnpackedPathsBySharing(RandIt first, Ran
...
+                return duration + edge_duration;
+            }
+            return duration;
+        };
 
-        const auto difference = different / static_cast<double>(unpacked.edges.size());
-        BOOST_ASSERT(difference >= 0.);
-        BOOST_ASSERT(difference <= 1.);
+        const auto shared_weight =
+            std::accumulate(begin(unpacked.edges), end(unpacked.edges), 0, add_if_seen);
```
daniel-j-h: `EdgeWeight{0}` otherwise `shared_weight` will be deduced to `int`

src/engine/routing_algorithms/alternative_path_mld.cpp
```diff
@@ -670,7 +745,7 @@ makeCandidateVias(SearchEngineData<Algorithm> &search_engine_data,
     return candidate_vias;
 }
 
-} // anon. ns
+} // namespace
```
daniel-j-h: :thought_balloon:

src/engine/routing_algorithms/alternative_path_mld.cpp
```diff
@@ -730,6 +808,9 @@ InternalManyRoutesResult alternativePathSearch(SearchEngineData<Algorithm> &sear
     NodeID shortest_path_via = shortest_path_via_it->node;
     EdgeWeight shortest_path_weight = shortest_path_via_it->weight;
 
+    const double duration_estimation = shortest_path_weight / facade.GetWeightMultiplier();
+    parameters.kAtMostLongerBy = getLongerByFactorBasedOnDuration(duration_estimation);
```
daniel-j-h: Hm, modifying the parameter attribute in-place here is a bit ugly.. but I get why we do it only here.

src/engine/routing_algorithms/alternative_path_mld.cpp
```diff
@@ -850,7 +932,9 @@ InternalManyRoutesResult alternativePathSearch(SearchEngineData<Algorithm> &sear
 
     if (routes.size() > 1)
     {
-        routes_last = filterAnnotatedRoutesByStretch(routes_first + 1, routes_last, *routes_first);
+        parameters.kAtMostLongerBy = getLongerByFactorBasedOnDuration(routes_first->duration());
```
daniel-j-h: Same here

src/engine/routing_algorithms/alternative_path_mld.cpp
TheMarex: Yeah I wanted to make it explicit what the value is that we really use and not just transform the default in some places.

src/engine/routing_algorithms/alternative_path_mld.cpp
TheMarex: Yeah this is just to bound the values to the range for which we provided samples of durations/parameters to the curve fitting. Outside of that range you don't really know how well this will approximate your intended behavior so we just clamp it.

src/engine/routing_algorithms/alternative_path_mld.cpp
```diff
@@ -400,27 +471,35 @@ template <typename RandIt> RandIt filterUnpackedPathsBySharing(RandIt first, Ran
...
-        const auto difference = different / static_cast<double>(unpacked.edges.size());
-        BOOST_ASSERT(difference >= 0.);
-        BOOST_ASSERT(difference <= 1.);
+        const auto shared_weight =
+            std::accumulate(begin(unpacked.edges), end(unpacked.edges), EdgeWeight{0}, add_if_seen);
 
-        const auto sharing = 1. - difference;
+        unpacked.sharing = shared_weight / static_cast<double>(total_duration);
+        BOOST_ASSERT(sharing >= 0.);
+        BOOST_ASSERT(sharing <= 1.);
```
daniel-j-h: These assertions fail to compile on debug because we no longer have a `sharing` var in scope; need to be adapted.

Issue #5027, opened 2018-04-13
Save graph weight for MLD graph by node


TheMarex:
To optimize the memory storage for metric for MLD we can split the weight of a turn into:

`node_weight[v] + turn_weight[u, v]`

Right now we save the sum of both in a single `32bit` integer for the MLD graph and save it [here](https://github.com/Project-OSRM/osrm-backend/blob/master/src/updater/updater.cpp#L784-L785).
Instead of using the pre-computed sum, we could also compute it during `relaxOutgoingEdge` [here](https://github.com/Project-OSRM/osrm-backend/blob/master/include/engine/routing_algorithms/routing_base_mld.hpp#L217). The overhead of an additional addition operations is neglectable.

Right now we need 4 bytes for each edge for the `duration` and 4 bytes for each edge for the `weight`, so in total `8 bytes * |E|`. If we change this to `8 bytes * |V|` we would save 66% of the metric storage since `|V| ~ 0.33 * |E|.`

That would be a memory reduction of `30% * 66% = 19%` of total memory usage.

PR #5049, opened 2018-04-23
Use node weights instead of edge weights in MLD algorithm

oxidase:
# Issue

PR implements issue #5027 and can be merged after review, but if #4876 will be merged before then the commit 061a314c3 won't be required.

PR  removes `weight` and `duration`, `forward`, `backward` members from `EdgeBasedGraphEdgeData` and adds a new `customizer::MultiLevelGraph` that has `node_weights`, `node_weights`, `is_forward_edge`, `is_backward_edge` vectors that are used now to compute edge weights and durations as a sum of node weights and turn penalties. The size of `.osrm.mldgr` files for `california-latest.osm.pbf` changes from 382468608 to 254924800 bytes or 33% reduction.

Performance check on `california-latest.osm.pbf` with random A->B routes does not show regression:
compare

In addition PR fixes two issues:
* e637d6963 adds an exception throw if a shared region is not found to prevent UB
* cda8b1b28 uses the sign bit of node weight to mark oneway streets and correctly set node weight value to `INVALID_EDGE_WEIGHT` if the value was updated. The change must fix bugs with missing routes after traffic updates in CH algorithm.

/cc @TheMarex @freenerd 

## Tasklist

 - [ ] review
 - [ ] adjust for comments

## Requirements / Relations

#4876

### DIFF
--- a/include/customizer/edge_based_graph.hpp
+++ b/include/customizer/edge_based_graph.hpp
@@ -76,6 +76,8 @@ class MultiLevelGraph : public partitioner::MultiLevelGraph<EdgeDataT, Ownership
 
     EdgeWeight GetNodeWeight(NodeID node) const { return node_weights[node]; }
 
+    EdgeWeight GetNodeDuration(NodeID node) const { return node_durations[node]; }
+
     friend void
     serialization::read<EdgeDataT, Ownership>(storage::tar::FileReader &reader,
                                               const std::string &name,
--- a/include/customizer/edge_based_graph.hpp
+++ b/include/customizer/edge_based_graph.hpp
@@ -16,8 +16,10 @@ namespace osrm
 namespace customizer
 {
 
-// TODO: Change to turn_id only
-using EdgeBasedGraphEdgeData = partitioner::EdgeBasedGraphEdgeData;
+struct EdgeBasedGraphEdgeData
+{
+    NodeID turn_id; // ID of the edge based node (node based edge)
+};
 
 template <typename EdgeDataT, storage::Ownership Ownership> class MultiLevelGraph;
 
@@ -39,8 +41,8 @@ class MultiLevelGraph : public partitioner::MultiLevelGraph<EdgeDataT, Ownership
 {
   private:
     using SuperT = partitioner::MultiLevelGraph<EdgeDataT, Ownership>;
-    using SuperC = partitioner::MultiLevelGraph<partitioner::EdgeBasedGraphEdgeData,
-                                                storage::Ownership::Container>;
+    using PartitionerGraphT = partitioner::MultiLevelGraph<partitioner::EdgeBasedGraphEdgeData,
+                                                           storage::Ownership::Container>;
     template <typename T> using Vector = util::ViewOrVector<T, Ownership>;
 
   public:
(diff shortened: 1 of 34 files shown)


oxidase:
For `europe-latest.osm.pbf` dataset:
* master
  * static part has a size of 12345436828 bytes
  * updateable part has a size of 18792362496 bytes

* branch
  * static part has a size of 12345429128 bytes
  * updateable part has a size of 17417163264bytes

PR reduces the complete dataset size by 5% and the updateable part of the dataset by 7.5%:
* size of `input.osrm.mldgr` changes from 6804054528 to 4580641280 bytes
* dataset has two additional files `input.osrm.end` and `input.osrm.enw` with node durations and weights of size 422718464  bytes

/cc @TheMarex

include/customizer/edge_based_graph.hpp
```diff
@@ -16,28 +16,105 @@ namespace osrm
...
+        SuperT::edge_array.reserve(original_edge_array.size());
+        for (const auto &edge : original_edge_array)
+        {
+            SuperT::edge_array.push_back({edge.target, {edge.data.turn_id}});
+            is_forward_edge.push_back(edge.data.forward);
+            is_backward_edge.push_back(edge.data.backward);
+        }
+    }
+
+    MultiLevelGraph(Vector<typename SuperT::NodeArrayEntry> node_array_,
```
TheMarex: Using some class-scope type aliases might be easier to read, like `using NodeArrayEntry = typename SuperT::NodeArrayEntryT`.

include/engine/routing_algorithms/routing_base_mld.hpp
```diff
@@ -206,15 +206,22 @@ void relaxOutgoingEdges(const DataFacade<Algorithm> &facade,
...
         {
             const NodeID to = facade.GetTarget(edge);
 
             if (!facade.ExcludeNode(to) &&
                 checkParentCellRestriction(partition.GetCell(level + 1, to), args...))
             {
-                BOOST_ASSERT_MSG(edge_data.weight > 0, "edge_weight invalid");
-                const EdgeWeight to_weight = weight + edge_data.weight;
+                const auto node_weight =
+                    facade.GetNodeWeight(DIRECTION == FORWARD_DIRECTION ? node : to);
```
TheMarex: Good catch, didn't think about the reverse-case here. :+1:

include/partitioner/multi_level_graph.hpp
```diff
@@ -146,6 +149,14 @@ class MultiLevelGraph : public util::StaticGraph<EdgeDataT, Ownership>
         return max_border_node_id;
     }
 
+    auto data() &&
```
TheMarex: This is a kind of weird mechanism. Do you think we could also solve this by protected access to these members by the sub-class? Or is this a problem with templates?

include/storage/shared_data_index.hpp
```diff
@@ -46,33 +46,39 @@ class SharedDataIndex
...
     std::size_t GetBlockSize(const std::string &name) const
     {
-        const auto index_iter = block_to_region.find(name);
-        const auto &region = regions[index_iter->second];
+        const auto &region = GetBlockRegion(name);
         return region.layout.GetBlockSize(name);
     }
 
   private:
+    const AllocatedRegion &GetBlockRegion(const std::string &name) const
```
TheMarex: Nice refactor, thanks! :+1:

src/contractor/contractor.cpp
```diff
@@ -75,6 +75,12 @@ int Contractor::Run()
     EdgeID number_of_edge_based_nodes = updater.LoadAndUpdateEdgeExpandedGraph(
         edge_based_edge_list, node_weights, connectivity_checksum);
 
+    // Convert node weights for oneway streets to INVALID_EDGE_WEIGHT
+    for (auto &weight : node_weights)
+    {
+        weight = (weight & 0x80000000) ? INVALID_EDGE_WEIGHT : weight;
```
TheMarex: I don't understand this part? We change the weight to INVALID_EDGE_WEIGHT if we MSB is set, but why not save it as `INVALID_EDGE_WEIGHT` in the first place?

src/customize/customizer.cpp
```diff
@@ -119,8 +123,13 @@ int Customizer::Run(const CustomizationConfig &config)
...
     partitioner::files::readPartition(config.GetPath(".osrm.partition"), mlp);
 
+    std::vector<EdgeWeight> node_weights;
+    std::vector<EdgeDuration> node_durations; // TODO: to be removed later
     std::uint32_t connectivity_checksum = 0;
-    auto graph = LoadAndUpdateEdgeExpandedGraph(config, mlp, connectivity_checksum);
+    auto graph = LoadAndUpdateEdgeExpandedGraph(
+        config, mlp, node_weights, node_durations, connectivity_checksum);
+    BOOST_ASSERT(graph.GetNumberOfNodes() == node_weights.size());
+    std::for_each(node_weights.begin(), node_weights.end(), [](auto &w) { w &= 0x7fffffff; });
```
TheMarex: Same here, why can't we enforce the MSB clipping where this data is generated?

src/extractor/edge_based_graph_factory.cpp
```diff
@@ -138,9 +145,12 @@ NBGToEBG EdgeBasedGraphFactory::InsertEdgeBasedNode(const NodeID node_u, const N
     BOOST_ASSERT(nbe_to_ebn_mapping[edge_id_1] != SPECIAL_NODEID ||
                  nbe_to_ebn_mapping[edge_id_2] != SPECIAL_NODEID);
 
+    // TODO: use the sign bit to distinguish oneway streets that must
+    // have INVALID_EDGE_WEIGHT node weight values to enforce loop edges
+    // in contraction
     if (nbe_to_ebn_mapping[edge_id_1] != SPECIAL_NODEID &&
         nbe_to_ebn_mapping[edge_id_2] == SPECIAL_NODEID)
-        m_edge_based_node_weights[nbe_to_ebn_mapping[edge_id_1]] = INVALID_EDGE_WEIGHT;
+        m_edge_based_node_weights[nbe_to_ebn_mapping[edge_id_1]] |= 0x80000000;
```
TheMarex: I don't get what setting the MSB vs. `INVALID_EDGE_WEIGHT` gives us here?

include/partitioner/multi_level_graph.hpp
oxidase: It is not directly a problem with templates but with different `EdgeBasedGraphEdgeData` template arguments of customizer and partitioner graphs and `customizer::MultiLevelGraph` is not allowed to access protected members  of `partitioner::MultiLevelGraph`.

An rvalue qualifier `&&` is just a safety belt to prevent using an object instance after `graph.data()` call and explicit requirement for `std::move(graph).data()`. `&&` here and `std::move` at the calling place can be removed.

src/contractor/contractor.cpp
oxidase: In the extractor where the value is computed it is not possible to say how node weights will be used later:
* in contraction it must be INVALID_EDGE_WEIGHT to make loop edges at https://github.com/Project-OSRM/osrm-backend/blob/3ee94959eab7aacb314fc68220a67d842438bd13/src/contractor/graph_contractor.cpp#L193 If `node_weights[node]` is not `INVALID_EDGE_WEIGHT` then `path_weight` of the loop edge will be always larger than the node weight value and the condition will be always false.

* for MLD `node_weights[node]` must be a correct value  otherwise `GetNodeWeight` will return `INVALID_EDGE_WEIGHT` that will break routing

The MSB flag is used to distinguish these usage scenarios.

TheMarex: Oh the problem is the reverse search of MLD?  Okay yeah that is pretty ugly, but makes sense. I feel this deserves a better comment because this is a "hack".

TheMarex:
@oxidase thanks for answering my questions. I think this might only need a better comment and then this is ready to merge.

PR #5060, opened 2018-04-27
Improve speed of Map Matching

TheMarex:
# Issue

This PR optimizes the map matching plugin for speed. It uses two things to do that:
- The distance annotation work and improved distance calculation speed
- Caching the forward search space for every state transition.

For my test trace this yielded a speedup of 50%.

## Tasklist

 - [ ] run benchmark on real traces with bigger dataset 
 - [ ] review
 - [ ] adjust for comments

### DIFF
--- a/include/engine/routing_algorithms/routing_base.hpp
+++ b/include/engine/routing_algorithms/routing_base.hpp
@@ -321,24 +321,10 @@ void annotatePath(const FacadeT &facade,
     }
 }
 
-template <typename Algorithm>
-double getPathDistance(const DataFacade<Algorithm> &facade,
-                       const std::vector<PathData> &unpacked_path,
-                       const PhantomNode &source_phantom,
-                       const PhantomNode &target_phantom)
-{
-    double distance = 0;
-    auto prev_coordinate = source_phantom.location;
-    for (const auto &p : unpacked_path)
-    {
-        const auto current_coordinate = facade.GetCoordinateOfNode(p.turn_via_node);
-        distance += util::coordinate_calculation::fccApproximateDistance(prev_coordinate, current_coordinate);
-        prev_coordinate = current_coordinate;
-    }
-    distance += util::coordinate_calculation::fccApproximateDistance(prev_coordinate, target_phantom.location);
-
-    return distance;
-}
+void adjustPathDistanceToPhantomNodes(const std::vector<NodeID> &path,
+                                      const PhantomNode &source_phantom,
+                                      const PhantomNode &target_phantom,
+                                      EdgeDistance &distance);
 
 template <typename AlgorithmT>
 InternalRouteResult extractRoute(const DataFacade<AlgorithmT> &facade,
--- a/include/engine/routing_algorithms/routing_base.hpp
+++ b/include/engine/routing_algorithms/routing_base.hpp
@@ -44,50 +44,18 @@ bool needsLoopBackwards(const PhantomNode &source_phantom, const PhantomNode &ta
 bool needsLoopForward(const PhantomNodes &phantoms);
 bool needsLoopBackwards(const PhantomNodes &phantoms);
 
-template <typename Heap>
-void insertNodesInHeaps(Heap &forward_heap, Heap &reverse_heap, const PhantomNodes &nodes)
+namespace detail
 {
-    const auto &source = nodes.source_phantom;
-    if (source.IsValidForwardSource())
-    {
-        forward_heap.Insert(source.forward_segment_id.id,
-                            -source.GetForwardWeightPlusOffset(),
-                            source.forward_segment_id.id);
-    }
-
-    if (source.IsValidReverseSource())
-    {
-        forward_heap.Insert(source.reverse_segment_id.id,
-                            -source.GetReverseWeightPlusOffset(),
-                            source.reverse_segment_id.id);
-    }
-
-    const auto &target = nodes.target_phantom;
-    if (target.IsValidForwardTarget())
-    {
-        reverse_heap.Insert(target.forward_segment_id.id,
-                            target.GetForwardWeightPlusOffset(),
-                            target.forward_segment_id.id);
-    }
-
-    if (target.IsValidReverseTarget())
-    {
-        reverse_heap.Insert(target.reverse_segment_id.id,
-                            target.GetReverseWeightPlusOffset(),
-                            target.reverse_segment_id.id);
-    }
-}
-
-template <typename ManyToManyQueryHeap>
-void insertSourceInHeap(ManyToManyQueryHeap &heap, const PhantomNode &phantom_node)
+template <typename Algorithm>
+void insertSourceInHeap(typename SearchEngineData<Algorithm>::ManyToManyQueryHeap &heap, const PhantomNode &phantom_node)
 {
-    if (phantom_node.IsValidForwardSource())
+    if (phantom_node.IsValidForwardTarget())
     {
         heap.Insert(phantom_node.forward_segment_id.id,
                     -phantom_node.GetForwardWeightPlusOffset(),
                     {phantom_node.forward_segment_id.id, -phantom_node.GetForwardDuration()});
     }
-    if (phantom_node.IsValidReverseSource())
+    if (phantom_node.IsValidReverseTarget())
     {
         heap.Insert(phantom_node.reverse_segment_id.id,
                     -phantom_node.GetReverseWeightPlusOffset(),
@@ -95,8 +63,8 @@ void insertSourceInHeap(ManyToManyQueryHeap &heap, const PhantomNode &phantom_no
     }
 }
 
-template <typename ManyToManyQueryHeap>
-void insertTargetInHeap(ManyToManyQueryHeap &heap, const PhantomNode &phantom_node)
+template <typename Algorithm>
+void insertTargetInHeap(typename SearchEngineData<Algorithm>::ManyToManyQueryHeap &heap, const PhantomNode &phantom_node)
 {
     if (phantom_node.IsValidForwardTarget())
     {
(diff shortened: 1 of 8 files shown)


features/testbot/matching.feature
```diff
@@ -792,4 +792,4 @@ Feature: Basic Map Matching
         When I match I should get
           | trace | geometry             | a:distance | a:duration | a:weight | duration |
           | 2345  | 1.00018,1,1.000315,1 | 15.013264  | 1.5        | 1.5      | 1.5      |
-          | 4321  | 1.00027,1,1.000135,1 | 15.013264  | 1.5        | 1.5      | 1.5      |
\ No newline at end of file
+          | 4321  | 1.00027,1,1.000135,1 | 15.013264  | 1.5        | 1.5      | 1.5      |
```
daniel-j-h: Why do these tests change when the PR says it's only optimizing the map-matching performance?

include/engine/routing_algorithms/routing_base.hpp
```diff
@@ -44,71 +44,131 @@ bool needsLoopBackwards(const PhantomNode &source_phantom, const PhantomNode &ta
 bool needsLoopForward(const PhantomNodes &phantoms);
 bool needsLoopBackwards(const PhantomNodes &phantoms);
 
-template <typename Heap>
-void insertNodesInHeaps(Heap &forward_heap, Heap &reverse_heap, const PhantomNodes &nodes)
+namespace detail
+{
+template <typename Algorithm>
+void insertSourceInHeap(typename SearchEngineData<Algorithm>::ManyToManyQueryHeap &heap, const PhantomNode &phantom_node)
```
daniel-j-h: Hm, why only template the algorithm and not simply the heap? Does it make sense to restricti it to a specific heap impl.?

include/engine/routing_algorithms/routing_base.hpp
```diff
@@ -44,71 +44,131 @@ bool needsLoopBackwards(const PhantomNode &source_phantom, const PhantomNode &ta
...
     if (phantom_node.IsValidReverseTarget())
     {
         heap.Insert(phantom_node.reverse_segment_id.id,
                     phantom_node.GetReverseWeightPlusOffset(),
-                    {phantom_node.reverse_segment_id.id, phantom_node.GetReverseDuration()});
+                    phantom_node.reverse_segment_id.id);
+    }
+}
+}
+
```
daniel-j-h: Below: looks like we are doing the same for all algorithms? Do we need to be specific about the algorithm then or can we simply have generic functions passing the parameters on to the downstream functions?

include/engine/routing_algorithms/routing_base.hpp
```diff
@@ -44,71 +44,131 @@ bool needsLoopBackwards(const PhantomNode &source_phantom, const PhantomNode &ta
...
+{
+    detail::insertSourceInHeap<mld::Algorithm>(heap, phantom_node);
+}
+inline void insertSourceInHeap(typename SearchEngineData<ch::Algorithm>::QueryHeap &heap, const PhantomNode &phantom_node)
+{
+    detail::insertSourceInHeap<ch::Algorithm>(heap, phantom_node);
+}
+
+template <typename Heap>
+void insertNodesInHeaps(Heap &forward_heap, Heap &reverse_heap, const PhantomNodes &nodes)
```
daniel-j-h: Here we template the heap in contrast to above where we keep the heap impl. fixed and only template the algorithm - why?

include/engine/routing_algorithms/routing_base.hpp
```diff
@@ -321,58 +381,10 @@ void annotatePath(const FacadeT &facade,
...
-    const double aharv = sin_dlat * sin_dlat + prev_cos * current_cos * sin_dlon * sin_dlon;
-    const double charv = 2. * std::atan2(std::sqrt(aharv), std::sqrt(1.0 - aharv));
-    distance += EARTH_RADIUS * charv;
-
-    return distance;
-}
+void adjustPathDistanceToPhantomNodes(const std::vector<NodeID> &path,
+                                      const PhantomNode &source_phantom,
+                                      const PhantomNode &target_phantom,
+                                      EdgeDistance &distance);
```
daniel-j-h: out-ref? can you return by value?

include/engine/routing_algorithms/routing_base_mld.hpp
```diff
@@ -680,11 +682,20 @@ double getNetworkDistance(SearchEngineData<Algorithm> &engine_working_data,
         return std::numeric_limits<double>::max();
     }
 
-    std::vector<PathData> unpacked_path;
+    EdgeDistance distance = 0;
 
-    annotatePath(facade, phantom_nodes, unpacked_nodes, unpacked_edges, unpacked_path);
+    if (!unpacked_nodes.empty())
+    {
+        for (auto node_iter = unpacked_nodes.begin(); node_iter != std::prev(unpacked_nodes.end()); node_iter++)
```
daniel-j-h: Why not include the last here?

include/engine/routing_algorithms/routing_base_mld.hpp
```diff
@@ -680,11 +682,20 @@ double getNetworkDistance(SearchEngineData<Algorithm> &engine_working_data,
...
+        {
+            distance += computeEdgeDistance(facade, *node_iter);
+        }
+    }
```
daniel-j-h: Can this be `std::accumulate`?

src/engine/routing_algorithms/many_to_many_ch.cpp
```diff
@@ -240,74 +240,11 @@ void calculateDistances(typename SearchEngineData<ch::Algorithm>::ManyToManyQuer
         }
         if (!packed_leg.empty())
         {
-            auto annotation =
+            EdgeDistance annotation =
                 ch::calculateEBGNodeAnnotations(facade, packed_leg.begin(), packed_leg.end());
+            adjustPathDistanceToPhantomNodes(packed_leg, source_phantom, target_phantom, annotation);
```
daniel-j-h: Here: not sure if out-ref is a good idea. Maybe pass in and take return val again?

src/engine/routing_algorithms/routing_base.cpp
```diff
@@ -33,6 +33,75 @@ bool needsLoopBackwards(const PhantomNodes &phantoms)
...
+            //               s<---      <-- offsets
+            //         t<---------      <-- subtract source offset from target offset
+            //         ......           <-- want this distance as result
+            // entry 0---1---2---3---   <-- 3 is exit node
+            distance = target_phantom.GetReverseDistance() - source_phantom.GetReverseDistance();
+        }
+    }
+
+    // guard against underflow errors caused by rounding
+    distance = std::max<EdgeDistance>(0.0f, distance);
```
daniel-j-h: Hm why 0.0f literal? EdgeDistance should be integral? Or maybe use `EdgeDistance{0}`

src/engine/routing_algorithms/routing_base.cpp
```diff
@@ -33,6 +33,75 @@ bool needsLoopBackwards(const PhantomNodes &phantoms)
...
+        {
+            //               s<---      <-- offsets
+            //         t<---------      <-- subtract source offset from target offset
+            //         ......           <-- want this distance as result
+            // entry 0---1---2---3---   <-- 3 is exit node
+            distance = target_phantom.GetReverseDistance() - source_phantom.GetReverseDistance();
+        }
+    }
+
+    // guard against underflow errors caused by rounding
```
daniel-j-h: Should there be an assert here checking for an epsilon making sure it's _only_ due to rounding and not due to bugs in our code? Otherwise a negative erroneous distance will be silently ignored here.

src/engine/routing_algorithms/routing_base_ch.cpp
```diff
@@ -110,10 +110,14 @@ void search(SearchEngineData<Algorithm> & /*engine_working_data*/,
     weight = weight_upper_bound;
 
     // get offset to account for offsets on phantom nodes on compressed edges
-    const auto min_edge_offset = std::min(0, forward_heap.MinKey());
-    BOOST_ASSERT(min_edge_offset <= 0);
+    auto min_edge_offset = 0;
```
daniel-j-h: This will be int - do we want to be explicit about the type here?

src/engine/routing_algorithms/routing_base_ch.cpp
```diff
@@ -199,14 +198,31 @@ double getNetworkDistance(SearchEngineData<Algorithm> &engine_working_data,
...
+
+    std::vector<NodeID> unpacked_nodes;
+    unpacked_nodes.reserve(packed_path.size());
+    if (!packed_path.empty())
+    {
+        unpacked_nodes.push_back(packed_path.front());
+        unpackPath(facade,
+                   packed_path.begin(),
+                   packed_path.end(),
+                   [&](std::pair<NodeID, NodeID> &edge, const auto &) {
```
daniel-j-h: const ref for the pair? Maybe const auto& even

src/engine/routing_algorithms/routing_base_ch.cpp
```diff
@@ -199,14 +198,31 @@ double getNetworkDistance(SearchEngineData<Algorithm> &engine_working_data,
...
+                       BOOST_ASSERT(edge.first == unpacked_nodes.back());
+                       unpacked_nodes.push_back(edge.second);
+                   });
+
+        for (auto node_iter = unpacked_nodes.begin(); node_iter != std::prev(unpacked_nodes.end());
```
daniel-j-h: Same as above: last and accumulate

src/util/coordinate_calculation.cpp
```diff
@@ -22,6 +22,11 @@ namespace coordinate_calculation
 
 namespace
 {
+
+// earth radius varies between 6,356.750-6,378.135 km (3,949.901-3,963.189mi)
+// The IUGG value for the equatorial radius is 6378.137 km (3963.19 miles)
+const constexpr long double EARTH_RADIUS = 6372797.560856;
```
daniel-j-h: I don't think this has to be long double? Why do we need it?

features/testbot/matching.feature
TheMarex: It impacts numeric stability a little bit, which can leas to different matchings being chosen.

include/engine/routing_algorithms/routing_base.hpp
TheMarex: We need this in order to provide overloads for the normal QueryHeap and ManyToMany heap.

include/engine/routing_algorithms/routing_base_mld.hpp
TheMarex: In the edge-based-graph the last edge-based-node is not traveled over (we arrive at the beginning of that street segment).

include/engine/routing_algorithms/routing_base.hpp
TheMarex: The problem is that we can't infer the algorithm from the `SearchEngineData` since we only pass in sub-type `typename SearchEngineData<Algorithm>::QueryHeap`. I would love to do this with automatic code generation but I did not find a way to do it?

include/engine/routing_algorithms/routing_base.hpp
TheMarex: This a generic wrapper that dispatches to the overloaded functions for `insert{Source/Target}InHeap`.

src/util/coordinate_calculation.cpp
TheMarex: You are right up to 15 digits should be fine.

TheMarex:
@daniel-j-h Addressed your PR comments and added a measurements for the speedups on Berlin.

include/engine/routing_algorithms/routing_base.hpp
daniel-j-h: Hm can you simply put a member type into the sub-class as in `vector<T>::value_type` ?

Issue #4823, opened 2018-01-23
barrier=height_restriction should not prevent routing


osrmuser:
I think I have found a problem related to the handling of barrier=height_restriction on http://map.project-osrm.org/?z=18&center=52.140220%2C11.631485&loc=52.139836%2C11.630348&loc=52.139675%2C11.631544&hl=en&alt=0. It looks like OSRM avoids node https://www.openstreetmap.org/node/3247447825, probably beacause it is handled as an unpassable barrier without something like an explicit access=yes. But https://wiki.openstreetmap.org/wiki/Tag%3Abarrier%3Dheight_restrictor does not mention any explicit acces rule as required. 

Since I have found a barrier whitelist in https://github.com/Project-OSRM/osrm-backend/blob/master/profiles/car.lua#L49, which seems to take care of similar problems, height_restriction probably should be added there (and maybe in other profiles too).

MichalPP:
maybe it's time to switch from barrier_whitelist to barrier_blacklist #3890

osrmuser:
Verified fixed, thx MichalPP and TheMarex

Issue #3890, opened 2017-04-01
in foot profile, change barrier_whitelist to barrier_blacklist


MichalPP:
current whitelisting of barrier serves well for car routing, but due to man barrier types focused on pedestrians (and connected spelling errors), it is very inconvenient to keep a whitelist. 

the proposal is to change it to blacklist, and allow `foot` routing over all mapping and spelling errors. 

@emiltin what about bicycle profile? should it be blacklist or whitelist?

emiltin:
what are examples of barriers that are currently incorrectly blocking pedestrian routing?

MichalPP:
looking at data of Slovakia, it is rather a long list: `bump_gate,debris,derbis,door,fallen_tree, fallen_trees,fullheight_turnstile,full-height_turnstile,guard_rail,h,hampshire_gate,height_restrictor,chain, jersey_barrier,kissing_gate,lift_gate,log,organic_waste,rope,scrub,stile,sump_buster,swing_gata,swing:gate, swing_gate,turnstile,windthrow` 

some of them are spelling errors and wrong mapping, but still it is a long list. while the blacklist is rather small (yes, fence, wall).

database query was `select string_agg(distinct barrier, ',') from planet_osm_point where barrier is not null and barrier not in ('cycle_barrier', 'bollard', 'entrance', 'cattle_grid', 'border_control', 'toll_booth', 'sally_port', 'gate', 'no', 'kerb', 'block');`

emiltin:
for pedestrians i agree most of these should be allowed, although we should go through the list and consider if some should be blacklisted.

for bikes i think it also makes sense to move to a blacklist. stiles and certainly full-height turnstiles http://wiki.openstreetmap.org/wiki/Tag:barrier%3Dfull-height_turnstile should be blacklisted for bikes.

stuff like http://wiki.openstreetmap.org/wiki/Tag:barrier%3Dswing_gate is a grey area. if they are closed and you can't go around them, it's certainly not very practical to pass on a bike.

MichalPP:
I did some data checking for SK: 
```
 barrier     |   access    | foot | bicycle | count 
----------------+-------------+------+---------+-------
 block          |             | yes  | yes     |   161
 block          |             |      |         |   136
 block          |             | yes  |         |    21
 bollard        |             | yes  | yes     |   576
 bollard        |             |      |         |   373
 bollard        |             | yes  |         |    32
 border_control |             |      |         |    50
[...]

PR #5076, opened 2018-05-12
bicycle profile: change barrier_whitelist to barrier_blacklist

### LINKED ISSUE #5067: Route over barrier=kerb
Someone on IRC just flagged that we're not routing over `barrier=kerb` which might be tagged e.g. on driveways. It looks like it's only whitelisted in the foot profile. Investigate.

https://github.com/Project-OSRM/osrm-backend/blob/faff2c774d09a7227c77ae4fa40d22d54bb00b45/profiles/foot.lua#L38

MichalPP:
change barrier_whitelist to barrier blacklist  in foot profile
fix #5067 and #3890 and #4823

### DIFF
--- a/features/bicycle/barrier.feature
+++ b/features/bicycle/barrier.feature
@@ -19,7 +19,7 @@ Feature: Barriers
             | entrance       | x     |
             | wall           |       |
             | fence          |       |
-            | some_tag       |       |
+            | some_tag       | x     |
             | block          | x     |
 
     Scenario: Bike - Access tag trumphs barriers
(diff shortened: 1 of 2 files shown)


MichalPP:
and maybe routing over `barrier=some_tag` should be allowed.

TheMarex:
@MichalPP thanks for taking a stab at this! Can you explain a little bit more why you think a `blacklist` is more appropriate? Our reasoning was that we should optimize for safety and should thus avoid paths with unknown obstacles.

PR #5077, opened 2018-05-12
foot profile: change barrier_whitelist to barrier blacklist

### LINKED ISSUE #5067: Route over barrier=kerb

MichalPP:
change barrier_whitelist to barrier blacklist  in foot profile
fix #5067 and #3890 and #4823

### DIFF
--- a/features/foot/barrier.feature
+++ b/features/foot/barrier.feature
@@ -19,7 +19,7 @@ Feature: Barriers
             | entrance       | x     |
             | wall           |       |
             | fence          |       |
-            | some_tag       |       |
+            | some_tag       | x     |
             | block          | x     |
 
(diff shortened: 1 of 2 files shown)


TheMarex:
@MichalPP seems like there are also some test failures similar to the companion PR. Other than that we can go ahead and merge this. :-)

MichalPP:
@TheMarex I've added the missing bit.

not sure what the failing check means...

PR #5081, opened 2018-05-19
Bicycle route fix #5032

### LINKED ISSUE #5032: prefer ways with bicycle routes

Implementation:
**Add this function:**
```(lua)
        end
        return nil;
end
```
**Create handler**
```(lua)
    end
    end
end
```
and run this handler.

MichalPP:
# Issue

in bicycle profile, prefer ways which are a part of route=bicycle relation. fix #5032


MichalPP:
I'll try new PR without the barrier mess

PR #5082, opened 2018-05-19
prefer ways on route=bicycle

### LINKED ISSUE #5032: prefer ways with bicycle routes

MichalPP:
# Issue

change forward/backward rate on ways, that are a part of route=bicycle. fix  #5032

### DIFF
--- a/profiles/bicycle.lua
+++ b/profiles/bicycle.lua
@@ -32,6 +32,7 @@ function setup()
     turn_penalty              = 6,
     turn_bias                 = 1.4,
     use_public_transport      = true,
+    route_preference          = 1.1,
 
     allowed_start_modes = Set {
       mode.cycling,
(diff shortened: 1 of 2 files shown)


MichalPP:
@chaupow how do I ask travis/appveyor to run the checks again? (half of them were run before I included the third commit)

TheMarex:
@MichalPP hrm they should re-run after every commit. Looks like some failed. Do they work for you locally?

chaupow:
Sorry somehow this slipped through my mailbox attention time span.

As @TheMarex pointed out seems they failed? 🤔 

But if you encounter this problem again, you can try pushing 
`git commit --allow-empty -m "Trigger travis"`

MichalPP:
commit 2 and 3 should be just one commit (or 3 should be done before 2). 3 includes the function and 2 uses it.

(I haven't yet mastered git on command line)

MichalPP:
closing in favour of https://github.com/Project-OSRM/osrm-backend/pull/5092 

my git knowledge is increasing every minute :)

PR #5092, opened 2018-05-31
prefer ways on route=bicycle

### LINKED ISSUE #5032: prefer ways with bicycle routes

MichalPP:
change forward/backward rate on ways, that are a part of route=bicycle. fix #5032

@chaupow @emiltin 

open issues:
* it probably works only if weight_name is cyclability
* fix forward/backward roles (my lua knowledge is too short for this)
* some tests?

### DIFF
--- a/profiles/lib/relations.lua
+++ b/profiles/lib/relations.lua
@@ -258,4 +258,18 @@ function Relations.process_way_refs(way, relations, result)
   end
 end
 
+function Relations.filter_relations(relations, way, key, value, ret, forward)
+  -- if any of way's relation have key=value, return tag ret; else return nil
+  -- todo: check backward and forward
+  if not forward then forward = 'forward'; end -- ignored at the moment
+  local rel_id_list = relations:get_relations(way)
+  for i, rel_id in ipairs(rel_id_list) do
+    local rel = relations:relation(rel_id);
+    local p = rel:get_value_by_key(key);
+    if value == '*' and p then return rel:get_value_by_key(ret); end
+    if p == value then return rel:get_value_by_key(ret); end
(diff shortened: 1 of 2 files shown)


MichalPP:
travis reported errors that it was unable to download some files, otherwise everything passed.

MichalPP:
@TheMarex @chaupow any news on merging?

chaupow:
Looks good! Do you have screenshots of routes that were weird before and are now fixed (for documentation purposes)? 🙂

Also, it would be great if we can find some cucumber tests there are currently failing on master and are fixed with this PR

profiles/lib/relations.lua
```diff
@@ -258,4 +258,18 @@ function Relations.process_way_refs(way, relations, result)
   end
 end
 
+function Relations.filter_relations(relations, way, key, value, ret, forward)
+  -- if any of way's relation have key=value, return tag ret; else return nil
+  -- todo: check backward and forward
```
emiltin: is forward/backward actually handled by this function?

MichalPP: no, my lua knowledge is not that good. (plus I don't know of any such route in my local area).

could you add that check?

MichalPP:
@chaupow @emiltin any chance of merging? my fork is getting too away from the master...

emiltin:
i think forward/backward roles should be handled. other than than i think it would be great to get this merged.

DennisOSRM:

PR #5101, opened 2018-06-06
 Support maxlength and maxweight in car profile 

frodrigo:
Support in car profile:
* maxlength
* maxweight 

like maxheight and maxwidth.


frodrigo:
Thank

Issue #5146, opened 2018-07-17
add su-exec package and osrm user to docker container and bump alpine version


chesty:
add su-exec and osrm user, change README.md to use su-exec osrm in osrm-routed example
bump alpine version

PR #5147, opened 2018-07-17
Docker work

chesty:
# Issue  #5146

General security practices say you run with the least privileges required, network daemons are particularly at risk when running as root as a bug in the daemon will give a remote attacker root access in the container, it's easier then to break out and get root on the host. Docker says in https://docs.docker.com/engine/security/security/#conclusions

"Docker containers are, by default, quite secure; especially if you run your processes as non-privileged users inside the container."

add su-exec and osrm user, change README.md to use su-exec osrm in osrm-routed example
bump alpine version

### DIFF
--- a/README.md
+++ b/README.md
@@ -67,7 +67,7 @@ The flag `-v $(pwd):/data` creates the directory `/data` inside the docker conta
 
 Note that `berlin-latest.osrm` has a different file extension. 
 
-    docker run -t -i -p 5000:5000 -v $(pwd):/data osrm/osrm-backend osrm-routed --algorithm mld /data/berlin-latest.osrm
+    docker run -t -i -p 5000:5000 -v $(pwd):/data osrm/osrm-backend su-exec osrm osrm-routed --algorithm mld /data/berlin-latest.osrm
 
 Make requests against the HTTP server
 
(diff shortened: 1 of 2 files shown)


danpat:
@chesty Can you please update your commit message to explain the reason for the change?  What problem does this solve?

chesty:
I haven't sent many pull requests, noted for future reference to explain why not what, or why and what.

chesty:
It's also possible to run as non root with

`docker run -t -i -p 5000:5000 -v $(pwd):/data -u nobody osrm/osrm-backend osrm-routed --algorithm mld /data/berlin-latest.osrm`

So no need to add su-exec, maybe just update the README.md? only trick is berlin-latest.osrm.fileIndex gets created without world read access, so user nobody can't read it. `chmod +r  berlin-latest.osrm.fileIndex` fixes it. All the other osrm.* files are world readable, I'm not sure why osrm.fileIndex is special? I don't think it's a major deal to include su-exec though, it makes the container more flexible on how someone might want to use it.

I guess if you run docker as an ordinary user and you have write access to $(pwd) you can do all of the steps like

`docker run -t -v $(pwd):/data -u $UID osrm/osrm-backend osrm-extract -p /opt/car.lua /data/berlin-latest.osm.pbf`

then osrm.fileIndex is readable by $UID

TheMarex:
Running as root user inside the docker container is standard practice. The docker container might be run as non-root user which is what usually happens.

PR #5150, opened 2018-07-17
[WIP] In lua, replace access to speed by functions

frodrigo:
Replace the call to constant speed value by functions.
The idea is to allow more dynamic speed usage without the need to edit way_handlers.lua.
This patch originally come from the Uban Density contrib profile : https://github.com/Project-OSRM/osrm-profiles-contrib/tree/master/5/18/urban-density
But It should be applicable to other dynamic speeds setting like using DEM, raster or --location-dependent-data.
My goal is to use it as basses for Uban Density contrib profile, and make the speed table subject of --location-dependent-data based with countries.

After test, there is no noticeable overload of osrm-extraction.

### DIFF
--- a/profiles/car.lua
+++ b/profiles/car.lua
@@ -10,6 +10,7 @@ find_access_tag = require("lib/access").find_access_tag
 limit = require("lib/maxspeed").limit
 Utils = require("lib/utils")
 Measure = require("lib/measure")
+Tags = require('lib/tags')
 
 function setup()
   return {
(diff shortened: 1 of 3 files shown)


emiltin:
I'n not sure I see the benefit of this approach. Why insert functions in a table, when you can just modify the code?

Handlers is a way to simply the code. If they don't work for you, I think it's much simple to just write your own code/handlers.

frodrigo:
My point of view is that the handler are doing good job, and I want to keep it. Most of the time I just need to adjust the handlers, not remove or make new ones.
The handlers are already parametrable at some point, eg. for classes or other properties from setup.
The speeds is the more important point of profiles, and it's the main point of interest and change. I think it(s worth it to have more fine adjustment on speed setting while maintaining the rest of the code reusable.

Ahlid:
I think this is usefull and I will need this because I've more than one speed based on alot of variables

DennisOSRM:

Issue #5156, opened 2018-07-31
GDB can not debug routed after block all signals


wangyoucao577:
My test environment: 
- build on WSL(Windows Subsystem for Linux)
  - `Windows 10 Pro 1803`
  - `Ubuntu 18.04 LTS`
  - `gcc (Ubuntu 7.3.0-16ubuntu3) 7.3.0`
- Editor & Debugger
  - `Visual Studio Code 1.25.1`
  - `ms-vscode.cpptools 0.17.7`
  - `GNU gdb (Ubuntu 8.1-0ubuntu3) 8.1.0.20180409-git`

The remote debugger (`VSCode on Win10`, `gdb on Ubuntu 18.04 LTS`) works well from the beginning of the `main()` function. But when I step over the code `pthread_sigmask(SIG_BLOCK, &new_mask, &old_mask);` (`src/tools/routed.cpp(289)`), below breakpoints can not work and displayed `unverified breakpoint`.

wangyoucao577:
can be fixed by PR https://github.com/Project-OSRM/osrm-backend/pull/5157

PR #5157, opened 2018-07-31
Fix/signal block gdb

wangyoucao577:
# Issue

https://github.com/Project-OSRM/osrm-backend/issues/5156

## Tasklist

 - [ ] ~~update relevant [Wiki pages](https://github.com/Project-OSRM/osrm-backend/wiki)~~
 - [ ] ~~add tests (see [testing documentation](https://github.com/Project-OSRM/osrm-backend/blob/master/docs/testing.md)~~
 - [x] review
 - [x] adjust for comments
 - [ ] ~~cherry pick to release branch~~

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
   - Changes from 5.18.0:
     - Bugfixes:
       - FIXED: collapsing of ExitRoundabout instructions [#5114](https://github.com/Project-OSRM/osrm-backend/issues/5114)
+      - FIXED: fix osrm-routed gdb not work issue [#5156](https://github.com/Project-OSRM/osrm-backend/issues/5156)
 
 # 5.18.0
   - Changes from 5.17.0:
--- a/src/tools/routed.cpp
+++ b/src/tools/routed.cpp
@@ -273,10 +273,12 @@ int main(int argc, const char *argv[]) try
 
 #ifndef _WIN32
     int sig = 0;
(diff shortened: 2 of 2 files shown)


danpat:
@wangyoucao577 Can you update your commit message to explain *why* this change is necessary?  How does it solve the problem with gdb?

wangyoucao577:
As I mentioned in the issue https://github.com/Project-OSRM/osrm-backend/issues/5156, I met below issue on my Win10+WSL(`Ubuntu`) env:       
The remote debugger (`VSCode on Win10`, `gdb on Ubuntu 18.04 LTS`) works well from the beginning of the `main()` function. But when I step over the code `pthread_sigmask(SIG_BLOCK, &new_mask, &old_mask);` (`src/tools/routed.cpp(289)`), below breakpoints can not work and displayed `unverified breakpoint`. 

Then I found that `gdb breakpoint` need at least `SIGTRAP`, `SIGSTOP` to work (Please refer to [how debugger works](http://www.alexonlinux.com/how-debugger-works) for more details), but all signals are blocked in the source code until server initialized done.        

In my understanding, block all signals DO NOT make sense for this `osrm-routed` process. Only several signals (`SIGINT`, `SIGQUIT`, `SIGTERM`) are expected to wait. So I made the change and it works well for me then. 

@danpat Am I answer your question? And please correct me if I'm wrong. Thanks!

PR #5185, opened 2018-08-29
Increase allowed shared memory regions to 512 from ~120.

danpat:
# Issue

Currently, if you try to create more than 128 named datasets, some `std::uint8_t` values will overflow, and you'll end up overwriting other datasets with different names, and/or crashing.

This change increases the limit to 512.  Achieving this requires a few steps:

  1. `::ftok()` only uses the lowest 8 bits from the `id` field, so we need to use more than one lockfile name to differentiate unique keys.
  2. Various variables need to be increased from `std::uint8_t` to `std::uint16_t` to hold the larger values.
  3. The ring buffer for swapping datasets needs to be increased so that it doesn't run out of space when lots of datasets are active.

## Tasklist

 - [ ] review
 - [ ] adjust for comments

## Requirements / Relations

### DIFF
--- a/include/storage/shared_memory.hpp
+++ b/include/storage/shared_memory.hpp
@@ -34,10 +34,10 @@ namespace storage
 
 struct OSRMLockFile
 {
-    boost::filesystem::path operator()()
+    template <typename IdentifierT> boost::filesystem::path operator()(const IdentifierT &id)
     {
         boost::filesystem::path temp_dir = boost::filesystem::temp_directory_path();
-        boost::filesystem::path lock_file = temp_dir / "osrm.lock";
+        boost::filesystem::path lock_file = temp_dir / ("osrm-" + std::to_string(id) + ".lock");
         return lock_file;
     }
 };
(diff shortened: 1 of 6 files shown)


TheMarex:
> 128 shared memory regions should be enough for everybody

include/storage/shared_memory.hpp
```diff
@@ -34,10 +34,10 @@ namespace storage
 
 struct OSRMLockFile
 {
-    boost::filesystem::path operator()()
+    template <typename IdentifierT> boost::filesystem::path operator()(const IdentifierT &id)
```
chaupow: Oof, I was hoping that maybe we could get around to not templating this. I skimmed the calls a bit and it seems like this templated type **is** necessary. Just commenting on here in case you have an gedankenblitz (= an idea within a time span of < 1s) on how to avoid this template.

danpat: Meh, I could just make it support `operator()(int id)`, but I figured why not make it generally re-usable.

Whatever `id` is needs to be supported by `std::to_string(id)` lower down.

PR #5189, opened 2018-09-03
Make node bindings optionally return pre-rendered JSON buffers

danpat:
If you're using the NodeJS bindings with OSRM, and you're generating large responses, there can be significant overhead simply translating the OSRM JSON structure into a NodeJS object tree.

There are several use-cases where the next step you take is to simply render the NodeJS object into a JSON string, and send it over the network - using NodeJS as your OSRM HTTP server is one example of this.

This PR adds a new optional parameter to all the NodeJS API methods that allows you to configure the format of the response that OSRM returns.  It defaults to the current object tree, but if you supply `{format: 'json_buffer'}`, it will instead cause OSRM to pre-render the JSON tree on a worker thread, then return a NodeJS `Buffer` object containing the JSON response string.

In cases where you only wanted to return the result to a client over the network anyway, pre-rendering the JSON has a significant speedup.  It significantly unburdens the main NodeJS thread, allowing more concurrent work.

Timeline of a 5000x5000 `table` request and serialization over the network before this change:
screen shot 2018-09-01 at 12 30 04 pm

Timeline of a 5000x5000 `table` request and serialization over the network after this change:
screen shot 2018-09-01 at 12 29 19 pm

The `Main Thread` here represents the primary NodeJS event loop - less work done here means the event loop stays clear to handle i/O.

The time spent in the `pthread_body` has increased slightly in the second image - this shows the additional time spent rendering the JSON outside the main thread.

## Tasklist

 - [x] review
 - [x] adjust for comments

## Requirements / Relations

### DIFF
--- a/include/nodejs/node_osrm_support.hpp
+++ b/include/nodejs/node_osrm_support.hpp
@@ -2,6 +2,7 @@
 #define OSRM_BINDINGS_NODE_SUPPORT_HPP
 
 #include "nodejs/json_v8_renderer.hpp"
+#include "util/json_renderer.hpp"
 
 #include "osrm/approach.hpp"
 #include "osrm/bearing.hpp"
@@ -24,6 +25,7 @@
 #include <algorithm>
 #include <iostream>
 #include <iterator>
+#include <sstream>
 #include <string>
 #include <vector>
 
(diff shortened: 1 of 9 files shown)


include/nodejs/node_osrm_support.hpp
```diff
@@ -814,6 +831,54 @@ inline bool parseCommonParameters(const v8::Local<v8::Object> &obj, ParamType &p
...
+        }
+        else
+        {
+            Nan::ThrowError("format must be a string: \"object\" or \"json_buffer\"");
+            return {};
+        }
+    }
+    else
+    {
+        std::cout << "No format on config" << std::endl;
```
chaupow: is this necessary in release code or will this just fill our logs? :) 
Maybe we can remove this?

danpat: Yeah, this is debugging cruft, I'll 🔪 it.

PR #5019, opened 2018-04-12
Distances in MLD

ghoshkaj:
# Issue

https://github.com/Project-OSRM/osrm-backend/issues/5013 - Distances in Matrix, MLD

**Similar to unpacking in CH, we need to do two things for MLD**
1) From the bidirectional search and the unidirectional search in the table plugin for MLD, we need to recreate the "shortcut" or "packed" path.
2) We need to unpack the path and calculate the distance of each packed edge while doing so

**Details**
1) Finding the packed path
- ManyToMany/Bidirectional search: Path finding already happens in the forward routing steps and the backward routing steps in the MLD ManyToManySearch function. The idea is to use the heap from the ForwardRoutingStep and the search space buckets in the BackwardRoutingStep to recreate the packed or shortcutted path.

- OneToMany/Unidirectional Search: Path finding already happens during the unidirectional search - however, we need to extract the packed path from the heap and the multi-map that it uses.

2) We need as the original issue says, to repurpose the unpacking part of the Search function to compute the distances with. We also have to repurpose some of the heaps I believe, as the SearchEngineData heaps aren't `manyToMany` heaps, but rather, just regular heaps. 

## Tasklist
 - [ ] CHANGELOG.md entry ([How to write a changelog entry](http://keepachangelog.com/en/1.0.0/#how))
 - [ ] update relevant [Wiki pages](https://github.com/Project-OSRM/osrm-backend/wiki)
 - [ ] add tests (see [testing documentation](https://github.com/Project-OSRM/osrm-backend/blob/master/docs/testing.md)
 - [ ] review
 - [ ] adjust for comments
 - [ ] cherry pick to release branch

## Requirements / Relations

 Link any requirements here. Other pull requests this PR is based on?

PR #5197, opened 2018-09-05
Optionally cache CH/MLD edge distances

danpat:
After a long and sordid path, we added `annotations=distances` to the `table` plugin back in https://github.com/Project-OSRM/osrm-backend/pull/4990 and https://github.com/Project-OSRM/osrm-backend/pull/5019

These PRs implemented on-the-fly distance calculations - the table results are obtained, then the paths are unpacked to calculate the distances.

The upside to this is that it costs nothing in terms of resources (no extra preprocessing, no extra RAM) if you don't use this feature.

The downside is that it's a bit slow.  With CH, large matrices that take 1-2 seconds to complete returning just durations can take upwards of 60 seconds to complete when distances are included.

@niemeier-PSI submitted https://github.com/Project-OSRM/osrm-backend/pull/2764 way back which took the other approach - pre-calculate edge distances, and store them as an additional edge property.  This approach is nice and fast, but has the downside that it requires quite a bit more RAM.

Fast-forward to 2018, @TheMarex dropped https://github.com/Project-OSRM/osrm-backend/pull/4960 on us, which gives us a lot more flexibility with data formats.  Suddenly, it's significantly easier to make some things optional, without awful hacks in our file reading code.

This PR is a re-implementation of #2764 but the overhead is optional, and doesn't require compile-time modifications to the OSRM codebase to switch.

Once this is complete, #2764 will be redundant.

** WORK IN PROGRESS - DO NOT MERGE **

## Tasklist

 - [ ] review
 - [ ] adjust for comments

## Requirements / Relations


danpat:
I've implemented cached distance calculations for CH, roughly re-implementing #2764.  I haven't completed it for MLD yet, although some work has been done.

Next step is to refactor this to make it optional.

Performance-wise, it works as expected.  For a large 500x2000 matrix, the timing is roughly:

  Durations only - 1.5 seconds
  Durations+distances before this change - 55 seconds
  Durations+distances with this change - 1.9 seconds

All the `distance_matrix` tests pass in CH mode (but not MLD).

chaupow:
@danpat I hope you don't mind but as dev is happening over in https://github.com/Project-OSRM/osrm-backend/pull/5201, I am going to close this PR 🙏

PR #5214, opened 2018-09-27
In Belgium the maximum speed in rural areas is 70 in the region Flanders

frodrigo:
In Belgium the maximum speed in rural areas is 70 in the region Flanders
(hint from https://github.com/osm-fr/osmose-backend/issues/334)

### DIFF
--- a/profiles/car.lua
+++ b/profiles/car.lua
@@ -269,6 +269,7 @@ function setup()
       ["at:rural"] = 100,
       ["at:trunk"] = 100,
       ["be:motorway"] = 120,
+      ["be-vlg:rural"] = 70,
       ["by:urban"] = 60,
       ["by:motorway"] = 110,
       ["ch:rural"] = 80,
--- a/taginfo.json
+++ b/taginfo.json
@@ -148,6 +148,7 @@
         {"key": "maxspeed", "value": "AT:rural"},
         {"key": "maxspeed", "value": "AT:trunk"},
         {"key": "maxspeed", "value": "BE:motorway"},
+        {"key": "maxspeed", "value": "BE-VLG:rural"},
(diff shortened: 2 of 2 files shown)


chaupow:
Hi @hgcvm and @frodrigo !
thanks a lot for the PR ✨ 

I just tested this and noticed that `BE-VLG:rural` is often tagged as `source:maxspeed` which is not parsed by the current lua profiles at all. (This smells like a bug that is unrelated to this change).
See https://www.openstreetmap.org/way/593881430 as an example.

@frodrigo would you be interested in fixing that? I wil leave notes in your code on how to do it. I can then code review and merge 🙂

Let me know if you are time constrained, then I will take over.

chaupow:
I noticed I can't add proposals to your PR because the affected files weren't change so here my thoughts:

The relevant line is:
https://github.com/Project-OSRM/osrm-backend/blob/master/profiles/lib/way_handlers.lua#L435
```
local keys = Sequence { 'maxspeed:advisory', 'maxspeed' }
```

Per the wiki https://wiki.openstreetmap.org/wiki/Key:source:maxspeed we should add `source:maxspeed` and also `maxspeed:type` to the list of keys that we consider. Otherwise the tag will be ignored for maxspeed checks.

The function `Tags.get_forward_backward_by_set` [here](https://github.com/Project-OSRM/osrm-backend/blob/master/profiles/lib/way_handlers.lua#L436) will return the _first value_ found, so the order of `keys` matters.

I think it would be wise to check the tags in this order: `maxspeed, maxspeed:type, source:maxspeed, maxspeed:advisory` but open to other opinions here.

Let me know if you have questions.

frodrigo:
Maybe @alain-andre may be interested in implementing this.

danpat:
Superseded by #5217

PR #5241, opened 2018-10-19
connection.cpp: Check return value of write()

Lotterleben:
# Issue
Resolve one more instance of https://github.com/Project-OSRM/osrm-backend/issues/1682 .
I manually checked all other instances that

    ag '\.write\((.*),(.*)\);' | wc -l
    ag '\.read\((.*),(.*)\);' |wc -l

brought up and they were either minor/irrelevant (in examples or tests where this wasn't the focus) or resolved a few lines down (with another `if (!result)`).

I hope what I did makes sense; if not, I'd of course be happy to improve my PR according to your suggestions.

## Tasklist

 - [x] CHANGELOG.md entry ([How to write a changelog entry](http://keepachangelog.com/en/1.0.0/#how)) -- If I understand it correctly this doesn't qualify as a change that needs to be mentioned in the changelog
 - [x] update relevant [Wiki pages](https://github.com/Project-OSRM/osrm-backend/wiki) -- same
 - [x] add tests (see [testing documentation](https://github.com/Project-OSRM/osrm-backend/blob/master/docs/testing.md) -- same
 - [ ] review
 - [ ] adjust for comments

## Requirements / Relations

--

### DIFF
--- a/src/server/connection.cpp
+++ b/src/server/connection.cpp
@@ -1,3 +1,8 @@
+#include "osrm/error_codes.hpp"
+
+#include "util/exception.hpp"
+#include "util/exception_utils.hpp"
+
 #include "server/connection.hpp"
 #include "server/request_handler.hpp"
 #include "server/request_parser.hpp"
@@ -140,10 +145,15 @@ std::vector<char> Connection::compress_buffers(const std::vector<char> &uncompre
     boost::iostreams::filtering_ostream gzip_stream;
     gzip_stream.push(boost::iostreams::gzip_compressor(compression_parameters));
     gzip_stream.push(boost::iostreams::back_inserter(compressed_data));
-    gzip_stream.write(&uncompressed_data[0], uncompressed_data.size());
+    const auto &result = gzip_stream.write(&uncompressed_data[0], uncompressed_data.size());
+    if (!result)
+    {
+        throw util::RuntimeError(
+            "compress_buffers", ErrorCode::FileWriteError, SOURCE_REF, std::strerror(errno));
+    }
     boost::iostreams::close(gzip_stream);
 
     return compressed_data;
 }
-}
-}
+} // namespace server
+} // namespace osrm


src/server/connection.cpp
```diff
@@ -140,10 +145,15 @@ std::vector<char> Connection::compress_buffers(const std::vector<char> &uncompre
     boost::iostreams::filtering_ostream gzip_stream;
     gzip_stream.push(boost::iostreams::gzip_compressor(compression_parameters));
     gzip_stream.push(boost::iostreams::back_inserter(compressed_data));
-    gzip_stream.write(&uncompressed_data[0], uncompressed_data.size());
+    const auto &result = gzip_stream.write(&uncompressed_data[0], uncompressed_data.size());
+    if (!result)
+    {
+        throw util::RuntimeError(
```
danpat: Hmm, while I think it's good to check for error here, I'm not sure that throwing is the right solution - this will kill the server process, as I don't think we have a high-level `catch` on a per-request basis.

Write errors like this one should cause the client on the other end of the connection to receive an error message, but probably shouldn't cause the `osrm-routed` process to terminate.

Lotterleben: > Hmm, while I think it's good to check for error here, I'm not sure that throwing is the right solution - this will kill the server process, as I don't think we have a high-level `catch` on a per-request basis.
> 

Argh, of course! Sorry about that.

> Write errors like this one should cause the client on the other end of the connection to receive an error message, but probably shouldn't cause the `osrm-routed` process to terminate.

I'm probably missing something, but should the error message really trickle down all the way to the client? If I understand it correctly, `compress_buffers()` is only trying to apply the compression method set in the HTTP Accept-Encoding header, which, if I understand [this description](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Accept-Encoding) correctly, is more of a suggestion– "if you choose to compress, I understand the following methods". If the server fails to compress, no harm done to the client.

That being said, the compression failure should probably still be logged at the server?

TheMarex: Sorry for the belated answer, cleaning up the old PRs. :sparkles: 

@Lotterleben My reading of that description is the same, though I think we would need to return plaintext if compression fails, right?

So if anyone feels like picking this up again:
- We need to probably surface to the caller of this function that compression failed (either through the return value or through a custom exception).
- The call code needs to handle that and return plaintext instead.

DennisOSRM:

PR #5242, opened 2018-10-20
Support directly mmap-ing datafiles

danpat:
# Issue

This PR completes the vision initially described in https://github.com/Project-OSRM/osrm-backend/issues/1947.  It builds on the work done by @TheMarex in https://github.com/Project-OSRM/osrm-backend/pull/4881

Turns out, #4881 was almost there, it really only required the following changes:

  - ensuring the bit-vector on-disk layout matched the memory format expected by `vector_view<bool>`,
  - implementing a DataLayout override that understands data block offsets inside tarfiles directly, and
  - some special handling for the R-Tree `.fileIndex`

Fortunately, `.tar` files are block aligned on 512 byte boundaries - this conveniently means the data is properly word aligned if we directly `mmap` the entire `.tar` file.

There is no change here to the pre-processing tooling.  However, `osrm-routed` (and the NodeJS bindings), now support the `--mmap` (or `mmap_memory: true` for NodeJS) options.  When triggered, instead of loading all data into RAM, this causes OSRM to directly `mmap` datafiles.

This means you can run OSRM in quite constrained memory environments, if you're willing to sacrifice performance.  Supplying sufficient memory to allow page caching to have effect means that if your data has some common access patterns, you can achieve quite good performance with significantly less runtime RAM.  Realistic benchmarking for your particular scenario is necessary to decide how much RAM will give you acceptible performance, but at least now this knob exists to tune.

## Tasklist

 - [ ] Rename the ContiguousDataFacade - it no longer depends on the data being contiguous
 - [ ] Rename DataLayout to ContiguousDataLayout - this is where the contiguous idea is encalsulated
 - [x] review
 - [x] adjust for comments


include/storage/serialization.hpp
```diff
@@ -30,22 +30,29 @@ namespace serialization
...
+inline void
+unpackBits(T &data, const std::size_t base_index, const std::size_t count, const BlockT value)
 {
     static_assert(std::is_same<typename T::value_type, bool>::value, "value_type is not bool");
-    const BlockT mask = BlockT{1} << (count - 1);
-    for (std::size_t bit = 0; bit < count; value <<= 1, ++bit, ++index)
-        data[index] = value & mask;
+    for (std::size_t bit = 0; bit < count; ++bit)
+    {
+        data[base_index + bit] = value & (BlockT{1} << bit);
```
daniel-j-h: Potential for UB here in the left shift; maybe assert.

include/storage/serialization.hpp
```diff
@@ -266,14 +273,17 @@ template <typename K, typename V> void write(io::BufferWriter &writer, const std
     }
 }
 
-inline void read(io::BufferReader &reader, DataLayout &layout) { read(reader, layout.blocks); }
-
-inline void write(io::BufferWriter &writer, const DataLayout &layout)
+inline void read(io::BufferReader &reader, std::unique_ptr<BaseDataLayout> &layout)
```
daniel-j-h: `unique_ptr&` makes no sense; the unique ptr carries ownership semantics. If all you want is a writeable ref to the object you can simply pass a `BaseDataLayout&` and the call site de-refs the unique ptr. Then it's clear on the call site and the function's semantics are clear, too.

It's a lot better in rust :P

https://doc.rust-lang.org/book/second-edition/ch04-02-references-and-borrowing.html

include/storage/serialization.hpp
```diff
@@ -266,14 +273,17 @@ template <typename K, typename V> void write(io::BufferWriter &writer, const std
...
 {
-    write(writer, layout.blocks);
-}
-}
+    read(reader, layout->blocks);
 }
+
+inline void write(io::BufferWriter &writer, const std::unique_ptr<BaseDataLayout> &layout)
```
daniel-j-h: Same here: `const unique_ptr&` makes no sense. If you want a read-only ref to a `BaseDataLayout` then pass by `const BaseDataLayout&`. The unique ptr carries ownership semantics but you are misusing it here.

include/storage/shared_data_index.hpp
```diff
@@ -29,43 +29,43 @@ class SharedDataIndex
...
-            region.layout.List(name_prefix, out);
+            region.layout->List(name_prefix, out);
         }
     }
 
     template <typename T> auto GetBlockPtr(const std::string &name) const
     {
         const auto &region = GetBlockRegion(name);
-        return region.layout.GetBlockPtr<T>(region.memory_ptr, name);
+        return reinterpret_cast<T *>(region.layout->GetBlockPtr(region.memory_ptr, name));
```
daniel-j-h: Here and below: these cars are dangerous. Should we check for
- alignment
- T is trivially copyable

?

include/storage/shared_datatype.hpp
```diff
@@ -20,12 +20,12 @@ namespace osrm
 namespace storage
 {
 
-class DataLayout;
+class BaseDataLayout;
 namespace serialization
 {
-inline void read(io::BufferReader &reader, DataLayout &layout);
+inline void read(io::BufferReader &reader, std::unique_ptr<BaseDataLayout> &layout);
```
daniel-j-h: remove unique ptr

include/storage/shared_datatype.hpp
```diff
@@ -20,12 +20,12 @@ namespace osrm
...
 
-inline void write(io::BufferWriter &writer, const DataLayout &layout);
+inline void write(io::BufferWriter &writer, const std::unique_ptr<BaseDataLayout> &layout);
```
daniel-j-h: same

include/storage/shared_datatype.hpp
```diff
@@ -115,9 +92,59 @@ class DataLayout
         }
     }
 
+    virtual inline void *GetBlockPtr(char *shared_memory, const std::string &name) const = 0;
+
+    std::map<std::string, Block> blocks;
```
daniel-j-h: Can this be an unordered map?

include/storage/shared_datatype.hpp
```diff
@@ -115,9 +92,59 @@ class DataLayout
         }
     }
 
+    virtual inline void *GetBlockPtr(char *shared_memory, const std::string &name) const = 0;
```
daniel-j-h: Why a `char*`? The semantics are not clear here.

include/storage/shared_datatype.hpp
```diff
@@ -115,9 +92,59 @@ class DataLayout
...
+
+class DataLayout final : public BaseDataLayout
+{
+  public:
+    inline void SetBlock(const std::string &name, Block block) override final
+    {
+        blocks[name] = std::move(block);
+    }
+
+    inline uint64_t GetBlockEntries(const std::string &name) const override final
```
daniel-j-h: Here (and in other places): needs to be `std::uint64_t` and you need to include `<cstdint>`

include/storage/shared_datatype.hpp
```diff
@@ -157,7 +184,64 @@ class DataLayout
     }
 
     static constexpr std::size_t BLOCK_ALIGNMENT = 64;
-    std::map<std::string, Block> blocks;
+};
+
+class TarDataLayout final : public BaseDataLayout
+{
+  public:
+    inline void SetBlock(const std::string &name, Block block) override final
```
daniel-j-h: These functions seem to do more or less the same across data layout child classes. Do you think there should be a common base impl. so we don't have to repeat in child classes?

include/storage/storage.hpp
```diff
@@ -46,11 +46,15 @@ class Storage
     Storage(StorageConfig config);
 
     int Run(int max_wait, const std::string &name, bool only_metric);
-
-    void PopulateStaticLayout(DataLayout &layout);
-    void PopulateUpdatableLayout(DataLayout &layout);
     void PopulateStaticData(const SharedDataIndex &index);
     void PopulateUpdatableData(const SharedDataIndex &index);
+    void PopulateLayout(std::unique_ptr<storage::BaseDataLayout> &layout,
```
daniel-j-h: Here and below: unique ptr problem

include/storage/storage.hpp
```diff
@@ -46,11 +46,15 @@ class Storage
...
+                        std::vector<std::pair<bool, boost::filesystem::path>> files);
```
daniel-j-h: Semantics unclear - why pass the vector by value? Should this be a const ref?

include/util/mmap_file.hpp
```diff
@@ -15,14 +15,14 @@ namespace util
 
 namespace detail
 {
-template <typename T, typename RegionT>
-util::vector_view<T> mmapFile(const boost::filesystem::path &file, RegionT &region)
+template <typename T, typename MmapContainerT>
```
daniel-j-h: What is a `MmapContainerT`? What are the requirements? Should we have concepts for this? (example https://gist.github.com/daniel-j-h/af3b08a2a569d91d638006cbbe3fff83)

src/engine/datafacade/mmap_memory_allocator.cpp
```diff
@@ -16,47 +17,52 @@ namespace engine
...
+        // Now, we add one more AllocatedRegion, with it's start address as the start
+        // of the rtree_filename string we've saved.  In the fake_layout, we've
+        // stated that the data is at offset 0, which is where the string starts
+        // at it's own memory address.
+        // The syntax &(rtree_filename[0]) gets the memory address of the first char.
+        // We can't use the convenient `.data()` or `.c_str()` methods, because
+        // prior to C++17 (which we're not using), those return a `const char *`,
+        // which isn't compatible with the `char *` that AllocatedRegion expects
+        // for it's memory_ptr
+        allocated_regions.push_back({&(rtree_filename[0]), std::move(fake_layout)});
```
daniel-j-h: Why does AllocatedRegion expect a `char *` then in the first place?

src/engine/datafacade/mmap_memory_allocator.cpp
```diff
@@ -16,47 +17,52 @@ namespace engine
...
     }
-    else
-    {
-        mapped_memory = util::mmapFile<char>(memory_file, mapped_memory_file);
 
-        storage::DataLayout layout;
-        storage::io::BufferReader reader(mapped_memory.data(), mapped_memory.size());
-        storage::serialization::read(reader, layout);
-        auto layout_size = reader.GetPosition();
+    std::vector<std::pair<bool, boost::filesystem::path>> files = storage.GetStaticFiles();
```
daniel-j-h: Can be auto files?

src/engine/datafacade/mmap_memory_allocator.cpp
```diff
@@ -16,47 +17,52 @@ namespace engine
...
+    files.insert(files.end(), updatable_files.begin(), updatable_files.end());
 
-        index = storage::SharedDataIndex({{mapped_memory.data() + layout_size, std::move(layout)}});
+    for (const auto &file : files)
+    {
+        if (boost::filesystem::exists(file.second))
+        {
+            std::unique_ptr<storage::BaseDataLayout> layout =
+                std::make_unique<storage::TarDataLayout>();
+            boost::iostreams::mapped_file mapped_memory_file;
```
daniel-j-h: Quick q: what's the difference between iostream's mmaped file and interprocess's mapped region? And why are we using the iostream's one?

src/engine/datafacade/process_memory_allocator.cpp
```diff
@@ -15,20 +15,26 @@ ProcessMemoryAllocator::ProcessMemoryAllocator(const storage::StorageConfig &con
...
+    std::vector<storage::SharedDataIndex::AllocatedRegion> regions;
+    regions.push_back({internal_memory.get(), std::move(layout)});
+    index = {std::move(regions)};
 
     storage.PopulateStaticData(index);
     storage.PopulateUpdatableData(index);
 }
 
-ProcessMemoryAllocator::~ProcessMemoryAllocator() {}
+ProcessMemoryAllocator::~ProcessMemoryAllocator() { /* free(internal_memory) */}
```
daniel-j-h: I don't understand this - why is it commented out? What is this supposed to do? Do we need the empty dtor at all?

src/storage/storage.cpp
```diff
@@ -44,32 +44,15 @@ namespace
...
-
 struct RegionHandle
 {
     std::unique_ptr<SharedMemory> memory;
     char *data_ptr;
     std::uint16_t shm_key;
 };
 
-auto setupRegion(SharedRegionRegister &shared_register, const DataLayout &layout)
+auto setupRegion(SharedRegionRegister &shared_register,
```
daniel-j-h: Return type is bool? Maybe make explicit.

src/storage/storage.cpp
```diff
@@ -44,32 +44,15 @@ namespace
...
+                 const std::unique_ptr<storage::BaseDataLayout> &layout)
```
daniel-j-h: unique ptr problem

src/storage/storage.cpp
```diff
@@ -182,10 +165,29 @@ bool swapData(Monitor &monitor,
 
     return true;
 }
-}
+} // namespace
 
 Storage::Storage(StorageConfig config_) : config(std::move(config_)) {}
 
+void Storage::readBlocks(const boost::filesystem::path &path,
```
daniel-j-h: Looks like this is a helper function? Is it supposed to be in the anonymous namespace, too, then?

src/storage/storage.cpp
```diff
@@ -310,53 +306,74 @@ void Storage::PopulateStaticLayout(DataLayout &static_layout)
...
+}
+
+std::string Storage::PopulateLayoutWithRTree(std::unique_ptr<storage::BaseDataLayout> &layout)
+{
+    // Figure out the path to the rtree file (it's not a tar file)
+    auto absolute_file_index_path = boost::filesystem::absolute(config.GetPath(".osrm.fileIndex"));
+
+    // Convert the boost::filesystem::path object into a plain string
+    // that can then be stored as a member of an allocator object
+    std::string rtree_filename = absolute_file_index_path.string();
```
daniel-j-h: auto?

include/storage/serialization.hpp
ghoshkaj: danpat explained -- it's undefined behaviour. got it. 😄

danpat: Done

include/storage/serialization.hpp
danpat: Good idea - done.

include/storage/shared_datatype.hpp
danpat: Funnily enough no - there's calling code that depends on `std::for_each(blocks.begin(), blocks.end())` iterating over the keys in the same order every time.

I feel like the concept of this class has been somewhat damaged by introducing mmapping, and a refactoring is in order.

include/storage/serialization.hpp
daniel-j-h: http://catb.org/jargon/html/N/nasal-demons.html :P

include/util/mmap_file.hpp
danpat: In this case, it's one of:

  - `boost::iostream::mapped_file`
  - `boost::iostream::mapped_file_source`

The only change made in this PR is to improve the naming of the template type - a larger refactoring is probably in order at some point, but not critical for this PR.

src/engine/datafacade/mmap_memory_allocator.cpp
danpat: Historical reasons I believe - I've fixed and changed the stored base pointers to `void *`

src/engine/datafacade/process_memory_allocator.cpp
danpat: Removed

src/storage/storage.cpp
danpat: Fixed - it's used from outside this class.  I've made it a free function in the `osrm::storage` namespace.

include/storage/serialization.hpp
danpat: Range checking added.

include/storage/shared_data_index.hpp
danpat: Checks added

include/storage/shared_datatype.hpp
danpat: `char *` is old code - I've moved everything to `void *` which is more appropriate.

src/storage/storage.cpp
danpat: Done

PR #5251, opened 2018-10-29
Pre-calculate distance values

danpat:
After a long and sordid history, we added `annotations=distances` to the `table` plugin back in https://github.com/Project-OSRM/osrm-backend/pull/4990 and https://github.com/Project-OSRM/osrm-backend/pull/5019

These PRs implemented on-the-fly distance calculations - the table results are obtained, then the paths are unpacked to calculate the distances.

The upside to this is that it costs nothing in terms of resources (no extra preprocessing, no extra RAM) if you don't use this feature.

The downside is that it's a bit slow.  With CH, large matrices that take 1-2 seconds to complete returning just durations can take upwards of 60 seconds to complete when distances are included.

@niemeier-PSI submitted https://github.com/Project-OSRM/osrm-backend/pull/2764 way back which took the other approach - pre-calculate edge distances, and store them as an additional edge property.  This approach is nice and fast, but has the downside that it requires quite a bit more RAM.

This PR is a re-implementation of #2764 - and I will open a follow-up ticket to make this data optional.

## Tasklist

 - [X] add tests (see [testing documentation](https://github.com/Project-OSRM/osrm-backend/blob/master/docs/testing.md)
 - [ ] review
 - [ ] adjust for comments


include/extractor/files.hpp
```diff
@@ -462,6 +462,34 @@ void readEdgeBasedNodeWeights(const boost::filesystem::path &path, NodeWeigtsVec
     storage::serialization::read(reader, "/extractor/edge_based_node_weights", weights);
 }
 
+template <typename NodeWeigtsVectorT, typename NodeDurationsVectorT, typename NodeDistancesVectorT>
```
danpaz: typo `NodeWeigtsVectorT`, although not new to this PR

src/engine/routing_algorithms/many_to_many_ch.cpp
```diff
@@ -147,191 +160,36 @@ void backwardRoutingStep(const DataFacade<Algorithm> &facade,
...
 std::pair<std::vector<EdgeDuration>, std::vector<EdgeDistance>>
 manyToManySearch(SearchEngineData<ch::Algorithm> &engine_working_data,
                  const DataFacade<ch::Algorithm> &facade,
                  const std::vector<PhantomNode> &phantom_nodes,
                  const std::vector<std::size_t> &source_indices,
                  const std::vector<std::size_t> &target_indices,
-                 const bool calculate_distance,
-                 const bool calculate_duration)
+                 const bool /* calculate_distance */,
+                 const bool /* calculate_duration */)
```
danpaz: Can these comments be removed?

src/engine/routing_algorithms/many_to_many_mld.cpp
```diff
@@ -892,48 +626,29 @@ manyToManySearch(SearchEngineData<mld::Algorithm> &engine_working_data,
                  const std::vector<PhantomNode> &phantom_nodes,
                  const std::vector<std::size_t> &source_indices,
                  const std::vector<std::size_t> &target_indices,
-                 const bool calculate_distance,
-                 const bool calculate_duration)
+                 const bool /* calculate_distance */,
+                 const bool /* calculate_duration */)
```
danpaz: Can these comments be removed?

src/engine/routing_algorithms/many_to_many_ch.cpp
ghoshkaj: Oh good point. Yes these can be removed. The[ API code takes care of deciding whether to send back the distance or duration tables through node](https://github.com/Project-OSRM/osrm-backend/blob/82b5648c97edf1d2edec7aecebc35aa8a8033c82/include/engine/api/table_api.hpp#L74), so we don't need these flags here at all - both will always be computed. But only the ones requested will be returned int he response. Phew!

PR #5256, opened 2018-11-01
Save node ids in LegGeometry and inject them into the map matching re…

yairlempert:


yairplt:
Hi, thank you for the insight. I still get the node ids and what we need are node ids plus the node coordinates.
Am I missing something?

danpat:
@yairplt If you do:

`/route/v1/driving/<your coordinates>?geometry=geojson&overview=full&annotations=nodes`, then the `overview` geometry will map 1:1 with the `nodes` annotation.

The data is not perfectly together in the response, but linking them together is trivial.

roiaddi:
The result matches 1:1 the waypoints, not the nodes
Any reason not to enable this @danpat?

danpat:
@roiaddi If you use `annotations=nodes` in combination with `overview=full&geometries=geojson`, then your response will contain these 1:1 mappings between coordinates and nodes:

screen shot 2018-11-02 at 8 36 16 am

Note that the first/last node value in `annotations=nodes` are the values of the nodes at the end of the snapped segment, and don't align with the first/last coordinate, unless you perfectly drop your input coordinate at the node location.

If your route has multiple legs, then you will need to do some array concatenation on the client side to join these arrays together, but it's fairly straightforward.

From what I can tell, this PR

  1. Makes it more convenient to link nodes/coordinates together (+1)
  2. Duplicates already existing information (-1)
  3. Fetches the coordinate of the node before the snapped location at the start, and after the snapped location at the end.  This is +1/-1.

Particularly on point (3), what use-case does it solve for you, that isn't solved only by having the node ID and the snapped coordinate location?

yairplt:
Thanks, I see now how it's working in action. I created a multi leg matching route and the middle nodes were identical with the geometry coordinates as you explained. We still have a problem with the last coordinate (I guess we can ignore first one). Given that the response does add a bit of information which as far as I understand is not available at the moment (first, last nodes) and that join coordinates + nodes together is easier to work with, could you consider allowing this or similar functionality?(I can add an extra api flag or change however you think is best)

DennisOSRM:

PR #5258, opened 2018-11-02
Make returning waypoints in API response optional

danpat:
# Issue

Sometimes, there's some data in the response that you just don't need.

This PR adds a new parameter `return_waypoints=true|false` that allows you to suppress the `waypoints` structure in the various plugins.

It's probably not very useful for all of them, but it can significantly reduce the size of very large `table` requests when there are lots of coordinates being used, and you're not interested in the data in the `waypoints` property.

## Tasklist

 - [ ] review
 - [ ] adjust for comments

## Requirements / Relations


danpat:
Not going to merge this one for now, will dig it up again if we decide it's needed.

PR #5298, opened 2018-12-08
Add a multiplier to the matrix

ghoshkaj:
# Issue

Route optimization solvers often require custom inputs when feeding in the matrix. This is a simple dumb multiplier of table duration values.

## Tasklist

 - [x] review
 - [x] adjust for comments

## Requirements / Relations

### DIFF
--- a/include/engine/api/table_parameters.hpp
+++ b/include/engine/api/table_parameters.hpp
@@ -140,9 +140,6 @@ struct TableParameters : public BaseParameters
         if (fallback_speed < 0)
             return false;
 
-        if (scale_factor < 1)
-            return false;
-
         return true;
     }
 };
--- a/include/engine/api/table_parameters.hpp
+++ b/include/engine/api/table_parameters.hpp
@@ -79,6 +79,8 @@ struct TableParameters : public BaseParameters
 
     AnnotationsType annotations = AnnotationsType::Duration;
 
+    double scale_factor = 1;
+
     TableParameters() = default;
     template <typename... Args>
     TableParameters(std::vector<std::size_t> sources_,
@@ -105,10 +107,13 @@ struct TableParameters : public BaseParameters
                     const AnnotationsType annotations_,
                     double fallback_speed_,
                     FallbackCoordinateType fallback_coordinate_type_,
+                    double scale_factor_,
                     Args... args_)
         : BaseParameters{std::forward<Args>(args_)...}, sources{std::move(sources_)},
           destinations{std::move(destinations_)}, fallback_speed{fallback_speed_},
-          fallback_coordinate_type{fallback_coordinate_type_}, annotations{annotations_}
+          fallback_coordinate_type{fallback_coordinate_type_}, annotations{annotations_},
+          scale_factor{scale_factor_}
+
     {
     }
 
@@ -133,6 +138,9 @@ struct TableParameters : public BaseParameters
             return false;
 
         if (fallback_speed < 0)
+            return false;
+
+        if (scale_factor < 1)
             return false;
 
         return true;
--- a/include/engine/api/table_parameters.hpp
+++ b/include/engine/api/table_parameters.hpp
@@ -140,6 +140,9 @@ struct TableParameters : public BaseParameters
         if (fallback_speed < 0)
             return false;
 
+        if (scale_factor <= 0)
+            return false;
+
         return true;
     }
 };
--- a/include/nodejs/node_osrm_support.hpp
+++ b/include/nodejs/node_osrm_support.hpp
@@ -1238,6 +1238,11 @@ argumentsToTableParameter(const Nan::FunctionCallbackInfo<v8::Value> &args,
             Nan::ThrowError("scale_factor must be a number");
             return table_parameters_ptr();
         }
+        else if (scale_factor->NumberValue() < 0)
+        {
+            Nan::ThrowError("scale_factor must be > 0");
+            return table_parameters_ptr();
+        }
 
         params->scale_factor = static_cast<double>(scale_factor->NumberValue());
     }
--- a/include/nodejs/node_osrm_support.hpp
+++ b/include/nodejs/node_osrm_support.hpp
@@ -1229,6 +1229,24 @@ argumentsToTableParameter(const Nan::FunctionCallbackInfo<v8::Value> &args,
         }
     }
 
+    if (obj->Has(Nan::New("scale_factor").ToLocalChecked()))
+    {
+        auto scale_factor = obj->Get(Nan::New("scale_factor").ToLocalChecked());
+
+        if (!scale_factor->IsNumber())
+        {
+            Nan::ThrowError("scale_factor must be a number");
+            return table_parameters_ptr();
+        }
+        else if (scale_factor->NumberValue() < 1)
+        {
+            Nan::ThrowError("scale_factor must be > 1");
+            return table_parameters_ptr();
+        }
+
+        params->scale_factor = static_cast<double>(scale_factor->NumberValue());
+    }
+
     return params;
 }
 
--- a/include/nodejs/node_osrm_support.hpp
+++ b/include/nodejs/node_osrm_support.hpp
@@ -1238,7 +1238,7 @@ argumentsToTableParameter(const Nan::FunctionCallbackInfo<v8::Value> &args,
             Nan::ThrowError("scale_factor must be a number");
             return table_parameters_ptr();
         }
-        else if (scale_factor->NumberValue() < 0)
+        else if (scale_factor->NumberValue() <= 0)
         {
             Nan::ThrowError("scale_factor must be > 0");
             return table_parameters_ptr();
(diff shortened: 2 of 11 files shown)


src/engine/plugins/table.cpp
```diff
@@ -127,6 +128,12 @@ Status TablePlugin::HandleRequest(const RoutingAlgorithmsInterface &algorithms,
                         result_tables_pair.second[table_index] = distance_estimate;
                     }
                 }
+                if (params.scale_factor > 1.0 &&
+                    result_tables_pair.first[table_index] != MAXIMAL_EDGE_DURATION)
```
danpaz: Hmm should scale_factor also scale the fallback speed result?

ghoshkaj: The usecase for fallback_speed is a guestimate when there's no route. I'd imagine the solver would want realistic estimates for even the fallback to make realistic routes.

include/engine/api/table_parameters.hpp
```diff
@@ -135,6 +140,9 @@ struct TableParameters : public BaseParameters
         if (fallback_speed < 0)
             return false;
 
+        if (scale_factor < 1)
```
danpat: Docs say > 0, but code enforces > 1.

I think any value should be allowed - `scale_factor=-0.5` is weird, but possible to imagine.

include/nodejs/node_osrm_support.hpp
```diff
@@ -1229,6 +1229,24 @@ argumentsToTableParameter(const Nan::FunctionCallbackInfo<v8::Value> &args,
...
+    if (obj->Has(Nan::New("scale_factor").ToLocalChecked()))
+    {
+        auto scale_factor = obj->Get(Nan::New("scale_factor").ToLocalChecked());
+
+        if (!scale_factor->IsNumber())
+        {
+            Nan::ThrowError("scale_factor must be a number");
+            return table_parameters_ptr();
+        }
+        else if (scale_factor->NumberValue() < 1)
```
danpat: Same again - docs say > 0, code enforces >= 1 (not the same as above btw, which does > 1, not >=).

Should allow any number I think, no need to restrict this, it doesn't trigger any computation growth if it's large.

src/engine/plugins/table.cpp
```diff
@@ -96,15 +96,16 @@ Status TablePlugin::HandleRequest(const RoutingAlgorithmsInterface &algorithms,
     }
 
     // Scan table for null results - if any exist, replace with distance estimates
-    if (params.fallback_speed > 0)
+    if (params.fallback_speed > 0 || params.scale_factor > 1.0)
```
danpat: Should be `!= 1`

src/engine/plugins/table.cpp
```diff
@@ -127,6 +128,12 @@ Status TablePlugin::HandleRequest(const RoutingAlgorithmsInterface &algorithms,
                         result_tables_pair.second[table_index] = distance_estimate;
                     }
                 }
+                if (params.scale_factor > 1.0 &&
+                    result_tables_pair.first[table_index] != MAXIMAL_EDGE_DURATION)
+                {
+                    result_tables_pair.first[table_index] =
+                        result_tables_pair.first[table_index] * (double)params.scale_factor;
```
danpat: Should probably check for overflow here, and clamp the result to some very large number. - if value * scale_factor ends up > MAXIMAL_EDGE_DURATION (i.e. overflows), we should set `value` to `MAXIMAL_EDGE_DURATION-1`.

The value of `MAXIMAL_EDGE_DURATION` is reserved to indicate "noroute", so we have to clamp slightly lower than that.

Something like:

```
double result = result_tables_pair.first[table_index] * params.scale_factor;
if (double > MAXIMAL_EDGE_DURATION) {
  result_tables_pair.first[table_index] = MAXIMAL_EDGE_DURATION - 1;
} else {
  result_tables_pair.first[table_index] = result;
}
```

danpat: Also note - no need to cast `params.scale_factor` to `double` - it's already one of those.  Also, it's preferred in C++ to use `static_cast<double>(thingy)` instead of C-style `(double)thing` casts - the C++ style has better defined behaviour, which makes it easier to debug if something goes wrong.

src/engine/plugins/table.cpp
ghoshkaj: Yes fallback speeds should mirror real speeds. Scaling one and not the other is questionable when feeding into a optimzation solver.

include/engine/api/table_parameters.hpp
ghoshkaj: So we shouldn't really enforce anything here?

src/engine/plugins/table.cpp
ghoshkaj: 👍

ghoshkaj: @danpat after applying a clamp, should we return an error? Or return the raw really large value? For instance in this test case below, the really large unexplained number gets piped through. 

Also, we need a clamp for a negative multiplier too, yes?

```
 ✔ Given the profile "testbot"
  ✔ And the partition extra arguments "--small-component-size 1 --max-cell-sizes 2,4,8,16"
  ✔ Given the query options
      | scale_factor | 2147483647 |
  ✔ Given the node map
      """
      a b
      """
  ✔ And the ways
      | nodes |
      | ab    |
  ✖ When I request a travel time matrix I should get
      |   | a  | b  |
      | a | 0  | 20 |
      | b | 20 | 0  |

Failures:

1) Scenario: Testbot - Travel time matrix of minimal network with scale factor - features/testbot/duration_matrix.feature:611
   Step: When I request a travel time matrix I should get - features/testbot/duration_matrix.feature:624
   Step Definition: features/step_definitions/distance_matrix.js:10
   Message:
     Tables were not identical:
     |   |     a  |     b  |
     | a |     0  | (-) 20 |
     | a |     0  | (+) 214748364.6 |
     | b | (-) 20 |     0  |
     | b | (+) 214748364.6 |     0  |
```

include/engine/api/table_parameters.hpp
danpat: Yeah - I can’t think of a reason to restrict this, who knows how people might want to use it.

jcoupey:
Is there a reason to forbid `scale_factor` values under 1? Being able to shorten durations could also be useful.

ghoshkaj:
@jcoupey there is not a reason. I'm removing the greater than 1 restriction now.

src/engine/plugins/table.cpp
ghoshkaj: Hm I think we should clamp down the `scale_factor` itself. Try to clamp the result down after an overflow has occured is too late, and to check every element before doing the multiplication to prevent an overflow will be expensive. 

This is the kind of check that would happen before every multiplication:
```
EdgeDuration diff = MAXIMAL_EDGE_DURATION / std::lround(result_tables_pair.first[table_index]);
       if (params.scale_factor > diff) {
            // overflow will happen for std::lround(result_tables_pair.first[table_index]) * params.scale_factor so error
       }
result_tables_pair.first[table_index] =  std::lround(result_tables_pair.first[table_index] * params.scale_factor);
```

jcoupey:
@ghoshkaj I think you want to keep a `parameters.scale_factor > 0` check in `table_service.cpp`.

ghoshkaj:
> @ghoshkaj I think you want to keep a `parameters.scale_factor > 0` check in table_service.cpp.
@jcoupey I'm coming to same conclusion as I'm trying to figure out checking for overflows. Doing two multiplications for each overflow is quite expensive and I don't think a negative durations matrix is useful. 

So @danpat I'm going with @jcoupey's suggestion of clamping `parameters.scale_factor > 0`.

include/engine/api/table_parameters.hpp
ghoshkaj: @danpat I'm actually going to clamp to >0.  See https://github.com/Project-OSRM/osrm-backend/pull/5298#discussion_r240010511 for overflow checks and https://github.com/Project-OSRM/osrm-backend/pull/5298#issuecomment-445477266 for why doubling the number of overflow checks would trigger computation growth.

src/engine/plugins/table.cpp
ghoshkaj: I've gone with `return the raw really large value`.

ghoshkaj:
Okay I ran a number of requests with 1280 coordinates so 1280*1280 elements. It took about 3 seconds. I don't see any significant user of resources in the TableRequest Handler with coming from the division or the rounding.

I've confirmed that the values are being properly scaled.

screen shot 2018-12-08 at 2 30 10 pm

ghoshkaj:
Without `scale_factor` the same request takes 2.98 s.
screen shot 2018-12-08 at 2 39 59 pm

src/server/service/table_service.cpp
```diff
@@ -61,6 +61,11 @@ std::string getWrongOptionHelp(const engine::api::TableParameters &parameters)
         help = "fallback_speed must be > 0";
     }
 
+    if (parameters.scale_factor < 0)
```
danpat: This tests for < 0 but the error message says it must be > 0.

The `if` should probably be `<=` to match the message.

danpat:
@ghoshkaj So we're failing on coverage.

One more request: add some tests that exercise the failure conditions - `scale_factor=0` and `scale_factor=-1`.  This should increase coverage, *and* verify that our limit rules on the parameter are correct.

PR #5323, opened 2018-12-19
Add libc6 to Dockerfile in run stage

AliKarami:
we found that in v5.21.0 there is missing `GLIBC_2.28` requirement and pod crashes. so we add `libc6` package which solves the issue.

# Issue

fix crashes caused by missing `GLIBC_2.28` requirement.

## Tasklist

 - [ ] review
 - [ ] adjust for comments

## Requirements / Relations
-

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -33,7 +33,7 @@ RUN NPROC=$(grep -c ^processor /proc/cpuinfo 2>/dev/null || 1) && \
 FROM debian:buster-slim as runstage
 RUN mkdir -p /src  && mkdir -p /opt
 RUN apt-get update && \
-apt-get install -y --no-install-recommends libboost-program-options1.67.0 libboost-regex1.67.0 \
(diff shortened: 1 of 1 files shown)


amirmasud:
As @AliKarami said, the problem seems to be caused by `gcc` package dependencies where it uses `libc6-dev` as mentioned [here](https://packages.debian.org/buster/gcc) and on run-stage you didn't add `libc6` and the `libc6` version wasn't the same. so by adding the package as @AliKarami did, the problem should be solved.

danpat:
Ugh.  I missed that the `Dockerfile` depends on `buster`, which is still `testing` - dependencies may still change over time.  It might be preferable to switch to `stretch` which is considered stable.

We should also add some tests for this dockerfile to catch this stuff....I'll look into that.

danpat:
I decided to avoid the currently unstable state of `buster` and go with the downgrade to `stretch-slim` over in https://github.com/Project-OSRM/osrm-backend/pull/5338

PR #5332, opened 2019-01-03
WIP: Conditional behaviours with CH graphs

danpat:
# Issue

This PR aims to make some weighting options run-time adjustable on the CH graph.

The basic principle here is that if it's *ever* possible that a shortest path goes through a node, then shortcut edges should be inserted during CH construction.  If shortcut edges are available, then it's possible to adjust weights at query time.  This only works for situations where weights are *increased*, but fortunately, there are several useful scenarios where that's exactly what happens.

It also assumes that the edges that behave this way are fairly sparse, and the overall hierarchy of the graph isn't significantly changed by modifying their weights.

Goals:
  - disable private access penalties when map-matching, but not when routing
  - make private access penalties configurable at runtime with a URL parameter
  - implement adjustable `avoid=highway|toll|ferry` behaviour in CH with adjustable weighting

Long term:
  - Implement time-dependent conditional access `access:conditional=Tue` using this feature (https://wiki.openstreetmap.org/wiki/Conditional_restrictions)

## Tasklist

 - [ ] review
 - [ ] adjust for comments

## Requirements / Relations


TheMarex:
One wrinkle here is that each shortcut also needs to be marked as containing a restricted node (and each shortcut of a shortcut that contains a restricted node). I'm kind of curious what would happen with toll roads here, that would probably break the hierarchy quite a bit.

I've been wondering if it is possible to compute a CH that is valid and somewhat good for more than one metrics for a while. In that scenario an additional shortcut would be marked as "unnecessary" if e.g. tolls don't need to be avoided which could help to curtail the performance impact a bit. The node priority probably needs to be adjusted to balancing performance for the metrics. 

I'm not sure how this would shape up against the current approach of just keeping these things in the core and contracting that multiple times. Could be combined though to limit the insertion of a lot shortcuts.

DennisOSRM:

PR #5337, opened 2019-01-11
Add slow car profile

adityaseth09:
Experimenting with profiles


profiles/bicycle.lua
```diff
@@ -38,20 +39,10 @@ function setup()
...
-      'cattle_grid',
-      'border_control',
-      'toll_booth',
-      'sally_port',
-      'gate',
-      'lift_gate',
-      'no',
-      'block'
+    barrier_blacklist = Set {
+      'yes',
```
emiltin: why do you have 'yes' in the blacklist?

Issue #5115, opened 2018-06-21
Add graph version information to response


gardster:
I encountered such a problem while using OSRM. I have an OSRM cluster behind a load balancer that processes a continuous stream of requests. To update the road graph, I'm restarting servers one by one. At some point, several servers running with the old graph and some servers with the new one. As far as I understand, there is no way to find out on which version of the graph it was created from the OSRM answer. I suggest to add such field into the response. This can be a data timestamp from the pbf file, or the value of some parameter passed on the stage of osrm-extract.

gardster:
Sure. 
For example I have a "toll roads expenses" service. I have a matched tracks from the OSRM and some server with information about toll pricing. This server looks for toll roads in OSRM answer and returns prices if it can find toll boothes. So both services have geography information from the OSM (geometry, osm_ids and so on) and ideally need to deploy simultaneously. When someone make a vast graph modification and OSRMs answer has a different geometry than pricing server supposed it can lead to mistakes when the old pricing server tries to find objects on track. So we need additional information to be sure that all services works on same version of geography information.

TheMarex:
Ah interesting. So we did have something similar not too long ago but removed it because we did not have a use-case for it. :sweat_smile: 

To add this feature I would recommend the following steps:

- Get the timestamp from [here](https://github.com/Project-OSRM/osrm-backend/blob/master/src/extractor/extractor.cpp#L425)
- Add a new functions to `writeTimestamp`/`readTimestamp` to https://github.com/Project-OSRM/osrm-backend/blob/master/include/extractor/files.hpp and pass the timestamp there
- Add an entry `.osrm.timestamp` [here](https://github.com/Project-OSRM/osrm-backend/blob/master/src/storage/storage.cpp#L304)
- Create a `create_timestmap_view` function [here](https://github.com/Project-OSRM/osrm-backend/blob/master/include/storage/view_factory.hpp) that uses `index.GetBlockPtr<T>(name)` to get a shared memory region for the timestamp
- Call `create_timestamp_view` and `readTimestamp` from [somewhere here](https://github.com/Project-OSRM/osrm-backend/blob/master/src/storage/storage.cpp#L382)
- Extend the interface [BaseDatafacade](https://github.com/Project-OSRM/osrm-backend/blob/master/include/engine/datafacade/datafacade_base.hpp#L51) for a new method `GetTimestamp`
[...]

PR #5341, opened 2019-01-15
Data timestamp version in response

gardster:
# Issue

[5115](https://github.com/Project-OSRM/osrm-backend/issues/5115)

### DIFF
--- a/include/engine/api/route_api.hpp
+++ b/include/engine/api/route_api.hpp
@@ -71,7 +71,7 @@ class RouteAPI : public BaseAPI
         auto data_timestamp = facade.GetTimestamp();
         if (!data_timestamp.empty())
         {
-            response.values["data_version"] = facade.GetTimestamp();
+            response.values["data_version"] = data_timestamp;
         }
     }
 
--- a/include/engine/api/route_api.hpp
+++ b/include/engine/api/route_api.hpp
@@ -68,7 +68,11 @@ class RouteAPI : public BaseAPI
         response.values["waypoints"] = BaseAPI::MakeWaypoints(all_start_end_points);
         response.values["routes"] = std::move(jsRoutes);
         response.values["code"] = "Ok";
-        response.values["data_version"] = facade.GetTimestamp();
+        auto data_timestamp = facade.GetTimestamp();
+        if (!data_timestamp.empty())
+        {
+            response.values["data_version"] = facade.GetTimestamp();
+        }
     }
 
   protected:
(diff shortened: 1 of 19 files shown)


gardster:
Hi! Can anyone review my PR, please?
@TheMarex

danpat:
@gardster Overall, I think this idea is useful, but before merging it, I think it needs some changes:

  1. Make the `data_version` field optional in the response - don't return a fixed `n/a` value if it's missing, instead exclude the field altogether.
  2. Add a command-line flag to `osrm-extract` called something like `--data-version X`, where X can be one of:
    - `none` - would not save the version, and will cause API response to have no `data_version` field
    - `osmosis_replication_timestamp` - will extract the Osmosis timestamp from the OSM file - I think it would be reasonable to make this the default
    - `custom:<string>` - allow the user to supply a string on the command-line that'll get saved

The reason for these changes - I don't think users should be required to expose their data version to users if they don't want to.  Here at Mapbox, we don't even guarantee that we're exposing a single version of OSM data - we validate changesets individually, so some changes may get delayed while others come through.  Until we figure out what `data_version` would mean in that case, we wouldn't want `n/a` in every API response - it's not useful information (this is why I suggest the `custom:<string>` option above).

There's also the added complication of traffic updates.  It's possible to run `osrm-extract` once, then `osrm-customize` or `osrm-contract` multiple times with updated traffic data.  In this case, `data_version` wouldn't reflect all the information you might want to know about the data.  I'm not sure whether we'd want a second field for that, or whether `osrm-contract/osrm-customize` should be able to modify `data_version` in some way....

gardster:
Thank you for the comment.
I will implement this functionality.

gardster:
PTAL

include/engine/api/route_api.hpp
```diff
@@ -68,6 +68,11 @@ class RouteAPI : public BaseAPI
         response.values["waypoints"] = BaseAPI::MakeWaypoints(all_start_end_points);
         response.values["routes"] = std::move(jsRoutes);
         response.values["code"] = "Ok";
+        auto data_timestamp = facade.GetTimestamp();
+        if (!data_timestamp.empty())
+        {
+            response.values["data_version"] = facade.GetTimestamp();
```
danpat: This could just be `response.values["data_version"] = data_timestamp;`, right?

danpat:
@gardster If you could add a couple of tests, I think this is good to merge 👍

include/engine/api/route_api.hpp
gardster: Sure

gardster:
PTAL

danpat:
@gardster Merged with https://github.com/Project-OSRM/osrm-backend/commit/8b45ff7a186d158345304169768cd7d99acb8126 thanks for the feature!

PR #5342, opened 2019-01-15
Fix cpack config by pointing to a correct LICENSE file

Antti:
In `cmake/CPackConfig.cmake` we set `CPACK_RESOURCE_FILE_LICENSE` to be `LICENCE.TXT`, while this file is actually called `LICENSE.TXT`

* Fix `CPACK_RESOURCE_FILE_LICENSE` to point to a correct license file.


danpat:
Thanks!

PR #5345, opened 2019-01-23
Add waypoints parameter to viaroute API

danpat:
# Issue

This adds a `waypoints=` parameter to the `viaroute` plugin, similar to the one on the `match` plugin.

What does it do?  In a nutshell, it lets you suppress depart/arrive instructions for some waypoints.  This is helpful if you want to provide some coordinates to guide the path the route takes, but don't want them to be announced as waypoints.

## Tasklist

 - [ ] review
 - [ ] adjust for comments

## Requirements / Relations

### DIFF
--- a/features/testbot/via.feature
+++ b/features/testbot/via.feature
@@ -18,6 +18,23 @@ Feature: Via points
             | waypoints | route           |
             | a,b,c     | abc,abc,abc,abc |
 
+    Scenario: Simple via point with waypoints collapsing
+        Given the node map
+            """
+            a b c
+            """
+
+        And the ways
+            | nodes |
+            | abc   |
+
+       Given the query options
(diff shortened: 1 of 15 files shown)


features/testbot/via.feature
```diff
@@ -18,6 +18,23 @@ Feature: Via points
...
+        And the ways
+            | nodes |
+            | abc   |
+
+       Given the query options
+            | waypoints | 0;2   |
+
+        When I route I should get
+            | waypoints | route       |
+            | a,b,c     | abc,abc     |
```
1ec5: Let’s also test the case where the waypoint is located at an intersection and the route ends up making a non-obvious turn at that intersection; we’d expect the route to contain a maneuver at that point, rather than eliding it completely.

Issue #5353, opened 2019-02-05
Consider acceleration profiles for short trips


danpat:
When you request routes between two very close locations, we currently assume you'll be travelling at full speed all the way.

This leads us to under-estimate travel times for very short trips.  Imagine feeding a mix of close-by coordinates, and medium-distancey coordinates into a `/table` API call and then sending the results to an route optimizer - the results can be less than ideal, depending on the spread.

For car routing, we should consider the effects of acceleration/deceleration.  For longer journeys the effect is minimal, but for short trips, it can dominate the travel time.  I think we need two things:

  1.  A way to specify the acceleration/braking curves either as a URL parameter, or as part of the `car.lua` profile
  2. A way to decide whether all waypoints should be subject to the acceleration/braking penalties (e.g. waypoints where you stop vs ones where you don't)

We should also consider how this works with all plugins - i.e. I don't think acceleration should be considered for map matching, by default.

As a secondary feature, we could also consider acceleration when calculating turn costs.

PR #5354, opened 2019-02-07
[WIP] add stoppage penalties to table_params and use to penalize table results

kevinkreiser:
# Issue

This PR is meant as a step toward addressing #5353 

The gist is that I've added a place for new table parameters that allow you to add a penalty (in terms of time/duration) to each table result (source/target pair) based on how acceleration at the beginning of that trip might happen and how deceleration at the end of the trip might happen.

The additional parameters are a minimum and maximum penalty in seconds. Basically an upper and lower bound. The penalty will vary with respect to the expected speed of travel on the edge where the source/target was correlated in the graph. The slower the speed of the edge, the lesser the penalty. The idea being that it takes you less time to get up to speed or come to a stop for edges with slower speeds. The inverse should hold true for edges with higher speeds.

For edge speeds lower than a certain cutoff, no penalty is applied. For edge speeds greater than a certain cutoff, the user provided maximum penalty is applied (this probably doesnt make sense for space travel, but should be ok for personal vehicle travel). Speeds in between the two extrema result in a penalty that is a linear combination of the upper and lower user-supplied penalties.

## Tasklist

 - [ ] review
 - [ ] adjust for comments

## Requirements / Relations

### DIFF
--- a/include/util/typedefs.hpp
+++ b/include/util/typedefs.hpp
@@ -48,7 +48,7 @@ struct osm_way_id
 struct duplicated_node
 {
 };
-}
+} // namespace tag
 using OSMNodeID = osrm::Alias<std::uint64_t, tag::osm_node_id>;
 static_assert(std::is_pod<OSMNodeID>(), "OSMNodeID is not a valid alias");
 using OSMWayID = osrm::Alias<std::uint64_t, tag::osm_way_id>;
@@ -117,7 +117,13 @@ static const EdgeDuration MAXIMAL_EDGE_DURATION = std::numeric_limits<EdgeDurati
 static const EdgeDistance MAXIMAL_EDGE_DISTANCE = std::numeric_limits<EdgeDistance>::max();
 static const TurnPenalty INVALID_TURN_PENALTY = std::numeric_limits<TurnPenalty>::max();
 static const EdgeDistance INVALID_EDGE_DISTANCE = std::numeric_limits<EdgeDistance>::max();
-static const EdgeDistance INVALID_FALLBACK_SPEED = std::numeric_limits<double>::max();
+static const EdgeDistance INVALID_FALLBACK_SPEED = std::numeric_limits<EdgeDistance>::max();
+static const EdgeDuration INVALID_MINIMUM_STOPAGE_PENALTY =
+    std::numeric_limits<EdgeDuration>::max();
+static const EdgeDuration INVALID_MAXIMUM_STOPAGE_PENALTY =
+    std::numeric_limits<EdgeDuration>::max();
+constexpr double MINIMAL_ACCEL_DECEL_PENALIZABLE_SPEED = 10;
+constexpr double MAXIMAL_ACCEL_DECEL_PENALIZABLE_SPEED = 40;
 
 // FIXME the bitfields we use require a reduced maximal duration, this should be kept consistent
 // within the code base. For now we have to ensure that we don't case 30 bit to -1 and break any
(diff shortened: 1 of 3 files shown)


src/engine/plugins/table.cpp
```diff
@@ -106,6 +132,30 @@ Status TablePlugin::HandleRequest(const RoutingAlgorithmsInterface &algorithms,
...
+                if (stoppage_penalty_range >= 0 &&
+                    result_tables_pair.first[table_index] != MAXIMAL_EDGE_DURATION)
+                {
+                    const auto &source =
+                        snapped_phantoms[params.sources.empty() ? row : params.sources[row]];
+                    const auto &destination =
+                        snapped_phantoms[params.destinations.empty() ? column
+                                                                     : params.destinations[column]];
+
+                    // TODO: if phantom node is actual node, will the distance/duration be 0?
```
kevinkreiser: can anyone with more experience speak to this question? when the correlated point is close to a graph node, can the distance/duration be really small? do i need to worry about `reverse_distance/duration`?

include/util/typedefs.hpp
```diff
@@ -117,7 +117,13 @@ static const EdgeDuration MAXIMAL_EDGE_DURATION = std::numeric_limits<EdgeDurati
 static const EdgeDistance MAXIMAL_EDGE_DISTANCE = std::numeric_limits<EdgeDistance>::max();
 static const TurnPenalty INVALID_TURN_PENALTY = std::numeric_limits<TurnPenalty>::max();
 static const EdgeDistance INVALID_EDGE_DISTANCE = std::numeric_limits<EdgeDistance>::max();
-static const EdgeDistance INVALID_FALLBACK_SPEED = std::numeric_limits<double>::max();
+static const EdgeDistance INVALID_FALLBACK_SPEED = std::numeric_limits<EdgeDistance>::max();
```
kevinkreiser: wondering why these all aren't just `constexpr`?

danpat: This code predates C++11 and hasn't been updated :-)

src/engine/plugins/table.cpp
```diff
@@ -106,6 +132,30 @@ Status TablePlugin::HandleRequest(const RoutingAlgorithmsInterface &algorithms,
...
+
+                    auto source_penalty =
+                        stoppage_penalty(source.forward_distance / source.forward_duration,
+                                         params.min_stoppage_penalty,
+                                         stoppage_penalty_range);
```
danpat: Oh bother - `forward` here is relative to the edge direction in the graph - and there's no guarantee that the `forward` direction was the one used in the actual route.  I can't think of an easy way to get that information at this location either....

kevinkreiser: i wondered that :smile: 

does it mean that the duration/distance may be 0 depending on which one you pick?

danpat: Yes.  In general I think I'm going to take back my "let's just adjust the table results after the fact" - there's not going to be an easy way to get it right here.  I think it will make more sense to adjust the phantom node durations before routing begins, then we don't have to worry about which one was picked.

kevinkreiser: wont dont really care which direction they went, we only care the speed they went. can you have different durations depending on the direction you travel on the edge? i mean i know in real life its possible (but kind of absurd) but i mean in terms of the data model can we even have different speeds? also is the duration just the portion of the edge that was traveled or is it the full edge? 

seems like one of the directions had to be used, and if we can say an edge has the same speed in both directions we can just pick one to get the speed?

kevinkreiser:
changing the source branch for this work

PR #5357, opened 2019-02-07
Add Acceleration Profiles to Tune Route Duration (especially for short trips)

kevinkreiser:
# Issue

This PR is meant as a step toward addressing #5353

The gist is that we've added a new base parameter that allows you to select an acceleration profile. The idea being that this should increase the accuracy of the estimated duration of a trip by including some amount of penalty for starting/arriving from/at the given waypoints.

The additional parameter is the name of an acceleration profile which selects, at runtime, a profile appropriate for the vehicle being driven. Internally this essentially selects a coefficient of a larger set of equations used to estimate the time it takes to get up to or speed or come to a full stop. In the end this will add time to the routes duration. The penalty will vary with respect to the expected speed of travel on the edge where the waypoint was correlated in the graph. The slower the speed of the edge, the lesser the penalty. The idea being that it takes you less time to get up to speed or come to a stop for edges with slower speeds. The inverse should hold true for edges with higher speeds.

## Tasklist

 - [ ] review
 - [ ] adjust for comments

## Requirements / Relations

### DIFF
--- a/include/util/typedefs.hpp
+++ b/include/util/typedefs.hpp
@@ -119,6 +119,14 @@ static const TurnPenalty INVALID_TURN_PENALTY = std::numeric_limits<TurnPenalty>
 static const EdgeDistance INVALID_EDGE_DISTANCE = std::numeric_limits<EdgeDistance>::max();
 static const EdgeDistance INVALID_FALLBACK_SPEED = std::numeric_limits<double>::max();
 
+// Recommended value for passenger vehicles from
+// https://fdotwww.blob.core.windows.net/sitefinity/docs/default-source/content/rail/publications/studies/safety/accelerationresearch.pdf?sfvrsn=716a4bb1_0
+static const double ACCELERATION_ALPHA_CAR = 6.0;
+static const double ACCELERATION_ALPHA_FAST_CAR = 18;
+static const double ACCELERATION_ALPHA_SLOW_CAR = 2;
+static const double ACCELERATION_ALPHA_TRUCK = 1.5;
+static const double ACCELERATION_ALPHA_TRACTOR_TRAILER = 0.5;
+
 // FIXME the bitfields we use require a reduced maximal duration, this should be kept consistent
 // within the code base. For now we have to ensure that we don't case 30 bit to -1 and break any
 // min() / operator< checks due to the invalid truncation. In addition, using signed and unsigned
--- a/src/engine/plugins/table.cpp
+++ b/src/engine/plugins/table.cpp
@@ -86,8 +86,11 @@ Status TablePlugin::HandleRequest(const RoutingAlgorithmsInterface &algorithms,
     bool request_distance = params.annotations & api::TableParameters::AnnotationsType::Distance;
     bool request_duration = params.annotations & api::TableParameters::AnnotationsType::Duration;
 
+    // Because of the Short Trip ETA adjustments below, we need distances every time
+    const bool distances_required = request_distance || params.waypoint_acceleration_factor > 0.;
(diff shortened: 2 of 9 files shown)


include/engine/guidance/assemble_leg.hpp
```diff
@@ -182,6 +182,13 @@ inline RouteLeg assembleLeg(const datafacade::BaseDataFacade &facade,
         duration = std::max(0, duration);
     }
 
+    // Add start and stop penalties
+    if (distance > 0)
```
kevinkreiser: this is too un-nuanced. we should add a constant distance at which penalties make sense. maybe something like 50m?

include/engine/hint.hpp
```diff
@@ -63,11 +63,11 @@ struct Hint
     friend std::ostream &operator<<(std::ostream &, const Hint &);
 };
 
-static_assert(sizeof(Hint) == 80 + 4, "Hint is bigger than expected");
-constexpr std::size_t ENCODED_HINT_SIZE = 112;
+static_assert(sizeof(Hint) == sizeof(PhantomNode) + 4, "Hint is bigger than expected");
+constexpr std::size_t ENCODED_HINT_SIZE = 124;
```
kevinkreiser: this breaks compatability :frowning_face:

src/engine/plugins/table.cpp
```diff
@@ -98,14 +98,20 @@ Status TablePlugin::HandleRequest(const RoutingAlgorithmsInterface &algorithms,
...
+         params.max_stoppage_penalty != INVALID_MAXIMUM_STOPPAGE_PENALTY))
     {
         for (std::size_t row = 0; row < num_sources; row++)
         {
             for (std::size_t column = 0; column < num_destinations; column++)
             {
                 const auto &table_index = row * num_destinations + column;
                 BOOST_ASSERT(table_index < result_tables_pair.first.size());
+                // Zero out the diagonal
+                if (result_tables_pair.first[table_index] != MAXIMAL_EDGE_DURATION && row == column)
```
kevinkreiser: because each waypoint in the source/targets list is unique we ended up adding the penalty to all of them, which means even the diagonal of the matrix will have non-0 entries if the penalty was enabled. this is kind of messy to reset the diagonal here :frowning_face:

include/util/typedefs.hpp
```diff
@@ -119,6 +119,14 @@ static const TurnPenalty INVALID_TURN_PENALTY = std::numeric_limits<TurnPenalty>
 static const EdgeDistance INVALID_EDGE_DISTANCE = std::numeric_limits<EdgeDistance>::max();
 static const EdgeDistance INVALID_FALLBACK_SPEED = std::numeric_limits<double>::max();
 
+// Recommended value for passenger vehicles from
+// https://fdotwww.blob.core.windows.net/sitefinity/docs/default-source/content/rail/publications/studies/safety/accelerationresearch.pdf?sfvrsn=716a4bb1_0
+static const double ACCELERATION_ALPHA_CAR = 6.0;
```
kevinkreiser: any reason not to allow passing this value as the parameter in the query rather than matching the string to these constants?

kevinkreiser: is it becasue its hard to describe what the value actually means?

danpat: These values are just aliases - you can pass `car`, `fast_car`, etc, as strings in the query, or you can pass in an arbitrary number.  I'd like us to be able to adjust these numbers over time if we learn more about how cars behave, without having to have people change their code.

PR #5362, opened 2019-02-15
Quote command line options which may contain special characters

rjc:
I.e. the example commands fail if `$PWD` contains a space character.


danpat:
Thanks!

rjc:
No worries! :^)

PR #5364, opened 2019-02-15
Use an environment variable instead of command substitution

rjc:
`$PWD` is already there so use it instead of running `pwd(1)` in a subshell each time.


danpat:
Saving the world, 17 clock cycles at a time.  Thanks!

rjc:
:^D

Issue #5370, opened 2019-02-20
Brooklyn Bridge is not routable by default


Zverik:
OSRM by default does not include Brooklyn Bridge in New York to a routing graph. The reason is that it has [2.7 tons weight limit](https://www.openstreetmap.org/way/432550255#map=17/40.70573/-73.99628), while in `car.lua` the default weight is 3.5 tons.

danpat:
Fixed by #5371

PR #5371, opened 2019-02-20
Decrease default car weight to 2 tons

Zverik:
See issue #5370.

Googling shows that the average car weight is between 1.6 and 2 tons. The default 3.5 exceeds even the official limit for heavy cars in USA and Canada, which is 6000 lbs (2.7 tons). With most customer cars weighting under 1.5 tons, I consider 2 tons a good default.

Among other things, this would allow northbound passage through Brooklyn Bridge in New York, which is limited to 2.7 tons. People have not noticed this regression in OSRM 5.19, since most instances run on older versions: e.g. 5.14 for the FOSSGIS server.

References: [1](https://cars.lovetoknow.com/List_of_Car_Weights), [2](https://www.creditdonkey.com/average-weight-car.html), [3](https://www.quora.com/How-much-does-an-average-car-weigh), [4](https://en.wikipedia.org/wiki/Car#Weight).

## Tasklist

 - [x] CHANGELOG.md entry
 - [x] add tests
 - [ ] review
 - [ ] adjust for comments

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@
       - ADDED: new API parameter - `snapping=any|default` to allow snapping to previously unsnappable edges [#5361](https://github.com/Project-OSRM/osrm-backend/pull/5361)
     - Routing:
       - CHANGED: allow routing past `barrier=arch` [#5352](https://github.com/Project-OSRM/osrm-backend/pull/5352)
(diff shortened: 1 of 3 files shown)


danpat:
Nice catch - the demoserver is running a slightly custom version of the car profile, so it wasn't visible there either.

Issue #5380, opened 2019-02-28
Turn restriction not work some time


mostafa-j13:
When we have way with id >2^32 turn restriction is not work on this way

danpat:
Currently, the largest way ID in the public OSM database is about [565140676](https://www.openstreetmap.org/stats/data_stats.html), which fits within 2^30.  We've got a way to go before we run out of space for way IDs.  We added support for >2^32 node IDs back in December 2015, but it'll be quite a while before ways hit the same threshold.

I'm guessing you're using your own dataset?  If so, you should renumber your data so that the way ID values are less than 2^32.

PR #5381, opened 2019-02-28
Update extractor_callbacks.cpp

mostafa-j13:
way_id overflow with 64 bit number

# Issue

https://github.com/Project-OSRM/osrm-backend/issues/5380

### DIFF
--- a/src/extractor/extractor_callbacks.cpp
+++ b/src/extractor/extractor_callbacks.cpp
@@ -478,7 +478,7 @@ void ExtractorCallbacks::ProcessWay(const osmium::Way &input_way, const Extracti
                    });
 
     external_memory.way_start_end_id_list.push_back(
-        {OSMWayID{static_cast<std::uint32_t>(input_way.id())},
+        {OSMWayID{static_cast<std::uint64_t>(input_way.id())},
          OSMNodeID{static_cast<std::uint64_t>(nodes[0].ref())},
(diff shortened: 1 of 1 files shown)


src/extractor/extractor_callbacks.cpp
```diff
@@ -478,7 +478,7 @@ void ExtractorCallbacks::ProcessWay(const osmium::Way &input_way, const Extracti
                    });
 
     external_memory.way_start_end_id_list.push_back(
-        {OSMWayID{static_cast<std::uint32_t>(input_way.id())},
```
mostafa-j13: here this must be uint64_t
 #5380

mostafa-j13: i fixed in #5381

akashihi (approved):

akashihi (changes requested):
I think the issue it with 33bit node ids, not way ids?

wangyoucao577:
> I think the issue it with 33bit node ids, not way ids?

It's indeed `way_id` overflow, not only `node_id`. But it's mostly for non-OSM data set since the biggest OSM `way_id` value is about 565140676, which fits within 2^30. See comments in https://github.com/Project-OSRM/osrm-backend/issues/5380#issuecomment-468330377

SiarheiFedartsou:
done in https://github.com/Telenav/osrm-backend/commit/6283c6074066f98e6d4a9f774f21ea45407c0d52

Issue #5386, opened 2019-03-04
only routing endpoint returns `data_version` field


gardster:
This field might be in all plugins with JSON response

PR #5387, opened 2019-03-04
Add `data_version` field to responses of all plugins.

gardster:
# Issue

[#5386](https://github.com/Project-OSRM/osrm-backend/issues/5386)

## Requirements / Relations

[#5115](https://github.com/Project-OSRM/osrm-backend/issues/5115)


gardster:
Can somebody review my changes? It is hard to cover data_version with tests in the table plugin because table inside the cucumber DSL represents only table values. And there is no simple and readable cases (from my perspective) how to define additional value to check.

SiarheiFedartsou (approved):

Issue #5388, opened 2019-03-04
Default car height is too high


Zverik:
2.5 meters is okay for a van, but really, even the largest SUVs are under 2 meters. This default prevents routing over some US expressways, notably CT 15.

trigpoint:
Should SUVs be considered normal as a 2m height restriction causes unnecessary restrictions on routing for normal cars.
Eurotunnel has two routes through it teminals for passenger vehicles.

The trains are divided into two sections, a high vehicle section and a section for standard cars with a height limit of 1.85m. OSRM is unable to follow this route as car.lua has a maximum height of 2m.

The issue of car.lua also causes issues here https://map.project-osrm.org/?z=17&center=52.616433%2C-2.442065&loc=52.615993%2C-2.441604&loc=52.615302%2C-2.442162&hl=en&alt=0&srv=0

https://www.geograph.org.uk/photo/4893893

Height restrictions of 6ft/1.8m are not uncommon on the entrance to car parks.

Should normal cars be restricted to places where SUVs can go?

PR #5389, opened 2019-03-04
Decrease max height to 2 meters

Zverik:
See issue #5388

We again stumbled upon a default value that prevents routing over a major highway — in this case, [Highway 15](https://www.openstreetmap.org/way/186846100) in Connecticut. I again did some research and found that while vans can indeed be very high, they are less common that regular sedans and even SUVs.

A little googling showed that even the biggest SUVs (like Ford Expedition EL and VW Caravelle LG) are under two meters. See [this table](https://www.automobiledimension.com/car-search-engine.php) for example. In this pull request I am submitting a lesser default value for the average vehicle high, reflecting this research.


gardster:
Thank you for the contribution!

Issue #5448, opened 2019-05-08
Typo when pre-processed files not found


dburnsii:
When there are osrm files missing and osrm-routed is run, the error message reads "Could not find required filed" where it should probably read "Could not find required file(s)"

PR #5449, opened 2019-05-08
Fixed "Files not found" error message syntax

dburnsii:
# Issue

https://github.com/Project-OSRM/osrm-backend/issues/5448

## Tasklist

 - [ ] review
 - [ ] adjust for comments


src/storage/storage.cpp
```diff
@@ -312,7 +312,7 @@ std::vector<std::pair<bool, boost::filesystem::path>> Storage::GetStaticFiles()
     {
         if (file.first == REQUIRED && !boost::filesystem::exists(file.second))
         {
-            throw util::exception("Could not find required filed: " + std::get<1>(file).string());
+            throw util::exception("Could not find required file(s): " + std::get<1>(file).string());
```
gardster: Does this exception can happen for several files?

dburnsii: It looks like it fires on a per-file basis, so we may not need the plural, but it could potentially be multiple files that it can't find

Issue #5560, opened 2019-09-19
Pessimistic calls to std::move stop compilation on new gcc


peoplestom:
On up to date Arch Linux using GCC 9.1.0 I get errors due to the pessimistic use of std::move in 2 places.

I've removed them, the backend still builds and works for me without them, using CH routing.

PR #5561, opened 2019-09-19
Removed un-needed calls to std::move

peoplestom:
These calls were throwing a pessimistic move error and stopping compilation on GCC 9.1.0

# Issue
[5560](https://github.com/Project-OSRM/osrm-backend/issues/5560) pessimistic std::move calls

## Tasklist

 - [ ] review
 - [ ] adjust for comments

## Requirements / Relations
N/A


peoplestom:
First github PR in a while, let me know if I've missed something in the process.

src/server/api/parameters_parser.cpp
```diff
@@ -47,7 +47,7 @@ boost::optional<ParameterT> parseParameters(std::string::iterator &iter,
 
         // return move(a.b) is needed to move b out of a and then return the rvalue by implicit move
         if (ok && iter == end)
-            return std::move(parameters);
+            return parameters;
```
gardster: does comment above is still relevant?

include/updater/csv_file_parser.hpp
```diff
@@ -122,7 +122,7 @@ template <typename Key, typename Value> struct CSVFilesParser
 
             util::Log() << "Loaded " << filename << " with " << result.size() << "values";
 
-            return std::move(result);
+            return result;
```
gardster: result structure can be huge and it will be better to avoid copying.
Will it fix errors if you will add a std:move statement to the previous return at line 99?

peoplestom: I've removed these calls to std::move as the compiler throws an error saying the call to std::move is less efficient then letting the compiler optimize this move away completely with [copy elision](https://en.cppreference.com/w/cpp/language/copy_elision).

The other way to address this would be to turn off compiler warnings for the 'pessimistic move' error and keep the moves, I prefer to remove std::move calls, but I can have a look at the compiler warning options if that is preferable?

src/server/api/parameters_parser.cpp
peoplestom: No it wasn't, it's been removed.

include/updater/csv_file_parser.hpp
gardster: Thank you. I understand what's happened. I think we can leave it this way (without move) if there is no other occurrences of this warning.

Can you add this compiler to the travis config?

peoplestom:
I believe I've now added gcc9 to Travis  with my last push

gardster:
Thank you! It seems that CI found additional errors. Could you fix them?

peoplestom:
I missed that, I was working off the 5.22.0 tag. I have fixed the additional issues, it compiles and appears to work.
As I'm not familiar with flatbuffer's and that's were the issues were from could someone who know's what they're looking at review this?
It looks like @akashihi may have done the initial flatbuffer work.

akashihi:
I'll take a look on that flatbuffers issue next week. Hope it is not urgent :)

akashihi:
Added https://github.com/Project-OSRM/osrm-backend/pull/5566 on top of master that fixes flatbuffers caused compilation warnings

gardster:
@peoplestom Thank you for the fixes.
Can you format 
```include/engine/api/base_api.hpp include/engine/api/match_api.hpp include/engine/api/route_api.hpp``` with clang-format to pass formatting tests? 
Then I can merge this PR. You can find formatting script in `scripts/format.sh` but make sure that you have proper version (or at least old enough version) of the clang-format.

peoplestom:
I have clang-format 8.0.1, but the changes it kicked up were right were my edits were, so fingers crossed this is ready to merge!

gardster:
Thank you for your contribution! Now OSRM will be gcc9 compatible 🎉

peoplestom:
Awesome! glad I could contribute

Issue #5571, opened 2019-10-03
osrm-extract fails with very large size raster file.


Tomonobu3110:
When we execute osrm-extract with very large raster file, osrm-exract consume too much memory and failed to finish execution.

For example...

    raster_source = raster:load(
      "kanto13_ascii_osrm.asc",  -- file to load
      137.81250000000003,  -- longitude min
      141.328125,          -- longitude max
       34.88593094075317,  -- latitude min
       36.597889133070225, -- latitude max
      24576,               -- number of rows
      40960                -- number of columns

When I try to use raster file which size is 40960 x 24576 (zoom = 13 for whole Kanto (in Japan) area), I could not success the osrm-exract.

osrm-exract read more the one times and fails with below log message.

    [info] [source loader] Loading from kanto13_ascii_osrm.asc  ...
    terminate called after throwing an instance of 'tbb::captured_exception'
      what():  lua: error: Failed to read from raster source kanto13_ascii_osrm.asc: std::bad_allocinclude/extractor/raster_source.hpp:70
    中止 (コアダンプ)

PR #5572, opened 2019-10-03
Reduce memory usage for raster source handling.

Tomonobu3110:
# Issue

This PR solve Issue #5571 

1. Reading line by line instead of whole file.

osrm-extract try to read (load) whole raster file before parse it.
This consume very large memory when raster file is large (such as 40960 x 24576).
To reduce the memory usage, this PR change reading method 
from whole file to line by line.

2. Enable Raster Source cache system

osrm-extract had the cache system for raster sources.
If this works correct, from second time to load raster source,
first raster source (on memory) is reused and no more memory is required.
But unfortunately, this system is not worked by some bug.
I fix it. 

## Tasklist

 - [x] review
 - [x] adjust for comments

## No need Tasklist
- update relevant Wiki pages : No new feature added by this PR.
- add tests (see testing documentation : No new feature added by this PR.

## Requirements / Relations

No relations.


include/storage/io.hpp
```diff
@@ -60,29 +61,31 @@ class FileReader
...
+    template <typename T> void ReadLine(T *dest, const std::size_t count)
+    {
+        if (0 < count)
         {
-            return file_size;
+            const auto &ios =
+                input_stream.getline(reinterpret_cast<char *>(dest), count * sizeof(T));
+            for (std::size_t n = ios.gcount(); n < count; ++n)
+            {
+                reinterpret_cast<char *>(dest)[n] = '\0';
```
gardster: Can we init string with this value on `resize()`?

include/extractor/raster_source.hpp
```diff
@@ -43,37 +50,32 @@ class RasterGrid
         xdim = _xdim;
         ydim = _ydim;
         _data.reserve(ydim * xdim);
+        BOOST_ASSERT(ydim * xdim <= _data.capacity());
 
+        // Construct FileReader
         storage::io::FileReader file_reader(filepath, storage::io::FileReader::HasNoFingerprint);
+        std::string buf;
```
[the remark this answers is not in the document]
Tomonobu3110: Hi, Thanks for your review and comment.
I will fix as your comment. (buf --> buffer)

include/extractor/raster_source.hpp
```diff
@@ -43,37 +50,32 @@ class RasterGrid
...
-            throw util::exception("Failed to parse raster source: " + filepath.string() +
-                                  SOURCE_REF);
+            // read one line from file.
+            file_reader.ReadLine(&buf[0], xdim * 11);
+            boost::algorithm::trim(buf);
+
+            std::vector<std::string> result;
+            boost::split(
+                result, buf, boost::is_any_of(" \r\n\0"), boost::algorithm::token_compress_on);
+            // boost::split(result, buf, boost::is_any_of(" \r\n\0"));
```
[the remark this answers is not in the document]
Tomonobu3110: Thanks for your checking. As you said, this comment line is not needed.
I will remove soon.

include/extractor/raster_source.hpp
```diff
@@ -43,37 +50,32 @@ class RasterGrid
...
+            unsigned int x = 0;
+            BOOST_FOREACH (std::string s, result)
```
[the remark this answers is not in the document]
Tomonobu3110: Hi, thanks for your suggestion.
I will use 11 style loop. Soon, I will modify.

include/storage/io.hpp
Tomonobu3110: Thanks for your comment. I got it. I will update my code.

include/extractor/raster_source.hpp
```diff
@@ -143,8 +145,32 @@ class RasterContainer
     RasterDatum GetRasterInterpolateFromSource(unsigned int source_id, double lon, double lat);
 
   private:
+};
+
+// << singletone >> RasterCache
+class RasterCache
```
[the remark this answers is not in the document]
Tomonobu3110: Hi,
Thanks for your check and comment.

This is most important point of my pull request.
In original code, LoadedSources and LoadedSourcePaths are owned by RasterContainer class.
osrm-extract create more than one RasterContainer instances, and this causes more than one time to load same raster file (with rasterbot.lua).
My first idea is to change LoadesSources and LoadedSourcePaths to static.
But this cases "double free" problem. see error log at below URL, you can find "double free or corrption (fasttop)" error.
https://travis-ci.org/Project-OSRM/osrm-backend/jobs/593429031?utm_medium=notification&utm_source=github_status
This is the reason I introduce the singletone pattern.
Singletone can solve these two problems.
- Multiple load of same raster file.
- Double free error.

Best Regards,
Tomonobu Saito

include/extractor/raster_source.hpp
```diff
@@ -43,37 +50,32 @@ class RasterGrid
         xdim = _xdim;
         ydim = _ydim;
         _data.reserve(ydim * xdim);
+        BOOST_ASSERT(ydim * xdim <= _data.capacity());
 
+        // Construct FileReader
         storage::io::FileReader file_reader(filepath, storage::io::FileReader::HasNoFingerprint);
+        std::string buf;
+        buf.resize(xdim * 11); // INT32_MAX = 2147483647 = 10 chars + 1 white space = 11
+        BOOST_ASSERT(xdim * 11 <= buf.size());
```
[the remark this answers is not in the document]
Tomonobu3110: Thanks for your comment.
In original code, I found below code.
buffer.resized(file_reader.GetSize())
My understanding is "to expand the buffer size to same with raster file size."
By this pull request, I changed from whole size buffer to line buffer to reduce the memory usage.
So, just in case (and respect to original code), I expand the buffer size to "maximum size of one line (logically)".
Maybe this code is not needed or there is better way.
I'm glad if you could give me some advice.

Best Regards,
Tomonobu Saito.

include/storage/io.hpp
Tomonobu3110: The dest is not std::string but template T. So, I could not apply resize().
instead of that, I modified to use memset.
> memset(dest, 0, count * sizeof(T));
Is this satisfy your intention?

Tomonobu3110:
Hi,

https://travis-ci.org/Project-OSRM/osrm-backend/jobs/607969210?utm_medium=notification&utm_source=github_status

I think this failure is not because of my code.
Is there anyone who knows how can I kick build again?
Or do I need to do dummy commit to kick the build system?

Best Regards,
Tomonobu Saito

include/extractor/raster_source.hpp
```diff
@@ -43,37 +50,31 @@ class RasterGrid
...
+            file_reader.ReadLine(&buffer[0], xdim * 11);
+            boost::algorithm::trim(buffer);
+
+            std::vector<std::string> result;
+            boost::split(
+                result, buffer, boost::is_any_of(" \r\n\0"), boost::algorithm::token_compress_on);
+            unsigned int x = 0;
+            for (const auto &s : result)
+            {
+                if (x < xdim)
```
gardster: Hi! 
Do we need this check if we will cover it with  `BOOST_ASSERT(x == xdim);` ? Maybe this check can also be assertion?
I'm not confident with the file format and maybe it is normal to have more values in row than we expect in header: then assertion is not valid.

include/extractor/raster_source.hpp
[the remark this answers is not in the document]
gardster: Hi!
Thank you for the detailed explanation.
 
It seems that RasterContainer binds directly to the lua functions https://github.com/Project-OSRM/osrm-backend/blob/master/src/extractor/scripting_environment_lua.cpp#L188
So isn't it a lua problem that the profile creates several containers?

include/extractor/raster_source.hpp
```diff
@@ -143,8 +144,32 @@ class RasterContainer
...
+// << singletone >> RasterCache
+class RasterCache
+{
+  public:
+    // class method to get the instance
+    static RasterCache &getInstance()
+    {
+        if (NULL == g_instance)
+        {
+            g_instance = new RasterCache();
```
gardster: Where this object will be deleted?

include/extractor/raster_source.hpp
```diff
@@ -4,16 +4,23 @@
...
 #include <boost/assert.hpp>
 #include <boost/filesystem.hpp>
 #include <boost/filesystem/fstream.hpp>
+#include <boost/foreach.hpp>
 #include <boost/spirit/include/qi.hpp>
 #include <boost/spirit/include/qi_int.hpp>
+
 #include <storage/io.hpp>
 
+#include <iostream>
```
DennisOSRM: seems the header is not used. Is this a left-over from debugging?

include/extractor/raster_source.hpp
```diff
@@ -4,16 +4,23 @@
...
 #include <iterator>
+#include <list>
```
DennisOSRM: seems the header is not used. Is this a left-over from debugging?

include/storage/io.hpp
```diff
@@ -60,29 +61,27 @@ class FileReader
...
+            return std::size_t(boost::filesystem::file_size(path)) -
+                   ((fingerprint == FingerprintFlag::VerifyFingerprint) ? sizeof(util::FingerPrint)
+                                                                        : 0);
         }
-
-        // restore the current position
-        input_stream.seekg(position, std::ios::beg);
-
-        if (fingerprint == FingerprintFlag::VerifyFingerprint)
+        catch (boost::filesystem::filesystem_error &ex)
```
DennisOSRM: consider adding `const`

include/extractor/raster_source.hpp
Tomonobu3110: Thanks for your check and comment.
I will remove it after confirmation.

include/extractor/raster_source.hpp
Tomonobu3110: Thanks for your check and comment.
I will remove it after confirmation.

include/storage/io.hpp
Tomonobu3110: Hi, thanks for your comment.
I will add const (since ex should be read-only).

include/extractor/raster_source.hpp
Tomonobu3110: Hi, thanks for your inquiry.
Singletone object will be deleted after main() is finished or exit() is called.
Ref: §3.6.3, in ISO/IEC 14882:2003

include/extractor/raster_source.hpp
Tomonobu3110: Hi,
Thanks for your comment.

I afraid the difference of width (= the number of pixels) between actual raster file and setting in lua.
Especially, if (width of actual file) < (setting in lua) then we will face buffer overflow reading.
As you know, the result is not defined (beyond of control). 

To avoid this situation, I put the `if (x < xdim)` checking, just in case.
If you prefer to remove this checking, I can remove.

Which do you prefer?
a) leave `if (x < xdim)` just in case.
b) remove. we believe user's setting in lua :)

Best Regards,
Tomonobu Saito

include/extractor/raster_source.hpp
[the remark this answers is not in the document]
Tomonobu3110: Hi,
Thanks for your comment.

This is usage of `raster:load` in the lua file (rasterbot.lua).
https://github.com/Project-OSRM/osrm-backend/blob/master/profiles/rasterbot.lua

I think we cannot avoid to create more than one RasterContainer instance, since setup is called more than one time during osrm-extract (according to checking the log of osrm-extract).
But, just in case, I will check the possibility to avoid creating multiple RasterContainer by lua file.
Then, I will update the comment.

Best Regards,
Tomonobu Saito.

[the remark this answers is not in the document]
Tomonobu3110: Hi, 

The instance of RasterContainer is created at here.
https://github.com/Project-OSRM/osrm-backend/blob/master/include/extractor/scripting_environment_lua.hpp#L38

So, we cannot control the creation of RasterContainer instance in lua files (such as rasterbot.lua).

Best Regards,
Tomonobu Saito.

include/extractor/raster_source.hpp
gardster: HI! Thank you for the explanation.
Now I understand that we got xdim setting from the lua profile and it can be a source of mistakes.
I think we should keep this check.

include/extractor/raster_source.hpp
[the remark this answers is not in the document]
gardster: Hi,

thank you for this research. Now I am not against A) strategy. Can you add a briefly comment before this class to avoid confusion of future readers?

include/extractor/raster_source.hpp
gardster: Hi,
didn't know about such behavior.  Thank you for the reference.

include/extractor/raster_source.hpp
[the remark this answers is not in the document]
Tomonobu3110: Hi,
Thanks for your checking and comment.
Yes, I will add a comment then commit again.

include/extractor/raster_source.hpp
Tomonobu3110: Hi,
Thanks for your comment.
OK, I keep this modification.

Tomonobu3110:
Hi,
Thanks for your comment and approval !
I wait this PR is merged to main branch.

include/extractor/raster_source.hpp
Tomonobu3110: Hi, Mr. DennisOSRM,
Thanks for your strong support.
May I ask you to proceed reviewing?
I have already modified the code and committed.
Best Regards,
Tomonobu Saito

DennisOSRM: looks good from my side 👍

Tomonobu3110:
Hi, Mr. DennisOSRM,
Thanks for approval !

Issue #5515, opened 2019-08-13
Consider using embedded_rails tag in bicycle.lua profile


Nate-Wessel:
I was thinking that it would be nice to add consideration of whether or not streets have [embedded rails](https://wiki.openstreetmap.org/wiki/Key:embedded_rails) in the bicycle routing profile. This is a big issue (to my mind anyway) in Toronto where our large surface tram network makes cycling more difficult on many major streets. Taginfo indicates that the tag is fairly common in Europe as well. 

Specifically, embedded rails may require reduced speeds when travelling parallel to the tracks; any rail-crossing maneuvers have to be taken at a sufficient angle that the wheel doesn't risk getting caught in the gap - meaning generally slow but sharp turns. Turns from a lane with rails generally have to be taken more slowly, and travel straight along the tracks is usually slower than normal riding just because you need to consider obstacles ahead in light of a limited ability to maneuver (e.g. car doors, etc.). It can also be nearly impossible to overtake slower cyclists on such streets.

I think it would be sufficient to reduce default speeds by some factor on the relevant ways, but considering a turn penalty somehow could be useful as well. 

I don't think this should be the highest priority issue, but I offer it as a suggestion for the next time someone takes a stab at the bicycle.lua profile. I may update this issue with my own solution when I have a chance.

PR #5580, opened 2019-10-18
Penalizes roads with 'embedded_rails' for bicycle routes.

akashihi:
#5515

After almost failing on a road with tram tracks i thought that this is a good thing to do.

### DIFF
--- a/profiles/bicycle.lua
+++ b/profiles/bicycle.lua
@@ -509,6 +509,11 @@ function safety_handler(profile,way,result,data)
       safety_penalty = safety_penalty * 0.6
     end
 
+    -- embedded rails may require reduced speeds for safe passing
+    if (way:get_value_by_key('embedded_rails')) then
+      safety_penalty = safety_penalty * 0.25
+    end
+
     local forward_is_unsafe = is_unsafe and not data.has_cycleway_forward
     local backward_is_unsafe = is_unsafe and not data.has_cycleway_backward
     local is_undesireable = data.highway == "service" and profile.service_penalties[data.service]


gardster:
Hi!
Do I understand correctly, that this change will penalize routes with embedded rails when you will move along this route? Why do you want to handle such cases? I think that the main problem to cross such rails neither to move along them.

danpat:
@gardster I imagine this PR is meant to penalize situations like the image below which exist in cities with trams/street cars.  Having ridden my bike quite a lot in a city like this, I'm sort of torn.  Certainly, any time you have to cross the tracks it can be a bit sketchy (*especially* in the wet), and the narrower the angle of your crossing, the worse it is (perpendicular crossing is usually fine, even in the wet).

However, roads like this often have bike lanes too, as they're major thoroughfares.  Personally, I wouldn't avoid them, but I can see some people preferring to stay away from the roads with tracks altogether.

intersection-2262239_1280

akashihi:
@gardster Well, moving along those tracks could be unsafe too. For example here [1] you have to move close to tha tracks in parallel with them and it may happen that your bike wheel will slip into rail and you will fail. Same issue here [2] It's not just europe problem, cause Melbourne have same problem [3]

Therefore penalizing streets with tram rails (and now being equipped with the dedicated bike lanes) will result in probably longer, but safer bike routes.
[1] https://www.google.ru/maps/@49.2124499,16.5781748,3a,75y,86.46h,74.91t/data=!3m6!1e1!3m4!1sEOrRoIUAH0x5bAQeHwYKOA!2e0!7i13312!8i6656

[2] https://www.google.ru/maps/@60.1636482,24.9214603,3a,75y,102.78h,71.83t/data=!3m6!1e1!3m4!1sV_nYhj_BvDNY-IpxnSFv4g!2e0!7i13312!8i6656

[3] https://www.google.ru/maps/@-37.8128023,144.9526852,3a,75y,324.93h,55.09t/data=!3m6!1e1!3m4!1sRGkv4rkonufCXBXEpvxGBw!2e0!7i13312!8i6656

Zverik:
As an avid biker in past, when I lead hundreds of cyclist over city streets with plenty of tram tracks, I'm strongly against this change. Tram tracks almost always leave space for cycling. Photo [1] is a rare exception — more often falls happen in cases like [3], when a cyclist is breaking traffic law (you should not ever cross the yellow line).

This change also does not account for different rules in different countries: for example, in some you cannot ride a bicycle in any lanes but the first one, and in others road planners care about cyclists and make embedded lanes, like on the photo by danpat.

Basically introducing this penalty does not differ much from introducing a penalty on pedestrian routing for footways that border with cycleways (you have to walk carefully), or a penalty on car routing on streets with embedded rails, since you have to be careful not to ride into a tram car.

jarek:
(I came here from a link in weeklyOSM. I am an occasional OSRM user mostly via the openstreetmap.org website integration.)

I am a bicycle rider in a city with many stretches of legit dangerous embedded_rails streets: on-street parking and lane widths squeeze cyclists into a death zone between car doors on one side and rails on the other ([Mapillary link](https://www.mapillary.com/map/im/ZI1NDfe56GLF0fn1vUcbww)) and if you are particularly unlucky, the concrete that the rails are set in is broken ([Mapillary](https://www.mapillary.com/map/im/S2_BhogOsF-_7q1jMa-8Dg)).

(Checking now I see that this location also triggered issue #5515 which this aims to address.)

However, some of our (relatively) best bicycle routes also have embedded_rails, but fairly safely separated from the cycleway=track (or an approximation thereof) ([Mapillary view](https://www.mapillary.com/map/im/jTGWYMIUkIP_gK6aO-u-MA)). This is actually the best cycle route through the area, and it doesn't make much sense to penalize it.

It might make sense to penalize only the highways with embedded_rails that do not also have cycleway=lane or cycleway=track for the direction in question.

akashihi:
Speakin of case [1]: after falling here in fornt of car i decided to make that change :) That rail configuration is quite popular in Brno and Czech republic. Here is another one example from Praha [4]  I agree with @jarek that penalizing all roads with track could be too aggresive change and some additional filtering probably needs to be applied first.

[4] https://www.google.ru/maps/@50.0893978,14.4282223,3a,75y,355.2h,85.89t/data=!3m6!1e1!3m4!1ssycY0izZT8o19ds8djX3vw!2e0!7i13312!8i6656

Nate-Wessel:
I think that `embedded_rails` definitely need to be considered in a city like Toronto, per Jarek's examples above. My commute is made quite treacherous by various tram rails on otherwise reasonable streets. 

It seems though like the disagreement in the discussion here stems from applying this rule in a way that is not lane-aware. You could for example have a road with two general traffic lanes in each direction with embedded rails in the middle two lanes. This is very common in Toronto. If the outer lanes are clear, then cycling straight and making right turns is no extra challenge, but left turns from the lane with the tracks can be trouble. 

In the examples Jarek gives, the outer lane is used for car parking, often just at certain times of day (e.g. no parking during rush hour in the peak direction), meaning that cyclists will usually try to ride *between* the two lanes, right down the dashed line. OR, some brave souls will cross the rail to ride in the middle of the tracks... but with impatient drivers coming on from behind, and no safe way to change lanes without slowing down, this is just not a great place to be. 

To treat this case properly, we would need to know what lane we are routing the cyclist into. If there are two lanes in a direction and one is used for parking while the other has embedded rails, then we do have a safety issue. We should treat the parking lane as blocked, and the lane with rails as passable but unsafe. 

Fortunately, we do have [embedded_rails:lanes=*](https://wiki.openstreetmap.org/wiki/Key:embedded_rails:lanes)

Are lanes considered at all in the current profile?

i-ky:
Quoting [OSM Wiki on `embedded_rails`](https://wiki.openstreetmap.org/wiki/Key:embedded_rails#Motivation_and_Rationale):
> There are *many* good reasons for this convention. Nevertheless other traffic should be notified that there is a track, physically, running on the highway. The need for such notification is particularly acute for bicycles, mopeds and other vehicles that have narrow tyres. Railway tracks have deep indentations that narrow wheels can easily and dangerously slip into. **This danger alone warrants a note to routing programs so that they can avoid routing e.g. bicycles through such streets.**

DennisOSRM:

Issue #3427, opened 2016-12-08
env var or flag to enable debug logs


midhunkrishna:
Right now we have to rebuild the application with debug flag to get the debug logs. It would be great if we can toggle the log level with a flag or an env variable.

danpat:
We never return code 207, we only return 200, 400 and 500.  I suspect debug logging would not help you here, it sounds like your requests are not making it to the server.

OSRM debug logging is currently disabled at compile time, for performance reasons.  If I get a chance, I'll do some profiling - measure the cost of enabled-but-not-visible debug log messages.  If the hit is negligible, we could consider doing this and making `logDEBUG` always compiled in, but not visible by default.

wangyoucao577:
@danpat  I did some profiling for the enable/disable compile control for the debug logs: 

## Test env 
- aws r5.2xlarge (8 cpus, 64GB memory) 
- [latest us-west OSM, about 1.7GB](https://download.geofabrik.de/north-america/us-west-latest.osm.pbf) 
- osrm binaries build type: default(i.e. `-DCMAKE_BUILD_TYPE=Release`)
- algorithm: MLD
- commands
  - `time osrm-extract -p osrm-backend/profiles/car.lua us-west-latest.osm.pbf >1.log 2>&1`
  - `time osrm-partition us-west-latest.osrm >2.log 2>&1`
  - `time osrm-customize us-west-latest.osrm >3.log 2>&1`
  - NOTE: use `>1.log 2>&1` redirect logs to file to reduce output to possiblity bad performance on write to `stdout/stderr`. 

## Benchmark 
|branch/commit|`osrm-extract`|`osrm-partition`|`osrm-customize`|
|---------------|---------------|-----------------|------------------|
|[master-68dcab7](https://github.com/wangyoucao577/osrm-backend/commit/68dcab7c153b8ea5389426dee546fbd3cd076910)|real    8m39.028s<br>user    41m53.692s<br>sys     1m13.904s|real    5m11.227s<br>user    26m24.147s<br>sys     0m24.994s|real    1m16.134s<br>user    7m5.574s<br>sys     0m10.759s|
[...]

PR #5628, opened 2019-12-09
Control release mode debug logging output by ENABLE_DEBUG_LOGGING option

wangyoucao577:
# Issue

https://github.com/Project-OSRM/osrm-backend/issues/3427

## Tasklist

 - [ ] review
 - [ ] adjust for comments

## Requirements / Relations


wangyoucao577:
As I write in https://github.com/Project-OSRM/osrm-backend/issues/3427#issuecomment-563284389, the performance of debug log is acceptable.

akashihi:
Can it be a compile time option? So one will be able to choose, whether debug output should not be compiled at all in release mode, or compiled in but not visible by default?

wangyoucao577:
> Can it be a compile time option? So one will be able to choose, whether debug output should not be compiled at all in release mode, or compiled in but not visible by default?

Technically it's ok to be a compile time option, actually the `NDEBUG` is already the one. However, is it really valuable to control it by compiling? Intutively people will expect to have the debug log simply by `-l DEBUG` at runtime, that what I was thought before really deep dive the code. It really made bothered me for a long time. So that I stronglly suggest to remove the compile time control.

danpat:
I think whether performance is acceptable or not depents where the debug log output is being generated.  If someone places a debug message in one of the important loops during data processing, even if logging is suppressed, it could increase processing time.

My preference would be to leave the current behaviour by default - compile-time-disable debug logging, but may be add a CMake rule that *enables* debug logging for release builds if it's desired.

wangyoucao577:
> I think whether performance is acceptable or not depents where the debug log output is being generated. If someone places a debug message in one of the important loops during data processing, even if logging is suppressed, it could increase processing time.
> 
> My preference would be to leave the current behaviour by default - compile-time-disable debug logging, but may be add a CMake rule that _enables_ debug logging for release builds if it's desired.

@danpat  I have added cmake option `ENABLE_DEBUG_LOGGING` as you suggested. Its behavior is very similar with `ENABLE_ASSERTIONS` that defaully `ON` in debug mode and `OFF` in release mode. With `-DENABLE_DEBUG_LOGGING=ON` it can be enabled in release mode also. Please help to review it. Thanks!

akashihi:
Looks good for me!

@danpat @gardster Any objections to merge this?

PR #5729, opened 2020-04-30
Moved test data files to Wolt provided hosting.

akashihi:
# 5718

Mapbox disabled access to the testing data, needed by OSRM build process. This change moves location of those data to Wolt provided hosting.

Windows build dependencies will be provided later.


danpat:
Honestly, the monaco.osm.pbf file is only ~400kb - you could just add it directly to the repo instead of just moving it to a different external download location.

akashihi:
That would be next change :) I had to fix build quickly

akashihi:
@danpat @gardster Any objections to merge this?

akashihi:
@danpat I found some time and moved that monaco file right to the repo.

danpat:
I had started a branch locally to do this, but then one of my kids hit the other kid, and that was three months ago, and apparently this is my life now.

👍

Issue #5782, opened 2020-07-16
[Bug CMake] Link of the non-debug version of TBB


OgreTransporter:
After running CMake I have linker problems. More precisely with TBB, because in Visual Studio the wrong libraries are linked in the configuration "Debug". Example osrm-components _Debug_:
```
D:\test9\deps\TBB\install\lib\tbb.lib
D:\test9\deps\TBB\install\lib\tbbmalloc.lib
D:\test9\deps\boost\install\lib\libboost_date_time-vc141-mt-gd-x64-1_73.lib
D:\test9\deps\boost\install\lib\libboost_chrono-vc141-mt-gd-x64-1_73.lib
D:\test9\deps\boost\install\lib\libboost_filesystem-vc141-mt-gd-x64-1_73.lib
D:\test9\deps\boost\install\lib\libboost_iostreams-vc141-mt-gd-x64-1_73.lib
D:\test9\deps\boost\install\lib\libboost_thread-vc141-mt-gd-x64-1_73.lib
D:\test9\deps\boost\install\lib\libboost_system-vc141-mt-gd-x64-1_73.lib
D:\test9\deps\zlib\install\lib\zlibstaticd.lib
```
Release:
```
D:\test9\deps\TBB\install\lib\tbb.lib
D:\test9\deps\TBB\install\lib\tbbmalloc.lib
D:\test9\deps\boost\install\lib\libboost_date_time-vc141-mt-x64-1_73.lib
D:\test9\deps\boost\install\lib\libboost_chrono-vc141-mt-x64-1_73.lib
D:\test9\deps\boost\install\lib\libboost_filesystem-vc141-mt-x64-1_73.lib
D:\test9\deps\boost\install\lib\libboost_iostreams-vc141-mt-x64-1_73.lib
D:\test9\deps\boost\install\lib\libboost_thread-vc141-mt-x64-1_73.lib
D:\test9\deps\boost\install\lib\libboost_system-vc141-mt-x64-1_73.lib
D:\test9\deps\zlib\install\lib\zlibstatic.lib
```

The query `if(WIN32 AND CMAKE_BUILD_TYPE MATCHES Debug)` only works if you call CMake explicitly with the build configuration _Debug_. But if you generate _Debug_ and _Release_ all libraries are linked to the release version.

Proposal for a patch:
```
diff --git a/CMakeLists.txt b/CMakeLists.txt
index 0b1aaf3ab..0aaf06c2a 100644
[...]

PR #5784, opened 2020-07-16
Bugfix TBB linking #5782

OgreTransporter:
Issue #5782: The query if(WIN32 AND CMAKE_BUILD_TYPE MATCHES Debug) only works if you call CMake explicitly with the build configuration Debug. But if you generate Debug and Release all libraries are linked to the release version.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -530,8 +530,8 @@ else()
 
   find_package(TBB REQUIRED)
   add_dependency_includes(${TBB_INCLUDE_DIR})
-  if(WIN32 AND CMAKE_BUILD_TYPE MATCHES Debug)
-    set(TBB_LIBRARIES ${TBB_DEBUG_LIBRARIES})
+  if(WIN32)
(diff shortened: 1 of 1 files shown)


OgreTransporter:
> It's bad CI is still broken.

But this is not because of my patch ;-)

akashihi:
Of course not! It just makes it hard to test your patch :D

akashihi:
@OgreTransporter CI fails on you patch, but it think that's because we are missing TBB debug libraries in the pre-built dependencies. I'll take a look on how to compile them, update the deps and rerun CI.

OgreTransporter:
@akashihi You could also delete the outdated script cmake/FindTBB.cmake. There are configuration scripts which are delivered with TBB (source: https://github.com/oneapi-src/oneTBB/tree/tbb_2020/cmake/templates).

akashihi:
@OgreTransporter I just added tbb debug libs and everything looks fine. Will remove FindTBB with a separate PR, thank you for the advice.

PR #5815, opened 2020-09-02
Added GCC10 support to CI builds.

akashihi:
No issue here, just extending CI


akashihi:
I have to move it out of woltapp

PR #5836, opened 2020-09-22
Remove unused future

fofanov:
The serialization of the compressed node based graph was changed in
c410c200bd675be1cf5390b6abd52acd90837c9a to no longer be asynchronous.

This removes the unused future object.

## Tasklist

 - [ ] review
 - [ ] adjust for comments

### DIFF
--- a/src/extractor/extractor.cpp
+++ b/src/extractor/extractor.cpp
@@ -1,5 +1,6 @@
 #include "extractor/extractor.hpp"
 
+#include "extractor/compressed_edge_container.hpp"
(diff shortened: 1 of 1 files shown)


src/extractor/extractor.cpp
```diff
@@ -288,16 +288,6 @@ int Extractor::run(ScriptingEnvironment &scripting_environment)
     //
```
fofanov, line 288: Unclear if the above comment is still relevant.

akashihi (approved):

akashihi:
@danpat @gardster any objections to merge that?

PR #5849, opened 2020-10-06
Driving side property added to driving_side.geojson

### LINKED ISSUE #2269: Support Left-Sided Driving

akashihi:
Fixes #2269 

PS Will check, how new hacktoberfest rules are working

### DIFF
--- a/data/driving_side.geojson
+++ b/data/driving_side.geojson
@@ -4,7 +4,7 @@
     {
       "type": "Feature",
       "properties": {
-        
+        "driving_side": "left"
       },
       "geometry": {
         "type": "Polygon",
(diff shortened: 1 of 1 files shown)


mloskot (approved):
As I have just learned from https://github.com/Project-OSRM/osrm-backend/issues/2269#issue-148695754
> At the moment, we only support right-handed driving per default and assume it to be the general mode of driving.

I believe explicit is the right way to go :)

PR #5860, opened 2020-10-17
Undo libosrm API break by overloading method response types

### LINKED ISSUE #5548: libosrm interface broken due to public API changes
Hey folks, I have someone reporting my tiny C wrapper no longer compiling against libosrm.

Looks like you folks changed the libosrm public API in

https://github.com/Project-OSRM/osrm-backend/commit/75aadb0f3f13c4989c2da5aa5b7fad39fefd01db#diff-e8cabd3a392567e2a889841a1e789cc4

which was most likely by accident.

See for context: https://github.com/daniel-j-h/libosrmc/issues/11

Best,
Daniel

### LINKED ISSUE #5741: libosrm breaking change in master after v5.22.0
Hi there. I'm using `libosrm` with a simple consumer code that comes down to something like:

```
osrm::EngineConfig config;
osrm::OSRM osrm(config);

osrm::TableParameters params;
// Fill params with coordinates

osrm::json::Object result;
osrm::Status status = osrm.Table(params, result);
```

This works perfectly against tagged versions of OSRM up to `v5.22.0`, but fails when building `libosrm` from current `master` with something like:

```
error: no matching function for call to ‘osrm::OSRM::Table(osrm::engine::api::TableParameters&, osrm::util::json::Object&) const’
   osrm::Status status = _osrm.Table(params, result);
                                                   ^
/usr/local/include/osrm/osrm.hpp:96:12: note: candidate: osrm::engine::Status osrm::OSRM::Table(const osrm::engine::api::TableParameters&, osrm::engine::api::ResultT&) const
     Status Table(const TableParameters &parameters, osrm::engine::api::ResultT &result) const;
            ^~~~~
```

Indeed, the `Table` interface has been changed [in this commit](https://github.com/Project-OSRM/osrm-backend/commit/75aadb0f3f13c4989c2da5aa5b7fad39fefd01db#diff-701328b0889b0eeb309f95b6fa19428c).

I didn't dig enough to know whether switching from `osrm::json::Object` to `osrm::engine::api::ResultT` would be straithforward, but this looks like a breaking change anyway. All builds relying on previous `libosrm` interface will fail in the same way if this change makes it into a release.

FWIW, original problem spotted at https://github.com/VROOM-Project/vroom/issues/344.

mjjbell:
# Issue
Removes the breaking libosrm API change by replacing the use of variant response type with method overloading.

This adds a bit of boilerplate to convert to/from the variant response used in plugin code. I avoided templating as I think this would have implications on the use of the pimpl pattern. Perhaps someone can correct me on this.

This also has the added benefit of being explicit about the supported response types for each libosrm API method.

## Tasklist

 - [ ] review
 - [ ] adjust for comments

## Requirements / Relations
Closes #5741 
Closes #5548

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,6 @@
 # Unreleased
+    - API:
+      - FIXED: Undo libosrm API break by overloading method response types [#5860](https://github.com/Project-OSRM/osrm-backend/pull/5860)
 
 # 5.23.0
   - Changes from 5.22.0
--- a/example/example.cpp
+++ b/example/example.cpp
@@ -52,15 +52,14 @@ int main(int argc, const char *argv[])
     params.coordinates.push_back({util::FloatLongitude{7.419505}, util::FloatLatitude{43.736825}});
 
     // Response is in JSON format
-    engine::api::ResultT result = json::Object();
+    json::Object result;
 
     // Execute routing request, this does the heavy lifting
     const auto status = osrm.Route(params, result);
 
-    auto &json_result = result.get<json::Object>();
     if (status == Status::Ok)
     {
-        auto &routes = json_result.values["routes"].get<json::Array>();
+        auto &routes = result.values["routes"].get<json::Array>();
 
         // Let's just use the first route
         auto &route = routes.values.at(0).get<json::Object>();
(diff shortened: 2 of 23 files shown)


mjjbell:
Replacing with #5861

PR #5901, opened 2020-12-02
Upgrade ci to xenial

### LINKED ISSUE #5900: Trusty is EOL, update CI
Per this page: https://docs.travis-ci.com/user/reference/trusty/

The trusty dist used int eh Travis CI is EOL and should be updated to Xenial.

dburnsii:
# Issue

Closes #5900 

## Tasklist

 - [ ] review
 - [ ] adjust for comments

## Requirements / Relations

 #5887


mjjbell:
A few build/compiler changes are needed to get this to go green. Will make some fixes.

mjjbell:
The example program [fails](https://travis-ci.com/github/Project-OSRM/osrm-backend/jobs/507249258) to build for `mason-linux-release` with a CXX11 ABI linker failure.

```
CMakeFiles/osrm-example.dir/example.cpp.o:/home/travis/build/Project-OSRM/osrm-backend/example/example.cpp:function main: error: undefined reference to 'osrm::OSRM::Route(osrm::engine::api::RouteParameters const&, mapbox::util::variant<osrm::util::json::Object, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >, flatbuffers::FlatBufferBuilder>&) const'
```

Whilst this could be fixed by also building the example with `-D_GLIBCXX_USE_CXX11_ABI=0`, inspection of the node binary shows it contains libstdc++ symbols from GCC 5.1+

```
readelf -s ../lib/binding/node_osrm.node | grep GLIBCXX_3.4.21
    58: 0000000000000000     0 FUNC    GLOBAL DEFAULT  UND _ZNSt13runtime_errorC2EPK@GLIBCXX_3.4.21 (6)
    59: 0000000000000000     0 FUNC    GLOBAL DEFAULT  UND _ZNSt11logic_errorC1EPKc@GLIBCXX_3.4.21 (6)
    68: 0000000000000000     0 FUNC    GLOBAL DEFAULT  UND _ZNSt6thread15_M_start_th@GLIBCXX_3.4.21 (6)
   248: 0000000000000000     0 FUNC    GLOBAL DEFAULT  UND _ZNSt12out_of_rangeC1EPKc@GLIBCXX_3.4.21 (6)
   296: 0000000000000000     0 FUNC    GLOBAL DEFAULT  UND _ZNSt13runtime_errorC1EPK@GLIBCXX_3.4.21 (6)
   ```
   
This means maintaining standard lib compatibility by downgrading to libstdc++4.9 is not viable with Xenial builds (unless someone knows how to make this work).
 
Node 10 has GCC 4.9 as a minimum [toolchain requirement](https://github.com/nodejs/node/blob/v10.x/BUILDING.md#unix). So despite Trusty being EOL, I don't think we can upgrade CI if we want to keep the same level of support for Node 10.

**Comment by danpat** [comment]
@mjjbell I don't feel strongly that we need to continue supporting Node 10 - it's totally EOL.

**Comment by mjjbell** [comment]
:+1: that will make things easier. In which case we can bump the GCC toolchain to the Xenial default (GCC 5.4.0)

File: src/extractor/restriction_parser.cpp [hunk 1]
```diff
@@ -244,14 +242,15 @@ bool RestrictionParser::ShouldIgnoreRestriction(const std::string &except_tag_st
 
     // Be warned, this is quadratic work here, but we assume that
     // only a few exceptions are actually defined.
-    std::vector<std::string> exceptions;
-    boost::algorithm::split_regex(exceptions, except_tag_string, boost::regex("[;][ ]*"));
```
mjjbell, line 248: Removed usage of Boost.Regex as there's a bug in the version used by Xenial CI.
Some symbols are missing [abi:cxx11] tags which causes linking to fail if not also building OSRM with GCC5.
https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=823978

Used this as an opportunity to replace it with `std::regex`.

I didn't remove Boost.Regex as a dependency completely as it makes the Windows build fail - I'm not sure how to fix that with out changing the Wolt hosted dependencies.

mjjbell:
Superseded by #6071

Issue #4241, opened 2017-07-06
Handle `no_entry` and `no_exit` turn restrictions


daniel-j-h:
Currently we only handle `(fromWayId, viaNodeId, toWayId)` turn restrictions and work for handling way-vias is in progress https://github.com/Project-OSRM/osrm-backend/issues/2681.

What we don't consider at all is `no_entry` and `no_exit` which can have _multiple from and to_ ways, respectively - see spec: http://wiki.openstreetmap.org/wiki/Relation:restriction#Members

What we should do is
- [ ] detect `no_entry` and `no_exit` turn restrictions
- [ ] insert multiple `(from, via, to)` turn restrictions for all `from` and for all `to` ways
- [ ] add cucumber tests for `no_entry` and `no_exit` (might require updating support code)
- [ ] update `taginfo.json`
- [ ] update changelog

MoKob:
@daniel-j-h I just checked taginfo, it seems like there are 5 tags using `no_entry`, none for `no_exit`. I guess we can skip on supporting these?

daniel-j-h:
You need to check for `relation=no_entry/no_exit`

- https://taginfo.openstreetmap.org/tags/restriction=no_entry 419
- https://taginfo.openstreetmap.org/tags/restriction=no_exit 170

given that these transform to multiple simple turn restrictions I would love to have them in.

daniel-j-h:
@MoKob with your turn restriction work merged into master what's your assessment here? What specifically needs to be done and can we simply add multiple "simple" turn restrictions for these tags?

MoKob:
@daniel-j-h yes, we would need to translate the single restriction into multiple ones to allow this.
It requires:

 - [ ] translating a restriction into multiple node restrictions
        - adjust [this](https://github.com/Project-OSRM/osrm-backend/blob/master/src/extractor/restriction_parser.cpp#L133-L181) to not handle single from/via/to but multiple
        - output multiple node-restrictions for all n*m combinations

 - [ ] hand off multiple restrictions [here](https://github.com/Project-OSRM/osrm-backend/blob/master/src/extractor/scripting_environment_lua.cpp#L652-L657)

Issue #4102, opened 2017-05-30
Some issues in restrictions handling code


joto:
I had a look at the restrictions handling code and here are some mixed issues I found:

* https://github.com/Project-OSRM/osrm-backend/blob/master/src/extractor/restriction_parser.cpp#L53 Wrong documentation, return type is a vector (https://github.com/Project-OSRM/osrm-backend/blob/master/src/extractor/restriction_parser.cpp#L59) . That vector only ever contains at most one element!
* Building the KeyFilter https://github.com/Project-OSRM/osrm-backend/blob/master/src/extractor/restriction_parser.cpp#L69-L85 is expensive, the filter should be built once and reused.
* Iterating over the tags with the filter is expensive because of the string comparisons. Doing this three times (https://github.com/Project-OSRM/osrm-backend/blob/master/src/extractor/restriction_parser.cpp#L93 and https://github.com/Project-OSRM/osrm-backend/blob/master/src/extractor/restriction_parser.cpp#L107 and https://github.com/Project-OSRM/osrm-backend/blob/master/src/extractor/restriction_parser.cpp#L183) is suboptimal, as is creating strings just for a comparison (https://github.com/Project-OSRM/osrm-backend/blob/master/src/extractor/restriction_parser.cpp#L110 and https://github.com/Project-OSRM/osrm-backend/blob/master/src/extractor/restriction_parser.cpp#L186), or, even worse, not using them at all (https://github.com/Project-OSRM/osrm-backend/blob/master/src/extractor/restriction_parser.cpp#L109 and https://github.com/Project-OSRM/osrm-backend/blob/master/src/extractor/restriction_parser.cpp#L185).
* Checking the role multiple times (https://github.com/Project-OSRM/osrm-backend/blob/master/src/extractor/restriction_parser.cpp#L133 and https://github.com/Project-OSRM/osrm-backend/blob/master/src/extractor/restriction_parser.cpp#L142 for instance) is expensive. Better to only do this once.
* There must be a better way to split a string on semicolons than https://github.com/Project-OSRM/osrm-backend/blob/master/src/extractor/restriction_parser.cpp#L231.
[...]

PR #5907, opened 2020-12-09
Add support for multiple via-way restrictions

### LINKED ISSUE #4439: Support via-way restrictions for multiple vias
Over in https://github.com/Project-OSRM/osrm-backend/issues/2681, support was added for turn restricitons in OSM that use a `way` as the `via`.

In theory, it's possible to have multiple `via`s for a turn, although in practice they're pretty rare, most restructions are expressed with a single via `way` or `node`.

For completeness, we should support multiple vias for turn restrictions.

mjjbell:
# Issue

multi_via
[link](http://map.project-osrm.org/?z=18&center=50.777998%2C6.094392&loc=50.777765%2C6.094649&loc=50.777687%2C6.094360&hl=en&alt=0&srv=0)

OSRM currently supports restrictions for the following types of turns:
- Turns where the from/to ways are connected at an intersection via-node (NodeRestriction)
- Turns where the from/to ways are connected by a single via-way (WayRestriction) 

OSM allows for turn restrictions with [multiple via-ways](https://wiki.openstreetmap.org/wiki/Relation:restriction#Members) to represent longer and more complex restrictions. This PR adds support for multi via-way turn restrictions.

## Impact
OSM data has [approximately 8500](http://overpass-turbo.eu/s/x1l) multi via-way turn restrictions. These turns will now be restricted when route planning.

## Context
existing_restrictions

In the edge-based graph, a via-node restriction is applied by not creating an edge that represents the turn.

For single via-way restrictions, an edge-based node is created that duplicates the via-way node.
The from-via turn is replaced with a turn onto the duplicated node. The restriction instructions inform which
edges to create from the duplicate node back to the routing graph.

See #4255 and #2681 for more details about the current implementation.

## Supporting multi via-way restrictions
multi_via_new

This PR extends the use of duplicate nodes to support multi via-way restrictions.
In addition, with multi-via restrictions comes the possibility of nested and overlapping restrictions.
For this, we construct a _restriction graph_ to correctly represent these relationships when generating the edge-based graph.

# Restriction graph

The restriction graph is used to represent all possible turn restrictions within the routing graph.

Each node is equivalent to an edge-based graph node, along with the restriction path prefix that was taken to reach this node.

For example, the following three nodes duplicate the same edge-based node **cd**, but represent different restriction paths.

node_description

The restriction graph edges represent turns like they do in the edge-based routing graph. In particular, restriction paths that have the same prefix will share the same restriction graph nodes.

restriction_tree
   
The restriction graph also has an additional feature called _transfer edges_, which are edges between nodes on two different restriction paths. A transfer edge from node A to B can be added if **the path leading to node B is a suffix of the path represented by A.**

Transfer edges allow us to represent overlapping restrictions without needing additional logic for tracking path history or multiple path states.

transfer_edge

The final node in a restriction path stores the restriction instructions. Nested restrictions can lead to new instructions added to a node. For example, in this case the nested node restriction also applies to the via way restriction path.

nested_restriction

# Example Scenario
The construction and use of the restriction graph can be illustrated with the following scenario.

scenario

Take the above scene with the following restrictions applied.

input

Given a list of turn restrictions, the graph is created in multiple steps.

### Step 1: create prefix trees for all restriction paths.
The restriction instructions are added to the final node in its path.
step1

### Step 2: add transfer edges between restriction paths that overlap.
Traverse each restriction path, tracking where the suffix of our current path matches the prefix of any other. If it does, there's opportunity to transfer to the other path *if* the transfer would not be restricted *and* that turn does not take us
further on our current path. Nested restrictions are also added from paths that match the suffix.
step2

### Step 3: Index start and via nodes
- Start nodes are the entry points into a restriction path.
- Via nodes are the intermediate steps along a restriction path.

Start nodes and via nodes are indexed by the current edge of the path they represent for easy retrieval.

start_via

### Step 4: Apply to edge-based graph
During edge-based routing graph construction:
- Add all via nodes to the edge-based routing graph. 
- Add all edges in the restriction graph to the edge-based graph.
- For each start node, remove edges to a node if there is an edge to the via duplicate.
- For each via node, duplicate all edges to the original targets, if they are not restricted

step3
final_restricted

## Complexity
In the worst case, for each edge in a restriction via-way we have to create a duplicate node and duplicate all its turns in the edge-based graph. Given the low number of via-way restrictions in OSM, this will have low impact on the extracted graph size.

Whilst this level of turn restriction complexity is mostly hypothetical, it allows us to avoid building special-case logic for the subset of real-world occurrences, especially at complex intersections. Here is an [example](https://www.openstreetmap.org/#map=19/48.76987/11.43410).

### Non-compressible nodes in via-way

In the current implementation, a restriction with a via-way containing a non-compressible node (e.g. an intersection junction) would be filtered out.
To add support for these types of restrictions, we change a WayRestriction to track all nodes within the via way so that we know which edges remain after node compression. This requires changes to data preparation so that way nodes are tracked and not just turn masks.

### Conditional/unconditional restrictions
Conditional and unconditional restrictions are combined to use the same representation. This makes it easier to construct the 
restriction graph and deal with nested and overlapping restrictions. Given that turn restrictions all have conditional fields when parsed, this does not add any additional memory usage.

## Performance
Initial testing shows negligible change in memory and time usage by `osrm-extract`. I will run some more thorough tests and add in a separate post.

## Circumventing restrictions
When testing this I noticed some multi-via restrictions are not strict enough and could likely lead to
unexpected routing behaviour. A good example of this is performing an extra loop to circumvent the restriction because too much of the initial path is included ([example](https://github.com/Project-OSRM/osrm-backend/blob/5a435f947398123807eed2051a5a4882229d7398/features/car/multi_via_restrictions.feature#L746)).

## Tasklist
 - [ ] review
 - [ ] adjust for comments

## Requirements / Relations
Closes #4439

Possible next steps:
Restriction parsing optimisations: #4102
Supporting multiple from/to relations in no_entry and no_exit restrictions:  #4241


features/car/conditional_restrictions.feature
```diff
@@ -387,217 +387,37 @@ Feature: Car - Turn restrictions
             | m    | p  | mj,jp,jp       |
 
     @no_turning @conditionals
-    Scenario: Car - only_right_turn
+    Scenario: Car - Multiple conditional restrictions applicable to same turn
         Given the extract extra arguments "--parse-conditional-restrictions"
                                             # time stamp for 10am on Tues, 02 May 2017 GMT
         Given the contract extra arguments "--time-zone-file=test/data/tz/{timezone_names}/guinea.geojson --parse-conditionals-from-now=1493719200"
         Given the customize extra arguments "--time-zone-file=test/data/tz/{timezone_names}/guinea.geojson --parse-conditionals-from-now=1493719200"
-        Given the node map
```
mjjbell, line 395: A lot of these tests were duplicated by 879de346ea66. This is just removing the duplication.

woodpeck:
I don't claim to understand every detail but I would like to applaud the impressively detailed documentation & that you've kept an eye on resource usage. Well done!

nilsnolde:
Wow, this is pretty huge! I was always struggling with this in most foss routers. E.g. TomTom data is FULL of multi-via-way turn restrictions. Very sweet, thanks so much!

akashihi:
Impressive job! I will thoroughly review it on a weekend :)

In the meanwhile - all builds are failing with boost/container_hash/hash.hpp: No such file or directory, could you please take a look on that?

mjjbell:
[Here's a breakdown](https://docs.google.com/spreadsheets/d/1Q_XF0P-lVb2BWIlTzW60Izgko3o512ZiWF9pgVqE45s/edit?usp=sharing) of `osrm-extract` performance on a selection of geographic regions.

The PR has minimal effect on runtime and memory usage - within the normal variation between runs.

In terms of the graph output, all the regions show a minor increase in the number of small strongly connected components. This is to be expected - the multi-via implementation is creating more nodes that have reduced (and potentially no) outgoing edges.

I did not analyse the performance of subsequent CH or MLD steps, as there is no significant change in the extracted graph size or structure to warrant it.

akashihi (approved):

akashihi:
As change is big, let's wait for more opinions.

@mjjbell Thank you for fixing restrictions (and conditional restrictions!)

danpat:
> all the regions show a minor increase in the number of small strongly connected components. This is to be expected - the multi-via implementation is creating more nodes that have reduced (and potentially no) outgoing edges.

I think my only concern is unexpected side effects for routes that start/end within the new expanded graphs. I see the tests where you check routes that depart/arrive within the restriction graph - do you feel that these tests encompass all the possible behaviours?  How does a coordinate snapped to an EdgeBasedNode within the restriction graph decide which part of the new restriction graph to select?

mjjbell:
> I think my only concern is unexpected side effects for routes that start/end within the new expanded graphs. I see the tests where you check routes that depart/arrive within the restriction graph - do you feel that these tests encompass all the possible behaviours? How does a coordinate snapped to an EdgeBasedNode within the restriction graph decide which part of the new restriction graph to select?

Coordinates can't be snapped to duplicated nodes in the EBG. As the restriction graph nodes are added as duplicates, it shouldn't be possible.

My understanding is that the RTree will use the edge segments inserted [here](https://github.com/Project-OSRM/osrm-backend/blob/master/src/extractor/edge_based_graph_factory.cpp#L218) by the `InsertEdgeBasedNode` method during graph expansion.  Segments are not inserted for [duplicate nodes](https://github.com/Project-OSRM/osrm-backend/blob/master/src/extractor/edge_based_graph_factory.cpp#L377).

So this gives the desired behaviour that you can't be snapped into a restriction path.

I'm reasonably confident that the tests cover this, and that we would have seen issues with the single-via paths if it was happening. However, I can add a test that focusses specifically on this behaviour to avoid future regressions.

mjjbell:
Thinking this through now - I think there is an edge case when the target is snapped to a node that has a duplicate from the restriction graph. Let me see I can reproduce in the tests.

mjjbell:
This is the minimal test case. Snapping the target onto a via path fails.
It is a problem in the current single-via implementation too, which I think is why I somehow overlooked this.
```
    Scenario: Snap target to restriction via
        Given the node map
            """
            a-1-b-2-c--d
            """

        And the ways
            | nodes |
            | ab    |
            | bc    |
            | cd    |

        And the relations
            | type        | way:from | way:via | way:to  | restriction    |
            | restriction | ab       | bc      | cd      | no_straight_on |

        When I route I should get
            | from | to | route    |
            | 1    | 2  | ab,bc,bc |
```

In which case, we probably do need to include duplicate EBG nodes in the RTree. Which brings up your question of which of the original/duplicates to snap to.

I think the answer would be, all of them. They each have a path from the target that the others do not.

That leads me to believe this can't be fixed until there is routing support for multiple phantom candidates at endpoints.

danpat:
Right, I've been linking to https://github.com/Project-OSRM/osrm-backend/issues/4465 as kind of the gathering point for issues related to only supporting a single start location.

Given how relatively uncommon these types of restrictions are, and the fairly low chance that folks will be requesting start/endpoints within them, we can just live with the edge case.

👍 Thanks for this contribution, it's awesome :-)

matkoniecz:
Thanks! I was just asked by a mapper how to map this and he was a bit sad that OSRM and Graphhopper deployed at OSM website failed to work properly (so what is the point of mapping this? etc).

He was much happier when I found this PR  an was able to mention that it will change :)

Thanks again for implementing this!

zhysong:
Route detour when destination is on the via way of multiple via-ways complex restriction。
It should go through the red route.
image

matkoniecz:
@zhysong Version deployed there is likely using older version of OSRM software.

mjjbell:
@zhysong yes, this is the edge-case mentioned in the [earlier comment](https://github.com/Project-OSRM/osrm-backend/pull/5907#issuecomment-748648815).  Routes with destinations on via ways will be suboptimal or sometimes not found.

A combination of #5953 and https://github.com/mjjbell/osrm-backend/commit/3771e99f74d609735e61ca4fd81424b9d0fc663a will fix this.

zhysong:
@mjjbell I merged your code (#5953 and mjjbell@3771e99) to retest, it can fix detour issue (destination is on the via way of multiple via-ways complex restriction).  
Thank you very much.

zhudelong:
It seems that the cnbg-ebg mapping is not correctly modified to adapt to this new feature. According to my testing, the mapping does not include the duplicated nodes, and this leads to a large partition cell that is composed of all the duplicated nodes. This may need to be further analysed and fixed.

Issue #5892, opened 2020-11-23
Route provided is terrible, and different than demo server


caleb87:
I built OSRM on Ubuntu 18. I ran osrm-routed with MLD. 

OSRM demo gives a superb route. Example Route: http://map.project-osrm.org/?z=10&center=35.891831%2C-98.521957&loc=36.253133%2C-98.480072&loc=35.821432%2C-98.779449&hl=en&alt=0&srv=0

Mine OSRM server provides this route: http://map.project-osrm.org/?z=12&center=35.872777%2C-98.565731&loc=36.253133%2C-98.480072&loc=35.948273%2C-98.493805&loc=35.890441%2C-98.507881&loc=35.821432%2C-98.779449&hl=en&alt=0&srv=0

That route uses roads classified as residential, instead of 65mph secondary roads. 

This is using the default car profile. 

Traffic data shouldn't affect this, because there is no traffic data in my server, and the optimal route regardless of traffic is the one the OSRM demo provides.

![osrmbug](https://user-images.githubusercontent.com/6970514/100024163-80472f00-2da3-11eb-9437-983ecc573983.jpg)

danpat:
@datendelphin Is the demoserver running unmodified versions of `car.lua`?  If not, do you have a link to the changes?

datendelphin:
Hi! The profiles are stored here: https://github.com/fossgis-routing-server/cbf-routing-profiles/

systemed:
I wonder if this might be an issue with differing maxspeed handling. The residential roads in this area all have `maxspeed=45 mph` as a result of an automated edit. That's fine but in reality you're unlikely to hit 45mph * 0.8 (`speed_reduction`) on these unpaved roads. (As ever in the rural US, the surface has generally not yet been tagged.)

caleb87:
@systemed I suspected the same, and I actually made a PBF that removed all 45 MPH roads in Oklahoma counties. The route was still the same. I should have saved the PBF so others could have tested it, but I gave up and just ended up deleting the AWS instances I had it on.

I'm currently using Graphhopper for routing, but I'd still like to switch to OSRM because it seems to give better routes. I saw your debug script which seems like the best way to figure this out. Do you have a usage for dummies example? Is this something to run from the API or from terminal?

PR #5908, opened 2020-12-10
Profile debug script which fetches a way from OSM

systemed:
# Issue

When debugging profile issues (e.g. #5892), it's useful to be able to see instantly what the profile is calculating for a given way, without having to reprocess a whole OSM extract.

This short script fetches a way from the OSM API and runs the requested Lua profile over it, then prints the result:

````
$ lua5.1 debug_way.lua car 14672606
{ 
  highway = 'residential',
  maxspeed = '45 mph',
  name = 'E0770 Road',
  ['source:maxspeed'] = '45 in Blaine County unless otherwise posted.',
  ['tiger:cfcc'] = 'A41',
  ['tiger:county'] = 'Blaine, OK',
  ['tiger:name_base'] = 'E0770',
  ['tiger:name_type'] = 'Rd',
  ['tiger:reviewed'] = 'no' 
}
=>
{ 
  backward_classes = {},
  backward_mode = 1,
  backward_rate = 16.09,
  backward_speed = 57.924,
  duration = 0,
  forward_classes = {},
  forward_mode = 1,
  forward_rate = 16.09,
  forward_speed = 57.924,
  is_left_hand_driving = false,
  is_startpoint = true,
  name = 'E0770 Road',
  road_classification = { 
    may_be_ignored = false,
    road_priority_class = 10 
  } 
}
````

It uses the existing debug handler (tweaked to not error on the missing `get_location_tag` C++ method) and builds upon the example `profile_debugger.lua` script.

I use a similar script to this for cycle.travel and have found it enormously helpful in debugging profile issues.

## Tasklist

 - [ ] review
 - [ ] adjust for comments

## Requirements / Relations

Required Lua libraries are listed in the comments at the top of the script.

### DIFF
--- a/profiles/debug_way.lua
+++ b/profiles/debug_way.lua
@@ -1,0 +1,44 @@
+--
+-- Fetch a way from the OpenStreetMap API and run the given profile over it.
+--
+-- You'll need to install luasec and xml2lua first:
+--    > luarocks-5.1 install xml2lua
+--    > luarocks-5.1 install luasec
+-- [may require admin privileges]
+-- 
+-- Then to test way 2606296 using the foot profile:
+--    > lua5.1 debug_way.lua foot 2606296
+--
+
+-- initialise libraries
+local pprint = require('lib/pprint')
+local Debug = require('lib/profile_debugger')
+local xml2lua = require('xml2lua')
+local handler = require('xmlhandler.tree')
+local https = require('ssl.https')
+
+-- load the profile
+Debug.load_profile(arg[1])
+
+-- load way from the OSM API
+local url = 'https://www.openstreetmap.org/api/0.6/way/'..arg[2]
+local body, statusCode, headers, statusText = https.request(url)
+
+-- parse way tags
+local parser = xml2lua.parser(handler)
+parser:parse(body)
+
+-- convert XML-flavoured table to a simple k/v table
+local way = {}
+for i, p in pairs(handler.root.osm.way.tag) do
+  way[p._attr.k] = p._attr.v
+end
+
+-- call the way function
+local result = {}
+Debug.process_way(way,result)
+
+-- print input and output
+pprint(way)
+print("=>")
+pprint(result)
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: profiles/debug_way.lua [hunk 1]
```diff
@@ -0,0 +1,44 @@
+--
+-- Fetch a way from the OpenStreetMap API and run the given profile over it.
+--
+-- You'll need to install luasec and xml2lua first:
+--    > luarocks-5.1 install xml2lua
+--    > luarocks-5.1 install luasec
```
jcoupey, line 6: It looks like `luasec` isn't really required? I got an error all right prior to installing `xml2lua` but nothing related to not having run `luarocks install luasec`.

profiles/debug_way.lua
```diff
@@ -0,0 +1,44 @@
...
+-- [may require admin privileges]
+-- 
+-- Then to test way 2606296 using the foot profile:
+--    > lua5.1 debug_way.lua foot 2606296
```
jcoupey, line 10: My system (Ubuntu) currently does not have `lua5.1` installed but `lua5.2` which is also accessible via the `lua` generic command. If the same apply across other systems, pointing to simply `lua` would make it more generic.

jcoupey (commented):
Handy script indeed, just tested it.

profiles/debug_way.lua
akashihi: I guess it applies not only for ubuntu, so indeed, this needs to be just `lua`

akashihi (changes requested):
Please replace docs references to 5.1 with more generic references to lua

profiles/debug_way.lua
systemed: I've rephrased the comment to note that you may have this already (my Ubuntu box and Mac didn't).

profiles/debug_way.lua
systemed: No problem, changed (blame Homebrew for this one ;) ).

jcoupey (approved):

akashihi (approved):

PR #5923, opened 2021-01-01
Avoid copying ManyToMany table results

mjjbell:
# Issue
Regardless of any copy elision on the returned pair value, the duration and distance results are always copied.

Fix this by passing rvalue references to `std::make_pair`.

This could save copying a few megabytes on `Table` and `Trip` requests with hundreds of sources/destinations and waypoints respectively. I haven't profiled this to see if this affects performance in any other way.

## Tasklist

 - [ ] review
 - [ ] adjust for comments


jcoupey (approved):
Looks like a nice low-hanging fruit optimization. @mjjbell have you been able to notice a significant computing time reduction?

danpat:
The big benefit here is probably going to be reduce memory footprint under heavy load when making lots of concurrent matrix requests - I suspect the overall performance improvement will be really small.  The copy only happens once per request, not within the compute intensive processing loop.

Definitely a good change though, 👍

mjjbell:
@jcoupey sorry I don't have any numbers. I just noticed it and thought it would be a quick win.

PR #5926, opened 2021-01-06
Reduce copying in API parameter constructors

mjjbell:
# Issue
When using non-default constructors for the API parameter classes, vector arguments like `coordinates` and `hints` are copied at least once (twice when passed as lvalue arguments).
    
This PR enables perfect forwarding of `BaseParameters` arguments and pass-by-value/move in the constructor that uses the argument. This ensures we copy at most once (zero for rvalue arguments).

There is one case that I have omitted from forwarding - `waypoints` in `MatchParameters`, which inherits from `RouteParameters` and has a slightly different argument order that complicates things. I've left it as-is for readability.

## Impact
This will have no effect on requests made to `osrm-routed` or via NodeJS bindings as they both use the default constructor and build parameters incrementally. It only impacts libosrm API usage of non-default `*Parameters` constructors.

## Requirements / Relations

There is previous discussion in #2991 about enabling perfect forwarding in these constructors, suggesting this would break the libosrm API. Unless this is referring to a binary compatibility issue which I'm not aware of, I don't think this is a problem. 

With this change, lvalue arguments will still be copied, and rvalue arguments will be moved from, which should meet the expectation of the caller.

## Tasklist

 - [ ] review
 - [ ] adjust for comments

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
       - REMOVED: we no longer publish Node 8 binary modules (they are still buildable from source) [#5918](https://github.com/Project-OSRM/osrm-backend/pull/5918)
     - Routing:
       - FIXED: Avoid copying ManyToMany table results [#5923](https://github.com/Project-OSRM/osrm-backend/pull/5923)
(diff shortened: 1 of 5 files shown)


akashihi (approved):
Looks safe and clear

Issue #5209, opened 2018-09-18
Project status?


systemed:
It seems to be a reasonably open secret (by which I mean SOTM bar chat ;) ) that Mapbox is currently putting most of its development efforts behind Valhalla; and that many of the developers formerly working on OSRM are no longer involved with it.

From a distance it looks like OSRM is in "maintenance mode", which is obviously much better than being abandoned, and it's good to see PRs like #5197 moving the project on. Nonetheless the fast pace of development since 2013 has evidently abated for now.

OSM data does (slowly) evolve and if OSRM is to continue to be useful then it'll need to keep up with this. Much of this can be achieved via profiles, and keeping these up-to-date is fairly trivial, but there are some more significant data issues which will require real engineering commitment (I'm thinking in particular of #4439 to address the increasing number of multiple-via-way restrictions, but there will be others).

So it would be good to have a steer as to the future direction of the project. Does Mapbox intend to continue maintaining OSRM in perpetuity, or only until Valhalla reaches feature/performance parity, or...? Is there any interest in moving it to the community (a la CartoCSS)?

(FWIW I'm not concerned for cycle.travel, which uses an extensive fork of 4.9 and which is very unlikely to return to trunk, but I do use current versions of OSRM for some contract work. Looking at the issue tracker here, there is still a lot of interest in OSRM and people are building solutions on top of it.)

In any case, thanks for supporting so many great improvements over the last five years and being good stewards of the project!

systemed:
> From a distance it looks like OSRM is in "maintenance mode", which is obviously much better than being abandoned

18 months on, I think we can probably say that OSRM is to all intents and purposes abandoned. That's fine - priorities change and I certainly wouldn't mean this as a criticism of anyone. But Mapbox have evidently ceased development and promotion, yet the Project-OSRM organisation is still essentially Mapbox employees or alumni.

At the least, it would be good to have a maintainership model where there's some assurance that decent patches (even if trivial) are likely to get merged. At present, my reading is that potential new users are unlikely to use OSRM because it's been abandoned; therefore there are fewer patches; therefore people are less likely to use OSRM; etc. etc.

Mapbox has successfully transitioned legacy products to community maintainership in the past (carto, Tilemill) and it would be good to see the same happen here.

(For clarity, I'm personally happy to take on issue-gardening, profile stuff and so on, but my C++-fu and graph theory knowledge are certainly not strong enough for me to be sole maintainer!)

[...]

Issue #4741, opened 2017-12-17
Deprecate `hints` parameter


TheMarex:
The `hints` parameter is probably one of the worst concepts of our API that exposes internal data structures to the client and allows them to pass this information back to us. This breaks a few encapsulations, means we need to ship a clunky Base64 encoder and have our internal `PhantomNode` size impact the size of our response objects.

It is straight forward to remove this parameter from the HTTP API but the C++ API is another problem.
Since we encode internal state in the `Hints` object that is part of a public interface, we can't just remove this class. I don't see an easy way to do this without breaking the API.

daniel-j-h:
Related:
- generate hints api flag https://github.com/Project-OSRM/osrm-backend/pull/3456
- v6 api list https://github.com/Project-OSRM/osrm-backend/issues/3298
- osm id as hiint https://github.com/Project-OSRM/osrm-backend/issues/3219

PR #5953, opened 2021-02-03
Support snapping to multiple ways at an input location

mjjbell:
# Issue

before_after_snap_roundabout
[link](https://map.project-osrm.org/?z=18&center=51.509038%2C0.019113&loc=51.509095%2C0.019195&loc=51.509090%2C0.019441&hl=en&alt=0&srv=1)

This PR improves routing results by adding support for snapping to multiple ways at input locations.

## Context
before_snap_2

Input locations are snapped to positions on ways, which act as the source, target or via waypoint positions on the route.

Currently, OSRM only snaps an input location to **one** way. In certain scenarios, there are multiple ways to choose from:
- Intersections
- Connection points between two edges of a way
- Overlapping roads (e.g. bridges, tunnels)
- Edges that are part of via-ways in turn restrictions

The choice of selected way can impact the discovered route. Some good examples of the effect this has are discussed in #4465

This PR makes the change to use all ways at the snapped location as source/target candidates for routing search, ensuring we always find the best route, and not the one dependent on the way selected.

after_snap_2

### Design Choice
The decision to support multiple candidates for a snapped location was chosen for a few reasons:
- It doesn’t require special case logic for particular intersections
- It can be extended to support snapping to multiple equidistant locations from the input
- It acts as a first step towards supporting snapping to the k-nearest locations 

The downside of this approach is that using multiple source/target candidates increases the routing search space. However, in practice this has minimal effect on performance, given:
- Most locations will still only snap to one way
- Even with multiple candidates for an input, most candidates will be connected by a valid turn, so many paths will be quickly pruned.

### PhantomNodeCandidates

We extend the `PhantomNode` representation for a snapped location to `PhantomNodeCandidates`, a vector of snapped locations.

Algorithms that snap large numbers of locations, such as `table` requests, will generate many small vectors. The implementation reuses the vectors fetched from the geospatial query result throughout the whole request. The performance hit appears to be minimal. 

### Snapping to connection between two edges of a way segment
snap_mid_seg

All the routing algorithms assume a way segment only appears once in the search. 
When snapping to the connection between two edges of a segment, we have to decide which edge will represent the segment in the search. 

Segment edges are compressed and include forward and reverse directions. To decide which compressed edge to choose, we consider which is most likely to be useful as a source/target for a route. This can depend on edge traffic updates, approach parameters, etc.)

The implementation makes the choice between two compressed edges of a way segment in a consistent way. In very rare situations, it’s possible that some potential routes will not be found by having to select one over the other.

invalid_snapped

### Nearest vs Big SCC

The existing snapping logic considers whether the snapped way is part of a tiny strongly connected component with other snapped locations, and fallbacks to the nearest candidate from the big SCC if this is not the case. This PR keeps and extends this logic for snapping to multiple candidates.

### Geospatial Queries

The geospatial API for finding graph segments nearest to the location had many similar functions which represented permutations of parameter combinations. Given the logic for filtering is now more complex, I’ve consolidated the API functions and moved the parameter selection into one place.

### Routing Algorithms

`trip`, `table`, and `route` now use the multiple candidate snapping. I did not include `match`, as my depth of understanding of the map matching algorithm is insufficient to figure out how “nearest candidates at the same location” would be represented in the output. Maybe a good follow-up task.

### Shortest Path

Internally, the shortest path search function was passed in a list of endpoint candidates for each leg of the route. I have changed this to be a list of waypoint candidates instead, which makes it clear that the target candidates of the previous leg are the source candidates of the next leg. Now that candidates are vectors of `PhantomNodes`, it also avoids making a copy of each vector.

Not surprisingly, the dynamic programming approach for the shortest path search is made more complicated with multiple source/target nodes. I’ve tried as much as possible to avoid unnecessary vector allocations when tracking state between the legs.

### Libosrm / HTTP API changes

Returning waypoint information in API results is now trickier as we have multiple `PhantomNodes` representing each waypoint.

To avoid a breaking change, the solution I went with is to:
-  Use the location of the first candidate (this is fine as all candidates at a waypoint have the same location)
- Combine the names of the roads of each of the waypoint candidates
- Concatenate the hint strings (so hint size in the result is now `len(candidates)*HINT_SIZE`. I'm assuming changing the hint size is not an API break?

For hints in particular, this allows the hint string to contain all the information for multiple candidates. If we only returned the hint for the `PhantomNode` selected for the particular API call,  it could lead to suboptimal results as a hint for future requests.

EDIT: This is a breaking change in the request parameters. `hints` are now `vector<vector<hint>>`

## Tests

The changes to existing test values should be self explanatory.  A couple of them have weird properties that might need cleaning up:
- One relies on negative turn penalty duration (which creates a negative step duration in the result)
- One includes an `only_*` restriction into a one-way street going the other way

In both cases, they have been changed to the point that they still succeed as before, although they might need revisiting to question the assumptions being made.

Whilst creating new tests, I also noticed some weird behaviours when combining zero speed traffic updates with a source and target on the same segment. It’s quite a niche scenario though, so I omitted tests for this in the PR and will create an issue with details if anyone is interested.

## Performance
I’ll provide a full analysis for the effect on query performance in another comment. The initial findings show a small performance hit, at the expense of finding shorter routes, especially when many waypoints snap to intersections.

## Tasklist

 - [ ] review
 - [ ] adjust for comments

### DIFF
--- a/features/car/restrictions.feature
+++ b/features/car/restrictions.feature
@@ -411,7 +411,7 @@ Feature: Car - Turn restrictions
                     y
             i j f b x a e g h
 
-                  c   d
+                  c1   d
             """
 
         And the ways
@@ -438,7 +438,7 @@ Feature: Car - Turn restrictions
         When I route I should get
             | from | to | route                               |
             | e    | f  | ae,xa,bx,fb,fb                      |
-            | c    | f  | dc,da,ae,ge,hg,hg,ge,ae,xa,bx,fb,fb |
+            | 1    | f  | dc,da,ae,ge,hg,hg,ge,ae,xa,bx,fb,fb |
             | d    | f  | da,ae,ge,hg,hg,ge,ae,xa,bx,fb,fb    |
 
     @except
--- a/features/testbot/traffic_turn_penalties.feature
+++ b/features/testbot/traffic_turn_penalties.feature
@@ -47,7 +47,7 @@ Feature: Traffic - turn penalties applied to turn onto which a phantom node snap
             | 1    | e  | ab,be,be | 36 km/h  | 30s +-1 |
             | b    | f  | bc,cf,cf | 36 km/h  | 40s +-1 |
             | 2    | f  | bc,cf,cf | 36 km/h  | 30s +-1 |
-            | c    | g  | cd,dg,dg | 144 km/h | 10s +-1 |
+            | c    | g  | cd,dg,dg | 72 km/h  | 20s +-1 |
             | 3    | g  | cd,dg,dg | 54 km/h  | 20s +-1 |
 
     Scenario: Weighting based on turn penalty file with weights
@@ -65,5 +65,5 @@ Feature: Traffic - turn penalties applied to turn onto which a phantom node snap
             | 1    | e  | ab,be,be | 36 km/h  | 30s +-1 | 6.8,20,0   |
             | b    | f  | bc,cf,cf | 36 km/h  | 40s +-1 | 20,20,0    |
             | 2    | f  | bc,cf,cf | 36 km/h  | 30s +-1 | 10.1,20,0  |
-            | c    | g  | cd,dg,dg | 144 km/h | 10s +-1 | 120.8,20,0 |
+            | c    | g  | cd,dg,dg | 72 km/h  | 20s +-1 | 120.8,20,0 |
             | 3    | g  | cd,dg,dg | 54 km/h  | 20s +-1 | 110.9,20,0 |
--- a/include/engine/api/base_parameters.hpp
+++ b/include/engine/api/base_parameters.hpp
@@ -51,14 +51,14 @@ namespace api
  * Holds member attributes:
  *  - coordinates: for specifying location(s) to services
  *  - hints: hint for the service to derive the position(s) in the road network more efficiently,
- *           optional per coordinate
+ *           optional per coordinate. Multiple hints can be provided for a coordinate.
  *  - radiuses: limits the search for segments in the road network to given radius(es) in meter,
  *              optional per coordinate
  *  - bearings: limits the search for segments in the road network to given bearing(s) in degree
  *              towards true north in clockwise direction, optional per coordinate
  *  - approaches: force the phantom node to start towards the node with the road country side.
  *
- * \see OSRM, Coordinate, Hint, Bearing, RouteParame, RouteParameters, TableParameters,
+ * \see OSRM, Coordinate, Hint, Bearing, RouteParameters, TableParameters,
  *      NearestParameters, TripParameters, MatchParameters and TileParameters
  */
 struct BaseParameters
--- a/include/engine/hint.hpp
+++ b/include/engine/hint.hpp
@@ -47,8 +47,9 @@ namespace datafacade
 class BaseDataFacade;
 }
 
-// Is returned as a temporary identifier for snapped coodinates
-struct Hint
+// SegmentHint represents an individual segment position that could be used
+// as the waypoint for a given input location
+struct SegmentHint
 {
     PhantomNode phantom;
     std::uint32_t data_checksum;
@@ -57,16 +58,31 @@ struct Hint
                  const datafacade::BaseDataFacade &facade) const;
 
     std::string ToBase64() const;
-    static Hint FromBase64(const std::string &base64Hint);
+    static SegmentHint FromBase64(const std::string &base64Hint);
 
-    friend bool operator==(const Hint &, const Hint &);
-    friend std::ostream &operator<<(std::ostream &, const Hint &);
+    friend bool operator==(const SegmentHint &, const SegmentHint &);
+    friend bool operator!=(const SegmentHint &, const SegmentHint &);
+    friend std::ostream &operator<<(std::ostream &, const SegmentHint &);
 };
 
-static_assert(sizeof(Hint) == 80 + 4, "Hint is bigger than expected");
-constexpr std::size_t ENCODED_HINT_SIZE = 112;
-static_assert(ENCODED_HINT_SIZE / 4 * 3 >= sizeof(Hint),
-              "ENCODED_HINT_SIZE does not match size of Hint");
+// Hint represents the suggested segment positions that could be used
+// as the waypoint for a given input location
+struct Hint
+{
+    std::vector<SegmentHint> segment_hints;
+
+    bool IsValid(const util::Coordinate new_input_coordinates,
+                 const datafacade::BaseDataFacade &facade) const;
+
+    std::string ToBase64() const;
+    static Hint FromBase64(const std::string &base64Hint);
+};
+
+static_assert(sizeof(SegmentHint) == 80 + 4, "Hint is bigger than expected");
+constexpr std::size_t ENCODED_SEGMENT_HINT_SIZE = 112;
+static_assert(ENCODED_SEGMENT_HINT_SIZE / 4 * 3 >= sizeof(SegmentHint),
+              "ENCODED_SEGMENT_HINT_SIZE does not match size of SegmentHint");
+
 } // namespace engine
 } // namespace osrm
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -49,6 +49,7 @@
       - FIXED: Completed support for no_entry and no_exit turn restrictions. [#5988](https://github.com/Project-OSRM/osrm-backend/pull/5988)
       - ADDED: Add support for non-round-trips with a single fixed endpoint. [#6050](https://github.com/Project-OSRM/osrm-backend/pull/6050)
       - FIXED: Improvements to maneuver override processing [#6125](https://github.com/Project-OSRM/osrm-backend/pull/6125)
+      - ADDED: Support snapping to multiple ways at an input location. [#5953](https://github.com/Project-OSRM/osrm-backend/pull/5953)
 
 # 5.26.0
   - Changes from 5.25.0
--- a/features/bicycle/alley.feature
+++ b/features/bicycle/alley.feature
@@ -28,7 +28,7 @@ Feature: Bicycle - Route around alleys
 
         When I route I should get
             | from | to | a:nodes    | weight | #                                |
-            | a    | f  | 1:2:3:6    | 200.4  | Avoids d,e,f                     |
-            | a    | e  | 1:2:5      | 176.4  | Take the alley b,e if neccessary |
-            | d    | f  | 4:1:2:3:6  | 252.6  | Avoids the alley d,e,f           |
+            | a    | f  | 1:2:3:6    | 196.2  | Avoids d,e,f                     |
+            | a    | e  | 1:2:5      | 172.2  | Take the alley b,e if neccessary |
+            | d    | f  | 4:1:2:3:6  | 248.4  | Avoids the alley d,e,f           |
 
--- a/features/car/destination.feature
+++ b/features/car/destination.feature
@@ -7,10 +7,10 @@ Feature: Car - Destination only, no passing through
     Scenario: Car - Destination only street
         Given the node map
             """
-            a       e
-              b c d
+            a         e
+              b1 c 2d
 
-            x       y
+            x         y
             """
 
         And the ways
@@ -23,21 +23,21 @@ Feature: Car - Destination only, no passing through
         When I route I should get
             | from | to | route      |
             | a    | b  | ab,ab      |
-            | a    | c  | ab,bcd     |
-            | a    | d  | ab,bcd,bcd |
+            | a    | c  | ab,bcd,bcd |
+            | a    | 2  | ab,bcd,bcd |
             | a    | e  | axye,axye  |
             | e    | d  | de,de      |
-            | e    | c  | de,bcd     |
-            | e    | b  | de,bcd,bcd |
+            | e    | c  | de,bcd,bcd |
+            | e    | 1  | de,bcd,bcd |
             | e    | a  | axye,axye  |
 
     Scenario: Car - Destination only street
         Given the node map
             """
-            a       e
-              b c d
+            a         e
+              b1 c 2d
 
-            x       y
+            x         y
             """
 
         And the ways
@@ -51,12 +51,12 @@ Feature: Car - Destination only, no passing through
         When I route I should get
             | from | to | route       |
             | a    | b  | ab,ab       |
-            | a    | c  | ab,bc       |
-            | a    | d  | ab,cd       |
+            | a    | c  | ab,bc,bc    |
+            | a    | 2  | ab,bc,cd    |
             | a    | e  | axye,axye   |
             | e    | d  | de,de       |
-            | e    | c  | de,cd       |
-            | e    | b  | de,bc       |
+            | e    | c  | de,cd,cd    |
+            | e    | 1  | de,cd,bc    |
             | e    | a  | axye,axye   |
 
     Scenario: Car - Routing inside a destination only area
@@ -117,6 +117,7 @@ Feature: Car - Destination only, no passing through
                    +    \
                    +    |
                    d    |
+                   1    |
                     \___e
             """
 
@@ -129,7 +130,7 @@ Feature: Car - Destination only, no passing through
         When I route I should get
             | from | to | route         |
             | e    | a  | acbe,acbe     |
-            | d    | a  | de,acbe,acbe  |
+            | 1    | a  | de,acbe,acbe  |
             | c    | d  | cd,cd         |
 
     Scenario: Car - Routing through a parking lot tagged access=destination,service
--- a/features/guidance/merge-segregated-roads.feature
+++ b/features/guidance/merge-segregated-roads.feature
@@ -332,10 +332,11 @@ Feature: Merge Segregated Roads
               |
              .b.
             c   h
+            1   |
+            |   4
             |   |
-            |   |
-            1   2
-            |   |
+            2   |
+            |   3
             d   g
              'e'
               |
@@ -354,13 +355,13 @@ Feature: Merge Segregated Roads
             | hb    | road   | yes    |
 
         When I route I should get
-            | waypoints | turns         | route       | intersections									  |
+            | waypoints | turns         | route       | intersections                                     |
             | a,f       | depart,arrive | road,road   | true:180,false:0 true:180,false:0 true:180;true:0 |
-            | c,f       | depart,arrive | bridge,road | true:180,false:0 true:180;true:0				  |
             | 1,f       | depart,arrive | bridge,road | true:180,false:0 true:180;true:0                  |
+            | 2,f       | depart,arrive | bridge,road | true:180,false:0 true:180;true:0                  |
             | f,a       | depart,arrive | road,road   | true:0,true:0 false:180,true:0 false:180;true:180 |
-            | g,a       | depart,arrive | bridge,road | true:0,true:0 false:180;true:180                  |
-            | 2,a       | depart,arrive | bridge,road | true:0,true:0 false:180;true:180                  |
+            | 3,a       | depart,arrive | bridge,road | true:0,true:0 false:180;true:180                  |
+            | 4,a       | depart,arrive | bridge,road | true:0,true:0 false:180;true:180                  |
 
     @negative
     Scenario: Traffic Circle
--- a/features/testbot/bearing.feature
+++ b/features/testbot/bearing.feature
@@ -67,10 +67,10 @@ Feature: Compass bearing
     Scenario: Bearing in a roundabout
         Given the node map
             """
-            k d c j
-            e     b
-            f     a
-            l g h i
+            k d 1c j
+            e      b
+            f      a
+            l g2 h i
             """
 
         And the ways
@@ -94,8 +94,8 @@ Feature: Compass bearing
 
         When I route I should get
             | from | to | route                   | bearing                                                     |
-            | c    | b  | cd,de,ef,fg,gh,ha,ab,ab | 0->270,270->225,225->180,180->135,135->90,90->45,45->0,0->0 |
-            | g    | f  | gh,ha,ab,bc,cd,de,ef,ef | 0->90,90->45,45->0,0->315,315->270,270->225,225->180,180->0 |
+            | 1    | b  | cd,de,ef,fg,gh,ha,ab,ab | 0->270,270->225,225->180,180->135,135->90,90->45,45->0,0->0 |
+            | 2    | f  | gh,ha,ab,bc,cd,de,ef,ef | 0->90,90->45,45->0,0->315,315->270,270->225,225->180,180->0 |
 
     Scenario: Bearing should stay constant when zig-zagging
         Given the node map
--- a/features/testbot/distance_matrix.feature
+++ b/features/testbot/distance_matrix.feature
@@ -589,12 +589,12 @@ Feature: Basic Distance Matrix
 
         When I request a travel distance matrix I should get
             |   | a      | b     | c      | d      | e      | f      |
-            | a | 0      | 100   | 300    | 650    | 1934.5 | 1534.6 |
-            | b | 760.6  | 0     | 200    | 550.1  | 1834.6 | 1434.6 |
-            | c | 560.6  | 660.5 | 0      | 350    | 1634.6 | 1234.6 |
-            | d | 1484.6 | 1584.5| 1784.5 | 0      | 1284.5 | 884.6  |
-            | e | 200    | 300   | 500    | 710.6  | 0      | 1595.2 |
-            | f | 600    | 699.9 | 899.9  | 1110.5 | 399.9  | 0      |
+            | a | 0      | 100   | 300    | 650    | 660.5  | 1534.6 |
+            | b | 760.6  | 0     | 200    | 550.1  | 560.6  | 1434.6 |
+            | c | 560.6  | 660.5 | 0      | 350    | 360.5  | 1234.6 |
+            | d | 1484.6 | 1584.5| 1645.1 | 0      | 1284.5 | 884.6  |
+            | e | 200    | 300   | 360.5  | 710.6  | 0      | 1595.2 |
+            | f | 600    | 699.9 | 760.5  | 884.6  | 399.9  | 0      |
 
 
     Scenario: Testbot - Filling in noroutes with estimates (defaults to input coordinate location)
@@ -720,11 +720,10 @@ Feature: Basic Distance Matrix
             | 1    | 2  | abcdef,fghij,fghij | 1000.1m    |
             | 1    | 3  | abcdef,fghij,fghij | 1400.1m    |
             | 2    | 3  | fghij,fghij        | 400m       |
-            
+
 
         When I request a travel distance matrix I should get
-            |   |   1    | 2      | 3      |   
+            |   |   1    | 2      | 3      |
             | 1 |   0    | 1000.1 | 1400.1 |
             | 2 | 1000.1 | 0      | 400    |
             | 3 | 1400.1 | 400    | 0      |
-
--- a/features/testbot/duration_matrix.feature
+++ b/features/testbot/duration_matrix.feature
@@ -491,18 +491,18 @@ Feature: Basic Duration Matrix
 
         When I route I should get
           | from | to | route | distance | time | weight |
-          | a    | c  | ac,ac | 200m     | 5s   |      5 |
+          | a    | c  | ac,ac | 200m     | 1s   |      1 |
 
         When I request a travel time matrix I should get
-          |   | a | b | c |  d |
-          | a | 0 | 1 | 5 | 10 |
+          |   | a | b | c | d |
+          | a | 0 | 1 | 1 | 6 |
 
         When I request a travel time matrix I should get
           |   |  a |
           | a |  0 |
           | b |  1 |
-          | c | 15 |
-          | d | 10 |
+          | c |  1 |
+          | d |  6 |
 
     Scenario: Testbot - OneToMany vs ManyToOne
         Given the node map
--- a/features/testbot/snap_intersection.feature
+++ b/features/testbot/snap_intersection.feature
@@ -1,0 +1,629 @@
+Feature: Snapping at intersections
+
+    Background:
+        # Use turnbot so that we can validate when we are
+        # snapping to one of many potential candidate ways
+        Given the profile "turnbot"
+
+    # https://github.com/Project-OSRM/osrm-backend/issues/4465
+    Scenario: Snapping source to intersection with one-way roads
+        Given the node map
+            """
+            a   e   c
+              \ | /
+                d
+
+                1
+            """
+
+        And the ways
+            | nodes | oneway |
+            | da    | yes    |
+            | dc    | yes    |
+            | de    | yes    |
+
+
+        When I route I should get
+            | from | to | route  | time   |
+            | 1    | e  | de,de  | 20s    |
+            | 1    | a  | da,da  | 28.3s  |
+            | 1    | c  | dc,dc  | 28.3s  |
+
+        When I request a travel time matrix I should get
+            |   | a        | c        | e   |
+            | 1 | 28.3     | 28.3     | 20  |
+
+
+    Scenario: Snapping destination to intersection with one-way roads
+        Given the node map
+            """
+            a   e   c
+              \ | /
+                d
+
+                1
+            """
+
+        And the ways
+            | nodes | oneway |
+            | da    | -1     |
+            | dc    | -1     |
+            | de    | -1     |
+
+
+        When I route I should get
+            | from | to | route  | time   |
+            | e    | 1  | de,de  | 20s    |
+            | a    | 1  | da,da  | 28.3s  |
+            | c    | 1  | dc,dc  | 28.3s  |
+
+        When I request a travel time matrix I should get
+            |   | 1      |
+            | a | 28.3   |
+            | c | 28.3   |
+            | e | 20     |
+
+
+    Scenario: Snapping to intersection with bi-directional roads
+        Given the node map
+            """
+            a   e
+            | /
+            d---c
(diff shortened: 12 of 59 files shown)


danpat:
This will fix https://github.com/Project-OSRM/osrm-backend/issues/4465

@mjjbell I think you don't need to consider the `/match` service here - it already returns "all edges within radius N" and adds them to the Hidden Markov Model to find the best path.

Amazing contribution, thank you, this has been an issue for a very long time.

mjjbell:
# Performance Analysis

## Preparation

### OSRM
I used [germany-latest.osm.pbf](https://download.geofabrik.de/europe/germany.html) as input,  `osrm-extract` with the default car profile and `osrm-contract` to route using the CH algorithm (I plan to do an MLD analysis later).

###  Test Coordinates
I created a collection of test coordinates for use as request parameters in the performance tests.
Using the same `germany-latest.osm.pbf`, all nodes are processed with [libosmium](https://github.com/osmcode/libosmium) to ensure we get the same coordinate truncation as used by OSRM data files - this enables us to test out the snapping behaviour at intersections. Nodes are bucketed by vertex degree.

We then randomly generate waypoint lists (represented by the truncated coordinates of the selected nodes). For testing, we create five sets of waypoint lists to test performance against a variety of input sizes.
Given we are interested in how the snapping behaviour changes the results when we have multiple ways to chose from, most of the waypoint lists are filtered to include nodes with degree >= 3.

| Set      |  Number of waypoint lists |  Number nodes in each waypoint list | Node properties | 
| ----------- | ----------- | ----------- | ----------- | 
| 1 |  20408 | 2     |                                           |
| 2 | 19291  | 2     | All nodes have degree >= 3 |
| 3 | 17387  | 11   | All nodes have  degree >= 3|
| 4 | 7557    | 101 | All nodes have  degree >= 3|
| 5 | 762      | 1001 | All nodes have  degree >= 3|

The coordinate sets are available in this [gist](https://gist.github.com/mjjbell/77dc41810a0972c8892f41e9223923c2)

###  Test Requests

The test coordinates are used to run the following requests via libosrm:

- Route
- Route with `continue_straight=false`
- Table
- Trip

In the case of Route and Trip requests, the waypoint lists act as waypoints in the returned route.
For Table requests, a waypoint list acts as both source and target locations, creating an NxN table result.

Tests are performed using libosrm versions of `Project-OSRM:master` at commit 9d81eb3 and `mjjbell:mbell/snap_candidates`

The duration and route weight returned for each request is recorded and used to compare the performance of both branches.
N.B. for the purposes of comparing weights, any absolute difference of `<=` 0.2 is treated as the same weight - this is to accommodate for insignificant movements in the snapped locations between the branches.

mjjbell:
# Results - Route
## Coordinate set 1
### Duration
_route - coordinate set #1

Percentile |  master (μs) | snap_candidates(μs) | diff(μs) | % diff
--  | -- | -- | -- | --
0  | 189 | 201 | 12 | 6.35%
10  | 1372 | 1389 | 17 | 1.24%
20  | 1714 | 1729 | 15 | 0.88%
30  | 1978 | 1995 | 17 | 0.86%
40  | 2228 | 2244 | 16 | 0.72%
50  | 2460 | 2481 | 21 | 0.85%
60  | 2682 | 2702 | 20 | 0.75%
70  | 2906 | 2929 | 23 | 0.79%
80  | 3164 | 3183 | 19 | 0.60%
90  | 3553 | 3547 | -6 | -0.17%
99  | 4575 | 4590 | 15 | 0.33%
100  | 13183 | 13434 | 251 | 1.90%

### Smallest Route Weight
master | same | snap_candidates
-- | -- | --
0 | 20339 | 69

## Coordinate set 2
### Duration
_route - coordinate set #2


Percentile | master (μs) | snap_candidates(μs) | diff(μs)  | % diff
-- | -- | -- | -- | --
0  | 218 | 229 | 11 | 5.05%
10 | 1291 | 1313 | 22 | 1.70%
20 | 1606 | 1628 | 22 | 1.37%
30 | 1852 | 1879 | 27 | 1.46%
40 | 2095 | 2118 | 23 | 1.10%
50 | 2315 | 2334 | 19 | 0.82%
60 | 2529 | 2549 | 20 | 0.79%
70 | 2747 | 2772 | 25 | 0.91%
80 | 3001 | 3019 | 18 | 0.60%
90 | 3378 | 3362 | -16 | -0.47%
99 | 4371 | 4302 | -69 | -1.58%
100 | 16385 | 16270 | -115 | -0.70%

### Smallest Route Weight
master | same | snap_candidates
-- | -- | --
0 |  18985 | 234

## Coordinate set 3
### Duration
_route - coordinate set #3

Percentile | master (μs) | snap_candidates (μs) | diff (μs) | % diff
 -- | -- | -- | -- | --
0 | 1187 | 1251 | 64 | 5.39%
10 | 20472 | 20762 | 290 | 1.42%
20 | 21744 | 21951 | 207 | 0.95%
30 | 22684 | 22861 | 177 | 0.78%
40 | 23538 | 23694 | 156 | 0.66%
50 | 24318 | 24477 | 159 | 0.65%
60 | 25114 | 25316 | 202 | 0.80%
70 | 26017 | 26206 | 189 | 0.73%
80 | 27132 | 27352 | 220 | 0.81%
90 | 28843 | 28916 | 73 | 0.25%
99 | 69394 | 65226 | -4168 | -6.01%
100 | 82659 | 81083 | -1576 | -1.91%

### Smallest Route Weight
master | same | snap_candidates
-- | -- | --
0 |  16381 | 1006

## Coordinate set 4
### Duration
_route - coordinate set #4

Percentile | master (μs) | snap_candidates (μs) | diff (μs) | % diff
-- | -- | -- | -- | --
0 | 5977 | 5838 | -139 | -2.33%
10 | 218193 | 221424 | 3231 | 1.48%
20 | 222619 | 225899 | 3280 | 1.47%
30 | 225791 | 228930 | 3139 | 1.39%
40 | 228472 | 231357 | 2885 | 1.26%
50 | 231110 | 233607 | 2497 | 1.08%
60 | 233850 | 236019 | 2169 | 0.93%
70 | 237234 | 238720 | 1486 | 0.63%
80 | 242820 | 241960 | -860 | -0.35%
90 | 266953 | 246876 | -20077 | -7.52%
99 | 287465 | 282789 | -4676 | -1.63%
100 | 348364 | 361102 | 12738 | 3.66%

### Smallest Route Weight
master | same | snap_candidates
-- | -- | --
0 |  4562 | 2995

Reduction in route weight, master to snap_candidates(1)


## Coordinate set 5
### Duration
_route - coordinate set #5


Percentile | master (μs) | snap_candidates (μs) | diff (μs) | % diff
-- | -- | -- | -- | --
0 | 66182 | 65036 | -1146 | -1.73%
10 | 676659 | 679042 | 2383 | 0.35%
20 | 2250095 | 2271579 | 21484 | 0.95%
30 | 2268765 | 2287675 | 18910 | 0.83%
40 | 2280586 | 2299932 | 19346 | 0.85%
50 | 2290470 | 2310654 | 20184 | 0.88%
60 | 2304508 | 2324956 | 20448 | 0.89%
70 | 2320816 | 2343423 | 22607 | 0.97%
80 | 2347262 | 2370331 | 23069 | 0.98%
90 | 2375039 | 2405981 | 30942 | 1.30%
99 | 2452844 | 2463975 | 11131 | 0.45%
100 | 2623389 | 2523901 | -99488 | -3.79%

### Smallest Route Weight
master | same | snap_candidates
-- | -- | --
0 |  86 | 676

Reduction in route weight, master to snap_candidates

mjjbell:
# Results - Route `continue_straight=false`

We start the results from coordinate set 3, as ``continue_straight=false` will have no effect on waypoint lists of length two.

## Coordinate set 3
### Duration
_route continue_straight=false - coordinate set #3


Percentile | master (μs) | snap_candidates (μs) | diff (μs) | % diff
-- | -- | -- | -- | --
0 | 2223 | 2299 | 76 | 3.42%
10 | 18161 | 18016 | -145 | -0.80%
20 | 19383 | 19090 | -293 | -1.51%
30 | 20340 | 19963 | -377 | -1.85%
40 | 21160 | 20749 | -411 | -1.94%
50 | 21929 | 21511 | -418 | -1.91%
60 | 22739 | 22281 | -458 | -2.01%
70 | 23601 | 23127 | -474 | -2.01%
80 | 24658 | 24142 | -516 | -2.09%
90 | 26242 | 25603 | -639 | -2.44%
99 | 65394 | 29350 | -36044 | -55.12%
100 | 78483 | 78273 | -210 | -0.27%

### Smallest Route Weight
master | same | snap_candidates
-- | -- | --
0 |  15990 | 1397

## Coordinate set 4
### Duration
_route continue_straight=false - coordinate set #4

Percentile | master (μs) | snap_candidates (μs) | diff (μs) | % diff
-- | -- | -- | -- | --
0 | 7559 | 7843 | 284 | 3.76%
10 | 190380 | 193361 | 2981 | 1.57%
20 | 194161 | 197038 | 2877 | 1.48%
30 | 196893 | 199942 | 3049 | 1.55%
40 | 199394 | 202247 | 2853 | 1.43%
50 | 201640 | 204357 | 2717 | 1.35%
60 | 203800 | 206534 | 2734 | 1.34%
70 | 206402 | 208995 | 2593 | 1.26%
80 | 209833 | 211817 | 1984 | 0.95%
90 | 215237 | 216509 | 1272 | 0.59%
99 | 250923 | 241023 | -9900 | -3.95%
100 | 323250 | 326410 | 3160 | 0.98%

### Smallest Route Weight
master | same | snap_candidates
-- | -- | --
0 |  3465 | 4092

Reduction in route weight, master to snap_candidates(2)

## Coordinate set 5
### Duration
_route continue_straight=false - coordinate set #5

Percentile | master (μs) | snap_candidates (μs) | diff (μs) | % diff
-- | -- | -- | -- | --
0 | 74904 | 63922 | -10982 | -14.66%
10 | 1953409 | 1993940 | 40531 | 2.07%
20 | 1971011 | 2010467 | 39456 | 2.00%
30 | 1983935 | 2024351 | 40416 | 2.04%
40 | 1992464 | 2037172 | 44708 | 2.24%
50 | 2003905 | 2057112 | 53207 | 2.66%
60 | 2017461 | 2074207 | 56746 | 2.81%
70 | 2035575 | 2091534 | 55959 | 2.75%
80 | 2053559 | 2106061 | 52502 | 2.56%
90 | 2072882 | 2122799 | 49917 | 2.41%
99 | 2112569 | 2162705 | 50136 | 2.37%
100 | 2414390 | 2423875 | 9485 | 0.39%


### Smallest Route Weight
master | same | snap_candidates
-- | -- | --
0 | 15 | 747

Reduction in route weight, master to snap_candidates(3)

mjjbell:
# Results - Table

N.B. Here we are not able to compare result weights, as the table response does not contain them. Instead we compare durations. In some cases we do get shorter duration results with `master`. Upon inspection of these results, `snap_candidates` results have smaller weight. See the following illustrative example:

`master`
before_roundabout

`snap_candidates`
after_roundabout

## Coordinate set 1
### Duration
_table - coordinate set #1


Percentile | master (μs) | snap_candidates (μs) | diff (μs) | % diff
-- | -- | -- | -- | --
0 | 459 | 384 | -75 | -16.34%
10 | 786 | 771 | -15 | -1.91%
20 | 839 | 827 | -12 | -1.43%
30 | 878 | 870 | -8 | -0.91%
40 | 913 | 906 | -7 | -0.77%
50 | 946 | 941 | -5 | -0.53%
60 | 980 | 978 | -2 | -0.20%
70 | 1018 | 1016 | -2 | -0.20%
80 | 1065 | 1063 | -2 | -0.19%
90 | 1135 | 1134 | -1 | -0.09%
99 | 1367 | 1324 | -43 | -3.15%
100 | 19759 | 14208 | -5551 | -28.09%

### Smallest Route Duration
master | same | snap_candidates
-- | -- | --
3 |  81494 | 135

## Coordinate set 2
### Duration
_table - coordinate set #2

Percentile | master (μs) | snap_candidates (μs) | diff (μs) | % diff
-- | -- | -- | -- | --
0 | 379 | 384 | 5 | 1.32%
10 | 763 | 743 | -20 | -2.62%
20 | 818 | 800 | -18 | -2.20%
30 | 858 | 841 | -17 | -1.98%
40 | 892 | 878 | -14 | -1.57%
50 | 925 | 913 | -12 | -1.30%
60 | 960 | 949 | -11 | -1.15%
70 | 997 | 987 | -10 | -1.00%
80 | 1042 | 1035 | -7 | -0.67%
90 | 1109 | 1107 | -2 | -0.18%
99 | 1336 | 1354 | 18 | 1.35%
100 | 18832 | 18663 | -169 | -0.90%

### Smallest Route Duration
master | same | snap_candidates
-- | -- | --
1 | 76414 | 461

## Coordinate set 3
### Duration
_table - coordinate set #3

Percentile | master (μs) | snap_candidates (μs) | diff (μs) | % diff
-- | -- | -- | -- | --
0 | 3868 | 3713 | -155 | -4.01%
10 | 4799 | 4694 | -105 | -2.19%
20 | 4939 | 4840 | -99 | -2.00%
30 | 5039 | 4951 | -88 | -1.75%
40 | 5132 | 5040 | -92 | -1.79%
50 | 5215 | 5125 | -90 | -1.73%
60 | 5299 | 5212 | -87 | -1.64%
70 | 5395 | 5306 | -89 | -1.65%
80 | 5505 | 5417 | -88 | -1.60%
90 | 5676 | 5577 | -99 | -1.74%
99 | 6410 | 6047 | -363 | -5.66%
100 | 55243 | 55735 | 492 | 0.89%


### Smallest Route Duration
master | same | snap_candidates
-- | -- | --
81 | 2080444 | 23302

## Coordinate set 4
### Duration
_table - coordinate set #4


Percentile | master (μs) | snap_candidates (μs) | diff (μs) | % diff
-- | -- | -- | -- | --
0 | 45902 | 45955 | 53 | 0.12%
10 | 49471 | 48684 | -787 | -1.59%
20 | 49967 | 49205 | -762 | -1.53%
30 | 50355 | 49591 | -764 | -1.52%
40 | 50697 | 49903 | -794 | -1.57%
50 | 51029 | 50210 | -819 | -1.60%
60 | 51354 | 50536 | -818 | -1.59%
70 | 51677 | 50886 | -791 | -1.53%
80 | 52\068 | 51273 | -795 | -1.53%
90 | 52783 | 51873 | -910 | -1.72%
99 | 93967 | 53723 | -40244 | -42.83%
100 | 103247 | 101436 | -1811 | -1.75%

### Smallest Route Duration
master | same | snap_candidates
-- | -- | --
3944 | 76182571 | 902442

## Coordinate set 5
### Duration
_table - coordinate set #5

Percentile | master (μs) | snap_candidates (μs) | diff (μs) | % diff
-- | -- | -- | -- | --
0 | 675348 | 607830 | -67518 | -10.00%
10 | 682147 | 677855 | -4292 | -0.63%
20 | 686247 | 681076 | -5171 | -0.75%
30 | 689809 | 683411 | -6398 | -0.93%
40 | 692327 | 685563 | -6764 | -0.98%
50 | 696535 | 688071 | -8464 | -1.22%
60 | 710840 | 691110 | -19730 | -2.78%
70 | 724902 | 695422 | -29480 | -4.07%
80 | 728627 | 722568 | -6059 | -0.83%
90 | 733552 | 729589 | -3963 | -0.54%
99 | 743412 | 744403 | 991 | 0.13%
100 | 886189 | 763301 | -122888 | -13.87%

### Smallest Route Duration
master | same | snap_candidates
-- | -- | --
39196 | 754472544 | 9013022

mjjbell:
# Results - Trip

Running `Trip` on coordinate sets 1 and 2 is effectively finding the shortest route between two endpoints in either direction.
For coordinate sets 3,4 and 5 we begin to see some results having a significantly smaller weight when run on `master`. This is to be expected, given that these routes are just approximations of the optimal route, so it's possible that `snap_candidates` is finding worse approximations in some cases.

## Coordinate set 1
### Duration
_trip - coordinate set #1

Percentile | master (μs) | snap_candidates (μs) | diff (μs) | % diff
-- | -- | -- | -- | --
0 | 846 | 776 | -70 | -8.27%
10 | 3353 | 3354 | 1 | 0.03%
20 | 4047 | 4025 | -22 | -0.54%
30 | 4581 | 4560 | -21 | -0.46%
40 | 5097 | 5063 | -34 | -0.67%
50 | 5566 | 5544 | -22 | -0.40%
60 | 6027 | 5985 | -42 | -0.70%
70 | 6495 | 6459 | -36 | -0.55%
80 | 7039 | 7021 | -18 | -0.26%
90 | 7828 | 7840 | 12 | 0.15%
99 | 9844 | 10575 | 731 | 7.43%
100 | 19114 | 59637 | 40523 | 212.01%

### Smallest Route Weight
master | same | snap_candidates
-- | -- | --
0 | 20335 | 73


## Coordinate set 2
### Duration
_trip - coordinate set #2

Percentile | master (μs) | snap_candidates (μs) | diff (μs) | % diff
-- | -- | -- | -- | --
0 | 980 | 1004 | 24 | 2.45%
10 | 3172 | 3204 | 32 | 1.01%
20 | 3818 | 3890 | 72 | 1.89%
30 | 4332 | 4444 | 112 | 2.59%
40 | 4824 | 5065 | 241 | 5.00%
50 | 5281 | 5607 | 326 | 6.17%
60 | 5718 | 6076 | 358 | 6.26%
70 | 6165 | 6567 | 402 | 6.52%
80 | 6703 | 7117 | 414 | 6.18%
90 | 7469 | 7867 | 398 | 5.33%
99 | 9486 | 10655 | 1169 | 12.32%
100 |  23285 | 60152 | 36867 | 158.33%


### Smallest Route Weight
master | same | snap_candidates
-- | -- | --
0 | 18923 | 296


## Coordinate set 3
### Duration
_trip - coordinate set #3


Percentile | master (μs) | snap_candidates (μs) | diff (μs) | % diff
-- | -- | -- | -- | --
0 | 4845 | 4389 | -456 | -9.41%
10 | 17714 | 18073 | 359 | 2.03%
20 | 18394 | 18762 | 368 | 2.00%
30 | 18916 | 19283 | 367 | 1.94%
40 | 19391 | 19749 | 358 | 1.85%
50 | 19840 | 20188 | 348 | 1.75%
60 | 20305 | 20619 | 314 | 1.55%
70 | 20818 | 21123 | 305 | 1.47%
80 | 21432 | 21709 | 277 | 1.29%
90 | 22418 | 22552 | 134 | 0.60%
99 | 50839 | 24958 | -25881 | -50.91%
100 | 72122 | 74177 | 2055 | 2.85%

### Smallest Route Weight
master | same | snap_candidates
-- | -- | --
4 | 15913 | 1470

Reduction in route weight, master to snap_candidates(4)


## Coordinate set 4
### Duration
_trip - coordinate set #4

Percentile | master (μs) | snap_candidates (μs) | diff (μs) | % diff
-- | -- | -- | -- | --
0 | 50307 | 48584 | -1723 | -3.42%
10 | 110669 | 117229 | 6560 | 5.93%
20 | 111964 | 118632 | 6668 | 5.96%
30 | 113020 | 119616 | 6596 | 5.84%
40 | 113879 | 120459 | 6580 | 5.78%
50 | 114741 | 121297 | 6556 | 5.71%
60 | 115596 | 122210 | 6614 | 5.72%
70 | 116532 | 123186 | 6654 | 5.71%
80 | 117755 | 124439 | 6684 | 5.68%
90 | 119783 | 126681 | 6898 | 5.76%
99 | 162562 | 170316 | 7754 | 4.77%
100 | 223655 | 227874 | 4219 | 1.89%

### Smallest Route Weight
master | same | snap_candidates
-- | -- | --
69 | 3370 | 4118

Reduction in route weight, master to snap_candidates(5)


## Coordinate set 5
### Duration
_trip - coordinate set #5

Percentile | master (μs) | snap_candidates (μs) | diff (μs) | % diff
-- | -- | -- | -- | --
0 | 653879 | 625922 | -27957 | -4.28%
10 | 1474263 | 1580410 | 106147 | 7.20%
20 | 1480059 | 1587731 | 107672 | 7.27%
30 | 1483672 | 1593048 | 109376 | 7.37%
40 | 1487356 | 1596728 | 109372 | 7.35%
50 | 1491191 | 1602011 | 110820 | 7.43%
60 | 1496461 | 1609048 | 112587 | 7.52%
70 | 1506838 | 1621422 | 114584 | 7.60%
80 | 1520878 | 1643598 | 122720 | 8.07%
90 | 1558687 | 1673506 | 114819 | 7.37%
99 | 1576503 | 1699391 | 122888 | 7.79%
100 | 1753056 | 1799930 | 46874 | 2.67%

### Smallest Route Weight
master | same | snap_candidates
-- | -- | --
77 | 15 | 670

Reduction in route weight, master to snap_candidates(6)

mjjbell:
## Review

In summary, with this change we are able to find better results, especially when the inputs snap to locations with multiple ways. In the cases where we can compare optimal route weights, it does not produce any worse results.

The performance effect on request speed appears to differ on each service.
`route` has a small performance penalty.
`table` has no performance penalty.
`trip` has a significant slowdown for large numbers of waypoints. This appears to be disproportionately large and more than the expected penalty of compounding the internal `route` and `table` requests it performs. Further investigation is required.

jcoupey:
Thanks @mjjbell for this comprehensive impact study!

> For coordinate sets 3,4 and 5 we begin to see some results having a significantly smaller weight when run on master

Things are a bit fuzzy for the `trip` plugin and hard to evaluate based on weight only: my understanding is that the `trip` plugin does the TSP optimization based on durations, but the underlying `durations` matrix is computed on a weight-based objective (see #5930 for a discussion around this behavior).

> Trip has a significant slowdown for large numbers of waypoints. This appears to be disproportionately large and more than the expected penalty of compounding the internal Route and table requests it performs.

The expected changes are probably rather on the solving side than on internal `route` and `table` overhead. As you mentioned, only approximation solution are provided, and `master` and `snap_candidates` are likely to follow quite different solving paths, leading to varying computing times. As a result the subset of the "X% fastest requests" may be significantly different between the two branches.

mjjbell:
> Things are a bit fuzzy for the trip plugin and hard to evaluate based on weight only: my understanding is that the trip plugin does the TSP optimization based on durations, but the underlying durations matrix is computed on a weight-based objective (see #5930 for a discussion around this behavior).

Ah yes, that actually does explain it. Thanks!

mjjbell:
```
    Scenario: Middle Island Over Bridge
        Given the node map
            """
              a
              |
             .b.
            c   h
            |   |
            |   |
            1   2
            |   |
            d   g
             'e'
              |
              f
            """

        And the ways
            | nodes | name   | oneway |
            | ab    | road   | no     |
            | ef    | road   | no     |
            | bc    | road   | yes    |
            | cd    | bridge | yes    |
            | de    | road   | yes    |
            | eg    | road   | yes    |
            | gh    | bridge | yes    |
            | hb    | road   | yes    |

        When I route I should get
            | waypoints | turns         | route       | intersections						  |
            | a,f       | depart,arrive | road,road   | true:180,false:0 true:180,false:0 true:180;true:0 |
            | c,f       | depart,arrive | bridge,road | true:180,false:0 true:180;true:0		  |
            | 1,f       | depart,arrive | bridge,road | true:180,false:0 true:180;true:0                  |
            | f,a       | depart,arrive | road,road   | true:0,true:0 false:180,true:0 false:180;true:180 |
            | g,a       | depart,arrive | bridge,road | true:0,true:0 false:180;true:180                  |
            | 2,a       | depart,arrive | bridge,road | true:0,true:0 false:180;true:180                  |
```

When routing from `c` to `f` in this [merging segregated roads test graph](https://github.com/Project-OSRM/osrm-backend/blob/8b52c6c7ac6c4e7489625c0265f89fc1e88520f5/features/guidance/merge-segregated-roads.feature#L359), the test would fail only for contraction hierarchy search on OS X, as it finds the named route `road, road` instead of `bridge, road`. I investigated this as it seemed a bit puzzling.

#### Segregated Road Merging

With this PR there are now two possible start candidates on a `c` to `f` route - `bc` and `cd`. 
Linux/Ubuntu is selecting `cd`, and OS X is selecting `bc`.

Looking at the merging logic for combining segregated roads:
- The latter name of two connected edges is used. Hence, all the roads are merged and we're left with the name of `ef` as the second value in the route.
- The first value comes from first edge of the route, which is never merged, Therefore, for Linux `cd` is `bridge`, so we get `bridge,road`. For OS X, `bc` is `road`, so after merging we are left with `road,road`.

This explains why the merging logic creates different results given the selected start edge, but doesn’t explain why different start edges are selected on different platforms.

#### Contraction Order

I pulled out the contraction hierarchy forward and backward graphs from this test for both platforms.

Linux
forward_backward


OS X
forward_backward




Clearly the ordering is different. My understanding of the contraction logic is that when there are no criteria that indicate one node has higher rank than the other, one of the nodes is (deterministically) chosen at random.

I discovered this when looking at an edge case from a more trivial example, which I will discuss in a separate issue, but I expect this difference in randomness between the platforms is what creates the different contraction order here too.

CH search chooses the first path it finds when there are multiple candidates of equal weight.
If we follow the search path for finding the identical weight shortest paths `bc -> cd -> de -> ef`, and `cd -> de -> ef`, we will see that for Linux, `cd` as a starting point is found first, whereas for OS X, we will find `bc` first.

Linux
forward_backward_annotated


OS X
forward_backward_annotated


So this explains the difference. To fix this, I [explicitly added a numbered input location](https://github.com/Project-OSRM/osrm-backend/blob/d34f21e0af3fa3badc0cb16c41b00e97a06828c8/features/guidance/merge-segregated-roads.feature#L335) so that it only snaps to one edge. Other tests also required explicit inputs to continue to pass - it’s possible some of these will have displayed these platform differences too.

I expect this is one of many subtle differences that can occur from snapping to multiple candidates.

features/car/restrictions.feature
```diff
@@ -438,7 +438,7 @@ Feature: Car - Turn restrictions
         When I route I should get
             | from | to | route                               |
             | e    | f  | ae,xa,bx,fb,fb                      |
-            | c    | f  | dc,da,ae,ge,hg,hg,ge,ae,xa,bx,fb,fb |
+            | 1    | f  | dc,da,ae,ge,hg,hg,ge,ae,xa,bx,fb,fb |
```
mjjbell, line 441:
```
| restriction | cb       | bx     | b        | only_right_turn  |
```
This turn restriction routes you into a one-way road going in the other direction. I'm not sure if this was intended.
In any case, I added a numbered input location to ensure it snaps to the same `dc` as previously.

features/testbot/traffic_turn_penalties.feature
```diff
@@ -47,7 +47,7 @@ Feature: Traffic - turn penalties applied to turn onto which a phantom node snap
             | 1    | e  | ab,be,be | 36 km/h  | 30s +-1 |
             | b    | f  | bc,cf,cf | 36 km/h  | 40s +-1 |
             | 2    | f  | bc,cf,cf | 36 km/h  | 30s +-1 |
-            | c    | g  | cd,dg,dg | 144 km/h | 10s +-1 |
+            | c    | g  | cd,dg,dg | 72 km/h  | 20s +-1 |
```
mjjbell, line 50: The use of negative turn penalties is the cause of the increase in speed/time here.

Previously, the start would snap to `cf` (which would be combined with `cd` in the route result annotation). 
Now with multiple intersection candidates to choose from, the route does actually start from `cd`. 

There is [special case logic](https://github.com/Project-OSRM/osrm-backend/blob/919fe74c4064cd3de05b73be873b35c2d75a5052/include/engine/routing_algorithms/routing_base.hpp#L322) for preventing negative weights/durations on the first turn. Despite the intention of this test, it didn't apply previously when starting at `cf` as the negative weight was on the second turn. 
This leads to the multiple candidate route actually having larger weight/duration than previously.

I can see historical [discussions](https://github.com/Project-OSRM/osrm-backend/issues/3683) about the value of negative turn penalty weights, but I'm not sure if that also makes sense for negative durations too. In any case, I've just updated the expected value to reflect this first turn clamping logic.

mjjbell:
Also worth mentioning - this PR will enable a fix for snapping target locations to via ways used in turn restrictions, discussed [here](https://github.com/Project-OSRM/osrm-backend/pull/5907#issuecomment-748648815).

The fix is quite simple once this change is made: https://github.com/mjjbell/osrm-backend/commit/3771e99f74d609735e61ca4fd81424b9d0fc663a

mjjbell:
> (I plan to do an MLD analysis later)

I've completed the same analysis for the MLD algorithm. The relative performance change is very similar to the CH result detailed above, so to save some scrolling I've put the MLD results in [this gist](https://gist.github.com/mjjbell/85e9f4875935bf621a6c2ba341e96c02).

zhysong:
Route detour when destination is on the via way of multiple via-ways complex restriction。
It should go through the red route.
image

SiarheiFedartsou:
Hi @mjjbell 
Could you please summarize what is left to do to consider this PR as ready to merge?

mjjbell:
> Hi @mjjbell Could you please summarize what is left to do to consider this PR as ready to merge?

It's ready.
Just needs to be reviewed and rebased.

mjjbell:
> > It's ready.
> > Just needs to be reviewed and rebased.
> 
> Cool, I am not very familiar with this part, but I will try to do a review in the nearest days.
> 
> Do you think we will need someone from OSRM maintainers to approve it too? I guess it can be difficult taking into account that you are the only one active recently :)

Yes it will need a maintainer to approve for it to be mergeable. Repo ownership still resides with (ex-) Mapbox folk, so I can't change the situation myself. 
Related: #5209

SiarheiFedartsou:
Hey @mjjbell 

> EDIT: This is a breaking change in the request parameters. hints are now vector<vector<hint>>

Do you mean breaking change from libosrm point of view? Can we may be leave BaseParameters::hints parameters as is and add new one(may be `BaseParameters::extended_hints` - not sure what is the best name here) and use new if it is non-empty and fallback to the old one in the case if not? 

In general I had a chance to just briefly look through PR and it seems it is LGTM, but I think I will need to spend a bit more time on review, but just want to ask do you have time to rebase it? Or I can help you with it if you want? The same question is about other your PRs which seems to be open for a while:
Screenshot 2022-08-21 at 18 22 50

include/engine/api/base_parameters.hpp
```diff
@@ -77,7 +77,7 @@ struct BaseParameters
     };
 
     std::vector<util::Coordinate> coordinates;
-    std::vector<boost::optional<Hint>> hints;
+    std::vector<std::vector<Hint>> hints;
```
SiarheiFedartsou, line 80: @mjjbell 
It is a breaking change you mentioned, right? Is it possible to support both old and new version of API? I mean do not touch existing field and just add `std::vector<std::vector<Hint>> extended_hints;`? I would even make it `std::vector<WaypointHints> extended_hints` and put this vector of hints to `WaypointHints` structure to be able to extend this structure in the future if needed without breaking changes.

mjjbell:
Thanks, I'll rebase. 
Yes it would it be a breaking change to libosrm, so I like the idea of having an extended hints field.

The previous plan included a proposal to remove hints (#4741) as it exposes internal data structures, so I think a lightweight backwards compatible solution makes sense for now, and we can reserve the breaking change for when we have a hints replacement.

include/engine/api/base_parameters.hpp
mjjbell: :+1: I like this idea. This would make it backwards compatible with existing API users getting the existing (sub-optimal) behaviour, new users get the improved behaviour. Most people who use the HTTP API (such as osrm-frontend) will get the improvement for free.

SiarheiFedartsou:
This `bugprone-macro-parentheses` looks too annoying tbh 🤔

mjjbell:
So having tried to implement an additional field for hints in the request and response fields, I can see it complicates a lot of the code.

What I have realised however, is both the HTTP and libosrm responses contain the base64 encoding of hints. 
For HTTP, the hint is passed back as a base64 request argument without any changes. 
For libosrm, it needs to returned as a `Hint` argument. This is only possible by calling `Hint::FromBase64`. 

Therefore, I've inverted the change slightly. `Hint`now contains a vector of `SegmentHints`, that represents the multiple waypoints that can represent the input location.

From the `libosrm` user's perspective, nothing needs to change, they still call `Hint::FromBase64` on the base64 response to get the `Hint` to pass into the next request. This will still work on old base64 values that the user may have stored from previous OSRM versions. 

If the user happens to edit some of the internals of `Hint` or serializes the value separately from the base64 representation, then this would be a breaking change. However, this is the fundamental problem mentioned in #4741. Previous changes to `PhantomCandidate` have not been reflected in a breaking API change, so I'm satisfied that anyone who's tweaking/serializing internals will not be assuming it is stable.

In summary, I think this is good to go now :+1:

include/engine/hint.hpp
```diff
@@ -57,16 +58,31 @@ struct Hint
...
+    friend std::ostream &operator<<(std::ostream &, const SegmentHint &);
 };
 
-static_assert(sizeof(Hint) == 80 + 4, "Hint is bigger than expected");
-constexpr std::size_t ENCODED_HINT_SIZE = 112;
-static_assert(ENCODED_HINT_SIZE / 4 * 3 >= sizeof(Hint),
-              "ENCODED_HINT_SIZE does not match size of Hint");
+// Hint represents the suggested segment positions that could be used
+// as the waypoint for a given input location
+struct Hint
```
SiarheiFedartsou, line 70: Not sure if it is not breaking change too - as I understand API of Hint class is part of public API too, BUT tbh I don’t think it is a big problem: even though someone uses it for something(I doubt it) it will be quite easy to migrate.

SiarheiFedartsou (approved):

include/engine/hint.hpp
SiarheiFedartsou: Sorry didn’t notice comment https://github.com/Project-OSRM/osrm-backend/pull/5953#issuecomment-1222933925

ntzm:
Thanks everyone for your work on this, this has simplified something I'm doing significantly :)

Issue #5325, opened 2018-12-20
Store `way_id` values for all edges


danpat:
For debugging, it's very handy to be able to quickly cross-reference OSM.  Currently, you have to do this manually using coordinates - get the location of the issue from OSRM, then browse to that location in OSM to get the way tag information.

I propose a new `osrm-extract` flag called `--save-way-ids`.  When enabled, this will create a new, optional datafile called `.osrm.edge_ways`.  At runtime, this file can be optionally `mmap`-ed, and data returned with `annotations=ways` and as a property on edges in the debug tiles.

Related: https://github.com/Project-OSRM/osrm-backend/issues/5104

MichalPP:
#4071 is related as well

nnseva:
Hi people, I would like to implement this feature, but I need to have some explanations. Unfortunately, there is no enough documentation about the internal structure of the program data.

For the very first, I need to understand what does the Edge mean? Is it just a pair of Nodes directly connected? Or it is some sequence of nodes in one way from one branch of ways to another? Or something else? When the EdgeID appears? Is there some data storage indexed by the Edge ID?

I am looking into the code of `osrm::engine::guidance::assembleGeometry()` and don't understand, how to get there the `EdgeID` of the correspondent edges.

1. I've tried to use the `FindEdge` method of the facade (providing it to the `BaseDataFacade` from the `AlgorithmDataFacade` using virtual functions), and passing there the same parameters as passed to the `facade.GetOSMNodeIDOfNode()` in the code (they all typed as `NodeID`), but got an assertion error inside a `FindEdge` implementation of the `StaticGraph`. What I did improperly?
2. I've noticed that although the `GetUncompressedForwardGeometry()` (and several other methods of the BaseDataFacade class) parameter is marked as an EdgeID, the actual parameter passed to this (and others) method in the code is the geometry ID. Such code inconsistency leads to a full misunderstanding of what does the Edge (and EdgeID) mean really.
3. I could use a totally separate `map<pair<OSMNodeID,OSMNodeID>,OSMWayID>` in the function, but I am not sure, whether I can use it in such a multithreaded environment, and also, when and how to create this map. Additionally, I am afraid that this is not an optimal solution.

danpat:
OSRM uses a [line graph](https://en.wikipedia.org/wiki/Line_graph) to represent the road network.  This means that in the code where you're looking, a node is a road, and an edge is a turn.

[...]

Issue #5879, opened 2020-11-10
Lite version of OSRM instance


KushnirDV:
Is it possible to disable some not used data to get more lite (less size at least 2 times) OSRM instance? 
We use only rute and nearest API services. The reasson is to boot two backend OSRM entities on our server. 
Will thankful for your answer.

akashihi:
I guess you are asking if you can decrease dataset size?

jcoupey:
Related to #5838 (scoped more specifically on potentially removing the guidance dataset).

danpat:
@KushnirDV if you use `steps=true` in your Route API response, then *all* the data is used.

There are some files generated during `osrm-extract` that can be deleted, but I don't have a handy list.  One way to get the actual list of used files is to use `strace` and look for `open()` syscalls to get all the filenames that are used by routing.

If the issue is *memory* consumption, then you can use the `--mmap` option to lazy-load data from disk.  If you perform a warmup routine with a typical set of queries, this can help you use just the minimum RAM needed for fast queries.

KushnirDV:
You are doing a realy good job.
I will compile OSRM version higher than 5.20 to be able to use --mmap and will start generetion a new routing.

The main reason to have two instance is possibility to switch bettwen routing with higway and without.
Thanks for fast and helpfull support.

Issue #5838, opened 2020-09-23
Remove guidance data to reduce RAM usage?


jcoupey:
I'm usually not interested in route information apart from duration, distance and geometry, so I'm always using OSRM with `steps=false`. My understanding is that the matching data used to generate the `steps` array in response (`RouteStep.name` and `RouteStep.ref` values among others) is still loaded in RAM. Depending on the extract size, this may amount to a lot of unused data.

First of all, I'd like to figure whether trying not to load this data is worth the trouble. I can see `osrm.names` is not that big, but `osrm.icd` is. Is there a way to evaluate the potential gain here?

SiarheiFedartsou:
Done in https://github.com/Project-OSRM/osrm-backend/pull/6045

jcoupey:
#6045 is indeed related as it does save some calculation on route path data but as stated there: *this PR is a first step towards making the loading of these datasets optional* and *making these changes unlocks the optional loading of data*.

Optionally loading guidance data (or other parts of the data preprocessed by OSRM) is probably easier to implement now but it's still not possible.

jcoupey:
While there has been no recent progress on this, I think it's worth keeping the ticket open as a feature request. It would be quite useful in many situations to be able to reduce memory requirements when some data is unused, and this would be easier to build now after #6666.

Issue #5218, opened 2018-10-02
Make some data optional


danpat:
It'd be nice if you didn't need to load all the OSRM data if you didn't need it.

For example, if you never need turn-by-turn guidance, then it should be possible to skip the turn-by-turn data generation (save some time), and not bother loading it (save some memory).  If you never need geometry (`&geometries=false`), then it shouldn't be required to load it.

Steps to getting this done are:

  1. Figure out which data structures various API parameters trigger traversal of (this could be done by looking at calls to the Datafacade interface)
  2. Make data loading optional for various only-used-sometimes data (including how to actually tell OSRM which data not to load).
  3. Modify datafacade interface to return a "data not loaded" error if a not-loaded is accessed
  4. Modify API interface to return a descriptive error if data isn't loaded

For a first version of this, we can limit it to not loading data into `osrm-routed`/`node-osrm`.  As a second phase, we could possibly skip the creation of some data altogether (e.g. turn-by-turn data if it's not going to be used).

jcoupey:
Maybe others can comment here but this sounds like an interesting feature to keep track of.

PR #5968, opened 2021-02-22
#5325 Store and show way ids

nnseva:
# Issue #5325

- Features:
  - ADDED: a new `ways` subsection in the `annotations` section containing way IDs along the found path, with sign meaning direction [#5325]
- API:
  - Fixed item type for `nodes` section of the `annotations` flatbuffers output format


include/engine/api/flatbuffers/route.fbs
```diff
@@ -9,7 +9,8 @@ table Annotation {
     distance: [uint];
     duration: [uint];
     datasources: [uint];
-    nodes: [uint];
+    nodes: [ulong];
+    ways: [long];
```
akashihi, line 13: Should it be a `ulong` too?

include/engine/api/route_api.hpp
```diff
@@ -503,7 +503,17 @@ class RouteAPI : public BaseAPI
                 nodes.emplace_back(static_cast<uint64_t>(node_id));
             }
         }
+        std::vector<int64_t> ways;
```
akashihi, line 506: Are ways ID signed in OSM?

include/engine/guidance/assemble_geometry.hpp
```diff
@@ -56,9 +56,11 @@ inline LegGeometry assembleGeometry(const datafacade::BaseDataFacade &facade,
         reversed_source ? source_node.reverse_segment_id.id : source_node.forward_segment_id.id;
     const auto source_geometry_id = facade.GetGeometryIndex(source_node_id).id;
     const auto source_geometry = facade.GetUncompressedForwardGeometry(source_geometry_id);
+    // const auto source_osm_way_ids = facade.GetUncompressedForwardWayIDs(source_geometry_id);
```
akashihi, line 59: Hmm?

include/engine/guidance/assemble_geometry.hpp
```diff
@@ -56,9 +56,11 @@ inline LegGeometry assembleGeometry(const datafacade::BaseDataFacade &facade,
...
 
     geometry.osm_node_ids.push_back(
         facade.GetOSMNodeIDOfNode(source_geometry(source_segment_start_coordinate)));
+    // geometry.osm_way_ids.push_back(source_osm_way_ids(source_segment_start_coordinate)));
```
akashihi, line 63: Same hmm :)

include/engine/api/flatbuffers/route.fbs
nnseva: @akashihi I've made returned Way IDs signed to add the way direction info as explained in comments and proposed documentation. The OSM Way ID is unsigned, but Way IDs in the annotation are signed: the negative value means opposite moving direction, while positive means moving forward along the way.

include/engine/api/route_api.hpp
nnseva: @akashihi The OSM Way ID is unsigned, but Way IDs in the annotation are signed: the negative value means opposite moving direction, while positive means moving forward along the way.

include/engine/guidance/assemble_geometry.hpp
nnseva: @akashihi Forget to remove, sorry

include/engine/guidance/assemble_geometry.hpp
nnseva: The same :)

akashihi (approved):

akashihi:
@nnseva Thank you for your contribution :) Let's wait for more opinions before merging.

I'm also curious, have you estimated, what is the dataset size change?

nnseva:
@akashihi 

> I'm also curious, have you estimated, what is the dataset size change?

The unpacked dataset size should be changed to 8*N bytes, where N is a number of segments in the segments file. The segments file will contain an additional section `/osm_ways` with the way IDs array corresponding to the segments array.

mjjbell (commented):
LGTM, some minor code comments.

Not sure if it matters for your use-case, but you'll currently get duplicates returned in the annotation if the way can't be fully compressed (e.g. one of the nodes is an intersection).

For example:
```
    Scenario: way annotations
        Given the query options
            | annotations | true |

        Given the node map
            """
            a--b--c
               |
               d
            """

        And the ways
            | nodes |
            | abc   |
            | bd    |


        When I route I should get
          | from | to | route    | 
          |    a |  c | abc,abc  |
```

The response will contain:

```
"ways": [ 5, 5 ],
```

Deduplication could be performed client-side as an alternative.


With regards to data size, you can get an idea of segment count by looking at the `R-tree segs` column from this [spreadsheet](https://docs.google.com/spreadsheets/d/1Q_XF0P-lVb2BWIlTzW60Izgko3o512ZiWF9pgVqE45s/edit#gid=3259062), as it should be almost identical. 

For some of the bigger map extracts, unpacked size will be:
- Germany ~250MB
- North America  ~2.3 GB
- Europe ~2.5 GB

I don't have numbers on hand for how much RAM is currently needed for routing on those extracts, it's probably adding a few %?

mjjbell:
> Not sure if it matters for your use-case, but you'll currently get duplicates returned in the annotation if the way can't be fully compressed (e.g. one of the nodes is an intersection).

In fact, we'll get this even if the way is fully compressed. If a way contains N edges, we'll get that way returned N times in a row in the annotation.

nnseva:
@mjjbell 
> "ways": [ 5, 5 ],

Sure, every pair of sequentially neighbor nodes will have a separate member in the ways array of the annotations:
```
    ...
    "nodes": [1, 2, 3],
    "ways": [5, 5]
    ...
```

(meaning that nodes 1, 2, 3 all in the way 5 and connected sequentially).

Such a duplication allows recognizing the path structure without additional analysis of the OSM structure. The same role plays an additional sign of the returned way ID - positive means that the path segment direction corresponds to the way direction, while negative means that the path direction is opposite to the way direction. I.e. the response for the opposite direction request with the same OSM data might look like:
```
   ...
   "nodes": [3, 2, 1],
   "ways": [-5, -5]
   ...
```

include/extractor/segment_data_container.hpp
```diff
@@ -90,6 +94,21 @@ template <storage::Ownership Ownership> class SegmentDataContainerImpl
...
+    {
+        const auto begin = osm_ways.begin() + index[id];
+        const auto end = osm_ways.begin() + index[id + 1] - 1;
+
+        return boost::make_iterator_range(begin, end);
+    }
+
+    auto GetReverseOSMWayIDs(const DirectionalGeometryID id)
+    {
+        // return boost::adaptors::reverse(GetForwardOSMWayIDs(id));
```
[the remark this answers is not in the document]
nnseva: sure, just some forgotten footprint of my investigations ...

include/extractor/segment_data_container.hpp
```diff
@@ -151,6 +170,21 @@ template <storage::Ownership Ownership> class SegmentDataContainerImpl
...
+    {
+        const auto begin = osm_ways.cbegin() + index[id];
+        const auto end = osm_ways.cbegin() + index[id + 1] - 1;
+
+        return boost::make_iterator_range(begin, end);
+    }
+
+    auto GetReverseOSMWayIDs(const DirectionalGeometryID id) const
+    {
+        // return boost::adaptors::reverse(GetForwardOSMWayIDs(id));
```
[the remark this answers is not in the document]
nnseva: agree

src/extractor/compressed_edge_container.cpp
```diff
@@ -297,16 +300,68 @@ unsigned CompressedEdgeContainer::ZipEdges(const EdgeID f_edge_id, const EdgeID
...
         segment_data->fwd_durations.emplace_back(fwd_node.duration);
         segment_data->rev_durations.emplace_back(rev_node.duration);
         segment_data->fwd_datasources.emplace_back(LUA_SOURCE);
         segment_data->rev_datasources.emplace_back(LUA_SOURCE);
+        prev_node_id = node_id;
     }
 
     const auto &last_node = forward_bucket.back();
+    auto node_id = last_node.node_id;
+    if (node_id != prev_node_id)
```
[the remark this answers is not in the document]
nnseva: Ok

src/extractor/node_based_graph_factory.cpp
```diff
@@ -93,7 +94,7 @@ void NodeBasedGraphFactory::Compress(ScriptingEnvironment &scripting_environment
                               compressed_edge_container);
 }
 
-void NodeBasedGraphFactory::CompressGeometry()
+void NodeBasedGraphFactory::CompressGeometry(OSMWayIDMap &osm_way_id_map)
```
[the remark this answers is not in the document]
nnseva: Ok

src/extractor/compressed_edge_container.cpp
```diff
@@ -262,7 +262,9 @@ void CompressedEdgeContainer::InitializeBothwayVector()
     segment_data->rev_datasources.reserve(m_compressed_oneway_geometries.size());
 }
 
-unsigned CompressedEdgeContainer::ZipEdges(const EdgeID f_edge_id, const EdgeID r_edge_id)
+unsigned CompressedEdgeContainer::ZipEdges(const EdgeID f_edge_id,
+                                           const EdgeID r_edge_id,
+                                           OSMWayIDMap &osm_way_id_map)
```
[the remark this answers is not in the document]
nnseva: Ok

jcoupey:
>I don't have numbers on hand for how much RAM is currently needed for routing on those extracts, it's probably adding a few %?

I just did an extract/contract pass on the latest germany pbf using `v5.24.0` and the default car profile with excludables disabled. Then `osrm-routed` uses about 3.3G of RAM at startup so that sounds like a ~7.5% increase in RAM usage in that case.

Given the impact here and the fact that RAM consumption has always been a touchy subject (often a bottleneck when using OSRM), maybe we could further investigate the impact? I can report some figures on a bigger extract such as Europe when the preprocessing is done.

Maybe a silly question (I did not look up the implementation): would it be possible to make this change optional? There are definitely scenarios where this is super handy, but in other situations it would be just a pain to require extra RAM for an unused feature.

mjjbell:
> UPDATE 2: RAM consumption usage for osrm-routed on europe-210208 with the same setup is 26,2G for master and 29G for this PR, so it's more than a 10% increase in memory requirement. Quite significant.

Given this, I'm inclined to agree that this should be stored as a separate file that can be optionally loaded.

nnseva:
> UPDATE 2: RAM consumption usage for `osrm-routed` on europe-210208 with the same setup is 26,2G for `master` and 29G for this PR, so it's more than a 10% increase in memory requirement. Quite significant.

I should notice that the memory used for the additional segment data (Way IDs vector) is really included in mem-mapped storage (see https://github.com/Project-OSRM/osrm-backend/pull/5968/files#diff-ac92f7a34eb9d5b8b41e6a930db0f2e7356af695724842b390032f1e10842dc2R138). I'm not sure if we can rely on measuring memory consumption done this way.

eveyeti:
Hi @nnseva , 

A try to create a docker image with your PR but when it compile and always have the same error. 


c++: fatal error: Killed signal terminated program cc1plus
compilation terminated.
make[2]: *** [CMakeFiles/EXTRACTOR.dir/build.make:323: CMakeFiles/EXTRACTOR.dir/src/extractor/scripting_environment_lua.cpp.o] Error 1
make[1]: *** [CMakeFiles/Makefile2:717: CMakeFiles/EXTRACTOR.dir/all] Error 2
make: *** [Makefile:152: all] Error 2

how to fix the compilation error?

[CreatingDockerImage.log](https://github.com/Project-OSRM/osrm-backend/files/6135919/CreatingDockerImage.log)

Regards 
Etienne

nnseva:
Hi @eveyeti ,

> c++: fatal error: Killed signal terminated program cc1plus
> compilation terminated.

I don't know exactly, but usually such a problem means not enough memory to compile. Try to increase memory and/or reduce number of processor cores available for the docker in your dockerfile.

Check your dockerfile against an OSRM master branch.

nnseva:
@jcoupey @mjjbell

I've implemented skipping osm way ids in memory.

The extractor now can skip storing OSM Way IDs, the `--skip-way-ids` optional parameter of the extractor turns the feature on.

If the OSM Way IDs section is skipped, the router will skip showing them in annotations.

The correspondent data section is not filled (the 0-size vector is created instead) and doesn't spend memory space.

frodrigo:
Since this feature requires more memory than previous and not required in main use case. It is not preferable to have an option to enable it, rather than one to disable it?

nnseva:
@frodrigo ,

> Since this feature requires more memory than previous and not required in main use case. It is not preferable to have an option to enable it, rather than one to disable it?

1. Notice that the list of OSM Node IDs in the annotation is not required in the main use case also, but even more, is always on.
2. The annotation section list is determined by the enumeration values in the compile time. It would be not a desirable behavior if one of the sections will never appear by default.

mjjbell:
I agree that this should be opt-in rather than opt-out (that was the initial proposal in #5325).
I expect most users will use the default processing setup, so we should optimise for that and give advanced users the option to customise.

> Notice that the list of OSM Node IDs in the annotation is not required in the main use case also, but even more, is always on.

That's a fair point. However, there is interest in an alternative file loading mechanism that reduces memory footprint without a performance penalty (#5974, #5879, #5838). In which case, I can see OSM nodes also being an optionally loaded dataset in the not too distant future.

jcoupey:
> there is interest in an alternative file loading mechanism that reduces memory footprint without a performance penalty

That's a very interesting prospect indeed, which goes beyond the case of this PR. @mjjbell as you pointed out there are already a few tickets about various specific aspects of this question, maybe we should open a meta-ticket on the core principle of a new loading mechanism?

eveyeti:
> Hi @eveyeti ,
> 
> > c++: fatal error: Killed signal terminated program cc1plus
> > compilation terminated.
> 
> I don't know exactly, but usually such a problem means not enough memory to compile. Try to increase memory and/or reduce number of processor cores available for the docker in your dockerfile.
> 
> Check your dockerfile against an OSRM master branch.

Hi @nnseva  thanks a lot, it works your advice!

Regards
Etienne.-

nnseva:
@jcoupey 

> > there is interest in an alternative file loading mechanism that reduces memory footprint without a performance penalty
> 
> That's a very interesting prospect indeed, which goes beyond the case of this PR. @mjjbell as you pointed out there are already a few tickets about various specific aspects of this question, maybe we should open a meta-ticket on the core principle of a new loading mechanism?

I would like to notice that there is some open future in requirements to annotate nodes and segments along the found route. 

Looking for this future it might be appropriate to have a customizable annotations feature that might have an influence on memory consumption. This feature should allow customizing not only values (as it is now available) but keys of annotations.

Such a customizable annotation feature should define a set of available annotation keys, as well as some customizable procedures to get values for such annotations. Such a customizable procedure may be a scripting procedure calling on the extraction phase, as well as just a special open data file format associating source OSM IDs (for nodes) and OSM ID pairs (for segments) with annotated values.

Having such a feature, the deployment staff will have options to control memory consumption allowing only necessary annotation keys for the deployed OSRM instance.

nnseva:
> I agree that this should be opt-in rather than opt-out (that was the initial proposal in #5325).
> I expect most users will use the default processing setup, so we should optimise for that and give advanced users the option to customise.

Please take your attention to that the annotation key set is defined on the compilation time like:
```CPP
    enum class AnnotationsType
    {
        None = 0,
        Duration = 0x01,
        Nodes = 0x02,
        Distance = 0x04,
        Weight = 0x08,
        Datasources = 0x10,
        Speed = 0x20,
        Ways = 0x40,
        All = Duration | Nodes | Distance | Weight | Datasources | Speed | Ways
    };
```

The unsophisticated user using the default options values will probably expect that all available annotation keys including ways will be shown in case of all annotations are requested (`annotations=true` request parameter). Why the `"ways"` section (only!) is not appear? The optimization is not a good explanation for such a missing section.

jcoupey:
> Why the "ways" section (only!) is not appear? 

That would indeed be somehow inconsistent. The remarks about opt-in are rather a stand on a *status quo* position resulting from a concern on RAM requirements for standard usage. Of course you're right the same could (should?) apply to other datasets, but we have to deal with things as they are now and none of the rest is optional.

I think the core question here is whether we have the ability to add a generic mechanism to avoid loading a range of datasets. If this can be done realistically mid-term, we can probably merge this and live with the temporal increase until we provide opt-out mechanisms.

Probably others with more knowledge of the codebase and implications can comment on that last part?

mjjbell:
> The unsophisticated user using the default options values will probably expect that all available annotation keys including ways will be shown in case of all annotations are requested (annotations=true request parameter). Why the "ways" section (only!) is not appear? The optimization is not a good explanation for such a missing section from my point of view.

@nnseva I see your point. I was thinking it would it would have to return an error indicating the annotation was disabled. But if this means we're building a quasi optional data loader, then maybe it's better to design it properly to work for all optional datasets.


> I think the core question here is whether we have the ability to add a generic mechanism to avoid loading a range of datasets. If this can be done realistically mid-term, we can probably merge this and live with the temporal increase until we provide opt-out mechanisms.

> Probably others with more knowledge of the codebase and implications can comment on that last part?

There is an existing ticket (#5218) that outlines the steps required. It was a while ago, but @danpat maybe you have a sense for the effort required?  I don't think the data loading code has changed much since then.

PR #5988, opened 2021-03-19
Complete support for no_entry and no_exit turn restrictions

### LINKED ISSUE #4241: Handle `no_entry` and `no_exit` turn restrictions

mjjbell:
## Before
current

## After
after

([link](https://map.project-osrm.org/?z=18&center=48.806303%2C9.224643&loc=48.806528%2C9.225217&loc=48.806061%2C9.223891&hl=en&alt=0&srv=1))

## Issue

The internal representation of turn restrictions expects only one `from` way and only one `to` way.

According to the [OSM definition](https://wiki.openstreetmap.org/wiki/Relation:restriction#Members), `no_entry` and `no_exit` turn restrictions can have multiple `from` and `to` ways respectively. 
This means they are not fully supported by OSRM's restriction parser.

We complete support for these turn restriction types by parsing all ways and converting a valid restriction with multiple `from`/`to` members into multiple internal restrictions.

# Impact
There are approximately [200 restrictions](http://overpass-turbo.eu/s/15dh) in the OSM dataset which OSRM will now parse correctly.

## Tasklist

 - [ ] review
 - [ ] adjust for comments

## Requirements / Relations
Closes #4241

### DIFF
--- a/src/extractor/restriction_parser.cpp
+++ b/src/extractor/restriction_parser.cpp
@@ -50,13 +50,13 @@ RestrictionParser::RestrictionParser(bool use_turn_restrictions_,
  * in the corresponding profile. We use it for both namespacing restrictions, as in
  * restriction:motorcar as well as whitelisting if its in except:motorcar.
  */
-boost::optional<InputTurnRestriction>
+std::vector<InputTurnRestriction>
 RestrictionParser::TryParse(const osmium::Relation &relation) const
 {
     // return if turn restrictions should be ignored
     if (!use_turn_restrictions)
     {
-        return boost::none;
+        return {};
     }
 
     osmium::tags::KeyFilter filter(false);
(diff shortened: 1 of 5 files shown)


src/extractor/restriction_parser.cpp
```diff
@@ -111,21 +113,26 @@ RestrictionParser::TryParse(const osmium::Relation &relation) const
         else if (value.find("no_") == 0 && !boost::algorithm::ends_with(value, "_on_red"))
         {
             is_only_restriction = false;
+            if (value.find("no_exit") == 0)
```
SiarheiFedartsou, line 116: nit. 
can we use [`boost::algorithm::starts_with`](https://www.boost.org/doc/libs/1_41_0/doc/html/boost/algorithm/starts_with.html) instead? Should be a bit faster.

SiarheiFedartsou (approved):

Issue #5576, opened 2019-10-07
Use profile speeds without overwriting with max_speeds - enhancement


siljuovix:
According to the current version of car.lua,  WayHandlers.maxspeed is called after WayHandlers.speed in:

https://github.com/Project-OSRM/osrm-backend/blob/f520379419aa2d5f11b6c0dfa25fb16967f63e04/profiles/car.lua#L420-L425
which overwrites the speeds with the maximum speed values.
https://github.com/Project-OSRM/osrm-backend/blob/f520379419aa2d5f11b6c0dfa25fb16967f63e04/profiles/lib/way_handlers.lua#L440-L447
It seems more reasonable that, if we specify a default speed per road in:

https://github.com/Project-OSRM/osrm-backend/blob/f520379419aa2d5f11b6c0dfa25fb16967f63e04/profiles/car.lua#L140-L156

Then we only use (max speed *0.8)  if result.forward_speed or result.backward_speed don’t have a value assigned still. 

And finally, only assign the default speed if the parse_maxspeed function returns 0.

This could be done by assigning result.forward_speed = -1 instead of profile.default_speed

https://github.com/Project-OSRM/osrm-backend/blob/f520379419aa2d5f11b6c0dfa25fb16967f63e04/profiles/lib/way_handlers.lua#L293-L299

and then in WayHandlers.maxspeed function we can assign result.forward_speed = forward * profile.speed_reduction only if result.forward_speed == -1.

Looking forward to having your input on this. Thanks in advance.

PR #6002, opened 2021-04-02
Process maxspeed tag before surface, smoothness, and tracktype tags

flother:
Let's say we have a tertiary road with the following tags:

- `maxspeed=60`
- `surface=gravel`
- `smoothness=horrible`

The `maxspeed` tag tells us the legal speed limit, but the `surface` and `smoothness` tags have much more effect on the real-world speed of a car. In the default car profile, the [gravel surface sets a driving speed of 40kph](https://github.com/Project-OSRM/osrm-backend/blob/41dda32546399f1dc12af1de41668993de44c7dc/profiles/car.lua#L223), and the [horrible smoothness sets a driving speed of 10kph](https://github.com/Project-OSRM/osrm-backend/blob/41dda32546399f1dc12af1de41668993de44c7dc/profiles/car.lua#L255). It seems to me that the tags should be processed in this order:

1. `maxspeed`
2. `surface`
3. `smoothness`

That is, `maxspeed` can be overridden by the `surface` tag, which can in turn be overridden by the `smoothness` tag. In the case of our secondary road, it would start with a driving speed of 60kph (from `maxspeed`), which would be reduced to 40kph (from `surface`), and end at 10kph (from `smoothness`).

However, before this change, the `maxspeed` tag is processed *after* `surface` and `smoothness`: our road starts with a driving speed of 40kph (from `surface`), it's reduced to 10kph (from `smoothness`), and then *increased* to 60kph (from `maxspeed`).

This change alters the processing order so that `maxspeed` is processed first, giving it the lowest priority of these three tags.

### DIFF
--- a/profiles/car.lua
+++ b/profiles/car.lua
@@ -431,8 +431,8 @@ function process_way(profile, way, result, relations)
     WayHandlers.hov,
 
     -- compute speed taking into account way type, maxspeed tags, etc.
-    WayHandlers.maxspeed,
     WayHandlers.speed,
+    WayHandlers.maxspeed,
     WayHandlers.surface,
     WayHandlers.penalties,
 
--- a/profiles/car.lua
+++ b/profiles/car.lua
@@ -431,9 +431,9 @@ function process_way(profile, way, result, relations)
     WayHandlers.hov,
 
     -- compute speed taking into account way type, maxspeed tags, etc.
+    WayHandlers.maxspeed,
     WayHandlers.speed,
     WayHandlers.surface,
-    WayHandlers.maxspeed,
     WayHandlers.penalties,
 
     -- compute class labels


mjjbell:
It looks like there is [historical precedent](https://github.com/Project-OSRM/osrm-backend/blob/master/features/car/maxspeed.feature#L33) for `maxspeed` to override the speed assigned by highway type, which is why the tests are failing.

There aren't any tests that reference surface characteristics, so the ordering could be
```
    WayHandlers.speed,
    WayHandlers.maxspeed,
    WayHandlers.surface,
 ```
 
However, in principle I can see why `maxspeed` would always take precedence as its value is specific to the road, whereas surface speeds are estimates for all roads of that type.

Is the example provided you hypothetical, or do you see this issue occurring in OSM?

systemed:
> Is the example you provided hypothetical, or do you see this issue occurring in OSM?

It's moderately common in some US states where bulk edits have been done to add a county-wide speed limit to all roads. First example I stumbled across: https://www.openstreetmap.org/way/14761419

flother:
> Is the example you provided hypothetical, or do you see this issue occurring in OSM?

Similar to what @systemed said, in Iceland (where I use OSRM) the speed limit on rural gravel roads is 80kph (50mph) and many of the roads have the `maxspeed=80` tag set in OSM. But, to take [Hrafnseyrarvegur](https://www.openstreetmap.org/way/681497979) as an example, you wouldn't travel at 80kph if you value your paintwork or your suspension. Here's a [photo from the road](https://ja.is/kort/?x=296832&y=598739&nz=9.46&ja360=1&jh=166.1) as an example of the road conditions there.

I came across this while I was looking into why OSRM was recommending a poorly-maintained gravel road (`maxspeed=80`) over a parallel well-maintained asphalt road (`maxspeed=90`). It turns out that the asphalt road has advisory speed limits of 50kph on a few sharp bends, and they too were tagged on OSM (as `maxspeed=50`, though `maxspeed:advisory=50` would be more accurate). In this particular case taking the surface and smoothness would have been an improvement.

If there's a consensus on what should be done here (if anything) I'm willing to have a go at fixing the tests.

mjjbell:
> It's moderately common in some US states where bulk edits have been done to add a county-wide speed limit to all roads. First example I stumbled across: https://www.openstreetmap.org/way/14761419

>  in Iceland (where I use OSRM) the speed limit on rural gravel roads is 80kph (50mph) and many of the roads have the maxspeed=80 tag set in OSM. But, to take Hrafnseyrarvegur as an example, you wouldn't travel at 80kph if you value your paintwork or your suspension. Here's a photo from the road as an example of the road conditions there.

Thanks, sounds like relying on `maxspeed` is not the best option.

With regards to the tests, the use of `maxspeed` over highway speed seems to go [back a long way](https://github.com/Project-OSRM/osrm-backend/commit/572b17640167b8cdd9a9421e0094954e8e60f4b1) and its use as a heuristic has changed over time ([1](https://github.com/Project-OSRM/osrm-backend/commit/9e64ccdbf279b76030563421c7488d0b8f1f8f46), [2](https://github.com/Project-OSRM/osrm-backend/pull/989), [3](https://github.com/frodrigo/osrm-backend/commit/3c5b2286a3fc16aa69220350c7855d3a5229e4c9)), so it's not clear how much of original reasoning still applies in the face of non-specific bulk edits.  

In which case, the ordering mentioned above seems like a good intermediate step. It will fix your issue without breaking the tests (and anyone who may currently rely on that behaviour).

Related: #5576

flother:
Your suggested reordering makes much more sense: handle default profile speeds first, then maxspeed tags, and then surface/tracktype/smoothness tags. I've just updated the PR to use that order. Now, to use a hypothetical road with these tags:

- `highway=tertiary`
- `maxspeed=60`
- `surface=gravel`
- `smoothness=intermediate`

The speed would be processed like this:

1. Get the default speed from the profile (`tertiary = 40` on [line 150](https://github.com/Project-OSRM/osrm-backend/blob/ec36319232f6b6558080c5586d21fc0bd150de44/profiles/car.lua#L150) of `car.lua`)
2. Change speed to 60 using `maxspeed` tag (`WayHandlers.maxspeed` function in `way_handlers.lua`, [lines 434–447](https://github.com/Project-OSRM/osrm-backend/blob/ec36319232f6b6558080c5586d21fc0bd150de44/profiles/lib/way_handlers.lua#L434-L447))
3. Change speed to 40 using `surface` tag (`WayHandlers.surface` function in `way_handlers.lua`, [lines 360-363](https://github.com/Project-OSRM/osrm-backend/blob/ec36319232f6b6558080c5586d21fc0bd150de44/profiles/lib/way_handlers.lua#L360-L363))
4. Check speed according to `smoothness` tag — but because it's higher than the speed according to the `surface` tag, leave the speed unchanged (`WayHandlers.surface` function again, [lines 368-371](https://github.com/Project-OSRM/osrm-backend/blob/ec36319232f6b6558080c5586d21fc0bd150de44/profiles/lib/way_handlers.lua#L368-L371))

I've left a bit more detail in the message for [commit `ab471c51`](https://github.com/Project-OSRM/osrm-backend/pull/6002/commits/ab471c515d435d1489f627e207fa2026e2162dd9).

mjjbell (approved):
LGTM. Thanks :+1:

PR #6020, opened 2021-04-24
Increase PackedOSMIDs size to 34 bits

### LINKED ISSUE #6016: wrong node number in route
I would like to note you for a small mistake in your api.
When you run your api:
http://router.project-osrm.org/route/v1/driving/14.89857049,50.4026181;14.9366725,50.3229582?alternatives=false&annotations=nodes

All nodes are right from the same area in Czech Republic, but node nr.34819412 is from Norway!!!

I've treated it in my application, but it can be unpleasantly for others...

mjjbell:
# Issue
OSM node [2^33](https://www.openstreetmap.org/node/8589934592) was created in early April 2021. This and all subsequently created IDs will be overflowing OSRM node storage which only supports 33 bit IDs.
Bump the number of bits to 34 to double node ID capacity. This is a breaking change to the data format as it alters the layout of `.osrm.nbg_nodes`.

# Impact
Increases size of file `.osrm.nbg_nodes` and its in-memory representation by ~1%
E.g. `germany-latest.osrm.nbg_nodes` increases from 379,732,480 to 383,647,232 bytes.

## Tasklist

 - [ ] review
 - [ ] adjust for comments

## Requirements / Relations
Fixes #6016

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
       - FIXED: Upgrade to @mapbox/node-pre-gyp fix various bugs with Node 12/14 [#5991](https://github.com/Project-OSRM/osrm-backend/pull/5991)
       - FIXED: `valid` type in documentation examples [#5990](https://github.com/Project-OSRM/osrm-backend/issues/5990)
       - FIXED: Remove redundant loading of .osrm.cell_metrics [#6019](https://github.com/Project-OSRM/osrm-backend/issues/6019)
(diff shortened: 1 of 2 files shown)


danpat (approved):

Lv200:
I also met the problem several days ago, and I changed the bit parameter in my branch.
And I thought 34 is just enough for now, maybe the bit parameter should be set larger for future, 45 can be used for a long time.

PR #6045, opened 2021-06-03
Lazily generate optional route path data

mjjbell:
# Issue

Currently route results are annotated with additional path information, such as geometries, turn-by-turn steps and other metadata.

These annotations are generated even if they are not requested or returned in the response. Datasets needed to generate these annotations are loaded and available to the OSRM process even when unused.

This PR is a first step towards making the loading of these datasets optional. We refactor the code so that route annotations are only generated if explicitly requested and needed in the response

Specifically, we change the following annotations to be lazily generated:
- Turn-by-turn steps
- Route overview geometry
- Route segment metadata 

For example. a `/route/v1` request with
`steps=false&overview=false&annotations=false`
would no longer call the following data facade methods:
- `GetOSMNodeIDOfNode`
- `GetTurnInstructionForEdgeID`
- `GetNameIndex`
- `GetNameForID`
- `GetRefForID`
- `GetTurnInstructionForEdgeID`
- `GetClassData`
- `IsLeftHandDriving`
- `GetTravelMode`
- `IsSegregated`
- `PreTurnBearing`
- `PostTurnBearing`
- `HasLaneData`
- `GetLaneData`
- `GetEntryClass`

Requests that include segment metadata and/or overview geometry but not turn-by-turn instructions will also benefit from this, although there is some interdependency with the step instructions - a call to `GetTurnInstructionForEdgeID` is still required.
Requests for OSM annotations will understandably still need to call `GetOSMNodeIDOfNode`.

Making these changes unlocks the optional loading of data contained in
the following OSRM files:
- `osrm.names`
- `osrm.icd`
- `osrm.nbg_nodes` (partial)
- `osrm.ebg_nodes` (partial)
- `osrm.edges`

## Tasklist

 - [ ] review
 - [ ] adjust for comments

## Requirements / Relations
#5838

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -34,6 +34,8 @@
       - FIXED: Fixed Node docs generation check in CI. [#6058](https://github.com/Project-OSRM/osrm-backend/pull/6058)
       - CHANGED: Docker build, enabled arm64 build layer [#6172](https://github.com/Project-OSRM/osrm-backend/pull/6172)
       - CHANGED: Docker build, enabled apt-get update/install caching in separate layer for build phase [#6175](https://github.com/Project-OSRM/osrm-backend/pull/6175)
+    - Routing:
+      - CHANGED: Lazily generate optional route path data [#6045](https://github.com/Project-OSRM/osrm-backend/pull/6045)
 
 # 5.26.0
   - Changes from 5.25.0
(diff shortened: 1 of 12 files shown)


mjjbell:
After this change, and assuming a mechanism is implemented for optionally loading datasets, I estimate about ~15% memory saving when using the default car profile.
E.g. for `france-latest`, it would save about 700MB.

TheMarex:
This is great!

SiarheiFedartsou (approved):

PR #6050, opened 2021-06-08
Add support for non-round-trips with a single fixed endpoint

### LINKED ISSUE #5647: Issue with Trip service
We have an issue on the OSRM routing system which appears to be a bug. We have a set of points on which we'd like to use the optimize: true option in order to achieve the best path.
The problem is that when we position one of these points at the very center of a roundabout, the point is "ignored" and pushed to last in the line even though it is closest to the origin point
how it is supposed to work:

expected result:

actual result:

mjjbell:
# Issue

Currently `/trip` supports finding round-trip routes where only the start or end location is fixed. 
This PR extends this feature to non-round-trip requests.

We do this by a new table manipulation that simulates non-round-trip fixed endpoint requests as a round-trip request.

This PR also fixes a problem with the `trip` cucumber tests. If the response was not a `200` status code, the tests always passed. This corrects this and fixes the tests that were subsequently failing.

## Tasklist

 - [ ] review
 - [ ] adjust for comments

## Requirements / Relations
Fixes #5647

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -33,6 +33,8 @@
       - FIXED: Fixed Node docs generation check in CI. [#6058](https://github.com/Project-OSRM/osrm-backend/pull/6058)
       - CHANGED: Docker build, enabled arm64 build layer [#6172](https://github.com/Project-OSRM/osrm-backend/pull/6172)
       - CHANGED: Docker build, enabled apt-get update/install caching in separate layer for build phase [#6175](https://github.com/Project-OSRM/osrm-backend/pull/6175)
+    - Routing:
+      - ADDED: Add support for non-round-trips with a single fixed endpoint. [#6050](https://github.com/Project-OSRM/osrm-backend/pull/6050)
 
 # 5.26.0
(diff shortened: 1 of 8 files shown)


mjjbell:
It's ready for review, but the number of people with the time and understanding needed to review this is currently around zero, so if you really need it, I'd suggest trying it in your own fork.

tammarut:
> It's ready for review, but the number of people with the time and understanding needed to review this is currently around zero, so if you really need it, I'd suggest trying it in your own fork.

All right, I've just forked these.
Can I copy your commit? 
Does it work on you test?

tammarut:
> > All right, I've just forked these. Can I copy your commit?
> 
> Yes you can just cherry-pick this commit.
> 
> > Does it work on you test?
> 
> The tests indicate it implements the correct behaviour. But it's still experimental, so I'd be interested to know how it works for you.

I'll try it and tell you the results later.
Thanks 👏🏻

SiarheiFedartsou (approved):

PR #6071, opened 2021-07-04
Replace Travis with Github Actions for CI builds

### LINKED ISSUE #6056: Travis Integration is broken
Travis CI has not picked up the last two PRs (#6050, #6054), and I'm unable to trigger a build on `master`.

It looks like other open source projects are also losing support: https://daniel.haxx.se/blog/2021/06/14/bye-bye-travis-ci/

Github Actions seems like an obvious choice for replacement given we're now using Releases for hosting binaries. Any other suggestions?

mjjbell:
# Issue
This PR replaces Travis for continuous integration with Github Actions.

The Github Actions pipeline is functionally equivalent, with all the same build permutations supported. Whilst the Github Actions offering is broadly equivalent to Travis, a few changes have been made as part of the migration.

The `core` and `optional` Travis stages have been consolidated into one build matrix. This is due to the current inability in
Github Actions to [share build steps](https://github.community/t/reusing-sharing-inheriting-steps-between-jobs-declarations/16851) between jobs, so this avoids having to duplicate the steps.
Optional stage jobs will now run in parallel with core jobs, but they still remain optional in the sense that they don't fail the build.

A number of existing Github Action plugins are used to replace functionality provided by Travis or other tools:
- [Node setup](https://github.com/actions/setup-node)
- [Caching](https://github.com/actions/cache)
- [Code coverage](https://github.com/codecov/codecov-action)
- [Publishing release artifacts](https://github.com/ncipollo/release-action).

Linux builds are updated to build on `Ubuntu 18.04`. MacOS builds are updated to run on `10.15`.
Similar to the Travis Xenial [upgrade attempt](https://github.com/Project-OSRM/osrm-backend/pull/5901), some changes are required due to underlying platform and compiler upgrades.  This means some Node 10 toolchains will no longer be supported.

Whilst there is opportunity to upgrade build dependencies and make the CI steps more idiomatic, I've left this for future changes and just focussed on functional replication.

## Tasklist

 - [ ] review
 - [ ] adjust for comments

## Requirements / Relations
Closes #6056


CMakeLists.txt
```diff
@@ -514,7 +514,7 @@ if(ENABLE_MASON)
   set(CMAKE_MODULE_LINKER_FLAGS "${CMAKE_MODULE_LINKER_FLAGS} ${LINKER_FLAGS}")
 
   # current mason packages target -D_GLIBCXX_USE_CXX11_ABI=0
-  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -D_GLIBCXX_USE_CXX11_ABI=0")
+  add_dependency_defines(-D_GLIBCXX_USE_CXX11_ABI=0)
```
mjjbell, line 517: On Travis Trusty builds, the compilers all had `GLIBCXX_USE_CXX11_ABI=0` by default. This setting has been flipped in Bionic, so we need to explicitly pass this to pkg-config so that the example builds.

src/extractor/restriction_parser.cpp
```diff
@@ -244,14 +242,15 @@ bool RestrictionParser::ShouldIgnoreRestriction(const std::string &except_tag_st
 
     // Be warned, this is quadratic work here, but we assume that
     // only a few exceptions are actually defined.
-    std::vector<std::string> exceptions;
-    boost::algorithm::split_regex(exceptions, except_tag_string, boost::regex("[;][ ]*"));
+    const std::regex delimiter_re("[;][ ]*");
```
mjjbell, line 245: Similar problem to the Travis Xenial [upgrade attempt](https://github.com/Project-OSRM/osrm-backend/pull/5901#discussion_r640162103).
Boost.Regex and GCC5 appears to be buggy with respect to `[abi:cxx11]` tags, so using this as an opportunity to replace it with `std::regex`.

unit_tests/library/json.cpp
```diff
@@ -27,7 +26,7 @@ BOOST_AUTO_TEST_CASE(test_json_linestring)
     const auto coords = geom.values["coordinates"].get<util::json::Array>().values;
     BOOST_CHECK_EQUAL(coords.size(), 3); // array of three location arrays
 
-    for (const auto each : coords)
+    for (const auto &each : coords)
```
mjjbell, line 29: Unit test changes are due to compiler errors from updated macOS clang.

mjjbell:
It looks like Github Actions makes each job have a status check, rather than the action as a whole. 

This would be a problem if we want checks to be required because any change to the job matrix would break the required list. 

I think a solution is to have a no-op job that depends on all other jobs succeeding, and make this the only required check. That way we get required stability as the jobs evolve.

mjjbell:
> I think a solution is to have a no-op job that depends on all other jobs succeeding, and make this the only required check. That way we get required stability as the jobs evolve.

Added a `ci-complete` step that can be used as a required check.

mjjbell:
@danpat any thoughts on this?
Given the Travis integration is dead but required, we can't merge anything until it is disabled/replaced.

TheMarex:
Added myself as reviewer, I'll take a look Friday.

CMakeLists.txt
TheMarex: I forgot why we had this in the first place. I think this was mason related given that is afaik the only place where we care about ABI of our binaries.

src/extractor/restriction_parser.cpp
TheMarex: Nice!

unit_tests/library/json.cpp
TheMarex: This crashes clang?

unit_tests/library/match.cpp
```diff
@@ -184,7 +184,7 @@ BOOST_AUTO_TEST_CASE(test_match_fb_serialization)
     const auto matchings = fb->routes();
     const auto &number_of_matchings = matchings->size();
 
-    for (const auto &waypoint : *waypoints)
+    for (const auto waypoint : *waypoints)
```
TheMarex, line 187: Also clang-related?

.github/workflows/osrm-backend.yml
```diff
@@ -0,0 +1,588 @@
...
+          - name: mason-linux-release
+            continue-on-error: false
+            node: 10
+            runs-on: ubuntu-18.04
+            BUILD_TOOLS: ON
+            BUILD_TYPE: Release
+            CLANG_VERSION: 5.0.0
+            ENABLE_MASON: ON
+
+          - name: gcc-9-release
```
TheMarex, line 108: I think we can simplify the build matrix a bit and deprecate the older GCCs.

TheMarex (approved):
Looks great, thanks for doing this. I think we can also trim the matrix a bit, there is little sense in supporting the older GCC from 4-5 years ago.

unit_tests/library/json.cpp
mjjbell: Running `make tests` with  clang 12

> error: loop variable 'each' creates a copy from type 'const mapbox::util::variant<osrm::util::json::String, osrm::util::json::Number, mapbox::util::recursive_wrapper<osrm::util::json::Object>, mapbox::util::recursive_wrapper<osrm::util::json::Array>, osrm::util::json::True, osrm::util::json::False, osrm::util::json::Null>' [-Werror,-Wrange-loop-construct]
    for (const auto each : coords)

unit_tests/library/match.cpp
mjjbell: Yes, similar errors to above: https://github.com/mjjbell/osrm-backend/runs/2957246502?check_suite_focus=true#step:15:2784

> error: loop variable 'destination' is always a copy because the range of type 'const flatbuffers::Vector<flatbuffers::Offset<osrm::engine::api::fbresult::Waypoint> >' does not return a reference [-Werror,-Wrange-loop-analysis]
    for (const auto &destination : *destinations_array)
                     ^
/Users/runner/work/osrm-backend/osrm-backend/unit_tests/library/table.cpp:335:10: note: use non-reference type 'const osrm::engine::api::fbresult::Waypoint *'
    for (const auto &destination : *destinations_array)

CMakeLists.txt
mjjbell: Yep, the Mason Boosts are built against the old ABI .

mjjbell:
Thanks for the review @TheMarex!

unit_tests/library/json.cpp
TheMarex: Ah I see `-Werror` strikes again.

mjjbell:
@TheMarex would you be able to disable Travis CI as a required check? I believe only repo owners have the privileges needed.

It's configured in the Settings page: https://docs.github.com/en/github/administering-a-repository/defining-the-mergeability-of-pull-requests/managing-a-branch-protection-rule

TheMarex:
@mjjbell Done. I'll merge this now to enable the required `ci-complete` step.

PR #6083, opened 2021-07-17
Fix Boost link flags in pkg-config file.

### LINKED ISSUE #5687: find_package cannot find Boost
I have built osrm-backend successfully on my mac machine, but when CMake example I get multiple warning related to boost such as 
```
Target "osrm-example" links to target "Boost::regex" but the target was not found.  Perhaps a find_package() call is missing for an IMPORTED target, or an ALIAS target is missing?
```
and then when running build it throws error and fails

```
ld: library not found for -lBoost::regex
```
How to get around this issue? I have boost already installed. Thanks
```
ls /usr/local/Cellar/boost/
1.72.0
```

mjjbell:
# Issue

In newer versions of cmake, [FindBoost](https://cmake.org/cmake/help/latest/module/FindBoost.html) uses [Imported Targets ](https://cmake.org/cmake/help/latest/guide/importing-exporting/index.html)for library component variables, rather than file paths to the Boost libraries.

cmake uses these targets when linking (e.g. `target_link_library`) and knows how to correctly substitute the values. However, the OSRM pkg-config file that we generate doesn't do this, and ends up writing the actual target symbols, hence the errors trying to link `Boost::<component>`.

To fix this for newer cmake versions, we create an intermediate configure step that references [TARGET_LINKER_FILE](https://cmake.org/cmake/help/latest/manual/cmake-generator-expressions.7.html#genex:TARGET_LINKER_FILE) for each imported target. This is followed by a `file(GENERATE` step that performs the correct substitution.
See this thread for more details: https://cmake.org/pipermail/cmake/2018-December/068812.html

This is backwards compatible to the existing min cmake version (3.1). However, building using cmake 3.1 fails with an unrelated package.json parsing error:
```bash
CMake Error at cmake/JSONParser.cmake:287 (string):
  string end index: 1 is out of range -1 - 0
  cmake/JSONParser.cmake:232 (_sbeMoveToNextNonEmptyCharacter)
  cmake/JSONParser.cmake:71 (_sbeParseObject)
  cmake/JSONParser.cmake:22 (_sbeParse)
  CMakeLists.txt:69 (sbeParseJson)
```
Given 3.1 is nearly 7 years old, I don't think it's worth investigating. Instead, this PR also bumps the min version to 3.2.

## Tasklist

 - [ ] review
 - [ ] adjust for comments

## Requirements / Relations

Fixes #5687

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 3.1)
+cmake_minimum_required(VERSION 3.2)
 
 if(CMAKE_CURRENT_SOURCE_DIR STREQUAL CMAKE_CURRENT_BINARY_DIR AND NOT MSVC_IDE)
   message(FATAL_ERROR "In-source builds are not allowed.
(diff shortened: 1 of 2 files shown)


CMakeLists.txt
```diff
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 3.1)
+cmake_minimum_required(VERSION 3.2)
```
TheMarex, line 1: Seems fine, we can probably even go higher if that makes things easier, given most people build this in docker images anyway.

TheMarex (approved):
Looks good to me. :+1:

PR #6090, opened 2021-07-27
Fix the URL parser for special chars in profile

TheMarex:
# Issue

The URLs used for `osrm-routed` have the form `{service}/{version}/{profile}/{params}?{options...}`. The boost spirit grammar for parsing the `profile` and `service` identifiers was artificially limited to only allow alpha-numerical characters, but any URL-safe character would be fine.

This PR additionally allows for `._-~`.

## Tasklist

 - [x] add tests
 - [ ] review
 - [ ] adjust for comments

### DIFF
--- a/src/server/api/url_parser.cpp
+++ b/src/server/api/url_parser.cpp
@@ -26,10 +26,10 @@ struct URLParser final : qi::grammar<Iterator, Into>
     {
         using boost::spirit::repository::qi::iter_pos;
 
-        identifier = qi::char_("a-zA-Z0-9_..--~:");
+        identifier = qi::char_("a-zA-Z0-9_.~:-");
         percent_encoding =
             qi::char_('%') > qi::uint_parser<unsigned char, 16, 2, 2>()[qi::_val = qi::_1];
-        polyline_chars = qi::char_("a-zA-Z0-9_..--[]{}@?|\\~`^") | percent_encoding;
+        polyline_chars = qi::char_("a-zA-Z0-9_[]{}@?|\\~`^") | percent_encoding;
         all_chars = polyline_chars | qi::char_("=,;:&()..");
 
         service = +identifier;
--- a/src/server/api/url_parser.cpp
+++ b/src/server/api/url_parser.cpp
@@ -26,15 +26,15 @@ struct URLParser final : qi::grammar<Iterator, Into>
     {
         using boost::spirit::repository::qi::iter_pos;
 
-        alpha_numeral = qi::char_("a-zA-Z0-9");
+        identifier = qi::char_("a-zA-Z0-9_..--~:");
         percent_encoding =
             qi::char_('%') > qi::uint_parser<unsigned char, 16, 2, 2>()[qi::_val = qi::_1];
-        polyline_chars = qi::char_("a-zA-Z0-9_.--[]{}@?|\\~`^") | percent_encoding;
-        all_chars = polyline_chars | qi::char_("=,;:&().");
+        polyline_chars = qi::char_("a-zA-Z0-9_..--[]{}@?|\\~`^") | percent_encoding;
+        all_chars = polyline_chars | qi::char_("=,;:&()..");
 
-        service = +alpha_numeral;
+        service = +identifier;
         version = qi::uint_;
-        profile = +alpha_numeral;
+        profile = +identifier;
         query = +all_chars;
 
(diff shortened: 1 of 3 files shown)


src/server/api/url_parser.cpp
```diff
@@ -26,15 +26,15 @@ struct URLParser final : qi::grammar<Iterator, Into>
     {
         using boost::spirit::repository::qi::iter_pos;
 
-        alpha_numeral = qi::char_("a-zA-Z0-9");
+        identifier = qi::char_("a-zA-Z0-9_..--~:");
         percent_encoding =
             qi::char_('%') > qi::uint_parser<unsigned char, 16, 2, 2>()[qi::_val = qi::_1];
-        polyline_chars = qi::char_("a-zA-Z0-9_.--[]{}@?|\\~`^") | percent_encoding;
-        all_chars = polyline_chars | qi::char_("=,;:&().");
+        polyline_chars = qi::char_("a-zA-Z0-9_..--[]{}@?|\\~`^") | percent_encoding;
```
TheMarex, line 32: Turns out we need to specify `..` for a single dot to work.

src/server/api/url_parser.cpp
```diff
@@ -26,15 +26,15 @@ struct URLParser final : qi::grammar<Iterator, Into>
     {
         using boost::spirit::repository::qi::iter_pos;
 
-        alpha_numeral = qi::char_("a-zA-Z0-9");
+        identifier = qi::char_("a-zA-Z0-9_..--~:");
```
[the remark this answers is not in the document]
mjjbell: Yes, moving the `-` to the end seems to prevent it from forming a range, and removes the need for `..`
```cpp
identifier = qi::char_("a-zA-Z0-9_.~:-");
```

src/server/api/url_parser.cpp
mjjbell: I think a similar approach can be applied as above.

Perhaps out of scope, but I don't think `.` or `-` can appear in polylines, as their ASCII code is less than 63? Maybe that's why we've never noticed that the `.--` range doesn't work.

src/server/api/url_parser.cpp
```diff
@@ -26,15 +26,15 @@ struct URLParser final : qi::grammar<Iterator, Into>
...
         percent_encoding =
             qi::char_('%') > qi::uint_parser<unsigned char, 16, 2, 2>()[qi::_val = qi::_1];
-        polyline_chars = qi::char_("a-zA-Z0-9_.--[]{}@?|\\~`^") | percent_encoding;
-        all_chars = polyline_chars | qi::char_("=,;:&().");
+        polyline_chars = qi::char_("a-zA-Z0-9_..--[]{}@?|\\~`^") | percent_encoding;
+        all_chars = polyline_chars | qi::char_("=,;:&()..");
```
mjjbell, line 33: Shouldn't be needed as above.

unit_tests/server/url_parser.cpp
```diff
@@ -124,6 +124,16 @@ BOOST_AUTO_TEST_CASE(valid_urls)
     BOOST_CHECK_EQUAL(reference_8.profile, result_8->profile);
     CHECK_EQUAL_RANGE(reference_8.query, result_8->query);
     BOOST_CHECK_EQUAL(reference_8.prefix_length, result_8->prefix_length);
+
+    // profile with special characters
```
mjjbell, line 128: :+1:

mjjbell (changes requested):

src/server/api/url_parser.cpp
mjjbell:
```suggestion
        all_chars = polyline_chars | qi::char_("=,;:&().-");
```
With `-` removed from `polyline_chars`, it needs to be added here.
For example, the following test breaks otherwise:

https://github.com/Project-OSRM/osrm-backend/blob/dca35dcc8676df6fa93f491a7b157136918173d5/features/testbot/status.feature#L55

mjjbell:
@TheMarex one minor change left to prevent a regression.

(Ideally it would be picked up by CI but the Travis integration is broken - if you have some time, take a look at #6071 which uses Github Actions as a drop-in replacement)

mjjbell (changes requested):
One minor change to prevent regression.

mjjbell (approved):
:+1:

PR #6106, opened 2021-08-24
Fixed edge_speed_in_km_h as Double in Traffic Updates

juchengquan:
# Issue
Fixed an issue that /src/updater/csv_source.cpp only took edge_speed_in_km_h as unit when processing traffic updates. 
Now edge_speed_in_km_h can be set as double type. 

## Tasklist

 - [ ] review
 - [ ] adjust for comments

## Requirements / Relations

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,7 @@
 # Unreleased
+- Changes from 5.25.0
+    - API:
+      - FIXED: Allow for traffic updates to use edge_speed_in_km_h in decimals.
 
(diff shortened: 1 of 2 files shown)


danpat:
https://github.com/Project-OSRM/osrm-backend/blob/master/include/updater/source.hpp#L53

@juchengquan Any decimal places you include in the CSV will be silently truncated - OSRM's internal data structures only store speeds as integers.  We kept parsing as integers so that data wasn't silently discarded without warning the user.

Issue #6121, opened 2021-09-08
Division by zero in printCellStatistics()


pservit:
osrm-partition always fails processing planet after osrm-extract with foot profile.

After some debugging and adding logs I found that `partition.GetNumberOfCells` returns 0 for level=4 so division by zero occurs in [printCellStatistics() line 42](https://github.com/Project-OSRM/osrm-backend/blob/8e100ea1a019bff099cc9b8b15c5b4c0da0d2f35/include/partitioner/cell_statistics.hpp#L42)

```
  source /= num_cells;
  destination /= num_cells;
```

Here is full log for osrm-partition:

```
[info] Computing recursive bisection
[info] Loaded compressed node based graph: 736367212 edges, 1811657456 nodes
[info]  running partition: 128 1.2 0.25 10 1000 # max_cell_size balance boundary cuts small_component_size
[info] Found 1532677258 SCC (857 large, 1532676401 small)
[info] SCC run took: 93.7171s
[info] Full bisection done in 7846.82s
[info] Loaded node based graph to edge based graph mapping
[info] Loaded edge based graph for mapping partition ids: 2996357810 edges, 736366438 nodes
[info] Fixed 38099 unconnected nodes
[info] Edge-based-graph annotation:
[info]   level 1 #cells 2458529 bit size 22
[info]   level 2 #cells 258451 bit size 18
[info]   level 3 #cells 16310 bit size 14
[info]   level 4 #cells 534 bit size 10
[info] Renumbered data in 2297.4 seconds
[info] MultiLevelPartition constructed in 204.034 seconds
[info] CellStorage constructed in 48.4773 seconds
[info] MLD data writing took 64.3765 seconds
[info] Cells statistics per level
[info] Level 1
[info] num_cells 2458529
[info] source 42477944  num_cells 2458529
[info] destination 62028086  num_cells 2458529
[info] Level 1 #cells 2458529 #boundary nodes 62842547, sources: avg. 17, destinations: avg. 25, entries: 1581977611 (12655820888 bytes)
[info] Level 2
[info] num_cells 258451
[info] source 9323222  num_cells 258451
[info] destination 13481432  num_cells 258451
[info] Level 2 #cells 258451 #boundary nodes 13649358, sources: avg. 36, destinations: avg. 52, entries: 685047779 (5480382232 bytes)
[info] Level 3
[info] num_cells 16310
[info] source 1191758  num_cells 16310
[info] destination 1719894  num_cells 16310
[...]

PR #6124, opened 2021-09-13
Fix metric offset overflow for large MLD partitions

mjjbell:
# Issue
As discovered in #6121

Each MLD cell has source and destination nodes. MLD is keeping a |source| x |destination| sized table
for various metrics (distances, durations, etc) from each source to all destinations in a cell.

It stores all of the values for a metric in one large array, with an offset for each cell to find its values. The offset is currently
limited to 32 bit values, which overflows on very large graphs (e.g. Planet OSM).

We fix this by changing the offsets to be uint64_t types.

## Tasklist

 - [ ] review
 - [ ] adjust for comments

### DIFF
--- a/include/partitioner/cell_storage.hpp
+++ b/include/partitioner/cell_storage.hpp
@@ -52,11 +52,9 @@ namespace detail
 template <storage::Ownership Ownership> class CellStorageImpl
 {
   public:
-    using ValueOffset = std::uint32_t;
-    using BoundaryOffset = std::uint32_t;
+    using ValueOffset = std::uint64_t;
+    using BoundaryOffset = std::uint64_t;
     using BoundarySize = std::uint32_t;
-    using SourceIndex = std::uint32_t;
-    using DestinationIndex = std::uint32_t;
 
     static constexpr auto INVALID_VALUE_OFFSET = std::numeric_limits<ValueOffset>::max();
     static constexpr auto INVALID_BOUNDARY_OFFSET = std::numeric_limits<BoundaryOffset>::max();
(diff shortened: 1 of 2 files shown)


include/partitioner/cell_storage.hpp
```diff
@@ -52,11 +52,9 @@ namespace detail
 template <storage::Ownership Ownership> class CellStorageImpl
 {
   public:
-    using ValueOffset = std::uint32_t;
-    using BoundaryOffset = std::uint32_t;
+    using ValueOffset = std::size_t;
+    using BoundaryOffset = std::size_t;
     using BoundarySize = std::uint32_t;
```
mjjbell, line 57: OSM contains ~800 million ways, so 32 bit should be safe to use for boundary sizes..for now.

include/partitioner/cell_storage.hpp
```diff
@@ -52,11 +52,9 @@ namespace detail
 template <storage::Ownership Ownership> class CellStorageImpl
 {
   public:
-    using ValueOffset = std::uint32_t;
-    using BoundaryOffset = std::uint32_t;
+    using ValueOffset = std::size_t;
```
mjjbell, line 55: Still an overflow risk if building on 32 bit machine, but I expect it would fail earlier on for planet sized data.

danpat: Hmm, why not `std::uint64_t`?  I think, in general, when we're dealing with values that end up serialized, we should be explicit about sizes.

mjjbell: The intention was to make it clear that this type represents an index into the other cell arrays.
I had not considered the serialization aspect though, so I'll change to `std::uint64_t` - it will also prevent any possible issue on 32 bit builds.

danpat (approved):

danpat:
@mjjbell I think I'm going to make a new release once this is merged - any objections?

mjjbell:
@danpat sounds good :+1: 

I can see I tested the package publishing from Github Actions in https://github.com/mjjbell/osrm-backend/actions/runs/998677972, but I think I must have removed the generated artifacts to not confuse myself in the future.

mjjbell:
Proof that it is indeed a breaking change to the data format: #6128

PR #6128, opened 2021-09-21
Restrict test cache restoration

mjjbell:
# Issue
There is currently no way to invalidate the cache (https://github.com/actions/cache/issues/2), so master won't pass CI if a breaking change has been made to the binary data formats.

Fix this by only restoring test cache from the same commit SHA.

## Tasklist

 - [ ] review
 - [ ] adjust for comments


mjjbell:
Actually, I think this is because `osrm-contract` and `osrm-customize` are not being considered when creating the test directories.

mjjbell:
Closed in favour of #6129

Issue #6134, opened 2021-09-25
docker release for 5.26.0


cyril23:
Hey guys,

thanks for the release of 5.26.0 here. Great to see the OSRM project is still going strong! 

Would you be so kind and make a docker release version of 5.26.0, too? Currently it's stuck at 5.25.0

(If anyone shows me how to do it, I could do the job next time, too.)

If I am reading https://github.com/Project-OSRM/osrm-backend-docker correctly, then the whole process is automated and I just need to be a little bit more patient:

> The OSRM demo server periodically uses build-master.sh to grab the latest master code and publish a new image to DockerHub.

But does this work for the releases (in docker: tags), too? https://hub.docker.com/r/osrm/osrm-backend/tags

danpat:
@cyril23 Docker has changed their terms of service - honestly, I haven't read the details, but it seems that our automated builds aren't working.

Building docker images locally should be trivial though - just `docker build .` in the root of the code repo and you'll shortly have a local image to use.

hoerup:
Im thinking that somethings missing in the migration from travis to github actions ??

mjjbell:
IIUC, there's a Docker Hub [build hook](https://github.com/Project-OSRM/osrm-backend/blob/master/docker/hooks/build) that's no longer triggering. Docker Hub [documentation](https://docs.docker.com/docker-hub/builds/) says this now requires a subscription.

We can apply to re-enable the build pipeline by joining the [open source program](https://www.docker.com/community/open-source/application).

hoerup:
I assume that the container image build could be done as a GH action then pushed to either docker hub or ghcr.io

nilsnolde:
probably worth migrating to github packages instead of docker (until gh packages pulls the same move;)). we did the same for our images, from what I heard the "open source program" is pretty useless, @jcoupey tried it once IIRC

nilsnolde:
if you don't have a crack at it I might @hoerup . for reference, I use this to build and push valhalla: https://github.com/gis-ops/docker-valhalla/blob/master/.github/workflows/publish.yml. I omitted dockerhub for now, but could/should be done as well, if only for convenience reason (docker pull ghcr.io/... is somewhat more work on the command line;))

jcoupey:
> @jcoupey tried it once IIRC

[...]

PR #6137, opened 2021-09-27
Add GH actions script for docker image

hoerup:
# Issue
Related to https://github.com/Project-OSRM/osrm-backend/issues/6134

I'm new to github actions but I think this should do the trick ?

### DIFF
--- a/.github/workflows/osrm-backend-docker.yml
+++ b/.github/workflows/osrm-backend-docker.yml
@@ -3,6 +3,7 @@ on:
   push:
     tags:
       - 'v*'
+  workflow_dispatch:
 
 jobs:
   publish:
--- a/.github/workflows/osrm-backend-docker.yml
+++ b/.github/workflows/osrm-backend-docker.yml
@@ -1,0 +1,30 @@
+name: build and publish container image
+on:
+  create:
+    tags:
+      - v*
+
+jobs:
+  publish:
+    runs-on: ubuntu-latest
+    steps:       
+      - name: Check out the repo
+        uses: actions/checkout@v2
+
+
+      - name: Log in to GitHub Docker Registry
+        uses: docker/login-action@v1
+        with:
+          registry: ghcr.io
+          username: ${{ github.actor }}
+          password: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Build container image
+        uses: docker/build-push-action@v2
+        with:
+          push: true
+          file: ./docker/Dockerfile
+          tags: ghcr.io/${{ github.repository }}/osrm-backend:${{ env.GITHUB_REF }} 
+
+
+            
--- a/.github/workflows/osrm-backend-docker.yml
+++ b/.github/workflows/osrm-backend-docker.yml
@@ -55,9 +55,9 @@ jobs:
         with:
           push: true
           file: ./docker/Dockerfile
-          tags:  ${{ steps.meta.outputs.labels }}
+          tags:  ${{ steps.meta.outputs.tags  }}
           build-args: |
-            DOCKER_TAG=${{ join(steps.metaassertions.outputs.labels) }}
+            DOCKER_TAG=${{ join(steps.metaassertions.outputs.tags ) }}
 
 
       - name: Build container image - debug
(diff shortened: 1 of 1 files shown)


.github/workflows/osrm-backend-docker.yml
```diff
@@ -0,0 +1,30 @@
+name: build and publish container image
+on:
+  create:
+    tags:
+      - v*
```
mjjbell, line 5: I think tag filters are only supported on the `push` or `pull_request` events
https://docs.github.com/en/actions/learn-github-actions/workflow-syntax-for-github-actions#onpushpull_requestbranchestags

For example, it looks like this triggered when you created a branch with no tag.
https://github.com/hoerup/osrm-backend/actions/runs/1279643123

.github/workflows/osrm-backend-docker.yml
```diff
@@ -0,0 +1,30 @@
...
+        with:
+          registry: ghcr.io
+          username: ${{ github.actor }}
+          password: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Build container image
+        uses: docker/build-push-action@v2
+        with:
+          push: true
+          file: ./docker/Dockerfile
```
mjjbell, line 26: To support debug/assertion builds like the old hook, you'll need to pass build-args here
https://github.com/Project-OSRM/osrm-backend/blob/34877482e678e44d995454586191ee39d5c31c4b/docker/hooks/build#L9

.github/workflows/osrm-backend-docker.yml
```diff
@@ -0,0 +1,30 @@
...
+          tags: ghcr.io/${{ github.repository }}/osrm-backend:${{ GITHUB_REF#refs/*/ }}
```
mjjbell, line 27: Have you considered using [docker/metadata-action](https://github.com/docker/metadata-action)?
If has some nice Docker tag extraction functionality (e.g. we can push `v5.26.0`, and `latest` on the same container).

mjjbell (changes requested):

.github/workflows/osrm-backend-docker.yml
hoerup: docker/metadata-action seems nice - but I can't figure out how to combine that with the normal/debug/assertions variations

.github/workflows/osrm-backend-docker.yml
```diff
@@ -0,0 +1,84 @@
...
+    steps:       
+      - name: Check out the repo
+        uses: actions/checkout@v2
+
+
+      - name: Docker meta
+        id: meta
+        uses: docker/metadata-action@v3
+        with:
+          images: ghcr.io/${{ github.repository }}/osrm-backend
```
mjjbell, line 21: [github.repository](https://docs.github.com/en/actions/learn-github-actions/contexts#github-context) includes the repository name.

.github/workflows/osrm-backend-docker.yml
```diff
@@ -0,0 +1,84 @@
...
+
+
+      - name: Build container image - normal
+        uses: docker/build-push-action@v2
+        with:
+          push: true
+          file: ./docker/Dockerfile
+          tags:  ${{ steps.meta.outputs.labels }}
+          build-args: 
+          - DOCKER_TAG:  ${{ steps.meta.outputs.labels }}
```
mjjbell, line 61: Looks like `build-args` expects a list, not a dictionary.
>  build-args | List | List of build-time variables

With the `github.repository` change, this file should be fork agnostic, you can try building the containers on your own fork to confirm it works.

.github/workflows/osrm-backend-docker.yml
```diff
@@ -0,0 +1,84 @@
+name: build and publish container image
+on:
+  push:
+    tags:
+      - 'v*'
+    branches:
+      - master
```
mjjbell, line 7: Hmm, so the `latest` tag will would flip between stable release versions and master builds. I think this is confusing.

mjjbell (changes requested):

hoerup:
I think I got it to where we want:

https://github.com/hoerup/osrm-backend/runs/3743503746?check_suite_focus=true

https://github.com/hoerup/osrm-backend/pkgs/container/osrm-backend

Let me know if you want a clean PR with out the mess

PR #5509, opened 2019-08-07
osrm-routed monitoring in the prometheus format

gardster:
# Issue

[#5180](https://github.com/Project-OSRM/osrm-backend/issues/5180)
Endpoint to check OSRM status with monitoring solutions based on [prometheus](https://prometheus.io/docs/concepts/data_model/) format. 
This is a proof of concept implementation that has only several simple options. Additional metrics can be added with ongoing PRs.
Metrics endpoint is a separate thread that listens its own port and doesn't bother worker threads.

Issue #5180, opened 2018-08-24
Monitor OSRM with Prometheus


lasanthaa:
Is there any way to monitor osrm with Prometheus?

TheMarex:
There is probably nothing pre-build you would need to figure that out yourself (it is kind of out of the scope of this project).

PR #6113, opened 2021-08-29
Upgrade Boost to 1.70, fix inefficient connection handling

mjjbell:
# Issue

A request to `osrm-routed` can be assigned to a thread which is currently busy processing another request, even when there
are other threads/cores available. This unnecessarily delays the response, and can make requests appear to hang when awaiting CPU intensive requests to finish.

See #6039 for details of how to reproduce the issue.

`osrm-routed` server implementation is heavily influenced by the [HTTP server 3 example](https://www.boost.org/doc/libs/1_65_0/doc/html/boost_asio/examples/cpp03_examples.html) in the
Boost.Asio docs. By upgrading to Boost 1.70 and updating the server connections to match the [example](https://www.boost.org/doc/libs/1_70_0/doc/html/boost_asio/example/cpp03/http/server3/connection.cpp) provided in the 1.70
release, the problem is resolved.

The diff of the changes to the Boost.Asio stack are [vast](https://github.com/boostorg/asio/commit/59066d80b26e1d5b83b60d127ee17948d9ae9702), so it's difficult to identify the exact cause. However
the implementation change is to push the `strand` of execution into the socket (and timer) objects, which suggests it's
fixing the type of threading issue we are observing.

This PR therefore also bumps the Boost dependency to >= 1.70. Given this was released over two years ago and many people will have already been using this version as a dependency when building OSRM, I see this as low risk.

## Tasklist

 - [x] CHANGELOG.md entry ([How to write a changelog entry](http://keepachangelog.com/en/1.0.0/#how))
 - [ ] update relevant [Wiki pages](https://github.com/Project-OSRM/osrm-backend/wiki)
 - [ ] add tests (see [testing documentation](https://github.com/Project-OSRM/osrm-backend/blob/master/docs/testing.md)
 - [ ] review
 - [ ] adjust for comments
 - [ ] cherry pick to release branch

## Requirements / Relations

Fixes #6039

PR #6147, opened 2021-10-15
osrm-routed monitoring in the prometheus format

juliorenner:
#5180
Endpoint to check OSRM status with monitoring solutions based on prometheus format.
This is a proof of concept implementation that has only several simple options. Additional metrics can be added with ongoing PRs.
Metrics endpoint is a separate thread that listens its own port and doesn't bother worker threads.

#5509 Fixing conflicts


mjjbell:
> @mjbell could you take a look ?

I've looked over this briefly.
- I'm not a big fan of the scattering of counters around the codebase, it doesn't feel very maintainable.
- There are good C++ libraries like[ prometheus-cpp](https://github.com/jupp0r/prometheus-cpp) or maybe even [opentelemetry-cpp](https://github.com/open-telemetry/opentelemetry-cpp) which are worth evaluating as more maintainable alternatives.

If we want to go down the route of having built-in observability, it's probably also worth discussing the options for a more robust server alternative for `osrm-routed`. The current implementation as [pointed out recently](https://github.com/Project-OSRM/osrm-backend/pull/6113#pullrequestreview-745904617) is just a copy-paste job from the Boost docs, and not something that was intended to be a highly-available standalone server.

mjjbell:
I don't think anyone is planning anything for future releases at this point :slightly_smiling_face: , it's up to the community.

I agree that native observability would be beneficial, but I have reservations specifically about this PR's structure and about investing a lot of time in `osrm-routed`as mentioned above. 

Depending on your exact needs, you might get some immediate benefit from running a proxy like Envoy or Nginx in front of `osrm-routed` and exporting HTTP request metrics from the proxy instead.

DennisOSRM:

PR #6151, opened 2021-10-22
Fix for_each_indexed util function

mjjbell:
# Issue
- Fix typo in util function name `for_each_indexed`.
- Use the overloaded functions `for_each_indexed` and `for_each_pair`
  with a container argument where possible to improve readability.


TheMarex (approved):
Nice!

PR #6153, opened 2021-11-03
Support OSM traffic signal directions

### LINKED ISSUE #1317: traffic_signals:direction for traffic signal penalties
Would it be possible to use traffic_signals:direction informations when possible to avoid double count of traffic signals for complex intersections? see http://wiki.openstreetmap.org/wiki/Tag:highway=traffic_signals#Complex_intersections

mjjbell:
# Issue
Currently OSRM parses traffic signal nodes without consideration for the direction in which the signal applies. 
This can lead to duplicated routing penalties, especially when a forward and backward signal are in close proximity on a way.
    
 This commit adds support for directed signals to the extraction and graph creation. Signal penalties are only applied in the direction specified by the OSM tag.
    
We add the assignment of traffic directions to the lua scripts, maintaining backwards compatibility with the existing boolean traffic states.
    
As part of the changes to the internal structures used for tracking traffic signals during extraction, we stop serialising/deserialising signals to the `.osrm` file. 

The traffic signals are only used by `osrm-extract` so whilst this is a data format change, it will not break any existing user processes.

## Tasklist

 - [ ] review
 - [ ] adjust for comments

## Requirements / Relations

Fixes #1317

### DIFF
--- a/include/extractor/traffic_lights.hpp
+++ b/include/extractor/traffic_lights.hpp
@@ -1,0 +1,25 @@
+#ifndef OSRM_EXTRACTOR_TRAFFIC_LIGHTS_DATA_HPP_
+#define OSRM_EXTRACTOR_TRAFFIC_LIGHTS_DATA_HPP_
+
+namespace osrm
+{
+namespace extractor
+{
+
+namespace TrafficLightClass
+{
+// The traffic light annotation is extracted from node tags.
+// The directions in which the traffic light applies are relative to the way containing the node.
+enum Direction
+{
+    NONE = 0,
+    DIRECTION_ALL = 1,
+    DIRECTION_FORWARD = 2,
+    DIRECTION_REVERSE = 3
+};
+} // namespace TrafficLightClass
+
+} // namespace extractor
+} // namespace osrm
+
+#endif // OSRM_EXTRACTOR_TRAFFIC_LIGHTS_DATA_HPP_
--- a/include/extractor/traffic_signals.hpp
+++ b/include/extractor/traffic_signals.hpp
@@ -1,0 +1,28 @@
+#ifndef OSRM_EXTRACTOR_TRAFFIC_SIGNALS_HPP
+#define OSRM_EXTRACTOR_TRAFFIC_SIGNALS_HPP
+
+#include "util/typedefs.hpp"
+#include <unordered_set>
+
+#include <boost/unordered_set.hpp>
+
+namespace osrm
+{
+namespace extractor
+{
+
+struct TrafficSignals
+{
+    std::unordered_set<NodeID> bidirectional_nodes;
+    std::unordered_set<std::pair<NodeID, NodeID>, boost::hash<std::pair<NodeID, NodeID>>>
+        unidirectional_segments;
+
+    inline bool HasSignal(NodeID from, NodeID to) const
+    {
+        return bidirectional_nodes.count(to) > 0 || unidirectional_segments.count({from, to}) > 0;
+    }
+};
+} // namespace extractor
+} // namespace osrm
+
+#endif // OSRM_EXTRACTOR_TRAFFIC_SIGNALS_HPP
--- a/profiles/lib/traffic_signal.lua
+++ b/profiles/lib/traffic_signal.lua
@@ -1,0 +1,26 @@
+-- Assigns traffic light value to node as defined by
+-- include/extractor/traffic_lights.hpp
+
+local TrafficSignal = {}
+
+function TrafficSignal.get_value(node)
+    local tag = node:get_value_by_key("highway")
+    if "traffic_signals" == tag then
+        local direction = node:get_value_by_key("traffic_signals:direction")
+        if direction then
+            if "forward" == direction then
+                return traffic_lights.direction_forward
+            end
+            if "backward" == direction then
+                return traffic_lights.direction_reverse
+            end
+        end
+        -- return traffic_lights.direction_all
+        return true
+    end
+    -- return traffic_lights.none
+    return false
+end
+
+return TrafficSignal
+
--- a/src/extractor/graph_compressor.cpp
+++ b/src/extractor/graph_compressor.cpp
@@ -19,8 +19,10 @@ namespace osrm
 namespace extractor
 {
 
+static constexpr int SECOND_TO_DECISECOND = 10;
+
 void GraphCompressor::Compress(const std::unordered_set<NodeID> &barrier_nodes,
-                               const std::unordered_set<NodeID> &traffic_signals,
+                               const TrafficSignals &traffic_signals,
                                ScriptingEnvironment &scripting_environment,
                                std::vector<TurnRestriction> &turn_restrictions,
                                std::vector<UnresolvedManeuverOverride> &maneuver_overrides,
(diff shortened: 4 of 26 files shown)


include/extractor/traffic_lights.hpp
```diff
@@ -0,0 +1,23 @@
...
+#define OSRM_EXTRACTOR_TRAFFIC_LIGHTS_DATA_HPP_
+
+namespace osrm
+{
+namespace extractor
+{
+
+namespace TrafficLightClass
+{
+typedef std::uint8_t Direction;
```
SiarheiFedartsou, line 11: Nit. `using Direction = uint8_t;`

and why not make it enum btw?

profiles/lib/traffic_signal.lua
```diff
@@ -0,0 +1,25 @@
+local traffic_lights = traffic_lights
```
SiarheiFedartsou, line 1: Nit. I may be wrong, but it seems this is redundant(but probably I would somehow highlight that this enum declared from outside in C++ code)

src/extractor/graph_compressor.cpp
```diff
@@ -245,8 +249,17 @@ void GraphCompressor::Compress(const std::unordered_set<NodeID> &barrier_nodes,
                                                    roads_on_the_right,
                                                    roads_on_the_left);
                     scripting_environment.ProcessTurn(extraction_turn);
-                    node_duration_penalty = extraction_turn.duration * 10;
-                    node_weight_penalty = extraction_turn.weight * weight_multiplier;
+
+                    if (has_forward_signal)
```
SiarheiFedartsou, line 253: Nit: Let’s consider extracting small function to avoid code duplication here. Also for me 10 feels as magic number here - may be extract to constant?

include/extractor/traffic_signals.hpp
```diff
@@ -0,0 +1,28 @@
...
+namespace extractor
+{
+
+struct TrafficSignals
+{
+    std::unordered_set<NodeID> bidirectional_nodes;
+    std::unordered_set<std::pair<NodeID, NodeID>, boost::hash<std::pair<NodeID, NodeID>>>
+        unidirectional_segments;
+
+    inline bool hasSignal(NodeID from, NodeID to) const
```
SiarheiFedartsou, line 20: Nit. Not sure we follow such code style everywhere but it seems it should be HasSignal

SiarheiFedartsou (approved):
LGTM 🚀 please take a look on my comments

include/extractor/traffic_lights.hpp
mjjbell: I've just copied an existing type that interfaces with lua. 
I didn't actually explore if enums work. I'll take a look.

profiles/lib/traffic_signal.lua
mjjbell: I've no idea where that's come from, will remove.

include/extractor/traffic_lights.hpp
mjjbell: Yep, it appears to work :+1:

GitBenjamin:
Hi, @mjjbell. 
I did a lot of testing and found that most traffic signal problems can be solved, but there are some problems with one type of scenario. Like this one, [osm_data](https://github.com/GitBenjamin/Some-test-file/blob/main/test.osm).

- Data

image
This is an intersection with a traffic signal in the center that acts in all four directions. 

- Process (about traffic signal)

According to this [solution](https://github.com/Project-OSRM/osrm-backend/issues/1317#issuecomment-1259679903), I copy the traffic signal to the middle node of the four roads and add **"traffic_signals:direction"** info to traffic_signal node.

- Route Test

When I get to route _UnknownRoad_ to _NewRoad_, the duration is 13.5s, and when I get to route _NewRoad_ to _UnknownRoad_, the duration is 33.5s, traffic signal penalty is 20s. The two schemes covered the same number of miles and traffic lights, but took different time.

So, I tweaked **"traffic_signals:direction"** a lot, but nothing worked out. I hope I can get your help.

GitBenjamin:
> Hi, @mjjbell, @SiarheiFedartsou. I did a lot of testing and found that most traffic signal problems can be solved, but there are some problems with one type of scenario. Like this one, [osm_data](https://github.com/GitBenjamin/Some-test-file/blob/main/test.osm).
> 
> * Data
> 
> image This is an intersection with a traffic signal in the center that acts in all four directions.
> 
> * Process (about traffic signal)
> 
> According to this [solution](https://github.com/Project-OSRM/osrm-backend/issues/1317#issuecomment-1259679903), I copy the traffic signal to the middle node of the four roads and add **"traffic_signals:direction"** info to traffic_signal node.
> 
> * Route Test
> 
> When I get to route _UnknownRoad_ to _NewRoad_, the duration is 13.5s, and when I get to route _NewRoad_ to _UnknownRoad_, the duration is 33.5s, traffic signal penalty is 20s. The two schemes covered the same number of miles and traffic lights, but took different time.
> 
> So, I tweaked **"traffic_signals:direction"** a lot, but nothing worked out. I hope I can get your help.

Hi, @mjjbell, @SiarheiFedartsou . I reviewed the code for the branch [Support OSM traffic signal directions](https://github.com/Project-OSRM/osrm-backend/pull/6153/commits/7a00cbd756d06b87229b3d4859e9c0554e2ad080), i found a bug:

- firstly, the line is used to determine whether there is a traffic signal between **node_u** and **node_v**.
https://github.com/Project-OSRM/osrm-backend/blob/31e31a63d062fb804f5f4695ed3036ca7a269ead/src/extractor/graph_compressor.cpp#L210

- secondly, the line is used to update relation of nodes after the node_v is compressed.
https://github.com/Project-OSRM/osrm-backend/blob/31e31a63d062fb804f5f4695ed3036ca7a269ead/src/extractor/graph_compressor.cpp#L331

- example

Traffic_signal_bug

As shown in the picture above, when i first compress node_C, there will be no traffic signal penalty for D->A route. Because when i compress node_B, _traffic_signals.HasSignal(node_A, node_B) and traffic_signals.HasSignal(node_D, node_B) are both **false**_. Unless we compress node_B first, or _graph.SetTarget(forward_e1, node_c)_ not executed when node_C is being compressed.

mjjbell:
Yes, you're right. The traffic signals need a compressor wrapper to handle compression events, similar to the way turn path relations handle it.
https://github.com/Project-OSRM/osrm-backend/blob/31e31a63d062fb804f5f4695ed3036ca7a269ead/src/extractor/graph_compressor.cpp#L339

GitBenjamin:
> Yes, you're right. The traffic signals need a compressor wrapper to handle compression events, similar to the way turn path relations handle it.
> 
> https://github.com/Project-OSRM/osrm-backend/blob/31e31a63d062fb804f5f4695ed3036ca7a269ead/src/extractor/graph_compressor.cpp#L339

I try fix a bug caused by support OSM traffic signal directions. I hope to get you help and review.

mjjbell:
@GitBenjamin using your example, I created a minimal test case that currently fails.

```
@routing @car @traffic_light
Feature: Car - Handle traffic lights

    Background:
        Given the profile "car"

    Scenario: Car - Traffic signal direction with edge compression
        Given the node map
            """
            a-1-b - c - d-2-e

            """

        And the ways
            | nodes | highway |
            | abcde | primary |

        And the nodes
            | node | highway         | traffic_signals:direction |
            | c    | traffic_signals | forward                   |

        When I route I should get
            | from | to | time   | weight | #                   |
            | 1    | 2  |  35.3s | 35.3   | with traffic light  |
            | 2    | 1  |  33.3s | 33.3   | no traffic light    |
```

Let's get your PR updated so that this test passes.

Redy1908:
Is this change present in the docker image? I am encountering identical errors as reported in issues [#6665](https://github.com/Project-OSRM/osrm-backend/issues/6665) and [#6663](https://github.com/Project-OSRM/osrm-backend/issues/6663):

 `traffic_signal.lua:12: attempt to index global 'traffic_lights' (a nil value)`

These errors occur only when I manually import `profiles/lib/traffic_signal.lua` into `opt/lib`. Otherwise i get `traffic_signal.lua file not found`. I have also verified that the file is effectively not present within the container.

Below is the Dockerfile I am using:

```Dockerfile
FROM osrm/osrm-backend

COPY ./data /data
COPY profiles/car.lua /opt

ENV OSRM_FILE=sud-latest

RUN osrm-extract -p /opt/car-standard.lua /data/${OSRM_FILE}.osm.pbf && \
        osrm-partition /data/${OSRM_FILE}.osrm && \
        osrm-customize /data/${OSRM_FILE}.osrm;

EXPOSE 5000

CMD osrm-routed --algorithm mld /data/${OSRM_FILE}.osrm
```

mjjbell:
@Redy1908 can you confirm what version of `osrm-extract` and the `car.lua` profile you are using?
As discussed in #6663, this can happen when you use a newer profile on and older execuatable.

Redy1908:
I am currently using this [profile](https://github.com/Project-OSRM/osrm-backend/blob/master/profiles/car.lua) with the most recent Docker image, [osrm/osrm-backend](https://hub.docker.com/r/osrm/osrm-backend/tags) which as you said seems to not have this change included. Therefore, if I wish to continue using the Docker image, I should either revert to an older profile or remove the related bits of code to this commit from the profile right?

mjjbell:
> I am currently using this [profile](https://github.com/Project-OSRM/osrm-backend/blob/master/profiles/car.lua) with the most recent Docker image, [osrm/osrm-backend](https://hub.docker.com/r/osrm/osrm-backend/tags) which as you said seems to not have this change included. Therefore, if I wish to continue using the Docker image, I should either revert to an older profile or remove the related bits of code to this commit from the profile right?

The latest images are hosted on Github: https://github.com/Project-OSRM/osrm-backend/pkgs/container/osrm-backend
Docker Hub's last image was 5.25.0, so your case is an example of newer profile with older version.

If you are using a default profile, they come bundled with the container in `/opt/profiles`

See the [docker instructions ](https://github.com/Project-OSRM/osrm-backend?tab=readme-ov-file#using-docker)for full details.

Redy1908:
> The latest images are hosted on Github: https://github.com/Project-OSRM/osrm-backend/pkgs/container/osrm-backend
Docker Hub's last image was 5.25.0, so your case is an example of newer profile with older version.

I was missing that, thank you for your help.

PR #6170, opened 2021-12-11
Fix 32 bit CI builds

### LINKED ISSUE #6165: CI build  gcc-9-release-i686 is broken
https://github.com/Project-OSRM/osrm-backend/runs/4360200799?check_suite_focus=true

```
/usr/bin/ld.gold: error: /usr/lib/x86_64-linux-gnu/libexpat.so: incompatible target
```

`cmake` is finding the 64 bit version of the Expat library instead of the 32 bit version. 

During the CI migration to Github Actions, a bug was introduced such that Mason `cmake` is not being added to the `PATH` environment variable correctly, so it defaults to the `cmake` installed in the OS environment. This was recently auto-updated to version `3.22` in the Ubuntu Focal workers, leading to the regression in unrelated changes such as #6162

Fixing the use of Mason `cmake` also fails as it requires libssl 1.0 as a dependency, whilst the Ubuntu Focal runners are on the newer libssl 1.1

So the options to fix CI are to:
1. Understand the change in cmake 3.22 that leads to 32 bit libraries not being detected correctly and correct our CMakeLists.txt
2. Install libssl 1.0 on the Ubuntu CI workers
3. Add libssl 1.1 support to Mason cmake

1  seems like the preferred choice here.

mjjbell:
# Issue

This PR fixes two issues

1. Mason CMake was not being installed correctly due to a typo in the Github Actions script.
Once fixed, it also uncovers another problem. The version of Mason CMake requires libssl 1.0 as a dependency, whilst the Ubuntu Focal runners are on the newer libssl 1.1. Therefore, we also bump the version to `3.21.2` on CI.

2. A recent change to the Ubuntu Focal CI worker - possibly due to libexpat-dev being installed by default - leads to the 32-bit expat library to not be found by CMake ([example](https://github.com/Project-OSRM/osrm-backend/runs/4360200799?check_suite_focus=true)). `FindPackage(EXPAT)` finds the library via `pkg-config`, so the fix is to explicitly include the i386 pkgconfig directory in the `PKG_CONFIG_PATH` environment variable list.

## Tasklist

 - [ ] review
 - [ ] adjust for comments

## Requirements / Relations

Fixes #6165

### DIFF
--- a/.github/workflows/osrm-backend.yml
+++ b/.github/workflows/osrm-backend.yml
@@ -471,6 +471,7 @@ jobs:
           fi
         elif [[ $TARGET_ARCH == "i686" ]]; then
           source ./scripts/ci/before_install.${TARGET_ARCH}.sh
+          echo "PKG_CONFIG_PATH=/usr/lib/i386-linux-gnu/pkgconfig:${PKG_CONFIG_PATH}" >> $GITHUB_ENV
         fi
 
     - name: Prepare build
--- a/.github/workflows/osrm-backend.yml
+++ b/.github/workflows/osrm-backend.yml
@@ -16,7 +16,7 @@ env:
   CCACHE_COMPRESS: 1
   CASHER_TIME_OUT: 599 # one second less than 10m to avoid 10m timeout error: https://github.com/Project-OSRM/osrm-backend/issues/2742
   CCACHE_VERSION: 3.3.1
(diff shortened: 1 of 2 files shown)


mjjbell:
PTAL @danpat @TheMarex . Needed to unblock CI

TheMarex (approved):

PR #6185, opened 2021-12-29
Replace boost/function_output_iterator.hpp with boost/iterator/function_output_iterator.hpp

georgthegreat:
The former header is deprecated and causes a bunch of `#pragma message` to be issued during build.

The latter is present in at lease boost=1.67 which is at least 3 years old by now.

### DIFF
--- a/include/storage/serialization.hpp
+++ b/include/storage/serialization.hpp
@@ -10,7 +10,7 @@
 #include "storage/tar.hpp"
 
 #include <boost/assert.hpp>
-#include <boost/function_output_iterator.hpp>
+#include <boost/iterator/function_output_iterator.hpp>
 #include <boost/iterator/function_input_iterator.hpp>
 
 #include <cmath>
--- a/include/storage/shared_data_index.hpp
+++ b/include/storage/shared_data_index.hpp
@@ -3,7 +3,7 @@
 
 #include "storage/shared_datatype.hpp"
 
-#include <boost/function_output_iterator.hpp>
+#include <boost/iterator/function_output_iterator.hpp>
 
 #include <type_traits>
 #include <unordered_map>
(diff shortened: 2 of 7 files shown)


nilsnolde:
it seems it got introduced in boost v1.66. is it ok if you protect it with boost version checks? my reason would be: if you want to do python packaging for linux you need old systems and one of the latest ones you can use comes with boost 1.62. it'd be cool to keep that compatibility IMO.

georgthegreat:
Not sure if this particular block needs this check.
It's better to abandon PR and continue using deprecated header.

Checking boost version would require adding 6 lines in each place

```cpp
#include <boost/version.hpp>
#if BOOST_VERSION > 10066
#include <new_header>
#else
#include <old_header>
#endif
```
I'd like to keep the code simple. Is there any kind of list with supported platforms?

nilsnolde:
I'm not aware of an official list of supported distributions. also note, I'm not a osrm maintainer and also I still have to find the time to come up with at least high-level python bindings for osrm, so actually there's no use case yet, for me at least. there already (kinda unmaintained) [pyosrm](https://github.com/enricodvn/pyosrm), which would also run into problems I guess). I don't feel I have the authority to decide smth here, others would be more appropriate;)

mjjbell:
We're still supporting Boost versions as low as 1.60 for non HTTP server builds.

https://github.com/Project-OSRM/osrm-backend/blob/c0b79a8d0e518d297845815eedcbb4548d52f115/CMakeLists.txt#L529

georgthegreat:
I have checked boost releases.
`boost/iterator/function_output_iterator.hpp` was [released](https://github.com/boostorg/iterator/tree/boost-1.65.0/include/boost/iterator) in boost=1.65.

As OSRM tends to support boosts up to 1.60, there is no solution for the issue and I have to close this PR.

PR #6186, opened 2021-12-29
Do not depend on ADL

georgthegreat:


mjjbell:
Can you run `clang-format` on the file?

See https://github.com/Project-OSRM/osrm-backend/runs/4755342318?check_suite_focus=true

georgthegreat:
@mjjbell, I've fixed it.

mjjbell (approved):

PR #6212, opened 2022-02-27
Bicycle surface speeds

32Vache:
# Issue

Tweaked some of the bicycle surface speeds and added more surfaces.
From my own bicycling experience I think those speeds are more accurate. Ground surfaces for example are not very bad, 6km/h here was too low.

## Tasklist

 - [x] review
 - [ ] adjust for comments


profiles/bicycle.lua
```diff
@@ -173,21 +173,25 @@ function setup()
 
     surface_speeds = {
       asphalt = default_speed,
+      chipseal = default_speed,
```
SiarheiFedartsou, line 176: @mjjbell wdyt? At least adding new surfaces seems to be useful, not sure about changed speeds though.

mjjbell: Seems reasonable :+1:

PR #6215, opened 2022-03-05
Improvements to maneuver override processing

### LINKED ISSUE #6126: osrm-extract assert triggering on manoeuvre override relation
With the [Kentucky OSM extract](https://download.geofabrik.de/north-america/us/kentucky.html).

`osrm-extract` with the default bike profile triggers an assertion when compressing a manoeuvre override.

https://github.com/Project-OSRM/osrm-backend/blob/051e931091c19ce6f38c86202827e8ecd061aaa8/src/extractor/restriction_compressor.cpp#L228

The override is applied at an intersection between a private and public road: https://www.openstreetmap.org/relation/10547836#map=16/39.0284/-84.5417
Screenshot 2021-09-16 at 14 53 51

The private road is not accessible to the bike profile, so the node has degree two, allowing it to be compressed.
The via node of manoeuvre overrides are not exempt from being compressed, although the code appears to expect this to be the case. This inconsistency leads to the triggering of the assertion.

Furthermore, manoeuvre overrides do not support non-compressible nodes as part of the paths, meaning many are skipped or ignored. This was fixed for turn restrictions, a similar change is required for manoeuvres.

Other than fixing debug builds, there are only ~100 overrides in OSM data, so the impact on routing will be low.

mjjbell:
# Issue

This change unblocks the `osrm-extract` debug build, which is [currently failing](https://github.com/Project-OSRM/osrm-backend/issues/6126) on a maneuver override assertion.

The processing of maneuver overrides currently has three issues:
- It assumes the via node(s) can't be compressed (the failing assertion)
- It can't handle via-paths containing incompressible nodes
- It doesn't interop with turn restrictions on the same path

Turn restrictions and maneuver overrides both use the same `from-via-to` path representation.
Therefore, we can fix these issues by consolidating their structures and reusing the path representation for turn restrictions, which already is robust to the above issues.

This also simplifies some of the codebase by removing maneuver override specific path processing.

There are ~100 maneuver overrides in the OSM database, so the impact on processing and routing will be minimal.
But who knows, with improved support, it might encourage further adoption.

## Tasklist

 - [ ] review
 - [ ] adjust for comments

## Requirements / Relations

Fixes #6126


SiarheiFedartsou (approved):
LGTM, it is quite hard to review such PRs though - have no deep experience with OSRM codebase yet

PR #6270, opened 2022-06-29
Enable more clang-tidy checks.

SiarheiFedartsou:
# Issue

It is continuation of https://github.com/Project-OSRM/osrm-backend/pull/6261

## Tasklist

 - [x] CHANGELOG.md entry ([How to write a changelog entry](http://keepachangelog.com/en/1.0.0/#how))
 - [ ] update relevant [Wiki pages](https://github.com/Project-OSRM/osrm-backend/wiki)
 - [ ] add tests (see [testing documentation](https://github.com/Project-OSRM/osrm-backend/blob/master/docs/testing.md))
 - [ ] review
 - [ ] adjust for comments
 - [ ] cherry pick to release branch

## Requirements / Relations

 Link any requirements here. Other pull requests this PR is based on?

PR #6261, opened 2022-06-16
Configure clang-tidy job on CI

SiarheiFedartsou:
# Issue

Configures job on CI which runs clang-tidy on each pull request. Many of useful checks are disabled for the time being, because it was non-trivial to fix right in this PR, but it can be addressed later.

## Tasklist

 - [x] CHANGELOG.md entry ([How to write a changelog entry](http://keepachangelog.com/en/1.0.0/#how))
 - [ ] update relevant [Wiki pages](https://github.com/Project-OSRM/osrm-backend/wiki)
 - [ ] add tests (see [testing documentation](https://github.com/Project-OSRM/osrm-backend/blob/master/docs/testing.md)
 - [ ] review
 - [ ] adjust for comments
 - [ ] cherry pick to release branch

## Requirements / Relations

 Link any requirements here. Other pull requests this PR is based on?

PR #6273, opened 2022-07-02
Enable even more clang-tidy checks

SiarheiFedartsou:
# Issue

Continuation of https://github.com/Project-OSRM/osrm-backend/pull/6270 and https://github.com/Project-OSRM/osrm-backend/pull/6261
## Tasklist

 - [ ] review
 - [ ] adjust for comments

## Requirements / Relations


include/contractor/query_graph.hpp
```diff
@@ -15,7 +15,7 @@ namespace detail
 {
 template <storage::Ownership Ownership>
 using QueryGraph = util::StaticGraph<typename QueryEdge::EdgeData, Ownership>;
-}
+} // namespace detail
```
SiarheiFedartsou, line 18: clang-tidy has a check which fires if we skip it somewhere, I fixed some of them, but didn't enable it, because there are tons of such places and probably worth doing it in separate PR

src/extractor/extractor.cpp
```diff
@@ -531,7 +533,9 @@ std::
         });
 
     tbb::filter_t<SharedBuffer, std::shared_ptr<ExtractionRelationContainer>> buffer_relation_cache(
-        tbb::filter::parallel, [&](const SharedBuffer buffer) {
+        tbb::filter::parallel,
+        // NOLINTNEXTLINE(performance-unnecessary-value-param)
+        [&](const SharedBuffer buffer) {
```
SiarheiFedartsou, line 538: Was not sure if here we should pass it by reference because in case of multithreading often things are intentionally copied, so decided to just suppress it.

SiarheiFedartsou:
@mjjbell please take a look when you have a chance 🙂

include/extractor/intersection_bearings_container.hpp
```diff
@@ -46,9 +46,9 @@ template <storage::Ownership Ownership> class IntersectionBearingsContainer
     IntersectionBearingsContainer &operator=(IntersectionBearingsContainer &&) = default;
     IntersectionBearingsContainer &operator=(const IntersectionBearingsContainer &) = default;
 
-    IntersectionBearingsContainer(std::vector<BearingClassID> node_to_class_id_,
+    IntersectionBearingsContainer(const std::vector<BearingClassID> &node_to_class_id,
```
mjjbell, line 49: What was the error message here? Here we guarantee a copy.
By having this as by-value, we can avoid a copy by doing two `move` operations. 

It looks like the code isn't currently taking advantage of this, but it could.
https://github.com/Project-OSRM/osrm-backend/blob/3d2db20777a06ec064876f6c7526f485e36209ea/src/extractor/extractor.cpp#L863

src/extractor/intersection/mergable_road_detector.cpp
```diff
@@ -419,10 +419,10 @@ bool MergableRoadDetector::HaveSameDirection(const NodeID intersection_node,
     // sampling to correctly weight longer segments in regression calculations
     const auto constexpr SAMPLE_INTERVAL = 5;
     coordinates_to_the_left = coordinate_extractor.SampleCoordinates(
-        std::move(coordinates_to_the_left), distance_to_extract, SAMPLE_INTERVAL);
+        coordinates_to_the_left, distance_to_extract, SAMPLE_INTERVAL);
```
mjjbell, line 422: :+1:

include/extractor/intersection_bearings_container.hpp
SiarheiFedartsou: It was `performance-unnecessary-value-param` it seems it is because this `template <typename T> using Vector = util::ViewOrVector<T, Ownership>;` does not have constructor which accepts `std::vector<...>&&`, but tbh I am not sure and decided to add `NOLINT...` for the time being to preserve a status quo here.

src/extractor/extractor.cpp
SiarheiFedartsou: Do you think I should fix it here too? IMO it is still a bit risky, but won't give any significant benefits.

include/extractor/intersection_bearings_container.hpp
mjjbell: I see. The underlying container will depend on the ownership, so it can end up being `std::vector`. clang-tidy might not be able to figure it out.

src/extractor/extractor.cpp
mjjbell: I'd suggest leaving it as it is, unless you want to dig into the TBB internals to confirm.

include/extractor/intersection_bearings_container.hpp
mjjbell: Also, I can omit the linting comment on both my dev environments (Big Sur-clang14, Ubuntu 21.04-clang12) without error, so it does seem a bit unsure its relevance.

mjjbell (approved):
:+1:

include/extractor/intersection_bearings_container.hpp
SiarheiFedartsou: > Also, I can omit the linting comment on both my dev environments (Big Sur-clang14, Ubuntu 21.04-clang12) without error, so it does seem a bit unsure of its relevance.

It doesn't fire for me locally too, but does fire on CI.

mjjbell: Hmm, it must be a bug in that version. Even when I change the only invocation to be moving an r-value (see #6275), we get the same error. In any case, we'll need the nolint comment.

Issue #6266, opened 2022-06-21
Mason is unmaintained


nilsnolde:
I'll leave that here as well: https://github.com/mapbox/mason. doesn't mean it's going away or will be broken anytime soon, but since OSRM [is slowly getting really out of date on dependencies](https://github.com/Project-OSRM/osrm-backend/issues/6264), sooner or later another solution will have to be found. 

from my experience, [`conan`](https://conan.io/center/tbb) could be ok, been using it a few times. on first glance tbb is from 2019, could be working still. rest of the libs should be less of a problem.

SiarheiFedartsou:
I started migration to Conan here https://github.com/Project-OSRM/osrm-backend/pull/6284 
Would be great to have some initial feedback.

nilsnolde:
Thanks a ton @SiarheiFedartsou! I’d be happy to give this a shot and bits of review where I can (no write permission to this repo though). This has the potential to get rid of the crappy windows deps setup and let us install it all with conan!

SiarheiFedartsou:
Can be closed as we merged https://github.com/Project-OSRM/osrm-backend/pull/6284

mjjbell:
Let's keep it open until we've completely removed it. It's still being used to install CI worker dependencies.

PR #6278, opened 2022-07-09
Remove Boost.Regex from the list of dependencies

SiarheiFedartsou:
# Issue

Noticed that we do not use it.

## Tasklist

 - [ ] review
 - [ ] adjust for comments

## Requirements / Relations


mjjbell:
I've also tried this [in the past](https://github.com/Project-OSRM/osrm-backend/pull/5901#discussion_r640162103). 

The Windows build gets confused if the Boost regex library is in the dependency directory but is not included in the build. 
It needs to be removed from the zip file containing the dependencies, but it's hosted in a Wolt S3 bucket. I'm not sure who has access.

I think it's probably better to put effort into tackling #6266, and including the Windows build in any solution.

SiarheiFedartsou:
> I've also tried this [in the past](https://github.com/Project-OSRM/osrm-backend/pull/5901#discussion_r640162103).
> 
> The Windows build gets confused if the Boost regex library is in the dependency directory but is not included in the build. It needs to be removed from the zip file containing the dependencies, but it's hosted in a Wolt S3 bucket. I'm not sure who has access.
> 
> I think it's probably better to put effort into tackling #6266, and including the Windows build in any solution.

Okay, I've been already experimenting with Conan. :+1:

SiarheiFedartsou:
> > I've also tried this [in the past](https://github.com/Project-OSRM/osrm-backend/pull/5901#discussion_r640162103).
> > The Windows build gets confused if the Boost regex library is in the dependency directory but is not included in the build. It needs to be removed from the zip file containing the dependencies, but it's hosted in a Wolt S3 bucket. I'm not sure who has access.
> > I think it's probably better to put effort into tackling #6266, and including the Windows build in any solution.
> 
> Okay, I've been already experimenting with Conan. 👍

As promised :) https://github.com/Project-OSRM/osrm-backend/pull/6284

PR #6279, opened 2022-07-09
Migrate to C++17. Update sol2 to 3.3.0. Fix bug with reading Set values from Lua scripts.

SiarheiFedartsou:
# Issue
This PR migrates OSRM to C++17. Also I had to update sol2 to 3.3.0, because our ancient 2.x version couldn't be built under C++17.

We could upgrade to even C++20, but it would require rework of CI.

## Tasklist

 - [ ] review
 - [ ] adjust for comments

## Requirements / Relations


src/extractor/scripting_environment_lua.cpp
```diff
@@ -945,10 +945,6 @@ Sol2ScriptingEnvironment::GetStringListFromTable(const std::string &table_name)
     auto &context = GetSol2Context();
     BOOST_ASSERT(context.state.lua_state() != nullptr);
     std::vector<std::string> strings;
-    if (!context.profile_table[table_name])
```
SiarheiFedartsou, line 948: otherwise we have crash in runtime on some assertion inside sol2 that it expected boolean, but got table, i.e. it probably infers expected type from context(in our case `if` condition) and assert if real type corresponding it. As understand here our goal was to check that we really have this table in script, but this case is fully covered by condition a couple of lines below(`if (table.valid())`).

src/extractor/scripting_environment_lua.cpp
```diff
@@ -1171,14 +1163,14 @@ void LuaScriptingContext::ProcessNode(const osmium::Node &node,
     {
     case 4:
     case 3:
-        node_function(profile_table, node, result, relations);
```
SiarheiFedartsou, line 1174: Otherwise it complained that `Node` has no copy constructor and it cannot copy it.

SiarheiFedartsou:
Any ideas why do I have npm tests failing in runtime on CI on some jobs?

src/extractor/scripting_environment_lua.cpp
```diff
@@ -939,22 +939,37 @@ Sol2ScriptingEnvironment::GetStringListFromFunction(const std::string &function_
     return strings;
 }
 
+namespace
+{
+
+// string list can be defined either as a Set(see profiles/lua/set.lua) or as a Sequence (see
```
SiarheiFedartsou, line 945: It turned out we had a bug with `Set` values...

features/foot/restrictions.feature
```diff
@@ -29,7 +29,7 @@ Feature: Foot - Turn restrictions
         When I route I should get
             | from | to | route    |
             | s    | w  | sj,wj,wj |
-            | s    | n  | sj,nj,nj |
+            | s    | n  | sj,nj    |
```
SiarheiFedartsou, line 32: This test uses foot.lua profile, which in its turn uses `suffix_list` defined as `Set`. That's why fix of the bug with Set values made this test red.

https://github.com/Project-OSRM/osrm-backend/blob/06b1b980bbdfc838c4f8e8cb370cba6dfb0329b0/profiles/foot.lua#L67

CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
     - API:
       - FIXED: Fix inefficient osrm-routed connection handling [#6113](https://github.com/Project-OSRM/osrm-backend/pull/6113)
     - Build:
+      - CHANGED: Migrate to C++17. Update sol2 to 3.3.0. Fix bug with reading Set values from Lua scripts. [#6279](https://github.com/Project-OSRM/osrm-backend/pull/6279)
```
SiarheiFedartsou, line 6: Should I maybe extract bug fix into separate PR?

SiarheiFedartsou:
@mjjbell @danpat 
Please take a look when you have a chance. Please don't be afraid of the number of changes. Vast majority of them is in thirdparty/sol2:

Screenshot 2022-07-13 at 21 38 18

src/extractor/scripting_environment_lua.cpp
mjjbell: Nice catch!
I guess that means nobody has noticed the suffixes appearing in walking guidance (or nobody is using it).

features/foot/restrictions.feature
mjjbell: Hmm, it's not immediately obvious why this would change due to the bug fix, unless it's something very subtle in the code.

CHANGELOG.md
mjjbell: I think it would be good to do so. It's not clear to me how https://github.com/Project-OSRM/osrm-backend/pull/6279/files#r920390451 is related to the bug, so would be good to separate it from the other changes to understand.

SiarheiFedartsou: @mjjbell please see https://github.com/Project-OSRM/osrm-backend/pull/6285

mjjbell (approved):
:+1:

Zverik:
Looks like this PR invalidated the bundled Dockerfile. #6302

PR #6284, opened 2022-07-20
Use Conan instead of Mason to install code dependencies

SiarheiFedartsou:
# Issue

https://github.com/Project-OSRM/osrm-backend/issues/6266

Migration of CI dependencies(like clang-format) is out of scope of this PR, but seems can be done with a help of Conan too. 

## Tasklist

 - [ ] review
 - [ ] adjust for comments

## Requirements / Relations

### DIFF
--- a/.github/workflows/osrm-backend.yml
+++ b/.github/workflows/osrm-backend.yml
@@ -95,24 +95,24 @@ jobs:
             CUCUMBER_TIMEOUT: 60000
             ENABLE_CLANG_TIDY: ON
 
-          - name: mason-linux-debug-asan
+          - name: conan-linux-debug-asan
             continue-on-error: false
             node: 12
             runs-on: ubuntu-20.04
             BUILD_TOOLS: ON
             BUILD_TYPE: Release
             CLANG_VERSION: 5.0.0
-            ENABLE_MASON: ON
+            ENABLE_CONAN: ON
             ENABLE_SANITIZER: ON
 
-          - name: mason-linux-release
+          - name: conan-linux-release
             continue-on-error: false
             node: 12
             runs-on: ubuntu-20.04
             BUILD_TOOLS: ON
             BUILD_TYPE: Release
             CLANG_VERSION: 5.0.0
-            ENABLE_MASON: ON
+            ENABLE_CONAN: ON
 
           - name: gcc-11-release
             continue-on-error: false
@@ -173,7 +173,7 @@ jobs:
             CCOMPILER: gcc-7
             CXXCOMPILER: g++-7
 
-          - name: mason-osx-release-node-12
+          - name: conan-osx-release-node-12
             build_node_package: true
             continue-on-error: false
             node: 12
@@ -184,9 +184,9 @@ jobs:
             CXXCOMPILER: clang++
             CUCUMBER_TIMEOUT: 60000
             ENABLE_ASSERTIONS: ON
-            ENABLE_MASON: ON
+            ENABLE_CONAN: ON
 
-          - name: mason-osx-release-node-14
+          - name: conan-osx-release-node-14
             build_node_package: true
             continue-on-error: false
             node: 14
@@ -197,9 +197,9 @@ jobs:
             CXXCOMPILER: clang++
             CUCUMBER_TIMEOUT: 60000
             ENABLE_ASSERTIONS: ON
-            ENABLE_MASON: ON
+            ENABLE_CONAN: ON
 
-          - name: mason-osx-release-node-16
+          - name: conan-osx-release-node-16
             build_node_package: true
             continue-on-error: false
             node: 16
@@ -210,7 +210,7 @@ jobs:
             CXXCOMPILER: clang++
             CUCUMBER_TIMEOUT: 60000
             ENABLE_ASSERTIONS: ON
-            ENABLE_MASON: ON
+            ENABLE_CONAN: ON
 
           - name: gcc-7-release-shared
             continue-on-error: false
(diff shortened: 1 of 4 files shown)


CMakeLists.txt
```diff
@@ -593,19 +566,11 @@ if(ENABLE_CCACHE AND (CMAKE_CXX_COMPILER_ID MATCHES "Clang" OR CMAKE_CXX_COMPILE
   endif()
 endif()
 
-# even with mason builds we want to link to system zlib
+# even with conan builds we want to link to system zlib
 # to ensure that osrm binaries play well with other binaries like nodejs
 find_package(ZLIB REQUIRED)
 add_dependency_includes(${ZLIB_INCLUDE_DIRS})
 
-if(NOT WIN32 AND NOT Boost_USE_STATIC_LIBS)
```
SiarheiFedartsou, line 601: We have this here https://github.com/Project-OSRM/osrm-backend/blob/06b1b980bbdfc838c4f8e8cb370cba6dfb0329b0/unit_tests/CMakeLists.txt#L131

CMakeLists.txt
```diff
@@ -1,5 +1,11 @@
 cmake_minimum_required(VERSION 3.2)
 
+enable_language(CXX)
```
SiarheiFedartsou, line 3: Otherwise Conan doesn't work.

.github/workflows/osrm-backend.yml
```diff
@@ -500,7 +502,7 @@ jobs:
         echo "Using ${JOBS} jobs"
         pushd ${OSRM_BUILD_DIR}
         cmake .. -DCMAKE_BUILD_TYPE=${BUILD_TYPE} \
-                 -DENABLE_MASON=${ENABLE_MASON:-OFF} \
+                 -DENABLE_CONAN=${ENABLE_MASON:-OFF} \
```
mjjbell, line 505: s/MASON/CONAN/

CMakeLists.txt
```diff
@@ -1,5 +1,11 @@
...
+enable_language(C)
+if (NOT CMAKE_BUILD_TYPE)
```
mjjbell, line 5: We're already doing this here:
https://github.com/Project-OSRM/osrm-backend/blob/06b1b980bbdfc838c4f8e8cb370cba6dfb0329b0/CMakeLists.txt#L230-L233

CMakeLists.txt
mjjbell: The CMake docs say [this is the same](https://cmake.org/cmake/help/latest/command/enable_language.html) as using the `project()` command, which we already do here:
https://github.com/Project-OSRM/osrm-backend/blob/06b1b980bbdfc838c4f8e8cb370cba6dfb0329b0/CMakeLists.txt#L78

Any idea why Conan specifically needs this?

CMakeLists.txt
```diff
@@ -324,8 +304,9 @@ if(CMAKE_BUILD_TYPE MATCHES Release OR CMAKE_BUILD_TYPE MATCHES MinRelSize OR CM
   endif()
 endif()
 
-if(UNIX AND NOT APPLE AND ENABLE_MASON AND (LTO_WORKS OR ENABLE_GOLD_LINKER))
-  message(WARNING "ENABLE_MASON and ENABLE_LTO/ENABLE_GOLD_LINKER may not work on all linux systems currently")
+# TODO: this was added for Mason, we are not sure if it is true after migration to Conan
+if(UNIX AND NOT APPLE AND ENABLE_CONAN AND (LTO_WORKS OR ENABLE_GOLD_LINKER))
+  message(WARNING "ENABLE_CONAN and ENABLE_LTO/ENABLE_GOLD_LINKER may not work on all linux systems currently")
   message(WARNING "For more details see: https://github.com/Project-OSRM/osrm-backend/issues/3202")
```
mjjbell, line 310: Let's remove the second message as the issue is Mason specific.

CMakeLists.txt
mjjbell, line 601: These many still be needed for non-Conan builds?

CMakeLists.txt
```diff
@@ -830,7 +795,7 @@ JOIN("-I${DEPENDENCIES_INCLUDE_DIRS}" " -I" PKGCONFIG_OSRM_INCLUDE_FLAGS)
 # Boost uses imported targets, we need to use a generator expression to extract
 # the link libraries to be written to the pkg-config file.
 foreach(engine_lib ${ENGINE_LIBRARIES})
-  if("${engine_lib}" MATCHES "^Boost.*")
+  if("${engine_lib}" MATCHES "^Boost.*" OR "${engine_lib}" MATCHES "^CONAN_LIB.*")
```
mjjbell, line 798: What do the`CONAN_LIB.` targets represent?

mjjbell (commented):
Thanks for working on this :+1: 

I think it would be nice to have a [separate file](https://github.com/conan-io/cmake-conan#conanfile) for defining the dependency versions and Conan config (also cleaning up the CMake file a bit). What do you think?

CMakeLists.txt
SiarheiFedartsou: Its CMake targets for dependencies defined by Conan, otherwise we have problems with build of example. It tries to find libraries with name like CONAN_LIB::BZip2 and cannot. I guess we had the same issue with Boost in the past. https://github.com/Project-OSRM/osrm-backend/blob/06b1b980bbdfc838c4f8e8cb370cba6dfb0329b0/example/CMakeLists.txt#L34

I added a comment to make it more clear.

mjjbell (approved):
Nice work :+1: 

Please add a Changelog entry, and then I'll merge.

PR #6285, opened 2022-07-23
Fix bug with reading Set values from Lua scripts.

SiarheiFedartsou:
# Issue

Extracted from https://github.com/Project-OSRM/osrm-backend/pull/6279
## Tasklist

 - [ ] review
 - [ ] adjust for comments

## Requirements / Relations

### DIFF
--- a/features/foot/restrictions.feature
+++ b/features/foot/restrictions.feature
@@ -29,7 +29,7 @@ Feature: Foot - Turn restrictions
         When I route I should get
             | from | to | route    |
             | s    | w  | sj,wj,wj |
-            | s    | n  | sj,nj,nj |
+            | s    | n  | sj,nj    |
             | s    | e  | sj,ej,ej |
 
     @only_turning
@@ -55,7 +55,7 @@ Feature: Foot - Turn restrictions
         When I route I should get
             | from | to | route    |
             | s    | w  | sj,wj,wj |
-            | s    | n  | sj,nj,nj |
+            | s    | n  | sj,nj    |
             | s    | e  | sj,ej,ej |
 
     @except
--- a/src/extractor/scripting_environment_lua.cpp
+++ b/src/extractor/scripting_environment_lua.cpp
@@ -939,6 +939,26 @@ Sol2ScriptingEnvironment::GetStringListFromFunction(const std::string &function_
     return strings;
 }
 
+namespace
+{
+
+// string list can be defined either as a Set(see profiles/lua/set.lua) or as a Sequence (see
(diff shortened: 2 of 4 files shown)


src/extractor/scripting_environment_lua.cpp
```diff
@@ -939,6 +939,26 @@ Sol2ScriptingEnvironment::GetStringListFromFunction(const std::string &function_
...
 
+namespace
+{
+
+// string list can be defined either as a Set(see profiles/lua/set.lua) or as a Sequence (see
+// profiles/lua/sequence.lua) `Set` is a table with keys that are actual values we are looking for
+// and values that always `true`. `Sequence` is a table with keys that are indices and values that
+// are actual values we are looking for.
+
+std::string GetSetOrSequenceValue(const std::pair<sol::object, sol::object> &pair)
```
SiarheiFedartsou, line 950: https://github.com/Project-OSRM/osrm-backend/pull/6279/files#r920389235

features/foot/restrictions.feature
```diff
@@ -29,7 +29,7 @@ Feature: Foot - Turn restrictions
         When I route I should get
             | from | to | route    |
             | s    | w  | sj,wj,wj |
-            | s    | n  | sj,nj,nj |
+            | s    | n  | sj,nj    |
```
SiarheiFedartsou, line 32: Note that this test will start failing if we just change Set to Sequence here (without changes from this PR) https://github.com/Project-OSRM/osrm-backend/blob/06b1b980bbdfc838c4f8e8cb370cba6dfb0329b0/profiles/foot.lua#L67

mjjbell:
Thanks for separating this out. 
I'll do some digging to fully understand what's going on with the tests.

mjjbell:
It turns out that the suffixes listed in the profile are used for matching suffixes **and** prefixes

https://github.com/Project-OSRM/osrm-backend/blob/1a3d0f7c2077bc89dc47d2d03567cb78412b5628/include/util/guidance/name_announcements.hpp#L132-L144

In the test we have roads `sj` `nj` `ej` and `wj`. It filters out the common substring `j` and then decides the names of inbound and outbound roads to a turn are "the same" if the remaining prefix and suffix values are in the profile set. This is the case here (`n`, `s`, `e`, `w`), and wasn't before (`true`?).

In the case where the turn is straight ahead and the names are "the same", it will remove the announcement from the guidance, hence the change in the test route output.

I don't think it's worth changing any of this logic. This test road names are not representative of real data, so it shouldn't be replicated in OSM. I will though add a test that shows how the bugfix produces the expected turn collapse behaviour in guidance output.

mjjbell (approved):

PR #6290, opened 2022-07-28
Configure Undefined Behaviour Sanitizer

SiarheiFedartsou:
# Issue

Configures [undefined behaviour sanitizer](https://clang.llvm.org/docs/UndefinedBehaviorSanitizer.html). Reports trivial to fix were fixed, other ones are suppressed.

## Tasklist

 - [ ] review
 - [ ] adjust for comments

## Requirements / Relations

### DIFF
--- a/.github/workflows/osrm-backend.yml
+++ b/.github/workflows/osrm-backend.yml
@@ -64,7 +64,7 @@ jobs:
             CXXCOMPILER: g++-9
             ENABLE_COVERAGE: ON
 
-          - name: gcc-9-debug-asan
+          - name: gcc-9-debug-asan-ubsan
             continue-on-error: false
             node: 12
             runs-on: ubuntu-20.04
@@ -74,7 +74,9 @@ jobs:
             CUCUMBER_TIMEOUT: 20000
             CXXCOMPILER: g++-9
             ENABLE_SANITIZER: ON
-            TARGET_ARCH: x86_64-asan
+            TARGET_ARCH: x86_64-asan-ubsan
+            OSRM_CONNECTION_RETRIES: 10
+            OSRM_CONNECTION_EXP_BACKOFF_COEF: 1.5
 
           - name: clang-5.0-debug
             continue-on-error: false
@@ -95,13 +97,13 @@ jobs:
             CUCUMBER_TIMEOUT: 60000
             ENABLE_CLANG_TIDY: ON
 
-          - name: mason-linux-debug-asan
+          - name: mason-linux-debug-asan-ubsan
             continue-on-error: false
             node: 12
             runs-on: ubuntu-20.04
             BUILD_TOOLS: ON
             BUILD_TYPE: Release
-            CLANG_VERSION: 5.0.0
+            CLANG_VERSION: 11.0.0
             ENABLE_MASON: ON
             ENABLE_SANITIZER: ON
 
@@ -382,6 +384,8 @@ jobs:
       ENABLE_SANITIZER: ${{ matrix.ENABLE_SANITIZER }}
       NODE_PACKAGE_TESTS_ONLY: ${{ matrix.NODE_PACKAGE_TESTS_ONLY }}
       TARGET_ARCH: ${{ matrix.TARGET_ARCH }}
+      OSRM_CONNECTION_RETRIES: ${{ matrix.OSRM_CONNECTION_RETRIES }}
+      OSRM_CONNECTION_EXP_BACKOFF_COEF: ${{ matrix.OSRM_CONNECTION_EXP_BACKOFF_COEF }}
     steps:
     - uses: actions/checkout@v2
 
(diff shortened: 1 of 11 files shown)


SiarheiFedartsou:
Hi @mjjbell 
Sorry for pinging you too often, but kindly ask you to take a look please :)

include/extractor/extractor_config.hpp
```diff
@@ -84,14 +83,12 @@ struct ExtractorConfig final : storage::IOConfig
...
-    unsigned small_component_size;
+    unsigned requested_num_threads = 0;
+    unsigned small_component_size = 1000;
 
-    bool generate_edge_lookup;
-
-    bool use_metadata;
-    bool parse_conditionals;
-    bool use_locations_cache;
+    bool use_metadata = false;
```
SiarheiFedartsou, line 89: Curious why usual compiler warnings didn't catch that 🤔

src/util/log.cpp
```diff
@@ -91,8 +95,6 @@ Log::Log(LogLevel level_, std::ostream &ostream) : level(level_), stream(ostream
     }
 }
 
-Log::Log(LogLevel level_) : Log(level_, buffer) {}
```
SiarheiFedartsou, line 94: It seems this is considered as UB: we are using field of class itself(`buffer`) when calling constructor of this class.

.github/workflows/osrm-backend.yml
```diff
@@ -553,6 +554,12 @@ jobs:
       if: ${{ matrix.NODE_PACKAGE_TESTS_ONLY == 'ON' }}
       run: |
         npm run nodejs-tests
+    - name: Upload test logs
```
SiarheiFedartsou, line 557: Helpful if we want to see logs of nodejs tests.

.github/workflows/osrm-backend.yml
```diff
@@ -95,13 +95,13 @@ jobs:
...
 
-          - name: mason-linux-debug-asan
+          - name: mason-linux-debug-asan-ubsan
             continue-on-error: false
             node: 12
             runs-on: ubuntu-20.04
             BUILD_TOOLS: ON
-            BUILD_TYPE: Release
-            CLANG_VERSION: 5.0.0
+            BUILD_TYPE: Debug
```
SiarheiFedartsou, line 103: Or we probably should rename the job :)

.github/workflows/osrm-backend.yml
```diff
@@ -74,7 +74,8 @@ jobs:
             CUCUMBER_TIMEOUT: 20000
             CXXCOMPILER: g++-9
             ENABLE_SANITIZER: ON
-            TARGET_ARCH: x86_64-asan
+            TARGET_ARCH: x86_64-asan-ubsan
+            OSRM_CONNECTION_RETRIES: 30
```
SiarheiFedartsou, line 78: Some NodeJs tests became flaky after configuring this sanitizer, because as I understand OSRM startup time has increased and 10 retries we are using by default was not enough, so had to add this option and set it to 30.

include/extractor/extractor_config.hpp
```diff
@@ -84,14 +83,12 @@ struct ExtractorConfig final : storage::IOConfig
...
+    bool parse_conditionals = false;
+    bool use_locations_cache = true;
```
SiarheiFedartsou, line 91: Values are chosen as default values we use in [program_options](https://github.com/Project-OSRM/osrm-backend/blob/41dda32546399f1dc12af1de41668993de44c7dc/src/tools/extract.cpp#L66).

features/lib/osrm_loader.js
```diff
@@ -45,11 +45,11 @@ class OSRMBaseLoader{
         var retryCount = 0;
         let retry = (err) => {
           if (err) {
-            if (retryCount < 10) {
+            if (retryCount < this.scope.OSRM_CONNECTION_RETRIES) {
```
mjjbell, line 48: We could also add some exponential back-off, which might be more robust to fluctuations in CI startup times.

mjjbell (approved):
:+1: 

My only concern is the sanitizer job now taking an hour to run in CI.

I guess the activity in this repo is low enough that it's not going to slow down anyone's productivity though.

features/lib/osrm_loader.js
SiarheiFedartsou: Added `OSRM_CONNECTION_EXP_BACKOFF_COEF`, please take a look.

SiarheiFedartsou:
> 👍
> 
> My only concern is the sanitizer job now taking an hour to run in CI.
> 
> I guess the activity in this repo is low enough that it's not going to slow down anyone's productivity though.

Agree here. Also worth noting that it is only gcc based job has this slow down. Tbh I don't see a lot of value in having UBSAN configured for both GCC and UBSAN and we can simply introduce CMake flag to separately enable/disable ASAN and UBSAN and turn off it for GCC job. But not sure if worth spending time on it. WDYT?

mjjbell:
> Agree here. Also worth noting that it is only gcc based job has this slow down. Tbh I don't see a lot of value in having UBSAN configured for both GCC and UBSAN and we can simply introduce CMake flag to separately enable/disable ASAN and UBSAN and turn off it for GCC job. But not sure if worth spending time on it. WDYT?

I don't think it's worth spending time on. Some of the other jobs are already taking 40+ minutes, so it won't be a huge saving.
If it becomes a big hit on productivity then it can be explored, but I suspect that would also require finding a more performant CI service.

Issue #6297, opened 2022-08-07
Build from source Error: TBB not found.


handcath:
Hi,

Trying to build 5.26 on Mac OS and Ubuntu 22.04. I am getting the following error (on max os or ubuntu 22.04) when running cmake ..
The old fix (#6253) to build OSRM with the -DENABLE_MASON=ON does not work.

ERROR: Intel TBB NOT found!
-- Looked for Threading Building Blocks in /Library/Frameworks/Intel_TBB.framework/Versions
CMake Error at cmake/FindTBB.cmake:274 (message):
  Could NOT find TBB library.
Call Stack (most recent call first):
  CMakeLists.txt:518 (find_package)

nitko12:
Ran into the same issue, 

it seems they are migrating to Conan package manager instead of Mason.
https://github.com/Project-OSRM/osrm-backend/commit/589becbfec1f4ab19972bd2f445a8fdd0cc6ca25

You could use the last Mason commit until the docs are updated, that one works for me :)
https://github.com/Project-OSRM/osrm-backend/commit/4e8ee288d98bdf0e9725f11b5392f6276ede8bcb

Hope someone confirms this, possibly supplying Conan instructions, as my attempt building with it failed.

nitko12:
Hmm even
https://github.com/Project-OSRM/osrm-backend/commit/4e8ee288d98bdf0e9725f11b5392f6276ede8bcb
seems to have some bugs, ends up using too much memory and crashing upon extracting maps, 
commit before it doesn't.

SiarheiFedartsou:
> Hmm even [4e8ee28](https://github.com/Project-OSRM/osrm-backend/commit/4e8ee288d98bdf0e9725f11b5392f6276ede8bcb) seems to have some bugs, ends up using too much memory and crashing upon extracting maps, commit before it doesn't.

Could you elaborate here? What exactly are doing and have issues?

handcath:
[4e8ee28](https://github.com/Project-OSRM/osrm-backend/commit/4e8ee288d98bdf0e9725f11b5392f6276ede8bcb) appears to have successfully built, no reported memory usage issues, thank you.
However, when trying to run the example.cpp I get an include error:
`/usr/local/include/osrm/engine/api/match_parameters.hpp:31:10: fatal error: 'engine/api/route_parameters.hpp' file not found
#include "engine/api/route_parameters.hpp"
`

The `include/osrm/*` files don't seem to useable.

[...]

Issue #6181, opened 2021-12-20
Does not build with tbb 2021.4.0-1.1 on openSUSE_Tumbleweed


stoecker:
For the newer tbb a lot of filenames changed. FindTBB.cmake does not find it, because tbb/task_scheduler_init.h and tbb/tbb_stddef.h do no longer exist. When I patch CMAKE to use tbb/task.h and oneapi/tbb/version.h it fails in build with missing tbb/atomic.h in include/util/packed_vector.hpp.

zgdump:
I have the same trouble on the latest Majaro Linux. After patching FindTBB.cmake, build fails on including atomic headers

stoecker:
Is there a release planned in next time or should I think about building a patched version?

SiarheiFedartsou:
Yes, we are planning it (see https://github.com/Project-OSRM/osrm-backend/issues/6325), but not sure when it will happen.

Issue #6087, opened 2021-07-20
add Lua version 5.4 & tbb@2021 support


robertsjchen:
Hello, 

recently I tried to download & compile the entire project from my mac, I following those building instructions downloading dependency(boost, lua, tbb, etc) via brew, but the situation is, all dependency packages been downloaded latest by default, for example, as of today(Jul/2021), version of Lua is 5.4.3, version of tbb is 2021.3, cmake will fail with package can not found error. 

* In file "cmake/FindTBB.cmake", the hardcoded heading file "task_scheduler_init.h" been removed from tbb@2021:
```
find_path(TBB_INCLUDE_DIR
    tbb/task_scheduler_init.h
    HINTS ${TBB_INC_SEARCH_DIR} ENV CPATH
)
```
* In file "cmake/FindLua.cmake", the latest version of Lua been supported is 5.3(I added 5.4 locally fixed this issue):
```
function(_lua_set_version_vars)
    set(LUA_VERSIONS5 5.3 5.2 5.1 5.0)
    ...
```

Thanks
robert

danpat:
You can try doing `cmake -DENABLE_MASON=ON ..` when you run `cmake` - this will download pre-built binaries for known-to-work dependencies, rather than using those available on your system or via homebrew.

Of course, a patch to upgrade the dependencies would also be welcome.

mjjbell:
#5981 has been merged, which will fix the Lua 5.4 issue.

robertsjchen:
thank you so much @mjjbell & @danpat !

mjjbell:
For now, to continue building on a Mac, you'll need to install tbb@2020 and set a `TBB_INSTALL_DIR` environment variable to the install location (e.g. `/usr/local/opt/tbb@2020`), so that `FindTBB.cmake` can find it.

PR #6300, opened 2022-08-08
Migrate to modern TBB version

SiarheiFedartsou:
# Issue

closes https://github.com/Project-OSRM/osrm-backend/issues/6297
closes https://github.com/Project-OSRM/osrm-backend/issues/6181
closes https://github.com/Project-OSRM/osrm-backend/issues/6087

## Tasklist

 - [ ] review
 - [ ] adjust for comments

## Requirements / Relations


.github/workflows/osrm-backend.yml
```diff
@@ -155,7 +155,8 @@ jobs:
             CXXCOMPILER: g++-9
             CXXFLAGS: "-m32 -msse2 -mfpmath=sse"
             TARGET_ARCH: i686
-
+            ENABLE_CONAN: ON
```
SiarheiFedartsou, line 158: 32-bit build was complaining about some linker errors related to TBB(even though there is 32-bit binary and I also tried to build it from source), so it is easier to simply use Conan instead which seems to work.

CMakeLists.txt
```diff
@@ -459,22 +458,42 @@ if(ENABLE_CONAN)
...
+      bzip2/${CONAN_BZIP2_VERSION}
+      expat/${CONAN_EXPAT_VERSION}
+      lua/${CONAN_LUA_VERSION}
+      onetbb/${CONAN_TBB_VERSION}
+      BASIC_SETUP 
+      BUILD missing 
+      GENERATORS cmake_find_package
+      KEEP_RPATHS
+      NO_OUTPUT_DIRS
+      ARCH x86
```
SiarheiFedartsou, line 480: I am not CMake pro, so not sure how to reduce copy-paste here(the difference between branches is only this `ARCH x86` line)

cmake/FindTBB.cmake
```diff
@@ -1,286 +1,456 @@
-# Locate Intel Threading Building Blocks include paths and libraries
```
SiarheiFedartsou, line 1: Took it from here https://github.com/Kitware/VTK/blob/master/CMake/FindTBB.cmake

src/contractor/contractor.cpp
```diff
@@ -35,12 +35,7 @@
 #include <vector>
 
 #include <boost/assert.hpp>
-
-#if TBB_VERSION_MAJOR == 2020
```
SiarheiFedartsou, line 39: https://oneapi-src.github.io/oneTBB/main/tbb_userguide/Migration_Guide.html

include/util/packed_vector.hpp
```diff
@@ -89,6 +92,17 @@ inline WordT set_upper_value(WordT word, WordT mask, std::uint8_t offset, T valu
     return (word & ~mask) | ((static_cast<WordT>(value) >> offset) & mask);
 }
 
+inline bool compare_and_swap(uint64_t *ptr, uint64_t old_value, uint64_t new_value)
```
SiarheiFedartsou, line 95: Tbh I am not sure if we need PackedVector to be thread-safe at all - after brief look I couldn't find a place where we modify it concurrently.

CMakeLists.txt
```diff
@@ -496,7 +516,7 @@ if(ENABLE_CONAN)
   set(Boost_UNIT_TEST_FRAMEWORK_LIBRARY "${Boost_unit_test_framework_LIB_TARGETS}")
   
   find_package(BZip2 REQUIRED EXACT ${CONAN_BZIP2_VERSION})
-  find_package(EXPAT REQUIRED EXACT ${CONAN_EXPAT_VERSION})
+  find_package(EXPAT REQUIRED)
```
SiarheiFedartsou, line 519: https://github.com/Project-OSRM/osrm-backend/pull/6299

SiarheiFedartsou:
@mjjbell hi, would be great to have your opinion here

SiarheiFedartsou:
I see that Windows build fails. Most likely because it uses old TBB [stored on wolt.com](https://github.com/Project-OSRM/osrm-backend/blob/e7185b4bcb7eb17ea4f85dc3d087880dbd309646/appveyor-build.bat#L55). What I am thinking about is may be migrate it to use Conan too - should be probably a bit better approach rather than try to find someone from Wolt to update it now... Another option is to simply install TBB to the system just like I did for Linux/Mac.

SiarheiFedartsou:
> I see that Windows build fails. Most likely because it uses old TBB [stored on wolt.com](https://github.com/Project-OSRM/osrm-backend/blob/e7185b4bcb7eb17ea4f85dc3d087880dbd309646/appveyor-build.bat#L55). What I am thinking about is may be migrate it to use Conan too - should be probably a bit better approach rather than try to find someone from Wolt to update it now... Another option is to simply install TBB to the system just like I did for Linux/Mac.

Btw worth considering migration to GitHub actions for Windows CI 🤔

mjjbell:
Agreed, adding a Windows build to Github Actions CI and killing off the old one is probably the direction to take and will make it easier to perform library upgrades.
IIRC the motivation for moving to C++17 was to take advantage of the STL parallel algorithms. Removing TBB completely could solve these problems too, although might require a lot of surgery.

SiarheiFedartsou:
> IIRC the motivation for moving to C++17 was to take advantage of the STL parallel algorithms. Removing TBB completely could solve these problems too, although might require a lot of surgery.

Yep, but it turned out that there are a lot of compilers that still do not support STL parallel algorithms from C++17 in 2022. One example is Apple Clang. I haven't dug deep here though, but lets first migrate TBB in order to unblock local builds without Conan - it seems to be easier problem anyway.

src/storage/storage.cpp
```diff
@@ -290,27 +290,27 @@ int Storage::Run(int max_wait, const std::string &dataset_name, bool only_metric
 
 std::vector<std::pair<bool, boost::filesystem::path>> Storage::GetStaticFiles()
 {
-    constexpr bool REQUIRED = true;
-    constexpr bool OPTIONAL = false;
+    constexpr bool IS_REQUIRED = true;
+    constexpr bool IS_OPTIONAL = false;
```
SiarheiFedartsou, line 294: `windows.h` defines `OPTIONAL` macro 🤦🏻

docker/Dockerfile
```diff
@@ -4,13 +4,23 @@ ARG BUILD_CONCURRENCY
 RUN mkdir -p /src  && mkdir -p /opt
 
 RUN apt-get update && \
-    apt-get -y --no-install-recommends install cmake make git gcc g++ libbz2-dev libxml2-dev \
-    libzip-dev libboost1.74-all-dev lua5.2 liblua5.2-dev libtbb-dev -o APT::Install-Suggests=0 -o APT::Install-Recommends=0
+    apt-get -y --no-install-recommends install ca-certificates cmake make git gcc g++ libbz2-dev libxml2-dev wget \
+    libzip-dev libboost1.74-all-dev lua5.2 liblua5.2-dev -o APT::Install-Suggests=0 -o APT::Install-Recommends=0
+
+RUN NPROC=${BUILD_CONCURRENCY:-$(nproc)} && \
```
SiarheiFedartsou, line 10: We could download pre-built binaries, but TBB does not provide arm64 binaries at the moment

SiarheiFedartsou:
@mjjbell I think I am done with it

include/util/packed_vector.hpp
mjjbell: According to #4199 it can happen during edge updates, probably when running `osrm-customize`

mjjbell (approved):
LGTM. Thanks for working though all the required changes :+1:

include/util/packed_vector.hpp
SiarheiFedartsou: As I can see set_value is used on push_back to - curious if we are safe on reallocations there 🤔

Issue #6302, opened 2022-08-09
OSRM build fails on GCC 10.2.1


Zverik:
Building the OSRM in Debian Bullseye (the latest stable one) fails with the "use of deleted function" error. You can see it with the Dockerfile in this repo. Same with Ubuntu 22.04. But OSRM builds with Ubuntu 20.04 — although it uses much more memory for processing data for some reason.

Bisected it to #6279.

```
#9 78.05 /src/third_party/libosmium/include/osmium/osm/entity.hpp: In instantiation of 'TSubitem& osmium::detail::subitem_of_type(TIter, const TIter&) [with TSubitem = const osmium::TagList; TIter = osmium::memory::CollectionIterator<const osmium::memory::Item>]':
#9 78.05 /src/third_party/libosmium/include/osmium/osm/object.hpp:333:83:   required from here
#9 78.05 /src/third_party/libosmium/include/osmium/osm/entity.hpp:54:29: error: use of deleted function 'osmium::memory::Collection<osmium::Tag, osmium::item_type::tag_list>::Collection(osmium::memory::Collection<osmium::Tag, osmium::item_type::tag_list>&&)'
#9 78.05    54 |             static TSubitem subitem{};
#9 78.05       |                             ^~~~~~~
#9 78.05 In file included from /src/third_party/libosmium/include/osmium/osm/relation.hpp:36,
#9 78.05                  from /src/include/extractor/extraction_relation.hpp:6,
#9 78.05                  from /src/src/extractor/extractor.cpp:8:
#9 78.05 /src/third_party/libosmium/include/osmium/memory/collection.hpp:117:15: note: 'osmium::memory::Collection<osmium::Tag, osmium::item_type::tag_list>::Collection(osmium::memory::Collection<osmium::Tag, osmium::item_type::tag_list>&&)' is implicitly deleted because the default definition would be ill-formed:
#9 78.05   117 |         class Collection : public Item {
#9 78.05       |               ^~~~~~~~~~
#9 78.05 /src/third_party/libosmium/include/osmium/memory/collection.hpp:117:15: error: use of deleted function 'osmium::memory::Item::Item(osmium::memory::Item&&)'
#9 78.05 In file included from /src/third_party/libosmium/include/osmium/memory/buffer.hpp:36,
#9 78.05                  from /src/include/extractor/scripting_environment.hpp:9,
#9 78.05                  from /src/include/extractor/graph_compressor.hpp:4,
[...]

PR #6304, opened 2022-08-09
Fix performance issues after migration to sol2 3.3.0

SiarheiFedartsou:
# Issue

People are reporting performance issues after merge of https://github.com/Project-OSRM/osrm-backend/pull/6279, I reproduced it and it turned out that now we are copying the whole `relations` container on each invocation of node_function 🤦🏻 

Some discussions in:

https://github.com/Project-OSRM/osrm-backend/issues/6302
https://github.com/Project-OSRM/osrm-backend/issues/6297
## Tasklist

 - [ ] review
 - [ ] adjust for comments

## Requirements / Relations


KaSakee (approved):
It makes sense to use reference_wrapper, Nice job.

mjjbell:
Thanks, I missed this in the original PR.

To confirm my understanding:
- the `sol2` upgrade has changed the behaviour of passing const references to user-defined types. It will now copy the value (as indicated by the Node [copy constructor error](https://github.com/Project-OSRM/osrm-backend/pull/6279#discussion_r917304115))
- `ExtractionRelationContainer` had a copy constructor, so this is the performance regression.
- Wrapping non-const references is not strictly necessary as the behaviour has not changed, but it makes the arguments consistent / future-proofed.

Is this right?

SiarheiFedartsou:
> the sol2 upgrade has changed the behaviour of passing const references to user-defined types

Tbh I haven't seen it in sol2's changelog, but I assume so. It will now copy the value

>(as indicated by the Node https://github.com/Project-OSRM/osrm-backend/pull/6279#discussion_r917304115)

Yes, and it is my fault that I didn't pay enough attention to it 🤦🏻 

> ExtractionRelationContainer had a copy constructor, so this is the performance regression.

Yes.

> Wrapping non-const references is not strictly necessary as the behaviour has not changed, but it makes the arguments consistent / future-proofed.

In other words you are curious here why didn't we have problems for the cases when something was supposed to be changed by Lua script? I.e. if it was copied, i.e. we would change the copy in the script then behaviour was expected to be changed in this case? Tbh I didn't think about that. Probably worth looking a bit deeper into it to be 100% confident. 
There is [a paragraph about this](https://sol2.readthedocs.io/en/latest/functions.html#function-argument-handling) in sol2 docs. They say it copies C++ types by default 🤔 

Screenshot 2022-08-15 at 22 34 18

I'll probably be able to dig into it in the nearest days and have better answers. WDYT?

mjjbell:
> > Wrapping non-const references is not strictly necessary as the behaviour has not changed, but it makes the arguments consistent / future-proofed.
> 
> In other words you are curious here why didn't we have problems for the cases when something was supposed to be changed by Lua script? I.e. if it was copied, then we would change the copy in the script -> behaviour was expected to be changed in this case? Tbh I didn't think about that. Probably worth looking a bit deeper into it to be 100% confident. There is [a paragraph about this](https://sol2.readthedocs.io/en/latest/functions.html#function-argument-handling) in sol2 docs. They say it copies C++ types by default thinking
> 
> Screenshot 2022-08-15 at 22 34 18
> 
> I'll probably be able to dig into it in the nearest days and have better answers. WDYT?

Yes, although it has the exact [same wording](https://sol2.readthedocs.io/en/v2.20.6/functions.html) in the old docs, but we know something has changed since then.
I think it's fine as all of the tests would have failed. Judging by this relatively recent [comment](https://github.com/ThePhD/sol2/issues/1070#issuecomment-747869536), what we are seeing is now aligned with the intended behaviour.

mjjbell:
And the `std::ref` wrapping ensures it's not a problem in any case :+1:

mjjbell (approved):

PR #6312, opened 2022-08-16
Migrate Windows CI to GitHub Actions.

SiarheiFedartsou:
# Issue

It is quite difficult to maintain current Windows CI, because it uses pre-compiled dependencies stored on wolt.com server and there is low chance to find someone who is able to update it.

Needed for https://github.com/Project-OSRM/osrm-backend/pull/6300

## Tasklist

 - [ ] review
 - [ ] adjust for comments

## Requirements / Relations


scripts/ci/windows-build.bat
```diff
@@ -0,0 +1,99 @@
+@ECHO OFF
```
SiarheiFedartsou, line 1: It is modified old `appveyor-build.bat`

CMakeLists.txt
```diff
@@ -474,6 +484,7 @@ if(ENABLE_CONAN)
     GENERATORS cmake_find_package
     KEEP_RPATHS
     NO_OUTPUT_DIRS
+    OPTIONS tbb:shared=${TBB_SHARED}
```
SiarheiFedartsou, line 487: Tbh I spend several days trying to understand why `osrm-extract.exe` silently finishes with `-1073741515` exit code(Google says that it is something about DLL loading), then I found that it happens on the first access of any TBB symbol. I tried to statically compile it and it worked. I believe it is possible to make it working with dynamic linkage, but I have no Windows machine and had to debug all these things directly on CI, so debugging is really limited for me.

CMakeLists.txt
```diff
@@ -446,6 +445,10 @@ add_subdirectory(${FLATBUFFERS_SRC_DIR}
         ${CMAKE_CURRENT_BINARY_DIR}/flatbuffers-build
         EXCLUDE_FROM_ALL)
 
+if (MSVC)
```
SiarheiFedartsou, line 448: https://stackoverflow.com/questions/70898030/boost-link-error-using-conan-find-package

SiarheiFedartsou:
@mjjbell please take a look when you have a chance

SiarheiFedartsou:
Also probably someone with corresponding access rights should disable integration with AppVeyor:
Screenshot 2022-08-16 at 23 45 25

CMakeLists.txt
mjjbell: :+1: 
Can you add a comment to that effect in the file?

CMakeLists.txt
mjjbell: Let's add this to the file too, for future reference.

mjjbell:
I don't have access to the status checks. However I have access to the appveyor project, so I've removed the integration from that side.

CMakeLists.txt
SiarheiFedartsou: Added.

mjjbell:
> I don't have access to the status checks. However I have access to the appveyor project, so I've removed the integration from that side.

I think we still need someone to disable it in the repo. @danpat @TheMarex would you be able to disable the Appveyor check in the repo settings?

.github/workflows/osrm-backend.yml
```diff
@@ -20,6 +20,27 @@ env:
   ENABLE_NODE_BINDINGS: "ON"
 
 jobs:
+  windows:
```
mjjbell, line 23: I forgot about this with the Docker job.

[ci-complete](https://github.com/Project-OSRM/osrm-backend/pull/6312/files#diff-4f680fa9a2e9b55e339a5fc727a44e968318e00f54beb55da0e5c5b24bbd2f9dR642) is the meta-job that PRs will block merging on. We should add `windows` and `docker` to the list of `needs`, in the unlikely case that all the Linux/Mac jobs succeed before those two complete/fail.

SiarheiFedartsou: Okay, will do. Btw while we are waiting for someone to disable Appveyor check, may be just return that job and make it no-op(i.e. always green?). It would unblock this PR from being merged and would allow to work on other PRs which depend on it.

mjjbell: Good idea :+1:

mjjbell (approved):
LGTM :rocket: 

One step closer to being able to publish Windows releases too.

DennisOSRM:
Removed the appveyor web hook

PR #6314, opened 2022-08-18
Remove references to project-osrm.wolt.com

jpaalasm:
At Wolt, we are cleaning up subdomains under wolt.com, and project-osrm.wolt.com will go offline in the near future.

There are references to project-osrm.wolt.com also at https://github.com/Project-OSRM/osrm-backend/blob/master/appveyor-build.bat so that's another thing to fix.

### DIFF
--- a/docs/windows-deps.md
+++ b/docs/windows-deps.md
@@ -12,20 +12,14 @@ Prepare directories for dependencies, build and target file location.Target dire
 
 ### Bzip2
 
-1. Download either from Wolt OSRM mirror or original distribution and unpack.
-    * https://project-osrm.wolt.com/deps/bzip2-1.0.8.tar.gz
-    * https://sourceware.org/pub/bzip2/bzip2-1.0.8.tar.gz
-
+1. Download from https://sourceware.org/pub/bzip2/bzip2-1.0.8.tar.gz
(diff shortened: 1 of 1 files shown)


SiarheiFedartsou:
> There are references to project-osrm.wolt.com also at https://github.com/Project-OSRM/osrm-backend/blob/master/appveyor-build.bat

Btw there is PR which migrates Windows CI to Conan, so there will be no no need for this https://github.com/Project-OSRM/osrm-backend/pull/6312

jpaalasm:
Thanks for the info! Does this mean we can proceed decommissioning project-osrm.wolt.com without causing harm to builds?

jpaalasm:
And I suppose this PR could also be merged?

SiarheiFedartsou:
> Thanks for the info! Does this mean we can proceed decommissioning project-osrm.wolt.com without causing harm to builds?

I think so, we already merged that PR and we do not depend on wolt.com(at least AFAIR). Thanks for the care about community :)

mjjbell:
Thanks. Yes, the wiki will need to be updated again once we've completed the upgrade of Windows builds. 
But for now we can remove the references to unblock your decommissioning :+1: .

mjjbell (approved):

PR #6333, opened 2022-08-25
Fix TBB in case of Conan + NodeJS build

SiarheiFedartsou:
# Issue

Forgot to fix it during migration to Conan.

## Tasklist

 - [ ] review
 - [ ] adjust for comments

## Requirements / Relations

### DIFF
--- a/src/nodejs/CMakeLists.txt
+++ b/src/nodejs/CMakeLists.txt
@@ -34,7 +34,6 @@ endforeach(binary)
 # For mason-enabled builds we copy over tbb's shared objects for packaging.
 # TODO: consider using statically linked tbb library (for node_osrm only!)
 if (ENABLE_CONAN)
-  message(FATAL_ERROR "DFVRSDFGDFG ${CONAN_LIB_DIRS_ONETBB}")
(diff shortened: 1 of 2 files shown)


src/nodejs/CMakeLists.txt
```diff
@@ -33,8 +33,8 @@ endforeach(binary)
 
 # For mason-enabled builds we copy over tbb's shared objects for packaging.
 # TODO: consider using statically linked tbb library (for node_osrm only!)
```
SiarheiFedartsou, line 35: Btw I think it can be possible to try(we already doing that for Windows builds). @mjjbell do you see some downsides of it?

mjjbell: I don't have enough context on how the Node packaging works, but it doesn't look like we'd be losing much as we're not actually getting the benefit of a shared lib. If anything, statically linking might reduce the package size.

SiarheiFedartsou: Okay, will try to check how it will work :+1:

mjjbell (approved):

Issue #5936, opened 2021-01-14
Does nodejs bindings support flatbuffer 


xcTorres:

SiarheiFedartsou:
https://github.com/Project-OSRM/osrm-backend/pull/6338

PR #6338, opened 2022-08-27
Add Flatbuffers support to NodeJS bindings

SiarheiFedartsou:
# Issue

closes https://github.com/Project-OSRM/osrm-backend/issues/5936

## Tasklist

 - [ ] review
 - [ ] adjust for comments

## Requirements / Relations

### DIFF
--- a/docs/nodejs/api.md
+++ b/docs/nodejs/api.md
@@ -63,8 +63,8 @@ Returns the fastest route between two or more coordinates while visiting the way
     -   `options.approaches` **[Array](https://developer.mozilla.org/docs/Web/JavaScript/Reference/Global_Objects/Array)?** Keep waypoints on curb side. Can be `null` (unrestricted, default) or `curb`.
                          `null`/`true`/`false`
     -   `options.waypoints` **[Array](https://developer.mozilla.org/docs/Web/JavaScript/Reference/Global_Objects/Array)?** Indices to coordinates to treat as waypoints. If not supplied, all coordinates are waypoints.  Must include first and last coordinate index.
-    -   `options.snapping` **[String](https://developer.mozilla.org/docs/Web/JavaScript/Reference/Global_Objects/String)?** Which edges can be snapped to, either `default`, or `any`.  `default` only snaps to edges marked by the profile as `is_startpoint`, `any` will allow snapping to any edge in the routing graph.
     -   `options.format` **[String](https://developer.mozilla.org/docs/Web/JavaScript/Reference/Global_Objects/String)?** Which output format to use, either `json`, or [`flatbuffers`](https://github.com/Project-OSRM/osrm-backend/tree/master/include/engine/api/flatbuffers).
+    -   `options.snapping` **[String](https://developer.mozilla.org/docs/Web/JavaScript/Reference/Global_Objects/String)?** Which edges can be snapped to, either `default`, or `any`.  `default` only snaps to edges marked by the profile as `is_startpoint`, `any` will allow snapping to any edge in the routing graph.
 -   `callback` **[Function](https://developer.mozilla.org/docs/Web/JavaScript/Reference/Statements/function)** 
 
 **Examples**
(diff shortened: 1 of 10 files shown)


test/nodejs/route.js
```diff
@@ -5,6 +5,43 @@ var monaco_mld_path = require('./constants').mld_data_path;
 var monaco_corech_path = require('./constants').corech_data_path;
 var three_test_coordinates = require('./constants').three_test_coordinates;
 var two_test_coordinates = require('./constants').two_test_coordinates;
+const flatbuffers = require('../../features/support/flatbuffers').flatbuffers;
```
SiarheiFedartsou, line 8: Here I had to import flatbuffers related things from `features` folder, IMO it is a bit ugly, but not a big problem - as an alternative we probably can re-arrange folders somehow or add symlink to that folder in `test/nodejs`.

include/nodejs/node_osrm_support.hpp
```diff
@@ -96,6 +97,18 @@ inline void ParseResult(const osrm::Status &result_status, osrm::json::Object &r
 }
 
 inline void ParseResult(const osrm::Status & /*result_status*/, const std::string & /*unused*/) {}
+inline void ParseResult(const osrm::Status &result_status,
+                        const flatbuffers::FlatBufferBuilder &fbs_builder)
+{
+    auto fbs_result = osrm::engine::api::fbresult::GetFBResult(fbs_builder.GetBufferPointer());
```
SiarheiFedartsou, line 103: Done using [ParseResult for JSON](https://github.com/Project-OSRM/osrm-backend/blob/d74e7b66bdc5e9101ac3f1f74b969f5b6c84756a/include/nodejs/node_osrm_support.hpp#L78) as a reference. We also erase `message` and `code` there, but it is not so easy for flatbuffers(it is kind of readonly) and tbh I don't get why are we doing that at all.

test/nodejs/match.js
```diff
@@ -4,6 +4,24 @@ var data_path = require('./constants').data_path;
 var mld_data_path = require('./constants').mld_data_path;
 var three_test_coordinates = require('./constants').three_test_coordinates;
 var two_test_coordinates = require('./constants').two_test_coordinates;
+const flatbuffers = require('../../features/support/flatbuffers').flatbuffers;
+const FBResult = require('../../features/support/fbresult_generated').osrm.engine.api.fbresult.FBResult;
+
+
+test('match: match in Monaco with flatbuffers format', function(assert) {
```
SiarheiFedartsou, line 11: Added just a couple of smoke tests for each plugin: I assume that it is enough for the beginning.

src/nodejs/node_osrm.cpp
```diff
@@ -619,7 +645,7 @@ NAN_METHOD(Engine::trip) //
  *     [13.374481201171875, 52.506191342034576]
  *   ]
  * };
- * osrm.route(options, { format: "json_buffer" }, function(err, response) {
+ * osrm.route(options, { format: "buffer" }, function(err, response) {
```
SiarheiFedartsou, line 648: Renamed it to just `buffer` having `json_buffer` as deprecated alias.

docs/nodejs/api.md
```diff
@@ -332,12 +334,15 @@ specific behaviours.
...
     -   `plugin_config.format` **[String](https://developer.mozilla.org/docs/Web/JavaScript/Reference/Global_Objects/String)?** The format of the result object to various API calls.
-                                               Valid options are `object` (default), which returns a
-        standard Javascript object, as described above, and `json_buffer`, which will return a NodeJS
-        **[Buffer](https://nodejs.org/api/buffer.html)** object, containing a JSON string. The latter has
-        the advantage that it can be immediately serialized to disk/sent over the network, and the
-        generation of the string is performed outside the main NodeJS event loop.  This option is ignored
-        by the `tile` plugin.
+                                               Valid options are `object` (default if `options.format` is
+        `json`), which returns a standard Javascript object, as described above, and `buffer`(default if
+        `options.format` is `flatbuffers`), which will return a NodeJS
```
mjjbell, line 339: I think these need to be reworded to clarify that flatbuffers can only be output to `buffer`. It's not just the default case, it's the only case.

SiarheiFedartsou: Hm, agree, but not sure how to better express that. And also we already have

```
Also note that `options.format` set to `flatbuffers`
        cannot be used with `plugin_config.format` set to `object`
```

A bit below these lines.

Do you have good ideas?

src/nodejs/node_osrm.cpp
mjjbell: Unlikely, but I guess it's possible someone could ask for the `json_buffer` plugin format and the Flatbuffer response format. Maybe it's worth keeping the three plugin options separate for clarity.

- Object -> json object
- json_buffer, json string
- flatbuffer...buffer, flatbuffer bytes

SiarheiFedartsou: Well, at the moment all our Node bindings accept options and _plugin_ options. This `format` which currently exists in Node bindings and can be either `object` or `json_buffer` is the only citizen of these plugin options. At the same time options contain everything which we usually accept through our HTTP API and now we have new `format` parameter there which can be either `json` or `flatbuffers` and it 100% mirrors what we have in HTTP API(we have `format` there too). So my idea here is that: `format` from plugin options now controls the way how we receive data on Node side(in JSON or in byte buffer) and `format` from options controls representation of data itself(imagine tomorrow we can start supporting `yaml` as response representation and return it as both object and buffer)

That's the reason why I decided to do it exactly how I did it(btw initially I did almost like you proposed: `object/json_buffer/flat_buffer`, but then realized that plugin options and options are a bit different things and if we have this new `format` as separate field in options it will be a bit more consistent).

WDYT? In general I don't have any strong opinion here, IMO any solution here is good enough, but just sharing motivation behind current implementation.

SiarheiFedartsou: > imagine tomorrow we can start supporting yaml as response representation and return it as both object and buffer

I doubt that something except JSON may need `object` :)

SiarheiFedartsou: Also may be the fact that both fields are called `format` is a bit confusing 🤔

SiarheiFedartsou: Another option I see is to deprecate plugin options at all and move buffer/object to `format` in usual options.

mjjbell: Yeah I don't think the Node API was considered when flat buffers were added. I guess right now it would just return JSON.

I think the issue I have is aliasing `json_buffer` to `buffer`. If we can limit `json_buffer` to JSON responses, error when they try to set flatbuffers, and mark `json_buffer` as deprecated, then I'm ok with this.

SiarheiFedartsou: In other words you just propose to add condition to throw error in the case if `json_buffer` is used along with `flatbuffers`? Done.

> and mark json_buffer as deprecated

Already mention that in docstring, not sure if it is possible to do that somehow better.

WDYT?

mjjbell: Yep, sounds good :+1:

mjjbell (approved):

PR #6339, opened 2022-08-27
Fix snapping target locations to ways used in turn restrictions

### LINKED ISSUE #6069: Weird driving directions / path generated in certain areas
I try to use OSRM to generate driving directions but it gives a weird result. I have checked the OSM data and I didn't see anything special there.

Can someone please help me find out whether this is an OSRM bug, or it is just a map issue that I can solve it by updating the OSM data.

I got the problem using version 5.25 built from source. But same problem can be directly reproduced on OSM website too. Here is an example:

osm1

I don't know why it does not just go straight and make a left turn at the end. The expected driving path can be generated using graphhopper though:

osm2

Thanks a lot in advance!

mjjbell:
# Issue
before_after_via_end_1

[link](https://map.project-osrm.org/?z=18&center=50.778949%2C6.093239&loc=50.777613%2C6.094687&loc=50.778243%2C6.094258&hl=en&alt=0&srv=0)

Currently there is an edge-case in the turn restriction implementation, such that routes can not be found if the target input location snaps to a way used in a (multi) via restriction.

With the addition of snapping input locations to multiple ways, we can now also snap to the "duplicate" edges created for the restriction graph, thereby fixing the problem.
This is achieved by adding the duplicate restriction edges to the geospatial search RTree.

This does open up the possibility of multiple paths representing exactly the same route - one using the original edge as a source, the other using the duplicate restriction graph edge as source.
This is fine, as both edges are represented by the same geometry, so will generate the same result.

## Tasklist

 - [ ] review
 - [ ] adjust for comments

## Requirements / Relations

Fixes #6069
Fixes https://github.com/Project-OSRM/osrm-backend/pull/5907#issuecomment-797430372

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -50,6 +50,7 @@
       - ADDED: Add support for non-round-trips with a single fixed endpoint. [#6050](https://github.com/Project-OSRM/osrm-backend/pull/6050)
       - FIXED: Improvements to maneuver override processing [#6125](https://github.com/Project-OSRM/osrm-backend/pull/6125)
       - ADDED: Support snapping to multiple ways at an input location. [#5953](https://github.com/Project-OSRM/osrm-backend/pull/5953)
+      - FIXED: Fix snapping target locations to ways used in turn restrictions. [#6339](https://github.com/Project-OSRM/osrm-backend/pull/6339)
 
 # 5.26.0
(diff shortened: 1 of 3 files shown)


SiarheiFedartsou (approved):
:rocket:

Issue #5970, opened 2021-02-23
flatbuffers nodes subsection of annotations have inappropriate format


nnseva:
The `nodes` section of `flatbuffers` output data is declared as an array of `uint` having 32 bits, while the node identity of the OSM is now wider.

The proposed type of `nodes` section is an array of `ulong` having 64 bits.

Issue #5716, opened 2020-04-16
JSON Number(double) Insufficient for Long NodeID


wangyoucao577:
We use `annotations=true` to output `nodes` in route response, then we found the output `nodes` may invalid sometimes.     
After investigate, it's caused by the JSON util only has one number type `Number`, which defined as `double`:     
https://github.com/Project-OSRM/osrm-backend/blob/15f0ca8ddaa35c5b4d93c25afa72e81e1fb40c3e/include/util/json_container.hpp#L89-L94
It means whatever the internal number type is, it will be converted to `double` before output. For some long `std::uint64_t`, the value will be cut in this conversion.       

In our test, the nodeID `9280805980001101` will be cut to `9280805980001100`, which results invalid nodeID.     
Here's some test code:      
```c++
// Example program
#include <iostream>
#include <string>
#include <cstdint>
#include <iomanip>

int main()
{
  std::uint64_t id = 9280805980001101;
  double d_id = static_cast<double>(id);
  std::uint64_t convert_back_id = static_cast<std::uint64_t>(d_id);
  
  std::cout << id << std::endl;
  std::cout << std::fixed << std::setprecision(6) << d_id << std::endl;
  std::cout << convert_back_id << std::endl;
}

// output: 
9280805980001101
9280805980001100.000000
9280805980001100
``` 
Full story please refer to https://github.com/Telenav/osrm-backend/issues/276

Any suggestion? Is it possible to have more Number types in the JSON util?  
Many thanks in advance!

awgrover:
The JSON spec does not appear to limit the range of a number at all(!) https://www.json.org/json-en.html. JSON Schema seems to, I think, as 32 bit signed: -2,147,483,648 ... 2,147,483,647 . Google notes a work around of

> a 64-bit integer must be represented as a string in JSON requests/responses. The type property will be set to "string", but the format property will be set to "int64" to indicate that it is a 64-bit integer

As a workaround in this case, is it possible for you to use a string-id instead (i.e. trivially convert to-string for osrm, then convert to-int on return)?

wangyoucao577:
@awgrover  Actually I already returned `string` id as a workaround(see https://github.com/Telenav/osrm-backend/issues/276#issuecomment-615110149 and https://github.com/Telenav/osrm-backend/pull/285).      
However, refer to https://stackoverflow.com/questions/13502398/json-integers-limit-on-size, the number seems unlimited by JSON spec. I think return correctly `uint64` still will be better.

Issue #5644, opened 2020-01-09
Annotations give nodes that don't exist in the input osm file


flacombe:
Dear maintainers,

As you may find in the attached .osm file, all nodes have really big ids > 9999999999.
[loire.osm.zip](https://github.com/Project-OSRM/osrm-backend/files/4041205/loire.osm.zip)

When I process it with osrm-contract and make it available with osrm-routed 5.19.0, a query like

https://.../route/v1/driving/4.484138488769532,45.541510228130996;4.4870996475219735,45.54508696616638?overview=false&alternatives=true&steps=false&annotations=nodes

gives me the following annotations
```
{"code":"Ok","waypoints":[{"hint":"sQAAgNwAAIDlAQAAHQAAAFICAAAVBgAADxBCQjBdOEBL_m1C_bIqQzlEAAANBAAAilMAAB3bAAABAAAAhWpEACHptgIqbEQAhui2AgIAvwEuiaZu","location":[4.483717,45.541665],"name":"126623790"},{"hint":"SQEAgGcBAIA4AQAAmQAAAAAAAADuAAAAZvL5QSLDdEEAAAAA3yk6QtkrAAB6FQAAAAAAAHshAAABAAAArHdEAJH2tgK8d0QAf_a2AgAArwMuiaZu","location":[4.487084,45.545105],"name":"349379862"}],"routes":[{"legs":[{"steps":[],"weight":537.4,"distance":537.4,"annotation":{"nodes":[1604625344,2455888032,6010937664,7256645296,6208998656,4508343296,3635029600,939062000,6278481792,3977370144,5618166736,1851168448,3461536656,5512249904,7342764128,4935365664,1502573568,2345701904,7905357696,1214328448,4647535552,7875718336,4701186976,721583008,7450880000,7671176016,3106836464,7839090624]},"summary":"","duration":19344.7}],"weight_name":"surface","weight":537.4,"distance":537.4,"duration":19344.7}]}
```

Any of returned node don't exists in the input file.
Why?

What should I do to get a consistent answer with existing nodes please?

danpat:
Looks like we've got an overflow bug storing node IDs.

In your file, one of the node ID values is 99742940620696000 - this is *waay* bigger than 2^32, but is still less than 2^64.  I suspect we've got a signed-vs-unsigned integer problem in here somewhere.

The quick fix on your side would be to make sure node ID values are less than 2^32 (4294967296), which should avoid triggering the overflow.

[...]

PR #6341, opened 2022-08-27
Introduce osm_nodes annotations for 64-bit OSM Node Ids

SiarheiFedartsou:
# Issue

potentially closes https://github.com/Project-OSRM/osrm-backend/issues/5970 and https://github.com/Project-OSRM/osrm-backend/issues/5716

The question here is how to preserve backward compatibility, that's why I want to discuss this before doing final "polishing".

What I am doing at the moment I add additional field `osm_nodes` which in some cases may be returned along with old `nodes` field which is kind of ugly and wasteful(but we cannot just remove this old `nodes` without breaking backward compatibility). Another option would be to add parameter(e.g. `use_64bit_node_ids=true`) and start returning node ids as strings in this case. 

Also worth noting that JSON itself seems to support 64-bit numbers and we could just fix our JSON serialization to not convert numbers to double before serialization, but JS itself safely supports only integers up to 53-bit(https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/MAX_SAFE_INTEGER) and 

## Tasklist

 - [ ] review
 - [ ] adjust for comments

## Requirements / Relations

### DIFF
--- a/include/engine/api/flatbuffers/route.fbs
+++ b/include/engine/api/flatbuffers/route.fbs
@@ -10,6 +10,8 @@ table Annotation {
     duration: [uint];
     datasources: [uint];
     nodes: [uint];
+    osm_nodes: [ulong];
+    
     weight: [uint];
     speed: [float];
     metadata: Metadata;
--- a/include/engine/api/nearest_api.hpp
+++ b/include/engine/api/nearest_api.hpp
@@ -7,6 +7,7 @@
 
 #include "engine/api/json_factory.hpp"
 #include "engine/phantom_node.hpp"
+#include "util/json_container.hpp"
 
 #include <boost/assert.hpp>
 
@@ -109,6 +110,11 @@ class NearestAPI final : public BaseAPI
                                nodes.values.push_back(node_values.first);
                                nodes.values.push_back(node_values.second);
                                waypoint.values["nodes"] = std::move(nodes);
+
+                               util::json::Array osm_nodes;
+                               osm_nodes.values.emplace_back(std::to_string(node_values.first));
+                               osm_nodes.values.emplace_back(std::to_string(node_values.second));
+                               waypoint.values["osm_nodes"] = std::move(osm_nodes);
 
                                return waypoint;
                            });
--- a/include/engine/api/route_api.hpp
+++ b/include/engine/api/route_api.hpp
@@ -852,17 +852,6 @@ class RouteAPI : public BaseAPI
                         annotation.values["osmnodes"] = osm_nodes;
                     }
                 }
-                if (requested_annotations & RouteParameters::AnnotationsType::OSMNodes)
-                {
-                    util::json::Array nodes;
-                    nodes.values.reserve(leg_geometry.node_ids.size());
-                    for (const auto node_id : leg_geometry.node_ids)
-                    {
-                        nodes.values.push_back(std::to_string(
-                            static_cast<std::uint64_t>(facade.GetOSMNodeIDOfNode(node_id))));
-                    }
-                    annotation.values["nodes"] = std::move(nodes);
-                }
                 // Add any supporting metadata, if needed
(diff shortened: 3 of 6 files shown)


include/engine/api/nearest_api.hpp
```diff
@@ -110,6 +111,11 @@ class NearestAPI final : public BaseAPI
                                nodes.values.push_back(node_values.second);
                                waypoint.values["nodes"] = std::move(nodes);
 
+                               util::json::Array osm_nodes;
+                               osm_nodes.values.emplace_back(std::to_string(node_values.first));
+                               osm_nodes.values.emplace_back(std::to_string(node_values.second));
+                               waypoint.values["osm_nodes"] = std::move(osm_nodes);
```
SiarheiFedartsou, line 117: In the nearest API we unconditionally add this new `osm_nodes` array, i.e. always have kind of duplication.

include/engine/api/route_api.hpp
```diff
@@ -822,16 +823,34 @@ class RouteAPI : public BaseAPI
...
                 {
                     util::json::Array nodes;
+                    util::json::Array osm_nodes;
+
                     nodes.values.reserve(leg_geometry.node_ids.size());
                     for (const auto node_id : leg_geometry.node_ids)
                     {
-                        nodes.values.push_back(
-                            static_cast<std::uint64_t>(facade.GetOSMNodeIDOfNode(node_id)));
+                        const auto osm_node_id =
```
SiarheiFedartsou, line 835: In case of annotations we add this `osm_nodes` if it was explicitly requested(`annotations=osm_nodes,...`) or in case of `annotations=true`

include/engine/api/flatbuffers/route.fbs
```diff
@@ -10,6 +10,8 @@ table Annotation {
     duration: [uint];
     datasources: [uint];
     nodes: [uint];
+    osm_nodes: [ulong];
```
SiarheiFedartsou, line 13: flatbuffers implementation is not ready yet, but general idea will be the same

SiarheiFedartsou:
@mjjbell would be great if you will take a look and share your opinion

SiarheiFedartsou:
On another hand as per [this site](http://textual.ru/64/) the largest node id at the moment is only >2**33, i.e. we seems to be okay with 2**53 limitation of JS(that's the probably main reason why we would need to migrate to string from number).

SiarheiFedartsou:
Another option I see is to explicitly enable/disable this via CLI parameter/config(and have it disabled by default).

SiarheiFedartsou:
OSM Node IDs are 34-bit internally https://github.com/Project-OSRM/osrm-backend/blob/bfb74c2dad0d7c653919271b3329b70c9863e756/include/extractor/packed_osm_ids.hpp#L14
Related issue https://github.com/Project-OSRM/osrm-backend/issues/5644

mjjbell:
Let me summarise to see if I understand.

- OSM Node IDs are currently < 2^34

- JSON Numbers have precision up to 2^53, so our JSON API will be fine for a while / forever.

- Our Flatbuffers API is using `uint` for nodes and is therefore already overflowing. A breaking change is required to fix.

- Some people hack OSM IDs that our larger than JSON number precision. They would like to be able to return node IDs as strings.

Is this right?

SiarheiFedartsou:
> Let me summarise to see if I understand.
> 
> * OSM Node IDs are currently < 2^34
> * JSON Numbers have precision up to 2^53, so our JSON API will be fine for a while / forever.
> * Our Flatbuffers API is using `uint` for nodes and is therefore already overflowing. A breaking change is required to fix.
> * Some people hack OSM IDs that are larger than JSON number precision can support. They would like to be able to return node IDs as strings.
> 
> Is this right?

Exactly.

PR #6344, opened 2022-08-28
Optimize RestrictionParser performance

SiarheiFedartsou:
# Issue

closes https://github.com/Project-OSRM/osrm-backend/issues/4102

## Tasklist

 - [ ] review
 - [ ] adjust for comments

## Requirements / Relations

### DIFF
--- a/include/extractor/restriction_parser.hpp
+++ b/include/extractor/restriction_parser.hpp
@@ -5,7 +5,7 @@
 
 #include <boost/optional/optional.hpp>
 
-#include <osmium/tags/filter.hpp>
+#include <osmium/tags/tags_filter.hpp>
 #include <set>
 #include <string>
 #include <vector>
@@ -54,7 +54,7 @@ class RestrictionParser
     bool use_turn_restrictions;
     bool parse_conditionals;
     std::set<std::string> restrictions;
-    osmium::tags::KeyFilter filter;
+    osmium::TagsFilter filter;
 };
 } // namespace extractor
 } // namespace osrm
--- a/include/extractor/restriction_parser.hpp
+++ b/include/extractor/restriction_parser.hpp
@@ -6,7 +6,9 @@
 #include <boost/optional/optional.hpp>
 
 #include <string>
+#include <set>
 #include <vector>
+#include <osmium/tags/filter.hpp>
 
 namespace osmium
 {
@@ -43,7 +45,7 @@ class RestrictionParser
   public:
     RestrictionParser(bool use_turn_restrictions,
                       bool parse_conditionals,
-                      std::vector<std::string> &restrictions);
+                      const std::vector<std::string> &restrictions);
     std::vector<InputTurnRestriction> TryParse(const osmium::Relation &relation) const;
 
   private:
@@ -51,7 +53,8 @@ class RestrictionParser
 
     bool use_turn_restrictions;
     bool parse_conditionals;
-    std::vector<std::string> restrictions;
+    std::set<std::string> restrictions;
+    osmium::tags::KeyFilter filter;
 };
 } // namespace extractor
 } // namespace osrm
--- a/include/extractor/restriction_parser.hpp
+++ b/include/extractor/restriction_parser.hpp
@@ -5,10 +5,10 @@
 
 #include <boost/optional/optional.hpp>
 
-#include <string>
-#include <set>
-#include <vector>
 #include <osmium/tags/filter.hpp>
+#include <set>
+#include <string>
+#include <vector>
 
 namespace osmium
 {
--- a/src/extractor/restriction_parser.cpp
+++ b/src/extractor/restriction_parser.cpp
@@ -109,7 +109,8 @@ RestrictionParser::TryParse(const osmium::Relation &relation) const
         {
             is_only_restriction = true;
         }
-        else if (boost::algorithm::starts_with(value, "no_") && !boost::algorithm::ends_with(value, "_on_red"))
+        else if (boost::algorithm::starts_with(value, "no_") &&
+                 !boost::algorithm::ends_with(value, "_on_red"))
         {
             is_only_restriction = false;
             if (boost::algorithm::starts_with(value, "no_exit"))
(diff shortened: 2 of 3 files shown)


src/extractor/restriction_parser.cpp
```diff
@@ -59,31 +75,13 @@ RestrictionParser::TryParse(const osmium::Relation &relation) const
...
-        filter.add(true, "restriction:" + namespaced);
-    }
-
     const osmium::TagList &tag_list = relation.tags();
 
     osmium::tags::KeyFilter::iterator fi_begin(filter, tag_list.begin(), tag_list.end());
     osmium::tags::KeyFilter::iterator fi_end(filter, tag_list.end(), tag_list.end());
 
     // if it's not a restriction, continue;
-    if (std::distance(fi_begin, fi_end) == 0)
```
SiarheiFedartsou, line 86: It is O(N) for this iterator if I understand it correctly.

src/extractor/restriction_parser.cpp
```diff
@@ -126,6 +126,25 @@ RestrictionParser::TryParse(const osmium::Relation &relation) const
         {
             return {};
```
SiarheiFedartsou, line 127: Noticed that here we do `return {}` if found something unexpected in tag values, should we just ignore it instead? may be with some additional warning

src/extractor/restriction_parser.cpp
```diff
@@ -126,6 +126,25 @@ RestrictionParser::TryParse(const osmium::Relation &relation) const
...
+
+            if (parsed.empty())
+                continue;
+
+            for (const auto &p : parsed)
+            {
+                std::vector<util::OpeningHours> hours = util::ParseOpeningHours(p.condition);
+                // found unrecognized condition, continue
+                if (hours.empty())
+                    return {};
```
SiarheiFedartsou, line 143: The same, if we found something unexpected we just return, should it be just ignoring instead?

src/extractor/restriction_parser.cpp
```diff
@@ -99,18 +97,20 @@ RestrictionParser::TryParse(const osmium::Relation &relation) const
     bool is_multi_from = false;
     bool is_multi_to = false;
 
+    std::vector<util::OpeningHours> condition;
+
     for (; fi_begin != fi_end; ++fi_begin)
     {
-        const std::string key(fi_begin->key());
```
SiarheiFedartsou, line 104: Curious why didn’t we have warning about unused `key` variable here 🤔

include/extractor/restriction_parser.hpp
```diff
@@ -43,15 +45,16 @@ class RestrictionParser
...
     std::vector<InputTurnRestriction> TryParse(const osmium::Relation &relation) const;
 
   private:
     bool ShouldIgnoreRestriction(const std::string &except_tag_string) const;
 
     bool use_turn_restrictions;
     bool parse_conditionals;
-    std::vector<std::string> restrictions;
+    std::set<std::string> restrictions;
+    osmium::tags::KeyFilter filter;
```
mjjbell, line 57: The Osmium docs say this is [deprecated](https://docs.osmcode.org/libosmium/latest/namespaceosmium_1_1tags.html#a75df2aa39d225085eeffd438b701bfa1), and should be replaced with `osmium::TagsFilter`.

src/extractor/restriction_parser.cpp
mjjbell: In most cases you'll get at most one restriction returned per relation, it's only no entry/exits that could have multiple.

I think the way to view it is, if the relation is invalid, how correct will the remaining parsed restrictions be? It's difficult to know, so it's cleaner to not to try and deal with them.

We could add a log, but I would only make it a debug level. It's not OSRM's problem if input data is not correct.

mjjbell:
Can we run a benchmark on a reasonably sized OSM extract to evaluate the performance?

SiarheiFedartsou:
> Can we run a benchmark on a reasonably sized OSM extract to evaluate the performance?

Well, I ran it on 1.5GB poland-latest.osm.pbf(let me know if I should use something larger). I ran it for 3 times on both current master and this PR and used this value as a metric(because related code is executed only before this log):
Screenshot 2022-08-29 at 18 18 26



On current master(on my M1 Pro + 32GB + 10 threads with `./osrm-extract -p ../profiles/car.lua poland-latest.osm.pbf`):
```
(35.713 + 36.4149 +37.1714) / 3 = 36.4331
```

With this PR:
```
(36.2307 + 35.4937 + 36.8094) / 3 = 36.1780
```

Results are definitely not impressive.

src/extractor/restriction_parser.cpp
```diff
@@ -270,17 +264,25 @@ bool RestrictionParser::ShouldIgnoreRestriction(const std::string &except_tag_st
...
-        except_tag_string.begin(), except_tag_string.end(), delimiter_re, -1);
-    std::sregex_token_iterator except_tags_end;
-
-    return std::any_of(except_tags_begin, except_tags_end, [&](const std::string &current_string) {
-        return std::end(restrictions) !=
-               std::find(std::begin(restrictions), std::end(restrictions), current_string);
-    });
+    // split `except_tag_string` by semicolon and check if any of items is in `restrictions`
+    std::string current_string;
+    for (size_t index = 0; index < except_tag_string.size(); index++)
```
mjjbell, line 269:
```suggestion
    for (auto index : util::irange<size_t>(0, except_tag_string.size())
```

mjjbell (approved):

mjjbell:
> Results are definitely not impressive.

It's to be expected given the number of restrictions is small relative to other elements being parsed. In any case, it's tidying up the code for future improvements :+1:

SiarheiFedartsou:
@mjjbell it seems this new TagsParser is incompatible with clang 6. I'd propose to go back to KeysFilter(as I understand bumping the compiler is not an option). WDYT?
Screenshot 2022-08-29 at 23 33 36

mjjbell:
> I'd propose to go back to KeysFilter(as I understand bumping the compiler is not an option). WDYT?

:+1:

PR #6345, opened 2022-08-28
Use apt-get to install Clang on CI

SiarheiFedartsou:
# Issue

related to https://github.com/Project-OSRM/osrm-backend/issues/6266

Also will allow us to bump Clang version in the future.

## Tasklist

 - [ ] review
 - [ ] adjust for comments

## Requirements / Relations

### DIFF
--- a/.github/workflows/osrm-backend.yml
+++ b/.github/workflows/osrm-backend.yml
@@ -517,12 +517,12 @@ jobs:
         echo "$(${MASON} prefix ccache ${CCACHE_VERSION})/bin" >> $GITHUB_PATH
 
         # clang
-        if [[ -n ${CLANG_VERSION} ]]; then
+       # if [[ -n ${CLANG_VERSION} ]]; then
           echo "CCOMPILER=clang" >> $GITHUB_ENV
           echo "CXXCOMPILER=clang++" >> $GITHUB_ENV
-          sudo apt-get update -y && sudo apt-get install clang++-${CLANG_VERSION}
-          sudo update-alternatives --install /usr/bin/clang++ clang++ /usr/bin/clang++-${CLANG_VERSION} 100
-          sudo update-alternatives --install /usr/bin/clang clang /usr/bin/clang-${CLANG_VERSION} 100
+          sudo apt-get update -y && sudo apt-get install clang++-6
+          sudo update-alternatives --install /usr/bin/clang++ clang++ /usr/bin/clang++-6 100
+          sudo update-alternatives --install /usr/bin/clang clang /usr/bin/clang-6 100
 
           # we only enable lto for release builds
           # and therefore don't need to us ld.gold or llvm tools for linking
@@ -531,7 +531,7 @@ jobs:
             ${MASON} install binutils 2.27
             echo "$(${MASON} prefix binutils 2.27)/bin" >> $GITHUB_PATH
           fi
-        fi
+       # fi
 
         # Linux dev packages
         if [ "${TARGET_ARCH}" != "i686" ] && [ "${ENABLE_CONAN}" != "ON" ]; then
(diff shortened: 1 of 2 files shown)


.github/workflows/osrm-backend.yml
```diff
@@ -518,8 +518,14 @@ jobs:
         if [[ -n ${CLANG_VERSION} ]]; then
           echo "CCOMPILER=clang" >> $GITHUB_ENV
           echo "CXXCOMPILER=clang++" >> $GITHUB_ENV
-          ${MASON} install clang++ ${CLANG_VERSION}
-          echo "$(${MASON} prefix clang++ ${CLANG_VERSION})/bin" >> $GITHUB_PATH
+          sudo apt-get update -y && sudo apt-get install "clang++-${CLANG_VERSION}"
+          if [[ "${CLANG_VERSION}" = "6" ]]; then
```
SiarheiFedartsou, line 522: Clang 6 is installed as `/usr/bin/clang++-6.0` and Clang 11 as `/usr/bin/clang++-11`, so had to do this.
It is a bit ugly, but couldn't find better way.

.github/workflows/osrm-backend.yml
```diff
@@ -556,6 +561,12 @@ jobs:
         tar zxvf onetbb.tgz
         sudo cp -a oneapi-tbb-${TBB_VERSION}/lib/. /usr/local/lib/
         sudo cp -a oneapi-tbb-${TBB_VERSION}/include/. /usr/local/include/
+    - name: Check clang version
```
SiarheiFedartsou, line 564: Kind of self-check that actual clang version is the same as expected.

.github/workflows/osrm-backend.yml
```diff
@@ -138,7 +138,7 @@ jobs:
             runs-on: ubuntu-20.04
             BUILD_TOOLS: ON
             BUILD_TYPE: Debug
-            CLANG_VERSION: 6.0.0
+            CLANG_VERSION: 6
```
SiarheiFedartsou, line 141: We could just use `CCOMPILER/CXXCOMPILER` to use pre-installed clang, but the lowest pre-installed version is only 10: https://github.com/actions/runner-images/blob/main/images/linux/Ubuntu2004-Readme.md

.github/workflows/osrm-backend.yml
```diff
@@ -157,7 +157,7 @@ jobs:
             runs-on: ubuntu-20.04
             BUILD_TOOLS: ON
             BUILD_TYPE: Release
-            CLANG_VERSION: 11.0.0
+            CLANG_VERSION: 11
```
SiarheiFedartsou, line 160: Do we have any obligations to support certain old compilers btw?

mjjbell: Ideally for release builds we want to use a compiler that supports the minimum requirements of the oldest Node platform in our matrix. This is to avoid any ABI/stdlib incompatibilities with the host environments (example: #4567)

This is the toolchain info for Node 12: https://github.com/nodejs/node/blob/v12.x/BUILDING.md

.github/workflows/osrm-backend.yml
mjjbell: Is this the same package as https://packages.ubuntu.com/focal/clang-6.0?

SiarheiFedartsou: I think so.

SiarheiFedartsou: Screenshot 2022-08-29 at 17 42 08

mjjbell: Is it possible to set CCOMPILER/CXXCOMPILER to e.g. clang-6.0/clang++6.0 and then just install all compilers at L543? Or is there a step that's requiring the symlink?

SiarheiFedartsou: Will try

SiarheiFedartsou: It seems it works, a way more elegant solution, thanks :+1:

mjjbell (approved):

PR #6346, opened 2022-08-28
Use Lua 5.4 in Docker image

SiarheiFedartsou:
# Issue
Updates Lua in Dockerfile to the latest 5.4 version. [Some benchmarks](https://lwn.net/Articles/826134/) say it is up to 40% faster, but I got only improvement from ~8.1s to ~7.9s on my M1 Pro for `time docker run -t -v "${PWD}:/data" osrm-test osrm-extract -p /opt/car.lua /data/berlin-latest.osm.pbf`.

## Tasklist

 - [ ] review
 - [ ] adjust for comments

## Requirements / Relations

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@
     - NodeJS:
       - FIXED: Support `skip_waypoints` in Node bindings [#6060](https://github.com/Project-OSRM/osrm-backend/pull/6060)
     - Misc:
+      - CHANGED: Use Lua 5.4 in Docker image. [#6346](https://github.com/Project-OSRM/osrm-backend/pull/6346)
       - CHANGED: Remove redundant nullptr check. [#6326](https://github.com/Project-OSRM/osrm-backend/pull/6326)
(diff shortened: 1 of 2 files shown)


mjjbell:
Is that total running time for `osrm-extract` or just for the scripting stage?
I think it would be good to run it on a continental sized dataset to get a better gauge of performance.

sol2 is likely to have a big impact on the scripting performance in any case.

SiarheiFedartsou:
> Is that total running time for `osrm-extract` or just for the scripting stage? I think it would be good to run it on a continental sized dataset to get a better gauge of performance.
> 
> sol2 is likely to have a big impact on the scripting performance in any case.

It is total running time of osrm-extract. Tbh I had no goal to do it and have some significant performance update - the goal here is probably to just update dependency. I just noticed these promises and decided to check the difference just out of curiousity not spending more than 10 minutes on it. Do you think we should invest more time into it? 
Btw as per sol2 docs they have really small overhead over C API https://sol2.readthedocs.io/en/latest/benchmarks.html

mjjbell (approved):

Issue #6329, opened 2022-08-24
Deprecate .osrm file


mjjbell:
The `.osrm` file contains a node-based graph representation of the OSM data, prior to compression and edge-expansion.

It's generated by `osrm-extract` and only read by `osrm-extract`. I believe it's a hangover from when the data was too large for RAM, so was first serialised to disk. However the file contents are [immediately read back into memory](https://github.com/Project-OSRM/osrm-backend/blob/41dda32546399f1dc12af1de41668993de44c7dc/src/extractor/node_based_graph_factory.cpp#L37) anyway.

Assuming the file had value beyond the extraction phase, there are inconsistencies in which metadata components are serialised.
[Traffic signals and barriers](https://github.com/Project-OSRM/osrm-backend/blob/a98074a051ddd6fb95aa7e7d392420b9af0c0281/src/extractor/extraction_containers.cpp#L896-L933) are serialised to disk as part of the graph, but other metadata, [such as turn lanes, restrictions and manoeuvrer overrides ](https://github.com/Project-OSRM/osrm-backend/blob/91895604c9e458d1c804189b348e86dfdb2cc31c/src/extractor/extractor.cpp#L629-L631)are not. They are only kept in memory long enough for subsequent extraction steps.

Therefore, I propose that we scrap the disk serialise/deserialise step and stop generating this file.

`.osrm` is one of, if not the largest file generated during processing. Given we generate many other specific files for use in later processing stages, we can stop persisting the `.osrm` file entirely to save storage space, and if needed serialise its contents to more targeted files.

mjjbell:
The[ toolchain file overview ](https://github.com/Project-OSRM/osrm-backend/wiki/Toolchain-file-overview) refers to `.osrm` as a temporary file. However, there is a dependency on [the file](https://github.com/Project-OSRM/osrm-backend/blob/91895604c9e458d1c804189b348e86dfdb2cc31c/src/tools/components.cpp#L46) in `osrm-components`.

I'm not sure of the maintenance status of that tool, I've never used it myself. In any case, we could make serialising an optional setting on `osrm-extract` for debug purposes such as this.

SiarheiFedartsou:
Is it about the same? https://github.com/Project-OSRM/osrm-backend/issues/2055

mjjbell:
Yes it sounds similar.

I would suggest holding off until #6153 is applied, as it has related changes.
What is the proposed fix? To make serialising to disk optional?

[...]

PR #6354, opened 2022-08-31
Do not generate intermediate .osrm file in osrm-extract.

SiarheiFedartsou:
# Issue

closes https://github.com/Project-OSRM/osrm-backend/issues/6329

## Tasklist

 - [ ] review
 - [ ] adjust for comments

## Requirements / Relations


SiarheiFedartsou:
Some things to discuss:
1. Now in order to generate `.osrm` file user has to pass `--dump-nbg-graph` flag to `osrm-extract`. Is it a good naming? Should we may be make it default and allow to disable by CLI option? I can imagine that someone somehow relies on this file and uses `osrm-components` with it(it is really low chance IMO, but still).
2. The only place where this `.osrm` can be used is `osrm-components` tool which is as I understand purely for debug, so IMO `.osrm` feels a bit "loud" name for such kind of files. May be rename it to something?
3. We suggest to pass this `.osrm` file everywhere in the docs as if it is really used as input file(e.g. `./osrm-customize ./berlin-latest.osrm`). Actually it is just a base name. So I suggest to just remove this extension from everywhere: like `./osrm-partition ./berlin-latest`. But for backward compatibility leave the ability to work if it was passed in the old way(we may emit warning if it used this way). I can imagine that a lot of people have working pipelines with this `.osrm` in them. 

@mjjbell WDYT?

.github/workflows/osrm-backend.yml
```diff
@@ -95,10 +95,16 @@ jobs:
           # when `--memory-swap` value equals `--memory` it means container won't use swap
           # see https://docs.docker.com/config/containers/resource_constraints/#--memory-swap-details
           MEMORY_ARGS="--memory=1g --memory-swap=1g"
-          docker run $MEMORY_ARGS -t -v "${PWD}:/data" "${TAG}" osrm-extract -p /opt/car.lua /data/berlin-latest.osm.pbf
-          docker run $MEMORY_ARGS -t -v "${PWD}:/data" "${TAG}" osrm-partition /data/berlin-latest.osrm
-          docker run $MEMORY_ARGS -t -v "${PWD}:/data" "${TAG}" osrm-customize /data/berlin-latest.osrm
-          docker run $MEMORY_ARGS --name=osrm-container -t -p 5000:5000 -v "${PWD}:/data" "${TAG}" osrm-routed --algorithm mld /data/berlin-latest.osrm &
+          docker run $MEMORY_ARGS -t -v "${PWD}:/data" "${TAG}" osrm-extract --dump-nbg-graph -p /opt/car.lua /data/berlin-latest.osm.pbf
+          docker run $MEMORY_ARGS -t -v "${PWD}:/data" "${TAG}" osrm-components /data/berlin-latest.osrm /data/berlin-latest.geojson
```
SiarheiFedartsou, line 99: Added kind of smoke test for `osrm-components`.

include/extractor/extraction_containers.hpp
```diff
@@ -75,6 +72,8 @@ class ExtractionContainers
     std::vector<InputTrafficSignal> external_traffic_signals;
     TrafficSignals internal_traffic_signals;
 
+    std::vector<NodeBasedEdge> normal_edges;
```
SiarheiFedartsou, line 75: Not sure about naming here, tbh haven't tried to dive into the logic there, just made it available from outside.

include/extractor/extraction_containers.hpp
```diff
@@ -84,11 +83,12 @@ class ExtractionContainers
 
     std::vector<InputManeuverOverride> external_maneuver_overrides_list;
     std::vector<UnresolvedManeuverOverride> internal_maneuver_overrides;
+    std::unordered_set<NodeID> internal_barrier_nodes;
```
SiarheiFedartsou, line 86: The same about naming.

include/extractor/extractor.hpp
```diff
@@ -60,14 +60,25 @@ class Extractor
...
+
   private:
     ExtractorConfig config;
 
-    std::tuple<LaneDescriptionMap,
-               std::vector<TurnRestriction>,
-               std::vector<UnresolvedManeuverOverride>,
-               TrafficSignals>
-    ParseOSMData(ScriptingEnvironment &scripting_environment, const unsigned number_of_threads);
+    ParsedOSMData ParseOSMData(ScriptingEnvironment &scripting_environment,
```
SiarheiFedartsou, line 80: IMO this `ParsedOSMData` is a bit more readable than tuple we used before.

src/extractor/extractor.cpp
```diff
@@ -629,10 +624,38 @@ Extractor::ParseOSMData(ScriptingEnvironment &scripting_environment,
...
+    osm_node_ids.reserve(extraction_containers.internal_nodes.size());
+    for (size_t index = 0; index < extraction_containers.internal_nodes.size(); ++index)
+    {
+        const auto &current_node = extraction_containers.internal_nodes[index];
+        osm_coordinates[index].lon = current_node.lon;
+        osm_coordinates[index].lat = current_node.lat;
+        osm_node_ids.push_back(current_node.node_id);
+    }
+
+    if (config.dump_nbg_graph)
```
SiarheiFedartsou, line 640: Now we are dumping `.osrm` file only if it was explicitly requested.

include/extractor/node_based_graph_factory.hpp
```diff
@@ -38,11 +38,15 @@ class NodeBasedGraphFactory
     // node-based graph to represent the OSM network. This includes geometry compression, annotation
     // data optimisation and many other aspects. After this step, the edge-based graph factory can
     // turn the graph into the routing graph to be used with the navigation algorithms.
-    NodeBasedGraphFactory(const boost::filesystem::path &input_file,
-                          ScriptingEnvironment &scripting_environment,
+    NodeBasedGraphFactory(ScriptingEnvironment &scripting_environment,
```
mjjbell, line 41: Above comment will need updating.

README.md
```diff
@@ -62,12 +62,12 @@ Pre-process the extract with the car profile and start a routing engine HTTP ser
 
 The flag `-v "${PWD}:/data"` creates the directory `/data` inside the docker container and makes the current working directory `"${PWD}"` available there. The file `/data/berlin-latest.osm.pbf` inside the container is referring to `"${PWD}/berlin-latest.osm.pbf"` on the host.
 
-    docker run -t -v "${PWD}:/data" osrm/osrm-backend osrm-partition /data/berlin-latest.osrm
-    docker run -t -v "${PWD}:/data" osrm/osrm-backend osrm-customize /data/berlin-latest.osrm
+    docker run -t -v "${PWD}:/data" osrm/osrm-backend osrm-partition /data/berlin-latest
+    docker run -t -v "${PWD}:/data" osrm/osrm-backend osrm-customize /data/berlin-latest
 
-Note that `berlin-latest.osrm` has a different file extension. 
+Note that `berlin-latest` has no file extension.
```
mjjbell, line 68: Referring just to the base path feels a bit odd, but I don't really have a better suggestion. Maybe always use the OSM file as the argument?

mjjbell (changes requested):

src/extractor/node_based_graph_factory.cpp
```diff
@@ -16,32 +16,30 @@ namespace extractor
 {
 
 NodeBasedGraphFactory::NodeBasedGraphFactory(
-    const boost::filesystem::path &input_file,
     ScriptingEnvironment &scripting_environment,
     std::vector<TurnRestriction> &turn_restrictions,
     std::vector<UnresolvedManeuverOverride> &maneuver_overrides,
-    const TrafficSignals &traffic_signals)
+    const TrafficSignals &traffic_signals,
+    std::unordered_set<NodeID> &&barriers,
```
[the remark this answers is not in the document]
SiarheiFedartsou: Yes, only in this case we will have it actually moved https://stackoverflow.com/questions/63105858/what-happens-when-stdmove-is-called-on-a-rvalue-reference

include/extractor/extraction_containers.hpp
SiarheiFedartsou: Renamed to `used_barrier_nodes` as we already use `barrier_nodes` for barrier's OSM ids.

.github/workflows/osrm-backend.yml
```diff
@@ -95,10 +95,22 @@ jobs:
...
-          docker run $MEMORY_ARGS -t -v "${PWD}:/data" "${TAG}" osrm-extract -p /opt/car.lua /data/berlin-latest.osm.pbf
+          docker run $MEMORY_ARGS -t -v "${PWD}:/data" "${TAG}" osrm-extract --dump-nbg-graph -p /opt/car.lua /data/berlin-latest.osm.pbf
+          docker run $MEMORY_ARGS -t -v "${PWD}:/data" "${TAG}" osrm-components /data/berlin-latest.osrm /data/berlin-latest.geojson
+          if [ ! -s "${PWD}/berlin-latest.geojson" ]
+          then
+            >&2 echo "No berlin-latest.geojson found"
+            exit 1
+          fi
+
+          # here we check that `osrm-partition` accepts base file path with `.osrm` extension even if `.osrm` file doesn't exist
```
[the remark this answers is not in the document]
SiarheiFedartsou: Updated it

SiarheiFedartsou:
@mjjbell can you take a look again?

And also let's discuss these questions:

> Some things to discuss:
> 
> 1. Now in order to generate `.osrm` file user has to pass `--dump-nbg-graph` flag to `osrm-extract`. Is it a good naming? Should we may be make it default and allow to disable by CLI option? I can imagine that someone somehow relies on this file and uses `osrm-components` with it(it is really low chance IMO, but still).
> 2. The only place where this `.osrm` can be used is `osrm-components` tool which is as I understand purely for debug, so IMO `.osrm` feels a bit "loud" name for such kind of files. May be rename it to something?
> 3. We suggest to pass this `.osrm` file everywhere in the docs as if it is really used as input file(e.g. `./osrm-customize ./berlin-latest.osrm`). Actually it is just a base name. So I suggest to just remove this extension from everywhere: like `./osrm-partition ./berlin-latest`. But for backward compatibility leave the ability to work if it was passed in the old way(we may emit warning if it used this way). I can imagine that a lot of people have working pipelines with this `.osrm` in them.
> 
> @mjjbell WDYT?

For 1 I doubt someone really relies on it and it is super straightforward to just pass this additional flag if it is really needed.

For 3 I now more inclined to leave this `.osrm` extension everywhere with a note that it is not a specific file anymore, but base path for all other files we need(all of them have `.osrm.*` pattern in name)

src/extractor/node_based_graph_factory.cpp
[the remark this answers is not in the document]
mjjbell: Oops, yes of course.

mjjbell:
>  Now in order to generate `.osrm` file user has to pass `--dump-nbg-graph` flag to `osrm-extract`. Is it a good naming? Should we may be make it default and allow to disable by CLI option? I can imagine that someone somehow relies on this file and uses `osrm-components` with it(it is really low chance IMO, but still).

>  The only place where this `.osrm` can be used is `osrm-components` tool which is as I understand purely for debug, so IMO `.osrm` feels a bit "loud" name for such kind of files. May be rename it to something?

It's the internal node-based-graph representation prior to compression, creation of dummy edges and other stuff that doesn't exist in the OSM data. So I would say it's the correct arg name, but I agree that it's not obvious that this should generate an `.osrm` file. 

Renaming it to `.nbg` seems like a good fit with the other [toolchain output](https://github.com/Project-OSRM/osrm-backend/wiki/Toolchain-file-overview), given we have already have the OSM IDs and coordinates in `.nbg_nodes`, and the compressed graph in `.cnbg`.

>  We suggest to pass this `.osrm` file everywhere in the docs as if it is really used as input file(e.g. `./osrm-customize ./berlin-latest.osrm`). Actually it is just a base name. So I suggest to just remove this extension from everywhere: like `./osrm-partition ./berlin-latest`. But for backward compatibility leave the ability to work if it was passed in the old way(we may emit warning if it used this way). I can imagine that a lot of people have working pipelines with this `.osrm` in them.

Thinking longer-term, I can see an approach where we pass around a metadata/config file, that includes information about the settings used during processing, and detailing which datasets have been generated (e.g. if we wanted to support [optional loading](https://github.com/Project-OSRM/osrm-backend/issues/5838)), which maybe could be a good use for an `.osrm` file? In which case it would be beneficial to keep the instructions for passing that as the argument. What do you think?

mjjbell:
> For 3 I now more inclined to leave this .osrm extension everywhere with a note that it is not a specific file anymore, but base path for all other files we need(all of them have .osrm.* pattern in name)

Sorry I'm reading everything in the wrong order. In that's the case, then I agree, and it also aligns with the potential of a config/metadata file in the future.

README.md
```diff
@@ -65,7 +65,7 @@ The flag `-v "${PWD}:/data"` creates the directory `/data` inside the docker con
     docker run -t -v "${PWD}:/data" osrm/osrm-backend osrm-partition /data/berlin-latest.osrm
     docker run -t -v "${PWD}:/data" osrm/osrm-backend osrm-customize /data/berlin-latest.osrm
 
-Note that `berlin-latest.osrm` has a different file extension. 
+Note there is no `berlin-latest.osrm` file, but multiple `berlin-latest.osrm.*` files, i.e. `berlin-latest.osrm` is not file path, but "base" path referring to set of files and there is an option to omit this `.osrm` suffix completely(e.g. `osrm-partition /data/berlin-latest`).
```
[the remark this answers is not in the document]
mjjbell: Yep, agreed.

src/extractor/extractor.cpp
```diff
@@ -629,10 +624,36 @@ Extractor::ParseOSMData(ScriptingEnvironment &scripting_environment,
...
+    {
+        const auto &current_node = extraction_containers.used_nodes[index];
+        osm_coordinates[index].lon = current_node.lon;
+        osm_coordinates[index].lat = current_node.lat;
+        osm_node_ids.push_back(current_node.node_id);
+    }
+
+    if (config.dump_nbg_graph)
+    {
+        storage::tar::FileWriter writer(config.GetPath(".osrm").string(),
```
mjjbell, line 642: Only suggested change would be to make this `.nbg`

mjjbell (approved):

mjjbell:
> For 1 I doubt someone really relies on it and it is super straightforward to just pass this additional flag if it is really needed.

Agreed. Also, given the `.osrm` file is marked as temporary in the [toolchain](https://github.com/Project-OSRM/osrm-backend/wiki/Toolchain-file-overview), I don't think users can have expectations of stability.

PR #6372, opened 2022-09-21
Fix a couple of warnings found by PVS Studio

SiarheiFedartsou:
# Issue

I asked for a [PVS Studio](https://pvs-studio.com/en/) license for us(they give it for free for open source projects) and it was able to find a couple of issues like this.

## Tasklist

 - [ ] review
 - [ ] adjust for comments

## Requirements / Relations

### DIFF
--- a/.github/workflows/osrm-backend.yml
+++ b/.github/workflows/osrm-backend.yml
@@ -607,7 +607,7 @@ jobs:
       
         # Linux dev packages
         if [ "${TARGET_ARCH}" != "i686" ] && [ "${ENABLE_CONAN}" != "ON" ]; then
-          sudo add-apt-repository ppa:ubuntu-toolchain-r/test -y
+          #sudo add-apt-repository ppa:ubuntu-toolchain-r/test -y
           sudo apt-get update -y
           sudo apt-get install -y libbz2-dev libxml2-dev libzip-dev liblua5.2-dev libboost-all-dev
           if [[ "${CCOMPILER}" != clang-* ]]; then
--- a/.github/workflows/osrm-backend.yml
+++ b/.github/workflows/osrm-backend.yml
@@ -607,7 +607,6 @@ jobs:
       
         # Linux dev packages
         if [ "${TARGET_ARCH}" != "i686" ] && [ "${ENABLE_CONAN}" != "ON" ]; then
-          #sudo add-apt-repository ppa:ubuntu-toolchain-r/test -y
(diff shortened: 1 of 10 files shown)


include/util/coordinate_calculation.hpp
```diff
@@ -372,7 +372,7 @@ bool areParallel(const iterator_type lhs_begin,
     const auto slope_rhs = get_slope(null_island, rotated_difference_rhs);
     // the left hand side has a slope of `0` after the rotation. We can check the slope of the right
     // hand side to ensure we only considering slight slopes
-    return std::abs(slope_rhs) < 0.20; // twenty percent incline at the most
+    return std::fabs(slope_rhs) < 0.20; // twenty percent incline at the most
```
SiarheiFedartsou, line 375: This one is the most interesting - std::abs has overloads only for integer types. 
https://en.cppreference.com/w/cpp/numeric/math/abs

.github/workflows/osrm-backend.yml
```diff
@@ -607,7 +607,6 @@ jobs:
       
         # Linux dev packages
         if [ "${TARGET_ARCH}" != "i686" ] && [ "${ENABLE_CONAN}" != "ON" ]; then
-          sudo add-apt-repository ppa:ubuntu-toolchain-r/test -y
```
SiarheiFedartsou, line 610: It started failing unexpectedly, but it turned out that we probably don't need it :)

SiarheiFedartsou:
@mjjbell please take a look when you have a chance :)

.github/workflows/osrm-backend.yml
mjjbell: Yes I think we're lucky that 20.04 supports the 5 gcc versions in our matrix. If the distro and matrix get out of sync, we'll need to bring it back (or just limit it to the versions the distro supports).

mjjbell (approved):
Is this something we can run continuously?

I see they have a docs page for running via Github Actions: https://pvs-studio.com/en/docs/manual/6579/

SiarheiFedartsou:
> Is this something we can run continuously?
> 
> I see they have a docs page for running via Github Actions: https://pvs-studio.com/en/docs/manual/6579/

We definitely can, will try to look into it at some point. 

Btw will you be able to take a look into other opened PRs? you probably missed notifications about them(or just don’t have time for it, not sure if you have enough time at the moment)

Issue #6374, opened 2022-09-23
Timestamps for logs


nilsnolde:
I’d like to add timestamps to the processing logs of the various commands. I didn’t look into the code base yet, not sure if the logging is centralized, so it’d be quick to add, or it needs a logging abstraction.

We had some issues lately where custom PBFs take a whole longer to pass through osrm-contract. Timestamps would help to see in which phase of the pipeline it’s hanging. And it’s general good practice IMO. Shout out if any objections/thoughts.

SiarheiFedartsou:
At glance it looks like it will be as easy as modifying this `Log` class. https://github.com/Project-OSRM/osrm-backend/blob/be353630d5e845b1dc64a4b5e467b6a4bc1addbf/include/util/log.hpp#L56

nilsnolde:
Thanks, that was what I hoping for:)

SiarheiFedartsou:
Here is draft https://github.com/Project-OSRM/osrm-backend/pull/6375
How it looks now for osrm-extract:
<img width="964" alt="Screenshot 2022-09-23 at 16 25 16" src="https://user-images.githubusercontent.com/266271/191983595-8f8297df-f0be-4b8d-b752-eed6453653de.png">

I still see some logs which have no timestamps though and also not sure we need one more dependency because of it, but fmtlib is nice library(and actually is the same as `std::format` from C++20)

SiarheiFedartsou:
> I still see some logs which have no timestamps though

This `ok, after` is the "end" of `Collecting traffic signal information on 2292 signals...`. 

<img width="1492" alt="Screenshot 2022-09-23 at 16 35 53" src="https://user-images.githubusercontent.com/266271/191985936-2e213570-06f1-43ec-87e6-598d73d15203.png">

We use `UnbufferedLog` to output it https://github.com/Project-OSRM/osrm-backend/blob/be353630d5e845b1dc64a4b5e467b6a4bc1addbf/src/extractor/extraction_containers.cpp#L1196

I.e. if some other logs are happening in the middle(like these warnings in this case) they are "interleave" logs coming from this `UnbufferedLog`, tbh I'd stop using this.

SiarheiFedartsou:
Current log system is prone to race conditions, i.e. if we are doing something like:

```
util::Log() << "Hello, " << "OSRM";
```

we have no guarantee that log will be outputted "atomically", instead we may have it interleaved by some another log happening in another thread. 

[...]

PR #6375, opened 2022-09-23
Add timestamps for logs

SiarheiFedartsou:
# Issue

https://github.com/Project-OSRM/osrm-backend/issues/6374

## Tasklist

 - [ ] review
 - [ ] adjust for comments

## Requirements / Relations


CMakeLists.txt
```diff
@@ -455,6 +455,11 @@ add_subdirectory(${FLATBUFFERS_SRC_DIR}
         ${CMAKE_CURRENT_BINARY_DIR}/flatbuffers-build
         EXCLUDE_FROM_ALL)
 
+set(FMT_INCLUDE_DIR "${CMAKE_CURRENT_SOURCE_DIR}/third_party/fmt-9.1.0/include")
```
SiarheiFedartsou, line 458: Btw fmt promises to have super fast algorithm for floating point numbers rendering https://github.com/fmtlib/dtoa-benchmark#results

We can try to re-use it in our JSON renderer here https://github.com/Project-OSRM/osrm-backend/blob/ef8f3d7508dbdcaeac990d1b891d2e0775c594e6/include/util/json_renderer.hpp#L41

mjjbell:
Is there a benefit to vendoring `fmt` vs adding as a Conan dependency?

SiarheiFedartsou:
> Is there a benefit to vendoring `fmt` vs adding as a Conan dependency?

Well, we have to support ENABLE_CONAN=OFF use case and it is not clear how to deal with it.

mjjbell:
Ah, I thought we had started using it all the time for some reason.

mjjbell (approved):

SiarheiFedartsou:
> Ah, I thought we had started using it all the time for some reason.

Btw I would discuss that, tbh it is not clear what can be benefits of using dependencies from the system rather than "pinned" versions from Conan. I can just imagine that system dependencies may be of newer versions containing bug fixes etc. At the same time dependency management with Conan is much easier.

nilsnolde:
> tbh it is not clear what can be benefits of using dependencies from the system

My 2c: I think it's mostly about not duplicating libraries (esp linked ones). Also package maintainers (brew, apt, pacman etc) really don't like runtime linked dependencies to be packaged with a library but use the stuff that's inside the package repos and sometimes go to great lengths to patch sources to accomplish that. The biggest issue though is probably linked shared libraries when you distribute OSRM as a package.. OSRM will fail to find the runtime (shared lib) dependencies if they're not installed system-wide (and/or if conan sets some RPATH in the exes/libs). And then it'd have to be compatible with whatever the system packages provides. But I'd guess that's less of a problem for software like OSRM/Valhalla, I personally would go with using conan by default, it's so much less painful for the maintainers to pin versions. And this is FOSS, if people need compat with older versions of deps they can open PRs:)

PR #6376, opened 2022-09-25
Ensure u-turn exists in intersection view

### LINKED ISSUE #6218: osrm-extract assert triggering on guidance intersection handler
With the [Washington State OSM extract](https://download.geofabrik.de/north-america/us/washington.html).

Applying the fix #6215 to master.

`osrm-extract` with the default foot profile triggers an angle assertion when processing intersection guidance.

https://github.com/Project-OSRM/osrm-backend/blob/79d4363d5972ea4d69abd92b26cda80de7941365/src/guidance/turn_handler.cpp#L239-L243

The offending intersection is in a bus station: https://www.openstreetmap.org/node/3750788725

image

The fact that the assertion is triggered within `handleThreeWayTurn` on a four-way intersection hints at the problem here. 

When processing the turns from one of the incoming edges to the intersection, it incorrectly identifies the straight-ahead turn as a u-turn, and discards the actual u-turn. 
https://github.com/Project-OSRM/osrm-backend/blob/79d4363d5972ea4d69abd92b26cda80de7941365/src/extractor/intersection/intersection_analysis.cpp#L654

Later in the processing it tries to fix the inconsistency in the angles as it rotates around the turns, effectively removing the u-turn from the list and making two 90 degree turns. This is what triggers the assertion.
https://github.com/Project-OSRM/osrm-backend/blob/79d4363d5972ea4d69abd92b26cda80de7941365/src/extractor/intersection/intersection_analysis.cpp#L697-L705

#### Potential Fix

I'm not familiar with the guidance code, but increasing the coordinate precision to 7 decimal places is sufficient to fix the problem. Which probably means the nodes involved in the angle calculations are too close together to be calculated accurately at the current level of  coordinate precision.

https://github.com/Project-OSRM/osrm-backend/blob/79d4363d5972ea4d69abd92b26cda80de7941365/include/util/coordinate.hpp#L44

It's been noted previously (e.g. #3368) that coordinate precision needs increasing to improve consistency, but now we have a case where it's breaking logical assumptions.

mjjbell:
# Context

The full graphic details are in #6218

The motivation for this PR is to unblock the debug build on very large (i.e. global) datasets. Currently they will fail on this assertion, so we are unable to reach issues further down the processing chain.

# Issue
Due to some rather complex logic that tries to calculate intersection angles by looking further up the road, it's possible to return an intersection view that is missing a u-turn - something which is assumed to exist in later guidance calculations.

We apply a fix here by ensuring the u-turn is always included in the returned view.

Ideally we would refactor what looks like some quite brittle intersection code, but that requires:
- better understanding of the guidance code
- a large and complicated change

so this appears to be the better alternative in the short-term.

## Tasklist

 - [ ] review
 - [ ] adjust for comments

## Requirements / Relations

Fixes #6218

### DIFF
--- a/src/extractor/intersection/intersection_analysis.cpp
+++ b/src/extractor/intersection/intersection_analysis.cpp
@@ -617,12 +617,13 @@ IntersectionView convertToIntersectionView(const util::NodeBasedDynamicGraph &gr
     std::vector<IntersectionViewDataWithAngle> pre_intersection_view;
     IntersectionViewData uturn{{SPECIAL_EDGEID, 0., 0., 0.}, false, 0.};
     std::size_t allowed_uturns_number = 0;
+
+    const auto is_uturn = [](const auto angle) {
+        return std::fabs(angle) < std::numeric_limits<double>::epsilon();
+    };
+
     for (const auto &outgoing_edge : outgoing_edges)
     {
-        const auto is_uturn = [](const auto angle) {
-            return std::fabs(angle) < std::numeric_limits<double>::epsilon();
-        };
-
         const auto edge_it = findEdge(edge_geometries, outgoing_edge.edge);
         const auto is_merged = merged_edges.count(outgoing_edge.edge) != 0;
         const auto is_turn_allowed = intersection::isTurnAllowed(graph,
@@ -678,6 +679,7 @@ IntersectionView convertToIntersectionView(const util::NodeBasedDynamicGraph &gr
     BOOST_ASSERT(uturn.eid != SPECIAL_EDGEID);
     if (uturn.entry_allowed || allowed_uturns_number == 0)
     { // Add the true U-turn if it is allowed or no other U-turns found
+        BOOST_ASSERT(uturn.angle == 0.);
         pre_intersection_view.insert(pre_intersection_view.begin(), {uturn, 0});
     }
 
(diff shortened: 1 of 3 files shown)


src/extractor/intersection/intersection_analysis.cpp
```diff
@@ -678,6 +679,7 @@ IntersectionView convertToIntersectionView(const util::NodeBasedDynamicGraph &gr
     BOOST_ASSERT(uturn.eid != SPECIAL_EDGEID);
     if (uturn.entry_allowed || allowed_uturns_number == 0)
     { // Add the true U-turn if it is allowed or no other U-turns found
+        BOOST_ASSERT(uturn.angle == 0.);
```
mjjbell, line 682: `angle` is the perceived angle between the incoming and outgoing edges at the intersection. It can differ from the real angle because it does smart stuff like look further ahead down the road to understand where we're actually heading.

For the u-turn incoming/outgoing pair, any changes in angle will always match, so the difference remains `0.`. This is the important assumption that by adding it here guarantees we always have a 0° turn in the intersection view.

src/extractor/intersection/intersection_analysis.cpp
```diff
@@ -706,6 +708,21 @@ IntersectionView convertToIntersectionView(const util::NodeBasedDynamicGraph &gr
...
+        pre_intersection_view.end(),
+        [&is_uturn](IntersectionViewDataWithAngle road) { return is_uturn(road.first.angle); });
+    // After all of this, if we now don't have a u-turn, let's add one to the intersection.
+    // This is a hack to fix the triggered assertion ( see:
+    // https://github.com/Project-OSRM/osrm-backend/issues/6218 ). Ideally we would fix this more
+    // robustly, but this will require overhauling all of the intersection logic.
+    if (no_uturn)
+    {
+        BOOST_ASSERT(!uturn.entry_allowed && allowed_uturns_number > 0);
+        BOOST_ASSERT(uturn.angle == 0.);
```
mjjbell, line 722: Same comment as above. The other assertion is because we only manually insert here if we didn't above.

SiarheiFedartsou (approved):

src/extractor/intersection/intersection_analysis.cpp
```diff
@@ -706,6 +708,21 @@ IntersectionView convertToIntersectionView(const util::NodeBasedDynamicGraph &gr
         }
     }
 
+    auto no_uturn = std::none_of(
+        pre_intersection_view.begin(),
+        pre_intersection_view.end(),
+        [&is_uturn](IntersectionViewDataWithAngle road) { return is_uturn(road.first.angle); });
```
SiarheiFedartsou, line 714: nit. probably worth passing road by const reference: I.e. `const auto& road`

PR #6380, opened 2022-09-30
Improve performance of JSON rendering

SiarheiFedartsou:
- wip
- Refactor json renderer to share implementation for vector and string
- Add string
- Optimize strings
- Get rid of copy
- add check
- use std::string in node bindings
- add comments

# Issue

## Tasklist

 - [ ] review
 - [ ] adjust for comments

## Requirements / Relations


.github/workflows/osrm-backend.yml
```diff
@@ -715,6 +715,7 @@ jobs:
         pushd ${OSRM_BUILD_DIR}
```
SiarheiFedartsou, line 715: Baseline on my M1 Pro/32 gb ram:
initial

SiarheiFedartsou: "string" here is `std::stringstream ss; ... = ss.str()`

include/util/json_renderer.hpp
```diff
@@ -64,131 +78,94 @@ struct Renderer
...
-    std::ostream &out;
+    void write(const std::string &str);
+    void write(const char *str);
+    void write(char ch);
+
+  private:
+    Out &out;
 };
 
-struct ArrayRenderer
```
SiarheiFedartsou, line 109: # 1.

First of all I noticed that for some reason we use completely different implementation for `std::vector` and `std::ostream` outputs. I unified it using `std::ostream` implementation as common one.
vector uses the same implementation

include/util/json_renderer.hpp
```diff
@@ -64,131 +78,94 @@ struct Renderer
...
 
 inline void render(std::ostream &out, const Object &object)
 {
-    Value value = object;
-    mapbox::util::apply_visitor(Renderer(out), value);
+    Renderer renderer(out);
+    renderer(object);
+}
+
+inline void render(std::string &out, const Object &object)
```
SiarheiFedartsou, line 159: # 2.
Then I decided to add Renderer specialization for `std::string`(i.e. stopped using `std::stringstream` and started writing directly to `std::string`) and started using it in benchmark:
 
string added

include/util/json_renderer.hpp
```diff
@@ -27,18 +27,32 @@ namespace
...
-    explicit Renderer(std::ostream &_out) : out(_out) {}
+    explicit Renderer(Out &_out) : out(_out) {}
 
-    void operator()(const String &string) const
+    void operator()(const String &string)
     {
-        out << "\"";
-        out << escape_JSON(string.value);
-        out << "\"";
+        write('"');
```
SiarheiFedartsou, line 36: # 2.
Then I noticed that we spend quite a lot of time on escaping the strings which usually don't need to be escaped:
precalculate escapeed string size

include/util/json_renderer.hpp
```diff
@@ -64,131 +78,94 @@ struct Renderer
...
+
+inline void render(std::string &out, const Object &object)
+{
+    Renderer renderer(out);
+    renderer(object);
 }
 
 inline void render(std::vector<char> &out, const Object &object)
 {
-    Value value = object;
```
SiarheiFedartsou, line 190: # 3.
And finally I noticed that we are doing unneeded copy of the whole json here and it gave the most significant improvement:
get rid of copy

src/nodejs/node_osrm.cpp
```diff
@@ -167,13 +167,13 @@ inline void async(const Nan::FunctionCallbackInfo<v8::Value> &info,
                 ParseResult(status, json_result);
                 if (pluginParams.renderToBuffer)
                 {
-                    std::ostringstream buf;
-                    osrm::util::json::render(buf, json_result);
-                    result = buf.str();
+                    std::string json_string;
```
SiarheiFedartsou, line 170: Started using `std::string` here as it is faster. `osrm-routed` uses `std::vector`.

SiarheiFedartsou: We probably can even try to go with `std::vector` here too(which is slightly faster according to benchmarks).

include/util/json_renderer.hpp
```diff
@@ -27,18 +27,32 @@ namespace
...
+            escaped.reserve(size);
+            EscapeJSONString(string.value, escaped);
+
+            write(escaped);
+        }
+        write('"');
     }
 
-    void operator()(const Number &number) const
+    void operator()(const Number &number)
```
SiarheiFedartsou, line 55: We spend quite a lot on number formatting: I've tried to optimize it using `fmt`(which promises to have faster formatting than current algorithm), but I couldn't notice the difference.

danpat: Yeah - I optimized this part a long time ago - if you dig into this function, you'll notice we're _already_ using a highly optimized floating point -> string rendering implementation, rather than just doing `std::to_string(number)`.

Turning floats into strings is fairly expensive, which is why it still shows up in a profiler run.  There are a lot of papers about number->string rendering performance, you can _possibly_ find a better implementation than the one already here, but the gains over what we have will likely be small.

SiarheiFedartsou: Yes, but as I can see we currently use `milo` algorithm(did you modify it in some way btw?) and `fmtlib` author [says their algorithm is even faster](https://github.com/fmtlib/dtoa-benchmark).
Screenshot 2022-09-30 at 18 33 04

It seemed to be low-hanging fruit to use it here(we already depend on fmtlib anyway), but 100% agree - no reason to waste time on it since possible improvement is marginal.

SiarheiFedartsou: It seems we have UB in current implementation 🤔 
Screenshot 2022-09-30 at 18 36 56

https://github.com/Project-OSRM/osrm-backend/actions/runs/3160140786/jobs/5144255047

It seems we had no UBSAN alerts here before, because for `std::vector` `ArrayRenderer` was used, which used `std::stringstream` for float numbers rendering.

SiarheiFedartsou: Checked on separate PR that `Renderer` has this UB problem https://github.com/Project-OSRM/osrm-backend/pull/6381
So it seems to be another reason to switch to `fmt`...

SiarheiFedartsou: # 5.

With `fmt` is feels as it is slightly faster(but hard to notice the difference without statistical tests - I am judging just by presence of < 10ms measurements) and no UBs at the same time :)

Screenshot 2022-10-01 at 21 35 10

features/testbot/annotations.feature
```diff
@@ -115,4 +115,4 @@ Feature: Annotations
 
         When I route I should get
             | from | to | route    | a:speed | a:distance            | a:duration | a:nodes |
-            | a    | c  | abc,abc  | 10:10   | 249.987619:299.962882 | 25:30      | 1:2:3   |
+            | a    | c  | abc,abc  | 10:10   | 249.987618946:299.962882039 | 25:30      | 1:2:3   |
```
SiarheiFedartsou, line 118: Seems to be expected since we changed algorithm used for formatting.

include/util/json_renderer.hpp
```diff
@@ -4,191 +4,165 @@
...
-            }
-            ++pos;
-        }
-        while (pos < MAX_FLOAT_STRING_LENGTH && buffer[pos] != 0)
+        // `fmt::memory_buffer` stores first 500 bytes in the object itself(i.e. on stack in this
+        // case) and then grows using heap if needed
+        fmt::memory_buffer buffer;
+        fmt::format_to(std::back_inserter(buffer), FMT_COMPILE("{}"), number.value);
+
+        // Truncate to 10 decimal places
```
SiarheiFedartsou, line 59: Tbh I do not understand why do we truncate it - I couldn't find nothing about it in JSON spec. Just to save some bytes in corner cases?

include/nodejs/json_v8_renderer.hpp
```diff
@@ -60,8 +60,8 @@ struct V8Renderer
 
 inline void renderToV8(v8::Local<v8::Value> &out, const osrm::json::Object &object)
 {
-    osrm::json::Value value = object;
```
SiarheiFedartsou, line 63: This copy is expensive...

include/util/json_renderer.hpp
```diff
@@ -4,191 +4,164 @@
...
-    void operator()(const String &string) const
+    void operator()(const String &string)
     {
-        out << "\"";
-        out << escape_JSON(string.value);
-        out << "\"";
+        write('"');
+        // here we assume that vast majority of strings don't need to be escaped,
+        // so we check it first and escape only if needed
+        auto size = SizeOfEscapedJSONString(string.value);
```
mjjbell, line 35: Can we instead just search for the existence of a character that requires escaping?

mjjbell: Do we have a feel for how often OSRM responses do require escaping? I.e how representative are these benchmarks?

include/util/json_renderer.hpp
mjjbell: I'd assume it's because none of the API responses can attribute anything meaningful to more decimal places, so it saves response payload space + improves readability.

include/util/json_renderer.hpp
SiarheiFedartsou: > Can we instead just search for the existence of a character that requires escaping?

We can, but current implementation allows to preallocate string for escaped string if we need it(because we know required size). ANd also assuming that vast majority of string don’t need to be escaped(I.e. don’t contain special characters) we will have to scan the whole string most of the times anyway.

> Do we have a feel for how often OSRM responses do require escaping?

That’s the great question. I don’t have numbers, just a gut feeling… Will try to collect some statistics from real responses.

Btw this idea is inspired by implementation in https://github.com/nlohmann/json

SiarheiFedartsou: Actually probably should be a good idea to take huge real JSON response, convert it to our json:: objects using some script and try to check some of optimizations on it.

SiarheiFedartsou: Well, actually I found 2 "sources" of escaped strings in responses.

1. Geometries(tbh I didn't know we can have slashes there and they appear quite often...)
Screenshot 2022-10-02 at 22 37 42

But even though they appear quite often these strings are quite big usually and it should allow us to preallocate string buffer for them(but we have to scan string twice)...

2. Road names and pronunciations. I quickly hacked osrm-extract to count number of `name` and `name:pronunciation` tags which need to be escaped in osm.pbf and got the following numbers for Croatia and Poland respectively:
Screenshot 2022-10-02 at 22 10 58
Screenshot 2022-10-02 at 22 37 02

After that I updated benchmark to use real output of OSRM(huge route from Portugal to Korea) and "played" a bit with this optimisation. And my conclusion is that it seems this optimisation is still useful on real JSONs, but improvement is barely noticeable.

### Without this optimisation (i.e. just always call `EscapeJSONString`)
Screenshot 2022-10-02 at 22 38 44

### With optimisation
Screenshot 2022-10-02 at 22 38 23

So I don't have strong opinion here: we can remove it(less code == more readable code) or leave it as is since it seems to make things slightly better.

WDYT?

SiarheiFedartsou: > Can we instead just search for the existence of a character that requires escaping?

Decided to check this variant(i.e. without preallocation if string requires escaping) and it surprisingly even more better on this particular benchmark.
Screenshot 2022-10-02 at 23 10 59

SiarheiFedartsou: And I decided to try the following trick:

```
        if (RequiresJSONStringEscaping(string.value))
        {
            std::string escaped;
            // just a guess that 16 bytes for escaped characters will be enough to avoid
            // reallocations
            escaped.reserve(string.value.size() + 16);
            EscapeJSONString(string.value, escaped);

            write(escaped);
        }
        else
        {
            write(string.value);
        }
```

And it seems to be even better:
Screenshot 2022-10-02 at 23 18 17

include/util/json_renderer.hpp
```diff
@@ -4,191 +4,165 @@
...
-    }
-
-    void operator()(const Number &number) const
-    {
-        char buffer[MAX_FLOAT_STRING_LENGTH] = {'\0'};
-        ieee754::dtoa_milo(number.value, buffer);
+        write('"');
+        // here we assume that vast majority of strings don't need to be escaped,
+        // so we check it first and escape only if needed
+        if (RequiresJSONStringEscaping(string.value))
```
SiarheiFedartsou, line 35: We probably could return index from this function and start `EscapeJSONString` from it, but IMO it is overkill.

include/util/json_renderer.hpp
mjjbell: The `String` performance is worse in the final one?
In any case, the deltas seem quite minor and the optimisation code is relatively small, so we can go with whichever is best.

mjjbell (approved):

include/util/json_renderer.hpp
SiarheiFedartsou: Hm, interesting, tbh I was paying attention to Vector only. Will try to re-check that and will make a final decision. 👍🏻

SiarheiFedartsou: After re-running it several times with and without this preallocation it is really hard to say where `String` is better, so I decided to not waste time on it and just leave existing implementation.

PR #6388, opened 2022-10-03
add conanbuildinfo.json

nilsnolde:
Add a `conanbuildinfo.json` so it's easy for downstream projects to read and find OSRM's dependencies when `ENABLE_CONAN=ON`. I'm finally attempting releasing OSRM as a Python package (mainly for a QGIS plugin), this would be help a lot.

## Tasklist

 - [ ] review


CMakeLists.txt
```diff
@@ -500,7 +500,7 @@ if(ENABLE_CONAN)
     "lua/${CONAN_LUA_VERSION}"
     "onetbb/${CONAN_TBB_VERSION}"
     BASIC_SETUP 
-    GENERATORS cmake_find_package
+    GENERATORS cmake_find_package json
```
SiarheiFedartsou, line 503: It would just generate it in build folder, right? May be add brief comment why do we need this `json`? I mean someone in the future may just delete it because it is not used in this repo itself.

SiarheiFedartsou (approved):

CMakeLists.txt
nilsnolde: > It will just generate it in build folder, right?

Right. Added a comment why it's useful in the source.

SiarheiFedartsou:
Need to figure out why this `[conan-macos-x64-release-node-lts](https://github.com/Project-OSRM/osrm-backend/actions/runs/3180970542/jobs/5185204355#logs)` is flaky(it seems we have it only on this job)

PR #6387, opened 2022-10-03
Get rid of Mason

SiarheiFedartsou:
# Issue

closes https://github.com/Project-OSRM/osrm-backend/issues/6266

## Tasklist

 - [ ] CHANGELOG.md entry ([How to write a changelog entry](http://keepachangelog.com/en/1.0.0/#how))
 - [ ] update relevant [Wiki pages](https://github.com/Project-OSRM/osrm-backend/wiki)
 - [ ] add tests (see [testing documentation](https://github.com/Project-OSRM/osrm-backend/blob/master/docs/testing.md))
 - [ ] review
 - [ ] adjust for comments
 - [ ] cherry pick to release branch

## Requirements / Relations

 Link any requirements here. Other pull requests this PR is based on?

PR #6391, opened 2022-10-04
Use clang-format from CI base image

SiarheiFedartsou:
# Issue

As we are [getting rid of Mason](https://github.com/Project-OSRM/osrm-backend/pull/6387) we need a way to use clang-format on CI(we download it from Mason at the moment). Possible options can be:
1. Download pre-built binary from somewhere else (currently implemented in https://github.com/Project-OSRM/osrm-backend/pull/6387 and downloads it from https://github.com/muttleyxd/clang-tools-static-binaries which doesn't seem very reliable solution on my taste)
2. Build own binary from source? It is quite complicated.
3. Use pre-installed binary from CI image. Currently available on our 20.04 image is:
```
Clang-format 10.0.0, 11.0.0, 12.0.0
```
Btw 11.0.0 is not available on modern 22.04 and we will have to update in the future.

I decided to go with option 3 and just update CI and source code to use clang-format from base image(but 1 is also way to go, don't have strong opinion, just think that using it from the base image feels more reliable). 

## Tasklist

 - [ ] review
 - [ ] adjust for comments

## Requirements / Relations

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -67,6 +67,7 @@ Thumbs.db
 /.vs*
 /*.local.bat
 /CMakeSettings.json
+/.cache
 
 # Jetbrains related files #
 ###########################
--- a/.github/workflows/osrm-backend.yml
+++ b/.github/workflows/osrm-backend.yml
@@ -88,9 +88,9 @@ jobs:
     - name: Prepare environment
       run: |
         npm ci --ignore-scripts
-        export MASON=${GITHUB_WORKSPACE}/scripts/mason.sh
-        ${MASON} install clang-format 10.0.0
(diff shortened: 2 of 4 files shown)


.gitignore
```diff
@@ -67,6 +67,7 @@ Thumbs.db
 /.vs*
 /*.local.bat
 /CMakeSettings.json
+/.cache
```
SiarheiFedartsou, line 70: It is from VSCode plugin I've been using, I think it is okay to have it here, but not in separate PR(but let me know if it is not)

mjjbell (approved):
Makes sense. The format script is still enforcing the version, so if the base image version changes then we'll need to make a decision (update the script version vs alternative installation source).

PR #6399, opened 2022-10-11
Always use original weights when handling --parse-conditionals-from-now

SiarheiFedartsou:
# Issue

## Tasklist

 - [ ] review
 - [ ] adjust for comments

## Requirements / Relations

### DIFF
--- a/features/car/conditional_restrictions.feature
+++ b/features/car/conditional_restrictions.feature
@@ -353,6 +353,18 @@ Feature: Car - Turn restrictions
             | b    | a  | bj,jc,jc,aj,aj |
             | b    | d  | bj,jd,jd       |
 
+        # here we check that conditional restrictions can be updated via re-run of `contract`/`customize` with updated `--parse-conditionals-from-now`
+        # 10am utc, sat 
+        When I run "osrm-contract  {osm_file} --time-zone-file=test/data/tz/{timezone_names}/guinea.geojson --parse-conditionals-from-now=1494064800"
+        When I run "osrm-customize {osm_file} --time-zone-file=test/data/tz/{timezone_names}/guinea.geojson --parse-conditionals-from-now=1494064800"
+
+        When I route I should get
+            | from | to | route          |
+            | b    | c  | bj,jc,jc       |
+            | b    | a  | bj,aj,aj       |
+            | b    | d  | bj,jd,jd       |
+
+
     @no_turning @conditionals
     Scenario: Car - Conditional restriction with multiple time windows
         Given the extract extra arguments "--parse-conditional-restrictions"
--- a/src/updater/updater.cpp
+++ b/src/updater/updater.cpp
@@ -583,8 +583,15 @@ Updater::LoadAndUpdateEdgeExpandedGraph(std::vector<extractor::EdgeBasedEdge> &e
             [&] { extractor::files::readNodeData(config.GetPath(".osrm.ebg_nodes"), node_data); },
 
             [&] {
-                extractor::files::readTurnWeightPenalty(
-                    config.GetPath(".osrm.turn_weight_penalties"), turn_weight_penalties);
+                // we are going to overwrite this `.turn_weight_penalties` afterwards,
+                // so here we backup the original turn penalties if we didn't do that yet in order to
+                // guarantee that subsequent runs of this code will work on top of original weights
+                auto path = config.GetPath(".osrm.turn_weight_penalties").string() + ".original";
+                if (!boost::filesystem::exists(path))
+                {
+                    boost::filesystem::copy_file(config.GetPath(".osrm.turn_weight_penalties"), path);
+                }
+                extractor::files::readTurnWeightPenalty(path, turn_weight_penalties);
             },
             [&] {
                 extractor::files::readTurnDurationPenalty(


src/updater/updater.cpp
```diff
@@ -583,8 +583,15 @@ Updater::LoadAndUpdateEdgeExpandedGraph(std::vector<extractor::EdgeBasedEdge> &e
             [&] { extractor::files::readNodeData(config.GetPath(".osrm.ebg_nodes"), node_data); },
 
             [&] {
-                extractor::files::readTurnWeightPenalty(
-                    config.GetPath(".osrm.turn_weight_penalties"), turn_weight_penalties);
+                // we are going to overwrite this `.turn_weight_penalties` afterwards,
```
SiarheiFedartsou, line 586: This code modifies a couple of other files too and probably we should use similar approach for them too, but it is not so obvious as here: IIUC other files are related to traffic handling, i.e. it is more obvious for users that they need to provide data for all edges while updating speeds(but tbh I just had brief look into that code and not confident in it :) ).

SiarheiFedartsou: On another hand it seems to be good idea to preserve "default" speed weights for segments user didn't provide data for 🤔 So would appreciate opinions here.

src/updater/updater.cpp
```diff
@@ -583,8 +583,15 @@ Updater::LoadAndUpdateEdgeExpandedGraph(std::vector<extractor::EdgeBasedEdge> &e
...
+                // so here we backup the original turn penalties if we didn't do that yet in order to
+                // guarantee that subsequent runs of this code will work on top of original weights
+                auto path = config.GetPath(".osrm.turn_weight_penalties").string() + ".original";
```
SiarheiFedartsou, line 590: Also this problem made me think about if we can handle all these time based restrictions automatically without forcing users to make error-prone operations to handle this. It feels it should be relatively not so many time based restrictions in the graph and we probably can just handle it dynamically. It seems to be a bit against philosophy of OSRM when it tries to precalculate all weights to be as fast as possible, but at the same time it should add a lot of flexibility:
- time based restrictions will "just work"  without additional calls with `--parse-conditionals-from-now`
- we potentially will be able to implement "depart at" feature

features/car/conditional_restrictions.feature
```diff
@@ -353,6 +353,18 @@ Feature: Car - Turn restrictions
             | b    | a  | bj,jc,jc,aj,aj |
             | b    | d  | bj,jd,jd       |
 
+        # here we check that conditional restrictions can be updated via re-run of `contract`/`customize` with updated `--parse-conditionals-from-now`
```
SiarheiFedartsou, line 356: Decided to just extend existing test, but not add new one.

features/car/conditional_restrictions.feature
```diff
@@ -353,6 +353,18 @@ Feature: Car - Turn restrictions
...
+        # 10am utc, sat 
+        When I run "osrm-contract  {osm_file} --time-zone-file=test/data/tz/{timezone_names}/guinea.geojson --parse-conditionals-from-now=1494064800"
+        When I run "osrm-customize {osm_file} --time-zone-file=test/data/tz/{timezone_names}/guinea.geojson --parse-conditionals-from-now=1494064800"
```
SiarheiFedartsou, line 359: Also probably we should vendor geojson with time zones to the repo, not sure what can be problem here, but it would ease a lot usage of this feature for beginners.

src/updater/updater.cpp
SiarheiFedartsou: @mjjbell wdyt?

mjjbell: The current behaviour of overwriting the files on updates means that you can make multiple independent changes and they will all apply. 

This proposed change would alter that behaviour (only last update is applied), with the alternative being to apply the cumulative changes on every update, which will be slower. So I don't think we should go with this.

For someone who is updating turn weights via traffic updates, we can probably assume they have a model of the data for any time, so aren't too bothered about recovering the original weights.

For conditional penalties, I think what we need is an `else` case that can set the turn weight penalty back to the previous value.

https://github.com/Project-OSRM/osrm-backend/blob/2cf957148bc01a7dc31ba396b6c8c3c28117ae8d/src/updater/updater.cpp#L514-L521

Alternatively, rather then overwriting the weight value when the conditional is applied, we only set it in the local value used to update the edge weights? That way we aren't destroying the value used when the conditional doesn't apply.

https://github.com/Project-OSRM/osrm-backend/blob/2cf957148bc01a7dc31ba396b6c8c3c28117ae8d/src/updater/updater.cpp#L765

src/updater/updater.cpp
mjjbell: Can probably be discussed separately. There will be practical as well as philosophical issues.

features/car/conditional_restrictions.feature
TheMarex: For sure, if I remember correctly though there may be licensing issue. Not sure what happens if we bundle ODBL-derived files (country boundaries from OSM) with our docker images etc.

TheMarex:
I think an interesting hack here could be to to only set/reset the MSB of the turn_weight_penalty and limit the maximum turn weight accordingly. The checks would need to check if the MSB is set not `INVALID_TURN_WEIGHT`. I'll investigate how easy that would be.

TheMarex:
I looked into it an this is basically a non-starter. The main issue is that turn penalties are _signed_. This gives us only a maxium turn penalty of +-27min (or the rough equivalent in weight). This itself wouldn't be maybe too bad but what is actually happening is not that we completely disallow these turns, but it is more of a soft-penalty since we only add `INVALID_TURN_PENALTY` (0x7FFF = ~54min). By doing the "MSB" trick we would reduce this down to potentially only ~`0x4000` or `27min + original_turn_weight`. That means if taking this (forbidden) turn would save ~27min on the route, it would still happen.

That still sounds fairly uncommon but I can easily see this regressing working setups that simply reset the data before the update.

We could change this by making turn penalties unsigned (and honestly that would have been the right choice in retrospect), but doing so would be another breaking change.

I'll update the issue with information on a workaround and close this for now.

Issue #6400, opened 2022-10-12
mid-size table requests to osrm-routed lead to HTTP error 400 / bad request


lamaux:
# Issue

Starting with the newest release 5.27.0 osrm-routed fails with HTTP error status code `400 bad request` when sending mid-size table requests. The limit seems to be around 6800 characters or ~500 points.

# Steps to reproduce

Sending a table request with > 500 random points to osrm-routed (v5.27.0) produces the error. Works with earlier, unreleased versions of v5.27 and earlier releases.

[...]

PR #6403, opened 2022-10-12
Fix bug with large HTTP requests leading to Bad Request in osrm-routed.

SiarheiFedartsou:
# Issue

closes https://github.com/Project-OSRM/osrm-backend/issues/6400
## Tasklist

 - [ ] review
 - [ ] adjust for comments

## Requirements / Relations


features/step_definitions/requests.js
```diff
@@ -4,7 +4,7 @@ module.exports = function () {
     this.When(/^I request \/(.*)$/, (path, callback) => {
         this.reprocessAndLoadData((e) => {
             if (e) return callback(e);
```
SiarheiFedartsou, line 6: This was unused and probably contained obsolete assumptions because of it.

features/stress/large_request.feature
```diff
@@ -0,0 +1,39 @@
...
...725,47.89964;10.65969,49.30469;11.01374,48.55548;11.85932,47.67318;11.77681,48.82834;11.91272,48.76547;11.57922,49.2744;10.52352,47.58819;11.44351,49.98144;10.8438,48.29292;11.0613,49.23531;11.75374,47.77367;10.68599,48.88375;10.8381,48.89306;11.78983,48.57816;10.86193,47.86047;11.57646,48.46316;11.6558,48.6255;11.31754,48.56296;10.69917,47.52159;11.48864,47.93436;11.14443,49.34194;10.66421,48.15455;11.69012,49.71597;10.92805,49.39795;10.98026,48.62985;10.63252,49.95795;11.37468,48.32383;11.89213,49.41892;11.39476,48.36131;10.61696,47.97164;11.50761,48.21042;11.31284,48.29539;10.87814,48.67741;10.59776,49.1498;11.17827,48.8733;11.6018,49.45914;10.73621,49.5317;11.38794,49.0252;11.40146,49.55482;10.57116,48.86615;11.89119,48.7372;11.33492,47.66424;11.61472,49.9513;11.63065,48.67025;11.15378,48.6575;11.31485,47.94025;10.55891,47.48489;11.80567,49.04565;10.74453,49.99566;10.74512,47.61681;10.91615,48.93038;11.03106,49.41157;11.66623,48.85904;10.63791,47.50966;11.38506,47.85158;11.82981,48.99957;11.20183,48.28544;11.53728,49.93828;10.84794,49.48396;11.99693,48.55948;11.37453,48.06193;10.79992,48.12021;11.85343,48.03751?generate_hints=false&sources=0&destinations=all&skip_waypoints=true
```
SiarheiFedartsou, line 38: Request from original issue

src/server/connection.cpp
```diff
@@ -79,20 +86,23 @@ void Connection::handle_read(const boost::system::error_code &error, std::size_t
     }
 
     boost::beast::error_code ec;
-    http_request_parser->put(boost::asio::buffer(incoming_data_buffer, bytes_transferred), ec);
+    http_request_parser->put(boost::asio::buffer(incoming_data_buffer), ec);
```
SiarheiFedartsou, line 89: We used it wrong for the case when request size > 8192 too... 🤦🏻

SiarheiFedartsou: https://www.boost.org/doc/libs/develop/libs/beast/doc/html/beast/ref/boost__beast__http__parser/put.html

> This function attempts to incrementally parse the HTTP message data stored in the caller provided buffers. Upon success, a positive return value indicates that the parser made forward progress, consuming that number of bytes. In some cases there may be an insufficient number of octets in the input buffer in order to make forward progress. This is indicated by the code [error::need_more](https://www.boost.org/doc/libs/develop/libs/beast/doc/html/beast/ref/boost__beast__http__error.html). When this happens, the caller should place additional bytes into the buffer sequence and call [put](https://www.boost.org/doc/libs/develop/libs/beast/doc/html/beast/ref/boost__beast__http__parser/put.html) again. The error code [error::need_more](https://www.boost.org/doc/libs/develop/libs/beast/doc/html/beast/ref/boost__beast__http__error.html) is special. When this error is returned, a subsequent call to [put](https://www.boost.org/doc/libs/develop/libs/beast/doc/html/beast/ref/boost__beast__http__parser/put.html) may succeed if the buffers have been updated. Otherwise, upon error the parser may not be restarted.

SiarheiFedartsou:
@mjjbell don't see any problem with that - hope I didn't introduce new bugs there 🤞🏻

mjjbell:
I can only get it to accept ~5000 coordinates, before it errors with a `400 BadRequest - bad_target`

5000 coords is _probably_ enough for users, but it's not consistent with the change made here, so I think there's still something that needs investigating.

danpat:
@mjjbell how many bytes in that request?  Might give a hint as to where to look.

mjjbell:
Was just over 100KB, but it looks like it's [not just about the size](https://github.com/Project-OSRM/osrm-backend/issues/6400#issuecomment-1278556823).

PR #6411, opened 2022-10-15
Implement NodeJS based server fully replicating osrm-routed

SiarheiFedartsou:
# Issue
https://github.com/Project-OSRM/osrm-backend/issues/3832
## Tasklist

 - [ ] CHANGELOG.md entry ([How to write a changelog entry](http://keepachangelog.com/en/1.0.0/#how))
 - [ ] update relevant [Wiki pages](https://github.com/Project-OSRM/osrm-backend/wiki)
 - [ ] add tests (see [testing documentation](https://github.com/Project-OSRM/osrm-backend/blob/master/docs/testing.md))
 - [ ] review
 - [ ] adjust for comments
 - [ ] cherry pick to release branch

## Requirements / Relations

 Link any requirements here. Other pull requests this PR is based on?

PR #6415, opened 2022-10-18
Fix annotations=true handling in NodeJS bindings & libosrm

SiarheiFedartsou:
# Issue
Found this problem while working on https://github.com/Project-OSRM/osrm-backend/pull/6411 (it was actually caught by one of our tests when I started running them against NodeJS based server - one more reason to rewrite osrm-router in NodeJs :) ).

Now it works the same as in osrm-routed:
https://github.com/Project-OSRM/osrm-backend/blob/5e5f1f4add6d86b097569545b670de513ebc6b04/include/server/api/route_parameters_grammar.hpp#L94

Or in constructors of `RouteParameters`:
https://github.com/Project-OSRM/osrm-backend/blob/5e5f1f4add6d86b097569545b670de513ebc6b04/include/engine/api/route_parameters.hpp#L129

## Tasklist

 - [ ] review
 - [ ] adjust for comments

## Requirements / Relations

### DIFF
--- a/include/engine/api/route_api.hpp
+++ b/include/engine/api/route_api.hpp
@@ -778,7 +778,7 @@ class RouteAPI : public BaseAPI
                 util::json::Object annotation;
 
                 // AnnotationsType uses bit flags, & operator checks if a property is set
-                if (parameters.annotations_type & RouteParameters::AnnotationsType::Speed)
+                if (requested_annotations & RouteParameters::AnnotationsType::Speed)
                 {
                     double prev_speed = 0;
                     annotation.values["speed"] = GetAnnotations(
--- a/include/engine/api/route_api.hpp
+++ b/include/engine/api/route_api.hpp
@@ -439,7 +439,7 @@ class RouteAPI : public BaseAPI
     {
         // AnnotationsType uses bit flags, & operator checks if a property is set
         flatbuffers::Offset<flatbuffers::Vector<float>> speed;
-        if (parameters.annotations_type & RouteParameters::AnnotationsType::Speed)
+        if (requested_annotations & RouteParameters::AnnotationsType::Speed)
         {
             double prev_speed = 0;
             speed =
--- a/include/nodejs/node_osrm_support.hpp
+++ b/include/nodejs/node_osrm_support.hpp
@@ -799,6 +799,9 @@ inline bool parseCommonParameters(const v8::Local<v8::Object> &obj, ParamType &p
         if (annotations->IsBoolean())
         {
             params->annotations = Nan::To<bool>(annotations).FromJust();
+            params->annotations_type = params->annotations
+                        ? osrm::RouteParameters::AnnotationsType::All
+                        : osrm::RouteParameters::AnnotationsType::None;
         }
         else if (annotations->IsArray())
         {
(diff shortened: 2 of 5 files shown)


test/nodejs/route.js
```diff
@@ -322,6 +322,12 @@ test('route: routes Monaco with options', function(assert) {
         assert.ok(first.routes[0].legs[0]);
         assert.ok(first.routes[0].legs.every(l => { return l.steps.length > 0; }), 'every leg has steps');
         assert.ok(first.routes[0].legs.every(l => { return l.annotation;}), 'every leg has annotations');
+        assert.ok(first.routes[0].legs.every(l => { return l.annotation.distance;}), 'every leg has annotations for distance');
+        assert.ok(first.routes[0].legs.every(l => { return l.annotation.duration;}), 'every leg has annotations for durations');
+        assert.ok(first.routes[0].legs.every(l => { return l.annotation.nodes;}), 'every leg has annotations for nodes');
+        assert.ok(first.routes[0].legs.every(l => { return l.annotation.weight; }), 'every leg has annotations for weight');
+        assert.ok(first.routes[0].legs.every(l => { return l.annotation.datasources; }), 'every leg has annotations for datasources');
+        assert.ok(first.routes[0].legs.every(l => { return l.annotation.speed; }), 'every leg has annotations for speed');
```
SiarheiFedartsou, line 330: Btw problem was happening only for `speed`, because we have these checks here which it seems were added for backward compatibility, but I am not sure if we should use `parameters.annotations_type`(but not `requested_annotations`) when checking if speed annotations were requested.
https://github.com/Project-OSRM/osrm-backend/blob/5e5f1f4add6d86b097569545b670de513ebc6b04/include/engine/api/route_api.hpp#L770
Screenshot 2022-10-18 at 18 35 51

include/nodejs/node_osrm_support.hpp
```diff
@@ -845,6 +848,9 @@ inline bool parseCommonParameters(const v8::Local<v8::Object> &obj, ParamType &p
                     Nan::ThrowError("this 'annotations' param is not supported");
                     return false;
                 }
+
+                params->annotations =
```
SiarheiFedartsou, line 852: I would remove this `annotations` boolean field completely, but it seems it would break libosrm backward compatibility

include/engine/api/route_api.hpp
```diff
@@ -439,7 +439,7 @@ class RouteAPI : public BaseAPI
     {
         // AnnotationsType uses bit flags, & operator checks if a property is set
         flatbuffers::Offset<flatbuffers::Vector<float>> speed;
-        if (parameters.annotations_type & RouteParameters::AnnotationsType::Speed)
+        if (requested_annotations & RouteParameters::AnnotationsType::Speed)
```
SiarheiFedartsou, line 442: We can leave it "as is" and NodeJs tests will still pass, but this seems to fix the problem for libosrm(when `annotations = true` was used).

See:
https://github.com/Project-OSRM/osrm-backend/pull/6415/files#r998472037

SiarheiFedartsou:
Fix for CI issues https://github.com/Project-OSRM/osrm-backend/pull/6416

test/nodejs/route.js
mjjbell: Yep, looks like it.

include/nodejs/node_osrm_support.hpp
mjjbell: We should think about all the breaking changes we would make for a v6 release.

mjjbell (approved):

include/nodejs/node_osrm_support.hpp
mjjbell: Heh, it's already there: https://github.com/Project-OSRM/osrm-backend/issues/3644

PR #6426, opened 2022-10-27
Take stop signs into account during routing

SiarheiFedartsou:
# Issue

## Tasklist

 - [ ] review
 - [ ] adjust for comments

## Requirements / Relations


include/extractor/traffic_signals.hpp
```diff
@@ -22,6 +23,14 @@ struct TrafficSignals
         return bidirectional_nodes.count(to) > 0 || unidirectional_segments.count({from, to}) > 0;
     }
 };
+
+struct TrafficSignals final : public RoadObjects {};
+
+// TODO: better naming ?
```
SiarheiFedartsou, line 29: It feels like we can try to generalise the whole this logic to any kind of objects which follows the same "idea" as traffic lights, i.e. some nodes which can be directed and give additional penalty in routing.

features/car/stop_sign_penalties.feature
```diff
@@ -0,0 +1,39 @@
+@routing @car @stop_sign
```
SiarheiFedartsou, line 1: Hi @mjjbell !

May I ask you to take a look here? 
What I actually did is took your existing traffic lights logic and completely repeated it for stop signs. There is a lot of things to be done here(at least we need to get rid of copy-paste), but before starting doing it, would be great to have your opinion here. Don't you see any fundamental problems in this approach? I am not very familiar with OSRM internals yet(one of the goals of this PR is to educate myself :) ) and tbh some parts of PR were "blindly" copy-pasted from traffic lights, so asking.

include/extractor/traffic_signals.hpp
mjjbell: Maybe `ControlFlow`?

If we view them as node penalties applicable by traversal direction, then I can see them being part of the same logical grouping.

features/car/stop_sign_penalties.feature
mjjbell: Yes if we're viewing all of them as node penalties applicable by direction, then it's basically the same code.

If we wanted to model the impact of each sign/traffic light more accurately (e.g. junction size, unprotected turns, all the other stuff in #2652, etc) then we'd probably want to handle them differently, but I don't think we're at that stage yet.

include/extractor/traffic_lights.hpp
```diff
@@ -19,6 +20,36 @@ enum Direction
 };
 } // namespace TrafficLightClass
 
+// Stop Signs tagged on nodes can be present or not. In addition Stop Signs have
+// an optional way direction they apply to. If the direction is unknown from the
+// data we have to compute by checking the distance to the next intersection.
```
mjjbell, line 25: Hmm, looks like identifying the direction will require something additional to the traffic signal logic.

profiles/car.lua
```diff
@@ -472,8 +480,12 @@ function process_turn(profile, turn)
 
   if turn.has_traffic_light then
       turn.duration = profile.properties.traffic_light_penalty
+  elseif turn.has_stop_sign then
+    -- TODO: use another constant
```
mjjbell, line 484: :+1:

src/extractor/extraction_containers.cpp
```diff
@@ -979,6 +982,54 @@ void ExtractionContainers::PrepareTrafficSignals(
     log << "ok, after " << TIMER_SEC(prepare_traffic_signals) << "s";
 }
 
+// TODO: copy-paste
+void ExtractionContainers::PrepareStopSigns(const ReferencedStopSigns &referenced_stop_signs) {
```
mjjbell, line 986: Can probably be combined with the traffic signal functions.

include/extractor/traffic_lights.hpp
SiarheiFedartsou: > If the direction is unknown from the data we have to compute by checking the distance to the next intersection.

Do you mean this sentence? Tbh it was a copy-paste from some of older PRs and I noticed it only now...

src/extractor/extraction_containers.cpp
```diff
@@ -1159,6 +1210,95 @@ ExtractionContainers::ReferencedWays ExtractionContainers::IdentifyRestrictionWa
...
+    util::for_each_indexed(ways_list.cbegin(), ways_list.cend(), set_segments);
+
+    util::for_each_pair(
+        signal_segments, [](const auto pair_a, const auto pair_b) {
+            if (pair_a.first == pair_b.first)
+            {
+                // If a node is appearing multiple times in this map, then it's ambiguous.
+                // The node is an intersection and the traffic direction is being use for multiple
+                // ways. We can't be certain of the original intent. See:
+                // https://wiki.openstreetmap.org/wiki/Key:traffic_signals:direction
```
mjjbell, line 1285: Might not be ambiguous for stop signs, would have to see what the OSM consensus is.

src/extractor/extraction_containers.cpp
SiarheiFedartsou: Sure, I am going to refactor it a lot, just want to have preliminary "green light" about approach.

src/extractor/graph_compressor.cpp
```diff
@@ -212,11 +213,15 @@ void GraphCompressor::Compress(const std::unordered_set<NodeID> &barrier_nodes,
                 const bool has_forward_signal = traffic_signals.HasSignal(node_u, node_v);
                 const bool has_reverse_signal = traffic_signals.HasSignal(node_w, node_v);
 
+                const bool has_forward_stop_sign = stop_signs.HasSignal(node_u, node_v);
+                const bool has_reverse_stop_sign = stop_signs.HasSignal(node_w, node_v);
+                
+                // TODO: can we have a case when we have both traffic signal and stop sign? how should we handle it?
```
mjjbell, line 219: Anything is possible in OSM...

include/extractor/traffic_lights.hpp
mjjbell: Yep.

src/extractor/extraction_containers.cpp
mjjbell: Yep, looks good :green_circle:

include/extractor/traffic_lights.hpp
SiarheiFedartsou: It looks like ~77% of stop signs tagged with direction.

I will take a look what it will take to compute this "distance to the nearest intersection" - should be just a couple of heuristics, but not sure how to do that from code point of view, but if it is TOO difficult to be part of this PR I would probably propose to just ignore stop signs without direction for the time being and then return to it in separate PRs.

https://taginfo.openstreetmap.org/tags/highway=stop#combinations
Screenshot 2022-10-28 at 16 39 59

mjjbell:
Perhaps the other thing to consider is the navigation guidance. I didn't really think about it in #6153 as traffic signals already existed. Adding stop/give-way signs is something that the guidance code should be aware of.

src/extractor/scripting_environment_lua.cpp
```diff
@@ -284,17 +284,26 @@ void Sol2ScriptingEnvironment::InitContext(LuaScriptingContext &context)
         [&context, &get_location_tag](const osmium::Node &node, const char *key) {
             return get_location_tag(context, node.location(), key);
         });
-
+    // just for backward compatibility
+    // TODO: remove in v6
     context.state.new_enum("traffic_lights",
```
SiarheiFedartsou, line 289: I propose to use common `traffic_flow_control_direction` enum instead of existing `traffic_lights`

features/car/give_way_sign_penalties.feature
```diff
@@ -0,0 +1,77 @@
+@routing @car @give_way_sign
+Feature: Car - Handle give way signs
+
+    Background:
+        Given the profile "car"
+
+    Scenario: Car - Encounters a give way sign
```
SiarheiFedartsou, line 7: It is almost copy-paste of existing tests for traffic lights https://github.com/Project-OSRM/osrm-backend/blob/afe34456b9993d465d37ea3898fee1db6a2b5561/features/car/traffic_light_penalties.feature#L7
Probably worth having here all tests we have for traffic lights, but is there way to reduce copy-paste here somehow? E.g. somehow run the same test, but with different `highway` tag values for each run?

include/extractor/traffic_flow_control_nodes.hpp
```diff
@@ -11,17 +11,29 @@ namespace osrm
...
+enum class TrafficFlowControlNodeDirection : std::uint8_t
+{
+    NONE = 0,
+    ALL = 1,
+    FORWARD = 2,
+    REVERSE = 3
+};
+
+// represents traffic lights, stop signs, give way signs, etc.
+struct TrafficFlowControlNodes
```
SiarheiFedartsou, line 25: Not sure about naming here, if there are better ideas will be happy to discuss.

profiles/car.lua
```diff
@@ -28,6 +30,8 @@ function setup()
       use_turn_restrictions          = true,
       left_hand_driving              = false,
       traffic_light_penalty          = 2,
+      stop_sign_penalty              = 2,
```
SiarheiFedartsou, line 33: Just from own driving experience: stop signs are quite annoying and often you have to stay as much as on traffic lights, but "give ways" are usually more relaxed

profiles/car.lua
```diff
@@ -471,9 +481,14 @@ function process_turn(profile, turn)
   local turn_bias = turn.is_left_hand_driving and 1. / profile.turn_bias or profile.turn_bias
 
   if turn.has_traffic_light then
```
SiarheiFedartsou, line 483: Initially I had a question "what to do if node has traffic light and stop sign at the same time?", IMO the answer is quite simple: here we just apply one of penalties, but end users may go further and apply combination of penalties in this case if they want or change their priority.

src/extractor/extractor_callbacks.cpp
```diff
@@ -78,10 +78,23 @@ void ExtractorCallbacks::ProcessNode(const osmium::Node &input_node,
     {
         external_memory.barrier_nodes.push_back(id);
     }
-    if (result_node.traffic_lights != TrafficLightClass::NONE)
+    if (result_node.traffic_lights != TrafficFlowControlNodeDirection::NONE)
     {
         external_memory.external_traffic_signals.push_back({id, result_node.traffic_lights});
     }
+    // TODO: we ignore `ALL` for both stop signs and give way signs, because we cannot understand
```
SiarheiFedartsou, line 85: I have an idea how to implement it(at least for vast majority of cases), but I'd propose to do in separate PR.

include/extractor/traffic_flow_control_nodes.hpp
SiarheiFedartsou: Can we try to re-use the same mechanism in the future for example for railway crossings? https://wiki.openstreetmap.org/wiki/Tag:railway%3Dlevel_crossing

SiarheiFedartsou:
@mjjbell @DennisOSRM can we may be make second round of review here?

DennisOSRM (approved):

features/car/give_way_sign_penalties.feature
mjjbell: I expect there's some support for parametrization of tests, but we're on an ancient Cucumber version, so might take a lot of effort. Plus, having a separate file per penalty feature will allow the functionality to diverge if required.

include/extractor/extraction_node.hpp
```diff
@@ -1,7 +1,8 @@
 #ifndef EXTRACTION_NODE_HPP
 #define EXTRACTION_NODE_HPP
 
-#include "traffic_lights.hpp"
+#include "traffic_flow_control_nodes.hpp"
+#include <cstdint>
```
mjjbell, line 5: Needed?

profiles/car.lua
mjjbell: I think applying the max penalty out of the three would make sense to me. This logic achieves that too.

src/extractor/extraction_containers.cpp
```diff
@@ -1224,7 +1232,7 @@ ExtractionContainers::ReferencedTrafficSignals ExtractionContainers::IdentifyTra
     util::for_each_indexed(ways_list.cbegin(), ways_list.cend(), set_segments);
 
     util::for_each_pair(
-        signal_segments, [](const auto pair_a, const auto pair_b) {
+        node_segments, [](const auto pair_a, const auto pair_b) {
```
mjjbell, line 1235: Minor: Warning message below could be updated to refer to traffic control flow nodes.

src/extractor/extractor_callbacks.cpp
mjjbell: Why can't we understand all-way stop/give-ways?

Can we treat them the same as traffic lights without direction?

mjjbell (approved):
LGTM

kaligrafy:
Can someone verify why it has been approved but not merged already?

DennisOSRM:
Implemented by #7130

PR #6435, opened 2022-10-30
Delete CloudFormation templates

SiarheiFedartsou:
There is no need in them - it seems they were needed to manage some Mapbox S3 bucket which we have no access to anyway.


DennisOSRM (approved):
Good riddance

PR #6437, opened 2022-11-01
Fix `npm audit` warnings in NPM package

SiarheiFedartsou:
# Issue

At the moment if I run `npm audit` in project with installed `@project-osrm/osrm` it complains about a couple of warnings coming from our package:
Screenshot 2022-11-01 at 21 24 32

The reason of this is dependencies we have in package which actually can be made dev-dependencies as they are only needed for build/test purposes.

## Tasklist

 - [ ] review
 - [ ] adjust for comments

## Requirements / Relations

### DIFF
--- a/.github/workflows/osrm-backend.yml
+++ b/.github/workflows/osrm-backend.yml
@@ -95,6 +95,7 @@ jobs:
         ./scripts/format.sh && ./scripts/error_on_dirty.sh
         node ./scripts/validate_changelog.js
         npm run docs && ./scripts/error_on_dirty.sh
+        npm audit --production
 
   docker-image:
     needs: format-taginfo-docs
--- a/package.json
+++ b/package.json
@@ -4,12 +4,7 @@
   "private": false,
   "description": "The Open Source Routing Machine is a high performance routing engine written in C++14 designed to run on OpenStreetMap data.",
   "dependencies": {
-    "@mapbox/node-pre-gyp": "^1.0.9",
-    "cheap-ruler": "^3.0.2",
-    "mkdirp": "^0.5.6",
-    "nan": "^2.17.0",
-    "node-cmake": "^2.5.1",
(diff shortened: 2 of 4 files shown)


.github/workflows/osrm-backend.yml
```diff
@@ -95,6 +95,7 @@ jobs:
         ./scripts/format.sh && ./scripts/error_on_dirty.sh
         node ./scripts/validate_changelog.js
         npm run docs && ./scripts/error_on_dirty.sh
+        npm audit --production
```
SiarheiFedartsou, line 98: Will alert if we'll start having this problem again. Unfortunately it can be quite unexpected: e.g. if vulnerability will be found in version of `@mapbox/node-pre-gyp` we use and NPM will start reporting it.

package.json
```diff
@@ -4,12 +4,7 @@
   "private": false,
   "description": "The Open Source Routing Machine is a high performance routing engine written in C++14 designed to run on OpenStreetMap data.",
   "dependencies": {
-    "@mapbox/node-pre-gyp": "^1.0.9",
```
SiarheiFedartsou, line 7: `cheap-ruler`, `mkdirp` and `rimraf` are used in tests only
`nan` and `node-cmake` are needed to only build binary and not needed to use package after it

.github/workflows/osrm-backend.yml
SiarheiFedartsou: `--production` means "audit only non-dev dependencies". Our dev dependencies have a lot of problems at the moment and it is quite difficult to fix.

package.json
SiarheiFedartsou: Tbh it would be great to somehow check that package will 100% work without them on some clean test project, but not sure how to properly do that.

DennisOSRM (approved):
:+1:

PR #6444, opened 2022-11-06
[CPP17] Modernize: Replace typedef with using statements

DennisOSRM:
It's a mechanical change done by clang-tidy


SiarheiFedartsou (approved):

SiarheiFedartsou:
Btw we probably can enable https://clang.llvm.org/extra/clang-tidy/checks/modernize/use-using.html in https://github.com/Project-OSRM/osrm-backend/blob/fc12b6c3652612f177635d2d0240c9d3c7bb0006/.clang-tidy#L43

PR #6445, opened 2022-11-06
Drop GCC7 support

DennisOSRM:
Mininum compiler version required to compile Node 16+ is GCC 8.3 according to the [NodeJS release announcement](https://nodejs.org/ko/blog/release/v16.0.0/).

### DIFF
--- a/.github/workflows/osrm-backend.yml
+++ b/.github/workflows/osrm-backend.yml
@@ -264,15 +264,6 @@ jobs:
             CXXCOMPILER: g++-8
             CXXFLAGS: -Wno-cast-function-type
 
-          - name: gcc-7-release
-            continue-on-error: false
-            node: 16
-            runs-on: ubuntu-20.04
-            BUILD_TOOLS: ON
-            BUILD_TYPE: Release
-            CCOMPILER: gcc-7
-            CXXCOMPILER: g++-7
-
           - name: conan-macos-x64-release-node-16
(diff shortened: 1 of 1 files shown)


SiarheiFedartsou (approved):

.github/workflows/osrm-backend.yml
```diff
@@ -264,15 +264,6 @@ jobs:
             CXXCOMPILER: g++-8
             CXXFLAGS: -Wno-cast-function-type
 
-          - name: gcc-7-release
```
SiarheiFedartsou, line 267: We probably can add gcc-12 instead in separate PR

DennisOSRM: yes, good call with the separate PR. It seems the code doesn't compile cleanly with GCC12, yet. Limited testing started to throw a fair amount of compile errors.

SiarheiFedartsou:
> Mininum compiler version required to compile Node 16+ is GCC 8.3 according to the [NodeJS release announcement](https://nodejs.org/ko/blog/release/v16.0.0/).

Btw as I understand GCC 8 is minimum compiler version, but not required, i.e. nothing prevents us from using the very latest version of GCC to build bindings when we do it on our own(we actually use clang to do it btw), i.e. the only value of old compilers support is the fact that we enable other people to build OSRM(and probably Node bindings) with these compilers.

PR #6455, opened 2022-11-12
Add CI job which builds OSRM with gcc 12

SiarheiFedartsou:
# Issue

Adds CI job with Gcc 12.
## Tasklist

 - [ ] review
 - [ ] adjust for comments

## Requirements / Relations

### DIFF
--- a/.github/workflows/osrm-backend.yml
+++ b/.github/workflows/osrm-backend.yml
@@ -145,269 +145,279 @@ jobs:
     strategy:
       matrix:
         include:
-          - name: gcc-9-debug-cov
-            continue-on-error: false
-            node: 16
-            runs-on: ubuntu-20.04
-            BUILD_TOOLS: ON
-            BUILD_TYPE: Debug
-            CCOMPILER: gcc-9
-            CUCUMBER_TIMEOUT: 20000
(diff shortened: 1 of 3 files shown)


.github/workflows/osrm-backend.yml
```diff
@@ -212,6 +212,17 @@ jobs:
...
+          - name: gcc-12-release
+            continue-on-error: false
+            node: 16
+            runs-on: ubuntu-22.04
+            BUILD_TOOLS: ON
+            BUILD_TYPE: Release
+            CCOMPILER: gcc-12
+            CXXCOMPILER: g++-12
+            ENABLE_BENCHMARKS: ON
+            CXXFLAGS: '-Wno-array-bounds -Wno-uninitialized'
```
SiarheiFedartsou, line 224: Had to disable a couple of warnings as some of thirdparty dependencies(sol2 in particular) don't compile otherwise - probably we have some misconfiguration as in general warnings in third party deps shouldn't bother us, but for now it is easier option(at least for now we have some build with gcc 12 and afterwards we can try to understand what is wrong with sol2)

DennisOSRM (approved):
Love it

PR #6480, opened 2022-12-11
Nested namespace

DennisOSRM:
It's a mechanical change to modernize the code base


include/contractor/contract_excludable_graph.hpp
```diff
@@ -94,7 +92,6 @@ inline auto contractExcludableGraph(ContractorGraph contractor_graph_,
     return GraphAndFilter{QueryGraph{num_nodes, edge_container.edges},
                           edge_container.MakeEdgeFilters()};
 }
-} // namespace contractor
 } // namespace osrm
```
SiarheiFedartsou, line 95: nit. would be great to reflect this in such comments too(if possible to do in automated fashion ofc)

SiarheiFedartsou (approved):

include/contractor/contract_excludable_graph.hpp
DennisOSRM: fair point. I am afraid those are hard to do in an automated way, tho.

SiarheiFedartsou: Then I think it is not big problem if we keep it like this

DennisOSRM: I spoke too early. The latest push to this PR should fix the closing namespace annotation.

PR #6491, opened 2022-12-22
fix typo

SamuelBrucksch:
This bug was introduced 5 years ago in https://github.com/Project-OSRM/osrm-backend/commit/d166fc6a75503bbdc77f827b28967150e7dbebe9

guess it doesn't do anything important, if it has not been noticed, but still should be corrected.


SiarheiFedartsou:
Thanks!


Curious how could we avoid this? It seems Lua is quite tolerant to undefined variables and there some ways to avoid it http://lua-users.org/wiki/DetectingUndefinedVariables

SiarheiFedartsou (approved):

PR #6493, opened 2022-12-23
make it tolerant of TBB 2020

datendelphin:
I know I am too late to the party, as 5.27 is already released.
With regard to https://github.com/Project-OSRM/osrm-backend/pull/6300 it is not that difficult to make OSRM agnostic of tbb version. I implemented it for the fossgis routing server, as that one is on debian, which might take another few months until the next stable version is released.
So probably too late to be relevant, but if anyone is in a similar position, here would be the patch.

### DIFF
--- a/src/extractor/edge_based_graph_factory.cpp
+++ b/src/extractor/edge_based_graph_factory.cpp
@@ -32,7 +32,11 @@
 
 #include <tbb/blocked_range.h>
 #include <tbb/parallel_for.h>
+#if TBB_VERSION_MAJOR == 2020
+#include <tbb/pipeline.h>
+#else
 #include <tbb/parallel_pipeline.h>
+#endif
 
 namespace osrm
 {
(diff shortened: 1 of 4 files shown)


DennisOSRM:
@SiarheiFedartsou what's your opinion on this?

SiarheiFedartsou:
Well, on one hand I don’t see any issues with that - it is not a big deal to support both versions, but when do we think we can completely drop compatibility with tbb 2020? When Debian will officially start having newer versions in package repos?

btw at the moment we use Debian as base image too and just build newer TBB from sources https://github.com/Project-OSRM/osrm-backend/blob/95c45239873818ec5d982dd5f7bfb4f66930fec4/docker/Dockerfile#L10

datendelphin:
In my opinion, a good time to remove 2020 support would be in the next OSRM release after debian bookworm was released.
I have no problem with building a newer tbb from source. But it has drawbacks. The release coming with debian was tested with the rest of the system. And it's just more convenient. Less surprises for users (currently no clear error message that TBB is too old). If there is a compelling reason, like it is significantly more efficient, that would be another case.

SiarheiFedartsou:
> a good time to remove 2020 support would be in the next OSRM release after debian bookworm was released.

Google says Bookworm will likely be released in Mid 2023. Should we really spend time on support of both versions for 6 months? These days OSRM development is quite slow and I doubt we even will be able to release some new version of OSRM before release of new Debian. 

At the same time it is not a big deal to merge and maintain this and I can help with it if really needed - just sharing some doubts...

WDYT?

datendelphin:
Yes I feel the same way, that's why I wrote I am too late to the party ;)
Depends a bit on how often https://github.com/Project-OSRM/osrm-backend/issues/6497 pops up. Might be enough to just have this PR as a quick fix for people hitting it. Not sure about github behavior, but maybe I can just close it, but it will still be found via the bug report.

datendelphin:
Debian bookworm was released in the mean time

woodpeck:
I just tried to build OSRM on a stock RedHat 9.2 and the tbb coming with that was too old apparently, and this patch helped me to get this working.

Issue #6528, opened 2023-01-27
Following instructions on readme (docker), but curl returns empty result


wyzhao:
Hi, I follow the instructions on the Readme by starting a docker locally, but the following request returns empty result:

curl "http://127.0.0.1:5000/route/v1/driving/13.388860,52.517037;13.385983,52.496891?steps=true"

I am using Mac.

danpat:
@wyzhao Can you try the commands listed on the Docker wiki page:

https://github.com/Project-OSRM/osrm-backend/wiki/Docker-Recipes

those include some error checking to ensure that each step completed successfully.  If you're getting an empty result, then one of the earlier processing steps probably failed.  It is important that you confirm that each step completes properly, and the README doesn't exactly describe how to do that in precise detail.

wyzhao:
It works now.

Thank you.

Daniel Patterson ***@***.***> 于2023年1月26日周四 19:11写道：

PR #6529, opened 2023-01-28
Update Docker commands in README to also output error in case of failure

SiarheiFedartsou:
# Issue

People often get confused due to the fact that `osrm-extract` can just crash(e.g. due to OOM) and we output nothing in this case - they can have wrong impression that every step finished successfully, but it didn't in fact(example https://github.com/Project-OSRM/osrm-backend/issues/6528). So I propose to just include this `|| echo "... failed"` to make it more obvious. It is almost the same as we have in https://github.com/Project-OSRM/osrm-backend/wiki/Docker-Recipes, but a bit less verbose.

## Tasklist

 - [ ] review
 - [ ] adjust for comments

## Requirements / Relations

### DIFF
--- a/README.md
+++ b/README.md
@@ -58,12 +58,12 @@ Download OpenStreetMap extracts for example from [Geofabrik](http://download.geo
 
 Pre-process the extract with the car profile and start a routing engine HTTP server on port 5000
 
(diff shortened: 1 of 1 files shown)


danpat:
@SiarheiFedartsou thanks, I should've done this years ago, it's such a common problem....

danpat (approved):

Issue #6513, opened 2023-01-16
Why map matching returned confidence greater than 1 ?


Rejudge-F:
Hi team, When i use mm(map matching) to snip some gps trace i found some matching confidence greater than 1.f, I would like to know under what circumstances this result occurs. THX

SiarheiFedartsou:
Hey, could you give particular example? At least which exactly number greater than 1 you’ve seen.

Rejudge-F:
Sry, I just saw it once, haven't met recently, when I encounter this situation again, I will add it here.

THX

Rejudge-F:
Hi, I found I found it
URI:/match/v1/bick/116.186532,39.923911;116.186532,39.923911;116.186532,39.922552;116.186532,39.922552;116.186545,39.921992;116.186545,39.921992;116.186576,39.921185;116.186576,39.921185;116.186576,39.920503;116.186576,39.920503;116.186576,39.919327?steps=true&geometries=geojson&overview=full&annotations=true&tidy=true

Resposne:
```
{
  "code": "Ok",
  "matchings": [
    {
      "confidence": 1.711408792,
      "geometry": {
        "coordinates": [
          [
            116.186532,
            39.92386
          ],
          [
            116.18643,
            39.92386
          ],
          [
            116.18643,
            39.92351
          ],
          [
            116.18643,
            39.92284
          ],
          [
            116.18643,
            39.92255
          ],
          [
            116.18644,
            39.92218
          ],
          [
            116.186551,
            39.92207
          ],
          [
            116.18634,
            39.92208
          ],
          [
            116.18617,
            39.92208
          ],
          [
            116.18595,
            39.92208
          ],
          [
            116.18575,
            39.92208
          ],
          [
            116.18578,
            39.92108
          ],
          [
            116.18581,
            39.92048
          ],
          [
            116.18611,
            39.92048
          ],
          [
            116.18638,
            39.92048
          ],
          [
            116.18636,
            39.92116
          ],
          [
            116.18655,
            39.92116
          ],
          [
            116.186576,
            39.92116
          ],
          [
            116.18667,
            39.92116
          ],
          [
            116.18673,
            39.92116
          ],
          [
            116.18673,
            39.920503
          ],
          [
            116.18673,
            39.92049
          ],
          [
            116.18672,
            39.92042
          ],
          [
            116.18672,
            39.92033
          ],
          [
            116.18694,
            39.92029
          ],
          [
            116.18706,
            39.92025
[...]

PR #6531, opened 2023-01-30
fix(json_render): json_render is not accurate enough for extremely sm…

Rejudge-F:
fix(json_render): json_render is not accurate enough for extremely small numbers. eg: fmt::format_to(std::back_inserter(buffer), FMT_COMPILE("{}"), double(0.0000000000017114087924596788)); you will get a result of 1.7114087924, this is a completely wrong result.

formatter params(width & precision chapter): https://en.cppreference.com/w/cpp/utility/format/formatter 

# Issue

#6513 

## Tasklist

 - [ ] review
 - [ ] adjust for comments

## Requirements / Relations

### DIFF
--- a/features/car/conditional_restrictions.feature
+++ b/features/car/conditional_restrictions.feature
@@ -538,9 +538,9 @@ Feature: Car - Turn restrictions
 
         When I route I should get
             | from | to | route                                      | turns                                       |
-            | a    | e  | cap south,florida ne,florida ne | depart,turn left,arrive     |
+            | a    | e  | cap south,florida nw,florida nw,florida ne | depart,turn right,continue uturn,arrive     |
             | f    | d  | cap north,florida nw,florida nw            | depart,turn left,arrive                     |
-            | e    | c  | florida ne,cap south,cap south  | depart,turn left,arrive     |
+            | e    | c  | florida ne,florida nw,cap south,cap south  | depart,continue uturn,turn right,arrive     |
 
     @no_turning @conditionals
     Scenario: Car - DC North capitol situation, one on two off
--- a/features/car/conditional_restrictions.feature
+++ b/features/car/conditional_restrictions.feature
@@ -538,9 +538,9 @@ Feature: Car - Turn restrictions
 
         When I route I should get
             | from | to | route                                      | turns                                       |
-            | a    | e  | cap south,florida nw,florida nw,florida ne | depart,turn right,continue uturn,arrive     |
+            | a    | e  | cap south,florida ne,florida ne | depart,turn left,arrive     |
             | f    | d  | cap north,florida nw,florida nw            | depart,turn left,arrive                     |
-            | e    | c  | florida ne,florida nw,cap south,cap south  | depart,continue uturn,turn right,arrive     |
+            | e    | c  | florida ne,cap south,cap south  | depart,turn left,arrive     |
 
     @no_turning @conditionals
     Scenario: Car - DC North capitol situation, one on two off
@@ -586,7 +586,7 @@ Feature: Car - Turn restrictions
         When I route I should get
             | from | to | route                                      | turns                                         |
             | a    | e  | cap south,florida ne,florida ne            | depart,turn left,arrive                       |
-            | f    | d  | cap north,florida ne,florida ne,florida nw | depart,turn sharp right,continue uturn,arrive |
+            | f    | d  | cap north,florida nw,florida nw | depart,turn left,arrive |
             | e    | c  | florida ne,cap south,cap south             | depart,turn left,arrive                       |
 
     @only_turning @conditionals
--- a/features/car/conditional_restrictions.feature
+++ b/features/car/conditional_restrictions.feature
@@ -586,7 +586,7 @@ Feature: Car - Turn restrictions
         When I route I should get
             | from | to | route                                      | turns                                         |
             | a    | e  | cap south,florida ne,florida ne            | depart,turn left,arrive                       |
-            | f    | d  | cap north,florida nw,florida nw | depart,turn left,arrive |
+            | f    | d  | cap north,florida ne,florida ne,florida nw | depart,turn sharp right,continue uturn,arrive |
             | e    | c  | florida ne,cap south,cap south             | depart,turn left,arrive                       |
 
     @only_turning @conditionals
--- a/include/util/json_renderer.hpp
+++ b/include/util/json_renderer.hpp
@@ -47,20 +47,12 @@ template <typename Out> struct Renderer
 
     void operator()(const Number &number)
     {
+        // we don't want to print NaN or Infinity
+        BOOST_ASSERT(std::isfinite(number.value));
         // `fmt::memory_buffer` stores first 500 bytes in the object itself(i.e. on stack in this
         // case) and then grows using heap if needed
         fmt::memory_buffer buffer;
-        fmt::format_to(std::back_inserter(buffer), FMT_COMPILE("{}"), number.value);
-
-        // Truncate to 10 decimal places
-        size_t decimalpos = std::find(buffer.begin(), buffer.end(), '.') - buffer.begin();
-        if (buffer.size() > (decimalpos + 10))
-        {
-            buffer.clear();
-            fmt::format_to(std::back_inserter(buffer), FMT_COMPILE("{0:.10f}"), number.value);
-
-            // buffer.resize(decimalpos + 10);
-        }
+        fmt::format_to(std::back_inserter(buffer), FMT_COMPILE("{:.10g}"), number.value);
 
         write(buffer.data(), buffer.size());
     }
--- a/include/util/json_renderer.hpp
+++ b/include/util/json_renderer.hpp
@@ -56,7 +56,10 @@ template <typename Out> struct Renderer
         size_t decimalpos = std::find(buffer.begin(), buffer.end(), '.') - buffer.begin();
         if (buffer.size() > (decimalpos + 10))
         {
-            buffer.resize(decimalpos + 10);
+            buffer.clear();
+            fmt::format_to(std::back_inserter(buffer), FMT_COMPILE("{0:.10f}"), number.value);
+
+            // buffer.resize(decimalpos + 10);
         }
 
         write(buffer.data(), buffer.size());
--- a/unit_tests/util/json_render.cpp
+++ b/unit_tests/util/json_render.cpp
@@ -32,6 +32,62 @@ BOOST_AUTO_TEST_CASE(test_json_issue_6531) {
     osrm::util::json::Renderer<std::string> renderer(output);
     renderer(0.0000000000017114087924596788);
     BOOST_CHECK_EQUAL(output, "1.711408792e-12");
+
+    output.clear();
+    renderer(42.0);
+    BOOST_CHECK_EQUAL(output, "42");
+
+    output.clear();
+    renderer(42.1);
+    BOOST_CHECK_EQUAL(output, "42.1");
+
+    output.clear();
+    renderer(42.12);
+    BOOST_CHECK_EQUAL(output, "42.12");
+
+    output.clear();
+    renderer(42.123);
+    BOOST_CHECK_EQUAL(output, "42.123");
+
+    output.clear();
+    renderer(42.1234);
+    BOOST_CHECK_EQUAL(output, "42.1234");
+
+    output.clear();
+    renderer(42.12345);
+    BOOST_CHECK_EQUAL(output, "42.12345");
+
+    output.clear();
+    renderer(42.123456);
+    BOOST_CHECK_EQUAL(output, "42.123456");
+
+    output.clear();
+    renderer(42.1234567);
+    BOOST_CHECK_EQUAL(output, "42.1234567");
+
+    output.clear();
+    renderer(42.12345678);
+    BOOST_CHECK_EQUAL(output, "42.12345678");
+
+    output.clear();
+    renderer(42.123456789);
+    BOOST_CHECK_EQUAL(output, "42.12345679");
+
+    output.clear();
+    renderer(0.12345678912345);
+    BOOST_CHECK_EQUAL(output, "0.1234567891");
+
+    output.clear();
+    renderer(0.123456789);
+    BOOST_CHECK_EQUAL(output, "0.123456789");
+
+    output.clear();
+    renderer(0.12345678916);
+    BOOST_CHECK_EQUAL(output, "0.1234567892");
+
+    output.clear();
+    renderer(123456789123456789);
+    BOOST_CHECK_EQUAL(output, "1.234567891e+17");
 }
 
 BOOST_AUTO_TEST_SUITE_END()
--- a/unit_tests/util/json_render.cpp
+++ b/unit_tests/util/json_render.cpp
@@ -16,7 +16,7 @@ BOOST_AUTO_TEST_CASE(number_truncating)
 
     // this number would have more than 10 decimals if not truncated
     renderer(Number{42.9995999594999399299});
-    BOOST_CHECK_EQUAL(str, "42.999599959");
+    BOOST_CHECK_EQUAL(str, "42.99959996");
 }
 
 BOOST_AUTO_TEST_CASE(integer)
@@ -25,6 +25,13 @@ BOOST_AUTO_TEST_CASE(integer)
     Renderer<std::string> renderer(str);
     renderer(Number{42.0});
     BOOST_CHECK_EQUAL(str, "42");
+}
+
+BOOST_AUTO_TEST_CASE(test_json_issue_6531) {
+    std::string output;
+    osrm::util::json::Renderer<std::string> renderer(output);
+    renderer(0.0000000000017114087924596788);
+    BOOST_CHECK_EQUAL(output, "1.711408792e-12");
 }
 
 BOOST_AUTO_TEST_SUITE_END()
--- a/.github/workflows/osrm-backend.yml
+++ b/.github/workflows/osrm-backend.yml
@@ -376,9 +376,9 @@ jobs:
       uses: actions/cache@v3
       with:
         path: ${{github.workspace}}/test/cache
-        key: v3-test-${{ matrix.name }}-${{ github.sha }}
+        key: v4-test-${{ matrix.name }}-${{ github.sha }}
         restore-keys: |
-          v3-test-${{ matrix.name }}-
+          v4-test-${{ matrix.name }}-
 
     - name: Prepare environment
       run: |
--- a/features/testbot/annotations.feature
+++ b/features/testbot/annotations.feature
@@ -115,4 +115,4 @@ Feature: Annotations
 
         When I route I should get
             | from | to | route    | a:speed | a:distance            | a:duration | a:nodes |
-            | a    | c  | abc,abc  | 10:10   | 249.987618946:299.962882039 | 25:30      | 1:2:3   |
+            | a    | c  | abc,abc  | 10:10   | 249.9876189:299.962882 | 25:30      | 1:2:3   |
--- a/features/testbot/matching.feature
+++ b/features/testbot/matching.feature
@@ -810,6 +810,6 @@ Feature: Basic Map Matching
         # These should have the same weights/duration in either direction
         When I match I should get
           | trace | geometry             | a:distance    | a:duration | a:weight | duration |
-          | 2345  | 1.00018,1,1.000314,1 | 14.914666491  | 1.4        | 1.4      | 1.4      |
-          | 4321  | 1.00027,1,1.000135,1 | 15.025969972  | 1.5        | 1.5      | 1.5      |
+          | 2345  | 1.00018,1,1.000314,1 | 14.91466649  | 1.4        | 1.4      | 1.4      |
+          | 4321  | 1.00027,1,1.000135,1 | 15.02596997  | 1.5        | 1.5      | 1.5      |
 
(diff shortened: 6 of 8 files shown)


danpat:
@Rejudge-F This introduces a C++20 construct, but so far, this codebase is C++17.  @mjjbell @SiarheiFedartsou How do you feel about bumping the minimum C++ version needed?

Rejudge-F:
> @Rejudge-F This introduces a C++20 construct, but so far, this codebase is C++17. @mjjbell @SiarheiFedartsou How do you feel about bumping the minimum C++ version needed?

I don't think we need to bump the minimum C++ version needed, because the format_to function used by osrm is supported by the third-party fmt-9.1.0(code's here: third_party/fmt-9.1.0 ).

The above reference to the cpp reference is for the convenience of explaining the format_to parameters

SiarheiFedartsou:
@danpat indeed, it seems @Rejudge-F is right and we can support it with C++17(on another hand std::format from C++20 would allow us to get rid from dependency). But in general I don't see a lot of reasons to not use C++20 at the moment - we don't have obligations to maintain compatibility with old compilers.

include/util/json_renderer.hpp
```diff
@@ -56,7 +56,10 @@ template <typename Out> struct Renderer
         size_t decimalpos = std::find(buffer.begin(), buffer.end(), '.') - buffer.begin();
         if (buffer.size() > (decimalpos + 10))
         {
-            buffer.resize(decimalpos + 10);
+            buffer.clear();
+            fmt::format_to(std::back_inserter(buffer), FMT_COMPILE("{0:.10f}"), number.value);
```
SiarheiFedartsou, line 60: Why don't just straightway format with {0:.10f} a couple of lines above?

Rejudge-F: Because it is not necessary to preserve the precision of Number type numbers without a decimal point, for example, 116 will become 116.0000000000 after 10 decimal places are reserved

Rejudge-F: Response will like this:
image

SiarheiFedartsou: I think initial problem is caused by the fact that fmt uses so called scientific notation for such small numbers. Can we just try to turn it off? This fix is better than no fix, but it seems we can make it slightly better.

Something which I found after 1 min in Google: https://www.reddit.com/r/cpp_questions/comments/n61o3w/how_to_force_fmt_not_to_use_exponent_notation/?utm_source=share&utm_medium=ios_app&utm_name=iossmf

SiarheiFedartsou: Probably we can just use {:10f} and always do resize to cut trailing zeros - should work well I think.

SiarheiFedartsou: Or it seems that {:f} should force fmt to not use scientific notation - in this case we can just change format string in initial fmt::format call

Rejudge-F: I think you want use {:.10f}, {:10f} means 10 width, but {:.10f} do not cut trailing zeros too.

Rejudge-F: > I think initial problem is caused by the fact that fmt uses so called scientific notation for such small numbers. Can we just try to turn it off? This fix is better than no fix, but it seems we can make it slightly better.
> 
> Something which I found after 1 min in Google: https://www.reddit.com/r/cpp_questions/comments/n61o3w/how_to_force_fmt_not_to_use_exponent_notation/?utm_source=share&utm_medium=ios_app&utm_name=iossmf

I'm also thinking about whether there is a more elegant way, unfortunately I haven't found a better solution yet

SiarheiFedartsou: Besides that let's also add some simple test to avoid regressions in the future. I think this one should work in this case:
```
BOOST_AUTO_TEST_CASE(test_json_issue_6531) {
    std::string output;
    osrm::util::json::Renderer<std::string> renderer(output);
    renderer(0.0000000000017114087924596788);
    BOOST_CHECK_EQUAL(output, "0.0");
}
```

Rejudge-F: This test is very useful, I just checked [this](https://github.com/therion/therion/blob/devel/thdouble.h ) avoid zero, looks like it would be better to directly reformatter for fractional parts larger than 10 digits.
image

SiarheiFedartsou: I eventually came up with something like this:

```
    void operator()(const Number &number)
    {
        // we don't want to print NaN or Infinity
        BOOST_ASSERT(std::isfinite(number.value));
        // `fmt::memory_buffer` stores first 500 bytes in the object itself(i.e. on stack in this
        // case) and then grows using heap if needed
        fmt::memory_buffer buffer;
        fmt::format_to(std::back_inserter(buffer), FMT_COMPILE("{:f}"), number.value);

        // find out how many trailing zeros we have
        size_t zeros = 0;
        for (size_t index = buffer.size() - 1; index > 1; --index)
        {
            if (buffer[index] != '0' || buffer[index - 1] == '.') {
                break;
            }
            ++zeros;
        }
        // find position of the decimal point
        size_t pointpos = std::find(buffer.begin(), buffer.end(), '.') - buffer.begin();
        // we are removing trailing zeros and truncating the number to 10 digits after the decimal
        buffer.resize(std::min(buffer.size() - zeros, pointpos + 10));

        write(buffer.data(), buffer.size());
    }
 ```
 
 Not sure there are no bugs so - will return to it later(or feel free to continue with it please :) )

Comment (Rejudge-F) [reply include/util/json_renderer.hpp:60]: Nice, but `fmt::format_to(std::back_inserter(buffer), FMT_COMPILE("{:f}"), number.value);` only keep 6 decimal placesill.
I came up with like this 
```
void operator()(const Number &number)
  {
      // `fmt::memory_buffer` stores first 500 bytes in the object itself(i.e. on stack in this
      // case) and then grows using heap if needed
      fmt::memory_buffer buffer, tmp_buffer;

      fmt::format_to(std::back_inserter(tmp_buffer), FMT_COMPILE("{0:.10f}"), number.value);
      fmt::format_to(std::back_inserter(buffer), FMT_COMPILE("{}"), std::stod(to_string(tmp_buffer)));

      write(buffer.data(), buffer.size());
  }
```

Comment (Rejudge-F) [reply include/util/json_renderer.hpp:60]: response looks fine, the result will no longer be 0.0 but 0

Comment (SiarheiFedartsou) [reply include/util/json_renderer.hpp:60]: Doesn't `std::stod` rely on locales? E.g. on some systems it may expect `,` to be used as floating point separator?

Comment (Rejudge-F) [reply include/util/json_renderer.hpp:60]: https://en.cppreference.com/w/cpp/string/basic_string/stof
Looks like there are no rely on locales.

Comment (Rejudge-F) [reply include/util/json_renderer.hpp:60]: > I eventually came up with something like this:
> 
> ```
>     void operator()(const Number &number)
>     {
>         // we don't want to print NaN or Infinity
>         BOOST_ASSERT(std::isfinite(number.value));
>         // `fmt::memory_buffer` stores first 500 bytes in the object itself(i.e. on stack in this
>         // case) and then grows using heap if needed
>         fmt::memory_buffer buffer;
>         fmt::format_to(std::back_inserter(buffer), FMT_COMPILE("{:f}"), number.value);
> 
>         // find out how many trailing zeros we have
>         size_t zeros = 0;
>         for (size_t index = buffer.size() - 1; index > 1; --index)
>         {
>             if (buffer[index] != '0' || buffer[index - 1] == '.') {
>                 break;
>             }
>             ++zeros;
>         }
>         // find position of the decimal point
>         size_t pointpos = std::find(buffer.begin(), buffer.end(), '.') - buffer.begin();
>         // we are removing trailing zeros and truncating the number to 10 digits after the decimal
>         buffer.resize(std::min(buffer.size() - zeros, pointpos + 10));
> 
>         write(buffer.data(), buffer.size());
>     }
> ```
> 
> Not sure there are no bugs so - will return to it later(or feel free to continue with it please :) )

use this will be fine
fmt::format_to(std::back_inserter(buffer), FMT_COMPILE("{:.10f}"), number.value);

example:
fmt::format_to(std::back_inserter(buffer), FMT_COMPILE("{:.f}"), 116.123456789); will returns 116.123456 
fmt::format_to(std::back_inserter(buffer), FMT_COMPILE("{:.10f}"), 116.123456789); will returns 116.1234567890

Comment (Rejudge-F) [reply include/util/json_renderer.hpp:60]: @SiarheiFedartsou I found a more reasonable way
```
void operator()(const Number &number)
{
    // we don't want to print NaN or Infinity
    BOOST_ASSERT(std::isfinite(number.value));
    // `fmt::memory_buffer` stores first 500 bytes in the object itself(i.e. on stack in this
    // case) and then grows using heap if needed
    fmt::memory_buffer buffer;
    fmt::format_to(std::back_inserter(buffer), FMT_COMPILE("{:.10g}"), number.value);

    write(buffer.data(), buffer.size());
}
```
https://fmt.dev/latest/syntax.html#grammar-token-sf-precision

File: unit_tests/util/json_render.cpp [hunk 2]
```diff
@@ -27,4 +27,11 @@ BOOST_AUTO_TEST_CASE(integer)
     BOOST_CHECK_EQUAL(str, "42");
 }
 
+BOOST_AUTO_TEST_CASE(test_json_issue_6531) {
```
SiarheiFedartsou, line 30: @Rejudge-F  Cool that you found this `g` suffix! Thanks! May I ask you to add 3-4 test cases more in order to make sure that the rest of the things work as expected? e.g. `42.1`, `42.0`, `0.12345678912345`? 

Let me know if you don't have time on it and I will try to do that on my own.

Rejudge-F: I will do it.

DennisOSRM:
Nice change.

Rejudge-F:
I've no idea about this check item([clang-15.0-debug-clang-tidy](https://github.com/Project-OSRM/osrm-backend/actions/runs/4060677303/jobs/6993663397)) error, looks like dns can't resolve `apt.llvm.org`.

--2023-02-01 08:48:16--  https://apt.llvm.org/llvm-snapshot.gpg.key
Resolving apt.llvm.org (apt.llvm.org)... Warning: apt-key is deprecated. Manage keyring files in trusted.gpg.d instead (see apt-key(8)).
failed: Temporary failure in name resolution.
wget: unable to resolve host address ‘apt.llvm.org’
gpg: no valid OpenPGP data found.

SiarheiFedartsou:
Hey @Rejudge-F 
I just re-run failing job - it seems there was some one-time issue, but it still fails due to precision issues in some of tests:

```
2023-02-01T13:03:49.2459832Z 1) Scenario: [1mSpeed annotations should handle zero segments[22m - [90mfeatures/testbot/annotations.feature:96[39m
2023-02-01T13:03:49.2460578Z    Step: [1mWhen I route I should get[22m - [90mfeatures/testbot/annotations.feature:116[39m
2023-02-01T13:03:49.2461198Z    Step Definition: [90mfeatures/step_definitions/routing.js:4[39m
2023-02-01T13:03:49.2461529Z    Message:
2023-02-01T13:03:49.2463027Z      [31mTables were not identical:
2023-02-01T13:03:49.2463708Z      | from | to | route   | a:speed |     a:distance                  | a:duration | a:nodes |
2023-02-01T13:03:49.2464328Z      | a    | c  | abc,abc | 10:10   | (-) 249.987618946:299.962882039 | 25:30      | 1:2:3   |
2023-02-01T13:03:49.2464882Z      | a    | c  | abc,abc | 10:10   | (+) 249.9876189:299.962882      | 25:30      | 1:2:3   |[39m
```

Will you be able to update it too please?

Rejudge-F:
> Hey @Rejudge-F 
> 
> I just re-run failing job - it seems there was some one-time issue, but it still fails due to precision issues in some of tests:
> 
> 
> 
> ```
> 
> 2023-02-01T13:03:49.2459832Z 1) Scenario: [1mSpeed annotations should handle zero segments[22m - [90mfeatures/testbot/annotations.feature:96[39m
> 
> 2023-02-01T13:03:49.2460578Z    Step: [1mWhen I route I should get[22m - [90mfeatures/testbot/annotations.feature:116[39m
> 
> 2023-02-01T13:03:49.2461198Z    Step Definition: [90mfeatures/step_definitions/routing.js:4[39m
> 
> 2023-02-01T13:03:49.2461529Z    Message:
> 
> 2023-02-01T13:03:49.2463027Z      [31mTables were not identical:
> 
> 2023-02-01T13:03:49.2463708Z      | from | to | route   | a:speed |     a:distance                  | a:duration | a:nodes |
> 
> 2023-02-01T13:03:49.2464328Z      | a    | c  | abc,abc | 10:10   | (-) 249.987618946:299.962882039 | 25:30      | 1:2:3   |
> 
> 2023-02-01T13:03:49.2464882Z      | a    | c  | abc,abc | 10:10   | (+) 249.9876189:299.962882      | 25:30      | 1:2:3   |[39m
> 
> ```
> 
> 
> 
> Will you be able to update it too please?

Willing to

Rejudge-F:
I need to ask how I should find all the wrong cases

Rejudge-F:
I have fix all cucumber issue.
detail:
```
Failures:

1) Scenario: Car - DC North capitol situation, two on one off - features/car/conditional_restrictions.feature:499
   Step: When I route I should get - features/car/conditional_restrictions.feature:539
   Step Definition: features/step_definitions/routing.js:4
   Message:
     Tables were not identical:
     | from | to |     route                                      |     turns                                   |
     | a    | e  | (-) cap south,florida nw,florida nw,florida ne | (-) depart,turn right,continue uturn,arrive |
     | a    | e  | (+) cap south,florida ne,florida ne            | (+) depart,turn left,arrive                 |
     | f    | d  |     cap north,florida nw,florida nw            |     depart,turn left,arrive                 |
     | e    | c  | (-) florida ne,florida nw,cap south,cap south  | (-) depart,continue uturn,turn right,arrive |
     | e    | c  | (+) florida ne,cap south,cap south             | (+) depart,turn left,arrive                 |

2) Scenario: Car - DC North capitol situation, one on two off - features/car/conditional_restrictions.feature:546
   Step: When I route I should get - features/car/conditional_restrictions.feature:586
   Step Definition: features/step_definitions/routing.js:4
   Message:
     Tables were not identical:
     | from | to |     route                                      |     turns                                         |
     | a    | e  |     cap south,florida ne,florida ne            |     depart,turn left,arrive                       |
     | f    | d  | (-) cap north,florida ne,florida ne,florida nw | (-) depart,turn sharp right,continue uturn,arrive |
     | f    | d  | (+) cap north,florida nw,florida nw            | (+) depart,turn left,arrive                       |
     | e    | c  |     florida ne,cap south,cap south             |     depart,turn left,arrive                       |

3) Scenario: Speed annotations should handle zero segments - features/testbot/annotations.feature:96
   Step: When I route I should get - features/testbot/annotations.feature:116
   Step Definition: features/step_definitions/routing.js:4
   Message:
     Tables were not identical:
     | from | to | route   | a:speed |     a:distance                  | a:duration | a:nodes |
     | a    | c  | abc,abc | 10:10   | (-) 249.987618946:299.962882039 | 25:30      | 1:2:3   |
     | a    | c  | abc,abc | 10:10   | (+) 249.9876189:299.962882      | 25:30      | 1:2:3   |

4) Scenario: Regression test - add source phantom properly (two phantoms on one edge) - features/testbot/matching.feature:792
   Step: When I match I should get - features/testbot/matching.feature:811
   Step Definition: features/step_definitions/matching.js:7
   Message:
     Tables were not identical:
     | trace | geometry             |     a:distance   | a:duration | a:weight | duration |
     | 2345  | 1.00018,1,1.000314,1 | (-) 14.914666491 | 1.4        | 1.4      | 1.4      |
     | 2345  | 1.00018,1,1.000314,1 | (+) 14.91466649  | 1.4        | 1.4      | 1.4      |
     | 4321  | 1.00027,1,1.000135,1 | (-) 15.025969972 | 1.5        | 1.5      | 1.5      |
     | 4321  | 1.00027,1,1.000135,1 | (+) 15.02596997  | 1.5        | 1.5      | 1.5      |

5) Scenario: Shortest via path with continuation, multiple waypoint snapping candidates - features/testbot/snap_intersection.feature:529
   Step: When I route I should get - features/testbot/snap_intersection.feature:566
   Step Definition: features/step_definitions/routing.js:4
   Message:
     Tables were not identical:
     | waypoints | route             |     time           |     weight        |
     | a,f       | ad,df,df          |     40s            |     40            |
     | f,k       | fj,kj,kj          |     65.6s          |     65.6          |
     | a,f,k     | ac,cf,cf,fj,kj,kj |     132.8s         |     132.8         |
     | k,f       | ik,fi,fi          |     54.3s          |     54.3          |
     | f,a       | ef,ae,ae          |     66.6s          |     66.6          |
     | k,f,a     | kj,fj,fj,ef,ae,ae | (-) 141.399999999s | (-) 141.399999999 |
     | k,f,a     | kj,fj,fj,ef,ae,ae | (+) 141.4s         | (+) 141.4         |

6) Scenario: Shortest via path with uturns, multiple waypoint snapping candidates - features/testbot/snap_intersection.feature:582
   Step: When I route I should get - features/testbot/snap_intersection.feature:622
   Step Definition: features/step_definitions/routing.js:4
   Message:
     Tables were not identical:
     | waypoints | route             |     time           |     weight        |
     | a,f       | ad,df,df          |     40s            |     40            |
     | f,k       | fj,kj,kj          |     65.6s          |     65.6          |
     | a,f,k     | ad,df,df,fj,kj,kj |     105.6s         |     105.6         |
     | k,f       | ik,fi,fi          |     54.3s          |     54.3          |
     | f,a       | ef,ae,ae          |     66.6s          |     66.6          |
     | k,f,a     | ik,fi,fi,ef,ae,ae | (-) 120.899999999s | (-) 120.899999999 |
     | k,f,a     | ik,fi,fi,ef,ae,ae | (+) 120.9s         | (+) 120.9         |

7) Scenario: Start and target on the same and adjacent edge - features/testbot/weight.feature:39
   Step: When I route I should get - features/testbot/weight.feature:54
   Step Definition: features/step_definitions/routing.js:4
   Message:
     Tables were not identical:
     | waypoints | route   | distances | weights | times   |     a:distance                | a:duration | a:weight | a:speed |
     | s,t       | abc,abc | 20m,0m    | 2,0     | 2s,0s   | (-) 20.034626629              | 2          | 2        | 10      |
     | s,t       | abc,abc | 20m,0m    | 2,0     | 2s,0s   | (+) 20.03462663               | 2          | 2        | 10      |
     | t,s       | abc,abc | 20m,0m    | 2,0     | 2s,0s   | (-) 20.034626629              | 2          | 2        | 10      |
     | t,s       | abc,abc | 20m,0m    | 2,0     | 2s,0s   | (+) 20.03462663               | 2          | 2        | 10      |
     | s,e       | abc,abc | 40m,0m    | 3.9,0   | 3.9s,0s | (-) 29.940636463:10.017313314 | 3:0.9      | 3:0.9    | 10:11.1 |
     | s,e       | abc,abc | 40m,0m    | 3.9,0   | 3.9s,0s | (+) 29.94063646:10.01731331   | 3:0.9      | 3:0.9    | 10:11.1 |
     | e,s       | abc,abc | 40m,0m    | 3.9,0   | 3.9s,0s | (-) 10.017313314:29.940636463 | 0.9:3      | 0.9:3    | 11.1:10 |
     | e,s       | abc,abc | 40m,0m    | 3.9,0   | 3.9s,0s | (+) 10.01731331:29.94063646   | 0.9:3      | 0.9:3    | 11.1:10 |

1351 scenarios (7 failed, 1344 passed)
7246 steps (7 failed, 1 skipped, 7238 passed)
5m12.590s
```

features/car/conditional_restrictions.feature
```diff
@@ -538,9 +538,9 @@ Feature: Car - Turn restrictions
 
         When I route I should get
             | from | to | route                                      | turns                                       |
-            | a    | e  | cap south,florida nw,florida nw,florida ne | depart,turn right,continue uturn,arrive     |
+            | a    | e  | cap south,florida ne,florida ne | depart,turn left,arrive     |
```
SiarheiFedartsou, line 541: I am curious how it is possible that it was changed due to changes in JSON rendering? 🤔

SiarheiFedartsou: I guess it is due to some caches on CI because you have new failure which says you have "original" values here 
https://pipelines.actions.githubusercontent.com/serviceHosts/5f8b3a21-777f-4ccd-83b2-9ebc854aa1e0/_apis/pipelines/1/runs/1148/signedlogcontent/31?urlExpires=2023-02-02T09%3A15%3A35.9673538Z&urlSigningMethod=HMACV1&urlSignature=k5UH1girtk3krsR2TdceIF1EsrnxhL2RtXruekV8s%2BU%3D

Rejudge-F: This case is torturing me non-stop hahahaha, do you have a good solution?

SiarheiFedartsou: Yes, I think I can try dropping CI caches. Sorry, I am a bit busy right now, but will try to do that this evening(not sure which time zone you are, it will be in ~8-10 hours).

Rejudge-F: > 

Then this issue will be handed over to you, and I will no longer submit relevant code. I'm in GMT+8

SiarheiFedartsou:
@Rejudge-F many thanks for this contribution!

SiarheiFedartsou (approved):

DennisOSRM:
Very nice contribution. Tyvm

PR #6564, opened 2023-03-03
print tracebacks and line numbers for Lua runtime errors

### LINKED ISSUE #6563: No line numbers in Lua runtime errors
# Feature

When a runtime error occurs in a Lua profile script, OSRM exits with a message that says what the error was, but not a traceback or a line number, e.g.

```
$ osrm-extract -p ../profiles/bicycle_lts.lua bicycle_lts/bicycle_lts.osm.pbf
[lots of intermediate output]
libc++abi: terminating with uncaught exception of type sol::error: lua: error: attempt to compare two nil values
```

It would make debugging profiles much easier if there were a traceback and line number. I have a patch for this, will submit PR shortly.

This is related to #3084, but I think it's not a duplicate as the discussion there focuses on syntax errors, which in the latest OSRM already display line numbers.

mattwigway:
# Issue

This fixes #6563, and provides tracebacks and line number for Lua runtime errors. It does this by replacing all `sol::function`s with `sol::protected_function` and checking for errors after a Lua function is called. If one is detected, a traceback is printed to stderr, and an exception is thrown which causes OSRM to exit (as it did before when runtime errors were encountered).

Before:
```
$ osrm-extract -p ../profiles/bicycle_lts.lua bicycle_lts/bicycle_lts.osm.pbf
[lots of intermediate output]
libc++abi: terminating with uncaught exception of type sol::error: lua: error: attempt to compare two nil values
```

After:
```osrm-extract -p ../profiles/bicycle_lts.lua bicycle_lts/bicycle_lts.osm.pbf
[lots of intermediate output]
attempt to compare two nil values
stack traceback:
	[C]: in function 'math.max'
	...bc/git/vmt-mode-shift-study/src/routing/profiles/lts.lua:128: in function 'lts.lts_for_way'
	../profiles/bicycle_lts.lua:515: in function 'process_way'
attempt to compare two nil values
stack traceback:
	[C]: in function 'math.max'
	...bc/git/vmt-mode-shift-study/src/routing/profiles/lts.lua:128: in function 'lts.lts_for_way'
	../profiles/bicycle_lts.lua:515: in function 'process_way'
attempt to compare two nil values
stack traceback:
	[C]: in function 'math.max'
	...bc/git/vmt-mode-shift-study/src/routing/profiles/lts.lua:128: in function 'lts.lts_for_way'
	../profiles/bicycle_lts.lua:515: in function 'process_way'
attempt to compare two nil values
stack traceback:
	[C]: in function 'math.max'
	...bc/git/vmt-mode-shift-study/src/routing/profiles/lts.lua:128: in function 'lts.lts_for_way'
	../profiles/bicycle_lts.lua:515: in function 'process_way'
libc++abi: terminating with uncaught exception of type osrm::util::exception: Lua error (see stderr for traceback)
```

I believe the error message is printed multiple times because the error is getting hit in multiple threads before the exception kills the entire process. My understanding is that a single exception should kill the entire process, not just the thread, but I'm new to C++ so if that's incorrect changes will be required to ensure runtime errors actually terminate extraction.

Performance impact is negligible/nonexistent. With my small test PBF file, it took 3.511s to build and partition with this patch and 3.563 without.

~I ran the `format.sh` script as requested and it made some format changes elsewhere in the files. If this is an issue I can go back and manually revert those changes. It doesn't seem to pollute the diff too much.~ went ahead and reverted these for ease of review.

## Tasklist

I'm new to C++ and contributing to OSRM, so there's a few things I'm not completely sure of and would love feedback on.

 - [x] Make sure that errors in any Lua thread will bring the entire extraction process to a halt, not just that thread.
 - [x] I don't know what `source_function` and `GetStringListFromFunction` do. I changed these to have Lua error handling but it I'm not sure how to test them.
 - [x] Most places where profile functions are called there's a switch statement based on version; I define the result variable that I use to check for the error outside the statement. If the profile API version is increased, and the switch statements are not updated appropriately so that the lua function is always called, it's possible that the code would check for an error using an uninitialized result. I know C++ does weird things in this case - is it desirable to add an explicit check to each function to ensure the result has actually been initialized before checking if it is valid?

## Requirements / Relations

N/A

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@
       - CHANGED: Make edge metrics strongly typed [#6420](https://github.com/Project-OSRM/osrm-backend/pull/6420)
       - FIXED: Typo in file name src/util/timed_historgram.cpp -> src/util/timed_histogram.cpp [#6428](https://github.com/Project-OSRM/osrm-backend/issues/6428)
       - CHANGED: Replace boost::string_ref with std::string_view [#6433](https://github.com/Project-OSRM/osrm-backend/pull/6433)
+      - ADDED: Print tracebacks for Lua runtime errors [#6564](https://github.com/Project-OSRM/osrm-backend/pull/6564)
     - Routing:
       - FIXED: Fix adding traffic signal penalties during compression [#6419](https://github.com/Project-OSRM/osrm-backend/pull/6419)
 # 5.27.1
--- a/include/extractor/scripting_environment_lua.hpp
+++ b/include/extractor/scripting_environment_lua.hpp
@@ -41,10 +41,10 @@ struct LuaScriptingContext final
     bool has_way_function = false;
     bool has_segment_function = false;
 
-    sol::function turn_function;
-    sol::function way_function;
-    sol::function node_function;
-    sol::function segment_function;
+    sol::protected_function turn_function;
+    sol::protected_function way_function;
+    sol::protected_function node_function;
+    sol::protected_function segment_function;
 
     int api_version = 4;
     sol::table profile_table;
--- a/scripts/ci/leaksanitizer.conf
+++ b/scripts/ci/leaksanitizer.conf
@@ -6,3 +6,12 @@
 #     #1 0x7f7ae595d13e  (/usr/lib/x86_64-linux-gnu/libtbb.so.2+0x2213e)
 
 leak:libtbb.so
+
+# The extract-tests leak some memory in the tests to confirm that
+# lua errors print tracebacks.
+# This appears to be because when these tests throw exceptions, the
+# osmium objects being processed are not freed. In production this doesn't
+# matter, as the exceptions bring down the entire osrm-extract process. In the
+# tests, we catch the error to make sure it occurs, which is why the
+# leaksanitizer flags it.
+leak:extract-tests
--- a/src/extractor/scripting_environment_lua.cpp
+++ b/src/extractor/scripting_environment_lua.cpp
@@ -245,8 +245,7 @@ void Sol2ScriptingEnvironment::InitContext(LuaScriptingContext &context)
                                                  "valid",
                                                  &osmium::Location::valid);
 
-    auto get_location_tag = [](auto &context, const auto &location, const char *key)
-    {
+    auto get_location_tag = [](auto &context, const auto &location, const char *key) {
         if (context.location_dependent_data.empty())
             return sol::object(context.state);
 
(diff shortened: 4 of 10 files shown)


mattwigway:
@SiarheiFedartsou I think this is ready to go! I added unit tests, backed out the changes for those functions we weren't sure about, and confirmed that an error on any lua thread shuts OSRM down entirely. I don't see any wiki pages that would be affected by this change, so haven't made any updates there.

SiarheiFedartsou:
Thanks @mattwigway !
Can you please rebase/merge this PR with master? It contains fixes for CI which was not working before...

SiarheiFedartsou:
@mattwigway 
and you should reformat code here. It can be done with `./scripts/format.sh` script, make sure to use `clang-format-10` please(and let me know in case of any problems please)

mattwigway:
Ah, clang-format-10 is the piece I was missing before. Rebased and formatted.

SiarheiFedartsou:
@mattwigway you are probably missing #include <fstream> - that’s why CI is red. Please fix when you have a time… Or let me know and I will do it on my own in the nearest days.

mattwigway:
Oops, I meant to delete the function that uses fstream as it's from an earlier way I tried to specify the test. Will fix.

mattwigway:
I'm not sure what the memory leak test is flagging. I'll investigate later this week.

SiarheiFedartsou:
> I'm not sure what the memory leak test is flagging. I'll investigate later this week.

Yeah, would be nice, usually it does not have false positives, but taking into account use case(it happens only on error and we exit program anyway) I think we can just suppress it if we won’t find a reason or it will be too complicated to fix. Also I was going to look on my own, but I think I will be able to do that only on the next week in the best case.

SiarheiFedartsou:
@mattwigway then I think you can configure it somewhere here https://github.com/Project-OSRM/osrm-backend/blob/192d077adad5bc0d1a5fb9d185c5c4113deabacb/scripts/ci/leaksanitizer.conf#L8

AFAIR this file follows syntax described in https://clang.llvm.org/docs/SanitizerSpecialCaseList.html, so you can suppress this specific issue.

mattwigway:
@SiarheiFedartsou looks like that change fixed it. I rebased from master, LMK if anything else is needed.

SiarheiFedartsou:
@mattwigway many thanks for this contribution and I will be happy to review any other PRs from you :+1:

Issue #2983, opened 2016-10-02
Require a `radius` parameter other then `unlimited` when using `bearings`


TheMarex:
Currently every plugin will search the whole StaticRTree until a matching bearing segment is found - no matter how far away it is. For a lot of users this is an unexpected result and can have serious performance implications (even on accident).

Hence to make the user aware that it actually only makes sense to search in a specific search radius, say 100m when using the bearings filter, we should enforce this on the API level.

This means in case of no segment with the given bearing value in the search radius we will return `NoSegment` which is a better indicator then just returning a coordinate potentially across the Atlantic.

Two options for implementing this:

1. `bearings` without setting `radius` will error (invalid request)
2. `bearings` implies that `radius=SOME_DEFAULT_VALUE`

I would prefer option 1 because it makes it more explicit that we changed the behavior.

chaupow:
I guess in most cases when the segment with a matching bearing is too far away, the range of the bearing is probably too strict/narrow/small. Currently, we also do not recommend any value that should be used as a bearing range in osrm. We should at least recommend some range that makes sense in the osrm documentation (is `90` a good range? and a short sentence explaining range would be nice too...)

per chat with @TheMarex:
It might also make sense to decide on a default range `X` for bearings and use that default value. Three options for implementing this:

1. change the bearings option to `bearings={value}` which works like `bearings={value},X`
2. accept bearings in both forms, `bearings={value},{range}` and just `bearings={value}` which works like `bearings={value},X`
3. leave it as it is

I like option 1 as it makes things easier but I don't know if there are many use cases where people would want to set specific ranges.

Otherwise having a `radius` definitely sounds good - especially considering performance. 

@1ec5, @freenerd, @danpat any opinions on this?

1ec5:
I don't have a strong opinion on the syntax changes proposed here; libraries such as MapboxDirections.swift rationalize away any syntactic problems anyways. Rather, I think most of the confusion stems from bearing falling into an uncanny valley that's just different enough from the expected behavior, but coupling with a radius seems to address that.

whytro:
I'd like to take a crack at this if this issue is still open and available.
Unless I missed something, it looks like it should only require changes to the parameters' validation function/help message generation, as well as the corresponding adjustments for each service. 
This would be for the first option of "`bearings` without setting `radius` will error", with the goal of causing such requests to return an `InvalidOptions` code.

PR #6572, opened 2023-03-09
Require a radius parameter when using bearings

whytro:
# Issue

> #2983 Require a `radius` parameter other than unlimited when using `bearings`

## Tasklist

 - [ ] review
 - [ ] adjust for comments

## Requirements / Relations


nilsnolde:
Is this for GSoC by any chance?

nilsnolde:
It’s just easier to keep track of things that way. Anyways, thanks for the contribution. I’m not familiar with the codebase to give it a review, but I’m sure others will soonish.

Should we flag this as pre-gsoc so it’s easier to find when the time comes?

whytro:
> Should we flag this as pre-gsoc so it’s easier to find when the time comes?

I think that would be a good idea!

whytro:
I should note that I did fix the formatting by rerunning the format script - though it did seem to come with some strange looking results due to the increased length.

nilsnolde:
As this is for GSoC, would you have time to have a look @SiarheiFedartsou ?

SiarheiFedartsou:
Hey @whytro 
I took a quick look into this PR and it seems that everything is okay in general(but I will additionally look into details later), but it seems that we don't fix behaviour for NodeJs bindings here in any way. And it seems to be the reason why we have [NodeJs tests failure on CI](https://github.com/Project-OSRM/osrm-backend/actions/runs/4396762364/jobs/7760777515): 
```
✓ route: integer bearing values no longer supported
# route: valid bearing values
[assert][139638292518848] /home/runner/work/osrm-backend/osrm-backend/src/nodejs/node_osrm.cpp:117
in: void node_osrm::async(const Napi::CallbackInfo &, ParameterParser, ServiceMemFn, bool) [ParameterParser = std::unique_ptr<osrm::engine::api::RouteParameters, std::default_delete<osrm::engine::api::RouteParameters> > (*)(const Napi::CallbackInfo &, bool), ServiceMemFn = osrm::engine::Status (osrm::OSRM::*)(const osrm::engine::api::RouteParameters &, mapbox::util::variant<osrm::util::json::Object, std::__cxx11::basic_string<char>, flatbuffers::FlatBufferBuilder> &) const]: params->IsValid()
```

This assert comes from here:
https://github.com/Project-OSRM/osrm-backend/blob/82b73ed9ab9023f1477dd65a746fc68705c8077a/src/nodejs/node_osrm.cpp#L117

So I think we need to at least fix [NodeJs tests](https://github.com/Project-OSRM/osrm-backend/tree/master/test/nodejs) to respect this new requirement. In perfect scenario I would even add some user-friendly exception in NodeJs bindings for the case of this error(but we can do it in separate PR later). It can be done somewhere here:
https://github.com/Project-OSRM/osrm-backend/blob/82b73ed9ab9023f1477dd65a746fc68705c8077a/include/nodejs/node_osrm_support.hpp#L519

Let me please know if something is not clear or you will need a help with it :)

whytro:
Ah, that was sloppy of me! I've gone ahead and adjusted the bindings and the tests accordingly. Thank you for the pointers!

whytro:
I originally thought that the unit tests were for release builds (hence not asserting on parameter checks), but now I'm realizing that parameter checking tests are maybe not applicable for unit testing, and are mainly to be covered in the cucumber tests. I'll take a closer look at implementing those tests in the appropriate (cucumber) feature sections, and revert the unit tests change as soon as possible!

SiarheiFedartsou:
> I originally thought that the unit tests were for release builds (hence not asserting on parameter checks - which, in hindsight, is a nonsensical assumption), but now I'm realizing that parameter checking tests are maybe not applicable for unit testing, and are mainly to be covered in the cucumber tests. I'll take a closer look at implementing those tests in the appropriate (cucumber) feature sections, and revert the unit tests change as soon as possible!

But can we change unit tests in such a way to not cause those asserts?

danpat:
I also want to note that this is an API-breaking change - clients that were making requests using `bearings` before will suddenly stop working.  Merging this feature will require a new major release version - is anyone game for doing that?

SiarheiFedartsou:
> Perhaps merging this unblocks a period of other major breaking changes and starts the train rolling towards a reasonably significant major release.

It is what I am also thinking about 🤔 But the problem is that there are actually almost no people actively working on this project, so not sure we really want it... 
On another hand I don't see next release on a horizon at all and we can just agree that next release will be major update and it would unblock very slow work on adding other breaking changes we wanted to do.

whytro:
Hmm, I'm not sure if the assertion is avoidable in this case without significant alteration to the library. It also doesn't seem like similar tests (ie. for param validation checking) exist, and instead seem to focus on the calculation results (please let me know if I'm wrong here).

I did also consider that this change would probably be breaking given that it changes behavior significantly for bearings. Should I look to either implement a flag or add in default radius values when bearings are used, so that it can be easily changed should a major release happens - or rather what would be the best approach here?

SiarheiFedartsou:
> I did also consider that this change would probably be breaking given that it changes behavior significantly for bearings. Should I look to either implement a flag or add in default radius values when bearings are used, so that it can be easily changed should a major release happens - or rather what would be the best approach here?

I would probably start from flag, I totally agree with Daniel's points above. Sorry, that I didn't notice it in the beginning - it seems to be quite big rework of this PR... Also I still not sure if we want to go ahead with major version update - probably it can be a good idea to start new PR and to leave this PR "as is" as it seems what you did here is better default behaviour than it used to be (the only problem is backward compatibility).

whytro:
Okay, I'll look into adding the configurable default `radius` flags and open up a separate PR for it when ready!

whytro:
Sorry for the force-push, I messed up a rebase and polluted the commit history on the branch. 
These changes are mainly aimed at taking a cleaner approach to the problem overall, dependent on #6599, and expectant on a future change where the server doesn't default to unlimited (only one line would need a change) - and as the previous version of this would not have been drop-in compatible with the default_radius flag. 
Essentially, the goal for this was to just add support for if/when the server doesn't default to an infinite value, or have a flag in which radii must be specified - where the error becomes more relevant.

DennisOSRM:
latest build failed with

```
/home/runner/work/osrm-backend/osrm-backend/include/nodejs/node_osrm_support.hpp:593:37: error: use of undeclared identifier 'default_radius'
  593 |         if (!obj.Has("radiuses") && default_radius.IsUndefined())
      |                                     ^
1 error generated.
```

Issue #6558, opened 2023-02-28
GSoC 2023 announcement


nilsnolde:
The project will be: officially supported OSRM Python bindings. See more here: https://wiki.openstreetmap.org/wiki/Google_Summer_of_Code/2023/Project_ideas.

To regulate applications, please read the following wiki entry carefully, if you want to be considered:
https://github.com/gis-ops/valhalla-app/issues/67

SiarheiFedartsou:
@nilsnolde sure, done 👍

Zebreu:
Did anything happen here by the way?

I wrote python bindings as part of https://github.com/datakind/dk-routing/tree/main/dkroutingtool (you can build the docker image and interface directly with the table, nearest, and route C++ functions from the osrmbindings python library)

I can help anyone who wants to bring that out as an easily distributable library. I used pybind11 and it was straightforward enough. I'm about to make some changes for performance reasons so I'd be more than happy to work it into something more general and fast.

nilsnolde:
Right, sorry for not giving any updates here. The project was very successful thanks to @whytro and the bindings are here: https://github.com/gis-ops/py-osrm (might still change to OSRM namespace). I've been super busy before going on a longer vacation (where I am currently), I didn't have any time yet to set up PyPI distribution workflows. Will still take a few weeks, but it'll surely still happen in 2023:)

Thanks for sharing your project. I didn't look into it too much, but seems like you're just exposing the table endpoint right? Would be cool if you could try our library once it's published (or feel free now), we'll want some people using and testing it:)

Zebreu:
I'm exposing nearest, table and route and returns json, so I'll take a look at py-osrm and see if there are performance differences. Something I wanted to do was to have multiple route geometry results in a single request and potentially share memory objects (like numpy maybe?) to avoid parsing parameters and JSON every route, but if py-osrm is considerably faster I may be able to avoid that. Thanks!

PR #6575, opened 2023-03-19
Add support for a default_bearing_radius flag

whytro:
# Issue

Relevant to:
> https://github.com/Project-OSRM/osrm-backend/issues/2983 Require a radius parameter other than unlimited when using bearings

## Tasklist

 - [ ] review
 - [ ] adjust for comments

## Requirements / Relations

> Relevant to PR #6572: Require a radius parameter when using bearings (GSoC)

### DIFF
--- a/include/engine/plugins/plugin_base.hpp
+++ b/include/engine/plugins/plugin_base.hpp
@@ -28,9 +28,8 @@ class BasePlugin
 {
   protected:
     BasePlugin() = default;
-    
-    BasePlugin(boost::optional<double> default_radius_) 
-        : default_radius(std::move(default_radius_))
+
+    BasePlugin(boost::optional<double> default_radius_) : default_radius(std::move(default_radius_))
     {
     }
 
--- a/include/engine/plugins/plugin_base.hpp
+++ b/include/engine/plugins/plugin_base.hpp
@@ -237,7 +237,7 @@ class BasePlugin
             phantom_nodes[i] = facade.NearestPhantomNodes(
                 parameters.coordinates[i],
                 number_of_results,
-                use_radiuses ? parameters.radiuses[i] : boost::none,
+                use_radiuses ? parameters.radiuses[i] : default_bearing_radius,
                 use_bearings ? parameters.bearings[i] : boost::none,
                 use_approaches && parameters.approaches[i] ? parameters.approaches[i].get()
                                                            : engine::Approach::UNRESTRICTED);
@@ -279,7 +279,7 @@ class BasePlugin
 
             alternatives[i] = facade.NearestCandidatesWithAlternativeFromBigComponent(
                 parameters.coordinates[i],
-                use_radiuses ? parameters.radiuses[i] : boost::none,
+                use_radiuses ? parameters.radiuses[i] : default_bearing_radius,
                 use_bearings ? parameters.bearings[i] : boost::none,
                 use_approaches && parameters.approaches[i] ? parameters.approaches[i].get()
                                                            : engine::Approach::UNRESTRICTED,
@@ -320,6 +320,8 @@ class BasePlugin
         return std::string("Could not find a matching segment for coordinate ") +
                std::to_string(missing_index);
     }
+
+    boost::optional<double> default_bearing_radius;
 };
 } // namespace osrm::engine::plugins
 
--- a/include/engine/plugins/plugin_base.hpp
+++ b/include/engine/plugins/plugin_base.hpp
@@ -29,9 +29,7 @@ class BasePlugin
   protected:
     BasePlugin() = default;
 
-    BasePlugin(const boost::optional<double> default_radius_) : default_radius(default_radius_)
-    {
-    }
+    BasePlugin(const boost::optional<double> default_radius_) : default_radius(default_radius_) {}
 
     bool CheckAllCoordinates(const std::vector<util::Coordinate> &coordinates) const
     {
--- a/include/engine/plugins/plugin_base.hpp
+++ b/include/engine/plugins/plugin_base.hpp
@@ -237,7 +237,7 @@ class BasePlugin
             phantom_nodes[i] = facade.NearestPhantomNodes(
                 parameters.coordinates[i],
                 number_of_results,
-                use_radiuses ? parameters.radiuses[i] : default_bearing_radius,
+                use_radiuses ? parameters.radiuses[i] : default_radius,
                 use_bearings ? parameters.bearings[i] : boost::none,
                 use_approaches && parameters.approaches[i] ? parameters.approaches[i].get()
                                                            : engine::Approach::UNRESTRICTED);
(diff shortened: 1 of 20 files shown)


include/engine/plugins/plugin_base.hpp
```diff
@@ -237,7 +237,7 @@ class BasePlugin
             phantom_nodes[i] = facade.NearestPhantomNodes(
                 parameters.coordinates[i],
                 number_of_results,
-                use_radiuses ? parameters.radiuses[i] : boost::none,
+                use_radiuses ? parameters.radiuses[i] : default_bearing_radius,
```
SiarheiFedartsou, line 240: May be I don't get something, but by current semantics it is just `default_radius`, but not `default_bearing_radius`, right? I.e. if it is passed we always use it as default radius, isn't it? I don't mind about having such behaviour, but lets rename it to just `default_radius` may be then?

test/nodejs/index.js
```diff
@@ -112,6 +112,18 @@ test('constructor: throws if dataset_name is not a string', function(assert) {
     assert.throws(function() { new OSRM({dataset_name: "unsued_name___", shared_memory: true}); }, /Could not find shared memory region/, 'Does not accept wrong name');
 });
 
+test('constructor: takes a default_bearing_radius argument', function(assert) {
```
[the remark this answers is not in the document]
SiarheiFedartsou: https://github.com/nodejs/node-addon-api/blob/main/doc/object_wrap.md

> Caution: When the JavaScript object is garbage collected, the call to the C++ destructor may be deferred until a later time. Within that period, Value() will return an empty value.

include/engine/plugins/plugin_base.hpp
whytro: Oh, that's right. I could alternatively make a few adjustments to only apply the passed `default_bearing_radius` if bearings are used in the query - would that be preferable, as it's more inline with the original issue?

test/nodejs/index.js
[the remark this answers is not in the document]
whytro: Okay, I've commented those tests out for now.

include/engine/plugins/plugin_base.hpp
whytro: In this case, should this function be affected by this new default_radius behavior?

https://github.com/Project-OSRM/osrm-backend/blob/192d077adad5bc0d1a5fb9d185c5c4113deabacb/include/engine/plugins/plugin_base.hpp#L166-L171

I've left it alone since it has it's own radius behavior (even for queries not containing bearings), but I'm not sure if this is something the modified use case should affect or not.

src/engine/plugins/viaroute.cpp
```diff
@@ -15,9 +15,12 @@
 namespace osrm::engine::plugins
 {
 
-ViaRoutePlugin::ViaRoutePlugin(int max_locations_viaroute, int max_alternatives)
+ViaRoutePlugin::ViaRoutePlugin(int max_locations_viaroute,
+                               int max_alternatives,
+                               const boost::optional<double> default_radius)
```
SiarheiFedartsou, line 20: If you will make it non-const you will be able to `std::move` it.

src/engine/plugins/viaroute.cpp
```diff
@@ -15,9 +15,12 @@
...
     : max_locations_viaroute(max_locations_viaroute), max_alternatives(max_alternatives)
 {
+    this->default_radius = default_radius;
```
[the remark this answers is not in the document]
whytro: Okay, I've added a constructor for the BasePlugin so that default_radius can be initialized from within the derived class' initializer list.

src/engine/plugins/viaroute.cpp
whytro: Alright - I've taken a closer look at the passing of default_radius and applied copy & move across the board for default_radius.

whytro: A clang-tidy run seems to say that applying std::move to boost::optional<double> is redundant. It's somewhat unexpected for me since although an -O flag in godbolt seems to resolve them the same way, it did yield different results without. I've gone ahead and removed the const qualifier here though nonetheless.

SiarheiFedartsou:
Hey, now I think we are good with this PR in general, but can we try to understand what is wrong with those NodeJs tests failing on macOS? 

I guess the reason for it is simple: we close dataset files in desctructors of C++ object and the time when they are called in NodeJs is undetermined. And I think it is okay for production use cases since people don't create multiple `OSRM` instances. So I would propose trying to increase upper limit for number of file descriptors on CI https://superuser.com/questions/302754/increase-the-maximum-number-of-open-file-descriptors-in-snow-leopard
it probably can be done somewhere here https://github.com/Project-OSRM/osrm-backend/blob/192d077adad5bc0d1a5fb9d185c5c4113deabacb/.github/workflows/osrm-backend.yml#L383

We can check if it works and if it is not, then just create a ticket to resolve it later and merge this PR. WDYT @whytro ? I think you can uncomment those tests back and I can play with fix I proposed(feel free to do it on your own though, but it won't be really convenient - as you can see github asks for manual approval for each your CI run).

SiarheiFedartsou:
> Admittedly, I'm not too familiar on configuring CI workflows,

I didn't mean it would be hard to do for you, I meant it would not very convenient as each time you would have to wait until I push this button :) But thanks for the try :+1:

Screenshot 2023-03-22 at 14 19 08

whytro:
Oh no worries, I understood you there - I just wanted to mean that I'd given it a try but I'm not too sure if the change I made was correct because of my unfamiliarity with it!

SiarheiFedartsou (approved):

SiarheiFedartsou:
> @SiarheiFedartsou Thank you for the guidance! Should I close out the other PR (#6572)?

I think we can keep it for the future: I think we can merge it on the next major version update.

whytro:
Okay! I'll also work on refining it, as there are some things in that PR I've noticed while working on this one that I realized I didn't like, or missed. That way, it'll be in better shape for when the next major version update is being prepped.

nilsnolde:
Thanks for all the work and hanging in there @whytro ! I'd just like to remind you to not forget about the application document for GSoC:) There's still time of course, still wanted to mention it.

whytro:
Thanks! I've just been reading through the guidelines quite a bit regarding the application - I'll probably drop any remaining questions about the scale/expectations of #6558 after a day or two of some additional preliminary research!

nilsnolde:
Feel free to get in touch beforehand, @SiarheiFedartsou and me both have a public email address on Github. We could schedule a getting-to-know-each-other meeting if you want. I'd personally actually prefer that. Then we can also let you know what we'd expect in terms of application, so you don't have to spend too much time on it.

SiarheiFedartsou:
Hey @whytro 
I fixed those Node tests: not sure why we had this `Too many open files` error, but it has gone as soon as I added path to dataset when creating `OSRM` instance. So now I think we are good to merge it.

Thanks for this contribution and looking forward to work with you on our GSoC project 🚀

Issue #6592, opened 2023-04-03
Remove boost::phoenix and migrate from boost::spirit::qi to boost::spirit::x3


mugr1x:
Related #6551
I know @SiarheiFedartsou mentioned elsewhere it's not good style in general to use both boost::optional and std::optional in the same codebase. The removal is delayed due to spirit::qi and phoenix incompatiblity with std::optional. This would involve adpapting the current parsers (3 ones) completely new in x3, but I do have some progress in my fork.
Also for removing phoenix as per docs "Define small unnamed function objects at the actual call site" seems like something one would want to remove for a cleaner codebase and I can see it only being used in conjunction with qi so migrating to x3 would make it obsolete.

mugr1x:
Also I should probably mention I applied for GSoC 2023 (;

Issue #6551, opened 2023-02-23
Refactor codebase to use std::optional instead of boost::optional


SiarheiFedartsou:
# Description

We migrated to C++17 and can start using std::optional from standard library instead of boost::optional.

The main problem here is Boost.Spirit library which doesn’t support std::optional(see https://github.com/Project-OSRM/osrm-backend/pull/6429#issuecomment-1296207761).

Kaushik-Iyer:
Can I work on this?

Kaushik-Iyer:
Can I try to work on this? Which part of the codebase should I be looking at?

nilsnolde:
GSoC has an application process, which we will follow. We’ll make an announcement as soon as that starts.

code-sparsh:
I might be able to work on this. Is it available?

nilsnolde:
It is. Do you want to be assigned?

code-sparsh:
Yes, I want to be assigned.

97littleleaf11:
oops, I already have some progress on this issue

97littleleaf11:
@nilsnolde sorry for my previous misguiding comment. @code-sparsh It's fine that you take this issue! I will find another issue to work on later. Actually I haven't dig into the `Boost.Spirit` yet.

code-sparsh:
@97littleleaf11 Oh okay. If you already have some progress, it's fair that you take the issue because I haven't started yet. 

Regarding the Boost.Spirit, I did some research on it. I think it would require some extra workarounds to be able to support std::optional. That might add some extra boilerplate to the code. I don't know if that's feasible or not, because I am just a beginner in this.

97littleleaf11:
@code-sparsh Thanks for your information! Yeah, someone might add a boilerplate for std::optional like this https://github.com/boostorg/spirit/issues/270#issuecomment-376114135 . Also need to check the correct overloading func. I haven't made it and found it's a little bit beyond my ability. If would be fine that you can take this.

mugr1x:
Hello, I also have some progress on this issue. Can I submit a PR ?

SiarheiFedartsou:
> Hello, I also have some progress on this issue. Can I submit a PR ?

Sure, why not? :)

mugr1x:
Now I found out that boost::spirit::x3 works well with std::optional (and has only minor modifications needed for working). However is it intentional that one implements a whole CSV parser instead of linking against something like libcsv3 ?
[...]

PR #6593, opened 2023-04-03
Refactor: Replace parts of boost::optional with std::optional (#6551,…

mugr1x:
# Issue

Since C++17 added support for 'containers', #6551 targets at replacing current code that uses boost::optional to use std::optional. This PR replaces some parts of the codebase with std::optional, indeed the ones I could convert without having to rewrite some parsers, see #6592 . I try complete these parser, but without that any further migrations to std::optional cause only myriads of compiler errors.

## Tasklist

 - [X] update relevant [Wiki pages](https://github.com/Project-OSRM/osrm-backend/wiki) ==> none edited
 - [X] add tests (see [testing documentation](https://github.com/Project-OSRM/osrm-backend/blob/master/docs/testing.md)) ==> no new functions to be tested
 - [ ] review
 - [ ] adjust for comments

## Requirements / Relations

#6551 
#6592


mugr1x:
Tests fail with segfault.

mugr1x:
Had something to do with DecreaseKey in query_heap.hpp

mugr1x:
Relevant documentation includes:
https://www.boost.org/doc/libs/1_76_0/doc/html/boost/heap/d_ary_heap.html#id-1_3_18_6_2_1_1_1_22_17-bb

include/contractor/contract_excludable_graph.hpp
```diff
@@ -61,10 +61,12 @@ inline auto contractExcludableGraph(ContractorGraph contractor_graph_,
         // Add all non-core edges to container
         {
             auto non_core_edges = toEdges<QueryEdge>(contractor_graph);
-            auto new_end =
-                std::remove_if(non_core_edges.begin(), non_core_edges.end(), [&](const auto &edge) {
-                    return is_shared_core[edge.source] && is_shared_core[edge.target];
-                });
+            auto new_end = std::remove_if(non_core_edges.begin(),
```
SiarheiFedartsou, line 64: Hm, do I get it correctly that most of changes in this PR are format changes. Were they pushed by mistake?

mugr1x: Better now ?

include/util/query_heap.hpp
```diff
@@ -289,7 +289,7 @@ class QueryHeap
         return inserted_nodes[index].node == node;
     }
 
-    boost::optional<HeapNode &> GetHeapNodeIfWasInserted(const NodeID node)
+    std::optional<HeapNode> GetHeapNodeIfWasInserted(NodeID node)
```
SiarheiFedartsou, line 292: I.e. now on each call we will copy HeapNode, right? How do we know that it won’t impact performance? May be worth refactoring it to `HeapNode*` instead(I mean use just pointer instead of optional<HeapNode> here)? From some perspective optional<T&> is just overcomplicated way to express simple pointer…

WDYT?


P.S. I didn’t have a time yet to look over the rest of PR, just something important what I noticed…

SiarheiFedartsou: More than it: here you can see an example when we change contents of node in a heap and that by reference… https://github.com/Project-OSRM/osrm-backend/blob/d51631401eeab6e90fa58b29f2ba8c0f56dc1161/include/engine/routing_algorithms/routing_base_mld.hpp#L331

I.e. if we return it not by reference now most likely existing logic is broken.

mugr1x: I did run my build and it did at least not segfault or got killed by kernel for too much RAM.
To cite from cppreference.com:
"There are no optional references; a program is ill-formed if it instantiates an optional with a reference type."
Search engine would suggest std::reference_wrapper for this task.

SiarheiFedartsou: At the same time as I can see if we just change it to simple pointer instead of optional(I.e. won’t use neither boost::optional nor std::optional) the rest of code which uses this QueryHeap would “just work” without changes we had to make there…

mugr1x: Regarding the other comment std::optional defines an operator->. However in some cases you probably would need to change a deref etc. I'm just testing solely using pointers for these parts. Does not work out of the box, but the idea should be doable.

SiarheiFedartsou: `boost::optional<T&>` seems to be effectively the same as `T*`, isn’t it?

mugr1x: Yes it is. Still forking the main branch I need replace some of the idioms. Do you prefer nullptr or some void coercion ?

SiarheiFedartsou: What is void coercion? You mean `if (ptr == nullptr)` vs `if (!ptr)`? Personally I like first option as more explicit, but I don’t think it would be blocker to merge this PR :)

mugr1x: Actually it's dreadfully simple, memory safety should be fine since it tests. I got too fed up with the nullptr version.
I think you need to approve CI again.

include/util/query_heap.hpp
```diff
@@ -356,13 +356,13 @@ class QueryHeap
         const auto index = node_index.peek_index(node);
         auto &reference = inserted_nodes[index];
         reference.weight = weight;
-        heap.increase(reference.handle, std::make_pair(weight, index));
+        heap.decrease(reference.handle, std::make_pair(weight, index));
```
danpat, line 359: This change is pretty fundamental to how the whole routing algorithm works.  Can you explain exactly why this change is being made here?

danpat (changes requested):
Can you explain why the change from `heap.increase()` to `heap.decrease()` ?  "The old version was wrong because XYZ?" is an acceptable answer, but this is a pretty fundamental change to a core part of the routing algorithm, I think it needs a clear rationale.

mugr1x:
@SiarheiFedartsou @danpat
Can anyone please approve CI again ?

danpat:
@mugr1x You didn't address my question about the changes in `DecreaseKey` in `query_heap.hpp` - why are those changes included?  It's not obvious to me why changing to returning a pointer involves switching from `.increase()` to `.decrease()`.

mugr1x:
Well I linked the docs two times and explained my reasoning. I opened a separate issue #6603 for this but I would really advise you too look into this anyway soonish. As said I fixed a segfault at some point for me.
Also you are completely right it has nothing to do with just returning a pointer in GetHeapNodeIfWasInserted, however this is the same in master, because as Siarhei pointed out boost::optional<T&> is just a disguise for T*.
So the function always returned a pointer nothing is changed by me except the idiom.

mugr1x:
> > Well I linked the docs two times and explained my reasoning. I opened a separate issue #6603 for this but I would really advise you too look into this anyway soonish.
> 
> Can we discuss that in separate PR? This change is serious enough to deserve separate PR... Also if you see that it doesn't work in certain cases it is always a good practice to add test for such case to not regress it in the future.
Indeed

mugr1x:
> > Well I linked the docs two times and explained my reasoning. I opened a separate issue #6603 for this but I would really advise you too look into this anyway soonish.
> 
> Can we discuss that in separate PR? This change is serious enough to deserve separate PR... Also if you see that it doesn't work in certain cases it is always a good practice to add test for such case to not regress it in the future.

Yes

mugr1x:
I just noted that PR #6596 was merged, I had the same fix as that in this PR too. So I just resolved this one with master.

mugr1x:
@danpat Are you happy with the current status ? If so can you close the review or similar.
Also CI fails with a weird error with redifinition of block_counter, which is clearly not in the source.
However I had the same error (with my own but literally equivalent fix as the one merged into master) sometimes.

AlTimofeyev:
Hi @mugr1x , I took a glance at that error you mentioned and it looks like the file where `block_counter` is mentioned in your PR is redefined. Take a glance at your [commits](https://github.com/Project-OSRM/osrm-backend/pull/6593/commits/c42d32620eb97ed1ce637a9eb465d89fde011006). File `range_table.hpp` has two definitions for that variable on lines 84 and 86. It must've happened when you resolved with master (my bad about that).

mugr1x:
@AlTimofeyev Yes, I first merged my own and your fix. You are right. I just did look at my local source not the the diff.
Thank you very much.

mugr1x:
CI shows exactly the same error about block_counter, which is weird since it is clearly only defined once now, also in diff.

SiarheiFedartsou:
@mugr1x may I ask you to create one more empty commit to trigger CI again? I suspect it is some bug in GitHub Actions.

include/util/range_table.hpp
```diff
@@ -79,9 +79,9 @@ template <unsigned BLOCK_SIZE, storage::Ownership Ownership> class RangeTable
         unsigned last_length = 0;
         unsigned lengths_prefix_sum = 0;
         unsigned block_idx = 0;
-        unsigned block_counter = 0;
         BlockT block;
 #ifndef BOOST_ASSERT_IS_VOID
+        unsigned block_counter = 0;
```
danpat, line 84: These changes were merged to master in another PR https://github.com/Project-OSRM/osrm-backend/pull/6596 - they should not also be showing up in _this_ PR.  I suspect some merges have been incorrectly applied or lost on the branch.  This is probably also why CI is failing.

AlTimofeyev: The source code on master branch has `block_counter = 0` on line 85, meanwhile in the commit it is placed on line 84. Could that be what's causing CI to fail?

SiarheiFedartsou: @mugr1x may be try to squash all your commits into single one? Or open another PR? Tbh for me it is not clear why it is happening - looks more like some GH bug.

PR #6599, opened 2023-04-07
Allow -1.0 as unlimited for default_radius value

whytro:
# Issue

Relevant to:
> https://github.com/Project-OSRM/osrm-backend/issues/2983 Require a radius parameter other than unlimited when using bearings

This is an adjustment to PR #6575 to allow -1.0 to be used as an "unlimited" radius for the default radius flag. In effect, this would allow the user to define an unlimited radius default. This is useful for when the default starting behavior might be changed (ie. with #6572), and allow for the following interaction:
 - `-1.0`: unlimited radius range default
 - `unset`: no default radius range, return error when bearings are not accompanied by radiuses
 - `15`: default radius range of 15

Currently this PR sets the default value in the engine_config to be -1.0 in order to avoid being a breaking change, so the default behavior is still unlimited if not set.

## Tasklist

 - [ ] review
 - [ ] adjust for comments

## Requirements / Relations

Related to:
> https://github.com/Project-OSRM/osrm-backend/pull/6575 Add support for a default_radius flag


danpat:
It might be nice to align the UI here with what we support on the API - decimal values >= 0, and the string `unlimited`.  It's fine to use `-1` internally to represent the unlimited case, but it'll be easier to grok if the command-line options match the values users pass as API parameters.

whytro:
> It might be nice to align the UI here with what we support on the API - decimal values >= 0, and the string `unlimited`. It's fine to use `-1` internally to represent the unlimited case, but it'll be easier to grok if the command-line options match the values users pass as API parameters.

@danpat Okay, I've made a change that should allow "unlimited" to be passed and evaluated accordingly for both the `default-radius` flag and `max-matching-radius` flags.

SiarheiFedartsou (approved):
LGTM 
@whytro can you please resolve conflict in CHANGELOG?

PR #6602, opened 2023-04-15
keep libosrm*

### LINKED ISSUE #6601: Keep osrm libs in the docker image
I’d like to keep the OSRM libs in the docker image, that’d be very valuable for consuming projects who attempt to link osrm, see e.g. https://github.com/VROOM-Project/vroom/issues/919#issuecomment-1509718648.

nilsnolde:
fixes #6601 

The headers are also in the image, it'd make sense to keep the libs too. WDYT @SiarheiFedartsou @DennisOSRM ?

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -36,8 +36,7 @@ RUN NPROC=${BUILD_CONCURRENCY:-$(nproc)} && \
     make -j${NPROC} install && \
     cd ../profiles && \
     cp -r * /opt && \
-    strip /usr/local/bin/* && \
-    rm -rf /src /usr/local/lib/libosrm*
+    strip /usr/local/bin/*
 
 
 # Multistage build to reduce image size - https://docs.docker.com/engine/userguide/eng-image/multistage-build/#use-multi-stage-builds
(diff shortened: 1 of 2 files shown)


docker/Dockerfile
```diff
@@ -36,8 +36,7 @@ RUN NPROC=${BUILD_CONCURRENCY:-$(nproc)} && \
     make -j${NPROC} install && \
     cd ../profiles && \
     cp -r * /opt && \
-    strip /usr/local/bin/* && \
-    rm -rf /src /usr/local/lib/libosrm*
```
SiarheiFedartsou, line 40: Hm, but now we don’t remove /src as well. do we need that too? /src contains the whole repo contents as I remember, I.e. it will make resulting image quite big.

WDYT?

I am traveling at the moment and reviewing it from my phone’s screen, so could miss something :)

nilsnolde: Good catch, didn't see that at all :sweat_smile: `src` folder is irrelevant you're right.

nilsnolde: done:)

nilsnolde: although `/src` wouldn't even be copied over into the release image (it's multi-stage), but it'd probably still make the intermediate image quite a bit bigger

nilsnolde:
windows is failing on a known issue for conan's zlib: https://github.com/conan-io/conan-center-index/issues/15325

SiarheiFedartsou (approved):

PR #6429, opened 2022-10-29
Replace boost::unordered_{map/set} with std, also remove code duplication

DennisOSRM:

PR #6611, opened 2023-04-20
Replace boost::optional with std::optional

mugr1x:
# Issue
This PR target at (partially) resolving #6551.
That is exchanging boost::optional for std::optional from C++17.
As mentioned elsewhere (https://github.com/Project-OSRM/osrm-backend/pull/6429#issuecomment-1296207761)
boost::spirit::qi does not work with std::optional, however boost::spirit::x3 does.
This PR does NOT refactor these parts of the codebase. To my knowledge there are three parsers written with ::qi (and ::phoenix) and at least the time validator and CSV parser I was abled to at least work on. (not included in this PR)
However the core API of /server alses uses ::qi and changing this would be a more peculiar if not a bad decision (for now).
Note because of failing CI this is #6593 reopened.

## Tasklist

 - [ x] CHANGELOG.md entry
 - [ x] review
 - [ x] adjust for comments
 - [ x] cherry pick to release branch


src/nodejs/json_v8_renderer.hpp
```diff
@@ -0,0 +1,68 @@
+#ifndef OSRM_BINDINGS_NODE_JSON_V8_RENDERER_HPP
```
mjjbell, line 1: These `src/nodejs` additions should be removed.

src/guidance/intersection_handler.cpp
```diff
@@ -449,7 +449,7 @@ IntersectionHandler::getNextIntersection(const NodeID at, const EdgeID via) cons
     if (intersection_parameters.node == SPECIAL_NODEID ||
         intersection_parameters.edge == SPECIAL_EDGEID)
     {
-        return boost::none;
+        return {};
```
mjjbell, line 452: Any particular reason why we switch between `nullopt` and `{}`?

mjjbell (changes requested):
Please rectify the `src/nodejs` changes. Otherwise, seems fine.

DennisOSRM:
👍🏽

PR #6646, opened 2023-07-04
Fix operator overload on RouteParameters

whytro:
# Issue

This PR targets an error in the operator overload in RouteParameters, which made the `|=` operator nonfunctional. 

## Tasklist

 - [x] review
 - [x] adjust for comments

## Requirements / Relations

There are no requirements or relations relevant to this PR.

### DIFF
--- a/include/engine/api/route_parameters.hpp
+++ b/include/engine/api/route_parameters.hpp
@@ -202,8 +202,8 @@ inline RouteParameters::AnnotationsType operator|(RouteParameters::AnnotationsTy
         static_cast<std::underlying_type_t<RouteParameters::AnnotationsType>>(rhs));
 }
 
-inline RouteParameters::AnnotationsType operator|=(RouteParameters::AnnotationsType lhs,
-                                                   RouteParameters::AnnotationsType rhs)
+inline RouteParameters::AnnotationsType &operator|=(RouteParameters::AnnotationsType &lhs,
+                                                    RouteParameters::AnnotationsType rhs)
 {
     return lhs = lhs | rhs;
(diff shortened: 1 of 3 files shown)


include/engine/api/route_parameters.hpp
```diff
@@ -202,8 +202,8 @@ inline RouteParameters::AnnotationsType operator|(RouteParameters::AnnotationsTy
         static_cast<std::underlying_type_t<RouteParameters::AnnotationsType>>(rhs));
 }
 
-inline RouteParameters::AnnotationsType operator|=(RouteParameters::AnnotationsType lhs,
-                                                   RouteParameters::AnnotationsType rhs)
+inline RouteParameters::AnnotationsType &operator|=(RouteParameters::AnnotationsType &lhs,
+                                                    RouteParameters::AnnotationsType rhs)
```
mjjbell, line 206: Might be worth deleting this function instead. I assume given the bug that it's not in use.

whytro: I can commit a change to remove it instead if it's preferred, but I think that given the use of the `|` operator, it might be more helpful to keep around fixed, since it can help when calculating the AnnotationsType (ie. able to do `lhs |= rhs` rather than `lhs = lhs | rhs`).

mjjbell: Sounds good, thanks for spotting :+1:

mjjbell (approved):
LGTM.
@whytro Can you run `npm audit fix --production` on your branch and commit the changes? 
That will fix CI.

Issue #6145, opened 2021-10-12
Profile: Access `highway=` key in segment function


bda1986:
Within the `process_way()` function in a profile, it is possible to obtain relevant keys using, e.g., `way:get_value_by_key('highway')`.
Is there a similar possibility to do so for the segment function?

I have used a modified version of the `rasterbot.lua` profile to include third-party raster elevation data and would like the slope of a segment to affect duration conditional on the surface or road type. 
For example, the same slope should assign a higher penalty on ways with `highway=track` than on `highway=primary`. 

However, keys/tags such as `highway=primary` are not listed among the available attributes in the `process_segment()` function, as described in [profiles.md](https://github.com/Project-OSRM/osrm-backend/blob/master/docs/profiles.md).
As far as I understand, the segments are subsets or parts of the ways, so the information specific to each way should be available for the segment somehow?

mjjbell:
As the extraction is currently constructed this will be difficult, as the OSM tag context has been discarded by the time the extraction reaches `process_segment`.

Just to understand correctly, you want to _disproportionately_ apply a slope to the speed of a segment depending on the highway type?
E.g. Given a 5mph track and a 70mph motorway with identical uphill slope, you might want to scale the track to 3mph but motorways are scaled down less, so becomes 65mph?

bda1986:
> Just to understand correctly, you want to _disproportionately_ apply a slope to the speed of a segment depending on the highway type? E.g. Given a 5mph track and a 70mph motorway with identical uphill slope, you might want to scale the track to 3mph but motorways are scaled down less, so becomes 65mph?

Exactly, this is what I was trying to achieve.


> As the extraction is currently constructed this will be difficult, as the OSM tag context has been discarded by the time the extraction reaches `process_segment`.

So any information related to the way, for example, the way_id is no longer available in the `process_segment` function? 
Do the segments themselves have some sort of identifier (other than the coordinates) that would make it possible to join external data?

mjjbell:
Had a deeper look.

`process_segment` is called from `ProcessSegment` in the C++ extraction code.

https://github.com/Project-OSRM/osrm-backend/blob/41dda32546399f1dc12af1de41668993de44c7dc/src/extractor/extraction_containers.cpp#L439-L440

At this point, it's iterating over a list of `InternalExtractorEdge` values. In addition to edge coordinates, it also has an embedded `NodeBasedEdgeWithOSM` value.

[...]

PR #6658, opened 2023-07-20
pass flags into process_segment

mattwigway:
# Issue

This partially addresses #6145, by making the extractor edge flags available to `process_segment` as `segment.flags`. This allows accessing road classification, start point, etc. in `process_segment`. I'm using it by tagging bridges/tunnels as non-startpoints and then not applying elevation weighting to them since they are flat (ish) where I am but often cross deep valleys.

I'm not quite sure how to write an automated test for this. Maybe a profile that uses the flags information in a cucumber test?

## Tasklist

 - [ ] review
 - [ ] adjust for comments

## Requirements / Relations

### DIFF
--- a/include/extractor/extraction_segment.hpp
+++ b/include/extractor/extraction_segment.hpp
@@ -12,9 +12,10 @@ struct ExtractionSegment
                       const osrm::util::Coordinate target_,
                       double distance_,
                       double weight_,
-                      double duration_)
+                      double duration_,
+                      const osrm::extractor::NodeBasedEdgeClassification flags_)
         : source(source_), target(target_), distance(distance_), weight(weight_),
-          duration(duration_)
+          duration(duration_), flags(flags_)
     {
     }
 
@@ -23,6 +24,7 @@ struct ExtractionSegment
     const double distance;
     double weight;
     double duration;
+    const osrm::extractor::NodeBasedEdgeClassification flags;
 };
 } // namespace osrm::extractor
 
--- a/include/extractor/extraction_segment.hpp
+++ b/include/extractor/extraction_segment.hpp
@@ -13,7 +13,7 @@ struct ExtractionSegment
                       double distance_,
                       double weight_,
                       double duration_,
-                      const osrm::extractor::NodeBasedEdgeClassification flags_)
+                      const NodeBasedEdgeClassification flags_)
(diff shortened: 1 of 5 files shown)


include/extractor/extraction_segment.hpp
```diff
@@ -12,9 +12,10 @@ struct ExtractionSegment
                       const osrm::util::Coordinate target_,
                       double distance_,
                       double weight_,
-                      double duration_)
+                      double duration_,
+                      const osrm::extractor::NodeBasedEdgeClassification flags_)
```
mjjbell, line 16:
```suggestion
                      const NodeBasedEdgeClassification flags_)
```

include/extractor/extraction_segment.hpp
```diff
@@ -23,6 +24,7 @@ struct ExtractionSegment
     const double distance;
     double weight;
     double duration;
+    const osrm::extractor::NodeBasedEdgeClassification flags;
```
mjjbell, line 27:
```suggestion
    const NodeBasedEdgeClassification flags;
```

mjjbell (changes requested):
> I'm not quite sure how to write an automated test for this. Maybe a profile that uses the flags information in a cucumber test?

Yes, take a look at https://github.com/Project-OSRM/osrm-backend/blob/master/features/options/extract/lua.feature and https://github.com/Project-OSRM/osrm-backend/blob/master/features/options/extract/turn_function.feature.

You could write a test that uses the flags within the profile's `process_segment` function and confirm you get the expected result.

mattwigway:
I'm having some trouble getting the Cucumber tests running to finish this, which seems to have to do the node OSRM library. It looks like it's trying to find it locally, failing, and then trying to grab it from github, which fails as well since it's a non-release version. Do I need to do something to build the Node.js bindings locally?

mattwigway:
Okay, I figured out the Cucumber tests and added a test that confirms that segment flags get passed through to `process_segment`. @mjjbell let me know if anything else is needed and my apologies for the delay.

mjjbell (approved):
LGTM. Please add a changelog entry.

mattwigway:
@mjjbell I added the changelog entry and rebased to avoid merge conflicts in changelog. Thanks!

PR #6666, opened 2023-07-31
Add support for disabling feature datasets

mjjbell:
# Issue

This change adds support for disabling datasets, such that specific files are not loaded into memory when running OSRM. This enables users to not pay the memory cost for features they do not intend to use.

Initially, there are two options:
- `ROUTE_GEOMETRY`, for disabling overview, steps, annotations and waypoints.
- `ROUTE_STEPS`, for disabling steps only.

Attempts to query a feature for which the dataset is disabled will lead to a `DisabledDatasetException` being returned.

Full details are on a new wiki page: https://github.com/Project-OSRM/osrm-backend/wiki/Disabled-Datasets

## Design Choices
The implementation separates the "feature dataset" concept from the actual files loaded into memory. This gives some flexibility in how the data is stored, allowing for changes in the file structure without breaking the user facing API.

The feature datasets to disable are passed as an array argument to `osrm-routed`, `osrm-datastore` and the NodeJS engine configuration. Currently, you would only select at most one option (`ROUTE_GEOMETRY` is a superset of `ROUTE_STEPS`), but the array interface will allow us to add more options in the future whilst maintaining backwards compatibility.

On every data facade call to optional datasets, we check a boolean to see if its loaded. I attempted other solutions that don't perform a check on every call, such as virtual functions and lambdas. These didn't improve performance and were significantly more complicated, so I stuck with the simplest implementation.

## Performance
For a contraction hierarchy dataset, `--disable-feature-dataset ROUTE_GEOMETRY` reduces RAM usage by ~15%.

There appears to be no impact on query performance. This is likely due to the number of data facade calls affected being very small. Will add full query performance results in a subsequent post. 

## Future Direction
Some optional datasets are stored in files with required datasets. One such example is OSM node IDs being stored with coordinates. We could reduce memory usage further when disabling `ROUTE_GEOMETRY` by splitting such files, or refactoring how files are loaded into memory.

## Tasklist

 - [ ] review
 - [ ] adjust for comments

## Requirements / Relations
#5218
#5838
#6045

### DIFF
--- a/include/storage/storage_config.hpp
+++ b/include/storage/storage_config.hpp
@@ -31,9 +31,60 @@ SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 #include <boost/filesystem/path.hpp>
 
 #include "storage/io_config.hpp"
+#include "osrm/datasets.hpp"
+
+#include <set>
 
 namespace osrm::storage
 {
+
+std::istream &operator>>(std::istream &in, FeatureDataset &datasets);
+
+static std::vector<boost::filesystem::path>
+GetRequiredFiles(const std::vector<storage::FeatureDataset> &disabled_feature_dataset)
+{
+    std::set<boost::filesystem::path> required{
+        ".osrm.datasource_names",
+        ".osrm.ebg_nodes",
+        ".osrm.edges",
+        ".osrm.fileIndex",
+        ".osrm.geometry",
+        ".osrm.icd",
+        ".osrm.maneuver_overrides",
+        ".osrm.names",
+        ".osrm.nbg_nodes",
+        ".osrm.properties",
+        ".osrm.ramIndex",
+        ".osrm.timestamp",
+        ".osrm.tld",
+        ".osrm.tls",
+        ".osrm.turn_duration_penalties",
+        ".osrm.turn_weight_penalties",
+    };
+
+    for (const auto &to_disable : disabled_feature_dataset)
+    {
+        switch (to_disable)
+        {
+        case FeatureDataset::ROUTE_STEPS:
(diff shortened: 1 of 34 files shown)


include/storage/storage_config.hpp
```diff
@@ -31,40 +31,79 @@ SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 #include <boost/filesystem/path.hpp>
 
 #include "storage/io_config.hpp"
+#include "osrm/datasets.hpp"
+
+#include "set"
```
SiarheiFedartsou, line 36: Nit.
```suggestion
#include <set>
```

include/util/exception.hpp
```diff
@@ -84,6 +85,32 @@ static_assert(ErrorDescriptions.size() == ErrorCode::__ENDMARKER__,
...
 
+class DisabledDatasetException : public exception
+{
+  public:
+    explicit DisabledDatasetException(const std::string &dataset_)
+        : exception(BuildMessage(dataset_)), dataset(dataset_)
+    {
+    }
+
+    std::string Dataset() const { return dataset; }
```
SiarheiFedartsou, line 96: Nit. 
```suggestion
    const std::string& Dataset() const { return dataset; }
```

SiarheiFedartsou (approved):

mjjbell:
For completeness, I ran performance tests using the same setup as [here](https://github.com/Project-OSRM/osrm-backend/pull/5953#issuecomment-777082921) with the first three coordinate sets.

All requests had the parameters `overview=simplified&annotations=true&steps=true&skip_waypoints=false`, to check if there is any performance penalty for checking if a dataset is loaded (`optional`) in comparison to it always being loaded (`master`).

The results show no change in performance with this PR. Tests with request parameters that disabled geometries and steps had the same outcome.

image
image
image

PR #6713, opened 2023-10-11
Switch to api.openstreetmap.org API host

danieldegroot2:
Use `api.openstreetmap.org/api/` -and HTTPS- instead of `www.openstreetmap.org/api/*`.

(Is: https://github.com/openstreetmap/operations/issues/951)

Does not seem to be used anywhere else than in the debug code, based on a GitHub search within this repo.

cc: @systemed


systemed (approved):
LGTM!

PR #6724, opened 2023-10-23
Fix a bug caused by support OSM traffic signal directions

GitBenjamin:
I try fix a bug caused by support OSM traffic signal directions. With mjjbell's help, I learned a lot. I hope to get @mjjbell help and support.

# Issue
https://github.com/Project-OSRM/osrm-backend/pull/6153#issuecomment-1763937197

- Problem 1: When crossing a directed traffic signal, the directed traffic signal sometimes does not work.
- Problem 2: There is a bug in the duration penalty of traffic signal, and the master branch also supports https://github.com/Project-OSRM/osrm-backend/pull/6339. When **weight_name = duration** or **weight_name = routability**, a crash of the **macth_api** may be triggered. Because the directed traffic signal sometimes does not work in the Edge-expanded Graph, but it does in the "duplicate" edges created for the restriction graph. Finally, the sum of the forwad_weight and reverse_weight is not equal to total_weight. Last but not least, they are not loop.
https://github.com/Project-OSRM/osrm-backend/blob/31e31a63d062fb804f5f4695ed3036ca7a269ead/src/engine/routing_algorithms/routing_base_ch.cpp#L149
And then, it will crash in the line.
https://github.com/Project-OSRM/osrm-backend/blob/31e31a63d062fb804f5f4695ed3036ca7a269ead/include/engine/routing_algorithms/routing_base.hpp#L182
https://github.com/Project-OSRM/osrm-backend/blob/31e31a63d062fb804f5f4695ed3036ca7a269ead/include/engine/routing_algorithms/routing_base.hpp#L184

## Tasklist

 - [x] review
 - [ ] adjust for comments

## Requirements / Relations

https://github.com/Project-OSRM/osrm-backend/pull/6153
https://github.com/Project-OSRM/osrm-backend/pull/6339
https://github.com/Project-OSRM/osrm-backend/pull/6153#issuecomment-1772192537

### DIFF
--- a/src/extractor/graph_compressor.cpp
+++ b/src/extractor/graph_compressor.cpp
@@ -20,7 +20,7 @@ namespace osrm::extractor
 static constexpr int SECOND_TO_DECISECOND = 10;
 
 void GraphCompressor::Compress(const std::unordered_set<NodeID> &barrier_nodes,
-                               const TrafficSignals &traffic_signals,
+                               TrafficSignals &traffic_signals,
                                ScriptingEnvironment &scripting_environment,
                                std::vector<TurnRestriction> &turn_restrictions,
                                std::vector<UnresolvedManeuverOverride> &maneuver_overrides,
@@ -337,6 +337,9 @@ void GraphCompressor::Compress(const std::unordered_set<NodeID> &barrier_nodes,
 
                 // update any involved turn relations
                 turn_path_compressor.Compress(node_u, node_v, node_w);
+
+                // Update traffic signal paths containing compressed node.
+                traffic_signals.Compress(node_u, node_v, node_w);
 
                 // Forward and reversed compressed edge lengths need to match.
                 // Set a dummy empty penalty weight if opposite value exists.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -35,6 +35,7 @@
       - FIXED: Bicycle and foot profiles now don't route on proposed ways [#6615](https://github.com/Project-OSRM/osrm-backend/pull/6615)
     - Routing:
       - FIXED: Fix adding traffic signal penalties during compression [#6419](https://github.com/Project-OSRM/osrm-backend/pull/6419)
+      - FIXED: Correctly handle compressed traffic signals. [#6724](https://github.com/Project-OSRM/osrm-backend/pull/6724)
 # 5.27.1
   - Changes from 5.27.0
     - Misc:
--- a/features/car/traffic_light_penalties.feature
+++ b/features/car/traffic_light_penalties.feature
@@ -32,14 +32,14 @@ Feature: Car - Handle traffic lights
             | l    | traffic_signals |
 
         When I route I should get
-            | from | to | time   | # |
+            | from | to | time   | #                             |
             | 1    | 2  |  11.1s | no turn with no traffic light |
             | 3    | 4  |  13.1s | no turn with traffic light    |
             | g    | j  |  18.7s | turn with no traffic light    |
             | k    | n  |  20.7s | turn with traffic light       |
 
 
-    Scenario: Car - Traffic signal direction
+    Scenario: Car - Traffic signal direction straight
         Given the node map
             """
             a-1-b-2-c
@@ -112,14 +112,14 @@ Feature: Car - Handle traffic lights
 
 
 
-    Scenario: Car - Encounters a traffic light
+    Scenario: Car - Encounters a traffic light direction
         Given the node map
             """
-              a      f      k
-              |      |      |
-            b-c-d  h-g-i  l-m-n
-              |      |      |
-              e      j      o
+              a      f      k      p
+              |      |      |      |
+            b-c-d  h-g-i  l-m-n  q-r-s
+              |      |      |      |
+              e      j      o      t
 
             """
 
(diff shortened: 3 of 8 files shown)


GitBenjamin:
Hi, @SiarheiFedartsou. I added CHANGELOG and removed the comment. Thanks!

src/extractor/graph_compressor.cpp
```diff
@@ -327,6 +330,19 @@ void GraphCompressor::Compress(const std::unordered_set<NodeID> &barrier_nodes,
                                reverse_node_weight_penalty,
                                reverse_node_duration_penalty);
 
+                auto add_expanded_traffic_signal =
+                    [&expanded_traffic_signals,
```
mjjbell, line 334: The change to traffic signals needs to be made in-place, as they're subsequently used as part of the edge-expanded graph generation too. As mentioned in the [other thread](https://github.com/Project-OSRM/osrm-backend/pull/6153#issuecomment-1788694243), we could make a compressor wrapper for traffic signals, much like we do with TurnPaths.

https://github.com/Project-OSRM/osrm-backend/blob/31e31a63d062fb804f5f4695ed3036ca7a269ead/src/extractor/turn_path_compressor.cpp#L36

mjjbell:
It would be beneficial to add minimal test cases to the Cucumber suite that capture the two problems you have identified.

src/extractor/graph_compressor.cpp
GitBenjamin: > The change to traffic signals needs to be made in-place, as they're subsequently used as part of the edge-expanded graph generation too. As mentioned in the [other thread](https://github.com/Project-OSRM/osrm-backend/pull/6153#issuecomment-1788694243), we could make a compressor wrapper for traffic signals, much like we do with TurnPaths.
> 
> https://github.com/Project-OSRM/osrm-backend/blob/31e31a63d062fb804f5f4695ed3036ca7a269ead/src/extractor/turn_path_compressor.cpp#L36

Thank you very much for your guidance :smile:. I will try to make a compressor wrapper for traffic signals, like you do with TurnPaths. I need some time to complete the modification, and add test cases to the Cucumber suite.

GitBenjamin: Hi @mjjbell, I'm sorry that i only started working on this recently. I study the **TurnPathCompressor** class, but i find that the traffic signal problem seems to be different.

--- Related method ---
https://github.com/Project-OSRM/osrm-backend/blob/31e31a63d062fb804f5f4695ed3036ca7a269ead/src/extractor/edge_based_graph_factory.cpp#L622-L624
https://github.com/Project-OSRM/osrm-backend/blob/31e31a63d062fb804f5f4695ed3036ca7a269ead/src/extractor/edge_based_graph_factory.cpp#L630
The **TrafficSignals** info used during the generation of the edge-expanded graph. But the **HasSignal** method is not affected by [src/extractor/graph_compressor.cpp](https://github.com/Project-OSRM/osrm-backend/blob/31e31a63d062fb804f5f4695ed3036ca7a269ead/src/extractor/graph_compressor.cpp). As the **GetLastEdgeSourceID** method returns the original node_id of the segment (even if the node has been compressed), the _from_node_ & _intersection_node_ are the original relation in the OSM nodes. So we don't have to worry about the **HasSignal** losing traffic signal feature due to **_compression_**.

Do we still need to make a compressor wrapper for traffic signals?

GitBenjamin:
> It would be beneficial to add minimal test cases to the Cucumber suite that capture the two problems you have identified.

Hi, @mjjbell. I added minimal test cases to the Cucumber suite, but it only capture one problems. Another problem may require a more complex routing network to catch.😔

src/extractor/graph_compressor.cpp
mjjbell: Yes, well, this is currently working only **because** we're not considering the compression here.

If we start compressing the traffic signals, this becomes simpler, as we will not longer care if it's a trivial edge or not.

For example, this currently works as expected

```
@routing @car @traffic_light
Feature: Car - Handle traffic lights

    Background:
        Given the profile "car"

    Scenario: Car - Traffic signal direction turn with edge compression
        Given the node map
            """
                    d
                    |
                    2
                    |
            a-1-b - c--f
                    |
                    e

            """

        And the ways
            | nodes | highway |
            | abc   | primary |
            | ce    | primary |
            | cd    | primary |
            | ce    | primary |

        And the nodes
            | node | highway         | traffic_signals:direction | # |
            | c    | traffic_signals | forward                   | ambiguous, but happens in OSM

        When I route I should get
            | from | to | time   | weight | #                             |
            | 1    | 2  |  35.1s | 35.1   | turn with traffic light    |
            | 2    | 1  |  29.9s | 29.9   | turn with no traffic light |
```

But it would fail if we correctly compress the traffic signals and don't update the edge based graph factory logic.

The case that is currently not being handled properly is when there is no turn, as shown here
https://github.com/Project-OSRM/osrm-backend/pull/6153#issuecomment-2001979000

mjjbell:
@GitBenjamin I'm going to add the fix to this PR for problem 1.

It's unclear if problem 2 would also be fixed, but we can address that in a separate OR once we have identified a minimal reproduction test case.

mjjbell (approved):

GitBenjamin:
> @GitBenjamin I'm going to add the fix to this PR for problem 1.
> 
> It's unclear if problem 2 would also be fixed, but we can address that in a separate PR once we have identified a minimal reproduction test case.

Problem 2 no longer occurs when I call **[Match API](https://github.com/Project-OSRM/osrm-backend/blob/master/docs/http.md#match-service)** again, after I modify the traffic light compressed logic. I can try to do an analysis of that part of osm data again.

src/extractor/graph_compressor.cpp
GitBenjamin: ok, do you still need me to make a compressor wrapper for traffic signals ？

mjjbell: No, I included it in the traffic signal, as it's simple enough.
https://github.com/Project-OSRM/osrm-backend/blob/8526cc7d45adce9c352897acfd0366211e4f1d5e/include/extractor/traffic_signals.hpp#L23

PR #6726, opened 2023-11-03
Increment feature id in tile speed layer

datwelk:
The tile layer currently contains the same id value of "1" for all linestring features in the speeds layer. There's a comment in the codebase that says the id should be incremented for each feature, but this does not happen for the linestring features. It does happen for the turn restriction features a few lines down.

This PR addresses the issue where linestring feature ids are not incremented.


mjjbell (approved):

mjjbell:
LGTM. Can you add a changelog entry?

datwelk:
@mjjbell Added changelog entry and rebased on master. Do I need to squash both commits into one or is it fine like this?

mjjbell:
> @mjjbell Added changelog entry and rebased on master. Do I need to squash both commits into one or is it fine like this?

Thanks, it's fine, they will be squashed on merge.

mjjbell:
Looks like some node tests are failing, because the underlying PBF varints for the IDs now make it bigger than previously expected. Can you update?

https://github.com/Project-OSRM/osrm-backend/blob/8526cc7d45adce9c352897acfd0366211e4f1d5e/test/nodejs/constants.js#L13

Issue #6733, opened 2023-11-12
Multi maneuove relation bug!


rezashokry:
# Summary
Multiple maneuovre relations are not working. 

# Issue 
I was trying to add some maneuovre relation to the map; When I added 1 maneuovre to a node it worked well. (Sharp right is modified)
Before:
![Screenshot from 2023-11-12 14-51-14](https://github.com/Project-OSRM/osrm-backend/assets/45337530/e15f7722-981b-4e0e-8d75-33a49868f726)

After:
![image](https://github.com/Project-OSRM/osrm-backend/assets/45337530/b966403e-d22a-4ec4-b7b2-fd9ebad13607)

But when I added another relation to an irrelevant node the previous one stops working.

# Reproduce 
Here is the 3 versions of the map and the change set to reproduce the bug. 
- Baseline.osm.pbf is the cropped map (OSM release with no change)
- First_relation_added.osm.pbf is the map which the single maneuovre is added and works fine 
- Both_relation_added.osm.pbf is the map which both maneuovres are added and the first one is not working 
[maps.zip](https://github.com/Project-OSRM/osrm-backend/files/13328205/maps.zip)

OSRM version: 5.27.1 release (I also checked 5.22.0 and the bug still happens)

Origin: 
- latitude: 36.36486
- longitude:  59.53642

Destination:
- latitude: 36.36372
- longitude: 59.53701

First maneuovre Via Node Id: 4460228939
Second maneuovere Via Node id: 779031832

PR #4907, opened 2018-02-22
Renumber node IDs in .osrm.maneuver_overrides

oxidase:
# Issue

Fixes #4906

/cc @chaupow @TheMarex 

## Tasklist

 - [ ] CHANGELOG.md entry ([How to write a changelog entry](http://keepachangelog.com/en/1.0.0/#how))
 - [ ] update relevant [Wiki pages](https://github.com/Project-OSRM/osrm-backend/wiki)
 - [ ] add tests (see [testing documentation](https://github.com/Project-OSRM/osrm-backend/blob/master/docs/testing.md)
 - [x] review
 - [ ] adjust for comments
 - [ ] cherry pick to release branch

## Requirements / Relations

 Link any requirements here. Other pull requests this PR is based on?

PR #6739, opened 2023-11-19
Fix manuever overrides finding bug

rezashokry:
# Issue

I was trying to add some maneuver override relation to the map and I noticed some of them are not working. 
I tried to find out why and I figured out this is because there is a vector of maneuver overrides which is assumed to be sorted to use binary search on it in lookup but its not sorted at the binary search stage.
The vector is already sorted in extract function but after the renumber in partition function the vector becomes unsorted. 

Extraction:
https://github.com/Project-OSRM/osrm-backend/blob/master/src/extractor/edge_based_graph_factory.cpp#L1225

Renumbering in partitioner: (which doesn't maintain the sorted array)
https://github.com/Project-OSRM/osrm-backend/blob/master/src/partitioner/partitioner.cpp#L163C1-L163C1 

Finding maneuver overrides for each node:
https://github.com/Project-OSRM/osrm-backend/blob/master/src/engine/guidance/post_processing.cpp#L582

Binary search on the vector:
https://github.com/Project-OSRM/osrm-backend/blob/master/include/engine/datafacade/contiguous_internalmem_datafacade.hpp#L600

I just simply sort the vector after the rename function before the partition function writes it down.

## Tasklist

 - [ ] review
 - [ ] adjust for comments

## Requirements / Relations

This pull request fixes [this](https://github.com/Project-OSRM/osrm-backend/issues/6733) issue.


mjjbell:
Good spot. Looks like it happened in #4907 
I'll create a test case from your[ original example]([url](https://github.com/Project-OSRM/osrm-backend/issues/6733)) to confirm the fix.

mjjbell:
As this is a bug related to the multi-level partitioning, it's difficult to add a neat Cucumber test case.

However, I've verified the fix using the example PBFs from the issue details - thanks for providing them, it's very helpful.

mjjbell (approved):

mjjbell:
@rezashokry can you add an entry to CHANGELOG.md ?

Issue #6204, opened 2022-02-07
osrm postgis.lua - "attempt to concatenate a function value"


Spfeil2:
I would like to test the routing engine using postgis.lua profiles, so check the osm data against a postgis db.

The problem is, inside way_function, I am not able to use way.id. This is the error I get when I try to print it:
> 
> terminate called after throwing an instance of 'sol::error'
>   what():  lua: error: ../osrm-backend/profiles/postgis.lua:72: attempt to concatenate a function value (field 'id')
> Aborted (core dumped)

I use the following code:
```
function way_function (way, result)
  local highway = way:get_value_by_key('highway')
  if (not highway or highway=='') then
    return 0
  end
```
```
local sql_query = " " ..
    "SELECT SUM(SQRT(area.area)) AS val " ..
    "FROM import.osm_roads way " ..
    "LEFT JOIN import.osm_landusages area ON ST_DWithin(way.geometry, area.geometry, 100) " ..
    "WHERE area.type IN ('industrial') AND way.osm_id=" .. way.id .. " " .. --error see above
    "GROUP BY way.id"`
```
[https://github.com/Project-OSRM/osrm-backend/blob/master/profiles/examples/postgis.lua](url)

when I change `way.id` to `way:id()` in sql-query, I get the following error:

> Segmentation fault (core dumped)

Best,
Simon

PR #6745, opened 2023-11-30
Fix call to way:id() in PostGIS profile example

tumik:
Issue: #6204

## Tasklist

 - [ ] review
 - [ ] adjust for comments

## Requirements / Relations

No requirements or relations.


mjjbell:
In the issue details you mention that it segfaults on calling `way:id()`.
Can you clarify that you intend this to be a fix?

tumik:
I did not experience segfaults with this fix unlike the issue creator. This fix works for me. 

I suspect the segfault mentioned in the issue was unrelated but might be worth it for someone else to try it out too.

mjjbell:
I see. I think the PostGIS example is ancient however, and requires upgrading.

For example, when I run it, fails on an earlier line

```
 traceback:
	profiles/examples/postgis.lua:43: in function 'way_function'
profiles/examples/postgis.lua:43: attempt to index a nil value (field 'tags')
``` 

https://github.com/Project-OSRM/osrm-backend/blob/26d3206c58170fe5ac986e7068d4f8a3ddb87829/profiles/examples/postgis.lua#L48

Probably because the SOL bindings have changed in the last 8 years.

PR #6781, opened 2024-02-02
support rtree preheat

hy05190134:
# Issue

we found a lot of requests timeout when osrm is starting, but after a while the rt decrease, so we want to resolve the problem, and we found the reason is the preheat of rtree, because the leaf node of rtree is stored in the .fileIndex file and 
it use mmap so it will produce disk I/O for preheat

## Tasklist

 - [ ] review
 - [ ] adjust for comments

## Requirements / Relations

### DIFF
--- a/include/util/static_rtree.hpp
+++ b/include/util/static_rtree.hpp
@@ -260,6 +260,7 @@ class StaticRTree
     boost::iostreams::mapped_file_source m_objects_region;
     // This is a view of the EdgeDataT data mmap'd from the .fileIndex file
     util::vector_view<const EdgeDataT> m_objects;
+    EdgeDataT tmp;
 
   public:
     StaticRTree() = default;
@@ -462,6 +463,7 @@ class StaticRTree
         : m_coordinate_list(coordinate_list.data(), coordinate_list.size()),
           m_objects(mmapFile<EdgeDataT>(on_disk_file_name, m_objects_region))
     {
+              preheat();
     }
 
     /**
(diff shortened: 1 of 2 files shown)


danpat:
Hmm.  Why not run a simple `cat map.fileIndex > /dev/null` operation before launching `osrm-routed`?  That would have the same effect of loading all the files pages into the page cache....

hy05190134:
> Hmm. Why not run a simple `cat map.fileIndex > /dev/null` operation before launching `osrm-routed`? That would have the same effect of loading all the files pages into the page cache....

it's right. But I just want others do not sense the process when they first use the osrm-routed and encounter the same problem when started

hy05190134:
@danpat  another question, why the desinger of osrm just leave the .fileIndex using mmap

hy05190134:
> Hmm. Why not run a simple `cat map.fileIndex > /dev/null` operation before launching `osrm-routed`? That would have the same effect of loading all the files pages into the page cache....

DennisOSRM (changes requested):
the preheat is just a linear scan through the data. Seems one can do that using `cat` in any production setup.

fenwuyaoji:
@hy05190134 Hi, thanks for your work. we use your preheat method in our osrm. 
The osrm runs as a pod in k8s, and remote storage is used to save osrm data and mounted to the pod. the storage and k8s cluster are in different regions, leading to high latency, often needing to wait 10 mins until loading succeeds. 

We notice that during this init stage, the CPU usage (20m) is low but iowait (8*25%) & load (2.5) is high. I think this is due to the preheat and mmap. I'm wondering wether there is a way can solve this problem. do you have any advice?

image
image

PR #6842, opened 2024-04-02
Add support for opposite approach request parameter 

### LINKED ISSUE #5616: add approach on the opposite side of the road
This feature request is a continuation of #2201 issue and, specifically, #4039 pull request. Even though #4039 initially implemented three types of approaching a waypoint, it ended just adding `approach` of `curb` and `unrestricted`, because no use case was seen for the `opposite`. In this request, we would like to describe one use case for the `opposite` approach.

The use case is the same as in #5586, where we are solving a VRP problem with large, long trucks. In that problem, trucks generally like to continue in the direction they are going - if a truck approaches a waypoint heading West, then it continues from that waypoint also heading West.

In a routing problem, this may be solvable using `continue_straight_at_waypoint` profile parameter, but this is a VRP problem, meaning that we first have to calculate a distance matrix before solving. For each pair of waypoints, A and B, that matrix should contain the following four values for the distance from A to B:

- distance, if we leave A from the curb side of the road and approach B on the curb side
- distance, if we leave A from the curb side of the road and approach B on the opposite side
- distance, if we leave A from the opposite side of the road and approach B on the curb side
- distance, if we leave A from the opposite side of the road and approach B on the opposite side

With just `curb` and `unrestricted`, it is not possible to calculate this matrix. Besides, distance is not the only thing that matters - another thing that matters is whether the truck is carrying load or is driving empty. For instance, here OSRM by default approaches the red node on the right side using the shortest path:

example

However, it will be doing a pickup on the red node and the intention is to drive with as less load as possible, especially when making u-turns, because u-turns with heavy load are more damaging to the roads. This means that we would rather first drive past the red node, make a u-turn at the special loop, and then approach the red node from the opposite side. This is longer, but that makes driving with load from the red node to the next node shorter.

mjjbell:
# Issue
As outlined in #5616, adding support for restricting the waypoint approach to the `opposite` direction, relative to the input coordinate.

 It was included in the original approaches PR (#4039), but dropped due to lack of support at the time.
 This is beneficial for VRP use-cases in particular, and is a simple enough addition that I think it's worth supporting.
 
 This PR leverages @asaveljevs changes, I've just added some additional docs and test coverage. 

## Tasklist

 - [ ] review
 - [ ] adjust for comments

## Requirements / Relations
Fixes #5616


DennisOSRM (approved):

DennisOSRM:
Looks good, only one minor nitpick.

PR #6859, opened 2024-05-05
Upgrade clang-format to version 15

DennisOSRM:
Version 14 is the version available on ubuntu-latest in Github actions


.github/workflows/osrm-backend.yml
```diff
@@ -88,7 +88,7 @@ jobs:
     - name: Prepare environment
       run: |
         npm ci --ignore-scripts
-        clang-format-10 --version
```
SiarheiFedartsou, line 91: It seems ubuntu-22.04 has clang-format-15. May be worth updating to it as we have to reformat a lot of files anyway? https://github.com/actions/runner-images/blob/main/images/ubuntu/Ubuntu2204-Readme.md

Btw 24.04 seems to be already available as well…

DennisOSRM: Oh great, then I'm amend this PR to go straight to 24.04.

SiarheiFedartsou: I made this conclusion from existence of https://github.com/actions/runner-images/blob/main/images/ubuntu/Ubuntu2404-Readme.md, so not sure it is available in reality :)

DennisOSRM: It seems that 24.04 based runners are not available yet. Sticking to 22.04 for now.

SiarheiFedartsou (approved):

PR #6876, opened 2024-05-10
Optimise path distance calculation in MLD map matching

SiarheiFedartsou:
# Issue

### Benchmark
I used [this benchmark](https://github.com/Project-OSRM/osrm-backend/blob/master/src/benchmarks/match.cpp) -  it is the best what we have now :) 
#### Before
Screenshot 2024-05-10 at 20 18 24

#### After
Screenshot 2024-05-10 at 20 19 27

## Tasklist

 - [ ] review
 - [ ] adjust for comments

## Requirements / Relations

### DIFF
--- a/include/engine/routing_algorithms/routing_base_mld.hpp
+++ b/include/engine/routing_algorithms/routing_base_mld.hpp
@@ -681,7 +681,7 @@ double getNetworkDistance(SearchEngineData<Algorithm> &engine_working_data,
         distance += target_phantom.GetReverseDistance();
     }
 
-    return static_cast<double>(static_cast<float>(distance));
+    return from_alias<double>(distance);
 }
 
 } // namespace osrm::engine::routing_algorithms::mld
--- a/include/engine/routing_algorithms/routing_base_mld.hpp
+++ b/include/engine/routing_algorithms/routing_base_mld.hpp
@@ -650,25 +650,26 @@ double getNetworkDistance(SearchEngineData<Algorithm> &engine_working_data,
         return std::numeric_limits<double>::max();
     }
 
-
+    BOOST_ASSERT(unpacked_nodes.size() >= 1);
 
     EdgeDistance distance = {0.0};
 
-    if (source_phantom.forward_segment_id.id == unpacked_nodes[0])
+    if (source_phantom.forward_segment_id.id == unpacked_nodes.front())
     {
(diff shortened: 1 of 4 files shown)


include/engine/routing_algorithms/routing_base_mld.hpp
```diff
@@ -650,11 +650,38 @@ double getNetworkDistance(SearchEngineData<Algorithm> &engine_working_data,
         return std::numeric_limits<double>::max();
     }
 
-    std::vector<PathData> unpacked_path;
+    BOOST_ASSERT(unpacked_nodes.size() >= 1);
```
SiarheiFedartsou, line 653: I believe we could do similar optimisation for CH, but let's go step by step...

mjjbell: For my understanding, there are two changes?
1. Remove unnecessary annotations irrelevant to distance calculation.
2. Use pre-calculated distances

include/engine/routing_algorithms/routing_base_mld.hpp
```diff
@@ -650,11 +650,38 @@ double getNetworkDistance(SearchEngineData<Algorithm> &engine_working_data,
...
+        BOOST_ASSERT(target_phantom.forward_segment_id.enabled);
+        distance += target_phantom.GetForwardDistance();
+    }
+    else if (target_phantom.reverse_segment_id.id == unpacked_nodes.back())
+    {
+        BOOST_ASSERT(target_phantom.reverse_segment_id.enabled);
+        distance += target_phantom.GetReverseDistance();
+    }
+
+    return static_cast<double>(static_cast<float>(distance));
```
mjjbell, line 684:
```suggestion
    return from_alias<double>(distance);
```

mjjbell (approved):
Just one minor suggestion.

include/engine/routing_algorithms/routing_base_mld.hpp
SiarheiFedartsou: Yes, I think so

SiarheiFedartsou: Actually I think we can make much deeper optimization here: use 1 to N routing algorithm in map matching instead of running getNetworkDistance N times, but let's go with small steps...

PR #6943, opened 2024-06-11
Try to use boost::fast_pool_allocator in QueryHeap

SiarheiFedartsou:
# Issue

## Tasklist

 - [ ] review
 - [ ] adjust for comments

## Requirements / Relations

### DIFF
--- a/include/util/query_heap.hpp
+++ b/include/util/query_heap.hpp
@@ -5,6 +5,7 @@
 #include <boost/heap/d_ary_heap.hpp>
 
 #include <algorithm>
+#include <boost/pool/pool_alloc.hpp>
 #include <cstdint>
 #include <limits>
 #include <map>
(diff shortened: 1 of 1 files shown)


include/customizer/cell_customizer.hpp
```diff
@@ -116,8 +116,8 @@ class CellCustomizer
                    const std::vector<bool> &allowed_nodes,
                    CellMetric &metric) const
     {
-        Heap heap_exemplar(graph.GetNumberOfNodes());
```
SiarheiFedartsou, line 119: It used to work in a way that TBB copied `heap_exemplar` to each thread. QueryHeap started using thread local allocator, so here we had a situation when QueryHeap has PoolAllocator somewhere in its internals and this PoolAllocator has a pointer to its **thread local** memory pool, then we copy this QueryHeap to another thread and now new exemplar of QueryHeap has another PoolAllocator with a pointer to exactly the same memory pool, but now it is used on another thread! 
With such change we guarantee that each thread creates heap for itself on its own.

.github/workflows/osrm-backend.yml
```diff
@@ -653,7 +653,7 @@ jobs:
   benchmarks:
     if: github.event_name == 'pull_request'
     needs: [format-taginfo-docs]
-    runs-on: ubuntu-24.04
```
SiarheiFedartsou, line 656: copy-paste from https://github.com/Project-OSRM/osrm-backend/pull/6975

Just to make benchmark results more stable, I'll remove it before merge.

SiarheiFedartsou:
@DennisOSRM would love to have your opinion here :)

Issue #6706, opened 2023-10-02
evaluation of conditional restriction


ironspock:
# Feature
OSRM should evaluate access:conditional restrictions (e.g. in case of streets, ways are temporairly closed of construction)
   access:conditional = no @ (<startdate>-<enddate>)
   motor_vehicle:conditional = ...
   ...

At the moment a route will calculated, but is possibly no accessible in real life because of mentioned conditional restrictions (refer to screen dump and https://www.openstreetmap.org/way/138541665)
![osrm_issue](https://github.com/Project-OSRM/osrm-backend/assets/68661025/b7c35fe7-7169-4ae2-b7e6-1015fbefeeb3)

Issue #5801, opened 2020-08-17
access:conditional supports


jerrybach:
Hi Team,

I've set the access:conditional tags for a way and expected the vehicle will not be routed via this way. But it seems doesn't work (this issue relating to #5093, I supposed).
If I set the access:conditional = no, it works perfectly.

My steps below:

1) Set the conditional access : access: conditional = no @ (Mo-Su 00:00-23:00) -> for testing purpose
2) Save the osm file into my local , I've checked and found it has been exported successfully
3) osrm-extract with commend: docker run -t -v d:/osrm:/data osrm/osrm-backend osrm-extract -p /data/car.lua /data/customizemap.osm --parse-conditional-restrictions
4) osrm-contract with comment: docker run -t -v d:/osrm:/data osrm/osrm-backend osrm-contract /data/customizemap.osrm --time-zone-file /data/worldtimezones.json --parse-conditionals-from-now 1597488923
5) finally run the osrm-backend : docker run -t -i -p 5000:5000 -v d:/osrm:/data osrm/osrm-backend osrm-routed /data/customizemap.osrm

Everything works perfectly except conditional access.

Can you guys help to advise ? Do I need to modify the car.lua profile or include the "access:conditional" in wayhandlers?

Thank you very much.

Issue #4300, opened 2017-07-17
Do not route over invalid access tags


daniel-j-h:
We currently route over

http://www.openstreetmap.org/way/313005906#map=18/51.51021/-0.12293

which has an invalid tagging scheme - `access=Weekdays only` (should be a `access:conditional=`).

- [demo server](http://map.project-osrm.org/?z=18&center=51.510113%2C-0.122590&loc=51.510125%2C-0.123505&loc=51.510362%2C-0.121799&hl=en&alt=0)
- [debug map](http://map.project-osrm.org/debug/#19.38/51.50998/-0.12311)

Technically this is a data issue - but still we should not route over unknown access tags.

Cucumber test to reproduce

```cucumber
# http://www.openstreetmap.org/way/313005906#map=18/51.51021/-0.12293
Scenario: Car - Unknown access tags
    Then routability should be
        | highway    | access        | bothw |
        | footway    | Weekdays only |       |
```

cc @lily-chai @TheMarex

DennisOSRM:
The decision on this is to ignore _invalid_ access tags since this is mode dependent. There's also about a dozen of tags that are not wrong but would need special case handling to be ignored. See #7539 for more details.

Issue #4231, opened 2017-07-03
Support for conditional access tags


danpat:
With https://github.com/Project-OSRM/osrm-backend/pull/3841, we gained support for conditional turn restrictions.

We should also implement conditional `access:conditional` tags, example:

https://www.openstreetmap.org/way/242346384

daniel-j-h:
@karenzshea can you spec out the tasks that need to be done here?

karenzshea:
We can expand the meaning of our existing conditional restriction flags to include `access=conditional` tags on non-relation features.

- During `way` parsing in extraction, save ways tagged with `access:conditional` to disk [like we do with restrictions](https://github.com/Project-OSRM/osrm-backend/blob/master/src/extractor/extraction_containers.cpp#L639-L659) with data node ids, `is_only` and opening hours.
    - this will probably require a new osrm file, or we'll need to figure out how to make them share the `.osrm.restrictions` file
    - we can maybe genericize the serializer for both restricted turns and ways
- In the update handler (shared by contraction/customization) parse way restrictions and apply them [if the condition is valid](https://github.com/Project-OSRM/osrm-backend/blob/6d78c11fd2b80ae5273c53c8abb2c718d3f54120/src/updater/updater.cpp#L446)
- Tests

A lot of the hard work is already in place, like condition validation and time zone handling, so this shouldn't be as complicated of a lift as conditional turns.

xuruiray:
Do we support conditional access now?

joaommartins:
> Do we support conditional access now?

Still not supported, tested today. 

This feature is referenced by:
https://github.com/Project-OSRM/osrm-backend/issues/4300
https://github.com/Project-OSRM/osrm-backend/issues/5801
https://github.com/Project-OSRM/osrm-backend/issues/6706

PR #6984, opened 2024-07-08
Handle conditional access tags for date ranges

ivarbrek:
# Issue
OSRM currently don't support conditional access tags. This is a very useful feature, especially in the context of roads being closed due to construction work or events. 

Several people have requested this feature (#4231, #4300, #5801, #6706)

This PR takes a simple and lightweight approach to this problem, and handles it through the profile libs.

Note that we here only handle conditional time ranges, and require them to span for more than a week, using the format as described in the [OSM wiki](https://wiki.openstreetmap.org/wiki/Conditional_restrictions), e.g. `motor_vehicle:conditional=no @ 2018 May 22-2018 Oct 7`.

## Tasklist

 - [ ] review
 - [ ] adjust for comments

## Requirements / Relations


SiarheiFedartsou:
@ivarbrek could you please also extend tests? Feel free to ask for a help if it is not clear what to update…

ivarbrek:
I've tried to add a few test cases. 

However, the tests seem to fail in the `conan-linux-release` step, and I'm struggling with proceeding from this point. I've downloaded the test logs, but don't find any clear direction from the logs of `car/access.feature/.../309_car_conditional_restrictions.log`. It's shutting down with SIGINT, but that's similar to all other log files I checked. 

How do I proceed?

SiarheiFedartsou:
@ivarbrek this test works locally, right? Try to bump cache key version here https://github.com/Project-OSRM/osrm-backend/blob/4930d2ef054542851729c4ea5f8b7a61de2e0789/.github/workflows/osrm-backend.yml#L404 Sometimes it helps (you need to change v4 to v5 in 2 places there)

ivarbrek:
Finally got the tests running locally 😊 And think it should be good now! Who can I tag for review @SiarheiFedartsou?

DennisOSRM:
Happy to help with the review

features/car/access.feature
```diff
@@ -305,3 +305,16 @@ Feature: Car - Restricted access
...
             | primary    | customers    |   x   |
+
+    Scenario: Car - Conditional restrictions
+        Then routability should be
+            | highway      | motor_vehicle   |  motor_vehicle:conditional            | bothw |
+            | primary      | yes             |                                       |   x   |
+            | primary      | yes             |  no @ 2002 Jan 7 - 2002 Feb 8         |   x   |
+            | primary      | yes             |  no @ 2002 Jan 07 - 2002 Feb 08       |   x   |
+            | primary      | yes             |  no @ 2090 Jan 7 - 2100 Feb 8         |   x   |
+            | primary      | yes             |  no @ 2020 Jan 7 - 2050 Feb 8         |       |
```
DennisOSRM, line 316: Will this test break in 24 years? 😉

profiles/lib/access_conditional.lua
```diff
@@ -0,0 +1,79 @@
...
+ConditionalAccess = {}
+
+
+local function parse_conditional_access(way, key)
+    local conditional = way:get_value_by_key(key .. ':conditional')
+    if not conditional then
+      return nil
+    end
+
+    -- Examples of conditional tags: "no @ (2018 May 22-2018 Oct 7)
```
DennisOSRM, line 20: Consider aligning the examples for better legibility.

profiles/lib/way_handlers.lua
```diff
@@ -9,6 +9,7 @@ local set_classification = require("lib/guidance").set_classification
 local get_destination = require("lib/destination").get_destination
 local Tags = require('lib/tags')
 local Measure = require("lib/measure")
+local ConditionalAccess = require("lib/access_conditional")
```
DennisOSRM, line 12: Naming is ConditionalAccess vs access_conditional

profiles/lib/access_conditional.lua
```diff
@@ -0,0 +1,79 @@
...
+      if numericYear and numericDay then
+          return os.time({ year = numericYear, month = month, day = numericDay })
+      else
+          return nil
+      end
+    end
+
+    local start_date = parse_date(start_date_str)
+    local end_date = parse_date(end_date_str)
+    local current_date = os.time()
```
DennisOSRM, line 55: This probably needs some explanation that the extraction result depends on the date the extraction is run. Also, the date is not checked once but for each case if I am not mistaken. A little of folks run data updates over night and this may lead to edge cases. 

Consider pulling the date once at the start of extraction. Also, consider implementing an override, ie. supporting the use case of extracting data ahead of time.

features/car/access.feature
ivarbrek: Yep 😄 I can extend the ranges so they don't break that soon, but I don't see any simple way to get completely rid of this problem.

profiles/lib/access_conditional.lua
ivarbrek: Good points 👍 Note that the current suggestion excludes date ranges that span less than a week, partly to avoid weird cases like this. But that choice can be discussed I think. 

Anyways, I'll see if I manage to implement an overrride option in a reasonable way.

ivarbrek: Found a relevant example [here](https://github.com/ivarbrek/osrm-backend/blob/bf2ca0a4ee51b0c993039eda4b9e641c5fb5032e/features/car/conditional_restrictions.feature#L42-L48), will try to implement something similar 👍

features/car/access.feature
ivarbrek: Ah, this can be solved doing something similar to [this](https://github.com/ivarbrek/osrm-backend/blob/bf2ca0a4ee51b0c993039eda4b9e641c5fb5032e/features/car/conditional_restrictions.feature#L42-L48). I'll try to do that.

ivarbrek:
I've looked into this a bit more now, and here's my summary/understanding of the current situation:

**Some background I found out:**
Conditional _turn_ restrictions are already implemented at C++ level in the _contract_ step ([see test here](https://github.com/ivarbrek/osrm-backend/blob/bf2ca0a4ee51b0c993039eda4b9e641c5fb5032e/features/car/conditional_restrictions.feature#L42-L48)). A good thing about this implementation is that the `osrm-contract` command supports arguments to set a custom time for the parsing conditional turn restrictions.


Ideally we'd like the conditional access feature to be easily configurable, and here are the options as I see them:
- Make a date parameter in `car.lua` called `date_used_for_conditional_access`. If `nil`, we use the current date. This option is simple to implement, given that I've already started implementing this in Lua. But on the other hand, it does not provide the best interface for custom configuration.
- Implement in Lua, but make the date configurable through the `osrm extract` command. Requires some C++ coding and understanding of the code, so I would like some help if we decide to do this.
- Implement this in the same manner as conditional turn restrictions (see above). I think this would be the best solution, but I'm not very familiar with C++, so would need a great amount of help on that.

What I don't like about the two latter options is that we duplicate the date input. Would be interested to hear your take on that and your suggested next steps @DennisOSRM and @SiarheiFedartsou.

SiarheiFedartsou:
@ivarbrek having it in osrm-customize/osrm-contract is indeed more preferable. In practice you would need to re-run this logic many times per day for the same dataset and such kind of things are done with these tools in OSRM(e.g. live traffic or conditional turn restrictions you mentioned are implemented using it)

ivarbrek:
Allright, so we are maybe just closing this then?

Issue #6710, opened 2023-10-08
foot and bicycle profiles should support trunk access for some countries.


tombay:
# Issue

[Please describe the issue you are seeing with OSRM.
Images are a good way to illustrate your problem.]
This issue has been discussed - and has been reported  in  Bug report #6701.
 e.g. see
(https://map.project-osrm.org/?z=14&center=-41.230508%2C173.377562&loc=-41.212480%2C173.382840&loc=-41.210172%2C173.395951&hl=en&alt=0&srv=2)

tombay:
I have raised a PR (#6703) that  needs reviewing.

tombay:
I have created a new PR (#7013) that implements support for countries based on https://wiki.openstreetmap.org/wiki/OSM_tags_for_routing/Access_restrictions.
This PR is ready for review - but I see no way to assign  a reviewer %-?

tombay:
@frodrigo. Is this useful to solve the problem

Issue #6701, opened 2023-09-13
routing profile foot.lua results in spurious trails for New Zealand (suspect SH 6)


tombay:
# Issue
I have created a dataset using foot.lua as the profile
This is described in the community forum
https://community.openstreetmap.org/t/issues-with-foot-lua-profile/103765
---------------------------------------------------------------------------------------------------------------
The profile is from git source.
Downloaded new-zealand-latest.osm.pbf. And set up two osrm data trees
new-zealand/car
new-zealand/foot
cd car
osrm-extract -p /opt/osrm/car.opt new-zealand-latest.osm.pbf
osrm-partition new-zealand-latest.osrm
osrm-customize new-zealand-latest.osrm
cd …/foot
osrm-extract -p /opt/osrm/foot.opt new-zealand-latest.osm.pbf
osrm-partition new-zealand-latest.osrm
osrm-customize new-zealand-latest.osrm
If i run osrm-routed -a MLD car/new-zealand-latest - the trail draws correctly
If i run osrm-routed -a MLD foot/new-zealand-latest - the trail draws garbage
---------------------------------------------------------------------------------------------------------------
This can be seen here :
https://map.project-osrm.org/?z=14&center=-41.230508%2C173.377562&loc=-41.212480%2C173.382840&loc=-41.210172%2C173.395951&hl=en&alt=0&srv=2

danpat:
The issue is probably related to default access restrictions.  OSM has a wiki page on it here:

https://wiki.openstreetmap.org/wiki/OSM_tags_for_routing/Access_restrictions

If you look at the tables there, you'll see that different counties have different defaults for different road classes.  The OSRM foot profile was written for one set of rules, and it doesn't do country-by-country differences.

This is absolutely a bug in OSRM, and patches to fix it would be quite welcome.  Honestly, most of the work is probably just in the `.lua` files, enabling/disabling various road classes depending on the country.  You would need to build the routing files with country-level polygon file so that the Lua scripts can know which country the edges are in to change the access rules for the profile you're generating.

https://github.com/Project-OSRM/osrm-backend/issues/6008 is a similar bug report from a couple of years ago.  Obviously, nobody has submitted a patch to fix this yet.  You could be the one!

tombay:
[...]

PR #7013, opened 2024-08-06
Country specific highway access. fixes Project-OSRM / osrm-backend #6710

tombay:
#6710 
An issue https://github.com/Project-OSRM/osrm-backend/issues/6710 has been logged (reopening https://github.com/Project-OSRM/osrm-backend/issues/6701).
Address the issue of highway=trunk (and trunk_link) being blocked for foot and bicycle profiles. Discussed in https://wiki.openstreetmap.org/wiki/Key:motorroad and https://wiki.openstreetmap.org/wiki/OSM_tags_for_routing/Access_restrictions (esp #Alternative_ideas) proposed to address issue via
osrm-backend ... --location-dependent-data
# Tasklist

 - [ ] review
 - [ ] adjust for comments

## Requirements / Relations


tombay (commented):
Initial load

tombay:
Worldwide geojson file.
I have a geojson file containing the 32 countries involved. The file is 43Mb (13.5 gzipped). I have not uploaded it. The data was obtained from osm-boundaries.com. They source their data from openstreetmap.org, and prefer to have a link to them as the source of the data. I also assume that anybody making use of these profiles would only be serving one or two countries.

tombay:
Issues:
1/ Ways that cross country boundaries only have one country name associated.  This means that a way may be included (or dropped) inadvertently. It would overly complicate the solution to address a transient that should be rectified in the subsequent way in the route.
2/ Countries with intrastate differences. Australia has different rules for bicycles on footways. Two states (nsw and vic) have highway="footway" with the default bicycle="no". It is possible to make use of "bicycle_nsw" and "bicycle_vic" as 'Countries", but the assumption is that the correct "bicycle" tags have been addedd to OSM data. 
Austria has a region (tyrol) that has different access for vehicle traffic. If a "countryvehicle" profile is desired this can also be covered by using "tryol" as a country.  
3/ Incomplete and inconsistent data in the access tables. Some tables (e.g. Belarus) have footnotes to describe table entries. Other cases have caveats like "destination" or "service". The general assumption is that if does not say "no" then assume "yes".
As mentioned in "https://wiki.openstreetmap.org/wiki/OSM_tags_for_routing/Access_restrictions" - 
As per 2024, no mainstream routing engine is known to use this page directly to influence routing behavior.

tombay:
Fixes Project-OSRM/osrm-backend #6710

PR #7036, opened 2024-09-26
do not overwrite vehicle maxspeed with higher value, fixes #6979

### LINKED ISSUE #6979: maxspeed on road can lead to OSRM exceeding configured vehicle maxspeeds
# Issue

When working with the "truck-soft" profile I noticed that even when I changed the profile to have a truck go really slow on motorways, this would often not have any impact and the average speed in the computed result would be higher than my configured maximum.

I found out that this bug exists in the standard `car.lua` as well, it's just that it doesn't become apparent because speed limits tend to be below the configured car values. However, if some joker were to tag some unlimited motorways with a maxspeed of 9999999 it would quickly show implausible results.

In my opinion the root of the bug is https://github.com/Project-OSRM/osrm-backend/blob/master/profiles/lib/way_handlers.lua#L440-L446 where a maxspeed on the road unconditionally overwrites anything that was there before.

There is a line in `car.lua` that pulls in the `limit` function from `lib/maxspeed.lua` (https://github.com/Project-OSRM/osrm-backend/blob/master/profiles/car.lua#L11) and it appears that at some point this limit function was supposed to cap the maxspeed at the smaller of two values, however this function is then never used in `car.lua`. I have a hunch that somehow this functionality has been lost in some refactoring.

# Steps to reproduce

Set all road class based speeds in car.lua to 10 (https://github.com/Project-OSRM/osrm-backend/blob/master/profiles/car.lua#L142-L157) and notice how you're still getting long-distance routes with average speeds of far above 50.

# Fix

I have fixed the problem for me by changing the aforementioned lines in way_handlers.lua to
````
  if forward and forward > 0 then
    result.forward_speed = math.min(forward * profile.speed_reduction, result.forward_speed)
  end

  if backward and backward > 0 then
    result.backward_speed = math.min(backward * profile.speed_reduction, result.backward_speed)
  end
````
but it would probably be more elegant to fix the problem by putting the `limit` function to its intended use.

# Related issues

This bug might also be the source of issue  #5684 and part of the answer to #6872.

woodpeck:
# Issue

Fix the overwriting of a lower vehicle-based maxspeed by a higher signposted (edge-based) maxspeed

## Tasklist

 - [ ] review
 - [ ] adjust for comments

## Requirements / Relations

### DIFF
--- a/profiles/lib/way_handlers.lua
+++ b/profiles/lib/way_handlers.lua
@@ -438,11 +438,11 @@ function WayHandlers.maxspeed(profile,way,result,data)
   backward = WayHandlers.parse_maxspeed(backward,profile)
 
   if forward and forward > 0 then
-    result.forward_speed = forward * profile.speed_reduction
+    result.forward_speed = math.min(forward * profile.speed_reduction, result.forward_speed)
   end
 
   if backward and backward > 0 then
-    result.backward_speed = backward * profile.speed_reduction
+    result.backward_speed = math.min(backward * profile.speed_reduction, result.backward_speed)
   end
 end
 


DennisOSRM:
Implemented in separate PR linked above.

PR #7040, opened 2024-09-28
Remove unused bearing::get function

SiarheiFedartsou:
:)


.github/workflows/osrm-backend.yml
```diff
@@ -502,6 +502,13 @@ jobs:
 
         conan config init
         yq eval '.compiler.clang.version += ["18"]' -i "$HOME/.conan/settings.yml"
+    - name: Add Apple-clang 16 to list of Conan compilers # workaround for the issue that Conan 1.x doesn't know about Apple-clang 16
```
SiarheiFedartsou, line 505: I am going to migrate to Conan 2.x to get rid of such issues (1.x version is not supported anymore).

PR #7042, opened 2024-09-29
Migrate to Conan 2.x

SiarheiFedartsou:
Conan 1.x is deprecated and it complicates things more and more...


DennisOSRM:
love the change!

.github/workflows/osrm-backend.yml
```diff
@@ -558,7 +554,7 @@ jobs:
         fi
         popd
     - name: Build example
-      if: ${{ matrix.NODE_PACKAGE_TESTS_ONLY != 'ON' }}
+      if: ${{ matrix.NODE_PACKAGE_TESTS_ONLY != 'ON' && matrix.ENABLE_CONAN != 'ON' }}
```
SiarheiFedartsou, line 557: No a lot of reasons to support building libosrm example with Conan... IMO if we want to use Conan deps there instead of system ones we should make libosrm Conan package itself...

DennisOSRM (approved):
Looking good. Took the liberty to fix the build after merging in latest master branch.

PR #7043, opened 2024-09-29
uncomment "cyclability" weight name

szymon-szym:
# Issue

It looks like in `profile/bicycle.lua` the `weight_name` in properties was unintentionally commented out

Without it set to `cyclability` the `safety_check` is committed when using this profile. Turns' weights are also calculated differently

## Tasklist

 - [ ] review
 - [ ] adjust for comments

### DIFF
--- a/profiles/bicycle.lua
+++ b/profiles/bicycle.lua
@@ -18,8 +18,7 @@ function setup()
     properties = {
       u_turn_penalty                = 20,
       traffic_light_penalty         = 2,
-      --weight_name                   = 'cyclability',
-      weight_name                   = 'duration',
+      weight_name                   = 'cyclability',
       process_call_tagless_node     = false,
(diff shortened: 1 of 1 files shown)


kdziurman (approved):

frodrigo:
From the introduction of the concept `weight` in bicycle profile, it always be `duration` as default.

https://github.com/Project-OSRM/osrm-backend/commit/279f8aabfb#diff-c4b120b8978b0193746a4accb230bf085b89bca0fdda9f17920e056ddaee820fR58

I see this PR as a possible breaking change.

Nevertheless can can still use `cyclability` on you side if you prefer it.

TheMarex:
For context: This was not chosen as the default because it roughly x1.5 the memory usage since you need separate weight and duration values.

Issue #7089, opened 2024-12-31
highway=turning_circle should be used for u-turns


MarcelloPerathoner:
# Feature

OSRM should use nearby turning circles for u-turns.

## Current situation

OSRM ignores perfectly good turning opportunities and instead routes you literally through the woods, like here: (Note that there is no road where OSRM tells you to turn!)

![Screenshot from 2024-12-31 11-47-19](https://github.com/user-attachments/assets/c7995d83-05d0-4ca5-80de-5210bc8a09a5)

https://map.project-osrm.org/?z=18&center=46.562307%2C11.718265&loc=46.562099%2C11.718770&loc=46.562176%2C11.718206&loc=46.562121%2C11.718603&hl=en&alt=0&srv=0

Maybe duplicate of: #620, but that one was closed without proper resolution and I'm not allowed to reopen it.

PR #4828, opened 2018-01-24
Stop signs for the turn function in lua profiles

chaupow:
# Issue

Implementing https://github.com/Project-OSRM/osrm-backend/issues/2652

## Tasklist
 - [ ] CHANGELOG.md entry ([How to write a changelog entry](http://keepachangelog.com/en/1.0.0/#how))
 - [ ] update relevant [Wiki pages](https://github.com/Project-OSRM/osrm-backend/wiki)
 - [ ] add regression / cucumber cases (see docs/testing.md)
 - [ ] review
 - [ ] adjust for comments

PR #7130, opened 2025-03-15
Grand Unified Obstacle Treatment

MarcelloPerathoner:
# Issue

Currently OSRM:

- does not take into account stop signs, give-way signs, and crossings,
- does not take into account traffic calming,
- does not take into account designated turning points,
- uses separate code-paths for traffic signals and barriers.

## This patch

- implements stop signs and give-way signs,
  - understands 'minor' stop signs,
  - can infer stop direction from adjacent edge lengths,
- implements designated turning points,
- allows for multiple obstacles at a single node,
- implements a single code-path for all kinds of obstacles like: traffic signals, stop
  signs, barriers, traffic calming, and turning points,
  - reduces the (already crowded) signature of many functions.

## Related work

Similar to pull requests: #6426, #6153, #4828. Thanks to their authors, who gave many ideas.

Fixes issue: #7089

## Implementation description

A class `ObstacleMap` singleton is injected globally into the LUA scripting environment.
The `ObstacleMap` offers functions to add and query different kinds of 'obstacle' nodes.
(For want of a better name since 'waypoint' was already taken.)

The LUA `process_node()` function (repeatedly) calls `ObstacleMap::add()` to register
discovered obstacles. This function is thread safe. The data is directly stored in the
final place and not moved again after.

A compatibilty layer for `process_node()` has been provided to emulate the current
behaviour: to return `result.barrier` and `result.traffic_signal`.

The LUA `process_turn()` function can call `ObstacleMap::any()` and `ObstacleMap::get()` to
retrieve registered obstacles.

The length of the road has been added to the `ExtractionTurnLeg` LUA table and legs for
the start and target roads have been added to `ExtractionTurn`  (a parameter of `process_turn()`).

## Tasklist

 - [ ] review
 - [ ] adjust for comments


DennisOSRM:
Thanks so much for the pull request. This looks very nice, indeed. I love the simplification and the code, the documentation, and the added tests look well written at first glance. Once CI passes, I will do a deeper review. 

From the first impression, this patch has a high chance of getting merged. Well done. 👍

include/extractor/obstacles.hpp
```diff
@@ -0,0 +1,189 @@
...
+    // is there any obstacle of type 'type' at node 'to' when coming from node 'from'?
+    // pass SPECIAL_NODEID as 'from' to query all obstacles at 'to'
+    // 'type' can be a bitwise-or combination of Obstacle::Type
+    bool any(NodeID from, NodeID to, Obstacle::Type type = Obstacle::Type::All) const;
+
+    // get all obstacles at node 'to' when coming from node 'from'
+    // pass SPECIAL_NODEID as 'from' to get all obstacles at 'to'
+    // 'type' can be a bitwise-or combination of Obstacle::Type
+    using GetType =
+        boost::any_range<Obstacle, boost::forward_traversal_tag, Obstacle, std::ptrdiff_t>;
```
DennisOSRM, line 157: Any chance to implement this without boost?

profiles/lib/obstacles.lua
```diff
@@ -0,0 +1,53 @@
+-- Assigns obstacle value to node as defined by
```
DennisOSRM, line 1:
```suggestion
-- Assigns obstacle value to nodes as defined by
```

profiles/lib/obstacles.lua
```diff
@@ -0,0 +1,53 @@
...
+-- include/extractor/obstacles.hpp
+
+local Obstacles = {}
+
+-- process the obstacles at the given node
+-- note: does not process barriers
+function Obstacles.process_node(profile, node)
+    local highway = node:get_value_by_key("highway")
+    if highway then
+        local typ = obstacle_type[highway]
```
DennisOSRM, line 11: should the variable name be `type`?

src/extractor/obstacles.cpp
```diff
@@ -0,0 +1,180 @@
...
+        }
+    }
+
+    TIMER_STOP(preProcess);
+    log << "ok, after " << TIMER_SEC(preProcess) << "s";
+}
+
+void ObstacleMap::fixupNodes(const NodeIDVector &node_ids)
+{
+    const NodeIDVector::const_iterator begin = node_ids.cbegin();
```
DennisOSRM, line 138: perhaps use the `auto` keyword?

src/extractor/obstacles.cpp
```diff
@@ -0,0 +1,180 @@
...
+    const NodeIDVector::const_iterator end = node_ids.cend();
+
+    auto osm_to_internal = [&](const OSMNodeID &osm_node) -> NodeID
+    {
+        if (osm_node == SPECIAL_OSM_NODEID)
```
DennisOSRM, line 143: please consider putty curly braces even on one-line if statements

DennisOSRM (commented):
Looks pretty good to me. Left some minor comments that should be easy to address. The only bigger question is around the use of the `extern` keyword which should be avoided. Again, thanks so much for the contribution.

docs/src/content.js
```diff
@@ -11,6 +11,6 @@ var fs = require('fs');
  */
 module.exports =
       '# HTTP API\n' +
-      fs.readFileSync('./content/http.md', 'utf8') + '\n'+
-      '# libosrm C++ API\n' +
-      fs.readFileSync('./content/libosrm.md', 'utf8') + '\n';
+      fs.readFileSync('./content/http.md', 'utf8') + '\n'
+      // '# libosrm C++ API\n' +
```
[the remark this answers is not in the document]
MarcelloPerathoner: libosrm.md is gone and that line fails the CI. Was libosrm.md removed intentionally? There still are links to it from the README.md

include/extractor/obstacles.hpp
```diff
@@ -0,0 +1,189 @@
...
+// Note: A better name would have been 'WayPoint', but that name is ambiguous in the
+// OSRM-context.
+//
+struct Obstacle
+{
+    // clang-format off
+
+    // The type of an obstacle
+    // Note: must be kept in sync with scripting_environment_lua.cpp
+    enum class Type : uint16_t
```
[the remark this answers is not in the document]
MarcelloPerathoner: I can try. But then it is the same story with all enums passed to sol / LUA.

include/extractor/obstacles.hpp
MarcelloPerathoner: I can try

profiles/lib/obstacles.lua
MarcelloPerathoner: I thought `type` was reserved in LUA, but it is not. Will change.

include/extractor/obstacles.hpp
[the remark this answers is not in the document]
DennisOSRM: yeah, then let's not overthink it. We can leave the enum as is

DennisOSRM (approved):
👍 well done

include/extractor/extraction_turn.hpp
```diff
@@ -13,43 +15,79 @@ namespace osrm::extractor
...
     int highway_turn_classification;
     int access_turn_classification;
     int speed;
+    double distance;
     RoadPriorityClass::Enum priority_class;
     bool is_incoming;
     bool is_outgoing;
 };
 
+extern const std::vector<ExtractionTurnLeg> NO_OTHER_ROADS;
```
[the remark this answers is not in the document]
MarcelloPerathoner: Removed.

include/extractor/obstacles.hpp
```diff
@@ -0,0 +1,189 @@
...
+//
+// This also represents a good turning point like a mini_roundabout, turning_loop, or
+// turning_circle.
+//
+// Note: A better name would have been 'WayPoint', but that name is ambiguous in the
+// OSRM-context.
+//
+struct Obstacle
+{
+    // clang-format off
```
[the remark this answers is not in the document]
MarcelloPerathoner: Fixed.

include/extractor/obstacles.hpp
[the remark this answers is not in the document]
MarcelloPerathoner: I was able to pull it into the obstacles.cpp file but not into the header. In the header I had to use a constexpr, but the Microsoft compiler choked on it.

profiles/car.lua
```diff
@@ -476,8 +474,22 @@ function process_turn(profile, turn)
...
 
-  if turn.has_traffic_light then
-      turn.duration = profile.properties.traffic_light_penalty
+  for _, obs in pairs(obstacle_map:get(turn.from, turn.via)) do
+    -- disregard a minor stop if entering by the major road
+    if obs.type == obstacle_type.stop_minor and not Obstacles.entering_by_minor_road(turn) then
+        goto skip
+    end
+    -- heuristic to infer the direction of a stop without an explicit direction tag
+    -- rationale: a stop sign should not be placed farther than 10m from the intersection
```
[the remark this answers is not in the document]
MarcelloPerathoner: Ok. Changed.

profiles/car.lua
```diff
@@ -476,8 +474,22 @@ function process_turn(profile, turn)
   local turn_penalty = profile.turn_penalty
   local turn_bias = turn.is_left_hand_driving and 1. / profile.turn_bias or profile.turn_bias
 
-  if turn.has_traffic_light then
-      turn.duration = profile.properties.traffic_light_penalty
+  for _, obs in pairs(obstacle_map:get(turn.from, turn.via)) do
+    -- disregard a minor stop if entering by the major road
+    if obs.type == obstacle_type.stop_minor and not Obstacles.entering_by_minor_road(turn) then
```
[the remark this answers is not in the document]
MarcelloPerathoner: Done.

include/extractor/obstacles.hpp
MarcelloPerathoner: sol does not recognize a std::ranges::subrange as a container. It did recognize a boost::any_range though. As a compromise I used a std::vector hoping the performance hit is not too bad.

src/extractor/scripting_environment_lua.cpp
```diff
@@ -99,11 +99,12 @@ void handle_lua_error(const sol::protected_function_result &luares)
     const auto msg = luaerr.what();
     if (msg != nullptr)
     {
-        std::cerr << msg << "\n";
```
MarcelloPerathoner, line 102: I think this is why some tests are flaky. `std::cerr` is not thread-safe.

Issue #7016, opened 2024-08-16
Json_render is not accurate for osm node ID


fenwuyaoji:
# Issue

@Rejudge-F  @SiarheiFedartsou Hi, all. We are facing an inaccurate OSM node ID format issue and I found it's related to https://github.com/Project-OSRM/osrm-backend/pull/6531.

The case is:
original node ID: 11117421192, after format: 1.111742119e+10.

The reason I think is that `PackedOSMIDs` is 34 bits and the max length of it is 11. Why do we limit the number length to less than 10? I noticed that before https://github.com/Project-OSRM/osrm-backend/pull/6531, we only limited the decimal to less than 10, but now we limit the whole number. should we rollback to the previous version or just relax the length to 11?

# Steps to reproduce
Add below test case to `unit_tests/util/json_render.cpp`.
```
    output.clear();
    renderer(11117421192);
    BOOST_CHECK_EQUAL(output, "1.111742119e+10");
```

Rejudge-F:
If a rollback is performed, it will cause more serious problems, resulting in incorrect map match results. In addition, I found that the ID type used by OSRM elsewhere is int32, with a maximum of 10 digits. If only this place is modified, there seem to be problems elsewhere.
If you modify the type used by osrm, it would be a relatively large project, so I recommend that you could shrink the ID within the int32 range.

fenwuyaoji:
thanks for your quick response. I see your point. how about just relax it to 11? I think this part is only used for json format, right? then nothing will be affected I think.

Rejudge-F:
@fenwuyaoji I think you can fork this repository and modify it, but I don’t recommend doing so. Many parts of the code use the INT32 type for OSM IDs. Therefore, I suggest that you keep your IDs within the INT32 range and then use OSRM to process the data.
By the way, whether it’s 10-digit or 11-digit IDs, neither seems to be an ideal solution. If you can find a more universal way to solve this problem, that would be the best.

fenwuyaoji:
okay, I will look through the code and see whether I can do anything better to solve it.

fenwuyaoji:
[...]

Issue #6758, opened 2023-12-31
Truncated OSM node IDs on Raspberry Pi 4 with 64bit OS


matlen:
# Issue

I installed osrm-backend v5.28.0 on my Raspberry Pi 4 running Debian Bullseye in the 64 bit version. So in theory, the system should be able to deal with OSM IDs higher than 4294967295. However, I noticed that the IDs get truncated anyway. Is there anything I can do in the compilation to improve this?

# Steps to reproduce

An example is the query https://routing.openstreetmap.de/routed-bike/route/v1/driving/9.98356,53.59720;9.98357,53.59614?alternatives=true&overview=full&steps=true&annotations=true, which yields the following output. Pay attention to the OSM node ID 11455378494 (printed bolt). 

{"code":"Ok","routes":[{"geometry":"oecfIq~|{@jA?Lb@xBF","legs":[{"steps":[],"summary":"","weight":48.7,"duration":48.7,"annotation":{"metadata":{"datasource_names":["lua profile"]},"datasources":[0,0,0],"weight":[10.3,6.3,30.5],"nodes":[1490604279,1492031044,**11455378494**,5594295924],"distance":[42.738124380,13.967053929,67.826637997],"duration":[10.3,6.3,30.5],"speed":[4.1,2.2,2.2]},"distance":124.5}],"weight_name":"duration","weight":48.7,"duration":48.7,"distance":124.5}],"waypoints":[{"hint":"Un8ugO4W0oQjAAAAZwAAAOgAAAAAAAAANQ9rQdfzKkJVVodBAAAAACMAAABnAAAA6AAAAAAAAACSEAAAtVeYABDUMQNIVpgAENQxAwMAPwuDikf9","distance":24.178288347,"name":"Tarpenbekstraße","location":[9.983925,53.5972]},{"hint":"8BbShPO8Xo4xAQAAhgAAAD8AAAAAAAAAPaeHQuz9bEEOeV9BAAAAADEBAACGAAAAPwAAAAAAAACSEAAA4VaYAOnPMQNSVpgA7M8xAwEAbwuDikf9","distance":9.478472376,"name":"","location":[9.983713,53.596137]}]}

When I run the same query on my Raspi, the result looks like this. The OSM ID got truncated to 1.145537849e+10, which is 11455378490.

[...]

Issue #6594, opened 2023-04-04
node id is strange from annotation of route response 


aibear-huang:
i have some osm file
and i build one of them and start a osrm-routed service
and i got  some nodes from a route request , and the result seems all right.

and then i use osmconvert to combine all osm to pbf, and then i build this pbf and start a osrm-routed service
and i got some strange nodes, 
[1.021109170,1.437227874,5243432333480608,9252979683381472,8041568168893585,1.592471774,1.132044417,1.785935019,1.592109612,1187344792961079,687335956611697,1.749107293]
there are float node id in the array. and other ids is very big then my expect id. 
i change the osrm-extract code :osrm-backend/include/extractor/packed_osm_ids.hpp  i change the osrmid bit from 34 to 40
because my max node id  is 4e11

so i want to know why the node id from annotation is so strange?
is something unexpected happened in osmconvert? or osrm-extract?

thanks for your reply~
osrm version is 5.27.1

PR #7142, opened 2025-03-31
Fix json rendering of large osm ids

### LINKED ISSUE #7069: Node IDs > 2^34 are reported by route matching incorrect.
# Issue

Ran into an issue today with a client that is using an instance of OSRM to compute routes on a mix of data from OSM (and thus valid IDs) and data from a local source (which was given arbitrary IDs > 1quadrillion to avoid conflicts, we couldn't renumber everything right now). When performing a route match request against the higher node IDs the reported route coordinates were correct, but the node IDs were not. Upon further investigation it appeared that the reported IDs were overflowing beyond 2^34.

# Steps to reproduce

Start an OSRM match server with some node IDs higher than 2^34. Make a route match request (eg. v1/driving). Check the reported node IDs.

osrm-backend version: 5.26.0

# Specifications

- Unknown server the client operates. It's linux, some VM from a cloud provider.

# The good news

We have like 1600 days to resolve this before actual OSM data causes this issue.

MarcelloPerathoner:
# Issue

Fixes #7069 #7016 #6758 #6594 #5716

This patch introduces a new class for OSM ids to the JSON stack. 

The current JSON stack rounds OSM ids to a precision of 10 digits, which is insufficient to represent current OSM node ids.

Also, the current JSON stack converts OSM ids from `uint64_t` to `double`. The `double` type has ~53bits of accuracy, the rest going into exponent and sign, which are wasted here. Why is this important even if OSM is still far away from using 53 bits for node ids? Because some people use high ids for their own private data to avoid conflicts with the ids in the OSM database. See: #7069

This patch stores OSM ids into `uint64_t` and provides a different output format thus guaranteeing correct output under all circumstances.

This pull request is an alternative to:  #7096 

## Tasklist

 - [X] add tests (see [testing documentation](https://github.com/Project-OSRM/osrm-backend/blob/master/docs/testing.md))
 - [ ] review
 - [ ] adjust for comments

### DIFF
--- a/include/engine/api/nearest_api.hpp
+++ b/include/engine/api/nearest_api.hpp
@@ -65,7 +65,8 @@ class NearestAPI final : public BaseAPI
                                auto &phantom_node = phantom_with_distance.phantom_node;
 
                                auto node_values = MakeNodes(phantom_node);
-                               fbresult::Uint64Pair nodes{node_values.first, node_values.second};
+                               fbresult::Uint64Pair nodes{from_alias<uint64_t>(node_values.first),
+                                                          from_alias<uint64_t>(node_values.second)};
 
                                auto waypoint = MakeWaypoint(&fb_result, {phantom_node});
                                waypoint->add_nodes(&nodes);
@@ -123,10 +124,10 @@ class NearestAPI final : public BaseAPI
     const NearestParameters &parameters;
 
   protected:
-    std::pair<uint64_t, uint64_t> MakeNodes(const PhantomNode &phantom_node) const
+    std::pair<OSMNodeID, OSMNodeID> MakeNodes(const PhantomNode &phantom_node) const
     {
-        std::uint64_t from_node = 0;
-        std::uint64_t to_node = 0;
+        OSMNodeID from_node = OSMNodeID{0};
+        OSMNodeID to_node = OSMNodeID{0};
 
         datafacade::BaseDataFacade::NodeForwardRange forward_geometry;
         if (phantom_node.forward_segment_id.enabled)
@@ -135,9 +136,8 @@ class NearestAPI final : public BaseAPI
             const auto geometry_id = facade.GetGeometryIndex(segment_id).id;
             forward_geometry = facade.GetUncompressedForwardGeometry(geometry_id);
 
-            auto osm_node_id =
+            to_node =
                 facade.GetOSMNodeIDOfNode(forward_geometry[phantom_node.fwd_segment_position]);
-            to_node = static_cast<std::uint64_t>(osm_node_id);
         }
 
         if (phantom_node.reverse_segment_id.enabled)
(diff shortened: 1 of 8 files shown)


imannamdari:
Why not separate integers from doubles instead of introducing a new type (osm_id) in the JSON renderer package? Integers and doubles are well-defined types in general-purpose packages like JSON rendering, which should not have knowledge of osm ids. Otherwise, there would be no limit to custom-defined classes creeping into it.

Other popular JSON rendering frameworks also handle rendering based on types, as seen here:
[protobuf-go protojson encoding](https://github.com/protocolbuffers/protobuf-go/blob/master/encoding/protojson/encode.go#L296)

In my opinion, we should have separate Double and Integer classes instead of a single Number class, and we should not define an osm_id class within the JSON renderer, as it falls outside its scope.

MarcelloPerathoner:
Because doubles implicitly convert to integer so there would be conflicts between Json::Number and Json::Integer. Even if you fix it with templates and requires, you will still mess up a lot of tests that expect Number and not Integer.

imannamdari:
The point is that we should explicitly specify the type. If the value is an integer, we should use json::Integer; if it’s a double, we should use json::Float.

For example, in this line:
https://github.com/Project-OSRM/osrm-backend/blob/master/include/engine/api/nearest_api.hpp#L107
we should use emplace_back(json::Integer(node_values.first)) instead of just emplace_back(node_values.first), which leaves it up to the compiler to decide the type—potentially causing ambiguity between the two.

MarcelloPerathoner:
We should have done that. But programmers are lazy and the existing code relies on the compiler to select the right type. If we require explicit construction we'd have to change the code in a lot of places. All construction of Json::Value and all assignments to Json::Value will have to be changed.

plecompt446:
@MarcelloPerathoner and @imannamdari do you guys have an idea on whether this is incoorperated in the v6.0.0 version or not? I use the image [ghcr.io/project-osrm/osrm-backend:v6.0.0](http://ghcr.io/project-osrm/osrm-backend:v6.0.0) but get the issue that my id's appear in scientific notation

MarcelloPerathoner:
> @MarcelloPerathoner and @imannamdari do you guys have an idea on whether this is incoorperated in the v6.0.0 version or not? I use the image [ghcr.io/project-osrm/osrm-backend:v6.0.0](http://ghcr.io/project-osrm/osrm-backend:v6.0.0) but get the issue that my id's appear in scientific notation

AFAIK it is not included in 6.0.0.

plecompt446:
@MarcelloPerathoner that is too bad. I hoped it would have been and that someohow I missed something. Do you maybe know a docker image that does not ahve this issue?

pinkhamr:
I ran into this issue as well. It seems like there are many nodes now with >10 digits Any chance we'll see the fix in the next release?

TheMarex:
Also solved by https://github.com/Project-OSRM/osrm-backend/pull/7311

PR #7144, opened 2025-04-03
Fix initialization of tabulation hash

DennisOSRM:


TheMarex:
Have you benchmarked this? If I remember correctly the (technically) incorrect initialization actually yields better result for the particular use-case because there is such a high spatial correlation between node IDs in the graph that it is close to sequential. The MSB part will be basically fixed and most of the entropy comes from the LSB part. With random initialization the LSB part will have duplicates, so we are creating _more_ collisions with the same amount of bits.

DennisOSRM:
> Have you benchmarked this? If I remember correctly the (technically) incorrect initialization actually yields better result for the particular use-case because there is such a high spatial correlation between node IDs in the graph that it is close to sequential. The MSB part will be basically fixed and most of the entropy comes from the LSB part. With random initialization the LSB part will have duplicates, so we are creating _more_ collisions with the same amount of bits.

Now that you mention that, it rings a bell. Let me dig into my files if I can rediscover the benchmarks from way back when.

Issue #7146, opened 2025-04-10
Does not route on (the edge of) multipolygon area highways.


dreirund:
# Issue

OSRM does route on simple polygon area highways (to be precise, on their edge). But if it is a [multipolygon](https://wiki.openstreetmap.org/wiki/Relation:multipolygon), it does not even route on their edge.

# Steps to reproduce

1. Open [&nearr;&hairsp;this routing URL](https://map.project-osrm.org/?z=18&center=51.103596%2C13.770668&loc=51.103101%2C13.770123&loc=51.103626%2C13.769329&hl=en&alt=0&srv=2)
2. Observe that, instead of snapping to the closest edge of a footway, the area highway defined as multipolygon ([&nearr;&hairsp;this entity in OSM](https://www.openstreetmap.org/relation/14448718)), it snaps to an area highway further away ([&nearr;&hairsp;this entity in OSM](https://www.openstreetmap.org/way/217916783)) + resulting in a detour.  
   Screenshot:  
   ![Image](https://github.com/user-attachments/assets/94e6b377-8c7b-4a03-8b96-24e3b4bcc436)

# Further remarks

Similar issue for [OpenRouteService](https://github.com/GIScience/openrouteservice/issues/1944) and [GraphHopper](https://github.com/graphhopper/graphhopper/issues/3104).

Regards!

dreirund:
Addendum:

According to [this comment](https://community.openstreetmap.org/t/128765/6), the area in question is a correctly defined multipolygon area and thus should be handled the same way as single-polygon area highways.

mvl22:
> According to this comment, the area in question is a correctly defined multipolygon area and thus should be handled the same way as single-polygon area highways.

No, the problem is that the **Way** has no data on it, so there is nothing to route over. Data about physical real-world objects does not go on a Relation. This is not a router bug: it is simply incorrect data. Just because an OSM editor allows you to write arbitrary tags into a Relation doesn't mean anything will be done with that data.

To fix this:

Edit this Way, which is the actual physical object:
https://www.openstreetmap.org/way/1085465662

Add the four infrastructure tags currently on the Relation:

```
highway=footway
lit=yes
smoothness=intermediate
surface=compacted
```

The remove those tags from the Relation.

dreirund:
> No, the problem is that the Way has no data on it, so there is nothing to route over. Data about physical real-world objects does not go on a Relation.

[...]

Issue #2655, opened 2016-07-13
Enable routing over multipolygons (needed for pedestrian routing)


species:
Currently, when a pedestrian area has their highway=tags set on the relation and not on the outer ways, routing is not even possible on the boundaries of the area.

Please enable a mechanism to 'copy' all relation tags to all member ways, so at least routing around the edges of multipolygon areas would be possible.

danpat:
@species Patches welcome!

emiltin:
see #482 for discussion on processing relations

answerquest:
hello, any updates on this? I looked over the linked issue.. 
I marked a polygon which is accessible to pedestrians with `highway=pedestrian` and `area=yes`. That opened it up for routing, but the osrm router is sticking to the boundaries.

ref: https://www.openstreetmap.org/directions?engine=fossgis_osrm_foot&route=12.97990%2C80.24194%3B12.97861%2C80.24100#map=18/12.97913/80.24143

systemed:
This is #64 and unlikely to be solved unless someone steps up and provides code, I suspect.

ManDay:
I think this is more related to #482 (which is closed w/ fixed) than #64, routing *through* squares is not even asked for. Just that areas defined by multipolygon rels behave like their equivalent area counterparts.

OSRM routes *along* the `highway=pedestrian; area=yes` **area** but not *along* the `highway=pedestrian; area=yes` **relation** ([e.g.](https://www.openstreetmap.org/directions?engine=graphhopper_foot&route=42.01827%2C23.07887%3B42.01896%2C23.07849)). Shouldn't this already be fixed as by closure of #482?

dreua:
I find it quite confusing that the routing along the edges works for regular areas but as soon as it is a multipolygon it is just not routeable at all. Using multipolygons to cut out a part of the area is nowadays quite common, for example if there is a small building (like a kiosk) or a large fountain on a pedestrian area or a steps leading to an underground tunnel on a railway platform (Mainz Hbf has this on an all platforms).

PR #7161, opened 2025-05-01
feat: open area routing

### LINKED ISSUE #64: Route through squares (area=yes)
handling square might not be trivial, and are probably mostly important for pedestrian routing. so this might be for a future version.

the thing with a square is that you can cross it in any way. unfortunately squares can have odd shapes, so it's not always possible go in a straight line from where you enter to where you leave. for example a building might be in the way if the square has a hour-glass shape.

a simple example is "Prins Jørgens Gård" is a square, tagged with highway=pedestrian, area=yes:
http://www.openstreetmap.org/?lat=55.676763&lon=12.579427&zoom=18&layers=M

if i set up a pedestrain speedprofile, which includes 'pedestrian=5', osrm will route at the edge of the square. this is will get me across the square, but in a somewhat odd fashion.

one option would be to add routes connecting all entries to the square to all others. another would be to connect all entries to a central point. perhaps these to options could be improved by checking that routes are only added if they don't cross the square boundary.

a third option would be be still route along the edge of the square, but insert the route some reasonably distance from the edge, since in reality you cant walk right at the edge of the building.

MarcelloPerathoner:
This PR provides routing over areas. 

It builds a visibility graph of the area, then runs dijkstra from each entry point to reduce the number of edges, then adds the remaining edges as ways to the router.

&nbsp;&nbsp;

From the OSM site, without area routing, following mapped-for-the-router ways:

Screenshot from 2025-05-22 10-56-47

With area routing, also note the better directions:

Screenshot from 2025-05-22 10-55-44

[Documentation](/Project-OSRM/osrm-backend/blob/area-routing/docs/areas.md)

# Issue

Fixes #64 #2655 #7146

## Tasklist

 - [X] add tests (see [testing documentation](https://github.com/Project-OSRM/osrm-backend/blob/master/docs/testing.md))
 - [ ] review
 - [ ] adjust for comments

## TODO

- entrances to buildings along the perimeter of the area,
- PT stations in the middle of the area,
- alternate algorithms, eg. retraction.

---

The two follow-up commits on this branch (`8578e7aee`, `e2c57a772`) were written with AI
assistance: Claude Code, Claude Opus 5 🤖

Known defects and the review points deferred to follow-ups are tracked in #7683.


DennisOSRM:
cool feature!

seems the linked screenshots don't show tho in the PR

MarcelloPerathoner:
> seems the linked screenshots don't show tho in the PR

I re-uploaded them. Do they show now?

DennisOSRM:
> > seems the linked screenshots don't show tho in the PR
> 
> I re-uploaded them. Do they show now?

They do. Thanks a lot

MarcelloPerathoner:
Do we have a general dislike for boost? Just asking ...

DennisOSRM:
The good news is that heap implementations already exist in our code base:
 - include/util/binary_heap.hpp
 - include/util/d_ary_heap.hpp

DennisOSRM:
just tweaked the settings s.t. GitHub Actions should now run automatically on push.

MarcelloPerathoner:
I have now implemented Dijkstra and a binary heap that works with it. The d-ary heap in `include/util/d_ary_heap.hpp` looks like it could be coaxed into service too, but it is completely undocumented. 

Shouldn't we drop / update the macos-x64 build since it is based on apple-clang-14? All other builds use clang >= 16.

DennisOSRM:
Let me know when the change is ready for reviewing. 😉

unit_tests/mocks/mock_scripting_environment.hpp
```diff
@@ -33,14 +31,10 @@ class MockScriptingEnvironment : public extractor::ScriptingEnvironment
...
                          const extractor::RestrictionParser &,
-                         const extractor::ManeuverOverrideRelationParser &,
-                         const extractor::ExtractionRelationContainer &,
-                         std::vector<std::pair<const osmium::Node &, extractor::ExtractionNode>> &,
-                         std::vector<std::pair<const osmium::Way &, extractor::ExtractionWay>> &,
-                         std::vector<extractor::InputTurnRestriction> &,
-                         std::vector<extractor::InputManeuverOverride> &) override final
+                         const extractor::ManeuverOverrideRelationParser &) override final
     {
     }
```
justinormont, line 40: > **TODO**
> * entrances to buildings along the perimeter of the area,
> * PT stations in the middle of the area

This could also include other routable points like a bridge/tunnel/stairs starting in the middle of an area. 

For example: this, otherwise unconnected, bridge starting/ending in two parking lots ([OSM](https://openstreetmap.org/way/313601074)): 
image

Further, we could include POIs in the middle of routable areas. Perhaps include specific classes like `tourism=artwork` and `amenity=drinking_water`, or more simply anything with a `name=` tag?

My inbuilt assumption is mappers will lessen the use of artificial routes within a routable area, once apps support area routing, increasing the need of routing to individual POIs.

_(comment pinned to a line of code to allow for threaded discussion)_

MarcelloPerathoner: The main difficulty is to get all information into one place: in a typical OSM file all nodes come first, then all ways, then all relations. When you get to the area relation in the OSM file the node tags are long gone, so either you save all interesting nodes with their tags in memory (expensive in memory) or you save all interesting relations and make a second pass over the OSM file to get all the nodes and tags (less expensive in memory but expensive in cpu).

Further, if you want to include nodes that are 'inside' the area but not members of the area relation, you have to perform point-in-polygon tests for all nodes against all interesting areas. That means building an rtree of areas and a second pass over all nodes testing them all (very expensive in cpu).

justinormont: What percent of OSM files will the meshing run on? And what percent will have POIs inside? Perhaps we just have to keep the happy path fast, and the other to be bounded and reasonable?

Could do:
* Run all steps including the new area meshing (as your PR does currently)
* If meshing occurred – breaking off from the happy path
  * Create lat/lng bounding box for each meshed area (fast to check against; granted some corner cases at the poles and wrapping of the world)
  * Ignore any mesh area that is trivial small, like <3m width and <3m height, letting free space routing handle the case
  * Second pass:
    * Filter nodes: 
      * Fast filter: ignore nodes not in a one of the meshes' bounding boxes
      * Slower filter: full check for point-in-polygon
      * Slower filter: ignore nodes that are with-in 3m of an existing/new path in the mesh, letting free space routing handle the point
      * Limiting step: keep at max ~10k nodes that pass the filters (further reduced later)
        * Perhaps prioritizing keeping nodes with `name=` or other likely source/destination POI?
    * Store remaining nodes, then gather ways, relations
      * Filter to POIs w/ `name=` tags or other likely source/destination POI?
    * Limiting step: keep at max ~100 nodes _(not POIs)_ for each mesh areas
    * For any mesh area that we found POIs inside, rerun the meshing
    
Upside: happy path is fast, low and bounded memory, bounded compute
Downside: need to remesh if new POIs are added inside of a mesh area, two passes when area mesh code runs

Woazboat: > This could also include other routable points like a bridge/tunnel/stairs starting in the middle of an area.
> 
> For example: this, otherwise unconnected, bridge starting/ending in two parking lots ([OSM](https://openstreetmap.org/way/313601074)):

As a sidenote because it was used as an example, `amenity=parking` areas cannot safely be assumed to be fully traversible. Very often the mapped parking lot area also includes grassy medians, etc... that cannot be walked/driven over.

---

The parking lot in the example isn't even mapped as an area, just as single nodes, so this wouldn't work there either way. 

Attempting to automatically connect and route between two otherwise unconnected ways just because they are in close proximity but that are not known to lie in a routable area is absolutely not a good idea. That way lies madness.

sfkeller:
It's great that “area routing” is getting some attention again. The algorithms and implementations are available (see, for example, [this comment](https://github.com/graphhopper/graphhopper/issues/82#issuecomment-373587782) and the follow-up posts seven years(!) ago)...

I was recently made aware of this missing feature when I was testing my “[vampire mode](https://sosm.ch/the-coolest-way-to-find-shady-paths-vampire-routing-on-routing-osm-ch/)” - an OSRM profile for pedestrians that prefers shady streets...

safwat-halaby:
Taking full advantage of this may require new tags for areas where it's not clear whether they are traversable or not (for example a `natural=forest`). This discussion is relevant: https://community.openstreetmap.org/t/tagging-a-traversable-area/138357

docs/areas.md
```diff
@@ -0,0 +1,109 @@
...
+must be configured separately.
+
+### Meshing closed ways
+
+To mesh a closed way you must register it in your @ref process_way function. Insert
+following lines into your existing `process_way` function, immediately after the "quick
+initial test":
+
+```lua
+function process_way(profile, way, result, relations)
```
TheMarex, line 40: I'm wondering it it would be better to have an additional `process_area`. Right now the flow is a bit weird IMHO because we let this code decide which way/relation is considered an area, which all the consequences of messing that up. That would also safe us from having to expose the area_manager to lua at all from what I can tell.

features/bicycle/alley.feature
```diff
@@ -27,8 +27,7 @@ Feature: Bicycle - Route around alleys
             | cf    | residential |          |
 
         When I route I should get
-            | from | to | a:nodes    | weight | #                                |
-            | a    | f  | 1:2:3:6    | 196.2  | Avoids d,e,f                     |
-            | a    | e  | 1:2:5      | 172.2  | Take the alley b,e if neccessary |
-            | d    | f  | 4:1:2:3:6  | 248.4  | Avoids the alley d,e,f           |
-
+            | from | to | a:nodes | weight | #                                |
+            | a    | f  | abcf    | 196.2  | Avoids d,e,f                     |
```
TheMarex, line 31: Looks like an unrelated change? `1:2:3:6` -> names?

src/util/coordinate_calculation.cpp
```diff
@@ -55,15 +56,13 @@ static CheapRulerContainer cheap_ruler_container(1800);
 // Does not project the coordinates!
 std::uint64_t squaredEuclideanDistance(const Coordinate lhs, const Coordinate rhs)
 {
-    std::int64_t d_lon = static_cast<std::int32_t>(lhs.lon - rhs.lon);
-    std::int64_t d_lat = static_cast<std::int32_t>(lhs.lat - rhs.lat);
+    std::int64_t d_lon = from_alias<int64_t>(lhs.lon) - from_alias<int64_t>(rhs.lon);
```
TheMarex, line 59: Can you explain what is going on here? FYI we don't use the full range of the 32bits - only ~29 (28bits for the magnitude 1bit for the sign) are actually required for the encoding. It should be static cast to 64bit though I guess.

src/extractor/scripting_environment_lua.cpp
```diff
@@ -228,7 +239,7 @@ void Sol2ScriptingEnvironment::InitContext(LuaScriptingContext &context)
         "weight_precision",
         &ProfileProperties::weight_precision,
         "weight_name",
-        sol::property(&ProfileProperties::SetWeightName, &ProfileProperties::GetWeightName),
+        sol::property(&ProfileProperties::GetWeightName, &ProfileProperties::SetWeightName),
```
TheMarex, line 235: This is interesting - I'm fairly sure we reference this already in the `car` profile for example. So why does it work now? 🤔

src/extractor/scripting_environment_lua.cpp
```diff
@@ -209,6 +218,8 @@ void Sol2ScriptingEnvironment::InitContext(LuaScriptingContext &context)
                                                 "interpolate",
                                                 &RasterContainer::GetRasterInterpolateFromSource);
 
+    // the first parameter to the LUA process_* functions
+    // FIXME: shouldn't all of this be readonly?
```
TheMarex, line 212: V1 profiles have a global properties object. We kind of missed an opportunity with 6.0 to do some actual cleanup here which is sad.

src/extractor/area/area_manager.cpp
```diff
@@ -0,0 +1,219 @@
...
+/**
+ * @brief Second-pass handler for ways.
+ *
+ * This is called when a registered way was found in the input.
+ */
+void AreaManager::after_way(const osmium::Way &way)
+{
+    if (is_registered_closed_way(way.id()))
+    {
+        const char *name = way.get_value_by_key("name", "noname");
```
TheMarex: I think we usually use an empty name instead of `noname`?

src/extractor/area/area_manager.cpp
```diff
@@ -0,0 +1,219 @@
...
+        util::Log(logDEBUG) << "Completing way: " << name << " id: " << way.id();
```
TheMarex: Probably left-over?

include/extractor/area/index_priority_queue.hpp
```diff
@@ -0,0 +1,227 @@
+#ifndef OSRM_EXTRACTOR_AREA_INDEX_PRIORITY_QUEUE_HPP
```
TheMarex, line 1: Can you add more context why we can't use `query_heap.hpp` here? While the implementation is slightly simpler (no `Data`) it should be almost identical in function. One advantage of QueryHeap is that you do not need a visited array because it supports a `WasInserted` operation.

include/extractor/area/typedefs.hpp
```diff
@@ -0,0 +1,134 @@
...
+
+#include <osmium/osm/area.hpp>
+#include <osmium/osm/node_ref.hpp>
+#include <osmium/osm/types.hpp>
+#include <set>
+
+namespace osrm::extractor::area
+{
+
+namespace bg = boost::geometry;
```
TheMarex: In header files please always refer to the full namespace and not create aliases. Can get really confusing with other headers doing this.

src/extractor/area/area_mesher.cpp
```diff
@@ -0,0 +1,517 @@
...
+
+    util::Log(logDEBUG) << "  The polygon has " << poly_segments.size() << " edges:";
+    util::Log(logDEBUG) << "  The vis_map has " << vis_map.size() << " edges:";
+
+    // Add the segments of the visibility graph. Avoid duplicates.
+    for (const OsmiumSegment &s : vis_map)
+    {
+        if (!poly_segments.contains(s))
+        {
+            double weight = bg::distance(s.first, s.second);
```
TheMarex: We have our own haversine implementation in `utils/coordinate_calculation.hpp` - I suspect you can probably get of some (all?) of the boost geometry dependencies?

TheMarex (commented):
I've not completed the review yet, mostly focused on the structural things. Haven't looked at the algorithmic side at all so far. Overall really great contribution - but a large change with a lot of API implications lua-side so I'm trying to be a bit stricter. Let me know if you are interested in further evolving this in this PR (or if you have the time), otherwise we can figure out how to do that on `master`.

include/extractor/extraction_relation.hpp
```diff
@@ -3,202 +3,256 @@
...
-        members_role.clear();
+        return m.ref() == o.ref() && m.type() == o.type();
     }
-
-    const char *GetAttr(const std::string &attr) const
+    friend bool operator<(const RelationMember &m, const RelationMember &o)
     {
-        auto it = std::lower_bound(
-            attributes.begin(), attributes.end(), std::make_pair(attr, std::string()));
+        return m.ref() < o.ref() || (m.ref() == o.ref() && m.type() < m.type());
```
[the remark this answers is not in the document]
michaelblyons: Looks like the AI found a bug... but maybe not the one it reported? Should this be `... & m.type() == o.type()`?

src/util/coordinate_calculation.cpp
MarcelloPerathoner: It is safer from overflow.

src/extractor/area/area_manager.cpp
MarcelloPerathoner: We can remove both lines.

include/extractor/area/index_priority_queue.hpp
MarcelloPerathoner: I couldn't figure out how to use it. There's not a single comment in that whole file. And I need key update operations: the heap is sorted by distance from the sweep centerpoint and as you sweep the ray around the center obstacles don't maintain the same distance.

include/extractor/extraction_relation.hpp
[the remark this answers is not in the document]
MarcelloPerathoner: The AI is right.

DennisOSRM:
@copilot open a new pull request to apply changes based on the comments in [this thread](https://github.com/Project-OSRM/osrm-backend/pull/7161#pullrequestreview-3694222165)

src/extractor/area/area_manager.cpp
```diff
@@ -0,0 +1,219 @@
...
+
+#include <osmium/area/assembler.hpp>
+
+namespace osrm::extractor::area
+{
+
+/**
+ * @brief Initialize the area manager
+ *
+ * @param algorithm_name The alogorithm to use for meshing. At present only one
```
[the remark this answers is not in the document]
justinormont: Not sure why Copilot doesn't just offer a suggestion, that would make for a 1-click accept of the change:
```suggestion
 * @param algorithm_name The algorithm to use for meshing. At present only one
```

profiles/foot_area.lua
```diff
@@ -0,0 +1,284 @@
...
+
+  -- check if node is a traffic light
+  local tag = node:get_value_by_key("highway")
+  if "traffic_signals" == tag then
+    -- Direction should only apply to vehicles
+    result.traffic_lights = true
+  end
+end
+
+-- main entry point for processsing a way
```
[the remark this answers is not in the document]
justinormont: Not sure why Copilot doesn't just offer a suggestion, that would make for a 1-click accept of the change:

```suggestion
-- main entry point for processing a way
```

docs/profiles.md
```diff
@@ -311,14 +327,15 @@ end
 ```
 
 ### process_way(profile, way, result, relations)
+>>>>>>> origin/master
```
Ruben-VandeVelde, line 330:
```suggestion
```

Ruben-VandeVelde (commented):
This would be great to have! A drive-by comment below

src/extractor/area/area_manager.cpp
DennisOSRM: Changed to an empty default, as the rest of the codebase does.

src/extractor/area/area_manager.cpp
DennisOSRM: Removed. The lookup existed only to feed a debug log line, so both lines are gone.

DennisOSRM:
I have been working on and off an getting starting and ending a route on an open area right. I was able to achieve some progress on this despite the summer heat. 

image

DennisOSRM (approved):
Approving to move changes to master. More follow-up changes will come to PRs afterwards.

DennisOSRM:
Merged. Thank you, @MarcelloPerathoner, for this contribution and for your patience:
it has been open a long time, and you kept answering questions on it throughout.

Routing across pedestrian areas has been asked for since #64, and the visibility graph
plus Dijkstra approach here is a good foundation to build on. The mesher, the sweep and
the profile hooks are all yours; what came after is a small fix and some tests on top.

Work continues in follow-ups: snapping a coordinate that falls inside an area, pruning
the mesh to shortest-path trees, and solving a plaza at query time when both ends of a
journey are on it. The points raised in review that were not addressed here are recorded
in #7683 along with the defects known so far, so none of it is lost.

features/bicycle/alley.feature
DennisOSRM: Not unrelated, though the connection is easy to miss. This branch changes the route step to render `a:nodes` as node names instead of raw ids (`features/support/shared_steps.js`), so every existing assertion on that column had to be rewritten: 32 rows across four files, of which alley.feature is three. The weights are unchanged.

The rename earns its place here because meshing an area creates virtual nodes whose ids move whenever the mesh changes, so numeric assertions on them would be unstable.

It left a gap, though. `features/step_definitions/matching.js` resolves the same column by its own path and was not changed, so four cells in `features/testbot/matching.feature` still assert numeric ids and still pass. Two conventions for one column, now recorded in #7683.

src/util/coordinate_calculation.cpp
DennisOSRM: It widens each coordinate to 64 bits before the subtraction rather than after it, and drops a redundant cast on the result. Previously the difference was computed as an alias of `int32_t` and only then assigned to an `int64_t`.

Your reading is right that it cannot overflow in practice, since only about 29 bits are in use, so this is robustness and clarity rather than a fix for an observed defect. And yes, casting to 64 bit is exactly what it now does.

src/extractor/scripting_environment_lua.cpp
DennisOSRM: The two arguments were the wrong way round. `sol::property` takes the getter first and the setter second, and this had `SetWeightName` in the read slot and `GetWeightName` in the write slot.

It works today because nothing reaches it. All the profiles in the repo are `api_version = 4`, and for those the properties come back from `setup()` as a plain Lua table that is read key by key (`scripting_environment_lua.cpp:738` for `weight_name`), so the usertype accessor is never used. A v1 to v3 profile touching `properties.weight_name` through the usertype object would have hit it.

src/extractor/scripting_environment_lua.cpp
DennisOSRM: Agreed, and the split is awkward: v4 reads the same fields twice over, once through this usertype and once as a plain table further down. Not something to unpick here, but worth doing when the profile API is next opened up.

include/extractor/area/index_priority_queue.hpp
DennisOSRM: Fair question, and it is not answered in the header. Carried over to #7683: either the header says why `query_heap` does not fit, or the area code uses it.

docs/areas.md
DennisOSRM: Agreed that the current flow reads oddly, with areas registered from `process_way` and `process_relation` rather than from something of their own. It is a profile API change, so it wants its own discussion rather than being folded in here. Recorded in #7683.

unit_tests/mocks/mock_scripting_environment.hpp
DennisOSRM: Good list, and the unconnected bridge between two parking lots is a case the mesher has nothing to say about today. Entry points are only where another way meets the perimeter, so a way starting in the middle of an area is invisible to it, and POIs are not considered at all. Recorded in #7683 as a follow-up.

include/extractor/area/typedefs.hpp
DennisOSRM: Still open. `include/extractor/area/typedefs.hpp` keeps the `OsmiumPolygon`, `OsmiumMultiPolygon` and `NodeRefSet` aliases, and has a `using namespace osrm::extractor::area;` inside the `boost::geometry::traits` block, which leaks into every translation unit that includes it. The second of those is the one worth removing first. Recorded in #7683.

src/extractor/area/area_mesher.cpp
DennisOSRM: Still open, and there is a wrinkle worth recording before anyone changes it. `AreaMesher::run_dijkstra` weights edges with `boost::geometry::distance`, which for these coordinates yields an angle rather than a length. Only the ratios matter to Dijkstra so the routes come out right, but swapping in the haversine from `coordinate_calculation` rescales every weight, and `Dijkstra::distance_epsilon` is an absolute tolerance. There is a note in `docs/areas.md`; also recorded in #7683.

include/extractor/area/typedefs.hpp
DennisOSRM: Fixed in #7687, which is small and independent: the two using-directives are gone, `VisibilityGraph::Vertex` is named in full inside the traits block, and `COEFF` moves out of `boost::geometry::traits` and becomes `osrm::extractor::area::MERCATOR_COEFF`.

kaligrafy:
Wow, this is legendary work!

Issue #7209, opened 2025-07-18
Legs are not collapsed in /match if no waypoints are provided


Ephtolens:
# Feature

It would be great that legs in /match request are always collapsed, which is currently the case only if we provide waypoints in the request.

TheMarex:
@Ephtolens thanks for opening this. You are aware that this would be a breaking API change? The current behavior is clearly not ideal for your use-case, but as you mentioned you workaround this already by setting the waypoint index.

There is a good reason why we don't do this: The map matching API is written to be consumed by things that care about the actual path between trace points. This is an important building block for traffic data computation (vehicle traces -> paths between trace points -> record speed to each segment).

It sounds like your application mostly cares about the geometry?

PR #7211, opened 2025-07-18
feat(match): always collapse legs in sub_matchings route results

Ephtolens:
# Issue
 
[Legs are not collapsed in /match if no waypoints are provided](https://github.com/Project-OSRM/osrm-backend/issues/7209)

## Tasklist

 - [ ] review
 - [ ] adjust for comments


Ephtolens:
Yes this is still relevant and pending

TheMarex:
See my comment [here](https://github.com/Project-OSRM/osrm-backend/issues/7209) in that we can't accept API-breaking changes. Feel free to explain your use-case in more detail, maybe a flag could work but I doubt it is sensible given this is mostly for convince?

Issue #7238, opened 2025-09-07
Upgrade fmt from 10.2.1 to 11.2.0


afarber:
New fmt features since 10.2.1 (January 2024):

- 40-60% reduction in debug binary size
- Experimental fmt::writer API
- Enhanced C++20 module support
- format_as works with user-defined types
- Performance optimizations and bug fixes

jcoupey:
Since C++20, we now have `std::format` in the standard, which is fully supported by major compilers used in CI in this project (except probably gcc-12). Might be an interesting long-term goal to switch to the native support from the language and remove the dependency altogether.

nyllet:
Inspired by @jcoupey 's comment above, i took the liberty to try it out. not a lot of code changes and seems to work as far as i can see, but maybe issues with older compilers

nyllet:
Turned out that my PR with dropping libfmt requires gcc >= 14 and clang >= 15

DennisOSRM:
Sounds good to me. Let's bump fmt first and then work on compiler upgrades and migration to std::format independently.

PR #6877, opened 2024-05-10
Use C++20

SiarheiFedartsou:
I believe it is a time to bump C++ standard in this project. It has quite a lot of useful things OSRM can benefit from(e.g. std::span and also I see that even `json_render` benchmark became better due to some reason :) ). I hope there are no objections for this.

## Benchmark Results
| Benchmark | Base | PR |
|-----------|------|----|
| alias | aliased u32: 1094.05<br/>plain u32: 1092.11<br/>aliased double: 960.248<br/>plain double: 947.363 | aliased u32: 1091.75<br/>plain u32: 1089.74<br/>aliased double: 959.988<br/>plain double: 948.72 |
| json-render | String: 8.2719ms<br/>Stringstream: 11.5894ms<br/>Vector: 7.47187ms | String: 6.75741ms<br/>Stringstream: 10.5255ms<br/>Vector: 6.62243ms |
| match_ch | Default radius: <br/>4.43728ms/req at 82 coordinate<br/>0.0541131ms/coordinate<br/>Radius 5m: <br/>4.43271ms/req at 82 coordinate<br/>0.0540574ms/coordinate<br/>Radius 10m: <br/>15.0005ms/req at 82 coordinate<br/>0.182933ms/coordinate<br/>Radius 15m: <br/>36.5988ms/req at 82 coordinate<br/>0.446327ms/coordinate<br/>Radius 30m: <br/>312.388ms/req at 82 coordinate<br/>3.80961ms/coordinate | Default radius: <br/>4.53578ms/req at 82 coordinate<br/>0.0553144ms/coordinate<br/>Radius 5m: <br/>4.53172ms/req at 82 coordinate<br/>0.0552649ms/coordinate<br/>Radius 10m: <br/>15.343ms/req at 82 coordinate<br/>0.18711ms/coordinate<br/>Radius 15m: <br/>37.2938ms/req at 82 coordinate<br/>0.454803ms/coordinate<br/>Radius 30m: <br/>319.071ms/req at 82 coordinate<br/>3.89112ms/coordinate |
[...]

PR #7251, opened 2025-09-14
Replace libfmt with std::format

nyllet:
# Issue

This is an alternative solution to https://github.com/Project-OSRM/osrm-backend/issues/7238

## Tasklist

 - [ x] CHANGELOG.md entry ([How to write a changelog entry](http://keepachangelog.com/en/1.0.0/#how))
 - [ ] review
 - [ ] adjust for comments

## Requirements / Relations

 none


afarber:
Please do not forget to remove the fmt from scripts/update_dependencies.sh

afarber:
Hi @nyllet, your PR removes the fmt library entirely and switches directly to `std::format`...

However, this might cause build failures on systems where `std::format` is incomplete. For example:
- Clang with older libstdc++: The `<format>` header exists, but symbols are missing at link time
- Some GCC 13 installations: Have incomplete chrono formatting support
- Alpine Linux with GCC 14: Missing chrono formatters

I've just submitted #7261 which takes a different approach:
- Adds a thin compatibility layer (`util/format.hpp`) that automatically uses `std::format` when available
- Falls back to `fmt::format` on systems where `std::format` doesn't work
- Uses CMake runtime detection (`check_cxx_source_runs`) to test compile, link and execution
- Also fixes the node_osrm C++17/C++20 ABI mismatch that was overlooked in #6877

This way, we can:

1. Keep fmt as a fallback for systems with incomplete std::format
2. Gradually phase out fmt as older compilers are retired
3. Avoid breaking builds for users on platforms with partial C++20 support

Would you be open to reviewing #7261 as an alternative approach?

nyllet:
already fixed in https://github.com/Project-OSRM/osrm-backend/pull/7261

Issue #6171, opened 2021-12-12
MAX_COLLAPSE_DISTANCE and precise pedestrian steps


kaligrafy:
We are trying to route pedestrian with precise steps, including the road crossings which we mapped for every intersection on a small region. Our government partners and transit operators in the region want to be able to get precise routing for pedestrian to assess route security (probability of accidents). However, since MAX_COLLAPSE_DISTANCE is set to 30, any mapped crossing (footway=crossing) shorter than 30 m is merged and ignored in the steps, so that make our analysis impossible sinc we don't know the toal distance for which the pedestrian is crossing the road vs the distance the pedestrian is on a sidewalk or on a designated footpath.

Would it be possible to add MAX_COLLAPSE_DISTANCE to the main routing parameters so we do not need to fork and maintain a branch just for this? Thanks!

danpat:
> Would it be possible to add MAX_COLLAPSE_DISTANCE to the main routing parameters or in the lua profile so we do not need to fork and maintain a branch just for this? Thanks!

Seems reasonable - want to submit a PR that implements that?

kaligrafy:
Is it possible to add it to the lua profile (the real question is: is this parameter in the lua profile would be available suring the processing stage)?

danpat:
Take a look at the `max_speed_for_map_matching` and `continue_straight_at_waypoint` properties that are set in `car.lua` - these don't get used by _processing_, but they are saved into the routing files so that they can be used at query time - that's what you'd want for this parameter.

kaligrafy:
OK, thanks!

kaligrafy:
We will propose a PR soon for this, please keep open.

kaligrafy:
bump

roose242:
Is there any progress? I'm looking for this feature for the same reason. At least i need all road crossing edges untouched as separate steps in the output. If there is something started already i'm willing to help.

kaligrafy:
Created a PR #7252

PR #7252, opened 2025-09-19
Issue6171

kaligrafy:
# Issue

#6171


afarber:
Thanks for your contribution, I have noticed that CHANGELOG.md is missing

kaligrafy:
Hi! Can any maintainer check this PR, it has been a month with no review. Thanks!

kaligrafy:
I added the changelog entry as requested. Thanks!

kaligrafy:
Bump! Is there some timeline for review or commenting PRs? Thanks!

profiles/foot_test.lua
```diff
@@ -0,0 +1,272 @@
+-- Test Foot profile with configurable max_collapse_distance
```
TheMarex, line 1: Feel free to modify the foot profile directly, no need to create a separate profile. 10m makes sense in general, the values was intended for cars.

src/engine/guidance/collapse_turns.cpp
```diff
@@ -15,9 +16,11 @@ namespace osrm::engine::guidance
 using osrm::util::angularDeviation;
 using namespace osrm::guidance;
 
+// Thread-local storage for configurable max collapse distance
+thread_local double current_max_collapse_distance = DEFAULT_MAX_COLLAPSE_DISTANCE;
```
TheMarex, line 20: So this looks like a hack introduced by an LLM to me? The correct fix would simply pass the `max_collapse_distance` to the functions as parameter.

src/engine/guidance/collapse_turns.cpp
```diff
@@ -433,12 +436,15 @@ void suppressStep(RouteStep &step_at_turn_location, RouteStep &step_after_turn_l
 }
 
 // OTHER IMPLEMENTATIONS
-[[nodiscard]] RouteSteps collapseTurnInstructions(RouteSteps steps)
+[[nodiscard]] RouteSteps collapseTurnInstructions(const datafacade::BaseDataFacade &facade, RouteSteps steps)
```
TheMarex, line 439: `max_collapse_distance` should also be a parameter here.

unit_tests/extractor/profile_properties.cpp
```diff
@@ -0,0 +1,89 @@
+#include "extractor/profile_properties.hpp"
```
TheMarex, line 1: All of these tests are useless.

unit_tests/engine/guidance_collapse_integration.cpp
```diff
@@ -0,0 +1,130 @@
+#include "engine/guidance/collapse_turns.hpp"
```
TheMarex, line 1: These tests are also useless, please remove.

kaligrafy:
> I'm guessing this was mostly done by an LLM without real supervision? Look it probably took me longer to review this then it took you to generate this. None of this makes sense and it is a super hacky implementation that no sane human would allow. If you can not review your contributions yourself, then do not submit them here please. I know it is frustrating when nobody fixes your issues, but slop is not the answer and only drains more resources from open source projects.

I am really sorry and you are right. C++ is not one of the languages I master so I just tried using an LLM to generate the patch, thinking it was going to be very trivial and simple, but it was not (or at least the LLM made it more complicated). I will try to propose a better PR as soon as possible as this is a feature we need since a long time ago. Thanks and sorry for the inconvenience!

afarber:
Hi @TheMarex I have tried to follow your suggestions in #7344

TheMarex:
Closing here because we have a replacement in the works.

PR #7256, opened 2025-10-02
Allow returning geometry by leg

TheMarex:
# Issue

Right now it is kind of a pain to get the geometry by leg. You can request the steps and accumulate it, but that is annoying if you are not interested in the instructions. With hindsight that wasn't the greatest API decision, alas better fix it late then never. In general it doesn't make sense to have it by route and by leg at the same time, so it seems appropriate to extend the `overview` option for triggering the new behavior.

This is still work-in-progress, I need to spend some time to make sure the flatbuffer stuff isn't broken by this and add tests.

## Implementation

- [x] Add a new `overview=by_legs` parameter that will add the (unsimplified) geometry by leg.
- [ ] Add support to flatbuffer API

## Tasklist

 - [ ] CHANGELOG.md entry
 - [ ] review
 - [ ] adjust for comments


src/engine/api/json_factory.cpp
```diff
@@ -257,7 +257,10 @@ util::json::Object makeWaypoint(const util::Coordinate &location,
     return waypoint;
 }
 
-util::json::Object makeRouteLeg(guidance::RouteLeg leg, util::json::Array steps)
+util::json::Object makeRouteLeg(guidance::RouteLeg leg,
```
afarber, line 260: `makeRouteLeg` was removed from the header - should this be static or in an anonymous namespace now?

include/engine/api/route_api.hpp
```diff
@@ -186,26 +186,33 @@ class RouteAPI : public BaseAPI
...
-                json_geometry = json::makePolyline<1000000>(begin, end);
-            }
-            else
-            {
-                BOOST_ASSERT(parameters.geometries == RouteParameters::GeometriesType::GeoJSON);
-                json_geometry = json::makeGeoJSONGeometry(begin, end);
-            }
+            json_geometry = MakeGeometry(*geometry);
+        }
+        return json_geometry;
```
DennisOSRM: I wonder if this carries the intent better (and saves a temporary).

```suggestion
            return MakeGeometry(*geometry);
        }
        return std::nullopt;
```

include/engine/api/route_api.hpp
```diff
@@ -186,26 +186,33 @@ class RouteAPI : public BaseAPI
...
+    }
+
+    util::json::Value MakeGeometry(const std::vector<Coordinate> &geometry) const
+    {
+        util::json::Value json_geometry;
+        auto begin = geometry.begin();
+        auto end = geometry.end();
```
DennisOSRM: could this be `cbegin` and `cend`?

include/engine/api/route_api.hpp
```diff
@@ -723,6 +749,16 @@ class RouteAPI : public BaseAPI
...
         std::optional<util::json::Value> json_overview = MakeGeometry(MakeOverview(leg_geometries));
 
+        std::vector<util::json::Value> json_overview_by_legs;
+        if (parameters.overview == RouteParameters::OverviewType::ByLegs)
+        {
+            json_overview_by_legs.reserve(leg_geometries.size());
+            for (const auto idx : util::irange<std::size_t>(0UL, leg_geometries.size()))
+            {
+                json_overview_by_legs.emplace_back(MakeGeometry(leg_geometries[idx].locations));
+            }
```
DennisOSRM, line 759: would the following work? it could save the index variable and be a bit more legible, I think.

```suggestion
            json_overview_by_legs.reserve(leg_geometries.size());
            for (const auto& leg_geometry : leg_geometries)
            {
                json_overview_by_legs.emplace_back(MakeGeometry(leg_geometry.locations));
            }
```

Issue #5440, opened 2019-05-03
osrm_contract exception full planet car profile


lindowman:
Hello,
Running osrm-contract on the full planet car profile throws an exception in ContractedEdgeContainer::Merge

platform: Ubuntu Server 18.04 LTS
osrm-backend: v5.22.0  (Commit id: e86d93760) built Debug version from source
profile: car (using without any local changes)
coverage area: full planet (using planet-latest.osm.pbf)

Extraction was successful.
Contraction seemed to be going well. Until, the following...

....(stuff deleted)
[info] Converting contracted graph with 446059263 to edge list (10705422312 bytes)
[info] Getting edges of minimized graph .
 10%
..... (removed the incremental percent updates)
.
 90%
.
 100%
[assert][139931439740864] /osrm-backend/include/contractor/contracted_edge_container.hpp:130
in: void osrm::contractor::ContractedEdgeContainer::Merge(std::vector<osrm::contractor::QueryEdge>): static_cast<int>(edges_size) >= new_edges_size
terminate called without an active exception

Please let me know if any additional info needs to be provided to identify the problem source.
Also, if osrm were to be built without assertions enabled and assuming there are no other errors on retrying the process, would this be unsafe usage of the routing daemon? Unsafe meaning the router throws or loops forever or some such anomalous behaviour.
Thank you

danpat:
Here's the assertion in question - I think there's a bug in here somewhere, likely an overflow:

```
    126         // append new edges
    127         edges.insert(edges.end(), new_edges.begin(), new_end);
    128         auto edges_size = edges.size();
    129         auto new_edges_size = std::distance(new_edges.begin(), new_end);
    130         BOOST_ASSERT(static_cast<int>(edges_size) >= new_edges_size);
    131         flags.resize(edges_size);
    132         std::fill(flags.begin() + edges_size - new_edges_size, flags.end(), flag);
```

I can't see a fault in the merging logic, *but*, the `static_cast<int>(edges_size)` part of the assertion gives me pause - particularly the `static_cast` part.

Almost without exception, OSRM is using `std::uint32_t` or `unsigned` to represent edge/node indexes on internal types.  My suspicion is that the value here is overflowing the `int` type, causing the assertion to fail.

I know this will take a while, but can you try adding changing the `BOOST_ASSERT` to the following two assertions:

```
BOOST_ASSERT(edges_size >= new_edges_size);
[...]

PR #7257, opened 2025-10-10
fix an assertion that would fire erroneously fire due to int overflow

woodpeck:
I have recently had `osrm-contract` abort on a full-planet import with v6.0.0 with the problem described in issue #5440. The modification of the assertion described in that issue solves the problem but it appears it has never made it into the code proper. So here's a PR that does that; after the modification, the `osrm-contract` step runs normally and produces a functioning routing graph.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,7 +21,9 @@
       - CHANGED: Upgrade Cucumber-js to v12 [#7221](https://github.com/Project-OSRM/osrm-backend/pull/7221)
       - CHANGED: Add libcap-setcap to alpine dockerfile [#7241](https://github.com/Project-OSRM/osrm-backend/issues/7241)
       - FIXED: Minor misspellings found in source code, comments and documents [#7215](https://github.com/Project-OSRM/osrm-backend/pull/7215)
+    - Profile:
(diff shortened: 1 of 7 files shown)


Charmik:
@woodpeck I have the same problem - would be good to merge. I see that one test fails - is it normal?

include/contractor/contracted_edge_container.hpp
```diff
@@ -128,7 +128,8 @@ struct ContractedEdgeContainer
         edges.insert(edges.end(), new_edges.begin(), new_end);
         auto edges_size = edges.size();
         auto new_edges_size = std::distance(new_edges.begin(), new_end);
-        BOOST_ASSERT(static_cast<int>(edges_size) >= new_edges_size);
```
TheMarex, line 131: I think here is a better fix here: We want int64_t which will be equivalent to std::ptrdiff_t (the return type of std::distance). The casting to int was probably there originally because the return type of `std::distance` is _signed_.

TheMarex:
Thanks for your contribution! If you have a minute to rebase and fix, I can get this until next week. Otherwise let me know, happy to just cherry-pick into a fresh PR.

woodpeck:
TBH I've mentally swapped out OSRM at present so the context switch would take me a while, if you have just looked at this then your memory is fresher than mine & if you could just salvage this into a new PR that would be much appreciated!

PR #7261, opened 2025-10-15
Add std::format compatibility layer and fix node_osrm C++17/C++20 ABI mismatch

### LINKED ISSUE #7260: Use std::format when available, fallback to fmt::format for older compilers
Currently OSRM uses libfmt for formatting. With C++20, we can use `std::format` where available while maintaining backward compatibility with older compilers that  lack `<format>` support.

afarber:
Fix #7260 by addressing two issues:

### 1. std::format compatibility layer

Adds runtime detection and fallback for C++20 `std::format`:

- Add CMake runtime detection using check_cxx_source_runs to test compile, link and execution
- Create a thin compatibility wrapper include/util/format.hpp
- Use `std::format` when available and fully functional
- Fallback to `fmt::format` for incomplete implementations (e.g., Clang with older libstdc++ missing symbols)
- Include `<cmath>` before `<format>` to prevent `std::isfinite` conflicts

### 2. Fix node_osrm C++ standard mismatch

When #6877 moved the project to C++20 in May 2024, src/nodejs/CMakeLists.txt was overlooked.
This caused node_osrm to be compiled with C++17 while linking against libosrm compiled with C++20,
creating an ABI mismatch. This PR updates node_osrm to C++20 to match the rest of the codebase.

## Tasklist

 - [ ] review
 - [ ] adjust for comments

## Requirements / Relations

#7251

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
       - FIXED: Reduce MSVC compiler warnings by suppressing informational warnings while preserving bug-indicating warnings [#7253](https://github.com/Project-OSRM/osrm-backend/issues/7253)
     - Misc:
       - CHANGED: Add std::format compatibility layer with fallback to fmt::format [#7261](https://github.com/Project-OSRM/osrm-backend/pull/7261)
+      - FIXED: Update node_osrm to C++20 to fix ABI mismatch with libosrm (was overlooked in #6877) [#7261](https://github.com/Project-OSRM/osrm-backend/pull/7261)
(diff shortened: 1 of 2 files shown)


afarber:
Hi @DennisOSRM and @SiarheiFedartsou please review my PR.

It would allow moving users on newer systems to `std::format` without losing/upsetting users on older systems.

Also it fixes an ABI mismatch, which was causing not found `std::` methods errors while I was preparing the PR initially.

PR #7286, opened 2025-11-26
add argentine speed exceptions

### LINKED ISSUE #7285: Fix car profile for urban/rural speed limits in Argentina
# Issue

In Argentina the default speed limits for urban roads are 40 or 60 km/h, depending on the type of the road, instead of the car-profile's default of 50km/h. Similarly, on rural roads the limit is 110 km/h (or 120 km/h for semi-motorways), instead of the default of 90km/h.

gedankenstuecke:
# Issue

Tries to close #7285 by adding the corresponding exceptions [in line with the Article 51 of the local traffic laws](https://servicios.infoleg.gob.ar/infolegInternet/anexos/0-4999/818/texact.htm).

## Tasklist

 - [ ] review
 - [ ] adjust for comments

## Requirements / Relations

### DIFF
--- a/profiles/car.lua
+++ b/profiles/car.lua
@@ -276,8 +276,8 @@ function setup()
     maxspeed_table = {
       ["at:rural"] = 100,
       ["at:trunk"] = 100,
-      ["ar:urban"]: 40,
-      ["ar:rural"]: 110,      
+      ["ar:urban"] = 40,
+      ["ar:rural"] = 110,      
       ["be:motorway"] = 120,
       ["be-bru:rural"] = 70,
       ["be-bru:urban"] = 30,


frodrigo:
`ar:rural` looks to have very few usage https://taginfo.openstreetmap.org/search?q=maxspeed%3Dar%3Arural

`[country code]:[where]` is used for implicit speed limit. But it looks like to me that not used for AR, eg. comparing with AT. https://wiki.openstreetmap.org/wiki/Key:maxspeed

It rather looks like to me that AR is on the maxspeed per highway type schema ?

gedankenstuecke:
Yeah, the urban one is more common (https://taginfo.openstreetmap.org/search?q=maxspeed%3Dar%3Aurban) and is what triggered my PR, as osmose kept giving false-positives and in my PR over there it was suggested to also adapt it here, see https://github.com/osmose-qa/osmose-backend/pull/2635 :)

DennisOSRM:
Thanks for the contribution. `taginfo.json` needs to be updated to list the new tags, and perhaps a test covering the tags. Then we can merge.

gedankenstuecke:
I added the info into the `taginfo.json` - for the testing I'm not really sure how to go about it, as I didn't manage to find existing tests for the `maxspeed_table` to work from, would you have a pointer?

PR #7295, opened 2025-12-11
Add the PSV vehicle subtypes to the access_tag_blacklist of the car profile (taxi, share_taxi, minibus, and bus)

frodrigo:
Add the PSV vehicle subtypes to the access_tag_blacklist of the car profile (taxi, share_taxi, minibus, and bus).

Car access to highways restricted to PSV vehicles is currently blocked because psv is included in the car profile’s access_tag_blacklist.

However, `psv` is a generic value, and more specific PSV vehicle types can be defined. According to the OSM wiki, these include: taxi, share_taxi, minibus, and bus.

So, also blacklist this values.


frodrigo:
Rebased

DennisOSRM (approved):

PR #7300, opened 2025-12-18
Fix issue 7296

### LINKED ISSUE #7296: FakeTurn objects in ExtractionTurn are created as traffic_signals instead of specific obstacles
# Issue

While upgrading my OSRM setup to v6.0.0, I encountered a strange behavior that I eventually traced back to the new obstacles feature.

In my custom profile, I use a process_turn Lua function that adjusts turn weight based on the angle and on the presence of a traffic light. However, during testing, I noticed that nodes considered as obstacles (i.e. nodes registered in obstacles.lua) such as highway=crossing were receiving a turn penalty that should not have been applied.
This is unexpected because the obstacle created for a crossing has weight = 0 and duration = 0, so passing through that node should introduce no penalty.

Minimal test case

With a crossing node on a single continuous way, I get an unexpected 2-second penalty, even though my code only applies that penalty for traffic lights:
``` cucumber
Given the node map
    """
    a-b-c
    """

And a grid size of 100 meters

And the ways
    | nodes | highway  |
    | abc   | cycleway |

And the nodes
    | node | highway  |
    | b    | crossing |

When I route I should get
    | from | to | speed   | weight | time | #            |
    | a    | c  | 15 km/h | 32.8   | 50s  | no crossroad |
    | a    | b  | 15 km/h | 15.4   | 24s  | no crossroad |
    | b    | c  | 10 km/h | 17.4   | 26s  | no crossroad |
```

Surprisingly, if I split the way into two segments, the penalty disappears:
``` cucumber
Given the node map
    """
    a-b-c
    """

And a grid size of 100 meters

And the ways
    | nodes | highway  |
    | ab    | cycleway |
    | bc    | cycleway |

And the nodes
    | node | highway  |
    | b    | crossing |

When I route I should get
    | from | to | speed   | weight | time | #            |
    | a    | c  | 15 km/h | 30.8   | 48s  | no crossroad |
    | a    | b  | 15 km/h | 15.4   | 24s  | no crossroad |
    | b    | c  | 15 km/h | 15.4   | 24s  | no crossroad |
```
After digging deeper, I found the explanation in graph_compressor.cpp.

In the first test (single way), node b gets compressed, but before compression, turn penalties are computed because the node is detected as an obstacle. A fake turn is then created with has_traffic_light = true by default — even though this is not a traffic light but only a crossing.

This incorrect has_traffic_light = true is passed to my process_turn Lua function, which assumes there is a traffic light and applies the corresponding penalty.

What is even more surprising is that although obstacles allow specifying weight and duration during creation (in Lua), these values are not used during this phase.

# Question

Is this behavior expected?
Should obstacle-defined weight/duration be used here, or should the fake turn not assume has_traffic_light = true for crossings?
How can I be sure the weight and duration penalties defined for obstacles in Lua are correctly applied to the graph ? 
I’d be interested to know whether this is a bug or an intended design.
 
# Steps to reproduce

Please provide the steps required to reproduce your problem.
- osrm-backend v6.0.0
- Custom process_turn lua function that apply penalty on weight and duration if turn.has_traffic_light = true

Thanks in advance !

MarcelloPerathoner:
# Issue

Fixes #7296

### DIFF
--- a/.github/workflows/osrm-backend.yml
+++ b/.github/workflows/osrm-backend.yml
@@ -57,7 +57,7 @@ jobs:
 
         cmake -DCMAKE_BUILD_TYPE=Release -DENABLE_CONAN=ON -DENABLE_NODE_BINDINGS=ON ..
         cmake --build . --config Release
-        
+
     # TODO: MSVC goes out of memory when building our tests
     # - name: Run tests
     #   shell: bash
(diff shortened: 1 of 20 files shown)


TheMarex:
This makes sense to me, feel free to merge once rebased.

PR #7309, opened 2025-12-22
Parallel cucumbers

### LINKED ISSUE #7307: CI runs up from 0:45h to 4:30h
# Issue

CI runs now take ~4:30h to run. Before #7180 they took only ~0:45h.

# Steps to reproduce

Push anything that starts a CI run.

# Specifications

Please provide details of your development environment.
- CI for osrm-backend on github

MarcelloPerathoner:
# Issue

Fixes #7307 - CI runs up from 0:45h to 4:30h

## Improvements

Parallel execution of cucumber tests tremendously improves running times.

Note: when using `--parallel N` make sure there are `N` contiguous free ports at the
configured port number (eg. at ports 5000--5000+N).

A lot more can be configured with profiles (in `cucumber.mjs`). Usage examples:

``` bash
$ npx cucumber-js -p home -p ch -p mmap --format html:/tmp/testlog.html --parallel 8
$ npx cucumber-js -p github -p mld -p directly --parallel 8
$ npm test -- --parallel $JOBS --fail-fast
```
Github step summaries display the test results:

Screenshot 2026-01-08 at 21-33-21 WIP Parallel cucumbers · Project-OSRM_osrm-backend@6b55220

On github CI full logs, also containing the outputs from all binaries, are published to
`https://reports.cucumber.io/`.

Code refactored and simplified. ~150 LoC less.

### Other Fixes

During development other problems were found and fixed:

- The *datastore* load method now works as expected: one and the same `osrm-routed` will
be running in the background for the whole duration of the test. Previous behaviour: for
each scenario an `osrm-routed` would be spawned, loaded with datastore, and then killed
again.

- The data watchdog in `osrm-routed` now will emit the message: 'updated facade' right
after the update (previous behaviour: right before the update), so that it can be used
for process synchronization.

## Breaking changes

If you run the tests using the `npx cucumber-js ...` calling method you must add either
`-p home` or `-p github` as the first profile, eg.:

``` bash
$ npx cucumber-js -p home -p mld -p mmap --parallel 8
```
instead of:

``` bash
$ OSRM_LOAD_METHOD=mmap npx cucumber-js -p mld
```

Old cached files are not cleaned automatically anymore because that interfered with
concurrency. There is no hook in cucumber that runs before the workers start.

We don't support changing the load method in mid-run anymore. (eg. "Given data is loaded
with datastore" no longer works). Reason: With the datastore load method an
`osrm-routed` will be constantly running in the background. No second `osrm-routed` can
open the same port again.

### Retired Environment Variables

- `OSRM_CONNECTION_RETRIES`
- `OSRM_CONNECTION_EXP_BACKOFF_COEF`

## Other changes

Removed one cache level. CH and MLD are processed and cached separately.

The outputs of osrm binaries are not written into lots of small files anymore. Instead
use the `--format html:test/logs/testlog.html` commandline argument to get one big HTML
file with all outputs embedded.

## New configuration variables

New environment variables and/or cucumber `worldParameters` entries were introduced. The
whole configuration is done in `cucumber.mjs`.

| Environment Variable     | worldParameters | Defaults to  |                          |
| ------------------------ | --------------- | ------------ | ------------------------ |
| `CUCUMBER_TIMEOUT`       | `timeout`       | 5000         |  Scenario timeout in ms. |
| `CUCUMBER_HTTP_TIMEOUT`  | `httpTimeout`   | 2000         |  HTTP timeout in ms.     |
| `CUCUMBER_TEST_PATH`     | `testPath`      | `test`       |  The test directory      |
| `CUCUMBER_PROFILES_PATH` | `profilesPath`  | `profiles`   |  The profiles directory  |
| `CUCUMBER_LOGS_PATH`     | `logsPath`      | `test/logs`  |  The logs directory      |
| `CUCUMBER_CACHE_PATH`    | `cachePath`     | `test/cache` |  The cache directory     |
| `OSRM_BUILD_DIR`         | `buildPath`     | `build`      |  Path to the binaries    |
| `OSRM_LOAD_METHOD`       | `loadMethod`    | datastore    |  Data load method        |
| `OSRM_ALGORITHM`         | `algorithm`     | ch           |  Routing algorithm       |
| `OSRM_IP`                | `ip`            | 127.0.0.1    |  IP Address              |
| `OSRM_PORT`              | `port`          | 5000         |  IP Port                 |

## New tags

| Tag                                 | A scenario thus tagged ...                                   |
| ----------------------------------- | ------------------------------------------------------------ |
| `@isolated`                         | will not run while any other scenario is running in parallel |
| `@with_(datastore\|directly\|mmap)` | will be executed iff the load method matches                 |
| `@no_(datastore\|directly\|mmap)`   | will be executed unless the load method matches              |
| `@with_(ci\|mld)`                   | will be executed iff the algorithm matches                   |
| `@no_(ci\|mld)`                     | will be executed unless the algorithm matches                |

## Tasklist

 - [X] TODO: clean up and remove some unrelated stuff
 - [ ] review
 - [ ] adjust for comments

## Requirements / Relations

### DIFF
--- a/features/lib/osrm_loader.js
+++ b/features/lib/osrm_loader.js
@@ -29,47 +29,39 @@ class OSRMBaseLoader {
     const argsAsString = args.join(' ');
     log(`running ${cmd} ${argsAsString}`);
 
-    this.child = child_process.spawn(
-      cmd,
-      args,
-      { env : scenario.environment },
-    );
+    return new Promise((resolve) => {
+      this.child = child_process.spawn(
+        cmd,
+        args,
+        { env : scenario.environment },
+      );
 
-    this.child.on('exit', (code) => {
-      log(`osrm-routed completed with exit code ${code}`);
-      this.child = null;
-    });
-
-    return this.waitForConnection();
-  }
-
-  // Terminates OSRM server process gracefully
-  kill() {
-    if (this.osrmIsRunning()) {
-      const child = this.child;
-      const p = new Promise((resolve) => {
-        child.on('exit', resolve);
+      this.child.stdout.on('data', (data) => {
+        if (data.includes('running and waiting for requests')) {
+          log('Routed running and waiting for requests');
+          resolve();
+        }
       });
-      child.kill();
-      return p;
-    }
-    return Promise.resolve();
+
+      this.child.on('exit', (code) => {
+        log(`osrm-routed completed with exit code ${code}`);
+        this.child = null;
+      });
+    });
   }
 
-  /** Returns a promise resolved when the server is up. */
-  waitForConnection() {
-    const seconds = 10;
-    const waitOptions = {
-      resources: [`tcp:${env.wp.ip}:${env.wp.port}`],
-      delay:    10, // initial delay in ms
-      interval: 10, // poll interval in ms
-      timeout:  seconds * 1000, // timeout in ms
-    };
-    const p = waitOn(waitOptions);
-    p.catch(() => { throw new Error(
-      `Could not connect to osrm-routed after ${seconds} s.`
-    ); });
-    return p;
+  /**
+   * Terminates the OSRM server process gracefully
+   *
+   * @returns A promise resolved when routed has terminated.
+   */
+  kill() {
+    if (!this.osrmIsRunning())
+      return Promise.resolve();
+    return new Promise((resolve) => {
+      this.child.on('close', resolve);
+      this.child.kill();
+    });
   }
 
   osrmIsRunning() {
@@ -80,11 +72,19 @@ class OSRMBaseLoader {
 
   /** Called at the init of the cucumber run */
   beforeAll() { return Promise.resolve(); }
-  /** Called at the start of each scenario */
+  /**
+   * Called at the start of each scenario
+   *
+   * @returns A promise resolved when routed is ready for connections.
+   */
   before(_scenario) {
     return Promise.resolve();
   }
-  /** Called at the end of each scenario */
+  /**
+   * Called at the end of each scenario
+   *
+   * @returns A promise resolved when routed has terminated.
+   */
   after(_scenario) {
     return this.kill();
   }
@@ -187,38 +187,43 @@ export class OSRMDatastoreLoader extends OSRMBaseLoader {
     const argsAsString = args.join(' ');
     this.logSync(`running ${cmd} ${argsAsString}`);
 
-    this.child = child_process.spawn(
-      cmd,
-      args,
-      { env : scenario.environment },
-    );
+    return new Promise((resolve) => {
+      this.child = child_process.spawn(
+        cmd,
+        args,
+        { env : scenario.environment },
+      );
 
-    this.child.stdout.on('data', (data) => {
-      if (data.includes('updated facade')) {
-        this.logSync('Facade updated and promise resolved');
-        this.resolve();
-      }
+      this.child.stdout.on('data', (data) => {
+        this.logSync(`osrm-routed stdout:\n${data}`);
+        if (data.includes('updated facade')) {
+          this.logSync('Facade updated and promise resolved');
+          this.resolve();
+        }
+        else if (data.includes('running and waiting for requests')) {
+          this.logSync('Routed running and waiting for requests');
+          resolve();
+        }
+      });
+
+      // we MUST consume stdout and stderr or the osrm-routed process will block eventually
+      this.child.stderr.on('data', (data) => this.logSync(`osrm-routed stderr:\n${data}`));
+
+      this.child.on('exit', (code, signal) => {
+        this.child = null;
+        if (signal != null) {
+          const msg = `osrm-routed aborted with signal ${signal}`;
+          this.logSync(msg);
+          // throw new Error(msg);
+        }
+        if (code != null) {
+          const msg = `osrm-routed completed with exit code ${code}`;
+          this.logSync(msg);
+          if (code != 0)
+            throw new Error(msg);
+        }
+      });
     });
-
-    // we MUST consume these or the osrm-routed process will block eventually
-    this.child.stderr.on('data', (data) => this.logSync(`osrm-routed stderr:\n${data}`));
-    this.child.stdout.on('data', (data) => this.logSync(`osrm-routed stdout:\n${data}`));
-
-    this.child.on('exit', (code, signal) => {
-      this.child = null;
-      if (signal != null) {
-        const msg = `osrm-routed aborted with signal ${signal}`;
-        this.logSync(msg);
-        // throw new Error(msg);
-      }
-      if (code != null) {
-        const msg = `osrm-routed completed with exit code ${code}`;
-        this.logSync(msg);
-        if (code != 0)
-          throw new Error(msg);
-      }
-    });
-    return this.waitForConnection();
   }
   after () {
     this.current_scenario = null;
--- a/scripts/cucumber_test_matrix.sh
+++ b/scripts/cucumber_test_matrix.sh
@@ -2,7 +2,6 @@
 
 set -o errexit
 set -o pipefail
-# set -o nounset
 
 loadmethods=(mmap directly datastore)
 algorithms=(ch mld)


afarber:
Thanks for working on speeding up the CI! The parallel cucumber approach looks promising.

One concern: could we keep the CI matrix changes out of this PR? Switching from macos-13 to macos-15-intel drops test coverage for users on Ventura/Sonoma, and adding clang-20 (which isn't released yet) seems unrelated to the parallelization work. These changes could be discussed separately if there's a reason to drop older platform support.

Also spotted a few bugs in osrm_loader.js - there are references to an undefined err variable in the error handlers (lines 45, 118, 132) that would crash if osrm-routed fails.

MarcelloPerathoner:
The PR is still WIP. I will cleanup before releasing it. 

BTW clang-21 is current. The reason to drop macos-13 is that github does not support it anymore.

features/lib/osrm_loader.js
```diff
@@ -1,310 +1,247 @@
...
-      if (err) return callback(err);
-      if (!this.osrmIsRunning()) this.launch(callback);
-      else {
-        this.scope.setupOutputLog(
-          this.child,
-          fs.createWriteStream(this.scope.scenarioLogFile, { flags: 'a' }),
-        );
-        callback();
+    });
+    // process.on('SIGIO', this.resolve);
```
TheMarex: Left-over?

TheMarex:
Do you think the non-cucumber changes (datastore, routed, etc) can be broken out independently without messing with the tests? It is always quite the challenge to review large changes like this so I would appreciate the extra effort if possible. :pray:

MarcelloPerathoner:
The shared memory changes could easily go elsewhere, as they are not the primary reason for sync loss between the test suite and routed. But the change in routed is critical (and trivial too).

The changes to the github workflow should go elsewhere. I'm just playing around to get windows to run the test suite too. See if any surprises come up.

MarcelloPerathoner:
Ready for review. I have removed everything that is not strictly necessary. The shared memory fixes for macOS and Windows will go into another PR. Until that macOS tests with datastore will be flaky.

features/support/env.js
```diff
@@ -1,85 +1,78 @@
...
-import path from 'path';
-import util from 'util';
-import fs from 'fs';
-import d3 from 'd3-queue';
-import child_process from 'child_process';
-import tryConnect from '../lib/try_connect.js';
-import { setDefaultTimeout } from '@cucumber/cucumber';
-
-// Set global timeout for all steps and hooks
-setDefaultTimeout(
```
afarber, line 11: You have removed the setDefaultTimeout() method and that is why the `CUCUMBER_TIMEOUT: 60000` from the `conan-macos-arm64-release-node` section is not passed anywhere and default timeout of 5000 is used causing the macOS arm64 job to fail

docs/cucumber.md
```diff
@@ -0,0 +1,144 @@
...
+located by default in `test/cache` and should be cleaned periodically:
+
+``` bash
+$ rm -rf test/cache
+```
+
+## Configuration
+
+The whole configuration is done in `cucumber.mjs`. You can either edit `worldParameters`
+in `cucumber.mjs` or use environemnt variables to override single defaults.
```
afarber: Typo: environemnt

features/support/cache.js
```diff
@@ -1,187 +1,39 @@
...
+    // shorten uri to be relative to 'features/'
+    const featurePath = path.relative(path.resolve('./features'), uri);
+    /** eg. car/access.feature/{hash}/ */
+    this.featureID = path.join(featurePath, hexHash);
+    /** eg. test/cache/car/access.feature/{hash}/ */
+    this.featureCacheDirectory = path.join(this.env.wp.cachePath, this.featureID);
+
+    // ensure there is a make a cache directory
+    fs.mkdirSync(this.featureCacheDirectory, { recursive: true });
+  };
```
afarber: The semicolon is not needed

features/lib/osrm_loader.js
```diff
@@ -1,310 +1,246 @@
...
+  async logSync(msg) {
+    if (this.current_scenario)
+      await this.current_scenario.log(msg);
+    else
+      env.globalLog(msg);
+  }
+
+  before(scenario) {
+    this.current_scenario = scenario;
+    this.semaphore = new Promise((resolve) => this.resolve = resolve);
```
afarber, line 152: Consider adding a timeout fallback:

```js
  before(scenario) {
    this.current_scenario = scenario;
    this.semaphore = new Promise((resolve, reject) => {
      this.resolve = resolve;
      setTimeout(() => reject(new Error('Timed out waiting for facade update')), env.wp.timeout);
    });
    // ...
 }
```

Or use `Promise.race()` with the existing `waitForConnection()` timeout.

MarcelloPerathoner: In that case Cucumber will timeout. No need to complicate the code.

features/support/env.js
MarcelloPerathoner: The macOS builds fail randomly because of a race condition: see #7320 #7321. The build fails if `osrm-datastore` exits too early, not too late. Thus a longer timeout would not solve anything here.

And if `osrm-routed` dies for whatever reason, then a longer timeout would just slow down things because each scenario would take exactly timeout ms to fail.

OTOH the timeout was the reason the builds kept failing for such a long time: each time a build failed somebody just raised the timeout. Then, if the next CI run did by pure chance succeed, the problem was declared solved. Didn't you ever wonder how a totally absurd timeout of 60 seconds was reached?

I'm thinking of removing the timeout setting altogether. It induces people to fiddle with the parameters instead of finding the underlying cause. The default 5 sec is plenty.

docs/cucumber.md
MarcelloPerathoner: fixed

features/support/cache.js
MarcelloPerathoner: fixed

features/support/env.js
MarcelloPerathoner: Reinstated env var `CUCUMBER_TIMEOUT` with advice against using it.

afarber:
Looks like your PR #7321 resolved the root cause - nice work.

I notice you've added back `setDefaultTimeout()` with `CUCUMBER_TIMEOUT` support, but since all the `CUCUMBER_TIMEOUT: 60000` entries were removed from the workflow, the env var is never set. It will always default to 5000ms, which is also Cucumber's built-in default.

You could simplify by removing `setDefaultTimeout()` again.

MarcelloPerathoner:
Thanks. I put the timeout back in case somebody might want it badly, though I strongly advise against using it. Currently everything works fine with the default timeout.

features/options/datastore/datastore.feature
```diff
@@ -1,33 +1,63 @@
-@datastore @options @help
+@datastore @options @help @isolated @no_datastore
```
TheMarex, line 1: Hm I'm not so I follow: What does `@datastore` in conjunction with `@no_datastore` do?

features/step_definitions/data.js
```diff
@@ -66,9 +61,7 @@ Given(/^the shortcuts$/, function (table, callback) {
 });
 
 Given(/^the node map$/, function (docstring, callback) {
-  const q = d3.queue();
```
TheMarex, line 69: Hm interesting - any idea why this used to be async?

features/support/route.js
```diff
@@ -1,25 +1,43 @@
...
-    let uri;
-    if (service == 'timestamp') {
-      uri = [this.HOST, service].join('/');
-    } else {
-      uri = [this.HOST, service, 'v1', this.profile].join('/');
+  paramsToQuery(params) {
+    let query = '';
+    if (params.coordinates !== undefined) {
+      // FIXME this disables passing the output if its a default
+      // Remove after #2173 is fixed.
```
TheMarex, line 15: So actually that referenced issue _is_ fixed. Though honestly it may be fine to leave it implicitly as is.

features/testbot/alternative_loop.feature
```diff
@@ -1,4 +1,4 @@
-@routing @testbot @alternative
+@routing @testbot @alternative @isolated @todo
```
TheMarex, line 1: Why the `todo` here?

features/testbot/load.feature
```diff
@@ -1,5 +1,11 @@
-@routing @load @testbot
+@routing @load @testbot @todo @isolated @no_datastore
 Feature: Ways of loading data
+
+# TEST BROKEN: we don't support changing the load method in mid-run.
```
TheMarex, line 4: If we don't start a new `osrm-routed` does this work though? The core here is I think the spring clean.

scripts/cucumber_test_matrix.sh
```diff
@@ -2,17 +2,31 @@
 
 set -o errexit
 set -o pipefail
-set -o nounset
```
TheMarex, line 5: Left-over from debugging?

TheMarex (approved):
Looks great. 🙇 Tried it locally and while I'll miss the never ending row of green dots broken by the inevitable red `F` this is so nice. Massive amount of work, appreciate it. 🙌 Only minor questions / comments.

features/options/datastore/datastore.feature
MarcelloPerathoner: `@datastore` is a legacy tag, means: run all test concerning `osrm-datastore`. Those tests do ugly things like: `osrm-datastore --spring-clean`. So they cannot run while other tests are actually using the datastore. `@no_datastore` is a new tag, meaning: don't run if the selected load method is datastore. A bit confusing ...

features/step_definitions/data.js
MarcelloPerathoner: Because it was easier to call it this way instead of thru a chain of callbacks?

features/testbot/alternative_loop.feature
MarcelloPerathoner: That test must be rewritten. The result is unstable.

features/testbot/load.feature
MarcelloPerathoner: The logic for that test has been removed. While it may be possible to switch between `mmap` and `directly` I don't see the point. Most of those tests are duplicates.

scripts/cucumber_test_matrix.sh
MarcelloPerathoner: `$GITHUB_STEP_SUMMARY` may be undefined.

Issue #5703, opened 2020-04-02
Update traffic data without relaunching the web server "osrm-routed"


abdelhakimbendjabeur:
Hello,

I am experimenting on osrm locally using docker and I have been trying to customize the graph using custom segment speed files. I only see the effect of my speed when I relaunch the `osrm-routed`.

Here are the commands I used
```bash
wget http://download.geofabrik.de/europe/france/ile-de-france-latest.osm.pbf

docker run -t -v "${PWD}:/data" osrm/osrm-backend osrm-extract -p /opt/car.lua /data/ile-de-france-latest.osm.pbf

docker run -t -v "${PWD}:/data" osrm/osrm-backend osrm-partition /data/ile-de-france-latest.osrm

docker run -t -v "${PWD}:/data" osrm/osrm-backend osrm-customize /data/ile-de-france-latest.osrm

docker run -d -t -i -p 1000:5000 -v "${PWD}:/data" osrm/osrm-backend osrm-routed --algorithm mld /data/ile-de-france-latest.osrm
>>> <container-id>
```
and now I have a working web server on `localhost:1000`

Next I do is just get copy the speed file inside the container `<container-id>` and run the following
```
// My terminal
docker cp ~/Desktop/speedfile.csv <container-id>:/tmp

// Inside the container
root@79fc04d4079f:/opt# osrm-customize /data/ile-de-france-latest.osrm --segment-speed-file /tmp/speed_file.csv
```
log output
```
[info] Loaded /tmp/speed_file.csv with 16values
[info] In total loaded 1 file(s) with a total of 16 unique values
[info] Used 3504488 speeds from LUA profile or input map
[info] Used 16 speeds from /tmp/speed_file.csv
[warn] Speed values were used to update 16 segments for 'routability' profile
[info] Updating segment data took 412.522ms.
[info] In total loaded 0 file(s) with a total of 0 unique values
[info] Done reading edges in 2638.76ms.
[info] Loaded edge based graph: 3342718 edges, 936893 nodes
[info] Loading partition data took 3.77922 seconds
[info] Cells customization took 6.5012 seconds
[info] Cells statistics per level
[info] Level 1 #cells 5050 #boundary nodes 93417, sources: avg. 11, destinations: avg. 16, entries: 1171113 (9368904 bytes)
[info] Level 2 #cells 346 #boundary nodes 14907, sources: avg. 27, destinations: avg. 36, entries: 417244 (3337952 bytes)
[...]

Issue #5134, opened 2018-07-06
Is shared memory usefull in docker container ?


grib0:
Hi,

I followed this [wiki page for shared memory configuration](https://github.com/Project-OSRM/osrm-backend/wiki/Configuring-and-using-Shared-Memory)

I made the configuration in the server files
* /etc/sysctl.conf
* /etc/security/limits.conf

I ran docker container using the following parameter :
``` 
docker run --restart=always --ulimit memlock=68719476736:68719476736 --name osrm-backend -d -p 5000:5000 -v /opt/osrm:/data osrm/osrm-backend:v5.18.0 sh -c "osrm-datastore /data/france-latest.osrm && osrm-routed --shared-memory=yes --algorithm mld"
```

The startup output is :

``` 
[info] Data layout has a size of 1894 bytes
[info] Allocating shared memory of 1096107224 bytes
[info] Data layout has a size of 1661 bytes
[info] Allocating shared memory of 2630081913 bytes
[info] All data loaded. Notify all client about new data in:
[info] /static  0
[info] /updatable       1
[info] All clients switched.
[info] starting up engines, v5.18.0
[info] Loading from shared memory
[info] Threads: 1
[info] IP address: 0.0.0.0
[info] IP port: 5000
[info] http 1.1 compression handled by zlib version 1.2.11
[info] Listening on: 0.0.0.0:5000
[info] running and waiting for requests

```
Why data layout has a only size of 1894 bytes and 1661 bytes ?

Also I ran performance test with or without shared-memory and the results are the same.
So my question is, does the shared memory parameter increase performance and is it compatible with docker container of OSRM or it only applies to standalone start up ?

TheMarex:
The actual size of the shared memory is:

> [info] Allocating shared memory of 1096107224 bytes
> [info] Allocating shared memory of 2630081913 bytes

The other lines are just the size of the data structures that save the mappings.

> So my question is, does the shared memory parameter increase performance and is it compatible with docker container of OSRM or it only applies to standalone start up ?

It is used for enabling use-cases where you would need either several processes sharing the same data (e.g. when using OSRM through the node bindings, which several node processes), or for updating the data in-place while a process is running (important for traffic updates). It is generally not relevant for performance.

I'm not sure how you could setup docker to share memory between containers though.

klesniewski:
[...]

PR #7312, opened 2026-01-02
Add SHM_LOCK_DIR environment variable for containerized deployments (memory leak fix)

wes4m:
## Problem

OSRM crashes with `"lock file does not exist"` when using shared memory (`osrm-routed -s`) in containerized environments (Kubernetes) after a container restart.

The crash triggers a restart/crash loop, and each restart accumulates orphaned shm segments, leaking memory proportional to the graph size until the node is exhausted.

Related issues: #5134, #5703

## Reproduction

See gist: https://gist.github.com/wes4m/719cb69b72e26c09c7ff57ed71cf33d9

## Why This Happens in Containers

OSRM's shared memory implementation assumes `/tmp` and System V shared memory have the same lifecycle. This is true on traditional systems where both are cleared on reboot, but not in containers:

| Component | Traditional System | Container |
|-----------|-------------------|-----------|
| `/tmp` filesystem | Cleared on reboot | Cleared on container restart |
| System V shared memory | Cleared on reboot | Survives container restart (IPC namespace persists) |

When a container restarts:
1. `/tmp` is reset and lock files get deleted
2. IPC namespace survives with shm segments persisting (regardless of hostIPC setting)
3. `SharedRegionRegister` (in shm) still references old segments
4. OSRM checks for the old segment's lock file before attaching, doesn't find it, and crashes

## Solution

This PR adds `OSRM_LOCK_DIR` environment variable to specify a custom directory for lock files. This allows containerized deployments to place lock files in a volume that persists across container restarts (e.g. kubernetes `emptyDir`)

## Other Approaches

Mounting `/tmp` to a persistent volume fixes the issue but will persist all temporary files from the container, not just lock files. Setting `TMPDIR` will do the same, and cleaning orphaned shm segments on startup is a workaround that doesn't address the root cause.

`OSRM_LOCK_DIR` only affects lock file location and is backward compatible (tmp dir fallback), behavior is unchanged.

### DIFF
--- a/.github/workflows/osrm-backend.yml
+++ b/.github/workflows/osrm-backend.yml
@@ -68,13 +68,13 @@ jobs:
     # - name: Run node tests
     #   shell: bash
     #   run: |
-    #     ./lib/binding/osrm-extract.exe -p profiles/car.lua test/data/monaco.osm.pbf
+    #     ./lib/binding_napi_v8/osrm-extract.exe -p profiles/car.lua test/data/monaco.osm.pbf
 
     #     mkdir -p test/data/ch
     #     cp test/data/monaco.osrm* test/data/ch/
-    #     ./lib/binding/osrm-contract.exe test/data/ch/monaco.osrm
+    #     ./lib/binding_napi_v8/osrm-contract.exe test/data/ch/monaco.osrm
 
-    #     ./lib/binding/osrm-datastore.exe test/data/ch/monaco.osrm
+    #     ./lib/binding_napi_v8/osrm-datastore.exe test/data/ch/monaco.osrm
     #     node test/nodejs/index.js
     - name: Build Node package
       shell: bash
@@ -319,7 +319,7 @@ jobs:
             build_node_package: true
             continue-on-error: true
             node: 22
-            runs-on: macos-13 # x86_64
+            runs-on: macos-15-intel # x86_64
             BUILD_TYPE: Release
             CCOMPILER: clang
             CXXCOMPILER: clang++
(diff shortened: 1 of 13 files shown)


include/storage/shared_memory.hpp
```diff
@@ -29,13 +30,26 @@
...
 
+// Returns directory for OSRM lock files (OSRM_LOCK_DIR env var or system temp)
+inline std::filesystem::path getLockDir()
+{
+    if (const char *lock_dir = std::getenv("OSRM_LOCK_DIR"))
+    {
+        std::filesystem::path dir(lock_dir);
+        if (!std::filesystem::exists(dir))
+        {
+            std::filesystem::create_directories(dir);
```
TheMarex: IMHO this should error here. The directory should exist.

include/storage/shared_memory.hpp
```diff
@@ -29,13 +30,26 @@
 namespace osrm::storage
 {
 
+// Returns directory for OSRM lock files (OSRM_LOCK_DIR env var or system temp)
+inline std::filesystem::path getLockDir()
+{
+    if (const char *lock_dir = std::getenv("OSRM_LOCK_DIR"))
```
TheMarex: Generally speaking I think a prefix like `OSRM_` makes sense, but it is inconsistent with the other env variables. I would recommend `SHM_LOCK_DIR`.

wes4m:
> This change makes sense. Please add an entry to the changelog. Let me know if you have time to do the small edits yourself, otherwise I can try to put this on my TODO for the week.

Thanks for the review @TheMarex. changes pushed.

Issue #7129, opened 2025-03-06
Failing to compiler from master head


madwax:
# Issue

build of master failed.  
I have just setup a Ubuntu 24.04 machine, cloned master (9857dcfc7d792160cdc408985b7a984383185b7e) and tried to build using cmake as outline in the instructions.

I get the following errors:

In file included from /home/madwax/devel/tmp/osrm-backend/include/extractor/scripting_environment_lua.hpp:15,
                 from /home/madwax/devel/tmp/osrm-backend/src/extractor/scripting_environment_lua.cpp:1:
In member function ‘void sol::stack::field_getter<T, global, raw, <template-parameter-1-4> >::get(lua_State*, Key&&, int) [with Key = const char (&)[23]; T = char [23]; bool global = false; bool raw = false; <template-parameter-1-4> = void]’,
    inlined from ‘void sol::stack::get_field(lua_State*, Key&&, int) [with bool global = false; bool raw = false; Key = const char (&)[23]]’ at /home/madwax/devel/tmp/osrm-backend/third_party/sol2/include/sol/sol.hpp:12092:62,
    inlined from ‘sol::stack::probe sol::stack::probe_field_getter<T, P, global, raw, <template-parameter-1-5> >::get(lua_State*, Key&&, int) [with Key = const char (&)[23]; T = char [23]; P = int; bool b = false; bool raw = false; <template-parameter-1-5> = void]’ at /home/madwax/devel/tmp/osrm-backend/third_party/sol2/include/sol/sol.hpp:16329:21,
    inlined from ‘sol::stack::probe sol::stack::probe_field_getter<T, P, global, raw, <template-parameter-1-5> >::get(lua_State*, Key&&, int) [with Key = const char (&)[15]; T = char [15]; P = int; bool b = false; bool raw = false; <template-parameter-1-5> = void]’ at /home/madwax/devel/tmp/osrm-backend/third_party/sol2/include/sol/sol.hpp:16323:9,
    inlined from ‘sol::stack::probe sol::stack::probe_get_field(lua_State*, Key&&, int) [with bool global = false; bool raw = false; C = int; Key = const char (&)[15]]’ at /home/madwax/devel/tmp/osrm-backend/third_party/sol2/include/sol/sol.hpp:12112:78,
[...]

PR #7317, opened 2026-01-08
Disable array bounds warning until GCC is fixed

TheMarex:
# Issue

There are various issues around failing builds because of a false-positive in sol2. Simple "fix" is do disable the array bounds check for now. Workaround for #7129 

## Tasklist

 - [ ] review
 - [ ] adjust for comments


afarber:
Hi @TheMarex have you tried wrapping the` #include <sol/sol.hpp>` in scripting_environment_lua.hpp with `#pragma GCC diagnostic ignored "-Warray-bounds"`? 

That would keep the warning enabled for OSRM's own code.

TheMarex:
@afarber Excellent suggestion, didn't know GCC supported that.

PR #7321, opened 2026-01-09
Fix macOS and Windows shared mem wait for detach

### LINKED ISSUE #7320: osrm-datastore on macOS and win do not wait for clients to switch
# Issue

`osrm-datastore` on macOS and Windows does not wait for other processes to detach from shared
memory before exiting. Instead it sleeps for 50ms and then boldly announces: 'All clients switched.' 

Symptoms:

- Cucumber tests fail randomly because queries are executed against old data when 50ms were not enough for clients to switch
- Cucumber tests do not run as fast as they could run when 50ms is plenty to switch

# Steps to reproduce

This happens with the new Parallel Cucumbers PR #7309 

# Specifications

Please provide details of your development environment.

- github CI 
- with PR #7309 applied

MarcelloPerathoner:
# Issue

Fixes #7320

On macOS and Windows `osrm-datastore` now correctly waits for other processes to detach from shared memory before exiting. Previous behaviour: `osrm-datastore` slept for 50ms and then boldly announced: 'All clients switched.'

This caused

- tests to fail randomly because queries were executed against old data, and
- tests to run slower than on linux.

## Tasklist

 - [ ] review
 - [ ] adjust for comments

## Requirements / Relations

Related to PR #7309

### DIFF
--- a/include/engine/data_watchdog.hpp
+++ b/include/engine/data_watchdog.hpp
@@ -111,10 +111,6 @@ class DataWatchdogImpl<AlgorithmT, datafacade::ContiguousInternalMemoryDataFacad
                 updatable_region = *updatable_shared_region;
             }
 
-            util::Log() << "updated facade to regions " << (int)static_region.proj_id << " and "
-                        << (int)updatable_region.proj_id << " with timestamps "
-                        << static_region.timestamp << " and " << updatable_region.timestamp;
-
             {
                 boost::unique_lock<boost::shared_mutex> swap_lock(factory_mutex);
                 facade_factory =
@@ -123,6 +119,10 @@ class DataWatchdogImpl<AlgorithmT, datafacade::ContiguousInternalMemoryDataFacad
                             std::vector<storage::ProjID>{static_region.proj_id,
                                                          updatable_region.proj_id}));
             }
+
+            util::Log() << "updated facade to regions " << (int)static_region.proj_id << " and "
+                        << (int)updatable_region.proj_id << " with timestamps "
+                        << static_region.timestamp << " and " << updatable_region.timestamp;
         }
 
         util::Log() << "DataWatchdog thread stopped";


include/storage/shared_datatype.hpp
```diff
@@ -192,12 +186,18 @@ class TarDataLayout final : public BaseDataLayout
     }
 };
 
+#if defined __key_t_defined
```
TheMarex: Compatibility issue with OSX? Maybe worth to add a small comment why we can't expect this to be present?

include/storage/shared_datatype.hpp
```diff
@@ -207,7 +207,7 @@ struct SharedRegion
 
     char name[MAX_NAME_LENGTH + 1];
     std::uint64_t timestamp;
-    std::uint16_t shm_key = 0;
+    ShmKey shm_key = 0;
```
TheMarex: Just checking: We are not actually changing the size of this struct right? If so, I think this would be a breaking change since existing memory regions would be interpreted wrong.

include/storage/shared_memory.hpp
```diff
@@ -48,12 +44,34 @@ inline std::filesystem::path getLockDir()
...
+{
+  public:
+    OSRMShmName(const ShmKey shm_key) { name = "osrm-" + std::to_string(shm_key); }
+    operator const char *() const noexcept { return name.c_str(); }
+
+  private:
+    std::string name;
 };
 
+using namespace boost::interprocess;
```
TheMarex: Please do not use aliases like that in headers since it pollutes the namespace. it is fine to do this in cpp-files or function-local.

include/storage/shared_memory.hpp
```diff
@@ -95,118 +109,85 @@ class SharedMemory
...
+            OSRMLockFile lock_file(shm_key);
+            xsi_key xsi_key(lock_file, shm_key);
+            xsi_shared_memory xsi(open_only, xsi_key);
+            while (true)
             {
-                auto error_code = errno;
-                throw util::exception("shmctl encountered an error: " + errorToMessage(error_code) +
-                                      SOURCE_REF);
+                ::shmid_ds xsi_ds;
+                if (::shmctl(xsi.get_shmid(), IPC_STAT, &xsi_ds) < 0)
```
TheMarex: If I read this correctly, in the case of an error we swallow that and just return? Previously it would have thrown an exception.

include/storage/shared_memory.hpp
```diff
@@ -95,118 +109,85 @@ class SharedMemory
...
-    boost::interprocess::xsi_shared_memory shm;
-    boost::interprocess::mapped_region region;
+  private:
+    xsi_shared_memory shm;
+    mapped_region region;
 };
 #else
-// Windows - specific code
+
+// POSIX shared mem
```
TheMarex: I don't remember exactly why we used XSI over POSIX. Maybe issues with OSX at some point? To be honest we were originally planning to drop the whole hot-swapping thing in favor of mmap. The hotswap/shmem was conceived almost 15 years ago when the typical deployment process was "run it on a big server", not rolling deployments behind a load balancer.

Would have been nice for the 6.0 to clean this up but alas.

include/storage/shared_datatype.hpp
MarcelloPerathoner: Good catch. ShmKey is 32 bits. If that is an issue, maybe I cannot use the shm key here but must use the id. Looking into it ...

include/storage/shared_memory.hpp
MarcelloPerathoner: But we are expecting an error here. The error being that the region is gone.

include/storage/shared_memory.hpp
MarcelloPerathoner: Moved stuff into .cpp file since there were no more templates.

include/storage/shared_datatype.hpp
MarcelloPerathoner: Using the id now, which technically is also an int, but only the 8 lowest bits are used, so we can stuff it into a short. See `man 3 ftok`

include/storage/shared_datatype.hpp
MarcelloPerathoner: key_t no longer used.

include/storage/shared_memory.hpp
TheMarex: But can we check for a precise error code in that case? There may be other reasons this call errors.

MarcelloPerathoner: If we got there, the only plausible error that still can occur is: "EIDRM shmid points to a removed identifier". And that is exactly what we are waiting for to happen. I can test for this error and throw if any other error happens.

TheMarex: IMHO it makes it a bit more clear that we are waiting for something specific to happen.

PR #6407, opened 2022-10-14
Revert back to using custom HTTP parser instead of Boost.Beast

SiarheiFedartsou:
# Issue

Context: https://github.com/Project-OSRM/osrm-backend/issues/6400#issuecomment-1278556823
closes https://github.com/Project-OSRM/osrm-backend/issues/6400

## Tasklist

 - [ ] CHANGELOG.md entry ([How to write a changelog entry](http://keepachangelog.com/en/1.0.0/#how))
 - [ ] update relevant [Wiki pages](https://github.com/Project-OSRM/osrm-backend/wiki)
 - [ ] add tests (see [testing documentation](https://github.com/Project-OSRM/osrm-backend/blob/master/docs/testing.md))
 - [ ] review
 - [ ] adjust for comments
 - [ ] cherry pick to release branch

## Requirements / Relations

 Link any requirements here. Other pull requests this PR is based on?

PR #7328, opened 2026-01-12
Port osrm-routed to use boost::beast

TheMarex:
# Issue

We have long discussed sun-setting `osrm-routed` in favor of `osrm-routed.js`. (see https://github.com/Project-OSRM/osrm-backend/pull/6411 and https://github.com/Project-OSRM/osrm-backend/issues/3832). Now having written #3832 9 years ago, it is time to reflect if that was such a good idea.

1. `osrm-routed` is still around a most people use it inside a docker container behind some kind of HTTP proxy.
2. API-layers use a lot of different technologies, depending on NodeJS was mostly based on the MBX preferences at the time.

It is clear the current HTTP code is not great, but there are some good options in C++ land that are really mature and more then enough for a standalone HTTP server. For example `boost::beast` which makes a lot of sense since we use `boost::asio` anyway.

Thanks to our new AI overlords that transition was relatively smooth with some hand-holding. I've done some benchmarks and I think we should drop the NodeJS path in favor of this. It's pretty clear that we have some issues with initial connection setup, the new beast-based implementation is dramatically faster if we are not saturated.

hist_12

We need 128 concurrent connections to saturate my machine (4 cores, 8 threads) with the baseline implementation, we can do the same with far fewer connections using the new beast implementation. This is random MLD P2P queries on Bavaria.

hist_128

## Tasklist

 - [x] review
 - [x] adjust for comments

## Requirements / Relations

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@
       - FIXED: Work around compilation error due to a false-positive of array-bounds check in sol2 [#7317](https://github.com/Project-OSRM/osrm-backend/pull/7317)
       - FIXED: Fix compilation with gcc >14 in release with LTO. [#7268](https://github.com/Project-OSRM/osrm-backend/issues/7268)
     - Misc:
+      - CHANGED: Use boost::beast instead of own HTTP code for osrm-routed [#7328](https://github.com/Project-OSRM/osrm-backend/pull/7328)
       - ADDED: `SHM_LOCK_DIR` environment variable for shared memory lock file directory [#7312](https://github.com/Project-OSRM/osrm-backend/pull/7312)
       - FIXED: Fix JSON rendering of large OSM IDs (avoids scientific notation) and handle NaN/Infinity gracefully [#7016](https://github.com/Project-OSRM/osrm-backend/issues/7016)
       - FIXED: Replace removed --polyglot flag with JSDoc extraction script for documentation.js v14 [#7274](https://github.com/Project-OSRM/osrm-backend/issues/7274)


DennisOSRM:
That's a fascinating change. Thanks for preparing the insightful experiments.

DennisOSRM:
Looks like there's a segfault happening:

`       Error: osrm-routed killed by signal SIGSEGV: /home/runner/work/osrm-backend/osrm-backend/build-osrm/osrm-routed --dataset-name=cucumber -s -i 127.0.0.1 -p 5000 -a CH `

TheMarex:
Well this took a while to debug (only crashed every few thousand `osrm-routed` restarts...) but found the issue in a race condition between closing the request acceptor and accepting new requests. This should be ready to go now if all the tests pass.

CMakeLists.txt
```diff
@@ -367,6 +367,14 @@ endif()
 if (ENABLE_CONAN)
   set(Boost_USE_STATIC_LIBS ON)
   find_package(Boost REQUIRED COMPONENTS ${BOOST_COMPONENTS})
+  
+  # Boost.Beast requires at least Boost 1.66, but works best with 1.70+
```
DennisOSRM: We should probably be fine to just mandate a newer boost release. 1.70 was released 5+ years ago. And 1.66 in 2017.

TheMarex:
This is ready now.

include/server/server.hpp
```diff
@@ -35,7 +34,7 @@ class Server
                                                 unsigned requested_num_threads,
                                                 short keepalive_timeout)
     {
-        util::Log() << "http 1.1 compression handled by zlib version " << zlibVersion();
+        util::Log() << "HTTP/1.1 server using Boost.Beast, compression by zlib " << zlibVersion();
```
DennisOSRM, line 37: No blocker, but I could imagine this might break a few folks pragmatic uptime monitoring.

DennisOSRM (approved):

include/server/server.hpp
TheMarex: I would hope they would at least check for `[info] running and waiting for requests` instead. :laughing:

MarcelloPerathoner:
I'm afraid we have memory leaks:
[lsan.log](https://github.com/user-attachments/files/24724150/lsan.log)

TheMarex:
Are they new though? I can maybe squeeze in some time this week to take a closer look, might just be some `shared_ptr` stuff leaking somehow. Most of our memory usage is static so with enough requests we should see something more obvious pop up.

MarcelloPerathoner:
They are new since a few days ago. I have a debug-asan task running in my private repo, failing now.

danpat:
@TheMarex another thing to mention: a while back, @SiarheiFedartsou tried basically the same and we ended up reverting it, see https://github.com/Project-OSRM/osrm-backend/pull/6407

IIRC the main issue was that the boost::beast HTTP parser had some upper limit on the GET request size it would accept, and this broke quite a few users making very large GET `/table` requests.  There's some discussion in the linked PR above.

TheMarex:
@danpat Haha, nothing new under the sun. :laughing: Well, fortunately this is actually configurable in 2026 with the less obviously named `header_limit` on the request parser. The current limit works for the large test cases we configured in cucumber (that is how I discovered it) and should be the upper limit of what other HTTP servers would allow, but maybe we should make it configurable for running locally.

I'll prepare a separate PR for that, I think we can make this seamless by defaulting to some number derived from the coordinate limits.

TheMarex:
@MarcelloPerathoner Took a quick look at it, seems harmless since essentially the root cause is that the `std::shared_ptr<Server>` has ref count > 1 when it is supposed to be deallocated cleanly on shutdown. Looks like we have some lingering reference in the connection pool or something, I'll dive deeper to make sure we shutdown cleanly. But there should not be any real memory leak - at least running it with a memory profiler for an extensive period did not increase the memory usage.

PR #7331, opened 2026-01-13
Add audible fence exception for barrier routing

### LINKED ISSUE #6964: Audible fences should not be barriers
Investigating a routing issue I discovered that the node in question (431495501) was described thus:
barrier=fence
sensory=audio
------------------------------
 (I have also seen sensory=audible)

In Australia we have audible fences that are used to prevent livestock (and wild dogs etc) from passing the barrier. These fences do not stop traffic (and are not meant to) as there is no physical fence.
I question if these should be declared as barriers.
The alternative would be to add the access=yes but that looks wrong to me.
To reproduce use this  example - the route along the Balonne Highway.
https://map.project-osrm.org/?z=10&center=-27.957411%2C147.402191&loc=-27.965088%2C148.163815&loc=-27.993189%2C146.766357&hl=en&alt=0&srv=0
----------------------------------------------------------
One suggested change to car can address this issue
git diff  profiles/car.lua
index 95052d0ec..804087dc2 100644
--- a/profiles/car.lua
+++ b/profiles/car.lua
@@ -352,9 +352,14 @@ function process_node(profile, node, result, relations)
       local flat_kerb = kerb and ("lowered" == kerb or "flush" == kerb)
       local highway_crossing_kerb = barrier == "kerb" and highway and highway == "crossing"
 
+      -- make an exception for fence with sensory = audible
+      local sensory = node:get_value_by_key("sensory")
+      local fence_audible = "fence" == barrier and ("audible" == sensory or "audio" == sensory)
+
       if not profile.barrier_whitelist[barrier]
                 and not rising_bollard
                 and not flat_kerb
+                and not fence_audible
                 and not highway_crossing_kerb
                 or restricted_by_height then
         result.barrier = true
@@ -385,7 +390,6 @@ function process_way(profile, way, result, relations)
     bridge = way:get_value_by_key('bridge'),
     route = way:get_value_by_key('route')
   }

afarber:
# Issue

Fixes #6964

In Australia, audible fences use sound to deter livestock and wild animals but have no physical barrier. They are tagged in OSM as `barrier=fence` with `sensory=audible` or `sensory=audio`. OSRM currently treats all fences as impassable, causing incorrect routing.

This PR adds an exception for audible fences in the car, bicycle, and foot profiles.

## Tasklist

- [ ] review
- [ ] adjust for comments

## Requirements / Relations

None


DennisOSRM (commented):
Change looks good to me. Perhaps adding a test to the bike profile makes sense, too.

PR #7334, opened 2026-01-16
Apply configurable penalty to gates in car profile

### LINKED ISSUE #6757: barrier=gate and barrier=lift_gate should not be whitelisted in car profile
Hi,
currently in car profile **barrier=gate** and **barrier=lift_gate** are per-se whitelisted which means one can not use  access tag on that note. 

These 2 types of barriers should be processed with access tags in mind as they may be typically open so usable by the public, or they may be typically closed meaning its not usable by the public.

With the whitelisting in place there is no way to prohibit the OSRM car routing from using that path other then tagging the way, which may be completely okay to use for the public.

So please remove the  whitelisting for these 2 types of barriers.

This is the commit i am running now with:

https://github.com/flohoff/osrm-backend/commit/0eabdc96ff15ac2561edecb56dfa06361fcb228b

Flo

afarber:
# Issue

Fixes #6757                                                                                                                                                                    
                                                                                                                                                                                 
The car profile currently whitelists `barrier=gate` and `barrier=lift_gate`, making them unconditionally passable when no access tag is present. This is problematic because many real-world gates lack explicit access tags but are not publicly accessible.                                                                                                        
                                                                                                                                                                                     
This PR applies a configurable time penalty (default 60 seconds) to gates:                                                                                                         
                                                                                                                                                                                     
- Gates without access tags are routable but incur a 60-second penalty                                                                                                             
- Gates with `access=yes/permissive/designated` bypass the penalty (explicitly open)                                                                                               
- Gates with `access=no` remain blocked                                                                                                                                            
- Penalty is configurable via `profile.barrier_penalties` table                                                                                                                    
- Adds new `Gate` obstacle type for explicit handling                                                                         
                                                                                                                                                                                 
## Tasklist

 - [ ] review
 - [x] adjust for comments


## Requirements / Relations

- https://github.com/valhalla/valhalla/pull/3078
- https://github.com/valhalla/valhalla/pull/3154                                                                                      
- https://github.com/graphhopper/graphhopper/pull/2345
- https://github.com/graphhopper/graphhopper/issues/168

### DIFF
--- a/profiles/car.lua
+++ b/profiles/car.lua
@@ -364,17 +364,20 @@ function process_node(profile, node, result, relations)
       local sensory = node:get_value_by_key("sensory")
       local audible_fence = barrier == "fence" and sensory and (sensory == "audible" or sensory == "audio")
 
+      -- check if barrier has a configurable penalty (e.g., gates)
+      local barrier_penalty = profile.barrier_penalties[barrier]
+
       if not profile.barrier_whitelist[barrier]
                 and not rising_bollard
                 and not flat_kerb
                 and not highway_crossing_kerb
                 and not audible_fence
+                and not barrier_penalty
                 or restricted_by_height then
         obstacle_map:add(node, Obstacle.new(obstacle_type.barrier))
       end
 
       -- apply configurable penalty to gates/lift_gates
-      local barrier_penalty = profile.barrier_penalties[barrier]
       if barrier_penalty then
         obstacle_map:add(node, Obstacle.new(obstacle_type.gate,
                                             obstacle_direction.both, barrier_penalty, 0))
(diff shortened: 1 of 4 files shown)


TheMarex:
Hrm, can you say more why you think this is different in OSRM then in all other routing engines you referenced? I actually tried this on `openstreetmap.org` and the result seems the same between OSRM / Valhalla / Graphhopper.

https://www.openstreetmap.org/directions?engine=fossgis_osrm_car&route=52.551822%2C13.349682%3B52.552262%2C13.349802

The gate in question doesn't have an access tag and default accessibility is assumed for all engines.

> This is problematic because many real-world gates lack explicit access tags but are not publicly accessible.

Yeah but the problem is the inverse is also true: There are _tons_ of parking lots that are accessible and protected with barriers. Unfortunately a lot of POIs would need to snap to those parking lots for correct routing (e.g. to the Lidl above). Making these parking lots unaccessible will mean it snaps to the nearest road which can lead to odd results (snap to a backstreet).

A better fix would be to add a penalty in if the access is unclear. Looking at the profile I think we are already doing this? (see `obstacles`) So I'm not clear in which scenario this was causing wrong results.

afarber:
@TheMarex updated to use a 60-second penalty instead of blocking. Is this what you had in mind?

profiles/car.lua
```diff
@@ -369,6 +369,13 @@ function process_node(profile, node, result, relations)
                 or restricted_by_height then
         obstacle_map:add(node, Obstacle.new(obstacle_type.barrier))
       end
+
+      -- apply penalty to gates without explicit permissive access
+      -- discourages routing through gates while still allowing it when necessary
+      if barrier == 'gate' or barrier == 'lift_gate' then
```
TheMarex: This is in the right direction, couple of things though:
1. Penalty should to be configurable in the profile. Something like a map `profile.barrier_penalties` maybe?
2. You could add `Gate` in `obstacles.hpp` to make this more explicit.
3. The tests actually don't verify the new behavior, since access is not changed.

This would also unlock a new turn-instruction "go through gate" or something if we wanted to add that at some point.

PR #7336, opened 2026-01-21
Make maximum header size configurable

TheMarex:
# Issue

As @danpat pointed our [here](https://github.com/Project-OSRM/osrm-backend/pull/7328#issuecomment-3775213663) in a previous attempt to port to boost beast, large requests did not work. However in 2026 it is possible to configure the header size (which includes the GET line) so even humongous 10000 table requests are possible.

Configuring this is a bit of a pain, so this is automatically derived from the coordinate limits that people set for `osrm-routed` at start-up. The lower ceiling is 8KB which is well above the typical request. You can override it with `--max-header-size` if you are targeting a specific proxy's maximum header size.

I've verified 10k requests on Bavaria with MLD work fine, but maybe someone wants to throw their particular use-case at this to check if my header-size heuristic works well. (Note: I assume you are rounding your coordinates correctly to 6 decimals)

## Tasklist

 - [ ] review
 - [ ] adjust for comments

## Requirements / Relations

https://github.com/Project-OSRM/osrm-backend/pull/7328

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,9 +1,12 @@
 # Unreleased
   - Changes from 6.0.0
+    - Routing:
+      - FIXED: Crash when route starts or ends at `type=manoeuvre` relation via node [#7287](https://github.com/Project-OSRM/osrm-backend/issues/7287)
     - Profiles:
       - ADDED: Add exception for audible fences (`barrier=fence` with `sensory=audible` or `sensory=audio`) that deter livestock but do not block vehicles [#6964](https://github.com/Project-OSRM/osrm-backend/issues/6964)
       - ADDED: Use `is_sidepath:of:name` and `street:name` as fallback names for unnamed sidewalks and sidepaths in foot and bicycle profiles [#7259](https://github.com/Project-OSRM/osrm-backend/issues/7259)
     - Build:
+      - FIXED: Set `hwloc:shared=True` in Conan config as required by onetbb [#7342](https://github.com/Project-OSRM/osrm-backend/issues/7342)
       - CHANGED: Cucumber tests now can run in parallel and other improvements [#7309](https://github.com/Project-OSRM/osrm-backend/issues/7309)
       - FIXED: Update Node.js binding path from `lib/binding` to `lib/binding_napi_v8` to match node-pre-gyp versioning conventions [#7272](https://github.com/Project-OSRM/osrm-backend/pull/7272)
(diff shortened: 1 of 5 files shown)


danpat:
> I assume you are rounding your coordinates correctly to 6 decimals

https://xkcd.com/2170/ talks about this kind of assumption :-P

Why not just make it <very large number> by default, but allow it to be set smaller for people that want to put some DOS protection in place?

src/tools/routed.cpp
```diff
@@ -99,6 +99,21 @@ void validate(boost::any &v, const std::vector<std::string> &values, double *, d
...
+    constexpr unsigned MIN_HEADER_SIZE = 8 * 1024;
+
+    int max_coordinates = 0;
+    max_coordinates = std::max(max_coordinates, config.max_locations_trip);
+    max_coordinates = std::max(max_coordinates, config.max_locations_viaroute);
+    max_coordinates = std::max(max_coordinates, config.max_locations_distance_table);
+    max_coordinates = std::max(max_coordinates, config.max_locations_map_matching);
+
+    // for each coordinate we will need up to (3 (major) + 1 (dot) + 6 (decimals)) * 2 plus 1
+    // semi-colon = 21 chars we add a generous 1024 chars for the request line
```
jcoupey: The 21 chars estimate here only accounts for the "locations" part of the request, while there are request options (e.g. `approaches`  and `radiuses`) that, if used, also make the size increase linearly.

Not sure how to go about that?

TheMarex: Fair point, come to think of it also `sources=` and `destinations=`. So that would be an additional 5 chars for each `sources`+`destinations`, maybe 4 chars for radiuses and a whooping 13 chars for approaches per coordinate, 48 chars total. I'll update the heuristic to include that. Good catch!

TheMarex:
@danpat Yeah I've been going back and forth on this, but my argument is that switching to a production-grade HTTP server and then configuring it insecure by default doesn't make a lot of sense either.

I think the automatic limit should be fine for almost everyone, with @jcoupey's suggestions there should be very few sensible requests that would ever require touching `--max-header-size`. Maybe people will finally round their coordinates correctly and realize how much overhead that saves. :upside_down_face:

src/tools/routed.cpp
```diff
@@ -99,6 +99,25 @@ void validate(boost::any &v, const std::vector<std::string> &values, double *, d
 }
 } // namespace boost
 
+inline unsigned deriveMaxHeaderSize(const EngineConfig &config)
```
danpat: A unit test for this function would be nice :-)

TheMarex: Added some tests. :smile:

PR #7339, opened 2026-01-21
Fix crash when route starts/ends at manoeuvre relation via node

### LINKED ISSUE #7287: Assertion `!hasTurnType(steps.front()) && !hasTurnType(steps.back())' failed.
I am running 560ed5c0e82d7bd6a0887eb58188bf4fa8662372 on a planet-wide dataset with CH and the car profile, and there is one specific, but innocent-looking routing request (with steps=true&geometries=polyline&overview=full) that consistently triggers the above assertion.

I am presently debugging this further and will report findings.

afarber:
# Issue

Fixes #7287                                                                                                                    
                                                                                                                                 
Routes starting or ending exactly at the via node of a `type=manoeuvre` relation would crash with an assertion failure.

Manoeuvre overrides were applied to depart/arrive steps, setting a turn type on waypoint steps that should always have `NoTurn`.

- Skip manoeuvre overrides on waypoint steps (depart/arrive) in `post_processing.cpp`                                                                                  
- Keep defensive early return in `collapseSegregatedTurnInstructions` as defense-in-depth 

## Tasklist

 - [ ] review
 - [ ] adjust for comments


## Requirements / Relations

- https://github.com/valhalla/valhalla/pull/1796                                                                                                                               
- https://github.com/valhalla/valhalla/pull/2600                                                                                                                               
- https://wiki.openstreetmap.org/wiki/Relation:manoeuvre

### DIFF
--- a/src/engine/guidance/post_processing.cpp
+++ b/src/engine/guidance/post_processing.cpp
@@ -683,8 +683,8 @@ void applyOverrides(const datafacade::BaseDataFacade &facade,
                     // Depart/arrive steps mark route endpoints and should keep NoTurn type.
                     if (hasWaypointType(*step_to_update))
                     {
-                        util::Log(logDEBUG) << "Skipping maneuver override on waypoint step"
-                                            << std::endl;
+                        util::Log(logDEBUG)
+                            << "Skipping maneuver override on waypoint step" << std::endl;
                         continue;
                     }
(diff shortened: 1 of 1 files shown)


TheMarex:
Do you think there is a way to add a regression test for this to the cucumber test suite? While this may be more of a workaround I'm tempted to accept it since I don't think anyone truly understands the guidance code anymore.

PR #7348, opened 2026-01-29
Fix documentation build ESM/CommonJS conflict and tar vulnerability

### LINKED ISSUE #7347: Documentation build fails: "require is not defined in ES module scope"
The `npm run docs` command fails in CI with:                                                                                                             
                                                                                                                                                           
>  ReferenceError: require is not defined in ES module scope, you can use import instead

This file is being treated as an ES module because it has a '.js' file extension and '/home/runner/work/osrm-backend/osrm-backend/package.json' contains "type": "module". 
                                                                                                                                                           
Root cause: #7233 added `"type": "module"` to package.json. The docs build uses Babel to transpile docbox's files to CommonJS, but Node.js still treats the `.js` output as ESM.

afarber:
# Issue

Fixes #7347

- Fix docs build by using Babel's `--out-file-extension .cjs` flag so transpiled output is treated as CommonJS                                           
- Additionally, run "npm audit --fix" to fix GHSA-34x7-hfp2-rc4v vulnerability            

## Tasklist

 - [ ] review
 - [x] adjust for comments

## Requirements / Relations

- https://nvd.nist.gov/vuln/detail/CVE-2026-24842
- https://github.com/advisories/GHSA-34x7-hfp2-rc4v

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@
       - ADDED: Add exception for audible fences (`barrier=fence` with `sensory=audible` or `sensory=audio`) that deter livestock but do not block vehicles [#6964](https://github.com/Project-OSRM/osrm-backend/issues/6964)
       - ADDED: Use `is_sidepath:of:name` and `street:name` as fallback names for unnamed sidewalks and sidepaths in foot and bicycle profiles [#7259](https://github.com/Project-OSRM/osrm-backend/issues/7259)
     - Build:
+      - FIXED: Documentation build fails with ESM/CommonJS conflict after adding "type": "module" [#7347](https://github.com/Project-OSRM/osrm-backend/issues/7347)
       - FIXED: Set `hwloc:shared=True` in Conan config as required by onetbb [#7342](https://github.com/Project-OSRM/osrm-backend/issues/7342)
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -14,6 +14,9 @@
       "brfs"
     ]
   },
+  "overrides": {
```
TheMarex: Hm do you know why this was necessary? A bump to the lock file should be sufficient. `npm audit --fix` should do the right thing by default.

BTW this is the third critical node-tar vulnerability in 2 weeks WTF.

afarber: Because `npm audit --fix` only updates `package-lock.json` - and nobody looks closely at that file (it is quasi generated, but in the repo)

afarber: BTW please consider publishing 6.0.1 release

TheMarex: I don't follow. Nobody has to look at the lock file, npm uses it automatically for the consistent (transitive) dependency resolution. If you put this in the `package.json` it permanently override the version of indirect dependencies. That is not such a great idea if the dependency that relies on node-tar bumps to a higher minor version.

afarber: Ok, reverted

TheMarex:
Sorry forgot about this. Looks good now.

PR #7358, opened 2026-02-16
Replace boost::format with fmt::format across codebase

DennisOSRM:
Replace `boost::format` with `std::format` as part of the eternal quest to lower the number of dependencies.


DennisOSRM:
Superseded by #7363

PR #7401, opened 2026-03-05
Es6 nodejs tests

MarcelloPerathoner:
Update NodeJS tests

- Incorporate eslint suggestions: arrow operator, single quotes, indent=2, etc.
- Load OSRM once only
- Add an environment variable to load OSRM from a configurable directory.

The last point enables testing from a `cmake --install` directory, so as to catch missing dynamic libraries, wrong RPATH etc.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -8502,9 +8502,9 @@
       "license": "Apache-2.0"
     },
     "node_modules/flatted": {
-      "version": "3.3.3",
-      "resolved": "https://registry.npmjs.org/flatted/-/flatted-3.3.3.tgz",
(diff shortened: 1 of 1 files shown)


DennisOSRM:
Is there a chance we could separate this into two prs, one for reformatting and another for the functional change?

MarcelloPerathoner:
The functional change is contained entirely in the file `constants.js` and in the first few lines of each other file, those that say `import`.  The rest is formatting only.

DennisOSRM:
Could we do this in two PRs? The first changes formatting and the second the functional change. 

I know this is a bit cumbersome, but provides more clarity during debugging.

PR #7424, opened 2026-03-22
Exclude highway=road from foot and bicycle profiles

### LINKED ISSUE #7039: `highway=road` should be excluded from all routing profiles
### Issue

According to the OpenStreetMap `highway=road` [wiki](https://wiki.openstreetmap.org/wiki/Tag:highway%3Droad#Rendering_Support), roads with an unknown classification should be considered errors and shouldn't be routable.

Currently, on the OSM site, the OSRM demo shows that while the car profile correctly excludes such roads, the foot and bicycle profiles still include them.

### Steps to Reproduce

1. Go to: [OSM Directions](https://www.openstreetmap.org/directions?engine=fossgis_osrm_foot&route=18.723417%2C98.938268%3B18.721028%2C98.935925).
2. Switch between Bicycle (OSRM), Car (OSRM), and Foot (OSRM) profiles.
3. Notice that the Bicycle and Foot profiles don't exclude the unknown road between the start and end points, whereas the Car profile does.

DennisOSRM:
highway=road denotes a way with unknown classification and implies fixme=*. Per the OSM wiki it is an error marker and should not be routed over. The car profile already excluded it; this commit brings the foot and bicycle profiles in line.

Changes:
- Remove 'road' from the speeds table in profiles/foot.lua
- Remove 'road' from bicycle_speeds in profiles/bicycle.lua
- Update cucumber tests to expect highway=road to be non-routable

Fixes #7039

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
       - ADDED: Emit warning when ways reference nodes not present in input data [#1596](https://github.com/Project-OSRM/osrm-backend/issues/1596)
       - FIXED: Compilation error in raster_source [#7422](https://github.com/Project-OSRM/osrm-backend/issues/7422)
     - Profiles:
+      - FIXED: Exclude `highway=road` from foot and bicycle profiles since it is an OSM error marker with unknown classification [#7039](https://github.com/Project-OSRM/osrm-backend/issues/7039)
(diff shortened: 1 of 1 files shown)


DennisOSRM:
@copilot add a line to the change log in the style of the existing entries

PR #7433, opened 2026-03-26
Add Elevator obstacle type to prevent node collapsing

### LINKED ISSUE #7434: Elevator nodes are silently collapsed during graph compression
## Problem

OSRM's graph compressor collapses degree-2 nodes to reduce graph size. Elevators (highway=elevator in OSM) are typically degree-2 nodes connecting different levels, and they get collapsed away during this phase.

When an elevator node is compressed out of the graph, profiles cannot apply a duration/weight penalty for the elevator (wait + ride time).

This primarily affects foot routing in buildings, pedestrian underpasses, and transit stations where elevators are mapped.

## Current workarounds

There is no clean way to prevent this today. A profile could emit a false Barrier obstacle on elevator nodes via the obstacle_map API, but barriers are semantically different — they block traversal rather than penalize it — and downstream code checking for actual barriers would get false positives.

## Impact

- Foot profiles cannot accurately model elevator travel time, leading to incorrect ETAs on routes that include elevators
- No existing obstacle type maps to the elevator concept — it's neither a barrier (passable, but slow), nor a traffic signal (not directional), nor a turning point

mattrobmattrob:
# Description

Add `Obstacle::Type::Elevator` (`0x0800`) to the Incompressible bitmask so that elevator nodes are preserved during graph compression. This allows foot profiles to apply duration penalties and emit guidance for elevators (`highway=elevator`) instead of having them silently collapsed away.

Changes:
- Add `Elevator` to `Obstacle::Type` enum and Incompressible bitmask
- Register "elevator" in the Lua obstacle_type enum
- Add unit test verifying elevator nodes resist compression

Profile usage:
```lua
    function process_node(profile, node, result)
        local highway = node:get_value_by_key("highway")
        if highway == "elevator" then
            obstacle_map:add(node, Obstacle(obstacle_type.elevator))
        end
    end
```

# Issue

Closes #7434.

Was this change primarily generated using an AI tool?

Claude Code was used as part of this change.

## Tasklist

 - [x] self-review code for correctness and following the [coding guidelines](https://github.com/Project-OSRM/osrm-backend/wiki/Coding-Standards)
 - [x] add tests (see [testing](https://github.com/Project-OSRM/osrm-backend/blob/master/docs/testing.md))
 - [x] update relevant [wiki pages](https://github.com/Project-OSRM/osrm-backend/wiki)
 - [x] CHANGELOG.md entry (see [how to](http://keepachangelog.com/en/1.0.0/#how))
 - [ ] review
 - [ ] adjust for comments

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
       - FIXED: Prevent MLD route queries from updating removed heap nodes, avoiding `osrm-routed` segfaults/asserts [#7203](https://github.com/Project-OSRM/osrm-backend/issues/7203)
       - FIXED: Crash when route starts or ends at `type=manoeuvre` relation via node [#7287](https://github.com/Project-OSRM/osrm-backend/issues/7287)
     - Extraction:
+      - ADDED: `Elevator` obstacle type preventing degree-2 node collapsing during graph compression [#7434](https://github.com/Project-OSRM/osrm-backend/issues/7434)
       - ADDED: Emit warning when ways reference nodes not present in input data [#1596](https://github.com/Project-OSRM/osrm-backend/issues/1596)
       - FIXED: Compilation error in raster_source [#7422](https://github.com/Project-OSRM/osrm-backend/issues/7422)
     - Profiles:
--- a/include/extractor/obstacles.hpp
+++ b/include/extractor/obstacles.hpp
@@ -71,9 +71,10 @@ struct Obstacle
         TurningCircle = 0x0100,
         StopMinor = 0x0200,
         Gate = 0x0400,
+        Elevator = 0x0800,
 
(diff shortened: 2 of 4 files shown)


mattrobmattrob:
This isn't necessary (it turns out) because the following can work (even in 2-edge compressed scenarios):
```lua
function process_node(profile, node, result)
    local highway = node:get_value_by_key("highway")
    if highway == "elevator" then
        obstacle_map:add(node, Obstacle.new(
            obstacle_type.gate, obstacle_direction.both,
            profile.properties.elevator_penalty, profile.properties.elevator_penalty))
    end
end

function process_turn(profile, turn)
    for _, obs in pairs(obstacle_map:get(turn.from, turn.via)) do
        penalty = penalty + obs.duration
    end
end
```

Issue #7308, opened 2025-12-21
osrm-routed fail


lilishiying-shiying:
# Issue

(base) D:\osrm-backend\build\Release>osrm-customize.exe "d:\openstreet\asia\japan\japan.osrm"
[2025-12-21T10:47:42] [info] Loaded edge based graph: 98432974 edges, 23875107 nodes[0m
[2025-12-21T10:47:42] [info] Loading partition data took 24.5458 seconds[0m
[2025-12-21T10:48:22] [info] Cells customization took 40.253 seconds[0m
[2025-12-21T10:48:22] [info] Cells statistics per level[0m
[2025-12-21T10:48:23] [info] Level 1 #cells 106488 #boundary nodes 3145238, sources: avg. 19, destinations: avg. 28, entries: 73304496 (586435968 bytes)[0m
[2025-12-21T10:48:23] [info] Level 2 #cells 8863 #boundary nodes 562953, sources: avg. 43, destinations: avg. 60, entries: 27664239 (221313912 bytes)[0m
[2025-12-21T10:48:23] [info] Level 3 #cells 561 #boundary nodes 67817, sources: avg. 81, destinations: avg. 111, entries: 6410632 (51285056 bytes)[0m
[2025-12-21T10:48:23] [info] Level 4 #cells 23 #boundary nodes 2849, sources: avg. 82, destinations: avg. 108, entries: 282730 (2261840 bytes)[0m
[2025-12-21T10:48:23] [info] Unreachable nodes statistics per level[0m
[33m[2025-12-21T10:48:23] [warn] Level 1 unreachable boundary nodes per cell: 0.000122079 sources, 0.000431974 destinations[0m
[33m[2025-12-21T10:48:23] [warn] Level 2 unreachable boundary nodes per cell: 0.000451314 sources, 0.00180526 destinations[0m
[33m[2025-12-21T10:48:23] [warn] Level 3 unreachable boundary nodes per cell: 0.00356506 sources, 0.00891266 destinations[0m
[2025-12-21T10:48:23] [info] Unreachable nodes statistics per level[0m
[33m[2025-12-21T10:48:23] [warn] Level 1 unreachable boundary nodes per cell: 0.0788258 sources, 0.0612933 destinations[0m
[33m[2025-12-21T10:48:23] [warn] Level 2 unreachable boundary nodes per cell: 0.608146 sources, 0.448719 destinations[0m
[...]

PR #7437, opened 2026-03-29
Fix Windows handling of large files

861:
# Issue

This PR is targeting OSRM windows build.  issue #7308 

This issue was caused by fseek function in microtar library. According to microsoft [doc ](https://learn.microsoft.com/en-us/cpp/c-runtime-library/reference/fseek-fseeki64?view=msvc-170), on windows _fseeki64 is an better option.

### DIFF
--- a/third_party/microtar/src/microtar.c
+++ b/third_party/microtar/src/microtar.c
@@ -209,7 +209,12 @@ static int file_read(mtar_t *tar, void *data, mtar_size_t size) {
 }
 
 static int file_seek(mtar_t *tar, mtar_size_t offset) {
-  int res = fseek(tar->stream, offset, SEEK_SET);
+  int res;
+#if defined(_WIN32)
+  res = _fseeki64(tar->stream, (__int64) offset, SEEK_SET);
+#else
+  res = fseek(tar->stream, offset, SEEK_SET);
+#endif
   return (res == 0) ? MTAR_ESUCCESS : MTAR_ESEEKFAIL;
 }
 


861:
@DennisOSRM  I tested the entire US dataset today using OSRM V6.0.0 with this patch, it is working now on Windows 11 and Windows server 2019.

third_party/microtar/src/microtar.c
```diff
@@ -209,7 +209,12 @@ static int file_read(mtar_t *tar, void *data, mtar_size_t size) {
 }
 
 static int file_seek(mtar_t *tar, mtar_size_t offset) {
-  int res = fseek(tar->stream, offset, SEEK_SET);
+  int res;
+#if defined(_WIN32)
```
DennisOSRM, line 213: No need to support 32 bit environment. We dropped support some time ago.

861: Marco _WIN32 is defined for both 32-bit and 64-bit. Do you want me to update it to _WIN64 for 64-bit only?

DennisOSRM: If I understand this right, then we don‘t need the `if defined` block and can replace it with this single line:

`int res = _fseeki64(tar->stream, (__int64) offset, SEEK_SET); `

861: _fseeki64 is windows only. Not available in linux, so if the os is not windows, we still need to use fseek.

DennisOSRM: Ah! Thanks for the clarification. 

So, yes, if there's a WINDOWS or WIN64 define to indicate Windows environment, then let's use that. Otherwise, your patch looks good.

861:
> Thank you for your contribution to the code base

No problem. I am very glad i can make contribution.

PR #7442, opened 2026-04-02
fix(contractor): insert self-loops for all needed cases, not just one-ways

MarcelloPerathoner:
# Issue

The shortest distance between any two nodes must be invariant under contraction.  We must also keep invariant the shortest loop distance from any node back to itself.  This requirement arises for us from the need to "go around" if source and target are on the same node, with source downstream from target.  For this reason we must insert self-loops whenever this is the shortest path from self to self.

The current implementation only inserts self-loops for one-way streets, which is insufficient. Any road can be turned into an implicit one-way by any combination of turn restrictions, bearings, or continue-straight-at-waypoints.

## Tasklist

 - [ ] self-review code for correctness and following the [coding guidelines](https://github.com/Project-OSRM/osrm-backend/wiki/Coding-Standards)
 - [ ] add tests (see [testing](https://github.com/Project-OSRM/osrm-backend/blob/master/docs/testing.md))
 - [ ] update relevant [wiki pages](https://github.com/Project-OSRM/osrm-backend/wiki)
 - [ ] CHANGELOG.md entry (see [how to](http://keepachangelog.com/en/1.0.0/#how))
 - [ ] review
 - [ ] adjust for comments


MarcelloPerathoner:
I did some preliminary benchmarks. The added self-loops increase the no. of edges by ~3%. That in turn increases search time by ~8%. Surprisingly the contractor now uses far less memory ~57%.

Contractor benchmark

| log                                 |    edges |   norm |   time (s) |   norm |   mem (MB) |   norm |
|-------------------------------------|----------|--------|------------|--------|------------|--------|
| /tmp/origin-master-germany.log      | 96173352 |  1.000 |     883.02 |  1.000 |      13755 |  1.000 |
| /tmp/contractor-rewrite-germany.log | 99390191 |  1.033 |     853.71 |  0.967 |       7836 |  0.570 |

Routed benchmark (requests routes between randomly selected big cities in Germany)

| log                                        |   time (ms) |   norm |   distance |   norm |
|--------------------------------------------|-------------|--------|------------|--------|
| /tmp/routed-origin-master-germany.log      |        2.97 |  1.000 | 323295.350 |  1.000 |
| /tmp/routed-contractor-rewrite-germany.log |        3.21 |  1.082 | 323295.350 |  1.000 |

MarcelloPerathoner:
These two embarassingly simple tests fail in the current implementation:

https://github.com/MarcelloPerathoner/osrm-backend/blob/c768db25513a870e4e57eee908c066bb75a6cc15/features/testbot/self_loop.feature#L6-L45

Note that there are no one-way streets involved.

To illustrate the problem, I show the contraction of the single bidirectional road segment: 

With the current implementation:

dot

This contraction is clearly incorrect because it leaves no edge to exit node `1`. If source and target phantoms are both on node `1`, and source is downstream from target, there is no way for either forward nor backward search to leave node `1`. No route can be found. (The bidirectional edge is stored on node `0` only.)

With this PR:

dot

With the added self-loop, the forward search can now leave node `1` and return to it.

DennisOSRM:
Could you elaborate on the reason for this PR? I gave probably missed something, but what behavior does it correct?

MarcelloPerathoner:
The current implementation does not find routes that obviously exist. The reason is that the contractor does not always preserve the shortest path from a node back to itself.

The contractor inserts a self-loop only if the node represents a one-way street. That is insufficient, as the problem does surface most easily with one-way streets, but is not in any way limited to them.

The above mentioned test scenarios show how obvious routes are missed.

DennisOSRM:
This is looking good and will go in once CI comes back green. Thanks so much for the contribution and your patience. It took a bit to get this over the finish lines.

Added a regression test: features/testbot/self_loop.feature — new scenario "Waypoints snap to same node, source downstream"

Node map (matching the PR comment diagram):
```
 1   2
a-----b
```

A single bidirectional residential road between two nodes an and b. Both waypoints are on the same (upper) side.

Test cases:

| waypoints | approaches | expected route | notes                                                                                     |
|-----------|------------|----------------|-------------------------------------------------------------------------------------------|
| 1,2       | curb curb  | ab,ab,ab,ab    | source near a, target near b: source downstream on b→a edge → needs self-loop            |
| 2,1       | curb curb  | ab,ab          | Source near b, target near a: normal forward direction on b→a edge → no self-loop needed |

Verification results:
- With SELF_LOOPS (patch branch): ✅ 3 scenarios, 18 steps all pass
- Without SELF_LOOPS (master behavior): ❌ 1,2 curb curb returns empty route — forward search cannot leave the edge-based node, no route found

DennisOSRM (approved):

PR #7456, opened 2026-04-08
fix(mem usage): XORFastHash lost lines

MarcelloPerathoner:
Lines that should have been included in the previous PR.


src/contractor/graph_contractor.cpp
```diff
@@ -98,21 +98,19 @@ struct RemainingNodeData
...
+    explicit ThreadDataContainer() {}
 
     inline ContractorThreadData *GetThreadData()
     {
         bool exists = false;
         auto &ref = data.local(exists);
         if (!exists)
         {
-            ref = std::make_shared<ContractorThreadData>(number_of_nodes);
+            ref = std::make_shared<ContractorThreadData>(8000);
```
DennisOSRM: Is this the main trick to have things run faster and use less mem? How was the magic number 8000 determined?

MarcelloPerathoner: Mostly guesswork. If it is too small you will lock up in and endless loop. If it is too big you waste memory. We have FULL_SEARCH_SPACE_SIZE = 2000, and I run germany-latest with a value of 4000 without locking up. So I doubled that value for safety. 

Of course we could make it configurable: it is a trade-off between speed and memory. (and actually it rounds up internally to a power of 2, so it is in effect 8192)

src/contractor/graph_contractor.cpp
```diff
@@ -98,21 +98,19 @@ struct RemainingNodeData
...
         }
```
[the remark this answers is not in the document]
MarcelloPerathoner: We have a hard-coded search space limit of 2000. A storage capacity of 4000 was shown to accomodate a graph the size of germany and the value of 8000 is double that. The suggestion of using number of nodes is ridiculous as then we could use an array instead of a hash table.

DennisOSRM (commented):
There are two things that are worth implementing: guarding against a full table, and making the power of two more explicit.

First, let's focus on defending against the table running full. 

Detection (guard inside the loop):                                                                                                                                                                                                       
   
The simplest fix is a probe counter. Since the table has mask + 1 slots, if you've probed that many times without finding an empty slot or a match, the table is full:                                                                   
                  
```
  ValueType &operator[](const KeyType key)                                                                                                                                                                                                 
  {               
      std::size_t position = key & mask;
      std::size_t probes = 0;                                                                                                                                                                                                              
      while ((cells[position].time == current_timestamp) && (cells[position].key != key))
      {                                                                                                                                                                                                                                    
          ++position &= mask;
          BOOST_ASSERT_MSG(++probes <= mask, "LinearHashStorage is full");
      }                                                                                                                                                                                                                                    
      // ...
  }            
```                                                                                                                                                                                                                            
                  
This turns the infinite loop into a loud, diagnosable crash in debug builds.                                                                                                                                                             
   
Next, let's focus on the related case of preventing overflow to happen unnoticed.

Prevention (track occupancy):                                                                                                                                                                                                            
                  
Add an insertion counter. Increment it when a new key is claimed, reset it in Clear(). Then assert before probing that there's room:                                                                                                     
   
```
  std::size_t occupied = 0;                                                                                                                                                                                                                
                                                                                                                                                                                                                                           
  ValueType &operator[](const KeyType key)
  {                                                                                                                                                                                                                                        
      BOOST_ASSERT_MSG(occupied <= mask, "LinearHashStorage is full");
      std::size_t position = key & mask;                                                                                                                                                                                                   
      // ...
      if (p->time != current_timestamp)                                                                                                                                                                                                    
      {           
          ++occupied;
          // claim slot                                                                                                                                                                                                                    
      }
      return p->value;                                                                                                                                                                                                                     
  }               
                                                                                                                                                                                                                                           
  void Clear()    
  {
      occupied = 0;
      ++current_timestamp;                                                                                                                                                                                                                 
      // ...
  }            
```

Now, for the power of two being explicit. The cleanest approach is to define it as an explicit power of two and enforce the invariant at compile time, eg. something like the following:

```
static constexpr std::size_t HASH_MAP_CAPACITY = 1u << 13; // 8192

static_assert((HASH_MAP_CAPACITY & (HASH_MAP_CAPACITY - 1)) == 0,
              "HASH_MAP_CAPACITY must be a power of two");
```

2^13 is probably enough for the witness search to conclude even on the higher part of the hierarchy. Setting it to 2^14 should be safe for the planet-scale extracts.

MarcelloPerathoner:
I can add an occupancy counter and then the best would be to limit the search space to < 75% occupancy. We can do that by checking the number of relaxed nodes instead of settled nodes. When the table is full it is too late because performance already drops sharply when the table is nearly full.

As for the size, in the current implementation you can give any size you want and it will internally round up to the next power of two. Requiring it the user to supply an exact power of 2 makes it somehow different than the other storage implementations.

DennisOSRM:
Ok. Makes sense to check early and log a warning. Please note that this guards against running full has a bit of a different goal that the computation of shortcut paths could not be computed as expected. 

Writing to magic number SA power of two has the advantage of capturing the intent in a better way. The use of a telephone will not be able to set anything since it's hard-coded.

MarcelloPerathoner:
With a max. search space of 2000 settled nodes, the max. occupancy for contracting Berlin is 2815, for Germany is 3319. Using a heap size of 8192 and stopping the search at an occupancy of 6000 should be plenty. If you want we can double that. But even if we stop early, the graph is still correct, we only insert more shortcuts than strictly necessary. 

Can you explain why the search space limit is 1000 nodes for simulation and 2000 for real contraction?

DennisOSRM:
Originally, it was a full search without pruning. Limiting the search showed good speedups with a small impact on quality. The simulation case then was similar thinking. There was a time/quality tradeoff.

DennisOSRM:
> Can you explain why the search space limit is 1000 nodes for simulation and 2000 for real contraction?

The simulation feeds into finding the priority of a node. This is an NP hard problem and is thus only approximated. Setting the node budget to 1000 for simulation gives a good tradeoff to determine the priority of a node relative to its neighbours. Setting it to 2000 for the search during contraction is the time/quality tradeoff described above. The numbers themselves were determined empirically when the implementation was done.

Let me know if you have additional questions.

DennisOSRM:
> With a max. search space of 2000 settled nodes, the max. occupancy for contracting Berlin is 2815, for Germany is 3319. Using a heap size of 8192 and stopping the search at an occupancy of 6000 should be plenty. If you want we can double that. But even if we stop early, the graph is still correct, we only insert more shortcuts than strictly necessary.

Let's go with 8096 then. That fills the space ~25% max and should give enough head room.

DennisOSRM:
Do you intend to do further changes, here, or is this ready?

MarcelloPerathoner:
It's ready ... unless the copilot has more to complain

DennisOSRM (approved):

PR #7458, opened 2026-04-08
Replace d3-queue with native async/await across test infrastructure

DennisOSRM:
Removes `d3-queue` dependency and replaces all usages with native `async`/`await` and `Promise`-based patterns. Follow-up to review feedback requiring fully promise-based APIs with no callback mixing.

## Core changes

- **`features/support/data.js`**: `processRowsAndDiff`, `reprocessAndLoadData`, and `buildWaysFromTable` drop their `callback` parameters entirely — errors throw, callers `await`
- **`features/step_definitions/routability.js`**: `testRoutabilityRow` returns a `Promise` instead of calling `cb`; `testRow` is now `async`; step function is `async`
- **All call sites** (`matching.js`, `nearest.js`, `trip.js`, `distance_matrix.js`, `shared_steps.js`, `routing.js`, `requests.js`): step functions converted to `async`, inner `testRow`/`requestRow` functions wrap callback-based HTTP APIs in `new Promise()` and return results directly

## Before / After

```js
// Before: mixed async+callback
async processRowsAndDiff(table, fn, callback) {
  try { /* ... */ callback(); }
  catch (err) { callback(err); }
}
this.processRowsAndDiff(table, testRow, callback); // caller ignores returned Promise

// After: pure async
async processRowsAndDiff(table, fn) {
  for (const [i, row] of table.hashes().entries()) {
    actual.push(await fn(row, i)); // fn returns a Promise
  }
  if (diff) throw new Error(diff);
}
await this.processRowsAndDiff(table, testRow);
```

### DIFF
--- a/features/lib/hash.js
+++ b/features/lib/hash.js
@@ -1,32 +1,21 @@
 // File content hashing utilities for cache invalidation and content verification
 import fs from 'fs';
 import crypto from 'crypto';
-import d3 from 'd3-queue';
 
 // Computes MD5 hash of multiple files concatenated together
-const hashOfFiles = (paths, cb) => {
-  const queue = d3.queue();
-  for (let i = 0; i < paths.length; ++i) {
-    queue.defer(fs.readFile, paths[i]);
-  }
-  queue.awaitAll((err, results) => {
-    if (err) return cb(err);
-    const checksum = crypto.createHash('md5');
-    for (let i = 0; i < results.length; ++i) {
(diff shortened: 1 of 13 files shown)


DennisOSRM:
@copilot apply changes based on the comments in [this thread](https://github.com/Project-OSRM/osrm-backend/pull/7458#pullrequestreview-4080421676). Make it async/await all the way and remove the callback. The functions should be a promise and callsites should await

PR #7467, opened 2026-04-11
Fix microtar tar header size field on MinGW Windows

jrklasen:
On Windows (LLP64), `long` is 32-bit even on x86_64. The `_MSC_VER` guard in microtar correctly used `%llo` for MSVC but fell through to `%lo` for MinGW, causing the bug in the Julia binary of OSRM.

Replace the platform-specific `_MSC_VER` ifdef with portable `uint64_t` and `<inttypes.h>` format macros (`SCNo64`/`PRIo64`), which are correct on all platforms and compilers.

With this fix, we get the tests in the Julia wrapper passing. See [OpenSourceRoutingMachine.jl](https://github.com/moviro-hub/OpenSourceRoutingMachine.jl/pull/19) and the build [recipe](https://github.com/JuliaPackaging/Yggdrasil/pull/13479), which contains the patch currently.

PS: It would be great if we could get a new point release we could refer to downstream once this is merged.

### DIFF
--- a/third_party/microtar/src/microtar.c
+++ b/third_party/microtar/src/microtar.c
@@ -20,6 +20,7 @@
  * IN THE SOFTWARE.
  */
 
+#include <inttypes.h>
 #include <stdio.h>
 #include <stdlib.h>
 #include <stddef.h>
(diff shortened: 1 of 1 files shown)


DennisOSRM (approved):

jrklasen:
Thanks @DennisOSRM for swiftly merging this. Would it be possible to get a point release?

DennisOSRM:
> Thanks @DennisOSRM for swiftly merging this. Would it be possible to get a point release?

new release coming soon

PR #7476, opened 2026-04-13
ci: publish multi-arch docker images as part of monthly release

tete17:
## Summary

- The new monthly release pipeline (`release-monthly.yml`) pushes the release tag using the default `GITHUB_TOKEN`, which GitHub intentionally prevents from triggering downstream workflows. As a result, `osrm-backend-docker.yml` — which is gated on `push: tags: 'v*'` — never runs on automated releases, and no images land on `ghcr.io` for monthly versions.
- The release workflow already handles this gotcha for `osrm-backend.yml` by explicitly calling `gh workflow run ... --ref "$TAG"`. This PR applies the same pattern to the docker workflow so multi-arch images are built and published on every release.
- No changes to the Dockerfiles or build matrix — `osrm-backend-docker.yml` already produces `linux/amd64` + `linux/arm64` images for debian and alpine across normal/debug/assertions variants. The fix is purely trigger wiring.

### Changes

- `.github/workflows/osrm-backend-docker.yml`: add `workflow_dispatch` with a required `ref` input, and point `actions/checkout` at `inputs.ref || github.ref` so `docker/metadata-action` still emits tag-based image tags (e.g. `ghcr.io/project-osrm/osrm-backend:v26.4.0`) when invoked via dispatch.
- `.github/workflows/release-monthly.yml`:
  - New "Trigger docker image build for tag" step mirroring the existing `osrm-backend.yml` dispatch.
  - Refactor the pre-publish wait loop into a `wait_for_workflow` helper that now blocks `npm publish` on **both** `osrm-backend.yml` and `osrm-backend-docker.yml` reaching `completed`/`success`. A release is only "done" when all advertised artifacts exist.

### Why not push the tag with a PAT instead?

That would re-enable the existing `push: tags: 'v*'` trigger without any docker-workflow changes, but it requires maintaining a PAT secret, grants broader write access than `GITHUB_TOKEN`, and silently re-triggers every tag-gated workflow. Following the explicit-dispatch pattern already used for `osrm-backend.yml` has a smaller blast radius.

## Test plan

- [ ] `actionlint .github/workflows/release-monthly.yml .github/workflows/osrm-backend-docker.yml`
- [ ] Dry run: invoke `release-monthly.yml` via `workflow_dispatch` with a sentinel `version_override` on a test branch (with the final `npm publish` step temporarily disabled), and confirm both `osrm-backend.yml` and `osrm-backend-docker.yml` are dispatched and waited on.
- [ ] After the dry run, check `https://github.com/Project-OSRM/osrm-backend/pkgs/container/osrm-backend` for new tags (`vX.Y.Z`, `vX.Y.Z-debug`, `vX.Y.Z-assertions`, plus alpine equivalents).
- [ ] `docker buildx imagetools inspect ghcr.io/project-osrm/osrm-backend:vX.Y.Z` — manifest list should contain both `linux/amd64` and `linux/arm64`.
- [ ] Smoke test: `docker run --rm --platform linux/amd64 ghcr.io/project-osrm/osrm-backend:vX.Y.Z osrm-extract --help`, repeat with `--platform linux/arm64` under QEMU.
- [ ] Clean up the throwaway tag, release, and package versions.


.github/workflows/release-monthly.yml
```diff
@@ -174,57 +186,67 @@ jobs:
...
-                  exit 0
+
+          wait_for_workflow() {
+            local workflow="$1"
+            local elapsed=0
+            echo "--- Waiting for $workflow ---"
+            while [ "$elapsed" -lt "$MAX_WAIT" ]; do
+              RUN_JSON="$(gh run list --workflow="$workflow" --limit 50 --json databaseId,status,conclusion,headSha,displayTitle 2>/dev/null || echo '[]')"
+              MATCHING_RUN="$(echo "$RUN_JSON" | jq -r --arg sha "$TAG_SHA" '.[] | select(.headSha == $sha) | @json' | head -1)"
+
```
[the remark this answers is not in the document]
DennisOSRM: I think this is valuable feedback. We need to make sure that we wait for the assets to be available.

.github/workflows/release-monthly.yml
```diff
@@ -174,57 +186,67 @@ jobs:
...
-          
-          echo "Waiting for CI workflow to complete for tag $TAG (commit $TAG_SHA)"
-          
-          MAX_WAIT=3600  # 1 hour
-          ELAPSED=0
+
+          WORKFLOWS="osrm-backend.yml osrm-backend-docker.yml"
+          echo "Waiting for workflows ($WORKFLOWS) to complete for tag $TAG (commit $TAG_SHA)"
+
+          MAX_WAIT=3600  # 1 hour; multi-arch arm64 under QEMU is slow
```
DennisOSRM: Not sure this suffices. The docker build take somewhere around five hours to complete. 

What we probably need to do is to start the docker build at the end of this run, but not to wait for it to complete.

DennisOSRM (approved):

DennisOSRM (changes requested):
Now, thinking about it a bit more, it could be solvable by a oneliner in `realease-monthly.yml`

```
gh workflow run osrm-backend-docker.yml --ref "${{ steps.version.outputs.tag }}" --field ref="{{ steps.version.output.tag }}"
```

That should dispatch the docker job without waiting for it, which is fine.

.github/workflows/release-monthly.yml
DennisOSRM: now, that we have native ARM builds for docker images, my previous comment is not relevant anymore. Retracting this conversation.

DennisOSRM:
Thanks again for the contribution. We have no generation of multi-arch docker images. If you feel using a PAT is the wrong way, please reopen and continue working on the changes.

PR #7481, opened 2026-04-14
ci: use native ARM machines in docker build

DennisOSRM:
good riddance QEMU


DennisOSRM:
Planning to merge this soon and then debug any issue in a point release before the next scheduled release is made.

PR #7485, opened 2026-04-16
feat: integrate https://github.com/nilsnolde/osrm-bindings

nilsnolde:
forgot to add here: @whytro did the real work during GSoC 2023. this PR is "just" integrating it into upstream.

getting to build linux was less of a hassle than I imagined. it's more of a hassle to review I guess.. my proposal: since it's mostly copy/paste from https://github.com/nilsnolde/osrm-bindings, it's not too bad to not review the code toooo much. one can somewhat trust the python packaging pipeline. I'll add more details later. IMO it's more important to understand how things work conceptually, so I'll add quite some more text to the PR description once this is a final PR. and of course feel free to ask any understanding question you have:) packaging is always awful, but honestly, python got really usable after they introduced (and widely supported) pyproject.toml.

**General**
- `nanobind` for C++ binding (i.e. no setup.py, all CMake based), very close to libosrm API, only some plumbing python code
- for PyPI distribution, we only build python 3.12 SABI (stable ABI) wheels, older version build from source

**Repo Layout**

<details>

<summary>everything is properly namespaced to python:</summary>

```shell
src/python/
├── CMakeLists.txt
├── README.md
├── include/
│   └── python/
│       ├── engineconfig_nb.hpp
│       ├── parameters/
│       ├── types/
│       └── utility/
├── osrm/
│   ├── __init__.py
│   ├── __main__.py
│   └── osrm_ext.pyi
└── src/
    ├── engineconfig_nb.cpp
    ├── osrm_nb.cpp
    ├── parameters/
    ├── types/
    └── utility/

test/python/
├── constants.py
├── test_index.py
├── test_match.py
├── test_nearest.py
├── test_route.py
├── test_table.py
├── test_tile.py
└── test_trip.py
``` 

</details>

**Packaging Architecture**
- we distribute both sdist & wheels from release-monthly.yml
- `cibuildwheel` to orchestrate the build, packaging, repairing (mostly vendor dynamic libs into the wheel), can be run locally too
- cibuildwheel does the build directly on the system for macos & win; however, for linux it needs an old-ish glibc version which is the ABI compatibility barrier for python extensions. this project uses a `manylinux_2_28` docker image based on almalinux 8. I forked https://github.com/pypa/manylinux a while ago to https://github.com/nilsnolde/manylinux, where I added the OSRM build dependencies for both aarch64 & x86: https://github.com/nilsnolde/manylinux/pkgs/container/manylinux.
- we auth with PyPI over a "trusted publisher" so no secrets needed
- versioning & publishing is integrated with `release-monthly.yml`

**Open Questions**
1. what should we do about the package name? `osrm-bindings` is ok-ish, `osrm` would of course be much nicer:)
1. we should probably open a osrm account on PyPI?
2. is it ok to transfer https://github.com/nilsnolde/osrm-bindings/issues/40 here and have someone from the community solve the win map-matching failure? I really don't feel like going back to win and figure out how to properly debug there;)


docs/.vitepress/config.js
```diff
@@ -14,7 +14,11 @@ export default defineConfig({
     nav: [
       { text: 'Home', link: '/' },
       { text: 'HTTP API', link: '/http' },
-      { text: 'Node.js API', link: '/nodejs/api' }
+      { text: 'Node.js API', link: '/nodejs/api' },
+      { text: 'Python', items: [
+        { text: 'API', link: '/python/api' },
+        { text: 'Development', link: '/python/development' }
```
nilsnolde, line 20: I tried to be really thorough, so I had claude add some python development and API docs. python doesn't work very well with vitepress with pulling out docstrings etc. it's mostly `mkdocs` these days in python land. not sure, but I could imagine mkdocs support JS docstrings as well. I guess you just switched to vitepress @DennisOSRM , not sure you'd be open to ditch it again potentially?

nilsnolde: for reference, this is the mkdocs based documentation I configured for valhalla: https://valhalla.github.io/valhalla/

src/python/README.md
```diff
@@ -0,0 +1,68 @@
+# osrm-bindings
+
+**Python bindings for [osrm-backend](https://github.com/Project-OSRM/osrm-backend) using [nanobind](https://github.com/wjakob/nanobind).**
```
nilsnolde, line 3: this file is the project page on pypi.org

test/data/windows-build-test-data.bat
```diff
@@ -0,0 +1,44 @@
+@ECHO OFF
+SETLOCAL EnableDelayedExpansion
+
+SET DATA_DIR=%CD%
```
nilsnolde, line 4: could probably use the standard Makefile instead of duplicating the same logic here. I know there gmake for windows, could just work..

DennisOSRM:
The PR probably needs a rebase onto latest master to pick up a fix for macOS ci.

nilsnolde:
jeez, thanks, was just about to diagnose!

CMakeLists.txt
```diff
@@ -489,8 +490,18 @@ add_dependency_includes(${ZLIB_INCLUDE_DIRS})
...
+# header directory and breaks libc++'s <cstddef> guard on macOS.
+set(_osrm_system_includes ${DEPENDENCIES_INCLUDE_DIRS})
+if(CMAKE_CXX_IMPLICIT_INCLUDE_DIRECTORIES)
+  list(REMOVE_ITEM _osrm_system_includes ${CMAKE_CXX_IMPLICIT_INCLUDE_DIRECTORIES})
+endif()
+if(CMAKE_C_IMPLICIT_INCLUDE_DIRECTORIES)
+  list(REMOVE_ITEM _osrm_system_includes ${CMAKE_C_IMPLICIT_INCLUDE_DIRECTORIES})
+endif()
+include_directories(SYSTEM ${_osrm_system_includes})
+unset(_osrm_system_includes)
```
nilsnolde: this is a bit contrived.. basically there was duplication of `-isystem` paths which led to a config-time reordering of include paths ending in compile error bcs it'd first look into the SDK include headers instead of the /usr/include system one. that's bcs find_package(zlib) etc resolve to the SDK's include and implicitly add/duplicate `-isystem` entries to DEPENDENCIES_INCLUDE_DIRS, reordering priority.

.github/workflows/osrm-backend.yml
```diff
@@ -24,7 +24,10 @@ concurrency:
 jobs:
   conan-windows-release-node:
     needs: format-taginfo-docs
-    runs-on: windows-2025
+    strategy:
+      matrix:
+        os: [windows-2025]
+    runs-on: ${{ matrix.os }}
```
nilsnolde, line 38: just so we can use `${{ matrix.os }}` in the cache key, for when we update the runner OS

.github/workflows/osrm-backend.yml
```diff
@@ -43,7 +46,16 @@ jobs:
...
       - run: npm install --ignore-scripts
       - run: npm link --ignore-scripts
+      - name: Restore Conan cache
+        id: conan-restore
+        uses: actions/cache/restore@v5
+        with:
+          path: ~/.conan2
+          key: v11-conan-${{ matrix.os }}-${{ hashFiles('conanfile.py') }}
+          restore-keys: |
+            v11-conan-${{ matrix.os }}-
```
nilsnolde: restore/save conan cache for win

.github/workflows/osrm-backend.yml
```diff
@@ -91,6 +109,20 @@ jobs:
...
           replacesArtifacts: true
           token: ${{ secrets.GITHUB_TOKEN }}
+      - name: Run CIBuildWheel
+        if: github.event_name != 'pull_request'
+        uses: pypa/cibuildwheel@v3.4.0
+        with:
+          output-dir: dist
+        env:
+          CIBW_CONFIG_SETTINGS_WINDOWS: "cmake.define.ENABLE_CONAN=ON"
+      - name: Upload Windows wheel
```
nilsnolde, line 129: release-monthly.yml publishes the wheels

.github/workflows/osrm-backend.yml
```diff
@@ -345,21 +376,21 @@ jobs:
             ENABLE_LTO: OFF
 
           - name: conan-macos-x64-release-node
-            build_node_package: true
+            build_bindings: true
             continue-on-error: true
             node: 24
-            runs-on: macos-26-intel # x86_64
+            runs-on: macos-15-intel # x86_64
```
nilsnolde, line 400: the python bindings (and the node ones as well maybe? didn't look) prefer the oldest OS version we can build on, so the widest range of OSs are compatible (same story for linux, where we build on glibc 2.28, current 2.43).

should I create a new job for macos-26 so we cover both min & max? (also creates more CI tension for concurrent branches)

.github/workflows/osrm-backend.yml
```diff
@@ -401,20 +436,22 @@ jobs:
           key: ${{ runner.os }}-node-${{ hashFiles('package-lock.json') }}
           restore-keys: |
             ${{ runner.os }}-node-
-      - name: Enable compiler cache
-        uses: actions/cache@v5
+      - name: Restore compiler cache
+        id: ccache-restore
+        uses: actions/cache/restore@v5
```
nilsnolde, line 455: I split these up into restore/save to have more control which cache doesn't get uploaded (unsuccessful build stage)

.github/workflows/osrm-backend.yml
```diff
@@ -401,20 +436,22 @@ jobs:
...
             ccache-${{ matrix.name }}-
-      - name: Enable Conan cache
-        uses: actions/cache@v5
+      - name: Restore Conan cache
+        id: conan-restore
+        uses: actions/cache/restore@v5
         with:
           path: ~/.conan2
-          key: v10-conan-${{ matrix.name }}-${{ github.sha }}
+          key: v11-conan-${{ matrix.name }}-${{ hashFiles('conanfile.py') }}
```
nilsnolde: conan shouldn't build fresh on every PR, only when relevant files change

.github/workflows/osrm-backend.yml
```diff
@@ -484,17 +521,15 @@ jobs:
...
           fi
 
-          # TBB
-          TBB_VERSION=2021.12.0
+          # TBB (Linux only; macOS gets it from conan or brew)
           if [[ "${RUNNER_OS}" == "Linux" ]]; then
+            TBB_VERSION=2021.12.0
             TBB_URL="https://github.com/oneapi-src/oneTBB/releases/download/v${TBB_VERSION}/oneapi-tbb-${TBB_VERSION}-lin.tgz"
-          elif [[ "${RUNNER_OS}" == "macOS" ]]; then
-            TBB_URL="https://github.com/oneapi-src/oneTBB/releases/download/v${TBB_VERSION}/oneapi-tbb-${TBB_VERSION}-mac.tgz"
```
nilsnolde: installed with brew, so macos-15 is actually true. these releases were build against another macos version. for linux it doesn't matter because the manylinux docker image contains the deps like tbb to build the python extension. only on macos & win they build on the host

.github/workflows/osrm-backend.yml
```diff
@@ -505,14 +540,9 @@ jobs:
           fi
           echo "CC=${CCOMPILER}" >> $GITHUB_ENV
           echo "CXX=${CXXCOMPILER}" >> $GITHUB_ENV
-          if [[ "${RUNNER_OS}" == "macOS" ]]; then
-            # missing from GCC path, needed for conan builds of libiconv, for example.
-            sudo xcode-select --switch /Library/Developer/CommandLineTools
-            echo "LIBRARY_PATH=${LIBRARY_PATH}:/Library/Developer/CommandLineTools/SDKs/MacOSX.sdk/usr/lib" >> $GITHUB_ENV
-            echo "CPATH=${CPATH}:/Library/Developer/CommandLineTools/SDKs/MacOSX.sdk/usr/include" >> $GITHUB_ENV
```
nilsnolde, line 474: this created all kinds of problems with reproducible tool chain setups. I tried removing it and the problems went away.

.github/workflows/release-monthly.yml
```diff
@@ -228,9 +233,42 @@ jobs:
...
+        uses: actions/download-artifact@v4
+        with:
+          pattern: wheels-*
+          path: dist
+          merge-multiple: true
+          run-id: ${{ needs.release.outputs.run_id }}
+          github-token: ${{ github.token }}
+
+      - name: Publish to PyPI
+        uses: pypa/gh-action-pypi-publish@v1.13.0
```
nilsnolde, line 271: this action only works on a linux runner, so it's another job syncing the wheel artifacts from the osrm-backend.yml run's upload

docs/python/api.md
```diff
@@ -0,0 +1,243 @@
+# Python API
```
nilsnolde, line 1: there's no real support it seems for python docstrings -> vitepress. I added a line in AGENT.md to keep this doc up-to-date when there's python API changes

src/python/osrm/__main__.py
nilsnolde, line 1: this file makes sure one can do `python -m osrm contract`

src/python/osrm/osrm_ext.pyi
nilsnolde, line 1: this is generated at cmake configure time via a nanobind cmake function and committed to the git tree so e.g. docs don't need to compile the extension before knowing the python api. CI checks whether this file needs an update based on the PRs API (non)changes.

.github/workflows/osrm-backend.yml
```diff
@@ -627,13 +674,111 @@ jobs:
...
+      - uses: actions/checkout@v6
+      - uses: dorny/paths-filter@v3
+        id: changes
+        with:
+          filters: |
+            python_cpp:
+              - 'src/python/src/**'
+              - 'src/python/include/**'
+              - 'src/python/CMakeLists.txt'
+      - uses: actions/setup-python@v6
```
nilsnolde, line 711: only run the actual check if this file filter is true for changed files

.github/workflows/osrm-backend.yml
```diff
@@ -627,13 +674,111 @@ jobs:
...
+
   ci-complete:
     runs-on: ubuntu-latest
-    needs: [build-matrix, conan-windows-release-node, docker-image-matrix]
+    needs: [build-matrix, conan-windows-release-node, docker-image-matrix, check-python-stubs]
+    if: ${{ !cancelled() }}
     steps:
+      - name: Fail if any dependency failed
+        if: ${{ contains(needs.*.result, 'failure') }}
+        run: exit 1
```
nilsnolde: not sure why claude added this, but doesn't seem to hurt

src/python/CMakeLists.txt
```diff
@@ -0,0 +1,106 @@
...
+find_package(Python
+  REQUIRED COMPONENTS Interpreter Development.Module
+  OPTIONAL_COMPONENTS Development.SABIModule
+)
+
+# Find nanobind's cmake dir via the Python module
+execute_process(
+  COMMAND "${Python_EXECUTABLE}" -m nanobind --cmake_dir
+  OUTPUT_STRIP_TRAILING_WHITESPACE OUTPUT_VARIABLE NB_DIR
+)
```
nilsnolde, line 14: it's neat that we don't have to submodule (or vcpkg or whatever) nanobind

nilsnolde: note, if ENABLE_PYTHON_BINDINGS=ON locally and outside of a `pip`-ish context where nanobind gets auto-installed before cmake configures, it needs to be installed on the host somehow (either as a python package or system wide)

.pre-commit-config.yaml
```diff
@@ -0,0 +1,9 @@
+repos:
```
nilsnolde, line 1: this is harmless for most people. I really like `pre-commit`, avoids the loop over ci when triggered before committing locally

CMakeLists.txt
```diff
@@ -150,6 +151,22 @@ include(CheckCXXCompilerFlag)
...
+# instantiation diagnostics from headers under that prefix. Force -isystem
+# via add_compile_options so brew-installed boost/lua/tbb headers are
+# treated as system headers regardless of implicit status.
+if(APPLE)
+  foreach(_brew_prefix /usr/local /opt/homebrew)
+    if(EXISTS "${_brew_prefix}/include")
+      add_compile_options(SHELL:-isystem ${_brew_prefix}/include)
+    endif()
+  endforeach()
+endif()
```
nilsnolde, line 150: making sure all brew headers are `-isystem`

pyproject.toml
```diff
@@ -0,0 +1,103 @@
...
+  "pytest {project}/test/python"
+]
+
+[tool.cibuildwheel.linux]
+environment = "LD_LIBRARY_PATH=/usr/local/lib64:${LD_LIBRARY_PATH}"
+before-build = "ccache -s && ccache -M 500M"
+repair-wheel-command = "auditwheel repair -w {dest_dir} {wheel}"
+
+[tool.cibuildwheel.macos]
+environment = "MACOSX_DEPLOYMENT_TARGET=15.0 CCACHE_DIR=$HOME/.ccache"
```
nilsnolde: we build the wheels on macos 15.0 to easily satisfy the target platform, so that brew libs are compatible

nilsnolde:
@DennisOSRM this PR is final now. I could've sworn I added some links in the issue description. I ran the final state with CI to simulate a monthly release and it published to https://test.pypi.org/project/osrm-bindings/. only linux got published there bcs of a mismatch when publishing, that's just kind of a testing bug. it'll work fine in real life.

still some open questions:
 
> Open Questions
> - what should we do about the package name? osrm-bindings is ok-ish, osrm would of course be much nicer:)
> - we should probably open a osrm account on PyPI?
> - is it ok to transfer https://github.com/nilsnolde/osrm-bindings/issues/40 here and have someone from the community solve the win map-matching failure? I really don't feel like going back to win and figure out how to properly debug there;)

DennisOSRM:
Cool. I have this on my plate for the weekend to review.

.github/workflows/osrm-backend.yml
DennisOSRM: this probably needs a tweak now that we have switched to vcpkg.

.github/workflows/osrm-backend.yml
```diff
@@ -56,6 +68,12 @@ jobs:
 
           cmake -DCMAKE_BUILD_TYPE=Release -DENABLE_CONAN=ON -DENABLE_NODE_BINDINGS=ON ..
           cmake --build . --config Release
+      - name: Save Conan cache
```
DennisOSRM: similarly needs a vcpkg related change

.github/workflows/osrm-backend.yml
```diff
@@ -119,6 +151,11 @@ jobs:
         PR_TITLE: ${{ github.event.pull_request.title }}
       run: |
         node ./scripts/check_pr_title.js
+    - uses: actions/setup-python@v6
+      with:
+        python-version: '3.12'
```
DennisOSRM, line 161: is this resembling the minimally supported Python version?

.github/workflows/osrm-backend.yml
```diff
@@ -345,21 +381,21 @@ jobs:
             ENABLE_LTO: OFF
 
           - name: conan-macos-x64-release-node
-            build_node_package: true
+            build_bindings: true
             continue-on-error: true
             node: 24
-            runs-on: macos-26-intel # x86_64
+            runs-on: macos-15-intel # x86_64
```
DennisOSRM, line 400: we should probably need an additional job for this

.github/workflows/osrm-backend.yml
```diff
@@ -345,21 +381,21 @@ jobs:
...
             ENABLE_ASSERTIONS: ON
             ENABLE_CONAN: ON
 
           - name: conan-macos-arm64-release-node
-            build_node_package: true
+            build_bindings: true
             continue-on-error: true
             node: 24
-            runs-on: macos-26 # arm64
+            runs-on: macos-15 # arm64
```
DennisOSRM, line 410: similarly the macos-26 job should stay unchanged

DennisOSRM (commented):
The change looks good. This PR needs a rebase to reflect the most recent changes of the build system. We should be good to go then.

.github/workflows/osrm-backend.yml
nilsnolde: yeah I was hoping I'd be quicker than the vcpkg PR, but wasn't ;)

.github/workflows/osrm-backend.yml
nilsnolde: no, that's in the pyproject.toml. cibuildwheel builds the package. the wheels are only 3.12+, but any >=3.10 should be able to build from source/sdist.

nilsnolde:
oh boy, vcpkg sucks for linux python packaging (well, really c++ python packaging for linux is what sucks..) :disappointed: since vcpkg builds from source it needs to match the glibc etc versions which are used to build the bindings (and OSRM for the bindings). that happens in my [manylinux](https://github.com/nilsnolde/manylinux) image. ideally we'd somehow mount the vcpkg cache there bcs rebuilding is costly and completely unnecessary.

this will need a bit more thought and I currently don't have the patience/time. I'll get back to this in a few days, max weeks, when there's more time.

nilsnolde:
I have a really hard time letting things go ;) I didn't want to "waste" the changes and context I had already, so I continued until it worked. the main problem was that the manylinux image now ideally already ships the pre-built vcpkg binaries via local cache. the problem with vcpkg in this scenario is that it's highly allergic against the slightest differences when deciding whether there's an appropriate cache or not. so everything, really everything has to be the exact same when 1) building the vcpkg packages in the image and then later 2) use that image to build the OSRM bindings in this PR (where we naturally go through the vcpkg config again and hoping to find a vcpkg binary cache). 

anyways, all that means in practice that the [manylinux image hard-codes](https://github.com/nilsnolde/manylinux/blob/6aed21c7d04e6260f6f511001b54253ce6c424a1/.github/workflows/build.yml#L11-L14) the osrm-backend version when publishing the image. that's needed so we can get the right vcpkg config files from osrm-backend. eventually once vcpkg config changes in a backward incompatible way that PR will fail the python binding CI workflows. all we need to do is either merge a separate PR with only vcpkg changes, or simply use the PR's HEAD to build a new image (though there'll be some problems there too, bcs currently I use a single docker tag which gets overwritten, so I don't have to touch the pyproject.toml).

you see, this is proper python, a snake eating its own tail! so painful..

.github/workflows/osrm-backend.yml
```diff
@@ -48,9 +51,9 @@ jobs:
         uses: actions/cache@v5
         with:
           path: ${{ github.workspace }}/.vcpkg-bincache
-          key: vcpkg-${{ runner.os }}-${{ runner.arch }}-${{ hashFiles('vcpkg.json', 'vcpkg-configuration.json', 'vcpkg-overlay-ports/**') }}
```
nilsnolde, line 52: was copy/paste error I guess

.github/workflows/osrm-backend.yml
```diff
@@ -327,7 +352,6 @@ jobs:
             NODE_PACKAGE_TESTS_ONLY: ON
 
           - name: vcpkg-linux-debug-node
-            build_node_package: true
```
nilsnolde, line 328: is there a reason why debug was used to build the node package?

.github/workflows/osrm-backend.yml
```diff
@@ -356,7 +379,6 @@ jobs:
             ENABLE_ASSERTIONS: ON
 
           - name: vcpkg-macos-arm64-release-node
-            build_node_package: true
```
nilsnolde, line 357: both node & python packages are now built on lower macos-15. for python it's kinda necessary for ABI compat, not sure about node, but it can't hurt either.

nilsnolde:
@DennisOSRM this is ready for review again. I'll just have to revert the last commit before merging, where I'm testing if the wheels build properly, without uploading to pypi.

as for the "open questions": maybe it's best to continue releasing under my account. it's already on pypi and it's the most frictionless path IMO. pypi doesn't have the concept of "org" namespaces as e.g. npm does (otherwise we wouldn't have the mess with `osrm` pypi packages). but anyone wanting to be a co-admin, let me know, at least pypi lets me invite other admins for single packages.

nilsnolde:
I also wanna say: in effect osrm is not buildable anymore with just system packages in many cases. since it removed all vendored packages and e.g. brew removed `sol2` support, at least on macos vcpkg is a must. also on my manjaro linux `sol2` is not in the official repos, only on the AUR (which is generally a bad habit to use). most others only have it on the very newest version like `sid`, `rawhide` & `26.06`: https://pkgs.org/search/?q=sol2.

even if in theory vcpkg might be optional in cmake, in practice it's not on the majority of platforms. IMHO it was nicer before. what wasn't nice is that is was hard-vendored instead of submodule'd (in terms of maintenance). there's a way to have submodules (at least for header-only deps) to fall back to. I'm really more OSS than FOSS, but in such cases I'm all for "free". personally I really don't like being kinda locked in to smth like vcpkg (still, it's MS) without a hassle.

that's just my 2c from the packaging experience of the python bindings.

nilsnolde:
was mostly battling x64 macos in the past commits. now everything's working and all wheels are uploaded as artifacts: https://github.com/Project-OSRM/osrm-backend/actions/runs/25164784497. 

next I wanna try linux aarch64, but after merging:)

.github/workflows/osrm-backend.yml
DennisOSRM: Not sure. Might have been an oversight

docs/.vitepress/config.js
DennisOSRM: Sure, we can discuss the feasibility of switching. Perhaps independent of this PR, but whatever we choose should be visually on par with vitepress.

docs/python/development.md
```diff
@@ -0,0 +1,341 @@
...
+```
+
+## Platform-specific build requirements
+
+### Linux
+
+CI wheel builds run inside a custom manylinux image
+([nilsnolde/manylinux](https://github.com/nilsnolde/manylinux), branch
+`osrm_python`) that ships vcpkg pre-bootstrapped at the SHA pinned in
+`vcpkg-configuration.json`, plus a pre-warmed vcpkg binary cache compiled
```
DennisOSRM, line 47: What happens when the vcpkg pin changes? Will that incur a slow build of the many linux image once?

test/data/windows-build-test-data.bat
DennisOSRM: Perhaps after merging this change 😉

DennisOSRM (approved):
Great work!

docs/python/development.md
nilsnolde: mostly right. the barrier is basically whether any vcpkg update in upstream osrm-backend (only new version for whatever reason, new package etc) would break the OSRM compilation or python tests with the previous vcpkg still present in the previous manylinux image. if it does, the PR workflow will fail, I'll need to `workflow_dispatch` with the osrm-backend commit sha introducing that vcpkg change (i.e. build a new tagged image). if it doesn't, we don't have to care much about them being not in sync currently.

it's a bummer but kinda fair. the python builds failing won't be too obnoxious. they're the last steps to run, and only on a selection of jobs. 

I do feel responsible for this now, but I invited at least you @DennisOSRM as collaborator in case I'm not available. eventually we could move [the repo](https://github.com/nilsnolde/manylinux) to the Project-OSRM org?

nilsnolde:
thanks @DennisOSRM for the quick review. there was a master conflict, also changed a tiny thing in https://github.com/Project-OSRM/osrm-backend/pull/7485/commits/4ed0a957227b83c8ffd8a2bdf509ca05c39a0269.

DennisOSRM (approved):

nilsnolde:
urgh, are you open to a force merge (if you have to) after I resolve this conflict and merge master @DennisOSRM ? lots of conflicts :sweat_smile:

DennisOSRM:
Thanks so much for the contribution. Glad, we could get it in this weekend.

PR #7487, opened 2026-04-17
build: migrate from Conan + vendored deps to vcpkg manifest mode

tete17:
## Summary

Migrate osrm-backend's dependency management from the current three-path system (Conan 2.x + vendored `third_party/` sources + system packages) to a single **vcpkg manifest-mode** setup. One dependency path on every platform, vendored copies deleted, CMakeLists simplified.

### Why vcpkg over the status quo

The current setup has three parallel dependency paths that have diverged over time:

- **Conan** for Windows/macOS CI, with `ENABLE_CONAN` gating a separate CMakeLists branch
- **System packages** (`apt-get`, `apk`) for Linux CI and Docker, with custom Find modules
- **Vendored sources** in `third_party/` (flatbuffers, fmt, sol2, rapidjson, protozero, libosmium) as a fallback

This creates maintenance burden: dependency version bumps require touching 3+ places, the `if(ENABLE_CONAN)/else()` block in CMakeLists.txt has subtle differences (e.g., Lua 5.2 vs 5.4), and the vendored copies are perpetually stale.

**vcpkg manifest mode** collapses this into:
- `vcpkg.json` — single source of truth for all dependency versions
- `vcpkg-configuration.json` — pinned baseline for reproducibility
- `CMakePresets.json` — cross-platform build configurations
- One `find_package()` path in CMakeLists.txt — no more `ENABLE_CONAN` branching

### Why vcpkg specifically

- **Microsoft-backed, large port registry** — all OSRM dependencies (boost, tbb, expat, lua, flatbuffers, sol2, libosmium, etc.) are available
- **Manifest mode** pins exact versions per-project with a baseline commit, avoiding "works on my machine" drift
- **Binary caching** via GitHub Actions cache backend (`x-gha,readwrite`) — CI doesn't rebuild boost from source on every run
- **CMake-native** — the toolchain file integrates transparently with `find_package()`, no wrapper scripts needed
- **Cross-platform** — same workflow on Linux, macOS, Windows, and Docker with no platform-specific package install steps

### What changes

| Commit | Description |
|--------|-------------|
| 1. `build: add vcpkg manifest...` | `vcpkg.json` + `vcpkg-configuration.json` + `CMakePresets.json` |
| 2. `build: rewrite CMakeLists.txt...` | Single vcpkg-fed dep section, remove `ENABLE_CONAN`, add `cmake/FindOsmium.cmake` |
[...]

PR #7488, opened 2026-04-17
build: generate flatbuffers headers at build time

tete17:
## Summary

- Generate flatbuffers C++ headers at CMake build time from the `.fbs` schemas instead of committing pre-generated headers
- Remove the flatbuffers version pin (was 25.9.23) from `vcpkg.json` — any compatible version now works
- Delete the `generated/` directory (−1,800 lines of generated code)

### Why

The pre-committed `*_generated.h` files contain a `static_assert` checking for an exact flatbuffers version match. This means every flatbuffers upgrade requires manually re-running `flatc` and committing the results — a maintenance burden and source of version drift. With build-time generation, upgrading flatbuffers is just a vcpkg baseline bump.

### How it works

Uses `flatbuffers_generate_headers()` from flatbuffers' own `BuildFlatBuffers.cmake` module (shipped by vcpkg). This creates:
- A custom target `GENERATE_osrm_fbs` that runs `flatc` when `.fbs` schemas change
- An INTERFACE library `osrm_fbs` whose include directory points to the generated output
- Headers land in `${CMAKE_BINARY_DIR}/osrm_fbs/engine/api/flatbuffers/` — same include path as before, no source changes needed

> **Note:** This PR is stacked on #7487 (vcpkg migration). Should be merged after that PR lands.

## Test plan

- [x] CLion build with flatbuffers 25.12.19 (unpinned) — all 5 headers generated, full build succeeds
- [x] Generated headers contain version 25.12.19 assertions (was 25.9.23), confirming the pin is gone
- [ ] CI green

🤖 Generated with [Claude Code](https://claude.com/claude-code)


nilsnolde:
is there a fallback in case BuildFlatBuffers.cmake in not available?

tete17:
Hey @nilsnolde not realy but I wouldn't worry.

Any distribution of flatbuffers includes this cmake file https://packages.debian.org/trixie/amd64/libflatbuffers-dev/filelist#:~:text=/usr/lib/x86_64%2Dlinux%2Dgnu/cmake/flatbuffers/BuildFlatBuffers.cmake

It is required as many projects depend on it to autogenerate the targets based on their definitions. You probablly have it in your system if you install flatbuffers through your package manager

nilsnolde:
ah perfect, thanks @tete17 ! yeah, I was mainly worried about debian like distros. hmm, seems bookworm doesn't have it: https://packages.debian.org/bookworm/amd64/libflatbuffers-dev/filelist. it's EOL in a few months. since trixie is just a few weeks old, I'd strongly suggest to introduce a fall back.

DennisOSRM:
If this is pulled from vcpkg then there shouldn't be an issue on Debian bookworm, right? Or am I missing something here?

nilsnolde:
IMO it's a bit nicer generally to not _require_ vcpkg installations (always builds from source last I checked) and instead be able to use binary system packages for dependencies (except for win of course). I didn't really have the time to take a deeper look, I asked [here](https://github.com/Project-OSRM/osrm-backend/pull/7487#issuecomment-4270496719) if vcpkg is optional or required, but didn't get an answer yet.

tete17:
Hey @nilsnolde sorry for ignoring the questions.

So regarding bookworm I have a strong suspecious, but I can't test it that, the flatbuffers team just restructure the cmake files and are now in a different place. If we wanted to support a debian based system based on bookworm (Ubutnu 22.04 for that matter as well) we would need to introduce an if clause in the cmake to include the right folder.

Then again this would be to support an EOL operating system for native packages when they could just turn on the vcpkg integration and not worry about anything. Yes the initial cmake configure is annoyiong as it would take 1h in a medium system not going to lie but is a worth price to pay to ensure everybody uses the same library versions.

DennisOSRM:
I wouldn't worry too much about bookworm personally. It's ~~EOL~~ in this long term support phase in six weeks and since the beginning we had the policy of supporting the latest debian release only.

nilsnolde:
yeah sorry, I read the trixie release wrong. and since vcpkg remains optional (as conan was), there's no real problem anyways.

DennisOSRM:
Once #7487 is done, this needs a rebase. One idea to make this reviewable independently could be to use the other PR's branch as the base. And then after it's done to change the base branch to master.

DennisOSRM (approved):

PR #7504, opened 2026-04-27
fix: invalid TurnLaneData comparators

### LINKED ISSUE #7503: Comparator of `TurnLaneData` violates Strict Weak Ordering
# Issue

**Note: this issue is not targeting a real-world bug that actually happen, but raise an potential Undefined Behavior issue existing in the current codebase**

`TurnLaneData::operator<` has a standards-compliance issue in its `uturn` special-case handling: the comparator doesn't always satisfy the strict weak ordering required by `std::sort`.

https://github.com/Project-OSRM/osrm-backend/blob/105448a80ef44ca4c5e7c3ad93b5e83e89b4a9bc/src/guidance/turn_lane_data.cpp#L43-L57

In terms of the algorithm's contract, the comparator used by `std::sort` must induce a strict weak ordering. In particular:

- For all `a`, `comp(a, a) == false`.
- If `comp(a, b) == true` then `comp(b, a) == false`.

The current `uturn` branch violates these rules for entries with `tag == uturn` and `from == 0`:

- `comp(a, a)` can return `true`.
- Two equal right-edge `uturn` entries can compare `true` in both directions.

This means the comparator is not standards-compliant and may trigger undefined behavior during sorting.

Even if equal `TurnLaneData` `uturn` entries are unlikely or impossible to appear in a `LaneDataVector` under current construction paths, the comparator is still incorrect by definition. If such equal entries never occur in practice, the bug may stay latent, but the ordering relation remains invalid.

References:

- [C++ reference, C++ named requirements: Compare](https://en.cppreference.com/cpp/named_req/Compare)

qwe854896:
# Issue

Fix #7503

## Tasklist

 - [X] self-review code for correctness and following the [coding guidelines](https://github.com/Project-OSRM/osrm-backend/wiki/Coding-Standards)
 - [ ] add tests (see [testing](https://github.com/Project-OSRM/osrm-backend/blob/master/docs/testing.md))
 - [ ] update relevant [wiki pages](https://github.com/Project-OSRM/osrm-backend/wiki)
 - [ ] review
 - [ ] adjust for comments

## Requirements / Relations

No additional requirements.
Not based on any other pull request.

## Implementation Notes

Instead of touching the existing special case handling logic, I add an equality guard right after `from`/`to` tie comparison and before the original `tag` comparison.

### DIFF
--- a/src/guidance/turn_lane_data.cpp
+++ b/src/guidance/turn_lane_data.cpp
@@ -25,6 +25,9 @@ bool TurnLaneData::operator<(const TurnLaneData &other) const
     if (to > other.to)
         return false;
 
+    if (tag == other.tag)
+        return false;
+
     // the suppress-assignment flag is ignored, since it does not influence the order
     const constexpr TurnLaneType::Mask tag_by_modifier[] = {TurnLaneType::sharp_right,
                                                             TurnLaneType::right,


DennisOSRM (approved):
looks good. thanks for the contribution

PR #7506, opened 2026-04-27
refactor: drop boost dependencies replaceable by C++20 stdlib

tete17:
## Summary

Now that the codebase targets C++20, several Boost components have direct standard-library equivalents. This PR drops 11 Boost components from `vcpkg.json` (`boost-core`, `-dynamic-bitset`, `-foreach`, `-io`, `-mpl`, `-optional`, `-range`, `-serialization`, `-thread`, `-uuid`, `-variant`) by migrating each call site to its stdlib counterpart.

The work is split into 8 self-contained commits, one per component group, so each can be reviewed (or reverted) independently:

- `boost-uuid` → unused, removed outright
- `boost-core` / `-foreach` / `-optional` / `-mpl` → stdlib equivalents
- `boost::variant` → `std::variant`
- `boost-thread` (`shared_mutex`) → `std::shared_mutex` in `data_watchdog`
- `boost-io` (`ios_flags_saver`) → small local RAII shim
- `boost::dynamic_bitset` → `std::vector<bool>`
- `boost::range` → `std::ranges` / `std::views`
- `boost::starts_with` / `ends_with` / `contains` → native `std::string(_view)` members

Net diff is ~140 lines added / ~170 removed across 28 files.

Depends on #7487 (vcpkg migration) which provided the C++20 baseline.

## Test plan

- [ ] CI green on all platforms (Linux, macOS, Windows)
- [ ] Unit tests pass
- [ ] Cucumber feature tests pass
- [ ] No behavior changes — refactor only


DennisOSRM:
GCC 12 is the limiting factor most probably. The plan is to drop support for it (and also version 13) once GitHub action runners based on Ubuntu 26.04 becomes available. 

Debian Trixie ships GCC 14 und Ubuntu comes with 15.

include/engine/routing_algorithms/routing_base_mld.hpp
```diff
@@ -360,18 +353,12 @@ void relaxOutgoingEdges(const DataFacade<Algorithm> &facade,
             const auto &cell =
                 cells.GetCell(metric, level, partition.GetCell(level, heapNode.node));
             auto source = cell.GetSourceNodes().begin();
-            auto distance = [&cell, node = heapNode.node]() -> auto
+            auto distance = [&]() -> auto
             {
                 if constexpr (IS_MAP_MATCHING)
-                {
```
DennisOSRM, line 366: We should keep braces for all blocks

DennisOSRM (approved):

PR #7518, opened 2026-05-01
fix(api)!: prevent OSM node ID truncation in match/route annotations

### LINKED ISSUE #7069: Node IDs > 2^34 are reported by route matching incorrect.

timpara:
# Issue

Fixes #7069. OSM node IDs above 2^34 come back wrong in /match and /route annotations.

## Root cause

Two truncation bugs on the node-ID path:

1. Storage: `PackedOSMIDs` packs OSM IDs into 34 bits, so any ID at or above 2^34 is silently masked. The `BOOST_ASSERT` guard is compiled out in release.
2. FlatBuffers: `Annotation.nodes` was `[uint]` and `route_api.hpp` used `std::vector<uint32_t>`, truncating IDs above 2^32 again on the FB path. JSON output was already 64-bit.

## Changes

- Widen `PackedOSMIDs` from 34 to 40 bits (max ~1.1 * 10^12).
- Add `checkPackedOSMNodeIdFits()` and call it at every push site so overflow throws during osrm-extract instead of masking.
- Change FB `Annotation.nodes` to `[ulong]`, use `std::vector<uint64_t>` in `route_api.hpp`, regenerate `fbresult_generated.js`.
- Add unit tests for round-trip above 2^34 and the overflow guard.
- Update `docs/http.md`.

## Breaking changes

1. On-disk: `.osrm.nbg_nodes` layout changes; existing datasets must be re-extracted. Recommend bumping `package.json` minor before release so the fingerprint rejects stale data.
2. FB wire: `Annotation.nodes` changed from uint32 to uint64. FB clients must update accessors.

## Notes

- clang-format was not available in my environment; please run `./scripts/format.sh` if needed.
- No cucumber feature added: the framework auto-assigns small IDs and reworking it for huge IDs is out of scope. Unit tests cover the storage regression; the FB change is exercised by the existing FB cucumber matrix.
- Vector tile output already uses `OSMNodeID::value_type` (uint64_t) and benefits transparently.


DennisOSRM:
Thanks for the contribution. Please consider adding a test that covers the changed behavior.

DennisOSRM:
This needs a formatting fix to run CI checks:

```
diff --git a/unit_tests/util/packed_vector.cpp b/unit_tests/util/packed_vector.cpp
index 0762c9b..99425bc 100644
--- a/unit_tests/util/packed_vector.cpp
+++ b/unit_tests/util/packed_vector.cpp
@@ -215,13 +215,13 @@ BOOST_AUTO_TEST_CASE(packed_osm_ids_above_2pow34_roundtrip)
     const std::vector<std::uint64_t> inputs = {
         0ull,
         1ull,
-        (1ull << 33),                                          // 2^33
-        (1ull << 34),                                          // 2^34, the old truncation boundary
-        (1ull << 34) + 1,                                      // just above 2^34
-        20'000'000'000ull,                                     // realistic projected OSM node ID
-        (1ull << 35),                                          // 2^35
-        osrm::extractor::MAX_PACKED_OSM_NODE_ID - 1,           // just below the new limit
-        osrm::extractor::MAX_PACKED_OSM_NODE_ID,               // exactly at the new limit
+        (1ull << 33),                                // 2^33
+        (1ull << 34),                                // 2^34, the old truncation boundary
+        (1ull << 34) + 1,                            // just above 2^34
+        20'000'000'000ull,                           // realistic projected OSM node ID
+        (1ull << 35),                                // 2^35
+        osrm::extractor::MAX_PACKED_OSM_NODE_ID - 1, // just below the new limit
+        osrm::extractor::MAX_PACKED_OSM_NODE_ID,     // exactly at the new limit
     };
 
     for (auto raw : inputs)
```

DennisOSRM:
continuing the work in #7552

DennisOSRM:
Thanks again for the contribution. Halfway through i broke something in my setup and had continued in an additional pr. 

Please come again with more patches. 😉

PR #7519, opened 2026-05-03
build: replace Boost.ProgramOptions with CLI11

jrklasen:
## Summary

Alternative to #7511 (the `boost::program_options` `validate()` shim for `std::filesystem::path`). That PR fixes the path-with-spaces truncation in 23 lines of Boost glue. This one fixes the same bug as a side effect of replacing `boost::program_options` entirely with [CLI11](https://github.com/CLIUtils/CLI11). 

Pick whichever scope fits the project's appetite—they are mutually exclusive. 

## What changes

- **Removes** `boost-program-options` (the only non-header-only Boost component still linked into every CLI tool) from `vcpkg.json`, `BOOST_COMPONENTS`, and the link line of all 6 binaries.
- **Adds** `cli11` as a vcpkg dep and `find_package(CLI11)` in CMake.
- **Rewrites** argument parsing in all 6 tools (`osrm-extract`, `osrm-contract`, `osrm-customize`, `osrm-partition`, `osrm-routed`, `osrm-datastore`) — ~700 lines of boost::po replaced with CLI11 equivalents, net **−300 lines**.
- **Drops** dead code unlocked by the migration: `MaxCellSizesArgument` struct, two custom `operator>>` overloads (`EngineConfig::Algorithm`, `storage::FeatureDataset`), and a `boost::validate()` overload for `--max-matching-radius unlimited`. The whole of `src/storage/storage_config.cpp` becomes empty and is removed.
- **Adds** a small helper header `include/util/cli_helpers.hpp` (~50 lines) with shared CLI11 transforms for the two enums and the `unlimited` double.
- **Adds** `scripts/ci/test_cli_parsing.sh` — bash integration tests covering the high-risk parsing surface, wired into CI.

The path-with-spaces bug is fixed without any workaround: CLI11 doesn't go through `operator>>` for `std::filesystem::path`.

## Why this over the shim

| | Shim PR | This PR |
|---|---|---|
| Path-with-spaces fixed | yes (workaround) | yes (root-cause; bug class eliminated) |
| Boost surface added | `boost::any`, `validators::*`, custom `validate()` | none — Boost.po removed entirely |
| Build deps | unchanged | `boost-program-options` → `cli11` (smaller, header-only) |
| Linker output | unchanged | every tool drops the Boost.po `.o` files |
| CLI parsing test coverage | none | ~25 cases via new script |
| Continues OSRM's "drop Boost where stdlib/lighter alternatives exist" trend | no | yes |
| Diff size | small | medium |

## Breaking changes

User-visible:
- **`--help` and error message format change** — CLI11 formats both differently than boost::po. Scripts that grep tool output for exact strings will break. Option names, short flags, and parsing semantics are unchanged.

API-visible:
- **`storage::operator>>(istream&, FeatureDataset&)` is removed** from `include/storage/storage_config.hpp`. It was only used by boost::po's auto-discovery; downstream consumers are unlikely but possible. This is a header-level signature change.

Build-visible:
- `boost-program-options` removed from `vcpkg.json`, `cli11` added. First build after merge re-fetches deps automatically; no user action.

Not breaking:
- Same option names, same short flags, same field bindings, same exit codes, same control flow for `--help` / `--version` / `--list-inputs`. Existing scripts that invoke the tools (including `run_benchmarks.sh` and the cucumber matrix) keep working.

## Test plan

CI runs the full battery on this PR:
- Clean build with vcpkg fetching `cli11` for the first time
- Unit tests (`make tests` + glob loop)
- New `./scripts/ci/test_cli_parsing.sh` — covers the algorithm enum (default, case-insensitive, garbage rejection), `--max-matching-radius unlimited`, `--disable-feature-dataset` multitoken, `--max-cell-sizes` comma-split + sort check, the path-with-spaces regression, and unknown-flag rejection
- `make -C test/data benchmark` — end-to-end extract → partition → customize → contract pipeline
- Cucumber and Node bindings tests

If CI is green, the parsing surface should be fully verified. The only flag not covered is `--disable-location-cache`, which has no observable parse-time effect—if it's miswired, the only consequence is that the locations cache is on/off when the user expected the opposite. That's a performance/memory regression, not a correctness one (routes are unaffected).

## Out of scope

- Adding config-file support (boost::po had it; CLI11 has it via TOML—OSRM doesn't use it today, not adding it now).
- Removing other Boost components.
- Refactoring the inconsistent return-type conventions across the 6 tools (`return_code` enum vs `bool` vs `INIT_*` constants).


nilsnolde:
TBH, I was also thinking I'd replace program_options with smth else. I did that in a few of my other repos, but there the motivation was to remove all boost compiled libs and only have header-only. I usually use header-only https://github.com/jarro2783/cxxopts. It's almost the same syntax as program_options, which IMO goes a long way, on top of avoiding more vcpkg compilation, it's just a header download. Why I didn't do it in the end for OSRM was that IMO it'd "fix" smth that's not broken. 

Now, smth is evidently broken in program_options :sweat_smile: Just my 2c that your https://github.com/Project-OSRM/osrm-backend/pull/7511 is a more elegant workaround. And if replacing the cli lib: did you consider cxxopts?

jrklasen:
No, I don't have any strong reason for CLI11; I'm just trying to see how breaking such a change would be. I'm also absolutely fine if this attempt gets rejected or superseded by another attempt.

DennisOSRM:
I have used cxxopts in the past as well and found it a good replacement. Right now (and please change my opinion), I am thinking we should go with #7511 since it's a quite surgical fix.

jrklasen:
Sounds good to me!

jrklasen:
Okay, I'll close this one then for now

PR #7523, opened 2026-05-04
build: post-vcpkg cleanup of dead helper files

tete17:
## Summary

Follow-up cleanup to the vcpkg migration. Three small, independent build-system commits that remove dead weight left behind once Conan + the `third_party/` vendored deps were retired.

- **Remove `scripts/update_dependencies.sh`** — the script managed git-subtree updates for vendored deps under `third_party/`. After #7487 (vcpkg migration), #7488 (flatbuffers from vcpkg), #7495 (microtar → libarchive), and #7508 (vtzero unvendored via custom vcpkg overlay), `third_party/` is empty and every entry the script knew how to update is now provided by vcpkg. Nothing left to manage — delete the file.
- **Remove unused CMake helper files** (737 lines deleted across 4 files):
  - `cmake/FindTBB.cmake` — unused since `find_package(TBB CONFIG REQUIRED)` resolves via the vcpkg `tbb` port's `TBBConfig.cmake` and bypasses Find modules entirely.
  - `cmake/FindLua.cmake` — an older copy of CMake's upstream module that shadowed the built-in via `CMAKE_MODULE_PATH`. The vcpkg `lua` port ships a `vcpkg-cmake-wrapper.cmake` that redirects `find_package(Lua)` to `unofficial-lua` CONFIG and populates `LUA_INCLUDE_DIR` / `LUA_LIBRARIES`. Verified by rebuilding `osrm_extract` (which compiles `scripting_environment_lua.cpp`) with the file removed.
  - `cmake/FindDebArch.cmake` — never `include()`d; `CPackConfig.cmake` already inlines the equivalent `dpkg --print-architecture` logic.
  - `cmake/cmake_options_script.py` — a legacy SublimeClang helper for `compile_commands.json` consumption, superseded by clangd/LSP.
- **Modernise `cmake_uninstall.cmake.in`** — replace constructs deprecated since CMake 3.0 / 3.17 (project already requires 3.18): `exec_program` → `execute_process`, `cmake -E remove` → `cmake -E rm -f`. Also fix a stringly-typed return-value check (`STREQUAL` on a stringified int → numeric `EQUAL`), drop an unused `OUTPUT_VARIABLE`, and remove the legacy duplicate-condition args on `else()`/`endif()`/`endforeach()` closers. No behaviour change.

Net diff: **+10 / -811** across 6 files.

## Test plan

- [ ] CI green on all platforms (Linux, macOS, Windows)
- [ ] `cmake --build && cmake --install` still produces the expected artefacts
- [ ] `make uninstall` on an installed tree still removes everything in `install_manifest.txt`


DennisOSRM (approved):

DennisOSRM:
Does deleting the find_package helpers have an impact on folks embedding OSRM in a larger CMake project?

tete17:
> Does deleting the find_package helpers have an impact on folks embedding OSRM in a larger CMake project?

Not really since those files were really not used.

